PR #6799, opened 2025-08-18
Allow renaming .map functions in workflows

bismitpanda:
## Description

Added a parameter to `.map` method of workflow to add a custom id to the mapping step

## Related Issue(s)

Fixes #6795 

## Type of Change

- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [ ] Test update

## Checklist


packages/core/src/workflows/workflow.ts
```diff
@@ -738,12 +738,13 @@ export class Workflow<
             | DynamicMapping<TPrevSchema, z.ZodType<any>>;
         }
       | ExecuteFunction<z.infer<TPrevSchema>, any, any, any, TEngineType>,
+    idConfig?: { id?: string | null },
```
abhiaiyer91: Maybe we should call this stepOptions

bismitpanda: I changed it

taofeeq-deru (approved):
Looks good. Thanks for tackling this.

PR #6838, opened 2025-08-19
chore: optimize prettier checks to only run on changed files

TylerBarnes:
Claude says:
This makes our prettier checks much faster by only checking files that have changed instead of the entire codebase.

The scripts are now composable, so you can run:
- `pnpm list-changed-files` to see all changed files
- `pnpm list-prettier-files` to filter for files prettier should format
- `pnpm prettier:changed` to format only changed files

The GitHub workflow now uses this faster approach too, which should speed up CI checks on PRs.

Also added a global `lint` command that was missing from the root package.json.


package.json
```diff
@@ -58,10 +58,14 @@
     "test:evals:llm": "pnpm --filter ./packages/evals test src/metrics/llm",
     "test:observability": "pnpm --filter ./observability/* test",
     "lint-staged": "lint-staged",
+    "lint": "turbo lint --filter \"./packages/**\" --filter \"./stores/**\" --filter \"./auth/**\" --filter \"./voice/**\" --filter \"./deployers/**\" --filter \"./workflows/**\" --filter \"./observability/**\" --filter \"./client-sdks/**\" --filter \"!@mastra/playground\"",
```
wardpeet: You probably want to do this instead
```suggestion
    "lint": "turbo --filter \"!./examples/**/*\" --filter \"!./docs/**/*\" --filter \"!./integrations/**/*\" --filter \"!@mastra/playground\" lint",
```

package.json
```diff
@@ -58,10 +58,14 @@
...
     "preinstall": "npx only-allow pnpm",
     "prepare": "husky",
-    "prettier:format": "prettier --write .",
+    "prettier:format": "prettier --write . --log-level warn",
     "prettier:format:check": "prettier --check .",
+    "list-changed-files": "(git diff --name-only --diff-filter=ACMR $(git symbolic-ref refs/remotes/origin/HEAD | sed 's@^refs/remotes/origin/@@')...HEAD && git diff --name-only --diff-filter=ACMR && git diff --cached --name-only --diff-filter=ACMR) | sort -u",
```
wardpeet: not sure if I trust this :D It takes about 44s in CI, not sure if it's worth the optimisation

TylerBarnes: that's fair :) mostly useful locally where it takes half a second instead of 20

abhiaiyer91 (approved):

PR #6925, opened 2025-08-21
fix types

wardpeet:


DanielSLew (approved):

packages/core/src/stream/aisdk/v5/compat.ts
```diff
@@ -356,6 +356,7 @@ export function prepareToolsAndToolChoice<TOOLS extends Record<string, Tool>>({
           if ('inputSchema' in tool) {
             inputSchema = tool.inputSchema;
           } else if ('parameters' in tool) {
+            // @ts-ignore
```
DanielSLew: maybe an explanation as to why the ts-ignore will be helpful

wardpeet: because it wan't working :D

wardpeet:
```suggestion
            // @ts-ignore tool is not part
```

PR #6944, opened 2025-08-22
zod v4

wardpeet:
Add support for Zod v4, keeping compatability with zod v3

- Add v4 test suite and update schema compat
- Use toJsonSchema from zod
- upgrade autoform v4


cian:
@wardpeet Unfortunately I think I'm seeing a regression after this PR affecting next.js app using zod v3 and *not* using turbopack.

I've filed a bug here: https://github.com/mastra-ai/mastra/issues/7112

PR #6954, opened 2025-08-22
[CLOUD-412] Add AI trace db schema and create base observability storage class

YujohnNattrass:
## Description

Creates db schema for AI spans.

Implements base observability storage class which defaults to throwing for each method.

Only create AI span tables if supported

## Type of Change

- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [ ] Test update

## Checklist


packages/core/src/storage/constants.ts
```diff
@@ -129,7 +137,29 @@ export const SCORERS_SCHEMA: Record<string, StorageColumn> = {
...
+  attributes: { type: 'jsonb', nullable: true },
+  metadata: { type: 'jsonb', nullable: true },
+  events: { type: 'jsonb', nullable: true },
+  links: { type: 'jsonb', nullable: true },
+  other: { type: 'text', nullable: true },
+  input: { type: 'jsonb', nullable: true },
+  output: { type: 'jsonb', nullable: true },
+  error: { type: 'jsonb', nullable: true },
+  startAt: { type: 'timestamp', nullable: false }, // When the span started
+  endAt: { type: 'timestamp', nullable: true }, // When the span ended
```
YujohnNattrass: @epinzur should `endAt` be nullable for realtime insertion?

epinzur: yes I think so.

packages/core/src/storage/constants.ts
```diff
@@ -7,15 +7,23 @@ export const TABLE_THREADS = 'mastra_threads';
...
   | typeof TABLE_RESOURCES
-  | typeof TABLE_SCORERS;
+  | typeof TABLE_SCORERS
+  /**
+   * TODO: Remove this once we settle on the AI Span storage schema/apis
+   * Table names with optional new tables for gradual migration.
+   * Use `TABLE_NAMES<true>` to include newer tables like AI_SPANS.
+   * Defaults to original tables only for backward compatibility.
+   */
+  | (IncludeAiSpans extends true ? typeof TABLE_AI_SPANS : never);
```
YujohnNattrass: @wardpeet We’re only implementing libsql and the mock in-memory store for AI spans until we’re confident in the schema.

To avoid making a bunch of code changes throughout the codebase to fix the type error below (missing table ai spans), I added this temporary workaround.

What do you think of this approach? Or should I just update all store providers to solve the type error? Another option would be to use a temporary type like TABLE_NAMES_INCLUDE_AI_SPANS as a migration table, apply it only to the supported stores, and then later update the name back to TABLE_NAMES

cc: @epinzur 

image

packages/core/src/storage/constants.ts
```diff
@@ -129,7 +137,29 @@ export const SCORERS_SCHEMA: Record<string, StorageColumn> = {
   },
 };
 
-export const TABLE_SCHEMAS: Record<TABLE_NAMES, Record<string, StorageColumn>> = {
+export const AI_SPAN_SCHEMA: Record<string, StorageColumn> = {
+  // Composite primary key of traceId and spanId
+  traceId: { type: 'text', nullable: false },
+  spanId: { type: 'text', nullable: false },
```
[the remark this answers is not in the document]
epinzur: i agree we still need an `id` field here, that is the primary key

packages/core/src/storage/types.ts
```diff
@@ -159,3 +160,37 @@ export interface ThreadSortOptions {
 export type ThreadOrderBy = 'createdAt' | 'updatedAt';
 
 export type ThreadSortDirection = 'ASC' | 'DESC';
+
+export interface AISpanRecord {
```
epinzur, line 164: might need `id` (primary key)

packages/core/src/storage/types.ts
```diff
@@ -159,3 +160,37 @@ export interface ThreadSortOptions {
...
+  spanId: string;
+  parentSpanId: string | null;
+  name: string;
+  scope: Record<string, any> | null;
+  spanType: AISpanType;
+  attributes: Record<string, any> | null;
+  metadata: Record<string, any> | null;
+  events: Record<string, any> | null;
+  links: Record<string, any> | null;
+  other: string | null;
```
epinzur: drop events & other

packages/core/src/storage/constants.ts
```diff
@@ -129,6 +131,28 @@ export const SCORERS_SCHEMA: Record<string, StorageColumn> = {
...
+  parentSpanId: { type: 'text', nullable: true },
+  name: { type: 'text', nullable: false },
+  scope: { type: 'jsonb', nullable: true }, // Mastra package info {"core-version": "0.1.0"}
+  spanType: { type: 'integer', nullable: false }, // WORKFLOW_RUN, WORKFLOW_STEP, AGENT_RUN, AGENT_STEP, TOOL_RUN, TOOL_STEP, etc.
+  attributes: { type: 'jsonb', nullable: true },
+  metadata: { type: 'jsonb', nullable: true },
+  events: { type: 'jsonb', nullable: true },
+  links: { type: 'jsonb', nullable: true },
+  other: { type: 'text', nullable: true },
```
epinzur: drop events & other

packages/core/src/storage/types.ts
```diff
@@ -159,3 +160,37 @@ export interface ThreadSortOptions {
...
+  startAt: Date;
+  endAt: Date | null;
```
epinzur: i think these should be past-tense: `startedAt` & `endedAt`

packages/core/src/storage/types.ts
```diff
@@ -159,3 +160,37 @@ export interface ThreadSortOptions {
...
+  createdAt: Date;
+  updatedAt: Date | null;
+  input: Record<string, any> | null;
+  output: Record<string, any> | null;
+  error: Record<string, any> | null;
```
epinzur: lets make input, output, error & links just `any` type.

epinzur: or something json blob compatible... 

they can be a `string` or a `number` or a `list` for example, which wont fit into `Record<string, any>`

epinzur (changes requested):
generally looks good... a few things to update

packages/core/src/storage/types.ts
YujohnNattrass: since the primary key is a composite key of traceId and spanId there won't be an ID. To double check I checked out what the return data would be in the libsql implementation and verified there isn't an id.

packages/core/src/storage/types.ts
YujohnNattrass: good point, will update it!

epinzur (approved):

PR #7014, opened 2025-08-26
Setup context relevance scorer

adeniyii:
## Type of Change

- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [ ] Test update

## Checklist


packages/evals/src/scorers/llm/context-relevance/index.ts
```diff
@@ -0,0 +1,164 @@
...
+      outputSchema: analyzeOutputSchema,
+      createPrompt: ({ run }) => {
+        const userQuery = getUserMessageFromRunInput(run.input) ?? '';
+        const agentResponse = getAssistantMessageFromRunOutput(run.output) ?? '';
+
+        // Get context either from options or extractor
+        const context = options.contextExtractor ? options.contextExtractor(run.input!, run.output) : options.context!;
+
+        if (context.length === 0) {
+          throw new Error('No context available for evaluation');
```
YujohnNattrass: not quite sure if we want to throw an error here. maybe we can default to 0 or 1 and give a reason that no context was available.

What do you think? if it makes more sense to throw an error, I'm okay with that as well

YujohnNattrass (approved):

packages/evals/src/scorers/llm/context-relevance/index.ts
adeniyii: ah good point, this is a better approach.
updating!

PR #7039, opened 2025-08-26
Initialize scorers on Mastra instance to override scorers

YujohnNattrass:
## Description

This PR introduces support for registering scorers on the Mastra instance. This allows users to override scorers when calling methods such as Agent.generate, Agent.stream, Agent.generateVnext, and others.

### Initializing Scorers

Scorers can be initialized directly on the Mastra instance:

```
const testScorer = createScorer({
  name: 'scorer1',
  description: 'Scorer 1',
}).generateScore(() => {
  return 1
})

  agents: {
    chefAgent,
  },
  scorers: {
    testScorer,
  },
  // ...
})
```

### Using Scorers in Agent.generate

When calling Agent.generate, you can provide a scorers option:
```
const testScorer = mastra.getScorer('testScorer')

const result = chefAgent.generate('hello', { 
  scorers: {
    scorer: testScorer,
    sampling: { ... },
  },
})
```

### Why Initialization Is Required

Scorers that need to be overridden must be initialized on the Mastra instance. This is because scorer execution is event-driven: if a scorer isn’t registered on the instance, it can’t be fetched or triggered during the agent’s lifecycle.

## Type of Change

- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [x] Test update

## Checklist


packages/core/src/mastra/hooks.ts
```diff
@@ -80,3 +69,22 @@ export function createOnScorerHook(mastra: Mastra) {
...
+  } else if (entityType === 'WORKFLOW') {
+    const scorers = await mastra.getWorkflowById(entityId).getScorers();
+    scorerToUse = scorers[scorerId];
+  }
+
+  // Fallback to mastra-registered scorer
+  if (!scorerToUse) {
+    const mastraRegisteredScorer = mastra.getScorer(scorerId);
+    scorerToUse = mastraRegisteredScorer ? { scorer: mastraRegisteredScorer } : undefined;
+  }
```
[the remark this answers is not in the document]
YujohnNattrass: This is incorrect, if they are found on the agent or workflow they come back as a MastraEntity with the following properties: `{ scorer: MastraScorer, sampling: SamplingConfig }` We don't have to worry about sampling because that would have a happened before this call

packages/core/src/agent/agent.test.ts
```diff
@@ -6073,6 +6085,161 @@ function agentTests({ version }: { version: 'v1' | 'v2' }) {
...
+    let mastra: Mastra;
+    let scorerTest: any;
+    let scorer1: any;
+
+    beforeEach(() => {
+      vi.clearAllMocks();
+      scorerTest = createScorer({
+        name: 'scorerTest',
+        description: 'Test Scorer',
+      }).generateScore(() => 0.95);
```
epinzur, line 6177: maybe have your two dummy scorers return different scores?

packages/core/src/agent/agent.test.ts
```diff
@@ -6073,6 +6085,161 @@ function agentTests({ version }: { version: 'v1' | 'v2' }) {
...
+          }),
+        }),
+      );
+    });
+
+    it(`${version} - should use override scorers when provided in generate options`, async () => {
+      if (version === 'v1') {
+        await agent.generate('Hello world', {
+          scorers: {
+            scorer1: { scorer: mastra.getScorer('scorer1') },
```
epinzur, line 6223: any way you could just pass the scorer name here? and it did the `mastra.getScorer(name)` internally?

epinzur (approved):

packages/core/src/agent/agent.test.ts
YujohnNattrass: nice I idea, i'll take a look into it!

packages/core/src/agent/agent.test.ts
YujohnNattrass: We are testing if the correct scorers are called but not running the actual scorers so the scores are irrelevant.

packages/core/src/agent/agent.test.ts
YujohnNattrass: done!

PR #7079, opened 2025-08-27
added functions to disable/enable hot reload

NikAiyer:
## Description

This PR adds function to enable the disabling of hot reloading for certain server handlers.

## Type of Change

- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [ ] Test update

## Checklist


packages/deployer/src/server/handlers/client.ts
```diff
@@ -34,3 +35,18 @@ export function handleTriggerClientsRefresh(c: Context) {
   });
   return c.json({ success: true, clients: clients.size });
 }
+
+// Functions to control hot reload during template installation
+export function disableHotReload() {
+  isTemplateInstallationActive = true;
```
DanielSLew: I think we should make this more generalized. 
```suggestion
  isHotReadloadDisabled = true;
```

Issue #6773, opened 2025-08-15
[DISCUSSION] Refresh of AI Tracing (Observability) – Umbrella Thread


epinzur:
Hi all,

This issue will serve as an umbrella thread to capture and track all current issues, discussions, fixes, and improvements related to Mastra’s AI Observability and Tracing tooling.

I’m a new member of the Mastra team, and I’ll be focusing on improving and refining the observability tools within the product.

I know there are many related GitHub issues, Slack threads, and Discord conversations on this topic. My goal is to centralize those discussions here, highlight key problems, and align on priorities going forward.

Feel free to share links, ideas, known pain points, or open questions below.

We’re also aiming to ship an experimental refresh of the AI Tracing feature in the next release, so early feedback and suggestions are especially welcome.

DiegoGonzalezCruz:
Awesome! In our case, I have a setup consisting of nextjs, mastra cloud, and copilot kit (headless).
I want to be able to trace chat messages, and group sessions and users Ids.
Also I would like to trace the environment and add some optional metadata, such as the email or other info.

leo-paz:
This is awesome, i'm eager to get my hands on the experimental feature as traces in langfuse aren't working well right now.

Something you've already highlighted that would be great is being able to take runtimeContext and add it as metadata to the trace.

I'm also currently using the ai-v5 tags right now so it would be great if this new feature worked with ai sdk v5.

Some other features im missing:
- an easy way to attach arbitrary data mid workflow to the trace
   - eg. the llm outputs a sql query and you would like to attach the output running the sql to metadata.
- functionId is the name of our agent to make things more readable
- the traceId is returned as output of the workflow. this is useful for running custom evals on datasets where you then need to attach scores to the trace or attach user feedback to the trace.
- random functions like persistWorkflowSnapshot and mapping steps not being traced (this may be personal preference)
- one workflow run triggers exactly one trace with all agents under that trace
- specifying environment

epinzur:
@DiegoGonzalezCruz @leo-paz thanks for the comments and input.  These are super helpful.

Here is a sneak peak of where I am today:

<img width="1420" height="964" alt="Image" src="https://github.com/user-attachments/assets/d29000cd-1864-473a-a035-aaef46267bf5" />

@leo-paz to answer some of your points directly (and please give feedback if they make sense or a different approach might be better)

> - an easy way to attach arbitrary data mid workflow to the trace
[...]

PR #7085, opened 2025-08-27
Added a DefaultExporter for AI Tracing

epinzur:
## Description

Added a DefaultExporter for AI Tracing.  The exporter can run in one of 3 modes, as defined by the user and the storage provider:

- realtime
- batch-with-updates (for SQL-like DBs)
- insert-only (for high-volume prod DBs)

Currently the user must pass this exporter in the mastra Observability config to use it.  This will be fixed in another PR.

I also included the plan I made with Claude Code in the PR for reference on the implementation.

## Related Issue(s)

#6773

## Type of Change

- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [ ] Test update

## Checklist

Will make documentation updates in another PR.


packages/core/src/mastra/index.ts
```diff
@@ -479,9 +479,61 @@ do:
 
     registerHook(AvailableHooks.ON_SCORER_RUN, createOnScorerHook(this));
 
+    /*
+      Register Mastra instance with AI tracing exporters and initialize them
+    */
+    if (config?.observability) {
+      this.registerAITracingExporters();
+      // Fire-and-forget initialization - individual errors are already logged
+      void this.initAITracingExporters();
```
wardpeet: This will cause side effects, right? do you know if this work in serverless environments?

packages/core/src/mastra/index.ts
```diff
@@ -479,9 +479,61 @@ do:
...
+  /**
+   * Register this Mastra instance with AI tracing exporters that need it
+   */
+  private registerAITracingExporters(): void {
+    const allTracingInstances = getAllAITracing();
+    allTracingInstances.forEach(tracing => {
+      const exporters = tracing.getExporters();
+      exporters.forEach(exporter => {
+        // Check if exporter has __registerMastra method
+        if ('__registerMastra' in exporter && typeof (exporter as any).__registerMastra === 'function') {
```
wardpeet, line 502: shouldn't we do instanceof a class instead? Or Can't typescript safe us? Or does not all exporters have this method?

epinzur: most exporters won't need an instance of mastra and won't have this method.  Only ones  like the default that need storage access will.

packages/core/src/mastra/index.ts
epinzur: good catch!

changed to a synchronous method in this commit: https://github.com/mastra-ai/mastra/pull/7085/commits/41cc7603b869008cafe261b5d71ce8ef98c3a684

rase- (approved):

PR #7086, opened 2025-08-27
Support OTEL exporter env vars

YujohnNattrass:
## Description

in our documentation we mention that users can specify OTLP headers: https://mastra.ai/en/docs/observability/tracing#environment-variables

However this never worked because we only checked the endpoint and headers on the telemetry config which was set when their mastra instance was initialized and not the env vars. 

## Type of Change

- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [ ] Test update

## Checklist


packages/deployer/public/templates/instrumentation-template.js
```diff
@@ -15,6 +15,19 @@ import { telemetry } from './telemetry-config.mjs';
 
 globalThis.___MASTRA_TELEMETRY___ = true;
 
+function parseHeaders(headerString) {
```
YujohnNattrass, line 18: helper function to support when headers are set in the .env as:

"header1=value1,header2=value2"

according to otel configuration: https://opentelemetry.io/docs/languages/sdk-configuration/otlp-exporter/#otel_exporter_otlp_headers

abhiaiyer91 (approved):

PR #7095, opened 2025-08-28
update parallel steps video link

bookercodes:
i re-uploaded the video with improvements. this commit replaces the old
(now invalid) link with a link to the newly-uploaded and improved video,
which is also public.


bookercodes:
i made a dog's dinner of this commit, trying again later...

PR #7126, opened 2025-08-28
Taofeeq/cloud 416 add model fallbacks to core

taofeeq-deru:
## Description

This PR introduces model fallbacks in vnext  agent stream.
If there is a list of models, we loop through the list and return the stream of the successful one, if they all fail the last error/stream is returned.
It also introduces `maxRetries` to control the number of times one wants the model call to be retried before moving to the next one on the list or returning the response

## Type of Change

- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [x] Test update

## Checklist


packages/core/src/loop/workflow/llm-execution.ts
```diff
@@ -342,11 +348,49 @@ async function processOutputStream<OUTPUT extends OutputSchema | undefined = und
...
 
+function executeStreamWithFallbackModels<T>(models: ModelManagerModelConfig[]): ExecuteStreamModelManager<T> {
+  return async callback => {
+    let index = 0;
+    let finalResult: any;
+    let done = false;
+    for (const modelConfig of models) {
+      index++;
+      const maxRetries = modelConfig.retry || 0;
+      let attempt = 0;
```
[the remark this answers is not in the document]
taofeeq-deru: that's the expected behaviour, it means to retry just once, attempt - 0 is the first time, attempt - 1 is the retry

packages/core/src/agent/index.ts
```diff
@@ -617,24 +634,25 @@ export class Agent<
    * @param options Options for getting the LLM
    * @returns A promise that resolves to the LLM instance
    */
-  public getLLM({
+  public async getLLM({
```
wardpeet: We dont' want to break the current API
```suggestion
  public getLLM({
```

packages/core/src/agent/agent.test.ts
```diff
@@ -1320,6 +1320,94 @@ function agentTests({ version }: { version: 'v1' | 'v2' }) {
       expect(usedModelName).toBe('premium');
     });
 
+    if (version === 'v2') {
+      it('should use fallback model', async () => {
```
wardpeet:
```suggestion
      it.skip(() => version !== 'v2')('should use fallback model', async () => {
```

packages/core/src/agent/index.ts
```diff
@@ -623,24 +640,25 @@ export class Agent<
...
     runtimeContext = new RuntimeContext(),
     model,
   }: {
     runtimeContext?: RuntimeContext;
     model?: MastraLanguageModel | DynamicArgument<MastraLanguageModel>;
-  } = {}): MastraLLM | Promise<MastraLLM> {
+  } = {}): Promise<MastraLLM> {
```
wardpeet: You can't convert it to Promise yet

```
    const modelToUse = model
      ? typeof model === 'function'
        ? model({ runtimeContext, mastra: this.#mastra })
        : model
      : this.getModel({ runtimeContext });

    return resolveMaybePromise(modelToUse, resolvedModel => {
      let llm: MastraLLM | Promise<MastraLLM>;
      if (resolvedModel.specificationVersion === 'v2') {
        llm = this.prepareModels(runtimeContext, model).then(allModels =>  new MastraLLMVNext({ model: resolvedModel, mastra: this.#mastra, allModels }));
      } else {
        llm = new MastraLLMV1({ model: resolvedModel, mastra: this.#mastra });
      }

      return resolveMaybePromise(llm, (resolvedLLM) => {
      // Apply stored primitives if available
      if (this.#primitives) {
        resolvedLLM.__registerPrimitives(this.#primitives);
      }
      if (this.#mastra) {
        resolvedLLM(this.#mastra);
      }
      return resolvedLLM;
      })
    });
 ```

File: packages/core/src/agent/index.ts [hunk 5]
```diff
@@ -623,24 +640,25 @@ export class Agent<
...
     // If model is provided, resolve it; otherwise use the agent's model
     const modelToUse = model
       ? typeof model === 'function'
         ? model({ runtimeContext, mastra: this.#mastra })
         : model
       : this.getModel({ runtimeContext });
 
-    return resolveMaybePromise(modelToUse, resolvedModel => {
+    return resolveMaybePromise(modelToUse, async resolvedModel => {
```
wardpeet: you have to do the same resolveMaybe here like my comment above

packages/core/src/agent/types.ts
```diff
@@ -68,6 +68,12 @@ export interface AgentConfig<
   description?: string;
   instructions: DynamicArgument<string>;
   model: DynamicArgument<MastraLanguageModel>;
+  maxRetries?: number; //defaults to 0
+  fallbackModels?: {
```
wardpeet: I wonder if we should make this part of model instead?

```
model: DynamicArgument<MastraLanguageModel || { model: DynamicArgument<MastraLanguageModel>;
    maxRetries?: number; //defaults to 0
    enabled?: boolean; //defaults to true][]>;
 ```

File: packages/core/src/stream/aisdk/v5/execute.ts [hunk 7]
```diff
@@ -86,6 +88,10 @@ export function execute<OUTPUT extends OutputSchema | undefined = undefined>({
           console.log('Abort error', error);
         }
 
+        if (!isLastModel) {
```
wardpeet: Can you explain what's happening here with a comment?

taofeeq-deru: when the list of models is being looped though, if it's not the lastModel, we want to throw an error which will be caught when processing the stream in `llm.execution.ts`, and that will make `executeStreamWithFallbackModels` move to the next model on the list.

but if it's the last model, we don't want to throw an error, we just stream the error, and `executeStreamWithFallbackModels` will just return the stream

packages/core/src/agent/types.ts
taofeeq-deru: hmn, so it could be a list of models or just one model. Instead of a main model and a list of fallback models, we'll just have a list of models. We can do that.
Though I think we can still leave maxRetries outside, if one is using just one model but wants it to retry a number of times, or if they want the same maxRetries for all the models in the list, since we default to the agent maxRetries if a fallback model doesn't have maxRetries in its configuration

taofeeq-deru: this requires changes in multiple places where we reference model, and gets crazy pretty fast
I think we should leave fallbackModels separate for now

packages/core/src/agent/index.ts
taofeeq-deru: thanks, added this and pushed

packages/core/src/agent/index.ts
taofeeq-deru: thanks done

packages/core/src/agent/agent.test.ts
taofeeq-deru: thanks, added skip

packages/core/src/agent/types.ts
taofeeq-deru: trying this again

taofeeq-deru: done this

packages/core/src/stream/aisdk/v5/execute.ts
taofeeq-deru: renamed it to `shouldThrowError`

abhiaiyer91 (approved):
Lets wait till after release

examples/weather-agent/src/mastra/agents/index.ts
```diff
@@ -27,7 +28,12 @@ Your primary function is to help users get weather details for specific location
 - Keep responses concise but informative
 
 Use the weatherTool to fetch current weather data.`,
-  model: openai('gpt-4o'),
+  model: [
+    { model: anthropic('claude-3-5-sonnet-20241022') },
+    { model: openai('gpt-4o') },
+    { model: openai('gpt-4o-mini') },
```
TylerBarnes, line 34: does this work too? what's the reason for the object w/ model key?
```suggestion
    anthropic('claude-3-5-sonnet-20241022'),
    openai('gpt-4o'),
    openai('gpt-4o-mini'),
```

packages/core/src/agent/index.ts
```diff
@@ -639,23 +674,28 @@ export class Agent<
...
+          resolvedLLM.__registerPrimitives(this.#primitives);
+        }
+        if (this.#mastra) {
+          resolvedLLM.__registerMastra(this.#mastra);
+        }
+        return resolvedLLM;
+      });
 
-      return llm;
+      return resulvedLLM as MastraLLM;
```
TylerBarnes: theres a spelling mistake here

packages/core/src/agent/index.ts
```diff
@@ -683,7 +723,51 @@ export class Agent<
...
+            details: {
+              agentName: this.name,
+            },
+            text: `[Agent:${this.name}] - Empty model list provided`,
+          });
+          this.logger.trackException(mastraError);
+          this.logger.error(mastraError.toString());
+          throw mastraError;
+        }
+        modelToUse = this.model[0]!.model;
```
TylerBarnes: shouldn't use `!` here, that can lead to runtime errors

```suggestion
        if (!this.model[0]?.model) throw new Error(`expected model provider but none were found`)
        modelToUse = this.model[0].model;
```

packages/core/src/llm/model/model.loop.ts
```diff
@@ -59,19 +60,19 @@ export class MastraLLMVNext extends MastraBase {
...
   }
 
   getModel() {
-    return this.#model;
+    return this.#models[0]!.model;
   }
 
   private _applySchemaCompat(schema: OutputSchema): Schema {
-    const model = this.#model;
+    const model = this.#models[0]?.model;
```
TylerBarnes: same for these, probably we should have a class method for getting the initial default method that throws if one isn't found (to make it so you don't need `!`)

packages/core/src/loop/loop.ts
```diff
@@ -47,8 +47,8 @@ export function loop<Tools extends ToolSet = ToolSet, OUTPUT extends OutputSchem
   const { rootSpan } = getRootSpan({
     operationId: mode === 'stream' ? `mastra.stream` : `mastra.generate`,
     model: {
-      modelId: model.modelId,
-      provider: model.provider,
+      modelId: models[0]?.model?.modelId!,
+      provider: models[0]?.model?.provider!,
```
TylerBarnes: these too

TylerBarnes: we need to assert that a model exists since one is always required for everything to work

packages/core/src/loop/test-utils/fullStream.ts
```diff
@@ -44,45 +44,51 @@ export function fullStreamTests({ loopFn, runId }: { loopFn: typeof loop; runId:
...
-              {
-                role: 'assistant',
-                content: [{ type: 'text', text: 'test-input' }],
-              },
-              {
-                role: 'user',
-                content: [{ type: 'text', text: 'test-input' }],
-              },
-            ]);
+        models: [
```
TylerBarnes, line 48: instead of changing these tests you should make new ones. We need to make sure the old way continues working too. You can duplicate existing tests and add multiple model versions of each test

packages/core/src/loop/test-utils/fullStream.ts
```diff
@@ -883,7 +913,7 @@ export function fullStreamTests({ loopFn, runId }: { loopFn: typeof loop; runId:
 
       const result = await loopFn({
         runId,
-        model: createTestModel({
+        models: createTestModels({
```
TylerBarnes, line 926: same for all tests

packages/core/src/loop/workflow/llm-execution.ts
```diff
@@ -355,11 +361,50 @@ async function processOutputStream<OUTPUT extends OutputSchema | undefined = und
   }
 }
 
+function executeStreamWithFallbackModels<T>(models: ModelManagerModelConfig[]): ExecuteStreamModelManager<T> {
```
TylerBarnes, line 364: @taofeeq-deru did you consider making a new class to encapsulate all the multi-model logic into one file? something like `CombinedLLM`

TylerBarnes (changes requested):
Left some comments but this PR needs documentation and new tests. The test updates here are lowering our test coverage for single model setups and doesn't add any new tests that are specific to this new feature (unless I missed them)

packages/core/src/agent/agent.test.ts
```diff
@@ -1397,6 +1397,100 @@ function agentTests({ version }: { version: 'v1' | 'v2' }) {
       expect(usedModelName).toBe('premium');
     });
 
+    it(
```
taofeeq-deru: @TylerBarnes , test for model list in agent.test.ts

examples/weather-agent/src/mastra/agents/index.ts
taofeeq-deru: yes, this works. I tested it. And the reason for the object is it also takes `maxRetries` which determines the number of times a model will be retried if it fails, and `enabled` which is defaulted to `true` and can be used to easily disable a model without having to delete it from the list

packages/core/src/loop/workflow/llm-execution.ts
taofeeq-deru: yeah, didn't want to change the existing implementation too much. So i left most things as they were, and just changed `model` to `models`

packages/core/src/loop/test-utils/fullStream.ts
taofeeq-deru: the `loop` function the tests use has changed though, instead of `model`, it now takes `models`. That's why the tests had to be updated. From the `Agent` side of things, nothing seem to really change as an `Agent` can still take just one model and work fine, or the new model list

packages/core/src/agent/index.ts
taofeeq-deru: thanks, corrected this

packages/core/src/agent/index.ts
taofeeq-deru: checking this now, thanks

packages/core/src/llm/model/model.loop.ts
taofeeq-deru: checking this now, thanks

packages/core/src/loop/loop.ts
taofeeq-deru: doing this now, thanks

packages/core/src/agent/index.ts
```diff
@@ -706,7 +706,7 @@ export class Agent<
         return resolvedLLM;
       });
 
-      return resulvedLLM as MastraLLM;
+      return resolvedLLM as MastraLLM;
```
TylerBarnes: does this lead to scope issues with the nested variable name? maybe it makes more sense to do something like:
```ts
      return resolveMaybePromise(llm, resolvedLLM => { ... }) as MastraLLM;
```

packages/core/src/loop/test-utils/options.ts
```diff
@@ -1666,156 +1666,7 @@ export function optionsTests({ loopFn, runId }: { loopFn: typeof loop; runId: st
...
-                    "cachedInputTokens": 3,
-                    "inputTokens": 3,
-                    "outputTokens": 10,
-                    "reasoningTokens": 10,
-                    "totalTokens": 23,
-                  },
-                  "warnings": [],
-                },
-              ],
+              "steps": [],
```
TylerBarnes: is this a regression?

packages/core/src/loop/test-utils/options.ts
```diff
@@ -1869,38 +1720,8 @@ export function optionsTests({ loopFn, runId }: { loopFn: typeof loop; runId: st
...
-                            "toolCallId": "call-1",
-                            "toolName": "tool1",
-                            "type": "tool-result",
-                          },
-                        ],
-                        "role": "tool",
-                      },
-                    ],
-                     "modelId": "mock-model-id",
+                    "messages": [],
```
TylerBarnes: this too

packages/core/src/loop/workflow/llm-execution.ts
```diff
@@ -502,11 +552,11 @@ export function createLLMExecutionStep<
...
             model: {
-              modelId: modelResult.model.modelId,
-              provider: modelResult.model.provider,
-              version: modelResult.model.specificationVersion,
+              modelId: model.modelId,
+              provider: model.provider,
+              version: model.specificationVersion,
             },
-            stream: modelResult.stream as ReadableStream<ChunkType>,
+            stream: modelResult as unknown as ReadableStream<ChunkType>,
```
TylerBarnes: is this right?

packages/core/src/agent/agent.test.ts
```diff
@@ -1397,6 +1397,100 @@ function agentTests({ version }: { version: 'v1' | 'v2' }) {
...
+            {
+              model: premiumModel,
+            },
+          ],
+        });
+
+        await agent.streamVNext('Test message');
+
+        await new Promise(resolve => setTimeout(resolve, 1000));
+        expect(usedModelName).toBe('premium');
```
TylerBarnes: should add an assertion on the output stream results as well

packages/core/src/agent/agent.test.ts
```diff
@@ -1397,6 +1397,100 @@ function agentTests({ version }: { version: 'v1' | 'v2' }) {
...
+            },
+            {
+              model: premiumModel,
+            },
+          ],
+        });
+
+        await agent.streamVNext('Test message');
+
+        await new Promise(resolve => setTimeout(resolve, 1000));
```
TylerBarnes:
```suggestion
        const result = await agent.streamVNext('Test message');
        await result.consumeStream();
```

packages/core/src/agent/agent.test.ts
```diff
@@ -6459,6 +6553,124 @@ function agentTests({ version }: { version: 'v1' | 'v2' }) {
...
+        model: [
+          {
+            model: openaiModel,
+          },
+          {
+            model: openai('gpt-4o-mini'),
+          },
+          {
+            model: openai('gpt-4.1'),
+          },
```
TylerBarnes: can you add a test that makes sure object and direct model config here both work? To make sure we don't regress that later.

So a separate test that checks that `model: [mockModel1, mockModel2, mockModel3]` works as well as `model: [{model: mockModel1}, {model: mockModel2}, {model: mockModel3}]` and `model: [{model: mockModel1}, mockModel2, {model: mockModel3}]`

packages/core/src/agent/agent.test.ts
```diff
@@ -6459,6 +6553,124 @@ function agentTests({ version }: { version: 'v1' | 'v2' }) {
...
```
TylerBarnes: we need tests for the other config options here as well

packages/core/src/llm/model/model.loop.ts
```diff
@@ -1,4 +1,4 @@
-import type { LanguageModelV2 } from '@ai-sdk/provider-v5';
+// import type { LanguageModelV2 } from '@ai-sdk/provider-v5';
```
TylerBarnes:
```suggestion
```

packages/core/src/loop/test-utils/fullStream.ts
TylerBarnes: ah got it, maybe to keep the diff small it'd be better to change from:
```
model: new MockLanguageModelV2({
```
to
```
models: [new MockLanguageModelV2({
```

packages/core/src/loop/test-utils/fullStream.ts
```diff
@@ -593,7 +607,7 @@ export function fullStreamTests({ loopFn, runId }: { loopFn: typeof loop; runId:
       const result = await loopFn({
         runId,
         messageList,
-        model: modelWithFiles,
+        models: [{ maxRetries: 0, id: 'test-model', model: modelWithFiles }],
```
TylerBarnes, line 616: what's the id here used for? doesn't the model provider on the model key already have an id on it?

packages/core/src/loop/test-utils/streamObject.ts
```diff
@@ -863,6 +919,11 @@ export function streamObjectTests({ loopFn, runId }: { loopFn: typeof loop; runI
               "error": undefined,
               "files": [],
               "finishReason": "stop",
+              "model": {
+                "modelId": "",
+                "provider": "",
+                "specificationVersion": "",
```
TylerBarnes: shouldn't this have values set?

packages/core/src/loop/test-utils/streamObject.ts
```diff
@@ -1022,6 +1089,11 @@ export function streamObjectTests({ loopFn, runId }: { loopFn: typeof loop; runI
             Error message: Validation failed],
               "files": [],
               "finishReason": "error",
+              "model": {
+                "modelId": "",
+                "provider": "",
+                "specificationVersion": "",
+              },
```
TylerBarnes: and this one

packages/core/src/loop/workflow/llm-execution.ts
TylerBarnes: looks like it was modelResult.stream before

packages/core/src/loop/workflow/llm-execution.ts
```diff
@@ -385,206 +430,227 @@ export function createLLMExecutionStep<
...
-          if (options?.prepareStep) {
-            try {
-              const prepareStepResult = await options.prepareStep({
-                stepNumber: inputData.output?.steps?.length || 0,
-                steps: inputData.output?.steps || [],
-                model,
-                messages: messageList.get.all.aiV5.model(),
-              });
+      const { outputStream, runState, request, warnings, rawResponse, callBail } =
+        await executeStreamWithFallbackModels<{
```
TylerBarnes: adding this here makes the diff huge, maybe this should be part of the workflow outside the step instead?

TylerBarnes: the diff being so large here makes me nervous that any changes from main were merged in properly

packages/core/src/loop/workflow/llm-execution.ts
TylerBarnes: ah yeah that's why I'm wondering, you could make even fewer changes to the codebase if you create a single composite model from the other models, similar to how packages like this work https://github.com/nakasyou/ai-fallback. Then there's still only 1 model and the code would generally be the same as before, only the public api would change to allow multiple models in an array. So the diff for this PR would probably be 90% smaller and not touch everything

TylerBarnes (changes requested):

packages/core/src/agent/agent.test.ts
TylerBarnes: to make sure the mocked messages are returned in the stream and final result

TylerBarnes:
@taofeeq-deru looks like tests are failing still too

packages/core/src/loop/workflow/llm-execution.ts
taofeeq-deru: at that point, we can't really catch the model stream error anymore because it'd have been processed and returned fully. the only place i could successfully catch the stream error is here.
That's why I couldn't create a composite model like the package you recommended - https://github.com/nakasyou/ai-fallback. Because when there's an error in stream, it doesn't throw it, it just streams it down. So have to catch the error in the process stream flow

packages/core/src/loop/test-utils/fullStream.ts
taofeeq-deru: the `id` is different from the `modelId`, it's a random uuid generated in `agent.ts` and used to make it possible to easily reorder the models or enable/disable a model
if we want to use the `modelId`, then dynamic models will have to be resolved first before we can do anything, and if for whatever reason, there's an instance where a model is repeated in the list, then it messes it things up

packages/core/src/agent/agent.test.ts
taofeeq-deru: `model: [mockModel1, mockModel2, mockModel3]` this won't work
has to be object
it can either be `model: mockModel1` or `model: [{model: mockModel1}, {model: mockModel2}]`, that object can then take other things like maxRetries and enabled, so like `model: [{model: mockModel1, enabled: true, maxRetries: 3}, {model: mockModel2, enabled: true, maxRetries: 2}]`,  `enabled` is defaulted to true and `maxRetries` is defaulted to 1

packages/core/src/loop/workflow/llm-execution.ts
taofeeq-deru: resolved this, thanks

packages/core/src/agent/index.ts
taofeeq-deru: yeah, changed it to this, thanks

packages/core/src/loop/workflow/llm-execution.ts
taofeeq-deru: yeah, i originally wanted to do that. And all of my changes would probably have been in core/agent.ts file, and it wouldn't have been much. But this is the only place where I can catch a stream error, before it gets streamed down

packages/core/src/loop/test-utils/fullStream.ts
TylerBarnes: do we need that? It seems ok to me to repeat models in the list, maybe each one has a different API key attached to it for ex

packages/core/src/loop/workflow/llm-execution.ts
TylerBarnes: do we need to catch it though? the model just needs to fallback to the next one

taofeeq-deru: yeah, but we need to know when to fallback. Typically, we want to fallback to the next one when a model fails, to know when a model fails, we need to catch the stream error. And stream error can't be caught anymore after the processing is done.
After processing is done, the only way to know if a stream has error in it is by reading the stream, using something like `toTextResponse` or so, which delays the process, because then we'll need to get the whole stream, check for error in it, before returning it, if there's no error. This will make `stream` feel like `generate`

packages/core/src/loop/test-utils/fullStream.ts
taofeeq-deru: so, I'm not saying not to repeat models in list, I'm saying we need a unique id for each of them, to be able to update them with ease, one can update the model in the object it self, or the enabled status of the model, or the maxRetries for the model. With that unique id, there's no mistake of updating the wrong model. When `agent.getModelList` is called, it returns the unique id generated for each model config

packages/core/src/loop/test-utils/streamObject.ts
taofeeq-deru: fixed this, thanks

packages/core/src/loop/test-utils/streamObject.ts
taofeeq-deru: fixed this, thanks

packages/core/src/agent/agent.test.ts
taofeeq-deru: I've added tests for them

packages/core/src/agent/agent.test.ts
taofeeq-deru: checking for the stream result now too

packages/core/src/agent/agent.test.ts
```diff
@@ -7028,6 +7028,704 @@ function agentTests({ version }: { version: 'v1' | 'v2' }) {
     }
   }, 10000);
 
+  describe(
+    'model list',
+    {
+      skip: version === 'v1',
```
TylerBarnes, line 7035: does this mean it only works with streamVNext? if so that's not a problem but is there logic in here to make sure users know that?

packages/core/src/loop/workflow/llm-execution.ts
```diff
@@ -385,206 +430,223 @@ export function createLLMExecutionStep<
     inputSchema: llmIterationOutputSchema,
     outputSchema: llmIterationOutputSchema,
     execute: async ({ inputData, bail, tracingContext }) => {
-      const runState = new AgenticRunState({
-        _internal: _internal!,
-        model,
-      });
```
TylerBarnes, line 391: the diff here is hard to review cause it's so long, can you double check what's on main vs this branch to be extra sure everything is merged right?

TylerBarnes (approved):
After seeing your demo this makes a lot more sense to me! Approved but left a couple questions

packages/core/src/agent/agent.test.ts
taofeeq-deru: yeah, only works with `v2`. This will be indicated in the docs, and will also add a check, so it throws whenever v1 models are added to the list

packages/core/src/loop/workflow/llm-execution.ts
taofeeq-deru: will do

PR #7204, opened 2025-08-29
Resumable streams

taofeeq-deru:
## Type of Change

- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [ ] Test update

## Checklist


.changeset/easy-kids-enjoy.md
```diff
@@ -0,0 +1,10 @@
+---
+'@mastra/client-js': patch
+'@mastra/deployer': patch
+'@mastra/server': patch
+'@mastra/core': patch
+'mastra': patch
+'create-mastra': patch
```
wardpeet:
```suggestion
```

wardpeet (commented):
How do we feel about adding storage to the server key instead of creating a new thing?
This means we can use exisitng storage as a key-value basically? It won't be set by default when adding storage to the mastra key

PR #7218, opened 2025-08-29
load env files with mastra start

TheIsrael1:
- **mastra start - load env files, custom env**
- **Deprecate redundant mastra build --env flag**
- **changeset**

## Related Issue(s)

Fixes #7056

## Type of Change

- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [ ] Test update

## Checklist


docs/src/content/en/reference/cli/build.mdx
```diff
@@ -35,12 +35,6 @@ mastra build [options]
       description: "Comma-separated list of paths to tool files to include",
       isOptional: true,
     },
-    {
```
abhiaiyer91, line 38: Whats the reason we got rid of this?

TheIsrael1: It is kind of redundant today, we don't use the env files at build time for the build bundler. and since the mastra start command will now load the env files while running the build output, it will be good to deprecate this to prevent confusion.

abhiaiyer91 (approved):

Issue #7137, opened 2025-08-29
Pipe abortcontroller to agent alls #5428 not working !


zackshen:
I’m dealing with a situation where the user asks a question, and the agent responds in a streaming manner. If the user clicks Cancel on the client side (browser), then the agent on the server should stop, and any tools the agent is calling should also stop.

I read this document [Tool Advanced Usage](https://mastra.ai/en/docs/tools-mcp/advanced-usage), and it explains how to handle the cancellation event inside a tool call after an agent has been canceled.

I found a Pull Request that seems to address a similar requirement: [#5428](https://github.com/mastra-ai/mastra/pull/5428/). However, it looks like there are some issues with its changes.
	1.	Incorrect parameter name – In both [generateHandler handling abort signal](https://github.com/mastra-ai/mastra/pull/5428/files#diff-71438ce2ee79815921f228d205f4ddfc54b77d672594f577622e76fa38f751d3R253) and [streamGenerateHandler handling abort signal](https://github.com/mastra-ai/mastra/pull/5428/files#diff-71438ce2ee79815921f228d205f4ddfc54b77d672594f577622e76fa38f751d3R301), the parameter should be `abortSignal: abortSignal`, not `signal: abortSignal`.
After I corrected these two parameter names, I was able to successfully receive the `abortSignal` in the second argument of `createTool`’s `execute` method.
	2.	Abort not propagating – However, when I call `controller.abort()`  in the browser, the agent in Mastra on the backend does not stop, and the `execute` method in `createTool` continues running step by step. Also, the `abortSignal.aborted` state always remains `false`. So the second problem is that the abort operation on the client side does not seem to cancel the agent running on the backend.

TheIsrael1:
Hi @zackshen, thanks for flagging this. We now have a PR up for the bad `signal` property set in the mastra server.

For the second issue of stopping tool execution on abort. The tools are separate function calls, so once invoked it keeps running as a regular javascript function. The tool execute does expose the abortSignal though for you to so special handling.
we have that documented here https://mastra.ai/en/docs/tools-mcp/advanced-usage#abort-signals

There also seems to have been a regression with createTool exposure of that option
I have a fix here #7267 and would be in the next release

PR #7267, opened 2025-09-02
Fix - add missing tool options to createTool

TheIsrael1:
## Related Issue(s)
#7137

## Type of Change

- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [ ] Test update

## Checklist

### DIFF
--- a/packages/core/src/tools/tool.ts
+++ b/packages/core/src/tools/tool.ts
@@ -1,3 +1,4 @@
+import type { ToolCallOptions } from '@ai-sdk/provider-utils-v5';
 import type { ToolExecutionOptions } from 'ai';
 import type { z } from 'zod';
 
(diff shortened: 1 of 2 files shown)


packages/core/src/tools/tool.ts
```diff
@@ -1,3 +1,4 @@
+import type { ToolCallOptions } from '@ai-sdk/provider-utils-v5';
```
wardpeet: I wonder if it should come from 
`import type { ToolCallOptions, Tool as ToolV5 } from 'ai-v5';`
instead

wardpeet (approved):

packages/core/src/tools/tool.ts
TheIsrael1: looks like we use both rn in the core package,  i'm not sure which is best haha, but 'ai-v5' is concise

PR #7361, opened 2025-09-03
stopped creating event spans for llm_chunks in ai-observability

epinzur:
## Description
 
It turns out that a single chunk from an LLM streaming response is generally a single word.  Creating a LLM_Chunk event span for each returned word is too verbose.

This PR removes our creation of LLM_Chunks spans for streaming Agents in legacy & vnext streaming.

## Related Issue(s)

#6773

## Type of Change

- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [ ] Test update

## Checklist

### DIFF
--- a/packages/core/src/loop/workflow/stream.ts
+++ b/packages/core/src/loop/workflow/stream.ts
@@ -27,18 +27,6 @@ export function workflowLoopStream<
     start: async controller => {
       const writer = new WritableStream<ChunkType>({
         write: chunk => {
-          // Create event spans for streaming chunks on the LLM span (not agent span)
-          if (llmAISpan && chunk.type === 'text-delta') {
-            llmAISpan.createEventSpan({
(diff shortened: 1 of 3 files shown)


packages/core/src/loop/workflow/stream.ts
```diff
@@ -27,18 +27,6 @@ export function workflowLoopStream<
...
-            llmAISpan.createEventSpan({
-              type: AISpanType.LLM_CHUNK,
-              name: `llm chunk: ${chunk.type}`,
-              output: chunk.payload.text,
-              attributes: {
-                chunkType: chunk.type,
-              },
-            });
-          }
-
```
epinzur, line 41: we already capture the output for vNext streaming elsewhere, so no need to re-implement for this version.

greglobinski (approved):

PR #7376, opened 2025-09-03
cloud exporter for ai traces

YujohnNattrass:
## Description

 Implements CloudAITracingExporter for streaming AI trace data to Mastra Cloud
 observability platform.

Things to note
  - Only supports batch inserts
  - Extracts team/project IDs from JWT access token

## Type of Change

- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [ ] Test update

## Checklist


packages/cloud/src/ai-tracing/index.ts
```diff
@@ -0,0 +1,313 @@
...
+    const accessToken = config.accessToken ?? process.env.MASTRA_CLOUD_ACCESS_TOKEN;
+    if (!accessToken) {
+      throw new MastraError({
+        id: `CLOUD_AI_TRACING_EXPORTER_ACCESS_TOKEN_REQUIRED`,
+        domain: ErrorDomain.MASTRA_OBSERVABILITY,
+        category: ErrorCategory.USER,
+      });
+    }
+
+    const { teamId, projectId, cloudEndpoint } = this.parseAccessToken(accessToken);
```
YujohnNattrass: I'm thinking of adding the cloudEndpoint to the access token. This way they would only need to add one env var. Any opinions here?

YujohnNattrass:
@greptileai can you do a review on the whole PR again

YujohnNattrass:
@greptilai can you do another review?

I have fixed the api endpoint mismatch issue.
AnyAISpan does have the attributes and metadata fields so there are no type issues there.
We want to specify the path of ai spans when we import

packages/cloud/src/ai-tracing/index.ts
```diff
@@ -0,0 +1,311 @@
...
+      this.logger.error('Batch upload failed after all retries, dropping batch', mastraError);
+      // Don't re-throw - we want to continue processing new events
+    }
+  }
+
+  /**
+   * Uploads spans to cloud API using fetchWithRetry for all retry logic
+   */
+  private async batchUpload(spans: CloudSpanRecord[]): Promise<void> {
+    const url = `${this.config.endpoint}/aiTracing/publish`;
```
YujohnNattrass: @rphansen91 just wanted to double check if this is the correct path

packages/cloud/src/ai-tracing/index.ts
```diff
@@ -0,0 +1,311 @@
...
+    this.buffer.totalSize++;
+  }
+
+  private formatSpan(span: AnyAISpan): CloudSpanRecord {
+    const spanRecord: CloudSpanRecord = {
+      teamId: this.teamId,
+      projectId: this.projectId,
+      traceId: span.traceId,
+      spanId: span.id,
+      parentSpanId: span.parent?.id ?? null,
```
YujohnNattrass, line 123: @rphansen91 should we be transforming the parentSpanId here or in cloud? I was thinking in cloud but I'm open to either way.

rphansen91: @YujohnNattrass I am transforming this into an empty string on cloud

packages/cloud/src/ai-tracing/index.ts
rphansen91: @YujohnNattrass I have it as `/ai/spans/publish`

rphansen91 (changes requested):

packages/cloud/src/ai-tracing/index.ts
rphansen91: @YujohnNattrass We will be configuring this in the deployer so I think we should follow the same patttern where we can supply the headers and a url

This makes it usable from outside of cloud as well

rphansen91 (changes requested):

rphansen91 (approved):

PR #7382, opened 2025-09-03
fix(@mastra/core): prevent text deltas from being stored as individual parts

### LINKED ISSUE #7050: [DISCORD:1408759298871394304] Bug or feature? mastra_messages parts are stored as streamed chunks
This issue was created from Discord post: https://discord.com/channels/1309558646228779139/1408759298871394304

After switching to @mastra/core 0.14.1,  @mastra/memory@0.13.1 and agent.streamVNext with format: 'aisdk'  I noticed that mastra_messages parts are stored in LibSQLStore as an array of streamed chunks. Before 0.14.0  the parts were stored as an array of entire messages. Is it a bug or a feature of aisdk mode?

TylerBarnes:
Fixes https://github.com/mastra-ai/mastra/issues/7050
LLM providers that send response-metadata chunks after each text-delta would cause text to be split into individual parts instead of being combined.

Some providers like OpenRouter add response-metadata after each text-delta chunk. This was triggering our flush logic, resulting in messages with separate text parts for each delta:
```
[{ type: 'text', text: 'Hello' }, { type: 'text', text: ' world' }]
```

Now we properly combine them into a single text part:
```
[{ type: 'text', text: 'Hello world' }]
```

This ensures cleaner message formatting in streamVNext when working with providers that include frequent metadata chunks.


NikAiyer (approved):

packages/core/src/stream/base/output.ts
```diff
@@ -261,7 +261,7 @@ export class MastraModelOutput<OUTPUT extends OutputSchema = undefined> extends
                 isContinued: chunk.payload.stepResult.isContinued,
                 logprobs: chunk.payload.stepResult.logprobs,
                 finishReason: chunk.payload.stepResult.reason,
-                response: { ...otherMetadata, messages: chunk.payload.messages.nonUser } as any,
+                response: { ...otherMetadata, messages: messageList.get.response.aiV5.model() } as any,
```
DanielSLew: was this fix needed?

PR #7445, opened 2025-09-04
fix(@mastra/core): output processors not running before saving to memory

### LINKED ISSUE #7087: [BUG] Memory unaffected by Agent Output Processors
### Describe the Bug

When adding an Output Processor to an agent, you are able to modify or filter the output of the response before it goes to the user. For example, if you modified the text of the result, you can see the changed result return to the UI as expected. 

In a situation where you are loading the previous messages from the thread and displaying them to a user, all you have to do is hit refresh, and you'll see the original unmodified text of the result. 

The expectation is that any changes made via an Output Processor would propagate into the memory database itself. 

Especially when an Output Processor is being used to redact or remove PII or other sensitive material, this is a major oversight. 

### Steps To Reproduce

In a mastra project, create a new Output Processor that for demo purposes, removes the response text completely:

```
import type { Processor } from "@mastra/core/processors";
import type { ChunkType } from "@mastra/core";

export class MetadataProcessor implements Processor {
  readonly name = "metadata-processor";

  constructor() {}

  async processOutputStream({
    part,
    streamParts,
    state,
    abort,
  }: {
    part: ChunkType;
    streamParts: ChunkType[];
    state: Record<string, any>;
    abort: (reason?: string) => never;
  }): Promise<ChunkType | null | undefined> {
    if (part.type === "text-delta") {
      return null;
    }

    return part; // Emit the part
  }
}
```

Add that to your agent:
```
export const myAgent = new Agent({
  {...}
  outputProcessors: [new MetadataProcessor()],
});
```

Next, use generate or stream or streamVNext to generate a response. You'll see a response without any text. Be sure to set a threadId and resourceId
```
agent.streamVNext('hi there', {
   format: "aisdk",
   memory:  {
      thread: 'thread-id-here',
      resource: 'resource-id-here'
   }
}
```

Finally, get all the messages from the thread:
```
const messages =
    (await storage?.getMessages({
      threadId: 'thread-id-here',
      format: "v2",
    })) || [];

console.log(messages);
```
In the above, you'll see the the unmodified original message content as if the Output Processor never ran
```

### Expected Behavior

If we are modifying the output before it returns to the user, we'd want to ensure that the message's content retains that modification beyond the initial request-- so those changes should propagate into the database.

### Environment Information

- Mastra Version: 0.14.1

### Verification

### PR DESCRIPTION
**Pull request by TylerBarnes**
Output processors weren't being applied before messages were saved to memory, causing unprocessed content (like sensitive data) to be stored.

The fix ensures output processors transform messages before they hit memory storage. This was a simple ordering issue in the streaming pipeline.

Before:
```typescript
// Messages saved first, then processed
await self.saveMemory(self.messageList);
self.messageList = await self.processorRunner.runOutputProcessors(self.messageList);
```

After:
```typescript
// Process first, then save
self.messageList = await self.processorRunner.runOutputProcessors(self.messageList);
await self.saveMemory(self.messageList);
```

Added integration tests to verify the fix works correctly with various output processors including PII redaction and pattern removal.

Fixes #7087


abhiaiyer91 (approved):

DanielSLew (approved):

TylerBarnes:
merging with failing core tests cause those are from this other PR that was merged with failing tests https://github.com/mastra-ai/mastra/pull/7437

PR #7486, opened 2025-09-04
fix(@mastra/memory): memory pagination not working when using query() method

### LINKED ISSUE #6787: [BUG] The pagination of memory method query() not work
### Describe the Bug

When using memory, executing `query({selectBy:{pagination: { page: 1, perPage: 1 }}})`  should got paginated data, but not work

Image

### Steps To Reproduce

```typescript

const store = new PostgresStore({
     connectionString: 'your connection string',
});

const memory = new Memory({
        storage: store, // 
});

// add messages
// memory.saveMessages() 

const messages = await memory.query({
     resourceId: resourceId, // replace it
     threadId: threadId, // replace it
     selectBy: { pagination: { page: 1, perPage: 1 } },
});
```

### Expected Behavior

`pagination: { page: 1, perPage: 1 }` should get one message

### Environment Information

-  "@mastra/core": "^0.13.2"
-  "@mastra/memory": "^0.12.2"
- "@mastra/pg": "^0.14.0"
- node: 22.14.0
- postgres: 16.9

### Verification

TylerBarnes:
This fixes the pagination bug where Memory.query() was ignoring pagination parameters and returning all messages instead of the requested page.

The issue was that query() always called storage.getMessages() which doesn't support pagination, instead of storage.getMessagesPaginated() which does.

After:
```typescript  
// Now correctly returns only 10 messages from page 0
const result = await memory.query({
  threadId,
  resourceId,
  selectBy: { pagination: { page: 0, perPage: 10 } }
});
```

The fix checks if pagination is requested and uses the appropriate storage method. Backward compatibility is maintained for queries without pagination.

Fixes #6787

### DIFF
--- a/.changeset/witty-oranges-fly.md
+++ b/.changeset/witty-oranges-fly.md
@@ -1,0 +1,7 @@
+---
+'@mastra/memory': patch
+---
+
+Fix memory pagination not working when using query() method
+
+The Memory.query() method was ignoring pagination parameters in the selectBy option, always returning all messages instead of the requested page. This fix ensures pagination works correctly by using storage.getMessagesPaginated() when pagination is requested, while maintaining backward compatibility with storage.getMessages() for non-paginated queries.
--- a/packages/memory/integration-tests/src/with-pg-storage.test.ts
+++ b/packages/memory/integration-tests/src/with-pg-storage.test.ts
@@ -1,8 +1,9 @@
+import { randomUUID } from 'crypto';
 import { fastembed } from '@mastra/fastembed';
 import { Memory } from '@mastra/memory';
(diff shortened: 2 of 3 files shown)


NikAiyer (approved):

TylerBarnes (commented):
OpenCode says: Thanks for the suggestion @greptileai! While I appreciate the code cleanup recommendation, we're planning to deprecate the non-paginated getMessages() method in the future, so the duplication is temporary. Once we remove the deprecated path, this conditional logic and duplication will be eliminated entirely. For now, keeping them separate makes the deprecation path clearer.

OpenCode

PR #7496, opened 2025-09-05
Playground UI - components update for Observability 

greglobinski:
## Type of Change

- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [ ] Test update

## Checklist


packages/playground-ui/src/domains/observability/components/trace-span-usage.tsx
```diff
@@ -0,0 +1,124 @@
...
+      const { promptTokens, completionTokens, totalTokens } = tokensByProvider[provider];
+      acc.promptTokens += promptTokens;
+      acc.completionTokens += completionTokens;
+      acc.totalTokens += totalTokens;
+      return acc;
+    },
+    { promptTokens: 0, completionTokens: 0, totalTokens: 0 },
+  );
+
+  const tokensByProviderValid = JSON.stringify(traceUsage) === JSON.stringify(traceTokensBasedOnSpans);
```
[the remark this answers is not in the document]
greglobinski: in next iteration

packages/playground-ui/src/components/ui/elements/entry-list/entry-list-cell.tsx
```diff
@@ -1,3 +1,41 @@
...
-  return <span className="text-icon4 text-[0.875rem] truncate ">{children}</span>;
+import { cn } from '@/lib/utils';
+import { VisuallyHidden } from '@radix-ui/react-visually-hidden';
+import { useMemo } from 'react';
+
+export function EntryListTextCell({ children, isLoading }: { children: React.ReactNode; isLoading?: boolean }) {
+  // Generate random width once per component render
+  const randomWidth = useMemo(() => {
+    return Math.floor(Math.random() * (90 - 50 + 1)) + 50;
+  }, []);
```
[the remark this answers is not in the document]
greglobinski: yeah :)

packages/playground-ui/src/components/ui/elements/buttons/button.tsx
```diff
@@ -2,36 +2,35 @@ import { cn } from '@/lib/utils';
...
   children: React.ReactNode;
   variant?: 'primary' | 'outline' | 'ghost';
-  target?: string;
 }
 
-export const Button = ({ className, as, variant = 'outline', ...props }: ButtonProps) => {
-  const Component = as || 'button';
+export const Button = React.forwardRef<HTMLButtonElement, ButtonProps>(
+  ({ className, variant = 'outline', ...props }: ButtonProps, ref) => {
+    return (
```
mfrachet, line 12: I think we still need the "as" and "href" so that we can have links looking like buttons?

packages/playground-ui/src/components/ui/elements/entry-list/entry-list-cell.tsx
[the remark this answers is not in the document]
mfrachet: I don't think I understand why we want random width but regarding the greptile comment, I think it is right, math.random on the server and client will mismatch when rendering the width

packages/playground-ui/src/components/ui/elements/entry-list/entry-list-item.tsx
```diff
@@ -3,20 +3,20 @@ import { Column, getColumnTemplate } from './shared';
...
+  selectedItemId?: string;
+  onClick?: (itemId: string) => void;
   children?: React.ReactNode;
   columns?: Column[];
+  isLoading?: boolean;
 }) {
-  const isSelected = selectedItem && selectedItem?.id === item.id;
-
-  console.log('selectedItem', selectedItem, item, isSelected);
+  const isSelected = selectedItemId === item.id;
```
mfrachet, line 19: you can pass isSelected from the parent, this will allow to remove the need for `item` and `selectedId` which will make the component more generic

packages/playground-ui/src/components/ui/elements/entry-list/entry-list-item.tsx
```diff
@@ -3,20 +3,20 @@ import { Column, getColumnTemplate } from './shared';
 
 export function EntryListItem({
   item,
-  selectedItem,
+  selectedItemId,
   onClick,
   children,
   columns,
+  isLoading,
```
mfrachet, line 10: isLoading can become isDisabled in this case I think. isLoading for an "ClickableEntry" means nothing while disabled does 😊

packages/playground-ui/src/components/ui/elements/entry-list/entry-list.tsx
```diff
@@ -1,15 +1,17 @@
...
 import { EntryListItem } from './entry-list-item';
 import { getColumnTemplate, type Column } from './shared';
 
 import { cn } from '@/lib/utils';
+import { isValidElement } from 'react';
 
 export function EntryList({
-  items,
-  selectedItem,
+  items: dataItems,
```
mfrachet, line 10: I don't think we need this. We could probably keep EntryList as a UI component (with no information about any entities type or whatever.

What we could do instead is
- reducing the EntryList props list
- add a {children} instead
- create a "business component" that knows about domain types, that uses EntryList and all the other things

So that we can have a generic EntryList.

packages/playground-ui/src/components/ui/elements/entry-list/entry-list.tsx
```diff
@@ -31,16 +35,22 @@ export function EntryList({
   perPage?: number;
   columns?: Column[];
   searchTerm?: string;
+  setEndOfListElement: (element: HTMLDivElement | null) => void;
 }) {
-  console.log('selectedItem', selectedItem);
+  const loadingItems: Record<string, any>[] = Array.from({ length: 3 }).map((_, index) => {
+    return {
```
mfrachet, line 41: for instance, this should live in the parent component and not in the UI one

packages/playground-ui/src/components/ui/elements/headers/page-header.tsx
```diff
@@ -0,0 +1,27 @@
...
+    <header
+      className={cn(
+        'grid gap-[.5rem] pt-[2rem] pb-[2rem]',
+        '[&>h1]:text-icon6 [&>h1]:text-[1.25rem] [&>h1]:font-normal [&>h1]:flex [&>h1]:items-center [&>h1]:gap-[0.5rem]',
+        '[&_svg]:w-[1.4rem] [&_svg]:h-[1.4rem] [&_svg]:text-icon3',
+        '[&>p]:text-icon4 [&>p]:text-[0.875rem] [&>p]:m-0',
+        className,
+      )}
+    >
+      <h1>
```
mfrachet, line 21: careful with this one, maybe we can add a "titleAs: ELementyType" that defalts to h1. Pretty sure in some situations this won't work

packages/playground-ui/src/components/ui/elements/entry-list/entry-list-cell.tsx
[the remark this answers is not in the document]
greglobinski: the random is used only if `isLoading=true` it's job is to generate fake shape of data for loading state. so there is no change it can mess with ssr

packages/playground-ui/src/components/ui/elements/buttons/button.tsx
greglobinski: in this part we do not, I will work on it in the next iteration, for now this Button is not exported from playground-ui, and it's used only in these component

packages/playground-ui/src/components/ui/elements/entry-list/entry-list-item.tsx
greglobinski: no because isLoading in this case provide information to generate "loading" state and disabled could/is only a part of this state, isLoading trigger the component to render different html/css version

packages/playground-ui/src/components/ui/elements/entry-list/entry-list-item.tsx
greglobinski: will remember of this, for the next iteration

packages/playground-ui/src/components/ui/elements/entry-list/entry-list.tsx
greglobinski: will think of it in the next iteration

packages/playground-ui/src/components/ui/elements/headers/page-header.tsx
greglobinski: you are right, it's still work in progress

mfrachet (approved):

PR #7523, opened 2025-09-05
feat(pg): add Cloud SQL + IAM authentication support

nuritizra:
## Description

Add support for Google Cloud SQL + IAM by allowing a raw `pg.ClientConfig` to be passed to `PostgresStore`.

Currently, `PostgresConfig` only accepts either:

* a connection string, or
* a static TCP config (`{host, port, database, user, password, ssl}`).

This breaks when using the Cloud SQL Node.js Connector with IAM auth, because it doesn’t expose a resolvable host.

When Backstage is configured with:

```yaml
database:
  client: pg
  connection:
    type: cloudsql
    instance: ${DATABASE_INSTANCE}
    user: ${DATABASE_USER}
    ipAddressType: PRIVATE
```

the Cloud SQL connector is used ([[see `postgres.ts` in Backstage](https://github.com/backstage/backstage/blob/master/packages/backend-defaults/src/entrypoints/database/connectors/postgres.ts#L111)]
That connector establishes a secure stream and returns a `pg.ClientConfig` with:

* `stream`: a TLS socket connected to Cloud SQL
* `ssl`: TLS options
* `user`, `database`
* `password`: often a **function** that dynamically generates IAM tokens

There is **no stable DNS host** to extract. Forcing `{host, port}` is incorrect and fails at runtime.
`pg` and `pg-promise` already support this shape — we just need to accept and forward it.

### Changes in this PR

* Extend `PostgresConfig` to accept a raw `pg.ClientConfig`.
* Relax constructor validation to allow configs with `stream` or a functional `password`.
* Forward connector configs directly to `pg-promise`.

This preserves full backward compatibility with connection strings and TCP configs.

## Type of Change

* [ ] Bug fix (non-breaking change that fixes an issue)
* [ ] Documentation update
* [ ] Code refactoring
* [ ] Performance improvement
* [ ] Test update

---

## Checklist


stores/pg/src/storage/index.ts
```diff
@@ -79,23 +88,45 @@ export class PostgresStore extends MastraStorage {
...
+      } else if ('host' in config) {
+        this.#config = {
+          host: (config as any).host,
+          port: (config as any).port,
+          database: (config as any).database,
+          user: (config as any).user,
+          password: (config as any).password,
+          ssl: (config as any).ssl,
+          max: config.max,
+          idleTimeoutMillis: config.idleTimeoutMillis,
```
NikAiyer: Is there a way we can get away from using `as any` for a lot of these config changes?

Maybe we can add some type guards like these?

```
      const isConnectionStringConfig = (cfg: PostgresConfig): cfg is PostgresConfig & { connectionString: string } => {
        return 'connectionString' in cfg;
      };

      const isHostConfig = (
        cfg: PostgresConfig,
      ): cfg is PostgresConfig & {
        host: string;
        port: number;
        database: string;
        user: string;
        password: string;
        ssl?: boolean | ISSLConfig;
      } => {
        return 'host' in cfg && 'database' in cfg && 'user' in cfg && 'password' in cfg;
      };

      const isCloudSqlConfig = (cfg: PostgresConfig): cfg is PostgresConfig & ClientConfig => {
        return 'stream' in cfg || ('password' in cfg && typeof cfg.password === 'function');
      };
      ```

NikAiyer: That way they're more specifically typed and we won't need to rely on using assertions as much.

nuritizra:
Hi @NikAiyer, I’ve addressed your comments—could you please take another look when you have a moment? Thanks!

NikAiyer (approved):
LGTM! Thanks for making this PR!

PR #7539, opened 2025-09-05
fix(@mastra/pg): only return Mastra-managed tables from listIndexes()

TylerBarnes:
Fixes #6691 

The listIndexes() method was returning all tables with vector columns, causing initialization to fail when external pgvector tables existed in the database.

Now it only returns tables with the exact Mastra structure (vector_id, embedding, metadata columns), allowing PgVector to coexist with other applications using pgvector.

Added tests to verify the fix handles external vector tables correctly.


rphansen91 (approved):
Looks good, nice idea matching the table structure

NikAiyer (approved):

PR #7555, opened 2025-09-06
Reproduce/issue 6827 repetitive tool calling error of gemini 2.5 pro

OoO256:
## Description
reproduction of https://github.com/mastra-ai/mastra/issues/6827

## Related Issue(s)

https://github.com/mastra-ai/mastra/issues/6827

## Type of Change

- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [ ] Test update

## Checklist

### DIFF
--- a/packages/core/src/agent/agent-repetitive-tools-simple.test.ts
+++ b/packages/core/src/agent/agent-repetitive-tools-simple.test.ts
@@ -107,6 +107,74 @@ describe('Agent - Repetitive Tool Calls Issue #6827', () => {
     }
   }, 20_000);
 
+  it('should call a final options tool exactly once at end of conversation', async () => {
+    let toolExecutionCount = 0;
+    const callIds: string[] = [];
+
+    const presentChoicesTool = createTool({
+      id: 'present-choices',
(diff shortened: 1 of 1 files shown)


TylerBarnes:
Hey @OoO256 it's your system prompt, took me way longer than I'd like to admit to figure that out haha. It fails as `You are a helpful assistant. Provide a concise final answer. Then, as the very last action, call the 'present-choices' tool exactly once to present user options and then stop. Do not call any tool more than once.` but succeeds as `You are a helpful assistant.`

I think you're confusing the model by telling it the tool call must be the very last action. It's trying to do that but it can't because there's always another action it could take. So you're asking it to do nothing and it doesn't know how to do that

TylerBarnes:
Adding on here that you have a few better options:
1. Allow the tool call to not be the final step, account for that in your code/logic
2. Use `structuredOutput` in `streamVNext` https://mastra.ai/en/reference/agents/streamVNext. This allows an agent to take multiple steps and then reliably output JSON at the end (which you can then use to call the tool/function/api yourself)
3. Set `maxSteps` to 1 so multiple tool calls can't happen

PR #7514, opened 2025-09-05
paul/grwth-766-update-memory-api-references

PaulieScanlon:
## Description

- docs: update all memory api references

## Related Issue(s)

- https://github.com/mastra-ai/mastra/issues/7428
- https://github.com/mastra-ai/mastra/issues/7372

## Type of Change

- [ ] Bug fix (non-breaking change that fixes an issue)
- [ ] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [x] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [ ] Test update

## Checklist

- [ ] I have made corresponding changes to the documentation (if applicable)
- [ ] I have added tests that prove my fix is effective or that my feature works

PR #7557, opened 2025-09-07
paul/grwth-768-memory-examples-docs

PaulieScanlon:
## Description

- add examples removed from API reference to docs and refactor memory overview
- add missing front matter to all /memory pages

## Related Issue(s)

- https://github.com/mastra-ai/mastra/pull/7514

## Type of Change

- [x] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [ ] Test update

## Checklist


TylerBarnes (changes requested):
I'm not sure about this. We had a ton of info like this in the overview before and spent some time moving it out into other pages. Perhaps we should have a memory/storage page?
This seems like less of an overview now and more a dump of information 😅

PaulieScanlon:
Thanks @TylerBarnes, whilst that may be true, and these pages could use some further refinement, I had to remove those code snippets from the API references, because that's not where they should live, and rather than lose them. I gave them a home under overview. We've used a similar patter for [Agents](https://mastra.ai/en/docs/agents/overview). 

Are either of these things blocking changes? Could we get both PRs merged and i'll make a note to come back and refine Memory pages?

TylerBarnes:
> Thanks @TylerBarnes, whilst that may be true, and these pages could use some further refinement, I had to remove those code snippets from the API references, because that's not where they should live, and rather than lose them. I gave them a home under overview. We've used a similar patter for [Agents](https://mastra.ai/en/docs/agents/overview).
> 
> Are either of these things blocking changes? Could we get both PRs merged and i'll make a note to come back and refine Memory pages?

@PaulieScanlon I think we should do it right the first time

TylerBarnes (approved):

PR #7566, opened 2025-09-08
fix(braintrust-exporter): add `projectName` as a top-level field and transform usage metrics for Braintrust

knjiang:
## Description

This PR adds `projectName` as a top-level field on the exporter and transforms usage metrics from AI SDK format to `BraintrustUsageMetrics` format (e.g.: `inputTokens` -> `prompt_tokens`). 

built package with my changes locally, exported traces and confirmed in the ui that metrics show up with usage metrics:
image

## Type of Change

- [] New feature (non-breaking change that adds functionality)
- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [ ] Test update

## Checklist


observability/braintrust/src/ai-tracing.ts
```diff
@@ -330,3 +387,9 @@ export class BraintrustExporter implements AITracingExporter {
     this.traceMap.clear();
   }
 }
+
+function isWorkflowStepLLMExecutionSpan(span: AnyAISpan): boolean {
+  return (
+    span.type === AISpanType.WORKFLOW_STEP && (span.attributes as WorkflowStepAttributes)?.stepId === 'llm-execution'
+  );
+}
```
knjiang: is there an existing type that handles this step? i couldn't find any and the token usage was associated with this span.

observability/braintrust/src/ai-tracing.test.ts
```diff
@@ -446,6 +446,115 @@ describe('BraintrustExporter', () => {
...
+          stepId: 'llm-execution',
+        },
+        output: {
+          output: {
+            usage: {
+              inputTokens: 10,
+              outputTokens: 5,
+              totalTokens: 15,
+            },
+          },
```
epinzur: We generally only expect usage data to be returned from `LLM_GENERATION` SpanTypes as part of the attributes field. 

If you are currently see usage info in `WORKFLOW_STEP` span types, this is a bug and this data will be removed in the future.

observability/braintrust/src/ai-tracing.ts
epinzur: this should be returned as an LLM_GENERATION span inside the WORKFLOW_STEP span.

epinzur (changes requested):
please remove the extra logic around `usage` tracking of `WORKFLOW_STEP` spans.

observability/braintrust/src/ai-tracing.test.ts
knjiang: thanks! should i file an issue in github or is there a known case? i think the code path i'm hitting is https://github.com/mastra-ai/mastra/blob/main/packages/core/src/loop/workflow/llm-execution.ts

epinzur (approved):
looks great. thanks for these updates!

epinzur:
looks like you need to do a lint check... `pnpm lint -- --fix`

PR #7599, opened 2025-09-08
Fixed ai tracing for workflows nested directly in agents

epinzur:
## Description

AI Tracing propegation was broken for workflows run directly inside agents like the following example:

```
const simpleWorkflow = createSimpleWorkflow();

const workflowAgent = new Agent({
  name: 'Workflow Agent',
  instructions: 'You can execute workflows that exist in your config',
  model,
  workflows: {
    simpleWorkflow,
  },
});
```

It is fixed now.

## Related Issue(s)

#6773

## Type of Change

- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [ ] Test update

## Checklist

### DIFF
--- a/packages/core/src/agent/index.ts
+++ b/packages/core/src/agent/index.ts
@@ -1413,6 +1413,7 @@ export class Agent<
                 const streamResult = run.streamVNext({
                   inputData: context,
                   runtimeContext,
+                  tracingContext: innerTracingContext,
                   format,
                 });
 
@@ -2741,10 +2742,10 @@ Message ${msg.threadId && msg.threadId !== threadObject.id ? 'from previous conv
           resourceId,
           runId,
           runtimeContext,
+          tracingContext: { currentSpan: agentAISpan },
           writableStream: options.writableStream,
           methodType,
           format,
-          tracingContext: { currentSpan: agentAISpan },
         });
 
         return {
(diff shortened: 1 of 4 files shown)


packages/core/src/agent/index.ts
```diff
@@ -1413,6 +1413,7 @@ export class Agent<
                 const streamResult = run.streamVNext({
                   inputData: context,
                   runtimeContext,
+                  tracingContext: innerTracingContext,
```
epinzur, line 1416: this is the primary fix.

packages/core/src/workflows/workflow.ts
```diff
@@ -1498,8 +1498,14 @@ export class Run<
   streamVNext({
     inputData,
     runtimeContext,
+    tracingContext,
     format,
-  }: { inputData?: z.infer<TInput>; runtimeContext?: RuntimeContext; format?: 'aisdk' | 'mastra' | undefined } = {}) {
+  }: {
+    inputData?: z.infer<TInput>;
+    runtimeContext?: RuntimeContext;
+    tracingContext?: TracingContext;
```
epinzur, line 1506: along with passing tracingContext thru here

abhiaiyer91 (approved):

PR #7616, opened 2025-09-09
Adjust workflow `.execute()` docs

rase-:
- Remove reference doc
- Add code comments


PaulieScanlon:
Hi @rase- can you also remove a reference to page from: https://github.com/mastra-ai/mastra/blob/main/docs/src/content/en/reference/workflows/_meta.ts#L16

Otherwise the page metadata.title will appear in the navigation list.

PaulieScanlon (changes requested):
Remove `.execute()` from [_meta.ts](https://github.com/mastra-ai/mastra/blob/main/docs/src/content/en/reference/workflows/_meta.ts#L16)

PaulieScanlon (approved):

PR #7647, opened 2025-09-09
fix(@mastra/core): apply output processors before saving to memory

### LINKED ISSUE #7087: [BUG] Memory unaffected by Agent Output Processors

TylerBarnes:
Fixes #7087

Output processors were modifying the streamed response but the original unmodified text was still being saved to memory. This was particularly problematic for PII redaction use cases.

The fix moves output processor application earlier in the stream pipeline, ensuring transformations happen before messages are persisted to storage.

Before this fix, if you had a processor that redacted sensitive data:
```typescript
class RedactionProcessor implements Processor {
  async processOutputStream({ part }) {
    if (part.type === 'text-delta' && part.text) {
      return { ...part, text: part.text.replace(/SSN-\d{3}-\d{2}-\d{4}/g, '[REDACTED]') };
    }
    return part;
  }
}
```

The user would see `[REDACTED]` in the stream, but refreshing the page would show the original SSN from memory.

Now the processor runs before memory storage, so both the stream and stored messages show `[REDACTED]` consistently.


packages/core/src/agent/index.ts
```diff
@@ -279,6 +279,22 @@ export class Agent<
...
+  private async getResolvedOutputProcessors(runtimeContext?: RuntimeContext): Promise<OutputProcessor[]> {
+    if (!this.#outputProcessors) {
+      return [];
+    }
+
+    if (typeof this.#outputProcessors === 'function') {
+      if (!runtimeContext) {
+        // Create a default runtime context if none provided
+        return [];
+      }
```
DanielSLew: The comment contradicts what the code is doing, don't we want to return the output processors even if there's no runtime context

packages/core/src/stream/base/output.ts
```diff
@@ -560,41 +602,10 @@ export class MastraModelOutput<OUTPUT extends OutputSchema = undefined> extends
 
     let fullStream = this.teeStream();
 
-    const processorStates = new Map<string, ProcessorState>();
+    // Output processors are now applied in the baseStream transform
+    // so we don't need to apply them here anymore
```
DanielSLew:
```suggestion
```

packages/core/src/loop/loop.ts
```diff
@@ -110,7 +110,8 @@ export function loop<Tools extends ToolSet = ToolSet, OUTPUT extends OutputSchem
       onStepFinish: rest.options?.onStepFinish,
       includeRawChunks: !!includeRawChunks,
       output: rest.output,
-      outputProcessors,
+      // Don't pass outputProcessors here - they're applied in the inner MastraModelOutput
+      // created in llm-execution.ts to avoid double processing
```
NikAiyer: Do we need this comment?

packages/core/src/stream/base/output.ts
NikAiyer: Probably can remove this

packages/core/src/loop/loop.ts
DanielSLew:
```suggestion
```

DanielSLew (approved):

packages/core/src/agent/index.ts
TylerBarnes: good catch, yeah this looks weird

NikAiyer (approved):

packages/core/src/loop/loop.ts
TylerBarnes: Added this here intentionally since we initialize MastraModelOutput in multiple places, wanted to leavea note on why we don't do output processing in this place anymore and where it does happen instead

CalebBarnes (approved):

packages/core/src/loop/loop.ts
TylerBarnes: yeah, wanted to leave a note for the future https://github.com/mastra-ai/mastra/pull/7647/files/9e04c49b9de4e54f9e2e7fc5ce443fd2af7718a3#r2334682806

PR #7712, opened 2025-09-10
Validate SQL queries to Prevent Postgres Function Calls

alliehowe29:
## Description

In the sqlExecutionTool call we validate that we only allow SELECT statements but this still allows internal postgres functions to be called.

These can be dangerous such as when pg_sleep is called and creates long running queries that could result in a denial of service.

This PR adds validation around which function calls are allowed as defined by an allow list.

## Related Issue(s)

This was originally fixed here: https://github.com/mastra-ai/template-text-to-sql/pull/3

## Type of Change

- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [ ] Test update

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Added SQL query validation layer to enhance security and prevent unauthorized database operations.


abhiaiyer91 (approved):

templates/template-text-to-sql/src/mastra/tools/sql-execution-tool.ts
```diff
@@ -43,6 +154,8 @@ export const sqlExecutionTool = createTool({
         throw new Error('Only SELECT queries are allowed for security reasons');
       }
```
wardpeet: unused code
```suggestion
```

templates/template-text-to-sql/src/mastra/tools/sql-execution-tool.ts
```diff
@@ -43,6 +166,8 @@ export const sqlExecutionTool = createTool({
...
```
wardpeet:
```suggestion
```

PR #3807, opened 2025-04-24
[MASTRA-3131] Storage Workflow Runs

NikAiyer:
## Related Issue(s)

This PR adds getWorkflowRunByID and adds resourceID as a filter for getWorkflowRuns for storage providers.

## Type of Change

- [ ] Bug fix (non-breaking change that fixes an issue)
- [ ] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [ ] Test update

## Checklist

- [ ] I have performed a self-review of my own code
- [ ] I have made corresponding changes to the documentation (if applicable)
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] New and existing unit tests pass locally with my changes
- [ ] I have generated a changeset for this PR

PR #7740, opened 2025-09-11
Add resource id to workflow run snapshots

### LINKED ISSUE #5856: Attach ResourceID to Workflow/Run
### Problem Statement

When working with workflow and steps, the requirement is to resume/suspend the workflow for a user.
Workflow Run snapshots are persisted in the storage, but is not associated with any resourceID (could be specific to user like userID).

### Proposed Solution

There is method to `getWorkflowRuns` which accepts resourceID as param
`
getWorkflowRuns(args?: {
        fromDate?: Date;
        toDate?: Date;
        limit?: number;
        offset?: number;
        resourceId?: string;
    }): Promise<WorkflowRuns>;
`

But there is no way to attach a resourceID to a workflow run, which can be expected in `createRun` method.
`
createRun(params?: {
        runId?: string;
    }): Promise<{
        runId: string;
    }>;
`

This is similar to Thread which is created with resourceID, so the resource threads can be fetched using:
`mastra.getThreadsByResourceId` OR `mastraClient.getMemoryThreads`

### Component

Workflows

### Alternatives Considered

_No response_

### Example Use Case

As a developer, i can list out all the user's completed/suspended workflows, and resume suspended workflow steps.

### Additional Context

_No response_

### Verification

CalebBarnes:
## Description

This PR implements support for associating workflow runs with `resourceId` identifiers, addressing the need to track workflow runs per user or resource. This enables developers to query, filter, and manage workflow runs based on the resource they belong to, similar to how memory threads work with `resourceId`.

## Related Issue(s)

Closes #5856

### Core Workflow Engine Changes
- **Updated `Run` class** to accept and store `resourceId` parameter in constructor
- **Enhanced `createRun()` and `createRunAsync()`** methods to accept `resourceId` option
- **Modified execution pipeline** to pass `resourceId` through all execution methods:
  - `DefaultExecutionEngine.execute()`
  - `executeStep()`, `executeConditional()`, `executeParallel()`, `executeLoop()`, `executeForeach()`
  - `executeEntry()` and `persistStepUpdate()`
- **Added `resourceId` to step execution context** via `ExecuteFunctionParams`

### Storage Layer Updates
- **Updated base storage interfaces** to accept `resourceId` parameter in `persistWorkflowSnapshot()`
- **Enhanced all storage providers** to handle `resourceId`:
  - PostgreSQL, LibSQL, MongoDB, DynamoDB, ClickHouse
  - Cloudflare D1, Cloudflare KV, MSSQL, Lance, Upstash
  - In-memory storage and MockStore
- **Added backward compatibility** via `alterTable()` to add `resourceId` column to existing databases

### Type System Updates
- **Extended `StorageWorkflowRun`** interface to include optional `resourceId` field
- **Updated workflow execution types** to thread `resourceId` through the execution pipeline
- **Enhanced storage method signatures** across all providers

### Testing & Validation
- **Added comprehensive test suite** for `resourceId` persistence and retrieval
- **Created example implementation** demonstrating `resourceId` usage
- **Updated storage test utilities** to validate `resourceId` functionality across all providers

### Associating workflow runs to a resource id:
```typescript
// Create workflow run with resourceId (e.g., userId)
const run = await workflow.createRunAsync({ 
  resourceId: 'user-123' 
});

await run.start({ inputData: { /* ... */ } });

// Query runs by resourceId
const { runs } = await workflow.getWorkflowRuns({ 
  resourceId: 'user-123' 
});
```

### Access resourceId in workflow step functions
```typescript
// Access resourceId in step functions
const step = createStep({
  id: 'myStep',
  execute: async ({ resourceId }) => {
    console.log('Running for resource:', resourceId);
    // Step logic here
  }
});
```

## Type of Change

- [x] Database migration (adds optional `resourceId` column to workflow snapshot tables)

## Backward Compatibility

✅ **Fully backward compatible** - all existing workflows continue to work unchanged. The `resourceId` parameter is optional everywhere and existing workflow runs will have `resourceId: undefined`.

## Database Changes

The PR includes automatic database migrations that add an optional `resourceId` column to the `workflow_snapshot` table across all storage providers. This is handled via the existing `alterTable()` mechanism with `ifNotExists` checks.


packages/core/src/storage/base.ts
```diff
@@ -372,22 +372,30 @@ export abstract class MastraStorage extends MastraBase {
       schema: TABLE_SCHEMAS[TABLE_MESSAGES],
       ifNotExists: ['resourceId'],
     });
+    await this?.alterTable?.({
+      tableName: TABLE_WORKFLOW_SNAPSHOT,
+      schema: TABLE_SCHEMAS[TABLE_WORKFLOW_SNAPSHOT],
+      ifNotExists: ['resourceId'],
+    });
```
CalebBarnes, line 379: @abhiaiyer91 🤔 It looks like `resourceId` was added to the workflow snapshots table schema [4 months ago~ in this PR](https://github.com/mastra-ai/mastra/pull/3807/files#diff-b77405b3c883d5fb422360f6ffc3a910d4ebb347f1bb8241d9af07d6645fe1bc)

But I guess if anyone still had a DB from before that they might not have the `resourceId` field and we should still run `alterTable` here. Does that seem right? Unless theres something I'm missing somewhere else that handles migrations for outdated db schemas.

stores/mssql/src/storage/domains/workflows/index.ts
```diff
@@ -96,17 +98,19 @@ export class WorkflowsMSSQL extends WorkflowsStorage {
       const request = this.pool.request();
       request.input('workflow_name', workflowName);
       request.input('run_id', runId);
+      request.input('resourceId', resourceId);
```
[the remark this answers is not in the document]
CalebBarnes: The db column name was already added as `resourceId` camel cased months ago so we probably shouldn't change that at this point.

packages/core/src/storage/base.ts
abhiaiyer91: Yes this is good

abhiaiyer91 (approved):

PR #7750, opened 2025-09-11
fix(core): deprecate output property in favor of structuredOutput

DanielSLew:
This PR deprecates the `output` property in `generateVNext` and `streamVNext` methods in favor of using the `structuredOutput` property.

When `maxSteps` is set to 1, the system now internally maps `structuredOutput.schema` to `output` and handles the model override from `structuredOutput.model`. This maintains backward compatibility while encouraging users to adopt the new API.

Before:
```typescript
agent.streamVNext(messages, {
  output: schema,
  maxSteps: 1
})
```

After:
```typescript
agent.streamVNext(messages, {
  structuredOutput: {
    schema: schema,
    model: customModel // optional
  },
  maxSteps: 1
})
```

The old `output` property still works but is marked as deprecated and will be removed in a future version.

### DIFF
--- a/packages/core/src/agent/agent.test.ts
+++ b/packages/core/src/agent/agent.test.ts
@@ -8196,7 +8196,7 @@ describe('Stream ID Consistency', () => {
   }, 10000); // Increase timeout to 10 seconds
 });
 
-describe.only('Agent structuredOutput to output deprecation mapping', () => {
+describe('Agent structuredOutput to output deprecation mapping', () => {
   it('should map structuredOutput to output when maxSteps is 1', async () => {
     let structuredOutputProcessorGotCalled = false;
     const mockModel = new MockLanguageModelV2({
--- a/packages/core/src/agent/agent.test.ts
+++ b/packages/core/src/agent/agent.test.ts
@@ -8195,3 +8195,280 @@ describe('Stream ID Consistency', () => {
     expect(onFinishResult.object).toEqual({ name: 'John', age: 30 });
   }, 10000); // Increase timeout to 10 seconds
 });
+
+describe.only('Agent structuredOutput to output deprecation mapping', () => {
+  it('should map structuredOutput to output when maxSteps is 1', async () => {
+    let structuredOutputProcessorGotCalled = false;
+    const mockModel = new MockLanguageModelV2({
+      doGenerate: async () => {
+        return {
+          content: [
+            {
+              type: 'text',
+              text: '{"name": "John", "age": 30}',
+            },
+          ],
+          finishReason: 'stop',
+          usage: { inputTokens: 10, outputTokens: 20, totalTokens: 30 },
+          warnings: [],
+        };
+      },
+      doStream: async _options => {
+        return {
+          stream: convertArrayToReadableStream([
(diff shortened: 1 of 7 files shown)


packages/core/src/agent/agent.ts
```diff
@@ -3557,6 +3576,7 @@ export class Agent<
       ...mergedStreamOptions,
       messages,
       methodType: 'streamVNext',
+      _modelOverride: modelOverride,
```
TylerBarnes: do we need a `_modelOverride`? can't you just add a different model in the input arg?

DanielSLew: no, because the model is set on the Agent level, and this is the generate/stream level, there's no model input arg

TylerBarnes: ah ok, it should just be called `model` then, meaning for this execute call, use this model

DanielSLew:
```suggestion
      model: modelOverride,
```

packages/core/src/agent/agent.types.ts
```diff
@@ -138,4 +142,6 @@ export type InnerAgentExecutionOptions<
   writableStream?: WritableStream<ChunkType>;
   messages: MessageListInput;
   methodType: 'generate' | 'stream' | 'streamVNext';
+  /** Internal: Model override for when structuredOutput.model is used with maxSteps=1 */
+  _modelOverride?: MastraLanguageModel;
```
DanielSLew:
```suggestion
  /** Internal: Model override for when structuredOutput.model is used with maxSteps=1 */
  model?: MastraLanguageModel;
```

packages/core/src/agent/agent.ts
```diff
@@ -2697,7 +2697,7 @@ export class Agent<
       );
     }
 
-    const llm = (await this.getLLM({ runtimeContext })) as MastraLLMVNext;
+    const llm = (await this.getLLM({ runtimeContext, model: options._modelOverride })) as MastraLLMVNext;
```
DanielSLew:
```suggestion
    const llm = (await this.getLLM({ runtimeContext, model: options.model })) as MastraLLMVNext;
```

packages/core/src/agent/agent.ts
DanielSLew: fair enough, made that change!

packages/core/src/agent/agent.test.ts
```diff
@@ -8195,3 +8195,280 @@ describe('Stream ID Consistency', () => {
...
+      ],
+      {
+        structuredOutput: {
+          schema: outputSchema,
+        },
+        maxSteps: 1,
+      },
+    );
+
+    expect(structuredOutputProcessorGotCalled).toBe(false);
```
CalebBarnes, line 8269: It doesn't look like this `structuredOutputProcessorGotCalled` is used in this test.

CalebBarnes (approved):

packages/core/src/agent/agent.test.ts
DanielSLew: oh good catch, added that in

PR #7835, opened 2025-09-15
fix: structuredOutput returns undefined object in generateVNext result

nehaaprasad:
### Description

fixes: #7780

- This PR fixes a bug where `result.object` was returning `undefined` when using `structuredOutput` with `agent.generateVNext()`, even though the structured output was being generated correctly.

**Root Cause:** The issue occurred because `streamVNext` was always using "stream" mode for output processors, but the `StructuredOutputProcessor` only implements `processOutputResult` (for "result" mode), not `processOutputStream` (for "stream" mode). This meant the structured output processor was never being executed, so the structured data was never extracted and stored in the result object.

**Solution:** 
1. Modified the processor mode selection logic to detect when a `StructuredOutputProcessor` is present and automatically switch to "result" mode
2. Updated the stream output logic to properly extract structured data from message metadata regardless of whether an explicit output schema is provided

- This ensures that when `structuredOutput` is used, the processor runs in the correct mode and the structured data is properly returned in `result.object` with the expected TypeScript types.

## Related Issue(s)

- Fixes the bug reported where `result.object` type is undefined when using `structuredOutput` with `agent.generateVNext()`


CalebBarnes:
Hi @naaa760 thanks for the PR! 

This looks like it fixes a separate issue, not #7780. The #7780 issue is about the result.object type not being correctly inferred from the provided structuredOutput.schema.

nehaaprasad:
Thanks @CalebBarnes  You're right about the issue numbers 
this PR fixes the rerank compatibility bug where newer AI SDK versions throw "V2 models are not supported for stream" error.
 The fix is working correctly by using the right generation method for V2 vs V1 models. 
Should I update the issue reference or create a new issue for proper tracking?

CalebBarnes:
The referenced issue #7780 will be resolved by #7969

If you are also experiencing a separate issue of `structuredOutput` not working with `generateVNext` or another issue please create an issue detailing it and if you'd also like to make a PR for that it'd be awesome if you write a test that reproduces the issue that is being fixed.

I'm currently working through some changes to structured output processor so for now I'll close this PR. Thanks again for your help! I'll reference the change you made here.

PR #7862, opened 2025-09-15
fix(core): pass correct step results to stopWhen condition

### LINKED ISSUE #7771: [DISCORD:1415804120857575485] tool calls and text don't always stream as separate steps, making stopWhen less useful
This issue was created from Discord post: https://discord.com/channels/1309558646228779139/1415804120857575485

original text
```
Sometimes we don’t need generated text — we only need the tool call data to render a UI related to it. In those cases, the generated text becomes redundant and even looks unprofessional to end users, especially when a cleaner and more effective custom tool call UI is already available.

In previous versions, I was extending the tool input schemas just before registering them to the agent by adding a "direct_return" boolean property, supported by a description in the system prompt. Then, in the agent’s ".stream" method, I used the "stopWhen" function to detect whether a step included any tool call with the direct_return property. That approach worked well, since tool call and text parts were separated into different steps.

However, in the current versions of Mastra, both tool calls and generated text appear within the same step. Because of this, "stopWhen" no longer provides the control I need. As a workaround, I’ve started handling this through the outputProcessor, which works to some extent — but it doesn't prevent the LLM from wasting tokens. It only modifies the output during streaming and storage, not during generation itself.

This feels like a core use case, so having a native way to stop generation during tool actions or within output processors would be extremely useful.

Also, I don’t think AbortSignal helps much here — it just behaves like an error interrupt, not a natural way to stop generation cleanly.
```

Issue:
"However, in the current versions of Mastra, both tool calls and generated text appear within the same step. Because of this, "stopWhen" no longer provides the control I need. As a workaround, I’ve started handling this through the outputProcessor, which works to some extent — but it doesn't prevent the LLM from wasting tokens. It only modifies the output during streaming and storage, not during generation itself.
"

DanielSLew:
Fixes #7771 where stopWhen wasn't working properly because it wasn't receiving the accumulated steps from previous iterations.

The issue was that each workflow iteration was only passing the current step to stopWhen, not the full history. This meant stopWhen conditions that needed to check multiple steps (like stopping after a tool call followed by text) couldn't work correctly.

The fix accumulates all steps across iterations and passes them to stopWhen. Also refactored the workflow structure to be cleaner - extracted the loop logic into a separate agentic-loop workflow and split the execution steps into more focused modules.

- On the first iteration stopWhen used to get an empty step, instead of the actual result data
- On step 2 it would get the step 1 results, so it was always a step behind
- We would also call stopWhen an extra time when the LLM execution was already stopping
- Did a big refactor to make the workflows file structure make more sense and changed the naming to better reflect what the workflows are doing.


packages/core/src/loop/workflow/stream.ts
```diff
@@ -1,188 +0,0 @@
-import { ReadableStream } from 'stream/web';
```
DanielSLew, line 1: This file was moved to `/workflows/stream.ts` and I simplified it, GitHub isn't smart enough to detect that

packages/core/src/loop/workflows/agentic-execution/index.ts
```diff
@@ -0,0 +1,69 @@
+import type { ToolSet } from 'ai-v5';
```
DanielSLew, line 1: This used to be outerLLMWorkflow but I thought that name didn't make any sense as to what it was actually doing. This workflow runs a single execution against the LLM

packages/core/src/loop/workflows/agentic-execution/llm-execution-step.ts
```diff
@@ -651,13 +651,20 @@ export function createLLMExecutionStep<
...
+      const currentIterationContent = allResponseContent.slice(existingResponseCount);
+
       steps.push(
         new DefaultStepResult({
           warnings: outputStream._getImmediateWarnings(),
           providerMetadata: providerOptions,
           finishReason: runState.state.stepResult?.reason,
-          content: messageList.get.response.aiV5.modelContent(),
-          // @ts-ignore this is how it worked internally for transformResponse which was removed TODO: how should this actually work?
+          content: currentIterationContent,
```
DanielSLew, line 729: We were passing the wrong content to the step

packages/core/src/loop/workflows/agentic-execution/llm-mapping-step.ts
```diff
@@ -1,38 +1,18 @@
 import type { ToolSet } from 'ai-v5';
```
DanielSLew, line 1: This file wasn't renamed, it was created, GitHub is confused. This step I just extracted from the agentic-execution workflow. It has 2 other steps extracted into separate files, but the third step was created in file. So I moved it out for consistency and readability

packages/core/src/loop/workflows/agentic-loop/index.ts
```diff
@@ -0,0 +1,151 @@
+import type { StepResult, ToolSet } from 'ai-v5';
```
DanielSLew, line 1: This workflow used to get created inside of `stream.ts` and it was just called `mainWorkflow` but I moved it out and renamed it to better organize the code

packages/core/src/loop/workflows/schema.ts
```diff
@@ -0,0 +1,143 @@
...
+  warnings: z.array(z.any()),
+  isContinued: z.boolean(),
+  logprobs: z.any().optional(),
+  totalUsage: languageModelUsageSchema.optional(),
+  headers: z.record(z.string()).optional(),
+  messageId: z.string().optional(),
+  request: z.record(z.any()).optional(),
+});
+
+export const llmIterationOutputSchema = z.object({
```
DanielSLew, line 97: I tried to add some zod types but the more I went down this path it got way too hairy. I ended up just casting the inputData correct type inside of the workflow code

DanielSLew: Which also makes me think that this code might be the wrong use case for using workflows, but that's a different convo!

packages/core/src/loop/workflows/agentic-execution/llm-execution-step.ts
```diff
@@ -651,13 +651,20 @@ export function createLLMExecutionStep<
 
       const steps = inputData.output?.steps || [];
 
+      // Only include content from this iteration, not all accumulated content
+      // Get the number of existing response messages to know where this iteration starts
+      const existingResponseCount = inputData.messages?.nonUser?.length || 0;
+      const allResponseContent = messageList.get.response.aiV5.modelContent();
+
+      // Extract only the content added in this iteration
+      const currentIterationContent = allResponseContent.slice(existingResponseCount);
```
TylerBarnes: I think the right fix might be to pass the step number into `stepContent(number)` to only return content for a specific step, and use `stepContent` instead of `modelContent` here. I meant to add that to `stepContent()` but didn't get around to it

packages/core/src/stream/base/output.ts
```diff
@@ -316,7 +316,7 @@ export class MastraModelOutput<OUTPUT extends OutputSchema = undefined> extends
                 request: request,
                 usage: chunk.payload.output.usage,
                 // TODO: need to be able to pass a step id into this fn to get the content for a specific step id
```
TylerBarnes:
```suggestion
```

packages/core/src/agent/message-list/index.ts
```diff
@@ -431,7 +431,140 @@ export class MessageList {
...
+            {
+              id: `step-${stepNumber}`,
+              role: 'assistant',
+              parts: stepParts,
+            },
+          ];
+
+          // Convert to model messages without adding extra step-start markers
+          const modelMessages = AIV5.convertToModelMessages(this.sanitizeV5UIMessages(stepUiMessages));
+          return modelMessages.flatMap(m => this.response.aiV5.stepContent(m));
```
TylerBarnes:
```suggestion
          return modelMessages.flatMap(this.response.aiV5.stepContent);
```

TylerBarnes (approved):

Issue #6852, opened 2025-08-20
Monorepos Support mega issue


LekoArts:
# Mastra 🤝🏻 Monorepos

Hello everyone! 👋🏻

We love to see how many of you want to use Mastra. A lot of you are using Mastra as part of their monorepo setup, often authored in TypeScript. We hear you loud and clear (e.g. through issue like [#1996](https://github.com/mastra-ai/mastra/issues/1996)) that there are currently pain points when trying to use TypeScript monorepos with Mastra.

The gist of it is that Mastra currently doesn't support parsing TypeScript files by default. You either have to add a build step to your packages, compiling them to JavaScript or use the `transpilePackages` option.

**This is where you come in!**

To be able to _make moves_ and fix these issues, we need a more structured approach to tackling them. In various GitHub issues people mentioned multiple, distinct setups and their respective bugs. There doesn't seem to be _one_ single issue or setup that doesn't work.

Therefore we'd like to ask you to:

**Send in your minimal reproductions of your bugs with Mastra + TypeScript monorepos.** This is a smaller, simplified Mastra project that demonstrates your issue. Having a working reproduction in a new project helps to confirm that this is a repeatable problem, and is not caused by something else in your personal environment or existing project.

## How to create a minimal reproduction

1. Setup your project
	- Follow [Mastra's install instructions](https://mastra.ai/en/docs/getting-started/installation) and add your monorepo setup
	- Alternatively, you can fork [this example project](https://github.com/lekoarts-demos/mastra-monorepo-01) and try to reproduce your issue there
3. Verify that your minimal reproduction shows your issue
4. Share the link to the public GitHub repository in this issue
5. Write down the clear reproduction steps (e.g. "Run `mastra dev` to see this error x")

> [!TIP]
> Use the `@alpha` version of Mastra packages to receive the latest published version from our `main` branch. This way you can quickly test out bug fixes we ship.

If you can't / don't want to create a minimal reproduction but have an existing project showing a bug, we can take a look at it if:
- We can run the project without us requiring to add environment variables or having access to private npm packages
- It's a public GitHub repository

We prefer not to look at your private company project as that can come with legal headaches.

## Our game plan

With the help of your reproductions we want to squash more of these monorepo bugs. By working through them we should also be able to see if our documentation needs updates (e.g. specific instructions for how to setup a monorepo that works with Mastra) or if we wan fix it all in code.

You'll be able to test out updates for this regularly by installing the `@alpha` version.

## Related issues

[...]

PR #7943, opened 2025-09-17
fix(deployer): Workspaces on Windows

LekoArts:
## Description

This reproduction (https://github.com/mastra-ai/mastra/issues/6852#issuecomment-3299560795) failed on Windows. I booted up my Windows machine and was able to reproduce.

I added a new util `slash()` that converts double backslash to forward slash. In codepaths where we manually did this already I used `slash()` now.

The first root cause of things not working was that we didn't populate our own workspacesMap because in the comparison `ws.location === location` the different path styles were falsy. Once that was fixed it was a matter of finding other instances were we relied on the assumption for such comparisons.

## Related Issue(s)

https://github.com/mastra-ai/mastra/issues/6852

## Type of Change

- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [ ] Test update

## Checklist


packages/deployer/src/build/bundler.ts
```diff
@@ -80,19 +81,28 @@ export async function getInputOptions(
...
 
           const filename = analyzedBundleInfo.dependencies.get(id)!;
-          // also add projectRoot
-          const resolvedPath = join(workspaceRoot || projectRoot, filename);
+          const absolutePath = join(workspaceRoot || projectRoot, filename);
+
+          // During `mastra dev` we want to keep deps as external
+          if (isDev) {
+            return {
+              id: process.platform === 'win32' ? pathToFileURL(absolutePath).href : absolutePath,
```
LekoArts, line 89: Without this the imports would be `C:\\users` which ESM doesn't allow.

During build though Rollup didn't like these `file://` paths, so we only do this for dev

abhiaiyer91 (approved):

wardpeet (approved):

PR #8009, opened 2025-09-18
filter scorers by type

YujohnNattrass:
## Description

When scorering on a agent trace only show scorers with type `agent`

When scoring on spans filter out scorers with type `agent`

## Type of Change

- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [ ] Test update

## Checklist

### DIFF
--- a/packages/playground-ui/src/domains/scores/components/score-dialog.tsx
+++ b/packages/playground-ui/src/domains/scores/components/score-dialog.tsx
@@ -58,11 +58,15 @@ export function ScoreDialog({
                   value: <Link href={computeTraceLink(score?.traceId)}>{score?.traceId}</Link>,
                   key: 'traceId',
                 },
-                {
-                  label: 'Span ID',
-                  value: <Link href={computeTraceLink(score?.traceId, score?.spanId)}>{score?.spanId}</Link>,
-                  key: 'spanId',
-                },
+                ...(score?.spanId
+                  ? [
+                      {
+                        label: 'Span ID',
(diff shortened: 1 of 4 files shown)


packages/playground-ui/src/domains/scores/hooks/use-scorers.tsx
```diff
@@ -82,5 +82,7 @@ export const useScorers = () => {
   return useQuery({
     queryKey: ['scorers'],
     queryFn: () => client.getScorers(),
+    staleTime: 0,
+    gcTime: 0,
```
YujohnNattrass, line 86: I noticed that when navigating between spans the scorer list was not being updated.

packages/playground-ui/src/domains/scores/components/score-dialog.tsx
```diff
@@ -58,11 +58,15 @@ export function ScoreDialog({
...
-                },
+                ...(score?.spanId
+                  ? [
+                      {
+                        label: 'Span ID',
+                        value: <Link href={computeTraceLink(score?.traceId, score?.spanId)}>{score?.spanId}</Link>,
+                        key: 'spanId',
+                      },
+                    ]
+                  : []),
```
YujohnNattrass, line 69: When scoring on a trace I used to add the parent span id as the spanID. However I have removed that so we can differentiate between scoring on a trace vs scoring on a span.

abhiaiyer91 (approved):

PR #8010, opened 2025-09-18
fix(core): Add comprehensive types to streamVNext code path

DanielSLew:
## Description

- adds lots of types for things previously untyped or typed wrong
- Aligned the return types to what they should be (we were returning and/or missing lots of properties that would likely be undefined or missing)
- Cleaned up and consolidated dead code
- Tried to use existing types and clean up as much as possible

TODO: Fix up docs with the public reference types for MastraModelOutput

## Type of Change

- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [ ] Test update

## Checklist


packages/core/src/agent/agent.ts
```diff
@@ -3073,19 +3073,19 @@ export class Agent<
...
-          object: result?.object,
-          toolResults: result?.toolResults,
-          toolCalls: result?.toolCalls,
-          usage: result?.usage,
+          stepType: s.stepType,
+          text: result.text,
+          object: result.object,
+          toolResults: result.toolResults,
+          toolCalls: result.toolCalls,
+          usage: result.usage,
```
epinzur: should these be all s.property?

seems odd to return the same text, object, toolResults for each step.

DanielSLew: hmm, this is what it was before, but that's probably true, it's fairly harmless as this is just a debug log

packages/core/src/agent/agent.ts
```diff
@@ -3124,9 +3124,7 @@ export class Agent<
         }
 
         if (responseMessages) {
-          // @TODO: PREV VERSION DIDNT RETURN USER MESSAGES, SO WE FILTER THEM OUT
-          const filteredMessages = responseMessages.filter((m: any) => m.role !== 'user');
-          messageList.add(filteredMessages, 'response');
+          messageList.add(responseMessages, 'response');
```
DanielSLew, line 3174: @TylerBarnes it seems weird that we had to filter this, as response messages don't have user messages in it

packages/core/src/stream/types.ts
```diff
@@ -74,32 +82,37 @@ interface SourcePayload {
...
   mimeType: string;
   providerMetadata?: SharedV2ProviderMetadata;
 }
 
-interface ToolCallPayload {
+export interface ToolCallPayload<TArgs = unknown, TOutput = unknown> {
   toolCallId: string;
   toolName: string;
-  args?: Record<string, any>;
+  input?: TArgs;
```
CalebBarnes: Not 100% sure about this change but I think we should either keep this as args or if we change to input then we should update everywhere that currently uses args to input.

packages/core/src/loop/__snapshots__/loop.test.ts.snap
```diff
@@ -1368,48 +1364,73 @@ exports[`Loop Tests > AISDK v5 > result.pipeUIMessageStreamToResponse > should s
...
+        "testProvider": {
+          "signature": "1234567890",
+        },
       },
+      "text": "I will open the conversation with witty banter.",
     },
-    "text": "I will open the conversation with witty banter.",
+    "runId": "test-run-id",
     "type": "reasoning",
   },
```
CalebBarnes, line 1374: This snapshot looks like it was updated to be in the mastra format but isn't this for an aisdk test? "AISDK v5"

TylerBarnes (approved):

PR #8027, opened 2025-09-19
added ai-tracing support to postgres storage

epinzur:
## Description

Implements complete AI tracing/observability support for the PostgreSQL storage
adapter, matching functionality available in LibSQL store. 

Also:
* Changes SpanType from `int` to `string` throughout all stores, so that we don't need to worry about SpanType enum order changing
* Stops setting `createdAt` and `updatedAt` in the exporter... these should be set by the database in most cases.

The postgres code was entirely generated by Claude Code, pls check it carefully.

## Related Issue(s)

#6773

## Type of Change

- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [ ] Test update

## Checklist

### DIFF
--- a/stores/pg/src/storage/domains/observability/index.ts
+++ b/stores/pg/src/storage/domains/observability/index.ts
@@ -27,23 +27,16 @@ export class ObservabilityPG extends ObservabilityStorage {
 
   async createAISpan(span: AISpanRecord): Promise<void> {
     try {
-      // Convert Date objects to ISO strings for PostgreSQL
       const startedAt = span.startedAt instanceof Date ? span.startedAt.toISOString() : span.startedAt;
-      const endedAt = span.endedAt instanceof Date ? span.endedAt?.toISOString() : span.endedAt;
-      const createdAt = span.createdAt instanceof Date ? span.createdAt.toISOString() : span.createdAt;
-      const updatedAt = span.updatedAt instanceof Date ? span.updatedAt?.toISOString() : span.updatedAt;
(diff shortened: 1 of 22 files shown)


epinzur:
@greptileai re-review pls

stores/pg/src/storage/domains/observability/index.ts
```diff
@@ -0,0 +1,353 @@
...
+      parentSpanId: null, // Only get root spans for traces
+    };
+
+    const whereClause = prepareWhereClause(filtersWithDateRange, AI_SPAN_SCHEMA);
+
+    let actualWhereClause = whereClause.sql;
+    let currentParamIndex = whereClause.args.length + 1;
+
+    // Handle entity filtering
+    if (entityId && entityType) {
```
YujohnNattrass, line 186: The frontend is already filtering by entityId and entityType. We may want an index for entityType + entityId + started At. Then we should make sure the sql query being created has those statements args in the right order to hit the index.

What do you think?

YujohnNattrass (dismissed):
Overall looks good, we should fix this bug I found with endedAt being null and we can merge:

image

image

stores/pg/src/storage/domains/observability/index.ts
epinzur: these don't exist in the tableSchema yet, so we can't add an index for them.

I decided to do the update to add them to the schema in a future PR, and modify all stores at that point.

epinzur:
@greptileai

YujohnNattrass (approved):

PR #8033, opened 2025-09-19
Batch trace scoring implementation and internal event based workflow support

YujohnNattrass:
## Description

Add Trace Scoring System with Batch Processing Support

Overview
This PR introduces a trace scoring system to Mastra, enabling batch processing of traces with configurable scorers and workflow-based execution.

Also adds internal workflows to mastra instance so we can use event based workflows internally

Core Features
•  Batch Trace Scoring: New functionality to score multiple traces in batch operations
•  Utility Functions: Comprehensive utilities for building scorer payloads for agent-type scorers
•  Internal Workflows: Enhanced Mastra core with internal workflow capabilities

Key Components Added
•  scoreBatchTraces/ module with complete workflow implementation

## Type of Change

- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [ ] Test update

## Checklist


packages/core/src/scores/scoreBatchTraces/scoreBatchTracesWorkflow.ts
```diff
@@ -0,0 +1,229 @@
...
+
+  const scorerRun = buildScorerRun({
+    scorerType: scorer.type === 'agent' ? 'agent' : undefined,
+    tracingContext,
+    trace,
+    targetSpan: span,
+  });
+
+  const result = await scorer.run(scorerRun);
+  const traceId = `${target.traceId}${target.spanId ? `-${target.spanId}` : ''}`;
```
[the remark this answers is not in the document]
YujohnNattrass: Trace ID temporarily contains both trace id and span id for now, until we update the storage adapters and add a spanId column

packages/core/src/scores/scoreBatchTraces/utils.ts
```diff
@@ -0,0 +1,326 @@
...
+    const systemMessages = extractSystemMessages(primaryLLMSpan);
+
+    // Extract remembered messages from LLM span (excluding current input)
+    const currentInputContent = inputMessages[0]?.content || '';
+    const rememberedMessages = extractRememberedMessages(primaryLLMSpan, currentInputContent);
+
+    return {
+      // We do not keep track of the tool call ids in traces, so we need to cast to UIMessageWithMetadata
+      inputMessages: inputMessages as UIMessageWithMetadata[],
+      rememberedMessages: rememberedMessages as UIMessageWithMetadata[],
```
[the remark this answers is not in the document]
YujohnNattrass: spans do not contain Ids for specific spans so we omit them

packages/core/src/scores/scoreBatchTraces/utils.test.ts
```diff
@@ -0,0 +1,439 @@
...
+    it('should include tool invocations in response', () => {
+      const trace = TransformerTestScenarios.agentWithToolCalls().buildTrace();
+      const result = transformTraceToScorerOutput(trace);
+
+      expect(result[0]?.toolInvocations).toHaveLength(1);
+      expect(result[0]?.toolInvocations?.[0]?.toolName).toBe('weatherAPI');
+      expect(result[0]?.toolInvocations?.[0]?.args).toEqual({ location: 'Seattle' });
+
+      // @ts-ignore
+      expect(result[0]?.toolInvocations?.[0]?.result).toEqual({ temperature: 72, condition: 'sunny' });
```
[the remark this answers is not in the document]
YujohnNattrass: Since spans do not contain all the information to rebuild the UIMessages some of the types in utils and utils.test will be ignored.

packages/core/src/mastra/index.ts
```diff
@@ -489,6 +491,19 @@ do:
...
+    };
+    Object.entries(internalWorkflows).forEach(([key, workflow]) => {
+      workflow.__registerMastra(this);
+      workflow.__registerPrimitives({
+        logger: this.getLogger(),
+      });
+      // @ts-ignore
+      this.#internalMastraWorkflows[key] = workflow;
+    });
+
```
epinzur: i think it would be better if the scorer code registered its workflow with the mastra instance as being internal... 

Instead of importing the workflow into mastra constructor and hard coding it as internal.

packages/core/src/mastra/index.ts
```diff
@@ -684,6 +699,28 @@ do:
...
+        details: {
+          status: 404,
+          workflowId: String(id),
+        },
+      });
+    }
+
+    return workflow;
+  }
+
```
epinzur, line 718: would be cool to add a `__getInternalWorkflow()` or `__registerInternalWorkflow()` method that could be used in scorers and other systems to register workflows as internal.

packages/core/src/scores/scoreBatchTraces/scoreBatchTracesWorkflow.ts
```diff
@@ -0,0 +1,233 @@
...
+    scoreId: scoreRecord.id,
+    scorerName: scoreRecord.scorer.name,
+    score: scoreRecord.score,
+    createdAt: scoreRecord.createdAt,
+  };
+  await storage.updateAISpan({
+    spanId: span.spanId,
+    traceId: span.traceId,
+    updates: { links: [...existingLinks, link] },
+  });
```
epinzur, line 209: how will this work for databases like clickHouse that won't have `updateAISpan()` implemented.

packages/core/src/scores/scoreBatchTraces/scoreBatchTracesWorkflow.ts
```diff
@@ -0,0 +1,233 @@
...
+}
+
+export const scoreBatchTracesWorkflow = createWorkflow({
```
epinzur: do we need 'batch' in the exported name?  doesn't plural 'Traces' infer it's a batch?

packages/core/src/scores/scoreBatchTraces/utils.ts
```diff
@@ -0,0 +1,324 @@
+import type { ToolInvocation } from 'ai';
```
epinzur, line 1: on the file naming, seems like this folder can just be `scoreTraces`

packages/core/src/scores/scoreBatchTraces/utils.ts
```diff
@@ -0,0 +1,324 @@
...
+  parentSpanId: string,
+  spanType: AISpanType,
+): T[] {
+  const children = spanTree.childrenMap.get(parentSpanId) || [];
+  return children.filter(span => span.spanType === spanType) as T[];
+}
+
+/**
+ * Convert AI SDK v5 message content to v4 format
+ */
```
epinzur, line 67: don't we already have helpers defined elsewhere to do this format conversion?

packages/core/src/scores/scoreBatchTraces/utils.ts
```diff
@@ -0,0 +1,324 @@
...
+    if (span.parentSpanId && !spanIds.has(span.parentSpanId)) {
+      throw new Error(`Span ${span.spanId} references non-existent parent ${span.parentSpanId}`);
+    }
+  }
+}
+
+/**
+ * Find the most recent LLM span that contains conversation history
+ */
+function findPrimaryLLMSpan(spanTree: SpanTree, rootAgentSpan: AISpanRecord): AISpanRecord {
```
epinzur, line 240: there should only be a single `LLM_GENERATION` span per agentSpan... and it should be a direct child of the `rootAgentSpan`.

if there are others in the trace, those are associated with nested agents.

packages/core/src/scores/scoreBatchTraces/utils.ts
```diff
@@ -0,0 +1,324 @@
...
+
+    // Extract system messages from LLM span
+    const systemMessages = extractSystemMessages(primaryLLMSpan);
+
+    // Extract remembered messages from LLM span (excluding current input)
+    const currentInputContent = inputMessages[0]?.content || '';
+    const rememberedMessages = extractRememberedMessages(primaryLLMSpan, currentInputContent);
+
+    return {
+      // We do not keep track of the tool call ids in traces, so we need to cast to UIMessageWithMetadata
```
epinzur, line 274: we could start keeping track of the tool call ids in traces if that would be helpful for scoring.

packages/core/src/scores/scoreBatchTraces/utils.ts
```diff
@@ -0,0 +1,324 @@
...
+    throw new Error(
+      `Failed to transform trace to scorer input: ${error instanceof Error ? error.message : 'Unknown error'}`,
+    );
+  }
+}
+
+/**
+ * Transform trace to scorer output format
+ */
+export function transformTraceToScorerOutput(trace: AITraceRecord): ScorerRunOutputForAgent {
```
epinzur, line 290: should you build the input & output data for scoring in the same method?

you seem to be duplicating a bunch of checks & queries in each method.

packages/core/src/mastra/index.ts
YujohnNattrass: Makes sense, great idea!

packages/core/src/scores/scoreBatchTraces/scoreBatchTracesWorkflow.ts
YujohnNattrass: Good point. For some context `links` here is being used by the front end to find which scores belong to a span.

A better long term solution is to create a storage api for the front end to fetch scores by trace & span id. Which would also include adding a column for `spanId`.

Created a fast follow ticket for it: https://linear.app/kepler-crm/issue/CLOUD-497/add-storage-api-to-fetch-scores-by-trace-and-span-id

packages/core/src/scores/scoreBatchTraces/utils.ts
YujohnNattrass: If users have tool call id's as part of their scoring process it will be useful. However I highly doubt users would use the tool call id.

I typecasted mainly because the tool call id was missing so TS complained about it. If you think users would use that then I can definitely make that change in a follow up.

packages/core/src/scores/scoreBatchTraces/scoreBatchTracesWorkflow.ts
YujohnNattrass: makes sense!

packages/core/src/scores/scoreBatchTraces/utils.ts
YujohnNattrass: You are right validating the trace, building the span tree and finding the parent span in each method.

I'll create a fast follow ticket to remove the duplication, but for now I would say we can ship this as is.

YujohnNattrass: https://linear.app/kepler-crm/issue/CLOUD-500/reduce-the-duplication-when-creating-the-scorer-run-input-and-output

rase- (approved):

PR #8069, opened 2025-09-22
fix(core): Fix Gemini message ordering validation errors (#7287, #8053)

### LINKED ISSUE #7287: [DISCORD:1410989513076899963] Gemini function-call turn ordering errors after upgrading Mastra (0.10.13 → 0.10.23)
This issue was created from Discord post: https://discord.com/channels/1309558646228779139/1410989513076899963

Hey Mastra Team!

We upgraded Mastra 0.10.13 → 0.10.23 (with Gemini Flash via @ai-sdk/google) and started intermittently getting:
“Please ensure that function call turn comes immediately after a user turn or after a function response turn.”

Setup:
- agent.stream(...) with clientTools { citeSourcesTool, ..., proposeSuggestionsTool }, onFinish persists messages.
- Tool-backed RAG on the agent.

Behavior:
- Error seems to appear when assistant streams text and then a tool call (e.g., citeSources or proposeSuggestions) is triggered in the same assistant turn.
- Same app code on 0.10.13 didn’t surface this.

Questions:
1) Any stream/turn sequencing changes between 0.10.13 and 0.10.23 that affect tool-call ordering with Gemini?
2) Recommended pattern for “post-answer” tools (e.g., separate follow-up call in onFinish) or a config to ensure tools execute before any streamed text?
3) Would streamVNext help here?

Notes:
- Changing maxSteps/toolChoice or moving tools to tools vs clientTools didn’t help.
- Didn’t find anything relevant in release notes; no minimal repro yet, but I can DM redacted raw turn logs.

taofeeq-deru:
## Description

Fixes Gemini API validation errors by ensuring proper message ordering in the message list for Gemini models. This PR addresses specific error cases that occur when using Gemini models with edge case message patterns.

**Breaking Change**: Empty or system-only message lists now throw an error instead of adding a placeholder message. This prevents confusing LLM responses to placeholder "." messages and provides clear feedback to developers.

## Related Issue(s)

Fixes #7287 - "Please ensure that function call turn comes immediately after a user turn"
Fixes #8053 - "Please ensure that single turn requests end with a user role"

### Core Fix: Message List Updates (`packages/core/src/agent/message-list/`)

Created new utility function `ensureGeminiCompatibleMessages()` in `packages/core/src/agent/message-list/utils/ai-v5/gemini-compatibility.ts` and integrated it into both `prompt()` and `llmPrompt()` methods for AI SDK v4 and v5.

**Ensure first non-system message is user role**
- When the first non-system message is `assistant` or `assistant-with-tool-call`, inserts a placeholder user message before it
- When only system messages exist or the array is empty, throws an error: "This request does not contain any user or assistant messages. At least one user or assistant message is required to generate a response."
- Prevents **Issue #7287**: "function call turn comes immediately after a user turn" error
- Prevents **Issue #8053**: "single turn requests end with a user role" error  
- Handles memory truncation scenarios where tool-call history starts with assistant
- Also fixes "contents is not specified" error by ensuring at least one message with valid content exists
- Uses TypeScript generics to support both v4 `CoreMessage[]` and v5 `ModelMessage[]`

Through testing with real Gemini API calls, we confirmed that:
- `[assistant]` → `[user(.), assistant]` ✅ Accepted by Gemini
- `[system, assistant]` → `[system, user(.), assistant]` ✅ Accepted by Gemini
- `[user, assistant, tool]` → No change needed ✅ Already accepted by Gemini
- `[]` or `[system]` → Error thrown ⚠️ BREAKING CHANGE

Gemini accepts messages ending with assistant or tool roles as long as a user message exists in the conversation, so only ensuring the first non-system message is from the user is necessary.

### Test Coverage

Added and updated 29 tests total:
- **message-list-gemini.test.ts**: 4 new error-throwing tests (12/12 passing)
- **message-list.test.ts**: 4 new validation tests (79/79 passing)
- **message-list-v5.test.ts**: 4 updated tests expecting errors (59/59 passing, 2 skipped)
- **agent-gemini.test.ts**: 1 updated integration test (16/16 passing)
- All tests cover both `.generate()` and `.network()` methods with real Gemini API calls

## Type of Change

- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [ ] Test update

## Checklist


leosilberg:
Hi just checking if there's an ETA for getting this merged? 

Thanks

packages/core/src/agent/message-list/index.ts
```diff
@@ -218,21 +218,36 @@ export class MessageList {
...
+        // Check 1: Ensure first non-system message is user
+        const firstNonSystemIndex = messages.findIndex(m => m.role !== 'system');
+        if (firstNonSystemIndex === -1) {
+          // Only system messages or empty, add user message at end
+          messages.push({
             role: 'user',
             content: '.',
-          };
-          messages.unshift(defaultMessage);
+          });
```
TylerBarnes: This is ok but I kinda wonder if we should just throw if there are no messages. Cause if you only have a system prompt it's a useless request, the LLM will just respond and say something like "I saw you just wrote '.', how can I help you?"

packages/core/src/agent/message-list/index.ts
```diff
@@ -218,21 +218,36 @@ export class MessageList {
...
+        } else if (messages[firstNonSystemIndex]?.role === 'assistant') {
+          // First non-system is assistant, insert user message before it
+          messages.splice(firstNonSystemIndex, 0, {
+            role: 'user',
+            content: '.',
+          });
```
TylerBarnes: this case makes sense to me because there are more messages, it's just that gemini is weird and requires an initial user message

packages/core/src/agent/message-list/index.ts
```diff
@@ -218,21 +218,36 @@ export class MessageList {
...
+        }
+
+        // Check 2: Ensure last message is user
+        if (messages.length > 0 && messages[messages.length - 1]?.role !== 'user') {
+          messages.push({
+            role: 'user',
+            content: '.',
+          });
```
TylerBarnes: this I'm not sure about. Why does the last message have to be a user message? Is it that the error from gemini is confusing? Because agentic requests by definition will have multiple requests where the last message is not a user message.
Even with the logic added here, once the loop starts, many requests will be made where the last message is not a user message. So I think we shouldn't add this - if we're seeing an error it's likely just that the gemini API is giving us a confusing error message

packages/core/src/agent/message-list/index.ts
```diff
@@ -295,14 +310,28 @@ export class MessageList {
...
+        }
+
+        // Check 2: Ensure last message is user
+        if (messages.length > 0 && messages[messages.length - 1]?.role !== 'user') {
+          messages.push({
+            role: 'user',
+            content: '.',
```
TylerBarnes: maybe we should make a method since we're duplicating this logic? Want to avoid folks doing a bug fix on one of these in the future and not the other

TylerBarnes (changes requested):

packages/core/src/agent/message-list/index.ts
NikAiyer: I'll double check how the tests function without this. It could be that the error is misleading. We get this usually:
`Please ensure that single turn requests end with a user role or the role field is empty` which could be leading us astray.

TylerBarnes: I think the key is probably "single turn"

TylerBarnes: https://www.perplexity.ai/search/what-does-please-ensure-that-s-Qy_NfrLMRZOWlFqTvoCtBQ#0

TylerBarnes: "means that when you send a single message (single-turn request) to the API, the last entry in your request must be from the user"

TylerBarnes: so the problem is sending a single assistant message to gemini

NikAiyer: Yeah I can confirm that all the tests seem to pass even when I remove that check so it's probably what you say.

rase- (approved):

PR #8122, opened 2025-09-23
fix(core): generateVNext tripwrire

DanielSLew:
## Description

Because we were casting the return types as unknown, we missed that the response we return for the tripwire case is actually a stream response, not a generate response.

I also removed the condition for returning tripwire as it should just return a normal response, there shouldn't be anything special with the return for it.

## Type of Change

- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [ ] Test update

## Checklist

### DIFF
--- a/.changeset/quiet-symbols-move.md
+++ b/.changeset/quiet-symbols-move.md
@@ -1,0 +1,5 @@
+---
+'@mastra/core': patch
+---
+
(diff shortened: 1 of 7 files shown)


CalebBarnes (approved):
Awesome

PR #8129, opened 2025-09-23
fix navigating between scores and entity types

YujohnNattrass:
## Description

Instead of using scorerId (the ID of the scorer from the scorer config object), we will now use the scorer name.

This PR fixes two bugs:

Navigating between scorers and entities was broken.

The scorers page displayed multiple scorers for the same scorer instance.

For example, if you had two agents like this:
```
const scorer = createScorer({ name: 'test-scorer' })
const agent1 = new Agent({ ...scorer: { scorerForAgent1: { scorer } } })
const agent2 = new Agent({ ...scorer: { scorerForAgent2: { scorer } } })
```

The scorers page would show scorerForAgent1 and scorerForAgent2 as separate scorers.

This PR changes the logic so that we use the scorer name (test-scorer) when collecting scores, preventing duplicates.

heres a clip to show the navigation works

https://github.com/user-attachments/assets/4ba8bfce-d719-4dbe-9350-a427fe486b51

## Type of Change

- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [ ] Test update

## Checklist

### DIFF
--- a/.changeset/ready-houses-stand.md
+++ b/.changeset/ready-houses-stand.md
@@ -1,0 +1,6 @@
+---
+'@mastra/playground-ui': patch
+'@mastra/core': patch
+---
+
+Fix navigating between scores and entity types
--- a/.changeset/ready-houses-stand.md
+++ b/.changeset/ready-houses-stand.md
@@ -1,6 +1,7 @@
 ---
 '@mastra/playground-ui': patch
 '@mastra/core': patch
+'mastra': patch
 ---
 
 Fix navigating between scores and entity types
--- a/packages/core/src/mastra/hooks.ts
+++ b/packages/core/src/mastra/hooks.ts
@@ -17,7 +17,7 @@ export function createOnScorerHook(mastra: Mastra) {
     const entityType = hookData.entityType;
     const scorer = hookData.scorer;
     try {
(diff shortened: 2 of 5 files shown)


packages/core/src/mastra/hooks.ts
```diff
@@ -43,7 +43,7 @@ export function createOnScorerHook(mastra: Mastra) {
         ...rest,
         ...runResult,
         entityId,
-        scorerId: hookData.scorer.id,
+        scorerId: hookData.scorer.name,
```
YujohnNattrass, line 46: we save scorerId in the db so we leave it as is here

TylerBarnes: is the name the id?

TylerBarnes (approved):

CalebBarnes (approved):

.changeset/ready-houses-stand.md
```diff
@@ -0,0 +1,6 @@
+---
+'@mastra/playground-ui': patch
+'@mastra/core': patch
+---
```
CalebBarnes, line 5: @YujohnNattrass I think you need to also bump `mastra` because of the `playground` package change.

PR #8178, opened 2025-09-25
feat(pg): add SSL support to connection string configuration

roaminro:
## Description

Add optional `ssl` property to connection string config type and pass it through to the pool configuration. This enables SSL/TLS encryption for PostgreSQL connections when using connection strings.

Also refactor config type guards for better type safety and replace fallback 'any' casting with proper error throwing for invalid configurations.

## Related Issue(s)

Fixes https://discord.com/channels/1309558646228779139/1420774343625150556

## Type of Change

- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [ ] Test update

## Checklist

### DIFF
--- a/stores/pg/src/storage/index.ts
+++ b/stores/pg/src/storage/index.ts
@@ -48,6 +48,7 @@ export type PostgresConfig = {
     }
   | {
       connectionString: string;
+      ssl?: boolean | ISSLConfig;
     }
   // Support Cloud SQL Connector & pg ClientConfig
   | ClientConfig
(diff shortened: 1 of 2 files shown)


stores/pg/src/storage/index.ts
```diff
@@ -138,14 +142,12 @@ export class PostgresStore extends MastraStorage {
...
-        } as any;
+        };
       } else {
         // This should never happen due to validation above, but included for completeness
-        this.#config = {
-          ...(config as ClientConfig),
-          max: (config as any).max,
-          idleTimeoutMillis: (config as any).idleTimeoutMillis,
-        } as any;
+        throw new Error(
```
DanielSLew, line 148: we should throw a mastra error here instead

DanielSLew (approved):

stores/pg/src/storage/index.ts
roaminro: there's actually a catch that's already converting the errors into MastraError a few lines below

PR #8202, opened 2025-09-26
feat: show react sdk to people

mfrachet:
## Description

Prepare a very early version of the react sdk package. It currently exposes `useAgent` that allows to use

- streamVNext
- network

and provides a function to map back (mostly accumulating) the streaming chunk into other formats. We provide support for "ThreadMessageLike" (we extend the assistant-ui type for supporting workflows and networks) and UIMessage (ai-sdk)

With some tweaking and adjustment, I was able to show a dynamic workflow execution in a greenfield assistant-ui.

This package still needs a lot of love, iteration and all but we're already using it inside the playground itself. Might be good to release an early version soon.

**Why custom and not just `useChat`?**

When using .network, we use a set of custom streaming events. It means that the main UI library are not able to deal with that. This `useAgent` hook allows to convert the network call into whatever format (UIMessage + Assistant UI are available for now) + it's closer to our primitives than what "useChat" suggests.

Also, since we're in control over what's happening at the streaming level, we can introduce any type of backend events and be able to deal with them our way.

**Example**

```tsx
import { toStreamAssistantUIMessage , useAgent, MastraReactProvider } from '@mastra/react'

// App.tsx
export default function App() {
  return (
    <MastraReactProvider baseUrl="http://localhost:4111">
      <YourComponent />
    </MastraReactProvider>
  );
}

// YourComponent.tsx
export default function YourComponent() {
  const [setMessages, messages] = useState<ThreadMessageLike[]>([]);

  const { streamVNext, network, isRunning, cancelRun } = useAgent({
    agentId: 'chefModelV2Agent',
  });

  const startStreaming = (input: string) => {
    // messages object is dynamically populated from here
    return streamVNext({
      coreUserMessages: [{ role: 'user', content: input }],
      onChunk: ({ chunk }) => {
        setMessages(currentConversation => toStreamAssistantUIMessage({ chunk, conversation: currentConversation }));
      },
    });
  };

  return <div>...</div>
}
```

## Type of Change

- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [ ] Test update

## Checklist


client-sdks/react/src/agent/lib/assistant-ui/utils/mapWorkflowStreamChunkToWatchResult.ts
```diff
@@ -0,0 +1,172 @@
+import { WorkflowWatchResult } from '@mastra/client-js';
```
mfrachet, line 1: ℹ️ already exists in the playground, moving shit around

client-sdks/react/src/agent/lib/assistant-ui/toStreamAssistantUIMessage.ts
```diff
@@ -0,0 +1,289 @@
+import { ChunkType } from '@mastra/core';
```
mfrachet, line 1: ℹ️ already exists in the playground, moving shit around

client-sdks/react/src/agent/hooks.ts
```diff
@@ -4,7 +4,9 @@ import { MastraClient } from '@mastra/client-js';
 import { CoreUserMessage } from '@mastra/core/llm';
 import { RuntimeContext } from '@mastra/core/runtime-context';
 import { ChunkType, NetworkChunkType } from '@mastra/core/stream';
-import { useState } from 'react';
+import { UIMessage } from 'ai';
```
adeleke5140: do we only need the types from `ai`? Curious about how much it would add to our bundle.

mfrachet: 0, it's just for the types (treeshaking should do the job here) and it's referenced as a dev deps

client-sdks/react/src/agent/lib/assistant-ui/toNetworkAssistantMessage.ts
```diff
@@ -0,0 +1,348 @@
+import { ThreadMessageLike } from '@assistant-ui/react';
```
mfrachet, line 1: ℹ️ already exists in the playground, moving stuff

PR #8229, opened 2025-09-26
Structured output stream processor

CalebBarnes:
## Description

Convert structured output from an output processor to a stream processor.

### Key Changes

- **EventEmitter-based streams**: Replaced direct stream consumption with `#createEventedStream()` in `MastraModelOutput` to enable multiple consumers without stream locking issues
- **Structured output as stream processor**: Converted structured output from an output processor to a stream processor, allowing structured agent stream chunks to be enqueued directly into the main agent stream
- Remove structuredOutput from generateLegacy

## Usage example
```ts
const stream = await agent.streamVNext('prompt', {
  structuredOutput: {
    schema: z.object({
      name: z.string(),
      age: z.number(),
    }),
  },
});

// consume fullStream with partial `object` chunks 
for await (const chunk of stream.fullStream) {
  if (chunk.type === 'text-delta') {
    console.log(chunk.payload.text);
  }
  if (chunk.type === 'object') {
    const partialObject = chunk.object
                //  ^? Partial<{ name: string; age: number }>
    console.log(partialObject);
  }
}

// OR consume objectStream directly
for await (const chunk of stream.objectStream) {
                //  ^? Partial<{ name: string; age: number }>
  console.log(chunk);
}

// Accessing result object promise
const object = await stream.object;
           // ^? { name: string; age: number; }
console.log(object);


```

## Type of Change

- [ ] Documentation update
- [x] Code refactoring
- [ ] Performance improvement
- [x] Test update

## Checklist


packages/core/src/loop/workflows/schema.ts
```diff
@@ -48,6 +48,7 @@ export interface LLMIterationOutput<Tools extends ToolSet = ToolSet> {
   dynamicToolResults?: DynamicToolResult[];
   usage: LanguageModelUsage;
   steps: StepResult<Tools>[];
+  // object?: InferSchemaOutput<OUTPUT>;
```
DanielSLew: What was this commented out for?

packages/core/src/stream/aisdk/v5/output.ts
```diff
@@ -296,61 +280,63 @@ export class AISDKV5OutputStream<OUTPUT extends OutputSchema = undefined> {
...
+          }
 
-            if ('payload' in chunk) {
-              const transformedChunk = convertMastraChunkToAISDKv5({
-                chunk,
-              });
+          if ('payload' in chunk) {
+            const transformedChunk = convertMastraChunkToAISDKv5({
+              // @ts-ignore
+              chunk, // TODO: fix types here (is not assignable to type 'ChunkType'.)
```
DanielSLew: what changed with the type here?

packages/core/src/stream/aisdk/v5/output.ts
```diff
@@ -296,61 +280,63 @@ export class AISDKV5OutputStream<OUTPUT extends OutputSchema = undefined> {
...
             }
-          },
+          }
         },
-      ),
-    );
-
-    return transformedStream as AIV5FullStreamType<OUTPUT>;
+      }),
+    ) as any as AIV5FullStreamType<OUTPUT>;
```
DanielSLew: why do we need to introduce an 'any' here?

packages/core/src/stream/base/output-format-handlers.ts
```diff
@@ -287,7 +291,10 @@ class ArrayFormatHandler<OUTPUT extends OutputSchema = undefined> extends BaseFo
     return { shouldEmit: false };
   }
 
-  async validateAndTransformFinal(_finalValue: string): Promise<ValidateAndTransformFinalResult<OUTPUT>> {
+  async validateAndTransformFinal(
+    _finalValue: string,
+    _previousObject: any,
```
DanielSLew: I don't see previousObject being used in this function at all? I was going to complain that it's typed as `any` but I think it can just be removed?

packages/core/src/stream/base/output-format-handlers.ts
```diff
@@ -544,29 +550,51 @@ export function createObjectStreamTransformer<OUTPUT extends OutputSchema = unde
...
-        // Not JSON mode, no final validation needed - exit
+      // Bypass final validation if there is no output schema provided
+      // or if we are not in the LLM execution step (inner stream)
+      if (!isLLMExecutionStep || !schema) {
         return;
       }
 
       if (['tool-calls'].includes(finishReason ?? '')) {
-        onFinish(undefined as InferSchemaOutput<OUTPUT>);
+        controller.enqueue({
```
DanielSLew, line 554: I'm not sure I understand this change here

packages/core/src/stream/base/output-format-handlers.ts
```diff
@@ -544,29 +550,51 @@ export function createObjectStreamTransformer<OUTPUT extends OutputSchema = unde
...
+          from: ChunkFrom.AGENT,
+          runId: currentRunId ?? '',
+          type: 'object-result',
+          object: undefined as InferSchemaOutput<OUTPUT>,
+        });
         return;
       }
 
-      const finalResult = await handler.validateAndTransformFinal(accumulatedText);
+      const finalResult = await handler.validateAndTransformFinal(accumulatedText, previousObject);
```
DanielSLew:
```suggestion
      const finalResult = await handler.validateAndTransformFinal(accumulatedText);
```

packages/core/src/stream/base/output.ts
```diff
@@ -172,22 +196,30 @@ export class MastraModelOutput<OUTPUT extends OutputSchema = undefined> extends
...
       const processorStates = new Map<string, ProcessorState>();
+
       processedStream = stream.pipeThrough(
         new TransformStream<ChunkType<OUTPUT>, ChunkType<OUTPUT>>({
           async transform(chunk, controller) {
+            /**
+             * Add base stream controller to structured output processor state
+             * so it can be used to enqueue chunks into the main stream from the structuring agent stream
+             */
+            const STRUCTURED_OUTPUT_PROCESSOR_NAME = 'structured-output';
```
DanielSLew: nitpick, can we use the same const for here and to actually name the processor?

packages/core/src/stream/base/output.ts
```diff
@@ -216,6 +255,18 @@ export class MastraModelOutput<OUTPUT extends OutputSchema = undefined> extends
               self.#status = 'suspended';
               self.#delayedPromises.suspendPayload.resolve(chunk.payload);
               break;
+            case 'raw':
```
DanielSLew, line 259: what is this? Were we just missing emitting this entirely?

packages/core/src/stream/aisdk/v5/output.ts
CalebBarnes: Oops, I must have accidentally included this when merging.

packages/core/src/stream/aisdk/v5/output.ts
CalebBarnes: `ChunkType` takes the `OUTPUT` generic now. So now I updated `convertMastraChunkToAISDKv5` to use that generic instead of just `ChunkType`. Thanks!

packages/core/src/loop/workflows/schema.ts
CalebBarnes: I had to pass OUTPUT generic through all of these types and update anywhere that was using them. Added now! Thanks.

packages/core/src/stream/base/output.ts
CalebBarnes: Previously this logic was in its own stream transformer on the  fullStream only. Now all of this kind of processing is moved up to the baseStream transformer. 

You can see this in main here https://github.com/mastra-ai/mastra/blob/a7243e2e58762667a6e3921e755e89d6bb0a3282/packages/core/src/stream/base/output.ts#L809-L812

packages/core/src/stream/base/output-format-handlers.ts
CalebBarnes: On main the `createObjectTransformer` function takes an `onFinish` function and uses that to resolve the `object` promise. 

With these new changes the `onFinish` callback was removed now that any chunk with type `object-result` will just resolve the `object` promise. This makes it more flexible, allowing us to easily resolve the `object` promise by enqueuing `object-result` chunks from anywhere else like stream processors / structuring agent.

In this specific part of the code, we're just handling the case where the stream ends early to call client side tools. In that case we just resolved undefined to avoid throwing errors.

CalebBarnes:
Thanks for the review @DanielSLew !

I've addressed this feedback:
  1. ✅ **Removed unused `previousObject` parameter** from
  `validateAndTransformFinal` - it wasn't being used anywhere in the method (commit
   25e9e10d25)

  2. ✅ **Used shared constant for processor name** - extracted
  `STRUCTURED_OUTPUT_PROCESSOR_NAME` and used it consistently in both the processor
   definition and state assignment (commit 2e44a77fe4)

  3. ✅ **Removed accidental `as any` cast** - cleaned up the type assertion
  (commit 4ea6f35055)

  4. ✅ **Uncommented `object` field in `LLMIterationOutput`** - this is now
  properly typed with the `OUTPUT` generic parameter since structured output is
  added to the iteration output when using the processor (commit a08ba84058edf17ff56d05b98d000a7686cb216b)

  5. ✅ **Fixed `@ts-ignore` in output.ts** - the function now properly accepts
  `ChunkType<OUTPUT>` with a generic parameter (commit 05f786eab3cac60f17df97e5114bad31dc6b0933)
  
   Let me know if you'd like me to clarify anything else!

DanielSLew (approved):
woo

PR #8235, opened 2025-09-26
feat(@mastra/core): model router

TylerBarnes:
Adds a model router system that lets you use any OpenAI-compatible provider with simple magic strings. You can now specify models like `openai/gpt-4o`, `anthropic/claude-3-opus`, or `groq/llama-3.1-70b-versatile` instead of manually configuring each provider.

The system fetches provider configs from `models.dev/api.json` and auto-generates a registry with all available providers. API keys are resolved automatically from environment variables (e.g., `OPENAI_API_KEY`, `ANTHROPIC_API_KEY`).

```typescript
// Before: 
import { anthropic } from '@ai-sdk/anthropic'
const agent = new Agent({
  name: 'my-agent',
  model: anthropic("claude-4-1-opus")
});

// After: just use a magic string
const agent = new Agent({
  name: 'my-agent',
  model: 'anthropic/claude-4-1-opus' // IDE TS autocomplete here works for all known provider/models
});
```

You can also pass a full config object for a custom openai-compatible provider

```typescript
const agent = new Agent({
  name: 'my-agent',
  model: {
    modelId: 'gpt-4o',
    url: 'https://api.openai.com/v1/chat/completions',
    apiKey: 'sk-...',
    headers: { 'Custom-Header': 'value' }
  }
});
```

The `OpenAICompatibleModel` class handles SSE parsing, tool call buffering, and provider-specific headers automatically.

Limitations:

- For now it only supports the chat completions API
- We're using models.dev as the source of model information, but the provider/model file is generated manually and published in `@mastra/core`. In a follow up I'll add support for dynamically fetching the list during local dev every ~1hr. Unknown model IDs are also supported already if for example you do `openai/totally-new-model`

Immediate follow up PRs:
- Playground support
- Docs


packages/core/scripts/generate-providers.ts
```diff
@@ -0,0 +1,160 @@
...
+ */
+export type ProviderModels = typeof PROVIDER_MODELS;
+export type Provider = keyof ProviderModels;
+export type ModelForProvider<P extends Provider> = ProviderModels[P][number];
+
+/**
+ * OpenAI-compatible model ID type
+ * Full provider/model paths (e.g., "openai/gpt-4o", "anthropic/claude-3-5-sonnet-20241022")
+ */
+export type ModelRouterModelId = {[P in Provider]: \`\${P}/\${ModelForProvider<P>}\`}[Provider] | (string & {});
```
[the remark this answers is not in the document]
TylerBarnes: It does the opposite actually, the point is to get TS to show autocomplete for the provider/model list but also allow unknown strings since we don't know every possible model id @greptileai

abhiaiyer91 (approved):
🔥

packages/core/src/llm/model/openai-compatible.ts
```diff
@@ -0,0 +1,600 @@
...
+  apiKeyHeader?: string,
+  customHeaders?: Record<string, string>,
+  provider?: string,
+): Record<string, string> {
+  const headers: Record<string, string> = {
+    'Content-Type': 'application/json',
+    ...customHeaders,
+  };
+
+  if (apiKey) {
```
DanielSLew, line 38: I'm confused here, didn't we have this exact same logic already in models-dev.ts? Are we doing this twice?

packages/core/src/llm/model/openai-compatible.ts
```diff
@@ -0,0 +1,600 @@
...
+  async doStream(options: LanguageModelV2CallOptions): Promise<{
+    stream: ReadableStream<LanguageModelV2StreamPart>;
+    request?: { body: string };
+    response?: { headers: Record<string, string> };
+    warnings: LanguageModelV2CallWarning[];
+  }> {
+    this.validateApiKey(); // Validate API key before making the request
+    const { prompt, tools, toolChoice, providerOptions } = options;
+
+    // TODO: real body type, not any
```
DanielSLew, line 343: I assume this type will come from OpenAI?

packages/core/src/llm/model/openai-compatible.ts
```diff
@@ -0,0 +1,600 @@
...
+              if (line.startsWith('data: ')) {
+                try {
+                  const data: OpenAIStreamChunk = JSON.parse(line.slice(6));
+
+                  // Send response metadata from first chunk
+                  if (!sentStart && data.id) {
+                    controller.enqueue({
+                      type: 'response-metadata',
+                      id: data.id,
+                      modelId: data.model || (this as any).modelId,
```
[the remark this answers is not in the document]
DanielSLew: this seems right

DanielSLew (approved):

packages/core/src/llm/model/openai-compatible.ts
TylerBarnes: yes most likely, or we can make an expanded one if there are some things other providers also support in this endpoint (not 100% sure)

packages/core/src/llm/model/openai-compatible.ts
TylerBarnes: the models.dev provider only runs right now when we run the generate providers script. I'm going to clean it up in a future PR when I make it live fetch/update the provider/model list during local dev

TylerBarnes:
OpenCode says: Thanks for the thorough review @greptileai! We've addressed all the issues you identified:

**1. Multi-slash model ID parsing (generate-providers.ts:105-108)**
Fixed in commit c874498db6. Changed the logic to use `indexOf('/')` instead of `lastIndexOf('/')` so that for multi-slash model IDs like 'fireworks/accounts/etc/model', we correctly extract provider='fireworks' and modelId='accounts/etc/model'. Added a test case to verify this behavior.

**2. Type assertion for modelId (openai-compatible.ts:464)**
Fixed in commit 548d65cc2c. The root cause was that `modelId` was declared `readonly` but never assigned in the constructor. We now properly assign `this.modelId` in the constructor, and captured it in a local variable before the ReadableStream definition to avoid `this` context issues. No more `(this as any)` needed!

**3. Fallback 'unknown' ID (agent.ts:2912)**
Fixed in commit 2631a9aadd. Instead of falling back to 'unknown', we now throw a descriptive `MastraError` when we can't determine a model ID. This makes debugging much easier.

**4. global.fetch override in tests (models-dev.integration.test.ts:119)**
No fix needed! The test already uses a `try/finally` block to properly restore the original fetch, so it's safe even if tests run in parallel or the test fails.

All commits have been pushed to the branch. Thanks again for the detailed feedback!

OpenCode

packages/core/src/agent/agent.ts
```diff
@@ -792,126 +800,88 @@ export class Agent<
...
-            category: ErrorCategory.USER,
-            details: {
-              agentName: this.name,
-            },
-            text: `[Agent:${this.name}] - Only v2 models are allowed when an array of models is provided`,
-          });
-          this.logger.trackException(mastraError);
-          this.logger.error(mastraError.toString());
-          throw mastraError;
-        }
```
NikAiyer, line 851: Do we still need to handle these kinds of checks? Or are we just routing to prepareModels now instead of dealing with an array of models in getModel?

TylerBarnes: A lot of the logic here was redundant because it checked array + functions + single model + functions.
I refactored it to just process one array item fully and then if it's an array call it on the used array model

TylerBarnes: so we had multiple error messages for similar code paths before, there are fewer error messages now but they cover the same paths as before

NikAiyer: Cool makes sense!

PR #8238, opened 2025-09-26
fix: workflow events compatibility with AI SDK v5

elliotmassen:
## Description

Currently when you try to use an agent that runs a workflow with AI SDK v5, the connection is dropped once the front-end processes a `workflow-*` event type as it doesn't match the [schema](https://github.com/vercel/ai/blob/55e094b8b85581b284fe8e55557542489ff4c2eb/packages/ai/src/ui-message-stream/ui-message-chunks.ts#L14-L150). This means that the frontend only receives and renders events prior to the workflow starting and then stops unexpectedly. 

This PR fixes this issue by updating the AI SDK v5 compatibility support to convert any `workflow-*` event types to `data-workflow-*`.  `data-*` is Vercel's [documented](https://ai-sdk.dev/docs/ai-sdk-ui/stream-protocol?utm_source=chatgpt.com#data-parts) way of introducing custom event data.

This PR can be manually tested by registering a [chat route](https://mastra.ai/en/docs/frameworks/agentic-uis/ai-sdk#enabling-stream-compatibility) for an agent that runs a workflow in Mastra config like so:

```typescript
  agents: { 'test-agent': testAgent },
  workflows: { 'test-workflow': testWorkflow },
  ...,
  server: {
    apiRoutes: [chatRoute({ path: '/chat', agent: 'test-agent' })],
  },
})
```

If you have a front-end that use's Vercel's `useChat` function with a `/chat` endpoint as shown above, then you should see that without this fix the frontend does **not** receive any data after the workflow starts, and that _with_ this fix the frontend receives all expected data.

## Related Issue(s)

This PR fixes: https://github.com/mastra-ai/mastra/issues/8013

## Type of Change

- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [ ] Test update

## Checklist


wardpeet:
Thank you for the PR but we're moving everything to our AI sdk package - https://github.com/mastra-ai/mastra/pull/7829

Closing this one in favor of the above PR. Thank you for your contribution!

Issue #8195, opened 2025-09-26
[FEATURE] Expose traceId in streamVnext and generateVNext


khelif96:
### Problem Statement

The [docs state](https://mastra.ai/en/docs/observability/ai-tracing/overview#agent-trace-ids) that you can access the `traceId` for a given stream or generate call by accessing the `result.traceId` method. This doesn't appear to be supported in `streamVNext` or `generateVNext` Could we expand this feature out to these methods if they are the way forward and only method for supporting `AI SDK v5` and `LanguageModelV2` models

### Proposed Solution

Support the traceId field on `streamVnext` calls

```
const result = await agent.streamVNext(message, { format: 'aisdk' });

console.log(result.traceId); // Should correspond to the ai tracing trace id
// Do something with traceId
// (In our case return the traceId to the client so that they can use it to rate responses)
```

### Component

Agents

### Alternatives Considered

Right now we are using the old telemetry functionality which while hacky allows us to do this to extract the current span id
```
    const spanId: string;
    const stream = await agent.streamVNext(message, {
          runtimeContext,
          memory: memoryOptions,
          format: 'aisdk',
          onFinish: () => {
            const span = currentSpan();
            spanId = span?.id;
          },
        })
    );

    // Get the UIMessage stream and pipe it to user
    pipeUIMessageStreamToResponse({
      response: res,
      stream: stream.toUIMessageStream({
        messageMetadata: () => ({
          spanId,
        }),
      }),
    });
```

### Example Use Case

As a developer I want to be able to take the traceIds that are used to trace the llm calls and return them to the client.

The client has some special logic that can associate ratings with traceIds and be used to rate llm responses. 

### Verification

- [x] I have searched the existing issues to make sure this is not a duplicate
- [x] I have provided sufficient context for the team to understand the request

khelif96:
I believe this is the same feature as https://github.com/mastra-ai/mastra/pull/7663 except with support for `streamVNext`

khelif96:
Hey @abhiaiyer91 This issue and https://github.com/mastra-ai/mastra/issues/8214 are both the same issues. I'm not sure how you plan on consolidating but we can close one of them.

epinzur:
I'm confused because this test is passing, and I think this verifies that `traceId` is set on the return for `generate`, `stream`, `generateVNext`, `streamVNext`.

https://github.com/mastra-ai/mastra/blob/main/packages/core/src/ai-tracing/integration-tests.test.ts#L1286

[...]

PR #8263, opened 2025-09-29
fixed issue with missing traceId with aisdk output format

### LINKED ISSUE #8113: [DISCORD:1419984871412990013] Traces ofstructured output parser end up in different tracing context
This issue was created from Discord post: https://discord.com/channels/1309558646228779139/1419984871412990013

This report is about Langfuse AI tracing specifically. I have a workflow step, and use the steps `tracingContext` to call an agent's `generateVNext`, the agent's span is nicely put within the workflow tracing context in Langfuse. However, if you use structured output, the trace ends up as a seperate trace in Langfuse. I would prefer (and I think most if not all people would) if the structured output parser trace ended up within the context of the agent run span

epinzur:
## Description

Fixes issue with missing traceId output when using `aisdk` output format

Also disables tracing in all input/output processors (so they don't appear as separate traces until we specifically want to include them in the trace output).

## Related Issue(s)

#6773 

Fixes #8113 and #8195

## Type of Change

- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [ ] Test update

## Checklist

- [X] have added tests that prove my fix is effective or that my feature works

### DIFF
--- a/packages/core/src/processors/processors/language-detector.ts
+++ b/packages/core/src/processors/processors/language-detector.ts
@@ -4,6 +4,7 @@ import type { MastraMessageV2 } from '../../agent/message-list';
 import { TripWire } from '../../agent/trip-wire';
 import type { MastraLanguageModel } from '../../llm/model/shared.types';
 import type { Processor } from '../index';
+import { InternalSpans } from '../../ai-tracing';
 
 /**
  * Language detection result for a single text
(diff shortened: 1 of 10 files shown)


packages/core/src/processors/processors/language-detector.ts
```diff
@@ -172,6 +173,7 @@ export class LanguageDetector implements Processor {
       name: 'language-detector',
       instructions: options.instructions || this.createDefaultInstructions(),
       model: options.model,
+      options: { tracingPolicy: { internal: InternalSpans.ALL } },
```
epinzur, line 176: this disables trace output for the agent.

or really marks it "internal", which is the equivalent as disabled for now.

wardpeet (approved):

PR #8335, opened 2025-10-01
Support adding scores to traces for external providers

YujohnNattrass:
## Description

This PR Adds the method addScoreToTrace for AI Trace exporters to add scores to the trace after the run has been completed.

Screenshot of score added to langfuse workflow step:
image

Screenshot of score added for agent:
image

## Type of Change

- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [ ] Test update

## Checklist


packages/core/src/workflows/default.ts
```diff
@@ -894,7 +894,7 @@ export class DefaultExecutionEngine extends ExecutionEngine {
             stepId: step.id,
             runtimeContext,
             disableScorers,
-            tracingContext: { currentSpan: tracingContext.currentSpan },
+            tracingContext: { currentSpan: stepAISpan },
```
YujohnNattrass, line 939: We were adding the workflow run span however scorers run on workflow steps.

packages/core/src/ai-tracing/types.ts
```diff
@@ -659,6 +659,22 @@ export interface AITracingExporter {
...
+    metadata,
+  }: {
+    traceId: string;
+    spanId?: string;
+    score: number;
+    reason?: string;
+    scorerName: string;
+    metadata?: Record<string, any>;
+  }): Promise<void>;
+
```
epinzur, line 677: Adding this method to the `AITracingExporter` makes sense. 

But I don't think we should iterate over the set of attached exporters to call this method directly.

I think the score output should flow the standard AI tracing pipeline... so that things like the score metadata can go through the SpanProcessors.

Instead I'd expect an `addScore()` event to be added to the `AISpan` interface and pipe the score event through the AI Tracing system.

packages/core/src/workflows/default.ts
epinzur: This is for tracing of the scorer run... which is currently disabled.

However I think this brings up another issue... if you have scorers defined on a step, we are currently waiting for the scorer runs to complete before moving onto the next step.

packages/core/src/mastra/hooks.ts
```diff
@@ -50,6 +51,28 @@ export function createOnScorerHook(mastra: Mastra) {
...
+                scorerName: scorerToUse.scorer.name,
+                metadata: {
+                  ...(currentSpan.metadata ?? {}),
+                },
+              });
+            }
+          },
+          { concurrency: 3 },
+        );
+      }
```
epinzur, line 75: i would have expected something here like:

```
const currentSpan = hookData.tracingContext?.currentSpan;
currentSpan.addScore({
	score: runResult.score,
    reason: runResult.reason,
    scorerName: scorerToUse.scorer.name,
    metadata: runResult.metadata,
    });
```

epinzur (approved):

epinzur (approved):

PR #8353, opened 2025-10-01
fix(deployer): Transitive deps in dev

wardpeet:
Add support for transitive dependency resolution in monorepos


packages/deployer/src/build/analyze/bundleExternals.ts
```diff
@@ -114,6 +123,8 @@ async function getInputPlugins(
 
     if (dir) {
       transpilePackagesMap.set(pkg, slash(dir));
+    } else {
```
wardpeet, line 126: need to check if this is actually necessary

eddyhdzg (approved):

abhiaiyer91 (approved):

packages/deployer/src/build/analyze/analyzeEntry.ts
```diff
@@ -14,6 +14,7 @@ import { getPackageName, getPackageRootPath, slash } from '../utils';
 import { type WorkspacePackageInfo } from '../../bundler/workspaceDependencies';
 import type { DependencyMetadata } from '../types';
 import { DEPS_TO_IGNORE } from './constants';
+import { getPackageInfo } from 'local-pkg';
```
LekoArts, line 17: Unused import

packages/deployer/src/build/analyze/bundleExternals.ts
LekoArts: 👍

LekoArts (approved):

PR #8372, opened 2025-10-01
fix: model router docs/ui fixes

TylerBarnes:
Fixed several issues with the model router documentation generation and the playground UI's model picker component.

- The docs generation script had inverted logic for selecting example models
- cleaned up the copy to be more action-oriented
- The playground UI model picker had a bug where typing in the model input would clear the provider field visually
- fixed an environment variable display bug where multiple env vars would concatenate into one string like `NETLIFY_TOKENNETLIFY_SITE_ID`. Now properly handles arrays and shows correct singular/plural text.
- Added responsive design to the model picker so it stacks nicely on mobile instead of disappearing, plus general sizing/spacing improvements for the items in the list


packages/playground-ui/src/components/assistant-ui/network-threads.tsx
```diff
@@ -116,7 +116,7 @@ const Composer: FC<{ hasMemory?: boolean }> = ({ hasMemory }) => {
           <ComposerPrimitive.Input asChild className="w-full">
             <textarea
               className="text-ui-lg leading-ui-lg placeholder:text-icon3 text-icon6 bg-transparent resize-none"
-              autoFocus
+              autoFocus={document.activeElement === document.body}
```
TylerBarnes: don't autofocus if the user clicks something (like an input)

packages/playground-ui/src/components/ui/popover.tsx
```diff
@@ -19,7 +19,8 @@ const PopoverContent = React.forwardRef<
       align={align}
       sideOffset={sideOffset}
       className={clsx(
-        'z-50 w-72 rounded-md border bg-popover p-4 text-popover-foreground shadow-md data-[state=open]:animate-in data-[state=closed]:animate-out data-[state=closed]:fade-out-0 data-[state=open]:fade-in-0 data-[state=closed]:zoom-out-95 data-[state=open]:zoom-in-95 data-[side=bottom]:slide-in-from-top-2 data-[side=left]:slide-in-from-right-2 data-[side=right]:slide-in-from-left-2 data-[side=top]:slide-in-from-bottom-2',
+        'z-50 w-72 rounded-md border bg-popover text-popover-foreground shadow-md data-[state=open]:animate-in data-[state=closed]:animate-out data-[state=closed]:fade-out-0 data-[state=open]:fade-in-0 data-[state=closed]:zoom-out-95 data-[state=open]:zoom-in-95 data-[side=bottom]:slide-in-from-top-2 data-[side=left]:slide-in-from-right-2 data-[side=right]:slide-in-from-left-2 data-[side=top]:slide-in-from-bottom-2',
+        className?.includes(` p-`) ? false : `p-4`,
```
TylerBarnes, line 23: just allowed changing p-* from outside the component

LekoArts (commented):
We should change the focus behavior of the dropdowns like so:

- Do not automatically change focus to next dropdown
- Do not automatically switch focus to chat window after selecting model

https://github.com/user-attachments/assets/5562b149-5769-4c8e-9f53-caa160ba4240

This is basically a focus trap, so with a keyboard you can never get past the model selector.

Compared to our other dropdowns these ones don't have a chevron pointing down:

Image

So we could do:

- Replace the (i) with the chevron when it's closed (provider selector)
- Add the chevron to the model selector

DanielSLew (approved):

TylerBarnes:
@LekoArts I made a story to follow up on the accessibility fixes, thanks for the advice!

PR #8407, opened 2025-10-02
feat: add progress feature support to MCPClient and fix document title positioning

karnjj:
## Description
- adds mcp progress to MCPClient
- adds example on how to use progress in examples/agent.
- adds a document about the progress property in MCPClient and fixes the wrong prompts property heading position

## Related Issue(s)

https://github.com/mastra-ai/mastra/issues/7309

## Type of Change

- [x] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [ ] Test update

## Checklist

### DIFF
--- a/.changeset/smart-dogs-grab.md
+++ b/.changeset/smart-dogs-grab.md
@@ -1,0 +1,5 @@
+---
+'@mastra/mcp': patch
+---
+
+Add progress feature support to MCPClient. Also fix wrong document title wrong position (prompts property)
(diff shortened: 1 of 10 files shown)


DanielSLew:
This looks great! Thank you for the contribution. I think my only thought is that we should just always send a progress token and the server can choose to ignore it or not, so we wouldn't need an `enableProgressTracking` boolean.

abhiaiyer91 (approved):

karnjj:
> This looks great! Thank you for the contribution. I think my only thought is that we should just always send a progress token and the server can choose to ignore it or not, so we wouldn't need an `enableProgressTracking` boolean.

I see your point, so I set the default value of enableProgressTracking to true. I also kept the option to disable it for cases where the client doesn’t want progress updates and needs to reduce network transfer from server to client.

abhiaiyer91:
Ported this here: https://github.com/mastra-ai/mastra/pull/10637/files

PR #8441, opened 2025-10-03
Add more tsdoc to memory class

wardpeet:
Update functions in memory class with tsdoc

### DIFF
--- a/packages/core/src/memory/memory.ts
+++ b/packages/core/src/memory/memory.ts
@@ -31,7 +31,7 @@ export type MemoryProcessorOpts = {
 };
 
 /**
- * Base MemoryProcessor classs for implementing message processors that can filter or transform messages
+ * Base MemoryProcessor class for implementing message processors that can filter or transform messages
  * before they're sent to the LLM. Processors can be chained together to create complex
  * message transformation pipelines.
  */
--- a/packages/core/src/memory/memory.ts
+++ b/packages/core/src/memory/memory.ts
@@ -18,20 +18,41 @@ import type {
   WorkingMemoryTemplate,
 } from './types';
 
+/**
+ * Options for processing messages through memory processors
+ */
 export type MemoryProcessorOpts = {
+  /** Optional system message to provide context for processing */
   systemMessage?: string;
+  /** Optional memory-specific system message for additional context */
   memorySystemMessage?: string;
+  /** New messages being added to the conversation */
   newMessages?: CoreMessage[];
 };
+
 /**
- * Interface for message processors that can filter or transform messages
- * before they're sent to the LLM.
+ * Base MemoryProcessor classs for implementing message processors that can filter or transform messages
+ * before they're sent to the LLM. Processors can be chained together to create complex
+ * message transformation pipelines.
  */
 export abstract class MemoryProcessor extends MastraBase {
   /**
    * Process a list of messages and return a filtered or transformed list.
-   * @param messages The messages to process
-   * @returns The processed messages
+   * Override this method to implement custom message processing logic.
+   *
+   * @param messages - The messages to process
+   * @param _opts - Processing options including system messages and new messages
+   * @returns The processed messages (can be synchronous or asynchronous)
+   *
+   * @example
+   * ```typescript
+   * class TokenLimitProcessor extends MemoryProcessor {
+   *   process(messages: CoreMessage[]): CoreMessage[] {
+   *     // Keep only last 10 messages to stay within token limits
+   *     return messages.slice(-10);
+   *   }
+   * }
+   * ```
    */
   process(messages: CoreMessage[], _opts: MemoryProcessorOpts): CoreMessage[] | Promise<CoreMessage[]> {
     return messages;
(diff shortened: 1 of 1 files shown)


packages/core/src/memory/memory.ts
```diff
@@ -71,13 +92,13 @@ export const memoryDefaultOptions = {
  * - Handles memory processors to manipulate messages before they are sent to the LLM
  */
 export abstract class MastraMemory extends MastraBase {
-  MAX_CONTEXT_TOKENS?: number;
-
   protected _storage?: MastraStorage;
-  vector?: MastraVector;
-  embedder?: EmbeddingModel<string> | EmbeddingModelV2<string>;
+  protected vector?: MastraVector;
```
wardpeet, line 96: @TylerBarnes is it ok to make these protected?

packages/core/src/memory/memory.ts
```diff
@@ -71,13 +92,13 @@ export const memoryDefaultOptions = {
  * - Handles memory processors to manipulate messages before they are sent to the LLM
  */
 export abstract class MastraMemory extends MastraBase {
-  MAX_CONTEXT_TOKENS?: number;
```
wardpeet, line 74: Are we using this @TylerBarnes ?

packages/core/src/memory/memory.ts
```diff
@@ -159,8 +250,29 @@ export abstract class MastraMemory extends MastraBase {
    * Get tools that should be available to the agent.
    * This will be called when converting tools for the agent.
    * Implementations can override this to provide additional tools.
+   *
+   * @param _config - Optional memory configuration to customize tool selection
```
wardpeet:
```suggestion
   * @param config - Optional memory configuration to customize tool selection
```

packages/core/src/memory/memory.ts
```diff
@@ -274,33 +446,93 @@ export abstract class MastraMemory extends MastraBase {
...
+   * @param text - The text to estimate tokens for
+   * @returns Estimated number of tokens
+   *
+   * @example
+   * ```typescript
+   * const tokenCount = memory.estimateTokens("Hello world!");
+   * console.log(tokenCount); // ~3
+   * ```
+   */
   estimateTokens(text: string): number {
```
wardpeet, line 462: @TylerBarnes is this one used?

packages/core/src/memory/memory.ts
TylerBarnes: I think we should avoid it for now and do it in a breaking change, I know there are people who extend the memory class and write their own implementation and this might mess with them

packages/core/src/memory/memory.ts
```diff
@@ -130,14 +155,64 @@ export abstract class MastraMemory extends MastraBase {
...
+   * @example
+   * ```typescript
+   * import { PgStorage } from "@mastra/pg";
+   *
+   * const storage = new PgStorage({
+   *   connectionString: "postgresql://..."
+   * });
+   * memory.setStorage(storage);
+   * ```
+   */
```
TylerBarnes, line 174: I think some of these methods are used by us but shouldn't be recommended to users in reference docs. For example users shouldn't call this - which of course means this class shouldn't have it or it should be something like `__setStorage`.

Memory should always be configured via constructor config or in stream/generate args

packages/core/src/memory/memory.ts
```diff
@@ -130,14 +155,64 @@ export abstract class MastraMemory extends MastraBase {
...
+   * ```typescript
+   * import { PineconeVector } from "@mastra/vector";
+   *
+   * const vector = new PineconeVector({
+   *   apiKey: "your-api-key",
+   *   environment: "us-east-1"
+   * });
+   * memory.setVector(vector);
+   * ```
+   */
```
TylerBarnes, line 196: same with this one

packages/core/src/memory/memory.ts
```diff
@@ -146,6 +221,22 @@ export abstract class MastraMemory extends MastraBase {
...
+   *
+   * @example
+   * ```typescript
+   * class CustomMemory extends MastraMemory {
+   *   async getSystemMessage({ threadId }) {
+   *     const context = await this.getContext(threadId);
+   *     return `You are assisting with: ${context.projectName}`;
+   *   }
+   * }
+   * ```
```
TylerBarnes, line 239: we shouldn't show examples of extending the memory class imo. For most people they'll have a bad time doing it

packages/core/src/memory/memory.ts
```diff
@@ -159,8 +250,29 @@ export abstract class MastraMemory extends MastraBase {
...
+   *         description: "Search through conversation history",
+   *         schema: z.object({ query: z.string() }),
+   *         execute: async ({ query }) => {
+   *           // Search implementation
+   *         }
+   *       }
+   *     };
+   *   }
+   * }
+   * ```
```
TylerBarnes, line 272: also shouldn't show this, it's part of the public API but only really meant to be called by us, 99% of users don't need to know about it

packages/core/src/memory/memory.ts
```diff
@@ -201,6 +313,23 @@ export abstract class MastraMemory extends MastraBase {
...
+   *
+   * @example
+   * ```typescript
+   * const memory = new Memory();
+   * const config = memory.getMergedThreadConfig({
+   *   lastMessages: 20,
+   *   semanticRecall: true
+   * });
+   * ```
+   */
```
TylerBarnes, line 332: same with this one

packages/core/src/memory/memory.ts
```diff
@@ -130,14 +155,64 @@ export abstract class MastraMemory extends MastraBase {
...
+   * @internal
+   *
+   * @example
+   * ```typescript
+   * import { openai } from "@ai-sdk/openai";
+   *
+   * const embedder = openai.embedding("text-embedding-3-small");
+   * memory.setEmbedder(embedder);
+   * ```
+   */
```
TylerBarnes, line 215: and this one

packages/core/src/memory/memory.ts
TylerBarnes: no looks like this was a mistake or an early version of the token processor or something

TylerBarnes (commented):
generally I think for memory the only areas that should show up in reference docs are
- any get/query/delete/update/etc methods for threads/messages
- the constructor input args
- the MemoryProcessor class

everything else shouldn't be documented and we should change how it's exposed on the class in a breaking change

PR #8546, opened 2025-10-06
fix(core): pass processors to agent server endpoints

DanielSLew:
## Description

Agent get/list endpoints in server now return a serialized version of input/output processors in order for the client to access them. We can now see what processors an agent has attached to it


mfrachet (approved):

packages/server/src/server/handlers/agents.ts
```diff
@@ -24,29 +25,81 @@ type GetHITLBody<
...
+  outputSchema?: string;
+  [key: string]: unknown;
+}
+
+export interface SerializedWorkflow {
+  name: string;
+  steps?: Record<string, { id: string; description?: string }>;
+}
+
+export interface SerializedAgent {
```
wardpeet, line 46: Should this be tied to Agent class with Pick?

DanielSLew: I looked into that, there wasn't much we could use from the class instance other than `name`, I reused some of the same types that the Agent class uses though

PR #8558, opened 2025-10-06
core: add comprehensive tests for structuredOutputs

DanielSLew:
## Description

- Add real world test schemas for structuredOutputs tests (they are skipped for now while we finish up improving structuredOutputs).
- Set maxSteps to 1 for structuredOutput processor
- don't rely on `supportsStructuredOutputs` property from model for schema compat

### DIFF
--- a/packages/core/src/processors/processors/structured-output.ts
+++ b/packages/core/src/processors/processors/structured-output.ts
@@ -92,6 +92,7 @@ export class StructuredOutputProcessor<OUTPUT extends OutputSchema> implements P
 
       const structuringAgentStream = await this.structuringAgent.stream(prompt, {
         output: this.schema,
+        maxSteps: 1,
       });
 
       const excludedChunkTypes = [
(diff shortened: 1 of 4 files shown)


packages/core/src/processors/processors/structured-output.ts
```diff
@@ -92,6 +92,7 @@ export class StructuredOutputProcessor<OUTPUT extends OutputSchema> implements P
 
       const structuringAgentStream = await this.structuringAgent.stream(prompt, {
         output: this.schema,
+        maxSteps: 1,
```
CalebBarnes, line 95: Could this break if there was reasoning before generating the text output? Also could this affect providers like anthropic that adds a tool for handling response format?

DanielSLew: I think reasoning and text would be the same step and I'm not sure about the anthropic case

PR #8559, opened 2025-10-07
feat(@mastra/core): Use ai sdk providers for openai/anthropic/google/openrouter/xai in model router

TylerBarnes:
Long term we want to have our own provider code for this, but to make sure model router works perfectly right now, I installed the ai sdk provider packages for each of these and dynamically route to the underlying provider. I set this up for models.dev sources as well as Netlify

This makes multimodal inputs work properly, as well as the responses api (to support openai reasoning/gpt-5 models), and provider options (like anthropic cache breakpoints for ex).

These packages are quite small and have no additional external deps so it seems ok to include them in the server bundle:

https://bundlephobia.com/package/@ai-sdk/xai@2.0.23
https://bundlephobia.com/package/@openrouter/ai-sdk-provider@1.2.0
https://bundlephobia.com/package/@ai-sdk/google@2.0.17
https://bundlephobia.com/package/@ai-sdk/openai-compatible@1.0.19
https://bundlephobia.com/package/@ai-sdk/openai@2.0.42

Image input:
Screenshot 2025-10-06 at 5 24 44 PM
Responses-only model:
Screenshot 2025-10-06 at 5 28 53 PM


packages/core/src/llm/model/gateway-resolver.ts
```diff
@@ -1,71 +1,35 @@
...
   }
 
-  // remove the provider id
-  const firstSlashIndex = resolvedModelId.indexOf('/');
-  if (firstSlashIndex !== -1) {
-    resolvedModelId = resolvedModelId.substring(firstSlashIndex + 1);
+  const providerId = idParts.at(gatewayPrefix ? 1 : 0);
+  const modelId = idParts.slice(gatewayPrefix ? 2 : 1).join(`/`);
+
+  console.info({ providerId, modelId, idParts });
```
DanielSLew:
```suggestion
```

packages/core/src/llm/model/gateways/models-dev.ts
```diff
@@ -83,6 +82,8 @@ export class ModelsDevGateway extends MastraModelGateway {
     const providerConfigs: Record<string, ProviderConfig> = {};
 
     for (const [providerId, providerInfo] of Object.entries(data)) {
+      // Skip this provider, it doesn't work
+      if (providerId === `github-copilot`) continue;
```
DanielSLew: we should remove this from the provider generation if it doesn't work, we're still creating docs for it

DanielSLew: It might be worth it pulling this out into a separate array for `UNSUPPORTED_PROVIDERS = ['github-copilot'` to make it more clear in the code what we don't support, and also to add new unsupported providers if we run into any

packages/core/src/llm/model/gateways/constants.ts
```diff
@@ -0,0 +1 @@
+export const SUPPORTED_PROVIDERS = ['anthropic', 'google', 'openai', 'openrouter', 'xai'];
```
DanielSLew: Does this mean that these are the only supported providers?

DanielSLew: If that's the case we should remove docs for all of the other providers

packages/core/src/llm/model/gateways/models-dev.ts
```diff
@@ -95,23 +96,25 @@ export class ModelsDevGateway extends MastraModelGateway {
         providerInfo.npm === '@ai-sdk/gateway' || // Vercel AI Gateway is OpenAI-compatible
         normalizedId in OPENAI_COMPATIBLE_OVERRIDES;
 
+      const isOtherProvider = SUPPORTED_PROVIDERS.includes(providerId);
```
DanielSLew: hmm it's not clear to me what `isOtherProvider` mean, what is the "other" in this case?

packages/core/src/llm/model/gateways/netlify.ts
```diff
@@ -91,19 +83,19 @@ export class NetlifyGateway extends MastraModelGateway {
...
         category: 'UNKNOWN',
-        text: `Missing NETLIFY_SITE_ID environment variable required for model: ${modelId}`,
+        text: `Missing NETLIFY_SITE_ID environment variable required for model: ${routerId}`,
       });
     }
 
     try {
       const tokenData = await this.getOrFetchToken(siteId, netlifyToken);
-      return `${tokenData.url}chat/completions`;
+      return tokenData.url.endsWith(`/`) ? tokenData.url.substring(0, tokenData.url.length - 1) : tokenData.url;
```
DanielSLew, line 92: why'd we get rid of the chat/completions part?

packages/core/src/llm/model/gateways/models-dev.ts
TylerBarnes: docs are generated from the providers file so the doc doesn't get generated, but that's a good idea about making a constant

packages/core/src/llm/model/gateways/constants.ts
TylerBarnes: It means they're directly supported instead of using openai compat, I'll add a comment and rename to `DIRECTLY_SUPPORTED_PROVIDERS`

packages/core/src/llm/model/gateways/models-dev.ts
TylerBarnes: I'll rename, it's related to the comment I left on your comment above

packages/core/src/llm/model/gateways/netlify.ts
TylerBarnes: we're using openai/google/anthropic AI SDK providers directly and only falling back to openai compat

packages/core/src/llm/model/gateways/models-dev.ts
TylerBarnes: https://github.com/mastra-ai/mastra/pull/8559/commits/e1ff8fc0e8840dc745e145e080c9f2e48a6c2a89

packages/core/src/llm/model/gateways/constants.ts
TylerBarnes: https://github.com/mastra-ai/mastra/pull/8559/commits/e1ff8fc0e8840dc745e145e080c9f2e48a6c2a89

DanielSLew (approved):

PR #8570, opened 2025-10-07
add custom changeset CLI with peer dependency management

wardpeet:
Adds a custom changeset CLI that makes version management easier. Instead of manually editing changeset files, you can now interactively select version bumps and it automatically handles peer dependency updates.

Previously you'd need to know the changeset format and manually track peer deps. Now just run the CLI and pick your versions from a friendly prompt.


packages/_changeset-cli/src/changeset/getChangesetMessage.ts
```diff
@@ -0,0 +1,67 @@
...
+  }
+
+  try {
+    const message = await openEditor(template);
+
+    if (!message) {
+      return onCancel('⚠️  No changeset message provided. Aborting...');
+    }
+
+    return message;
```
LekoArts, line 66: This doesn't work for me, I never get past this step


https://github.com/user-attachments/assets/78d2e381-fae4-4d89-bff4-2e1646d9b85c

wardpeet: It opens vim for me . I'm using a package for it, you could try -m 'some message' to see it in action

LekoArts: Got it resolved, I was missing a `--wait` for my `EDITOR=code --wait`

wardpeet:
@greptilebot can you check again

PR #8582, opened 2025-10-07
chore: keep cloud deployer peer dependencies in sync

wardpeet:
Updates the cloud deployer peer dependencies to stay in sync with the core package versions. This prevents version mismatches when deploying to cloud platforms.

We sync the dependencies right before we publish so we have to correct versions assigned.

### DIFF
--- a/deployers/cloud/package.json
+++ b/deployers/cloud/package.json
@@ -29,7 +29,7 @@
     "build:watch": "tsup --watch --silent --config tsup.config.ts",
     "test": "vitest run",
     "lint": "eslint .",
-    "prepare": "node scripts/sync-versions.mjs"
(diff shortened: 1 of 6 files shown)


deployers/cloud/package.json
```diff
@@ -27,7 +28,8 @@
     "build": "tsup --silent --config tsup.config.ts",
     "build:watch": "tsup --watch --silent --config tsup.config.ts",
     "test": "vitest run",
-    "lint": "eslint ."
+    "lint": "eslint .",
+    "prepare": "node scripts/sync-versions.mjs"
```
wardpeet:
```suggestion
    "prepack": "node scripts/sync-versions.mjs"
```

rphansen91 (approved):

PR #8589, opened 2025-10-07
fix(@mastra/core, @mastra/playground-ui): model router OpenAI compat urls

TylerBarnes:
In https://github.com/mastra-ai/mastra/pull/8559 I added `@ai-sdk/openai-compat` but forgot to remove the `/chat/completions` we were adding to each API url.

Also fixed an issue where playground ui provider picker was showing `openai.responses` instead of `OpenAI`


docs/src/content/en/models/providers/xai.mdx
```diff
@@ -294,7 +294,6 @@ for await (const chunk of stream) {
 const agent = new Agent({
   name: "custom-agent",
   model: {
-    url: "https://api.x.ai/v1/chat/completions",
```
NikAiyer, line 297: Did we want to remove this whole url string?

TylerBarnes: yeah, xai has a provider package installed now

NikAiyer (approved):

CalebBarnes (approved):

PR #8590, opened 2025-10-07
fix(@mastra/agent-builder): support model router

TylerBarnes:
Before this it looks like agent builder template UI would just fallback to openai/gpt-4.1 when a provider like openrouter was selected


CalebBarnes (approved):
🚀

roaminro (approved):

PR #8604, opened 2025-10-08
loop iterationCount documentation

taofeeq-deru:
## Description

Adds documentation for loop iteration count.

## Type of Change

- [x] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [ ] Test update

## Checklist

### DIFF
--- a/docs/src/content/en/docs/workflows/control-flow.mdx
+++ b/docs/src/content/en/docs/workflows/control-flow.mdx
@@ -125,6 +125,44 @@ export const testWorkflow = createWorkflow({...})
   .commit();
 ```
 
+### Useful patterns for loop conditions
+
+#### Condition based on `inputData`
+
+The `inputData` of the loop step is the same as the output of the previous iteration of that step. For that reason, most often what we need to decide if the loop is done or not will be passed as `inputData` to the condition function.
+
+```typescript
+.dountil(nestedWorkflowStep, async ({ inputData: { userResponse } }) => userResponse === "yes")
+```
+
+If the `nestedWorkflowStep` is a workflow that suspends and resumes with a `userResponse` and eventually returns it as the result of the workflow, the condition will have access to it through `inputData`.
+
+#### Maximum iterations with a successful result
+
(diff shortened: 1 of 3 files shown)


docs/src/content/en/docs/workflows/control-flow.mdx
```diff
@@ -125,6 +125,44 @@ export const testWorkflow = createWorkflow({...})
...
+
+```typescript
+.dountil(nestedWorkflowStep, async ({ inputData: { userResponse } }) => userResponse === "yes")
+```
+
+If the `nestedWorkflowStep` is a workflow that suspends and resumes with a `userResponse` and eventually returns it as the result of the workflow, the condition will have access to it through `inputData`.
+
+#### Maximum iterations with a successful result
+
+The `iterationCount` is the number of times the loop has been executed. This is useful when we need to decide if the loop is done or not based on the number of iterations.
```
PaulieScanlon: Consider refining explaination.

The `iterationCount` is the number of times the loop has executed. This is useful when determining whether the loop should end based on the number of iterations.

docs/src/content/en/docs/workflows/control-flow.mdx
```diff
@@ -125,6 +125,44 @@ export const testWorkflow = createWorkflow({...})
...
+```typescript
+.dountil(nestedWorkflowStep, async ({ inputData: { userResponse, iterationCount } }) => userResponse === "yes" || iterationCount >= 10)
+```
+
+Here, even if the user does not respond with "yes", after 10 iterations the loop will be done and the workflow will continue with the last iteration's result.
+
+
+#### Abort workflow execution when maximum iterations is reached
+
+The `iterationCount` is the number of times the loop has been executed. This is useful when we need to decide if the loop is done or not based on the number of iterations.
```
PaulieScanlon: Perhaps to avoid repeating information, this section could be bumped up above the two code examples. If they share the same or similar explanation group them under a heading, then add a specific heading to introduce what's being seen in the code example.

PaulieScanlon (approved):

PR #8626, opened 2025-10-08
feat: standardize model configuration across scorers and processors

TylerBarnes:
This PR standardizes model configuration across all components to use the same flexible `MastraModelConfig` type that the `Agent` class uses.

Now you can use model router magic strings, custom configs, or dynamic functions anywhere you configure a model:

```typescript
// Before - only accepted LanguageModel instances
const scorer = createScorer({
  judge: {
    model: createOpenAI({ apiKey: '...' })('gpt-4o-mini'),
  },
});

// After - accepts magic strings, configs, or dynamic functions
const scorer = createScorer({
  judge: {
    model: 'openai/gpt-4o-mini',
    // or { providerId: 'openai', modelId: 'gpt-4o-mini' }
    // or { id: 'custom/model', url: 'https://...', apiKey: '...' }
    // or (ctx) => 'openai/gpt-4o-mini'
  },
});
```

This applies to all scorers (including built-in ones like `createAnswerRelevancyScorer`), input/output processors (`ModerationProcessor`, `PIIDetector`), and relevance scorers.


DanielSLew (approved):

TylerBarnes:
Reese says: @greptileai Thanks for catching the duplicate gateway resolution issue in `router.ts`! You were absolutely right - the gateway was being resolved twice with the same ID (`normalizedConfig.id` and `parsedConfig.routerId`), which was redundant and could cause inconsistency.

I've fixed this in commit 3fb35fdf7f by removing the second gateway resolution call. The gateway is now only resolved once using `normalizedConfig.id`, and that instance is reused throughout the constructor.

Regarding the peer dependency version mismatch you mentioned - the user is investigating that separately, so I've left that for them to handle.

Reese

PR #8631, opened 2025-10-08
feat(create-mastra): use model-router instead of provider packages

DanielSLew:
## Description

Use model router as default in create-mastra create command, remove ai sdk provider packages.


TylerBarnes (approved):
nice to see mostly deletions!! 🔥 😄

wardpeet (approved):

PR #8680, opened 2025-10-09
fix(playground): model picker autotab fix

DanielSLew:
## Description

Fix auto tab for model picker in playground-ui, the UI no longer auto tabs to the next selector when selecting a model/provider.


packages/playground-ui/src/domains/agents/components/agent-metadata/agent-metadata-model-switcher.tsx
```diff
@@ -675,20 +648,7 @@ export const AgentMetadataModelSwitcher = ({
...
-                            chatInput.focus();
-                          } else {
-                            // Fallback to any textarea if data-chat-input not found
-                            const anyTextarea = document.querySelector('textarea') as HTMLTextAreaElement;
-                            if (anyTextarea) {
-                              anyTextarea.focus();
-                            }
-                          }
-                        }, 0);
+                        // Don't auto-focus chat input on mouse click - only on Enter key
```
TylerBarnes: I think on click it still makes sense to highlight the chat, mainly we just want to avoid a keyboard trap where you can't properly tab through to all elements

packages/playground-ui/src/domains/agents/components/agent-metadata/agent-metadata-model-switcher.tsx
```diff
@@ -208,13 +208,7 @@ export const AgentMetadataModelSwitcher = ({
...
     }
 
-    // Only auto-focus model input if provider is connected
-    if (provider.connected) {
-      setTimeout(() => {
-        modelInputRef.current?.focus();
-        modelInputRef.current?.click();
-      }, 100);
-    }
+    // Don't auto-focus - let user navigate with Tab or Enter as they prefer
```
TylerBarnes: we should keep this one, I think jumping from provider to model picker on select is ok, you don't get trapped and you always need to select a model after you select the provider

TylerBarnes (changes requested):

packages/playground-ui/src/domains/agents/components/agent-metadata/agent-metadata-model-switcher.tsx
DanielSLew: added back in

packages/playground-ui/src/domains/agents/components/agent-metadata/agent-metadata-model-switcher.tsx
DanielSLew: added back in

TylerBarnes (approved):

PR #8687, opened 2025-10-09
chore(core): rename ai-v5 packages to be default

DanielSLew:
## Description

- Renamed ai sdk related packages that used to have the `-v5` to remove it so they become the defaults.
- Renamed ai sdk related packages that used to be the default to have the `-v4` suffix.
- Moved ai-sdk provider packages to dev deps


voice/azure/src/index.ts
```diff
@@ -219,7 +219,7 @@ export class AzureVoice extends MastraVoice {
       const chunkSize = 4096;
       for (let i = 0; i < audioData.length; i += chunkSize) {
         const chunk = audioData.slice(i, i + chunkSize);
-        pushStream.write(chunk);
+        pushStream.write(chunk.buffer);
```
DanielSLew, line 222: I'm not sure why this change was needed, but packages wouldn't build anymore because this was the wrong type. Maybe we just haven't had to build this in a while or some transient dependency forced a package to update? A little confused, because this package doesn't use ai-sdk packages either.

packages/core/package.json
```diff
@@ -224,25 +224,28 @@
...
   "peerDependencies": {
     "zod": "^3.25.0 || ^4.0.0"
   },
   "devDependencies": {
-    "@ai-sdk/openai": "^1.3.24",
+    "@ai-sdk/anthropic": "^2.0.23",
+    "@ai-sdk/google": "^2.0.17",
+    "@ai-sdk/openai": "^2.0.42",
+    "@ai-sdk/openai-compatible": "^1.0.19",
+    "@ai-sdk/xai": "^2.0.23",
```
CalebBarnes, line 237: @DanielSLew @TylerBarnes Don't we need these in deps for model router to work?

DanielSLew: Our bundlerooski says that dev deps gets bundled in core

TylerBarnes (approved):

packages/core/package.json
```diff
@@ -224,25 +224,28 @@
...
+    "zod-to-json-schema": "^3.24.6"
   },
   "peerDependencies": {
     "zod": "^3.25.0 || ^4.0.0"
   },
   "devDependencies": {
-    "@ai-sdk/openai": "^1.3.24",
+    "@ai-sdk/anthropic": "^2.0.23",
+    "@ai-sdk/google": "^2.0.17",
+    "@ai-sdk/openai": "2.0.42",
```
wardpeet, line 235: they all have a ^ except this one, is that intentional?

.changeset/wet-hoops-float.md
```diff
@@ -0,0 +1,6 @@
+---
+'@mastra/core': patch
+'@mastra/voice-azure': patch
```
wardpeet: Can you create a different changeset for voice alone?

One that just says, fixed typescript type or something

packages/core/package.json
DanielSLew: Yes, they shipped a bug in their latest version 2.0.46 where provider tools don't work anymore, so I just pinned to the version we previously had 2.0.42 that works

.changeset/wet-hoops-float.md
DanielSLew: done!

PR #8688, opened 2025-10-09
feat(core): dynamic model router with auto-refresh and runtime type generation

TylerBarnes:
Implements dynamic provider registry loading with automatic hourly refresh during development.

**What changed:**

The model router now fetches provider data from gateways at runtime instead of using a static generated file. When running `mastra dev`, it automatically refreshes every hour to pick up new models and providers.

**Before:**
- Static `provider-registry.generated.ts` file
- Manual regeneration required for new models
- No runtime updates

**After:**
- Dynamic `provider-registry.json` loaded at runtime
- Auto-refresh every hour in dev mode (configurable via `MASTRA_AUTO_REFRESH_PROVIDERS`)
- Runtime `.d.ts` regeneration for TypeScript autocomplete
- Persistent cache at `~/.cache/mastra/` to avoid unnecessary refreshes
- Dynamic refreshing only happens in `mastra dev`

**Key features:**
- `ModelRegistry.syncGateways()` for manual refresh
- Auto-refresh enabled by default when `MASTRA_DEV=true`
- Graceful error handling (uses stale data if refresh fails)
- Custom OpenAI-compatible endpoint support
- Comprehensive test coverage (12 new tests + integration tests)

Types are automatically updated when the registry refreshes, so your IDE autocomplete stays current with the latest models.


packages/core/scripts/generate-model-docs.ts
```diff
@@ -2,15 +2,12 @@ import fs from 'fs/promises';
...
-import { PROVIDER_REGISTRY } from '../src/llm/model/provider-registry.generated.js';
 
 /**
  * Generate a comment indicating the file was auto-generated
  */
 function getGeneratedComment(): string {
-  const timestamp = new Date().toISOString();
-  return `{/* This file is auto-generated by generate-model-docs.ts - DO NOT EDIT MANUALLY */}
-{/* Generated at: ${timestamp} */}`;
+  return `{/* This file is auto-generated by generate-model-docs.ts - DO NOT EDIT MANUALLY */}`;
```
TylerBarnes, line 11: I removed the timestamp so it doesn't make such huge diffs each time

packages/core/scripts/generate-providers.ts
```diff
@@ -1,132 +1,31 @@
...
- */
-
-import type { ProviderConfig } from './gateways/base';
-
-/**
- * Provider configurations for OpenAI-compatible APIs
- */
-export const PROVIDER_REGISTRY: Record<string, ProviderConfig> = ${JSON.stringify(allProviders, null, 2)} as const;
-
-/**
```
TylerBarnes, line 43: lots removed here because we now delegate to writeRegistryFiles()

packages/core/src/llm/model/gateways/models-dev.ts
```diff
@@ -130,16 +127,12 @@ export class ModelsDevGateway extends MastraModelGateway {
...
-        console.info(`Skipped provider ${providerInfo.name}`);
       }
     }
 
     // Store for later use in buildUrl and buildHeaders
     this.providerConfigs = providerConfigs;
 
-    console.info(`Found ${Object.keys(providerConfigs).length} OpenAI-compatible providers`);
-    console.info('Providers:', Object.keys(providerConfigs).sort());
     return providerConfigs;
```
TylerBarnes, line 136: removed these so it's not noisy when it regens at runtime

packages/core/src/llm/model/provider-registry.generated.ts
```diff
@@ -1,1706 +0,0 @@
-/**
```
TylerBarnes, line 1: instead of this file we now have a .d.ts + a .json file.
That's to make it easier to regen at runtime without rebundling

packages/core/src/llm/model/provider-registry.ts
```diff
@@ -0,0 +1,410 @@
...
+      const gateways = [new ModelsDevGateway({}), new NetlifyGateway()];
+
+      // Fetch provider data
+      const { providers, models } = await fetchProvidersFromGateways(gateways);
+
+      // Write to dist/ (the bundled location that gets distributed)
+      // Use createRequire to find the @mastra/core package root reliably
+      const require = createRequire(import.meta.url);
+      const packageJsonPath = require.resolve('@mastra/core/package.json');
+      const packageRoot = path.dirname(packageJsonPath);
```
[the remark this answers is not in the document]
TylerBarnes: @greptileai I don't think that's true though, it works fine when I'm developing mastra

TylerBarnes:
Reese says: @greptileai Thanks for the review! I've addressed the redundant `isDev` check you flagged on line 407. The `autoRefreshEnabled` variable already includes the `isDev` condition, so the additional `&& isDev` was unnecessary.

Fixed in commit 4513edc272.

Regarding the `require.resolve('@mastra/core/package.json')` concern - I appreciate you following up and confirming that approach is correct. The `createRequire(import.meta.url)` pattern works well for our use case.

All tests pass, lint is clean, and the build succeeds.

Reese

packages/core/scripts/generate-model-docs.ts
DanielSLew: amazing haha, I almost asked for that before

packages/core/scripts/generate-model-docs.ts
```diff
@@ -189,11 +192,11 @@ async function fetchProviderInfo(providerId: string): Promise<{ models: any[]; p
   }
 }
 
-async function generateProviderPage(provider: ProviderInfo): Promise<string> {
+async function generateProviderPage(provider: ProviderInfo, providerRegistry: Record<string, any>): Promise<string> {
```
DanielSLew: Don't we know the type of the providerRegistry?

DanielSLew (approved):

packages/core/scripts/generate-model-docs.ts
TylerBarnes: ah yeah we do, but we only use it for one line, don't want to run CI again for that 😆 I'll fix in a different PR

TylerBarnes: nvm CI failed anyway

PR #8714, opened 2025-10-10
Preserve Mastra span id

mjh-sakh:
## Description

Now Mastra and Braintrust spans cannot be matched because all span ids are being re-generated before being exported to Braintrust. This makes it impossible to add new leaf spans to already finished spans. 

This change preserves Mastra's OTEL style span id when corresponding Braintrust spans are created. It enables Mastra users to record tracingContext span ids and emit new spans that will be attached to them. 

### Example

User wants to capture customer feedback. This becomes possible by recording root span id together with thread id in Mastra's memory. This can be retrieved and new span emitted attached to original conversation:

```ts
braintrustLogger
  .startSpan({
    parentSpanIds: {
      spanId: spanId, // from memory
      rootSpanId: spanId,
    },
    name: 'attached-span',
    event: {
      metadata,
      scores: {
        feedback: score,
      },
    },
  })
  .end()
```

image

## Type of Change

- [ x ] New feature (non-breaking change that adds functionality)
- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [ ] Test update

## Checklist

### DIFF
--- a/.changeset/fast-cups-turn.md
+++ b/.changeset/fast-cups-turn.md
@@ -1,0 +1,5 @@
+---
+'@mastra/braintrust': patch
+---
+
+preserve Mastra span id when exported to Braintrust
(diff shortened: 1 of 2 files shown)


epinzur (approved):
looks great, thanks for the contribution.

Note: I'm surprised we aren't passing our TraceId to braintrust anywhere either.  I looked around in braintrust docs and couldn't find a good way to do it.

rltvty (approved):

PR #8766, opened 2025-10-13
tests: improve e2e tests

mfrachet:
## Description

What has been tested:

`/agents`
- link computation is accurate in the playground (since cloud and playground don't resolve the same links)
- verify the top level page (to avoid the 500 situation we were last week because of agent prompt being structured differently)
- verify that the information showing the entity count is visible to the screen

`/agents/:id`
- verify that thread list is showing when memory is enabled
- verify the "New chat" link computation because of differences between playground and cloud link computation
- verify the data resolution in the agent metadata page to make sure the data is valid from the DB to the UI

**Next steps**
- Mocking openai package at the agent config level
- Adding some fixtures in order to create valid tests
- Add a way to switch fixture depending on test case
- Tests generate / stream in a simple scenario
- Test generate / stream in a complex scenario (tool / workflow / network)

## Type of Change

- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [x] Test update

## Checklist

### DIFF
--- a/e2e-tests/kitchen-sink/tests/agents.$agentId.spec.ts
+++ b/e2e-tests/kitchen-sink/tests/agents.$agentId.spec.ts
@@ -21,7 +21,9 @@ test('overall layout information', async ({ page }) => {
   await expect(agentCrumb).toHaveAttribute('href', '/agents/weatherAgent');
 
   // Thread history (with memory)
-  await expect(page.locator('text=New Chat')).toBeVisible();
+  const newChatButton = await page.locator('text=New Chat');
+  await expect(newChatButton).toBeVisible();
+  await expect(newChatButton).toHaveAttribute('href', /agents\/weatherAgent\/chat\/new.*/);
   await expect(page.locator('text=Your conversations will appear here once you start chatting!')).toBeVisible();
 
   // Information side panel
--- a/e2e-tests/kitchen-sink/tests/agents.$agentId.spec.ts
+++ b/e2e-tests/kitchen-sink/tests/agents.$agentId.spec.ts
@@ -55,7 +55,7 @@ test.describe('agent panels', () => {
       const toolBadge = await page.locator('[data-testid="tool-badge"]');
       await expect(toolBadge).toBeVisible();
       await expect(toolBadge).toHaveText('get-weather');
-      expect(toolBadge).toHaveAttribute('href', '/agents/weatherAgent/tools/get-weather');
+      await expect(toolBadge).toHaveAttribute('href', '/agents/weatherAgent/tools/get-weather');
 
       await expect(page.locator('text=No workflows')).toBeVisible();
       await expect(page.locator('text=No scorers')).toBeVisible();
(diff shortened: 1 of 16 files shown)


e2e-tests/kitchen-sink/tests/agents.$agentId.spec.ts
```diff
@@ -1,55 +1,136 @@
...
-  await page.click('text=Advanced Settings');
-
-  // Assert
-  await expect(page.getByLabel('Top K')).toHaveValue('9');
-  await expect(page.getByLabel('Frequency Penalty')).toHaveValue('0.7');
-  await expect(page.getByLabel('Presence Penalty')).toHaveValue('0.6');
-  await expect(page.getByLabel('Max Tokens')).toHaveValue('44');
-  await expect(page.getByLabel('Max Steps')).toHaveValue('3');
-  await expect(page.getByLabel('Max Retries')).toHaveValue('2');
+test('overall layout information', async ({ page }) => {
```
wardpeet, line 3: these tests are to specifc, will be painful on redesigns

wardpeet: try to test functionality with these tests. Like does the sending a message work and can i add a second message, is the agent outcome present.

You could than also check if the correct api request got send instead of actually checking if all UI elements have the correct attributes.

maybe toMatchAriaSnapshot is what we can use as well

mfrachet: My intention in these specific tests ("layout XXX") is to verify that fetching the data actually resolves someting on the screen in an e2e way (and reaching the db) => I wanted to know if something breaks far in the backend that the UI does not respond accordingly.

This was to avoid the situation we got last week where /agents was throwing in a 500 and we didn't get any info.

Do you think it's a problem to make the test at this level? What is the drawback?

mfrachet: (also adding messages is next step on my todo 😄, I wanted to lock default agents behaviour and then play with the mocking lib we talked about last time)

wardpeet: Main issue is that these selectors are very explicit so these tests will break easily.

if we can move them to snapshots its easier to update
```
  const breadcrumb = page.locator('header>nav');
  const agentsCrumb = breadcrumb.locator('li:nth-child(1)>a');
  const separatorCrumb = breadcrumb.locator('li:nth-child(2)');
  const agentCrumb = breadcrumb.locator('li:nth-child(3)>a');
```

mfrachet: sure, let me try that 👍🏻

mfrachet: Just pushed the aria snapshot :)

abhiaiyer91 (approved):

PR #8778, opened 2025-10-13
fix(deployer): Remove validation step to fix transpilation with binary deps

wardpeet:
The validation step in the bundler was causing transpilation to fail for TypeScript files that depend on packages with binaries. This removes that validation logic to allow proper bundling.

Before, the bundler would try to validate each chunk by importing it, which failed for files with binary dependencies.

After, we skip validation and let the bundler proceed, which allows TS files with binary deps to transpile correctly.

image
image


LekoArts:
@greptile check it again

PR #8834, opened 2025-10-14
feat: add scheduled GitHub Action to regenerate provider registry

TylerBarnes:
Adds a scheduled workflow that automatically regenerates the provider registry every 6 hours, keeping the model/provider data fresh without manual intervention.

The workflow fetches the latest data from models.dev and Netlify, generates the registry files, formats them with Prettier, and commits directly to main when changes are detected. It also creates a changeset for @mastra/core to trigger a patch version bump.

This works alongside the existing runtime auto-refresh system - the GitHub Action keeps the repo source files up-to-date while the runtime system keeps local dev environments fresh between git pulls.

Can be manually triggered from the Actions tab for testing.

### DIFF
--- a/.github/workflows/regenerate-provider-registry.yml
+++ b/.github/workflows/regenerate-provider-registry.yml
@@ -62,7 +62,7 @@ jobs:
           # Generate a unique changeset filename with timestamp
           CHANGESET_ID="automated-provider-registry-$(date +%Y%m%d-%H%M%S)"
           CHANGESET_FILE=".changeset/${CHANGESET_ID}.md"
-          
+
           # Create changeset for @mastra/core with patch bump
           cat > "${CHANGESET_FILE}" << 'EOF'
           ---
@@ -71,7 +71,7 @@ jobs:
 
           Update provider registry with latest models and providers
           EOF
-          
+
           echo "Created changeset: ${CHANGESET_FILE}"
           echo "CHANGESET_FILE=${CHANGESET_FILE}" >> $GITHUB_ENV
 
--- a/.github/workflows/regenerate-provider-registry.yml
+++ b/.github/workflows/regenerate-provider-registry.yml
@@ -17,11 +17,35 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
+      - name: Get Github App Token
+        uses: actions/create-github-app-token@v2
+        id: app-token
+        with:
+          app-id: ${{ vars.DANE_APP_ID }}
+          private-key: ${{ secrets.DANE_APP_PRIVATE_KEY }}
+
+      - name: Get GitHub App User ID
+        id: get-user-id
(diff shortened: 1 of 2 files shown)


TylerBarnes:
Reese says: Thanks for the review @greptileai! Agree with #1 and #2 - will fix the changeset filename uniqueness and the git add wildcard issue.

For #3 (optimizing pnpm install), I'm going to skip this one. Looking at other workflows in the repo (lint.yml, version-packages.yml, etc.), they all do full `pnpm install` without filters. This is because workspace dependencies need to be resolved, and pnpm caching makes subsequent runs fast anyway. The generation script also has transitive dependencies that need the full workspace context.

Implementing the other two fixes now!

Reese

abhiaiyer91 (approved):

.github/workflows/regenerate-provider-registry.yml
```diff
@@ -0,0 +1,95 @@
...
+  # Allow manual triggering for testing
+  workflow_dispatch:
+
+jobs:
+  regenerate:
+    # Only run on the main repository, not on forks
+    if: ${{ github.repository == 'mastra-ai/mastra' }}
+    runs-on: ubuntu-latest
+
+    steps:
```
wardpeet, line 19: Can you use the following template or look at other workflows. We should use Dane as the PR creator so our actions get triggered. GH action bot is a special user in github which doesn't trigger all subactions correctly.

```
- uses: actions/create-github-app-token@v1
        id: app-token
        with:
          app-id: ${{ vars.DANE_APP_ID }}
          private-key: ${{ secrets.DANE_APP_PRIVATE_KEY }}

      - name: Get GitHub App User ID
        id: get-user-id
        run: echo "user-id=$(gh api "/users/${{ steps.app-token.outputs.app-slug }}[bot]" --jq .id)" >> "$GITHUB_OUTPUT"
        env:
          GH_TOKEN: ${{ steps.app-token.outputs.token }}

      - run: |
          cat <<- EOF > $HOME/.netrc
            machine github.com
            login ${{ steps.app-token.outputs.app-slug }}
            password ${{ steps.app-token.outputs.token }}
            machine api.github.com
            login ${{ steps.app-token.outputs.app-slug }}
            password ${{ steps.app-token.outputs.token }}
          EOF
          git config --global user.name '${{ steps.app-token.outputs.app-slug }}[bot]'
          git config --global user.email '${{ steps.get-user-id.outputs.user-id }}+${{ steps.app-token.outputs.app-slug }}[bot]@users.noreply.github.com'
```

wardpeet (approved):
LGTM, only need to change the user to use the dane bot

.github/workflows/regenerate-provider-registry.yml
TylerBarnes: Thanks, updated it in https://github.com/mastra-ai/mastra/pull/8834/commits/565de8b3fc57a7e33ef53ea263a2cd30fe9878f0

PR #8835, opened 2025-10-14
fix(core): allow input processors to add non-user role messages

### LINKED ISSUE #8656: [DISCORD:1425455372097032192] Input processor can't add non user role message
This issue was created from Discord post: https://discord.com/channels/1309558646228779139/1425455372097032192

I have tried to add system role message in my input processor and it led to the error:

```
Error executing step prepare-memory-step: Error: [Agent:gitlab-agent] - Input processor error
    at Agent.__runInputProcessors (file:///Users/[REDACTED]/node_modules/@mastra/core/dist/chunk-DQISKQDE.js:672:2179)
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
    at async Object.execute (file:///[REDACTED]/node_modules/@mastra/core/dist/chunk-DQISKQDE.js:641:271)
    at async DefaultExecutionEngine.executeStep (file:///[REDACTED]/node_modules/@mastra/core/dist/chunk-DQISKQDE.js:230:3675)
    at async DefaultExecutionEngine.executeEntry (file:///[REDACTED]/node_modules/@mastra/core/dist/chunk-DQISKQDE.js:238:359)
    at async Promise.all (index 1)
    at async DefaultExecutionEngine.executeParallel (file:///[REDACTED]/node_modules/@mastra/core/dist/chunk-DQISKQDE.js:234:3132)
    at async DefaultExecutionEngine.executeEntry (file:///[REDACTED]/node_modules/@mastra/core/dist/chunk-DQISKQDE.js:238:2757)
    at async DefaultExecutionEngine.execute (file:///[REDACTED]/node_modules/@mastra/core/dist/chunk-DQISKQDE.js:228:904)
    at async Run._start (file:///[REDACTED]/node_modules/@mastra/core/dist/chunk-DQISKQDE.js:310:1894)
ERROR [2025-10-08 14:47:37.177 +0300] (Mastra): Error in stream generate: Unknown error
```

There is no statements in documentation that it is disallowed

DanielSLew:
This fixes an issue where input processors couldn't add system or assistant messages. Previously all messages from input processors were forced to be user messages, causing an error when trying to add other role types.

The fix checks the role of each message from the processor and handles system messages specially by converting them to CoreMessage format, while other messages (user/assistant) are added normally.

Before:
```typescript
// This would throw an error
processInput: async ({ messages }) => {
  const systemMessage = { role: 'system', content: { ... } };
  return [systemMessage, ...messages];
}
```

After:
```typescript
// Now works correctly
processInput: async ({ messages }) => {
  const systemMessage = { role: 'system', content: { ... } };
  const assistantMessage = { role: 'assistant', content: { ... } };
  return [systemMessage, assistantMessage, ...messages]; // ✅ System and assistant messages handled properly
}
```

Fixes #8656


packages/core/src/processors/runner.ts
```diff
@@ -383,7 +384,17 @@ export class ProcessorRunner {
     }
 
     if (processableMessages.length > 0) {
-      messageList.add(processableMessages, 'user');
+      // Process messages based on their actual roles
+      for (const message of processableMessages) {
+        if (message.role === 'system') {
+          const coreMessage = convertMessages([message]).to('AIV4.Core')[0];
+          messageList.addSystem(coreMessage!);
```
TylerBarnes: we shouldn't have any message role checking/conversion logic like this outside of message list. we should just add all the messages at once to the list

packages/core/src/processors/runner.ts
```diff
@@ -383,7 +384,17 @@ export class ProcessorRunner {
...
+        } else if (message.role === 'user') {
+          messageList.add(message, 'user');
+        } else if (message.role === 'assistant') {
+          messageList.add(message, 'response');
```
TylerBarnes: this isn't what the second arg means. Second arg is the message source, here they're all `input`

TylerBarnes (changes requested):

packages/core/src/agent/message-list/index.ts
```diff
@@ -910,7 +920,11 @@ export class MessageList {
       // In the past system messages were accidentally stored in the db. these should be ignored because memory is not supposed to store system messages.
       if (messageSource === `memory`) return null;
 
-      if (MessageList.isAIV4CoreMessage(message) || MessageList.isAIV5CoreMessage(message))
+      if (
+        MessageList.isAIV4CoreMessage(message) ||
+        MessageList.isAIV5CoreMessage(message) ||
+        MessageList.isMastraMessageV2(message)
```
TylerBarnes: if we're converting to core messages, why are we checking for v2 here?

also we should do something cleaner instead of reassigning a variable, it was ok as a simple ternary before but now it's looking rough

```ts
if (isv2) return this.addSystem(MessageList.mastraMessageV2SystemToV4Core(message)) // <- new static method
if (isv4) return x
if (isv5) return y
```

TylerBarnes (changes requested):

packages/core/src/agent/message-list/index.ts
TylerBarnes: oh these are different fns, looked at the diff wrong, I mean up above

DanielSLew: Hmm it's a pretty small function and the reason we reassign a variable is because there is handling afterwards

TylerBarnes: it's messy, should organize it something like:

```
const message = MessageList.systemMessageToAICore(inputMessage) // takes any kind of system message format and returns a single one
```

pretty much any time you need to do `let something: type` without defining it, it's a code smell imo. Makes it harder to follow when you get 100 different snippets of code like that over time

DanielSLew: okay I made this change!

packages/core/src/agent/message-list/index.ts
```diff
@@ -873,6 +884,30 @@ export class MessageList {
...
+    let content = '';
+
+    if (typeof message.content.content === 'string') {
+      content = message.content.content;
+    } else if (message.content.parts) {
+      // Concatenate all text parts
+      content = message.content.parts
+        .filter(part => part.type === 'text')
+        .map(part => (part as any).text)
+        .join('');
```
TylerBarnes, line 858: if there are multiple text content parts we should retain that as-is. These wouldn't be streaming chunks, they could be paragraphs or sentences, we don't really know. For the LLM it's all the same but if someone uses the message elsewhere the rendering will be messed up. Even if we joined like this, joining on an empty string will join all the text parts together like `theseareallthetextpartsthatwereseparate`

TylerBarnes (changes requested):

TylerBarnes (approved):

Guria:
This issue is not fixed. I've added details to #8656

PR #8880, opened 2025-10-15
fix: use pnpm prettier in regenerate workflow

TylerBarnes:
The workflow was failing with `prettier: not found` when trying to format the generated files.

Changed from `prettier --write` to `pnpm prettier --write` to match the pattern used in the root package.json scripts (e.g., `prettier:format` uses `prettier --write` but is run via pnpm).

This ensures prettier is available from the workspace dependencies.


TylerBarnes:
closing in favour of https://github.com/mastra-ai/mastra/pull/8882

PR #8885, opened 2025-10-15
chore(cli): Weather agent scorers and consolidation

graysonhicks:
## Description

This unifies the weather agent between template, example and CLI string.  Template is now source of truth, example has been deleted, and CLI uses the template files.

Also adds three OOTB scorers and one custom scorer to the agent example.

Deletes other used files.

## Type of Change

- [ ] Documentation update
- [x] Code refactoring
- [ ] Performance improvement
- [ ] Test update

## Checklist


templates/weather-agent/src/mastra/agents/index.ts
LekoArts, line 1: This file has some commented out code and greptile is also complaining about something

templates/weather-agent/src/mastra/agents/index.ts
```diff
@@ -1,21 +1,60 @@
...
+- Keep responses concise but informative
 
-      Use the weatherTool to fetch current weather data.
-`,
-  model: openai('gpt-4o'),
+Use the weatherTool to fetch current weather data.`,
+  model: [
+    { model: anthropic('claude-3-5-sonnet-20241022') },
+    { model: openai('gpt-4o') },
+    { model: openai('gpt-4o-mini') },
```
LekoArts: We can use our new model router (https://mastra.ai/en/models) so you don't need to install the ai-sdk packages for this template 👍

templates/weather-agent/package.json
```diff
@@ -13,9 +13,12 @@
   "description": "One Agent, one Workflow and one Tool to bring you the weather in your city.",
   "type": "module",
   "dependencies": {
-    "@ai-sdk/openai": "^1.3.24",
+    "@ai-sdk/anthropic": "^2.0.9",
+    "@ai-sdk/openai": "^2.0.23",
```
LekoArts: See model router comment further below

packages/cli/src/commands/init/init.ts
```diff
@@ -70,6 +70,17 @@ export const init = async ({
...
+      // Scorers require evals dependencies; install when agent is added
+      if (components.includes('agents')) {
+        const needsEvals = (await depService.checkDependencies(['@mastra/evals'])) !== `ok`;
+        if (needsEvals) {
+          await depService.installPackages(['@mastra/evals']);
+        }
+        const needsAiSdk = (await depService.checkDependencies(['@ai-sdk/openai'])) !== `ok`;
+        if (needsAiSdk) {
+          await depService.installPackages(['@ai-sdk/openai']);
+        }
```
LekoArts: See model router comment further below

packages/cli/src/commands/init/utils.ts
```diff
@@ -56,249 +55,68 @@ export const getModelIdentifier = (llmProvider: LLMProvider) => {
...
-  inputSchema: z.object({
-    city: z.string().describe('The city to get the weather for'),
-  }),
-  outputSchema: forecastSchema,
-  execute: async ({ inputData }) => {
-    if (!inputData) {
-      throw new Error('Input data not found');
-    }
+// Resolve a template file from templates/weather-agent/src/mastra
+const readWeatherTemplateFile = async (subpath: string): Promise<string> => {
```
LekoArts, line 59: Can't we adjust the `copyStarterFile` method on the FileService class and use that? If the template uses the model router syntax then you also don't need to adjust the string in this file here

LekoArts (changes requested):

templates/weather-agent/src/mastra/agents/index.ts
graysonhicks: The commented out code was already there, I assumed it was used for quick demo purposes by commenting back in.  Will check and remove if not needed

packages/cli/src/commands/init/utils.ts
graysonhicks: We still have some transforms (will need to put their chosen model provider in even with router change) so don't think its the right fit

templates/weather-agent/src/mastra/agents/index.ts
graysonhicks: Sam asked to keep for demo purposes

graysonhicks:
closing for a simpler approach

PR #8924, opened 2025-10-16
fix(core): add `strictJsonSchema` to openai `providerOptions` when using json response format

CalebBarnes:
## Description

When using OpenAI models with JSON response format, the provider does not automatically enable strict schema validation. This causes the LLM to sometimes not fully respect the provided JSON schema, resulting in issues like missing required fields or incorrect data types.

This PR adds  to the OpenAI provider options when:
- The model provider is OpenAI
- Response format is JSON

This ensures consistent, schema-compliant JSON responses from OpenAI models.

## Type of Change

- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [ ] Test update

## Checklist

### DIFF
--- a/packages/core/src/stream/aisdk/v5/execute.ts
+++ b/packages/core/src/stream/aisdk/v5/execute.ts
@@ -110,6 +110,12 @@ export function execute<OUTPUT extends OutputSchema = undefined>({
     });
   }
 
+  /**
+   * Enable OpenAI's strict JSON schema mode to ensure schema compliance.
+   * Without this, OpenAI may omit required fields or violate type constraints.
+   * @see https://platform.openai.com/docs/guides/structured-outputs#structured-outputs-vs-json-mode
+   * @see https://ai-sdk.dev/docs/ai-sdk-core/generating-structured-data#accessing-reasoning
+   */
   const providerOptionsToUse =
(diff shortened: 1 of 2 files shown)


packages/core/src/stream/aisdk/v5/execute.ts
```diff
@@ -109,6 +109,24 @@ export function execute<OUTPUT extends OutputSchema = undefined>({
...
   }
+
+  /**
+   * Enable OpenAI's strict JSON schema mode to ensure schema compliance.
+   * Without this, OpenAI may omit required fields or violate type constraints.
+   * @see https://platform.openai.com/docs/guides/structured-outputs#structured-outputs-vs-json-mode
+   * @see https://ai-sdk.dev/docs/ai-sdk-core/generating-structured-data#accessing-reasoning
+   */
+  const providerOptionsToUse =
+    model.provider.startsWith('openai') && responseFormat?.type === 'json' && !structuredOutput?.jsonPromptInjection
```
CalebBarnes, line 120: Using `startsWith('openai')` because this could be just "openai" or something like "openai.responses", "openai.completions" etc.

DanielSLew (approved):

PR #8926, opened 2025-10-16
fix `model.loop.test.ts` tests to use `structuredOutput.schema` and add assertions

CalebBarnes:
## Description

Fixed  to properly test functionality instead of just logging output. The tests were previously using the deprecated `output` parameter which no longer exists internally in model.stream args, causing them to silently pass without actually testing anything.

### Changes:
- **Replaced `output` with `structuredOutput.schema`** - Updated all tests to use the current API
- **Added proper assertions** - Replaced `console.log` statements with `expect` assertions to actually validate:
  - Text generation returns defined strings
  - Object generation returns properly typed objects with expected fields
  - Streaming produces chunks
- **Added missing `agentId`** - Required parameter for proper test execution
- **Increased timeouts** - Changed from 10s to 20s for more reliable test execution
- **Updated type definitions** - Fixed `AIV5FullStreamPart` and `AIV5FullStreamType` to properly type structured output
- **Fixed return type** - `MastraLLMVNext.stream` now returns `MastraModelOutput<OUTPUT>` instead of `MastraModelOutput<OUTPUT | undefined>`

## Type of Change

- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [x] Test update

## Checklist


packages/core/src/llm/model/model.loop.test.ts
```diff
@@ -102,18 +110,21 @@ describe('MastraLLMVNext', () => {
       ),
       runtimeContext: new RuntimeContext(),
       tracingContext: {},
-      output: z.object({
-        name: z.string(),
-        age: z.number(),
-      }),
+      structuredOutput: {
```
DanielSLew, line 113: It looks like this test still isn't asserting anything

DanielSLew (approved):
Nice! Good catch, just one of the tests still isn't asserting anything

packages/core/src/llm/model/model.loop.test.ts
CalebBarnes: Oops missed one! Thanks 😁

PR #8984, opened 2025-10-17
update manual install guide

bookercodes:


docs/src/content/en/docs/getting-started/installation.mdx
```diff
@@ -15,7 +15,7 @@ For more control over setup, or to add Mastra to an existing project, see the [m
 
 ## Before you start
 
-- You'll need an API key from a [model provider](/models) to complete setup. We recommend starting with [Gemini](https://aistudio.google.com/app/api-keys), as you likely already have a Google account and they don't require a card.
+- You'll need an API key from a [model provider](/models) to complete setup. We suggest starting with [Gemini](https://aistudio.google.com/app/api-keys), as you likely already have a Google account and they don't require a card.
```
PaulieScanlon, line 18: I don't know if we need to make assumptions like this. Perhaps a better follow up is:

We suggest starting with Gemini, as it's free to sign up and quick to get going (or something similar)

PaulieScanlon (approved):

PaulieScanlon:
All good old chum. Some things to consider though. I think with Kapa (the new search tool) it favours headings like: Creating, and Adding, rather than Create and Add

docs/src/content/en/docs/getting-started/installation.mdx
bookercodes: Can you elaborate? I'm not sure what you mean by assumption

bookercodes:
@PaulieScanlon Can you elaborate on that one and how to balance it? Crate an agent is a step, Creating an agent doesn't sound quite right to me

docs/src/content/en/docs/getting-started/installation.mdx
bookercodes: Some context here https://kepler-bej6556.slack.com/archives/C0919NR870W/p1760367353505869?thread_ts=1760367302.907109&cid=C0919NR870W

I am planning to come back to it!  Want to make sure I understand your feedback so I can consider it properly

PR #8989, opened 2025-10-17
fix(@mastra/core): allow custom providers with custom URLs in model router

TylerBarnes:
Fixes an issue where the model router would fail with "Could not find config for provider" error even when a custom URL was specified.

Previously, if you tried to use a custom provider like Ollama with a custom URL, the router would still validate against the provider registry and throw an error:

```typescript
const agent = new Agent({
  model: {
    id: 'ollama/llama3.2',
    url: 'http://localhost:11434/v1',
  },
});
// ❌ Error: Could not find config for provider ollama with model id ollama/llama3.2
```

Now it works as expected.

The fix skips provider registry validation when a custom URL is provided, allowing any provider name to be used for display/identification purposes while routing to the custom endpoint.


packages/core/src/llm/model/router-custom-provider.test.ts
```diff
@@ -0,0 +1,203 @@
...
+      });
+
+      await agent.generate('test', { maxSteps: 1 });
+
+      // Verify createOpenAICompatible was called with the right params
+      expect(createOpenAICompatible).toHaveBeenCalledWith({
+        name: 'my-custom-provider',
+        apiKey: 'test-key',
+        baseURL: 'http://fake-test-server-that-does-not-exist.local:9999/v1',
+        headers: undefined,
```
CalebBarnes, line 52: This might also have supportsStructuredOutputs: true

TylerBarnes: does it matter for this test?

TylerBarnes: ah I see with changes from main it might fail

CalebBarnes (approved):

Issue #8920, opened 2025-10-16
[FEATURE] Show agent tool outputs better in playground


rase-:
### Problem Statement

Right now `agents` passed to an Agent are rendered as a json payload when called as tools with `generate` or `stream`.

### Proposed Solution

We should show the output as a streamed string, similarly to how we do with `Agent.network()`.

Easiest way to do this would be to prefix the `toolName` when calling an agent or a workflow with `workflow-` or `agent-` to identify the tool type in playground.

### Component

Agents

### Example Use Case

Playground

### Verification

- [x] I have searched the existing issues to make sure this is not a duplicate
- [x] I have provided sufficient context for the team to understand the request

rase-:
Related issue https://github.com/mastra-ai/mastra/issues/8216

rase-:
This is fixed in https://github.com/mastra-ai/mastra/pull/9021

Issue #8216, opened 2025-09-26
[DISCORD:1420814550290141334] Playground agent chat is making failing API calls looking for tools as workflows (core 0.16.3)


wardpeet:
This issue was created from Discord post: https://discord.com/channels/1309558646228779139/1420814550290141334

```
Error getting workflow, searching agents for workflow MastraError: Workflow with ID getConversationTool not found
    at Mastra.getWorkflow (file:///Users/alex/projects/my-project/backend/node_modules/@mastra/core/dist/chunk-46ZMVRXN.js:568:21)
    at file:///Users/alex/projects/my-project/backend/node_modules/@mastra/core/dist/chunk-TLJPVRO5.js:146:57
    at NoopContextManager.with (/Users/alex/projects/my-project/backend/node_modules/@opentelemetry/api/build/src/context/NoopContextManager.js:25:19)
    at ContextAPI.with (/Users/alex/projects/my-project/backend/node_modules/@opentelemetry/api/build/src/api/context.js:60:46)
    at descriptor.value (file:///Users/alex/projects/my-project/backend/node_modules/@mastra/core/dist/chunk-TLJPVRO5.js:146:30)
    at getWorkflowsFromSystem (file:///Users/alex/projects/my-project/backend/.mastra/output/index.mjs:4778:25)
    at getWorkflowByIdHandler$1 (file:///Users/alex/projects/my-project/backend/.mastra/output/index.mjs:4811:32)
    at getWorkflowByIdHandler (file:///Users/alex/projects/my-project/backend/.mastra/output/index.mjs:51392:28)
    at dispatch (file:///Users/alex/projects/my-project/backend/.mastra/output/index.mjs:1543:23)
    at file:///Users/alex/projects/my-project/backend/.mastra/output/index.mjs:1543:46 {
  id: 'MASTRA_GET_WORKFLOW_BY_ID_NOT_FOUND',
  domain: 'MASTRA',
  category: 'USER',
  details: {
    status: 404,
    workflowId: 'getConversationTool',
    workflows: 'fooWorkflow, barWorkflow, bazWorkflow'
  },
  [cause]: undefined
}
Workflow not found, searching agents for workflow { workflowId: 'getConversationTool' }

mfrachet:
[...]

PR #9021, opened 2025-10-20
Show agent tool output better in playground

taofeeq-deru:
## Description

This fixes the UI of agent tool output in playground, and also prevent get workflow api call that occurs and fails when the tool is not a workflow

https://github.com/user-attachments/assets/ab704f53-62d2-495d-a453-b5ca64797a00

## Related Issue(s)

#8216 
#8920 

## Type of Change

- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [ ] Test update

## Checklist


rase- (approved):

packages/playground-ui/src/hooks/use-agent-messages.ts
```diff
@@ -0,0 +1,16 @@
+import { useMastraClient } from '@mastra/react';
+import { useQuery } from '@tanstack/react-query';
+
+export const useAgentMessages = ({ threadId, agentId }: { threadId: string; agentId: string }) => {
+  const client = useMastraClient();
+
+  return useQuery({
```
mfrachet, line 15: we already have `useMessages` in packages/playground/src/hooks/use-memory.ts

packages/playground-ui/src/components/assistant-ui/tools/tool-fallback.tsx
```diff
@@ -25,15 +25,17 @@ const ToolFallbackInner = ({ toolName, result, args, metadata, ...props }: ToolF
...
+  const isWorkflow = (metadata?.mode === 'network' && metadata.from === 'WORKFLOW') || toolName.startsWith('workflow-');
 
-  const isAgent = metadata?.mode === 'network' && metadata.from === 'AGENT';
+  const agentToolName = toolName.startsWith('agent-') ? toolName.substring('agent-'.length) : toolName;
+  const workflowToolName = toolName.startsWith('workflow-') ? toolName.substring('workflow-'.length) : toolName;
 
-  if (isAgent) {
-    const messages = result?.childMessages ?? [];
+  useWorkflowStream(result);
+  const { data: workflow, isLoading } = useWorkflow(workflowToolName, isWorkflow);
```
mfrachet: instead of having useWorkflow here, we should probably modify the WorkflowBadge and handle the fetching request instead. It should be clearer.

The stream accumulation in useWorkflowStream should stay there though

packages/playground-ui/src/hooks/use-workflows.ts
```diff
@@ -12,13 +12,13 @@ export type ExtendedWorkflowWatchResult = WorkflowWatchResult & {
...
 
-export const useWorkflow = (workflowId?: string) => {
+export const useWorkflow = (workflowId?: string, enabled: boolean = true) => {
   const client = useMastraClient();
   const { runtimeContext } = usePlaygroundStore();
   return useQuery({
     queryKey: ['workflow', workflowId],
     queryFn: () => (workflowId ? client.getWorkflow(workflowId).details(runtimeContext) : null),
-    enabled: Boolean(workflowId),
+    enabled: Boolean(workflowId) && enabled,
```
mfrachet: if you make the modification about moving the call to useWorfklow, this modification is not needed

packages/playground-ui/src/hooks/use-agent-messages.ts
taofeeq-deru: this is used for agents that were used as tool, each agent tool created has it's own unique threadId which is returned in the toolResult. Same way we're calling useWorkflowRuns in playground-ui to get the run information for a workflow used as tool, this is something similar. For an agent used as tool, to get the messages for that agent, here in playground-ui

packages/playground-ui/src/components/assistant-ui/tools/tool-fallback.tsx
taofeeq-deru: thanks, moved it to WorkflowBadge

packages/playground-ui/src/hooks/use-workflows.ts
taofeeq-deru: removed this too

packages/playground-ui/src/hooks/use-agent-messages.ts
taofeeq-deru: removed the one in packages/playground/src/hooks/use-memory.ts

mfrachet (approved):

PR #9034, opened 2025-10-20
paul/grwth-884-docstoberfest-workflows-agents-tools

PaulieScanlon:
## Description

- Agents should really be introduced much earlier with workflows, including in the Overview page
- Typo “withing” in “Using workflows as tools” section
- Exposing workflows with MCPServer
- Can we expose agents and tools in MCP as well? This potentially should just be moved to an MCP section

## Type of Change

- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [ ] Test update

## Checklist


docs/src/content/en/docs/workflows/overview.mdx
```diff
@@ -81,11 +88,9 @@ export const testWorkflow = createWorkflow({
 
 ### Understanding control flow
 
-Workflows can be composed using a number of different methods. The method you choose determines how each step's schema should be structured.
-
-> See [Control Flow](./control-flow.mdx) for more information.
+Workflows can be composed using a number of different methods. The method you choose determines how each step's schema should be structured. Visit the [Control Flow](./control-flow.mdx) page for more information.
```
bookercodes, line 91: We should be looking for more opportunities to replace callouts with inline links like this, nice one!

docs/src/content/en/docs/workflows/overview.mdx
```diff
@@ -200,23 +205,21 @@ The workflow output includes the full execution lifecycle, showing the input and
 {
   "status": "success",
   "steps": {
-    "input": {
-      "message": "Hello world"
-    },
+    // ...
     "step-1": {
+      "status": "success",
```
bookercodes, line 210: Oh, interesting, we had the output documented totally wrong all this time?

docs/src/content/en/docs/workflows/using-with-agents-and-tools.mdx
```diff
@@ -5,346 +5,119 @@ description: "Steps in Mastra workflows provide a structured way to manage opera
...
-Workflow steps are composable and typically run logic directly within the `execute` function. However, there are cases where calling an agent or tool is more appropriate. This pattern is especially useful when:
+Workflow steps can call agents to leverage LLM reasoning or run custom functions from your codebase. You can either invoke them from within a step's `execute` function or compose them directly as steps using `createStep()`.
 
-- Generating natural language responses from user input using an LLM.
-- Abstracting complex or reusable logic into a dedicated tool.
-- Interacting with third-party APIs in a structured or reusable way.
+## When to use agents and tools
 
-Workflows can use Mastra agents or tools directly as steps, for example: `createStep(testAgent)` or `createStep(testTool)`.
+Use agents in workflow steps when you need to solve open-ended tasks that require reasoning or language generation. Use tools when you need to call APIs, query databases, or run custom functions that perform specific operations.
```
bookercodes: I'm not sure if this recommendation makes sense. Personally, I don’t understand why I would call a tool from a step instead of a function, unless I already happened to have a tool. Even then, I might prefer to call a function that the step shares directly. Could you help me (and other readers) understand the reasoning here?

docs/src/content/en/docs/workflows/using-with-agents-and-tools.mdx
```diff
@@ -5,346 +5,119 @@ description: "Steps in Mastra workflows provide a structured way to manage opera
 
 # Agents and Tools
```
bookercodes: In the Agent section, you have a page called Using Tools, so maybe this should be using Tools and Agents or using Tools & Agents (if using &, we should be consistent about that in all headings)

docs/src/content/en/docs/workflows/using-with-agents-and-tools.mdx
```diff
@@ -5,346 +5,119 @@ description: "Steps in Mastra workflows provide a structured way to manage opera
...
+  }
 });
 ```
 
-### Adding an agent as a step
-
-In this example, `step1` uses the `testAgent` to generate an interesting fact about the country based on a given city.
-
-The `.map` method transforms the workflow input into a `prompt` string compatible with the `testAgent`.
+> See [Calling Agents](../../examples/agents/calling-agents.mdx) for more information.
```
bookercodes: Another opportunity to inline the link?

docs/src/content/en/docs/workflows/using-with-agents-and-tools.mdx
```diff
@@ -5,346 +5,119 @@ description: "Steps in Mastra workflows provide a structured way to manage opera
...
+const step1 = createStep({
+  // ...
+  execute: async ({ inputData, mastra }) => {
+    const { message } = inputData;
 
-```typescript filename="src/mastra/agents/test-agent.ts" showLineNumbers copy
-import { openai } from "@ai-sdk/openai";
-import { Agent } from "@mastra/core/agent";
+    const testAgent = mastra.getAgent("testAgent");
+    const response = await testAgent.generate(`Convert this to uppercase: ${message}`);
```
bookercodes: We should use an example of something an LLM is actually well-suited for, like processing a unstructured (PDF) document and extracting insights (it can be super contrived)

docs/src/content/en/docs/workflows/using-with-agents-and-tools.mdx
```diff
@@ -5,346 +5,119 @@ description: "Steps in Mastra workflows provide a structured way to manage opera
...
-
-In this example, `step1` uses the `testAgent` to generate an interesting fact about the country based on a given city.
-
-The `.map` method transforms the workflow input into a `prompt` string compatible with the `testAgent`.
+> See [Calling Agents](../../examples/agents/calling-agents.mdx) for more information.
 
-The step is composed into the workflow using `.then()`, allowing it to receive the mapped input and return the agent's structured output. The workflow is finalized with `.commit()`.
+### Using agents as steps
 
+When using an agent as a step, it expects a `prompt` as input. Use `.map()` to transform the workflow input or previous step’s output into this shape before invoking the agent.
```
bookercodes: I found this a bit hard to understand

Suggest:

> When using an agent as a step, it expects a prompt as input. Before invoking the agent, use ⁠.map() to transform your workflow input or the output from a previous step into an object with a ⁠prompt property.

docs/src/content/en/docs/workflows/using-with-agents-and-tools.mdx
```diff
@@ -5,346 +5,119 @@ description: "Steps in Mastra workflows provide a structured way to manage opera
...
     return {
-      country_name: country
+      formatted: response.formatted
     };
   }
 });
 ```
 
-### Adding a tool as a step
+> See [Calling Tools](../../examples/tools/calling-tools.mdx) for more information.
```
bookercodes: Can we denote this is an example? 

We have this pattern of adding a quote/callout with a link at the bottom of sections and linking to something more detailed

It feels confusing when these links take the user to different places (documentation or examples), and especially jarring if you're going from one to the other, as you lose your place

bookercodes: In this case, there's nothing interesting on that page you haven't covered already

docs/src/content/en/docs/workflows/using-with-agents-and-tools.mdx
```diff
@@ -5,346 +5,119 @@ description: "Steps in Mastra workflows provide a structured way to manage opera
...
 
-In this example, `step1` uses the `testTool`, which performs a geocoding lookup using the provided `city` and returns the resolved `country`.
+### Using tools as steps
 
-The step is added to the workflow using the sequential `.then()` method, allowing it to receive input from the workflow and return structured output. The workflow is finalized with `.commit()`.
+When using a tool as a step, it must be wrapped with `createStep()`. The tool's input must match the shape expected by its schema. Use `.map()` to transform the workflow input or previous step’s output as needed before invoking the tool.
```
bookercodes: Nit, but using "must" back to back doesn't read as well as it can

I'm also not sure about the language "When using a tool as a step, it must be wrapped with `createStep()`" 

"Must" sounds a tad forceful - could sound more collaborative and supportive (nit, but could be impactful if we apply this tone throughout  our writing)

docs/src/content/en/docs/workflows/using-with-agents-and-tools.mdx
```diff
@@ -5,346 +5,119 @@ description: "Steps in Mastra workflows provide a structured way to manage opera
...
 
 ![Tool as step](/image/workflows/workflows-agent-tools-tool-step.jpg)
 
-```typescript {1,3,6} filename="src/mastra/workflows/test-workflow.ts" showLineNumbers copy
+```typescript {1,3,12} filename="src/mastra/workflows/test-workflow.ts" showLineNumbers copy
```
bookercodes: should it b lines 9-15?

bookercodes (approved):
These updates look better - much tighter and focused. Solid first pass. 



The idea that agents can call workflows, workflows can call agents, agents can call tools, workflows can call tools, tools can call agents... etc. is really mind-bending for new users. 
2025-10-21 at 10 27 57@2x

I wonder if it might make sense to group the Agents and Tools sections under h2s? now instead of 4 concepts, it's just 2 - calling agents and calling tools - you can just go about it in different ways

You can take this page up a level by providing more guidance about when to use which and walking the user through it. At the moment, this feels more like a reference than a guide

PR #9040, opened 2025-10-20
fix: prevent test file modifications and changeset spam

TylerBarnes:
Fixed two annoying CI/test issues with the model router:

Tests were modifying provider-registry.json and provider-types.generated.d.ts locally. Now they use selective mocks that only block writes to these specific files while allowing temp file writes for testing. Files are restored after each test.

The CI workflow was creating a new automated changeset on every provider registry update. Now it checks if one already exists and skips creation if found, so we only get one changeset per alpha release cycle.


TylerBarnes:
Reese says: Great suggestion! Added the directory check to handle the case where `.changeset` doesn't exist yet. Also added a conditional around the `git add` to handle empty CHANGESET_FILE (when an existing changeset is found).

Reese

wardpeet (commented):
This means even between alphas we only publish this once?

TylerBarnes:
> This means even between alphas we only publish this once?

@wardpeet it'll only add a new changeset if one isn't already present, to prevent it adding 100 different ones that are all the same
cleanshot_2025-10-20_at_14 41 40_720

wardpeet:
@TylerBarnes i know but it's for each alpha. These changesets are only being cleared on a stable release

Imagine this scenario:
1. updated modals
2. next day, update model - skipped changeset
3. publish alpha
4. update modal - skipped changeset
5. stable release
6. updated modals
7. next day, update model - skipped changeset

TylerBarnes:
I think that's ok as long as the changeset triggers bumping the core version in the latest release

wardpeet (approved):

PR #9045, opened 2025-10-21
feat!: change generateTitle default to false

TylerBarnes:
Changed the default value of `threads.generateTitle` from `true` to `false`. Automatic title generation now requires explicit opt-in to avoid unexpected LLM API calls. I found it confusing when this would happen by default in playground. Sometimes I would be debugging the title llm call instead of the main one (I also saw others doing this)

Before:
```typescript
// Title generation happened automatically
const agent = new Agent({
  memory: new Memory({})
});
```

After:
```typescript
// Explicitly enable if you want title generation
const agent = new Agent({
  memory: new Memory({
    options: {
      threads: {
        generateTitle: true
      }
    }
  })
});
```

Also updated the playground UI to display thread IDs (last 5 chars) instead of "Chat from" when titles aren't generated, making it easier to identify threads.

Screenshot 2025-10-20 at 4 53 32 PM

Keeping this a draft for now until after this weeks release. I have multiple breaking change PRs for memory that I'm getting queued up.


mfrachet:
IIRC the intention was to have it by default on and force opting out, but I might be wrong cc @abhiaiyer91

TylerBarnes:
> IIRC the intention was to have it by default on and force opting out, but I might be wrong cc @abhiaiyer91

@mfrachet I confirmed in slack that we're good to make this change. I do like your idea of using the first x chars of the user message though, instead of the id

client-sdks/ai-sdk/package.json
```diff
@@ -22,7 +22,7 @@
     "CHANGELOG.md"
   ],
   "peerDependencies": {
-    "@mastra/core": ">=0.21.0-0 <0.24.0-0",
+    "@mastra/core": ">=0.21.0-0 <2.0.0-0",
```
TylerBarnes, line 25: hmm @wardpeet claude did this, I assume it's right but shouldn't main be like this already?

packages/cli/src/commands/dev/DevBundler.ts
```diff
@@ -107,9 +107,8 @@ export class DevBundler extends Bundler {
           }
         },
         plugins: [
-          // @ts-ignore - types are good
-
-          ...inputOptions.plugins,
+          // inputOptions.plugins is guaranteed to be Plugin[] by getWatcherInputOptions
+          ...((inputOptions.plugins as Plugin[]) || []),
```
TylerBarnes, line 111: @wardpeet local lint always fails on this too so claude ended up changing it

TylerBarnes:
Reese says: @greptileai Thanks for the thorough review! I've addressed all the issues you identified:

1. ✅ Fixed syntax error in `working-memory.test.ts` - corrected the missing closing brace and nesting
2. ✅ Updated all 6 documentation files to use the new top-level `generateTitle` API instead of the deprecated `threads.generateTitle` path

All changes have been committed separately with links to the specific review comments. The code should now be safe to merge.

Reese

LekoArts (commented):
You can drop all changes to the `ja` docs 😅

DanielSLew (approved):

TylerBarnes:
> You can drop all changes to the `ja` docs 😅

@LekoArts I know they're generated, claude updated those and I didn't think it really mattered since they'll be regenerated anyway

PR #9056, opened 2025-10-21
fix(deployer): Improve error message w/ binary dep

cian:
This is particularly useful for workspace dependencies ending with `.node` where the current error is a bit misleading.

## Description
I hit an interesting issue this morning, I have a workspace import like this:
`import { serverSideOnlyDbUtility } from "@our-monorepo/internalUtils/dbActions.node";

We use the `.node` convention alongside webpack in a next.js app to prevent accidental import of server-side code in any webapps.

This inadvertently triggered the `!id.endsWith('.node')` [here](https://github.com/mastra-ai/mastra/blob/959c6e9896e5edcfd99d03c1661baadb84a4842b/packages/deployer/src/build/analyze/bundleExternals.ts#L214), resulting in a rather misleading message:

```
We found a binary dependency in your bundle. Please add `@our-monorepo/anotherPackage` to your externals.
  
  bundler: {
    externals: ["@our-monorepo/anotherPackage"],
  }
})
```

Once I ran `mastra dev` in a debugger I could immediately see what was going on, and adjust the import statement above to not have `.node` in it. I didn't want to propose a change to the actual bundler behavior.

## Related Issue(s)

#6852

## Type of Change

- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [ ] Test update

## Checklist

- [NA] I have made corresponding changes to the documentation (if applicable)
- [NA] I have added tests that prove my fix is effective or that my feature works

### DIFF
--- a/.changeset/small-bees-heal.md
+++ b/.changeset/small-bees-heal.md
@@ -1,0 +1,5 @@
+---
+'@mastra/deployer': patch
+---
+
+Improve error related to finding possible binary dependencies
--- a/packages/deployer/src/build/analyze/bundleExternals.ts
+++ b/packages/deployer/src/build/analyze/bundleExternals.ts
@@ -227,7 +227,7 @@ async function getInputPlugins(
             },
             text: `We found a possible binary dependency in your bundle. ${id} was not found when imported at ${importer}.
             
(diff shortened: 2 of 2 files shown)


LekoArts (approved):
Thanks!

PR #9084, opened 2025-10-21
breaking(mcp): Remove deprecated options

DanielSLew:
- Remove MastraMCPClient/MCPConfigurationOptions/MCPConfiguration options that have been deprecated for a while
- Update tests to not use stream/generate legacy
- Update jsdocs/implementation to use model router


explorations/mcp-registry-client/packages/mcp-configuration/src/index.ts
```diff
@@ -225,17 +225,18 @@ export class MCPClient {
...
+								command: config.command,
+								args: [...(config.args || []), ...(config.runtimeArgs || [])],
+								env: config.env,
+							},
 						},
 					})
 
 			if (!exists) {
-				await mcpClient.connect()
+				// MCPClient connects automatically when tools are requested
```
TylerBarnes, line 239: don't we connect here so that it's connected before it's added to the by id map? Or does that not matter?

packages/mcp/package.json
```diff
@@ -47,15 +47,13 @@
...
-    "@ai-sdk/anthropic": "^1.2.12",
-    "@ai-sdk/openai": "^1.3.24",
     "@hono/node-server": "^1.19.5",
     "@internal/lint": "workspace:*",
     "@mastra/core": "workspace:*",
     "@mendable/firecrawl-js": "^1.29.3",
     "@microsoft/api-extractor": "^7.52.8",
     "@types/node": "^20.19.0",
-    "ai": "4.3.19",
+    "ai": "^5.0.60",
```
TylerBarnes, line 56: does mcp need `ai` as a dep?

packages/mcp/src/client/client.ts
```diff
@@ -720,15 +720,3 @@ export class InternalMastraMCPClient extends MastraBase {
...
- */
-
-export class MastraMCPClient extends InternalMastraMCPClient {
-  constructor(args: InternalMastraMCPClientOptions) {
-    super(args);
-    this.logger.warn(
-      '[DEPRECATION] MastraMCPClient is deprecated and will be removed in a future release. Please use MCPClient instead.',
-    );
-  }
-}
```
TylerBarnes, line 734: probably better to keep it for now but make it throw an error. That way we don't need a migration guide for it

packages/mcp/src/client/configuration.ts
```diff
@@ -875,47 +874,3 @@ To fix this you have three different options:
...
- * ```
- */
-export class MCPConfiguration extends MCPClient {
-  /**
-   * @deprecated Use MCPClient constructor instead
-   */
-  constructor(args: MCPClientOptions) {
-    super(args);
-    this.logger.warn(
-      `MCPConfiguration has been renamed to MCPClient and MCPConfiguration is deprecated. The API is identical but the MCPConfiguration export will be removed in the future. Update your imports now to prevent future errors.`,
```
TylerBarnes, line 918: same with this one

packages/mcp/src/client/index.ts
```diff
@@ -1,3 +1,2 @@
 export type { LoggingLevel, LogMessage, LogHandler, MastraMCPServerDefinition, ElicitationHandler } from './client';
-export { MastraMCPClient } from './client';
```
TylerBarnes, line 2: and here

TylerBarnes (changes requested):

packages/mcp/package.json
DanielSLew: I believe so, we use some of the imports in tests

.changeset/vast-rats-beg.md
```diff
@@ -0,0 +1,5 @@
+---
+'@mastra/mcp': patch
```
TylerBarnes: needs to be minor

explorations/mcp-registry-client/packages/mcp-configuration/src/index.ts
```diff
@@ -225,17 +225,18 @@ export class MCPClient {
...
 				this.mcpClientsById.set(id, mcpClient)
 				process.on(`exit`, () => {
 					mcpClient.disconnect()
```
TylerBarnes, line 242: ah yeah if we don't connect above and this exit handler is attached, mcpClient.disconnect() will throw an error when the process exits

TylerBarnes (changes requested):

explorations/mcp-registry-client/packages/mcp-configuration/src/index.ts
TylerBarnes: nvm didn't realize this is explorations, you can ignore 😄

TylerBarnes (approved):

PR #9085, opened 2025-10-21
feat(observability): Add parent span context to MastraSpan

cephalization:
## Description

When a MastraSpan is instantiated with a parent span id, also create a ParentSpanContext.

This allows for trace UIs to construct a tree of trace spans using parent hierarchy, with semantic attribute access via `parentId`.

This is what a trace tree looks like in a third party collector UI with parent span context added and no further changes needed.
 
image (7)

Prior to this change, the tree would be completely flat.

## Type of Change

- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [ ] Test update

## Checklist

### DIFF
--- a/.changeset/kind-falcons-allow.md
+++ b/.changeset/kind-falcons-allow.md
@@ -1,0 +1,6 @@
+---
+'@mastra/otel-exporter': patch
+'@mastra/arize': patch
+---
+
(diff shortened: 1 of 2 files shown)


epinzur (approved):
amazing. thanks for finding this and submitting the PR.

PR #9089, opened 2025-10-21
docs: provider options generation fix for Google/OpenAI + docs gen tests

TylerBarnes:
Fixes the provider options docs generation for Google and OpenAI providers. The script wasn't finding their types because they use `InferValidator<typeof schema>` instead of direct type aliases.

Updated the script to check `exportedDeclarations` to properly resolve these types, and added tests to verify all providers work correctly.

Now the Google and OpenAI provider docs correctly show their provider-specific options like `cachedContent`, `thinkingConfig`, `instructions`, etc.


docs/src/content/en/models/providers/google.mdx
```diff
@@ -353,6 +353,87 @@ const agent = new Agent({
...
+        "description": "",
+        "isOptional": true
+    },
+    {
+        "name": "structuredOutputs",
+        "type": "boolean | undefined",
+        "description": "",
+        "isOptional": true
+    },
+    {
```
TylerBarnes, line 398: these went missing when deps were updated! tests will catch it next time

DanielSLew (approved):

bookercodes (approved):

PR #9091, opened 2025-10-21
paul/grwth-899-workflow-state

PaulieScanlon:
## Description

- docs: first pass at adding state

## Type of Change

- [x] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [ ] Test update

## Checklist


docs/src/content/en/docs/workflows/overview.mdx
```diff
@@ -129,6 +141,66 @@ export const testWorkflow = createWorkflow({
...
+  })
+})
+  .then(step1)
+  .then(step2)
+  .commit();
+```
+
+<Callout type="info">
+  Workflow state is currently supported only when using [Run.start()](../../reference/workflows/run-methods/start.mdx). Support for `Run.stream()` will be added soon.
+</Callout>
```
PaulieScanlon, line 202: Added a callout to explain state is currently only supported by `.start()` we can circle back and remove this when `.stream()` is supported.

bookercodes: maybe that's a `warn`

docs/src/content/en/docs/workflows/overview.mdx
```diff
@@ -129,6 +141,66 @@ export const testWorkflow = createWorkflow({
   .commit();
 ```
 
+### Workflow state
+
+Workflow state lets you share values across steps without passing and defining them explicitly using `inputSchema` and `outputSchema`. Define all state values in the workflow’s `stateSchema`, but each step only needs to declare the values it uses. To set initial values, use `initialState` when running the workflow. See [Running workflows](#running-workflows) for details.
```
bookercodes: > Workflow state lets you share values across steps without passing and defining them explicitly using `inputSchema` and `outputSchema`.
the user might wonder, "why would I want to do that?"

docs/src/content/en/docs/workflows/overview.mdx
```diff
@@ -164,22 +236,36 @@ Workflows can be run in two modes: start waits for all steps to complete before
 
 <Tabs items={["Start", "Stream"]}>
   <Tabs.Tab>
-Create a workflow run instance using `createRunAsync()`, then call `.start()` with `inputData` matching the workflow's `inputSchema`. The workflow executes all steps and returns the final result.
+Create a workflow run with `createRunAsync()`, then call `.start()` with `inputData` matching the workflow's `inputSchema`. To initialize workflow state, include `initialState` alongside `inputData`. The workflow runs all steps and returns the final result.
+
+To return state values in the result, set `includeState: true` in `outputOptions`.
```
bookercodes, line 241: this is tricky

it does make sense to co-locate with start

but this doesn't apply unless you're using state, and most users aren't (it's new) and probably that won't change

docs/src/content/en/docs/workflows/overview.mdx
```diff
@@ -164,22 +236,36 @@ Workflows can be run in two modes: start waits for all steps to complete before
...
 
 console.log(result);
 ```
+
+> See [Run.start()](../../reference/workflows/run-methods/start.mdx) for a full list of configuration options.
+
   </Tabs.Tab>
   <Tabs.Tab>
-Create a workflow run instance using `.createRunAsync()`, then call `.stream()` with `inputData` matching the workflow's `inputSchema`. The workflow emits events as each step executes, which you can iterate over to track progress.
+Create a workflow run with `createRunAsync()`, then call `.stream()` with `inputData` matching the workflow's `inputSchema`. The workflow emits events as steps run, allowing you to monitor progress in real time.
```
bookercodes: don't think real time is right, maybe more like "as they happen"

bookercodes (approved):

PR #9100, opened 2025-10-22
[docs] Moving Memory, Workflows, and Tools examples

justindeguzman:


docs/src/content/en/docs/workflows/using-with-agents-and-tools.mdx
```diff
@@ -37,6 +37,19 @@ export const testAgent = new Agent({
...
+```json
+{
+  inputSchema: {
+    prompt: string
+  },
+  outputSchema: {
+    text: string
+  }
+}
+```
```
PaulieScanlon: I just made some big changes to Agents and Tools, i think this part should be added though. I can either create a PR or, are you able to resolve the merge conflicts @justindeguzman?

PaulieScanlon (changes requested):

docs/src/content/en/docs/workflows/human-in-the-loop.mdx
PaulieScanlon, line 1: Does this file belong in docs? If we do want a human-in-the-loop section in the docs i think this example will need to be refactored. It feels out of place being moved over as it is.

PaulieScanlon (changes requested):

docs/src/content/en/docs/memory/storage/memory-with-libsql.mdx
PaulieScanlon, line 1: Same as the workflows comments. If we are going to move examples into docs, they will need some refactoring. I'm also not convinced storage should be nested under memory. Could the dir be named Memory & Storage perhaps?

PaulieScanlon (changes requested):

PaulieScanlon (approved):

PR #9106, opened 2025-10-22
get and work with sub agents via api

taofeeq-deru:
## Description

When get agent api is called, we now also check for subAgents registered on a parent agent, making it possible to work with them via api

## Type of Change

- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [ ] Test update

## Checklist


packages/server/src/server/handlers/agents.ts
```diff
@@ -249,6 +249,47 @@ async function formatAgentList({
...
+  let agent;
+
+  try {
+    agent = mastra.getAgent(agentId);
+  } catch (error) {
+    logger.debug('Error getting agent from mastra, searching agents for agent', error);
+  }
+
+  if (!agent) {
+    logger.debug('Agent not found, searching agents for agent', { agentId });
```
TheIsrael1:
```suggestion
    logger.debug(`Agent ${agentId} not found, looking through sub-agents`);
```

TheIsrael1 (approved):

PR #9134, opened 2025-10-22
break(core, mcp): add mcp property to tool type

DanielSLew:
## Description

Rather than the mcp package monkeypatching the tool type to create it's own tool type use one unified tool type that has an optional `mcp` property that provides the tool with any mcp related features that come from being attached to an MCPServer class.

Tools used in an mcp server now have a different signature (but now the properties are also typed properly)

Tools can be easily written in both the mcp context and non mcp context now fully typed.

If added to an MCPServer it will use the `mcp` condition, otherwise it will use the other condition.

```
export const accountBalance = createTool({
  id: 'account-balance',
  description: 'Fetches the current account balance for a given account',
  inputSchema: z.object({
    accountId: z.string(),
  }),
  outputSchema: z.object({
    balance: z.number(),
  }),
  suspendSchema: z.object({
    message: z.string(),
  }),
  resumeSchema: z.object({
    confirm: z.boolean(),
  }),
  execute: async ({ context, mcp, suspend, resumeData}) => {
    const question = `Is it ok to fetch the account balance for account ${context.accountId}?`;
    
    if (mcp) {
      const result = await mcp.elicitation.sendRequest({
        message: question,
        requestedSchema: {
          type: 'object',
          properties: {
            confirm: { type: 'boolean' },
          },
          required: ['confirm'],
        }
      });
      if (result.action === 'accept') {
        return {
          balance: getAccountBalance(context.accountId),
        };
      } else if (result.action === 'decline') {
        return 'User declined to provide account balance';
      } else {
        return 'User cancelled account balance request';
      }
    } else {
      if (!resumeData) {
        return await suspend?.({
          message: question,
        });
      }
      if (!resumeData.confirm) {
        return 'User declined to fetch the account balance';
      }
      return {
        balance: getAccountBalance(context.accountId),
      };
    }
  },
});

```

## Type of Change

- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [ ] Test update

## Checklist


packages/mcp/src/server/server.ts
```diff
@@ -351,7 +349,7 @@ export class MCPServer extends MCPServerBase {
       return {
         tools: Object.values(this.convertedTools).map(tool => {
           const toolSpec: any = {
-            name: tool.name,
+            name: tool.id || 'unknown',
```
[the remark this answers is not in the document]
TylerBarnes: @DanielSLew why switch from using name here?

packages/mcp/src/server/server.ts
```diff
@@ -417,12 +415,16 @@ export class MCPServer extends MCPServerBase {
...
           messages: [],
           toolCallId: '',
-          elicitation: sessionElicitation,
-          extra,
-        });
+          // Pass MCP-specific context through the mcp property
+          mcp: {
+            elicitation: sessionElicitation,
+            extra,
+          },
```
TylerBarnes, line 425: since this might be hard to figure out that the properties moved, I think we should add getter functions for the old ones that just throw errors:

```ts
get elicitation() { throw new Error(`The elicitation key is now nested under mcp.elicitation in tool arguments`) }
get extra() { same }
```

packages/mcp/src/server/server.ts
```diff
@@ -854,13 +856,12 @@ export class MCPServer extends MCPServerBase {
...
-        parameters: coreTool.parameters,
-        outputSchema: coreTool.outputSchema,
-        execute: coreTool.execute!,
-        toolType: 'workflow',
-      };
+        ...coreTool,
+        id: workflowToolName,
+        mcp: {
+          toolType: 'workflow',
+        },
```
TylerBarnes, line 871: is toolType from the mcp spec?

packages/mcp/src/server/server.ts
```diff
@@ -1642,11 +1640,11 @@ export class MCPServer extends MCPServerBase {
     return {
       tools: Object.entries(this.convertedTools).map(([toolId, tool]) => ({
         id: toolId,
-        name: tool.name,
+        name: tool.id || toolId,
```
TylerBarnes, line 1651: why change this part?

TylerBarnes (changes requested):

packages/mcp/src/server/server.ts
DanielSLew: No this is something that we use specifically for our MCP tools to figure out if they're from agents or workflows so we can render it in UI differently. Before it was just a property called `toolType` but I thought I'd make it more clear that this is for an MCP tool type only

packages/mcp/src/server/server.ts
DanielSLew: The type changed. Before we created a new internal tool type called ConvertedTool which was just because our mcp tools and non mcp tool types were slightly different. We had a name property on it instead of `id` but it was essentially the same thing

packages/mcp/src/server/server.ts
DanielSLew: yeah this is a good idea

packages/mcp/src/server/server.ts
TylerBarnes: ahh got it, thanks

TylerBarnes (approved):

PR #9151, opened 2025-10-23
refactor(ai-tracing): consolidate span operations in ModelSpanTracker

epinzur:
## Description

Centralized all LLM generation span operations within ModelSpanTracker to eliminate redundant llmAISpan parameter passing.

**Changes:**

- **ModelSpanTracker API**: Added public methods for span operations
  - `getTracingContext()`: Returns current step span or model span
  - `reportGenerationError()`: Report errors on generation span
  - `endGeneration()`: End generation span with output/attributes
  - `updateGeneration()`: Update generation span
  - Made internal methods private (startStepSpan, endStepSpan, etc.)

- **Removed llmAISpan parameter**: No longer passed through call chain
  - Removed from LoopOptions, LoopRun, OuterLLMRun types
  - Removed from loop(), workflowLoopStream() signatures
  - llmAISpan now only used locally in model.loop.ts

- **Updated call sites**: Use ModelSpanTracker methods instead
  - model.loop.ts: Use reportGenerationError() and endGeneration()
  - loop.ts, stream.ts: Use getTracingContext() for tracing context
  - tool-call-step.ts: Use getTracingContext() for tool spans

**Benefits:**
- Single source of truth for generation span operations
- Cleaner API with fewer parameters
- Better encapsulation of span lifecycle
- Consistent pattern for accessing tracing context

## Related Issue(s)

#6773

## Type of Change

- [ ] Documentation update
- [X] Code refactoring
- [ ] Performance improvement
- [ ] Test update

## Checklist


packages/core/src/ai-tracing/model-tracing.ts
```diff
@@ -33,10 +34,50 @@ export class ModelSpanTracker {
     this.modelSpan = modelSpan;
   }
 
+  /**
+   * Get the tracing context for creating child spans.
+   * Returns the current step span if active, otherwise the model span.
+   */
+  getTracingContext(): { currentSpan?: AnyAISpan } {
+    return {
+      currentSpan: this.currentStepSpan ?? this.modelSpan,
```
wardpeet: What's the difference between currentSpan and modelspan?

packages/core/src/ai-tracing/model-tracing.ts
```diff
@@ -33,10 +34,50 @@ export class ModelSpanTracker {
...
+    metadata?: Record<string, any>;
+  }): void {
+    this.modelSpan?.update(options);
+  }
+
   /**
    * Start a new Model execution step
    */
-  startStepSpan(payload?: StepStartPayload) {
+  private startStepSpan(payload?: StepStartPayload) {
```
wardpeet: use actual private methods
```suggestion
  #startStepSpan(payload?: StepStartPayload) {
```

packages/core/src/ai-tracing/model-tracing.ts
epinzur: currentStepSpan is the span for the currentStep, which doesn't exist before the first chunk arrives.  

modelSpan is the span that owns all stepsSpans.

packages/core/src/ai-tracing/model-tracing.ts
epinzur: learn something new every day. thanks!

wardpeet (approved):

PR #9153, opened 2025-10-23
Rename runCount to retryCount

taofeeq-deru:
## Description

This PR deprecates runCount and replaces it with retryCount

## Type of Change

- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [x] Test update

## Checklist


.changeset/bright-taxis-raise.md
```diff
@@ -0,0 +1,6 @@
+---
+'@mastra/inngest': patch
+'@mastra/core': patch
+---
+
+Deprecate runCount and replace it with retryCount
```
LekoArts: Please explain in more detail what is deprecated and why. On which classes is runCount used? Give an example on how to migrate (a before/after), either through two code blocks or a `diff` block

LekoArts (changes requested):

.changeset/bright-taxis-raise.md
taofeeq-deru: thanks, updated the changeset

.changeset/bright-taxis-raise.md
```diff
@@ -0,0 +1,60 @@
+---
+'@mastra/inngest': patch
+'@mastra/core': patch
```
LekoArts, line 3: When you add both packages to this changeset they will both receive this changelog. The inngest package only received an internal change by us, so it should be removed from here

.changeset/bright-taxis-raise.md
```diff
@@ -0,0 +1,60 @@
...
+    result: z.string()
+  }),
+  execute: async ({ retryCount, ...params }) => {
+    console.log(`Retry attempt: ${retryCount}`);
+    // ... rest of your logic
+  }
+});
+```
+
+This also applies to condition functions and loop condition functions that use this parameter.
```
LekoArts: This entry will be added to the CHANGELOG file and will be part of an unordered list in there. So with that context in mind, we should use headings sparsely and be concise (but have all the information in there). What's the relevant public API that a user would care about?

Here's how I'd imagine a more concise write-up:

````suggestion
Deprecate `runCount` parameter in favor of `retryCount` for better naming clarity. The name `runCount` was misleading as it doesn't represent the total number of times a step has run, but rather the number of retry attempts made for a step.

`runCount` is available in `execute()` functions and methods that interact with the step execution. This also applies to condition functions and loop condition functions that use this parameter. If your code uses `runCount`, change the name to `retryCount`.

Here's an example migration:

```diff
const myStep = createStep({
  // Rest of step...
-  execute: async ({ runCount, ...params }) => {
+  execute: async ({ retryCount, ...params }) => {
    // ... rest of your logic
  }
});
```
````

LekoArts (dismissed):

.changeset/bright-taxis-raise.md
taofeeq-deru: the same thing applies to the inngest package though, for anyone using `createStep` from `@mastra/inngest` `init`

LekoArts: Ah, you're right, my bad 👍

rase- (approved):

PR #9203, opened 2025-10-23
feat!: remove deprecated APIs and complete VNext transition

DanielSLew:
- Removes modelSettings.abortSignal in favour of top-level abortSignal only. Also removes the deprecated output field - use structuredOutput.schema instead.

- The deprecated generateVNext() and streamVNext() methods have been removed since they're now the stable generate() and stream() methods.

- The deprecated `output` option has been removed entirely, in favour of `structuredOutput`.

Method renames to clarify the API surface:
- getDefaultGenerateOptions → getDefaultGenerateOptionsLegacy  
- getDefaultStreamOptions → getDefaultStreamOptionsLegacy
- getDefaultVNextStreamOptions → getDefaultStreamOptions

These are breaking changes that require users to update their code if using any of the removed/renamed APIs.


TylerBarnes (commented):
generally looks good! but needs a callout in the v1 migration guide and tests are failing

LekoArts (changes requested):
Please add a migration guide to https://github.com/mastra-ai/mastra/blob/main/docs/src/content/en/guides/migrations/upgrade-to-v1.mdx and make the changeset a `major`

TylerBarnes (approved):

PR #9211, opened 2025-10-24
fix(core): Don't call `os.homedir()` at top level

NickCrews:
## Description

I am trying to use this library in the deno Slack SDK. Slack is running my code in a sandboxed environment, and if I merely import @mastra/core then I get

```
   error: Uncaught (in promise) NotCapable: Requires sys access to "homedir", run again with the --allow-sys flag
       at Object.homedir (node:os:94:10)
       at file:///Users/nc/Library/Caches/deno/npm/registry.npmjs.org/@mastra/core/0.22.2/dist/chunk-A3TQYIIR.js:2019:30
```

This fixes this by refactoring these constants to be simple functions that lazily call os.homedir

## Type of Change

- [ ] Documentation update
- [x] Code refactoring
- [ ] Performance improvement
- [ ] Test update

## Checklist

(this isn't really testable, as long as we think we have adequate existing coverage for this refactor then I think that is adequate)


wardpeet (approved):
oh definitly, we should even refactor it more but this looks good! Thank you!

wardpeet:
Created backport PR: https://github.com/mastra-ai/mastra/pull/9357

PR #9258, opened 2025-10-24
fix(core): use schema-compat transformed schemas for tool input validation

### LINKED ISSUE #8805: [DISCORD:1427319015478657034] update working memory tool call errors on searchString
This issue was created from Discord post: https://discord.com/channels/1309558646228779139/1427319015478657034

The update working memory tool call regularly errors in the first call w/ this error.

```
"state": "result",
            "toolCallId": "call_W84bP8Lo2qCwIYe03QDLCgTJ",
            "toolName": "updateWorkingMemory",
            "args": {
                "newMemory": "#User\n- First Name: Randy\n- Last Name: Lynn",
                "searchString": null,
                "updateReason": "append-new-memory"
            },
"result": {
                "error": true,
                "message": "Tool validation failed for updateWorkingMemory. Please fix the following errors and try again:\n- searchString: Expected string, received null\n\nProvided arguments: {\n  \"newMemory\": \"#User\\n- First Name: Randy\\n- Last Name: Lynn\",\n  \"searchString\": null,\n  \"updateReason\": \"append-new-memory\"\n}",
                "validationErrors": {
                    "_errors": [],
                    "searchString": {
                        "_errors": [
                            "Expected string, received null"
                        ]
                    }
                }
            }
```

from what I understand this is internal tools and prompting within Mastra.
the agent definitely figures this out and makes a tool call again, passing `searchstring = ""` and successfully updates the memory. 
Would be nice to have this looked at to avoid the multiple tool calls for updateWorkingMemory.

TylerBarnes:
Fixes a validation bug where tools would reject valid LLM responses because validation used the original Zod schema while the LLM received a schema-compat transformed version.

This caused issues with OpenAI o3 reasoning models (which convert `.optional()` to `.nullable()` and send `null` values) and Claude 3.5 Haiku (which strips `min`/`max` string constraints). Validation would reject these responses even though they matched what the LLM was told was valid.

The fix restructures `CoreToolBuilder.build()` to apply schema compatibility before creating the execute function, then passes the processed Zod schema to `createExecute()` for validation. Now validation uses the same schema that was sent to the LLM.

Added comprehensive tests covering Anthropic string constraints, OpenAI o3 nullable conversion, and the specific `updateWorkingMemory` bug that triggered this investigation.

Before:

Screenshot 2025-10-24 at 1 31 27 PM

After:
Screenshot 2025-10-24 at 1 30 26 PM

Fixes https://github.com/mastra-ai/mastra/issues/8805


TylerBarnes:
Reese says:

@greptileai Good catch on the code path! However, this is actually correct behavior for provider-defined tools.

**Why provider-defined tools don't need `processedSchema`:**

1. **Provider tools are already provider-specific**: Tools like `google.tools.googleSearch()` come from the provider's SDK (e.g., `@ai-sdk/google`) and are already designed for that specific provider's schema expectations.

2. **They skip schema compatibility entirely**: The `build()` method returns early at line 335 when `buildProviderTool` succeeds, meaning provider-defined tools intentionally bypass the entire schema compatibility process (lines 340-422).

3. **No transformation needed**: At line 114, `buildProviderTool` uses `convertZodSchemaToAISDKSchema` for basic conversion, but doesn't apply provider-specific compatibility layers. This is correct because the tool is already provider-compatible.

4. **Execute function is rare**: While provider-defined tools *can* have an `execute` function (line 118-124), this is uncommon. When present, it's defined by the provider SDK, not by Mastra, so it expects the original schema format.

**The bug this PR fixes** only affects Mastra-defined tools (ToolAction, VercelTool, VercelToolV5) where:
- Mastra applies schema compatibility transformations for cross-provider support
- The LLM receives the transformed schema
- Validation must use the same transformed schema

Provider-defined tools don't have this mismatch because they're never transformed in the first place.

DanielSLew (approved):

wardpeet:
Created backport PR: https://github.com/mastra-ai/mastra/pull/9360

PR #9266, opened 2025-10-25
fix: improve reliability of string constraints in schema compat, and make tests not flaky

TylerBarnes:
Before these changes tons of models would fail to generate structured output making the tests flaky.
I improved the reliability a few ways, while also re-enabling tool tests which were disabled:

1. ~~only use json prompt injection. It seems schema response api is actually not as good, I think we need to enable schema compatibility for it to work better. Without prompt injection many simple field tests fail, with it there's 100% success~~ edit: just disabled these for now
2. for tools, the schema compatibility constraints instruction was confusing and made the models think the constraint was the schema (especially for regex), ~~so I added a note on that, and made regex constraints clearer~~ edit: I ended up making all constraints into natural language descriptions
3. I updated the tests so each field description is not an instruction "I need x" but is a description "an x that follows y pattern"

With these changes, 100% success rate. Screenshot shows with all the models but I commented a bunch out because I think it's a waste to run these tests on all of them always
Screenshot 2025-10-24 at 6 43 56 PM


DanielSLew (approved):

wardpeet:
Created backport PR: https://github.com/mastra-ai/mastra/pull/9362

PR #9396, opened 2025-10-28
 Fix duplicate assistant messages when using useChat

### LINKED ISSUE #9370: [BUG] Mastra with ai-sdk v5 may cause duplicate assistant messages
### Describe the Bug

When using `useChat` from `@ai-sdk/react` to call Mastra's API, assistant messages may have duplicate parts. In the messages list, some assistant messages contain previous parts.

It occurred in my project and can also be reproduced 100% in the [Official AI SDK v5 example](https://github.com/mastra-ai/mastra/tree/main/examples/ai-sdk-v5) when sending user messages quickly. It reproduces less frequently at normal chat speed.

After some debugging, I found that it might be caused in the call stack `addOne() -> addPartsToLatestMessage() -> pushNewMessagePart()` in [message-list](https://github.com/mastra-ai/mastra/blob/main/packages/core/src/agent/message-list/index.ts#L1163) in `prepare-memory-step`.

100% reproduced when calling `/chat` API with messages that contain `{"type":"step-start"}` part in assistant message. 

**Possible calling trace:**

1. user send message A
2. ai respond message B
3. user send message C(call `/chat` API with the following message with step-start part)
```json
{
  "agentName": "editorAgent",
  "threadId": "fed58330-517f-465a-95fc-fc80302e4b08",
  "id": "mw89qp2duTDiuRpA",
  "messages": [
    {
      "role": "user",
      "parts": [{ "type": "text", "text": "hi" }],
      "id": "N8lo8qlwX3rtuPb1"
    },
    {
      "id": "519c0c40-459e-4623-8ecc-bbd0a5459388",
      "metadata": {
        "totalUsage": {
          "inputTokens": 1872,
          "outputTokens": 81,
          "totalTokens": 1953,
          "cachedInputTokens": 1
        }
      },
      "role": "assistant",
      "parts": [
        { "type": "step-start" },
        {
          "type": "text",
          "text": "Hello! I am the AI assistant of the Galacean editor, and I am happy to help you.",
          "state": "done"
        }
      ]
    },
    {
      "role": "user",
      "parts": [{ "type": "text", "text": "my name is Jim" }],
      "id": "2FkPWZ0FDA6xQng9"
    }
  ],
  "trigger": "submit-message"
}
```
4. In the `prepare-memory-step`, it first initializes MemoryList with messages from storage ([A, B]), and then adds messages from the client ([A, B, C]) into it.

https://github.com/mastra-ai/mastra/blob/ee04303344ea6fc38a69e7624fe66e6bf4e47467/packages/core/src/agent/workflows/prepare-stream/prepare-memory-step.ts#L225-L230

5. It finally goes into the function `pushNewMessagePart`. When adding message B from the client, it contains an extra part `{ "type": "step-start" }`, and the original message part index is 1, which is bigger than message B in storage, so in this function it decides to merge parts.

https://github.com/mastra-ai/mastra/blob/ee04303344ea6fc38a69e7624fe66e6bf4e47467/packages/core/src/agent/message-list/index.ts#L1163-L1171

Do you have any advice to avoid it?

### Steps To Reproduce

1. Run example https://github.com/mastra-ai/mastra/tree/main/examples/ai-sdk-v5
2. When chatting with the agent, send messages frequently without calling `initial-chat`.

### Expected Behavior

_No response_

### Environment Information

- Mastra 0.23.1
- Node.js 22.11.0
- macOS
- LLM Provider(alibaba-cn/deepseek-r1)
- Chrome

### Verification

Higangssh:
## Fix duplicate assistant messages when using useChat

  Fixes #9370

  ### Changes

  Added ID check to prevent merging assistant messages from memory with new responses.

  The issue occurred because `shouldAppendToLastAssistantMessage` didn't verify message IDs, causing old assistant
  messages (retrieved from storage) to merge with new ones when `useChat` resent the message history.

  ### Solution

  Only enforce ID matching when `latestMessage` is from memory:
  ```typescript
  (!this.memoryMessages.has(latestMessage) || !latestMessage.id || !messageV2.id || latestMessage.id ===
  messageV2.id)
```
  This preserves existing behavior for tool-calls while preventing the duplicate issue.

  Testing

  - Added 2 regression tests for the issue
  - All tests passing (81/81)

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevents incorrect merging of assistant replies when past responses are replayed from memory, so distinct assistant messages remain separate while preserving correct streaming merges.
* **Tests**
  * Added tests validating duplicate-assistant prevention and correct merging for streaming updates.
* **Changelog**
  * Patch entry documenting the fix for duplicate assistant messages during history resends.

### CODE REVIEW

**Comment by Higangssh** [comment]
Hi! Just wanted to explain my approach here.

  I found that the issue happens when `useChat` resends messages with `step-start` parts. The old message from
  storage (1 part) would merge with the new one (2 parts) even though they have different IDs.

  I tried a few approaches:
  - First attempted strict ID checking for all input messages → broke tool-call merging
  - Then realized the issue only happens with messages from memory

  So I added the ID check specifically for `memoryMessages`:
  ```typescript
  (!this.memoryMessages.has(latestMessage) || ID check)
```
  This fixes the duplicate while keeping tool-calls working normally. Let me know if you'd prefer a different
  approach!

Higangssh (commented):
Fixed as suggested.

Higangssh:
Hi! Just wanted to explain my approach here.

  I found that the issue happens when `useChat` resends messages with `step-start` parts. The old message from
  storage (1 part) would merge with the new one (2 parts) even though they have different IDs.

  I tried a few approaches:
  - First attempted strict ID checking for all input messages → broke tool-call merging
  - Then realized the issue only happens with messages from memory

  So I added the ID check specifically for `memoryMessages`:
  ```typescript
  (!this.memoryMessages.has(latestMessage) || ID check)
```
  This fixes the duplicate while keeping tool-calls working normally. Let me know if you'd prefer a different
  approach!

**Comment by abhiaiyer91** [comment]
Hi @Higangssh thank you so much for the PR! We'll take a look!

**Comment by wardpeet** [comment]
I'm still learning all the differences between the frameworks, but in ai-sdk we could do, to dupllcate. Can't we do it for assistant ui as well?
```
      const result = await agentObj.stream<OUTPUT, 'mastra'>(messages, {
        ...defaultOptions,
        ...rest,
        runtimeContext: runtimeContext || defaultOptions?.runtimeContext,
      });

      const uiMessageStream = createUIMessageStream({
        originalMessages: messages,
        execute: async ({ writer }) => {
          for await (const part of toAISdkFormat(result, { from: 'agent' })!) {
            writer.write(part);
          }
        },
      });

      return createUIMessageStreamResponse({
        stream: uiMessageStream,
      });
```

Is there a similar approach we can do?

TheIsrael1:
@Higangssh  👆

Higangssh:
Hi @wardpeet and @TheIsrael1! The Thanks for the great question - it made me dig deeper into how the streaming patterns work.

  I looked into the code you mentioned, and I think I found something interesting. The `createUIMessageStream`
  pattern is already being used in `client-sdks/ai-sdk/src/chat-route.ts`, and there's a similar pattern for
  Assistant UI with `toAssistantUIMessage`.

  **Here's my understanding of the issue** (please correct me if I'm wrong!):

  I believe the duplicate bug happens at a different layer than what those streaming wrappers handle. From what I
  can see:
  - The streaming patterns (`createUIMessageStream`, `toAssistantUIMessage`) focus on format conversion
  - But the duplicate issue occurs in the `MessageList` class itself, where it was merging messages without checking
   if IDs matched
  - When `useChat` resends stored messages, old assistant messages were getting merged with new responses

  So my thinking is: even with the streaming patterns in place, we still need this MessageList fix to prevent the
  duplication at the core level.

  **Does this understanding make sense?** I might be missing something about how the streaming wrappers could handle
   this differently. Would love to hear your thoughts on whether there's a better approach at the streaming layer
  instead!

kmk142789 (approved):

Higangssh:
@coderabbitai generate docstrings

abhiaiyer91 (approved):

wardpeet:
Thank you for the pull request. Upon investigation, the issue is due to not passing initialMessages to the createUIMessageStream. I have created a pull request where you can reuse our logic from chatRoute with Next.js, which handles deduplication for you.

https://github.com/mastra-ai/mastra/pull/10628

Issue #5848, opened 2025-07-11
[BUG] workflow startAsync in mastra client incorrect error type


Nikamura:
### Describe the Bug

Typescript types indicate that result.error should be of type Error, but it seems to always be string.
`    error: Error;`

### Steps To Reproduce

```


      const mastra = new MastraClient();

      const workflow = mastra.getWorkflow("x");

      const run = await workflow.createRun();


      const result = await workflow.startAsync({
        runId: run.runId,
        inputData: {
          data
        },
      });

   if (result.status === "failed") {
        const error: unknown = result.error;
        if (error instanceof Error) {
          throw error;
        } else if (typeof error === "string") {
          throw new Error(error); // this throws
        }
}
```

### Verification

- [x] I have searched the existing issues to make sure this is not a duplicate
- [x] I have included sufficient information for the team to reproduce and understand the issue

wardpeet:
Feel free to update the types yourself, I think we should rethrow the error on the client, like we do on the server.

DanielSLew:
Closing this in favour of https://github.com/mastra-ai/mastra/issues/9348, a better proposed solution for this

PR #9399, opened 2025-10-28
fix: correct WorkflowResult error type from Error to string

RichardLoRicco:
Fixes issue #5848 where the `WorkflowResult` type declared `error` as an `Error` object, but the workflow system actually returns error messages as strings at runtime.

**Primary fix:** Updated `WorkflowResult` interface to correctly type the error field as `string` instead of `Error`.

**Secondary fix:** While investigating, discovered the agent-builder was also making incorrect assumptions about the error type. Added a type guard to safely handle both strings and Error objects.

**Before:**
```typescript
// Type definition - incorrect
interface WorkflowResult {
  error: Error;  // Runtime value is actually string
}

// Agent builder transformWorkflowResult - assumes Error object
} else if (result.status === 'failed') {
  return {
    success: false,
    applied: false,
    message: `Agent builder action failed: ${result.error.message}`,  // Would throw on string
    error: result.error.message,
  };
}
```

After:
```typescript
// Type definition - corrected
interface WorkflowResult {
  error: string;  // Matches runtime behavior
}

// Agent builder transformWorkflowResult - handles both safely
} else if (result.status === 'failed') {
  const errorMessage = typeof result.error === 'string' ? result.error : result.error.message;
  return {
    success: false,
    applied: false,
    message: `Agent builder action failed: ${errorMessage}`,
    error: errorMessage,
  };
}
```

Added tests to verify the corrected behavior.

Co-Authored-By: Harold Camacho <harold.camacho.diaz@gmail.com>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved error handling consistency by ensuring all errors are properly serialized and typed as strings over HTTP, replacing variable error object formats.

* **Tests**
  * Added test case to verify error messages are consistently handled as strings in workflow results.


client-sdks/client-js/src/resources/agent-builder.test.ts
```diff
@@ -130,3 +130,63 @@ describe('AgentBuilder.runs', () => {
     expect(params.get('resourceId')).toBe('test-resource-456');
   });
 });
+
+describe('AgentBuilder.transformWorkflowResult', () => {
```
DanielSLew: I don't think we need these tests, I think the workflow test is enough that we're passing through the correct type

DanielSLew (approved):
Just a small nit but looks good

PR #9410, opened 2025-10-28
fix: reduce flakiness in DynamoDB resource test

CalebBarnes:
## Summary

Fixes flaky DynamoDB resource test that was failing intermittently due to millisecond precision differences in timestamps when data round-trips through DynamoDB.

The test was using strict equality (`toEqual()`) to compare the entire resource object, including timestamps. DynamoDB can sometimes introduce microsecond-level precision differences (e.g., `2025-10-28T20:30:06.662Z` vs `2025-10-28T20:30:06.663Z`), causing the test to fail randomly.

## Changes

- Compare timestamp fields (`createdAt`, `updatedAt`) separately with 5ms tolerance
- Compare non-timestamp fields (`id`, `workingMemory`, `metadata`) individually
- Add spread-based comparison as additional safety check

## Test Plan

- [x] Run DynamoDB tests multiple times to verify flakiness is resolved
- [x] Verify typecheck passes
- [x] Verify linting passes

## Summary by CodeRabbit

* **Tests**
  * Strengthened internal tests to validate resource and thread fields individually and to use tolerant timestamp comparisons to reduce flakiness.

**Note:** This update affects internal testing only and does not change end-user functionality.


stores/_test-utils/src/domains/memory/resources.ts
```diff
@@ -10,7 +10,16 @@ export function createResourcesTest({ storage }: { storage: MastraStorage }) {
 
       // Save resource
       const savedResource = await storage.saveResource({ resource });
-      expect(savedResource).toEqual(resource);
+      expect(savedResource.id).toEqual(resource.id);
+      expect(savedResource.workingMemory).toEqual(resource.workingMemory);
+      expect(savedResource.metadata).toEqual(resource.metadata);
+      // Compare timestamps within 5ms tolerance to avoid flakiness from precision differences
+      expect(Math.abs(savedResource.updatedAt.getTime() - resource.updatedAt.getTime())).toBeLessThan(5);
+      expect(Math.abs(savedResource.createdAt.getTime() - resource.createdAt.getTime())).toBeLessThan(5);
```
TylerBarnes, line 18: probably need to check that it's greater than -5 too

DanielSLew: I got around this in other tests by setting the vitest timers and expecting that same value

CalebBarnes: From claude:

`Math.abs()` already handles that case perfectly. `Math.abs()` returns the absolute value (the distance from zero), which always makes negative numbers positive:
- If `savedResource.updatedAt` is 5ms later than `resource.updatedAt`:
`savedResource.updatedAt.getTime()` - `resource.updatedAt.getTime()` = `5`
`Math.abs(5) = 5`

- If `savedResource.updatedAt` is 5ms earlier than `resource.updatedAt`:
`savedResource.updatedAt.getTime()` - `resource.updatedAt.getTime() = -5`
`Math.abs(-5) = 5`

So whether the saved resource timestamp is ahead or behind, `Math.abs()` converts it to a positive number, and then `.toBeLessThan(5)` checks if it's within the 5ms tolerance in either direction.

CalebBarnes: @DanielSLew I think if we use mock/fake dates then we could possibly miss real timestamp precision issues introduced at some point in the future.  So far I've only seen dynamodb be off by 1 microsecond and everything else is spot on.

kmk142789 (approved):

PR #9471, opened 2025-10-29
fix: remove Japanese content from llms.txt files to reduce size

TylerBarnes:
This removes Japanese content from the llms.txt generation script since LLMs are multilingual and English documentation is sufficient.

The script was processing both English and Japanese directories, resulting in very large files (llms.txt was 78k tokens, llms-full.txt was 1.7M tokens).

Before:
```typescript
const enDir = path.join(contentDir, 'en');
const jaDir = path.join(contentDir, 'ja');

const enDocs = await processDirectory(enDir, 'en');
const jaDocs = await processDirectory(jaDir, 'ja');
const allDocs = [...enDocs, ...jaDocs];
```

After:
```typescript
const enDir = path.join(contentDir, 'en');

const enDocs = await processDirectory(enDir, 'en');
const allDocs = [...enDocs];
```

This should significantly reduce the file sizes while maintaining full functionality since the English docs contain all the necessary information for LLMs.

## Chores
* Simplified documentation build process by consolidating to English-only support, removing Japanese documentation processing to streamline the build workflow.


mfrachet (approved):
LGTM

DanielSLew (approved):

PR #9481, opened 2025-10-29
fix: restore original agent model when resetting model in UI

CalebBarnes:
## Problem

When an agent is created with a `LanguageModelV2` object (e.g., `openai()`), switching models in the playground UI and then clicking the reset button fails to restore the original model. Instead, it throws an error:

```
Error: Could not find config for provider ... with model id ...
```

**Root cause**: When `__updateModel()` is called from the UI, it overwrites `agent.model` with a string (e.g., `"openai/gpt-4"`), permanently losing the original `LanguageModelV2` object reference. The reset button tried to reconstruct the model using model router format with the extracted provider/modelId, which doesn't work for models that weren't created via model router.

## Solution

Store the original model configuration in a private `#originalModel` field when the agent is constructed, and add a server endpoint to restore it.

### Changes

**Core** (`packages/core`)
- Add `#originalModel` field to Agent class to preserve initial model config
- Add `__resetToOriginalModel()` method to restore from stored original

**Server** (`packages/server`)
- Add `resetAgentModelHandler` endpoint: `POST /api/agents/:agentId/model/reset`

**Deployer** (`packages/deployer`)
- Add route definition with OpenAPI documentation
- Add Hono handler wrapper

**Client SDK** (`client-sdks/client-js`)
- Add `resetModel()` method to Agent resource class

**Playground UI** (`packages/playground-ui`)
- Add `useResetAgentModel()` hook
- Wire up reset button to call new endpoint
- Add `useEffect` to sync UI state when props change after reset

## Testing

1. Create an agent with a `LanguageModelV2` (e.g., `openai()`)
2. Switch to a different model via the UI picker
3. Click the reset button
4. Verify the original `LanguageModelV2` is restored (not a model router string)

## Type of Change

## Summary by CodeRabbit

* **New Features**
  * Agent model reset: restore an agent's model to its original configuration from the model picker UI (supports all model types), fixing the reset behavior.
* **API & SDK**
  * Programmatic reset endpoint and client SDK method to trigger agent model resets.
* **UX**
  * Model switcher now uses the reset flow to keep UI state consistent after a reset.


packages/core/src/agent/agent.ts
```diff
@@ -983,6 +987,20 @@ export class Agent<
...
+  /**
+   * Resets the agent's model to the original model that was set during construction.
+   * This restores the original LanguageModelV2 or model configuration that may have
+   * been overwritten by __updateModel().
+   * @internal
+   */
+  __resetToOriginalModel() {
+    // Restore from snapshot; clone to avoid aliasing future mutations
+    this.model = Array.isArray(this.#originalModel)
+      ? this.#originalModel.map(mdl => ({ ...mdl }))
```
TylerBarnes:
```suggestion
      ? [...this.#originalModel]
```

TylerBarnes (approved):

PR #9587, opened 2025-10-31
fix: update all tool signatures to v1.0 (input, context) format

DanielSLew:
Updated all tool execution signatures across the entire codebase from the old `({ context })` format to the new v1.0 `(inputData, context)` format.

This makes it easier to reason what properties will be available inside a tool depending on which context you're using the tool in. This fixes up the typescript inference too with tools, before you had to do a lot of casting and add as any's a lot.

Consolidated the 3 different execution contexts to one 
```typescript
// before depending on the context the tool was executed in
tool.execute({ context: data })
tool.execute({ context: { inputData: data } })
tool.execute(data)

// now, for all contexts
tool.execute(data, context)
```

**Before:**
```typescript
inputParams: z.object({ something: z.string() }),
execute: async ({ context, tracingContext, runId, ... }) => {
  return doSomething(context.string);
}
```

**After:**
```typescript
inputParams: z.object({ something: z.string() }),
execute: async (inputData, context) => {
  const { agent, mcp, workflow, ...sharedContext } = context
  
  // context that only an agent would get like toolCallId, messages, suspend, resume, etc
  if (agent) {
    doSomething(inputData.something, agent)
  // context that only a workflow would get like runId, state, suspend, resume, etc
  } else if (workflow) {
    doSomething(inputData.something, workflow)
  // context that only a workflow would get like "extra", "elicitation"
  } else if (mcp) {
    doSomething(inputData.something, mcp)
  } else {
    // Running a tool in no execution context
    return doSomething(inputData.something);
  }
}
```

## Summary by CodeRabbit

* **Breaking Changes**
  * Tool execute signature changed: tools now receive input as the first argument and execution context as a separate second argument — custom tools and integrations must be updated.

* **Refactor**
  * Unified input/context handling across agents, workflows, and server tools for consistent invocation and context nesting.

* **Documentation**
  * Examples, guides, and references updated to show the new input-first execute signature and updated context access.


examples/dane/src/mastra/workflows/link-checker.ts
```diff
@@ -32,8 +32,8 @@ const getBrokenLinks = new Step({
   outputSchema: z.object({
     brokenLinks: z.array(linkSchema),
   }),
-  execute: async ({ context }) => {
-    const targetUrl = context?.triggerData?.targetUrl;
+  execute: async (input, context) => {
```
abhiaiyer91: i think this should be inputData

epinzur (approved):

.changeset/famous-keys-type.md
```diff
@@ -0,0 +1,62 @@
...
+tool.execute(data);
+
+// now, for all contexts
+tool.execute(data, context);
+```
+
+**Before:**
+
+```typescript
+inputParams: z.object({ something: z.string() }),
```
DanielSLew:
```suggestion
inputSchema: z.object({ something: z.string() }),
```

.changeset/famous-keys-type.md
```diff
@@ -0,0 +1,62 @@
...
+inputSchema: z.object({ something: z.string() }),
+execute: async ({ context, tracingContext, runId, ... }) => {
+  return doSomething(context.string);
+}
+```
+
+**After:**
+
+```typescript
+inputParams: z.object({ something: z.string() }),
```
DanielSLew:
```suggestion
inputSchema: z.object({ something: z.string() }),
```

PR #9595, opened 2025-11-01
feat: add feature flags export to @mastra/core

TylerBarnes:
Adds a new  export that allows dependent packages to check for feature support in the core package.

Dependent packages can now import and check feature flags like this:

```typescript
import { coreFeatures } from "@mastra/core/features";

if (coreFeatures.someNewFeature) {
  // Use the new feature
}
```

This makes it easier to maintain backwards compatibility and feature detection across the Mastra ecosystem as new features are added to core.

## Summary by CodeRabbit

* **New Features**
  * Added a new public "features" export to access available core features.
  * Enables runtime querying of feature availability by other packages.
  * The feature registry is initialized for runtime population so available features can be discovered dynamically.

* **Documentation**
  * Included usage guidance and examples for consuming the new features export.


abhiaiyer91 (approved):

packages/core/src/features/index.ts
```diff
@@ -0,0 +1,19 @@
...
+ *
+ * if (coreFeatures.someNewThing) {
+ *   doWhatever()
+ * }
+ * ```
+ */
+export const coreFeatures: Record<string, boolean> = {
+  // Add feature flags here as new features are introduced
+  // Example: someNewFeature: true,
+};
```
wardpeet: I would even do a Set instead
```suggestion
// Add feature flags here as new features are introduced
export const coreFeatures  = new Set<string>();
```

TylerBarnes: good call! changed it

PR #9703, opened 2025-11-04
fix: combine assistant messages in stream() calls

DanielSLew:
This fixes the message splitting behavior in stream() calls by combining consecutive assistant messages into a single message with multiple parts.

Previously, assistant messages with different content types (text, tool-calls) were split into separate messages for backwards compatibility. Since we no longer need v1 message compatibility inside stream(), we can now combine them.

The main change is removing the tool-call exception that prevented assistant messages from being combined, and adding logic to recalculate the message type when combining messages.

Before:
- Assistant message with text
- Assistant message with tool-call (separate)
- Tool result
- Assistant message with text (separate)

After:
- Assistant message with text + tool-call (combined)
- Tool result
- Assistant message with text

This results in cleaner message handling and fewer split messages with __split-N suffixes.

### DIFF
--- a/packages/core/src/agent/message-list/prompt/convert-to-mastra-v1.ts
+++ b/packages/core/src/agent/message-list/prompt/convert-to-mastra-v1.ts
@@ -16,18 +16,15 @@ const makePushOrCombine = (v1Messages: MastraMessageV1[]) => {
 
   return (msg: MastraMessageV1) => {
     const previousMessage = v1Messages.at(-1);
-    if (
-      msg.role === previousMessage?.role &&
-      Array.isArray(previousMessage.content) &&
-      Array.isArray(msg.content) &&
-      // we were creating new messages for tool calls before and not appending to the assistant message
-      // so don't append here so everything works as before
-      (msg.role !== `assistant` || (msg.role === `assistant` && msg.content.at(-1)?.type !== `tool-call`))
-    ) {
+    if (msg.role === previousMessage?.role && Array.isArray(previousMessage.content) && Array.isArray(msg.content)) {
       for (const part of msg.content) {
         // @ts-ignore needs type gymnastics? msg.content and previousMessage.content are the same type here since both are arrays
         // I'm not sure what's adding `never` to the union but this code definitely works..
         previousMessage.content.push(part);
+      }
+      // Recalculate the type when combining assistant messages
+      if (previousMessage.role === 'assistant' && Array.isArray(previousMessage.content)) {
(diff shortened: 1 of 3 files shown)


packages/core/src/agent/message-list/prompt/convert-to-mastra-v1.ts
```diff
@@ -16,19 +16,16 @@ const makePushOrCombine = (v1Messages: MastraMessageV1[]) => {
...
+    if (msg.role === previousMessage?.role && Array.isArray(previousMessage.content) && Array.isArray(msg.content)) {
       for (const part of msg.content) {
         // @ts-ignore needs type gymnastics? msg.content and previousMessage.content are the same type here since both are arrays
         // I'm not sure what's adding `never` to the union but this code definitely works..
         previousMessage.content.push(part);
       }
+      // Recalculate the type when combining assistant messages
+      if (previousMessage.role === 'assistant' && Array.isArray(previousMessage.content)) {
+        previousMessage.type = previousMessage.content.some((c: any) => c.type === 'tool-call') ? 'tool-call' : 'text';
+      }
```
TylerBarnes, line 28: I don't think this is the right way to do this, v1 always gets split, we want to combine multiple messages into one MastraDBMessage with parts, which wouldn't happen in this file

packages/core/src/agent/message-list/tests/message-list.test.ts
```diff
@@ -3426,14 +3426,13 @@ describe('MessageList', () => {
...
+      // assistant tool-call and subsequent text are now combined into one message
+      expect(v1Again.length).toBe(4); // 4 messages because assistant parts are combined
       expect(v1Again[0].id).toBe('user-1');
-      expect(v1Again[1].id).toBe('msg-1');
-      expect(v1Again[2].id).toBe('msg-1__split-1'); // assistant tool-call (preserved)
-      expect(v1Again[3].id).toBe('msg-1__split-1'); // tool result (preserved - no double suffix!)
-      expect(v1Again[4].id).toBe('msg-1__split-2'); // assistant text (preserved)
+      expect(v1Again[1].id).toBe('msg-1'); // assistant with tool-call now combined with text
+      expect(v1Again[2].id).toBe('msg-1__split-1'); // tool result
+      expect(v1Again[3].id).toBe('msg-1__split-2'); // assistant text (now combined with previous assistant)
```
TylerBarnes, line 3435: v1 should stay as-is, this should go back. Actually might be worth closing this PR and making a new one since I think we don't want any of the changes in this PR

TylerBarnes (changes requested):

PR #9705, opened 2025-11-04
fix: remove deprecated content and toolInvocations from MastraMessageContentV2

DanielSLew:
Removes the deprecated `content` and `toolInvocations` fields from `MastraMessageContentV2` type. All message content is now stored exclusively in the `parts` array, aligning with the AI SDK v5 pattern.

**Before:**
```typescript
const message = {
  content: {
    format: 2,
    content: 'hello world',  // deprecated
    parts: [{ type: 'text', text: 'hello world' }]
  }
}
```

**After:**
```typescript
const message = {
  content: {
    format: 2,
    parts: [{ type: 'text', text: 'hello world' }]
  }
}
```

Updated all storage adapters, test utilities, processors, and core message handling to extract content from the `parts` array. Added type casts where needed to handle TypeScript's strict union type inference when accessing `.text` on text parts.

## Summary by CodeRabbit

* **New Features**
  * Added a utility to reconstruct text from structured message parts.

* **Bug Fixes**
  * Content extraction now exclusively uses parts, improving consistency for moderation, PII/language detection, embeddings, and processors.

* **Refactor**
  * Unified message model: text, tool calls, and attachments consolidated into a parts-based content representation, aligning UI and storage shapes.

* **Tests**
  * Test suites and integration tests updated to validate the new parts-based content shape.

* **Chores**
  * Storage update APIs adjusted to accept parts-based content payloads.


packages/core/src/agent/message-list/index.ts
```diff
@@ -1036,17 +1037,8 @@ export class MessageList {
...
-              const toolInvocationIndex = latestMessage.content.toolInvocations.findIndex(
-                t => t.toolCallId === existingCallPart.toolInvocation.toolCallId,
-              );
-              if (toolInvocationIndex === -1) {
-                latestMessage.content.toolInvocations.push(existingCallPart.toolInvocation);
-              } else {
-                latestMessage.content.toolInvocations[toolInvocationIndex] = existingCallPart.toolInvocation;
-              }
+              // Tool invocations are now managed only within parts array
+              // No need to maintain a separate toolInvocations array
```
TylerBarnes:
```suggestion
```

packages/core/src/agent/message-list/index.ts
```diff
@@ -1068,16 +1060,22 @@ export class MessageList {
...
-      if (
-        latestMessage.content.content &&
-        messageV2.content.content &&
-        latestMessage.content.content !== messageV2.content.content
-      ) {
-        // Match what AI SDK does - content string is always the latest text part.
-        latestMessage.content.content = messageV2.content.content;
+      // Extract text content from parts for comparison and update
+      const latestTextContent = latestMessage.content.parts.find(p => p.type === 'text')?.text;
+      const messageV2TextContent = messageV2.content.parts.find(p => p.type === 'text')?.text;
```
TylerBarnes: looks like this behaviour doesn't match what it did before, it said content.content was the latest text part, but now we're grabbing the first one

I wonder though, do we even need any of the code here anymore?

packages/core/src/agent/message-list/index.ts
```diff
@@ -1068,16 +1060,22 @@ export class MessageList {
...
+        const textPartIndex = latestMessage.content.parts.findIndex(p => p.type === 'text');
+        if (textPartIndex === -1) {
+          latestMessage.content.parts.unshift({ type: 'text', text: messageV2TextContent });
+        }
+      } else if (latestTextContent && messageV2TextContent && latestTextContent !== messageV2TextContent) {
+        // Update existing text part with new content
+        const textPart = latestMessage.content.parts.find(p => p.type === 'text');
+        if (textPart && textPart.type === 'text') {
+          textPart.text = messageV2TextContent;
+        }
```
TylerBarnes: I don't think this makes sense, we shouldn't be updating text parts, the whole point is you can have multiple text parts in the array and they're ordered in the order they happened in. This code looks like it assumes there's only a single text part. Maybe we can just delete all of this code that was added? All new parts should be added into the end of the parts array. The only time we should mutate parts is when we get a tool result and need to update a tool invocation to change the state and add the result

packages/core/src/agent/message-list/index.ts
```diff
@@ -1415,7 +1420,7 @@ export class MessageList {
       parts: message.parts,
     };
 
-    if (message.toolInvocations) content.toolInvocations = message.toolInvocations;
+    // Tool invocations are now stored only in parts array
```
TylerBarnes: should probably remove these comments

packages/core/src/agent/message-list/index.ts
```diff
@@ -2079,29 +2083,8 @@ export class MessageList {
...
-        } else {
-          parts.push({
-            type: `tool-${invocation.toolName}`,
-            toolCallId: invocation.toolCallId,
-            state: invocation.state === 'call' ? 'input-available' : 'input-streaming',
-            input: invocation.args,
-          } as unknown as AIV5Type.UIMessage['parts'][number]);
-        }
-      }
-    }
```
TylerBarnes, line 2114: we do actually still need to check for incoming deprecated fields from the db. So if the message source is `memory` we should check each deprecated field and check if wee have parts yet for each deprecated item. Otherwise we'll have to document how people need to do db migrations to keep their old data working, but it's easier for us/them if we just pull the data in at runtime.

Not sure if this specific code location is where we should do it, but we should do it for all deprecated fields somewhere in message list

packages/core/src/agent/message-list/index.ts
```diff
@@ -898,18 +893,24 @@ export class MessageList {
...
+      .map(p => p.text)
+      .join('\n');
+
+    if (message.role !== `system` || !textContent)
       throw new MastraError({
         id: 'INVALID_SYSTEM_MESSAGE_FORMAT',
         domain: ErrorDomain.AGENT,
         category: ErrorCategory.USER,
-        text: `Invalid system message format. System messages must include 'role' and 'content' properties. The content should be a string.`,
+        text: `Invalid system message format. System messages must include 'role' and text content in parts.`,
```
TylerBarnes, line 951: system messages shouldn't get this treatment, they only have a single content string

TylerBarnes: the reason for the change is regular messages have many ordered parts where tool calls, reasoning, text, can be mixed in in any order, system messages don't have that, they're static

packages/core/src/agent/message-list/index.ts
```diff
@@ -2291,9 +2274,11 @@ export class MessageList {
...
 
-    // 5. Handle text content (fallback if no parts)
-    if (dbMsg.content.content && !hasNonToolReasoningParts) {
-      parts.push({ type: 'text', text: dbMsg.content.content });
+    // 5. Text content is now only stored in parts array
+    // Extract text from parts if needed for fallback
+    const textFromParts = dbMsg.content.parts?.find(p => p.type === 'text')?.text;
+    if (textFromParts && !hasNonToolReasoningParts && !parts.some(p => p.type === 'text')) {
+      parts.push({ type: 'text', text: textFromParts });
     }
```
TylerBarnes: this looks like it's taking all text parts, combining them into one and pushing a duplicate combined text part? definitely remove this

TylerBarnes: we need to maintain the order of all parts, we shouldn't be combining or reordering anything

packages/core/src/agent/message-list/prompt/convert-to-mastra-v1.ts
```diff
@@ -88,10 +88,13 @@ export function convertToV1Messages(messages: Array<MastraDBMessage>) {
...
-          const userContent = experimental_attachments
-            ? [{ type: 'text', text: content || '' }, ...attachmentsToParts(experimental_attachments)]
-            : { type: 'text', text: content || '' };
+        if (parts == null || parts.length === 0) {
+          // Extract text content from parts if available
+          const textContent = inputParts?.find(p => p.type === 'text')?.text || '';
+          const userContent =
+            experimental_attachments.length > 0
+              ? [{ type: 'text', text: textContent }, ...attachmentsToParts(experimental_attachments)]
+              : { type: 'text', text: textContent };
```
TylerBarnes, line 104: wont this always evaluate to false? when we're going from db -> v1 I think we'd actually have to create attachments from file parts

packages/core/src/agent/message-list/prompt/convert-to-mastra-v1.ts
```diff
@@ -115,11 +118,8 @@ export function convertToV1Messages(messages: Array<MastraDBMessage>) {
             ...fields,
             type: 'text',
             content:
-              Array.isArray(userContent) &&
-              userContent.length === 1 &&
-              userContent[0]?.type === `text` &&
-              typeof content !== `undefined`
-                ? content
+              Array.isArray(userContent) && userContent.length === 1 && userContent[0]?.type === `text`
+                ? userContent[0].text
```
TylerBarnes, line 131: this should be the last text part or all text parts combined, but really maybe we should just output a content array for v1? pretty sure v1 still supported a content array

packages/core/src/agent/message-list/prompt/convert-to-mastra-v1.ts
```diff
@@ -252,95 +252,33 @@ export function convertToV1Messages(messages: Array<MastraDBMessage>) {
...
 
           break;
         }
 
-        const toolInvocations = message.content.toolInvocations;
+        // Extract tool invocations from parts
+        const partsArray = (message.content as any).parts || [];
+        const toolInvocationParts = partsArray.filter(
+          (p: any) => p.type === 'tool-invocation' && p.toolInvocation?.toolName !== 'updateWorkingMemory',
+        ) as any[];
```
TylerBarnes: this should be done without casting as `any`

packages/core/src/agent/message-list/prompt/convert-to-mastra-v1.ts
```diff
@@ -252,95 +252,33 @@ export function convertToV1Messages(messages: Array<MastraDBMessage>) {
...
+          pushOrCombine({ role: 'assistant', ...fields, content: textContent, type: 'text' });
           break;
         }
 
-        const maxStep = toolInvocations.reduce((max, toolInvocation) => {
-          return Math.max(max, toolInvocation.step ?? 0);
+        const maxStep = toolInvocationParts.reduce((max: number, part: any) => {
+          const step = part.toolInvocation?.step ?? 0;
+          return Math.max(max, step);
         }, 0);
```
TylerBarnes: should remove any for any newly added loc, also anywhere we're accessing text parts we either need to combine all of them into one or we need to just grab the very last text part, but probably combining is the way to go

packages/core/src/agent/message-list/prompt/convert-to-mastra-v1.ts
```diff
@@ -252,95 +252,33 @@ export function convertToV1Messages(messages: Array<MastraDBMessage>) {
...
 
         for (let i = 0; i <= maxStep; i++) {
-          const stepInvocations = toolInvocations.filter(
-            toolInvocation => (toolInvocation.step ?? 0) === i && toolInvocation.toolName !== 'updateWorkingMemory',
-          );
+          const stepInvocations = toolInvocationParts.filter((part: any) => (part.toolInvocation?.step ?? 0) === i);
```
TylerBarnes, line 305: no any and why remove the working memory check? seems like it was probably there for a reason

packages/core/src/agent/message-list/prompt/convert-to-mastra-v1.ts
```diff
@@ -352,26 +290,28 @@ export function convertToV1Messages(messages: Array<MastraDBMessage>) {
             ...fields,
             type: 'tool-call',
             content: [
-              ...(isLastMessage && content && i === 0 ? [{ type: 'text' as const, text: content }] : []),
-              ...stepInvocations.map(({ toolCallId, toolName, args }) => ({
+              ...(isLastMessage && textContent && i === 0 ? [{ type: 'text' as const, text: textContent }] : []),
+              ...stepInvocations.map((part: any) => ({
```
TylerBarnes, line 318: no any

packages/core/src/agent/message-list/prompt/convert-to-mastra-v1.ts
```diff
@@ -352,26 +290,28 @@ export function convertToV1Messages(messages: Array<MastraDBMessage>) {
...
 
           if (invocationsWithResults.length > 0) {
             pushOrCombine({
               role: 'tool',
               ...fields,
               type: 'tool-result',
-              content: invocationsWithResults.map((toolInvocation): ToolResultPart => {
-                const { toolCallId, toolName, result } = toolInvocation;
+              content: invocationsWithResults.map((part: any): ToolResultPart => {
+                const { toolCallId, toolName, result } = part.toolInvocation;
```
TylerBarnes, line 338: no any

packages/core/src/agent/message-list/tests/message-list.test.ts
```diff
@@ -393,19 +383,7 @@ describe('MessageList', () => {
...
-          state: 'result',
-          toolCallId: 'toolu_01Y9o5yfKqKvdueRhupfT9Jf',
-          toolName: 'weatherTool',
-          args: { location: 'Paris' },
-          result: {
-            temperature: 24.3,
-            conditions: 'Partly cloudy',
-          },
-        },
-      ]);
```
TylerBarnes, line 408: should remove all the llms comments from the whole pr, also anywhere we're removing an assertion like this, we need a new one to make sure the new code where these are in parts is working

packages/core/src/processors/runner.ts
```diff
@@ -341,10 +341,7 @@ export class ProcessorRunner {
 
       // Add system messages using addSystem
       for (const sysMsg of systemMessages) {
-        messageList.addSystem(
-          (sysMsg.content.content as string) ||
-            sysMsg.content.parts.map(p => (p.type === 'text' ? p.text : '')).join('\n'),
-        );
+        messageList.addSystem(sysMsg.content.parts.map(p => (p.type === 'text' ? p.text : '')).join('\n'));
```
TylerBarnes: system messages should still have content string

packages/memory/integration-tests/src/processors.test.ts
```diff
@@ -33,8 +33,13 @@ function getTextContent(message: any): string {
...
   }
-  if (typeof message.content?.content === 'string') {
-    return message.content.content;
+  // Extract text from parts array
+  if (message.content?.parts) {
+    const textParts = message.content.parts
+      .filter((p: any) => p.type === 'text')
+      .map((p: any) => p.text)
+      .join(' ');
+    if (textParts) return textParts;
```
TylerBarnes, line 42: maybe we should export a util for this, we're doing it manually all over our tests and likely people will need this kind of thing too anyway

stores/_test-utils/src/domains/memory/data.ts
```diff
@@ -89,9 +93,11 @@ export const createSampleMessageV2 = ({
     createdAt: createdAt || new Date(),
     content: {
       format: 2,
-      parts: content?.parts || [{ type: 'text', text: content?.content ?? '' }],
-      content: content?.content || `Sample content ${randomUUID()}`,
-      ...content,
+      parts: defaultParts,
+      reasoning: content?.reasoning,
+      experimental_attachments: content?.experimental_attachments,
+      annotations: content?.annotations,
```
TylerBarnes: aren't we removing these too?

stores/clickhouse/src/storage/domains/memory/index.ts
```diff
@@ -989,7 +989,7 @@ export class MemoryStorageClickhouse extends MemoryStorage {
     messages: (Partial<Omit<MastraDBMessage, 'createdAt'>> & {
       id: string;
       threadId?: string;
-      content?: { metadata?: MastraMessageContentV2['metadata']; content?: MastraMessageContentV2['content'] };
+      content?: { metadata?: MastraMessageContentV2['metadata']; content?: MastraMessageContentV2['parts'] };
```
TylerBarnes: these all look wrong

TylerBarnes (changes requested):

Issue #9244, opened 2025-10-24
[BUG]


calebeaires:
### Describe the Bug

Mastra clickhouse storage breaks when select at the table mastra_threads is done. 

### Steps To Reproduce

1. Implement @mastra/clickhouse storage
2. Start any chat
3. Mastra stores data into mastra_x tables
4. On select, mastra tries to select data from the table mastra_threads, the column metadata is empty

### Expected Behavior

@mastra/clickhouse inserts all values, but metadata is empty, it should be "{}", since it is going be parsed on select

Potential options to resolve
1. at insert, make sure metadata columns will have the values '{}'
2. on mastra_thread create table, use the option DEFAULT '{}'

### Environment Information

@mastra/clickhouse = ^0.15.6
@mastra/core=0.21.1

### Verification

- [x] I have searched the existing issues to make sure this is not a duplicate
- [x] I have included sufficient information for the team to reproduce and understand the issue

abhiaiyer91:
Hi @calebeaires we prob should do both! PRs are welcome!

graysonhicks:
Should be fixed by #9708

PR #9708, opened 2025-11-05
fix/9244 clickhouse metadata

graysonhicks:
## Description

Fixes duplicate thread rows appearing in ClickHouse and consolidates messy metadata serialization logic.
Problem
ClickHouse's ReplacingMergeTree engine only deduplicates during background merges (unpredictable timing). When title generation updates threads, duplicate rows appear until merges run. Additionally, metadata serialization/deserialization was inconsistent across the codebase.
### Solution
1. Thread deduplication:
Removed FINAL clauses (rely on unpredictable merges)
getThreadById: Added ORDER BY updatedAt DESC LIMIT 1 to always get newest row
listThreadsByResourceId: Uses COUNT(DISTINCT id) and ROW_NUMBER() OVER (PARTITION BY id) window function to deduplicate
getResourceById: Same pattern applied
saveThread: Simplified to just INSERT
2. Metadata handling:
Added serializeMetadata() helper: object → JSON string with '{}' fallback
Added parseMetadata() helper: JSON string → object with graceful error handling
Replaced all repetitive metadata logic with these helpers
Consistent behavior across all read/write operations

## Related Issue(s)

#9244

## Type of Change

- [x ] Bug fix (non-breaking change that fixes an issue)
- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [ ] Test update

## Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved metadata handling for threads to ensure consistency and proper data integrity.
  * Enhanced duplicate prevention in query results.

* **Improvements**
  * Optimized database queries to reliably fetch the latest version of each thread.
  * Enhanced data accuracy for pagination and thread listing operations.


abhiaiyer91 (approved):

graysonhicks:
The `SORT` / `LIMIT` queries work, but feel like a hack.  We were using `FINAL`, which should be returning a single unique result, but it was returning multiple.  I think this is because we weren't passing a `version` to the `ReplacingMergeTree` in `TABLE_ENGINES`.

Before merging this, need to test/determine best option here.

abhiaiyer91:
@graysonhicks oops i merged it

Issue #6477, opened 2025-08-01
[FEATURE] Unified Workflow API endpoint to create, run and get resull


mjh-sakh:
### Problem Statement

Three calls are needed for a single workflow run:
1. Creating a run: api/workflows/#{workflow_name}/create-run?runId=#{run_id}
2. Streaming to know when it finishes: api/workflows/#{workflow_name}/stream?runId=#{run_id}
3. Getting result: api/workflows/#{workflow_name}/runs/#{run_id}/execution-result

This adds latency for each extra network call that makes workflows unnecessary slower. 

### Proposed Solution

Add an endpoint that combines all three steps and returns final result. 

### Component

Workflows

### Example Use Case

Unified api endpoint will simplify workflow use by developers by having a single POST call in their code. 

### Verification

- [x] I have searched the existing issues to make sure this is not a duplicate
- [x] I have provided sufficient context for the team to understand the request

abhiaiyer91:
I can see the value in this if the workflow execution is happening in the same process and you are awaiting it. We can add a convenience method for this.

johnpccd:
I also noticed that the calls to check the run status or result returns the full data passed between the steps which can be huge.
An option to enable/disable that would be great.

abhiaiyer91:
With some changes down the pipe, doing this will be inconvenient in reality. Closing for now

PR #9739, opened 2025-11-05
feat: add unified workflow execute() method and /execute endpoint

RichardLoRicco:
Adds a new `execute()` method to the workflow client SDK and corresponding `/execute` endpoint for single-call workflow execution, addressing issue #6477.

This provides a more intuitive API for developers who want to run a workflow and get the result in one call, instead of the current 3-call pattern.

**Before:**
```typescript
const run = await workflow.createRunAsync();
const stream = await run.stream({ inputData: { value: 42 } });
for await (const chunk of stream) {
  if (chunk.type === 'workflow-finish') break;
}
const result = await workflow.runExecutionResult(run.runId);
```

**After:**
```typescript
const result = await workflow.execute({
  inputData: { value: 42 }
});
```

The implementation delegates to the existing `startAsyncWorkflowHandler`, maintaining consistency with current behavior while providing better discoverability through clearer naming and API design.

Related: #6477

Co-Authored-By: Harold Camacho <harold.camacho.diaz@gmail.com>

## Summary by CodeRabbit

* **New Features**
  * Added single-call workflow execution that creates and runs a workflow, returning status, runId, result and steps.
  * Supports optional custom runId and tracing options.

* **Documentation**
  * Added "Execute Workflow" reference with examples for basic use, custom runId, and tracing options.

* **Tests**
  * Added tests covering default and custom runId, input propagation, and result shape.


RichardLoRicco:
@coderabbitai review

Issue #7464, opened 2025-09-04
[DISCORD:1412844130832220170] Strategy when a workflow stuck in running state due to server restart


wardpeet:
This issue was created from Discord post: https://discord.com/channels/1309558646228779139/1412844130832220170

Hello,

A conversation was in progress, when server has restarted. That lead to the run being in state `running` and doesn't change to anything else by itself.

How would I approach this problem? Is there an automatic check mechanism in Mastra that can resolve this for me? Or the only way is to write a custom solution?

taofeeq-deru:
#9750 adds this enhancement.

PR #9750, opened 2025-11-05
Restart interrupted workflows

taofeeq-deru:
## Description

- Add restart method to workflow run that allows restarting an active workflow run
- Add status filter to `listWorkflowRuns`
- Add automatic restart to restart active workflow runs when server starts

## Related Issue(s)

#7464 

## Type of Change

- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [x] Test update

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Manual restart API for active workflow runs.
  * Status-based filtering when listing workflow runs.
  * Automatic restart of active workflow runs on server startup.

* **Tests**
  * Added tests for restart behavior and engine-specific restart constraints.

* **Chores**
  * Bumped multiple packages to patch versions.


packages/core/src/workflows/types.ts
```diff
@@ -9,6 +9,8 @@ import type { ConditionFunction, ExecuteFunction, LoopConditionFunction, Step }
 export type { ChunkType, WorkflowStreamEvent } from '../stream/types';
 export type { MastraWorkflowStream } from '../stream/MastraWorkflowStream';
 
+export type WorkflowEngineType = 'default' | 'evented' | 'inngest';
```
rase-: I think we should maybe just make this a string.

rase- (changes requested):

packages/core/src/workflows/types.ts
taofeeq-deru: okay, that makes sense, since we're adding more engines
made the change

rase- (approved):

PR #9790, opened 2025-11-06
fix: URL downloading to respect provider capabilities

### LINKED ISSUE #9584: [BUG] should not convert files into base64 for LLM providers that support providing documents as URL
### Describe the Bug

When providing messages like this:
```ts
{ role: "user", content: [ { type: "text", text: "analyze the image", }, 
{ type: "file", mediaType: "image/jpeg", filename: "image.jpg", data: "https://link-to-image.com/image.jpg", }, ], } 
```
They get converted to this before being sent to the LLM:
```ts
{ role: "user", content: [ { type: "text", text: "analyze the image", }, 
 {\"type\":\"image_url\",\"image_url\":{\"url\":\"data:image/jpeg;base64,/9j/4AA...}}
```
document urls need to be preserved for LLM providers that support passing documents that way.

original message from user:

1. In the message conversion logic, HTTP/HTTPS URLs ARE initially categorized and preserved:
In aiV5ModelMessageToMastraMessageV3, when handling file parts with string data, the code calls categorizeFileData() which checks if the data is a URL using isValidUrl().
[packages/core/src/agent/message-list/index.ts:2737-2748](https://github.com/mastra-ai/mastra/blob/main/packages/core/src/agent/message-list/index.ts#L2737-L2748)
[image-utils.ts:186-222](https://github.com/mastra-ai/mastra/blob/main/packages/core/src/agent/message-list/prompt/image-utils.ts#L186-L222)
Similarly, in mastraMessageV2ToMastraMessageV3, URLs are detected and preserved in the url field: https://github.com/mastra-ai/mastra/blob/main/packages/core/src/agent/message-list/index.ts#L2350-L2365
2. The Actual conversion happens later
However, before messages are sent to the provider, the llmPrompt() method calls downloadAssetsFromMessages(), which downloads ALL HTTP/HTTPS URLs unconditionally: [index.ts:234-255](https://github.com/mastra-ai/mastra/blob/main/packages/core/src/agent/message-list/index.ts#L234-L255)
3. All URLs Are Downloaded Unconditionally
The downloadAssetsFromMessages() function finds all URLs in file/image parts and downloads them, regardless of provider support: [download-assets.ts:56-95](https://github.com/mastra-ai/mastra/blob/main/packages/core/src/agent/message-list/prompt/download-assets.ts#L55-L95)
Critical Issue: The function calculates isUrlSupportedByModel (lines 78-84) but never uses this flag to filter which URLs to download. All URLs are downloaded unconditionally at lines 87-95.
4. Downloaded Data Replaces URLs
The convertImageFilePart() function then replaces the URLs with the downloaded binary data (Uint8Array): [convert-file.ts:28-35](https://github.com/mastra-ai/mastra/blob/main/packages/core/src/agent/message-list/prompt/convert-file.ts#L28-L35)
This binary data is then used in the LanguageModelV2FilePart that gets sent to the provider: [convert-file.ts:48-54](https://github.com/mastra-ai/mastra/blob/main/packages/core/src/agent/message-list/prompt/convert-file.ts#L48-L54)
5. Integration Point
The downloaded assets are applied to the messages before conversion to the final provider format: https://github.com/mastra-ai/mastra/blob/main/packages/core/src/agent/message-list/index.ts#L259-L288

Conclusion
External HTTP/HTTPS URLs are NOT preserved when sent to the OpenAI provider. They are downloaded and converted to base64 data URIs by the downloadAssetsFromMessages() function in the message preparation pipeline, which is called before messages reach any LLM provider.
The presence of the unused isUrlSupportedByModel flag suggests that the original design may have intended to only download URLs that are NOT natively supported by the provider (e.g., for Anthropic Claude and for OpenAI which supports external URLs). However, this conditional logic was never implemented, resulting in all URLs being downloaded regardless of provider capabilities.

### Steps To Reproduce

send a message to an LLM with a document URL

### Expected Behavior

send the document as URL if LLM provider is compatible

### Environment Information

_No response_

### Verification

nehaaprasad:
### Description

fixes: #9584 

- fixed bug where URLs were downloaded unconditionally even for LLM providers that support them natively. Now URLs are preserved for supported providers (like OpenAI) and only downloaded for unsupported ones.

**Changes:**
- Added filter in `downloadAssetsFromMessages()` to skip downloading supported URLs
- Added tests verifying correct behavior for both provider types

## Type of Change
- [x] Test update

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved URL handling so providers that natively support specific file types retain original URLs; only unsupported URLs are downloaded.

* **Tests**
  * Added tests verifying that supported URLs remain as URLs and that unsupported URLs are downloaded/converted to binary.

* **Chores**
  * Added a patch changeset documenting the adjusted download behavior for unsupported file types.


abhiaiyer91 (approved):

roaminro:
Hi @naaa760 ! The changes look good, but it looks like some "required" test are failing

wardpeet (approved):

PR #9792, opened 2025-11-06
feat(cli): Init new git repository

nehaaprasad:
### Description

Add git initialization option during the create wizard

### Related Issue(s)

Fixes #9547 

### Type of Change

## Summary by CodeRabbit

* **New Features**
  * Optional Git initialization during project creation — the setup wizard now prompts to initialize a Git repo and, if accepted, creates an initial commit (applies to both template and non-template flows).

* **Chores**
  * Added a changelog entry documenting the new Git initialization prompt.

* **Tests**
  * Updated test mocks to support the new confirmation prompt in creation flows.


packages/create-mastra/src/index.ts
LekoArts, line 1: Let's not add CLI flags for this, should only be in the wizard

packages/cli/src/commands/create/create.ts
LekoArts, line 1: Since we don't want CLI flags we can drop the changes here

packages/cli/src/commands/create/utils.ts
```diff
@@ -226,6 +228,35 @@ export const createMastraProject = async ({
     }
     s.stop('.gitignore added');
 
+    // Handle git initialization
+    let shouldInitGit = gitInit;
+    if (shouldInitGit === undefined && needsInteractive) {
+      const gitConfirmResult = await p.confirm({
+        message: 'Would you like to initialize a git repository?',
```
LekoArts:
```suggestion
        message: 'Initialize a new git repository?',
```

packages/cli/src/commands/create/utils.ts
```diff
@@ -226,6 +228,35 @@ export const createMastraProject = async ({
     }
     s.stop('.gitignore added');
 
+    // Handle git initialization
+    let shouldInitGit = gitInit;
+    if (shouldInitGit === undefined && needsInteractive) {
+      const gitConfirmResult = await p.confirm({
```
LekoArts: Put the prompt after this:

```ts
  if (needsInteractive) {
    result = await interactivePrompt({
      options: { showBanner: false },
      skip: { llmProvider: llmProvider !== undefined, llmApiKey: llmApiKey !== undefined },
    });
  }
```

This way all questions are asked upfront

LekoArts (changes requested):
Thanks for the PR, left some comments

LekoArts (approved):

PR #9798, opened 2025-11-06
fix: add exponential backoff to model retry logic

karamvirsingh1998:
## Problem
When AI model calls fail, the system immediately retries without any delay, causing:
- Rate limit , cascading failurs, api quota wastages

## Solution
 i have implemented exponential backoff (1s, 2s, 4s, 8s, max 10s) between retry attempts.

## Changes
- Modified `llm-execution-step.ts` to add delay before retrying failed model calls

this is critical fix and with no breaking changes . code is passing linting and changeset created

## Release Notes

* **New Features**
  * AI model operations now use exponential backoff retries with progressively longer delays (1-10 seconds) to handle temporary failures and rate limits more gracefully.


karamvirsingh1998:
@smthomas  PRM

rase- (approved):

Issue #9674, opened 2025-11-04
[FEATURE] allow configuring the API root path `/api`


roaminro:
### Problem Statement

as a developer I would like to be able to configure the root path of the API provided by the Mastra server

### Proposed Solution

Let users pass the API root path using the Mastra instance `server` configuration

### Component

CLI, Deployment

### Example Use Case

run Mastra server along another API server that already use the `/api` route.

### Verification

- [x] I have searched the existing issues to make sure this is not a duplicate
- [x] I have provided sufficient context for the team to understand the request

nehaaprasad:
Hi! @roaminro 
fix it , please let me know

PR #9858, opened 2025-11-07
feat: allow configuring API root path

nehaaprasad:
### Description

- Add `apiRootPath` configuration option to allow customizing the API root path from `/api` to any custom path (e.g., `/my-api`).

## Related Issue(s)

#9674

## Type of Change

**Changes:**
- Added `apiRootPath?: string` to `ServerConfig` type
- Updated all API routes to use configurable path instead of hardcoded `/api`
- Modified auth middleware to dynamically handle custom API paths
- Updated logging and HTML route handling for custom paths

## Summary by CodeRabbit

* **New Features**
  * Added configurable API root path setting to server configuration, allowing customization of the API endpoint root. Defaults to '/api' if not specified.
  * Authentication and authorization systems now respect the custom API root path configuration across all protected and public route handling.


ffMathy:
See also #10089

packages/core/src/server/types.ts
```diff
@@ -94,6 +94,11 @@ export type ServerConfig = {
    * Timeout for the server
    */
   timeout?: number;
+  /**
+   * Root path for the API endpoints
+   * @default '/api'
+   */
+  apiRootPath?: string;
```
LekoArts, line 101: Thanks for the PR!

In https://github.com/mastra-ai/mastra/pull/10441 I'm adding a `studioBase` option. Let's name this one in your PR `apiBase` to make it consistent 👍

PR #9950, opened 2025-11-10
feat(mcp) configurable connection timeout for http servers (#9891)

### LINKED ISSUE #9891: MCP client: hard coded initial connection timeout limit
the comment over this line  
https://github.com/mastra-ai/mastra/blob/2c4a7074d3765d54f3bad4dc2d7d9f7ab6bb2034/packages/mcp/src/client/client.ts#L352. 
reads  

>  this is hardcoded to 3s because the long default timeout would be extremely slow for sse backwards compat (60s)

I'm running a streaming http server on Vercel with https://github.com/vercel/mcp-handler#readme and it indeed has a significant latency before it becomes available (think mostly due to running in different AZs), certainly higher than the hardcoded 3 seconds. I'd suggest to make that parameter configurable and eventually warn when it conflicts with the reason you've given.

elmariachi111:
## Description

Makes the connect timeout for connecting mcp http streaming servers configurable. Was hardcoded to 3 seconds, mcp servers on Vercel *can* have longer latencies. 

## Related Issue(s)

Fixes https://github.com/mastra-ai/mastra/issues/9891#issuecomment-3513702167

## Type of Change

- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [ ] Test update

## Checklist

## Summary by CodeRabbit

* **New Features**
  * HTTP server connections now support a configurable connection timeout, with a 3-second default when not specified.
  * The new option applies to HTTP/streamable transports; other transports (e.g., stdio) explicitly disallow setting this timeout to avoid misconfiguration.


abhiaiyer91 (approved):

wardpeet (approved):
Thank you!

PR #9961, opened 2025-11-11
SEO: mark old v0 docs clearly as v0

axe312ger:
related to #9960

### DIFF
--- a/docs/docusaurus.config.js
+++ b/docs/docusaurus.config.js
@@ -29,6 +29,7 @@ const config = {
     },
     parseFrontMatter: async (params) => {
       const result = await params.defaultParseFrontMatter(params);
+      result.frontMatter.title = `${result.frontMatter.title} | Mastra v0`;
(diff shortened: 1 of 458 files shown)


docs/docusaurus.config.js
```diff
@@ -10,7 +10,7 @@ import "dotenv/config";
 
 /** @type {import('@docusaurus/types').Config} */
 const config = {
-  title: "Mastra Documentation",
+  title: "Mastra v0 Documentation",
```
PaulieScanlon: I'm wondering if this should be v0 Stable, since v1 says Beta. These are the labels we've used on the drop down switcher.

PaulieScanlon (approved):

docs/docusaurus.config.js
axe312ger: FAIR!

PaulieScanlon (approved):

PR #10023, opened 2025-11-12
feat: setup easier access to chatbot

adeniyii:
- Refactored ChatbotSidebar to use context for state management of the visibility state of the chatbot sidebar.
- Introduced AskAI button in the Navbar to toggle the chatbot sidebar using a keyboard shortcut.

## Type of Change

- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [ ] Test update

## Checklist

### DIFF
--- a/docs/src/theme/DocRoot/Layout/ChatbotSidebar/context.tsx
+++ b/docs/src/theme/DocRoot/Layout/ChatbotSidebar/context.tsx
@@ -1,0 +1,109 @@
+import {
+  createContext,
+  useCallback,
+  useContext,
+  useState,
+  type ReactNode,
+} from "react";
+
+export interface ChatbotSidebarContextValue {
+  /**
+   * Whether the chatbot sidebar is hidden
+   */
+  isHidden: boolean;
+  /**
+   * Set the hidden state of the chatbot sidebar
+   */
+  setIsHidden: (value: boolean | ((prev: boolean) => boolean)) => void;
+  /**
+   * Toggle the chatbot sidebar open/closed
+   */
+  toggle: () => void;
+  /**
+   * Open the chatbot sidebar
+   */
+  open: () => void;
+  /**
+   * Close the chatbot sidebar
+   */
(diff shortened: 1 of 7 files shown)


docs/src/theme/DocRoot/Layout/ChatbotSidebar/context.tsx
```diff
@@ -0,0 +1,109 @@
...
+  children,
+  defaultHidden = true,
+}: ChatbotSidebarProviderProps) {
+  const [isHidden, setIsHidden] = useState(defaultHidden);
+
+  console.log("isHidden", isHidden);
+
+  const toggle = useCallback(() => {
+    setIsHidden((prev) => !prev);
+  }, []);
```
adeleke5140, line 59: if there's no deps, wouldn't the toggle be recreated on every render since the array will change?

adeniyii: Hm, I don't follow. Toggle is wrapped in a callback, so it will not be recreated.
Are you referring to the dependency array of the callback?

adeleke5140: yeah im referring to the deps, since we don't have a stable reference, is the `useCallback` necessary?

adeleke5140: Never mind.

adeleke5140 (approved):
lgtm

docs/src/theme/DocRoot/Layout/ChatbotSidebar/context.tsx
adeniyii: Still don't follow. Toggle would only be recreated on every render if there was no deps array.
The callback is necessary because toggle is a dep of an effect elsewhere

CleanShot 2025-11-12 at 4  38 02@2x

PR #10122, opened 2025-11-14
fix(core): download assets

DanielSLew:
## Description

We removed the check in a previous PR that would check if there were any assets to be downloaded. Instead of running this logic for all messages always, we only run it if there's an image that needs to be handled.

Also fixes a vitest upgrade issue

## Summary by CodeRabbit

* **Bug Fixes**
  * Optimized image and file asset transformation to only process when media is present in messages, improving efficiency while preserving behavior for text-only content.

### DIFF
--- a/packages/core/src/agent/message-list/index.ts
+++ b/packages/core/src/agent/message-list/index.ts
@@ -311,34 +311,44 @@ export class MessageList {
 
         let messages = [...systemMessages, ...modelMessages];
 
-        messages = messages.map(message => {
-          if (message.role === 'user') {
-            if (typeof message.content === 'string') {
+        // Check if any messages have image/file content that needs processing
+        const hasImageOrFileContent = modelMessages.some(
+          message =>
+            message.role === 'user' &&
+            typeof message.content !== 'string' &&
+            message.content.some(part => part.type === 'image' || part.type === 'file'),
+        );
+
+        if (hasImageOrFileContent) {
+          messages = messages.map(message => {
+            if (message.role === 'user') {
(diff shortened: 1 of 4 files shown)


packages/core/src/agent/message-list/index.ts
```diff
@@ -311,34 +311,44 @@ export class MessageList {
...
-            if (typeof message.content === 'string') {
+        // Check if any messages have image/file content that needs processing
+        const hasImageOrFileContent = modelMessages.some(
+          message =>
+            message.role === 'user' &&
+            typeof message.content !== 'string' &&
+            message.content.some(part => part.type === 'image' || part.type === 'file'),
+        );
+
+        if (hasImageOrFileContent) {
```
DanielSLew, line 322: The only change is adding these lines here

packages/core/src/llm/model/router.integration.test.ts
```diff
@@ -182,34 +182,30 @@ describe('ModelRouter Integration Tests', () => {
...
-        }
-
-        expect(chunks.length).toBeGreaterThan(0);
-        const fullText = chunks.join('');
-        expect(fullText).toBeDefined();
-        expect(typeof fullText).toBe('string');
-      },
-      { timeout: 30000 },
-    );
+    it.skipIf(skipInCI)('should support streaming', { timeout: 30000 }, async () => {
```
DanielSLew, line 185: we upgraded to vitest v4 but missed this, change with how this fn receives args, it was throwing an error

CalebBarnes (approved):

PR #10203, opened 2025-11-17
fix(core): Correctly track usage

abhiaiyer91:
- Fix usage reporting
- Fix usage tracking

## Type of Change

- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [ ] Test update

## Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Stream usage now reports the most recent step's tokens, fixing incorrect cumulative tracking.

* **Improvements**
  * Provider metadata (providerMetadata/providerOptions) is propagated and preserved more consistently across message parts, UI/DB conversions, and tool-call payloads.
  * Usage objects expose additional optional token fields (cachedInputTokens, reasoningTokens).

* **Tests**
  * Added and expanded usage-tracking and provider-metadata preservation tests; adjusted snapshots and expectations.

* **Chores**
  * Added a patch changeset and bumped several dev dependency versions.


packages/core/src/agent/message-list/index.ts
```diff
@@ -1497,6 +1497,13 @@ export class MessageList {
             break;
 
           case 'tool-call':
+            // Merge part-level and message-level providerOptions
+            // Part-level takes precedence over message-level
+            const toolCallProviderMetadata = {
+              ...('providerOptions' in coreMessage && coreMessage.providerOptions ? coreMessage.providerOptions : {}),
+              ...('providerOptions' in part && part.providerOptions ? part.providerOptions : {}),
+            };
```
CalebBarnes: @TylerBarnes I think you had some thoughts on adding the providerOptions here. Maybe we only want to include the providerOptions on the individual parts instead of also adding from the top level message?

TylerBarnes: yeah @abhiaiyer91 we should retain each but separately. Any on the message should stay there, same for parts. Big reason is you can add cache breakpoints (eg for anthropic) and I think you'll start getting anthropic API errors if there are too many breakpoints - if for ex that plugin supports adding breakpoints on individual parts (not sure if it does)

DanielSLew (approved):

PR #10209, opened 2025-11-18
paul/grwth-1152-replace-ai-sdk-mastra-model-strings-1.x

PaulieScanlon:
## Description

- Fix check-deleted-pages.js
- docs: update model strings

## Type of Change

- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [ ] Test update

## Checklist

## Summary by CodeRabbit

* **Documentation**
  * Switched example model specs from dynamic provider calls to string identifiers (e.g., gpt-5.1 / gpt-5.1-realtime).
  * Replaced direct provider embedding calls with the ModelRouterEmbeddingModel pattern across RAG, memory, and embedding examples.
  * Removed now-unused provider import lines from many examples.
  * Updated public-facing examples and usage snippets to reflect the new model and embedding patterns for consistency.


docs/src/content/en/guides/guide/web-search.mdx
```diff
@@ -54,29 +54,31 @@ Create a new file `src/mastra/agents/searchAgent.ts` and define your agent:
...
+    import { google } from "@ai-sdk/google";
     import { Agent } from "@mastra/core/agent";
 
     export const searchAgent = new Agent({
         id: "search-agent",
         name: "Search Agent",
         instructions:
             "You are a search agent that can search the web for information.",
-        model: 'google/gemini-2.5-flash',
+        model: google("gemini-2.5-flash"),
```
bookercodes: nit, but prefer to use model string here

docs/src/content/en/guides/migrations/agentnetwork.mdx
```diff
@@ -61,7 +61,7 @@ const agent = new Agent({
   name: "agent-network",
   agents: { agent1, agent2 },
   tools: { tool1, tool2 },
-  model: openai("gpt-4o"),
+  model: "openai/gpt-4.1-mini",
```
bookercodes: 👋🏻 You switched several instances of gpt‑4o to the mini variant, 4.1‑mini. According to our [models overview](https://mastra.ai/models/providers/openai), 4.1‑mini doesn’t support reasoning.

A couple of concerns:
* I’m not sure 4.1‑mini will work with some features, like Agent Network  
* It might also break other features, like voice, and that would need testing

Rather than testing all of that, why not use 4.1 instead of 4.1‑mini? It has the best chance of working reliably and performing well. 

probably safe to assume users will see better results with 4.1 than 4.1‑mini, so we should encourage users to use the latest and most powerful models supported by their API key so they have the best experience with Mastra

bookercodes: i mean, we should use the GPT‑5–equivalent models

although, worth double‑checking that they’re available in the API and accessible to new accounts, since OpenAI sometimes restricts access until an account is a certain number of days old.

docs/src/content/en/docs/agents/overview.mdx
```diff
@@ -446,8 +446,8 @@ export const testAgent = new Agent({
     const userTier = requestContext.get("user-tier") as UserTier["user-tier"];
 
     return userTier === "enterprise"
-      ? openai("gpt-4o-mini")
-      : openai("gpt-4.1-nano");
+      ? "openai/gpt-4.1-mini"
+      : "openai/gpt-4.1-nano";
```
bookercodes, line 448: nit, this example would make sense if you were switching between a cheap (mini or nano) and a more expensive model (e.g. mini and 5)

docs/src/content/en/examples/rag/usage/graph-rag.mdx
```diff
@@ -129,7 +128,7 @@ Generate embeddings for the chunks and store them in the vector database:
 
 ```typescript copy showLineNumbers{56} title="index.ts"
 const { embeddings } = await embedMany({
-  model: openai.embedding("text-embedding-3-small"),
+  model: "openai/text-embedding-3-small",
```
bookercodes: I _think_ you need to wrap the model string in ⁠ModelRouterEmbeddingModel, as shown here: https://mastra.ai/models/embeddings#quick-start

that sounds annoying to update in all these files - you can probably skip ⁠examples/rag/* since we’re overhauling that later, but it’s worth double‑checking the docs. In other words, it's okay not to update examples/rag/* in this PR

docs/src/content/en/guides/guide/web-search.mdx
bookercodes: just so happens we’re importing the provider to access the search tool, but there’s no need to pass the model provider instead of a model string.

docs/src/content/en/guides/migrations/agentnetwork.mdx
PaulieScanlon: Good catch old sport. I did some research, 4.1-mini is safer for a couple of reasons. 

- Lower cost
- Lower access barriers
- Good sweet spot between perf and compat 

... and mini's are apparently considered more stable.

It would be cool to use 5 models, but there are some restrictions on usage, some relating to regions. @TylerBarnes do you have any suggestions here?

PaulieScanlon: And gpt-4o doesn't support reasoning. Not sure if that was ever a problem, was it?

bookercodes (approved):

PR #10233, opened 2025-11-18
fix: preserve providerMetadata for Gemini 3 thought_signature support

TylerBarnes:
Fixes Gemini 3 models throwing `AI_APICallError: Function call is missing a thought_signature in functionCall parts` when using tools in multi-turn conversations.

**The Problem**

Gemini 3 models (like `gemini-2.0-flash-thinking-exp-01-21`) require a `thoughtSignature` field to be preserved across multi-turn function calling conversations. When the model makes a tool call, it includes a `thoughtSignature` in the response that must be sent back in subsequent turns to maintain context. Without it, the API rejects the request.

The AI SDK Google provider correctly captures `thoughtSignature` from the model's response and stores it in `providerMetadata.google.thoughtSignature`. However, Mastra was losing this metadata during message storage and retrieval, causing the error on the second turn when the assistant's tool call message was sent back to the API.

**The Root Cause**

Tool call messages were being created with AI SDK v4 format (`args` property) instead of v5 format (`input` property). This caused Mastra's message classification logic to misidentify them as v4 messages, which took a different code path that didn't preserve `providerMetadata`. When these messages were loaded from the database and converted back to model messages, the `thoughtSignature` was missing.

**The Solution**

Changed tool call message creation to use AI SDK v5 format (`input` instead of `args`) so messages are correctly classified as v5. This ensures `providerMetadata` is preserved when storing tool calls to the message list and carried forward as `providerOptions` when converting messages back to model format for LLM execution.

Also updated AI SDK packages to latest versions which include improvements to Gemini reasoning model support.

**Testing**

Tested with `gemini-2.0-flash-thinking-exp-01-21` model using tools and `thinkingConfig` in the playground. Multi-turn function calling conversations now work correctly without `AI_APICallError`.


avaer:
Thanks! We need this.

Without support for this gemini-3 models are unusable, since the first memory tool will effectively crash the whole conversation.

PR #10270, opened 2025-11-19
fix(client-js): added agentId to listMessages API call

bala-intron:
## Description

Fixed listMessages API call to include agentId.

## Type of Change

## Summary by CodeRabbit

* **Bug Fixes**
  * Ensures the message listing API always includes the agentId parameter, fixing inconsistent requests.
  * Stabilizes query parameter ordering so filters and pagination are applied reliably across calls.


NikAiyer (approved):
LGTM! Can you add a changeset?

PR #10282, opened 2025-11-19
Fix generate toolResults and mismatch in provider tool names

abhiaiyer91:
Problem 1: Missing tool-result handling in doGenerate
Issue: After the recent commit that added doGenerate support for v2 models, the wrapper in model.ts was converting the generate response to a stream but wasn't handling tool-result chunks. This meant provider-executed tools worked in stream() but toolResults was empty in generate().

Problem 2: Inconsistent tool names between doStream and doGenerate
Issue: AI SDK v5 has an inconsistency in how it handles provider-defined tool names:
doStream: Returns tool name as just the suffix part (e.g., "web_search" from ID "openai.web_search")
doGenerate: Returns the name field we provided (e.g., "search")
This meant the same tool had different names depending on whether you used stream() or generate().

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved streaming: handle tool-result items, normalize provider tool names for matching, and distinguish URL vs. document sources so outputs show correct source metadata.

* **Tests**
  * Expanded streaming and generate tests to validate tool calls/results, sources, structured outputs, finish reasons, and richer mock generate responses.

* **Chores**
  * Added a changeset describing this patch release.

### DIFF
--- a/packages/core/src/stream/aisdk/v5/compat/prepare-tools.ts
+++ b/packages/core/src/stream/aisdk/v5/compat/prepare-tools.ts
@@ -62,7 +62,7 @@ export function prepareToolsAndToolChoice<TOOLS extends Record<string, Tool>>({
                 inputSchema: asSchema(sdkTool.inputSchema).jsonSchema,
                 providerOptions: sdkTool.providerOptions,
               };
-            case 'provider-defined':
+            case 'provider-defined': {
               // For provider-defined tools, extract the name from the ID to match doStream behavior
               // ID format is typically "provider.toolName" (e.g. "openai.web_search")
               // doStream returns just "toolName" part, so we use that as name for consistency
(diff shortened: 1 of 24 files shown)


wardpeet (approved):

packages/core/src/stream/aisdk/v5/compat/prepare-tools.ts
```diff
@@ -63,11 +63,20 @@ export function prepareToolsAndToolChoice<TOOLS extends Record<string, Tool>>({
...
               };
             case 'provider-defined':
+              // For provider-defined tools, extract the name from the ID to match doStream behavior
+              // ID format is typically "provider.toolName" (e.g. "openai.web_search")
+              // doStream returns just "toolName" part, so we use that as name for consistency
+              const providerId = (sdkTool as any).id;
+              const providerToolName =
+                providerId && providerId.includes('.')
+                  ? providerId.split('.').slice(1).join('.') // Take everything after first dot
+                  : name; // Fallback to our key if no dot in ID
```
TylerBarnes: if there's no `.` should we fallback to providerId if it exists?

abhiaiyer91: Yeah i'll do that

PR #10366, opened 2025-11-20
fix(core, schema-compat): nested ZodNullable and structuredOutput opts for schema compat

### LINKED ISSUE #10006: Invalid schema for response_format 'response'
This issue was created from Discord post 1437803077162762416:

[Open in Discord](https://discord.com/channels/1309558646228779139/1437803077162762416)

Claude Code is constantly generating workflows that result in this error. 

Any TLDR on what I need to do to make sure things work for openai? Something to do with strict mode and mastra .24?

ERROR [2025-11-11 13:54:43.411 +0000] (Mastra): Error in agent stream
    runId: "e8d974d7-75e9-41f9-8356-e95df0ab6c35"
    error: {
      "message": "Invalid schema for response_format 'response': In context=(), 'required' is required to be supplied and to be an array including every key in properties. Missing 'keyConnections'.",
      "name": "AI_APICallError",

DanielSLew:
Fixes https://github.com/mastra-ai/mastra/issues/10006

Some model providers have specific schema requirements. For example, OpenAI's strict mode requires all properties to be in the `required` array. Use `processSchema` to transform your Zod schema before passing it to the agent, ensuring the same schema works for both backend generation and frontend validation.

```typescript showLineNumbers copy
import { z } from "zod";
import { processSchema } from "@mastra/core";

// Define your schema once
const userSchema = z.object({
  name: z.string(),
  age: z.number().optional(), // Would cause issues with OpenAI strict mode
});

// Transform it for OpenAI
const openaiSchema = processSchema.openai(userSchema);

// Use in agent
const response = await testAgent.generate("Extract user information", {
  structuredOutput: {
    schema: openaiSchema,
  },
});

console.log(response.object);

// Use the same transformed schema on frontend for validation
const validatedData = openaiSchema.parse(response.object);
```

The `processSchema` utility converts `.optional()` fields to `.nullable()` for OpenAI models, ensuring all fields are in the required array but can accept null values. Other providers have their own transformations: `processSchema.anthropic()`, `processSchema.google()`, `processSchema.deepseek()`, `processSchema.meta()`, and `processSchema.openaiReasoning()` for reasoning models.

## Summary by CodeRabbit

* **New Features**
  * Automatic Zod schema transformation for OpenAI strict-mode compatibility in agents with structured output; exposes a ModelInformation type.

* **Bug Fixes**
  * Optional fields converted to nullable with a transform that maps null → undefined; existing nullable fields preserved; applies deeply across nested schemas and arrays.

* **Tests**
  * Expanded suites covering transformations, nesting, serialization, reasoning-model behavior, and agent integration.

* **Documentation**
  * Added changeset entry documenting the patch release.


CalebBarnes (approved):

Vulcanostrol:
@DanielSLew Would this change mean you would have to do `const openaiSchema = processSchema.openai(userSchema);` before every agent.generate in our codebases? Why wouldn't mastra (in the agent.generate function) just notice the agent's provider is openai, and do the compatibility transformation internally? The feature as it is right now would still have a lot of new Mastra users running into these API call errors I think.

Vulcanostrol:
> @Vulcanostrol nice suggestion! I just tried it out and it works great, helps me scope down this PR also to only handle openai issues

awesome! no worries, I think this is definitely cleaner for users of mastra as well 👍

PR #10370, opened 2025-11-20
fix: ci to run when its supposed to

DanielSLew:
Lots of tests haven't been running when they were supposed to.

I did an audit of all the packages and their dependencies and when their tests run. There were a lot of missing cases where tests should be running

Now:
- Every package that has a mastra fixture will run ALWAYS, this is because it tests multiple mastra packages and it's easier to do that than to whitelist 8 different packages and then also have to remember to go in and update the CI workflows everytime we add a new package to the integration tests
- Every package that has a mastra dependency but not integration tests, will run when that dependency changes.

## Summary by CodeRabbit

* **Chores**
  * Updated test automation trigger conditions across multiple workflows to improve test coverage and execution efficiency by adjusting package-level path filtering patterns.

### DIFF
--- a/.github/workflows/secrets.test-agent-builder.yml
+++ b/.github/workflows/secrets.test-agent-builder.yml
@@ -40,8 +40,8 @@ jobs:
           predicate-quantifier: 'every'
           filters: |
             agent-builder:
-              - 'packages/agent-builder/**'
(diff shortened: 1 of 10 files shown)


wardpeet (approved):
Worth a try!

PR #10392, opened 2025-11-21
fix(core): resolve netlify gateway double-prefix in model docs generation

DanielSLew:
Fixes the CI error where the model docs generation script was failing with `ENOENT: no such file or directory, open 'docs/src/content/en/models/providers/netlify/netlify.mdx'`.

The issue was that `NetlifyGateway` had both a `prefix = 'netlify'` and returned a provider with id `netlify`, causing the registry generator to create `netlify/netlify` instead of just `netlify`. This made the docs script treat it as a regular provider instead of a gateway.

Since we both have prefix and id, I removed prefix in favour of using `id` as the prefix/id. `models.dev` gets special treatment now as it's not really a gateway, it's just an aggregator. This makes it easier to reason when writing your own gateways now as well. Models are now correctly accessed as `netlify/anthropic/claude-...` instead of `netlify/netlify/anthropic/claude-...`.

Also preserved the "Custom Gateways" documentation link in the auto-generated sidebars and gateway index page so it doesn't get removed on future regenerations.

## Summary by CodeRabbit

* **New Features**
  * Added Poe and Cohere providers plus many new model variants (Gemini, Grok, Opencode, etc.); new provider options for Google, OpenAI, and xAI.

* **Bug Fixes**
  * Fixed provider registry generation.

* **Documentation**
  * Updated provider docs, model counts, provider grid, examples, and model lists; removed legacy alpha/beta entries.

* **Refactor**
  * Gateway ID now used as the model-ID prefix.

* **Tests**
  * Updated tests/fixtures to reflect naming and ID changes.

* **Chores**
  * Added release changeset for a patch update.


packages/core/src/llm/model/gateways/netlify.ts
```diff
@@ -37,7 +37,7 @@ interface TokenData {
 export class NetlifyGateway extends MastraModelGateway {
   readonly id = 'netlify';
   readonly name = 'Netlify AI Gateway';
-  readonly prefix = 'netlify'; // All providers will be prefixed with "netlify/"
+  readonly prefix = undefined; // No prefix - netlify is the gateway itself
```
TylerBarnes: I don't think this is the right fix, we want all gateways to be required to set a prefix (and we should probably enforce that it's not already used by another gateway)

DanielSLew: isn't that what the id is for? models.dev also has an undefined prefix

TylerBarnes: models.dev is undefined cause it's an aggregator, maybe we need a `readonly isGateway`?
basically gateways should only return a single provider where the provider is the same as their prefix. We could make id/prefix the same thing but we should fully take out prefix in that case. The idea was "prefix" meant gateway as in all models are prefixed with this string

DanielSLew: hmm yeah, if it has an id, I kind of like the idea of removing `prefix` to make it easier to reason about especially now that we've made it so you can add your own gateways

abhiaiyer91 (approved):

PR #10420, opened 2025-11-22
Fix/nested embedding path

### LINKED ISSUE #6563: [DISCORD:1401227347793018991] MongoDB embedding field
This issue was created from Discord post: https://discord.com/channels/1309558646228779139/1401227347793018991

Using MongoDBVector, is there any way to specify the path of the field that hold the embeddings?

I have seen that the default field is named "`embedding`" but what if I want to save the embedding inside an inner field. For example in a collection `emails`:

`text:{
  content: "",
  contentEmbedding: [],
}`

In this case field will be from collection `emails` > path `text.contentEmbedding`.

Is the recommended way to handle this situations to just use MongoDB tools instead of Mastra abstractions?

intojhanurag:
## Description

Added support for upserting and querying vectors stored at nested paths in `MongoDBVector`. Previously, only flat embedding fields were supported, which caused issues when using deeply nested embeddings. This update allows specifying a custom `embeddingPath` for both `upsert` and `query` operations. The change also updates tests to ensure nested embedding functionality works correctly.

## Related Issue(s)
Closed: #6563

## Type of Change

- [x] Test update

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Support for nested embedding paths when storing, updating, and querying vectors — embeddings and metadata can be placed under nested object fields.
  * Embedding field name can be customized.

* **Tests**
  * Added tests validating nested embedding-path upsert and query behavior.
  * Identical duplicate test suite was also added (cleanup recommended).

### DIFF
--- a/stores/mongodb/src/vector/index.test.ts
+++ b/stores/mongodb/src/vector/index.test.ts
@@ -481,6 +481,41 @@ describe('MongoDBVector Integration Tests', () => {
     });
   });
 
+  describe('Nested embedding path support', () => {
+    const indexName = 'nested_embedding_test_' + Date.now();
+
+    beforeAll(async () => {
+      await createIndexAndWait(vectorDB, indexName, 4, 'cosine');
+    });
+
+    afterAll(async () => {
+      await deleteIndexAndWait(vectorDB, indexName);
+    });
+
+    it('should upsert and query vectors stored at nested path', async () => {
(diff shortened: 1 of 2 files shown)


intojhanurag:
Hey  @abhiaiyer91 , Can you review this pr ? If you have a moment.

abhiaiyer91:
This looks good! just a conflict to address!

PR #10488, opened 2025-11-25
message metadata

### LINKED ISSUE #8556: [FEATURE] Allow adding metadata to messages
### Problem Statement

Users would like to have the ability to add metadata to messages for later retrieval.
While the `stream/generate` makes you believe you can pass metadata, these don't get actually save to the database:

When calling stream this way
```ts
const response = await agent.stream([
      {
        role: 'user',
        content: prompt,
        metadata: { userId: 'user-1234' },
      },
    ], {
      memory: {
        resource: 'organization-xyz',
        thread: 'weather-activity-planning',
      }
    });
```

Retrieving the message afterward will look like this:
```ts
[
  {
    id: '8414e145-cc5e-43a4-a9e6-7c5b2836c7e3',
    role: 'user',
    metadata: { createdAt: 2025-10-06T18:41:27.127Z },
    parts: [ { type: 'text', text: 'Test message' } ]
  }
]
```
Where `metadata` seems to be automatically populated with `createdAt` but, the initial metadata are not there (and they are not in DB either)

### Proposed Solution

.

### Component

Memory

### Alternatives Considered

_No response_

### Example Use Case

.

### Additional Context

_No response_

### Verification

abhiaiyer91:
Fix message metadata not persisting when using simple message format. Previously, custom metadata passed in messages (e.g., `{role: 'user', content: 'text', metadata: {userId: '123'}}`) was not being saved to the database. This occurred because the CoreMessage conversion path didn't preserve metadata fields.

Now metadata is properly preserved for all message input formats:
- Simple CoreMessage format: `{role, content, metadata}`
- Full UIMessage format: `{role, content, parts, metadata}`
- AI SDK v5 ModelMessage format with metadata

Fixes #8556

## Summary by CodeRabbit

* **Bug Fixes**
  * Message metadata is now reliably persisted and preserved across all message format conversions and SDK versions.

* **Tests**
  * Added coverage for string-content messages preserving metadata.
  * Updated test fixtures and URL expectations to a new stable image source.

* **Documentation**
  * Added a changelog entry noting the metadata persistence fix.

### DIFF
--- a/packages/core/src/agent/__tests__/uimessage.test.ts
+++ b/packages/core/src/agent/__tests__/uimessage.test.ts
@@ -298,7 +298,7 @@ function uiMessageTest(version: 'v1' | 'v2') {
 
     it('should handle content as string', async () => {
       const agent = new Agent({
-        name: 'mixed-metadata-agent',
+        name: 'simple-content-agent',
         instructions: 'You are a helpful assistant',
         model: dummyModel,
         memory: mockMemory,
@@ -317,9 +317,9 @@ function uiMessageTest(version: 'v1' | 'v2') {
           ],
           {
             memory: {
-              resource: 'mixed-user',
+              resource: 'simple-content-user',
               thread: {
-                id: 'mixed-thread',
+                id: 'simple-content-thread',
               },
             },
           },
@@ -337,9 +337,9 @@ function uiMessageTest(version: 'v1' | 'v2') {
           ],
           {
             memory: {
-              resource: 'mixed-user',
+              resource: 'simple-content-user',
               thread: {
-                id: 'mixed-thread',
+                id: 'simple-content-thread',
               },
             },
           },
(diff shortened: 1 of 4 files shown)


packages/core/src/agent/__tests__/uimessage.test.ts
```diff
@@ -295,6 +295,65 @@ function uiMessageTest(version: 'v1' | 'v2') {
...
+        );
+      }
+
+      const result = await mockMemory.recall({
+        threadId: 'mixed-thread',
+        resourceId: 'mixed-user',
+        perPage: 10,
+      });
+
+      expect(result?.messages.length).toBeGreaterThan(0);
```
DanielSLew, line 355: I think for this test we want to fetch the actual message and assert that the metadata is on it

abhiaiyer91: .recall fetches the message, ill check for metadata

packages/core/src/agent/message-list/index.ts
```diff
@@ -1707,6 +1707,11 @@ export class MessageList {
       content.providerMetadata = coreMessage.providerOptions;
     }
 
+    // Preserve metadata field if present (matches aiV4UIMessageToMastraDBMessage behavior)
+    if ('metadata' in coreMessage && coreMessage.metadata !== null && coreMessage.metadata !== undefined) {
```
wardpeet, line 1711:
```suggestion
    if ('metadata' in coreMessage && !coreMessage.metadata) {
```

packages/core/src/agent/message-list/index.ts
```diff
@@ -2818,8 +2823,11 @@ export class MessageList {
       .map(p => p.text)
       .join('\n');
 
-    // Store original content in metadata for round-trip
-    const metadata: Record<string, unknown> = {};
+    // Preserve metadata from the input message if present
+    const metadata: Record<string, unknown> =
+      'metadata' in modelMsg && modelMsg.metadata !== null && modelMsg.metadata !== undefined
```
wardpeet, line 2828:
```suggestion
      'metadata' in modelMsg && !modelMsg.metadata
```

PR #10543, opened 2025-11-26
fix(memory): recall() with lastMessages returns newest messages, not oldest

m0rdras:
## Summary

When using `lastMessages: N` config without an explicit `orderBy`, the `recall()` function was returning the **OLDEST** N messages instead of the **NEWEST** N messages. This completely breaks conversation history for any thread that grows beyond the lastMessages limit.

### Root Cause

All storage adapters default to ASC ordering for `getMessages`/`listMessages`. When `recall()` doesn't specify `orderBy`, it uses ASC, so `LIMIT 3` on a 10-message thread returns rows 1,2,3 (oldest) instead of the last 3.

### Fix

When `perPage` is set and no explicit `orderBy` is provided:
1. Query with DESC to get the newest N messages
2. Reverse the results to restore chronological order for the LLM

### Backward Compatibility

- Explicit `orderBy` still works as before
- `perPage: false` (get all) still works as before  
- Only the implicit "get last N" case is fixed

## Test Plan

- [x] Added integration test that verifies recall returns newest messages
- [x] Test creates 10 messages, recalls with `lastMessages: 3`, expects messages 8,9,10 (not 1,2,3)
- [x] Verified messages are returned in chronological order (oldest to newest within the window)

## Summary by CodeRabbit

* **Bug Fixes**
  * Recall now returns the most recent messages when a message limit is used without explicit ordering, preserving correct conversation history for long threads.

* **Tests**
  * Added integration tests across all storage backends to verify the corrected message ordering behavior.


NikAiyer (commented):
hey @m0rdras thanks for making this change! Could you also add the same test to with-libsql-storage.test.ts and with-upstash-storage.test.ts? Also a changeset (pnpm changeset). Then I think this should be good to merge!

NikAiyer (approved):
LGTM! Can you fix that lint issue?

Issue #10517, opened 2025-11-26
[FEATURE] 401 errors should not trigger SSE fallback


yamadashy:
### Problem Statement

When Streamable HTTP fails, the client falls back to SSE for all errors including 401.

The MCP spec has been clarified:
- Old (2025-06-18): "HTTP 4xx status code (e.g., 405 or 404)" 
- New (2025-11-25): "400 Bad Request, 404 Not Found or 405 Method Not Allowed"

Spec: https://modelcontextprotocol.io/specification/2025-11-25/basic/transports

Current implementation triggers SSE fallback for all errors:
https://github.com/mastra-ai/mastra/blob/main/packages/mcp/src/client/client.ts#L348-L386

### Proposed Solution

Only trigger SSE fallback for 400, 404, 405.
Other errors (401, 403, 500, etc.) should be thrown directly.

### Component

Other

### Alternatives Considered

Patching @mastra/mcp locally to throw 401 directly.

### Example Use Case

Catching 401 to trigger re-authentication flow in OAuth-protected MCP servers.

Related: #7497

### Verification

- [x] I have searched the existing issues to make sure this is not a duplicate
- [x] I have provided sufficient context for the team to understand the request

DanielSLew:
@yamadashy thanks for reporting this, we'll get around to doing this one but it's a lower priority for us. We'd love a PR if you have time as it seems like you already know the fix 😄

intojhanurag:
Hey @yamadashy , I fixed this issue . You can review , if you have a minute.

abhiaiyer91:
Looks like @yamadashy got their PR into upstream!

PR #10550, opened 2025-11-26
fix(mcp): restrict SSE fallback to 400/404/405 instead of all errors

intojhanurag:
## Description

Restrict SSE fallback in the MCP client to only occur for HTTP status codes 400, 404, and 405.  
Previously, any error, including 401 Unauthorized triggered a fallback, which could break OAuth re-authentication flows.  
This update ensures that 401, 403, 500, and other errors are thrown immediately, following the updated MCP specification.

## Related Issue(s)

#10517

## Type of Change

## Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * SSE fallback for streaming connections now limited to HTTP statuses 400, 404, and 405 to avoid inappropriate fallbacks.
  * Connection errors for other HTTP statuses are now surfaced immediately, improving clarity.
  * Improved error handling and reporting around failed connection attempts for faster diagnosis.


packages/mcp/src/client/client.ts
```diff
@@ -364,7 +364,22 @@ export class InternalMastraMCPClient extends MastraBase {
...
         this.log('debug', `Streamable HTTP transport failed: ${error}`);
-        shouldTrySSE = true;
+        
+        const status =
+          error?.response?.status ??
+          error?.status ??
+          error?.cause?.statusCode ??
+          null;
+
+        const allowedFallback = [400, 404, 405];
```
abhiaiyer91, line 375: Are these the right codes?

abhiaiyer91 (approved):

packages/mcp/src/client/client.ts
abhiaiyer91: a comment with the link to the spec would be good!

abhiaiyer91:
@intojhanurag you have CI failures

abhiaiyer91:
@intojhanurag we need to add tests to back up or prove your change works

intojhanurag:
> @intojhanurag we need to add tests to back up or prove your change works

@abhiaiyer91 , Yeah but we need to wait for the resolving the upstream issue.
Till then we can keep it open.

PR #10601, opened 2025-11-27
fix(memory): guard against non-object content in updateMessageToHideWorkingMemoryV2

m0rdras:
## Summary
- The function assumed `message.content` was always a V2 object (`{ format: 2, parts: [...] }`)
- Legacy or malformed messages could have string, array, or undefined content
- Spreading non-object content corrupted the message structure: `{ ...undefined }` → `{}`, `{ ..."string" }` → `{}`
- This caused crashes when subsequently accessing `.parts` or `.content`

## Changes
- Only spread content if it's a proper object (not array/null/undefined)
- Use optional chaining for `content?.content` and `content?.parts`
- Safe access for part properties with `part?.type`, `part?.toolInvocation?.toolName`, `part?.text`

## Test plan
- [x] Build passes (`pnpm build:memory`)
- [x] Unit tests added covering edge cases (non-object content, null parts, etc.)


NikAiyer:
hey @m0rdras thanks for making this PR! Can you fix the lint issues? After that it should be good to merge

NikAiyer (approved):

PR #10605, opened 2025-11-27
fix(0.x@core): Fix corrupt provider-registry.json in globalCach

DanielSLew:
## Description

Previous versions before this was released https://github.com/mastra-ai/mastra/pull/10455 some people saw corrupted provider-registry.json files. these files get saved to a global cache so we don't need to always regenerate it. Because of this if the global cache is corrupted the user gets stuck in this state where they have to manually delete their global cache, and it's not clear how to do that.

This fix will check if the global cache is corrupted, and if it is it will delete it so it can regenerate it.


packages/core/src/llm/model/provider-registry.test.ts
```diff
@@ -626,6 +626,149 @@ describe('GatewayRegistry Auto-Refresh', () => {
   });
 });
 
+describe('Corrupted JSON recovery', () => {
+  const originalReadFileSync = fs.readFileSync.bind(fs);
+  const originalWriteFileSync = fs.writeFileSync.bind(fs);
+  const originalExistsSync = fs.existsSync.bind(fs);
+  const originalMkdirSync = fs.mkdirSync.bind(fs);
+  const originalRmSync = fs.rmSync.bind(fs);
+  const originalUnlinkSync = fs.unlinkSync.bind(fs);
```
DanielSLew:
```suggestion
```

packages/core/src/llm/model/provider-registry.test.ts
```diff
@@ -626,6 +626,148 @@ describe('GatewayRegistry Auto-Refresh', () => {
...
+    // Mock fs.existsSync to return true for our test path
+    const existsSyncSpy = vi.spyOn(fs, 'existsSync').mockImplementation(filePath => {
+      if (typeof filePath === 'string' && filePath === corruptedJsonPath) {
+        return true;
+      }
+      return originalExistsSync(filePath);
+    });
+
+    // Mock fs.unlinkSync to track deletion
+    let deletedPath: string | null = null;
```
DanielSLew:
```suggestion
```

packages/core/src/llm/model/provider-registry.test.ts
```diff
@@ -626,6 +626,148 @@ describe('GatewayRegistry Auto-Refresh', () => {
...
+  "models": { "test": ["a", "b"] },
+  "version": "1.0.0"
+}
+}
+]}
+}`; // <-- extra garbage from concurrent write
+    originalWriteFileSync(corruptedJsonPath, corruptedContent, 'utf-8');
+
+    // Mock getPackageRoot to return our temp directory
+    const packageRootMock = vi.fn().mockReturnValue(tempDir);
```
DanielSLew:
```suggestion
```

abhiaiyer91 (approved):

PR #10606, opened 2025-11-27
fix(core): corrupted global cache provider registry

DanielSLew:
## Description

Previous versions before this was released https://github.com/mastra-ai/mastra/pull/10455 some people saw corrupted provider-registry.json files. these files get saved to a global cache so we don't need to always regenerate it. Because of this if the global cache is corrupted the user gets stuck in this state where they have to manually delete their global cache, and it's not clear how to do that.

This fix will check if the global cache is corrupted, and if it is it will delete it so it can regenerate it.

## Summary by CodeRabbit

* **Bug Fixes**
  * Detects and removes corrupted provider registry files in the global cache, falling back to the bundled registry and issuing warnings instead of hard failures.

* **Tests**
  * Added extensive tests for corrupted-registry recovery, deletion behavior, and concurrent write scenarios to ensure robustness.

* **Chores**
  * Added a changeset entry for a patch release.

* **Documentation**
  * Minor wording/formatting cleanup in example content.

* **Style**
  * Cosmetic workflow YAML string quoting consistency.


packages/core/src/llm/model/provider-registry.test.ts
```diff
@@ -626,6 +626,145 @@ describe('GatewayRegistry Auto-Refresh', () => {
...
+    // Mock fs.existsSync to return true for our test path
+    const existsSyncSpy = vi.spyOn(fs, 'existsSync').mockImplementation(filePath => {
+      if (typeof filePath === 'string' && filePath === corruptedJsonPath) {
+        return true;
+      }
+      return originalExistsSync(filePath);
+    });
+
+    // Mock fs.unlinkSync to track deletion
+    let deletedPath: string | null = null;
```
DanielSLew:
```suggestion
```

abhiaiyer91 (approved):

PR #10657, opened 2025-11-28
feat: add reserved keys in RequestContext for secure resourceId/threadId setting

### LINKED ISSUE #4296: [FEATURE] Cleaner way to set `resourceId` from `RuntimeContext`
### Problem Statement

I'm using the Mastra server as-is (not using Mastra as a library inside another web server).
I've an authentication middleware that pokes some user data into the `RuntimeContext`.
Now I'd like to set the `resourceId` based on that user data.

### Proposed Solution

Maybe an optional `getResourceId?: (options: {runtimeContext: RuntimeContext}) => string | undefined` callback could be added to `MastraConfig`. The callback would be called after the middlewares have been run, to get chance derive a `resourceId` from the runtime context.

### Component

Agents

### Alternatives Considered

In Discord, @abhiaiyer91 [suggested](https://discord.com/channels/1309558646228779139/1357964923522650285/1358334943305928714) to use a middleware to change the `req.body` of the request.

This feels a little dirty and wonky to me: the middleware now needs to figure out for what routes to set `resourceId` on the `req.body` or skip doing that.

### Example Use Case

As a developer, I want to be able to set `resourceId` based on `runtimeContext` in a clean way.

### Additional Context

_No response_

### Verification

abhiaiyer91:
## Summary

This PR adds support for reserved keys in `RequestContext` that allow middleware to securely set `resourceId` and `threadId` for memory operations.

## Problem

When using Mastra server with authentication middleware, users want to set `resourceId` based on authenticated user data rather than trusting client-provided values. Currently, `resourceId` and `threadId` come from the request body, which could allow attackers to hijack another user's memory by sending a different resourceId.

## Solution

Introduce two reserved keys in `RequestContext`:
- `MASTRA_RESOURCE_ID_KEY` (`'mastra__resourceId'`)
- `MASTRA_THREAD_ID_KEY` (`'mastra__threadId'`)

When set by middleware, these values **take precedence** over any values passed in the request body, ensuring security.

## Usage Example

```typescript
import { MASTRA_RESOURCE_ID_KEY, MASTRA_THREAD_ID_KEY } from '@mastra/core';

  server: {
    middleware: [
      async (c, next) => {
        const user = await authenticateUser(c);
        const requestContext = c.get('requestContext');
        
        // Securely set resourceId/threadId from authenticated user
        requestContext.set(MASTRA_RESOURCE_ID_KEY, user.id);
        requestContext.set(MASTRA_THREAD_ID_KEY, user.preferredThread);
        
        await next();
      },
    ],
  },
});
```

## Changes

- `packages/core/src/request-context/index.ts`: Added `MASTRA_RESOURCE_ID_KEY` and `MASTRA_THREAD_ID_KEY` constants
- `packages/core/src/agent/agent.ts`: Modified `#execute` to check for reserved keys first
- `packages/core/src/agent/agent-legacy.ts`: Same changes for v1 API
- `packages/core/src/di/index.ts`: Export the reserved key constants
- Added comprehensive tests for both v1 and v2 APIs

## Tests

Added 9 tests covering:
- Using reserved keys from RequestContext
- Reserved keys taking precedence over body values (security)
- Reserved keys taking precedence over memory option values
- Both `generate` and `stream` methods
- Both v1 (legacy) and v2 APIs

Closes #4296

## Summary by CodeRabbit

* **New Features**
  * RequestContext supports reserved keys for thread and resource identity that take precedence over client-supplied values across standard and legacy execution paths to protect memory targeting.

* **Tests**
  * Comprehensive unit tests covering reserved-key precedence and propagation across generate/stream and legacy variants.

* **Documentation**
  * Added changelog entry describing the reserved-keys behavior and security precedence.

### DIFF
--- a/.changeset/clever-keys-guard.md
+++ b/.changeset/clever-keys-guard.md
@@ -1,0 +1,8 @@
+---
+'@mastra/core': minor
+---
+
+Add reserved keys in RequestContext for secure resourceId/threadId setting from middleware
+
+This allows middleware to securely set `resourceId` and `threadId` via reserved keys in RequestContext (`MASTRA_RESOURCE_ID_KEY` and `MASTRA_THREAD_ID_KEY`), which take precedence over client-provided values for security.
+
--- a/packages/core/src/agent/__tests__/request-context-reserved-keys.test.ts
+++ b/packages/core/src/agent/__tests__/request-context-reserved-keys.test.ts
@@ -1,0 +1,388 @@
+/**
+ * Tests for RequestContext reserved keys feature
+ *
+ * This feature allows middleware to securely set resourceId and threadId
+ * via reserved keys in RequestContext, which take precedence over
+ * client-provided values for security.
+ *
+ * Reserved keys:
+ * - mastra__resourceId: Sets the resourceId for memory operations
+ * - mastra__threadId: Sets the threadId for memory operations
+ *
+ * @see https://github.com/mastra-ai/mastra/issues/4296
+ */
+import { simulateReadableStream, MockLanguageModelV1 } from '@internal/ai-sdk-v4';
+import { convertArrayToReadableStream, MockLanguageModelV2 } from 'ai-v5/test';
+import { beforeEach, describe, expect, it } from 'vitest';
+
+import { MockMemory } from '../../memory/mock';
+import { RequestContext, MASTRA_RESOURCE_ID_KEY, MASTRA_THREAD_ID_KEY } from '../../request-context';
+import { Agent } from '../index';
+
+describe('RequestContext reserved keys for resourceId and threadId', () => {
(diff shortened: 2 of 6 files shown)


NikAiyer (approved):

wardpeet (commented):
Instead of accepting these values directly from the request, I'd recommend adding custom properties to the Hono context (`c`) specifically for this purpose:
- `c.setRequestId(id: string)`
- `c.setThreadId(id: string)`

These methods should be used in middleware to set values on the `requestContext` using `Symbol` keys. This approach ensures that:

1. The properties can only be set through controlled middleware
2. The Symbol keys prevent external code from accessing or modifying these values
3. The implementation remains internal to `@mastra/core`

This way, we maintain a clear security boundary

PR #10735, opened 2025-12-01
fix(core): return processed text in result.text when using output processors

### LINKED ISSUE #9916: Textual response with OutputProcessors applied?
This issue was created from Discord post 1437347410144919634:

[Open in Discord](https://discord.com/channels/1309558646228779139/1437347410144919634)

Silly question... I have written some custom output processors that modify the original textual output of the Agent, following the same pattern as the Mastra native ones(replacing content.content and content.parts). However, when reading `result.text` I still get the raw original response.
This is how I'm obtaining the response:
```
      const result = await mastraAgent.generate(inputText);
      const message = result.tripwire ? result.tripwireReason : result.text;
```

I see the processed parts in the `result.response` data structure, that I could use to build what I need (e.g.: result.response.messages, result.response.uiMessages), but I'm wondering what is the proper way to go about just getting a textual response with processors applied? 
Thanks!

DanielSLew:
When using output processors with `agent.generate()`, `result.text` was returning the unprocessed LLM response instead of the processed text.

**Before:**
```ts
const result = await agent.generate('hello');
result.text // "hello world" (unprocessed)
result.response.messages[0].content[0].text // "HELLO WORLD" (processed)
```

**After:**
```ts
const result = await agent.generate('hello');
result.text // "HELLO WORLD" (processed)
result.object // also now reflects processed values for structured output
```

The bug was caused by the `text` delayed promise being resolved twice - first correctly with the processed text, then overwritten with the unprocessed buffered text. Also moved `object` resolution to the finish step so structured output can also be processed.

Fixes #9916

## Summary by CodeRabbit

* **Bug Fixes**
  * result.text now reliably returns the final processed text and structured outputs reflect processor-modified values by deferring object resolution until processing completes.

* **Refactor**
  * Consolidated output validation into a single, consistent validation pathway for schema-backed outputs.

* **Tests**
  * Added tests verifying processor transformations for both plain text and structured (JSON-like) outputs.

* **Changelog**
  * Added a patch entry documenting the output processing fix.

### DIFF
--- a/packages/core/src/stream/base/output.ts
+++ b/packages/core/src/stream/base/output.ts
@@ -18,7 +18,11 @@ import type {
   MastraModelOutputOptions,
   MastraOnFinishCallbackArgs,
 } from '../types';
-import { createJsonTextStreamTransformer, createObjectStreamTransformer } from './output-format-handlers';
+import {
+  createJsonTextStreamTransformer,
+  createObjectStreamTransformer,
+  validateOutputValue,
+} from './output-format-handlers';
 import { getTransformedSchema } from './schema';
 import type { InferSchemaOutput, OutputSchema, PartialSchemaOutput } from './schema';
 
@@ -302,9 +306,14 @@ export class MastraModelOutput<OUTPUT extends OutputSchema = undefined> extends 
               }
               break;
             case 'object-result':
-              // Buffer the object - it will be resolved in the 'finish' step
-              // after output processors have had a chance to modify the text
               self.#bufferedObject = chunk.object;
+              // Track if the object came from a processor (e.g., StructuredOutputProcessor)
+              // If so, we should resolve immediately since the processor already produced the final object
+              // Otherwise, defer resolution to finish step so output processors can modify the text
+              const isFromProcessor = chunk.metadata?.from === 'structured-output';
+              if (isFromProcessor && self.#delayedPromises.object.status.type === 'pending') {
+                self.#delayedPromises.object.resolve(chunk.object);
+              }
               break;
             case 'source':
               self.#bufferedSources.push(chunk);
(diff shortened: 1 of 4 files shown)


packages/core/src/stream/base/output.ts
```diff
@@ -586,6 +584,17 @@ export class MastraModelOutput<OUTPUT extends OutputSchema = undefined> extends
                   self.#delayedPromises.text.resolve(outputText);
                   self.#delayedPromises.finishReason.resolve(self.#finishReason);
 
+                  // Resolve object with processed text parsed as JSON if possible
+                  if (self.#delayedPromises.object.status.type === 'pending') {
+                    try {
+                      const parsedObject = JSON.parse(outputText);
```
CalebBarnes, line 619: @DanielSLew I think this wont validate the object before we resolve it if we just call JSON.parse(). The processor could modify the text to not match the schema and then we would resolve it anyways as long as its valid JSON.

CalebBarnes: We could prob do something like this
```ts
const { success, value, error } = createOutputHandler({ schema }).validateAndTransformFinal(outputText)
if (!success) { 
  handleValidationError(error, controller) // <-- enqueue a validation error to the controller
  self.#delayedPromises.object.reject(error)
}
self.#delayedPromises.object.resolve(value)

```

packages/core/src/stream/base/output.ts
```diff
@@ -303,8 +307,11 @@ export class MastraModelOutput<OUTPUT extends OutputSchema = undefined> extends
               break;
             case 'object-result':
               self.#bufferedObject = chunk.object;
-              // Only resolve if not already rejected by validation error
-              if (self.#delayedPromises.object.status.type === 'pending') {
+              // Track if the object came from a processor (e.g., StructuredOutputProcessor)
+              // If so, we should resolve immediately since the processor already produced the final object
+              // Otherwise, defer resolution to finish step so output processors can modify the text
+              const isFromProcessor = chunk.metadata?.from === 'structured-output';
+              if (isFromProcessor && self.#delayedPromises.object.status.type === 'pending') {
```
CalebBarnes, line 314: What happens if I have 

```json
{
    structuredOutput: { schema, model }
}
```
provide a model (triggers use of StructuredOutputProcessor) and then also use an output processor to change the object

DanielSLew: that won't work in this setup. I think it's going to take a bunch of rearchitecting to make that work, it's probably not worth it atm

CalebBarnes (approved):

CalebBarnes:
🎉👏

Issue #10172, opened 2025-11-17
Tracing prompts with Langfuse


daneatmastra:
This issue was created from Discord post 1439974803250348052:

[![Open in Discord](https://img.shields.io/badge/Open_in_Discord-5865F2?style=for-the-badge&logo=discord&logoColor=white)](https://discord.com/channels/1309558646228779139/1439974803250348052)

I am using Langfuse for prompt management, and I can consume the prompts just fine. However, I cannot have the tracing link to the prompts to the ones created in langfuse, and versions. Here is an example of how I set it up
```
const headerAuditing = createStep({
  id: "header-auditing",
  description: "Diagnose the header",
  inputSchema: productDataSchema,
  outputSchema: scoreSchema,
  execute: async ({ mastra, inputData }) => {
    const agent = mastra.getAgent("headerAuditingAgent")
    const message = `
    ### Attribute to evaluate
    
    ${inputData.header}
    `
    const score = await observe(async () => {
      updateActiveObservation(
        { prompt: instructionsHeaderAuditing },
        { asType: "generation" },
      )
      return await agent.generate(message, {
        structuredOutput: {
          schema: scoreSchema,
        },
      })
    })()

    return score.object
  },
})
```

However, this is still not working. I can just see the raw prompt in Langfuse. Do we have any other tracing integration with Langfuse?

graysonhicks:
I have a [possible solution](https://github.com/mastra-ai/mastra/pull/8762#issuecomment-3559800894) that would make this work by having the exporter automatically handle prompt linking.

## What You'd Need to Change

Instead of using `observe()` and `updateActiveObservation()`, you'd just set the prompt info in metadata:

```typescript
const headerAuditing = createStep({
  id: "header-auditing",
  description: "Diagnose the header",
  inputSchema: productDataSchema,
  outputSchema: scoreSchema,
  execute: async ({ mastra, inputData, tracingContext }) => {
    const agent = mastra.getAgent("headerAuditingAgent");
    
    // Set prompt metadata
    tracingContext.currentSpan?.update({
      metadata: {
        promptName: instructionsHeaderAuditing.name,
        promptVersion: instructionsHeaderAuditing.version,
      }
    });
    
    const message = `
    ### Attribute to evaluate
    
    ${inputData.header}
    `;
    
    const score = await agent.generate(message, {
      structuredOutput: {
        schema: scoreSchema,
      },
      tracingContext,
    });

    return score.object;
  },
})
```

[...]

PR #8762, opened 2025-10-13
Add langfuse prompt tracing

zack-ashen:
## Description
This adds the ability to link prompts from llm generations to stored langfuse prompts. Specifically following this [reference](https://langfuse.com/integrations/native/opentelemetry#prompt). When you add a Langfuse object to the metadata of a generation:

```ts
tracingOptions: {
  metadata: {
    langfuse: {
      observation: {
        prompt: {
          name: <langfuse prompt name>,
          version: <langfuse prompt version>,
        },
      },
    },
  },
},
```

You can then add a processor to propagate it to the llm_generation span:

```ts
// Propagate root span `langfuse` metadata down to each LLM generation span
export const propagateLangfusePromptProcessor = {
    name: "propagate-langfuse-prompt-to-llm",
    process(span?: AnyAISpan): AnyAISpan | undefined {
        if (!span?.isValid) return span;
        if (span.metadata?.langfusePrompt) return span;

        try {
            if (span.type === AISpanType.LLM_GENERATION) {
                // Find the nearest ancestor with the metadata
                let cursor: AnyAISpan | undefined = span.parent;
                while (cursor) {
                    const langfuseData = cursor.metadata?.langfuse;
                    if (langfusePrompt || langfuse) {
                        span.attributes = {
                            ...span.attributes,
                            langfuse: langfuseData.
                        };
                        break;
                    }
                    cursor = cursor.parent;
                }
            }
        } catch {
            // no-op
        }
        return span;
    },
```

## Related Issue(s)
#8075 

## Type of Change

- [ ] Bug fix (non-breaking change that fixes an issue)
- [x] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [ ] Test update

## Checklist
- [ ] I have made corresponding changes to the documentation (if applicable)
- [ ] I have added tests that prove my fix is effective or that my feature works

PR #10738, opened 2025-12-01
fix langfuse prompt linking

graysonhicks:
## Description

This PR adds support for linking Langfuse Prompt Management prompts to model generation spans. Users can now associate their agent/LLM calls with prompts managed in Langfuse using a simple helper function. The implementation also supports **multiple different prompts within a single trace**, enabling workflows that call multiple agents each using different Langfuse-managed prompts.

### Changes:

**Composable tracing options builder (`observability/mastra/src/tracing-options.ts`):**
- Added `TracingOptionsUpdater` type for creating tracing option updaters
- Added `buildTracingOptions()` function for composing updaters from different providers
- Exported from `@mastra/observability`

**Langfuse prompt helper (`observability/langfuse/src/helpers.ts`):**
- Added `withLangfusePrompt()` helper that works with Langfuse SDK prompt objects or manual fields
- Automatically extracts `name`, `version`, and `id` from prompt objects
- Exported from `@mastra/langfuse`

**Langfuse exporter prompt linking (`observability/langfuse/src/tracing.ts`):**
- Added `LangfusePromptData` type for prompt metadata
- Added `SpanMetadata` type and `spanMetadata` map to track parent relationships and prompt data per span
- Added `findParentLangfusePrompt()` method to traverse up the span tree and find the closest ancestor with a Langfuse prompt
- Extracts `metadata.langfuse.prompt` from `MODEL_GENERATION` spans
- Maps prompt data to Langfuse's native `prompt` field on generations
- Supports linking by `name`+`version`, by `id` alone, or with all fields
- Supports prompt inheritance from parent spans (e.g., AGENT_RUN → MODEL_GENERATION)
- Supports multiple different prompts in a single trace (workflows with multiple agents)

**Documentation:**
- Updated Langfuse exporter docs (`docs/src/content/en/docs/observability/tracing/exporters/langfuse.mdx`) with prompt linking examples using the new helpers
- Updated Langfuse reference docs (`docs/src/content/en/reference/observability/tracing/exporters/langfuse.mdx`) with helper function documentation

**Tests:**
- Added tests for the `withLangfusePrompt` helper
- Added tests for prompt linking in the exporter
- Added tests for prompt inheritance from parent spans
- Added tests for multiple prompts in a single trace

### Test Cases Added:

**`withLangfusePrompt` helper tests:**
- `should add prompt metadata with name and version`
- `should add prompt metadata with id only`
- `should add prompt metadata with all fields`
- `should work with Langfuse SDK prompt object format`
- `should compose with other updaters`
- `should preserve existing metadata`

**Langfuse Prompt Linking tests:**
- `should link prompt to generation when metadata.langfuse.prompt is set`
- `should link prompt with all fields when name, version, and id are set`
- `should link prompt with id alone`
- `should not include prompt field for non-MODEL_GENERATION spans`
- `should omit langfuse property from metadata after extracting prompt`
- `should handle metadata.langfuse without prompt gracefully`
- `should inherit langfuse prompt from AGENT_RUN root span to child MODEL_GENERATION span`
- `should inherit langfuse prompt to MODEL_GENERATION even when root span has other metadata`
- `should prefer span-level langfuse prompt over inherited root span prompt`
- `should not add prompt to MODEL_GENERATION when root span has no langfuse data`

**Multiple Langfuse Prompts in Single Trace tests:**
- `should handle workflow calling multiple agents with different prompts`
- `should handle nested agents with different prompts`
- `should traverse up the span tree to find prompt from grandparent`
- `should handle mixed scenario: some agents with prompts, some without`
- `should handle deeply nested structure with prompt at different levels`
- `should use closest ancestor prompt when multiple ancestors have prompts`

### Usage:

```typescript
import { buildTracingOptions } from "@mastra/observability";
import { withLangfusePrompt } from "@mastra/langfuse";

const prompt = await langfuse.getPrompt("customer-support");

const agent = new Agent({
  instructions: prompt.prompt,
  defaultGenerateOptions: {
    tracingOptions: buildTracingOptions(withLangfusePrompt(prompt)),
  },
});
```

## Related Issue(s)

- #8762
- #10172

## Type of Change

- [x] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [x] Test update

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Added Langfuse prompt linking for LLM generations, enabling integration with Langfuse Prompt Management.
  * New `withLangfusePrompt` helper function for streamlined prompt metadata integration.
  * Added `buildTracingOptions` for composable tracing configuration.

* **Documentation**
  * Added "Prompt Linking" sections to Langfuse exporter documentation with usage examples and linking strategies.


packages/core/src/observability/types/tracing.ts
```diff
@@ -751,12 +751,49 @@ export interface TraceState {
...
+  /** Langfuse-specific metadata */
+  langfuse?: LangfuseMetadata;
+  /** User ID for the trace (used by Langfuse) */
+  userId?: string;
+  /** Session ID for the trace (used by Langfuse) */
+  sessionId?: string;
+  /** Allow arbitrary additional metadata fields */
+  [key: string]: any;
+}
+
```
epinzur: would it be possible to only have these types inside the `@mastra/langfuse` package?

docs/src/content/en/docs/observability/tracing/exporters/langfuse.mdx
```diff
@@ -107,7 +107,93 @@ new LangfuseExporter({
...
+import { Langfuse } from "langfuse";
+
+const langfuse = new Langfuse({
+  publicKey: process.env.LANGFUSE_PUBLIC_KEY!,
+  secretKey: process.env.LANGFUSE_SECRET_KEY!,
+});
+
+// Fetch the prompt from Langfuse Prompt Management
+const prompt = await langfuse.getPrompt("customer-support");
+
```
epinzur, line 132: Last night I thought about this some more, and was wondering if we could add some helper to our `@mastra/langfuse` package to build the `tracingOptions` for us... so we don't have to rely on the user getting the exact format correct.

Again it might be good to establish some re-usable pattern here that can be also used elsewhere.

epinzur: If you add this to `observability/mastra/src` and add it to the export at `observability/mastra/src/index.ts`:

[tracing-options.ts](https://github.com/user-attachments/files/23879308/tracing-options.ts)

Then you should be able to create in `@mastra/langfuse` something like:
```
import type { TracingOptions } from '@mastra/core/observability';
import type { TracingOptionsUpdater } from '@mastra/observability';
import type { LangfusePrompt } from 'langfuse'; // adjust if needed

/**
 * Adds Langfuse prompt metadata to the tracing options
 * to enable Langfuse Prompt Tracing
 *
 * The metadata is added under:
 *
 *   metadata.langfuse.prompt
 *
 * and includes:
 *   - name
 *   - version
 *   - id
 *
 * All fields are deeply merged with any existing metadata.
 *
 * @example
 * const prompt = await langfuse.getPrompt('customer-support');
 *
 * const tracingOptions = buildTracingOptions(
 *   withLangfusePrompt(prompt)
 * );
 *
 * @param prompt - A Langfuse prompt object containing name, version, and id.
 * @returns A function that adds Langfuse prompt metadata to TracingOptions.
 */
export function withLangfusePrompt(
  prompt: LangfusePrompt,
): TracingOptionsUpdater {
  return (opts) => ({
    ...opts,
    metadata: {
      ...opts.metadata,
      langfuse: {
        ...opts.metadata?.langfuse,
        prompt: {
          name: prompt.name,
          version: prompt.version,
          id: prompt.id,
        },
      },
    },
  });
}
```

And the example below becomes:

```

export const dynamicAgent = new Agent({
  name: "dynamic-agent",
  instructions: prompt.prompt, // Use the prompt text from Langfuse
  model: openai("gpt-4o"),
  defaultGenerateOptions: {
    tracingOptions: buildTracingOptions(withLangfusePrompt(prompt)),
  },
});
```

packages/core/src/observability/types/tracing.ts
epinzur: Or maybe these aren't needed at all?

docs/src/content/en/docs/observability/tracing/exporters/langfuse.mdx
graysonhicks: Nice, yea that would work as we add to other exporters.

packages/core/src/observability/types/tracing.ts
graysonhicks: yea, if we do the helper function way, dont think we need this type

observability/langfuse/src/tracing.ts
```diff
@@ -443,12 +445,35 @@ export class LangfuseExporter extends BaseExporter {
...
+
+      if (hasNameAndVersion || hasId) {
+        payload.prompt = {};
+
+        if (promptData?.name !== undefined) payload.prompt.name = promptData.name;
+        if (promptData?.version !== undefined) payload.prompt.version = promptData.version;
+        if (promptData?.id !== undefined) payload.prompt.id = promptData.id;
+
+        metadataToOmit.push('langfuse');
+      }
```
epinzur, line 536: how does this data get to the `MODEL_GENERATION` span?

in:
```
const agent = new Agent({
  name: "support-agent",
  instructions: prompt.prompt,
  model: openai("gpt-4o"),
  defaultGenerateOptions: {
    tracingOptions: buildTracingOptions(withLangfusePrompt(prompt)),
  },
});
```

`tracingOptions` is only passed to the initial creation of the `AGENT_RUN` span.

graysonhicks: Good catch.  Pushed a change to get the prompt info from the parent span for those.

observability/langfuse/src/tracing.ts
```diff
@@ -270,12 +273,17 @@ export class LangfuseExporter extends BaseExporter {
 
   private initTrace(span: AnyExportedSpan): void {
     const trace = this.client.trace(this.buildTracePayload(span));
+
+    // Extract langfuse prompt data from root span for inheritance by child spans
+    const langfuseData = span.metadata?.langfuse as { prompt?: LangfusePromptData } | undefined;
+
```
epinzur, line 300: I don't think this will be sufficient.

this works when there is only one prompt, and it is defined at the root level of the trace.

but what if it is a nested agent, or a workflow calling a bunch of agents, and each uses a separate prompt from langfuse prompt manager?

observability/langfuse/src/tracing.ts
```diff
@@ -270,12 +273,17 @@ export class LangfuseExporter extends BaseExporter {
...
     this.traceMap.set(span.traceId, {
       trace,
       spans: new Map(),
       events: new Map(),
       activeSpans: new Set(),
       rootSpanId: span.id,
+      langfusePrompt: langfuseData?.prompt,
```
epinzur: instead if you have a `MODEL_GENERATION` span, and it doesn't have `langfuse` metadata (the normal case).

Then get the parent `AGENT_RUN` span from this map, from the model generation span, via the parentSpanId.

``` (pseudo code)
let langfuseMetadata = undefined;

switch (span.type) {
  case: "MODEL_GENERATION":
    langfuseMetadata = span.metadata?.langfuse;
    if (!langfuseMetadata) {
       const parentAgentSpan = this.traceMap(span.TraceId).spans(span.parentSpanId);
       langfuseMetadata = parentAgentSpan?.metadata?.langfuse;
    break;
}
```

epinzur: i think there is even a helper around to navigate up traceMaps to find the closest span of a certain type.

but that might only exist on normal spans... not the lightweight exportedSpans.

epinzur (changes requested):
nearly there.  Now need to be able to handle multiple different langfuse prompts inside a single trace.

I proposed a solution below.

I really like your verbose test cases.

epinzur (approved):

PR #10781, opened 2025-12-02
fix: add time-to-first-token (TTFT) support for Langfuse integration

### LINKED ISSUE #9444: [DISCORD:1433005133951664179] langfuse/mastra integration - time to first token data is missing
This issue was created from Discord post: https://discord.com/channels/1309558646228779139/1433005133951664179

Hi, i have an agent built using matra, and i have integrated it with langfuse for observability.
In langfuse im trying to see the "time to first token" metric but i get only 0 there.
Other metrics seem to work fine.
is this issue on mastra's side or did i integrated in a wrong way?

abhiaiyer91:
Fixes #9444

Previously, Langfuse showed 0 for the "time to first token" metric because we weren't setting `completionStartTime` on generation spans.

This adds `completionStartTime` to model generation spans. The timestamp is captured automatically when the first content chunk arrives during streaming (text, reasoning, tool-call, or object chunks).

```typescript
// Before: TTFT always showed 0 in Langfuse
// After: TTFT is calculated from completionStartTime - startTime
```

Changes:
- Added `completionStartTime` field to `ModelGenerationAttributes`
- Updated Langfuse exporter to include `completionStartTime` as a top-level field
- Added first-chunk timestamp capture in `ModelSpanTracker`

## Release Notes

* **New Features**
  * Added time-to-first-token (TTFT) support for Langfuse observability. Streaming model responses now automatically capture the timestamp when the first token arrives, enabling precise TTFT metrics within Langfuse for performance monitoring.

* **Tests**
  * Added comprehensive test coverage for TTFT functionality, validating automatic timestamp capture in streaming responses and proper behavior for non-streaming generation scenarios.

### DIFF
--- a/observability/mastra/src/model-tracing.ts
+++ b/observability/mastra/src/model-tracing.ts
@@ -32,9 +32,28 @@ export class ModelSpanTracker {
   #accumulator: Record<string, any> = {};
   #stepIndex: number = 0;
   #chunkSequence: number = 0;
+  /** Tracks whether completionStartTime has been captured for this generation */
+  #completionStartTimeCaptured: boolean = false;
 
   constructor(modelSpan?: Span<SpanType.MODEL_GENERATION>) {
     this.#modelSpan = modelSpan;
+  }
+
+  /**
+   * Capture the completion start time (time to first token) when the first content chunk arrives.
+   * This is used by observability providers like Langfuse to calculate TTFT metrics.
+   */
+  #captureCompletionStartTime(): void {
+    if (this.#completionStartTimeCaptured || !this.#modelSpan) {
+      return;
+    }
+
+    this.#completionStartTimeCaptured = true;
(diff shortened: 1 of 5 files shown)


observability/mastra/src/model-tracing.ts
```diff
@@ -317,24 +336,40 @@ export class ModelSpanTracker {
             case 'text-start':
             case 'text-delta':
             case 'text-end':
+              // Capture completion start time on first text content
+              if (chunk.type === 'text-delta') {
```
abhiaiyer91, line 340: I dont think these are the right places, im going to adjust them

epinzur: can we just do this on the first chunk, independent of chunk type? Like on line 333?

epinzur (approved):

PR #10801, opened 2025-12-03
docs: move Cloud out of Deployment subcategory

bookercodes:
## Summary by CodeRabbit

* **Documentation**
  * Reorganized Mastra Cloud into a new top-level docs section for clearer navigation.
  * Updated internal documentation links to new relative paths to match the reorganization.
  * Adjusted site redirects to preserve existing links and route traffic to the new Mastra Cloud locations.


docs/src/content/en/docs/sidebars.js
```diff
@@ -631,6 +605,32 @@ const sidebars = {
         },
       ],
     },
+    {
+      type: "category",
```
PaulieScanlon, line 609: Not sure Mastra Cloud should be moved right to the bottom of the nav. Perhaps positioning it under Deployment, or even before would make more sense?

PaulieScanlon (changes requested):

PR #10828, opened 2025-12-03
paul/grwth-887-structured-output

PaulieScanlon:
## Description

- docs: add structured output

## Type of Change

- [x] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [ ] Test update

## Checklist

## Release Notes

* **Documentation**
  * Added comprehensive guide for structured output in agents, covering schema definition, generation methods, streaming, and error handling strategies.
  * Reorganized agents documentation navigation for improved navigation clarity.
  * Fixed formatting issues in agents documentation pages.


docs/src/content/en/docs/agents/processors.mdx
PaulieScanlon, line 1: I don't know why this file is here, we replaced this with Guardrails a long time ago.

DanielSLew: we need to keep this, processors aren't just guardrails

DanielSLew: guardrails are a type of processor

docs/src/content/en/docs/agents/structured-output.mdx
```diff
@@ -0,0 +1,203 @@
...
+  {
+    "name": "Evening",
+    "activities": ["Dinner", "Relax", "Read a book", "Sleep by 10pm"]
+  }
+]
+```
+
+## Streaming
+
+Streaming also supports structured output. The final structured object is available on `stream.object` after the stream completes. Text stream chunks are still emitted, but they contain natural language text rather than structured data.
```
DanielSLew: it's also available in a `object-result` chunk

docs/src/content/en/docs/agents/structured-output.mdx
```diff
@@ -0,0 +1,203 @@
...
+for await (const chunk of stream.textStream) {
+  process.stdout.write(chunk);
+}
+
+console.log(await stream.object)
+```
+
+## Structuring agent
+
+For more complex data shapes, using a structuring agent works better. When you provide a `model` to `structuredOutput`, Mastra uses a second agent under the hood to extract structured data from the main agent's natural language response. This makes two LLM calls, one to generate the response and another to turn that response into the structured object, which adds some latency and cost but can improve accuracy for complex schemas.
```
DanielSLew:
```suggestion
When your main agent isn't proficient at creating structured output you can provide a `model` to `structuredOutput`. In this case, Mastra uses a second agent under the hood to extract structured data from the main agent's natural language response. This makes two LLM calls, one to generate the response and another to turn that response into the structured object, which adds some latency and cost but can improve accuracy for complex structuring tasks.
```

DanielSLew (changes requested):
made some comments for changes
- We can't delete the processors doc page, we need that, that is separate from guardrails
- you removed the gemini note, we need to add that back in, Im' not picky about how it gets added in but it needs to exist somewhere 

:::info[Gemini 2.5 with tools]

Gemini 2.5 models do not support combining `response_format` (structured output) with function calling (tools) in the same API call. If your agent has tools and you're using `structuredOutput` with a Gemini 2.5 model, you must set `jsonPromptInjection: true` to avoid the error `Function calling with a response mime type: 'application/json' is unsupported`.

```typescript
const response = await agentWithTools.generate("Your prompt", {
  structuredOutput: {
    schema: yourSchema,
    jsonPromptInjection: true, // Required for Gemini 2.5 when tools are present
  },
});
```

:::

DanielSLew (approved):

PR #10850, opened 2025-12-04
fix(core): improve RequestContext type inference for entries/keys/values

### LINKED ISSUE #4467: [FEATURE] RuntimeContext supports more accurate type definition
### Problem Statement

The type definition of RuntimeContext is type-safe and accurate when set(). However, it is safe but inaccurate when get(), which is very inconvenient to use.
For example:
```typescript
type MyRuntimeContext = {
  name: string;
  age: number;
};

const runtimeContext = new RuntimeContext<MyRuntimeContext>();
runtimeContext.set('name', 'John');
runtimeContext.set('age', 20);

// The following get('age') infers string|number, but expects number
const age = runtimeContext.get('age');

```
### Proposed Solution

The following is the type redeclare I used in the project:

```typescript
declare module '@mastra/core/runtime-context' {
  export class RuntimeContext<Values extends Record<string, any> = Record<string, any>> {
    private registry: Map<string, any>;
    constructor(iterable?: Iterable<{ [K in keyof Values]: [K, Values[K]] }[keyof Values]>);
    set<K extends keyof Values>(key: K, value: Values[K]): void;
    set(key: string, value: unknown): void;
    get<K extends keyof Values>(key: K): Values[K];
    get(key: string): unknown;
    has(key: string): boolean;
    delete(key: string): boolean;
    clear(): void;
    keys(): IterableIterator<keyof Values>;
    values(): IterableIterator<Values[keyof Values]>;
    entries(): IterableIterator<{ [K in keyof Values]: [K, Values[K]] }[keyof Values]>;
    forEach<K extends keyof Values>(callbackfn: (value: Values[K], key: K, map: Map<string, any>) => void): void;
    size(): number;
  }
}
```

### Component

Workflows

### Alternatives Considered

_No response_

### Example Use Case

```typescript
for (const [key, value] of runtimeContext.entries()) {
  if (key === 'age') {
    // value is correctly inferred as number
  } else {
    // value is correctly inferred as string
  }
}
```

### Additional Context

_No response_

### Verification

abhiaiyer91:
Fixes the type inference for `RequestContext.entries()`, `keys()`, `values()`, and `forEach()` methods.

Previously, `entries()` returned `[string, union]` which didn't allow type narrowing when iterating. Now it returns a discriminated union of tuples:

```typescript
const ctx = new RequestContext<{ name: string; age: number }>();

for (const [key, value] of ctx.entries()) {
  if (key === 'age') {
    // value is now correctly inferred as `number`
  }
}
```

Fixes #4467

## Summary by CodeRabbit

* **New Features**
  * Improved RequestContext iteration methods so keys(), values(), entries(), and forEach() provide stronger type narrowing and IDE autocompletion for typed contexts.
* **Tests**
  * Added TypeScript type tests to validate precise key/value inference, discriminated entry tuples, nested contexts, and forEach typings.
* **Chores**
  * Added a changelog entry documenting the type-inference improvements.


packages/core/src/request-context/request-context.test-d.ts
LekoArts, line 1: Coderabbit gave a good comment to add these tests, too:

```ts
it('should infer correct value types from values()', () => {
  const context = new RequestContext<MyContext>();
  const values = context.values();
  expectTypeOf(values).toEqualTypeOf<IterableIterator<string | number | boolean>>();
});

it('should provide typed callback in forEach()', () => {
  const context = new RequestContext<MyContext>();
  context.forEach((value, key) => {
    if (key === 'age') {
      expectTypeOf(value).toEqualTypeOf<number>();
    }
  });
});
```

packages/core/src/request-context/request-context.test-d.ts
```diff
@@ -0,0 +1,116 @@
...
+      // set() has correct typing
+      context.set('name', 'John');
+      context.set('age', 25);
+      context.set('isActive', true);
+
+      // @ts-expect-error - age should be number, not string
+      context.set('age', 'invalid');
+
+      // @ts-expect-error - unknown key should not be allowed
+      context.set('unknownKey', 'value');
```
LekoArts: That's not using vitest's type testing. Try this (untested) code instead:

```ts
it('should accept correct value types in set()', () => {
  const context = new RequestContext<MyContext>();

  // Verify correct types work
  context.set('name', 'John');
  context.set('age', 25);
  context.set('isActive', true);

  // Assert the value type for 'age' must be number
  expectTypeOf<string>().not.toMatchTypeOf<Parameters<typeof context.set<'age'>>[1]>();

  // Assert 'unknownKey' is not a valid key
  expectTypeOf<'unknownKey'>().not.toMatchTypeOf<keyof MyContext>();
});
```

roaminro (approved):

Issue #7923, opened 2025-09-17
[FEATURE] Better support for guardrails


Vulcanostrol:
### Problem Statement

First of all, let me start by saying our team at our company is enjoying using Mastra so far in general: thanks for all the hard work you all are putting into this.

Mastra supports input and output processors with strategies like block, warn, and filtering. Processors can abort and Mastra then returns a “successful” response carrying a tripwire reason instead of throwing, which forces callers to inspect result.tripwireReason on every agent.generate() call. There is no unified way to 1) append a guardrail failure as a message and continue the agent as normal (with an added message like: "your tone-of-voice was deemed incorrect because ..." so the agent can make a better response without having to re-run and generate all context/invoke all tools etc), or 2) cancel the internal agent workflow immediately on trigger - for example: a prompt injection should simply just excalate to an error throw, which would just cancel the whole workflow. 

Specific gaps:
- Need an option where a guardrail failure is appended to the conversation and the agent proceeds.
- Need a way for a guardrail to cancel the active agent workflow execution. Mastra workflows expose abort controls, but processors cannot tap these directly.
- Current block halts invocation yet still returns a result object. Callers must always check processors’ outcomes instead of relying on structured failure handling.

### Proposed Solution

I am willing to go into discussion about how to do this 'the Mastra way' (I do not know your design ideas/philosophy behind Mastra), but I have some experience in implementing guardrails in production projects now. You can contact me on the Discord (Vulcano) if you want to chat as well, or I can help on a PR if needed.

A very simple short-term fix would be to let processors pass more information when calling `abort`.

In short: I believe input- and output-processors could be used as guardrails, but it would be best I think to structure it in the following way:
- Input- and output processors handle transformation, filtering, etc. of input and output. For example: PII, toxicity. But: processors do NOT alter the **flow** of the agent. Only the data.
- Guardrails, on the other hand, I think **should be able to alter the agent flow**. I would even go as far as to say: **guardrails should define what happens when it is set off**. And the logic for this should be with the guardrail's implementation. Examples include: add the error to the messageList and continue running the agent; throw an error and completely stop executing any workflow; do nothing special (but maybe invoke a webhook or smth for an alert)

There are of course infinite ways to implement this, but one straight-forward way to do this now might be to have something alongside `abort(message: string)`. Possibly something named `triggerGuardrail`, so it can be used in input- and output processors.

[...]

PR #10947, opened 2025-12-05
feat: enhanced processor system with retry, workflow orchestration, and tripwire improvements

DanielSLew:
Addresses #7923 - Better processor support in Mastra.

There were a lot of missing features of processors, all these new features are added to close the loop. The big ones are
- Workflows can now return tripwires, they bubble up from agents that return tripwires in a step
- You can write processors as workflows using the existing Workflow primitive, every processor flow is now a workflow.
- tripwires that you throw can now return additional information including ability to retry the step
- New processor method `processOutputStep` added which runs after every step.

**What's new:**

**1. Retry mechanism with LLM feedback** - Processors can now request retries with feedback that gets sent back to the LLM:
```typescript
processOutputStep: async ({ text, abort, retryCount }) => {
  if (isLowQuality(text)) {
    abort('Response quality too low', { retry: true, metadata: { score: 0.6 } });
  }
  return [];
};
```
Configure with `maxProcessorRetries` (default: 3). Rejected steps are preserved in `result.steps[n].tripwire`. Retries are only available in `processOutputStep` and `processInputStep`. It will replay the step with additional context added.

**2. Workflow orchestration for processors** - Processors can now be composed using workflow primitives:
```typescript
import { createStep, createWorkflow } from '@mastra/core/workflows';
import {
  ProcessorStepSchema,
} from '@mastra/core/processors';

const moderationWorkflow = createWorkflow({ id: 'moderation', inputSchema: ProcessorStepSchema, outputSchema: ProcessorStepSchema })
  .then(createStep(new lengthValidator({...})))
  .parallel([createStep(new piiDetector({...}), createStep(new toxicityChecker({...}))])
  .commit();

const agent = new Agent({ inputProcessors: [moderationWorkflow] });
```
Every processor array that gets passed to an agent gets added as a workflow
image

**3. Extended tripwire API** - `abort()` now accepts options for retry control and typed metadata:
```typescript
abort('reason', { retry: true, metadata: { score: 0.8, category: 'quality' } });
```

**4. New `processOutputStep` method** - Per-step output processing with access to step number, finish reason, tool calls, and retry count.

**5. Workflow tripwire status** - Workflows now have a `'tripwire'` status distinct from `'failed'`, properly bubbling up processor rejections.

**6. UI updates** - 

- Processor workflows show a "Processor" badge in the playground.

image

- Tripwire events display in agent chat with reason/metadata.

image

- Tripwire rendering added to workflow run UI

image

- Tripwire data in modal 

image

## Summary by CodeRabbit

* **New Features**
  * Processors can be authored/executed as workflows; new processor-backed demo agents, examples, and a UI processor badge.
  * Per-step post-LLM processing with retry support and configurable maxProcessorRetries; UI shows Tripwire notices with retry, processor ID, and metadata.

* **Bug Fixes / Behavior**
  * Tripwire events now carry structured details (reason, retry, metadata, processorId) and prevent blocked content from final outputs.
  * Tripwire propagation and processor-driven retries work across streaming and workflow runs.

* **Documentation**
  * Migration guide, processor enhancements plan, and docs updated with examples and usage notes.


packages/core/src/loop/workflows/schema.ts
```diff
@@ -24,7 +24,8 @@ import type { InferSchemaOutput, OutputSchema } from '../../stream/base/schema';
 
 // Type definitions for the workflow data
 export interface LLMIterationStepResult {
-  reason: LanguageModelV2FinishReason | 'abort';
+  /** Includes 'tripwire' and 'retry' for processor scenarios */
+  reason: LanguageModelV2FinishReason | 'tripwire' | 'retry';
```
DanielSLew, line 28: abort was the old name for the finish reason, changed it to tripwire here to be more consistent with naming

packages/core/src/workflows/workflow.ts
```diff
@@ -336,6 +380,379 @@ export function createStep<
...
+      execute: async ({
+        inputData,
+        requestContext,
+      }: {
+        inputData: z.infer<typeof ProcessorStepSchema>;
+        requestContext: RequestContext;
+      }): Promise<ProcessorStepOutput> => {
+        // Cast to output type for easier property access - the discriminated union
+        // ensures type safety at the schema level, but inside the execute function
+        // we need access to all possible properties
```
CalebBarnes, line 404: I wonder if it'd be better if this whole execute function lived in the ../processors dir and just import here `executeProcessorWorkflowStep` or something. Just so the processor related code is separated from the rest of generic workflows stuff since a lot of this code used to live in the processor runner. Not a huge deal though.

CalebBarnes: maybe even `ProcessorRunner.executeWorkflow({ inputData, requestContext })` ?

docs/processor-enhancements-plan.md
```diff
@@ -0,0 +1,425 @@
+# Processor Enhancements Implementation Plan
```
CalebBarnes, line 1: oh yeah we can remove this

packages/core/src/agent/agent-legacy.ts
```diff
@@ -292,22 +300,19 @@ export class AgentLegacyHandler {
...
             convertedTools,
             threadExists: false,
             thread: undefined,
             messageList,
             agentSpan,
-            ...(tripwireTriggered && {
-              tripwire: true,
-              tripwireReason,
-            }),
+            tripwire,
```
CalebBarnes, line 315: for agent legacy should we just leave it as the old tripwireReason? 🤷‍♂️

workflows/inngest/src/execution-engine.ts
```diff
@@ -337,8 +337,27 @@ export class InngestExecutionEngine extends DefaultExecutionEngine {
               payload: {},
             },
           };
+        } else if (result.status === 'tripwire') {
```
CalebBarnes, line 340: Did you try out tripwire in inngest? It might need some special handling since inngest has its own error retrying built in. If you throw any `Error` that is not a `NonRetriableError`, it will retry the step by default. Might need to throw a `NonRetriableError` with `cause` of tripwire data or something. Worth testing out though.

packages/core/src/agent/message-list/index.ts
```diff
@@ -288,12 +288,14 @@ export class MessageList {
...
     input: Set<string>;
     output: Set<string>;
+    context: Set<string>;
     getSource: (message: MastraDBMessage) => MessageSource | null;
   } {
     const sources = {
       memory: new Set(Array.from(this.memoryMessages.values()).map(m => m.id)),
       output: new Set(Array.from(this.newResponseMessages.values()).map(m => m.id)),
       input: new Set(Array.from(this.newUserMessages.values()).map(m => m.id)),
+      context: new Set(Array.from(this.userContextMessages.values()).map(m => m.id)),
```
CalebBarnes, line 298: Do you remember why this was needed?

packages/core/src/agent/agent-legacy.ts
DanielSLew: eh, I think we should just have it all the same so we don't need to manage two ways of things.

workflows/inngest/src/execution-engine.ts
DanielSLew: I left a comment in slack, but I'd rather do the inngest stuff in a follow up. Tripwire doesn't work today, so this won't make it any worse

CalebBarnes (approved):

PR #10950, opened 2025-12-05
om eval

CalebBarnes:
## Type of Change

- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [ ] Test update

## Checklist

### DIFF
--- a/explorations/longmemeval/src/commands/prepare.ts
+++ b/explorations/longmemeval/src/commands/prepare.ts
@@ -491,11 +491,11 @@ export class PrepareCommand {
       observationalMemory = new ObservationalMemory({
         storage: omStorage,
         observer: {
-          model: 'openai/gpt-4o-mini', // Use OpenAI to avoid Gemini rate limits
+          model: 'openai/gpt-4o-mini',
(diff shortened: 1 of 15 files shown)


explorations/longmemeval/src/commands/prepare.ts
```diff
@@ -617,14 +619,43 @@ export class PrepareCommand {
...
+          chalk.gray(`   Evidence sessions included: ${excludedEvidenceSessions.map(s => s.sessionId).join(', ')}`),
+        );
+      } else {
+        sessionsToProcess = sessionsInRange;
+        console.log(
+          chalk.yellow(
+            `\n📊 Processing limited to ${options.sessionLimit} sessions (${startIndex + 1} to ${endIndex})`,
+          ),
+        );
+      }
```
CalebBarnes, line 658: This stuff is here so that if you run a tiny benchmark like `--subset 1 --session-limit 3` it will make sure the evidence is always included in the history. Otherwise if you limit the sessions you might just not include the answers in the chat history so that the observer has a chance to make an observation on it. This is only here for small testing and doesn't do anything when running the whole dataset.

PR #9710, opened 2025-11-05
feat: add configurable tool concurrency option

clemparpa:
## Summary
Adds a `toolConcurrency` parameter to Agent configuration to allow controlling parallel tool execution.

## Motivation
Currently, tools execute with a hard-coded concurrency of 10 in the agentic execution workflow. This PR allows users to configure this value, including setting it to 1 for sequential execution when needed.

## Changes
- Add `toolConcurrency` option to `AgentCreateOptions` interface (`packages/core/src/agent/types.ts`)
- Propagate `toolConcurrency` through agent execution workflow chain:
  - `packages/core/src/agent/agent.ts`
  - `packages/core/src/agent/workflows/prepare-stream/index.ts`
  - `packages/core/src/agent/workflows/prepare-stream/stream-step.ts`
  - `packages/core/src/llm/model/model.loop.ts`
  - `packages/core/src/loop/types.ts`
- Use configurable concurrency in tool execution loop (`packages/core/src/loop/workflows/agentic-execution/index.ts`)
- Maintain backward compatibility with default value of 10

## Usage
\`\`\`typescript
const agent = new Agent({
  name: 'my-agent',
  options: {
    toolConcurrency: 1  // Sequential tool execution
  }
});
\`\`\`

## Testing
- ✅ TypeScript compilation successful
- ✅ All packages build without errors
- ✅ Backward compatible (defaults to 10)

---

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **New Features**
  * Added configurable tool concurrency control (default: 10) to limit the maximum number of tools executing simultaneously across agent workflows, streaming, and execution loops.

Issue #9704, opened 2025-11-04
[BUG] make tool call concurrency configurable


clemparpa:
### Describe the Bug

Hi, it's the really first time i write an issue on an open source project, i hope i do it well, 

## Problem Statement

I tried to use the very new HITL system in agent tools  (see blog post <https://mastra.ai/blog/tool-approval>).
but when the agent tries to execute multiple tools, it seems to have a race bug between event. 
Events are emitted in this order

[tool-call-1] -> [tool-call-2] -> [tool-call-suspended (from tool 1)]

and then the agent can't be resumed. 

After investigation, i think the problem is the parallel execution added in this PR <https://github.com/mastra-ai/mastra/pull/7524>
And discussed in this issue <https://github.com/mastra-ai/mastra/issues/7230>

Specifically this commit <https://github.com/mastra-ai/mastra/pull/7524/commits/2f71f7cf9711d5e3d08c2a3569585cafc1f7600a>

## Proposed Solution

I think a solution is to enable to change the concurrency parameters in the Agent loop, in the Agent configurations, 
like a boolean "enableToolConcurrency".  

I can try to work on a PR if you think it's a great idea. 

## Component
Tools, HITL

### Steps To Reproduce

1. Create a Mastra project
2.  Create Agent 
3.  Attach tools with HITL (human in the loop)
4.  execute a stream call with multiple tools 

### Expected Behavior

tool parallelism and HITL are incompatible . 
either the user should be able to configure one or the other

### Verification

- [x] I have searched the existing issues to make sure this is not a duplicate
- [x] I have included sufficient information for the team to reproduce and understand the issue

clemparpa:
after tests with the main branch of the project, it seems the race problem has already been patched
ie when multiple calls arrives the resume call validate all the tool calls. I don't know if its the expected behavior. 

However it could be good to be able to chose the concurrency of the agent .  it allow to validate each tool one at a time with concurrency = 1.

clemparpa:
I just made a PR for this issue. See <https://github.com/mastra-ai/mastra/pull/9710>

ShirazC:
is there an eta on when this will be merged? I'm blocked by the same issue

clemparpa:
Hi,
Even if it's not merged yet you can mitigate the bug with parallel tool call config in your llm provider. Some providers enable to use only 1 tool call at a time.

NikAiyer:
[...]

PR #10998, opened 2025-12-09
fix(core): enable conditionally triggered sequential tool calling #9704

astout55555:
## Description

Rather than always using a default limit of 10 concurrent tool calls during an agent's execution workflow, we've added support for sequential execution conditioned on whether the tools involved may require suspension for human approval. When approval is required, concurrent tool execution can cause race conditions that can break the workflow, preventing it from being resumed. This is now automatically guarded against by checking for either a global flag requiring tool approval, or if any tool has the `requireApproval` option set to `true` (or if it has the `suspendSchema` indicating that it may conditionally require approval).

Note that this is not the same approach as previously pursued in the [related (closed) PR](https://github.com/mastra-ai/mastra/pull/9710). Instead of taking the approach of making the concurrency configurable, putting the requirement on the user to identify the conflict between human-in-the-loop workflows and the default concurrency setting, this is abstracted away and handled behind the scenes, preventing the error from occurring.

An alternative approach would be to expand this change to include both options, allowing for configuration in addition to the automatic safeguard (which might be desirable to add increased functionality in case the user wants to only reduce or possibly increase the concurrency limit). Changes have already been identified in the code which would implement this approach, and could be added if desired, but for now the changes have been kept minimal.

## Related Issue(s)

Issue #9704
PR (closed): https://github.com/mastra-ai/mastra/pull/9710

## Type of Change

- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [ ] Test update

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Added a toolCallConcurrency option to control parallel tool execution (default 10).
  * Automatic sequential execution (concurrency = 1) when workflows include tools that require approval or suspend/wait-for-human input.
  * Tools surface a flag indicating presence of a suspend schema.

* **Tests**
  * Added integration tests validating sequential vs. concurrent tool execution across approval, suspension, and invalid-concurrency scenarios.


NikAiyer (commented):
LGTM! Can you add some tests for this as well?

astout55555:
The additional functionality (user controlled configuration of tool call concurrency at the level of agent creation) has been added. 

We chose to prioritize error prevention over user control in the following 2 scenarios:

1. If the user attempts to create an agent with an invalid `toolCallConcurrency` value (a negative number), it instead defaults to 10 like before.
2. If the user attempts to set the `toolCallConcurrency` value even though approval is required (which would cause the race condition error), the configuration is ignored and the value still defaults to 1 for sequential tool handling.

We'll begin work on adding tests for all this expected behavior.

NikAiyer (approved):

PR #11149, opened 2025-12-13
feat(core): add support for passing custom `assistantMessageId`

RostiMelk:
## Description

Adds support for passing a custom `assistantMessageId` when calling `agent.stream()` or `agent.generate()`. This allows you to specify the ID that will be used for the assistant response message when it's saved to memory.

```ts
const result = await agent.stream('Hello!', {
  memory: {
    thread: 'thread-123',
    resource: 'user-456',
    assistantMessageId: 'my-custom-id', // now works!
  },
});
```

This is useful when you need to know the assistant message ID upfront (e.g., for optimistic UI updates, offline mode reconciliation, or allowing users to interact with messages while they're still streaming).

Note: This only applies to the new message format used by `stream()` and `generate()`. The legacy methods are unaffected since they can split responses into multiple message records, making a single ID impractical.

## Related Issue(s)

Addresses #7222

## Type of Change

- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [ ] Test update

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Added support for passing an optional `assistantMessageId` through agent execution and streaming workflows, enabling better tracking and identification of assistant messages throughout the execution lifecycle.

* **Tests**
  * Added test coverage verifying that the `assistantMessageId` option is properly respected when streaming messages.

### DIFF
--- a/packages/core/src/agent/agent.types.ts
+++ b/packages/core/src/agent/agent.types.ts
@@ -133,9 +133,6 @@ export type AgentExecutionOptions<
 
   /** Whether to include raw chunks in the stream output (not available on all model providers) */
   includeRawChunks?: boolean;
-
-  /** Custom ID to use for the assistant response message when saving to memory */
-  assistantMessageId?: string;
 };
 
 export type InnerAgentExecutionOptions<
--- a/packages/core/src/agent/agent.types.ts
+++ b/packages/core/src/agent/agent.types.ts
@@ -133,6 +133,9 @@ export type AgentExecutionOptions<
 
   /** Whether to include raw chunks in the stream output (not available on all model providers) */
   includeRawChunks?: boolean;
+
+  /** Custom ID to use for the assistant response message when saving to memory */
+  assistantMessageId?: string;
 };
 
 export type InnerAgentExecutionOptions<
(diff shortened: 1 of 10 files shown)


packages/core/src/agent/agent.types.ts
```diff
@@ -133,6 +133,9 @@ export type AgentExecutionOptions<
 
   /** Whether to include raw chunks in the stream output (not available on all model providers) */
   includeRawChunks?: boolean;
+
+  /** Custom ID to use for the assistant response message when saving to memory */
+  assistantMessageId?: string;
```
RostiMelk: `assistantMessageId` could potentially live under the `memory` option as `memory.assistantMessageId`(?)

RostiMelk: edit: I made the move

wardpeet:
i commented on the original bug, i dont think we should need to do this

TylerBarnes:
Thanks for opening this.

I’m going to close this because a single caller-provided `assistantMessageId` doesn’t fit the current direction of the memory + streaming architecture.

The memory system can rotate response message IDs internally within a single agent loop, allowing it to chunk and create multiple response messages based on token size or other factors. Also, a single stream will soon be able to have user messages injected into the middle of it, so there may not be one stable assistant response ID for the whole stream.

For the optimistic UI / reconciliation use case, a better fit would be to provide an ID-generation function instead of one fixed assistant message ID. Callers that need reconciliation could keep track of generated IDs with a hoisted array/closure around that generator. There may also be a way to inspect the stream itself to see which message IDs were actually used.

Closing this PR for now, but the use case is valid — I think the implementation should be redesigned around generated/observed message IDs rather than a single fixed `assistantMessageId`.

Issue #11015, opened 2025-12-09
Workflows with suspend/resume cannot be called by Agents


daneatmastra:
This issue was created from Discord post 1447645218374553753:

[![Open in Discord](https://img.shields.io/badge/Open_in_Discord-5865F2?style=for-the-badge&logo=discord&logoColor=white)](https://discord.com/channels/1309558646228779139/1447645218374553753)

I have an agent which can use a workflow, which in itself has some steps that ask for HIL input, calling for some suspend and resume. The issue is that there's no current way of giving that input from the Mastra Studio right now. Even if I manually go into the workflow thread and put the values there, the main thread where I have the conversation doesn't get updated with that information.
Expected behavior from my UX point of view would be:
- If a workflow is suspended, for the agent to be able to consume more messages
- The agent being able to use the information from those new messages to interact with the suspended workflow and resume it.
- The agent view from Mastra Studio should see the state of the workflow at all times.

I'm using the latest version of the beta with the following packages:
    "@mastra/core": "1.0.0-beta.8",
    "@mastra/evals": "1.0.0-beta.2",
    "@mastra/libsql": "1.0.0-beta.5",
    "@mastra/loggers": "1.0.0-beta.3",
    "@mastra/memory": "1.0.0-beta.4",
    "@mastra/observability": "1.0.0-beta.3",

daneatmastra:
Thank you for reporting this issue! We have assigned it to the trio-tnt and we will look into it as soon as possible.

If you're experiencing an error, please provide a [minimal reproducible example](https://github.com/mastra-ai/mastra/blob/main/CONTRIBUTING.md#minimal-reproduction) whenever possible to help us resolve it quickly.

Issue #10957, opened 2025-12-06
When using agent.suspend() inside a tool, the message is not persisted.


daneatmastra:
This issue was created from Discord post 1446774155684085884:

[![Open in Discord](https://img.shields.io/badge/Open_in_Discord-5865F2?style=for-the-badge&logo=discord&logoColor=white)](https://discord.com/channels/1309558646228779139/1446774155684085884)

using `agent.suspend()` in tool has no any suspend message or metadata in db mastra_messages
unlike when using `requireApproval:true`  has  `pendingToolApprovals` in message.

taofeeq-deru:
#11157 addresses this

PR #11157, opened 2025-12-14
Agent tool resumption

taofeeq-deru:
## Description

Auto resume suspended tools when new message is sent if `autoResumeSuspendedTools: true`
This makes the agent resume a suspended tool call by using the message sent to construct `resumeData` for the suspended tool call.

The flag can be added to defaultAgentOptions when creating the agent or to options in agent.stream or agent.generate

```typescript
const agent = new Agent({
  //...agent information,
  defaultAgentOptions: {
    autoResumeSuspendedTools: true,
  }
})
```

Also persists message when tool is suspended

https://github.com/user-attachments/assets/7d5e612e-ec0f-4c97-94b0-f456b430ed8f

## Related Issue(s)

#10957 
#11015 

## Type of Change

- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [x] Test update

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Added autoResumeSuspendedTools option to automatically resume suspended tools during agent runs.
  * New workflow and example demonstrating suspend/resume flows.

* **Behavior**
  * Suspend/resume payloads and resume-schema handling enriched; suspended tool entries now tracked by tool name.
  * Workflow results may include consolidated suspendPayload data.

* **UI**
  * Assistant UI now displays tool suspend payloads alongside tool results.

* **Tests**
  * Expanded coverage for approval, suspension, and resume flows with memory propagation.


packages/core/src/loop/workflows/agentic-execution/tool-call-step.ts
```diff
@@ -49,18 +51,26 @@ export function createToolCallStep<
               ? (lastAssistantMessage.content.metadata as Record<string, any>)
               : {};
           metadata.pendingToolApprovals = metadata.pendingToolApprovals || {};
-          metadata.pendingToolApprovals[toolCallId] = {
+          metadata.pendingToolApprovals[toolName] = {
```
NikAiyer: What was the reason we switched from toolCallId to toolName? Is it easier to use elsewhere?

packages/core/src/loop/workflows/agentic-execution/tool-call-step.ts
```diff
@@ -107,6 +117,94 @@ export function createToolCallStep<
...
+
+            // Flush to persist the metadata removal
+            try {
+              await saveQueueManager.flushMessages(messageList, threadId, memoryConfig);
+            } catch (error) {
+              console.error('Error removing tool suspension metadata:', error);
+            }
+          }
+        }
+      };
```
NikAiyer: I wonder if we should merge these and the approval metadata helpers together. They seem very similar outside of a couple of code differences.

NikAiyer (approved):
LGTM. Just had a couple of questions about parts of the PR.

packages/core/src/loop/workflows/agentic-execution/tool-call-step.ts
taofeeq-deru: true, let me do that and confirm all works still

taofeeq-deru: merged them

packages/core/src/loop/workflows/agentic-execution/tool-call-step.ts
taofeeq-deru: yeah, for when we want to remove the pending tool approval/suspension metadata
toolName doesn't change but the toolCallId does because a new one is always created when the llm triggers a tool call.
This way, we can easily find the right tool metadata and remove it

PR #11189, opened 2025-12-15
fix(core): remove redundant toolCalls from network agent finalResult (#11059)

### LINKED ISSUE #11059: Network finalResult massive excess token usage
This issue was created from Discord post 1448099020944703639:

[Open in Discord](https://discord.com/channels/1309558646228779139/1448099020944703639)

In the agent network loop, [`packages/core/src/loop/network/index.ts`](https://github.com/mastra-ai/mastra/blob/798d0c740232653b1d754870e6b43a55c364ffe2/packages/core/src/loop/network/index.ts#L644):

```ts
await memory?.saveMessages({
    messages: [
        {
        id: generateId(),
        type: 'text',
        role: 'assistant',
        content: {
            parts: [
            {
                type: 'text',
                text: JSON.stringify({
                isNetwork: true,
                selectionReason: inputData.selectionReason,
                primitiveType: inputData.primitiveType,
                primitiveId: inputData.primitiveId,
                input: inputData.prompt,
                finalResult: { text: await result.text, toolCalls: await result.toolCalls, messages },
                }),
            },
            ],
            format: 2,
        },
        createdAt: new Date(),
        threadId: initData?.threadId || runId,
        resourceId: initData?.threadResourceId || networkName,
        },
    ] as MastraDBMessage[],
});
```

`finalResult` is redundantly storing `toolCalls` whose data will already be (to the extent desired after output processing!) already present in `messages`. I think this is a wrong implementation on its face because it deprives the user of control over what goes into the routing agent when it dumps `finalResult` into the network prompt ...but this is in source now and it's easy to imagine that there are users who are inadvertantly relying on this data being there multiple times to really convince a low perplexity routing LLM of the state of the network... from my POV it seems like this is both a) a MASSIVE waste of users resources, and b) a breaking change, so not sure how you folks want to handle a high priority breaking change.

NikAiyer:
## Description

Removes redundant `toolCalls` from network agent's `finalResult` to reduce token waste.

The `finalResult` object was storing `toolCalls` separately even though all tool call information is already present in the `messages` array (as `tool-call` and `tool-result` type messages). This caused significant token waste since the routing agent reads this data from memory on every iteration.

**Before:** `finalResult: { text, toolCalls, messages }`  
**After:** `finalResult: { text, messages }`

## Related Issue(s)

Fixes #11059

## Type of Change

- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [ ] Test update

## Checklist

## Summary by CodeRabbit

* **Refactor**
  * finalResult now contains only text and messages, removing redundant nested tool call fields to reduce token usage.
  * Client memory resolution updated to derive tool-call and tool-result data from messages rather than a separate toolCalls field.

* **Tests**
  * Added integration tests verifying finalResult no longer includes redundant tool call fields while preserving tool interactions within messages.


rase- (approved):

wardpeet (commented):
Do we have to update ai-sdk data-network as well? Or did the data- events stay untouched?

PR #11198, opened 2025-12-16
Seo: fix links to unsupported features

axe312ger:
This will get rid of 404s introduced by ourself when allowing v1 docs to be indexed again.

# Example:

Screenshot 2025-12-16 at 12 43 37

https://mastra-docs-git-seo-fix-links-to-unsuppo-0e66d0-kepler-ab1bf5ab.vercel.app/guides/v1/getting-started/next-js

## Summary by CodeRabbit

* **New Features**
  * Version pages show availability indicators—unsupported versions display a non-clickable "Not available" label and supported versions are clickable.
  * Version labels now include explicit version numbers (Stable v0, Beta v1).

* **Chores**
  * Added a feature-version mapping to drive availability checks.
  * Added many permanent redirects to route legacy documentation, guides, models, and reference URLs to their v1 locations.


docs/feature-versioning.json
```diff
@@ -0,0 +1,40 @@
+{
+  "/docs/agents/agent-approval": "1.0.0",
```
axe312ger, line 2: We might need a regex instead of a string as index, so we can match all versions.

But for now, this is not needed.

Same for the version string, I don't use it in code right now, but later we could use the `semver` package to reliable check if something is/was available in a certain version.

docs/src/components/version-control.tsx
```diff
@@ -11,9 +11,11 @@ import { useLocation } from "@docusaurus/router";
 import { cn } from "../lib/utils";
 import { BetaIcon, StableIcon, TriggerIcon, VersionLabel } from "./icons/icon";
 
+import FeatureVersioning from "../../feature-versioning.json";
+
 const versions = [
-  { value: "stable", label: "Stable" },
-  { value: "beta", label: "Beta" },
+  { value: "stable", label: "Stable (v0)" },
+  { value: "beta", label: "Beta (v1)" },
```
axe312ger, line 18: Took the liberty to add this. I think it helps the user to know which one they are looking at.

docs/src/components/version-control.tsx
```diff
@@ -86,27 +88,38 @@ export default function VersionControl({
...
+                  )}
+                </a>
+              ) : (
+                <div>
+                  <div className="inline-flex dark:text-white text-black gap-2">
+                    {version.value === "stable" ? <StableIcon /> : <BetaIcon />}
+                    <span>Not available in {version.label}</span>
+                  </div>
+                </div>
+              )}
```
axe312ger, line 122: Not a beauty, but we will replace this switcher soon when we switch to docusaurus internal doc version management

docs/vercel.json
```diff
@@ -434,6 +434,196 @@
...
+    },
+    {
+      "source": "/reference/vectors/duckdb",
+      "destination": "/reference/v1/vectors/duckdb",
+      "permanent": true
+    },
+    {
+      "source": "/reference/vectors/elasticsearch",
+      "destination": "/reference/v1/vectors/elasticsearch",
+      "permanent": true
```
axe312ger: We introduced the paths to the internet, so potentially they could be linked or indexed.. so lets redirect! :)

docs/src/components/version-control.tsx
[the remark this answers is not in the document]
axe312ger: This is irrelevant as the change is v1 only and v1 does not have any i18n

bookercodes (approved):
not had a chance to look in detail; approving so you're unblocked

LekoArts (approved):

PR #11234, opened 2025-12-17
Improved CWD for dev command

### LINKED ISSUE #11081: Improve file handling of storage files (cwd/relative paths)
### Problem Statement

When creating a new Mastra project our current docs recommendation is to write storage files like this:

```ts
import { LibSQLStore } from "@mastra/libsql";

const mastra = new Mastra({
  storage: new LibSQLStore({
    id: 'mastra-storage',
    url: "file:../mastra.db",
  }),
});
```

This will write the file to `.mastra` folder because the path is relative to the `.mastra/output` directory (the `process.cwd()` of `mastra dev` is that folder).

This can cause **really annoying papercuts**. This works if you use `mastra dev` but the moment you e.g. use your agents inside an API route of your preferred framework the `process.cwd()` will be the project root.

This creates two problems:
- The `mastra.db` file you use for `mastra dev` and for your framework are different
- The framework `mastra.db` file ends up one directory higher than your project root, it lands outside of your project!

### Proposed Solution

We should use the shell cwd (so e.g. if you run `npm run dev` in your project root, the cwd will be your project root). This is even more important now that you'll be able to use Mastra without bundling, e.g. with server adapters.

### Component

Other

### Alternatives Considered

_No response_

### Example Use Case

-

### Additional Context

_No response_

### Verification

wardpeet:
## Description

CWD of mastra dev is now the public dir. Means that all relative files are saved from public folder instead of `.mastra/output`. Files will now be persisted and the user can ignore them using gitignore or commit them.

Will store it inside public folder
```ts
new LibSQLStore({
    id: "mastra-storage",
    // stores observability, scores, ... into memory storage, if it needs to persist, change to file:../mastra.db
    url: "file:./mastra.db",
  }),
```
## Related Issue(s)

Fixes https://github.com/mastra-ai/mastra/issues/11081

## Type of Change

- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [ ] Test update

## Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Playground assets and HTML now load reliably via stable module-relative path resolution.

* **Changes**
  * Dev workflow writes generated files into the project's public directory and uses that directory as the dev server working directory.
  * Public directory contents are no longer automatically copied into build output during watch/build.
  * Server start now uses the running Node binary for consistent behavior.

* **Tests**
  * Integration tests run servers with the current Node runtime for more reliable test runs.

### DIFF
--- a/packages/cli/src/commands/dev/dev.ts
+++ b/packages/cli/src/commands/dev/dev.ts
@@ -95,6 +95,7 @@ const startServer = async (
 
     // Write mastra packages to a file and pass the file path via env var
     const packagesFilePath = join(dotMastraPath, '..', 'mastra-packages.json');
+    await mkdir(dotMastraPath, { recursive: true });
     if (startOptions.mastraPackages) {
       await writeFile(packagesFilePath, JSON.stringify(startOptions.mastraPackages), 'utf-8');
     }
--- a/packages/cli/src/commands/dev/dev.ts
+++ b/packages/cli/src/commands/dev/dev.ts
@@ -1,5 +1,5 @@
 import type { ChildProcess } from 'node:child_process';
-import { writeFileSync } from 'node:fs';
+import { mkdir, writeFile } from 'node:fs/promises';
 import { join } from 'node:path';
 import process from 'node:process';
 import devcert from '@expo/devcert';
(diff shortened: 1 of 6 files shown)


LekoArts (approved):

packages/cli/src/commands/dev/dev.ts
```diff
@@ -92,22 +91,22 @@ const startServer = async (
       commands.push(...startOptions.customArgs);
     }
 
-    commands.push('index.mjs');
+    commands.push(join(dotMastraPath, 'index.mjs'));
 
     // Write mastra packages to a file and pass the file path via env var
     const packagesFilePath = join(dotMastraPath, '..', 'mastra-packages.json');
     if (startOptions.mastraPackages) {
```
wardpeet, line 99:
```suggestion
    await mkdir(dotMastraPath, { recursive: true });
    if (startOptions.mastraPackages) {
```

PR #11257, opened 2025-12-18
fix: handle empty overrideScorers in agent scoring

YujohnNattrass:
When `overrideScorers` was passed as an empty object `{}`, the agent would throw a "No scorers found" error because `{}` is truthy but has no entries.

Now:
- `overrideScorers: undefined` → uses default scorers via `listScorers()`
- `overrideScorers: {}` → early return, skips scoring entirely  
- `overrideScorers: { ... }` → resolves and uses the provided scorers

```typescript
// Before: throws error
await agent.generate({ overrideScorers: {} })

// After: skips scoring gracefully
await agent.generate({ overrideScorers: {} })
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed scorer override handling to gracefully skip scoring when provided with an empty override configuration.
  * Enhanced error handling for cases where specified override scorers cannot be resolved, with improved error messaging.

### DIFF
--- a/packages/core/src/agent/agent.ts
+++ b/packages/core/src/agent/agent.ts
@@ -2412,11 +2412,6 @@ export class Agent<TAgentId extends string = string, TTools extends ToolsInput =
     resourceId?: string;
     tracingContext: TracingContext;
   }) {
-    // If overrideScorers is explicitly empty, skip scoring
-    if (overrideScorers && Object.keys(overrideScorers).length === 0) {
-      return;
-    }
-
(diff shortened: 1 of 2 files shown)


packages/core/src/agent/agent.ts
```diff
@@ -2412,6 +2412,11 @@ export class Agent<TAgentId extends string = string, TTools extends ToolsInput =
     resourceId?: string;
     tracingContext: TracingContext;
   }) {
+    // If overrideScorers is explicitly empty, skip scoring
+    if (overrideScorers && Object.keys(overrideScorers).length === 0) {
```
abhiaiyer91: `overrideScorers || {}`

PR #11267, opened 2025-12-18
Fix: telemetry decorators ignoring `enabled: false` configuration

miguelff:
## Problem

When users configured `telemetry: { enabled: false }` in their Mastra initialization, telemetry decorators would still execute and create spans despite the explicit opt-out. This occurred because the `hasActiveTelemetry()` utility function only checked for the existence of a tracer, not whether telemetry was actually enabled.

This was observed at [Omnia](useomnia.com) when we saw excess heap allocation due to the creation of spans, when we had telemetry disabled in mastra. 

### Root Cause

The previous implementation of `hasActiveTelemetry()` in [utility.ts](packages/core/src/telemetry/utility.ts) used:

```typescript
export function hasActiveTelemetry(tracerName: string = 'default-tracer'): boolean {
  try {
    return !!trace.getTracer(tracerName);
  } catch {
    return false;
  }
}
```

The OpenTelemetry `trace.getTracer()` API [**always returns a tracer object**](https://github.com/open-telemetry/opentelemetry-js/blob/main/packages/opentelemetry-sdk-trace-base/src/BasicTracerProvider.ts#L64-L84) Therefore, `!!trace.getTracer('default-tracer')` always evaluated to `true`.

The new implementation combines both checks:

```typescript
export function hasActiveTelemetry(tracerName: string = 'default-tracer'): boolean {
  try {
    const isEnabled = globalThis.__TELEMETRY__?.isEnabled() ?? true;
    return isEnabled && !!trace.getTracer(tracerName);
  } catch {
    return false;
  }
}
```

This ensures telemetry is only disabled when explicitly configured. The logic:
- If `__TELEMETRY__` is not initialized → `isEnabled` is `true` (backward compatible)
- If `__TELEMETRY__` exists → calls `isEnabled()` method (respects configuration)

The short-circuit evaluation provides an early exit for better performance when telemetry is disabled. 

### Impact

All telemetry decorators (`@withSpan`, `@InstrumentClass`) and runtime instrumentation methods (`traceClass()`, `traceMethod()`) invoked throughout the codebase would:

1. Execute instrumentation logic even when `enabled: false` was set
2. Create spans and serialize arguments/results (performance overhead)
3. Process these spans through the no-op tracer (wasted CPU cycles)
4. Ignore the user's explicit configuration to disable telemetry

This was particularly problematic for:
- Performance-sensitive applications where users explicitly disabled telemetry
- Development environments where telemetry was unnecessary overhead
- Privacy-focused deployments where span data should never be collected

## Solution

Uses the existing `globalThis.__TELEMETRY__` instance to check the enabled state directly instead of introducing a new global flag

### Disable at Initialization

```typescript

const mastra = new Mastra({
  telemetry: {
    enabled: false  // Completely disables telemetry
  },
  agents: { myAgent }
});

// hasActiveTelemetry() → false
// No spans created, zero overhead
```

### Runtime Control

```typescript
import { Telemetry } from '@mastra/core';

// Disable telemetry dynamically
Telemetry.setEnabled(false);

// Do work without telemetry
await agent.generate('test');

// Re-enable telemetry
Telemetry.setEnabled(true);

// Spans are created again
await agent.generate('test 2');
```

### Environment-Based Configuration

```typescript

const mastra = new Mastra({
  telemetry: {
    enabled: process.env.NODE_ENV === 'production',
    serviceName: 'my-service'
  }
});
```

### Before Fix

```typescript
const mastra = new Mastra({
  telemetry: { enabled: false },
  agents: { myAgent }
});

// hasActiveTelemetry() → true ❌ (wrong!)
// Spans were created and processed by no-op tracer
// Performance overhead from serialization
```

### After Fix

```typescript
const mastra = new Mastra({
  telemetry: { enabled: false },
  agents: { myAgent }
});

// hasActiveTelemetry() → false ✅ (correct!)
// No spans created, zero overhead
// Decorators are complete no-ops
```

### DIFF
--- a/packages/core/src/telemetry/utility.ts
+++ b/packages/core/src/telemetry/utility.ts
@@ -1,22 +1,17 @@
 import { propagation, trace } from '@opentelemetry/api';
 import type { Context } from '@opentelemetry/api';
 
-declare global {
-  var __TELEMETRY_ENABLED__: boolean | undefined;
-}
-
 /**
  * Check if telemetry is active and enabled.
- * This function checks both if a tracer exists AND if telemetry is enabled in the configuration.
+ * This function checks both if telemetry is enabled in the configuration AND if a tracer exists.
  *
  * @param tracerName - Optional tracer name to check (currently unused, kept for backwards compatibility)
  * @returns true if telemetry is both initialized and enabled, false otherwise
  */
 export function hasActiveTelemetry(tracerName: string = 'default-tracer'): boolean {
   try {
-    // Check if telemetry is explicitly enabled via the global flag
-    // This flag is set by Telemetry.init() based on the enabled config option
-    return globalThis.__TELEMETRY_ENABLED__ === true && !!trace.getTracer(tracerName);
+    const isEnabled = globalThis.__TELEMETRY__?.isEnabled() ?? true;
+    return isEnabled && !!trace.getTracer(tracerName);
   } catch {
     return false;
   }
--- a/packages/core/src/telemetry/utility.ts
+++ b/packages/core/src/telemetry/utility.ts
@@ -1,10 +1,22 @@
-import { propagation, trace } from '@opentelemetry/api';
+import { propagation } from '@opentelemetry/api';
 import type { Context } from '@opentelemetry/api';
 
-// Helper function to check if telemetry is active
-export function hasActiveTelemetry(tracerName: string = 'default-tracer'): boolean {
+declare global {
+  var __TELEMETRY_ENABLED__: boolean | undefined;
+}
+
+/**
+ * Check if telemetry is active and enabled.
+ * This function checks both if a tracer exists AND if telemetry is enabled in the configuration.
+ *
+ * @param _tracerName - Optional tracer name to check (currently unused, kept for backwards compatibility)
+ * @returns true if telemetry is both initialized and enabled, false otherwise
+ */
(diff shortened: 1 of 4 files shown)


packages/core/src/telemetry/utility.ts
```diff
@@ -1,10 +1,17 @@
...
+ * This function checks both if telemetry is enabled in the configuration AND if a tracer exists.
+ *
+ * @param tracerName - Optional tracer name to check (currently unused, kept for backwards compatibility)
+ * @returns true if telemetry is both initialized and enabled, false otherwise
+ */
 export function hasActiveTelemetry(tracerName: string = 'default-tracer'): boolean {
   try {
-    return !!trace.getTracer(tracerName);
+    const isEnabled = globalThis.__TELEMETRY__?.isEnabled() ?? true;
+    return isEnabled && !!trace.getTracer(tracerName);
```
miguelff, line 14: Actually, `trace.getTracer` is always going to return a Tracer instance, and hence `!!trace.getTracer(tracerName)` always evaluates to true, regardless of the value of `tracerName` passed-in.

The none-default call, is only made by the `InstrumentClass` decorator, in particular when decorating an `Agent`. We could change the signature of `hasActiveTelemetry` and made the caller rely only on the value of the `telemetry.enabled` property, If that's preferred. The effect will be the same, because what comes after the `&&` is always true.

abhiaiyer91 (approved):

wardpeet (approved):

Issue #6182, opened 2025-07-23
[INTEGRATION] Support Datadog for telemetry


ctamulonis-glossy:
### Service Information

Datadog LLM observability
API: https://docs.datadoghq.com/llm_observability/setup/api/?tab=model#span
SDK: https://docs.datadoghq.com/llm_observability/sdk/?tab=nodejs#advanced-tracing

### Use Case

Datadog has an LLM observability feature. We use Datadog for most of our observability needs and would be interested in using it for AI Agent observability as well.

### Verification

- [x] I have searched the existing issues to make sure this integration hasn't already been requested
- [x] I have checked that this service has a public API that can be integrated

epinzur:
hi @ctamulonis-glossy. We currently don't have plans to make a Datadog exporter for the new ai Observability stack.

If you or someone else wanted to make a contribution to add this, we would be very grateful.  You could look at the existing providers in this location for the general structure of what is needed:

https://github.com/mastra-ai/mastra/tree/main/observability

Alternatively, maybe the Otel exporter that will be released soon would function well enough for use in DataDog.

epinzur:
@ctamulonis-glossy we have decided to now work on a Datadog Exporter... but it will still be a few weeks before we get to it.

abarone-btf:
Hi! @epinzur , I dont want to put presion or anything, just to have it on the radar.

Do you know an approximate/expected date for this feature?

Thanks!

epinzur:
There is no specific date, but here is an overall timeline with the current tasks we know about, and rough info about what we are currently doing and what is next. 

https://github.com/mastra-ai/mastra/issues/8149

abarone-btf:
awesome! @epinzur , in your experience that would be more than say, 3 months? (roughly, nothing fixed or anything)

epinzur:
I would expect it to exist in the next month or so.

epinzur:
this should probably be based on the existing [Langfuse](https://github.com/mastra-ai/mastra/tree/main/observability/langfuse) and/or [Braintrust](https://github.com/mastra-ai/mastra/tree/main/observability/braintrust) exporters.

edgar-zigis:
+1 on this.

epinzur:
@abarone-btf our native exporter isn't ready yet, but a few users showed us that you can already send traces to Datadog APM with the Otel Exporter. We updated the docs to include the specific config info to get you started:

[...]

PR #11305, opened 2025-12-19
feat(observability): added Datadog exporter #6182

astout55555:
## Description

A native Datadog exporter has been built which maps Mastra span types onto Datadog span kinds and organizes them recursively into Datadog traces reflecting the proper tree structure. (Datadog's SDK requires span context to be inferred from callbacks, so spans cannot simply have their parent span assigned after the trace is created.) The DatadogExporter class can be instantiated and added to the config options' list of exporters when creating a new Mastra instance. See the README or the Mastra docs or Mastra reference doc for details on basic setup, config options, and available methods and utility functions.

Metrics are handled separately and then added to an annotations object on each span (where tags can also be added by the user). Utility functions (including the span type mapping) have been extracted to their own `utils.ts` file, leaving just the `DatadogExporter` class definition to be the dedicated purpose of the main `tracing.ts` file. Comprehensive tests have been added for each of these files, checking internal function and for correct output. Documentation has also been added.

### Recommended Future Work

Change the maximum time-to-live for spans not yet emitted from a hardcoded 30 minutes to a configurable amount. This could be handled as part of a cross-exporter memory leak management effort (ideally), or most simply as a configuration option for the Datadog exporter alone (add it as an option during exporter instantiation referenced by the cleanup function, rather than the maximum ttl constant). The default standard cleanup timer of 60 seconds could also be adjustable.

### Special Thanks

Thanks to @RichardLoRicco and @camachoh1 for building the skeleton for this exporter before @zijin1994 and I took over!

## Related Issue(s)

#6182 

## Type of Change

- [X] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [ ] Test update

## Checklist

- [X] I have made corresponding changes to the documentation (if applicable)

## Summary by CodeRabbit

* **New Features**
  * Datadog LLM Observability exporter (1.0.0‑beta) — hierarchical parent‑first tracing, span kind mapping, message formatting, dual delivery modes (agentless or local agent), token metrics, error tracking, and optional evaluation scoring.

* **Configuration**
  * Code and environment-variable configuration with sensible defaults for service/env/session/user.

* **Documentation**
  * Full docs, reference, and README with setup and usage examples.

* **Tests**
  * Comprehensive tests for formatting, metrics, tracing lifecycles, and error scenarios.

* **Chores**
  * Packaging and build/test config for the new exporter.

---

> [!NOTE]
> Introduces a Datadog exporter for LLM tracing (with metrics, scoring, hierarchy) and updates server schema to carry span tags in traces.
> 
> - **Observability**:
>   - **Datadog exporter**: New package `@mastra/datadog` implementing `DatadogExporter` to send Mastra traces to Datadog LLM Observability.
>     - Maps Mastra `SpanType` to Datadog kinds; emits hierarchical traces; formats LLM inputs/outputs; captures token usage metrics; supports agentless/agent modes; evaluation scoring via `addScoreToTrace()`; default user/session IDs; error tagging.
>     - Core implementation in `observability/datadog/src/tracing.ts` with helpers in `src/utils.ts` and `src/metrics.ts`.
> - **Server**:
>   - Adds `tags: string[]` to `tracingOptionsSchema` in `packages/server/src/server/schemas/common.ts` to preserve/export span tags.
> - **Docs & Reference**:
>   - Adds setup/usage docs and API reference for `DatadogExporter` under `docs/src/content/en/docs/.../datadog.mdx` and `.../reference/.../datadog.mdx`; package `README.md` and `CHANGELOG.md`.
> - **Tests**:
>   - Comprehensive unit tests for tracing behavior, utilities, and metrics (`src/*.test.ts`).
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 7ba477a9c6ebbf8cf8638c75bdd9fc37bc050c65. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


.changeset/polite-shrimps-cheer.md
```diff
@@ -0,0 +1,56 @@
...
+            apiKey: process.env.DD_API_KEY,
+          }),
+        ],
+      },
+    },
+  }),
+});
+```
+
+This is an initial beta release. The API is stable but may receive enhancements in future versions.
```
epinzur: pls change this wording to indicate this is an initial experimental release. (don't mention the API is stable).

.changeset/polite-shrimps-cheer.md
```diff
@@ -0,0 +1,56 @@
...
+
+Also, a bug was patched in the mastra server package to prevent tags from being stripped from exported traces. This fix may be implemented differently in ongoing work implementing Mastra's tagging capabilities.
```
epinzur: please remove this line entirely.

docs/src/content/en/docs/observability/tracing/exporters/datadog.mdx
```diff
@@ -0,0 +1,150 @@
...
+  },
+});
+```
+
+Tags appear in Datadog's trace view and can be used for filtering. Common use cases include:
+
+- Environment labels: `"production"`, `"staging"`
+- Experiment tracking: `"experiment-v1"`, `"control-group"`
+- Priority levels: `"priority-high"`, `"batch-job"`
+- User segments: `"beta-user"`, `"enterprise"`
```
epinzur: I think you can omit the `## Using Tags` section.  this should be in the root-level tracing overview docs.

docs/src/content/en/docs/observability/tracing/exporters/datadog.mdx
```diff
@@ -0,0 +1,150 @@
...
+**Important**: Evaluations must be submitted within 60 seconds of the trace completing. Scores submitted after trace cleanup will be dropped with a warning log.
+
+## Span Type Mapping
+
+Mastra span types are automatically mapped to Datadog LLMObs span kinds:
+
+| Mastra SpanType      | Datadog Kind |
+| -------------------- | ------------ |
+| `AGENT_RUN`          | `agent`      |
+| `MODEL_GENERATION`   | `workflow`   |
```
epinzur, line 100: why is `MODEL_GENERATION` a workflow?

docs/src/content/en/docs/observability/tracing/exporters/datadog.mdx
```diff
@@ -0,0 +1,150 @@
...
+| `MODEL_STEP`         | `llm`        |
```
epinzur, line 101: this should probably be a `task`?

observability/datadog/src/metrics.test.ts
```diff
@@ -0,0 +1,98 @@
+/**
+ * Tests for metrics transformation utilities
+ */
+
+import { describe, it, expect } from 'vitest';
+import { formatUsageMetrics } from './metrics';
+
+describe('formatUsageMetrics', () => {
+  it('transforms AI SDK v5 usage metrics format to Datadog metrics format', () => {
```
epinzur: just a nit, but this isn't `AI SDK v5` usage metrics format.  It is `OpenInference` format.

observability/datadog/src/tracing.ts
```diff
@@ -0,0 +1,746 @@
...
+
+  /**
+   * Default user ID applied to all spans if not specified in metadata.
+   */
+  defaultUserId?: string;
+
+  /**
+   * Default session ID applied to all spans if not specified in metadata.
+   */
+  defaultSessionId?: string;
```
epinzur: please remove these from the config, unless they are required parts of datadog.

`defaultUserId`, `defaultSessionId`

these would be better suited in the top-level observability setup.

observability/datadog/src/tracing.ts
```diff
@@ -0,0 +1,746 @@
...
+      annotations.inputData = formatInput(span.input, span.type);
+    }
+
+    // Format and add output (dd-trace expects 'outputData')
+    if (span.output !== undefined) {
+      annotations.outputData = formatOutput(span.output, span.type);
+    }
+
+    // Normalize and add token usage metrics
+    const usage = (span.attributes as ModelGenerationAttributes | undefined)?.usage;
```
epinzur: you should be able to switch on `spanType`.  `usage` will only exist on `MODEL_GENERATION` spans.

observability/datadog/src/tracing.ts
```diff
@@ -0,0 +1,746 @@
...
+      if (!state.rootEnded) return;
+
+      // Find the root span in the buffer
+      let rootSpan: AnyExportedSpan | undefined;
+      for (const span of state.buffer.values()) {
+        if (span.isRootSpan) {
+          rootSpan = span;
+          break;
+        }
+      }
```
epinzur: maybe you want to store the `rootSpan` outside of the `buffer`? then you don't need to do this iteration on every `tryEmitReadySpans()` call.

it is possible for traces to have hundreds of spans.

observability/datadog/src/tracing.ts
```diff
@@ -0,0 +1,746 @@
...
+    const options: LLMObsSpanOptions = {
+      kind,
+      name: span.name,
+      sessionId: traceCtx.sessionId,
+      userId: traceCtx.userId,
+      startTime,
+      endTime,
+      ...(kind === 'llm' && attrs?.model ? { modelName: attrs.model } : {}),
+      ...(kind === 'llm' && attrs?.provider ? { modelProvider: attrs.provider } : {}),
+    };
```
epinzur: this logic seems duplicate of the logic to build a `LLMObsSpanOptions` object from `emitSpanTree()`

can we extract this into a helper function?

observability/datadog/src/utils.ts
```diff
@@ -0,0 +1,179 @@
...
+  [SpanType.WORKFLOW_STEP]: 'task',
+  [SpanType.WORKFLOW_CONDITIONAL]: 'task',
+  [SpanType.WORKFLOW_CONDITIONAL_EVAL]: 'task',
+  [SpanType.WORKFLOW_PARALLEL]: 'task',
+  [SpanType.WORKFLOW_LOOP]: 'task',
+  [SpanType.WORKFLOW_SLEEP]: 'task',
+  [SpanType.WORKFLOW_WAIT_EVENT]: 'task',
+  [SpanType.PROCESSOR_RUN]: 'task',
+  [SpanType.GENERIC]: 'task',
+};
```
epinzur, line 24: I'd prob just map the non-task ones... and then allow a fallback to `task`.  

There will be more SpanTypes in the future and having them default to `task` without having to make a code change to the exporters, is preferred.

packages/server/src/server/schemas/common.ts
```diff
@@ -82,6 +82,7 @@ export const tracingOptionsSchema = z.object({
   requestContextKeys: z.array(z.string()).optional(),
   traceId: z.string().optional(),
   parentSpanId: z.string().optional(),
+  tags: z.array(z.string()).optional(),
```
epinzur: pls undo. this will be fixed outside this PR.

epinzur (changes requested):
Awesome start. I've added a bunch of comments below, but y'all are really close to having this ready.

docs/src/content/en/docs/observability/tracing/exporters/datadog.mdx
zijin1994: We found [this](https://docs.datadoghq.com/llm_observability/terms/#workflow-span) in their doc, and thought this is the most fitting option for this. It groups stuff together (llm steps + tool calls etc).


model generation span

docs/src/content/en/docs/observability/tracing/exporters/datadog.mdx
zijin1994: We felt that [llm](https://docs.datadoghq.com/llm_observability/terms/#llm-span) is more accurate in this case because it represents a llm call. 
step 0 span

observability/datadog/src/utils.ts
zijin1994: We removed the hard coded mapping, the `kindFor` function handles the fallback.

observability/datadog/src/tracing.ts
astout55555: We also found that it exists on `MODEL_STEP` spans, so we made it conditional on either of those two (still more efficient than the previous code).

observability/datadog/src/tracing.ts
astout55555: We ended up removing this check entirely, rather than storing the `rootSpan` reference on the `traceState`. Checking for the existence of the `rootSpan` in the buffer served the same purpose as checking that the `rootSpan` was completed, and had the same consequence. Condensing the logic to a check on the `traceState` made it clear that it was redundant, since it would be set at the same time as the `rootEnded` boolean. There didn't seem to be a real possibility of the `rootSpan` ending without being added to the buffer, so the inefficient iteration on every call didn't seem worthwhile (even though it's technically safer to check in both ways).

docs/src/content/en/docs/observability/tracing/exporters/datadog.mdx
epinzur: eh, looks good i guess. seems odd tho.

docs/src/content/en/docs/observability/tracing/exporters/datadog.mdx
epinzur: In the other providers, when the type is `llm` or `model_generation`, they expect other information like usage data. Because we don't have that info available here, I would expect this to be a different type.

epinzur (approved):

vvatanabe:
@astout55555 I was looking for a way to integrate Mastra with Datadog and came across this PR. I pulled the code locally and verified its behavior, and overall it works without any issues. 

Thank you for the excellent implementation!

Note: When tracing streams, I noticed that the output-processor span can be generated in large numbers on a per-chunk basis. On the consumer side, using SpanOutputProcessor to filter out noisy spans made the traces much easier to read.

epinzur (approved):

epinzur:
@vvatanabe 

> @epinzur I tried mastra/core 1.0.0-beta.20, and the spans are now much better organized and significantly easier to read. Thank you!

Awesome, thanks for the feedback!

PR #11328, opened 2025-12-21
feat(deployer-cloud): add studio option for bundling playground UI

YujohnNattrass:
Adds a `studio` option to `CloudDeployer` that bundles the playground UI into the cloud deployment output, similar to the `--studio` flag in the CLI.

```typescript
// before: no studio option
const deployer = new CloudDeployer();

// after: enable studio bundling
const deployer = new CloudDeployer({ studio: true });
```

When enabled, the playground assets get copied to the output directory and the server is configured to serve the studio UI.

### DIFF
--- a/deployers/cloud/src/index.ts
+++ b/deployers/cloud/src/index.ts
@@ -3,7 +3,7 @@ import { fileURLToPath } from 'node:url';
 import type { Config } from '@mastra/core/mastra';
 import { Deployer } from '@mastra/deployer';
 import { IS_DEFAULT } from '@mastra/deployer/bundler';
-import { readJSON } from 'fs-extra/esm';
+import { copy, readJSON } from 'fs-extra/esm';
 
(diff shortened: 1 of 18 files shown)


deployers/cloud/src/index.ts
```diff
@@ -12,8 +12,11 @@ import { getMastraEntryFile } from './utils/file.js';
 import { successEntrypoint } from './utils/report.js';
 
 export class CloudDeployer extends Deployer {
-  constructor() {
+  private studio: boolean;
+
+  constructor({ studio }: { studio?: boolean } = {}) {
     super({ name: 'cloud' });
+    this.studio = studio ?? false;
```
YujohnNattrass, line 19: made this configurable so that we can target ourselves when testing. We need to enforce auth before we make it public otherwise anyone can visit the studio

abhiaiyer91 (approved):

PR #11336, opened 2025-12-22
fix(client-js): fix listMemoryThreads orderBy serialization and type

hanjeahwan:
## Summary

Fixes the `listMemoryThreads` method in `@mastra/client-js` where the `orderBy` parameter was not working correctly.

## Problem

1. **Type mismatch**: The client type defined `orderBy` as a string (`'createdAt' | 'updatedAt'`) with a separate `sortDirection` field, but the server expects an object: `{ field: 'createdAt' | 'updatedAt', direction: 'ASC' | 'DESC' }`

2. **Serialization bug**: When passing an object to `orderBy`, it was being converted to `[object Object]` in the query string because `URLSearchParams` calls `.toString()` on values

## Changes

- Updated `ListMemoryThreadsParams.orderBy` type to match server schema:
  ```typescript
  orderBy?: {
    field?: 'createdAt' | 'updatedAt';
    direction?: 'ASC' | 'DESC';
  };
  ```
- Removed unused `sortDirection` field
- Added `JSON.stringify()` for proper serialization of `orderBy` in query params

## Usage

```typescript
const result = await client.listMemoryThreads({
  resourceId,
  agentId,
  page: 1,
  perPage: 20,
  orderBy: { field: "createdAt", direction: "DESC" },
});
```

## Testing

- All existing client SDK tests pass (`pnpm test:clients`)

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed serialization of ordering parameters so ordering is transmitted correctly to the API.
* **Breaking Changes**
  * The ordering parameter shape has changed: it now uses a nested object with `field` and `direction` (previous separate fields removed).

### DIFF
--- a/.changeset/fix-listmemorythreads-orderby-serialization.md
+++ b/.changeset/fix-listmemorythreads-orderby-serialization.md
@@ -1,0 +1,5 @@
+---
+'@mastra/client-js': patch
+---
+
+Fix listMemoryThreads orderBy parameter serialization and type definition. The `orderBy` parameter now correctly expects an object with `field` and `direction` properties matching the server schema, and is properly JSON-serialized in the query string instead of being converted to `[object Object]`.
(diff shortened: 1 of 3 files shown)


wardpeet:
going to close it anyway cause it's a breaking change

PR #11452, opened 2025-12-27
fix(core): Support AI SDK v6 (v3) models in sub-agent calls

dev-vishnu:
## Problem

When using AI SDK v6 models (`specificationVersion: "v3"`) with sub-agents through the `agents` property, the version check in `listAgentTools` only matched "v2", causing v3 models to incorrectly route to legacy methods and throw:

```
V2 models are not supported for streamLegacy. Please use stream instead.
```

## Root Cause

In `agent.ts`, the `listAgentTools` method checks `modelVersion === 'v2'` but doesn't handle `'v3'`, causing v3 models to fall through to the legacy `streamLegacy()` path.

## Solution

Updated version checks to use the centralized `supportedLanguageModelSpecifications` array from `utils.ts` which already includes both v2 and v3:

```typescript
// Before (bug)
if (modelVersion === 'v2') { ... }

// After (fix)
if (supportedLanguageModelSpecifications.includes(modelVersion)) { ... }
```

## Changes

- `agent.ts`: Updated lines 1863 and 1889 to use `supportedLanguageModelSpecifications.includes()`
- `llm-mapping-step.ts`: Updated line 177 for `onChunk` callback
- `tool-handling.test.ts`: Added missing `toolhandlingTests('v3')` call
- `mock-model.ts`: Added v3 support to `getOpenAIModel()` helper

## Testing

- Existing v3 tests pass
- Added v3 test coverage for sub-agent tool handling

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed v3 language model detection in nested/sub-agent flows so v3 models no longer fall back to legacy streaming and no longer trigger related errors.

* **Tests**
  * Added v3-focused test coverage to tool handling to prevent regressions and ensure correct behavior across model versions.


abhiaiyer91 (approved):

abhiaiyer91:
Amazing find @dev-vishnu thank you!

PR #11533, opened 2026-01-02
feat(otel-exporter): add flush() method to force export span

### LINKED ISSUE #11372: [FEATURE] Flushing of AI tracing exporters
### Problem Statement

We use a hybrid serverless environment (vercel's fluid compute) and we have to flush spans before the instance is terminated, so we ensure all spans are properly exported. We currently use mastra's official `OtelExporter`.

The thing is, we cannot call `shutdown` to cleanup resources, as the runtime instance can be reused for another request in the future and the span exporter must be operative. But, there is no other way of flushing things as of right now.

Otel processors have `shutdown` and `forceFlush` method to handle this use case -> https://opentelemetry.io/docs/specs/otel/trace/sdk/#forceflush

### Proposed Solution

I propose the addition of a `flush` method to the `BaseExporter` class, and the implementation of that method in `OtelExporter` that simply calls `this.processor.forceFlush()`, in a similar way to https://github.com/mastra-ai/mastra/blob/b56d3858fbb285cfd0ec33a950c9acad577a5578/observability/otel-exporter/src/tracing.ts#L216

### Component

Other

### Alternatives Considered

_No response_

### Example Use Case

force flushing in serverless environment

### Additional Context

_No response_

### Verification

intojhanurag:
## Description

This PR adds a `flush()` method to the Mastra AI observability exporters, allowing manual flushing of spans without shutting down the exporter. This is critical for serverless environments (like Vercel) where instances may be reused for multiple requests.

Changes include:  
- Updated the `ObservabilityExporter` interface to include an optional `flush(): Promise<void>` method.  
- Added a default no-op `flush()` method in `BaseExporter` for backward compatibility.  
- Implemented `flush()` in `OtelExporter` to call `this.processor.forceFlush()`, ensuring all spans are exported immediately.  
- Created a changeset documenting the minor version bump for affected packages.

## Related Issue(s)

- Closes #11372

## Type of Change

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Observability exporters now support an explicit flush operation to manually flush queued spans; default exporters are no-ops unless implemented, and the OTEL exporter performs an immediate flush when invoked.

* **Chores**
  * Bumped observability and exporter packages to minor versions.

### DIFF
--- a/observability/otel-exporter/src/tracing.ts
+++ b/observability/otel-exporter/src/tracing.ts
@@ -112,7 +112,7 @@ export class OtelExporter extends BaseExporter {
         } catch (grpcError) {
           this.logger.error(
             `[OtelExporter] Failed to load gRPC metadata. Install required packages:\n` +
-              `  npm install @opentelemetry/exporter-trace-otlp-grpc @grpc/grpc-js\n`,
+            `  npm install @opentelemetry/exporter-trace-otlp-grpc @grpc/grpc-js\n`,
             grpcError,
           );
           this.isDisabled = true;
@@ -210,6 +210,12 @@ export class OtelExporter extends BaseExporter {
       );
     } catch (error) {
       this.logger.error(`[OtelExporter] Failed to export span ${span.id}:`, error);
+    }
+  }
+
+  async flush(): Promise<void> {
+    if (this.processor) {
+      await this.processor.forceFlush();
     }
   }
 
--- a/observability/otel-exporter/src/tracing.ts
+++ b/observability/otel-exporter/src/tracing.ts
@@ -112,7 +112,7 @@ export class OtelExporter extends BaseExporter {
         } catch (grpcError) {
           this.logger.error(
             `[OtelExporter] Failed to load gRPC metadata. Install required packages:\n` +
-              `  npm install @opentelemetry/exporter-trace-otlp-grpc @grpc/grpc-js\n`,
+            `  npm install @opentelemetry/exporter-trace-otlp-grpc @grpc/grpc-js\n`,
             grpcError,
           );
           this.isDisabled = true;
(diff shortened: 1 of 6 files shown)


observability/otel-exporter/src/tracing.ts
```diff
@@ -213,6 +213,23 @@ export class OtelExporter extends BaseExporter {
...
+      return;
+    }
+    if (this.processor) {
+      try {
+        await this.processor.forceFlush();
+      } catch (error) {
+        this.logger.error(`[OtelExporter] Failed to flush spans:`, error);
+      }
+    }
+  }
```
wardpeet, line 231: Add similar style as above 
```suggestion
    // Skip if disabled
    if (this.isDisabled || !this.processor) {
      return;
    }

    try {
      await this.processor.forceFlush();
    } catch (error) {
      this.logger.error(`[OtelExporter] Failed to flush spans:`, error);
    }
  }
```

wardpeet (approved):

wardpeet:
@intojhanurag they are, seems like we need to make flush a protected fn
```
> tsup --silent --config tsup.config.ts

Error: src/exporters/cloud.ts(43,14): error TS2415: Class 'CloudExporter' incorrectly extends base class 'BaseExporter'.
  Property 'flush' is private in type 'CloudExporter' but not in type 'BaseExporter'.
Error: src/exporters/default.ts(90,14): error TS2415: Class 'DefaultExporter' incorrectly extends base class 'BaseExporter'.
  Property 'flush' is private in type 'DefaultExporter' but not in type 'BaseExporter'.
 ELIFECYCLE  Command failed with exit code 2.
```

intojhanurag:
Hey @wardpeet , it is ready now .

PR #11561, opened 2026-01-04
fix(core): respect sub-agent defaultOptions.memory when called as tool

dev-vishnu:
## Summary

Fixes a bug where sub-agent's `defaultOptions.memory` configuration is overridden when called as a tool from a parent agent.

## Problem

When a sub-agent is called via the `agents: {}` option, its `defaultOptions.memory` is unconditionally replaced if the parent has `resourceId && threadId`:

```typescript
// Before: Always overrides sub-agent's memory config
...(resourceId && threadId
  ? { memory: { resource: subAgentResourceId, thread: subAgentThreadId } }
  : {}),
```

### Impact
- Sub-agents lose their configured memory isolation strategy
- Memory continuity across multiple tool calls to the same sub-agent is broken
- `defaultOptions` function runs but its `memory` value is silently discarded
- Sub-agents work correctly when called directly, but break when used via `agents: {}`

## Reproduction

```typescript
const ChildAgent = new Agent({
  id: 'child-agent',
  defaultOptions: ({ requestContext }) => ({
    memory: {
      resource: requestContext.get('resourceId'),
      thread: uuidv4(), // Intended custom thread
    }
  }),
  memory: new Memory({ ... })
});

const ParentAgent = new Agent({
  agents: { child: ChildAgent }, // Child's memory config ignored
});
```

## Solution

Check if the sub-agent has its own `defaultOptions.memory` before applying parent-derived settings:

```typescript
const subAgentDefaultOptions = await agent.getDefaultOptions?.({ requestContext });
const subAgentHasOwnMemoryConfig = subAgentDefaultOptions?.memory !== undefined;

...(resourceId && threadId && !subAgentHasOwnMemoryConfig
  ? { memory: { resource: subAgentResourceId, thread: subAgentThreadId } }
  : {}),
```

## Changes

- `packages/core/src/agent/agent.ts` - Check sub-agent's `defaultOptions.memory` before overriding (both generate and stream paths)

## Testing

- Ran `pnpm test:core` - all tests pass
- No breaking changes for existing consumers

## Summary by CodeRabbit

* **Bug Fixes**
  * Sub-agent memory handling corrected so sub-agents with explicit memory settings retain their own configuration; parent agent memory no longer unintentionally overrides them. Sub-agents without explicit memory still receive parent-derived IDs as a fallback.

* **Tests**
  * Added test coverage for sub-agent creation and tool handling across various defaultOptions configurations.


wardpeet (approved):
Looks good. Thank you for finding this bug and fixing it!

Any chance you can add a test?

PR #11625, opened 2026-01-05
fix(evals): remove unnecessary ai peer dependency

roaminro:
The `@mastra/evals` package had a peer dependency on `ai` just for the `ToolInvocation` type import. This was only used to type the `state` parameter in `createToolInvocation()`.

Replaced it with an inline string literal union type:

```ts
// before
import type { ToolInvocation } from 'ai';
state?: ToolInvocation['state'];

// after
state?: 'call' | 'partial-call' | 'result';
```

This removes an unnecessary peer dependency for users of `@mastra/evals`.

## Summary by CodeRabbit

* **Chores**
  * Removed the external "ai" dependency from the evals package so users no longer need to install it.
  * Included the necessary types internally to preserve existing behavior and compatibility.
  * Performed a minor internal type refinement with no user-facing API changes.
  * Reduced dependency surface and package size to simplify setup and avoid version conflicts.


packages/evals/src/scorers/utils.ts
```diff
@@ -348,7 +347,7 @@ export const createToolInvocation = ({
   toolName: string;
   args: Record<string, any>;
   result: Record<string, any>;
-  state?: ToolInvocation['state'];
+  state?: 'call' | 'partial-call' | 'result';
```
DanielSLew, line 350: don't we have this type exported already from somewhere in mastra core? we should reuse the type instead of creating it again here inline

DanielSLew (approved):

packages/evals/src/scorers/utils.ts
roaminro: the type doesn't seem to be re-exported from @mastra/core, also this utils is only used for the tests here, so I think that's ok to create it inline?

PR #11736, opened 2026-01-08
fix(braintrust): include tool calls in LLM span output for Thread view [#11735]

### LINKED ISSUE #11735: [BUG] Thread view truncated for last turn in a conversation in Braintrust
### Describe the Bug

When in the thread view on Braintrust, the view shows the all the message history fed to the model and the final response. Is omits the full tool call and model response history and doesn't reflect the trace view.

Image

For multi turn, we see the full message history for all N-1 turns, but turn N only has the final response.

I have prepared a fix and will be submitting it shortly.

### Steps To Reproduce

Run an agent that executes tool calls and exports to BT. Open the trace in thread view and observe.

N/A

### Expected Behavior

Image

### Environment Information

```block
Installed from head 2025-01-08
```

### Verification

ctamulonis-glossy:
## Summary

Fixes the Braintrust Thread view to display the complete conversation history including tool calls and results.

**Problem**: The LLM span output only included the final text, missing intermediate tool calls and results needed for the Thread view.

### 1. Accumulate step data in `ModelSpanTracker` (`observability/mastra/src/model-tracing.ts`)

- Added `#steps` array to collect completed `MODEL_STEP` spans
- Added `#currentStepToolResults` to accumulate `tool-result` chunks per step
- On `step-finish`: include `toolResults` in step output, push to `#steps`
- On `endGeneration`: attach `modelSteps` array to `MODEL_GENERATION` span output

### 2. Transform to OpenAI format in Braintrust exporter (`observability/braintrust/src/tracing.ts`)

- Updated `transformOutput` to detect `modelSteps` in `MODEL_GENERATION` spans
- For each step, convert `toolCalls` → assistant message with `tool_calls`
- Convert `toolResults` → tool messages with `tool_call_id`
- Append final `text` as assistant content

### Result

The LLM span output is now an array of OpenAI-format messages that Braintrust renders correctly in Thread view:

```json
[
  { "role": "assistant", "content": "", "tool_calls": [...] },
  { "role": "tool", "content": "...", "tool_call_id": "..." },
  { "role": "assistant", "content": "Final response" }
]
```

## Testing

- Added unit tests for `modelSteps` accumulation with `toolResults`
- Added unit test for Braintrust message reconstruction from step outputs

Fixes #11735

## Summary by CodeRabbit

* **New Features**
  * Improved conversion of AI SDK messages to OpenAI-like thread format with tool_call and tool_result support.
  * Accumulation of model generation steps and tool results for richer trace reconstruction.

* **Bug Fixes**
  * Fixed LLM output reconstruction for Thread view (issue #11735).

* **Tests**
  * Added tests for message reconstruction, tool-result accumulation, and more flexible mock span construction.

### DIFF
--- a/observability/braintrust/src/tracing.test.ts
+++ b/observability/braintrust/src/tracing.test.ts
@@ -637,6 +637,86 @@ describe('BraintrustExporter', () => {
         }),
       );
     });
+
+    /**
+     * Test for GitHub issue #11735: Thread view truncated for last turn in Braintrust
+     *
+     * When an LLM generation has steps with messages, the output should be
+     * converted to OpenAI format so Braintrust's Thread view displays properly.
+     *
+     * Each step's output contains text, toolCalls, and messages (AI SDK format).
+     *
+     * @see https://github.com/mastra-ai/mastra/issues/11735
+     */
+    it('should reconstruct LLM output from steps for Thread view (issue #11735)', async () => {
+      const traceId = 'step-output-reconstruction-trace';
+
+      // Create the MODEL_GENERATION span with steps
+      // Each step has output containing text, toolCalls, and toolResults
+      const llmSpan = createMockSpan({
+        id: 'model-gen-span',
+        traceId,
+        name: 'gpt-4-call',
+        type: SpanType.MODEL_GENERATION,
+        isRoot: true,
+        input: [{ role: 'user', content: 'What is 2+2?' }],
+        output: {
+          text: 'The answer is 4.',
+          // Steps accumulated by ModelSpanTracker (named 'modelSteps' to avoid deepClean stripping)
+          modelSteps: [
+            {
+              output: {
+                text: '',
+                toolCalls: [{ toolName: 'calculator', toolCallId: 'tc-1', args: { a: 2, b: 2 } }],
+                toolResults: [{ toolCallId: 'tc-1', toolName: 'calculator', result: { result: 4 } }],
+              },
+            },
+            {
+              output: {
+                text: 'The answer is 4.',
+              },
+            },
+          ],
+        },
+        attributes: { model: 'gpt-4', provider: 'openai' },
+      });
+
+      // Start MODEL_GENERATION span
+      await exporter.exportTracingEvent({
+        type: TracingEventType.SPAN_STARTED,
+        exportedSpan: llmSpan,
+      });
+
+      // Reset mock to capture only the end call
+      mockSpan.log.mockClear();
+
+      // End MODEL_GENERATION span - this should convert messages from steps
+      await exporter.exportTracingEvent({
+        type: TracingEventType.SPAN_ENDED,
+        exportedSpan: { ...llmSpan, endTime: new Date() },
+      });
+
+      // The output should be converted from AI SDK format to OpenAI format
+      expect(mockSpan.log).toHaveBeenCalledWith(
+        expect.objectContaining({
+          output: [
+            {
+              role: 'assistant',
+              content: '',
+              tool_calls: [
+                {
+                  id: 'tc-1',
+                  type: 'function',
+                  function: { name: 'calculator', arguments: '{"a":2,"b":2}' },
+                },
+              ],
+            },
+            { role: 'tool', content: '{"result":4}', tool_call_id: 'tc-1' },
+            { role: 'assistant', content: 'The answer is 4.' },
+          ],
+        }),
+      );
+    });
   });
 
   describe('AI SDK v5 Message Conversion', () => {
@@ -2275,6 +2355,8 @@ function createMockSpan({
   output,
   errorInfo,
   tags,
+  traceId,
+  parentSpanId,
 }: {
   id: string;
   name: string;
@@ -2286,6 +2368,8 @@ function createMockSpan({
   output?: any;
   errorInfo?: any;
   tags?: string[];
+  traceId?: string;
+  parentSpanId?: string;
 }): AnyExportedSpan {
   const mockSpan = {
     id,
@@ -2299,11 +2383,11 @@ function createMockSpan({
     tags,
     startTime: new Date(),
     endTime: undefined,
-    traceId: isRoot ? `${id}-trace` : 'parent-trace-id',
+    traceId: traceId ?? (isRoot ? `${id}-trace` : 'parent-trace-id'),
     get isRootSpan() {
       return isRoot;
     },
-    parentSpanId: isRoot ? undefined : 'parent-id',
+    parentSpanId: parentSpanId ?? (isRoot ? undefined : 'parent-id'),
     isEvent: false,
   } as AnyExportedSpan;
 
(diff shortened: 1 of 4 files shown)


epinzur (changes requested):
Thanks for the PR! I appreciate your help trying to resolve these open Braintrust issues.

However, I have some concerns about the approach of adding the full messages array to the MODEL_GENERATION span output:

### Data Volume

Adding `messageList.get.response.aiV5.model()` to every MODEL_GENERATION span output could significantly increase data volume, especially for multi-step agent interactions. For conversations with many tool calls, this array would include:
- Multiple assistant messages with tool call requests
- All tool result messages  
- Intermediate and final text responses

This data compounds across all traced generations in a session.

### Data Duplication

The tool call details are already captured in dedicated `TOOL_CALL` child spans, each with their full input arguments and output results. The span hierarchy already contains:
- `MODEL_GENERATION` (parent) → input messages, text output, usage stats
- `TOOL_CALL` (children) → individual tool inputs/outputs with metadata

Adding messages to the parent duplicates what's available in the child spans.

### Format Coupling

Including `aiV5.model()` format in the core tracing output couples the observability pipeline to a specific AI SDK message format. The tracing layer should ideally remain format-agnostic, with format transformations handled by individual exporters based on their target platform's requirements.

### Suggested Alternative

Could the Braintrust exporter reconstruct the conversation thread from the existing span hierarchy instead? The exporter already has access to:
1. Input messages on the `MODEL_GENERATION` span
2. Tool call inputs/outputs from child `TOOL_CALL` spans
3. The final text output

This would keep the core tracing pipeline lean while giving Braintrust (and potentially other exporters) the flexibility to construct whatever view they need from the span tree. It also avoids adding Braintrust-specific requirements to the core framework.

Happy to discuss further if there are constraints with Braintrust's API that make the span-reconstruction approach difficult!

ctamulonis-glossy:
Thanks for the feedback!
 
> ### Data Volume
> Adding `messageList.get.response.aiV5.model()` to every MODEL_GENERATION span output could significantly increase data volume, especially for multi-step agent interactions. For conversations with many tool calls, this array would include:
> 
> * Multiple assistant messages with tool call requests
> * All tool result messages
> * Intermediate and final text responses
> 
> This data compounds across all traced generations in a session.

The volume should be relatively low, as we are only adding output to the single MODEL_GENERATION span. Also, I think this extra data makes the input and output fields consistent: the input field contains every message loaded from memory at the start of the turn (text, tool calls and tool results), and now the output also has the corresponding messages generated by the agent for the turn.


> ### Data Duplication
> The tool call details are already captured in dedicated `TOOL_CALL` child spans, each with their full input arguments and output results. The span hierarchy already contains:
> 
> * `MODEL_GENERATION` (parent) → input messages, text output, usage stats
> * `TOOL_CALL` (children) → individual tool inputs/outputs with metadata
> 
> Adding messages to the parent duplicates what's available in the child spans.

This is true, but I don't see a way around it. Braintrust doesn't seem to look at the child spans when constructing the Thread view.

> ### Format Coupling
> Including `aiV5.model()` format in the core tracing output couples the observability pipeline to a specific AI SDK message format. The tracing layer should ideally remain format-agnostic, with format transformations handled by individual exporters based on their target platform's requirements.
> 
> ### Suggested Alternative
> Could the Braintrust exporter reconstruct the conversation thread from the existing span hierarchy instead? The exporter already has access to:
> 
> 1. Input messages on the `MODEL_GENERATION` span
> 2. Tool call inputs/outputs from child `TOOL_CALL` spans
> 3. The final text output

I re-did the PR to follow this approach. I had to add data to the exported MODEL_GENERATION span, as the Braintrust exporter doesn't seem to have access to the full tree. I can see parent span ids, but not children, and there doesn't seem to be a way to fetch the span given an id.

epinzur (changes requested):
Thanks for the iteration on this! I appreciate the effort, but I think there may have been a misunderstanding about the suggested approach.

### Current Implementation

The changes still modify the core `@mastra/observability` package to accumulate step data (`#steps`, `#currentStepToolResults`) and inject `modelSteps` into the MODEL_GENERATION span output. This means:

1. **Core tracing changes for exporter-specific needs** - The `ModelSpanTracker` is being modified to accommodate Braintrust's Thread view requirements, but other exporters don't need this data structure.

2. **Data duplication remains** - Tool results now exist in both:
   - TOOL_CALL child spans (existing)
   - `modelSteps[].output.toolResults` (new)

3. **Data volume still increases** - Every MODEL_GENERATION span now carries the accumulated step outputs.

### What I Had in Mind

The suggestion was for the Braintrust exporter itself to reconstruct messages from the **existing** span hierarchy, without any changes to core tracing. The exporter already:
- Receives `span_started`/`span_ended` events for all spans including TOOL_CALL children
- Maintains a `traceMap` with `spans: Map<string, Span>` tracking all spans per trace
- Has access to parent-child relationships via `parentSpanId`

When a MODEL_GENERATION span ends, the exporter could look up its child TOOL_CALL spans from the trace's span map and reconstruct the conversation from those.

### Timing Consideration

I'm currently working on a refactor in the `esp/base_exporter` branch that changes how the local span cache functions in exporters. This type of span-hierarchy-based reconstruction would actually be easier to implement cleanly after that work merges.

I'm happy to take a crack at implementing this approach once that foundation is in place. (Probably Wednesday) I'll use your existing tests as the spec/requirements for success—they do a good job of capturing the expected OpenAI message format output.

What do you think?

ctamulonis-glossy:
> Thanks for the iteration on this! I appreciate the effort, but I think there may have been a misunderstanding about the suggested approach.
> 
> ### Current Implementation
> The changes still modify the core `@mastra/observability` package to accumulate step data (`#steps`, `#currentStepToolResults`) and inject `modelSteps` into the MODEL_GENERATION span output. This means:
> 
> 1. **Core tracing changes for exporter-specific needs** - The `ModelSpanTracker` is being modified to accommodate Braintrust's Thread view requirements, but other exporters don't need this data structure.
> 2. **Data duplication remains** - Tool results now exist in both:
>    
>    * TOOL_CALL child spans (existing)
>    * `modelSteps[].output.toolResults` (new)
> 3. **Data volume still increases** - Every MODEL_GENERATION span now carries the accumulated step outputs.
> 
> ### What I Had in Mind
> The suggestion was for the Braintrust exporter itself to reconstruct messages from the **existing** span hierarchy, without any changes to core tracing. The exporter already:
> 
> * Receives `span_started`/`span_ended` events for all spans including TOOL_CALL children
> * Maintains a `traceMap` with `spans: Map<string, Span>` tracking all spans per trace
> * Has access to parent-child relationships via `parentSpanId`
> 
> When a MODEL_GENERATION span ends, the exporter could look up its child TOOL_CALL spans from the trace's span map and reconstruct the conversation from those.
> 
> ### Timing Consideration
> I'm currently working on a refactor in the `esp/base_exporter` branch that changes how the local span cache functions in exporters. This type of span-hierarchy-based reconstruction would actually be easier to implement cleanly after that work merges.
> 
> I'm happy to take a crack at implementing this approach once that foundation is in place. (Probably Wednesday) I'll use your existing tests as the spec/requirements for success—they do a good job of capturing the expected OpenAI message format output.
> 
> What do you think?

I tried to use the traceMap, but it stores Braintrust spans, which are write-only objects.  I could have created a separate one, but thought this would be cleaner.

I'm happy to hand this over to you, esp since you are working on some changes that will make this more straight forward. Feel free to close this PR when it's no longer needed.

Issue #10896, opened 2025-12-04
[FEATURE] Add eval scoring support to all exporters


epinzur:
### Problem Statement

When I eval a trace or span in mastra, I want the score to show up on my langfuse/braintrust/etc... traces and spans.

### Proposed Solution

Design and build a new system for passing Eval data through the observability event stack. Update the exporters to consume this data, and save scores to traces & spans.

### Component

Other

### Example Use Case

I want to see my mastra evals in braintrust/langfuse/arize.

### Additional Context

We already do this for langfuse, but we are circumventing the observability pipeline. We should refactor and upgrade observability to natively handle evals and scores. 

Also consider if the design should include dataset/experimentation support.

### Verification

- [x] I have searched the existing issues to make sure this is not a duplicate
- [x] I have provided sufficient context for the team to understand the request

zijin1994:
@epinzur We made a initial attempt of this. Now the default exporter + playground would see scores on spans, and they are persistent for the default exporter as well. If this is the right direction, we will start implementing exporters logics (maybe langfuse first and then datadog once that gets merged?)

ps: this is how it looks atm (inside span details):

<img width="832" height="794" alt="Image" src="https://github.com/user-attachments/assets/2681440a-de50-4e89-91c5-fd664fcf93db" />

epinzur:
@zijin1994 Actually, I made a goof in my last comment.

We already have a UI for Scores on spans.  You don't need to create any new UI or experience in studio/playground for scores on spans.  The existing system is sufficient.

But we do want to change the backend path to send the scores through the observability system when going out to mastra storage (via DefaultExporter) or to mastra API (via CloudExporter).

We also may need a way to save a score on a **span** vs on a **trace**. I think we currently allow both of these scenarios in the system. You can confirm with @YujohnNattrass

intojhanurag:
Going to work on this now :)

PR #11774, opened 2026-01-09
feat(observability): Added eval scores to spans (Integrated with Datadog/langfuse/default exporters) #10896

zijin1994:
## Description

Add eval scores as part of the spans so that they flow through the span processor pipeline. The scores are attached to spans when scorer hook function gets called, and gets filtered, exported like other properties on spans.
We also made default exporter stores scores and removed the direct database call in the scorer hook.

Default/Langfuse/Datadog exporters now process scores that are attached to spans by listening to `span_update` events that has an non-empty span.scores field. Default exporter then stores scores into the database, while Langfuse/Datadog export them using SDK methods.

For Cloud exporter, we weren't able to find an api endpoint that accepts exported scores. Therefore for now, scores are sent along with spans and are directly attached to spans as a property.

## Related Issue(s)
#10896 

## Type of Change

- [ ] Documentation update
- [x] Code refactoring
- [ ] Performance improvement
- [x] Test update

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Span scoring: attach multiple evaluation scores (value, reason, metadata, timestamp, scorer) to spans; scores now appear in exported span payloads.
  * Exporters submit scores on span updates/ends and handle late/out-of-order score arrivals.
  * Added label sanitization for metric/tag submissions.

* **Bug Fixes**
  * Late-arriving scores are persisted and submitted instead of being skipped.

* **Tests**
  * Extensive tests for scoring flows, exporter submissions, buffering/late updates, and redaction of sensitive score metadata.


observability/mastra/src/spans/base.ts
```diff
@@ -139,6 +141,8 @@ export abstract class BaseSpan<TType extends SpanType = any> implements Span<TTy
   protected parentSpanId?: string;
   /** Deep clean options for serialization */
   protected deepCleanOptions: DeepCleanOptions;
+  /** Scores attached to this span */
+  public scores: SpanScore[] = [];
```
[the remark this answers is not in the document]
zijin1994: I don't think client.score() needs to be awaited. SDK handles batching/queueing behind the scene. Their examples don't seem to await this.
Also looks like langfuse has a newer version of sdk, where packages are modularized. Mastra is still using the older version atm, might be something worth upgrading in the future.

As for `deepClean`, looks like it only runs on user input/output. scores are internally generated values, and they seem to only run through the sensitive data filter.

zijin1994:
@epinzur This is still in progress, but the skeleton is here and we have implemented scores for default exporter/langfuse/datadog. 
That `span.update` call without any arguments passed in is there to notify the exporters. Since scorers run async after `agent run` span ended, we need this update event to inform exporters that scores are updated.

observability/langfuse/src/tracing.test.ts
```diff
@@ -1247,122 +1249,174 @@ describe('LangfuseExporter', () => {
...
+      // When client is disabled, nothing should happen
+      await disabledExporter.exportTracingEvent({
+        type: TracingEventType.SPAN_STARTED,
+        exportedSpan: spanWithScores,
       });
 
-      mockLoggerError.mockRestore();
+      expect(mockLangfuseClient.score).not.toHaveBeenCalled();
     });
   });
```
epinzur, line 1702: for `langfuse` only, we should probably keep the old tests for `addScoreToTrace()` for a while... and also keep the `addScoreToTrace()` method in langfuse but mark it deprecated. 

If the method calls the same method as the new stuff, that works too.

observability/langfuse/src/tracing.ts
```diff
@@ -509,41 +515,34 @@ export class LangfuseExporter extends BaseExporter {
     return payload;
   }
 
-  async addScoreToTrace({
```
epinzur, line 330: make sure to keep the old method around but mark it deprecated.

It can just be a wrapper around the new method.

observability/mastra/src/exporters/default.ts
```diff
@@ -235,8 +235,22 @@ export class DefaultExporter extends BaseExporter {
...
+              traceId: span.traceId,
+              spanId: span.id,
+              updates: this.buildUpdateRecord(span),
+            }).catch(error => {
+              this.logger.error('Failed to update span with scores (out-of-order)', {
+                error: error instanceof Error ? error.message : String(error),
+                spanId: span.id,
+                traceId: span.traceId,
+              });
+            });
```
epinzur: im confused why this exists... Can you explain it more?

observability/mastra/src/spans/base.test.ts
```diff
@@ -536,6 +536,170 @@ describe('Span', () => {
...
+
+      span.end();
+
+      const endEvent = testExporter.events.find(e => e.type === 'span_ended');
+      expect(endEvent).toBeDefined();
+      expect(endEvent?.exportedSpan.scores).toHaveLength(1);
+      expect(endEvent?.exportedSpan.scores?.[0].scorerName).toBe('quality-scorer');
+    });
+  });
+
```
epinzur, line 764: test cases look good.

observability/mastra/src/exporters/default.ts
zijin1994: For cases where the evals arrive late (span_ended => scorer runs, takes few seconds + => previously ended spans already sent), without this default exporter would generate out of order span warning and skipping the span_update event that has the scores attached to span. This makes that these score updates get registered into the database instead of being ignored(scores on span in this case).

packages/core/src/observability/types/tracing.ts
```diff
@@ -368,6 +368,42 @@ export interface SpanTypeMap {
...
+  /** ID of the scorer that generated this score */
+  scorerId: string;
+  /** Display name of the scorer (defaults to scorerId if not provided) */
+  scorerName: string;
+  /** Numeric score value */
+  score: number;
+  /** Optional explanation for the score */
+  reason?: string;
+  /** Scorer-specific metadata from runResult */
+  metadata?: Record<string, any>;
```
epinzur: metadata should be `Record<string, unknown>`

packages/core/src/observability/types/tracing.ts
```diff
@@ -368,6 +368,42 @@ export interface SpanTypeMap {
...
+export interface AddScoreArgs {
+  /** ID of the scorer */
+  scorerId: string;
+  /** Display name of the scorer (defaults to scorerId if not provided) */
+  scorerName?: string;
+  score: number;
+  reason?: string;
+  /** Scorer-specific metadata (from runResult, not span metadata) */
+  metadata?: Record<string, any>;
+}
```
epinzur: consider changing these types to be `zod` based. See the new `zod` based schemas in observability storage.

https://github.com/mastra-ai/mastra/blob/main/packages/core/src/storage/domains/observability/types.ts

I think this is generally a much better way to build types up. It locks the description to code instead of comments.

packages/core/src/observability/types/tracing.ts
```diff
@@ -429,6 +465,12 @@ export interface Span<TType extends SpanType> extends BaseSpan<TType> {
   /** Trace-level state shared across all spans in this trace */
   traceState?: TraceState;
 
+  /** Scores attached to this span */
+  scores: SpanScore[];
+
+  /** Add an evaluation score to this span */
+  addScore(args: AddScoreArgs): void;
+
```
epinzur, line 471: should you also add `traceScores` and `addScoreToTrace` (and reject the method if not on the root span.)?

It is possible to both score a specific span and/or an entire trace.

packages/core/src/observability/types/tracing.ts
```diff
@@ -565,6 +607,8 @@ export interface ExportedSpan<TType extends SpanType> extends BaseSpan<TType> {
    * Tags are string labels used to categorize and filter traces.
    */
   tags?: string[];
+  /** Scores attached to this span */
+  scores?: SpanScore[];
```
epinzur, line 609: potentially add `traceScores` as well.  Or make the type `SpanScore[] | TraceScore[]`

packages/core/src/observability/types/tracing.ts
epinzur: some third-party observability tools may only support one of these types.

packages/core/src/storage/domains/observability/types.ts
```diff
@@ -60,6 +60,17 @@ const inputField = z.unknown().describe('Input data passed to the span');
...
+/** Schema for evaluation scores attached to spans */
+const spanScoreSchema = z.object({
+  scorerId: z.string().describe('ID of the scorer that generated this score'),
+  scorerName: z.string().describe('Display name of the scorer (defaults to scorerId if not provided)'),
+  score: z.number().describe('Numeric score value'),
+  reason: z.string().optional().describe('Optional explanation for the score'),
+  metadata: z.record(z.unknown()).optional().describe('Scorer-specific metadata'),
+  timestamp: z.number().describe('Timestamp when score was added'),
+});
+const scoresField = z.array(spanScoreSchema).describe('Evaluation scores attached to this span');
```
epinzur, line 64: potentially move this type up to `@core/mastra/observability` and use everywhere.

observability/mastra/src/exporters/default.ts
epinzur: sending scores via `span_update` event might not be the best approach.

we might want a new event entirely.  

several of our exporters ignore `update` events entirely.  

even the defaultExporter ignores `update` events when storage is ClickHouse.

epinzur: also for defaultExporter, scores aren't actually stored in the span on the DB... they are stored in a separate table and linked back to spans.

zijin1994: We noticed that for defaultExporter, scores are stored separately, but we still notified the exporter about the attachment update event and stored the span.scores. If this is not needed we can remove that.

packages/core/src/observability/types/tracing.ts
zijin1994: Currently mastra scorer hooks only run against `agent run` spans, so `addScoreToTrace` implementation would require adding a configuration that changes when hooks run.

.changeset/odd-yaks-peel.md
```diff
@@ -0,0 +1,14 @@
...
+'@mastra/core': minor
+---
+
+Added scores to spans so that they flow through the observability pipeline and are accessible for all the registered exporters.
+
+An `addScore` function was added onto the spans. Scores from live evaluations now flow through the observability pipeline. The scorer hook now calls `currentSpan.addScore()` to attach scores to spans, which then emits a SPAN_UPDATED event. This allows all configured exporters to receive and process scores.
+
+DefaultExporter now handles score persistence by listening for scores on span.scores and saving them to the scores database.
+
+LangfuseExporter and DatadogExporter now process scores from span.scores and submits them using the SDK's score API.
```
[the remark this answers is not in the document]
zijin1994: addScore is meant to be an automatic process, not an api feature.

observability/datadog/src/tracing.test.ts
```diff
@@ -1363,4 +1363,284 @@ describe('DatadogExporter', () => {
...
+        scores: [
+          {
+            scorerId: 'scorer-1',
+            scorerName: 'Relevance',
+            score: 0.85,
+            reason: 'Good relevance',
+            timestamp: Date.now(),
+          },
+        ],
+      });
```
epinzur, line 1381: I don't think we will have scores at `span_end` time.  I'd expect scores to be added sometime after the span ends.

PR #11816, opened 2026-01-12
Make Datadog exporter zero config compatible

epinzur:
## Description

Similar to recent changes in other exporters:

The Datadog exporter can now be instantiated without any configuration by reading credentials from environment variables:

- `DD_LLMOBS_ML_APP` - ML application name
- `DD_API_KEY` - Datadog API key
- `DD_SITE` - Datadog site (defaults to `datadoghq.com`)
- `DD_ENV` - Environment name

## Type of Change

- [ ] Documentation update
- [X] Code refactoring
- [ ] Performance improvement
- [ ] Test update

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Datadog exporter supports zero-configuration via environment variables (DD_LLMOBS_ML_APP, DD_API_KEY, DD_SITE with default datadoghq.com, DD_ENV).

* **Documentation**
  * Added zero-config setup examples and preserved explicit-configuration guides.

* **Improvements**
  * Environment selection is more consistently applied and validation/error messages were clarified for easier troubleshooting.

### DIFF
--- a/observability/datadog/src/tracing.ts
+++ b/observability/datadog/src/tracing.ts
@@ -158,9 +158,7 @@ export class DatadogExporter extends BaseExporter {
 
     // Validate required configuration
     if (!mlApp) {
-      this.setDisabled(
-        `Missing required mlApp. Set DD_LLMOBS_ML_APP environment variable or pass mlApp in config.`,
-      );
+      this.setDisabled(`Missing required mlApp. Set DD_LLMOBS_ML_APP environment variable or pass mlApp in config.`);
       this.config = config as any;
       return;
     }
--- a/observability/datadog/src/tracing.ts
+++ b/observability/datadog/src/tracing.ts
@@ -146,9 +146,10 @@ export class DatadogExporter extends BaseExporter {
     super(config);
 
     // Resolve configuration from config object and environment variables
-    const mlApp = config.mlApp || process.env.DD_LLMOBS_ML_APP;
-    const apiKey = config.apiKey || process.env.DD_API_KEY;
-    const site = config.site || process.env.DD_SITE || 'datadoghq.com';
+    const mlApp = config.mlApp ?? process.env.DD_LLMOBS_ML_APP;
+    const apiKey = config.apiKey ?? process.env.DD_API_KEY;
+    const site = config.site ?? process.env.DD_SITE ?? 'datadoghq.com';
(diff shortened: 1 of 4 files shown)


observability/datadog/src/tracing.ts
```diff
@@ -157,18 +158,26 @@ export class DatadogExporter extends BaseExporter {
 
     // Validate required configuration
     if (!mlApp) {
-      this.setDisabled('Missing required mlApp (set config.mlApp or DD_LLMOBS_ML_APP)');
+      const mlAppSource = config.mlApp ? 'from config' : process.env.DD_LLMOBS_ML_APP ? 'from env' : 'missing';
```
astout55555: This logic is flawed, and will only ever set `mlAppSource` to 'missing', since to get here the `mlApp` variable above has already been determined to be falsey.

observability/datadog/src/tracing.ts
```diff
@@ -157,18 +158,26 @@ export class DatadogExporter extends BaseExporter {
...
+      );
       this.config = config as any;
       return;
     }
 
     if (agentless && !apiKey) {
-      this.setDisabled('Missing required apiKey (set config.apiKey or DD_API_KEY)');
+      const apiKeySource = config.apiKey ? 'from config' : process.env.DD_API_KEY ? 'from env' : 'missing';
+      this.setDisabled(
+        `Missing required apiKey for agentless mode (apiKey: ${apiKeySource}). ` +
```
astout55555: There's the same logic issue here with the value of `apiKeySource`. So the messages in both blocks won't actually inform us of the source as written.

astout55555 (changes requested):
Thanks for improving the exporter! This seems like a good change to make, overall. The current implementation is already zero config compatible, and relies on environment variables as backup values. However, this change in approach seems worthwhile to have more consistent code and better documentation.

There's a logic issue in the two validation blocks updated in `tracing.ts`, though, setting the value of `mlAppSource` and `apiKeySource`. However, the messages passed to `setDisabled` do seem useful if they can be given the source of the value to include in the message.

One way to implement this functionality would be to update the block logic to use `hasOwnProperty` for `config` and `process.env` to check the source of the value separate from confirming that the value itself is truthy.

astout55555 (approved):
Thanks for the fix, LGTM!

roaminro (approved):

PR #11842, opened 2026-01-12
fix(memory): sync vector database when updating messages

### LINKED ISSUE #6195: [FEATURE]  memory.updateMessage() updates the vector db
### Problem Statement

Semantic recall doesn't update when I use memory.updateMessage(), so my agent's memory is different than the updated messages.

Here's the discord slack thread:
https://discord.com/channels/1309558646228779139/1397589594911801536/1397589594911801536

### Proposed Solution

I need to update the vector db when I update a message using memory.updateMessage() so that semantic recall matches the updated message history.

### Component

Memory

### Alternatives Considered

_No response_

### Example Use Case

As a developer, I need to have consistency between updated messages and semantic recall so that my agent behaves consistently across interactions.

### Additional Context

_No response_

### Verification

roaminro:
Fixes #6195

When calling `memory.updateMessages()`, vector embeddings weren't being updated, so semantic recall would return stale results based on old message content.

Now when you update a message:

```ts
await memory.updateMessages({
  messages: [{
    id: messageId,
    content: { parts: [{ type: 'text', text: 'new content here' }] }
  }]
});
```

The old vector embeddings are deleted and new ones are created for the updated content. Semantic recall will now match the current message content instead of the original.

Also handles clearing content (deletes vectors) and updating multiple messages at once.

## Summary by CodeRabbit

* **New Features**
  * Message update API accepts an optional memory configuration to better control semantic recall behavior.

* **Bug Fixes**
  * Message updates now synchronize the vector database so semantic search reflects edited or cleared content and avoids stale results.

* **Tests**
  * Added integration tests validating vector sync across updates, deletions, and multi-message update scenarios.

### DIFF
--- a/packages/memory/src/index.ts
+++ b/packages/memory/src/index.ts
@@ -1036,7 +1036,11 @@ ${
         }> = [];
         let dimension: number | undefined;
 
-        // Delete old vectors and prepare new embeddings
+        // Track which messages will have new embeddings vs cleared content
+        const messageIdsWithNewEmbeddings = new Set<string>();
+        const messageIdsWithClearedContent = new Set<string>();
+
+        // Prepare new embeddings and track which messages need vector operations
         await Promise.all(
           messagesWithContent.map(async message => {
             const existingMessage = existingMessagesMap.get(message.id);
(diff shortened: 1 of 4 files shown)


packages/memory/src/index.ts
```diff
@@ -998,20 +998,147 @@ ${
...
+                this.logger.debug(
+                  `No existing vectors found for message ${message.id} in ${indexName}, skipping delete`,
+                );
+              }
+            }
+          }
+        } catch {
+          // Ignore errors if no indexes exist yet
+          this.logger.debug(`No memory indexes found to delete from`);
+        }
```
NikAiyer: @roaminro Is there a reason we need to delete vectors even when no content is updated for those?

NikAiyer (approved):

packages/memory/src/index.ts
roaminro: @NikAiyer — made some changes, thanks!

PR #11846, opened 2026-01-13
fix(schema-compat): handle passthrough schemas for OpenAI compatibility

NikAiyer:
## Description

Fixed OpenAI schema validation error when using passthrough schemas with tools like `vectorQueryTool`.

OpenAI rejects tool schemas with `additionalProperties: {}` (empty object), which is what Zod v4's `.passthrough()` and `z.looseObject()` schemas produce. This caused the error:

```
"Invalid schema for function 'vectorQueryTool': In context=('additionalProperties',), 
schema must have a 'type' key."
```

**Changes:**
- `OpenAISchemaCompatLayer` now passes `{ passthrough: false }` to the object handler to prevent preserving passthrough behavior
- `defaultZodObjectHandler` in the v4 layer now skips adding `ZodUnknown` catchall when passthrough is disabled

This converts passthrough schemas to strict object schemas for OpenAI, producing valid `additionalProperties: false` instead of `{}`.

## Related Issue(s)

Fixes #11823

## Type of Change

- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [ ] Test update

## Checklist

---

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed OpenAI schema validation errors when using passthrough/loose-object patterns (including partial().passthrough()) with tools like vectorQueryTool. These schemas are now converted to OpenAI-compatible JSON Schema forms (valid additionalProperties shapes), preventing previously seen validation failures.


packages/schema-compat/src/provider-compats/openai.ts
```diff
@@ -89,7 +89,10 @@ export class OpenAISchemaCompatLayer extends SchemaCompatLayer {
 
       return value;
     } else if (isObj(z)(value)) {
-      return this.defaultZodObjectHandler(value);
+      // OpenAI doesn't support additionalProperties: {} (empty object)
+      // which is what Zod v4's looseObject/passthrough produces.
+      // Setting passthrough: false converts these to strict objects instead.
+      return this.defaultZodObjectHandler(value, { passthrough: false });
```
DanielSLew: Does this remove any existing valid object passthroughs? Because a passthrough can also be keyed instead of just a boolean

NikAiyer: It should retain valid object passthroughs. I'll add a test to confirm.

DanielSLew (approved):

PR #11885, opened 2026-01-13
Inngest observability fixes

rase-:
# Inngest Workflow Tracing Fix Report

## Solution: Deferred Span Collection

Instead of creating spans during execution (which happens on every Inngest replay), we now:

1. **Collect span metadata during execution** - Using a SpanCollector class
2. **Create all spans in finalize step** - With proper parent-child relationships
3. **Use memoization for timing** - Span start times captured via step.run()

### Architecture

```
During Execution:
  - proxyWorkflowSpan.createChildSpan() returns a CollectorSpan
  - CollectorSpan captures: name, type, entityId, startTime, input, etc.
  - When CollectorSpan.end() is called, it captures: endTime, output, status
  - Span data stored in collector with parent-child hierarchy

In Finalize Step (memoized):
  - Create REAL workflow span first (with memoized IDs)
  - For each collected span: create REAL child span from workflow span
  - Proper hierarchy: workflow → step → conditional → step
  - All spans created with correct parent references
```

## How It Works Now

```
INNGEST ENGINE (AFTER FIX):
  1. Create SpanCollector with trace ID
  2. Create CollectorSpan for workflow (collects metadata)
  3. proxyWorkflowSpan delegates to CollectorSpan
  4. During execution: CollectorSpan collects child span metadata in tree structure
  5. In finalize (memoized - runs only ONCE):
     a. Create REAL workflow span first
     b. Recursively create real child spans from collected data
     c. Each child created via parent.createChildSpan() → proper hierarchy
     d. Look up step results by entityId (step ID)
     e. Use step result timing (startedAt/endedAt) for accurate durations
  6. Result: Proper hierarchical traces with correct timing
```

## Why This Works With Multiple Replicas

A key concern with any Inngest solution is multi-replica safety. Here's why SpanCollector works correctly even when multiple server replicas are running:

### Inngest's Replay Model

When an Inngest function runs:
1. The function is invoked from the beginning on every replay
2. `step.run()` calls check if their result is memoized - if yes, they return instantly
3. Only NEW (not-yet-memoized) steps actually execute
4. The function continues until it hits a new step or completes

### Why SpanCollector Is Multi-Replica Safe

```
Replica A (Invocation 1):
  - Executes step1.run() → actually runs, takes 5 seconds
  - Function suspends, result memoized

Replica B (Invocation 2 - replay):
  - step1.run() → returns instantly (memoized)
  - Executes step2.run() → actually runs, takes 3 seconds
  - Function suspends, result memoized

Replica A (Invocation 3 - replay):
  - step1.run() → returns instantly (memoized)
  - step2.run() → returns instantly (memoized)
  - Executes finalize.run() → creates all spans
  - Function completes
```

**The key insight**: SpanCollector captures the span hierarchy on EVERY invocation (it's just collecting metadata, very cheap). When finalize runs (memoized, so only ONCE), it has access to the full collected hierarchy from that specific invocation.

**No shared state needed** - each invocation:
1. Creates its own SpanCollector (local variable)
2. Collects the same hierarchy (deterministic based on workflow structure)
3. Only one invocation will execute finalize (Inngest memoization)
4. That invocation creates all the real spans with proper hierarchy

**Timing**: Step results already have memoized `startedAt`/`endedAt` from when steps actually executed. In finalize, we look up each step's timing by `entityId` (the step ID) and use that for accurate span durations.

## Key Insights

1. **Span identity vs span objects**: Inngest's model requires separating span IDs (memoized) from span objects (created in finalize)

2. **Deferred creation**: By collecting metadata during execution and creating real spans in finalize, we avoid duplication from Inngest replays

3. **Tree structure preservation**: CollectorSpan maintains parent-child relationships during collection, then recreates them with real spans

4. **Multi-replica safe by design**: No shared state across replicas - each invocation collects the hierarchy independently, and only one will execute finalize

5. **Agent spans work differently**: Agent/model spans are created inside step.run() and use normal observability - they don't go through the collector

## Summary by CodeRabbit

* **New Features**
  * Added an Inngest example app with weather-driven activity planning, a planning agent, and local package config.
  * Added Docker Compose for local Inngest development.

* **Bug Fixes**
  * Improved observability: traces now collect span hierarchy and accurate timings, then materialize spans safely (multi-replica safe).

* **Documentation**
  * Expanded observability docs and configuration examples for production exporters.

* **Tests**
  * Added tracing-focused tests to validate observability integration.


workflows/inngest/src/index.test.ts
```diff
@@ -14882,4 +14884,197 @@ describe('MastraInngestWorkflow', () => {
...
+        status: 'success',
+        output: { result: 'processed: test' },
+      });
+
+      // Verify tracing context was provided
+      expect(capturedTracingContext).toBeDefined();
+
+      // THIS IS THE CRITICAL TEST: currentSpan should be defined for workflow step tracing
+      // Currently this FAILS because InngestWorkflow doesn't create a workflow span
+      expect(capturedTracingContext.currentSpan).toBeDefined();
```
epinzur, line 14985: remove the comment, assuming this is passing now.

epinzur (approved):

PR #11897, opened 2026-01-13
fix(core): remove deprecated threadId/resourceId from AgentExecutionOptions

DanielSLew:
Removes the deprecated `threadId` and `resourceId` options from `AgentExecutionOptions`, these have been deprecated for months. These were deprecated in favour of the `memory` option

Before:
```ts
await agent.stream('Hello', {
  threadId: 'thread-123',
  resourceId: 'user-456',
});
```

After:
```ts
await agent.stream('Hello', {
  memory: {
    thread: 'thread-123',
    resource: 'user-456',
  },
});
```

Also updates the migration guide, API reference docs, and tests to use the new format. The legacy methods (`generateLegacy`, `streamLegacy`) still support the old fields.

## Summary by CodeRabbit

* **Breaking Changes**
  * Removed top-level threadId/resourceId options — use memory: { thread, resource } instead.

* **Documentation**
  * Added migration guides and updated examples to show memory-based usage.
  * Removed deprecated Memory.query() docs and added redirect to Memory.recall().

* **Compatibility**
  * Legacy endpoints continue accepting threadId/resourceId for backward compatibility.

* **Client SDKs**
  * Client and React SDKs now send memory objects; React auto-converts legacy fields.

* **Tests**
  * Updated tests and integration suites to use the memory object shape.


docs/src/content/en/reference/memory/query.mdx
```diff
@@ -1,191 +0,0 @@
----
```
DanielSLew, line 1: This file was removed because this was changed to `recall` from `query` but it looks like while we created a new doc for `recall` we never removed the `query` api reference doc.

abhiaiyer91 (approved):

docs/src/content/en/reference/memory/query.mdx
LekoArts, line 1: You need to add a redirect to the vercel.json inside `docs` to redirect from `query` to `recall`

LekoArts (changes requested):

LekoArts (approved):

wardpeet (approved):

packages/server/src/server/handlers/agents.ts
```diff
@@ -840,7 +841,7 @@ export const STREAM_NETWORK_ROUTE = createRoute({
   responseType: 'stream' as const,
   streamFormat: 'sse' as const,
   pathParamSchema: agentIdPathParams,
-  bodySchema: agentExecutionBodySchema.extend({ thread: z.string().optional() }),
+  bodySchema: agentExecutionBodySchema,
```
DanielSLew, line 844: I think this was an error that we added thread there. It should've been nested (and it was already) under params.memory.thread. We were just misusing it down below and we were also ignoring the resourceId if they provided it under the memory option

PR #11941, opened 2026-01-14
fix(core): TokenLimiterProcessor now filters memory messages

### LINKED ISSUE #11902: Does TokenLimiterProcessor work?
This issue was created from Discord post 1460753031544176700:

[Open in Discord](https://discord.com/channels/1309558646228779139/1460753031544176700)

I'm creating an input processor and I was using the impl of [TokenLimiterProcessor](https://github.com/mastra-ai/mastra/blob/e22026e2af44c8b90312a419edd36bd597f731f4/packages/core/src/processors/processors/token-limiter.ts#L11) as an example. looking at `processInput` It assumes `{messages}`has the complete message history, even tho the args interface has system messages separated out. Also, when I run in the debugger I see that only the latest user message is passed in `messages`, the full history is in `messagesList`. How does this work?

DanielSLew:
Fixes #11902

The `TokenLimiterProcessor.processInput()` wasn't filtering memory messages because it only received input messages, not the full conversation history.

Now it correctly uses `messageList.get.all.db()` to access all messages (memory + input) and `args.systemMessages` for system messages, then removes filtered messages directly from the messageList.

```typescript
// Before: processor only saw input messages, couldn't filter history
const messages = args.messages; // just the latest user message

// After: processor sees all messages and can filter them
const messages = args.messageList?.get.all.db() ?? args.messages;
const systemTokens = countTokens(args.systemMessages);
// ... filter and remove old messages to fit token budget
```

Added end-to-end tests via ProcessorRunner to verify memory messages get filtered and system messages are counted in the budget.

## Summary by CodeRabbit

* **Bug Fixes**
  * Token budget enforcement now accounts for all messages (including system messages) and prioritizes recent conversation context to fit within limits.
  * Improved error handling when no messages are available.

* **Tests**
  * Added comprehensive end-to-end tests covering token limitation, system messages, and message filtering behavior.


packages/core/src/processors/processors/token-limiter.ts
```diff
@@ -113,14 +124,43 @@ export class TokenLimiterProcessor implements Processor<'token-limiter'> {
...
+    }
+
+    // Fallback: return array of filtered non-system messages
+    return messagesToKeep;
+  }
+
+  /**
+   * Count tokens for a CoreMessageV4 (system messages from args.systemMessages)
+   */
+  private countCoreMessageTokens(message: CoreMessageV4): number {
```
TylerBarnes: should rename this to `countCoreSystemMessageTokens` and throw an error if it's used on anything that doesn't have role: system

packages/core/src/processors/processors/token-limiter.ts
```diff
@@ -68,41 +67,53 @@ export class TokenLimiterProcessor implements Processor<'token-limiter'> {
...
+    // remove all non-system messages
     if (systemTokens + TokenLimiterProcessor.TOKENS_PER_CONVERSATION >= limit) {
-      return systemMessages;
+      if (messageList) {
+        const idsToRemove = nonSystemMessages.map(m => m.id);
+        messageList.removeByIds(idsToRemove);
+        return messageList;
+      }
+      return [];
     }
```
TylerBarnes, line 111: I don't think this case makes sense. You can't send an llm a request with no messages. We had some logic before where no messages would send a `.` to the llm - but no matter the system prompt the llm always responds with "I see you just sent a "." character. how can I help you?"

So this should either throw or do something else entirely

TylerBarnes (changes requested):

packages/core/src/processors/processors/token-limiter.ts
TylerBarnes: maybe tripwire is the way to go here

packages/core/src/processors/processors/token-limiter.ts
```diff
@@ -68,41 +68,55 @@ export class TokenLimiterProcessor implements Processor<'token-limiter'> {
...
+    // If system messages alone exceed the limit (accounting for conversation overhead),
+    // remove all non-system messages
     if (systemTokens + TokenLimiterProcessor.TOKENS_PER_CONVERSATION >= limit) {
-      return systemMessages;
+      if (messageList) {
+        const idsToRemove = nonSystemMessages.map(m => m.id);
+        messageList.removeByIds(idsToRemove);
+        return messageList;
+      }
+      return [];
```
TylerBarnes: I think we need a tripwire here too, right? cause removing all non-system messages is the same issue

TylerBarnes (changes requested):
looking good! just one more tripwire, and also the docs for token limiter should show this new behaviour

packages/core/src/processors/processors/token-limiter.ts
DanielSLew: yeah good catch!

packages/core/src/processors/processors/token-limiter.ts
```diff
@@ -102,14 +102,12 @@ export class TokenLimiterProcessor implements Processor<'token-limiter'> {
...
+    // throw TripWire - can't send LLM a request with only system messages
     if (systemTokens + TokenLimiterProcessor.TOKENS_PER_CONVERSATION >= limit) {
-      if (messageList) {
-        const idsToRemove = nonSystemMessages.map(m => m.id);
-        messageList.removeByIds(idsToRemove);
-        return messageList;
-      }
-      return [];
+      throw new TripWire(
+        'TokenLimiterProcessor: System messages alone exceed token limit. Cannot send LLM a request with only system messages.',
```
TylerBarnes: probably want to word this a little differently since this could be confusing: "TokenLimiterProcessor: System messages alone exceed token limit. Request cannot be completed because it would require removing system messages." or something like that

TylerBarnes (approved):

PR #12011, opened 2026-01-16
fix(@mastra/react): Use new syntax for generate/stream REST API calls

wardpeet:
## Summary
- Updates `useChat` hook in `@mastra/react` to use the new function signature for `agent.generate()` and `agent.stream()` methods
- Changes from `{ messages: coreUserMessages, ...options }` to `(coreUserMessages, { ...options })` syntax

## Summary by CodeRabbit

* **Breaking Changes**
  * Core messaging call signatures changed: message lists are now passed as the first argument rather than inside the options object — update SDK call sites to the new syntax.

* **Chores**
  * Patch release entry added documenting compatibility adjustments with updated client streaming/generation signatures.

### DIFF
--- a/.changeset/kind-turtles-serve.md
+++ b/.changeset/kind-turtles-serve.md
@@ -2,8 +2,4 @@
 '@mastra/react': patch
 ---
 
----
-'@mastra/react': patch
----
-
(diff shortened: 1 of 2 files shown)


DanielSLew (approved):
Just need to update `.network` as well then this is good!

PR #12020, opened 2026-01-16
fix(deployer): resolve circular dependency errors in Mastra Cloud dep…

roaminro:
## Description

…loyments

Added @mastra/loggers and @mastra/libsql to GLOBAL_EXTERNALS to prevent circular dependency issues during bundling. This change ensures successful deployments to Mastra Cloud, even when these packages are not explicitly installed. Included tests to validate the configuration of GLOBAL_EXTERNALS.

## Related Issue(s)

fixes https://github.com/mastra-ai/mastra/issues/11982

## Type of Change

- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [ ] Test update

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Added a deployer hook to let deployers specify extra packages to exclude from bundling.

* **Bug Fixes**
  * Cloud deployments now externalize specific runtime dependencies to prevent circular re-export bundling errors.

* **Tests**
  * Added tests covering the new externalization behavior for cloud bundling.

* **Chores**
  * Added patch-level release metadata updates.


packages/deployer/src/build/analyze/constants.ts
```diff
@@ -4,6 +4,8 @@ export const GLOBAL_EXTERNALS = [
   'pino',
   'pino-pretty',
   '@libsql/client',
+  '@mastra/loggers',
+  '@mastra/libsql',
```
YujohnNattrass: This seems correct but I would double check by creating a snapshot and test in cloud

YujohnNattrass (approved):

packages/deployer/src/build/analyze/constants.ts
wardpeet: We should not do this, this will make it external for everyone

wardpeet (commented):
This PR is not doing the correct fix, @roaminro what is actually happening?

packages/deployer/src/build/analyze/constants.ts
roaminro: hmm, maybe we only need this fix to be for mastra cloud?

PR #12113, opened 2026-01-20
feat(core): add getProjectRoot and resolveFromProjectRoot utilities

CalebBarnes:
## Summary

Added `getProjectRoot()` and `resolveFromProjectRoot()` utilities for path resolution.

These utilities find the project root by searching upward for the nearest `package.json` and can resolve relative paths from that location. The search skips `.mastra` directories (and subdirectories like `.mastra/output`) since bundled output has its own `package.json`.

```typescript
import { getProjectRoot, resolveFromProjectRoot } from '@mastra/core/utils';

// Find project root
getProjectRoot(); // → /Users/me/project

// Resolve relative paths from project root
resolveFromProjectRoot('./data/db.sqlite'); // → /project/data/db.sqlite

// Absolute paths returned as-is
resolveFromProjectRoot('/var/data/db.sqlite'); // → /var/data/db.sqlite
```

## Test plan

- [x] Unit tests for `getProjectRoot` and `resolveFromProjectRoot`

## Release Notes

* **New Features**
  * Added utilities for resolving paths relative to project root in core package.

* **Tests**
  * Added comprehensive tests for project root utilities.

* **Chores**
  * Updated example agent dependencies to latest versions.
  * Updated example agent scripts and storage configuration.


examples/agent-v6/src/mastra/storage/libsql.ts
```diff
@@ -0,0 +1,6 @@
+import { LibSQLStore } from '@mastra/libsql';
+
+export const storage = new LibSQLStore({
+  id: 'mastra-storage',
+  url: `file:./data/mastra.db`,
```
CalebBarnes: Before this change, depending on execution context it would resolve this path to many different locations:
- `mastra dev` : `examples/agent-v6/src/mastra/public/mastra.db`
- `mastra start` : `examples/agent-v6/.mastra/output/mastra.db`
- `tsx path/to/file.ts` : `<current cwd>/mastra.db`
- `bun path/to/file.ts` : `<current cwd>/mastra.db`

After this change:
  - `mastra dev` : `<project-root>/mastra.db`
  - `mastra start` : `<project-root>/mastra.db`
  - `tsx path/to/file.ts` : `<project-root>/mastra.db`
  - `bun path/to/file.ts` : `<project-root>/mastra.db`

CalebBarnes: @wardpeet I saw in https://github.com/mastra-ai/mastra/pull/11234 you changed `mastra dev` to set cwd to `./src/mastra/public` so that files created during dev get persisted there.

This PR adds `resolveFromProjectRoot()` to ensure relative paths (like `file:./mastra.db` in LibSQLStore) resolve consistently from the project root regardless of execution context (`mastra dev`, `mastra start`, `tsx`, `bun`, even using mastra in an existing hono/express server, etc.).

My thinking is that db files should always resolve to the same location regardless of how you run the code. Was the public dir change intended for other files, or were db files expected to go there too?

Curious to hear your thoughts on how these should interact.

wardpeet: This PR is wrong :)  It should come from project cwd, and it should resolve from public when in master dev. It is consistent with `mastra build` and how any production application works.

If people want to enforce it, it should be done using their own env vars or path magic. I'm not opposed to creating a utility that resolves to the master dir.

CalebBarnes: @wardpeet Makes sense! After looking into LibSQLStore I realized we should just continue to pass through the url directly to createClient to keep the same behaviour.  I'll revert the LibSQLStore changes and just keep the utilities exported for users who want project-root resolution for other use cases.

PR #12211, opened 2026-01-22
fix(workflows): add generic bail() signature with overloads

TheIsrael1:
## Summary

Implements flexible `bail()` signature using method overloads, allowing users to pass any type while maintaining type compatibility for workflow step chaining.

## Changes

**File:** `packages/core/src/workflows/step.ts`

```typescript
// Before
bail: (result: TStepOutput) => InnerOutput;

// After (method overloads)
bail(result: TStepOutput): InnerOutput;
bail<T>(result: T extends Workflow<any, any, any, any, any, any, any> ? InferZodLikeSchema<T['outputSchema']> : T): InnerOutput;
```

## How It Works

**First overload:** `bail(result: TStepOutput)`
- Maintains backward compatibility
- Preserves type checking for workflow step chaining
- No type errors when using `.then()` after steps with bail

**Second overload:** `bail<T>(result: ...)`
- Allows explicit type passing for better type inference
- When passed a Workflow type, infers from outputSchema
- Default `T = any` allows flexibility

## Usage Examples

```typescript
// Default - accepts any object
bail({ reason: "User rejected" });

// With workflow type - infers from workflow.outputSchema
bail<typeof myWorkflow>({ status: "rejected", message: "..." });

// With custom type
bail<{ status: string }>({ status: "rejected" });
```

## Follow-up

Runtime validation using Zod schemas will be added in a separate PR.

## Issue

Resolves the bail() portion of #12166

## Type

- [ ] Documentation
- [x] Bug fix
- [ ] Feature
- [ ] Breaking change

## Summary by CodeRabbit

* **New Features**
  * Enhanced bail() with overloads for improved type safety and generic inference in workflows, enabling more flexible early exits while remaining backward compatible.

* **Chores**
  * Removed an inline type-ignore directive, allowing stricter type checking during development.


packages/core/src/workflows/step.ts
```diff
@@ -41,7 +41,7 @@ export type ExecuteFunctionParams<TState, TStepInput, TStepOutput, TResume, TSus
   suspend: unknown extends TSuspend
     ? (suspendPayload?: TSuspend, suspendOptions?: SuspendOptions) => InnerOutput | Promise<InnerOutput>
     : (suspendPayload: TSuspend, suspendOptions?: SuspendOptions) => InnerOutput | Promise<InnerOutput>;
-  bail: (result: TStepOutput) => InnerOutput;
+  bail(result: TStepOutput | Record<string, any>): InnerOutput;
```
wardpeet: we probably want something like this, you can define it yourself by passing a workflow or T and by default we accept anything
```suggestion
  bail<T = any>(result: T extends Workflow<any, any, any, any, any, any, any> ? InferZodLikeSchema<T['outputSchema']> : T : InnerOutput;
```

packages/core/src/agent/workflows/prepare-stream/map-results-step.ts
```diff
@@ -113,7 +113,6 @@ export function createMapResultsStep<OUTPUT = undefined>({
         messageList: memoryData.messageList,
       });
 
-      // @ts-expect-error - TODO: types are wrong here, maybe wrong in general?
```
wardpeet: it should be able to remove ts-expect erorr now

wardpeet (commented):
as a follow up please add client side validation with zod. Maybe it happends automatically already at the end of a run.

packages/core/src/agent/workflows/prepare-stream/map-results-step.ts
TheIsrael1: yeah this is a removed line

packages/core/src/workflows/step.ts
TheIsrael1: done!

wardpeet (approved):

PR #12227, opened 2026-01-22
fix: Cleanup Memory message in vector store

### LINKED ISSUE #12225: Cleanup Memory Messages on Vector Store
This issue was created from Discord post 1463988887126020297:

[Open in Discord](https://discord.com/channels/1309558646228779139/1463988887126020297)

Hi! When I delete a thread using `memory.deleteThread` I noticed that message vectors on the table `memory_messages` are not deleted as well. After a couple of months we now have a significant number of entries left behind from deleted threads. I was looking for a way to "cascade" the removal of the thread and also remove the message vectors there too. Could you help me "cascade" that removal too, please?

intojhanurag:
## Description

This PR adds cascading cleanup for vector embeddings when memory threads or messages are deleted.

Previously, calling `memory.deleteThread` or deleting messages removed records from the memory store but left behind orphaned vector embeddings in the vector store. Over time, this caused significant buildup of stale vectors.

This change ensures that:
- Deleting a thread also removes all associated vector embeddings.
- Deleting messages also removes their corresponding vector embeddings.
- Cleanup is **fire-and-forget** (non-blocking) — storage deletion completes first, then vector cleanup runs in the background without blocking the caller.
- Works across all supported vector store backends.

This prevents long-term vector store bloat and keeps memory storage consistent.

## Related Issue(s)

Fixes #12225

## Type of Change

- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [ ] Test update

## Checklist


DanielSLew (commented):
Looks good but can we add tests and also we should do a promise.all for both deleteThread and deleteMessages to parallelize the deletion!

packages/memory/src/index.ts
```diff
@@ -1279,13 +1322,51 @@ Notes:
       throw new Error('All message IDs must be non-empty strings');
     }
 
-    // Delete from storage
+    // Delete from storage and vector store in parallel
     const memoryStore = await this.getMemoryStore();
+
     await memoryStore.deleteMessages(messageIds);
+    if (this.vector && messageIds.length > 0) {
+      await this.deleteMessageVectors(messageIds);
```
abhiaiyer91: i dont think we want to await this?

abhiaiyer91 (changes requested):
We dont want to block on the deletion!

packages/memory/src/index.ts
```diff
@@ -1191,26 +1227,26 @@ Notes:
 
         if (messageIdsNeedingDeletion.size > 0) {
           try {
+            const separator = this.vector.indexSeparator ?? '_';
+            const prefix = `memory${separator}messages`;
             const indexes = await this.vector.listIndexes();
-            const memoryIndexes = indexes.filter(name => name.startsWith('memory_messages'));
+            const memoryIndexes = indexes.filter(name => name.startsWith(prefix));
```
intojhanurag: Fixed a hardcoded separator bug in updateMessages that broke Pinecone/Vectorize.

NikAiyer (approved):
LGTM!

abhiaiyer91 (approved):

PR #12229, opened 2026-01-23
fix(playground-ui): generate unique runId for legacy agent methods

YujohnNattrass:
Legacy agent methods (`generateLegacy` and `streamLegacy`) were hardcoding `runId: agentId`, causing observability traces to show the agent ID instead of a unique run identifier.

**Before:**
```typescript
const generateResponse = await agent.generateLegacy({
  runId: agentId, // Always shows agent ID in traces
  ...
});
```

**After:**
```typescript
const generateResponse = await agent.generateLegacy({
  runId: uuid(), // Unique ID per request
  ...
});

```

This matches how the v5 model path generates runIds via the React SDK hooks.

## Summary by CodeRabbit

* **Bug Fixes**
  * Legacy agent runs now generate a unique run ID per request, improving trace readability and observability.

* **Chores**
  * Added a changelog entry recording the patch-level fix for the playground UI package.


packages/playground-ui/src/services/mastra-runtime-provider.tsx
```diff
@@ -352,7 +353,7 @@ export function MastraRuntimeProvider({
               },
               ...attachments,
             ],
-            runId: agentId,
+            runId: uuid(),
```
epinzur: i think if you just don't pass any id at all, then the code will generate an Id.

epinzur (changes requested):
pls try just not passing `runId` at all

greglobinski (approved):
Thank you @YujohnNattrass

packages/playground-ui/src/services/mastra-runtime-provider.tsx
YujohnNattrass: @epinzur Yup you are correct the code generates a run ID so I removed it.

I was generating the UUID on the client since that’s how the non-legacy generate calls worked, but it looks like for non-legacy generate/stream, the UUID needs to be created on the client for tool call approvals.

epinzur (approved):

PR #12251, opened 2026-01-23
fix(server): enforce reserved context keys in memory endpoints

roaminro:
Server API endpoints now respect `MASTRA_RESOURCE_ID_KEY` and `MASTRA_THREAD_ID_KEY` when set by middleware.

Previously, these endpoints ignored the reserved keys and used client-provided values directly. This allowed authenticated users to potentially access other users' threads, messages, and workflow runs by manipulating the `resourceId` query parameter.

Now when middleware sets these reserved keys, they take precedence over client values:

```typescript
// In middleware
const user = c.get('requestContext').get('user');
c.get('requestContext').set(MASTRA_RESOURCE_ID_KEY, user.id);

// All endpoints now use this value instead of client-provided resourceId
// Attempts to access other users' data return 403
```

## Memory Endpoints

- Add `getEffectiveResourceId`/`getEffectiveThreadId` helpers that prefer reserved keys
- Add `validateThreadOwnership` helper to check thread access (returns 403 if mismatch)
- Update all memory endpoints to use effective values
- Validate thread ownership on single-thread operations (GET, UPDATE, DELETE, etc.)
- `DELETE_MESSAGES_ROUTE` now queries messages to validate thread ownership before deletion
- `DELETE_MESSAGES_ROUTE` fails closed when storage unavailable (throws 403 instead of bypassing validation)
- Add `resourceId` query parameter to `GET_THREAD_BY_ID_ROUTE`, `DELETE_THREAD_ROUTE`, `DELETE_MESSAGES_ROUTE` for ownership validation fallback
- Update network route schemas to match for consistency

## Workflow Endpoints

- Add `validateRunOwnership` helper to check workflow run access
- Update 15+ workflow routes to use `effectiveResourceId` and validate run ownership
- `START_ASYNC_WORKFLOW_ROUTE` now passes resourceId to `createRun`
- `TIME_TRAVEL_STREAM_WORKFLOW_ROUTE` validates run ownership before time traveling
- `STREAM_LEGACY_WORKFLOW_ROUTE` uses effectiveResourceId when creating run
- Centralize authorization utilities in `utils.ts`

## Agent Endpoints

- `GENERATE_AGENT_ROUTE` and `STREAM_GENERATE_ROUTE` validate thread ownership when `memory` option is provided
- Override `memory.resource` and `memory.thread` with effective values from context
- Legacy routes (`GENERATE_LEGACY_ROUTE`, `STREAM_GENERATE_LEGACY_ROUTE`) also validate thread ownership

## Documentation

- Update middleware path from `/api/memory/*` to `/api/*` (covers all endpoints)
- Add "Authorization (User Isolation)" section to middleware docs
- Add "Reserved keys" section to request-context docs
- Update storage docs to link to authorization middleware

## Tests

- 16+ new tests for authorization behavior across memory endpoints
- 5 new tests for agent route authorization
- Workflow authorization tests

## Summary by CodeRabbit

* **Bug Fixes**
  * Memory, agent, and workflow APIs now honor middleware-provided reserved user/thread keys, enforce ownership checks, and return 403 for unauthorized cross-resource/thread access.

* **Documentation**
  * Added Authorization (User Isolation) and Reserved Keys guidance with examples showing middleware-enforced ID precedence and related usage notes.

* **Tests**
  * Added comprehensive tests validating reserved-key precedence and resource/thread isolation across routes.

* **Chores**
  * Added a patch changelog entry documenting the fix.

fixes https://github.com/mastra-ai/mastra/issues/12383


docs/src/content/en/docs/server/middleware.mdx
```diff
@@ -112,6 +112,68 @@ export const mastra = new Mastra({
...
+  server: {
+    auth: {
+      authenticateToken: async (token) => {
+        // Your auth logic returns the user
+        return verifyToken(token); // { id: 'user-123', ... }
+      },
+    },
+    middleware: [
+      {
+        path: '/api/memory/*',
```
davidthor: I believe this should be `/api/*`. There are several endpoints in `/api/agents/*` for creating messages that take in a resourceId to upsert a thread during message creation.

packages/server/src/server/handlers/memory.test.ts
```diff
@@ -1329,6 +1358,433 @@ describe('Memory Handlers', () => {
...
+        const spy = vi.spyOn(mockMemory, 'listThreads');
+
+        // Client tries to access user-b's threads, but middleware set user-a
+        const result = await LIST_THREADS_ROUTE.handler({
+          ...createTestContextWithReservedKeys({ mastra, resourceId: 'user-a' }),
+          agentId: 'test-agent',
+          resourceId: 'user-b', // Client tries to access different user
+          page: 0,
+          perPage: 10,
+        });
```
davidthor, line 1388: According to the OpenAPI spec, the resourceId for the list threads route (and other getters) is passed as a query string parameter, not as requestContext. I think that means this test doesn't cover the case properly.

docs/src/content/en/docs/server/middleware.mdx
```diff
@@ -112,6 +112,68 @@ export const mastra = new Mastra({
...
+          const requestContext = c.get('requestContext');
+          const user = requestContext.get('user');
+
+          if (!user) {
+            return c.json({ error: 'Unauthorized' }, 401);
+          }
+
+          // Force all memory operations to use this user's ID
+          // This takes precedence over any client-provided resourceId
+          requestContext.set(MASTRA_RESOURCE_ID_KEY, user.id);
```
davidthor, line 146: Most GET operations, like listing/getting threads and messages, pass in the resourceId as a query string parameter. In those cases this override wouldn't get the job done, would it?

davidthor (commented):
I'm not entirely sure that setting the resourceId in the requestContext exclusively solves all the necessary cases. Can you confirm?

PR #12378, opened 2026-01-27
feat(core): add datasets for agent/workflow evaluation

YujohnNattrass:
Adds datasets as a new storage domain for systematic evaluation of agents, workflows, and scorers.

**What it does:**
- Create datasets with test inputs and expected outputs
- Run datasets against agents, workflows, or scorers
- Auto-score results and compare runs to detect regressions
- Import/export items via CSV
- Bulk operations (select items → export/create new dataset/delete)

**Usage:**

```ts
// Create dataset and add items
const dataset = await mastra.storage.datasets.create({ name: 'qa-tests' });
await mastra.storage.datasets.addItem(dataset.id, {
  input: { question: 'What is 2+2?' },
  expectedOutput: '4'
});

// Run against an agent with scorers
const run = await mastra.runDataset({
  datasetId: dataset.id,
  targetId: 'my-agent',
  scorerIds: ['accuracy', 'latency']
});

// Compare runs to detect regressions
const comparison = await mastra.compareRuns(run1.id, run2.id);
```

Full playground UI included under Observability → Datasets.


.changeset/dark-emus-reply.md
```diff
@@ -0,0 +1,11 @@
+---
+'@mastra/core': minor
+'@mastra/client-js': minor
+'@mastra/playground-ui': minor
+'@mastra/server': minor
```
YujohnNattrass, line 5: We need to review the consequences of these changesets

PR #12400, opened 2026-01-28
feat: support Anthropic programmatic tool calling (PTC)

nehaaprasad:
### Description

- fixes stream freezing with Anthropic's Programmatic Tool Calling (PTC). When tool calls arrive via streaming `tool-input-start/delta/end` from code execution, Mastra now synthesizes a `tool-call` chunk and merges matching `tool-result` chunks for `providerExecuted` tools, allowing the stream to continue processing.

fix: #12390

### Type of Change

### Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed streams that could hang when a tool provided results via input-only streaming by ensuring end-of-input is forwarded and synthesized tool-call events complete the stream.

* **Refactor**
  * Merge provider-produced tool results into corresponding tool invocation entries and track per-tool streaming metadata with persistence.

* **Tests**
  * Added regression tests for provider-executed tool-input streaming (start/delta/end) with merged outputs.

* **Documentation**
  * Updated comments describing streaming-end handling and result-merging behavior.


ahnaflol:
@naaa760 Just wanted to flag that the studio may not pick up the streaming changes in the PR (not 100% sure)

The studio's stream structure and tool calling structure seemed to become jumbled when I added some patch changes to try make the PTC work last week. 

It's worth verifying that the streamed results arrive and render correctly in the studio.

DanielSLew (approved):

packages/core/src/stream/base/output.ts
```diff
@@ -416,7 +420,52 @@ export class MastraModelOutput<OUTPUT = undefined> extends MastraBase {
               break;
             case 'tool-call-input-streaming-start':
               self.#toolCallDeltaIdNameMap[chunk.payload.toolCallId] = chunk.payload.toolName;
+              self.#toolCallStreamingMeta[chunk.payload.toolCallId] = {
+                toolName: chunk.payload.toolName,
+                providerExecuted: chunk.payload.providerExecuted,
+                providerMetadata: chunk.payload.providerMetadata,
```
wardpeet, line 448: This is a breakin change but maybe adding toString makes it less breaking cause to string works
```suggestion
                providerMetadata: chunk.payload.providerMetadata,
                toString() {
                  return chunk.payload.toolName
                }
```

DanielSLew: Thanks for the suggestion! Looking at this more closely, `#toolCallStreamingMeta` is a private field — the only consumer is the synthetic `ToolCallChunk` construction a few lines below (line ~474), which reads `meta.toolName` as a string property. The object is never string-coerced anywhere, so a `toString()` wouldn't be exercised.

The synthetic chunk itself conforms to the `ToolCallChunk` type which already has `toolName: string` on the payload, so downstream consumers (studio, etc.) see the same shape as a real `tool-call` chunk.

If there's a specific scenario you're worried about, happy to revisit!

PR #12416, opened 2026-01-28
fix(core): handle split tool call args across messages

### LINKED ISSUE #12405: [BUG] Tool call args lost when tool invocation split across messages (client tools)
### Describe the Bug

When a tool invocation is split across two messages in storage, the tool call args are lost when sending to the LLM.

I have added a unit test to reproduce the bug (see link below). Then attempted to fix it with this small commit https://github.com/scoutqa-dot-ai/mastra/commit/eb87c61, I can open a PR if this approach aligns with Mastra team's.

### Steps To Reproduce

1. Use an agent with a client tool
2. Agent generates a tool call (saved with `state='call'` and args)
3. Client executes and sends result (saved as separate message with `state='result'` and empty args)
4. The `tool-result` part in LLM request has `input: {}` instead of original args after loading messages from memory

https://github.com/scoutqa-dot-ai/mastra/blob/a1e1b8f/packages/core/src/agent/message-list/tests/split-tool-call-args.test.ts

### Expected Behavior

The `tool-result` should have `input` populated with the original args from the `state='call'` message.

### Environment Information

```block
N/A - this is a core logic bug reproducible via unit test.
```

### Verification

shobhitsahani:
Fixes #12405

In some cases, tool call arguments are split across messages.
A later message can include the same tool call with empty args, which was
being returned too early.

This change skips empty args and continues until a valid payload is found.

## Summary by CodeRabbit

* **Bug Fixes**
  * Tool invocation arguments now skip empty, null, or invalid entries in later messages so earlier valid argument objects are preserved.

* **Tests**
  * Added tests for multi-message argument aggregation, regression cases, multiple tool-call scenarios, and all-empty-args behavior.

* **Chores**
  * Added a changeset documenting this patch fix.

### DIFF
--- a/.changeset/open-moose-hope.md
+++ b/.changeset/open-moose-hope.md
@@ -1,0 +1,5 @@
+---
+'@mastra/core': minor
+---
+
+Added logger support to Workspace filesystem and sandbox providers. Providers extending MastraFilesystem or MastraSandbox now automatically receive the Mastra logger for consistent logging of file operations and command executions.
--- a/packages/core/src/mastra/index.ts
+++ b/packages/core/src/mastra/index.ts
@@ -2599,6 +2599,10 @@ export class Mastra<
       this.#serverAdapter.__setLogger(this.#logger);
     }
 
+    if (this.#workspace) {
+      this.#workspace.__setLogger(this.#logger);
+    }
+
     this.#observability.setLogger({ logger: this.#logger });
   }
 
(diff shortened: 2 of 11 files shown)


shobhitsahani:
@graysonhicks Regression tests added! ✅

I've added 10 comprehensive test cases that validate the fix for split tool call arguments:

**Test Coverage:**
- ✅ Valid args in first message, empty in second
- ✅ Empty args in first message, valid in second  
- ✅ Undefined/null args handling
- ✅ Primitive value rejection
- ✅ Multiple tool calls with different patterns
- ✅ 3+ message chains
- ✅ Real-world scenario from issue #12405

**Results:** All 10 tests passing locally ✓

The tests ensure this bug won't be reintroduced and document the expected behavior for future development.

Ready for review!

PR #12423, opened 2026-01-28
fix(arize): map mastra span types to correct OpenInference span kinds

### LINKED ISSUE #12397: OtelExporter adds gen_ai.agent.* attributes to MODEL_GENERATION spans, breaking OpenInference span kind classification
## Summary

When using Mastra's observability layer, `MODEL_GENERATION` spans are incorrectly classified as `AGENT` type in OpenInference-compatible observability tools (like Phoenix). This prevents cost calculation since Phoenix only calculates costs for `LLM` span kinds.

## Root Cause

The `@mastra/otel-exporter` adds `gen_ai.agent.id` and `gen_ai.agent.name` attributes to `MODEL_GENERATION` spans if `entityId`/`entityName` are present on the span:

https://github.com/mastra-ai/mastra/blob/main/packages/otel-exporter/src/gen-ai-semantics.ts#L67-L72

```typescript
if (span.type === SpanType.MODEL_GENERATION && span.attributes) {
    // ... model/provider handling ...
    
    if (span.entityId) {
      attributes[ATTR_GEN_AI_AGENT_ID] = span.entityId;  // <-- problematic
    }
    if (span.entityName) {
      attributes[ATTR_GEN_AI_AGENT_NAME] = span.entityName;  // <-- problematic
    }
}
```

Combined with automatic inheritance in `@mastra/observability`:

```typescript
this.entityId = options.entityId ?? options.parent?.entityId;
this.entityName = options.entityName ?? options.parent?.entityName;
```

This means any LLM span that's a child of a workflow or agent span inherits `entityId`/`entityName`, which then gets converted to `gen_ai.agent.*` attributes.

The OpenInference converter (`@arizeai/openinference-genai`) then correctly interprets these attributes as indicating an AGENT span:

```typescript
if (AGENT_KIND_PREFIXES.some((prefix) => spanAttributes[prefix])) {
    spanKind = OpenInferenceSpanKind.AGENT;
}
```

## Impact

- Phoenix shows LLM calls as `span_kind: "AGENT"` instead of `"LLM"`
- Cost calculation doesn't work (Phoenix only calculates costs for LLM spans)
- Token usage data is present but costs show as $0.00

## Current Workaround

Pass empty strings to prevent inheritance (empty string is not nullish, so `??` doesn't fall back):

```typescript
const llmSpan = tracingInstance?.startSpan({
  type: SpanType.MODEL_GENERATION,
  name: `llm:${templateId}`,
  parent: parentSpan,
  entityId: '',    // Prevent parent inheritance
  entityName: '',  // Prevent parent inheritance
  // ...
});
```

## Suggested Fix

For `MODEL_GENERATION` spans, don't add `gen_ai.agent.*` attributes. The `entityId`/`entityName` on an LLM span indicates which agent *initiated* the call, not that the span itself is an agent. This metadata could be preserved in a different attribute (e.g., `mastra.initiated_by.agent_id`) without affecting span kind classification.

```typescript
if (span.type === SpanType.MODEL_GENERATION && span.attributes) {
    // model/provider handling stays the same
    
    // Don't add agent attributes to LLM spans - they break OpenInference classification
    // If needed, use a different attribute to track the initiating agent
    if (span.entityId) {
      attributes['mastra.initiated_by.agent_id'] = span.entityId;
    }
    if (span.entityName) {
      attributes['mastra.initiated_by.agent_name'] = span.entityName;
    }
}
```

## Environment

- `@mastra/core`: 1.0.4
- `@mastra/otel-exporter`: 1.0.0
- `@mastra/arize`: 1.0.0
- Phoenix (Arize)

### LINKED ISSUE #12404: @mastra/arize: Workflow spans incorrectly mapped to LLM span kind in Phoenix
## Problem

When using `@mastra/arize` to export traces to Phoenix (or Arize), **all workflow spans are incorrectly mapped to `LLM` span kind** instead of `CHAIN`.

### Expected

| `mastra.span.type` | `openinference.span.kind` |
|--------------------|---------------------------|
| `workflow_run`, `workflow_step`, etc. | `CHAIN` |
| `model_generation`, `model_step`, `model_chunk` | `LLM` |
| `tool_call`, `mcp_tool_call` | `TOOL` |
| `agent_run` | `AGENT` |

### Actual

All spans default to `LLM` unless they have `gen_ai.tool.*` or `gen_ai.agent.*` attributes.

## Root Cause

The `mapSpanKind` function in `@arizeai/openinference-genai` unconditionally defaults to `LLM`:

```javascript
// @arizeai/openinference-genai/src/attributes.ts
export const mapSpanKind = (spanAttributes: Attributes): Attributes => {
  let spanKind = OpenInferenceSpanKind.LLM;  // ← Default to LLM
  
  // Only overrides for GenAI-specific attributes
  if (AGENT_KIND_PREFIXES.some((prefix) => spanAttributes[prefix])) {
    spanKind = OpenInferenceSpanKind.AGENT;
  }
  if (TOOL_EXECUTION_PREFIXES.some((prefix) => spanAttributes[prefix])) {
    spanKind = OpenInferenceSpanKind.TOOL;
  }
  // No way to get CHAIN!
  
  set(attrs, SemanticConventions.OPENINFERENCE_SPAN_KIND, spanKind);
  return attrs;
};
```

Workflow spans have no `gen_ai.*` attributes, so they incorrectly default to `LLM`.

## Suggested Fix

In `@mastra/arize`, map `mastra.span.type` to `openinference.span.kind` **after** the OpenInference conversion:

```javascript
// After: mutableSpan.attributes = { ...processedAttributes, ...mastraOther };

const SPAN_KIND_MAP = {
  workflow_run: "CHAIN", workflow_step: "CHAIN", workflow_conditional: "CHAIN",
  workflow_conditional_eval: "CHAIN", workflow_parallel: "CHAIN", workflow_loop: "CHAIN",
  workflow_sleep: "CHAIN", workflow_wait_event: "CHAIN",
  model_generation: "LLM", model_step: "LLM", model_chunk: "LLM",
  tool_call: "TOOL", mcp_tool_call: "TOOL",
  agent_run: "AGENT",
  processor_run: "CHAIN", generic: "CHAIN",
};

const spanType = mastraOther["mastra.span.type"];
if (typeof spanType === "string" && SPAN_KIND_MAP[spanType]) {
  mutableSpan.attributes["openinference.span.kind"] = SPAN_KIND_MAP[spanType];
}
```

## Workaround

We're currently using a `pnpm patch` on `@mastra/arize` to apply this fix locally.

## Environment

- `@mastra/arize`: 1.0.0
- `@mastra/core`: 1.0.4
- `@arizeai/openinference-genai`: 0.1.0
- Phoenix: latest

roaminro:
Fixes #12404
Fixes #12397

When exporting traces to Phoenix/Arize, Mastra span types need to be mapped to the appropriate OpenInference span kinds for proper categorization.

This adds a mapping from `mastra.span.type` to `openinference.span.kind`. Only non-CHAIN types are explicitly mapped - all other span types (including future ones) default to `CHAIN`:

| Span Type | OpenInference Kind |
|-----------|-------------------|
| `model_generation`, `model_step`, `model_chunk` | `LLM` |
| `tool_call`, `mcp_tool_call` | `TOOL` |
| `agent_run` | `AGENT` |
| Everything else | `CHAIN` (default) |

This ensures model spans are always classified as `LLM` (enabling cost calculation in Phoenix), even when they inherit `gen_ai.agent.*` attributes from parent agent spans.

Added tests covering all span type mappings plus a test for unknown types defaulting to CHAIN.

### DIFF
--- a/observability/arize/src/openInferenceOTLPExporter.ts
+++ b/observability/arize/src/openInferenceOTLPExporter.ts
@@ -52,17 +52,10 @@ const MASTRA_SPAN_TYPE = 'mastra.span.type';
  * The @arizeai/openinference-genai library defaults all spans to LLM kind,
  * which is incorrect for workflow, agent, tool, and other non-LLM spans.
  * This mapping overrides the span kind based on the mastra.span.type attribute.
+ *
+ * Only non-CHAIN types are mapped here - all other span types default to CHAIN.
  */
 const SPAN_TYPE_TO_KIND: Record<string, OpenInferenceSpanKind> = {
-  // Workflow spans -> CHAIN
-  workflow_run: OpenInferenceSpanKind.CHAIN,
-  workflow_step: OpenInferenceSpanKind.CHAIN,
-  workflow_conditional: OpenInferenceSpanKind.CHAIN,
-  workflow_conditional_eval: OpenInferenceSpanKind.CHAIN,
-  workflow_parallel: OpenInferenceSpanKind.CHAIN,
-  workflow_loop: OpenInferenceSpanKind.CHAIN,
-  workflow_sleep: OpenInferenceSpanKind.CHAIN,
-  workflow_wait_event: OpenInferenceSpanKind.CHAIN,
   // Model spans -> LLM
   model_generation: OpenInferenceSpanKind.LLM,
   model_step: OpenInferenceSpanKind.LLM,
@@ -72,9 +65,6 @@ const SPAN_TYPE_TO_KIND: Record<string, OpenInferenceSpanKind> = {
   mcp_tool_call: OpenInferenceSpanKind.TOOL,
   // Agent spans -> AGENT
   agent_run: OpenInferenceSpanKind.AGENT,
-  // Other spans -> CHAIN
-  processor_run: OpenInferenceSpanKind.CHAIN,
-  generic: OpenInferenceSpanKind.CHAIN,
 };
 
 /**
@@ -250,10 +240,11 @@ export class OpenInferenceOTLPTraceExporter extends OTLPTraceExporter {
 
         // Override span kind based on mastra.span.type
         // The @arizeai/openinference-genai library incorrectly defaults all spans to LLM kind.
-        // We need to map based on the actual Mastra span type.
+        // We need to map based on the actual Mastra span type, defaulting to CHAIN.
         const spanType = mastraOther[MASTRA_SPAN_TYPE];
-        if (typeof spanType === 'string' && SPAN_TYPE_TO_KIND[spanType]) {
-          mutableSpan.attributes[SemanticConventions.OPENINFERENCE_SPAN_KIND] = SPAN_TYPE_TO_KIND[spanType];
+        if (typeof spanType === 'string') {
+          mutableSpan.attributes[SemanticConventions.OPENINFERENCE_SPAN_KIND] =
+            SPAN_TYPE_TO_KIND[spanType] ?? OpenInferenceSpanKind.CHAIN;
         }
       }
 
--- a/observability/arize/src/openInferenceOTLPExporter.ts
+++ b/observability/arize/src/openInferenceOTLPExporter.ts
@@ -47,11 +47,7 @@ const MASTRA_MODEL_CHUNK_OUTPUT = 'mastra.model_chunk.output';
 const MASTRA_SPAN_TYPE = 'mastra.span.type';
 
 /**
- * Maps Mastra span types to OpenInference span kinds.
- *
- * The @arizeai/openinference-genai library defaults all spans to LLM kind,
- * which is incorrect for workflow, agent, tool, and other non-LLM spans.
- * This mapping overrides the span kind based on the mastra.span.type attribute.
+ * Maps Mastra span types to OpenInference span kinds for proper trace categorization.
  *
  * Only non-CHAIN types are mapped here - all other span types default to CHAIN.
  */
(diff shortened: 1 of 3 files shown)


observability/arize/src/openInferenceOTLPExporter.ts
```diff
@@ -42,6 +44,38 @@ const MASTRA_METADATA_PREFIX = 'mastra.metadata.';
...
+  model_chunk: OpenInferenceSpanKind.LLM,
+  // Tool spans -> TOOL
+  tool_call: OpenInferenceSpanKind.TOOL,
+  mcp_tool_call: OpenInferenceSpanKind.TOOL,
+  // Agent spans -> AGENT
+  agent_run: OpenInferenceSpanKind.AGENT,
+  // Other spans -> CHAIN
+  processor_run: OpenInferenceSpanKind.CHAIN,
+  generic: OpenInferenceSpanKind.CHAIN,
+};
```
epinzur, line 64: can this map just have the LLM, TOOL, and AGENT types, and then we default to CHAIN for the others?

If we add new span types, I don't want to have to change every exporter.

cephalization: Agree

epinzur (changes requested):
minor changes, but otherwise looks good.

observability/arize/src/openInferenceOTLPExporter.ts
```diff
@@ -213,6 +247,14 @@ export class OpenInferenceOTLPTraceExporter extends OTLPTraceExporter {
         Object.assign(processedAttributes, usageMetrics);
 
         mutableSpan.attributes = { ...processedAttributes, ...mastraOther };
+
+        // Override span kind based on mastra.span.type
+        // The @arizeai/openinference-genai library incorrectly defaults all spans to LLM kind.
```
cephalization: It defaults unknown genai span types to LLM, but it could probably be chain instead.

https://github.com/Arize-ai/openinference/blob/9f8c42657be41709df4d48d5ca461f9dc441a8af/js/packages/openinference-genai/src/attributes.ts#L261-L277

Do you have opinions on other heuristics we could use from the gen ai span shape to handle this on the openinference side?

epinzur: i think your conversion on the GenAI side looks good for Agent/Tool/LLM stuff... 

the issue is that Mastra also has workflows which don't fit well into any of those buckets.

and I don't think there is any good way in GenAI to indicate a workflow-type process.

roaminro: @epinzur what do you think?

epinzur: I would default to `CHAIN` for our things. 

Since we have specific span types, its easier for us to just switch directly to specific output types instead of trying to guess which is correct based on other attributes in the trace.

epinzur (approved):

PR #12542, opened 2026-01-30
fix: CloudFlare Workers deployment with lazy snapshot directory initialization

### LINKED ISSUE #12536: CloudFlare Workers deployment fails with @mastra/observability@1.1.0 - fileURLToPath called at module init
## Description

Deploying to CloudFlare Workers fails after upgrading to `@mastra/observability@1.1.0`. The worker fails to start due to `fileURLToPath` being called with `undefined` at module initialization time.

## Error Message

```
Uncaught TypeError: The "path" argument must be of type string or an instance of URL. Received undefined
  at null.<anonymous> (node-internal:internal_url:155:15) in fileURLToPath
  at null.<anonymous> (assets/worker-entry-Mq6la_v4.js:335559:20)
 [code: 10021]
```

## Root Cause

The file `observability/mastra/src/exporters/json.ts` (introduced in commit `ab3c190198` via PR #12370) contains Node.js-specific code that executes at **module load time**:

```typescript
// Lines 14-18

const __filename = fileURLToPath(import.meta.url);  // <-- Fails here
const __dirname = dirname(__filename);
const SNAPSHOTS_DIR = join(__dirname, '..', '__snapshots__');
```

In CloudFlare Workers (V8 runtime), `import.meta.url` is `undefined` during worker startup, causing `fileURLToPath(undefined)` to throw.

Since this code runs at module initialization (top-level, not inside a function), it executes as soon as `@mastra/observability` is imported, even if the `JsonExporter` class is never used.

## Environment

- Runtime: CloudFlare Workers (V8)
- Package: `@mastra/observability@1.1.0`
- Previous working version: `@mastra/observability@1.0.0`

## Steps to Reproduce

1. Create a Mastra application targeting CloudFlare Workers
2. Install `@mastra/observability@1.1.0` (or `@mastra/core@1.1.0` which depends on it)
3. Build and deploy to CloudFlare using Wrangler
4. Worker fails to start with the above error

## Suggested Fix

Move the `SNAPSHOTS_DIR` initialization inside the method that uses it (`assertMatchesSnapshot`), making it lazy:

```typescript
// Instead of top-level initialization:
// const __filename = fileURLToPath(import.meta.url);
// const __dirname = dirname(__filename);
// const SNAPSHOTS_DIR = join(__dirname, '..', '__snapshots__');

// Use lazy initialization inside the method:
async assertMatchesSnapshot(snapshotName: string, options?: { updateSnapshot?: boolean }): Promise<void> {
  const getSnapshotsDir = () => {
    if (typeof import.meta.url === 'undefined') {
      throw new Error('assertMatchesSnapshot is not supported in non-Node.js environments');
    }
    const __filename = fileURLToPath(import.meta.url);
    const __dirname = dirname(__filename);
    return join(__dirname, '..', '__snapshots__');
  };
  
  const snapshotPath = join(getSnapshotsDir(), snapshotName);
  // ... rest of method
}
```

This ensures the Node.js-specific code only runs when the testing functionality is actually used, not at module load time.

## Workaround

Users can apply a pnpm patch to `@mastra/observability` to move the initialization into the method until an official fix is released.

## Related

- Introduced in: PR #12370 (commit `ab3c190198`)
- File: `observability/mastra/src/exporters/json.ts`

graysonhicks:
## Description

Fixed CloudFlare Workers deployment failure where `fileURLToPath(import.meta.url)` was called at module load time. The JsonExporter now lazily initializes the snapshots directory path inside `assertMatchesSnapshot()`, allowing the package to be imported in non-Node.js environments like CloudFlare Workers where `import.meta.url` is undefined during startup.

## Related Issue(s)

Fixes #12536

## Type of Change

- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [ ] Test update

## Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Resolved CloudFlare Workers deployment failures by deferring snapshot directory path initialization from module load time to runtime, preventing undefined reference errors in non-Node environments.

* **Tests**
  * Added CloudFlare Workers compatibility tests to validate snapshot functionality and JSON exporter operations in non-Node environments.

### DIFF
--- a/observability/mastra/src/exporters/json.ts
+++ b/observability/mastra/src/exporters/json.ts
@@ -13,10 +13,6 @@ import { readFile, writeFile } from 'node:fs/promises';
 import { dirname, join } from 'node:path';
 import { fileURLToPath } from 'node:url';
 
-const __filename = fileURLToPath(import.meta.url);
-const __dirname = dirname(__filename);
-const SNAPSHOTS_DIR = join(__dirname, '..', '__snapshots__');
-
 import type {
   TracingEvent,
   TracingEventType,
@@ -850,8 +846,22 @@ export class JsonExporter extends BaseExporter {
    * @throws Error if the snapshot doesn't match (and updateSnapshot is false)
    */
   async assertMatchesSnapshot(snapshotName: string, options?: { updateSnapshot?: boolean }): Promise<void> {
+    // Lazily compute the snapshots directory to avoid calling fileURLToPath at module load time.
+    // This is necessary for CloudFlare Workers compatibility where import.meta.url is undefined
+    // during worker startup. See: https://github.com/mastra-ai/mastra/issues/12536
+    const getSnapshotsDir = (): string => {
+      if (typeof import.meta.url === 'undefined') {
+        throw new Error(
+          'assertMatchesSnapshot is not supported in non-Node.js environments (import.meta.url is undefined)',
+        );
+      }
+      const __filename = fileURLToPath(import.meta.url);
+      const __dirname = dirname(__filename);
+      return join(__dirname, '..', '__snapshots__');
+    };
+
     // Resolve snapshot path relative to the __snapshots__ directory
-    const snapshotPath = join(SNAPSHOTS_DIR, snapshotName);
+    const snapshotPath = join(getSnapshotsDir(), snapshotName);
     const normalizedTree = this.buildNormalizedTree();
     const structureGraph = this.generateStructureGraph(normalizedTree);
 
--- a/.changeset/proud-heads-check.md
+++ b/.changeset/proud-heads-check.md
@@ -1,0 +1,5 @@
+---
+'@mastra/observability': patch
+---
+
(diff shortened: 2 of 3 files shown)


observability/mastra/src/exporters/json.ts
```diff
@@ -850,8 +846,22 @@ export class JsonExporter extends BaseExporter {
    * @throws Error if the snapshot doesn't match (and updateSnapshot is false)
    */
   async assertMatchesSnapshot(snapshotName: string, options?: { updateSnapshot?: boolean }): Promise<void> {
+    // Lazily compute the snapshots directory to avoid calling fileURLToPath at module load time.
+    // This is necessary for CloudFlare Workers compatibility where import.meta.url is undefined
+    // during worker startup. See: https://github.com/mastra-ai/mastra/issues/12536
+    const getSnapshotsDir = (): string => {
```
theianjones, line 852: this probably doesnt need to be a function if you define the snap shots here 👀

PR #12557, opened 2026-01-31
feat(core,inngest): durable agent architecture with resumable streams

rase-:
This PR introduces durable agents - a way to make agent execution resilient to disconnections, crashes, and long-running operations.

## The Problem

Standard agent streaming has two fragility points:

1. **Connection drops** - If a client disconnects mid-stream (network blip, browser refresh, mobile app backgrounded), all subsequent events are lost. The client has no way to "catch up" on what they missed.

2. **Long-running operations** - Agent loops with tool calls can take minutes. Holding an HTTP connection open that long is unreliable. If the server restarts or the connection times out, the work is lost.

## The Solution

Durable agents address both problems:

**Resumable streams** solve connection drops. Every event is cached with a sequential index. If a client disconnects at event 5, they can reconnect and request events starting from index 6. They receive cached events immediately, then continue with live events as they arrive.

```typescript
// Start streaming
const { runId, output } = await durableAgent.stream('Analyze this data...');

// Client disconnects at event 5...

// Reconnect and resume from where we left off
const { output: resumed } = await durableAgent.observe(runId, { offset: 6 });
// Receives events 6, 7, 8... from cache, then continues with live events
```

**Durable execution** solves long-running operations. Instead of executing the agent loop directly in the HTTP request, execution happens in a workflow engine (built-in evented engine or Inngest). The HTTP request just subscribes to events. If the connection drops, execution continues. The client can reconnect anytime to observe progress.

## Usage

Wrap any existing `Agent` with durability using factory functions:

```typescript
import { createDurableAgent } from '@mastra/core/agent/durable';

const agent = new Agent({
  id: 'my-agent',
  model: openai('gpt-4'),
  instructions: 'You are helpful',
});

const durableAgent = createDurableAgent({ agent });
```

**Factory functions for different execution strategies:**

| Factory | Execution | Use Case |
|---------|-----------|----------|
| `createDurableAgent({ agent })` | Local, synchronous | Development, simple deployments |
| `createEventedAgent({ agent })` | Fire-and-forget via workflow engine | Long-running operations |
| `createInngestAgent({ agent, inngest })` | Inngest-powered | Production, distributed systems |

## PubSub and Cache

Durable agents use two infrastructure components:

| Component | Purpose | Default |
|-----------|---------|---------|
| **PubSub** | Real-time event delivery during streaming | `EventEmitterPubSub` |
| **Cache** | Stores events for replay on reconnection | `InMemoryServerCache` |

When `stream()` is called, events flow through pubsub in real-time. The cache stores each event with a sequential index. When `observe()` is called, missed events replay from cache before continuing with live events.

**Configure via Mastra instance (recommended):**
```typescript
const mastra = new Mastra({
  cache: new RedisServerCache({ url: 'redis://...' }),
  pubsub: new RedisPubSub({ url: 'redis://...' }),
  agents: {
    // Inherits cache and pubsub from Mastra
    myAgent: createDurableAgent({ agent }),
  },
});
```

**Configure per-agent (overrides Mastra):**
```typescript
const durableAgent = createDurableAgent({
  agent,
  cache: new RedisServerCache({ url: 'redis://...' }),
  pubsub: new RedisPubSub({ url: 'redis://...' }),
});
```

**Disable caching (streams won't be resumable):**
```typescript
const durableAgent = createDurableAgent({ agent, cache: false });
```

For single-instance deployments, the defaults work fine. For multi-instance deployments (load balancer, horizontal scaling), use Redis-backed implementations so any instance can serve reconnection requests.

## Class Hierarchy

- `DurableAgent` extends `Agent` - base class with resumable streams
- `EventedAgent` extends `DurableAgent` - fire-and-forget execution
- `InngestAgent` extends `DurableAgent` - Inngest-powered execution

## Other Additions

- `observe()` method on all durable agents for stream reconnection
- `CachingPubSub` - decorator that wraps any PubSub with automatic event caching
- Model fallback support with retry logic
- 284 tests covering all durable agent functionality
- Docker-free test setup for InngestAgent


packages/core/src/agent/durable/workflows/steps/tool-call.ts
taofeeq-deru, line 1: we need to handle tool suspension here too

packages/core/src/agent/durable/workflows/steps/tool-call.ts
```diff
@@ -0,0 +1,184 @@
...
+          type: 'approval',
+          resumeSchema: JSON.stringify({
+            type: 'object',
+            properties: {
+              approved: { type: 'boolean' },
+            },
+            required: ['approved'],
+          }),
+        });
+
```
taofeeq-deru, line 221: Also, we persist messages before suspension in the original agent implementation, are we doing that some other way here, not sure i've seen it anywhere

taofeeq-deru (commented):
all looks good, apart from the suspension related comments in tool-call step

taofeeq-deru (approved):

packages/core/src/agent/durable/workflows/steps/tool-call.ts
```diff
@@ -0,0 +1,483 @@
...
+      const bgToolsConfig = bgConfig?.tools;
+      const isToolEnabled =
+        bgToolsConfig === 'all' ||
+        (typeof bgToolsConfig === 'object' && bgToolsConfig !== null && (bgToolsConfig as any)[toolName]?.enabled);
+      const shouldRunBackground =
+        bgManager && bgConfig && !bgConfig.disabled && (argsBackground || toolBgConfig?.enabled || isToolEnabled);
+
+      if (shouldRunBackground) {
+        try {
+          const { createBackgroundTask } = await import('../../../../background-tasks/create');
```
taofeeq-deru: not sure this is right, it should import it at the top

packages/core/src/agent/durable/workflows/steps/tool-call.ts
```diff
@@ -0,0 +1,483 @@
...
+                    workspace,
+                    requestContext,
+                    abortSignal: taskContext?.abortSignal,
+                  });
+                },
+              },
+            },
+          });
+          await bgTask.dispatch();
+          const taskResult = await bgManager.waitForNextTask([(bgTask as any).taskId ?? (bgTask as any).task?.id]);
```
taofeeq-deru: looks like we want to always wait for the background task to be done before the loop continues?

packages/core/src/agent/durable/workflows/steps/llm-execution.ts
```diff
@@ -0,0 +1,550 @@
...
+                const { ProcessorRunner } = await import('../../../../processors/runner');
+                const runner = new ProcessorRunner({
+                  inputProcessors: registryEntry.inputProcessors ?? [],
+                  outputProcessors: registryEntry.outputProcessors ?? [],
+                  errorProcessors: registryEntry.errorProcessors,
+                  logger: logger as any,
+                  agentName: typedInput.agentName ?? typedInput.agentId,
+                  processorStates: registryEntry.processorStates,
+                });
+                const currentMessageList = new (await import('../../../message-list')).MessageList();
```
taofeeq-deru: seeing another inline import here, instead of importing at the top

taofeeq-deru (approved):

abhiaiyer91 (approved):

rase-:
Going to merge this work as a base to continue iterating in smaller chunks.

PR #12575, opened 2026-02-02
Atomic db level merge operations in workflow snapshot

taofeeq-deru:
## Description

Add `updateupdateWorkflowState` and `updateWorkflowResults` to all storage databases.
Confirm that the storage supports atomic merge operations before creating run in evented workflows

## Type of Change

- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [ ] Test update

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Workflows can update run state and record per-step results across all storage backends; stores now report if they support concurrent atomic updates.
  * The workflow engine will surface a clear error when used with stores that don't support concurrent updates.
* **Tests**
  * Added comprehensive tests for sequential and concurrent updates, snapshot integrity, merged contexts, and missing-run behavior.
* **Bug Fixes / Reliability**
  * Improved atomicity, retries/backoff, and standardized error handling for snapshot updates.


stores/cloudflare/src/storage/domains/workflows/index.ts
```diff
@@ -39,35 +45,141 @@ export class WorkflowsStorageCloudflare extends WorkflowsStorage {
...
+  async updateWorkflowState({
+    workflowName,
+    runId,
+    opts,
+  }: {
+    workflowName: string;
+    runId: string;
+    opts: UpdateWorkflowStateOptions;
+  }): Promise<WorkflowRunState | undefined> {
+    try {
```
rase-: I think the problem here is that this isn't really transactional.

stores/cloudflare-d1/src/storage/domains/workflows/index.ts
```diff
@@ -35,35 +42,153 @@ export class WorkflowsStorageD1 extends WorkflowsStorage {
...
+    workflowName: string;
+    runId: string;
+    opts: UpdateWorkflowStateOptions;
+  }): Promise<WorkflowRunState | undefined> {
+    try {
+      const fullTableName = this.#db.getTableName(TABLE_WORKFLOW_SNAPSHOT);
+      const now = new Date().toISOString();
+
+      // Load existing snapshot
+      const existingDoc = await this.#db.load<{ snapshot: unknown }>({
```
rase-: This doesn't seem transactional, because we are loading, then merging, and then updating. But it doesn't happen within one transaction.

stores/libsql/src/storage/domains/workflows/index.ts
```diff
@@ -44,6 +44,10 @@ export class WorkflowsLibSQL extends WorkflowsStorage {
     );
   }
 
+  supportsConcurrentUpdates(): boolean {
+    return true;
+  }
+
   private parseWorkflowRun(row: Record<string, any>): WorkflowRun {
```
rase-, line 51: Is the libsql support transactional?

rase- (changes requested):

stores/cloudflare/src/storage/domains/workflows/index.ts
taofeeq-deru: yeah was unable to make it transactional for cloudflare, that's why `supportsConcurrentUpdates` returns `false` for both cloudflare and cloudflare-d1.

stores/cloudflare-d1/src/storage/domains/workflows/index.ts
taofeeq-deru: yeah was unable to make it transactional for cloudflare, that's why supportsConcurrentUpdates returns false for both cloudflare and cloudflare-d1.

stores/libsql/src/storage/domains/workflows/index.ts
taofeeq-deru: yeah, i believe it is. The test added attempts concurrent update which i believe would have failed if it isn't transactional

NikAiyer (commented):
Do we have tests in our shared test suite that cover these functions? I believe we do already, but we might need to make sure that these stores are running those tests.

Also we might need to split up the changeset into multiple changesets.

NikAiyer (approved):
LGTM! Looks like you can actually group changesets together if they have the exact same message, so you can remove some of these changesets.

stores/pg/package.json
```diff
@@ -54,7 +54,7 @@
     "vitest": "catalog:"
   },
   "peerDependencies": {
-    "@mastra/core": ">=1.4.0-0 <2.0.0-0"
+    "@mastra/core": ">=1.0.0-0 <2.0.0-0"
```
NikAiyer: I guess we already bumped the changesets before. In which case, if this doesn't require a new version of core, we can just leave it at 1.4

PR #12599, opened 2026-02-02
feat: add observational memory

TylerBarnes:
Adds Observational Memory (OM) — a system that keeps agent context windows small while building long-term memory across conversations. Two background agents handle this: an Observer watches conversations and extracts observations, and a Reflector periodically reorganizes/condenses those observations into reflections.

```ts
const memory = new Memory({
  storage: new PostgresStore({ connectionString }),
  options: {
    observationalMemory: true, // that's it
  },
});
```

Three-tier memory: recent messages -> observations (5-40x compression) -> reflections (full rewrite). Supports thread-scoped (per-conversation) and resource-scoped (shared across threads) modes. Token thresholds are configurable.

Extended config example:

```ts
const memory = new Memory({
  storage: new PostgresStore({ connectionString }),
  options: {
    observationalMemory: {
      scope: 'resource', // shared memory across all threads for a user (default: 'thread')
      model: 'google/gemini-2.5-flash', // model for both observer and reflector
      shareTokenBudget: true, // let messages and observations share token space
      observation: {
        // model: 'deepseek/deepseek-chat', // override model for observer only
        messageTokens: 20_000, // trigger observation after this many unobserved tokens (default: 30k)
        maxTokensPerBatch: 10_000, // max tokens per batch in resource scope (default: 10k)
      },
      reflection: {
        // model: 'deepseek/deepseek-reasoner', // override model for reflector only
        observationTokens: 90_000, // trigger reflection after this many observation tokens (default: 40k)
      },
    },
  },
});
```

This is the core/storage/memory portion of OM — playground UI and evals are separate PRs.

**What's here:**
- `@mastra/core`: OM types, storage interfaces, feature flag, processor enhancements (state persistence, abort signals, stream writer)
- `@mastra/memory`: `ObservationalMemory` class, observer/reflector agents, token counting, `observe()` API
- Storage adapters: pg, libsql, mongodb get full OM support; inmemory for testing; other adapters are unaffected
- `@mastra/server`: OM status endpoint
- `@mastra/client-js`: `getMemoryStatus` updated (backward compatible)

All new interfaces are optional/additive. Storage adapters guard on `TABLE_OBSERVATIONAL_MEMORY` existence so they work fine with older `@mastra/core`. `@mastra/memory` checks `coreFeatures.has('observationalMemory')` before loading OM.

## Summary by CodeRabbit

* **New Features**
  * Observational Memory (OM): three-tier memory (recent → observations → reflections), per-thread/resource scope, manual observe control, OM UI panel, and new observational-memory API endpoint.
  * Client: fetch observational memory and richer memory status/config responses.
  * Agent: find processor by ID.

* **Improvements**
  * Processors: abort propagation, custom streaming events, and persistent processor state across turns.
  * Storage: OM support in in-memory, Postgres/LibSQL, and MongoDB adapters.

* **Bug Fixes**
  * Sealed messages preserved during merges; safer system-message replacement.


.changeset/cold-roses-teach.md
```diff
@@ -0,0 +1,44 @@
...
+});
+```
+
+**What's new:**
+
+- `observationalMemory: true` enables the three-tier memory system (recent messages → observations → reflections)
+- Thread-scoped (per-conversation) and resource-scoped (shared across all threads for a user) modes
+- Manual `observe()` API for triggering observation outside the normal agent loop
+- New OM storage methods for pg, libsql, and mongodb adapters (conditionally enabled)
+- `Agent.findProcessor()` method for looking up processors by ID
```
DanielSLew, line 41: Haven't looked at what this is for yet but we already have a `mastra.getProcessorById`. Can we not just use that? And if we can't, we should make the naming conventions the same so `agent.getProcessorById`

packages/core/src/agent/message-list/merge/MessageMerger.ts
```diff
@@ -14,6 +14,15 @@ import type { MastraDBMessage, MastraMessageContentV2 } from '../state/types';
  * throughout the MessageList.addOne method.
  */
 export class MessageMerger {
+  /**
+   * Check if a message is sealed (should not be merged into).
+   * Messages are sealed after observation to preserve observation markers.
+   */
+  static isSealed(message: MastraDBMessage): boolean {
```
DanielSLew, line 21: Took me a while to figure out this meant. Maybe rename it to `isLocked` and `metadata.mastra.locked` instead, I think that's easier to grok terminology

packages/core/src/agent/workflows/prepare-stream/schema.ts
```diff
@@ -56,6 +56,8 @@ export const prepareMemoryStepOutputSchema = z.object({
   threadExists: z.boolean(),
   thread: storageThreadSchema.optional(),
   messageList: z.instanceof(MessageList),
+  /** Shared processor states map that persists across loop iterations for both input and output processors */
+  processorStates: z.instanceof(Map<string, ProcessorState>),
```
DanielSLew, line 60: I thought we already had processor states? Isn't that how the BatchParts processor works, or was processor states only in the processStreamParts method before

packages/core/src/memory/types.ts
```diff
@@ -5,12 +5,41 @@ import type { ZodObject } from 'zod';
...
+};
+
+/**
+ * Model configuration for Observational Memory Observer/Reflector agents.
+ * Supports the same patterns as Agent's model configuration:
+ * - Static model config (string or object)
+ * - Dynamic function that receives requestContext
+ * - Array of models with retry configuration
+ */
+export type ObservationalMemoryModelConfig =
```
DanielSLew: Do we need to recreate this model type? Wouldn't it be the same type as all of our other model property fields?

packages/core/src/memory/types.ts
```diff
@@ -303,6 +390,222 @@ export type SemanticRecall = {
   indexName?: string;
 };
 
+/**
+ * Model settings for Observer/Reflector agents in Observational Memory.
+ */
+export interface ObservationalMemoryModelSettings {
```
DanielSLew: Is there a reason here also why we dont' reuse the normal modelSettings type for this instead of creating a copy?

packages/memory/src/processors/observational-memory/observational-memory.ts
```diff
@@ -0,0 +1,3250 @@
...
+  DataOmObservationEndPart,
+  DataOmObservationFailedPart,
+  DataOmProgressPart,
+  ObservationMarkerConfig,
+} from './types';
+
+/**
+ * Format a relative time string like "5 days ago", "2 weeks ago", "today", etc.
+ */
+function formatRelativeTime(date: Date, currentDate: Date): string {
```
DanielSLew, line 49: We should maybe clean up this file into multiple files like `/observational-memory/[types]|[utils]|[index]` other logical parts etc, I think it's going to be hard for coding agents (and humans 😅) to work on this. I'm just worried that if we don't do it now it's just going to become patches upon patches in this. I'm fine with this being a fast follow up after it's merged but it should happen!

stores/clickhouse/src/storage/db/utils.ts
```diff
@@ -1,5 +1,6 @@
-import type { TABLE_NAMES, TABLE_SCHEMAS, StorageColumn } from '@mastra/core/storage';
+import type { CORE_TABLE_NAMES, TABLE_NAMES, TABLE_SCHEMAS, StorageColumn } from '@mastra/core/storage';
```
CalebBarnes: @TylerBarnes  Was this renamed? I don't think core exports `CORE_TABLE_NAMES`?

packages/memory/src/processors/observational-memory/types.ts
```diff
@@ -0,0 +1,389 @@
...
+}) => Promise<MastraModelConfig> | MastraModelConfig;
+
+type ModelWithRetries = {
+  id?: string;
+  model: MastraModelConfig | DynamicModel;
+  maxRetries?: number; //defaults to 0
+  enabled?: boolean; //defaults to true
+};
+
+export type ObservationalMemoryModelConfig = MastraModelConfig | DynamicModel | ModelWithRetries[];
```
DanielSLew: can't we just import this type from core?

DanielSLew (approved):

packages/memory/src/processors/observational-memory/observational-memory.ts
TylerBarnes: definitely agree! I'm planning on doing it when I add async buffering (probably later this week)

stores/clickhouse/src/storage/db/utils.ts
TylerBarnes: needed to change the types to accept a partial of this because OM table might not exist

.changeset/cold-roses-teach.md
TylerBarnes: good call with naming! this one is a little different cause it doesn't check for registered processors, it resolves them from the config and returns a promise (so calling functions with request context if needed). I just renamed it to resolveProcessorById() instead.

packages/core/src/agent/message-list/merge/MessageMerger.ts
TylerBarnes: it's not locked, the message is sealed (as in it no more parts can be added to it)

TylerBarnes: locking indicates a temporary state where you get a lock, do something, and then unlock

packages/core/src/agent/workflows/prepare-stream/schema.ts
TylerBarnes: processor states reset between each step before and weren't properly shared between methods. So they mostly didn't work

TylerBarnes: I think it might've only worked for output streaming and nothing else

packages/core/src/memory/types.ts
TylerBarnes: needed this for backwards compat, it's a breaking change since the needed types aren't exported from core, adding a TODO V2 comment rn

TylerBarnes: actually nvm I fixed it, thanks!

packages/core/src/memory/types.ts
TylerBarnes: fixed, ty!

PR #12609, opened 2026-02-03
feat(playground-ui, playground): add Observational Memory UI

TylerBarnes:
Adds the playground UI for Observational Memory on top of the core PR (#12599).

The chat thread shows observation/reflection markers inline with messages, with expand/collapse to see what the Observer extracted. The agent info panel gets a new Observational Memory section showing current observations, reflection history, token usage, and config.

All OM UI state is gated behind a context provider that no-ops when OM is not configured, so nothing changes for agents without it.

Includes E2E tests with fixture-based mock models that simulate the full observation -> tool call -> reflection cycle.

Targets obs-mem-core — will retarget to main after that merges.


packages/playground-ui/src/domains/agents/components/agent-information/agent-memory-config.tsx
```diff
@@ -72,6 +72,28 @@ export const AgentMemoryConfig = ({ agentId }: AgentMemoryConfigProps) => {
       });
     }
 
+    // Observational Memory section
+    const omConfig = (config as any).observationalMemory;
```
mfrachet: can we type this appropriately?

packages/playground-ui/src/domains/agents/components/agent-information/agent-memory.tsx
```diff
@@ -27,6 +28,17 @@ export function AgentMemory({ agentId, threadId }: AgentMemoryProps) {
...
+  // Check if observational memory is enabled
+  const { data: omStatus } = useMemoryWithOMStatus({
+    agentId,
+    resourceId: agentId, // In playground, agentId is the resourceId
+    threadId,
+  });
+  const isOMEnabled = omStatus?.observationalMemory?.enabled ?? false;
+
+  // Debug: log OM status
+  console.log('[AgentMemory] OM Status:', { omStatus, isOMEnabled, agentId, threadId });
```
mfrachet: to remove

packages/playground-ui/src/domains/agents/context/agent-observational-memory-context.tsx
```diff
@@ -0,0 +1,94 @@
+'use client';
```
mfrachet: to remove

packages/playground-ui/src/domains/memory/hooks/use-memory.ts
```diff
@@ -124,3 +126,94 @@ export const useCloneThread = () => {
...
+    retry: false,
+    refetchOnWindowFocus: false,
+    refetchInterval: isActive && pollWhenActive ? 2000 : false, // Poll every 2 seconds when active
+    placeholderData: previousData => previousData, // Keep previous data during refetch to prevent skeleton flash
+  });
+
+  // Update isActive state when data changes
+  useEffect(() => {
+    const newIsActive =
+      query.data?.observationalMemory?.isObserving || query.data?.observationalMemory?.isReflecting || false;
```
mfrachet: for readability reasons, it sometiems a bit better to extract that variables outside the effect

const isObserving = query.data?.observationalMemory?.isObserving
const isReflecting = query.data?.observationalMemory?.isObserving

useEffect(() => {}; [isObserving, isReflecing])

packages/playground-ui/src/lib/ai-ui/components/bracket-overlay.tsx
```diff
@@ -0,0 +1,230 @@
...
+ * then for each badge measures the region from the previous same-type badge
+ * (or the top of the container) down to the bottom of the current badge.
+ *
+ * Observations span back to the previous badge of any type.
+ * Reflections span back to the previous reflection badge (or the beginning).
+ *
+ * - Only the most recent highlight (regardless of type) is always visible.
+ * - Other highlights appear on hover over their badge.
+ */
+export function BracketOverlay({ containerRef }: { containerRef: React.RefObject<HTMLDivElement | null> }) {
```
mfrachet, line 19: Everything related to targetting DOM ndoes should be put in an effect. In general, we tend to avoid modifying classes imperatively in react.

Instead, we tend to make the selected message ids available in a higher level context in the react tree, and we make the comparison where needed to apply style.

like <ObservedMessages ids={observedIds}>
<AgentChat/>
</ObservedMessages>

And in the place where we show messages:

<Message variant={observedIds===messageId && "glow"> using existing design system variants.

packages/playground-ui/src/lib/ai-ui/messages/assistant-message.tsx
```diff
@@ -39,8 +41,8 @@ export const AssistantMessage = ({ hasModelList }: AssistantMessageProps) => {
   const showModelUsed = hasModelList && modelMetadata;
 
   return (
-    <MessagePrimitive.Root className="max-w-full" data-message-id={messageId}>
-      <div className="text-neutral6 text-ui-lg leading-ui-lg">
+    <MessagePrimitive.Root className="max-w-full" data-message-id={messageId} data-message-index={data.index}>
```
mfrachet, line 44: The mentionned comparison above should probably be applied here

packages/playground-ui/src/lib/ai-ui/messages/observation-marker.tsx
```diff
@@ -0,0 +1,200 @@
...
+
+/**
+ * Renders an inline observation marker in the chat history.
+ * Shows different states: in-progress, completed, or failed.
+ */
+export const ObservationMarker = ({ part, onObservationComplete, onObservationFailed }: ObservationMarkerProps) => {
+  // Trigger callbacks when appropriate
+  if (part.type === 'data-om-observation-end' && onObservationComplete) {
+    // Use setTimeout to avoid calling during render
+    setTimeout(() => onObservationComplete(part.data), 0);
```
mfrachet: browser APIs should be put in useEffect, this is not valid react

packages/playground-ui/src/lib/tanstack-query.tsx
```diff
@@ -6,7 +7,8 @@ export interface PlaygroundQueryClientProps {
 }
 
 export const PlaygroundQueryClient = ({ children, options }: PlaygroundQueryClientProps) => {
-  const queryClient = new QueryClient(options);
+  // Use useState to ensure QueryClient is only created once per component instance
+  const [queryClient] = useState(() => new QueryClient(options));
```
mfrachet: Don't do that. PlaygroundQueryClient is high enough in the react tree to now rerender

packages/playground-ui/tailwind.config.ts
```diff
@@ -93,5 +92,5 @@ export default {
       },
     },
   },
-  plugins: [animate, assistantUi, containerQueries, typography],
+  plugins: [animate, assistantUi, containerQueries],
```
mfrachet: we should live it here. the tailwind config in playground-ui is a preset that is used in packages where typography is used

packages/playground/src/App.tsx
```diff
@@ -191,7 +192,8 @@ function App() {
     return <PlaygroundConfigGuard />;
   }
 
-  const router = createBrowserRouter(routes, { basename: studioBasePath });
+  // Memoize router to prevent recreation on every render (which causes component remounts)
+  const router = useMemo(() => createBrowserRouter(routes, { basename: studioBasePath }), [studioBasePath]);
```
mfrachet: don't do that, high enough in the react tree too :)

mfrachet (changes requested):
Thanks a lot for adding e2e tests and for this awesome feature 🙏🏻 . I got a bunch of small changes to make it a bit more react-ish

packages/playground/src/App.tsx
TylerBarnes: ahh I think opus added this while debugging the sidebar continually reloading, forgot to remove it! thanks 🙏

packages/playground-ui/src/lib/ai-ui/components/bracket-overlay.tsx
TylerBarnes: ahh so I think the problem here is we're adding absolutely positioned highlights across multiple elements. I think it would probably be much more complex with useState, wouldn't it? It doesn't highlight individual messages but across multiple messages

TylerBarnes: going to merge it like this but happy to follow up and make it better 🙏

PR #12657, opened 2026-02-04
fix(workflows): resolve type mismatch with optional default schemas

### LINKED ISSUE #12634: [Bug]: Workflow-to-step type checking uses z.input instead of z.output for schemas with .optional().default()
## Bug Description

When a workflow's `inputSchema` contains fields with `.optional().default()`, the first step in the workflow chain receives a type error. The workflow's type checking uses `z.input` semantics (where optional fields are `T | undefined`) but steps expect `z.output`/`z.infer` semantics (where defaults are applied and the field is `T`).

## Reproduction

```typescript
import { z } from "zod";
import { createWorkflow, createStep } from "@mastra/core/workflows";

// Schema with optional field that has a default
const workflowInputSchema = z.object({
  requiredField: z.string(),
  optionalWithDefault: z.number().optional().default(10),
});

// Step that expects the OUTPUT type (after defaults applied)
const myStep = createStep({
  id: "my-step",
  inputSchema: workflowInputSchema,
  outputSchema: z.object({ result: z.string() }),
  execute: async ({ inputData }) => {
    // At runtime, optionalWithDefault is always a number (default applied)
    // But TypeScript thinks it could be undefined
    return { result: `Value: ${inputData.optionalWithDefault}` };
  },
});

// This causes a TypeScript error
const myWorkflow = createWorkflow({
  id: "my-workflow",
  inputSchema: workflowInputSchema,
  outputSchema: z.object({ result: z.string() }),
})
  .then(myStep); // ❌ Type error here
```

## Error Message

```
Argument of type 'Step<"my-step", { requiredField: string; optionalWithDefault: number; }, { result: string; }, ...>' 
is not assignable to parameter of type 'Step<string, { requiredField: string; optionalWithDefault?: number | undefined; }, ...>'
```

The workflow passes `{ optionalWithDefault?: number | undefined }` (z.input) but the step expects `{ optionalWithDefault: number }` (z.output/z.infer).

## Root Cause Analysis

Looking at the type definitions in `@mastra/core@1.1.0`:

### In `workflows/types.d.ts` (lines 447-462)
Step's `StepParams` uses `z.infer<TInputSchema>` for the execute function's inputData:
```typescript
execute: (params: {
    inputData: z.infer<TInputSchema>;  // Uses z.infer (output type, defaults applied)
    // ...
}) => Promise<z.infer<TOutputSchema>> | z.infer<TOutputSchema>;
```

### In `workflows/workflow.d.ts` (lines 151-155)
The `.then()` method's type checking compares the workflow's TInput against the step's input, but the workflow's TInput comes from the workflow config which preserves `z.input` semantics (optional fields are `T | undefined`).

## Expected Behavior

The workflow-to-step type compatibility check should use `z.output`/`z.infer` semantics consistently, since:
1. Zod's `.parse()` applies defaults before the data reaches the step
2. The step's `execute` function correctly expects the output type
3. At runtime, everything works correctly (defaults are applied)

## Workarounds

Currently using `@ts-expect-error` comments:
```typescript
const myWorkflow = createWorkflow({...})
  // @ts-expect-error Mastra v1 bug: workflow passes z.input type but step expects z.output
  .then(myStep);
```

## Environment

- `@mastra/core`: 1.1.0
- `zod`: 3.x
- TypeScript: 5.9.2

## Impact

This affects any workflow that uses `.optional().default()` in its input schema, which is a common pattern for providing sensible defaults while allowing user customization. The build succeeds and runtime behavior is correct, but developers must add type suppression comments to every affected `.then()` call.

shobhitsahani:
Fixes #12634

TypeScript was throwing errors when using `.optional().default()` in schemas, even though the code worked fine. The type checker treated optional fields as potentially undefined, but Zod applies defaults at runtime.

Fixed by simplifying type constraints in `.then()` and `.foreach()` methods to match runtime behavior. Build passes, no breaking changes.

## Summary by CodeRabbit

* **Refactor**
  * Refined type definitions in workflow methods to provide clearer compile-time error messages and stricter type constraints.
  * Streamlined generic type signatures for improved type inference accuracy during workflow step composition and iteration.


NikAiyer (changes requested):
hey @shobhitsahani I think this change removes the type-level compatibility check entirely, and removes validation in foreach. We might want to take a different approach for this. I made a draft PR here for this issue: #12778

PR #12667, opened 2026-02-04
Lock vendored ai packages

wardpeet:
- **fix vendored**
- **update typescript**

## Summary by CodeRabbit

* **Chores**
  * Standardized TypeScript and development tool versions across packages through centralized dependency management.
  * Updated API extractor tooling version.
  * Removed unused development dependencies from vendored packages.
  * Improved monorepo dependency consistency and maintainability.

### DIFF
--- a/docs/package.json
+++ b/docs/package.json
@@ -77,7 +77,7 @@
     "tailwindcss-animate": "^1.0.7",
     "tsx": "^4.21.0",
     "tw-animate-css": "^1.4.0",
-    "typescript": "catalog:"
+    "typescript": "^5.9.3"
   },
   "browserslist": {
     "production": [
--- a/docs/package.json
+++ b/docs/package.json
@@ -77,7 +77,7 @@
     "tailwindcss-animate": "^1.0.7",
     "tsx": "^4.21.0",
     "tw-animate-css": "^1.4.0",
-    "typescript": "^5.9.3"
+    "typescript": "catalog:"
   },
   "browserslist": {
     "production": [
--- a/package.json
+++ b/package.json
@@ -96,7 +96,7 @@
     "pnpm": ">=10.18.0"
   },
   "resolutions": {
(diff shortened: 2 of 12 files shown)


docs/package.json
```diff
@@ -77,7 +77,7 @@
     "tailwindcss-animate": "^1.0.7",
     "tsx": "^4.21.0",
     "tw-animate-css": "^1.4.0",
-    "typescript": "^5.9.3"
+    "typescript": "catalog:"
```
LekoArts: This will break the docs build for now because it uses npm. Let's not use catalog because then in the docs build we have to run the monorepo install + build

wardpeet: didn't you update it to be in the workspace?

wardpeet:
```suggestion
    "typescript": "^5.9.3"
```

packages/agent-builder/package.json
```diff
@@ -63,7 +63,7 @@
   },
   "peerDependencies": {
     "@mastra/core": ">=1.0.0-0 <2.0.0-0",
-    "typescript": ">=5.7.0",
+    "typescript": "catalog:",
```
wardpeet:
```suggestion
     "typescript": ">=5.7.0",
```

package.json
```diff
@@ -96,7 +96,7 @@
     "pnpm": ">=10.18.0"
   },
   "resolutions": {
-    "typescript": "^5.9.3",
+    "typescript": "catalog:",
```
wardpeet:
```suggestion
    "typescript": "^5.9.3",
```

docs/package.json
LekoArts: It is, but the vercel build uses "npm install" and "npm run build" on the docs folder. This way the build only takes 50sec

LekoArts (approved):

PR #12675, opened 2026-02-04
docs: add notes about using absolute paths for shared database access

roaminro:
When running `mastra dev` alongside a Next.js app, both processes need to access the same database file. Relative paths like `file:./mastra.db` resolve based on each process's working directory, which can differ.

Added tips to the storage docs and Next.js guide recommending absolute paths:

```typescript
url: "file:/absolute/path/to/your/project/mastra.db"
```

## Summary by CodeRabbit

* **Documentation**
  * Added guidance on sharing a database file between Mastra Studio and your app by using absolute file URLs in storage configuration.
  * Included Next.js-specific tips to ensure Mastra Studio and Next.js access the same database during development.
  * Clarified that relative paths resolve differently per process (dev servers may use different working directories).

### DIFF
--- a/docs/src/content/en/docs/memory/storage.mdx
+++ b/docs/src/content/en/docs/memory/storage.mdx
@@ -32,7 +32,7 @@ export const mastra = new Mastra({
 When running `mastra dev` alongside your application (e.g., Next.js), use an absolute path to ensure both processes access the same database:
 
 ```typescript
-url: "file:/Users/yourname/project/mastra.db"
+url: "file:/absolute/path/to/your/project/mastra.db"
 ```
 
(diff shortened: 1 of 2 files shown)


epinzur (approved):

docs/src/content/en/docs/memory/storage.mdx
```diff
@@ -27,6 +27,17 @@ export const mastra = new Mastra({
   }),
 });
 ```
+
+:::tip[Sharing the database with Mastra Studio]
+When running `mastra dev` alongside your application (e.g., Next.js), use an absolute path to ensure both processes access the same database:
+
+```typescript
+url: "file:/Users/yourname/project/mastra.db"
```
epinzur: super minor... but I'd use the other example path in both places.

PR #12676, opened 2026-02-04
fix(core): preserve workspace skills when custom inputProcessors are used

### LINKED ISSUE #12612: Workspace + Skills not working with custom apiRoutes
This issue was created from Discord post 1468065794222067723:

[Open in Discord](https://discord.com/channels/1309558646228779139/1468065794222067723)

When open the playground works correctly, /api/:agentId/stream too.

But when use a custom router not working,  workspace not work, not found skills or files.

NikAiyer:
## Description

When users passed `inputProcessors` to `generate()`/`stream()` options, the auto-derived processors (memory, skills) were completely replaced instead of being merged. This caused skill tools (`skill-activate`, `skill-search`) to be unavailable.

The fix ensures that custom `inputProcessors` override only user-configured processors (from Agent constructor) while preserving implicit processors for memory and skills.

## Related Issue(s)

Fixes #12612

## Type of Change

- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [ ] Test update

## Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Custom input processors passed to generate()/stream() now override only user-configured processors while preserving implicit processors (memory and skills), restoring expected tool availability.

* **Tests**
  * Added tests confirming skill-related tools remain available when custom input processors are supplied via agent configuration or per-call options (including streaming).

### DIFF
--- a/packages/core/src/agent/agent.ts
+++ b/packages/core/src/agent/agent.ts
@@ -417,8 +417,8 @@ export class Agent<
     outputProcessorOverrides?: OutputProcessorOrWorkflow[];
     processorStates?: Map<string, ProcessorState>;
   }): Promise<ProcessorRunner> {
-    // Use overrides if provided, otherwise resolve from agent config + memory
-    const inputProcessors = inputProcessorOverrides ?? (await this.listResolvedInputProcessors(requestContext));
+    // Resolve input processors - overrides replace user-configured but auto-derived (memory, skills) are kept
+    const inputProcessors = await this.listResolvedInputProcessors(requestContext, inputProcessorOverrides);
 
     const outputProcessors = outputProcessorOverrides ?? (await this.listResolvedOutputProcessors(requestContext));
 
@@ -544,13 +544,19 @@ export class Agent<
    * All processors are combined into a single workflow for consistency.
    * @internal
    */
-  private async listResolvedInputProcessors(requestContext?: RequestContext): Promise<InputProcessorOrWorkflow[]> {
-    // Get configured input processors
-    const configuredProcessors = this.#inputProcessors
-      ? typeof this.#inputProcessors === 'function'
-        ? await this.#inputProcessors({ requestContext: requestContext || new RequestContext() })
-        : this.#inputProcessors
-      : [];
+  private async listResolvedInputProcessors(
+    requestContext?: RequestContext,
+    configuredProcessorOverrides?: InputProcessorOrWorkflow[],
+  ): Promise<InputProcessorOrWorkflow[]> {
+    // Get configured input processors - use overrides if provided (from generate/stream options),
+    // otherwise use agent constructor processors
+    const configuredProcessors = configuredProcessorOverrides
+      ? configuredProcessorOverrides
+      : this.#inputProcessors
+        ? typeof this.#inputProcessors === 'function'
+          ? await this.#inputProcessors({ requestContext: requestContext || new RequestContext() })
+          : this.#inputProcessors
+        : [];
 
     // Get memory input processors (with deduplication)
     // Use getMemory() to ensure storage is injected from Mastra if not explicitly configured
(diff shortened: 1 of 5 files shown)


packages/core/src/agent/workflows/prepare-stream/schema.ts
```diff
@@ -20,7 +21,12 @@ export type AgentCapabilities = {
   runInputProcessors: Agent['__runInputProcessors'];
   executeOnFinish: (args: AgentExecuteOnFinishOptions) => Promise<void>;
   outputProcessors?: DynamicArgument<OutputProcessorOrWorkflow[]>;
-  inputProcessors?: DynamicArgument<InputProcessorOrWorkflow[]>;
+  inputProcessors?:
+    | InputProcessorOrWorkflow[]
+    | ((args: {
+        requestContext: RequestContext;
+        overrides?: InputProcessorOrWorkflow[];
+      }) => Promise<InputProcessorOrWorkflow[]> | InputProcessorOrWorkflow[]);
```
TylerBarnes, line 33: why did the type change here?

packages/core/src/agent/agent.ts
```diff
@@ -417,8 +417,8 @@ export class Agent<
     outputProcessorOverrides?: OutputProcessorOrWorkflow[];
     processorStates?: Map<string, ProcessorState>;
   }): Promise<ProcessorRunner> {
-    // Use overrides if provided, otherwise resolve from agent config + memory
-    const inputProcessors = inputProcessorOverrides ?? (await this.listResolvedInputProcessors(requestContext));
+    // Resolve input processors - overrides replace user-configured but auto-derived (memory, skills) are kept
+    const inputProcessors = await this.listResolvedInputProcessors(requestContext, inputProcessorOverrides);
 
     const outputProcessors = outputProcessorOverrides ?? (await this.listResolvedOutputProcessors(requestContext));
```
DanielSLew: don't we need to do the same here?

DanielSLew: We need to do all this same hanlding for output processors too

DanielSLew (changes requested):

packages/core/src/agent/workflows/prepare-stream/map-results-step.ts
```diff
@@ -142,16 +142,15 @@ export function createMapResultsStep<OUTPUT = undefined>({
...
-          : capabilities.inputProcessors
-        : []);
+    // Resolve input processors - overrides replace user-configured but auto-derived (memory, skills) are kept
+    const effectiveInputProcessors = capabilities.inputProcessors
+      ? typeof capabilities.inputProcessors === 'function'
+        ? await capabilities.inputProcessors({
+            requestContext: result.requestContext!,
+            overrides: options.inputProcessors,
+          })
+        : options.inputProcessors || capabilities.inputProcessors
```
TylerBarnes, line 152: is this backwards? if capabilities -> if not function -> use options instead (or capabilities). Wouldn't it be if capabilities -> if not function -> use capabilities (and then could the || even trip??)

packages/core/src/agent/workflows/prepare-stream/schema.ts
NikAiyer: We needed to pass overrides so the function can merge user-provided processors with auto-derived ones.

packages/core/src/agent/workflows/prepare-stream/map-results-step.ts
NikAiyer: I think that part was the original behavior. I don't think that fallback ever usually gets reached, since capabilities is usually a function.

TylerBarnes: I guess I'm confused then, its capabilities ? function ? not a function -> use options (instead of capabilities). Why do we only prefer capabilities when it's a function?

NikAiyer: We prefer capabilities when its a function so we can pass the options.inputProcessors. If those are passed and are defined, then we override the existing user input processors while leaving the memory and skills processors alone. When it's static we don't have a good way to tell, so for that I just left the original logic.

TylerBarnes: ahh got it, probably worth adding an inline comment about it

NikAiyer: Agent seems to always set capabilities as functions anyway:

```
  const capabilities = {
    // ...other stuff...
    inputProcessors: async ({ requestContext, overrides }) =>
      this.listResolvedInputProcessors(requestContext, overrides),
    outputProcessors: async ({ requestContext, overrides }) =>
      this.listResolvedOutputProcessors(requestContext, overrides),
    // ...
  };
```

TylerBarnes (approved):

DanielSLew (approved):

PR #12687, opened 2026-02-04
fix(storage): prevent long pg constraint names

### LINKED ISSUE #12679: [BUG] Dev server errors when restarted with custom schema names over 13 chars in length
### Describe the Bug

When starting the dev server after the first time, get the following error:

```
MastraError: relation "mastra_winston_ai_mastra_workflow_snapshot_workflow_name_run_id" already exists
....
````

With some digging the issue appears to be here https://github.com/mastra-ai/mastra/blob/9b32ee697cd796f73b1ec1475e2a69616d028f87/stores/pg/src/storage/db/index.ts#L196C3-L216C20

The index name `${constraintPrefix}mastra_workflow_snapshot_workflow_name_run_id_key` is 49 chars long plus `constraintPrefix`

When the tables and constraints are first created, if schemaName is 14 chars long and therefore 'constraintPrefix' is 15 (including the "_" suffix) a constraint of greater than 64 chars is created. Postgres automatically truncates the constraint name to 63.

On subsequent runs the `IF NOT EXISTS` uses the full 64 char plus string which will never match and so assumes the constraint does not exist, then tries to recreate it, causing the error.

Possible fix:
Remove the `constraintPrefix` prefix altogether, constraints are already schema scoped. but this might break existing databases. Otherwise truncate the constraint identifier before checking if it exists.

### Steps To Reproduce

Use postgres storage with connectionString, and a schemaName 14 characters or longer
Start the dev server for the first time, starts correctly.
Stop, and restart the dev server, error adding constraint.

https://github.com/DuncanLHS/bug-reproduction

### Expected Behavior

No error, correct constraint names checked in `IF NOT EXISTS`

### Environment Information

```block
- Bun
- Supabase local for postgres
- Windows
```

### Verification

skylarkoo7:
## Summary
- add a shared constraint-name helper with Postgres length guard
- use truncated constraint names in PG table creation and migrations
- keep schema-scoped constraints consistent across restarts

## Test plan
- not run (not requested)

Fixes #12679

## Summary by CodeRabbit

* **Refactor**
  * Standardized database constraint naming via a new schema-aware, lowercase, length-limited naming utility; runtime checks and constraint creation now use these standardized names while preserving legacy-name detection to avoid duplicates. No functional changes to runtime behavior.

* **Tests**
  * Added comprehensive tests covering identifier truncation, multibyte safety, casing, legacy compatibility, and overall naming behavior.


stores/pg/src/storage/db/index.ts
```diff
@@ -193,7 +194,14 @@ function generateTableSQL({
...
   const parsedSchemaName = schemaName ? parseSqlIdentifier(schemaName, 'schema name') : '';
-  const constraintPrefix = parsedSchemaName ? `${parsedSchemaName}_` : '';
+  const workflowSnapshotConstraint = buildConstraintName({
+    baseName: 'mastra_workflow_snapshot_workflow_name_run_id_key',
+    schemaName: parsedSchemaName || undefined,
+  });
+  const spansPrimaryKeyConstraint = buildConstraintName({
+    baseName: 'mastra_ai_spans_traceid_spanid_pk',
+    schemaName: parsedSchemaName || undefined,
+  });
```
wardpeet, line 208:
```suggestion
  const workflowSnapshotConstraint = buildConstraintName({
    baseName: 'mastra_wf_snapshot_wf_name_run_id_key',
    schemaName: parsedSchemaName || undefined,
  });
  const spansPrimaryKeyConstraint = buildConstraintName({
    baseName: 'mastra_ai_spans_traceid_spanid_pk',
    schemaName: parsedSchemaName || undefined,
  });
```

stores/pg/src/storage/db/index.ts
```diff
@@ -8,6 +8,7 @@ import {
   TABLE_SCHEMAS,
   getSqlType,
   getDefaultValue,
+  buildConstraintName,
```
wardpeet: we have to move these utils into the pg store itself, else it's a breaking change with core

NikAiyer (commented):
Hey @skylarkoo7, this looks good, but can you take a look at some of the coderabbit comments that are outside the diff range? I think some of them might be worth addressing.

NikAiyer (changes requested):
Can you add some unit tests for the new constraint-utils? Also we probably want to test how this will work for existing users with the old constraint name to make sure nothing breaks for them.

skylarkoo7:
Added unit tests for `constraint-utils` in `stores/pg/src/storage/db/constraint-utils.test.ts` — 22 tests covering:

**`truncateIdentifier`**
- ASCII values under/at/over the 63-byte limit
- UTF-8 byte-aware truncation (multibyte chars like `€` and `😀`)
- Custom `maxLength` and empty string edge cases

**`buildConstraintName`**
- Schema prefix assembly, lowercase normalization, truncation
- Production constraint names (workflow snapshot + spans PK) with short and long schemas

**Backward compatibility**
- Verifies that no-schema and short-schema constraint names match the old naming convention (lowercased)
- Long schema names are truncated to 63 bytes, matching PostgreSQL's silent truncation behavior
- Confirms the `lower($1)` SQL pattern works correctly with the generated names

All 22 tests pass. This doesn't require a running database since the utilities are pure functions.

NikAiyer (changes requested):
Since the workflow snapshot constraint got its base name shortened from `mastra_workflow_snapshot_workflow_name_run_id_key` to `mastra_wf_snapshot_wf_name_run_id_key`, we might see some issues for existing users who have the old constraint name. Can we add some tests for that as well? We might need to keep the old name, or add some logic to migrate the constraint.

NikAiyer (approved):
LGTM!

PR #12690, opened 2026-02-04
feat(langsmith): add vendor metadata support for dynamic project routing

roaminro:
Adds `withLangsmithMetadata` helper to dynamically configure LangSmith tracing options per-span. This, for example, enables routing traces to different projects based on runtime conditions (e.g., user tier, environment).

```typescript
import { buildTracingOptions } from '@mastra/observability';
import { withLangsmithMetadata } from '@mastra/langsmith';

const tracingOptions = buildTracingOptions(
  withLangsmithMetadata({
    projectName: 'my-project',
    tags: ['production'],
    sessionId: 'user-123',
  }),
);
```

Works with `requestContext` for dynamic values from middleware. Follows the same pattern as `withLangfusePrompt` in the Langfuse exporter.

Tested with https://smith.langchain.com/ 

## Summary by CodeRabbit

* **New Features**
  * Exposed a public metadata helper to set per-span overrides (projectName, sessionId, sessionName) for dynamic trace routing and richer payload metadata via tracing options.

* **Documentation**
  * Added docs and examples for dynamic per-span metadata, merge semantics, and runtime routing via tracing options.

* **Tests**
  * Added tests covering metadata merging, precedence, tag preservation, and end-to-end vendor metadata routing.

### DIFF
--- a/observability/langsmith/src/helpers.ts
+++ b/observability/langsmith/src/helpers.ts
@@ -1,0 +1,102 @@
+/**
+ * LangSmith Tracing Options Helpers
+ *
+ * These helpers integrate with the `buildTracingOptions` pattern from
+ * `@mastra/observability` to add LangSmith-specific tracing features.
+ *
+ * @example
+ * ```typescript
+ * import { buildTracingOptions } from '@mastra/observability';
+ * import { withLangsmithMetadata } from '@mastra/langsmith';
+ *
+ * const agent = new Agent({
+ *   defaultGenerateOptions: {
+ *     tracingOptions: buildTracingOptions(
+ *       withLangsmithMetadata({ projectName: 'my-project' })
+ *     ),
+ *   },
+ * });
+ * ```
+ */
+
+import type { TracingOptionsUpdater } from '@mastra/observability';
+
+/**
+ * LangSmith vendor metadata that can be passed via span metadata.
+ * These fields are extracted by the LangSmith exporter and used
+ * to override default configuration on a per-span basis.
+ */
+export interface LangSmithMetadataInput {
+  /**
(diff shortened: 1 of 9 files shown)


observability/langsmith/src/helpers.ts
```diff
@@ -0,0 +1,102 @@
...
+export interface LangSmithMetadataInput {
+  /**
+   * Override the project name for this span and its children.
+   * This allows dynamically routing traces to different LangSmith projects.
+   */
+  projectName?: string;
+  /**
+   * Additional tags to add to this span in LangSmith.
+   */
+  tags?: string[];
```
epinzur: was there a request for vendor-specific `tags`? This seems like an un-needed complication to me.

observability/langsmith/src/tracing.ts
```diff
@@ -204,12 +205,49 @@ export class LangSmithExporter extends TrackingExporter<
...
+          spanId: span.id,
+          metadata: providerMetadata,
+        });
+        return providerMetadata;
+      }
+      currentSpanId = traceData.getParentId({ spanId: currentSpanId });
+    }
+
+    return undefined;
+  }
```
epinzur, line 252: this logic is the duplicate from `langfuse` exporter?

if yes, maybe we could extract this to the base class `TrackingExporter` ? like `findVendorMetadata()` shared method.

but I guess that would be a breaking change... I guess leave it. 

Perhaps leave a comment about a future refactor for 2.0:

```
// TODO(2.0): Create a shared `findVendorMetadata()` in the base TrackingExporter class and use here (and LangfuseExporter)
```

epinzur (approved):

PR #12693, opened 2026-02-04
fix(core, editor): fix stale agent data in CMS pages

mfrachet:
When editing an agent in the CMS, the agent list and detail pages showed outdated data even after a hard refresh. This happened because the agent remained in Mastra's internal registry after updates.

**The fix:**

Added `removeAgent(keyOrId)` to Mastra, and updated `clearStoredAgentCache` in the editor to also remove the agent from Mastra's registry:

```ts
// packages/core - new method
mastra.removeAgent('agent-id'); // removes by key or ID

// packages/editor - updated behavior
editor.clearStoredAgentCache('agent-id'); // now also clears Mastra registry
```

This ensures fresh data is loaded when the agent is accessed again after a CMS update.

## Summary by CodeRabbit

* **New Features**
  * Added capability to remove agents from the Mastra registry by key or ID.

* **Improvements**
  * Editor cache clearing now also removes specific agents from the Mastra registry and guards against missing caches to avoid no-ops.

* **Tests**
  * Added comprehensive tests covering agent removal, cache clearing, re-registration, and edge cases.


packages/core/src/mastra/index.ts
```diff
@@ -867,6 +867,41 @@ export class Mastra<
...
+   * @example
+   * ```typescript
+   * // Remove by key
+   * mastra.removeAgent('myAgent');
+   *
+   * // Remove by ID
+   * mastra.removeAgent('agent-123');
+   * ```
+   */
+  public removeAgent(keyOrId: string): boolean {
```
DanielSLew, line 886: Maybe deleteAgent instead of remove

DanielSLew (approved):

packages/core/src/mastra/index.ts
wardpeet: i would use remove as we also use add.

PR #12747, opened 2026-02-06
feat: add datasets and experiments

YujohnNattrass:
Adds a Datasets and Experiments system for evaluation workflows. Datasets store versioned collections of test inputs (with optional JSON Schema validation), and Experiments run AI outputs against those inputs with configurable scorers to track quality over time.

**Packages changed:**
- `@mastra/core` — `DatasetsManager`, `Dataset` classes, `DatasetsStorage`/`ExperimentsStorage` abstract base classes, SCD-2 item versioning, JSON Schema validation
- `@mastra/client-js` — new client methods for all dataset/experiment CRUD + bulk ops + experiment comparison
- `@mastra/server` — REST API routes under `/datasets` for datasets, items, versions, experiments, results, and comparison
- `@mastra/libsql` — full `DatasetsLibSQL` and `ExperimentsLibSQL` storage domain implementations
- `@mastra/pg` — composite primary key support in `generateTableSQL`/`createTable` for SCD-2 tables
- `@mastra/clickhouse` — table engine configs for dataset/experiment tables
- `@mastra/cloudflare` — type mappings for dataset/experiment tables
- `@mastra/playground-ui` — full datasets UI: CRUD, CSV/JSON import/export, version browsing, experiment triggering with scorer selection, results with trace visualization, cross-experiment comparison with score deltas

```ts
import { Mastra, DatasetsManager } from '@mastra/core';

const mastra = new Mastra({ /* ... */ });
const dataset = await mastra.datasets.create({ name: 'my-eval-set' });
await dataset.addItems([
  { input: { query: 'What is 2+2?' }, groundTruth: { answer: '4' } },
]);

const result = await dataset.runExperiment({
  targetType: 'agent',
  targetId: 'my-agent',
  scorerIds: ['accuracy'],
});
```

Bumps `@mastra/core` peerdep to `>=1.4.0-0` on all store packages


.changeset/datasets-core.md
```diff
@@ -0,0 +1,33 @@
...
+- `Dataset` — single-dataset handle for adding items and running experiments
+
+**New storage domains:**
+- `DatasetsStorage` — abstract base class for dataset persistence (datasets, items, versions)
+- `ExperimentsStorage` — abstract base class for experiment lifecycle and result tracking
+
+**Example:**
+
+```ts
+import { Mastra, DatasetsManager } from '@mastra/core';
```
DanielSLew: We do'nt want to have any imports other than `Mastra` from @mastra/core, it should be from a subpath

packages/server/src/server/handlers/datasets.ts
```diff
@@ -0,0 +1,713 @@
+import { SchemaValidationError, SchemaUpdateValidationError } from '@mastra/core/datasets';
```
CalebBarnes: @YujohnNattrass are these errors the only reason we need peerdep bump for server? 
If yes then maybe we should either just duplicate these into server, OR check coreFeatures.has("datasets") and dynamic import, OR put these into a shared internal package that gets bundled in?

packages/server/src/server/handlers/datasets.ts
```diff
@@ -0,0 +1,713 @@
...
+  queryParamSchema: paginationQuerySchema,
+  responseSchema: listDatasetsResponseSchema,
+  summary: 'List all datasets',
+  description: 'Returns a paginated list of all datasets',
+  tags: ['Datasets'],
+  requiresAuth: true,
+  handler: async ({ mastra, ...params }) => {
+    try {
+      const { page, perPage } = params;
+      const result = await mastra.datasets.list({ page: page ?? 0, perPage: perPage ?? 10 });
```
CalebBarnes, line 102: @YujohnNattrass If `mastra.datasets` only exists in core ^1.4.0, we can either import coreFeatures from core and check if it has "datasets", OR check if typeof mastra?.datasets === 'function'. If it doesnt each dataset handler can return 501 not implemented.

packages/server/src/server/handlers/datasets.ts
YujohnNattrass: sounds good, I remove these changes

packages/server/src/server/handlers/datasets.ts
YujohnNattrass: great idea, I'll do that

.changeset/datasets-core.md
YujohnNattrass: sounds good thank you

stores/libsql/src/storage/index.ts
```diff
@@ -141,6 +145,8 @@ export class LibSQLStore extends MastraCompositeStore {
     const memory = new MemoryLibSQL(domainConfig);
     const observability = new ObservabilityLibSQL(domainConfig);
     const agents = new AgentsLibSQL(domainConfig);
+    const datasets = new DatasetsLibSQL(domainConfig);
```
wardpeet, line 150: this will be breaking as well if the import is not available.  There is no real good way around it cause you can only dynamically import these classes. 

This should be async but constructors cannot be made async...

abhiaiyer91 (approved):

PR #12794, opened 2026-02-06
feat: add tool description overrides for stored agents

DanielSLew:
Stored agents can now override tool descriptions. The `tools` field changes from a `string[]` to a `Record<string, { description?: string }>`.

Before:
```ts
tools: ["my-tool"]
```

After:
```ts
tools: {
  "my-tool": { description: "Custom description for this agent" },
  "other-tool": {}  // keeps original description
}
```

When a stored agent specifies a `description` for a tool, it gets applied at resolution time in `MastraEditor.resolveStoredTools()`. If no description override is provided, the original tool description is used as-is.

Updated across `@mastra/core` (storage types), `@mastra/server` (API schemas + handlers), `@mastra/editor` (resolution logic), and `@mastra/client-js` (SDK types). Added tests for both the override and no-override cases.

## Summary by CodeRabbit

* **New Features**
  * Per-agent per-tool description overrides: stored agents can include custom tool descriptions that are applied at resolution.
  * Editor/CMS flows accept, normalize, and persist per-tool config so saved agents retain custom tool metadata.

* **Chores**
  * Client SDK and server schemas updated to support the new per-tool config structure.

* **Tests**
  * Added tests ensuring overrides are applied and originals preserved when absent.


abhiaiyer91 (approved):
This looks good to me but we'll need to make sure we migrate the schema properly

mfrachet (approved):

PR #12799, opened 2026-02-07
fix(core): remove rejected response from message list on processor retry

roaminro:
When a `processOutputStep` calls `abort({ retry: true })`, the rejected assistant response was staying in the `messageList`. On the next LLM call, the model would see its own rejected response in the prompt — which confused it and often caused empty text on retry.

The fix is a single `messageList.removeByIds([messageId])` call when `shouldRetry` is true, placed after the rejected step is captured for history but before the next iteration begins.

```ts
// Before: LLM sees on retry
// [system] instructions
// [system] [Processor Feedback] Your previous response was not accepted...
// [user]   What is the capital of France?
// [assistant] <the rejected response is still here>

// After: LLM sees on retry
// [system] instructions
// [system] [Processor Feedback] Your previous response was not accepted...
// [user]   What is the capital of France?
```

Added unit tests for both generate and stream paths that assert the rejected response is absent from the retry prompt.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed a processor retry bug where rejected assistant responses were being incorrectly resent to the LLM on retry, causing confusing model behavior and empty text outputs.

* **Tests**
  * Added comprehensive test coverage to verify that processor retries no longer include rejected responses and properly include retry feedback messages across both standard and streaming execution paths.


DanielSLew (approved):
Nice catch!

Issue #12550, opened 2026-01-30
[BUG] mastra dev fails to build package aliases in Mono Repo


sabinayakc:
### Describe the Bug

I have a monorepo setup with apps and packages.
I am using relative imports inside tsconfig.json since the `workspace import is not working`.

I have a package called lib which imports `better-auth` and has an auth helper.

Now I want to use this auth helper in mastra.

I run `mastra dev`

I get the following issue.
```

> server@1.0.0 dev ../apps/server
> mastra dev

ERROR [2026-01-30 16:46:53.942 -0600] (Mastra CLI): Cannot find package 'better-auth' imported from ../apps/server/.mastra/output/server-config.mjs
Did you mean to import "better-auth/dist/index.mjs"?
```

I also tinkered with transpileOptions.

### Steps To Reproduce

1. Create a turbo repo with default 
```
apps/web 
apps/mastra
packages/lib
```
2. Add path alias to mastra server tsconfig json.
```
  "compilerOptions": {
    "outDir": "dist",
    "paths": {
      "@lib/*": ["../../packages/lib/*"]
    }
  },
```
3. Import the package in a file
```
import {auth} from "@lib/auth";
```
4. Run `mastra dev`

### Link to Minimal Reproducible Example

https://github.com/envoy1084/mastra-monorepo-repro

### Expected Behavior

Should be able to build successfully

### Environment Information

```block
System:
    OS: macOS 26.0.1
    CPU: (10) arm64 Apple M1 Pro
    Memory: 114.38 MB / 16.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 22.21.0 - 
    npm: 10.9.4 - 
    pnpm: 10.0.0 
    bun: 1.3.5 -
```

### Verification

- [x] I have searched the existing issues to make sure this is not a duplicate
- [x] I have included sufficient information for the team to reproduce and understand the issue

daneatmastra:
Thank you for reporting this issue! We have labeled it for trio-wp and we will look into it as soon as possible.

If you're experiencing an error, please provide a [minimal reproducible example](https://github.com/mastra-ai/mastra/blob/main/CONTRIBUTING.md#minimal-reproduction) whenever possible to help us resolve it quickly.

PR #12803, opened 2026-02-07
fix(deployer): resolve cross-package tsconfig path alias deps in dev

intojhanurag:
## Description
- When tsconfig path aliases point to files in a different monorepo package, their transitive dependencies (e.g., `better-auth`) were unresolvable at runtime from `.mastra/output/`
- Use `createRequire` to resolve bare imports from the importer's location and emit absolute paths

## Before
image

## After
Screenshot 2026-02-26 173636

## Related Issue(s)
#12550 

## Test plan
- [x] Added regression test for cross-package tsconfig alias resolution
- [x] All 304 deployer tests pass

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed path alias resolution in development — resolves "Cannot find package" errors when running mastra dev. Transitive dependencies of aliased workspace packages (e.g., @lib/*) are now properly discovered at runtime and no longer treated as external during bundling.

* **Tests**
  * Added integration tests to verify transitive dependency resolution for aliased packages during builds.

* **Chores**
  * Added a changeset entry and bumped the package version.

### DIFF
--- a/.changeset/fix-monorepo-tsconfig-aliases.md
+++ b/.changeset/fix-monorepo-tsconfig-aliases.md
@@ -1,0 +1,5 @@
+---
+'@mastra/deployer': patch
+---
+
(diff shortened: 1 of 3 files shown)


intojhanurag:
I reproduced this issue here https://github.com/intojhanurag/reprodution-issues/tree/master/repro-12550

abhiaiyer91 (approved):

Issue #10889, opened 2025-12-04
[FEATURE] Add Client-side tool tracing


epinzur:
### Problem Statement

When I run a client-side tool, I don't see it in my trace.

### Proposed Solution

Add client-side tool tracing.

### Component

Other

### Example Use Case

I might have client side tools. It would be cool to have them traced.

### Verification

- [x] I have searched the existing issues to make sure this is not a duplicate
- [x] I have provided sufficient context for the team to understand the request

epinzur:
@intojhanurag, could you take this implementation over from me?  

I created a draft PR that might be ready, or it might need more work. See: https://github.com/mastra-ai/mastra/pull/16425

intojhanurag:
I will work on it this weekend , so I need to just review and add additional cases if missing with testing .

PR #12804, opened 2026-02-07
feat: enable tracing for tool executions through mcp server

Shaik-Sirajuddin:
## Description

The Pr udpates core tool to create a default span if not passed , which enables tracing for all core tool executions
one of output is Observability section in UI , traces are now logged during execution of mcp server tools
## Type of Change

- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [ ] Test update

## Checklist

## Summary by CodeRabbit

* **New Features**
  * MCP server tool executions now generate observable traces that appear in the Observability UI, providing enhanced visibility into tool performance and execution flow.


epinzur (changes requested):
This seems to be a fix for tracing MCP tools that don't have a parent Agent span, and not an implementation the linked issue:  [Client Side tool tracing](https://github.com/mastra-ai/mastra/issues/10889)

For fixing the MCP trace, this is the correct approach, however there are some things to fix before merge:

### Critical: Missing `mastra` parameter

`getOrCreateSpan()` needs a `mastra` instance to create root spans. From `observability/types/tracing.ts`:

```typescript
export interface GetOrCreateSpanOptions<TType extends SpanType> {
  // ...
  mastra?: Mastra;  // Required for root span creation
}
```

The fallback path in `getOrCreateSpan()` (`observability/utils.ts:31`) does:

```typescript
const instance = options.mastra?.observability?.getSelectedInstance({ requestContext });
return instance?.startSpan<T>({ ... });
```

Without `mastra`, when there's no parent span this returns `undefined` — making the change effectively a no-op for the MCP server use case. The `mastra` instance **is** available in `createExecute` (destructured as `_mastra` at `builder.ts:251`), so it should be forwarded:

```typescript
const toolSpan = getOrCreateSpan({
  // ...existing props...
  tracingContext,
  requestContext,
  tracingPolicy: options.tracingPolicy,
  mastra: options.mastra,  // <-- Needs to be added
});
```

---

### Medium: Test assertion weakened

The test "should not inject tracingContext when agentSpan is not available" is renamed to "should inject tracingContext when agentSpan is not available" and the assertion is changed from a specific equality check to a loose truthiness check:

**Before:**
```typescript
expect(receivedTracingContext).toEqual({ currentSpan: undefined });
```

**After:**
```typescript
expect(receivedTracingSpan).toBeTruthy();
```

Suggestions:
- Verify span properties (type should be `SpanType.TOOL_CALL`, name should match, entityType should be `EntityType.TOOL`)
- Verify span lifecycle (`end()` is called after execution completes)
- The test setup (lines 624–636) doesn't provide a mock `mastra` with observability — so if `mastra` is intended to be passed to `getOrCreateSpan`, the test needs a mock that returns a span from `observability.getSelectedInstance().startSpan()`

---

### Minor: Extra fields spread into `createChildSpan`

When a parent span **does** exist, `getOrCreateSpan` spreads `...rest` into `createChildSpan`. This now includes `mastra` and `requestContext` fields that weren't previously passed to `createChildSpan`. These extra fields are likely ignored but could have unintended side effects depending on the span implementation. Worth verifying.

---

### Missing: Changeset

This PR needs a changeset for `@mastra/core` since it's a behavioral change (new feature). Should be a `minor` bump:
```
feat: enable tracing for tool executions through MCP server when no parent span exists
```

packages/core/src/tools/tool-builder/builder.ts
```diff
@@ -286,6 +287,9 @@ export class CoreToolBuilder extends MastraBase {
           toolType: logType || 'tool',
         },
         tracingPolicy: options.tracingPolicy,
+        tracingContext: tracingContext,
+        requestContext: requestContext,
+        mastra: options.mastra instanceof Mastra ? options.mastra : undefined,
```
Shaik-Sirajuddin, line 348: options.mastra can also be a MastrPrimities type , observability is present for mastra instance type, 
passing here with a type check

Shaik-Sirajuddin:
Updated tests to have strict checks

DanielSLew (approved):

epinzur (approved):

PR #12831, opened 2026-02-09
fix(memory): skip memory tools when no thread context is provided

roaminro:
When an agent has `workingMemory.enabled` on its Memory instance, calling `agent.stream()` without passing `memory` options (no thread/resource) still injected the `updateWorkingMemory` tool. If the model decided to call it, it would throw:

> "Thread ID, Memory instance, and resourceId are required for working memory updates"

This adds a `threadId` guard in `listMemoryTools()` so memory tools are simply not injected when there's no thread context. This matches what the working memory input processor already does.

```ts
// Before: this would inject updateWorkingMemory, which errors if the model calls it
const stream = await agent.stream('hello', { maxSteps: 3 });

// After: memory tools are skipped, no runtime error possible
const stream = await agent.stream('hello', { maxSteps: 3 });
```

When you do pass thread context, everything works as before:

```ts
const stream = await agent.stream('hello', {
  memory: { thread: threadId, resource: resourceId },
  maxSteps: 3,
});
// updateWorkingMemory tool is available and works correctly
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Stop injecting working-memory tools when no thread or resource context exists.
  * Enforce scope-aware validation for working memory: thread-scoped requires threadId; resource-scoped requires resourceId. Threads are only fetched/created when threadId is provided.

* **Tests**
  * Added comprehensive tests for memory-tool injection, scope/context propagation, and absence-of-context scenarios.

* **Chores**
  * Added a changeset recording patch-level updates.


packages/core/src/agent/agent.ts
```diff
@@ -1999,6 +1999,13 @@ export class Agent<
 
     // Get memory tools if available
     const memory = await this.getMemory({ requestContext });
+
+    // Skip memory tools if there's no thread context — they require threadId to function
+    if (!threadId) {
+      this.logger.debug(`[Agent:${this.name}] - Skipping memory tools (no thread context)`, { runId });
+      return convertedMemoryTools;
+    }
```
TylerBarnes: they should still be added if it's in resource scope and there's a resource id. Likely will never happen but good to add just in case

TylerBarnes (approved):

PR #12835, opened 2026-02-09
fix(core): prevent trailing assistant message with structured output

### LINKED ISSUE #12800: Your API request included an `assistant` message in the final position
This issue was created from Discord post 1469545457679208498:

[Open in Discord](https://discord.com/channels/1309558646228779139/1469545457679208498)

```
Error in tryGenerateWithJsonFallback. Attempting fallback. Error [AI_APICallError]: Your API request included an `assistant` message in the final position, which would pre-fill the `assistant` response. When using output format, pre-filling the `assistant` response is not supported.
```

I got this above error when i use `agent.network()` with structuredOutput when i'm using AI SDK Anthropic Provider directly.

[Reproduction]

```
const router = new Agent({
  id: "router",
  model: anthropic("claude-haiku-4-5-20251001"),
  instructions: "...",
  agents: { specialistA, specialistB },
  memory: new Memory({ options: { lastMessages: 20 } }), // <-- key ingredient
});

const stream = await router.network(
  [{ role: "user", content: "..." }],
  {
    structuredOutput: {
      schema: myZodSchema,
      model: anthropic("claude-haiku-4-5-20251001"),
      instructions: "Structure the response...",
    },
    memory: { thread: "thread-123", resource: "user-456" },
  },
);

const result = await stream.object; // undefined — structuring silently failed
```

graysonhicks:
## Description

Fixed structured output failing with Anthropic models when memory is enabled. The error "Your API request included an `assistant` message in the final position" occurred because the prompt sent to Anthropic ended with an assistant-role message, which is not supported when using output format (responseFormat).

The fix adds a `TrailingAssistantGuard` input processor that detects when a prompt ends with an assistant message under Claude 4.6 + native structured output, and appends a user message to satisfy the API constraint.

The guard is dynamically appended to the processor chain in `ProcessorRunner.runProcessInputStep` only when the **resolved model for the current step** is Claude 4.6. This means:
- It never runs for non-Claude 4.6 models, even if Claude 4.6 is elsewhere in a fallback array
- It runs at the end of the processor chain so it catches trailing assistant messages regardless of their source (memory, skills, user input)
- It goes through the normal processor loop — no special invocation path

## Related Issue(s)

Fixes #12800

## Type of Change

- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [ ] Test update

## Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevents structured-output failures with Anthropic/Claude 4.6-like models when memory is enabled by ensuring prompts never end with assistant messages.

* **Tests**
  * Added regression tests for both generate and streaming flows to confirm structured output works correctly with memory enabled.

* **Chores**
  * Patch version bump for the core package.


TylerBarnes (commented):
nice, it looks like it'll work, but feels like it doesn't belong in the execute file. Probably worth making a message processor and attaching the processor to the agent under the conditions your checking

DanielSLew (approved):

PR #12942, opened 2026-02-11
fix(client-js): convert structuredOutput.schema in ClientAgent.network() (#12941)

### LINKED ISSUE #12941: [BUG] ClientAgent.network() does not convert structuredOutput.schema before HTTP serialization
### Describe the Bug

`ClientAgent.network()` in `@mastra/client-js` sends `structuredOutput.schema` raw to the server without converting Zod schemas to JSON Schema first. Both `generate()` and `stream()` call `zodToJsonSchema()` on `structuredOutput.schema` before serialization, but `network()` spreads `params` directly into the request body.

When a Zod v4 schema is passed, non-enumerable properties (`_zod`, `safeParse`, `parse`, etc.) are stripped by `JSON.stringify`. The server receives a garbled object (`{ _def: {...}, type: "object" }`) that passes the `asJsonSchema()` check in `@mastra/core` (because `safeParse` is absent), gets treated as a pre-converted JSON Schema, and is forwarded as-is to the model provider. OpenAI then rejects the malformed schema when `strictJsonSchema: true` is set.

The same issue applies to `requestContext`, which is not processed through `parseClientRequestContext()` in `network()` but is in both `generate()` and `stream()`.

**Workaround:**

Pre-convert the schema before passing to `network()`:

```typescript
const response = await agent.network(messages, {
  structuredOutput: {
    schema: z.toJSONSchema(outputSchema),
  },
});
```

### Steps To Reproduce

```typescript
import { MastraClient } from '@mastra/client-js';
import { z } from 'zod';

const client = new MastraClient({ baseUrl: 'http://localhost:4111' });
const agent = client.getAgent('my-agent');

const outputSchema = z.object({
  name: z.string(),
  summary: z.string(),
});

// This fails — schema is not converted before serialization
const response = await agent.network(
  [{ role: 'user', content: 'Summarize the project' }],
  {
    structuredOutput: { schema: outputSchema },
  },
);
```

Error from OpenAI: schema validation failure due to malformed JSON Schema (missing `properties`, `required`, `additionalProperties`; contains Zod internals like `_def` and `typeName` instead).

https://github.com/raulpineda/mastra-repro-network-structured-output-serialization

### Expected Behavior

`network()` should convert `structuredOutput.schema` from Zod to JSON Schema before sending the request, matching the behavior of `generate()` and `stream()`. The server should receive a valid JSON Schema object.

### Environment Information

```block
- LLM Provider: OpenAI (gpt4o)
- Envinfo

  System:
    OS: macOS 26.2
    CPU: (14) arm64 Apple M4 Pro
    Memory: 164.34 MB / 24.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 24.11.1 - /Users/raul/.nvm/versions/node/v24.11.1/bin/node
    npm: 11.6.2 - /Users/raul/.nvm/versions/node/v24.11.1/bin/npm
    pnpm: 10.0.0 - /opt/homebrew/bin/pnpm
  Browsers:
    Chrome: 144.0.7559.133
    Safari: 26.2
  npmPackages:
    tsx: ^4.21.0 => 4.21.0 
    typescript: ^5.9.3 => 5.9.3 
  npmGlobalPackages:
    corepack: 0.34.2
    npm: 11.6.2

- Libs

@mastra/core: 1.2.0
@mastra/client-js: 1.2.0
zod: 4.x
@ai-sdk/openai: 3.0.26
Model: openai/gpt-4o (Responses API)
Node.js: v22+
```

### Verification

raulpineda:
Fixes https://github.com/mastra-ai/mastra/issues/12941

`ClientAgent.network()` sends `structuredOutput.schema` and `requestContext` raw to the server — unlike `generate()` and `stream()` which process them first.

With Zod v4 schemas, `JSON.stringify` strips non-enumerable properties (`safeParse`, `_zod`, etc.), so the server receives a garbled object like `{ _def: {...}, type: "object" }` instead of valid JSON Schema. It passes the `asJsonSchema()` check (because `safeParse` is missing) and gets forwarded as-is to OpenAI, which rejects it.

The fix adds the same `zodToJsonSchema()` and `parseClientRequestContext()` calls that `generate()` and `stream()` already have:

```typescript
// before — raw params
const response = await this.request(`/agents/${this.agentId}/network`, {
  method: 'POST',
  body: { messages, ...params },
  stream: true,
});

// after — processed params
const processedParams = {
  ...params,
  messages,
  requestContext: parseClientRequestContext(params.requestContext),
  structuredOutput: params.structuredOutput
    ? { ...params.structuredOutput, schema: zodToJsonSchema(params.structuredOutput.schema) }
    : undefined,
};
const response = await this.request(`/agents/${this.agentId}/network`, {
  method: 'POST',
  body: processedParams,
  stream: true,
});
```

Also makes `NetworkStreamParams` generic so it carries the `OUTPUT` type through to `structuredOutput`.

## Example

Applying the changes on this branch as a patch fixes the issue in reproduction case https://github.com/raulpineda/mastra-repro-network-structured-output-serialization/tree/fix/verify-pr-12942-patch

## Summary by CodeRabbit

* **Bug Fixes**
  * Agent.network() now correctly converts structured output schemas and normalizes request context before sending, preventing schema misidentification and server-side validation failures.

* **Tests**
  * Added comprehensive tests for Agent.network() covering schema conversion, request context processing, proper request payloads, and streamed event handling.


taofeeq-deru (changes requested):
Thanks for the PR, looks good. But you'll need to change the json schema back to zod schema here `https://github.com/mastra-ai/mastra/blob/52cc4024b75d49bf745a90376db8940d83396825/packages/server/src/server/handlers/agents.ts#L1372-L1373` for it to work fully

taofeeq-deru (approved):

PR #12963, opened 2026-02-11
 fix(deployer): skip analyzeBundle in dev mode to prevent hang in large monorepos

### LINKED ISSUE #12843: [BUG] `mastra dev` hangs indefinitely in large monorepos due to recursive `analyzeBundle` in `getWatcherInputOptions`
### Describe the Bug

`mastra dev` hangs indefinitely (never starts) in monorepos with many workspace packages (~50+). The hang occurs in `getWatcherInputOptions` (`@mastra/deployer`, `chunk-DIQKRYF4.js`) which calls `analyzeBundle` with `isDev: true`.

The root cause is in `analyzeBundle` → `captureDependenciesToOptimize` → `checkTransitiveDependencies`. When `isDev: true`, `shouldCheckTransitiveDependencies` is set to `true`, which triggers **recursive calls to `analyzeEntry`** for every workspace dependency. Each call creates a new Rollup instance. In a monorepo with 50+ workspace packages, this cascades into dozens of concurrent Rollup passes that never complete.

**The analysis is unnecessary for dev mode**: after `analyzeBundle` completes, `getWatcherInputOptions` filters deps to only workspace packages, then passes them to `getInputOptions` where they're all marked as `external: true`. The `node-resolve` plugin is also removed. So all imports are effectively external anyway — Node.js resolves them at runtime.

### Call chain

```
mastra dev
  → DevBundler.watch()
    → getWatcherInputOptions()         # chunk-DIQKRYF4.js
      → analyzeBundle(isDev: true)     # chunk-UA7HQFWN.js:1130
        → analyzeEntry()               # chunk-UA7HQFWN.js:418 — one Rollup pass (fast)
        → captureDependenciesToOptimize()
          → checkTransitiveDependencies()  # line 343 — RECURSIVE
            → analyzeEntry() per workspace dep  # creates new Rollup instance each time
            → checkTransitiveDependencies()     # recurses up to depth 10
              → ... (50+ workspace packages = dozens of Rollup instances = HANG)
```

### Relevant code

`@mastra/deployer/dist/chunk-UA7HQFWN.js` — `captureDependenciesToOptimize`:

```js
// line 387-389
if (shouldCheckTransitiveDependencies) {    // true when isDev=true
  await checkTransitiveDependencies(initialDepsToOptimize);
}
```

```js
// line 343-386 — checkTransitiveDependencies
async function checkTransitiveDependencies(internalMap, maxDepth = 10, currentDepth = 0) {
  // ...
  for (const [dep, meta] of depsSnapshot) {
    if (!meta.isWorkspace || internalMap.has(dep)) continue;
    // Creates a NEW Rollup instance for EACH workspace dependency:
    const analysis = await analyzeEntry({ entry: resolvedPath, isVirtualFile: false }, "", {
      workspaceMap, projectRoot, logger: noopLogger, sourcemapEnabled: false,
      initialDepsToOptimize: depsToOptimize
    });
    // ... adds new deps, recurses
  }
  if (hasAddedDeps) {
    await checkTransitiveDependencies(internalMap, maxDepth, currentDepth + 1);
  }
}
```

`@mastra/deployer/dist/chunk-DIQKRYF4.js` — `getWatcherInputOptions`:

```js
// After analyzeBundle, deps are filtered to workspace-only, then marked external:
const deps = new Map();
for (const [dep, metadata] of analyzeEntryResult.dependencies.entries()) {
  const pkgName = getPackageName(dep);
  if (pkgName && workspaceMap.has(pkgName)) {
    deps.set(dep, metadata);  // Only workspace deps kept
  }
}
// Then in getInputOptions, isDev=true makes them external:
// alias-optimized-deps plugin: return { id: absolutePath, external: true }
```

### Steps To Reproduce

1. Create a pnpm monorepo with 50+ workspace packages
2. Create a Mastra app that imports from several workspace packages (which transitively import the rest)
3. Run `mastra dev`
4. The CLI outputs `Logger updated [component=BUNDLER] [name=Dev]` and hangs forever

Alternatively, you can reproduce the hang directly:

```js
// test-hang.mjs — run from your mastra app directory
import { posix } from 'path';
import path from 'path';

const cwd = process.cwd();
const entryFile = path.join(cwd, 'src', 'mastra', 'index.ts');
const deployerPath = import.meta.resolve('@mastra/deployer');
const distDir = new URL(deployerPath).pathname.split('/dist/')[0] + '/dist';

const chunk = await import(distDir + '/chunk-UA7HQFWN.js');
const { noopLogger } = await import('@mastra/core/logger');

const { workspaceMap, workspaceRoot } = await chunk.getWorkspaceInformation({ mastraEntryFile: entryFile });
console.log('workspaceMap size:', workspaceMap.size); // 50+ = will hang

const t = setTimeout(() => { console.log('TIMEOUT — analyzeBundle hung'); process.exit(1); }, 60000);

const result = await chunk.analyzeBundle(
  [entryFile], entryFile,
  {
    outputDir: posix.join(cwd, '.mastra', '.build'),
    projectRoot: workspaceRoot,
    platform: 'node',
    isDev: true,
  },
  noopLogger,
);
clearTimeout(t);
console.log('Done — deps:', result.dependencies?.size); // Never reaches here
```

No standalone repo — the issue requires a monorepo with 50+ workspace packages. The reproduction script above demonstrates the hang when run inside such a monorepo.

### Expected Behavior

`mastra dev` should start within a few seconds regardless of monorepo size. The dev server should not perform expensive recursive dependency analysis since all imports are treated as external in dev mode anyway.

In `getWatcherInputOptions` (`chunk-DIQKRYF4.js`), skip the `analyzeBundle` call entirely and pass empty deps:

```js
// Before:
const analyzeEntryResult = await analyzeBundle([entryFile], entryFile, {
  outputDir: posix.join(process.cwd(), ".mastra", ".build"),
  projectRoot: workspaceRoot || process.cwd(),
  platform,
  isDev: true
}, noopLogger);
const deps = new Map();
for (const [dep, metadata] of analyzeEntryResult.dependencies.entries()) {
  const pkgName = getPackageName(dep);
  if (pkgName && workspaceMap.has(pkgName)) {
    deps.set(dep, metadata);
  }
}

// After:
const deps = new Map();
```

This works because:
- The `node-resolve` plugin is already removed by code later in the function
- Bare module imports are unresolved by Rollup → treated as external → resolved by Node.js at runtime
- `nodeModulesExtensionResolver` still handles subpath imports
- `mastra build` uses a completely separate code path and is unaffected

**Workaround**: Apply a `pnpm patch` on `@mastra/deployer` to remove the `analyzeBundle` call in `getWatcherInputOptions`.

### Environment Information

```block
System:
  OS: macOS 26.2
  CPU: (12) arm64 Apple M2 Max
  Memory: 32.00 GB
  Shell: 5.9 - /bin/zsh
Binaries:
  Node: 24.12.0
  pnpm: 10.4.1
npmPackages:
  mastra: 1.2.0
  @mastra/core: 1.2.0
  @mastra/deployer: 1.2.0
  @mastra/memory: 1.2.0


Monorepo: 65 workspace packages, pnpm workspaces, Turborepo.

## Additional Context

- `mastra.config.js` with `bundler: { externals: true }` does NOT help — `getWatcherInputOptions` ignores bundler options entirely (they're only used by `mastra build`)
- The issue is NOT Node.js version specific — tested on Node 22 and 24, same behavior
- The existing `@mastra/deployer/server` + `tsx watch` approach (bypassing the CLI) works perfectly, confirming no bundling analysis is needed for dev mode
- A secondary issue: `getWatcherInputOptions` also passes `workspaceRoot` (monorepo root) instead of `projectRoot` (closest package.json) to `analyzeBundle`, making the initial `analyzeEntry` scope unnecessarily large
```

### Verification

intojhanurag:
## Description

In large monorepos, `checkTransitiveDependencies` spawns a new Rollup instance for every workspace dependency it encounters. When multiple packages share
the same transitive dependency, the same entry file gets analyzed repeatedly — each spawning its own Rollup process. This causes `mastra dev` to hang.

This PR adds a shared `analyzeCache` Map created in `analyzeBundle()` and threaded through all recursive `analyzeEntry()` calls. Before spawning a Rollup
instance, the cache is checked — if the entry was already analyzed, the cached result is returned. Virtual file entries are not cached as they have no
stable path key.

## Related Issue(s)

Fixes #12843

## Type of Change
- [ ] Documentation update
- [ ] Code refactoring
- [x] Performance improvement
- [x] Test update

## Checklist

ELI5: This change stops the dev server from getting stuck in very large projects by remembering which files were already analyzed so the bundler doesn't redo expensive work for the same package over and over. It adds a small shared cache that prevents duplicate Rollup analyses during dev startup, making mastra dev fast again.

- What changed
  - analyzeBundle now creates a shared analyzeCache (Map) and passes it into analyzeEntry / captureDependenciesToOptimize to deduplicate recursive analysis across workspace dependencies.
  - analyzeEntry checks analyzeCache before spawning a Rollup instance and returns cached results when available; non-virtual entries are cached by stable path key.
  - Virtual-file entries are explicitly not cached (they lack a stable path), so in-memory/virtual sources behave the same as before.
  - Tests added to verify: (a) a shared analyzeCache prevents multiple Rollup instances for the same non-virtual entry; (b) shared transitive workspace dependencies are analyzed once when cached; (c) virtual entries are not cached.

- Why
  - In large monorepos, checkTransitiveDependencies would recursively call analyzeEntry for each workspace dependency, causing many redundant Rollup instances and sometimes hanging mastra dev. The shared cache deduplicates those analyses and restores fast dev startup without altering build semantics.

- Known concerns and follow-up
  - Reviewer noted a regression where TypeScript workspace files were no longer transpiled in dev (runtime unknown ".ts" extension). The author acknowledged the concern and agreed to investigate a queue + dedupe approach as an alternative if needed; tests were added and an e2e failing case was reported. Reviewers should validate TypeScript workspace dev behavior (runtime transpilation) in real monorepo e2e scenarios before merge.

- Impact and compatibility
  - Internal change within @mastra/deployer analysis/bundling flow; no public API changes.
  - Restores performance for mastra dev in large monorepos while aiming to preserve TypeScript workspace handling by not caching virtual entries; additional validation recommended for TS workspaces.
  - Includes tests and a changeset; labeled bugfix/performance improvement.

- Linked issue
  - Fixes issue #12843.


TheIsrael1 (dismissed):
## Review: APPROVE

The analysis is correct and the change is safe. Here's the reasoning:

### Why this works

1. **`alias-optimized-deps` is the only consumer** of the `dependencies` Map in the bundler. Passing an empty Map makes it a no-op — `dependencies.has(id)` is always false, so the plugin returns `null` for everything.

2. **No `node-resolve` plugin in dev mode** — the watcher strips it (`watcher.ts:71-73`), so bare specifiers were always going to be unresolved by Rollup. Previously `alias-optimized-deps` was the only thing resolving workspace deps to absolute paths. Without it, Rollup treats unresolvable bare specifiers as external in ESM output (its default behavior).

3. **Build path is unaffected** — `mastra build` uses `bundler.ts` directly with full `analyzeBundle`, not the watcher.

4. **The behavioral difference is fine**: output imports change from absolute paths (`import { Mastra } from '/workspace/packages/core/dist/index.mjs'`) to bare specifiers (`import { Mastra } from '@mastra/core'`). Both resolve correctly in dev mode — Node.js handles workspace packages via symlinks in `node_modules`.

5. **Better for Windows** — the previous path used `pathToFileURL()` to handle Windows paths. Bare specifiers sidestep that entirely.

### Minor suggestions

Left two inline comments — a changeset wording nit and a test improvement suggestion. Neither blocks merge.

Nice fix — eliminating the recursive `analyzeBundle` calls is the right approach for dev mode.

wardpeet:
I'm pretty sure we need to keep deps to make sure TypeScript transpiles ts files in workspaces. If `mastra dev` works for typescript workspaces, this is good to go but i doubt it

packages/deployer/lint-staged.config.js
```diff
@@ -1,5 +1,5 @@
 export default {
-  '*.{ts,tsx}': ['eslint --fix --max-warnings=0', 'prettier --write'],
-  '*.{js,jsx}': ['eslint --fix', 'prettier --write'],
+  '*.{ts,tsx}': ['eslint --fix --max-warnings=0 --no-warn-ignored', 'prettier --write'],
```
wardpeet: why do you need this?

wardpeet (approved):

packages/deployer/lint-staged.config.js
intojhanurag: lol , due to pre-commit warning , just because this file ignored by eslint . Not any specific issue , I can remove it :)

intojhanurag:
This @socket-security is our bot ?

PR #12969, opened 2026-02-11
fix: persist modelId in assistant message content.metadata

catdalfonso:
## Description

Currently, `llm-execution-step.ts` captures `modelId` in `runState.state.responseMetadata` during streaming but never propagates it to `MastraDBMessage.content.metadata` before messages are saved. This means downstream consumers (storage adapters, processors) that read `content.metadata.modelId` receive `undefined`.

Inject `responseMetadata.modelId` into `content.metadata` of all assistant response messages after they are added to the `MessageList` and before output processors run.

## Type of Change

- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [ ] Test update

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Assistant messages now include a modelId in their metadata so stored and streamed responses indicate which model produced them.

* **Tests**
  * Added a test ensuring modelId is persisted on streamed assistant messages (v2 flow).

* **Chores**
  * Added a patch release entry documenting the metadata persistence change.


packages/core/src/loop/workflows/agentic-execution/llm-execution-step.ts
```diff
@@ -1013,6 +1013,18 @@ export function createLLMExecutionStep<TOOLS extends ToolSet = ToolSet, OUTPUT =
...
       }
 
+      // Persist the model identifier from response metadata into assistant message content.
+      // Only update messages produced in this execution step (matched by messageId) so that
+      // earlier iterations in a multi-turn agentic loop retain their own modelId values.
+      const responseModelId = runState.state.responseMetadata?.modelId;
+      if (responseModelId) {
+        for (const msg of messageList.get.response.db()) {
+          if (msg.id === messageId && msg.role === 'assistant') {
+            msg.content.metadata = { ...msg.content.metadata, modelId: responseModelId };
```
TylerBarnes: hmm I'm not sure this is the right way to do it, I think we should add it when we add messages to message list during streaming. The problem here is om "seals" messages now, breaking up the same agents response message into smaller chunks (each with unique ids) so that background buffering/retrieval of parts of messages works properly

TylerBarnes (changes requested):

packages/core/src/loop/workflows/agentic-execution/llm-execution-step.ts
catdalfonso: Good call, I moved this around to better support OM.

TylerBarnes (approved):

PR #13000, opened 2026-02-12
fix: Handle undefined assignees in gh-issue-triage workflow

### LINKED ISSUE #12996: [BUG] GH_Issue_Triage github workflow fails
### Describe the Bug

When opening a Github issue the `GH_Issue_Triage` Github Action runs and fails. 

### Steps To Reproduce

1. Open a Github Issue
2. `GH_Issue_Triage` Github Action runs and fails

https://github.com/mastra-ai/mastra/actions/workflows/issue-triage.yml

### Expected Behavior

The `result.assignees` property is not defined in `scripts/gh-issue-triage/index.ts:91` making it so that the .map on line 92 fails.

Image

Assuming this is for internal use. Either fix the returned object from the workflow or if not needed anymore I would remove the entire `if (result.status === 'success')` block

### Environment Information

```block
N/A
```

### Verification

nehaaprasad:
### Description:

- fixes a TypeError in the GitHub issue triage workflow when assignees is undefined. The code attempted to call .map() on result.assignees without checking if it exists.

- The fix adds safety checks to handle cases where assignees may be missing or structured differently, preventing the workflow from crashing and only sending Slack notifications when valid assignees are present.

### Related Issue(s):

fixes #12996

### Type of Change:
[x] Bug fix (non-breaking change that fixes an issue)

### DIFF
--- a/scripts/gh-issue-triage/index.ts
+++ b/scripts/gh-issue-triage/index.ts
@@ -88,21 +88,27 @@ async function main() {
 
   if (result.status === 'success') {
     const workflowOutput = result.result;
-    const assignees = workflowOutput.result.assignees
-      .map((assignee: string) => mappings[assignee])
-      .filter(Boolean)
-      .map(assignee => `<@${assignee}>`);
+    const assigneesList = workflowOutput?.assignees ?? workflowOutput?.result?.assignees;
 
-    await tools['slack_slack_post_message'].execute({
-      context: {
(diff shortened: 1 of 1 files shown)


rburgt (changes requested):
This seems to be resolving symptoms but not the cause. I feel this PR would resolve the issue https://github.com/mastra-ai/mastra-triage/pull/13

PR #13021, opened 2026-02-12
feat(core): typed Workspace generics for filesystem, sandbox, and mounts

CalebBarnes:
## Summary

Workspace accessors now return the concrete provider types you configured instead of the base interface. Full autocomplete, no casts, and type-safe mount access — all inferred from the constructor, no explicit generics needed.

**Before:**
```ts
workspace.filesystem  // WorkspaceFilesystem | undefined
workspace.sandbox     // WorkspaceSandbox | undefined
```

**After:**
```ts
  filesystem: new LocalFilesystem({ basePath: './data' }),
  sandbox: new E2BSandbox({ timeout: 60000 }),
});
workspace.filesystem  // LocalFilesystem
workspace.sandbox     // E2BSandbox

// Mounts get typed per-key access:
const ws = new Workspace({
  mounts: {
    '/local': new LocalFilesystem({ basePath: './data' }),
    '/s3': new S3Filesystem({ bucket: 'my-bucket' }),
  },
});
ws.filesystem                      // CompositeFilesystem<{ '/local': LocalFilesystem, '/s3': S3Filesystem }>
ws.filesystem.mounts.get('/local') // LocalFilesystem
ws.filesystem.mounts.get('/s3')    // S3Filesystem
```

## Changes

- Make `Workspace`, `WorkspaceConfig`, and `CompositeFilesystem` generic — `workspace.filesystem` / `workspace.sandbox` return the exact type passed to the constructor
- Sensible defaults: bare `Workspace` gives `filesystem: WorkspaceFilesystem | undefined` with proper optional chaining, no `any` needed
- When `mounts` is used, `workspace.filesystem` returns `CompositeFilesystem<TMounts>` with typed per-key access via `mounts.get()`
- Add `ReadonlyMountMap<TMounts>` (replaces `ReadonlyMap<string, WorkspaceFilesystem>`) and `MountMapEntry<TMounts>` for correlated tuple narrowing in `entries()` / iterators
- Implement `getInfo()` on `CompositeFilesystem` — aggregates mount info into `metadata.mounts`
- Derive `CompositeFilesystem.readOnly` from mounts — `true` when every mount is read-only
- Remove unnecessary `!` assertions and `as CompositeFilesystem` casts in tests
- Add compile-time type tests in `workspace.test-d.ts`

Closes COR-524

## Test plan

- [x] All 70 workspace runtime tests pass
- [x] All 91 composite-filesystem tests pass
- [x] 8 compile-time type tests in `workspace.test-d.ts`
- [x] Core and server packages typecheck clean


packages/core/src/workspace/workspace.ts
```diff
@@ -323,8 +335,37 @@ export interface WorkspaceInfo {
...
+ *
+ * // With mounts — filesystem becomes CompositeFilesystem
+ * const ws2 = new Workspace({
+ *   mounts: {
+ *     '/local': new LocalFilesystem({ basePath: './data' }),
+ *   },
+ * });
+ * ws2.filesystem                        // CompositeFilesystem<{ '/local': LocalFilesystem }>
+ * ws2.filesystem.mounts.get('/local')    // LocalFilesystem
+ * ```
```
CalebBarnes: Cut down this huge comment. Don't need so much info about generics since you never really need to pass in generics manually. They should always be inferred.

NikAiyer (approved):

packages/core/src/workspace/workspace.ts
CalebBarnes:
```suggestion

```

PR #13027, opened 2026-02-12
fix(core): prevent workflow hangs and memory leaks in evented workflows

roaminro:
## Summary

Fixes workflow hanging and a critical memory leak (1.5 GB heap growth) affecting deeply nested and parallel evented workflows.

### Fix 1: Race condition in parallel completion check
`processWorkflowStepEnd` used the return value of `updateWorkflowResults` to check if all parallel branches completed. With a real database, concurrent branches can get stale return values (each only sees its own result), so both return early and nobody advances the workflow — permanent hang.

**Fix:** Re-read from storage via `loadWorkflowSnapshot` after writing.

### Fix 2: Unhandled errors in `process()`
If `updateWorkflowResults` throws, the error escapes `process()`, is silently dropped by `EventEmitterPubSub` (`emit()` doesn't await async listeners), and the workflow hangs because `workflow.fail` is never published.

**Fix:** try/catch around the switch statement in `process()`.

### Fix 3: `processWorkflowFail` itself throwing
If the `workflow.fail` handler throws (e.g. `updateWorkflowState` not implemented on PostgreSQL), the `workflows-finish` event is never published and the execution engine's result promise hangs forever.

**Fix:** Catch block publishes `workflows-finish` directly as a last resort.

### Fix 4: Strip `stepResults` from `parentWorkflow` events
When a nested workflow starts, the event data carried the **full parent `stepResults`** recursively through `parentWorkflow.parentWorkflow`. This data was carried in every pubsub event but never used — all consumers loaded stepResults from storage instead.

For `forEach` with N items and D nesting levels, each event at iteration K carried all K-1 previous results at each nesting level → **O(N² × depth) memory growth**. Heap snapshot showed 297,216 copies of a single execution path string (50 MB).

**Fix:** Remove `stepResults` from all `parentWorkflow` construction sites. Load from storage on nested workflow completion instead.

### Fix 5: Release buffered stream events on workflow completion
**Root cause of the 1.5 GB heap growth.** `WorkflowRunOutput.#bufferedChunks` accumulated every stream event so that `fullStream` could replay history. If nobody read `fullStream` (user only awaited `.result`), all events stayed in memory forever.

Heap snapshot analysis of a user's 3.8 GB snapshot revealed:

| Object Shape | Instances | Self Size |
|---|---|---|
| `{ from, payload, runId, type }` — stream events | **5,377,135** | 287 MB |
| `{ id, payload, status, stepCallId, stepName }` — step payloads | **5,376,214** | 205 MB |
| `{ value }` — value wrappers | **5,074,938** | 194 MB |
| V8 PropertyArrays backing the above | **5,406,266** | 330 MB |
| Concatenated strings in event data | **14,310,523** | 437 MB |

**~1.5 GB** retained in a single `#bufferedChunks` array on one `WorkflowRunOutput` instance.

**Fix:** Clear `#bufferedChunks` in the WritableStream `close()` handler. Also: conditional buffering (skip when active consumer exists), scoped listener removal (prevent listener leaks on `fullStream` cancel).

### Additional findings (not addressed in this PR)
- PostgreSQL `updateWorkflowResults` and `updateWorkflowState` throw "Method not implemented" — Fix 3 prevents the hang, but workflows on PG will always fail at step completion.
- `forEach` with concurrency has the same stale-read race condition (separate `updateWorkflowResults` call).
- Sleep steps use fire-and-forget `setTimeout` — errors in the callback are lost.

## Test plan

- [x] `workflow-timeout-repro.test.ts` — 5 regression tests:
  - **Fix 1:** Mocks `updateWorkflowResults` to return stale data per caller, verifies workflow completes instead of hanging
  - **Fix 2:** Makes `updateWorkflowResults` throw, verifies workflow fails cleanly instead of hanging
  - **Fix 3:** Makes both `updateWorkflowResults` and `updateWorkflowState` throw, verifies workflow fails instead of hanging
  - **Fix 4:** Spies on pubsub.publish, verifies `parentWorkflow.stepResults` is undefined in nested workflow events
  - **Fix 4:** 3-level deep nested workflow completes correctly with stepResults loaded from storage
- [x] 210/210 evented workflow tests passing
- [x] 89/89 stream tests passing

### DIFF
--- a/packages/core/src/stream/RunOutput.ts
+++ b/packages/core/src/stream/RunOutput.ts
@@ -25,6 +25,7 @@ export class WorkflowRunOutput<
   #baseStream: ReadableStream<WorkflowStreamEvent>;
   #emitter = new EventEmitter();
   #bufferedChunks: WorkflowStreamEvent[] = [];
+  #hasActiveConsumer = false;
 
   #streamFinished = false;
 
@@ -71,12 +72,16 @@ export class WorkflowRunOutput<
               },
             } as WorkflowStreamEvent;
 
-            self.#bufferedChunks.push(chunk);
+            if (!self.#hasActiveConsumer) {
+              self.#bufferedChunks.push(chunk);
+            }
             self.#emitter.emit('chunk', chunk);
           },
           write(chunk) {
             if (chunk.type !== 'workflow-step-finish') {
-              self.#bufferedChunks.push(chunk);
+              if (!self.#hasActiveConsumer) {
+                self.#bufferedChunks.push(chunk);
+              }
               self.#emitter.emit('chunk', chunk);
             }
 
@@ -223,6 +228,7 @@ export class WorkflowRunOutput<
     this.#baseStream = stream;
     this.#streamFinished = false;
     this.#consumptionStarted = false;
+    this.#hasActiveConsumer = false;
     this.#status = 'running';
     this.#delayedPromises = {
       usage: new DelayedPromise<LanguageModelUsage>(),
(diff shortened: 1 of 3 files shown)


LekoArts:

PR #13047, opened 2026-02-13
Fix Gemini API errors from empty reasoning parts

### LINKED ISSUE #12980: Gemini API call errors due to empty reasoning message
This issue was created from Discord post 1471289943681925325:

[Open in Discord](https://discord.com/channels/1309558646228779139/1471289943681925325)

Using Gemini 3 flash + a mastra-built agent. I recently turned on reasoning tokens for my agent. Since that change we've seen some "poisoned" conversations due to a message in the history having an emty reasoning part
```
{"role":"assistant","content":[{"type":"reasoning","text":""}]}
```

Our agent has message memory on, so after the user gets 1 bad message, all subsequent messages fail with: `Unable to submit request because it must include at least one parts field, which describes the prompt input`

Ideally we'd filter out these invalid message parts when sending the next message to the agent. However, there's no way I can filter them out on my side unless we manually handle message memory ourselves.

graysonhicks:
## Description

Fixed Gemini API call failures caused by empty reasoning parts stored in message memory. When using Gemini Flash with reasoning tokens enabled, empty reasoning parts `{"type":"reasoning","text":""}` get stored in memory. When these messages are sent back to Gemini on subsequent turns, Gemini rejects them with: `Unable to submit request because it must include at least one parts field`. This poisons the entire conversation — after one bad message, all subsequent messages fail.

**Root cause:** The `reasoning-end` handler in `llm-execution-step.ts` always stored reasoning parts, even when empty and without `providerMetadata`. This created `details: [{ type: 'text', text: '' }]` entries that poisoned conversation history. Additionally, `AIV5Adapter.toUIMessage()` would pass these through because the `part.details?.length` check was truthy even when all text was empty.

**Fix (three layers — source fix + guard fix + safety net):**

1. **`llm-execution-step.ts` (reasoning-end handler)** — Fixed at the source: only store reasoning when there's actual text content or `providerMetadata` (OpenAI needs empty reasoning for `item_reference`). When there's only `providerMetadata` and no text, the `details` array is empty instead of containing `{ type: 'text', text: '' }`.

2. **`AIV5Adapter.ts` (line 175)** — Fixed the guard in `toUIMessage()` to properly retain reasoning parts with: text content (all providers), `providerMetadata` (OpenAI encrypted reasoning), or redacted details (Anthropic/OpenAI).

3. **`output-converter.ts` (lines 77-83)** — Retained as a safety net for legacy stored data that may predate the source fix. Filters empty reasoning parts without `providerMetadata`.

**Tests added (4):**
- Filters empty reasoning parts mixed with text in `prompt()`
- Filters empty reasoning parts in `llmPrompt()`
- Removes assistant message entirely when only part is empty reasoning
- Preserves non-empty reasoning parts (regression guard)

## Related Issue(s)

Fixes #12980

## Type of Change

- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [ ] Test update

## Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented empty reasoning entries from being stored, avoiding Gemini API errors.
  * Preserved reasoning entries when provider metadata is present (e.g., OpenAI-encrypted reasoning).
  * Added a sanitize safety net for legacy stored messages.

* **Tests**
  * Added comprehensive tests to ensure empty reasoning parts are filtered and valid reasoning/metadata are preserved across prompts and LLM flows.


packages/core/src/agent/message-list/adapters/AIV5Adapter.ts
TylerBarnes, line 186: I'm confused about this part, wouldn't this make it happen more often?

packages/core/src/agent/message-list/conversion/output-converter.ts
```diff
@@ -74,6 +74,14 @@ export function sanitizeV5UIMessages(
           if (hasNonEmptyParts) return false;
         }
 
+        // Filter out empty reasoning parts without providerMetadata (Issue #12980)
+        // Gemini with reasoning tokens can produce empty reasoning parts that get stored in memory.
+        // When sent back, Gemini rejects with "must include at least one parts field".
+        // Preserve empty reasoning parts that have providerMetadata (e.g. OpenAI encrypted reasoning).
+        if (p.type === 'reasoning' && (!('text' in p) || !p.text) && !('providerMetadata' in p && p.providerMetadata)) {
+          return false;
+        }
```
TylerBarnes: I think this is the wrong fix, we need to figure out how to retain these parts, because we get them from the api

TylerBarnes (changes requested):

joofsh:
Hey @graysonhicks any update here? I see this branch is closed now but I was eagerly awaiting this change.

PR #13051, opened 2026-02-13
fix: Restore parent agent's MastraMemory after sub-agent tool execution

### LINKED ISSUE #12884: Observational Memory + Sub-Agents: "Exhausted all fallback models" on primary agent's final response
This issue was created from Discord post 1470630439671562240:

[Open in Discord](https://discord.com/channels/1309558646228779139/1470630439671562240)

**SETUP**
I have a primary research agent with **Observational Memory (OM)** that uses **sub-agents** via the agents config (which wraps them as tools). Both the primary agent and its sub-agents have their own OM configured. **I'm routing via Portkey**.

**The Problem**
When the sub-agent **finishes** its work and the primary agent is **about to produce its final response**, the Mastra Studio UI shows:
`{message:Exhausted all fallback models and reached the maximum number of retries.,name:Error}`

**What works**
- OM works correctly on both agents individually — traces confirm observations and reflections fire as expected.
- If I flatten the primary agent (move all tools directly onto it, remove agents config entirely), everything works and the final response is produced.
- Portkey model configs are valid — requests are sent and responses received throughout the conversation. The error only occurs at the final response step when OM + sub-agents are combined.

roaminro:
Fixes #12884

When a parent agent with Observational Memory uses a sub-agent (also with OM via `agents: { researcher: subAgent }`), the sub-agent's `prepare-memory-step` overwrites the shared `RequestContext`'s `MastraMemory` key with its own thread/resource identity. When control returns to the parent, its OM reads the stale sub-agent context, hits a resourceId mismatch, and cascades through all fallback models until it throws "Exhausted all fallback models."

The fix saves and restores the `MastraMemory` key on `RequestContext` around sub-agent tool execution in `listAgentTools()`, so the parent's processors always see the correct memory context on subsequent steps.

Also added a test that sets up a primary agent + sub-agent both with OM, verifies the primary completes successfully, and checks that each agent's OM observations land under the correct thread/resource.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed memory context pollution issue with sub-agents using Observational Memory that caused "Exhausted all fallback models" errors.
  * Ensured parent agent's memory context is properly preserved during sub-agent tool executions.

* **Tests**
  * Added comprehensive test coverage for multi-agent scenarios with Observational Memory to verify proper memory isolation.

### DIFF
--- a/.changeset/slick-poems-post.md
+++ b/.changeset/slick-poems-post.md
@@ -1,0 +1,6 @@
+---
+'@mastra/memory': patch
+'@mastra/core': patch
+---
+
+Fixed sub-agent memory context pollution that caused 'Exhausted all fallback models' errors when using Observational Memory with sub-agents. The parent agent's memory context is now preserved across sub-agent tool execution.
--- a/packages/core/src/agent/agent.ts
+++ b/packages/core/src/agent/agent.ts
@@ -2298,6 +2298,12 @@ export class Agent<
           // manually wrap agent tools with tracing, so that we can pass the
           // current tool span onto the agent to maintain continuity of the trace
           execute: async (inputData: z.infer<typeof agentInputSchema>, context) => {
+            // Save the parent agent's MastraMemory before the sub-agent runs.
+            // The sub-agent's prepare-memory-step will overwrite this key with
+            // its own thread/resource identity. We restore it after the sub-agent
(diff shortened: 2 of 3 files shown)


TylerBarnes (approved):
oh good find! 💪

PR #13086, opened 2026-02-15
fix(auth): `CompositeAuth` ignores `public` and `protected` routes configuration of the underlying providers

### LINKED ISSUE #13083: Bypassing certain routes on Mastra works locally but not in the production environment
This issue was created from Discord post 1472381584916218028:

[Open in Discord](https://discord.com/channels/1309558646228779139/1472381584916218028)

Server auth setup:
```js
auth: new MastraJwtAuth({
        secret: JWT_SECRET,
        public: ["/api/mcp/*"]
    })
```
Endpoint we're hitting: `/api/mcp/tenderIntelligence/mcp`

When hitting that endpoint locally it works but when deployed to prod `*.mastra.cloud` we're hit with `error: Authentication required`

kavins14:
## Description

When deploying to mastra cloud, `CompositeAuth` wraps around the user configured auth providers and it ignores their `public` & `protected` options. This means that a public & protected route configurations won't be respected in production but is in local development.

## Related Issue(s)

fixes #13083

## Type of Change

- [x] Test update

## Checklist

## Summary by CodeRabbit

* **Improvements**
  * CompositeAuth now correctly aggregates public and protected access paths from all configured providers, preserving route configurations.

* **Tests**
  * Added comprehensive tests covering aggregation of mixed path types, handling of missing fields, integration scenarios, short-circuiting, and token/request propagation across providers.

* **Chores**
  * Added changelog entry documenting the patch-level fix.


rphansen91 (approved):
One thing to be aware of: this changes the zero-provider case from undefined to [] for both fields. I checked the middleware helpers and they all guard with || [], so it's functionally equivalent, but worth noting since an empty array is truthy while undefined is not (in case any downstream code does if (authConfig.protected) checks).

PR #13120, opened 2026-02-16
fix(playground): deduplicate traces and workflow runs to prevent table instability

WolffM:
## Summary

Fixes #13057 — duplicate rows in the Logs page table after refetch cycles.

**Root cause:** `useInfiniteQuery` with `refetchInterval: 3000` uses offset-based pagination. When new traces arrive between cycles, page boundaries shift and the same `traceId` appears on multiple pages. The `select` callback did `data.pages.flatMap(page => page)` with no deduplication, producing duplicate React keys. Additionally, `fetchTracesFn` discarded the server's `pagination` metadata (`return res.spans || []`), so `getNextPageParam` used a `!lastPage?.length` heuristic that never terminated for live data.

**What this fixes:**
- Duplicate rows after scrolling/refetch — dedup by `traceId`/`runId` in `select`
- `getNextPageParam` using server's `hasMore` flag instead of length heuristic
- Missing `fetchNextPage` in `useEffect` dependency array
- Removed `gcTime: 0` (traced to commit `eb09742`, a bulk observability refactor — it destroys the query cache on unmount, losing all loaded pages when navigating away; `refetchInterval` already handles freshness)

### Changeset

- `@mastra/playground-ui: patch` — the changeset bot correctly shows 4 packages (`@mastra/playground-ui`, `@internal/playground`, `mastra`, `create-mastra`) because `@internal/playground` is bumped transitively as a dependent, and it's in a `fixed` group with `mastra` and `create-mastra` per `.changeset/config.json`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Removed duplicate workflow run rows and duplicate traces when navigating paginated results.
  * Improved pagination stability and infinite-scroll behavior to reliably fetch next pages without overlap and prevent repeated entries during refetches.

* **Tests**
  * Added unit tests covering pagination logic and cross-page deduplication for runs and traces, including pagination boundary cases and preservation of first-seen items.


WolffM:
the remaining code rabbit comment are trivial

taofeeq-deru (approved):

PR #13126, opened 2026-02-17
fix(core): prevent stream bail for provider-executed tools called in parallel

### LINKED ISSUE #13125: [BUG] Provider-executed tools (web_search) cause stream to bail when called in parallel with regular tools
## Describe the Bug

When a provider-executed tool like Anthropic's `web_search_20250305` is called **in parallel** with regular `createTool` tools, three things go wrong:

1. **Stream bail** — the agentic loop terminates after one step. All tool results are silently dropped and no assistant response is generated.
2. **Web search results never delivered** — even after fixing the bail, the model re-calls `web_search` on the next step because the `server_tool_use` block was stripped from the assistant message sent back to the Anthropic API. The model never receives actual search results.
3. **Conversation resume fails with HTTP 400** — provider-executed tool invocations persist as `state: "call"` in the messageList DB because their results are never written back. On conversation resume, the sanitizer keeps these orphaned `server_tool_use` blocks without matching `web_search_tool_result`, causing the Anthropic API to reject the request.

This affects any agent that mixes provider-executed tools (`providerExecuted: true`) with standard Mastra tools in the same step.

### Impact

- **Zero tool results** emitted to the stream despite tools executing successfully
- **No assistant response** generated (the agentic loop terminates at the tool-call step)
- After fixing the bail, the model **re-calls web_search** in every subsequent step, wasting tokens and API calls, because the deferred search is never executed by the provider
- **Conversation resume breaks** — follow-up messages to the same agent fail with HTTP 400 because the message history contains `server_tool_use` without the corresponding result

### Root Cause

There are three independent bugs in the agentic execution pipeline:

#### Bug 1: Stream bail (tool-call-step → llm-mapping-step)

```
LLM streams response → tool-call-step (foreach tool) → llm-mapping-step (aggregate results)
```

When the AI SDK emits a `tool-call` chunk for a provider-executed tool, `tool-call-step` returns `result: inputData.output`. But `output` is always `undefined` for provider-executed tools — the AI SDK's tool call type doesn't include it. The `llm-mapping-step` then sees `result === undefined` with no error, misidentifies the tool as a pending HITL (human-in-the-loop) interaction, and **bails the entire stream** — dropping results from all parallel tools, including ones that completed successfully.

#### Bug 2: Web search results stripped (message pipeline)

```
llm-execution-step (store tool calls in messageList)
    → MessageMerger (merge assistant messages)
    → AIV5Adapter.toUIMessage (convert to UI format)
    → sanitizeV5UIMessages (filter for LLM prompt)
    → Anthropic API
```

After fixing the bail, the model correctly calls both tools in step 1. But the Anthropic API uses **deferred execution** for `web_search` called in parallel — the `server_tool_use` block appears in step 1's response, but actual search results aren't returned until step 2, after the client sends only the custom tool's result.

The problem: `sanitizeV5UIMessages` filters out `input-available` tool parts when building the LLM prompt (designed to remove incomplete client-side calls). But provider-executed tools legitimately have no client-side result — they stay in `input-available` state. So `server_tool_use(web_search)` gets **stripped from the assistant message**. Without it, the Anthropic API can't execute the deferred search, and the model has to re-call `web_search`.

The `providerExecuted` flag already exists on the AI SDK v5 `UIToolInvocation` type, but it wasn't being propagated through Mastra's message pipeline (DB message → UI part → sanitization filter).

#### Bug 3: Provider-executed tool results never persisted (resume failure)

```
llm-mapping-step (tool results come back)
    → clientExecutedToolCalls only → messageList.add (state: "result")
    → providerExecutedToolCalls → SKIPPED — never written to messageList
```

After the agentic loop completes, `llm-mapping-step` filters provider-executed tools OUT of the tool-result message added to `messageList`. This means provider-executed tool invocations remain stuck at `state: "call"` in the DB forever. On conversation resume:

1. `MessageMerger` converts `state: "call"` → `input-available` UI part with `providerExecuted: true`
2. `sanitizeV5UIMessages(filterIncompleteToolCalls=true)` keeps `input-available + providerExecuted` parts (correct for mid-loop, wrong for resume)
3. The Anthropic API receives `server_tool_use(web_search)` without a matching `web_search_tool_result`
4. **HTTP 400** — the API rejects the malformed request

Additionally, completed provider-executed tools that reached `output-available` state were being kept by the sanitizer and sent back to the LLM, producing `tool_result` blocks for `server_tool_use` IDs — also invalid.

## Steps To Reproduce

Requires an Anthropic model that supports `web_search_20250305` (e.g. `claude-sonnet-4-5-20250929`).

```bash
git clone https://github.com/KaiKloepfer/mastra-provider-tool-bail-repro.git
cd mastra-provider-tool-bail-repro
pnpm install
ANTHROPIC_API_KEY=sk-ant-... npx tsx src/test-resume-bug.ts  # 048048a
```

The repro (`src/test-resume-bug.ts`, commit `048048a`) creates an agent with `web_search_20250305` and a trivial `get_company_info` tool, runs a conversation, inspects the messageList for provider-executed tool state, then attempts a follow-up message to trigger the resume failure.

## Expected Behavior

```
Step 1:
  tool-calls:   [web_search_20250305, get_company_info]
  tool-result:  web_search_20250305 (providerExecuted=true)
  tool-result:  get_company_info: {"name":"Anthropic",...}

Step 2:
  (text generation only — model writes response using real web search results)

Resume:
  Follow-up message succeeds, model responds with context from conversation

PASS [Bug 1] Stream completed with tool-result chunks
PASS [Bug 2] web_search not re-called (deferred search results preserved)
PASS [Bug 3] Resume succeeds — provider-executed tool has state: "result" in DB
```

## Actual Behavior

**Against `@mastra/core@1.4.0`:**

```
Step 1:
  tool-calls:   [web_search_20250305, get_company_info]

Text length: 0 chars

FAIL [Bug 1] Stream bailed: 0 tool-result chunks received
```

Stream bails after step 1. Zero `tool-result` chunks, zero text, no assistant response.

**After fixing Bug 1 + Bug 2 (`1e523aec47` + `1976925c95`), before Bug 3 fix (`09fe94a680`):**

```
PASS [Bug 1] Stream completes
PASS [Bug 2] web_search not re-called
FAIL [Bug 3] Resume failed: provider-executed tool stuck at state: "call"
```

## Fix

See PR #13126.

## Environment Information

```
Node.js: v22.x
@ai-sdk/anthropic: ^3.0.0 => 3.0.44
@mastra/core: ^1.4.0 => 1.4.0
zod: ^3.24.0 => 3.25.76
LLM Provider: Anthropic (claude-sonnet-4-5-20250929)
```

## Verification

KaiKloepfer:
## Summary

Fixes #13125

When a provider-executed tool (e.g. Anthropic `web_search_20250305`) is called **in parallel** with regular `createTool` tools, the agentic execution pipeline has three bugs:

1. **Stream bail** — `tool-call-step` returns `result: undefined` for provider-executed tools, and `llm-mapping-step` misidentifies these as pending HITL interactions and bails the entire stream.

2. **Web search results never delivered** — the `providerExecuted` flag isn't propagated through the message pipeline, so `sanitizeV5UIMessages` strips provider-executed tools from the LLM prompt. Without the `server_tool_use` block in the assistant message, the Anthropic API can't execute its deferred search, and the model re-calls `web_search` on every subsequent step.

3. **Conversation resume fails with HTTP 400** — provider-executed tool invocations are never written back to `messageList` with `state: "result"`, so they persist as `state: "call"` in the DB. On resume, the sanitizer keeps these orphaned `server_tool_use` blocks without matching `web_search_tool_result`, and the Anthropic API rejects the request. Additionally, completed provider-executed tools at `output-available` state were being sent back to the LLM, producing invalid `tool_result` blocks for `server_tool_use` IDs.

**After this PR:** The stream completes, the Anthropic API sees `server_tool_use(web_search)` in the assistant message and executes the deferred search, provider-executed tool results are persisted with `state: "result"`, and conversation resume works correctly.

### Changes

**`1e523aec47` — Prevent stream bail** (`tool-call-step.ts`)

Provide a fallback result when `output` is undefined for provider-executed tools, preventing the bail path:

```diff
- result: inputData.output,
+ result: inputData.output ?? { providerExecuted: true, toolName: inputData.toolName },
```

**`1976925c95` — Propagate `providerExecuted` through the message pipeline** (4 files + tests)

The AI SDK v5 `UIToolInvocation` type already has `providerExecuted?: boolean` — Mastra just wasn't propagating it. This commit connects the plumbing through three layers:

| File | Change |
|------|--------|
| `llm-execution-step.ts` | Store `providerExecuted` on tool-invocation parts in the messageList |
| `AIV5Adapter.ts` | Carry `providerExecuted` through to `ToolUIPart` in both `output-available` and `input-available` branches |
| `output-converter.ts` | Keep `providerExecuted` + `input-available` parts in `sanitizeV5UIMessages` (these are server-executed tools, not incomplete client calls) |
| `llm-mapping-step.ts` | Exclude provider-executed tools from the client tool-result message (prevents sending fabricated results that conflict with provider deferred execution) |

**`09fe94a680` — Persist provider-executed tool results and fix resume sanitization** (`llm-mapping-step.ts`, `output-converter.ts`)

Two changes to fix conversation resume:

| File | Change |
|------|--------|
| `llm-mapping-step.ts` | Add a **separate** `messageList.add` call for provider-executed tools with `state: "result"` and `providerExecuted: true`. `MessageMerger` uses this to update the tool invocation from `state: "call"` to `state: "result"` in the DB, preventing orphaned `server_tool_use` blocks on resume. |
| `output-converter.ts` | Strip `output-available` + `providerExecuted` parts in `sanitizeV5UIMessages` when `filterIncompleteToolCalls=true`. Completed provider-executed tools should not be sent back to the LLM — the provider already handled them internally. Also strip `output-error` + `providerExecuted` parts for the same reason. |

### Unit tests — `1e523aec47` (3 pre-existing tests)
- [x] Provider-executed tool without output returns fallback result
- [x] Provider-executed tool with output passes through actual output
- [x] Provider-executed tools mixed with regular tools don't trigger bail in mapping step

### Unit tests — `1976925c95` (12 tests, all RED→GREEN verified)

**`output-converter-provider-executed.test.ts`** (5 → 8 tests after `09fe94a680`)
- [x] Regular `input-available` parts filtered when `filterIncompleteToolCalls=true`
- [x] Provider-executed `input-available` parts kept
- [x] `output-available` parts kept for client-executed tools
- [x] Mixed mid-loop: client output-available + provider input-available kept, regular input-available dropped
- [x] No filtering when `filterIncompleteToolCalls=false`

**`AIV5Adapter-provider-executed.test.ts`** (4 tests)
- [x] `providerExecuted` propagated on `call` → `input-available`
- [x] `providerExecuted` propagated on `result` → `output-available`
- [x] Not added when absent from DB message
- [x] Mixed: provider-executed tool has flag, regular tool doesn't

**`llm-mapping-step.test.ts`** (3 tests, updated by `09fe94a680`)
- [x] Split client-executed and provider-executed tools into separate messageList entries (2 calls: one for client, one for provider with `providerExecuted: true` and `state: "result"`)
- [x] Provider-executed-only turns produce a single messageList entry with `providerExecuted: true` and `state: "result"`
- [x] Stream chunks still emitted for both tools

### Unit tests — `09fe94a680` (3 new output-converter tests, RED→GREEN verified)

**`output-converter-provider-executed.test.ts`**
- [x] `output-available` + `providerExecuted` parts stripped (completed server tools don't go back to LLM)
- [x] `output-error` + `providerExecuted` parts stripped
- [x] Resume scenario: client `output-available` kept, provider `output-available` stripped in same message

### Existing test suites
- [x] 12/12 llm-mapping-step tests pass
- [x] 8/8 output-converter-provider-executed tests pass
- [x] 20/20 message-list-gemini tests pass
- [x] 28/28 stream tests pass (11 pre-existing skips)
- [x] No type errors

### Live verification with [standalone reproduction](https://github.com/KaiKloepfer/mastra-provider-tool-bail-repro)

| State | Bug 1 (stream bail) | Bug 2 (web search re-called) | Bug 3 (resume fails) |
|-------|--------------------|-----------------------------|--------------------|
| `@mastra/core@1.4.0` (npm) | **FAIL** — 0 tool-results, stream bails after step 1 | unreachable | unreachable |
| `1e523aec47` only | **PASS** — 2 tool-results | **FAIL** — web_search re-called in step 2 | **FAIL** — HTTP 400 |
| `1e523aec47` + `1976925c95` | **PASS** — 3 tool-results | **PASS** — web_search not re-called | **FAIL** — state: "call" persists |
| All three (`1e523aec47` + `1976925c95` + `09fe94a680`) | **PASS** — 3 tool-results | **PASS** — real results delivered | **PASS** — state: "result" in DB, resume succeeds |

## Summary by CodeRabbit

* **Bug Fixes**
  * Provider-executed tools no longer cause bailouts when mixed with regular tools; missing outputs get a fallback and provider results are tracked separately (not treated as pending HITL or returned as completed server tool uses).

* **New Features**
  * Tool-invocation and UI messages may include a providerExecuted flag to mark provider-run tool calls; provider-executed input/results are preserved appropriately.

* **Tests**
  * Added extensive tests covering mixed provider/client flows, result propagation, error scenarios, and filtering behavior.


packages/core/src/agent/message-list/adapters/AIV5Adapter.ts
```diff
@@ -158,6 +159,7 @@ export class AIV5Adapter {
               input: inv.args,
               state: 'input-available',
               callProviderMetadata: part.providerMetadata,
+              ...((part as { providerExecuted?: boolean }).providerExecuted ? { providerExecuted: true } : {}),
```
wardpeet: This is a bit shorter and easier to read

```suggestion
              providerExecuted: ( part as { providerExecuted?: boolean }).providerExecuted
```

packages/core/src/loop/workflows/agentic-execution/llm-execution-step.ts
```diff
@@ -1005,6 +1005,7 @@ export function createLLMExecutionStep<TOOLS extends ToolSet = ToolSet, OUTPUT =
                   args: toolCall.args,
                 },
                 ...(toolCall.providerMetadata ? { providerMetadata: toolCall.providerMetadata } : {}),
+                ...(toolCall.providerExecuted ? { providerExecuted: true as const } : {}),
```
wardpeet: easier to read than the spreading

```suggestion
                providerMetadata: toolCall.providerMetadata,
                providerExecuted: toolCall.providerExecuted
```

packages/core/src/loop/workflows/agentic-execution/llm-mapping-step.ts
```diff
@@ -258,12 +258,17 @@ export function createLLMMappingStep<Tools extends ToolSet = ToolSet, OUTPUT = u
 
         const toolResultMessageId = rest.experimental_generateMessageId?.() || _internal?.generateId?.();
 
+        // Exclude provider-executed tools from the tool-result message. These tools (e.g.
+        // Anthropic web_search) are executed server-side — sending a client-fabricated result
+        // would conflict with the provider's deferred execution.
+        const clientExecutedToolCalls = inputData.filter(toolCall => !toolCall.providerExecuted);
+
         const toolResultMessage: MastraDBMessage = {
```
wardpeet: We should put the whole cosnt in the if case below

wardpeet (commented):
Some small nitpicks but looks good to me. Can you also check code rabbits comment :pray:

KaiKloepfer:
Thanks for the review @wardpeet, I'll try to find some time to circle back around to this in the next few days, but you are also welcome to make any direct modifications you like to the branch if that is faster, access should be enabled.

packages/core/src/loop/workflows/agentic-execution/llm-execution-step.ts
```diff
@@ -1004,7 +1004,8 @@ export function createLLMExecutionStep<TOOLS extends ToolSet = ToolSet, OUTPUT =
                   toolName: toolCall.toolName,
                   args: toolCall.args,
                 },
-                ...(toolCall.providerMetadata ? { providerMetadata: toolCall.providerMetadata } : {}),
+                providerMetadata: toolCall.providerMetadata,
+                providerExecuted: toolCall.providerExecuted,
```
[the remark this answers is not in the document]
TylerBarnes: Reese says: Confirmed there are no `"providerExecuted" in part` checks anywhere in the codebase, so the `in` operator concern is moot. The `MastraMessagePart` type extends `UIMessageV4["parts"][number]` from the AI SDK — adding `providerExecuted` to it would be consistent with how `providerMetadata` was added, but is a broader type-system change that should be its own PR rather than part of this feature PR. The type assertion approach matches the existing pattern used throughout the adapter.

Reese

PR #13140, opened 2026-02-17
fix(core): allow model self-recovery from malformed tool calls

### LINKED ISSUE #9815: Allow model to recover from malformed tool calls
### Problem Statement

Both an API regression and feature request.

In case a model emits a tool call that is not valid JSON or refers to a non-existing tool, the tool call loop now throws an error.

Example stacktrace:
```
Error executing step toolCallStep: Error: Tool wrong_tool_name not found
  at <anonymous> (/path/node_modules/@mastra/core/dist/chunk-EEPYVE6F.js:420:090)
  at execute (/path/node_modules/@mastra/core/dist/chunk-EEPYVE6F.js:420:248)
  at <anonymous> (/path/node_modules/@mastra/core/dist/chunk-EEPYVE6F.js:259:2989)
  at executeStep (/path/node_modules/@mastra/core/dist/chunk-EEPYVE6F.js:259:3747)
  at processTicksAndRejections (native:7:39)

Error executing step step_name: Error: promise 'text' was not resolved or rejected when stream finished
  at <anonymous> (/path/node_modules/@mastra/core/dist/chunk-EEPYVE6F.js:33:2479)
  at forEach (native:1:11)
  at flush (/path/node_modules/@mastra/core/dist/chunk-EEPYVE6F.js:33:2398)
  at <anonymous> (native:1:11)
  at <anonymous> (native:1:11)
  ...
```

Earlier the AI SDK's [experimental_repairToolCall](https://ai-sdk.dev/docs/reference/ai-sdk-core/generate-text#experimental_repair-tool-call) was exposed to get around this, which seems is not anymore supported.

Now there may be an argument that today's tool calling models are reliable enough to make this concern obsolete I can confirm this happening even with Claude Haiku 4.5 (hallucinating not existing tools when input is around 20k tokens, including 20-30 tools)

### Proposed Solution

To support agentic applications, where it is highly preferred to just let the LLM recover from malformed responses (which they can quite robustly can, assuming the error message is descriptive enough), Mastra framework should:
- at a minimum bring back support for `experimental_repairToolCall`
- even better, support (or change to) a behavior, where model just gets returned plain text error responses to malformed tool calls, so they can recover later down the tool call loop

### Component

Tools

### Alternatives Considered

_No response_

### Example Use Case

Allowing today's smart agentic models to recover themselves is I'd say a quite fundamental expectation from any agent execution framework today.

### Additional Context

_No response_

### Verification

graysonhicks:
## Description

When a tool call fails (invalid JSON args, tool execution error, etc.), the agentic loop now continues instead of terminating, letting the model see the error and self-correct.

Two changes:
- `tool-call-step.ts`: Added a guard for null/undefined args (from malformed JSON) that returns a descriptive error instead of calling `tool.execute(undefined)`
- `llm-mapping-step.ts`: Broadened the self-recovery condition from only `ToolNotFoundError` to all tool execution errors — errors are already added to the messageList, so the model sees them on the next turn

## Related Issue(s)

Fixes #9815

## Type of Change

- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [ ] Test update

## Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of tool execution errors so the agent can emit error details and continue/self-recover instead of bailing.
  * Added validation to detect malformed or null JSON tool-call arguments and return a clear error without invoking the tool.

* **Tests**
  * Expanded coverage for tool execution error self-recovery, mixed success/error scenarios, and multiple concurrent errors.
  * Added tests for malformed/null tool-call arguments.


packages/core/src/loop/workflows/agentic-execution/tool-call-step.ts
```diff
@@ -233,6 +233,15 @@ export function createToolCallStep<Tools extends ToolSet = ToolSet, OUTPUT = und
         };
       }
 
+      if (inputData.args == null) {
```
TylerBarnes: what if there's a tool that doesn't take input args? I think also instead of == it'd be better to do `inputData.args === null || typeof inputData.args === "undefined"` since double equals is less clear (in this case it'd match null or undefined)

graysonhicks: good catch.  tools with no input args are fine because the AI SDK always passes {} for argument-less tool calls, so neither null nor undefined would be seen in that case. This should only catche genuinely malformed JSON from the model.

TylerBarnes (approved):

PR #13146, opened 2026-02-17
fix(agent): preserve thread metadata when MASTRA_THREAD_ID_KEY is set

kosukeoya:
When MASTRA_THREAD_ID_KEY is set in RequestContext, the thread metadata passed by the user via options.memory.thread was being completely ignored. This fix preserves the user-provided metadata while still overriding the id with threadIdFromContext for security purposes.

Changes:
- agent.ts: Spread thread object properties when threadIdFromContext exists
- agent-legacy.ts: Apply same fix with existing type casting pattern

## Type of Change

- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [ ] Test update

## Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Thread metadata provided by callers is now preserved even when a thread ID is present in the request context; the context thread ID still takes precedence.

* **Refactor**
  * Thread ID resolution was unified so an explicit context thread ID consistently overrides resolved IDs.

* **Tests**
  * Added comprehensive tests covering thread resolution and override scenarios.


packages/core/src/agent/agent-legacy.ts
```diff
@@ -694,7 +694,10 @@ export class AgentLegacyHandler {
     const threadIdFromContext = requestContext.get(MASTRA_THREAD_ID_KEY) as string | undefined;
 
     const threadFromArgs = threadIdFromContext
-      ? { id: threadIdFromContext }
+      ? {
+          ...(typeof (args.memory as any)?.thread === 'object' ? (args.memory as any).thread : {}),
```
abhiaiyer91: is there a nicer way to do this?

kosukeoya: I changed it so that the thread ID is resolved by ‎`resolveThreadIdFromArgs`.

wardpeet (approved):

PR #13166, opened 2026-02-17
feat(workspace): remove outputSchema, return raw text, extract tools into individual files

CalebBarnes:
## Summary

Closes [COR-525](https://linear.app/kepler-crm/issue/COR-525/prefer-raw-text-over-json-in-workspace-tool-outputs-to-reduce-token)

- **Remove `outputSchema`** from all workspace tools — eliminates the schema definition overhead that AI SDK includes in every tool call
- **Raw text returns** — tools return CLI-style strings instead of structured JSON, reducing token usage
- **Extract 10 workspace tools** from monolithic `tools.ts` (755 lines) into individual files — each tool is now standalone and importable
- **Context-based** — tools use `requireFilesystem(context)` / `requireSandbox(context)` instead of closure-captured workspace
- **UI metadata sideband** — tools emit `data-workspace-metadata` chunks via `context.writer.custom()` for studio UI badges, scoped by `toolCallId` for parallel execution
- **Streaming via data chunks** — execute-command streams stdout/stderr as `sandbox-stdout` / `sandbox-stderr` / `sandbox-exit` message parts, scoped by `toolCallId`
- **Silent command support** — execute command badge now shows running state for commands that produce no output (e.g. `sleep`, `mkdir`)
- **Log improvements** — LocalSandbox logs use `[LocalSandbox]` prefix, `Command completed` downgraded from `info` to `debug`

### Why raw text over JSON?

When tools define `outputSchema`, AI SDK includes the schema definition in the tool call overhead. For workspace tools that return simple results (file contents, "Wrote 42 bytes to /path"), structured JSON with schema validation adds unnecessary tokens. Raw text is more natural for LLM consumption and reduces per-call token cost.

Structured metadata that was previously returned in tool output (workspace info, sandbox status, filesystem provider) is now emitted as `data-workspace-metadata` chunks via `writer.custom()`, keeping it available for UI badge consumption without passing it to the LLM.

### Breaking changes

Workspace tool return values change from structured objects to strings:

| Tool | Before | After |
|------|--------|-------|
| `read_file` | `{ content, size, path }` | `"/path (42 bytes)\n     1→content"` |
| `write_file` | `{ success, path, size }` | `"Wrote 42 bytes to /path"` |
| `edit_file` | `{ success, replacements }` | `"Replaced 1 occurrence in /path"` |
| `list_files` | `{ tree, summary }` | `"tree\n\nsummary"` |
| `delete` | `{ success, path }` | `"Deleted /path"` |
| `file_stat` | `{ exists, type, size, modifiedAt }` | `"/path Type: file Size: 42 bytes Modified: ISO"` |
| `mkdir` | `{ success, path }` | `"Created directory /path"` |
| `search` | `{ results, count, mode }` | `"path: content\n---\nN results (mode search)"` |
| `index` | `{ success, path }` | `"Indexed /path"` |
| `execute_command` | `{ success, stdout, exitCode }` | stdout text or `"stdout\nstderr\nExit code: N"` |

### New exports

Individual tools can now be imported directly:

```ts
import { readFileTool, writeFileTool, editFileTool } from '@mastra/core/workspace';
```

## Test plan

- [x] `packages/core` — 77 tests across 13 test files (per-tool tests, helpers, metadata emission, execute-command data chunks)
- [x] `packages/core` — workspace safety tests pass
- [x] `packages/core` — workspace logger tests pass (updated for `[LocalSandbox]` prefix)
- [x] `packages/core` — builds successfully
- [x] `packages/playground-ui` — builds successfully
- [x] Manual smoke test — streaming output works for execute_command
- [x] Manual smoke test — parallel execute_command calls with toolCallId scoping
- [x] Manual smoke test — error handling (exit codes, timeouts)
- [x] Manual smoke test — file tree badge displays correctly


packages/playground-ui/src/lib/ai-ui/tools/badges/sandbox-execution-badge.tsx
```diff
@@ -51,16 +55,6 @@ export interface SandboxExecutionBadgeProps extends Omit<ToolApprovalButtonsProp
   args: Record<string, unknown> | string;
   result: any;
   metadata?: MastraUIMessage['metadata'];
-  toolOutput?: Array<{
```
NikAiyer, line 54: Was tool output only used by our workspace tools? Was anything else using that?

packages/playground-ui/src/lib/ai-ui/tools/badges/sandbox-execution-badge.tsx
```diff
@@ -146,44 +140,23 @@ const TerminalBlock = ({ command, content, maxHeight = '20rem', onCopy, isCopied
   );
 };
 
-// Extract error message from various possible locations
-// Priority: error.message > error (string) > message > stderr (for failed commands)
-const extractErrorMessage = (result: any): string | null => {
```
NikAiyer, line 151: Same for this error message. What else was using this?

NikAiyer (approved):
LGTM!

packages/playground-ui/src/lib/ai-ui/tools/badges/sandbox-execution-badge.tsx
CalebBarnes: It's still used in ToolBadge (the generic fallback) and AgentBadge. It was only removed from SandboxExecutionBadgeProps because the sandbox badge doesn't need it — it reads streaming data directly from message data parts instead. The prop still exists on the other badges that need it. Nothing else broke.

packages/playground-ui/src/lib/ai-ui/tools/badges/sandbox-execution-badge.tsx
CalebBarnes: No other references in the codebase. It was only used inside SandboxExecutionBadge itself. Replaced it with the `exitChunk.data.success` check from the data chunks, which is more reliable.

PR #13178, opened 2026-02-18
fix(pg): set replica identity on workflow_snapshot table

### LINKED ISSUE #13097: Replica identity
This issue was created from Discord post 1472910209272774667:

[Open in Discord](https://discord.com/channels/1309558646228779139/1472910209272774667)

Hello guys, we are hitting a blocker in our environment with this one

error: cannot update table "mastra_workflow_snapshot" because it does not have a replica identity and publishes updates

Should you include in the pg library the addition of this replica identity? my suggestion is to use the already unique key you have there

ALTER TABLE mastra_workflow_snapshot REPLICA IDENTITY USING INDEX public_mastra_workflow_snapshot_workflow_name_run_id_key

what do you think?

intojhanurag:
## Description

The `mastra_workflow_snapshot` table does not have a PRIMARY KEY and instead relies on a UNIQUE constraint on `(workflow_name, run_id)`.  
This causes PostgreSQL logical replication to fail with the error:

> cannot update table "mastra_workflow_snapshot" because it does not have a replica identity and publishes updates

This change sets **REPLICA IDENTITY USING INDEX** on the existing unique constraint index so PostgreSQL can correctly identify rows during replication.  

All other Mastra tables already have PRIMARY KEYs and are unaffected.

## Related Issue(s)

Fixes #13097

## Type of Change

- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [ ] Test update

## Checklist

## Test Plan

- [x] Existing constraint-utils tests pass (24/24)
- [x] Build succeeds
- [ ] Manual verification with logical replication enabled PostgreSQL instance

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed workflow snapshot storage compatibility with PostgreSQL logical replication. The workflow snapshot table now correctly tracks row updates during replication, preventing errors that occurred when replication was enabled on schemas lacking proper replication identity.


NikAiyer (commented):
Could you add a unit test for this as well?

Something like:

```
  it('includes REPLICA IDENTITY USING INDEX for workflow snapshot table', () => {
    const sql = generateTableSQL({
      tableName: TABLE_WORKFLOW_SNAPSHOT,
      schema: TABLE_SCHEMAS[TABLE_WORKFLOW_SNAPSHOT],
    });
    expect(sql).toContain('REPLICA IDENTITY USING INDEX');
  });
```
  And a second test verifying it works with a custom schema name too.

NikAiyer (approved):

PR #13253, opened 2026-02-19
fix(schema-compat): fix zodToJsonSchema routing for v3/v4 schemas

abhiaiyer91:
## Summary
- Fixes `zodToJsonSchema` to reliably detect Zod v3 vs v4 schemas regardless of which version the ambient `zod` import resolves to
- Explicitly imports `z as zV4` from `zod/v4` and routes based on the schema's `_zod` property instead of checking `'toJSONSchema' in z`
- Migrates all mastracode tool files from `zod/v3` to `zod` imports now that schema-compat handles both correctly

## Problem
The previous implementation relied on `'toJSONSchema' in z` where `z` came from `import { z } from 'zod'`. This import resolves to v3 (3.25.76) in the monorepo but v4 (4.3.6) when installed globally via npm, causing environment-dependent failures:
- **Global install**: v3 schemas routed to v4's `toJSONSchema()` → crash: "Cannot read properties of undefined (reading 'def')"
- **Monorepo**: v4 schemas routed to v3's converter → schemas missing the `type` field, rejected by providers

## Fix
- Import `z as zV4` from `zod/v4` explicitly to guarantee access to `toJSONSchema`
- Route based on `zodSchema._zod` (present only on v4 schemas) instead of ambient import detection
- All 709 schema-compat tests pass

## Test plan
- [x] Build `@mastra/schema-compat` and run test suite (709 tests pass)
- [x] Test mastracode locally with v3 tool schemas — works
- [x] Test mastracode locally with v4 tool schemas (all tools migrated) — works
- [x] Verify tool calls succeed end-to-end (file operations, search, shell, etc.)

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed Zod v3/v4 schema routing so schemas are reliably detected and processed by the correct converter, preventing misrouting and improving compatibility and stability.

* **Chores**
  * Updated internal tooling imports to align with the fix and added a changeset documenting the patch releases.

### DIFF
--- a/packages/schema-compat/src/zod-to-json.ts
+++ b/packages/schema-compat/src/zod-to-json.ts
@@ -1,6 +1,6 @@
 import type { JSONSchema7 } from 'json-schema';
-import { z } from 'zod';
 import type { ZodSchema as ZodSchemaV3 } from 'zod/v3';
+import { z as zV4 } from 'zod/v4';
 import type { Targets } from 'zod-to-json-schema';
 import zodToJsonSchemaOriginal from 'zod-to-json-schema';
 
@@ -27,7 +27,7 @@ function patchRecordSchemas(schema: any): any {
     // The bug: z.record(valueSchema) puts the value in keyType instead of valueType
     // Fix: move it to valueType and set keyType to string (the default)
     def.valueType = def.keyType;
-    def.keyType = (z as any).string();
+    def.keyType = zV4.string();
   }
 
   // Recursively patch nested schemas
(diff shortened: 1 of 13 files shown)


rase- (approved):

packages/schema-compat/src/zod-to-json.ts
```diff
@@ -188,13 +188,16 @@ export function zodToJsonSchema(
...
 ): JSONSchema7 {
-  const fn = 'toJSONSchema';
-
-  if (fn in z) {
+  // Route based on whether the schema is v4 (has _zod) or v3 (only has _def).
+  // We use zV4.toJSONSchema (imported from 'zod/v4') for v4 schemas, since the
+  // default 'zod' import may resolve to v3 depending on the environment.
+  // Without this check, v3 schemas passed to v4's toJSONSchema would throw
+  // "Cannot read properties of undefined (reading 'def')".
+  if (zodSchema?._zod) {
```
wardpeet, line 196: Are you sure this fixes it cause it does exactly like what was previous

wardpeet (approved):

PR #13286, opened 2026-02-19
 feat: support metadata-only queries in vector stores

### LINKED ISSUE #13188: [Feature request + Bug report] How to query vector store with metadata filter only?
This issue was created from Discord post 1473626775065071666:

[Open in Discord](https://discord.com/channels/1309558646228779139/1473626775065071666)

I am using a postgres database with the pgvector extension. I can query records semantically using embeddings and cosine similarity, but there are also cases where i want to query using metadata filters only. Is there some kind of trick or workaround for this? This would be a nice feature.

I also found a **small inconsistency in the docs:** The `.query()` method of the `PgVector` class expects an input parameter `queryVector: number[]` but the latest [Mastra docs](https://mastra.ai/reference/vectors/pg#query:~:text=index%20to%20query-,vector%3A,number%5B%5D,-Query%20vector) (18th february 2026) still state that the parameter is called `vector: number[]`.

intojhanurag:
## Description

This PR enables metadata-only queries for vector stores that support them.

- Makes `queryVector` optional in the `QueryVectorParams` interface.
- Implements full metadata-only query support in **PgVector**:
  - Allows querying by `filter` without providing an embedding.
  - Returns results with `score: 0` since no similarity is computed.
  - Validates that at least one of `queryVector` or `filter` is provided.
  - Keeps the existing similarity search path unchanged.
- Adds explicit guards in the other 15 vector stores to throw:
  `"Metadata-only queries are not supported by this vector store"` when `queryVector` is omitted.
- Fixes documentation bug where `query()` parameter was incorrectly named `vector` instead of `queryVector` in `pg.mdx` and `pinecone.mdx`.
- Marks `queryVector` as optional in `pg.mdx`.

Includes 9 new PgVector integration tests covering:
- Simple and complex metadata filters
- `topK` handling
- `includeVector` behavior
- Proper error cases
- Regression check for combined `queryVector + filter`

## Related Issue(s)

Fixes #13188

## Type of Change

- [x] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [x] Test update

## Checklist

## Summary by CodeRabbit

* **New Features**
  * PostgreSQL vector store: support metadata-only queries using filters without a query vector.

* **Improvements**
  * Docs: parameter renamed to "queryVector" and usage clarified.
  * Many vector stores now return clear, consistent user-facing errors when a query vector is required.

* **Tests**
  * Added metadata-only query tests for the PostgreSQL vector store.

* **Chores**
  * Package version bumps.

### DIFF
--- a/.changeset/feat-metadata-only-vector-query.md
+++ b/.changeset/feat-metadata-only-vector-query.md
@@ -1,0 +1,23 @@
+---
+"@mastra/core": minor
+"@mastra/pg": minor
+"@mastra/pinecone": patch
+"@mastra/qdrant": patch
+"@mastra/chroma": patch
+"@mastra/astra": patch
+"@mastra/upstash": patch
+"@mastra/mongodb": patch
+"@mastra/elasticsearch": patch
+"@mastra/opensearch": patch
+"@mastra/duckdb": patch
+"@mastra/turbopuffer": patch
+"@mastra/vectorize": patch
+"@mastra/convex": patch
+"@mastra/couchbase": patch
+---
+
+Support metadata-only queries in vector stores by making `queryVector` optional in the `QueryVectorParams` interface.
+
+PgVector now supports querying by metadata filters alone without providing a query vector — useful when you need to retrieve records by metadata without performing similarity search. Other vector stores will throw a clear error if `queryVector` is omitted, since their backends require a vector for queries.
+
+Also fixes documentation where the `query()` parameter was incorrectly named `vector` instead of `queryVector`.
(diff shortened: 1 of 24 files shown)


NikAiyer (commented):
This looks good to me! I think we just need to make separate changesets for each package, and maybe move the docs changes to another PR, since those get published immediately. The naming fix in the docs can probably remain though. 

Also can you check out the additional coderabbit comments and see if they're worth addressing?

Edit: Actually most of the stores changesets are fine to be grouped together. I think @mastra/core and @mastra/pg should be separated though.

NikAiyer (approved):
LGTM!

PR #13311, opened 2026-02-20
fix(mastracode): wire mcpManager to TUI so /mcp command works

RainerAtSpirit:
## Problem

The `/mcp` slash command always shows **"MCP system not initialized"** even when MCP servers are configured and working. This is because the handler calls `this.harness.getMcpManager?.()` — a method that doesn't exist on the `Harness` class — so it always returns `undefined`.

MCP tools work fine during conversations (they're injected via `createDynamicTools(mcpManager)`), but the `/mcp` status display and `/mcp reload` are broken.

## Fix

Pass `mcpManager` directly to `MastraTUI` via `MastraTUIOptions`, following the same pattern already used for `hookManager` and `authStorage`.

**2 files, +5/-1 lines:**
- `mastracode/src/tui/mastra-tui.ts` — add `mcpManager?` to options interface, store as class property, use in `/mcp` handler
- `mastracode/src/main.ts` — pass `mcpManager` to TUI constructor

## Test Plan

- All existing tests pass (9/9)
- Zero new TypeScript errors introduced
- Verified `/mcp` handler now receives the manager instance

## Summary by CodeRabbit

* **New Features**
  * Improved MCP integration in the CLI/UI initialization to better surface server status.

* **Bug Fixes**
  * Fixed the /mcp slash command always showing "MCP system not initialized"; status now reflects real server state and `/mcp reload` updates status correctly.

### DIFF
--- a/.changeset/fix-mcp-status-display.md
+++ b/.changeset/fix-mcp-status-display.md
@@ -2,6 +2,6 @@
 "mastracode": patch
 ---
 
-Fix /mcp slash command always showing "MCP system not initialized"
+Fix /mcp slash command now correctly displays MCP server status
 
-The `/mcp` command was calling `this.harness.getMcpManager?.()` but the Harness class has no such method, so it always returned undefined. Fixed by passing the `mcpManager` instance directly to the TUI via `MastraTUIOptions`, following the same pattern used for `hookManager` and `authStorage`.
+The `/mcp` command always showed "MCP system not initialized" even when MCP servers were configured and working. Server status and `/mcp reload` now work as expected.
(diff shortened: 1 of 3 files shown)


RainerAtSpirit:
Thanks for the thorough review @coderabbitai.

**Import style + changeset wording**: Fixed in 21b3734.

**`stop()` and MCP disconnect**: Valid observation, but this is outside the scope of this PR — `stop()` is pre-existing code we didn't modify. MCP disconnect is already handled in `main.ts` via the SIGINT/SIGTERM signal handlers (lines 62–78). Whether `stop()` should also own cleanup is a separate discussion/refactor.

abhiaiyer91 (approved):

PR #13322, opened 2026-02-20
fix(observability): handle non-Set keysToStrip in deepClean()

### LINKED ISSUE #13295: Bug: keysToStrip.has is not a function — any agent with memory crashes on Mastra Cloud
This issue was created from Discord post 1474146084929536111:

[Open in Discord](https://discord.com/channels/1309558646228779139/1474146084929536111)

@mastra/core@1.5.0 + @mastra/observability@1.2.0 + mastra@1.3.2                               

  Minimal reproduction

  // index.ts — deploy this to Mastra Cloud
  import { openai } from '@ai-sdk/openai';

  const memory = new Memory({
    storage: new PostgresStore({ connectionString: process.env.DATABASE_URL! }),
  });

  const agent = new Agent({
    id: 'test-agent',
    name: 'Test Agent',
    instructions: 'Say hello.',
    model: openai('gpt-4o-mini'),
    memory, // <-- removing this makes the crash go away
  });

    agents: { agent },
    bundler: { externals: true },
  });

  Then call the agent:

  curl -X POST https://your-cloud-url/api/agents/test-agent/generate \
    -H "Content-Type: application/json" \
    -d '{"messages":[{"role":"user","content":"Hello"}]}'

  Result: Agent fails immediately with [Agent:Test Agent] - Input processor error

  Expected: Agent responds normally.

  Works locally with mastra dev. Only crashes when deployed to Mastra Cloud.

  What happens internally

  Adding memory to an agent causes @mastra/core to auto-add a MessageHistory input processor.
  After that processor runs, the ProcessorRunner finalizes the span:

  @mastra/core ProcessorRunner.runInputProcessors()
    → processorSpan.end({ output })           // line 1562 of chunk-MZIMV2BB.js — NO try-catch
      → DefaultSpan.end()                      // @mastra/observability line 7781
        → deepClean(options.output)            // @mastra/observability line 7787
          → keysToStrip.has(key)               // @mastra/observability line 7524 — CRASH

  keysToStrip is defined as new Set([...]) at line 7371 of @mastra/observability/dist/index.js.
  After Cloud bundling/deployment, it's no longer a Set, so .has() doesn't exist.

roaminro:
Bundlers (e.g. when deploying agents with memory to Mastra Cloud) can transform `new Set([...])` into a plain object or array, which makes `keysToStrip.has(key)` blow up in `deepClean()`.

The fix builds a resilient lookup function once at `deepClean()` scope that handles all three forms:

```ts
const strip: unknown = keysToStrip;
const shouldStripKey =
  strip instanceof Set
    ? (key: string) => (strip as Set<string>).has(key)
    : Array.isArray(strip)
      ? (key: string) => (strip as string[]).includes(key)
      : (key: string) => key in (strip as Record<string, unknown>);
```

Tests confirm the crash reproduces without the fix and passes with it.

Closes #13295

## Summary by CodeRabbit

* **Bug Fixes**
  * Resolved a crash for agents with memory on Mastra Cloud by making key-stripping tolerant of bundler-produced formats.

* **Tests**
  * Added tests to verify key-stripping handles different formats created by bundlers.

* **Chores**
  * Added a patch release entry documenting the fix.

### DIFF
--- a/observability/mastra/src/spans/serialization.ts
+++ b/observability/mastra/src/spans/serialization.ts
@@ -196,6 +196,16 @@ function compressJsonSchema(schema: any, depth: number = 0): any {
 export function deepClean(value: any, options: DeepCleanOptions = DEFAULT_DEEP_CLEAN_OPTIONS): any {
   const { keysToStrip, maxDepth, maxStringLength, maxArrayLength, maxObjectKeys } = options;
 
+  // Bundlers can transform `new Set([...])` into a plain object or array,
+  // so we build a resilient lookup once (no per-object overhead).
+  const strip: unknown = keysToStrip;
+  const shouldStripKey: (key: string) => boolean =
+    strip instanceof Set
+      ? (key: string) => (strip as Set<string>).has(key)
+      : Array.isArray(strip)
+        ? (key: string) => (strip as string[]).includes(key)
+        : (key: string) => key in (strip as Record<string, unknown>);
+
   const seen = new WeakSet<any>();
 
   function helper(val: any, depth: number): any {
@@ -294,7 +304,7 @@ export function deepClean(value: any, options: DeepCleanOptions = DEFAULT_DEEP_C
     let keyCount = 0;
 
     for (const [key, v] of entries) {
-      if (keysToStrip.has(key)) {
+      if (shouldStripKey(key)) {
         continue;
       }
 
--- a/observability/mastra/src/spans/serialization.ts
+++ b/observability/mastra/src/spans/serialization.ts
@@ -41,7 +41,7 @@ export const DEFAULT_KEYS_TO_STRIP = new Set([
 ]);
 
 export interface DeepCleanOptions {
-  keysToStrip: Set<string>;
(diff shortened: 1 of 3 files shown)


observability/mastra/src/spans/serialization.ts
```diff
@@ -196,6 +196,15 @@ function compressJsonSchema(schema: any, depth: number = 0): any {
...
 
+  // Bundlers can transform `new Set([...])` into a plain object or array,
+  // so we build a resilient lookup once (no per-object overhead).
+  const shouldStripKey: (key: string) => boolean =
+    keysToStrip instanceof Set
+      ? (key: string) => keysToStrip.has(key)
+      : Array.isArray(keysToStrip)
+        ? (key: string) => keysToStrip.includes(key)
+        : (key: string) => key in keysToStrip;
+
```
epinzur, line 206: alternatively could we convert keysToStrip to a Set when we parse DeepCleanOptions on startup?

SET.has() is a O(1) operation, while ARRAY.includes is a O(n) operation.

epinzur (approved):
approved with one comment. Feel free to leave as is or update.

PR #13338, opened 2026-02-20
fix(schema-compat): handle ZodNull in Anthropic and OpenAI Reasoning providers

### LINKED ISSUE #13315: bug(schema-compat): MCP servers with null types in tool schemas crash mastracode
## Summary

MCP servers that use `{ "type": "null" }` in their tool JSON Schemas cause mastracode to crash on startup with:

```
Error: claude-opus-4-6 does not support zod type: ZodNull
```

## Reproduction

1. Configure any MCP server that has `null` types in its tool schemas (e.g. `chrome-devtools-mcp`)
2. Start mastracode
3. Crash on startup

## Root cause

`@mastra/schema-compat` lists `ZodNull` as an unsupported type and throws unconditionally. When `@mastra/mcp` converts MCP tool JSON Schemas to Zod, `{ "type": "null" }` becomes `z.null()`, which then hits this error in schema-compat.

`{ "type": "null" }` is valid JSON Schema. MCP servers using it shouldn't crash the client.

## Expected behavior

`@mastra/schema-compat` should coerce `ZodNull` to a compatible type (e.g. `z.any().optional()`) rather than throwing, or `@mastra/mcp` should handle the conversion before schemas reach the compatibility layer.

## Environment

- mastracode from source (main branch, commit 19b07c02e2)
- MCP server: `chrome-devtools-mcp@latest` (via npx)
- Same server works fine in Claude Code

intojhanurag:
## Description

When an MCP server (e.g. `chrome-devtools-mcp`) declares `{ "type": "null" }` in its tool JSON Schema, Mastra's `@mastra/mcp` package converts it to `z.null()`. The `@mastra/schema-compat` layer then throws unconditionally because `ZodNull` is listed as unsupported for both Anthropic and OpenAI Reasoning providers.

This PR adds a `ZodNull` handler to both providers that coerces `z.null()` into `z.any().refine(v => v === null).describe('must be null')`. This preserves the null-only constraint at runtime while using a Zod type that all LLM providers accept. Any existing `.describe()` on the original `z.null()` is also preserved.

## Related Issue(s)

Fixes #13315

## Type of Change
- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [x] Test update

## Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Schemas containing explicit null types no longer cause startup crashes for Anthropic and OpenAI Reasoning tool integrations; null fields are now accepted and validated.

* **Tests**
  * Added test coverage verifying null-type handling, optional nulls, validation behavior, and preservation of field descriptions.

* **Documentation**
  * Added a changelog entry documenting the null-handling patch.


packages/schema-compat/src/provider-compats/anthropic.ts
NikAiyer, line 53: @intojhanurag Doesn't the isNull check already exist below here? Same for the openai-reasoning. Do we still need both?

NikAiyer (commented):
Should we apply this to all providers equally? I think right now that check exists for antrhopic, google, and openai-reasoning, and you added additional checks to anthropic and openai-reasoning. Should we just handle this in the base class instead?

PR #13363, opened 2026-02-21
fix: resolve skill file paths relative to skill root, not hardcoded subdirectories

### LINKED ISSUE #13270: [FEATURE] `skill-read-reference`: resolve paths relative to skill root, not hardcoded `./references/` directory
### Problem Statement

`skill-read-reference` hardcodes `references/` as a mandatory path prefix inside `getReference()` in `workspace-skills.ts`. This means the tool can only read files that live under `<skill>/references/`, but the AgentSkills spec defines file references as **relative paths from the skill root** — e.g. `references/colors.md`, `docs/schema.md`, or `./config.json` are all valid.

When an agent reads a SKILL.md body and encounters a backlink like `See [schema](docs/schema.md)`, calling `skill-read-reference` with `referencePath: "docs/schema.md"` silently fails because it resolves to `<skill>/references/docs/schema.md` instead of `<skill>/docs/schema.md`. The model has no way to follow spec-compliant backlinks.

Many skills come packaged with `./rules` rather than `./references`

### Proposed Solution

Change `skill-read-reference` (and `getReference()` in `workspace-skills.ts`) to resolve `referencePath` relative to the **skill root**, not a hardcoded `references/` subdirectory. The `references/` folder should remain a recommended convention per the spec, but must not be a hard constraint enforced by the implementation.
The path traversal guard (`#assertRelativePath`) should be kept — just anchor it to the skill root rather than the `references/` subfolder.

### Component

Agents, Tools

### Alternatives Considered

Skill authors can work around this today by putting everything under `references/` and writing SKILL.md backlinks as `references/foo.md` — but this conflicts with the AgentSkills spec's intent and prevents skill reuse across other spec-compliant agents that don't impose this restriction.

### Example Use Case

As a skill author, I want to write a SKILL.md with `See [colors](references/colors.md)` or `See [schema](docs/schema.md)` and have the agent resolve either path correctly via `skill-read-reference`, so my skill is portable across any AgentSkills-compatible runtime.

### Additional Context

Relevant code: [`packages/core/src/workspace/skills/workspace-skills.ts` line 307](https://github.com/mastra-ai/mastra/blob/b9035521d0e7655dec481fe6926ba292d01f2184/packages/core/src/workspace/skills/workspace-skills.ts#L307):
```ts
const refFilePath = this.#joinPath(skill.path, 'references', safeRefPath);
```
Should be:
```ts
const refFilePath = this.#joinPath(skill.path, safeRefPath);
```
The same issue applies symmetrically to `scripts/` and `assets/` — worth auditing those too.

### Verification

netbrah:
## Description

`getReference()` (and symmetrically `getScript()`, `getAsset()`) unconditionally prepended `references/`, `scripts/`, or `assets/` to the caller-supplied path, making it impossible to reference files outside those exact subdirectories. The AgentSkills spec treats `referencePath` as skill-root-relative.

### Changes

- **`packages/core/src/workspace/skills/workspace-skills.ts`** — Resolve paths relative to skill root instead of hardcoded subdirectories
- **`packages/core/src/workspace/skills/workspace-skills.test.ts`** — Updated tests for new path resolution
- **`packages/core/src/processors/processors/skills.ts`** — Aligned processor with new path handling

## Related Issue(s)

Fixes #13270

## Type of Change

## Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed path resolution in skill reference, script, and asset tools. Files are now resolved relative to the skill root directory instead of fixed subdirectories, enabling access to files at any location (e.g., docs/schema.md, references/colors.md, config.json) while maintaining protection against path traversal attacks.


abhiaiyer91 (approved):

NikAiyer (approved):

NikAiyer:
thanks @netbrah for making this PR!

PR #13478, opened 2026-02-25
fix(agent): use unique runId per parallel workflow tool call (#13473)

echoVic:
## Summary

Fixes #13473 — parallel workflow tool calls share the same Run instance due to agent runId reuse in `createRun`.

## Root Cause

In `listWorkflowTools`, the `execute` closure captures the agent's `runId` and passes it to `workflow.createRun()`. When the LLM makes multiple parallel tool calls to the same workflow, all calls use the same `runId`, so `createRun()` returns the cached Run instance from the `#runs` Map. Only the first `run.start()` processes its input; the rest receive identical results.

## Fix

Generate a unique `runId` (via `randomUUID()`) for each non-resume workflow tool call:

```diff
- const runIdToUse = suspendedToolRunId || runId;
+ const runIdToUse = suspendedToolRunId || randomUUID();
```

This ensures each parallel invocation creates an independent Run instance. Resume cases still use `suspendedToolRunId` to correctly identify the existing run.

## Verification

- `randomUUID` is already imported at the top of the file
- The resume path (`suspendedToolRunId`) is unchanged
- Matches the workaround confirmed working by the issue reporter

## Summary by CodeRabbit

* **Bug Fixes**
  * Each parallel workflow/tool invocation now receives a unique execution identifier to avoid reusing cached runs and cross-call interference.
  * Suspension and resume now correctly track and propagate the proper run identifier (including workflow-type tools), so resumed runs restore prior state while new runs start fresh, improving concurrency, traceability, and logs.


packages/core/src/agent/agent.ts
```diff
@@ -2744,7 +2744,10 @@ export class Agent<
           execute: async (inputData, context) => {
             try {
               const { initialState, inputData: workflowInputData, suspendedToolRunId } = inputData as any;
-              const runIdToUse = suspendedToolRunId || runId;
+              // Use a unique runId for each workflow tool call to prevent parallel calls
+              // from sharing the same cached Run instance (see #13473).
+              // For resume cases, suspendedToolRunId correctly identifies the existing run.
```
taofeeq-deru: this is only valid for when resuming via `autoResumeSuspendedTools` though. When you resume using `agent.resumeStream()` or `agent.resumeGenerate()`, `suspendedToolRunId` won't exist because the llm won't add it to the tool inputData. It's only added if `autoResumeSuspendedTools: true` .

I think you'll have to pass the `runId` from the workflowRun as one of the `suspendOptions` here
https://github.com/echoVic/mastra/blob/65b9f42e9d2babd2e573f024df1ed97f0e26bd90/packages/core/src/agent/agent.ts#L2848.

then in tool-call-step - https://github.com/mastra-ai/mastra/blob/death-to-agent-network/packages/core/src/loop/workflows/agentic-execution/tool-call-step.ts#L30, you'll have to make sure it's saved to memory and get it from there to pass to the tool args when resuming a workflow tool.

You can run the tool-approval.e2e.test.ts to confirm that workflow suspension and resumption works fine for both autoResume and manual resumption

echoVic: Good catch! You're right — `suspendedToolRunId` is only injected when `autoResumeSuspendedTools: true`.

I've updated the logic to handle all three cases:

```ts
const runIdToUse = suspendedToolRunId || (resumeData ? runId : randomUUID());
```

- **autoResume**: uses `suspendedToolRunId` from inputData
- **manual resume** (resumeStream/resumeGenerate): falls back to the outer `runId`
- **fresh call**: generates a new `randomUUID()` to prevent parallel sharing

Pushed in 4e692af.

taofeeq-deru (changes requested):

packages/core/src/agent/agent.ts
taofeeq-deru: Hi, sorry, was updating the comment. Didn't know you were working on it already. Can you confirm that all work fine by running this test please - tool-approval.e2e.test.ts to?
I believe it should catch it if the resumption doesn't work fine

taofeeq-deru: Also, unfortunately, the `runId` for when there's `resumeData` won't be the same one for when the workflow tool actually suspended, because that was a `randomUUID` and not the actual agent's `runId`, so i'm not sure it'll find a suspended workflow to resume

echoVic: Good point — that's exactly what the latest push (263ad3c) addresses!

Instead of falling back to the outer `runId` for manual resume, the workflow tool now passes `runId: runIdToUse` via `suspendOptions` when it suspends. `tool-call-step` stores this in message metadata (same pattern as agent tools), and on manual resume it retrieves the `suspendedToolRunId` from metadata and injects it into the tool args.

So the flow is:
1. Fresh call → `randomUUID()` generates `runIdToUse`
2. Workflow suspends → `runIdToUse` stored in metadata via `addToolMetadata`
3. Manual resume → `tool-call-step` finds `suspendedToolRunId` from metadata → injects into `args.suspendedToolRunId`
4. Workflow tool receives `suspendedToolRunId` → uses it to `createRun` with the correct cached Run

I wasn't able to run `tool-approval.e2e.test.ts` locally (3.8GB RAM server can't install the full monorepo). Could you verify via CI or locally?

taofeeq-deru: tried via CI, it failed
So I checked locally, the `suspendedToolRunId` gets removed from the tool inputData because it isn't part of the inputSchema. Need to update the builder to make sure it is - https://github.com/mastra-ai/mastra/blob/main/packages/core/src/tools/tool-builder/builder.ts#L61.

Good work by the way!

echoVic: Great catch! Pushed aaf9681 — added `workflow-` prefix check in `CoreToolBuilder` (builder.ts:61) alongside the existing `agent-` check, so `suspendedToolRunId` is now included in the inputSchema for workflow tools too.

Also updated the changeset to be more outcome-focused per coderabbit's suggestion.

I can't run `tool-approval.e2e.test.ts` locally (3.8GB RAM), would appreciate if you could verify via CI. Thanks for the thorough review! 🙏

echoVic:
Addressed CodeRabbit's feedback in 7690410: hoisted `isWorkflowTool` before `resumeDataToPassToToolOptions` and added it to the exclusion guard. This ensures approval-resumed workflow tools don't skip the `suspendedToolRunId` metadata lookup.

Regarding CI failures:
- **Prebuild (Build)**: fails at "Setup custom action" — infra issue, not code-related
- **Core Package Tests**: Gemini e2e tests fail due to `resumeData` schema having `{ type: 'array' }` without `items` (Gemini API rejects it with 400). This is a pre-existing schema compat issue, not caused by this PR.
- **flakeyTool test**: known flaky test, unrelated

packages/core/src/loop/workflows/agentic-execution/tool-call-step.ts
```diff
@@ -363,8 +363,9 @@ export function createToolCallStep<Tools extends ToolSet = ToolSet, OUTPUT = und
         // For agent tools, always pass resume data so the agent tool wrapper knows to call
         // resumeStream instead of stream (otherwise the sub-agent restarts from scratch)
         const isAgentTool = inputData.toolName?.startsWith('agent-');
+        const isWorkflowTool = inputData.toolName?.startsWith('workflow-');
         const resumeDataToPassToToolOptions =
-          !isAgentTool && toolRequiresApproval && Object.keys(resumeData).length === 1 && 'approved' in resumeData
+          !(isAgentTool || isWorkflowTool) && toolRequiresApproval && Object.keys(resumeData).length === 1 && 'approved' in resumeData
```
taofeeq-deru: oh this actually isn't needed. coderabbit was wrong about that.
The agent tool (subagent) might have gotten suspended because it has some tool in it that requires approval, that's why we have to make sure to confirm that the tool isn't a subagent, to ensure the resumeData { approved: true/false } is sent down to the subagent.

This does not apply to workflows

taofeeq-deru (approved):

Issue #12757, opened 2026-02-06
[BUG] Tool input validation failed for mastra_workspace_execute_command. Please fix the following errors and try again


sandysong:
### Describe the Bug

使用GLM4.7模型生成mastra_workspace_execute_command的工具参数时生成内容如下：
```
{
  "command": "python3",
  "args": "[\"parse_excel.py\"]",
  "timeout": 60000
}
```
其中args被当成了字符串，导致输入参数校验失败

### Steps To Reproduce

简单的让Agent执行一个python命令。

### Link to Minimal Reproducible Example

无

### Expected Behavior

无

### Environment Information

```block
无
```

### Verification

- [x] I have searched the existing issues to make sure this is not a duplicate
- [x] I have included sufficient information for the team to reproduce and understand the issue

Issue #11457, opened 2025-12-28
Input Validation Bug?


daneatmastra:
This issue was created from Discord post 1454662146003832993:

[![Open in Discord](https://img.shields.io/badge/Open_in_Discord-5865F2?style=for-the-badge&logo=discord&logoColor=white)](https://discord.com/channels/1309558646228779139/1454662146003832993)

Hey guys! Running into what looks like a bug with tool input validation when using nested objects with .partial(). 

The issue: When I create a tool with a Zod schema that has a nested object field marked as .partial(), the schema validates correctly when I test it directly, but Mastra's internal validation is still failing with "expected string, received undefined" errors for all the partial fields. 

Versions:
@mastra/core: 0.24.8
zod: 4.1.13

Minimal repro:
```
import { createTool } from '@mastra/core';
import { z } from 'zod';

// Schema with nested partial object
const inputSchema = z.object({
  eventId: z.string(),
  request: z.object({
    City: z.string(),
    Name: z.string(),
    Slug: z.string(),
  }).partial().passthrough(),
  eventImageFile: z.any().optional()
});

// This validates successfully:
const testData = {
  eventId: '123',
  request: { Name: 'Test' }, // Only providing one field
  eventImageFile: null
};
console.log(inputSchema.safeParse(testData).success); // :white_check_mark: true

// But when passed to Mastra:
const tool = createTool({
  id: 'updateEvent',
  description: 'Update event',
  inputSchema,
  execute: async (input) => {
    return { success: true };
  }
});

// LLM calls with: { eventId: '123', request: { Name: 'Test' }, eventImageFile: null }
// :x: Mastra validation fails with:
// Tool input validation failed for 'updateEvent'
// errors: {
//   request: {
//     City: { _errors: ['Invalid input: expected string, received undefined'] },
//     Slug: { _errors: ['Invalid input: expected string, received undefined'] },
//   }
// }
```

What I've confirmed:
- The nested request.shape.City is correctly a ZodOptional after .partial()
- Direct validation with inputSchema.safeParse() succeeds
- But... Mastra's internal validation (in builder.ts createExecute) fails

daneatmastra:
Thank you for reporting this issue! We have labeled it for trio-tb and we will look into it as soon as possible.

If you're experiencing an error, please provide a [minimal reproducible example](https://github.com/mastra-ai/mastra/blob/main/CONTRIBUTING.md#minimal-reproduction) whenever possible to help us resolve it quickly.

[...]

PR #13481, opened 2026-02-25
fix: fallback to original schema when compat layer breaks optional fields (#13480)

echoVic:
## Summary

Fixes #13480

When OpenAI compat layers process tool schemas, `.optional()` is converted to `.nullable().transform()`. This removes the optional wrapper, making the field **required** (but nullable) in the processed schema. If the LLM omits optional fields entirely (key not present in input), all existing validation steps fail because the processed schema expects the key to exist.

## Solution

Add a Step 6 fallback to `validateToolInput`: after all other attempts fail, try validating against the original (pre-compat) schema which still has `.optional()` and accepts missing keys.

### Changes

- **`validation.ts`**: Add optional `originalSchema` parameter to `validateToolInput`. After Steps 1-5 fail, attempt validation with the original schema as a last resort.
- **`builder.ts`**: Pass `originalSchema` through `createExecute` to `validateToolInput`.
- **`validation.test.ts`**: 6 new tests covering:
  - Single optional field omitted → fallback succeeds
  - Multiple optional fields omitted → fallback succeeds
  - Valid input against processed schema → no fallback needed
  - Invalid input against both schemas → still fails correctly
  - Nested optional fields → fallback succeeds
  - No originalSchema provided → backward compatible (still fails)

## Validation pipeline (updated)

1. `normalizeNullishInput` — top-level null/undefined → {} or []
2. `convertUndefinedToNull` — undefined → null for compat layers (#11457)
3. First validation with null preserved
4. Retry with stringified JSON coercion (#12757)
5. Retry with null values stripped (#12362)
6. **NEW: Fallback to original schema (#13480)**

## Summary by CodeRabbit

* **New Features**
  * Improved tool input validation: when compatibility transforms make optional fields behave oddly, the system now falls back to the original input schema during execution so valid inputs are accepted.

* **Tests**
  * Added comprehensive test coverage for schema-fallback behavior, nested optionals, multiple omissions, null-stripping edge cases, and backward-compatibility scenarios.

* **Chores**
  * Added release metadata for publishing the fix.

### DIFF
--- a/.changeset/original-schema-fallback.md
+++ b/.changeset/original-schema-fallback.md
@@ -1,0 +1,5 @@
+---
+"@mastra/core": patch
+---
+
+fix: fallback to original schema when compat layer breaks optional fields (#13480)
--- a/packages/core/src/tools/tool-builder/builder.ts
+++ b/packages/core/src/tools/tool-builder/builder.ts
@@ -245,6 +245,7 @@ export class CoreToolBuilder extends MastraBase {
     options: ToolOptions,
     logType?: 'tool' | 'toolset' | 'client-tool',
     processedSchema?: z.ZodTypeAny,
+    originalSchema?: z.ZodTypeAny,
   ) {
     // don't add memory, mastra, or tracing context to logging (tracingContext may contain sensitive observability credentials)
     const {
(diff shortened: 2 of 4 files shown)


echoVic (commented):
The null-stripping + original schema fallback edge case raised by CodeRabbit is already addressed in commit 17ebf7d (Step 6b). Test case `should fallback to original schema with null-stripped input (Step 6b)` covers this scenario.

PR #13492, opened 2026-02-25
fix(react): send dev playground header for localhost servers

### LINKED ISSUE #13488: Studio: x-mastra-dev-playground header not sent when using mastra studio with separate server
## Problem

When running `mastra studio` pointed at a separate local server (e.g. Hono on port 4000), the `x-mastra-dev-playground: true` header is never sent with API requests. This means the auth middleware's dev bypass (`MASTRA_DEV=true`) doesn't work in this scenario.

The root cause is in `client-sdks/react/src/mastra-client-context.tsx`:

```typescript
const createMastraClient = (baseUrl?: string, mastraClientHeaders: Record<string, string> = {}, apiPrefix?: string) => {
  return new MastraClient({
    baseUrl: baseUrl || '',
    headers: !baseUrl ? { ...mastraClientHeaders, 'x-mastra-dev-playground': 'true' } : mastraClientHeaders,
    apiPrefix,
  });
};
```

The header is only added when `baseUrl` is falsy (i.e. Studio and API on the same origin, the `mastra dev` case). When `baseUrl` is set — which is always the case with `mastra studio --server-port 4000` — the header is skipped.

## Expected behavior

When `mastra studio` points at a localhost server, the `x-mastra-dev-playground` header should be sent so the server-side auth bypass works with `MASTRA_DEV=true`.

## Steps to reproduce

1. Set up a Hono server with Mastra mounted at `/mastra` with auth configured
2. Run the Hono server with `MASTRA_DEV=true` on port 4000
3. Run `mastra studio --server-port 4000 --server-api-prefix /mastra`
4. Open Studio — all API calls fail with 401 because the dev playground header is missing

## Suggested fix

Update the condition in `createMastraClient` to also send the header when `baseUrl` points to localhost, e.g.:

```typescript
const isLocal = !baseUrl || baseUrl.includes('localhost') || baseUrl.includes('127.0.0.1');
headers: isLocal ? { ...mastraClientHeaders, 'x-mastra-dev-playground': 'true' } : mastraClientHeaders,
```

Related: #12371

graysonhicks:
## Description

Fixed the dev playground header not being sent when using `mastra studio --server-port`. The `x-mastra-dev-playground` header is now correctly included for localhost and 127.0.0.1 URLs, enabling the `MASTRA_DEV=true` auth bypass to work with separate local servers.

## Related Issue(s)

Fixes #13488

## Type of Change

- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [ ] Test update

## Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Dev-auth now correctly applies local development auth for requests to localhost, 127.0.0.1, IPv6 loopback, and hostnames ending with .local or .localhost; local behavior also applies when no base URL is provided.

* **Tests**
  * Added comprehensive tests validating local URL detection for empty/undefined, IPv4/IPv6 loopback, .local/.localhost, remote hosts, and edge-case hostnames.

* **Chores**
  * Added a changelog entry recording the patch.

### DIFF
--- a/client-sdks/react/src/mastra-client-context.tsx
+++ b/client-sdks/react/src/mastra-client-context.tsx
@@ -21,11 +21,22 @@ export const MastraClientProvider = ({ children, baseUrl, headers, apiPrefix }: 
 
 export const useMastraClient = () => useContext(MastraClientContext);
 
+export const isLocalUrl = (url?: string): boolean => {
+  if (!url) return true;
+  try {
+    const { hostname } = new URL(url);
+    return hostname === 'localhost' || hostname === '127.0.0.1';
+  } catch {
+    return false;
+  }
+};
+
 const createMastraClient = (baseUrl?: string, mastraClientHeaders: Record<string, string> = {}, apiPrefix?: string) => {
   return new MastraClient({
     baseUrl: baseUrl || '',
-    // only add the header if the baseUrl is not provided i.e it's a local dev environment
-    headers: !baseUrl ? { ...mastraClientHeaders, 'x-mastra-dev-playground': 'true' } : mastraClientHeaders,
+    headers: isLocalUrl(baseUrl)
+      ? { ...mastraClientHeaders, 'x-mastra-dev-playground': 'true' }
+      : mastraClientHeaders,
     apiPrefix,
   });
 };
--- a/client-sdks/react/src/mastra-client-context.tsx
+++ b/client-sdks/react/src/mastra-client-context.tsx
@@ -25,7 +25,14 @@ export const isLocalUrl = (url?: string): boolean => {
   if (!url) return true;
   try {
     const { hostname } = new URL(url);
-    return hostname === 'localhost' || hostname === '127.0.0.1' || hostname === '::1' || hostname === '[::1]';
+    return (
+      hostname === 'localhost' ||
+      hostname === '127.0.0.1' ||
+      hostname === '::1' ||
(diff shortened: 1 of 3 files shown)


client-sdks/react/src/mastra-client-context.tsx
```diff
@@ -21,11 +21,20 @@ export const MastraClientProvider = ({ children, baseUrl, headers, apiPrefix }:
 
 export const useMastraClient = () => useContext(MastraClientContext);
 
+export const isLocalUrl = (url?: string): boolean => {
```
wardpeet, line 32: i wonder if this is better, actually resolving to localhost so mastra.local also works 🤷 
```
import { lookup } from 'node:dns/promises';
import { isLoopback } from 'node:net';

async function isLocalhostResolved(hostname) {
  // Quick check first
  if (isLocalhost(hostname)) return true;
  
  try {
    const { address } = await lookup(hostname);
    return net.isIP(address) && isLoopbackAddress(address);
  } catch {
    return false;
  }
}
```

graysonhicks: Hm, this is client context, don't think those node packages are available.

graysonhicks: I can still update to handle mastra.local I think

taofeeq-deru (approved):

PR #13517, opened 2026-02-26
feat(core): add SDK setters for platform config injection

rphansen91:
## Summary
- Adds `setObservability()` and `setServer()` public methods to the `Mastra` class for post-construction configuration
- Adds `registerObservabilityInstance()` which safely extends existing observability (or bootstraps from NoOp) without replacing user config
- These methods enable platform tooling (e.g. the studio runner) to inject defaults like `CloudExporter` and `MastraAuthWorkos` at deploy time without requiring users to configure them manually

## Test plan
- [ ] `setObservability()` replaces NoOp with a real instance, shuts down existing non-NoOp before replacing
- [ ] `setServer()` assigns server config
- [ ] `registerObservabilityInstance()` bootstraps from NoOp when no observability exists
- [ ] `registerObservabilityInstance()` extends existing observability via `registerInstance` when user already has one
- [ ] Publish snapshot and verify wrapper integration in studio runner

## Summary by CodeRabbit

* **New Features**
  * Added dynamic configuration for observability settings
  * Added dynamic configuration for server settings
  * Support for registering and managing multiple observability instances

### DIFF
--- a/packages/core/src/mastra/index.ts
+++ b/packages/core/src/mastra/index.ts
@@ -471,6 +471,57 @@ export class Mastra<
   }
 
   /**
+   * Sets the observability instance for this Mastra instance, replacing any existing one.
+   *
+   * If a non-noop observability instance already exists, it will be shut down before
+   * the new one is set.
+   *
+   * @param observability - An ObservabilityEntrypoint instance with a `getDefaultInstance` method
+   * @throws {MastraError} When the provided observability instance is invalid
+   *
+   * @example
+   * ```typescript
+   * import { Observability, DefaultExporter } from '@mastra/observability';
+   *
+   * const mastra = new Mastra();
+   * mastra.setObservability(new Observability({
+   *   configs: { default: { serviceName: 'mastra', exporters: [new DefaultExporter()] } },
+   * }));
+   * ```
+   */
+  public setObservability(observability: ObservabilityEntrypoint): void {
+    if (typeof observability.getDefaultInstance !== 'function') {
+      throw new MastraError({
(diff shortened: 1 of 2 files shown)


packages/core/src/mastra/index.ts
```diff
@@ -470,6 +470,82 @@ export class Mastra<
...
+      });
+    }
+    // Shutdown existing observability if it's not the no-op
+    if (!(this.#observability instanceof NoOpObservability)) {
+      this.#observability.shutdown().catch(() => {});
+    }
+    this.#observability = observability;
+    this.#observability.setLogger({ logger: this.#logger });
+    this.#observability.setMastraContext({ mastra: this });
+  }
```
epinzur, line 508: we also might want the ability to dyamically add/remove exporters... 

Some of my current changes will make this easier... exporters are moving to a subscription interface, where they can be added and subscribe to events, or un-subscribe from events.

epinzur:
maybe we hold off on this until my changes to observability merge?  I'd rather get the new system in place and then figure out how to do this in that world... instead of having to do this twice.

unless this is currently blocking your work

packages/core/src/mastra/index.ts
epinzur: from the description:

>  registerObservabilityInstance() extends existing observability via registerInstance when user already has one

this really isn't helpful,  in this case, observability will still flow to only the first instance unless you add a `configSelector` to choose which instance to export to.

see: https://mastra.ai/docs/observability/tracing/overview#multi-config-setup

epinzur (changes requested):

PR #13534, opened 2026-02-26
fix(convex): use by_thread/by_resource indexes in queryTable for messages and threads

### LINKED ISSUE #13533: convex: queryTable on mastra_messages ignores by_thread index, causes 16MB read limit
## Bug

`handleTypedOperation`'s `queryTable` case does a full table scan on typed tables when no `indexHint` matches:

```ts
// current behavior for mastra_messages:
docs = await ctx.db.query(convexTable).take(maxDocs) // reads ALL messages across ALL threads
// then filters in-memory
if (request.filters && request.filters.length > 0) {
  docs = docs.filter(...)
}
```

`mastra_messages` has `by_thread` and `by_thread_created` indexes defined in the schema, but `queryTable` never uses them. When Memory queries messages for a specific thread, it reads every message across every thread first, then filters in-memory.

With enough messages (especially from agents that use tools returning large payloads like transcripts), this exceeds Convex's 16MB function execution read limit:

```
Uncaught Error: Too many bytes read in a single function execution (limit: 16777216 bytes)
    at async handleTypedOperation (storage.ts:171:10)
```

This is the same class of bug that was fixed in v1.0.3 for workflow snapshots (#12849), where `IndexHint` was added for `by_workflow` / `by_workflow_run`. The fix just wasn't extended to messages or threads.

## Suggested fix

Use existing indexes when matching filters are present in `queryTable`:

```ts
case "queryTable": {
  const maxDocs = request.limit ? Math.min(request.limit * 2, 1e4) : 1e4;
  let docs;

  // Use by_thread index for messages
  if (convexTable === "mastra_messages") {
    const threadFilter = request.filters?.find(f => f.field === "thread_id");
    if (threadFilter) {
      docs = await ctx.db.query(convexTable)
        .withIndex("by_thread", q => q.eq("thread_id", threadFilter.value))
        .take(maxDocs);
      // apply remaining filters (excluding thread_id) in-memory
      // ...
      return { ok: true, result: docs };
    }
  }

  // Use by_resource index for threads
  if (convexTable === "mastra_threads") {
    const resourceFilter = request.filters?.find(f => f.field === "resourceId");
    if (resourceFilter) {
      docs = await ctx.db.query(convexTable)
        .withIndex("by_resource", q => q.eq("resourceId", resourceFilter.value))
        .take(maxDocs);
      // ...
      return { ok: true, result: docs };
    }
  }

  // existing indexHint / fallback logic...
}
```

The same pattern used for `by_workflow` / `by_workflow_run` in v1.0.3 — just extended to the other typed tables that already have indexes defined.

## Environment

- `@mastra/convex`: 1.0.4 (also verified identical in 1.0.3)
- Convex function execution read limit: 16MB
- Trigger: agent with Memory (working memory, thread scope) accumulating messages over many conversations

harjot1singh:
## Summary

- `handleTypedOperation`'s `queryTable` performs a full table scan on `mastra_messages` and `mastra_threads` because no `indexHint` matches — even though both tables have `by_thread`, `by_thread_created`, and `by_resource` indexes defined in the schema
- With enough messages (agents using tools that return large payloads), this exceeds Convex's 16MB function execution read limit
- This is the same class of bug fixed in v1.0.3 for workflow snapshots (#12849, `by_workflow` / `by_workflow_run`), extended to messages and threads

## Changes

When `queryTable` is called on `mastra_messages` or `mastra_threads`, check for `thread_id` / `resourceId` in the filters array and use the corresponding index. Filters consumed by the index are removed from the in-memory filter pass. Falls back to existing behavior (full scan or `indexHint`) when no matching filter is present.

## Test plan

- [ ] Verify `queryTable` on `mastra_messages` with `thread_id` filter uses `by_thread` index
- [ ] Verify `queryTable` on `mastra_messages` with `resourceId` filter uses `by_resource` index
- [ ] Verify `queryTable` on `mastra_threads` with `resourceId` filter uses `by_resource` index
- [ ] Verify `queryTable` without matching filters still falls back to full scan
- [ ] Verify workflow snapshot `indexHint` path is unaffected

Fixes #13533

## Summary by CodeRabbit

* **Refactor**
  * Optimized query paths to utilize targeted index-based lookups instead of full table scans, improving data retrieval performance and reducing resource consumption across the system.


NikAiyer (commented):
Looks good! Can you add some tests for this as well as a changeset? Then we can merge this in.

stores/convex/src/server/storage.ts
```diff
@@ -148,9 +148,44 @@ async function handleTypedOperation(
...
+        if (resourceFilter) {
+          docs = await ctx.db
+            .query(convexTable)
+            .withIndex('by_resource', (q: any) => q.eq('resourceId', resourceFilter.value))
+            .take(maxDocs);
+          remainingFilters = remainingFilters.filter(f => f.field !== 'resourceId');
+        } else {
+          docs = await ctx.db.query(convexTable).take(maxDocs);
+        }
+      } else if (request.indexHint) {
```
NikAiyer, line 187: I believe this check is already handled in a more generalized way down below:
https://github.com/harjot1singh/mastra/blob/e7b28a1843d4c1dd9e6eda2f95500c5ea233381a/stores/convex/src/server/storage.ts#L203

NikAiyer (commented):
I believe this is actually already fixed in another PR: https://github.com/mastra-ai/mastra/pull/13630

I think we can go ahead and close this, but if you see any issues with the current approach and want to modify something, feel free to reopen.

PR #13563, opened 2026-02-26
fix: make Codex models work with OM and mastracode streams

TylerBarnes:
This PR makes Codex models work reliably with observational memory in both @mastra/memory and mastracode. OM observer/reflector calls now use stream output instead of generate so Codex-compatible requests work, OM only defaults maxOutputTokens when using the built-in default model, Codex model reasoning is clamped so gpt-5.3 variants never run with reasoning "off", and mastracode now aborts the stream when OM observation/buffering fails.

Before:
```ts
const result = await agent.generate(prompt, options)
maxOutputTokens: 100_000 // defaulted even on custom models
reasoning: { effort: "off" } // could be used for gpt-5.3 codex routing
// OM failure events were surfaced but stream could continue
```

After:
```ts
const output = await agent.stream(prompt, options)
const result = await output.getFullOutput()
maxOutputTokens: 100_000 // only when default OM model selection is used
reasoning: { effort: "low" | "medium" | "high" } // gpt-5.3 codex floor
// OM failure events abort the mastracode stream
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Observational memory failures now emit appropriate OM error events and abort active streams to prevent continued assistant output.
  * Default maxOutputTokens is no longer auto-applied when a custom observation or reflection model is supplied.

* **New Features**
  * Observational memory runs use streaming-compatible model execution so stream-only providers work seamlessly.

* **Documentation**
  * Clarified modelSettings defaults for observational memory.

* **Tests**
  * Added tests covering Codex model routing, thinking-level enforcement, streaming OM behavior, and failure aborts.

### DIFF
--- a/.changeset/better-wings-type.md
+++ b/.changeset/better-wings-type.md
@@ -2,4 +2,6 @@
 '@mastra/memory': patch
 ---
 
-Adjusted observational memory defaults so maxOutputTokens is only auto-applied when using the built-in default model selection.
+Stop auto-applying `maxOutputTokens: 100_000` to observer and reflector models when users supply their own model.
+
+Previously, the 100k default was always injected regardless of model choice. Now it is only applied when using the built-in default model (`google/gemini-2.5-flash`). If you set a custom `observation.model` or `reflection.model`, no `maxOutputTokens` default is added — pass it explicitly in `modelSettings` if your model needs it.
--- a/.changeset/better-wings-type.md
+++ b/.changeset/better-wings-type.md
@@ -1,0 +1,5 @@
+---
+'@mastra/memory': patch
+---
+
+Adjusted observational memory defaults so maxOutputTokens is only auto-applied when using the built-in default model selection.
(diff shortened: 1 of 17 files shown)


abhiaiyer91 (approved):

TylerBarnes:
@coderabbitai All five review items addressed in 6415b785d7:

1. **Debug logs removed** — all `console.log` statements stripped from `mid-loop-observation.test.ts`
2. **Changeset rewritten** — `better-wings-type.md` now action-oriented, clarifies maxOutputTokens only auto-applied for default model
3. **Changeset rewritten** — `shaggy-zebras-fail.md` uses user-facing language; `@mastra/core` changes already covered by separate `funny-rivers-smile.md` changeset
4. **cycleId guard fixed** — `harness.ts` buffering-failed handler changed from `if (payload && payload.cycleId)` to `if (payload)`
5. **Mock hardened** — `createStreamCapableMockModel` now throws on `doGenerate()` calls to catch accidental generate-path regressions

PR #13564, opened 2026-02-26
fix(mastracode): wire extraTools into tool builder and filter denied tools

### LINKED ISSUE #13550: mastracode: extraTools not merged; per-tool deny still advertises tools
## Affected package

`mastracode` (observed in `0.4.0`)

## Summary

Two behavior mismatches in `createMastraCode()`:

1. `extraTools` is exposed in docs/types, but not merged into the dynamic tool set at runtime.
2. `initialState.permissionRules.tools[toolName] = "deny"` blocks execution, but denied tools remain in advertised tool availability (and in tool guidance text), causing UX/runtime mismatch.

## Reproduction

```ts
await createMastraCode({
  initialState: {
    permissionRules: {
      tools: {
        string_replace_lsp: "deny",
      },
    },
  },
  extraTools: {
    my_tool: someTool,
  },
});
```

Observed:
- `my_tool` does not appear in available tools.
- `string_replace_lsp` still appears as available in guidance/listing.

Expected:
- `my_tool` appears and is callable.
- denied tools are omitted from surfaced tool list/guidance.

## Implementation direction

In `createDynamicTools(...)`:
- merge `extraTools` into the returned `tools` map,
- apply per-tool `deny` filtering before returning.

In prompt/tool guidance:
- conditionally omit guidance entries for tools denied via `permissionRules.tools`.

## Notes

We currently carry a local patch downstream to implement the above. Happy to open a PR with this change set if this direction is accepted.

NikAiyer:
## Description

Two bugs fixed in Mastra Code tool handling:

1. **extraTools not merged** — The `extraTools` parameter in `createMastraCode` was accepted via `MastraCodeConfig` but never passed through to `createDynamicTools`, so extra tools were silently dropped. They are now correctly merged into the tool set (without overwriting built-in tools).

2. **Denied tools still advertised** — Tools with a per-tool `deny` policy in `permissionRules` were still included in the tool set and system prompt guidance, causing the model to attempt using them only to be blocked at execution time. Denied tools are now filtered from both the dynamic tool map and the tool guidance sections, so the model never sees them.

## Related Issue(s)

Fixes #13550

## Type of Change

## Checklist

## Test Plan

- **14 new automated tests** in `mastracode/src/agents/__tests__/extra-tools.test.ts` covering:
  - extraTools merge into tool set
  - Built-in tools take precedence over extraTools with the same name
  - extraTools work without MCP manager
  - Per-tool deny filtering removes tools from tool set
  - Multiple denied tools filtered simultaneously
  - Allow/ask policies preserve tools
  - Denied extraTools also filtered
  - Denied tools omitted from `buildToolGuidance` output
  - All tools included when no denied set is provided
- **Manual testing** confirmed:
  - Bug #1: `test_extra_tool` injected via `extraTools` appears and executes correctly
  - Bug #2: `execute_command` with deny rule is completely invisible to the model — model uses alternative tools and reports it doesn't have `execute_command` available

## Bug Fixes
* Fixed extra tools not merging properly with built-in tools—they are now combined without overwriting existing functionality.
* Fixed tools with deny policies still appearing in guidance—denied tools are now excluded from prompts and execution.

### DIFF
--- a/mastracode/src/agents/prompts/tool-guidance.ts
+++ b/mastracode/src/agents/prompts/tool-guidance.ts
@@ -6,9 +6,12 @@
 
 interface ToolGuidanceOptions {
   hasWebSearch?: boolean;
+  /** Tool names that have been denied — omit their guidance sections. */
+  deniedTools?: Set<string>;
 }
 
 export function buildToolGuidance(modeId: string, options: ToolGuidanceOptions = {}): string {
+  const denied = options.deniedTools ?? new Set<string>();
   const sections: string[] = [];
 
   sections.push(`# Tool Usage Rules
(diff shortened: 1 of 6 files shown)


mastracode/src/agents/prompts/tool-guidance.ts
```diff
@@ -19,86 +22,133 @@ You have access to the following tools. Use the RIGHT tool for the job:`);
...
 - \`old_str\` must be an exact match of existing text in the file.
 - Provide enough surrounding context in \`old_str\` to make it unique.
 - For creating new files, use \`write_file\` instead.
 - Good: Include 2-3 lines of surrounding context to ensure uniqueness.
-- Bad: Using just \`return true;\` — too common, will match multiple places.
+- Bad: Using just \`return true;\` — too common, will match multiple places.`);
+    }
 
+    if (!denied.has('write_file')) {
+      writeTools.push(`
```
TylerBarnes, line 88: we should generally avoid changing the system prompt as much as possible, but perhaps preventing it is for another PR. anytime we change the system prompt it invalidates the prompt cache and makes the next request much more expensive. In the future we should probably always keep tools and tool descriptions stable, and just have the tool reject with an error if it's denied. For making the model know what's allowed and what isn't, we can inject user messages like `<system-reminder>you are now in X mode, the only tools you can call are $list</system-reminder>`

TylerBarnes (approved):

mastracode/src/agents/prompts/tool-guidance.ts
NikAiyer: Yeah I think a lot of this will be fixed by Caleb's PR when that's merged: https://github.com/mastra-ai/mastra/pull/13437

PR #13568, opened 2026-02-27
feat(memory): observer context optimization

TylerBarnes:
Reduces observer token costs by truncating the "Previous Observations" context to a configurable budget and automatically folding in buffered reflections.

### `observation.previousObserverTokens`

Controls how many tokens of previous observations the observer sees. Defaults to `2000`. Set to `0` to send nothing, `false` to explicitly disable (full history).

```ts
observation: {
  previousObserverTokens: 3000,
}
```

When truncation is active:
- Keeps the most recent observations as a raw tail
- Preserves important (🔴) observations from older context
- Shows inline `[N hidden observations]` markers at gaps (skipped when only 1 observation is hidden)
- Injects prior `current-task` and `suggested-response` from thread metadata for continuity
- Automatically replaces reflected observation lines with the pending buffered reflection summary

### Other changes
- Empty/metadata-only messages are now filtered from observer input
- Non-empty `reasoning` parts are included in observer context
- Multi-thread observer skips empty `<thread>` wrappers
- Truncation state passed explicitly (`wasTruncated` boolean) instead of regex detection
- `previousObserverTokens` exposed via `get config()` and `getResolvedConfig()`

Example:
Screenshot 2026-02-27 at 3 17 40 PM


.changeset/thin-knives-admire.md
LekoArts, line 1: That changeset shouldn't be necessary since it's opt-in and shouldn't be breaking for people if they have old configuration

docs/src/content/en/reference/memory/observational-memory.mdx
```diff
@@ -163,6 +163,22 @@ OM performs thresholding with fast local token estimation. Text uses `tokenx`, a
             isOptional: true,
             defaultValue: '1.2 (when bufferTokens is set)',
           },
+          {
+            name: 'observer.previousObservationTokens',
```
LekoArts: You mustn't use dot notation for these values, you have to properly nest them.

Here's an example from some lines above:

```md
{
  name: 'modelSettings',
  type: 'ObservationalMemoryModelSettings',
  description: 'Model settings for the Observer agent.',
  isOptional: true,
  defaultValue: '{ temperature: 0.3, maxOutputTokens: 100_000 }',
  properties: [
    {
      type: 'ObservationalMemoryModelSettings',
      parameters: [
        {
          name: 'temperature',
          type: 'number',
          description: 'Temperature for generation. Lower values produce more consistent output.',
          isOptional: true,
          defaultValue: '0.3',
        },
        {
          name: 'maxOutputTokens',
          type: 'number',
          description: 'Maximum output tokens. Set high to prevent truncation of observations.',
          isOptional: true,
          defaultValue: '100000',
        },
      ],
    },
  ],
},
```

packages/memory/package.json
```diff
@@ -79,7 +79,7 @@
     "zod": "catalog:"
   },
   "peerDependencies": {
-    "@mastra/core": ">=1.4.1-0 <2.0.0-0",
+    "@mastra/core": ">=1.8.0-0 <2.0.0-0",
```
LekoArts: See changeset comment

LekoArts (changes requested):

.changeset/thin-knives-admire.md
TylerBarnes: ah missed this, thanks

packages/memory/src/processors/observational-memory/observational-memory.ts
```diff
@@ -1659,32 +1687,274 @@ export class ObservationalMemory implements Processor<'observational-memory'> {
...
+    // 2. Apply tail truncation to previousObservationTokens (keep most recent observations)
+    if (tokenBudget !== undefined) {
+      if (tokenBudget === 0) {
+        return '';
+      }
+
+      const currentTokens = this.tokenCounter.countObservations(optimized);
+      if (currentTokens > tokenBudget) {
+        optimized = this.truncateObservationsToTokenBudget(optimized, tokenBudget);
+      }
```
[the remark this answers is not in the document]
TylerBarnes: Reese says: Good catch @coderabbitai — you're right that applying `useBufferedReflection` before truncation meant small budgets (especially `0`) would silently eat the reflection content. Fixed: truncation now runs first on the raw observations, and buffered reflection is appended *after* — completely outside the token budget. Added a dedicated test for the `previousObservationTokens: 0` + `useBufferedReflection: true` edge case.

— The reflection shall not perish 🪞

packages/memory/src/processors/observational-memory/observational-memory.ts
```diff
@@ -4997,9 +5340,13 @@ ${formattedMessages}
...
       }
 
-      const existingObservations = freshRecord?.activeObservations ?? record.activeObservations ?? '';
+      const rawExistingObservations = freshRecord?.activeObservations ?? record.activeObservations ?? '';
 
-      // ═════════════════════════════════════════���══════════════════
+      // Apply observer context optimization (truncation, buffered-reflection inclusion, gating)
+      const existingObservations =
+        this.prepareObserverContext(rawExistingObservations, freshRecord ?? record) ?? rawExistingObservations;
+
```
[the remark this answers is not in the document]
TylerBarnes: Reese says: Excellent find @coderabbitai — this was a real data-loss bug. The multi-thread path was using the truncated `existingObservations` (meant only for the prompt) as the base for building `currentObservations` that gets persisted to storage. Fixed: `currentObservations` now starts from `rawExistingObservations`, so the full observation history is always preserved in the DB.

— Your canonical observations are safe with us 🏛️

packages/memory/src/processors/observational-memory/__tests__/observational-memory.test.ts
```diff
@@ -2826,6 +2875,83 @@ Ask about favorite vegetarian dishes
...
+    const userPromptText = Array.isArray(userMessage.content)
+      ? userMessage.content
+          .map((part: any) => (part?.type === 'text' ? part.text : ''))
+          .filter(Boolean)
+          .join('\n')
+      : String(userMessage.content ?? '');
+    expect(userPromptText).toContain('Prior Thread Metadata');
+    expect(userPromptText).toContain('prior current-task: Implement observer context optimization');
+    expect(userPromptText).toContain('prior suggested-response: I will update the observer prompt context next.');
+  });
```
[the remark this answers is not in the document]
TylerBarnes: Reese says: Respectfully disagree on this one @coderabbitai — prior thread metadata injection (`currentTask`, `suggestedResponse`) is a core feature of multi-thread observational memory, not gated behind `previousObservationTokens`. The truncation path controls how much *history* the observer sees, but metadata continuity ("what was this thread working on?") is always useful regardless of whether context optimization is enabled. These tests correctly validate the always-on behavior.

— Prior metadata: gating not required, conviction strong 💪

packages/memory/src/processors/observational-memory/__tests__/observational-memory.test.ts
```diff
@@ -8545,6 +8803,328 @@ describe('threadId validation in thread scope', () => {
...
+      const om = createOM({
+        previousObservationTokens: budget,
+        useBufferedReflection: true,
+      });
+      const record = { bufferedReflection: '- Reflection summary' };
+
+      const result = prepareObserverContext(om, observations, record);
+      expect(result).not.toBe(observations);
+      expect(tc.countObservations(result!)).toBeLessThanOrEqual(budget);
+    });
```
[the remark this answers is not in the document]
TylerBarnes: Reese says: Good suggestion @coderabbitai — this was actually a valid concern because the old code *did* apply buffered reflection before truncation, meaning tight budgets could silently eat it. Fixed: truncation now runs first on raw observations, and buffered reflection is appended afterward (outside the budget). Added explicit `toContain('- Reflection summary')` assertion to the combined test, plus a new edge-case test for `previousObservationTokens: 0` + `useBufferedReflection: true`.

— Reflections: now truncation-proof 🛡️

docs/src/content/en/docs/memory/observational-memory.mdx
```diff
@@ -253,6 +253,55 @@ Async buffering isn't supported with `scope: 'resource'`. It's automatically dis
 
 :::
 
+## Observer Context Optimization
+
+By default, the Observer receives the full observation history as context when processing new messages. The Observer also receives prior `current-task` and `suggested-response` metadata (when available), so it can stay oriented even when observation context is truncated. For long-running conversations where observations grow large, you can opt into context optimization to reduce Observer input costs.
+
+Both settings are opt-in and disabled by default — existing behavior is preserved unless you explicitly enable them.
```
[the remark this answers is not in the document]
TylerBarnes: Reese says: Appreciate the wording suggestion @coderabbitai, but the prior `current-task` and `suggested-response` metadata is actually *always* injected into the observer prompt (not just when truncation is enabled). It helps the observer maintain continuity across observation cycles regardless of whether `previousObservationTokens` is set. The current wording is accurate.

— Metadata: always present, always helpful 📋

.changeset/thin-knives-admire.md
```diff
@@ -0,0 +1,5 @@
+---
+'@mastra/memory': minor
+---
+
+Update peer dependencies to match core package version bump (1.8.0)
```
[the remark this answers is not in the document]
TylerBarnes: Reese says: Moot point @coderabbitai — this changeset has been removed per LekoArts' review feedback (the only changeset needed is the existing `slick-clouds-lead.md` for `@mastra/memory`).

— Changeset: deleted, gone, dust in the wind 🌬️

docs/src/content/en/reference/memory/observational-memory.mdx
TylerBarnes: Fixed — restructured to use proper nested `properties` / `parameters` arrays instead of dot notation. Thanks for the example!

packages/memory/package.json
TylerBarnes: Reverted peer dep back to `>=1.4.1-0 <2.0.0-0` 👍

packages/memory/src/processors/observational-memory/observer-agent.ts
```diff
@@ -746,6 +770,35 @@ export function buildMultiThreadObserverTaskPrompt(existingObservations: string
...
+    .filter(Boolean)
+    .join('\n');
+
+  if (threadMetadataLines) {
+    prompt += `## Prior Thread Metadata\n\n${threadMetadataLines}\n\n`;
+    if (hasTruncatedObservations) {
+      prompt += `Previous observations were truncated for context budget reasons.\n`;
+      prompt += `The main agent still has full memory context outside this observer window.\n`;
+    }
+    prompt += `Use each thread's prior current-task and suggested-response as continuity hints, then update them based on that thread's new messages.\n\n---\n\n`;
```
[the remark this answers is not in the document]
TylerBarnes: Reese says: Done in d2a0254 — `prepareObserverContext` now returns `{ context, wasTruncated }` and the boolean is threaded through `callObserver`/`callMultiThreadObserver` into both prompt builders. The regex is gone.

@coderabbitai

packages/memory/src/processors/observational-memory/observer-agent.ts
```diff
@@ -639,6 +645,11 @@ function formatObserverMessage(
     content = maybeTruncate(msg.content.content, maxLen);
   }
 
+  // Skip messages that produced no visible content (e.g. only data-* or encrypted reasoning parts)
+  if (!content && attachments.length === 0) {
+    return { text: '', attachments };
+  }
```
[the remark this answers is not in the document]
TylerBarnes: Reese says: Fixed in 2e14bc5 — threads with no visible content after filtering are now skipped entirely. No empty `<thread>` wrappers emitted.

@coderabbitai

packages/memory/src/processors/observational-memory/__tests__/observational-memory.test.ts
```diff
@@ -8545,6 +8860,397 @@ describe('threadId validation in thread scope', () => {
...
+
+      const result = prepareObserverContext(om, observations, record);
+      expect(result).toBeDefined();
+      // Old raw lines are not present (were replaced)
+      expect(result!).not.toContain('Old line 1');
+      // Most recent observations preserved (tail)
+      expect(result!).toContain('New line 40');
+      // Total fits within budget
+      expect(tc.countObservations(result!)).toBeLessThanOrEqual(budget);
+    });
```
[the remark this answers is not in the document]
TylerBarnes: Reese says: This comment was based on the old `useBufferedReflection` API which has since been removed. `previousObservationTokens` now implicitly handles buffered reflection when available. The test suite has been updated accordingly — 320 tests passing.

@coderabbitai

mastracode/src/agents/memory.ts
```diff
@@ -49,7 +49,8 @@ export function getDynamicMemory(storage: MastraCompositeStore) {
     const obsThreshold = state?.observationThreshold ?? DEFAULT_OBS_THRESHOLD;
     const refThreshold = state?.reflectionThreshold ?? DEFAULT_REF_THRESHOLD;
 
-    const cacheKey = `${obsThreshold}:${refThreshold}:${omScope}`;
+    const observerPreviousObservationTokens = 1000;
```
CalebBarnes, line 52: Is that not configurable here?

packages/memory/src/processors/observational-memory/observational-memory.ts
```diff
@@ -1659,32 +1681,233 @@ export class ObservationalMemory implements Processor<'observational-memory'> {
...
+    const lines = observations.split('\n');
+    const totalCount = lines.length;
+
+    // tokenx is lightweight (regex-based), so measure each line directly.
+    const lineTokens: number[] = new Array(totalCount);
+    const isImportant: boolean[] = new Array(totalCount);
+    for (let i = 0; i < totalCount; i++) {
+      lineTokens[i] = this.tokenCounter.countString(lines[i]!);
+      isImportant[i] = lines[i]!.includes('🔴');
```
DanielSLew, line 1766: This is a nit, but I feel like the emoji's 🔴 should be consts so it's easier to tell what they mean. Because to me red might mean like not as important, rather than important. It's harder to grok what the semantic meaning of the stop light colours are.

CalebBarnes (approved):

packages/memory/src/processors/observational-memory/observational-memory.ts
```diff
@@ -1659,32 +1681,233 @@ export class ObservationalMemory implements Processor<'observational-memory'> {
...
+        }
+        outputLines.push(lines[keptIndex]!);
+        previousKeptIndex = keptIndex;
+      }
+
+      const trailingHiddenCount = totalCount - previousKeptIndex - 1;
+      if (trailingHiddenCount === 1) {
+        outputLines.push(lines[totalCount - 1]!);
+      } else if (trailingHiddenCount > 1) {
+        outputLines.push(`[${trailingHiddenCount} hidden observations]`);
```
DanielSLew: I saw the image and I thought the `hidden observations` text was a little confusing, not sure if it would confuse LLM's or not. I thought it meant that everything under it was the hidden observation. Maybe like `292 observations removed from context`, that doesn't sound quite right either, not sure if I have a better suggestions, but i found it confusing.

packages/memory/src/processors/observational-memory/types.ts
```diff
@@ -147,6 +147,14 @@ export interface ObservationConfig {
    */
   blockAfter?: number;
 
+  /**
+   * Optional token budget for observer context.
+   * When set, "Previous Observations" is tail-truncated to preserve the most recent entries,
+   * and pending buffered reflections replace the raw observations they summarized.
+   * Set to `0` for full truncation (omit previous observations entirely), or `false` to disable.
+   */
+  previousObserverTokens?: number | false;
```
DanielSLew, line 156: This property name is a also a little confusing, I didn't know what it meant at first. I don't know if I have a better name for it. `includeTokensFromPreviousObservations` is too verbose haha, maybe even like `tokensFromPreviousObservations`

DanielSLew (approved):

mastracode/src/agents/memory.ts
TylerBarnes: there's no mc settings ui wired up for it, but from my testing it doesn't make a difference in output quality so I didn't add one

packages/memory/src/processors/observational-memory/observational-memory.ts
TylerBarnes: good point, maybe `[x observations truncated here]` or something

packages/memory/src/processors/observational-memory/types.ts
TylerBarnes: before I had it as `observation.observer.previousObservationTokens` lmao which is even worse. The thing I'm trying to avoid too is any name that might sound like it can apply to the actor agent which I think your suggestions could be read as

TylerBarnes: but I guess mine can too

mastracode/src/agents/memory.ts
TylerBarnes: Reese says: Tyler's right — no MC settings UI for this one. The default of `2000` should be a good fit for most conversations without needing a knob. If we start hearing feedback that people want to tweak it, we can always wire up a setting later.

Yours in token frugality,
Claude

packages/memory/src/processors/observational-memory/observational-memory.ts
TylerBarnes: Reese says: Good call — just pushed `cc59979` which changes the marker from `[N hidden observations]` to `[N observations truncated here]`. Should be much clearer that it's a gap indicator, not a label for what follows.

Clarity over cleverness,
Claude

packages/memory/src/processors/observational-memory/types.ts
TylerBarnes: Reese says: We've since flattened this from `observation.observer.previousObservationTokens` to `observation.previousObserverTokens`, which reads a bit better. The "observer" qualifier helps distinguish it from any actor-facing budget. Open to further bikeshedding if anyone has a stronger name — naming is hard and this one lives in user-facing config.

Still workshopping,
Claude

packages/memory/src/processors/observational-memory/observational-memory.ts
TylerBarnes: Reese says: Good nit — the 🔴/🟡/🟢 priority system is a pre-existing convention baked into the observer's prompt templates (used ~20 times across `observer-agent.ts`). Extracting them to named constants would make sense as a follow-up, but it's a broader refactor touching the whole observer prompt system rather than just this PR's truncation logic. Filed it as a TODO!

Seeing red (in a good way),
Claude

PR #13569, opened 2026-02-27
feat(memory): clone Observational Memory when forking threads

TylerBarnes:
When `memory.cloneThread()` is called and Observational Memory is enabled, OM records are now cloned along with the thread.

**Thread-scoped OM** gets a full copy with message IDs remapped to the cloned messages. **Resource-scoped OM** is shared when the `resourceId` stays the same, and cloned with remapped thread tags when it changes. Multi-generation history (including reflections) is preserved, and transient processing flags are reset.

```ts
const { thread, clonedMessages, messageIdMap } = await memory.cloneThread({
  sourceThreadId: 'thread-123',
  resourceId: 'new-resource', // optional — triggers OM clone for resource-scoped
});
// OM records are automatically cloned with remapped IDs
```

Also adds a `/clone` slash command to MastraCode and a "Clone thread" option when a thread lock is encountered. The thread selector now supports pressing `c` to clone a thread.

Storage adapters (PG, LibSQL, MongoDB) implement `insertObservationalMemoryRecord` and return `messageIdMap` from `cloneThread`. Harness gets a `cloneThread()` method with thread lock handling.

9 unit tests cover thread-scoped cloning, resource-scoped sharing/cloning, multi-generation history, tag remapping, and transient flag reset. Integration test coverage added to shared reusable-tests.
Screenshot 2026-02-27 at 2 43 01 PM

## Summary by CodeRabbit

* **New Features**
  * Added a /clone command and TUI clone workflow with optional rename; thread selector now prioritizes threads in the current directory and auto-resumes single-directory threads.

* **Bug Fixes / Behavior Changes**
  * Thread lock prompts now include "Switch thread" and cloning options. Memory provided as a dynamic factory is resolved before cloning.

* **New Functionality**
  * Observational Memory (OM) cloning with remapped message IDs; clone results include a message ID map and OM records are persisted.

* **Documentation**
  * Docs updated for harness.cloneThread, Memory.cloneThread (messageIdMap/OM behavior), and Mastra Code /clone.

* **Tests**
  * New unit and integration tests covering thread cloning and OM remapping.


.changeset/cyan-feet-act.md
```diff
@@ -0,0 +1,17 @@
+---
+'@mastra/memory': minor
```
DanielSLew: gotta fix the changesets, memory, core, and MC should probably each be their own and the storage providers can all be in the same one. each of the bold sections you have here is exactly how they can be split up

DanielSLew (approved):

PR #13579, opened 2026-02-27
fix(core): restore stdout pipe inheritance for Windows LocalSandbox

wumingaizhou:
## Summary

On Windows, `LocalSandbox` command execution (`mastra_workspace_execute_command`) can only capture output from **shell built-in commands** (e.g., `echo`, `dir`). All **external executables** (e.g., `python`, `git`, `node`, `npm`, `pnpm`) run successfully (exit code 0) but produce **no stdout output**, making the workspace sandbox effectively unusable on Windows.

### Root Cause

`detached: true` in `child_process.spawn()` sets the `DETACHED_PROCESS` flag on Windows `CreateProcessW`. This flag detaches the child from the parent's console and breaks stdout pipe handle inheritance for sub-processes of `cmd.exe`. While `detached: true` is harmless on Linux/macOS (it just calls `setsid()` for process group management), on Windows it silently discards stdout from any external executable launched by the shell.

### Changes

All changes are in `packages/core/src/workspace/sandbox/local-process-manager.ts`:

- **Make `detached` platform-conditional** (`detached: !isWindows`) — on Windows, `detached: false` restores normal pipe inheritance; on Linux/macOS, behavior is unchanged
- **Add `killProcessTree()` helper** — uses `taskkill /T /F /PID` on Windows for proper process tree cleanup, while preserving the existing `process.kill(-pid)` (process group signal) behavior on Linux/macOS
- **Update `kill()` method and timeout handler** to use the new `killProcessTree()` helper

### Before / After (Windows)

| Command | Before | After |
|---|---|---|
| `echo hello` | `hello` ✅ | `hello` ✅ |
| `git --version` | *(no output)* ❌ | `git version 2.x.x` ✅ |
| `python --version` | *(no output)* ❌ | `Python 3.11.x` ✅ |
| `node --version` | *(no output)* ❌ | `v22.x.x` ✅ |

before:
image
after:
image

### Risk Assessment

- **Linux / macOS regression**: None — `detached: true` + `process.kill(-pid)` path is completely unchanged
- **`taskkill` availability**: Built-in Windows command since Windows XP
- **`execSync` blocking**: Negligible — `taskkill` executes fast; `kill()` is not a hot path
- **Process already exited**: `taskkill` throws on non-existent PID, caught by existing `try/catch` fallback

## Test plan

- [x] Verified on Windows 11 that external executables (`git`, `python`, `node`) now produce stdout output
- [x] Verified shell built-in commands (`echo`, `dir`) continue to work
- [ ] Verify Linux/macOS behavior is unchanged (existing CI should cover this)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved process termination reliability and stability across Windows and POSIX.
  * Resolved Windows-specific stdout/pipe inheritance and cleanup issues to prevent orphaned processes.
  * Safer sandbox process spawning to avoid platform-specific termination failures, reducing crashes and hangs during process cleanup.


wardpeet (approved):

packages/core/src/workspace/sandbox/local-process-manager.ts
```diff
@@ -5,6 +5,9 @@
  * Tracks processes in-memory since there's no server to query.
  */
 
+import { execSync } from 'node:child_process';
+
+import { execa } from 'execa';
```
abhiaiyer91: why was this added?

wumingaizhou: Sorry about that — this was a leftover from a merge conflict.

wardpeet (approved):

PR #13600, opened 2026-02-27
feat(mastracode): support Anthropic API key as fallback auth for model resolution

DanielSLew:
Fixes #13575

`resolveModel()` previously always routed `anthropic/*` models through the Claude Max OAuth provider, ignoring any `ANTHROPIC_API_KEY` that was set.

Now it checks auth in this order:

1. Claude Max OAuth (if logged in via `/login`) — primary path
2. `ANTHROPIC_API_KEY` env var or stored API key credential — fallback
3. Falls through to OAuth provider to prompt login if nothing is configured

This means users with a Claude Max subscription get the OAuth experience by default, while users with direct API keys can still use mastracode without OAuth.

Also exports `promptCacheMiddleware` from the Claude Max provider so the API key path can reuse it.

## Summary by CodeRabbit

* **New Features**
  * Anthropic models now fall back to an ANTHROPIC_API_KEY (env var or stored API key) when not authenticated via Claude Max OAuth; env var takes precedence.

* **Documentation**
  * Authentication docs and README updated to clarify Anthropic credential lookup order and how to switch between OAuth and API-key usage.

* **Tests**
  * Added tests for model credential resolution, provider selection, and fallback behavior.

### DIFF
--- a/.changeset/fiery-roses-beam.md
+++ b/.changeset/fiery-roses-beam.md
@@ -1,5 +1,0 @@
----
-'mastracode': minor
----
-
-Added support for ANTHROPIC_API_KEY as the primary authentication method for Anthropic models. When the environment variable is set, mastracode now uses it directly instead of requiring Claude Max OAuth. OAuth remains as a fallback when no API key is configured.
--- a/.changeset/fiery-roses-beam.md
+++ b/.changeset/fiery-roses-beam.md
@@ -1,0 +1,5 @@
+---
+'mastracode': minor
+---
+
+Added support for ANTHROPIC_API_KEY as the primary authentication method for Anthropic models. When the environment variable is set, mastracode now uses it directly instead of requiring Claude Max OAuth. OAuth remains as a fallback when no API key is configured.
--- a/docs/src/mastra-code/configuration.mdx
+++ b/docs/src/mastra-code/configuration.mdx
@@ -27,15 +27,13 @@ export CEREBRAS_API_KEY=...
 
 Mastra Code uses Mastra's [model router](https://mastra.ai/models), which routes requests to the correct provider based on the model ID and available API keys. No OAuth login is required when using API keys.
 
-For Anthropic models, the API key is the **primary** authentication method. If `ANTHROPIC_API_KEY` is set, Mastra Code uses it directly with the Anthropic API. This is the recommended path for most users. Claude Max OAuth is only used as a fallback when no API key is available.
-
 ### OAuth login (optional)
 
 If you have an Anthropic Claude Max or OpenAI ChatGPT Plus subscription, you can authenticate via OAuth to use your subscription instead of API keys. Run `/login` in the TUI to start the flow.
 
 | Provider | OAuth flow | What you get |
 | - | - | - |
-| **Anthropic** | Claude Max subscription | Access to Claude models via your subscription (used only when no API key is set) |
+| **Anthropic** | Claude Max subscription | Access to Claude models via your subscription |
 | **OpenAI** | ChatGPT Plus / Codex | Access to OpenAI models via your subscription |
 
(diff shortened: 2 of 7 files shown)


.changeset/fiery-roses-beam.md
```diff
@@ -0,0 +1,5 @@
+---
+'mastracode': minor
+---
+
+Added support for ANTHROPIC_API_KEY as the primary authentication method for Anthropic models. When the environment variable is set, mastracode now uses it directly instead of requiring Claude Max OAuth. OAuth remains as a fallback when no API key is configured.
```
TylerBarnes: loooks like this is a duplicate changeset

docs/src/mastra-code/configuration.mdx
```diff
@@ -27,17 +27,29 @@ export CEREBRAS_API_KEY=...
...
+For Anthropic models, the API key is the **primary** authentication method. If `ANTHROPIC_API_KEY` is set, Mastra Code uses it directly with the Anthropic API. This is the recommended path for most users. Claude Max OAuth is only used as a fallback when no API key is available.
+
 ### OAuth login (optional)
 
 If you have an Anthropic Claude Max or OpenAI ChatGPT Plus subscription, you can authenticate via OAuth to use your subscription instead of API keys. Run `/login` in the TUI to start the flow.
 
 | Provider | OAuth flow | What you get |
 | - | - | - |
-| **Anthropic** | Claude Max subscription | Access to Claude models via your subscription |
+| **Anthropic** | Claude Max subscription | Access to Claude models via your subscription (used only when no API key is set) |
```
TylerBarnes: this should be opposite, only use api key when not oauthed

docs/src/mastra-code/configuration.mdx
```diff
@@ -27,17 +27,29 @@ export CEREBRAS_API_KEY=...
...
 
 OAuth credentials are stored in `auth.json` alongside the database in the app data directory. During onboarding, you can skip the OAuth step if you already have API keys configured.
 
+### Authentication priority for Anthropic
+
+When resolving Anthropic models, Mastra Code checks for credentials in this order:
+
+1. `ANTHROPIC_API_KEY` environment variable
+1. Stored API key credential (set via the TUI)
+1. Claude Max OAuth (fallback)
```
TylerBarnes: same here

abhiaiyer91 (approved):

docs/src/mastra-code/configuration.mdx
TylerBarnes: the way to use api key should be to logout of oauth

TylerBarnes: cause for ex anyone who already had a key set but was also oauthed will now get a surprise bill thanks to your change 😅

TylerBarnes (changes requested):

TylerBarnes (changes requested):
need to swap the loading order, oauth always takes precedence

docs/src/mastra-code/configuration.mdx
DanielSLew: Claude says: Good call — the code was already updated to do OAuth-first in a follow-up (the commit you reviewed was the intermediate state). Just pushed a fix for the docs too, they were still describing the old API-key-first priority. Now everything is aligned: OAuth first → API key fallback → OAuth login prompt.

Claude

.changeset/fiery-roses-beam.md
DanielSLew: Claude says: Fixed — removed the duplicate changeset in an earlier commit.

Claude

docs/src/mastra-code/configuration.mdx
DanielSLew: Claude says: Agreed — the code already does OAuth-first (was updated before this review landed). The docs were lagging behind but just pushed a fix for those too.

Claude

TylerBarnes (approved):

PR #13602, opened 2026-02-27
fix(core): improve structured output processor error handling and observability

roaminro:
## Summary

Improves error handling and observability in structured output processor mode (`structuredOutput.model`):

- **Forward `errorStrategy`/`fallbackValue` to inner structuring agent** — the inner agent's `createObjectStreamTransformer` was always defaulting to strict mode because these options were never passed through, causing unnecessary error chunk propagation and "Error in agent stream" logging
- **Add TripWire guard in catch block** — prevents double error handling when strict mode's `abort()` throws a TripWire that gets caught and re-processed
- **Add `maxRetries` option to `StructuredOutputOptionsBase`** — lets users control the inner agent's p-retry count (defaults to 2 retries unchanged)
- **Add `onFailedAttempt` logging to p-retry** — LLM call retries are no longer silent; users will now see `[Mastra] LLM call to <modelId> failed (attempt N/M): <message>` on retry

## Context

A user reported that `result.object` is always `undefined` when using processor mode with a custom OpenAI-compatible provider, with the inner agent retrying silently. The retry logging added here will give us (and the user) visibility into what's actually failing. This is a draft for the user to test — the retry logs should reveal the root cause.

## Test plan

- [x] Unit tests for errorStrategy/fallbackValue forwarding
- [x] Unit test for TripWire guard (no double error logging)
- [x] Unit tests for maxRetries forwarding and omission
- [x] Integration repro tests with MockLanguageModelV2 (fallback, strict, warn, throwing model)
- [x] Verified "no error in agent stream" test fails without the fix, passes with it
- [x] All existing structured-output and output-format-handlers tests pass

### DIFF
--- a/packages/core/src/agent/types.ts
+++ b/packages/core/src/agent/types.ts
@@ -85,6 +85,13 @@ type StructuredOutputOptionsBase<OUTPUT = {}> = {
   logger?: IMastraLogger;
 
   /**
+   * Maximum number of retries for the internal structuring agent's LLM calls.
+   * Controls the p-retry count when the inner agent encounters transient errors.
+   * If not set, the default retry behavior (2 retries) is used.
+   */
+  maxRetries?: number;
+
+  /**
    * Provider-specific options passed to the internal structuring agent.
    * Use this to control model behavior like reasoning effort for thinking models.
    *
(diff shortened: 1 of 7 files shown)


LekoArts:

PR #13603, opened 2026-02-27
feat(mastracode): auto-update prompt on session start

DanielSLew:
Prompts the user to update when a newer version of mastracode is available on npm. On accept, runs the install command and exits so the new version takes over. On decline, the dismissed version is saved so the prompt doesn't repeat — instead a passive one-liner shows the manual update command.

Detects the package manager used for global installation (npm, pnpm, yarn, bun) so the update/instructions match how the user installed it. Detection uses `npm_config_user_agent`, `NODE_PATH` (set by pnpm global stubs), resolved script path, and a `pnpm list -g` fallback.

**Files changed:**
- `mastracode/src/utils/update-check.ts` (new) — version fetching, semver comparison, PM detection, update runner
- `mastracode/src/main.ts` — reads version from `package.json` instead of hardcoding
- `mastracode/src/onboarding/settings.ts` — adds `updateDismissedVersion` to persisted settings
- `mastracode/src/tui/mastra-tui.ts` — `checkForUpdate()` and `showUpdatePrompt()` integrated into `init()`

## Summary by CodeRabbit

* **New Features**
  * Automatic update prompt at startup and passive periodic checks during long sessions when a newer release exists.
  * Inline Yes/No prompt to run the update; if declined or canceled, shows a one-line manual install command.
  * Detects and suggests the appropriate command for npm, pnpm, yarn, or bun.
  * Remembers dismissed versions to suppress repeat prompts; successful updates advise restart.


mastracode/src/tui/mastra-tui.ts
```diff
@@ -299,6 +301,9 @@ export class MastraTUI {
     } else if (this.shouldShowOnboarding()) {
       await this.showOnboarding();
     }
+
+    // Check for updates (after onboarding so it doesn't interfere)
+    await this.checkForUpdate();
```
TylerBarnes, line 317: should probably also set an interval to recheck every 30-60 mins, it's common to leave a coding agent tui running for days/weeks

DanielSLew: updated!

TylerBarnes (approved):

PR #13625, opened 2026-02-28
feat: add deleteThread method to Harness class

graysonhicks:
## Summary
- Add `deleteThread({ threadId })` method to the `Harness` class that deletes a thread and all its messages from storage
- When deleting the current thread, releases the thread lock and clears active thread/token state
- Add `thread_deleted` event type to `HarnessEvent` union with display state reset handling
- Expose all thread management operations (`create`, `switch`, `list`, `rename`, `delete`) via `harness.memory` getter
- Fix `createThread` lock/state consistency: `saveThread` now runs **after** the lock swap, wrapped in try/catch with proper rollback — releases the new lock, re-acquires the old lock (best-effort), and only restores `currentThreadId` to the previous value if re-acquire succeeds (otherwise sets it to `null`)
- Fix `deleteThread` lock release resilience: `threadLock.release()` is wrapped in try/catch so that state cleanup (`currentThreadId = null`, `tokenUsage` reset) and `thread_deleted` event emission always occur regardless of lock release failure
- Fix `switchThread` lock ordering: lock acquire/release now happens before the storage lookup, so lock operations are adjacent with no intermediate awaits

## Test plan
- [x] 6 new tests covering: storage deletion, lock release on current thread, clearing `currentThreadId`, no lock release on non-current thread, error on nonexistent thread, event emission
- [x] All 128 harness tests pass
- [x] TypeScript compiles cleanly


graysonhicks:
@abhiaiyer91 redid this so you can do `harness.memory.[method]`.  also fixed some broken tests related to thread locking as well

abhiaiyer91 (approved):

PR #13634, opened 2026-03-01
feat(core): show first user message as thread title in chat history

gourine:
## Summary

The chat history sidebar currently shows only a timestamp for each conversation (e.g. "Feb 23 at 5:18:36 PM"), making it impossible to identify what a chat was about without clicking into it.

This PR makes thread history scannable by automatically using the user's first message as the conversation title — no LLM call required.

**How it works:**
- When `generateTitle` is **not** configured (the default), the agent now saves the user's first message (truncated to 100 chars) as the thread title after the first turn
- When `generateTitle` **is** configured (LLM-based), the existing behaviour is unchanged — the LLM title takes priority
- The `playground-ui` sidebar now renders the title above the timestamp, and styles the timestamp as smaller secondary text when a title is present

**Before:** Only a timestamp shown (`Feb 23 at 5:18:36 PM`)
**After:** Title from first message + timestamp as secondary info (`"Help me debug my authentication flow" · Feb 23 at 5:18 PM`)

## Changes

- **`packages/core/src/agent/agent.ts`** — in `executeOnFinish`, added an `else if` branch: when `shouldGenerate` is `false` and the thread is new (`!threadExists`) and has no title, extract the first text part from the most recent user message and save it as the thread title
- **`packages/playground-ui/src/domains/agents/components/chat-threads.tsx`** — when a thread has a title, style the timestamp span as `text-xs text-neutral3 font-normal` to visually demote it to secondary info

## Test plan

- [ ] Start a new chat with an agent that has memory enabled
- [ ] Send a first message (e.g. "Help me plan a trip to Tokyo")
- [ ] Navigate away and back — the sidebar should show "Help me plan a trip to Tokyo" as the thread title
- [ ] Timestamp should appear below the title in smaller muted text
- [ ] Sending subsequent messages should NOT change the thread title (title is only set on first turn)
- [ ] Agents with `generateTitle: true` configured should still use LLM-generated titles (existing behaviour unchanged)
- [ ] Messages with no text content (image/file only) should gracefully fall back to showing only the timestamp

## Summary by CodeRabbit

* **New Features**
  * Thread conversations are now automatically titled using your first message (truncated to 100 characters). LLM-generated titles remain prioritized when configured.

* **UI/Styling**
  * Chat sidebar timestamps now display with secondary styling (smaller, muted appearance) when a conversation title is present, improving visual hierarchy.


packages/core/src/agent/agent.ts
```diff
@@ -4397,6 +4397,24 @@ export class Agent<
               });
             }
           }
+        } else if (!shouldGenerate && !thread.title && !threadExists) {
+          // No LLM title generation configured — use the user's first message as the title directly
+          const userMessage = this.getMostRecentUserMessage(messageList.get.all.ui());
+          if (userMessage) {
+            const textPart = userMessage.parts.find((p): p is TextPart => p.type === 'text');
```
TylerBarnes, line 4404: this should probably be an opt in feature. personally I don't find it very useful, it used to work this way but the sidebar would look like this most of the time:
- hi
- hi
- test
- hello
- hi

But since it's new behaviour we should make it opt in

TylerBarnes: playground-ui can opt in as part of this pr, but we shouldn't opt all users in

TylerBarnes (changes requested):

abhiaiyer91:
closing! we are starting to enforce issues with PRs

PR #13671, opened 2026-03-02
fix(core): fix TypeError in generateTitleFromUserMessage for file parts

### LINKED ISSUE #13670: Bug: generateTitleFromUserMessage fails with TypeError when message contains file parts
### Description

`generateTitleFromUserMessage` in `Agent` throws a `TypeError` when the user message contains file parts (e.g., images). The error is silently caught, resulting in the thread title being set to an empty string `""`.

### Steps to Reproduce

1. Create an agent with memory enabled
2. Send a message containing a file part (e.g., an image) along with text:

```typescript
await agent.generate([
  { type: "file", data: "data:image/png;base64,...", mimeType: "image/png" },
  { type: "text", text: "Describe this image" }
], { threadId: "test-thread", resourceId: "user-1" });
```

3. The thread title will be `""` instead of an auto-generated title.

### Root Cause

In [`packages/core/src/agent/agent.ts` (line 1696)](https://github.com/mastra-ai/mastra/blob/main/packages/core/src/agent/agent.ts#L1696), `generateTitleFromUserMessage` converts the message to UI format using `.ui()`:

```typescript
const normMessage = new MessageList().add(message, 'user').get.all.ui().at(-1);
```

In UI format, file parts have a `url` property (not `data`):

```javascript
// UI format file part (from ui-message.ts):
{ type: 'file', mediaType: 'image/png', url: 'data:image/png;base64,...' }
```

However, at [line 1713](https://github.com/mastra-ai/mastra/blob/main/packages/core/src/agent/agent.ts#L1713), the code references `part.data` which is `undefined` in UI format:

```typescript
} else if (part.type === `file`) {
    partsToGen.push({
      type: 'text',
      text: `User added ${part.mimeType} file: ${part.data.substring(0, 100)}`,
      //                  ^^^^^^^^^^^^^^         ^^^^^^^^^
      //                  also 'mediaType'       undefined in UI format (should be 'url')
    });
}
```

This causes a `TypeError: Cannot read properties of undefined (reading 'substring')`, which is caught by the surrounding try/catch, leaving the title as `""`.

### Expected Behavior

The thread title should be auto-generated even when the user message contains file parts.

### Suggested Fix

Use `part.url` (UI format property) instead of `part.data`, and `part.mediaType` instead of `part.mimeType`:

```typescript
} else if (part.type === `file`) {
    partsToGen.push({
      type: 'text',
      text: `User added ${part.mediaType} file: ${part.url?.substring(0, 100) ?? '(binary)'}`,
    });
}
```

### Environment

- `@mastra/core` version: 1.7.0+ (bug exists on `main`)
- Affects any agent with memory enabled that receives file input

tik-son:
## Summary

Fixes a `TypeError` in `generateTitleFromUserMessage` when user messages contain file parts (e.g., images), which causes thread titles to silently become empty strings.

## Problem

In [`agent.ts` L1696](https://github.com/mastra-ai/mastra/blob/main/packages/core/src/agent/agent.ts#L1696), `generateTitleFromUserMessage` converts the message to UI format using `.ui()`. In UI format, file parts have the shape:

```javascript
{ type: 'file', mediaType: 'image/png', url: 'data:image/png;base64,...' }
```

However, the code at [L1713](https://github.com/mastra-ai/mastra/blob/main/packages/core/src/agent/agent.ts#L1713) references `part.data` (core format) and `part.mimeType` instead of `part.url` and `part.mediaType` (UI format):

```javascript
text: `User added ${part.mimeType} file: ${part.data.substring(0, 100)}`
//                  ^^^^^^^^^^^^^^         ^^^^^^^^^
//                  undefined (UI uses     undefined (UI uses 'url')
//                  'mediaType')
```

This causes `TypeError: Cannot read properties of undefined (reading 'substring')`, which is caught silently, leaving the thread title as `""`.

## Fix

- Access `part.url` (UI format) instead of `part.data`
- Use `part.mediaType ?? part.mimeType` for compatibility with both formats
- Add optional chaining and fallback to prevent future undefined errors

## Related

Fixes #13670

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed thread title generation when messages include file parts. Titles now detect file type and include a shortened preview or show "(binary)" if data/URL isn’t available, preventing empty titles.

* **Tests**
  * Added tests to ensure reliable title generation for messages with mixed file/text parts and UI-format file parts.


roaminro:
Hi @tik-son ! Thanks for the PR!

Looking at the CI tests, it looks like `@mastra/core` isn't building, could you have a look?
Also, could you add a small test? 

Thanks!

tik-son:
Hi @roaminro, thanks for the review!
I've addressed the issues.

PTAL !

NikAiyer (changes requested):
hey @tik-son, can you fix the lint issues? I think there might also be an issue with the tests you added to title-generation file. I think the tests need to wrap the file parts in a proper user message.

packages/core/src/agent/agent.ts
```diff
@@ -1708,9 +1708,14 @@ export class Agent<
...
+        // The .ui() conversion may produce parts with `url`/`mediaType` (AIV5 format)
+        // instead of `data`/`mimeType` (AIV4 FileUIPart type), so handle both.
+        const filePart = part as unknown as Record<string, unknown>;
+        const fileData = typeof filePart.url === 'string' ? filePart.url : typeof filePart.data === 'string' ? filePart.data : undefined;
+        const fileType = typeof filePart.mediaType === 'string' ? filePart.mediaType : typeof filePart.mimeType === 'string' ? filePart.mimeType : 'unknown';
         partsToGen.push({
           type: 'text',
-          text: `User added ${part.mimeType} file: ${part.data.substring(0, 100)}`,
+          text: `User added ${fileType} file: ${fileData ? fileData.slice(0, 100) : '(binary)'}`,
         });
```
NikAiyer: I believe `.get.all.ui()` delegates to AIV4Adapter.toUIMessage, which strips file parts out of parts and moves them into experimental_attachments. The part.type === 'file' branch never fires.

 I think we could do one of the following:
 1. Switch to .get.all.aiV5.ui(). The AIV5 adapter preserves file parts in parts as { type: 'file', url, mediaType }, which matches your handling code, but we would also need to rename source parts to source-url and flattens part.source.url to part.url.
 2. Keep .get.all.ui() and also iterate normMessage.experimental_attachments after the parts loop to pick up the files that AIV4 moved there. Each attachment has { url, contentType }.

NikAiyer (approved):
I updated the tests and changed the title generation logic. Should be good to go!

PR #13703, opened 2026-03-03
fix(react,playground): externalize @mastra/core and stub Node builtins

CalebBarnes:
## Summary
- `nodeExternals()` in the react vite config wasn't catching workspace-linked packages (`workspace:*` resolves to `link:` paths which bypass the plugin)
- This caused `@mastra/core` to be fully inlined into `@mastra/react`'s dist — **902KB → 99KB** after fix
- Any new dep added to core could either bloat the react bundle or break the build (e.g., adding `execa` broke Rollup due to incompatible export conditions in a transitive dep)
- Fix: switched build from Vite/Rollup to tsup with explicit `external: [/^@mastra\/core/]` so core is properly externalized
- Moved `@mastra/core` to `peerDependencies` (`>=1.8.0-0 <2.0.0-0`) since it is externalized and should be provided by consumers, not bundled inline
- Added a Vite plugin to `@internal/playground` that stubs Node.js builtins during production builds — core's dist chunks mix server-only code (voice, workspace tools) with browser-safe code, and the playground needs to resolve those imports without erroring

## Test plan
- [x] `@mastra/react` builds successfully with 99KB dist (down from 902KB)
- [x] `@mastra/core` appears as external imports in dist (not bundled inline)
- [x] `@internal/playground` production build succeeds
- [x] React lint passes
- [x] Full monorepo build passes

## Summary by CodeRabbit

* **Bug Fixes**
  * Reduce distributed bundle size by treating the core package as external (no longer inlined).

* **Chores**
  * Move core package to runtime dependencies for correct resolution.
  * Add browser stubs for Node.js built-ins so the playground builds without errors from server-only imports.
  * Remove automatic UI CSS side-effect import so consumers no longer get global UI styles on import.
  * Update build/test tooling and configs to a new build pipeline.


client-sdks/react/package.json
```diff
@@ -44,6 +44,7 @@
   ],
   "dependencies": {
     "@mastra/client-js": "workspace:*",
+    "@mastra/core": "workspace:*",
```
wardpeet: not sure this is the final fix for this. Shouldn't it be a peerdep?

CalebBarnes: Yeah I think peerdep makes sense here. I only put it there because it was being bundled previously so I think it wasn't actually required to be a dep for consumers.

wardpeet (approved):

client-sdks/react/package.json
CalebBarnes: I'm gonna move it to peerdeps.

client-sdks/react/package.json
```diff
@@ -52,31 +53,34 @@
     "tailwind-merge": "^3.4.1"
   },
   "peerDependencies": {
+    "@mastra/core": ">=1.8.0-0 <2.0.0-0",
```
DanielSLew, line 56: does this mean it's a breaking change?

DanielSLew (approved):

client-sdks/react/package.json
CalebBarnes: I'm not sure. It was never a peer dep before so it was never working correctly. It was just bundling its own version in with it. 

I added 1.8 as minimum because I have no idea what versions added the core APIs this package is using.

PR #13719, opened 2026-03-03
fix: CI test failures in message-list and evals/run

abhiaiyer91:
## Summary

Fixes two test failures surfaced in the CI pipeline for the changeset-release PR (#13523).

### Changes

**1. message-list tests — filename preservation**
- `AIV4Adapter.fromCoreMessage()` now correctly preserves `filename` on file content parts (added in #13574)
- Tests were not updated to expect the new `filename` property
- Also added `filename?: string` to the `MastraMessagePart` type so the assertion is type-safe

**2. evals/run — step scoring guard**
- The guard condition `stepResult.payload && stepResult.output` prevented scorers from running on step results
- Root cause: the workflow engine's `fmtReturnValue` strips `payload` when it matches the previous step's output (optimization to avoid redundant data)
- Fix: removed `payload` from the guard and use `scoringData.input` as fallback input for the scorer

### Test plan
```
pnpm test packages/core/src/agent/message-list/tests/message-list.test.ts  # 89 pass
pnpm test packages/core/src/evals/run/index.test.ts                        # 26 pass
```

### Not addressed (environment/infra issues)
- Memory tests: corrupted fastembed model cache in CI (`ZlibError`)
- Core LLM tests: flaky external API calls (Gemini timeouts/schema errors)
- E2E deployer tests: wrangler config mismatch (`projectName` field)
- Observational-memory tests: broken from introduction (XML formatting, missing setup)
- Agent Builder tests: `createTool` export and `@mastra/mcp` module resolution

## Summary by CodeRabbit

* **Bug Fixes**
  * Relaxed per-step validation in evaluation flows and added input fallback so scorers receive available input when step payload is missing.

* **Tests**
  * Made tests tolerant of an optional filename on file parts to avoid brittle equality checks.
  * Removed a noisy debug log from a test to reduce output.
  * Broadened template-merge validation to accept multiple generated file naming conventions and flexible commit message matching.

* **Chores**
  * Updated an example project's tool import path and simplified a contributor workflow matrix entry.

### DIFF
--- a/packages/core/src/evals/run/index.ts
+++ b/packages/core/src/evals/run/index.ts
@@ -350,12 +350,12 @@ async function runScorers(
       const stepScorerResults: Record<string, any> = {};
       for (const [stepId, stepScorers] of Object.entries(scorers.steps)) {
         const stepResult = targetResult.scoringData.stepResults?.[stepId];
-        if (stepResult?.status === 'success' && stepResult.payload && stepResult.output) {
+        if (stepResult?.status === 'success' && stepResult.output) {
           const stepResults: Record<string, any> = {};
           for (const scorer of stepScorers) {
             try {
               const score = await scorer.run({
-                input: stepResult.payload,
+                input: stepResult.payload ?? targetResult.scoringData.input,
                 output: stepResult.output,
                 groundTruth: item.groundTruth,
                 requestContext: item.requestContext,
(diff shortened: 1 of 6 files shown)


YujohnNattrass (approved):

packages/core/src/evals/run/index.ts
```diff
@@ -350,12 +350,12 @@ async function runScorers(
...
       for (const [stepId, stepScorers] of Object.entries(scorers.steps)) {
         const stepResult = targetResult.scoringData.stepResults?.[stepId];
-        if (stepResult?.status === 'success' && stepResult.payload && stepResult.output) {
+        if (stepResult?.status === 'success' && stepResult.output !== undefined) {
           const stepResults: Record<string, any> = {};
           for (const scorer of stepScorers) {
             try {
               const score = await scorer.run({
-                input: stepResult.payload,
+                input: stepResult.payload !== undefined ? stepResult.payload : targetResult.scoringData.input,
```
wardpeet, line 358: are we sure this is correct?

PR #13730, opened 2026-03-03
fix(core): compact list-files output, lower token caps, suppress empty LSP diagnostics

TylerBarnes:
list-files now returns tab-indented paths instead of tree glyphs, saving tokens:

```
src
	index.ts
	utils
		helpers.ts
package.json

2 directories, 3 files
```

Also adds a `respectGitignore` toggle (default true) so callers can opt out.

Default output token cap reduced from 3k to 2k globally, and list-files uses a 1k cap since directory listings rarely need more.

Empty LSP diagnostics no longer emit "No errors or warnings" — just returns empty string.

## Summary by CodeRabbit

* **New Features**
  * list-files now respects .gitignore by default (toggle to show ignored files) and returns a compact, tab-indented path listing for clearer output.

* **Bug Fixes**
  * LSP diagnostics no longer show a "No errors or warnings" placeholder when there are no issues.

* **Chores**
  * Default output token limits reduced to 2,000 overall and 1,000 for the list-files tool.

### DIFF
--- a/packages/core/src/workspace/tools/tree-formatter.test.ts
+++ b/packages/core/src/workspace/tools/tree-formatter.test.ts
@@ -241,6 +241,22 @@ visible.txt`,
       expect(result.paths).toEqual(['visible.txt']);
       expect(result.fileCount).toBe(1);
     });
+
+    it('should match root-qualified .gitignore patterns when listing a subdirectory', async () => {
+      await fs.writeFile(path.join(tempDir, '.gitignore'), 'apps/web/dist/\n');
+      await fs.mkdir(path.join(tempDir, 'apps', 'web', 'dist'), { recursive: true });
+      await fs.mkdir(path.join(tempDir, 'apps', 'web', 'src'), { recursive: true });
+      await fs.writeFile(path.join(tempDir, 'apps', 'web', 'dist', 'bundle.js'), '');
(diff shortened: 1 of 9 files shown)


packages/core/src/workspace/tools/tree-formatter.test.ts
```diff
@@ -225,6 +225,23 @@ describe('tree-formatter', () => {
...
+      await fs.mkdir(path.join(tempDir, 'node_modules'));
+      await fs.writeFile(path.join(tempDir, 'node_modules', 'index.js'), '');
+      await fs.writeFile(path.join(tempDir, 'debug.log'), '');
+      await fs.writeFile(path.join(tempDir, 'visible.txt'), '');
+
+      const result = await formatAsTree(filesystem, '/');
+
+      expect(result.tree).toBe(
+        `.
+└── visible.txt`,
```
CalebBarnes: Should this not have `└── ` anymore?

TylerBarnes: good call, this made me realize the fix was applied in the wrong place, I'm updating the tree building code instead

TylerBarnes: this is gonna save hundreds of tokens for long lists tho, so that's nice
Screenshot 2026-03-03 at 11 09 19 AM

NikAiyer (approved):

PR #13755, opened 2026-03-04
fix: propagate OTEL context into step.execute() for correct span nesting

Sanrigas:
Without executeWithContext, any auto-instrumented code inside a workflow step (e.g. AI SDK's ai.* spans) sees the outer parent span via context.active() rather than the workflow step span. This causes all AI spans to appear as siblings of the workflow step in the trace rather than nested children.

Wrapping step.execute() in executeWithContext sets the step's OTEL span as the active context during execution, enabling correct parent-child nesting.

## Description

Workflow step spans are created correctly, but `step.execute()` runs outside their OTEL context. Any auto-instrumented code inside a step (e.g. AI SDK's `ai.*` spans) calls `context.active()` and finds the outer parent span rather than the workflow step span — causing all nested spans to appear as siblings instead of children.

Wrapping `step.execute()` in `executeWithContext` sets the step's OTEL span as the active context during execution, enabling correct parent-child nesting.

## Type of Change

- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [ ] Test update

## Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed OTEL context propagation so auto-instrumented operations inside workflow steps are correctly nested under each step's trace span.

* **Improvements**
  * Improved observability and tracing for workflow step execution to provide clearer, more accurate step-level telemetry.


epinzur (commented):
Looks good.  Thanks for finding this and fixing it!

epinzur (approved):

PR #13759, opened 2026-03-04
fix: deliver onIterationComplete feedback when continue is false

### LINKED ISSUE #13756: [BUG] onIterationComplete feedback is silently discarded when continue: false
## Bug Description

When `onIterationComplete` returns `{ continue: false, feedback: "some message" }`, the feedback message is **never added to the conversation**. The model never sees the feedback and never gets a chance to generate a text reply, resulting in an empty `text` output.

## Root Cause

In `packages/core/src/loop/workflows/agentic-loop/index.ts` (around line 107-127):

```typescript
if (iterationResult.continue === false && !hasFinishedSteps) {
    hasFinishedSteps = true;  // ← Set to true FIRST
}

// This condition can NEVER be true when continue === false,
// because hasFinishedSteps was just set to true above
if (iterationResult.feedback && typedInputData.stepResult?.isContinued && !hasFinishedSteps) {
    messageList.add(/* feedback message */);  // ← Never executes
}
```

The `!hasFinishedSteps` condition in the feedback block is always `false` when `continue === false`, because `hasFinishedSteps` was set to `true` in the preceding block.

## Impact

1. `feedback` string is silently discarded — the model never sees it
2. The AI SDK loop stops immediately — the model has no opportunity to generate a text reply
3. `streamResult.text` resolves to `""` (empty string)
4. Users who return `{ continue: false, feedback: "Please summarize your findings" }` expecting the agent to produce a final text response get nothing

## Reproduction

```typescript
const result = await agent.stream("Find entity X", {
  maxSteps: 30,
  onIterationComplete: (ctx) => {
    if (ctx.iteration >= 3) {
      return {
        continue: false,
        feedback: "Stop retrying and summarize what you found."
      };
    }
  }
});

const reader = result.fullStream.getReader();
while (true) {
  const { done } = await reader.read();
  if (done) break;
}

console.log(await result.text); // "" ← empty, feedback was never delivered
```

## Expected Behavior

When `{ continue: false, feedback: "..." }` is returned:
- The feedback should be added to the conversation
- The model should get **one more turn** to produce a text reply incorporating the feedback
- Then the loop should stop

## Suggested Fix

**Option A**: Reorder — add feedback before setting `hasFinishedSteps`:

```typescript
// Add feedback first (if provided)
if (iterationResult.feedback && typedInputData.stepResult?.isContinued) {
    messageList.add(/* feedback */);
}
// Then check continue flag
if (iterationResult.continue === false && !hasFinishedSteps) {
    hasFinishedSteps = true;
}
```

**Option B**: Remove `!hasFinishedSteps` from the feedback condition:

```typescript
if (iterationResult.continue === false && !hasFinishedSteps) {
    hasFinishedSteps = true;
}
// Feedback should always be added when provided, regardless of hasFinishedSteps
if (iterationResult.feedback && typedInputData.stepResult?.isContinued) {
    messageList.add(/* feedback */);
}
```

**Option C** (ideal): When `continue: false` with feedback, give the model one final generation turn before stopping.

## Environment

- `@mastra/core`: 1.8.0
- Bug also present on latest `main` branch

intojhanurag:
## Description

When `onIterationComplete` returned `{ continue: false, feedback: "..." }`, the feedback was silently discarded because `hasFinishedSteps` was
set to `true` before the feedback condition was evaluated, making `!hasFinishedSteps` always false.

This fix restructures the logic to add feedback before checking the continue flag and introduces a `pendingFeedbackStop` mechanism that allows
one more LLM turn for the model to produce a text response incorporating the feedback, then stops the loop. Also fixes a secondary bug where
`stopWhen` evaluation could overwrite the pending stop via direct assignment.

## Related Issue(s)

Fixes #13756

## Type of Change
- [x] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [x] Test update

## Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed feedback handling when `continue: false` is returned; feedback is now appended to the conversation and the model is granted one final turn to incorporate it before stopping, improving iteration termination reliability.

* **Documentation**
  * Clarified iteration-monitoring guidance to explain how feedback + `continue: false` can yield a final response only when the current iteration is still active.

* **Tests**
  * Added integration tests covering the feedback-then-final-turn and force-stop scenarios to ensure correct iteration behavior.


taofeeq-deru (approved):
lgtm

PR #13798, opened 2026-03-04
fix: suppress tool result JSON leak in text-delta stream during multi-step execution

NikAiyer:
## Description

Some models (e.g. gpt-oss-120b via OpenRouter) echo tool-result JSON as `text-delta` content in intermediate agent steps during multi-step execution. This leaked JSON was being sent to clients via SSE `text-delta` events and persisted in the message list, polluting the response stream.

The fix buffers `text-delta` chunks in `processOutputStream` instead of immediately enqueuing them. When a `tool-call` chunk arrives, the buffer is discarded. On `finish`, if the step had no tool calls (a pure text response), the buffer is flushed to the client stream. This ensures leaked JSON from tool-call steps never reaches the client or message list.

**Changes:**

- **`packages/core/src/loop/workflows/agentic-execution/llm-execution-step.ts`**: Buffer text-start/text-delta/text-end chunks; flush on finish for text-only steps, discard for tool-call steps. Clear `textDeltas` state and `immediateText` when tool calls are present. After `hasToolCalls` is set, text chunks are dropped immediately without accumulating in state or buffer.
- **`packages/core/src/stream/base/output.ts`**: Discard accumulated step text when the step has tool calls (mirrors the buffering fix for the `stepResult` layer).
- **`client-sdks/ai-sdk/src/transformers.ts`**: Clear accumulated `text` in `transformAgent` when `tool-call` chunks arrive; skip `text-delta` accumulation when tool calls are already present; reset per-step state (text, toolCalls, toolResults, reasoning, sources, files) on `step-finish`.
- **`client-sdks/ai-sdk/src/__tests__/transformers.test.ts`**: Added comprehensive tests covering text suppression in tool-call steps, normal text emission in text-only steps, multi-step scenarios, step-finish state reset, and edge cases (legitimate thinking text, post-tool-call text, interleaved text with multiple tool calls).

### Known trade-off

**All text in tool-call steps is unconditionally discarded.** Some models legitimately emit text like "Let me search for that" before making tool calls. This text is currently discarded along with leaked JSON. This matches how Vercel AI SDK's `streamText` treats tool-call steps (text field is empty), and most UIs don't render intermediate step text. The edge case is documented with explicit tests.

We investigated whether this matters in practice and concluded it does not:

- The Vercel AI SDK itself treats text and tool-call steps as **mutually exclusive**. In `streamText`, `stepResult.text` is always `''` for tool-call steps — the `responseMessages` for a tool-call step contain `{ type: "text", text: "" }` alongside the tool-call content. Our behavior matches this upstream convention.
- The original issue report itself showed `"text":"","toolCalls":[]` in step results, confirming this is the expected shape.
- Models that emit "thinking" text before tool calls are a minority, the text is inconsistent and not guaranteed, and no UI framework relies on it.
- The only scenario where this could matter is raw SSE stream consumers who want to show pre-tool-call thinking text as a UX indicator. If that becomes a real need, a future improvement could preserve the text in step metadata without sending it as `text-delta` SSE events, or attempt to distinguish legitimate text from leaked JSON by comparing against prior tool results.

Edge case tests are included documenting this behavior explicitly so it's easy to revisit.

## Related Issue(s)

Fixes #13268

## Type of Change

## Checklist


LekoArts:

PR #13804, opened 2026-03-04
refactor: remove virtual-root path convention from LocalFilesystem

CalebBarnes:
## Summary

Absolute paths (starting with `/`) are now always treated as real filesystem paths. Relative paths are resolved against `basePath`. The old "virtual-root" convention where `/file.txt` silently mapped to `basePath/file.txt` is removed.

Also fixes two bugs in `LocalFilesystem` discovered during testing:
- `allowedPaths` resolved against `process.cwd()` instead of `basePath`, causing incorrect containment checks
- `assertPathContained` rejected access to non-existent `allowedPaths` directories (e.g., skills discovery paths)

## Problem

In contained mode, `LocalFilesystem` treated absolute paths as virtual-root paths (e.g., `/file.txt` -> `basePath/file.txt`). This caused:

- **Split brain**: `view` and `write_file` could silently resolve the same absolute path to different locations
- **Misleading errors**: paths outside `basePath` showed "File not found" instead of "Permission denied"
- **Silent nesting**: `/Users/caleb/.config/file.txt` would resolve to `basePath/Users/caleb/.config/file.txt`

### Virtual-root removal (5 source files)

| File | Change |
|------|--------|
| `local-filesystem.ts` | `resolvePath` simplified via new `resolveToBasePath` utility. Absolute paths go through containment check directly. `toRelativePath` no longer prepends `/` |
| `composite-filesystem.ts` | `resolveMount` passes relative `fsPath` to underlying filesystems. `normalizePath` handles `.` as root |
| `glob.ts` | `extractGlobBase` returns `.` instead of `/` for root patterns. `walkAll` builds relative paths |
| `tree-formatter.ts` | `getRelativePath` and `joinPath` handle `.` as root-equivalent |
| `workspace.ts` | `getAllFiles(".")`, updated JSDoc examples, simplified directory coverage check |

### Bug fixes (2 source files)

| File | Change |
|------|--------|
| `local-filesystem.ts` | `allowedPaths` now resolve against `basePath` (not `cwd`). `assertPathContained` handles non-existent roots correctly |
| `fs-utils.ts` | New `resolveToBasePath` utility replaces `resolveWorkspacePath`, used by both `resolvePath` and `allowedPaths` |

### Tests (19 files)

All test files updated to use relative paths instead of virtual-root paths (e.g., `readFile("test.txt")` instead of `readFile("/test.txt")`). New tests for `allowedPaths` resolution against `basePath` and non-existent allowed paths.

## Verification

All **1440 tests pass** across workspace test suite, zero failures, zero type errors.

## Summary by CodeRabbit

* **Bug Fixes**
  * allowedPaths now resolve from the workspace base path; non-existent allowed roots no longer cause false permission errors.

* **New Features**
  * Tilde (~) expansion for path inputs.
  * Default path shorthand changed from "/" to "." across tools and UI.

* **Documentation**
  * Clarified path-handling rules and added runtime examples for updating allowed paths.

* **Chores**
  * UI and tools standardized to use workspace-root-relative paths (".") consistently.

### DIFF
--- a/.changeset/no-virtual-root.md
+++ b/.changeset/no-virtual-root.md
@@ -1,6 +1,8 @@
 ---
 '@mastra/core': minor
 '@mastra/playground-ui': patch
+'@mastra/gcs': patch
+'@mastra/s3': patch
 ---
 
 **Absolute paths now resolve to real filesystem locations instead of being treated as workspace-relative.**
@@ -23,3 +25,5 @@ await filesystem.readFile('/src/index.ts');
 // After
 await filesystem.readFile('src/index.ts');
 ```
+
+GCS and S3 filesystem providers now also accept `.` as a root path, ensuring compatibility when used as mounts in `CompositeFilesystem`.
--- a/.changeset/no-virtual-root.md
+++ b/.changeset/no-virtual-root.md
@@ -23,3 +23,7 @@ await filesystem.readFile('/src/index.ts');
 await filesystem.readFile('src/index.ts');
 ```
 
+Also fixed:
+- `allowedPaths` resolving against the working directory instead of `basePath`, causing unexpected permission errors when `basePath` differed from `cwd`
+- Permission errors when accessing paths under `allowedPaths` directories that don't exist yet (e.g., during skills discovery)
+
--- a/.changeset/no-virtual-root.md
+++ b/.changeset/no-virtual-root.md
@@ -1,6 +1,5 @@
 ---
 '@mastra/core': minor
-'@mastra/playground-ui': patch
 ---
 
 **Absolute paths now resolve to real filesystem locations instead of being treated as workspace-relative.**
--- a/.changeset/no-virtual-root.md
+++ b/.changeset/no-virtual-root.md
@@ -4,7 +4,7 @@
 
 **Absolute paths now resolve to real filesystem locations instead of being treated as workspace-relative.**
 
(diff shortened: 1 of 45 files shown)


packages/core/src/workspace/filesystem/composite-filesystem.ts
```diff
@@ -179,9 +179,10 @@ export class CompositeFilesystem<
   }
 
   private normalizePath(path: string): string {
-    if (!path || path === '/') return '/';
+    if (!path || path === '/' || path === '.') return '/';
```
TylerBarnes, line 182: isn't this wrong? `.` should stay as is? cwd, / is root

packages/core/src/workspace/filesystem/composite-filesystem.ts
```diff
@@ -179,9 +179,10 @@ export class CompositeFilesystem<
...
     // posix.normalize resolves dot segments (./foo → foo, a/../b → b)
     let n = posixPath.normalize(path);
+    if (n === '.') return '/';
```
TylerBarnes, line 185: and this

packages/core/src/workspace/filesystem/composite-filesystem.ts
CalebBarnes: This is a bit different for composite filesystem. It's not a real filesystem here. Will have to check why this was here but I think it had something to do with mapping to the underlying FS

.changeset/no-virtual-root.md
LekoArts, line 1: You need to split up this changeset because the description isn't relevant to `@mastra/playground-ui` at all, the UI only got a small update

NikAiyer (approved):

ginlink:
This will cause files to fail to load properly in the studio:

image

PR #13834, opened 2026-03-05
feat(filesystem): add prefix scoping and STS session token support for S3/GCS mounts

okxiaoliang4:
## Description

- Add prefix option to S3 and GCS mount configs to scope FUSE mounts to
  a subdirectory (s3fs bucket:/prefix syntax, gcsfuse --only-dir flag)
- Add sessionToken support to S3 mounts for STS/AssumeRole temporary credentials
  via AWS credentials file (INI format) rather than passwd file
- Add validatePrefix() to guard against shell injection in prefix paths
- Expose prefix from getMountConfig() in both S3Filesystem and GCSFilesystem
- Add tests for prefix inclusion, trailing-slash stripping, and omission

## Type of Change

- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [x] Test update

## Checklist

## New Features
- GCS mounts now support optional path-based scoping to limit access to specific directories
- S3 mounts now support optional path-based scoping to limit access to specific key prefixes
- S3 authentication now supports optional session tokens for temporary credential management
- Added validation for mount path prefixes to enforce safe configuration standards

### DIFF
--- a/workspaces/e2b/src/sandbox/mounts/gcs.ts
+++ b/workspaces/e2b/src/sandbox/mounts/gcs.ts
@@ -1,6 +1,6 @@
 import type { FilesystemMountConfig } from '@mastra/core/workspace';
 
-import { LOG_PREFIX, validateBucketName } from './types';
+import { LOG_PREFIX, validateBucketName, validatePrefix } from './types';
 import type { MountContext } from './types';
 
 /**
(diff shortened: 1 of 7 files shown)


abhiaiyer91:
Thanks for the contribution! After pulling latest main, most of this PR's features have already landed:

- **S3 prefix scoping** — `E2BS3MountConfig.prefix`, `bucket:/prefix` syntax in `mountS3()`, and `validatePrefix()` are all on main
- **SDK-level prefix** — Both `S3Filesystem` and `GCSFilesystem` support `prefix` in their config and `getMountConfig()`
- **S3 sessionToken (SDK)** — `S3MountConfig` already includes `sessionToken` and `getMountConfig()` returns it

**What's still missing** (and would make great focused follow-up PRs):

1. **E2B sessionToken wiring** — `E2BS3MountConfig` doesn't include `sessionToken`, and `mountS3()` doesn't write `~/.aws/credentials` in INI format for STS. This is the main gap — s3fs needs the INI credentials file with `aws_session_token` plus the `-o use_session_token` flag.

2. **GCS prefix in E2B mount** — `E2BGCSMountConfig` doesn't have `prefix`, and `mountGCS()` doesn't pass `--only-dir` to gcsfuse. The SDK layer has the field but the E2B mount layer doesn't use it.

Closing since the bulk of the work is already on main. If you'd like to tackle either of the remaining gaps above, a smaller PR focused on just that feature would be easy to review and merge. 🙏

PR #13858, opened 2026-03-05
fix(deployer): use export * for workspace packages to fix missing transitive exports

gourine:
## Problem

When workspace packages use shared rollup chunks internally (as `@mastra/core` does), static analysis of the user entry only captures directly-imported named exports. Exports referenced *transitively* via intra-package chunk imports are missed.

**Concrete failure:** `InMemoryStore` was recently added to `@mastra/core/storage`. The `Mastra` class in `@mastra/core/mastra` references it via a shared chunk (`chunk-E7S6YPRX.js`). During `mastra dev`, the bundler:
1. Correctly maps the shared chunk to the `@mastra/core/storage` cache in the main bundle
2. But the cache was generated *without* `InMemoryStore` because static analysis of the user entry never directly imports it

Result: runtime crash on `mastra dev` with any project using the local workspace:
```
SyntaxError: The requested module '.../@mastra__core__storage.mjs' does not provide an export named 'InMemoryStore'
```

## Fix

For workspace packages, emit `export * from 'pkg'` in the virtual module instead of an explicit named-export list. This guarantees all named exports are present in the cache regardless of how they are referenced transitively.

Non-workspace (third-party) packages retain the named-export optimisation for tree-shaking.

## Tests

Updated 2 existing unit tests that encoded the old (buggy) behavior. All 304 deployer tests pass.

## Summary by CodeRabbit

* **Refactor**
  * Changed how workspace dependencies are emitted during builds: now uses broader re-exports with conditional default export handling for workspace packages while preserving previous behavior for external dependencies.
* **Tests**
  * Added and updated tests to validate the new emission patterns, including behavior when a workspace package has no default export.


wardpeet:
Pretty sure this is fixed

PR #13873, opened 2026-03-05
fix(server): only bypass auth in dev mode when no auth provider is configured

rphansen91:
## Summary

The dev playground auth bypass was unconditionally skipping auth for all localhost requests with `MASTRA_DEV=true`, which prevented auth development/testing locally. Now the bypass only applies when no real auth provider (with `authenticateToken`) is configured.

## Changes

- **Capabilities endpoint** (`auth.ts` handler): Return `{ enabled: false }` when no auth provider exists, regardless of dev playground status. Removes the `isDevPlaygroundRequest` check entirely — if `getAuthProvider()` returns null, auth is disabled.
- **`coreAuthMiddleware`** (`helpers.ts`): Only use dev playground bypass when no auth provider is configured (`!hasAuthProvider`), so the full auth flow runs when auth IS configured — ensuring user/roles/permissions are set in `requestContext`.
- **Tests**: Updated to cover both scenarios (no provider → disabled, provider configured → real capabilities).
- **Bonus**: Fix pre-existing TS errors in `permission-generator.ts` (default values for destructured `split()` result).

## Problem

When auth was configured and running locally (`MASTRA_DEV=true` + localhost), the `isDevPlaygroundRequest()` check in `coreAuthMiddleware` was returning `pass` immediately — without setting user context. Downstream route handlers checking permissions would then fail with 403 because `userPermissions` was undefined.

## Test Plan

- `pnpm vitest run src/server/handlers/auth.test.ts` — 2/2 passing
- `pnpm vitest run src/server/auth/helpers.test.ts` — 60/60 passing
- `tsc --noEmit` on server package — clean (0 errors)

## Summary by CodeRabbit

* **New Features**
  * Added support flag for fine-grained authorization (FGA) in capability reporting.

* **Bug Fixes**
  * Prevented undefined values when parsing permission patterns, improving permission descriptions.

* **Refactor**
  * Dev-mode authentication bypass now only applies when no auth provider is configured; auth capabilities reflect actual provider presence.

* **Tests**
  * Updated auth tests to validate capability behavior with and without an auth provider.


rphansen91:
## Does this rebreak #13794?

**No.** The original issue was a user with plain auth config (e.g. `auth: { protected: ['/api/*'] }`) — no `authenticateToken` method. With this PR:

- **Plain config** (`auth: { protected: [...] }`) → `getAuthProvider()` returns `null` → returns `{ enabled: false, login: null }` → no login gate. Same behavior as the #13801 fix. ✅
- **Real provider** (`auth: new MastraAuthStudio(...)`) → `getAuthProvider()` returns the provider → proceeds to `buildCapabilities()` → returns real capabilities with SSO/credentials login. ✅

The key distinction is `getAuthProvider()` checks for the `authenticateToken` method. Plain config objects don't have it, so they're treated as "no auth provider" and the playground stays accessible. Only when a user has explicitly configured a full auth provider (which provides a login method) does the login gate appear — which is the correct behavior in that case.

graysonhicks (approved):

abhiaiyer91 (approved):

PR #13881, opened 2026-03-05
fix(core): prevent supervisor messages from leaking into sub-agent thread

### LINKED ISSUE #14583: Supervisor/Subagent Persistence Duplication in Delegated Chat Threads
This issue was created from Discord post 1485421279275257866:

[Open in Discord](https://discord.com/channels/1309558646228779139/1485421279275257866)

We ran into a persistence issue with a supervisor/subagent setup in Mastra and wanted to share what we found.

Setup:
- Supervisor agent: `supervisorAgent`
- Delegated subagent: `workerAgent`
- Native Mastra delegation
- Chat thread memory enabled on the supervisor
- Goal: only the supervisor’s final assistant reply should be persisted in the main thread

Observed issue:
- Runtime UI looked correct: only one assistant reply showed up
- But after reloading the thread, we consistently got 2 assistant messages with different IDs and identical text
- This only happened when delegation was involved

What we found:
The duplicate was actually two separate persisted assistant messages:
1. one persisted assistant message from the delegated `workerAgent`
2. one persisted assistant message from the `supervisorAgent`

So both the subagent output and the supervisor output were being saved as assistant turns.

How we verified that:
- the persisted “subagent” message contained the direct workspace/tool call
- the persisted “supervisor” message contained the `agent-workerAgent` delegation tool result
- so it was not just one stored message rendered twice

Workaround that fixed it for us:
- added an output processor directly on the delegated subagent
- that processor filters out assistant messages from the subagent before persistence
- result: the subagent still executes and returns its result to the supervisor, but only the supervisor persists the final assistant reply

gourine:
Fixes #14583

## Summary

When a supervisor agent delegates work to a sub-agent, the supervisor's full conversation history (`filteredContextMessages`) was being passed as part of the `messages` argument. Because `MessageHistory.processOutputResult` saves all messages with `source: 'input'` to the sub-agent's thread, this caused the supervisor's own tool call records (e.g. `agent-codeExpert`) to be written into the sub-agent's thread.

**Effect in the studio UI:** When viewing chat history, the sub-agent badge showed a nested duplicate of itself instead of showing the sub-agent's actual tool calls (e.g. MCP tools like `search-code`, `get-file-contents`). Opening the inner duplicate badge showed nothing.

## Root cause

```ts
// BEFORE — supervisor history included in sub-agent 'messages', persisted to sub-agent thread:
const messagesForSubAgent = [...filteredContextMessages, { role: 'user', content: effectivePrompt }];
agent.stream(messagesForSubAgent, { ... });
```

## Fix

Pass `filteredContextMessages` as the `context` option instead. The `context` source is added to the `MessageList` as `source: 'context'`, which `MessageHistory` explicitly excludes from persistence. The sub-agent's LLM still receives the full conversation history, but the supervisor's messages are never written to the sub-agent's thread.

```ts
// AFTER — supervisor history passed as context, NOT persisted:
const messagesForSubAgent = [{ role: 'user', content: effectivePrompt }];
agent.stream(messagesForSubAgent, { context: filteredContextMessages, ... });
```

## Test plan

- [ ] Send a message to a supervisor agent that delegates to a named sub-agent
- [ ] Verify the sub-agent badge in the studio shows the sub-agent's own tool calls (e.g. MCP tools), not a duplicate `agent-<name>` badge
- [ ] Send multiple messages in the same thread — confirm no duplicate badges appear in history
- [ ] Confirm the sub-agent still produces correct responses (context is still passed to the LLM)


taofeeq-deru (approved):

wardpeet (approved):
Thank you, great catch!

PR #13882, opened 2026-03-05
fix(studio): keep sub-agent badge expanded while streaming

gourine:
## Summary

When a supervisor agent delegates to a sub-agent, the studio UI shows a collapsible badge representing the sub-agent's work. This PR fixes two related issues with that badge:

1. **Badge collapsed immediately** — as soon as the first streaming chunk arrived, `assistant-ui` saw a `result` field on the tool-call content part and reported `status='complete'`, causing the badge to auto-collapse before the sub-agent had done any work.

2. **Badge showed no content while open** — fixing issue 1 by removing `result` during streaming left the badge empty (no child tool calls or sub-agent responses visible while streaming).

## Root cause

`toAssistantUIMessage.ts` populates a `result` field on the tool-call content part as soon as `output` is available on the dynamic-tool part. During streaming this happens incrementally (the `output.childMessages` array grows with each event), but the part's `state` is still `'input-available'`. Because `assistant-ui` infers `status='complete'` from the presence of any `result`, the badge collapsed prematurely.

## Fix

**`toAssistantUIMessage.ts`**: Always include `result` when `output` is present (so child tool calls are visible during streaming), but tag it with `_mastraStreaming: true` when `state !== 'output-available'`.

**`tool-fallback.tsx`**: Gate the `isComplete` flag on the absence of `_mastraStreaming` so the badge stays expanded for the full duration of sub-agent work and only auto-collapses when the sub-agent truly finishes.

```ts
// toAssistantUIMessage.ts
if ('output' in part) {
  if (part.state !== 'output-available') {
    return { ...baseToolCall, result: { ...part.output, _mastraStreaming: true } };
  }
  return { ...baseToolCall, result: part.output };
}

// tool-fallback.tsx
const isComplete = props.status?.type === 'complete' && result?._mastraStreaming !== true;
```

## Test plan

- [ ] Send a message to a supervisor agent that calls a sub-agent with multiple tool calls
- [ ] Verify the sub-agent badge **opens** when the sub-agent starts and shows its child tool calls (e.g. MCP tools) and streaming response while active
- [ ] Verify the badge **auto-collapses** only after the sub-agent finishes and the result is returned to the parent agent
- [ ] Reload the page and confirm previously completed sub-agent badges load already-collapsed (history view)

## Related

Companion to #13881 (prevents supervisor messages from leaking into the sub-agent thread).

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented sub-agent badges from collapsing prematurely during streaming: badges now remain expanded while sub-agents stream tool calls and responses and only auto-collapse once the sub-agent finishes.
  * No other visible UI behavior changes in this release.


abhiaiyer91:
Closing for now, please open an issue

PR #13888, opened 2026-03-06
fix: add OM runtime backpressure near the token threshold

TylerBarnes:
This adds runtime backpressure to observational memory as turns approach the message token threshold.

## What changed

- ramp async buffering more aggressively near the threshold so buffered chunks are created at tighter boundaries
- pre-seal eligible messages shortly before threshold handling so buffered activation has cleaner swap points
- add tiered backpressure waits near/above threshold
- only wait when buffered observations still look insufficient for activation, and short-circuit the wait when in-flight buffering finishes early
- keep the runtime covered with targeted OM tests

## Why

The goal is to reduce bad threshold crossings where OM reaches the threshold before buffered observations are ready to activate cleanly. In practice this should make threshold handling less collapse-prone and reduce unnecessary waiting when buffering has already done enough work.


LekoArts:

PR #13891, opened 2026-03-06
feat(mastracode): allow overriding memory instance via config

tulga-bytes:
## Summary

Adds an optional `memory` property to `MastraCodeConfig` that lets callers supply their own memory instance (or dynamic factory) instead of relying on the built-in `getDynamicMemory`.

## Motivation

The default `getDynamicMemory(storage)` uses mastracode's built-in model resolution (Anthropic OAuth, OpenAI Codex, models.dev gateway). When models are served by a custom provider that `resolveModel` cannot resolve, there is no way to inject a custom memory configuration.

## Changes

- Added `memory?: HarnessConfig['memory']` to `MastraCodeConfig` interface with JSDoc documentation
- Updated `createMastraCode` to use `config.memory` when provided, falling back to the default `getDynamicMemory(storage)`

## Testing

This is a backward-compatible change — when `memory` is not provided, the existing behavior is preserved.

---
Pull Request opened by [Augment Code](https://www.augmentcode.com/) with guidance from the PR author

ELI5: You can now tell MastraCode to use your own memory system instead of its built-in one by passing a memory option when creating it. This is useful when your models or provider setup can't be resolved by MastraCode's defaults.

- Added an optional memory?: HarnessConfig['memory'] property to MastraCodeConfig with JSDoc explaining its purpose and when to use it.
- createMastraCode now initializes memory using the provided config.memory when present; otherwise it falls back to the existing getDynamicMemory(storage) behavior (memory = config?.memory ?? getDynamicMemory(storage)).
- Motivation: allows callers to inject custom memory instances or factories when models are served by custom providers that resolveModel cannot handle.
- Backward-compatible: if memory is not provided, behavior is unchanged.


TylerBarnes (approved):
looks good! just needs a changeset

PR #13895, opened 2026-03-06
fix(core): remap provider tool types when routing to V3 models via gateway

### LINKED ISSUE #13667: Provider tools broken when using custom gateway with AI SDK v6
This issue was created from Discord post 1477996465530994709:

[Open in Discord](https://discord.com/channels/1309558646228779139/1477996465530994709)

We're using Portkey and have implemented a custom gateway to route things that way.

With AI SDK v6, it seems provider side tools are not working. AFAICT the problem is that 
`ModelRouterLanguageModel` hardcodes `specificationVersion = "v2"` regardless of what the underlying model actually is. The gateway returns a v3 model (from `.responses()`), but Mastra thinks it's v2.

This matters because the tool prep code uses this version to decide the tool type format:

- v3 (AI SDK v6): `type: "provider"`
- v2 (AI SDK v5): `type: "provider-defined"`

Since the version is always "v2", tools get `type: "provider-defined"`. AI SDK v6 rejects this unrecognized type, so the tools array ends up empty.

- [`packages/core/src/llm/model/router.ts#L80`](https://github.com/mastra-ai/mastra/blob/main/packages/core/src/llm/model/router.ts#L80) - `ModelRouterLanguageModel` hardcodes `specificationVersion = "v2"`
- [`packages/core/src/stream/aisdk/v5/compat/prepare-tools.ts#L128`](https://github.com/mastra-ai/mastra/blob/main/packages/core/src/stream/aisdk/v5/compat/prepare-tools.ts#L128) - uses `targetVersion` to decide tool type format

roaminro:
Fixes #13667

`ModelRouterLanguageModel` hardcodes `specificationVersion = 'v2'`, so when a custom gateway returns an AI SDK v6 (V3) model (e.g. via `.responses()`), provider tools like `openai.tools.webSearch()` get prepared with `type: 'provider-defined'` instead of `type: 'provider'`. AI SDK v6 rejects the unrecognized type and silently drops all provider tools.

The fix adds a `remapToolsToV3()` step in `AISDKV6LanguageModel` — the V2→V3 model wrapper that all V3 model calls flow through. This ensures provider tool types are normalized from `'provider-defined'` to `'provider'` regardless of how the tools were prepared upstream, benefiting all V3 model usage (not just the router path).

## Test plan
- [x] Added 6 regression tests covering doStream, doGenerate, function tools, and mixed tool scenarios
- [x] Existing router tests pass (router-supported-urls, custom-gateway-integration)
- [x] Existing prepare-tools tests pass


mbirkegaard:
This is great, @roaminro. Thanks! Do you have a rough idea about when it'll get merged and released?

wardpeet (approved):

PR #13920, opened 2026-03-06
fix(server): serialize plain JSON Schema tool schemas

### LINKED ISSUE #13909: zodToJsonSchema fails on tools with pre-converted JSON Schema (e.g. Composio)
## Description

`getSerializedAgentTools` in `@mastra/server` calls `zodToJsonSchema()` on tool schemas (input, output, requestContext). This fails when a tool's schema is already a plain JSON Schema object rather than a Zod schema — which happens with Composio tools.

## Root cause

The `@composio/mastra` provider wraps tools using `applyCompatLayer` from `@mastra/schema-compat` with `mode: "jsonSchema"`. This converts the original JSON Schema → Zod → a hybrid object. When `getSerializedAgentTools` then calls `zodToJsonSchema()` on this hybrid, it fails with:

```
Cannot read properties of undefined (reading 'typeName')
```

The round-trip (JSON Schema → Zod → JSON Schema) breaks because the intermediate Zod representation doesn't carry the metadata that `zodToJsonSchema` expects.

## Impact

- Tools from Composio (and potentially other providers using `applyCompatLayer`) fail to serialize
- Mastra Studio can't list these tools — they silently disappear from the UI
- Tool *execution* works fine (Zod validates inputs correctly); only *listing/serialization* for the Studio UI breaks

## Suggested fix

Before calling `zodToJsonSchema()`, check if the schema is already a plain JSON Schema object (has `type`, `properties`, or `$schema`) and return it directly:

```js
function safeZodToJsonSchema(schema) {
  try {
    return zodToJsonSchema(schema);
  } catch {
    if (schema && typeof schema === "object" && ("type" in schema || "properties" in schema || "$schema" in schema)) {
      return schema;
    }
    throw new Error("Failed to serialize tool schema");
  }
}
```

This applies to the three `zodToJsonSchema()` calls in `getSerializedAgentTools` (for `inputSchema`, `outputSchema`, and `requestContextSchema`).

## Workaround

We're using a pnpm patch on `@mastra/server` that wraps the three calls with the fallback above.

## Related

- Composio issue: https://github.com/ComposioHQ/composio/issues/2109
- Affects `@mastra/server@1.8.0` and `@mastra/server@1.10.0` (likely all versions)

## Environment

- `@mastra/server`: 1.10.0
- `@composio/mastra`: 0.6.3
- `zod-to-json-schema`: 3.25.1

TheIsrael1:
## Summary
- fix `getSerializedAgentTools` to serialize plain JSON Schema objects without routing them through `zodToJsonSchema`
- reuse core's `asJsonSchema` via `@mastra/core/utils/zod-to-json` and resolve lazy schema functions before conversion
- add regression test for JSON Schema input/output/requestContext schemas in agent tool serialization
- add changesets for `@mastra/server` and `@mastra/core` (including server peer dependency minimum update)

## Verification
- `pnpm --filter @mastra/server test -- src/server/handlers/agent.test.ts`

Fixes #13909

## Release Notes

* **Bug Fixes**
  * Fixed server-side tool serialization for plain JSON Schema objects to prevent tools from disappearing in Studio.


packages/server/src/server/handlers/agents.ts
```diff
@@ -197,53 +204,21 @@ export async function getSerializedAgentTools(
...
-            const inputSchema = tool.inputSchema();
-            if (inputSchema && inputSchema.jsonSchema) {
-              inputSchemaForReturn = stringify(inputSchema.jsonSchema);
-            }
-          } else if (tool.inputSchema) {
-            inputSchemaForReturn = stringify(
-              zodToJsonSchema(tool.inputSchema as Parameters<typeof zodToJsonSchema>[0]),
-            );
-          }
+        const inputSchema = asJsonSchema(resolveLazySchema(tool.inputSchema) as Parameters<typeof asJsonSchema>[0]);
```
wardpeet: you should be able to use `import { toStandardSchema, standardSchemaToJSONSchema } from '@mastra/schema-compat/schema';`

wardpeet (approved):

PR #13934, opened 2026-03-06
fix(playground-ui): lazy-load prettier to reduce main bundle size

rphansen91:
## What

Convert static prettier imports to dynamic `import()` in `formatting.ts` so prettier + plugins (~600KB) are loaded on-demand when the user formats JSON, instead of blocking the critical rendering path on every page load.

## Why

The studio main bundle is 5.8MB. Prettier (standalone + babel plugin + estree plugin) accounts for ~600KB of that, but is only needed when a user clicks "format JSON" — an uncommon interaction that shouldn't delay initial render.

## Changes

- **1 file changed**: `packages/playground-ui/src/lib/formatting.ts`
- Replaced 3 static `import` statements with `Promise.all([import(...)])` inside `formatJSON()`

## Bundle impact

| | Before | After | Diff |
|---|---|---|---|
| Main bundle (raw) | 5,834 KB | 5,201 KB | **-634 KB** |
| Main bundle (gzip) | 1,736 KB | 1,557 KB | **-180 KB** |

Prettier is now split into 3 lazy-loaded chunks:
- `standalone-*.js` (87 KB)
- `babel-*.js` (317 KB)  
- `estree-*.js` (204 KB)

## Test plan

1. Build passes (`pnpm build` + `tsc --noEmit`)
2. Open studio → main bundle no longer contains prettier
3. Navigate to request context / agent settings / workflow time-travel → click format JSON → prettier loads on demand, formatting works as before

## Summary by CodeRabbit

* **Refactor**
  * JSON formatter now lazy-loads the code formatter and its plugins at runtime to reduce initial bundle size; formatting behavior and public API remain unchanged.
  * Adds robust error handling that clears cached loaders on failure.

* **Chores**
  * Adds a changeset recording a patch version bump and noting the lazy-load performance improvement.


abhiaiyer91 (approved):

wardpeet (approved):

packages/playground-ui/src/lib/formatting.ts
```diff
@@ -1,8 +1,26 @@
...
+        typeof import('prettier/plugins/estree'),
+      ]
+    >
+  | undefined;
+
+const loadPrettier = () =>
+  (prettierModules ??= Promise.all([
+    import('prettier/standalone'),
+    import('prettier/plugins/babel'),
+    import('prettier/plugins/estree'),
```
wardpeet, line 15: A more optimized way is to create a single file (less requests used)

```
import prettier from 'prettier/standalone'
import prettierPluginBabel from 'prettier/plugins/babel'
import prettierPluginEstree from 'prettier/plugins/estree'

export { prettier, prettierPluginBabel, prettierPluginEstree }
```

and do a dynamic import of the above file. that means only one file is created during bundling instead of 3, which means less http bloat

PR #13989, opened 2026-03-09
fix(server): fallback to agent toolsResolver in EXECUTE_TOOL_ROUTE

### LINKED ISSUE #13987: [BUG] /tools/:toolId/execute cannot find dynamically-resolved agent tools (toolsResolver)
## Description

The `EXECUTE_TOOL_ROUTE` handler (`POST /tools/:toolId/execute`) cannot find tools that are dynamically resolved via an agent's `toolsResolver`. It only searches:

1. `registeredTools` (set via middleware, always `{}` for the global tools endpoint)
2. `mastra.getToolById()` (only searches statically registered tools)

This means any tools provided by `toolsResolver` (e.g. Composio integration tools) are invisible to the global tool execute endpoint, even though they show up correctly when listing agent tools via `/agents/:agentId/tools`.

## Steps to Reproduce

1. Create an agent with a `toolsResolver` that returns dynamic tools (e.g. Composio integration tools resolved per-request based on user context)
2. Open Mastra Studio
3. Navigate to the agent — tools are listed correctly
4. Try to execute one of the dynamically-resolved tools
5. Error: `Tool with id TOOLNAME not found`

## Expected Behavior

The `/tools/:toolId/execute` endpoint should fall back to searching agent tools (via `agent.listTools({ requestContext })`) when the tool is not found in the global registry.

## Current Behavior

```
Error: Tool with id GOOGLECALENDAR_LIST_CALENDARS not found
    at Mastra.getToolById (chunk-GTA5BKXZ.js:1390:19)
    at Object.handler (tools.ts:158:23)
```

## Suggested Fix

In the `EXECUTE_TOOL_ROUTE` handler, after the global lookup fails, iterate through agents and call `agent.listTools({ requestContext })` to find dynamically-resolved tools — similar to how `EXECUTE_AGENT_TOOL_ROUTE` already works:

```javascript
// After global lookup fails:
if (!tool) {
  const agents = mastra.listAgents() || {};
  for (const agent of Object.values(agents)) {
    try {
      const agentTools = await agent.listTools({ requestContext });
      tool = Object.values(agentTools || {}).find((t) => t.id === toolId);
      if (tool) break;
    } catch {
      // continue
    }
  }
}
```

## Context

- Mastra version: `1.10.0`
- The `EXECUTE_AGENT_TOOL_ROUTE` (`/agents/:agentId/tools/:toolId/execute`) works correctly because it calls `agent.listTools({ requestContext })`
- We're currently working around this with a patch to `@mastra/server`

Aayush-engineer:
## What does this PR do?

Fixes #13987

`EXECUTE_TOOL_ROUTE` (`POST /tools/:toolId/execute`) was throwing 
404 for tools dynamically resolved via `toolsResolver` because it 
only searched `registeredTools` and `mastra.getToolById()` — both 
of which only contain statically registered tools.

## Fix

After the global lookup fails, the handler now iterates through all 
agents and calls `agent.listTools({ requestContext })` to find 
dynamically-resolved tools — the same approach already used by 
`EXECUTE_AGENT_TOOL_ROUTE`.

## Changes

- `packages/server/src/server/handlers/tools.ts` — added fallback 
  lookup in `EXECUTE_TOOL_ROUTE` after global tool search fails

## Testing

1. Create an agent with `toolsResolver` returning dynamic tools
2. Call `POST /tools/:toolId/execute` with a dynamic tool ID
3. Tool executes correctly instead of returning 404

## Summary by CodeRabbit

* **Bug Fixes**
  * Tool lookup now falls back to dynamically-resolved agent tools in both GET and EXECUTE flows, reducing false 404s and improving tool discovery.

* **Tests**
  * Added tests for executing tools supplied dynamically by agents and for resilience when an agent's resolver throws errors.

* **Chores**
  * Added a patch changeset to bump server and core packages.


packages/server/src/server/handlers/tools.ts
```diff
@@ -158,6 +158,20 @@ export const EXECUTE_TOOL_ROUTE = createRoute({
         tool = mastra.getToolById(toolId);
       }
 
+      // Fallback: search dynamically-resolved agent tools (toolsResolver)
+      if (!tool) {
+        const agents = mastra.listAgents() || {};
```
abhiaiyer91: should pass requestContext

Aayush-engineer: Hi @abhiaiyer91 — `requestContext` is already available from the 
handler params (line 146) and is being passed to 
`agent.listTools({ requestContext })` on line 166, which is where 
dynamic tool resolution actually happens.

`mastra.listAgents()` has no parameters — it simply returns 
`this.#agents` directly (`packages/core/src/mastra/index.ts` line 832).

Let me know if you had something else in mind!

Aayush-engineer:
Hi @abhiaiyer91 — rebased with main, PR now contains only the server toolsResolver fallback fix. All feedback from your previous review has been addressed. Would appreciate a re-review when you get a chance!

abhiaiyer91 (approved):

PR #14025, opened 2026-03-09
Add input message preview column to trace list

### LINKED ISSUE #14006: [FEATURE] Show input message preview column in Observability trace list
### Problem Statement

The trace list only shows ID, date, time, name, entity, and status. To find a specific trace, you have to click into each one individually. When a user reports an issue and describes what they said, there is no way to visually scan for that message in the list.

### Proposed Solution

Add a column to the trace list showing a truncated preview (~100 chars) of the input message/prompt from the root span. This data is already available in the span attributes.

### Component

Playground UI

### Alternatives Considered

Currently clicking through each trace individually to inspect the input. This is time-consuming with high trace volumes.

### Example Use Case

As a developer, I want to see a preview of the user's input message directly in the trace list so that when a user reports "I asked the agent about X and got a wrong answer", I can visually scan the list and find the relevant trace immediately.

### Verification

graysonhicks:
## Description

Added an "Input" column to the observability trace list showing a truncated preview of user input messages. This enables developers to visually scan traces and find specific ones based on user input without clicking into each trace individually.

The implementation includes a new `getInputPreview()` utility that extracts user message content from agent traces, handles multipart content arrays, and truncates long text to 100 characters with an ellipsis.

image

## Related Issue(s)

Fixes #14006

## Type of Change

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Added an input message preview column to the observability trace list, allowing users to view truncated previews of user input messages directly within the trace table for better debugging and monitoring visibility.

* **Tests**
  * Added comprehensive test coverage for input preview functionality, including edge cases such as null inputs, multiple messages, content truncation, and various input formats.

### DIFF
--- a/packages/playground-ui/src/domains/observability/components/traces-list.tsx
+++ b/packages/playground-ui/src/domains/observability/components/traces-list.tsx
@@ -2,18 +2,21 @@ import { EntryList } from '@/ds/components/EntryList';
 import { getShortId } from '@/ds/components/Text';
 import { SpanRecord } from '@mastra/core/storage';
 import { format, isToday } from 'date-fns';
+import { getInputPreview } from '../utils/span-utils';
 
 export const tracesListColumns = [
   { name: 'shortId', label: 'ID', size: '6rem' },
   { name: 'date', label: 'Date', size: '4.5rem' },
   { name: 'time', label: 'Time', size: '6.5rem' },
   { name: 'name', label: 'Name', size: '1fr' },
+  { name: 'input', label: 'Input', size: '1fr' },
   { name: 'entityId', label: 'Entity', size: '10rem' },
   { name: 'status', label: 'Status', size: '3rem' },
 ];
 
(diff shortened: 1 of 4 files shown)


packages/playground-ui/src/domains/observability/components/traces-list.tsx
```diff
@@ -64,6 +65,7 @@ export function TracesList({
                     date: isTodayDate ? 'Today' : format(createdAtDate, 'MMM dd'),
                     time: format(createdAtDate, 'h:mm:ss aaa'),
                     name: trace?.name,
+                    input: getInputPreview(trace?.input),
```
rltvty, line 68: maybe we could try to be a little smart about what to show here?

like if there is a single user message with just a content field, only show that instead of the full messages array?

I think that would be much easier to read

greglobinski (approved):

PR #14030, opened 2026-03-09
fix(core): persist step messages to storage when savePerStep is enabled

### LINKED ISSUE #13984: savePerStep does not persist messages to storage — user message lost on stream abort
## Bug Description

`savePerStep: true` does not actually persist messages to the storage backend (e.g. Convex). It only accumulates messages in the in-memory `MessageList` via `saveStepMessages()`, which calls `messageList.add(result.response.messages, "response")` but never calls `saveQueueManager.flushMessages()` or `persistUnsavedMessages()`.

The actual persistence only happens in `executeOnFinish` → `#executeOnFinish`, which is gated by `!abortSignal.aborted` (chunk-YC6Z75K2.js:17972):

```js
const aborted = options.abortSignal?.aborted;
if (!aborted) {
    await capabilities.executeOnFinish({ ... });
}
```

This means if the user refreshes the page mid-stream, the abort signal fires, `executeOnFinish` is skipped entirely, and **no messages are persisted** — including the user's original message that was added to the `MessageList` in the `before()` step.

## Expected Behavior

When `savePerStep: true` is set, messages should be incrementally flushed to storage after each step completes. At minimum, the user message should be persisted before the first LLM step begins, so it survives a stream abort.

## Actual Behavior

`saveStepMessages()` (line 21169) only does:
```js
messageList.add(result.response.messages, "response");
```

It does **not** call `persistUnsavedMessages()` or `enqueueSave()` / `flushMessages()`. The `SaveQueueManager.flushMessages()` is only called in the tool suspension/approval code path (lines 13325, 13349), not in the `onStepFinish` callback for `savePerStep`.

## Code Path Analysis

1. User message added to `MessageList` in `__primitive().before()` at line 16945: `messageList.add(messages, "user")`
2. `onStepFinish` with `savePerStep` (line 17872-17888) calls `saveStepMessages` which only adds to in-memory list
3. On normal completion: `executeOnFinish` persists all messages (user + assistant) — works fine
4. On abort (page refresh): `executeOnFinish` is skipped → **nothing is persisted**, user message is lost

## Reproduction

1. Configure an agent with memory (e.g. Convex storage) and `savePerStep: true`
2. Send a message that triggers a multi-step response (e.g. tool calls + text)
3. Refresh the page while the stream is still active
4. Check the storage — the user message is missing, only partial or no assistant messages are saved

## Suggested Fix

In the `onStepFinish` callback (around line 17872), after `saveStepMessages`, also call `saveQueueManager.flushMessages()` to actually persist to storage:

```js
onStepFinish: async (props) => {
    if (options.savePerStep && !memoryConfig?.readOnly) {
        // ... thread creation logic ...
        await capabilities.saveStepMessages({
            result: props,
            messageList: memoryData.messageList,
            runId
        });
        // Actually persist to storage
        if (saveQueueManager && memoryData.thread?.id) {
            await saveQueueManager.flushMessages(
                memoryData.messageList,
                memoryData.thread.id,
                memoryConfig
            );
        }
    }
    return options.onStepFinish?.({ ...props, runId });
},
```

Additionally, the user message should ideally be flushed to storage in the `before()` step (after `messageList.add(messages, "user")`) so it's persisted before any LLM processing begins.

## Workaround

We manually pre-save the user message to Convex before calling `handleChatStream`. Convex's `batchInsert` is an upsert by record ID, so when `executeOnFinish` later saves the same message on normal completion, it patches the existing record without duplicates.

## Environment

- `@mastra/core`: 1.1.x (checked dist/chunk-YC6Z75K2.js)
- `@mastra/ai-sdk`: latest
- Storage: `@mastra/convex`
- Stream method: `handleChatStream` from `@mastra/ai-sdk`

graysonhicks:
## Description

Fixed issue #13984 where `savePerStep: true` wasn't actually persisting messages to storage during step execution. The problem was that `onStepFinish` callback in the map-results-step only added messages to the in-memory MessageList via `saveStepMessages()` but never flushed them to persistent storage via `saveQueueManager.flushMessages()`. This meant that when streams were aborted or pages were refreshed, the user's original message and any in-progress responses would be lost since `executeOnFinish` (the main persistence path) is gated by `!abortSignal.aborted`.

The fix makes the SaveQueueManager available to the map-results-step and calls `flushMessages()` after `saveStepMessages()` to immediately persist accumulated messages, ensuring they survive page refreshes and stream aborts.

## Related Issue(s)

Fixes #13984

## Type of Change

- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [ ] Test update

## Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * savePerStep now reliably persists messages to storage after each step, ensuring data survives page refreshes and aborted streams.

* **Tests**
  * Added a regression test validating per-step persistence when streams are aborted.

### DIFF
--- a/packages/core/src/agent/workflows/prepare-stream/map-results-step.ts
+++ b/packages/core/src/agent/workflows/prepare-stream/map-results-step.ts
@@ -10,6 +10,7 @@ import { StructuredOutputProcessor } from '../../../processors';
 import type { RequestContext } from '../../../request-context';
 import type { Step } from '../../../workflows';
 import type { InnerAgentExecutionOptions } from '../../agent.types';
+import type { SaveQueueManager } from '../../save-queue';
 import { getModelOutputForTripwire } from '../../trip-wire';
 import type { AgentMethodType } from '../../types';
 import { isSupportedLanguageModel } from '../../utils';
@@ -26,6 +27,7 @@ interface MapResultsStepOptions<OUTPUT = undefined> {
   agentSpan: Span<SpanType.AGENT_RUN>;
   agentId: string;
   methodType: AgentMethodType;
+  saveQueueManager?: SaveQueueManager;
   /**
    * Shared processor state map that persists across agent turns.
    */
@@ -43,6 +45,7 @@ export function createMapResultsStep<OUTPUT = undefined>({
   agentSpan,
   agentId,
   methodType,
+  saveQueueManager,
 }: MapResultsStepOptions<OUTPUT>): Step<
   string,
   unknown,
@@ -89,6 +92,15 @@ export function createMapResultsStep<OUTPUT = undefined>({
             messageList: memoryData.messageList!,
             runId,
           });
+
+          // Flush accumulated messages to persistent storage so they survive
+          // an abort/disconnect.  Without this, saveStepMessages only adds to
+          // the in-memory MessageList and the only persistence path is
+          // executeOnFinish which is gated by !abortSignal.aborted.
+          // See: https://github.com/mastra-ai/mastra/issues/13984
+          if (saveQueueManager && memoryData.thread?.id) {
+            await saveQueueManager.flushMessages(memoryData.messageList!, memoryData.thread.id, memoryConfig);
+          }
         }
 
         return options.onStepFinish?.({ ...props, runId });
--- a/packages/core/src/agent/workflows/prepare-stream/map-results-step.ts
+++ b/packages/core/src/agent/workflows/prepare-stream/map-results-step.ts
@@ -93,11 +93,6 @@ export function createMapResultsStep<OUTPUT = undefined>({
             runId,
           });
 
-          // Flush accumulated messages to persistent storage so they survive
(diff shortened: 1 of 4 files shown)


packages/core/src/agent/workflows/prepare-stream/map-results-step.ts
```diff
@@ -89,6 +92,15 @@ export function createMapResultsStep<OUTPUT = undefined>({
             messageList: memoryData.messageList!,
             runId,
           });
+
+          // Flush accumulated messages to persistent storage so they survive
+          // an abort/disconnect.  Without this, saveStepMessages only adds to
+          // the in-memory MessageList and the only persistence path is
+          // executeOnFinish which is gated by !abortSignal.aborted.
+          // See: https://github.com/mastra-ai/mastra/issues/13984
```
DanielSLew: this commen tisn't really necessary

DanielSLew (approved):

PR #13056, opened 2026-02-13
fix(core): pass writer to output processors in processOutputStream and processOutputResult

graysonhicks:
## Description

Fixes the `writer` object being `undefined` in both `processOutputResult` and `processOutputStream` when implementing custom output processors. This enables use cases like running moderation on output and streaming updated UI events back to the client.

**Changes:**
- Thread `writer` parameter through `ProcessorRunner.processPart()` and `runOutputProcessorsForStream()`
- Create a `ProcessorStreamWriter` from the stream controller in `MastraModelOutput` and `runOutputProcessorsForStream` so processors always have a working writer
- Wire up `streamWriter` in `createLLMMappingStep` for agentic workflow execution
- Add tests verifying writer is passed to both `processOutputResult` and `processOutputStream`

## Type of Change

- [x] Bug fix (non-breaking change that fixes an issue)
- [ ] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [ ] Test update

## Checklist

- [ ] I have made corresponding changes to the documentation (if applicable)
- [x] I have added tests that prove my fix is effective or that my feature works

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed writer being undefined in output processing so processors can emit stream events reliably.

* **New Features**
  * Output processors can emit custom data events during streaming (e.g., moderation updates or UI events) via an injectable stream writer.

* **Documentation**
  * Added docs and examples showing how to emit and consume custom stream events and required chunk naming.

* **Tests**
  * Added tests verifying writer availability in output processors.

PR #14111, opened 2026-03-10
fix(processors): pass writer to output processors in all execution paths

graysonhicks:
## Summary

Fixes `writer` being `undefined` inside `processOutputStream` for all output processors (e.g. GuardRailProcessor) when using `agent.stream()`. This was reported as still broken after PRs #13056 and #13454.

## Root Cause

There are two gaps where `writer` was not being passed through:

### 1. Primary: `processPart` in `ProcessorRunner` (runner.ts)

`combineProcessorsIntoWorkflow` in `agent.ts` wraps **all** user processors into workflows via `createWorkflow` + `createStep(processor)`. This means in `processPart`, every processor hits the `isProcessorWorkflow` branch (line 405), never the regular processor branch (line 457+).

The workflow branch called `executeWorkflowAsProcessor` **without** the `writer` parameter:

```typescript
// BEFORE (runner.ts:420-432)
const result = await this.executeWorkflowAsProcessor(
  processorOrWorkflow,
  { phase: 'outputStream', part, streamParts, state, messageList, retryCount },
  observabilityContext,
  requestContext,
  // ← writer was missing here (5th parameter)
);
```

Without `writer`, the chain breaks:
- `executeWorkflowAsProcessor` sets `outputWriter: undefined` (line 187)
- `workflow.ts:764` sets `processorWriter = undefined`
- `workflow.ts:792` puts `writer: undefined` into `baseContext`
- `workflow.ts:1047` calls `processOutputStream({ ...baseContext })` — writer is undefined

**Fix**: Pass `writer` as the 5th argument to `executeWorkflowAsProcessor`.

### 2. Secondary: `data-*` chunks in workflow loop stream (stream.ts)

The `outputWriter` callback in `workflowLoopStream` processes `data-*` chunks (custom data emitted via `writer.custom()`) through output processors, but didn't create or pass a `ProcessorStreamWriter` to `processPart`.

**Fix**: Create a `dataChunkStreamWriter` that enqueues chunks back into the stream controller, and pass it as the 7th argument to `processPart`.

## Why previous PRs didn't fix this

PRs #13056 and #13454 added writer plumbing and a test in `runner.test.ts` that called `processPart` directly with a bare `Processor` object. That test exercised the **regular processor branch** (line 457+) of `processPart`, which correctly passes writer. But in production, `combineProcessorsIntoWorkflow` wraps all processors into workflows, so they always hit the **workflow branch** (line 405) instead — which was the one missing writer. The test gave false confidence by testing a code path that's never actually reached in the real agent pipeline. That misleading test has been removed in this PR.

## Changes

| File | Change |
|------|--------|
| `packages/core/src/processors/runner.ts` | Pass `writer` to `executeWorkflowAsProcessor` in the `outputStream` phase of `processPart` |
| `packages/core/src/loop/workflows/stream.ts` | Create `dataChunkStreamWriter` and pass to `processPart` for `data-*` chunks |
| `packages/core/src/processors/runner.test.ts` | Remove misleading test that tested writer on bare processors via `processPart` (a code path never hit in production) |
| `packages/core/src/agent/agent-writer-repro.test.ts` | End-to-end test: creates an Agent with a mock model + WriterCaptureProcessor, verifies `writer` is defined for every chunk type through the full `agent.stream()` pipeline |
| `packages/core/src/loop/workflows/stream.test.ts` | Unit test: verifies `writer` is defined when output processors run on `data-*` chunks in workflowLoopStream |
| `.changeset/brave-clouds-mate.md` | Patch changeset for `@mastra/core` |

## Test plan

- [x] `agent-writer-repro.test.ts` — end-to-end verification that writer=DEFINED for all chunk types through `agent.stream()` (response-metadata, text-start, text-delta, text-end, finish)
- [x] `stream.test.ts` — verifies writer is passed for data-* chunks in the workflow loop stream
- [x] `runner.test.ts` — 42 existing tests still pass after removing the misleading test
- [x] `output.test.ts` — existing output stream tests still pass
- [ ] Manual verification with a GuardRailProcessor using `writer.custom()` in `processOutputStream`

### DIFF
--- a/.changeset/brave-clouds-mate.md
+++ b/.changeset/brave-clouds-mate.md
@@ -2,4 +2,4 @@
 '@mastra/core': patch
 ---
 
-Fixed writer being undefined in processOutputStream when output processors run on data-\* chunks in the workflow loop stream. This ensures custom output processors (like guardrail processors) can emit stream events via writer.custom() in all execution paths.
+Fixed `writer` being undefined in `processOutputStream` for all output processors. The root cause was that `processPart` in `ProcessorRunner` did not pass the `writer` to `executeWorkflowAsProcessor` in the outputStream phase. Since all user processors are wrapped into workflows via `combineProcessorsIntoWorkflow`, this meant no processor ever received a `writer`. Custom output processors (like guardrail processors) can now reliably use `writer.custom()` to emit stream events.
--- a/.changeset/brave-clouds-mate.md
+++ b/.changeset/brave-clouds-mate.md
@@ -1,0 +1,5 @@
+---
+'@mastra/core': patch
+---
+
+Fixed writer being undefined in processOutputStream when output processors run on data-\* chunks in the workflow loop stream. This ensures custom output processors (like guardrail processors) can emit stream events via writer.custom() in all execution paths.
--- a/packages/core/src/agent/agent-writer-repro.test.ts
+++ b/packages/core/src/agent/agent-writer-repro.test.ts
@@ -1,0 +1,82 @@
+/**
+ * Reproduction test for writer being undefined in processOutputStream.
+ *
+ * Tests every chunk type that flows through processOutputStream to verify
+ * writer is always defined. This covers the full agent.stream() path
+ * (handleChatStream → agent.stream() → workflowLoopStream → MastraModelOutput).
+ */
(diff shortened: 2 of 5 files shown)


wardpeet (approved):

wardpeet (changes requested):
The coderabbit is actual correct, this makes it so data- tags are not stored so no replayed

PR #14171, opened 2026-03-11
Fix generateEmptyFromSchema nested object initialization

VasuBansal7576:
## Description

Fixes `generateEmptyFromSchema` so working memory schema initialization behaves correctly when the input schema is already parsed and when it contains nested object properties.

This change:
- accepts both stringified and pre-parsed JSON schema objects
- recursively initializes nested object properties instead of returning empty `{}` placeholders
- respects schema `default` values during initialization
- adds regression tests covering all three cases

## Related Issue(s)

Fixes #14152

## Type of Change

- [x] Test update

## Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed schema initialization to correctly handle pre-parsed JSON schema objects, with improved support for nested properties and default value handling.
* **Behavioral Improvements**
  * Memory handling now applies shared schema defaults for parsed templates, ensuring defaulted fields appear in generated content.
* **Public API**
  * generateEmptyFromSchema now accepts both schema strings and parsed schema objects.

### DIFF
--- a/.changeset/fix-generate-empty-from-schema-14152.md
+++ b/.changeset/fix-generate-empty-from-schema-14152.md
@@ -1,0 +1,5 @@
+---
+'@mastra/core': patch
+---
+
+Fixed `generateEmptyFromSchema` so working memory schema initialization handles pre-parsed JSON schema objects, nested object properties, and schema defaults correctly.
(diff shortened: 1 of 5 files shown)


wardpeet:
Should be fixed by https://github.com/mastra-ai/mastra/pull/14310

PR #14177, opened 2026-03-11
fix(core): stop dowhile loop for client tools without execute function

DanielSLew:
## Summary

When a tool created with `createTool()` has no `execute` function (intended for client-side execution), the agent's `dowhile` loop incorrectly continued calling the model repeatedly. This led to repeated model calls up to `maxSteps` instead of returning control to the caller after the first iteration.

## Root Cause

In `llm-execution-step.ts`, the `hasPendingToolCalls` check only verified that tool calls weren't provider-executed, but didn't check whether the resolved tool actually has an `execute` function. Client tools (created without `execute`) would always pass this check, keeping the loop alive.

## Changes

**`packages/core/src/loop/workflows/agentic-execution/llm-execution-step.ts`**
- Added `resolveTool` helper to look up tools by name or ID (matching existing tool resolution logic)
- Refined `hasPendingToolCalls` to also check `resolveTool(tc.toolName)?.execute` — tools without execute no longer count as pending
- Added `'tool-calls'` to the stop-reasons list as a secondary safety check

**`packages/core/src/agent/__tests__/tools.test.ts`** — 4 new tests:
- Tool without execute stops loop (stream mode)
- Tool without execute stops loop (generate mode)
- Mixed tools (one with execute, one without) stops loop
- Regression test for #14043: tool WITH execute and `finishReason: 'stop'` still continues the loop

## Verification

- TypeScript compiles clean (`tsc --noEmit`)
- All 28 tool tests pass (1 skipped)
- All 142 loop tests pass (no regressions)

Closes #14093

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed an infinite loop in agent execution when tools intended for client-only use lacked a server execute path; the agent now stops and returns control to the caller instead of repeatedly invoking the model.

* **Tests**
  * Added end-to-end tests covering client-only tool scenarios across streaming and generation flows, mixed tool configurations, and regression cases to ensure correct termination and single model invocation.

### DIFF
--- a/.changeset/swift-cloths-slide.md
+++ b/.changeset/swift-cloths-slide.md
@@ -1,0 +1,5 @@
+---
+'@mastra/core': patch
+---
+
+Fixed agent loop not stopping for client tools without an execute function. When a tool created with createTool() has no execute function (intended for client-side execution), the agent's do-while loop would incorrectly continue calling the model repeatedly. The loop now properly detects that client tools without execute should stop the loop and return control to the caller. Fixes #14093
--- a/.changeset/swift-cloths-slide.md
+++ b/.changeset/swift-cloths-slide.md
@@ -2,4 +2,4 @@
 '@mastra/core': patch
 ---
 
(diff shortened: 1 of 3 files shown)


LekoArts:

PR #14221, opened 2026-03-12
fix(schema-compat): replace CJS require with ESM import for zod/v4

intojhanurag:
## Description

Replace `require('zod/v4')` with a static ESM `import { toJSONSchema } from 'zod/v4'` in the zod-v4 standard-schema adapter. The bare `require()` call was being bundled by tsup into a `__require` shim that fails in pure ESM runtimes (where `require` is undefined), causing all Zod 4 schema conversions (structuredOutput, tool inputSchema, workflow step schemas) to throw `"z.toJSONSchema is not available"`. Removed the now-unnecessary lazy-loading cache (`_toJSONSchema`, `_toJSONSchemaResolved`, `getToJSONSchema()`).

## Related Issue(s)

Fixes #14217

## Type of Change
- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [ ] Test update

## Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed Zod 4 schema conversions failing in ESM projects. Structured outputs, tool input schemas, and workflow step schemas now work correctly when using Zod 4 in modules with `"type": "module"`.

### DIFF
--- a/.changeset/fix-zod4-esm-import.md
+++ b/.changeset/fix-zod4-esm-import.md
@@ -1,0 +1,5 @@
+---
+'@mastra/schema-compat': patch
+---
+
+Fixed Zod 4 schema conversions failing in ESM projects with the error `"z.toJSONSchema is not available. Ensure zod >= 3.25.0 is installed."` This affected structured outputs, tool input schemas, and workflow step schemas when using Zod 4 in a project with `"type": "module"`.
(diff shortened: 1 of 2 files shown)


LekoArts (changes requested):
I don't think this is the correct fix for this issue, the `require` is there on purpose to have a dynamic require that then fails in the try/catch and sets another value.

I've pinged @wardpeet about it to look over it

PR #14224, opened 2026-03-12
chore(docs): Re-organize agent & network docs

LekoArts:
## Description

 - overview.mdx — Simplified to match overview page template. Added "Multi-agent systems" section with supervisor pattern guidance and decision context.                  
 - using-tools.mdx — Added "Execution context" and "Control tool selection" sections. Removed duplicate content covered elsewhere.                                        
 - agent-memory.mdx — Added "Memory in multi-agent systems" section explaining isolation mechanics and how to share memory between agents. Added                          
 listThreads/getThreadById query example.                                                                                                                                 
 - processors.mdx — Restructured around user goals. Fixed all import paths against source code. Updated "Use workflows as processors" example with correct .map() step.   
 - guardrails.mdx — Removed duplicate content already in processors.mdx. Added "Speed up guardrails" section (parallelization, model choice, batching).                   
 - agent-approval.mdx — Condensed from 31 to 18 headings. Removed redundant sections, added "When to use" and "Quickstart".                                               
 - networks.mdx — Consolidated network-approval.mdx into this file. Reduced from ~590 to ~210 lines with a single deprecation banner.                                     
 - network-approval.mdx — Deleted (content merged into networks.mdx, redirect already exists).                                                                            
 - structured-output.mdx — Fixed gerund headings.                                                                                                                         
 - supervisor-agents.mdx — Minor link and terminology updates.

## Type of Change

- [X] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [ ] Test update

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Deprecated badge added to UI badges.
  * Processor interface can opt into handling data-* chunks.

* **Improvements**
  * Major docs rewrites: agents, memory, guardrails, processors, tools, workflows with clearer Quickstarts and examples.
  * Standardized headings (Quickstart) and updated model references across examples.

* **Updates**
  * Navigation and warnings de-emphasize agent networks in favor of supervisor agents; network-approval content removed.
  * Tab label shortened from "Guides & Migrations" to "Guides".


docs/src/content/en/docs/agents/agent-memory.mdx
```diff
@@ -139,11 +151,9 @@ const response = await memoryAgent.generate("What's my favorite color?", {
 Each thread has an owner (`resourceId`) that can't be changed after creation. Avoid reusing the same thread ID for threads with different owners, as this will cause errors when querying.
```
graysonhicks: Here would be useful to actually see how to query this.

docs/src/content/en/docs/agents/agent-memory.mdx
```diff
@@ -161,37 +171,58 @@ export const memoryAgent = new Agent({
...
-      },
-    },
-  }),
+:::
+
+## Memory in multi-agent systems
+
+When a [supervisor agent](/docs/agents/supervisor-agents) delegates to a subagent, Mastra scopes memory so each agent has its own conversation history. Understanding how this scoping works lets you decide what stays private and what gets shared.
+
+### Default behavior: isolation
```
graysonhicks: Really great section overall.  Not clear if `isolation` is an option you pass in or a pattern.  Or what the alternatives are (the sections after don't feel clearly connected)

docs/src/content/en/docs/agents/processors.mdx
```diff
@@ -199,39 +154,151 @@ export class DynamicModelProcessor implements Processor {
...
-- `providerOptions`: Modify provider options
-- `modelSettings`: Modify model settings
-- `structuredOutput`: Modify structured output configuration
-
-#### Ensuring a final response with `maxSteps`
+The method receives the current `stepNumber`, `model`, `tools`, `toolChoice`, `messages`, and more. Return an object with any properties you want to override for that step, for example `{ model, toolChoice, tools, systemMessages }`.
+
+See the [`Processor` reference](/reference/processors/processor-interface#processinputstep) for all available arguments and return types.
+
+### Use the `prepareStep` callback
```
DanielSLew: I dno't think this is clear that it just implements the processInputStep fn

docs/src/content/en/docs/agents/agent-memory.mdx
```diff
@@ -12,58 +12,45 @@ import StepItem from "@site/src/components/StepItem";
 
 # Agent memory
 
-Agents use memory to maintain context across interactions. LLMs are stateless and don't retain information between calls, so agents need memory to track message history and recall relevant information.
+Agents use memory to maintain context across interactions. Without memory, each call to an agent is independent, and the agent has no access to past messages.
```
bookercodes: Suggest - 

> Memory enables your agent to remember user messages, agent replies, and tool results across interactions, giving it the context it needs to stay consistent, maintain conversation flow, and produce better answers over time.

(More specific)

docs/src/content/en/docs/agents/supervisor-agents.mdx
```diff
@@ -120,8 +122,8 @@ The `context` object includes:
 
 Called after a delegation finishes. Use it to inspect results, provide feedback, or stop execution:
 
-- `context.bail()` — stop the supervisor loop immediately
-- Return `{ feedback: '...' }` — add feedback that gets saved to the supervisor's memory and is visible to subsequent iterations
+- `context.bail()`: Stop the supervisor loop immediately
+- Return `{ feedback: '...' }`: Add feedback that gets saved to the supervisor's memory and is visible to subsequent iterations
```
bookercodes, line 132: As a reader, I wondered "how do I access this?"

Iis there a reference doc we can link to?

docs/src/content/en/docs/agents/agent-memory.mdx
bookercodes: We have a a dedicated **Memory** section with information on this: https://mastra-docs-1x-git-docs-agent-network-rework-kepler-ab1bf5ab.vercel.app/docs/memory/message-history#accessing-memory

Perhaps we can link to it?

docs/src/content/en/docs/agents/agent-memory.mdx
```diff
@@ -139,11 +151,9 @@ const response = await memoryAgent.generate("What's my favorite color?", {
 Each thread has an owner (`resourceId`) that can't be changed after creation. Avoid reusing the same thread ID for threads with different owners, as this will cause errors when querying.
 :::
 
-To learn more about memory see the [Memory](../memory/overview) documentation.
-
 ## Observational memory
 
-For long-running conversations, raw message history grows until it fills the context window, degrading agent performance. [Observational Memory](../memory/observational-memory) solves this by running background agents that compress old messages into dense observations, keeping the context window small while preserving long-term memory.
+For long-running conversations, raw message history grows until it fills the context window, degrading agent performance. [Observational memory](../memory/observational-memory) solves this by running background agents that compress old messages into dense observations, keeping the context window small while preserving long-term memory.
```
bookercodes: Is the lowercase "m" intentional here? You use uppercase everywhere else 👀

docs/src/content/en/docs/agents/agent-memory.mdx
```diff
@@ -12,58 +12,45 @@ import StepItem from "@site/src/components/StepItem";
 
 # Agent memory
 
-Agents use memory to maintain context across interactions. LLMs are stateless and don't retain information between calls, so agents need memory to track message history and recall relevant information.
+Agents use memory to maintain context across interactions. Without memory, each call to an agent is independent, and the agent has no access to past messages.
 
-Mastra agents can be configured to store message history, with optional [working memory](../memory/working-memory) to maintain recent context, [semantic recall](../memory/semantic-recall) to retrieve past messages based on meaning, or [observational memory](../memory/observational-memory) for automatic long-term memory that compresses conversations as they grow.
+Memory stores message history. Mastra supports three additional memory modes:
```
bookercodes: I found the first sentence a bit confusing

Try:

> Mastra supports four complementary memory types:

And list them? For example: 


> 1. [Message history](https://mastra-docs-1x-git-docs-agent-network-rework-kepler-ab1bf5ab.vercel.app/docs/memory/message-history) - keeps recent messages from the current conversation so they can be rendered in the UI and used to maintain short-term continuity within the exchange.
> 2. [Observational memory](https://mastra-docs-1x-git-docs-agent-network-rework-kepler-ab1bf5ab.vercel.app/docs/memory/observational-memory) - uses background Observer and Reflector agents to maintain a dense observation log that replaces raw message history as it grows, keeping the context window small while preserving long-term memory across conversations.
> 3. [Working memory](https://mastra-docs-1x-git-docs-agent-network-rework-kepler-ab1bf5ab.vercel.app/docs/memory/working-memory) - stores persistent, structured user data such as names, preferences, and goals.
> 4. [Semantic recall](https://mastra-docs-1x-git-docs-agent-network-rework-kepler-ab1bf5ab.vercel.app/docs/memory/semantic-recall) - retrieves relevant messages from older conversations based on semantic meaning rather than exact keywords, mirroring how humans recall information by association. Requires a [vector database](https://mastra-docs-1x-git-docs-agent-network-rework-kepler-ab1bf5ab.vercel.app/docs/memory/semantic-recall#storage-configuration) and an [embedding model](https://mastra-docs-1x-git-docs-agent-network-rework-kepler-ab1bf5ab.vercel.app/docs/memory/semantic-recall#embedder-configuration).

docs/src/content/en/docs/agents/agent-memory.mdx
```diff
@@ -161,37 +171,58 @@ export const memoryAgent = new Agent({
...
+
+- **Thread ID**: Unique per delegation. The subagent starts with a clean message history every time it's called.
+- **Resource ID**: Derived as `{parentResourceId}-{agentName}`. Because the resource ID is stable across delegations, resource-scoped memory persists between calls. A subagent remembers facts from previous delegations by the same user.
+- **Memory instance**: If a subagent has no memory configured, it inherits the supervisor's `Memory` instance. If the subagent defines its own, that takes precedence.
+
+The supervisor forwards its conversation context to the subagent so it has enough background to complete the task. Only the delegation prompt and the subagent's response are saved — the full parent conversation is not stored. You can control which messages reach the subagent with the [`messageFilter`](/docs/agents/supervisor-agents) callback.
+
+### Share memory between agents
+
+Memory sharing in Mastra is controlled by two identifiers: `resourceId` and `threadId`. Agents that use the same values read and write to the same data. This is useful when agents collaborate on a shared context — for example, a researcher that saves notes and a writer that reads them.
```
bookercodes: Resource ID and thread ID are currently introduced in the message history section, but they apply to all memory types. Since they're foundational to how memory is scoped, it would make more sense to define them in their own section early in the doc, before any specific memory type is discussed.

docs/src/content/en/docs/agents/supervisor-agents.mdx
```diff
@@ -308,9 +310,7 @@ Success criteria:
 
 ## Related
```
bookercodes, line 323: Add a link to the **Memory** page and the anchor on multi-agent memory? Feels highly relevant

docs/src/content/en/docs/agents/agent-memory.mdx
```diff
@@ -161,37 +171,58 @@ export const memoryAgent = new Agent({
...
+
+### Default behavior: isolation
+
+Each delegation creates a fresh `threadId` and a deterministic `resourceId` for the subagent:
+
+- **Thread ID**: Unique per delegation. The subagent starts with a clean message history every time it's called.
+- **Resource ID**: Derived as `{parentResourceId}-{agentName}`. Because the resource ID is stable across delegations, resource-scoped memory persists between calls. A subagent remembers facts from previous delegations by the same user.
+- **Memory instance**: If a subagent has no memory configured, it inherits the supervisor's `Memory` instance. If the subagent defines its own, that takes precedence.
+
+The supervisor forwards its conversation context to the subagent so it has enough background to complete the task. Only the delegation prompt and the subagent's response are saved — the full parent conversation is not stored. You can control which messages reach the subagent with the [`messageFilter`](/docs/agents/supervisor-agents) callback.
```
bookercodes: Can we link somewhere more specific for `messageFilter`, such a a section on the page or a reference doc? As a reader, I clicked this link and I wasn't sure where to go next.

docs/src/content/en/docs/agents/agent-memory.mdx
```diff
@@ -161,37 +171,58 @@ export const memoryAgent = new Agent({
...
+
+### Share memory between agents
+
+Memory sharing in Mastra is controlled by two identifiers: `resourceId` and `threadId`. Agents that use the same values read and write to the same data. This is useful when agents collaborate on a shared context — for example, a researcher that saves notes and a writer that reads them.
+
+**Resource-scoped sharing** is the most common pattern. [Observational memory](../memory/observational-memory), [working memory](../memory/working-memory), and [semantic recall](../memory/semantic-recall) all default to `scope: 'resource'`. If two agents share a `resourceId`, they share observations, working memory, and embeddings — even across different threads.
```
bookercodes: Important - OM uses thread scope by default btw, resource scoping is in beta.

docs/src/content/en/docs/agents/agent-memory.mdx
bookercodes: This section seems _so_ specific to supervisor agent, it might make more sense to put in that page? 

We should at least add a link to this section to the Supervisor Agent page under "Resources"

It's so important, you might even consider a subheading that briefly describes the options before  linking here.

docs/src/content/en/docs/agents/overview.mdx
```diff
@@ -1,127 +1,47 @@
...
 
 import Tabs from "@theme/Tabs";
 import TabItem from "@theme/TabItem";
-import Steps from "@site/src/components/Steps";
-import StepItem from "@site/src/components/StepItem";
 
-# Using agents
+# Agents overview
 
 Agents use LLMs and tools to solve open-ended tasks. They reason about goals, decide which tools to use, retain conversation memory, and iterate internally until the model emits a final answer or an optional stop condition is met. Agents produce structured responses you can render in your UI or process programmatically. Use agents directly or compose them into workflows or agent networks.
```
bookercodes: o we still want to mention Networks at the end of the introduction? 👀 I think it's still good to capture the idea - perhaps the phrasing just needs updating.

docs/src/content/en/docs/agents/overview.mdx
```diff
@@ -1,127 +1,47 @@
...
+export const testAgent = new Agent({
+  id: 'test-agent',
+  name: 'Test Agent',
+  instructions: 'You are a helpful assistant.',
+  model: 'openai/gpt-5.4',
+})
+```
 
-:::
+Specify a unique `id` for your agent and provide a descriptive `name`.
```
bookercodes: This feels repetitive, since you already mentioned it at the top (then in the code).

docs/src/content/en/docs/agents/overview.mdx
```diff
@@ -1,127 +1,47 @@
...
-Visit [Agent reference](/reference/agents/agent) for more information.
+Watch an introduction to agents, and how they compare to workflows on [YouTube (7 minutes)](https://youtu.be/0jg2g3sNvgw).
 
 :::
 
-### Dynamic instructions
+## Quickstart
 
-Instructions can be provided as an async function, allowing you to resolve prompts at runtime. This enables patterns like personalizing instructions based on user context, fetching prompts from external registry services, and running A/B tests with different variants.
+Create an agent by instantiating the `Agent` class from `@mastra/core` and provide the required properties: `id`, `name`, `instructions`, and `model`.
```
bookercodes: Some of the prose explanations before code blocks feel redundant - the code is simple enough to speak for itself. Consider trimming or removing the lead-in text where the snippet already makes the behaviour obvious 👍🏻

docs/src/content/en/docs/agents/overview.mdx
```diff
@@ -1,127 +1,47 @@
...
+  instructions: 'You are a helpful assistant.',
+  model: 'openai/gpt-5.4',
+})
+```
 
-:::
+Specify a unique `id` for your agent and provide a descriptive `name`.
 
-### Registering an agent
+The `instructions` define the agent's behavior, personality, and capabilities. They're system-level prompts that establish the agent's core identity and expertise. The `model` is specified as `provider/model-name` using Mastra's [model router](/models).
```
bookercodes: We refer to these as "model router strings" elsewhere. This could be a good opportunity to introduce that terminology here so it feels more familiar when readers encounter it later.

Also, I'm not sure whether this should be formatted as `provider/model` or `"provider/model"` (with quotes) — I am leaning towards quotes.

docs/src/content/en/docs/agents/overview.mdx
```diff
@@ -132,85 +52,37 @@ export const mastra = new Mastra({
...
 
-### Registering subagents
-
-You can register subagents within an agent using the `agents` property. Adding subagents makes the parent agent a supervisor. A supervisor can delegate tasks to subagents, control delegations with hooks, filter context, and monitor progress using the [supervisor pattern](/docs/agents/supervisor-agents). Read the [using agents as tools](/docs/agents/using-tools#using-agents-as-tools) documentation for setup details.
-
-### Registering subworkflows
+Once registered, it can be called from workflows, tools, or other agents, and has access to shared resources such as memory, logging, and observability features.
 
-You can register subworkflows within an agent. They'll be executed as tools from the parent agent. Read the [using workflows as tools](/docs/agents/using-tools#using-workflows-as-tools) documentation to learn more.
+Use [Studio](/docs/getting-started/studio) to test your agent with different messages, inspect tool calls and responses, and debug agent behavior.
```
bookercodes, line 57: Good addition! IIt feels a bit out of left field - might be worth making it a `::: tip` so it doesn't catch the reader off guard.

docs/src/content/en/docs/agents/overview.mdx
```diff
@@ -132,85 +52,37 @@ export const mastra = new Mastra({
...
-`mastra.getAgent()` is preferred over a direct import, since it provides access to the Mastra instance configuration (logger, telemetry, storage, registered agents, and vector stores).
-
-::::
+:::
 
-## Generating responses
+## Use your agent
 
-Agents can return results in two ways: generating the full output before returning it or streaming tokens in real time. Choose the approach that fits your use case: generate for short, internal responses or debugging, and stream to deliver pixels to end users as quickly as possible.
+After registration, retrieve your agent with [`mastra.getAgentById()`](/reference/core/getAgentById). Call `.generate()` for a complete response or `.stream()` to deliver tokens in real time. You can call agents from workflow steps, tools, the Mastra Client, or the command line.
```
bookercodes: > You can call agents from workflow steps, tools, the Mastra Client, or the command line.

Should we also mention route handlers here? I think that's a pretty common use case worth calling out.

There's also an opportunity to link to the relevant docs - for example, how to use Mastra with a Next.js route handler. That way readers can easily dive deeper if needed.

docs/src/content/en/docs/agents/overview.mdx
```diff
@@ -219,130 +91,59 @@ Agents can return results in two ways: generating the full output before returni
...
-
-## Structured output
-
-Agents can return structured, type-safe data using Zod or JSON Schema. The parsed result is available on `response.object`.
-
-:::info
+:::note
 
-Visit [Structured Output](/docs/agents/structured-output) for more information.
+`mastra.getAgentById()` is preferred over a direct import, since it provides access to the Mastra instance configuration (logger, telemetry, storage, registered agents, etc.).
```
bookercodes: I think this undersells it a bit - this is a highly recommended pattern. Rather than framing it as a side note in an admonition, it would read better as just "the way you do it" in the main body of the text. That way it comes across as the default approach rather than a suggestion.

(Appreciate that this has been here a while and it's not something you introduced!)

docs/src/content/en/docs/agents/overview.mdx
```diff
@@ -219,130 +91,59 @@ Agents can return results in two ways: generating the full output before returni
...
 
-`onStepFinish` is only available when streaming or generating text without structured output.
+| Goal | Start here |
+| - | - |
+| Give your agent tools to call external APIs or services | [Tools](/docs/agents/using-tools) |
+| Keep context and preferences across conversations | [Memory](/docs/agents/agent-memory) |
+| Get typed objects back instead of plain text | [Structured output](/docs/agents/structured-output) |
+| Intercept or transform messages before and after generation | [Processors](/docs/agents/processors) |
+| Swap instructions or models based on request context | [Dynamic configuration](/docs/server/request-context) |
+| Have one agent delegate tasks to other agents | [Supervisor agents](/docs/agents/supervisor-agents) |
```
bookercodes: The goal here is probably better described as building a multi-agent network . (See my comment below on "human-in-the-loop" for rationale.)

docs/src/content/en/docs/agents/overview.mdx
```diff
@@ -219,130 +91,59 @@ Agents can return results in two ways: generating the full output before returni
...
+| Register subagents | [Tools](/docs/agents/using-tools#agents-as-tools) |
```
bookercodes: To me, this reads more like a task than a goal - a goal should describe a desired outcome. 

Since a sub-agent is a type of multi-agent system, if we added a new top-level page about multi-agent networks, we could consolidate two of these goals into one? That would reduce duplication while also making the distinction clearer.

docs/src/content/en/docs/agents/overview.mdx
```diff
@@ -219,130 +91,59 @@ Agents can return results in two ways: generating the full output before returni
...
+| Validate what goes in or comes out of an agent | [Guardrails](/docs/agents/guardrails) |
```
bookercodes: Consider rephrasing this to something like "keep your agent safe". Tying it to a tangible value makes it more meaningful to the reader.

docs/src/content/en/docs/agents/overview.mdx
```diff
@@ -219,130 +91,59 @@ Agents can return results in two ways: generating the full output before returni
...
+| Pause execution and wait for human approval | [Approval](/docs/agents/agent-approval) |
```
bookercodes: "Human-in-the-loop" is the keyword here - please make sure to include it 👍🏻

It's important that we signpost things based on how readers think about the concept, not how we name the feature internally.

Along the same lines, I'd like to consider renaming "Agent Approval" to "Human-in-the-loop" in the sidebar. We already use "human-in-the-loop" in the Workflows section (even though we could technically call it "suspend and resume" or something similar). Why should the treatment be different here?

docs/src/content/en/docs/agents/overview.mdx
```diff
@@ -219,130 +91,59 @@ Agents can return results in two ways: generating the full output before returni
...
-export const testAgent = new Agent({
-  id: 'test-agent',
-  name: 'Test Agent',
-  tools: { testTool },
-})
-```
+## Multi-agent systems
 
-:::info
+A multi-agent system uses multiple agents to solve a task that is too broad or too specialized for a single agent. Instead of building one agent with dozens of tools and a long instruction set, you split responsibilities across focused agents and let a coordinator bring results together.
```
bookercodes, line 117: This section feels *very* similar to the **Supervisor Agent** page. It makes me wonder whether they should live on the same page, or at the very least share a natural intro that then branches off into each approach.

To be clear, duplication isn't inherently bad - I think it's fine to repeat information elsewhere in the docs if it's surfaced at the right time. In this case though, consolidating could create a clearer path for the reader and make the differences between the two approaches more obvious upfront.

docs/src/content/en/docs/agents/overview.mdx
```diff
@@ -219,130 +91,59 @@ Agents can return results in two ways: generating the full output before returni
...
-  'user-tier': 'enterprise' | 'pro'
-}
+### Supervisor agents
 
-export const testAgent = new Agent({
-  id: 'test-agent',
-  name: 'Test Agent',
-  model: ({ requestContext }) => {
-    const userTier = requestContext.get('user-tier') as UserTier['user-tier']
+[Supervisor agents](/docs/agents/supervisor-agents) are the recommended way to build multi-agent systems in Mastra. A supervisor agent owns a set of subagents. During execution, the supervisor's LLM decides when to delegate to a subagent, what prompt to send, and how to combine results. Subagents are automatically exposed as callable tools.
```
bookercodes: I don't think this is quite accurate - a multi-agent system doesn't necessarily require a Supervisor. For example, a workflow where each step uses its own sub-agent is a very common pattern, as is an agent that calls other agents as tools. Both would qualify as multi-agent systems.

That said, I'm curious - do you not consider those to be multi-agent systems? I am interested  to hear your thinking here!

docs/src/content/en/docs/agents/overview.mdx
```diff
@@ -219,130 +91,59 @@ Agents can return results in two ways: generating the full output before returni
...
 
-See [Request Context](/docs/server/request-context) for more information.
-
-:::
+You can control delegation behavior with hooks (`onDelegationStart`, `onDelegationComplete`), filter which messages are forwarded (`messageFilter`), and score task completion (`taskCompletionCriteria`). See [Supervisor agents](/docs/agents/supervisor-agents) for a full walkthrough with code examples.
 
 :::tip
 
-For type-safe request context schema validation, see [Schema Validation](/docs/server/request-context#schema-validation).
+Follow the [Build a research coordinator with supervisor agents](/guides/guide/research-coordinator) guide for a step-by-step walkthrough.
```
bookercodes: This tip is visually dominant and draws the eye away from the link to the Supervisor Agents page, which is more important.

docs/src/content/en/docs/agents/overview.mdx
bookercodes: For what it's worth, I don't fully understand why you need to use ⁠getAgent instead of importing it directly. Kristian (our YouTube partner) asked me the same question once, so I suspect other readers will wonder too. Might be out of scope here, but it could be worth addressing somewhere in the docs.

docs/src/content/en/docs/agents/agent-memory.mdx
bookercodes: We have a section on this under **Memory** (top-level page).

docs/src/content/en/docs/agents/overview.mdx
bookercodes: The table order looks generally correct - definitely agree with tools at the top and voice at the bottom. A couple of suggestions:

* I'd group **Processors** and **Guardrails** together and move them towards the bottom, since they're more advanced and a less likely path for someone reading the introduction.
* Maybe move **Human-in-the-loop** up? It feels like a more common concern for readers at this stage.

docs/src/content/en/docs/agents/agent-memory.mdx
LekoArts: I use the lowercase everywhere 🤔

docs/src/content/en/docs/agents/overview.mdx
LekoArts: We do not use "model router strings" anywhere in the docs, we use "Model router"

docs/src/content/en/docs/agents/overview.mdx
LekoArts: > a workflow where each step uses its own sub-agent is a very common pattern, as is an agent that calls other agents as tools. Both would qualify as multi-agent systems.

I don't think a workflow qualifies as what most people understand under this term as it's "too deterministic". I understand it as multiple agents autonomously working on something together, not that if you use 2 agents to solve a problem you then have a multi-agent system.

But we don't say require? We recommend using supervisor agents

DanielSLew (approved):

docs/src/content/en/docs/agents/agent-memory.mdx
```diff
@@ -12,58 +12,46 @@ import StepItem from "@site/src/components/StepItem";
...
 
-Mastra agents can be configured to store message history, with optional [working memory](../memory/working-memory) to maintain recent context, [semantic recall](../memory/semantic-recall) to retrieve past messages based on meaning, or [observational memory](../memory/observational-memory) for automatic long-term memory that compresses conversations as they grow.
+Mastra supports four complementary memory types:
+
+- [Message history](/docs/memory/message-history) keeps recent messages from the current conversation so they can be rendered in the UI and used to maintain short-term context.
+- [Observational memory](../memory/observational-memory) uses background agents (Observer and Reflector) to maintain a dense observation log that replaces raw message history as it grows. This keeps the context window small while preserving long-term memory.
+- [Working memory](../memory/working-memory) stores persistent, structured user data such as names, preferences, and goals.
+- [Semantic recall](../memory/semantic-recall) retrieves relevant past messages on semantic meaning rather than exact keywords.
```
TylerBarnes: I kinda preferred this before the change, it was shorter, and we don't actually necessarily want to tell people to use them all ("complementary")

docs/src/content/en/docs/agents/agent-memory.mdx
TylerBarnes: it should probably be uppercase since "Observational Memory" is the name of it

docs/src/content/en/docs/agents/processors.mdx
LekoArts: Good callout, added a note for that 👍

PR #14232, opened 2026-03-12
fix(core): extend TrailingAssistantGuard to all request types for Claude 4.6

saddlepaddle:
## Summary

- Claude 4.6 rejects assistant message prefill in **all** request types, not just structured output. The existing `TrailingAssistantGuard` only fired when `structuredOutput.schema` was set, leaving normal streaming/generate calls vulnerable to the API error: *"This model does not support assistant message prefill. The conversation must end with a user message."*
- This extends the guard to append a synthetic user message whenever the last message is an assistant message, regardless of structured output. Uses `"Continue."` for plain requests and `"Generate the structured response."` for structured output (unchanged).
- Adds a regression test for the non-structured-output case.

## Context

The original `TrailingAssistantGuard` (PR #12835) was scoped to structured output because that was the initially reported scenario. However, Claude Opus 4.6 enforces the no-prefill restriction universally. This surfaces in practice during the agentic loop — e.g., after tool approval/resume, when memory loads a conversation ending with an assistant message, or when the network routing agent constructs an assistant-role prompt.

## Test plan

- [x] Existing structured output tests pass (no behavioral change for that path)
- [ ] New test: `should not send prompt ending with assistant message even WITHOUT structuredOutput (plain streaming)` — verifies the guard fires for non-structured-output cases with Claude 4.6

## Summary by CodeRabbit

* **Bug Fixes**
  * Trailing assistant-message guard now applies universally across all request types (including plain streaming), ensuring assistant-prefill messages are consistently rejected and prompts end in a user message when required.

* **Tests**
  * Added and updated test coverage to verify assistant message prefill rejection in plain streaming and other scenarios, reinforcing the guard's behavior.

### DIFF
--- a/packages/core/src/processors/trailing-assistant-guard.ts
+++ b/packages/core/src/processors/trailing-assistant-guard.ts
@@ -38,12 +38,12 @@ export function isMaybeClaude46(
 }
 
 /**
- * Guards against trailing assistant messages when using native structured output
- * with Anthropic Claude 4.6.
+ * Guards against trailing assistant messages with Anthropic Claude 4.6.
  *
- * Claude 4.6 rejects requests where the last message is an assistant message when
- * using output format (structured output), interpreting it as pre-filling the response.
- * This processor appends a user message to prevent that error.
+ * Claude 4.6 rejects requests where the last message is an assistant message,
+ * interpreting it as pre-filling the response. This applies to all request types,
+ * not just structured output. This processor appends a user message to prevent
+ * that error.
  *
  * This processor should only be added when the agent uses a Claude 4.6 model.
  * Use {@link isMaybeClaude46} to check before adding.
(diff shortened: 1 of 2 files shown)


packages/core/src/processors/trailing-assistant-guard.ts
```diff
@@ -71,7 +68,7 @@ export class TrailingAssistantGuard implements Processor<'trailing-assistant-gua
           role: 'user' as const,
           content: {
             format: 2 as const,
-            parts: [{ type: 'text' as const, text: 'Generate the structured response.' }],
+            parts: [{ type: 'text' as const, text: promptText }],
```
TylerBarnes, line 71: I hadn't seen this processor before, but this looks like the wrong way to fix it. I think this will mess with prompt caching and likely the way the agent responds because after each assistant step there will be a temporary user message saying continue.

Better fix imo is to catch when anthropic rejects the request and add 1 user message onto the end and retry, rather than on every step.

TylerBarnes:
superseded by https://github.com/mastra-ai/mastra/pull/14435

PR #14237, opened 2026-03-12
feat(memory): enforce resourceId isolation across SQL storage drivers

Sigmabrogz:
This implements Phase 2 of the memory resource scoping, enforcing resourceId isolation across Postgres and LibSQL storage drivers. It references PR #1 which introduced the interface changes.

ELI5: The PR makes thread lookups respect a "resource" label so a thread from one tenant/resource can't be read by another. It updates thread-getting methods across storage drivers and adds tests to ensure the isolation works.

* **What changed**
  * Enforce resourceId scoping for memory/thread retrieval across SQL drivers (Postgres, LibSQL) and all memory-backed storage adapters by extending getThreadById to accept an optional resourceId and return null when the stored thread’s resourceId does not match.
  * Updated the abstract MemoryStorage API (packages/core) and in-memory implementation to include the optional resourceId parameter.
  * Applied the same optional resourceId check in adapters for: Postgres, LibSQL, ClickHouse, Cloudflare (D1, DO, KV), Convex, DynamoDB, Lance, MongoDB, MSSQL, Redis, Upstash and related memory wrappers.
  * Updated packages/memory to forward resourceId in Memory.getThreadById and adjusted the wrapper signature accordingly.

* **Tests**
  * Moved/added resourceId isolation tests into the shared threads test suite (stores/_test-utils/src/domains/memory/threads.ts) verifying: matching resource returns the thread, mismatched resource returns null, and omission of resourceId remains backward-compatible.

* **Documentation / Release**
  * Added changeset (.changeset/memory-isolation-phase-2.md) documenting the change and bumping affected packages.

* **Notes for reviewers**
  * Public API: MemoryStorage.getThreadById signature changed to accept optional resourceId — adapter implementations were updated accordingly.
  * Behaviour: When resourceId is provided, lookups are now resource-scoped and will return null for non-matching resourceIds; when omitted, existing behavior is preserved.


TylerBarnes:
@coderabbitai full review

TylerBarnes:
@coderabbitai how are we looking now?

TylerBarnes (approved):

PR #14243, opened 2026-03-12
fix(core,memory): fix provider-executed tool persistence for deferred execution

CalebBarnes:
## Problem

Anthropic non-deterministically defers provider tool execution to subsequent API calls — usually when called in parallel with client tools. When this happens, our code had a few issues:

- `llm-mapping-step` created a **separate message** for the provider tool result (splitting call/result across messages)
- `output-converter` then stripped the result when building the next API request, assuming provider tools should always be filtered out
- This left an orphaned `server_tool_use` without its `web_search_tool_result`, causing the Anthropic 400 error: `web_search tool use with id ... was found without a corresponding web_search_tool_result block`
- Even when it didn't error, the persisted result was a stub `{ providerExecuted: true }` instead of real search data

### `llm-execution-step.ts`
- **Same-turn execution**: when a provider-executed tool call has its result in the same stream, store it directly as `state:'result'` (call + result in one part, one message)
- **Deferred execution**: when a deferred result arrives in a later turn (orphaned `tool-result` chunk with no matching `tool-call` in that turn's stream), find the existing `state:'call'` part in the messageList and update it in-place to `state:'result'` — no new message created

### `llm-mapping-step.ts`
- Remove the code that created separate messages for provider-executed tool results — no longer needed since `llm-execution-step` handles both scenarios

### `output-converter.ts`
- Keep `input-available + providerExecuted` parts (deferred calls waiting for results)
- Keep `output-available + providerExecuted` parts (completed results)

### `to-prompt.ts`
- Allow `tool-result` parts on assistant messages — AI SDK's `convertToModelMessages` correctly places both the `server_tool_use` call and its result on the assistant message for provider-executed tools. Our `to-prompt.ts` was rejecting this.

### `observational-memory.ts`
- Exclude messages with pending `state:'call'` tool invocations from buffering batches
- Skip sealing the newest message during sync observation if it has pending calls
- This ensures OM observes tool results with complete data rather than sealing messages before the deferred result arrives

## Test

Recorded the exact deferred API response (2 API calls — first returns the deferred `web_search` + client `tool_use`, second returns the `web_search_tool_result` after client tool execution) and wrote a replay test that exercises the full multi-step agentic loop. It verifies:
- Real result data is persisted (not a stub)
- State is updated from `call` → `result`
- Call and result stay in the same message (no split)

**Fails on main, passes on this branch.**

All existing tests pass (284 OM tests, 294 message-list tests, 230 loop tests, 132 agent e2e tests, 11 provider-tools e2e tests).


packages/memory/src/processors/observational-memory/observational-memory.ts
```diff
@@ -4381,6 +4142,18 @@ ${formattedMessages}
       `[OM:bufferCursor] cursor=${bufferCursor?.toISOString() ?? 'null'}, unobserved=${unobservedMessages.length}, afterExcludeBuffered=${preFilterCount}, afterCursorFilter=${candidateMessages.length}`,
     );
 
+    // Exclude messages with pending tool calls (state: 'call').
+    // Provider-executed tools (e.g. Anthropic web_search) may have their result
+    // arrive in a later turn. If we buffer/seal now, the result would be added
+    // via in-place mutation but OM would have already observed the message with
+    // the incomplete call — the completed result would never be observed.
+    candidateMessages = candidateMessages.filter(msg => {
```
TylerBarnes: this proves it but I don't think the fix is quite right. we should either make a way to split behind the call, or defer until the next step so it splits after the result

TylerBarnes (changes requested):

packages/memory/src/processors/observational-memory/observational-memory.ts
CalebBarnes: Yeah that makes sense. Originally I thought of maybe splitting the parts in the message but that might've caused more issues later on like reordering the tool calls/cache busting. It probably makes the most sense to just observe the candidate messages up to the point where theres a pending tool call and just cut it off there for this step.

PR #14246, opened 2026-03-12
fix(core): rehydrate suspended tool state data parts

### LINKED ISSUE #14216: data-tool-call-suspended part getting removed when saved to memory / storage
This issue was created from Discord post 1481644842365878292:

[Open in Discord](https://discord.com/channels/1309558646228779139/1481644842365878292)

I followed the NextJS guide for setting up Mastra. I added memory + storage for storing the conversation history.

I built a simple chat bot UI (via AI SDK UI guide) for showing the chat history and sending messages.

I attached images where the one with data-tool-call-suspended is after calling the tool. Then the other image is when I refreshed after the tool is called.

---

Screenshot 1

Screenshot 2

roaminro:
Fixes #14216.

Restores suspended and approval tool state data parts when AI SDK v5 messages are rebuilt from stored metadata after reload. This keeps the UI state consistent with what was streamed before persistence.

Before:
```ts
AIV5Adapter.toUIMessage(dbMessage).parts
// => [tool-process-data]
```

After:
```ts
AIV5Adapter.toUIMessage(dbMessage).parts
// => [tool-process-data, data-tool-call-suspended]
```

Also adds a focused adapter test covering suspended and approval state rehydration.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed AI SDK v5 message rehydration so suspended and pending-approval tool states are restored from persisted message metadata after reload.
  * Prevent duplicate suspended/approval indicators on messages and suppress them when tools have been resumed.


NikAiyer (approved):
## Review: fix(core): rehydrate suspended tool state data parts

**Overall: Approve** — Clean, focused bugfix with good test coverage.

### What this does
When messages are persisted to DB then reloaded via `AIV5Adapter.toUIMessage()`, the `data-tool-call-suspended` and `data-tool-call-approval` data parts were lost because they were only streamed to the client and never serialized into `content.parts`. The metadata (`suspendedTools` / `pendingToolApprovals`) *was* persisted though, so this PR reconstructs those data parts from metadata during rehydration.

### What I liked
- **Defensive programming**: Thorough null/type guards at every level (`typeof suspendedTool === 'object'`, `'toolCallId' in ...`, etc.). This is appropriate since the metadata shape isn't statically enforced.
- **Deduplication via `existingToolStateDataPartIds`**: Correctly prevents duplicates when the parts *do* happen to already exist in `content.parts` (they fall through the catch-all `else` branch at line 295 of the original code).
- **Good test coverage**: 5 test cases covering the happy paths (suspended, approval), dedup for both types, and the negative case (already-resumed tools with empty metadata).

### Minor observations (non-blocking)

1. **Duplicate code structure**: The `suspendedTools` and `pendingToolApprovals` loops are nearly identical — only the metadata key name and the part `type` string differ. A small helper like `rehydrateToolStateParts(metadata, key, partType, existingIds, parts)` could DRY this up. Not a blocker — the current form is clear and easy to follow.

2. **`as` casts on push**: The `as AIV5Type.DataUIPart<AIV5.UIDataTypes>` casts are necessary since the data shape isn't typed, but worth noting that if the `DataUIPart` generic changes upstream, these casts will silently accept mismatched shapes. The tests mitigate this risk well.

3. **Ordering**: Rehydrated data parts are appended after all other parts. During streaming, `data-tool-call-suspended` typically appears right after its corresponding `tool-*` part. If any consumer relies on part ordering (e.g., rendering suspended state adjacent to its tool invocation), the current append-at-end behavior could produce a slightly different part order on reload vs. live stream. Unlikely to matter in practice, but worth keeping in mind.

LGTM — ship it.

PR #13860, opened 2026-03-05
fix(core): persist provider-executed tool calls in stream order

TheIsrael1:
## Summary
- persist provider-executed `tool-call` parts at stream time so memory ordering matches stream ordering
- hydrate tool call args/output from matching `tool-result` payloads when available
- avoid double-persisting tool calls already written during streaming
- add regression coverage for provider-executed tool ordering + result persistence
- include a patch changeset for `@mastra/core`

## Verification
- `pnpm --filter @mastra/core run build:lib`
- `pnpm --filter @mastra/core test -- src/loop/loop.test.ts -t "providerExecuted tools should not be re-executed"`
- `pnpm --filter @mastra/core test -- src/loop/loop.test.ts -t "should persist provider-executed tool calls in stream order with results"`
- `pnpm --filter @mastra/core test -- src/loop/loop.test.ts`

Fixes #13762

## Summary by CodeRabbit

* **Bug Fixes**
  * Memory replay now correctly persists provider-executed tool calls with their results and in the correct streaming order.

* **Tests**
  * Added a test verifying provider-executed tool calls and their results are persisted in stream order.

* **Chores**
  * Added a changelog entry for the patch release documenting the fix.

PR #14282, opened 2026-03-13
fix(core): support deferred provider-executed tool calls & fix tool call persistence

CalebBarnes:
## Problem

When Anthropic non-deterministically defers provider tool execution (e.g. `web_search` alongside a client tool), several issues arose:

1. **Tool call part ordering** — provider-executed tool calls were batched at the end of messages instead of being persisted in stream order (`text → tool-call → text`), breaking the expected part ordering
2. **Output converter stripping provider tool results** — `output-converter.ts` was explicitly filtering out completed provider-executed tools (`if (p.providerExecuted) return false`), meaning `web_search` results were actively discarded when building outbound prompts. This was the root cause of lost provider tool data.
3. **Split messages** — if Observational Memory observed a message while a tool was still `state:'call'`, the deferred result couldn't be merged back, causing Anthropic 400 errors on subsequent requests
4. **Missing provider tool results** — deferred `web_search` results could be lost or persisted as stubs instead of real data

### New `updateToolInvocation` method (`message-list.ts`)
- Added `MessageList.updateToolInvocation()` — walks backwards through messages to find a tool-invocation part by `toolCallId` and replaces it with the updated state (e.g. `call → result`), preserving the original `args`. Moves the message to the response source so it gets re-saved.
- Dedicated test suite in `update-tool-invocation.test.ts`

### Stream-order tool call persistence (`llm-execution-step.ts`)
- Added eager `case 'tool-call'` in `processOutputStream` to add tool-invocation parts to the message list inline as they arrive in the stream, preserving `text → tool-call → text` ordering
- Removed `tool-call` from the text-flush exclusion so pending text deltas are flushed before a tool-call part is added
- Added `providerExecuted` and `providerMetadata` to `case 'tool-result'` so provider tool metadata is preserved through the `call → result` transition
- Removed the old post-stream batch block that added tool calls after stream processing
- Removed the `deferredProviderResults` block (from PR #13860) since the inline `case 'tool-result'` already handles deferred results as they arrive in the stream

### Refactored tool result persistence (`llm-mapping-step.ts`)
- Replaced `messageList.add()` with `messageList.updateToolInvocation()` for both error and success tool results — tool results now update the existing `state:'call'` part in-place instead of creating separate messages
- Added `!toolCall.providerExecuted` guard to the error results filter and success path so provider tools are skipped (handled by `llm-execution-step` stream path)
- Removed split client/provider result handling — both now flow through the same `updateToolInvocation` path
- Removed `MastraDBMessage` import (no longer constructing full messages)
- Updated all tests in `llm-mapping-step.test.ts` to expect `updateToolInvocation` calls instead of `add` calls

### Provider tool handling cleanup (`tool-call-step.ts`)
- Removed dead `result: inputData.output` mapping for provider tools (provider results come via `tool-result` chunks, not on `tool-call`)

### Output converter fix (`output-converter.ts`)
- **Removed the filter that was stripping completed provider-executed tools** — the old code had `if (p.providerExecuted) return false` which actively discarded `web_search` results from outbound prompts
- Now keeps completed provider tools (`output-available`/`output-error`) so Anthropic sees `server_tool_use` + `tool_result` blocks on follow-up turns
- Still keeps deferred provider tools (`input-available` + `providerExecuted`) so the provider API can execute them on the next request
- Still filters incomplete client-side tool calls

### Prompt conversion fix (`to-prompt.ts`)
- Removed `assistant` from the `tool-result` role incompatibility check — assistant messages can contain `tool-result` parts when provider-executed tools return results inline

### OM observation deferral (`observational-memory.ts`)
- Added `hasIncompleteToolCalls` gate at the three observation decision points in `process()` to defer observation when the latest step of the last message has a pending tool call (`state:'call'`)
- Added `getLatestStepParts` helper to scope the check to only the current step (avoids false positives from earlier steps)

## Tests

- **Unit test** (`loop/test-utils/tools.ts`): verifies `text → tool-call → text` stream order is preserved in persisted message parts
- **E2E test** (`provider-tools-persistence.e2e.test.ts`): two recording-backed tests using real Anthropic `web_search`:
  - Deferred `web_search` alongside a client tool — verifies real data persistence and correct part order
  - Single-turn `web_search` (no client tools) — verifies non-deferred persistence and part order
- **OM tests** (`observational-memory.test.ts`): async buffering deferral and sync observation deferral when provider tool calls are pending
- **Update tool invocation tests** (`update-tool-invocation.test.ts`): dedicated test suite for `messageList.updateToolInvocation`
- **Snapshot update** (`options.ts`): updated `prepareStep` snapshot to reflect leaner message shape from inline tool call processing
- **Mapping step tests** (`llm-mapping-step.test.ts`): updated all assertions from `messageList.add` to `messageList.updateToolInvocation`

## Summary by CodeRabbit

* **New Features**
  * In-place updates of existing tool invocations so tool results update the original message instead of creating duplicates.
  * Improved streaming handling so tool-call and tool-result events are reflected immediately and in correct order.
  * Per-call override to control recorder/mock test mode.

* **Bug Fixes**
  * Preserve completed provider-executed tool results and retain deferred provider-executed parts for the next response.
  * Prevent split/orphaned tool calls by persisting parts in actual stream order.
  * Defer observation/buffering while pending tool calls exist.

* **Tests**
  * Expanded coverage for update flows, provider-executed behavior, ordering, persistence, and edge cases.


packages/core/src/agent/message-list/message-list.ts
```diff
@@ -626,6 +627,50 @@ export class MessageList {
...
 
+  /**
+   * Replace a tool-invocation part matching the given toolCallId with the
+   * provided result part. Walks backwards through messages to find the match.
+   * If the message was already persisted (e.g. as a memory message), it is
+   * moved to the response source so it will be re-saved.
+   *
+   * @returns true if the tool call was found and updated, false otherwise.
+   */
+  public updateToolInvocation(inputPart: Extract<MastraMessagePart, { type: 'tool-invocation' }>): boolean {
```
DanielSLew, line 638: this probably makes more sense in the stateManager

TylerBarnes: I'm not sure cause it's updating the actual message list message values

TylerBarnes: it's like a specialized .add() just for tool invocation results

TylerBarnes (approved):

packages/core/src/agent/message-list/message-list.ts
CalebBarnes: Yeah its actually mutating the tool invocation in place. Just removing & adding in state manager to update state from memory -> response for deferred provider tool calls.

NikAiyer (approved):

PR #14297, opened 2026-03-14
fix(pg): use timezone-aware columns for thread/message ordering and filtering

abhiaiyer91:
## Summary

Fixes thread `updatedAt` being set to `updatedAtZ` value after the second message, causing incorrect ordering in `listThreads()` with PostgresStore.

## Root Cause

The PostgreSQL store had TIMESTAMP (`createdAt`, `updatedAt`) and TIMESTAMPTZ (`createdAtZ`, `updatedAtZ`) columns, but ORDER BY and date range queries used the TIMESTAMP columns. When the PG server timezone differs from UTC, this produces incorrect ordering. Additionally, ISO strings were passed to UPDATE queries for TIMESTAMP columns, stripping timezone info.

## Changes

- **`listThreads`**: `ORDER BY COALESCE("fieldZ", "field")` to prefer timezone-aware columns
- **`listMessages`**: Same COALESCE fix for message ordering  
- **`listMessagesByResourceId`**: COALESCE fix for both ORDER BY and date range filtering conditions
- **`saveMessages` / `updateThread`**: Pass `Date` objects instead of ISO strings for `updatedAt`/`updatedAtZ` to let pg-promise handle timezone conversion correctly

## Test Plan

- Verified existing PG memory tests still pass
- The fix ensures timezone-aware columns are used for sorting, which correctly handles UTC/local timezone differences

Closes #14120

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed timezone handling in database operations to ensure correct sorting and filtering of records when the server timezone differs from UTC. Enhanced timestamp storage and retrieval mechanisms to properly preserve timezone information across all database interactions, improving accuracy of date-range filtering, record ordering, and other critical query operations.


wardpeet (approved):

packages/playground-ui/src/domains/agents/hooks/use-agent-trace-scores.ts
```diff
@@ -24,7 +24,7 @@ export function useAgentTraceScores({ agentId, scorerId, enabled }: UseAgentTrac
       let page = 0;
       const scores: ScoreRecord[] = [];
 
-      // eslint-disable-next-line no-constant-condition
+
```
wardpeet:
```suggestion
      // eslint-disable-next-line no-constant-condition
```

PR #14298, opened 2026-03-14
fix(core): prevent merging consecutive tool-only turns from different steps

### LINKED ISSUE #14124: [BUG] Consecutive tool-only turns are incorrectly merged into one
### Describe the Bug

When an agent performs two consecutive turns with only tool calls (no output text, thinking) those tool calls are merged into a single turn instead of being kept separate.

This leads the agent to the false assumption that it ran all the tools in parallel, which can affect its output in subsequent messages.

For example, here the agent thinks it ran the tools in parallel:

Image

But it didn't:

Image

You can see here why it is getting it wrong - mastra has merged the two turns into one when it sends subsequent LLM calls:

```json
{
      "role": "assistant",
      "content": [
        {
          "type": "tool-call",
          "toolCallId": "toolu_01NtUVX3zKx3DKpZLxZVdNPM",
          "toolName": "weatherTool",
          "input": {
            "location": "London"
          }
        },
        {
          "type": "tool-call",
          "toolCallId": "toolu_012a27zjYRaLbugehnFGmVM6",
          "toolName": "weatherTool",
          "input": {
            "location": "Paris"
          }
        }
      ]
    },
    {
      "role": "tool",
      "content": [
        {
          "type": "tool-result",
          "toolCallId": "toolu_01NtUVX3zKx3DKpZLxZVdNPM",
          "toolName": "weatherTool",
          "output": {
            "type": "json",
            "value": {
              "temperature": 10,
              "feelsLike": 7.4,
              "humidity": 95,
              "windSpeed": 16.2,
              "windGust": 34.9,
              "conditions": "Light drizzle",
              "location": "London"
            }
          },
          "input": "[Circular]"
        },
        {
          "type": "tool-result",
          "toolCallId": "toolu_012a27zjYRaLbugehnFGmVM6",
          "toolName": "weatherTool",
          "output": {
            "type": "json",
            "value": {
              "temperature": 11,
              "feelsLike": 8.2,
              "humidity": 81,
              "windSpeed": 15,
              "windGust": 32.4,
              "conditions": "Overcast",
              "location": "Paris"
            }
          },
          "input": "[Circular]"
        }
      ]
    },
    {
      "role": "user",
      "content": [
        {
          "type": "text",
          "text": "Did you run them in separate turns or in parallel?"
        }
      ]
    }
```

It may not be immediately obvious why this is so problematic. Here's a motivating example: I had an agent with two tools - query data, and reply to user. It would query the data, then reply as two separate turns. Then it would incorrectly see it did the query and reply in parallel, and reply again "correcting" itself with the actual updated (but actually identical) data.

### Steps To Reproduce

Create a new project.

Send the weather agent this message: "Do not say anything. Just run your tool once, see the result but do not say anything, and then run the tool again"

Then ask it if it ran the tools in parallel - it will answer yes

However, check the actual LLM calls using observability: It ran them as separate turns

Reproducible with default-project weather agent

### Expected Behavior

Mastra should leave the two assistant turns as separate turns, instead of condensing them into one.

### Environment Information

```block
Claude sonnet 4.6 (but any will do)
```

### Verification

abhiaiyer91:
## Summary

Fixes consecutive tool-only turns being incorrectly merged into a single assistant message, which caused agents to falsely perceive sequential tool calls as parallel execution.

## Root Cause

`MessageMerger.shouldMerge()` returned `true` for consecutive assistant messages containing tool invocations from different LLM steps. When an agent performed Tool A in Step 1 and Tool B in Step 2 (both without text), they were merged into one message. The agent then interpreted this as having called both tools at once and could produce incorrect corrections.

## Changes

- Added `hasNewToolCalls()` method to `MessageMerger` that detects incoming tool invocations with state `'call'` and unrecognized `toolCallId`s (indicating a new LLM step)
- `shouldMerge()` now returns `false` when new tool calls from a different step are detected
- Tool result updates (state `'result'` with matching `toolCallId`) still merge correctly
- Added 263 lines of new test coverage for the merging behavior

## Test Plan

- All 92 message-list tests pass (including new tests for the merge fix)
- All 158 agent tests pass (5 pre-existing skips)

Closes #14124

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed an issue where consecutive tool operations from different steps were incorrectly merged into a single message. Tool calls are now properly kept as separate messages while tool result updates continue to merge correctly.

* **Tests**
  * Added comprehensive test coverage for tool operation message handling scenarios.

### DIFF
--- a/packages/core/src/agent/message-list/merge/MessageMerger.ts
+++ b/packages/core/src/agent/message-list/merge/MessageMerger.ts
@@ -24,6 +24,38 @@ export class MessageMerger {
   }
 
   /**
+   * Check if the incoming message introduces new tool calls (state: "call")
+   * that don't correspond to existing tool calls in the latest message.
+   * This indicates a new LLM step's tool calls, which should NOT be merged
+   * into the previous step's message.
+   */
+  private static hasNewToolCalls(latestMessage: MastraDBMessage, incomingMessage: MastraDBMessage): boolean {
+    const incomingParts = incomingMessage.content?.parts || [];
+    const latestParts = latestMessage.content?.parts || [];
+
+    // Collect toolCallIds from the latest message
+    const latestToolCallIds = new Set<string>();
+    for (const part of latestParts) {
+      if (part.type === 'tool-invocation') {
+        latestToolCallIds.add(part.toolInvocation.toolCallId);
+      }
+    }
+
+    // Check if the incoming message has tool-invocation "call" parts with new toolCallIds
+    for (const part of incomingParts) {
+      if (
+        part.type === 'tool-invocation' &&
+        part.toolInvocation.state === 'call' &&
+        !latestToolCallIds.has(part.toolInvocation.toolCallId)
+      ) {
+        return true;
+      }
+    }
+
+    return false;
+  }
+
+  /**
    * Check if we should merge an incoming message with the latest message
    *
    * @param latestMessage - The most recent message in the list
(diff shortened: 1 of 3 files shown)


packages/core/src/agent/message-list/merge/MessageMerger.ts
```diff
@@ -23,6 +23,38 @@ export class MessageMerger {
     return metadata?.mastra?.sealed === true;
   }
 
+  /**
+   * Check if the incoming message introduces new tool calls (state: "call")
+   * that don't correspond to existing tool calls in the latest message.
+   * This indicates a new LLM step's tool calls, which should NOT be merged
+   * into the previous step's message.
+   */
+  private static hasNewToolCalls(latestMessage: MastraDBMessage, incomingMessage: MastraDBMessage): boolean {
```
TylerBarnes, line 32: It should be ok for them to be merged actually! This is just about the location of `step-start` parts. The db merging isn't related to parallel tool calling, just the order of messages and step-start

TylerBarnes: I think this pr is more correct https://github.com/mastra-ai/mastra/pull/14126/changes

TylerBarnes: hmm actually that ones not quite right either. I think we need to solve this in the agent loop and make it add step-start parts when an actual new step is started

PR #14301, opened 2026-03-14
fix(core): normalize custom gateway provider keys to prevent duplicates in Studio

ivancho2:
---
  Bug report

  When registering a custom MastraModelGateway, the provider appears twice in Mastra Studio's provider selector.

  Root cause

  GET_PROVIDERS_ROUTE (/agents/providers) merges two sources with inconsistent keys:

  1. Static registry (PROVIDER_REGISTRY) — populated by fetchProvidersFromGateways(), which normalizes keys to
  ${gateway.id}/${providerId} (e.g. "melioffice/genai")
  2. Live fetchProviders() call — uses the raw key returned by the gateway (e.g. "genai")

  Since the keys differ, allProviders ends up with two separate entries for the same provider, and both are rendered in the
   UI.

  // Before (buggy) — two different keys → two UI entries
  allProviders["melioffice/genai"] = ...  // from static registry
  allProviders["genai"] = ...             // from live fetchProviders()

  Fix

  Apply the same key normalization logic used in fetchProvidersFromGateways when merging live results:

  const normalizedKey = providerId === gateway.id
    ? gateway.id
    : `${gateway.id}/${providerId}`;
  allProviders[normalizedKey] = config;

  To reproduce

  1. Implement a custom MastraModelGateway with any provider key in fetchProviders()
  2. Register it in new Mastra({ gateways: { myGateway: new MyGateway() } })
  3. Open Mastra Studio → select an agent → open the provider dropdown
  4. The custom provider appears twice

  Expected: one entry per provider
  Actual: two entries — one from static registry, one from live call

## Summary by CodeRabbit

* **Bug Fixes**
  * Gateway-provided providers are now exposed with normalized keys to avoid duplicate or conflicting provider entries in the Studio UI.

* **Chores**
  * Broadened compatibility with the core runtime across many packages to support newer core versions (reduces install/version conflicts).
  * Added release metadata entries documenting version bumps and dependency alignment notes.


abhiaiyer91:
Closing due to inactivity! Will open an issue instead

PR #14305, opened 2026-03-15
feat: add memory operation tracing with playground support

### LINKED ISSUE #10897: [FEATURE] Add Memory tracing
### Problem Statement

When I run an agent, and it uses memory, it would be helpful to see info about it in my trace.

### Proposed Solution

Add Memory tracing.

### Component

Other

### Alternatives Considered

_No response_

### Example Use Case

Memory is confusing. It would be great to gain some insight into how it is working in order to debug my agent.

### Additional Context

_No response_

### Verification

intojhanurag:
## Description

Add observability tracing to Memory operations (recall, save, delete, update working memory). Memory spans appear as child spans under agent runs, capturing operation type, message
counts, embedding token usage, semantic recall status, and vector result counts. The playground trace timeline now recognizes memory spans with a dedicated icon, color, and filter toggle.

  Changes:
  - Add `MEMORY_OPERATION` span type and `MEMORY` entity type to core observability types
  - Add `MemoryOperationAttributes` interface for memory-specific span data
  - Add optional `observabilityContext` parameter to abstract `MastraMemory` methods and concrete `Memory` implementations
  - Thread `observabilityContext` from agent and network call sites into memory operations
  - Register memory as a recognized span type in the playground UI with `MemoryIcon`
 
image

  ## Related Issue(s)
  
  Fixes #10897

  ## Type of Change
  - [ ] Documentation update
  - [ ] Code refactoring
  - [ ] Performance improvement
  - [x] Test update

  ## Checklist

## Summary by CodeRabbit

* **New Features**
  * Memory operations (recall, save, delete, update) now support opt-in tracing, emitting memory-operation traces that capture operation type, counts, token/embedding metrics and success status.
  * Tracing is opt-in and preserves existing behavior when not used.
  * Playground UI now shows Memory spans in the trace timeline with a dedicated icon, color, and filter.

* **Tests**
  * Added tests validating memory tracing behavior for success and error paths.

### DIFF
--- a/packages/core/src/agent/agent.ts
+++ b/packages/core/src/agent/agent.ts
@@ -36,7 +36,7 @@ import { networkLoop } from '../loop/network';
 import type { Mastra } from '../mastra';
 import type { MastraMemory } from '../memory/memory';
 import type { MemoryConfigInternal } from '../memory/types';
-import type { DefinitionSource, TracingProperties, ObservabilityContext } from '../observability';
+import type { DefinitionSource, TracingContext, TracingProperties, ObservabilityContext } from '../observability';
 import {
   EntityType,
   InternalSpans,
@@ -2642,12 +2642,14 @@ export class Agent<
     vectorMessageSearch,
     memoryConfig,
     requestContext,
+    tracingContext,
   }: {
     resourceId?: string;
     threadId: string;
     vectorMessageSearch: string;
     memoryConfig?: MemoryConfigInternal;
     requestContext: RequestContext;
+    tracingContext?: TracingContext;
   }): Promise<{ messages: MastraDBMessage[] }> {
     const memory = await this.getMemory({ requestContext });
     if (!memory) {
@@ -2669,6 +2671,7 @@ export class Agent<
       threadConfig: memoryConfig,
       // The new user messages aren't in the list yet cause we add memory messages first to try to make sure ordering is correct (memory comes before new user messages)
       vectorSearchString: threadConfig.semanticRecall && vectorMessageSearch ? vectorMessageSearch : undefined,
+      tracingContext,
     });
   }
 
--- a/packages/core/src/agent/agent.ts
+++ b/packages/core/src/agent/agent.ts
@@ -36,7 +36,7 @@ import { networkLoop } from '../loop/network';
 import type { Mastra } from '../mastra';
 import type { MastraMemory } from '../memory/memory';
 import type { MemoryConfigInternal } from '../memory/types';
-import type { DefinitionSource, TracingContext, TracingProperties, ObservabilityContext } from '../observability';
+import type { DefinitionSource, TracingProperties, ObservabilityContext } from '../observability';
 import {
   EntityType,
   InternalSpans,
@@ -2642,14 +2642,12 @@ export class Agent<
     vectorMessageSearch,
     memoryConfig,
     requestContext,
-    tracingContext,
   }: {
     resourceId?: string;
     threadId: string;
     vectorMessageSearch: string;
     memoryConfig?: MemoryConfigInternal;
     requestContext: RequestContext;
-    tracingContext?: TracingContext;
   }): Promise<{ messages: MastraDBMessage[] }> {
     const memory = await this.getMemory({ requestContext });
     if (!memory) {
@@ -2671,7 +2669,6 @@ export class Agent<
       threadConfig: memoryConfig,
       // The new user messages aren't in the list yet cause we add memory messages first to try to make sure ordering is correct (memory comes before new user messages)
       vectorSearchString: threadConfig.semanticRecall && vectorMessageSearch ? vectorMessageSearch : undefined,
-      tracingContext,
     });
   }
 
--- a/packages/core/src/loop/network/index.ts
+++ b/packages/core/src/loop/network/index.ts
@@ -10,7 +10,7 @@ import type { StructuredOutputOptions } from '../../agent/types';
 import { ErrorCategory, ErrorDomain, MastraError } from '../../error';
 import type { MastraLLMVNext } from '../../llm/model/model.loop';
 import { noopLogger } from '../../logger';
-import type { ObservabilityContext } from '../../observability';
+import type { ObservabilityContext, TracingContext } from '../../observability';
 import { createObservabilityContext, resolveObservabilityContext } from '../../observability';
 import { ProcessorRunner } from '../../processors/runner';
 import type { RequestContext } from '../../request-context';
@@ -307,6 +307,7 @@ export async function prepareMemoryStep({
               resourceId: thread?.resourceId,
             },
           ] as MastraDBMessage[],
+          tracingContext: observabilityContext.tracingContext,
         }),
       );
     }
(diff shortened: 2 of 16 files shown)


packages/core/src/agent/agent.ts
```diff
@@ -32,7 +32,7 @@ import { networkLoop } from '../loop/network';
 import type { Mastra } from '../mastra';
 import type { MastraMemory } from '../memory/memory';
 import type { MemoryConfigInternal } from '../memory/types';
-import type { TracingProperties, ObservabilityContext } from '../observability';
+import type { TracingContext, TracingProperties, ObservabilityContext } from '../observability';
```
epinzur: tracingContext should exist in ObservabilityContext... no need to pass it separately

packages/core/src/observability/types/tracing.ts
```diff
@@ -370,6 +374,22 @@ export interface WorkflowWaitEventAttributes extends AIBaseAttributes {
   waitDurationMs?: number;
 }
 
+/**
+ * Memory operation attributes
+ */
+export interface MemoryOperationAttributes extends AIBaseAttributes {
+  operationType?: 'recall' | 'save' | 'delete' | 'update_working';
+  threadId?: string;
+  resourceId?: string;
```
epinzur: i think `threadId` and `resourceId` might already be saved in the span metadata. If so, no need to duplicate them here.

packages/core/src/processors/memory/message-history.ts
```diff
@@ -80,43 +100,68 @@ export class MessageHistory implements Processor {
...
+          messageList.add(msg, 'memory');
+        }
+      }
 
-    if (chronologicalMessages.length === 0) {
-      return messageList;
-    }
+      span?.end({
+        output: { messageCount: chronologicalMessages.length },
+        attributes: { success: true, messageCount: chronologicalMessages.length },
```
epinzur: avoid duplicating data.  put `messageCount` in `attributes` or `output` but not both.

packages/core/src/processors/memory/message-history.ts
```diff
@@ -208,11 +250,29 @@ export class MessageHistory implements Processor {
...
+    );
+
+    try {
+      await this.persistMessages({ messages: messagesToSave, threadId, resourceId });
+      // add extra 1ms latency to make sure the next generate has not the same input
+      await new Promise(resolve => setTimeout(resolve, 10));
+
+      span?.end({
+        attributes: { success: true, messageCount: messagesToSave.length },
+      });
```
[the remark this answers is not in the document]
intojhanurag: Hey @coderabbitai Span logs messagesToSave.length but persistMessages filters- Valid but not worth fixing

persistMessages calls filterMessagesForPersistence (line 292) which can reduce the count. So the span may overcount. However, fixing this requires changing persistMessages to return a count — that's a public method used by other processors (ObservationalMemory). The discrepancy is minor (partial tool calls and working memory tags are rare), and changing a shared API signature for trace accuracy is overkill.

packages/core/src/loop/network/index.ts
```diff
@@ -307,6 +307,7 @@ export async function prepareMemoryStep({
               resourceId: thread?.resourceId,
             },
           ] as MastraDBMessage[],
+          tracingContext: observabilityContext.tracingContext,
```
epinzur: can you change to pass the full `observabilityContext` through instead of just tracing context?

In general we now try to always pass `observabilityContext` instead of `tracingContext`.

This allows for context-aware tracing, metrics & logging.

packages/core/src/observability/types/tracing.ts
```diff
@@ -352,6 +354,20 @@ export interface WorkflowWaitEventAttributes extends AIBaseAttributes {
   waitDurationMs?: number;
 }
 
+/**
+ * Memory operation attributes
+ */
+export interface MemoryOperationAttributes extends AIBaseAttributes {
+  operationType?: 'recall' | 'save' | 'delete' | 'update_working';
```
epinzur: maybe change `update_working` to just `update` ?

packages/core/src/observability/types/tracing.ts
```diff
@@ -352,6 +354,20 @@ export interface WorkflowWaitEventAttributes extends AIBaseAttributes {
...
+  messageCount?: number;
+  embeddingTokens?: number;
+  semanticRecallEnabled?: boolean;
+  vectorResultCount?: number;
+  workingMemoryEnabled?: boolean;
+  lastMessages?: number | false;
```
epinzur, line 403: do we need `false` option here? is lastMessages == 'undefined` equivalent?

packages/core/src/processors/memory/message-history.ts
```diff
@@ -70,6 +89,7 @@ export class MessageHistory implements Processor {
     } & Partial<ObservabilityContext>,
   ): Promise<MessageList | MastraDBMessage[]> {
     const { messageList, requestContext } = args;
+    const tracingContext = args.tracingContext ?? args.tracing;
```
epinzur: i don't think you need this.  i think these two are just alias's of each other anyway

packages/core/src/processors/memory/message-history.ts
```diff
@@ -61,6 +62,24 @@ export class MessageHistory implements Processor {
     return null;
   }
 
+  private createMemorySpan(
+    operationType: MemoryOperationAttributes['operationType'],
+    tracingContext?: TracingContext,
```
epinzur: i know this is just for creating the span, but passing the observabilityCOntext here too, probably simplifies code elsewhere in this file

packages/core/src/observability/types/tracing.ts
intojhanurag: false means "history explicitly disabled" which is different from undefined or not set. Dropping it loses the info.

packages/core/src/processors/memory/message-history.ts
```diff
@@ -61,6 +62,24 @@ export class MessageHistory implements Processor {
...
+    input?: any,
+    attributes?: Partial<MemoryOperationAttributes>,
+  ) {
+    if (!tracingContext) return undefined;
+    return getOrCreateSpan({
```
epinzur: I think in most cases you prob should be using `tracingContext?.currentSpan?.createChildSpan()`

and not `getOrCreateSpan()`

getOrCreate will create a new trace if the tracingContext doesn't exist, and we don't want that.

we only want to add memory spans to existing traces.

packages/core/src/observability/types/tracing.ts
```diff
@@ -388,6 +390,20 @@ export interface WorkflowWaitEventAttributes extends AIBaseAttributes {
...
+ */
+export interface MemoryOperationAttributes extends AIBaseAttributes {
+  operationType?: 'recall' | 'save' | 'delete' | 'update';
+  messageCount?: number;
+  embeddingTokens?: number;
+  semanticRecallEnabled?: boolean;
+  vectorResultCount?: number;
+  workingMemoryEnabled?: boolean;
+  lastMessages?: number | false;
+  success?: boolean;
```
epinzur: `success` would probably be better as a span output instead of an attribute.

epinzur (approved):

PR #14333, opened 2026-03-16
[Docs] Clarify disconnect behavior for chat routes

### LINKED ISSUE #6715: How to access consumeStream functionality for handling client disconnects during streaming?
## Problem

We're building a chat application with Mastra that needs to persist AI messages even when clients disconnect during streaming (e.g., page refresh, network issues). Currently, when a client disconnects, the stream gets aborted and the `onFinish` callback never executes, resulting in lost messages.

## Question

The Vercel AI SDK provides a `consumeStream()` method specifically for this use case (see [Chatbot Message Persistence docs](https://ai-sdk.dev/docs/ai-sdk-ui/chatbot-message-persistence)). This method removes backpressure and allows streams to complete server-side even when clients disconnect.

Since Mastra uses the Vercel AI SDK internally, how can we access this `consumeStream` functionality? We don't see it exposed in the Mastra documentation or API.

## Minimal Example

```typescript

const agent = new Agent({
  name: 'ChatBot',
  model: openrouter('gpt-4'),
  defaultStreamOptions({ runtimeContext }) {
    return {
      onFinish: async (response) => {
        // This callback needs to execute even if client disconnects
        await saveMessageToDatabase(response.text);
      }
    };
  }
});

// How to ensure onFinish executes even on client disconnect?
// In Vercel AI SDK this would be: await consumeStream(result.stream);
```

## Use Case

- User sends message to AI agent
- Client refreshes page mid-streaming
- AI response should still be saved to database via `onFinish`
- Currently: message is lost because stream aborts

## Expected Behavior

Mastra should either:
1. Expose access to `consumeStream` method
2. Handle this automatically in agent streaming
3. Provide alternative mechanism for guaranteed message persistence

Thanks for any guidance on this!

TheIsrael1:
## Summary
- clarify that `chatRoute()` forwards the request abort signal and aborts generation on client disconnect
- document a custom route pattern that keeps generation running server-side with `consumeStream()`
- add a Hono integration test that aborts the client mid-stream and verifies the full response is still persisted

Fixes #6715

## Testing
- pnpm vitest run server-adapters/hono/src/__tests__/custom-stream-route.test.ts
- pnpm vitest run packages/core/src/agent/__tests__/save-and-errors.test.ts

## Summary by CodeRabbit

* **Documentation**
  * Clarified chatRoute() behavior on client disconnect: generation is aborted when the client disconnects.
  * Added documentation and example for continuing server-side generation after client disconnect using a custom route pattern.


docs/src/content/en/docs/server/custom-api-routes.mdx
```diff
@@ -40,6 +41,66 @@ curl http://localhost:4111/my-custom-route
 
 Each route's handler receives the Hono `Context`. Within the handler you can access the `Mastra` instance to fetch or call agents and workflows.
 
+## Continue generation after client disconnect
```
LekoArts: Move this below "Authentication" please, I wouldn't deem it as important to move it that high compared to the rest

TheIsrael1: That makes sense, done now!

LekoArts (approved):

PR #14354, opened 2026-03-16
feat(core): add filesystem-level optimistic concurrency for file writes [COR-726]

jq-ln:
### Problem

The existing per-file write lock (`InMemoryFileWriteLock`) and read tracker (`InMemoryFileReadTracker`) operate at the tool wrapper level. Direct calls to `filesystem.writeFile()` have no concurrency protection, and external writes (e.g., LSP editors saving a file) can silently overwrite in-flight changes because the mtime check happens at tool invocation time, not at the actual moment of write.

### Solution

Add filesystem-level optimistic concurrency as a second line of defense alongside the existing tool-level checks:

1. **New `expectedMtime` option on `WriteOptions`** — any caller can opt into mtime validation at write time
2. **`StaleFileError`** — thrown when the on-disk mtime doesn't match the expected value
3. **Automatic mtime pass-through** — workspace tools (`edit_file`, `write_file`, `ast_edit`) automatically pass the recorded mtime from the read tracker when `requireReadBeforeWrite` is enabled

### Why both layers?

- **Tool-level lock** prevents the agent's own concurrent writes from interleaving
- **Filesystem-level mtime check** catches external modifications that happen between the tool-level check and the actual write
- Together they provide defense-in-depth

### Changes

| File | Change |
|------|--------|
| `errors.ts` | Added `StaleFileError` class (code: `ESTALE`) |
| `filesystem.ts` | Added `expectedMtime` to `WriteOptions` |
| `local-filesystem.ts` | Mtime check before write in `writeFile()` |
| `tools.ts` | Inject `__expectedMtime` from read tracker into context |
| `edit-file.ts`, `write-file.ts`, `ast-edit.ts` | Pass `expectedMtime` to `filesystem.writeFile()` |
| `overview.mdx` | Documented two-layer defense and `expectedMtime` usage |
| `filesystem.mdx`, `local-filesystem.mdx` | Added `expectedMtime` to writeFile options reference |

### Tests

- 4 filesystem-level tests in `local-filesystem.test.ts` (stale mtime rejected, matching mtime succeeds, new file succeeds, backward compat)
- 4 integration tests in `write-lock.test.ts` (external modification detected via tool pipeline, happy path, write_file rejection, new file with requireReadBeforeWrite)
- All 128 tests pass, typecheck clean, build clean

## Summary by CodeRabbit

* **New Features**
  * Optimistic concurrency for file writes: supply an expected modification time (expectedMtime) to avoid overwriting concurrent changes; mismatches raise StaleFileError.

* **Bug Fixes**
  * Workspace write operations now consistently propagate recorded modification times as an extra safety check to prevent accidental overwrites.

* **Documentation**
  * Added guidance and examples for read-before-write safety and the expectedMtime option.

* **Tests**
  * Added tests covering optimistic concurrency and related write behaviors.


packages/core/src/workspace/tools/tools.ts
```diff
@@ -120,6 +120,12 @@ function wrapWithReadTracker(
           if (check.needsReRead) {
             throw new FileReadRequiredError(input.path, check.reason!);
           }
+          // Pass the mtime from the last read through to the tool so it can
+          // forward it to filesystem.writeFile() for optimistic concurrency.
+          const record = readTracker.getReadRecord(input.path);
+          if (record) {
+            context = { ...context, __expectedMtime: record.modifiedAtRead };
+          }
```
NikAiyer: Right now this only runs when requireReadBeforeWrite is configured. Do we want to track this on all writes regardless of that config option?

jq-ln: Good catch! Yes, we should track this on all writes. I'll fix that.

NikAiyer (approved):

PR #14360, opened 2026-03-16
fix(core): parallelize serial I/O in skills processor

### LINKED ISSUE #14290: Abnormally long time spent on Skills Processor
This issue was created from Discord post 1482180532174454956:

[Open in Discord](https://discord.com/channels/1309558646228779139/1482180532174454956)

In Agent runs, the Skills Processor steps takes disproportionately long, even if no skill is ultimately used. In observability, it's labelled "input step processor: skills-processor", and it takes a minimum of 25 seconds each agent run.

The Agent only has 6 basic skills mounted in its `/skills` directory in its workspace, so I'm not sure why the processor takes so long each time. If it helps, I've noticed that going to `http://localhost:4111/workspaces?tab=skills` in Mastra Studio also takes ~30 seconds to load, so maybe that's correlated? Opening Skills in the studio also takes much longer than opening other markdown files in the Agent's workspace from the studio.

Let me know if there's anything I can optimize to speed this step up 🙏

roaminro:
Fixes #14290

The skills processor was running all filesystem I/O sequentially — ~46 serial operations for 6 skills. On remote/cloud backends (S3, GCS, AgentFS) where each op takes 200-500ms, this added up to 9-23 seconds per agent run.

This PR parallelizes the independent I/O operations using `Promise.all` / `Promise.allSettled` in 6 places across `WorkspaceSkillsImpl`: subdirectory discovery, reference reads, skill directory processing, staleness checks, glob resolution, and reference indexing. The outer loop over resolved paths stays serial since later paths intentionally overwrite earlier ones.

Also fixes a bug in `CompositeVersionedSkillSource.stat()` that returned `new Date()` for the root path, making every `maybeRefresh()` think skills were stale and triggering a full re-discovery. Now uses the max `versionCreatedAt` across entries.

Test results show peak concurrency going from 1 → 18 and elapsed time dropping from ~2300ms to ~460ms for 6 skills with 50ms simulated I/O latency.

ELI5: This change makes the skills-loading step much faster by doing many file reads at the same time instead of one-by-one, and fixes a bug that made the system think skills were always out-of-date. It also adds tests to prove the speedup.

## Summary

- Performance: Parallelizes independent filesystem I/O in the skills processor (WorkspaceSkillsImpl) to reduce serial latency on remote/cloud-backed filesystems. Key areas parallelized: subdirectory discovery, reference reads, per-skill directory processing, staleness checks, glob resolution, and reference indexing. The outer loop over resolved paths remains serial to preserve deterministic overwrite ordering.
- Bug fix: Corrects CompositeVersionedSkillSource.stat() so the root path's created/modified timestamps use the maximum versionCreatedAt across entries instead of returning the current time; this prevents unnecessary re-discovery on refresh.
- Tests: Adds a performance-focused test suite that simulates slow I/O, measures concurrency and elapsed time, and validates that peak concurrency increases and wall-clock time drops (reported: peak concurrency up to 18 and elapsed time reduced from ~2300 ms to ~460 ms for a 6-skill scenario with simulated latency).
- Changesets: Adds a changeset note documenting the fix and performance improvement.
- Risk/compatibility: No public API or exported signatures changed. Error handling was preserved by allowing per-entry failures to be skipped while keeping overall discovery resilient.

## Release Notes

* Bug Fixes
  * Fixed incorrect root stat in CompositeVersionedSkillSource that caused unnecessary re-discovery.
* Performance
  * Significantly improved skills discovery performance by parallelizing filesystem I/O operations.
* Tests
  * Added tests that simulate slow I/O and validate concurrency and speed improvements.


abhiaiyer91 (approved):

roaminro:
@coderabbitai - can you please do one last review

PR #14401, opened 2026-03-17
fix(schema-compat): map draft target names for Zod v4 compatibility

brick-pixel:
## What

Added a target name mapping in the Zod v4 adapter before calling `z.toJSONSchema()`.

## Why

Zod v4's `z.toJSONSchema()` expects `"draft-7"` as the target name, but `@mastra/schema-compat` passes `"draft-07"`. This causes repeated console warnings on every schema conversion:

```
Invalid target: draft-07
Invalid target: draft-07
```

The mismatch is because Mastra uses `"draft-07"` (JSON Schema convention) while Zod v4 uses `"draft-7"` (shortened form).

## Changes

**File:** `packages/schema-compat/src/standard-schema/adapters/zod-v4.ts`

Added a mapping object that translates between naming conventions before passing the target to `z.toJSONSchema()`:

```typescript
const ZOD_V4_TARGET_MAP: Record<string, string> = {
  'draft-07': 'draft-7',
  'draft-04': 'draft-4',
};
```

`"draft-2020-12"` is unchanged since both Mastra and Zod v4 use the same name for it.

Fixes #14399

## Summary by CodeRabbit

* **Bug Fixes**
  * Normalize JSON Schema draft target names (e.g., draft-07, draft-04) to the expected format to avoid invalid-target handling and ensure the $schema field is emitted.
* **Tests**
  * Added tests to confirm generating JSON Schemas for those targets no longer produces invalid-target warnings and that $schema is present.


wardpeet (approved):

wardpeet:
@coderabbitai can you do another review and tell me what a good changeset description would be?

wardpeet (approved):

PR #14407, opened 2026-03-17
fix(schema-compat): ensure all anyOf branches have a type key for OpenAI strict mode

### LINKED ISSUE #14403: OpenAI “invalid_function_parameters” when parent agent uses structuredOutput + agents delegation
This issue was created from Discord post 1483383559241601105:

[Open in Discord](https://discord.com/channels/1309558646228779139/1483383559241601105)

When a parent agent is configured with agents: { ...subAgents } (so Mastra exposes sub-agents as tool/functions like agent-userFlowBuilderAgent) and the parent call uses structuredOutput (JSON schema / strict: true), OpenAI rejects the request with:

Invalid schema for function 'agent-userFlowBuilderAgent': In context=('properties', 'resumeData', 'anyOf', '0'), schema must have a 'type' key.
param: tools[0].parameters / code: invalid_function_parameters

This only happens when the parent call requests structured output. If the parent returns a single markdown/text output (no structuredOutput), the same delegation flow works.

It looks like the generated tool schema for delegated agents contains an invalid JSON Schema branch for resumeData:

- resumeData.anyOf[0] has a description but no type
- OpenAI requires each anyOf branch to include a type

**Why this seems like a Mastra bug**
From the request payload sent to OpenAI (/v1/responses), Mastra generates tool parameters like:

`{
  "type": "function",
  "name": "agent-userFlowBuilderAgent",
  "parameters": {
    "type": "object",
    "properties": {
      "resumeData": {
        "anyOf": [
          {
            "description": "The resumeData object created from the resumeSchema of suspended tool"
          },
          {
            "type": "null"
          }
        ]
      }
    }
  }
}`

Rest of the details are in the attached file, since I was hitting a character limit when putting every thing here.

I have worked around this in the past by just accepting a single markdown output from the agent, rather than structured output. But I need a structured output now.

As a workaround, I can think of making a separate agent call post the main call, just to destructure the output, but that seems like a hack.

intojhanurag:
## Description

This fixes OpenAI invalid_function_parameters errors when a parent agent uses structuredOutput and delegates to sub-agents exposed as tools. 

The real user issue was that delegated tool schemas could generate resumeData branches without a valid OpenAI-strict type/object shape, so OpenAI rejected the request before execution.

This update fixes that by expanding typeless schemas into concrete OpenAI-safe branches, recursively handling anyOf/oneOf/allOf, and ensuring the object variant is strict (additionalProperties: false). Verified with a real OpenAI API repro and added regression coverage.

## Related Issue(s)

Fixes #14403

## Type of Change
- [ ] Documentation update
- [x] Code refactoring
- [ ] Performance improvement
- [x] Test update

## Checklist

### DIFF
--- a/.changeset/fix-typeless-anyof-branches.md
+++ b/.changeset/fix-typeless-anyof-branches.md
@@ -9,6 +9,7 @@ fix: ensure all anyOf branches have a type key for OpenAI strict mode
 producing `anyOf: [{ description: "..." }, { type: "null" }]` where the first
 branch had no `type` key. OpenAI rejects this when using structuredOutput.
 
-The OpenAI compat layer now assigns `type: "object"` to typeless anyOf branches,
-and `fixTypelessProperties` now recurses into anyOf/oneOf/allOf branches as a
-safety net.
+The OpenAI compat layer now expands typeless branches into concrete typed
+branches, and `fixTypelessProperties` now recurses into anyOf/oneOf/allOf
+branches while preserving a strict object variant with
+`additionalProperties: false`.
--- a/.changeset/fix-typeless-anyof-branches.md
+++ b/.changeset/fix-typeless-anyof-branches.md
@@ -1,0 +1,14 @@
+---
+'@mastra/core': patch
+'@mastra/schema-compat': patch
+---
+
+fix: ensure all anyOf branches have a type key for OpenAI strict mode
+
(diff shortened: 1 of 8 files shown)


wardpeet:
@coderabbitai, isn't this fixed by https://github.com/mastra-ai/mastra/pull/14624

intojhanurag:
closing this due, I think , already fixed by #14624

PR #14417, opened 2026-03-17
fix(core): bundle @internal/core to fix E2E monorepo test

rphansen91:
## Summary
- `@internal/core` (added in #14270) is `private: true` and never published to the local test registry, causing `ERR_PNPM_FETCH_404` in the E2E monorepo test
- Added `noExternal: ['@internal/core']` to `tsup.config.ts` so it gets bundled into `@mastra/core` dist
- Moved `@internal/core` from `dependencies` to `devDependencies`, matching how all other `@internal/*` packages are handled

## Test plan
- [ ] E2E monorepo test passes (was failing with 404 for `@internal/core`)
- [ ] `pnpm build:core` succeeds and dist files contain no `@internal/core` imports

## Summary by CodeRabbit

* **Chores**
  * Updated internal dependencies to improve project configuration.


packages/core/tsup.config.ts
```diff
@@ -66,6 +66,7 @@ export default defineConfig({
   format: ['esm', 'cjs'],
   clean: true,
   dts: false,
+  noExternal: ['@internal/core'],
```
CalebBarnes: I don't think this is necessary, at least we didn't do this for other internal packages we bundle in to core. Tsup only externalizes dependencies so it should bundle it as soon as you move it to dev deps.

wardpeet: This is automatic
```suggestion
```

PR #14442, opened 2026-03-18
fix(observability): handle plain object requestContext in BaseSpan to prevent crash on HTTP API workflows

intojhanurag:
## Description
  
When a workflow is invoked via the HTTP API (through `@mastra/client-js`) with observability enabled, the `BaseSpan` constructor crashes with `TypeError: requestContext.size is not a function`. This happens because the evented workflow engine converts `RequestContext` class instances to plain objects via `Object.fromEntries()` before publishing events, and the
Inngest durable execution engine's `deserializeRequestContext` incorrectly returns a `Map` cast as `RequestContext`. Both lose the `.size()` method and `.all` getter that `BaseSpan` relies on.

**Changes:**
  - `observability/mastra/src/spans/base.ts`: Made `BaseSpan` constructor defensive — duck-types `.size` and `.all` to handle both `RequestContext` instances and plain objects
  - `observability/mastra/src/instances/base.ts`: Made `extractKeys` defensive — duck-types `.get()` to handle both forms
  - `packages/core/src/workflows/default.ts`: Fixed `deserializeRequestContext` to return an actual `RequestContext` instance instead of a `Map` incorrectly cast as `RequestContext`

  ## Related Issue(s)
  
  Fixes #14425

  ## Type of Change
  - [ ] Documentation update
  - [ ] Code refactoring
  - [ ] Performance improvement
  - [x] Test update

  ## Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed crashes during Inngest durable execution replay when observability is enabled
  * Fixed runtime errors in observability-related HTTP API workflow invocations

### DIFF
--- a/.changeset/spicy-fans-fix.md
+++ b/.changeset/spicy-fans-fix.md
@@ -1,0 +1,5 @@
+---
+'@mastra/observability': patch
+---
+
+Fixed crash when workflow is invoked via HTTP API with observability enabled. The BaseSpan constructor now handles both RequestContext class instances and plain objects, preventing the 'requestContext.size is not a function' error.
--- a/.changeset/spicy-fans-fix.md
+++ b/.changeset/spicy-fans-fix.md
@@ -1,5 +1,0 @@
----
-'@mastra/observability': patch
----
-
-Fixed crash when workflow is invoked via HTTP API with observability enabled. The BaseSpan constructor now handles both RequestContext class instances and plain objects, preventing the 'requestContext.size is not a function' error.
--- a/observability/mastra/src/instances/base.ts
+++ b/observability/mastra/src/instances/base.ts
@@ -569,12 +569,16 @@ export abstract class BaseObservabilityInstance extends MastraBase implements Ob
    */
   protected extractKeys(requestContext: RequestContext, keys: string[]): Record<string, any> {
     const result: Record<string, any> = {};
+    const getValue =
+      typeof requestContext.get === 'function'
+        ? (k: string) => requestContext.get(k)
+        : (k: string) => (requestContext as Record<string, any>)[k];
 
     for (const key of keys) {
       // Handle dot notation: get first part from RequestContext, then navigate nested properties
       const parts = key.split('.');
       const rootKey = parts[0]!; // parts[0] always exists since key is a non-empty string
-      const value = requestContext.get(rootKey);
+      const value = getValue(rootKey);
 
       if (value !== undefined) {
         // If there are nested parts, extract them from the value
--- a/observability/mastra/src/instances/base.ts
+++ b/observability/mastra/src/instances/base.ts
@@ -569,16 +569,12 @@ export abstract class BaseObservabilityInstance extends MastraBase implements Ob
    */
   protected extractKeys(requestContext: RequestContext, keys: string[]): Record<string, any> {
     const result: Record<string, any> = {};
-    const getValue =
-      typeof requestContext.get === 'function'
-        ? (k: string) => requestContext.get(k)
-        : (k: string) => (requestContext as Record<string, any>)[k];
 
     for (const key of keys) {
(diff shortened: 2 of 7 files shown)


observability/mastra/src/instances/base.ts
```diff
@@ -569,12 +569,16 @@ export abstract class BaseObservabilityInstance extends MastraBase implements Ob
    */
   protected extractKeys(requestContext: RequestContext, keys: string[]): Record<string, any> {
     const result: Record<string, any> = {};
+    const getValue =
+      typeof requestContext.get === 'function'
+        ? (k: string) => requestContext.get(k)
+        : (k: string) => (requestContext as Record<string, any>)[k];
```
epinzur, line 572: it seems like your fix in `deserializeRequestContext` is sufficient, and this change here is probably not needed.

observability/mastra/src/spans/base.ts
```diff
@@ -157,8 +157,13 @@ export abstract class BaseSpan<TType extends SpanType = any> implements Span<TTy
...
     );
-    if (options.requestContext && options.requestContext.size() > 0) {
-      this.requestContext = deepClean(options.requestContext.all, this.deepCleanOptions);
+    if (options.requestContext) {
+      const ctx = options.requestContext;
+      const size = typeof ctx.size === 'function' ? ctx.size() : Object.keys(ctx).length;
+      if (size > 0) {
+        const all = typeof ctx.all !== 'undefined' ? ctx.all : ctx;
+        this.requestContext = deepClean(all, this.deepCleanOptions);
+      }
```
epinzur: same here:

it seems like your fix in `deserializeRequestContext` is sufficient, and this change here is probably not needed.

observability/mastra/src/tracing.test.ts
```diff
@@ -1972,6 +1972,49 @@ describe('Tracing', () => {
...
+        exporters: [testExporter],
+      });
+
+      const plainObj = { userId: 'user-123', tenantId: 'tenant-456' };
+
+      const span = observability.startSpan({
+        type: SpanType.AGENT_RUN,
+        name: 'test-agent',
+        attributes: {},
+        requestContext: plainObj as any,
```
epinzur: does this work as `requestContext: plainObj as unknown as RequestContext` or similar?

if not, feel free to leave as-is

observability/mastra/src/tracing.test.ts
```diff
@@ -1972,6 +1972,49 @@ describe('Tracing', () => {
...
+        serviceName: 'test-service',
+        name: 'test',
+        exporters: [testExporter],
+      });
+
+      const span = observability.startSpan({
+        type: SpanType.AGENT_RUN,
+        name: 'test-agent',
+        attributes: {},
+        requestContext: {} as any,
```
epinzur: same here?

observability/mastra/src/tracing.test.ts
epinzur: nevermind, pls ignore this comment

intojhanurag:
No, the evented engine does not use deserializeRequestContext. It manually reconstructs RequestContext via new RequestContext() at its consumption points (workflow-event-processor/index.ts:1042, loop.ts:43, sleep.ts:91/184, parallel.ts:104) before step execution. So the crash was specific to the Inngest path only.

.changeset/spicy-fans-fix.md
```diff
@@ -0,0 +1,5 @@
+---
+'@mastra/observability': patch
+---
+
+Fixed crash when workflow is invoked via HTTP API with observability enabled. The BaseSpan constructor now handles both RequestContext class instances and plain objects, preventing the 'requestContext.size is not a function' error.
```
epinzur: i think we can remove this changeset now too

intojhanurag:
@epinzur Now is it good to go ?

epinzur (approved):

PR #14443, opened 2026-03-18
fix(schema-compat): use dynamic createRequire to avoid breaking Vite browser builds

guoyangzhen:
## Summary

Fixes #14440

Remove static `import { createRequire } from "node:module"` from `@mastra/schema-compat` which causes Vite browser builds to fail with:

```
"createRequire" is not exported by "__vite-browser-external", imported by
"node_modules/@mastra/schema-compat/dist/chunk-XI3YONZL.js"
```

Rollup resolves the static import at build time even when the code path is never executed in the browser. The fix lazily loads `createRequire` via `Function("m", "return require(m)")("node:module")` which is invisible to static analysis.

## Root Cause

`zod-v4.ts` had a static `import { createRequire } from "node:module"` at the top level. When `@mastra/client-js` → `@mastra/core` → `@mastra/schema-compat/schema` → `standard-schema.ts` → `zod-v4.ts` is bundled by Vite, Rollup sees this import and fails because `node:module` is a Node.js builtin.

The code using `createRequire` is only meant to run in Node.js (for server-side zod v4 JSON Schema conversion), but Rollup does static analysis before tree-shaking.

## Fix

1. Remove the static `import { createRequire } from "node:module"`
2. Lazily load `createRequire` at runtime using `Function("m", "return require(m)")("node:module")` which:
   - Works in CJS (where `require` is global) — returns `require` directly
   - Works in ESM Node.js — gets `createRequire` dynamically
   - Is invisible to Rollup static analysis (no static import)
   - In browser builds, the code path is never reached, so no error at runtime
3. Fallback throws a clear error message if somehow called in a browser

## Changes

- **`packages/schema-compat/src/standard-schema/adapters/zod-v4.ts`**: Replace static `createRequire` import with lazy runtime initialization that works in both CJS and ESM without breaking browser bundlers

## Summary by CodeRabbit

* **Refactor**
  * Enhanced module resolution for improved compatibility across CommonJS, ES Modules, and browser environments.
  * Strengthened build-time safety with dynamic runtime resolution of the module loader.
  * Improved error handling and diagnostics for better support in non-Node environments.

### DIFF
--- a/packages/schema-compat/src/standard-schema/adapters/zod-v4.ts
+++ b/packages/schema-compat/src/standard-schema/adapters/zod-v4.ts
@@ -1,4 +1,3 @@
-import { createRequire } from 'node:module';
 import type { StandardSchemaV1, StandardJSONSchemaV1 } from '@standard-schema/spec';
 import type { StandardSchemaWithJSON, StandardSchemaWithJSONProps } from '../standard-schema.types';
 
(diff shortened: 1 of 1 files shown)


LekoArts:
Hello, thanks for the PR. We need to discuss first how we want to solve this in the issue and as such I'm closing your PR.

PR #14447, opened 2026-03-18
fix(core): return top-level agent run span ID instead of nested model spans

lforst:
### Summary

Fix `stream.spanId` / `stream.traceId` so public stream results point to the root exported span, not whichever nested span happens to be active during execution.

### Problem

MastraModelOutput was deriving traceId and spanId from tracingContext.currentSpan. In real agent.stream() execution, that span can be a workflow step, model generation, or model step span, so the returned spanId could point to an internal or nested child instead of the root span users see in e.g. Braintrust.

### Reasoning

`currentSpan` is an execution-local cursor, not a stable public API identifier. For public-facing APIs, the returned spanId should match the root span of the trace tree that is actually exported. That keeps lookup and correlation consistent even when internal workflow/model spans are active.

### What changed

- Added a helper to resolve the top-most non-internal exported span from the current tracing context.
- Updated `MastraModelOutput` to use that exported root span for `traceId` / `spanId`.
- Kept the behavior generic so it applies to public APIs beyond just `agent.stream()`.

### Tests

- Added a regression test covering the real traced `agent.stream()` path.
- Added a unit test proving nested/internal current spans still resolve to the exported root span.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed agent.stream() to correctly return the root agent run span ID instead of nested model span ID for improved trace correlation in observability integrations.
  * Results now properly expose rootSpanId for clearer span context across traced operations.

* **Tests**
  * Added integration tests validating proper span hierarchy, parent-child span relationships, and root span exposure in agent stream operations.

### DIFF
--- a/packages/core/src/agent/__tests__/stream-id.test.ts
+++ b/packages/core/src/agent/__tests__/stream-id.test.ts
@@ -4,8 +4,10 @@ import { MockLanguageModelV1 } from '@internal/ai-sdk-v4/test';
 import { convertArrayToReadableStream, MockLanguageModelV2 } from '@internal/ai-sdk-v5/test';
 import { beforeEach, describe, expect, it, vi } from 'vitest';
 import { z } from 'zod';
+import { Observability, TestExporter } from '../../../../../observability/mastra/src';
 import { Mastra } from '../../mastra';
 import { MockMemory } from '../../memory/mock';
+import { SpanType } from '../../observability';
 import type { Processor } from '../../processors';
 import { createTool } from '../../tools';
 import { Agent } from '../agent';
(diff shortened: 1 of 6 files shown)


packages/core/src/agent/__tests__/stream-id.test.ts
```diff
@@ -4,8 +4,10 @@ import { MockLanguageModelV1 } from '@internal/ai-sdk-v4/test';
 import { convertArrayToReadableStream, MockLanguageModelV2 } from '@internal/ai-sdk-v5/test';
 import { beforeEach, describe, expect, it, vi } from 'vitest';
 import { z } from 'zod';
+import { Observability, TestExporter } from '../../../../../observability/mastra/src';
```
epinzur: this is a cross-package import...

I think this test should instead be inside 'observability/mastra` package

lforst: moved it to integration tests!

packages/core/src/stream/base/output.ts
```diff
@@ -281,8 +281,9 @@ export class MastraModelOutput<OUTPUT = undefined> extends MastraBase {
     this.#transportRef = options.transportRef;
     this.#returnScorerData = !!options.returnScorerData;
     this.runId = options.runId;
-    this.traceId = options.tracingContext?.currentSpan?.externalTraceId;
-    this.spanId = options.tracingContext?.currentSpan?.id;
+    const resultSpan = getRootExportSpan(options.tracingContext?.currentSpan);
+    this.traceId = resultSpan?.externalTraceId;
+    this.spanId = resultSpan?.id;
```
epinzur, line 286: can you pls change the name of this property from `spanId` to `rootSpanId`, or `traceSpanId`.

Also pls update the comment from:
```
  /**
   * Span ID used on the execution (if the execution was traced).
   */
```

to something like:

```
  /**
   * RootSpanId of the trace used on the execution (if the execution was traced).
   */
```

as its currently defined it shouldn't be changed... but luckily its only used here, so we can easily change its meaning.

then in the `async getFullOutput():` method, around line 1342, adjust to:

```
spanId: this.rootSpanId,
```

epinzur (dismissed):

packages/core/src/stream/base/output.ts
epinzur: on 2nd thought, lets not do this.

epinzur (approved):

PR #14466, opened 2026-03-18
fix(core): strip tools array when toolChoice is 'none' to fix Gemini structured output

Aayush-engineer:
## Summary
Fixes #14459

## Problem
When an agent is in a Mastra instance with workflows registered,
`listWorkflowTools()` auto-injects workflow tools into every `agent.generate()`
call. When `structuredOutput` is set, Gemini rejects the request with:

> Function calling with a response mime type: 'application/json' is unsupported

Setting `toolChoice: 'none'` didn't help because `prepareToolsAndToolChoice()`
still serialized the full tools array — it only omitted tools when the object
was already empty.

## Root Cause
This is not a Gemini limitation — it's a Mastra bug. When the caller explicitly
sets `toolChoice: 'none'`, Mastra should honor that and not send the tools array
to the provider regardless of which provider is used.

## Fix
Added an early return in `prepareToolsAndToolChoice()` when `toolChoice === 'none'`
that returns `tools: undefined` while preserving `toolChoice: { type: 'none' }`.

Only 2 files changed:
- `packages/core/src/stream/aisdk/v5/compat/prepare-tools.ts` — the fix
- `packages/core/src/stream/aisdk/v5/compat/prepare-tools.test.ts` — regression test

## Tests
- 2 pre-existing failing tests now pass ✅
- 1 new regression test added for the exact bug scenario
  (toolChoice: 'none' with non-empty injected workflow tools) ✅
- 21/21 tests passing, 0 type errors ✅

## Summary by CodeRabbit

* **Bug Fixes**
  * Structured-output agents no longer fail when workflow tools are present.
  * When tool selection is set to "none", tools are omitted from provider requests to prevent provider errors (including Gemini).

* **Tests**
  * Added a regression test covering the "none" tool selection with non-empty workflow tools.

* **Chores**
  * Prepared a patch release changelog entry.


Aayush-engineer:
Addressed all 3 CodeRabbit review comments:

Rewrote both changesets to be user-facing outcome focused (no internal implementation details)
Fixed both catch {} blocks in tools.ts to only swallow not-found errors — real server faults now rethrow correctly
Scoped server changeset to @mastra/server only

wardpeet (approved):

PR #14490, opened 2026-03-19
fix(datadog): use ddSpan.finish() to set correct span end times

### LINKED ISSUE #14474: [BUG] Datadog LLM Observability exporter: all spans share the same end time, breaking request timing metrics
## Describe the Bug

When using the Datadog LLM Observability exporter (`@mastra/observability-datadog`), all emitted spans appear to have the same end time. This breaks request timing metrics in Datadog, making it impossible to accurately measure span durations or identify performance bottlenecks.

In the Datadog trace view (see screenshot below), spans that should have different durations all terminate at the same point, resulting in incorrect timing data.

Datadog trace view showing spans with identical end times

### Root Cause (Suspected)

In `observability/datadog/src/tracing.ts` line 587, the `buildSpanOptions()` method falls back to `new Date()` when `span.endTime` is not set:

```typescript
const endTime = span.endTime ? toDate(span.endTime) : span.isEvent ? startTime : new Date();
```

When spans are exported in a batch, `new Date()` is called at roughly the same moment for all spans in that batch, giving them all the same end time — the export time rather than the actual span completion time. This causes all spans to appear to end simultaneously regardless of when they actually finished.

## Steps To Reproduce

1. Set up a Mastra project with the Datadog LLM Observability exporter
2. Configure an agent with tools (e.g., an orchestrator agent with OpenAI calls and tool invocations)
3. Run the agent and let it complete a multi-step interaction
4. View the resulting trace in Datadog LLM Observability
5. Observe that all spans share the same end time

_Investigating — will update with a minimal reproduction_

## Expected Behavior

Each span should have an accurate end time reflecting when that span actually completed. In the Datadog trace view, spans should show correct relative durations — e.g., an LLM call span should end before its parent agent span, and tool call spans should show their actual execution duration.

## Environment Information

- Package: `@mastra/observability-datadog`
- Relevant file: `observability/datadog/src/tracing.ts` (line 587, `buildSpanOptions` method)
- LLM Provider: OpenAI (gpt-5.4 observed in trace, but likely affects all providers)

## Verification

intojhanurag:
## Description
  
dd-trace's `llmobs.trace()` does not honor `endTime` in span options — it auto-finishes spans with the current wall-clock time when the synchronous callback returns. Since all spans in a trace tree are emitted in a single synchronous loop, they all receive the same end time, making duration metrics in Datadog LLM Observability inaccurate.
This fix calls `ddSpan.finish(endTimeMs)` explicitly inside the `llmobs.trace()` callback, using dd-trace's own API to set the correct end time before auto-finish kicks in. Once `finish()` is called, dd-trace's subsequent auto-finish is a no-op since `_duration` is already set.

Changes:
  - `buildSpanOptions()` now returns `endTimeMs` separately (not in trace options, since dd-trace ignores it)
  - `emitSpanTree()` and `emitSingleSpan()` call `ddSpan.finish(endTimeMs)` after annotations
  - Removed `endTime` from `LLMObsSpanOptions` interface (not part of dd-trace's API)
  - Updated mock and test to verify `finish()` is called with correct timestamp

## Before 
image

## After
Screenshot 2026-03-19 150228

  ## Related Issue(s)
  
  Fixes #14474

  ## Type of Change
  - [ ] Documentation update
  - [ ] Code refactoring
  - [ ] Performance improvement
  - [x] Test update

  ## Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed Datadog LLM Observability trace span duration accuracy. Spans now correctly report their individual end times, improving visibility into LLM calls and agent run timing in Datadog.

### DIFF
--- a/.changeset/fix-datadog-span-end-time.md
+++ b/.changeset/fix-datadog-span-end-time.md
@@ -1,0 +1,5 @@
+---
+'@mastra/datadog': patch
+---
+
+Fixed span duration accuracy in Datadog LLM Observability traces. Previously, all spans in a trace appeared to have the same end time because dd-trace's `llmobs.trace()` does not honor `endTime` in span options. Now uses `ddSpan.finish()` to explicitly set each span's correct end time, so tools, LLM calls, and agent runs show their actual durations in Datadog.
--- a/observability/datadog/src/tracing.test.ts
+++ b/observability/datadog/src/tracing.test.ts
@@ -51,6 +51,7 @@ const {
         id: `mock-dd-span-${parents.length}`,
         options,
         setTag: vi.fn(),
+        finish: vi.fn(),
       };
       spans.push(ddSpan);
       // Activate this span in scope for the duration of the callback
@@ -494,10 +495,13 @@ describe('DatadogExporter', () => {
       expect(mockTrace).toHaveBeenCalledWith(
         expect.objectContaining({
           startTime,
-          endTime: startTime,
         }),
         expect.any(Function),
       );
+
+      // endTime is set via ddSpan.finish() instead of trace options
+      const ddSpan = capturedSpans[0];
+      expect(ddSpan.finish).toHaveBeenCalledWith(startTime.getTime());
     });
 
     it('buffers event spans until parent exists and emits with parent context', async () => {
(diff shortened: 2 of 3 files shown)


epinzur (approved):
Looks great! Thanks for the quick fix.

PR #14526, opened 2026-03-19
fix(pg): set search_path before index creation and queries with custom schema

roaminro:
When PgVector is configured with a custom `schemaName` and the pgvector extension gets installed in that schema, `CREATE INDEX` and vector similarity queries fail because operator classes (`vector_cosine_ops`) and distance operators (`<=>`) can't be resolved — the custom schema isn't on PostgreSQL's default `search_path`.

The existing `search_path` logic only ran before `CREATE TABLE` and had a condition that skipped when the extension schema matched the table schema (which is the common case when using a custom schema).

This extracts an `ensureSearchPath()` helper that sets the path whenever the extension is in a non-default schema, and calls it in `createIndex()`, `setupIndex()`, and `query()`.

```ts
// before — only ran when schemas differed, skipping the common case
if (this.schema && this.vectorExtensionSchema && this.schema !== this.vectorExtensionSchema) {
  await client.query(`SET search_path TO ...`);
}

// after — runs whenever extension is in a non-default schema
private async ensureSearchPath(client) {
  if (this.vectorExtensionSchema && this.vectorExtensionSchema !== 'public' && this.vectorExtensionSchema !== 'pg_catalog') {
    await client.query(`SET search_path TO ...`);
  }
}
```

Fixes #14357

## Release Notes

* **Bug Fixes**
  * Fixed vector operations in PostgreSQL when `pgvector` extension is installed in a custom schema, ensuring correct operator and distance function resolution.

* **Tests**
  * Added test coverage for vector operations with custom schema configurations.


wardpeet (approved):

stores/pg/src/vector/index.ts
```diff
@@ -580,6 +608,9 @@ export class PgVector extends MastraVector<PGVectorFilter> {
     // Start a transaction
     const client = await this.pool.connect();
     try {
+      // Set search path so vector type casts (e.g. ::vector, ::halfvec) resolve correctly
+      await this.ensureSearchPath(client);
```
wardpeet, line 612: Isn't it enough to do it on the pool connection?

wardpeet: @coderabbitai what do you think?

PR #14528, opened 2026-03-19
feat(core): dynamic function support for workspace tool config [COR-435]

cmphilpot:
The `enabled`, `requireApproval`, and `requireReadBeforeWrite` options now accept async functions in addition to static booleans, enabling context-aware tool behavior like disabling tools based on user tier or requiring approval only for certain file paths.

## Type of Change

- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [ ] Test update

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Tool configuration fields (enabled, requireApproval, requireReadBeforeWrite) can be dynamic functions (sync or async) evaluated with runtime context.

* **Behavior**
  * enabled is resolved when listing tools; requireApproval and requireReadBeforeWrite are evaluated at execution time with access to call args and request context.

* **Documentation**
  * Added examples and guidance for context-aware tool configuration.

* **Tests**
  * Test suites updated to exercise async/dynamic tool initialization and evaluation.


packages/core/src/workspace/tools/tools.ts
```diff
@@ -127,11 +181,20 @@ function wrapWithReadTracker(
...
-              throw new FileReadRequiredError(input.path, check.reason!);
+          // Policy gate: require the agent to have read the file first.
+          // Only evaluate when explicitly configured (opt-in policy).
+          // Safe default true = fail-closed if a dynamic function throws.
+          if (config.requireReadBeforeWrite !== undefined) {
+            const shouldRequireRead = await resolveDynamicValue(
+              config.requireReadBeforeWrite,
+              { args: input, requestContext: context.requestContext ?? {}, workspace },
+              true,
+            );
```
NikAiyer, line 192: Should the safe default here be true? That might be the opposite of what we had before.

packages/server/src/server/constants.ts
```diff
@@ -63,22 +79,23 @@ export function resolveToolConfig(
 
   if (toolsConfig) {
     if (toolsConfig.enabled !== undefined) {
-      enabled = toolsConfig.enabled;
+      enabled = typeof toolsConfig.enabled === 'function' ? true : toolsConfig.enabled;
```
NikAiyer: This might be the opposite of the default in core. Can you double check that? We'll probably want to make sure these configs are consistent between core and server.

NikAiyer (commented):
LGTM overall, but we'll probably want to add a changeset for the server changes.

packages/server/src/server/constants.ts
NikAiyer: It looks like this might be the current setup between core and server:

```  ┌────────────────────────┬─────────────────────┬──────────────────┐
  │ Config                 │ Core default        │ Server default   │
  ├────────────────────────┼─────────────────────┼──────────────────┤
  │ enabled                │ false (fail-closed) │ true (fail-open) │
  ├────────────────────────┼─────────────────────┼──────────────────┤
  │ requireApproval        │ true                │ true             │
  ├────────────────────────┼─────────────────────┼──────────────────┤
  │ requireReadBeforeWrite │ true                │ true             │
  └────────────────────────┴─────────────────────┴──────────────────┘
  ```

File: packages/core/src/workspace/tools/tools.ts (hunk 1 above)
Comment (NikAiyer) [reply packages/core/src/workspace/tools/tools.ts:192]: The changes here might be fine, since we're most restrictive when something fails:

```
  ┌────────────────────────┬───────────┬─────────────────────┬────────────────────────┬─────────────────────────────────┐
  │                        │ main      │ PR (static boolean) │ PR (function succeeds) │ PR (function throws/no context) │
  ├────────────────────────┼───────────┼─────────────────────┼────────────────────────┼─────────────────────────────────┤
  │ enabled                │ true      │ uses value          │ uses returned value    │ false                           │
  ├────────────────────────┼───────────┼─────────────────────┼────────────────────────┼─────────────────────────────────┤
  │ requireApproval        │ false     │ uses value          │ uses returned value    │ true                            │
  ├────────────────────────┼───────────┼─────────────────────┼────────────────────────┼─────────────────────────────────┤
  │ requireReadBeforeWrite │ undefined │ uses value          │ uses returned value    │ true                            │
  └────────────────────────┴───────────┴─────────────────────┴────────────────────────┴─────────────────────────────────┘
  ```

packages/server/src/server/constants.ts
```diff
@@ -73,8 +73,8 @@ export function resolveToolConfig(
   toolsConfig: WorkspaceToolsConfig | undefined,
   toolName: WorkspaceToolName,
 ): { enabled: boolean; requireApproval: boolean; requireReadBeforeWrite?: boolean } {
-  let enabled = true;
-  let requireApproval = false;
+  let enabled = false;
+  let requireApproval = true;
```
NikAiyer: I think the base values themselves should remain unchanged, since the base defaults match core. But the part where we check the` typeof toolsConfig.[field[=== 'function'` might cause issues if that's hardcoded to true, especially if the enabled function itself throws or returns false in some cases. Could we support adding workspace and requestContext to this function? Then we can probably determine what the actual return value should be by calling the enabled function

.changeset/public-turtles-speak.md
```diff
@@ -0,0 +1,6 @@
+---
+'@mastra/server': patch
+'@mastra/core': patch
```
DanielSLew: can remove this, core is represented in the other changeset
```suggestion
```

.changeset/puny-clubs-push.md
```diff
@@ -0,0 +1,22 @@
...
+tools: {
+  [WORKSPACE_TOOLS.FILESYSTEM.WRITE_FILE]: {
+    requireApproval: async ({ args }) => {
+      return (args.path as string).startsWith('/protected')
+    },
+  },
+  [WORKSPACE_TOOLS.SANDBOX.EXECUTE_COMMAND]: {
+    enabled: async ({ requestContext }) => {
+      return requestContext.get('allowExecution') === 'true'
+    },
```
NikAiyer, line 19: I think the docs and this changeset don't match in terms of notations. This uses `.get` while the other uses bracket notions.

NikAiyer (approved):
Got a small nitpick, but not a blocker

PR #14533, opened 2026-03-20
fix(core): gracefully fallback search mode when requested mode not supported

### LINKED ISSUE #14531: Hybrid search requires both vector and BM25 configuration
This issue was created from Discord post 1484388184015241329:

[Open in Discord](https://discord.com/channels/1309558646228779139/1484388184015241329)

I have workspace configured with bm25 and no vector configuration but I'm getting following error when agent try to call mastra_workspace_search tool. Vector search should not be attempted if not configured
```
   error: {
      "message": "Hybrid search requires both vector and BM25 configuration.",
      "domain": "TOOL",
      "category": "USER",
      "code": "TOOL_EXECUTION_FAILED",
      "details": {
        "errorMessage": "Error: Hybrid search requires both vector and BM25 configuration.",
        "argsJson": "{\"query\":\"thread title\",\"topK\":5,\"mode\":\"hybrid\"}",
        "model": "gemini-3-flash-preview"
      },
      "cause": {
        "message": "Hybrid search requires both vector and BM25 configuration.",
        "name": "Error"
      }
    }
```

Aayush-engineer:
## Summary
Fixes #14531

## Problem
When the workspace search tool is called with `mode: 'hybrid'` but only 
BM25 is configured, it throws:
> Hybrid search requires both vector and BM25 configuration.

The agent has no way to know which modes are supported, so it should 
never throw — it should fall back gracefully.

## Fix
Two-layer fix:

**1. Root cause (`tools.ts`):** The search tool input schema now dynamically 
restricts available `mode` values to only what the workspace supports. 
The LLM can never pick an unsupported mode in the first place.

**2. Safety net (`search.ts`):** Runtime fallback when requested mode isn't supported:
- `hybrid` requested but not supported → falls back to `vector` or `bm25`
- `vector` requested but not supported → falls back to `bm25`

This matches the fallback logic already present in the server handler.

## Changes
- `packages/core/src/workspace/tools/search.ts` — runtime fallback safety net
- `packages/core/src/workspace/tools/__tests__/search.test.ts` — regression test
- `packages/core/src/workspace/tools/tools.ts` — dynamic enum restriction (root cause fix)
- `.changeset/fix-workspace-search-mode-fallback.md` — changeset

## Tests
- 1 new regression test added ✅
- 3/3 tests passing, 0 type errors ✅

## Summary by CodeRabbit

* **Bug Fixes**
  * Search now falls back to an available supported mode when the requested mode isn’t supported, preventing errors.

* **Improvements**
  * Search mode choices and reported mode now reflect the workspace’s actual supported backends.

* **Tests**
  * Added a test confirming fallback behavior returns results instead of throwing when a requested mode is unavailable.

* **Chores**
  * Patch release entry added for the workspace search behavior fix.


Aayush-engineer:
Fixed — when mode === 'vector' and vector is unsupported, it now always falls back to 'bm25' directly instead of the incorrect 'hybrid' path. Thanks for catching this!

abhiaiyer91 (approved):

PR #14574, opened 2026-03-21
fix: enable HNSW/IVFFlat index usage in PgVector queries

### LINKED ISSUE #14421: Mastra PG Vector generating suboptimal query
This issue was created from Discord post 1483545354119942246:

[Open in Discord](https://discord.com/channels/1309558646228779139/1483545354119942246)

Hi,
I am on latest version of mastra pg Vector. I use it as follows:
```
export async function searchCandidateIndex(
  queryVector: number[],
  options: CandidateIndexSearchOptions = {},
): Promise<CandidateIndexResult[]> {
  const pgVector = getPgVector();

  const filterConditions: VectorFilter = {};

  const whitelist = options.profileIdWhitelist?.filter(id => id.trim().length > 0);
  const blacklist = options.profileIdBlacklist?.filter(id => id.trim().length > 0);

  if (whitelist && whitelist.length > 0 && blacklist && blacklist.length > 0) {
    filterConditions.profile_id = {$in: whitelist, $nin: blacklist};
  } else if (whitelist && whitelist.length > 0) {
    filterConditions.profile_id = {$in: whitelist};
  } else if (blacklist && blacklist.length > 0) {
    filterConditions.profile_id = {$nin: blacklist};
  }

  const queryParams = {
    indexName: options.indexName ?? DEFAULT_INDEX_NAME,
    queryVector,
    topK: options.topK ?? 500,
    minScore: options.minScore ?? 0,
    filter: filterConditions,
  };

  const results = await pgVector.query({...queryParams});

  return results.map(result => ({
    id: result.id,
    score: result.score,
    metadata: (result.metadata ?? {}) as Record<string, unknown>,
  }));
}```

I am using this through supabase. Queries are taking 10-20 seconds. For context, my dataset is sized around 45k and I am trying to retrieve 5000 items. When I go into Supabase, this query pattern is flagged:
```
with
  vector_scores as (
    select
      vector_id as id,
      $3 - (embedding <=> $4::vector) as score,
      metadata
    from
      "candidate_index"
  )
select
  *
from
  vector_scores
where
  score > $1
order by
  score desc
limit
  $2```

This is the query patter generated by mastra pgvector. I believe to use my HNSW index the order by and limit need to be placed within the vector_scores expression? My understanding here is that the vector_scores expression is computing values for all entries in the table.

intojhanurag:
## Description
Move `ORDER BY` and `LIMIT` inside the CTE in PgVector similarity queries so PostgreSQL can use HNSW and IVFFlat indexes for approximate nearest neighbor search instead of scanning all rows.

  ## Related Issue(s)

  Fixes #14421

  ## Type of Change

  - [ ] Documentation update
  - [ ] Code refactoring
  - [x] Performance improvement
  - [ ] Test update

  ## Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * PostgreSQL vector similarity queries now properly leverage HNSW and IVFFlat indexes when no filters are applied, avoiding full-row scans and restoring faster approximate nearest-neighbor search behavior.
* **Tests**
  * Added tests to verify ORDER BY/LIMIT placement for filtered vs. unfiltered vector queries.
* **Chores**
  * Added release metadata for a patch release.


abhiaiyer91 (approved):

wardpeet (approved):

wardpeet:
@intojhanurag seems like pg is failing

PR #14578, opened 2026-03-22
fix: strip Anthropic thinking block signatures from LLM prompt history

### LINKED ISSUE #14559: [BUG] Harness drops thinking block signatures from thread history, causing API errors on replay
## Describe the Bug

When using Anthropic models with extended thinking enabled, the Harness `convertToHarnessMessage()` function converts stored `reasoning` parts to `{ type: "thinking", thinking: text }` objects but **drops the `providerOptions.anthropic.signature`** metadata. When these messages are replayed from thread history in subsequent API calls, the Anthropic API rejects them with:

```
Invalid request: messages.1.content.0: Invalid `signature` in `thinking` block
```

This is especially problematic when switching models mid-thread (e.g. from one Anthropic model to another, or from a non-Anthropic model to Anthropic), since thinking block signatures are cryptographic and model-specific. But even without model switching, the dropped signatures cause failures on the very next message in the same thread.

A secondary issue: if all content parts in an assistant message were thinking blocks, stripping them leaves an empty message which Anthropic also rejects with:

```
Invalid request: messages: text content blocks must be non-empty
```

## Steps To Reproduce

1. Create an agent with Anthropic model and extended thinking enabled
2. Send a message that triggers thinking (reasoning) in the response
3. The Harness stores the response to thread history via `convertToHarnessMessage()`, which drops `providerOptions.anthropic.signature` from reasoning parts
4. Send a follow-up message — the thread history is loaded and the previous thinking blocks are replayed without valid signatures
5. Anthropic API returns `Invalid signature in thinking block` error

Alternatively:
1. Start a thread with one Anthropic model (or any model that produces thinking blocks)
2. Switch to a different Anthropic model mid-thread
3. Send a message — old thinking blocks have signatures from the wrong model
4. Same error

No standalone repro repo — the bug is in the Harness internals (`convertToHarnessMessage`). The relevant code path is in the compiled harness where reasoning stream parts are converted to message content but `providerOptions` (containing the signature) is not carried through.

## Expected Behavior

Either:
1. **Preserve signatures**: `convertToHarnessMessage()` should carry `providerOptions.anthropic.signature` through to stored reasoning parts, so they can be replayed with the same model, OR
2. **Strip thinking blocks on retrieval**: When loading thread history, strip reasoning/thinking blocks since they are not needed for conversation continuity (the model's text response already contains the distilled reasoning)

We implemented option 2 as a workaround via `LanguageModelMiddleware.transformParams` that strips all `reasoning`/`thinking` content parts from assistant messages before sending to the API.

## Environment Information

```
@mastra/core: fork (based on recent main)
@mastra/memory: fork (based on recent main)
AI SDK: @ai-sdk/anthropic with interleaved-thinking-2025-05-14 beta header
LLM Provider: Anthropic (Claude models with extended thinking)
OS: macOS
Node: v22
```

## Verification

majiayu000:
## Description

Strip Anthropic reasoning parts (thinking blocks with cryptographic signatures) from LLM prompt history in `sanitizeV5UIMessages()`, matching the established OpenAI reasoning stripping pattern.

When Anthropic thinking blocks with `providerMetadata.anthropic.signature` are stored in thread history and later replayed as input to the Anthropic API, the signatures fail validation (they are cryptographic, model-specific, and ephemeral), causing `Invalid signature in thinking block` errors. If all parts in a message were thinking blocks, stripping them can leave an empty message causing `text content blocks must be non-empty` errors.

The fix adds `hasAnthropicReasoning` detection alongside the existing `hasOpenAIReasoning` check, and filters out Anthropic reasoning parts when `filterIncompleteToolCalls=true` (i.e., when building prompts TO the LLM). Reasoning data is preserved in the database and UI output — only stripped from LLM input. The existing empty-message removal (`if (!safeParts.length) return false`) handles the edge case where all parts were reasoning blocks.

## Related Issue(s)

Fixes #14559

## Type of Change

- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [x] Test update

## Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed "Invalid signature in thinking block" errors with Anthropic: signed reasoning blocks are now removed from prompts sent to the LLM, preventing replayed signatures from causing API errors while keeping those reasoning blocks visible in the UI and stored in the database, matching the existing behavior for other providers.

### DIFF
--- a/.changeset/early-crews-clap.md
+++ b/.changeset/early-crews-clap.md
@@ -1,0 +1,5 @@
+---
+'@mastra/core': patch
+---
+
+Fixed Anthropic thinking block signatures being replayed to the LLM, causing 'Invalid signature in thinking block' API errors. Anthropic reasoning parts with cryptographic signatures are now stripped from LLM prompts (matching the existing OpenAI reasoning stripping pattern), while being preserved in database storage and UI display.
--- a/packages/core/src/agent/message-list/conversion/output-converter-anthropic-reasoning.test.ts
+++ b/packages/core/src/agent/message-list/conversion/output-converter-anthropic-reasoning.test.ts
@@ -1,0 +1,112 @@
+import { describe, expect, it } from 'vitest';
+import type { AIV5Type } from '../types';
+import { sanitizeV5UIMessages } from './output-converter';
+
+/**
+ * Tests for Anthropic reasoning (thinking block) stripping in sanitizeV5UIMessages.
+ *
+ * Anthropic thinking blocks contain cryptographic signatures that are model-specific
+ * and ephemeral. Replaying them from thread history causes 'Invalid signature in
+ * thinking block' errors. These parts are stripped when building a prompt TO the LLM
+ * (filterIncompleteToolCalls=true) but preserved for UI display and DB storage.
+ *
+ * See: https://github.com/mastra-ai/mastra/issues/14559
+ */
+describe('sanitizeV5UIMessages — Anthropic reasoning stripping', () => {
+  const makeMessage = (parts: AIV5Type.UIMessage['parts']): AIV5Type.UIMessage => ({
+    id: 'msg-1',
+    role: 'assistant',
+    parts,
+  });
+
+  it('should strip Anthropic reasoning parts from LLM prompt', () => {
+    const msg = makeMessage([
+      {
+        type: 'reasoning',
+        reasoning: 'thinking...',
+        providerMetadata: {
+          anthropic: { signature: 'sig_abc123' },
+        },
+      } as AIV5Type.UIMessage['parts'][number],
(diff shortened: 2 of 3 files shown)


TylerBarnes (changes requested):
hmm this isn't quite right, for anthropic models we do want to send these parts back to maintain prompt caching. The approach should be totally different than this, we should store metadata about which provider made each request as message part metadata, then downstream (in harness, mastracode, new Agent, or elsewhere) a processor like `ModelHistoryCompatibility` can be written that conditionally does transforms like this based on if the currently active model doesn't match what was stored. Even if we don't store metadata like that, this should be a processor and not part of message list.

PR #14608, opened 2026-03-23
fix(deployer): discover transitive npm deps of workspace packages

roaminro:
Fixes #14557

In monorepo setups, `mastra dev` / `mastra build` would fail with:

> We couldn't load "@prisma/client" from "@eddi/forge"

This happened because `checkTransitiveDependencies()` in `analyzeEntry.ts` only kept transitive deps that were workspace packages — npm packages like `@prisma/client` were silently dropped. The fix includes them in the analysis results so the caller can externalize or install them at runtime. Workspace packages still trigger further recursive discovery; non-workspace packages don't.

Added a test with a fixture that reproduces the scenario (app → workspace pkg → npm pkg).

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed bundle analysis to correctly discover transitive npm dependencies in monorepo workspaces. This resolves "couldn't load" errors during `mastra dev` and `mastra build` operations.

* **Tests**
  * Added test coverage for transitive dependency discovery in workspace scenarios.


roaminro:
Closing as superseded — the bug this PR addressed was fixed on main by #18879 and #19808. `checkTransitiveDependencies()` no longer does a recursive Rollup walk; workspace manifests are read directly from `workspaceMap`, and npm deps of workspace packages now surface through the `bundleExternals` output import scan. The scenario is pinned by `e2e-tests/monorepo` (`@inner/inner-tools` → `bcrypt`). Thanks for the review!

PR #14643, opened 2026-03-24
feat(observability): add @mastra/arthur exporter and extract @mastra/openinference

videetparekh:
## Summary

- **New `@mastra/arthur` package**: Arthur AI observability provider for Mastra. Extends `OtelExporter` with Arthur-specific config (`ARTHUR_API_KEY`, `ARTHUR_BASE_URL`, `ARTHUR_TASK_ID` env vars, Bearer auth, `/api/v1/traces` endpoint).
- **New `@mastra/openinference` package**: Extracts the `OpenInferenceOTLPTraceExporter` (Mastra → OpenInference span conversion) from `@mastra/arize` into a shared package so any OpenInference-compatible sink can reuse it.
- **Updated `@mastra/arize`**: Imports `OpenInferenceOTLPTraceExporter` from `@mastra/openinference` instead of its local copy.

### Problem

[Arthur AI](https://arthur.ai) is an OpenInference-compatible telemetry sink. Mastra users who want to send traces to Arthur today have to use a vendored copy of our custom exporter in every application because there's no native Mastra integration. Mastra provides out-of-the-box OTel support with an existing OpenInference translation layer (used by `@mastra/arize`), and we want to leverage that to reduce our customers' overhead.

Fixes #14600 

### Usage

```typescript
import { ArthurExporter } from '@mastra/arthur';

const mastra = new Mastra({
  observability: {
    configs: {
      arthur: {
        serviceName: 'my-service',
        exporters: [
          new ArthurExporter({
            apiKey: process.env.ARTHUR_API_KEY,
            endpoint: process.env.ARTHUR_BASE_URL,
            taskId: process.env.ARTHUR_TASK_ID, // optional
          }),
        ],
      },
    },
  },
});
```

Or zero-config with environment variables:

```bash
ARTHUR_API_KEY=your-key
ARTHUR_BASE_URL=https://app.arthur.ai
ARTHUR_TASK_ID=your-task-id  # optional
```

```typescript
new ArthurExporter()
```

## Test plan

- [x] `@mastra/arthur` — 13 unit tests covering config resolution, env var fallbacks, endpoint formatting, auth headers, taskId resource attributes, and missing-config warnings
- [x] `@mastra/openinference` — builds successfully, all packages depending on it build cleanly
- [x] `@mastra/arize` — builds successfully with refactored import from `@mastra/openinference`
- [x] Live tested: sent a real span from `ArthurExporter` to Arthur's development environment and verified it appeared in the trace viewer

## Summary by CodeRabbit

* **New Features**
  * Added Arthur AI exporter (API key, endpoint, optional task ID and metadata) for sending traces
  * Added OpenInference package with a shared OTLP trace exporter and semantic conventions

* **Documentation**
  * Added full setup guide and examples for the Arthur AI exporter

* **Bug Fixes**
  * Fixed token-counting to reliably compute total tokens when inputs are non-numeric

* **Tests**
  * Added unit tests covering exporter configuration and span/attribute mapping

* **Chores**
  * Updated observability manifests, build/test configs, and dependency wiring for new packages


epinzur:
sorry I haven't been able to look at this yet.  I'll do the review early next week.

epinzur (changes requested):
## What this PR does

1. **New `@mastra/arthur` package** — An `OtelExporter` subclass that sends Mastra traces to Arthur AI using OpenInference semantic conventions. Supports zero-config via environment variables (`ARTHUR_API_KEY`, `ARTHUR_BASE_URL`, `ARTHUR_TASK_ID`).

2. **New `@mastra/openinference` package** — Extracts the `OpenInferenceOTLPTraceExporter` class (previously internal to `@mastra/arize`) into a shared package so both Arize and Arthur can reuse it.

3. **Refactors `@mastra/arize`** — Swaps the internal `OpenInferenceOTLPTraceExporter` for an import from `@mastra/openinference`. Removes now-unnecessary direct dependencies on `@arizeai/openinference-genai` and `@opentelemetry/exporter-trace-otlp-proto`.

## What looks good

- **Arthur exporter implementation is clean.** It follows the same patterns as the Arize exporter — env var fallback, `setDisabled()` on missing credentials, consistent config types. Easy to review and understand.

- **Good test coverage.** 13 unit tests for Arthur covering config precedence, disabled states, header merging, resource attributes, and taskId warnings. The OpenInference exporter also gets a standalone 340-line test suite covering span kind mapping, token count conversion, metadata handling, and input/output extraction.

- **README is solid.** Shows both zero-config and explicit setup, optional configuration, and custom metadata usage.

- **JSDoc additions are helpful.** The class and constructor are well-documented with `@example` blocks.

## Concerns

### 1. Premature extraction of `@mastra/openinference`

The PR extracts `OpenInferenceOTLPTraceExporter` into a shared package, but there are only two consumers today: `@mastra/arize` and `@mastra/arthur`. Both happen to use OpenInference conventions because Arthur accepts them — but that's an implementation detail, not a signal that this needs to be a shared abstraction yet.

Additionally `@mastra/openinference` is un-usable on its own.  We currently don't publish any packages like this.

A simpler alternative: Arthur could depend on the `@arizeai/openinference-*` packages directly (like Arize originally did) and have its own copy of the exporter. The duplication is small and contained.

### 2. Protocol lock-in via `@opentelemetry/exporter-trace-otlp-proto`

`@mastra/openinference` has a hard dependency on `@opentelemetry/exporter-trace-otlp-proto` because `OpenInferenceOTLPTraceExporter` extends `OTLPTraceExporter` from that package. This locks the "shared" package to the protobuf transport.

If the package is meant to enable reuse across multiple telemetry sinks, different backends may need different transports (HTTP/JSON, gRPC). As-is, any consumer pulls in the protobuf exporter whether they want it or not.

The actual value of the class is the `export()` method that transforms Mastra span attributes into OpenInference semantic conventions. That logic is entirely transport-independent — it just maps `ReadableSpan` objects. A cleaner design would separate the span transformation from the transport, letting consumers compose it with whatever OTLP exporter they need.

### 5. Missing docs

I'd expect to see an entry in docs/src/content/en/docs/observability/tracing/exporters and docs/src/content/en/reference/observability/tracing/exporters for the new package. And add it to the sidebars.

## Minor notes

- The changeset `warm-keys-take.md` bundles `@mastra/arthur` (minor), `@mastra/openinference` (minor), and `@mastra/arize` (patch) together. Given the arize change is a refactor and arthur is a new package, these should be separate changesets for cleaner changelogs.

## Suggestion


1. Ship `@mastra/arthur` as a standalone package that depends on `@arizeai/openinference-*` directly and contains its own copy of the OTLP exporter or just depend on `@mastra/arize`.
2. Skip the `@mastra/openinference` extraction for now.
3. Leave `@mastra/arize` unchanged.

If OpenInference reuse becomes a real need (a third consumer, or transport flexibility requirements), extract at that point with a transport-agnostic design.

epinzur (approved):

epinzur (approved):

PR #14648, opened 2026-03-24
fix: split client/provider tool blocks to fix Anthropic ordering error

CalebBarnes:
## Problem

When Anthropic returns `tool_use` (client) before `server_tool_use` (provider) in the same turn, and the provider result is inlined in the assistant message, the client `tool_result` doesn't immediately follow its `tool_use`. Anthropic rejects this with:

> `tool_use ids were found without tool_result blocks immediately after`

This happens non-deterministically when using web_search alongside client tools (e.g. find_files, execute_command). Once it occurs, the thread becomes unrecoverable.

## Root cause

`convertToModelMessages` groups consecutive tool parts into a single assistant message block. When a client tool precedes a provider tool, the resulting structure is:

```
assistant: [tool_use(client), server_tool_use(provider), web_search_tool_result(provider)]
user:      [tool_result(client)]
```

Anthropic requires `tool_result(client)` to **immediately follow** `tool_use(client)`, but the inlined provider result sits between them.

## Fix

Insert a `step-start` in `addStartStepPartsForAIV5` between client tool parts and completed provider-executed tool parts. This forces them into separate message blocks:

```
assistant: [tool_use(client)]
user:      [tool_result(client)]
assistant: [server_tool_use(provider), web_search_tool_result(provider)]
```

The reverse order (provider before client) is already safe and is not split.

## Tests

- 6 unit tests for `addStartStepPartsForAIV5` splitting logic (fail without fix)
- 1 e2e test with recorded two-turn flow (verified against real Anthropic API: 400 without fix, passes with fix)

## Test plan

```bash
cd packages/core
pnpm vitest run src/agent/message-list/conversion/output-converter-provider-executed.test.ts
pnpm vitest run src/tools/provider-tools-ordering.e2e.test.ts
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed a critical issue where parallel execution of client and provider tools would cause Anthropic integration to reject requests, making conversation threads unrecoverable. Tool message handling has been improved to ensure proper sequencing compatibility with Anthropic's requirements.

### DIFF
--- a/packages/core/src/agent/message-list/conversion/output-converter.ts
+++ b/packages/core/src/agent/message-list/conversion/output-converter.ts
@@ -195,6 +195,23 @@ export function addStartStepPartsForAIV5(messages: AIV5Type.UIMessage[]): AIV5Ty
       if (nextPart && nextPart.type !== `step-start` && !AIV5.isToolUIPart(nextPart)) {
         message.parts.splice(index + 1, 0, { type: 'step-start' });
       }
+
+      // Split client tools from completed provider-executed tools.
+      // Anthropic requires tool_result to immediately follow tool_use. When a client tool_use and
+      // a server_tool_use (with inline result) are in the same block, convertToModelMessages produces:
+      //   assistant: [tool_use(client), server_tool_use(provider), tool_result(provider)]
+      //   user:      [tool_result(client)]
+      // Anthropic rejects this because tool_result(client) doesn't immediately follow tool_use(client).
+      // Splitting them into separate blocks fixes the ordering.
+      if (
+        nextPart &&
+        AIV5.isToolUIPart(nextPart) &&
+        !part.providerExecuted &&
+        nextPart.providerExecuted &&
+        (nextPart.state === 'output-available' || nextPart.state === 'output-error')
+      ) {
+        message.parts.splice(index + 1, 0, { type: 'step-start' });
+      }
     }
   }
   return messages;
--- a/.changeset/swift-dogs-act.md
+++ b/.changeset/swift-dogs-act.md
@@ -1,0 +1,5 @@
+---
+'@mastra/core': patch
+---
+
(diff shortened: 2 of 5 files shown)


packages/core/src/agent/message-list/conversion/output-converter.ts
```diff
@@ -195,6 +195,23 @@ export function addStartStepPartsForAIV5(messages: AIV5Type.UIMessage[]): AIV5Ty
...
+      // a server_tool_use (with inline result) are in the same block, convertToModelMessages produces:
+      //   assistant: [tool_use(client), server_tool_use(provider), tool_result(provider)]
+      //   user:      [tool_result(client)]
+      // Anthropic rejects this because tool_result(client) doesn't immediately follow tool_use(client).
+      // Splitting them into separate blocks fixes the ordering.
+      if (
+        nextPart &&
+        AIV5.isToolUIPart(nextPart) &&
+        !part.providerExecuted &&
+        nextPart.providerExecuted &&
```
TylerBarnes, line 210: what about the opposite case where it's a provider tool and the next one is not?

CalebBarnes: The opposite case (provider → client) is already safe. It produces:

```
assistant: [server_tool_use(web_search), web_search_tool_result, tool_use(find_files)]
user:      [tool_result(find_files)]
```

`tool_use(find_files)` is the last block in the assistant message so its `tool_result` in the next user message satisfies Anthropic's "immediately after" rule. This is the order our existing test recordings happened to use, which is why prior tests passed.

The bug only triggers the other way around — `[tool_use(client), server_tool_use, server_tool_result]` — because the client's `tool_result` in the user message no longer immediately follows its `tool_use`.

There's also an explicit unit test for the safe order: `should NOT split provider tool followed by client tool (safe order)`.

TylerBarnes: cool, thanks!

TylerBarnes (approved):

PR #14652, opened 2026-03-24
fix: insert step-start between consecutive loop iterations

### LINKED ISSUE #14124: [BUG] Consecutive tool-only turns are incorrectly merged into one

CalebBarnes:
## Problem

When the agentic loop runs multiple iterations that each produce only tool calls, they get merged into a single assistant message with no boundary. The LLM then misinterprets sequential tool calls as parallel calls from a single turn.

This was reported in [#14124](https://github.com/mastra-ai/mastra/issues/14124).

## Root cause

`MessageMerger` merges consecutive assistant messages that share the same `messageId`. Since `messageId` is generated once per `agent.stream()` call and reused across all loop iterations, tool-only turns from different iterations collapse into a single message block.

## Fix

- Add `messageList.stepStart()` — finds the most recent unsealed assistant message and appends a `{ type: 'step-start' }` part, handling source migration for persistence.
- Call `messageList.stepStart()` at the start of each loop iteration after the first in `llm-execution-step.ts`.

This ensures `convertToModelMessages` splits them into separate blocks, so the LLM sees sequential steps rather than parallel tool calls.

## Changes

- `message-list.ts`: Added `stepStart()` method
- `llm-execution-step.ts`: Call `stepStart()` when `currentIteration > 1`
- `tools.ts`: Test with 3-iteration mock (tool → tool → text) verifying `step-start` boundary
- `options.ts`: Updated inline snapshot

## Test plan

```bash
cd packages/core
pnpm vitest run src/loop/loop.test.ts -t "step-start"
pnpm vitest run src/agent/__tests__/supervisor-integration.test.ts
pnpm vitest run --project unit:packages/core  # full suite: 5696 passed
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented consecutive tool-only iterations from being merged into a single assistant response by inserting explicit step-start boundaries, preserving sequential tool-call ordering.

* **Tests**
  * Added tests to verify step-start boundaries are inserted between consecutive tool-only iterations and that insertion is idempotent and respects sealed/invalid states.

* **Chores**
  * Added release metadata for a patch.

### DIFF
--- a/packages/core/src/loop/workflows/agentic-execution/llm-execution-step.ts
+++ b/packages/core/src/loop/workflows/agentic-execution/llm-execution-step.ts
@@ -642,6 +642,14 @@ export function createLLMExecutionStep<TOOLS extends ToolSet = ToolSet, OUTPUT =
     execute: async ({ inputData, bail, tracingContext }) => {
       currentIteration++;
 
+      // Insert a step-start boundary between loop iterations so that
+      // consecutive tool-only turns are not collapsed into a single block
+      // by convertToModelMessages. This ensures the LLM sees them as
+      // sequential steps rather than parallel tool calls.
+      if (currentIteration > 1) {
+        messageList.stepStart();
+      }
+
       let currentMessageId = inputData.isTaskCompleteCheckFailed
         ? `${messageIdPassed}-${currentIteration}`
         : inputData.messageId || messageIdPassed;
--- a/.changeset/loud-turtles-fly.md
+++ b/.changeset/loud-turtles-fly.md
@@ -1,0 +1,5 @@
+---
+'@mastra/core': patch
+---
+
+Fix consecutive tool-only loop iterations being merged into a single assistant message block. When the agentic loop runs multiple iterations that each produce only tool calls, the LLM would misinterpret them as parallel calls from a single turn. A `step-start` boundary is now inserted between iterations to ensure they are treated as sequential steps.
--- a/packages/core/src/agent/message-list/message-list.ts
+++ b/packages/core/src/agent/message-list/message-list.ts
@@ -699,25 +699,30 @@ export class MessageList {
    * source so the updated content is re-saved.
    */
   public stepStart(): boolean {
-    for (let m = this.messages.length - 1; m >= 0; m--) {
-      const msg = this.messages[m]!;
-      if (msg.role !== 'assistant' || !msg.content?.parts) continue;
-
-      if (MessageMerger.isSealed(msg)) {
-        return false;
-      }
-
-      msg.content.parts.push({ type: 'step-start' as const });
-
-      // Ensure the mutated message is persisted
-      if (!this.stateManager.isResponseMessage(msg)) {
-        this.stateManager.removeMessage(msg);
-        this.stateManager.addToSource(msg, 'response');
-      }
-
-      return true;
+    const lastMsg = this.messages[this.messages.length - 1];
+    if (!lastMsg || lastMsg.role !== 'assistant' || !lastMsg.content?.parts) {
+      return false;
     }
-    return false;
+
+    if (MessageMerger.isSealed(lastMsg)) {
+      return false;
+    }
+
+    // Don't add a duplicate step-start
+    const lastPart = lastMsg.content.parts[lastMsg.content.parts.length - 1];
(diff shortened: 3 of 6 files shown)


TylerBarnes (approved):
hell yea! 🔥

packages/core/src/loop/workflows/agentic-execution/llm-execution-step.ts
```diff
@@ -642,6 +642,14 @@ export function createLLMExecutionStep<TOOLS extends ToolSet = ToolSet, OUTPUT =
     execute: async ({ inputData, bail, tracingContext }) => {
       currentIteration++;
 
+      // Insert a step-start boundary between loop iterations so that
+      // consecutive tool-only turns are not collapsed into a single block
+      // by convertToModelMessages. This ensures the LLM sees them as
+      // sequential steps rather than parallel tool calls.
+      if (currentIteration > 1) {
```
TylerBarnes, line 649: oh is this 1 indexed? I think first step is step 0 right?

TylerBarnes: nvm lol just had to expand the gh diff one time above haha

CalebBarnes: Yeah I think it starts at 0 `let currentIteration = 0;` but then this check is right after calling `currentIteration++` so it should be at least 1

PR #14653, opened 2026-03-24
fit: catch errors from persistMessages in runAsyncBufferedObservation

### LINKED ISSUE #14651: [BUG] ObservationalMemory can produce unhandled promise rejection
### Describe the Bug

`startAsyncBufferedObservation` in `ObservationalMemory` stores a detached promise (asyncOp) in a static Map with .finally() but no .catch(). If the underlying storage call fails, the promise rejects without a handler, producing an unhandled promise rejection.

The bug is masked during multi-step agent execution because a subsequent `runAsyncBufferedObservation` call accidentally catches the previous rejection via `try { await existingOp } catch {}`. However, when the failing `asyncOp` is the last one, meaning no future step creates a new `runAsyncBufferedObservation` to consume it, the rejection is unhandled.

## When the bug manifests
The rejection goes unhandled whenever the rejected `asyncOp` is the last one for that buffer key.

```
startAsyncBufferedObservation(record, threadId, ...) {
    // ...
    const asyncOp = this.runAsyncBufferedObservation(
      record, threadId, unobservedMessages, bufferKey, writer, requestContext
    ).finally(() => {
      ObservationalMemory.asyncBufferingOps.delete(bufferKey);
      unregisterOp(record.id, 'bufferingObservation');
      // ...
    });
    ObservationalMemory.asyncBufferingOps.set(bufferKey, asyncOp);  // ❌ no .catch()
}
```

The promise is stored but never has a `.catch()` attached.

Inside `runAsyncBufferedObservation`, the `persistMessages` call sits outside the try/catch block:

```
async runAsyncBufferedObservation(...) {
    const existingOp = ObservationalMemory.asyncBufferingOps.get(bufferKey);
    if (existingOp) {
      try {
        await existingOp;    // ← accidentally catches PREVIOUS asyncOp's rejection
      } catch {}              //   (mutex-wait pattern, not intentional error handling)
    }
    // ...
    
    // ❌ UNPROTECTED — no try/catch
    await this.messageHistory.persistMessages({ messages, threadId, resourceId });
    
    try {
      // ✅ This IS protected
      await this.doAsyncBufferedObservation(...);
    } catch (error) {
      // handles failure markers, logs with omError
    }
}
```

## Why it's masked during normal multi-step execution
The `try { await existingOp } catch {}` at the top of runAsyncBufferedObservation waits for the previous buffering op to finish before starting a new one. As a side effect, it catches any rejection from the previous `asyncOp`. So during a multi-step agent run where step N+1 triggers another startAsyncBufferedObservation, the rejection from step N is consumed. But when the failing op is the last one, no subsequent call exists to catch it.

### Steps To Reproduce

Described here: https://github.com/trevor-zablocki-anrok/mastra/tree/unhandled-persist-messages-promise

https://github.com/trevor-zablocki-anrok/mastra/tree/unhandled-persist-messages-promise

### Expected Behavior

I would expect failures to persist messages would be caught and logged, then runAsyncBufferedObservation would return.

### Environment Information

```block
System:
    OS: macOS 15.7.4
    CPU: (14) arm64 Apple M4 Pro
    Shell: 5.9
  Binaries:
    Node: 22.20.0
    npm: 10.9.3
  npmPackages:
    @mastra/core: ^1.15.0 => 1.15.0 
    @mastra/evals: ^1.1.2 => 1.1.2 
    @mastra/libsql: ^1.7.1 => 1.7.1 
    @mastra/loggers: ^1.0.3 => 1.0.3 
    @mastra/memory: ^1.9.0 => 1.9.0 
    @mastra/observability: ^1.5.1 => 1.5.1 
    @types/node: ^25.5.0 => 25.5.0 
    mastra: ^1.3.14 => 1.3.14 
    typescript: ^6.0.2 => 6.0.2 
    zod: ^4.3.6 => 4.3.6 
  npmGlobalPackages:
    corepack: 0.34.0
    npm: 10.9.3
```

### Verification

trevor-zablocki-anrok:
## Description

The `persistMessages` call in `runAsyncBufferedObservation` that saves sealed messages to storage had no error handling. Since this method runs as a background async operation, a storage failure would surface as an unhandled promise rejection. This wraps the call in a try/catch with an early return so the method logs the error and bails out before registering sealed IDs or starting the observation cycle, preventing a state inconsistency where in-memory state diverges from storage. Includes a regression test verifying that neither sealed ID tracking nor the observation cycle runs when the persist fails.

## Related Issue(s)

Fixes #14651 

## Type of Change

- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [x] Test update

## Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced error handling for message persistence failures. The system now gracefully exits and prevents subsequent operations when message persistence encounters errors, ensuring data integrity.

### DIFF
--- a/packages/memory/src/processors/observational-memory/__tests__/observational-memory.test.ts
+++ b/packages/memory/src/processors/observational-memory/__tests__/observational-memory.test.ts
@@ -2,7 +2,7 @@ import { MockLanguageModelV2 } from '@internal/ai-sdk-v5/test';
 import type { MastraDBMessage, MastraMessageContentV2 } from '@mastra/core/agent';
 import { coreFeatures } from '@mastra/core/features';
 import { InMemoryMemory, InMemoryDB } from '@mastra/core/storage';
-import { describe, it, expect, beforeEach, afterEach } from 'vitest';
+import { describe, it, expect, beforeEach, afterEach, vi } from 'vitest';
 
 import { injectAnchorIds, parseAnchorId, stripEphemeralAnchorIds } from '../anchor-ids';
 import { ModelByInputTokens } from '../model-by-input-tokens';
@@ -6600,6 +6600,46 @@ describe('Async Buffering Processor Logic', () => {
       // Should not throw
       (om as any).sealMessagesForBuffering([msg]);
       expect(msg.content.metadata).toBeUndefined();
+    });
+  });
+
+  describe('runAsyncBufferedObservation — persist failure', () => {
+    afterEach(() => {
+      vi.restoreAllMocks();
+      (ObservationalMemory as any).sealedMessageIds.clear();
+    });
+
+    it('should return early and not track sealed IDs when persistMessages fails', async () => {
+      const storage = createInMemoryStorage();
+      const om = new ObservationalMemory({
+        storage,
+        scope: 'thread',
+        model: createStreamCapableMockModel({ defaultObjectGenerationMode: 'json' }),
+        observation: {
+          messageTokens: 50000,
+          bufferTokens: 10000,
+        },
+        reflection: { observationTokens: 20000 },
+      });
+
+      const threadId = 'thread-1';
+      await storage.initializeObservationalMemory({
+        threadId,
+        resourceId: 'resource-1',
+        scope: 'thread',
+        config: {},
+      });
+      const record = await storage.getObservationalMemory(threadId, 'resource-1');
+
+      const messages = createTestMessages(20);
+
+      vi.spyOn((om as any).messageHistory, 'persistMessages').mockRejectedValue(new Error('Storage write failed'));
+      const doBufferedSpy = vi.spyOn(om as any, 'doAsyncBufferedObservation');
+
+      await (om as any).runAsyncBufferedObservation(record, threadId, messages, `obs:thread:${threadId}`);
+
+      expect((ObservationalMemory as any).sealedMessageIds.has(threadId)).toBe(false);
+      expect(doBufferedSpy).not.toHaveBeenCalled();
     });
   });
 
--- a/packages/memory/src/processors/observational-memory/observational-memory.ts
+++ b/packages/memory/src/processors/observational-memory/observational-memory.ts
@@ -4863,11 +4863,16 @@ ${formattedMessages}
     // 2. When MessageList creates new messages for streaming content after the seal,
     //    those new messages have their own IDs and don't overwrite the sealed messages
     // 3. The sealed messages remain intact with their content at the time of buffering
-    await this.messageHistory.persistMessages({
-      messages: messagesToBuffer,
-      threadId,
-      resourceId: freshRecord.resourceId ?? undefined,
-    });
+    try {
+      await this.messageHistory.persistMessages({
+        messages: messagesToBuffer,
+        threadId,
+        resourceId: freshRecord.resourceId ?? undefined,
+      });
+    } catch (error) {
+      omError('[OM] Failed to persist messages for buffering', error);
+      return;
+    }
 
     // Track sealed message IDs in the static map so saveMessagesWithSealedIdTracking
     // generates new IDs for any future saves of these messages.


packages/memory/src/processors/observational-memory/__tests__/observational-memory.test.ts
```diff
@@ -6603,6 +6603,46 @@ describe('Async Buffering Processor Logic', () => {
...
+
+      const messages = createTestMessages(20);
+
+      vi.spyOn((om as any).messageHistory, 'persistMessages').mockRejectedValue(new Error('Storage write failed'));
+      const doBufferedSpy = vi.spyOn(om as any, 'doAsyncBufferedObservation');
+
+      await (om as any).runAsyncBufferedObservation(record, threadId, messages, `obs:thread:${threadId}`);
+
+      expect((ObservationalMemory as any).sealedMessageIds.has(threadId)).toBe(false);
+      expect(doBufferedSpy).not.toHaveBeenCalled();
```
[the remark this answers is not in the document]
trevor-zablocki-anrok: The sealed messages are keyed with the thread ID, I think.

packages/memory/src/processors/observational-memory/observational-memory.ts
```diff
@@ -4863,11 +4863,16 @@ ${formattedMessages}
...
-      resourceId: freshRecord.resourceId ?? undefined,
-    });
+    try {
+      await this.messageHistory.persistMessages({
+        messages: messagesToBuffer,
+        threadId,
+        resourceId: freshRecord.resourceId ?? undefined,
+      });
+    } catch (error) {
+      omError('[OM] Failed to persist messages for buffering', error);
```
TylerBarnes, line 4873: omError here just logs to a debug log when debugging is enabled, we should reject the whole request using tripWire instead

TylerBarnes: cause the new change is almost worse, it will silently swallow any error here

TylerBarnes (changes requested):

packages/memory/src/processors/observational-memory/observational-memory.ts
trevor-zablocki-anrok: Should the entire request be rejected if this fails, though? It would cause the observation to fail, but it could be handled on the next cycle. Tripwire isn't currently available here and with it being run in the background I would think we don't want it to block the request.

TylerBarnes: this is saving the regular thread messages, not observations. We could add a retry here or something, but if the messages don't save, the rest of the system probably wont work anyway

trevor-zablocki-anrok: Yeah, I just meant that they are only being saved here for the sake of observations and they will be saved elsewhere in the code, in a location where failures will be handled properly from what I see.

Since the code is fire and forget, it complicates things a little and I'm not sure how TripWire could be used here or how it would would eventually get handled. Would it be appropriate to write a `data-om-` chunk when this fails?

trevor-zablocki-anrok: Functionally it is no different than all of the returns that occur before this one in this method, the buffered operations will just not happen. As far as I can tell, the only side effect would be the messages being marked as sealed with `this.sealMessagesForBuffering(messagesToBuffer);`, which should be harmless.

TylerBarnes: where it's different is if it continues failing each time, there will be a (potentially long) history which gets lost when the agent finishes. Failing early is better in that case so it can be fixed, and either way we never want to silently fail. `omError` here is more of a local debugging util that just calls `omDebug`

TylerBarnes:
From Mastra Code: Thanks for taking a look at this. I’m going to close this PR since it’s stale and I don’t think this is the right direction for the failure mode.

The main concern is that this changes regular thread message persistence failures into a swallowed/log-only path. If `persistMessages` keeps failing, the agent can appear to continue normally while the conversation history is silently not being saved. That’s worse than surfacing the failure, especially because `omError` only goes to debug logging and most users won’t see it.

If we revisit this, I’d prefer a design that distinguishes observation-memory background failures from primary thread message persistence failures, and only suppresses errors where losing the write is genuinely non-critical.

PR #14687, opened 2026-03-25
fix(core): use API token for Cloudflare Workers AI

TheIsrael1:
## Summary
This fixes the Cloudflare Workers AI provider metadata generated from `models.dev`.

`models.dev` currently lists `CLOUDFLARE_ACCOUNT_ID` first in the provider env var list, and Mastra treats the first env var as the auth env var during registry generation. That causes the generated provider registry to use the account ID as the API key.

This change adds a provider-specific override for `cloudflare-workers-ai` so generation keeps using `CLOUDFLARE_ACCOUNT_ID` in the URL path, but uses `CLOUDFLARE_API_TOKEN` for authentication.

Fixes #14510

## Changes
- add a `cloudflare-workers-ai` override in `ModelsDevGateway`
- regenerate the checked-in provider registry and generated Cloudflare provider doc
- add a regression test for the override in `models-dev.test.ts`

## Validation
- `pnpm --filter @mastra/core generate:model-router`
- `pnpm --filter @mastra/core exec vitest run src/llm/model/gateways/models-dev.test.ts`

## Summary by CodeRabbit

* **Bug Fixes**
  * Cloudflare Workers AI auth now prefers API-key style environment variables over account identifiers when multiple auth options exist.

* **Documentation**
  * Provider guide updated to show both account ID and API key examples and now lists all required env vars with clearer placeholders extracted from URL templates.

* **Tests**
  * Added tests covering env-var selection priority and placeholder-preserving URL behavior.


packages/core/src/llm/model/gateways/models-dev.ts
```diff
@@ -43,6 +43,11 @@ const PROVIDER_OVERRIDES: Record<string, Partial<ProviderConfig>> = {
   groq: {
     url: 'https://api.groq.com/openai/v1',
   },
+  'cloudflare-workers-ai': {
+    // models.dev lists CLOUDFLARE_ACCOUNT_ID first, but that belongs in the URL path.
+    // Cloudflare Workers AI auth should use the API token env var instead.
+    apiKeyEnvVar: 'CLOUDFLARE_API_TOKEN',
```
TylerBarnes: I wonder if instead of doing an override we should sort any array that has multiple env var names by any that end in `_API_TOKEN` first, because then it'll work for any other future providers

TheIsrael1: Makes sense. I switched it to use models.dev as the source of truth and just made our env selection
 smarter, so we now prefer auth-like env vars instead of assuming env[0] is always the credential.

wardpeet (approved):

TylerBarnes (approved):

PR #14727, opened 2026-03-26
fix(mastracode): fix custom slash commands not loading

### LINKED ISSUE #14719: [BUG] [Mastra Code] Custom slash commands not loaded from .claude/commands/ or .mastracode/commands/; skills not loaded from .claude/skills/
## Description

Mastra Code does not load custom slash commands from either `.claude/commands/` or `.mastracode/commands/`. Typing `/` in the TUI does not show any project-defined commands.

Additionally, skills defined in `.claude/skills/` are not loaded by the agent, while skills in `.mastracode/skills/` work correctly. The `--version` output confirms all four skill directories are scanned, but only `.mastracode/skills/` actually works.

## Expected Behavior

Per the [configuration docs](https://code.mastra.ai/configuration#custom-slash-commands):

1. Custom slash commands in `.mastracode/commands/` and `.claude/commands/` should be discoverable via `/` in the TUI
2. Skills in `.claude/skills/` should be loaded with the same behavior as `.mastracode/skills/`

### Commands
1. Create `.mastracode/commands/review.md` (or `.claude/commands/review.md`) with valid frontmatter:
   ```yaml
   ---
   name: review
   description: Review the current diff
   ---
   Review my current git diff.
   ```
2. Start Mastra Code
3. Type `/` — the custom command does not appear

### Skills
1. Place a valid skill directory in `.claude/skills/my-skill/SKILL.md`
2. Start Mastra Code — skill is listed in startup output but not activated by the agent
3. Move the same skill to `.mastracode/skills/my-skill/SKILL.md` — it works

## Additional Context

- Symlinks from `.mastracode/commands/` → `.claude/commands/` were also tried; same result
- Hard-copied files in `.mastracode/commands/` also do not work
- Frontmatter with and without `name` field was tested; neither works
- The `--version` output shows all expected directories are scanned:
  ```
  Skills loaded from:
    - /path/to/project/.claude/skills
    - ~/.mastracode/skills
    - ~/.agents/skills
    - ~/.claude/skills
  ```

## Environment

- **Mastra Code version:** 0.9.2
- **Node.js:** v22.14.0
- **OS:** macOS (Darwin)

intojhanurag:
## Description
Custom slash commands from `.mastracode/commands/`, `.claude/commands/`, and other configured directories were silently failing to load. The root cause was a CJS/ESM interop issue with the `js-yaml` import — `import * as yaml from 'js-yaml'` results in `yaml.load` being `undefined` at runtime since `js-yaml` v3 is a CJS module. The error was silently caught, causing all command files to be skipped.

Additionally fixed two minor bugs in the same file:
- **Priority inversion**: The dedup logic used a `Set` (first-come-first-served), but sources load
lowest-to-highest priority — so the lowest priority source incorrectly won. Switched to a `Map` so
higher-priority sources correctly override.
- **Namespace loss in subdirectories**: Recursive directory scanning lost the root directory
reference, causing `git/commit.md` to be named `commit` instead of `git:commit`.

  ## Related Issue(s)

  Fixes #14719

  ## Type of Change

  - [ ] Documentation update
  - [ ] Code refactoring
  - [ ] Performance improvement
  - [ ] Test update

  ## Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Custom slash commands now load correctly from all configured directories.

### DIFF
--- a/mastracode/src/utils/slash-command-loader.ts
+++ b/mastracode/src/utils/slash-command-loader.ts
@@ -1,6 +1,6 @@
 import { promises as fs } from 'node:fs';
 import * as path from 'node:path';
-import yaml from 'js-yaml';
+import { parse as parseYaml } from 'yaml';
 
 /**
  * Metadata for a slash command
@@ -51,7 +51,7 @@ export async function parseCommandFile(filePath: string, baseDir?: string): Prom
     const template = parts.slice(2).join('---').trim();
 
     // Parse YAML frontmatter
-    const metadata = yaml.load(frontmatter) as Record<string, string>;
+    const metadata = parseYaml(frontmatter) as Record<string, string>;
 
     // Derive name from file path if not specified in frontmatter
     let name: string;
--- a/mastracode/src/utils/slash-command-loader.ts
+++ b/mastracode/src/utils/slash-command-loader.ts
@@ -1,6 +1,6 @@
 import { promises as fs } from 'node:fs';
 import * as path from 'node:path';
-import * as yaml from 'js-yaml';
+import yaml from 'js-yaml';
 
 /**
  * Metadata for a slash command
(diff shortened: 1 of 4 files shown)


mastracode/src/utils/slash-command-loader.ts
```diff
@@ -1,6 +1,6 @@
 import { promises as fs } from 'node:fs';
 import * as path from 'node:path';
-import * as yaml from 'js-yaml';
+import yaml from 'js-yaml';
```
LekoArts: This is an extra request, feel free to ignore:

Can you swap out the dependency to a more modern one: https://e18e.dev/docs/replacements/js-yaml.html#yaml

intojhanurag: I swapped js-yaml to yaml per your link, but it adds package.json and pnpm-lock.yaml changes which makes the PR noisier for a one-line bugfix. Let me know if you'd prefer I revert and keep just the import fix, or if this is
fine.

LekoArts (approved):

wardpeet (approved):

PR #14730, opened 2026-03-26
feat(core,observability): add registerExporter for runtime exporter registration

rphansen91:
## Summary

Adds `registerExporter()` method throughout the observability stack, enabling platform tooling (e.g. studio runner) to inject exporters like `CloudExporter` at runtime without requiring users to configure them upfront.

### Core (`@mastra/core`)
- **`ObservabilityInstance` interface** — added `registerExporter(exporter)` method
- **`ObservabilityEntrypoint` interface** — added `registerExporter(exporter)` method
- **`NoOpObservability`** — no-op implementation (silently discards)
- **`Mastra.registerExporter(exporter, entrypoint?)`** — top-level method that:
  - If observability is `NoOpObservability` and an `entrypoint` is provided → replaces no-op, sets logger/context, then registers
  - If observability is already real → delegates directly
  - If no-op without entrypoint → silently returns

### Observability (`@mastra/observability`)
- **`BaseObservabilityInstance.registerExporter()`** — registers on the bus, tracks in config exporters, propagates logger, and calls `init()` with mastra context if available
- **`Observability.registerExporter()`** — delegates to the default instance

### Tests
- 7 new tests in `register-exporter.test.ts` covering:
  - Registration on default instance
  - No-throw when no default instance exists
  - Named config registration
  - Duplicate exporter suppression
  - Init with mastra context
  - No init without mastra context
  - Error handling on init failures

## Test Plan

```bash
cd observability/mastra && pnpm test -- --run register-exporter
```

All 26 test files pass (594 tests passed, 2 skipped).

## Summary by CodeRabbit

* **New Features**
  * Added registerExporter() to allow runtime registration of observability exporters — enables adding or switching exporters after initialization and updates observability behavior dynamically.
* **Tests**
  * Added unit tests covering registerExporter() across uninitialized, initialized-with-default, and already-configured observability states.
* **Chores**
  * Added release metadata to trigger a minor version release for observability-related packages.


packages/core/src/mastra/index.ts
```diff
@@ -494,6 +500,36 @@ export class Mastra<
...
+    exporter: ObservabilityExporter,
+    instance: ObservabilityInstance,
+    entrypoint: ObservabilityEntrypoint,
+  ): void {
+    if (this.#observability instanceof NoOpObservability) {
+      this.#observability = entrypoint;
+      this.#observability.setLogger({ logger: this.#logger });
+      this.#observability.setMastraContext({ mastra: this });
+      this.#observability.registerInstance('default', instance, true);
+    } else {
```
epinzur: should this not be an `else` ?

ie, let the flow continue, registering the exporter on the newly created default instance?

packages/core/src/observability/types/core.ts
```diff
@@ -230,6 +230,14 @@ export interface ObservabilityInstance {
...
    */
   getMetricsContext?(span?: AnySpan): MetricsContext;
+
+  /**
+   * Register an additional exporter to this instance at runtime.
+   * Duplicate registrations (same instance) are silently ignored.
+   *
+   * @param exporter - The exporter to register
+   */
+  registerExporter(exporter: ObservabilityExporter): void;
```
epinzur: this should prob be an optional method, so it's not a breaking change.

epinzur (changes requested):
generally looks good

needs fixing:

* `registerExporter` should be optional on `@mastra/core` to ensure this is not a breaking change.

also added a few other comments

epinzur (approved):

PR #14778, opened 2026-03-27
feat(memory): add minMessages option to generateTitle config

Akash504-ai:
## Description

This PR adds support for a `minMessages` option in the `generateTitle` configuration for Memory.

Currently, title generation is triggered on the very first message, which often results in duplicate or generic titles (e.g., "Hello", "New Chat") in many agent workflows. This is especially problematic for onboarding or template-based conversations where initial messages are identical.

With this change, developers can delay title generation until a minimum number of messages have been exchanged, ensuring titles are more meaningful and context-aware.

---

## Changes

* Added optional `minMessages` field to `generateTitle` config
* Updated `resolveTitleGenerationConfig` to include `minMessages`
* Enforced `minMessages` threshold in both:

  * `agent.ts`
  * `agent-legacy.ts`
* Default behavior remains unchanged (`minMessages` defaults to `1`)

---

## Example Usage

```ts
memory: new Memory({
  options: {
    generateTitle: {
      model: TITLE_MODEL,
      minMessages: 5, // Title generated only after 5 messages
    }
  }
})
```

---

## Why this matters

* Prevents duplicate/generic titles in chat history
* Reduces unnecessary LLM calls (token savings)
* Improves UX by generating titles based on meaningful user intent

---

## Related Issue

Fixes #14755

---

## Type of Change

---

## Checklist

* [ ] I have added tests that prove my feature works
* [x] I have addressed all Coderabbit comments on this PR

## Summary by CodeRabbit

* **Improvements**
  * Title generation supports an optional minMessages setting to control when automatic thread titles are created.
  * Automatic titles are produced only when enabled, no title exists, a recent user message is present, and the message count meets or exceeds minMessages (defaults to 1).
  * Logic now separately evaluates visible messages and core message count, ensuring the most recent visible user message is used when deciding to generate a title.


packages/core/src/agent/agent.ts
```diff
@@ -5133,7 +5137,7 @@ export class Agent<
       });
     }
 
-    const fullOutput = (await result.result.getFullOutput()) as Awaited<
+    const fullOutput = (await (result.result as MastraModelOutput<OUTPUT>).getFullOutput()) as Awaited<
```
TylerBarnes: why is this change needed?

TylerBarnes: all of these seem random

packages/core/src/agent/agent.ts
```diff
@@ -4388,27 +4388,31 @@ export class Agent<
           shouldGenerate,
           model: titleModel,
           instructions: titleInstructions,
-        } = this.resolveTitleGenerationConfig(config.generateTitle);
+          minMessages,
+        } = this.resolveTitleGenerationConfig(config?.generateTitle);
 
-        if (shouldGenerate && !thread.title) {
-          const userMessage = this.getMostRecentUserMessage(messageList.get.all.ui());
+        const messages = messageList.get.all.ui();
```
TylerBarnes: I wouldn't count ui messages, one assistant ui message can contain basically unlimited LLM message parts, you want to count model messages

TylerBarnes (changes requested):

packages/core/src/agent/agent-legacy.ts
```diff
@@ -589,9 +594,13 @@ export class AgentLegacyHandler {
...
               model: titleModel,
               instructions: titleInstructions,
+              minMessages,
             } = this.capabilities.resolveTitleGenerationConfig(config?.generateTitle);
 
-            if (shouldGenerate && !thread.title && userMessage) {
+            const messages = messageList.get.all.ui();
+            const requiredMessages = minMessages ?? 1;
+
+            if (shouldGenerate && !thread.title && userMessage && messages.length >= requiredMessages) {
```
TylerBarnes: need to update this one too

packages/core/src/agent/agent.ts
```diff
@@ -4388,27 +4388,34 @@ export class Agent<
...
-                this.logger.error('Error persisting generated title:', error);
-              },
+            const title = await this.genTitle(
+              userMessage,
+              requestContext,
+              observabilityContext,
+              titleModel,
+              titleInstructions,
             );
+
```
TylerBarnes:
```suggestion
```

packages/core/src/agent/agent.ts
```diff
@@ -5334,12 +5338,13 @@ export class Agent<
   resolveTitleGenerationConfig(
     generateTitleConfig:
       | boolean
-      | { model: DynamicArgument<MastraModelConfig>; instructions?: DynamicArgument<string> }
+      | { model: DynamicArgument<MastraModelConfig>; instructions?: DynamicArgument<string>; minMessages?: number }
       | undefined,
```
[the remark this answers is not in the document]
TylerBarnes: this seems legit actually

TylerBarnes (changes requested):

CleverFool77 (commented):
What happens if minMessages is 10. Does the agent waits till 10th message then to show the title ?  Can we not try fire on first message and then also fire at 10th message in such case. More like initial title now, then update later. 
And what would be our initial title based on our current implementation ? is it "New Thread 2026-03-04T11:48:22.587Z" ?

wardpeet (approved):

TylerBarnes (approved):

PR #14782, opened 2026-03-27
Fix notifications sent at end of httpMCPs

noahjohnhay:
## Summary by CodeRabbit

* **Bug Fixes**
  * MCP HTTP transport now streams notifications incrementally (e.g., SSE) as they are produced instead of buffering until request handling completes, improving delivery responsiveness.
  * Failures when initiating the HTTP response are now handled asynchronously and logged; if headers are already sent, in-progress streams continue uninterrupted.


wardpeet (approved):
LGTM, thank you!

PR #14783, opened 2026-03-27
feat(playground): attach scorers to datasets

DanielSLew:
Adds persistent scorer attachment to datasets. When scorers are attached to a dataset, they automatically run alongside any scorers specified at experiment trigger time.

**Core/Storage:** New `scorerIds` field on `DatasetRecord`, `CreateDatasetInput`, and `UpdateDatasetInput`. Added to `DATASETS_SCHEMA` and all storage backends (InMemory, LibSQL, PostgreSQL, MongoDB). At experiment time, dataset-attached scorers are merged with explicitly-provided scorers and deduplicated.

**Server/Client SDK:** `scorerIds` flows through CREATE/UPDATE handlers and request/response schemas. Client types updated to match.

**Playground UI:** New collapsible "Scorers" section in the dataset detail view with attach/detach via a dialog picker. `ScorerDetailView` now also shows datasets where the scorer is attached via `scorerIds`.

Also fixes a crash on page refresh in the evaluate tab — `ReviewQueueProvider` was conditionally unmounted during loading states, causing `useReviewQueue` to throw.

## Summary by CodeRabbit

* **New Features**
  - Datasets can store associated scorer IDs; create/update responses include scorer lists.
  - Experiments now merge dataset-attached scorers with user-configured scorers (string IDs deduplicated).
  - UI: attach/detach scorers from a dataset via the Dataset detail view, with an attach dialog and per-scorer controls.

* **Bug Fixes**
  - Prevented a crash on refresh in the evaluate tab by ensuring required providers render during loading.

### DIFF
--- a/.changeset/attach-scorer-ui.md
+++ b/.changeset/attach-scorer-ui.md
@@ -1,0 +1,5 @@
+---
+'@mastra/playground-ui': patch
+---
+
+Added a Scorers section to the dataset detail view for attaching and detaching scorers. Fixed a crash on page refresh in the evaluate tab caused by `ReviewQueueProvider` not being mounted during loading states.
--- a/.changeset/calm-bags-sit.md
+++ b/.changeset/calm-bags-sit.md
@@ -1,0 +1,5 @@
+---
+'@mastra/core': minor
+---
+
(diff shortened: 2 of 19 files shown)


YujohnNattrass (approved):

DanielSLew:
**Re: [dataset.ts:360 — Pin targetVersion when handing off to runExperiment](https://github.com/mastra-ai/mastra/pull/14783#discussion_r3002898249)**

AI says: Fixed in 9f8f01c — `targetVersion` is now explicitly passed as `version` in the `runExperiment` call, so the background worker runs against the same dataset snapshot recorded on the experiment row.

— Mastra Code

PR #14787, opened 2026-03-27
fix(server): handle non-Zod schemas in tool serialization

DanielSLew:
When MCP tools (or any tool with a JSON Schema-backed `inputSchema`) were registered, `GET /tools` would crash with:

```
TypeError: Cannot read properties of undefined (reading 'typeName')
```

`serializeTool` in `tools.ts` assumed `tool.inputSchema` was always a Zod schema and called `zodToJsonSchema` directly. MCP tools use `StandardSchemaWithJSON` (a `JsonSchemaWrapper`) which doesn't have `._def.typeName`.

Added a `serializeSchema` helper that checks `isStandardSchemaWithJSON` first, then lazy AI SDK schemas, and only falls back to `zodToJsonSchema` for raw Zod schemas.

Note: `getSerializedAgentTools` in `agents.ts` had the same bug but was already fixed on main via the `schemaToJsonSchema` refactor, so after merging main I took that version.

## Bug Fixes
- Fixed a runtime crash that occurred when retrieving tools or agents containing MCP tools or other tools that use custom JSON schemas instead of standard formats. Tool information now correctly displays and is properly accessible through API endpoints for all supported tool types.


packages/server/src/server/handlers/tools.ts
```diff
@@ -60,9 +89,9 @@ function serializeTool(tool: any): any {
 
   return {
     ...tool,
-    inputSchema: tool.inputSchema ? stringify(zodToJsonSchema(tool.inputSchema)) : undefined,
-    outputSchema: tool.outputSchema ? stringify(zodToJsonSchema(tool.outputSchema)) : undefined,
-    requestContextSchema: tool.requestContextSchema ? stringify(zodToJsonSchema(tool.requestContextSchema)) : undefined,
+    inputSchema: serializeSchema(tool.inputSchema),
+    outputSchema: serializeSchema(tool.outputSchema),
```
wardpeet, line 74:
```
import { standardSchemaToJSONSchema, toStandardSchema } from '@mastra/schema-compat/schema';
```

This should do the trick as well
```suggestion
    inputSchema: standardSchemaToJSONSchema(toStandardSchema(tool.inputSchema)),
    outputSchema: standardSchemaToJSONSchema(toStandardSchema(tool.outputSchema)),
```

wardpeet (approved):

PR #14804, opened 2026-03-28
fix(mastracode): show configured subagents in /subagents

ts-76:
## Summary

Fix `/subagents` so it reflects configured subagents from `createMastraCode({ subagents })` instead of always showing the hardcoded built-in list.

Closes #14803.

## Problem

`createMastraCode()` accepts a `subagents` configuration, but the `/subagents` TUI command was building its selector options from a fixed built-in list. As a result, custom configured subagents were not shown in the UI.

## Solution

- read `ctx.state.harness.config?.subagents` in `/subagents`
- map configured subagents to the selector options
- fall back to the built-in list only when no subagents are configured
- add a targeted test for both the fallback and configured-subagents cases

## Testing

- `npx pnpm@latest --dir mastracode test src/tui/commands/__tests__/subagents.test.ts`

## Summary by CodeRabbit

* **New Features**
  * Subagent selection now reflects harness configuration; when absent or empty, the UI falls back to built-in default options.

* **Tests**
  * New tests validate the subagent prompt, option rendering (configured vs. default), and command control flow including cancel/submit behavior.

* **Chores**
  * Added a changeset entry for a patch release documenting the subagents update.


TylerBarnes (commented):
looks great, thank you! only thing needed is a changeset

TylerBarnes (approved):
approving but just need that changeset, thanks again!

PR #14830, opened 2026-03-30
fix(core): normalize bare '*' pattern to '**/*' in list_files tool

### LINKED ISSUE #14823: mastra_workspace_list_files: pattern "*" silently filters out all files at depth > 0
## Bug Description

When an LLM calls the `mastra_workspace_list_files` tool with `pattern: "*"`, all files in subdirectories are silently excluded from the output. Only top-level directories are returned.

The glob `*` does not match paths containing `/`. Since files at depth > 0 have relative paths like `subdir/file.txt`, the glob filter rejects them. Directories are exempt (the filter has an explicit `if (e.type === "directory") return true` check), so they pass through — but their child files don't.

This results in output like `26 directories, 0 files` even though every directory contains files.

## Steps to Reproduce

1. Have a workspace with files in subdirectories (e.g. `dir-a/file1.txt`, `dir-b/file2.pdf`)
2. Call `mastra_workspace_list_files` with `{ path: ".", maxDepth: 3, pattern: "*" }`
3. Result: directories are listed but all files are missing — `N directories, 0 files`
4. Same call without `pattern` (or with `pattern: "**/*"`) returns files correctly

## Root Cause

In `src/workspace/tools/tree-formatter.ts`, the glob filter matches against the full relative path:

```ts
if (globMatcher && !dirsOnly) {
  filtered = filtered.filter((e) => {
    if (e.type === "directory") return true;
    const relativePath = getRelativePath(path8, currentPath, e.name);
    return globMatcher(relativePath);
  });
}
```

At depth > 0, `relativePath` is `subdir/filename.ext`. The glob `*` doesn't cross `/` boundaries, so it rejects all such paths.

## Impact

LLMs commonly pass `pattern: "*"` as a "match everything" intent. The tool schema describes `pattern` as "Glob pattern(s) to filter files" — passing `*` for "all files" is a reasonable interpretation. The tool silently returns 0 files instead of erroring or matching correctly, making this hard to diagnose.

## Environment

- `@mastra/core`: 1.3.15

Aayush-engineer:
## Summary
Fixes #14823

## Problem
When `list_files` is called with `pattern: "*"`, all files in 
subdirectories are silently excluded. Only top-level directories 
are returned, resulting in "N directories, 0 files".

Root cause: bare `*` doesn't cross `/` boundaries in glob matching,
so `subdir/file.txt` never matches `*`.

## Fix
Normalize bare `*` to `**/*` before compiling the glob matcher.
This matches LLM intent — `*` as "match everything".

## Changes
- `packages/core/src/workspace/tools/tree-formatter.ts` — 4-line fix
- `packages/core/src/workspace/tools/__tests__/tree-formatter.test.ts` — regression test

## Tests
- 1 new regression test added ✅
- 54/54 tests passing, 0 type errors ✅

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed `list_files` tool glob pattern matching. The bare `*` pattern now correctly matches files at any directory depth, including files in subdirectories, instead of failing to match nested files.

### DIFF
--- a/.changeset/fix-list-files-bare-star-pattern.md
+++ b/.changeset/fix-list-files-bare-star-pattern.md
@@ -1,0 +1,5 @@
+---
+"@mastra/core": patch
+---
+
+Workspace list_files tool now correctly handles pattern "*" by treating it as "**/*", matching files at all depths instead of silently returning 0 files.
--- a/packages/core/src/workspace/tools/__tests__/tree-formatter.test.ts
+++ b/packages/core/src/workspace/tools/__tests__/tree-formatter.test.ts
@@ -592,6 +592,18 @@ src
       expect(result.fileCount).toBe(2);
     });
 
+    it('should treat bare "*" as "**/*" to match files at any depth (#14823)', async () => {
+      await fs.mkdir(path.join(tempDir, 'src'));
+      await fs.mkdir(path.join(tempDir, 'src', 'utils'));
+      await fs.writeFile(path.join(tempDir, 'index.ts'), '');
+      await fs.writeFile(path.join(tempDir, 'src', 'app.ts'), '');
+      await fs.writeFile(path.join(tempDir, 'src', 'utils', 'helpers.ts'), '');
+      const result = await formatAsTree(filesystem, '.', { pattern: '*' });
+      expect(result.tree).toContain('index.ts');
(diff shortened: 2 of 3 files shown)


wardpeet (approved):
LGTM, thank you!

PR #14831, opened 2026-03-30
fix(mastracode): apply thinkingLevel to Anthropic models

gnodet:
## Summary

- Pass `thinkingLevel` from harness state to both `opencodeClaudeMaxProvider` and `anthropicApiKeyProvider` in `resolveModel()` — previously only the OpenAI Codex path used it
- Add `createAnthropicThinkingMiddleware()` that maps `ThinkingLevel` to `providerOptions.anthropic.thinking` with appropriate `budgetTokens` (low→4096, medium→16384, high→65536, xhigh→unlimited)
- Add 14 unit tests covering the mapping, middleware behavior, and option preservation

## Problem

When a user sets `thinkingLevel` (e.g. via `harness.setState({ thinkingLevel: 'high' })`), the setting only takes effect for OpenAI Codex models. For Anthropic models, extended thinking is silently dropped — `opencodeClaudeMaxProvider` and `anthropicApiKeyProvider` are called without `thinkingLevel`, and neither configures the Anthropic `thinking` parameter.

This is the root cause behind:
- https://github.com/anthropics/claude-code/issues/30726 — "max" accepted but treated as "high"
- https://github.com/anthropics/claude-code/issues/37864 — `--effort max` accepted but not applied
- https://github.com/anthropics/claude-code/issues/34171 — effort resets every session

## Test plan

- [x] New test suite `anthropic-thinking.test.ts` — 14 tests covering thinking level mapping, middleware creation for all levels, `transformParams` injection, preservation of existing provider options, and `xhigh` having no budget limit
- [x] Updated `model.test.ts` — existing expectations updated and 4 new tests verifying `thinkingLevel` is passed through `resolveModel()` to both OAuth and API key Anthropic providers
- [ ] Manual: set `thinkingLevel` to `high` or `xhigh` with an Anthropic model and verify extended thinking output appears

_Claude Code on behalf of Guillaume Nodet_

## Summary by CodeRabbit

* **New Features**
  * Added configurable thinking levels for Anthropic Claude models, letting users control extended thinking depth.
  * Supports intensity options: off, low, medium, high, and xhigh with corresponding computational budgets.
  * Works with both OAuth and API-key authentication methods.

* **Tests**
  * Added comprehensive test coverage for thinking-level mapping and middleware integration.

### DIFF
--- a/mastracode/src/providers/claude-max.ts
+++ b/mastracode/src/providers/claude-max.ts
@@ -10,6 +10,7 @@ import type { MastraModelConfig } from '@mastra/core/llm';
 import { wrapLanguageModel } from 'ai';
 import type { LanguageModelMiddleware } from 'ai';
 import { AuthStorage } from '../auth/storage.js';
+import type { ThinkingLevel } from './openai-codex.js';
 
 // Required for Claude Max plan OAuth - the endpoint checks for this system message
 const claudeCodeIdentity = "You are Claude Code, Anthropic's official CLI for Claude.";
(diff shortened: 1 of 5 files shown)


mastracode/src/providers/claude-max.ts
```diff
@@ -189,6 +237,6 @@ export function opencodeClaudeMaxProvider(modelId: string = 'claude-sonnet-4-202
   // Wrap with middleware to inject Claude Code identity and enable prompt caching
   return wrapLanguageModel({
     model: anthropic(modelId),
-    middleware: [claudeCodeMiddleware, promptCacheMiddleware],
+    middleware: [claudeCodeMiddleware, promptCacheMiddleware, ...(thinkingMiddleware ? [thinkingMiddleware] : [])],
```
TylerBarnes, line 240: rather than a middleware this should be set on the agent.stream() call, so that if we started caching this function in the future, or we added the ability for subagents to have a different thinking level than the main one, the thinking level could still be dynamically changed per request

TylerBarnes (changes requested):
one inline comment but there are also two general problems, which is why this hasn't been wired up yet
1. There was a mastra core bug with thinking signatures not always being sent back properly, I'm not sure if it's been fixed. With these changes have you found that it's working well?
2. We generally have an issue where we don't prepare the message history yet per-provider. For most providers this isn't an issue, however with anthropic, if you turn thinking on and then send back signatures/thinking-blocks that were made by a different provider (say you did planning mode with a non-anthropic model, then switched to anthropic for build mode), anthropic will reject the request. See this comment for more info https://github.com/mastra-ai/mastra/pull/14578#pullrequestreview-4002538427

abhiaiyer91:
Closing due to inactivity

PR #14847, opened 2026-03-30
fix: default agent version resolution to published instead of draft

DanielSLew:
Agents were resolving to the draft version by default across all execution endpoints. Now they default to the latest published version, which is what users expect in the playground chat tab and versioned endpoints.

The editor tab still uses the draft/selected version — it passes `agentVersionId` via `requestContext` so the server knows which version to resolve.

Also wires `requestContext` through the client-js tool approval/decline methods (`approveToolCall`, `declineToolCall`, etc.) so the version context isn't lost when resuming after a tool approval pause.

Adds `resolvedVersionId` to agent run trace span attributes so you can see which version was actually used during execution.

## Summary by CodeRabbit

* **New Features**
  * Tool approval/decline flows now carry per-request context so agent version info is preserved when resuming.
  * Agent execution traces include a resolvedVersionId for clearer version traceability.
  * UI, chat, playground, and runtime accept an optional agentVersionId to target specific agent versions.

* **Documentation**
  * Agent retrieval docs updated: published versions are described as the default "active" selection.

### DIFF
--- a/client-sdks/react/src/agent/hooks.ts
+++ b/client-sdks/react/src/agent/hooks.ts
@@ -65,6 +65,7 @@ export const useChat = ({ agentId, resourceId, initialMessages }: MastraChatProp
   const _onChunk = useRef<((chunk: ChunkType) => Promise<void>) | undefined>(undefined);
   const _networkRunId = useRef<string | undefined>(undefined);
   const _onNetworkChunk = useRef<((chunk: NetworkChunkType) => Promise<void>) | undefined>(undefined);
+  const _requestContext = useRef<RequestContext | undefined>(undefined);
   const [messages, setMessages] = useState<MastraUIMessage[]>([]);
   const [toolCallApprovals, setToolCallApprovals] = useState<{
     [toolCallId: string]: { status: 'approved' | 'declined' };
@@ -104,6 +105,7 @@ export const useChat = ({ agentId, resourceId, initialMessages }: MastraChatProp
       maxSteps,
       requireToolApproval,
     } = modelSettings || {};
+    _requestContext.current = requestContext;
     setIsRunning(true);
 
     // Create a new client instance with the abort signal
@@ -205,6 +207,7 @@ export const useChat = ({ agentId, resourceId, initialMessages }: MastraChatProp
       requireToolApproval,
     } = modelSettings || {};
 
+    _requestContext.current = requestContext;
     setIsRunning(true);
 
     // Create a new client instance with the abort signal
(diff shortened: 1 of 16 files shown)


client-sdks/react/src/agent/hooks.ts
```diff
@@ -104,6 +115,7 @@ export const useChat = ({ agentId, resourceId, initialMessages }: MastraChatProp
       maxSteps,
       requireToolApproval,
     } = modelSettings || {};
+    _requestContext.current = requestContext;
```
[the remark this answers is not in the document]
DanielSLew: AI says: This concern is already addressed in the current code. Lines 116, 219, and 282 use `const resolvedRequestContext = requestContext ?? propsRequestContext` before assigning to `_requestContext.current`, so the persistent prop-based context is used as a fallback when per-call `requestContext` is omitted — it never gets overwritten with `undefined`.

— Mastra Code

packages/playground-ui/src/services/mastra-runtime-provider.tsx
```diff
@@ -389,6 +390,13 @@ export function MastraRuntimeProvider({
     setLegacyMessages(initializeMessageState(initialLegacyMessages || []));
   }, [initialLegacyMessages]);
 
+  const chatRequestContext = useMemo(() => {
```
wardpeet, line 393: i'm not sure how this code actually is written but not enough context to say no. Isnt this only rendered once anyway?

client-sdks/react/src/agent/hooks.ts
```diff
@@ -60,11 +62,17 @@ const extractRunIdFromMessages = (messages: ExtendedMastraUIMessage[]): string |
...
+  agentId,
+  resourceId,
+  initialMessages,
+  requestContext: propsRequestContext,
+}: MastraChatProps) => {
   const _currentRunId = useRef<string | undefined>(undefined);
   const _onChunk = useRef<((chunk: ChunkType) => Promise<void>) | undefined>(undefined);
   const _networkRunId = useRef<string | undefined>(undefined);
   const _onNetworkChunk = useRef<((chunk: NetworkChunkType) => Promise<void>) | undefined>(undefined);
+  const _requestContext = useRef<RequestContext | undefined>(propsRequestContext);
```
wardpeet, line 75: I find this setup with _requestContext a bit strange but not enough context 🤷

wardpeet (approved):
Not sure about these 2 comments but not against merging it, I don't know enough about studio and its setup to actually say no to it. This looks like non-react patterns.

PR #14859, opened 2026-03-30
fix(client-js): move @mastra/core to peerDependencies

intojhanurag:
## Summary

`@mastra/client-js` listed `@mastra/core` as a bundled dependency instead of a peer dependency. This caused package managers to install two separate copies of `@mastra/core` — one nested inside `client-js` and one hoisted for the rest of the app. Because `RequestContext` uses a private property (`registry`), TypeScript treated the two copies as incompatible types, producing `ts(2769)`.

Every other `@mastra/*` package (`memory`, `mcp`, `rag`, `server`) already declares `@mastra/core` as a peer dependency. This PR aligns `client-js` with that pattern.

  ## Context

  - `@mastra/core` was originally moved to `peerDependencies` across all packages by Ward
  Peeters in May 2025 (`83da932`).
  - It was reverted to a regular dependency in `client-js` only, as a quick fix for Vercel
  example builds in June 2025 (`b2810ab`, PR #4774).
  - That workaround was never cleaned up, and later refactors continued building on it.

## Issue

  Fixes #14855 

## Test plan

  - [x] `client-js` build passes (`pnpm build:lib`)
  - [x] All 343 `client-js` unit tests pass (`pnpm test:unit`)
  - [x] `e2e-tests/pkg-outputs/bundle.test.ts` now validates `client-js` follows the peer
  dependency rule

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed RequestContext type mismatch when using `@mastra/client-js` with `@mastra/core` together
  * Resolved TypeScript errors caused by duplicate dependency installations by updating dependency structure

* **Tests**
  * Enhanced validation for `@mastra/client-js` package compatibility

### DIFF
--- a/.changeset/fix-client-js-core-peer-dep.md
+++ b/.changeset/fix-client-js-core-peer-dep.md
@@ -2,4 +2,4 @@
 '@mastra/client-js': patch
 ---
 
-Fixed RequestContext type mismatch when using @mastra/client-js alongside @mastra/core. Moved @mastra/core from a bundled dependency to a peer dependency, matching the pattern used by all other @mastra packages. This prevents duplicate installations that caused TypeScript error `ts(2769): Types have separate declarations of a private property 'registry'`.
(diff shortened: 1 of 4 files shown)


wardpeet (commented):
It should be embedded, we shouldn't need to install `@mastra/core`. It mostly means there is a different problem. Maybe we are not embedding the types properly

Issue #14907, opened 2026-03-31
makePushOrCombine merges consecutive user messages into single turn, breaking conversation history


sawyer-denari:
## Bug Description

When using `useChat` from `@ai-sdk/react` with Mastra memory (Observational Memory enabled), consecutive user messages are merged into a single API turn by `makePushOrCombine` in `convertToV1Messages`. This drops intermediate assistant responses from the model's perspective, causing the model to re-greet and ignore conversation context.

## Reproduction

1. Configure an agent with `Memory` (Observational Memory enabled) and `useChat` from `@ai-sdk/react`
2. Send a greeting message: "Hello, how are you?"
3. Wait for the assistant response
4. Send a follow-up: "What can you do?"
5. The model re-greets instead of answering the follow-up

## What's Happening

The Anthropic API request body shows the messages are merged:

```json
"messages": [
  { "role": "user", "content": [{ "text": "." }] },
  { "role": "assistant", "content": [{ "text": "Hi Sawyer! I'm doing great..." }] },
  {
    "role": "user",
    "content": [
      { "text": "Hello how are you" },
      { "text": "What can you do?" }
    ]
  }
]
```

The UI shows 5 separate messages (3 user, 2 assistant), but the API receives only 3 — the two user messages and the assistant response between them are collapsed into a single user turn with multiple text parts.

## Root Cause

**`makePushOrCombine`** in `convertToV1Messages` (`src/agent/message-list/prompt/convert-to-mastra-v1.ts` or equivalent):

```js
var makePushOrCombine = (v1Messages) => {
  // ...
  return (msg) => {
    const previousMessage = v1Messages.at(-1);
    if (msg.role === previousMessage?.role && Array.isArray(previousMessage.content) && Array.isArray(msg.content) &&
        (msg.role !== `assistant` || msg.role === `assistant` && msg.content.at(-1)?.type !== `tool-call`)) {
      // Merges consecutive same-role messages
      for (const part of msg.content) {
        previousMessage.content.push(part);
      }
    } else {
      v1Messages.push(msg);
    }
  };
};
```

This function merges **any** consecutive same-role messages except assistant messages ending with tool calls. When Observational Memory's `loadHistoricalMessagesIfNeeded` adds DB messages as "memory" and the frontend messages are added as "input", the internal ordering in `MessageList` can produce consecutive user messages (with the assistant message between them lost or reordered). `makePushOrCombine` then merges them.

## Expected Behavior

[...]

PR #14916, opened 2026-04-01
fix(core): prevent MessageList sort from breaking message order with mixed timestamp sources

### LINKED ISSUE #14912: docs: show that when using memory, the client should only send new messages, not the full history
## Bug

`MessageList.addOne` sorts all messages by `createdAt` after every insertion. When the frontend (AI SDK v6's `useChat`) doesn't send `createdAt` on messages, Mastra assigns `Date.now()` to input messages but retains the original DB timestamp for messages that hit the `shouldReplace` path. The sort then puts the older DB-timestamped assistant message before the newer user messages, breaking the alternating user/assistant pattern.

## Result

The model receives a mangled conversation:
```
user: "."              ← injected by ensureGeminiCompatibleMessages (assistant was first)
assistant: "Hi!"       ← sorted first due to old DB timestamp
user: ["msg1", "msg2"] ← two user messages merged by makePushOrCombine
```

## Root cause

```typescript
// message-list.ts, addOne() — added in refactor commit 73b0bb39 (2026-01-06)
// "make sure messages are always stored in order of when they were created!"
this.messages.sort((a, b) => a.createdAt.getTime() - b.createdAt.getTime());
```

The sort assumes reliable timestamps. This breaks when:
1. AI SDK v6 (`@ai-sdk/react` 3.x) removed `createdAt` from the transport body
2. Observational Memory's `loadHistoricalMessagesIfNeeded` triggers `shouldReplace`, which keeps the DB version's old `createdAt`
3. Input user messages get `Date.now()` via `generateCreatedAt`

The assistant message (DB timestamp from seconds/minutes ago) sorts before user messages (`Date.now()`), producing consecutive user messages that `makePushOrCombine` then merges.

## Workaround

Commenting out the sort in `addOne` fixes the issue — the frontend's array order is preserved.

## Suggested fix

Either:
- Don't sort when timestamps are synthetic/missing — trust insertion order
- Or ensure `shouldReplace` propagates the input message's `createdAt` instead of keeping the DB version's

## Environment

- `@mastra/core`: 1.14.0
- `@mastra/memory`: 1.6.1 (Observational Memory enabled)
- `ai`: 6.0.89
- `@ai-sdk/react`: 3.0.91

## Related

- #14907 (our earlier report focusing on `makePushOrCombine`, same root bug)
- #3735 (message order swap due to similar timestamps)

intojhanurag:
## Description

  `MessageList.addOne()` sorted all messages by `createdAt` after every insertion. When input messages from
  the frontend (AI SDK v6's `useChat`) have no `createdAt` and receive `Date.now()` timestamps, while memory
   messages retain their old DB timestamps, the sort reorders the old assistant message before the newer
  user messages. This breaks the user/assistant alternating pattern, causing `makePushOrCombine` to merge
  consecutive user messages and `ensureGeminiCompatibleMessages` to inject a dummy `"."` user message.

  **Changes:**
  - Removed the two `this.messages.sort()` calls in `addOne()` that caused the reordering
  - Memory messages are now inserted at their chronological position via `findIndex` + `splice` instead of
  append + sort
  - When `shouldReplace` fires, the newer `createdAt` is preserved so a DB message's old timestamp does not
  regress an input message's position

  ## Related Issue(s)

  Fixes #14912
  Related to #14907

  ## Type of Change

  - [ ] Documentation update
  - [ ] Code refactoring
  - [ ] Performance improvement
  - [x] Test update

  ## Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed message reordering when mixing newly-supplied input messages with previously persisted memory messages, preserving chronological order and correct user/assistant alternation.

* **Tests**
  * Added tests covering mixed input/memory timestamp scenarios, replacement cases, and edge conditions to validate ordering, role alternation, and message updates.

### DIFF
--- a/.changeset/better-regions-sin.md
+++ b/.changeset/better-regions-sin.md
@@ -1,0 +1,5 @@
+---
+'@mastra/core': patch
+---
+
+Fixed message reordering when mixing input messages (Date.now() timestamps) with memory messages (old DB timestamps). The full-array sort in MessageList.addOne() could move historical assistant messages before current user messages, breaking the user/assistant alternating pattern and causing consecutive user messages to be merged.
--- a/packages/core/src/agent/message-list/message-list.ts
+++ b/packages/core/src/agent/message-list/message-list.ts
@@ -1003,7 +1003,15 @@ export class MessageList {
             if (messageV2.createdAt <= existingMessage.createdAt) {
               messageV2.createdAt = new Date(existingMessage.createdAt.getTime() + 1);
             }
-            this.messages.push(messageV2);
+            // Insert at the correct chronological position so that removing the
+            // full-array sort does not leave split-off messages out of order.
+            const splitTime = messageV2.createdAt.getTime();
+            const splitIdx = this.messages.findIndex(m => m.createdAt.getTime() > splitTime);
+            if (splitIdx === -1) {
+              this.messages.push(messageV2);
+            } else {
+              this.messages.splice(splitIdx, 0, messageV2);
+            }
           }
           // If no new parts, don't add anything (the sealed message already has all the content)
         } else {
--- a/packages/core/src/agent/message-list/message-list.ts
+++ b/packages/core/src/agent/message-list/message-list.ts
@@ -1018,21 +1018,34 @@ export class MessageList {
           if (shouldMergeIntoExisting) {
             MessageMerger.merge(existingMessage, messageV2);
             this.pushMessageToSource(existingMessage, messageSource);
-            // Sort messages and return early — existingMessage stays in messages[] and its Sets
-            this.messages.sort((a, b) => a.createdAt.getTime() - b.createdAt.getTime());
             return this;
+          }
+          // Preserve the newer createdAt so that replacing a message with one
+          // that has an older DB timestamp does not shift it before other messages.
+          if (existingMessage.createdAt > messageV2.createdAt) {
+            messageV2.createdAt = existingMessage.createdAt;
           }
           this.messages[existingIndex] = messageV2;
         }
       } else if (!exists) {
-        this.messages.push(messageV2);
+        if (messageSource === `memory`) {
+          // Insert memory messages at their chronological position instead of
+          // appending and sorting. A full-array sort breaks ordering when memory
+          // messages carry old DB timestamps while input messages use Date.now().
+          const insertTime = messageV2.createdAt.getTime();
+          const insertIndex = this.messages.findIndex(m => m.createdAt.getTime() > insertTime);
+          if (insertIndex === -1) {
+            this.messages.push(messageV2);
+          } else {
+            this.messages.splice(insertIndex, 0, messageV2);
+          }
+        } else {
+          this.messages.push(messageV2);
+        }
       }
 
       this.pushMessageToSource(messageV2, messageSource);
     }
-
-    // make sure messages are always stored in order of when they were created!
-    this.messages.sort((a, b) => a.createdAt.getTime() - b.createdAt.getTime());
 
     return this;
   }
--- a/packages/core/src/agent/message-list/tests/message-list-ordering.test.ts
+++ b/packages/core/src/agent/message-list/tests/message-list-ordering.test.ts
@@ -614,13 +614,12 @@ describe('Message ordering with identical timestamps (Issue #10683)', () => {
 
       const result = messageList.get.all.db();
       const roles = result.map(m => m.role);
+      const texts = result.map(m => m.content?.parts?.[0]?.text);
 
-      for (let i = 1; i < roles.length; i++) {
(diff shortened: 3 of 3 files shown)


.changeset/better-regions-sin.md
```diff
@@ -0,0 +1,5 @@
+---
+'@mastra/core': patch
+---
+
+Fixed message reordering when mixing input messages (Date.now() timestamps) with memory messages (old DB timestamps). The full-array sort in MessageList.addOne() could move historical assistant messages before current user messages, breaking the user/assistant alternating pattern and causing consecutive user messages to be merged.
```
TylerBarnes, line 5: I'm a bit confused by this description, the old behaviour you're describing sounds correct. Historical assistant messages should be moved before current user messages

packages/core/src/agent/message-list/tests/message-list-ordering.test.ts
```diff
@@ -476,4 +476,150 @@ describe('Message ordering with identical timestamps (Issue #10683)', () => {
...
+
+      // No consecutive user messages should appear
+      for (let i = 1; i < result.length; i++) {
+        if (result[i].role === result[i - 1]?.role && result[i].role !== 'assistant') {
+          throw new Error(
+            `Consecutive ${result[i].role} messages at index ${i - 1} and ${i}: ` +
+              `"${result[i - 1]?.content?.parts?.[0]?.text}" and "${result[i].content?.parts?.[0]?.text}"`,
+          );
+        }
+      }
```
TylerBarnes, line 518: better to just directly assert the order of messages

packages/core/src/agent/message-list/tests/message-list-ordering.test.ts
```diff
@@ -476,4 +476,150 @@ describe('Message ordering with identical timestamps (Issue #10683)', () => {
...
+      const inputUser2: MastraDBMessage = {
+        id: 'msg-user-2',
+        role: 'user',
+        content: { format: 2, parts: [{ type: 'text', text: 'What can you do?' }] },
+        createdAt: new Date(now + 2),
+      };
+
+      messageList.add([inputUser, inputAssistant, inputUser2], 'input');
+
+      // Memory loads same assistant with old timestamp + different content (triggers shouldReplace)
```
TylerBarnes, line 546: hmm if this situation actually happened, the memory message should be sorted older. memory createdAt timestamp is higher priority than whatever is in message list

packages/core/src/agent/message-list/tests/message-list-ordering.test.ts
```diff
@@ -476,4 +476,150 @@ describe('Message ordering with identical timestamps (Issue #10683)', () => {
...
+
+      // Content should be updated via shouldReplace
+      const assistantMsg = result.find(m => m.role === 'assistant');
+      expect(assistantMsg?.content.parts).toHaveLength(2);
+
+      // createdAt should be the newer timestamp from the input message
+      expect(assistantMsg?.createdAt.getTime()).toBeGreaterThanOrEqual(now);
+    });
+
+    it('should handle useChat without createdAt combined with observational memory', () => {
```
TylerBarnes, line 573: I don't see anything related to observational memory here

packages/core/src/agent/message-list/message-list.ts
```diff
@@ -1018,22 +1026,35 @@ export class MessageList {
...
+        } else {
+          this.messages.push(messageV2);
+        }
       }
 
       this.pushMessageToSource(messageV2, messageSource);
     }
 
-    // make sure messages are always stored in order of when they were created!
-    this.messages.sort((a, b) => a.createdAt.getTime() - b.createdAt.getTime());
```
TylerBarnes, line 1035: message should be sorted by date, if that breaks the order something else is wrong

packages/core/src/agent/message-list/message-list.ts
```diff
@@ -1018,22 +1026,35 @@ export class MessageList {
...
+          // messages carry old DB timestamps while input messages use Date.now().
+          const insertTime = messageV2.createdAt.getTime();
+          const insertIndex = this.messages.findIndex(m => m.createdAt.getTime() > insertTime);
+          if (insertIndex === -1) {
+            this.messages.push(messageV2);
+          } else {
+            this.messages.splice(insertIndex, 0, messageV2);
+          }
+        } else {
+          this.messages.push(messageV2);
```
TylerBarnes, line 1051: the messages.sort below should already handle this

TylerBarnes (changes requested):
The bug is real but the suggested/implemented fix is wrong. Timestamp ordering isn't the issue here, the issue seems to be merging a new user message onto an older one, though I'm not 100% sure. Asked for more info here https://github.com/mastra-ai/mastra/issues/14912#issuecomment-4171029825

PR #14917, opened 2026-04-01
fix(core): use path to directory where SKILL.md file is instead of to file itself

pzerelles:
https://github.com/mastra-ai/mastra/blob/9a43b476465e86c9aca381c2831066b5c33c999a/packages/core/src/processors/processors/skills.ts#L108

The `skill` tool tries to lookup a skill by name or location, but the location in the list of skills is the path to the SKILL.md file, not including the file's name. Lookup based on location fail currently because of this.

## Release Notes

* **Bug Fixes**
  * Updated the location field in available-skill metadata to reference skill paths directly instead of constructed file paths. This affects how skill locations are displayed in XML, JSON, and Markdown formatted outputs.


pzerelles:
I created the bug report -> #14918
Can you please reopen the PR?

PR #14944, opened 2026-04-01
fix(datadog): set native error tags to fix [object   Object] in Datadog Error Tracking UI

intojhanurag:
## Description
When a Mastra tool throws an error, the Datadog exporter only set `ddSpan.setTag('error', true)` — the error message, type, and stack trace were passed as LLM Obs annotations but never as native dd-trace span tags. This caused error details to appear as `[object Object]` in the Datadog UI and prevented Datadog's native Error Tracking UI from displaying the error
banner.

This PR:
- Adds `name` and `stack` fields to `SpanErrorInfo` so the original error class name and stack trace are preserved through the tracing pipeline
- Captures `error.name` and `error.stack` from the original Error object in `DefaultSpan.error()`
- Sets native dd-trace error tags (`error.message`, `error.type`, `error.stack`) on the span in both `emitSpanTree` and `emitSingleSpan`, with a fallback chain for `error.type`: `name →
category → 'Error'`

## Before
image

## After 
image

## Related Issue(s)

Fixes #14934

## Type of Change

- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [x] Test update

## Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Datadog error spans now show full error details (message, type, stack) instead of opaque objects, restoring complete error banners.

* **New Features**
  * Span error data now includes error class/name and stack trace so monitoring and error tracking display richer diagnostic context.

### DIFF
--- a/packages/core/src/observability/types/tracing.ts
+++ b/packages/core/src/observability/types/tracing.ts
@@ -425,6 +425,10 @@ export type AnySpanAttributes = SpanTypeMap[keyof SpanTypeMap];
 export interface SpanErrorInfo {
   message: string;
   id?: string;
+  /** Error class name (e.g. "TypeError", "ValidationError") */
+  name?: string;
+  /** Stack trace string */
+  stack?: string;
   domain?: string;
   category?: string;
   details?: Record<string, any>;
--- a/.changeset/fiery-ears-like.md
+++ b/.changeset/fiery-ears-like.md
@@ -1,0 +1,5 @@
+---
+'@mastra/datadog': patch
+---
+
+Fixed error spans showing as [object Object] in Datadog UI. The exporter now sets native dd-trace error tags (error.message, error.type, error.stack) so Datadog's Error Tracking UI displays the full error banner with message, type, and stack trace.
--- a/.changeset/smooth-clouds-glow.md
+++ b/.changeset/smooth-clouds-glow.md
@@ -1,0 +1,6 @@
+---
+'@mastra/core': patch
+'@mastra/observability': patch
+---
+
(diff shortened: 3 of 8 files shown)


packages/core/src/observability/types/tracing.ts
```diff
@@ -425,6 +425,10 @@ export type AnySpanAttributes = SpanTypeMap[keyof SpanTypeMap];
 export interface SpanErrorInfo {
   message: string;
   id?: string;
+  /** Error class name (e.g. "TypeError", "ValidationError") */
+  name?: string;
+  /** Stack trace string */
+  stack?: string;
```
epinzur, line 431: when this is set, does it go through any processing to ensure it doesn't get to long?  like the in the serializationOptions stuff around string length?

epinzur (approved):

packages/core/src/observability/types/tracing.ts
intojhanurag: Hey , Like message and name fields are typically short, but stack can grow large in deeply nested call chains . Should I add any truncation  for the stack ?

epinzur: first check if our auto-truncation stuff would already take care of it... like if `deepClean()` is called on it somewhere in the stack, then we are safe already.

epinzur: seems maybe we don't call `deeepClean()` on the error object... pls add it, probably in `observability/mastra/src/spans/default.ts`

intojhanurag: Yeah adding it

PR #14953, opened 2026-04-01
fix(core): auto-retry on assistant message prefill error

DanielSLew:
When a model doesn't support assistant message prefill (conversation ends with an assistant message), the harness now catches the error and automatically retries by sending a `[System] Continue where you left off.` user message. This follows the same pattern already used for the "Tool not found" auto-retry.

Previously this would just crash the session

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed session crashes caused by assistant message prefill incompatibilities: the app now automatically detects models that don't support assistant prefill and retries using a user-prefill fallback so runs no longer fail.
  * Outcome: more reliable conversations, preserved follow-up prompts, and fewer interrupted sessions.

### DIFF
--- a/packages/core/src/harness/harness.ts
+++ b/packages/core/src/harness/harness.ts
@@ -1372,6 +1372,20 @@ export class Harness<TState = {}> {
           requestContext: requestContextInput,
         });
         this.emit({ type: 'agent_end', reason: 'error' });
+      } else if (
+        error instanceof Error &&
+        /does not support assistant message prefill|must end with a user message/i.test(error.message)
+      ) {
+        this.emit({
+          type: 'error',
+          error: new Error('Model does not support assistant message prefill. Retrying with a user message.'),
+          retryable: true,
+        });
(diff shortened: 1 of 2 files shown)


CalebBarnes (approved):

packages/core/src/harness/harness.ts
```diff
@@ -1372,6 +1372,20 @@ export class Harness<TState = {}> {
...
+        error instanceof Error &&
+        /does not support assistant message prefill|must end with a user message/i.test(error.message)
+      ) {
+        this.emit({
+          type: 'error',
+          error: new Error('Model does not support assistant message prefill. Retrying with a user message.'),
+          retryable: true,
+        });
+        this.followUpQueue.push({
+          content: '[System] Continue where you left off.',
```
TylerBarnes: if you do `<system-reminder>There was an API error, please continue</system-reminder>` it will render in a nice box in the TUI

PR #15026, opened 2026-04-03
fix(cli): derive gateway and studio URLs from platform URL

roaminro:
Gateway and studio URLs now automatically follow the platform URL environment. If `MASTRA_PLATFORM_API_URL` points at staging, `MASTRA_GATEWAY_URL` and `MASTRA_STUDIO_URL` default to their staging equivalents. Each can still be overridden via its own env var.

Also fixes hardcoded `https://platform.mastra.ai` and `https://app.mastra.ai` in deploy error messages to use the derived studio URL instead.


.changeset/tired-geckos-invent.md
```diff
@@ -0,0 +1,41 @@
...
+'@mastra/deployer-vercel': patch
+'@mastra/deployer-cloud': patch
+'@mastra/editor': patch
+'@mastra/memory': patch
+'@mastra/server': patch
+'@mastra/core': patch
+'mastra': patch
+'@mastra/mcp': patch
+'@mastra/rag': patch
+'mastracode': patch
```
wardpeet:
```suggestion
'mastra': patch
```

wardpeet (approved):

PR #15037, opened 2026-04-04
adding initial /all-the-smoke slash command for feedback

mikhael28:
## Description

This is the code for the demo I gave earlier today, for the slash command `/all-the-smoke`. It's a good thing that i had a two-for-one demo promotion, because this one has a lot in common with the existing Claude skill /smoke-test which I believe has room for improvement. Running it locally with Claude Code Opus 4.6, it failed to initially detect that a Studio instance was running on localhost (which it was) and in-general took too long to finish with inconsistent results. Perhaps I wasn't running it correctly, but the time it took to run seemed excessive as it kept hitting various tool call timeouts.

This PR is meant to be a prototype, a jumping off point for discussing how we might want to 'productionize' smoke-testing. This command spins up a new mastra studio project with pnpm, and then walk through a few headless browser testing scenarios with Playwright based off of different Markdown files that detail the possible surface area of the runner to walk through. My thinking was that engineers who have greater visibility into different parts of the codebase would update the markdown with their specific smoke testing instruction set, so that if we wanted to walk through the entire application, it would have a bunch of high-signal test cases to work through. 

@roaminro you mentioned in the demo call that you had some Playwright automation set up that you use for smoke testing - what are your thoughts on this? What is your current approach towards using headless browser testing? Is Markdown the right medium? Is this even the right solution?

Again, it's a prototype, not production quality, but in my limited testing it worked a bit better (and faster) than the previous /smoke-test command.

## Related Issue(s)

NA

## Type of Change

- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [x] Test update

## Checklist

### DIFF
--- a/.claude/commands/all-the-smoke.md
+++ b/.claude/commands/all-the-smoke.md
@@ -1,0 +1,1 @@
+../../.mastracode/commands/all-the-smoke.md
--- a/.claude/skills/smoke-test/SKILL.md
+++ b/.claude/skills/smoke-test/SKILL.md
@@ -180,7 +180,7 @@ Based on the selected LLM provider, check for the required API key:
 
 ### Step 4: Start the Development Server
 
-Navigate to the project directory and start the dev server:
+Navigate to the generated project directory and start the dev server:
 
 ```sh
 cd <directory>/<project-name>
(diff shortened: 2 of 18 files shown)


mikhael28:
Closing, as @NikAiyer smoke tests in prod do a much finer job :)

PR #15050, opened 2026-04-04
fix(core): Corrupt multi-step message tracking behavior

### LINKED ISSUE #15049: `ProcessorRunner.applyMessagesToMessageList` corrupts response message tracking in multi-step agent
This issue was created from Discord post 1489894903981936753:

[Open in Discord](https://discord.com/channels/1309558646228779139/1489894903981936753)

Hey team, found a message persistence bug that I believe is in `@mastra/core`'s `ProcessorRunner.applyMessagesToMessageList`.

**TL;DR:** In multi-step agent executions (router delegates to sub-agent), the final assistant response is lost from memory when Observational Memory's token threshold isn't reached. The message streams fine but disappears on refresh.

The issue: when a `processOutputStep` processor returns messages unchanged, `applyMessagesToMessageList` does an unconditional remove-and-re-add of every message. This corrupts the `newResponseMessages` Set tracking when there's a single merged assistant message spanning both steps. `response.db()` returns empty after the re-add.

When OM's token threshold IS reached (large delegation results), `rotateResponseMessageId` fires, step 2 gets a new ID, creates separate messages — and those survive the re-add. So the bug only manifests with small/empty delegation results.

Disabling OM on the router eliminates it. 

Fix suggestion: skip the remove/re-add in `applyMessagesToMessageList` when the message object reference is already in the messageList (same ref = no-op, processor mutated in-place or returned unchanged).

1. Create a router agent with `observationalMemory: { enabled: true }` and a `processOutputStep` processor
2. Configure a sub-agent with tools (search, etc.)
3. Ask a question that triggers delegation where the sub-agent returns a **small** result (few tokens — e.g., empty search results, or a small structured output)
4. The router streams the full response correctly
5. Refresh the page — **the final assistant text is gone**
6. Repeat with a query that produces a **large** delegation result (many tokens) — it works

**I am currently testing a solution and testing the implementation, can open a PR once testing is finished. ** and after dane 🐶  finishes the issue report (I do not have time to create a reproducable repo at the moment & working directly with forked mastra to test interally)

kevintyj:
## Description

`ProcessorRunner.applyMessagesToMessageList` unconditionally removes every message from the `MessageList` by ID and re-adds it when a `processOutputStep` processor returns a `MastraDBMessage[]`. This corrupts the `MessageStateManager`'s `newResponseMessages` Set tracking, causing `messageList.get.response.db()` to return empty after the re-add. The result is that the final assistant text from step 2 streams correctly to the client but is missing on page refresh.

The bug only manifests when:
1. Step 1 and step 2 content are merged into a single assistant message (same ID — `rotateResponseMessageId` wasn't called because Observational Memory's token threshold wasn't reached)
2. A `processOutputStep` processor returns the messages array unchanged (common no-op case)
3. The `removeByIds` → `add` round-trip creates a new object reference that mismatches the Set

**Fix:** Before the loop, build a lookup of current messages by ID. Skip the remove/re-add for any message whose object reference is already in the `MessageList`. Only perform the remove/re-add for messages that are genuinely new or replaced (different object with the same ID).

## Related Issue(s)

Fixes #15049 

## Type of Change

- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [x] Test update

## Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed message-tracking corruption where multi-step assistant responses could be lost after refresh when processors returned unchanged messages.

* **Tests**
  * Added comprehensive tests ensuring message-list state remains correct across unchanged, merged, replaced, deleted, new, input-tracked, and mixed-message scenarios.

### DIFF
--- a/.changeset/fix-corrupt-message-tracking.md
+++ b/.changeset/fix-corrupt-message-tracking.md
@@ -2,4 +2,4 @@
 '@mastra/core': patch
 ---
 
-fix(processors): prevent applyMessagesToMessageList from corrupting response message tracking when processor returns unchanged messages
+Fixed an issue where multi-step assistant responses could disappear after refresh when an output processor returned messages unchanged.
--- a/.changeset/fix-corrupt-message-tracking.md
+++ b/.changeset/fix-corrupt-message-tracking.md
@@ -1,0 +1,5 @@
+---
+'@mastra/core': patch
+---
+
+fix(processors): prevent applyMessagesToMessageList from corrupting response message tracking when processor returns unchanged messages
--- a/packages/core/src/processors/runner.test.ts
+++ b/packages/core/src/processors/runner.test.ts
@@ -2045,4 +2045,161 @@ describe('ProcessorRunner', () => {
       expect(stateInOutputResult.errorInfo).toEqual({ toolName: 'myTool', error: 'something broke' });
     });
   });
+
+  describe('applyMessagesToMessageList', () => {
+    it('should preserve response message tracking when processor returns unchanged messages', () => {
+      // Simulate: add a response message, then call applyMessagesToMessageList
+      // with the same objects (no-op processor). The response tracking must survive.
+      const msg = createMessage('assistant reply', 'assistant');
+      messageList.add(msg, 'response');
+
+      const allBefore = messageList.get.all.db();
+      expect(allBefore).toHaveLength(1);
+      expect(messageList.get.response.db()).toHaveLength(1);
+
+      const check = messageList.makeMessageSourceChecker();
+      const idsBeforeProcessing = allBefore.map(m => m.id);
+
+      // Processor returns the exact same array/objects — common no-op case
+      ProcessorRunner.applyMessagesToMessageList(allBefore, messageList, idsBeforeProcessing, check, 'response');
+
+      // Response tracking must still work
+      expect(messageList.get.response.db()).toHaveLength(1);
+      expect(messageList.get.response.db()[0].id).toBe(msg.id);
+    });
+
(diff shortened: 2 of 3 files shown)


TylerBarnes (changes requested):
From Mastra Code: This looks like the right shape overall.

Could we add one focused test for the object-identity skip path with an in-place mutation? The fix relies on the premise that if a processor returns the same object reference, the `MessageList` already reflects any mutation and we should avoid remove/re-add to preserve source tracking.

A cloned no-op same-ID assistant response test would also be useful to confirm the non-identity path still preserves response tracking via the source snapshot.

graysonhicks:
I could not reproduce the bug on latest `main` (`95148b0e41`).

Evidence:

- Exact OM lifecycle simulation passed: response → persisted/cleared → re-added as memory → same-ID step-2 merge → no-op `processOutputStep`.
- Final text remained in `response.db()` and `drainUnsavedMessages()`.
- History ordering and source classification remained intact.
- Existing OM and output-processor suites passed: 14/14 tests.
- Temporary regression checks passed: 2/2 tests, then removed. Worktree is clean.

The original behavior appears fixed by:

- `e06a1598ca`: corrects memory→response source promotion.
- `fbeda0c0f3`: preserves processor message boundaries with `merge: false`.

Conclusion: #15050 should not merge as a bug fix without a new failing reproduction on current `main`. The remaining helper consolidation would need to be proposed separately as a refactor.

abhiaiyer91:
Thanks for working on this! This has since been addressed by #22146, which is now merged, so I'm closing this PR as superseded.

PR #15065, opened 2026-04-05
fix(core): infer TState from stateSchema in createWorkflow

octo-patch:
Fixes #14627

## Problem

When `stateSchema` is provided to `createWorkflow()`, the `TState` generic is not inferred from it. This causes the `state` parameter in `dowhile`/`dountil` condition callbacks (and branch conditions) to be typed as `unknown` instead of the schema's inferred type.

## Solution

Changed `WorkflowConfig.stateSchema` from `PublicSchema<TState>` to `StandardSchemaWithJSON<TState>`.

The root cause: `PublicSchema<TState>` is a union type that includes `JSONSchema7` (which has no type parameter). When TypeScript tries to infer `TState` from this union, the presence of the unparameterized `JSONSchema7` member prevents reliable inference, causing `TState` to fall back to its default of `unknown`.

`StandardSchemaWithJSON<TState>` is a simpler intersection type that allows TypeScript to directly infer `TState` — the same pattern already used in `createStep`, where `state` is correctly typed.

All existing callers pass Zod schemas (which implement the Standard Schema interface), so this is backwards-compatible. `toStandardSchema()` in the constructor still works since `StandardSchemaWithJSON` extends `PublicSchema`.

## Testing

Added type-level tests in `workflow-schema-types.test-d.ts` that verify:
1. `state` is correctly typed when `stateSchema` is provided to `createWorkflow`
2. `state` is `unknown` when `stateSchema` is not provided (expected behavior)

## Bug Fixes
* Improved TypeScript type inference for workflow state parameters in loop conditions. The state object is now properly typed based on the provided state schema definition, enabling better IDE autocomplete, type checking, and developer experience.

### DIFF
--- a/.changeset/fix-workflow-state-schema-inference.md
+++ b/.changeset/fix-workflow-state-schema-inference.md
@@ -1,0 +1,28 @@
+---
+'@mastra/core': patch
+---
+
+Fixed `createWorkflow` to correctly infer `TState` from `stateSchema`, so `state` is properly typed in `dowhile`/`dountil` conditions instead of being typed as `unknown`.
+
+**Before:** Passing `stateSchema` to `createWorkflow` did not propagate the type to loop conditions.
+
+```ts
+const workflow = createWorkflow({
+  id: 'my-workflow',
+  inputSchema: z.object({}),
+  outputSchema: z.object({}),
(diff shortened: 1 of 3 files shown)


wardpeet:
Public schema is actually correct, thanks for opening the PR but it seems like stateSchema is working.

PR #15082, opened 2026-04-05
fix: prune mastracode chat memory growth

TylerBarnes:
This keeps mastracode from holding onto every old TUI chat component in long sessions, which was causing memory usage to grow over time.

Before, the chat container kept growing without a limit. Now it prunes older components once it gets too large (200 components) and keeps the most recent 100 around. Pruning from 200->100 should keep flicker down, if we always pruned 200+->200 it would do a full re-render on every change above 200 components

```ts
pruneChatContainer(state)
```

I also included the follow-up mastracode harness API/type fixes that were needed after merging main, plus the prune test cleanup and lint/import fixes so the branch is green again.

## Summary by CodeRabbit

* **New Features**
  * Automatic conversation history optimization during long sessions prevents memory issues while preserving recent messages and important UI elements

* **Refactor**
  * Subagent selection now uses standardized type options
  * Improved default value handling for missing configuration parameters

* **Tests**
  * Added test coverage for conversation history management


mastracode/src/agents/tools.ts
```diff
@@ -2,16 +2,18 @@ import { createAnthropic } from '@ai-sdk/anthropic';
...
 
+type MastraCodeState = z.infer<typeof stateSchema>;
+
 /** Minimal shape for tools passed to createDynamicTools. */
-interface ToolLike {
-  execute?: (input: unknown, context?: unknown) => Promise<unknown> | unknown;
-  [key: string]: unknown;
-}
+type ToolLike = {
+  execute?: (...args: any[]) => Promise<unknown> | unknown;
```
DanielSLew, line 15: why this change?

mastracode/src/tools/request-sandbox-access.ts
```diff
@@ -17,6 +18,8 @@ function expandTilde(p: string): string {
   return p;
 }
 
+type MastraCodeState = z.infer<typeof stateSchema>;
```
DanielSLew, line 21: You define this in so many places, maybe should just be exported from schema.js

abhiaiyer91 (approved):

mastracode/src/tui/commands/subagents.ts
```diff
@@ -116,30 +116,23 @@ async function showSubagentScopeThenList(
...
-          label: 'Plan',
-          description: 'Read-only analysis and planning',
-        },
-        {
-          id: 'execute',
-          label: 'Execute',
-          description: 'Task execution with write access',
-        },
-      ];
+  const agentTypes: Array<{ id: string; label: string; description: string }> = [
```
DanielSLew, line 119: do we not let subagents get configured anymore?

PR #15084, opened 2026-04-05
fix(deployer-cloudflare): mark Node.js builtins as external in Rollup bundle

abhiaiyer91:
Temporary debug logging to diagnose why cloudflare e2e tests return HTML instead of JSON in CI.

Logs:
- Response body and status for both test endpoints
- Output dir contents, package.json, entry file, wrangler config
- node_modules contents

Will revert after diagnosis.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed improper bundling of Node.js builtin modules in Cloudflare Worker deployments.


wardpeet (approved):
could be an issue becaue we swapped to web deploying, i think @NikAiyer made that swtich. But i can approve this

PR #15102, opened 2026-04-06
feat(memory): per-record config overrides for observation & reflection thresholds

abhiaiyer91:
## Summary

OM threshold decisions now check the record's stored config first, falling back to the instance-level config. This enables per-thread (or per-resource) threshold overrides without restarting the process or invalidating the OM instance cache.

## What changed

Added two private helpers to `ObservationalMemory`:
- `getEffectiveMessageTokens(record)` — resolves observation `messageTokens` from `record.config.observation.messageTokens`, falling back to `this.observationConfig.messageTokens`
- `getEffectiveReflectionTokens(record)` — resolves reflection `observationTokens` from `record.config.reflection.observationTokens`, falling back to `this.reflectionConfig.observationTokens`

Patched **6 call sites** across threshold-decision points:

| Method | What changed |
|--------|-------------|
| `meetsObservationThreshold()` | Uses per-record observation threshold |
| `getStatus()` | Observation threshold, reflection threshold, and shared budget UI all use per-record values |
| `reflect()` | Reflection token budget passed to reflector uses per-record value |
| `swapBufferedToActive()` | Buffer activation ratio uses per-record observation threshold |

## Motivation

The gateway needs to expose an API for users to set per-thread OM thresholds (e.g. lower the observation threshold for a specific thread to trigger faster observations). Today the OM instance caches config at creation time, and updating the record's config in MongoDB has no effect because all threshold checks read from instance-level config.

## Backward compatibility

Fully backward-compatible. If a record has no per-record override (the default), the helpers fall back to the instance-level config — identical to previous behavior.

## Test plan

- `pnpm tsc --noEmit` passes
- All existing OM tests pass (404/405, 1 unrelated failure due to missing `OPENROUTER_API_KEY`)
- Verified end-to-end: update a record's `config.observation.messageTokens` in MongoDB → next request respects the override

## Summary by CodeRabbit

* **New Features**
  * Added gateway observational memory progress monitoring and polling for improved memory status visibility.
  * Enhanced thread cloning with full gateway support for multi-thread operations.

* **Bug Fixes**
  * Fixed task and plan state carryover when switching between threads.

* **Tests**
  * Improved mock harness implementation for better test coverage.

### DIFF
--- a/packages/memory/src/processors/observational-memory/observational-memory.ts
+++ b/packages/memory/src/processors/observational-memory/observational-memory.ts
@@ -800,13 +800,31 @@ export class ObservationalMemory {
 
   /**
    * Resolve the effective messageTokens for a record.
-   * Uses the record's stored config if it has a per-record override,
-   * otherwise falls back to the instance-level config.
+   * Only explicit per-record overrides (stored under `_overrides`) win;
+   * the initial config snapshot written by getOrCreateRecord() is ignored
+   * so that later instance-level changes still take effect.
+   *
+   * Overrides that fall below the instance-level buffering floor
+   * (bufferTokens / absolute bufferActivation) are clamped to the
+   * instance threshold to preserve buffering invariants.
    */
   private getEffectiveMessageTokens(record: ObservationalMemoryRecord): number | ThresholdRange {
-    const recordConfig = record.config as { observation?: { messageTokens?: number | ThresholdRange } } | undefined;
-    const recordTokens = recordConfig?.observation?.messageTokens;
-    if (recordTokens != null) {
+    const overrides = (record.config as { _overrides?: { observation?: { messageTokens?: number | ThresholdRange } } })
+      ?._overrides;
+    const recordTokens = overrides?.observation?.messageTokens;
+    if (recordTokens) {
+      const maxOverride = getMaxThreshold(recordTokens);
+
+      // Clamp: override must not violate instance-level buffering invariants
+      const bufferTokens = this.observationConfig.bufferTokens;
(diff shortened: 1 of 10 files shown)


packages/core/src/harness/harness.ts
```diff
@@ -1013,6 +1013,19 @@ export class Harness<TState = {}> {
...
+    // Try external provider first (e.g., gateway)
+    if (this.config.omProgressProvider) {
+      try {
+        const event = await this.config.omProgressProvider(this.currentThreadId, this.resourceId);
+        if (event) {
+          this.emit(event);
+          return;
+        }
+      } catch {
+        // Fall through to local storage
```
TylerBarnes: I wonder if we should actually error here

packages/memory/src/processors/observational-memory/observational-memory.ts
```diff
@@ -798,6 +798,34 @@ export class ObservationalMemory {
...
 
+  /**
+   * Resolve the effective messageTokens for a record.
+   * Uses the record's stored config if it has a per-record override,
+   * otherwise falls back to the instance-level config.
+   */
+  private getEffectiveMessageTokens(record: ObservationalMemoryRecord): number | ThresholdRange {
+    const recordConfig = record.config as { observation?: { messageTokens?: number | ThresholdRange } } | undefined;
+    const recordTokens = recordConfig?.observation?.messageTokens;
+    if (recordTokens != null) {
```
TylerBarnes: maybe we should just do `if (recordTokens)`?

TylerBarnes (approved):
looks good! just one question and needs a changeset too

PR #15103, opened 2026-04-06
fix(core): isolate subagent from parent's OM thread context

abhiaiyer91:
## Problem

When a parent agent spawns a subagent via `createSubagentTool`, the parent's `requestContext` was forwarded as-is. This meant the subagent's conversation would trigger OM enrichment (observe/reflect) on the **parent's** memory thread, corrupting the parent's observation history with subagent conversation fragments.

## Fix

Create a shallow copy of the `requestContext` for the subagent with:
- `threadId` set to `null`
- `resourceId` set to `''`

All other harness state (sandbox allowed paths, `emitEvent`, etc.) is preserved so the subagent still inherits the execution environment — it just won't write to the parent's OM thread.

## Test plan

Updated `subagent-tool.test.ts`:
- Verifies subagent receives a **new** RequestContext instance (not the parent's)
- Verifies `threadId` is `null` and `resourceId` is `''` on the subagent's harness context
- Verifies other harness fields (`emitEvent`) are preserved
- Verifies non-harness context data (`custom-key`) is forwarded

ELI5: When a parent agent starts a helper agent (subagent), the helper used to write notes into the parent's notebook by accident. This change gives the helper its own clean notebook so it can't mess up the parent's notes.

* What changed
  * Fix: createSubagentTool now forwards a shallow-cloned RequestContext to subagents. The clone preserves harness fields but sets harness.threadId = null and harness.resourceId = '' so subagents cannot write to the parent's observation/memory thread.
  * Implementation: tools.ts constructs subagentRequestContext (new RequestContext) when a parent context exists and passes that to the subagent stream call instead of the original reference.

* Tests
  * Updated subagent-tool.test.ts to assert:
    - the subagent receives a new RequestContext instance (not the parent's),
    - harness.threadId is null and harness.resourceId is '' on the forwarded context,
    - other harness fields (e.g., emitEvent, abortSignal) are preserved,
    - non-harness context entries (e.g., custom-key) are copied through,
    - stream options use a RequestContext instance and preserve maxSteps/stopWhen/abortSignal/requireToolApproval semantics.

* Other
  * Added a changeset entry documenting the patch release and the fix preventing subagents from corrupting parent agent memory threads.


packages/core/src/harness/tools.ts
```diff
@@ -445,15 +436,24 @@ Use this tool when:
...
       const toolCallLog: Array<{ name: string; toolCallId: string; isError?: boolean }> = [];
 
+      // Build a request context for the subagent that inherits sandbox paths
+      // and harness state but strips threadId/resourceId so the subagent
+      // doesn't trigger OM enrichment on the parent's memory thread.
+      let subagentRequestContext: RequestContext | undefined;
+      if (context?.requestContext) {
+        subagentRequestContext = new RequestContext(context.requestContext.entries());
+        if (harnessCtx) {
+          subagentRequestContext.set('harness', { ...harnessCtx, threadId: null, resourceId: '' });
```
wardpeet, line 456: We should try to rename these in the future to something private with `_` or something

wardpeet (approved):
You need a changeset @abhiaiyer91

CalebBarnes (approved):

PR #15119, opened 2026-04-07
fix(client-js): correct Vector SDK return types to match server responses

octo-patch:
Fixes #15089

## Problem

Three methods in `@mastra/client-js` Vector resource had TypeScript return types that didn't match what the server actually returns, causing silent runtime failures:

| Method | SDK declared | Server returns |
|---|---|---|
| `getIndexes()` | `Promise<{ indexes: string[] }>` | `string[]` |
| `upsert()` | `Promise<string[]>` | `{ ids: string[] }` |
| `query()` | `Promise<QueryVectorResponse>` | `QueryResult[]` |

## Solution

Updated return types in `client-sdks/client-js/src/resources/vector.ts` to match the server's Zod response schemas (in `packages/server/src/server/schemas/vectors.ts`):

- `getIndexes()` → `Promise<string[]>`
- `upsert()` → `Promise<{ ids: string[] }>`
- `query()` → `Promise<QueryResult[]>` (using `QueryResult` from `@mastra/core/vector`)

Also marked `QueryVectorResponse` as `@deprecated` in `types.ts` since it describes a shape that never existed in the server response.

## Testing

The fix aligns the SDK types with the existing server Zod schemas (`listIndexesResponseSchema`, `upsertVectorsResponseSchema`, `queryVectorsResponseSchema`) which serve as the source of truth for what the server returns.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed incorrect TypeScript return types in Vector SDK methods (`getIndexes()`, `upsert()`, `query()`) to match actual server responses.
  * Deprecated `QueryVectorResponse` type; use `QueryResult[]` from `@mastra/core/vector` instead.

### DIFF
--- a/.changeset/fix-vector-sdk-return-types.md
+++ b/.changeset/fix-vector-sdk-return-types.md
@@ -1,0 +1,13 @@
+---
+'@mastra/client-js': patch
+---
+
+Fixed Vector SDK return types to match actual server responses.
+
+Three methods had incorrect TypeScript return types that caused runtime failures even though TypeScript compilation succeeded:
+
+- `getIndexes()` now returns `Promise<string[]>` instead of `Promise<{ indexes: string[] }>`
+- `upsert()` now returns `Promise<{ ids: string[] }>` instead of `Promise<string[]>`
+- `query()` now returns `Promise<QueryResult[]>` instead of `Promise<QueryVectorResponse>`
+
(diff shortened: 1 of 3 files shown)


intojhanurag:
Hey @octo-patch Thanks for the PR ! It needs discussion , then we will raise PR .

PR #15124, opened 2026-04-07
Support exporting new signal types in CloudExporter

noahcrowley:
### Summary

- Extend `CloudExporter` to handle all five Observability Bus signals (traces, logs, metrics, scores, feedback) with per-signal endpoints, buffering, and batch upload
- Add `sanitizeForJson` to safely serialize payloads containing circular references, `Map`, `Set`, `BigInt`, `Symbol`, `Function`, and `Error` objects — replacing raw `JSON.stringify` across all signals
- Change `MastraCloudSpanRecord` from a hand-picked interface to an `AnyExportedSpan & {...}` intersection so all current and future `ExportedSpan` fields (tags, entityType/Id/Name, isRootSpan, etc.) flow through automatically
- Fix flush semantics: size-triggered flushes no longer block `exportTracingEvent` callers (`void` instead of `await`), and `flush()` now tracks in-flight uploads so it drains fully before resolving — critical for serverless shutdown

### Details

**Multi-signal endpoints**

Each signal gets its own configurable endpoint (`tracesEndpoint`, `logsEndpoint`, `metricsEndpoint`, `scoresEndpoint`, `feedbackEndpoint`). When not explicitly set, sibling endpoints are derived from the traces endpoint by replacing `/spans/publish` with `/{signal}/publish`. Falls back to the traces endpoint if the URL pattern doesn't match.

**Safe serialization**

`safeJsonStringify` pre-walks the payload tree before `JSON.stringify`, handling:
- Circular references → `[Circular]`
- `Error` → `{ name, message, stack, cause }`
- `Map` → plain object, `Set` → array
- `BigInt` → string, `Function` → `[Function name]`, `Symbol` → `Symbol.toString()`

**Flush reliability**

- `handleBufferedEvent` uses `void this.flush()` so event ingestion is non-blocking
- `startTrackedFlush` / `inFlightFlushes` tracks background uploads
- `flush()` loops until both buffer and in-flight set are empty
- Tested with deferred promises to verify `exportTracingEvent` resolves before upload completes, and `flush()` blocks until uploads finish

### Test plan

- [ ] Verify all five signals buffer, batch, and upload to correct endpoints
- [ ] Verify disabled exporter ignores all signal types
- [ ] Verify shutdown clears timers and flushes remaining events

## ELI5

This PR upgrades CloudExporter so it can send five kinds of observability data (traces, logs, metrics, scores, feedback) to Mastra Cloud instead of only traces. It adds per-signal endpoints (with sensible derivation from the traces endpoint), a safe JSON serializer to avoid payload crashes, and fixes flush/upload behavior so events aren't lost during concurrent uploads or shutdown.

## Overview

Extends CloudExporter from traces-only to multi-signal support (traces, logs, metrics, scores, feedback) with per-signal buffering, batching, concurrent uploads, configurable/derived endpoints, robust serialization, and corrected flush/shutdown semantics.

### Multi‑Signal Support
- New public handlers: onLogEvent(event), onMetricEvent(event), onScoreEvent(event), onFeedbackEvent(event).
- Each signal maintains its own buffer and batching rules; flushes perform concurrent per-signal uploads (one request per signal).
- Shared helpers added for buffering/flush decision logic (handleBufferedEvent, batch/flush helpers).

### Endpoint Resolution & Config
- `endpoint` is now documented/treated as a "base cloud endpoint" (docs default updated to https://api.mastra.ai).
- Added explicit per-signal config fields: tracesEndpoint?, logsEndpoint?, metricsEndpoint?, scoresEndpoint?, feedbackEndpoint?.
- If a sibling endpoint is not explicitly set, it is derived from the traces endpoint by replacing the path segment `/spans/publish` with `/{signal}/publish`; if that pattern doesn't match, the traces endpoint is used as a fallback.
- Validation enforces the base endpoint is an origin-like URL (no path/search/hash) where appropriate and prefers explicit overrides when provided.

### Safe JSON Serialization
- Replaced raw JSON.stringify with sanitizeForJson / safeJsonStringify which pre-walks payloads to handle:
  - Circular references → "[Circular]"
  - Error objects → { name, message, stack, cause }
  - Map → plain object; Set → array
  - BigInt → string; Function → "[Function name]"; Symbol → Symbol.toString()

### Span Record Shape & Types
- MastraCloudSpanRecord changed from a hand-picked interface to an intersection type AnyExportedSpan & { ... } so existing and future ExportedSpan fields (tags, entityType/entityId/entityName, isRootSpan, etc.) are preserved.
- Span formatting now largely forwards ExportedSpan fields and normalizes error → errorInfo ?? null.

### Flush Semantics & Concurrency
- Size-triggered flushes no longer block ingestion callers (handleBufferedEvent uses void this.flush()).
- Background uploads are tracked with startTrackedFlush / inFlightFlushes so flush() waits for both buffer drains and any in-flight uploads — improving reliability for serverless shutdown and overlapping flush calls.
- flushBuffer() concurrently attempts uploads for all signal batches; individual signal failures are logged and recorded rather than aborting the whole flush.
- Ensures timers are cleared and shutdown triggers a draining flush of remaining events.

## Tests
- observability/mastra/src/exporters/cloud.test.ts expanded significantly:
  - Added typed factory helpers for LogEvent, MetricEvent, ScoreEvent, FeedbackEvent and createDeferred to test concurrency.
  - New tests validate buffering, batching, and upload for all five signals to correct derived or overridden endpoints (e.g., /ai/logs/publish, /ai/metrics/publish, /ai/scores/publish, /ai/feedback/publish, /ai/spans/publish).
  - Tests cover endpoint derivation/override behavior and rejecting legacy/invalid endpoint path forms.
  - Concurrency tests assert size-triggered flushes do not block ingestion and that flush() waits for already-started in-flight uploads.
  - Tests ensure disabled exporter ignores all signal handlers and that shutdown clears timers and flushes remaining events.

## Docs & Release
- Updated CloudExporter reference docs to document base endpoint semantics, tracesEndpoint override, new per-signal handlers, and generalized "Signal Processing" behavior.
- Added changeset (minor) for @mastra/observability describing multi-signal CloudExporter and base-endpoint derivation; notes call out the endpoint format/derivation behavior as a breaking consideration.

## Risk / Review Notes
- Large surface-area change: new public handlers, buffers, endpoint logic, serialization, and flush concurrency — review effort: High.
- Verification checklist: end-to-end buffering/batching/upload for all five signals; endpoint derivation/overrides; safe serialization of complex payloads; non-blocking ingestion and proper drain on flush/shutdown.


observability/mastra/src/exporters/cloud.ts
```diff
@@ -13,36 +20,187 @@ export interface CloudExporterConfig extends BaseExporterConfig {
...
+function serializeError(error: Error, activeRefs: WeakSet<object>): Record<string, unknown> {
+  return {
+    name: error.name,
+    message: error.message,
+    stack: error.stack,
+    ...(error.cause !== undefined ? { cause: sanitizeForJson(error.cause, activeRefs) } : {}),
+  };
+}
+
+function sanitizeForJson(value: unknown, activeRefs = new WeakSet<object>()): unknown {
```
epinzur: I think this is basically a duplicate of the existing [deepClean](https://github.com/mastra-ai/mastra/blob/main/observability/mastra/src/spans/serialization.ts#L211) method in observability.

Maybe re-use that instead of re-inventing here?

Also I think we are already calling `deepClean()` on most (if not all) paths when exporting observability signals. So you could potentially skip this entirely here.

epinzur (approved):
looks good to me. Added one comment about potentially using the existing `deepClean()` instead of making a new json sanitizer here.

Also I'm wondering if the split of the different endpoints is an over optimization at this point.

The buffer logic would def be simpler if everything could be pushed to the same url.

observability/mastra/src/exporters/cloud.ts
```diff
@@ -171,6 +441,20 @@ export class CloudExporter extends BaseExporter {
...
+        this.inFlightFlushes.delete(flushPromise);
+      },
+      () => {
+        this.inFlightFlushes.delete(flushPromise);
+      },
+    );
+    return flushPromise;
+  }
+
   private async flushBuffer(): Promise<void> {
```
epinzur, line 448: pls override `flush()` and `shutdown()` from the based class and make sure we are calling `flushBuffer()` inside both.

see: https://github.com/mastra-ai/mastra/blob/main/observability/mastra/src/exporters/base.ts#L248-L257

noahcrowley: Done!

PR #15159, opened 2026-04-07
feat(cli): add `mastra server env` commands

mikhael28:
Adds `mastra server env` CLI commands so you can manage environment variables on deployed servers directly from the terminal, similar to `vercel env`.

**New commands:**

```bash
mastra server env list                    # list env vars (values masked)
mastra server env set FOO bar             # set a single var
mastra server env unset FOO               # remove a var
mastra server env import .env.production  # merge vars from a .env file
```

All commands resolve the linked project from `.mastra-project.json` (created on first deploy). The `import` command merges with existing vars — it won't remove anything not in the file.

This was a pain point where after initial deploy, updating env vars required going to the production UI. Now it's a one-liner from the CLI.

## ELI5

This PR adds simple terminal commands so you can view, add, remove, or bulk-import environment variables on a deployed Mastra server without opening the web UI.

## Changes Overview

Adds a new CLI subtree: mastra server env with four subcommands:
- list — lists environment variables (values masked: shows up to first 4 chars, then "..." if longer)
- set <key> <value> — sets or updates a single env var
- unset <key> — removes an env var if present
- import <file> — parses a .env-style file and merges its vars into the server (new values override; does not delete other vars)

All commands resolve authentication (uses getToken and MASTRA_ORG_ID or the current org) and determine the target project ID from MASTRA_PROJECT_ID or by loading the linked project config (.mastra-project.json created on first deploy). Each command supports an optional -c/--config flag.

## Implementation Details

- New CLI handlers: packages/cli/src/commands/server/env.ts
  - envListAction: fetches env map, sorts keys, prints each key with masked value; prints a friendly message if none set.
  - envSetAction: loads current env map, sets/overwrites the given key, PUTs updated map, and prints success.
  - envUnsetAction: loads current env map, no-ops (with message) if key absent, otherwise deletes key, PUTs updated map, and prints success.
  - envImportAction: resolves and reads the provided file (errors if unreadable), parses into key/value pairs, exits with a message if the file contains no variables, merges parsed vars into existing env map (new values override), PUTs the merged map, and prints the imported keys and count.

- New platform API helpers: packages/cli/src/commands/server/platform-api.ts
  - getServerProjectEnv(token, orgId, projectId): GET /v1/server/projects/{id}/env → returns data.envVars; throws on API error.
  - updateServerProjectEnv(token, orgId, projectId, envVars): PUT /v1/server/projects/{id}/env with JSON body { envVars }; throws on API error.

- CLI registration: packages/cli/src/index.ts
  - Registers the four subcommands under mastra server env and wires them through existing wrapAction helper; subcommands accept -c/--config where applicable.

## Other Notes

- Import is merge-only (does not remove existing server variables not present in the file).
- Values are masked in listings to reduce secret leakage (first 4 chars shown, "..." appended when longer).
- Changeset: .changeset/lucky-rivers-lie.md marks this as a minor release for the mastra package.
- Related API/type updates (platform declarations) add/adjust server env endpoints and request/response typing to support the new operations.


mikhael28:
@coderabbitai review

abhiaiyer91 (approved):

PR #14624, opened 2026-03-24
fix(schema-compat): improve provider structured output and tool-call compatibility

wardpeet:
## Summary

This PR improves schema compatibility across provider adapters, with a focus on structured outputs and tool calls that use optional, nullable, defaulted, and date-like fields.

### User-facing changes

- OpenAI-compatible schemas now encode nullable optional fields using `anyOf` instead of JSON Schema type arrays, which improves compatibility with LanguageModel v1 and OpenAI strict-mode style validators.
- Date-like fields are now validated more reliably when providers return ISO date strings for `z.date()` fields.
- Provider e2e coverage has been expanded so the same many-schema structured-output and tool-call flows are exercised across Anthropic, Google, OpenAI, and OpenAI-reasoning adapters.

### `packages/schema-compat`

- Restored the OpenAI nullable-optional conversion so optional properties become nullable via `anyOf` instead of `type: [..]` arrays.
- Preserved nested property schema details during that `anyOf` conversion so complex object/union shapes continue to validate correctly.
- Added schema-guided date coercion before validation in provider compat flows where raw model output can return strings for `z.date()` fields.
- Refined `x-date` propagation so strict `z.date()` fields can be converted before validation while `z.coerce.date()` fields remain provider-returned strings until schema validation handles coercion.
- Synced provider e2e tests so Anthropic, Google, and OpenAI-reasoning exercise the same structured-output and tool input/output cases as OpenAI.
- Updated snapshots and recordings to match the current compat behavior.

### `packages/core`

- Updated tool-builder schema validation to work correctly with compat-processed standard schemas.
- Added exact-match replay controls to the recorder/mock path for replay-sensitive tool-call tests.
- Refreshed affected core recordings/tests to align with the current validation behavior.

## Why

The branch started as provider-compat test cleanup, but the test coverage surfaced a few real compatibility gaps:

- some OpenAI-compatible paths were still emitting nullable optional fields as type arrays,
- some providers were validating raw string dates directly against `z.date()`,
- and replay-sensitive tool-call tests in core needed stricter matching to avoid stale recordings masking current schema behavior.

This PR fixes those issues together so the compat layer behavior and the recorded integration tests stay aligned.

### `packages/schema-compat`

```bash
cd packages/schema-compat
pnpm exec vitest run src/provider-compats/openai.test.ts src/provider-compats/groq.test.ts src/provider-compats/openai-reasoning.test.ts -u
pnpm exec vitest run src/provider-compats/google.test.ts
[...]

PR #15169, opened 2026-04-08
fix(core): use .optional() instead of .nullish() in sub-agent tool schemas

octo-patch:
Fixes #15076

## Problem

Agent network sub-agent tool declarations are rejected by Google Gemini's function-calling API with `property is not defined`. Root cause: `agentInputSchema` uses `.nullish()` which generates `anyOf:[{type:"string"},{type:"null"}]` JSON Schema that Gemini rejects.

## Solution

Changed `.nullish()` to `.optional()` for `threadId`, `resourceId`, `instructions`, `maxSteps`. OpenAI null values are already handled by `validateToolInput` Step 5 (GitHub #12362).

## Testing

Updated `tools.test.ts` schema copy to match. Existing tests pass.

## ELI5 (Explain Like I'm 5)

This PR fixes a compatibility problem where Google Gemini can't handle the way optional fields were being described in sub-agent tool schemas. The fix changes how optional fields are marked (using `.optional()` instead of `.nullish()`), which makes the schema descriptions simpler and compatible with Gemini's requirements, while still working properly with other AI providers like OpenAI.

## Summary

This PR resolves issue #15076 by updating sub-agent tool input schemas to use `.optional()` instead of `.nullish()` for optional fields, ensuring compatibility with Google Gemini's function-calling API.

### Problem
Agent network sub-agent tool declarations were being rejected by Google Gemini's function-calling API with errors indicating that properties were not defined. The root cause was that the `agentInputSchema` used Zod's `.nullish()` method, which generates JSON Schema containing `anyOf` constructs (e.g., `anyOf: [{type: "string"}, {type: "null"}]`). Gemini's function-calling API does not support `anyOf` in parameter schema definitions.

### Solution
Replaced `.nullish()` with `.optional()` for optional input fields (`threadId`, `resourceId`, `instructions`, `maxSteps`) while keeping `prompt` as a required field. This change eliminates the `anyOf` construct from the generated JSON Schema, producing simpler schema definitions that Gemini can process.

The solution maintains compatibility with OpenAI by relying on existing runtime validation and coercion through the `validateToolInput` step (referenced in PR #12362), which already handles transformation of null values to undefined at execution time. The execute handler implements falsy checks that safely coerce null to undefined for non-required fields, so OpenAI sending null for unfilled fields remains safe.

### Changes
- **packages/core/src/agent/agent.ts**: Updated the dynamically constructed sub-agent tool input Zod schema to use `.optional()` instead of `.nullish()` for `threadId`, `resourceId`, `instructions`, and `maxSteps` fields. Added clarifying comments explaining that `.optional()` avoids `anyOf` constructs that break Gemini function calling, and that the execute handler already handles null coercion via falsy checks for OpenAI compatibility.
- **packages/core/src/agent/__tests__/tools.test.ts**: Updated the test-local `agentInputSchema` in the "sub-agent prompt input normalization (GitHub #14154)" test to match the schema changes, changing the same four fields from `.nullish()` to `.optional()`.
- **.changeset/fix-agent-schema-optional-gemini.md**: Added a changeset entry documenting the patch-level update to `@mastra/core`.

### Impact
- Fixes Gemini function-calling compatibility without breaking existing OpenAI integration
- Simplifies generated JSON schemas by removing `anyOf` constructs that some providers cannot process
- Existing validation and coercion logic continue to handle null/undefined values appropriately at runtime
- Tests continue to pass with the updated schema definitions

### DIFF
--- a/.changeset/fix-agent-schema-optional-gemini.md
+++ b/.changeset/fix-agent-schema-optional-gemini.md
@@ -1,0 +1,5 @@
+---
+'@mastra/core': patch
+---
+
+Fixed sub-agent tool input schema to use `.optional()` instead of `.nullish()` for optional fields (`threadId`, `resourceId`, `instructions`, `maxSteps`). The previous `.nullish()` produced `anyOf: [T, {type: "null"}]` in JSON Schema, which Google Gemini's function calling API rejects. Using `.optional()` generates clean JSON Schema without `anyOf`, while the execute handler already handles null values via falsy checks so OpenAI compatibility is preserved.
--- a/packages/core/src/agent/__tests__/tools.test.ts
+++ b/packages/core/src/agent/__tests__/tools.test.ts
@@ -1395,10 +1395,10 @@ describe('sub-agent prompt input normalization (GitHub #14154)', () => {
   // by CoreToolBuilder.createExecute) must normalize it before validation fails.
   const agentInputSchema = z.object({
     prompt: z.string().describe('The prompt to send to the agent'),
-    threadId: z.string().nullish().describe('Thread ID'),
-    resourceId: z.string().nullish().describe('Resource ID'),
-    instructions: z.string().nullish().describe('Additional instructions'),
-    maxSteps: z.number().min(3).nullish().describe('Max steps'),
+    threadId: z.string().optional().describe('Thread ID'),
+    resourceId: z.string().optional().describe('Resource ID'),
+    instructions: z.string().optional().describe('Additional instructions'),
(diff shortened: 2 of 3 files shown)


wardpeet:
Thank you for creating this pull request, I believe https://github.com/mastra-ai/mastra/pull/14624 is fixing this issue as well. Can you double-check?

intojhanurag:
> Thank you for creating this pull request, I believe #14624 is fixing this issue as well. Can you double-check?

Hey @wardpeet it is bot , we can close

PR #15173, opened 2026-04-08
Add MastraCode product analytics

smthomas:
## Summary
- add a dedicated PostHog analytics wrapper for MastraCode with  opt-out support
- track session start, prompt submission, thread/model changes, selected slash commands, and key interactive prompts in the TUI
- add targeted analytics and command/thread prompt tests and update the lockfile/changeset

## Test plan
- pnpm --filter ./mastracode test src/__tests__/analytics.test.ts src/tui/__tests__/command-dispatch.test.ts src/tui/commands/__tests__/threads.test.ts
- pnpm --filter ./mastracode check

## ELI5

This PR adds silent, opt-out product analytics to MastraCode so the app can record how people use it (sessions, commands, prompts, thread/model changes) while letting users disable telemetry with an environment variable.

## Overview

Adds a PostHog-backed analytics wrapper for MastraCode with a noop fallback when telemetry is disabled. Analytics are resilient to errors (swallowed) and threaded through the TUI so session, prompt, command, thread, model, and selected interactive-prompt events are captured. Includes targeted tests and a changeset.

## Key Changes

- New analytics core (mastracode/src/analytics.ts)
  - Exports MastraCodeAnalytics interface and MastraCodeAnalyticsEvent union.
  - PostHogMastraCodeAnalytics implementation using posthog-node with per-instance sessionId, hostname-based distinctId, and rich base properties.
  - NoopMastraCodeAnalytics for disabled telemetry.
  - isTelemetryDisabled(env?) helper and createMastraCodeAnalytics(options) factory.
  - All capture/shutdown calls wrapped in try/catch to avoid affecting runtime.

- Initialization and wiring
  - main.ts initializes analytics via createMastraCodeAnalytics(...) and captures session start.
  - TUI state and options accept analytics; analytics instance is threaded into command and event contexts.

- Event tracking added throughout TUI
  - Prompt submission: captures payload including thread/resource/mode, hasImages, pendingNewThread, isFirstPromptInThread.
  - Harness events: thread_created/thread_changed/model_changed → mastracode_thread_changed / mastracode_model_changed.
  - Slash commands: dispatchSlashCommand tracks selected commands (e.g., /models, /custom-providers, /new) with thread/resource/mode.
  - Interactive prompts: showThreadLockPrompt, login provider selector, and tool-related events (tool_approval_required, tool_input_start for ask_user/request_access/submit_plan) emit trackInteractivePrompt events.

- Type updates
  - Added optional analytics?: MastraCodeAnalytics to SlashCommandContext, EventHandlerContext, MastraTUIOptions, and TUIState so handlers can access analytics safely.

## Testing

- Added mastracode/src/__tests__/analytics.test.ts to verify telemetry-disable behavior and that the noop client is safe and shutdown resolves.
- Updated tests:
  - mastracode/src/tui/__tests__/command-dispatch.test.ts to assert trackCommand calls for routed slash commands.
  - mastracode/src/tui/commands/__tests__/threads.test.ts to assert trackInteractivePrompt for thread lock prompt.

## Dependencies

- Added dependency: posthog-node@^5.28.3
- package.json entries adjusted (minor reordering; yaml re-added)

## Changeset

- .changeset/few-bobcats-notice.md added noting a minor release and that MASTRA_TELEMETRY_DISABLED=1 disables telemetry.

## Lines changed (approx.)
- Core analytics: ~125 added
- TUI integration & handlers: ~100+ modified across multiple files
- Tests & other: ~60 added
- Total: several hundred lines across analytics module, TUI wiring, and tests


TylerBarnes (changes requested):
From Mastra Code: Thanks for putting this together — the analytics wrapper is nicely isolated, and threading it through the existing TUI/state/context paths keeps the call sites testable and optional.

A few things I’d consider before merge:

1. **Telemetry identity/privacy wording**
   The docs describe the analytics as anonymous, but events include a hostname-derived distinct ID and also send the raw hostname as `machineId` from `getBaseProperties()`. Hostnames can include user, company, or device identifiers, so I’d either hash/remove the hostname-derived fields or adjust the docs to avoid calling this anonymous.

2. **Lazy analytics initialization**
   `createMastraCodeAnalytics()` currently runs at module scope in `main.ts`, before `main()` decides whether to route to `--help` or headless mode. Even if no events are captured there, this still constructs the PostHog client unnecessarily. It seems safer to create the analytics client inside `tuiMain()` once we know the interactive TUI is actually starting.

3. **CI status**
   `statusCheckRollup` currently shows several failing checks, including Build/Lint/peer dependency/docs-related jobs. Some may be unrelated/flaky, but I’d want those verified before merge.

Minor note: command telemetry currently fires before the handler succeeds, so those events should be interpreted as “command selected/attempted” rather than “command completed.” That may be exactly what we want, but worth being explicit about in analysis.

TylerBarnes:
Thanks for the updates — lazy initialization and the explicit `action: "attempted"` command telemetry semantics look good.

I think the hostname-based identity still needs to change before merge.

Right now `os.hostname()` is used both for the PostHog `distinctId` and as `machineId` in the event payload. On my machine it returns `Mac.lan`, which highlights the issue: it’s not personally identifying in this case, but it’s also not unique. Other users may have hostnames like `tylers-mbp`, `jane-doe-laptop`, or `CORP-ASSET-1234`, which may be identifying.

So hostname-based telemetry identity has two problems:

1. **Privacy:** hostnames are environment-dependent and can contain usernames, company asset tags, or other real-world identifiers.
2. **Correctness:** generic hostnames like `Mac.lan`, `MacBook-Pro.local`, `ubuntu`, or `localhost` can collide across many users, causing unrelated users to be merged under the same PostHog identity.

In other words, this is neither reliably anonymous nor reliably unique.

Could we replace the hostname-derived ID with a generated anonymous install ID instead? For example, generate a UUID once, store it in the MastraCode app/config data directory, and use that persisted UUID as the PostHog `distinctId`. I’d also remove the raw hostname from event properties entirely rather than sending it as `machineId`.

Once that’s changed, the docs can accurately describe the telemetry as anonymous product analytics.

TylerBarnes (approved):
discussed on slack, we'll make IDs unique (for mc and framework) as a follow up

PR #15195, opened 2026-04-09
fix(core): use .optional() instead of .nullish() in sub-agent tool schemas

YgorLeal:
## Summary
Fixes #15076 — Sub-agent tool schemas in agent networks now use `.optional()` instead of `.nullish()`, restoring compatibility with Google Gemini's function calling API.

## Problem
The `agentInputSchema` used `.nullish()` (= `.nullable().optional()`) for optional fields like `threadId`, `resourceId`, `instructions`, and `maxSteps`. This produces `anyOf` unions in JSON Schema:

```json
{ "threadId": { "anyOf": [{ "type": "string" }, { "type": "null" }] } }
```

Google Gemini does **not** support `anyOf` in tool parameter schemas, causing all sub-agent tools to be rejected with:
```
GenerateContentRequest.tools[0].function_declarations[N].parameters.required[0]: property is not defined
```

## Solution
Replaced `.nullish()` with `.optional()` in `agentInputSchema` (and matching test). Since these are **input** schemas (what the parent agent sends to the sub-agent), omitting the field (`undefined`) is sufficient — explicit `null` values are unnecessary.

### Files changed
- `packages/core/src/agent/agent.ts` — production schema
- `packages/core/src/agent/__tests__/tools.test.ts` — matching test schema

## Test plan
- [x] Existing tests in `tools.test.ts` updated to match
- [ ] Verify sub-agent delegation works with Google Gemini models
- [ ] Verify sub-agent delegation still works with OpenAI models (regression check)

## ELI5

Some auto-generated tool schemas were rejected because they explicitly allowed "null". This PR makes those fields optional (they can be left out), so the generated schema no longer includes null unions and is accepted by Google Gemini.

---

## Changes Overview

Replaces `.nullish()` (equivalent to `.nullable().optional()` which produced JSON Schema anyOf unions like `{ "anyOf": [{ "type": "string" }, { "type": "null" }] }`) with `.optional()` in the agent input schema used to generate sub-agent tool parameter schemas. This removes `anyOf`/explicit `null` from the generated JSON Schema so Google Gemini's function-calling API accepts the tools.

Fields changed: threadId, resourceId, instructions, maxSteps — still optional, but no longer explicitly accept null.

Rationale: For parent→sub-agent inputs, omitting a field (undefined) is sufficient; explicit null support is unnecessary and broke Gemini. If null support is required for other providers (e.g., OpenAI), consider provider-specific schema transformation converting `anyOf: [T, {type: "null"}]` into a provider-supported nullable representation.

---

## Modified Files

- packages/core/src/agent/agent.ts (+4/-5)
  - Switched Zod schema fields from `.nullish()` to `.optional()` for: threadId, resourceId, instructions, maxSteps.
  - Required fields (e.g., prompt) and overall schema structure unchanged.

- packages/core/src/agent/__tests__/tools.test.ts (+16/-4)
  - Updated test-local Zod schema to match the change and added a regression test that converts the schema to JSON Schema and asserts no `anyOf` containing `{ type: 'null' }` is emitted for the optional fields.

- .changeset/optional-nullish-gemini-compat.md (+5/-0)
  - Added changeset documenting the fix for Gemini compatibility.

---

## Testing & Verification

- Updated tests included (regression test ensuring no `anyOf` with `null` in generated JSON Schema).
- Remaining verification: confirm sub-agent delegation works with Google Gemini models and run regression checks against OpenAI models.


wardpeet:
Thank you for creating this pull request, i believe https://github.com/mastra-ai/mastra/pull/14624 is fixing this issue as well. Can you double check?

YgorLeal:
> Thank you for creating this pull request, i believe #14624 is fixing this issue as well. Can you double check?

Thanks for the pointer! I checked #14624 and confirmed it already fixes #15076 at the schema-compat layer in a much more comprehensive way. Closing this PR as a duplicate.

PR #15137, opened 2026-04-07
Add RAG observability: tracing for ingestion and query operations

epinzur:
## Description

Surfaces RAG ingestion and query operations in Mastra's AI observability system, enabling users to debug retrieval quality and ingestion pipelines end-to-end through distributed tracing.

### New Span Types

Added 5 new span types to `@mastra/core/observability`:

- **`RAG_INGESTION`** (root) — wraps a complete ingestion pipeline run (load → chunk → extract → embed → upsert)
- **`RAG_EMBEDDING`** — embedding calls used by both ingestion and query, with `mode: 'ingest' | 'query'`
- **`RAG_VECTOR_OPERATION`** — vector store I/O with `operation: 'query' | 'upsert' | 'delete' | 'fetch'`
- **`RAG_ACTION`** — RAG-specific operations (`chunk`, `extract_metadata`, `rerank`) distinguished by `action` attribute
- **`GRAPH_ACTION`** — non-RAG graph operations (`build`, `traverse`, `update`, `prune`) for knowledge graphs, memory graphs, etc.

Each type includes a corresponding attributes interface with relevant metadata (model, provider, dimensions, counts, etc.).

### New Observability Helpers

Exported from `@mastra/core/observability`:

- **`startRagIngestion(opts)`** — manual control: returns `{ span, observabilityContext }` for users who want to manage span lifecycle
- **`withRagIngestion(opts, fn)`** — scoped helper: runs `fn(observabilityContext)`, automatically attaches return value as span output, routes errors to `span.error(...)`

Both accept optional `mastra` instance and attributes; no-op gracefully when observability is disabled.

### Instrumentation in `@mastra/rag`

- **`vectorQuerySearch`** — emits `RAG_EMBEDDING` (mode: `query`) and `RAG_VECTOR_OPERATION` (operation: `query`) child spans
- **`rerank` / `rerankWithScorer`** — emits `RAG_ACTION` (action: `rerank`) with scorer metadata
- **`MDocument.chunk`** — emits `RAG_ACTION` (action: `chunk`) with strategy/size/overlap; nested `RAG_ACTION` (action: `extract_metadata`) for metadata extraction
- **`createVectorQueryTool`** — threads `observabilityContext` from agent's `TOOL_CALL` span automatically
- **`createGraphRAGTool`** — emits `GRAPH_ACTION` (action: `build`) for graph initialization and `GRAPH_ACTION` (action: `traverse`) for query walks

[...]

PR #15199, opened 2026-04-09
feat(core): trace embed and upsert steps in RAG ingestion pipeline

### LINKED ISSUE #10898: [FEATURE] Add RAG tracing
### Problem Statement

When I run an agent, and it uses RAG, it would be helpful to see info about it in my trace.

### Proposed Solution

Add RAG tracing.

### Component

Other

### Alternatives Considered

_No response_

### Example Use Case

RAG is confusing. It would be great to gain some insight into how it is working in order to debug my agent.

### Additional Context

_No response_

### Verification

intojhanurag:
## Description

Follow-up to #15137 — adds the missing observability tracing for the **embed** and **upsert** steps during RAG
ingestion. Previously only chunking and the query path were instrumented.

- `embedForIngestion()` — wraps `embedV1`/`embedV2`/`embedV3` with a `RAG_EMBEDDING` span (`mode: 'ingest'`),
accumulates token usage across the batch
- `upsertWithTracing()` on `MastraVector` — wraps `upsert()` with a `RAG_VECTOR_OPERATION` span (`operation: 'upsert'`)
- Fixes `withRagIngestion` / `startRagIngestion` JSDoc examples to reference the actual APIs

Both are opt-in via `observabilityContext`. No breaking changes — existing callers are unaffected.

## Related Issue(s)

Follow-up to #15137 (Fixes #10898)

## Type of Change

- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [ ] Test update

## Checklist

## ELI5

This PR adds the ability to track and observe what's happening when RAG systems embed text and store vectors in a database. Think of it like adding checkpoints in a factory assembly line so you can see how long each step takes and spot problems—it helps developers understand and debug how their RAG-powered agents are performing.

## Overview

This PR adds observability tracing for the embed and upsert steps of the RAG ingestion pipeline, following up on PR #15137. It introduces two new instrumented operations:

1. **`embedForIngestion()`** - Wraps embedding calls with RAG observability spans
2. **`upsertWithTracing()`** - Wraps vector upsert calls with RAG observability spans

Tracing is completely optional and controlled via `observabilityContext`, so existing code is unaffected.

### Embedding Instrumentation
- New `embedForIngestion()` function in `embed-instrumented.ts` that:
  - Creates a `RAG_EMBEDDING` span with `mode: 'ingest'` when observability is enabled
  - Routes to the appropriate embed function (`embedV1`, `embedV2`, or `embedV3`) based on model specification version
  - Accumulates token usage across the batch and includes `inputTokens` in span metadata
  - Records embedding dimensions and vector count on success
  - Properly handles errors with span recording

### Vector Upsert Instrumentation
- New `upsertWithTracing()` method on `MastraVector` class that:
  - Wraps the existing `upsert()` call with a `RAG_VECTOR_OPERATION` span (`operation: 'upsert'`)
  - Includes index name, vector count, and dimensions in span attributes
  - Records output vector count on success
  - Passes through errors while recording them on the span

### Documentation Updates
- Updated JSDoc examples in `rag-ingestion.ts` to:
  - Correct the `doc.chunk()` API signature to use `doc.chunk({}, { observabilityContext })`
  - Replace raw `embed()` calls with `embedForIngestion()`
  - Replace raw `vectorStore.upsert()` calls with `vectorStore.upsertWithTracing()`

### Public API Exports
- Both new functions are exported from the main vector module index for public use

## Testing

Comprehensive test coverage was added:
- `embed-instrumented.test.ts`: Tests routing logic, span creation, token accumulation, and error handling for `embedForIngestion()`
- `upsert-with-tracing.test.ts`: Tests conditional span creation, attribute recording, and error propagation for `upsertWithTracing()`

## Changeset

Added a minor version bump entry for `@mastra/core` documenting the new opt-in observability features.

### DIFF
--- a/packages/core/src/vector/embed-instrumented.ts
+++ b/packages/core/src/vector/embed-instrumented.ts
@@ -1,0 +1,114 @@
+/**
+ * Instrumented embedding helpers for RAG ingestion.
+ *
+ * Wraps the raw AI SDK `embed` functions with `RAG_EMBEDDING` child spans
+ * carrying `mode: 'ingest'`. This is the ingestion counterpart of the
+ * query-path embedding instrumented inside `vectorQuerySearch`.
+ *
+ * Accepts an optional `observabilityContext`; no-ops when absent.
+ */
+
+import type { ObservabilityContext } from '../observability';
+import { SpanType } from '../observability';
+import { embedV1, embedV2, embedV3 } from './embed';
+import type { MastraEmbeddingModel, MastraEmbeddingOptions } from './vector';
+
+export interface EmbedForIngestionParams {
+  /** The embedding model to use */
+  model: MastraEmbeddingModel<string>;
+  /** Array of text values to embed */
+  values: string[];
+  /** Optional embedding options (maxRetries, providerOptions, etc.) */
+  options?: MastraEmbeddingOptions;
+  /** Observability context for tracing. When absent, no spans are emitted. */
+  observabilityContext?: ObservabilityContext;
+}
+
+export interface EmbedForIngestionResult {
+  /** The computed embedding vectors, one per input value */
+  embeddings: number[][];
+}
+
+/**
+ * Embed an array of texts for ingestion, optionally emitting a
+ * `RAG_EMBEDDING` span with `mode: 'ingest'`.
+ *
+ * This mirrors the embedding logic in `vectorQuerySearch` but handles
+ * batch inputs and tags spans as ingestion rather than query.
+ */
+export async function embedForIngestion({
+  model,
+  values,
+  options,
+  observabilityContext,
+}: EmbedForIngestionParams): Promise<EmbedForIngestionResult> {
+  const parentSpan = observabilityContext?.tracingContext?.currentSpan;
+
+  const embedSpan = parentSpan?.createChildSpan({
+    type: SpanType.RAG_EMBEDDING,
+    name: 'rag embed: ingest',
+    input: { count: values.length },
+    attributes: {
+      mode: 'ingest',
+      model: (model as any)?.modelId,
+      provider: (model as any)?.provider,
+      inputCount: values.length,
+    },
+  });
+
+  try {
+    const embeddings: number[][] = [];
+    let totalTokens = 0;
+
+    for (const value of values) {
+      let result;
+      if (model.specificationVersion === 'v3') {
+        result = await embedV3({
+          model: model,
+          value,
+          maxRetries: options?.maxRetries,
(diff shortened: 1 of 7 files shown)


packages/core/src/vector/embed-instrumented.ts
```diff
@@ -0,0 +1,114 @@
+/**
+ * Instrumented embedding helpers for RAG ingestion.
+ *
+ * Wraps the raw AI SDK `embed` functions with `RAG_EMBEDDING` child spans
+ * carrying `mode: 'ingest'`. This is the ingestion counterpart of the
+ * query-path embedding instrumented inside `vectorQuerySearch`.
+ *
+ * Accepts an optional `observabilityContext`; no-ops when absent.
+ */
```
epinzur, line 9: here we should probably just wrap all the AI-SDK methods... embed, embedMany, etc..

I didn't realize that the RAG stuff still relied so heavily on non-mastra libraries.  This makes tracing harder.


As best as possible the user code to get this to work should remain looking the same as the old way... with maybe a few minor changes:

like: 

```
import { withRagIngestion } from '@mastra/core/observability';
import { embed } from 'ai'
import { PgVector } from '@mastra/pg'

...

await withRagIngestion(
  {
    mastra,
    name: 'docs ingestion',
    attributes: { vectorStore: 'pgvector', indexName: 'docs' },
  },
  async (observabilityContext) => {
    const chunks = await doc.chunk({ observabilityContext });
    const { embeddings } = await embed(chunks, { observabilityContext });
    await vectorStore.upsert({
      indexName: 'docs',
      vectors: embeddings,
      observabilityContext,
    });
    return { chunkCount: chunks.length };
  },
);
```

but change:

```
import { embed } from 'ai'
```

to 

```
import { embed } from '@mastra/rag'
```

does that make sense?

epinzur: I want to avoid creating lot of `withSomething` helpers... tracing should just work transparently with minimal changes.

intojhanurag: Thanks Eric! I actually thought about this approach initially. I skipped the embed wrap thinking it would bloat @mastra/rag, and skipped the in-place upsert tracing because it meant touching the abstract method and all 16 store implementations.

intojhanurag: I can go ahead with your approach if you agree with this refactor

intojhanurag:
closing this , I don't think it is good approach

PR #15217, opened 2026-04-09
feat(azure): add Azure Blob Storage workspace filesystem provider

intojhanurag:
## Description

Adds `@mastra/azure`, a new Azure Blob Storage workspace filesystem provider, so teams on Azure can use native Blob containers for workspace persistence without S3/GCS workarounds.

- Implements the full `WorkspaceFilesystem` interface on top of `@azure/storage-blob`
- Auth: connection string, account key, SAS token, `DefaultAzureCredential` (optional `@azure/identity`), anonymous
- Prefix namespacing, read-only mode, MastraEditor provider descriptor
- 56 unit tests + 90 integration tests (verified against real Azure, all passing)
- Docker Compose setup for Azurite emulator
- Adds `'azure-blob'` to the mount-type list in `@internal/workspace-test-utils`

Follows the same architecture as `@mastra/gcs` and `@mastra/s3`.

## Related Issue(s)

Fixes #14104

## Type of Change

- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [ ] Test update

## Checklist

ELI5: This adds a new plugin so the app can store and manage files directly in Microsoft Azure Blob Storage — like giving the app a new cloud drawer. It supports multiple login methods, optional subfolder-prefixing, and a read-only mode, plus lots of unit and integration tests and a local Azurite emulator for safe testing.

* **New Features**
  * New package @mastra/azure implementing an Azure Blob Storage WorkspaceFilesystem.
  * Supports authentication via connection string, accountName/accountKey (StorageSharedKeyCredential), SAS token, optional DefaultAzureCredential (dynamically imported from @azure/identity), and anonymous access. Requires a container.
  * Supports optional prefix namespacing and a readOnly mode (mutations assert writable).
  * Exposes provider descriptor azureBlobFilesystemProvider and public exports: AzureBlobFilesystem, AzureBlobFilesystemOptions, AzureBlobMountConfig.
  * Implements core filesystem methods: init/destroy, readFile, writeFile, appendFile, deleteFile, copyFile, moveFile, readdir, exists, stat, mkdir, rmdir, isFile, isDirectory, getMountConfig, getInfo, getInstructions.

* **Implementation details**
  * Maps filesystem paths to blob keys with normalized prefix handling.
  * Lazily constructs and caches Azure SDK clients (ContainerClient/BlobServiceClient); dynamic import for DefaultAzureCredential when requested.
  * readFile decodes by encoding when requested and maps Azure 404 → FileNotFoundError, 403 → access handling.
  * writeFile enforces readOnly, supports overwrite semantics and sets Content-Type from file extension.
  * appendFile implemented as read-modify-write with binary-safe concatenation.
  * copyFile/moveFile attempt server-side copy (sync vs async thresholds, SAS URL generation) with fallback to download+reupload and special handling for zero-length blobs.
  * Directory semantics via blob prefix listing; mkdir is a no-op; rmdir supports non-recursive rejection and recursive batched deletion.
  * readdir supports hierarchical listing (listBlobsByHierarchy) and flat listing (listBlobsFlat) and converts results to file/directory entries.

* **Tests**
  * Adds unit tests (mocking @azure/storage-blob) and credential-gated integration tests that run against real Azure or the Azurite emulator.
  * Integration test suite includes core operation flows, prefix isolation tests, and wiring into higher-level workspace integration/conformance suites.
  * Test infra: Vitest config, package.json pretest/posttest hooks that bring up/tear down Azurite via Docker Compose (workspaces/azure/docker-compose.yml).

* **Packaging & tooling**
  * New workspace package at workspaces/azure with ESM/CJS exports, tsup build config, TypeScript configs, Vitest config, and ESLint config.
  * Declares runtime dependency on @azure/storage-blob and optional peer @azure/identity; peer dependency on @mastra/core.
  * Adds 'azure-blob' mount type to internal test utils allowlist.
  * Changeset added to publish the new package.

* **Notes**
  * Integration tests are gated by environment variables (AZURE_STORAGE_CONNECTION_STRING, TEST_AZURE_CONTAINER) and package.json test scripts use the Azurite default connection string for local runs.
  * PR marked "Not ready for review yet" by the author; docs and some CodeRabbit comments remain to be addressed.


workspaces/azure/src/blob/filesystem/index.ts
```diff
@@ -0,0 +1,646 @@
...
+ * Azure Blob mount configuration.
+ * Returned by AzureBlobFilesystem.getMountConfig() for FUSE mounting in sandboxes.
+ */
+export interface AzureBlobMountConfig extends FilesystemMountConfig {
+  type: 'azure-blob';
+  container: string;
+  accountName?: string;
+  accountKey?: string;
+  connectionString?: string;
+}
```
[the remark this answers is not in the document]
intojhanurag: Addressed 4 of the 5 fields in the follow-up commits:

- Added `sasToken`, `endpoint`, `readOnly`, and `useDefaultCredential` to `AzureBlobMountConfig` and `getMountConfig()`.
- Deliberately did **not** add `prefix`.

**Reason for excluding `prefix`:** Neither `S3MountConfig` nor `GCSMountConfig` includes `prefix` — in both existing providers it's handled by the logical filesystem layer, not the mount layer. `getMountConfig()` is specifically documented for "FUSE mounting in sandboxes" where the whole container/bucket is mounted and prefix filtering happens in the Mastra abstraction on top.

Adding `prefix` only to Azure would deviate from the established pattern. If `prefix` should be part of mount config across all providers, I think it's better addressed in a separate cross-provider PR. Happy to include it here if maintainers prefer — let me know.

workspaces/azure/src/blob/filesystem/index.ts
```diff
@@ -0,0 +1,646 @@
...
+      throw new FileExistsError(path);
+    }
+
+    const body = typeof content === 'string' ? Buffer.from(content, 'utf-8') : Buffer.from(content);
+    const contentType = getMimeType(path);
+    const blobClient = containerClient.getBlockBlobClient(this.toKey(path));
+
+    await blobClient.upload(body, body.length, {
+      blobHTTPHeaders: { blobContentType: contentType },
+    });
```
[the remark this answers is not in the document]
intojhanurag: Verified this against `@mastra/s3` and `@mastra/gcs`: **neither existing provider implements `expectedMtime`**. The field is defined in `WriteOptions` in `@mastra/core/workspace` but no provider currently enforces it.

Implementing it only in Azure would create inconsistency with the existing providers. I think this is better addressed in a separate PR that adds optimistic-concurrency support uniformly across S3, GCS, and Azure — or the field should be removed from `WriteOptions` if it's not intended to be a required contract.

Leaving out of scope for this PR, but flagging for maintainers to decide how they want to handle this at the interface level.

workspaces/azure/src/blob/filesystem/index.integration.test.ts
```diff
@@ -0,0 +1,288 @@
...
+    capabilities: {
+      supportsAppend: true,
+      supportsBinaryFiles: true,
+      supportsMounting: false,
+      supportsForceDelete: true,
+      supportsOverwrite: true,
+      supportsConcurrency: true,
+      supportsEmptyDirectories: false,
+      deleteThrowsOnMissing: true,
+    },
```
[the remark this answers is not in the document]
intojhanurag: You're right — the conformance test at `mount-config.ts:54` explicitly asserts `expect(fs.getMountConfig).toBeUndefined()` when `supportsMounting: false`. Since `AzureBlobFilesystem` implements `getMountConfig()`, this would fail.

Changed back to `supportsMounting: true`. The conformance suite's mount tests verify config shape, not actual sandbox mounting capability — which is consistent with how S3 and GCS use this flag.

workspaces/azure/src/blob/filesystem/index.ts
```diff
@@ -0,0 +1,726 @@
...
+      if (error instanceof FileNotFoundError) {
+        // File doesn't exist, start fresh with empty buffer
+      } else {
+        throw error;
+      }
+    }
+
+    const appendBuffer = typeof content === 'string' ? Buffer.from(content, 'utf-8') : Buffer.from(content);
+    await this.writeFile(path, Buffer.concat([existing, appendBuffer]));
+  }
```
NikAiyer: I notice we already create a buffer in write file. I wonder if there's a way we can dedupe this logic so we only create a buffer once.

workspaces/azure/package.json
```diff
@@ -0,0 +1,86 @@
...
+    },
+    "./package.json": "./package.json"
+  },
+  "scripts": {
+    "build": "tsup --silent --config tsup.config.ts",
+    "build:lib": "pnpm build",
+    "build:watch": "pnpm build --watch",
+    "test:unit": "vitest run --exclude '**/*.integration.test.ts'",
+    "test:watch": "vitest watch",
+    "pretest": "docker compose up -d && sleep 3",
```
NikAiyer: Since you have a healthcheck in the docker-compose file, should we just use this for pretest?

```suggestion
    "pretest": "docker compose up -d --wait",
```

workspaces/azure/src/blob/filesystem/index.ts
```diff
@@ -0,0 +1,726 @@
...
+    await this.copyFile(src, dest, options);
+    await this.deleteFile(src, { force: true });
+  }
+
+  // ---------------------------------------------------------------------------
+  // Directory Operations
+  // ---------------------------------------------------------------------------
+
+  async mkdir(path: string, _options?: { recursive?: boolean }): Promise<void> {
+    this.assertWritable(path, 'mkdir');
```
NikAiyer: Do we need this guard? This function is a no-op.

workspaces/azure/src/blob/filesystem/index.ts
```diff
@@ -0,0 +1,726 @@
...
+    }
+
+    if (blobNames.length > 0) {
+      await this.deleteBlobBatch(containerClient, blobNames);
+    }
+  }
+
+  private async deleteBlobBatch(containerClient: ContainerClient, blobNames: string[]): Promise<void> {
+    await Promise.all(blobNames.map(name => containerClient.getBlobClient(name).delete()));
+  }
```
NikAiyer, line 519: Is there an API for batch deletes already in Azure? If there is, we might want to use that, unless there are issues with it in testing.

NikAiyer (approved):

Issue #14932, opened 2026-04-01
Supervisor nested agent streaming emits huge cumulative data-tool-agent payloads and OOMs


dawgster:
## Summary

Using the supervisor pattern with a delegated sub-agent appears to generate extremely large `data-tool-agent` stream payloads and can drive Node into OOM even within a single delegation run.

This does **not** look like normal tool-result size alone. In our repro, the nested agent stream emits repeated cumulative `data-tool-agent` snapshots whose size grows rapidly as steps accumulate.

## Versions

- `mastra`: `1.3.15`
- `@mastra/core`: `1.20.0`
- `@mastra/ai-sdk`: `1.3.0`
- `@mastra/memory`: `1.12.1`
- `@mastra/deployer`: `1.17.0`
- Node: `v24.14.0`
- OS: `Linux pop-os 6.17.4-76061704-generic x86_64`

## What we observed

We have a top-level routing/supervisor agent delegating to a  sub-agent. During the delegated run, the streamed response becomes enormous.

I reproduced this locally against the live dev server and captured:

- heap rising from ~`190 MB` to ~`2.43 GB` used heap
- RSS rising to ~`2.8 GB`
- eventual process death / server no longer accepting connections
- a heap snapshot captured during the inflating phase

The streamed response files for 3 concurrent requests were:

- request 1: ~`443 MB`
- request 2: ~`769 MB`
- request 3: ~`337 MB`

Those files were dominated by repeated `data-tool-agent` events:

- file 1: `109`
- file 2: `134`
- file 3: `106`

In one request, `data-tool-agent` payload sizes grew like this:

- event `#40`: ~`16 KB`, `1` step
- event `#80`: ~`562 KB`, `6` steps
- event `#120`: ~`18 MB`, `11` steps
- event `#133`: ~`72 MB`, `13` steps, `14` tool results

That looks like cumulative state being re-emitted on every update, not normal answer streaming.

## Suspected cause

The issue seems to be in the AI SDK transformer for nested agent output.

Relevant installed code:

- `node_modules/@mastra/ai-sdk/dist/index.js`
- `createAgentStreamToAISDKTransformer()` around `12254`
- `transformAgent()` around `12388`
- `data-tool-agent` emission around `12527`

The important behavior is:

1. `tool-output` from nested agents is converted to `tool-agent`
2. `transformAgent()` keeps cumulative buffered state per run (`text`, `toolCalls`, `toolResults`, `steps`, etc.)
[...]

PR #15218, opened 2026-04-09
fix(ai-sdk): compact workflow stream snapshots

graysonhicks:
## Description

This is a follow-up to `#14986` for the explicit workflow streaming path in `@mastra/ai-sdk`.

The prior fix removed cumulative growth from nested `data-tool-agent` payloads, but `workflowRoute()` / `handleWorkflowStream()` still emitted a full `data-workflow` snapshot on every workflow update. Once steps had large outputs, each later snapshot repeated all previously completed step outputs, which made total stream size and client-side retained data grow much faster than intended and could still lead to OOM in large real-world workflows.

This PR changes the workflow transformer so intermediate `data-workflow` parts remain lightweight and focus on workflow status plus step structure, while new `data-workflow-step` / `data-tool-workflow-step` parts carry the full payload for the step that just changed. The final `workflow-finish` snapshot still includes full step outputs, so end-of-run workflow state remains intact.

The PR also updates nested workflow handling inside network streams so the new workflow-step delta parts continue to propagate, exports the new `WorkflowStepDataPart` type, adds a regression test that proves intermediate workflow snapshots no longer repeat completed step outputs, and updates the AI SDK UI guide to document the new streaming shape.

Compatibility note: this only affects the `@mastra/ai-sdk` workflow stream transformer. Core workflow execution, stored workflow state, usage accounting, and observability data are unchanged. The main behavior change is that consumers who relied on intermediate `data-workflow.steps[*].output` during a live run must now also read `data-workflow-step` parts; final workflow snapshots still contain the full outputs.

Verification performed:
- `pnpm build:core`
- `pnpm --filter @mastra/ai-sdk exec tsc --noEmit`
- `pnpm --filter @mastra/ai-sdk run build:lib`
- `pnpm --filter @mastra/ai-sdk test -- src/__tests__/transform-agent-cumulative-growth.test.ts src/__tests__/transform-workflow-cumulative-growth.test.ts src/__tests__/transformers.test.ts`

The relevant transformer suites passed, including the new workflow regression test. A few unrelated `@mastra/ai-sdk` tests in this checkout still fail because of pre-existing package resolution issues (`@mastra/libsql`, `ai/test`) outside the files changed here.

## Related Issue(s)

Follow-up to #14932 and #14986

## Type of Change

- [x] Documentation update
- [ ] Code refactoring
- [x] Performance improvement
- [x] Test update

## Checklist

## ELI5

Instead of re-sending all previous step outputs every time a workflow updates (like reprinting all previous pages when you add a new one), this PR sends small workflow snapshots and only sends the full data for the single step that changed — preventing streams from growing without bound and avoiding OOMs.

## Changes Overview

This PR changes the @mastra/ai-sdk workflow stream transformer so intermediate workflow snapshots no longer duplicate completed step outputs; per-step full payloads are emitted as separate step-delta parts and final snapshots still include full outputs.

### Key Changes

- Streaming behavior
  - Intermediate data-workflow / data-tool-workflow snapshots now contain workflow status and step structure with step outputs nulled.
  - New data-workflow-step / data-tool-workflow-step parts carry the full payload for the single step that changed.
  - workflow-finish still emits a final snapshot including full step outputs.

- Types & exports
  - Added WorkflowStepDataPart and re-exported it from client-sdks/ai-sdk/src/index.ts.

- Transformer & network updates
  - Introduced buffered workflow state utilities and creators (BufferedWorkflowState, cloneWorkflowStep, serializeWorkflowSteps, createWorkflowDataPart, createWorkflowStepDataPart).
  - transformWorkflow now returns arrays in some cases and emits:
    - workflow-start / workflow-step-start → workflow-level part only.
    - workflow-step-result / workflow-step-suspended → [workflow-level part, step-level part].
    - workflow-finish → workflow-level part with includeOutputs: true.
  - createWorkflowStreamToAISDKTransformer and agent stream transformer now enqueue array results item-by-item.
  - transformNetwork updated so nested tool/network workflow streams propagate the new workflow-step delta parts.

- Tests
  - New regression test (transform-workflow-cumulative-growth.test.ts) verifies intermediate snapshots do not repeat completed step outputs and that exactly one data-workflow-step is emitted per step.
  - Added custom-data.test.ts asserting nested tool-workflow streams emit data-tool-workflow-step with the step's full output.

- Documentation
  - AI SDK UI guide updated: data-workflow redefined as status/structure snapshots; data-workflow-step documented and example UI code updated to handle step-delta parts.

### Compatibility & Verification

- Scope: Only the @mastra/ai-sdk workflow stream transformer and its conversion behavior changed. Core workflow execution, stored workflow state, usage accounting, and observability are unchanged.
- Consumer impact: Live-run consumers that previously read step outputs from intermediate data-workflow snapshots must also read data-workflow-step / data-tool-workflow-step parts for per-step payloads; final workflow snapshots are unchanged.
- Verification: pnpm build/typecheck/build/test for core and @mastra/ai-sdk (including the new regression test). A few unrelated @mastra/ai-sdk tests still fail locally due to pre-existing package resolution issues.


graysonhicks:
Marking as draft until confirmation that this upstream bug isn't the root cause: https://github.com/vercel/ai/issues/13839

client-sdks/ai-sdk/src/transformers.ts
```diff
@@ -322,23 +402,19 @@ export function createAgentStreamToAISDKTransformer<OUTPUT>(
...
+                controller.enqueue(workflowChunk);
               }
-            } else if (workflowChunk) {
-              controller.enqueue(workflowChunk);
             }
           } else if (transformedChunk.type === 'tool-network') {
             const payload = transformedChunk.payload;
             const networkChunk = transformNetwork(payload, bufferedSteps, true);
-            if (Array.isArray(networkChunk)) {
-              for (const c of networkChunk) {
```
TheIsrael1, line 336: This looks like a regression. transformNetwork() can still return an array in valid cases (for example the fallback
 text path, and now also the nested workflow propagation path), so enqueueing networkChunk directly here can emit a raw
 array chunk instead of individual AI SDK parts. The previous Array.isArray(networkChunk) flattening logic seems like
  it still needs to stay in this branch.

graysonhicks: thanks, addressed 👍

TheIsrael1 (approved):

PR #15227, opened 2026-04-09
fix(tools): preserve args for programmatic tool calls when merging synthetic tool-call

Akash504-ai:
## Description

Fixes an issue where programmatic tool calls (PTC) receive empty `{}` arguments during streaming.

In the streaming pipeline, a synthetic tool-call is created from `tool-call-input-streaming-end` with empty args. When the real `tool-call` event arrives (which contains the correct args), it is ignored if a synthetic already exists.

This change ensures that if a synthetic tool-call has empty args, the args from the real tool-call event are merged into it.

## Related Issue(s)

Fixes #15224

## Type of Change

## Checklist

## ELI5 (Explain Like I'm 5)

When AI tools are called programmatically while the model is thinking, the system was receiving the tool arguments as empty `{}` instead of the actual values the model provided. This fix detects when a partially-created tool call (made during streaming) has empty arguments, and fills in the correct arguments when the complete tool call information arrives later.

## Problem Summary

Programmatic tool calls (PTCs) were receiving empty `{}` arguments instead of the correct values during streaming. The issue occurred because:

- The streaming pipeline creates a synthetic `tool-call` from `tool-call-input-streaming-end` events with empty arguments.
- When the real `tool-call` event arrives containing the actual arguments, the existing synthetic tool-call is found and the handler returns early.
- Previously, only `providerMetadata` and `dynamic` properties were merged; the critical `args` property was not merged, so the real tool-call arguments were discarded.
- This caused programmatic tool invocations to fail validation since they received `{}` instead of expected parameters.

### packages/core/src/stream/base/output.ts

Enhanced the `tool-call` chunk handler in `MastraModelOutput` to preserve arguments when merging synthetic and real tool-calls:

- If an existing synthetic tool-call is found and its `payload.args` is empty or undefined, the code now copies/merges the incoming real `tool-call` chunk's `payload.args` into the synthetic tool-call.
- Existing merge behavior for `providerMetadata` and `dynamic` properties is preserved.
- The function still returns early after merging to avoid duplicate tool-calls.

### packages/core/src/stream/base/output.test.ts

Added a unit test that validates the fix:

- Simulates the streaming sequence: `tool-call-input-start` → `tool-call-input-end` (creating a synthetic with empty args) → real `tool-call` chunk with actual arguments.
- Verifies that after consuming the stream, the recorded tool call contains the correctly merged arguments from the real event (`{ query: 'SELECT 1' }`).

### .changeset/fix-ptc-args.md

Added a changeset announcing a patch release for `@mastra/core` describing the bug fix.

## Impact

- Type: Bug fix (non-breaking)
- Scope: Fixes programmatic tool calling in streaming scenarios (e.g., Anthropic provider) so tools receive the correct JSON args.
- Tests: New test covers the reported failure scenario and validates the merge behavior.


TylerBarnes (changes requested):
looks good! but it needs a changeset

Akash504-ai:
Added a changeset 👍

TylerBarnes (approved):

PR #15239, opened 2026-04-10
fix: ensure listVectorStores always returns a string id

### LINKED ISSUE #15238: Bug: listVectorStores handler returns undefined id instead of fallback to vector name
### Description

The legacy `listVectorStores` function in `packages/server/src/server/handlers/vector.ts` (lines 190-207) maps vector entries using `id: vector.id` without any fallback, which can produce `undefined` values when a `MastraVector` instance does not have an `id` property set.

This directly violates the `listVectorsResponseSchema` which declares `id` as a **required** `z.string()` field (non-optional):

```ts
// packages/server/src/server/schemas/vectors.ts
export const listVectorsResponseSchema = z.object({
  vectors: z.array(z.object({
    id: z.string(),   // <-- required, not optional
    name: z.string(),
    type: z.string(),
    ...
  }))
});
```

### Root Cause

The newer `LIST_VECTORS_ROUTE` (lines 383-410) correctly uses:
```ts
id: vector.id || name,  // fallback to the registration key if id is not set
```

But the older standalone `listVectorStores` function uses:
```ts
id: vector.id,  // no fallback — can be undefined!
```

### Impact

- When a vector store is registered without an explicit `id` property, `listVectorStores` returns `{ id: undefined, name, type }`
- This fails Zod schema validation (schema expects a string, gets undefined)
- Any client calling the legacy route gets malformed data or a runtime error

### Fix

Apply the same `id: vector.id || name` fallback pattern used in `LIST_VECTORS_ROUTE` to the legacy `listVectorStores` function:

```ts
const vectorList = Object.entries(vectors).map(([name, vector]) => ({
  name,
  id: vector.id || name,  // fallback to registration key
  type: vector.constructor.name,
}));
```

### File

`packages/server/src/server/handlers/vector.ts` — `listVectorStores` function (~line 198)

kuishou68:
Closes #15238

## Problem
The legacy `listVectorStores` function (lines ~198-202) maps vector entries using `id: vector.id` with no fallback, which returns `undefined` when a `MastraVector` instance has no `id` property. This violates the `listVectorsResponseSchema` which declares `id: z.string()` as required (non-optional), causing validation failures or `undefined` values leaking into the API response.

The newer `LIST_VECTORS_ROUTE` handler (lines ~401) correctly uses `id: vector.id || name` but the legacy function was never updated to match.

## Fix
Changed `id: vector.id` to `id: vector.id || name` in the legacy handler, making it consistent with the newer route handler and ensuring `id` is always a defined string.

## ELI5
Some stored "vector" items were missing an ID and showed up as "undefined", which broke responses. This change ensures every item always has an ID by using the registration name as a fallback when the item's own ID is missing.

## Changes
- packages/server/src/server/handlers/vector.ts
  - listVectorStores: map each vector entry to include `id: vector.id || name` (use registration key `name` as fallback), ensuring `id` is always a string.

- .changeset/fix-vector-store-undefined-id.md
  - Add changeset for @mastra/core with a `patch` bump and entry: "fix: ensure listVectorStores always returns a string id".

## Impact
- Prevents schema validation failures by guaranteeing `id` satisfies `listVectorsResponseSchema` (`id: z.string()`).
- Aligns legacy `listVectorStores` behavior with the newer `LIST_VECTORS_ROUTE`.
- No public API/signature changes; internal response mapping only.


kuishou68:
Added a changeset to address the `changeset-bot` warning. The file `.changeset/fix-vector-store-undefined-id.md` has been committed to this branch, marking `@mastra/core` for a `patch` version bump:

```
---
"@mastra/core": patch
---

fix: ensure listVectorStores always returns a string id
```

wardpeet (approved):

PR #15242, opened 2026-04-10
feat(observability): Add unique IDs to all observability signals

### LINKED ISSUE #15223: Add IDs to all observability signals
Add unique IDs to all observability signals, created at event emission time. 

Spans already have unique IDs (traceId + spanId)

But we should create unique IDs for each log, metric, score, feedback, etc..

LogId would ideally use the same format as logId in otel...if a standard exists. 

Others can be just uuids. 

Primary use case today is use as a de-duplication key in our OLAP stores, but also can help with linking signals to other systems.

intojhanurag:
## Description

Added unique IDs to all observability signals (logs, metrics, scores, feedback). Each event now carries a framework-generated `logId`, `metricId`, `scoreId`, or `feedbackId` created at emission time using `crypto.randomUUID()`. This aligns with the OTel `log.record.uid` semantic convention (which allows UUID). Framework-managed emission APIs (`logger.info()`, `metrics.emit()`, `addScore()`, `addFeedback()`) are unchanged — IDs are populated automatically by the framework. IDs flow end-to-end from emission through the bus, exporters, and record builders into ClickHouse and DuckDB storage with dedicated columns.

## Related Issue(s)

Fixes #15223

## Type of Change

- [x] Breaking change (TypeScript-level: see Design notes)
- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [x] Test update

## Checklist

## Overview

Framework-generated UUID identifiers (`logId`, `metricId`, `scoreId`, `feedbackId`) are created at emission time via `generateSignalId()` (`crypto.randomUUID()`) and propagated end-to-end through the observability pipeline: emission points, event bus, exporters, record builders, and persistence layers (ClickHouse and DuckDB). Core transport types include the new ID fields. Storage schemas include the new ID columns with non-destructive migration paths for existing deployments.

### ID generation
- Added `generateSignalId()` in `packages/core/src/observability/ids.ts` (uses `crypto.randomUUID()`) and re-exported from the observability index.
- Vitest tests validate UUID format and uniqueness.

### Signal emission
- `LoggerContextImpl.log()` adds `logId` to `ExportedLog`.
- `MetricsContextImpl.emit()` adds `metricId` to `ExportedMetric`.
- `buildScoreEvent()` and `buildFeedbackEvent()` include `scoreId` and `feedbackId` via `generateSignalId()`.
- Server handlers for `CREATE_SCORE` / `CREATE_FEEDBACK` inject `scoreId` / `feedbackId` when absent.

### Transport type changes
- `ExportedLog`: `+logId: string`
- `ExportedMetric`: `+metricId: string`
- `ExportedScore`: `+scoreId: string`
- `ExportedFeedback`: `+feedbackId: string`

### Storage schema & persistence

**ClickHouse v-next**
- DDL adds `String` (non-nullable) columns for `metricId` / `logId` / `scoreId` / `feedbackId`.
- Signal tables switched from `MergeTree` to `ReplacingMergeTree` with the signal ID appended to `ORDER BY`. Dedup guarantees apply to the framework exporter retry path (same buffered event + original timestamp → ReplacingMergeTree collapses). External HTTP writes are not idempotent under the current schema because the server re-stamps `timestamp` per request and reads do not dedupe by signal ID at query time — tracked as follow-up.
- **Non-destructive migration** (per maintainer feedback — no drop+recreate): `migrateSignalTables()` detects tables still on `MergeTree`, renames each to `<table>_backup_<timestamp>`, creates the new `ReplacingMergeTree` table, then `INSERT … SELECT` with explicit columns — backfilling missing IDs via `toString(generateUUIDv4())` and writing `NULL` for columns absent in the backup — and finally drops the backup. On failure, the original table is restored from the backup. Follows the same pattern as `migrateSpansTableSortingKey`.
- Row ↔ record helpers updated to map the new columns.

**DuckDB**
- DDL adds `VARCHAR NOT NULL PRIMARY KEY` ID columns.
- **Non-destructive migration**: `migrateSignalTables()` detects tables missing the primary-key ID column, renames to `<table>_backup_<timestamp>`, creates the new table with the PK, then `INSERT … SELECT` with explicit columns — backfilling IDs via `CAST(uuid() AS VARCHAR)` and writing `NULL` for columns absent in the backup — and drops the backup. On failure, the original table is restored from the backup.
- INSERT paths use `|| generateSignalId()` fallback so callers without IDs still produce valid rows.

### Tests & fixtures
- Updated test helpers and exporter/record-builder tests to include ID fields in fixtures and assertions.
- Added tests for `generateSignalId()`.
- Changeset added to bump package versions for coordinated release.

## Design notes

- **Backward compatibility (data)**: Existing rows are preserved through the copy-and-swap migration; new rows always carry IDs.
- **OTel alignment**: `logId` follows the OpenTelemetry `log.record.uid` semantic convention (UUID allowed).
- **Runtime API unchanged**: `logger.info()`, `metrics.emit()`, `addScore()`, `addFeedback()` work exactly as before — IDs are injected by the framework.
- **TS type surface is breaking**: `ExportedLog`, `ExportedMetric`, `ExportedScore`, `ExportedFeedback` now require `logId` / `metricId` / `scoreId` / `feedbackId` when constructed directly in TypeScript. Callers relying on framework-managed APIs are unaffected. Deferring exact semver bump (minor vs major) to maintainer preference.
- **Migration safety**: Both backends use a rename → create-new → backfill-from-backup → drop-backup sequence so init is idempotent and existing data is never dropped.


wardpeet:
@coderabbitai full review

packages/_internal-core/src/storage/domains/observability/logs.ts
```diff
@@ -31,6 +31,7 @@ const logDataField = z.record(z.string(), z.unknown()).describe('Structured data
  */
 export const logRecordSchema = z
   .object({
+    logId: z.string().nullish().describe('Unique id for this log event (de-duplication key)'),
```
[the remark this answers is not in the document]
intojhanurag: Hey @coderabbitai  The framework guarantees IDs at every emission point (generateSignalId() in logger, metrics, recorded.ts, and server endpoints), so the create path always has a non-null ID in practice

stores/clickhouse/src/storage/domains/observability/v-next/ddl.ts
```diff
@@ -556,6 +559,17 @@ export const DISCOVERY_MV_DDL = [DISCOVERY_VALUES_MV_DDL, DISCOVERY_PAIRS_MV_DDL
...
+/**
+ * Signal tables that need to be migrated from MergeTree to ReplacingMergeTree.
+ * Used by init() to check the current engine and drop+recreate only if needed.
+ */
+export const SIGNAL_TABLES_REQUIRING_ENGINE_MIGRATION = [
+  TABLE_METRIC_EVENTS,
+  TABLE_LOG_EVENTS,
+  TABLE_SCORE_EVENTS,
+  TABLE_FEEDBACK_EVENTS,
+];
```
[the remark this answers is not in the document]
intojhanurag: Hey @coderabbitai , The drop+recreate is intentional and guarded — init() checks system.tables.engine and
only drops tables that are still on the old MergeTree engine. Tables already on
ReplacingMergeTree are skipped. Per maintainer (@epinzur): "I don't think anyone is using
Clickhouse vNext in production yet," so a copy-and-swap migration is unnecessary overhead.
When vNext reaches production, the team will need the rename-and-swap pattern for future
schema changes, but that's out of scope for this PR.

stores/clickhouse/src/storage/domains/observability/v-next/index.ts
```diff
@@ -124,6 +125,20 @@ export class ObservabilityStorageClickhouseVNext extends ObservabilityStorage {
...
+      // Tables already on ReplacingMergeTree (or not yet created) are skipped.
+      for (const table of SIGNAL_TABLES_REQUIRING_ENGINE_MIGRATION) {
+        const result = await this.#client.query({
+          query: `SELECT engine FROM system.tables WHERE database = currentDatabase() AND name = '${table}'`,
+          format: 'JSONEachRow',
+        });
+        const rows = (await result.json()) as Array<{ engine: string }>;
+        if (rows.length > 0 && rows[0]!.engine !== 'ReplacingMergeTree') {
+          await this.#client.command({ query: `DROP TABLE IF EXISTS ${table}` });
+        }
```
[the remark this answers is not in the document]
intojhanurag: Hey @coderabbitai The drop+recreate is intentional and guarded — init() checks system.tables.engine and only drops tables that are still on the old MergeTree engine. Tables already on ReplacingMergeTree are skipped. Per maintainer (@epinzur): "I don't think anyone is using Clickhouse vNext in production yet," so a copy-and-swap migration is unnecessary overhead. When vNext reaches production, the team will need the rename-and-swap pattern for future schema changes, but that's out of scope for this PR.

.changeset/stale-plums-occur.md
```diff
@@ -0,0 +1,8 @@
+---
+'@mastra/core': minor
+'@mastra/observability': minor
+'@mastra/clickhouse': minor
+'@mastra/duckdb': minor
+---
+
+Added unique IDs to all observability signals (logs, metrics, scores, feedback). Each event now carries a framework-generated `logId`, `metricId`, `scoreId`, or `feedbackId` created at emission time, acting as a de-duplication key in downstream OLAP stores and letting external systems link a specific signal back to Mastra. Log IDs align with the OpenTelemetry `log.record.uid` semantic convention. User code that calls `logger.info()`, `metrics.emit()`, `addScore()`, and `addFeedback()` does not change; the IDs are populated automatically. Closes #15223.
```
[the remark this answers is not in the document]
intojhanurag: Hey @coderabbitai, this was already addressed in an earlier commit — the changeset text was simplified to remove jargon and focus on the outcome.

stores/duckdb/src/storage/domains/observability/ddl.ts
```diff
@@ -66,6 +66,7 @@ CREATE TABLE IF NOT EXISTS metric_events (
   timestamp TIMESTAMP NOT NULL,
 
   -- IDs
+  metricId VARCHAR,
   traceId VARCHAR,
```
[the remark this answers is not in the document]
intojhanurag: Hey @coderabbitai, this was addressed in a later commit — DuckDB migrations now include `ALTER TABLE ADD COLUMN IF NOT EXISTS` for all 4 signal ID columns, plus the DDL uses `PRIMARY KEY` on the signal ID columns for dedup enforcement. Tables without a PRIMARY KEY are dropped and recreated on init.

stores/duckdb/src/storage/domains/observability/feedback.ts
```diff
@@ -161,6 +161,7 @@ function rowToFeedbackRecord(row: Record<string, unknown>): Record<string, unkno
   if (!isNaN(numValue)) value = numValue;
 
   return {
+    feedbackId: (row.feedbackId as string) ?? '',
     timestamp: toDate(row.timestamp),
```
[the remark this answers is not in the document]
intojhanurag: Hey @coderabbitai, the `?? ''` is a last-resort fallback that should never trigger — every code path that writes signals already generates an ID via `generateSignalId()`. If it somehow does trigger, the `PRIMARY KEY` constraint catches it: the first `''` insert succeeds, the second fails with a duplicate-key error. This is fail-fast by design — a loud error that surfaces the real bug (missing ID at emission) rather than silently masking it by generating an ID in the store layer.

stores/duckdb/src/storage/domains/observability/logs.ts
```diff
@@ -39,6 +40,7 @@ const COLUMNS_SQL = COLUMNS.join(', ');
 
 function rowToLogRecord(row: Record<string, unknown>): Record<string, unknown> {
   return {
+    logId: (row.logId as string) ?? '',
     timestamp: toDate(row.timestamp),
```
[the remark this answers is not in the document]
intojhanurag: Hey @coderabbitai, same reasoning as the feedback comment — the `?? ''` is a defensive fallback that should never trigger. All emission paths generate IDs via `generateSignalId()`. The `PRIMARY KEY` constraint on `logId` makes this fail-fast: if an empty string somehow gets written twice, the duplicate-key error surfaces the upstream bug immediately.

stores/duckdb/src/storage/domains/observability/metrics.ts
```diff
@@ -201,6 +202,7 @@ function resolveGroupBy(groupBy: string[]): ResolvedGroupBy[] {
 
 function rowToMetricRecord(row: Record<string, unknown>): Record<string, unknown> {
   return {
+    metricId: (row.metricId as string) ?? '',
```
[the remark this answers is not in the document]
intojhanurag: Hey @coderabbitai, same pattern as the other signal ID comments — the `?? ''` fallback is defensive and never triggers in practice since all emission paths call `generateSignalId()`. The `PRIMARY KEY` on `metricId` enforces uniqueness at the DB level, so a duplicate empty string would fail fast with a constraint violation rather than silently corrupting data.

stores/duckdb/src/storage/domains/observability/scores.ts
```diff
@@ -151,6 +151,7 @@ function toSeriesName(values: unknown[]): string {
 
 function rowToScoreRecord(row: Record<string, unknown>): Record<string, unknown> {
   return {
+    scoreId: (row.scoreId as string) ?? '',
```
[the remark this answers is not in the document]
intojhanurag: Hey @coderabbitai, same reasoning as the other signal ID fallback comments — every write path generates an ID via `generateSignalId()`, so the `?? ''` never triggers. The `PRIMARY KEY` on `scoreId` makes this fail-fast by design: if an empty string somehow gets written twice, the constraint violation surfaces the real bug immediately.

intojhanurag:
Hey @wardpeet , can you confirm that no-one is using Clickhouse and duckdb in production , if luckily it is true then this PR is good to go otherwise we need to think any other workaround .

epinzur:
@intojhanurag we have no way to confirm usage... we will need to provide a migration tool.  I'll discuss on the call.

packages/core/src/observability/ids.test.ts
```diff
@@ -0,0 +1,18 @@
+import { describe, expect, it } from 'vitest';
```
epinzur: this test can be dropped.  we don't need to test external impelmentations.

packages/core/src/observability/ids.ts
```diff
@@ -0,0 +1,4 @@
+/** Generate a unique id for an observability signal (log, metric, score, feedback). */
+export function generateSignalId(): string {
+  return crypto.randomUUID();
+}
```
epinzur: maybe this can be elsewhere instead of in its own file?

stores/clickhouse/src/storage/domains/observability/v-next/ddl.ts
```diff
@@ -253,13 +254,13 @@ CREATE TABLE IF NOT EXISTS ${TABLE_METRIC_EVENTS} (
   metadata           Nullable(String),
   scope              Nullable(String)
 )
-ENGINE = MergeTree
+ENGINE = ReplacingMergeTree
 PARTITION BY toDate(timestamp)
-ORDER BY (name, timestamp)
+ORDER BY (name, metricId)
```
epinzur: keep timestamp, and add metricId. 

```
ORDER BY (name, timestamp, metricId)
```

stores/clickhouse/src/storage/domains/observability/v-next/ddl.ts
```diff
@@ -310,14 +312,13 @@ CREATE TABLE IF NOT EXISTS ${TABLE_LOG_EVENTS} (
   metadata           Nullable(String),
   scope              Nullable(String)
 )
-ENGINE = MergeTree
+ENGINE = ReplacingMergeTree
 PARTITION BY toDate(timestamp)
-ORDER BY (timestamp, traceId)
-SETTINGS allow_nullable_key = 1
+ORDER BY (logId)
```
epinzur:
```
ORDER BY (timestamp, logId)
```

stores/clickhouse/src/storage/domains/observability/v-next/ddl.ts
```diff
@@ -375,14 +377,14 @@ CREATE TABLE IF NOT EXISTS ${TABLE_SCORE_EVENTS} (
   metadata           Nullable(String),
   scope              Nullable(String)
 )
-ENGINE = MergeTree
+ENGINE = ReplacingMergeTree
 PARTITION BY toDate(timestamp)
-ORDER BY (traceId, timestamp)
+ORDER BY (traceId, scoreId)
```
epinzur:
```
ORDER BY (traceId, timestamp, scoreId)
```

stores/clickhouse/src/storage/domains/observability/v-next/ddl.ts
```diff
@@ -443,9 +446,9 @@ CREATE TABLE IF NOT EXISTS ${TABLE_FEEDBACK_EVENTS} (
   metadata           Nullable(String),
   scope              Nullable(String)
 )
-ENGINE = MergeTree
+ENGINE = ReplacingMergeTree
 PARTITION BY toDate(timestamp)
-ORDER BY (traceId, timestamp)
+ORDER BY (traceId, feedbackId)
```
epinzur:
```
ORDER BY (traceId, timestamp, feedbackId)
```

stores/clickhouse/src/storage/domains/observability/v-next/feedback.ts
```diff
@@ -189,13 +189,13 @@ export async function listFeedback(client: ClickHouseClient, args: ListFeedbackA
 
   const countResult = await queryJson<{ total?: number }>(
     client,
-    `SELECT count() AS total FROM ${TABLE_FEEDBACK_EVENTS} AS f ${whereClause}`,
+    `SELECT count() AS total FROM ${TABLE_FEEDBACK_EVENTS} AS f FINAL ${whereClause}`,
```
epinzur: what does `FINAL` do?

stores/clickhouse/src/storage/domains/observability/v-next/helpers.ts
```diff
@@ -252,6 +253,7 @@ export function spanRecordToRow(span: CreateSpanRecord): Record<string, unknown>
 
 export function rowToLogRecord(row: Record<string, any>): LogRecord {
   return {
+    logId: nullableString(row.logId),
```
epinzur: `logId` should never be null after the migration.

stores/clickhouse/src/storage/domains/observability/v-next/helpers.ts
```diff
@@ -289,6 +291,7 @@ export function rowToLogRecord(row: Record<string, any>): LogRecord {
 
 export function logRecordToRow(log: CreateLogRecord): Record<string, unknown> {
   return {
+    logId: log.logId || generateSignalId(),
```
epinzur: `logId` should never be null after the migration.

stores/clickhouse/src/storage/domains/observability/v-next/index.ts
```diff
@@ -125,6 +126,9 @@ export class ObservabilityStorageClickhouseVNext extends ObservabilityStorage {
 
   async init(): Promise<void> {
     try {
+      // Non-destructive migration: MergeTree signal tables → ReplacingMergeTree with signal-ID.
+      await migrateSignalTables(this.#client, this.logger);
```
epinzur: we should not migrate at init() time.  instead we should force the user to quit and migrate in a separate path.

stores/clickhouse/src/storage/domains/observability/v-next/migration.ts
```diff
@@ -0,0 +1,112 @@
...
+async function getTableColumns(client: ClickHouseClient, table: string): Promise<string[]> {
+  const result = await client.query({ query: `DESCRIBE TABLE ${table}`, format: 'JSONEachRow' });
+  const rows = (await result.json()) as Array<{ name: string }>;
+  return rows.map(r => r.name);
+}
+
+/**
+ * Migrate signal tables from MergeTree to ReplacingMergeTree without dropping data.
+ * Copy-and-swap: rename → create new → INSERT…SELECT (generating IDs) → drop backup.
+ * On failure the original table is restored from the backup. Idempotent.
```
epinzur: we should probably do the opposite logic here. 

* create new under a temp name
* insert..select (generating IDs) into the new table
* rename old to backup
* rename new to correct name

This makes failure case easier before the final 2 steps... especially if the user just ctrl-c's out of the process.

stores/duckdb/src/storage/domains/observability/feedback.ts
```diff
@@ -242,12 +244,13 @@ export async function createFeedback(db: DuckDBConnection, args: CreateFeedbackA
...
   await db.execute(
     `INSERT INTO feedback_events (
-      timestamp, traceId, spanId, experimentId,
+      feedbackId, timestamp, traceId, spanId, experimentId,
       entityType, entityId, entityName, entityVersionId, parentEntityVersionId, parentEntityType, parentEntityId, parentEntityName, rootEntityVersionId, rootEntityType, rootEntityId, rootEntityName,
       userId, organizationId, resourceId, runId, sessionId, threadId, requestId, environment, executionSource, serviceName,
       feedbackUserId, sourceId, feedbackSource, feedbackType, value, comment, tags, metadata, scope
     )
      VALUES (${[
+       v(f.feedbackId || generateSignalId()),
```
epinzur: we should always assume feedbackId exists after migration... if we get a feedback without an Id, we should error.

same with the other signals

stores/duckdb/src/storage/domains/observability/index.ts
```diff
@@ -96,6 +97,9 @@ export class ObservabilityStorageDuckDB extends ObservabilityStorage {
 
   /** Create all observability tables if they don't exist. */
   async init(): Promise<void> {
+    // Non-destructive migration for signal tables missing the signal-ID PK.
+    await migrateSignalTables(this.db, this.logger);
```
epinzur: no migrate on init.  instead return msg to user that they should use the `mastra migrate` command.

stores/clickhouse/src/storage/domains/observability/v-next/feedback.ts
intojhanurag: like Final forces query-time dedup for `ReplacingMergeTree`. Background merges eventually collapse rows with the same sorting key, but not synchronously, so without `FINAL` recent retries can still appear twice. I used it here so list/count reads see the deduplicated view immediately.

Actually deduplicated can be see without Final as well but not immediately .

intojhanurag:
@coderabbitai full review

stores/clickhouse/src/storage/domains/observability/v-next/index.ts
```diff
@@ -124,6 +157,19 @@ export class ObservabilityStorageClickhouseVNext extends ObservabilityStorage {
...
+      throw new MastraError({
+        id: createStorageErrorId('CLICKHOUSE', 'MIGRATION_REQUIRED', 'SIGNAL_TABLES'),
+        domain: ErrorDomain.STORAGE,
+        category: ErrorCategory.USER,
+        text: buildSignalMigrationRequiredMessage({
+          store: 'ClickHouse',
+          tables: migrationStatus.tables.map(({ table, engine }) => ({ table, engine })),
+        }),
+      });
+    }
```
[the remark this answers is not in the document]
intojhanurag: Intentionally out of scope. The preflight `checkSignalTablesMigrationStatus` throws a `MastraError` when migration is needed, which is the expected user-facing error. Wrapping raw client-connection errors from the probe itself is a broader adapter-error-consistency cleanup that should land separately from this signal-ID migration PR.

stores/clickhouse/src/storage/domains/observability/v-next/metrics.ts
```diff
@@ -245,13 +245,13 @@ export async function listMetrics(client: ClickHouseClient, args: ListMetricsArg
...
-    `SELECT count() AS total FROM ${TABLE_METRIC_EVENTS} AS m ${whereClause}`,
+    `SELECT count() AS total FROM ${TABLE_METRIC_EVENTS} AS m FINAL ${whereClause}`,
     filter.params,
   );
 
   const rows = await queryJson<Record<string, any>>(
     client,
-    `SELECT * FROM ${TABLE_METRIC_EVENTS} AS m ${whereClause} ORDER BY ${orderBy} LIMIT {limit:UInt32} OFFSET {offset:UInt32}`,
+    `SELECT * FROM ${TABLE_METRIC_EVENTS} AS m FINAL ${whereClause} ORDER BY ${orderBy} LIMIT {limit:UInt32} OFFSET {offset:UInt32}`,
     { ...filter.params, limit: pagination.limit, offset: pagination.offset },
```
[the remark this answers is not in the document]
intojhanurag: This is valid ClickHouse syntax. Both `FROM table AS alias FINAL` and `FROM table FINAL AS alias` are accepted. The full integration suite runs against ClickHouse 24.8 and all 416 tests pass — every `listMetrics`/`listScores`/`listFeedback`/`listLogs` query uses this exact form successfully.

stores/clickhouse/src/storage/domains/observability/v-next/migration.ts
```diff
@@ -0,0 +1,155 @@
...
+
+      logger?.info?.(`Successfully migrated ${table}`);
+    } catch (error) {
+      logger?.error?.(`Migration of ${table} failed: ${(error as Error).message}`);
+      try {
+        await dropTableIfExists(client, temp);
+      } catch (restoreError) {
+        logger?.error?.(`Failed to clean up temporary table ${temp}: ${(restoreError as Error).message}`);
+      }
+      throw createMigrationError({ table, idColumn }, error);
```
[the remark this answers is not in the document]
intojhanurag: `RENAME TABLE a TO b, c TO d` is atomic in ClickHouse (single statement under the DDL lock). The explicit backup is deliberate — it's only dropped after a successful swap, so the data is recoverable if anything goes wrong. `EXCHANGE TABLES` would eliminate that recoverable backup without simplifying the code path.

stores/clickhouse/src/storage/domains/observability/v-next/scores.ts
```diff
@@ -184,13 +184,13 @@ export async function listScores(client: ClickHouseClient, args: ListScoresArgs)
...
     client,
-    `SELECT count() AS total FROM ${TABLE_SCORE_EVENTS} AS s ${whereClause}`,
+    `SELECT count() AS total FROM ${TABLE_SCORE_EVENTS} AS s FINAL ${whereClause}`,
     filter.params,
   );
 
   const rows = await queryJson<Record<string, any>>(
     client,
-    `SELECT * FROM ${TABLE_SCORE_EVENTS} AS s ${whereClause} ORDER BY ${orderBy} LIMIT {limit:UInt32} OFFSET {offset:UInt32}`,
+    `SELECT * FROM ${TABLE_SCORE_EVENTS} AS s FINAL ${whereClause} ORDER BY ${orderBy} LIMIT {limit:UInt32} OFFSET {offset:UInt32}`,
```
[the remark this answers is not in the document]
intojhanurag: Same as the metrics.ts comment — valid ClickHouse syntax, covered by passing integration tests against ClickHouse 24.8.

stores/duckdb/src/storage/domains/observability/ddl.ts
```diff
@@ -220,6 +223,7 @@ CREATE TABLE IF NOT EXISTS feedback_events (
   timestamp TIMESTAMP NOT NULL,
 
   -- IDs
+  feedbackId VARCHAR NOT NULL PRIMARY KEY,
```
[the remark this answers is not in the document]
intojhanurag: Already handled — `stores/duckdb/src/storage/domains/observability/feedback.ts:289` has `ON CONFLICT DO NOTHING` on the single-row `createFeedback` INSERT (and the batch path too). No duplicate-key errors on retry.

stores/duckdb/src/storage/domains/observability/migration.test.ts
```diff
@@ -0,0 +1,179 @@
...
+            throw new Error('Simulated INSERT failure');
+          }
+          // @ts-expect-error proxied call
+          return target.execute(sql, ...rest);
+        };
+      }
+      return Reflect.get(target, prop, receiver);
+    },
+  }) as DuckDBConnection;
+}
```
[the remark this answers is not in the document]
intojhanurag: Intentional — this test exercises the INSERT-failure → rollback branch, which is a real failure mode (constraint violation, OOM, schema mismatch). The RENAME-failure branch is a separate scenario; replacing this test with a RENAME-target version would trade one form of coverage for another rather than adding to it.

stores/duckdb/src/storage/domains/observability/migration.ts
```diff
@@ -0,0 +1,171 @@
...
+  return rows.length > 0;
+}
+
+async function hasPrimaryKey(db: DuckDBConnection, table: string): Promise<boolean> {
+  const rows = await db.query<{ constraint_type: string }>(
+    `SELECT constraint_type FROM information_schema.table_constraints
+     WHERE table_name = ? AND constraint_type = 'PRIMARY KEY'`,
+    [table],
+  );
+  return rows.length > 0;
```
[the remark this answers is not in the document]
intojhanurag: Legacy DuckDB signal tables have no PRIMARY KEY constraint at all — the shape is plain `CREATE TABLE ... ORDER BY timestamp`. A PK on a different column is not a state this migration can encounter in practice, so the table-level check is sufficient.

packages/_internal-core/src/storage/domains/observability/feedback.ts
```diff
@@ -56,6 +56,7 @@ function normalizeLegacyFeedbackActor<T>(input: T): T {
  * Includes all fields from ExportedFeedback plus storage-specific fields.
  */
 const feedbackRecordObjectSchema = z.object({
+  feedbackId: z.string().nullish().describe('Unique id for this feedback event (de-duplication key)'),
```
epinzur: can we remove the "(de-duplication key)" from these describes()?

stores/clickhouse/src/storage/domains/observability/v-next/index.ts
```diff
@@ -2,8 +2,8 @@
  * ClickHouse v-next observability storage domain.
  *
  * Insert-only model: only completed spans stored. Uses ReplacingMergeTree for
- * tracing tables with dedupeKey for retry-idempotency. Append-only MergeTree
- * for metric/log/score/feedback signals.
+ * tracing tables with dedupeKey for retry-idempotency, and ReplacingMergeTree
+ * for metric/log/score/feedback signals with signal IDs for de-duplication.
```
epinzur: i think this can be simplified:

```
 * Insert-only model: Uses ReplacingMergeTree for all signals
 * with dedupeKey for retry-idempotency,
```

stores/clickhouse/src/storage/domains/observability/v-next/migration.ts
```diff
@@ -0,0 +1,155 @@
...
+          }
+          return currentColumns.has(c) ? `"${c}"` : `NULL AS "${c}"`;
+        })
+        .join(', ');
+
+      await client.command({
+        query: `INSERT INTO ${temp} (${columnList}) SELECT ${selectExprs} FROM ${table}`,
+      });
+
+      await client.command({ query: `RENAME TABLE ${table} TO ${backup}, ${temp} TO ${table}` });
```
epinzur: i wasn't familiar with this double rename trick, so I looked at the clickhouse docs. pretty neat.

I also saw its possible to do an exchange:

```
query: `EXCHANGE TABLES ${temp} AND ${table}`
```

which might even be better for our use case. But I'll leave it up to you.

stores/duckdb/src/storage/domains/observability/index.ts
```diff
@@ -73,9 +74,39 @@ import * as discoveryOps from './discovery';
...
+    `\n` +
+    `  npx mastra migrate\n` +
+    `\n` +
+    `This command will:\n` +
+    `  1. Create replacement signal tables with signal-ID primary keys\n` +
+    `  2. Backfill missing signal IDs for legacy rows\n` +
+    `  3. Swap the migrated tables into place\n` +
+    `\n` +
+    `WARNING: This migration recreates the signal tables and may take significant\n` +
+    `time for large datasets. Please ensure you have a backup before proceeding.\n` +
```
epinzur: large databases

stores/clickhouse/src/storage/domains/observability/v-next/index.ts
```diff
@@ -104,10 +104,43 @@ import * as discoveryOps from './discovery';
...
+    `\n` +
+    `  npx mastra migrate\n` +
+    `\n` +
+    `This command will:\n` +
+    `  1. Create replacement signal tables with signal-ID dedupe keys\n` +
+    `  2. Backfill missing signal IDs for legacy rows\n` +
+    `  3. Swap the migrated tables into place\n` +
+    `\n` +
+    `WARNING: This migration recreates the signal tables and may take significant\n` +
+    `time for large datasets. Please ensure you have a backup before proceeding.\n` +
```
epinzur: large databases

epinzur (approved):

PR #14576, opened 2026-03-21
fix: capture stream errors via onError callback instead of onFinish payload

askadityapandey:
## Description

When an upstream error occurs during streaming (e.g. timeout from Google Vertex), 
Mastra was logging `Error in agent stream { error: undefined }` because the Vercel 
AI SDK does not include an `error` property on the `OnFinishEvent` payload by design. 
The actual error is passed exclusively to the `onError` callback.

This PR removes the dead error-handling code from `onFinish` and moves it to the 
`onError` callback where the actual error object is available. User-provided 
`options.onError` callbacks are still forwarded correctly.

## Related Issue(s)

Fixes #14575

## Type of Change

- [x] Bug fix (non-breaking change that fixes an issue)

## Checklist

- [ ] I have made corresponding changes to the documentation (if applicable)
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] I have addressed all Coderabbit comments on this PR

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved error handling for agent streams with clearer, more consistent reporting of upstream API errors.
  * Removed inconsistent or misleading metadata from stream error logs to make messages easier to interpret.

PR #15252, opened 2026-04-10
fix: capture stream errors via onError callback instead of onFinish payload

TheIsrael1:
## Summary
- move stream error capture to `onError`, which is where the loop delivers the real error object
- preserve upstream provider and model metadata from `APICallError.data` in logs when available
- keep `onFinish` focused on finalization while still avoiding finish-time persistence on failed streams
- add regression coverage for upstream stream-error logging and keep AGENT_RUN span-ending coverage intact
- add the missing `@mastra/core` patch changeset

## Testing
- `pnpm vitest run src/agent/__tests__/save-and-errors.test.ts`
- `pnpm check`

## Supersedes
- Supersedes #14576
- Fixes #14575

## ELI5

This PR fixes a bug so Mastra captures real streaming errors from AI providers by listening to the stream's onError callback (where the SDK delivers the actual error) instead of reading an absent error from the finish event. That lets Mastra log the true error and which provider/model caused it.

## Overview

Move upstream stream-error capture from onFinish to the onError callback so the loop receives the actual error object and preserved provider/model metadata. End agent spans on stream failure in the onError path and avoid persisting or treating finish-time payloads as the source of upstream errors.

## Key changes

- map-results-step.ts
  - Added normalized error conversion via getErrorFromUnknown and moved upstream error handling into the stream onError wrapper.
  - Extracts provider/modelId from APICallError.data and logs upstream LLM errors with provider/model context.
  - Records error on agentSpan with endSpan: true inside onError, then forwards to downstream options.onError.
  - Removed the previous onFinish-based error-logging/persistence for finishReason === 'error'.

- Loop types & network loop
  - Added LoopErrorInfo type: { error: Error | string; provider?: string; modelId?: string }.
  - Updated LoopConfig.onError signature and createNetworkLoop onError callback to accept provider/modelId alongside error.

- llm-execution-step.ts / processOutputStream
  - Threaded provider and modelId into processOutputStream and enriched error callbacks so onError receives { error, provider?, modelId? } when a stream error occurs.

- Tests
  - Added regression test (packages/core/src/agent/__tests__/save-and-errors.test.ts) verifying:
    - onError receives the original APICallError instance for v2 streams,
    - resolved provider (`google-vertex`) and modelId (`gemini-2.5-pro`) are passed through,
    - agent logger.error is called with an object containing error, runId, provider, and modelId.
  - Test is v2-only (returns early for v1).

- Changeset
  - .changeset/four-poems-jump.md: patch for @mastra/core documenting the fix and trace behavior.

## Rationale / linked issue

Fixes issue where Vercel AI SDK does not include errors in OnFinishEvent; the SDK delivers actual upstream errors via onError. This PR ensures Mastra logs the real error object and preserves upstream provider/model metadata when available. Supersedes #14576 and fixes #14575.

## Testing

- New unit test added: pnpm vitest run src/agent/__tests__/save-and-errors.test.ts
- pnpm check (CI expectations)

### DIFF
--- a/.changeset/four-poems-jump.md
+++ b/.changeset/four-poems-jump.md
@@ -1,0 +1,5 @@
+---
+'@mastra/core': patch
+---
+
+Fixed agent streaming errors to log upstream provider details correctly and end agent traces when the stream fails.
--- a/.changeset/four-poems-jump.md
+++ b/.changeset/four-poems-jump.md
@@ -2,4 +2,4 @@
 '@mastra/core': patch
 ---
 
-Fixed agent streaming errors to log upstream provider details correctly and end agent traces when the stream fails.
+Fixed agent streaming errors to use `onError` for logging, preserve resolved model metadata, and end agent traces when the stream fails.
--- a/packages/core/src/agent/__tests__/save-and-errors.test.ts
+++ b/packages/core/src/agent/__tests__/save-and-errors.test.ts
@@ -1286,7 +1286,7 @@ function saveAndErrorTests(version: 'v1' | 'v2') {
         expect(caughtError.message).toMatch(/Simulated stream error/);
       });
 
-      it('should log upstream stream errors with provider and model metadata from APICallError.data', async () => {
+      it('should pass resolved model metadata through onError for stream failures', async () => {
         if (version === 'v1') return;
 
         const upstreamError = new APICallError({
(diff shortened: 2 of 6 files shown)


TheIsrael1:
Closing this in favor if #15435

PR #15269, opened 2026-04-12
fix(memory): fall back to OM record lastObservedAt in getOtherThreadsContext

### LINKED ISSUE #15265: [BUG] Resource-scoped OM: getOtherThreadsContext ignores lastObservedAt for threads not observed in the triggering cycle
## Bug Description

When using **resource-scoped** Observational Memory (`scope: "resource"`), `getOtherThreadsContext()` loads **all messages from all threads** on every request — even after observation has successfully run — because `lastObservedAt` is missing from thread metadata for most threads.

## Root Cause

`getOtherThreadsContext()` (in the `ObservationalMemory` class) reads `lastObservedAt` from **thread metadata** (`thread.metadata.mastra.om.lastObservedAt`) to decide whether to apply a date filter on `listMessages`:

```js
const omMetadata = getThreadOMMetadata(thread.metadata);
const threadLastObservedAt = omMetadata?.lastObservedAt;
const startDate = threadLastObservedAt
  ? new Date(new Date(threadLastObservedAt).getTime() + 1)
  : void 0; // <-- undefined = no date filter = loads ALL messages
```

`ResourceScopedObservationStrategy.persist()` **does** write `lastObservedAt` to thread metadata — but only for threads in `threadMetadataUpdates`, i.e., threads that had **unobserved messages in that particular observation cycle**.

Threads that were already fully observed (no new messages since their `observedMessageIds` were recorded) are **skipped** by the observer and never get `lastObservedAt` written to their metadata. On subsequent requests, `getOtherThreadsContext()` sees `lastObservedAt = undefined` for those threads and loads their **entire message history** with no date filter.

Note that `SyncObservationStrategy.persist()` (thread-scoped) writes `lastObservedMessageCursor` to thread metadata but **never** writes `lastObservedAt`.

## Impact

For a user with 18 threads where only 1 was observed in the latest cycle:
- 17 threads have `metadata.mastra.om.lastObservedAt = undefined`
- Every request triggers 17 × `listMessages` with **no date filter** (full history scan)
- Produces ~72 DB spans and ~1.1s of pure DB I/O per request
- This is an **N+1 query pattern** that scales linearly with thread count

## Reproduction

1. Create an agent with resource-scoped OM
2. Send messages across multiple threads (e.g., 10+ threads)
3. Wait for observation to trigger (or lower `messageTokens` threshold)
4. After observation completes, send a new message to any thread
5. Query `mastra_threads` and check `metadata` column: most threads will NOT have `mastra.om.lastObservedAt`

**Verification query:**
```sql
SELECT id, title,
       metadata->'mastra'->'om'->>'lastObservedAt' as thread_lastObservedAt,
       metadata->'mastra'->'om'->>'lastObservedMessageCursor' as cursor
FROM mastra.mastra_threads
WHERE "resourceId" = '<your-resource-id>';
```

Compare with the OM record:
```sql
SELECT id, "lastObservedAt", "observationTokenCount"
FROM mastra.mastra_observational_memory
WHERE "resourceId" = '<your-resource-id>';
```

## Expected Behavior

After observation runs successfully, `getOtherThreadsContext()` should apply date filtering for ALL threads — not just the ones that happened to have unobserved messages during the triggering cycle.

## Suggested Fix

Two complementary fixes:

1. **In `ResourceScopedObservationStrategy.persist()`**: After processing `threadMetadataUpdates`, also update `lastObservedAt` on threads that were skipped (already fully observed). Use the OM record's `lastObservedAt` as the value.

2. **In `getOtherThreadsContext()`**: Fall back to the OM record's `lastObservedAt` when `thread.metadata.mastra.om.lastObservedAt` is missing:
   ```js
   const threadLastObservedAt = omMetadata?.lastObservedAt ?? record?.lastObservedAt;
   ```

## Environment

- `@mastra/memory`: 1.15.0
- `@mastra/core`: 1.24.0
- `scope: "resource"`
- Storage: `@mastra/pg` (PostgreSQL)

Yanhu007:
## Summary

Fixes #15265

`getOtherThreadsContext()` reads `lastObservedAt` from each thread's metadata to apply a date filter when loading messages. However, `ResourceScopedObservationStrategy.persist()` only stamps `lastObservedAt` on threads that had unobserved messages in that particular observation cycle. Threads that were already fully observed never get the field.

This causes `getOtherThreadsContext` to load the **entire message history** of those threads with no date filter on every request — an N+1 query pattern that scales linearly with thread count.

### Impact example (from issue)
For a user with 18 threads where only 1 was observed:
- 17 threads → 17 × `listMessages` with **no date filter** (full history scan)
- ~72 DB spans and ~1.1s of pure DB I/O per request

## Fix

Fetch the OM record once at the start of `getOtherThreadsContext()` and use its `lastObservedAt` as a fallback when per-thread metadata is missing:

```ts
const threadLastObservedAt = omMetadata?.lastObservedAt ?? recordLastObservedAt;
```

This is a minimal, safe change — the OM record's `lastObservedAt` is a conservative lower bound that eliminates the full-history scans while preserving correctness.

## ELI5

Imagine you're checking what happened in multiple chat rooms. Instead of just asking "what's new since we last checked?" for each room, you were accidentally loading ALL messages from each room on every request. This PR adds a backup plan: if a room doesn't have a "last checked" timestamp, use the one from the main record instead, so you always know where to start looking for new messages.

## Summary

This PR fixes an N+1 query and excessive DB I/O in getOtherThreadsContext() by falling back to the observational memory (OM) record's lastObservedAt when a thread's per-thread metadata lacks lastObservedAt.

Root cause
- ResourceScopedObservationStrategy.persist() only writes lastObservedAt for threads that had unobserved messages in a cycle. Threads that were already fully observed never receive this field.
- getOtherThreadsContext() used per-thread lastObservedAt directly; when it's undefined, no date filter was applied and the code scanned full thread histories, causing N+1 queries and high I/O.

Fix
- Read the OM record once at the start of getOtherThreadsContext() and cache its lastObservedAt.
- When iterating sibling threads, use:
  const threadLastObservedAt = omMetadata?.lastObservedAt ?? recordLastObservedAt;
  so a missing per-thread cursor falls back to the record-level timestamp, providing a conservative lower bound and avoiding full-history scans while preserving correctness.

Tests and release note
- Added a regression test in packages/memory/.../__tests__/observational-memory-api.test.ts that sets the record's lastObservedAt and verifies getOtherThreadsContext() only returns messages newer than that cursor (ensures older sibling messages are filtered out when thread metadata is missing).
- Added a changeset (.changeset/fresh-oranges-listen.md) documenting a patch release for @mastra/memory describing the fallback behavior.

Files changed (high level)
- packages/memory/src/processors/observational-memory/observational-memory.ts — use record-level lastObservedAt as fallback in getOtherThreadsContext (small, focused change).
- packages/memory/src/processors/observational-memory/__tests__/observational-memory-api.test.ts — new test covering the fallback behavior.
- .changeset/fresh-oranges-listen.md — release note for the patch.

Impact
- Minimal, safe change that prevents full-history scans for threads missing per-thread lastObservedAt and reduces DB queries/spans and I/O per request.


TylerBarnes (approved):
looks good, thanks! I added a changeset and test

PR #15277, opened 2026-04-12
fix(core): OM multi-step merge must drop memory source when promoting to response

itskrishna21:
## Summary

When Observational Memory clears step-1 assistant output to memory and step-2 text merges into the same assistant row, the message must live in a single source set. If it stayed in `memoryMessages` while also being added to `newResponseMessages`, a later `clear.response.db()` could strand merged text.

## Changes

- `MessageStateManager.addToSource('response')`: remove the message from `memoryMessages` before adding it to the response sets.
- Add a `MessageList` regression test that mirrors the OM handoff (response → memory → merged response → clear).

Fixes #14926

## ELI5
When an AI assistant performs a task in multiple steps, it sometimes needs to save its first response to memory before generating a second response. The bug was that the message could end up stored in two places at once (memory and response), causing the second response to get lost when the system cleaned up. This fix ensures each message only lives in one place at a time, preventing the second response from disappearing.

## Summary
This PR fixes an issue where assistant text produced during multi-step operations with ObservationalMemory would be lost after reload. The problem occurred when a step-1 assistant message was moved from the response source to memory source, and then step-2 text attempted to merge into the same message. Because the message existed in both `memoryMessages` and `newResponseMessages`, a subsequent clear operation would strand the merged content.

## Changes
The fix modifies `MessageStateManager.addToSource()` to check if a message already exists in `memoryMessages` when promoting it to the `'response'` source. If found, the message is removed from `memoryMessages` before being added to the response sets, ensuring a single canonical source location.

## Testing
A new regression test (`message-list-om-multistep.test.ts`) validates the complete handoff sequence: an initial step-1 assistant message is added to response, cleared to memory, then a merged step-2 message is promoted back to response. The test verifies that the merged content persists through the subsequent clear operation and confirms the message is no longer tracked in memory.

## Metadata
- Changeset added for @mastra/core (patch bump)
- Fixes issue #14926


TylerBarnes (approved):
From Mastra Code: This looks good to me.

The state looks reachable through the existing MessageList merge path: a message can be re-added as memory after OM clears/persists step output, then later merged back into the same object as a response message. Without removing the memory source, the object can be tracked in both source sets and later response clearing/persistence can miss the merged content.

This is a narrow fix, main does not appear to already cover it, and the regression test captures the source handoff clearly.

itskrishna21:
@TylerBarnes can this pr be merged if everything looks good?

TylerBarnes:
@itskrishna21 yes definitely! just waiting for ci to pass first. Apologies on the delay btw, my gh notifications were way out of control and I finally got a handle on it

itskrishna21:
@TylerBarnes no worries, glad to be a contributor🫡

PR #15284, opened 2026-04-13
fix(core): remove hardcoded temperature=0 default in agent model settings

### LINKED ISSUE #15240: Remove hardcoded temperature=0 default in prepareCallSettings — breaks models like Kimi K2.5
## Bug Description

`prepareCallSettings()` in `packages/core/src/core/prompt/prepare-call-settings.ts` defaults `temperature` to `0` when not explicitly set:

```typescript
// TODO v5 remove default 0 for temperature
temperature: temperature != null ? temperature : 0,
```

This causes **hard failures** with models that restrict temperature values. For example, Moonshot's **Kimi K2.5** only accepts `temperature=1` and rejects any other value with:

```
400 Bad Request: "invalid temperature: only 1 is allowed for this model"
```

Since most users don't explicitly set temperature (expecting the model's own default), every `agent.stream()` / `agent.generate()` call silently forces `temperature=0`, which breaks these models.

## Reproduction

```typescript

const provider = createOpenAI({
  baseURL: 'https://api.moonshot.cn/v1',
  apiKey: '...',
});

const agent = new Agent({
  name: 'test',
  model: provider.chat('kimi-k2-5'),
});

// Fails: 400 "invalid temperature: only 1 is allowed for this model"
```

## Expected Behavior

When `temperature` is not set by the user, it should be `undefined` so that the AI SDK provider omits it from the request body, letting the model use its own default.

## Proposed Fix

One-line change — the TODO comment already exists in the code:

```diff
- // TODO v5 remove default 0 for temperature
- temperature: temperature != null ? temperature : 0,
+ temperature,
```

This is safe because:
- AI SDK providers (`@ai-sdk/openai`, etc.) correctly strip `undefined` fields from requests
- Users who explicitly set `temperature: 0` still get deterministic behavior
- Only the "didn't set anything" case changes: from forced `0` to model default
- Aligns with how every other parameter (`topP`, `topK`, `presencePenalty`, etc.) is already handled in the same function — none of them have a forced default

## Current Workaround

We intercept the fetch call and strip `temperature: 0` from the request body before sending, which is fragile and shouldn't be necessary.

## Environment

- `@mastra/core`: 1.22.0
- Model: Kimi K2.5 (Moonshot AI) via `@ai-sdk/openai` compatible provider

I'm happy to submit a PR for this if the team is open to it — it's a one-line change.

### LINKED ISSUE #15246: [BUG] Nested workflow runs drop resourceId before persistWorkflowSnapshot
### Describe the Bug

When a parent workflow run is created with resourceId, nested child workflows do not pass that resourceId through to persistWorkflowSnapshot. This causes child workflow snapshots/runs to lose tenant/resource association.
https://github.com/mastra-ai/mastra/blob/067b2250b5aaba9334d06ca3eaa1f230942ce5b3/packages/core/src/workflows/workflow.ts#L2277

### Steps To Reproduce

```
const child = createWorkflow({...}).then(childStep)
child.commit()

const parent = createWorkflow({...}).then(child)
parent.commit()

const run = await parent.createRun({
  runId: 'run-1',
  resourceId: 'workspace-1',
})

await run.start({ inputData: { ok: true } })

```
Observed result:

parent snapshot calls include resourceId: 'workspace-1'
child snapshot calls have resourceId: undefined

https://github.com/amitava82/mastra-demo

### Expected Behavior

If a parent workflow is started with:

```
const run = await workflow.createRun({ runId: 'run-1', resourceId: 'workspace-1' })
```
then nested workflows invoked inside that workflow should also persist snapshots with resourceId: 'workspace-1'.

### Environment Information

```block
@mastra/core: 1.24.1
```

### Verification

octo-patch:
Fixes #15240

## Problem

In `packages/core/src/agent/workflows/prepare-stream/map-results-step.ts`, the `modelSettings` object was built with a hardcoded `temperature: 0` default:

```typescript
modelSettings: {
  temperature: 0,
  ...(options.modelSettings || {}),
},
```

This meant that every `agent.stream()` / `agent.generate()` call silently forced `temperature: 0` when the user did not explicitly set a temperature. This breaks models that only accept specific temperature values — for example, Moonshot's Kimi K2.5 only accepts `temperature=1` and rejects any other value with a `400 Bad Request`.

## Solution

Removed the `temperature: 0` hardcoded default so that when the user does not explicitly set a temperature, no temperature override is injected into the model settings. The user's explicit temperature (if provided via `options.modelSettings.temperature`) is still passed through correctly via the spread.

```typescript
modelSettings: {
  ...(options.modelSettings || {}),
},
```

## Testing

- No existing tests relied on the `temperature: 0` default behavior.
- Users who explicitly set `temperature` in `modelSettings` are unaffected — their value is still passed through.
- Users who do not set `temperature` will now get the model provider's own default instead of a forced `0`.

## ELI5

Imagine you're at a restaurant and always tell the kitchen to cook at zero temperature, even though you never asked for that—it would break everything! This PR fixes that by letting the kitchen (model provider) use its normal cooking temperature when you don't specify one. It also makes sure that when a workflow runs another workflow inside it, both remember which restaurant (tenant) they're working for.

## Overview

This PR addresses two related issues in the core workflow and agent execution logic:

1. **Removes hardcoded temperature default** that was preventing certain models from being used (fixes #15240)
2. **Propagates resourceId to nested workflows** to preserve tenant/resource associations in snapshots (fixes #15246)

### Temperature Default Removal

**File: `packages/core/src/agent/workflows/prepare-stream/map-results-step.ts`**

Removed the hardcoded `temperature: 0` injection from `modelSettings`. Previously, the code always added `temperature: 0` to model settings regardless of user input. Now it only includes temperature if explicitly provided by the user:

- **Before:** `modelSettings: { temperature: 0, ...(options.modelSettings || {}) }`
- **After:** `modelSettings: { ...(options.modelSettings || {}) }`

This allows models that restrict acceptable temperature ranges (e.g., Moonshot Kimi K2.5 which only accepts `temperature=1`) to use their own default values when users don't explicitly set temperature. This aligns temperature handling with other parameters like `topP` and `topK` which already follow this pattern.

### ResourceId Propagation for Nested Workflows

**File: `packages/core/src/workflows/workflow.ts`**

Updated `Workflow.execute()` to accept and forward an optional `resourceId` parameter to `createRun`, ensuring nested workflows inherit the parent workflow's `resourceId`. This preserves the tenant/resource association in workflow snapshots across nested workflow boundaries:

- Added optional `resourceId?: string` parameter to `execute()` method
- Forwards `resourceId` to `createRun()` in both resume and non-resume execution paths

### Documentation Files

**Files: `.changeset/fix-nested-workflow-resource-id.md` and `.changeset/fix-remove-hardcoded-temperature-default.md`**

Added changeset entries documenting both patch-level fixes for `@mastra/core`.

## Impact

- **User-provided temperature values** remain unchanged and are still passed through
- **Model provider defaults** now apply when users don't specify temperature, fixing compatibility with models that have restricted temperature ranges
- **Nested workflows** now maintain proper resource context in snapshots

### DIFF
--- a/.changeset/fix-nested-workflow-resource-id.md
+++ b/.changeset/fix-nested-workflow-resource-id.md
@@ -1,0 +1,5 @@
+---
+"@mastra/core": patch
+---
+
+Fixed nested workflow runs not inheriting `resourceId` from the parent workflow. When a workflow is invoked as a step inside a parent workflow that was created with a `resourceId`, child workflow snapshots are now correctly persisted with the parent's `resourceId`, maintaining tenant/resource association across nested workflows.
--- a/.changeset/fix-remove-hardcoded-temperature-default.md
+++ b/.changeset/fix-remove-hardcoded-temperature-default.md
@@ -1,0 +1,5 @@
+---
+"@mastra/core": patch
+---
+
+Fixed agents incorrectly defaulting `temperature` to `0` when no temperature is explicitly set by the user. Previously, `agent.stream()` and `agent.generate()` always sent `temperature: 0` to the model provider, which broke models that only accept specific temperature values (e.g., Kimi K2.5 which requires `temperature: 1`). Now, when temperature is not explicitly set, the model provider's own default is used instead.
--- a/packages/core/src/agent/workflows/prepare-stream/map-results-step.ts
+++ b/packages/core/src/agent/workflows/prepare-stream/map-results-step.ts
@@ -330,7 +330,6 @@ export function createMapResultsStep<OUTPUT = undefined>({
       inputProcessors: effectiveInputProcessors,
       outputProcessors: effectiveOutputProcessors,
       modelSettings: {
-        temperature: 0,
         ...(options.modelSettings || {}),
       },
       messageList: memoryData.messageList!,
--- a/packages/core/src/workflows/workflow.ts
+++ b/packages/core/src/workflows/workflow.ts
@@ -2193,6 +2193,7 @@ export class Workflow<
   // To run a workflow use `.createRun` and then `.start` or `.resume`
   async execute({
     runId,
+    resourceId,
     inputData,
     resumeData,
     state,
@@ -2215,6 +2216,7 @@ export class Workflow<
     ...rest
   }: {
     runId?: string;
+    resourceId?: string;
     inputData: TInput;
     resumeData?: unknown;
     state: TState;
@@ -2274,7 +2276,9 @@ export class Workflow<
 
     const isTimeTravel = !!(timeTravel && timeTravel.steps.length > 0);
 
-    const run = isResume ? await this.createRun({ runId: resume.runId }) : await this.createRun({ runId });
+    const run = isResume
+      ? await this.createRun({ runId: resume.runId, resourceId })
+      : await this.createRun({ runId, resourceId });
     const nestedAbortCb = () => {
       abort();
     };


coderberry:
Thank you to everyone who contributed to this PR — the fix for removing the hardcoded temperature default is really valuable work! 🙏

I wanted to follow up since this was closed: I'm running into what appears to be the same issue when using `⁠anthropic/opus-4.7`, where the forced default temperature is causing compatibility problems. Could anyone point me to whether an alternative fix has already been merged, or if there's a recommended workaround in the meantime?

Really appreciate all the hard work going into Mastra — thanks so much!

intojhanurag:
> Thank you to everyone who contributed to this PR — the fix for removing the hardcoded temperature default is really valuable work! 🙏
> 
> I wanted to follow up since this was closed: I'm running into what appears to be the same issue when using `⁠anthropic/opus-4.7`, where the forced default temperature is causing compatibility problems. Could anyone point me to whether an alternative fix has already been merged, or if there's a recommended workaround in the meantime?
> 
> Really appreciate all the hard work going into Mastra — thanks so much!

Hey there is not any intention behind this PR to close . It was auto generated by random Bot , so we closed :)
Any PRs are welcome or we will look into it soon !

PR #15285, opened 2026-04-13
fix(agent-browser): initialize browser for default thread ID in thread scope

### LINKED ISSUE #15283: [BUG] @mastra/agent-browser: Browser fails to initialize with thread scope (default)
## Bug Description

When using `@mastra/agent-browser` with the default `scope: 'thread'` configuration, the browser fails to initialize. The `launch()` method completes successfully but `getPage()` throws "Browser not launched. Call launch first."

## Environment

- `@mastra/agent-browser`: v0.1.0
- `agent-browser`: v0.19.0  
- Node.js: $(node --version 2>/dev/null || echo "unknown")
- Platform: macOS (Darwin)
- Chrome installed: Yes

## Steps to Reproduce

```typescript
import { AgentBrowser } from '@mastra/agent-browser';

const browser = new AgentBrowser({ headless: false }); // scope defaults to 'thread'
await browser.launch();
const page = await browser.getPage(); // throws: "Browser not launched. Call launch first."
```

## Root Cause Analysis

In `@mastra/agent-browser`, the `doLaunch()` method (lines 561-580) has this logic:

```typescript
async doLaunch() {
  const scope = this.threadManager.getScope();
  if (scope === "thread") {
    this.sharedManager = new BrowserManager();
    this.threadManager.setSharedManager(this.sharedManager);
    return; // Returns early WITHOUT launching the browser!
  }
  // shared scope continues to launch...
  await this.sharedManager.launch(launchOptions);
}
```

For thread scope, the browser is supposed to launch lazily when a thread session is created via `createSession()`. However, in `AgentBrowserThreadManager.createSession()`:

```typescript
if (this.scope === "thread") {
  // ...
  if (threadId !== DEFAULT_THREAD_ID && !existingSession) {
    await this.getManagerForThread(threadId);  // Only creates if threadId !== DEFAULT_THREAD_ID
  }
}
```

The lazy launch only triggers when `threadId !== DEFAULT_THREAD_ID`. When using the default thread ID, no browser is ever launched.

## Workaround

Use `scope: 'shared'` instead:

```typescript
const browser = new AgentBrowser({ 
  headless: false,
  scope: 'shared'
});
```

## Expected Behavior

Browser should initialize properly with the default thread scope configuration, or the documentation should clearly state that thread scope requires explicit thread ID management.

## Impact

- All users relying on default configuration are affected
- Breaking change for anyone migrating to this package
- No clear error message indicating the configuration issue

Aayush-engineer:
## Summary
Fixes #15283

## Problem
When `AgentBrowser` is used with default `scope: 'thread'` config, 
`launch()` succeeds but `getPage()` throws "Browser not launched".

Two root causes:

1. In `ThreadManager.getManagerForThread()`, `DEFAULT_THREAD_ID` was 
treated the same as `shared` scope — falling back to `getSharedManager()` 
which is just an uninitialized placeholder in thread scope.

2. In `AgentBrowser.ensureReady()`, the session creation guard 
`threadId !== DEFAULT_THREAD_ID` prevented a browser from ever being 
launched for the default thread.

## Fix
**`packages/core/src/browser/thread-manager.ts`** — Remove 
`DEFAULT_THREAD_ID` from the shared scope condition so thread scope 
always creates a dedicated session, even for the default thread.

**`browser/agent-browser/src/agent-browser.ts`** — Remove 
`threadId !== DEFAULT_THREAD_ID` guard so `ensureReady()` correctly 
creates a session for the default thread.

## Changes
- `packages/core/src/browser/thread-manager.ts` — 1 line fix
- `browser/agent-browser/src/agent-browser.ts` — 2 line fix  
- `.changeset/fix-agent-browser-default-thread-scope.md` — patch changeset

## Notes
Existing `scope: 'shared'` behavior is completely unchanged.

### DIFF
--- a/.changeset/fix-agent-browser-default-thread-scope.md
+++ b/.changeset/fix-agent-browser-default-thread-scope.md
@@ -3,4 +3,4 @@
 "@mastra/core": patch
 ---
 
-Fixed AgentBrowser failing to initialize when using default thread scope. Browser now correctly creates a dedicated session for the default thread ID instead of falling back to an uninitialized shared manager.
+AgentBrowser with default thread scope now initializes correctly. Previously, calling launch() followed by getPage() would throw "Browser not launched" when no explicit thread ID was provided.
--- a/.changeset/fix-agent-browser-default-thread-scope.md
+++ b/.changeset/fix-agent-browser-default-thread-scope.md
@@ -1,0 +1,6 @@
+---
+"@mastra/agent-browser": patch
+"@mastra/core": patch
+---
+
+Fixed AgentBrowser failing to initialize when using default thread scope. Browser now correctly creates a dedicated session for the default thread ID instead of falling back to an uninitialized shared manager.
--- a/browser/agent-browser/src/__tests__/thread-manager.test.ts
+++ b/browser/agent-browser/src/__tests__/thread-manager.test.ts
@@ -229,6 +229,32 @@ describe('AgentBrowserThreadManager', () => {
       expect(threadManager.hasActiveThreadManagers()).toBe(false);
     });
 
+    it('creates dedicated session for DEFAULT_THREAD_ID in thread scope', async () => {
+      const threadManager = new AgentBrowserThreadManager({
+        scope: 'thread',
+        browserConfig: { headless: true },
+      });
+
+      // Should create a session for DEFAULT_THREAD_ID without throwing
+      await threadManager.getManagerForThread();
+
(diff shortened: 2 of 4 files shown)


intojhanurag:
@Aayush-engineer pls Add test if you can .

Aayush-engineer:
Added tests for the DEFAULT_THREAD_ID case and updated changeset wording.

NikAiyer (approved):

PR #15299, opened 2026-04-13
fix(cli): allow headless server deploys to use project config

TheIsrael1:
Fix `mastra server deploy` so headless CI can rely on `.mastra-project.json` again, matching the existing deploy contract and docs intent.

## What changed

- Keep the existing headless preflight behavior that accepts org/project from:
  - env vars
  - `--org` / `--project` flags
  - `.mastra-project.json`
- Fix the later org resolution step to trust `projectConfig.organizationId` in headless mode instead of re-fetching org membership first
- Add server deploy tests covering:
  - headless failure when no org/project source is available
  - headless success when `.mastra-project.json` provides org/project
  - no `fetchOrgs()` call when config is used in headless mode

## Why

- `serverDeployAction()` already implied `.mastra-project.json` was sufficient in headless mode
- but `resolveOrg()` could still fail later with `No organizations found`
- this made the server deploy contract internally inconsistent and caused CI failures

## Files changed

- `packages/cli/src/commands/server/deploy.ts`
- `packages/cli/src/commands/server/deploy.test.ts`
- `.changeset/deep-boxes-brush.md`

## Validation

- `pnpm --filter ./packages/cli test -- src/commands/server/deploy.test.ts src/commands/studio/deploy.test.ts`
- `pnpm --filter ./packages/cli typecheck`

## ELI5

This PR fixes a problem where the server deploy command wasn't working in automated environments (like CI/CD pipelines). The fix lets the deploy command use organization and project information that's already saved in a config file (`.mastra-project.json`) instead of trying to look it up online, which was failing. The PR also adds tests to make sure this works correctly.

## Overview

This PR restores support for headless CI deployments when using `mastra server deploy` with a `.mastra-project.json` configuration file. Previously, headless deployments would fail when trying to resolve organization information, even when the org and project were already specified in the project config.

### Modified: `packages/cli/src/commands/server/deploy.ts`
The `resolveOrg` function now detects headless mode (by checking for `MASTRA_API_TOKEN` environment variable) and, when a `projectConfig.organizationId` is available, returns the organization immediately without attempting to fetch organization data via the API. This prevents unnecessary API calls and eliminates failures that could occur when `fetchOrgs()` returns no results in headless environments.

### Added: `packages/cli/src/commands/server/deploy.test.ts`
New test suite covering headless deployment scenarios:
- Validates that the action fails appropriately when running headless without org/project configuration
- Confirms the action succeeds when organization and project are sourced from `.mastra-project.json`
- Verifies that `fetchOrgs()` is not called when project config is available, avoiding unnecessary network calls

### Added: `.changeset/deep-boxes-brush.md`
Changelog entry documenting this patch-level fix for headless server deployments.

## Behavior

The fix maintains backward compatibility with existing deployment configuration sources (environment variables and command-line flags) while now properly trusting the `.mastra-project.json` configuration in headless CI environments.

### DIFF
--- a/.changeset/deep-boxes-brush.md
+++ b/.changeset/deep-boxes-brush.md
@@ -1,0 +1,5 @@
+---
+'mastra': patch
+---
+
+Fixed headless server deploys so they can use .mastra-project.json for org and project resolution in CI.
--- a/packages/cli/src/commands/server/deploy.test.ts
+++ b/packages/cli/src/commands/server/deploy.test.ts
@@ -1,0 +1,140 @@
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
+
+vi.mock('node:child_process', () => ({
+  execSync: vi.fn().mockReturnValue('my-app'),
+}));
+
+let closeHandler: (() => void) | undefined;
+
+vi.mock('node:fs', () => ({
+  createWriteStream: vi.fn(() => ({
+    on: vi.fn((event: string, callback: () => void) => {
+      if (event === 'close') {
+        closeHandler = callback;
+      }
+    }),
+  })),
+}));
+
+vi.mock('node:fs/promises', () => ({
+  mkdir: vi.fn().mockResolvedValue(undefined),
+  rm: vi.fn().mockResolvedValue(undefined),
+  stat: vi.fn().mockResolvedValue({ size: 1024 }),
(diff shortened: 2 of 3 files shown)


wardpeet (approved):

intojhanurag:
Related to #15321

PR #15315, opened 2026-04-13
feat(mcp): add requireToolApproval to MCP server configuration

roaminro:
Adds a `requireToolApproval` option to MCP server definitions so you can require human approval for MCP tool calls at the server level.

Supports a boolean for blanket approval, or a function for dynamic per-tool logic:

```ts
const mcp = new MCPClient({
  servers: {
    github: {
      url: new URL('http://localhost:3000/mcp'),
      requireToolApproval: ({ toolName, args, requestContext }) => {
        if (toolName === 'list_repos') return false
        return requestContext?.userRole !== 'admin'
      },
    },
  },
})
```

This hooks into the existing human-in-the-loop approval flow — no changes to the core Tool class or execution pipeline. The MCP client sets `requireApproval` and `needsApprovalFn` on tool instances directly (same pattern as the AI SDK tool-builder).

## ELI5

This PR adds a server-level "ask a human first" setting for MCP servers so tool calls can require human approval; it can apply to all tools or decide per-call with a function.

---

## Summary

Adds requireToolApproval to MCP server definitions in @mastra/mcp to enable server-level human approval gating for MCP tool calls. The option accepts:
- a boolean: true requires approval for all tools, false/omitted means no approval; or
- a function (sync or async) called per invocation with { toolName, args, requestContext } that returns a boolean to decide whether approval is required.

The MCP client translates server-level requireToolApproval into per-tool properties (requireApproval and an attached needsApprovalFn) so the existing human-in-the-loop approval flow is reused without changing the Tool class or core execution pipeline.

### Key Changes

- Types
  - New types in packages/mcp/src/client/types.ts:
    - RequireToolApprovalContext
    - RequireToolApprovalFn
    - RequireToolApproval (boolean | function)
  - BaseServerOptions now includes requireToolApproval?: RequireToolApproval.

- MCP Client
  - InternalMastraMCPClient stores server.requireToolApproval and computes per-tool approval settings when creating tools.
  - When requireToolApproval is a function, the client wraps it into a needsApprovalFn that forwards { toolName, args, requestContext } (defaults requestContext to {} when absent) and sets requireApproval = true.
  - When requireToolApproval is true, requireApproval is set to true with no custom needsApprovalFn.
  - When unset/false, requireApproval is left undefined so tools default to no approval enforcement.
  - The computed requireApproval is passed into createTool, and the computed needsApprovalFn (when present) is attached to the tool instance for downstream use.

- Execution Flow Integration
  - Wires server-level config into existing approval suspension/resume logic. Tool-call step tests exercise behavior when needsApprovalFn is present (suspends on throw, proceeds when it returns false).

- Documentation & Changeset
  - docs updated: new "Tool approval" section in MCP overview and MCPClient reference with examples for static and dynamic configuration.
  - Added changeset documenting the new option.

- Tests
  - New tests in packages/mcp/src/client/client.test.ts covering boolean and function variants (including async) and verifying needsApprovalFn is set and gets correct context.
  - Extended core tool-call-step tests to verify behavior when needsApprovalFn is present.

### Commit / Implementation notes

- Small robustness/follow-up fixes: the wrapper defaults ctx/requestContext to {} to support network-path callers, added comments documenting the ctx limitation and false/undefined fall-through behavior, and simplified the tool-creation wiring to use a plain requireApproval property.
- No changes to the public Tool class or core execution pipeline APIs; changes are types, client wiring, docs, and tests.

### DIFF
--- a/packages/mcp/src/client/client.ts
+++ b/packages/mcp/src/client/client.ts
@@ -687,17 +687,20 @@ export class InternalMastraMCPClient extends MastraBase {
         let needsApprovalFn: ((args: any, ctx: any) => boolean | Promise<boolean>) | undefined;
 
         if (typeof this.requireToolApproval === 'function') {
-          // Wrap the server-level function to match the per-tool needsApprovalFn signature
-          // (args, ctx) => the server function receives { toolName, args, ...ctx }
+          // Wrap the server-level function to match the per-tool needsApprovalFn signature.
+          // Note: ctx may be undefined when called via network/index.ts (which only passes args).
+          // We default ctx to {} so the spread doesn't fail and approval fn receives partial context.
           const serverApprovalFn = this.requireToolApproval;
           const toolName = tool.name;
           requireApproval = true; // Signal that approval check is needed
-          needsApprovalFn = (args: Record<string, unknown>, ctx: Record<string, unknown>) => {
+          needsApprovalFn = (args: Record<string, unknown>, ctx: Record<string, unknown> = {}) => {
             return serverApprovalFn({ toolName, args, ...ctx });
           };
         } else if (this.requireToolApproval === true) {
           requireApproval = true;
         }
+        // When requireToolApproval is false/undefined, requireApproval stays undefined
+        // and createTool defaults it to false
 
         const mastraTool = createTool({
           id: `${this.name}_${tool.name}`,
@@ -707,7 +710,7 @@ export class InternalMastraMCPClient extends MastraBase {
           // already validates structuredContent against the tool's outputSchema using AJV.
           // Passing it here causes Zod to strip unrecognized keys from the CallToolResult
           // envelope, returning {} for tools without structuredContent.
-          ...(requireApproval !== undefined ? { requireApproval } : {}),
+          requireApproval,
           mcpMetadata: {
             serverName: this.name,
             serverVersion: this.client.getServerVersion()?.version,
--- a/packages/mcp/src/client/client.ts
+++ b/packages/mcp/src/client/client.ts
@@ -54,6 +54,7 @@ import type {
   MastraMCPServerDefinition,
   InternalMastraMCPClientOptions,
   Root,
+  RequireToolApproval,
 } from './types';
 
 // Re-export types for convenience
@@ -67,6 +68,9 @@ export type {
   MastraMCPServerDefinition,
   InternalMastraMCPClientOptions,
   Root,
+  RequireToolApproval,
+  RequireToolApprovalFn,
+  RequireToolApprovalContext,
 } from './types';
 
 const DEFAULT_SERVER_CONNECT_TIMEOUT_MSEC = 3000;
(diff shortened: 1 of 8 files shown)


packages/mcp/src/client/client.ts
```diff
@@ -676,6 +682,23 @@ export class InternalMastraMCPClient extends MastraBase {
...
       try {
+        // Resolve requireToolApproval for this tool
+        let requireApproval: boolean | undefined;
+        let needsApprovalFn: ((args: any, ctx: any) => boolean | Promise<boolean>) | undefined;
+
+        if (typeof this.requireToolApproval === 'function') {
+          // Wrap the server-level function to match the per-tool needsApprovalFn signature
+          // (args, ctx) => the server function receives { toolName, args, ...ctx }
+          const serverApprovalFn = this.requireToolApproval;
+          const toolName = tool.name;
```
wardpeet, line 694: The `ctx` param will be `undefined` when called from `network/index.ts` (line 1536), which only passes one argument: `needsApprovalFn(inputDataToUse)`. The spread `...undefined` is a JS no-op so it won't crash, but `requestContext` will silently never reach the server-level approval function through the network execution path.

Consider defaulting `ctx` to `{}`:
```ts
needsApprovalFn = (args: Record<string, unknown>, ctx: Record<string, unknown> = {}) => {
```

Or at minimum document this as a known limitation — the dynamic approval function won't receive `requestContext` when tools execute via the agent network path.

---
✅ Resolved: Added default `ctx = {}` and documented the limitation.

packages/mcp/src/client/client.ts
```diff
@@ -684,6 +707,7 @@ export class InternalMastraMCPClient extends MastraBase {
           // already validates structuredContent against the tool's outputSchema using AJV.
           // Passing it here causes Zod to strip unrecognized keys from the CallToolResult
           // envelope, returning {} for tools without structuredContent.
+          ...(requireApproval !== undefined ? { requireApproval } : {}),
```
wardpeet: Nit: The conditional spread isn't strictly necessary — passing `requireApproval: undefined` to `createTool()` produces the same result since the `Tool` constructor does `opts.requireApproval || false`. A plain `requireApproval,` would be simpler and equivalent.

---
✅ Resolved: Simplified to plain `requireApproval,`

packages/mcp/src/client/client.ts
```diff
@@ -676,6 +682,23 @@ export class InternalMastraMCPClient extends MastraBase {
...
+        if (typeof this.requireToolApproval === 'function') {
+          // Wrap the server-level function to match the per-tool needsApprovalFn signature
+          // (args, ctx) => the server function receives { toolName, args, ...ctx }
+          const serverApprovalFn = this.requireToolApproval;
+          const toolName = tool.name;
+          requireApproval = true; // Signal that approval check is needed
+          needsApprovalFn = (args: Record<string, unknown>, ctx: Record<string, unknown>) => {
+            return serverApprovalFn({ toolName, args, ...ctx });
+          };
+        } else if (this.requireToolApproval === true) {
```
wardpeet, line 699: When `requireToolApproval` is explicitly `false`, we fall through both branches and `requireApproval` stays `undefined`. This works because `createTool` defaults it to `false`, but it's a bit subtle. Worth either an explicit `else` branch or a brief comment noting the intentional fall-through — makes the exhaustive handling of all three cases (function / true / false|undefined) obvious to the next reader.

---
✅ Resolved: Added comment explaining the intentional fall-through.

wardpeet (commented):
Nice work — clean types and good test coverage. A few things I noticed:

wardpeet (approved):
Small nitpicks

PR #15318, opened 2026-04-13
feat(core): enable structuredOutput agent to re-use the parent agent config

TylerBarnes:
With this change users can now opt into parent-agent reuse for the structuring pass with `useAgent: true`:

```ts
agent.stream(prompt, {
  memory: { thread, resource },
  structuredOutput: { schema, model: 'openai/gpt-5.4-mini', useAgent: true },
});
```

That enables the separate structuring model to use the same agent, allowing it to also use agent memory, while keeping the main response stream clean (by making memory read only in the structuring agent). In a breaking change we should make this the default behaviour but for now it's opt-in.

Docs and regression coverage are included for the streaming, memory, and opt-in agent reuse behavior.

## ELI5

This PR adds an opt-in switch so the structured-output step can reuse the main agent and its conversation memory (read-only) when producing structured data, making the structured output aware of prior conversation context when desired.

## Summary of changes

Adds StructuredOutputOptionsBase.useAgent?: boolean to let callers opt a separate structuredOutput.model into reusing the parent Agent (with read-only memory routing from thread/resource when available). Leaving useAgent unset preserves the prior behavior where the structuring pass does not inherit conversation memory.

Key code changes
- Types
  - Added useAgent?: boolean to StructuredOutputOptionsBase.
  - Made AgentMemoryOption.resource optional.
- Agent capabilities / wiring
  - AgentCapabilities now includes agent: Agent.
  - Agent.#execute populates capabilities.agent so processors can access the running Agent instance.
- StructuredOutputProcessor
  - Stores structuring model, structuring instructions, and the useAgent flag.
  - Added setAgent(agent) to inject the parent Agent for the structuring pass.
  - processOutputStream/processAndEmitStructuredOutput now accept requestContext and messageList so memoryInfo can be derived.
  - getStructuringStream conditional behavior:
    - If useAgent is true, an injected agent exists, and a thread id is available → stream via the injected agent with a model override and memory.options.readOnly = true (memory routed from requestContext or messageList.serialize().memoryInfo).
    - Otherwise → fall back to the internal structuring agent (no inherited conversation memory).
  - Guarded controller.enqueue calls and unified abort typing for streaming.
- RequestContext forwarding
  - Forward active request context into the nested structuring agent re-entry so per-request values (thread/resource) are preserved for memory routing.
- Prompt injection / execution changes
  - Processor-mode JSON-schema prompt injection is skipped when structuredOutput.useAgent is true (only injected when structuredOutputMode === 'processor' && responseFormat.type === 'json' && responseFormat.schema && !structuredOutput?.useAgent), preventing processor-inserted schema instructions from being merged into the main agent prompt when reusing the agent.
- Persistence / JSON echo handling
  - Ensures JSON-only structured-output chunks are treated by the object stream transformer so persisted assistant messages retain natural-language text and avoid persisting raw JSON-echo chunks.

Documentation
- Updated docs to explain useAgent behavior, requirement to pair it with structuredOutput.model for a separate structuring model, and included TypeScript examples demonstrating memory.thread/memory.resource with useAgent: true. Clarified default behavior when useAgent is unset.

Tests
- Added/updated tests covering:
  - execute structured-output prompt handling: verifies prompt differs when useAgent is enabled vs omitted.
  - StructuredOutputProcessor unit tests: assert injected agent usage, memory derived from requestContext/messageList, and memory.options.readOnly across scenarios (thread+resource, thread-only, fallback).
  - Agent-level and e2e tests (including structured-output-memory.e2e.test.ts): validate structured output memory inheritance across turns, streaming behavior (text-delta vs object-result), and that JSON-only structured-output does not persist as assistant text.
  - stream-id and related tests assert persisted assistant text handling when structured output returns parsed objects only.

Commits / changesets
- Added changesets documenting the opt-in useAgent behavior for @mastra/core and JSON persistence behavior for @mastra/ai-sdk.

Behavioral notes
- The feature is opt-in (useAgent must be explicitly enabled). Making it the default would be breaking and is deferred.
- When enabled, the structuring agent sees the same per-request context but with memory routed read-only so the main response stream and persisted assistant text remain unchanged.


packages/core/src/agent/__tests__/structured-output-memory.e2e.test.ts
```diff
@@ -0,0 +1,124 @@
+import { randomUUID } from 'node:crypto';
+import { createGatewayMock } from '@internal/test-utils';
+import { config } from 'dotenv';
+import { afterAll, beforeAll, describe, expect, it, vi } from 'vitest';
+import { z } from 'zod/v4';
+import { MockMemory } from '../../memory/mock';
+import { Agent } from '../agent';
+
+config();
```
wardpeet: lets not do dotenv in tests
```suggestion
import { afterAll, beforeAll, describe, expect, it, vi } from 'vitest';
import { z } from 'zod/v4';
import { MockMemory } from '../../memory/mock';
import { Agent } from '../agent';
```

packages/core/src/agent/__tests__/structured-output-memory.e2e.test.ts
```diff
@@ -0,0 +1,124 @@
...
+
+const mock = createGatewayMock();
```
wardpeet, line 12: Also add setupDummyKeys to credate a dummy OPEN_API_KEY  so recordings are ran

packages/core/src/agent/__tests__/structured-output-memory.e2e.test.ts
```diff
@@ -0,0 +1,124 @@
...
+afterAll(() => mock.saveAndStop());
+
+const profileSchema = z.object({
+  favoriteColor: z.string(),
+  hometown: z.string(),
+  petName: z.string(),
+});
+
+describe('Structured output memory inheritance (e2e)', { timeout: 180_000 }, () => {
+  it.skipIf(!process.env.OPENAI_API_KEY)(
```
wardpeet: we shouldn't do skips
```suggestion
  it(
```

wardpeet (approved):
small test changes

packages/core/src/agent/agent.ts
```diff
@@ -4574,6 +4574,7 @@ export class Agent<
 
     // Create a capabilities object with bound methods
     const capabilities = {
+      agent: this,
```
DanielSLew, line 4602: hah, if we're passing the whole agent, we kind of don't need this capabilities obj anymore, but that's a bigger re-factor.

packages/core/src/processors/processors/structured-output.ts
```diff
@@ -174,13 +182,70 @@ export class StructuredOutputProcessor<OUTPUT extends {}> implements Processor<'
...
+      return this.agent.stream(prompt, {
+        model: this.structuringModel,
+        structuredOutput: {
+          schema: this.schema,
+          jsonPromptInjection: this.jsonPromptInjection,
+        },
+        memory: {
+          thread: threadId,
+          ...(resourceId ? { resource: resourceId } : {}),
+          options: { readOnly: true },
```
DanielSLew, line 235: I wonder if some people might have their own memory implementation and this might start adding memory messages to their system because they don't acknowledge the readOnly param.

packages/core/src/processors/processors/structured-output.ts
```diff
@@ -174,13 +182,70 @@ export class StructuredOutputProcessor<OUTPUT extends {}> implements Processor<'
...
+      typeof requestResourceId === 'string'
+        ? requestResourceId
+        : typeof serializedMemoryInfo?.resourceId === 'string'
+          ? serializedMemoryInfo.resourceId
+          : undefined;
+
+    // When opted in and an explicit agent is available with thread info,
+    // use that agent with a model override and read-only memory.
+    // This gives the structuring model full conversation context.
+    if (this.useAgent && this.agent && threadId) {
```
DanielSLew, line 223: I think we need to guard here for `this.structuringModel` too right? Otherwise they might have passed useAgent but not passed a model and this will throw i think?

.changeset/six-times-start.md
```diff
@@ -0,0 +1,6 @@
+---
+'@mastra/ai-sdk': minor
+'@mastra/core': minor
+---
+
+You can now opt into parent-agent reuse for the structuring agent pass with `structuredOutput: { schema, model, useAgent: true }`, which lets the separate structuring request re-use the parent agents config, including memory.
```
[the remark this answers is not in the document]
TylerBarnes: Fixed in e0f530aa02. I split the release notes so the existing changeset is core-only and added a separate ai-sdk entry for the direct structured-output streaming change. Reese says: thanks, good catch.

packages/core/src/processors/processors/structured-output.ts
TylerBarnes: I did not add an extra branch guard here. The processor already throws in its constructor when model is missing, and the map-results-step path only instantiates it when structuredOutput.model is present. I did update the docs and type comment in e0f530aa02 to make that contract explicit: useAgent only applies to the separate structuring pass when model is also provided. Reese says: thanks.

packages/core/src/processors/processors/structured-output.ts
```diff
@@ -174,13 +182,70 @@ export class StructuredOutputProcessor<OUTPUT extends {}> implements Processor<'
...
+        },
+        memory: {
+          thread: threadId,
+          ...(resourceId ? { resource: resourceId } : {}),
+          options: { readOnly: true },
+        },
+        providerOptions: this.providerOptions,
+        ...observabilityContext,
+      });
+    }
```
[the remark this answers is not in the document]
TylerBarnes: I intentionally kept the thread gate here. useAgent is only meant to opt into parent-agent reuse for the separate structuring pass when we can also carry over the request memory context. Without a thread, reusing the parent agent would change behavior beyond the scoped memory fix this PR is shipping. I clarified that contract in docs/types in e0f530aa02 so stateless runs are not implied to reuse the parent agent. Reese says: thanks.

packages/core/src/agent/__tests__/stream-id.test.ts
```diff
@@ -527,6 +527,76 @@ describe('Stream ID Consistency', () => {
...
+        const recalled = await memory.recall({ threadId, resourceId });
+        const assistantMessage = [...recalled.messages].reverse().find(message => message.role === 'assistant');
+        const assistantText =
+          assistantMessage?.content.parts
+            .filter(part => part.type === 'text')
+            .map(part => part.text)
+            .join('') ?? '';
+
+        expect(assistantText).toBe('');
+      });
```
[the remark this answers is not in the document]
TylerBarnes: I tried the suggested hardening, but this test can legitimately persist no assistant message at all for the object-only case, so requiring one made the focused test fail while the behavior stayed correct. I kept the assertion on the persisted assistant text instead. Reese says: thanks.

packages/core/src/processors/processors/structured-output.ts
TylerBarnes: I left this as-is. The separate pass is using the normal readOnly memory contract that other agent flows already rely on; if a custom memory implementation ignores readOnly, that backend is already out of contract. I did not want to add provider-specific branching in this fix. Reese says: thanks.

packages/core/src/agent/agent.ts
TylerBarnes: Agreed. Passing the full agent makes parts of capabilities feel redundant now, but I kept this PR surgical and did not fold that refactor into the structured-output fix. Reese says: thanks.

TylerBarnes:
Addressed the remaining small review items on this branch:\n- split the changesets by package-facing behavior\n- clarified in docs/types that useAgent applies to the separate structuring pass when model is provided\n- replied inline on the comments I intentionally left unchanged\n\nFocused verification:\n- pnpm exec vitest run src/agent/__tests__/stream-id.test.ts src/processors/processors/structured-output.test.ts --bail 1 --reporter=dot\n\nNote: pnpm check in packages/core still hits the pre-existing unrelated tool-builder error in src/tools/tool-builder/builder.ts:666 about processToCompatSchema/processToJSONSchema.\n\nWould appreciate another look. Reese says:

PR #15335, opened 2026-04-14
fix(channels): support serverless cold starts for webhook handlers

### LINKED ISSUE #15300: Channels don't work on Vercel serverless — AgentChannels.initialize() is fire-and-forget
## Description

When deploying a Mastra application with Channels (Slack adapter) to **Vercel using `@mastra/deployer-vercel`**, every webhook request returns **503 `{"error":"Chat not initialized"}`**. The same configuration works perfectly in local development (`mastra dev`).

## Root Cause

In `@mastra/core`, the `Mastra` constructor calls `AgentChannels.initialize()` with `void` (fire-and-forget):

```js
// @mastra/core — Mastra constructor (addAgent method)
void agentChannels.initialize(this);  // line ~609
```

In a **long-running server** (local dev, Railway, etc.), this is fine — the async init completes before the first request arrives.

In **Vercel serverless**, each cold start creates a new instance. The first request arrives **before** `initialize()` completes, so `this.chat` is still `null` when the webhook handler checks it:

```js
// AgentChannels.getWebhookRoutes()
const sdkInstance = self.chat;
if (!sdkInstance) {
  return c.json({ error: "Chat not initialized" }, 503);  // ← always hits this
}
```

## Additional Issue: Storage init also fails

`LibSQLStore.init()` (creating tables) runs during module import and fails in serverless with:

```
Unhandled Rejection: Error: fetch failed
    at LibSQLDB.createTable (...)
```

This compounds the problem — even if Channels waited, the storage isn't ready either.

## Environment

- `@mastra/core`: 1.24.1
- `@mastra/deployer-vercel`: 1.1.16
- `@chat-adapter/slack`: 4.25.0
- `@mastra/libsql`: 1.8.0 (Turso remote URL)
- Vercel serverless function (Node.js)
- Config: `gateway: false`, `maxDuration: 300`, `memory: 1024`

## Reproduction

1. Create agents with `channels.adapters.slack` using `createSlackAdapter()`
2. Add `storage` (LibSQLStore with remote Turso URL) to Mastra instance
3. Add `deployer: new VercelDeployer()` to Mastra config
4. Deploy to Vercel via `mastra build`
5. Send POST to `/api/agents/{agentId}/channels/slack/webhook` with Slack `url_verification` payload
6. Response: `{"error":"Chat not initialized"}` (503)

## Workarounds Attempted

1. **`gateway: false`** on adapter — didn't help (already webhook-only)
2. **Middleware polling** — added Hono middleware that waits up to 10s for `channels.sdk` to be non-null. Didn't help because storage init also fails before channels can initialize
3. **Increased `maxDuration` to 300s and `memory` to 1024MB** — didn't help, issue is not timeout

## Expected Behavior

Channels should work in Vercel serverless. Either:
- `AgentChannels.initialize()` should be `await`-ed (not `void`) so the server is ready before handling requests
- Or there should be a documented pattern for lazy initialization in serverless environments

## Questions

1. Is there an official way to use Channels with `@mastra/deployer-vercel`?
2. Should we use a different deployment strategy for Channels (e.g., long-running server only)?
3. Is there a planned fix to make initialization synchronous/awaited for serverless compatibility?

Any guidance would be appreciated. Happy to provide more details or test proposed fixes.

intojhanurag:
## Description

Fixes Channels (Slack, Discord, etc.) failing on Vercel and other serverless platforms due to a race condition where webhook requests arrive before `AgentChannels.initialize()` completes, and background agent processing being killed after the HTTP response is sent.

- Store the initialization promise on `AgentChannels` via new `__startInitialize()` method so webhook handlers can await readiness instead of immediately returning 503
- Pass the platform's `executionCtx.waitUntil` from Hono context to the Chat SDK webhook handler so agent processing survives after the serverless function responds
- Replace fire-and-forget `void agentChannels.initialize(this)` with `agentChannels.__startInitialize(this)` in `Mastra.addAgent()`
- No behavior change for long-running servers — the init promise resolves instantly since initialization completes before any request arrives

## Related Issue(s)

Fixes #15300

## Type of Change

- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [ ] Test update

## Checklist

## ELI5

This PR fixes a timing problem where webhook requests arrive at serverless platforms (like Vercel) before the chat system is fully set up, causing them to fail with an error. The solution makes the initialization process "awaitable" so handlers can wait for it to complete, and also ensures background work isn't stopped after the HTTP response is sent.

### Changeset (.changeset/vast-masks-juggle.md)
- Added a patch-level release note for `@mastra/core`
- Documents that Channels functionality is now fixed for Vercel serverless and other serverless platforms
- Notes that webhook handlers now await initialization instead of returning 503 errors on cold starts
- Documents that webhook handlers now pass the platform's `waitUntil` capability to the Chat SDK for continued background processing

### AgentChannels (packages/core/src/channels/agent-channels.ts)

**Initialization improvements:**
- Added `initPromise` field to store the in-flight initialization promise, making initialization idempotent and memoized
- Refactored `initialize()` method to return the stored promise when called multiple times
- Implemented initialization logic as an async IIFE that clears the promise on failure and rethrows the error
- Early-returns when `this.chat` is already initialized

**Webhook handler improvements:**
- Updated request handlers in `getWebhookRoutes()` to await `initPromise` before using `this.chat`
- Returns a 503 JSON response if initialization fails
- Modified webhook handler invocation to forward the serverless platform's execution context: passes `c.executionCtx.waitUntil` as options to the Chat SDK webhook handler when available
- Enables background agent processing to continue after the HTTP response is sent in serverless environments

## Impact

- Eliminates the race condition where concurrent cold-start requests could return 503 errors
- Allows multiple concurrent requests to share a single initialization promise
- Enables Channels to work reliably on serverless platforms by preventing premature function termination of background jobs
- Addresses issue #15300

### DIFF
--- a/packages/core/src/channels/agent-channels.ts
+++ b/packages/core/src/channels/agent-channels.ts
@@ -343,6 +343,8 @@ async function headContentType(url: string, logger?: IMastraLogger): Promise<str
 export class AgentChannels {
   readonly adapters: Record<string, Adapter>;
   private chat: Chat | null = null;
+  /** Stored initialization promise so webhook handlers can await readiness on serverless cold starts. */
+  private initPromise: Promise<void> | null = null;
   private agent!: Agent<any, any, any, any>;
   private logger?: IMastraLogger;
   private customState: StateAdapter | undefined;
@@ -409,6 +411,19 @@ export class AgentChannels {
   __setLogger(logger: IMastraLogger): void {
     this.logger =
       'child' in logger && typeof (logger as any).child === 'function' ? (logger as any).child('CHANNEL') : logger;
+  }
+
+  /**
+   * Begin initialization and store the promise so webhook handlers can
+   * await readiness on serverless cold starts instead of returning 503.
+   * @internal Called by Mastra.addAgent.
+   */
+  __startInitialize(mastra: Mastra): void {
+    const promise = this.initialize(mastra);
+    // Attach a no-op catch to prevent unhandled rejection warnings.
+    // Errors are re-surfaced when the webhook handler awaits initPromise.
+    promise.catch(() => {});
+    this.initPromise = promise;
   }
 
   /**
@@ -679,6 +694,16 @@ export class AgentChannels {
         requiresAuth: false,
         createHandler: async () => {
           return async c => {
+            // Await initialization to handle serverless cold starts where
+            // the first request arrives before initialize() completes.
+            if (self.initPromise) {
+              try {
+                await self.initPromise;
+              } catch {
+                return c.json({ error: 'Chat initialization failed' }, 503);
+              }
+            }
+
             const sdkInstance = self.chat;
             if (!sdkInstance) {
               return c.json({ error: 'Chat not initialized' }, 503);
@@ -688,7 +713,12 @@ export class AgentChannels {
             if (!webhookHandler) {
               return c.json({ error: `No webhook handler for ${platform}` }, 404);
             }
-            return webhookHandler(c.req.raw);
+
+            // Pass platform execution context (e.g. Vercel/Cloudflare waitUntil)
+            // to the Chat SDK so background processing survives serverless responses.
+            const execCtx = c.executionCtx as { waitUntil?: (p: Promise<unknown>) => void } | undefined;
+            const waitUntilFn = execCtx?.waitUntil?.bind(execCtx);
+            return webhookHandler(c.req.raw, waitUntilFn ? { waitUntil: waitUntilFn } : undefined);
           };
         },
       });
--- a/packages/core/src/channels/agent-channels.ts
+++ b/packages/core/src/channels/agent-channels.ts
@@ -414,19 +414,6 @@ export class AgentChannels {
   }
 
   /**
-   * Begin initialization and store the promise so webhook handlers can
-   * await readiness on serverless cold starts instead of returning 503.
-   * @internal Called by Mastra.addAgent.
-   */
-  __startInitialize(mastra: Mastra): void {
-    const promise = this.initialize(mastra);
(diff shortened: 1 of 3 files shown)


wardpeet:
I"m making a couple of changes so it's al inside the initialize fn

packages/core/src/channels/agent-channels.ts
```diff
@@ -432,241 +434,258 @@ export class AgentChannels {
...
   /**
    * Returns API routes for receiving webhook events from each adapter.
    * One POST route per adapter at `/api/agents/{agentId}/channels/{platform}/webhook`.
    */
-  getWebhookRoutes(): ApiRoute[] {
+  getWebhookRoutes(mastra?: Mastra): ApiRoute[] {
     if (!this.agent) return [];
+    if (mastra) {
+      this.initialize(mastra).catch(() => {});
+    }
```
wardpeet:
```suggestion
  getWebhookRoutes(): ApiRoute[] {
if (!this.agent) return [];
```

packages/core/src/mastra/index.ts
```diff
@@ -1088,14 +1088,13 @@ export class Mastra<
     const agentChannels = mastraAgent.getChannels();
     if (agentChannels) {
       agentChannels.__setLogger(this.#logger);
-      const channelRoutes = agentChannels.getWebhookRoutes();
+      const channelRoutes = agentChannels.getWebhookRoutes(this);
```
wardpeet:
```suggestion
      const channelRoutes = agentChannels.getWebhookRoutes();
```

wardpeet (approved):

intojhanurag:
Thanks @wardpeet for working on this !

PR #15343, opened 2026-04-14
Fix Sentry error stack traces to point to actual error origin

### LINKED ISSUE #15337: Mastra Sentry bug - wrong stacktrace
This issue was created from Discord post 1493572073329655988:

[Open in Discord](https://discord.com/channels/1309558646228779139/1493572073329655988)

Bug: @mastra/sentry passes string to Sentry.captureException — stack traces always point to exporter, never to application code

Hey, we've been setting up Sentry with @mastra/sentry@1.0.15 and found that all error stack traces in Sentry point to SentryExporter.handleSpanEnded instead of our actual workflow code.

The issue:

In @mastra/sentry/dist/index.js line 193:

Sentry.captureException(span.errorInfo.message, { ... })
span.errorInfo.message is a string. When Sentry.captureException receives a string, it creates a synthetic stack trace from the current call site — which is handleSpanEnded inside the exporter. So every error in Sentry looks like this:

Error: Our workflow error message
    at SentryExporter.handleSpanEnded (node_modules/@mastra/sentry/dist/index.js:193:14)
    at SentryExporter._exportTracingEvent (node_modules/@mastra/sentry/dist/index.js:115:20)
What we verified:

@mastra/observability already stores the original error.stack in span.errorInfo.stack (line 19378)
But the sentry exporter never uses it — it only reads .message
We also found that @mastra/core wraps thrown errors in MastraError, and observability stores the wrapper's stack (error.stack) instead of the original cause stack (error.cause.stack)
Suggested fix in @mastra/sentry:

// Instead of:
Sentry.captureException(span.errorInfo.message, { ... })
// Use:
const err = new Error(span.errorInfo.message);
if (span.errorInfo.stack) err.stack = span.errorInfo.stack;
Sentry.captureException(err, { ... })
And in @mastra/observability (error method):

// Instead of:
stack: error48.stack
// Use the original error's stack when available:
stack: error48.cause?.stack || error48.stack
Our setup:

@mastra/sentry@1.0.15, @mastra/observability@1.9.0, @mastra/core@1.5.0
Vanilla config, no custom hooks
Also enabled bundler: { sourcemap: true } — didn't help since the stack itself is wrong
Is this a known issue? Is there a fix planned or a workaround we're missing?

epinzur:
## Description

Fixed Sentry error stack traces so they point to the actual error origin instead of the Mastra exporter internals.

Two related fixes:

1. **@mastra/sentry**: The exporter now passes an `Error` object (preserving the captured stack) to `Sentry.captureException` instead of a bare message string. Previously, passing a string caused Sentry to synthesize a stack trace from the exporter's call site, so every error in Sentry appeared to originate from `SentryExporter.handleSpanEnded`.

2. **@mastra/observability**: When a span is failed with a `MastraError` that wraps another error, `span.errorInfo.stack` now prefers the original cause's stack over the wrapper's stack, so downstream exporters receive the stack pointing to the true error origin.

## Related Issue(s)

Fixes #15337

## Type of Change

## Checklist

- [x] Tests added to verify stack trace handling in both observability and Sentry exporter

https://claude.ai/code/session_01EcjkUuaWpXuwA3J9NAdEF7

## ELI5

When Mastra caught errors and reported them to Sentry (an error tracking service), Sentry was showing that the error came from deep inside Mastra's code instead of from the actual place where the error originally happened. This PR fixes that by properly passing the original error information to Sentry, so developers can see where problems actually started.

## Overview

This PR fixes Sentry error stack traces to point to the original error origin rather than Mastra exporter internals, addressing issue #15337. The issue occurred because stack trace information was being lost or misattributed when errors were wrapped and reported to Sentry.

## Root Cause

The problem had two contributing factors:
1. **@mastra/sentry**: was passing only an error message string to `Sentry.captureException`, causing Sentry to synthesize a stack trace from the exporter call site rather than preserving the original stack
2. **@mastra/observability**: was storing the wrapper error's stack (`MastraError.stack`) instead of the original cause's stack when building `span.errorInfo`

### observability/mastra/src/spans/default.ts
Updated `DefaultSpan.error()` to prefer the original error's stack when available. When a `MastraError` wraps another error as its cause, the span now captures `error.cause.stack` instead of the wrapper's stack, preserving the true error origin for downstream exporters.

### observability/sentry/src/tracing.ts
Modified the Sentry exporter to construct a proper `Error` object (with message, name, and stack from `span.errorInfo`) before passing it to `Sentry.captureException`, instead of passing just the error message string. This ensures Sentry receives an Error instance with the correct stack trace rather than synthesizing one.

## Testing

Added comprehensive test coverage:
- **observability/mastra/src/tracing.test.ts**: Two new tests verifying that `span.errorInfo.stack` correctly uses the original cause's stack when a `MastraError` wraps another error, and falls back to the wrapper's stack when no cause exists
- **observability/sentry/src/tracing.test.ts**: Updated existing test and added new test to verify that `SentryMock.captureException` receives a proper `Error` instance with correct `message`, `name`, and `stack` properties, including handling the case where `span.errorInfo` has no stack

## Release

Marked as patch releases for both `@mastra/observability` and `@mastra/sentry` packages.

### DIFF
--- a/.changeset/tough-geese-relax.md
+++ b/.changeset/tough-geese-relax.md
@@ -3,11 +3,11 @@
 '@mastra/sentry': patch
 ---
 
-Fixed Sentry error stack traces so they point to the actual error origin instead of the Mastra exporter internals.
+Fixed stack traces for errors reported to Sentry. Exceptions now point to the code that threw the error instead of `SentryExporter.handleSpanEnded` inside the exporter, so issues in Sentry are actually debuggable.
 
-Two related fixes:
+This was caused by two issues, both fixed:
 
-- **@mastra/sentry**: The exporter now passes an `Error` object (preserving the captured stack) to `Sentry.captureException` instead of a bare message string. Previously, passing a string caused Sentry to synthesize a stack trace from the exporter's call site, so every error in Sentry appeared to originate from `SentryExporter.handleSpanEnded`.
-- **@mastra/observability**: When a span is failed with a `MastraError` that wraps another error, `span.errorInfo.stack` now prefers the original cause's stack over the wrapper's stack, so downstream exporters receive the stack pointing to the true error origin.
+- `@mastra/sentry` passed the error message as a string to `Sentry.captureException`, which made Sentry synthesize a stack trace from the exporter's call site. It now passes an `Error` instance with the captured stack attached.
+- `@mastra/observability` stored the wrapping `MastraError`'s stack on the span, hiding the original error's location. When the `MastraError` has a cause, the cause's stack is now preserved.
 
-Fixes #15337.
+Fixes [#15337](https://github.com/mastra-ai/mastra/issues/15337).
--- a/.changeset/tough-geese-relax.md
+++ b/.changeset/tough-geese-relax.md
@@ -1,0 +1,13 @@
+---
+'@mastra/observability': patch
+'@mastra/sentry': patch
+---
+
+Fixed Sentry error stack traces so they point to the actual error origin instead of the Mastra exporter internals.
+
+Two related fixes:
+
+- **@mastra/sentry**: The exporter now passes an `Error` object (preserving the captured stack) to `Sentry.captureException` instead of a bare message string. Previously, passing a string caused Sentry to synthesize a stack trace from the exporter's call site, so every error in Sentry appeared to originate from `SentryExporter.handleSpanEnded`.
+- **@mastra/observability**: When a span is failed with a `MastraError` that wraps another error, `span.errorInfo.stack` now prefers the original cause's stack over the wrapper's stack, so downstream exporters receive the stack pointing to the true error origin.
+
+Fixes #15337.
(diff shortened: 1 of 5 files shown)


wardpeet (approved):

reactsaas:
jsut a quick question here . since iam often following the status of commits / merges here and i am wondering in terms of package releases how to tell if the fix landed in a newly released version yet or not . Like lets say there is a fix merged into core package . What is your pipeline for making a release / updating package . so i know when the fix is inclued so i can pull that new version :). like you seemed to have dropped making release notes as well in discord so I find it hard to tell whats included

intojhanurag:
> jsut a quick question here . since iam often following the status of commits / merges here and i am wondering in terms of package releases how to tell if the fix landed in a newly released version yet or not . Like lets say there is a fix merged into core package . What is your pipeline for making a release / updating package . so i know when the fix is inclued so i can pull that new version :). like you seemed to have dropped making release notes as well in discord so I find it hard to tell whats included

Drop this in discord and tag Misha , he will explain you :)

PR #15350, opened 2026-04-14
fix(cli): retry transient fetch failures in platform deploy commands

TheIsrael1:
## Summary
- retry transient transport failures while polling platform deploy status in `mastra studio deploy` and `mastra server deploy`
- keep the retry loop local to these CLI platform commands
- share only the polling error predicate in a small CLI util
- add targeted tests for retryable and non-retryable polling failures

## Scope
CLI, platform commands only.

This change is intentionally limited to deploy status polling after a deploy has been accepted.

It does not change:
- artifact upload behavior
- deploy log streaming behavior
- general platform client retry behavior

## Why
Initial deploy polling could fail on transient fetch errors even when the deploy itself later succeeded. This makes the CLI tolerate up to 3 retries after the initial attempt for retryable transport failures before exiting.

## Verification
- pnpm --filter ./packages/cli typecheck
- pnpm --filter ./packages/cli test -- src/commands/studio/platform-api.test.ts src/commands/server/platform-api.test.ts

## ELI5

When you deploy with the CLI it polls the server for status; sometimes those status checks fail due to temporary network hiccups even though the deploy succeeds. This PR makes the CLI retry those transient polling failures up to 3 times after the initial attempt, for up to 4 total attempts, before giving up, while still failing immediately for real errors.

### New Files
- **.changeset/fresh-points-occur.md**: Patch release notes documenting the fix for transient polling failures in deploy commands.

### New Utilities
- **packages/cli/src/utils/polling.ts**
  - Added `isRetryablePollingError(error: unknown): boolean` to classify retryable polling errors (AbortError by name, nested `cause.code` matching `ECONNRESET|ETIMEDOUT|ECONNREFUSED|ENOTFOUND`, or `TypeError` messages containing \"fetch failed\", case-insensitive).
  - Added `withPollingRetries<T>(fn: () => Promise<T>, maxRetries = 3): Promise<T>` — an internal retry wrapper that retries retryable errors with exponential backoff (`500 * 2^retryCount` ms) for up to 3 retries after the initial attempt, rethrowing for non-retryable errors or when retries are exhausted.

### CLI Deploy Command Updates
- **packages/cli/src/commands/studio/platform-api.ts**
  - `pollDeploy` now calls platform GET requests via `withPollingRetries`, so transient transport/fetch failures during polling are retried per the utility. Existing polling control flow (token refresh, terminal status handling, background log streaming, 2s loop wait, timeout) is preserved.

- **packages/cli/src/commands/server/platform-api.ts**
  - `pollServerDeploy` receives the same retry wrapping for its deploy-status GET calls; behavior mirrors the studio command changes.

Notes:
- Retry loop is local to these CLI platform deploy polling paths; global platform client retry behavior, artifact upload, and deploy log streaming are unchanged.

### Tests
- **packages/cli/src/commands/studio/platform-api.test.ts**
  - Added tests for `pollDeploy` covering:
    - Transient network failure retried then succeed (expects 4 GET calls total).
    - Transient failures exhausted and rejected after retries (expects 4 GET calls).
    - Non-transient error fails immediately (expects 1 GET call).
  - Restored real timers after tests with `afterEach`.

- **packages/cli/src/commands/server/platform-api.test.ts**
  - Added equivalent tests for `pollServerDeploy` with the same three scenarios and timer handling.

### Design Decisions & Scope
- Scope is limited to CLI polling for platform deploy status after a deploy has been accepted.
- Does not change artifact upload behavior, deploy log streaming, or global client retry behavior.
- Exposes only a small polling utility (error predicate and retry wrapper) to share retryability logic between the two CLI commands.
- Default retry limit is 3 retries after the initial attempt with exponential backoff; non-retryable errors still fail immediately.

### DIFF
--- a/.changeset/fresh-points-occur.md
+++ b/.changeset/fresh-points-occur.md
@@ -1,0 +1,5 @@
+---
+'mastra': patch
+---
+
+Fixed `mastra studio deploy` and `mastra server deploy` so transient polling failures are retried up to 3 times before the CLI exits.
--- a/packages/cli/src/commands/server/platform-api.test.ts
+++ b/packages/cli/src/commands/server/platform-api.test.ts
@@ -1,4 +1,4 @@
-import { describe, expect, it, vi, beforeEach } from 'vitest';
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
 
 const mockGET = vi.fn();
 const mockPOST = vi.fn();
@@ -24,6 +24,10 @@ vi.mock('../auth/credentials.js', () => ({
 beforeEach(() => {
   vi.resetAllMocks();
   vi.stubGlobal('fetch', vi.fn());
+});
+
+afterEach(() => {
+  vi.useRealTimers();
 });
 
 describe('fetchServerProjects', () => {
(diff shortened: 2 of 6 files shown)


rphansen91 (commented):
## Review comments

### 1. Retry count naming (nit)

The PR description says "retry up to 3 times" and the code does `retryCount >= 3` — meaning initial attempt + 3 retries = **4 total attempts**. The test assertions (`toHaveLength(4)`) are correct, so this is consistent in practice. Just calling out that "3 retries" ≠ "3 attempts" in case anyone reads the description expecting 3 total calls.

### 2. Duplicated retry loop

The retry `while` loop in `studio/platform-api.ts` and `server/platform-api.ts` is identical (~15 lines each). Consider extracting a small helper:

```ts
async function withPollingRetries<T>(fn: () => Promise<T>, maxRetries = 3): Promise<T> { ... }
```

This would make future changes (e.g., adjusting backoff or max retries) less error-prone since you'd only need to update one place.

### 3. `'fetch failed'` case sensitivity (nit)

In `polling.ts:17`:
```ts
return error instanceof TypeError && error.message.toLowerCase().includes('fetch failed');
```

Node's native `TypeError` for failed fetches already says `"fetch failed"` (lowercase), so the `.toLowerCase()` is technically unnecessary for the known case. Defensive and harmless — just noting it.

TheIsrael1:
Good catches.

  1. On the retry count wording: the implementation is 1 initial
     attempt plus up to 3 retries, for 4 total attempts. The tests
     were already asserting that correctly, and I updated the PR
     description to make the wording explicit.
  2. On the duplicated retry loop: agreed. I extracted the shared
     loop into withPollingRetries() in the polling util so the
     retry/backoff behavior now lives in one place, while keeping
     the deploy polling behavior scoped to these CLI platform
     commands.
  3. On toLowerCase(): agreed it isn’t strictly necessary for the
     known Node fetch failed case, but I left it in as harmless
     defensive normalization.

> ## Review comments
> ### 1. Retry count naming (nit)
> The PR description says "retry up to 3 times" and the code does `retryCount >= 3` — meaning initial attempt + 3 retries = **4 total attempts**. The test assertions (`toHaveLength(4)`) are correct, so this is consistent in practice. Just calling out that "3 retries" ≠ "3 attempts" in case anyone reads the description expecting 3 total calls.
> 
> ### 2. Duplicated retry loop
> The retry `while` loop in `studio/platform-api.ts` and `server/platform-api.ts` is identical (~15 lines each). Consider extracting a small helper:
> 
> ```ts
> async function withPollingRetries<T>(fn: () => Promise<T>, maxRetries = 3): Promise<T> { ... }
> ```
> 
> This would make future changes (e.g., adjusting backoff or max retries) less error-prone since you'd only need to update one place.
> 
> ### 3. `'fetch failed'` case sensitivity (nit)
> In `polling.ts:17`:
> 
> ```ts
> return error instanceof TypeError && error.message.toLowerCase().includes('fetch failed');
> ```
> 
> Node's native `TypeError` for failed fetches already says `"fetch failed"` (lowercase), so the `.toLowerCase()` is technically unnecessary for the known case. Defensive and harmless — just noting it.

abhiaiyer91 (approved):

PR #15352, opened 2026-04-14
refactor: refine Mastra Code autonomy prompts

TylerBarnes:
This updates Mastra Code's base and build prompts so the agent is more autonomous by default, but still stops to ask when ambiguity would materially change the outcome and reports blockers instead of thrashing.

Before:
```ts
- If you're unsure about something, ask the user rather than guessing.
- If the approach is risky or ambiguous, ask the user before proceeding
- If stuck after 2 attempts, tell the user what you've tried
```

After:
```ts
- Make reasonable assumptions when the most sensible path is clear, but ask the user when ambiguity is material and could change the outcome.
- If multiple approaches are plausible and the choice would materially affect scope, behavior, or risk, stop and ask the user
- If progress is blocked after reasonable attempts, briefly explain the blocker, what you've tried, and the next best option
```

Also adds a patch changeset for mastracode.

### DIFF
--- a/mastracode/src/agents/prompts/base.ts
+++ b/mastracode/src/agents/prompts/base.ts
@@ -85,6 +85,7 @@ You are an autonomous AI assistant with strong common sense reasoning capabiliti
 **Common Sense Reasoning**
 - Apply implicit knowledge about how the world works (cause-and-effect, social norms, practical constraints)
 - Consider the user's likely intent, not just literal words
+- Make reasonable assumptions when the most sensible path is clear, but ask the user when ambiguity is material and could change the outcome.
 - Bias towards action, but be flexible in your rules. If you think the user would want you to ask them, then do! Especially if they've previously stated a preference that you do in the specific situation.
 
 **Decision Framework**
--- a/mastracode/src/agents/prompts/base.ts
+++ b/mastracode/src/agents/prompts/base.ts
@@ -35,16 +35,6 @@ ${ctx.toolGuidance}
 - For unfamiliar codebases, check git log to understand recent changes and patterns.
 - Identify existing conventions (naming, structure, error handling) and follow them.
 
-## Work Incrementally
-- Focus on ONE thing at a time. Complete it fully before moving to the next.
-- Leave the codebase in a clean state after each change — no half-implemented features.
-- For multi-step tasks, use tasks to track progress and ensure nothing is missed.
-
-## Verify Before Moving On
-- After each change, verify it works. Don't assume — actually test it.
-- Run the relevant tests, check for type errors, or manually verify the behavior.
-- If something breaks, fix it immediately. Don't pile more changes on top of broken code.
-
 # Coding Philosophy
 
 - **Avoid over-engineering.** Only make changes that are directly requested or clearly necessary.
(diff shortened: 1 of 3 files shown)


mastracode/src/agents/prompts/base.ts
```diff
@@ -35,16 +35,6 @@ ${ctx.toolGuidance}
...
 
-## Work Incrementally
-- Focus on ONE thing at a time. Complete it fully before moving to the next.
-- Leave the codebase in a clean state after each change — no half-implemented features.
-- For multi-step tasks, use tasks to track progress and ensure nothing is missed.
-
-## Verify Before Moving On
-- After each change, verify it works. Don't assume — actually test it.
-- Run the relevant tests, check for type errors, or manually verify the behavior.
-- If something breaks, fix it immediately. Don't pile more changes on top of broken code.
```
TylerBarnes, line 46: this was already in the build prompt

mastracode/src/agents/prompts/base.ts
```diff
@@ -70,7 +59,7 @@ Don't commit files likely to contain secrets (\`.env\`, \`*.key\`, \`credentials
 Write commit messages that explain WHY, not just WHAT. Match the repo's existing style. Include \`Co-Authored-By: Mastra Code${ctx.modelId ? ` (${ctx.modelId})` : ''} <noreply@mastra.ai>\` in the message body.
 
 ## Pull Requests
-Use \`gh pr create\`. Include a summary of what changed and a test plan.
+Use \`gh pr create\`. Include a summary of what changed and a test plan. Word the pull request title/description to explain the entire unit of work being shipped, worded to explain it to someone who doesn't know anything about the work being shipped. Do not add details of fixes that were needed along the way.
```
TylerBarnes, line 62: I see gpt-5.4 often making pr descriptions that are about individual commits rather than the full pr

mastracode/src/agents/prompts/base.ts
```diff
@@ -80,11 +69,10 @@ Use \`gh pr create\`. Include a summary of what changed and a test plan.
 - NEVER guess file paths or function signatures. Use search_content/find_files to find them.
 - NEVER make up URLs. Only use URLs the user provides or that you find in the codebase.
 - When referencing code locations, include the file path and line number.
-- If you're unsure about something, ask the user rather than guessing.
 
 # File Access & Sandbox
 
-By default, you can only access files within the current project directory. If you get a "Permission denied" or "Access denied" error when trying to read, write, or access files outside the project root, do NOT keep retrying. Instead, use the \`request_access\` tool to request access to the external directory. Only tell the user to run \`/sandbox\` themselves if the tool is unavailable or the request cannot be made from the current context.
```
TylerBarnes, line 87: the user basically never needs to run `/sandbox`

mastracode/src/agents/prompts/build.ts
```diff
@@ -50,9 +50,10 @@ For each change you make:
 1. **Understand** — Read the relevant code. Check how similar things are done elsewhere.
 2. **Implement** — Make the change. Follow existing patterns and conventions.
 3. **Verify** — Test that it works. Don't assume — actually run it.
-4. **Clean up** — Ensure no broken code, no debug statements, no half-done features.
+4. **Prove** — Prove that it works. The burden of proof is on YOU.
```
TylerBarnes, line 53: proving it to the user is an important step that can be overlooked

DanielSLew (approved):

PR #15366, opened 2026-04-14
fix(client-js,react): support HeadersInit for custom client headers

aryanbarde80:
## Description

Allow client SDK custom headers to use the full fetch `HeadersInit` shapes instead of only plain string maps.

This updates both typing and runtime behavior so callers can pass:
- plain objects
- `Headers` instances
- header tuple arrays

It also fixes the React provider merge path so local dev header injection still works with non-object header inputs.

## Related Issue(s)

Fixes the client SDK custom headers typing/runtime issue.

## Type of Change

- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [x] Test update

## Checklist

## ELI5

The PR lets developers pass custom HTTP headers to the Mastra SDK in more flexible ways—not just as a simple list of pairs, but also as JavaScript objects or arrays. It also fixes a bug where the React version couldn't properly add its own dev headers when developers used these flexible formats.

### Core Type Updates
- Introduced `ClientHeaders` type alias equal to `HeadersInit` in `client-sdks/client-js/src/types.ts`
- Updated `ClientOptions.headers` and `RequestOptions.headers` from `Record<string, string>` to `ClientHeaders`, enabling support for:
  - Plain objects
  - `Headers` instances
  - Header tuple arrays `[key, value][]`

### Header Normalization
- Added `normalizeHeaders()` helper function in `client-sdks/client-js/src/resources/base.ts` to convert any `HeadersInit` form into a `Record<string, string>`
- Updated `BaseResource.request()` to use this helper when merging client-level and request-level headers
- Preserved conditional `content-type: application/json` header setting for POST/PUT/PATCH/DELETE requests without `FormData`

### React Provider Fix
- Updated `MastraClientProvider` props type to accept `ClientHeaders` instead of `Record<string, string>`
- Fixed the local dev header injection (x-mastra-dev-playground) to work with non-object header inputs by converting to `Headers` instance, then normalizing back to object form

### Testing & Documentation
- Added two new test cases validating:
  - `BaseResource` with `Headers` instance in client options
  - Tuple-array headers merged at request level
- Updated README to document the expanded `HeadersInit` support for the `headers` configuration option
- Added Changeset entry for patch-level releases

## Impact
- **Backward compatible:** Existing code using plain object headers continues to work unchanged
- **Bug fix:** Resolves the React provider's inability to merge dev-injected headers with non-object header inputs
- **Enhanced flexibility:** SDK users can now use standard Fetch API header patterns across their codebase


abhiaiyer91:
Thanks for the contribution! We looked at this and decided not to merge it.

The current `Record<string, string>` type for headers works well and is the simplest API for our users. The "bug" in the React provider (spreading non-object headers) can't actually occur today because the type system enforces plain objects — nobody can pass a `Headers` instance or tuple array without bypassing types.

Widening to `HeadersInit` adds a normalization layer and type alias without solving a real problem. We'd rather keep the API surface simple.

Appreciate the thorough PR though!

PR #15407, opened 2026-04-15
Fix: Is there an ETA/plan for AI SDK 4.0 -> v5.0 migration?

shahyashish:
The change modifies how the root path is set for nested projects in the vitest configuration. Previously, the root was set to `'./${projectDir}'`. This was incorrect, as it would resolve the root relative to the current working directory instead of the vitest config file. The corrected code sets the root to just `${projectDir}`. This ensures that the root path is correctly set to the directory containing the vitest config file, and nested projects can correctly locate their files.

Test: To verify the fix, you can create a vitest config file with nested projects in a subdirectory. Run vitest and ensure that the tests in the nested projects are able to find their source files. Before the fix, the tests would fail because the root path was incorrect.

## ELI5 (Explain Like I'm 5)

Imagine you have a box with smaller boxes inside it, and each smaller box has instructions on where to find things. Before this fix, the instructions said "look in ./folder" (starting from wherever you happen to be), which got confusing. Now they say "look in folder" (starting from the box itself), so everything works correctly.

## Summary

This PR adds a root-level Vitest configuration that discovers and runs all test projects across the monorepo. The key fix corrects how nested inline Vitest project configurations set the `test.root` path.

**Before**: `root: './${projectDir}'` — resolved paths relative to the current working directory
**After**: `root: '${projectDir}'` — resolves paths relative to the Vitest config file location

This ensures nested projects can correctly locate their source files, since the root now properly points to the directory containing the Vitest config file rather than relying on the process working directory.

## Changes Made

- **vitest.config.ts (new)**: Added root-level Vitest configuration with automatic project discovery
  - Scans multiple directories for nested `vitest.config.ts` files using glob patterns
  - Expands inline nested project configurations with the corrected `root` path format
  - Excludes specific directories from discovery (e.g., playground, vendored packages)
  - Falls back to directory path if config parsing fails

The fix specifically addresses the `root` path assignment for inline nested projects (lines that now use `root: '${projectDir}'` instead of `root: './${projectDir}'`), ensuring paths are resolved correctly relative to the config file location.

## Testing Instructions

To verify the fix:
1. Create a Vitest config with nested projects placed in a subdirectory
2. Run Vitest
3. Before the fix: tests in nested projects would fail due to incorrect root resolution
4. After the fix: tests should find their source files correctly

### DIFF
--- a/vitest.config.ts
+++ b/vitest.config.ts
@@ -86,7 +86,7 @@ async function discoverProjects(): Promise<TestProjectConfiguration[]> {
             ...projectConfig,
             test: {
               ...projectConfig.test,
-              root: `./${projectDir}`,
+              root: `${projectDir}`,
             },
           });
         }


itskrishna21 (changes requested):
@shahyashish  wanted to understand how these changes reflect the core issue?

PR #15416, opened 2026-04-15
fix(core): remove noisy browser not running reminder

NikAiyer:
## Description

Remove the "Browser is not currently running" system reminder that was injected on every turn even when the user wasn't doing browser-related tasks. Now the system-reminder only appears when there's meaningful browser context (URL or page title).

## Related Issue(s)

N/A - Internal improvement

## Type of Change

- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [x] Test update

## Checklist

## ELI5 Explanation

Previously the agent would nag "Browser is not currently running" on every turn even when you weren't doing anything with the browser. This PR stops that noise so the agent only injects browser reminders when there's actual browser context (like a page URL or title).

## Changes

- Remove `isRunning` from `BrowserContext`
  - The optional `isRunning?: boolean` field was removed from the `BrowserContext` interface in packages/core/src/browser/processor.ts.

- Only inject per-request browser reminders when meaningful context exists
  - The processor no longer prepends a "Browser is not currently running" reminder on every turn. processInputStep now injects a <system-reminder> only when per-request browser data is present (currentUrl and/or pageTitle).

- Agent browser context initialization adjusted
  - In packages/core/src/agent/agent.ts, the agent now fetches browserCtx.currentUrl only when the relevant browser thread is running (isThreadRunning), aligning initialization with the conditional reminder injection.

- Tests updated
  - packages/core/src/browser/processor.test.ts: removed tests that expected the global "browser not running" reminder and added/updated tests to cover URL+title and pageTitle-only injection behavior and the "no per-request data" case.

- Changeset
  - .changeset/remove-browser-not-running-reminder.md added to document the patch release for @mastra/core describing the fix.

This is classified as a bug fix with accompanying test updates.

### DIFF
--- a/packages/core/src/browser/processor.ts
+++ b/packages/core/src/browser/processor.ts
@@ -45,9 +45,6 @@ export interface BrowserContext {
 
   /** Current page title (updated per-request) */
   pageTitle?: string;
-
-  /** Whether browser is currently running */
-  isRunning?: boolean;
 }
 
 /**
@@ -90,10 +87,6 @@ export class BrowserContextProcessor {
 
     if (ctx.pageTitle) {
       parts.push(`Page title: ${ctx.pageTitle}`);
-    }
-
-    if (ctx.isRunning === false) {
-      parts.push('Browser is not currently running.');
     }
 
     if (parts.length === 0) return;
--- a/.changeset/remove-browser-not-running-reminder.md
+++ b/.changeset/remove-browser-not-running-reminder.md
@@ -2,12 +2,4 @@
 '@mastra/core': patch
 ---
 
-Remove noisy "Browser is not currently running" system reminder.
-
-**What changed**
-
-- Removed the `isRunning` field from `BrowserContext` interface
-- The `<system-reminder>` tag is now only injected when there's meaningful browser context (current URL or page title)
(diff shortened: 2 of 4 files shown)


CalebBarnes (approved):

packages/core/src/browser/processor.ts
TylerBarnes, line 96: I know this isn't part of this pr, but we shouldn't modify history cause it breaks the prompt cache. System reminders should always be added as a brand new user message on the very end of history. Take a look at how agents md injector does it. You can also set a metadata key on the reminder so that it doesn't leak into the UI

TylerBarnes: basically you want to rotateResponseMessageId?.() so that if we're in the middle of an assistant turn it will split it into a new message (new ID), then just messageList.add(reminder, "input")

NikAiyer: @TylerBarnes I'll try looking into this too, and see if we can update this processor in this PR as well.

NikAiyer: @TylerBarnes Actually I'll make a followup for this

PR #15417, opened 2026-04-15
fix(core): add browser reminders as new messages to preserve prompt cache

NikAiyer:
## Description

Instead of prepending browser context to existing user messages (which breaks prompt cache), browser context reminders are now added as **new user messages** at the end of history. This follows the same pattern as `AgentsMDInjector`.

The messages include `systemReminder` metadata to prevent them from leaking into the UI.

**Changes:**
- Add new user message with `systemReminder: { type: 'browser-context' }` metadata
- Call `rotateResponseMessageId()` after adding reminder
- Skip duplicate reminders if same content already exists
- Remove `isRunning` field from `BrowserContext` interface (also removes the noisy "Browser is not currently running" reminder from #15416)

## Related Issue(s)

Follow-up to #15416 (review comment from @TylerBarnes)

## Type of Change

- [ ] Documentation update
- [ ] Code refactoring
- [x] Performance improvement
- [x] Test update

## Checklist

## Testing

The unit tests cover:
- Adding new user message with correct metadata
- URL and title in reminder content
- Title-only reminder
- No reminder when no browser context
- Duplicate prevention (same content)
- New reminder when URL changes

**To verify prompt caching works:** We should test that the message history before the browser reminder is unchanged, allowing the LLM provider to cache it. This could be validated by checking that `messageList.add()` is called rather than modifying existing messages.

## ELI5

Instead of changing the user's last message to add a browser reminder, the code now appends the reminder as a brand-new user message so the AI's prompt cache isn't broken and the UI doesn't accidentally show system reminder data.

## Overview

This PR changes browser context reminder injection to preserve prompt caching by appending reminders as new user messages (with explicit reminder metadata) and adding deduplication and escaping logic. It also removes an obsolete BrowserContext field and adds tests covering the new behavior.

### Core Implementation (packages/core/src/browser/processor.ts)
- Reminder injection now appends a new user message (MastraMessageContentV2) containing a <system-reminder type="browser-context"> payload instead of prepending markup into the last user message.
- Reminder messages include content.metadata.systemReminder (at minimum { type: 'browser-context' } — tests contain { type: 'browser-context', url, title }) so UI rendering won’t leak system reminder content.
- XML-escapes URL/title before inserting them into reminder markup.
- Adds helper functions: createBrowserReminderMessage, escapeXml, hasTrailingBrowserReminder, and a REMINDER_TYPE constant.
- Implements deduplication: reads persisted history (messageList.get.all.db()) and skips injection when the trailing reminder already matches current url/title; injects a new reminder when the URL changes (including navigation sequences like A→B→A).
- Calls rotateResponseMessageId() after successfully adding a reminder.
- Returns the messageList (MessageList | undefined) flow instead of mutating/returning modified existing messages.
- Removes isRunning from BrowserContext (and thereby removes the "Browser is not currently running" reminder path introduced earlier).

### Tests (packages/core/src/browser/processor.test.ts)
- Adds createMockMessageList test helper and updates createInputStepArgs to use it.
- Asserts reminder insertion via messageList.add() (verifying prompt cache preservation) and that rotateResponseMessageId() is invoked.
- Tests include:
  - Creation of new user reminder message with correct metadata and markup.
  - URL + title content and title-only reminders.
  - No reminder when no browser context.
  - Deduplication: identical reminders are skipped.
  - New reminder added when URL changes (including A→B→A sequences and when trailing message is non-reminder).
  - XML escaping of special characters in URL/title.
- Adjusts assertions to validate messageList.add() usage rather than mutating existing messages.

### Changeset
- .changeset/browser-reminder-prompt-cache.md added for @mastra/core (patch).

## Impact
- Preserves prompt cache by keeping prior message history immutable before reminder injection.
- Prevents duplicate reminders while allowing updates on URL changes.
- Keeps system reminder data separated from UI-visible message content via metadata.
- Removes obsolete BrowserContext field and its related "browser not running" reminder.

## Testing notes
- Verify prompt caching by asserting messageList.add() is called (i.e., prior message objects are not modified).
- Confirm rotateResponseMessageId() is called after adding the reminder.

### DIFF
--- a/packages/core/src/browser/processor.ts
+++ b/packages/core/src/browser/processor.ts
@@ -93,14 +93,14 @@ export class BrowserContextProcessor {
     const reminderText = parts.join(' | ');
     const reminderMarkup = `<system-reminder type="${REMINDER_TYPE}">${reminderText}</system-reminder>`;
 
-    // Check if we already have this exact reminder to avoid duplicates
+    // Check if we already have a reminder with the same URL/title (scan from newest)
     const existingMessages = args.messageList.get.all.db();
-    if (hasExistingBrowserReminder(existingMessages, reminderMarkup)) {
+    if (hasExistingBrowserReminder(existingMessages, ctx.currentUrl, ctx.pageTitle)) {
       return;
     }
 
     // Add as a new user message at the end of history to preserve prompt cache
-    const reminderMessage = createBrowserReminderMessage(reminderMarkup);
+    const reminderMessage = createBrowserReminderMessage(reminderMarkup, ctx.currentUrl, ctx.pageTitle);
     args.messageList.add(reminderMessage, 'user');
     args.rotateResponseMessageId?.();
 
@@ -108,14 +108,28 @@ export class BrowserContextProcessor {
   }
 }
 
-function createBrowserReminderMessage(reminderMarkup: string): MastraDBMessage {
+interface BrowserReminderMetadata {
+  type: typeof REMINDER_TYPE;
+  url?: string;
+  title?: string;
+}
+
+function createBrowserReminderMessage(
+  reminderMarkup: string,
+  url: string | undefined,
+  title: string | undefined,
+): MastraDBMessage {
+  const reminderMeta: BrowserReminderMetadata = {
+    type: REMINDER_TYPE,
+    url,
+    title,
+  };
+
   const content: MastraMessageContentV2 = {
     format: 2,
     parts: [{ type: 'text', text: reminderMarkup }],
     metadata: {
-      systemReminder: {
-        type: REMINDER_TYPE,
-      },
+      systemReminder: reminderMeta,
     },
   };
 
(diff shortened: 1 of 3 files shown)


packages/core/src/browser/processor.ts
```diff
@@ -92,43 +88,60 @@ export class BrowserContextProcessor {
...
+function hasExistingBrowserReminder(messages: MastraDBMessage[], reminderMarkup: string): boolean {
+  for (const msg of messages) {
+    if (msg.role !== 'user') continue;
+
+    const metadata = msg.content.metadata;
+    if (typeof metadata === 'object' && metadata !== null && 'systemReminder' in metadata) {
+      const reminder = (metadata as { systemReminder?: { type?: string } }).systemReminder;
+      if (reminder?.type === REMINDER_TYPE) {
+        // Check if the content matches (same URL/title)
+        const textPart = msg.content.parts?.find((p): p is { type: 'text'; text: string } => p.type === 'text');
```
TylerBarnes: probably worth using metadata for this too instead of matching string content

TylerBarnes: could add the url/title to metadata as well

TylerBarnes (approved):
awesome! 🎉 looks great, one optional comment

PR #15420, opened 2026-04-15
feat: activate observations on provider changes

TylerBarnes:
This adds provider/model-change activation for observational memory and wires it through MastraCode.

After this, switching models can immediately activate observations instead of waiting for the next threshold or idle timeout.

```ts
observationalMemory: {
  activateOnProviderChange: true,
}
```

Under the hood, step-start parts now carry the model used for each step, so model changes can be detected across steps, even within a single assistant message. The PR also includes MastraCode TUI updates, docs, and changesets.
Screenshot 2026-04-15 at 3 47 39 PM

## ELI5

When you switch to a different AI model/provider, the system can immediately turn recent buffered observations into active memory so the new model starts with the same compact context—saving tokens and avoiding re-sending large unactivated buffers.

## Overview

Adds provider/model-change activation for Observational Memory and wires it through MastraCode so buffered observations and reflections can be activated immediately when execution moves to a different provider/model (config-gated).

## Key Changes

- New config flag: observationalMemory.activateOnProviderChange?: boolean (default false) — when true, OM activates buffered observations/reflections on provider/model switch.
- Activation logic:
  - Activation triggers now include 'provider_change' in addition to 'threshold' and 'ttl'.
  - Activation metadata and markers carry previousModel and currentModel when triggered by provider change.
  - Helpers added: getLastModelFromMessages and getCurrentModel to derive model context.
- Message/step metadata:
  - step-start parts now optionally include model?: string so model changes can be detected across steps and within a single assistant message.
  - MessageMerger stamps synthetic step-start parts with model metadata when available.
  - MessageList.enrichLastStepStart(model: string): boolean added to annotate the last step-start.
- Observational memory & reflection plumbing:
  - ObservationalMemory.activate and ReflectorRunner.maybeReflect accept currentModel and can trigger activation/reflection when actor model differs from lastModel.
  - Activation marker emission and OM processor forward previousModel/currentModel through events/markers and harness events.
  - ObservationTurn and OM processor state carry actorModelContext for per-step model tracking.
- MastraCode TUI integration:
  - New OM marker type om_activation_provider_change with previousModel/currentModel and UI text "Model changed previous → current, activating observations".
  - Event dispatch and handlers extended to handle triggeredBy: 'provider_change' and forward model info; TUI state tracks activeActivationProviderChangeMarker and clears it during buffering to avoid stale UI.
  - Tests added to validate marker rendering and TUI behavior.
- Execution flow & metadata:
  - LLM execution step threads model context into output processing and enriches last step-start for fallback iterations.
  - Response/DB/UI message metadata now include provider alongside modelId when available (tests/snapshots updated).
- Storage/serialization:
  - Observational memory serialization includes activateOnProviderChange so the flag persists in stored config.
- Docs & changesets:
  - Documentation updated to document activateOnProviderChange and rationale (preserve prompt-cache savings across provider/model switches).
  - Changesets added for mastracode, @mastra/core, and @mastra/memory describing the provider/model-change activation behavior.

## Tests

- New and updated tests/snapshots:
  - OM activation on provider-change integration/unit tests.
  - MessageList step-start enrichment tests.
  - MastraCode TUI OM marker rendering tests.
  - Updated snapshots to include provider metadata where applicable.

## Rationale / Impact

When enabled, activateOnProviderChange ensures switching providers/models immediately activates buffered observations/reflections so the new provider receives a compact, activated context rather than a large raw token window—reducing token waste and preserving prompt-cache savings.

### DIFF
--- a/docs/src/content/en/docs/memory/observational-memory.mdx
+++ b/docs/src/content/en/docs/memory/observational-memory.mdx
@@ -379,7 +379,7 @@ const memory = new Memory({
 
 With a 5-minute prompt cache TTL, this activates buffered context after 5 minutes of inactivity so the next uncached prompt uses observations and reflections instead of a larger raw message window. If you prefer, `300_000` works the same way.
 
-If your agent can switch between providers or models mid-thread, `activateOnProviderChange: true` forces buffered context to activate before the new provider runs. That avoids sending a large raw window to a provider that cannot reuse the previous prompt cache.
+Changing model or providers mid-thread will invalidate the prompt cache. If your agent can switch between providers or models mid-thread, `activateOnProviderChange: true` forces buffered context to activate before the new provider runs. That avoids sending a large raw window to a provider that cannot reuse the previous prompt cache.
 
 ### Disabling
 
--- a/docs/src/content/en/docs/memory/observational-memory.mdx
+++ b/docs/src/content/en/docs/memory/observational-memory.mdx
@@ -359,6 +359,7 @@ Reflection works similarly — the Reflector runs in the background when observa
 | `observation.bufferActivation` | `0.8` | How aggressively to clear the message window on activation. `0.8` means remove enough messages to keep only 20% of `messageTokens` remaining. Lower values keep more message history. |
 | `observation.blockAfter` | `1.2` | Safety threshold as a multiplier of `messageTokens`. At `1.2`, synchronous observation is forced at 36k tokens (1.2 × 30k). Only matters if buffering can't keep up. |
 | `activateAfterIdle` | none | Forces buffered observations and buffered reflections to activate after a period of inactivity, even if their token thresholds have not been reached yet. Accepts milliseconds or duration strings like `300_000`, `"5m"`, or `"1hr"`. Set this to your prompt cache TTL if you want activation to happen before the next cold prompt. |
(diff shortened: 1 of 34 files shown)


CalebBarnes (approved):

docs/src/content/en/docs/memory/observational-memory.mdx
```diff
@@ -370,13 +371,16 @@ const memory = new Memory({
...
       activateAfterIdle: '5m',
+      activateOnProviderChange: true,
     },
   },
 })
 ```
 
 With a 5-minute prompt cache TTL, this activates buffered context after 5 minutes of inactivity so the next uncached prompt uses observations and reflections instead of a larger raw message window. If you prefer, `300_000` works the same way.
 
+If your agent can switch between providers or models mid-thread, `activateOnProviderChange: true` forces buffered context to activate before the new provider runs. That avoids sending a large raw window to a provider that cannot reuse the previous prompt cache.
```
DanielSLew: I feel like the reason for this was buried at the end of the sentence. I think it'll catch more peoples eyes to say it up front the reason for the reason.
```suggestion
Changing model or providers mid-thread will invalidate the prompt cache. If your agent can switch between providers or models mid-thread, `activateOnProviderChange: true` forces buffered context to activate before the new provider runs. That avoids sending a large raw window to a provider that cannot reuse the previous prompt cache.
```

DanielSLew (approved):

PR #15437, opened 2026-04-16
feat(s3): add AWS credential provider chain support

### LINKED ISSUE #14289: S3Filesystem workspace actions fail with SignatureDoesNotMatch
This issue was created from Discord post 1482178409164705996:

[Open in Discord](https://discord.com/channels/1309558646228779139/1482178409164705996)

Workspace write actions using the S3Filesystem fail with this message:
```
SignatureDoesNotMatch: The request signature we calculated does not match the signature you provided. Check your key and signing method.
```

This only happens in my deployed app where I'm using the container metadata service for temporary credentials, and not locally where I'm using the SSO credential provider. I pass the `accessKeyId`, `secretAccessKey`, and `sessionToken` to the S3Filesystem in both instances.

I thought it might be because the aws credentials providers provide some additional fields that are necessary for the signature, so I patched @mastra/s3 to also pass through `credentialScope` and `accountId` to the underlying S3 client, but that also did not work.

Perhaps if the S3Filesystem could just take an S3 credentials provider as a parameter, or use the default credentials chain, or even take an entire S3 client as a parameter, that would circumvent all of these issues.

Here are my package versions:
```json
    "@mastra/core": "^1.11.0",
    "@mastra/hono": "^1.2.0",
    "@mastra/lance": "^1.0.4",
    "@mastra/loggers": "^1.0.2",
    "@mastra/mcp": "^1.1.0",
    "@mastra/memory": "^1.6.2",
    "@mastra/observability": "^1.4.0",
    "@mastra/pg": "^1.8.0",
    "@mastra/s3": "^0.2.2",
    "mastra": "^1.3.8",

```

KiraPC:
Fixes https://github.com/mastra-ai/mastra/issues/14289

`S3Filesystem` only accepted static `accessKeyId`/`secretAccessKey` credentials. When omitted, it fell back to anonymous access (empty credentials + no-op signer), which actively prevented the AWS SDK default credential provider chain from working. This caused `SignatureDoesNotMatch` errors on ECS/Lambda deployments where temporary credentials expire.

This PR adds three-tier credential resolution to both `S3Filesystem` and `S3BlobStore`:

1. New `credentials` option — pass a provider function (e.g. `fromNodeProviderChain()`) for auto-refreshing credentials
2. Static `accessKeyId`/`secretAccessKey` — works exactly as before (backward compatible)
3. SDK default chain — when all credential options are omitted, the SDK discovers credentials from the environment automatically (env vars, `~/.aws`, IMDS, ECS container credentials)

```typescript
import { S3Filesystem } from '@mastra/s3'
import { fromNodeProviderChain } from '@aws-sdk/credential-providers'

// Auto-refreshing credentials (ECS task role, SSO, etc.)
const fs = new S3Filesystem({
  bucket: 'my-bucket',
  region: 'us-east-1',
  credentials: fromNodeProviderChain(),
})

// Or just let the SDK figure it out
const fs2 = new S3Filesystem({
  bucket: 'my-bucket',
  region: 'us-east-1',
})
```

**A note on credential patterns across the repo:** the DynamoDB store takes a different approach — it accepts a pre-configured `DynamoDBDocumentClient` directly, which lets users set up credential providers at the client level. Here we went with a `credentials` option on the constructor instead. It might be worth discussing whether these two patterns should converge into a shared, consistent approach across all AWS-backed stores (e.g. always accept either a pre-configured client or a credentials provider). Happy to align if there's a preference.

## Test plan
- [x] All 115 existing + new unit tests pass (`pnpm test:unit` in `workspaces/s3`)
- [ ] Verify with real ECS/Lambda task role credentials
- [x] Verify with `fromNodeProviderChain()`

## ELI5

S3Filesystem and S3BlobStore no longer fall back to "anonymous" when no keys are provided — they now accept an auto-refreshing credentials provider, static keys, or defer to the AWS SDK's built-in discovery so temporary credentials (ECS/Lambda/SSO/AssumeRole) keep working.

---

## Changes Summary

This PR replaces the previous anonymous fallback and implements a three-tier AWS credential resolution for S3 consumers to fix SignatureDoesNotMatch failures with expiring temporary credentials.

Core implementation
- Add credentials?: AwsCredentialIdentity | AwsCredentialIdentityProvider to S3FilesystemOptions and S3BlobStoreOptions.
- Make accessKeyId and secretAccessKey optional.
- getClient() credential precedence:
  1. options.credentials (provider function or static identity)
  2. static accessKeyId/secretAccessKey (+ optional sessionToken) only if both static keys present
  3. otherwise leave credentials unset so the AWS SDK default credential provider chain is used (env vars, AWS config, IMDS/ECS, etc.)
- Remove the anonymous/public-bucket shortcut (empty credentials + no-op signer) that blocked SDK discovery and credential refresh.

API / Types
- S3BlobStoreOptions and S3FilesystemOptions: added credentials? and made static keys optional.
- Internal class fields updated accordingly; S3Client is constructed conditionally with resolved credentials.

Documentation
- Updated docs and workspaces/s3 README to document credential precedence, examples showing credentials: fromNodeProviderChain(), and guidance that omitting credentials uses the SDK default chain.
- JSDoc on getMountConfig() clarifies mount configs only include static accessKeyId/secretAccessKey/sessionToken; provider-based credentials are not included since FUSE mounts cannot call provider functions.
- Notes that @aws-sdk/credential-providers must be installed separately when using fromNodeProviderChain().

Tests
- Added/updated unit tests in workspaces/s3 to cover credential-resolution cases:
  - provider forwarded to S3Client,
  - static keys assembled into static credentials (sessionToken handled accordingly),
  - no credentials results in S3Client without explicit credentials (allowing SDK discovery),
  - credentials option takes precedence over static keys.
- All existing + new s3 tests pass locally; real ECS/Lambda task-role verification is pending.

Packaging
- Added @aws-sdk/types as a peer dependency in workspaces/s3/package.json.

Notes & follow-ups
- Callers relying on the previous enforced anonymous behavior must adjust.
- Both accessKeyId and secretAccessKey are still required to be treated as static; providing only one falls through to the SDK chain and will surface request-time errors (by design).
- Reviewer suggestions: one JSDoc change applied; two refactor nits declined (examples retained per changeset rules; shared helper extraction deferred).
- Consider aligning S3 behavior with DynamoDB store (which accepts a pre-configured client) as a future improvement.


KiraPC:
## Review feedback notes

Addressed one of the three nitpick suggestions from review. Here's the reasoning for each:

### 1. Remove code examples from changeset — **Not applied**

The project's own changeset guidelines (`.claude/commands/changeset.md`) explicitly state: *"If the change is a breaking change or is adding a new feature, ensure that a code example is provided."* Since this is a new feature (`minor` bump), the code examples are required per convention. Keeping them as-is.

### 2. Extract shared credential resolution helper — **Not applied**

The credential resolution logic is ~10 lines of a straightforward if/else if in both `S3Filesystem.getClient()` and `S3BlobStore.getClient()`. These two classes don't share a common base class (`MastraFilesystem` vs `BlobStore`), and no other AWS-backed store in the repo uses a shared credential helper either. Extracting a utility for this would be a premature abstraction — if the two classes diverge in credential handling later (e.g., blob-store accepting a pre-configured client), the shared helper becomes a constraint rather than a benefit.

### 3. JSDoc on `getMountConfig()` for credential provider limitation — **Applied** ✓

Added a JSDoc comment documenting that `getMountConfig()` only includes static `accessKeyId`/`secretAccessKey`/`sessionToken` in the returned config. When credentials are provided only via the `credentials` option (provider function), the returned config will have no credentials because FUSE mounts cannot call a provider function. Commit: 1840ce59c9.

KiraPC:
## Re: partial static credentials warning

I verified the original code at commit 42ebd644db — the `this.accessKeyId && this.secretAccessKey` check was already the existing behavior (line 444 in the original file). This PR only renamed the variable from `hasCredentials` to `hasStaticCredentials` and changed the else branch: previously it fell back to anonymous access (empty credentials + no-op signer), now it falls back to the SDK default credential provider chain.

In practice, if someone passes only one of the two static keys, the SDK default chain kicks in and the partial key is simply ignored. The first actual S3 request will then fail with a clear error from the AWS SDK itself (e.g. `InvalidAccessKeyId`, `SignatureDoesNotMatch`), which is more informative than a generic warning from our side. This is consistent with how the AWS SDK handles credential validation — at request time, not at configuration time.

abhiaiyer91 (approved):

KiraPC:
Hi maintainers 👋

I noticed the jobs (`githubIssueManagerWorkflow` and `discordSyncWorkflow`) have been retrying since this morning, which is also burning Actions minutes for no reason.

Them don't seem related to my changes. Looking at the logs of both, `BASE_URL` and the `Authorization: Bearer` token are empty

Is there anything I can do?

Thanks!

KiraPC:
Guys any updates on this?

abhiaiyer91:
Addressed the three CodeRabbit suggestions on `docs/src/content/en/reference/workspace/s3-filesystem.mdx` in 98958314fd:

- Added an install note for `@aws-sdk/credential-providers` next to the `fromNodeProviderChain()` example.
- Documented that provider functions only apply to `S3Filesystem` API calls and aren't preserved when mounting into E2B sandboxes (mount config supports static credentials only).
- Clarified the `sessionToken` parameter description — it's for static temporary credentials; auto-refreshing SSO/AssumeRole/container credentials should go through the `credentials` provider parameter or the SDK default chain.

PR #15454, opened 2026-04-16
fix(core): build step messages after stream completes instead of mid-stream

CalebBarnes:
## Summary

`processOutputStream` previously tried to construct assistant messages on-the-fly during streaming using a state machine that tracked text/reasoning buffers and flushed them eagerly whenever a different chunk type arrived. This was fragile — flush timing depended on chunk interleaving, which varies by provider and by request.

This PR replaces that approach: raw chunks are collected during streaming, then messages are built once per step from the complete chunk sequence.

## What this fixes

- **Duplicate OpenAI item IDs** (`rs_*`, `msg_*`): Interleaved chunks triggered premature flushes creating multiple parts with the same provider item ID, causing "Duplicate item found" errors on subsequent turns
- **Empty text parts**: Empty-string `text-delta` artifacts were flushed as `{ type: 'text', text: '' }` parts, causing Anthropic "text content blocks must be non-empty" errors
- **Wrong provider metadata**: Flush logic used metadata from the *interrupting* chunk instead of the originating `text-start`, misattributing item IDs
- **Interleaved text corruption**: A single global text accumulator mixed deltas from different spans, producing corrupted output like `"aworld!"`
- **Prompt cache instability**: Non-deterministic flush timing meant the same model response could produce different message structures, busting provider prompt caches

## How it works

1. **During stream**: `processOutputStream` collects chunks into an ordered array while handling side effects that must happen inline — tool input streaming callbacks (`onInputStart`/`onInputDelta`), deferred provider tool result patching (e.g. Anthropic `web_search`), response metadata capture, error handling, and abort checks
2. **After stream**: `buildMessagesFromChunks` walks the complete chunk sequence and deterministically builds messages — one part per span ID, correct metadata attribution, empty spans skipped, tool-call + tool-result merged

## What changed

- `llm-execution-step.ts` — Removed all mid-stream message construction (text/reasoning buffer flushing, inline `messageList.add` calls). `processOutputStream` now collects chunks and handles only side effects that require inline execution. Message assembly happens post-stream via `buildMessagesFromChunks`
- `build-messages-from-chunks.ts` — New module that converts collected chunks into `MastraDBMessage`(s)
- `run-state.ts` — Removed dead streaming state fields (`reasoningBuffers`, `textDeltas`, `isReasoning`, `isStreaming`, `hasToolCallStreaming`)

## Test plan

- `build-messages-from-chunks.test.ts`: 25 new tests covering text/reasoning spans, metadata cascading, tool merging, empty filtering, redacted reasoning
- Regression test in `fullStream.ts`: proves duplicate `rs_*` reasoning IDs fail on `main`, pass on this branch
- Existing suites: `loop.test.ts`, `message-list/`, `stream/` — all passing
- Clean `tsc --noEmit`


packages/core/src/loop/workflows/agentic-execution/build-messages-from-chunks.ts
```diff
@@ -0,0 +1,354 @@
...
+        let span = textSpans.get(p.id);
+        // Auto-create span if delta arrives without a matching text-start
+        if (!span) {
+          span = { deltas: [], providerMetadata: p.providerMetadata };
+          textSpans.set(p.id, span);
+        }
+        span.deltas.push(p.text);
+        // AI SDK semantics: latest non-null providerMetadata wins
+        if (p.providerMetadata) {
+          span.providerMetadata = p.providerMetadata;
```
TylerBarnes, line 104: any idea if merging is ever needed? I'm curious what would exist here that gets overwritten

packages/core/src/loop/workflows/agentic-execution/llm-execution-step.ts
```diff
@@ -1181,6 +845,34 @@ export function createLLMExecutionStep<TOOLS extends ToolSet = ToolSet, OUTPUT =
...
+            // are added to messageList, so it can't find the message. We apply it here.
+            const bufferedObject = outputStream._getImmediateObject();
+            if (bufferedObject !== undefined) {
+              const responseMessages = messageList.get.response.db();
+              const lastAssistant = [...responseMessages].reverse().find(m => m.role === 'assistant');
+              if (lastAssistant) {
+                if (!lastAssistant.content.metadata) {
+                  lastAssistant.content.metadata = {};
+                }
+                lastAssistant.content.metadata.structuredOutput = bufferedObject;
```
TylerBarnes, line 873: any reason we shouldn't add the structured output in buildMessagesFromChunks? or even in the builtMessages loop?

TylerBarnes (approved):

abhiaiyer91 (approved):

packages/core/src/loop/workflows/agentic-execution/llm-execution-step.ts
CalebBarnes: Yeah we could add it in buildMessagesFromChunks and collect the object result chunks as well. I only didn't add it there (yet?) because there are also a ton of other things that should be cleaned up too and ended up just deciding to do a bunch more in a follow up PR to clean up / simplify / remove dead code etc.

TylerBarnes: ah fair

packages/core/src/loop/workflows/agentic-execution/build-messages-from-chunks.ts
CalebBarnes: Confirmed — merging is never needed. Checked the AI SDK source (`ai@6.0.116`) and they use the same "latest non-null wins" pattern for both text and reasoning spans:

```js
// text-delta (line 5387)
textPart.providerMetadata = chunk.providerMetadata ?? textPart.providerMetadata;

// text-end (line 5401)
textPart.providerMetadata = chunk.providerMetadata ?? textPart.providerMetadata;

// reasoning-delta (line 5428) & reasoning-end (line 5441) — same pattern
```

Each chunk either provides a complete replacement `providerMetadata` object or nothing. No deep merge, no key-by-key combination. Makes sense since `providerMetadata` is keyed by provider name (e.g. `{ anthropic: { cacheControl: ... } }`) — the provider controls the full shape so a later value represents current state, not an incremental diff.

PR #15473, opened 2026-04-17
fix(core): propagate tracingContext through built-in subagent tool

### LINKED ISSUE #15461: [BUG] Built-in subagent tool drops tracingContext, breaking nested spans in Langfuse/OTel.
## Summary

When a Harness agent delegates work through the built-in `subagent` tool,
the ephemeral subagent's activity is **not traced** in any OTel exporter
(verified with Langfuse). The parent agent's run creates a proper `agent_run`
span, but the child subagent produces no spans at all — breaking the
expected parent → child relationship in the trace UI.

Root cause: `createSubagentTool` forwards every other contextual field into
the ephemeral agent's `.stream(...)` call except `tracingContext`.

## Affected version

- `@mastra/core@1.17.0` (and earlier 1.x releases with harness)
- Confirmed via `dist/harness/index.js`

## Location

`packages/core/src/harness/tools.ts` → `createSubagentTool`
(in the built package: `dist/harness/index.js` line ~334).

The `subagent.stream(task, { ... })` call forwards:

- `maxSteps`, `stopWhen`
- `abortSignal`
- `requireToolApproval`
- `requestContext`  ← propagated
- `prepareStep`

but omits `tracingContext: context?.tracingContext`.

```js
// Current — inside createSubagentTool.execute
const response = await subagent.stream(task, {
  maxSteps: definition.maxSteps ?? (definition.stopWhen ? void 0 : 50),
  stopWhen: definition.stopWhen,
  abortSignal,
  requireToolApproval: false,
  requestContext: context?.requestContext,
  prepareStep: /* ... */,
  // ❌ tracingContext missing
});

### Steps To Reproduce

1. Configure Mastra with Observability + LangfuseExporter.
2. Register a parent agent on the Mastra instance so it gets the observability reference.
3. Build a Harness with the parent agent and at least one subagent (e.g. a "worker").
4. Send a message that causes the parent to call the built-in subagent tool.
5. Open the resulting trace in Langfuse

I currently cannot provide the project I am working on that.- But it is a single line of code, you can create a simple agent with at least one subagent to delegate from and trace the result.

### Expected Behavior

Expected: agent_run (parent) → tool (subagent) → agent_run (child subagent) → ...

Actual: agent_run (parent) → tool (subagent) — the child agent_run span and everything under it is missing.

### Environment Information

```block
@mastra/core: 1.17.0
@mastra/langfuse: ^1.0
@mastra/observability: ^1.5.1
Node: 22.x
OS: macOS / Linux (docker)
```

### Verification

intojhanurag:
## Description

The built-in `subagent` tool dropped the parent agent run's `tracingContext` when invoking the inner subagent, causing the subagent's `agent_run`, `invoke_agent`, and `model_step` spans to surface as a separate orphan trace in observability backends (Langfuse, etc.) instead of nesting under the parent. This PR forwards the tool's `tracingContext` to `subagent.stream()` so the full call tree stays linked in a single trace.

- Forward `context.tracingContext` to `subagent.stream()` in `createSubagentTool`, mirroring the existing pattern used elsewhere in `harness.ts`
- Add unit tests in `subagent-tool.test.ts` covering both the forwarded and the absent `tracingContext` cases
- Verified end-to-end against Langfuse Cloud: with the fix, the parent agent run trace contains the nested `invoke_agent → chat → model_step → model_chunk` chain; without it, those spans are orphaned to a separate root trace

## Before
image

## After
image

## Related Issue(s)

Fixes #15461

## Type of Change

- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [ ] Test update

## Checklist

## ELI5

When a parent agent asked a smaller agent to do work, the system that records the whole process lost the connection between them. This change makes sure the child agent keeps the parent's tracing information so the whole process appears together in monitoring tools.

## Overview

Fixes a bug where the built-in subagent tool omitted the parent agent run's tracingContext when invoking a nested subagent, causing child spans (agent_run, invoke_agent, model_step) to become orphaned in observability backends (e.g., Langfuse, OTel).

## Changes

- packages/core/src/harness/tools.ts
  - createSubagentTool now conditionally forwards context.tracingContext into subagent.stream(...) alongside other forwarded options (abortSignal, requireToolApproval, requestContext, prepareStep).
- packages/core/src/harness/subagent-tool.test.ts
  - Added tests verifying that tracingContext is forwarded when present and omitted entirely when not present.
  - Includes beforeEach/afterEach hooks to manage mocks.
- .changeset/eager-regions-taste.md
  - New changeset entry to release a patch for @mastra/core documenting the fix.

## Verification

- Unit tests added to cover both presence and absence of tracingContext.
- End-to-end verification against Langfuse Cloud shows the nested invoke_agent → chat → model_step → model_chunk chain is now nested under the parent agent_run instead of appearing as an orphan trace.

## Impact

Restores expected parent→child trace relationships for ephemeral subagent activity in OTel exporters. Fixes issue #15461. Classified as a bug fix with tests added.

### DIFF
--- a/.changeset/eager-regions-taste.md
+++ b/.changeset/eager-regions-taste.md
@@ -1,0 +1,5 @@
+---
+'@mastra/core': patch
+---
+
+Fixed subagent traces being orphaned from the parent agent's trace when the built-in `subagent` tool is invoked from inside an agent run. Subagent spans now nest correctly under the parent agent's trace in observability backends like Langfuse. Fixes #15461.
--- a/.changeset/eager-regions-taste.md
+++ b/.changeset/eager-regions-taste.md
@@ -2,4 +2,4 @@
 '@mastra/core': patch
 ---
 
-Fixed subagent traces being orphaned from the parent agent's trace when the built-in `subagent` tool is invoked from inside an agent run. Subagent spans now nest correctly under the parent agent's trace in observability backends like Langfuse. Fixes #15461.
+Fixed Harness subagent tracing so delegated runs keep the parent tracing context and show up in the same trace in observability exporters. Fixes #15461.
--- a/packages/core/src/harness/subagent-tool.test.ts
+++ b/packages/core/src/harness/subagent-tool.test.ts
@@ -241,6 +241,63 @@ describe('createSubagentTool requestContext forwarding', () => {
   });
 });
 
+describe('createSubagentTool tracingContext forwarding', () => {
+  beforeEach(() => {
+    vi.clearAllMocks();
+  });
+
+  afterEach(() => {
+    vi.restoreAllMocks();
+  });
+
+  it('forwards tracingContext to subagent.stream() when provided', async () => {
+    mockStream.mockResolvedValue(createMockStreamResponse('done'));
+
+    const tool = createSubagentTool({
+      subagents,
+      resolveModel,
+      fallbackModelId: 'test-model',
+    });
+
+    const mockSpan = { spanContext: () => ({ traceId: 'abc', spanId: 'def' }) };
+    const tracingContext = { currentSpan: mockSpan } as any;
+
+    const requestContext = new RequestContext();
+    requestContext.set('harness', { emitEvent: vi.fn() });
+
+    await (tool as any).execute(
+      { agentType: 'explore', task: 'Investigate' },
+      { requestContext, tracingContext, agent: { toolCallId: 'tc-trace-1' } },
+    );
+
+    expect(mockStream).toHaveBeenCalledTimes(1);
+    const streamOpts = mockStream.mock.calls[0]![1];
+    expect(streamOpts.tracingContext).toBe(tracingContext);
+  });
+
+  it('does not include tracingContext when not provided', async () => {
+    mockStream.mockResolvedValue(createMockStreamResponse('done'));
+
+    const tool = createSubagentTool({
+      subagents,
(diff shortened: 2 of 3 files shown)


tripplen23:
Hi, this looks good, several evidence of testing on this:

1. After wire up this patch:
image


2. Before that:
image

3. General trace, down one is before and the upper one is after:
image

.changeset/eager-regions-taste.md
```diff
@@ -0,0 +1,5 @@
+---
+'@mastra/core': patch
+---
+
+Fixed subagent traces being orphaned from the parent agent's trace when the built-in `subagent` tool is invoked from inside an agent run. Subagent spans now nest correctly under the parent agent's trace in observability backends like Langfuse. Fixes #15461.
```
tripplen23: may be make it shorter and less noise like:

`Fixed Harness subagent tracing so delegated runs keep the parent tracing context and show up in the same trace in observability exporters. Fixes #15461.` - Wdyt?

tripplen23 (approved):
other things look good to me and tested well so, you can consider just only my comment above.

.changeset/eager-regions-taste.md
intojhanurag: Fixed ! Thanks :)

epinzur (approved):

PR #15487, opened 2026-04-18
Skip heavy field serialization for filtered spans

epinzur:
## Description

Optimize span serialization by skipping heavy payload fields (attributes, input, output, errorInfo, requestContext) for spans that will be filtered out before export. Previously, these fields were serialized and retained for the lifetime of the span even though they would never be exported.

The optimization applies when:
- `excludeSpanTypes` drops the span type
- The span is internal and `includeInternalSpans` is false
- The span is a NoOpSpan (never exported)

Metadata is still attached and deep-cleaned because it is read in-process by `getCorrelationContext()` and by `getLoggerContext()`/`getMetricsContext()`, and is inherited by non-filtered child spans.

### Changes

- **BaseSpan**: Add `isExcluded` flag determined at construction time based on config. Skip attaching heavy fields when excluded, but always attach and deep-clean metadata.
- **DefaultSpan**: Reorder `end()` and `update()` to apply metadata updates before checking `isExcluded`, ensuring metadata is always current for in-process readers.
- **NoOpSpan**: Override `alwaysExcluded` to return true since NoOpSpans are never exported.
- **Tests**: Add comprehensive test suite covering excluded span behavior, metadata inheritance, and field attachment on included spans.

## Type of Change

- [ ] Documentation update
- [x] Code refactoring
- [x] Performance improvement
- [x] Test update

## Checklist

https://claude.ai/code/session_01K9wGzZ44BhT8CZcfKaeGng

## ELI5 Explanation
The PR stops doing expensive work for trace spans that will be ignored anyway—filtered spans no longer get bulky payloads serialized—while still keeping small metadata the app needs for in-process uses. This reduces CPU and memory overhead without changing exported data.

## Overview
Avoid attaching and deep-cleaning heavy payload fields (attributes, input, output, errorInfo, requestContext) for spans that will be filtered out before export. Short-circuiting occurs when a span matches configured `excludeSpanTypes`, when an internal span is created while `includeInternalSpans` is false, or when the span is a NoOpSpan. Metadata is still deep-cleaned and attached because it’s consumed in-process (correlation, logging, metrics) and is inherited by child spans.

### BaseSpan
- Adds an `isExcluded` flag computed at construction from `alwaysExcluded`, `excludeSpanTypes`, and internal-span settings.
- If `isExcluded` is true, the constructor returns early after setting `attributes = {}`, leaving heavy fields (`requestContext`, `input`, `output`, `errorInfo`) unset so they are never deep-cleaned or serialized.
- `metadata` is still deep-cleaned and attached.
- Ensures `isInternal` is computed before determining exclusion.

### DefaultSpan
- Reorders operations so metadata merges/updates occur before the exclusion check:
  - `update()` and `end()` now apply metadata first, then short-circuit if excluded, preventing updates to `input`, `output`, and `attributes` for excluded spans.
  - `error()` applies metadata unconditionally but only computes `errorInfo` and merges attributes when not excluded.

### NoOpSpan
- Overrides `alwaysExcluded` to return `true`, making NoOpSpan always treated as excluded (so it never attaches heavy fields).

### Tests
- Adds tests verifying:
  - Excluded spans do not attach heavy fields and have `attributes = {}`, while still exposing and inheriting `metadata`.
  - Internal-span behavior under `includeInternalSpans: false` (no `input`) and `includeInternalSpans: true` (`input` attached and deep-cleaned).
  - Non-excluded spans still have deep-clean applied to payloads (e.g., functions become "[Function]").

## Impact
- No public API or exported span-data changes; exporters and custom processors remain unchanged for spans that are exported.
- Reduces CPU and memory overhead by avoiding unnecessary serialization and retention for filtered spans, improving performance in high-throughput tracing.
- In-process readers (correlation, logging, metrics) continue to observe up-to-date metadata.


intojhanurag:
LGTM

intojhanurag (approved):

PR #15505, opened 2026-04-19
fix(memory): propagate Vertex thought signatures for OM split tool calls

### LINKED ISSUE #15294: [BUG] Observational Memory causes Gemini 3 thought_signature error on multi-step parallel tool calls
## Describe the Bug

When using a Mastra Agent with Observational Memory enabled and Gemini 3 Flash Preview via `@ai-sdk/google-vertex`, multi-step conversations with parallel tool calls fail with Vertex AI's `thought_signature` validation error (400):

```
Unable to submit request because function call `query_data` in the 5. content block
is missing a `thought_signature`
```

Gemini streaming returns `thoughtSignature` only on the first parallel function call part (known behavior — see [n8n fix](https://github.com/n8n-io/n8n/commit/3dbca3e), [google/adk-python#3705](https://github.com/google/adk-python/issues/3705)). Without OM, the AI SDK's step loop keeps all parallel tool calls in one content block, so at least one has the signature and Vertex accepts the request. With OM, the `processInputStep` runs between steps and reconstructs the message list, splitting tool calls into separate assistant messages. Each becomes its own content block — those without `thoughtSignature` are rejected.

**Disabling Observational Memory eliminates the error entirely.** Same agent, same prompt, same model — no OM → completes successfully with text output.

## Steps To Reproduce

1. Set up env vars: `GOOGLE_VERTEX_PROJECT`, `GOOGLE_VERTEX_LOCATION`, `GOOGLE_VERTEX_CREDENTIALS_BASE64`, `DATABASE_URL` (PostgreSQL)
2. Install deps: `@mastra/core@1.23.0`, `@mastra/memory@1.14.0` (also fails on `1.15.0`), `@mastra/pg@1.8.5`, `@ai-sdk/google-vertex@4.0.95`, `ai@6.0.116`, `zod`
3. Run the reproduction script below — stderr will show the 400 error

```javascript
import { randomUUID } from "node:crypto"
import { createVertex } from "@ai-sdk/google-vertex"
import { Agent } from "@mastra/core/agent"
import { Memory } from "@mastra/memory"
import { PgVector, PostgresStore } from "@mastra/pg"
import { tool } from "ai"
import { z } from "zod"

const PROJECT = process.env.GOOGLE_VERTEX_PROJECT
const LOCATION = process.env.GOOGLE_VERTEX_LOCATION
const CREDS_B64 = process.env.GOOGLE_VERTEX_CREDENTIALS_BASE64
const DB_URL = process.env.DATABASE_URL

const creds = JSON.parse(Buffer.from(CREDS_B64, "base64").toString("utf-8"))
const provider = createVertex({ project: PROJECT, location: LOCATION, googleAuthOptions: { credentials: creds } })
const MODEL = "gemini-3-flash-preview"

// Large tool results (~2000 tokens each) trigger OM observer between steps
function makeLargeResult(id) {
  return {
    id,
    total_rows: 30,
    data: Array.from({ length: 30 }, (_, i) => ({
      row_id: `${id}-row-${i}`,
      name: `Record ${id} item ${i} with detailed description and metadata`,
      status: i % 3 === 0 ? "active" : i % 3 === 1 ? "pending" : "inactive",
      amount: Math.floor(Math.random() * 10000),
      notes: `Detailed notes for item ${i} in record ${id}. Additional context to trigger observational memory buffering.`,
    })),
  }
}

const memory = new Memory({
  storage: new PostgresStore({ id: "s", connectionString: DB_URL, schemaName: "mastra_thought_sig_repro" }),
  vector: new PgVector({ id: "v", connectionString: DB_URL, schemaName: "mastra_thought_sig_repro" }),
  embedder: provider.textEmbeddingModel("text-embedding-005"),
  options: {
    observationalMemory: { enabled: true, model: provider(MODEL), scope: "thread" },
    semanticRecall: false,
    workingMemory: { enabled: false },
  },
})

let callCount = 0
const agent = new Agent({
  name: "repro",
  instructions: "Follow instructions precisely. Use the provided tools. Call tools in parallel when instructed.",
  model: provider(MODEL),
  tools: {
    query_data: tool({
      description: "Query data by ID. Returns a large dataset. Call in parallel when multiple IDs are needed.",
      parameters: z.object({ id: z.string().describe("The record ID to query"), _pad: z.string().optional() }),
      execute: async ({ id }) => {
        callCount++
        console.log(`  [tool] id=${id} (#${callCount})`)
        return makeLargeResult(id)
      },
    }),
  },
  memory,
})

// Multi-step: 1 call -> 2 calls -> 8 parallel calls
const stream = await agent.stream(
  [{ role: "user", content: `Follow these steps EXACTLY:
Step 1: Call query_data with id="index-001".
Step 2: Call query_data for BOTH "rec-A" and "rec-B".
Step 3: Call query_data for ALL IN PARALLEL: det-1, det-2, det-3, det-4, det-5, det-6, det-7, det-8
After all results, write a summary. Each step must wait for previous results.` }],
  {
    maxSteps: 15,
    memory: { thread: `repro-${randomUUID()}`, resource: "repro-resource" },
    providerOptions: { vertex: { thinkingConfig: { includeThoughts: true } } },
  },
)

let text = ""
for await (const chunk of stream.textStream) text += chunk
console.log(`\nDone: text=${text.length}ch, toolCalls=${callCount}`)
if (text.length === 0) console.log("text is empty -- check stderr for thought_signature error")
```

The script above is self-contained. Save as `repro.mjs` in a project with the listed dependencies installed, set env vars, and run `node repro.mjs`.

## Expected Behavior

The agent should complete all 11 tool calls and return a text summary. Disabling Observational Memory (removing the `memory` option) produces this expected behavior with the exact same configuration.

## Environment Information

```
@mastra/core: 1.23.0
@mastra/memory: 1.14.0 (also tested 1.15.0 -- same issue)
@mastra/pg: 1.8.5
@ai-sdk/google-vertex: 4.0.95
@ai-sdk/google: 3.0.53
ai: 6.0.116
Node.js: v24.14.1
OS: Linux (Manjaro)
LLM Provider: Google Vertex AI (gemini-3-flash-preview)
```

## Additional Context

- Gemini streaming returns `thoughtSignature` only on the first parallel function call -- known across SDKs
- Without OM: AI SDK keeps parallel calls in one content block -> partial signatures OK -> Vertex accepts
- With OM: `processInputStep` splits tool calls into individual messages -> individual content blocks -> blocks with zero signatures -> Vertex rejects (400)
- Observed 12 occurrences in production over 22 hours across 6 different agents, all with OM active

h30s:
## Description

Observational Memory can split parallel Gemini tool calls into separate assistant messages. Vertex only attaches a `thought_signature` to the first call in a batch, so when those calls end up in different blocks the API rejects the next request. After each OM input step we now reuse the last signature we saw and attach it to any tool-invocation parts that are still missing one, so the history stays valid for thinking mode.

## Related Issue(s)

Fixes #15294

## Type of Change

- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [ ] Test update

## Checklist

## ELI5
Observational Memory sometimes splits multiple tool calls into separate assistant messages but Gemini only gives a special "thought signature" to the first one; later pieces get rejected. This PR copies the last-seen thought signature onto subsequent assistant tool-invocation parts (until a new user message), so Vertex accepts split parallel calls.

## Overview
Fixes a bug where Observational Memory (OM) reconstruction of Gemini/Vertex streaming could produce assistant content blocks for parallel tool-invocations that lack the required Vertex `thought_signature`, causing Vertex 400 validation errors (issue #15294). The fix propagates the most recently observed thought signature across assistant message parts within the same assistant turn so multi-step parallel tool-call workflows validate correctly.

## Changes
- Added propagateVertexThoughtSignaturesToToolInvocations(messages: MastraDBMessage[]) in packages/memory/src/processors/observational-memory/vertex-thought-signature.ts.
  - Scans messages, resets its last-known signature on every user message, records signatures from assistant parts (Vertex then Google fallback), and writes the last-known signature into tool-invocation parts that lack one (mutating messages in place).
- Invoke propagation inside ObservationalMemoryProcessor.processInputStep (packages/memory/src/processors/observational-memory/processor.ts) immediately after persisting token counts and before repro-capture/other repro logic so OM-updated message objects include signatures before further processing.
- Added tests:
  - Unit tests (vertex-thought-signature.test.ts) covering propagation behavior, Google fallback, respecting user-turn boundaries (no cross-turn propagation), and not overwriting existing signatures.
  - Integration test (vertex-thought-signature-om.integration.test.ts) asserting processInputStep performs propagation on OM-split assistant messages.
- Changeset entry (.changeset/metal-mirrors-design.md) bumping @mastra/memory patch with a short note describing the fix.

## Impact
Resolves #15294 — Observational Memory no longer produces Vertex validation errors for multi-step parallel tool calls on Gemini models. Classified as a bug fix; includes tests proving the behavior.

## Notes
- The propagation deliberately stops at user message boundaries (scope limited to assistant-run segments following user turns).
- Documentation was not updated in this PR; some reviewer comments (coderabbit) were not fully addressed.

### DIFF
--- a/.changeset/metal-mirrors-design.md
+++ b/.changeset/metal-mirrors-design.md
@@ -1,0 +1,5 @@
+---
+'@mastra/memory': patch
+---
+
+Fixed Observational Memory triggering Gemini Vertex thought signature validation errors when parallel tool calls were stored as separate assistant messages. Tool invocations now receive the last known thought signature so multi-step runs with thinking enabled complete successfully.
--- a/packages/memory/src/processors/observational-memory/__tests__/vertex-thought-signature-om.integration.test.ts
+++ b/packages/memory/src/processors/observational-memory/__tests__/vertex-thought-signature-om.integration.test.ts
@@ -29,7 +29,8 @@ function createStubMemoryProvider(): MemoryContextProvider {
   };
 }
 
-function toolResultPart(
+/** Mirrors OM-split parallel calls: still in `call` before results are merged. */
+function toolInvocationPart(
   callId: string,
   toolName: string,
   providerMetadata?: { vertex?: { thoughtSignature?: string }; google?: { thoughtSignature?: string } },
@@ -37,11 +38,10 @@ function toolResultPart(
   return {
     type: 'tool-invocation',
     toolInvocation: {
-      state: 'result',
+      state: 'call',
       toolCallId: callId,
       toolName,
       args: {},
-      result: { ok: true },
     },
     ...(providerMetadata ? { providerMetadata } : {}),
   } as MastraDBMessage['content']['parts'][number];
@@ -92,7 +92,7 @@ describe('ObservationalMemoryProcessor + Vertex thought signatures (#15294)', ()
       resourceId,
       content: {
         format: 2,
-        parts: [toolResultPart('c1', 'query_data', { vertex: { thoughtSignature: 'sig-from-stream' } })],
+        parts: [toolInvocationPart('c1', 'query_data', { vertex: { thoughtSignature: 'sig-from-stream' } })],
       },
     };
 
@@ -104,7 +104,7 @@ describe('ObservationalMemoryProcessor + Vertex thought signatures (#15294)', ()
       resourceId,
       content: {
         format: 2,
-        parts: [toolResultPart('c2', 'query_data')],
+        parts: [toolInvocationPart('c2', 'query_data')],
       },
     };
 
--- a/packages/memory/src/processors/observational-memory/__tests__/vertex-thought-signature-om.integration.test.ts
+++ b/packages/memory/src/processors/observational-memory/__tests__/vertex-thought-signature-om.integration.test.ts
@@ -1,0 +1,144 @@
+/**
+ * Ensures #15294 fix runs inside processInputStep (not only the standalone helper).
+ * Live Vertex/Gemini is not invoked — credentials are not available in CI.
+ */
+import type { MastraDBMessage } from '@mastra/core/agent';
+import { InMemoryMemory, InMemoryDB } from '@mastra/core/storage';
+import { describe, it, expect, beforeEach, vi } from 'vitest';
+
+import { ObservationalMemory } from '../observational-memory';
+import { ObservationalMemoryProcessor } from '../processor';
+import type { MemoryContextProvider } from '../processor';
+
+function createInMemoryStorage(): InMemoryMemory {
+  const db = new InMemoryDB();
+  return new InMemoryMemory({ db });
+}
+
+function createStubMemoryProvider(): MemoryContextProvider {
+  return {
+    getContext: vi.fn().mockResolvedValue({
+      systemMessage: undefined,
+      messages: [],
+      hasObservations: false,
+      omRecord: null,
+      continuationMessage: undefined,
+      otherThreadsContext: undefined,
+    }),
+    persistMessages: vi.fn().mockResolvedValue(undefined),
+  };
+}
+
+function toolResultPart(
+  callId: string,
+  toolName: string,
+  providerMetadata?: { vertex?: { thoughtSignature?: string }; google?: { thoughtSignature?: string } },
+): MastraDBMessage['content']['parts'][number] {
+  return {
+    type: 'tool-invocation',
+    toolInvocation: {
+      state: 'result',
+      toolCallId: callId,
+      toolName,
+      args: {},
+      result: { ok: true },
+    },
+    ...(providerMetadata ? { providerMetadata } : {}),
+  } as MastraDBMessage['content']['parts'][number];
+}
+
+describe('ObservationalMemoryProcessor + Vertex thought signatures (#15294)', () => {
+  const threadId = 'thread-15294';
+  const resourceId = 'resource-15294';
+
+  let om: ObservationalMemory;
+  let processor: ObservationalMemoryProcessor;
+
+  beforeEach(() => {
+    const storage = createInMemoryStorage();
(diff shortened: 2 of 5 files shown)


h30s:
@abhiaiyer91 could you please review when you get a chance?

TylerBarnes (changes requested):
From Mastra Code: I don’t think this should live inside ObservationalMemoryProcessor.

This looks like a provider history compatibility issue, not OM behavior itself. OM may expose it by splitting parallel tool calls into separate assistant messages, but the invariant is that when Mastra replays stored/split assistant tool-call history to Gemini Vertex, the outbound provider prompt needs to satisfy Vertex’s thoughtSignature requirements.

We now have `ProviderHistoryCompat` in `packages/core/src/processors/provider-history-compat.ts` for exactly this class of issue. It already supports preemptive outbound-only prompt rewrites via `CompatRule.applyToPrompt`, so persisted memory/UI/observability history stays unchanged while the provider-specific prompt is repaired before the request.

Could you move this into a Gemini/Vertex-scoped compat rule there instead of mutating `messageList.get.all.db()` from OM? A regression test should build split assistant tool-call history and verify the `ProviderHistoryCompat.processLLMRequest` output includes the required thoughtSignature metadata for the Vertex/Gemini prompt.

roaminro:
Hey @h30s ! Could you have a look at the review left by Tyler?

TylerBarnes:
Going to close this! The PR is valuable but please re-open it as a provider compat processor feature. We shouldn't add any provider specific code into memory or core packages except in the compat processor

PR #15548, opened 2026-04-20
feat(server): add gateway memory support to agent responses API

TheIsrael1:
This updates the server-side Responses and Conversations adapters so gateway-backed agents follow the same stored-response flow as local memory. Initial turns work with `store: true`, follow-up turns work with `previous_response_id`, and direct thread access works with `conversation_id`. The server resolves gateway thread and resource context from stored messages, known conversations, or `requestContext`

It also fixes gateway-backed follow-up turns to return the persisted assistant message ID instead of a synthetic route ID, so `previous_response_id` can be reused reliably on the next turn.

```ts
const first = await client.responses.create({
  agent_id: 'gateway-agent',
  input: 'Hello',
  store: true,
})

const second = await client.responses.create({
  previous_response_id: first.id,
  input: 'Continue',
})
```

Conversation inspection and deletion continue to work through `conversation_id`, and explicit resource scoping still comes from `requestContext` when needed.

## ELI5 Summary

This PR makes remote ("gateway") agents behave like local agents for conversations and responses: you can store an assistant reply, refer to it later for follow-ups, and inspect or delete threads without putting resource IDs into every request—resource scoping is handled via requestContext.

## Overview

Adds gateway-backed memory support so Responses and Conversations routes treat gateway agents the same as local-memory agents. The server resolves thread and resource context from stored messages, known conversations, or requestContext; callers no longer need to include resource_id in public payloads. Follow-up turns using previous_response_id now return the persisted assistant message ID (not a synthetic route ID), enabling reliable reuse.

## Key Changes

- Server handlers
  - Conversations: POST now creates gateway threads for gateway-backed agents (requires resource_id when creating), GET/GET items/DELETE now fall back to gateway thread lookup and mapping when local threads are not found; supports cursor-like listing of gateway messages.
  - Responses: Create flow recognizes gateway-backed agents and skips local-memory-only checks; resolveThreadExecutionContext accepts resourceId and isGatewayBacked and enforces conversation_id + resource_id for direct gateway thread context; GET/DELETE response routes can lookup or delete gateway messages when DB lookup misses.

- Gateway client
  - Added GatewayMemoryClient.deleteMessages(threadId, { messageIds }) to delete specific messages.
  - toLocalMessage now returns MastraDBMessage, coerces roles, and normalizes string content into structured content objects.

- Schemas & types
  - Added conversationLookupQuerySchema (agent_id?, resource_id?).
  - Added responseLookupQuerySchema (agent_id?, conversation_id?, resource_id?).
  - createResponseBodySchema now accepts optional resource_id.
  - Removed resource_id from client-js CreateConversationParams (client API surface).

- Tests
  - Added gateway-backed tests for Conversations (create, get, list items, delete) and Responses (create validation, create with memory.thread+resource, get/delete mapping and deletion), plus test scaffolding for gateway client and Mastra instance creation; restored vi mocks between tests.

- Docs & changesets
  - Documentation updated to state gateway-backed agents follow stored-response patterns and to recommend requestContext for resource scoping; changesets added for server/client packages reflecting these changes.

## Notable behavior / compatibility notes

- Callers no longer need to supply resource_id in public Conversations/Responses payloads; explicit scoping remains available via requestContext.
- For gateway-backed direct lookups or deletions, conversation_id and resource_id are required and validated.
- Follow-up turns using previous_response_id now reliably reference persisted assistant message IDs for gateway-backed agents.


TheIsrael1:
@copilot resolve the merge conflicts in this pull request

Issue #15517, opened 2026-04-20
Agent editor fail when adding skills to the workspace


daneatmastra:
This issue was created from Discord post 1495733103128477756:

[![Open in Discord](https://img.shields.io/badge/Open_in_Discord-5865F2?style=for-the-badge&logo=discord&logoColor=white)](https://discord.com/channels/1309558646228779139/1495733103128477756)

I am suing the default weather app example from the init project. when i deploy mastra studio to the platform or self hosted (same behaviour) if I add a skill to the workspace(any) when I try to open the editor in an agent it fails with this error message
 Unexpected Application Error!
Cannot convert undefined or null to object
TypeError: Cannot convert undefined or null to object
    at Object.entries (<anonymous>)
    at hSn (https://mastra-local.7vvsz4jtykzrg.eu-west-2.cs.amazonlightsail.com/assets/main-BY2pdxL7.js:512:61364)
    at vSn (https://mastra-local.7vvsz4jtykzrg.eu-west-2.cs.amazonlightsail.com/assets/main-BY2pdxL7.js:516:3567)
    at https://mastra-local.7vvsz4jtykzrg.eu-west-2.cs.amazonlightsail.com/assets/main-BY2pdxL7.js:516:3922
    at Object.useMemo (https://mastra-local.7vvsz4jtykzrg.eu-west-2.cs.amazonlightsail.com/assets/main-BY2pdxL7.js:9:61846)
    at NOt.gn.useMemo (https://mastra-local.7vvsz4jtykzrg.eu-west-2.cs.amazonlightsail.com/assets/main-BY2pdxL7.js:2:8354)
    at G9 (https://mastra-local.7vvsz4jtykzrg.eu-west-2.cs.amazonlightsail.com/assets/main-BY2pdxL7.js:516:3908)
    at Okn (https://mastra-local.7vvsz4jtykzrg.eu-west-2.cs.amazonlightsail.com/assets/main-BY2pdxL7.js:520:192784)
    at q8 (https://mastra-local.7vvsz4jtykzrg.eu-west-2.cs.amazonlightsail.com/assets/main-BY2pdxL7.js:9:49083)
    at uB (https://mastra-local.7vvsz4jtykzrg.eu-west-2.cs.amazonlightsail.com/assets/main-BY2pdxL7.js:9:72200)
💿 Hey developer 👋

You can provide a way better UX than this when your app throws errors by providing your own ErrorBoundary or errorElement prop on your route.

PR #15559, opened 2026-04-20
fix: agent editor skills bug

mikhael28:
## Description

Fixed a crash in the Agent Editor when opening a code-defined agent whose `Workspace` discovers skills. Addresses https://github.com/mastra-ai/mastra/issues/15517 

Previously the editor threw `TypeError: Cannot convert undefined or null to object` because workspace-discovered skill metadata (`SkillMetadata[]`) was being fed into the stored-skill-config form mapper, which expects a `Record<string, StoredAgentSkillConfig>`. Code-defined agents don't carry stored skill overrides, so this field is now left unset when loading them into the edit form.

Screenshot 2026-04-20 at 11 59 17 AM

You can recreate it by going in the monorepo, installing the `agent` example, adding a skills attribute to the workspace, adding a skill to the chef-v2 agent and then opening the Agent Editor for that agent.

## ELI5 (Explain Like I'm 5)

The Agent Editor crashed when opening certain agents because it was given the wrong kind of skill data. The fix makes the editor skip passing that wrong data so it loads normally.

---

## Changes Summary

This PR fixes a crash in the Agent Editor that occurred when opening code-defined agents whose Workspace discovers skills. The root cause was workspace-discovered skill metadata (SkillMetadata[]) being passed into a form mapper that expects a Record<string, StoredAgentSkillConfig>, causing `TypeError: Cannot convert undefined or null to object`. When loading code-defined agents into the edit form, the stored-skill-overrides field is now left unset, preventing the mapper from receiving the wrong shape.

### Key Changes

- packages/playground/src/domains/agents/utils/compute-agent-initial-values.ts
  - Updated `mapAgentResponseToDataSource` to stop mapping `agent.skills` into the edit form data and explicitly set `skills: undefined` for code-defined agents so the stored-skill-config mapper is not given workspace-discovered SkillMetadata.

- examples/agent/src/mastra/agents/model-v2-agent.ts
  - Added `skills: ['.agents/skills']` to the Workspace configuration (example change to reproduce/verify behavior).

- examples/agent/src/mastra/workflows/content-moderation.ts
  - Updated workflow step schemas to use `ProcessorStepOutputSchema` for both `inputSchema` and `outputSchema` and cast `.commit()` results to `ProcessorWorkflow` (type-level adjustments).

- .changeset/fix-agent-editor-workspace-skills-crash.md
  - Added a changeset entry documenting the crash fix at patch level for the mastra package.

### Type of Change

Bug fix

### Notes

Documentation updates, tests, and responses to CodeRabbit comments remain open.


abhiaiyer91 (approved):

cyajae:
+1

PR #15317, opened 2026-04-13
feat(core): unify versions, evals, and observability

DanielSLew:
Closes the loop between agent versions, experiments, traces, scores, and feedback so observability becomes the single source of truth for runtime data.

Experiments now actually apply the specified `agentVersion` during execution — previously it was stored as metadata but the agent ran with its default config. `experimentId` is propagated to agent spans, and scorers receive full correlation context so scores carry experiment info.

Added `entityVersionId` as a first-class observability dimension on all records (spans, metrics, scores, feedback, logs). This replaces the old `resolvedVersionId` buried in span attributes — it's now filterable and usable as a `groupBy` key in OLAP queries. Designed to work for any versioned entity type (agents today, workflows/tools in the future).

New query filters: `listExperiments` supports `targetType`, `targetId`, `agentVersion`, `status`. `listExperimentResults` supports `traceId`, `status`.

Playground UI improvements: disabled tabs now show tooltips with docs links explaining how to enable them. Experiment headers link to the agent version in the editor. Trace details show `experimentId` and `entityVersionId` as clickable cross-links.

Storage adapters (DuckDB, ClickHouse, libsql, pg, mongodb) updated with new columns, filters, and row mappings.

Implements: #14930

## ELI5

When you run an experiment, it now actually runs the exact agent version you picked and tags every runtime signal (traces, metrics, scores, feedback, logs) with the producing version and experiment so you can easily link, filter, and group results by version and experiment.

### Core Experiment & Execution
- Experiments apply the specified agentVersion at execution time (resolveTarget supports agentVersion and resolves that exact version).
- executeTarget/executeAgent accept and forward experimentId into agent.generate()/generateLegacy via tracingOptions.metadata so agent spans include experiment context.
- Scorers receive a full targetCorrelationContext (including experimentId, traceId, entity identity) so emitted scores include experiment linkage.

### Observability Model
- Introduces entityVersionId as the canonical, first-class correlation field on spans, metrics, scores, feedback, and logs; adds parentEntityVersionId and rootEntityVersionId for lineage.
- BaseSpan.getCorrelationContext and CorrelationContext now include entityVersionId, parentEntityVersionId, and rootEntityVersionId.
- Legacy resolvedVersionId attribute is still emitted (marked deprecated) for backward compatibility; entityVersionId is filterable and groupable for OLAP queries.
- Record builders and span creation populate the new version fields; tests updated to expect them (nullable when absent).
[...]

PR #15570, opened 2026-04-21
fix(core): prevent dataset experiment hang on workflow targets

### LINKED ISSUE #15453: [BUG] dataset.startExperiment with targetType: 'workflow' hangs forever; step never runs and itemTimeout is ignored
### Describe the Bug

`dataset.startExperiment({ targetType: "workflow", targetId, ... })` hangs indefinitely. The item's workflow step is never executed, `itemTimeout` never fires, and the returned promise never resolves or rejects.

The same workflow works fine in two adjacent call paths:

- **A.** `workflow.createRun().start({ inputData })` directly — succeeds.
- **B.** `dataset.startExperiment({ task: async ({ input }) => workflow.createRun().start(...) })` — succeeds for all items.
- **C.** `dataset.startExperiment({ targetType: "workflow", targetId: "..." })` — hangs forever, even with a trivial echo step (no LLM, no I/O).

Instrumentation points to the gap between `resolveTarget()` and `executeTarget()` in `@mastra/core` 1.25.0:

- `mastra.getWorkflowById(targetId)` **is called** by `resolveTarget` (verified by spying on `Mastra.prototype.getWorkflowById`) and returns the expected workflow instance (identity-equal to `mastra.getWorkflow(targetId)`).
- `workflow.createRun` on that same resolved instance is **never called** (verified by patching `createRun` on the returned ref before starting the experiment).
- `itemTimeout: 15_000` does not abort the hang — the item never enters `execFn` at all, so `raceWithSignal` never gets to race anything.

So something in `runExperiment` / `executeTarget` for `targetType: "workflow"` blocks before `executeWorkflow(target, item)` is invoked, without throwing and without honouring the item timeout.

Workaround: use the `task:` option and call `workflow.createRun().start()` inside it. Works reliably.

### Steps To Reproduce

1. Fresh project, install `@mastra/core@1.25.0`, `@mastra/pg@1.9.1`, `zod@^4`.
2. Point `DATABASE_URL` at a Postgres instance.
3. Run the script below with `tsx`.

```ts
import "dotenv/config";
import { Mastra } from "@mastra/core/mastra";
import { createStep, createWorkflow } from "@mastra/core/workflows";
import { PostgresStore } from "@mastra/pg";
import { z } from "zod";

const inputSchema = z.object({ name: z.string() });
const outputSchema = z.object({ text: z.string() });

const echoStep = createStep({
  id: "echo",
  inputSchema,
  outputSchema,
  execute: async ({ inputData }) => {
    console.log("[step] executing:", inputData.name);
    return { text: `hello ${inputData.name}` };
  },
});

const echoWorkflow = createWorkflow({
  id: "echo-wf",
  inputSchema,
  outputSchema,
}).then(echoStep).commit();

const mastra = new Mastra({
  workflows: { "echo-wf": echoWorkflow },
  storage: new PostgresStore({ id: "repro", connectionString: process.env.DATABASE_URL! }),
});

async function main() {
  const dataset = await mastra.datasets.create({
    name: "repro-" + Date.now(),
    inputSchema,
    groundTruthSchema: z.object({}),
  });
  await dataset.addItems({ items: [{ input: { name: "world" }, groundTruth: {} }] });

  // A — works
  console.log("A) standalone workflow run...");
  const run = await echoWorkflow.createRun({ disableScorers: true });
  const r: any = await run.start({ inputData: { name: "world" } });
  console.log("   status:", r.status); // "success"

  // B — works
  console.log("B) experiment with inline task...");
  const s1 = await dataset.startExperiment({
    task: async ({ input }: any) => {
      const run2 = await echoWorkflow.createRun({ disableScorers: true });
      const rr: any = await run2.start({ inputData: input });
      return rr.result;
    },
    scorers: [],
    maxConcurrency: 1,
    itemTimeout: 15_000,
  });
  console.log("   status:", s1.status, "ok:", s1.succeededCount); // completed, 1

  // C — hangs
  console.log("C) experiment with targetType: 'workflow'...");
  const t0 = Date.now();
  const hb = setInterval(() => console.log(`   [${Math.round((Date.now() - t0) / 1000)}s] still running...`), 3000);
  hb.unref();
  const s2 = await dataset.startExperiment({
    targetType: "workflow",
    targetId: "echo-wf",
    scorers: [],
    maxConcurrency: 1,
    itemTimeout: 15_000,
  });
  console.log("   status:", s2.status, "ok:", s2.succeededCount); // never reached
}

main().then(() => process.exit(0)).catch((e) => { console.error(e); process.exit(1); });
```

Output:

```
A) standalone workflow run...
[step] executing: world
   status: success
B) experiment with inline task...
[step] executing: world
   status: completed ok: 1
C) experiment with targetType: 'workflow'...
   [3s] still running...
   [6s] still running...
   ... (forever — itemTimeout is ignored, step never runs)
```

The script above is a complete standalone reproduction. Happy to push it to a public repo if needed — the full file is self-contained and requires only `@mastra/core`, `@mastra/pg`, `zod`, `dotenv`, and a Postgres URL.

### Expected Behavior

`startExperiment({ targetType: "workflow", targetId })` should resolve the workflow via `mastra.getWorkflowById(targetId)` and, for each dataset item, call `workflow.createRun().start({ inputData: item.input })` — the same behaviour as the `task:` path that calls the workflow manually.

At minimum, `itemTimeout` should abort the hang and surface a failure instead of leaving the promise pending forever.

### Environment Information

```
System:
  OS: macOS 26.3.1
  CPU: (10) arm64 Apple M1 Max
  Shell: zsh 5.9
Binaries:
  Node: 24.14.1
  npm: 11.11.0
  pnpm: 10.32.1
npmPackages:
  @mastra/core: 1.25.0
  @mastra/pg: 1.9.1
  zod: 4.3.6
```

- Storage: Postgres (via `@mastra/pg`)
- LLM provider: N/A — the repro uses a pure in-memory echo step, no LLM.

intojhanurag:
## Description

`dataset.startExperiment({ targetType: 'workflow', targetId })` hung forever because `resolveTarget` is an `async` function and a `Workflow` is thenable (it exposes a `.then(step)` builder for chaining). When the async resolver returned the workflow, JavaScript's promise machinery invoked `.then(resolve, reject)` on it to unwrap — but the builder `.then` never calls those callbacks, so the outer promise never settled.

Regression introduced in #15317, which made `resolveTarget` async to support agent version pinning.

- Wrap the resolved target in a plain `{ target }` object so the Promise machinery cannot treat it as a thenable.
- Update the single call site in `runExperiment` to destructure the wrapper.
- Replace the misleading inline NOTE with a docstring that explains *why* the wrapper exists.
- Add a regression test that builds a real `createWorkflow().then(step).commit()` instance (the prior test used a plain mock without `.then`, which is why CI didn't catch the hang).

## Related Issue(s)

Fixes #15453

## Type of Change

- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [ ] Test update

## Checklist

## ELI5
Running an experiment that targeted a workflow could hang forever because workflow objects expose a .then builder that confused Promise resolution. The fix wraps the workflow in a plain object ({ target }) so JavaScript won't try to "unwrap" it, preventing the hang.

---

## Summary of Changes

This PR fixes a hang in dataset.startExperiment() when the targetType is "workflow". The hang occurred because resolveTarget returned a Workflow instance directly; as an async return value, Promises treated that instance as a thenable and called its .then builder, which never invoked the provided resolve/reject callbacks—leaving the outer promise pending.

### Root Cause
An async resolveTarget returned a Workflow instance (which implements a .then builder). Promise resolution attempted to treat the Workflow as a thenable, invoking its builder and stalling, so workflow.createRun was never called and the experiment never settled.

### Solution
- resolveTarget now returns a plain wrapper object { target } (or null) instead of the Target instance directly, preventing it from being recognized as a thenable.
- runExperiment updated to use the wrapper and destructure const { target } = resolved before calling executeTarget.
- Replaced a misleading inline NOTE with a docstring explaining why the wrapper is required.
- Added a regression test that constructs a real workflow via createWorkflow().then(step).commit() (previous test used a mock without .then and missed the hang).

### Files Changed
1. .changeset/seven-books-peel.md — Changeset documenting the patch and referencing Fixes #15453.
2. packages/core/src/datasets/experiment/index.ts — resolveTarget now returns `{ target } | null`; runExperiment updated to destructure and use the wrapped target.
3. packages/core/src/datasets/experiment/__tests__/runExperiment.test.ts — Added regression test that builds a real committed workflow (with .then) and verifies runExperiment completes and produces expected outputs.
4. packages/core/src/server/server.test-d.ts — minor import ordering change (no behavioral impact).

### Behavior Impact
- Before: startExperiment({ targetType: 'workflow', targetId }) could hang indefinitely; workflow.createRun was never invoked; itemTimeout didn't fire.
- After: workflow-targeted experiments complete normally; runs start as expected; itemTimeout and failure propagation work correctly.

### Testing
- Added regression test asserting experiment completes (status: completed), with expected succeeded/failed counts and outputs. The test uses a real workflow instance (created via createWorkflow().then(...).commit()) and includes a 10s timeout.

Type: Bug fix (non-breaking)  
Related: Fixes #15453

### DIFF
--- a/.changeset/seven-books-peel.md
+++ b/.changeset/seven-books-peel.md
@@ -2,4 +2,4 @@
 '@mastra/core': patch
 ---
 
-Fixed `dataset.startExperiment({ targetType: 'workflow', targetId })` hanging forever. Internally, the target resolver returned a workflow from an async function, but workflows expose a `.then(step)` builder method that makes them look like a thenable to JavaScript. The promise machinery then tried to unwrap the workflow and never settled. The resolver now returns a non-thenable wrapper so workflow experiments complete normally, honour `itemTimeout`, and surface failures. Fixes #15453.
+Fixed `dataset.startExperiment` hanging forever when `targetType` is `'workflow'`. Workflow experiments now complete normally, honour `itemTimeout`, and surface failures. Fixes #15453.
--- a/.changeset/seven-books-peel.md
+++ b/.changeset/seven-books-peel.md
@@ -1,0 +1,5 @@
+---
+'@mastra/core': patch
+---
+
+Fixed `dataset.startExperiment({ targetType: 'workflow', targetId })` hanging forever. Internally, the target resolver returned a workflow from an async function, but workflows expose a `.then(step)` builder method that makes them look like a thenable to JavaScript. The promise machinery then tried to unwrap the workflow and never settled. The resolver now returns a non-thenable wrapper so workflow experiments complete normally, honour `itemTimeout`, and surface failures. Fixes #15453.
--- a/packages/core/src/datasets/experiment/__tests__/runExperiment.test.ts
+++ b/packages/core/src/datasets/experiment/__tests__/runExperiment.test.ts
@@ -1,4 +1,5 @@
 import { describe, it, expect, beforeEach, vi } from 'vitest';
+import { z } from 'zod';
 import type { MastraScorer } from '../../../evals/base';
 import type { Mastra } from '../../../mastra';
 import { RequestContext } from '../../../request-context';
@@ -6,6 +7,7 @@ import type { MastraCompositeStore, StorageDomains } from '../../../storage/base
 import { DatasetsInMemory } from '../../../storage/domains/datasets/inmemory';
 import { ExperimentsInMemory } from '../../../storage/domains/experiments/inmemory';
 import { InMemoryDB } from '../../../storage/domains/inmemory-db';
+import { createStep, createWorkflow } from '../../../workflows';
 import { runExperiment } from '../index';
 
 // Mock agent that returns predictable output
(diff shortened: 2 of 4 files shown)


intojhanurag:
One file is unrelated , due to linting issue

DanielSLew (approved):

PR #15571, opened 2026-04-21
fix(core): preserve tool execution errors through history reload

### LINKED ISSUE #15569: Tool execution errors stored as successful results — lost on history reload
## Bug Description

## Root Cause

```js
const chunk = {
  // ...
};

  type: "tool-invocation",
  toolInvocation: {
    args: toolCall.args,
  },
});
```

```js
if (inv.state === "result") {
} else {
}
```

## Reproduction

## Expected Behavior

## Suggested Fix

```js
  type: "tool-invocation",
  toolInvocation: {
    state: "output-error",
    args: toolCall.args,
  },
});
```

```js
if (inv.state === "result") {
  // existing success path
  parts.push({
    type: `tool-${inv.toolName}`,
    toolCallId: inv.toolCallId,
    input: inv.args,
    state: "output-error",
  });
} else {
}
```

## Versions

- `@mastra/core@1.25.0`
- `@mastra/ai-sdk@1.4.0`
- `ai@5.1.3`

kagura-agent:
## Summary

Fixes #15569 — Tool execution errors were stored as successful results and lost on history reload.

## Problem

When a tool's `execute()` throws an error:
- ✅ During live streaming: correctly shows `state: 'output-error'` with `errorText`
- ❌ **Bug**: When loaded from history via `toAISdkV5Messages()`, appears as successful (`state: 'output-available'` with error string as `output`)

## Root Cause

The `AIV5Adapter` had three gaps in error state handling:

1. **`toUIMessage()`**: Only checked `state === 'result'`, missing `'output-error'` — fell through to the else branch which treated it as a call/streaming state
2. **`fromModelMessage()`**: `updateMatchingCallInvocationResult()` always set `state: 'result'`, ignoring the `isError` flag on `ToolResultPart`
3. **`fromUIMessage()`**: Similar to #1, didn't preserve `output-error` state when converting UI messages back to DB format

## Fix

- **`toUIMessage()`**: Added `output-error` branch in both toolInvocations and parts handling, emitting `errorText` instead of `output`
- **`fromModelMessage()`**: Check `isError` on tool-result parts; when true, set `state: 'output-error'` with `errorText` instead of `state: 'result'`
- **`fromUIMessage()`**: Preserve `output-error` state in both conversion paths (toolInvocations array and V2 parts)

All changes follow the existing V6 adapter pattern (lines 442-448, 579-585).

## Tests

Added `AIV5Adapter-tool-error.test.ts` covering:
- `toUIMessage` converts `output-error` state correctly from both toolInvocations and parts
- `fromModelMessage` handles `isError=true` on tool-result parts
- Round-trip preservation: error stored → loaded from history → still shows as error
- Edge cases: empty errorText, missing isError field

## Changeset

Included — `@mastra/core` patch.

## ELI5

When tools fail during execution, the errors now stay labeled as errors instead of getting mixed up with successful results. This way, when you reload the conversation history, the errors show up correctly as errors rather than appearing like the tool worked but returned an error message.

## Overview

This PR fixes issue #15569 by ensuring tool execution errors are properly preserved through history reloads. The problem occurred when tool invocations that threw errors during execution were stored with an `output-error` state during live streaming, but appeared as successful results when messages were later loaded from history. The root cause was in the `AIV5Adapter` message conversion logic, which didn't properly handle the `output-error` state.

### Changeset
- Added a patch-level changeset for `@mastra/core` documenting the tool error state preservation fix.

### AIV5Adapter Enhancements (`AIV5Adapter.ts`)
The adapter now correctly handles `output-error` state across all conversion paths:

- **`toUIMessage()`**: Added handling for `output-error` state when converting tool invocations from the database. Tool invocations with `state: 'output-error'` now produce UI parts with `state: 'output-error'` and include `errorText` (defaulting to empty string if not present).

- **`fromModelMessage()`**: Updated to check the `isError` flag on `ToolResultPart`. When `isError === true`, it sets the matching tool invocation to `state: 'output-error'` with `errorText` derived from the output, instead of always setting `state: 'result'`.

- **`fromUIMessage()`**: Extended to preserve `output-error` state when converting UI parts back to database format, creating `MastraDB` tool invocation entries with the appropriate error state and `errorText`.

### Test Coverage (`AIV5Adapter-tool-error.test.ts`)
A comprehensive test suite validates:

- **State conversion**: `toUIMessage()` correctly converts DB tool invocations with `output-error` state to UI parts with `errorText`.
- **Error flag handling**: `fromModelMessage()` properly respects the `isError` flag and sets the appropriate state.
- **Round-trip preservation**: Error state and `errorText` survive a complete `fromModelMessage` → `toUIMessage` conversion cycle without being misinterpreted as successful output.
- **Edge cases**: Proper handling of missing or empty `errorText` values, and behavior when the `isError` flag is absent.

## Result

Tool execution errors are now consistently preserved through save/load cycles. Live streaming errors and history-loaded errors now have parity, ensuring users see the correct error state whether a conversation is being streamed in real-time or loaded from history.

### DIFF
--- a/.changeset/fix-tool-error-state-history.md
+++ b/.changeset/fix-tool-error-state-history.md
@@ -1,0 +1,7 @@
+---
+'@mastra/core': patch
+---
+
+Fix tool execution errors being stored as successful results on history reload
+
+When a tool's `execute()` threw an error, the error state was correctly shown during live streaming but was lost when loading from history. The `AIV5Adapter` now correctly handles `output-error` state in `toUIMessage`, `fromUIMessage`, and `fromModelMessage` conversions, preserving `errorText` through save/load cycles.
--- a/packages/core/src/agent/message-list/adapters/AIV5Adapter-tool-error.test.ts
+++ b/packages/core/src/agent/message-list/adapters/AIV5Adapter-tool-error.test.ts
@@ -1,0 +1,267 @@
+import { describe, it, expect } from 'vitest';
+
+import type { AIV5Type } from '../types';
+import type { MastraDBMessage } from '../state/types';
+import { AIV5Adapter } from './AIV5Adapter';
+
+describe('AIV5Adapter - tool error handling', () => {
+  describe('toUIMessage - handling output-error state', () => {
+    it('should convert output-error state from toolInvocations array', () => {
+      const dbMsg: MastraDBMessage = {
+        id: 'test-msg',
+        role: 'assistant',
+        createdAt: new Date('2024-01-01T00:00:00Z'),
+        content: {
+          format: 2,
+          toolInvocations: [
+            {
+              toolCallId: 'call-123',
+              toolName: 'testTool',
+              args: { input: 'test' },
+              state: 'output-error',
+              errorText: 'Tool execution failed',
+            },
+          ],
+        },
+      };
+
+      const uiMsg = AIV5Adapter.toUIMessage(dbMsg);
+
+      expect(uiMsg.parts).toHaveLength(1);
+      const part = uiMsg.parts[0] as AIV5Type.ToolUIPart;
+      expect(part.type).toBe('tool-testTool');
+      expect(part.toolCallId).toBe('call-123');
+      expect(part.state).toBe('output-error');
+      expect(part.errorText).toBe('Tool execution failed');
+      expect(part).not.toHaveProperty('output');
+    });
+
+    it('should convert output-error state from parts array', () => {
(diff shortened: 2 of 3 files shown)


kagura-agent:
Note: I wasn't able to run the test suite locally (monorepo dependencies OOM during `pnpm install` on my machine). The changes follow the exact same pattern as the V6 adapter's `output-error` handling. CI results will confirm correctness — happy to address any failures.

PR #15575, opened 2026-04-21
fix(memory): surrogate-safe string truncation prevents Anthropic JSON parse errors

### LINKED ISSUE #15573: Observational memory truncation splits UTF-16 surrogate pairs → Anthropic rejects as invalid JSON
This issue was created from Discord post 1496041127222972476:

[Open in Discord](https://discord.com/channels/1309558646228779139/1496041127222972476)

Hitting a reproducible bug: observational memory + Anthropic provider → intermittent 400s:

```
AI_APICallError: The request body is not valid JSON: no low surrogate in string: line 1 column 20240 (char 20239)
  at AnthropicMessagesLanguageModel.doStream (@ai-sdk/anthropic/src/anthropic-messages-language-model.ts:1270)
  at AISDKV6LanguageModel.doStream (@mastra/core/src/llm/model/aisdk/v6/model.ts:80)
```

Error is from Anthropic's server-side JSON parser — fires when the body contains an unpaired UTF-16 high surrogate.

**Root cause (strong suspicion)**
The observations system prompt embeds past threads and truncates with `... [truncated N characters]`. That looks like `str.slice(0, N)`, which cuts by UTF-16 code unit. Emoji outside the BMP (🔥 U+1F525, 🎭, 😄) are surrogate pairs — if the cut lands between the high and low surrogate, the body contains a lone high surrogate. `JSON.stringify` emits it; Anthropic rejects it. The error offset consistently lands inside `<other-conversation>` blocks containing emoji.

**Repro**
1. Agent with Anthropic + working memory + observational memory (semantic recall across threads).
2. Seed a past thread with emoji (org named `🔥🔥🔥`, or 🎭/😄 in replies) long enough to hit the truncation threshold.
3. New thread, send "hello" → intermittent `no low surrogate` 400. Clearing memory fixes it until emoji content is re-ingested.

Other providers likely silently replace lone surrogates so the bug is masked — Anthropic validates strictly.

**Fix**
Surrogate-safe truncation:

```ts
function safeTruncate(s: string, max: number) {
  if (s.length <= max) return s;
  let end = max;
  const c = s.charCodeAt(end - 1);
  if (c >= 0xd800 && c <= 0xdbff) end -= 1;
  return s.slice(0, end);
}
```

Ideally truncate on code-point / token boundaries.

**Env**
`@mastra/core@1.25.0`, `@ai-sdk/anthropic@3.0.69`, `@ai-sdk/provider-utils@4.0.23`, Node 20, macOS, Anthropic `claude-opus-4-6` streaming + extended thinking.

---

Screenshot 1

kagura-agent:
## Summary

Observational memory truncation uses `str.slice(0, n)` which operates on UTF-16 code units. When the cut lands between a surrogate pair (emoji outside the BMP like 🔥🎭😄), the resulting string contains an unpaired high surrogate. Anthropic's JSON parser strictly validates UTF-16 and rejects the body with `no low surrogate in string`.

## Fix

Added a `surrogateSafeSlice()` helper that checks whether the last code unit before the cut is a high surrogate (U+D800–U+DBFF) and backs off by one position when it is. Applied to all three truncation sites in the memory package:

- `truncateStringByTokens()` in `tool-result-helpers.ts` — binary-search token truncation
- `maybeTruncate()` in `observer-agent.ts` — character-limit truncation
- `sanitizeObservationLines()` in `observer-agent.ts` — per-line length guard

## Tests

Added 4 test cases covering:
- Truncation landing mid-emoji in mixed text
- Strings consisting entirely of surrogate-pair emoji
- Passthrough when text fits within budget
- Per-line sanitization with emoji at the boundary

All tests pass locally (`vitest run`).

Closes #15573

## ELI5

This PR fixes a bug where cutting text at certain positions could accidentally split emoji characters in half, creating broken data that Anthropic's parser rejects. The fix checks if we're about to cut in the middle of an emoji and moves the cut point back one position to keep them whole.

## Overview

This PR addresses issue #15573 by implementing surrogate-safe string truncation across the memory package. The issue occurred because UTF-16 encodes characters outside the Basic Multilingual Plane (like many emoji) using surrogate pairs—two code units that must stay together. When truncation using `slice(0, n)` cut at the wrong position, it could leave an unpaired high surrogate, which Anthropic's strict JSON parser rejects with a "no low surrogate in string" error.

### Core Fix: Surrogate-Safe Truncation

A new helper function `surrogateSafeSlice(str, end)` was added to both `tool-result-helpers.ts` and `observer-agent.ts`. This function detects when a truncation boundary falls immediately after a high surrogate code unit (U+D800–U+DBFF) and automatically backs off by one position to preserve the surrogate pair.

### Applied Locations

- **`truncateStringByTokens()` in `tool-result-helpers.ts`**: Binary-search token truncation now uses `surrogateSafeSlice` instead of the unsafe `slice()` operation.
- **`maybeTruncate()` in `observer-agent.ts`**: Character-limit truncation updated to use `surrogateSafeSlice` with corrected remaining character count calculation.
- **`sanitizeObservationLines()` in `observer-agent.ts`**: Per-line length guard now truncates overlong lines using `surrogateSafeSlice` before appending the ellipsis marker.

### Tests

A new test file `surrogate-safe-truncation.test.ts` was added with four test cases:
- Truncation within mixed text containing emoji at the boundary
- Truncation of strings composed entirely of surrogate-pair emoji
- Passthrough behavior when text fits within the token budget
- Per-line sanitization with emoji at length boundaries

All tests validate that truncated output contains no unpaired surrogates and can be safely JSON-stringified.

### Changeset

A changeset entry (`.changeset/surrogate-safe-truncation.md`) was added marking a patch release for `@mastra/memory` with the bugfix description.

### DIFF
--- a/.changeset/surrogate-safe-truncation.md
+++ b/.changeset/surrogate-safe-truncation.md
@@ -1,0 +1,15 @@
+---
+'@mastra/memory': patch
+---
+
+fix(memory): surrogate-safe string truncation prevents Anthropic JSON parse errors
+
+Observational memory truncation used `str.slice(0, n)` which operates on UTF-16 code units. When the cut lands between a surrogate pair (emoji outside the BMP like 🔥🎭😄), the resulting string contains an unpaired high surrogate. Anthropic's JSON parser strictly validates UTF-16 and rejects the body with "no low surrogate in string".
+
+Fixed by checking whether the last code unit before the cut is a high surrogate (U+D800–U+DBFF) and backing off by one position when it is. Applied to all three truncation sites in the memory package:
+
+- `truncateStringByTokens()` (tool-result-helpers.ts)
+- `maybeTruncate()` (observer-agent.ts)
+- `sanitizeObservationLines()` (observer-agent.ts)
+
+Fixes #15573
--- a/packages/memory/src/processors/observational-memory/__tests__/surrogate-safe-truncation.test.ts
+++ b/packages/memory/src/processors/observational-memory/__tests__/surrogate-safe-truncation.test.ts
@@ -1,0 +1,81 @@
+import { describe, it, expect } from 'vitest';
+import { truncateStringByTokens } from '../tool-result-helpers';
+import { sanitizeObservationLines } from '../observer-agent';
+
+/**
+ * UTF-16 surrogate pair safety tests.
+ *
+ * Emoji outside the BMP (e.g. 🔥 U+1F525) are encoded as two UTF-16 code
+ * units (a surrogate pair).  Naive `str.slice(0, n)` can split a pair,
+ * leaving an unpaired high surrogate that Anthropic's JSON parser rejects
+ * with "no low surrogate in string".
+ *
+ * @see https://github.com/mastra-ai/mastra/issues/15573
+ */
+
+// 🔥 = U+1F525, encoded as \uD83D\uDD25 in UTF-16 (2 code units, .length === 2)
+const FIRE = '🔥';
+
+describe('truncateStringByTokens – surrogate safety', () => {
+  it('should not produce unpaired surrogates when truncation lands mid-emoji', () => {
+    // Build a string where truncation is likely to land on the emoji
+    const text = 'a'.repeat(50) + FIRE + 'b'.repeat(50);
+    // Force truncation by using a very small token budget
+    const result = truncateStringByTokens(text, 10);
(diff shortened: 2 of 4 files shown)


roaminro:
closing in favor of https://github.com/mastra-ai/mastra/pull/15634

PR #15583, opened 2026-04-21
fix(core): include partial text in onAbort when stream is aborted mid-generation

### LINKED ISSUE #15535: bug: onAbort callback receives empty steps and no partial text when stream is aborted mid-generation
## Bug Description

When a user cancels an in-progress LLM stream, the `onAbort` callback receives `{ steps: [] }` with no accumulated text, even if the model had already generated a partial response before cancellation.

## Steps to Reproduce

```ts
const result = await agent.stream(messages, {
  abortSignal: controller.signal,
  onAbort: (event) => {
    console.log(event.text)  // undefined — partial text is lost
    console.log(event.steps) // [] — always empty
  },
})

// abort mid-stream
controller.abort()
```

## Expected Behavior

`onAbort` should receive the text that was generated before the abort signal fired:

```ts
onAbort: ({ steps, text }) => {
  console.log(text) // "Hello, I can help you with..." (partial response)
}
```

## Actual Behavior

`text` is `undefined` and `steps` is always `[]`, regardless of how much text was generated before cancellation.

## Root Cause

In `llm-execution-step.ts`, both abort code paths construct the abort event without reading `runState.state.textDeltas`, so accumulated text is never included in the event payload.

## Impact

Applications that want to save or display partial responses when users cancel generation cannot do so — the generated text is silently discarded.

TheIsrael1:
Supersedes #15521. Fixes #15535.

When a stream is aborted mid-generation, `onAbort` was called with an empty `text` field even if the model had already streamed some tokens. This fix accumulates text deltas into `runState` as they arrive, so `onAbort` receives whatever partial text was generated before cancellation.

**Before**
```ts
agent.stream('...', {
  abortSignal: controller.signal,
  onAbort({ text }) {
    console.log(text); // always ""
  },
});
```

**After**
```ts
agent.stream('...', {
  abortSignal: controller.signal,
  onAbort({ text }) {
    console.log(text); // "Hello wor" — whatever was streamed before abort
  },
});
```

Also tightens the `onAbort` event type from `any` to `{ steps: any[]; text?: string }`.

## ELI5

If the AI is cancelled while it's answering, you no longer lose what it already said — the system now returns the partial text it had produced before the cancel.

## Changes Overview

Fixes a streaming cancellation bug where `onAbort` was called with empty `text` even if tokens had already streamed. The change accumulates text deltas during streaming and passes the joined partial text to `onAbort` in both abort paths.

## Key Technical Changes

- packages/core/src/loop/workflows/agentic-execution/llm-execution-step.ts
  - Consume buffered chunks on abort and append `text-delta` payloads into `runState.state.textDeltas`.
  - Pass `text: runState.state.textDeltas?.join('') ?? ''` to `options.onAbort` in both the abort catch and the post-stream abort check.

- packages/core/src/loop/types.ts
  - Tighten LoopConfig: `onAbort?: (event: { steps: any[]; text?: string }) => Promise<void> | void`.

- packages/core/src/stream/base/input.ts
  - When `MastraModelInput.initialize()` catches an abort error, enqueue an `'error'` chunk with abort context (instead of generic error handling) so the loop can process abort-related chunks.

- .changeset
  - Add changeset documenting that `onAbort` now includes partial response text when streaming is cancelled.

## Tests

- packages/core/src/loop/workflows/agentic-execution/llm-execution-step.test.ts
  - Added tests ensuring `onAbort` is invoked once and receives accumulated `text` when abort occurs after text-deltas, and empty `text` when abort happens before any text-delta.
- Updated snapshots and test utilities to expect non-empty `text` in abort-related payloads where applicable.

## Observable Behavior Change

Callers of `onAbort` will now receive any partial text produced before a cancellation via `event.text`, enabling saving or displaying partial responses that were previously lost.


abhiaiyer91:

We're currently testing out our agentic factory, and it implemented a fix for the linked issue (#15535), which has since been closed. So I'm closing this PR out as no longer needed.

PR #15593, opened 2026-04-21
fix(cli): deploy to existing Observe project instead of always creating new

DanielSLew:
Closes [PLTFRM-866](https://linear.app/kepler-crm/issue/PLTFRM-866).

## Summary

Both `mastra studio deploy` and `mastra server deploy` now let users pick an existing Observe project instead of always creating a new one.

- `mastra studio deploy` previously always proposed creating a brand new project named after the local `package.json`.
- `mastra server deploy` looked up projects but silently matched by name/slug — which could target the wrong project on collision.

Both commands now, in order:

1. Honor `MASTRA_PROJECT_ID`, `--project <id-or-slug>`, and `.mastra-project.json` same-org auto-match (unchanged).
2. Fall back to a `@clack/prompts` selector listing every project in the org. Initial selection defaults to any project whose name/slug matches the local `package.json`.
3. `-y/--yes` mode auto-selects only when there is exactly one name/slug match; otherwise fails with a clear instruction to pass `--project <id-or-slug>`.

The "create new project" flow now runs only when the user picks "Create new project" in the selector (or when the org has zero projects and `-y` is passed with a valid name).

Headless deploys using `MASTRA_API_TOKEN` still require `MASTRA_ORG_ID + MASTRA_PROJECT_ID` (unchanged).

## Test plan

- \`pnpm --filter ./packages/cli typecheck\` — clean
- \`pnpm test:cli\` — 287/287 including the new selector / \`--project\` / auto-accept tests on both commands
- Manual: deploy an unconfigured project into an existing Observe project with and without \`--project\` and \`-y\`

### DIFF
--- a/.changeset/pltfrm-866-deploy-to-existing-project.md
+++ b/.changeset/pltfrm-866-deploy-to-existing-project.md
@@ -1,0 +1,14 @@
+---
+'mastra': patch
+---
+
+`mastra studio deploy` and `mastra server deploy` now prompt you to pick from existing projects in the selected organization instead of silently matching by `package.json` name or always creating a new project.
+
+**What changed**
+
+- When existing projects are found, both commands show an interactive selector listing them (plus a "Create new project" option).
+- `--project <id-or-slug>` still bypasses the selector for non-interactive use.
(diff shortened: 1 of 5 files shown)


DanielSLew:
Superseded by consolidated #15728 (feat/mastra-observe).

PR #15594, opened 2026-04-21
feat(create-mastra): ask about enabling Mastra Observe during project bootstrap

DanielSLew:
## Summary

Both `mastra init` and `create-mastra` now ask "Do you want to enable Mastra Observe?" during interactive onboarding. If the user answers yes, the generated \`.env\` gets a \`MASTRA_CLOUD_ACCESS_TOKEN=\` entry and the CLI prints a pointer to https://cloud.mastra.ai so the user can mint a token.

The existing generated \`src/mastra/index.ts\` already registers a \`CloudExporter\` that no-ops when the env var is empty, so enabling Observe is a pure env-var concern for now.

Also adds \`--observe\` / \`--no-observe\` flags to both commands so the prompt can be suppressed in non-interactive flows.

## Test plan

- \`pnpm --filter ./packages/cli typecheck\` — clean
- \`pnpm --filter ./packages/cli test -- src/commands/init/\` — 285/285, including the new \`writeObserveEnv\` tests (appends to existing \`.env\`, writes real token, creates \`.env\` when missing)
- \`pnpm --filter ./packages/cli test -- src/commands/create/\` — 285/285, no regressions
- \`pnpm build:cli\` — succeeds
- \`pnpm --filter ./packages/create-mastra build\` — succeeds
- Manual: run \`create-mastra\` interactively, choose yes, confirm \`.env\` contains the entry and the prompt appears

### DIFF
--- a/.changeset/observe-prompt-create-mastra.md
+++ b/.changeset/observe-prompt-create-mastra.md
@@ -1,0 +1,10 @@
+---
+'mastra': minor
+'create-mastra': minor
+---
+
+Add "Enable Mastra Observe?" prompt to `create-mastra` and `mastra init`.
+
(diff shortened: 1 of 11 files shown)


DanielSLew:
Superseded by consolidated #15728 (feat/mastra-observe).

PR #15603, opened 2026-04-22
fix(core): pin @vercel/oidc to 3.1.0 as direct dependency

abhiaiyer91:
## Summary

`@vercel/oidc` 3.2.0 ships a build that esbuild cannot safely re-bundle as ESM — it emits `__require("path")` stubs that throw `Dynamic require of "path" is not supported` at runtime. This breaks E2E tests and any consumer project that installs `@mastra/core`.

The previous override-based fix (#15601) only pinned `@vercel/oidc` within this monorepo's own lockfile. Consumer installs (and our E2E temp projects) resolve `@vercel/oidc` independently via `@ai-sdk/gateway`'s transitive dep, so they still picked up 3.2.0.

## Fix

- Add `@vercel/oidc` as a direct, exact-version (`3.1.0`) dependency of `@mastra/core` so downstream installs resolve 3.1.0 alongside core.
- Remove the monorepo-only `pnpm.overrides` entry that didn't affect consumers.

## Test plan

- Local `pnpm install` + `pnpm build:lib` on `@mastra/core` ✅
- E2E `no-bundling` (which was failing with the override) should now resolve 3.1.0 in the generated project's install and start cleanly.

## ELI5 (Explain Like I'm 5)

A newer version of a library called `@vercel/oidc` (version 3.2.0) breaks when bundled with the Mastra project, so this PR locks the @mastra/core package to use an older working version (3.1.0) instead. By making this version requirement explicit in @mastra/core's own dependencies, anyone who installs Mastra will automatically get the working version, just like specifying which toy goes with which toy set.

## Changes

- **`.changeset/pin-vercel-oidc-core.md`** (new file): Added a changeset entry documenting this patch-level fix for @mastra/core
- **`package.json`**: Removed the monorepo-only `pnpm.overrides` entry for `@vercel/oidc`, which was ineffective for downstream consumers
- **`packages/core/package.json`**: Added `@vercel/oidc@3.1.0` as a direct runtime dependency, ensuring all downstream installations resolve the working version

## Context

The root cause: @vercel/oidc 3.2.0 contains build artifacts that esbuild cannot safely re-bundle, emitting dynamic require stubs (`__require("path")`) that fail at runtime. While a prior monorepo-only override partially addressed the issue, it did not affect consumers installing @mastra/core as a dependency, since transitive resolution could still pull in the broken 3.2.0 through @ai-sdk/gateway. Making @vercel/oidc a direct dependency ensures all downstream projects—E2E test scenarios and consumer applications alike—receive the compatible 3.1.0 version.

### DIFF
--- a/.changeset/pin-vercel-oidc-core.md
+++ b/.changeset/pin-vercel-oidc-core.md
@@ -1,0 +1,11 @@
+---
+'@mastra/core': patch
+---
+
+fix(deps): pin @vercel/oidc to 3.1.0 in @mastra/core
+
+@vercel/oidc 3.2.0 ships a build that breaks when bundled as ESM from CJS
+(emits `__require("path")` stubs that throw "Dynamic require of path is not
+supported" at runtime). Declaring @vercel/oidc as a direct, exact-version
+dependency of @mastra/core forces downstream installs to resolve 3.1.0,
+unblocking E2E and consumer projects.
--- a/package.json
+++ b/package.json
@@ -119,8 +119,7 @@
       "lodash@<=4.17.23": "^4.18.0",
       "prismjs@<1.30.0": "^1.30.0",
       "serialize-javascript@<7.0.5": "^7.0.5",
-      "@tootallnate/once@<3.0.1": "^3.0.1",
-      "@vercel/oidc": "3.1.0"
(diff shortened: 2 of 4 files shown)


abhiaiyer91:
Superseded by the revert PR — rolling back the lockfile maintenance instead of patching around it.

PR #15611, opened 2026-04-22
fix(core): stop forcing temperature: 0 into agent modelSettings

### LINKED ISSUE #15240: Remove hardcoded temperature=0 default in prepareCallSettings — breaks models like Kimi K2.5

truffle-dev:
Fixes #15240.

The agent's stream workflow seeds `modelSettings` with `temperature: 0` before spreading the caller's values on top. Callers that pass `temperature` are unaffected, but callers that omit it silently get `0` in the outgoing request. That breaks providers that restrict the accepted temperature values. Moonshot Kimi K2.5 is the reported case. It returns `400 Bad Request` for any value other than `1`, so agents pointed at that model can't stream at all.

Re @TylerBarnes's v4/v5/v6 question: this reproduces on all three. The injection happens in `packages/core/src/agent/workflows/prepare-stream/map-results-step.ts` before the request is routed to `AISDKV4LegacyLanguageModel` / `AISDKV5LanguageModel` / `AISDKV6LanguageModel`. By the time the vendored SDK is called, the default is already in `modelSettings`.

@hikariming's v4 finding is also real. `ai@4.3.19`'s `prepareCallSettings` has `temperature != null ? temperature : 0`, while v5 and v6 have dropped that default upstream. But that SDK-level fallback only runs after Mastra's own seed, so fixing it in the vendored v4 wouldn't help v5/v6 callers. Removing the Mastra-side seed fixes all three paths at once.

The fix drops the seed. When the caller passes `temperature` it's forwarded unchanged (including `temperature: 0` for deterministic output, which no longer relies on the default). When the caller omits it, the provider's own default is used.

Regression tests in `packages/core/src/agent/__tests__/agent-model-settings-defaults.test.ts`, patterned on the existing `MockLanguageModelV2` + `doStreamCalls` setup used by `per-model-fallback-settings.test.ts`:

- No `temperature` set by the caller: `doStreamCalls[0].temperature` is `undefined` (was `0`).
- `temperature: 0` set by the caller: `doStreamCalls[0].temperature` is `0`.

Changeset bumps `@mastra/core` as a patch.

## ELI5

Mastra stopped secretly setting temperature to 0 for agent calls when users didn't pick one. Now if you don't provide a temperature the provider's own default is used; if you do provide a temperature (including 0), it is forwarded unchanged.

### Core Fix
- Removed Mastra-side seeding of `temperature: 0` into `modelSettings` for agent workflows. The change is in packages/core/src/agent/workflows/prepare-stream/map-results-step.ts so omitted `temperature` values are forwarded as `undefined` and provider defaults apply; explicit `temperature` values (including `0`) are preserved.

### Tests
- Added regression tests in packages/core/src/agent/__tests__/agent-model-settings-defaults.test.ts covering both `agent.stream()` and `agent.generate()`:
  - Omitting `temperature` yields no `temperature` field in outgoing provider calls.
  - Explicit `temperature: 0` is preserved and forwarded.
  - Tests ensure the provider call completed before validating recorded call data.

### Observability / Snapshots
- Updated many observability snapshots (observability/mastra/src/__snapshots__/*) to reflect that span `attributes.parameters` no longer contains `{ "temperature": 0 }` and now serializes as `{}` when temperature is omitted.

### Recordings & Integration Tests
- Several recorded fixtures and E2E recordings were refreshed to match new request hashes caused by removing the hardcoded temperature.
- Memory integration tests were adapted: a local test wrapper injects `modelSettings.temperature = 0` for V5+ models during test runs to preserve prior recording expectations where necessary.
- The author noted additional recording re-generation may be required and requested access/permission to refresh scoped recordings.

### Release Notes
- Added changeset .changeset/fix-remove-hardcoded-temperature-default.md to publish a patch release for @mastra/core; references issue #15240.

## Impact
- Fixes 400 errors from providers that reject `temperature=0` (example: Moonshot Kimi K2.5) by allowing provider defaults when callers omit temperature.
- Preserves behavior for callers that explicitly set `temperature` (including `0`).

---

<a href="https://app.devin.ai/review/mastra-ai/mastra/pull/15611" target="_blank">
  <picture>
    <source media="(prefers-color-scheme: dark)" srcset="https://static.devin.ai/assets/gh-open-in-devin-review-dark.svg?v=1">
    Open in Devin Review
  </picture>
</a>


truffle-dev:
1. Parameterized the test across both `agent.stream()` and `agent.generate()` so the regression is pinned at both entry points named in the changeset. 790ded3
2. Added `expect(model.doStream/doGenerateCalls).toHaveLength(1)` before each index-0 assertion so a future regression that skips the provider call can't pass the temperature check. 790ded3

wardpeet (approved):
Thank you!

TylerBarnes:
@truffle-dev looks like some snapshots need to be updated and some other potentially related tests are failing

truffle-dev:
1. Observability tracing snapshots — 13eafbdda10b7c11d62f8d2fe35f3a4ddae9b03c. 29 files captured `"parameters": { "temperature": 0 }` from the old default injection, now emit `"parameters": {}` to match what `agent.generate()` actually passes when callers don't set `modelSettings`.

2. The other failures (agent-memory, message-ordering, streaming-memory, `builder.e2e` openrouter v2, and related) are the same root cause one layer down: `llm-recorder` hashes the request body, so removing the hardcoded `temperature: 0` changes the hash. I reproduced CI's reported `4bcb022ba4cbaf31 → ee7b002b3a5658f2` transition locally against the memory-integration recordings to confirm. Fix requires re-recording, which `working-memory.test.ts` complicates (it sets `modelSettings: { temperature: 0 }` explicitly at `integration-tests/src/shared/working-memory.ts:733,743` for some calls but not others, so the recordings mix entries that should keep the hash and entries that shouldn't). Happy to run `UPDATE_RECORDINGS=true` if there's a way to get access to the keys, or push a scoped recording refresh for just the files where the test file never sets `modelSettings` — whichever fits your workflow better.

TylerBarnes:
@truffle-dev I'll look at re-running them locally, one sec

abhiaiyer91 (approved):

truffle-dev:
CI at 19:24Z on 708bdc1 still fails `src/working-memory.test.ts > V5 > Working Memory Additive Updates (openai/gpt-5.2) > Large Real-World Schema - User Context > should preserve people array when adding work details`. Hash misses cascade into `APICallError: API Key not found` when fuzzy fall-through lands on a recording that doesn't serve the request.

Ten unique unmatched hashes from the failing run:

```
7a1379b7c7f6d0c9  c200e7c32a26098e  625dba3fae48be45
ab8b80baac83064f  b0985dc33f5f9eed  ab28f6758480550b
8d440e2c864a2381  18a5c140f4f93427  3ddc4c3e4fb845b1
e6831c92e93fde74
```

None appear in any of the 13 recording files under `packages/memory/__recordings__/memory-integration-tests-*.json`. The fuzzy fall-through target `fc442148efdf06c7` only exists in `memory-integration-tests-src-agent-memory-v4.json`, so the fallback crosses namespaces and then fails when the served recording doesn't match the actual request.

Points at the V5 Additive-Updates (openai/gpt-5.2) scenarios never having been captured under a namespace the working-memory test file discovers. Ball stays with you on UPDATE_RECORDINGS access.

TylerBarnes:
@truffle-dev yeah it takes a while to regen some of these so I have a bg agent doing it, committing/pushing, waiting for ci, and looping

truffle-dev:
Got it. Standing down while you loop. Happy to re-trace if the unmatched set shifts in a way that needs a second eye.

truffle-dev:
Thanks for landing it. The bg-agent regen loop you described (commit, push, wait for CI, repeat) is a clean shape for any test suite where deterministic LLM-output fixtures get hash-coupled to model behavior. Useful pattern to remember next time I run into a similar regen wall. Glad the temperature ceiling is gone.

PR #15635, opened 2026-04-22
fix(hono): align handlerParams tools with fastify and express adapters

Akash504-ai:
## Description

Expose `tools` in Hono adapter handlerParams to align with Fastify and Express adapters.  Fixes inconsistency where `tools` was missing in Hono.

## Related Issue(s)

N/A

## Type of Change

## Checklist

## ELI5 (Explain Like I'm 5)

Imagine you have a toolbox that stores tools you can use later. The problem was that if someone gave you a request with a field also called "tools" in it, the two would get mixed up and cause problems. The fix is to call the stored tools "registeredTools" instead so there's no confusion.

## Overview

This PR fixes a naming collision in the Fastify adapter where exposing registered tools as `tools` in handler parameters was silently overwriting `tools` fields in request bodies for endpoints like `POST /stored/agents` and `POST /stored/workspaces`.

### Fastify Adapter (`server-adapters/fastify/src/index.ts`)
- **Module augmentation**: Renamed the `FastifyRequest` property from `tools: ToolsInput` to `registeredTools: ToolsInput` to avoid conflicts with request body fields
- **Context middleware**: Updated to populate `request.registeredTools` instead of `request.tools`
- **Handler parameters**: Modified route handler parameter mapping to pass `registeredTools: request.registeredTools` instead of `tools: request.tools`

### Changeset Documentation
- Added `.changeset/fix-fastify-registered-tools.md` documenting the patch-level fix for `@mastra/fastify`
- Documents that registered tools are now exposed via `registeredTools` parameter, aligning with Express and Hono adapter contracts

## Alignment
The change aligns the Fastify adapter's handler contract with the Express and Hono adapters, which already use `registeredTools` to prevent request body field collisions.


abhiaiyer91:
Fastify should actually do `registeredTools`, i will take over this PR

Akash504-ai:
Got it, thanks for taking a look! Happy to adjust if needed.

PR #15660, opened 2026-04-23
fix(arize): emit llm.tools.* from mastra.model_step.input when native…

jd-avinya:
… tool definitions are absent

Mastra model_step spans carry tool definitions inside mastra.model_step.input as a tools[] array, but the OpenInferenceOTLPTraceExporter did not convert these to the OpenInference llm.tools.{i}.tool.json_schema format.

As a result, observability backends that read llm.tools.* (including Arize and Phoenix) showed no tool definitions for Mastra agents even when tools were configured and used.

This change adds a fallback: when llm.tools.0.tool.json_schema is not already present on the span, parse mastra.model_step.input and emit each tool's name, description, and parameters as llm.tools.{i}.tool.json_schema. The type field and other wrapper-specific keys are intentionally excluded to keep the output schema-agnostic.

The guard ensures native llm.tools.* attributes (from Vercel AI SDK or other OpenInference-native sources) are never overwritten.

## Type of Change

- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [ ] Test update

## Checklist

## ELI5 Explanation
When Mastra agents use tools, the information about those tools wasn't being properly sent to observability tools like Arize. This PR adds code that takes the tool information that Mastra has and converts it to the format that Arize and other tools expect, so you can actually see what tools your agent is using.

## Changes

**observability/arize/src/openInferenceOTLPExporter.ts**
- Added fallback logic to extract and emit tool definitions from `mastra.model_step.input` when native OpenInference `llm.tools.*.tool.json_schema` attributes are not already present on the span
- When tools array exists in the input, the exporter now iterates through each tool and sets `llm.tools.{index}.tool.json_schema` with a JSON string containing the tool's `name`, `description`, and `parameters`
- Intentionally omits the `type` field and wrapper-specific keys to keep the schema agnostic
- Includes a guard to prevent overwriting pre-existing native `llm.tools.*` attributes from other sources (e.g., Vercel AI SDK)
- All parsing/processing errors are silently ignored

**observability/arize/src/tracing.test.ts**
- Added test coverage for tool schema emission from `mastra.model_step.input.tools` when no native OpenInference `llm.tools.*` attributes exist, verifying correct `name`, `description`, and `parameters` in the emitted schema and that extraneous wrapper fields like `type` are omitted
- Added test coverage to ensure pre-existing native `llm.tools.*` attributes are preserved unchanged and not overwritten by the fallback logic

### DIFF
--- a/observability/arize/src/openInferenceOTLPExporter.ts
+++ b/observability/arize/src/openInferenceOTLPExporter.ts
@@ -234,6 +234,27 @@ export class OpenInferenceOTLPTraceExporter extends OTLPTraceExporter {
 
         mutableSpan.attributes = { ...processedAttributes, ...mastraOther };
 
+        // Emit tool definitions from mastra.model_step.input as OpenInference llm.tools.*
+        // Only runs when native llm.tools.* attributes are not already present.
+        if (!mutableSpan.attributes['llm.tools.0.tool.json_schema']) {
+          const modelStepRaw = mastraOther[MASTRA_MODEL_STEP_INPUT];
+          if (modelStepRaw) {
+            try {
+              const modelStep = typeof modelStepRaw === 'string' ? JSON.parse(modelStepRaw) : modelStepRaw;
+              if (Array.isArray(modelStep?.tools)) {
+                (modelStep.tools as Record<string, unknown>[]).forEach((tool, i) => {
+                  const { name, description, parameters } = tool;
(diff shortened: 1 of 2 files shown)


abhiaiyer91:
Thanks for this PR! Please reopen it with a valid issue!

PR #15666, opened 2026-04-23
Fix public origin resolution for AWS ALB deployments

### LINKED ISSUE #14886: Datasets not visible in Mastra Studio (only in production env)
This issue was created from Discord post 1488486646944108664:

[Open in Discord](https://discord.com/channels/1309558646228779139/1488486646944108664)

## Environment:
- Storage: PostgreSQL on AWS RDS
- Deployment: Task on ECR
- @mastra packages version
```
@mastra/core": "^1.16.0",
@mastra/evals": "^1.1.2",
@mastra/libsql": "1.7.2",
@mastra/loggers": "^1.0.3",
@mastra/mcp": "^1.3.1",
@mastra/memory": "^1.10.0",
@mastra/observability": "^1.5.1",
@mastra/opensearch": "^1.0.1",
@mastra/pg: "^1.8.3"
```
## Observed behavior:
When running locally, everything works as expected:
- `mastra dev` → Datasets menu visible
- `mastra build --studio` + `node .mastra/output/index.mjs` → Datasets menu visible
- Building a Docker image locally using the same Dockerfile used in production → Datasets menu visible
- After deploying the same Docker image to production (ECS), the Datasets menu does not appear in Mastra Studio
- Observability (traces) works correctly in production, confirming the PostgreSQL storage is reachable and functional

## What we tried:
- Confirmed storage is configured at the Mastra instance level (not agent-level only)
- Added explicit await storage.init() call at startup — no change
- Storage uses PostgresStore with a valid connectionString pointing to RDS
- tried differente @mastra packages versions

## Expected behavior:
The Datasets menu should appear in Studio in production, as it does locally

## Questions for the team:

1.  Are there specific tables for the datasets and experiments domains that init() should create? We could not find them documented.
2. **Is there a runtime condition** (feature flag, version check, etc.) that controls the visibility of the Datasets menu in Studio?
3. Does PostgresStore fully support the datasets domain, or is a MastraCompositeStore with an explicit datasets domain required?

Thanks in advance for the help!
cc: <@1287363396576215153>

cannarocks:
fixes https://github.com/mastra-ai/mastra/issues/14886

This PR improves the logic used to reconstruct the public origin in server-side
handlers by extending `getPublicOrigin()` to correctly handle environments where
`X-Forwarded-Host` is not injected by the proxy/load balancer.

Some infrastructures (e.g. AWS Application Load Balancer with “Preserve Host
Header” enabled) forward the original `Host` header but do not add
`X-Forwarded-Host`. In these cases, the current implementation falls back to
`new URL(request.url).origin`, which results in internal container URLs such as
`http://10.x.x.x:3000` and breaks redirect flows (OAuth, SSO, callback URLs,
etc.).

### What this PR changes
- Uses `X-Forwarded-Host` when present (unchanged behavior)
- Falls back to `Host` + `X-Forwarded-Proto` when `X-Forwarded-Host` is missing
- Falls back to `request.url` only when no headers are available (dev mode)

### Why this is safe
- No API changes
- No breaking behavior
- Existing deployments that rely on `X-Forwarded-Host` continue to work
- Improves compatibility with ALB, Cloudflare, Nginx, and other proxies that
  preserve the `Host` header

This is a patch-level fix that corrects origin reconstruction in real-world
deployments without introducing breaking changes.

## ELI5

The server sometimes used its internal address (like http://10.x.x.x:3000) for redirects when behind certain load balancers. This PR makes the server detect the real public address from proxy headers so redirects (OAuth/SSO) use the correct public origin.

---

### Core Implementation: Public Origin Resolution Logic
- Enhanced getPublicOrigin(request) to build the public origin with a cascading strategy:
  1. Use X-Forwarded-Host (when present) and always treat it as HTTPS (preserves existing behavior and avoids relying on X-Forwarded-Proto).
  2. If X-Forwarded-Host is missing, use the Host header combined with X-Forwarded-Proto (first comma-separated value) when available.
  3. If X-Forwarded-Proto is absent but Host exists, derive the scheme from request.url and combine it with Host.
  4. If no Host header is present, fall back to new URL(request.url).origin (development/direct access fallback).
- Adds JSDoc and comments documenting the trusted-proxy assumption and the resolution priority.

### Tests
- Added get-public-origin.test.ts (Vitest) covering:
  - X-Forwarded-Host parsing (ports preserved, comma-separated lists use first value, whitespace trimmed).
  - X-Forwarded-Host takes precedence and forces HTTPS (ignores X-Forwarded-Proto).
  - Host + X-Forwarded-Proto handling and behavior when X-Forwarded-Proto is absent (scheme comes from request.url).
  - Fallback to request.url.origin when no proxy headers exist.

### Documentation / Changeset
- Added a Changeset (.changeset/warm-bottles-wish.md) describing the patch fix for @mastra/server and the header/url fallback behavior.

---

## Impact & Compatibility
- No API or exported signature changes.
- Non-breaking; existing deployments using X-Forwarded-Host continue to work.
- Fixes incorrect internal-container origins (e.g., http://10.x.x.x:3000) observed with AWS ALB Preserve Host Header and similar proxies, improving OAuth/SSO redirect correctness for ALB, Cloudflare, Nginx, and other proxy setups.

### DIFF
--- a/.changeset/warm-bottles-wish.md
+++ b/.changeset/warm-bottles-wish.md
@@ -1,0 +1,13 @@
+---
+'@mastra/server': patch
+---
+
+Fix: Public origin resolution for AWS ALB deployments
+
+Implement cascading header resolution in getPublicOrigin() to properly handle:
+
+- X-Forwarded-Host (traditional reverse proxies) → always HTTPS
+- Host header (AWS ALB with Preserve Host Header) → respect X-Forwarded-Proto or default HTTPS
+- request.url (local development) → fallback
+
+Fixes OAuth callback URLs being resolved to http:// instead of https:// when deployed behind AWS ALB with Preserve Host Header enabled.
--- a/packages/server/src/server/handlers/auth.ts
+++ b/packages/server/src/server/handlers/auth.ts
@@ -65,17 +65,21 @@ function getAuthProvider(mastra: any): MastraAuthProvider | null {
 
 /**
  * Get the public-facing origin from a request, respecting reverse proxy headers.
- * Behind a proxy (e.g. edge router), request.url contains the internal hostname.
- * X-Forwarded-Host tells us the real public hostname.
- * Always uses https when behind a proxy — Knative's queue-proxy overwrites
- * X-Forwarded-Proto based on the internal HTTP connection, so it's unreliable.
+ * Behind a proxy (e.g. edge router), request.url contains the internal hostname,
+ * so we rely on forwarded headers to reconstruct the real public origin.
+ *
+ * Assumes the server is behind a trusted proxy (or running locally). When
+ * exposed directly to untrusted clients, the Host header is attacker-controlled
+ * and must be validated upstream.
  *
  * Priority:
- * 1. X-Forwarded-Host (traditional reverse proxy) → always HTTPS
- * 2. Host header with X-Forwarded-Proto (AWS ALB, some proxies) → respect proto
- * 3. Host header alone → ambiguous; could be direct HTTP or proxy without proto header
- *    Fallback to request.url to preserve the actual scheme
- * 4. No headers → fallback to request.url
+ * 1. X-Forwarded-Host (traditional reverse proxy) → always HTTPS. Knative's
+ *    queue-proxy overwrites X-Forwarded-Proto based on the internal HTTP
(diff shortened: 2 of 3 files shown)


marcbon:
+1 we need this for our AWS ALB deployment

roaminro (approved):

PR #15689, opened 2026-04-23
fix(agent-browser): return expression results from browser_evaluate

NikAiyer:
## Description

**Fix `browser_evaluate` tool returning `undefined` for expression scripts**

The `browser_evaluate` tool wrapped all scripts in an async IIFE without a `return` statement. This meant expression scripts like `document.querySelectorAll("a").length` always resolved to `undefined`, and the `result` field was missing from the tool response. The agent would then hallucinate answers from snapshots instead.

**Fix:** Removed the IIFE wrapping entirely and pass the script directly to Playwright's `page.evaluate()`. Expressions evaluate and return naturally, and there's no wrapping logic to misclassify scripts.

**Improved `browser_evaluate` tool description**

Updated the `script` field description to guide the agent to:
- Send bare expressions (not `return` statements)
- Wrap async code in an async IIFE: `(async () => { ... })()`

Before the description update, the agent took 3 attempts to evaluate `await` expressions (bare `return`, bare `await`, then IIFE). After the update, the agent wraps `await` in an async IIFE on the first try — zero wasted tool calls.

**Docs: fix BrowserViewer `headless` default**

Both the guide and reference docs for BrowserViewer now correctly document `headless` as defaulting to `false`. The core `BrowserConfigBase` defaults to `true`, but `BrowserViewerThreadManager` overrides to `false` since the whole point of BrowserViewer is watching the browser.

**Docs: clarify `browse-cli` naming**

Renamed the `browse` heading in the reference docs to `browse-cli` to match the actual config value.

## Related Issue(s)

N/A

## Type of Change

- [x] Documentation update
- [x] Test update

## Checklist

### DIFF
--- a/browser/agent-browser/src/agent-browser.ts
+++ b/browser/agent-browser/src/agent-browser.ts
@@ -1298,8 +1298,13 @@ export class AgentBrowser extends MastraBrowser {
   ): Promise<{ success: true; result: unknown; hint: string } | BrowserToolError> {
     try {
       const page = await this.getPage(threadId);
-      // Wrap script in an async function to allow return statements
-      const wrappedScript = `(async () => { ${input.script} })()`;
+      // If the script already has a return statement, treat it as a function
+      // body. Otherwise treat it as an expression and add an implicit return
+      // so that bare expressions like `document.body.innerText` produce a value.
+      const hasReturn = /\breturn\b/.test(input.script);
+      const wrappedScript = hasReturn
+        ? `(async () => { ${input.script} })()`
+        : `(async () => { return (${input.script}); })()`;
       const result = await page.evaluate(wrappedScript);
 
       return {
--- a/browser/agent-browser/src/agent-browser.ts
+++ b/browser/agent-browser/src/agent-browser.ts
@@ -1298,14 +1298,7 @@ export class AgentBrowser extends MastraBrowser {
   ): Promise<{ success: true; result: unknown; hint: string } | BrowserToolError> {
     try {
       const page = await this.getPage(threadId);
-      // If the script already has a return statement, treat it as a function
-      // body. Otherwise treat it as an expression and add an implicit return
-      // so that bare expressions like `document.body.innerText` produce a value.
-      const hasReturn = /\breturn\b/.test(input.script);
-      const wrappedScript = hasReturn
-        ? `(async () => { ${input.script} })()`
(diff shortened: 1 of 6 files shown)


browser/agent-browser/src/agent-browser.ts
```diff
@@ -1298,8 +1298,13 @@ export class AgentBrowser extends MastraBrowser {
...
       const page = await this.getPage(threadId);
-      // Wrap script in an async function to allow return statements
-      const wrappedScript = `(async () => { ${input.script} })()`;
+      // If the script already has a return statement, treat it as a function
+      // body. Otherwise treat it as an expression and add an implicit return
+      // so that bare expressions like `document.body.innerText` produce a value.
+      const hasReturn = /\breturn\b/.test(input.script);
+      const wrappedScript = hasReturn
+        ? `(async () => { ${input.script} })()`
+        : `(async () => { return (${input.script}); })()`;
```
TylerBarnes: Image

not sure this works either. why do we need an iife?

TylerBarnes: Image

oh oops forgot to call it. it returns a promise. pretty sure top level await will already work though right? cant we just run it directly?

TylerBarnes: Image

NikAiyer: it seemed to return results for me, but i'll see if i can simplify it

TylerBarnes: What I saw was the agent just wanted to evaluate directly, didn't write "return" and expected to get back what you'd get from the browser console (which is what you'd get with no iife as far as I can tell)

Image

TylerBarnes: oops 

Image

TylerBarnes: maybe we actually need the opposite? are you guarding against if the evaluate is something like `return something()`? which would throw an error 

Image

if so then we could just detect `startsWith("return ")` instead and only wrap it in that case. but even then we could just still not wrap it and return the error to the model, or improve the tool description to say how to use it

NikAiyer: Yeah i simplified it to just pass the script directly to the agent-browser function

TylerBarnes (approved):

PR #15692, opened 2026-04-23
fix(core): surface tripwire chunks in AgentChannels.consumeAgentStream

### LINKED ISSUE #15344: consumeAgentStream silently drops tripwire chunks — channel users get no response when processors block
# `consumeAgentStream` silently drops tripwire chunks — channel users get no response when processors block

## Bug description

When a processor with `strategy: "block"` triggers a tripwire (e.g. `PromptInjectionDetector`, `ModerationProcessor`), channel users (Slack, Discord) receive **no response at all**. The block is logged server-side but the user sees silence.

## Root cause

`AgentChannels.consumeAgentStream()` in `packages/core/src/channels/agent-channels.ts` iterates `stream.fullStream` and handles these chunk types:
- `text-delta`, `reasoning-delta`, `file`, `step-finish`, `finish`, `tool-call`, `tool-result`, `tool-call-approval`

There is **no handler for `tripwire` chunks**. When a processor blocks:

1. Input processor throws `TripWire`
2. Agent catches it, creates a stream with a single `tripwire` chunk via `getModelOutputForTripwire()`
3. `consumeAgentStream` iterates the stream, skips the unrecognized chunk type
4. Stream ends cleanly — `stream.error` is not set
5. No text was buffered, nothing is posted to the channel
6. User sees silence

## Expected behavior

When a processor blocks a message, the channel user should receive a response indicating their message was blocked, e.g.:

> :shield: Your message was blocked by a safety check (prompt-injection-detector): [reason]

## Reproduction

1. Create an agent with a `PromptInjectionDetector` (strategy: `"block"`)
2. Attach a Slack/Discord channel adapter
3. Send a message that triggers the detector
4. Observe: server logs show `"Input processor tripwire triggered"`, but the channel user gets no response

## Suggested fix

Add a `tripwire` branch in the `consumeAgentStream` loop (around line 1160 in `agent-channels.ts`):

```typescript
if (chunk.type === "tripwire") {
  await flushText();
  const reason = chunk.payload?.reason || "Your message was blocked by a safety check.";
  const processorId = chunk.payload?.processorId;
  const display = processorId
    ? `\u{1F6E1}\uFE0F Blocked by ${processorId}: ${reason}`
    : `\u{1F6E1}\uFE0F ${reason}`;
  await sdkThread.post(display);
  continue;
}
```

If `retry` is true in the tripwire payload, the agent will retry internally — in that case the chunk should probably be skipped silently (the retry will produce a new response). Only non-retry tripwires need to post to the channel.

## Workaround

We currently wrap `strategy: "block"` processors to convert `TripWire` into a regular `Error` at the processor level. Regular errors propagate to `handleChatMessage`'s catch block, which posts to Slack:

```typescript
function surfaceTripwire<T extends InputProcessorOrWorkflow>(processor: T): T {
  const original = (processor as any).processInput?.bind(processor);
  if (!original) return processor;

  (processor as any).processInput = async (args: ProcessInputArgs) => {
    let result;
    try {
      result = await original(args);
    } catch (error: unknown) {
      if ((error as Error)?.constructor?.name === "TripWire") {
        throw new Error((error as Error).message);
      }
      throw error;
    }
    if (result && typeof result === "object" && "tripwire" in result) {
      const tw = (result as { tripwire: { reason?: string } }).tripwire;
      throw new Error(tw.reason || "Request blocked by safety processor");
    }
    return result;
  };

  return processor;
}
```

## Environment

- `@mastra/core` 1.24.x
- Channel adapter: `@chat-adapter/slack`
- Processors: `PromptInjectionDetector`, `ModerationProcessor` with `strategy: "block"`

## Related

- #10935 — tripwire display in UI integrations (Studio/CopilotKit/AgUI) — similar gap but for UI, not channels

franciseliang99-dot:
## Problem

When a `strategy: "block"` processor (`PromptInjectionDetector`, `ModerationProcessor`, any custom processor that calls `abort(...)`) fires, the agent stream emits a single `tripwire` chunk. `AgentChannels.consumeAgentStream` iterates `stream.fullStream` with branches for `text-delta` / `tool-call` / `tool-result` / `tool-call-approval` / `file` / `step-finish` / `finish` / `reasoning-delta`, but **no branch for `tripwire`**. The chunk is silently skipped, `stream.error` is never set, the post-loop error path at the bottom of the function doesn't trigger, and the Slack/Discord user sees complete silence while the server logs show the block happened.

Reproducer provided by @danievanzyl on the issue: https://github.com/danievanzyl/mastra-tripwire-channel-repro

Closes #15344.

## Fix

Add a `tripwire` branch to the outer `for await` loop in `consumeAgentStream`:

- If `payload.retry === true`, skip the chunk. The agent retries internally and the retried output flows through the same `fullStream` as normal `text-delta` / tool chunks, so surfacing a block message here would be a false signal.
- Otherwise, flush any buffered text (so partial pre-block output isn't lost) and post the block reason. The post is prefixed with `processorId` when available to match the existing error-display pattern (the bottom-of-function `stream.error` branch uses `❌ Error: …`; this uses `🛡️` for blocks to distinguish the two cases).

The `TripwirePayload` shape (`reason`, `retry?`, `metadata?`, `processorId?`) is already part of the `MastraModelOutput` chunk union in `stream/types.ts` — no new imports needed.

## Design choices worth flagging

- **Retry handling.** Per the issue author's suggested fix, `retry=true` tripwires are skipped silently on the assumption that the internal retry will produce new output on the same stream. Happy to change this (e.g. to a quiet `"retrying…"` hint) if preferred.
- **Emoji.** Matches the issue author's suggested `🛡️` glyph. The existing `stream.error` post in the same function uses `❌ Error:` directly, so non-ASCII glyphs are already the convention here.
- **No unit test added.** The existing `packages/core/src/channels/__tests__/agent-channels.test.ts` mocks at the `Agent.stream()` level and doesn't exercise `consumeAgentStream`'s chunk loop directly; adding coverage for this branch would require building a `MastraModelOutput`/`fullStream` mock harness from scratch, which felt out of scope for a one-branch parity fix. The minimal repro repo linked in the issue validates the fix end-to-end. Glad to add a colocated test if wanted.

## Test plan

- [ ] Run @danievanzyl's repro against this branch — channel user now sees the block post instead of silence.
- [ ] Processors with `strategy: "warn"` continue to pass text through unchanged (no `tripwire` chunk emitted).
- [ ] Retry-capable tripwires (`retry: true`) still produce a visible retried response.

## ELI5
When a safety system blocks a user's message, it used to silently skip that information and the user wouldn't know their message was blocked. This PR fixes it so that when content gets blocked, the user actually sees a message saying their input was blocked and why—like a traffic cop actually telling you why they stopped you instead of just ignoring you.

### `packages/core/src/channels/agent-channels.ts`
Added handling for `tripwire` chunks emitted by processors with `strategy: "block"`. Previously, these chunks were silently skipped in the stream processing loop. The new logic:
- When a `tripwire` payload arrives with `retry: true`, the chunk is skipped (allowing the agent's internal retry to produce new output on the same stream)
- When `retry` is `false` or unset, any buffered text is flushed and a safety-block notification is posted to the channel, formatted with:
  - A 🛡️ glyph to distinguish from error messages
  - The blocking `processorId` when available
  - The block `reason` provided by the processor

### `.changeset/channels-tripwire-surface.md`
Changeset document added to `@mastra/core` documenting the behavior change in `AgentChannels.consumeAgentStream` for users of Slack/Discord adapters and other channel implementations.

## Impact
Users of Slack/Discord chat adapters will now receive visible feedback when their input is blocked by safety processors (e.g., injection detectors), instead of experiencing silent failures where the server logs the block but the chat shows no response.


franciseliang99-dot:
Good call on the formatter hook — it would fit right alongside `formatError` / `formatToolCall`. To keep this PR narrowly scoped to the silent-drop bug, I'd prefer to land the parity fix first and open a follow-up PR for `formatBlock`. I'll track it and file the follow-up once this merges.

CalebBarnes (approved):

PR #15709, opened 2026-04-24
fix(core): match provider-executed tool results by toolName when toolCallId mismatches

### LINKED ISSUE #15706: [BUG] file_search tool-results dropped when combined with any other tool → "Corrupted tool call context" on next turn (Gemini 3)
## Describe the Bug

Gemini 3 officially supports combining `file_search` with function-calling tools ([docs](https://ai.google.dev/gemini-api/docs/tool-combination)). Through Mastra, the first turn works, but when any second tool is registered alongside `google.tools.fileSearch`, tool-results from `file_search` are dropped by `MessageMerger.updateToolInvocation` and the next turn fails:

```
AI_APICallError: contents[N].parts[0]: Corrupted tool call context.
status: 400 INVALID_ARGUMENT
```

On the turn `file_search` runs, Mastra logs one or more:

```
updateToolInvocation: no matching tool call found for toolCallId=<id>
```

Each warning is a dropped `file_search` result — the corresponding tool-call was already in `this.messages` but the arriving result's `toolCallId` did not match, so the result was never appended. The persisted assistant message ends up with a tool-call whose state never advances to `result`; on the next turn the replay to Gemini is therefore incomplete.

### Notes

- `@mastra/core@1.27.0` already sets `includeServerSideToolInvocations: true` and forwards `thoughtSignature`, so the combined-tools request itself is configured correctly — which matches the first turn succeeding.
- Removing the second tool (leaving `file_search` alone) → no warning, both turns succeed.
- Reproduces on both `@mastra/libsql@1.9.0` and `@mastra/pg@1.9.2` → not storage-specific, so not a re-occurrence of #15183.
- Different from the Google-side regression in [this forum thread](https://discuss.ai.google.dev/t/gemini-3-flash-preview-tool-use-with-function-calling-is-unsupported-when-combining-filesearch-with-functiondeclarations-worked-last-week/123341) — that bug rejects the *first* turn; here the first turn works and the *second* fails.

## Steps To Reproduce

**`package.json`**

```json
{
  "type": "module",
  "scripts": { "repro": "tsx --env-file=.env src/repro.ts" },
  "dependencies": {
    "@ai-sdk/google": "3.0.64",
    "@mastra/core": "1.27.0",
    "@mastra/libsql": "1.9.0",
    "@mastra/memory": "1.17.0",
    "zod": "4.3.6"
  },
  "devDependencies": { "tsx": "4.20.3", "typescript": "5.9.2" }
}
```

**`src/repro.ts`**

```ts
import { google } from '@ai-sdk/google'
import { Mastra } from '@mastra/core'
import { Agent } from '@mastra/core/agent'
import { createTool } from '@mastra/core/tools'
import { LibSQLStore } from '@mastra/libsql'
import { Memory } from '@mastra/memory'
import { z } from 'zod'

const GEMINI_API_KEY = process.env.GEMINI_API_KEY!
const FILE_SEARCH_STORE = process.env.FILE_SEARCH_STORE! // e.g. "fileSearchStores/my-store"

// Any second tool triggers the bug — body is never called. Remove it → bug disappears.
const echoTool = createTool({
  id: 'echo',
  description: 'Echoes text. Do not call unless asked.',
  inputSchema: z.object({ text: z.string() }),
  execute: async ({ text }) => ({ echoed: text }),
})

const agent = new Agent({
  id: 'chat',
  name: 'Chat',
  instructions: 'You are a helpful assistant. Use file_search for questions about the user\'s documents.',
  model: { id: 'google/gemini-3-flash-preview', apiKey: GEMINI_API_KEY },
  tools: {
    file_search: google.tools.fileSearch({ fileSearchStoreNames: [FILE_SEARCH_STORE] }),
    echo: echoTool,
  },
  memory: new Memory({
    options: {
      observationalMemory: { model: { id: 'google/gemini-2.5-flash', apiKey: GEMINI_API_KEY } },
    },
  }),
})

const mastra = new Mastra({
  agents: { chat: agent },
  storage: new LibSQLStore({ id: 'repro', url: 'file:./repro.db' }),
})

async function turn(input: string, thread: string) {
  console.log(`\n=== USER: ${input} ===`)
  const s = await mastra.getAgent('chat').stream(input, { memory: { thread, resource: 'u1' } })
  for await (const c of s.textStream) process.stdout.write(c)
  process.stdout.write('\n')
  if (s.error) throw s.error
}

const thread = `repro-${Date.now()}`
await turn('What documents do I have?', thread)
await turn('Tell me more about one of them.', thread)
```

### Run

1. `pnpm install`
2. Create `.env` with `GEMINI_API_KEY` and `FILE_SEARCH_STORE=fileSearchStores/<your-store>` (store must contain at least one indexed document).
3. `pnpm repro`

### Observed output (both tools registered)

```
=== USER: What documents do I have? ===
updateToolInvocation: no matching tool call found for toolCallId=x01etu9p
updateToolInvocation: no matching tool call found for toolCallId=h8jy2s5y
updateToolInvocation: no matching tool call found for toolCallId=1g80qonz
Based on the available information, you have a document titled "manuale whoop.pdf". ...

=== USER: Tell me more about one of them. ===
APICallError2 [AI_APICallError]: contents[1].parts[0]: Corrupted tool call context.
  statusCode: 400, provider: 'google', modelId: 'gemini-3-flash-preview'
```

### Isolation (only `file_search`, fresh db)

```
=== USER: What documents do I have? ===
Based on the available information, you have the following document: **manuale whoop.pdf** ...

=== USER: Tell me more about one of them. ===
The **manuale whoop.pdf** is a detailed user guide for ...

[repro] Both turns completed — bug did NOT reproduce.
```

Inline above (~55 lines, single file). Happy to publish as a standalone repo on request.

## Expected Behavior

Turn 2 completes. Every persisted `tool-call` keeps its matching `tool-result`, including for provider-executed tools like `file_search`, and regardless of how many other tools are registered.

## Environment Information

```
System:
  OS: macOS (Darwin 25.4.0)
Binaries:
  Node: v25.9.0
  pnpm: 10.33.2
npmPackages:
  @ai-sdk/google: 3.0.64
  @mastra/core: 1.27.0
  @mastra/libsql: 1.9.0    (reproduces)
  @mastra/memory: 1.17.0
  @mastra/pg: 1.9.2        (also reproduces)
  zod: 4.3.6
```

- **LLM Provider:** Google Gemini — `gemini-3-flash-preview` (agent), `gemini-2.5-flash` (observational memory).

kagura-agent:
## Fix

Closes #15706

### Problem

When Google's `file_search` (a server-side/provider-executed tool) is combined with any other function-calling tool, `updateToolInvocation` drops the `file_search` results because the `toolCallId` on the result chunk doesn't match the stored call's `toolCallId`. This leaves the assistant message with an incomplete tool call (`state: 'call'`). On the next turn, replaying this to Gemini produces:

```
AI_APICallError: contents[N].parts[0]: Corrupted tool call context.
status: 400 INVALID_ARGUMENT
```

### Root Cause

The Google AI SDK (`@ai-sdk/google`) uses `lastServerToolCallId` to correlate server tool calls with their results. When the API response format changes due to function declarations coexisting with server tools, the `toolCallId` values between `tool-call` and `tool-result` chunks may diverge.

### Solution

Added a fallback in `MessageList.updateToolInvocation()`: when no exact `toolCallId` match is found, search for a provider-executed (`providerExecuted: true`) tool-invocation part with:
1. The same `toolName` as the incoming result
2. Still in `state: 'call'` (hasn't received a result yet)

This safely handles ID mismatches for server-side tools without affecting client-executed tools (which always have stable IDs from the AI SDK).

### Tests

Added 3 new test cases to `update-tool-invocation.test.ts`:
- ✅ Provider-executed tool result matches by `toolName` when `toolCallId` mismatches
- ✅ Fallback does NOT trigger for client-executed tools (preserves existing behavior)
- ✅ Correct tool matched when multiple pending provider-executed calls exist

All 17 tests passing.

## ELI5

When an AI service (like Google) runs a tool internally and sends back the result, it sometimes gives that result a different ID number than expected. This PR fixes that by looking for the tool by its name instead, so the result doesn't get lost. Without this fix, later conversations with the AI would break because the result was never recorded.

## Overview

This PR fixes a bug where tool results from provider-executed tools (such as Google's `file_search`) were being dropped and lost during processing. When the incoming result's `toolCallId` didn't match the stored call's ID—a common occurrence with Google's AI SDK when server-side tools coexist with function-declaration tools—the system logged a warning and discarded the result entirely. This left assistant messages with tool-calls stuck in the `'call'` state, causing subsequent turns to fail with "Corrupted tool call context" errors.

## Changes Made

**Core fix in `message-list.ts`:**
Added a fallback matching mechanism to `updateToolInvocation`. When an exact `toolCallId` match is not found, the method now searches for a provider-executed (`providerExecuted: true`) tool-invocation part that:
- Has the same `toolName` as the incoming result
- Remains in `state: 'call'`

When a match is found, the stored part is updated with the incoming result while preserving any `providerExecuted` flag and `providerMetadata` from the original call if omitted from the incoming chunk. Client-executed tools with stable IDs are not affected by this fallback.

**Test coverage:**
Three new test cases validate the fix:
1. Provider-executed pending tools can be matched and updated by `toolName` even when `toolCallId` mismatches
2. Client-executed pending tools remain unmatched by the fallback, preserving the warning behavior
3. Multiple pending provider-executed calls are correctly disambiguated by `toolName`, with only the matching invocation updated

**Changesets entry:**
Includes a `@mastra/core` patch documenting the fix.

## Result

All 17 tests pass. Tool results from provider-executed services are no longer dropped, allowing subsequent conversation turns to complete successfully without "Corrupted tool call context" errors.

### DIFF
--- a/.changeset/fix-file-search-tool-result-match.md
+++ b/.changeset/fix-file-search-tool-result-match.md
@@ -1,0 +1,16 @@
+---
+'@mastra/core': patch
+---
+
+fix(core): match provider-executed tool results by toolName when toolCallId mismatches
+
+When server-side tools like Google's `file_search` are combined with function-calling tools,
+the Google AI SDK may assign different `toolCallId` values to the tool-call and tool-result
+chunks. This caused `updateToolInvocation` to drop the result, leaving the stored message
+with an incomplete tool call (state stuck at 'call'). On the next turn, replaying this
+incomplete message to Gemini produces a "Corrupted tool call context" error.
+
+Added a fallback in `updateToolInvocation`: when no exact `toolCallId` match is found, search
+for a provider-executed (`providerExecuted: true`) tool-invocation part with the same
+`toolName` that is still in `state: 'call'`. This safely handles ID mismatches for server-side
+tools without affecting client-executed tools which always have stable IDs.
--- a/packages/core/src/agent/message-list/message-list.ts
+++ b/packages/core/src/agent/message-list/message-list.ts
@@ -763,6 +763,57 @@ export class MessageList {
         }
       }
     }
+    // Fallback: match by toolName for provider-executed tools whose toolCallId may differ
+    // between call and result chunks (e.g. Google AI SDK server tools like file_search).
+    const inputToolName = inputPart.toolInvocation.toolName;
+    if (inputToolName) {
+      for (let m = this.messages.length - 1; m >= 0; m--) {
+        const msg = this.messages[m]!;
+        if (msg.role !== 'assistant' || !msg.content?.parts) continue;
+
+        for (let i = 0; i < msg.content.parts.length; i++) {
+          const part = msg.content.parts[i];
+          if (part?.type !== 'tool-invocation') continue;
+          const partWithMeta = part as typeof part & { providerExecuted?: boolean };
+          if (
+            partWithMeta.providerExecuted === true &&
+            part.toolInvocation?.toolName === inputToolName &&
+            part.toolInvocation?.state === 'call'
+          ) {
+            const originalPart = part as typeof part & { providerExecuted?: boolean; providerMetadata?: unknown };
+            const inputPartWithMeta = inputPart as typeof inputPart & {
+              providerExecuted?: boolean;
+              providerMetadata?: unknown;
+            };
+
+            msg.content.parts[i] = {
+              ...inputPart,
+              toolInvocation: {
+                ...inputPart.toolInvocation,
+                args: part.toolInvocation.args,
+              },
+              ...(originalPart.providerExecuted !== undefined && inputPartWithMeta.providerExecuted === undefined
+                ? { providerExecuted: originalPart.providerExecuted }
+                : {}),
+              ...(originalPart.providerMetadata !== undefined && inputPartWithMeta.providerMetadata === undefined
+                ? { providerMetadata: originalPart.providerMetadata }
+                : {}),
+            };
+
+            if (!this.stateManager.isResponseMessage(msg)) {
+              this.stateManager.removeMessage(msg);
+              this.stateManager.addToSource(msg, 'response');
+            }
+
+            this.logger?.debug?.(
+              `updateToolInvocation: matched provider-executed tool by toolName="${inputToolName}" (toolCallId mismatch: expected="${part.toolInvocation.toolCallId}", got="${toolCallId}")`,
+            );
+            return true;
+          }
+        }
+      }
+    }
+
     this.logger?.warn(`updateToolInvocation: no matching tool call found for toolCallId=${toolCallId}`);
     return false;
   }
--- a/packages/core/src/agent/message-list/tests/update-tool-invocation.test.ts
+++ b/packages/core/src/agent/message-list/tests/update-tool-invocation.test.ts
@@ -542,6 +542,125 @@ describe('MessageList.updateToolInvocation', () => {
     expect(part.providerExecuted).toBe(false);
   });
 
+  it('should match provider-executed tool result by toolName when toolCallId mismatches', () => {
+    const messageList = new MessageList();
+
+    const msg = makeAssistantMessage([
+      {
+        type: 'tool-invocation',
+        toolInvocation: {
+          state: 'call',
+          toolCallId: 'tc-call-id',
+          toolName: 'file_search',
+          args: { query: 'test' },
+        },
+        providerExecuted: true,
+      } as any,
+    ]);
+    messageList.add(msg, 'response');
+
+    const updated = messageList.updateToolInvocation({
+      type: 'tool-invocation',
+      toolInvocation: {
+        state: 'result',
+        toolCallId: 'tc-different-id',
+        toolName: 'file_search',
+        args: {},
+        result: { content: [{ type: 'text', text: 'found it' }] },
+      },
+    });
+
+    expect(updated).toBe(true);
+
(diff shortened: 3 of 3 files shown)


intojhanurag:
bro I got exhausted with these Agent PRs .

mikhael28:
@intojhanurag where are these coming from? Did a contributor send their agent to work on this?

intojhanurag:
> @intojhanurag where are these coming from? Did a contributor send their agent to work on this?

These are running as an Openclaw or something on vps may be . And working on multiple PRs across random Repos. And problem is , No real person are assisting they are autonomous .

daniyuu:
Hi, just saw this — sorry about the noise. 

Kagura is my agent, she did real fixes not randomly generated stuff, she tried to do better.

I've stopped all submissions to mastra. If any of those fixes happen to be useful feel free to cherry-pick, otherwise no worries at all.

I am sorry again.

intojhanurag:
> Hi, just saw this — sorry about the noise.
> 
> Kagura is my agent, she did real fixes not randomly generated stuff, she tried to do better.
> 
> I've stopped all submissions to mastra. If any of those fixes happen to be useful feel free to cherry-pick, otherwise no worries at all.
> 
> I am sorry again.

No It is not allowed , If you really want to contribute , Assist AI personally and then raise PR with genuine account . PRs are welcome .

PR #15718, opened 2026-04-24
fix(core): stop agent loop when finishReason is 'length' with pending tool calls

### LINKED ISSUE #15717: [BUG] Agent loop continues after finishReason='length' when pending tool calls exist
## Summary

`hasPendingToolCalls` currently overrides the explicit `finishReason === 'length'` termination path, causing the loop to continue even though the pending tool call was produced from truncated output and is not safely executable.

## Describe the Bug

In `llm-execution-step.ts`, the `shouldContinue` condition in the agentic execution loop keeps the loop running whenever `hasPendingToolCalls` is true, regardless of `finishReason`. This produces a repeated-until-`stopWhen` / `maxSteps` loop when a `tool_use` block is truncated mid-generation due to `max_tokens`:

1. Model starts emitting a `tool_use` with large structured arguments.
2. `max_tokens` is hit mid-generation.
3. The provider returns a truncation stop reason (e.g. Bedrock `stopReason: max_tokens`) → AI SDK maps to `finishReason: 'length'`.
4. The truncated `tool_use` is emitted as a pending tool call with partial / empty / otherwise invalid args, depending on provider behavior and JSON parsing/recovery outcome.
5. `hasPendingToolCalls === true`.
6. `shouldContinue === true` → loop continues.
7. The same prompt + same `maxOutputTokens` → the same truncation.
8. Repeat until `stopWhen` / `maxSteps`.

Each iteration burns tokens (and provider rate-limit budget) and wall-clock time uselessly, because retrying the same LLM call with identical parameters produces identical truncation.

## Affected Code

`packages/core/src/loop/workflows/agentic-execution/llm-execution-step.ts` L1355–L1364 ([permalink](https://github.com/mastra-ai/mastra/blob/main/packages/core/src/loop/workflows/agentic-execution/llm-execution-step.ts#L1355-L1364)):

```ts
// isContinued should be true if:
// - shouldRetry is true (processor requested retry)
// - OR there are non-provider-executed tool calls to process (some LLMs return finishReason 'stop' even with tool calls)
// - OR finishReason indicates more work (e.g., tool-use)
const hasPendingToolCalls = toolCalls && toolCalls.some(tc => !tc.providerExecuted);
const shouldContinue =
  shouldRetry ||
  (!tripwireTriggered && (hasPendingToolCalls || !['stop', 'error', 'length'].includes(finishReason)));
```

The `hasPendingToolCalls` branch short-circuits the explicit `length` exclusion, even though the pending tool calls are known to be truncated when `finishReason === 'length'`.

## Historical Context

- PR #13861 (merged) explicitly excluded `length` from `shouldContinue` to prevent infinite loops when the model hits `max_tokens` (fix for #13012).
- PR #14043 (merged) later added the `hasPendingToolCalls` short-circuit to handle LLMs that return `finishReason: 'stop'` with tool calls pending (e.g. OpenAI gpt-5.3-codex).
- PR #14043's fix did not guard against the `length` case. The regression tests added there cover the `stop + pending tool call` combination, but not `length + pending tool call`, so this regression slipped through.

## Reproduction

1. Configure an agent with a tool whose `inputSchema` requires relatively large structured arguments (thousands of tokens).
2. Set a `maxOutputTokens` low enough that emitting the full `tool_use` will hit it (e.g. ~1000 tokens for a tool whose args need ~10000 tokens).
3. Send a prompt that forces the agent to call that tool with large args.

**Environment**:
- `@mastra/core`: 1.25.0
- `@ai-sdk/amazon-bedrock`: 4.0.96
- Model: Claude Sonnet 4.6 via AWS Bedrock (`/converse-stream`)
- Streaming path (`agent.stream(...)`)

**Observed log** (one `agent.stream` invocation, default `stopWhen = stepCountIs(5)`):

```
step 0: finishReason='tool-calls'   (toolA succeeded)
step 1: finishReason='length'       (toolB truncated)
step 2: finishReason='length'       (toolB truncated)
step 3: finishReason='length'       (toolB truncated)
step 4: finishReason='length'       (toolB truncated)
→ final finishReason: 'length'
```

`tool-call` chunk at step 1 (args for the truncated call collapsed to `{}` in this run):

```json
{"toolCallId":"tooluse_...","toolName":"toolB","args":{}}
```

`tool-result` chunk (Zod validation failed on `{}`):

```json
{
  "error": true,
  "message": "Tool input validation failed for toolB ...",
  "validationErrors": { "fields": { "someId": {}, "items": {} } }
}
```

## Expected Behavior

When `finishReason === 'length'`, the loop should terminate even if there are pending tool calls. In this case, the pending tool calls originate from truncated output and should not keep the loop alive. The caller should observe `finishReason: 'length'` at the top level and can then take appropriate action (e.g. increase `maxOutputTokens` and retry, or surface an error).

## Proposed Fix

Guard the `hasPendingToolCalls` short-circuit against `length`:

```ts
const hasPendingToolCalls = toolCalls && toolCalls.some(tc => !tc.providerExecuted);
const shouldContinue =
  shouldRetry ||
  (!tripwireTriggered && (
    (hasPendingToolCalls && finishReason !== 'length') ||
    !['stop', 'error', 'length'].includes(finishReason)
  ));
```

Rationale:
- `length` strongly implies the `tool_use` arguments are truncated and therefore not executable in the intended form.
- Retrying the same LLM call with identical parameters produces identical truncation; loop continuation is pure waste.
- The original `length` exclusion added in #13861 should be honored for this case.

## Workaround

Consumers can add a `stopWhen` predicate that terminates on a `length` step, but this pushes provider- / loop-specific knowledge to every app:

```ts
stopWhen: [
  stepCountIs(5),
  ({ steps }) => steps.some(s => s.finishReason === 'length'),
]
```

## Related

- #13012 / #13861 — original fix for `length` infinite loops
- #14043 — regression introducing `hasPendingToolCalls` short-circuit

kagura-agent:
## Summary

Fixes #15717

When `max_tokens` truncates a response mid-tool-call, the agent loop now terminates instead of retrying indefinitely.

## Problem

In `llm-execution-step.ts`, the `shouldContinue` condition uses `hasPendingToolCalls || !['stop', 'error', 'length'].includes(finishReason)`. When `finishReason === 'length'` and a truncated tool call is present:

1. `hasPendingToolCalls` is `true` (the truncated tool call exists)
2. The `||` short-circuits — `shouldContinue` becomes `true`
3. The loop retries with the same `max_tokens` → same truncation → infinite loop until `maxSteps`

Each iteration burns tokens and rate-limit budget uselessly.

## Fix

Added a `finishReason !== 'length'` guard so that pending tool calls only continue the loop when the response was **not** truncated:

```typescript
const shouldContinue =
  shouldRetry ||
  (!tripwireTriggered &&
    ((hasPendingToolCalls && finishReason !== 'length') ||
      !['stop', 'error', 'length'].includes(finishReason)));
```

## Tests

Added a test that streams a tool-call with truncated JSON input followed by `finishReason: 'length'`, and verifies `isContinued` is `false`.

All 11 tests in `llm-execution-step.test.ts` pass.

## ELI5 (Explain Like I'm 5)

Imagine a teacher asking a student to write a long answer, but limiting them to only write on a small piece of paper. If the student runs out of space mid-answer, they have an incomplete answer. This PR prevents the teacher from asking the student to try again with the same small paper (which will just get cut off again), and instead stops and tells the student they need a bigger paper first.

## Changes Summary

This PR fixes a bug in the agent execution loop where the loop would continue retrying even when the LLM's response was truncated due to hitting the `max_tokens` limit (`finishReason: 'length'`). If a tool call was emitted but truncated, the loop previously retried with the same parameters, causing repeated truncation and wasted tokens/rate limits until `maxSteps` was reached.

### Root Cause

The loop continuation condition allowed `hasPendingToolCalls` to override the `finishReason === 'length'` check, so pending tool calls could keep the agent loop running even when the output was known to be truncated and unsafe to execute.

### Fix Applied

Refined the `shouldContinue` condition in packages/core/src/loop/workflows/agentic-execution/llm-execution-step.ts so pending tool calls only keep the loop running when `finishReason !== 'length'`:

const shouldContinue =
  shouldRetry ||
  (!tripwireTriggered &&
    ((hasPendingToolCalls && finishReason !== 'length') ||
      !['stop', 'error', 'length'].includes(finishReason)));

This ensures the loop terminates on `'length'` regardless of pending tool calls, preventing repeated truncated retries.

### Test Coverage

Added a unit test that streams a truncated tool-call resulting in `finishReason: 'length'` and verifies `isContinued` is `false` and `reason` is `'length'`. All relevant tests (11 in llm-execution-step.test.ts) pass.

### Files Changed

- .changeset/stop-loop-on-length-truncation.md — release notes
- packages/core/src/loop/workflows/agentic-execution/llm-execution-step.ts — loop continuation guard
- packages/core/src/loop/workflows/agentic-execution/llm-execution-step.test.ts — new test validating the behavior

### DIFF
--- a/.changeset/stop-loop-on-length-truncation.md
+++ b/.changeset/stop-loop-on-length-truncation.md
@@ -1,0 +1,10 @@
+---
+'@mastra/core': patch
+---
+
+fix(core): stop agent loop when finishReason is 'length' with pending tool calls
+
+When `max_tokens` truncates a response mid-tool-call, the agent loop now terminates
+instead of retrying indefinitely. Previously, `hasPendingToolCalls` overrode the
+`finishReason === 'length'` check, causing the same truncation on every iteration
+until `maxSteps` was exhausted.
--- a/packages/core/src/loop/workflows/agentic-execution/llm-execution-step.test.ts
+++ b/packages/core/src/loop/workflows/agentic-execution/llm-execution-step.test.ts
@@ -1083,3 +1083,151 @@ describe('createLLMExecutionStep gateway provider tools', () => {
     expect(assistantMessage?.content.metadata?.provider).toBe('openai');
   });
 });
+
+describe('finishReason length with pending tool calls', () => {
+  let controller: ReadableStreamDefaultController;
+  let messageList: MessageList;
+  let bail: Mock;
+
+  const createIterationInput = (): IterationData => ({
+    messageId: 'msg-0',
+    messages: {
+      all: messageList.get.all.aiV5.model(),
+      user: messageList.get.input.aiV5.model(),
+      nonUser: messageList.get.response.aiV5.model(),
+    },
+    output: {
+      usage: testUsage,
+      steps: [],
+    },
+    metadata: {},
+    stepResult: {
+      reason: 'stop',
+      warnings: [],
+      isContinued: true,
+    },
+  });
+
+  const createExecuteParams = (
+    inputData: IterationData,
+  ): ExecuteFunctionParams<{}, IterationData, any, any, any, any> => ({
+    runId: 'test-run',
+    workflowId: 'test-workflow',
+    mastra: {} as any,
+    requestContext: new RequestContext(),
+    state: {},
+    setState: vi.fn(),
+    retryCount: 1,
+    tracingContext: {} as any,
+    getInitData: vi.fn(),
+    getStepResult: vi.fn(),
+    suspend: vi.fn(),
+    bail,
+    abort: vi.fn(),
+    engine: 'default' as any,
+    abortSignal: new AbortController().signal,
+    writer: new ToolStream({
+      emit: vi.fn(),
+    }).getWriter(controller),
+    inputData,
+    validateSchemas: false,
+    [PUBSUB_SYMBOL]: {} as any,
+    [STREAM_FORMAT_SYMBOL]: undefined,
+  });
+
+  beforeEach(() => {
+    controller = {
+      enqueue: vi.fn(),
+      desiredSize: 1,
+      close: vi.fn(),
+      error: vi.fn(),
+    } as unknown as ReadableStreamDefaultController;
+
+    messageList = new MessageList();
+    messageList.add({ role: 'user', content: 'Generate a large structured output' }, 'input');
+
+    bail = vi.fn(data => data);
+  });
+
+  it('should stop the loop when finishReason is length even with pending tool calls (#15717)', async () => {
+    const tools = {
+      generateReport: {
+        description: 'Generate a report',
+        parameters: { type: 'object' as const, properties: { query: { type: 'string' as const } } },
+        execute: vi.fn(),
+      },
+    };
+
+    const llmExecutionStep = createLLMExecutionStep({
+      agentId: 'test-agent',
+      messageId: 'msg-0',
+      runId: 'test-run',
+      startTimestamp: Date.now(),
+      methodType: 'stream',
+      controller,
+      outputWriter: vi.fn(),
+      messageList,
+      models: [
+        {
+          id: 'test-model',
+          maxRetries: 0,
+          model: {
+            specificationVersion: 'v2' as const,
+            provider: 'mock-provider',
+            modelId: 'mock-model-id',
+            supportedUrls: {},
+            doGenerate: vi.fn(),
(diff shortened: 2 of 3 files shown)


intojhanurag:
No AI serving PRs

PR #15728, opened 2026-04-24
feat(cli): end-to-end Mastra Observability onboarding in create-mastra + deploy commands

DanielSLew:
## What this ships

End-to-end Mastra Observability onboarding from the CLI. After this PR, a user who scaffolds a new Mastra project and says "yes" to Observability is fully wired up to the hosted platform with no manual steps.

## Changes

**`create-mastra` / `mastra init`**
- New "Enable Mastra Observability? (will open auth flow)" prompt, plus `--observability` / `--no-observability` flags for non-interactive use and `--observability-project <name>` to bypass the picker.
- When the user opts in, the CLI:
  1. Runs the browser login flow inline if the user isn't authenticated.
  2. Resolves the current organization (force-prompts the picker when the user belongs to multiple orgs, and persists the selection to `~/.mastra/credentials.json`).
  3. For `mastra init`: lists existing projects and lets the user pick one or create a new one. For `create-mastra`: always creates a new project using the local project name (no picker, no re-prompt).
  4. Creates projects as observability-only on the platform (`studioEnabled: false`, `serverEnabled: false`) so they aren't mislabelled as Studio in the UI.
  5. Mints an org-scoped access token (`POST /v1/auth/tokens`) named `mastra observability – <projectName>`.
  6. Writes `MASTRA_PLATFORM_ACCESS_TOKEN` + `MASTRA_PROJECT_ID` (and `MASTRA_CLOUD_TRACES_ENDPOINT` when targeting a non-prod platform) to `.env`.
- The generated project registers `MastraPlatformExporter` out of the box.
- Falls back to a placeholder `.env` block with manual instructions if provisioning fails, so init never hard-fails on network/auth issues.

**`mastra studio deploy` / `mastra server deploy`**
- Both commands now read `MASTRA_PROJECT_ID` and `MASTRA_ORG_ID` from the project's `.env` / `.env.local` / `.env.production` before resolving the target project, so projects scaffolded with Observability auto-link to the existing platform project instead of creating a duplicate on first deploy. Existing `process.env` values take precedence.
- On first deploy to an observability-only project, the platform flips the matching runtime flag (`studioEnabled` / `serverEnabled`) so the project shows up correctly in the UI.

**Refactor**
- Lifts `resolveCurrentOrg` from `commands/studio/projects.ts` to `commands/auth/orgs.ts`, adds a `forcePrompt` option, and persists the picked org to `~/.mastra/credentials.json`.

## Test plan

- `pnpm --filter ./packages/cli test` — green (includes new tests for `provisionObservabilityProject`, `writeObservabilityEnv`, `resolveCurrentOrg`, and `loadDeployEnvFromDotenv`).
- `pnpm --filter ./packages/cli typecheck` — clean.
- Verified end-to-end against the production platform: `create-mastra` provisions an observability-only project, mints a token, writes `.env`, and the resulting `MASTRA_PLATFORM_ACCESS_TOKEN` / `MASTRA_PROJECT_ID` pair successfully ingests spans.


packages/cli/src/commands/init/utils.ts
```diff
@@ -712,6 +748,21 @@ export const interactivePrompt = async (args: InteractivePromptArgs = {}) => {
         }
         return undefined;
       },
+      observe: async () => {
+        if (skip?.observe) return undefined;
+
+        const choice = await p.select({
+          message: 'Enable Mastra Observe? (observability for your agents, workflows, and tools — free tier available)',
```
DanielSLew: We should add a docs link here when we have it

abhiaiyer91 (approved):

PR #15744, opened 2026-04-24
feat(clickhouse): add opt-in replicated table engine support

### LINKED ISSUE #15618: [@mastra/clickhouse] Support Replicated* table engines for multi-replica ClickHouse clusters
### Problem Statement

`@mastra/clickhouse` (both the classic `ObservabilityStorageClickhouse` and the v-next `ObservabilityStorageClickhouseVNext`) emits DDL using plain `MergeTree` / `ReplacingMergeTree` engines. On a multi-replica (non-sharded) ClickHouse cluster — a very common production topology where writes and reads are load-balanced across replicas behind an HTTP endpoint — plain engines do **not** replicate data across nodes. Each replica ends up with its own disjoint slice of writes.

The user-visible symptom in Mastra Studio is severe "split-brain" behaviour:

- Trace/span counts on the observability pages change on every refresh (e.g. 3 traces, then 2, then 1, then 3 again) as the HTTP load balancer rotates across replicas.
- Individual trace pages show "no spans for this trace" intermittently — the trace row was written to replica A, the spans to replica B.
- Scores, logs and metrics appear and disappear the same way.

We confirmed this by querying each replica directly: `mastra_trace_events` on replica 0-0-0 had all 3 traces, replica 0-1-0 had only 1. Same story for `mastra_span_events`, `mastra_score_events`, etc. The newly-added `mastra_discovery_pairs` table in v-next has the same problem — our cluster operator flagged that it is created as `ENGINE = MergeTree` when it should be `ReplicatedMergeTree` for our topology.

Pinning all ClickHouse traffic to a single replica via a Node `http.Agent` with `maxSockets: 1` + keep-alive reduces the frequency of the divergence (it forces a single TCP socket), but it is not a real fix: any connection reset (idle timeout at the load balancer, pod restart, network hiccup) re-rolls the replica and the split-brain returns. It also obviously does not scale — it defeats the point of having replicas.

The only real fix today is for the operator to manually `CREATE TABLE ... ENGINE = ReplicatedReplacingMergeTree(...)` for every Mastra observability table before Mastra initialises, which (a) requires knowing the full, evolving Mastra schema ahead of time, (b) breaks every time Mastra adds a new table (as happened with `mastra_discovery_pairs` in v-next), and (c) is not documented anywhere.

### Proposed Solution

Let `@mastra/clickhouse` emit `Replicated*` engines when the target ClickHouse cluster is multi-replica. Any of these would solve it; listed in rough order of preference:

1. **Config option on the storage class.** E.g.

   ```ts
   new ObservabilityStorageClickhouseVNext({
     client,
     engine: "replicated", // or: { cluster: "{cluster}", zooPath: "..." }
   });
   ```

   Internally this swaps `MergeTree` → `ReplicatedMergeTree(zooPath, replica)` and `ReplacingMergeTree(ver)` → `ReplicatedReplacingMergeTree(zooPath, replica, ver)` for every table it owns, and optionally appends `ON CLUSTER '{cluster}'` to DDL. Using the `{shard}` / `{replica}` macros keeps the zoo path generic.

2. **Honor server-side `default_table_engine`.** ClickHouse lets you set `default_table_engine = 'ReplicatedMergeTree'` at the cluster level, which auto-rewrites `CREATE TABLE ... ENGINE = MergeTree` on the fly. Mastra currently passes explicit engine strings which bypass this. If the storage class had an option to emit "unspecified" engines (or used `CREATE TABLE ... AS ...` without `ENGINE =`), operators could handle replication entirely server-side with no library changes.

3. **Auto-detect.** Query `system.clusters` / `system.replicas` on init and pick `Replicated*` automatically when the connected instance is part of a replicated cluster. More magic, but would "just work".

Whichever path is chosen, please apply it uniformly — including to any tables added in future versions (the recent `mastra_discovery_pairs` addition is what re-broke us). A single internal helper that wraps every `CREATE TABLE` would keep this consistent going forward.

Docs should also call out explicitly that the current defaults are **not safe** on multi-replica clusters — right now this is a silent footgun.

### Component

Memory

### Alternatives Considered

Two workarounds we've tried, neither of which is a real fix:

1. **Pin client traffic to a single replica.** Pass a custom `http.Agent` with `keepAlive: true, maxSockets: 1, keepAliveMsecs: 60_000` to `@clickhouse/client`'s `createClient` via `http_agent`. This serialises all Mastra's ClickHouse traffic onto one TCP socket, so as long as the socket stays alive it lands on one replica and reads/writes are consistent. Any socket reset (ingress idle timeout, pod restart) re-rolls the replica and the split-brain returns. It also obviously kills the point of having replicas.

2. **Pre-create the tables with `Replicated*` engines before Mastra starts.** Works, but is brittle: requires mirroring Mastra's internal schema (column types, ordering keys, TTLs) and breaks on every Mastra release that adds or changes a table — which is how we hit `mastra_discovery_pairs` in v-next.

Neither workaround addresses the new tables Mastra adds in future versions, which is the core of the feature request: we need the library itself to emit replication-safe DDL (or at least provide a supported hook for operators to control engine choice).

### Example Use Case

As a platform engineer running Mastra against a production multi-replica ClickHouse cluster, I want `@mastra/clickhouse` to emit replication-safe DDL (either automatically or via an explicit config option) so that traces, spans, scores, logs and metrics are consistent across all replicas and Mastra Studio shows stable, correct data regardless of which replica the HTTP load balancer picks for a given request.

### Additional Context

**Environment / versions affected**

- `@mastra/clickhouse` 1.5.0 (`ObservabilityStorageClickhouseVNext`) — confirmed reproducer
- `@mastra/core` 1.26.0, `@mastra/observability` 1.10.0
- `@clickhouse/client` 1.18.2
- ClickHouse 24.x, Altinity Operator, 2 replicas, no sharding, HTTP load balancer in front
- Node 22 / TypeScript

**Tables we observed the split-brain on**

- `mastra_trace_events`
- `mastra_span_events`
- `mastra_log_events`
- `mastra_metric_events`
- `mastra_score_events`
- `mastra_feedback_events`
- `mastra_discovery_pairs` (new in v-next; called out explicitly by our cluster operator as being created with `ENGINE = MergeTree` when it should be `ReplicatedMergeTree`)

### Verification

intojhanurag:
## Description

Adds an opt-in replicated table engine mode to `@mastra/clickhouse` so stores can initialize tables on replicated ClickHouse clusters. When configured, DDL emits `ReplicatedMergeTree` / `ReplicatedReplacingMergeTree` engines and can optionally apply `ON CLUSTER` to `CREATE TABLE`, `CREATE MATERIALIZED VIEW`, and `ALTER TABLE`
statements.

## Related Issue(s)

Fixes #15618

## Type of Change

- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [ ] Test update

## Checklist

## ELI5

This PR adds an opt-in setting so Mastra's ClickHouse integration can create replicated tables (copies coordinated across multiple ClickHouse servers) and optionally add ON CLUSTER to DDL — making multi-replica deployments safe from split‑brain.

## Overview

Adds opt-in replicated table engine support to @mastra/clickhouse so stores can emit ReplicatedMergeTree / ReplicatedReplacingMergeTree engines (and optional ON CLUSTER clauses) instead of plain MergeTree variants. Implemented via a single engine/DDL helper threaded through all ClickHouse-managed DDL paths; includes tests and docs.

## Key Changes

- New engine/DDL module: stores/clickhouse/src/storage/db/engine.ts
  - Exported type ClickhouseTableEngineConfig (shorthand 'replicated' and object form with optional cluster, zooPath, replica).
  - buildClickhouseTableEngine(baseEngine, tableName, config): converts MergeTree / ReplacingMergeTree → ReplicatedMergeTree / ReplicatedReplacingMergeTree, preserves engine args, supports {table} templating and replica defaults.
  - applyClickhouseDDLConfig(ddl, config): rewrites CREATE TABLE, CREATE MATERIALIZED VIEW, and ALTER TABLE (ADD COLUMN) to replace engine clauses and inject ON CLUSTER when configured.
  - Helpers: isReplicatedEngineConfig, isReplicatedTableEngineName.

- Config & propagation
  - Adds ClickhouseConfig.engine?: ClickhouseTableEngineConfig and re-exports the type.
  - resolveClickhouseConfig returns engine and ClickhouseDB stores it; domain constructors (background-tasks, memory, observability, v-next observability, scores, workflows) forward engine into ClickhouseDB so all DDL goes through applyClickhouseDDLConfig/getTableEngine.
  - TABLE_ENGINES usage replaced by getTableEngine that respects configured engine.

- Migration safety
  - migrateSignalTables and spans sorting-key migration paths observe engine config and abort with a MastraError (CLICKHOUSE/MIGRATE_SIGNAL_TABLES/REPLICATED_ENGINE_UNSUPPORTED) if a replicated engine is configured or detected to avoid unsafe automatic destructive changes.
  - Spans sorting-key migration is blocked when replicated engines are configured or detected, requiring operator review.

- DDL execution changes
  - CREATE TABLE and ALTER TABLE ... ADD COLUMN execution paths apply applyClickhouseDDLConfig so created/altered tables reflect configured replicated engines and include ON CLUSTER when provided.

- Tests & docs
  - Tests: stores/clickhouse/src/storage/db/engine.test.ts covers engine detection, conversion, zooPath/replica templating, and DDL rewriting (CREATE TABLE, MATERIALIZED VIEW, ALTER TABLE).
  - Documentation/changeset: updates in stores/clickhouse/README.md, docs composite example, and a changeset describing the feature and usage (examples show engine: 'replicated' or object with cluster/zooPath/replica).

## Notable behavioral details

- Opt-in and backwards compatible: default behavior unchanged; operators must enable replicated engines.
- zooPath supports {table} substitution and defaults to /clickhouse/tables/{shard}/{database}/{table}; replica defaults to '{replica}'.
- When cluster is provided, ON CLUSTER '<cluster>' is injected into CREATE TABLE, CREATE MATERIALIZED VIEW IF NOT EXISTS, and ALTER TABLE statements.
- Original MergeTree/ReplacingMergeTree engine arguments (e.g., ReplacingMergeTree(updatedAt)) are preserved inside the replicated wrapper.
- DDL rewriting is conservative: only affects recognized MergeTree / ReplacingMergeTree engine patterns; other engines/DDL are left unchanged.

## Types / API additions

- New exported type: ClickhouseTableEngineConfig
- New exported functions: isReplicatedEngineConfig, isReplicatedTableEngineName, buildClickhouseTableEngine, applyClickhouseDDLConfig
- ClickhouseConfig.engine?: ClickhouseTableEngineConfig (new optional property)

## Tests & Checklist

- Tests added for engine logic and DDL rewriting; docs and a changeset included.
- PR metadata notes Coderabbit comments are not fully addressed.

### DIFF
--- a/.changeset/some-cloths-beg.md
+++ b/.changeset/some-cloths-beg.md
@@ -13,7 +13,7 @@ const storage = new ClickhouseStore({
   engine: {
     type: 'replicated',
     cluster: 'production_cluster',
-    zooPath: '/clickhouse/tables/{shard}/{table}',
+    zooPath: '/clickhouse/tables/{shard}/{database}/{table}',
     replica: '{replica}',
   },
 });
--- a/.changeset/some-cloths-beg.md
+++ b/.changeset/some-cloths-beg.md
@@ -1,0 +1,20 @@
+---
+'@mastra/clickhouse': minor
+---
+
+Added an opt-in replicated table engine mode for ClickHouse stores. Configure replicated engines to create `ReplicatedMergeTree` and `ReplicatedReplacingMergeTree` tables, with optional `ON CLUSTER` DDL for replicated ClickHouse clusters.
+
+```ts
+const storage = new ClickhouseStore({
+  id: 'clickhouse',
+  url: process.env.CLICKHOUSE_URL!,
+  username: process.env.CLICKHOUSE_USERNAME!,
+  password: process.env.CLICKHOUSE_PASSWORD!,
+  engine: {
+    type: 'replicated',
+    cluster: 'production_cluster',
+    zooPath: '/clickhouse/tables/{shard}/{table}',
+    replica: '{replica}',
+  },
+});
+```
--- a/docs/src/content/en/reference/storage/composite.mdx
+++ b/docs/src/content/en/reference/storage/composite.mdx
@@ -298,7 +298,7 @@ const storage = new MastraCompositeStore({
       engine: {
         type: 'replicated',
         cluster: 'production_cluster',
-        zooPath: '/clickhouse/tables/{shard}/{table}',
+        zooPath: '/clickhouse/tables/{shard}/{database}/{table}',
         replica: '{replica}',
       },
     }),
--- a/docs/src/content/en/reference/storage/composite.mdx
+++ b/docs/src/content/en/reference/storage/composite.mdx
@@ -277,3 +277,31 @@ const storage = new MastraCompositeStore({
 This approach is also required when using storage providers that don't support observability (like Convex, DynamoDB, or Cloudflare). See the [DefaultExporter documentation](/docs/observability/tracing/exporters/default#storage-provider-support) for the full list of supported providers.
 
 :::
+
+For multi-replica ClickHouse clusters behind a load balancer, configure replicated table engines when Mastra initializes the ClickHouse tables. Without replicated engines, ClickHouse stores each table locally on the replica that receives the write, so Studio can show different observability data depending on which replica serves a read.
+
+```typescript
+import { MastraCompositeStore } from '@mastra/core/storage'
+import { MemoryPG, WorkflowsPG, ScoresPG } from '@mastra/pg'
+import { ObservabilityStorageClickhouse } from '@mastra/clickhouse'
+
+const storage = new MastraCompositeStore({
+  id: 'composite',
+  domains: {
+    memory: new MemoryPG({ connectionString: process.env.DATABASE_URL }),
+    workflows: new WorkflowsPG({ connectionString: process.env.DATABASE_URL }),
+    scores: new ScoresPG({ connectionString: process.env.DATABASE_URL }),
+    observability: new ObservabilityStorageClickhouse({
+      url: process.env.CLICKHOUSE_URL,
+      username: process.env.CLICKHOUSE_USERNAME,
+      password: process.env.CLICKHOUSE_PASSWORD,
+      engine: {
+        type: 'replicated',
+        cluster: 'production_cluster',
+        zooPath: '/clickhouse/tables/{shard}/{table}',
+        replica: '{replica}',
+      },
+    }),
+  },
+})
+```
--- a/docs/src/content/en/reference/storage/composite.mdx
+++ b/docs/src/content/en/reference/storage/composite.mdx
@@ -278,7 +278,7 @@ This approach is also required when using storage providers that don't support o
 
 :::
 
-For multi-replica ClickHouse clusters behind a load balancer, configure replicated table engines when Mastra initializes the ClickHouse tables. Without replicated engines, ClickHouse stores each table locally on the replica that receives the write, so Studio can show different observability data depending on which replica serves a read.
+For multi-replica ClickHouse clusters behind a load balancer, configure replicated table engines when Mastra initializes the ClickHouse tables. Without replicated engines, ClickHouse stores each table locally on the replica that receives the write, so Studio can show different observability data depending on which replica serves a read. For self-managed clusters, set `cluster` so DDL includes `ON CLUSTER` and creates the tables on every replica.
 
 ```typescript
 import { MastraCompositeStore } from '@mastra/core/storage'
--- a/docs/src/content/en/reference/storage/composite.mdx
+++ b/docs/src/content/en/reference/storage/composite.mdx
@@ -298,8 +298,6 @@ const storage = new MastraCompositeStore({
       engine: {
         type: 'replicated',
         cluster: 'production_cluster',
-        zooPath: '/clickhouse/tables/{shard}/{database}/{table}',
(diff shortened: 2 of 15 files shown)


intojhanurag:
Not planned

andreev-artem:
@intojhanurag hi, what is correct way for now to use mastra on clickhouse cluster?

epinzur:
@andreev-artem pls see the documentation here: https://mastra.ai/reference/storage/clickhouse#observability-with-vnext-recommended

andreev-artem:
@epinzur sorry, didn't fully get - couldn't find something about setup with Clickhouse Cluster. For Clickhouse that looks like pretty common case - to have cluster with replication. Currently I vibecoded workaroung with redifining `client.command` to change instructions on the fly. But that looks pretty hacky,

PR #15748, opened 2026-04-24
fix(mastracode): clear task list on new/switched thread

TylerBarnes:
## Bug

The task list state lives on `Harness.state.tasks` (and `Harness.state.activePlan`), which is a single global object held by the harness process — it is **not** persisted per-thread. `resetThreadDisplayState()` in the core harness correctly clears the per-thread *display* state on thread switch/create, but the schema-level harness state is never reset.

As a result, when you switch threads (or create a new one), the previous conversation's tasks and active plan remain on `harness.getState()` and get re-injected into the next thread's system prompt, causing stale tasks/plans to "leak" across threads.

## Fix

Clear the ephemeral per-thread fields on the mastracode side (keeping the fix out of the core harness public API):

- `mastracode/src/tui/event-dispatch.ts` — call `state.harness.setState({ tasks: [], activePlan: null })` inside both the `thread_changed` and `thread_created` handlers.
- `mastracode/src/tui/commands/new.ts` — do the same inside `/new`, so the next system prompt (which is built before the subsequent `thread_created` event fires) does not carry over stale tasks/plan. The command is now `async` and awaited in `command-dispatch.ts`.
- `mastracode/src/tui/setup.ts` — `renderExistingTasks` now reads from `harness.getDisplayState().tasks` (already per-thread) instead of `getState().tasks`, so the UI matches the cleared harness state.
- `mastracode/src/schema.ts` — the `tasks` comment claimed "persisted per-thread", which was misleading. Updated to accurately describe that it is ephemeral, lives on the single global state object, and must be cleared by callers on thread transitions.

`omProgress` / `om status` were already handled via the display state reset and are not touched.

## Tests

- `mastracode/src/tui/__tests__/event-dispatch-thread-clear.test.ts` — dispatches `thread_changed` and `thread_created` events directly and asserts `harness.setState` is called with `{ tasks: [], activePlan: null }` and that the underlying state is cleared.
- `mastracode/src/tui/commands/__tests__/new.test.ts` — calls `handleNewCommand` and verifies tasks/activePlan are cleared plus the other `/new` side effects.

Both new test files pass locally:

```
✓ src/tui/__tests__/event-dispatch-thread-clear.test.ts (2 tests)
✓ src/tui/commands/__tests__/new.test.ts (1 test)
✓ src/tui/__tests__/command-dispatch.test.ts (7 tests)
✓ src/tui/commands/__tests__/thread.test.ts (4 tests)
✓ src/tui/commands/__tests__/threads.test.ts (3 tests)
```

`pnpm --filter mastracode lint` is clean. `pnpm --filter mastracode check` shows only pre-existing type errors unrelated to this change (they reproduce on `main`). The rest of the `mastracode` test suite has environmental flakiness tied to `process.stdout.columns` / `isolate: false` that is also present on `main`.

---

<a href="https://app.devin.ai/review/mastra-ai/mastra/pull/15748" target="_blank">
  <picture>
    <source media="(prefers-color-scheme: dark)" srcset="https://static.devin.ai/assets/gh-open-in-devin-review-dark.svg?v=1">
    Open in Devin Review
  </picture>
</a>

## ELI5

Switching or creating a new conversation thread sometimes showed leftover tasks and plans from a previous thread. This PR makes sure those task lists and the active plan are cleared whenever a thread is created or changed so every thread starts fresh.

## Overview

Fixes a bug where transient task list and activePlan state leaked across threads in the mastracode TUI because those fields lived on the single global harness state rather than being scoped/cleared per-thread. The PR ensures harness-level ephemeral fields are explicitly reset on thread creation/switch and during the `/new` command so stale tasks/activePlan are not injected into subsequent threads' system prompts.

## Changes Made

- mastracode/src/tui/event-dispatch.ts
  - In both `thread_changed` and `thread_created` handlers call `state.harness.setState({ tasks: [], activePlan: null, sandboxAllowedPaths: [] })` to clear harness-level ephemeral state.
  - Retain and align UI refresh logic (taskProgress update, render requests) with the cleared state.

- mastracode/src/tui/commands/new.ts
  - Convert `handleNewCommand` to async and await the harness state reset, explicitly clearing `tasks`, `activePlan`, and `sandboxAllowedPaths` during new-thread initialization so the next system prompt cannot include stale data.

- mastracode/src/tui/command-dispatch.ts
  - Await `handleNewCommand(...)` in the `/new` slash-command path to ensure async clearing completes before returning.

- mastracode/src/tui/setup.ts
  - `renderExistingTasks` now reads tasks from the harness display state (per-thread) via `harness.getDisplayState().tasks` instead of global `getState().tasks` to keep UI consistent with cleared per-thread state.

- mastracode/src/schema.ts
  - Updated comment clarifying that `tasks` are ephemeral and stored on the global harness state and must be cleared on thread transitions.

- mastracode/src/tui/commands/clone.ts
  - Clarifying comment about when ephemeral per-thread harness state must be cleared in clone flows.

- .changeset/clear-thread-task-list.md
  - Adds a changeset note describing the clearing behavior on thread start/switch/clone.

## Tests Added

- mastracode/src/tui/__tests__/event-dispatch-thread-clear.test.ts
  - Asserts `harness.setState` is called to clear `{ tasks: [], activePlan: null, sandboxAllowedPaths: [] }` on `thread_changed` and `thread_created`, verifies in-memory state and UI/task-progress side effects, and checks created-thread title handling.

- mastracode/src/tui/commands/__tests__/new.test.ts
  - Verifies `/new` clears `tasks`/`activePlan`/`sandboxAllowedPaths`, and checks related command-side state updates (pending new thread flag, taskWriteInsertIndex reset).

## Impact

- Prevents task list and active plan from leaking between threads; each thread starts with a clean ephemeral state.
- Keeps existing per-thread display-state reset mechanisms intact while ensuring the underlying harness-level ephemeral fields are also reset.
- No exported/public APIs changed except `handleNewCommand` now returns a Promise (signature changed from `void` → `Promise<void>`).

### DIFF
--- a/.changeset/clear-thread-task-list.md
+++ b/.changeset/clear-thread-task-list.md
@@ -1,0 +1,5 @@
+---
+'mastracode': patch
+---
+
+Fix(mastracode): clear task list, active plan, and sandbox allowed paths when starting, switching, or cloning a thread so ephemeral per-thread harness state from a previous conversation no longer leaks across threads.
--- a/mastracode/src/schema.ts
+++ b/mastracode/src/schema.ts
@@ -32,7 +32,11 @@ export const stateSchema = z.object({
   smartEditing: z.boolean().default(true),
   // Notification mode — alert when TUI needs user attention
   notifications: z.enum(['bell', 'system', 'both', 'off']).default('off'),
-  // Task list (persisted per-thread)
+  // Task list — ephemeral and scoped to the active thread.
+  // NOTE: this lives on the harness's single global state (one per process)
+  // and is NOT persisted with the thread; callers (see mastracode TUI
+  // thread_changed / thread_created / /new handlers) must clear it when
+  // switching threads so tasks do not leak across conversations.
   tasks: z
     .array(
       z.object({
--- a/mastracode/src/tui/__tests__/event-dispatch-thread-clear.test.ts
+++ b/mastracode/src/tui/__tests__/event-dispatch-thread-clear.test.ts
@@ -9,11 +9,17 @@ vi.mock('../../utils/project.js', () => ({
 }));
 
 function createHarness() {
-  const state: { tasks: unknown[]; activePlan: unknown | null; escapeAsCancel?: boolean } = {
+  const state: {
+    tasks: unknown[];
+    activePlan: unknown | null;
+    sandboxAllowedPaths: string[];
+    escapeAsCancel?: boolean;
+  } = {
     tasks: [
       { content: 'leftover task', status: 'pending', activeForm: 'leftover task' },
     ],
     activePlan: { title: 'stale plan', plan: 'stale', approvedAt: new Date().toISOString() },
+    sandboxAllowedPaths: ['/tmp/leftover-allowed-path'],
   };
 
   return {
@@ -64,7 +70,7 @@ describe('event-dispatch thread state clearing', () => {
     state = createState(harness);
   });
 
-  it('clears tasks and activePlan on thread_changed', async () => {
+  it('clears tasks, activePlan, and sandboxAllowedPaths on thread_changed', async () => {
     const event: HarnessEvent = {
       type: 'thread_changed',
       threadId: 'thread-1',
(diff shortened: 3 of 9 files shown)


abhiaiyer91 (approved):

TylerBarnes (commented):
Review from Mastra Code:

Thanks for the fix. This is the cleaner implementation overall: awaiting the state reset in `/new`, awaiting it from command dispatch, and clearing state before rendering the switched thread all look right. The schema comment update is also helpful because it documents why this state has to be cleared manually.

A few follow-ups I’d recommend before landing:

- If `sandboxAllowedPaths` is intended to be per-thread, it should be cleared alongside `tasks` and `activePlan`. The schema describes it as per-thread, so leaving it out may preserve a related state leak.
- Consider also clearing harness state in the clone reset path, or explicitly confirm clone always flows through `thread_created` / `thread_changed` after reset.
- Please add a changeset for the user-visible bug fix.

Overall: this is the implementation I’d use as the base, with the related per-thread state coverage filled in.

PR #15690, opened 2026-04-23
feat: ownership, visibility & access control for stored agents and skills

CalebBarnes:
## Summary

Adds ownership (`authorId`) and visibility (`private` / `public`) to stored agents and stored skills, with full access-control enforcement on server routes, storage backends, client SDK, and Agent Builder UI.

Closes COR-831, COR-835, COR-837, COR-826, COR-836, COR-887, COR-886, COR-884

### Server — user-aware request context (COR-831)

- Server adapters (Hono, Express, Fastify, Koa) propagate authenticated user identity into request context
- Reserved context keys (`user`, `userPermissions`, `userRoles`) namespaced with `mastra__` prefix to prevent collisions with app-level `requestContext`
- New `authorship.ts` module with helpers: `getCallerAuthorId`, `assertReadAccess`, `assertExecuteAccess`, `assertWriteAccess`, `resolveAuthorFilter`, `matchesAuthorFilter`

### Core + Stores — ownership & visibility on storage (COR-835, COR-826)

- `authorId` and `visibility` fields added to storage types for both agents and skills
- Schema constants (`AGENTS_SCHEMA`, `SKILLS_SCHEMA`) updated with new columns
- All 5 storage backends (inmemory, filesystem, libsql, pg, mongodb) persist and filter on `authorId` and `visibility`
- `StorageVisibility` type (`'private' | 'public'`) shared between agents and skills

### Server — route enforcement (COR-835, COR-837, COR-826, COR-836)

- **Stored agents:** All `/agents/:agentId/*` routes enforce ownership + visibility via `assertReadAccess`, `assertExecuteAccess`, or `assertWriteAccess`
- **Stored skills:** All `/stored/skills/*` routes enforce the same ownership + visibility checks
- **Unowned resources:** Code-defined agents and stored resources without an `authorId` are treated as public and accessible to all (COR-837)
- **List filtering (COR-836):** `GET /stored/agents` and `GET /stored/skills` support `?authorId=X` and `?visibility=public` query params. Default listing returns caller's own agents + public agents from others. Admins see everything.
- **Create defaults:** `authorId` injected from authenticated caller; `visibility` defaults to `private` when owned, `public` when unowned (no auth)
- Broad role grants (e.g. `agents:read`) no longer bypass per-record ownership — scoped grants (`agents:read:<id>`) are required to override

### Client SDK

- `visibility` and `authorId` added to response types, create/update params, and list filter params for both agents and skills
[...]

PR #15750, opened 2026-04-24
Add publish permission action and split stored resources into per-family permissions

YujohnNattrass:
## Summary

Studio admins can now delegate version activation (publish, activate, restore) without handing out full write access, and can scope CRUD to individual stored resource families instead of the catch-all 'stored' key.

## What changed

- New 'publish' permission action, derived automatically for POST routes ending in `/publish`, `/activate`, or `/restore`.
- `stored` resource split into per-family resources: `stored-agents`, `stored-skills`, `stored-prompt-blocks`, `stored-mcp-clients`, `stored-scorers`, `stored-workspaces`. Each supports the standard CRUD actions; all except `stored-workspaces` also support `publish`.
- Roles that previously granted `stored:read` / `stored:write` / `stored:delete` / `stored:*` continue to work via a backwards-compatible alias in `matchesPermission`. The legacy patterns are kept in `PERMISSION_PATTERNS` and marked `@deprecated` so IDEs surface the migration path.
- Playground route registry updated so the new per-family keys are used when gating UI routes.

## Test plan

- `pnpm --filter @mastra/server test permissions.test.ts` — 167 tests pass (includes new coverage for all 6 stored families, the `publish` action, and `/publish`, `/activate`, `/restore` derivation).
- `pnpm --filter @mastra/core test roles.test.ts` — 34 new tests pass covering the legacy alias across every family, resource-id preservation, and negative cases.
- `tsc --noEmit` clean for `@mastra/core` and `@mastra/playground`.

Base: #15690 (`agent-builder-stuff`).


YujohnNattrass:
closing in favor of https://github.com/mastra-ai/mastra/pull/15878

PR #15768, opened 2026-04-26
core: block phantom nested-subpath imports with explicit null exports

### LINKED ISSUE #15758: Phantom exports: wildcard ./* in @mastra/core exposes 61 non-existent runtime subpaths
## Summary

`@mastra/core` uses a wildcard export `"./*"` in `package.json` that, combined with tsc emitting `.d.ts` files for every source module while tsup bundles JS into single entry files, creates **61 importable subpaths that compile in TypeScript but fail at runtime**.

Two customer orgs have already hit this with `@mastra/core/auth/ee/defaults`.

## Root Cause

The exports field wildcard `"./*"` maps to `./dist/*/index.{js,d.ts}`. Per Node.js spec, `*` is a **string replacement** that matches **any characters including `/`**. So:

```
import { StaticRBACProvider } from "@mastra/core/auth/ee/defaults"
```

Matches the wildcard: `*` captures `auth/ee/defaults`, resolving to:
- **Types**: `./dist/auth/ee/defaults/index.d.ts` → **EXISTS** (tsc emits individual `.d.ts` per source file)
- **Runtime**: `./dist/auth/ee/defaults/index.js` → **DOES NOT EXIST** (tsup bundles into `dist/auth/ee/index.js`)

TypeScript compiles fine, deployment crashes with `MODULE_NOT_FOUND`.

## Scope

61 phantom subpaths exist, including:
- `auth/ee/defaults`, `auth/ee/interfaces`, `auth/defaults`, `auth/interfaces`
- `llm/model`, `llm/model/gateways`
- `workspace/sandbox`, `workspace/filesystem`, `workspace/tools`, `workspace/skills`
- `storage/domains`, `storage/domains/*` (12+ sub-domains)
- `agent/message-list/adapters`, `agent/message-list/cache`, etc.
- `evals/run`, `processors/memory`, `stream/aisdk/v5/compat`, and more

Any of these can be imported in TypeScript without error, but will crash at runtime.

## Why deleting orphaned `.d.ts` files doesn't work

The orphaned `.d.ts` files are **referenced by parent `.d.ts` files** via relative imports. For example, `dist/auth/ee/index.d.ts` contains `export * from './defaults'` — deleting `dist/auth/ee/defaults/index.d.ts` would break type resolution for `@mastra/core/auth/ee` itself.

30+ valid parent `.d.ts` files have relative imports to orphaned subdirectories.

## Recommended Fix

Use **`null` exports** to block phantom subpaths. Per [Node.js docs](https://nodejs.org/api/packages.html):

```json
{
  "exports": {
    "./*": { "import": { "types": "./dist/*/index.d.ts", "default": "./dist/*/index.js" } },
    "./auth/ee/defaults": null,
    "./auth/ee/interfaces": null
  }
}
```

This causes:
- **Node.js**: `ERR_PACKAGE_PATH_NOT_EXPORTED` (clear error at import time)
- **TypeScript**: `TS2307: Cannot find module` (fails at compile time, not deploy time)

Parent modules like `@mastra/core/auth/ee` continue working — their `export * from './defaults'` is a relative import within the package, unaffected by exports.

### Implementation options

1. **Static null entries**: Add all 61 null entries to `package.json` manually. Requires maintenance when source modules change.

2. **Auto-generate null entries at build time**: In `generateTypes()`, after tsc completes, scan for `index.d.ts` without matching `index.js`, compute subpaths, and write null entries into `package.json` exports. Automatic but modifies `package.json` during build.

3. **CI validation**: Add a build check that fails if any `.d.ts`-only subpath exists without a corresponding null export entry.

Option 2 (auto-generate) + option 3 (CI check) would be the most robust.

## Reproduction

```ts
// This compiles fine in TypeScript but crashes at runtime
import { StaticRBACProvider, DEFAULT_ROLES } from "@mastra/core/auth/ee/defaults";
```

Error at deploy time:
```
Cannot find module '/app/.mastra/output/node_modules/@mastra/core/dist/auth/ee/defaults/index.js'
```

## Verified behavior of null exports

Tested with Node.js and TypeScript 5.9:
- `"./auth/ee/defaults": null` blocks both runtime and compile-time resolution
- Parent `@mastra/core/auth/ee` continues to export all types from `./defaults` via relative import
- `DEFAULT_ROLE` is accessible through `@mastra/core/auth/ee` (the correct import path)

joaquinhuigomez:
The wildcard `./*` export in `@mastra/core` matches multi-segment paths (per Node.js spec, `*` captures `/`), so nested type-only files emitted by tsc resolve to importable subpaths whose `.js` bundle never exists in `dist`. TypeScript compiles them; the program crashes at runtime with `MODULE_NOT_FOUND`.

Add `null` entries for the subpaths called out in the issue as breaking customers (`@mastra/core/auth/ee/defaults` and friends), so resolution fails at import time with `ERR_PACKAGE_PATH_NOT_EXPORTED` / `TS2307` instead of silently mapping a phantom `.d.ts`. Build-time auto-generation of all phantom entries (Option 2 in the issue) can land in a follow-up.

Fixes #15758

## ELI5

This PR prevents broken imports from appearing to work in TypeScript. The @mastra/core package had some "phantom" subpaths where TypeScript definition files existed but the actual JavaScript code didn't, causing code to compile successfully but crash when run. The fix adds explicit "this path is blocked" entries in package.json for these phantom paths so TypeScript and Node.js both reject them at import time instead of failing silently at runtime.

## Changes

**packages/core/package.json**

Adds 16 explicit null export entries to the `exports` map for phantom subpaths that have TypeScript definitions but no corresponding runtime JavaScript bundles:

- `./auth/defaults`
- `./auth/interfaces`
- `./auth/ee/defaults`
- `./auth/ee/interfaces`
- `./agent/message-list/adapters`
- `./agent/message-list/cache`
- `./evals/run`
- `./llm/model`
- `./llm/model/gateways`
- `./processors/memory`
- `./processors/processors`
- `./stream/aisdk/v5/compat`
- `./storage/domains`
- `./workspace/filesystem`
- `./workspace/sandbox`
- `./workspace/skills`
- `./workspace/tools`

This targeted fix ensures these known phantom subpaths are explicitly marked as unexported in the package, causing Node.js to raise `ERR_PACKAGE_PATH_NOT_EXPORTED` errors and TypeScript to report `TS2307` compile errors when these paths are imported, rather than allowing them to resolve through the wildcard export and fail mysteriously at runtime.

## Impact

Prevents module resolution failures at deployment time by making import errors visible at compile time. This is a targeted mitigation for the 61 phantom subpaths identified across @mastra/core; build-time auto-generation of null entries for all such paths is deferred to a follow-up effort.

### DIFF
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -31,6 +31,23 @@
         "default": "./dist/*/index.cjs"
       }
     },
+    "./auth/defaults": null,
+    "./auth/interfaces": null,
+    "./auth/ee/defaults": null,
+    "./auth/ee/interfaces": null,
+    "./agent/message-list/adapters": null,
+    "./agent/message-list/cache": null,
+    "./evals/run": null,
+    "./llm/model": null,
+    "./llm/model/gateways": null,
+    "./processors/memory": null,
+    "./processors/processors": null,
+    "./stream/aisdk/v5/compat": null,
+    "./storage/domains": null,
+    "./workspace/filesystem": null,
+    "./workspace/sandbox": null,
+    "./workspace/skills": null,
+    "./workspace/tools": null,
     "./auth/ee": {
       "import": {
         "types": "./dist/auth/ee/index.d.ts",


LekoArts:
Thanks for the PR! But we'd like to solve this on our own 👍

PR #15769, opened 2026-04-26
fix(core): reject plain functions as individual tool entries in AgentConfig.tools

### LINKED ISSUE #15229: [BUG] `AgentConfig.tools` typing is unnecessarily wide
### Describe the Bug

AgentConfig.tools accepts an invalid value that should fail type checking. In a static tools map,
  each entry should already be a tool object, but TypeScript currently allows a nested resolver
  function like myTool: () => realTool.

  That compiles without error, even though it is not a valid static tool definition. The bad value
  then survives to runtime, where agent.listTools() crashes instead of the mistake being caught by
  TypeScript.

### Steps To Reproduce

1. Clone the reproduction repo: https://github.com/abrahamcl/mastra-bug-agent-tools

## Reproduction

```sh
npm install
npx tsc --noEmit
```

Expected: TypeScript reports an error for `myTool: () => realTool`.

Actual: `tsc` exits successfully with no type error.

To show that an actual runtime crash occurs when using the invalid config, run:

```sh
npm run test
```

https://github.com/abrahamcl/mastra-bug-agent-tools

### Expected Behavior

Typescript should flag an invalid tool value

### Environment Information

```block
System:
    OS: macOS 26.3
    CPU: (8) arm64 Apple M3
    Memory: 128.28 MB / 24.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 25.8.1 - /opt/homebrew/bin/node
    Yarn: 1.22.22 - /opt/homebrew/bin/yarn
    npm: 11.11.0 - /opt/homebrew/bin/npm
    pnpm: 10.32.1 - /opt/homebrew/bin/pnpm
    bun: 1.3.10 - /opt/homebrew/bin/bun
  Browsers:
    Chrome: 146.0.7680.165
    Safari: 26.3
  npmPackages:
    @mastra/core: ^1.24.1 => 1.24.1 
    @types/node: ^25.5.2 => 25.5.2 
    mastra: ^1.5.0 => 1.5.0 
    typescript: ^6.0.2 => 6.0.2 
    vitest: ^4.1.4 => 4.1.4 
    zod: ^4.3.6 => 4.3.6 
  npmGlobalPackages:
    @openai/codex: 0.75.0
    codex: 0.2.3
    httpyac: 6.16.7
    mint: 4.2.60
    npm: 11.11.0
```

### Verification

gaurav0107:
## Description

`AgentConfig.tools` was silently accepting function values as individual tool entries, which then crashed at runtime inside `agent.listTools()`. The union member `ProviderDefinedTool` had all fields optional plus an `[key: string]: any` index signature, so a plain function matched it structurally.

Before:

```ts
new Agent({
  tools: { myTool: () => realTool }, // no type error — crashes at runtime
});
```

After:

```ts
new Agent({
  tools: { myTool: () => realTool }, // TS error: missing `id`
});

new Agent({
  tools: { myTool: realTool }, // still fine
});

new Agent({
  tools: ({ requestContext }) => ({ myTool: realTool }), // still fine
});
```

The fix makes `id: string` required on the v5 branch of `ProviderDefinedTool`. Functions don't declare an `id` on their type, so TypeScript now rejects them. This mirrors the existing runtime guard in `isProviderDefinedTool`, which already requires `id: string`. Real provider tools (`google.tools.googleSearch()`, `openai.tools.webSearch()`, etc.) always carry an `id` of the form `'<provider>.<tool_name>'`, so this is backwards-compatible.

Added three type-level tests in `agent-types.test-d.ts` covering the rejection, the positive case, and the whole-map resolver pattern on `AgentConfig.tools`.

Note: this revisits the same fix direction as the self-closed #15261; the CodeRabbit changeset-wording feedback from that round is addressed up front.

## Related Issue(s)

Fixes #15229

## Type of Change

- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [ ] Test update

## Checklist

- [x] I have made corresponding changes to the documentation (if applicable) — changeset added
- [x] I have added tests that prove my fix is effective — 3 type-level tests in `agent-types.test-d.ts`

## ELI5

TypeScript used to let you put a plain function where a “tool object” belongs, which looked OK at compile time but could crash later. This PR tightens the `AgentConfig.tools` typing so individual entries must be real tool objects, while still allowing `tools` to be provided via a resolver function that returns the whole tools map.

---

### Type Definition Update
- **File:** `packages/core/src/agent/types.ts`
- Tightened `ToolsInput` by introducing an internal `ToolsInputValue` union that only allows provider-defined tools when they include an `id: string` locally:
  - `ToolAction | VercelTool | VercelToolV5 | (ProviderDefinedTool & { id: string })`
- Result: function-valued **individual** map entries like `tools: { myTool: () => realTool }` are rejected at compile time, matching the runtime guard behavior.

### External Types
- **File:** `packages/_external-types/src/index.ts`
- Kept `ProviderDefinedTool`’s public shape unchanged (`id?: string`), with a TODO noting `id` should become required in a future breaking change.

### Type-level Tests
- **File:** `packages/core/src/agent/agent-types.test-d.ts`
- Added compile-time checks for Issue `#15229`:
  - Rejects a function value as a static tool entry (`@ts-expect-error`)
  - Also rejects resolver functions returning non-tool values (e.g. `() => 42` and `async () => ({})`) via additional `@ts-expect-error`
  - Confirms valid static tool objects created with `createTool` are accepted
  - Confirms the dynamic “whole-map resolver” pattern remains valid: `tools` can be provided as a function returning a `ToolsInput` map (via the `DynamicArgument` pattern)

### Changeset
- **File:** `.changeset/fix-agent-tools-reject-function-values.md`
- Added a patch changeset documenting the TypeScript typing fix.

## Impact

- Prevents mis-typed configs like `myTool: () => realTool` from compiling and later crashing during `agent.listTools()`.
- Non-breaking for correct usage: valid provider tools (with `id`) and `tools` as a dynamic resolver function are still supported.

## Related

- Fixes: `#15229`
- Revisits: `#15261`

### DIFF
--- a/.changeset/fix-agent-tools-reject-function-values.md
+++ b/.changeset/fix-agent-tools-reject-function-values.md
@@ -1,0 +1,5 @@
+---
+'@mastra/core': patch
+---
+
+Fixed typing for `AgentConfig.tools` so each tool entry must be a tool object. Functions like `tools: { myTool: () => realTool }` are now rejected at compile time, while setting `tools` itself to a resolver function is still supported. Fixes [#15229](https://github.com/mastra-ai/mastra/issues/15229).
--- a/packages/_external-types/src/index.ts
+++ b/packages/_external-types/src/index.ts
@@ -31,10 +31,18 @@ export type ProviderDefinedTool =
     }
   | {
       // ToolV5 structure
+      // TODO(v2): make `id` non-optional in a breaking change. The runtime
+      // check in `isProviderDefinedTool` already requires `id: string`, and
+      // every real provider tool (e.g. `google.tools.googleSearch()`,
+      // `openai.tools.webSearch()`) carries an `id` of the form
+      // `'<provider>.<tool_name>'`. Tightening here would let TypeScript
+      // catch bare function values at compile time instead of relying on the
+      // narrower `ToolsInput` intersection, but it is a public-type change
+      // and deferred to the next major.
+      id?: string;
       inputSchema?: unknown;
       description?: string;
       type?: string;
-      id?: string;
       name?: string;
       providerOptions?: any;
       execute?: ((...args: any[]) => any) | undefined;
--- a/packages/core/src/agent/agent-types.test-d.ts
+++ b/packages/core/src/agent/agent-types.test-d.ts
@@ -3,9 +3,10 @@ import { assertType, describe, expectTypeOf, it } from 'vitest';
 import { z } from 'zod/v4';
 import type { RequestContext } from '../request-context';
 import type { PublicSchema } from '../schema';
+import { createTool } from '../tools';
 import { Agent } from './agent';
 import type { AgentExecutionOptions, PublicAgentExecutionOptions } from './agent.types';
-import type { AgentConfig } from './types';
+import type { AgentConfig, ToolsInput } from './types';
 
 /**
  * Type tests for Agent configuration types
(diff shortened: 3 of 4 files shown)


gaurav0107:
@coderabbitai review

TylerBarnes (changes requested):
From Mastra Code: The runtime behavior this PR is encoding makes sense: bare function entries in `tools` already throw through `ensureToolProperties()`, so rejecting `tools: { myTool: () => realTool }` at compile time is the right direction.

My concern is the fix changes the exported `ProviderDefinedTool` V5 branch from `id?: string` to `id: string`. That is a public type tightening, even though it matches what `isProviderDefinedTool()` requires at runtime.

Could you avoid changing the public type in this PR and leave `id` optional for now? Please add a TODO above it instead, e.g. `TODO(v2): make this non-optional in a breaking change`, and use a narrower approach for the `ToolsInput` / `AgentConfig.tools` rejection so bare function entries are still rejected without making `ProviderDefinedTool` stricter in a patch release.

gaurav0107:
@TylerBarnes thanks — fair point on the patch-version constraint. Just pushed `9d2d945`:

- Reverted the public `ProviderDefinedTool` V5 branch in `@internal/external-types` back to `id?: string` and added a `TODO(v2): make \`id\` non-optional in a breaking change` with the rationale (runtime `isProviderDefinedTool` already requires it; every real provider tool carries one).
- Moved the bare-function rejection into `ToolsInput` itself by narrowing *only* the provider branch locally: `(ProviderDefinedTool & { id: string })`. Plain functions have no `id`, so they still fail at compile time, but `ProviderDefinedTool` stays loose for external consumers.
- `AgentConfig.tools` as a `DynamicArgument<ToolsInput>` resolver continues to work; the three `@ts-expect-error` cases (`() => realTool`, `() => 42`, `async () => ({})`) still fire.

Changeset stays a patch.

gaurav0107:
@TylerBarnes thanks for the careful read — agreed on both points, and I reworked it so the exported type is untouched.

- `@internal/external-types`: the `ProviderDefinedTool` V5 branch stays `id?: string`. I added a `TODO(v2): make id non-optional in a breaking change` above it, so there is no public-type tightening in this patch release.
- The compile-time rejection now lives only in `ToolsInput` (`packages/core/src/agent/types.ts`): the provider-defined branch is narrowed locally to `(ProviderDefinedTool & { id: string })` via a new internal `ToolsInputValue` alias. That is scoped to the agent tools-input map and changes no exported type shape.

Behavior is unchanged for real usage: provider tools (e.g. `google.tools.googleSearch()`) always carry an `id`, so valid entries still typecheck, and setting `AgentConfig.tools` to a whole-map resolver (`DynamicArgument<ToolsInput>`) still works. Only individual `myTool: () => realTool` entries are rejected — mirroring what `isProviderDefinedTool()` / `ensureToolProperties()` already enforce at runtime.

The `agent-types.test-d.ts` type suite covers all cases (bad function entry, non-tool resolvers `() => 42` and `async () => ({})`, a valid tool object, and the dynamic resolver) and passes 9/9 with no type errors. Repushed as a single commit `c4902f58`.

abhiaiyer91:
Thanks for working on this! This has since been addressed by #21705, which is now merged, so I'm closing this PR as superseded.

PR #15792, opened 2026-04-27
 fix(core): close dataset.startExperiment workflow gaps

### LINKED ISSUE #15613: [BUG] `dataset.startExperiment` workflow target drops `stepResults` / `stepExecutionPath` and does not propagate observability context
## Describe the Bug

The `executeWorkflow` implementation used by `dataset.startExperiment` is a stripped-down version of the one used by `runEvals`. It returns `{ output: result.result, traceId }` only — it omits `stepResults` and `stepExecutionPath`, and it does not propagate `requestContext` or the observability context to `run.start(...)`.

Two concrete consequences:

1. **Scorers can't do per-step evaluation on a persisted experiment.** The step outputs are never surfaced to `runScorersForItem`, so there's no `stepResult.output` to feed to a step-scoped scorer.

2. **The persisted observability trace is nearly empty for workflow targets.** A workflow whose step internally calls `agent.stream(...)` produces the expected nested `agent_run` / `model_generation` / `tool_call` spans during execution (they appear in Langfuse and other OTEL-style exporters), but only the outermost `workflow_run` + `workflow_step` spans land in the Mastra observability store. Any scorer or diagnostic that walks `storage.getStore('observability').getTrace({ traceId })` sees a trace with just those two spans; `extractTrajectoryFromTraceStore(...)` therefore returns a `Trajectory` with a single empty `workflow_step` entry.

Same workflow, run under `runEvals` instead of `dataset.startExperiment`, produces complete nested spans in the observability store and exposes `stepResults` to scorers — so the surface where this matters is clearly the persisted-experiments runner.

## Steps To Reproduce

1. Set up a Mastra instance with storage + observability so traces persist to the observability store:

   ```ts
   import { createStep, createWorkflow } from '@mastra/core/workflows';
   import { createScorer } from '@mastra/core/evals';
   import { LibSQLStore } from '@mastra/libsql';
   import { Observability, DefaultExporter } from '@mastra/observability';
   import { z } from 'zod';

   const echoTool = createTool({
     id: 'echo',
     description: 'echoes its input',
     inputSchema: z.object({ value: z.string() }),
     outputSchema: z.object({ value: z.string() }),
     execute: async ({ context }) => ({ value: context.value }),
   });

   const agent = new Agent({
     name: 'echo-agent',
     instructions: 'Call the echo tool, then answer.',
     model: 'openai/gpt-5-mini',
     tools: { echoTool },
   });

   const chatStep = createStep({
     id: 'chat',
     inputSchema: z.object({ prompt: z.string() }),
     outputSchema: z.object({ text: z.string() }),
     execute: async ({ inputData, requestContext }) => {
       const res = await agent.stream(
         [{ role: 'user', content: inputData.prompt }],
         { requestContext, maxSteps: 5 },
       );
       return { text: await res.text };
     },
   });

   const workflow = createWorkflow({
     id: 'repro-workflow',
     inputSchema: z.object({ prompt: z.string() }),
     outputSchema: z.object({ text: z.string() }),
   }).then(chatStep).commit();

   // Probe scorer — prints what it actually receives.
   const probe = createScorer({
     id: 'probe',
     description: 'prints run payload',
     type: 'agent',
   })
     .preprocess(async ({ run }) => {
       console.log('run.output      =', run.output);
       console.log('run.input       =', run.input);
       console.log('run.targetTraceId =', run.targetTraceId);
       console.log('Any stepResults in run? →', Object.keys(run));
       // Try to read trace directly from storage to inspect spans.
       const obs = await mastra.getStorage()?.getStore('observability');
       const trace = obs && run.targetTraceId
         ? await obs.getTrace({ traceId: run.targetTraceId })
         : null;
       console.log(
         'persisted spans =',
         trace?.spans?.map((s) => ({ type: s.spanType, name: s.name })),
       );
       return { ok: true };
     })
     .generateScore(() => 1);

   const mastra = new Mastra({
     agents: { 'echo-agent': agent },
     workflows: { 'repro-workflow': workflow },
     scorers: { probe },
     storage: new LibSQLStore({ id: 's', url: ':memory:' }),
       configs: {
         default: { serviceName: 'repro', exporters: [new DefaultExporter()] },
       },
     }),
   });
   ```

2. Seed a dataset and start the experiment:

   ```ts
   const dataset = await mastra.datasets.create({ name: 'repro-ds' });
   await dataset.addItems({
     items: [{
       input: { prompt: 'say hello' },
       groundTruth: { text: 'hello' },
     }],
   });

   await dataset.startExperiment({
     name: 'repro',
     targetType: 'workflow',
     targetId: 'repro-workflow',
     scorers: [probe],
   });
   ```

3. Observe the probe's logs:
   - `run.output = { text: '...' }` — the workflow's declared output. No `stepResults` / `stepExecutionPath` reachable from the scorer.
   - `persisted spans = [ { type: 'workflow_run', ... }, { type: 'workflow_step', ... } ]` — only 2 spans. No nested `agent_run`, `model_generation`, or `tool_call`, even though the agent inside the step called `echoTool` and produced a model response during execution (visible in third-party exporters like Langfuse).

4. For contrast, invoke the same workflow through `runEvals({ target: workflow, scorers: { workflow: [probe], steps: { chat: [probe] }, trajectory: [probe] }, data: [...] })` — the step scorer receives `stepResult.output`, and the trace contains the full nested span hierarchy.

N/A — the snippet in **Steps To Reproduce** is self-contained. Paste it into a fresh Mastra project (with an OpenAI-compatible key for the agent model) to reproduce both observations. Happy to spin up a repo on request if preferred.

## Expected Behavior

`dataset.startExperiment` should have parity with `runEvals` for workflow targets:

- `executeWorkflow` should expose `stepResults` and `stepExecutionPath` in its return value so `runScorersForItem` can forward per-step `input` / `output` to scorers.
- `executeWorkflow` should propagate `requestContext` and the observability context into `run.start(...)` so that nested agent/tool spans land in the same trace and reach the observability store (matching what third-party exporters like Langfuse already receive).

As a result, the probe scorer in step 6 above would have access to the step's input/output, and the trace in step 7 would contain the nested `agent_run` → `tool_call` hierarchy.

## Environment Information

- OS: macOS 15.7.4
- Node.js: v25.2.1
- `@mastra/core`: 1.26.0
- `@mastra/observability`: 1.10.0
- `@mastra/libsql`: 1.9.0
- Storage: `MastraCompositeStore` with `LibSQLStore` default + observability domain on `DuckDBStore`
- Observability exporters: `DefaultExporter` (local) + `LangfuseExporter`

intojhanurag:
## Description

`dataset.startExperiment` for workflow targets had two gaps versus `runEvals`: scorers received only the workflow's final output (no per-step input/output), and the persisted observability trace contained only the outer `workflow_run` and `workflow_step` spans because direct agent calls inside steps started detached child traces. This PR surfaces step-level data to scorers and makes inner spans automatically nest under the workflow step.

## Before
image

## After
image

## Related Issue(s)

Fixes #15613

## Type of Change

- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [ ] Test update

## Checklist

### DIFF
--- a/.changeset/silent-lions-draw.md
+++ b/.changeset/silent-lions-draw.md
@@ -1,0 +1,5 @@
+---
+'@mastra/core': patch
+---
+
+Fixed `dataset.startExperiment` for workflow targets to match `runEvals`. Previously, scorers running inside a persisted experiment could not access per-step input or output, `requestContext` configured on the experiment was not forwarded into the workflow run, and direct agent calls inside workflow steps could start detached traces instead of nesting under the workflow step span. Step-level data is now exposed to scorers via `run.targetMetadata.stepResults` and `run.targetMetadata.stepExecutionPath`, the workflow's root span ID is available as `run.targetSpanId`, `requestContext` propagates into every step, and ambient workflow step tracing is used when creating nested spans. Fixes [#15613](https://github.com/mastra-ai/mastra/issues/15613).
--- a/packages/core/src/datasets/experiment/__tests__/executor.test.ts
+++ b/packages/core/src/datasets/experiment/__tests__/executor.test.ts
@@ -310,6 +310,69 @@ describe('executeTarget', () => {
       expect(result.output).toEqual({ processed: true });
       expect(result.error).toBeNull();
     });
+
+    it('surfaces stepResults, stepExecutionPath and spanId from a successful run', async () => {
+      const stepResults = {
+        chat: { status: 'success', payload: { prompt: 'hi' }, output: { text: 'hello' } },
+      };
+      const mockWorkflow = createMockWorkflow({
+        status: 'success',
+        result: { text: 'hello' },
+        steps: stepResults,
+        stepExecutionPath: ['chat'],
+        traceId: 'trace-1',
+        spanId: 'span-1',
+      });
+
+      const result = await executeTarget(mockWorkflow, 'workflow', {
+        id: 'item-7',
+        datasetId: 'ds-1',
+        input: { prompt: 'hi' },
+        groundTruth: null,
+        version: new Date(),
+        createdAt: new Date(),
+        updatedAt: new Date(),
+      });
+
+      expect(result.output).toEqual({ text: 'hello' });
+      expect(result.stepResults).toEqual(stepResults);
+      expect(result.stepExecutionPath).toEqual(['chat']);
+      expect(result.traceId).toBe('trace-1');
+      expect(result.spanId).toBe('span-1');
+    });
+
+    it('forwards requestContext and observability context into run.start', async () => {
+      const startSpy = vi.fn().mockResolvedValue({ status: 'success', result: {}, steps: {} });
+      const mockWorkflow = {
+        id: 'test-workflow',
+        name: 'Test Workflow',
+        createRun: vi.fn().mockResolvedValue({ start: startSpy }),
+      } as unknown as Workflow;
+
+      await executeTarget(
+        mockWorkflow,
+        'workflow',
+        {
+          id: 'item-8',
+          datasetId: 'ds-1',
+          input: { prompt: 'hi' },
+          groundTruth: null,
+          version: new Date(),
+          createdAt: new Date(),
+          updatedAt: new Date(),
+        },
+        { requestContext: { tenantId: 't-1' } },
+      );
+
+      expect(startSpy).toHaveBeenCalledTimes(1);
+      const startArgs = startSpy.mock.calls[0][0];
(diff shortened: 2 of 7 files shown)


intojhanurag:
there are two different issues mentioned by user that I fixed in the same PR :
- fix dataset.startExperiment to return step data ( related to Dataset )
- fix getOrCreateSpan to use ambient parent span  ( related to observability )

YujohnNattrass (approved):

PR #15797, opened 2026-04-27
fix(server,deployer): preload agent-builder during server init

mfrachet:
## Summary

When a user opts into the agent-builder via `MastraEditor`, the agent should appear in Studio's agent list on first load. Previously it didn't — it would only show up after a refresh, because the builder agent was registered lazily on the first call to `GET /editor/builder/settings`, which Studio fetches *after* `GET /agents`.

This PR makes the builder agent register eagerly during server startup, behind the existing license-and-opt-in gates, so it's already on the Mastra instance before the first HTTP request hits.

## Changes

- **`@mastra/server`** — Add `MastraServer.preloadBuilder()` to the server adapter base class. It calls `editor.resolveBuilder()` only if the user has explicitly enabled builder (`hasEnabledBuilderConfig()`). Wired into `init()` after `validateAgentBuilderLicense()`, so the EE license check runs first.
- **`@mastra/deployer`** — The Hono server in `createHonoServer` inlines its own init flow rather than calling `init()`, so it was bypassing both `validateAgentBuilderLicense()` and the new `preloadBuilder()`. Both calls are now added there too, immediately after `validateEELicense()`.
- **`@mastra/editor`** — Builder-agent registration plumbing.

## Gating (unchanged)

The new code path is reachable only when **both** existing gates pass:

1. **Explicit opt-in**: user must set `builder: { enabled: true }` in `MastraEditorConfig`. Otherwise `preloadBuilder()` short-circuits, the dynamic `import('./ee')` never runs, and the EE bundle is never loaded.
2. **Production EE license**: `validateAgentBuilderLicense()` runs immediately before `preloadBuilder()`. In production without `MASTRA_EE_LICENSE`, it throws and aborts server boot — `preloadBuilder()` is never reached.
3. Dev/test environments work the same as today: opt-in is enough.

This PR does not widen who sees the builder agent; it just moves the registration earlier in the boot sequence.

## Test plan

- `pnpm --filter ./packages/server test` — 11 license/preload tests pass (4 new `preloadBuilder` cases).
- `pnpm --filter ./packages/deployer test` — 365 tests pass (3 mock files updated to include `getEditor`).
- `pnpm --filter ./packages/editor test` — existing implicit-registration test still passes.
- Manual: `examples/agent` with `builder.enabled: true`, run `pnpm dev`, open Studio. `agent-builder` appears in the agent list on first load.
- Manual: `builder.enabled: false`, restart, confirm `agent-builder` does NOT appear and no `@mastra/editor/ee` code is loaded.

## Changeset

`@mastra/deployer`, `@mastra/server` patch.


wardpeet:
@mfrachet is this platform only PR?

mfrachet:
It's an EE feature that should go OSS: goal is to expose an agent that is only available for EE users and that we can query to create agents with a chat interface (CF last friday demo)

PR #15798, opened 2026-04-27
fix(core): send plain text to title generation model instead of JSON

brandonlbarros:
## Summary
- Title generation was passing `JSON.stringify(partsToGen)` to the model, which included `TextPart` object structure and `providerOptions` metadata (e.g. `"mastra"`, `"createdAt"`). This caused the title model to produce titles referencing internal framework details.
- Changed to `partsToGen.map(p => p.text).join('\n')` so only plain text content reaches the title model.
- Added regression tests verifying the model receives plain text without JSON artifacts or metadata leakage.

## Test plan
- [x] Existing title generation tests pass (38/38)
- [x] New test: plain text message sends only text content, not JSON-serialized TextPart objects
- [x] New test: file + text message doesn't leak `providerOptions`, `mastra`, or `createdAt` metadata

## ELI5

This PR stops the title-generation model from seeing internal JSON or metadata and instead sends only the plain human-readable text (including formatted multi-turn conversation context when needed), preventing titles that mention internal framework details.

## Overview

Title generation previously sent JSON.stringify(partsToGen) (including TextPart objects and providerOptions/mastra/createdAt metadata) to the model, which caused titles to reference internal framework details. This change extracts and joins plain text from UI parts and messages before sending to the title model, and adds regression tests to ensure no JSON or internal metadata leaks.

## Changes

- packages/core/src/agent/agent.ts
  - Replace JSON-serialized part payloads for title generation with plain-text formatting helpers that extract role-prefixed, human-readable content from UI message parts.
  - Support multi-message (multi-turn) title inputs: `generateTitleFromUserMessage` accepts an optional `messages` parameter (making `message` optional), formats messages and parts (text, tool calls/results, reasoning, files, URLs) into newline-delimited plain text, and returns `undefined` when formatted content is empty.
  - `genTitle` extended with an optional `uiMessages` argument and forwards `uiMessages` to title generation so both single-message and multi-message flows use the plain-text path.
  - Ensures both title model code paths receive a string assembled from parts text (partsToGen.map(p => p.text).join('\n')) rather than JSON.stringify.

- packages/core/src/agent/__tests__/title-generation.test.ts
  - Add three regression tests to verify title prompt hygiene:
    - Assert the title model receives only plain human-readable text (captures the user prompt and verifies no JSON markers like "type" or '{"').
    - Ensure mixed file + text parts produce readable text and file descriptions without JSON structure or internal metadata keys (e.g., `providerOptions`, `mastra`, `createdAt`).
    - Verify multi-turn conversations (including tool call/result parts) are formatted with role prefixes and labels while excluding internal identifiers/metadata (`providerOptions`, `toolCallId`, etc.).
  - Tests instrument mock models to capture raw prompts sent to the title model.

- .changeset/pretty-days-sort.md
  - Add patch-level changeset for @mastra/core describing the fix and multi-turn plain-text formatting.

## Testing

- Existing title generation tests continue to pass (38/38).
- New regression tests confirm:
  - No JSON-serialized TextPart objects are sent.
  - No leakage of providerOptions/mastra/createdAt/toolCallId or similar internal metadata.
  - Multi-turn context is correctly formatted as plain text.

## Public API / Exports

- None (behavioral changes internal to agent and added tests only).

## Review notes

- Focus review on the message-to-text conversion helpers and formatting for each part type (text, file, URL, tool call/result, reasoning) to ensure no metadata can leak and multi-turn formatting meets expectations.

### DIFF
--- a/packages/core/src/agent/agent.ts
+++ b/packages/core/src/agent/agent.ts
@@ -2254,7 +2254,7 @@ export class Agent<
           [
             {
               role: 'user',
-              content: JSON.stringify(partsToGen),
+              content: partsToGen.map(p => p.text).join('\n'),
             },
           ],
           'input',
@@ -2280,7 +2280,7 @@ export class Agent<
           },
           {
             role: 'user',
-            content: JSON.stringify(partsToGen),
+            content: partsToGen.map(p => p.text).join('\n'),
           },
         ],
       });
--- a/packages/core/src/agent/agent.ts
+++ b/packages/core/src/agent/agent.ts
@@ -1,5 +1,5 @@
 import { randomUUID } from 'node:crypto';
-import type { TextPart, UIMessage } from '@internal/ai-sdk-v4';
+import type { UIMessage } from '@internal/ai-sdk-v4';
 import type { ModelMessage } from '@internal/ai-sdk-v5';
 import { wrapSchemaWithNullTransform } from '@mastra/schema-compat';
 import type { StandardSchemaWithJSON } from '@mastra/schema-compat/schema';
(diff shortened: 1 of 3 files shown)


packages/core/src/agent/agent.ts
```diff
@@ -2280,7 +2280,7 @@ export class Agent<
           },
           {
             role: 'user',
-            content: JSON.stringify(partsToGen),
```
TylerBarnes, line 2283: this does more than remove the json structure, take a look at how observational memory converts message json into prompts for some ideas

TylerBarnes (changes requested):

packages/core/src/agent/agent.ts
brandonlbarros: Got it will do

TylerBarnes: oh reason I mention this btw is someone added an option for title gen to happen after x number initial messages. So I think the parts can now include assistant messages

brandonlbarros: So if I understand correctly, there are 2 issues:

1. (reported by Netanel) too much noise from passing in the JSON
2. Generating the title when message # > 1 misses context from previous messages

So, can copy similar structure to the formatMessagesForObserver() function, looping over all the messages and extracting text + some other relevant info. Will implement and get the update up asap

TylerBarnes (approved):

PR #15800, opened 2026-04-27
fix(core): pass requestContext and workspace to needsApprovalFn in network loop

### LINKED ISSUE #15799: [BUG] requireApproval predicate doesn't receive { requestContext, workspace } in network loop
## Describe the Bug

The `requireApproval` predicate type signature on tools accepts an optional second argument `{ requestContext?, workspace? }`, and the agentic-execution path passes both. The **network loop** (`agent.network(...)`) does not — it calls `needsApprovalFn` with only `inputData`. This is a runtime/type inconsistency between the two execution paths.

**Type signature** ([`packages/core/src/tools/types.ts:402-407`](https://github.com/mastra-ai/mastra/blob/main/packages/core/src/tools/types.ts#L402-L407)) — promises both args:

```ts
requireApproval?:
  | boolean
  | ((
      input: TSchemaIn,
      ctx?: { requestContext?: Record<string, unknown>; workspace?: Workspace },
    ) => boolean | Promise<boolean>);
```

**Agentic-execution path** ([`packages/core/src/loop/workflows/agentic-execution/tool-call-step.ts:293-296`](https://github.com/mastra-ai/mastra/blob/main/packages/core/src/loop/workflows/agentic-execution/tool-call-step.ts#L293-L296)) — passes both:

```ts
const needsApprovalResult = await (tool as any).needsApprovalFn(args, {
  requestContext: requestContext ? Object.fromEntries(requestContext.entries()) : {},
  workspace: _internal?.stepWorkspace,
});
```

**Network loop** ([`packages/core/src/loop/network/index.ts:1536`](https://github.com/mastra-ai/mastra/blob/main/packages/core/src/loop/network/index.ts#L1536)) — passes only `inputData`:

```ts
const needsApprovalResult = await (tool as any).needsApprovalFn(inputDataToUse);
```

Function predicates that read user role, workspace tier, or other request-scoped state work in `agent.generate()` / `agent.stream()` but silently fail in `agent.network()` (the second arg is `undefined`).

## Steps To Reproduce

```ts
import { Agent, createTool, RequestContext } from '@mastra/core';

const gated = createTool({
  id: 'gated',
  description: 'gated by role',
  requireApproval: (_input, ctx) => {
    // Works in agent.generate / agent.stream — `ctx` is populated.
    // Fails in agent.network — `ctx` is undefined.
    return ctx?.requestContext?.role !== 'admin';
  },
  execute: async () => ({ ok: true }),
});

const agent = new Agent({
  id: 'demo',
  name: 'demo',
  instructions: 'Use the gated tool when asked.',
  model: /* any model */,
  tools: { gated },
});

const requestContext = new RequestContext();
requestContext.set('role', 'admin');

// Network path — second arg is undefined, predicate sees no role and treats user as non-admin.
await agent.network('use the gated tool', { requestContext });
```

The fix PR (with a unit test that exercises the predicate's argument shape) is the minimal repro: see the unit test added in `packages/core/src/loop/network/index.test.ts` — `evaluateNetworkToolApproval` "passes inputData and { requestContext, workspace } to needsApprovalFn".

(Happy to spin up a separate `create-mastra` reproduction repo if a maintainer prefers.)

## Expected Behavior

The network loop should pass `{ requestContext, workspace }` to `needsApprovalFn` as the second argument, identical to the agentic-execution path. Function-form `requireApproval` predicates should be able to gate approval on request-scoped state regardless of which execution path the tool runs through.

## Environment Information

```block
Internal Mastra inconsistency — reproducible from any environment running @mastra/core.

Verified against `main` at commit 2082925 (latest as of report).
Affected file: packages/core/src/loop/network/index.ts:1536
Type signature: packages/core/src/tools/types.ts:402-407
Reference behavior: packages/core/src/loop/workflows/agentic-execution/tool-call-step.ts:293-296
```

## Verification

mbenhamd:
Fixes #15799.

The `requireApproval` predicate's type signature already declares an optional second argument (`{ requestContext, workspace }`), and the agentic-execution path at `loop/workflows/agentic-execution/tool-call-step.ts:293` passes it. The network loop at `loop/network/index.ts:1536` was still calling with only the input — function predicates running through `agent.network()` could not gate approval on user role / workspace tier.

This brings the two paths into agreement.

Before:

```ts
const needsApprovalResult = await (tool as any).needsApprovalFn(inputDataToUse);
```

After:

```ts
const toolRequiresApproval = await evaluateNetworkToolApproval({
  tool,
  inputData: inputDataToUse,
  requestContext,
  agent,
  logger,
  toolId,
});
```

`evaluateNetworkToolApproval` is a small exported helper that mirrors the agentic-execution shape and gives us a clean unit-test surface (the network loop's `toolStep` is otherwise internal). `agent.getWorkspace({ requestContext })` is wrapped in `.catch(() => undefined)` so predicate evaluation never throws on workspace lookup failure — the type already declares `workspace?:` optional.

Tests cover: boolean-only `requireApproval`, missing predicate, predicate input/ctx shape, true/false returns, predicate throwing (defaults to requiring approval), and workspace fallback when `getWorkspace` rejects.

## ELI5

The network execution path in Mastra wasn't giving approval predicates the same information that the regular agentic execution path was giving them. This meant predicates couldn't make approval decisions based on who was making the request or what workspace they were using. This PR fixes that by making both paths pass the same context information to approval predicates.

## Changes

**Changeset Documentation**
- Added release note documenting the observable behavioral change: `requireApproval` predicates evaluated in `agent.network()` now receive `{ requestContext, workspace }` as the second argument, matching the agentic-execution behavior. This enables approval gating based on user role, workspace tier, or other request-scoped state.

**New Exported Function**
- Added `evaluateNetworkToolApproval()` in `packages/core/src/loop/network/index.ts` — a reusable helper that evaluates approval predicates with proper context. It accepts tool configuration, input data, request context, agent instance, logger, and tool ID. Returns `Promise<boolean | undefined>`.

**Implementation Details**
- The function passes `{ requestContext, workspace }` to `needsApprovalFn`, where workspace is retrieved via `agent.getWorkspace({ requestContext })` 
- Wraps the workspace lookup in `.catch(() => undefined)` so failures don't throw; the predicate still runs with `workspace: undefined`
- Defaults to requiring approval (`true`) if the predicate throws or rejects, with errors logged via `logger.error` including the tool ID
- Simplified the tool-execution step's approval-check control flow to use the new helper instead of inline predicate evaluation

**Test Coverage**
- Added comprehensive test suite covering:
  - Boolean-only `requireApproval: true` returning `true` when no predicate exists
  - Missing predicate returning `undefined`
  - Predicate receiving correct argument shape: `(inputData, { requestContext, workspace })`
  - Boolean predicate results (`true`/`false`) being propagated correctly
  - Predicate errors/rejections defaulting to approval required with error logging
  - Workspace lookup failures allowing predicate to run with `workspace: undefined`

### DIFF
--- a/.changeset/metal-bottles-serve.md
+++ b/.changeset/metal-bottles-serve.md
@@ -1,0 +1,15 @@
+---
+'@mastra/core': patch
+---
+
+Fixed: `requireApproval` predicate now receives `{ requestContext, workspace }` as the second argument when invoked from the network loop, matching the existing behavior in the agentic-execution path. Function predicates can now gate approval on user role, workspace tier, or other request-scoped state regardless of whether the tool runs in `agent.network()` or the standard agentic loop.
+
+```ts
+createTool({
+  id: 'delete-account',
+  requireApproval: (input, { requestContext, workspace }) => {
+    return requestContext.role !== 'admin';
+  },
+  execute: async () => { /* ... */ },
+});
+```
--- a/packages/core/src/loop/network/index.test.ts
+++ b/packages/core/src/loop/network/index.test.ts
@@ -5,7 +5,7 @@ import type { Processor } from '../../processors';
 import { RequestContext } from '../../request-context';
 import { createTool } from '../../tools';
 import { createWorkflow } from '../../workflows';
-import { getLastMessage, getRoutingAgent } from './index';
+import { evaluateNetworkToolApproval, getLastMessage, getRoutingAgent } from './index';
 
 describe('getLastMessage', () => {
   it('returns string directly', () => {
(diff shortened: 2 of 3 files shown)


mbenhamd:
Closing this — I just realized `.network()` is deprecated and will be removed in a future major release, so adding code (even a small bugfix) to that path is the wrong move. The recommended supervisor pattern (`agent.stream()` / `agent.generate()`) already passes `{ requestContext, workspace }` through the agentic-execution path at `loop/workflows/agentic-execution/tool-call-step.ts`, so this gap doesn't exist on the path you're directing users toward.

Leaving #15799 open as a migration-time gotcha so anyone hitting this while still on `.network()` finds the answer ("migrate, don't patch"). Sorry for the noise.

mikhael28:
Thank you for self-closing @mbenhamd !

PR #15829, opened 2026-04-28
feat(core): throttle background task progress output

### LINKED ISSUE #15803: [FEATURE] BackgroundTaskManager: optional progressThrottleMs to coalesce high-frequency progress chunks
## Problem Statement

`BackgroundTaskManager.onProgress` (`packages/core/src/background-tasks/manager.ts:561-566`) calls `publishLifecycleEvent('task.output', ...)` once per chunk emitted by the tool executor. There's no built-in coalescing — every chunk is serialized, pushed through pubsub, and reaches every subscriber.

For long-running tools that emit fine-grained progress (research/scan-style executors emitting 50–200 progress events per second), this floods the SSE stream, downstream stores, and the consumer's render loop. Today every consumer that cares has to reinvent the same coalescing wrapper.

A sibling `progressIntervalMs` already exists at `manager.ts:277` for `waitForNextTask` elapsed-time pings, but it doesn't apply to chunk-based progress emission.

## Proposed Solution

Add an optional `progressThrottleMs?: number` field to `BackgroundTaskManagerConfig`. When set, the chunk-based `onProgress` builder applies a leading-edge throttle — drops intermediate chunks within the window. Default `undefined` keeps current behavior. The terminal `task.completed` / `task.failed` events are unaffected, so completion is never lost.

```ts
// types.ts (extend BackgroundTaskManagerConfig)
progressThrottleMs?: number;

// manager.ts:561 (replace onProgress builder)
const throttleMs = this.config.progressThrottleMs;
let lastProgressEmitMs = 0;
const onProgress = async (chunk: any) => {
  if (throttleMs) {
    const now = Date.now();
    if (now - lastProgressEmitMs < throttleMs) return;
    lastProgressEmitMs = now;
  }
  await this.publishLifecycleEvent('task.output', { ...task, chunk });
};
```

## Alternatives Considered

- **AI SDK `smoothStream`** — server-side but typed `TransformStream<TextStreamPart<TOOLS>, ...>`. Only sees text-delta chunks. Mastra's `BackgroundTaskProgressChunk` (`packages/core/src/stream/types.ts:762`) doesn't pass through this transform.
- **AI SDK `experimental_throttle` on `useChat`** — client-side render throttle. Doesn't reduce server-side serialization or pubsub volume.
- **Per-consumer wrapping** — works but every subscriber has to reinvent it.

## Example Use Case

A research-style tool emits ~100 fine-grained progress events per second while traversing results. Setting `progressThrottleMs: 500` on the manager coalesces upstream — fewer pubsub publishes, smaller SSE frames on the wire, smoother UI without the consumer needing custom debounce logic.

```ts
const manager = new BackgroundTaskManager({
  enabled: true,
  progressThrottleMs: 500,
});
```

## Additional Context

Naming mirrors existing siblings in `BackgroundTaskManagerConfig`: `defaultTimeoutMs`, `waitTimeoutMs`, and `progressIntervalMs`. Layer sits below AI SDK's transform pipeline and below `useChat` re-render throttle, so it doesn't conflict with the v4/v5/v6 message-format compatibility surface.

Verified no existing throttle on chunk-based `onProgress` by reading `manager.ts` (975 lines) and grepping for `throttle|debounce|minUpdateInterval|lastEmit|progressInterval` — only the elapsed-time `progressIntervalMs` came up, which targets a different callback signature.

Happy to send a PR (~30 LOC + 3 tests + changeset, leading-edge throttle, default off) once a maintainer confirms direction.

mbenhamd:
Adds an optional progressThrottleMs setting to background task manager config so high-frequency progress output chunks can be coalesced before pubsub and stream consumers. Terminal completion and failure events are unchanged.

Example:

```ts
const mastra = new Mastra({
  backgroundTasks: {
    enabled: true,
    progressThrottleMs: 500,
  },
});
```

Closes #15803

## ELI5
Imagine a machine that works on tasks and keeps telling you how it's progressing. If it talks *really* fast (hundreds of times per second), that can overwhelm other systems listening to it. This PR adds a "slow down" setting that makes the machine only report progress once every X milliseconds, while always reporting when the task is finally done. It's like batching progress reports instead of sending them one-by-one.

## Changes Overview
This PR introduces a new `progressThrottleMs` configuration option to `BackgroundTaskManager` that applies optional rate-limiting to high-frequency progress output before it reaches pubsub and streaming consumers.

## Key Changes

**Configuration**
- Added `progressThrottleMs?: number` property to `BackgroundTaskManagerConfig` (optional, undefined by default to preserve existing behavior)

**Implementation**
- Modified the `onProgress` handler in `BackgroundTaskManager` to check if throttling is enabled and configured
- When throttling is active, intermediate progress chunks emitted within the throttle window are skipped
- The throttle uses a leading-edge strategy: the first chunk is always emitted, and subsequent chunks are dropped until the configured interval has elapsed
- Terminal events (`task.completed` and `task.failed`) remain unaffected and always emit immediately, regardless of throttle state

**Testing**
- Added comprehensive test coverage including:
  - Default behavior verification (no throttling) with three progress chunks and completion event
  - Throttled behavior verification using mocked timestamps to confirm intermediate chunks are suppressed while allowing chunks after the throttle window, and completion is always emitted

## Impact
Reduces pubsub and SSE churn for high-frequency progress-emitting tools (tens–hundreds of events/sec), lowering downstream load and smoothing client render loops while maintaining task completion reliability.

### DIFF
--- a/.changeset/cool-pots-fall.md
+++ b/.changeset/cool-pots-fall.md
@@ -1,0 +1,14 @@
+---
+'@mastra/core': minor
+---
+
+Added `progressThrottleMs` to background task configuration so high-frequency progress output can be coalesced before it reaches pubsub and stream consumers.
+
+```ts
+const mastra = new Mastra({
+  backgroundTasks: {
+    enabled: true,
+    progressThrottleMs: 500,
+  },
+});
+```
--- a/packages/core/src/background-tasks/manager.test.ts
+++ b/packages/core/src/background-tasks/manager.test.ts
@@ -884,6 +884,124 @@ describe('BackgroundTaskManager', () => {
       abortController.abort();
     });
 
+    it('emits every progress output chunk by default', async () => {
+      const abortController = new AbortController();
+      const stream = manager.stream({ abortSignal: abortController.signal });
+      const reader = stream.getReader();
+
+      const chunk = (output: string) => ({
+        type: 'tool-output',
+        runId: 'run-1',
+        from: 'AGENT',
+        payload: { output, toolCallId: 'c1', toolName: 'tool' },
+      });
+
+      const executeFn = vi.fn().mockImplementation(async (_args: any, opts: { onProgress?: (chunk: any) => void }) => {
+        await opts.onProgress?.(chunk('first'));
+        await opts.onProgress?.(chunk('second'));
+        await opts.onProgress?.(chunk('third'));
+        return 'done';
+      });
+
+      await manager.enqueue(
+        { toolName: 'tool', toolCallId: 'c1', args: {}, agentId: 'a1', runId: 'run-1' },
+        ctx(executeFn),
+      );
+
+      await tick();
+
+      await reader.read(); // running
+
+      const firstOutput = await reader.read();
+      expect(firstOutput.value).toMatchObject({
+        type: 'background-task-output',
+        payload: { payload: { payload: { output: 'first' } } },
+      });
+
+      const secondOutput = await reader.read();
+      expect(secondOutput.value).toMatchObject({
+        type: 'background-task-output',
+        payload: { payload: { payload: { output: 'second' } } },
+      });
+
+      const thirdOutput = await reader.read();
+      expect(thirdOutput.value).toMatchObject({
+        type: 'background-task-output',
+        payload: { payload: { payload: { output: 'third' } } },
+      });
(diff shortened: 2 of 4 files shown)


taofeeq-deru (approved):
lgtm

PR #15835, opened 2026-04-28
fix(core): filter tool calls during agentic loop steps

### LINKED ISSUE #15811: ToolCallFilter run only once on processInput
This issue was created from Discord post 1498148138303623230:

[Open in Discord](https://discord.com/channels/1309558646228779139/1498148138303623230)

In a tool heavy agent, tool call results are added to each step in the agentic loop which adds up in the context. For example an agentic loop involved 10 steps and 2nd step had a tool call result adding 5k input token, that'd be added to each subsequent steps. 

I'm not sure if this is right approach, I added a custom processor with `processInputStep` which drops previous tool call parts from messages and then added an instruction bit to agent to summerize the result for later steps. It'd be helpful to have some built in mechanism to handle this. Perhaps update  `ToolCallFilter` with more options to control such behaviour.

mbenhamd:
This fixes ToolCallFilter so it also runs during each agentic loop step, not only on the initial input. Filtering now uses prompt-only modelContextMessages, so older tool calls and results can be removed from the next model prompt without deleting canonical MessageList history, memory data, UI data, or final tool results. The latest completed tool step is preserved by default so the model can use fresh tool output once before it is pruned from later context.

It also adds validation for ambiguous processInputStep returns, covers markerless harness-style histories, and updates the processor reference docs.

Closes https://github.com/mastra-ai/mastra/issues/15811

Validated with focused core Vitest suites for ToolCallFilter, processInputStep, agent processors, and AI SDK v5/v6 message-list compatibility, plus core typecheck, core lint, docs validation, touched docs remark/prettier, and git diff --check.

## ELI5
Old tool outputs kept piling into the model's prompts and made them huge. This PR makes the filter hide older tool outputs at every agent loop step (so prompts stay small) while keeping the full history intact and letting the latest tool output be seen once by default.

## Changes Overview
ToolCallFilter now runs during every agentic loop step via processInputStep and filters against a prompt-only context (modelContextMessages). This prunes prior tool outputs from the next model prompt without mutating canonical MessageList, memory, UI data, or final tool results.

## Key Features
- Step-aware filtering: ToolCallFilter.processInputStep executes on initial input and each agent loop step.
- Latest-step preservation: new preserveLatestStep option (default true) preserves the most recent completed tool step’s output for one model consumption; setting false removes it every step.
- Prompt-only context: processors can return modelContextMessages to produce a prompt-only message set used for the next model prompt; it does not persist to the MessageList.
- Stronger validation: rejects processor returns that include both messages and modelContextMessages and errors when a processor mutates MessageList after entering model-context-only mode.
- Broad compatibility: generalized filtering supports AI SDK v5/v6 part shapes and legacy toolInvocations; generic tool parts without toolName are not inferred for exclusion.

## Implementation Details
- Added optional modelContextMessages to ProcessInputStepResult, step types/schemas, and runtime validation.
- Runner/workflow: runProcessInputStep/workflow prefer stepInput.modelContextMessages as the prompt source and avoid applying returned messages to MessageList when modelContextMessages is active; validatedResult.messages are applied to MessageList only when not using modelContextMessages.
- ToolCallFilter: refactored to a unified filterMessages pipeline over "tool-like" parts, prunes content.toolInvocations, computes excluded tool-call IDs across parts and invocations, preserves IDs from the latest completed step (with step-start boundary logic), and trims orphan step markers.
- LLM step: llm-execution-step builds a derived MessageList from modelContextMessages (preserving system messages and source metadata) and uses it for aiV5.llmPrompt input when present.
- Provider & exports: ToolCallFilterOptions type added and exported; processor provider config accepts preserveLatestStep and includes processInputStep in available phases; ToolCallFilter constructor updated and processInputStep added.

## Documentation Updates
- Updated docs for processors, processor interface, and ToolCallFilter to document modelContextMessages, step-aware filtering, preserveLatestStep semantics, and agentic-loop behavior; usage examples and formatting adjusted.

## Testing & Validation
- Added/expanded Vitest coverage:
  - ToolCallFilter tests for processInput and processInputStep across v5/v6 part shapes, legacy toolInvocations, preserveLatestStep behavior, and markerless histories.
  - processInputStep tests for modelContextMessages propagation, isolation from MessageList, invalid-return guardrails, and mutation guards.
  - Agent/workflow/processor-step tests ensuring prompt-only context is used for LLM input and MessageList mutations are prevented.
  - agent-processor test validating prompt filtering while preserving final toolResults.
- CI checks updated: typecheck, lint, docs validation, remark/prettier formatting, and git diff --check.

## Public API Surface
- Added exported type: ToolCallFilterOptions.
- ToolCallFilter constructor signature updated and public processInputStep method added.
- Re-exported ToolCallFilterOptions from processors barrel.
- modelContextMessages fields added to processor step types/schemas.

## Resolves
Closes issue #15811: prevents repeated accumulation of prior tool call results across agent loop steps.

### DIFF
--- a/.changeset/quick-bags-tap.md
+++ b/.changeset/quick-bags-tap.md
@@ -1,0 +1,5 @@
+---
+'@mastra/core': patch
+---
+
+Added prompt-only processor context for agent loop messages so processors can reduce model prompt size without mutating canonical run history.
--- a/docs/src/content/en/docs/agents/processors.mdx
+++ b/docs/src/content/en/docs/agents/processors.mdx
@@ -375,7 +375,7 @@ See the [`TokenLimiterProcessor` reference](/reference/processors/token-limiter-
 
 Removes tool calls and results from messages sent to the LLM, saving tokens on verbose tool interactions. Optionally exclude only specific tools. This filter only affects the LLM input, filtered messages are still saved to memory.
 
-By default, `ToolCallFilter` filters the initial input before the agent loop starts. Use `filterAfterToolSteps` to also filter during each loop step while preserving recent tool-producing steps.
+By default, `ToolCallFilter` filters the initial input before the agent loop starts. Use `filterAfterToolSteps` to also filter the next model prompt during each loop step while preserving recent tool-producing steps. Step filtering does not mutate canonical run history or remove final tool results.
 
 ```typescript
 new ToolCallFilter({
--- a/docs/src/content/en/reference/processors/processor-interface.mdx
+++ b/docs/src/content/en/reference/processors/processor-interface.mdx
@@ -170,7 +170,7 @@ Most processor methods receive both `messages` and `messageList`. They point to 
 
 ### `messages` vs `messageList`
 
-- `messages`: A plain array of `MastraDBMessage` objects, scoped to the current stage. For `processInput` and `processInputStep` this excludes system messages. For `processOutputResult` and `processOutputStep` this includes the latest LLM response. The array is backed by `messageList`, so editing a message's `content.parts` in place is visible to downstream processors and to persistence.
+- `messages`: A plain array of `MastraDBMessage` objects, scoped to the current stage. For `processInput` this excludes system messages. For `processInputStep`, this is usually derived from `messageList`; after an earlier step processor returns `modelContextMessages`, it is the prompt-only message context for the current step. For `processOutputResult` and `processOutputStep` this includes the latest LLM response.
 - `messageList`: The live `MessageList` instance backing the run. It exposes filtered views (input, response, remembered, all), multiple output formats (db, ui, core), and methods for mutating the conversation.
 
 Use `messages` when you only need to read, map over, or lightly edit fields on the current stage's messages. Use `messageList` when you need to:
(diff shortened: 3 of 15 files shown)


mbenhamd:
Benchmark update for the prompt-only step context change.

I measured the new branch flow with temporary scripts outside the repo so no benchmark files are added here.

Processor-step CPU overhead, median vs baseline:

| Synthetic tool steps | Active filter | Active filter + downstream processor |
|---:|---:|---:|
| 10 | +0.036 ms | +0.038 ms |
| 50 | +0.077 ms | +0.082 ms |
| 200 | +0.289 ms | +0.302 ms |
| 500 | +0.724 ms | +0.727 ms |

Pre-model prompt preparation after the follow-up optimization in `6f55f1e`:

| Synthetic tool steps | Added prompt-prep latency | Prompt messages |
|---:|---:|---:|
| 10 | +0.030 ms | 31 -> 13 |
| 50 | +0.103 ms | 151 -> 53 |
| 200 | +0.930 ms | 601 -> 203 |
| 500 | +5.919 ms | 1501 -> 503 |

Interpretation: the new CPU cost is linear and small. In large tool-heavy histories, it trades a few local milliseconds for a much smaller model prompt, which should reduce provider latency/cost more than the local prep cost adds.

Validation after the perf change:

- `pnpm --filter ./packages/core exec vitest run src/processors/processors/tool-call-filter.test.ts src/processors/process-input-step.test.ts src/agent/agent-processor.test.ts src/workflows/processor-step.test.ts src/loop/workflows/agentic-execution/llm-execution-step.test.ts` -> 189 tests passed
- `pnpm --filter ./packages/core check` -> passed
- `pnpm --filter ./packages/core lint` -> passed

amitava82:
Any update on this getting merged?

PR #15842, opened 2026-04-28
fix(playground): persist Review tab comments across reload

BrandiATMuhkuh:
## Summary

On the studio Review tab, tags survive a page reload but typed comments disappear. The rehydrate hooks were hardcoding `comment: ''` when mapping experiment results into review items, so the comment that gets written to the feedback store on textarea blur was never read back.

## Fix

Comments are persisted via `client.createFeedback({ feedbackType: 'comment', sourceId: resultId, ... })` on blur. This PR adds the matching read path: a small helper that uses the existing `listFeedback` endpoint to fetch the latest comment-type feedback per experiment result, and wires it into the four rehydrate hooks.

- New `packages/playground/src/domains/review/utils/fetch-result-comments.ts` — one fetch per experiment, indexed by `sourceId`, latest-by-timestamp wins.
- Updated `use-review-items`, `use-completed-items` (agents) and `use-dataset-review-items` (both review and completed queries) to call the helper alongside `listDatasetExperimentResults` and populate `comment` from the returned map instead of the hardcoded `''`.

No schema or server changes; the write path is unchanged.

## Test plan

- [ ] Open Review tab on an agent, add a comment to an item, blur the textarea, reload — comment is restored.
- [ ] Same flow on the dataset Review tab.
- [ ] Same flow for a completed (Mark as complete) item — comment still shown read-only.
- [ ] Items with no comment continue to show an empty textarea.
- [ ] Items whose experiment has no traceId continue to work for tags/rating (no regression).

## ELI5
When you type a comment in the Review tab and refresh the page, your comment was disappearing—even though it was actually being saved. This PR fixes it by making sure comments are loaded back from storage when the page reloads, the same way other feedback (like tags) already were.

---

## Summary
This PR fixes a client-side rehydration bug in the playground's Review tab where typed comments were not being restored after a page reload. While comment writes were working correctly (saved via `client.createFeedback()` on textarea blur), the rehydrate/load hooks were hardcoding `comment: ''` instead of reading the saved comments back from the feedback store.

## Changes
- **New utility**: `packages/playground/src/domains/review/utils/fetch-result-comments.ts` — Exports `fetchResultCommentsByExperiment()`, which queries the existing `listFeedback` endpoint to retrieve all comment-type feedback for an experiment, returns the latest comment per result (keyed by `sourceId`, sorted by descending timestamp), and silently returns an empty `Map` on errors.

- **Updated rehydrate hooks** — Three hooks now call the new helper alongside `listDatasetExperimentResults` and populate `comment` from the fetched comments map instead of hardcoding empty strings:
  - `packages/playground/src/domains/agents/hooks/use-completed-items.ts`
  - `packages/playground/src/domains/agents/hooks/use-review-items.ts`
  - `packages/playground/src/domains/review/hooks/use-dataset-review-items.ts` (both review and completed query variants)

- **Changeset**: Documents the patch-level fix for `@internal/playground`.

## Impact
No schema or server changes. The feedback write path remains unchanged. Comments typed in Review tabs (agent and dataset) will now persist across page reloads, while completed items show read-only comments and items without comments display empty textareas as expected.

### DIFF
--- a/.changeset/review-tab-comment-persistence.md
+++ b/.changeset/review-tab-comment-persistence.md
@@ -1,0 +1,5 @@
+---
+'@internal/playground': patch
+---
+
+Fixed Review tab comments not persisting across page reload in the studio. Tags would survive a reload but typed comments would disappear, because the rehydrate path always reset the comment field to an empty string. Comments are now read back from the feedback store on load, matching how they are written when the textarea is blurred.
--- a/packages/playground/src/domains/agents/hooks/use-completed-items.ts
+++ b/packages/playground/src/domains/agents/hooks/use-completed-items.ts
@@ -2,6 +2,7 @@ import { useMastraClient } from '@mastra/react';
 import { useQuery } from '@tanstack/react-query';
 import type { ReviewItem } from '../context/review-queue-context';
 import { useAgentExperiments } from './use-agent-experiments';
+import { fetchResultCommentsByExperiment } from '@/domains/review/utils/fetch-result-comments';
 
 /**
  * Loads completed review items (status='complete') for auditing.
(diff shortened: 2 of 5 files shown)


BrandiATMuhkuh:
I'm closing this PR. I could not fully validate if it worked. As it seems there is also a server side problem.

PR #15871, opened 2026-04-28
ci: reuse prebuild outputs in vitest-changed

abhiaiyer91:
## Summary
- upload prebuild dist/.next outputs as a reusable artifact
- restore that artifact in vitest-changed unit shards and e2e jobs
- remove the duplicate vitest-changed repo build that was rerunning after Prebuild

## Test Plan
- actionlint .github/workflows/prebuild.yml .github/workflows/vitest-changed.yml
- node - <<'NODE'
  const fs = require('fs');
  const yaml = require('js-yaml');
  for (const path of ['.github/workflows/prebuild.yml', '.github/workflows/vitest-changed.yml']) {
    yaml.load(fs.readFileSync(path, 'utf8'));
    console.log(`YAML OK: ${path}`);
  }
  NODE
- git diff --check -- .github/workflows/prebuild.yml .github/workflows/vitest-changed.yml

## ELI5
Tests now reuse the build outputs the Prebuild job already made instead of rebuilding everything, so CI runs faster and uses fewer resources.

## Summary
Add uploading of prebuilt dist/.next outputs from the Prebuild workflow and restore those outputs in vitest-changed jobs so unit and e2e shards consume the prebuilt artifact instead of performing a duplicate repository build; tighten workflow-run origin checks and move shared pnpm install responsibility into the pnpm setup action.

### .github/workflows/prebuild.yml
- Require PR head repository to equal github.repository in addition to existing has_code == 'true' and repository checks.
- After pnpm turbo build, locate dist and .next directories (exclude node_modules and .next/cache*), fail if none are found, package them into prebuild-outputs.tar.gz, and upload as artifact prebuild-outputs (actions/upload-artifact) with 1-day retention.

### .github/workflows/vitest-changed.yml
- Restrict set-pending-status, test, and e2e-test jobs to workflow_run.conclusion == 'success' and workflow_run.head_repository.full_name == github.repository so only trusted runs from the main repo trigger downstream jobs.
- test and e2e-test download the prebuild-outputs artifact tied to the triggering workflow_run.id and extract prebuild-outputs.tar.gz instead of running a full pnpm turbo build, removing the duplicate build step.
- Explicit minimal permissions (actions: read, contents: read) are requested where needed for artifact access.

### .github/actions/setup-pnpm-node/action.yaml
- Stop running pnpm install in the composite action shell step; delegate installation to pnpm/action-setup by setting run_install based on inputs.install-dependencies. When enabled, installs are executed recursively with --frozen-lockfile and --ignore-scripts (suppresses lifecycle/package scripts).

## Commit note
One commit: "ci: move shared install into pnpm setup action" — delegate pnpm install to pnpm/action-setup with the same frozen-lockfile and add --ignore-scripts when enabled. Co-Authored-By: Mastra Code (openai/gpt-5.4).

## Rationale
Reusing Prebuild artifacts removes redundant rebuilds, speeds CI, reduces resource use, and narrows artifact consumption to trusted workflow runs while preventing untrusted package lifecycle scripts during installs.

### DIFF
--- a/.github/actions/setup-pnpm-node/action.yaml
+++ b/.github/actions/setup-pnpm-node/action.yaml
@@ -29,7 +29,7 @@ runs:
     - uses: pnpm/action-setup@v4
       name: Install pnpm
       with:
-        run_install: false
+        run_install: ${{ inputs.install-dependencies == 'true' && '[{"recursive":true,"args":["--frozen-lockfile","--ignore-scripts"]}]' || 'false' }}
         package_json_file: ${{ inputs.package-json-file || 'package.json' }}
         cache: ${{ inputs.package-manager-cache == 'true' }}
         cache_dependency_path: ${{ inputs.cache-dependency-path || 'pnpm-lock.yaml' }}
@@ -38,8 +38,3 @@ runs:
       uses: actions/setup-node@v6
       with:
         node-version: 22.13.0
-
-    - name: Install dependencies
-      if: ${{ inputs.install-dependencies == 'true' }}
-      shell: bash
-      run: pnpm install --frozen-lockfile --ignore-scripts
--- a/.github/actions/setup-pnpm-node/action.yaml
+++ b/.github/actions/setup-pnpm-node/action.yaml
@@ -42,4 +42,4 @@ runs:
     - name: Install dependencies
       if: ${{ inputs.install-dependencies == 'true' }}
       shell: bash
-      run: pnpm install --frozen-lockfile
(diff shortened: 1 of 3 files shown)


wardpeet:
you probably want to fallback to https://turborepo.dev/docs/guides/ci-vendors/github-actions#remote-caching-with-github-actionscache when turbo envs are not available for external contributors[

Today the vercel remote cache is working for us. a pnpm build takes 5s

PR #15897, opened 2026-04-29
fix(core): avoid duplicate agent loop response items

TylerBarnes:
This fixes duplicate provider response items in agent loops by removing the extra message-list append of the `messages.nonUser` snapshot. `llmExecutionStep` already builds stream chunks into response DB messages, so the follow-up map was adding the same response again after converting it through model-message shape.

Before:
```ts
.then(llmExecutionStep)
.map(addResponseSnapshotBackToMessageList)
.map(mapToolCalls)
```

After:
```ts
.then(llmExecutionStep)
.map(mapToolCalls)
```

This also removes the hidden duplicate file-part artifacts from loop snapshots. I added a regression test for duplicate OpenAI provider item IDs.

Tested with `pnpm exec vitest run --exclude '**/tool-builder/**' $(find src/loop -name '*.test.ts' -print) --bail=1 --reporter=dot`, `pnpm --filter ./packages/core check`, and `pnpm build:core`.

## ELI5

When the agent ran, the same AI response could get saved twice. This PR removes the extra step that re-added an already-built response so each model response is stored only once.

---

### Core fix: stop re-appending streamed assistant responses
- Root cause: llmExecutionStep already assembles streamed chunks into DB messages; a subsequent mapping step (addResponseSnapshotBackToMessageList) was re-appending the same assistant response converted to model-message shape, producing duplicates.
- Fix: removed the snapshot-and-reappend mapping so the pipeline is now .then(llmExecutionStep).map(mapToolCalls) instead of .then(llmExecutionStep).map(addResponseSnapshotBackToMessageList).map(mapToolCalls).
- File: packages/core/src/loop/workflows/agentic-execution/index.ts — removed ~35 lines of snapshot-and-reappend logic.

### Preserve consolidated streamed text and avoid hidden duplicate artifacts
- buildMessagesFromChunks now joins final text parts into a top-level content string when non-empty, ensuring streamed text is preserved as a single message and avoiding hidden duplicate file-part artifacts.
- File: packages/core/src/loop/workflows/agentic-execution/build-messages-from-chunks.ts — adds optional top-level content alongside parts/format.

### Regression test preventing duplicate provider item IDs
- New unit test constructs a streaming model that emits chunks with an OpenAI provider itemId already present in the input snapshot and asserts the persisted messages contain exactly one occurrence of that itemId.
- File: packages/core/src/loop/workflows/agentic-execution/index.test.ts — +129 lines (new test).

### Resume/tool-run id handling and tool-call changes
- Tool-call step refactors suspended run-id handling: lookup/removal of suspension metadata is centralized and resumed tool runs preserve locally stored suspended run ids before clearing suspension metadata to avoid replayed/stale run ids.
- File: packages/core/src/loop/workflows/agentic-execution/tool-call-step.ts — refactored logic and conditional ordering.

### Tests & snapshots adjustments
- Updated inline snapshots and expectations (normalizing createdAt timestamps and removing expectations for now-removed duplicated fields).
- Files: packages/core/src/loop/test-utils/streamObject.ts and packages/core/src/loop/test-utils/options.ts.

### Memory recordings & transform updates
- Updated several memory integration test recordings (hash and minor content normalizations) and improved transform-request.replaceField to redact numeric literals and JSON-string embedded values.
- Files: packages/memory/__recordings__/* and packages/memory/integration-tests/src/transform-request.ts.

### Release note / Changeset
- .changeset/kind-socks-smoke.md added documenting a patch release for @mastra/core describing the duplicate-response fix.

### Tests & Validation
- Added regression test for duplicate provider item IDs and updated snapshots.
- Local verification: targeted vitest loop tests, pnpm --filter ./packages/core check, and pnpm build:core.

### Commit messages of note
- fix(core): avoid duplicate agent loop response items by removing extra append of response snapshot
- fix(core): preserve suspended tool run ids on resume
- fix(core): populate top-level content when building messages from streamed chunks
- Memory recording/normalization updates; Co-Authored-By: Mastra Code (openai/gpt-5.5) <noreply@mastra.ai>


packages/core/src/loop/__snapshots__/loop.test.ts.snap
```diff
@@ -12781,7 +12700,6 @@ exports[`Loop Tests > AISDK v6 (V3 models) > result.steps > should add the files
       "dbMessages": [
         {
           "content": {
-            "content": "Hello!",
```
TylerBarnes, line 12784: all the "content" fields were because we were re-adding messages after converting them to ai sdk format (which has this duplicated content field)

CalebBarnes (approved):

abhiaiyer91 (approved):

packages/core/src/loop/workflows/agentic-execution/tool-call-step.ts
```diff
@@ -280,6 +280,46 @@ export function createToolCallStep<Tools extends ToolSet = ToolSet, OUTPUT = und
...
+
+            const foundTool = message.content.parts?.find(
+              (part: any) =>
+                (part.type === 'data-tool-call-suspended' || part.type === 'data-tool-call-approval') &&
+                !(part.data as any).resumed &&
+                part.data.toolName === inputData.toolName,
+            );
+            if (foundTool) {
+              return (foundTool as any).data.runId as string | undefined;
+            }
```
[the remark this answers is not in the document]
TylerBarnes: Adjusted in the latest push. Runtime metadata lookup still runs for agent/workflow resumes so the live suspended runId can win, but the older data-tool-call-* parts fallback is now scoped to the previous condition: only when this is not a resume tool call, or when the resume call omitted args.suspendedToolRunId. That avoids broadening stale historical part matching while preserving the resume fix.

packages/core/src/loop/workflows/agentic-execution/tool-call-step.ts
```diff
@@ -527,16 +523,18 @@ export function createToolCallStep<Tools extends ToolSet = ToolSet, OUTPUT = und
...
+                part =>
+                  (part.type === 'data-tool-call-suspended' || part.type === 'data-tool-call-approval') &&
+                  !(part.data as any).resumed,
+              );
+              if (dataToolSuspendedParts && dataToolSuspendedParts.length > 0) {
+                const foundTool = dataToolSuspendedParts.find((part: any) => part.data.toolName === inputData.toolName);
+                if (foundTool) {
+                  suspendedToolRunId = (foundTool as any).data.runId;
+                  break;
+                }
```
TylerBarnes, line 537: for pr history: this change was needed as the previous logic relied on ordering that only worked because of the bug that was fixed

PR #15307, opened 2026-04-13
feat: background task execution for agents

taofeeq-deru:
## Summary

Agents can now dispatch slow tool calls to run asynchronously while the conversation keeps streaming. The LLM receives a placeholder result immediately and continues generating; when the background tool finishes, its result flows back into the loop (if still open) or the message list for the LLM to process when queried.

```
User: "Research Solana for me, and what's Bitcoin's price?"

Agent: crypto-price runs synchronously → "Bitcoin is at $68,434"
       crypto-research dispatched to background → "I'm researching Solana…"
       (stream closes, UI shows 1 background task running)
       when background task is done, result is injected into the agent message list 
       UI shows 1 background task compeleted
//User can ask agent other question and even when background task is still running
User: what's the result of the research
Agent: agent processes solana research result → full analysis
```

## Architecture

Three subsystems working together:

1. **PubSub** (reuses the existing workflows pubsub) — two topics:
   - `background-tasks` with group subscription so workers compete for each dispatch / cancel message (only one worker picks up each task)
   - `background-tasks-result` with fan-out so every process hears lifecycle events (`task.running`, `task.completed`, `task.failed`, `task.cancelled`, `task.output`)

   Retries are delegated to the pubsub layer via `nack()` + `deliveryAttempt` tracking — no manual re-dispatch on the manager.

2. **Storage** — new `BackgroundTasksStorage` domain. `TaskFilter` supports status, agent / thread / resource / run / tool filters, date range (`fromDate` / `toDate` / `dateFilterBy`), ordering, and pagination (`page` / `perPage`). `listTasks()` returns `{ tasks, total }`.

3. **Agentic loop** — `tool-call-step` resolves bg config (LLM `_background` override → agent → tool → manager defaults), dispatches via `createBackgroundTask()`, and returns a placeholder. `backgroundTaskCheckStep` pauses the loop between LLM iterations with configurable `waitTimeoutMs` (Strategy B — process results as they arrive). `isTaskCompleteStep` skips scoring while a task is still pending.

## Key features

- **`createBackgroundTask(manager, options)`** — self-contained per-task handle with `dispatch()`, `cancel()`, `waitForCompletion()`. Per-task `TaskContext` (executor, `onChunk`, `onResult`, `onComplete`, `onFailed`) is keyed by task ID so concurrent streams never share state.
- **Schema + prompt injection** — `_background` field added to bg-eligible tool schemas (both JSON Schema and Zod) + generated system prompt section listing bg-eligible tools and defaults.
[...]

PR #15902, opened 2026-04-29
fix(clickhouse): use lightweight DELETE and push deleteTasks filter into ClickHouse

dbworku:
## Description

Two related improvements to the background-tasks delete paths in `@mastra/clickhouse`. One is a race condition and the other is a performance improvement. 

###  Switch deleteTask / deleteTasks from ALTER TABLE ... DELETE to lightweight DELETE FROM

Lightweight `DELETE FROM` (ClickHouse 23.3+) is synchronous by default, marks rows via a tombstone mask instead of rewriting parts, and is immediately visible to subsequent reads — so the mutations_sync workaround
is no longer needed and per-delete I/O drops.
- PR  #15307 introduce Clickhouse store for background tasks with a latent bug -- deletes are async.
- PR #15895 made deletes await mutation completion by setting `mutations_sync: '1'` on the `ALTER TABLE ... DELETE` mutation, and explicitly deferred the lightweight-DELETE alternative.

###  Push `deleteTasks` filtering into ClickHouse

Previously `deleteTasks(filter)` called `listTasks(filter)` and looped one mutation per matching id — every matching task id round-tripped through Node.js, followed by N separate delete statements (and N mutations
under the old ALTER TABLE path).

## Related Issue(s)

None filed.

## Type of Change

- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [ ] Test update

## Checklist

## ELI5

Deletes for background tasks in ClickHouse now run entirely in the database in one fast step, so removals are immediate and don't overload Node.js by shuttling lots of IDs back and forth.

### Implementation (stores/clickhouse/src/storage/domains/background-tasks/index.ts)
- Added private buildFilterClause(filter) to produce a SQL WHERE fragment and query params reused by listTasks (count + data queries).
- listTasks now reuses the filter clause for consistent counting and paginated queries.
- deleteTask now uses lightweight synchronous DELETE FROM ... WHERE id = {var_id} via client.command instead of ALTER TABLE ... DELETE + mutations_sync.
- deleteTasks now issues a single DELETE FROM ... WHERE ... with the filter pushed into ClickHouse; it returns early (no-op) when no filters are provided to avoid accidental full-table deletes (use dangerouslyClearAll for full wipes).

### Tests (stores/_test-utils/src/domains/background-tasks/index.ts)
- Added test coverage including a test that creates a task, deletes it with deleteTask, and asserts getTask returns null.

### Docs / Changeset
- README updated to require ClickHouse 23.3+ for delete operations (22.8–23.2 can work with allow_experimental_lightweight_delete = 1).
- Changeset documents the migration from ALTER TABLE ... DELETE (mutations_sync) to lightweight DELETE and the improvement from pushing filtering into ClickHouse; deleteTasks with no filters is now a no-op.

## Notes
- Fixes async-delete race and reduces per-delete I/O and network/OOM risk by avoiding per-id round trips.
- No public API surface changes.
- Change type: Bug fix.

### DIFF
--- a/stores/clickhouse/README.md
+++ b/stores/clickhouse/README.md
@@ -10,7 +10,9 @@ npm install @mastra/clickhouse
 
 ## Prerequisites
 
-- Clickhouse server (version 21.12 or higher recommended)
+- Clickhouse server (version 23.3 or higher recommended)
+  - Lightweight `DELETE FROM` requires ClickHouse 22.8+ with `allow_experimental_lightweight_delete = 1` (for 22.8–23.2), or 23.3+ where it is generally available.
+  - The `deleteTask` and `deleteTasks` methods use `DELETE FROM` on the background tasks table — ensure your server supports lightweight delete before using those operations.
 - Node.js 22.13.0 or later
 
 ## Usage
@@ -128,7 +130,6 @@ The store uses different table engines for different types of data:
 
 ### Operations Not Currently Supported
 
-- `deleteMessages(messageIds)`: Message deletion is not supported in ClickHouse
 - AI Observability (traces/spans): Not currently supported
 
 ## Data Types
--- a/stores/clickhouse/README.md
+++ b/stores/clickhouse/README.md
@@ -10,9 +10,9 @@ npm install @mastra/clickhouse
 
 ## Prerequisites
 
-- Clickhouse server (version 23.3 or higher recommended)
+- Clickhouse server (version 23.3 or higher required for delete operations; earlier versions may work for read/write operations)
   - Lightweight `DELETE FROM` requires ClickHouse 22.8+ with `allow_experimental_lightweight_delete = 1` (for 22.8–23.2), or 23.3+ where it is generally available.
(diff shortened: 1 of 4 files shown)


stores/clickhouse/README.md
```diff
@@ -10,7 +10,9 @@ npm install @mastra/clickhouse
 
 ## Prerequisites
 
-- Clickhouse server (version 21.12 or higher recommended)
+- Clickhouse server (version 23.3 or higher recommended)
+  - Lightweight `DELETE FROM` requires ClickHouse 22.8+ with `allow_experimental_lightweight_delete = 1` (for 22.8–23.2), or 23.3+ where it is generally available.
+  - The `deleteTask` and `deleteTasks` methods use `DELETE FROM` on the background tasks table — ensure your server supports lightweight delete before using those operations.
```
wardpeet:
```suggestion
  - The `deleteTask`, `deleteTasks`, and `deleteMessages` methods use `DELETE FROM` — ensure your server supports lightweight delete before using those operations.
```

PR #15915, opened 2026-04-29
feat(server): add GET /agents/summary that never invokes dynamic getters

wardpeet:
## Summary

Follow-up to #15911. Adds a new lean endpoint **`GET /agents/summary`** that returns only `{ id, name, description }` per agent. The endpoint **never invokes any per-request dynamic getter** (`getInstructions`, `getLLM`, `listTools`, `listWorkflows`, `listAgents`, `getDefaultOptions`, `getModelList`, `getWorkspace`, …), so it cannot 500 from a misconfigured `requestContext` — by construction, not by `try/catch`. It's the right endpoint for clients that only need a list of agent identities (e.g. populating a navigation list, an autocomplete, or a sub-agent picker) and want a hard guarantee that the response shape is independent of the request context.

The existing `GET /agents` route is **untouched** (response shape, query params, behavior) — all current consumers keep working.

## What's added

- **Server:** `GET /agents/summary` returning `Record<id, { id, name, description }>`. Reuses `agents:read` permission. Includes both code-defined and stored agents (same merge rule as `/agents` — code-defined wins).
- **Client SDK:** `client.listAgentsSummary()` returning `Record<string, { id; name; description? }>`.
- **Schema:** `agentSummarySchema` + `listAgentsSummaryResponseSchema` in `packages/server/src/server/schemas/agents.ts`.
- **Adapter registration** in `packages/server/src/server/server-adapter/routes/agents.ts`. Registered immediately after `LIST_AGENTS_ROUTE` and before `GET_AGENT_BY_ID_ROUTE` so `/agents/summary` resolves before `/agents/:agentId` regardless of the adapter's matching strategy.

## Why a new endpoint instead of a query param

We can't change the existing `/agents` shape without breaking the playground (which renders `instructions`, `provider`, `modelId` and the workflow/tool/agent counts in its list view). A separate, narrowly-typed endpoint keeps the contract crystal-clear: callers that don't want dynamic-getter risk get a guaranteed-safe payload; callers that need the rich shape keep using `/agents`.

## Test plan

- New tests in `packages/server/src/server/handlers/agent.test.ts`:
  - Returns only `{ id, name, description }` for healthy agents (no extra keys).
  - Throwing `getInstructions`/`getLLM`/`listTools` are **never invoked** (asserted via `vi.spyOn().not.toHaveBeenCalled()`).
  - Endpoint works with no `requestContext` on the handler context.
- New SDK test in `client-sdks/client-js/src/resources/agent.test.ts` verifying it hits `/agents/summary`.
- `pnpm test:server` — 1068 tests pass.
- `pnpm build:server` + `pnpm build:core` — clean.
- `pnpm --filter ./packages/server check:permissions` — passes (regenerated permissions).

## Studio impact

None. Studio's `useAgents()` continues to use `GET /agents`. A separate follow-up could switch the agents list page to `/agents/summary` + lazy detail fetches, but that's out of scope for this PR.

## ELI5

This PR adds a lightweight new endpoint (`GET /agents/summary`) that returns a simple list of agents with just their id, name, and description. Unlike the existing full agents endpoint, it avoids calling any complex per-request configuration logic, so it won't break when things are misconfigured.

## Changes Overview

**New Endpoint & SDK Support**
- Added `GET /agents/summary` server route that returns a record mapping agent IDs to summaries (`{ id, name, description? }`)
- Added `client.listAgentsSummary()` SDK method in the JavaScript client to consume the endpoint
- Route uses the `agents:read` permission (reuses existing permission)

**Implementation Details**
- The endpoint avoids invoking per-request dynamic getters (e.g., `getInstructions`, `getLLM`, `listTools`, `listWorkflows`, etc.) that could fail due to misconfigured `requestContext`
- Applies stored editor overrides to code-defined agents so name/description customizations are preserved
- Returns both code-defined and stored agents with the same merge rule as the existing `/agents` endpoint (code-defined takes precedence)
- Individual agent serialization failures are caught and logged without aborting the overall response

**Schema & Route Registration**
- Added `agentSummarySchema` and `listAgentsSummaryResponseSchema` Zod schemas for validation
- Registered `LIST_AGENTS_SUMMARY_ROUTE` immediately after `LIST_AGENTS_ROUTE` and before `GET_AGENT_BY_ID_ROUTE` to ensure correct route matching (`/agents/summary` resolves before `/agents/:agentId`)

**Permissions & Generated Files**
- Updated permission models to add `agents:create` and `agents:delete` patterns, and introduced `background-tasks` as a new resource

**Testing**
- Added server handler tests asserting the endpoint:
  - Returns only `{ id, name, description }` 
  - Never invokes dynamic getters (even when mocked to throw)
  - Works without a `requestContext`
- Added SDK test verifying the client correctly calls the `/api/agents/summary` endpoint with proper headers
- All existing tests pass (1068+ tests); builds remain clean

**No Impact on Existing Code**
- The existing `GET /agents` endpoint and Studio's `useAgents()` hook remain unchanged

### DIFF
--- a/.changeset/agents-list-resilient-instructions.md
+++ b/.changeset/agents-list-resilient-instructions.md
@@ -1,0 +1,5 @@
+---
+'@mastra/server': patch
+---
+
+Fix `GET /agents` returning 500 when an agent's dynamic `instructions` (or other dynamic config such as `getLLM`, `getModelList`, `getDefaultOptions`, `listTools`, sub-agent `listAgents`) throws under the active request context. Each per-agent dynamic getter is now isolated: failures are logged with the agent name and the agent is returned with safe defaults so the rest of the list still renders. The list handler also uses `Promise.allSettled` so a catastrophic failure in one agent's serialization can no longer reject the whole response, matching the existing behavior for stored agents.
--- a/.changeset/agents-summary-endpoint.md
+++ b/.changeset/agents-summary-endpoint.md
@@ -1,0 +1,10 @@
+---
+'@mastra/server': minor
+'@mastra/client-js': minor
+---
+
+Add a new `GET /agents/summary` endpoint and `client.listAgentsSummary()` SDK method that return a lean `{ id, name, description }` listing of agents.
+
+Unlike `GET /agents`, the new endpoint never invokes any per-agent dynamic getter (instructions, llm, tools, default options, model list, sub-agents, workflows, processors, workspace, …), so it cannot fail when a user-supplied dynamic-config callback throws under the active `requestContext`. Use this endpoint when you only need a list of agent identities (e.g. populating a navigation list) and want a hard guarantee that the response shape is independent of the request context.
+
+The existing `GET /agents` route and its response shape are unchanged.
--- a/client-sdks/client-js/src/client.ts
+++ b/client-sdks/client-js/src/client.ts
@@ -213,6 +213,18 @@ export class MastraClient extends BaseResource {
     return this.request(`/agents${queryString ? `?${queryString}` : ''}`);
   }
 
+  /**
+   * Retrieves a lean summary { id, name, description } for every agent.
+   *
+   * Unlike {@link listAgents}, this does not invoke any per-request dynamic
+   * getters (instructions, llm, tools, default options, etc.), so it cannot
+   * fail when a user-supplied dynamic-config callback throws under the
+   * active request context.
+   */
+  public listAgentsSummary(): Promise<Record<string, { id: string; name: string; description?: string }>> {
+    return this.request(`/agents/summary`);
+  }
+
   public listAgentsModelProviders(): Promise<ListAgentsModelProvidersResponse> {
     return this.request(`/agents/providers`);
   }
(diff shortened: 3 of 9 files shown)


TylerBarnes (changes requested):
From Mastra Code: The summary endpoint shape and “don’t invoke dynamic getters” approach look good, and the focused tests/checks pass locally.

One blocker: the latest commit says `/agents/summary` applies stored overrides so code-defined agent summaries reflect Studio-edited name/description, but `editor.agent.applyStoredOverrides()` currently does not apply `name` or `description` — it only applies instructions/tools. So this still appears to return the code-defined `agent.name` / `getDescription()` for code-defined agents even when a stored config has edited summary fields.

Could you add a regression test for a code-defined agent with stored name/description overrides and either:

1. update `applyStoredOverrides()` to apply those static fields safely, or
2. have the summary endpoint read those static fields from the resolved stored config directly without invoking dynamic getters?

The endpoint should keep the invariant that it doesn’t call dynamic getters, but the stored override behavior needs to be real/tested rather than just calling `applyStoredOverrides()`.

abhiaiyer91:
closing

Issue #15908, opened 2026-04-29
DX: Prevent duplicate skill context when SkillSearchProcessor-loaded skills are also read via workspace file tools


panliji:
### Description

We are using `SkillSearchProcessor` as the primary skill discovery/loading mechanism for a production Mastra agent.

The expected flow works well:

1. The agent calls `search_skills`.
2. The agent calls `load_skill`.
3. The loaded skill instructions are injected as system messages on subsequent steps.

However, we observed a context duplication pattern in production:

After a skill is loaded through `SkillSearchProcessor`, the agent may still call workspace file tools such as `mastra_workspace_list_files` / `mastra_workspace_read_file` to read the same `SKILL.md` file from the workspace, for example under `./skills/.../SKILL.md` or `./default-skills/.../SKILL.md`.

This means the same skill content can enter the model context through multiple routes:

- as system messages injected by `SkillSearchProcessor`
- as `load_skill` tool transcript / result
- as normal workspace file tool result from reading `SKILL.md`

This increases input tokens and can make the source of instruction authority less clear. It also interacts poorly with memory/history replay if large skill file reads or skill tool results are persisted and later recalled.

### Why this matters

For normal task execution, once `load_skill` has succeeded, the model already has the skill instructions. Reading the same `SKILL.md` again through workspace file tools is usually redundant.

The extra content can:

- increase token usage significantly, especially with long skills
- reduce available context for user input and business data
- introduce duplicate or stale instruction copies into future steps
- make tool result replay / memory persistence more expensive and harder to reason about

At the same time, reading `SKILL.md` is still legitimate in explicit file-oriented workflows, such as when the user asks to inspect, edit, compare, or debug a skill file.

### Question

Is there an official Mastra-recommended pattern for avoiding duplicate skill context when using `SkillSearchProcessor` together with workspace file tools?

In particular:

1. Should `SkillSearchProcessor` expose loaded skill state in `requestContext` or processor state so applications can identify that a `SKILL.md` has already been injected?
2. Is there an intended way for workspace file tools to suppress or summarize reads of already-loaded `SKILL.md` files during normal agent execution?
3. Should `load_skill` / `skill_read` / workspace reads of `SKILL.md` be marked as non-replayable or filtered from memory by default?
4. Is `ToolCallFilter` / `TokenLimiterProcessor` the recommended approach here, or should this be handled with a custom `processInputStep()` processor?
[...]

PR #15916, opened 2026-04-29
fix(core): treat skill search as on-demand skill discovery

mbenhamd:
Refs #15908.

When `SkillSearchProcessor` is configured directly as an input processor, agents now treat it as an on-demand skill discovery path instead of also adding the eager `SkillsProcessor` by default.

Before this change, `SkillSearchProcessor` could coexist with auto-added eager skill context, so the model could see `<available_skills>` and overlapping `skill` / `skill_search` tools while also having `search_skills` / `load_skill`.

After this change, `SkillSearchProcessor` owns the on-demand discovery flow. The agent skips the auto-added eager `SkillsProcessor` in that mode, hides the overlapping skill tools, and keeps `skill_read` available for supporting skill files.

This also threads per-call input processors through the legacy generate path so `SkillSearchProcessor` can add its tools there without dropping existing custom tools. Explicitly configuring `SkillsProcessor` still opts back into eager skill activation.

This should reduce one major source of duplicate skill context. It does not block workspace file tools from reading an already-loaded `SKILL.md`, and it does not change memory or tool-result replay behavior; the docs call out the workspace-file caveat and recommend reserving workspace reads for explicit file inspection/editing workflows.

Verified with:

```sh
npx vitest run packages/core/src/agent/__tests__/skills-with-custom-processors.test.ts
pnpm --filter ./packages/core check
pnpm --dir docs lint:remark
git diff --check
```

ELI5 (Explain Like I'm 5)

Some agents learn about skills all at once; others look for skills only when needed. This PR makes the "look-when-needed" processor (SkillSearchProcessor) take ownership when configured so the agent doesn't show duplicate or confusing skill tools and instead exposes search/load/read tools for on-demand discovery.

Overview

When SkillSearchProcessor is configured as an input processor (including when wrapped inside a processor workflow), the agent treats it as the authoritative on-demand skill-discovery flow and suppresses the auto-added eager SkillsProcessor unless SkillsProcessor is explicitly configured. In on-demand mode the agent exposes search_skills and load_skill (and keeps skill_read for reading supporting files), hides eager skill tools (skill, skill_search), and updates prompts to omit <available_skills> and instruct callers to use search_skills/load_skill. Workspace file tools remain able to read SKILL.md during explicit inspection/edit workflows. Memory and tool-result replay semantics are unchanged.

Key Changes

- Processor metadata
  - Added readonly providesSkillDiscovery?: 'on-demand' to Processor interface.
  - SkillSearchProcessor now sets providesSkillDiscovery = 'on-demand'.

- Agent/tool provisioning
  - Agent inspects top-level processors and nested workflow steps to detect eager (skills-processor) and on-demand discovery providers.
  - When on-demand discovery exists and no explicit eager SkillsProcessor is present, agent suppresses eager skill tools (skill, skill_search) and surfaces search_skills, load_skill, and skill_read.
  - Prompt generation omits <available_skills> and adds guidance to call search_skills in on-demand mode.
  - Suppression logic handles workflow-wrapped processors and avoids false positives from id collision or misinterpreting a steps field as a workflow.

- SkillSearchProcessor behavior
  - Declares providesSkillDiscovery and owns on-demand flow.
  - Validates search_skills.query by trimming and rejecting blank/whitespace-only queries.
  - Maintains thread-scoped loaded-skill state with TTL cleanup and exposes clearState/clearAllState APIs.

- Legacy & per-call support
  - Threaded per-call inputProcessors through legacy pathways:
    - AgentLegacyCapabilities.convertTools now accepts inputProcessors?: InputProcessorOrWorkflow[].
    - AgentLegacyHandler.__primitive/__runProcessInputStep accept inputProcessorOverrides and extended execution context (tools, runId, threadId, resourceId, outputWriter, autoResumeSuspendedTools, backgroundTaskEnabled).
    - __runProcessInputStep may return overridden tools; AgentLegacyHandler updates converted tools accordingly.
  - Tool conversion (convertTools) and prepare-tools workflow forward options.inputProcessors so per-call processors can add/override tools without dropping existing custom tools.

- Workflows & processor runner
  - createStep/createStepFromProcessor propagation updated to copy providesSkillDiscovery onto workflow steps so on-demand metadata survives workflow wrapping.
  - Processor skill-discovery metadata preserved across workflow steps and normalized during conversion.

- Tests & verification
  - Extensive test coverage added/updated to assert:
    - On-demand mode across Agent constructor inputProcessors, per-call generate/stream inputProcessors, workflow-wrapped processors, and legacy generateLegacy.
    - Presence of search_skills/load_skill/skill_read and absence of eager skill tools in on-demand mode.
    - Prompt text changes (no <available_skills>, includes guidance to call search_skills).
    - Handling of processors with steps fields, id collisions, and per-call stream parity.
    - Legacy generate path parity (MockLanguageModelV1/V2 capture prompt and tools).

- Docs & minor fixes
  - SkillSearchProcessor reference updated to document providesSkillDiscovery: 'on-demand', default tool exposure differences vs SkillsProcessor, guidance on workspace file tool usage, and thread-state TTL behavior.
  - Minor typing simplifications and tool-normalization fixes; deferred metadata caching optimization for a follow-up.

Behavioral notes

- Explicitly configuring SkillsProcessor still opts into eager skill activation even when SkillSearchProcessor is present.
- Workspace file tools remain usable for explicit inspection/edit flows; SkillSearchProcessor does not block reading SKILL.md when explicitly invoked.
- No changes to memory semantics or tool-result replay; per-call processors are threaded through both legacy and new generation paths so custom tools are preserved.


packages/core/src/agent/agent.ts
```diff
@@ -141,6 +141,21 @@ function resolveMaybePromise<T, R = void>(value: T | Promise<T> | PromiseLike<T>
...
+  return processors.some(processor => !isProcessorWorkflow(processor) && predicate(processor));
+}
+
+function hasEagerSkillsProcessor(processors: InputProcessorOrWorkflow[]): boolean {
+  return hasConfiguredProcessor(processors, processor => processor.id === 'skills-processor');
+}
+
+function hasOnDemandSkillDiscoveryProcessor(processors: InputProcessorOrWorkflow[]): boolean {
+  return hasConfiguredProcessor(processors, processor => processor.providesSkillDiscovery === 'on-demand');
+}
```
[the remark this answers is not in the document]
mbenhamd: Addressed in 314c30201e. Workflow-wrapped processor steps are now inspected via processor:* step ids, with regression coverage for a workflow-wrapped SkillSearchProcessor and for the explicit SkillsProcessor + SkillSearchProcessor workflow escape hatch.

docs/src/content/en/reference/processors/skill-search-processor.mdx
```diff
@@ -9,6 +9,13 @@ packages:
 
 The `SkillSearchProcessor` is an **input processor** that enables on-demand skill discovery and loading. Instead of injecting all skill metadata into the system prompt upfront (like `SkillsProcessor`), it gives the agent two meta-tools (`search_skills` and `load_skill`) that let it find and load skills on demand. This reduces context token usage when workspaces have many skills.
 
+When you attach `SkillSearchProcessor` to an agent with a skill-enabled `Workspace`, the agent treats skills as on-demand:
+
+- It does not auto-add the default eager `SkillsProcessor`.
+- It exposes `search_skills` and `load_skill` for discovery and instruction loading.
+- It does not expose the overlapping `skill` and `skill_search` tools.
+- It still exposes `skill_read` so the agent can read supporting skill files such as references, scripts, and assets.
```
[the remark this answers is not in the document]
mbenhamd: Addressed in 314c30201e by qualifying this as the default behavior when only SkillSearchProcessor is attached and documenting the explicit SkillsProcessor escape hatch.

mbenhamd:
Addressed CodeRabbit feedback in `314c30201e`:

- Documented the explicit `SkillsProcessor` escape hatch in the `SkillSearchProcessor` reference.
- Detect workflow-wrapped skill processors so `SkillSearchProcessor` still suppresses eager skill mode when wrapped in a processor workflow. Added regression coverage for both workflow-wrapped `SkillSearchProcessor` and explicit `SkillsProcessor` + `SkillSearchProcessor` workflows.
- Reject blank/whitespace `search_skills` queries.

I left the `skills.list()` metadata lookup optimization as-is for this PR. The current path uses `list()` to fetch descriptions without loading full skill instructions via `get()`; adding a metadata cache would need refresh/maybeRefresh invalidation semantics and seems better as a separate performance follow-up.

mbenhamd:
Addressed the latest CodeRabbit nitpicks in `3836794fcf`:

- Added per-call `stream()` parity coverage for `SkillSearchProcessor` on-demand mode.
- Reworded the `SkillSearchProcessor` reference bullets to avoid repeated openings.

Verified again with:

```sh
npx vitest run packages/core/src/agent/__tests__/skills-with-custom-processors.test.ts
pnpm --filter ./packages/core check
pnpm --dir docs lint:remark
git diff --check
```

mbenhamd:
@coderabbitai review

packages/core/src/agent/agent.ts
```diff
@@ -2826,8 +2914,42 @@ export class Agent<
...
+                  workspace,
+                },
+                undefined,
+              );
+            } else {
+              convertedTools[name] = tool as CoreTool;
+            }
+          }
+
+          nextTools = convertedTools;
```
NikAiyer, line 2962: Do we want to fuilly replace the tools here, or merge them together like so:
```suggestion
          nextTools = { ...nextTools, ...convertedTools };
```

packages/core/src/processors/processors/skill-search.ts
```diff
@@ -88,6 +88,7 @@ export class SkillSearchProcessor implements Processor<'skill-search'> {
   readonly id = 'skill-search';
   readonly name = 'Skill Search Processor';
   readonly description = 'Enables on-demand skill discovery and loading via search';
+  readonly providesSkillDiscovery = 'on-demand' as const;
```
NikAiyer: DO we need `as const` here?

packages/core/src/agent/agent.ts
```diff
@@ -4490,6 +4615,7 @@ export class Agent<
       mastraProxy,
       autoResumeSuspendedTools,
       backgroundTaskEnabled,
+      inputProcessors,
```
NikAiyer: Do we need to pass down `inputProcessors` like this?

If we're only using it to determine these variables:

```
    const configuredInputProcessors = inputProcessors ?? (await this.listConfiguredInputProcessors(requestContext));
    const hasOnDemandProcessor = hasOnDemandSkillDiscoveryProcessor(configuredInputProcessors);
    const hasSkillsProcessor = hasEagerSkillsProcessor(configuredInputProcessors);
```

we can probably just pass down a boolean flag instead. Though it depends on what else we're using `inputProcessors` for when we're passing it down.

packages/core/src/agent/agent.ts
```diff
@@ -2826,8 +2914,42 @@ export class Agent<
...
+                  ...observabilityContext,
+                  model: await this.getModel({ requestContext }),
+                  tracingPolicy: this.#options?.tracingPolicy,
+                  requireApproval: (tool as any).requireApproval,
+                  backgroundConfig: (tool as any).background,
+                  workspace,
+                },
+                undefined,
+              );
+            } else {
```
NikAiyer, line 2957: I think we're missing these fields:
```
mastra, memory, outputWriter, autoResumeSuspendedTools, and backgroundTaskEnabled
```

from this makeCoreTool callsite that is present in other callsites.

packages/core/src/agent/agent.ts
```diff
@@ -141,6 +141,67 @@ function resolveMaybePromise<T, R = void>(value: T | Promise<T> | PromiseLike<T>
...
+    if (!isWorkflowLike || workflowSteps.length === 0) {
+      const processorId =
+        typeof (processor as Processor).id === 'string' && (processor as Processor).id.startsWith('processor:')
+          ? (processor as Processor).id.slice('processor:'.length)
+          : (processor as Processor).id;
+      return predicate({
+        ...(processor as Processor),
+        id: processorId,
+        providesSkillDiscovery:
+          (processor as Processor).providesSkillDiscovery ?? (processorId === 'skill-search' ? 'on-demand' : undefined),
```
NikAiyer: This might cause an edge case if a user creates their own processor with the name skill-search. I wonder if there's a better way to account for this.

packages/core/src/agent/agent.ts
```diff
@@ -141,6 +141,67 @@ function resolveMaybePromise<T, R = void>(value: T | Promise<T> | PromiseLike<T>
...
+
+      const stepId = typeof (step as { id?: unknown }).id === 'string' ? (step as { id: string }).id : undefined;
+      if (!stepId?.startsWith('processor:')) {
+        return false;
+      }
+
+      const processorId = stepId.slice('processor:'.length);
+      return predicate({
+        id: processorId,
+        providesSkillDiscovery: processorId === 'skill-search' ? 'on-demand' : undefined,
```
NikAiyer: Same for this location as well.

NikAiyer (commented):
Hey this looks good to me, just had some suggestions for some tweaks we could make here.

packages/core/src/agent/agent.ts
NikAiyer: Something like:

```
    const skillTools = await this.listSkillTools({
      // ...existing params...
      suppressEagerSkillTools: hasOnDemandProcessor && ! hasSkillsProcessor,
    });
    ```

File: packages/core/src/agent/agent.ts (hunk 7 above)
Comment (mbenhamd) [reply packages/core/src/agent/agent.ts]: Addressed in `3c915c3f21`: on-demand mode now relies on explicit `providesSkillDiscovery` metadata instead of inferring it from the `skill-search` id, and there is a regression test for a user processor with a colliding id.

File: packages/core/src/agent/agent.ts (hunk 3 above)
Comment (mbenhamd) [reply packages/core/src/agent/agent.ts:2962]: Addressed in `3c915c3f21`: processor-returned tools are now merged over the existing tool set instead of replacing it wholesale.

File: packages/core/src/agent/agent.ts (hunk 6 above)
Comment (mbenhamd) [reply packages/core/src/agent/agent.ts:2957]: Addressed in `3c915c3f21`: the `makeCoreTool` call now includes the same runtime context fields used by the other conversion paths, including `mastra`, `memory`, `outputWriter`, `autoResumeSuspendedTools`, and `backgroundTaskEnabled` where available.

File: packages/core/src/agent/agent.ts (hunk 8 above)
Comment (mbenhamd) [reply packages/core/src/agent/agent.ts]: Addressed in `3c915c3f21` as well: workflow-created processor steps now preserve explicit `providesSkillDiscovery` metadata, and the workflow detection reads that metadata instead of inferring from the step id.

File: packages/core/src/processors/processors/skill-search.ts (hunk 4 above)
Comment (mbenhamd) [reply packages/core/src/processors/processors/skill-search.ts]: Addressed in `55391a0ec7`: removed the redundant `as const` and typed `providesSkillDiscovery` against `Processor["providesSkillDiscovery"]` directly.

File: packages/core/src/agent/agent.ts (hunk 5 above)
Comment (mbenhamd) [reply packages/core/src/agent/agent.ts]: Addressed in `41b4a2def2`: `convertTools()` now derives `suppressEagerSkillTools` and passes that boolean to `listSkillTools()`, so `listSkillTools()` no longer receives the full `inputProcessors` array.

File: packages/core/src/agent/agent.ts [hunk 9]
```diff
@@ -2826,8 +2910,42 @@ export class Agent<
...
+                  tracingPolicy: this.#options?.tracingPolicy,
+                  requireApproval: (tool as any).requireApproval,
+                  backgroundConfig: (tool as any).background,
+                  workspace,
+                },
+                undefined,
+              );
+            } else {
+              convertedTools[name] = tool as CoreTool;
+            }
```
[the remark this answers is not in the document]
mbenhamd: Fixed in `3c915c3f21`: provider-shaped tools returned from `processInputStep` are now normalized via `makeCoreTool` rather than being cast directly to `CoreTool`.

packages/core/src/agent/agent.ts
```diff
@@ -141,6 +141,67 @@ function resolveMaybePromise<T, R = void>(value: T | Promise<T> | PromiseLike<T>
...
+      const stepId = typeof (step as { id?: unknown }).id === 'string' ? (step as { id: string }).id : undefined;
+      if (!stepId?.startsWith('processor:')) {
+        return false;
+      }
+
+      const processorId = stepId.slice('processor:'.length);
+      return predicate({
+        id: processorId,
+        providesSkillDiscovery: processorId === 'skill-search' ? 'on-demand' : undefined,
+      } as Processor);
```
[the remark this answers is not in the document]
mbenhamd: Fixed in `3c915c3f21`: workflow steps now carry explicit `providesSkillDiscovery` metadata from the wrapped processor, and detection reads that metadata instead of hard-coding `skill-search`.

packages/core/src/agent/agent.ts
```diff
@@ -2826,8 +2916,51 @@ export class Agent<
...
+                autoResumeSuspendedTools,
+                backgroundTaskEnabled,
+              );
+            } else {
+              convertedTools[name] = tool as CoreTool;
+            }
+          }
+
+          nextTools = { ...(nextTools ?? {}), ...convertedTools };
+        }
```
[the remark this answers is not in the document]
mbenhamd: Fixed in `4c5155df9f`: after normalizing returned tools, `processInputStep` tool output is again treated as the authoritative tool set, so intentionally removed tools are not reintroduced.

packages/core/src/agent/agent.ts
```diff
@@ -141,6 +141,67 @@ function resolveMaybePromise<T, R = void>(value: T | Promise<T> | PromiseLike<T>
...
+      ...Object.values(maybeWorkflow.steps ?? {}),
+      ...(maybeWorkflow.stepGraph ?? []).flatMap(entry => {
+        if (entry.type === 'step') {
+          return entry.step ? [entry.step] : [];
+        }
+        return entry.steps?.map(stepEntry => stepEntry.step).filter(Boolean) ?? [];
+      }),
+    ];
+
+    if (!isWorkflowLike || workflowSteps.length === 0) {
```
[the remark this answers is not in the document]
mbenhamd: Fixed in `4c5155df9f`: `hasConfiguredProcessor()` now only treats objects passing `isProcessorWorkflow()` as workflows, and there is a regression test for an on-demand processor that owns a regular `steps` field.

mbenhamd:
@coderabbitai review

abhiaiyer91 (approved):

Issue #9005, opened 2025-10-19
[BUG] Memory incompatibility with GPT-5 mini reasoning models causes "required reasoning item" errors


realaman90:
### Describe the Bug

When using GPT-5 mini (OpenAI's reasoning model) with Mastra agents that have tools/function calls enabled, the agent fails on subsequent messages with conversation history due to missing reasoning items.
 Error Message

```
Item 'fc_...' of type 'function_call' was provided without its required 'reasoning' item: 'rs_...'.
```

Full error:

```javascript
Error [AI_APICallError]: Item 'fc_08bb11e324d4e4680068f4c72182808193b7ba870871731d72' of type 'function_call' was provided without its required 'reasoning' item: 'rs_08bb11e324d4e4680068f4c71de92c81939e298c28f83703b2'.

cause: undefined,
url: 'https://api.openai.com/v1/responses',
statusCode: 400,
responseBody: {
  "error": {
    "message": "Item 'fc_...' of type 'function_call' was provided without its required 'reasoning' item: 'rs_...'.",
    "type": "invalid_request_error",
    "param": "input",
    "code": null
  }
}
```
## Root Cause

GPT-5 mini uses OpenAI's **Responses API** (not Chat Completions API) and generates reasoning items for every action, including tool/function calls. The Responses API requires that when a `function_call` item is included in conversation history, its associated `reasoning` item must also be present.

**Current behavior**: Mastra's Memory system stores messages and function calls but doesn't preserve the associated reasoning items, causing the API to reject the conversation history on subsequent turns.

---

### 1. Create an agent with GPT-5 mini and tools

```typescript
import { openai } from "@ai-sdk/openai";
import { Agent } from "@mastra/core/agent";
import { Memory } from "@mastra/memory";
import { createTool } from "@mastra/core/tools";
import { z } from "zod";

const memory = new Memory({
  storage: sharedStorage,
  options: {
    lastMessages: 10,
  },
});

export const testAgent = new Agent({
  name: "Test Agent",
  instructions: "You are a helpful assistant.",
  model: openai("gpt-5-mini"),
  memory: memory,
  tools: {
    testTool: createTool({
      id: "testTool",
      description: "A test tool that returns a message",
      inputSchema: z.object({
        message: z.string(),
      }),
      execute: async ({ context }) => {
        return { result: `Tool executed: ${context.message}` };
      },
    }),
  },
});
```

### 2. Make a conversation with tool calls

```typescript
// First message - works fine
[...]

PR #15934, opened 2026-04-29
fix(core): preserve semantic stream order in buildMessagesFromChunks (#15914)

### LINKED ISSUE #15914: [BUG] buildMessagesFromChunk function uses streaming order while some internal codes expect semantic order
### Describe the Bug

Background: Currently I'm using mastracode with local ollama setup (model: qwen3.6:35b-a3b-q8_0), the conversation frequently get cutoff due to LLM returning a response without tool call, which will stop the agentic loop. Later, I found that the input prompt from mastra to LLM contains history messages in weird order (tool call before text before reasoning), the culprit causing LLM to malfunction and to start giving back responses following this input pattern, where text response comes last without tool call.

After some digging, and through debugger, I can see the chunks streaming back from LLM is in the following order:
```
[response-metadata]
[text-start]
[reasoning-start]
[reasoning-delta]
...
[reasoning-delta]
[text-delta]
...
[text-delta]
[tool-call-input-streaming-start]
[tool-call-delta]
[tool-call-input-streaming-end]
[tool-call]
[tool-call-input-streaming-start]
[tool-call-delta]
[tool-call-input-streaming-end]
[tool-call]
[text-end]
[reasoning-end]
[finish]
```

As a result, the messages built from the collected chunks become:
```
{type: 'tool-invocation', ...},
{type: 'tool-invocation', ...},
{type: 'step-start'},
{type: 'text', ...},
{type: 'reasoning', ...}
```

Since `buildMessagesFromChunk` intends to keep messages in the streaming order, not enforcing any semantic order, we can consider this as expected behavior? However, taking a look at the LLM response, even though the end span is emitted in tool-call - text - reasoning order, the bulk of the parts is in reasoning - text - tool-call order, thus we can argue that the end span does not necessarily reflect the streaming order of the parts.

This becomes a problem when mastra build LLM promt for next step in the loop, which internally calls `convertAIV5UIToModelMessages`, and adds 'step-start' to the part sequence before converting, and since the part sequence is tool-call -> text -> reasoning, `addStartStepPartsForAIV5` erroneously add 'step-start' in between message parts from the same turn.

In addition, I noticed that the messages already contain 'step-start' parts added in LLM Execution Step when `currentIteration > 1`, so probably the `addStartStepPartsForAIV5` is redundant?

By removing `addStartStepPartsForAIV5` line from `convertAIV5UIToModelMessages` function, I'm able to improve my agent's loop continuation from annoying cutoffs to almost fine, but the illogical ordering of message parts still causes cutoffs once in a while, and some other cosmetic issues in mastracode (like reasoning showing below the text response in TUI), I believe in most places we just assume the ordering of messages is reasoning - text - tool calls.

I think if we intend to keep `buildMessagesFromChunk` building messages in streaming order, we should at least warn people not to assume the orders of message parts, but anyway, to use end span to represent the streaming order of parts doesn't look right in this case. I'm not aware of any specs requiring the LLM to stream the chunks in certain order, so `buildMessagesFromChunk` should handle this case.

One possible fix is to use the last-seen delta as the ordering of a part (whenever we see a new delta coming in, we update the order of the part to be at the end of the sequence) for maximum compatibilities with the current behavior, with fallback to the current logic of `buildMessagesFromChunk` for spans without delta. But naturally, I think the ordering shall be based on the first-seen delta, with fallback to the appearing order for tool-call, source, file spans.

### Steps To Reproduce

Add a test in packages/core/src/loop/workflows/agentic-execution/build-messages-from-chunks.test.ts:
```
it('should produce reasoning → text → tool-call 1 → tool-call 2 when reasoning-end arrives after text-end', () => {
  const result = parts([
    { type: 'response-metadata', payload: { id: 'rm1', modelId: 'test-model' } },
    { type: 'text-start', payload: { id: 't1' } },
    { type: 'reasoning-start', payload: { id: 'r1' } },
    { type: 'reasoning-delta', payload: { id: 'r1', text: 'Thinking...' } },
    { type: 'text-delta', payload: { id: 't1', text: 'Hello' } },
    { type: 'tool-call-input-streaming-start', payload: { toolCallId: 'tc1', toolName: 'myTool', args: {} } },
    { type: 'tool-call-delta', payload: { toolCallId: 'tc1', argsTextDelta: "{'q':'first'}" } },
    { type: 'tool-call-input-streaming-end', payload: { toolCallId: 'tc1' } },
    {
      type: 'tool-call',
      payload: { toolCallId: 'tc1', toolName: 'myTool', args: { q: 'first' } },
    },
    { type: 'tool-call-input-streaming-start', payload: { toolCallId: 'tc2', toolName: 'myTool', args: {} } },
    { type: 'tool-call-delta', payload: { toolCallId: 'tc2', argsTextDelta: "{'q':'second'}" } },
    { type: 'tool-call-input-streaming-end', payload: { toolCallId: 'tc2' } },
    {
      type: 'tool-call',
      payload: { toolCallId: 'tc2', toolName: 'myTool', args: { q: 'second' } },
    },
    { type: 'text-end', payload: { id: 't1' } },
    { type: 'reasoning-end', payload: { id: 'r1' } },
    { type: 'finish', payload: { finishReason: 'stop', usage: {} } },
  ]);

  expect(result).toHaveLength(4);
  expect(result.map((p: any) => p.type)).toEqual(['reasoning', 'text', 'tool-invocation', 'tool-invocation']);

  expect(result[0]).toMatchObject({ type: 'reasoning', details: [{ type: 'text', text: 'Thinking...' }] });
  expect(result[1]).toMatchObject({ type: 'text', text: 'Hello' });
  expect(result[2]).toMatchObject({ toolInvocation: { state: 'call', toolCallId: 'tc1', toolName: 'myTool', args: { q: 'first' } } });
  expect(result[3]).toMatchObject({ toolInvocation: { state: 'call', toolCallId: 'tc2', toolName: 'myTool', args: { q: 'second' } } });
});
```

then run
```
pnpm --filter ./packages/core test -- src/loop/workflows/agentic-execution/build-messages-from-chunks.test.ts
```

https://github.com/jmzhang/mastra/tree/messages-build-order-issue

### Expected Behavior

The messages built from the chunks shall be:
```
{type: 'reasoning', ...},
{type: 'text', ...},
{type: 'tool-invocation', ...},
{type: 'tool-invocation', ...}
```

### Environment Information

```block
- LLM Provider: Ollama 0.21.2, qwen3.6:35b-a3b, ollama-ai-provider-v2 version 3.5.0
```

### Verification

franciseliang99-dot:
## Summary

`buildMessagesFromChunks` previously emitted each text/reasoning span at its **end-event**, so the resulting `parts` array followed end-event timing rather than first-content arrival. When a stream interleaves spans —

- `text-start` before `reasoning-start`,
- `reasoning-delta` arrives before `text-delta`,
- `reasoning-end` arrives after `text-end`

— the parts came out as `[tool, tool, text, reasoning]` even though the semantic order is reasoning → text → tool-call → tool-call. Downstream `convertAIV5UIToModelMessages` + `addStartStepPartsForAIV5` then injected `step-start` markers in the wrong places, breaking the agent loop continuation described in the issue.

Closes #15914.

## Fix

Each text and reasoning span now reserves its slot in `parts` at **first-seen-delta**, then fills the slot at end-event. Redacted reasoning (which never receives a delta) reserves its slot at `reasoning-start` instead.

```ts
case 'text-delta': {
  // ...
  if (span.placeholder === null) {
    span.placeholder = { type: 'text', text: '' } as MastraMessagePart;
    parts.push(span.placeholder);
  }
  span.deltas.push(p.text);
}

case 'text-end': {
  // ...
  if (span.placeholder) {
    if (text.length > 0) {
      (span.placeholder as { text: string }).text = text;
      // ...
    } else {
      // Drop empty placeholder so "skip empty text spans" behavior is preserved
      const idx = parts.indexOf(span.placeholder);
      if (idx >= 0) parts.splice(idx, 1);
    }
  }
  // ...
}
```

The same pattern applies to `reasoning-delta` / `reasoning-end` and to the unclosed-span flush loops at the end of `buildMessagesFromChunks`.

## Behaviour preserved

- **"Always emit reasoning even if empty"** (#9005) — `reasoning-end`'s no-delta branch falls back to `parts.push` so empty reasoning still emits.
- **"Skip empty text spans"** — empty text drops its placeholder rather than emitting an empty text part.
- **Redacted reasoning** — placeholder reserved at `reasoning-start` (no delta arrives for redacted spans).

## Regression test

One test in `build-messages-from-chunks.test.ts` mirroring the issue's chunk sequence — pure chunk-array assertion, no model needed. Verifies `[reasoning, text, tool-invocation, tool-invocation]` on the interleaved-spans + reasoning-end-after-text-end stream from the issue body.

## Local verification

Did NOT run `pnpm typecheck` / `vitest` locally (monorepo install absent — same posture as #15904). Hand-verified:

- New test placed inside `describe('buildMessagesFromChunks', ...)` before "Empty stream" section to match local convention.
- Existing `should preserve correct order: reasoning, text, tool-call` (line 307) still expects the same output — its case has end-events arriving in semantic order, so placeholder mutation produces an identical result to the pre-fix behavior.

Happy to address CodeRabbit / typecheck / test feedback.

## ELI5

When the AI streamed pieces of a message out of order, the code ordered parts by when they finished arriving, which scrambled the intended sequence. This PR fixes that by reserving slots when a part first starts streaming so final parts appear in the order their content actually began.

## What's Changed

- buildMessagesFromChunks now reserves placeholders for text and reasoning spans at first-seen-delta (or at reasoning-start for redacted reasoning) and fills or removes those placeholders on end/flush, instead of only pushing parts at end events.
- Text spans: placeholder inserted on first text-delta, filled at text-end, removed if text is empty (preserves skipping empty text).
- Reasoning spans: placeholder reserved at reasoning-start for redacted reasoning or at first reasoning-delta for non-redacted; reasoning-end/flush populates the placeholder or pushes via fallback for empty-but-required reasoning.
- Unclosed-span flush logic uses the same placeholder pattern.
- Regression test added to assert semantic ordering for an interleaved sequence reproducing #15914.

## Why It Matters

Fixes semantic ordering so downstream consumers (e.g., convertAIV5UIToModelMessages + addStartStepPartsForAIV5) can rely on first-seen ordering when assembling prompts and step markers, preventing malformed prompts and broken agent loops.

## Tests

- New/updated tests in build-messages-from-chunks.test.ts:
  - Interleaved text spans now assert first-seen-delta ordering.
  - New regression test reproduces reasoning-delta arriving before text-delta but reasoning-end after text-end and asserts parts order [reasoning, text, tool-invocation, tool-invocation].

## Files Modified

- .changeset/fix-build-messages-semantic-order.md — release note
- packages/core/src/loop/workflows/agentic-execution/build-messages-from-chunks.ts — implementation (+105/-37)
- packages/core/src/loop/workflows/agentic-execution/build-messages-from-chunks.test.ts — tests (+~36)

## Notes for reviewers

- The author did not run monorepo typecheck/tests locally and requested CI/typecheck feedback. This change is focused on ordering semantics and preserves previous behaviors (empty reasoning still emits; empty text skipped).
- Closes #15914.

### DIFF
--- a/.changeset/fix-build-messages-semantic-order.md
+++ b/.changeset/fix-build-messages-semantic-order.md
@@ -1,0 +1,5 @@
+---
+'@mastra/core': patch
+---
+
+Fix `buildMessagesFromChunks` to preserve semantic stream order. Previously the `parts` array followed end-event timing, so when `text-end` arrived before `reasoning-end` the reasoning part was emitted after the text even though its first delta arrived first — confusing downstream prompt assembly. Each text and reasoning span now reserves its slot at first-seen-delta (or at start for redacted reasoning) and fills the slot at end. Closes #15914.
--- a/.changeset/fix-build-messages-semantic-order.md
+++ b/.changeset/fix-build-messages-semantic-order.md
@@ -2,4 +2,4 @@
 '@mastra/core': patch
 ---
 
-Fix `buildMessagesFromChunks` to preserve semantic stream order. Previously the `parts` array followed end-event timing, so when `text-end` arrived before `reasoning-end` the reasoning part was emitted after the text even though its first delta arrived first — confusing downstream prompt assembly. Each text and reasoning span now reserves its slot at first-seen-delta (or at start for redacted reasoning) and fills the slot at end. Closes #15914.
+Assistant message `parts` now preserve the correct order when reasoning, text, and tool chunks interleave. Previously, parts followed the order of end events, so a span whose end arrived first appeared earlier in the message even when its content actually started streaming later — for example a tool call could be emitted before reasoning that began first. Order now tracks when content actually started arriving. Closes #15914.
--- a/packages/core/src/loop/workflows/agentic-execution/build-messages-from-chunks.test.ts
+++ b/packages/core/src/loop/workflows/agentic-execution/build-messages-from-chunks.test.ts
@@ -322,6 +322,42 @@ describe('buildMessagesFromChunks', () => {
     expect(types).toEqual(['reasoning', 'text', 'tool-invocation']);
   });
 
+  it('should produce reasoning → text → tool-call when reasoning-end arrives after text-end (#15914)', () => {
+    // Regression for #15914: when spans are interleaved (text-start before reasoning-start
+    // but reasoning-delta arriving before text-delta, and reasoning-end arriving after
+    // text-end), the previous emit-on-end pattern produced [tool, tool, text, reasoning].
+    // The fix reserves each span's slot at first-seen-delta so order tracks semantic arrival.
+    const result = parts([
+      { type: 'response-metadata', payload: { id: 'rm1', modelId: 'test-model' } },
+      { type: 'text-start', payload: { id: 't1' } },
+      { type: 'reasoning-start', payload: { id: 'r1' } },
+      { type: 'reasoning-delta', payload: { id: 'r1', text: 'Thinking...' } },
+      { type: 'text-delta', payload: { id: 't1', text: 'Hello' } },
+      { type: 'tool-call-input-streaming-start', payload: { toolCallId: 'tc1', toolName: 'myTool', args: {} } },
+      { type: 'tool-call-delta', payload: { toolCallId: 'tc1', argsTextDelta: "{'q':'first'}" } },
+      { type: 'tool-call-input-streaming-end', payload: { toolCallId: 'tc1' } },
+      { type: 'tool-call', payload: { toolCallId: 'tc1', toolName: 'myTool', args: { q: 'first' } } },
+      { type: 'tool-call-input-streaming-start', payload: { toolCallId: 'tc2', toolName: 'myTool', args: {} } },
+      { type: 'tool-call-delta', payload: { toolCallId: 'tc2', argsTextDelta: "{'q':'second'}" } },
+      { type: 'tool-call-input-streaming-end', payload: { toolCallId: 'tc2' } },
+      { type: 'tool-call', payload: { toolCallId: 'tc2', toolName: 'myTool', args: { q: 'second' } } },
+      { type: 'text-end', payload: { id: 't1' } },
+      { type: 'reasoning-end', payload: { id: 'r1' } },
+      { type: 'finish', payload: { finishReason: 'stop', usage: {} } },
+    ]);
+
+    expect(result).toHaveLength(4);
+    expect(result.map((p: any) => p.type)).toEqual(['reasoning', 'text', 'tool-invocation', 'tool-invocation']);
+    expect(result[0]).toMatchObject({ type: 'reasoning', details: [{ type: 'text', text: 'Thinking...' }] });
+    expect(result[1]).toMatchObject({ type: 'text', text: 'Hello' });
+    expect(result[2]).toMatchObject({
+      toolInvocation: { state: 'call', toolCallId: 'tc1', toolName: 'myTool', args: { q: 'first' } },
+    });
+    expect(result[3]).toMatchObject({
+      toolInvocation: { state: 'call', toolCallId: 'tc2', toolName: 'myTool', args: { q: 'second' } },
+    });
+  });
+
   // ── Empty stream / no parts ─────────────────────────────────
 
   it('should return empty array for empty chunks', () => {
--- a/packages/core/src/loop/workflows/agentic-execution/build-messages-from-chunks.test.ts
+++ b/packages/core/src/loop/workflows/agentic-execution/build-messages-from-chunks.test.ts
@@ -80,9 +80,10 @@ describe('buildMessagesFromChunks', () => {
       { type: 'text-end', payload: { id: 't1' } },
     ]);
     expect(result).toHaveLength(2);
-    // Parts are emitted in text-end order
-    expect(result[0]).toMatchObject({ type: 'text', text: 'Goodbye' });
-    expect(result[1]).toMatchObject({ type: 'text', text: 'Hello, world!' });
+    // Parts are emitted in first-seen-delta order (#15914): t1 receives its first delta
+    // before t2, so 'Hello, world!' precedes 'Goodbye' even though t2 ends first.
+    expect(result[0]).toMatchObject({ type: 'text', text: 'Hello, world!' });
+    expect(result[1]).toMatchObject({ type: 'text', text: 'Goodbye' });
   });
 
   // ── ProviderMetadata cascading ──────────────────────────────
--- a/packages/core/src/loop/workflows/agentic-execution/build-messages-from-chunks.ts
+++ b/packages/core/src/loop/workflows/agentic-execution/build-messages-from-chunks.ts
@@ -63,13 +63,23 @@ export function buildMessagesFromChunks({
     }
   }
 
-  // State for text span accumulation (keyed by text ID to handle interleaved spans)
-  const textSpans = new Map<string, { deltas: string[]; providerMetadata: Record<string, any> | undefined }>();
+  // State for text span accumulation (keyed by text ID to handle interleaved spans).
+  // `placeholder` reserves the part's position at first-seen-delta so the resulting
+  // parts array reflects semantic stream order rather than end-event order. See #15914.
+  const textSpans = new Map<
+    string,
+    { deltas: string[]; providerMetadata: Record<string, any> | undefined; placeholder: MastraMessagePart | null }
+  >();
 
   // State for reasoning span accumulation (keyed by reasoning ID)
   const reasoningSpans = new Map<
     string,
-    { deltas: string[]; providerMetadata: Record<string, any> | undefined; redacted: boolean }
+    {
+      deltas: string[];
+      providerMetadata: Record<string, any> | undefined;
+      redacted: boolean;
+      placeholder: MastraMessagePart | null;
+    }
   >();
   for (const chunk of chunks) {
     switch (chunk.type) {
@@ -80,6 +90,7 @@ export function buildMessagesFromChunks({
           textSpans.set(p.id, {
             deltas: [],
             providerMetadata: p.providerMetadata,
+            placeholder: null,
           });
         } else {
           // Update providerMetadata if this start has it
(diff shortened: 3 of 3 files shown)


TylerBarnes:
any chance this fix is related? https://github.com/mastra-ai/mastra/pull/15897

It fixes similar sounding issues for openai where reasoning/tool parts would get duplicated or put out of order

jmzhang:
> any chance this fix is related? #15897
> 
> It fixes similar sounding issues for openai where reasoning/tool parts would get duplicated or put out of order

I've came across this while debugging, in addResponseSnapshotBackToMessageList, the new messages were always already added during LLM Execution Step, but the wrong order of parts comes from buildMessagesFromChunk right before the messages are added to messageList, so probably two separate issues.

jmzhang:
@franciseliang99-dot I would generally agree with the first-seen ordering for parts, but the existing test cases indicate text spans shall follow text-end ordering (`should correctly separate interleaved text spans by ID` case), changing the ordering might unknowingly break other codes. On second thought, text spans might be the only exception to first-seen-delta order, and we shall use last-seen-delta as the order, since we can't stop providers from sending text-delta after tool call (conclusive remarks maybe?), and the last few comments in `buildMessagesFromChunks` says this exactly:
```
  // Insert step-start markers between tool-invocation and subsequent text parts.
  // This matches the convention used by MessageMerger.pushNewPart when merging messages,
  // and is required so that AI SDK convertToModelMessages splits them into separate steps.
  const finalParts: MastraMessagePart[] = [];
  for (let i = 0; i < parts.length; i++) {
    const part = parts[i]!;
    if (
      part.type === 'text' &&
      finalParts.length > 0 &&
      finalParts[finalParts.length - 1]?.type === 'tool-invocation'
    ) {
      finalParts.push({ type: 'step-start' } as MastraMessagePart);
    }
    finalParts.push(part);
  }
```

franciseliang99-dot:
Thanks for the thorough read @jmzhang — agreed on first-seen ordering as the general rule. Let me lay out why I think the unified rule still holds for text spans too, since I read the three pieces of evidence as pointing the other way:

1. **The existing `should correctly separate interleaved text spans by ID` expectation is itself a symptom of the same bug.** In that test, span `t1` starts and deltas first but `t2` ends first, and the old logic emits `[Goodbye, Hello, world!]`. Structurally that is identical to the `[tool, tool, text, reasoning]` ordering #15914 describes — both come from end-event timing dictating order. If first-seen-delta is correct for the cross-type case, the same argument applies to text-vs-text.

2. **The `step-start` insertion is a post-pass and is orthogonal to ordering.** The `finalParts` loop walks `parts[]` and inserts a `step-start` whenever a `text` follows a `tool-invocation`, regardless of how `parts[]` got ordered. So whether text uses first-seen or text-end, that loop behaves identically — it does not constrain the upstream ordering decision.

3. **Provider-emitted conclusive text after a tool call works correctly under first-seen too.** If a `text-delta` arrives after `tool-call`, the placeholder for that text gets pushed *after* the tool part — exactly the order we want. First-seen-delta means "first delta we observed", not "first start event", so late conclusive text naturally lands after tools.

Just pushed two updates: the text-spans test expectation now reflects first-seen semantics, and the changeset is rewritten to be outcome-focused per CodeRabbit's other comment.

Happy to revisit if maintainers see a downstream consumer that actually depends on text-end ordering specifically — in that case I can put a text-only branch behind a flag, but I would want to see the consumer first before adding a dual rule.

jmzhang:
Since I can't find any docs regarding the streaming order of chunks, I take this to Gemini and this is from its perspective:

> While there is no formal RFC "specification" for the sequence, the Vercel AI SDK follows a strict logical lifecycle based on how LLMs generate tokens. The streaming order is governed by the state of the model's output.
> 1. Logical Generation Order
In a standard request, chunks generally follow this chronological flow:
`text-delta` / `reasoning-delta`: These arrive first as the model begins its thought process or verbal response.
`tool-call` related chunks: If the model decides to use a tool, text generation usually pauses. You will see:
`tool-call-streaming-start`
`tool-call-delta` (repeated)
`tool-call` (the complete object)
`finish`: This is always the terminal chunk for a specific model turn. It contains the finishReason (e.g., "stop" or "tool-calls") and usage metrics.
> 2. Common Sequencing Rules
**Text vs. Reasoning**
Reasoning First: Models like DeepSeek or O1 typically emit reasoning chunks before the visible text chunks.
No Interleaving: Most providers do not interleave reasoning and text; they complete the reasoning "thought block" before starting the prose.
**Tool Call Integrity**
Mutual Exclusion: In a single "choice" or "turn," a model typically emits either text or a tool call. If it does both, the text almost always precedes the tool call.
Atomic Completion: A tool-call (complete) chunk will only arrive after all corresponding tool-call-delta chunks for that specific call have been emitted.
**Finish Chunks**
The Final Word: A finish chunk signals the end of the provider's stream.
Multi-step exception: If you have maxSteps set to >1, the SDK may internally handle multiple finish chunks (one for each tool call/response loop) before sending the final state to your UI.

So yes, I think this PR with the unified rule can fix the bug, and should be the expected behavior, only there's an edge case (maybe non-existent but possible) where the model streams interleaving text-delta with tool-call related chunks (notebook style?), such as:
```
...
[text-delta]
...
[text-delta]
[tool-call-input-streaming-start]
[tool-call-delta]
[tool-call-input-streaming-end]
[tool-call]
[text-delta]
...
[text-delta]
...
```

Neither first-seen-delta or last-seen-delta can pin the text part at the correct location, but we can probably skip this edge case for now, as it's imaginary, only a limitation of the current implementation.

franciseliang99-dot:
Thanks for the Gemini reference @jmzhang — the Vercel AI SDK lifecycle described there matches what downstream `addStartStepPartsForAIV5` and `step-start` insertion already assume (text/reasoning precede tool-call within a single turn, no interleaving), so the unified rule lands consistently with the rest of the pipeline.

And agreed on skipping the imaginary interleaved-text-around-tool case for now — first-seen vs last-seen both fail to pin a span split across a tool call. That shape would need chunk-level span fragmentation, which is a separate mechanism worth waiting for an actual repro to design.

Leaving the PR as-is for maintainer review.

TylerBarnes:
any reason we can't just push directly to an ordered array and not maintain separate tracking for different types of parts? if the array ends in text-delta (or some other delta) and a part besides text-end comes in, we can push a text-end, and then the next part

TylerBarnes:
superseded by https://github.com/mastra-ai/mastra/pull/16073 thanks for the work here though! ❤️

franciseliang99-dot:
@TylerBarnes thanks for taking a look. The separate tracking is there because spans typically emit multiple deltas before their end:

```
text-start(s1)
text-delta(s1, "hello, ")
text-delta(s1, "world")     // second delta of s1
text-end(s1)
```

With auto-close-on-type-change, the second `text-delta(s1)` is "a part besides text-end" and the array ends in `text-delta`, so we'd push a synthetic `text-end` and start a new text part — splitting s1 into `["hello, ", "world"]` instead of `"hello, world"`. Same bisect happens when a reasoning span interleaves between two deltas of the same text span.

The placeholder reserves s1's slot on the first delta and only fills it at `text-end(s1)`, which keeps same-span deltas in one part regardless of what arrives in between — that's also what the existing `should correctly separate interleaved text spans by ID` test asserts (span identity preserved across interleaving).

If you have a provider where streams are guaranteed single-delta-per-span and the simpler form is preferable, happy to gate the placeholder behind a multi-delta check or refactor altogether — let me know which consumer you have in mind and I'll match.

TylerBarnes:
@franciseliang99-dot all good, I was misunderstanding some things, but https://github.com/mastra-ai/mastra/pull/16073 looks like it achieves the same thing in a slightly cleaner way so lets go with that one

Issue #11519, opened 2025-12-31
Memory readOnly: true broken in 1.0.0-beta, saves to memory anyway


daneatmastra:
This issue was created from Discord post 1455989471131467887:

[![Open in Discord](https://img.shields.io/badge/Open_in_Discord-5865F2?style=for-the-badge&logo=discord&logoColor=white)](https://discord.com/channels/1309558646228779139/1455989471131467887)

Very easy to reproduce. Setup an agent, provide memory to it, call `.stream()` with `memory.readOnly: true`. It'll write to memory and it shouldn't.

We're on the latest beta version.

PR #15949, opened 2026-04-30
fix: honor readOnly in semantic recall indexing

### LINKED ISSUE #15950: [BUG] Semantic recall indexes messages when memory readOnly mode is enabled
## Describe the Bug

When memory read-only mode is enabled for an agent call, semantic recall still indexes new messages during output processing. Message persistence is skipped, but the semantic recall output processor can still create embeddings, create/validate the vector index, and upsert vectors for the current input/response messages.

This means `memory.options.readOnly: true` is not fully read-only when semantic recall is enabled.

## Steps To Reproduce

1. Configure memory with semantic recall enabled, including a vector store and embedder.
2. Call `agent.stream()` or `agent.generate()` with memory options containing `readOnly: true` and a valid thread/resource context.
3. Let `SemanticRecall.processOutputResult()` run after the model response.
4. Observe that the embedder and vector store can still be called even though memory is in read-only mode.

A focused regression test is included in PR #15949: `packages/core/src/processors/memory/semantic-recall.test.ts` asserts that with `memoryConfig.readOnly: true`, semantic recall does not call the embedder, `createIndex`, or `upsert`.

https://github.com/mastra-ai/mastra/pull/15949

## Expected Behavior

When `memoryConfig.readOnly` is true, semantic recall output processing should return the existing `MessageList` or messages unchanged and avoid all write-like side effects, including embedding generation, vector index creation, and vector upserts.

## Environment Information

- Package: `@mastra/core`
- Area: memory semantic recall output processor
- Reproduced against the current `main` branch via a unit test added in PR #15949

## Verification

KyleBoyer:
## Summary
- Skip semantic recall output indexing when memoryConfig.readOnly is true
- Add regression coverage to ensure read-only mode does not call embedder, createIndex, or upsert
- Add a patch changeset for @mastra/core

## Fixes
Fixes #15950

## Context
This is a follow-up to #11519 / #11523. That fix stopped message persistence and savePerStep writes from ignoring `memory.options.readOnly`, but the semantic recall output processor was already a separate write path at the time and still embedded/upserted new messages into the vector store.

## Tests
- pnpm --filter ./packages/core test -- src/processors/memory/semantic-recall.test.ts
- pnpm --filter ./packages/core test -- src/agent/__tests__/memory-readonly.test.ts


KyleBoyer:
@coderabbitai review

KyleBoyer:
Historical context after checking #11519 / #11523:

#11523 fixed the main message persistence paths for `memory.options.readOnly` (including `savePerStep`) and added tests around `MockMemory.saveMessages`, but it did not touch `packages/core/src/processors/memory/semantic-recall.ts`.

At the merge commit for #11523 (`e0941c3d`), `SemanticRecall.processOutputResult()` already parsed `MastraMemory` from the request context and then proceeded to embed new messages, ensure/create the vector index, and `vector.upsert(...)` without checking `memoryConfig.readOnly`. So this PR is a targeted follow-up for the same read-only contract, specifically covering semantic recall's vector indexing side effect.

mikhael28:
@coderabbitai please review!

abhiaiyer91 (approved):

PR #15956, opened 2026-04-30
Add top-level environment config to Mastra for observability tagging

epinzur:
## Description

This PR adds a top-level `environment` configuration option to the `Mastra` class that automatically attaches the deployment environment to every observability signal (traces, logs, metrics) without requiring `tracingOptions.metadata.environment` on each call.

### Key Changes

- **`Mastra` class**: Added `environment` config field and `getEnvironment()` method that resolves to the explicit config value, falls back to `process.env.NODE_ENV`, or returns `undefined` if neither is set
- **`Observability` entrypoint**: Modified `setMastraContext` to propagate the resolved environment to all registered observability instances via `__setMastraEnvironment()`
- **`BaseObservabilityInstance`**: Added `getMastraEnvironment()` and `__setMastraEnvironment()` methods to store and retrieve the Mastra-level environment
- **Span correlation context**: Updated to use the Mastra environment as a fallback when `metadata.environment` is not set on a specific span
- **Documentation**: Added configuration guide and updated reference docs

### Behavior

- Per-call `tracingOptions.metadata.environment` always takes precedence
- Falls back to `process.env.NODE_ENV` when `environment` config is not set
- Child spans inherit the environment through metadata propagation
- Observability instances receive the environment during `setMastraContext` initialization

## Type of Change

- [x] Documentation update
- [x] Test update

## Checklist

- [x] I have made corresponding changes to the documentation
- [x] I have added tests that prove my feature works
  - Added comprehensive test suite in `packages/core/src/mastra/environment.test.ts` covering config resolution, fallback behavior, and observability propagation
  - Added integration tests in `observability/mastra/src/tracing.test.ts` verifying span correlation context behavior with Mastra environment fallback

https://claude.ai/code/session_012fW46J3P9uuA3v8XsjcCYA

## ELI5 (Explain Like I'm 5)

Set the environment once on Mastra (like "production" or "test"), and that label will automatically be attached to all traces, logs, metrics, and persisted telemetry so you don't have to pass it on every call.

## Overview

Adds a top-level `environment` configuration to Mastra, resolves it (explicit config → `process.env.NODE_ENV` → `undefined`), exposes it via `getEnvironment()`, and propagates the resolved value into observability instances during `setMastraContext` so traces, logs, metrics, and persisted span records include the deployment environment unless a per-call override is provided.

## Key Changes

- Mastra core
  - Adds optional `environment?: string` to `Config`.
  - Stores resolved environment on construction (explicit config → `process.env.NODE_ENV` → `undefined`).
  - Adds `getEnvironment(): string | undefined`.

- Observability entrypoint
  - `setMastraContext` captures `mastra.getEnvironment?.()` and calls `__setMastraEnvironment()` on each registered observability instance during initialization.

- Observability instance API & implementation
  - `ObservabilityInstance` interface gains optional `getMastraEnvironment?(): string | undefined` and `__setMastraEnvironment?(environment: string | undefined): void`.
  - `BaseObservabilityInstance` implements private storage plus `__setMastraEnvironment()` and `getMastraEnvironment()`.

- Span correlation & persistence
  - Root span creation (`startSpan`) injects Mastra-level environment into root span `metadata` when `tracingOptions.metadata.environment` is not provided; child spans inherit via existing metadata propagation.
  - `BaseSpan.getCorrelationContext()` prefers `metadata.environment` and falls back to `observabilityInstance.getMastraEnvironment()`.
  - Persisted RecordedSpan/RecordedTrace consumers (scores, feedbacks, downstream readers) preserve and surface the environment via `correlationContext.environment`.

- Tests & snapshots
  - Unit tests: `packages/core/src/mastra/environment.test.ts` (config resolution, precedence, and propagation into observability).
  - Integration tests: `observability/mastra/src/tracing.test.ts`, `recorded.test.ts`, and other integration fixtures updated to assert environment presence across traces, logs, metrics, recorded scores/feedbacks, and exporter correlation contexts.
  - Many trace snapshots updated to include `metadata.environment` values (e.g., `"test"` or `"production"`).

## Precedence

1. per-call `tracingOptions.metadata.environment` (highest)  
2. Mastra-level `environment` config  
3. `process.env.NODE_ENV`  
4. `undefined` if none are present

## Documentation

- Tracing overview and Mastra class reference updated to document the new `environment` config, its resolution rules, precedence, and observability propagation.

## Release / Changesets

- Changesets added documenting the Mastra-level `environment` behavior; intended minor releases for core/observability and documentation-only changesets for snapshots.


epinzur:
@coderabbitai pls review

observability/mastra/src/default.ts
```diff
@@ -152,7 +152,13 @@ export class Observability extends MastraBase implements ObservabilityEntrypoint
     const { mastra } = options;
     this.#mastra = mastra;
 
+    const mastraEnvironment = mastra.getEnvironment?.();
+
     instances.forEach(instance => {
+      // Propagate the Mastra-level environment so spans can fall back to it
+      // when `metadata.environment` isn't set on a specific span.
+      instance.__setMastraEnvironment?.(mastraEnvironment);
```
noahcrowley, line 160: When an observability instance is registered after `setMastraContext()` has already run, it never receives this environment value. This shows up in `Mastra.registerExporter()`, which calls `setMastraContext({ mastra })` before `registerInstance('default', instance, true)`, so the registry is still empty during this loop. It also affects any post-construction `observability.registerInstance(...)` call. Consider propagating `this.#mastra?.getEnvironment()` inside `Observability.registerInstance()` when context is already available, so late-registered instances also auto-tag spans with `environment`.

noahcrowley (approved):

observability/mastra/src/default.ts
epinzur: Good catch — confirmed the bug and fixed in 430d1f7.

`Observability.registerInstance` now pushes `mastra.getEnvironment()` to the new instance whenever `#mastra` is already set, which covers both the `Mastra.registerExporter` bootstrap path and any post-construction `observability.registerInstance(...)`.

Added a regression test in `registry.test.ts` ("propagates the Mastra environment to instances registered after setMastraContext") that fails on the prior code and passes with the fix.

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #15981, opened 2026-04-30
fix(cli): use HTTPS for dev internal fetches

### LINKED ISSUE #15953: `mastra dev` hardcodes `http://` for internal fetches, breaks under `server.https` (workflow restart, refresh, hot-reload)
# `mastra dev` hardcodes `http://` for internal fetches, breaks when `server.https` is configured

## Summary

When a Mastra app is configured with `server.https` (TLS in dev via `mkcert` or similar), `mastra dev` logs:

```
✗ Failed to restart all active workflow runs: TypeError: fetch failed
```

shortly after `mastra <version> ready`. The cause is that the dev CLI hardcodes `http://` for several internal post-startup fetches, even though the dev server is HTTPS-only on the same port. The retry uses the same hardcoded scheme, so it also fails. The hot-reload / refresh fetches have the same issue (silent there because the retry catches without logging).

## Version

- `mastra@1.7.1` (also reproduces on `main`, `1.6.2`, and `1.7.2-alpha.0`)
- Node 24
- macOS 25.4 (Darwin), but not OS-specific

## Repro

Minimal Mastra app with `server.https` set:

```ts
import fs from 'node:fs';

  // ...storage, agents, workflows...
  server: {
    port: 4060,
    host: '0.0.0.0',
    https: {
      cert: fs.readFileSync('localhost.pem'),  // mkcert-generated
      key:  fs.readFileSync('localhost-key.pem'),
    },
  },
});
```

Run `mastra dev`. Observe in the output:

```
[ai-agent] Workspace init complete
 mastra  1.7.1 ready in 8112 ms
│ Studio: https://0.0.0.0:4060
│ API:    https://0.0.0.0:4060/api
◯ watching for file changes...
✗ Failed to restart all active workflow runs: TypeError: fetch failed
```

Note that the warning fires *after* the server reports ready, ruling out a startup race — storage and the HTTP listener are both fully up. The `fetch` call fails because the CLI sends plain HTTP to a TLS-only port.

## Source pointers

In `packages/cli/src/commands/dev/dev.ts` on `main`:

- L77, L85 — `restartAllActiveWorkflowRuns()` posts to `http://${host}:${port}/__restart-active-workflow-runs` (initial + retry).
- L213, L223 — `__refresh` posts (also hardcoded `http://`).
- L295 — `__hot-reload` fetch (same).

The `https` config is already in scope at the call site: L206 passes `startOptions.https` to `devLogger.ready(...)`, but it isn't threaded into `restartAllActiveWorkflowRuns` or the refresh/hot-reload fetches.

## Suggested fix

Thread the protocol through the existing `startOptions.https` flag at the call sites. Either:

1. Pass `https` into `restartAllActiveWorkflowRuns` and the inline fetches, e.g.

   ```ts
   const scheme = startOptions.https ? 'https' : 'http';
   await fetch(`${scheme}://${host}:${port}/__restart-active-workflow-runs`, { method: 'POST' });
   ```

2. Or compute a `selfBaseUrl` once when starting the server and reuse it for all internal fetches.

When `https` is enabled with a self-signed/mkcert cert, `fetch` may also need `rejectUnauthorized: false` (via a custom `dispatcher`) since the dev cert chain isn't in Node's trust store. Worth handling for completeness.

## Workaround

For users hitting this today: the warning is cosmetic in dev (no real consumer of `restart-active-workflow-runs` for ephemeral local sessions), so it can be ignored. Alternatively, `yarn patch mastra` to flip the four hardcoded `http://` strings, but that breaks across upgrades.

## Why we hit it

We run the Mastra dev server on `https://localhost:4060` because the consuming SPA is served from `https://localhost:9000`, and it calls the API directly (not via a proxy) — so plain HTTP on the API would be blocked as mixed content. There's no straightforward way to keep the existing dev topology without `server.https`.

arjunrawal1:
## Description

Fixes #15953.

`mastra dev` starts the local server with HTTPS when `--https` or `server.https` is configured, but its internal workflow restart, Studio refresh, and hot-reload status fetches were still hardcoded to `http://`. This derives the internal dev server base URL from the active `StartOptions.https` value and reuses it for those fetches.

## Validation

- `corepack pnpm@10.18.0 --dir packages/cli exec vitest run src/commands/dev/dev.test.ts --project unit:packages/cli`
- `corepack pnpm@10.18.0 --filter ./packages/cli typecheck`
- `corepack pnpm@10.18.0 --filter ./packages/cli lint -- src/commands/dev/dev.ts src/commands/dev/dev.test.ts`

## ELI5

When you run `mastra dev` with a secure (HTTPS) server, the CLI tool was trying to talk to it using regular, unencrypted (HTTP) requests, which failed. This fix makes the CLI automatically detect whether the server is using HTTPS and use secure requests too, so everything works together properly.

### Test Coverage (`dev.test.ts`)
- Added test case verifying that when `dev` is started with `https: true`, all internal dev-server interactions use HTTPS endpoints
- Test validates that workflow restart, refresh, and hot-reload status requests are sent to `https://localhost:4111/...` with appropriate request headers

### Implementation (`dev.ts`)
- Introduced helper function to derive dev server base URL from `startOptions.https` (selecting `https://` or `http://` scheme accordingly)
- Refactored internal dev HTTP calls to use the dynamically-determined base URL instead of hardcoded `http://`
- Updated endpoints:
  - `restartAllActiveWorkflowRuns`: now posts to `${baseUrl}/__restart-active-workflow-runs`
  - Studio refresh: now targets `${serverBaseUrl}${studioBasePath}/__refresh`
  - Hot-reload status check: now fetches `${serverBaseUrl}${studioBasePath}/__hot-reload-status`

### DIFF
--- a/packages/cli/src/commands/dev/dev.test.ts
+++ b/packages/cli/src/commands/dev/dev.test.ts
@@ -359,4 +359,49 @@ describe('dev command - inspect flag behavior', () => {
       expect(commands.some(cmd => cmd.startsWith('--inspect-brk='))).toBe(false);
     });
   });
+
+  describe('HTTPS internal dev server requests', () => {
+    it('uses https for ready and hot-reload fetches when dev HTTPS is enabled', async () => {
+      const fetchMock = vi.fn().mockResolvedValue({
+        ok: true,
+        json: vi.fn().mockResolvedValue({ disabled: true, timestamp: 'now' }),
+      });
+      vi.stubGlobal('fetch', fetchMock);
+
+      const { dev } = await import('./dev');
+
+      await dev({
+        dir: undefined,
+        root: process.cwd(),
+        tools: undefined,
+        env: undefined,
+        inspect: false,
+        inspectBrk: false,
+        customArgs: undefined,
+        https: true,
+        debug: false,
+      });
+
+      fetchMock.mockClear();
+
+      await vi.waitFor(() => {
+        expect(fetchMock).toHaveBeenCalledWith('https://localhost:4111/__restart-active-workflow-runs', {
+          method: 'POST',
+        });
+      });
+      expect(fetchMock).toHaveBeenCalledWith('https://localhost:4111/__refresh', {
+        method: 'POST',
+        headers: {
+          'Content-Type': 'application/json',
+        },
+      });
+
+      const bundleEndHandler = mockWatcher.on.mock.calls.find(([event]) => event === 'event')?.[1];
+      expect(bundleEndHandler).toBeDefined();
+
+      await bundleEndHandler({ code: 'BUNDLE_END' });
+
+      expect(fetchMock).toHaveBeenCalledWith('https://localhost:4111/__hot-reload-status');
+    });
+  });
 });
--- a/packages/cli/src/commands/dev/dev.ts
+++ b/packages/cli/src/commands/dev/dev.ts
@@ -72,9 +72,14 @@ type ProcessOptions = {
   publicDir: string;
 };
 
-const restartAllActiveWorkflowRuns = async ({ host, port }: { host: string; port: number }) => {
+const getDevServerBaseUrl = ({ host, port, https }: { host: string; port: number; https?: HTTPSOptions }) => {
+  const protocol = https ? 'https' : 'http';
+  return `${protocol}://${host}:${port}`;
+};
+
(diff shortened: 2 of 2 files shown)


mikhael28:
We have an internal PR in progress for this, closing.

PR #16036, opened 2026-05-01
fix(client-js): align Vector return types with server

intojhanurag:
## Description

Aligns three Vector resource return types in `@mastra/client-js` with what the server actually returns. The previous types described shapes that did not exist on the wire, leading to silent runtime failures with no TypeScript errors.

- Changed `vector.getIndexes()` from `Promise<{ indexes: string[] }>` to `Promise<string[]>`
- Changed `vector.upsert()` from `Promise<string[]>` to `Promise<{ ids: string[] }>`
- Changed `vector.query()` from `Promise<QueryVectorResponse>` to `Promise<QueryResult[]>`
- Updated the public `QueryVectorResponse` type from `{ results: QueryResult[] }` to a `QueryResult[]` alias so external imports still resolve
- Fixed unit-test mocks in `vector.test.ts` to match real server response shapes, closing the loophole that hid the bug at unit level
- Removed `: any` casts and `TODO(#15089)` workarounds in `e2e-tests/client-js/_test-utils/src/vector-tests.ts`
- Server is unchanged — alignment was done on the SDK side to preserve the public HTTP contract

## Related Issue(s)

Fixes #15089

## Type of Change

- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [ ] Test update

## Checklist

## ELI5

This PR fixes the client library so its TypeScript types match what the server actually sends. Previously the SDK promised different shapes than the server returned, causing runtime crashes; now the SDK types and tests reflect the real responses.

### Method Return Type Corrections
- getIndexes(): Promise<{ indexes: string[] }> → Promise<string[]>
- upsert(): Promise<string[]> → Promise<{ ids: string[] }>
- query(): Promise<QueryVectorResponse> → Promise<QueryResult[]>

### Type Definition Updates
- client-sdks/client-js/src/types.ts: `QueryVectorResponse` changed from `interface { results: QueryResult[] }` to `export type QueryVectorResponse = QueryResult[]` (keeps external imports resolvable while matching server shape).

### Tests and E2E Test Utilities
- client-sdks/client-js/src/resources/vector.test.ts: mocked fetch payloads and assertions updated to the real server response shapes for getIndexes, upsert, and query.
- e2e-tests/client-js/_test-utils/src/vector-tests.ts: removed `: any` casts and TODO(#15089) workarounds so inferred types reflect corrected SDK types.

### Release Notes / Changeset
- .changeset/vector-sdk-return-types.md added documenting the breaking changes with before/after examples and referencing issue #15089.

## Breaking Changes
- This is a breaking change for @mastra/client-js: consumers must update call sites that expected the old wrapped/ unwrapped shapes (e.g., stop using `result.indexes` and use the array directly, handle `{ ids: string[] }` for upsert, treat query responses as an array).

## Server Compatibility
- No server changes: the SDK is updated to match the existing server HTTP responses (non-breaking to the server/API).

### DIFF
--- a/client-sdks/client-js/src/resources/vector.ts
+++ b/client-sdks/client-js/src/resources/vector.ts
@@ -1,9 +1,9 @@
 import type { RequestContext } from '@mastra/core/request-context';
-import type { QueryResult } from '@mastra/core/vector';
 import type {
   CreateIndexParams,
   GetVectorIndexResponse,
   QueryVectorParams,
+  QueryVectorResponse,
   ClientOptions,
   UpsertVectorParams,
 } from '../types';
@@ -82,7 +82,7 @@ export class Vector extends BaseResource {
    * @param params - Query parameters including query vector and search options
    * @returns Promise containing query results
    */
-  query(params: QueryVectorParams): Promise<QueryResult[]> {
+  query(params: QueryVectorParams): Promise<QueryVectorResponse> {
     return this.request(`/vector/${encodeURIComponent(this.vectorName)}/query`, {
       method: 'POST',
       body: params,
--- a/client-sdks/client-js/src/resources/vector.ts
+++ b/client-sdks/client-js/src/resources/vector.ts
@@ -1,9 +1,9 @@
 import type { RequestContext } from '@mastra/core/request-context';
+import type { QueryResult } from '@mastra/core/vector';
 import type {
   CreateIndexParams,
   GetVectorIndexResponse,
   QueryVectorParams,
-  QueryVectorResponse,
   ClientOptions,
   UpsertVectorParams,
 } from '../types';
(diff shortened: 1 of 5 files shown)


intojhanurag:
Review the changeset carefully .

client-sdks/client-js/src/resources/vector.ts
```diff
@@ -1,9 +1,9 @@
 import type { RequestContext } from '@mastra/core/request-context';
+import type { QueryResult } from '@mastra/core/vector';
 import type {
   CreateIndexParams,
   GetVectorIndexResponse,
   QueryVectorParams,
-  QueryVectorResponse,
```
NikAiyer, line 6: It looks we're still exporting this type, so we might as well still use it rather than import `QueryResult`.

NikAiyer (approved):

PR #13150, opened 2026-02-17
 feat(core): support dynamic filesystem resolution via requestContext in Workspace

intojhanurag:
## Description

Workspace options like skills, instructions, model, tools, and memory already support dynamic functions that receive `requestContext`, but the filesystem only accepted a static instance.  

This PR adds support for a resolver function  
`({ requestContext }) => WorkspaceFilesystem`, allowing a single Workspace instance to serve different filesystems per request.  

This enables multi-role agents to scope each role to a different filesystem root or permissions.

### Key Changes
- Added `WorkspaceFilesystemResolver` type and widened filesystem config to accept it  
- Added `hasFilesystemConfig()` and `resolveFilesystem()` public methods on `Workspace`  
- All auto-injected workspace tools now resolve filesystem dynamically at execution time via a Proxy  
- Runtime `readOnly` enforcement for dynamically resolved filesystems on write/edit/delete/mkdir tools  
- Class constructor guard — passing `LocalFilesystem` (class) instead of `new LocalFilesystem(...)` throws a clear error  
- Fully backward compatible — static filesystem works exactly as before  

## Related Issue(s)

Fixes #13133

## Type of Change

- [ ] Bug fix (non-breaking change that fixes an issue)
- [x] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [x] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [ ] Test update

## Checklist

- [x] I have added tests that prove my fix is effective or that my feature works
- [ ] I have made corresponding changes to the documentation (if applicable)

## Summary by CodeRabbit

* **New Features**
  * Workspace filesystem can be a static instance or a per-request resolver for dynamic routing.
  * New APIs to detect and resolve per-request filesystems at runtime (hasFilesystemConfig, resolveFilesystem).
  * Workspace tools receive execution context so filesystem selection occurs per request.
  * Write tools remain visible for dynamic filesystems; writes are blocked at runtime when filesystem is read-only.

* **Tests**
  * Added tests for per-request filesystem resolution and read-only enforcement.

* **Documentation**
  * Docs updated with dynamic filesystem guidance and examples.

PR #16048, opened 2026-05-01
feat(core): support dynamic sandbox resolution via requestContext

### LINKED ISSUE #15664: [FEATURE] Support dynamic sandbox resolution via requestContext in Workspace
### Problem Statement

Workspace options like `skills`, `instructions`, `model`, `tools`, and `memory` all support dynamic functions that receive `requestContext`:

```ts
skills: (context) => { ... }          // ✅ dynamic
```

But `sandbox` only accepts a static instance:

```ts
// no way to resolve dynamically per request
sandbox: new LocalSandbox({
    workingDirectory: './workspace',
})
```

### Proposed Solution

Allow `sandbox` to accept a function that receives `requestContext`, matching the pattern of every other dynamic option:

```ts
    sandbox: ({ requestContext }) => {
        const role = requestContext.get('agent-role');
        return new LocalSandbox({
            workingDirectory: `/workspaces/${role}`,
        });
    },
    // ...
});
```

### Component

Agents

### Alternatives Considered

_No response_

### Example Use Case

We run a multi-role agent on a shared workspace. Each role should be scoped to a different working directory or have different execution permissions. Today we're forced to create separate `Workspace` instances per role at runtime. A dynamic `sandbox` resolver would let a single `Workspace` instance scope execution per request — different working directories, different permissions, or even different sandbox implementations.

### Additional Context

This is essentially the same request as #13133 for `filesystem` — both options suffer from the same limitation.

### Verification

intojhanurag:
## Description

Workspace `sandbox` now accepts a resolver function `({ requestContext }) => WorkspaceSandbox` in addition to a static instance. This enables per-request sandbox routing from a single Workspace, mirroring the dynamic-filesystem pattern shipped in #13150 and unblocking the multi-tenant use case described in #15664.

- Added `WorkspaceSandboxResolver` type and widened `config.sandbox` on `WorkspaceConfig` to accept it.
- Added `Workspace.hasSandboxConfig()` and `Workspace.resolveSandbox({ requestContext })` to check availability and resolve per-request.
- Constructor rejects passing a class constructor (clear `INVALID_CONFIG` error) and rejects `mounts` + sandbox resolver, since mounts need a concrete sandbox at construction time.
- LSP warns and disables when combined with a sandbox resolver, since LSP attaches to the process manager at construction time.
- `getInfo()` reports `provider: 'dynamic'` for resolver-backed sandbox without invoking the resolver, so per-tool metadata emission stays cheap.
- `init()` and `destroy()` skip lifecycle calls for resolver-backed providers — the caller owns the lifecycle of any sandbox returned by the resolver.
- Refactored `resolveEffectiveWorkspace` in `tools/tools.ts` to take a `targets` argument so each tool only resolves the providers it actually uses (filesystem tools never invoke the sandbox resolver and vice versa).
- Sandbox tools register from `hasSandboxConfig()`. Dynamic sandbox path registers `execute_command` (with background), `get_process_output`, and `kill_process`; missing capabilities surface as `SandboxFeatureNotSupportedError` at runtime.
- Added `dynamic-sandbox.test.ts` (7 tool-level tests) and a new `dynamic sandbox` describe block in `workspace.test.ts` (11 unit tests). All 153 tests pass across affected files; zero typecheck errors in workspace package.
- Updated `docs/workspace/sandbox.mdx`, `docs/workspace/overview.mdx`, and `reference/workspace/workspace-class.mdx` with the new option, methods, and tool-registration behavior.

## Related Issue(s)

Fixes #15664

## Type of Change

- [x] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [ ] Test update

## Checklist

## ELI5

A Workspace can now choose a different execution sandbox for each request by calling a small resolver function that looks at the request context (who/what made the request). This lets one Workspace serve multiple users/roles with different execution environments without creating separate Workspace instances.

## Summary

This PR adds dynamic per-request sandbox resolution to Workspace by allowing WorkspaceConfig.sandbox to be either a static WorkspaceSandbox or a resolver function (WorkspaceSandboxResolver). It preserves static behavior, introduces validation and lifecycle semantics for resolver-backed sandboxes, updates tool registration and resolution to support dynamic sandboxes, adds async instruction/path-context APIs, expands tests and types, and updates documentation.

### Core API Changes
- New exported type: WorkspaceSandboxResolver: ({ requestContext }) => WorkspaceSandbox | Promise<WorkspaceSandbox>.
- WorkspaceConfig.sandbox now accepts WorkspaceSandbox | WorkspaceSandboxResolver.
- New Workspace instance methods:
  - hasSandboxConfig(): boolean
  - resolveSandbox({ requestContext }): Promise<WorkspaceSandbox | undefined>
  - getInstructionsAsync(opts?: { requestContext?: RequestContext }): Promise<string>
  - getPathContextAsync(opts?: { requestContext?: RequestContext }): Promise<PathContext>
- When a resolver is configured, workspace.sandbox remains undefined; callers should use hasSandboxConfig()/resolveSandbox().

### Validation & Constraints
- Constructor rejects class constructors passed as sandbox (INVALID_CONFIG).
- Providing mounts together with a sandbox resolver throws INVALID_CONFIG (mounts require a concrete sandbox at construction).
- lsp: true is warned and disabled when used with a sandbox resolver (LSP requires a concrete sandbox/process manager at construction time).

### Lifecycle Semantics
- Resolver-returned sandboxes are caller-owned: Workspace.init() and destroy() do not call start()/destroy() on resolver results; the caller manages those lifecycles.
- getInfo() reports provider: 'dynamic' (status: 'pending') for resolver-backed sandboxes without invoking the resolver.

### Tools & Runtime Behavior
- resolveEffectiveWorkspace now accepts provider targets so each tool resolves only the providers it needs.
- Filesystem tools request { filesystem: true } and do not invoke the sandbox resolver.
- Sandbox tools request { sandbox: true } and are registered when hasSandboxConfig() is true (even if workspace.sandbox is not statically set).
- For dynamic sandboxes, the workspace registers execute_command (including background support), get_process_output, and kill_process. Missing capabilities on the resolved sandbox surface at runtime as SandboxFeatureNotSupportedError.
- Resolver results (filesystem and sandbox) are memoized per RequestContext so a single request shares one resolved provider instance across consumers.
- init()/destroy() skip lifecycle calls for resolver-backed providers.

### Async Instructions & Agent Integration
- getInstructionsAsync/getPathContextAsync resolve dynamic providers using optional requestContext and forward the synthesized requestContext to provider instruction hooks.
- WorkspaceInstructionsProcessor and durable agent preparation prefer the async instruction APIs when present and forward requestContext accordingly.
- resolveEffectiveWorkspace/proxying ensures tools see resolved provider instances for the current request.

### Tests & Types
- New/updated tests:
  - packages/core/src/workspace/tools/__tests__/dynamic-sandbox.test.ts (tool-level tests covering resolver behavior, background processes, caching, and errors).
  - Expanded dynamic sandbox tests in packages/core/src/workspace/workspace.test.ts (constructor validation, lifecycle, LSP interaction, resolver behavior, instructions integration).
  - Type-level test: packages/core/src/workspace/workspace.test-d.ts verifying resolver typing and method signatures.
- Reported: all affected tests (153) pass and the workspace package typechecks with zero errors.

### Documentation
- Updated docs: docs/workspace/sandbox.mdx, docs/workspace/overview.mdx, reference/workspace/workspace-class.mdx, and related filesystem/instructions docs to document resolver usage, lifecycle semantics, tool-registration behavior, examples, and constraints.

Fixes #15664.


packages/core/src/workspace/workspace.ts
```diff
@@ -490,14 +511,34 @@ export class Workspace<
     this.lastAccessedAt = new Date();
 
     this._config = config;
-    this._sandbox = config.sandbox;
+
+    if (typeof config.sandbox === 'function') {
+      if (/^class\s/.test(Function.prototype.toString.call(config.sandbox))) {
```
wardpeet: why do we care about this? Because  they are class instances?

intojhanurag: It is just a guard , In JS, typeof SomeClass === 'function' is true, so the resolver-detection branch (typeof config.sandbox === 'function') accidentally accepts the class itself if someone writes
sandbox: LocalSandbox instead of sandbox: new LocalSandbox() or sandbox: () => new LocalSandbox().

I think , it doesn't need ,I copied it from dynamic filesystem part where this type of guard was . ( Since I also implemented that so  )

intojhanurag: btw I am removing this , I don't think any user will do this mistake lol

TylerBarnes (dismissed):
I like the overall shape of this feature — especially keeping tool registration stable for resolver-backed sandboxes instead of changing the tool set per request. I think there are two dynamic-sandbox semantics we should tighten before merge.

### 1. `getInstructionsAsync()` should not resolve/provision dynamic sandboxes by default

The biggest concern is that dynamic sandbox resolution now happens while building workspace instructions.

`WorkspaceInstructionsProcessor` injects workspace instructions into the system message, and for resolver-backed sandboxes that path calls `workspace.getInstructionsAsync({ requestContext })`. `getInstructionsAsync()` then calls `resolveSandbox({ requestContext })`.

That means a sandbox resolver may create/start/provision a sandbox just to build the prompt, even if the model never calls `execute_command`.

This is especially concerning because resolver-backed sandboxes are caller-owned and skipped by `Workspace.init()` / `Workspace.destroy()`. If prompt construction alone can resolve a sandbox, the workspace may trigger external resource allocation without also owning cleanup for that resource. That makes lifecycle expectations harder to reason about.

There’s also a related prompt-cache concern: because these instructions become part of the system message, concrete sandbox instructions can make the prompt request-specific. Different tenants/users/roles may produce different working directories, mount descriptions, or provider instruction text. Tool registration is stable for dynamic sandboxes, which is good, but the system message can still vary per request.

Could dynamic sandbox instructions be configurable separately from dynamic sandbox execution?

For example, the default for resolver-backed sandboxes could avoid calling the resolver and emit stable placeholder text:

> Dynamic sandbox configured. Shell commands execute in a request-scoped sandbox resolved at tool execution time.

Then users who need concrete per-request sandbox details in the prompt could opt in explicitly, accepting the eager-provisioning, caller-owned lifecycle, and prompt-cache trade-offs. Something along these lines:

```ts
new Workspace({
  sandbox: ({ requestContext }) => resolveSandbox(requestContext),

  instructions: {
    dynamicSandbox: 'placeholder', // default, does not call resolver
    // or 'resolve' for current behavior
    // or ({ requestContext }) => 'custom instruction text'
  },
});
```

The exact API can differ, but I think it’s important to separate “resolve sandbox for tool execution” from “resolve sandbox while constructing system prompt”.

### 2. Background process continuity should not rely on `RequestContext` object identity

`resolveSandbox()` caches resolver results in a `WeakMap<RequestContext, Promise<WorkspaceSandbox>>`.

That fixes the prompt/tool mismatch only when the exact same `RequestContext` instance flows through the whole request. But background processes can outlive a single tool call. If `execute_command({ background: true })` runs with one `RequestContext` instance and `get_process_output` later runs with a different-but-equivalent instance, the resolver may return a different sandbox and lose access to the process.

For process-capable dynamic sandboxes, continuity should be keyed by stable request/session/tenant identity, not by object identity. Otherwise the behavior depends on whether callers preserve the same `RequestContext` object across follow-up calls, which is easy to get wrong.

Could we add an explicit stable cache-key mechanism for resolver-backed sandboxes that support background processes? For example:

```ts
new Workspace({
  sandbox: ({ requestContext }) => resolveSandbox(requestContext),
  sandboxCacheKey: ({ requestContext }) => requestContext.get('threadId') ?? requestContext.get('tenantId'),
});
```

or make this part of the resolver contract/options somehow. The important part is that `execute_command(background: true)`, `get_process_output`, and `kill_process` should resolve to the same sandbox based on stable logical identity, not only the same `RequestContext` object instance.

### Summary

I’m supportive of the feature, but I think we should avoid two footguns before merge:

- prompt construction should not resolve/provision caller-owned dynamic sandboxes by default
- background process continuity should not depend on `RequestContext` object identity

My preferred direction would be:

- keep dynamic sandbox tool registration stable
- keep resolver-backed sandbox lifecycle caller-owned
- use stable placeholder instructions for dynamic sandboxes by default
- add an explicit opt-in/config path for concrete dynamic sandbox instructions
- add a stable cache-key mechanism for process-capable dynamic sandboxes

packages/core/src/workspace/workspace.ts
```diff
@@ -723,10 +805,61 @@ export class Workspace<
...
+      }
+      return keyed;
+    }
+
+    let pending = this._sandboxRequestCache.get(requestContext);
+    if (!pending) {
+      pending = Promise.resolve(this._sandboxResolver({ requestContext }));
+      this._sandboxRequestCache.set(requestContext, pending);
+    }
+    return pending;
```
NikAiyer, line 871: **Rejected resolver promises poison the cache.**

The in-flight promise is stored into `_sandboxKeyCache` / `_sandboxRequestCache` *before* awaiting (lines 837 and 845). If the resolver throws, the rejected promise stays cached and every subsequent lookup with the same `cacheKey` or `RequestContext` re-throws without retry.

Either:
- `pending.catch(() => this._sandboxRequestCache.delete(requestContext))` (and the same for the keyed cache), or
- Document this as intentional so callers know they need a new `RequestContext` (or to call `clearSandboxCache(key)`) after a failure.

Not covered by the new tests.

packages/core/src/workspace/tools/tools.ts
```diff
@@ -448,23 +486,30 @@ export async function createWorkspaceTools(
...
+      await addTool(WORKSPACE_TOOLS.SANDBOX.GET_PROCESS_OUTPUT, getProcessOutputTool, { targets: { sandbox: true } });
+      await addTool(WORKSPACE_TOOLS.SANDBOX.KILL_PROCESS, killProcessTool, { targets: { sandbox: true } });
     }
+  } else if (hasSandboxConfig(workspace)) {
+    await addTool(WORKSPACE_TOOLS.SANDBOX.EXECUTE_COMMAND, executeCommandWithBackgroundTool, {
+      targets: { sandbox: true },
+    });
+    await addTool(WORKSPACE_TOOLS.SANDBOX.GET_PROCESS_OUTPUT, getProcessOutputTool, { targets: { sandbox: true } });
+    await addTool(WORKSPACE_TOOLS.SANDBOX.KILL_PROCESS, killProcessTool, { targets: { sandbox: true } });
   }
```
NikAiyer, line 507: **Background-process tools register unconditionally without `sandboxCacheKey`.**

`get_process_output` and `kill_process` always register when a sandbox resolver is configured. If the resolver returns a fresh sandbox per call and the user hasn't set `sandboxCacheKey`, the WeakMap fallback in `resolveSandbox` keys by `RequestContext` identity — so a PID spawned in one tool call won't survive to the next.

Two reasonable mitigations:
- Warn at construction time when a resolver is configured without `sandboxCacheKey` and process tools would be exposed.
- Have the error from these tools mention `sandboxCacheKey` when the resolved sandbox does support processes, so the failure mode points users to the fix.

The existing test at `dynamic-sandbox.test.ts:129` proves the API works *with* an external cache, but the misuse path isn't guarded.

packages/core/src/workspace/workspace.ts
```diff
@@ -723,10 +805,61 @@ export class Workspace<
...
+   * The workspace does not own resolver-returned sandboxes, so this only drops
+   * references from the workspace cache. Callers remain responsible for
+   * destroying any sandbox instances they created.
+   */
+  clearSandboxCache(cacheKey?: string): void {
+    if (cacheKey === undefined) {
+      this._sandboxKeyCache.clear();
+      return;
+    }
+    this._sandboxKeyCache.delete(cacheKey);
```
NikAiyer, line 889: **Doc clarification.**

The JSDoc says it clears "cached resolver-backed sandboxes", but it only clears the keyed `Map` — the per-`RequestContext` `WeakMap` entries are GC-managed and not cleared by this method. Worth a sentence noting that explicitly so the API doesn't look incomplete.

NikAiyer (commented):
A few follow-ups worth tightening before merge. Inline comments below.

**Test coverage to add alongside the inline items** (both fit the `dynamic-sandbox.test.ts` style):
- Resolver throws on first call → assert second call with the same `RequestContext` retries (or asserts the documented locked-in behavior).
- Workspace with `sandbox: () => new LocalSandbox(...)` and no `sandboxCacheKey` → spawn background process → call `get_process_output` with a fresh `RequestContext` → assert clear error (or success once fixed). The existing test at `dynamic-sandbox.test.ts:129` proves the API works *with* an external cache, but the misuse path isn't guarded.

NikAiyer (approved):

PR #16059, opened 2026-05-01
fix(memory): load read-only observational context

### LINKED ISSUE #16037: [BUG] Read-only Observational Memory skips context loading
## Describe the Bug

When Observational Memory is enabled in read-only mode, the input processor returns before starting an observation turn. That skips writes and observation side effects, but it also skips the normal context/history load path.

Observed behavior: in the read-only path, no OM-loaded stored context/history is added to the `MessageList`, although the non-read-only OM path would load it through `ObservationTurn.start()`.

Expected behavior: read-only Observational Memory should still read existing OM/memory context and skip only write/observation/persistence side effects.

This is related to, but distinct from, #14613. That issue discusses how OM and `lastMessages` interact. This report is specifically about the `readOnly` branch returning before the OM context load path runs.

Checked against `mastra-ai/mastra` `main` at commit `8f9e3822e3cb7a6cf8e070e6a01205cbc4dca701`.

## Steps To Reproduce

Source-level path:

1. Configure memory with Observational Memory enabled and `memoryConfig.readOnly: true` for a request.
2. Ensure the thread has existing stored messages/context.
3. Run an agent turn that relies on the automatic memory processors.
4. Observe that Observational Memory is present, so automatic `MessageHistory` is not added:
   - `packages/core/src/memory/memory.ts` checks Observational Memory and skips `MessageHistory` input when OM is enabled.
   - The same file also skips `MessageHistory` output when OM is enabled.
5. The OM input processor reads `readOnly` and returns before creating/starting the turn:
   - `packages/memory/src/processors/observational-memory/processor.ts` reads `memoryConfig.readOnly`.
   - It returns immediately in the read-only branch before `this.engine.beginTurn(...)`.
   - It therefore never reaches `await this.turn.start(this.memory)`.
6. `ObservationTurn.start()` is the path that calls `memory.getContext(...)` and adds historical messages to the `MessageList`.

Relevant source links:

- https://github.com/mastra-ai/mastra/blob/8f9e3822e3cb7a6cf8e070e6a01205cbc4dca701/packages/core/src/memory/memory.ts#L698-L704
- https://github.com/mastra-ai/mastra/blob/8f9e3822e3cb7a6cf8e070e6a01205cbc4dca701/packages/core/src/memory/memory.ts#L857-L863
- https://github.com/mastra-ai/mastra/blob/8f9e3822e3cb7a6cf8e070e6a01205cbc4dca701/packages/memory/src/processors/observational-memory/processor.ts#L123-L146
- https://github.com/mastra-ai/mastra/blob/8f9e3822e3cb7a6cf8e070e6a01205cbc4dca701/packages/memory/src/processors/observational-memory/processor.ts#L170-L182
- https://github.com/mastra-ai/mastra/blob/8f9e3822e3cb7a6cf8e070e6a01205cbc4dca701/packages/memory/src/processors/observational-memory/observation-turn/turn.ts#L122-L129

I do not have a standalone repo, but this now has a focused in-repo repro against `packages/memory`.

Repro shape:

1. Create an OM-enabled memory instance with a stored thread message.
2. Call `ObservationalMemoryProcessor.processInputStep()` with `memoryConfig.readOnly: true`.
3. Spy on the memory context provider / `getContext`.
4. Assert that `getContext` is called and the stored message is loaded into the `MessageList`.

Actual result on the current code:

```json
{
  "getContextCalls": 0,
  "loadedIds": []
}
```

Expected result:

```json
{
  "getContextCalls": 1,
  "loadedIds": ["stored-user-1"]
}
```

As a control, running the same focused repro without `memoryConfig.readOnly` loads the stored message correctly:

```json
{
  "getContextCalls": 1,
  "loadedIds": ["stored-user-1"]
}
```

Suggested regression test location: `packages/memory/src/processors/observational-memory/__tests__/observational-memory.test.ts` or a nearby OM processor test. The regression test should also assert that read-only mode still skips write-side observation, buffering, reflection, and persistence side effects.

## Expected Behavior

Read-only Observational Memory should still perform read-side context loading and only skip write-side observation/persistence behavior.

One possible fix would be to move the read-only branch after `turn.start()` has loaded context. Alternative fallbacks should be evaluated carefully to avoid duplicating history.

The output processor also has a read-only early return, which seems intentional for skipping persistence. The issue here is the input processor returning before the read-side context load.

## Environment Information

Source review against:

- Repo: `mastra-ai/mastra`
- Branch: `main`
- Commit: `8f9e3822e3cb7a6cf8e070e6a01205cbc4dca701`
- Relevant packages: `@mastra/core`, `@mastra/memory`

Focused repro reverified locally on 2026-05-01 against the current checkout.

## Verification

mbenhamd:
## Summary

Closes #16037.

Read-only Observational Memory returned before the normal OM context load path ran. Because OM also suppresses automatic `MessageHistory`, read-only turns could miss existing persisted thread context entirely.

This PR keeps read-only mode side-effect-free while still loading stored OM context into the `MessageList` before returning.

## Changes

- Adds a shared `loadMemoryContextMessages()` helper for loading persisted OM context into a `MessageList`.
- Uses that helper in the read-only input processor path before exiting.
- Reuses the same helper in normal turn startup so context loading stays centralized.
- Adds regression coverage for the read-only path, including the important negative assertions that observation, buffering, persistence, and progress side effects are still skipped.
- Adds a patch changeset for `@mastra/memory`.

## Validation

- Added focused regression coverage for read-only OM context loading and skipped write-side side effects.

## ELI5

The PR fixes a bug where read-only memory mode wasn't loading previously saved messages. It creates a shared helper function to load these saved messages that both read-only and normal modes can use, so old conversations are always restored while keeping read-only mode from making any changes.

## Problem

Read-only Observational Memory had an early-return path that prevented the normal context-loading flow from executing. This caused persisted thread context and message history to be omitted from the MessageList, and prevented automatic MessageHistory in read-only mode. The issue required that read-only mode continue to load context from storage while still skipping write-side behaviors (observation, buffering, persistence, and progress).

## Solution

Introduces a shared helper function `loadMemoryContextMessages()` that:
- Retrieves observational memory context for a given thread (optionally scoped by resource ID)
- Filters out system messages and appends remaining messages to the MessageList tagged as `'memory'`
- Returns the context object from the MemoryContextProvider

This helper is:
- Used in the read-only input processor path to load persisted context before returning
- Reused in normal turn startup to centralize context loading and eliminate duplication
- Paired with an `injectObservationContextMessages` helper to inject system and continuation messages

## Changes

**New file:**
- `packages/memory/src/processors/observational-memory/observation-turn/load-memory-context.ts`: New helper function for loading memory context messages

**Updated files:**
- `packages/memory/src/processors/observational-memory/processor.ts`: Read-only branch now loads and injects context; normal flow refactored to use shared injection helper
- `packages/memory/src/processors/observational-memory/observation-turn/turn.ts`: Turn startup now uses the shared helper instead of inline context-loading logic
- `packages/memory/src/processors/observational-memory/__tests__/observational-memory.test.ts`: Added comprehensive test suite for read-only mode covering context loading and verification that write-side effects remain skipped

**Metadata:**
- `.changeset/tricky-dryers-yell.md`: Patch version update for `@mastra/memory`

## Testing

Regression tests verify that in read-only mode:
- `memoryProvider.getContext` is called to load persisted context
- Stored user messages are loaded into the MessageList
- An `om-continuation` message is added when observational context exists
- No write-side side effects occur (`beginTurn`, `observe`, `buffer`, `persistMessages`, `emitProgress` are not called)
- `state.__omTurn` remains undefined
- No observational continuation is added when system context is not observational-memory-related

## Closes

Issue #16037


packages/memory/src/processors/observational-memory/processor.ts
```diff
@@ -140,8 +141,14 @@ export class ObservationalMemoryProcessor implements Processor<'observational-me
...
-      // ── Read-only fast path: skip turn creation and observation lifecycle ──
+      // ── Read-only path: load existing context, skip observation lifecycle ──
       if (readOnly) {
+        await loadMemoryContextMessages({
+          memory: this.memory,
+          messageList,
+          threadId,
+          resourceId,
+        });
         return messageList;
```
TylerBarnes, line 206: instead of this, perhaps we should let the regular logic run, but just guard to prevent persistence? the current logic here isn't quite right because (unless I'm reading it wrong) it wont load the observations, just messages, right?

TylerBarnes (changes requested):

packages/memory/src/processors/observational-memory/processor.ts
mbenhamd: Indeed, I focused on getting the stored messages back into the list, but this still skips the part that actually injects the observation context.

I'll change this so read-only uses the normal context preparation/injection path and only skips the write/persistence side effects.

abhiaiyer91 (approved):

PR #16073, opened 2026-05-01
fix(core): fix message part ordering in buildMessagesFromChunks to use first-delta position

### LINKED ISSUE #15914: [BUG] buildMessagesFromChunk function uses streaming order while some internal codes expect semantic order

### LINKED ISSUE #16007: Message persistence ordering issue
This issue was created from Discord post 1499489631811403927:

[Open in Discord](https://discord.com/channels/1309558646228779139/1499489631811403927)

Hey all!! I’m debugging what looks like a Mastra + AI SDK v6 message persistence ordering issue and wanted to check whether this is expected or a known regression.

Setup
Recently mastra related packages: 

@mastra/ai-sdk: 1.3.1 → 1.4.1
@mastra/core:   1.14.0 → 1.29.0
@mastra/memory: 1.13.1 → 1.17.2
@mastra/pg:     1.8.6 → 1.9.2

mastra@1.7.2

ai@6.0.158

@ai-sdk/react@3.0.148

Postgres via Neon

Streaming with toAISdkStream(stream, { from: "agent", version: "v6", sendReasoning: true })

During streaming, reasoning and tool calls appear in the correct trace order. After the assistant message is persisted and reloaded from storage, content.parts often comes back reordered, with tool invocations saved before earlier reasoning parts.

Expected/live order:
reasoning →  tool:read  →  tool: grep  → reasoning → text

Saved DB order:
tool:read →  tool:grep  →  tool:grep  →  reasoning  →  step-start  →  reasoning  →  text

I compared main vs development:

Older Mastra versions in production mostly persist recent messages correctly.

Newer Mastra packages in development often persist tools first and reasoning later.

Updating to the even latest Mastra packages did not resolve it.

The app-level chat route and model code are unchanged between branches, so this seems more like stream-to-memory/storage assembly than frontend rendering.

Questions

Should persisted mastra_messages.content.parts preserve the exact stream order of reasoning and tool-invocation parts?

What is the recommended way to preserve reasoning + tool trace order with AI SDK v6?

Should I switch to format: "aisdk" / stream.toUIMessageStreamResponse() instead of manually wrapping toAISdkStream() in createUIMessageStreamResponse()?

Could this be a known regression related to MessageList or memory persistence for reasoning models?

CalebBarnes:
## Problem

`buildMessagesFromChunks` emitted text and reasoning parts at `*-end` time, not when they first appeared in the stream. When spans overlapped with tool calls (e.g. `text-start` → `tool-call` → `text-end`), the tool call would appear before the text part in the final message — breaking step-start insertion logic and agent loop behavior.

## Solution

Refactored part emission to use **first-delta ordering**: parts are pushed into the `parts` array when their first delta arrives (not on `*-start` or `*-end`), then mutated in-place as subsequent deltas arrive. This preserves the true stream order without accumulating deltas into intermediate arrays.

Key changes:

- **First-delta positioning** — a part is created and pushed to the array on the first `text-delta` or `reasoning-delta`, establishing its position relative to tool calls and other parts
- **In-place mutation** — subsequent deltas append directly to the part via a reference map (`textRefs`/`reasoningRefs`), avoiding `deltas[]` accumulation and `join("")` at the end
- **Metadata maps** — `*-start` chunks only stash `providerMetadata` in lightweight maps (`textMeta`/`reasoningMeta`); no part is created until content actually arrives
- **Redacted reasoning** — still pushed immediately on `reasoning-start` since redacted spans have no deltas
- **Empty reasoning** — emitted on `reasoning-end` if no deltas arrived, preserving OpenAI `item_reference` requirements (#9005)
- **Empty text cleanup** — text parts where every delta was an empty string `''` are filtered out at the end

## Testing

- Added regression test for interleaved `text-start` → `reasoning-start` → `reasoning-delta` → `text-delta` ordering (from #15914)
- Updated existing interleaved text/tool-call test to verify first-delta ordering
- All 29 tests pass

## ELI5

When the model streams bits of text, thoughts, or tool calls, this change makes sure each saved message piece appears when its first real content shows up so the saved order matches what happened live.

## Overview

Refactors buildMessagesFromChunks to use first-delta ordering: text and reasoning parts are created/pushed when their first content delta arrives (not at span start or end), preserving true stream-relative ordering with tool calls and other parts. Provider metadata from span-start is stashed and applied when content appears; redacted reasoning is still emitted immediately; empty reasoning parts are emitted when required to satisfy OpenAI item_reference semantics; purely-empty text spans are filtered out except when they carry providerMetadata.

## Changes

- First-delta positioning: create and push text/reasoning parts on first text-delta/reasoning-delta so part ordering reflects actual content arrival (#15914).
- In-place mutation: subsequent deltas update the already-pushed part via reference maps (textRefs/reasoningRefs) instead of buffering and joining at end.
- *-start metadata handling: text-start and non-redacted reasoning-start stash providerMetadata in lightweight meta maps (textMeta/reasoningMeta) without creating parts until content arrives.
- Redacted reasoning: redacted reasoning spans are emitted immediately on reasoning-start (since no deltas will arrive).
- Empty reasoning: reasoning-end emits an empty reasoning part if no deltas arrived to satisfy downstream item_reference requirements (consistent with #9005 rationale).
- Empty text filtering: drop text parts whose deltas were all empty strings while preserving empty text parts that carry providerMetadata.
- Stream-end/cleanup: text-end cleans refs/meta without forcing a flush; unclosed reasoning spans with no deltas are emitted as empty reasoning at stream end.
- Step-start insertion: step-start markers are inserted after filtering so they remain correctly ordered relative to actual content.

## Tests

- Added regression test for interleaved reasoning/text ordering and updated tests for interleaved text/tool-call/step-start scenarios.
- Updated inline snapshot expectations for mixed text/reasoning blocks in options tests.
- Test run: 29 tests pass.

## Changesets

- Adds a patch changeset for @mastra/core documenting the fix: streaming message parts (text, reasoning, tool calls) are now ordered by actual arrival sequence.

## Issues Fixed

- Fixes #16007 (message persistence ordering) and #15914 (buildMessagesFromChunks ordering bug).

Fixes #16007
Fixes #15914


packages/core/src/loop/workflows/agentic-execution/build-messages-from-chunks.ts
```diff
@@ -286,44 +278,36 @@ export function buildMessagesFromChunks({
...
+  // Unclosed text spans that had deltas are already in `parts`.
+  // Clean up undefined providerMetadata on any that are still open.
+  for (const [, ref] of textRefs) {
+    if (!ref.providerMetadata) {
+      delete ref.providerMetadata;
     }
   }
 
+  // Remove text parts that ended up empty (e.g. spans where every delta was '')
+  const nonEmptyParts = parts.filter(p => !(p.type === 'text' && (p as any).text === ''));
```
TylerBarnes, line 305: could there be other empty parts? like reasoning

packages/core/src/loop/workflows/agentic-execution/build-messages-from-chunks.test.ts
```diff
@@ -304,6 +305,52 @@ describe('buildMessagesFromChunks', () => {
...
+      },
+      {
+        type: 'tool-result',
+        payload: { toolCallId: 'tc1', toolName: 'myTool', args: {}, result: 'ok' },
+      },
+      // Text span t1 closes after the tool-call
+      { type: 'text-end', payload: { id: 't1' } },
+      { type: 'text-start', payload: { id: 't2' } },
+      { type: 'text-delta', payload: { id: 't2', text: 'After tool' } },
+      { type: 'text-end', payload: { id: 't2' } },
```
TylerBarnes: can you add a test for if we get two reasoning starts with different ids, no reasoning-end, and tool call in between?

TylerBarnes: nvm that would be multiple requests

packages/core/src/loop/workflows/agentic-execution/build-messages-from-chunks.ts
CalebBarnes: We kept empty reasoning parts because of https://github.com/mastra-ai/mastra/issues/9005 where previously we stripped empty reasoning part and lost the provider metadata on that message. Maybe we should only strip out empty parts if they also have no providerMetadata?

CalebBarnes:
Addressed Tyler's feedback on line 303: updated the empty text filter to also preserve empty text parts that carry `providerMetadata` (same rationale as #9005 for reasoning parts).

packages/core/src/loop/workflows/agentic-execution/build-messages-from-chunks.ts
```diff
@@ -286,44 +278,39 @@ export function buildMessagesFromChunks({
...
+    if (!ref.providerMetadata) {
+      delete ref.providerMetadata;
     }
   }
 
+  // Remove text parts that ended up empty (e.g. spans where every delta was ''),
+  // but keep them if they carry providerMetadata (same rationale as #9005 for reasoning).
+  const nonEmptyParts = parts.filter(
+    p => !(p.type === 'text' && (p as any).text === '' && !(p as any).providerMetadata),
+  );
```
TylerBarnes: Oh, isn't this about reasoning parts not text?

TylerBarnes: I think we need a separate condition for each, not to check if text parts don't have metadata

CalebBarnes: reverted that `!(p as any).providerMetadata)` part, 
But the empty reasoning parts are already in the parts array at this point and not being filtered out at all. 

Do you think we should only keep empty reasoning parts if they have providerMetadata?

TylerBarnes: wasn't it the other way around? filter out empty reasoning unless it has provider metadata

CalebBarnes: Before these changes (on main): Empty reasoning was kept unconditionally, no `providerMetadata` check. The original fix for #9005 (PR #10585) also kept them unconditionally. It just removed the `if (reasoningDeltas.length > 0)` guard with no `providerMetadata` check. 

So the existing behaviour, both in `build-messages-from-chunks.ts` and the original `llm-execution-step.ts` fix (when we were constructing messages mid stream) has always kept empty reasoning parts without checking for `providerMetadata`. 

Maybe we _should_ check for that and only keep empty ones if they have `providerMetadata` but I think that's a behavioural change and outside of this PR.

TylerBarnes: ah ok cool

TylerBarnes (approved):

PR #16078, opened 2026-05-01
fix(core): merge tool invocation provider metadata

### LINKED ISSUE #16075: MessageList.updateToolInvocation() drops existing provider metadata when tool result adds new metada
This issue was created from Discord post 1499837717922451556:

[Open in Discord](https://discord.com/channels/1309558646228779139/1499837717922451556)

**Summary**
In @mastra/core@1.30.0, MessageList.updateToolInvocation() replaces a tool-call part’s existing providerMetadata when the tool result includes its own providerMetadata. This drops provider-specific state required by later model calls, notably Gemini vertex.thoughtSignature.

**Impact**
Gemini thinking models can fail on the next tool-call turn with:

```Function call is missing a thought_signature in functionCall parts.```
This happens because the original tool-call metadata:

```{ vertex: { thoughtSignature: "..." } }```
is replaced by result-side metadata such as:

```{ mastra: { modelOutput: ... } }```
instead of being merged.

**Minimal Repro**
Create a MessageList with an assistant tool-invocation part containing:
```providerMetadata: {
  vertex: { thoughtSignature: "sig-from-gemini" },
}```
Call messageList.updateToolInvocation() for the same toolCallId with result-side metadata:
```providerMetadata: {
  mastra: { modelOutput: { /* ... */ } },
}```
Inspect the updated tool-invocation part.
**Actual**
providerMetadata.vertex.thoughtSignature is gone.

**Expected**
Metadata namespaces should be merged:

```{
  vertex: { thoughtSignature: "sig-from-gemini" },
  mastra: { modelOutput: { /* ... */ } },
}```
**Suggested Fix**
In MessageList.updateToolInvocation(), merge original and incoming providerMetadata by namespace instead of only preserving the original when incoming metadata is undefined.

```const mergedProviderMetadata =
  originalPart.providerMetadata !== undefined || inputPart.providerMetadata !== undefined
    ? {
        ...(originalPart.providerMetadata ?? {}),
        ...(inputPart.providerMetadata ?? {}),
      }
    : undefined;```
Then set providerMetadata to mergedProviderMetadata when defined.

### PR DESCRIPTION
**Pull request by roaminro**
Fixes #16075.

This fixes tool results replacing provider metadata from the original tool call when both parts include metadata.

Before, a result with its own provider metadata could drop unrelated metadata from the original call:

```ts
// original call
{ providerMetadata: { vertex: { thoughtSignature: '...' } } }

// result
{ providerMetadata: { mastra: { modelOutput: true } } }

// final metadata only kept mastra
{ providerMetadata: { mastra: { modelOutput: true } } }
```

After, metadata is merged by provider namespace, so unrelated provider data is preserved:

```ts
{
  providerMetadata: {
    vertex: { thoughtSignature: '...' },
    mastra: { modelOutput: true },
  },
}
```

Tested with:

```bash
pnpm --filter ./packages/core test src/agent/message-list/tests/update-tool-invocation.test.ts
pnpm --filter ./packages/core check
```

## ELI5 (Explain Like I'm 5)

If a tool call and the tool's result each had labels from different providers, we used to throw away the original labels and only keep the result's. Now we combine the labels so nothing important from the original call gets lost.

---

## Summary of Changes

- Fix MessageList.updateToolInvocation to merge providerMetadata by provider namespace instead of replacing it. The updated logic constructs mergedProviderMetadata by spreading the original part.providerMetadata and the incoming result's providerMetadata (when present) so namespaces from the original call are preserved alongside namespaces added by the result.
- Update tests to assert the merged providerMetadata behavior.
- Extend serializeRequestContext to accept RequestContext | Record<string, any> | Map<string, unknown> and ensure JSON-safe serialization for non-serializable values; add tests for these cases.
- Add a changeset marking a patch release for @mastra/core.

### Files Changed

- .changeset/quiet-clowns-worry.md
  - New changeset: patch for @mastra/core describing the fix for dropped provider metadata.

- packages/core/src/agent/message-list/message-list.ts
  - updateToolInvocation now builds mergedProviderMetadata by spreading original and incoming providerMetadata and assigns it to the updated tool-invocation part, preserving both provider namespaces. (+9 / -4)

- packages/core/src/agent/message-list/tests/update-tool-invocation.test.ts
  - Tests updated to expect merged providerMetadata (original metadata preserved and combined with result metadata). (+5 / -2)

- packages/core/src/workflows/default.ts
  - serializeRequestContext signature widened to accept RequestContext | Record<string, any> | Map<string, unknown> and implementation updated to route serialization through RequestContext.toJSON() when available; Map and plain-record inputs are converted into RequestContext before serialization. (+11 / -8)

- packages/core/src/workflows/default.test.ts
  - New tests verifying serializeRequestContext handles plain objects (including circular refs and function-valued fields) and Map inputs by producing JSON-safe serialized output that retains only JSON-safe fields (e.g., userId). (+30 / 0)

### Behavior Example

Before: provider namespaces on the original tool call (e.g., vertex) could be lost when the result provided its own namespace (e.g., mastra).  
After: merged providerMetadata retains both vertex and mastra namespaces so provider-specific state (e.g., thought signatures) is not lost.

---

## Tests / Verification

Run:
- pnpm --filter ./packages/core test src/agent/message-list/tests/update-tool-invocation.test.ts

### DIFF
--- a/packages/core/src/workflows/default.test.ts
+++ b/packages/core/src/workflows/default.test.ts
@@ -625,36 +625,6 @@ describe('DefaultExecutionEngine.deserializeRequestContext', () => {
     expect(serialized).toEqual({ userId: 'user-123' });
   });
 
-  it('should serialize plain object request context values', () => {
-    const engine = new TestableExecutionEngine({ mastra: undefined });
-    const circular: any = { name: 'service' };
-    circular.self = circular;
-    const plainObj = {
-      userId: 'user-123',
-      progressEmitter: () => undefined,
-      service: circular,
-      forEach: () => undefined,
-    };
-
-    const serialized = engine.serializeRequestContext(plainObj);
-
-    expect(() => JSON.stringify(serialized)).not.toThrow();
-    expect(serialized).toEqual({ userId: 'user-123' });
-  });
-
-  it('should serialize Map request context values', () => {
-    const engine = new TestableExecutionEngine({ mastra: undefined });
-    const requestContext = new Map<string, unknown>([
-      ['userId', 'user-123'],
-      ['progressEmitter', () => undefined],
-    ]);
-
-    const serialized = engine.serializeRequestContext(requestContext);
-
-    expect(() => JSON.stringify(serialized)).not.toThrow();
-    expect(serialized).toEqual({ userId: 'user-123' });
-  });
-
   it('should return a RequestContext instance with all entries from the plain object', () => {
     const engine = new TestableExecutionEngine({ mastra: undefined });
     const plainObj = { userId: 'user-123', tenantId: 'tenant-456', nested: { flag: true } };
--- a/packages/core/src/workflows/default.test.ts
+++ b/packages/core/src/workflows/default.test.ts
@@ -625,6 +625,18 @@ describe('DefaultExecutionEngine.deserializeRequestContext', () => {
     expect(serialized).toEqual({ userId: 'user-123' });
   });
 
+  it('should serialize plain object request context values', () => {
+    const engine = new TestableExecutionEngine({ mastra: undefined });
+    const circular: any = { name: 'service' };
+    circular.self = circular;
+    const plainObj = { userId: 'user-123', progressEmitter: () => undefined, service: circular };
+
+    const serialized = engine.serializeRequestContext(plainObj);
+
+    expect(() => JSON.stringify(serialized)).not.toThrow();
+    expect(serialized).toEqual({ userId: 'user-123' });
+  });
+
   it('should return a RequestContext instance with all entries from the plain object', () => {
     const engine = new TestableExecutionEngine({ mastra: undefined });
     const plainObj = { userId: 'user-123', tenantId: 'tenant-456', nested: { flag: true } };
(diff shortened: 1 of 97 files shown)


packages/core/src/workflows/default.ts
```diff
@@ -614,15 +614,18 @@ export class DefaultExecutionEngine extends ExecutionEngine {
...
-    requestContext.forEach((value, key) => {
-      obj[key] = value;
-    });
-    return obj;
+
+    if (requestContext instanceof Map) {
+      return new RequestContext<unknown>(Array.from(requestContext.entries())).toJSON();
+    }
+
+    return new RequestContext(Object.entries(requestContext)).toJSON();
```
TylerBarnes: did this get in here accidentally?

packages/core/src/workflows/default.test.ts
```diff
@@ -625,6 +625,36 @@ describe('DefaultExecutionEngine.deserializeRequestContext', () => {
     expect(serialized).toEqual({ userId: 'user-123' });
   });
 
+  it('should serialize plain object request context values', () => {
```
TylerBarnes: and this

TylerBarnes (approved):

CalebBarnes (approved):

PR #16084, opened 2026-05-01
feat(server): add ZodOpenAPIRouteConfig — Zod-first OpenAPI for custom routes

### LINKED ISSUE #5454: [FEATURE] Hono Zod OpenAPI Support
### Problem Statement

I'd like to improve Custom API Routes experience of server deployment 

### Proposed Solution

Simply say, 

1. mastra/core server supports zod-openapi plus to hono-openapi

---
Mastra is using hono for providing api server and workflows can receive zod schema input/output. however, custom api route doesn't support zod openapi schema, yet I think. 
https://mastra.ai/en/docs/deployment/custom-api-routes

Hono already supports zod-based openapi declaration. It's like below 
https://hono.dev/examples/zod-openapi

```typescript 
const route = createRoute({
  method: 'get',
  path: '/users/{id}',
  request: {
    params: ParamsSchema,
  },
  responses: {
    200: {
      content: {
        'application/json': {
          schema: UserSchema,
        },
      },
      description: 'Retrieve the user',
    },
  },
})
```

but, mastra server support only hono-openapi, not zod-openapi. hono-openapi is more foundational package prior to validator-based ones.
https://github.com/mastra-ai/mastra/blob/main/packages/core/src/server/types.ts#L10

### Component

Other

### Alternatives Considered

There are several ways 

- use workflow/agent api directory
  - that means we don't provide custom api routes. 
  - it's straightforward, but it's not bit easy to use because workflow and agent interface doesn't provide concrete workflow input zod schema, just providing 
  - we cannot handle ai api interface as more deterministic
- integrate mastra package into hono, nextjs and so on 
  - that means we use other platform as a agent provider server. 
  - this is also effective, but sometimes it's not feasible due to organization structure or something business affairs

Image
Image

### Example Use Case

For example, 

- The case that we want to call specific workflow(s) directory with enjoying zod-based api clients
- The case we want to separate api endpoints `/workflows` into more specific endpoints due to access control

### Additional Context

_No response_

### Verification

piiiico:
## Summary

Closes #5454.

`registerApiRoute()` now accepts a Zod-first OpenAPI config (inspired by `@hono/zod-openapi`'s `createRoute` API) as an alternative to the raw `DescribeRouteOptions` format. Schemas are converted to JSON Schema automatically when building the OpenAPI spec.

**Before** (raw `DescribeRouteOptions` — still works):
```typescript
registerApiRoute('/users/:id', {
  method: 'GET',
  openapi: {
    summary: 'Get user',
    parameters: [{ name: 'id', in: 'path', required: true, schema: { type: 'string' } }],
    responses: { 200: { description: 'User' } },
  },
  handler: async (c) => c.json({ id: c.req.param('id') }),
});
```

**After** (new `ZodOpenAPIRouteConfig` — Zod schemas inline):
```typescript
import { z } from 'zod';

registerApiRoute('/users/:id', {
  method: 'GET',
  openapi: {
    summary: 'Get user',
    tags: ['users'],
    request: {
      params: z.object({ id: z.string().describe('User ID') }),
      query: z.object({ include: z.string().optional() }),
    },
    responses: {
      200: {
        description: 'User found',
        content: {
          'application/json': {
            schema: z.object({ id: z.string(), name: z.string() }),
          },
        },
      },
      404: { description: 'User not found' },
    },
  },
  handler: async (c) => {
    const { id } = c.req.param();
    return c.json({ id, name: 'Alice' });
  },
});
```

## Changes

- **`packages/core/src/server/types.ts`**: adds `ZodOpenAPIRouteConfig` type; `ApiRoute.openapi` is now `DescribeRouteOptions | ZodOpenAPIRouteConfig` (fully backwards-compatible)
- **`packages/core/src/server/index.ts`**: exports `ZodOpenAPIRouteConfig`; `registerApiRoute` accepts either openapi format
- **`packages/server/src/server/server-adapter/openapi-utils.ts`**:
  - `isZodOpenAPIRouteConfig()` type guard (detects `request` key — absent from `DescribeRouteOptions`)
  - `isZodSchema()` helper covering both Zod v3 (`_def`) and v4 (`_zod`)
  - `convertZodRouteConfigToOperation()` converts `request.params/query/body` + `responses` to JSON Schema
  - `convertCustomRoutesToOpenAPIPaths` dispatches on format
  - Fix: the existing `DescribeRouteOptions` path now uses `isZodSchema()` instead of checking `_def` only, so Zod v4 schemas work there too
- **`packages/server/src/server/server-adapter/openapi-utils.test.ts`**: new tests for the new format (path params, query params, request body, response schemas, `hide` flag, mixed routes, legacy format unchanged)

## Notes

- No new dependencies — uses the already-present `zodToJsonSchema` from `@mastra/core/utils/zod-to-json`
- The discriminator between formats is the `request` key: `DescribeRouteOptions` (which extends `OpenAPIV3_1.OperationObject`) never has a `request` property; `ZodOpenAPIRouteConfig` always does (even if `request: {}`)
- Both Zod v3 and v4 schemas are handled via the existing `zodToJsonSchema` adapter

## Test plan

- [ ] `pnpm test --filter @mastra/server` — new `openapi-utils.test.ts` tests pass
- [ ] Existing custom route OpenAPI tests unaffected (backwards compat)
- [ ] Manual: define a route with `ZodOpenAPIRouteConfig`, hit `/openapi.json`, confirm path params / query params / body / responses appear correctly

## ELI5

You can now describe custom API routes using Zod schemas (for params, query, body, and responses) and the server will auto-convert them into OpenAPI docs—while the old format still works. The PR also fixes edge-cases so operation IDs aren’t lost and prevents prototype-pollution in the generated OpenAPI paths.

## Overview

Adds Zod-first OpenAPI support for custom routes via a new ZodOpenAPIRouteConfig type and conversion logic that turns Zod schemas into JSON Schema (using the existing zodToJsonSchema) when building the OpenAPI spec. The change is backwards-compatible: the existing DescribeRouteOptions format remains supported.

### Types & Exports
- packages/core/src/server/types.ts
  - New exported type: ZodOpenAPIRouteConfig (OpenAPI metadata + optional request { params, query, body } and responses keyed by status).
  - ApiRoute.openapi now accepts DescribeRouteOptions | ZodOpenAPIRouteConfig.
- packages/core/src/server/index.ts
  - Exports ZodOpenAPIRouteConfig and updates RegisterApiRouteOptions to allow the new format.

### OpenAPI Conversion & Helpers
- packages/server/src/server/server-adapter/openapi-utils.ts
  - Added isZodOpenAPIRouteConfig type-guard and convertZodRouteConfigToOperation to convert Zod request/response schemas to OpenAPI operations (via zodToJsonSchema).
  - Added isZodSchema helper that recognizes Zod v3 (_def) and v4 (_zod).
  - convertCustomRoutesToOpenAPIPaths now dispatches between Zod and legacy DescribeRouteOptions conversion.
  - Broadened Zod-config detection to handle configs missing request (3-step detection so entries like { operationId, responses } are treated as Zod configs).
  - Replaced plain object maps with Object.create(null) for the top-level paths container and per-path objects to prevent prototype pollution.
  - Legacy path generation updated to use isZodSchema so Zod v4 schemas work in DescribeRouteOptions flows.

### Tests
- packages/server/src/server/server-adapter/openapi-utils.test.ts
  - New tests covering Zod format conversions (path params, query, body, responses), hide flag, mixed routes, routes with metadata-only, and backwards compatibility.
  - Regression tests for operationId-preservation and prototype-pollution prevention.

### Release notes / changeset
- .changeset/fix-zod-openapi-detection-and-proto-pollution.md updates for @mastra/core and @mastra/server describing the detection broadening and prototype-pollution fix.

## Impact

- Backwards-compatible: existing DescribeRouteOptions users unaffected.
- No new dependencies.
- Developer experience improved: Zod-first route schemas produce deterministic OpenAPI output and can be used alongside existing formats.
- Fixes: ensures operationId and other metadata aren’t lost for Zod-shaped configs without request, and defends against prototype-pollution in OpenAPI path keys.

### DIFF
--- a/.changeset/fix-zod-openapi-detection-and-proto-pollution.md
+++ b/.changeset/fix-zod-openapi-detection-and-proto-pollution.md
@@ -1,0 +1,17 @@
+---
+'@mastra/core': patch
+'@mastra/server': patch
+---
+
+fix(server): broaden ZodOpenAPIRouteConfig detection and prevent prototype pollution
+
+`isZodOpenAPIRouteConfig` now recognises configs without a `request` field
+(e.g. `{ operationId, responses }`) by checking for the absence of
+`DescribeRouteOptions`-specific fields (`parameters[]` / `requestBody`).
+Previously those configs silently fell through to the legacy converter and
+lost `operationId` and any other Zod-specific processing.
+
+`generateOpenAPIDocument` and `convertCustomRoutesToOpenAPIPaths` now use
+`Object.create(null)` for both the top-level `paths` map and per-path
+objects, eliminating prototype-pollution risk when a route path is
+`__proto__`, `constructor`, or `toString`.
--- a/packages/core/src/server/index.ts
+++ b/packages/core/src/server/index.ts
@@ -3,7 +3,7 @@ import type { DescribeRouteOptions } from 'hono-openapi';
 import { MastraError, ErrorDomain, ErrorCategory } from '../error';
 import type { Mastra } from '../mastra';
 import type { RequestContext } from '../request-context';
-import type { ApiRoute, MastraAuthConfig, Methods } from './types';
+import type { ApiRoute, MastraAuthConfig, Methods, ZodOpenAPIRouteConfig } from './types';
 
 export type {
   MastraAuthConfig,
@@ -13,6 +13,7 @@ export type {
   ValidationErrorContext,
   ValidationErrorResponse,
   ValidationErrorHook,
+  ZodOpenAPIRouteConfig,
 } from './types';
 export { MastraAuthProvider } from './auth';
 export type { MastraAuthProviderOptions } from './auth';
@@ -40,7 +41,7 @@ type CustomRouteVariables = {
 
 type RegisterApiRouteOptions<P extends string> = {
   method: Methods;
-  openapi?: DescribeRouteOptions;
+  openapi?: DescribeRouteOptions | ZodOpenAPIRouteConfig;
   handler?: Handler<
     {
       Variables: CustomRouteVariables;
--- a/packages/core/src/server/types.ts
+++ b/packages/core/src/server/types.ts
@@ -9,13 +9,120 @@ import type { MastraAuthProvider } from './auth';
 
 export type Methods = 'GET' | 'POST' | 'PUT' | 'DELETE' | 'PATCH' | 'ALL';
 
+/**
+ * Zod-based OpenAPI route configuration.
+ * An alternative to `DescribeRouteOptions` that lets you define request/response schemas
+ * using Zod schemas directly (inspired by @hono/zod-openapi's `createRoute` API).
+ *
+ * @example
+ * ```typescript
+ * import { z } from 'zod';
+ * import { registerApiRoute } from '@mastra/core/server';
+ *
+ * export const getUserRoute = registerApiRoute('/users/:id', {
+ *   method: 'GET',
+ *   openapi: {
+ *     summary: 'Get user by ID',
+ *     tags: ['users'],
+ *     request: {
+ *       params: z.object({ id: z.string() }),
+ *     },
+ *     responses: {
+ *       200: {
+ *         description: 'User found',
+ *         content: {
+ *           'application/json': {
+ *             schema: z.object({ id: z.string(), name: z.string() }),
+ *           },
+ *         },
+ *       },
+ *       404: { description: 'User not found' },
+ *     },
+ *   },
+ *   handler: async (c) => {
+ *     const { id } = c.req.param();
+ *     return c.json({ id, name: 'Alice' });
+ *   },
+ * });
+ * ```
+ */
+export type ZodOpenAPIRouteConfig = {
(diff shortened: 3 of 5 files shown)


piiiico:
Fixed both issues in f40cda3:

**Issue 1 — operationId lost on ZodOpenAPIRouteConfig without request field**

Broadened `isZodOpenAPIRouteConfig` with a 3-step detection strategy:
1. `request` present → ZodOpenAPIRouteConfig (unambiguous)
2. `parameters[]` or `requestBody` present → DescribeRouteOptions
3. `responses` present and neither of the above → ZodOpenAPIRouteConfig

Configs like `{ operationId, responses }` now route through `convertZodRouteConfigToOperation` correctly.

**Issue 2 — prototype pollution via paths object**

`generateOpenAPIDocument` and `convertCustomRoutesToOpenAPIPaths` now use `Object.create(null)` for both the top-level `paths` map and per-path objects.

Added regression tests for both cases.

intojhanurag:
see the latest comment on issue

PR #16088, opened 2026-05-01
feat(core): add request-aware tool search filtering

### LINKED ISSUE #16045: [FEATURE] Add request-context-aware tool filter hook to ToolSearchProcessor
## Description

`ToolSearchProcessor` is useful for large tool catalogs, but production apps often need to filter the searchable/loadable tool catalog per request.

Today, `ToolSearchProcessorOptions` accepts:

- `tools`
- `search.topK`
- `search.minScore`
- `ttl`

The processor can preserve existing per-step tools and loaded tools, but there does not appear to be a public hook for request-context-aware tool gating inside `ToolSearchProcessor` itself.

This makes apps wrap `ToolSearchProcessor` when tool availability depends on runtime context such as tenant, plan, entitlement, workspace mode, safety policy, or feature flags.

## Current Behavior

`ToolSearchProcessor` builds its searchable catalog at construction time.

At runtime, `processInputStep(...)` receives `requestContext`, but the processor only uses it to resolve the thread id.

There is no hook to say:

- this tool should not appear in `search_tools` results for this request
- this tool should not be loadable through `load_tool` for this request
- this already-loaded tool should be removed for this request if policy no longer allows it

Apps can avoid registering a tool globally, but that does not help when the same agent/tool catalog should be filtered differently per request.

Mastra can already provide tools dynamically at the agent level, but that workaround requires constructing or wrapping ToolSearch behavior outside the processor and can complicate thread-loaded state. A first-class hook inside `ToolSearchProcessor` would let apps keep the built-in search/load state machine while enforcing request-time policy.

## Proposed Behavior

Add an optional request-context-aware filter hook to `ToolSearchProcessor`.

Example shape:

```ts
const toolSearch = new ToolSearchProcessor({
  tools,
  search: { topK: 8, minScore: 0.05 },
  filterTool: ({ toolName, tool, requestContext, phase }) => {
    if (phase === "search" || phase === "load" || phase === "active") {
      return canUseTool(requestContext, toolName);
    }

    return true;
  },
});
```

Possible type:

```ts
type ToolSearchFilterPhase = "search" | "load" | "active";

type ToolSearchFilterArgs = {
  toolName: string;
  tool: Tool<any, any>;
  requestContext?: RequestContext;
  phase: ToolSearchFilterPhase;
};

type ToolSearchProcessorOptions = {
  tools: Record<string, Tool<any, any>>;
  search?: {
    topK?: number;
    minScore?: number;
  };
  ttl?: number;
  filterTool?: (args: ToolSearchFilterArgs) => boolean | Promise<boolean>;
};
```

Expected behavior:

- `phase: "search"` filters results returned by `search_tools`
- `phase: "load"` blocks `load_tool` from loading disallowed tools
- `phase: "active"` removes already-loaded tools that are no longer allowed for the current request
- existing tools passed through `args.tools` can continue to be preserved by default unless the hook explicitly applies to them

The exact API shape may differ, but the important capability is a per-request policy hook with access to `requestContext`.

## Why This Matters

Some tool catalogs are stable at the agent level but dynamic at the request level.

Examples:

- plan-tier or entitlement-gated tools
- tenant-specific tool availability
- workspace mode differences
- feature-flagged tools
- policy-gated integrations
- tools that should be searchable in one request but hidden in another

Without a hook, apps need to wrap `ToolSearchProcessor` and re-implement parts of `search_tools` / `load_tool` behavior to enforce policy consistently.

## Non-goals

This is not asking for a full category policy system.

This is also not asking for:

- `toModelOutput` support for dynamically loaded tools
- `requireApproval` preservation
- preserving externally provided tools
- always-loaded tool support

Those appear to already be handled by current Mastra behavior.

The request is specifically for runtime, request-context-aware filtering of ToolSearch search/load/active behavior.

## Compatibility

This can be additive.

If no `filterTool` hook is provided, current behavior should remain unchanged.

## Verified

Checked against latest `origin/main`:

```text
1c2dda805fbfccc0abf55d4cb20cc34402dc3f0c
```

`ToolSearchProcessorOptions` currently exposes `tools`, `search`, and `ttl`, and `requestContext` is only used for thread id resolution.

mbenhamd:
Adds a request-aware filterTool hook to ToolSearchProcessor so apps can hide tools from search results, block tool loading, and hide already-loaded tools for the current request. This keeps the existing search/load state machine while letting runtime policy use requestContext.

Example:

```ts
const toolSearch = new ToolSearchProcessor({
  tools,
  filterTool: ({ toolName, requestContext, phase }) => {
    const plan = requestContext?.get('plan');
    return plan === 'pro' || !toolName.startsWith('premium_');
  },
});
```

Fixes #16045

## ELI5

This PR adds a per-request "gatekeeper" to ToolSearchProcessor so apps can decide, at request time, which tools an agent can see, load, or use (for example, to enforce a user's plan) without rewriting the processor.

## Overview

Adds an optional, request-context-aware filter hook to ToolSearchProcessor that can allow or deny tools at three phases: search (filter search results), load (block load attempts), and active (hide already-loaded tools for the current request). The change is additive and preserves existing behavior when the hook is not provided; meta-tools (search_tools, load_tool) remain available and thread state is preserved even when per-request visibility differs.

### New Public Types
- ToolSearchFilterPhase = 'search' | 'load' | 'active'
- ToolSearchFilterArgs = { toolName: string; tool?: Tool; requestContext?: RequestContext; phase: ToolSearchFilterPhase }

### ToolSearchProcessor Enhancements
- New option: filterTool?: (args: ToolSearchFilterArgs) => boolean | Promise<boolean>
- Search phase:
  - search_tools now performs async filtering using filterTool with phase 'search' for BM25 candidates.
  - When filtering is enabled, BM25 inspection expands up to the full index so high-ranked denied items don't hide allowed lower-ranked matches.
  - Returns up to topK allowed results.
  - Empty-result messaging updated (removes “Try different keywords” text).
- Load phase:
  - load_tool consults filterTool with phase 'load' and treats disallowed tools as not found (failed load).
  - “Did you mean” / single-tool suggestion behavior is filtered using phase 'load'.
- Active phase:
  - Injected / already-loaded tools are filtered per-request using filterTool with phase 'active' and disallowed tools are hidden for that request without clearing thread state.
- Failure semantics:
  - If filterTool throws or rejects, the processor "fails closed" and treats the tool as disallowed for that request.
- Scope:
  - Tools provided directly to the agent / via processInputStep remain available unless filtered externally; meta-tools search_tools and load_tool remain available.

### Docs, Tests & Exports
- Documentation: new "Request-aware filtering" section and examples for filterTool, its args, phases, execution semantics, and failure behavior (docs/src/content/en/reference/processors/tool-search-processor.mdx).
- Exports: re-exports ToolSearchFilterArgs and ToolSearchFilterPhase from processors barrel (packages/core/src/processors/processors/index.ts).
- Tests: Vitest coverage added for search/load/active phases, async predicates, failure/closed semantics, suggestion behavior, BM25 inspection behavior, and backward compatibility (packages/core/src/processors/processors/tool-search.test.ts).
- Changeset: added .changeset/six-badgers-drop.md documenting the minor change.

## Backward Compatibility

If no filterTool is provided, behavior is unchanged. The change is additive and reduces the need for apps to wrap/reimplement ToolSearchProcessor to enforce per-request tool policies.

## Notes / Non-goals

Not a full policy/categorization system. This PR does not change toModelOutput, requireApproval, externally provided tools handling, or always-loaded tool support.


LekoArts (commented):
Docs changes LGTM

.changeset/six-badgers-drop.md
```diff
@@ -0,0 +1,15 @@
+---
+'@mastra/core': minor
+---
+
+Added request-aware filtering for ToolSearchProcessor search, load, and active tools.
+
+```ts
+new ToolSearchProcessor({
+  tools,
+  filterTool: ({ toolName, requestContext }) => {
```
abhiaiyer91: this should just be called `filter`

abhiaiyer91 (changes requested):
rename to `filter`

abhiaiyer91 (approved):

PR #16108, opened 2026-05-02
fix: collect all tool invocations from streamed tool-calls step

### LINKED ISSUE #15576: [BUG] @mastra/client-js only picks one tool-invocation from a tool-calls step before continuing
### Describe the Bug

I think there is a bug in the client-side continuation logic in @mastra/client-js for streamed client tools.

  ### Core issue

  In the processStreamResponse(...) continuation path, when a step finishes with finishReason === 'tool-calls', the client appears to collect only one tool invocation from the assistant
  message:

```typescript
  const toolCall = [...(message?.parts ?? [])]
    .reverse()
    .find(part => part.type === 'tool-invocation')?.toolInvocation;
  if (toolCall) {
    toolCalls.push(toolCall);
  }
```

  Because this uses reverse().find(...), it selects only the last tool-invocation part from the message.

  So even though the code later loops over toolCalls, it is not actually collecting all tool invocations from the finished step.

  ### Why this matters

  If a streamed assistant step contains multiple client tool invocations, the client should:

  1. collect all pending tool-invocation parts from that step
  2. execute each pending client tool once
  3. patch all of their results into the assistant message
  4. recurse once with the fully updated message set

  The current code does not appear to do that.

  ### Secondary problem

  Executed tool calls are not removed from the toolCalls array before later continuations. That means the current structure can also reprocess previously seen tool calls.

  So the current behavior seems to be:

  1. collect one tool invocation from the finished step
  2. append it to an accumulated toolCalls array
  3. loop over that accumulated array
  4. recurse once per tool call

  That is different from “handle all tool calls from the current step.”

  ### Why I’m filing this

  We tested this in a real frontend using native MastraClient.getAgent(...).stream(...) and a single client meta-tool. That still failed whenever the run required repeated client-tool
  continuation.

  Server-only prompts succeeded.
  Client-tool runs that required repeated client continuations did not.

  This lines up with the code above: using a single meta-tool name does not help if the assistant can emit that same client tool multiple times in the same step, because only one
  invocation is harvested from the message.

  ### Suggested fix direction

  Instead of:

```typescript
  const toolCall = [...(message?.parts ?? [])]
    .reverse()
    .find(part => part.type === 'tool-invocation')?.toolInvocation;
```

  the client should gather all pending tool invocations from the message, something conceptually like:

```typescript
  const pendingToolCalls = (message?.parts ?? [])
    .filter(part => part.type === 'tool-invocation')
    .map(part => part.toolInvocation)
    .filter(toolInvocation => toolInvocation?.state === 'call');
```

  Then:

  1. execute each matching client tool once
  2. patch all results into a cloned lastMessage
  3. recurse once with the updated messages

  ### Question

  Is the current behavior intentional, or is this a bug in the continuation logic for multiple client tool invocations in a streamed step?

### Steps To Reproduce

1. Create an agent that can emit multiple client-side tool invocations in a streamed step.
  2. Call it through native @mastra/client-js using MastraClient.getAgent(...).stream(...).
  3. Pass at least one clientTools entry with an execute(...) function.
  4. Use a prompt that causes the model to issue more than one client tool invocation before it can finish the response.
  5. Observe the client continuation path when the stream finishes with finishReason === 'tool-calls'.

  A minimal repro pattern is:

  - the assistant emits multiple tool-invocation parts in one streamed assistant message
  - those tool invocations correspond to client tools
  - the client should continue after tool execution

  Example high-level shape:

```typescript
  const agent = mastraClient.getAgent('myAgent');

  await agent.stream(
    [{ role: 'user', content: [{ type: 'text', text: 'run multiple client tool operations' }] }],
    {
      clientTools: {
        my_client_tool: {
          type: 'function',
          parameters: { /* valid schema */ },
          async execute(args) {
            return { ok: true, args };
          },
        },
      },
      memory: {
        resource: 'resource-1',
        thread: 'thread-1',
      },
    },
  );
```

  The important part is that the model emits multiple client tool invocations during the streamed step.

can be provided, if necessary

### Expected Behavior

When a streamed step ends with finishReason === 'tool-calls', the client should:

  1. collect all pending tool-invocation parts from the assistant message for that step
  2. execute each matching client tool exactly once
  3. patch all returned results into the assistant message
  4. make a single continuation call with the fully updated message set
  5. continue until there are no pending client tool calls left

  The run should not stop early with unfinished or cancelled client tool calls just because multiple client tool invocations were emitted in the same step.

### Environment Information

```block
System:
    OS: macOS 26.4.1
    CPU: (12) arm64 Apple M2 Max
    Shell: /bin/zsh

  Binaries:
    Node: 24.14.1
    Yarn: 4.13.0
    npm: 11.11.0

  Browsers:
    Chrome: 147.0.7727.56

  Mastra Packages:
    mastra: 1.3.15
    @mastra/auth: 1.0.2
    @mastra/client-js: 1.10.0
    @mastra/core: 1.16.0
    @mastra/deployer: 1.16.0
    @mastra/evals: 1.1.2
    @mastra/express: 1.3.0
    @mastra/loggers: 1.0.3
    @mastra/mcp: 1.3.1
    @mastra/memory: 1.10.0
    @mastra/observability: 1.5.1
    @mastra/pg: 1.8.3
    @mastra/rag: 2.1.2

  Related AI SDK Packages:
    ai: 6.0.138
```

### Verification

aayushbaluni:
## Problem
`processStreamResponse` in `@mastra/client-js` uses `reverse().find()` which only picks the first tool-invocation from a tool-calls step, ignoring parallel tool calls.

## Change
Replace `find()` with `filter()` to collect all tool-invocation parts so parallel tool calls are all executed.

Fixes #15576.

## ELI5 (Explain Like I'm 5)

When the assistant asked to run more than one tool at the same time, the old code ran only one and lost the others. This change runs all requested tools in parallel, collects every result, and continues the stream once with all results included.

## Summary

Fixes a bug in @mastra/client-js where a streamed assistant step with multiple client tool invocations (finishReason === 'tool-calls') only executed/propagated a single invocation. Previously the code picked one invocation and recursed per-tool, causing lost results, duplicate/reordered processing, and writable-stream races in the legacy path.

Key changes:
- processStreamResponse (vNext) and processStreamResponseLegacy now gather all pending tool-invocation parts (from message parts and message.toolInvocations), deduplicate by toolCallId, and filter for client tools that implement execute.
- Executable client tools are run concurrently (Promise.all). The last assistant message is cloned once and patched so every matching invocation is marked state: 'result' and contains its result.
- Both flows perform exactly one recursive continuation call with the fully-updated messages containing all tool results. The legacy streaming path writes all tool-result chunks under a single writable.getWriter() lock and awaits a single recursive request to avoid concurrent writer races.
- Adds a Vitest streaming test covering a streamed assistant step that emits two tool-call chunks (weatherTool + newsTool), asserting both tools execute and the single recursive /stream call includes both results (fetch called exactly twice).

Impact:
- Ensures all parallel client tool invocations emitted in one assistant step are executed and their results are included together in the single continuation.
- Eliminates per-iteration recursion and legacy writable-stream race conditions.
- Fixes issue #15576.

Files changed (high level):
- client-sdks/client-js/src/resources/agent.ts — batching/deduping of tool invocation collection, concurrent execution, single-message patching, and single recursion.
- client-sdks/client-js/src/resources/agent.vnext.test.ts — new streaming tests for multi-tool invocations.
- .changeset/fiery-groups-accept.md — changeset bump for @mastra/client-js.

[Review Change Stack](https://app.coderabbit.ai/change-stack/mastra-ai/mastra/pull/16108)

### DIFF
--- a/.changeset/fiery-groups-accept.md
+++ b/.changeset/fiery-groups-accept.md
@@ -1,0 +1,5 @@
+---
+'@mastra/client-js': patch
+---
+
+Fixed client-side stream continuations after tool calls so recursive requests include executed tool results.
--- a/client-sdks/client-js/src/resources/agent.ts
+++ b/client-sdks/client-js/src/resources/agent.ts
@@ -1975,21 +1975,13 @@ export class Agent extends BaseResource {
         },
         onFinish: async ({ finishReason, message }) => {
           if (finishReason === 'tool-calls') {
-            const toolCall = [...(message?.parts ?? [])]
-              .reverse()
-              .find(part => part.type === 'tool-invocation')?.toolInvocation;
-            if (toolCall) {
-              const toolInvocationWithMetadata = message?.toolInvocations?.find(
-                invocation => invocation.toolCallId === toolCall.toolCallId,
-              );
-              toolCalls.push({
-                ...toolInvocationWithMetadata,
-                ...toolCall,
-                ...(!getClientToolObservabilityContext(toolCall) && toolInvocationWithMetadata
-                  ? { observability: getClientToolObservabilityContext(toolInvocationWithMetadata) }
-                  : {}),
-              });
-            }
+            const pendingToolCalls = (message?.parts ?? [])
+              .filter((part): part is ToolInvocationUIPart => part.type === 'tool-invocation')
+              .map(part => part.toolInvocation)
+              .filter((inv): inv is ToolInvocation => inv != null && inv.state === 'call');
+
+            toolCalls.length = 0;
+            toolCalls.push(...pendingToolCalls);
 
             let shouldExecuteClientTool = false;
             // Handle tool calls if needed
@@ -3000,21 +2992,13 @@ export class Agent extends BaseResource {
         },
         onFinish: async ({ finishReason, message }) => {
           if (finishReason === 'tool-calls') {
-            const toolCall = [...(message?.parts ?? [])]
-              .reverse()
-              .find(part => part.type === 'tool-invocation')?.toolInvocation;
-            if (toolCall) {
-              const toolInvocationWithMetadata = message?.toolInvocations?.find(
-                invocation => invocation.toolCallId === toolCall.toolCallId,
-              );
-              toolCalls.push({
-                ...toolInvocationWithMetadata,
-                ...toolCall,
-                ...(!getClientToolObservabilityContext(toolCall) && toolInvocationWithMetadata
-                  ? { observability: getClientToolObservabilityContext(toolInvocationWithMetadata) }
-                  : {}),
-              });
-            }
+            const pendingToolCalls = (message?.parts ?? [])
+              .filter((part): part is ToolInvocationUIPart => part.type === 'tool-invocation')
+              .map(part => part.toolInvocation)
+              .filter((inv): inv is ToolInvocation => inv != null && inv.state === 'call');
+
+            toolCalls.length = 0;
+            toolCalls.push(...pendingToolCalls);
 
             // Handle tool calls if needed
             for (const toolCall of toolCalls) {
--- a/client-sdks/client-js/src/resources/agent.ts
+++ b/client-sdks/client-js/src/resources/agent.ts
@@ -176,6 +176,9 @@ async function executeClientToolWithObservability({
   };
 }
 
+type ClientToolCallResult = { toolCall: ToolInvocation; result: any };
+
+
 async function executeToolCallAndRespond<OUTPUT>({
   response,
   params,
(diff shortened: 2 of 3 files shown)


mbenhamd:
Thanks for tackling this. The underlying bug is real, and collecting all pending `tool-invocation` parts is the right first step.

I don’t think this is complete yet, though, because the continuation logic still runs inside the per-tool loop.

In `processStreamResponse`, each iteration does roughly this:

```ts
for (const toolCall of toolCalls) {
  const result = await clientTool.execute(...);

  const lastMessage =
    lastMessageRaw != null ? JSON.parse(JSON.stringify(lastMessageRaw)) : undefined;

  // only this toolCallId is marked as result
  toolInvocationPart.toolInvocation = {
    ...toolInvocationPart.toolInvocation,
    state: "result",
    result,
  };

  await this.processStreamResponse({
    ...processedParams,
    messages: updatedMessages,
  }, controller, route);
}
```

So with two parallel client tool calls from the same streamed assistant step, the first recursive request contains only result A, and the second recursive request contains only result B. Because `lastMessage` is cloned fresh from `messages[messages.length - 1]` each time, prior loop mutations are not accumulated into the next iteration.

The legacy implementation has the same shape, but with an extra risk: it starts recursive calls inside the loop without awaiting them:

```ts
this.processStreamResponseLegacy(
  { ...processedParams, messages: updatedMessages },
  writable,
).catch(...)
```

With multiple tool calls, that can create multiple recursive streams writing to the same `writable`.

I think the fix should batch this step:

1. collect all pending executable client tool calls;
2. execute them;
3. clone the assistant message once;
4. mark every matching invocation as `state: "result"` with its result;
5. for legacy, write all result chunks;
6. recurse exactly once with the complete updated message.

That would preserve the intended “parallel tool calls in one assistant step -> one continuation containing all tool results” behavior.

Could we also add streaming coverage for both implementations? The key case is one streamed assistant step with two pending client tool invocations, asserting both tools execute and the recursive request includes both results in the same assistant message.

@coderabbitai could you also review this batching/recursive continuation case?

aayushbaluni:
@mbenhamd Thanks for the detailed review — you're absolutely right about the batching issue.

I've pushed a fix that implements exactly the approach you outlined:

**`processStreamResponse` (vNext path):**
1. After `onFinish` collects all pending `state === 'call'` invocations, the code now filters for executable client tools.
2. Executes all of them concurrently via `Promise.all`.
3. Clones the assistant message **once**.
4. Marks every matching invocation as `state: "result"` with its result in the single clone.
5. Recurses exactly **once** with the complete updated message.

**`processStreamResponseLegacy`:**
Same batching logic, plus writes all `a:` result chunks in a single batch before the one recursive call — eliminating the risk of multiple concurrent recursive streams writing to the same `writable`.

**Test coverage:**
Added a streaming test for the key case: one assistant step emitting two `tool-call` chunks (`weatherTool` + `newsTool`), asserting both tools execute and the single recursive request contains both results in the same assistant message (`fetch` is called exactly twice, not three times).

TheIsrael1:
Thanks for the PR! closing this! issue will be fixed in https://github.com/mastra-ai/mastra/pull/16981

PR #16111, opened 2026-05-02
feat(observability): expose dropped observability events

### LINKED ISSUE #16040: [FEATURE] Expose structured telemetry for dropped observability events
## Problem Statement

`DefaultExporter` can currently lose observability signals in ways that are hard for applications or external observability bridges to detect.

Two cases seem important:

1. If storage does not implement a signal type, `DefaultExporter` logs a warning once, marks that signal unsupported, and then skips future flushes for that signal.
2. If retryable storage failures continue past `maxRetries`, `EventBuffer` drops the events, but there is no structured drop telemetry/counter/callback that another exporter or bridge can observe.

This is especially relevant when traces work with a storage backend but native logs or metrics are unsupported. In that case a partial observability setup can look healthy if logs/metrics are unavailable except for local logger warnings or internal state.

Checked against `mastra-ai/mastra` `main` at commit `8f9e3822e3cb7a6cf8e070e6a01205cbc4dca701`.

Current behavior evidence:

- `DefaultExporterConfig` exposes batch/retry/strategy options, but no drop callback or telemetry hook: https://github.com/mastra-ai/mastra/blob/8f9e3822e3cb7a6cf8e070e6a01205cbc4dca701/observability/mastra/src/exporters/default.ts#L26-L31
- `DefaultExporter` tracks unsupported signals internally: https://github.com/mastra-ai/mastra/blob/8f9e3822e3cb7a6cf8e070e6a01205cbc4dca701/observability/mastra/src/exporters/default.ts#L83
- Unsupported signals return early on later flushes: https://github.com/mastra-ai/mastra/blob/8f9e3822e3cb7a6cf8e070e6a01205cbc4dca701/observability/mastra/src/exporters/default.ts#L224
- `_NOT_IMPLEMENTED` disables the signal after logging once: https://github.com/mastra-ai/mastra/blob/8f9e3822e3cb7a6cf8e070e6a01205cbc4dca701/observability/mastra/src/exporters/default.ts#L231-L234
- Logs/metrics flush via `batchCreateLogs` and `batchCreateMetrics`: https://github.com/mastra-ai/mastra/blob/8f9e3822e3cb7a6cf8e070e6a01205cbc4dca701/observability/mastra/src/exporters/default.ts#L371-L374
- `EventBuffer` re-adds events only while `retryCount <= maxRetries`; after that they are omitted from the retryable list: https://github.com/mastra-ai/mastra/blob/8f9e3822e3cb7a6cf8e070e6a01205cbc4dca701/observability/mastra/src/exporters/event-buffer.ts#L129-L145
- Base observability storage methods throw `_NOT_IMPLEMENTED` for logs/metrics by default: https://github.com/mastra-ai/mastra/blob/8f9e3822e3cb7a6cf8e070e6a01205cbc4dca701/packages/core/src/storage/domains/observability/base.ts#L269-L299
- `ObservabilityEvents` exposes tracing/log/metric/score/feedback hooks, but I do not see a drop/exporter-health event: https://github.com/mastra-ai/mastra/blob/8f9e3822e3cb7a6cf8e070e6a01205cbc4dca701/packages/core/src/observability/types/core.ts#L520-L543

## Proposed Solution

Expose structured exporter-health/drop telemetry when observability data is not persisted.

One possible API would be an `ObservabilityEvents` hook for exporter/storage drops, so external observability sinks can forward the event.

Alternatives could include a callback on `DefaultExporter`, counters exposed by the exporter, or a structured warning/event that bridges can forward.

Minimum useful metadata:

- Signal type: tracing, log, metric, score, feedback.
- Reason: unsupported storage, retry exhausted, validation failure, etc.
- Count dropped or disabled.
- Last error id/message/cause where safe.

## Component

Other

## Alternatives Considered

The current once-only warning is helpful for local logs, but it is not enough for production alerting. If an app forwards observability health to an external sink, it needs a structured signal rather than an internal set and a one-time console warning.

Applications can wrap storage/exporters locally, but that is hard to do consistently because retry exhaustion happens inside `EventBuffer` and unsupported-signal state is internal to `DefaultExporter`.

## Example Use Case

As a production Mastra user, I want to alert if native Mastra logs or metrics are unsupported or dropped, even if traces continue to work. This prevents a partial observability setup from looking healthy when some signal types are silently unavailable.

For example, an app using a storage backend that supports spans/traces but not native metrics/logs should be able to emit an exporter-health event saying metrics/logs are unsupported, with counts and storage/error metadata.

## Additional Context

Focused tests that would prove the behavior:

Tests should cover unsupported-signal disabling in `DefaultExporter` and retry exhaustion in `EventBuffer`.

## Verification

- [x] I have provided sufficient context for the team to understand the request.

mbenhamd:
Adds structured drop events for observability data that `DefaultExporter` cannot persist. Exporters and bridges can implement `onDroppedEvent` to observe unsupported storage drops and retries being exceeded with signal, reason, count, storage, exporter, timestamp, and sanitized error details.

Before, unsupported logs or metrics could be dropped after a local warning, and retry exhaustion inside the event buffer was not visible to other observability sinks.

After:

```ts
class DropAlertExporter extends BaseExporter {
  name = 'drop-alerts'

  async onDroppedEvent(event: ObservabilityDropEvent) {
    await fetch('https://monitoring.example.com/observability-drops', {
      method: 'POST',
      headers: { 'content-type': 'application/json' },
      body: JSON.stringify({
        count: event.count,
        signal: event.signal,
        reason: event.reason,
        exporterName: event.exporterName,
      }),
    })
  }
}
```

Includes focused tests for bus fan-out, `DefaultExporter` unsupported-storage and retry-exhausted paths, and `EventBuffer` dropped-event return values.

Fixes #16040

## ELI5

When observability systems (which collect logs, metrics, and traces from applications) can't store or send data, they usually just silently fail. This PR adds a way for exporters to be notified about these failures so that external monitoring systems can alert when observability is only partially working—for example, when traces are being captured but logs aren't.

---

## Overview

This PR introduces a structured event notification system for dropped observability data. When the observability pipeline cannot persist events (due to unsupported storage or exhausted retries), it now emits standardized `ObservabilityDropEvent` objects to exporters and bridges implementing the new `onDroppedEvent` handler, enabling external monitoring and alerting on partial observability setups.

---

### New Type Definitions

Four new types are introduced in `packages/core/src/observability/types/core.ts`:
- `ObservabilityDropSignal`: Identifies the signal type (`'tracing' | 'log' | 'metric' | 'score' | 'feedback'`)
- `ObservabilityDropReason`: Indicates why the event was dropped (`'unsupported-storage' | 'retry-exhausted'`)
- `ObservabilityDropError`: Sanitized error details (`id`, `domain`, `message`)
- `ObservabilityDropEvent`: Complete drop notification with signal, reason, count, timestamp, exporter/storage names, and optional error

### API Extensions

- `ObservabilityExporter` interface now includes optional `onDroppedEvent?(event: ObservabilityDropEvent): void | Promise<void>` handler
- `InitExporterOptions` now includes optional `emitDropEvent` callback for exporters to emit drop events

### Bus & Routing

- `ObservabilityBus` adds `emitDropEvent(event: ObservabilityDropEvent): void` method to broadcast drop events to all registered exporters and bridges
- New `routeDropToHandler` function in `route-event.ts` safely routes drop events to exporter handlers with error handling and async support

### DefaultExporter Enhancements

`DefaultExporter` now:
- Tracks unsupported signals per storage instance using typed `ObservabilityDropSignal`
- Emits drop events with `reason: 'unsupported-storage'` when a storage adapter doesn't support a signal
- Emits drop events with `reason: 'retry-exhausted'` when flush retries exceed the configured maximum
- Includes sanitized `MastraError` details in drop events for debugging
- Threads drop emission through create-flush and span update/end paths
- Uses singular signal names (`log`, `metric`, `score`) instead of plural in drop events

### EventBuffer Updates

`EventBuffer.reAddCreates()` and `EventBuffer.reAddUpdates()` now return `BufferedEvent[]` of events that exceeded `maxRetries` instead of silently dropping them. This allows callers (DefaultExporter) to explicitly emit drop events for exhausted retries.

### Documentation

- **Configuration examples** section added to DefaultExporter documentation with a new "Dropped observability events" subsection explaining the two drop reasons and providing a `DropAlertExporter` example that forwards drop events to an external monitoring endpoint
- Reference documentation updated to describe the new drop event types, callback signature, and metadata included in drop notifications

### Testing

- `ObservabilityBus` tests verify drop event routing to exporters and bridges, async handler handling, and safe handling of missing `onDroppedEvent` implementations
- `DefaultExporter` tests extend coverage to verify drop events for unsupported-storage and retry-exhausted scenarios, including error details and correct signal/reason values
- `EventBuffer` tests verify that `reAddCreates` and `reAddUpdates` properly return dropped events when retry limits are exceeded

---

## Documentation Updates

Two changesets document the additions:
- `@mastra/observability`: DefaultExporter now emits drop-event telemetry to custom exporters and integrations
- `@mastra/core`: New structured drop-event types and `onDroppedEvent` hook for bridge/exporter integrations


LekoArts (commented):
Docs changes LGTM

abhiaiyer91 (approved):

docs/src/content/en/docs/observability/tracing/exporters/default.mdx
```diff
@@ -183,7 +183,37 @@ The DefaultExporter includes robust error handling for production use:
...
+
+class DropCounterExporter extends BaseExporter {
+  name = 'drop-counter'
+
+  onDroppedEvent(event: ObservabilityDropEvent) {
+    metricsCounter.add(event.count, {
+      signal: event.signal,
+      reason: event.reason,
+      exporterName: event.exporterName,
+    })
```
epinzur, line 214: can we use some other example here... Our metrics doesn't support `Counter.add()` so I'd rather not show this here and potentially confuse users.

docs/src/content/en/docs/observability/tracing/exporters/default.mdx
```diff
@@ -183,7 +183,37 @@ The DefaultExporter includes robust error handling for production use:
 - **Persistent Failures**: Drop batch after 4 failed attempts
 - **Buffer Overflow**: Prevent memory issues during storage outages
 
-### Configuration Examples
+## Dropped event telemetry
+
+`DefaultExporter` emits structured drop telemetry when it cannot persist observability data. Register an exporter or bridge with `onDroppedEvent` to forward these drops to alerting or metrics.
```
epinzur: nit: can we avoid using `telemetry` in the description? 

maybe: "`DefaultExporter` emits structured drop event when it cannot persist observability data."

docs/src/content/en/reference/observability/tracing/interfaces.mdx
```diff
@@ -150,6 +150,9 @@ interface ObservabilityExporter {
   /** Handle feedback events */
   onFeedbackEvent?(event: FeedbackEvent): void | Promise<void>
 
+  /** Handle exporter pipeline drop telemetry */
```
epinzur: `/** Handle exporter pipeline droppedEvent */`

docs/src/content/en/docs/observability/tracing/exporters/default.mdx
epinzur: in Mastra: 

* `telemetry` refers to standard infrastructure observability/telemtry
* `observability` refers to AI related observability/telemetry

We try very hard to not use these words interchangeably in docs, code, etc...

epinzur (changes requested):
looks great, but pls make the few wording & example changes

thanks for adding this safegaurd!

docs/src/content/en/docs/observability/tracing/exporters/default.mdx
mbenhamd: AI says: Updated the docs example to avoid the unsupported metrics counter API. It now uses a generic monitoring endpoint example instead of Counter.add(), and the PR body has been updated to match. Commit 48571fcca2 includes the docs fix. Mohamed

docs/src/content/en/docs/observability/tracing/exporters/default.mdx
mbenhamd: AI says: Replaced telemetry wording in the docs, changesets, code comments, test names, PR title, and PR body so the feature is described as dropped observability events. Commit 48571fcca2 includes the code/docs wording fixes. Mohamed

docs/src/content/en/reference/observability/tracing/interfaces.mdx
mbenhamd: AI says: Updated the generated reference docs and the source TypeScript interface comment to use the requested text: /** Handle exporter pipeline droppedEvent */. Commits 48571fcca2 and 70894ac69b include the fix. Mohamed

epinzur (approved):

PR #16113, opened 2026-05-02
fix(core,nestjs): register scheduled workflows and preserve route inputs

### LINKED ISSUE #16112: [BUG] Scheduled workflows and NestJS routes mishandle prefixes and query IDs
Scheduled workflows and the NestJS adapter have a few related request-path correctness issues that should be tracked separately from observability work.

Problems:
- Importing createWorkflow from @mastra/core/workflows and declaring schedule can fail because the public barrel does not load the evented workflow registration side effect.
- In @mastra/nestjs, a configured prefix such as /api is not enforced when matching Mastra routes, so unprefixed requests like /agents can still be handled by Mastra.
- The NestJS adapter coerces numeric-looking query values to numbers before route schema validation, which can reject or change string identifiers such as resourceId=123 or runId=123.

Expected behavior:
- The public @mastra/core/workflows entry point should support scheduled workflow promotion without a separate evented import.
- NestJS Mastra routes should only match inside the configured prefix.
- Query identifiers should remain strings and let the route schema decide coercion.

I have a focused fix ready with tests for the workflow barrel registration, route-prefix matching, and query identifier preservation.

mbenhamd:
This fixes scheduled workflows created through the public `@mastra/core/workflows` barrel by loading the evented workflow registration when the barrel is imported. It also tightens `@mastra/nestjs` routing so configured prefixes are enforced, and keeps numeric-looking query identifiers as strings so route schemas can validate or coerce them consistently.

Before:
```ts
MastraModule.register({ mastra, prefix: "/api" })
// GET /agents could still match Mastra routes
// GET /api/agents?resourceId=123 passed resourceId as 123
```

After:
```ts
MastraModule.register({ mastra, prefix: "/api" })
// GET /api/agents matches Mastra routes
// GET /agents is left outside the Mastra route matcher
// GET /api/agents?resourceId=123 passes resourceId as "123"
```

Fixes #16112

## ELI5

This PR fixes three things: (1) scheduled workflows you create using the public `@mastra/core/workflows` entry point now automatically work without needing extra imports, (2) when you set up Mastra with a route prefix like `/api`, it now strictly enforces that prefix—requests to `/agents` are ignored and only `/api/agents` is handled, and (3) query parameters that look like numbers (like `resourceId=123`) stay as text strings so your route validation can control what type they should be.

---

### 1. Scheduled Workflows Auto-Registration
- **`packages/core/src/workflows/index.ts`**: Added a side-effect import of `./evented` after base workflow exports to ensure `EventedWorkflow` can extend `Workflow` without triggering ESM circular dependency issues. This allows scheduled workflows declared via `createWorkflow` from the public `@mastra/core/workflows` barrel to be registered automatically.
- **`packages/core/src/workflows/schedule-promotion.test.ts`**: Removed the local side-effect import of `./evented` since it's now handled in the barrel index.

### 2. NestJS Route Prefix Enforcement
The `@mastra/nestjs` adapter now strictly constrains route matching to a configured prefix:

- **`server-adapters/nestjs/src/utils/route-path.ts`**: Updated `getMastraRoutePath` return type from `string` to `string | null`. The function now returns `null` when the incoming path doesn't match the configured prefix (not just a partial match like `/apiish/...`), and normalizes prefix variants (e.g., `'api'` and `'/api/'`) before matching. Matching paths have the prefix stripped and are normalized to start with `/`, defaulting to `'/'` when empty.

- **`server-adapters/nestjs/src/guards/mastra-route.guard.ts`**: Added an early return (`return true`) when `getMastraRoutePath` returns a falsy value, allowing the request to pass through without attempting route matching, authentication, or rate limiting.

- **`server-adapters/nestjs/src/controllers/mastra.controller.ts`**: Now throws `NotFoundException` when `getMastraRoutePath` returns no route path instead of proceeding to further checks.

- **`server-adapters/nestjs/src/interceptors/tracing.interceptor.ts`**: Only calls `routeHandler.matchRoute` when `routePath` is truthy; otherwise sets `matchResult` to `null`.

- **Test coverage added in `server-adapters/nestjs/src/__tests__/route-path.test.ts`**: New Vitest suite verifying prefix behavior, including normalization of no-leading-slash (`'api'`), trailing-slash (`'/api/'`), undefined/empty prefix cases, exact-prefix matching, stripping, and rejection of non-matching paths (including partial matches like `/apiish/agents`).

### 3. Query Parameter String Preservation
- **`server-adapters/nestjs/src/controllers/mastra.controller.ts`**: Removed numeric-string → `number` coercion logic in `coerceQueryValue`. Numeric-looking strings (e.g., `resourceId=123`) now remain as strings, allowing route schemas to explicitly handle type validation or coercion.

- **`server-adapters/nestjs/src/__tests__/request-context.test.ts`**: Added tests verifying that numeric-looking query identifiers (e.g., `resourceId=123`, `runId=456`) are preserved as strings in `queryParams`.

### 4. Test Updates
Updated multiple NestJS test files to use the `/api/` prefix path pattern:
- `auth-rate-limit.test.ts`: Routes updated to `/api/agents`
- `mastra-module.test.ts`: Routes updated to `/api/agents/...`
- `request-context.test.ts`: Routes updated to `/api/...` with string-preservation assertions
- `server-app-access.test.ts`: Routes updated to `/api/agents`

### 5. Changesets
- `.changeset/silly-mugs-do.md`: Patch release for `@mastra/core` documenting the scheduled workflow registration fix.
- `.changeset/famous-rules-begin.md`: Patch release for `@mastra/nestjs` documenting route prefix enforcement and query parameter string preservation.
- `.changeset/ten-experts-check.md`: Patch release for `@mastra/core` documenting additional suspend/resume fixes in evented workflows (parallel, `.branch()`, `dountil`/`dowhile`, and nested workflow scenarios).

---

## Impact

- **`@mastra/core`**: Scheduled workflows are now automatically registered when importing from the public barrel, removing a footgun for users.
- **`@mastra/nestjs`**: Route matching is now properly scoped to the configured prefix; unprefixed requests are rejected early. Query parameters remain as strings, deferring type coercion decisions to route schemas.

Fixes issue #16112.

[Review Change Stack](https://app.coderabbit.ai/change-stack/mastra-ai/mastra/pull/16113)


mbenhamd:
@coderabbitai I added the requested prefix normalization coverage in e17f183faf: no-leading-slash prefix, trailing-slash prefix, undefined prefix, exact-prefix matching, and a non-matching prefixed route case. The @mastra/core changeset finding appears stale: .changeset/silly-mugs-do.md already targets '@mastra/core': patch, and changeset-bot detects @mastra/core for this PR, so I did not add a duplicate changeset.

mbenhamd:
Closing this in favor of a smaller replacement PR: #16637.

The query-parameter portion is now covered by #16268, and this branch conflicts with newer evented workflow changes on main. The replacement keeps only the still-relevant scheduled workflow barrel registration and NestJS prefix enforcement fixes.

PR #16117, opened 2026-05-03
fix(memory): hoist image-data tool-result blocks into observer attachments

### LINKED ISSUE #15883: [BUG] Observational memory system doesn't handle tools which output image-data
### Describe the Bug

When a tool produces a `type: 'image-data'` (e.g. using `toModelOutput`), the observational memory system feeds that message into the observer as base64 text. This causes all sorts of issues, from exceeding token limits, to death spirals in the observer output.

### Steps To Reproduce

- Define a tool which produces `type: 'image-data'` messages via `toModelOutput`
- Add it to an OM agent
- Get the agent to call the tool, and then wait for OM to observe the tool call

Will create if requested, but this is straight-forward to reproduce

### Expected Behavior

Any `type: 'image-data'` blocks should be shown to the observer as `type: 'image-data'` blocks, instead of being shown as base64 text.

It's tempting to consider stripping `type: 'image-data'` blocks, but this is a poor solution as the agent may not verbally comment on what it sees, so data would be lost in this case.

### Environment Information

```block
Anthropic
```

### Verification

intojhanurag:
## Description

Tools that return AI SDK v5 `image-data` content blocks via `toModelOutput` were having their base64 payloads stringified into the observational-memory observer's text prompt, blowing past token limits and producing degenerate observer output. This change extracts media blocks out of tool-result outputs and passes them to the observer as real attachments — mirroring the existing handling for user-side `image` and `file` message parts.

- Added `mapToolResultBlockToAttachment` to translate AI SDK v5 content blocks (`image-data`, `image-url`, `media`, `file-data`, `file-url`) into `ObserverAttachmentPart` form, building `data:<mediaType>;base64,<data>` URIs for inline blocks.
- Added `extractToolResultAttachments` which walks `{ type: 'content', value: [...] }` tool outputs, hoists media blocks as attachments, and replaces each block in the residual with `{ type, placeholder: '[Image #N: ...]' }` so the JSON-stringified body stays small and shape-stable.
- Wired the extractor into the `tool-invocation` result branch of `formatObserverMessage` so attachments flow through to `buildObserverHistoryMessage` alongside the existing image/file part path.
- Non-content outputs (`text`, `json`, `error-text`, `error-json`, `execution-denied`) and unknown block types (`image-file-id`, `file-id`, `custom`) pass through unchanged.
- Added three regression tests in `observational-memory.test.ts`: image-data block becomes a placeholder in `formatMessagesForObserver`, plain object tool results still serialize as JSON, and `buildObserverHistoryMessage` emits a real `{ type: 'image' }` attachment with the correct data URI and `mimeType`.

## Related Issue(s)

Fixes #15883

## Type of Change

- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [ ] Test update

## Checklist

## ELI5

Tools were embedding full images/files as huge text blobs, which made messages too large and broke prompts. This PR pulls those images/files out of the text and sends them as attachments, leaving small numbered placeholders in the text to keep order and context without the bloat.

## Overview

Detects AI SDK v5 content blocks inside tool-result outputs (image-data, image-url, media, file-data, file-url), hoists recognized media/file blocks into ObserverAttachmentPart entries, and for inline binary blocks builds data:<mediaType>;base64,<data> URIs. The extractor replaces the original blocks in the residual content with compact numbered placeholder blocks (e.g., { type, placeholder: '[Image #1: image/png]' }) so the JSON-stringified body stays small and shape-stable. The extractor is integrated into the tool-invocation result branch of formatObserverMessage so attachments flow into buildObserverHistoryMessage alongside existing user-attached image/file parts. Non-content outputs (text, json, error-text, error-json, execution-denied) and unknown/unsupported block types (image-file-id, file-id, custom) pass through unchanged.

## Key changes

- Added mapToolResultBlockToAttachment: translates content blocks into ObserverAttachmentPart (creates data: URIs for inline blocks).
- Added extractToolResultAttachments: walks content outputs ({ type: 'content', value: [...] }), hoists media/file blocks into attachments, and replaces them with numbered placeholder blocks in the residual content.
- Wired the extractor into formatObserverMessage/observer-agent so hoisted attachments are appended to message attachments and the textual "Tool Result …" parts contain placeholders instead of raw base64.
- Ensures image/file numbering counter is shared between user-attached parts and tool-result attachments so placeholders are sequential and stable.

## Tests

- Added regression tests in observational-memory.test.ts:
  - formatMessagesForObserver produces placeholders for image-data and ensures base64 does not appear in formatted text.
  - Plain-object tool results still serialize to JSON correctly.
  - buildObserverHistoryMessage emits { type: 'image' } attachments with correct data: URI and mimeType, and verifies shared numbering between user-attached and tool-result images.
  - Coverage includes image/file URL blocks, generic media blocks, and image-data without mediaType.

## Docs / Changeset

- .changeset entry describing the fix to prevent token-limit failures by hoisting media/file blocks into attachments and inserting placeholders in text.

## Issue resolved

Fixes #15883 — observational memory now delivers tool-produced image/file data as attachments with placeholder references in text, preventing token-limit overflows while preserving positional context.

### DIFF
--- a/.changeset/huge-papers-rest.md
+++ b/.changeset/huge-papers-rest.md
@@ -1,0 +1,7 @@
+---
+'@mastra/memory': patch
+---
+
+Fixed an issue where tool results containing AI SDK v5 `image-data` content blocks (returned via `toModelOutput`) were stringified into the observational memory prompt as raw base64 text. The base64 data overflowed the observer's context, causing token-limit errors and degenerate output.
+
+Image and file blocks (`image-data`, `image-url`, `file-data`, `file-url`, and `media`) inside tool results are now hoisted into the observer's input as proper attachments, the same way image and file message parts already are. The text body shows a placeholder like `[Image #1: image/png]` so the observer keeps positional context without seeing the bytes.
--- a/packages/memory/src/processors/observational-memory/__tests__/observational-memory.test.ts
+++ b/packages/memory/src/processors/observational-memory/__tests__/observational-memory.test.ts
@@ -1404,6 +1404,71 @@ describe('Observer Agent Helpers', () => {
       expect(formatted).not.toContain('x'.repeat(200));
     });
 
+    // Regression test for https://github.com/mastra-ai/mastra/issues/15883
+    // Tools that return AI SDK v5 image-data blocks via toModelOutput must not
+    // dump the base64 payload into the observer's text body — it blows the
+    // observer's context and causes degenerate output.
+    it('should replace image-data tool-result blocks with attachment placeholders', () => {
+      const base64 = 'A'.repeat(2000);
+      const msg = createTestMessage('ignored', 'assistant');
+      msg.content = {
+        format: 2,
+        parts: [
+          {
+            type: 'tool-invocation',
+            toolInvocation: {
+              state: 'result',
+              toolCallId: 'tool-1',
+              toolName: 'screenshot',
+              args: { url: 'https://example.com' },
+              result: {},
+            },
+            providerMetadata: {
+              mastra: {
+                modelOutput: {
+                  type: 'content',
+                  value: [
+                    { type: 'text', text: 'Captured screenshot of the homepage.' },
+                    { type: 'image-data', data: base64, mediaType: 'image/png' },
+                  ],
+                },
+              },
+            },
+          },
+        ],
+      } as any;
+
+      const formatted = formatMessagesForObserver([msg]);
+      expect(formatted).toContain('Tool Result screenshot');
+      expect(formatted).toContain('Captured screenshot of the homepage.');
+      expect(formatted).toContain('[Image #1: image/png]');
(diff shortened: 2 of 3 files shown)


TylerBarnes:
From mastracode: I tightened the tool-result attachment hoisting code and added focused regression coverage. Specifically, I replaced the loose Record casts in the observer formatter with explicit record narrowing, switched the content-block mapper to a clearer switch, and renamed the extracted text payload from residual to resultWithoutAttachments so the flow is easier to reason about. I also added tests for image/file URL blocks, generic media blocks, and image-data without mediaType to document that media stays out of observer text while still being passed to the model as attachments. This keeps the PR scoped to the base64/token-overflow fix while improving confidence in the supported AI SDK v5 content block paths.

TylerBarnes (approved):

PR #16118, opened 2026-05-03
fix(playground): prevent IME composition Enter from submitting agent …

### LINKED ISSUE #16109: [BUG] The Agent Chat interface of Studio UI cannot input Chinese normally.
### Describe the Bug

The 1.7.0 version of mastra is normal, but after upgrading to 1.7.1, the agent chat input box in the studio UI cannot input Chinese normally.

### Steps To Reproduce

1. Install version 1.7.1 or higher.
2. Open any Agent and directly input Chinese in the input box.
3. No response.
4. Downgrade to 1.7.0, and input works normally.

---

### Expected Behavior

it works

### Environment Information

```block
- Browser: chrome 147
```

### Verification

guangyang1206:
…chat

When using a Chinese/Japanese/Korean IME to input text, pressing Enter to confirm character selection was triggering the chat form submission instead of completing the IME composition.

Root cause: ComposerPrimitive.Input uses Radix's asChild Slot pattern which merges the primitive's internal onKeyDown (which submits on Enter) with the textarea's handlers. IME composition events fire before the keydown, but without checking isComposing the Enter key always submits.

Fix: Add an isComposingRef (useRef) and wire onCompositionStart/End to flip the flag. In onKeyDown, stop propagation when Enter fires during an active IME composition so the primitive never sees it as a submit.

Fixes #16109

## Description

When using a Chinese/Japanese/Korean IME to input text in the Agent Chat
interface, pressing **Enter** to confirm character selection was triggering
form submission instead of completing the IME composition cycle.

This regression was introduced in v1.7.1 (background task feature, #15307)
which added `onChange={e => setThreadInput?.(e.target.value)}` to the
`<textarea asChild>` inside `ComposerPrimitive.Input`.

**Root cause:**

`ComposerPrimitive.Input` uses Radix's `asChild` / `Slot` pattern, which
merges the primitive's internal `onKeyDown` (Enter → submit) with the
textarea's handlers.  IME composition fires `compositionstart` /
`compositionend` events around character selection, but without a
`isComposing` guard the primitive's Enter handler fires unconditionally,
submitting the message mid-composition.

**Fix:**

Track composition state via `isComposingRef` (a `useRef` to avoid
re-renders) and wire `onCompositionStart` / `onCompositionEnd` on the
textarea to flip it.  In `onKeyDown`, call `e.stopPropagation()` when
Enter fires during active composition so `ComposerPrimitive.Input`'s
submit handler never sees it.

## Related Issue(s)

Fixes #16109

## Type of Change

### `packages/playground/src/lib/ai-ui/thread.tsx`

```tsx
// Added inside Composer component:
const isComposingRef = useRef(false);

// Added to <textarea>:
onCompositionStart={() => { isComposingRef.current = true; }}
onCompositionEnd={() => { isComposingRef.current = false; }}
onKeyDown={e => {
  if (e.key === 'Enter' && isComposingRef.current) {
    e.stopPropagation();
  }
}}
```

## Checklist

## ELI5

When typing with Chinese/Japanese/Korean input, pressing Enter both confirmed a character and accidentally sent the chat message. This change detects when IME composition is active and prevents Enter from submitting the form so you can finish typing characters normally.

## Changes

**File: `packages/playground/src/lib/ai-ui/thread.tsx`**
- Track IME composition with `isComposingRef` (`useRef(false)`).
- Attach `onCompositionStart` / `onCompositionEnd` to the textarea to toggle `isComposingRef`.
- Add `onKeyDown` on the textarea that, when Enter is pressed during composition, calls `e.preventDefault()` and `e.stopPropagation()` to stop the Radix primitive’s Enter→submit handler; also checks `e.nativeEvent.isComposing` as a defensive fallback.

**File: `packages/playground/e2e/tests/agents/$agentId/ime-composition.spec.ts`**
- Add Playwright e2e test simulating IME composition: asserts Enter during composition is prevented (no submit) and Enter after `compositionend` submits as expected.

**File: `client-sdks/client-js/src/route-types.generated.ts`**
- Extend metrics-related route types: `aggregation` now includes `'count_distinct'`.
- Add optional `distinctColumn?: ...` (restricted literal union) tied to `count_distinct`.
- Add optional query fields for metrics grouping: `limit?: number` and `orderDirection?: 'ASC' | 'DESC'`.

## Rationale / Root cause
Radix’s asChild/Slot pattern merged the primitive’s internal Enter→submit handler with the textarea’s handlers. IME composition events happen before keydown; because composition state wasn’t tracked, the primitive’s Enter handler ran while composition was active and submitted the form.

## Impact
- Type: Bug fix (non-breaking).
- Affected area: Agent Chat textarea (playground UI) and e2e tests; also unrelated type additions in client-js route types.
- No exported/public API removals; added optional route-type fields.
- Approx. lines changed: +23/-0 in thread.tsx, +92/-0 for the e2e spec, +64/-9 in route-types.generated.ts.

## Fixes
Closes issue #16109 — restores correct CJK IME composition behavior in the Agent Chat input.


roaminro (approved):

roaminro:
Thanks for the contribution @guangyang1206 !

PR #16122, opened 2026-05-03
fix(editor): forward userId from request context to tool providers

### LINKED ISSUE #16120: Editor tools are not passed userId
This issue was created from Discord post 1500504283118833755:

[Open in Discord](https://discord.com/channels/1309558646228779139/1500504283118833755)

According to the docs, the editor's tool providers, such as composio and arcade, should receive the userId from the requestContext but it is never passed.

in agent.ts:
`const providerTools = await provider.resolveTools(slugsToResolve, providerConfig.tools, { requestContext });`

Request context is passed as an option, but the userId arg is already separate by then.

In both composio.ts and arcade.ts the userId is taken directly off the options and not from the userId.

`const userId = options?.userId ?? 'default';`

So it is always set to "default".

intojhanurag:
## Description

Editor tool providers (Composio, Arcade) were receiving a hardcoded `'default'` userId because the agent hydration path passed only `requestContext` to `provider.resolveTools`, never extracting `userId` from it. Per-user OAuth scoping silently broke for every stored agent using `integrationTools`. This PR forwards `userId` from `requestContext.get('userId')` to providers, matching the documented behavior in `docs/src/content/en/docs/editor/tools.mdx`.

- `resolveStoredIntegrationTools` now accepts `RequestContext` (matching `resolveStoredMCPTools`) and pulls `userId` from it once before the provider loop.
- Passes `{ requestContext: requestContext.toJSON(), userId }` to `provider.resolveTools` so both the top-level `userId` field and the JSONified context reach Composio/Arcade.
- Integration tools are forced through the dynamic resolution path (`isDynamicTools` gate now includes `hasIntegrationTools`). Without this, purely-stored agents bake integration tools at creation time with no `requestContext`, so `userId` would still default to `'default'` for the most common case.
- All four call sites pass the `RequestContext` through, including the previously-missing pass at the `applyStoredOverrides` static branch.

### Usage

```typescript
import { RequestContext } from '@mastra/core/request-context';

const ctx = new RequestContext();
ctx.set('userId', 'user-123');

await agent.generate('list my repos', { requestContext: ctx });
```

## Related Issue(s)

Fixes #16120

## Type of Change

- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [ ] Test update

## Checklist

## ELI5

This PR fixes a bug where all users' tool calls (like Composio and Arcade integrations) were incorrectly routed to a shared 'default' user account instead of each user's own account. The fix passes the authenticated user's ID from the request context through to the tool providers, so each user's OAuth connections work correctly.

## Problem

Editor tool providers (Composio and Arcade) were receiving a hardcoded `'default'` userId when resolving tools during agent execution. This caused all tool calls to be scoped to a shared account rather than the authenticated user's account, breaking per-user OAuth scoping for integrations.

## Solution

The PR forwards `userId` from the request context through the entire tool resolution chain:

- **`resolveStoredIntegrationTools` refactor**: Now accepts a `RequestContext` parameter (matching the pattern of `resolveStoredMCPTools`), extracts the `userId` once before iterating through providers, and passes it along with the full request context to each provider's `resolveTools` call.

- **Dynamic resolution enforcement**: Integration tools are now always treated as "dynamic" (resolved at request-time rather than agent-creation time) whenever `integrationTools` are present, ensuring they have access to the request context with the authenticated user.

- **Provider-level identity resolution**: Both `ComposioToolProvider` and `ArcadeToolProvider` now check for `resourceId` in the request context first (`MASTRA_RESOURCE_ID_KEY`), fall back to `options.userId`, and finally default to `'default'` if neither is available.

- **Comprehensive call-site updates**: All four call sites that invoke `resolveStoredIntegrationTools` now pass the full `RequestContext`, including the previously-missing pass in the `applyStoredOverrides` static branch.

## Changes Made

**Core Logic** (`packages/editor/src/namespaces/agent.ts`):
- `resolveStoredIntegrationTools` now accepts an optional `RequestContext` parameter
- Integration tools are treated as dynamic whenever `hasIntegrationTools` is true
- Providers receive both the JSONified context and the extracted `userId` in options

**Provider Integration** (`packages/editor/src/providers/`):
- `composio.ts`: Updated to read `userId` from `requestContext[MASTRA_RESOURCE_ID_KEY]` with fallbacks
- `arcade.ts`: Updated with the same identity resolution strategy

**Testing** (`packages/editor/src/editor-integration-tools.test.ts`):
- Updated existing tests to expect `requestContext: {}` instead of `undefined`
- Added new test verifying that `agent.listTools({ requestContext })` properly forwards the context to tool providers for resource-scoped identity resolution
- Updated warning test to explicitly call `agent.listTools()`

**Documentation**:
- Updated Composio and ToolProvider documentation to reflect that tool scoping uses `resourceId` from request context instead of a top-level `userId`

**Release** (`.changeset/wicked-experts-care.md`):
- Added patch changeset for `@mastra/editor` documenting the fix

## Outcome

Editor tool integrations now correctly scope tool execution to the authenticated user from the request context, enabling proper per-user OAuth connections for multi-tenant deployments.

### DIFF
--- a/packages/editor/src/providers/arcade.ts
+++ b/packages/editor/src/providers/arcade.ts
@@ -309,8 +309,8 @@ export class ArcadeToolProvider implements ToolProvider {
     if (toolSlugs.length === 0) return {};
 
     const client = this.getClient();
-    const userId =
-      (options?.requestContext?.[MASTRA_RESOURCE_ID_KEY] as string | undefined) ?? options?.userId ?? 'default';
+    const resourceId = options?.requestContext?.[MASTRA_RESOURCE_ID_KEY];
+    const userId = typeof resourceId === 'string' ? resourceId : (options?.userId ?? 'default');
 
     // Fetch tool definitions for the requested slugs
     const toolDefs = await Promise.all(toolSlugs.map(slug => client.tools.get(slug).catch(() => null)));
--- a/packages/editor/src/providers/arcade.ts
+++ b/packages/editor/src/providers/arcade.ts
@@ -9,6 +9,7 @@ import type {
 } from '@mastra/core/tool-provider';
 import type { ToolAction } from '@mastra/core/tools';
 import type { StorageToolConfig } from '@mastra/core/storage';
+import { MASTRA_RESOURCE_ID_KEY } from '@mastra/core/request-context';
 
 import { Arcade } from '@arcadeai/arcadejs';
 import type {
@@ -18,7 +19,6 @@ import type {
 } from '@arcadeai/arcadejs/resources';
 import { toZodToolSet, executeOrAuthorizeZodTool } from '@arcadeai/arcadejs/lib/index';
 import type { ZodTool, ToolAuthorizationResponse } from '@arcadeai/arcadejs/lib/index';
-import { getProviderUserId } from './identity';
 
 export interface ArcadeToolProviderConfig {
   /** Arcade AI API key */
@@ -309,7 +309,8 @@ export class ArcadeToolProvider implements ToolProvider {
     if (toolSlugs.length === 0) return {};
 
     const client = this.getClient();
-    const userId = getProviderUserId(options);
+    const userId =
+      (options?.requestContext?.[MASTRA_RESOURCE_ID_KEY] as string | undefined) ?? options?.userId ?? 'default';
 
     // Fetch tool definitions for the requested slugs
     const toolDefs = await Promise.all(toolSlugs.map(slug => client.tools.get(slug).catch(() => null)));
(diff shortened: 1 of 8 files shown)


packages/editor/src/providers/arcade.ts
```diff
@@ -308,7 +309,8 @@ export class ArcadeToolProvider implements ToolProvider {
     if (toolSlugs.length === 0) return {};
 
     const client = this.getClient();
-    const userId = (options?.userId as string) ?? 'default';
+    const resourceId = options?.requestContext?.[MASTRA_RESOURCE_ID_KEY];
+    const userId = typeof resourceId === 'string' ? resourceId : (options?.userId ?? 'default');
```
DanielSLew, line 313: Where does `options.userId` come from? Is that something configured on the top level?

intojhanurag: options.userId is just a field on the ResolveToolProviderToolsOptions type. Nothing at the top level sets it.
The only place it gets set is the Composio e2e tests, which call the provider directly with 
{ userId: 'default' }. I kept it as a fallback so those tests keep working.

DanielSLew (approved):

PR #16123, opened 2026-05-04
fix: preserve text when filtering tool-only parts

### LINKED ISSUE #16052: [BUG] ToolCallFilter drops messages with top-level text when all tool parts are filtered
## Bug

`ToolCallFilter` can drop an assistant message that still has valid top-level text content.

When a message uses v2 content with:

- `content.content` containing assistant text, and
- `content.parts` containing only filtered `tool-invocation` parts,

the processor removes all parts and then returns `null` for the whole message. The top-level text fallback is lost.

## Expected behavior

Filtering tool invocation parts should not remove a message that still has non-tool text content.

For example, after filtering tool history, this message should survive with its text:

```ts
{
  role: 'assistant',
  content: {
    format: 2,
    content: 'I found three relevant papers.',
    parts: [
      {
        type: 'tool-invocation',
        toolInvocation: {
          state: 'result',
          toolCallId: 'call-1',
          toolName: 'search_papers',
          args: { query: 'attention mechanisms' },
          result: { papers: ['paper-1', 'paper-2', 'paper-3'] },
        },
      },
    ],
  },
}
```

## Current behavior

The message is removed entirely.

A focused regression test against `ToolCallFilter.processInput()` currently fails with:

```text
AssertionError: expected [] to have a length of 1 but got +0
```

## Why this happens

Verified against latest `origin/main` at commit:

```text
9efa13f3575d681f3231087de80c480b7408d64f
```

In `packages/core/src/processors/processors/tool-call-filter.ts`, both filtering paths drop the message when the filtered parts array is empty:

- `filterAllToolCalls()` returns `null` when `nonToolParts.length === 0`: https://github.com/mastra-ai/mastra/blob/9efa13f3575d681f3231087de80c480b7408d64f/packages/core/src/processors/processors/tool-call-filter.ts#L145-L157
- `filterSpecificToolCalls()` returns `null` when `filteredParts.length === 0`: https://github.com/mastra-ai/mastra/blob/9efa13f3575d681f3231087de80c480b7408d64f/packages/core/src/processors/processors/tool-call-filter.ts#L214-L242

Those checks only consider `parts`, not the top-level `content.content` text that is later spread into `updatedContent`.

## Minimal reproduction shape

A focused test can be added near `packages/core/src/processors/processors/tool-call-filter.test.ts`:

```ts
it('preserves top-level content when all tool invocation parts are filtered', async () => {
  const filter = new ToolCallFilter();
  const messages: MastraDBMessage[] = [
    {
      id: 'assistant-with-text-fallback',
      role: 'assistant',
      content: {
        format: 2,
        content: 'I found three relevant papers.',
        parts: [
          {
            type: 'tool-invocation',
            toolInvocation: {
              state: 'result',
              toolCallId: 'call-1',
              toolName: 'search_papers',
              args: { query: 'attention mechanisms' },
              result: { papers: ['paper-1', 'paper-2', 'paper-3'] },
            },
          },
        ],
      },
      createdAt: new Date(),
    },
  ];

  const messageList = new MessageList();
  messageList.add(messages, 'input');

  const result = await filter.processInput({
    messages,
    messageList,
    abort: ((reason?: string) => {
      throw new Error(reason || 'Aborted');
    }) as (reason?: string) => never,
  });

  const resultMessages = Array.isArray(result) ? result : result.get.all.db();
  expect(resultMessages).toHaveLength(1);
  expect(resultMessages[0]?.content.content).toBe('I found three relevant papers.');
});
```

Current result: the test fails because `resultMessages` is `[]`.

## Related issues

This is related to the general invariant in #16047 that processors should not silently erase all meaningful model input, but it is a different processor and a different failure path.

It is also different from #16039, which is about preserving compact `toModelOutput` history. This issue is about preserving existing top-level assistant text when tool parts are filtered.

## Verification

- [x] I checked the latest `origin/main`.
- [x] I reproduced the behavior with a focused test.
- [x] I searched existing issues and did not find a duplicate.

arjunrawal1:
## Summary
- keep messages with top-level text content after `ToolCallFilter` removes all tool-invocation parts
- cover both default all-tool filtering and named-tool filtering paths

Fixes #16052

## Validation
- `git diff --check`
- `corepack pnpm --filter ./packages/core test:unit src/processors/processors/tool-call-filter.test.ts` could not run in the fresh clone because dependencies are not installed (`vitest` not found)

## ELI5 Summary
When a message contains some spoken text and some tool instructions, a filter was accidentally throwing away the entire message if all the tool parts got filtered out. This fix makes sure the spoken text stays in the message even if all the tools get removed.

### Implementation (`tool-call-filter.ts`)
Modified `ToolCallFilter` to preserve messages containing top-level text content when all tool-invocation parts are filtered out:

- **`filterAllToolCalls` method**: Changed the early return condition from `if (nonToolParts.length === 0) return null` to only return `null` when both there are no non-tool parts AND no text content exists. This ensures messages with text fallback are preserved.
- **`filterSpecificToolCalls` method**: Removed the early return `if (filteredParts.length === 0) return null`, allowing the existing downstream logic (`hasNoToolParts && hasNoTextContent`) to determine whether to discard the message. This ensures text-only messages survive filtering.

### Tests (`tool-call-filter.test.ts`)
Added comprehensive test coverage with two new Vitest cases:
1. Default behavior: verifies `ToolCallFilter` preserves top-level text content while filtering all tool-invocation parts, resulting in empty `parts` array
2. Named tool filtering: verifies the same preservation behavior when excluding specific tool names via the `exclude` parameter

Both tests assert that resulting messages retain expected text content and confirm `parts` becomes an empty array when all tool-related content is removed.

## Related Issue
Fixes #16052 – `ToolCallFilter` was incorrectly dropping assistant messages that had top-level text content (`content.content`) but whose `content.parts` consisted entirely of tool-invocation parts.

### DIFF
--- a/packages/core/src/processors/processors/tool-call-filter.test.ts
+++ b/packages/core/src/processors/processors/tool-call-filter.test.ts
@@ -121,6 +121,50 @@ describe('ToolCallFilter', () => {
       }
     });
 
+    it('should preserve top-level text when all tool parts are filtered', async () => {
+      const filter = new ToolCallFilter();
+
+      const messages: MastraDBMessage[] = [
+        {
+          id: 'assistant-with-text-fallback',
+          role: 'assistant',
+          content: {
+            format: 2,
+            content: 'I found three relevant papers.',
+            parts: [
+              {
+                type: 'tool-invocation' as const,
+                toolInvocation: {
+                  state: 'result' as const,
+                  toolCallId: 'call-1',
+                  toolName: 'search_papers',
+                  args: { query: 'attention mechanisms' },
+                  result: { papers: ['paper-1', 'paper-2', 'paper-3'] },
+                },
+              },
+            ],
+          },
+          createdAt: new Date(),
+        },
+      ];
+
+      const messageList = new MessageList();
+      messageList.add(messages, 'input');
+
+      const result = await filter.processInput({
+        messages,
+        messageList,
+        abort: mockAbort,
+      });
+
+      const resultMessages = Array.isArray(result) ? result : result.get.all.db();
+      expect(resultMessages).toHaveLength(1);
+      expect(resultMessages[0]!.content).toMatchObject({
+        content: 'I found three relevant papers.',
+        parts: [],
+      });
+    });
+
     it('should handle messages without tool calls', async () => {
       const filter = new ToolCallFilter();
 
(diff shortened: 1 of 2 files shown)


intojhanurag:
Closing in respect of #16077

PR #16141, opened 2026-05-04
perf(memory): reuse per-step token count and cross-thread context

### LINKED ISSUE #15677: ObservationalMemoryProcessor.processInputStep blocks every agentic loop step with DB reads and token counting even when far below thresholds
### Describe the Bug

Every agentic loop step, `ObservationalMemoryProcessor.processInputStep` unconditionally executes:

1. `this.turn.start(this.memory)` — reads **all** thread messages from DB
2. `step.prepare()` — computes token counts
3. `tokenCounter.countMessagesAsync(allDbMsgs)` — re-counts all messages

These are blocking `await` calls on every step, even for a **single-message conversation**, causing **2–4 seconds latency overhead per step** even when the configured thresholds (e.g. `messageTokens: 280,000`) are far from being reached.

**Observed in Langfuse traces**: `processor_run Observational Memory` span appears with 2–4s latency on every agentic loop step. The span has `usageDetails: {}` (no LLM calls inside), confirming it is pure DB I/O + token counting overhead.

**Expected behavior**: When message token count is clearly below the `bufferActivation` threshold, the processor should short-circuit the heavy DB reads and token counting, or at minimum perform them asynchronously without blocking the main agentic loop.

### Steps To Reproduce

1. Configure an agent with `observationalMemory` enabled and high thresholds (e.g. `messageTokens: 280_000`)
2. Send a single-message conversation to the agent
3. Enable Langfuse observability tracing
4. Observe the Langfuse trace — `processor_run Observational Memory` span appears with **2–4 seconds latency** on every agentic loop step
5. Note: the span has `usageDetails: {}` (zero tokens) confirming no LLM call was made — just DB I/O + token counting

https://github.com/mastra-ai/mastra

### Expected Behavior

When the conversation is a single message and total token count is far below `bufferActivation` threshold, `processInputStep` should either:
- **Short-circuit** early after a quick lightweight check (e.g. check cached token count before loading all messages), OR
- **Run the DB reads non-blocking** so they don't add latency to the main agentic loop

A single-message conversation should not incur 2–4 seconds of mandatory overhead per step due to unconditional DB reads.

### Environment Information

```block
- @mastra/memory: 1.27.0
- @mastra/core: 1.27.0
- LLM Provider: OpenAI (gpt-5, gpt-5-mini)
- Storage: LibSQL (local file)
- Runtime: Bun 1.x
- OS: macOS
```

### Verification

intojhanurag:
## Description

Removes duplicate work performed every agentic loop step inside `ObservationalMemoryProcessor.processInputStep`. Both the message token count and the resource-scope cross-thread context were being computed once inside `getStatus(...)` and then re-computed by the processor, doubling per-step latency for no behavioral gain. This PR threads the already-computed values through the step context so each is computed at most once per step.

- `getStatus(...)` now returns `otherThreadsContext` alongside the token tally so callers can reuse it
- `ObservationStep.prepare()` reuses `statusSnapshot.otherThreadsContext` in resource scope and falls back to `refreshOtherThreadsContext()` only when the snapshot did not compute one
- `processor.ts` post-prepare block drops its own `countMessagesAsync(allDbMsgs)` and `countString(otherThreadsContext)` and reads `ctx.status.pendingTokens` instead
- `StepContext.status` extended with optional `otherThreadsContext` so downstream consumers can read the value without a second computation
- Tests added in `observational-memory.test.ts` covering the dedup paths

Note: the original report claimed `turn.start(memory)` runs every step, which is not the case (it is gated to once per turn at `processor.ts:165`). This PR addresses the real per-step duplication identified in the issue thread by @truffle-dev. It does not by itself fully explain a 2-4s per-step latency without the reporter confirming `scope` and other-thread count.

## Related Issue(s)

Fixes #15677

## Type of Change

- [ ] Documentation update
- [ ] Code refactoring
- [x] Performance improvement
- [ ] Test update

## Checklist

## ELI5

The processor was doing expensive calculations (like counting tokens and checking other threads) multiple times every single step, adding 2-4 seconds of unnecessary delay. This PR makes it compute those values once and reuse them throughout the step, eliminating the duplicate work and speeding things up significantly.

## Summary

This PR optimizes the observational memory processor by eliminating duplicate per-step token counting and context fetching. The optimization threads already-computed values through the step context so each expensive operation happens at most once per step, removing 2-4 seconds of latency on simple agent interactions.

### Key Changes

**Core Logic Refactoring:**
- `getStatus()` now returns `otherThreadsContext` alongside the token tally, enabling downstream consumers to reuse it instead of re-fetching
- `ObservationStep.prepare()` conditionally reuses `statusSnapshot.otherThreadsContext` when available for resource scope, falling back to `refreshOtherThreadsContext()` only when needed
- `processor.ts` post-prepare block now reads `ctx.status.pendingTokens` directly instead of re-calling `countMessagesAsync(allDbMsgs)` and `countString(otherThreadsContext)`

**Type System Updates:**
- Extended `StepContext.status` with optional `otherThreadsContext?: string` field to pass fetched context through the step pipeline

**Test Coverage:**
- Added integration test verifying `processInputStep` persists `pendingMessageTokens` by reusing the pending token count path (asserting `countMessagesAsync` is called exactly once)
- Added integration test verifying the processor reuses resource-scope context loaded during status checks (asserting `getOtherThreadsContext` is called once)
- Added `saveTestThread` and `createBasicMockModel` test helpers for seeding thread records and creating deterministic mock model responses

### Impact

Addresses #15677 by eliminating the blocking DB reads and token counting calls that were occurring unconditionally on every agentic loop step. The deduplication ensures single-message conversations and cases where total tokens are clearly below thresholds no longer incur the full latency penalty of these expensive operations.

### DIFF
--- a/packages/memory/src/processors/observational-memory/observation-turn/step.ts
+++ b/packages/memory/src/processors/observational-memory/observation-turn/step.ts
@@ -246,7 +246,9 @@ export class ObservationStep {
     }
 
     // ── Refresh cross-thread context (resource scope) ──────────
-    const otherThreadsContext = await this.turn.refreshOtherThreadsContext();
+    const otherThreadsContext =
+      om.scope === 'resource' ? statusSnapshot.otherThreadsContext : await this.turn.refreshOtherThreadsContext();
+    this.turn.context.otherThreadsContext = otherThreadsContext;
 
     // ── Build system messages (one per cache-stable chunk) ────
     const systemMessage = await om.buildContextSystemMessages({
@@ -282,6 +284,7 @@ export class ObservationStep {
         pendingTokens: statusSnapshot.pendingTokens,
         threshold: statusSnapshot.threshold,
         effectiveObservationTokensThreshold: statusSnapshot.effectiveObservationTokensThreshold,
+        otherThreadsContext,
         shouldObserve: statusSnapshot.shouldObserve,
         shouldBuffer: statusSnapshot.shouldBuffer,
         shouldReflect: statusSnapshot.shouldReflect,
--- a/packages/memory/src/processors/observational-memory/processor.ts
+++ b/packages/memory/src/processors/observational-memory/processor.ts
@@ -256,12 +256,7 @@ export class ObservationalMemoryProcessor implements Processor<'observational-me
         });
 
         // ── Token persistence (processor-specific) ──────────
-        const allDbMsgs = messageList.get.all.db();
-        const tokenCounter = this.engine.getTokenCounter();
-        const contextTokens = await tokenCounter.countMessagesAsync(allDbMsgs);
-        const otherThreadsContext = this.turn.context.otherThreadsContext;
-        const otherThreadTokens = otherThreadsContext ? tokenCounter.countString(otherThreadsContext) : 0;
-        const finalTotalPending = contextTokens + otherThreadTokens;
+        const finalTotalPending = ctx.status.pendingTokens;
 
         await this.engine
           .getStorage()
--- a/.changeset/some-bears-dream.md
+++ b/.changeset/some-bears-dream.md
@@ -1,0 +1,5 @@
+---
+'@mastra/memory': patch
+---
+
+Improved observational memory step performance by reusing token and resource context checks that were already computed during step preparation.
--- a/packages/memory/src/processors/observational-memory/__tests__/observational-memory.test.ts
+++ b/packages/memory/src/processors/observational-memory/__tests__/observational-memory.test.ts
@@ -145,6 +145,31 @@ function createInMemoryStorage(): InMemoryMemory {
   return new InMemoryMemory({ db });
 }
 
+async function saveTestThread(storage: InMemoryMemory, threadId: string, resourceId: string) {
+  await storage.saveThread({
+    thread: {
+      id: threadId,
+      resourceId,
+      title: threadId,
+      createdAt: new Date('2025-01-01T08:00:00Z'),
+      updatedAt: new Date('2025-01-01T08:00:00Z'),
+      metadata: {},
+    },
+  });
+}
+
+function createBasicMockModel() {
+  return new MockLanguageModelV2({
+    doGenerate: async () => ({
+      rawCall: { rawPrompt: null, rawSettings: {} },
+      finishReason: 'stop' as const,
+      usage: { inputTokens: 10, outputTokens: 5, totalTokens: 15 },
+      content: [{ type: 'text' as const, text: 'ok' }],
+      warnings: [],
+    }),
+  });
+}
+
 function createStreamCapableMockModel(config: Record<string, any>) {
   if (config.doGenerate && !config.doStream) {
     const originalDoGenerate = config.doGenerate;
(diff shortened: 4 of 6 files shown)


intojhanurag:
@coderabbitai review this whole PR

truffle-dev:
The dedup shape here matches what the trace surfaced: `ctx.status.pendingTokens` already accounts for `contextWindowTokens + otherThreadTokens`, so the post-prepare recount in `processor.ts` was paying for work `getStatus` had just done. Same story for `otherThreadsContext` in resource scope. Tests cover both paths via the spy on `countMessagesAsync` and `getOtherThreadsContext`.

One thing worth checking that's outside this PR's scope and may matter for the 2-4s number @LengYXin reported: the post-prepare `getOrCreateRecord` at `processor.ts` (the second call after `step.prepare()` returns) is intentional, because it picks up fire-and-forget `buffer()` writes that may have landed during prepare. So that DB read stays per-step. With many threads in resource scope, the first `getOtherThreadsContext` pass inside `getStatus` itself is a `listThreads` + per-thread `listMessages` + `formatUnobservedContextBlocks` chain. That's the load this dedup PR doesn't shave; it just stops paying for it twice.

For the residual latency question, the most useful thing @LengYXin could share alongside the scope/thread count is a single per-step `console.time/timeEnd` around the `om.getStatus(...)` call at `step.ts` (or whatever the equivalent is on their fork) versus the post-prepare block. If `getStatus` alone is 1-2s with N other threads, the bottleneck is in `formatUnobservedContextBlocks` / per-thread `listMessages`, not in the duplication this PR removes.

packages/memory/src/processors/observational-memory/processor.ts
```diff
@@ -256,12 +256,7 @@ export class ObservationalMemoryProcessor implements Processor<'observational-me
         });
 
         // ── Token persistence (processor-specific) ──────────
-        const allDbMsgs = messageList.get.all.db();
-        const tokenCounter = this.engine.getTokenCounter();
-        const contextTokens = await tokenCounter.countMessagesAsync(allDbMsgs);
-        const otherThreadsContext = this.turn.context.otherThreadsContext;
-        const otherThreadTokens = otherThreadsContext ? tokenCounter.countString(otherThreadsContext) : 0;
-        const finalTotalPending = contextTokens + otherThreadTokens;
```
TylerBarnes, line 264: we want to keep this one, token counter uses cached counts from metadata so it's cheap to call again. it runs it again in case anything has changed

packages/memory/src/processors/observational-memory/observation-turn/step.ts
```diff
@@ -246,7 +246,9 @@ export class ObservationStep {
     }
 
     // ── Refresh cross-thread context (resource scope) ──────────
-    const otherThreadsContext = await this.turn.refreshOtherThreadsContext();
+    const otherThreadsContext =
+      om.scope === 'resource' ? statusSnapshot.otherThreadsContext : await this.turn.refreshOtherThreadsContext();
```
TylerBarnes, line 250: refreshOtherThreadsContext already does this internally, so I don't think this makes a difference

TylerBarnes (changes requested):
I don't think these changes prevent any work, token counter already caches on metadata to prevent duplicate counting, and refreshOtherThreadsContext already only does the lookup in resource scope

PR #16149, opened 2026-05-04
perf(convex): batch bulk storage mutations

### LINKED ISSUE #16147: Optimize Convex storage bulk mutation throughput
`stores/convex/src/server/storage.ts` has several Convex bulk mutation paths that await indexed lookups and writes/deletes one item at a time. This creates a database waterfall where latency grows linearly with the number of records in the request.

## Affected Paths

- Typed table `batchInsert`
- Typed table `deleteMany`
- Typed table `clearTable` / `dropTable`
- Vector table `batchInsert`
- Vector table `deleteMany`
- Vector table `clearTable` / `dropTable`
- Generic fallback table `batchInsert`
- Generic fallback table `deleteMany`
- Generic fallback table `clearTable` / `dropTable`

## Current Behavior

`deleteMany` currently resolves each id with an indexed query, waits for that lookup, deletes the matching document, waits for that delete, and only then moves to the next id.

`batchInsert` has the same shape: for each record, it waits for the indexed lookup and then waits for the patch/insert before starting the next record.

`clearTable` and `dropTable` already limit each mutation call to a small page of documents, which is good for Convex mutation limits, but each document in that page is still deleted sequentially.

## Why This Matters

Bulk storage mutations should not behave like `N * database round-trip latency` when the operations are independent. The current implementation is especially expensive for memory/message cleanup, vector index maintenance, and fallback document-table cleanup because those paths may operate on many ids or records.

The optimization should keep Convex mutation limits in mind by using bounded concurrency, not unbounded `Promise.all()` over arbitrary input sizes.

## Desired Change

- Add a shared bounded batching helper for independent storage mutation work.
- Keep the existing page size of 25 for clear/drop operations.
- Use that same 25-item bound for concurrent lookup/write/delete batches.
- Deduplicate `deleteMany` ids before lookup so repeated ids do not issue repeated queries or deletes.
- Coalesce duplicate `batchInsert` records without changing observed write semantics:
  - Typed tables should merge duplicate records by id before patch/insert, preserving sequential patch merge behavior.
  - Vector and generic fallback tables should keep the last duplicate record because those paths replace the vector payload or stored generic record.
- Apply the behavior consistently across typed, vector, and generic fallback storage handlers.

## Edge Cases To Cover

- Empty `deleteMany` id list does not query or delete.
- Duplicate `deleteMany` ids are looked up once and deleted once.
- Missing ids are ignored without failing the request.
- Lookup and write/delete concurrency is capped at 25.
- `clearTable` / `dropTable` delete only the current 25-document page and preserve `hasMore`.
- Vector deletes and inserts stay scoped by `indexName`.
- Generic fallback deletes and inserts stay scoped by `table` and `primaryKey`.
- Typed `batchInsert` duplicate records preserve merge semantics.
- Vector/generic `batchInsert` duplicate records keep the last replacement.

## Benchmark Baseline And Target

Local benchmark method:

- Compare compiled baseline `HEAD` against compiled optimized working tree.
- Use the actual compiled Convex storage handler.
- Use a mocked Convex ctx that adds 5ms latency to each DB read/write.
- Run 100 records per operation, 7 measured iterations after warmup.

Baseline results:

| Path | Baseline avg | Baseline concurrency |
| --- | ---: | ---: |
| `deleteMany` | 1024.4ms | reads 1, writes 1 |
| `batchInsert` | 1023.5ms | reads 1, writes 1 |
| `clearTable` | 133.0ms | reads 1, writes 1 |

Target result after bounded batching:

| Path | Optimized avg | Optimized concurrency |
| --- | ---: | ---: |
| `deleteMany` | about 41ms | reads 25, writes 25 |
| `batchInsert` | about 41ms | reads 25, writes 25 |
| `clearTable` | about 10ms | reads 1, writes 25 |

This is a focused local latency benchmark, not a production Convex deployment benchmark. It is intended to prove that the serial await waterfall has been removed and bounded concurrency is enforced.

## Verification

Run the narrow Convex package checks:

- `pnpm --filter ./stores/convex exec vitest run src/server/storage.test.ts`
- `pnpm --filter ./stores/convex test`
- `pnpm --filter ./stores/convex lint`
- `pnpm --filter ./stores/convex build:lib`
- `git diff --check`

mbenhamd:
Closes #16147.

This PR removes the serial await waterfall from Convex storage bulk mutations by batching independent lookup, insert, patch, and delete work in bounded chunks of 25. It applies the same behavior across typed tables, vector tables, and the generic fallback table, while preserving the existing clear/drop page size and `hasMore` behavior.

`deleteMany` now deduplicates ids before indexed lookup. `batchInsert` now coalesces duplicate records before writing: typed tables merge duplicate records by id to preserve the previous sequential patch behavior, while vector and generic fallback tables keep the last duplicate because those paths replace the stored payload.

The local benchmark compared compiled baseline `HEAD` against the compiled optimized branch using the actual storage handler and a mocked Convex ctx with 5ms latency per DB read/write. This is a focused local latency benchmark, not a production Convex deployment benchmark.

| Path | Baseline avg | Optimized avg | Change | Concurrency |
| --- | ---: | ---: | ---: | --- |
| `deleteMany` with 100 ids | 1024.4ms | 41.2ms | ~24.9x faster | reads/writes `1 -> 25` |
| `batchInsert` with 100 records | 1023.5ms | 41.4ms | ~24.7x faster | reads/writes `1 -> 25` |
| `clearTable` with 100 docs available | 133.0ms | 10.3ms | ~12.9x faster | deletes `1 -> 25` |

Single-item bulk calls were also checked against the same benchmark setup and did not show a measurable slowdown: `deleteMany` stayed at 10.3ms, `batchInsert` moved from 10.2ms to 10.3ms, and `clearTable` stayed at 10.2ms.

Verified with `pnpm --filter ./stores/convex exec vitest run src/server/storage.test.ts`, `pnpm --filter ./stores/convex test`, `pnpm --filter ./stores/convex lint`, `pnpm --filter ./stores/convex build:lib`, and `git diff --check`.

### DIFF
--- a/.changeset/lazy-doors-listen.md
+++ b/.changeset/lazy-doors-listen.md
@@ -1,0 +1,5 @@
+---
+'@mastra/convex': patch
+---
+
+Improved Convex bulk insert and delete throughput.
--- a/stores/convex/src/server/storage.test.ts
+++ b/stores/convex/src/server/storage.test.ts
@@ -1,29 +1,43 @@
 import { TABLE_WORKFLOW_SNAPSHOT } from '@mastra/core/storage/constants';
+import type { GenericId } from 'convex/values';
 import { describe, expect, it, vi } from 'vitest';
 
-import { handleTypedOperation } from './storage';
+import type { StorageRequest, StorageResponse } from '../storage/types';
+import { handleTypedOperation, mastraStorage } from './storage';
+
+type TypedOperationCtx = Parameters<typeof handleTypedOperation>[0];
+type StorageHandlerForTest = typeof mastraStorage & {
+  _handler: (ctx: TypedOperationCtx, request: StorageRequest) => Promise<StorageResponse>;
+};
+type TestDoc = { _id: GenericId<string>; id?: string };
+type TestQueryBuilder = {
+  eq: (field: string, value: string) => TestQueryBuilder;
+};
+
+const asConvexId = (id: string) => id as GenericId<string>;
 
 describe('mastraStorage typed load', () => {
   it('uses by_workflow_run for workflow snapshot composite keys', async () => {
     const workflowRun = {
+      _id: asConvexId('snapshot-doc'),
       workflow_name: 'workflow-a',
       run_id: 'run-1',
       snapshot: {},
     };
 
-    const builder = {
+    const builder: TestQueryBuilder = {
       eq: vi.fn((_field: string, _value: string) => builder),
     };
     const unique = vi.fn(async () => workflowRun);
     const take = vi.fn(async () => {
       throw new Error('load should not scan workflow snapshots for composite keys');
     });
-    const withIndex = vi.fn((_indexName: string, queryBuilder: (q: typeof builder) => typeof builder) => {
+    const withIndex = vi.fn((_indexName: string, queryBuilder: (q: TestQueryBuilder) => TestQueryBuilder) => {
       queryBuilder(builder);
       return { unique, take };
     });
     const query = vi.fn(() => ({ withIndex, take }));
-    const ctx = { db: { query } } as any;
+    const ctx = { db: { query } } as unknown as TypedOperationCtx;
 
     const result = await handleTypedOperation(ctx, 'mastra_workflow_snapshots', {
       op: 'load',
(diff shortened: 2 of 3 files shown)


mbenhamd:
Addressing CodeRabbit's nitpick about records without `id` in `batchInsert`: this PR intentionally preserves the existing batch behavior. Before this change, `batchInsert` already skipped records with missing/falsy ids while single `insert` threw. Changing `batchInsert` to throw would be a behavior change outside this performance PR, so the coalescing helpers keep the same lenient batch semantics while removing the serial DB waterfall.

wardpeet (approved):

PR #16152, opened 2026-05-04
fix(server): read Studio mode from request context

### LINKED ISSUE #16151: Retrieve Studio serialization mode from request context
## Problem

`GET /agents/:agentId` already receives `requestContext`, and `formatAgent` already has Studio-specific serialization behavior. The route currently hard-codes `isStudio` to `false`, which leaves that behavior unreachable from this handler even when the request context identifies the request as coming from Studio.

## Impact

The TODO makes the handler harder to reason about because the serialization mode is disconnected from the context already available in the handler. It also means dynamic agent serialization cannot use the Studio placeholder behavior for missing request-context values through this endpoint.

This is a code-health and correctness cleanup, not a performance optimization.

## Desired change

Read a strict boolean Studio flag from `requestContext` and pass it to `formatAgent`. Preserve existing behavior when the flag is missing or not exactly `true`.

## Edge cases

- Missing `isStudio` should behave like the current route and serialize as non-Studio.
- `isStudio: true` should enable the existing placeholder request-context proxy.
- Non-boolean values such as `"true"` or `1` should not enable Studio mode.
- Existing version selection behavior through `status`, `versionId`, and request context should remain unchanged.

## Verification target

Add focused handler coverage for the Studio flag path and non-true edge cases, then run the package-scoped server tests, lint, and build.

mbenhamd:
Closes #16151.

This PR removes the hard-coded `isStudio = false` from `GET /agents/:agentId` and reads the flag from `requestContext` with a strict `=== true` check before passing it into `formatAgent`.

The behavior stays unchanged unless the context explicitly sets `isStudio` to `true`. I added handler coverage for the Studio placeholder path and for false, string, and number edge cases so non-boolean values do not opt into Studio mode.

Verified with `pnpm --filter ./packages/server exec vitest run src/server/handlers/agent.test.ts`, `pnpm --filter ./packages/server test`, `pnpm --filter ./packages/server lint`, `pnpm --filter ./packages/server build:lib`, and `git diff --check`.

## ELI5

The server stopped pretending Studio mode is always off and now checks a server-set flag in the request context; Studio-specific output only happens when that flag is exactly true.

## Summary

This PR removes the hard-coded isStudio = false in the GET /agents/:agentId handler and instead reads Studio mode from the server-populated request context using a strict === true check. Studio-specific serialization (placeholder proxying for instructions) runs only when the server sets MASTRA_IS_STUDIO_KEY === true; all other values (missing, false, "true", 1, or user-supplied isStudio) do not enable Studio mode.

### Changes

- packages/server/src/server/handlers/agents.ts
  - Added getIsStudioFromContext(requestContext) which returns requestContext.get(MASTRA_IS_STUDIO_KEY) === true.
  - GET_AGENT_BY_ID_ROUTE now uses that derived isStudio value when calling formatAgent; the proxied request-context behavior is applied only for agent.getInstructions(...) when isStudio === true, while subsequent serialization (LLM defaults, sub-agent serialization, etc.) uses the original requestContext.
  - Added mergeBodyRequestContext to merge non-reserved keys from body-provided requestContext into server RequestContext only when the server context does not already have a value.
- packages/server/src/server/constants.ts
  - Added and exported constants: MASTRA_IS_STUDIO_KEY, MASTRA_CLIENT_TYPE_HEADER, MASTRA_STUDIO_CLIENT_TYPE.
  - Added isReservedRequestContextKey and isStudioClientTypeHeader helpers; reserved keys set updated to include the studio flag.
- server adapter & adapter tests
  - MastraServer.applyRequestMetadataToContext hook added; adapters call it during context construction to set MASTRA_IS_STUDIO_KEY when the client-type header indicates Studio.
  - Adapters now filter reserved request-context keys when merging body/params context; tests updated to assert reserved keys (including MASTRA_IS_STUDIO_KEY) are stripped from client input and set from headers where appropriate.
- packages/server/src/server/handlers/agent.test.ts
  - Tests added/updated to cover:
    - MASTRA_IS_STUDIO_KEY === true enables Studio placeholder behavior for instructions (placeholder interpolation).
    - MASTRA_IS_STUDIO_KEY false, 'true' (string), or 1 (number) do not enable Studio behavior.
    - A user-provided isStudio key (not the reserved MASTRA_IS_STUDIO_KEY) does not enable Studio placeholder behavior.
- Other: changeset added that documents Studio serialization behavior and bumps related packages; playground now supplies studio header when appropriate.

### Behavior

- Studio serialization is enabled only when the server-populated request context contains MASTRA_IS_STUDIO_KEY === true.
- Non-boolean or client-supplied values do not enable Studio mode.
- Reserved request-context keys (including mastra__isStudio) cannot be set from client body/query; adapters set them from headers.
- Existing version-selection and other handler logic remain unchanged.

### Verification

- Added handler test coverage for Studio true and non-true edge cases.
- Local verification included package-scoped server tests, lint, library build, and a git diff check.

### Related issue

Closes #16151


packages/server/src/server/handlers/agents.ts
```diff
@@ -86,6 +86,12 @@ function stashVersionOverrides(ctx: RequestContext, versions: VersionOverrides |
   }
 }
 
+const IS_STUDIO_CONTEXT_KEY = 'isStudio';
```
wardpeet: We should make it unique so it doesn't clash with the real context. Or am I wrong here? Also, shouldn't this be a header of some sort?

``` suggestion
const IS_STUDIO_CONTEXT_KEY = '__mastra_isStudio';
```

mbenhamd: Good call. I agree `isStudio` should not be a plain request-context key.

`requestContext` is user-controlled in several paths: adapters merge query/body `requestContext` into the runtime `RequestContext`. So `isStudio` can collide with a real app variable, and a non-Studio caller could set it accidentally or intentionally.

I’m going to move this behind an internal reserved key and keep the transport signal separate:

- use a reserved internal key following the existing convention, e.g. `mastra__isStudio`
- add that key to the reserved request-context filter so query/body `requestContext` cannot set it
- have adapter context middleware translate a Studio/client header into that internal key
- prefer `x-mastra-client-type: studio` over `x-mastra-dev-playground`, since the latter is dev/auth flavored and `x-mastra-client-type` is already the Cloud-neutral client signal
- narrow the placeholder context so it only affects instruction serialization, not workspace/skill/tool/default-option resolution

So yes: unique key internally, header as the external signal.

abhiaiyer91 (approved):

PR #16174, opened 2026-05-04
fix(tools): correctly skip output validation when suspendData is falsy

Akash504-ai:
## Description

Fixes an incorrect truthiness check for `suspendData` when determining whether to skip output validation.

Previously, falsy values such as `""`, `0`, or `false` would not skip validation, even though suspend was triggered. This caused unintended output validation errors when tools correctly suspended execution.

This change ensures validation is skipped based on existence (`null` check) rather than truthiness.

Additionally, tests have been added to cover edge cases where `suspendData` is falsy.

---

## Related Issue(s)

N/A

---

## Type of Change

* [ ] Documentation update
* [ ] Code refactoring
* [ ] Performance improvement
* [ ] Test update

---

## Checklist

## ELI5 (Explain Like I'm 5)

When a tool pauses execution and gives you some data—even if that data is empty or zero—the system should skip checking whether the result is valid. The old code only skipped this check if the data was "truthy" (like `true`, non-zero numbers, non-empty strings), which meant empty data would still get checked and cause errors. This PR fixes that by checking whether data exists at all, instead of checking whether it's truthy.

### Tool Execution Output Validation Logic

**packages/core/src/tools/tool.ts**

Updated the logic that determines when to skip output schema validation. The `skipOutputValidation` calculation now skips validation when `suspendData` is `null` (rather than falsy), meaning falsy-but-present values like empty strings (`""`), zero (`0`), or `false` will correctly trigger validation skipping when suspend is triggered with these values.

### Test Coverage

**packages/core/src/tools/tools.test.ts**

Added a new test suite covering suspend output validation edge cases with two test scenarios:
- Validates that output validation is skipped when `context.suspend('')` (empty string) is called
- Validates that output validation is skipped when `context.suspend(0)` (zero) is called

Both tests verify that the tool's `execute` method returns `undefined` when these falsy suspend values are used.


abhiaiyer91:
Closing! We are requiring issues for non Mastra core team PRs!

PR #16175, opened 2026-05-04
feat(server): add A2A push notifications

TheIsrael1:
This adds baseline A2A push notification support to the server. Tasks can now register push configs through message/send, message/stream, or the tasks/pushNotificationConfig/* methods, and the server will POST the current task snapshot to registered webhooks when a task reaches completed, failed, canceled, or input-required. Delivery is in-memory and best-effort for now, and it no longer blocks the response or final stream event while the webhook call is in flight.

Example:
```ts
await a2a.setTaskPushNotificationConfig({
  taskId: "task-123",
  pushNotificationConfig: {
    url: "https://example.com/a2a-webhook",
    token: "session-token",
  },
});
```

## ELI5

This PR lets tasks tell other services when their status changes by sending HTTP webhooks (push notifications), so external systems don't need to keep asking the server whether a task finished or needs input.

## Overview

Adds baseline Agent-to-Agent (A2A) push notification support: tasks can register webhook configs (via message/send, message/stream, or new tasks/pushNotificationConfig methods) and the server will POST full task snapshots to those webhooks when tasks transition to completed, failed, canceled, or input-required. Delivery is in-memory and best-effort; notifications are sent asynchronously after task state is persisted so slow/failing webhooks do not block responses or final stream events. Outbound delivery is pinned to validated hosts/addresses to mitigate DNS rebinding/SSRF risk.

## Key Changes

- New in-memory store and sender
  - InMemoryPushNotificationStore: per-(agentId,taskId) in-memory store with set/get/list/delete, normalizes config IDs, and returns structured clones.
  - DefaultPushNotificationSender: validates and normalizes destination URLs, rejects non-http(s), local/internal hostnames, and disallowed private/local IPs; performs DNS resolution and rejects if any resolved address is disallowed, then rewrites requests to the validated IP while preserving the original Host header. Supports optional allowedHosts allowlist, configurable token header (DEFAULT_PUSH_NOTIFICATION_TOKEN_HEADER = 'X-A2A-Notification-Token'), optional Bearer/Basic auth, injectable fetch and lookup, and a default timeout (5000ms). Concurrently POSTs task JSON to configured endpoints; non-2xx responses and delivery failures are logged and surfaced per-destination.

- Handler integration and APIs
  - message/send and message/stream accept optional pushNotificationConfig and persist it when provided.
  - Centralized saveTaskAndMaybeSendPushNotification helper saves task state and schedules notifications for status transitions: completed, failed, canceled, input-required.
  - New JSON-RPC endpoints for per-task push config CRUD:
    - handleSetTaskPushNotificationConfig
    - handleGetTaskPushNotificationConfig
    - handleListTaskPushNotificationConfig
    - handleDeleteTaskPushNotificationConfig
  - getAgentCardByIdHandler now advertises pushNotifications capability; pushNotificationStore/sender are threaded through cancellation and other routes.

- Delivery model & safety
  - Delivery is best-effort and in-memory; webhook calls are asynchronous and do not block client flows.
  - Safety: rejects localhost/single-label hosts, private IP ranges, and DNS-resolved private addresses; outbound requests are pinned to validated resolved addresses to mitigate DNS rebinding. There is an optional allowedHosts allowlist and injected lookup to control resolution.

- Tests
  - Tests for DefaultPushNotificationSender (unsafe URL and private-IP resolution blocking, pinning to resolved IP while preserving Host header).
  - Tests for InMemoryPushNotificationStore (collision-safety).
  - Extended a2a handler tests covering agent card push capability, persisting push configs via message/send, webhook delivery attempts and token header behavior, non-blocking failure behavior, and full CRUD for tasks/pushNotificationConfig (including handling missing tasks).

- Documentation/versioning
  - Added a .changeset entry for @mastra/server (minor) documenting the feature.

## Why this matters / reviewer notes

- Delivery is intentionally non-durable and best-effort (in-memory). If guaranteed delivery is required later, the design must be revisited to add persistence/retries.
- Security: reviewers should inspect the allowlist semantics, DNS lookup/timeouts, and the host/ip filtering/pinning logic (commit notes indicate explicit pinning to validated hosts to mitigate DNS rebinding).
- API design: wardpeet questioned whether push config registration should be server-side or provided by callers and suggested updating A2A client-js—consider whether registration belongs in client libraries or should be restricted/configured server-side.


wardpeet:
I'm not sure how to configure my server. A2A adds all necessary handlers that the protocol needs but shouldn't push config be done on the server side? Instead of setting it using `setTaskPushNotificationConfig`. Or is it the caller that actually needs to tell us? 

Should we also update a2a client-js parts?

PS: i did not read the a2a spec

TheIsrael1:
> I'm not sure how to configure my server. A2A adds all necessary handlers that the protocol needs but shouldn't push config be done on the server side? Instead of setting it using `setTaskPushNotificationConfig`. Or is it the caller that actually needs to tell us?
> 
> Should we also update a2a client-js parts?
> 
> PS: i did not read the a2a spec



> I'm not sure how to configure my server. A2A adds all necessary handlers that the protocol needs but shouldn't push config be done on the server side? Instead of setting it using `setTaskPushNotificationConfig`. Or is it the caller that actually needs to tell us?
> 
> Should we also update a2a client-js parts?
> 
> PS: i did not read the a2a spec

Fair points @wardpeet the a2a JS SDK client factory had it like this where you configure it from the client and that's the standard we followed here... and it's keyed by taskId. 

we can make it configurable from the server too as i have just exposed a server.a2a config for agent signing in a follow up PR here - https://github.com/mastra-ai/mastra/pull/16207 


so we can for sure expose that config too in a follow up PR

wardpeet:
We should add these things to client-js so we can set it. We don't have to set it as a config if that's not following the spec

wardpeet (approved):

TheIsrael1:
> We should add these things to client-js so we can set it. We don't have to set it as a config if that's not following the spec

Yeah the a2a js SDK is not necessarily the "spec" just an SDK too, I think we can for sure still make convenient patterns for 👏

Issue #16180, opened 2026-05-04
CLI lint should not execute next.config.js while checking serverExternalPackages


mbenhamd:
## Problem

The CLI Next.js lint rule currently parses `next.config.js` by extracting the `nextConfig` object with a regex and evaluating it with `eval()`.

Affected file:

`packages/cli/src/commands/lint/rules/nextConfigRule.ts`

Current behavior:

```ts
const configMatch = nextConfigContent.match(/const nextConfig = ({[\s\S]*?});/);
if (!configMatch?.[1]) {
  return null;
}
const configStr = configMatch[1].replace(/\n/g, "").replace(/\s+/g, " ");
return eval(`(${configStr})`);
```

Because `next.config.js` belongs to the project being linted, this lets project-controlled JavaScript run inside the Mastra CLI process.

## Impact

This is a real code execution issue in the CLI lint path. A malicious or compromised project can execute arbitrary Node.js code when a developer or CI job runs the Mastra lint command.

Potential impact includes reading local files, reading environment variables, accessing CI secrets available to the process, or modifying files in the workspace.

This is not remote unauthenticated execution. Exploitation requires running the CLI lint command against a malicious or compromised project config. The severity is still high for developer and CI environments because the executed code inherits the lint process permissions.

## Expected behavior

The lint rule should inspect `serverExternalPackages` without executing any code from `next.config.js`.

## Suggested fix

Use static parsing for `next.config.js` and only read supported literal config shapes, for example:

- `const nextConfig = { serverExternalPackages: ["@mastra/*"] }`
- `module.exports = { serverExternalPackages: ["@mastra/*"] }`
- `export default { serverExternalPackages: ["@mastra/*"] }`

The parser should reject or ignore dynamic expressions instead of evaluating them.

## Acceptance criteria

- Remove `eval()` from the Next.js config lint rule.
- Preserve lint behavior for simple literal `serverExternalPackages` configs.
- Add a regression test proving code embedded in `next.config.js` is not executed during linting.
- Add a changeset for the CLI package.

daneatmastra:
Thank you for reporting this issue! We have labeled it for trio-wp and we will look into it as soon as possible.

If you're experiencing an error, please provide a [minimal reproducible example](https://github.com/mastra-ai/mastra/blob/main/CONTRIBUTING.md#minimal-reproduction) whenever possible to help us resolve it quickly.

mikhael28:
[...]

PR #16184, opened 2026-05-04
fix(cli): parse next config without eval

mbenhamd:
What

The CLI Next.js lint rule no longer evaluates project next.config.js content while checking serverExternalPackages. It now parses the file with acorn and extracts supported literal string arrays.

Risk

The old eval path could execute arbitrary Node.js code when a developer or CI job ran Mastra lint against a malicious or compromised next.config.js. That code would inherit the lint process permissions, including access to local files and environment variables.

Solution

This replaces eval with static AST parsing for simple const nextConfig, module.exports, and export default object configs. Dynamic expressions are not evaluated. A regression test covers a config with an IIFE payload and verifies it does not run.

Validation

- pnpm --filter ./packages/cli test src/commands/lint/rules/nextConfigRule.test.ts
- pnpm --filter ./packages/cli lint
- pnpm --filter ./packages/cli typecheck

Note: pnpm build:cli was attempted and failed in @mastra/server#generate:route-types because zod-to-ts could not be resolved in that dependent package. The failure is outside the CLI diff.

Refs #16180

## ELI5 (Explain Like I'm 5)

The CLI used to run a project's next.config.js to read a setting, which could let that file run code. This PR reads the file safely by parsing its structure (without executing it), so nothing in the file can run during linting.

## Overview

Replaces eval-based reading of next.config.js in the CLI lint rule with static AST parsing using @babel/parser (matching the repo parser stack). The new logic extracts literal string arrays for serverExternalPackages from simple, common config forms and explicitly avoids evaluating any dynamic expressions or executing the file.

## Changes

- packages/cli/src/commands/lint/rules/nextConfigRule.ts
  - Replaced eval/regex approach with AST parsing via @babel/parser and @babel/types.
  - Collects top-level const object initializers (e.g., const nextConfig = {...}) and resolves module.exports or export default to those object literals or inline object literals.
  - Extracts serverExternalPackages only when its value is an array of string literals; ignores dynamic or computed values.
  - Adds an early-return optimization: skip parsing entirely when the file content does not include the identifier serverExternalPackages.
  - readNextConfig returns {} when the file contains no serverExternalPackages, and returns null on parse/read/resolution failures.
  - No external behavioral changes to the lint rule’s decision logic (still validates Mastra entries and logs errors).

- packages/cli/src/commands/lint/rules/nextConfigRule.test.ts
  - Adds tests verifying detection when serverExternalPackages includes Mastra packages.
  - Adds a regression test ensuring potentially malicious constructs (IIFE that would set a global marker) are not executed during parsing.

- .changeset/grumpy-places-notice.md
  - Adds a changeset documenting the fix as a patch-level release for mastra.

- packages/cli/package.json
  - Minor reordering of dependency entries (no version changes).

## Validation

- Ran: pnpm --filter ./packages/cli test src/commands/lint/rules/nextConfigRule.test.ts
- Ran: pnpm --filter ./packages/cli lint
- Ran: pnpm --filter ./packages/cli typecheck
- Note: pnpm build:cli failed due to an unrelated dependency resolution issue in @mastra/server (zod-to-ts).

## Risk / Security Impact

- Eliminates the previous security risk of executing arbitrary Node.js code from next.config.js during linting by switching to non-executing AST parsing.


packages/cli/src/commands/lint/rules/nextConfigRule.ts
```diff
@@ -1,18 +1,210 @@
...
+        return exportedConfig;
+      }
+    }
+  }
+
+  return null;
+}
+
+function parseNextConfig(nextConfigContent: string): NextConfig | null {
+  const program = parseProgram(nextConfigContent);
```
wardpeet, line 164: Can we short circuit with searching for the string serverExternalPackages

packages/cli/package.json
```diff
@@ -52,9 +52,10 @@
   "dependencies": {
     "@clack/prompts": "^1.1.0",
     "@expo/devcert": "^1.2.1",
-    "archiver": "^7.0.1",
     "@mastra/deployer": "workspace:^",
     "@mastra/loggers": "workspace:^",
+    "acorn": "^8.16.0",
```
wardpeet: Can we move to the babel parser? We use it everywhere else in the repo. Even if acorn is smaller and faster lets keep things consistent

mbenhamd:
Addressed the review feedback in two follow-up commits.

Switched the Next.js config lint parser from acorn to @babel/parser/@babel/types, using the same dependency specs already used elsewhere in the repo.

Added the serverExternalPackages string short-circuit before AST parsing so configs without that key avoid parser work while preserving the missing-config lint result.

Validation:
- pnpm --filter ./packages/cli test src/commands/lint/rules/nextConfigRule.test.ts
- pnpm --filter ./packages/cli typecheck
- pnpm --filter ./packages/cli lint

abhiaiyer91 (approved):

wardpeet (approved):

PR #16229, opened 2026-05-05
feat(core): add Agent signals

TylerBarnes:
## Summary

Adds Agent signals as a first-class way to send context into an agent thread.

This introduces the core runtime, server, and client-js APIs for sending a signal into either:

- an active thread run, where the signal is drained at a stream boundary and interjected into the loop, or
- an idle thread, where the signal starts the run directly as the initial input.

Example:

```ts
await agent.sendSignal(
  { type: 'user-message', contents: 'please also account for this' },
  { resourceId, threadId },
);
```

Consumers can also subscribe to a thread stream independently:

```ts
const subscription = await agent.subscribeToThread({ resourceId, threadId });

for await (const chunk of subscription.stream) {
  // handle raw stream chunks
}
```

## Stack

This is PR 1 of the Agent signals stack:

1. #16229 — Core Agent signal runtime, server routes, and client-js API
2. #16438 — Route system reminders through Agent signals
3. #16231 — Use Agent signals for MastraCode follow-up chat
4. #16338 — Use Agent signals for Studio/playground follow-up chat

## What changed

- Added signal types, persistence helpers, XML conversion for non-user context signals, and stream data-part conversion.
- Added `agent.sendSignal()` and `agent.subscribeToThread()` runtime APIs.
- Added server routes for sending signals and subscribing to thread streams.
- Added client-js helpers for the new signal and subscription routes.
- Drained pending signals at stream boundaries (`text-end`, `reasoning-end`, `tool-result`, `finish`) instead of mid-token text deltas.
- Rotated response message IDs before inserting interjected user input so assistant output is persisted after the new user message.
- Allowed `agent.stream()` to accept created signal inputs directly for idle signal runs.
- Kept signal metadata isolated in the persisted signal envelope so caller metadata does not get polluted by internal signal bookkeeping.
- Added abort signal plumbing for subscription stream processing.

## Test plan

- `pnpm --filter ./packages/core exec vitest run src/agent/__tests__/agent-signals.test.ts --bail 1 --reporter=dot`
- `pnpm --filter ./packages/server test src/server/handlers/agents.test.ts --bail 1 --reporter=dot`
- `pnpm --filter ./client-sdks/client-js test src/utils/process-mastra-stream.test.ts src/resources/agent.test.ts --bail 1 --reporter=dot`
- `pnpm build:server`
- `pnpm --filter ./client-sdks/client-js build`


TylerBarnes:
Updated this stack with the API/schema fixes: the server signal body now uses a typed structured message schema instead of , accepts , and route types were regenerated. I also addressed the highlighted review items for  thread serialization and  handling, and moved the client-js subscription abort processing down into this foundational PR.

TylerBarnes:
Updated this stack after rebasing onto current main. #16229 is no longer DIRTY; GitHub now reports mergeStateStatus=BLOCKED pending checks/review.\n\nAddressed the remaining high-signal signal API/review items:\n- Server signal schema is typed (no z.any()) for user-message MessageListInput variants, structured text/image/file parts, string arrays, and attributes passthrough.\n- Regenerated client-js route types after the rebase.\n- Client-js subscribeToThread returns processDataStream and passes AbortSignal through processMastraStream; reader cancellation is covered by tests.\n- resumeStream now serializes against active thread runs before preparing resumed runs.\n- SEND_AGENT_SIGNAL_ROUTE and SUBSCRIBE_AGENT_THREAD_ROUTE use HTTPException(400) for request validation failures.\n- Signal/XML tag and attribute names are validated before XML emission.\n- Internal DB signal bookkeeping is isolated under metadata.signal and no longer round-trips as user metadata.\n- ifIdle requires thread context, existing active-thread behavior is covered by the cross-agent queued signal test, and failed idle starts clean prepared abort state.\n- waitForActiveRun test helper is bounded with a 500ms deadline.\n\nLocal validation run on the rebased stack:\n- packages/core agent signal + signal UI conversion tests: 21/21\n- packages/server agents handler tests: 55/55\n- client-js process-mastra-stream + agent resource tests: 42/42\n- @mastra/react toUIMessage tests: 93/93\n- @mastra/playground-ui tests: 117/117\n- pnpm --filter ./packages/core check\n- pnpm build:server\n- pnpm --filter ./client-sdks/client-js build:lib

TylerBarnes (commented):
I found two additional issues worth addressing before merging:

1. **Cross-agent idle signals can start duplicate concurrent runs**

In `packages/core/src/agent/thread-stream-runtime.ts`, when `sendSignal()` is called with `ifIdle` while another agent is active on the same thread, it returns a new `runId` and starts `agent.stream()`, but does not reserve that returned run ID because `#activeThreadRunIds` already contains the other agent’s run. A second signal sent before the first active run finishes can’t attach to that returned run ID, and multiple `ifIdle` calls can all wait on the active run and then start together once it finishes. That seems to violate the intended invariant that only one thread run should be active/queued per thread.

2. **Signal UI data parts drop `attributes` after persistence/history conversion**

`signalToDataPart()` includes signal attributes, but the adapter reconstruction paths in `AIV4Adapter`/`AIV5Adapter` build non-user signal `data-*` UI parts without copying `attributes`. That means persisted signals with structured attributes lose them when converted back for UI/subscribers after reload/history conversion. This is likely important for signals like `system-reminder` where attributes may carry context such as type/path/enabled state.

Note: I tried the focused core signal test, but it failed before running because the local test setup artifact was missing (`packages/core/@internal/test-utils/setup`).

docs/src/content/en/reference/agents/agent.mdx
```diff
@@ -108,6 +108,143 @@ export const agent = new Agent({
...
+    console.log(chunk)
+  }
+})()
+
+await agent.sendSignal(
+  { type: 'user-message', contents: 'Use the latest customer note too.' },
+  {
+    resourceId: 'user-123',
+    threadId: 'thread-abc',
+    ifIdle: {
```
abhiaiyer91, line 134: streamOptions here for resource and thread are redundant no?

abhiaiyer91 (approved):
Just a small comment on ifIdle options

docs/src/content/en/reference/agents/agent.mdx
TylerBarnes: oh good call i'll pull em out!

PR #16233, opened 2026-05-05
fix(pg): keep agent listing resilient to jsonb scalar strings (#16224)

### LINKED ISSUE #16224: [BUG] @mastra/pg: agent listing crashes when model is stored as jsonb scalar string (one bad row breaks entire endpoint)
## Summary

`@mastra/pg`'s `parseJson` in agent listing endpoints crashes with **HTTP 500** when any row in `mastra_agent_versions.model` is stored as a **JSONB scalar string** (e.g. `"google/gemini-3-flash"`). One bad row breaks the entire listing for all agents (fail-fast).

## Environment

- `@mastra/pg`: 1.9.1
- `@mastra/core`: ^1.24.1
- Postgres: Neon (jsonb columns)
- Driver: `pg` (default behavior auto-parses jsonb)

## Reproduction

1. Create an agent version where `model` is written as a JSON scalar string. This can happen via `createVersion({ model: "anthropic/claude-haiku-4.5" })` because `createVersion` does `JSON.stringify(input.model)` without normalizing scalar inputs to objects (see `dist/index.js:3995`).
2. The DB ends up with `jsonb_typeof(model) = 'string'`.
3. Call any endpoint that lists agents (e.g. the Mastra Editor's `/api/stored/agents?status=published`).
4. Response:

```
HTTP/2 500
{"error":"Failed to parse JSON for field \"model\": Unexpected token 'g', \"google/gem\"... is not valid JSON"}
```

## Root cause

Two problems compound:

### 1. Double-deserialization in `parseJson` (read path)

`dist/index.js:3690`:

```js
parseJson(value, fieldName) {
  if (!value) return void 0;
  if (typeof value !== "string") return value;
  try { return JSON.parse(value); } catch (error) { ... throw }
}
```

For a `jsonb` column, the `pg` driver already deserializes the value:

| DB value (jsonb)                  | Driver returns (JS)                      |
| --------------------------------- | ---------------------------------------- |
| `{"name":"x","provider":"y"}`     | `{ name: "x", provider: "y" }` (object)  |
| `"google/gemini-3-flash"`         | `"google/gemini-3-flash"` (**JS string**) |

For the second case, `parseJson` sees `typeof === "string"` and runs `JSON.parse("google/gemini-3-flash")` — which throws on the unquoted `g`.

`parseJson` cannot distinguish "string containing JSON" (text column) from "JSON scalar string already deserialized" (jsonb column).

### 2. Fail-fast in `mapAgentRow` (listing endpoint)

`dist/index.js:4237`:

```js
model: this.parseJson(row.model, "model"),
```

`mapAgentRow` runs synchronously over every row in the listing. One row that throws kills the entire endpoint, hiding **all** agents from the Editor — even ones with valid `model`.

### 3. Asymmetric write path

`createVersion` (`dist/index.js:3995`) does:

```js
JSON.stringify(input.model)
```

with no type validation. Passing `input.model = "slug"` is silently accepted and stored as a jsonb scalar string, guaranteeing the read path will later fail.

## Impact

In our deployment, **9 of 21 agents** had model stored as scalar string (a mix of historical UPDATEs and edits via the Mastra Editor). The result was a fully broken Editor listing in production. Worked around with:

```sql
UPDATE mastra_agent_versions
SET model = jsonb_build_object('slug', model #>> '{}')
WHERE jsonb_typeof(model) = 'string';
```

Runtime kept working only because we have a custom override layer that tolerates scalar strings; the Editor was the only consumer that crashed.

## Suggested fixes

1. **`parseJson`**: when the storage column type is known to be `jsonb` (which is the case for `model`, `tools`, `defaultOptions`, etc.), skip `JSON.parse` and return the driver-provided value as-is. Reserve `JSON.parse` for `text`-typed columns.
2. **Listing endpoints**: don't fail-fast. If a row fails to parse, log + skip + return the rest. A single bad row shouldn't take down the whole UI.
3. **`createVersion` (write path)**: normalize string inputs to a consistent shape (e.g. `{ slug: input.model }`) or reject with a typed error. Symmetric write/read.

Happy to send a PR if there's interest in any of the three.

## Affected callers (in our codebase)

```
node_modules/@mastra/pg/dist/index.js:4237   // mapAgentRow → parseJson(row.model)
node_modules/@mastra/pg/dist/index.js:3995   // createVersion → JSON.stringify(input.model)
node_modules/@mastra/pg/dist/index.js:3690   // parseJson definition
```

tarun-khatri:
## Summary

Fixes #16224.

`@mastra/pg`'s agent listing currently crashes with HTTP 500 when any row in `mastra_agent_versions` has a jsonb column stored as a scalar string (e.g. `model = "google/gemini-3-flash"`). One bad row hides every other agent from the Editor.

The bug compounds across three sites in [`stores/pg/src/storage/domains/agents/index.ts`](https://github.com/mastra-ai/mastra/blob/main/stores/pg/src/storage/domains/agents/index.ts):

1. **Read path — `parseJson`.** The `pg` driver auto-deserialises jsonb columns, so a jsonb scalar string arrives as a JS string. `parseJson` then tries `JSON.parse("google/gemini-3-flash")`, which throws on the unquoted `g`.
2. **Listing fail-fast.** `list()` and `listVersions()` map every row through `parseRow` / `parseVersionRow`. A single throwing row takes down the whole listing.
3. **Asymmetric write path — `createVersion`.** `JSON.stringify(input.model)` silently accepts a string and stores it as a jsonb scalar — exactly the shape the read path can't handle.

### Read path — `parseJson` returns the value as-is on parse failure
Every caller in this file is reading a jsonb column. The only legitimate failure mode of `JSON.parse` here is a scalar the driver has already materialised. Returning `value` unchanged matches what the driver gives us for non-string scalars and stops the crash.

### Listing — `list` and `listVersions` skip + warn instead of failing fast
Per-row try/catch with `this.logger?.warn?.(...)`. If a future regression introduces a different mapping failure, one bad row never takes down the whole listing again.

### Write path — `createVersion` rejects a string `model` early
A typed `INVALID_MODEL` `MastraError` (category `USER`) so bad data can't reach jsonb storage. Existing scalar-string rows remain readable thanks to the read-path change above.

## Test plan

Added a new `describe('AgentsPG jsonb scalar string resilience (#16224)')` block in [`stores/pg/src/storage/test-utils.ts`](https://github.com/mastra-ai/mastra/blob/main/stores/pg/src/storage/test-utils.ts) (inside `pgTests()`). Three tests:

- [x] `listVersions` returns both a valid row and a bad-jsonb-scalar-model row, with the scalar deserialised — listing no longer crashes.
- [x] `getVersion` returns a jsonb-scalar `model` as the deserialised scalar instead of throwing.
- [x] `createVersion` rejects a string `model` with `INVALID_MODEL` and writes nothing.

Each test seeds the bad row via raw SQL (`'"google/gemini-3-flash"'::jsonb`) so it reproduces the exact production shape from the issue.

To run locally:

```bash
pnpm --filter @mastra/pg test -t "jsonb scalar string"
```

(`pretest` brings up Postgres via docker-compose automatically.)

## Out of scope

The same `parseJson` pattern exists in 11 other PG domains (`mcp-servers`, `skills`, `schedules`, `prompt-blocks`, `scorer-definitions`, `experiments`, `datasets`, `workspaces`, `mcp-clients`, `blobs`, `channels`). They are vulnerable to the same bug class but were not reported as broken in production. Happy to extend in a follow-up PR if maintainers want the broader sweep — kept this PR scoped to the issue label `effort:medium` and the reporter's recommended fix shape.

## ELI5

If a database row stored a model as plain text instead of an object, agent listing could crash. This PR makes reads tolerant (skip or return malformed-but-materialized scalar values) so listings keep working and don't fail on a single bad row.

## Summary

Fixes HTTP 500 crashes caused when Postgres jsonb columns are materialized as JSON scalar strings (e.g. "google/gemini-3-flash") by making JSON parsing and listing resilient across several PG domains.

### Root causes
1. parseJson unconditionally called JSON.parse on driver-deserialized jsonb scalar strings and threw.
2. list()/listVersions() mapped rows synchronously so a single throwing row aborted the whole listing.
3. (Initially) createVersion accepted string models and could write jsonb scalar strings — later refactor removed the write-side guard to prioritize read-side resilience.

### Changes
- New utility: parseJsonResilient(value, _fieldName?) that:
  - returns undefined for null/undefined,
  - returns non-strings unchanged,
  - attempts JSON.parse for strings and falls back to the original value on parse failure (avoids throwing on driver-provided scalar strings).
- Read resilience:
  - Replace local parseJson with parseJsonResilient in agents plus six other PG domains (mcp-clients, mcp-servers, prompt-blocks, scorer-definitions, skills, workspaces).
  - list() and listVersions() now map rows with per-row try/catch (using flatMap), log a warning via this.logger?.warn?.(...), and skip problematic rows rather than failing the entire listing.
- Write path:
  - A prior commit added rejection of string models on write, but a later refactor removed that write-side guard; this PR focuses on read-side resilience so malformed scalar rows can be read/restored without blocking restores or reverts.
- Tests:
  - Adds PG regression tests that seed pathological jsonb scalar-string rows via raw SQL ('"google/gemini-3-flash"'::jsonb) to verify:
    - listVersions returns the valid rows and does not crash when a scalar-model row is present.
    - getVersion returns the deserialized scalar value for a scalar-model row.
  - Follow-up domain tests ensure listVersions in other updated domains return deserialized scalar values for targeted jsonb columns.

### Impact
- Listing endpoints no longer crash when encountering jsonb scalar-string rows: valid rows are returned and malformed-but-materialized scalar values are returned where applicable.
- Prior attempted write-side guards were removed in favor of read-side resilience to avoid interfering with restore/revert flows.
- Similar resilient parsing and per-row error handling applied across seven PG domains.

### Out of scope
- Other PG domains still using the old parseJson pattern (beyond the seven updated domains) were not changed in this PR.

[Review Change Stack](https://app.coderabbit.ai/change-stack/mastra-ai/mastra/pull/16233)


tarun-khatri:
Thanks @coderabbitai — addressed the changeset nit in bee6db98af: rewrote it as a short, outcome-focused note per the `.mastracode/commands/changeset.md` style guide (lead with user-facing outcome, drop internal mechanics).

abhiaiyer91 (approved):

PR #16236, opened 2026-05-05
fix(core): pass writer to durable output processors

### LINKED ISSUE #16221: Custom chat route with durable agent and ai-sdk
This issue was created from Discord post 1501234448664887469:

[Open in Discord](https://discord.com/channels/1309558646228779139/1501234448664887469)

Hi team! I have a working custom chat route that responds a steam that my client ingests with Vercel's ai-sdk. Now, I'm trying to implement durable agent so we'll have resumable streams but I'm noticing a change in behaviour and I'm not sure if I'm missing anything or if it's a valid issue.

- `data-om-status` is no longer saved in storage but is streamed to the client
- the last assistant message is not streamed to the client. This includes the `text`, `data-om-thread-update`, and `data-om-buffering-end` types -- needs a reload to fetch historical messages

Here is how the custom route is defined:

```
registerApiRoute('/chat/:agentId', {
method: 'POST',
handler: async c => {
const { messages, memory } = await c.req.json();
const mastra = c.get('mastra');
const agent = mastra.getAgentById(c.req.param('agentId'));
const durableAgent = new DurableAgent({ agent });

const {
cleanup,
runId,
output
} = await durableAgent.stream(messages, {
memory
});

void output.consumeStream().catch(error => {
mastra
.getLogger()
?.error('...', { error });
});

const uiStream = createUIMessageStream({
originalMessages: messages,
execute: async ({ writer }) => {
const reader = toAISdkStream(output, {
from: 'agent',
sendReasoning: true,
version: 'v6'
}).getReader();
let result = await reader.read();
while (!result.done) {
writer.write(result.value as InferUIMessageChunk<UIMessage>);
result = await reader.read();
}
reader.releaseLock();
}
// onFinish: () => cleanup() // TODO: review, ends the stream prematurely and causes errors
});

return createUIMessageStreamResponse({ stream: uiStream });
}
});
```

Some notes for above:
- this doesn't implement resuming any streams yet
- we're instantiating the durable agent within the route just to avoid strict Typescript lint errors
- still figuring out where to call `cleanup()` without affecting the stream

Additional context on the agent:
- OM is enabled
- with output processor
- save per step is enabled

KrishavRajSingh:
## Description

Fixes DurableAgent output processors so they can write custom stream chunks and update the final response text correctly.
Durable input processors already received a stream writer, but output processors did not. This meant `writer.custom(...)` calls from `processOutputResult` never reached the client stream.

## Related Issue(s)

Fixes #16221 

## Type of Change

- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [ ] Test update

## Checklist

## ELI5

Durable agents couldn't send small custom streamed messages or the final processed assistant text from output processors to clients; this change gives output processors a stream writer and wires their emitted chunks back into the durable agent stream so those custom chunks and the final processed message reach the client.

## Changes Overview

Fixes issue #16221 by passing a stream writer into DurableAgent output processors and re-emitting processor-emitted chunks into the durable PubSub stream so OM-related chunks, processor custom chunks, and the final assistant message are delivered through the client stream while preserving resumable stream semantics. Marked as a breaking change.

### Key Changes

- packages/core/src/agent/durable/workflows/create-durable-agentic-workflow.ts
  - Import emitChunkEvent and emitFinishEvent.
  - Make finalText mutable and update it from processed output when processors modify the last assistant message.
  - Dynamically import and run ProcessorRunner at workflow finish, constructing a MessageList from persisted state.
  - Provide an outputProcessorWriter that forwards processor.custom(...) events to PubSub via emitChunkEvent.
  - Serialize updated MessageList back to state and derive processed final text; on processor failure, log a warning and continue.
  - Ensure finish events use the possibly updated final output.

- packages/core/src/agent/durable/__tests__/durable-agent-stream.test.ts
  - Add OutputProcessor import and a new integration test verifying an OutputProcessor receives a writer, can emit a data-durable-output event before finish, mutate the last assistant text to "processed response", and that the durable stream's finish payload and output (text/fullOutput/steps) reflect the processed text. Confirms event ordering (processor chunk before finish).

- packages/core/src/stream/base/output.ts
  - In the non-processor / non-LLM-execution finish path, prefer chunk.payload.output.text (when present) to resolve final text; reconcile buffered steps/text with that final text and update last step text and buffered text accordingly.

- .changeset/kind-donkeys-sell.md
  - Add changeset (patch) noting the fix and marking the change as breaking.

## Tests & Docs

- Tests: Added an integration test covering output-processor streaming and final-text processing.
- Changeset/Docs: Changeset added; docs/notes updated to reflect the breaking change.

## Impact

Restores expected streaming behavior for DurableAgent-based routes: output processors can emit custom stream chunks to clients and update the final assistant message that is streamed to clients (fixes the resumable-stream regressions). This is a breaking change in output processing/stream emission semantics.

### DIFF
--- a/.changeset/kind-donkeys-sell.md
+++ b/.changeset/kind-donkeys-sell.md
@@ -1,0 +1,5 @@
+---
+'@mastra/core': patch
+---
+
+Fixed DurableAgent output processors so custom chunks stream and processed final text is used.
--- a/packages/core/src/agent/durable/__tests__/durable-agent-stream.test.ts
+++ b/packages/core/src/agent/durable/__tests__/durable-agent-stream.test.ts
@@ -10,6 +10,7 @@ import { MockLanguageModelV2, convertArrayToReadableStream } from '@internal/ai-
 import { describe, it, expect, beforeEach, afterEach } from 'vitest';
 import { z } from 'zod';
 import { EventEmitterPubSub } from '../../../events/event-emitter';
+import type { OutputProcessor } from '../../../processors';
 import { createTool } from '../../../tools';
 import { Agent } from '../../agent';
 import { AGENT_STREAM_TOPIC, AgentStreamEventTypes } from '../constants';
@@ -228,6 +229,55 @@ describe('DurableAgent streaming execution', () => {
       // Cleanup should remove from registry
       cleanup();
       expect(durableAgent.runRegistry.has(runId)).toBe(false);
+    });
+
+    it('should stream custom chunks emitted by output processors before finish', async () => {
+      let writerWasProvided = false;
+      const outputProcessor: OutputProcessor = {
+        id: 'durable-output-writer',
+        name: 'Durable Output Writer',
+        processOutputResult: async ({ messages, writer }) => {
+          writerWasProvided = !!writer;
+          const lastAssistantMessage = [...messages].reverse().find(message => message.role === 'assistant');
+          const textPart = lastAssistantMessage?.content.parts.find(part => part.type === 'text');
+
+          if (textPart && 'text' in textPart) {
+            textPart.text = 'processed response';
+          }
+
+          await writer?.custom({
+            type: 'data-durable-output',
+            data: { status: 'processed' },
+          });
+
+          return messages;
+        },
+      };
+
+      const baseAgent = new Agent({
+        id: 'output-writer-agent',
+        name: 'Output Writer Agent',
+        instructions: 'Test durable output writer',
+        model: createTextStreamModel('draft response') as LanguageModelV2,
+        outputProcessors: [outputProcessor],
+      });
+
+      const durableAgent = createDurableAgent({ agent: baseAgent, pubsub });
+      const { output, cleanup } = await durableAgent.stream('Test');
+      const chunks: any[] = [];
+
+      for await (const chunk of output.fullStream) {
+        chunks.push(chunk);
+      }
+
+      cleanup();
+
+      const finishChunk = chunks.find(chunk => chunk.type === 'finish');
+
+      expect(writerWasProvided).toBe(true);
+      expect(chunks.some(chunk => chunk.type === 'data-durable-output')).toBe(true);
+      expect(finishChunk?.payload.output.text).toBe('processed response');
+      await expect(output.text).resolves.toBe('processed response');
     });
   });
 
--- a/packages/core/src/agent/durable/__tests__/durable-agent-stream.test.ts
+++ b/packages/core/src/agent/durable/__tests__/durable-agent-stream.test.ts
@@ -272,12 +272,24 @@ describe('DurableAgent streaming execution', () => {
 
       cleanup();
 
+      const dataIndex = chunks.findIndex(chunk => chunk.type === 'data-durable-output');
+      const finishIndex = chunks.findIndex(chunk => chunk.type === 'finish');
+      const dataChunk = chunks[dataIndex];
       const finishChunk = chunks.find(chunk => chunk.type === 'finish');
 
       expect(writerWasProvided).toBe(true);
-      expect(chunks.some(chunk => chunk.type === 'data-durable-output')).toBe(true);
+      expect(dataChunk).toBeDefined();
+      expect(finishChunk).toBeDefined();
+      expect(dataIndex).toBeLessThan(finishIndex);
       expect(finishChunk?.payload.output.text).toBe('processed response');
       await expect(output.text).resolves.toBe('processed response');
+
+      const steps = await output.steps;
+      const fullOutput = await output.getFullOutput();
+
+      expect(steps.at(-1)?.text).toBe('processed response');
+      expect(fullOutput.text).toBe('processed response');
+      expect(fullOutput.steps.at(-1)?.text).toBe('processed response');
     });
   });
 
--- a/packages/core/src/agent/durable/__tests__/durable-agent-stream.test.ts
+++ b/packages/core/src/agent/durable/__tests__/durable-agent-stream.test.ts
@@ -279,6 +279,7 @@ describe('DurableAgent streaming execution', () => {
 
       expect(writerWasProvided).toBe(true);
       expect(dataChunk).toBeDefined();
+      expect(dataChunk.data).toEqual({ status: 'processed' });
       expect(finishChunk).toBeDefined();
       expect(dataIndex).toBeLessThan(finishIndex);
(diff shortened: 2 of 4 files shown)


abhiaiyer91 (changes requested):
**Verdict: request changes**

Thanks for the fix — the durable side of the change (giving output processors a `writer` and forwarding their `custom()` chunks into the run's PubSub topic before `finish`) is the right shape and mirrors how the non-durable path in `MastraModelOutput` already builds an `outputResultWriter`. The blockers are the `output.ts` change and the fact that the branch is now stale against `main` in exactly the regions it edits.

## Findings

**1. The branch is based on pre-rewrite `main` and conflicts in the code it changes (blocking).**
GitHub reports `mergeable: CONFLICTING`. This is not a trivial textual conflict: the finish-path branch this PR rewrites was reworked on `main` by "Fix tool-call narration leaking into stream.text" (#18644) and by the durable `resolveFinalPromises` work. On `main` the branch now reads:

```ts
} else if (!self.#options.isLLMExecutionStep || self.#options.resolveFinalPromises) {
  const lastStep = self.#bufferedSteps[self.#bufferedSteps.length - 1];
  const hasToolStep = self.#bufferedSteps.some(s => s.toolCalls.length > 0 || s.toolResults.length > 0);
  this.resolvePromises({
    text: hasToolStep && !self.#wasSuspended && lastStep ? lastStep.text : self.#bufferedText.join(''),
    finishReason: self.#finishReason,
  });
}
```

The PR replaces the pre-#18644 version (`text: self.#bufferedText.join('')`). Resolving that conflict by taking this PR's side would revert both the narration-exclusion fix and the suspended/resumed aggregate-text behavior. Please rebase on `main` and reconcile explicitly with `hasToolStep` / `#wasSuspended`, then re-run the durable and narration/suspend suites.

**2. The `output.ts` change is not scoped to durable runs (blocking).**
After the rebase, that branch is shared by *every* agent stream that reaches finish without a batch processor runner — not just durable. Two consequences:

- `chunk.payload.output.text` is populated by the regular loop too (`llmIterationOutputSchema.output.text`, plumbed through the `finish` chunk in `loop/workflows/stream.ts` via `...executionResult.result`). So `finishOutput.text ?? ...` changes non-durable final-text resolution as a side effect, overriding the `hasToolStep`/`#wasSuspended` decision above.
- The change mutates shared state rather than just resolving a promise: `self.#bufferedSteps = finishOutput.steps`, `lastStep.text = finalText`, `self.#bufferedText = [finalText]`. Those buffers feed `steps`, `getFullOutput()`, and the `onFinish` payload, so the blast radius is much wider than `await output.text`.

Please gate this on the durable path (e.g. `self.#options.resolveFinalPromises`) or otherwise make it explicit, and avoid clobbering `#bufferedSteps`/`#bufferedText` when `finishOutput.text` is absent.

**3. `chunk.payload.output.text` is read through a cast, not the contract.**
`FinishPayload['output']` (`packages/core/src/stream/types.ts`) declares only `usage` and `steps?`. The PR reads `text` via `as { steps?: ...; text?: string }`. If the durable finish event is meant to carry processed text, add `text?: string` to `FinishPayload['output']` so the producer (`map-final-output`) and the consumer agree in the type system.

**4. Final-text derivation in the durable workflow can blank valid text.**
`processedText` is the join of the last assistant message's text parts, and it is applied whenever it is `!== undefined`. If the final assistant message has no text parts (tool-call-only final step, tripwire/abort), `processedText` is `''` and `finalText` becomes empty where it previously fell back to `lastStep.text`. The non-durable path only overwrites when the processor actually changed the text (`outputText && outputText !== originalText`, falling back to `originalText`); please mirror that.

**5. `(params as any).tracingContext` is a step backwards.**
`main` already destructures `tracingContext` from `params` in this `.map()` (added so `processor_run` spans parent to `AGENT_RUN`). Use the destructured value after rebasing and drop the cast.

**6. Changeset vs. breaking-change claim.**
The changeset is `patch` while the PR description marks this a breaking change. Pick one: if final-text/stream-emission semantics genuinely change for existing output processors, the changeset should say so explicitly (and be sized accordingly); if it is a bug fix, drop the breaking-change checkbox.

**Tests.** The new durable integration test is genuine — it asserts writer presence, chunk payload, `data-*`-before-`finish` ordering, and that `text` / `steps` / `getFullOutput()` all agree on the processed text. That is the right test for the durable path. What is missing is coverage for the shared `output.ts` behavior this PR alters: a processor that clears text, a tool-only final step, and confirmation that the existing narration (#18644) and suspend/resume text expectations still hold post-rebase.

**CI.** No core unit-test check is currently reported on this head (only pending E2E/Memory/store suites); the two Vercel failures are deploy-authorization only and are not a signal about this change.

## Requested changes

1. Rebase onto current `main` and reconcile the `output.ts` finish branch with the `hasToolStep` / `#wasSuspended` / `resolveFinalPromises` logic instead of replacing it.
2. Scope the `output.text` handling to the durable path and stop unconditionally overwriting `#bufferedSteps`, `#bufferedText`, and `lastStep.text`.
3. Add `text?: string` to `FinishPayload['output']` (or otherwise remove the cast) so the durable finish contract is typed.
4. Only adopt `processedText` when output processors actually modified the last assistant message; fall back to the pre-processor text otherwise.
5. Use the destructured `tracingContext` from `params` rather than `(params as any).tracingContext`.
6. Align the changeset with the breaking-change claim (or drop the claim).
7. Add tests for the cleared-text and tool-only-final-step cases, and confirm the existing narration/suspend text tests pass after the rebase.

## Assumptions

- Treated the merge conflict as blocking rather than mechanical, because the conflicting hunk is the exact logic this PR rewrites and the semantics diverged after the base commit.
- Assumed `chunk.payload.output.text` is present for non-durable runs based on `llmIterationOutputSchema` and the `finish` chunk spread in `loop/workflows/stream.ts`; if the loop strips `text` before finish, finding 2's first bullet is weaker but the shared-state mutation still stands.
- Treated the empty-`processedText` case as a real regression path rather than theoretical, since durable runs can end on a tool-only or tripwire step.
- Did not treat the pending E2E/Memory suites or the Vercel authorization failures as findings.

## Open questions

- Should the durable finish event be the authoritative source of final text (requiring a `FinishPayload` contract change), or should the durable adapter apply processed text locally and leave `output.ts` untouched? That is an architecture call worth a maintainer's input before the rebase.

abhiaiyer91:
Closing, will have factory resolve

PR #16259, opened 2026-05-06
fix(core): set threadExists from post-step state, not pre-step lookup (#16216)

### LINKED ISSUE #16216: saveThread upsert replaces thread metadata instead of merging
## Bug

`PostgresStore.saveThread` uses `ON CONFLICT (id) DO UPDATE SET metadata = EXCLUDED.metadata` — a full replacement. `#executeOnFinish` calls `memory.createThread({ metadata: thread.metadata })` where `thread.metadata` is the stale in-memory object (often `undefined`), which overwrites any metadata written during the run by processors via `updateThread`.

`updateThread` already merges metadata correctly. `saveThread` should do the same.

## Why `#executeOnFinish` triggers a second save

In `prepare-memory-step.ts`, when a thread is new:

1. `existingThread` is `null`
2. `memory.createThread({ saveThread: true })` creates the thread in DB ([line 155](https://github.com/mastra-ai/mastra/blob/main/packages/core/src/agent/workflows/prepare-stream/prepare-memory-step.ts#L155))
3. But `threadExists` is set to `!!existingThread` — still `false` ([line 194](https://github.com/mastra-ai/mastra/blob/main/packages/core/src/agent/workflows/prepare-stream/prepare-memory-step.ts#L194))

This `false` propagates to `#executeOnFinish` via `map-results-step.ts` ([line 302](https://github.com/mastra-ai/mastra/blob/main/packages/core/src/agent/workflows/prepare-stream/map-results-step.ts#L302)), which calls `memory.createThread()` again ([line 5378](https://github.com/mastra-ai/mastra/blob/main/packages/core/src/agent/agent.ts#L5378)) — triggering the upsert that replaces metadata.

## Reproduction

```typescript
const memory = new Memory({ storage: new PostgresStore({ connectionString }) });

// 1. Create thread (simulating prepare-memory-step.ts creating the thread)
await memory.saveThread({ thread: { id: "t1", resourceId: "r1", title: "", metadata: undefined, createdAt: new Date(), updatedAt: new Date() } });

// 2. Write metadata via updateThread (simulating a processor writing metadata mid-run)
const store = await memory.storage.getStore("memory");
await store.updateThread({ id: "t1", title: "", metadata: { myKey: "myValue" } });

// 3. saveThread again with undefined metadata (simulating #executeOnFinish with threadExists=false)
await memory.saveThread({ thread: { id: "t1", resourceId: "r1", title: "", metadata: undefined, createdAt: new Date(), updatedAt: new Date() } });

// 4. Check — myKey is gone
const thread = await store.getThreadById({ threadId: "t1" });
console.log(thread.metadata); // null — expected: { myKey: "myValue" }
```

Also reproduces with LibSQLStore — not PostgresStore-specific.

## Environment

- `@mastra/core`: 1.28.0
- `@mastra/memory`: 1.17.1
- `@mastra/pg`: 1.9.0

Related: #4167

tarun-khatri:
## Summary

Fixes #16216.

When an agent run creates a brand new thread, any metadata a processor writes to that thread mid-run via `updateThread` is silently wiped at end-of-run. The thread itself survives; only the metadata is lost. Reproduces on `@mastra/pg`, `@mastra/libsql`, and the in-process `InMemoryStore`.

## Root cause

`prepare-memory-step` calls `memory.createThread({ saveThread: true })` to persist the thread immediately, then returns:

```ts
threadExists: !!existingThread,
```

`existingThread` came from a lookup *before* the create — so for new threads it is `null`, and the step returns `threadExists: false` even though the thread is now persisted in the DB.

That stale `false` propagates through `map-results-step` to `agent.ts#executeOnFinish`, which guards a redundant create on `if (!threadExists)`:

```ts
if (!threadExists) {
  await memory.createThread({
    threadId: thread.id,
    metadata: thread.metadata,   // often stale or undefined
    title: thread.title,
    memoryConfig,
    resourceId: thread.resourceId,
  });
}
```

The storage backend's upsert (`ON CONFLICT (id) DO UPDATE SET metadata = EXCLUDED.metadata` in [`@mastra/pg`](https://github.com/mastra-ai/mastra/blob/main/stores/pg/src/storage/domains/memory/index.ts), equivalent in `libsql`, the in-process `InMemoryStore`, etc.) wipes any metadata a processor wrote via `updateThread` mid-run.

## Fix

Derive `threadExists` from the post-step persisted state (`!!threadObject`) instead of the pre-step lookup. The redundant end-of-run create is then correctly skipped for newly created threads.

The same bug pattern exists in [`AgentLegacyHandler`](https://github.com/mastra-ai/mastra/blob/main/packages/core/src/agent/agent-legacy.ts) (legacy v4 generate path) at two return sites — fixed identically there.

The durable execution path ([`agent/durable/preparation.ts`](https://github.com/mastra-ai/mastra/blob/main/packages/core/src/agent/durable/preparation.ts)) was already correct (sets `threadExists = true` unconditionally after the create/lookup) — left alone.

## Test plan

Added [`prepare-memory-step.test.ts`](https://github.com/mastra-ai/mastra/pull/16263/files) with three regression tests against `MockMemory` + `InMemoryStore`:

- [x] `threadExists` is `true` after creating a brand new thread
- [x] `threadExists` is `true` when the thread already existed
- [x] **Metadata-survival end-to-end**: simulate the exact scenario from the issue body — `prepare-memory-step` creates a new thread, a processor writes metadata via `updateThread`, then the `if (!threadExists)` guard from `executeOnFinish` runs. With the fix, the second create is skipped and the metadata is preserved.

**Verified:**
- Pre-fix: 2 of 3 tests fail. Test 1 fails with `expected false to be true` (the flag bug). Test 3 fails with `expected undefined to deeply equal { writtenByProcessor: 'value1' }` (the actual data loss reproduces in-process).
- Post-fix: all 3 pass.
- Adjacent suites green: [agent.test.ts](https://github.com/mastra-ai/mastra/blob/main/packages/core/src/agent/agent.test.ts) (160 passed, 5 skipped), title-generation, base64-images-with-threads, durable/tool-call-bg, processors/runner — 269 tests across 6 files, no regressions.

```bash
pnpm --filter @mastra/core test src/agent/workflows/prepare-stream/prepare-memory-step.test.ts
```

## Out of scope

The reporter also suggests making `saveThread` itself merge metadata (so it behaves like `updateThread`). That's a behavioural contract change touching every storage adapter (PG, LibSQL, MongoDB, MSSQL, Redis, ClickHouse, DynamoDB, Cloudflare D1, Convex, Lance — 10+ adapters). I've left it out of this PR. Fixing the root cause here closes the data-loss path that issue #16216 is actually filed about; happy to follow up with the broader `saveThread` semantics change as a separate PR if maintainers want it.

## ELI5

The agent was creating a new conversation thread but then forgot it had already created it and tried to create it again at the end, which wiped out any updates done in the middle. This change makes the agent recognize the thread after creation so it won't recreate and erase those updates.

## Summary

Fixes a silent metadata-loss bug where metadata written to a newly created thread during an agent run (via updateThread) could be overwritten at end-of-run.

Root cause: prepare-memory-step returned threadExists based on a pre-step lookup (`existingThread`) instead of the post-step persisted thread state. For newly created threads that lookup was false, so executeOnFinish redundantly called memory.createThread again with stale/undefined metadata, allowing storage adapters’ upserts to replace processor-written metadata.

Fix: derive threadExists from the post-step persisted state (`!!threadObject`) so newly created threads are marked existing and the redundant end-of-run create is skipped. The same change was applied in AgentLegacyHandler (legacy v4 generate path). Durable execution path already handled this correctly and was left unchanged.

Changes:
- packages/core/src/agent/workflows/prepare-stream/prepare-memory-step.ts: return threadExists: !!threadObject (with explanatory comment).
- packages/core/src/agent/agent-legacy.ts: same fix in two legacy return paths.
- packages/core/src/agent/workflows/prepare-stream/prepare-memory-step.test.ts: added three regression tests (MockMemory + InMemoryStore) verifying threadExists behavior and end-to-end metadata survival.
- .changeset/prepare-memory-thread-exists-flag.md: changeset documenting the patch.

Verification: Before the fix 2/3 regression tests failed; after the fix all 3 pass. No public API changes. Changing saveThread semantics (merge vs replace) is intentionally out of scope.


tarun-khatri:
Thanks @coderabbitai — addressed the nit in bde7ba6118: added an explicit `expect(stepResult.threadExists).toBe(true)` to the metadata-survival test before the simulated guard, so a future regression localizes to the flag itself rather than the downstream metadata effect.

PR #16268, opened 2026-05-06
fix(nestjs): forward raw query values without pre-schema coercion (#16114)

### LINKED ISSUE #16114: [BUG] NestJS adapter coerces query values before schema validation unlike other adapters
## Describe the bug

The NestJS adapter currently performs controller-level, schema-agnostic query value coercion before route schema validation.

This differs from the other server adapters. Express/Fastify/Hono/Koa normalize the framework query object and then rely on the shared route query parsing path. That shared path can JSON-parse values when the route `queryParamSchema` expects complex query fields such as objects, arrays, or records, but it does not blindly coerce every JSON-looking or scalar-looking string before schema validation.

In NestJS, query values are coerced before the route schema gets to decide the intended type. This means a route that explicitly declares a string query field can fail validation because NestJS converts the raw string before the schema can validate it. For routes without query schema validation, handlers may receive coerced values instead of raw query strings.

PR #16113 addresses one part of this by preserving numeric-looking query identifiers as strings, but NestJS still currently coerces:

- `"true"` -> `true`
- `"false"` -> `false`
- `"null"` -> `null`
- `'{"a":1}'` -> `{ a: 1 }`
- `"[1,2]"` -> `[1, 2]`

## Reproduction

Define a route with a string query parameter:

```ts
queryParamSchema: z.object({
  filter: z.string().optional(),
})
```

Send:

```http
GET /api/example?filter={"a":1}
```

## Expected behavior

For parity with Hono/Express/Fastify/Koa, the schema should receive and accept:

```ts
filter === '{"a":1}'
```

Because the route schema says `filter` is a string.

Schema-guided JSON parsing should only happen when the route schema expects a complex query field, for example `z.object(...)`, `z.array(...)`, or `z.record(...)`.

## Actual behavior

With the NestJS adapter, `filter` is JSON-parsed to `{ a: 1 }` before schema validation. Since the schema declares `filter` as `z.string()`, validation fails instead of accepting the raw query string.

For routes without a `queryParamSchema`, the handler may receive the coerced value directly.

## Cross-adapter behavior

This table is about **adapter pre-schema coercion**, not whether the shared route schema parser may later parse a value because the schema expects a complex query field.

| Request query | Intended schema example | Hono/Express/Fastify/Koa adapter pre-schema behavior | NestJS current behavior | Expected NestJS behavior |
|---|---|---|---|---|
| `?flag=true` | `z.string().optional()` or schema decides | `"true"` | `true` before validation | `"true"` |
| `?flag=false` | `z.string().optional()` or schema decides | `"false"` | `false` before validation | `"false"` |
| `?value=null` | `z.string().optional()` or schema decides | `"null"` | `null` before validation | `"null"` |
| `?filter={"a":1}` | `z.string().optional()` | `'{"a":1}'` | `{ a: 1 }` before validation, causing `z.string()` to fail | `'{"a":1}'` |
| `?items=[1,2]` | `z.string().optional()` or schema decides | `"[1,2]"` | `[1, 2]` before validation | `"[1,2]"` |
| `?id=123` | `z.string().optional()` | `"123"` | previously `123`; addressed by #16113 | `"123"` |

## Why this matters

Query strings are strings at the HTTP boundary. Mastra route schemas should control whether and how those strings are parsed or coerced.

Blind adapter-level coercion causes surprising validation behavior and breaks adapter parity. For example, a user may intentionally declare `z.string()` because they want to receive raw JSON text, IDs, flags, or other string-like values exactly as sent.

## Recommended fix

- Remove NestJS controller-level blind query value coercion.
- Keep `normalizeQueryParams()`.
- Let shared route schema parsing decide whether to parse/coerce query values.
- Users who want coercion should opt in via schema-level tools such as:
  - `z.coerce.boolean()`
  - `z.coerce.number()`
  - JSON-specific query helpers / preprocessors for fields that are intended to be objects or arrays.

Related PR: https://github.com/mastra-ai/mastra/pull/16113

tarun-khatri:
## Summary

Fixes #16114.

`@mastra/nestjs` ran a private `coerceQueryValue` helper over every query value *before* the route's `queryParamSchema` validated them. That helper turned `'true'` / `'false'` into booleans, `'null'` into `null`, numeric strings into numbers, and JSON-shaped strings into parsed objects / arrays. A route declaring `queryParamSchema: z.object({ filter: z.string() })` would then reject a valid request like `?filter={"a":1}` because the adapter had already turned the string into `{ a: 1 }`.

The other adapters (`@mastra/hono`, `@mastra/express`, `@mastra/fastify`, `@mastra/koa`) keep query values as the raw strings the HTTP layer delivered and let the route schema decide whether to parse or coerce. This PR pins NestJS to the same contract.

## What changed

[`server-adapters/nestjs/src/controllers/mastra.controller.ts`](https://github.com/mastra-ai/mastra/blob/main/server-adapters/nestjs/src/controllers/mastra.controller.ts):

- Removed the private `coerceQueryValue` helper and the call from `parseQueryParams`.
- Kept `normalizeQueryParams` from `@mastra/server/server-adapter` (same import other adapters use), so single-element arrays still collapse to strings.
- Kept the `requestContext` skip and the prototype-pollution key strip (`__proto__` / `prototype` / `constructor`).

The public contract for routes that already opted in to type coercion via `z.coerce.boolean()`, `z.coerce.number()`, etc. is unchanged.

## Test plan

Added [`server-adapters/nestjs/src/__tests__/query-coercion.test.ts`](https://github.com/mastra-ai/mastra/pull/16264/files) with 9 unit-level assertions on `MastraController.parseQueryParams`:

- [x] `'true'` stays as `'true'`
- [x] `'false'` stays as `'false'`
- [x] `'null'` stays as `'null'`
- [x] `'{"a":1}'` (JSON object) stays as `'{"a":1}'`
- [x] `'[1,2]'` (JSON array) stays as `'[1,2]'`
- [x] numeric strings stay as strings (no auto-`Number()`)
- [x] dangerous keys (`__proto__`, `prototype`, `constructor`) still stripped
- [x] `requestContext` key still removed (handled separately)
- [x] array-of-strings shape preserved for repeated query params

**Verified**:
- Pre-fix: 6 of 9 fail with assertions like `expected { items: [ 1, 2 ] } to deeply equal { items: '[1,2]' }`. The bug reproduces in-process with no HTTP layer needed.
- Post-fix: all 9 pass.
- Full nestjs adapter suite green: **2132 / 2132 tests pass across 14 files** (integration suite, MCP transports, auth/rate limit, multipart, abort signal, server-app access, stream redaction, request context, route handler, malformed JSON, etc.).
- One existing test in [`request-context.test.ts`](https://github.com/mastra-ai/mastra/pull/16264/files) was codifying the old buggy `[1, 2]` shape for `?tag=1&tag=2`. Updated to assert the correct cross-adapter `['1', '2']` shape (and a comment explains why).

```bash
pnpm --filter @mastra/nestjs test src/__tests__/query-coercion.test.ts
pnpm --filter @mastra/nestjs test
```

## Migration

If a NestJS-backed user route was relying on pre-schema coercion, they should update the field's schema to opt in:

```ts
// Before — relied on adapter pre-coercion
queryParamSchema: z.object({ flag: z.boolean().optional() });

// After — explicit, parity with other adapters
queryParamSchema: z.object({ flag: z.coerce.boolean().optional() });
```

This matches the existing guidance for all other Mastra server adapters.

## Related

PR #16113 (merged) addressed one part of this same issue — preserving numeric-looking IDs as strings — by special-casing routes that pass an `:id` path param. This PR removes the underlying blind coercion, so that special-case is now covered by the general parity rule.

## ELI5

The NestJS adapter used to convert query text like "true", "null", "123", or '{"a":1}' into booleans, null, numbers, or objects before route validation, which broke routes expecting raw strings. This PR stops that automatic conversion and forwards raw query strings/arrays to the route schema so the schema decides parsing/coercion.

---

## Problem

The NestJS server adapter performed blind, controller-level coercion of query parameters before route schema validation, causing:
- `?flag=true` → boolean true (instead of string "true")
- `?filter={"a":1}` → parsed object (instead of JSON string)
- `?id=123` → number 123 (instead of string "123")
- `?tag=1&tag=2` → [1, 2] (instead of ["1", "2"])

This broke routes declaring string types (e.g., z.string()) and diverged from other adapters that preserve raw HTTP query strings.

---

## Solution

- Removed the private coerceQueryValue helper and its use in MastraController.parseQueryParams so query values are no longer coerced at the adapter level.
- Retained normalizeQueryParams to keep framework normalization behavior (single-element arrays collapse to scalars).
- Preserved prototype-pollution key stripping (`__proto__`, `prototype`, `constructor`) and removal of `requestContext`.
- Added runtime guard in tests around parseQueryParams access to fail loudly if the private API is renamed.
- Added a smoke assertion in tests that pins the single-element-array → scalar collapse contract.

After the change, type coercion/parsing is the responsibility of route-level schemas (e.g., z.coerce.boolean(), z.coerce.number(), or a JSON preprocessor).

---

## Changes

Added:
- .changeset/nestjs-no-pre-schema-query-coercion.md — documents the fix and migration guidance
- server-adapters/nestjs/src/__tests__/query-coercion.test.ts — regression tests for raw-string preservation, dangerous-key stripping, requestContext removal, and repeated-param array behavior

Modified:
- server-adapters/nestjs/src/controllers/mastra.controller.ts — removed pre-schema coercion (-60/+8)
- server-adapters/nestjs/src/__tests__/request-context.test.ts — updated expectations for repeated params to be string arrays

---

## Testing

- New query-coercion tests: pre-fix 6/9 failing, post-fix all passing (10/10 in updated test file).
- Full NestJS adapter suite: 2132/2132 tests pass.
- Bot review (coderabbitai) added supportive automated review noting the runtime guard and smoke test.

---

## Migration

If you relied on previous adapter-level coercion, opt into coercion at the schema level:
- Booleans: z.coerce.boolean()
- Numbers: z.coerce.number()
- JSON: preprocessor or transform/refine with JSON.parse()

This aligns NestJS behavior with other adapters and preserves raw HTTP semantics.


tarun-khatri:
Thanks @coderabbitai — addressed both quick-win nits in 6bd6a0a4d5:
1. Added a runtime guard around the private `parseQueryParams` access so a future rename throws loudly instead of silently returning `undefined` for every assertion.
2. Added a smoke test pinning `normalizeQueryParams`'s single-element array → scalar collapse contract.

Full test file is now 10/10 green.

tarun-khatri:
Thanks @coderabbitai. Skipping the colocate suggestion — declining with reasoning per CONTRIBUTING.md.

The repo-wide AGENTS.md does say tests should colocate, but the established convention specifically in `server-adapters/` is centralized `src/__tests__/` directories — every adapter package follows it:

- `server-adapters/express/src/__tests__/` (9 test files)
- `server-adapters/fastify/src/__tests__/` (8 test files)
- `server-adapters/hono/src/__tests__/` (7+ test files)
- `server-adapters/nestjs/src/__tests__/` (13 test files including the existing `request-context.test.ts` and `route-handler.service.test.ts` that test the same controller surface)

Moving `query-coercion.test.ts` alone to `src/controllers/` would make it the lone outlier in this package and across the whole `server-adapters/` family. Happy to defer to maintainers if they'd rather flip the entire adapter-tests convention as a separate cleanup, but doing it just for this one new file feels worse than matching the local pattern.

The runtime guard + the single-element-array contract test (commit 6bd6a0a4d5) cover the substantive nits from the previous review.

TheIsrael1 (approved):

PR #16280, opened 2026-05-06
Fix CLI next.config lint parsing

### LINKED ISSUE #16180: CLI lint should not execute next.config.js while checking serverExternalPackages

graysonhicks:
## Description

Fixes the CLI Next.js lint rule so it statically parses `next.config.js` and reads literal `serverExternalPackages` values without executing project code. Adds regression coverage for supported literal config shapes and for dynamic code that must not run during linting.

## Related Issue(s)

Fixes #16180

## Type of Change

- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [x] Test update

## Checklist

## ELI5

This PR fixes a security issue in the Mastra CLI where it was using dangerous `eval()` code to read Next.js configuration files, which could execute any code embedded in those files. The fix replaces that with safe static code analysis (reading the file structure without running it), so configuration can still be read properly but no hidden code gets executed.

---

## Overview

The PR replaces unsafe `eval()`-based parsing with TypeScript AST (Abstract Syntax Tree) parsing in the Next.js config lint rule, eliminating the risk of executing arbitrary code from `next.config.js` during linting while preserving all lint functionality.

---

### Core Implementation (`packages/cli/src/commands/lint/rules/nextConfigRule.ts`)

Refactored the config parsing logic to use TypeScript's AST parser instead of regex extraction and `eval()`:

- **New helper functions**:
  - `readStaticStringArray()`: Safely extracts string arrays from AST nodes, returning `null` for any non-literal expressions
  - `readConfigObject()`: Extracts `serverExternalPackages` from an object literal AST node
  - `findNextConfigObject()`: Locates the config object in three supported literal patterns:
    - `const nextConfig = { ... }; module.exports = nextConfig;`
    - `module.exports = { ... };`
    - `export default { ... };`
  - `isModuleExports()`: Checks if an expression is `module.exports`
  - `getPropertyNameText()`: Safely reads property names from AST nodes

- **Behavior**: Parses `next.config.js` using `ts.createSourceFile()`, locates a literal object config, and extracts only statically-determinable string arrays for `serverExternalPackages`. Dynamic expressions (IIFE, function calls, variables) are rejected, causing the rule to fail if no valid static config is found.

### Test Coverage (`packages/cli/src/commands/lint/rules/nextConfigRule.test.ts`)

Added comprehensive regression tests:

- **Literal config support**: Validates all three supported config patterns correctly read `serverExternalPackages`
- **Security regression test**: Proves that dynamic code (a self-executing function that sets an environment variable) is **not executed** during linting, with assertion that the env variable remains undefined

Test infrastructure:
- Creates isolated temporary project directories with synthetic `next.config.js` files
- Mocks the logger module to avoid side effects
- Cleans up temp directories and environment state after each test

### Dependency Update (`packages/cli/package.json`)

Added TypeScript as a dependency (`"typescript": "catalog:"`) to enable AST parsing capabilities.

### Changeset Entry

Added `.changeset/fine-oranges-switch.md` documenting the patch fix: "Fixed CLI linting of Next.js config so it reads serverExternalPackages without running project code."

---

## Security Impact

- **Before**: Executing arbitrary code in `next.config.js` during linting
- **After**: Static analysis only; no code execution regardless of what's in the config file

---

## Testing

- 4 test cases covering all supported literal config shapes and security regression
- Dynamic code rejection validated with IIFE mutation test

### DIFF
--- a/.changeset/fine-oranges-switch.md
+++ b/.changeset/fine-oranges-switch.md
@@ -1,0 +1,5 @@
+---
+'mastra': patch
+---
+
+Fixed CLI linting of Next.js config so it reads serverExternalPackages without running project code.
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -69,6 +69,7 @@
     "serve-handler": "^6.1.7",
     "shell-quote": "^1.8.3",
     "strip-json-comments": "^5.0.3",
+    "typescript": "catalog:",
     "yocto-spinner": "^1.1.0"
   },
   "devDependencies": {
@@ -90,7 +91,6 @@
     "rollup": "^4.59.0",
     "tsup": "^8.5.1",
     "type-fest": "^5.4.4",
-    "typescript": "catalog:",
     "vitest": "catalog:"
   },
   "peerDependencies": {
--- a/packages/cli/src/commands/lint/rules/nextConfigRule.test.ts
+++ b/packages/cli/src/commands/lint/rules/nextConfigRule.test.ts
@@ -1,0 +1,95 @@
+import { mkdtempSync, rmSync, writeFileSync } from 'node:fs';
+import { tmpdir } from 'node:os';
+import { join } from 'node:path';
+import { afterEach, describe, expect, it, vi } from 'vitest';
+import { nextConfigRule } from './nextConfigRule.js';
+import type { LintContext } from './types.js';
+
+vi.mock('../../../utils/logger.js', () => ({
+  logger: {
+    error: vi.fn(),
+    info: vi.fn(),
+    warn: vi.fn(),
+  },
+}));
+
+const evalMarker = 'MASTRA_NEXT_CONFIG_RULE_EVAL_MARKER';
+
+function createContext(rootDir: string): LintContext {
+  return {
+    rootDir,
+    mastraDir: join(rootDir, 'src', 'mastra'),
+    outputDirectory: join(rootDir, '.mastra'),
+    discoveredTools: [],
+    packageJson: {},
+    mastraPackages: [{ name: '@mastra/core', version: '1.0.0', isAlpha: false }],
+  };
+}
+
+function createProject(nextConfigContent: string): string {
+  const rootDir = mkdtempSync(join(tmpdir(), 'mastra-next-config-rule-'));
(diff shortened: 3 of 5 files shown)


LekoArts:
We shouldn't use typescript for this, plus we have a PR already open for this?
https://github.com/mastra-ai/mastra/pull/16184

PR #16284, opened 2026-05-06
fix(core): recover prior tool-call args when converting stand-alone tool-results (#16017)

### LINKED ISSUE #16017: [BUG] AIV5Adapter.fromModelMessage fabricates empty args on resume tool-results, poisons LLM context after 3-4 cycles
## Summary

`AIV5Adapter.fromModelMessage` in `@mastra/core@1.28.0` (`packages/core/dist/chunk-DLUNYVZY.js` ESM / `chunk-SRAWH6R4.cjs` CJS) fabricates `args: {}` on every `tool-result` whose matching `tool-call` is not in the same model message. The canonical case is `agent.resumeStream(resumeData)`: the resumed stream emits a tool-result chunk on its own, with no preceding tool-call in that message, so the fabrication branch fires every cycle and persists `args: {}` for the call.

After 3–4 such cycles, Anthropic in-context-learns the empty pattern and starts emitting empty-args tool-calls itself. Validation fails (`Required: <field>`) on every subsequent invocation.

## Reproduction

Tested with a multi-tool slide-build agent using `agent.suspend()` / `agent.resumeStream()` (Mastra v1 native HITL pattern):

1. Tool calls `agentCtx.suspend(args)` → relay maps to a custom event → host runs effect → POSTs result to `/eko/resume` → `agent.resumeStream(resumeData)` resolves the suspend.
2. Postgres `mastra_messages` rows show: original tool-call with **full args** (call message), then the result message with **`args: {}`** for the same `toolCallId`.
3. After 3–4 successful slides, the LLM emits an empty-args tool-call as its primary call. Tool input validation fails. Subsequent slides all fail the same way.

Persisted-args breakdown for one 12-slide attempt:
- 7 toolCallIds: full args persisted on call, `args: {}` on result row (Mastra running OK)
- 5 toolCallIds: only empty args present (LLM emitted empty after pattern-matching)
- 5 of 12 final tool invocations failed validation

## Trace into the LLM input

Inspecting `mastra_ai_spans.input` confirms the corrupted `args: {}` entries reach Anthropic in subsequent turn payloads:

```
[8.1]  good-call  toolu_011H5b... keys=[slideIndex, title, brandKitId, layoutName, dsl]
[10.0] EMPTY-call toolu_011H5b... keys=[]      ← persisted from fabrication branch
[12.0] OK-result  toolu_011H5b... slideIndex=0
[18.0] EMPTY-call toolu_01LhVL... keys=[]      ← LLM has now learned the pattern
[21.0] ERR-result toolu_017jmi... msg="Tool input validation failed: Required: slideIndex, title, dsl"
```

Each empty-args call shares its `toolCallId` with a real (full-args) call earlier in the same conversation — so the data is recoverable from thread storage; it's just not consulted on the persistence path.

## Root cause

`packages/core/dist/chunk-DLUNYVZY.js` (and the CJS sibling `chunk-SRAWH6R4.cjs`), in `AIV5Adapter.fromModelMessage`, around line 1217:

```js
} else if (part.type === "tool-result") {
    const toolResultPart = part;
    const matchingCall = toolInvocations.find(
        (inv) => inv.toolCallId === toolResultPart.toolCallId
    );
    // ...
    if (matchingCall) {
        updateMatchingCallInvocationResult(toolResultPart, matchingCall);
    } else {
        const call = {
            state: "call",
            toolCallId: toolResultPart.toolCallId,
            toolName: sanitizeToolName(toolResultPart.toolName),
            args: {}    // ← FABRICATION
        };
        updateMatchingCallInvocationResult(toolResultPart, call);
        toolInvocations.push(call);
    }
```

`toolInvocations` is a per-message-batch local array — only contains tool-calls present in *the current* model message. On resume, the resumed stream emits the tool-result without re-emitting the original tool-call (which is already in DB memory), so `matchingCall` is undefined, the fabrication branch fires, and `args: {}` is what gets persisted.

The `AIV6Adapter` already accepts a `context` parameter and uses `context.dbMessages` for tool-approval matching at line 1585 (`findApprovalRequest`). The same lookup pattern can recover the prior tool-call's args before falling back to `{}`. Mastra also already has a `findToolCallArgs(dbMessages, toolCallId)` helper used on the **outbound** Anthropic-conversion path (`enrichToolResultsWithInput` at chunk-DLUNYVZY.js:282). It just isn't applied on the inbound persistence path.

## Patch

Threads `context.dbMessages` through to `AIV5Adapter.fromModelMessage` and adds a `findPriorToolCallArgs(toolCallId)` helper that searches prior persisted messages by `toolCallId` and reuses real args. Two fabrication branches updated; two callsites pass `context` through.

Patch file (5 hunks, applies cleanly to `1.28.0`):

```diff
diff --git a/node_modules/@mastra/core/dist/chunk-DLUNYVZY.js b/node_modules/@mastra/core/dist/chunk-DLUNYVZY.js
index 281875b..c944600 100644
--- a/node_modules/@mastra/core/dist/chunk-DLUNYVZY.js
+++ b/node_modules/@mastra/core/dist/chunk-DLUNYVZY.js
@@ -1164,13 +1164,46 @@ var AIV5Adapter = class {
   }
   /**
    * Direct conversion from AIV5 ModelMessage to MastraDBMessage
+   *
+   * INNOKO PATCH (ISD-667): the fabrication branches at lines ~1220 and ~1236
+   * inject `args: {}` when a tool-result arrives in a model message that does
+   * not also contain its matching tool-call (the canonical case is Mastra's
+   * own resume path: `agent.resumeStream()` re-invokes execute and emits the
+   * tool-result chunk on its own). Persisting `args: {}` poisons the LLM via
+   * in-context learning — after 3-4 cycles Anthropic pattern-completes empty
+   * args and every subsequent tool-call ships with no input, breaking the
+   * a2x slide build loop. The fix threads `context.dbMessages` through and
+   * reuses the prior tool-call's args before falling through to `{}`.
    */
-  static fromModelMessage(modelMsg, _messageSource) {
+  static fromModelMessage(modelMsg, _messageSource, context = {}) {
     const content = Array.isArray(modelMsg.content) ? modelMsg.content : [{ type: "text", text: modelMsg.content }];
     const mastraDBParts = [];
     const toolInvocations = [];
     const reasoningParts = [];
     const experimental_attachments = [];
+    // INNOKO PATCH: locate a prior tool-call's args by toolCallId across the
+    // already-persisted thread messages.
+    const findPriorToolCallArgs = (toolCallId) => {
+      const dbMessages = context && context.dbMessages;
+      if (!dbMessages || !toolCallId) return void 0;
+      for (const message of [...dbMessages].reverse()) {
+        const parts = (message && message.content && message.content.parts) || [];
+        for (const part of [...parts].reverse()) {
+          if (
+            part &&
+            part.type === "tool-invocation" &&
+            part.toolInvocation &&
+            part.toolInvocation.toolCallId === toolCallId &&
+            part.toolInvocation.args &&
+            typeof part.toolInvocation.args === "object" &&
+            Object.keys(part.toolInvocation.args).length > 0
+          ) {
+            return part.toolInvocation.args;
+          }
+        }
+      }
+      return void 0;
+    };
     for (const part of content) {
       if (part.type === "text") {
         const textPart = {
@@ -1217,11 +1250,13 @@ var AIV5Adapter = class {
         if (matchingCall) {
           updateMatchingCallInvocationResult(toolResultPart, matchingCall);
         } else {
+          // INNOKO PATCH: fall back to args from prior persisted tool-call.
+          const recoveredArgs = findPriorToolCallArgs(toolResultPart.toolCallId);
           const call = {
             state: "call",
             toolCallId: toolResultPart.toolCallId,
             toolName: sanitizeToolName(toolResultPart.toolName),
-            args: {}
+            args: recoveredArgs ?? {}
           };
           updateMatchingCallInvocationResult(toolResultPart, call);
           toolInvocations.push(call);
@@ -1233,12 +1268,14 @@ var AIV5Adapter = class {
             matchingV2Part.createdAt = getMastraCreatedAt(toolResultPart.providerOptions) ?? matchingV2Part.createdAt;
           }
         } else {
+          // INNOKO PATCH: same recovery for the parts-array branch.
+          const recoveredArgs = findPriorToolCallArgs(toolResultPart.toolCallId);
           const toolInvocationPart = {
             type: "tool-invocation",
             toolInvocation: {
               toolCallId: toolResultPart.toolCallId,
               toolName: sanitizeToolName(toolResultPart.toolName),
-              args: {},
+              args: recoveredArgs ?? {},
               state: "call"
             }
           };
@@ -1542,7 +1579,9 @@ var AIV6Adapter = class _AIV6Adapter {
         ...modelMsg,
         content: compatibleContent
       },
-      _messageSource
+      _messageSource,
+      // INNOKO PATCH (ISD-667): forward context for prior-args recovery.
+      context
     );
     const parts = [...baseDb.content.parts];
     if (modelMsg.role === "assistant") {
@@ -2341,7 +2380,8 @@ function inputToMastraDBMessage(message, messageSource, context) {
     };
   }
   if (TypeDetector.isAIV5CoreMessage(message)) {
-    const dbMsg = AIV5Adapter.fromModelMessage(message, messageSource);
+    // INNOKO PATCH (ISD-667): pass context for prior-args recovery.
+    const dbMsg = AIV5Adapter.fromModelMessage(message, messageSource, context);
     const rawCreatedAt = "metadata" in message && message.metadata && typeof message.metadata === "object" && "createdAt" in message.metadata ? message.metadata.createdAt : void 0;
     return stampMessageParts(
       {
diff --git a/node_modules/@mastra/core/dist/chunk-SRAWH6R4.cjs b/node_modules/@mastra/core/dist/chunk-SRAWH6R4.cjs
index a8a4164..b07abb5 100644
--- a/node_modules/@mastra/core/dist/chunk-SRAWH6R4.cjs
+++ b/node_modules/@mastra/core/dist/chunk-SRAWH6R4.cjs
@@ -1166,13 +1166,49 @@ var AIV5Adapter = class {
   }
   /**
    * Direct conversion from AIV5 ModelMessage to MastraDBMessage
+   *
+   * INNOKO PATCH (ISD-667): the fabrication branches at lines ~1222 and ~1238
+   * inject `args: {}` when a tool-result arrives in a model message that does
+   * not also contain its matching tool-call (the canonical case is Mastra's
+   * own resume path: `agent.resumeStream()` re-invokes execute and emits the
+   * tool-result chunk on its own). Persisting `args: {}` poisons the LLM via
+   * in-context learning — after 3-4 cycles Anthropic pattern-completes empty
+   * args and every subsequent tool-call ships with no input, breaking the
+   * a2x slide build loop ("Diapositive 1" / slide 4 collapse). The fix
+   * threads `context.dbMessages` through (the same lookup used by AIV6's
+   * approval matching at line 1585) and reuses the prior tool-call's args
+   * before falling through to `{}`.
    */
-  static fromModelMessage(modelMsg, _messageSource) {
+  static fromModelMessage(modelMsg, _messageSource, context = {}) {
     const content = Array.isArray(modelMsg.content) ? modelMsg.content : [{ type: "text", text: modelMsg.content }];
     const mastraDBParts = [];
     const toolInvocations = [];
     const reasoningParts = [];
     const experimental_attachments = [];
+    // INNOKO PATCH: locate a prior tool-call's args by toolCallId across the
+    // already-persisted thread messages. Used by the tool-result fabrication
+    // branches below to recover real args on resume instead of saving `{}`.
+    const findPriorToolCallArgs = (toolCallId) => {
+      const dbMessages = context && context.dbMessages;
+      if (!dbMessages || !toolCallId) return void 0;
+      for (const message of [...dbMessages].reverse()) {
+        const parts = (message && message.content && message.content.parts) || [];
+        for (const part of [...parts].reverse()) {
+          if (
+            part &&
+            part.type === "tool-invocation" &&
+            part.toolInvocation &&
+            part.toolInvocation.toolCallId === toolCallId &&
+            part.toolInvocation.args &&
+            typeof part.toolInvocation.args === "object" &&
+            Object.keys(part.toolInvocation.args).length > 0
+          ) {
+            return part.toolInvocation.args;
+          }
+        }
+      }
+      return void 0;
+    };
     for (const part of content) {
       if (part.type === "text") {
         const textPart = {
@@ -1219,11 +1255,15 @@ var AIV5Adapter = class {
         if (matchingCall) {
           updateMatchingCallInvocationResult(toolResultPart, matchingCall);
         } else {
+          // INNOKO PATCH: fall back to args from a prior persisted tool-call
+          // with the same toolCallId before injecting `{}`. See the patch
+          // header comment at the top of fromModelMessage for rationale.
+          const recoveredArgs = findPriorToolCallArgs(toolResultPart.toolCallId);
           const call = {
             state: "call",
             toolCallId: toolResultPart.toolCallId,
             toolName: sanitizeToolName(toolResultPart.toolName),
-            args: {}
+            args: recoveredArgs ?? {}
           };
           updateMatchingCallInvocationResult(toolResultPart, call);
           toolInvocations.push(call);
@@ -1235,12 +1275,14 @@ var AIV5Adapter = class {
             matchingV2Part.createdAt = getMastraCreatedAt(toolResultPart.providerOptions) ?? matchingV2Part.createdAt;
           }
         } else {
+          // INNOKO PATCH: same recovery as above for the parts-array branch.
+          const recoveredArgs = findPriorToolCallArgs(toolResultPart.toolCallId);
           const toolInvocationPart = {
             type: "tool-invocation",
             toolInvocation: {
               toolCallId: toolResultPart.toolCallId,
               toolName: sanitizeToolName(toolResultPart.toolName),
-              args: {},
+              args: recoveredArgs ?? {},
               state: "call"
             }
           };
@@ -1544,7 +1586,10 @@ var AIV6Adapter = class _AIV6Adapter {
         ...modelMsg,
         content: compatibleContent
       },
-      _messageSource
+      _messageSource,
+      // INNOKO PATCH (ISD-667): forward context so the AIV5 adapter can
+      // recover prior tool-call args during fabrication.
+      context
     );
     const parts = [...baseDb.content.parts];
     if (modelMsg.role === "assistant") {
@@ -2343,7 +2388,9 @@ function inputToMastraDBMessage(message, messageSource, context) {
     };
   }
   if (TypeDetector.isAIV5CoreMessage(message)) {
-    const dbMsg = AIV5Adapter.fromModelMessage(message, messageSource);
+    // INNOKO PATCH (ISD-667): pass context so the AIV5 adapter can recover
+    // prior tool-call args during fabrication.
+    const dbMsg = AIV5Adapter.fromModelMessage(message, messageSource, context);
     const rawCreatedAt = "metadata" in message && message.metadata && typeof message.metadata === "object" && "createdAt" in message.metadata ? message.metadata.createdAt : void 0;
     return stampMessageParts(
       {

```

Also covers the duplicate CJS variant `chunk-SRAWH6R4.cjs` for projects that load via `require()`.

## Environment

- `@mastra/core@1.28.0`
- `@mastra/memory@1.17.1`
- `@mastra/pg@1.9.2`
- Node 22, PostgreSQL 17
- ESM consumer (`"type": "module"`)
- Anthropic Claude Sonnet 4.5

## Workaround

Apply the patch via [`patch-package`](https://github.com/ds300/patch-package). Patch attached above (covers both ESM and CJS chunks).

## Suggested fix in upstream

The proper fix likely belongs in `MessageList` rather than `fromModelMessage` — looking at `hydrateMastraDBMessageFields` (chunk-DLUNYVZY.js:~2402) you already have a *intra-message* repair for this same shape. Extending that repair to look at prior persisted messages would be the cleaner upstream fix. Happy to open a PR if useful.

tarun-khatri:
## Summary

Fixes #16017.

`AIV5Adapter.fromModelMessage` fabricated `args: {}` on every `tool-result` whose matching `tool-call` lived in a different model message. The canonical case is `agent.resumeStream(...)` after an HITL suspend: the resumed stream emits a stand-alone tool-result chunk, the persisted DB row stores `args: {}` for that toolCallId, and after **3-4 such cycles Anthropic in-context-learns the empty pattern** and starts emitting empty-args tool-calls itself — breaking every subsequent invocation with `Required: <field>` validation errors.

The reporter's production trace shows ~5 of 12 slides in a slide-build agent failing with this exact pattern.

## Root cause

The infrastructure to fix this already existed in the repo:

- [`InputConversionContext.dbMessages`](https://github.com/mastra-ai/mastra/blob/main/packages/core/src/agent/message-list/conversion/input-converter.ts#L24) is documented as *"Messages array for looking up tool call args"*.
- [`AIV6Adapter.fromModelMessage`](https://github.com/mastra-ai/mastra/blob/main/packages/core/src/agent/message-list/adapters/AIV6Adapter.ts#L288) accepts a context with `dbMessages`.

But:
- [`AIV5Adapter.fromModelMessage`](https://github.com/mastra-ai/mastra/blob/main/packages/core/src/agent/message-list/adapters/AIV5Adapter.ts#L637) had no context parameter — the AIV5 entry point on `inputToMastraDBMessage` dropped it.
- The internal AIV6 → AIV5 forwarding at line 301 also dropped context.

So even though the right data was sitting one stack frame away, the recovery path was unreachable, and both fabrication sites at `AIV5Adapter.ts:706` and `:724` produced `args: {}`.

## Fix

- `AIV5AdapterContext` gains an optional `previousMessages` field, fully documented as the resume-recovery hook.
- `AIV5Adapter.fromModelMessage` accepts a `Pick<AIV5AdapterContext, 'previousMessages'>` third argument and uses a new private helper `findPriorToolCallArgs` to scan it newest-first for a matching `toolCallId`.
- Both `args: {}` fabrication sites now use the recovered args when available; both fall back to `{}` for backwards compatibility when no context is supplied.
- `inputToMastraDBMessage` (AIV5 branch) and `AIV6Adapter.fromModelMessage` now forward `dbMessages` as `previousMessages`.

Each new identifier (interface field, helper, updated method) carries explicit JSDoc so docstring coverage stays high on the changed surface.

## Test plan

Added [`AIV5Adapter-resume-tool-args.test.ts`](https://github.com/mastra-ai/mastra/pull/16284/files) covering 6 scenarios:

- [x] **Default `{}` fallback** — no context supplied, behaviour matches today.
- [x] **Recovery from a prior message** — the resume-flow positive case from the issue body.
- [x] **`toolInvocations` array also recovers** — the second fabrication site.
- [x] **No-match fallback** — `previousMessages` doesn't contain the id, no regression.
- [x] **Intra-message unchanged** — same-message tool-call/tool-result pair still works.
- [x] **Most-recent wins** on duplicate `toolCallId`s.

**Verified:**
- Pre-fix: **3 of 6 fail** with assertions like `expected {} to deeply equal { id: 42 }` — bug reproduces.
- Post-fix: **all 6 pass**.
- Adjacent suites stay green: full message-list directory (28 files, **378 tests**), `agent.test.ts` + `title-generation.test.ts` (**194 tests**, 5 skipped). Combined: **578 tests pass, 0 fail**.

```bash
pnpm --filter @mastra/core test src/agent/message-list/adapters/AIV5Adapter-resume-tool-args.test.ts
pnpm --filter @mastra/core test src/agent/message-list/
```

## Backwards compatibility

The new third parameter is optional. Every existing call site that doesn't pass `context.previousMessages` keeps the legacy `args: {}` behaviour.

The two internal call sites I updated (`inputToMastraDBMessage` AIV5 branch and `AIV6Adapter.fromModelMessage`) now forward the `dbMessages` array that was already being threaded through `InputConversionContext`. No new public API and no external callers need to change.

## Related

The reporter ([@jfsiroisleclerc](https://github.com/jfsiroisleclerc)) shipped a `patch-package` workaround in their issue body and suggested the upstream fix could also live in `MessageList.hydrateMastraDBMessageFields` (which already does intra-message repair). I went with the source-level fix in `fromModelMessage` because it stops the bad data at the moment it would have been written — there's no window in which `args: {}` ever exists. The `hydrateMastraDBMessageFields` post-fact path is still useful for messages that were already persisted with the bug, but that's a follow-up consideration.

## ELI5

When a resumed AI run delivers a tool's result in a new message, the system used to forget the tool's original inputs and invent empty ones. This PR makes the converter look back at earlier messages to recover the real inputs so resumed runs keep correct tool-invocation data.

## Overview

Fixes #16017 by updating the AIV5 conversion path so stand-alone tool-result parts no longer get fabricated empty args (`args: {}`) when their matching tool-call lived in a prior persisted message (a common case during agent.resumeStream after HITL suspend). Instead, the adapter can recover original tool-call args from earlier messages, preventing LLMs from learning and repeating an empty-args pattern and avoiding later validation failures.

## Root cause

AIV5's conversion previously dropped prior-message context (InputConversionContext.dbMessages), so fromModelMessage had no way to locate prior tool-call args. AIV6 already threaded this context; AIV5 now does too.

## Changes made

- packages/core/src/agent/message-list/adapters/AIV5Adapter.ts
  - Added exported interface AIV5AdapterContext with optional previousMessages.
  - fromModelMessage now accepts context?: Pick<AIV5AdapterContext,'previousMessages'>.
  - Added private findPriorToolCallArgs(...) which scans previousMessages newest-first for a matching toolCallId and extracts args.
  - Both fabrication sites that previously emitted args: {} now use recoveredArgs ?? {}.
  - Added JSDoc for new identifiers and nearby helpers.

- packages/core/src/agent/message-list/adapters/AIV6Adapter.ts
  - Forwards dbMessages → previousMessages when calling AIV5Adapter.fromModelMessage so the V6 → V5 bridge can recover args.

- packages/core/src/agent/message-list/conversion/input-converter.ts
  - inputToMastraDBMessage now passes context.dbMessages as previousMessages to AIV5Adapter.fromModelMessage for AIV5CoreMessage inputs.

- Tests
  - Added packages/core/src/agent/message-list/adapters/AIV5Adapter-resume-tool-args.test.ts with regression tests covering resume scenarios (default fallback, recovery, toolInvocations recovery, no-match fallback, intra-message unchanged, legacy/restored shape, most-recent-wins).
  - Pre-fix: several regressions; post-fix: all regression tests pass (7/7).

- Changeset
  - .changeset/aiv5adapter-recover-resume-tool-args.md documents the patch and links #16017.

## Additional reviewer notes

- findPriorToolCallArgs was extended (per review) to also inspect content.toolInvocations to recover args from legacy/restored DB rows where the call exists only on the flat toolInvocations array.
- The new previousMessages parameter is optional; behavior is backward-compatible when omitted (falls back to args: {}).
- The fix prevents bad data at conversion time; hydrateMastraDBMessageFields remains available for repairing already-persisted rows.

## Public API / Signature changes

- New exported interface: AIV5AdapterContext (adds previousMessages).
- AIV5Adapter.fromModelMessage signature now accepts an optional third parameter: context?: { previousMessages?: MastraDBMessage[] }.

## Backwards compatibility

Optional parameter; default legacy behavior preserved when context is not supplied.

## Testing & Impact

- New regression tests added and passing (7/7). Full affected suites remain green.
- No public exports removed; changes are additive and focused on conversion correctness.


tarun-khatri:
Thanks @coderabbitai. All three items addressed in 160f60298c:

1. **`findPriorToolCallArgs` toolInvocations fallback** — applied your suggested patch verbatim. The helper now also scans `content.toolInvocations` so legacy/restored DB rows (where the call only lives on the flat array) still recover args. Added a 7th regression test pinning that shape.

2. **Changeset shortened to outcome-focused** — dropped the in-context-learning / persisted-row mechanics and rewrote it as two short sentences per `.coderabbit.yaml` path-instructions for changeset files.

3. **Docstring coverage** — pushed it above the 80% threshold by documenting the two nearby utility helpers (`mergeMastraCreatedAt`, `getMastraCreatedAt`) that were previously undocumented in `AIV5Adapter.ts`.

Tests still 7/7 green.

PR #16293, opened 2026-05-06
Enforce Harness tool permissions through Tool Gate

mbenhamd:
## Summary
- route Harness permission rules through the internal Tool Gate policy instead of the old global approval-only path
- keep explicit denies stronger than yolo/session grants and propagate the policy into approve/decline/resume/subagent runs
- hide provider-executed tools that require local approval before model exposure and reject stale provider-executed calls that still require local approval
- make needsApprovalFn additive in both non-durable and durable tool approval paths

## Validation
- pnpm --filter ./packages/core test -- src/harness/harness-tool-gate-policy.test.ts src/harness/subagent-tool.test.ts src/loop/workflows/agentic-execution/tool-call-step.test.ts src/agent/__tests__/active-tools-enforcement.test.ts src/agent/durable/utils/resolve-runtime.test.ts src/agent/durable/workflows/steps/llm-execution.test.ts
- git diff --check

## Review
- Reviewed locally with Claude, Gemini, and Codex. Codex found the duplicate Harness approval path and durable additive-approval gap; both are fixed in this PR.


mbenhamd:
Closing this PR because it was opened against official main but this branch is stacked on the fork's ToolGate/Flow Signals baseline. I will keep the clean PR against the fork main instead.

PR #16302, opened 2026-05-07
fix(core): handle special tokens in TokenLimiterProcessor

### LINKED ISSUE #16301: TokenLimiterProcessor crashes on LLM-generated content containing special tokens (e.g. `<|endoftext|>`) — intentional?
## Context

We're using `TokenLimiterProcessor` as an input processor in our agent setup. Our agent has a tool that calls `gpt-4.1-nano` via `generateText` and returns the generated content as a tool result. Occasionally, the model leaks special tokens like `<|endoftext|>` into its output.

On the next agent step, when `TokenLimiterProcessor` processes the message history (which now includes the tool result), `js-tiktoken`'s `encode()` throws:

```
The text contains a special token that is not allowed: <|endoftext|>
```

This crashes the entire agent workflow with `PROCESSOR_WORKFLOW_FAILED`.

## Root cause

`TokenLimiterProcessor` calls `encoder.encode(text)` without passing `allowedSpecial`, so the tokenizer defaults to rejecting all special tokens. Since the processor is only counting tokens (not interpreting them), the rejection seems unnecessarily strict.

Relevant code paths:
- `countInputMessageTokens` — `encoder.encode(tokenString).length`
- `countCoreSystemMessageTokens` — `encoder.encode(tokenString).length`
- `countTokensInChunk` — `encoder.encode(...)` in multiple branches

## Question

Is this behavior intentional? We can think of a few perspectives:

1. **Intentional as a safety guard** — rejecting special tokens prevents prompt injection payloads from propagating through the message history. If so, is there a recommended way to handle LLM outputs that contain leaked special tokens before they re-enter the processor?

2. **Unintentional** — the processor is just counting tokens, so it should tolerate any content. A fix could be passing `allowedSpecial: "all"` to `encode()`, or wrapping in a try/catch with a fallback estimation.

We're leaning toward option 2 since the token limiter's job is truncation, not content validation — but wanted to check before working around it.

## Reproduction

1. Configure an agent with `inputProcessors: [new TokenLimiterProcessor({ limit: N })]`
2. Add a tool that calls an OpenAI model (e.g. `gpt-4.1-nano`) and returns the raw output
3. Wait for the model to leak `<|endoftext|>` (or simulate by returning it from the tool)
4. The next agent step crashes in the token limiter

## Environment

- `@mastra/core`: 1.16.0

intojhanurag:
## Description

Fixed `TokenLimiterProcessor` crashing when input or output text contains tiktoken
special-token strings (e.g. `<|endoftext|>`). The processor now passes `'all'` to
`encoder.encode`, which treats those sequences as ordinary text instead of raising.

- Added a `countTokens` helper on `TokenLimiterProcessor` that wraps `encoder.encode(text,
  'all').length`
- Replaced every direct `encoder.encode(...).length` call (input-step counting,
  output-result counting, and stream-chunk counting) with the helper
- Added tests for text-delta and tool-result chunks, output result text, system messages,
  and tool results that contain special token strings

## Related Issue(s)

Fixes #16301

## Type of Change

- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [ ] Test update

## Checklist

## ELI5

The TokenLimiterProcessor is a tool that counts the length of conversations to prevent them from getting too long. When AI models generate responses, they sometimes include special marker sequences (like `<|endoftext|>`) that were causing the processor to crash. This fix tells the token counter to treat these special markers as regular text instead of rejecting them, so conversations with AI-generated content can be measured without errors.

## Overview

This PR fixes a bug in `TokenLimiterProcessor` where the component would crash when processing text containing tiktoken special-token strings such as `<|endoftext|>`. The root cause was that token counting calls to `encoder.encode()` were rejecting these special tokens by default. The solution centralizes token counting through a new helper function that calls `encoder.encode(text, 'all')`, which treats special-token sequences as ordinary text for counting purposes.

## Changes

**Implementation (`token-limiter.ts`):**
- Added a `countTokens()` helper function that wraps `encoder.encode(text, 'all').length` to safely count tokens in any text, including special-token strings
- Refactored all token-counting calls throughout the processor to use this helper:
  - System message token accounting
  - Input message content token accounting (across various content types and tool handling paths)
  - Streaming chunk token counting (text deltas, object chunks, tool-call results, tool-result chunks)
  - Output truncation calculations (non-streaming text parts and binary-search truncation loop)

**Tests (`token-limiter.test.ts`):**
- Added test coverage for text-delta chunks containing `"<|endoftext|>"` to verify they are processed without truncation
- Added test coverage for tool-result chunks containing special token strings
- Added test coverage for final output text containing special token strings (ensuring no unintended truncation)
- Added test scenarios for system messages and tool-result messages with special-token strings to verify successful processing

**Metadata:**
- Added changeset entry for a patch release of `@mastra/core`

## Impact

The processor now gracefully handles AI-generated content that includes special-token sequences, preventing crashes during token counting while maintaining the intended behavior of tracking token budgets across conversations and managing output truncation.

### DIFF
--- a/packages/core/src/processors/processors/token-limiter.ts
+++ b/packages/core/src/processors/processors/token-limiter.ts
@@ -93,6 +93,10 @@ export class TokenLimiterProcessor implements Processor<'token-limiter', TokenLi
     return this.encoderPromise;
   }
 
+  private countTokens(encoder: Tiktoken, text: string): number {
+    return encoder.encode(text, 'all').length;
+  }
+
   /**
    * Process input messages at each step of the agentic loop, before they are sent to the LLM.
    * Runs at every step (including tool call continuations), preventing the conversation history
@@ -195,7 +199,7 @@ export class TokenLimiterProcessor implements Processor<'token-limiter', TokenLi
     const encoder = await this.getEncoder();
     const tokenString = message.role + message.content;
 
-    return encoder.encode(tokenString).length + TokenLimiterProcessor.TOKENS_PER_MESSAGE;
+    return this.countTokens(encoder, tokenString) + TokenLimiterProcessor.TOKENS_PER_MESSAGE;
   }
 
   /**
@@ -266,7 +270,7 @@ export class TokenLimiterProcessor implements Processor<'token-limiter', TokenLi
       overhead += toolResultCount * TokenLimiterProcessor.TOKENS_PER_MESSAGE;
     }
 
-    const tokenCount = encoder.encode(tokenString).length;
+    const tokenCount = this.countTokens(encoder, tokenString);
     const total = tokenCount + overhead;
     return total;
   }
@@ -326,12 +330,12 @@ export class TokenLimiterProcessor implements Processor<'token-limiter', TokenLi
     const encoder = await this.getEncoder();
     if (part.type === 'text-delta') {
       // For text chunks, count the text content directly
-      return encoder.encode(part.payload.text).length;
+      return this.countTokens(encoder, part.payload.text);
     } else if (part.type === 'object') {
       // For object chunks, count the JSON representation
       // This is similar to how the memory processor handles object content
       const objectString = JSON.stringify(part.object);
-      return encoder.encode(objectString).length;
+      return this.countTokens(encoder, objectString);
     } else if (part.type === 'tool-call') {
       // For tool-call chunks, count tool name and args
       let tokenString = part.payload.toolName;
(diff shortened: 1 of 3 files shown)


packages/core/src/processors/processors/token-limiter.ts
```diff
@@ -93,6 +93,10 @@ export class TokenLimiterProcessor implements Processor<'token-limiter', TokenLi
     return this.encoderPromise;
   }
 
+  private countTokens(encoder: Tiktoken, text: string): number {
+    return encoder.encode(text, 'all').length;
```
wardpeet, line 97: is 'all' the fix here?

wardpeet (approved):

packages/core/src/processors/processors/token-limiter.ts
intojhanurag: Yes

PR #16321, opened 2026-05-07
fix(server): pass auth context to dataset experiments

smthomas:
This fixes dataset-triggered workflow experiments so they receive the same authenticated request context as direct workflow runs.

The route now keeps the adapter-injected auth context separate from any body-provided experiment context, then merges them before starting the experiment. Body requestContext values still win for explicit overrides.

Before, a workflow run started directly from Studio could read auth/user context, but the same workflow started through a dataset experiment could lose it.

After, dataset experiments pass the authenticated context through to workflow execution while preserving explicit experiment context values.

Tested with:
- pnpm --filter ./packages/server test src/server/handlers/datasets.test.ts
- pnpm build:server
- adapter builds for express, hono, fastify, koa, nestjs
- pnpm --filter ./packages/server check:core-imports

## ELI5

This PR makes sure dataset-started workflows use the server’s authenticated user information. It also prevents client data from replacing protected authentication data.

## Changes

- Added a Hono adapter regression test.
- Verified that authenticated request context overrides client-supplied `MASTRA_USER_KEY` values.
- Verified that non-reserved client context values remain available.
- Added the `MASTRA_USER_KEY` import.


YujohnNattrass:
Possible identity override via the new merge

After this PR the handler merges as:

```ts
const requestContext = {
  ...authRequestContext,
  ...(rawBodyRequestContext ?? bodyRequestContext),
};
```

The catch is that auth middleware writes **two** user keys (see `packages/server/src/server/auth/helpers.ts:444-450`):

```ts
requestContext.set(MASTRA_USER_KEY, user); // 'mastra__user' — reserved
// Backward-compat: also write the legacy `'user'` key so existing
// middleware and integrations that read `requestContext.get('user')`
// (including built-in FGA route enforcement, memory handlers, and the
// documented public surface) keep working.
requestContext.set('user', user);          // 'user' — NOT reserved
```

`'user'` is **not** in `RESERVED_REQUEST_CONTEXT_KEYS` (`packages/server/src/server/constants.ts:36-45`, only `mastra__*` keys are). So a client POSTing:

```json
{ "requestContext": { "user": { "id": "spoofed", "roles": ["admin"] } } }
```

…will produce a merged `requestContext` of:

```js
{ mastra__user: <real>, user: <spoofed>, ... }
```

Anything downstream in the experiment's workflow that reads `.get('user')` (the comment in `helpers.ts` explicitly calls out memory handlers and FGA route enforcement as consumers of the legacy key) sees the spoofed value. Consumers that read `MASTRA_USER_KEY` are safe.

Pre-PR this was harder to hit because the adapter-injected `RequestContext` instance was set *after* the body spread in `handlerParams`, so the auth-set `'user'` on that instance was authoritative. The new explicit shallow merge changes that.

The new test `should merge body request context over adapter-injected request context` even codifies the new precedence for `tenantId` — the same rule applies to `user`.

Possible fixes:

- Strip auth-owned keys from `bodyRequestContext` before merging (e.g. extend `omitReservedKeys` / a separate allowlist that also covers `user`, `userPermissions`, `userRoles`, `userId`).
- Reverse the merge order so auth wins: `{ ...body, ...auth }` (matches the pre-PR effective behaviour).
- If body-overrides-auth is intentional, document it on the route and make sure workflow consumers prefer `MASTRA_USER_KEY` over `'user'` for any authz decision.

NikAiyer (commented):
The protected merge behavior looks correct: reserved and auth-owned values remain server-controlled, while non-auth experiment context retains the intended body-wins precedence.

My remaining feedback is about keeping the generic context contract synchronized, exercising it through the actual routing boundary, and making the raw-context trust boundary maintainable. I'm leaving this as a comment review rather than requesting changes.

packages/server/src/server/server-adapter/routes/index.ts
```diff
@@ -50,6 +50,8 @@ import { WORKSPACE_ROUTES } from './workspace';
 export type ServerContext = {
   mastra: Mastra;
   requestContext: RequestContext;
+  /** The unmerged request context supplied in the request body. */
+  bodyRequestContext?: Record<string, unknown>;
```
NikAiyer: Since `bodyRequestContext` is being added to the generic handler context, should the parallel `ServerContext` declaration in `packages/_internal-core/src/routes/index.ts` be updated as well?

The direct handler tests currently use `as any`, so they will not detect drift between these route-context contracts.

packages/server/src/server/handlers/datasets.test.ts
```diff
@@ -88,6 +96,92 @@ describe('Datasets Handlers', () => {
...
+          user: { id: 'spoofed-user' },
+          [MASTRA_AUTH_TOKEN_KEY]: 'spoofed-token',
+          [MASTRA_USER_PERMISSIONS_KEY]: ['*:write'],
+          userPermissions: ['*:write'],
+          [MASTRA_USER_ROLES_KEY]: ['admin'],
+          userRoles: ['admin'],
+          tenantId: 'body-tenant',
+          traceId: 'trace-1',
+        },
+      } as any);
```
NikAiyer: The handler-level assertions cover filtering and precedence well. Could we also add one representative routing or adapter-boundary regression test?

The current tests inject `bodyRequestContext` directly with `as any`, so they do not verify that an HTTP body's `requestContext` is preserved separately from adapter/auth context and reaches this handler correctly. One representative adapter should be sufficient if the implementations share the same contract.

packages/server/src/server/server-adapter/routes/index.ts
```diff
@@ -50,6 +50,8 @@ import { WORKSPACE_ROUTES } from './workspace';
 export type ServerContext = {
   mastra: Mastra;
   requestContext: RequestContext;
+  /** The unmerged request context supplied in the request body. */
```
NikAiyer: Could we document that `bodyRequestContext` is the unmerged, caller-controlled HTTP value and must not be consumed without applying the appropriate reserved/auth-owned-key policy?

Unlike `requestContext`, this value has intentionally not passed through the normal protected merge. Making that distinction explicit would reduce the chance that another handler later spreads it directly into trusted context.

packages/server/src/server/handlers/datasets.ts
```diff
@@ -59,6 +60,20 @@ function assertDatasetsAvailable(): void {
   }
 }
 
+// Auth middleware owns these legacy aliases in addition to the reserved
+// mastra__* keys. Do not let caller-supplied experiment context spoof them.
+const AUTH_OWNED_BODY_REQUEST_CONTEXT_KEYS = new Set(['user', 'userPermissions', 'userRoles']);
```
NikAiyer: Could the auth-owned aliases be exposed through a shared helper or constant near the authentication/request-context definitions?

Keeping a separate dataset-local list means a future auth compatibility alias could be introduced without automatically becoming protected here.

abhiaiyer91 (changes requested):
**Verdict: request changes** — this supersedes my earlier comment approving the merge semantics. I went back and ran the PR's tests against the base handler, and the result contradicts the stated motivation.

## The regression test for the reported bug passes without the fix

I checked out the base `packages/server/src/server/handlers/datasets.ts` (`a440c311`) into the PR branch, leaving the new test file in place:

```
✓ TRIGGER_EXPERIMENT_ROUTE > should forward adapter-injected auth request context to workflow experiments
× TRIGGER_EXPERIMENT_ROUTE > should allow body request context to override non-auth keys without spoofing auth-owned keys
```

The test that encodes the bug described in the PR body — "a workflow run started directly from Studio could read auth/user context, but the same workflow started through a dataset experiment could lose it" — **already passes on base**. Only the override test fails, and that test asserts new behavior rather than the reported defect.

## Why the auth context was never lost

The path was already intact end to end:

1. `createContextMiddleware()` merges body + query context into a single `RequestContext` instance and stores it (`server-adapters/express/src/index.ts:116-123`). Same shape in fastify/hono/koa/nestjs.
2. `checkRouteAuth()` runs per route *after* that middleware and sets `mastra__user`, `user`, `mastra__authToken`, `mastra__userPermissions`, `userPermissions`, `mastra__userRoles`, `userRoles` on that **same** instance.
3. `handlerParams` spreads the body first and assigns `requestContext:` afterwards (base `express/src/index.ts:542-546`), so a body-supplied `requestContext` field could never clobber the injected instance.
4. The base handler already normalized it: `rawRequestContext instanceof RequestContext ? rawRequestContext.all : rawRequestContext`.

Auth keys were written last, so they were present and authoritative.

## What the PR actually changes: a precedence flip on one route

`{ ...authRequestContext, ...filterBodyRequestContext(bodyRequestContext) }` inverts the ordering. Body now wins over adapter-injected values for every key that isn't reserved (`mastra__*`) or one of the three aliases in `AUTH_OWNED_BODY_REQUEST_CONTEXT_KEYS`.

The built-in auth keys are all covered by that filter, so token/user/role spoofing is genuinely blocked. The gap is everything else. Custom `server.middleware` and custom auth handlers are a documented pattern (`docs/src/content/en/docs/server/middleware.mdx`) and routinely set non-reserved keys derived from a validated token. The PR's own test locks in the risky case:

```ts
requestContext.set('tenantId', 'auth-tenant');   // trusted, set server-side
bodyRequestContext: { tenantId: 'body-tenant' }  // caller-supplied
// asserts: tenantId === 'body-tenant'
```

A tenant identifier resolved by trusted middleware becomes client-overridable — on this route only. On every other route the merge-then-auth ordering still makes the server value win. Same key, same deployment, two different trust models depending on the endpoint.

## Cost side

`bodyRequestContext` is added to `ServerContext` and the extraction logic is duplicated across all five adapters, for a single consumer. That is a permanent widening of the adapter/handler contract in exchange for behavior the base already provided.

## What I'd need to move to approve

Either:

- **(a)** A failing repro for the original bug — a test or trace where the base code drops auth context for a dataset experiment. If there's a real path I've missed (a specific adapter, custom route registration, or client-js call shape), point me at it and I'll re-run against it. If it exists, the fix likely belongs where the context is dropped, not in a per-route re-merge.
- **(b)** Re-frame this as an intentional precedence change: retitle away from `fix(server)`, justify why explicit experiment context should outrank middleware-derived context, and apply it consistently rather than only for `TRIGGER_EXPERIMENT_ROUTE`. `tenantId` overriding specifically needs an explicit call — I'd expect tenancy to stay server-owned.

Happy to be wrong on (a); the reading above is from the base code and the run is reproducible with the swap described.

packages/server/src/server/server-adapter/routes/index.ts
smthomas: After merging current main, I removed `bodyRequestContext` entirely. The adapters and dataset handler now use the single authoritative `ServerContext.requestContext`, so there is no parallel field to add to `_internal-core` and no route-context contract drift.

packages/server/src/server/handlers/datasets.test.ts
smthomas: Added a representative authenticated Hono boundary regression in `server-adapters/hono/src/__tests__/hono-adapter.test.ts`. It posts spoofed namespaced and legacy `user` values plus a benign custom value, then verifies authentication remains authoritative while the custom body context survives.

packages/server/src/server/server-adapter/routes/index.ts
smthomas: Resolved by removing the raw `bodyRequestContext` field and plumbing rather than exposing caller-controlled context as a second handler contract. Handlers now receive only the normal merged/authenticated `requestContext`, so this unsafe consumption path no longer exists.

packages/server/src/server/handlers/datasets.ts
smthomas: The dataset-local auth-owned alias list and re-merge have been removed. Current main merges caller context through the adapter protection path, then auth writes the trusted namespaced and legacy aliases onto that same `RequestContext`; the new Hono regression covers both alias forms.

PR #16331, opened 2026-05-07
fix(playground): stabilize streaming message rendering

TylerBarnes:
This fixes two Studio rendering issues that showed up during fast streaming output.

Reasoning chunks now stay separated when a model interleaves reasoning, tool calls, and text instead of appending everything back into the first reasoning block.

Before:
```tsx
reasoning: "first thought second thought third thought"
tool call
text
```

After:
```tsx
reasoning: "first thought"
tool call
reasoning: "second thought"
text
reasoning: "third thought"
```

Autoscroll also keeps following rapid output from parallel tool calls unless the user intentionally scrolls away, so content growth alone does not disable follow mode.

## Test plan
- `pnpm --filter ./client-sdks/react test src/lib/ai-sdk/utils/toUIMessage.test.ts -- --bail 1 --reporter=dot`
- `pnpm --filter ./packages/playground-ui test -- --bail 1 --reporter=dot`
- `pnpm build:playground-ui`
- `pnpm --filter ./packages/playground build`

### DIFF
--- a/packages/playground-ui/src/hooks/use-autoscroll.tsx
+++ b/packages/playground-ui/src/hooks/use-autoscroll.tsx
@@ -4,8 +4,13 @@ export interface UseAutoscrollOptions {
   enabled?: boolean;
 }
 
+const SCROLL_END_THRESHOLD = 8;
+
 export const useAutoscroll = (ref: React.RefObject<HTMLElement | null>, { enabled = true }: UseAutoscrollOptions) => {
   const shouldScrollRef = useRef(enabled);
+  const scrollFrameRef = useRef<number | null>(null);
+  const userScrollIntentRef = useRef(false);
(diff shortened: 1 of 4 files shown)


abhiaiyer91 (approved):

TylerBarnes:
Split the changeset into separate  and  entries and reran the narrow validations.

packages/playground-ui/src/hooks/use-autoscroll.tsx
```diff
@@ -4,45 +4,125 @@ export interface UseAutoscrollOptions {
...
+        return;
+      }
+
+      if (userScrollIntentRef.current) {
         shouldScrollRef.current = false;
+        cancelPendingScroll();
       }
     };
 
+    area.addEventListener('wheel', handleWheel, { passive: true });
```
mfrachet, line 105: why do we need all those listeners?

PR #16365, opened 2026-05-08
Fix ToolSearchProcessor approval resume

### LINKED ISSUE #16350: ToolSearchProcessor and requireApproval causing tool approval loop
This issue was created from Discord post 1502242184030584933:

[Open in Discord](https://discord.com/channels/1309558646228779139/1502242184030584933)

Any tool that is being lazy loaded with ToolSearchProcessor causes loop since after approval the tool isn't available in the agent context.  I see `tool-error` in the chunk with ToolNotFoundError.

After approval the tool should be loaded into the agent context automatically.

graysonhicks:
## Summary

Fixes approval resume for dynamically loaded tools from `ToolSearchProcessor`, including when the processor is wrapped in a processor workflow.

Previously, a tool loaded through `load_tool` could suspend for approval, but `approveToolCall()` resumed into a tool map that no longer contained the dynamically loaded executor. The resumed run could then loop or emit a tool error because the approval path rebuilt agent tools without consulting processor-local loaded tool state.

Closes #16350.

## What changed

- Added a `ToolSearchProcessor#getLoadedToolsForRequestContext()` hook so the agent can ask a processor for the tools it has already loaded for the active request/thread.
- Updated agent tool assembly to collect loaded tools from configured input processors before formatting the final tool map.
- Converted processor-loaded Mastra/provider tools through `makeCoreTool()` with the same execution context used by other agent tools, preserving:
  - `runId`, `threadId`, `resourceId`
  - `requestContext`
  - memory, model, workspace, logger, and Mastra proxy
  - `requireApproval`
  - background task settings
  - tracing and output writer context
- Added processor workflow traversal so loaded tools can also be recovered when `ToolSearchProcessor` is nested inside a workflow.
- Updated processor-to-workflow-step conversion to preserve the loaded-tool hook on generated workflow steps.
- Added changesets for the `@mastra/core` patch behavior.

## Why this fixes the issue

The initial stream has the dynamic tool available because `ToolSearchProcessor` runs during input processing and adds the loaded tool into its own state. Approval resume is different: the resumed workflow enters the suspended tool-call path and does not re-run the original `processInputStep` that loaded the executor.

This patch makes `Agent.convertTools()` rebuild the tool map from processor-loaded state as part of resume setup. That means the approved tool call can resolve the same dynamic tool executor that was available before suspension.

## Test coverage

Added regression coverage in `packages/core/src/agent/__tests__/tool-approval-standalone-repro.test.ts` for:

- direct `ToolSearchProcessor` usage
- `ToolSearchProcessor` wrapped in a processor workflow
- a dynamically loaded tool with `requireApproval: true`
- streaming until `tool-call-approval`
- resuming through `agent.approveToolCall(...)`
- asserting no `tool-error` chunks are emitted after resume
- asserting the dynamically loaded tool actually executes with the approved input

## Review notes

The main review risk is in `packages/core/src/agent/agent.ts`, where processor-loaded tools are merged into the final tool map. Please check that the conversion context matches the rest of agent tool conversion and that precedence is appropriate.

The workflow-specific risk is in `packages/core/src/workflows/workflow.ts`, where the processor hook is copied onto the generated step so workflow-wrapped processors remain discoverable.

This does not make a `ToolSearchProcessor` configured only on a supervisor agent available inside a subagent. The processor must be configured on the agent whose suspended tool call is being resumed.

## ELI5

Imagine you have a tool that gets loaded on-demand (like finding a tool in a toolbox when needed) and it needs approval to run. The problem was: after someone approved it, the agent forgot where the tool was and couldn't find it to actually use it. This fix teaches the agent to remember where it got the tool from, so after approval it can find and run it successfully.

## Summary

This PR fixes a bug in approval-resume flows for tools dynamically loaded via `ToolSearchProcessor`. Previously, when a tool loaded from a processor required approval and the agent suspended to ask for it, resuming after approval would fail because the dynamically loaded tool was no longer available in the agent's tool context, causing tool-not-found errors or loops.

### Changes

**ToolSearchProcessor** now exposes `getLoadedToolsForRequestContext()`, a public method that returns tools the processor has already loaded for the active request context (extracted from thread ID).

**Agent** was updated to collect dynamically loaded tools from configured input processors during tool assembly:
- Added `listInputProcessorLoadedTools()` to resolve processors and call their `getLoadedToolsForRequestContext()` methods
- These loaded tools are converted to `CoreTool` instances with the same execution context as other agent tools (runId, threadId, resourceId, requestContext, memory, model, workspace, logger, Mastra proxy, requireApproval, background tasks, tracing, output writer)
- Loaded tools are merged into the aggregated tool set before formatting, ensuring they're available for resumed runs

**Workflow** processor handling was enhanced to support nested `ToolSearchProcessor` workflows:
- Added recursion to traverse nested processor workflows and recover loaded tools
- `createStepFromProcessor()` now preserves the `getLoadedToolsForRequestContext` function when converting processors to workflow steps, binding it to the original processor

**Test coverage** added for approval-resume flows with `ToolSearchProcessor`:
- Direct processor configuration in inputProcessors
- Processor embedded in a processor workflow  
- Nested workflow wrapped with an additional `createStep(...)` layer
- Tests verify successful tool execution post-approval and confirm no tool-error chunks appear after resume

### Scope

The fix applies to any agent with a suspended tool call, including subagents in supervisor patterns. Regression testing covers the nested processor workflow case on a direct agent, though explicit end-to-end supervisor integration tests were not added. The processor must be configured on the agent whose tool call is being resumed (not inherited from a parent supervisor).

### Risk Areas

Primary review focus on `agent.ts` (tool merging/conversion context and precedence) and `workflow.ts` (copying processor hook onto generated steps).

[Review Change Stack](https://app.coderabbit.ai/change-stack/mastra-ai/mastra/pull/16365)


amitava82:
Does this address case where subagent has ToolSearchProcessor in Supervisor agent pattern?

graysonhicks:
> Does this address case where subagent has ToolSearchProcessor in Supervisor agent pattern?

Yes, for the important case: **a subagent itself has `ToolSearchProcessor`, loads a tool, suspends for approval, and is later resumed through the supervisor’s agent-tool flow.**

Why:
- In the supervisor path, subagent resume calls go through `resolvedAgent.resumeStream(...)` / `resumeGenerate(...)`, so tool rebuilding happens on the **subagent**, not the supervisor.
- The fix is in `Agent.convertTools()` via [agent.ts](/Users/grayson/conductor/workspaces/mastra/calgary-v5/packages/core/src/agent/agent.ts:2979), so it applies to any `Agent` instance, including subagents.
- If the subagent’s `ToolSearchProcessor` was wrapped in a processor workflow, the new recursion finds the nested processor step and calls `getLoadedToolsForRequestContext`.
- [workflow.ts](/Users/grayson/conductor/workspaces/mastra/calgary-v5/packages/core/src/workflows/workflow.ts:1443) preserves that hook when processors are converted into workflow steps.

Caveat: I added regression coverage for the nested processor workflow approval-resume path on a direct agent, not a full supervisor integration test. So the code path should cover supervisor subagents, but there is not yet an explicit supervisor-pattern test asserting it end to end.

It does **not** make a `ToolSearchProcessor` configured only on the supervisor available inside the subagent. The processor must be on the subagent whose tools are being resumed.

mbenhamd (commented):
LGTM - Thank you :)

mbenhamd (approved):

PR #12573, opened 2026-02-02
fix(core): use toJSON for requestContext serialization in all workflow paths

owk-owk130:
## Summary

Fixes #12301 (follow-up to #12344)

While `RequestContext.toJSON()` was fixed in #12344 to skip non-serializable values, there were other code paths that serialized `requestContext` without calling `toJSON()`:

- `DefaultExecutionEngine.serializeRequestContext()` - used `forEach` to copy values directly
- `EventedWorkflow.start()` / `startAsync()` - used `Object.fromEntries()`
- `EventedExecutionEngine.execute()` - used `Object.fromEntries()`

**Changes:**
- `DefaultExecutionEngine.serializeRequestContext()` now uses `RequestContext.toJSON()`
- `EventedWorkflow.start()` and `startAsync()` now use `requestContext.toJSON()`
- `EventedExecutionEngine.execute()` now uses `requestContext.toJSON()`

## Test plan

- [x] Add tests for `serializeRequestContext()` with serializable values
- [x] Add tests for `serializeRequestContext()` with non-serializable values (functions, circular references, RPC proxies)
- [x] Run `pnpm test` to verify all tests pass

## Summary by CodeRabbit

* **Bug Fixes**
  * Workflow snapshots and event/publish payloads now skip or sanitize non-serializable values (functions, symbols, circular refs, proxy objects), preventing persistence and restore errors.

* **Tests**
  * Added tests verifying serializable values are preserved and non-serializable values are excluded.

* **Chores**
  * Introduced a safer JSON stringification helper and switched snapshot/database serialization to use it.

* **Documentation**
  * Added a changelog entry describing the fix.

PR #16368, opened 2026-05-08
fix(libsql): prevent safeStringify from dropping shared object references

roaminro:
Workflow runs were being persisted without their top-level `result` field because libsql's `safeStringify` was treating shared (non-circular) object references as cycles.

The cycle-detection used a single `WeakSet` that was added to but never cleared, so any object that appeared more than once in the graph got dropped. In workflow snapshots, `snapshot.result` and `context[step].output` are the same reference on success, so `snapshot.result` was being silently stripped on every persist.

This broke `listWorkflowRuns` (snapshots came back with `result: undefined`) and workflow resume.

Before:
```ts
const seen = new WeakSet();
const sanitize = (val) => {
  if (seen.has(val)) return undefined; // dropped on second visit
  seen.add(val);
  // ...recurse...
};
```

After:
```ts
const ancestors = new Set();
const sanitize = (val) => {
  if (ancestors.has(val)) return undefined; // only true cycles
  ancestors.add(val);
  try { /* ...recurse... */ } finally { ancestors.delete(val); }
};
```

Regression from #12573.

## ELI5

A serializer was mistakenly throwing away shared toys when it saw the same object twice. The fix makes it only remove objects that actually form a loop, so shared objects are kept and workflow snapshots no longer lose data.

### Fix: correct circular-reference detection in safeStringify
- File: stores/libsql/src/storage/db/utils.ts
- Replaced a global WeakSet that tracked every visited object with a per-call ancestor Set that tracks only objects on the current recursion path. Values are added on entry and removed on exit (try/finally), so only true cycles are detected.
- Result: shared (non-circular) object references are no longer treated as cycles and are preserved during serialization. This fixes a regression where snapshot.result was being stripped when it referenced the same object as context[step].output.

### Tests added
- File: stores/libsql/src/storage/db/safe-stringify.test.ts
  - New unit tests for safeStringify covering: true cycles, functions, symbols, BigInt, toJSON, throwing getters/Proxy behavior, arrays, undefined input, and importantly shared-but-non-circular reference preservation and regression cases.
- File: stores/libsql/src/storage/domains/workflows/index.test.ts
  - Regression test using an in-memory LibSQL instance verifying that persisting/loading workflow snapshots preserves the shared reference between snapshot.result and context[step].output across loadWorkflowSnapshot, getWorkflowRunById, and listWorkflowRuns.

### Changeset
- File: .changeset/fast-pianos-grow.md
  - New changeset documenting the @mastra/libsql patch that fixes workflow run snapshot persistence where repeated references were incorrectly dropped.

## Notes
- No public API signatures changed.
- This restores correct behavior for listing and resuming workflow runs that relied on shared object references in persisted snapshots.

### DIFF
--- a/.changeset/fast-pianos-grow.md
+++ b/.changeset/fast-pianos-grow.md
@@ -1,0 +1,5 @@
+---
+'@mastra/libsql': patch
+---
+
+**Fixed** Workflow run snapshots no longer lose fields when serialized for storage. The libsql `safeStringify` cycle-detection treated any object that appeared more than once in a snapshot as a circular reference and dropped it. Because `snapshot.result` and the final step's `context[step].output` share the same reference on success, `snapshot.result` was being silently stripped on every persist. This caused `listWorkflowRuns` to return runs with `snapshot.result === undefined` and broke workflow resume when suspended-state fields were shared elsewhere in the snapshot.
--- a/stores/libsql/src/storage/db/safe-stringify.test.ts
+++ b/stores/libsql/src/storage/db/safe-stringify.test.ts
@@ -1,0 +1,125 @@
+import { describe, it, expect } from 'vitest';
+
+import { safeStringify } from './utils';
+
+describe('safeStringify', () => {
+  describe('preserves the original sanitization behaviour', () => {
+    it('drops true circular references', () => {
+      const a: Record<string, any> = { name: 'a' };
+      a.self = a;
+
+      const json = safeStringify(a);
+
+      expect(JSON.parse(json)).toEqual({ name: 'a' });
+    });
+
+    it('drops circular references nested deeper in the graph', () => {
+      const root: Record<string, any> = { name: 'root' };
+      const child: Record<string, any> = { name: 'child' };
+      root.child = child;
+      child.parent = root;
+
+      const json = safeStringify(root);
+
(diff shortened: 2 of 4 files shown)


TylerBarnes (changes requested):
seems like it makes sense, though from reading this I'm not sure if the original thing the code was doing has regressed. Do we have tests for the original behaviour? And we should add tests for the new behaviour too

roaminro:
AI says: Thanks for the review!

The original behavior is preserved — I confirmed by reverting just the fix and running the new tests: the 8 tests covering the original behavior (true circular refs, functions, symbols, BigInt, toJSON-honoring, RPC-proxy-style throwing getters, recursive arrays, undefined input) still pass with the buggy version. Only the 4 new shared-reference tests fail without the fix.

Added two test files in 251812c72f:

- `stores/libsql/src/storage/db/safe-stringify.test.ts` — direct unit coverage for both the original sanitization behavior and the shared-reference regression
- `stores/libsql/src/storage/domains/workflows/index.test.ts` — workflows-domain regression test that persists a snapshot whose top-level `result` and `context[step].output` point at the same object, then asserts the field round-trips through `loadWorkflowSnapshot`, `getWorkflowRunById`, and `listWorkflowRuns` (this is exactly the scenario that was failing on CI for `should list runs for a workflow`).

Mastra Code

TylerBarnes (approved):
awesome, thanks!

PR #16185, opened 2026-05-04
feat(observability): route eval scores through ScoreEvent bus

intojhanurag:
## Description

Wires Mastra eval scores through the unified observability bus so every exporter implementing `onScoreEvent` automatically receives them. Replaces the deprecated per-exporter `addScoreToTrace` side-channel that bypassed the pipeline. Also unblocks Mastra spans appearing under Braintrust experiment eval spans (#11097) as a side effect of the unified path.

- Producer alignment: `packages/core/src/mastra/hooks.ts` no longer iterates exporters with `addScoreToTrace`. `MastraScorer.run()` (already on main from PR #14920) remains the single producer of `ScoreEvent`s through `mastra.observability.addScore()`.
- Consumer side: implemented `onScoreEvent` on Langfuse (`LangfuseClient.score.create`), Laminar (`/v1/evaluators/score`), Braintrust (`logger.logFeedback`), Datadog (`tracer.llmobs.submitEvaluation`), and LangSmith (`Client.createFeedback`).
- Backwards compatibility: Langfuse `addScoreToTrace` and Laminar `_addScoreToTrace` are preserved as `@deprecated` wrappers that forward to a shared private `submitScore` helper.
- Bug fix in `observability/mastra/src/recorded.ts`: `buildScoreEvent` now forwards `scorerName` and `targetEntityType` from `ScoreInput` to `ExportedScore` (without this, exporters silently fell back to `scorerId`).
- Tests: rewrote core hook test to assert the hook no longer publishes a `ScoreEvent`; added per-exporter `onScoreEvent` tests; added a regression test for `buildScoreEvent` field forwarding.
- Seven single-package changesets created (`@mastra/core`, `@mastra/observability`, `@mastra/langfuse`, `@mastra/laminar`, `@mastra/braintrust`, `@mastra/datadog`, `@mastra/langsmith`) per the project's changeset style guide.

## Related Issue(s)

Fixes #10896 (also unblocks #11097)

## Type of Change

- [ ] Bug fix (non-breaking change that fixes an issue)
- [x] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [ ] Test update

## Checklist

- [ ] I have made corresponding changes to the documentation (if applicable)
- [x] I have added tests that prove my fix is effective or that my feature works
- [ ] I have addressed all Coderabbit comments on this PR

## ELI5

[...]

PR #16371, opened 2026-05-08
fix(core): register workflow step scorers with Mastra

intojhanurag:
## Description

Registers workflow step scorers with the owning Mastra instance so workflow scorer runs can emit scores through the observability score pipeline.

This keeps static workflow scorers discoverable after workflow registration and registers dynamic workflow scorers at execution time, when request context is available.

## Related Issue(s)

Follow-up to #16185.

## Type of Change

- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [x] Test update

## Checklist

## Testing

- pnpm --filter ./packages/core test src/mastra/scorer-registration.test.ts
- git diff --check

## ELI5

Workflow step scorers are little checks that rate each step of a process. This PR makes sure those checkers are registered with Mastra (either when the workflow is added or at execution time) so their scores are emitted into the observability pipeline and tracked.

## Changes

**Changeset Addition** (`.changeset/workflow-scorers-register.md`)
- Adds a changelog entry (minor) for `@mastra/core` noting that configured workflow step scorers now automatically emit scores through the observability score pipeline.

**Core Registration Logic** (`packages/core/src/mastra/index.ts`)
- Added a private helper to register static workflow step scorers found on workflow steps.
- `Mastra.addWorkflow()` now invokes that helper after committing/registering a workflow so static (non-function) `step.scorers` are registered with source `'code'`.
- Removed the previous declarative schedule auto-registration side-effect (it no longer sets `#hasScheduledWorkflow` or triggers scheduler bootstrap/registration based on collected schedules).
- Documented `addScorer()` behavior regarding duplicate keys.

**Workflow Step Execution** (`packages/core/src/workflows/handlers/step.ts`)
- `runScorersForStep()` now ensures dynamic scorers are registered with `engine.mastra` before execution by invoking the scorer's internal register helper and calling `mastra.addScorer(...)`.
- When emitting scores, scorer identification now uses `scorerObject.scorer.id` (instead of the prior `scorerObject.name`).

**Tests** (`packages/core/src/mastra/scorer-registration.test.ts`)
- Expanded tests to cover workflow step scorer registration (static and dynamic) in addition to existing agent-level tests.
- New/updated test cases verify:
  - Static workflow step scorers are registered after `addWorkflow()` returns and are findable via `getScorerById`.
  - Shared static scorers are not duplicated (one registration per id).
  - Dynamic workflow step scorers are registered before they run.
  - Score events are emitted through `mastra.observability.addScore` with expected scorer id/name, `targetEntityType: 'workflow_run'`, and numeric score values.

## Impact

Fixes a bug where workflow step scorers (both static and dynamic) could be undiscoverable by Mastra; ensures they are registered so their emitted scores flow through the observability score pipeline and are captured for reporting and analysis.

[Review Change Stack](https://app.coderabbit.ai/change-stack/mastra-ai/mastra/pull/16371)

### DIFF
--- a/packages/core/src/mastra/index.ts
+++ b/packages/core/src/mastra/index.ts
@@ -2447,7 +2447,8 @@ export class Mastra<
    * This method allows dynamic registration of scorers after the Mastra instance
    * has been created.
    *
-   * @throws {MastraError} When a scorer with the same key already exists
+   * If a scorer with the same key already exists, this method leaves the existing
+   * scorer registered and returns.
    *
    * @example
    * ```typescript
--- a/packages/core/src/mastra/index.ts
+++ b/packages/core/src/mastra/index.ts
@@ -3278,6 +3278,8 @@ export class Mastra<
     }
     workflows[workflowKey] = workflow;
 
+    this.registerStaticWorkflowScorers(workflow);
+
     // If a schedule is declared, mark the flag and either register into the
     // running scheduler or trigger a lazy ensure.
     if (hasSchedule) {
@@ -3297,6 +3299,19 @@ export class Mastra<
         })();
       } else {
         this.#ensureScheduler();
+      }
+    }
+  }
+
+  private registerStaticWorkflowScorers(workflow: AnyWorkflow): void {
+    for (const step of Object.values(workflow.steps ?? {})) {
+      const scorers = step.scorers;
+      if (!scorers || typeof scorers === 'function') {
+        continue;
+      }
+
+      for (const [, entry] of Object.entries(scorers)) {
+        this.addScorer(entry.scorer, undefined, { source: 'code' });
       }
     }
   }
--- a/packages/core/src/workflows/handlers/step.ts
+++ b/packages/core/src/workflows/handlers/step.ts
@@ -572,8 +572,12 @@ export async function runScorersForStep(params: RunScorersParams): Promise<void>
 
   if (!disableScorers && scorersToUse && Object.keys(scorersToUse || {}).length > 0) {
     for (const [_id, scorerObject] of Object.entries(scorersToUse || {})) {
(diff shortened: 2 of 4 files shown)


packages/core/src/mastra/index.ts
```diff
@@ -3301,6 +3303,19 @@ export class Mastra<
...
 
+  private registerStaticWorkflowScorers(workflow: AnyWorkflow): void {
+    for (const step of Object.values(workflow.steps ?? {})) {
+      const scorers = step.scorers;
+      if (!scorers || typeof scorers === 'function') {
+        continue;
+      }
+
+      for (const [, entry] of Object.entries(scorers)) {
+        this.addScorer(entry.scorer, undefined, { source: 'code' });
```
epinzur, line 3315: how does this behave when a scorer with the same name is added more than once?

epinzur (approved):

packages/core/src/workflows/handlers/step.ts
```diff
@@ -572,8 +572,12 @@ export async function runScorersForStep(params: RunScorersParams): Promise<void>
 
   if (!disableScorers && scorersToUse && Object.keys(scorersToUse || {}).length > 0) {
     for (const [_id, scorerObject] of Object.entries(scorersToUse || {})) {
+      if (engine.mastra) {
+        scorerObject.scorer.__registerMastra(engine.mastra);
+        engine.mastra.addScorer(scorerObject.scorer, undefined, { source: 'code' });
+      }
       runScorer({
-        scorerId: scorerObject.name,
+        scorerId: scorerObject.scorer.id,
```
intojhanurag, line 580: I think it will cover , when name will be same . I don't know in the past anyone changed to name , but I reverted that .

Issue #16220, opened 2026-05-05
[BUG] @mastra/client-js drops providerMetadata on recursive client-tool stream — Gemini thought_signature lost


SebastianPodgajny:
## Summary

`@mastra/client-js` drops `providerMetadata` (e.g. `vertex.thoughtSignature`) when rebuilding the message stack for a recursive client-tool stream call. The metadata is placed inside the `toolInvocation` object instead of on the parent V2 part, so the server's `MessageList` adapters (which read it from `part.providerMetadata`) never see it on the next turn.

This was almost-but-not-quite fixed by #16078. That PR added the namespace merge in `MessageList.updateToolInvocation()` server-side, but the merge can only preserve metadata that arrives at `part.providerMetadata` — and the client SDK puts it elsewhere. So for any flow that goes through a `clientTool` continuation (the SDK's `agent.stream(..., { clientTools })` path), the signature is still lost.

## Impact

With Gemini thinking models (`gemini-3-flash-preview`, `gemini-3-pro-preview`, etc.) the recursive tool-result continuation fails:

```
AI_APICallError: Function call is missing a thought_signature in functionCall parts.
This is required for tools to work correctly...
```

Any `agent.stream()` call that uses `clientTools` against a Gemini thinking model fails on the second turn.

## Reproduction

```ts
import { MastraClient, createTool } from '@mastra/client-js';
import { z } from 'zod';

const client = new MastraClient({ baseUrl: 'http://127.0.0.1:4111' });
const agent = client.getAgent('myAgent'); // agent must use a Gemini thinking model

const res = await agent.stream(
  [{ role: 'user', content: 'Call presentList with kind=link items=[{href:"https://a",label:"A"}]' }],
  {
    memory: { thread: { id: crypto.randomUUID() }, resource: 'user-1' },
    clientTools: {
      presentList: createTool({
        id: 'presentList',
        description: 'Render a list',
        inputSchema: z.object({ kind: z.string(), items: z.array(z.unknown()) }),
        execute: async () => ({ ok: true }),
      }),
    },
  },
);
await res.processDataStream({ onChunk: c => console.log(c.type) });
// Stream emits a `tool-call` with providerMetadata, FE execute returns,
// recursion POSTs back without providerMetadata at part level → 400 from Gemini.
```

## Root cause

`processChatResponse_vNext` in `client-sdks/client-js/src/resources/agent.ts`:

```ts
const invocation = {
  state: 'call',
  step,
[...]

PR #16403, opened 2026-05-11
fix(core): preserve providerMetadata on durable tool-invocation parts

### LINKED ISSUE #17579: [BUG] providerMetadata and providerExecuted not persisted breaking Gemini models
### Describe the Bug

DurableAgent drops providerMetadata and providerExecuted when it persists tool-invocation parts into the message list, breaking Gemini multi-step tool flows.

### Steps To Reproduce

Use a gemini model agent with a durable agent and have a multi turn chat 

N/A

### Expected Behavior

Gemini models work with durable agent 

### Environment Information

```block
Not relevant (only making this issue to link a PR to)
```

### Verification

wouter-leistra:
Related: #15294 (`@mastra/memory` ObservationalMemory variant of the same Gemini `thoughtSignature` loss). Patches against `@mastra/memory` posted there don't cover this site — the durable executor builds tool-invocation parts on the initial step, before any memory processor runs.

## Summary

`DurableAgent` drops `providerMetadata` and `providerExecuted` when it persists tool-invocation parts into the message list, breaking Gemini multi-step tool flows.

In `createDurableLLMExecutionStep` (`packages/core/src/agent/durable/workflows/steps/llm-execution.ts`) the chunk handler captures both fields from each `tool-call` chunk into `toolCalls[]`:

```ts
toolCalls.push({
  toolCallId: payload.toolCallId,
  toolName: payload.toolName,
  args: payload.args || {},
  providerMetadata: payload.providerMetadata as Record<string, unknown> | undefined,
  providerExecuted: payload.providerExecuted,
  output: payload.output,
});
```

…but the V2 `tool-invocation` part built ~80 lines later only copies `state / toolCallId / toolName / args`. The metadata is silently dropped.

## Why this breaks Gemini

For Gemini thinking models (`gemini-3-flash-preview`, `gemini-3-pro-preview`, …) the AI-SDK normalizes `thoughtSignature` into `providerMetadata.google.thoughtSignature`. Vertex requires at least one `functionCall` per model content block to carry a `thoughtSignature`.

There are **two ways** this metadata gets lost:

1. **At persist time.** Even when `payload.providerMetadata` is set on the `tool-call` chunk, the V2-part construction drops it (described above).
2. **At capture time.** Vertex/Gemini frequently emits `thoughtSignature` on a *sibling* chunk (`reasoning` / `finish-step`) that precedes the `tool-call` chunk for the same call — so `payload.providerMetadata` on the tool-call chunk itself is `undefined`. Even if we fixed #1, the signature wouldn't be there to spread.

Once any part lands in the message list without that metadata, the next durable step rehydrates it, `convertToModelMessages` maps `part.providerMetadata` → `providerOptions`, and the Gemini adapter emits a `functionCall` without `thoughtSignature`. Vertex rejects with:

```
AI_APICallError: Function call is missing a thought_signature in functionCall parts.
This is required for tools to work correctly, and missing thought_signature may lead to
degraded model performance. Additional data, function call `default_api:<toolName>`, position N.
```

Non-durable `Agent` is unaffected because the AI-SDK keeps state in-process across its step loop. Durable execution round-trips through serialized state, so it has to carry provider metadata explicitly.

## Relation to existing issues

- **#15294** (`@mastra/memory` Observational Memory) — same class of bug at a different site: OM's `processInputStep` splits parallel tool calls into separate messages so blocks without `thoughtSignature` get sent. Community patch there only touches the memory site.
- **#16220** (`@mastra/client-js`) — recursive client-tool stream loses `providerMetadata`. Orthogonal site.
- **#16078** already fixed the *merge* in `MessageList.updateToolInvocation`, but the gap remains at the **initial** construction site in the durable executor — confirmed still present on `1.33.0-alpha.9`.

## Fix

Two complementary changes in `createDurableLLMExecutionStep`:

**1. Carry chunk-level metadata forward across the stream.** Track the most recent `providerMetadata` seen on any chunk and use it as a fallback when the `tool-call` chunk has none of its own. This handles the Vertex pattern where the signature arrives on a `reasoning` / `finish-step` chunk that precedes the tool-call chunk.

```ts
let pendingProviderMetadata: Record<string, unknown> | undefined = undefined;

for await (const chunk of trackedStream) {
  // …

  const chunkPayloadMeta = (chunk as { payload?: { providerMetadata?: Record<string, unknown> } })?.payload
    ?.providerMetadata;
  if (chunkPayloadMeta !== undefined) {
    pendingProviderMetadata = chunkPayloadMeta;
  }

  switch (chunk.type) {
    case 'tool-call': {
      toolCalls.push({
        // …
        providerMetadata:
          (payload.providerMetadata as Record<string, unknown> | undefined) ?? pendingProviderMetadata,
        // …
      });
      break;
    }
  }
}
```

**2. Spread the optional fields onto the persisted V2 part.** The interface `DurableToolCallInput` already declares both as optional, so no type changes are needed:

```ts
parts.push({
  type: 'tool-invocation' as const,
  toolInvocation: { state: 'call' as const, toolCallId: tc.toolCallId, toolName: tc.toolName, args: tc.args },
  ...(tc.providerMetadata !== undefined ? { providerMetadata: tc.providerMetadata } : {}),
  ...(tc.providerExecuted !== undefined ? { providerExecuted: tc.providerExecuted } : {}),
});
```

Both spreads are conditional — providers that don't attach `providerMetadata` see no behavior change.

## Test plan

- [x] Verified on `@mastra/core@1.32.1` via a local `yarn` patch carrying the equivalent diff: a `DurableAgent` over Gemini 3 Flash with a workflow tool + multiple per-step tool calls now completes follow-up steps cleanly. Without either part of the fix, the second step throws the `thought_signature` error from `dist/agent/durable/index.js` (the catch at "Error processing LLM stream").
- [ ] Happy to add a focused unit test on the part-construction site if you'd like — point me at the right pattern (didn't want to guess the test harness conventions for this step).

## ELI5

DurableAgent was throwing away small "thinking" notes Gemini puts into streamed tool-call events, which broke the next step of a multi-step flow; this change keeps those notes when saving the conversation so Gemini flows continue to work.

---

## Summary

This PR fixes a bug in DurableAgent where optional tool-invocation metadata (`providerMetadata` and `providerExecuted`) could be lost when persisting V2 `tool-invocation` parts, causing multi-step interactions with Gemini "thinking" models to fail.

### Problem

During stream processing, `createDurableLLMExecutionStep` captured chunk-level `providerMetadata` (e.g., `providerMetadata.google.thoughtSignature`) and used it when populating `toolCalls[]`, but the code that serialized the final assistant message parts persisted only `state`, `toolCallId`, `toolName`, and `args`. Dropping `providerMetadata` (which Gemini may emit on a sibling chunk) removed the `thought_signature` Vertex/Gemini expect, causing errors like: `AI_APICallError: Function call is missing a thought_signature ...`. Non-durable `Agent` instances were unaffected.

### Solution

When building persisted `tool-invocation` parts, the step now includes `providerMetadata` and `providerExecuted` when present. The stream processor also tracks the most recent chunk-level `providerMetadata` as a fallback so a preceding "thinking" chunk's metadata is preserved on the persisted part.

Example of the persisted part construction:
```typescript
parts.push({
  type: 'tool-invocation' as const,
  toolInvocation: { state: 'call' as const, toolCallId: tc.toolCallId, toolName: tc.toolName, args: tc.args },
  ...(tc.providerMetadata !== undefined ? { providerMetadata: tc.providerMetadata } : {}),
  ...(tc.providerExecuted !== undefined ? { providerExecuted: tc.providerExecuted } : {}),
});
```

No public API/type changes were required because `DurableToolCallInput` already declares these fields optional.

### Verification & Impact

- Locally verified on @mastra/core@1.32.1: multi-step Gemini tool flows complete.
- Scope: small patch in packages/core (packages/core/src/agent/durable/workflows/steps/llm-execution.ts) and an accompanying changeset entry.
- Affects durable agents created via \`createDurableAgent\` / \`new DurableAgent\` using Gemini "thinking" models.
- Author can add a unit test if maintainers provide guidance on test patterns.

[Review Change Stack](https://app.coderabbit.ai/change-stack/mastra-ai/mastra/pull/16403)

Fixes #17579


wouter-leistra:
Problem in core, also exists in @mastra/memory package. That will need a proper fix there, for now i run with the following hacky non-PR'able patch

```diff
diff --git a/dist/chunk-BPJLUC2F.js b/dist/chunk-BPJLUC2F.js
index 64dd35add46c9e51c5e3aee40e8d2dd6fc56f022..5360a50a177fc3992ac625ad127320718d0abfb4 100644
--- a/dist/chunk-BPJLUC2F.js
+++ b/dist/chunk-BPJLUC2F.js
@@ -9306,6 +9306,37 @@ var ObservationalMemoryProcessor = class {
           });
         }
       }
+      try {
+        var _allMsgs = messageList.get.all.db();
+        var _lastSig = void 0;
+        for (var _mi = 0; _mi < _allMsgs.length; _mi++) {
+          var _msg = _allMsgs[_mi];
+          if (!_msg || !_msg.content || !_msg.content.parts) continue;
+          for (var _pi = 0; _pi < _msg.content.parts.length; _pi++) {
+            var _part = _msg.content.parts[_pi];
+            if (!_part || !_part.providerMetadata) continue;
+            var _g = _part.providerMetadata.google || _part.providerMetadata.vertex;
+            if (_g && _g.thoughtSignature) _lastSig = _g.thoughtSignature;
+          }
+        }
+        if (_lastSig) {
+          for (var _mi2 = 0; _mi2 < _allMsgs.length; _mi2++) {
+            var _msg2 = _allMsgs[_mi2];
+            if (!_msg2 || !_msg2.content || !_msg2.content.parts) continue;
+            for (var _pi2 = 0; _pi2 < _msg2.content.parts.length; _pi2++) {
+              var _part2 = _msg2.content.parts[_pi2];
+              if (!_part2 || _part2.type !== "tool-invocation") continue;
+              if (!_part2.providerMetadata) _part2.providerMetadata = {};
+              var _hasGoogle = _part2.providerMetadata.google && _part2.providerMetadata.google.thoughtSignature;
+              var _hasVertex = _part2.providerMetadata.vertex && _part2.providerMetadata.vertex.thoughtSignature;
+              if (!_hasGoogle && !_hasVertex) {
+                if (!_part2.providerMetadata.google) _part2.providerMetadata.google = {};
+                _part2.providerMetadata.google.thoughtSignature = _lastSig;
+              }
+            }
+          }
+        }
+      } catch (_e) {}
       return messageList;
     });
   }
diff --git a/dist/chunk-UZDSNIGD.cjs b/dist/chunk-UZDSNIGD.cjs
index d132f169976675b0c03f765e39b19cf384825898..d43652f1b1170660ab083b9f637db4afc5df2d39 100644
--- a/dist/chunk-UZDSNIGD.cjs
+++ b/dist/chunk-UZDSNIGD.cjs
@@ -9313,6 +9313,37 @@ var ObservationalMemoryProcessor = class {
           });
         }
       }
+      try {
+        var _allMsgs = messageList.get.all.db();
+        var _lastSig = void 0;
+        for (var _mi = 0; _mi < _allMsgs.length; _mi++) {
+          var _msg = _allMsgs[_mi];
+          if (!_msg || !_msg.content || !_msg.content.parts) continue;
+          for (var _pi = 0; _pi < _msg.content.parts.length; _pi++) {
+            var _part = _msg.content.parts[_pi];
+            if (!_part || !_part.providerMetadata) continue;
+            var _g = _part.providerMetadata.google || _part.providerMetadata.vertex;
+            if (_g && _g.thoughtSignature) _lastSig = _g.thoughtSignature;
+          }
+        }
+        if (_lastSig) {
+          for (var _mi2 = 0; _mi2 < _allMsgs.length; _mi2++) {
+            var _msg2 = _allMsgs[_mi2];
+            if (!_msg2 || !_msg2.content || !_msg2.content.parts) continue;
+            for (var _pi2 = 0; _pi2 < _msg2.content.parts.length; _pi2++) {
+              var _part2 = _msg2.content.parts[_pi2];
+              if (!_part2 || _part2.type !== "tool-invocation") continue;
+              if (!_part2.providerMetadata) _part2.providerMetadata = {};
+              var _hasGoogle = _part2.providerMetadata.google && _part2.providerMetadata.google.thoughtSignature;
+              var _hasVertex = _part2.providerMetadata.vertex && _part2.providerMetadata.vertex.thoughtSignature;
+              if (!_hasGoogle && !_hasVertex) {
+                if (!_part2.providerMetadata.google) _part2.providerMetadata.google = {};
+                _part2.providerMetadata.google.thoughtSignature = _lastSig;
+              }
+            }
+          }
+        }
+      } catch (_e) {}
       return messageList;
     });
   }
```

wouter-leistra:
Fixes #17579

TobyColeman:
Would love to see this in a patch soon if possible. Maintaining a few patches for Mastra right now and it's starting to get a bit hairy 😬. Happy to help out if there's any way I can do that!

wardpeet:
Thank you for the PR, i'm closing this for now because it gotten a bit stale. Feel free to re-open.

PR #16448, opened 2026-05-11
fix(aiv5-adapter): pass through OpenAI Files API file IDs without corrupting them into invalid base64 data URIs

### LINKED ISSUE #16408: [BUG]  OpenAI Files API file IDs corrupted into invalid base64 data URIs in AIV5Adapter
### Describe the Bug

When an input processor returns a file part with `data: "file-abc123"` (an OpenAI Files API file ID), `AIV5Adapter.toUIMessage` silently corrupts it into an invalid base64 data URI, causing a runtime crash.

  Root cause: `categorizeFileData("file-abc123")` calls `new URL("file-abc123")` which throws because the string has no URL scheme. The code then falls through to `createDataUri("file-abc123", mimeType)`, producing `"data:application/pdf;base64,file-abc123"`. This is not valid base64 — the colon in
  `"file-abc123"` makes it unparseable. `downloadAssetsFromMessages` subsequently calls `fetch("data:...")` and throws:
  > MastraError: Failed to download asset – failed to fetch the data URL

  Note: `@ai-sdk/openai` already handles `"file-"` prefixed strings correctly by sending `{ file_id: "file-abc123" }` to the API. The bug is that the string never reaches the provider intact — `AIV5Adapter` corrupts it first.

related PR request is : https://github.com/mastra-ai/mastra/pull/16352

### Steps To Reproduce

Steps To Reproduce:

  1. Configure an agent with an input processor that uploads a file to the OpenAI Files API and returns `data: "file-<id>"` on the file part (e.g. using a processor that calls `client.files.create()`).
  2. Send a message to the agent with a file attachment.
  3. Observe the error: `MastraError: Failed to download asset – failed to fetch the data URL`

https://github.com/erkanonatudemy/mastra-openai-fileid-repro

### Expected Behavior

File parts with `data` values starting with `"file-"` should be passed through to `@ai-sdk/openai` as-is so the provider can send `{ file_id: "file-abc123" }` to the OpenAI API, using the Files API reference rather than re-encoding the string as base64.

### Environment Information

```block
provider: OpenAI

sample Postman request :

curl --location 'http://localhost:8080/api/agents/in-course-assistant/generate' \
--header 'Content-Type: application/json' \
--data '{
      "messages": [{
        "role": "user",
        "content": [
          {
              "type": "file",
              "mediaType": "application/pdf",
              "data": "file-XkZk6RV6jeACpVewBphWEX" (firstly upload file and  set the fileId to openAI-file-ID with output of custom our processor)
            },
          { "type": "text", "text": "summarize this file ?" }
        ]
      }],
      "memory": { "thread": "test-scenario1122", "resource": "user-1122" }
    }


 '
```

### Verification

nancysangani:
When a file part's `data` value is an OpenAI Files API file ID (e.g. `"file-abc123"`), `categorizeFileData` in `image-utils.ts` misclassified it as raw data because `new URL("file-abc123")` throws (no scheme). This caused the code to fall through to `createDataUri`, producing `"data:application/pdf;base64,file-abc123"` — invalid base64 — which crashed `downloadAssetsFromMessages` with `MastraError: Failed to download asset`.

**Fix:** added an early-return guard in `categorizeFileData` for strings starting with `"file-"`, returning `type: 'url'`. The existing `url` branch in `AIV5Adapter.toUIMessage` already passes `part.data` through untouched, so `@ai-sdk/openai` receives the file ID intact and forwards it as `{ file_id: "file-abc123" }` to the OpenAI API.

**Files changed:**
- `packages/core/src/agent/message-list/prompt/image-utils.ts` — guard in `categorizeFileData` before `isValidUrl`
- `packages/core/src/agent/message-list/adapters/AIV5Adapter-file-ui-part.test.ts` — test case covering file ID passthrough

Fixes #16408 

## ELI5

OpenAI file IDs such as `file-abc123` are references, not file contents. This PR keeps these IDs unchanged so OpenAI can retrieve the files correctly.

## Overview

Fixes incorrect conversion of OpenAI Files API IDs into invalid `data:` URIs. The IDs now remain unchanged across v1, v4, and v5 message and prompt conversion paths.

## Changes

- Recognize `file-` values as provider file IDs.
- Preserve provider file IDs and MIME types during attachment and prompt conversion.
- Restore provider file IDs after SDK conversion when required.
- Use `application/octet-stream` when no MIME type is available.
- Add regression tests for direct conversion, round trips, attachments, prompts, and split assistant/tool messages.
- Add a `@mastra/core` patch changeset.

## Impact

- Passes file IDs unchanged to `@ai-sdk/openai`, which can send them as `{ file_id: "<id>" }`.
- Prevents invalid base64 data URIs and related asset download failures.


intojhanurag (approved):

intojhanurag:
@coderabbitai bro how you forgot to drop your review

PR #16449, opened 2026-05-11
fix(core): normalize `toModelOutput` media types for AI SDK provider compatibility

CalebBarnes:
## Summary

Fixes tools with `toModelOutput` that return image content (like screenshot tools) not being visible to the model.

## Problem

`toModelOutput` returns `{ type: 'media', mediaType: 'image/png', data: ... }` which matches the `@ai-sdk/provider` TypeScript types (`LanguageModelV2ToolResultOutput`). However, AI SDK providers at runtime expect `type: 'image-data'` — and silently drop any unrecognized content part type.

AI SDK normally normalizes `media` → `image-data` internally in `mapToolResultOutput`, but Mastra calls `toModelOutput` directly via `getProviderMetadataWithModelOutput` and stores the result in `providerMetadata.mastra.modelOutput`, bypassing that normalization step.

### Affected providers

| Provider | `type: 'image-data'` (normalized) | `type: 'media'` (unnormalized) |
|---|---|---|
| **Anthropic** | Native `image` block | Silently dropped |
| **Google/Gemini** | `inlineData` | Silently dropped |
| **OpenAI (Responses API)** | `input_image` with data URI | Dropped + warning |
| **OpenAI (Chat Completions)** | `input_image` with data URI | Dropped + warning |
| **OpenAI-compatible** | `JSON.stringify` (no image support) | Same — no difference |

Any tool using `toModelOutput` to return image/file content was affected across all major providers.

## Fix

Add `normalizeModelOutput()` in `llm-mapping-step.ts` that runs immediately after `tool.toModelOutput()` and converts:
- `type: 'media'` with image MIME → `type: 'image-data'`
- `type: 'media'` with non-image MIME → `type: 'file-data'`

This mirrors the same mapping AI SDK does in its `mapToolResultOutput` function.

## Test Plan

- Existing `llm-mapping-step.test.ts` tests pass (26/26)
- Typecheck clean
- Manually verified: screenshot tool images now visible to Claude models via Anthropic provider

## ELI5

Tools were giving images/files to the AI in a format the AI providers didn't understand, so those images/files were dropped. This change converts the tool output into the format providers expect so images/files actually reach the model.

## Overview

Fixes a compatibility bug where tool results returned by toModelOutput() using Mastra's internal media shape (e.g., { type: 'media', mediaType: 'image/png', data: ... }) were not recognized by AI SDK providers and were silently dropped. The PR normalizes tool outputs immediately after toModelOutput() so they match provider-expected types (image-data/file-data), preserving media parts across Anthropic, Google/Gemini, OpenAI (Responses/Chat), and OpenAI-compatible providers.

## Changes

packages/core/src/loop/workflows/agentic-execution/llm-mapping-step.ts
- Add normalizeModelOutput() helper and run it right after tool.toModelOutput(toolCall.result).
- convert any part with type: 'media' into:
  - type: 'image-data' when mediaType starts with 'image/'
  - type: 'file-data' for other media types
- Preserve data and mediaType fields and avoid unsafe runtime assertions by narrowing at runtime.
- Persist the normalized value into providerMetadata.mastra.modelOutput so Mastra no longer bypasses AI SDK normalization.

packages/core/src/loop/workflows/agentic-execution/llm-mapping-step.test.ts
- Add unit tests:
  - Normalizes media parts from toModelOutput into image-data and file-data while preserving non-media parts.
  - Handles malformed content.value entries robustly (null/undefined/non-objects) without throwing, ensuring only valid media entries are normalized.

.changeset/fifty-doors-go.md
- Add changeset metadata for a patch release of @mastra/core describing the media normalization fix.

## Testing

- Existing tests pass (llm-mapping-step tests: 26/26).
- TypeScript typecheck clean.
- Manual verification: screenshot tool images are visible to Claude via the Anthropic provider.

[Review Change Stack](https://app.coderabbit.ai/change-stack/mastra-ai/mastra/pull/16449)

### DIFF
--- a/packages/core/src/loop/workflows/agentic-execution/llm-mapping-step.ts
+++ b/packages/core/src/loop/workflows/agentic-execution/llm-mapping-step.ts
@@ -16,31 +16,6 @@ import { createStep } from '../../../workflows';
 import type { OuterLLMRun } from '../../types';
 import { llmIterationOutputSchema, toolCallOutputSchema } from '../schema';
 
-/**
- * Normalize modelOutput from toModelOutput() so that `type: 'media'` parts are
- * converted to `type: 'image-data'` or `type: 'file-data'` as the AI SDK
- * provider layer expects. AI SDK performs this same normalization internally in
- * `mapToolResultOutput`, but Mastra calls toModelOutput directly and stores
- * the result, so we need to replicate this mapping.
- */
-function normalizeModelOutput(output: unknown): unknown {
-  if (output == null || typeof output !== 'object') return output;
-
-  const obj = output as Record<string, unknown>;
-  if (obj.type !== 'content' || !Array.isArray(obj.value)) return output;
-
-  return {
-    ...obj,
-    value: (obj.value as Array<Record<string, unknown>>).map(item => {
-      if (item.type !== 'media') return item;
-      if (typeof item.mediaType === 'string' && item.mediaType.startsWith('image/')) {
-        return { type: 'image-data', data: item.data, mediaType: item.mediaType };
-      }
-      return { type: 'file-data', data: item.data, mediaType: item.mediaType };
-    }),
-  };
-}
-
 export function createLLMMappingStep<Tools extends ToolSet = ToolSet, OUTPUT = undefined>(
   { models, _internal, ...rest }: OuterLLMRun<Tools, OUTPUT>,
   llmExecutionStep: any,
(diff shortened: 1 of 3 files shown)


packages/core/src/loop/workflows/agentic-execution/llm-mapping-step.ts
```diff
@@ -16,6 +16,35 @@ import { createStep } from '../../../workflows';
...
+function normalizeModelOutput(output: unknown): unknown {
+  if (
+    output == null ||
+    typeof output !== 'object' ||
+    (output as any).type !== 'content' ||
+    !Array.isArray((output as any).value)
+  ) {
+    return output;
+  }
+  const value = (output as any).value as Array<Record<string, unknown>>;
```
DanielSLew: The types seem funky here, is output always going to have a .value and is that always an error of items with types etc?

CalebBarnes: The types look ugly here but we do check that output.value is an array right above this. Will see if I can clean up the types there though.

wardpeet (approved):

PR #16472, opened 2026-05-12
fix(workflows): use getWorkflowById in event processor for scheduled workflow lookup

### LINKED ISSUE #16471: bug: scheduled workflow fails with "not found" when workflow id differs from registration key
## Bug

When a scheduled workflow has an `id` that differs from its registration key (the property name in the `workflows` object passed to `new Mastra()`), the cron fires but immediately fails with:

```
MastraError: Workflow with ID <workflow-id> not found
```

## Reproduction

The docs show this pattern as valid:

```ts
export const dailyReport = createWorkflow({
  id: 'daily-report',          // kebab-case id
  inputSchema: z.object({ userId: z.string() }),
  schedule: { cron: '0 9 * * *', inputData: { userId: 'system' } },
})
  .then(sendReport)
  .commit()

const mastra = new Mastra({
  workflows: { dailyReport },  // registration key = 'dailyReport'
})
```

When the cron fires, it throws:
```
Workflow with ID daily-report not found
Available: dailyReport, ...
```

## Root cause

`WorkflowScheduler` publishes `workflow.start` events with `workflowId` set to the workflow's `.id` property (e.g. `'daily-report'`).

`WorkflowEventProcessor.process()` then looks up the workflow using `mastra.getWorkflow(workflowData.workflowId)`, which performs a **registration key** lookup (`this.#workflows[id]`). Since `'daily-report' !== 'dailyReport'`, the lookup fails.

The correct method is `mastra.getWorkflowById()`, which searches by `.id` property first (with a fallback to registration key):

```ts
// packages/core/src/workflows/evented/workflow-event-processor/index.ts

// Line ~1888 (user-event handler):
workflow: this.mastra.getWorkflow(workflowData.workflowId)  // ❌
workflow: this.mastra.getWorkflowById(workflowData.workflowId)  // ✅

// Line ~1905 (main process handler):
workflow = this.mastra.getWorkflow(workflowData.workflowId)  // ❌
workflow = this.mastra.getWorkflowById(workflowData.workflowId)  // ✅
```

## Workaround

Until fixed, the workflow `id` must exactly equal the registration key:

```ts
export const dailyReport = createWorkflow({
  id: 'dailyReport',  // must match registration key exactly
  ...
})
const mastra = new Mastra({ workflows: { dailyReport } })
```

## Fix

PR with the two-line fix: https://github.com/mastra-ai/mastra/pull/5740 _(link to be updated)_

Affects: `@mastra/core@1.32.1` and current `main`.

gourine:
Fixes #16471

## Problem

`WorkflowEventProcessor.process()` calls `mastra.getWorkflow(workflowData.workflowId)` to look up the workflow when an event fires. `getWorkflow()` looks up by **registration key** (the property name in the `workflows: {}` object passed to `new Mastra()`).

The scheduler publishes `workflow.start` events with `workflowId` set to the workflow's `.id` field. When `id !== registrationKey` (e.g. docs example: `id: 'daily-report'` registered as `{ dailyReport }`), the lookup fails:

```
MastraError: Workflow with ID daily-report not found
Available: dailyReport, ...
```

This makes the docs example non-functional — the docs show `id: 'daily-report'` with shorthand registration `{ dailyReport }`, but the scheduler can't find the workflow.

## Fix

Replace both `getWorkflow()` calls with `getWorkflowById()`, which searches by `.id` property first and falls back to registration key — exactly what's needed here.

```diff
- workflow: this.mastra.getWorkflow(workflowData.workflowId),
+ workflow: this.mastra.getWorkflowById(workflowData.workflowId),

- workflow = this.mastra.getWorkflow(workflowData.workflowId);
+ workflow = this.mastra.getWorkflowById(workflowData.workflowId);
```

Two occurrences in `packages/core/src/workflows/evented/workflow-event-processor/index.ts`.

## ELI5

When you schedule a workflow, the scheduler tells the event processor which workflow to run using the workflow's unique ID. But the event processor was looking it up using a different name (the registration key), so it couldn't find the workflow and would fail. This PR fixes it by making the lookup use the correct ID instead.

## Problem

The `WorkflowEventProcessor` uses `mastra.getWorkflow()` to resolve workflows when events are processed. However, `getWorkflow()` looks up workflows by their registration key (the property name in the workflows object passed to `new Mastra()`), while the scheduler publishes `workflow.start` events using the workflow's `.id` property. When these differ (e.g., a workflow with `id: 'daily-report'` registered as `{ dailyReport }`), the lookup fails with "Workflow with ID ... not found", causing scheduled workflows to fail immediately.

## Solution

Replaced two `getWorkflow()` calls with `getWorkflowById()` in `packages/core/src/workflows/evented/workflow-event-processor/index.ts`:

- When handling `workflow.user-event.*` events for `processWorkflowWaitForEvent`
- When selecting the workflow for non-internal, non-nested execution

`getWorkflowById()` searches by the workflow `.id` first and falls back to the registration key, addressing the mismatch and restoring functionality for workflows where `id !== registration key`.

## Changes

- **File modified**: `packages/core/src/workflows/evented/workflow-event-processor/index.ts`
- **Lines changed**: +2/-2
- **Scope**: Low impact, targeted fix with no changes to control flow, error handling, or event publishing logic

[Review Change Stack](https://app.coderabbit.ai/change-stack/mastra-ai/mastra/pull/16472)


abhiaiyer91:
Thanks for this but addressed this and another issue in https://github.com/mastra-ai/mastra/pull/16510

Issue #16292, opened 2026-05-06
[BUG] @mastra/client-js: SDK ↔ server contract drift (MemoryThread.update missing agentId, plus systematic audit)


roaminro:
# SDK ↔ server contract drift in `@mastra/client-js`

## Summary

`@mastra/client-js` exposes typed methods that produce HTTP requests rejected by the server's Zod route schemas. The SDK type system does not catch this because resource classes and parameter types in `client-sdks/client-js/src/types.ts` are hand-rolled and diverge from the schemas declared in `SERVER_ROUTES` (`packages/server/src/server/server-adapter/routes/index.ts`).

Generated route types already exist at `client-sdks/client-js/src/route-types.generated.ts` (produced by `packages/server/scripts/generate-route-types.ts`), but the SDK does not consume them.

This issue tracks both a **confirmed user-reported bug** and a **systematic audit** that uncovered additional drift of the same class.

---

## Confirmed bug (user-reported)

```ts
const client = new MastraClient({ baseUrl });
const thread = client.getMemoryThread('thread-1');
await thread.update({ title: 'renamed' }); // ❌ rejected by server
```

The server's route schema for `PATCH /memory/threads/:threadId` requires `agentId` as a query param. The SDK's `MemoryThread` constructor declares `agentId` as `private agentId?: string` (optional), so TypeScript permits constructing a thread without it. When `update` / `delete` / `deleteMessages` is called, no `?agentId=...` is appended and the request fails Zod validation server-side.

**Files**
- SDK: `client-sdks/client-js/src/resources/memory-thread.ts`
- Server route: `packages/server/src/server/server-adapter/routes/index.ts` (memory thread routes)

**Expected**: either the SDK requires `agentId` at the type level, or the server route does not require it.

---

## Systematic audit (this issue's broader scope)

A contract audit script was written to enumerate every public SDK method, synthesize valid inputs from the SDK's TypeScript types, capture the resulting `fetch` request, match it to a `SERVER_ROUTES` entry, and run the server's Zod `bodySchema` / `querySchema` against the captured request.

### Audit script

The full script (713 lines) is inlined at the bottom of this issue under "Audit script source".

To run it, save it to `client-sdks/client-js/scripts/audit-contract.ts`, then:

```bash
pnpm install
pnpm --filter @mastra/client-js add -D ts-morph tsx
pnpm --filter @mastra/client-js add -D @mastra/server
[...]

PR #16489, opened 2026-05-12
fix(core,server): coerce limit on metric discovery query schemas

roaminro:
## Summary

The `GET /api/observability/discovery/metric-names` and `GET /api/observability/discovery/metric-label-values` routes use `getMetricNamesArgsSchema` / `getMetricLabelValuesArgsSchema` as the query-param schema, and both declared `limit` as `z.number()`. URL query values arrive as strings, so any caller passing `?limit=10` got rejected with:

> `Invalid input: expected number, received string`

Switched `limit` to `z.coerce.number().int().min(1).optional()` so HTTP callers no longer have to pre-parse numeric values, matching the pattern used by `paginationArgsSchema` and other query schemas.

Surfaced by the audit work in #16292.

## Test plan

- Unit: 2 new tests in `metrics.test.ts` confirming `getMetricNamesArgsSchema.parse({ limit: '25' })` and `getMetricLabelValuesArgsSchema.parse({ limit: '50' })` coerce to numbers (33/33 in file pass).
- E2E: built monorepo packages, ran `mastra dev` against a minimal Mastra app and verified the routes via HTTP:
  - `GET /observability/discovery/metric-names?limit=10` → 500 from storage layer (`This storage provider does not support metric name discovery`), i.e. **passes schema validation** (previously: 400 "expected number, received string").
  - `GET /observability/discovery/metric-label-values?metricName=test&labelKey=agent&limit=25` → also passes schema validation.
  - `GET /observability/discovery/metric-names?limit=-5` → still rejected with 400 (min(1) enforced).
  - `GET /observability/discovery/metric-names?limit=abc` → still rejected with 400 (coerce fails to NaN).
  - `GET /observability/discovery/metric-names?limit=100` → passes schema validation.

## ELI5 release note

If you called the metric-discovery routes with `?limit=10`, the server rejected it because the schema required a true number and URL params are always strings. The schema now coerces strings to numbers (and still rejects bogus values like `-5` or `abc`), so the routes accept normal HTTP calls.

## Summary by CodeRabbit

* **Bug Fixes**
  * Metric-discovery endpoints now accept numeric-looking URL query values like "?limit=10" instead of rejecting them.

* **Tests**
  * Added tests to verify query parameter coercion for metric discovery limits.

* **Chores**
  * Added release metadata documenting the fix.

[Review Change Stack](https://app.coderabbit.ai/change-stack/mastra-ai/mastra/pull/16489)


roaminro:
Addressed in 1fc0c8090b — rewrote the changeset to describe the user-visible behavior (endpoints accept `limit` as a URL param) rather than the schema implementation detail. cc @coderabbitai

abhiaiyer91 (approved):

PR #16532, opened 2026-05-13
feat(playground): auto-title chat threads from first user message

damien-schneider:
## Summary

Replaces the date fallback label in the chat-thread sidebar with the first user message when no other title is available. Hierarchy is now: Observational Memory title > auto-title from first message > date (last resort).

- Extract `isDefaultThreadName` into a shared util (`packages/playground/src/domains/agents/utils/thread-title.ts`) and add `deriveThreadTitleFromMessage` (collapse whitespace, truncate to 80 chars with ellipsis).
- In `MastraRuntimeProvider.onNew`, snapshot the prior message count before send. If it's the first user message in a memory-enabled thread, fire-and-forget a PATCH to `thread.title` after the stream resolves.
- Only applies when the current title is empty or matches the default `New Thread <ISO>` pattern, so OM-suggested titles (which write directly to `thread.title` via `observation-strategies/sync.ts`) are preserved. OM, if active, will still overwrite the auto-title with its smarter summary on a later cycle.
- Triggers `refreshThreadList` on successful update so the sidebar reflects the new title immediately.

Internal-only package (`@internal/playground`), no changeset needed.

## Test plan

- [ ] Memory disabled: thread list still shows date fallback for default-named threads.
- [ ] Memory enabled, OM disabled: send first message in a new chat → sidebar updates from `May 11 at 5:58 PM`-style date to the truncated first message.
- [ ] Memory enabled, OM enabled: auto-title appears first, then gets replaced by OM-suggested title once the observation cycle runs.
- [ ] Existing thread with non-default title: title is not overwritten.
- [ ] Long / multi-line input: title collapses whitespace, truncates at 80 chars with `…`.
- [ ] Send failure / network error: no title update, no UI regression.


wardpeet:
We had this in before, you get a lot of message titles saying hey, hi, what's up so we shoudl not do this

Issue #7111, opened 2025-08-28
[FEATURE] Allow Agent instructions to use a full CoreSystemMessage


steven-range:
### Problem Statement

We use Mastra agents heavily in our code as part of several workflows.  One drawback of the agents is there's not a great way to signal to Anthropic to cache the system prompt given by the `instructions` property, since it's just a `string` and it's constructed into a generic `CoreSystemMessage`.  We have several system prompts where we'd like to use prompt caching because they're long and complex enough to justify doing so for speed and cost reasons.  Our current solution is adding a nearly empty message to the `defaultStreamContext` or `defaultGenerateContext`, and having the ability to cache would allow us to simplify that configuration.

### Proposed Solution

I would like the `instructions` property to accept a `string | CoreSystemMessage | CoreSystemMessage[]` type.  The `CoreSystemMessage[]` option is not mandatory for my use case, but I think it is a reasonable thing to include while touching this particular section of code to allow composition of instructions for an agent.

### Component

Agents

### Alternatives Considered

We are currently using a hacky workaround because it happens to append these messages after the `CoreSystemMessage` created from the `instructions`.  Note that we'd have to repeat this for `defaultGenerateOptions` if we needed the Agent to operate in both modes.

```typescript
return new Agent({
      name: `myAgent`,
      instructions,
      defaultStreamOptions: {
        context: [
          {
            role: 'system',
            content: '<></>',
            providerOptions: ANTHROPIC_1H_CACHE_OPTIONS,
          },
        ],
      },
      model,
    });
```

### Example Use Case

```typescript
new Agent({
      name: `myAgent`,
      instructions: [
          {
            role: 'system',
            content: MY_AGENT_SYSTEM_PROMPT,
            providerOptions: ANTHROPIC_1H_CACHE_OPTIONS,
          },
        ],
      model,
    })
```

### Additional Context

I have a PR open that does this, according to unit tests and local testing: https://github.com/mastra-ai/mastra/pull/6845.  Let me know if it would be helpful to rebase on main, since I know the repo is moving quickly.

### Verification

- [x] I have searched the existing issues to make sure this is not a duplicate
- [x] I have provided sufficient context for the team to understand the request

abhiaiyer91:
Thanks for this @steven-range that seems reasonable! PRs are welcome!

steven-range:
@abhiaiyer91 https://github.com/mastra-ai/mastra/pull/6845 is ready, barring some merge conflicts and review by Mastra team.

[...]

PR #6845, opened 2025-08-20
Instructions core message

steven-range:
## Description

Updates the Agent's instructions parameter to accept CoreSystemMessage and CoreSystemMessage[] types alongside strings, enabling providerOptions on `instructions` without requiring workarounds via the `context` parameter (which does accept full CoreMessage instances).

Key changes:
- Create InstructionsInput type that allows `string | CoreSystemMessage | CoreSystemMessage[]`
- Add comprehensive test coverage for single messages, arrays, and dynamic instructions
- Maintain backward compatibility with existing string instructions
- Ensure voice integration properly converts instructions to text
- Fix message list building to use proper instruction resolution

## Related Issue(s)

https://github.com/mastra-ai/mastra/issues/7111

[This Discord thread](https://discord.com/channels/1309558646228779139/1404828771902554142) is what prompted it.

## Type of Change

- [ ] Bug fix (non-breaking change that fixes an issue)
- [x] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [x] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [x] Test update

## Checklist

- [x] I have made corresponding changes to the documentation (if applicable)
- [x] I have added tests that prove my fix is effective or that my feature works

PR #16545, opened 2026-05-13
fix(durable): handle object form of instructions in preparation.ts

### LINKED ISSUE #16514: [BUG] DurableAgent silently drops Agent.instructions when given as a single CoreSystemMessage object
### Describe the Bug

`packages/core/src/agent/durable/preparation.ts` (lines 154-164) only handles two of the four shapes that `AgentInstructions = SystemMessage` can take (`string` and arrays). When `Agent.instructions` resolves to a single `CoreSystemMessage` / `SystemModelMessage` (the object form — needed to attach `providerOptions` such as Anthropic's `cacheControl: ephemeral` for prompt caching, see #7111 / #6845), the durable preparation falls through every branch and the system prompt is **silently dropped** before the LLM request is built. Only the workspace's system instructions (appended just below at line 168) end up in the outbound `messages[]`.

The non-durable `Agent` path is unaffected — its `prepare-stream` handles every `SystemMessage` shape via `MessageList`.

Affected since the durable architecture landed in #12557. Confirmed against `@mastra/core@1.32.1`.

### Steps To Reproduce

1. Build an Agent with `instructions` as a `CoreSystemMessage` object (the shape supported since #6845):
   ```ts
   const inner = new Agent({
     name: 'Test',
     instructions: {
       role: 'system' as const,
       content: 'You are a strict JSON-only assistant. Respond only with {"ok":true}.',
       providerOptions: { anthropic: { cacheControl: { type: 'ephemeral' as const } } },
     },
     model: /* any provider */,
   });
   ```
2. Wrap with `createDurableAgent({ agent: inner })`.
3. Call `durable.stream('hello')`.
4. Inspect the outbound LLM request body (e.g. via Langfuse, or by logging the provider call). The system message from step 1 is absent. The model returns prose instead of the instructed JSON.

The same agent **without** `createDurableAgent` sends the system prompt correctly.

https://github.com/tmdeim-git/mastra-durable-instructions-drop

### Expected Behavior

Calling `durable.stream(...)` on an agent whose `instructions` is a single `CoreSystemMessage` (or `SystemModelMessage`) should add it to the outbound `messages[]` exactly as the non-durable path does — preserving `content` and `providerOptions`. The full `SystemMessage` union (`string | string[] | CoreSystemMessage | SystemModelMessage | CoreSystemMessage[] | SystemModelMessage[]`) declared in `packages/core/src/llm/index.ts` should be honored.

### Suggested fix (one line)

```diff
       for (const inst of instructions) {
         messageList.addSystem(inst);
       }
+    } else {
+      messageList.addSystem(instructions);
     }
   }
```

`messageList.addSystem(obj)` already accepts the object form — it is used three lines below for workspace instructions: `messageList.addSystem({ role: 'system', content: wsInstructions })`. Happy to send a PR.

### Environment Information

```block
System:
  OS: Windows 11 10.0.26200
  CPU: (20) x64 12th Gen Intel(R) Core(TM) i7-12700KF
  Memory: 984.21 MB / 31.82 GB
Binaries:
  Node: 22.17.0 - C:\nvm4w\nodejs\node.EXE
  npm: 10.9.2 - C:\nvm4w\nodejs\npm.CMD
Browsers:
  Chrome: 147.0.7727.138
  Edge: Chromium (145.0.3800.82)
  Internet Explorer: 11.0.26100.8115
npmPackages:
  @mastra/core: 1.32.1
  @mastra/memory: 1.17.5
  @biomejs/biome: 2.4.13
  drizzle-orm: 0.45.2
  esbuild: 0.27.7
  patch-package: 8.0.1
  tsx: 4.20.5
  turbo: 2.9.6
  typescript: 5.9.3
  vite: 8.0.10
  zod: 4.3.6
```

### Verification

guangyang1206:
Fixes #16514 — DurableAgent silently dropped Agent.instructions when given as a single CoreSystemMessage object (the form needed for providerOptions such as Anthropic cacheControl).

Root cause: preparation.ts only handled string and Array forms, not the object form. Added an else branch to pass the object directly to messageList.addSystem().

### Problem

`packages/core/src/agent/durable/preparation.ts` (lines 154-164) only handles two of the four shapes that `AgentInstructions = SystemMessage` can take (`string` and arrays). When `Agent.instructions` resolves to a single `CoreSystemMessage` / `SystemModelMessage` (the object form — needed to attach `providerOptions` such as Anthropic's `cacheControl: ephemeral` for prompt caching, see #7111 / #6845), the durable preparation falls through every branch and the system prompt is **silently dropped** before the LLM request is built.

The non-durable `Agent` path is unaffected — its `prepare-stream` handles every `SystemMessage` shape via `MessageList`.

Affected since the durable architecture landed in #12557. Confirmed against `@mastra/core@1.32.1`.

### Solution

Add an `else` branch to handle the object form:

```diff
   } else if (Array.isArray(instructions)) {
     for (const inst of instructions) {
       messageList.addSystem(inst);
     }
+  } else {
+    // Handle object form (CoreSystemMessage / SystemModelMessage) for providerOptions
+    messageList.addSystem(instructions);
   }
```

`messageList.addSystem(obj)` already accepts the object form — it is used three lines below for workspace instructions: `messageList.addSystem({ role: 'system', content: wsInstructions })`.

### Reproduction

1. Build an Agent with `instructions` as a `CoreSystemMessage` object:
   ```ts
   const inner = new Agent({
     name: 'Test',
     instructions: {
       role: 'system' as const,
       content: 'You are a strict JSON-only assistant. Respond only with {"ok":true}.',
       providerOptions: { anthropic: { cacheControl: { type: 'ephemeral' as const } } },
     },
     model: /* any provider */,
   });
   ```
2. Wrap with `createDurableAgent({ agent: inner })`.
3. Call `durable.stream('hello')`.
4. Inspect the outbound LLM request body — the system message from step 1 is absent.

The same agent **without** `createDurableAgent` sends the system prompt correctly.

### Verification

- [x] `messageList.addSystem()` already accepts `CoreSystemMessage | SystemModelMessage | string | string[]`
- [x] Non-durable path (direct `Agent.stream()`) is unaffected
- [x] All four `AgentInstructions` union members now handled

Fixes #16514

## ELI5
Imagine you're giving instructions to an AI assistant, and you want to attach special settings to those instructions (like how fast it should think). The code was ignoring these special instructions when they were packaged a certain way. This PR fixes it by making sure the code handles all ways of packaging those instructions.

## Overview
This PR fixes an issue where `DurableAgent` was silently dropping `Agent.instructions` when provided as a CoreSystemMessage or SystemModelMessage object (which can attach provider-specific options like Anthropic's cacheControl).

## Changes
**File: `packages/core/src/agent/durable/preparation.ts`**

The `prepareForDurableExecution` function now handles the object form of instructions. Previously, it only explicitly handled string and string[] forms—if instructions came in any other format (like a CoreSystemMessage object), they were silently omitted from the prepared request.

The fix adds logic to catch this case and pass it to `messageList.addSystem()`, which already supports CoreSystemMessage, SystemModelMessage, strings, and string arrays.

## Validation
- `messageList.addSystem` already accepts the object form
- The non-durable execution path is unaffected
- All union members of `AgentInstructions` are now handled

[Review Change Stack](https://app.coderabbit.ai/change-stack/mastra-ai/mastra/pull/16545)


TheIsrael1:
Thanks for the PR! i'm closing this for https://github.com/mastra-ai/mastra/pull/16599

PR #16556, opened 2026-05-13
fix: replay transformed recording hashes

TylerBarnes:
It fixes exact LLM replay when recordings use request transforms. Loaded recordings now recompute their in-memory hash after applying transformRequest, so exact replay compares transformed live requests with transformed saved requests.

This does not loosen matching or rewrite recording files; if the transformed request still differs, exact replay still fails.

Verified with the llm-recorder test/build, rebuilt test-utils, and the focused memory Thread Cloning integration test.

## ELI5

When replaying saved LLM recordings that have been modified (transformed) before use, the system now applies the same modifications to the saved recording before comparing it with the live request. This ensures that the "before" and "after" versions are compared fairly—just like making sure two recipes go through the same cooking process before checking if the results match.

## Changes

The PR updates LLM recording replay behavior to correctly handle request transforms in `exactMatch` mode:

**llm-recorder.ts**: When loading an existing recording file, the code now applies `options.transformRequest` (if provided) to each stored recording's request and recomputes the recording hash based on the transformed values. Previously, loaded recordings were used as-is without per-recording hash re-computation for transforms.

**llm-recorder.test.ts**: Added a new test case verifying that in `exactMatch` replay mode, `setupLLMRecording` uses the transformed request payload when looking up an existing recording hash. The test creates a temporary recordings directory with a stale recording containing a timestamp in the request body, sets `LLM_TEST_MODE` to `replay`, performs a fetch with a different timestamp, and verifies the replayed response matches the stored recording without any "No exact match" warnings.

## Impact

- Enables exact replay matching for recordings that use request transforms
- Preserves dynamic-field normalization stability
- Does not weaken exact matching behavior
- Does not mutate recording files
- If a transformed request still differs between live and saved, exact replay will continue to fail as expected

[Review Change Stack](https://app.coderabbit.ai/change-stack/mastra-ai/mastra/pull/16556)

### DIFF
--- a/packages/_llm-recorder/src/llm-recorder.test.ts
+++ b/packages/_llm-recorder/src/llm-recorder.test.ts
@@ -121,6 +121,78 @@ describe('transformRequest', () => {
       recorder.stop();
     }
   });
+
+  it('uses transformed hashes for existing recordings in exact replay mode', async () => {
+    const originalMode = process.env.LLM_TEST_MODE;
+    const tempDir = fs.mkdtempSync(path.join(os.tmpdir(), 'llm-recorder-transform-'));
+    const name = 'transformed-hash-replay';
+    const filePath = path.join(tempDir, `${name}.json`);
+
+    fs.writeFileSync(
+      filePath,
+      JSON.stringify(
+        {
+          meta: { name, createdAt: '2026-03-26T00:00:00.000Z' },
+          recordings: [
+            {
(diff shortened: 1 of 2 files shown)


wardpeet:
the goal is to only use this for hashes TBH

wardpeet:
what error are you seeying? I'm trying to fix the broken recordings though by using exactMatch

PR #16570, opened 2026-05-13
feat(core): return media parts and metadata fallback from workspace read_file

CalebBarnes:
## Summary

Reworks how the workspace `read_file` tool returns files to the model. Previously every file was either base64-dumped (for images/PDFs) or read as text — meaning unsupported binaries (zip, exe, video, etc.) would dump useless base64 into the conversation, and images were sent as base64 strings instead of as native file parts the model can actually look at.

`read_file` now branches on file type:

1. **Media files** matching `mediaTypes` (default `'image/png', 'image/jpeg', 'image/webp', 'application/pdf'`) are returned as native `media` parts the model can directly view, mirroring how the screenshot tool returns images.
2. **Text-readable files** (`text/*`, common code/config mime types, extensionless files like `Dockerfile`/`Makefile`/`LICENSE`, unknown extensions) are returned as text content, unchanged from before.
3. **Unsupported binaries** (zip, exe, mp4, docx, etc.) now return a short metadata string (path, size, mime type) instead of base64 dumping. An explicit `encoding: 'base64' | 'hex'` is still available as an escape hatch for the rare case the agent actually needs to inspect raw bytes.

## New config

The set of mime types treated as native media is configurable per workspace via a new `mediaTypes` option on the `read_file` tool config:

```ts
import { Workspace, WORKSPACE_TOOLS } from '@mastra/core/workspace';

  filesystem,
  tools: {
    [WORKSPACE_TOOLS.FILESYSTEM.READ_FILE]: {
      // Glob patterns (default)
      mediaTypes: ['image/*', 'application/pdf'],
      // Or a custom predicate
      // mediaTypes: (mime) => mime.startsWith('image/'),
      // Or disable entirely
      // mediaTypes: false,
    },
  },
});
```

The `['image/*', 'application/pdf']` default reflects the formats universally supported as native media parts across Anthropic, OpenAI, and Gemini.

## Additional changes

Extended `fs-utils` `MIME_TYPES` with common binary extensions (archives, audio, video, office docs, fonts, executables, disk images) so they can be distinguished from extensionless text files and correctly routed to the metadata branch. `.bin`/`.dat` use a dedicated `application/x-binary` mime so they aren't conflated with the `application/octet-stream` fallback used for extensionless text files.

Tool descriptions for `read_file` were rewritten to address the model directly and to clarify that `offset`/`limit`/`showLineNumbers` only apply when reading as text.

## Test plan

- 23 unit tests in `read-file.test.ts` covering all three branches, configurable `mediaTypes`, explicit `encoding`, extensionless files, known binary extensions, `.bin`/`.dat`, and unknown extensions.
- `pnpm vitest run src/workspace` — 1646 tests pass, no type errors.
- `pnpm --filter ./packages/core check` — clean.

## Changeset

`.changeset/slow-signs-teach.md` — minor bump on `@mastra/core`.

## ELI5

When the AI reads a file, it now checks the file type: supported images and PDFs are returned as viewable media parts, text files are returned as text (with line-range support), and other binaries return a short metadata summary instead of dumping raw bytes — you can still request raw bytes explicitly via encoding.

---

## Overview

Adds media-aware, per-workspace behavior to the workspace read_file tool. By default read_file classifies files by MIME/type and:
- returns native media parts (text header + media part) for configured mediaTypes (default: PNG, JPEG, WebP, GIF, PDF) when no encoding is requested and size ≤ maxMediaBytes;
- returns text with offset/limit/showLineNumbers applied for text-like MIME types;
- returns a concise metadata-only string (path, size, mime) for unsupported binaries or oversized media unless an explicit encoding ('base64' | 'hex') is provided or mediaTypes is configured to include them.

---

## Key Changes

- packages/core/src/workspace/tools/read-file.ts
  - Implements media-aware branching and a JSON-serialisable MediaToolResult wrapper.
  - Default inline media types: ['image/png','image/jpeg','image/webp','image/gif','application/pdf'] (cross-provider-safe).
  - Adds DEFAULT_MAX_MEDIA_BYTES (10 MiB) and per-workspace maxMediaBytes to cap inline media.
  - Treats many application/* mime types as text-like; unknown/extensionless files default to text when safe.
  - Adds validate/build helpers for mediaTypes which accept glob-like arrays, a predicate function, or false to disable media parts.
  - Returns metadata-only guidance for binaries when encoding is omitted; keeps explicit encoding override to force raw bytes.
  - Adds readFileTool.toModelOutput to convert MediaToolResult into structured model output (text + media part); plain string outputs return undefined to avoid redundant modelOutput storage.
  - Updates tool description and inputSchema to document that offset/limit/showLineNumbers apply only to text reads.

- packages/core/src/workspace/tools/types.ts
  - Adds ReadFileToolConfig with mediaTypes?: string[] | ((mimeType: string) => boolean) | false and maxMediaBytes?: number.
  - Registers a dedicated config entry for the read_file tool in WorkspaceToolsConfig.

- packages/core/src/workspace/filesystem/fs-utils.ts
  - Expands MIME_TYPES mapping across many extensions (images, TIFF/HEIC/AVIF, PDF, audio/video, archives, executables, office, fonts, compiled formats).
  - Maps .bin/.dat → application/x-binary to avoid conflating binary blobs with extensionless text fallbacks.
  - Improves text-like MIME detection logic used by read_file.

- packages/core/src/workspace/tools/__tests__/read-file.test.ts
  - Adds extensive Vitest coverage (23 focused tests within the file) validating: media inlining for images/PDFs, mediaTypes disable/restriction/predicate behavior, encoding overrides, metadata-only fallbacks for binaries and oversized media, text/extensionless handling, SVG behavior relative to mediaTypes, offset/limit/showLineNumbers semantics, and toModelOutput conversion.
  - Test run reported 1,646 tests passing; TypeScript checks clean.

- .changeset/slow-signs-teach.md
  - Changeset describing the behavior, defaults, and configuration (minor bump for @mastra/core).

---

## Notes for Reviewers

- Default mediaTypes are intentionally conservative (explicit image subtypes + PDF) to avoid provider incompatibilities; consumers can set mediaTypes: ['image/*'] or provide a predicate to broaden behavior, or mediaTypes: false to disable.
- maxMediaBytes default is 10 MiB and is configurable per workspace.
- read_file now returns MediaToolResult objects for inline media; toModelOutput converts these to model content. Plain string outputs return undefined from toModelOutput to prevent duplicate storage.
- Review the expanded MIME_TYPES and the .bin/.dat → application/x-binary mapping for any environment-specific expectations.

[Review Change Stack](https://app.coderabbit.ai/change-stack/mastra-ai/mastra/pull/16570)

### DIFF
--- a/packages/core/src/workspace/tools/read-file.ts
+++ b/packages/core/src/workspace/tools/read-file.ts
@@ -6,26 +6,127 @@ import { emitWorkspaceMetadata, requireFilesystem } from './helpers';
 import { applyTokenLimit } from './output-helpers';
 import { startWorkspaceSpan } from './tracing';
 
+/**
+ * Internal marker on the tool's text result that signals to `toModelOutput`
+ * that the file should be surfaced to the model as a media part (image or
+ * binary file) rather than as plain text. We attach this on a wrapper object
+ * but only the `text` field is shown to the model (via toModelOutput); the
+ * marker is stripped before it ever reaches the model.
+ *
+ * The shape is intentionally JSON-serialisable so it round-trips through
+ * storage layers that snapshot tool results.
+ */
+type MediaToolResult = {
+  __workspaceMedia: true;
+  text: string;
+  mediaType: string;
+  data: string;
+};
+
+function isMediaToolResult(value: unknown): value is MediaToolResult {
+  return (
+    typeof value === 'object' &&
+    value !== null &&
+    (value as Record<string, unknown>).__workspaceMedia === true &&
+    typeof (value as Record<string, unknown>).text === 'string' &&
+    typeof (value as Record<string, unknown>).mediaType === 'string' &&
+    typeof (value as Record<string, unknown>).data === 'string'
+  );
+}
+
+/**
+ * Default mime types surfaced to the model as media parts. Images become
+ * image parts the model can natively see; `application/pdf` becomes a file
+ * part for providers that accept native PDFs.
+ */
+const DEFAULT_MEDIA_TYPES: string[] = ['image/*', 'application/pdf'];
+
+/**
+ * `application/*` mime types that are actually text content and safe to read
+ * as utf-8. Anything not matching `text/*` and not in this list is treated
+ * as opaque binary when there's no explicit encoding and no media match.
+ */
+const TEXT_APPLICATION_TYPES = new Set([
+  'application/json',
+  'application/javascript',
+  'application/typescript',
+  'application/xml',
+  'application/graphql',
+  'application/x-sh',
+  'application/x-yaml',
+  'application/yaml',
+  'application/ld+json',
+  'application/sql',
+]);
+
+function isTextLikeMimeType(mimeType: string | undefined): boolean {
+  if (!mimeType) return true;
+  // `application/octet-stream` is the catch-all for unknown extensions; we
+  // optimistically treat it as text so files like `.log` or `.conf` (which
+  // have no registered mime mapping) still get read as utf-8.
+  if (mimeType === 'application/octet-stream') return true;
+  if (mimeType.startsWith('text/')) return true;
+  if (TEXT_APPLICATION_TYPES.has(mimeType)) return true;
+  if (mimeType.endsWith('+json') || mimeType.endsWith('+xml')) return true;
+  return false;
+}
+
+/**
+ * Build a predicate from the `mediaTypes` config option.
+ * Supports glob patterns (e.g. `'image/*'`), custom functions, and `false`
+ * to disable media parts entirely.
+ */
+function buildMediaTypeCheck(
+  config: string[] | ((mimeType: string) => boolean) | false | undefined,
+): (mimeType: string | undefined) => boolean {
+  if (config === false) return () => false;
+  if (typeof config === 'function') {
+    return (mimeType: string | undefined) => (mimeType ? config(mimeType) : false);
+  }
+  const patterns = config ?? DEFAULT_MEDIA_TYPES;
+  return (mimeType: string | undefined) => {
+    if (!mimeType) return false;
+    return patterns.some(pattern => {
+      if (pattern === '*' || pattern === '*/*') return true;
+      if (pattern.endsWith('/*')) {
+        return mimeType.startsWith(pattern.slice(0, -1));
+      }
+      return mimeType === pattern;
+    });
+  };
+}
+
 export const readFileTool = createTool({
   id: WORKSPACE_TOOLS.FILESYSTEM.READ_FILE,
   description:
-    'Read the contents of a file from the workspace filesystem. Use offset/limit parameters to read specific line ranges for large files.',
+    'Read a file from the workspace filesystem. Text files come back as text — use offset/limit to read a line range from large files. Supported media files come back as a native file part you can view directly. Other binary files return only their metadata (path, size, mime type) since their raw contents are not useful to read.',
   inputSchema: z.object({
     path: z.string().describe('The path to the file to read (e.g., "data/config.json")'),
-    encoding: z
-      .enum(['utf-8', 'utf8', 'base64', 'hex', 'binary'])
-      .optional()
-      .describe('The encoding to use when reading the file. Defaults to utf-8 for text files.'),
     offset: z
       .number()
       .optional()
-      .describe('Line number to start reading from (1-indexed). If omitted, starts from line 1.'),
-    limit: z.number().optional().describe('Maximum number of lines to read. If omitted, reads to the end of the file.'),
+      .describe(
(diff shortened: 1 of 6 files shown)


abhiaiyer91 (approved):

packages/core/src/workspace/tools/read-file.ts
```diff
@@ -90,4 +220,19 @@ export const readFileTool = createTool({
...
+      return {
+        type: 'content',
+        value: [
+          { type: 'text', text: output.text },
+          { type: 'media', data: output.data, mediaType: output.mediaType },
+        ],
+      };
+    }
+    if (typeof output === 'string') {
+      return { type: 'text', value: output };
```
TylerBarnes: I think if you return undefined here it wont save an extra copy in the db

CalebBarnes: Ooo I assumed if I returned undefined the LLM would just see an empty tool result. So if the toModelOutput returns undefined it'll just fallback to whatever the tool result was?

CalebBarnes: Yep, confirmed it works. Traced both paths:

1. **Storage side** — `llm-mapping-step.ts:188` only writes `mastra.modelOutput` when `modelOutput != null`, so returning undefined skips the duplicate write.
2. **Read side (sending to model)** — `message-list.ts:485-499` only overrides the tool-result `output` when `storedModelOutputs.has(part.toolCallId)`. With no stored modelOutput, the raw tool result string flows through `convertToModelMessages` unchanged.
3. **tool-call-filter** — `tool-call-filter.ts:168` explicitly checks `Object.hasOwn(mastraMetadata, 'modelOutput')` and falls back to the original result if it's missing.

So undefined = "no transform, use the raw result", which is exactly what we want for plain strings. Tests cover both branches (media tool result still returns content parts; string result returns undefined). 23 read-file tests pass.

packages/core/src/workspace/tools/read-file.ts
```diff
@@ -6,26 +6,153 @@ import { emitWorkspaceMetadata, requireFilesystem } from './helpers';
...
+    typeof (value as Record<string, unknown>).data === 'string'
+  );
+}
+
+/**
+ * Default mime types surfaced to the model as media parts. Images become
+ * image parts the model can natively see; `application/pdf` becomes a file
+ * part for providers that accept native PDFs.
+ */
+const DEFAULT_MEDIA_TYPES: string[] = ['image/*', 'application/pdf'];
```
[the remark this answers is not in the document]
CalebBarnes: Addressed in `f5b363ca09`. `DEFAULT_MEDIA_TYPES` is now the explicit cross-provider-safe set:

```ts
// packages/core/src/workspace/tools/read-file.ts
const DEFAULT_MEDIA_TYPES: string[] = [
  'image/png',
  'image/jpeg',
  'image/webp',
  'image/gif',
  'application/pdf',
];
```

Added regression tests verifying SVG falls back to text by default and only inlines as media when the user explicitly broadens to `image/*` (or a custom predicate). Users keep the same escape hatch via the public `mediaTypes` config.

packages/core/src/workspace/tools/read-file.ts
```diff
@@ -39,13 +166,42 @@ export const readFileTool = createTool({
...
+      if (!encoding && shouldReturnAsMedia(stat.mimeType)) {
+        const base64 = (await filesystem.readFile(path, { encoding: 'base64' })) as string;
+        const header = `${stat.path} (${stat.size} bytes, ${stat.mimeType})`;
+        span.end({ success: true }, { bytesTransferred: stat.size });
+        return {
+          __workspaceMedia: true,
+          text: header,
+          mediaType: stat.mimeType!,
+          data: base64,
+        } satisfies MediaToolResult;
```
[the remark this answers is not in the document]
CalebBarnes: Addressed in `f5b363ca09`. Added a configurable `maxMediaBytes` cap (default 10 MiB) on the media branch:

```ts
// packages/core/src/workspace/tools/read-file.ts
if (!encoding && shouldReturnAsMedia(stat.mimeType)) {
  const maxMediaBytes = readFileConfig?.maxMediaBytes ?? DEFAULT_MAX_MEDIA_BYTES;
  if (stat.size > maxMediaBytes) {
    // fall back to metadata-only without reading the file
    return `${stat.path} (${stat.size} bytes, ${stat.mimeType}) — exceeds maxMediaBytes (${maxMediaBytes}). ...`;
  }
  const base64 = (await filesystem.readFile(path, { encoding: 'base64' })) as string;
  ...
}
```

Large files now short-circuit to metadata before `readFile` is called, so we never materialize the bytes or persist a huge base64 string through storage. 10 MiB roughly matches per-image/per-PDF limits across providers (Anthropic ~5MB images / ~32MB PDFs, OpenAI ~20MB, Gemini ~20MB) — users can raise or lower it.

PR #16585, opened 2026-05-14
fix(observability): use configured model for OpenRouter pricing lookup

### LINKED ISSUE #16584: [BUG] OpenRouter responseModel uses different word ordering than pricing-registry (Anthropic models still miss after PR #16206)
## Summary

PR #16206 (released in `@mastra/core` 1.32) fixed pricing-registry lookup for OpenRouter routes with the `vendor/model` shape (e.g. `google/gemini-2.5-flash` → `gemini-2-5-flash`). The Google case now matches.

However for **Anthropic** routes the lookup still misses, because OpenRouter returns `responseModel` with **different word order** than the configured alias — and PR #16206's variant generator only does textual transforms (slash-flatten, vendor-strip, date-strip), not word reordering.

## Versions

- `@mastra/core@1.33.1`
- `@mastra/observability@1.12.0`
- `@mastra/duckdb@1.3.1` (observability store)

## Repro

```ts
import { z } from 'zod';

const agent = new Agent({
  id: 'demo',
  name: 'demo',
  instructions: 'Return strict JSON.',
  model: 'openrouter/anthropic/claude-sonnet-4-6',  // configured alias
});

await agent.generate('What is 2+2? Return {"r":<num>}.', {
  structuredOutput: { schema: z.object({ r: z.number() }) },
});
```

Inspect the resulting model span (`span_events.attributes`):

```json
{
  "provider":      "openrouter",
  "model":         "anthropic/claude-sonnet-4-6",
  "responseModel": "anthropic/claude-4.6-sonnet-20260217"
}
```

And the resulting `metric_events` row for `mastra_model_total_input_tokens`:

```
estimatedCost = null
costUnit      = null
costMetadata  = {"error":"no_matching_model"}
```

Studio's **Model Usage & Cost** panel shows a row keyed by `anthropic/claude-4.6-sonnet-20260217` with token counts populated but the Cost column empty.

## Why PR #16206's variants don't cover this

All variants generated from `responseModel`:

- `anthropic/claude-4.6-sonnet-20260217`
- `claude-4.6-sonnet-20260217`           (vendor-strip)
- `claude-4.6-sonnet`                    (vendor-strip + date-strip)
- `anthropic-claude-4.6-sonnet-20260217` (slash-flatten)
- `anthropic-claude-4.6-sonnet`          (slash-flatten + date-strip)

The bundled pricing-registry has the OpenRouter-alias form `anthropic/claude-sonnet-4-6` (or its bare variant). No textual transform maps `claude-4.6-sonnet` ↔ `claude-sonnet-4-6` — the word order is different.

## Suggested fix

When `provider === 'openrouter'`, use the **configured `model`** (the alias the user actually called) for pricing lookup, not the `responseModel`. The configured alias is what OpenRouter accepted on input and what OpenRouter billed the user for; resolving to a downstream provider's internal id and trying to price that is a layer-mixing error.

This also has a nice side effect of being insensitive to upstream provider renames (e.g. `claude-4.6-sonnet-20260217` → `claude-4.6-sonnet-20260218` on a model snapshot bump).

## Follow-up to

- #16205 / #16281 (the original report)
- PR #16206 (partial fix)

VINODvoid:
Fixes #16584

  ## Summary

  When using OpenRouter, the `responseModel` returned by the API uses the upstream
  provider's naming convention (e.g. `anthropic/claude-4.6-sonnet-20260217`) which may
  differ in word ordering from the user-configured alias (e.g.
  `anthropic/claude-sonnet-4-6`). The variant generator in the pricing registry cannot
  bridge this mismatch, resulting in `no_matching_model` cost estimation errors.

  The fix: when `provider === 'openrouter'`, use `attrs.model` (the user-configured alias)
  for the pricing registry lookup instead of `attrs.responseModel`. This aligns with how
  OpenRouter billing works — the configured alias is what OpenRouter accepted and charged
  for.

  ## What changed

  `observability/mastra/src/metrics/auto-extract.ts` — one-line change to the model
  selection logic for the pricing lookup.

  ## Test plan

  - [ ] New test added: `should use configured model (not responseModel) for OpenRouter
  pricing lookup`
  - [ ] All 17 existing `auto-extract` tests pass
  - [ ] All 29 existing `estimator` + `auto-extract` tests pass
  - [ ] Non-OpenRouter providers continue using `responseModel` first (existing behavior
  unchanged)

## ELI5

OpenRouter sometimes “names” a model in a slightly different way than what you configured. This PR makes the cost/pricing lookup use your configured model name for OpenRouter, so cost estimation won’t fail.

---

## Problem

OpenRouter’s `responseModel` may use upstream/internal naming (including different word order), e.g. `anthropic/claude-4.6-sonnet-20260217`, which doesn’t match the user-configured alias, e.g. `anthropic/claude-sonnet-4-6`. Because the pricing-registry variant generator can’t reconcile this mismatch, pricing lookup can fail with `{"error":"no_matching_model"}`, resulting in missing/null cost estimates.

## Solution

In `emitUsageMetrics`, when `provider === 'openrouter'`, use `attrs.model` (the user-configured alias) as the pricing-registry lookup key, falling back to `attrs.responseModel` only if `attrs.model` isn’t available. For all other providers, keep the existing lookup behavior unchanged.

## Changes

- `observability/mastra/src/metrics/auto-extract.ts`
  - Updated pricing lookup model selection in `emitUsageMetrics` to be OpenRouter-specific:
    - OpenRouter: `attrs.model ?? attrs.responseModel`
    - Others: `attrs.responseModel ?? attrs.model`
  - Added inline comments/JSDoc about the alias vs. `responseModel` word-order mismatch.
- `observability/mastra/src/metrics/auto-extract.test.ts`
  - Added a unit test verifying OpenRouter uses the span’s configured `model` (not `responseModel`) for pricing lookup, derives the expected `pricing_id`, and avoids the `no_matching_model` error.
- `observability/mastra/src/metrics/__fixtures__/pricing-data-test.jsonl`
  - Added pricing data for the `openrouter-anthropic-claude-sonnet-4-6` model.
- `.changeset/better-cows-sip.md`
  - Added a patch changeset documenting the OpenRouter pricing lookup fallback behavior.

## Impact

- Fixes OpenRouter cost estimation failures caused by `responseModel` naming/word-order mismatches.
- Ensures OpenRouter pricing lookup aligns with what the user/configured model is billed against.
- Preserves prior pricing lookup behavior for non-OpenRouter providers.


observability/mastra/src/metrics/__fixtures__/pricing-data-test.jsonl
```diff
@@ -3,4 +3,5 @@
 {"i":"anthropic-claude-sonnet-4-5","p":"anthropic","m":"claude-sonnet-4-5","s":{"v":"model_pricing/v1","d":{"u":"USD","t":[{"r":{"icrt":{"c":3e-7},"icwt":{"c":0.00000375},"it":{"c":0.000003},"ot":{"c":0.000015}}}]}}}
 {"i":"openrouter-xiaomi-mimo-v2-pro","p":"openrouter","m":"xiaomi-mimo-v2-pro","s":{"v":"model_pricing/v1","d":{"u":"USD","t":[{"r":{"icrt":{"c":2e-7},"it":{"c":1e-6},"ot":{"c":3e-6}}}]}}}
 {"i":"openrouter-gpt-5-mini","p":"openrouter","m":"gpt-5-mini","s":{"v":"model_pricing/v1","d":{"u":"USD","t":[{"r":{"icrt":{"c":2.5e-8},"it":{"c":2.5e-7},"ot":{"c":2e-6}}}]}}}
+{"i":"openrouter-anthropic-claude-sonnet-4-6","p":"openrouter","m":"anthropic/claude-sonnet-4-6","s":{"v":"model_pricing/v1","d":{"u":"USD","t":[{"r":{"icrt":{"c":3e-7},"icwt":{"c":0.00000375},"it":{"c":0.000003},"ot":{"c":0.000015}}}]}}}
```
intojhanurag: Change "m":"anthropic/claude-sonnet-4-6" to "m":"claude-sonnet-4-6" to match real production data and test the actual variant-stripping path.

intojhanurag:
On auto-extract.ts line 90: The fallback is provider-agnostic (not OpenRouter-only), btw but I think it should not cause problem

TheIsrael1:
> On auto-extract.ts line 90: The fallback is provider-agnostic (not OpenRouter-only), btw but I think it should not cause problem

Yeah a newer commit was made to make the fallback provider agnostic which may be better

TheIsrael1 (approved):

Issue #16561, opened 2026-05-13
[BUG] Observational Memory loses assistant message parts on multi-step turns (UPSERT collision between step boundary save and turn.end flush)


marlesson-santana:
# [BUG] Observational Memory loses assistant message parts on multi-step turns (UPSERT collision between step boundary save and turn.end flush)

### Describe the Bug

When Observational Memory (OM) is active and an assistant turn produces multiple LLM steps under the **same `messageId`** (e.g. tool-call step followed by text step), parts are **lost from storage** due to an uncoordinated double-write:

1. `ObservationStep.prepare(stepNumber > 0)` persists the assistant message mid-turn (with tool-invocation parts from step 0), then re-adds it to `MessageList` with `source="memory"`.
2. `ObservationTurn.end()` flushes only messages in the `"response"` source set. Because the message was moved to `"memory"` at step boundary, it either:
   - **Does NOT re-persist** the message → text part from step 1 is lost (production manifestation — see #15049).
   - **Re-persists a partial view** (only the new parts from after the seal) → tool-invocation parts from step 0 are overwritten and lost (reproducible in standalone).

Both manifestations stem from the same root cause: the boundary save and the final flush don't coordinate, so whichever upserts last wins and the previous content is silently discarded.

**Affected versions:** `@mastra/memory@1.17.2` with `@mastra/core@1.28.0`.

This is likely related to #15049 (same end-user symptom: "tool call responses are lost from memory"). The fix proposed in #15050 targets `ProcessorRunner` in `@mastra/core`, but our trace points at the OM-layer specifically (`ObservationStep` / `ObservationTurn` in `@mastra/memory`). Filing separately so maintainers can correlate both layers.

### Steps To Reproduce

A standalone reproduction repository is available: **https://github.com/marlesson-santana/bug-om-text-loss** (instructions in README).

The minimal setup:

1. An agent with 2+ tools that the LLM calls in the same step (forcing `maxSteps > 1`).
2. Observational Memory enabled (any thresholds — the boundary save fires unconditionally at `stepNumber > 0`).
3. A pinned `idGenerator` that returns the same assistant `messageId` across all steps within a turn (this is what real applications do to stream a stable id to the client early in the turn).

```ts
// src/agent.ts (simplified)
import { Agent } from '@mastra/core/agent'
import { Mastra } from '@mastra/core/mastra'
import { Memory } from '@mastra/memory'
import { PostgresStore } from '@mastra/pg'
[...]

Issue #16523, opened 2026-05-13
[BUG] Observational Memory never triggers observation for single large messages (step 0 dead zone)


drichfield:
### Description

Observational Memory (OM) never triggers observation generation when a **single message** exceeds the configured `messageTokens` threshold. Observations work correctly when multiple smaller messages accumulate past the threshold.

### Reproduction

1. `npx create-mastra@latest` — scaffold a new project with the stock weather agent
2. Configure OM with a low threshold on the agent:

```ts
import { Memory } from "@mastra/memory";

const memory = new Memory({
  options: {
    observationalMemory: {
      enabled: true,
      messageTokens: 1000,  // low threshold for easy reproduction
    },
  },
});
```

3. **Working case**: Send several short messages (each < 1,000 tokens) that total > 1,000 tokens. Observations are generated as expected.
4. **Failing case**: Send a **single message** with > 1,000 tokens of text. No observation is ever generated. `pendingTokens` keeps growing but `shouldObserve` is never acted upon.

### Expected Behavior

A single message exceeding `messageTokens` should trigger observation generation, just as accumulated messages do.

### Actual Behavior

The observation is silently skipped. `pendingTokens` accumulates indefinitely.

### Root Cause

In `packages/memory/src/processors/observational-memory/observation-turn/step.ts`, the `shouldObserve` check is gated behind `stepNumber > 0` (line 188):

```ts
if (this.stepNumber > 0) {
  // ...
  if (statusSnapshot.shouldObserve && !hasIncompleteToolCalls) {
    // observation logic
  }
}
```

When a turn has no tool calls, `stepNumber` stays `0` for the entire turn. The `shouldBuffer` path also won't fire because tokens are **above** the threshold (buffering requires tokens to be below threshold but above the buffer interval). This creates a dead zone:

- `stepNumber === 0` → skips the `shouldObserve` block
- `shouldBuffer` is `false` (tokens exceed threshold, not below it)
- Result: neither observation nor buffering triggers

### Proposed Fix

Add a step 0 handler before the existing `stepNumber > 0` guard that mirrors the observation logic:

```ts
if (this.stepNumber === 0 && statusSnapshot.shouldObserve && !hasIncompleteToolCalls) {
  const preObsGeneration = this.turn.record.generationCount;
  const obsResult = await this.runThresholdObservation();
  observerExchange = obsResult.observerExchange;
  if (obsResult.succeeded) {
    observed = true;
    didThresholdCleanup = true;

[...]

Issue #16522, opened 2026-05-13
[BUG] TokenCounter does not count file part content toward Observational Memory observation threshold


drichfield:
### Describe the Bug

`TokenCounter` does not include file part content when calculating token counts for the Observational Memory (OM) observation threshold. File parts are counted based only on their metadata descriptor (`type`, `mimeType`, `filename`) — approximately 8 tokens per file — regardless of actual content size.

This means if the majority of the observation threshold is consumed by one or more attached files, OM will not trigger. The full context — including all file tokens — continues to be sent on every subsequent LLM request until the threshold is eventually met by text parts alone. This can result in massive input token overages because the model receives far more tokens than the threshold was designed to gate.

OM is already set up to write observations on file parts (they are included in the observation context). It should estimate file part tokens just like it does text parts to determine when observations are written.

**Root cause in `TokenCounter`:**

`serializeNonImageFilePartForTokenCounting()` serializes only the descriptor:
```ts
JSON.stringify({ type: "file", mimeType, filename })
// → ~50 chars → ~8 tokens
```
The actual `part.data` is never read for token estimation. This affects all file parts, not just application-specific hydration flows.

**Call chain:**
```
observe()
  → tokenCounter.countMessagesAsync(unobservedMessages)
    → countAttachmentPartAsync(part)
      → countAttachmentPartSync(part)
        → serializeNonImageFilePartForTokenCounting(part)
          → JSON.stringify({ type, mimeType, filename })  ← ignores part.data
          → estimateTokenCount(serialized)                ← ~8 tokens
    → meetsObservationThreshold({ record, unobservedTokens })
      → pendingTokens >= threshold                        ← file content invisible
```

### A note on the complexity of file token estimation

We acknowledge that estimating tokens for file parts is not straightforward in Mastra's architecture. File content can take many forms — inline base64 data, URLs to remote resources, application-specific references that get hydrated by middleware before reaching the LLM. Mastra can't be expected to resolve and decode every possible file representation to count tokens, especially when content may live behind authentication, require external API calls, or be transformed by application-layer middleware at request time.

The application layer is often the only place that actually knows the true token cost of a file part — it knows how files are stored, how they're hydrated, and what content ultimately reaches the model.

### Suggested approach: a supported client-supplied token estimate on file parts

Rather than trying to make `TokenCounter` resolve file content internally, we'd suggest a stable, documented field on file parts — something like `estimatedTokens` — that `TokenCounter` reads when present.

[...]

PR #16588, opened 2026-05-14
fix(memory): observe threshold overflow on step zero

wesleysimplicio:
﻿## Problem

Observational Memory can start a turn already above the `messageTokens` threshold, but `step 0` currently never runs the threshold observation path. That creates a dead zone where a single oversized starting message is neither buffered nor observed until a later step happens.

## Root cause

The threshold observation path lived entirely inside the `stepNumber > 0` branch, while `step 0` only initialized state and skipped the observe/cleanup flow even when `shouldObserve` was already true.

## Fix

- extract the threshold observe/cleanup flow into a small helper
- run that helper on `step 0` too when the threshold is already exceeded and there are no incomplete tool calls
- keep previous-step message persistence inside the `step > 0` branch only
- update the focused regression test to cover a single oversized message on `step 0`
- add a changeset for `@mastra/memory`

## Solution sketch

```mermaid
flowchart TD
    A[processInputStep] --> B{stepNumber == 0?}
    B -- no --> C[save previous-step messages]
    B -- yes --> D[skip previous-step save]
    C --> E{shouldObserve and no incomplete tool calls?}
    D --> E
    E -- no --> F[continue]
    E -- yes --> G[run threshold observation]
    G --> H[cleanup observed messages]
    H --> I[refresh record and status]
```

## Duplicate check

- built for issue `#16523`
- checked adjacent OM work like `#16522`, `#16561`, `#14747`; related context, but not the same step-0 threshold bug
- did not find an open PR specifically fixing the `step 0` dead zone

## Tests

Attempted:

- `pnpm vitest packages/memory/src/processors/observational-memory/__tests__/mid-loop-observation.test.ts`
- `corepack pnpm --filter @mastra/memory exec vitest run src/processors/observational-memory/__tests__/mid-loop-observation.test.ts`
- `corepack pnpm --filter @mastra/memory run test:unit -- src/processors/observational-memory/__tests__/mid-loop-observation.test.ts`

Environment note:

- the local workspace does not have dependencies installed (`node_modules` absent), so `vitest` could not run in this environment

## Risk

Medium. The change intentionally allows observation to happen earlier in a turn, so the main thing to watch is preserving the existing guard around incomplete tool calls and keeping step-0 behavior below the threshold unchanged.

## ELI5

Imagine you're a note-taker in a meeting. You have a rule: "Once you've taken too many notes, you should save the important bits and throw away the rest to keep room for new notes." This PR fixes a bug where on the very first moment of a meeting, if someone gives you a HUGE note right away, you'd just ignore the rule and keep the massive note around anyway. Now, you'll check and follow the rule even at the very start.

---

## Overview

This PR fixes a bug in Observational Memory where `step 0` fails to observe messages that already exceed the configured `messageTokens` threshold. Previously, the threshold observation logic was completely gated behind `stepNumber > 0`, leaving oversized initial messages unobserved and unbuffered until a subsequent step.

### Test Updates (`mid-loop-observation.test.ts`)

- **Removed**: Test asserting that observation does **not** trigger on step 0 when context exceeds threshold
- **Added**: 
  - New test verifying observation **does** trigger on step 0 when a single message exceeds the token threshold (e.g., a ~2500-character message against a 500-token limit)
  - New test verifying observation does **not** trigger on step 0 when messages remain below the threshold
  - Coverage now includes assertions on the observational memory record state after step 0

### Core Logic Refactor (`step.ts`)

- **Extracted helper method**: `handleThresholdObservation()` consolidates the threshold observation/cleanup pipeline (waiting for buffering, re-checking status, attempting activation, running sync observation if needed, sealing/persisting observed messages)
- **Step 0 enhancement**: Now calls `handleThresholdObservation()` when `statusSnapshot.shouldObserve` is true and no incomplete tool calls are pending, enabling synchronous observation for oversized initial messages
- **Step > 0 preservation**: Maintains previous behavior—saves messages from the prior step, then runs threshold observation only if needed
- **Guard condition**: Both step 0 and step > 0 observation check for incomplete tool calls (e.g., provider-executed tools still in `state:'call'`) before attempting observation

### Changeset

- `@mastra/memory` marked for patch release with description: "Fix observational memory so step 0 can synchronously observe when a single oversized message already exceeds the configured message token threshold"

## Risk & Testing Notes

**Risk Level**: Medium — extends observation to an earlier point in the turn lifecycle; important to preserve the incomplete tool call guard and ensure behavior remains unchanged when step 0 is below threshold.

Tests in `mid-loop-observation.test.ts` could not be executed locally due to missing `node_modules`, but the test suite covers both the happy path (observation on oversized step 0 message) and the guard condition (no observation when below threshold).

[Review Change Stack](https://app.coderabbit.ai/change-stack/mastra-ai/mastra/pull/16588)

### DIFF
--- a/.changeset/dry-trains-hunt.md
+++ b/.changeset/dry-trains-hunt.md
@@ -1,0 +1,5 @@
+---
+'@mastra/memory': patch
+---
+
+Fix observational memory so step 0 can synchronously observe when a single oversized message already exceeds the configured message token threshold.
--- a/packages/memory/src/processors/observational-memory/__tests__/mid-loop-observation.test.ts
+++ b/packages/memory/src/processors/observational-memory/__tests__/mid-loop-observation.test.ts
@@ -318,28 +318,20 @@ describe('Mid-Loop Observation', () => {
       expect(sealedAtRotate).toEqual([true]);
     });
 
-    it('should NOT trigger observation on step 0', async () => {
+    it('should trigger observation on step 0 when a single large message already exceeds the threshold', async () => {
       const requestContext = createRequestContext(threadId, resourceId);
       const state: Record<string, unknown> = {};
 
-      // Create messageList with messages that exceed threshold
       const messageList = new MessageList({
         threadId,
         resourceId,
       });
 
-      // Add messages that will exceed 500 token threshold
-      for (let i = 0; i < 20; i++) {
-        const msg = createTestMessage(
-          `Step ${i}: `.padEnd(200, 'x'),
-          i % 2 === 0 ? 'user' : 'assistant',
-          `msg-${i}`,
-          new Date(Date.now() - (20 - i) * 1000),
-        );
-        messageList.add(msg, 'memory');
-      }
+      messageList.add(
+        createTestMessage(`Single oversized message: `.padEnd(2500, 'x'), 'user', 'oversized-msg', new Date()),
+        'memory',
+      );
 
-      // Step 0: Should NOT trigger observation (only initializes record)
       await processor.processInputStep({
         messageList,
         messages: messageList.get.all.db(),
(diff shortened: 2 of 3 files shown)


abhiaiyer91:
Closing! Requiring an issue before we take contributor PRs

PR #16603, opened 2026-05-14
feat(core): add metadata support for code-defined agents

### LINKED ISSUE #16387: [FEATURE] Add metadata support for code-defined Agent config
## Problem

Stored agents can carry metadata, but code-defined agents created with `new Agent({...})` cannot. This makes it harder for callers to classify and filter agents without encoding type information into agent IDs or names.

## Proposed change

Add optional static `metadata?: Record<string, unknown>` to `AgentConfig`, expose it through `agent.getMetadata()`, and include it in existing `/agents` and `/agents/:agentId` serialized responses.

This would keep server-side metadata filtering out of scope for the first change. Callers can filter the existing agent list client-side, and API-level filtering can be considered separately.

## Example

```ts
const supportAgent = new Agent({
  id: 'support-agent',
  name: 'Support Agent',
  instructions: 'You help customers with support requests.',
  model: 'openai/gpt-5.4',
  metadata: { type: 'support' },
})

supportAgent.getMetadata()?.type
```

## Compatibility

This is an additive API change and should not affect existing agent runtime behavior.

## Implemented version - ready for a PR

https://github.com/p-meier/mastra/tree/feature/agent-metadata

intojhanurag:
## Description

Adds optional `metadata` to code-defined agents so callers can classify and filter agents without encoding type info into IDs or names. Stored agents already supported metadata; this closes the parallel gap on `new Agent({...})` and surfaces the field through the agent API responses and the client SDK types.

- Added `metadata?: DynamicArgument<Record<string, unknown>>` to `AgentConfig` and a matching `#metadata` private field on `Agent`, mirroring the existing pattern for other dynamic config fields
- Added `Agent.getMetadata({ requestContext? })` returning the configured metadata, `undefined`, or a `Promise` when metadata is defined as a function
- Propagated metadata through `formatAgent` / `formatAgentList` so `/agents` and `/agents/:agentId` return it; extended the `SerializedAgent` interface and zod schema
- Bridged stored-agent metadata into hydrated runtime agents in `EditorAgentNamespace.createAgentFromStoredConfig`
- `EditorAgentNamespace.clone()` falls back to the source agent's metadata when the caller does not pass one explicitly
- Added `metadata?` to `@mastra/client-js` `GetAgentResponse` so SDK consumers see the typed field; regenerated `route-types.generated.ts` accordingly
- Added unit tests for both static and dynamic forms, an assertion in the existing editor stored-agent test, and route-handler tests covering list and get serialization
- Added `getMetadata.mdx` reference page (linked in the sidebar) and a `metadata` entry in `agent.mdx` showing both static and dynamic usage

```ts
// Static
const supportAgent = new Agent({
  id: 'support-agent',
  name: 'Support Agent',
  instructions: 'You help customers with support requests.',
  model: 'openai/gpt-5',
  metadata: { type: 'support' },
});

supportAgent.getMetadata(); // { type: 'support' }

// Dynamic (resolved per request)
const tenantAgent = new Agent({
  id: 'tenant-agent',
  name: 'Tenant Agent',
  instructions: '...',
  model: 'openai/gpt-5',
  metadata: ({ requestContext }) => ({
    type: 'support',
    tenant: requestContext.get('tenant'),
  }),
});

await tenantAgent.getMetadata({ requestContext });
// { type: 'support', tenant: 'acme' }
```

## Related Issue(s)

Fixes #16387

## Type of Change

- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [ ] Test update

## Checklist

## ELI5

This PR adds the ability to label agents with metadata (like tags or categories) when you create them in code. You can set these labels once when creating the agent, or make them change dynamically based on who's asking. The system then shows these labels in responses so you can identify and filter agents more easily.

---

### Agent Configuration & Runtime Support
- Added an optional `metadata` field to `AgentConfig` that accepts either a static `Record<string, unknown>` or a dynamic function returning metadata based on request context (following the `DynamicArgument` pattern used for other config fields)
- Added a new `Agent.getMetadata({ requestContext? })` method that returns configured metadata, `undefined` if none is set, or a `Promise` when metadata is function-based
- Metadata is stored in a private `#metadata` field on the `Agent` class instance

### API Serialization
- Extended the `SerializedAgent` interface to include `metadata?: Record<string, unknown>`
- Updated `formatAgent()` to fetch and include metadata via `agent.getMetadata({ requestContext })`
- Updated `formatAgentList()` to attempt fetching metadata for each agent, with errors gracefully caught and logged
- Updated the corresponding zod schema (`serializedAgentSchema`) to validate the new field
- Metadata is now included in `/agents` and `/agents/:agentId` API responses

### Stored Agent Integration
- Modified `EditorAgentNamespace.createAgentFromStoredConfig()` to pass stored agent metadata into the runtime `Agent` constructor
- Updated `EditorAgentNamespace.clone()` to use the provided `metadata` when available, or fall back to the source agent's metadata via `agent.getMetadata()`

### Client SDK Updates
- Added optional `metadata` field to `GetAgentResponse` type in `@mastra/client-js`
- Regenerated `route-types.generated.ts` with the `metadata` field in response types

### Testing
- Added unit tests in `Agent Tests` covering static and dynamic metadata retrieval
- Added route handler tests verifying metadata serialization in list and get-agent-by-id responses
- Added assertion in editor test confirming stored agents expose metadata via `getMetadata()`

### Documentation
- Created new `getMetadata.mdx` reference page documenting the method's signature and async/Promise behavior
- Updated `agent.mdx` to document the `metadata` parameter and its `DynamicArgument` form with examples
- Added `getMetadata` to the Agents reference sidebar

### Backward Compatibility
All changes are additive and non-breaking; existing agents without metadata continue to work unchanged.

[Review Change Stack](https://app.coderabbit.ai/change-stack/mastra-ai/mastra/pull/16603)


packages/core/src/agent/agent.ts
```diff
@@ -248,6 +248,7 @@ export class Agent<
   public source?: DefinitionSource;
   #instructions: DynamicArgument<AgentInstructions, TRequestContext>;
   readonly #description?: string;
+  readonly #metadata?: Record<string, unknown>;
```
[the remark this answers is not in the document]
intojhanurag: every other getter on Agent (getInstructions, getMemory, getModel, getDescription, listTools) returns the raw reference, so cloning only here would be inconsistent.

also a shallow freeze doesn't actually stop nested mutation like agent.getMetadata().nested.foo = 'x', and deep clone would break valid values like Map or Date since metadata is Record<string, unknown>. and formatAgent / formatAgentList call getMetadata() on every /agents request so cloning on each read isn't free.

contract is the same as the other getters, the test covers the return shape.

intojhanurag:
Not for durable agents.

packages/server/src/server/handlers/agents.ts
```diff
@@ -796,6 +805,7 @@ async function formatAgent({
   isStudio: boolean;
 }): Promise<SerializedAgent> {
   const description = agent.getDescription();
+  const metadata = await agent.getMetadata({ requestContext });
```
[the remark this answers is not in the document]
intojhanurag: skipping this one, the asymmetry is intentional.

formatAgent (get-by-id) doesn't try/catch any dynamic getter, not just getMetadata. getInstructions, getLLM, listTools, getDefaultGenerateOptionsLegacy all let errors propagate to a 500. that's by design for a single-agent endpoint, a partial response is worse than a clear error.

formatAgentList (list) try-wraps every dynamic getter because one bad agent shouldn't break the list of all of them. the comment in that function spells it out.

adding try/catch around just getMetadata in formatAgent would make it the only exception in that function. verified the current behavior with an e2e, a throwing dynamic metadata function returns 500 on /agents/:id and is silently skipped on /agents, which matches the documented intent of each handler.

packages/server/src/server/handlers/agents.ts
```diff
@@ -524,6 +525,13 @@ async function formatAgentList({
   // Wrap each independent getter so a single failure doesn't abort the whole
   // serialization — the agent will still be listed with safe defaults, and the
   // failure is logged so the user can see what went wrong in `mastra dev`.
+  let metadata: Record<string, unknown> | undefined;
+  try {
+    metadata = await agent.getMetadata({ requestContext });
+  } catch (error) {
+    logger.warn('Error getting metadata for agent', { agentName: agent.name, error });
```
abhiaiyer91: We need to check if agent has a method called getMetadata. This warn log will be noisy. We should just not log it

intojhanurag: done in 55ddc8c. typeof check in both formatAgentList and formatAgent, dropped the warn log in the list path. older core versions just skip silently now.

packages/server/src/server/handlers/agents.ts
```diff
@@ -796,6 +805,7 @@ async function formatAgent({
   isStudio: boolean;
 }): Promise<SerializedAgent> {
   const description = agent.getDescription();
+  const metadata = typeof agent.getMetadata === 'function' ? await agent.getMetadata({ requestContext }) : undefined;
```
abhiaiyer91: What if this throws?

intojhanurag: yes, need to catch ig

intojhanurag: fair point, you're right. fixed in 46d3327. wrapped both formatAgent and the editor clone fallback in try/catch like formatAgentList. metadata is descriptive, not load-bearing like instructions or model, so a buggy resolver shouldn't take down the endpoint when the agent itself is fine.

abhiaiyer91 (approved):

PR #16604, opened 2026-05-14
fix(core): support signal attributes on user-messages and multimodal signal contents

CalebBarnes:
## Overview

Signals carry user input and framework context (system reminders, screenshots, etc.) into the agent loop. Each signal has a loose `contents: BaseMessageListInput` field — string, `CoreMessage`, array of messages, `MastraDBMessage`, anything `MessageList` accepts as input.

Previously, every projection of a signal — to the LLM prompt, to a DB row, to a data part — had to re-walk that loose shape with its own duck-typed helpers (`signalContentsToText`, etc.). That meant two gaps in behavior:

1. **`user-message` signals never surfaced their `attributes` to the LLM.** `signalToLLMMessage` short-circuited for user messages and returned `signal.contents` verbatim, so `attributes` like `messageId` were invisible to the model — even though they were correctly written to the data part and DB metadata.
2. **Non-`user-message` signals with multimodal contents lost their file/image parts.** `signalContentsToText` flattened everything to a string before either the LLM prompt or the DB row saw it, so a `system-reminder` carrying a screenshot would never reach the model.

This PR normalizes `signal.contents` once at `createSignal` time (lazy + memoized) via `MessageList`'s input conversion. `signalToLLMMessage` and `signalToDBMessage` both read from the canonical `MastraDBMessage[]` form instead of re-walking the raw input. The external `signal.contents` value is unchanged — `toDataPart`, the metadata stash, and round-tripping via `mastraDBMessageToSignal` all still echo the original input.

### `user-message` signals + attributes

**Before:** `signalToLLMMessage` returned `signal.contents` verbatim, attributes dropped.
```
user: hello
```

**After:** the marker is inline-wrapped into the first text part so the attributes stay adjacent to the message they describe.
```
user: <user-message messageId="m-1">hello</user-message>
```

For multimodal user messages, the wrapper is injected into the first text part and file parts are preserved on the same logical turn:
```
user: <user-message messageId="m-2">Look at this</user-message> [image]
```

If there is no text part to wrap into (file-only), the marker is prefixed as its own self-closing message:
```
user: <user-message messageId="m-2d" />
user: [image]
```

### `system-reminder` (and other non-`user-message`) signals + multimodal contents

**Before:** `signalContentsToText` flattened multimodal contents to a string, dropping every file/image part before the model saw it.
```
user: <system-reminder kind="screenshot">The user is looking at this screen.</system-reminder>
# (no image — silently dropped)
```

**After:** the marker is inline-wrapped into the first text part, and the file part rides along on the same turn.
```
user: <system-reminder kind="screenshot">The user is looking at this screen.</system-reminder> [image]
```

### DB persistence

**Before:** `signalToDBMessage` called `signalContentsToText`, so the DB row only ever stored a single text part, dropping every file/image part for downstream UIs.

**After:** `signalToDBMessage` reads the canonical normalized parts (text + file + image) and stores them faithfully.

### External contract

Unchanged. `signal.contents` is still the original input the caller passed in. `toDataPart().data.contents`, `metadata.signal.contents`, and `mastraDBMessageToSignal(toDBMessage()).contents` all preserve the original shape.

## Implementation notes

- `createSignal` runs `new MessageList().add(contents, 'input').get.all.db()` lazily on first use. Control signals that never invoke `toLLMMessage`/`toDBMessage` pay nothing.
- `signalToLLMMessage` now has three branches: (1) `user-message` with no attributes → pass through unchanged, (2) text-only → single wrapped `CoreMessage`, (3) multimodal → `injectMarkerInline` wraps the first text part (or prefixes a self-closing marker if there is no text part).
- `signalToDBMessage` builds `content.parts` from `dbMessagesToParts(normalized)` instead of the old single-text-part shape, so file/image parts survive into storage.
- `mastraDBMessageToSignal` and `dataPartToSignal` route the round-trip through `normalizeContents` + `dbMessagesToText` so non-`user-message` round-trips still get a stable text form.

## Test plan

- 26/26 tests in `packages/core/src/agent/__tests__/agent-signals.test.ts` pass, including 7 new scenarios covering:
  - `user-message` + attributes inline-wrap with single `CoreMessage` contents
  - `user-message` + attributes inline-wrap with multimodal (text + file) contents
  - `user-message` + attributes with file-only contents (self-closing marker prefix)
  - `user-message` + attributes with array-of-strings contents
  - `user-message` with `null`/`undefined`-only attributes (treated as no attributes)
  - `system-reminder` + multimodal contents (text part inline-wrapped, file part preserved)
  - `system-reminder` + file-only contents (self-closing marker prefix)
  - DB persistence preserves file/image parts faithfully
- 2693/2693 tests pass across the full `@mastra/core` suite (agent + channels + processors + message-list).
- Typecheck clean.

## Stack

PR #16517 (channel signals migration) is now stacked on top of this PR. The channels work depends on `user-message` attributes reaching the LLM, which this PR fixes.

## ELI5

The PR makes the system convert a signal's contents into a single, standard internal format once when the signal is created, and then reuses that normalized form for both sending to the LLM and saving to the database—so attributes and multimodal parts are handled consistently and not lost.

### Release Notes
- Signals now normalize `contents` to `MastraDBMessage[]` once at `createSignal` time (lazy + memoized) using `MessageList().add(contents, 'input').get.all.db()`. Both `toLLMMessage` and `toDBMessage` read from that normalized representation.
- Behavioral fixes:
  - `user-message` signals now surface `attributes` (e.g., `messageId`) to the LLM by inlining an attribute marker into the first text part; if a user message contains only files, a self-closing marker message is emitted and file parts are preserved.
  - Non-`user-message` signals with multimodal contents no longer drop file/image parts; the attribute marker is inline-wrapped into the first text part and file/image parts are preserved to both LLM and DB.
  - DB persistence stores all normalized parts (text + file + image) instead of a single flattened text part.
- Forward-compat note: DB rows may now contain full `content.parts` arrays (multimodal parts) rather than always a single text part—consumers that assumed a single text part should be updated.

### Tests
- Added/updated tests (26 agent-signal tests; 2693@mastra/core tests still pass). New scenarios cover:
  - `user-message` attribute wrapping (text, multimodal, file-only, array contents).
  - Equivalent multimodal and file-only behaviors for non-user messages (e.g., `system-reminder`).
  - DB persistence fidelity for multimodal contents and edge cases (empty contents).
- All new and existing tests pass; typecheck clean.

### Implementation
- `ContextAgentSignalInput.contents` broadened: `string` → `AgentSignalContents`.
- `createSignal` memoizes normalized `MastraDBMessage[]`.
- New helpers: normalize contents via `MessageList`, dbMessages→text helper, and dbMessages→parts helper (with empty-text fallback).
- `signalToLLMMessage` branches:
  1. user-message + no attributes → pass-through
  2. text-only → single wrapped CoreMessage
  3. multimodal → inject marker inline into first text part, or prefix self-closing marker if no text part
- `signalToDBMessage` builds `content.parts` from normalized DB messages (preserving file/image parts).
- `mastraDBMessageToSignal` and `dataPartToSignal` round-trip via normalization + dbMessagesToText for stable text reconstruction.
- `signalToXmlMarkup` updated to accept optional `contents` and render escaped/validated tag/attribute names, using self-closing form for empty content.

### Comments / Notes
- Removed an earlier branch that prefixed markers based on original `signal.contents`; all multimodal messages now use the same inline-marker approach (no nested arrays).
- Stale doc comment and tests updated (commit notes referenced).
- Developer will add a forward-compat changeset/release-note advising consumers that DB rows may include full parts arrays.

[Review Change Stack](https://app.coderabbit.ai/change-stack/mastra-ai/mastra/pull/16604)

### DIFF
--- a/.changeset/signals-normalize-contents.md
+++ b/.changeset/signals-normalize-contents.md
@@ -1,0 +1,5 @@
+---
+'@mastra/core': patch
+---
+
+Signals now normalize `contents` to `MastraDBMessage[]` once at `createSignal` time (lazy + memoized) via MessageList's input conversion. Both `toLLMMessage` and `toDBMessage` read from this canonical form, fixing two gaps: `user-message` signals now surface their `attributes` to the LLM (inline-wrapped into the first text part), and non-`user-message` signals with multimodal contents (file/image parts) no longer get flattened to text — files survive into both the LLM prompt and DB persistence. The external `signal.contents` shape and `toDataPart` echo are unchanged.
--- a/.changeset/signals-normalize-contents.md
+++ b/.changeset/signals-normalize-contents.md
@@ -3,3 +3,5 @@
 ---
 
 Signals now normalize `contents` to `MastraDBMessage[]` once at `createSignal` time (lazy + memoized) via MessageList's input conversion. Both `toLLMMessage` and `toDBMessage` read from this canonical form, fixing two gaps: `user-message` signals now surface their `attributes` to the LLM (inline-wrapped into the first text part), and non-`user-message` signals with multimodal contents (file/image parts) no longer get flattened to text — files survive into both the LLM prompt and DB persistence. The external `signal.contents` shape and `toDataPart` echo are unchanged.
+
+Forward-compat note: new signal DB rows can now contain real multimodal `content.parts` (text + file/image), where before they were always a single text part. Consumers that walked signal rows assuming a single text part should be updated to handle the full parts array.
--- a/.changeset/signals-normalize-contents.md
+++ b/.changeset/signals-normalize-contents.md
@@ -2,6 +2,11 @@
 '@mastra/core': patch
 ---
 
-Signals now normalize `contents` to `MastraDBMessage[]` once at `createSignal` time (lazy + memoized) via MessageList's input conversion. Both `toLLMMessage` and `toDBMessage` read from this canonical form, fixing two gaps: `user-message` signals now surface their `attributes` to the LLM (inline-wrapped into the first text part), and non-`user-message` signals with multimodal contents (file/image parts) no longer get flattened to text — files survive into both the LLM prompt and DB persistence. The external `signal.contents` shape and `toDataPart` echo are unchanged.
+Two fixes to how signals reach the LLM and DB:
+
+- `user-message` signals now expose their `attributes` to the LLM. Previously `toLLMMessage()` short-circuited and dropped the wrapper for user messages, so attributes were invisible to the model even though they were correctly written to `toDataPart()` and DB metadata.
+- Non-`user-message` signals (`system-reminder`, etc.) with multimodal `contents` now preserve their file/image parts in both the LLM prompt and DB storage. Previously these parts were flattened to text before either projection saw them.
+
+Internally, `createSignal` normalizes `contents` once into the canonical `MastraDBMessage[]` form, and both `toLLMMessage` and `toDBMessage` read from that. The external `signal.contents` value and `toDataPart()` output are unchanged.
 
 Forward-compat note: new signal DB rows can now contain real multimodal `content.parts` (text + file/image), where before they were always a single text part. Consumers that walked signal rows assuming a single text part should be updated to handle the full parts array.
--- a/packages/core/src/agent/__tests__/agent-signals.test.ts
+++ b/packages/core/src/agent/__tests__/agent-signals.test.ts
@@ -1,3 +1,4 @@
+import type { CoreMessage } from '@internal/ai-sdk-v4';
 import { MockLanguageModelV2, convertArrayToReadableStream } from '@internal/ai-sdk-v5/test';
 import { beforeEach, describe, expect, it, vi } from 'vitest';
 
@@ -95,7 +96,9 @@ describe('Agent signals', () => {
       metadata: { source: 'test', signal: { userProvided: true } },
     });
 
-    expect(signal.toLLMMessage()).toBe('Signal contents');
+    expect(signal.toLLMMessage()).toEqual([
+      { role: 'user', content: '<user-message priority="high">Signal contents</user-message>' },
+    ]);
     expect(signal.toDataPart()).toEqual({
       type: 'data-user-message',
       data: {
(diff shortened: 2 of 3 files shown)


TylerBarnes (changes requested):
From Mastra Code: Nice refactor overall — normalizing signal contents once through `MessageList` makes the LLM and DB projections much easier to reason about, and preserving multimodal parts in DB/LLM projection seems like the right direction.

One issue I think is worth fixing before merge:

In `signalToLLMMessage()`, the multimodal non-`user-message` branch returns:

```ts
return [prefixMessage, signal.contents] as BaseMessageListInput;
```

If `signal.contents` is already an array, this produces a nested array shape:

```ts
[CoreMessage, BaseMessageListItem[]]
```

That does not match the declared `BaseMessageListInput` shape, and the cast hides the mismatch. It would be safer to flatten array-valued contents when constructing the prefixed message list:

```ts
return Array.isArray(signal.contents)
  ? [prefixMessage, ...signal.contents]
  : [prefixMessage, signal.contents];
```

I’d also add a test for a multimodal `system-reminder` whose `contents` is an array, to lock down that path.

I don’t see a major backwards-compat issue for old DB rows as long as historical signal rows saved full original `contents` in `content.metadata.signal.contents`, since `mastraDBMessageToSignal()` prefers that metadata before falling back to legacy `content.content` / `content.parts`. The main forward-looking compatibility change is that new signal DB rows can now have real multimodal `content.parts`, so any consumers assuming signal rows are always a single text part should be checked.

CalebBarnes:
Good eye — that `[prefixMessage, signal.contents]` branch was actually removed in commit `1312b9c4c1` ("inline-wrap multimodal signals so the marker stays adjacent to its payload"). Non-`user-message` multimodal signals now go through `injectMarkerInline` just like `user-message` ones, so the marker is inlined into the first text part across all messages instead of being prefixed as a separate message. No nested arrays anywhere.

The doc comment on `signalToLLMMessage` was stale and still described the old shape — fixed that in `9bbd44f3d4`, along with the test you asked for: `system-reminder` with array-valued multimodal contents (mixed text + file CoreMessages). It verifies the marker only wraps the very first text part across the whole array and subsequent messages pass through untouched.

Agreed on the forward-compat note about consumers that assume signal DB rows are always a single text part — worth flagging in the changeset/release notes. Will add that.

TylerBarnes (approved):
From Mastra Code: The follow-up commit addresses my requested change. Routing multimodal signals through the normalized `injectMarkerInline()` path removes the nested-array concern and keeps the marker adjacent to the payload, which is cleaner than the originally suggested spread fix. Looks good to me pending CI.

PR #16616, opened 2026-05-14
fix(schema-compat): produce OpenAI strict-mode compatible JSON Schema for optional fields

### LINKED ISSUE #16383: [BUG] zod schemas produce non-strict-mode-compatible JSON Schema for OpenAI structured outputs (HTTP 400 for v3, type mangling for v4)
## Describe the Bug

`@mastra/schema-compat`'s `zodToJsonSchema` produces JSON Schema that doesn't satisfy OpenAI's strict-mode structured outputs / strict tool calling. Two failure modes depending on which zod version you hand mastra:

**zod v3 schemas with `.optional()` → HTTP 400** from OpenAI:

> Invalid schema for response_format 'response': In context=(), 'required' is required to be supplied and to be an array including every key in properties. Missing 'optional'.

The v3 path in `zodToJsonSchema` (in `dist/zod-to-json.js`) returns standard JSON Schema where optional fields are NOT in `required` — but OpenAI strict mode requires every property to appear in `required`.

**zod v4 schemas with `.optional()` → HTTP 200, but type is mangled** to `["string","number","boolean","null"]` (the primitive any-or-null union). Comes from `fixAnyOfNullable`'s `defaultEmptySchemaHandler` in `@mastra/schema-compat/dist/index.js` filling empty `{}` schemas with the primitive union. The LLM sees the union and sometimes fills the field with a number or boolean placeholder instead of a string or null, breaking downstream validation.

**Cross-provider fallback amplifies this.** An Anthropic-primary agent with OpenAI fallback builds the schema once for Anthropic; if Anthropic fails and the OpenAI fallback runs, the same non-strict schema is sent to OpenAI and the call 400s.

The fix is sitting in `OpenAISchemaCompatLayer.processZodType` — it has correct `.optional()` → `.nullable().transform(null→undefined)` logic that handles BOTH v3 and v4. But it's gated:

```js
shouldApply() {
    return !this.getModel().supportsStructuredOutputs && (...openai/groq...);
}
```

So it only runs for legacy models that don't support structured outputs at all. Modern strict-mode-capable models (gpt-4o, gpt-4.1, gpt-5.x) bypass it entirely.

**Suggested fix**: apply `OpenAISchemaCompatLayer.processZodType` (or equivalent) for all OpenAI models, not just legacy ones — its existing logic already handles both v3 and v4 correctly.

## Steps To Reproduce

1. Clone the repro and `cd` in:
   ```sh
   git clone https://github.com/bluepnume/mastra-bug-repros
   cd mastra-bug-repros/openai-strict-mode
   ```
2. Install + run:
   ```sh
   npm install
   OPENAI_API_KEY=sk-... npm run repro
   ```
3. Observe:
   - **v3 path**: HTTP 400 from OpenAI with the strict-mode error
   - **v4 path**: HTTP 200 but wire schema has `optional: { type: ["string","number","boolean","null"] }` (description also stripped)

https://github.com/bluepnume/mastra-bug-repros/tree/main/openai-strict-mode

## Expected Behavior

Either path should produce a schema that OpenAI strict mode accepts AND that doesn't tell the LLM the wrong type for optional fields. Specifically:

- every property in `required`
- optional fields widened to `[T, "null"]` (NOT the `["string","number","boolean","null"]` primitive union)
- description and other metadata preserved

## Environment Information

```
System:
  OS: macOS 26.4
  CPU: arm64
Binaries:
  Node: 22.22.1
npmPackages:
  @mastra/core: 1.10.0
  @mastra/schema-compat: 1.1.3
  @ai-sdk/openai: 3.0.43
  @ai-sdk/provider-utils: 4.0.20
  zod: 4.3.6
LLM Provider: OpenAI (gpt-4.1, gpt-4o-mini, gpt-5.x — all affected)
```

abhilov23:
## Description

`@mastra/schema-compat`'s `zodToJsonSchema` produced JSON Schema that failed OpenAI strict-mode structured outputs in two ways:

**Zod v3 + `.optional()` → HTTP 400**
Optional fields were excluded from the `required` array. OpenAI strict mode requires every key in `properties` to appear in `required`.

**Zod v4 + `.optional()` → type mangled to `["string","number","boolean","null"]`**
The `fixAnyOfNullable` fallback for empty `{}` schemas replaced the type with a primitive union, causing the LLM to fill the field with wrong types and breaking downstream validation. Description metadata was also stripped.

### Changes

- **`zod-to-json.ts` — v4 path**: Added `ZodOptional` handler in the `override` callback to emit `[T, "null"]` at source instead of `{}`
- **`zod-to-json.ts` — v3 path**: Apply `fixAnyOfNullable` + `ensureAllPropertiesRequired` post-processing (was returning raw output before)
- **`zod-to-json.ts` — `fixAnyOfNullable`**: Stop mangling empty `{}` schemas to primitive union — leave them for the override to handle
- **`zod-to-json.ts` — `ensureAllPropertiesRequired`**: When adding a previously-optional field to `required`, widen its type to `[T, "null"]` to satisfy OpenAI strict mode
- **`zod-to-json.test.ts`**: Added regression test for issue #16383; updated two stale tests that were asserting the old broken behavior

### Result

Both v3 and v4 optional fields now produce:
```json
{
  "properties": {
    "nickname": { "type": ["string", "null"], "description": "..." }
  },
  "required": ["name", "nickname"],
  "additionalProperties": false
}
```

## Related issue(s)

Fixes #16383

Repro: https://github.com/bluepnume/mastra-bug-repros/tree/main/openai-strict-mode

## Type of change

- [x] Test update

## Checklist

## ELI5
This PR fixes schema generation so optional fields never get dropped or turned into "anything goes" types for OpenAI strict mode — every property is listed as required and optional fields are explicitly widened to allow null while keeping descriptions.

## Problem
zodToJsonSchema produced JSON Schema incompatible with OpenAI strict-mode:
- Zod v3: .optional() omitted fields from required, causing OpenAI strict-mode to reject the schema.
- Zod v4: .optional() produced empty {} schemas which prior post-processing mangled into primitive unions (["string","number","boolean","null"]), losing correct types and metadata.

Cross-provider reuse risk: schemas built for other providers could break when reused with OpenAI strict-mode.

## Solution
- Zod v4: added a ZodOptional override to emit the inner schema widened to include "null" (type: [T, "null"]) at conversion time and preserve description/metadata.
- Zod v3: run the stricter post-processing pipeline on v3 outputs (fixAnyOfNullable then ensureAllPropertiesRequired) instead of returning raw conversion results.
- fixAnyOfNullable: stop converting empty {} property schemas into primitive-type unions; leave that handling to optional-handling logic.
- ensureAllPropertiesRequired / widenOptionalForNull:
  - Ensures every key in properties appears in required (OpenAI strict-mode requirement).
  - Widens previously-optional property schemas to accept null, handling const/enum/anyOf/oneOf/type-array forms and preserving other metadata.
- Tests:
  - Added regression test for `#16383`.
  - Updated tests to assert optional fields are included in required and widened to include null while preserving descriptions.
  - Added nested test covering optional object properties inside optional array items (verifies items' required includes previously-optional fields and their types are widened to ['string','null']).

## Files changed (high level)
- packages/schema-compat/src/zod-to-json.ts
  - Added Zod v4 optional handling, changed fixAnyOfNullable behavior, implemented ensureAllPropertiesRequired (with widening), and applied post-processing to both v3 and v4 conversion paths.
- packages/schema-compat/src/zod-to-json.test.ts
  - Updated and expanded tests (regression `#16383`, nested array/item coverage, and direct helper tests).
- .changeset/fix-schema-compat-openai-strict-mode.md
  - Patch changeset documenting the fix.

## Result
Zod v3 and v4 now produce OpenAI strict-mode-compatible JSON Schema: every property in properties appears in required; optional fields are widened to include null (e.g., ["<type>","null"] or equivalent) and descriptions/metadata are preserved. Fixes `#16383`.

## Reviewer notes / remaining risks
- A reviewer initially pointed out recursion gate concerns for nested schemas; the author added a nested-array test that demonstrates the PR passes (the reviewer re-tested and deleted their objection).
- Additional strict-mode edge cases were later raised (optional unions, discriminated unions, enums, literals) and a follow-up helper (widenOptionalForNull) was proposed in a separate commit/branch; those changes are not part of this PR but were suggested to be copied/cherry-picked if desired.

[Review Change Stack](https://app.coderabbit.ai/change-stack/mastra-ai/mastra/pull/16616?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/.changeset/fix-schema-compat-openai-strict-mode.md
+++ b/.changeset/fix-schema-compat-openai-strict-mode.md
@@ -1,0 +1,5 @@
+---
+"@mastra/schema-compat": patch
+---
+
+fix(schema-compat): produce OpenAI strict-mode compatible JSON Schema for optional fields
--- a/packages/schema-compat/src/zod-to-json.test.ts
+++ b/packages/schema-compat/src/zod-to-json.test.ts
@@ -765,3 +765,48 @@ it('reproduce issue #16383 - optional fields must appear in required for OpenAI 
 
   expect(nicknameProp.description).toBe('The nickname');
 });
+
+it('should handle optional fields inside array items (nested strict-mode)', () => {
+  const schema = z.object({
+    list: z
+      .array(
+        z.object({
+          keep: z.string(),
+          omit: z.string().optional(),
+        }),
+      )
+      .optional(),
+  });
+
+  const result = zodToJsonSchema(schema) as any;
+
+  expect(result.required).toContain('list');
+
+  expect(result.properties.list.items.required).toContain('keep');
+  expect(result.properties.list.items.required).toContain('omit');
+
+  expect(result.properties.list.items.properties.omit.type).toEqual(['string', 'null']);
+});
+
+it('should handle optional fields inside array items (nested strict-mode)', () => {
+  const schema = z.object({
+    list: z
+      .array(
+        z.object({
+          keep: z.string(),
+          omit: z.string().optional(),
+        }),
+      )
+      .optional(),
+  });
+
+  const result = zodToJsonSchema(schema) as any;
+  expect(result.properties.list.items.type).toBe('object');
+  expect(result.required).toContain('list');
+  expect(result.properties.list.type).toEqual(['array', 'null']);
+
+  expect(result.properties.list.items.required).toContain('keep');
+  expect(result.properties.list.items.required).toContain('omit');
+
+  expect(result.properties.list.items.properties.omit.type).toEqual(['string', 'null']);
+});
--- a/packages/schema-compat/src/zod-to-json.test.ts
+++ b/packages/schema-compat/src/zod-to-json.test.ts
@@ -736,9 +736,6 @@ it('reproduce issue #16383 - optional fields must appear in required for OpenAI 
 
   const result = zodToJsonSchema(schema) as any;
 
-  console.log(JSON.stringify(result, null, 2));
-
-  // Bug 1
   expect(result.required).toBeDefined();
 
   expect(result.required).toEqual(expect.arrayContaining(['name', 'nickname']));
@@ -747,7 +744,6 @@ it('reproduce issue #16383 - optional fields must appear in required for OpenAI 
 
   expect(nicknameProp).toBeDefined();
 
-  // Bug 2
   if (nicknameProp.anyOf) {
     expect(nicknameProp.anyOf).toEqual(
       expect.arrayContaining([
@@ -767,6 +763,5 @@ it('reproduce issue #16383 - optional fields must appear in required for OpenAI 
     expect(nicknameProp.type).not.toContain('boolean');
   }
 
-  // Bug 3
   expect(nicknameProp.description).toBe('The nickname');
 });
--- a/packages/schema-compat/src/zod-to-json.test.ts
+++ b/packages/schema-compat/src/zod-to-json.test.ts
@@ -780,33 +780,10 @@ it('should handle optional fields inside array items (nested strict-mode)', () =
 
   const result = zodToJsonSchema(schema) as any;
 
-  expect(result.required).toContain('list');
-
-  expect(result.properties.list.items.required).toContain('keep');
-  expect(result.properties.list.items.required).toContain('omit');
-
-  expect(result.properties.list.items.properties.omit.type).toEqual(['string', 'null']);
-});
-
-it('should handle optional fields inside array items (nested strict-mode)', () => {
-  const schema = z.object({
-    list: z
-      .array(
-        z.object({
-          keep: z.string(),
-          omit: z.string().optional(),
-        }),
-      )
-      .optional(),
-  });
-
-  const result = zodToJsonSchema(schema) as any;
   expect(result.properties.list.items.type).toBe('object');
   expect(result.required).toContain('list');
   expect(result.properties.list.type).toEqual(['array', 'null']);
-
   expect(result.properties.list.items.required).toContain('keep');
   expect(result.properties.list.items.required).toContain('omit');
-
   expect(result.properties.list.items.properties.omit.type).toEqual(['string', 'null']);
 });
--- a/packages/schema-compat/src/zod-to-json.test.ts
+++ b/packages/schema-compat/src/zod-to-json.test.ts
@@ -787,3 +787,141 @@ it('should handle optional fields inside array items (nested strict-mode)', () =
   expect(result.properties.list.items.required).toContain('omit');
   expect(result.properties.list.items.properties.omit.type).toEqual(['string', 'null']);
 });
+
+describe('strict-mode widening for anyOf, oneOf, enum, and const', () => {
+  // GAP #1: ensureAllPropertiesRequired widens only fields that have a `type` string/array.
+  // A pure anyOf schema (z.union) has no `type` field, so it gets added to `required` but
+  // is NOT widened with a `null` branch. OpenAI strict mode then sees a required field
+  // that cannot be null, even though the source was .optional().
+  it('GAP #1a: optional z.union field gets a null branch', () => {
+    const schema = z.object({
+      id: z.union([z.string(), z.number()]).optional(),
+    });
+
+    const result = zodToJsonSchema(schema) as any;
+
+    expect(result.required).toContain('id');
+
+    const idProp = result.properties.id;
+
+    if (idProp.anyOf) {
+      const hasNull = idProp.anyOf.some(
+        (v: any) => v && (v.type === 'null' || (Array.isArray(v.type) && v.type.includes('null'))),
+      );
+      expect(hasNull).toBe(true);
+    } else if (idProp.oneOf) {
+      const hasNull = idProp.oneOf.some(
+        (v: any) => v && (v.type === 'null' || (Array.isArray(v.type) && v.type.includes('null'))),
+      );
+      expect(hasNull).toBe(true);
+    } else if (Array.isArray(idProp.type)) {
+      expect(idProp.type).toContain('null');
+    } else {
+      // Field is required but has no way to express null — strict mode would reject.
+      throw new Error(`optional union field has no null branch. Got: ${JSON.stringify(idProp)}`);
+    }
+  });
+
+  it('GAP #1b: optional z.discriminatedUnion field gets a null branch', () => {
+    const schema = z.object({
+      event: z
+        .discriminatedUnion('kind', [
+          z.object({ kind: z.literal('a'), value: z.string() }),
+          z.object({ kind: z.literal('b'), count: z.number() }),
+        ])
+        .optional(),
+    });
+
+    const result = zodToJsonSchema(schema) as any;
+
+    expect(result.required).toContain('event');
+
+    const eventProp = result.properties.event;
+
+    if (eventProp.anyOf) {
+      const hasNull = eventProp.anyOf.some(
+        (v: any) => v && (v.type === 'null' || (Array.isArray(v.type) && v.type.includes('null'))),
+      );
+      expect(hasNull).toBe(true);
+    } else if (eventProp.oneOf) {
+      const hasNull = eventProp.oneOf.some(
+        (v: any) => v && (v.type === 'null' || (Array.isArray(v.type) && v.type.includes('null'))),
+      );
+      expect(hasNull).toBe(true);
+    } else if (Array.isArray(eventProp.type)) {
+      expect(eventProp.type).toContain('null');
+    } else {
+      throw new Error(`optional discriminated union has no null branch. Got: ${JSON.stringify(eventProp)}`);
+    }
+  });
+
+  // GAP #2: the v4 override's typeMap only handles string/number/boolean/integer.
+  // For things like z.enum() and z.literal(), the override bails and the schema
+  // ends up with type 'string' + an enum/const constraint that doesn't include null.
+  // After ensureAllPropertiesRequired widens type to ['string','null'], the enum
+  // constraint still lacks 'null', so a null value would fail the enum/const check.
+  it('GAP #2a: optional z.enum produces a schema where null satisfies the type but not the enum', () => {
+    const schema = z.object({
+      status: z.enum(['a', 'b']).optional(),
+    });
+
+    const result = zodToJsonSchema(schema) as any;
+
+    expect(result.required).toContain('status');
+
+    const statusProp = result.properties.status;
+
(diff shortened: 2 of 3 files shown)


bluepnume:
Heads up — the `ensureAllPropertiesRequired` implementation in this PR has two gaps that the test cases don't currently catch. I posted the details on the linked issue but they apply directly here:

#16383 (comment): <https://github.com/mastra-ai/mastra/issues/16383#issuecomment-4463594732>

**TL;DR**:

1. **No recursion into arrays**: the function only walks `result.type === 'object'`, so the `items` schema of any array (e.g. `z.array(z.object({ ..., inner: z.string().optional() }))`) is never strict-mode-processed.

2. **Recursion gate breaks after widening**: when a field gets widened from `'array'` (or `'object'`) to `['array', 'null']`, any subsequent walk gated on `type === 'object'` / `type === 'array'` silently skips its children.

Minimal repro that the PR's current logic still fails (OpenAI returns HTTP 400):

```ts
z.object({
    list: z.array(z.object({
        keep: z.string(),
        omit: z.string().optional()   // ← will NOT be widened or added to required
    })).optional()
});
```

Expected wire schema is in the linked comment. We fixed the same bug in our local workaround by driving the recursion off `properties` / `items` presence instead of `type`.

abhilov23:
Hi @bluepnume  thanks for the detailed repro!

I traced through the implementation and this case is already handled correctly.

The key is that the recursion into `items` is driven off the presence of `result.items`, not gated on `type === 'array'`, so it continues to fire even after the parent type is widened to `['array', 'null']`. The recursion also happens before the widening in the parent, so children are always processed on their original type.

I added a test covering your exact repro and it passes on both v3 and v4:

```ts

it('should handle optional fields inside array items (nested strict-mode)', () => {
  const schema = z.object({
    list: z
      .array(
        z.object({
          keep: z.string(),
          omit: z.string().optional(),
        }),
      )
      .optional(),
  });

  const result = zodToJsonSchema(schema) as any;

  expect(result.properties.list.items.type).toBe('object');
  expect(result.required).toContain('list');
  expect(result.properties.list.type).toEqual(['array', 'null']);
  expect(result.properties.list.items.required).toContain('keep');
  expect(result.properties.list.items.required).toContain('omit');
  expect(result.properties.list.items.properties.omit.type).toEqual(['string', 'null']);
});

```

Based on that, I don't think there's a bug here—the nested optional field handling inside array items appears to already behave as expected.

bluepnume:
@abhilov23 you were right - my apologies. I cloned the PR branch, ran `pnpm vitest run zod-to-json.test.ts -t "nested strict-mode"`, and the test passes. The fix does cover the nested case correctly.

My mistake was in a hand-inlined trace where I missed that `ensureAllPropertiesRequired` already has an `if (result.items) { ... ensureAllPropertiesRequired(result.items) }` block further down (pre-dates this PR so wasn't in the diff). I deleted my earlier comment to avoid leaving noise. Thanks for the patient response.

bluepnume:
Following up on this PR with four additional strict-mode cases that the current `ensureAllPropertiesRequired` does not cover. All four produce schemas that OpenAI strict mode rejects (or that allow values incompatible with the declared shape).

I verified each by adding tests against the PR branch and running with both v3 and v4 (`pnpm vitest run zod-to-json.test.ts`). Failure mode and minimal repro for each below.

### 1. Optional `z.union(...)` (v4 only)

`ensureAllPropertiesRequired` only widens a field's `type` when `type` is a string or array. A `z.union(...)` produces an `anyOf` with no `type` field, so the field gets added to `required` but never gets a null branch.

```ts
const schema = z.object({
  id: z.union([z.string(), z.number()]).optional(),
});

const result = zodToJsonSchema(schema);
// Actual (v4):
// {
//   properties: { id: { anyOf: [{type:'string'}, {type:'number'}] } },
//   required: ['id'],
//   ...
// }
// Strict mode rejects: id is required but has no null branch.

const idProp = result.properties.id;
const hasNull = idProp.anyOf?.some(v => v.type === 'null')
  || (Array.isArray(idProp.type) && idProp.type.includes('null'));
expect(hasNull).toBe(true); // fails
```

v3 happens to pass because `zod-to-json-schema` emits a `type: ["string","number"]` array instead of `anyOf`, which the existing widening handles. Fix: when a previously-optional field has `anyOf` / `oneOf`, push `{type: 'null'}` into the variants.

### 2. Optional `z.discriminatedUnion(...)` (both v3 and v4)

Same root cause as #1 but for discriminated unions, which always emit `anyOf` (v3) or `oneOf` (v4) regardless of zod version.

```ts
const schema = z.object({
  event: z.discriminatedUnion('kind', [
    z.object({ kind: z.literal('a'), value: z.string() }),
    z.object({ kind: z.literal('b'), count: z.number() }),
  ]).optional(),
});

const result = zodToJsonSchema(schema);
// Actual (v4):
// {
//   properties: { event: { oneOf: [{...}, {...}] } },
//   required: ['event'],
//   ...
// }
// (v3 is identical with anyOf instead of oneOf.)

const eventProp = result.properties.event;
const variants = eventProp.anyOf ?? eventProp.oneOf ?? [];
const hasNull = variants.some(v => v.type === 'null');
expect(hasNull).toBe(true); // fails on both v3 and v4
```

Same fix as #1 covers this.

### 3. Optional `z.enum([...])` (both v3 and v4)

The widening adds `'null'` to `type` but leaves the `enum` constraint untouched. `null` then satisfies `type` but fails `enum`, so any value the LLM emits for the absent field is invalid.

```ts
const schema = z.object({
  status: z.enum(['a', 'b']).optional(),
});

const result = zodToJsonSchema(schema);
// Actual:
// {
//   properties: { status: { type: ['string', 'null'], enum: ['a', 'b'] } },
//   required: ['status'],
//   ...
// }
// null satisfies type but not enum.

const statusProp = result.properties.status;
if (statusProp.enum) {
  expect(statusProp.enum).toContain(null); // fails
}
```

Fix: when widening an optional field that has an `enum`, push `null` into the enum too (or wrap in `anyOf: [{...original}, {type: 'null'}]`).

### 4. Optional `z.literal(...)` (both v3 and v4)

Same shape as #3 but with `const` instead of `enum`. The widening adds `'null'` to `type` but `const` pins the value, so the field can only ever be the literal value, never null.

```ts
const schema = z.object({
  kind: z.literal('only').optional(),
});

const result = zodToJsonSchema(schema);
// Actual:
// {
//   properties: { kind: { type: ['string', 'null'], const: 'only' } },
//   required: ['kind'],
//   ...
// }
// const pins kind to 'only'; null is unreachable.

const kindProp = result.properties.kind;
if ('const' in kindProp) {
  // const cannot be widened in place; needs anyOf wrapping.
  expect(kindProp.anyOf?.some(v => v.type === 'null')).toBe(true); // fails
}
```

Fix: when widening an optional field that has `const`, wrap it: `{ anyOf: [{...originalSchema}, {type: 'null'}] }`.

bluepnume:
I went ahead and put a fix together with passing tests for all four cases above. If you'd like to fold it into this PR:

Commit: https://github.com/bluepnume/mastra/commit/db222b50fa
Branch: https://github.com/bluepnume/mastra/tree/fix/openai-strict-mode-anyof-enum-const-followup

The fix extracts a `widenOptionalForNull` helper inside `ensureAllPropertiesRequired` that handles `const` (wrap in anyOf), `anyOf`/`oneOf` (push null variant), scalar `type` (existing widening), and `enum` (append null). All 104 existing tests still pass; the 4 new cases live under `describe('strict-mode widening for anyOf, oneOf, enum, and const')`. Cherry-pick or copy whatever's useful.

abhilov23:
Hey @bluepnume, first of all really appreciate you taking the time to not just report the edge cases but also put together a fix with passing tests — that's super helpful especially for someone like me who's fairly new to contributing here!
I've cherry-picked your commit into the PR and all 104 tests pass on both v3 and v4. The widenOptionalForNull helper is a much cleaner approach than what I had — glad we could fold it in.
Thanks again!

percymcn:
There is a second strict-mode defect in the file this PR edits, and it fails in the opposite direction to the two above: instead of producing a schema OpenAI **rejects**, `prepareJsonSchemaForOpenAIStrictMode` produces one OpenAI **accepts** and that can never be populated.

`z.record(...)` renders as an open map (`additionalProperties: <valueSchema>`, no `properties`). The prep sets `additionalProperties: false` on it. On a node with no `properties` that does not *close* the object, it *empties* it — the only legal instance becomes `{}`, and the value schema is discarded.

Measured on `@mastra/schema-compat@1.3.5` + `zod@3.25.76`, verdicts from `openai@7.4.0`'s own transformer:

```
                    raw        after prep
record(string)      THROW  ->  ACCEPT   {"tags":{"type":"object","additionalProperties":false}}
record(object)      THROW  ->  ACCEPT   {"users":{"type":"object","additionalProperties":false}}
nested record       THROW  ->  ACCEPT   (same, one level down)
plain object        ACCEPT ->  ACCEPT   (untouched — the prep is not over-editing)
```

`z.record(z.object({ age: z.number() }))` is the one that hurts: the entire value model is gone, the request 200s, and the field comes back `{}` forever. The pre-fix behaviour was a loud 400 naming the exact path; this is silent.

Two things that matter for how you fix it:

**1. A regression test built on the shapes in this PR passes either way.** The `.optional()` fixtures have no `z.record()` in them, so they exercise none of this — the same pattern as a test that can't fail on the case it's meant to guard. A case that catches it needs a `record` field and must assert on the *content* (`additionalProperties` still describes the value type, or the schema is rejected), not merely that the call succeeded or that the vendor accepted the result. Accept/reject alone cannot see this bug — acceptance is precisely what went wrong.

**2. There is an offline oracle, so this is CI-testable with no API key:**

```js
const { toStrictJsonSchema } = require('openai/lib/transform');
toStrictJsonSchema(schema); // throws on strict-invalid input
```

On the fix itself: an open map genuinely cannot be expressed in OpenAI strict mode, so there is no local repair — which is why setting `false` is tempting and wrong. The honest options are to surface it to the caller, or to remodel: `Array<{ key: string, value: T }>` keeps both halves fully typed under strict mode, and declaring the keys you actually expect works when they're known. Silently emptying the field is the one outcome that can't be detected downstream — by the time anything sees the schema, the value type is already gone.

abhiaiyer91:
Thanks for putting this together! #16383 was resolved on `main` by #21294, so this PR is no longer needed. Closing it to keep the queue focused, but we really appreciate the work here. If you find a case that #21294 doesn't cover, please open a new issue with a repro and we'll pick it up.

PR #16618, opened 2026-05-14
feat(mastracode): add /skill/<name> command to activate skills explicitly

### LINKED ISSUE #16344: Add `/skill:<name>` slash command to invoke individual skills in Mastra Code
> *This was generated by AI during triage.*

## Description

I want to be able to explicitly call individual skills in Mastra Code via a slash command.

In Pi agent this is possible with:

```text
/skill:name-of-skill
```

Claude Code supports a similar workflow with:

```text
/name-of-skill
```

I prefer Pi agent's `/skill:<name>` form because it is explicit and avoids overloading the top-level slash-command namespace. Since Mastra Code is based on Pi, it would be useful to support the same command style.

## Expected behavior

Mastra Code should allow invoking a specific installed skill by name, for example:

```text
/skill:github-triage
/skill:browse
/skill:investigate
```

When entered, Mastra Code should activate/run that skill directly.

## Current behavior

Mastra Code does not appear to support directly invoking individual skills through a `/skill:<name>` slash command.

## Steps to reproduce

1. Open Mastra Code.
2. Ensure one or more skills are installed/available.
3. Try to invoke a skill directly with `/skill:name-of-skill`.
4. The command is not recognized or does not activate the named skill.

## Motivation

This would make skill invocation more discoverable and controllable, especially for users coming from Pi agent. It also keeps skill invocation under a clear namespace instead of competing with built-in slash commands.

## Environment

- Mastra Code v0.18.0-alpha.10
- Node: v26.0.0
- OS: Darwin

intojhanurag:
## What

Adds a `/skill/<name>` slash command to Mastra Code that explicitly activates any installed workspace skill in the current conversation.

```text
/skill/github-triage
/skill/release-check focus tests
```

Complements the existing automatic skill activation. Users get a deterministic way to invoke a specific skill without hoping the model picks it. Closes #16344.

## How

- New `handleSkillCommand` in `mastracode/src/tui/commands/skills.ts` resolves the workspace, loads the skill via `workspace.skills.get`, and sends a `<skill name="...">…</skill>` message containing the skill instructions plus any `references/`, `scripts/`, or `assets/` paths.
- Dispatcher branch in `mastracode/src/tui/command-dispatch.ts` routes any `/skill/<prefix>` to the new handler, gated behind the existing goal-judge input lock.
- `TUIState.skillCommands` powers per-skill autocomplete entries (`skill/hello-skill`, etc.) alongside the existing `goal/<name>` entries.
- `/skill/<name>` added to `/help` and to `docs/src/mastra-code/{index,configuration}.mdx`.
- Shared formatter: exported `formatSkillActivation(skill)` from `@mastra/core/workspace` so the built-in `skill` tool and the new slash command produce byte-identical output.
- Lazy autocomplete refresh: `setupAutocomplete` runs immediately with whatever's available; `refreshSkillsAutocomplete` then resolves the workspace in the background and rebuilds the provider so `skill/<name>` entries appear without a restart.

## Error handling

- `/skill/` → `Usage: /skill/<name>`
- `/skill/unknown` → `Skill not found: unknown. Available skills: …`
- No workspace skills configured → friendly inline message.

## Tests

- New `mastracode/src/tui/commands/__tests__/skills.test.ts`: happy path, `pendingNewThread` sequencing, not-found, empty-name, no-skills-configured.
- Updated `command-dispatch.test.ts`, `setup-keyboard-shortcuts.test.ts`, `help-overlay.test.ts`.
- Core skill-tool tests (`packages/core/src/workspace/skills/tools.test.ts`) confirm the extracted `formatSkillActivation` still produces the same output.
- Manually verified end-to-end against a demo project with three skills (plain, args, references): activation, args propagation, `skill_read` of reference files, autocomplete, error paths, and `/help` entry.

## Type of Change

## Checklist

- [x] I have made corresponding changes to the documentation
- [x] I have added tests that prove my feature works

### DIFF
--- a/.changeset/cozy-walls-teach.md
+++ b/.changeset/cozy-walls-teach.md
@@ -2,11 +2,11 @@
 'mastracode': minor
 ---
 
-Added the `/skill:<name>` command to explicitly activate an installed workspace skill in the current conversation. This complements automatic skill activation and mirrors the `/skill:<name>` form from Pi agent.
+Added the `/skill/<name>` command to explicitly activate an installed workspace skill in the current conversation. This complements automatic skill activation.
 
 ```text
-/skill:github-triage
-/skill:release-check focus tests
+/skill/github-triage
+/skill/release-check focus tests
 ```
 
 The command loads the skill's instructions (plus any `references/`, `scripts/`, and `assets/` paths the skill ships) and sends them to the agent. Use `/skills` to list available skills.
--- a/.changeset/cozy-walls-teach.md
+++ b/.changeset/cozy-walls-teach.md
@@ -11,4 +11,15 @@ Added the `/skill/<name>` command to explicitly activate an installed workspace 
 
 The command loads the skill's instructions (plus any `references/`, `scripts/`, and `assets/` paths the skill ships) and sends them to the agent. Use `/skills` to list available skills.
 
+Skills can opt out of direct user invocation by setting `metadata.userInvokable: false` in their frontmatter — those skills remain available for automatic activation by the agent but do not appear in `/skill/<name>` autocomplete, the `/skills` listing, or accept direct invocation.
+
+```md title=".mastracode/skills/internal-helper/SKILL.md"
+---
+name: internal-helper
+description: Used by the agent internally; not for direct user invocation.
+metadata:
+  userInvokable: false
+---
+```
+
 Closes #16344.
--- a/.changeset/cozy-walls-teach.md
+++ b/.changeset/cozy-walls-teach.md
@@ -11,14 +11,13 @@ Added the `/skill/<name>` command to explicitly activate an installed workspace 
 
 The command loads the skill's instructions (plus any `references/`, `scripts/`, and `assets/` paths the skill ships) and sends them to the agent. Use `/skills` to list available skills.
 
-Skills can opt out of direct user invocation by setting `metadata.userInvokable: false` in their frontmatter — those skills remain available for automatic activation by the agent but do not appear in `/skill/<name>` autocomplete, the `/skills` listing, or accept direct invocation.
+Skills can opt out of direct user invocation by setting `user-invocable: false` in their frontmatter — those skills remain available for automatic activation by the agent but do not appear in `/skill/<name>` autocomplete, the `/skills` listing, or accept direct invocation.
 
 ```md title=".mastracode/skills/internal-helper/SKILL.md"
 ---
 name: internal-helper
 description: Used by the agent internally; not for direct user invocation.
-metadata:
-  userInvokable: false
+user-invocable: false
 ---
 ```
 
--- a/.changeset/cozy-walls-teach.md
+++ b/.changeset/cozy-walls-teach.md
@@ -1,0 +1,14 @@
+---
+'mastracode': minor
+---
+
+Added the `/skill:<name>` command to explicitly activate an installed workspace skill in the current conversation. This complements automatic skill activation and mirrors the `/skill:<name>` form from Pi agent.
+
+```text
+/skill:github-triage
+/skill:release-check focus tests
+```
+
+The command loads the skill's instructions (plus any `references/`, `scripts/`, and `assets/` paths the skill ships) and sends them to the agent. Use `/skills` to list available skills.
+
+Closes #16344.
--- a/.changeset/shared-skill-formatter.md
+++ b/.changeset/shared-skill-formatter.md
@@ -4,6 +4,8 @@
 
 Exposed `formatSkillActivation(skill)` from `@mastra/core/workspace`. It returns the activation payload — instructions plus references, scripts, and assets listings — that the built-in `skill` tool uses, so callers (e.g. an explicit `/skill/<name>` slash command) can produce the same output without duplicating the formatting logic.
 
+Also preserves the `user-invocable` skill frontmatter field in workspace skill metadata.
+
 ```ts
 import { formatSkillActivation } from '@mastra/core/workspace';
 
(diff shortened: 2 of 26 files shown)


TylerBarnes (changes requested):
Nice feature overall — explicit/manual skill activation fits Mastra’s existing skill model well.

A few review notes:

1. **Consider `/skill/<name>` instead of `/skill:<name>`**

   Agent Skills doesn’t standardize slash-command invocation syntax, so Mastra Code can choose the syntax that works best for its own UX.

   I think `/skill/foo` is a better fit:
   - easier to type
   - reads like a command namespace

2. **Skill autocomplete should update lazily**

   From testing, autocomplete appears to show only the static `skill:` entry, not installed skills.

   The likely issue is that autocomplete setup depends on `state.skillCommands`, but skill loading currently relies on an already-resolved workspace. Initial autocomplete setup shouldn’t block on resolving workspace/skills.

   Suggested behavior:
   - initialize autocomplete immediately with static commands
   - resolve/load workspace skills in the background
   - refresh/reinstall the autocomplete provider once `state.skillCommands` is populated

   The current tests manually seed `state.skillCommands`, so they verify rendering but not the real async load path. It’d be good to add coverage where `getWorkspace()` is initially empty, skills load later, and autocomplete refreshes with `skill:<name>` entries.

3. **Avoid duplicating skill activation formatting**

   This PR duplicates the core skill tool’s activation formatting: instructions plus references/scripts/assets.

   Since manual activation and normal agent-driven activation should stay equivalent, it’d be better to share a helper with the core `skill` tool, or at least add parity coverage so these paths don’t drift.

4. **Docs/runtime supported skill directories look inconsistent**

   Runtime messages mention `.agents/skills/` and `~/.agents/skills/`, but the docs only mention `.mastracode/skills/`, `.claude/skills/`, and their global equivalents.

   If `.agents/skills/` is supported, the docs should include it too.

TylerBarnes (changes requested):
Thanks for the updates — the `/skill/<name>` syntax, shared formatter extraction, and async refresh direction are good improvements.

I still see a few issues before this is ready:

1. **Skill activation double-renders in the TUI**

`handleSkillCommand()` creates a `SlashCommandComponent` and inserts it into the chat immediately, then sends the model payload as:

```ts
<skill name="...">
...
</skill>
```

But `render-messages.ts` only has special persisted-message handling for `<system-reminder>` and `<slash-command>`. It does not recognize `<skill>`, so when the persisted user message comes back, the same skill content renders again as raw user text.

Since skills are distinct from slash commands, I don’t think this should be hidden inside the existing `<slash-command>` XML format. It should either:

- add first-class `<skill>` parsing/dedupe/rendering support in `render-messages.ts`, with tests for replaying persisted skill activations, or
- use the appropriate internal/system-reminder-style path if the intent is for skill instructions to affect the agent without appearing as a normal user message.

2. **Skill activation needs to respect the Agent Skills spec’s user-invokable metadata**

The Agent Skills spec defines user-invokable metadata for whether a skill should be directly invokable by the user. This command path needs to use that metadata. Right now it appears to activate any resolved skill via `workspace.skills.get(...)`, and autocomplete appears to include all discovered skills.

Please filter out non-user-invokable skills from:

- `/skill/<name>` execution
- dynamic autocomplete entries
- `/skills` listings, if that list is meant to advertise user-invokable skills

3. **The autocomplete refresh is still partly blocking startup**

The new `refreshSkillsAutocomplete(state).catch(...)` direction is good, but `init()` still awaits `loadCustomSlashCommands()` before installing autocomplete, and that load path can resolve the workspace. That means initial autocomplete can still block on workspace resolution. The initial static autocomplete setup should happen first, then workspace-backed custom command/skill entries should be refreshed lazily once the workspace resolves.

4. **Tests should cover the real persisted/render path**

The current tests cover direct command handling, but not the replay path that causes the double-render. Please add coverage that simulates a persisted `<skill ...>` message returning through `addUserMessage()` and verifies it renders as the intended skill component exactly once, not as raw text.

TylerBarnes (changes requested):
Thanks for the follow-up — the `<skill>` replay/dedupe handling, lazy autocomplete refresh, and direct invocation filtering all look much better.

One remaining issue: `workspace.skills.list()` still drops the `user-invocable` field, so the opt-out won’t work correctly for real `/skills` listings or dynamic autocomplete.

`#parseSkillFile()` preserves the frontmatter field:

```ts
'user-invocable': frontmatter['user-invocable'],
```

and `workspace.skills.get()` should return it for direct `/skill/<name>` execution, but `list()` currently returns:

```ts
{
  name: skill.name,
  path: skill.path,
  description: skill.description,
  license: skill.license,
  compatibility: skill.compatibility,
  metadata: skill.metadata,
}
```

without carrying over:

```ts
'user-invocable': skill['user-invocable'],
```

Since `/skills` and `refreshSkillsAutocomplete()` both filter the result of `workspace.skills.list()`, real opted-out skills can still appear there even though the tests pass with mocked `list()` results that include `'user-invocable': false`.

Please preserve `user-invocable` in `workspace.skills.list()` and add coverage against the real workspace parsing/listing path, not only mocked skill metadata.

TylerBarnes (approved):
Thanks for iterating on this. The remaining `user-invocable` list/autocomplete gap is fixed now, and the earlier concerns around `/skill/<name>` syntax, lazy autocomplete refresh, persisted `<skill>` replay/dedupe, and direct invocation opt-out handling all look addressed.

I also verified the focused core workspace skills coverage, core typecheck/build, mastracode typecheck, prettier, and lint. Approving from my side.

PR #16629, opened 2026-05-15
fix: make brightdata tools bun compatible

### LINKED ISSUE #16627: @mastra/brightdata is unusable under Bun runtime — tool execute throws 'dns is not a function'
## Summary

The `brightdata-search` and `brightdata-fetch` tools created by `createBrightDataTools()` / `createBrightDataSearchTool()` / `createBrightDataFetchTool()` throw synchronously when invoked under the Bun runtime: `TypeError: dns is not a function`.

Root cause is in the underlying `@brightdata/sdk`: its `Transport` constructor calls `interceptors.dns()` from `undici`, which Bun's bundled `undici` polyfill does not expose. Underlying Bun gap: [oven-sh/bun#19748](https://github.com/oven-sh/bun/issues/19748). I'm filing this against `@mastra/brightdata` separately because the integration is presented as drop-in for any agent and Bun is one of Mastra's supported runtimes — today the integration silently fails the first time an agent calls either tool.

## Environment

- `@mastra/brightdata@0.2.0`
- `@mastra/core@1.34.0`
- `@brightdata/sdk@1.1.0` (transitive)
- Bun `1.3.11`
- macOS `arm64`

## Repro

```js
// repro.mjs
import { createBrightDataSearchTool } from "@mastra/brightdata";

process.env.BRIGHTDATA_API_TOKEN = "any-string";
const tool = createBrightDataSearchTool();

try {
  await tool.execute(
    { query: "test" },
    { toolCallId: "t", messages: [], resourceId: "r" },
  );
  console.log("OK — execute returned");
} catch (err) {
  console.log("THREW:", err.message);
}
```

```text
$ bun run repro.mjs
THREW: dns is not a function. (In 'dns()', 'dns' is undefined)

$ node repro.mjs
THREW: invalid API key or insufficient permissions   # SDK constructor + request both succeeded; token rejection is the real API response
```

## Suggested fixes (any one would unblock callers on Bun)

1. **Re-implement against Bright Data's REST API.** Call `https://api.brightdata.com/serp/req` and `https://api.brightdata.com/request` directly from `@mastra/brightdata` using `fetch`, dropping the `@brightdata/sdk` dependency. Sidesteps the Bun gap entirely and removes the transitive `undici` dep.
2. **Add a runtime guard in `getBrightDataClient()`.** Detect Bun and throw a clear error mentioning Bun-incompatibility, instead of letting the obscure `dns is not a function` propagate from inside the SDK.
3. **Document Bun incompatibility.** Add a note to `@mastra/brightdata`'s README and tighten the `engines` field until upstream `@brightdata/sdk` lands a fix.

Rohithmatham12:
## Summary
- replace the Bright Data SDK runtime client with a fetch-based REST client for search and fetch tools
- remove the SDK dependency that triggers Bun's missing `undici.interceptors.dns()` path
- update existing brightdata unit tests for the REST-backed client
- add a patch changeset for `@mastra/brightdata`

Fixes #16627

## Tests
- `./node_modules/.bin/vitest run` from `integrations/brightdata`


Rohithmatham12:
Superseded by #16630 after renaming the source branch to `fix/brightdata-bun-runtime`.

PR #16653, opened 2026-05-15
fix(core): transform <system-reminder> tags for Azure OpenAI to avoid content moderation

### LINKED ISSUE #16531: Observational memory triggers Azure OpenAI content moderation
This issue was created from Discord post 1504029454085652561:

[Open in Discord](https://discord.com/channels/1309558646228779139/1504029454085652561)

Steps to reproduce: run an agent with observational memory on azure openai; trigger OM; then continue; Azure will always respond with "I'm sorry, but I cannot assist with that request.", which is a content moderation premade text.

I suspect it may have something to do with having <system-reminder> in the user message, but not entirely sure.

itskrishna21:
Fixes #16531

Azure's content moderation flags `<system-reminder>` as a prompt-injection signal and returns a canned refusal instead of the model's output. This happens for both Observational Memory continuation hints and temporal gap markers, which both inject `<system-reminder>` tags into user messages.

Added a new `azureSystemReminderTransform` CompatRule in `ProviderHistoryCompat` that renames the tag to `<memory-context>` in the outbound prompt when the resolved model is Azure. The rename is wire-only — persisted history and other providers (Anthropic, distilled models) are unaffected and keep seeing `<system-reminder>` as intended.

Before (sent to Azure):
```
<system-reminder>You are in the middle of an observation...</system-reminder>
```

After (sent to Azure):
```
<memory-context>You are in the middle of an observation...</memory-context>
```

Also exports `isMaybeAzure` for consistency with the existing `isMaybeCerebras` and `isMaybeAnthropic` helpers.

## ELI5

This PR fixes a problem where Azure OpenAI incorrectly rejects messages from an AI library because it thinks a special internal tag (`<system-reminder>`) is a hacking attempt. The fix renames these tags to `<memory-context>` only when sending requests to Azure, so Azure doesn't reject them. The change only affects what gets sent to Azure; everything stored in history and sent to other AI providers stays the same.

---

## Overview

This change fixes Azure OpenAI's false-positive content moderation flagging on Observational Memory continuation hints and temporal gap markers. When the library sends requests containing `<system-reminder>` tags to Azure OpenAI, Azure's content moderation layer treats them as potential prompt injection, causing the model to return canned refusals instead of intended outputs.

The fix implements a wire-only transformation: `<system-reminder>` tags are rewritten to `<memory-context>` only for outbound Azure OpenAI requests, while persisted history and all other providers (Anthropic, distilled models, etc.) continue using the original `<system-reminder>` tags.

## Changes Made

**Core Implementation (`provider-history-compat.ts`):**
- Added `isMaybeAzure()` helper function to detect Azure OpenAI models by string patterns (`azure/<...>` or `azure:<...>`), provider objects, and other model configurations
- Implemented `azureSystemReminderTransform` CompatRule that:
  - Targets user message text parts containing `<system-reminder>` tags
  - Rewrites opening tags to `<memory-context>` while preserving all attributes (e.g., temporal-gap markers)
  - Leaves non-user messages and non-Azure requests unchanged
  - Returns `undefined` when no transformation is needed (avoiding unnecessary mutations)
- Registered the new rule in `DEFAULT_COMPAT_RULES` for automatic application

**Test Coverage (`provider-history-compat.test.ts`):**
- Added `isMaybeAzure` test suite covering model ID patterns, provider-shaped objects, and non-Azure cases
- Added `azureSystemReminderTransform` test suite verifying:
  - Tag rewriting with attribute preservation
  - Correct handling of tags across multiple user messages
  - Non-mutation of original prompt
  - Proper behavior when tags are absent or model is non-Azure
- Added integration tests for `ProviderHistoryCompat.processLLMRequest` with Azure models

**Release Documentation (`.changeset`):**
- Added patch release entry documenting the Azure OpenAI content moderation fix

## Impact

- Resolves issue #16531: Observational memory can now continue on Azure OpenAI without triggering content moderation refusals
- No breaking changes; persisted history format remains unchanged
- Other providers and stored data unaffected
- Exports `isMaybeAzure` and `azureSystemReminderTransform` for potential reuse in other compatibility rules

[Review Change Stack](https://app.coderabbit.ai/change-stack/mastra-ai/mastra/pull/16653)

### DIFF
--- a/.changeset/some-regions-yawn.md
+++ b/.changeset/some-regions-yawn.md
@@ -1,0 +1,5 @@
+---
+'@mastra/core': patch
+---
+
+Fixed Azure OpenAI content moderation false-positives caused by `<system-reminder>` tags in Observational Memory and temporal gap markers. When using Azure OpenAI, these tags are now renamed to `<memory-context>` on the outbound request only — persisted history and other providers are unaffected.
--- a/packages/core/src/processors/provider-history-compat.test.ts
+++ b/packages/core/src/processors/provider-history-compat.test.ts
@@ -4,8 +4,10 @@ import { describe, expect, it } from 'vitest';
 import { MessageList } from '../agent/message-list';
 import {
   anthropicStripForeignReasoningContent,
+  azureSystemReminderTransform,
   cerebrasStripReasoningContent,
   isMaybeAnthropic,
+  isMaybeAzure,
   isMaybeCerebras,
   ProviderHistoryCompat,
 } from './provider-history-compat';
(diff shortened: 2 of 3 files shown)


intojhanurag (changes requested):
This is critical file , we don't consider modification without reproduction of issue . Thanks !

PR #16671, opened 2026-05-16
feat(memory): add observeAttachments filter to ObservationalMemory

CalebBarnes:
## Summary

Adds a new `observeAttachments` option on `ObservationConfig` for Observational Memory. It controls which image/file parts on observed messages get forwarded to the Observer model.

- `true` (default) — forward all attachments (existing behavior).
- `false` — drop all attachments; placeholder text lines (`[Image #1: foo.png]`, etc.) still appear in the Observer transcript.
- `string[]` — case-insensitive mimeType allowlist. Supports exact matches (`'application/pdf'`), wildcard subtypes (`'image/*'`), and bare `'*'`. Empty array drops everything.

```ts
memory: new Memory({
  options: {
    observationalMemory: {
      model: 'deepseek/deepseek-chat',
      observation: {
        // text-only Observer: drop binary attachments, keep placeholder lines
        observeAttachments: false,
      },
    },
  },
}),
```

## Motivation

After signals canonicalized message content to a single `content.parts` projection, observed messages now include image/file parts. If the Observer model is text-only (some DeepSeek endpoints, others), this surfaces as a hard runtime failure like _\"No endpoints found that support image input\"_. Previously, attachments would have been dropped at storage time and the Observer never saw them.

This adds a small, targeted escape hatch so OM stays stable across mixed-modality setups without forcing users onto a multimodal Observer model.

## Implementation

- New `ObserverAttachmentFilter` type and `shouldIncludeObserverAttachment` matcher in `observer-agent.ts`. Glob-style mimeType matching, case-insensitive; falls back to `image/*` for `type: 'image'` parts and image-like `type: 'file'` parts, otherwise `application/octet-stream`.
- Threaded as `attachmentFilter` on `ObserverFormatOptions` and applied in:
  - `formatObserverMessage` for image/file message parts.
  - `extractToolResultAttachments` for hoisted tool-result attachments.
- Placeholder text lines are always rendered — only the raw binary `ObserverInputAttachmentPart` is conditionally dropped.
- Surfaced as `observeAttachments?: boolean | string[]` on `ObservationConfig` and resolved as `true` by default on `ResolvedObservationConfig`. `observer-runner.ts` passes it through on both single-thread and multi-thread observer history calls.
- Not added to `ReflectionConfig` — the Reflector only consumes text summaries.

## Test plan

- New `attachmentFilter` describe in `observational-memory.test.ts` covering: `true`, `false`, `image/*` allowlist, empty array, case-insensitive exact match, and hoisted tool-result filtering.
- `pnpm --filter @mastra/memory test src/processors/observational-memory/__tests__/observational-memory.test.ts` → 442 passed, 1 todo.
- `pnpm --filter @mastra/memory exec tsc --noEmit` → clean.
- `pnpm --filter @mastra/memory build:lib` → clean.
- `pnpm --filter @mastra/memory lint` → clean.

🤖 Generated with [Mastra Code](https://github.com/mastra-ai/mastracode)

## ELI5
This adds a simple setting so the Observer can either receive all attachments, no attachments (but still see readable placeholders), or only an allowlisted set of MIME types — useful when the Observer is text-only or you only want certain file types forwarded.

## Overview
Add observeAttachments to ObservationConfig (default: true) to control which image/file parts observed messages forward to the Observer. Supports boolean or a case-insensitive MIME-type allowlist (string[]), with glob-style matching.

### Configuration & Types
- Added observeAttachments?: boolean | string[] to ObservationConfig.
- Added observeAttachments: boolean | string[] to ResolvedObservationConfig (resolved default: true).
- New exported type: ObserverAttachmentFilter = boolean | string[].

### Implementation
- New matcher shouldIncludeObserverAttachment in observer-agent.ts implementing case-insensitive, glob-style MIME matching:
  - exact (e.g., application/pdf), type/* (e.g., image/*), and '*' (allow-all).
  - Empty array => drop-all.
- Derives effective MIME when missing: image/* for image-like parts (and image-like file parts), otherwise application/octet-stream.
- Threaded attachmentFilter through ObserverFormatOptions and applied in:
  - formatObserverMessage: message-level image/file parts always emit placeholder lines; structured binary ObserverInputAttachmentPart may be omitted when filtered out.
  - extractToolResultAttachments: hoisted tool-result attachments filtered the same way; function tracks whether attachment blocks were present and returns the sanitized mapped value when blocks were encountered to avoid leaking raw base64 payloads while preserving placeholders.
- ObserverRunner.call and callMultiThread now pass observationConfig.observeAttachments into history-message builders so filtering applies to single- and multi-thread observer calls.
- ObservationalMemory resolves observeAttachments default to true when constructing its ResolvedObservationConfig.
- Reflection (Reflector) unchanged (it consumes text summaries only).

### Bugfix
- Fixed tool-result leak: extractToolResultAttachments now tracks attachment blocks and returns sanitized results when any attachment blocks exist, preventing raw payload leaks while keeping placeholders.

### Tests
- New tests in packages/memory/src/processors/observational-memory/__tests__/observational-memory.test.ts covering:
  - true (forward all), false (drop all but keep placeholders), image/* allowlist, empty array (drop-all), '*' allow-all, case-insensitive exact matches, and hoisted tool-result attachment filtering (placeholders preserved; raw payloads not leaked).
- Author reports local memory tests green and tsc/build/lint checks clean.

### Documentation
- Updated reference docs: docs/src/content/en/reference/memory/observational-memory.mdx (ObservationConfig docs).
- Updated feature guide: docs/src/content/en/docs/memory/observational-memory.mdx (new paragraph + example showing observeAttachments: false).
- Added changeset: .changeset/observer-attachment-filter.md.

### Scope
- Affects ObservationalMemory observer flows only; Reflector/Reflection unchanged.

### Reviewer notes
- Placeholders (text lines like “[Image `#1`: foo.png]”) are always emitted; only the structured/binary attachment parts may be dropped.
- Verify regression test for tool-result leak and that hoisted attachments do not expose raw payloads when filtered.

[Review Change Stack](https://app.coderabbit.ai/change-stack/mastra-ai/mastra/pull/16671?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


TylerBarnes (changes requested):
From Mastra Code: I found one issue with the tool-result attachment filtering path.

In `extractToolResultAttachments`, attachment blocks are mapped to placeholders in `newValue`, but the function returns the original `result` whenever `attachments.length === 0`:

`packages/memory/src/processors/observational-memory/observer-agent.ts:829-831`

That works when no attachment blocks were found, but it breaks when attachment blocks were found and all of them were intentionally excluded by `observeAttachments: false`, `[]`, or a non-matching allowlist. In those cases, the placeholder-substituted `newValue` is discarded, so `formatToolResultForObserver` can still receive the original tool-result payload and potentially include base64/file data in the Observer text.

The existing test covers partial filtering (`['image/*']`) where at least one attachment is still forwarded, so `attachments.length > 0` and the sanitized path is used. I’d add coverage for a filter that excludes all hoisted tool-result attachments.

Suggested direction: track whether any attachment block was encountered separately from whether any raw attachment was included, and return `{ ...record, value: newValue }` whenever attachment blocks were encountered.

One smaller docs request: this updates the reference page, but the regular Observational Memory docs should mention the new option too, likely in `docs/src/content/en/docs/memory/observational-memory.mdx`, so users looking at the feature guide see how to use `observeAttachments` with text-only Observer models.

CalebBarnes:
Hey @TylerBarnes — couple of these crossed wires with my push, so quick status:

**Tool-result leak fix**: agreed, exactly the same bug CodeRabbit flagged. Already fixed in `2a63c93588` (the commit your review is on) — `extractToolResultAttachments` now tracks `hadAttachmentBlocks` separately from `attachments.length` and returns the sanitized `newValue` whenever any attachment block was encountered. Added a regression test that asserts the raw base64 payload is gone (and the placeholders are present) when `observeAttachments: false` drops every hoisted tool-result attachment.

**Feature-guide docs**: good catch, added in `44b781dc14`. New paragraph in `docs/src/content/en/docs/memory/observational-memory.mdx` right after the existing attachment-forwarding paragraph, plus a full `new Agent({ ... new Memory({ ... observation: { observeAttachments: false } }) })` example. Reference page (`reference/memory/observational-memory.mdx`) was already updated earlier.

Memory unit suite still green (444/444 + 1 todo). Ready for another look when you have a sec.

TylerBarnes (approved):

PR #16675, opened 2026-05-16
fix(core): key suspendedTools/pendingToolApprovals by toolCallId

### LINKED ISSUE #16468: [BUG] same-tool parallel suspends always fail on the second resume
### Describe the Bug

If an LLM emits two parallel `tool_use` blocks targeting the same
workflow tool, both calls correctly suspend — but the second
`agent.resumeStream({ runId, toolCallId })` always fails with:

```
Error: This workflow run was not suspended
```

Single suspends work fine. Parallel suspends across *different* tools
work fine. It's specifically the case of two (or more) suspended calls
to the same tool name in the same run that breaks. The
[parallel suspend/resume docs](https://mastra.ai/docs/workflows/suspend-and-resume#parallel-suspended-steps)
only show distinct named steps, so this case isn't covered by the
existing tests or examples.

I'm seeing this in production with Gemini 3 Flash via Vertex AI on a
user request that the model fans out into two `tool_use` blocks for
the same workflow tool. After the second resume fails, the agent loop
advances to the next LLM call with one `tool_use` having no matching
`tool_result`, and Vertex returns a 400 (`Unable to submit request
because at least one contents field is required`). Other providers
will probably reject the same shape too.

It looks like this was introduced in
[#11157](https://github.com/mastra-ai/mastra/pull/11157), the
"Agent tool resumption" PR that added `autoResumeSuspendedTools`. That
PR changed the suspension bookkeeping in
`packages/core/src/loop/workflows/agentic-execution/tool-call-step.ts`
from being keyed on `toolCallId` to being keyed on `toolName`:

```ts
// Before #11157:
metadata.pendingToolApprovals[toolCallId] = { ... };

// After #11157 (current):
// Note: We key by toolName rather than toolCallId to track one suspension
// state per unique tool.
metadata.suspendedTools[toolName] = {
  toolCallId,
  toolName,
  suspendedToolRunId,
  suspendPayload,
  resumeSchema,
};
```

I'm guessing the keying change was so `autoResumeSuspendedTools` could
look up "the pending suspend for tool X" without needing a call id —
makes total sense for that feature, but it lost the ability to tell
two parallel suspends of the same tool apart, and the second one's
state silently gets clobbered by the first.

### Step-by-step failure

1. **Suspend A** writes `metadata.suspendedTools['workflow-sendEmail'] = { toolCallId: 'A', suspendedToolRunId: <run-A>, ... }`.
2. **Suspend B** overwrites the same slot with `{ toolCallId: 'B', suspendedToolRunId: <run-B>, ... }`. A's entry is gone.
3. **Resume A** with `toolCallId: 'A'`. The metadata lookup returns B's
   entry, so the resume path falls back to scanning message parts for
   the suspended part with `toolCallId === 'A'` and `resumed !== true`.
   That scan succeeds — A's inner `runId` is on the part — and the
   inner workflow resumes correctly.
4. After A completes, `removeToolMetadata` runs and does two things:
    - Deletes the only `metadata.suspendedTools['workflow-sendEmail']`
      entry — so B's bookkeeping is gone for good.
    - Iterates **every** suspended part with the matching `toolName`
      and marks them `resumed: true` — including B's part, which
      hasn't actually been resumed.
5. **Resume B** with `toolCallId: 'B'`. Metadata lookup misses. The
   part-scan finds B's part but skips it because `resumed: true`. The
   code falls through to `runIdToUse = randomUUID()` in `agent.ts`
   (~line 4463) and calls `workflow.resume(<random uuid>, ...)`.
   `workflow.ts:_resume` finds no snapshot for that UUID and throws
   `"This workflow run was not suspended"`.

### Steps To Reproduce

See the attached gist. In a repo w/ the requisite dependencies installed this should show this output on the second resume:

```
tool-error: {
  toolCallId: 'call-B',
  toolName: 'workflow-sendEmail',
  error: Error: [Agent:Repro Agent] - Failed workflow tool execution
    cause: Error: [Agent:Repro Agent] - Failed workflow tool execution
      cause: Error: This workflow run was not suspended
}
```

https://gist.github.com/rmadsen/d553a8aa2c2ad9aea8bb84a4634c61c8

### Expected Behavior

I should be able to approve one, then the next, tool calls that were issued in parallel.

### Environment Information

```block
Mastra @1.24.0, run against Gemini 3 flash via Vertex, but any LLM provider that can issue parallel tool calls should reproduce
```

### Verification

truffle-dev:
## Description

When an LLM emitted two `tool_use` blocks for the same tool in one assistant turn and both suspended, the second entry silently overwrote the first in `metadata.suspendedTools` (or `metadata.pendingToolApprovals`) because the record was keyed by `toolName`. Only one of the two calls could then be resumed; the other stayed orphaned, even after restart.

The fix keys those records by `toolCallId` so each parallel suspend remains a distinct entry. Readers iterate `Object.values()`, so runs persisted before this change keep resolving through a backward-compat lookup (`findEntryKey` scans values when the direct `[toolCallId]` lookup misses an older `toolName`-keyed write). `removeToolMetadata` clears only the matching `toolCallId` and marks only the matching `data-tool-call-suspended` (or `data-tool-call-approval`) part as `resumed: true`, leaving sibling suspends with their `resumed: false` flag intact so the autoResumeSuspendedTools loop can still find them.

Two regression tests were added: one at the adapter layer asserting the AIV5 adapter synthesizes one `data-tool-call-suspended` part per `toolCallId` when two parallel same-tool suspends exist in metadata, and one at the step layer asserting `addToolMetadata` writes two distinct entries to `metadata.suspendedTools` when the same tool is called twice in parallel.

## Related issue(s)

Fixes #16468.

## Type of change

## Checklist

## ELI5

When the AI tried to use the same tool twice at once, the second call erased the first one’s record so it couldn't be resumed. This PR tracks each call by a unique ID so parallel calls are tracked and resumed independently.

---

## Summary

Fixes a bug where parallel suspensions/approvals for the same tool in a single assistant turn overwrote each other because metadata was keyed by toolName. Tool suspension and approval records are now keyed by toolCallId so each parallel call becomes a distinct entry and can be resumed individually. Backward-compatible fallbacks let the system still find older toolName-keyed entries.

## Changes

- tool-call-step.ts
  - Key suspendedTools and pendingToolApprovals by toolCallId instead of toolName.
  - removeToolMetadata now deletes only the matching toolCallId entry and marks only the corresponding part as resumed, leaving sibling suspends intact.
  - Lookup logic updated to prefer direct toolCallId lookup with a fallback scan of values to support older toolName-keyed data.
  - Call sites updated to pass { toolCallId } where applicable.
  - Resume flow updated to derive suspended run IDs from the new shape, with compatibility fallback.

- llm-execution-step.ts
  - Auto-resume prompt augmentation now constructs suspended-tool objects keyed by toolCallId during legacy parts-only fallback.

- Tests
  - Added adapter-layer Vitest ensuring AIV5Adapter synthesizes one data-tool-call-suspended part per distinct toolCallId for parallel same-tool suspends.
  - Added step-layer Vitest verifying addToolMetadata writes two distinct suspendedTools entries keyed by toolCallId for parallel calls.

- Changeset
  - .changeset entry documenting the user-facing fix and compatibility behavior.

## Backward compatibility

- Readers already iterate Object.values(), so they remain compatible with either keying scheme.
- removeToolMetadata includes a fallback that scans stored values for a matching toolCallId when direct lookup by toolCallId fails, preserving resume behavior for runs persisted before this change.

## Impact

- Parallel suspends for the same tool no longer clobber each other.
- Each suspended tool call can be resumed independently via agent.resumeStream({ runId, toolCallId }).
- Auto-resume and existing persisted runs continue to work due to compatibility fallbacks.

[Review Change Stack](https://app.coderabbit.ai/change-stack/mastra-ai/mastra/pull/16675?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/.changeset/parallel-same-tool-suspend.md
+++ b/.changeset/parallel-same-tool-suspend.md
@@ -1,0 +1,9 @@
+---
+'@mastra/core': patch
+---
+
+Fixed parallel suspends of the same tool overwriting each other's bookkeeping. When an LLM emitted two `tool_use` blocks for the same tool in one assistant turn and both suspended, the second entry replaced the first in `metadata.suspendedTools` / `metadata.pendingToolApprovals` and only one of the two calls could be resumed.
+
+`addToolMetadata` now keys those records by `toolCallId` instead of `toolName`, so each parallel suspend remains a distinct entry. Readers iterate `Object.values()`, so older runs that persisted entries keyed by `toolName` continue to resolve through a backward-compat fallback. The AIV5 adapter renders one `data-tool-call-suspended` (or `data-tool-call-approval`) part per call, and `removeToolMetadata` clears only the matching `toolCallId` so parallel siblings keep their `resumed: false` flag.
+
+Fixes #16468.
--- a/.changeset/parallel-same-tool-suspend.md
+++ b/.changeset/parallel-same-tool-suspend.md
@@ -2,8 +2,8 @@
 '@mastra/core': patch
 ---
 
-Fixed parallel suspends of the same tool overwriting each other's bookkeeping. When an LLM emitted two `tool_use` blocks for the same tool in one assistant turn and both suspended, the second entry replaced the first in `metadata.suspendedTools` / `metadata.pendingToolApprovals` and only one of the two calls could be resumed.
+Fixed a bug where calling the same tool twice in parallel within a single assistant turn would only let you resume one of the two calls. The other call became orphaned and could never be resumed, even after a restart.
 
-`addToolMetadata` now keys those records by `toolCallId` instead of `toolName`, so each parallel suspend remains a distinct entry. Readers iterate `Object.values()`, so older runs that persisted entries keyed by `toolName` continue to resolve through a backward-compat fallback. The AIV5 adapter renders one `data-tool-call-suspended` (or `data-tool-call-approval`) part per call, and `removeToolMetadata` clears only the matching `toolCallId` so parallel siblings keep their `resumed: false` flag.
+Each suspended tool call (and each pending approval) is now tracked separately, so all parallel calls to the same tool can be resumed independently. Runs that were persisted before this fix continue to resume correctly.
 
 Fixes #16468.
--- a/packages/core/src/agent/message-list/adapters/AIV5Adapter-suspended-parts.test.ts
+++ b/packages/core/src/agent/message-list/adapters/AIV5Adapter-suspended-parts.test.ts
@@ -225,6 +225,74 @@ describe('AIV5Adapter — suspended tool state rehydration', () => {
     expect(approvalParts).toHaveLength(1);
   });
 
+  it('should produce one data-tool-call-suspended part per toolCallId when the same tool suspends twice in parallel', () => {
+    // Regression test for mastra-ai/mastra#16468 — when an LLM emits two parallel tool_use
+    // blocks for the same tool, addToolMetadata used to key the metadata record by toolName
+    // and silently overwrote the first entry. Writes now key by toolCallId so both suspends
+    // remain discoverable. The adapter should render one suspended part per distinct call.
+    const dbMessage: MastraDBMessage = {
+      id: 'msg-parallel',
+      role: 'assistant',
+      createdAt: new Date('2024-01-01'),
+      content: {
+        format: 2,
+        parts: [
+          {
+            type: 'tool-invocation',
+            toolInvocation: {
+              toolCallId: 'tc-parallel-A',
+              toolName: 'workflow-sendEmail',
+              args: { to: 'a@example.com' },
+              state: 'call',
+            },
+          },
+          {
+            type: 'tool-invocation',
+            toolInvocation: {
+              toolCallId: 'tc-parallel-B',
+              toolName: 'workflow-sendEmail',
+              args: { to: 'b@example.com' },
+              state: 'call',
+            },
+          },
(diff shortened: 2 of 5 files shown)


truffle-dev:
On the linked-issue ask: no pre-existing issue captures this bug — happy to file one if that is the preferred path. The PR body documents the failure mode in detail (parallel `tool_use` collision in `metadata.suspendedTools` / `metadata.pendingToolApprovals`, keyed by `toolName` so the second suspend silently overwrites the first).

truffle-dev:
Correction to my comment above: the PR body has linked Fixes #16468 since open, and that issue (reported 2026-05-12, still open) is exactly this bug. My earlier reply misstated that no issue existed.

The branch is still mergeable against main and CodeRabbit had no actionable comments. Happy to rebase if that would help a review.

truffle-dev:
Closing this — #18041 (merged 2026-06-23) implements the same fix and supersedes it. Same root cause (suspended-tool tracking keyed by tool name, so two parallel calls to the same tool collapsed into one entry and overwrote each other's suspended run id) and the same resolution (key the tracking by `toolCallId`), and the merged version also handles the legacy toolName-keyed fallback for pre-upgrade persisted metadata, which this PR didn't. Issue #16468 looks resolved on `main` by that change. Thanks to the team for landing it.

PR #16679, opened 2026-05-16
fix(core): continue durable streams after tool calls

### LINKED ISSUE #16277: `createEventedAgent` / `createDurableAgent`: does the wrapper support multi-step tool-call iteration, or does it only run one workflow turn per `.stream()`?
### Summary

Wrapping an `Agent` with `createEventedAgent({ agent })` or `createDurableAgent({ agent })` and calling `.stream(messages, { maxSteps: 4 })` against a tool-call-then-text prompt produces **only the first tool-call step**. The agent loop does not iterate to the second LLM call that would consume the tool result and emit a final text response. `agent.stream({ maxSteps: 4 })` on the same base agent and same prompt iterates correctly.

I'd like to confirm whether this is:

1. **Expected** — durable agents intentionally run one workflow turn per `.stream()`, and multi-step requires `createInngestAgent` or some other configuration we're missing.
2. **A bug** — the wrapper is supposed to iterate the agent loop, and our usage triggers an edge case.
3. **A docs gap** — multi-step works under specific conditions (e.g., requires `BackgroundTaskManager` + memory scope) that aren't surfaced in the [PR #12557](https://github.com/mastra-ai/mastra/pull/12557) description.

### Why I think this is worth filing

The [PR #12557 description](https://github.com/mastra-ai/mastra/pull/12557) describes durable agents as solving "long-running operations" where "agent loops with tool calls can take minutes." That implies the wrapper runs the full agent loop including multi-step iteration.

Looking at the shipped tests under `packages/core/src/agent/durable/__tests__/`:

- `durable-agent-stopwhen.test.ts` defines a `createTextThenToolModel` mock and tests like "should handle stopWhen with text and tool calls" — but every test in that file just asserts `runId` is defined and calls `cleanup()`. The streams are never drained to verify the loop iterates.
- `durable-agent-tools.test.ts` defines a `_createToolCallThenTextModel` helper (with leading underscore) for tool-call → text iteration. It does not appear to be used. The `should support maxSteps configuration` test asserts only that `prepare()` stores `workflowInput.options.maxSteps`, not that 4 steps actually run.
- The official example at `examples/durable-agents/src/mastra/agents/research-agent.ts` defines an agent whose instructions explicitly require multi-step ("1. Use the web-search tool / 2. Analyze the search results / 3. Provide a clear, well-organized summary") and wraps it three ways (regular, durable, evented, inngest). I couldn't find evidence in the test suite or example that the durable/evented variants are validated end-to-end against the same multi-step output as the regular variant.

So I want to make sure we're not building on an assumption that isn't actually guaranteed.

### Minimal repro

Self-contained, uses `MockLanguageModelV2` (no provider deps). The mock returns a tool-call on the first invocation, then text on the second. Regular `Agent.stream({ maxSteps: 4 })` produces 2 steps; `createEventedAgent({ agent }).stream({ maxSteps: 4 })` produces 1.

```ts
import { createEventedAgent, createDurableAgent } from '@mastra/core/agent/durable';
import { MockLanguageModelV2, convertArrayToReadableStream } from '@internal/ai-sdk-v5/test';
import { z } from 'zod';

// Mock that returns tool-call on call 1, text on call 2 (multi-step happy path)
function createToolThenTextModel() {
  let callCount = 0;
  return new MockLanguageModelV2({
    doStream: async () => {
      callCount += 1;
      if (callCount === 1) {
        return {
          stream: convertArrayToReadableStream([
            { type: 'stream-start', warnings: [] },
            { type: 'response-metadata', id: 'r1', modelId: 'mock', timestamp: new Date(0) },
            {
              type: 'tool-call',
              toolCallId: 'c1',
              toolName: 'echo',
              input: JSON.stringify({ text: 'hello' }),
              providerExecuted: false,
            },
            { type: 'finish', finishReason: 'tool-calls', usage: { inputTokens: 10, outputTokens: 10, totalTokens: 20 } },
          ]),
          rawCall: { rawPrompt: null, rawSettings: {} },
        };
      }
      return {
        stream: convertArrayToReadableStream([
          { type: 'stream-start', warnings: [] },
          { type: 'response-metadata', id: 'r2', modelId: 'mock', timestamp: new Date(0) },
          { type: 'text-start', id: 't1' },
          { type: 'text-delta', id: 't1', delta: 'tool returned hello' },
          { type: 'text-end', id: 't1' },
          { type: 'finish', finishReason: 'stop', usage: { inputTokens: 15, outputTokens: 10, totalTokens: 25 } },
        ]),
        rawCall: { rawPrompt: null, rawSettings: {} },
      };
    },
  });
}

const echoTool = createTool({
  id: 'echo',
  description: 'Echo back the input.',
  inputSchema: z.object({ text: z.string() }),
  outputSchema: z.object({ echoed: z.string() }),
  execute: async ({ context }) => ({ echoed: context.text }),
});

async function drain(stream: ReadableStream<any>): Promise<{ types: Record<string, number>; finish: string | null }> {
  const reader = stream.getReader();
  const types: Record<string, number> = {};
  let finish: string | null = null;
  while (true) {
    const { done, value } = await reader.read();
    if (done) break;
    types[value.type] = (types[value.type] ?? 0) + 1;
    if (value.type === 'finish') finish = value.payload?.stepResult?.reason ?? 'unknown';
  }
  return { types, finish };
}

// Run both
const baseAgent = new Agent({
  id: 'multi-step-test',
  instructions: 'Call echo, then describe what came back.',
  model: createToolThenTextModel(),
  tools: { echo: echoTool },
});

const baselineHandle = await baseAgent.stream('Run echo with "hello".', { maxSteps: 4 });
const baseline = await drain(baselineHandle.fullStream);

const eventedAgent = createEventedAgent({
  agent: new Agent({
    id: 'multi-step-test-evented',
    instructions: 'Call echo, then describe what came back.',
    model: createToolThenTextModel(),
    tools: { echo: echoTool },
  }),
});

const wrapperHandle = await eventedAgent.stream('Run echo with "hello".', { maxSteps: 4 });
const wrapper = await drain(wrapperHandle.fullStream);

console.log('baseline types:', baseline.types, 'finish:', baseline.finish);
console.log('wrapper  types:', wrapper.types, 'finish:', wrapper.finish);

// Expected (multi-step works):
//   baseline types include text-start/text-delta/text-end, finish='stop'
//   wrapper  types include text-start/text-delta/text-end, finish='stop'
//
// Actually observed:
//   baseline types include text-start/text-delta/text-end, finish='stop' ✓
//   wrapper  types do NOT include text-* chunks, finish='tool-calls'    ✗
```

### Production evidence

We ran the same comparison against real models through LiteLLM (no mocks). 4 of 4 production-critical model families regressed identically when wrapped:

| Model | baseline chunks | wrapper chunks | wrapper finish | wrapper text chunks |
|---|---|---|---|---|
| `anthropic/claude-opus-4-7` | 22 | 12 | `tool-calls` | 0 |
| `openai/gpt-5` | 52 | 14 | `tool-calls` | 0 |
| `anthropic/claude-haiku-4-5` | 36 | 10 | `tool-calls` | 0 |
| `anthropic/claude-sonnet-4-5` | 34 | 16 | `tool-calls` | reduced (1/2/1 vs 2/6/2) |

The wrapper consistently:
- Drops framework lifecycle chunks (`start`, `step-start`, `step-finish`).
- Drops `text-*` content chunks on Anthropic models.
- Adds a wrapper-specific `response-metadata` chunk.
- Has roughly half the chunk count of baseline.
- Finishes with reason `tool-calls` instead of `stop`.

### What we tried

- `eventedAgent.stream(messages, { maxSteps: 4 })` — single turn, ignores `maxSteps`.
- `eventedAgent.streamUntilIdle(messages, { maxSteps: 4 })` — identical output to `.stream()`. We found in `runDurableStreamUntilIdle` (`packages/core/src/agent/durable/durable-stream-until-idle.ts`) that it short-circuits to `agent.stream(messages, restStreamOptions)` when `!deps.bgManager || !scope`. Since we don't have a `BackgroundTaskManager` configured, this is the no-op path.
- `agent.stream(messages, { maxSteps: 4 })` (regular Agent, not wrapped) — works correctly: 2 steps, final text emitted, finish `stop`.

### Mastra version

`@mastra/core@1.31.0`.

### Questions

1. Is multi-step iteration in `createEventedAgent` / `createDurableAgent` supposed to work without a `BackgroundTaskManager` configured?
2. If yes, what's the suggested next step to debug? (Happy to send a PR with a real multi-step test like the one above against the durable variants if it'd help.)
3. If no — i.e., multi-step requires `BackgroundTaskManager` + memory scope or `createInngestAgent` — would you be open to a docs PR clarifying that on the [durable agents](https://github.com/mastra-ai/mastra/pull/12557) example README and the `createEventedAgent` reference page (which currently 404s at `https://mastra.ai/reference/agents/createEventedAgent`)? That'd save downstream consumers a fair amount of investigation time.

Thanks!

---

pragnyanramtha:
## Summary

Fixes #16277.

Durable and evented agent streams were stopping after the first tool-call step even when `maxSteps` allowed another model call. The mapping step appended tool results as a separate assistant message instead of completing the original assistant tool invocation, so the next LLM step did not receive the expected tool-result prompt shape.

This PR:

- merges durable tool results back into the matching assistant `tool-invocation` via `MessageList.updateToolInvocation`
- falls back to adding a tool-result message only when no matching invocation is present
- carries LLM text through durable step output so final stream output has the final text
- passes the current iteration as `stepIndex` for durable LLM execution
- adds a regression test covering both `createDurableAgent` and `createEventedAgent`, including an assertion that the second model prompt contains the tool result
- adds a patch changeset for `@mastra/core`

## Validation

Passed locally:

- `pnpm --filter @mastra/core test src/agent/durable/__tests__/durable-agent-stream.test.ts` (19 passed, no type errors)
- `pnpm --filter @mastra/core test src/agent/message-list/tests/update-tool-invocation.test.ts src/agent/message-list/tests/message-list-om-multistep.test.ts` (15 passed, no type errors)
- `pnpm build:core`
- `git diff --check`

Additional check:

- `pnpm --filter @mastra/core test src/agent/durable/` currently fails on `src/agent/durable/__tests__/resume-api.test.ts > should replay cached events on resume subscription` because cached event history is empty. I reproduced the same failure on a clean `origin/main` worktree at `49a9c88`, so this appears unrelated to this patch.

## ELI5

When an agent asked a tool for information and needed that answer to continue, the durable/evented stream stopped too early. This PR makes the agent put the tool's answer back into the original tool-invocation so the agent can continue and produce the final reply.

---

## Summary

Fixes `#16277`: durable and evented agent streams could stop after the first tool-call step even when maxSteps allowed further iterations. Root cause: tool results were appended as a separate assistant message previously, so the next LLM prompt lacked the expected tool-result shape and iteration stopped.

## Changes

Core fixes
- Merge durable tool results into the matching assistant `tool-invocation` via `MessageList.updateToolInvocation`, falling back to adding a `tool`-role `tool-result` message only when no matching invocation exists (packages/core/src/agent/durable/workflows/steps/llm-mapping.ts).
- Preserve and surface LLM-generated text from execution by adding `text?: string` to `DurableLLMStepOutput` and populating it from streamed text deltas (packages/core/src/agent/durable/types.ts and packages/core/src/agent/durable/workflows/steps/llm-execution.ts).
- Pass the current iteration as `stepIndex` into durable LLM execution input so the execution receives the correct step index (packages/core/src/agent/durable/workflows/create-durable-agentic-workflow.ts).
- Ensure durable mapping sets final output.text = llmOutput.text so final iteration output includes final text (packages/core/src/agent/durable/workflows/steps/llm-mapping.ts).
- Add a patch changeset for `@mastra/core` (.changeset/dark-words-greet.md).

Tests
- Added a regression test that uses a two-step mock model (tool-call then text) and asserts both createDurableAgent and createEventedAgent continue to the second LLM call: it records prompts, verifies the second prompt contains the tool result, collects stream chunks (tool-call, tool-result, text-delta), and asserts final finish chunk contains the final text (packages/core/src/agent/durable/__tests__/durable-agent-stream.test.ts).

## Validation

- Ran targeted suites locally: durable-agent-stream tests and message-list tests passed (no type errors).
- Core build and type checks passed (`pnpm --filter ./packages/core check`, `pnpm build:core`).
- Noted an unrelated failing durable test ("resume subscription") reproducible on clean origin/main; considered unrelated to this change.

[Review Change Stack](https://app.coderabbit.ai/change-stack/mastra-ai/mastra/pull/16679?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


pragnyanramtha:
The branch is refreshed to current `main` at `b5c16a6b`. I do not see any remaining CodeRabbit/Copilot action items on my side.

The visible blockers now look repo-controlled: Vercel authorization plus required `Combined store Tests (vector+storage)`, `E2E Tests`, and `Memory Tests` contexts that are still pending without target URLs. Could a maintainer rerun or refresh those pending required checks when convenient?

taofeeq-deru (approved):

Issue #16645, opened 2026-05-15
[FEATURE] Final Bundle Includes Tree-Shaken Dependencies


sgarfinkel:
### Describe the Bug

We are using postgresql for production mode and libsql for dev mode. The imports of these libraries are conditional around `process.env.NODE_ENV`. The final built files properly tree-shakes the unused import of libsql, but the process to generate the `package.json` persistently includes it. This is causing our production builds to be polluted with native binaries that we never even use.

### Steps To Reproduce

```ts
const getStorage = async () => {
  if(process.env.NODE_ENV === "development") {
    const { LibSQLStore } = await import('@mastra/libsql');
    return new LibSQLStore({ id: "libsql", url: "file:storage.db" });
  } else {
    const { PostgresStore } = await import('@mastra/pg');
    return new PostgresStore({
      id: 'pg-storage',
      connectionString: process.env.DATABASE_URL,
    })
  }
}
```

### Link to Minimal Reproducible Example

n/a

### Expected Behavior

The `package.json` should include ONLY the packages used in the transpiled code _and_ any `dynamicPackages` registered. It would also be helpful if an escape hatch could be added to force exclude a package in case scanning messes up (also not available at the moment).

### Environment Information

```block
System:
    OS: macOS 15.7.5
    CPU: (11) arm64 Apple M3 Pro
    Memory: 552.56 MB / 36.00 GB
    Shell: 4.3.3 - /opt/homebrew/bin/fish
  Binaries:
    Node: 22.22.3 - ~/.local/share/mise/installs/node/22.22.3/bin/node
    Yarn: 1.22.22 - ~/.local/share/mise/installs/npm-yarn/1.22.22/bin/yarn
    npm: 10.9.8 - ~/.local/share/mise/installs/node/22.22.3/bin/npm
    pnpm: 10.33.4 - ~/.local/share/mise/installs/pnpm/10.33.4/pnpm
    bun: 1.3.11 - ~/.local/share/mise/installs/bun/1.3.11/bin/bun
```

### Verification

- [x] I have searched the existing issues to make sure this is not a duplicate
- [x] I have included sufficient information for the team to reproduce and understand the issue

daneatmastra:
Thank you for reporting this issue! We have labeled it for trio-tnt and we will look into it as soon as possible.

[...]

PR #16680, opened 2026-05-16
feat(deployer): add bundler.excludePackages option to drop entries from generated package.json

shiminshen:
## Summary

Relates to #16645 (escape-hatch portion; the scanner-reachability half stays open).

Dependency analysis is intentionally conservative — packages reachable via conditional dynamic imports (e.g. `process.env.NODE_ENV`-gated branches) get flagged even when tree-shaking removes them from the production bundle. The user's report is the canonical example: dev/prod-branched `await import('@mastra/libsql')` / `await import('@mastra/pg')` leaves both packages in `.mastra/output/package.json`, polluting production installs with native binaries that never run.

## What this PR does

Adds a `bundler.excludePackages: string[]` opt-out on `BundlerConfig`:

```typescript
  bundler: {
    excludePackages: ['@mastra/libsql'],
  },
});
```

When the bundler builds `dependenciesToInstall`, entries whose import name is in `excludePackages` are dropped before the output `package.json` is written. The exclusion is logged at debug level so users can confirm the filter triggered.

## What this PR does NOT do

The issue also asks the scanner to **stop including** tree-shaken deps in the first place. That's a real bug and worth fixing — but the cause is deeper: `bundleExternals` declares every dep in `depsToOptimize` as its own Rollup entry, so each external gets its own entry chunk regardless of whether it's reachable from the user's actual entry after tree-shaking. The `o.imports` walk in `analyze.ts` then picks the dep up from that virtual entry chunk.

Fixing it cleanly needs reachability analysis on the post-tree-shake bundle that distinguishes user-entry chunks (the real ones) from the virtual entry chunks we created per external (the ones to skip). That's a larger change with bundler-correctness implications I'd rather not bundle into this PR. The escape hatch unblocks the immediate user pain; happy to follow up with the reachability fix in a separate PR.

## Test plan

- [x] `bundlerOptions.test.ts` — added a `with-exclude-packages.js` fixture and a test row verifying the option flows through the bundler-options extractor. New snapshot captured. 11 tests pass.
- [x] All other `packages/deployer/src/build/**` and `packages/deployer/src/bundler/**` tests pass (273 tests, 19 files) — unchanged behavior when `excludePackages` is unset.
- [x] `pnpm exec tsc --noEmit` in `packages/deployer` reports the same pre-existing `@mastra/hono` / `@mastra/server` errors (unrelated to this change) and no new errors. After `pnpm --filter @mastra/core` builds, `excludePackages` resolves cleanly against `BundlerConfig`.

Pre-existing env failures in `src/server/**` and `src/validator/**` tests (require dist of `@mastra/hono`, `@mastra/server`, `dist/validator/loader.js`) are unrelated to this change.

## Docs

JSDoc on `BundlerConfig.excludePackages` (in `packages/core/src/bundler/types.ts`) and `BundlerOptions.excludePackages` (in `packages/deployer/src/build/types.ts`) is the primary source. The reference doc (`docs/src/content/en/reference/core/mastra-class.mdx`) needs a small follow-up to list the option alongside `dynamicPackages` — the local prettier-mdx hook was failing on my machine and I didn't want to block this PR on the docs build env.

---
Written by an agent (Claude Code, claude-opus-4-7).

## ELI5

This PR lets developers remove specific packages from the generated `package.json` when the bundler incorrectly thinks they are needed. It helps with packages used only by code that the production build removes.

## Overview

Adds the optional `bundler.excludePackages: string[]` configuration. The deployer applies exclusions to the final dependency map before writing `.mastra/output/package.json`.

The option supports packages detected through conditional dynamic imports. It does not fix the underlying scanner reachability issue.

## Changes

- Adds `excludePackages` to `BundlerConfig` and `BundlerOptions`.
- Propagates the option through the bundling pipeline.
- Filters excluded packages after dependency analysis completes.
- Prevents workspace-linked packages from being re-added during dependency processing.
- Logs excluded packages at debug level.
- Uses exact package-name matching. Subpaths do not match.
- Keeps excluded packages external and omits them only from `package.json` dependencies.
- Keeps excluded workspace packages in `resolutions` and continues packing them with `packWorkspaceDependencies`.
- Adds JSDoc, reference documentation, a changeset, a fixture, and package-generation tests.

## Usage

```ts
export default {
  bundler: {
    excludePackages: ['`@mastra/libsql`'],
  },
};
```

Incorrect exclusions can cause runtime `MODULE_NOT_FOUND` errors.

## Validation

- 458 tests pass.
- One pre-existing `fs-routing` test fails.
- Tests cover excluded packages, unset configuration, and workspace dependency re-add behavior.

### DIFF
--- a/.changeset/exclude-packages-bundler-16645.md
+++ b/.changeset/exclude-packages-bundler-16645.md
@@ -1,0 +1,14 @@
+---
+'@mastra/core': minor
+'@mastra/deployer': minor
+---
+
+Added a `bundler.excludePackages` option to force-exclude packages from the generated `package.json` even when dependency analysis flagged them as in use. Useful when conditional dynamic imports (e.g. a dev-only `await import('@mastra/libsql')` gated by `process.env.NODE_ENV`) are tree-shaken out of the production bundle but still surface as dependencies. Addresses the "escape hatch" half of [#16645](https://github.com/mastra-ai/mastra/issues/16645).
+
+```typescript
+export const mastra = new Mastra({
+  bundler: {
+    excludePackages: ['@mastra/libsql'],
+  },
+});
+```
--- a/docs/src/content/en/reference/core/mastra-class.mdx
+++ b/docs/src/content/en/reference/core/mastra-class.mdx
@@ -153,9 +153,10 @@ Visit the [Configuration reference](/reference/configuration) for detailed docum
     name: 'bundler',
     type: 'BundlerConfig',
     description:
-      'Configuration for the asset bundler with options for externals, sourcemap, transpilePackages, and dynamicPackages.',
+      'Configuration for the asset bundler with options for externals, sourcemap, transpilePackages, dynamicPackages, and excludePackages.',
     isOptional: true,
-    defaultValue: '{ externals: [], sourcemap: false, transpilePackages: [], dynamicPackages: [] }',
+    defaultValue:
+      '{ externals: [], sourcemap: false, transpilePackages: [], dynamicPackages: [], excludePackages: [] }',
   },
   {
     name: 'scorers',
--- a/packages/core/src/bundler/types.ts
+++ b/packages/core/src/bundler/types.ts
@@ -31,5 +31,32 @@ export type BundlerConfig = {
    */
   dynamicPackages?: string[];
 
+  /**
+   * Packages to force-exclude from the generated `package.json` even if
+   * dependency analysis flagged them as in use.
+   *
+   * Useful when conditional dynamic imports (e.g. a dev-only
+   * `await import('@mastra/libsql')` gated by `process.env.NODE_ENV`) get
+   * picked up by static analysis but are tree-shaken out of the production
+   * bundle, polluting the output with packages your runtime never actually
+   * needs.
+   *
+   * Entries are matched by package name only — `'@mastra/libsql'` drops the
+   * package even when it was imported via a subpath, but `'@mastra/libsql/vector'`
+   * matches nothing and is silently ignored.
+   *
+   * This affects the generated `package.json` only; the package stays external
+   * to the bundle. Excluding something the runtime does reach therefore surfaces
+   * as a `MODULE_NOT_FOUND` at run time, not as a build error.
+   *
+   * @example
+   * ```typescript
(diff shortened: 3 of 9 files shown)


sgarfinkel:
Any chance we could get a review on this? @wardpeet

shiminshen:
Rebased onto current `main` and worked through the list.

**1 + 2.** Rebase done — `dependenciesToInstall` is `Map<string, ExternalDependencyInfo>` now, and the exclusion moved to a single sweep over the final map immediately before `writePackageJson`, so the transitive-workspace population can't re-add an excluded package. I dropped the in-loop `continue` instead of filtering at both insertion points: one place is harder to let drift on the next rebase.

**3.** Extended the coverage rather than narrowing the docs — filtering the final map picks up workspace-linked packages too. The JSDoc now states the three edges: package names only (a subpath entry like `@mastra/libsql/vector` matches nothing), `package.json` only so the package stays external to the bundle, and a mistaken exclusion surfacing as a runtime `MODULE_NOT_FOUND` rather than a build error.

One boundary I deliberately left alone: an excluded workspace package still appears under `resolutions`, and `packWorkspaceDependencies` still packs it. That felt like a different decision from "keep it out of `dependencies`" and I didn't want to widen the blast radius unasked — happy to make it consistent if you'd rather.

**4.** `Closes #16645` → `Relates to #16645`, so merging leaves the scanner half open.

**5.** `mastra-class.mdx` — added to both the bundler description and the `defaultValue` list, next to `dynamicPackages`.

**6.** Three tests in `bundler/index.test.ts` driving `_bundle` with `analyzeBundle` mocked and reading the generated `package.json`: excluded package absent, nothing dropped when the option is unset, and one aimed squarely at the re-add path from (2). I checked both assertions actually fail when the exclusion is removed, and that the third fails when the filter is moved back ahead of the workspace step — they're not green by accident.

Verification on the rebased head: `vitest run src/bundler src/build` → 458 passed, 1 failed. The failure is `src/build/fs-routing/fs-routing.test.ts`, which fails the same way on `cc85af24` without any of my changes. `tsc --noEmit` reports no errors in the files this touches (the `src/server/**` ones are unbuilt-workspace artifacts, same as you saw).

PR #16681, opened 2026-05-16
fix(deployer): support pnpm v11 in the deployer install path

### LINKED ISSUE #16613: [FEATURE] pnpm 11 support
### Problem Statement

I'd like to be able to use pnpm v11 with Mastra. `mastra build` fails on pnpm v11 with `ERR_PNPM_IGNORED_BUILDS` when the project depends on a package with postinstall scripts (e.g. `better-sqlite3`). This happens because the deployer writes incorrect pnpm metadata to the build output, then runs `pnpm install --ignore-workspace` in that directory, so the user's own build policy is dropped.

Related parts:
- `packages/deployer/src/services/deps.ts:148` runs `pnpm install --ignore-workspace` in that output dir, ignoring the user's configuration.
- `packages/deployer/src/bundler/index.ts:85-87` writes `pnpm: { neverBuiltDependencies: [] }` to `.mastra/output/package.json`, which is not respected in v11. This is also a security problem for v10, because `neverBuiltDependencies: []` bypasses the pnpm build allow/deny list. If a user originally disallowed the build script for a dependency for security reasons, then they would silently get hit by the problem since Mastra ignores this and allows all scripts.

It had security problems with v10, but worked without errors since it allowed all. In pnpm v10, `neverBuiltDependencies` and friends got merged into `allowBuilds`, so it doesn't work anymore.

### Proposed Solution

Replace `--ignore-workspace` + `pnpm: { neverBuiltDependencies: [] }` with a self-contained `pnpm-workspace.yaml` written into `.mastra/output/` before install:

- Even an empty `pnpm-workspace.yaml` file makes pnpm pick up that file while searching up the file tree. So, it will replace `--ignore-workspace`. `packages:` specified in the actual project will be ignored, just as the existing code intended.
- Carry over other relevant settings, e.g. `allowBuilds`(v11) and `onlyBuiltDependencies` / `neverBuiltDependencies` / `ignoredBuiltDependencies` (v10) into `pnpm-workspace.yaml`
  - A combination of `pnpm config list --json` (resolves `pnpm-workspace.yaml`, `.npmrc`, and global config) and the parent's `package.json > pnpm` field (which I believe `pnpm config` command does not take into account) should work to derive these
- Stop writing `pnpm: { neverBuiltDependencies: [] }` to the output `package.json` to avoid security problems on v10 and redundant configuration(which might decide to print warnings or errors later) on v11+

### Component

Deployment

### Alternatives Considered

- stay stuck on pnpm v10
- pass `pnpm_config_dangerously_allow_all_builds=true` env variable to make it work, same as Mastra's current security-problematic behavior with pnpm v10
- apply a local patch `@mastra/deployer` with `pnpm patch`

### Example Use Case

Self-explanatory

### Additional Context

_No response_

### Verification

shiminshen:
## Summary

Closes #16613.

`mastra build` broke on pnpm v11 with `ERR_PNPM_IGNORED_BUILDS` whenever the user's project depends on a package with a postinstall script (e.g. `better-sqlite3`). Two related issues, both flagged in the issue body:

1. **`pnpm: { neverBuiltDependencies: [] }` in the output `package.json`**
   - v10 silently treated this as "allow every postinstall script", which **overrode any allow/deny list the user had configured for security reasons** — a quiet security bypass.
   - v11 merged `neverBuiltDependencies` into `allowBuilds` in `pnpm-workspace.yaml`, so the key inside `package.json > pnpm` is **no longer honored**. The workaround stopped working but the security hole stayed in the source until someone hit a real package that needed scripts.

2. **`pnpm install --ignore-workspace`**
   - v11 doesn't isolate `.mastra/output` strongly enough on its own; the install still attaches to the user's parent workspace.

## What this PR does

- Drops the `pnpm: { neverBuiltDependencies: [] }` block from the output `package.json` in `packages/deployer/src/bundler/index.ts`.
- Writes a local `pnpm-workspace.yaml` (just `packages: []`) into the output directory before running `pnpm install`, in a new `Deps.ensurePnpmWorkspaceYaml` helper. pnpm walks up looking for a workspace manifest; the local file wins, and the output dir is treated as a self-contained install root.
- Removes `--ignore-workspace` from the pnpm install command. Replaced by the local workspace yaml.
- Adds explanatory comments at both edit sites so the historical bypass and its replacement are documented in code.

## Behavior change for users who relied on the bypass

Users who relied on the implicit "every build script runs" behavior to opt specific package scripts (e.g. `better-sqlite3`, native bindings) back in will now hit pnpm's default policy in the deployer output. They can:

- Configure `allowBuilds` (or `onlyBuiltDependencies` on v10) in their own `pnpm-workspace.yaml` and **re-run the build** — though right now the user's `allowBuilds` is NOT carried over into the deployer output; that's a deliberate follow-up.
- Run `pnpm config set allowBuilds <pkg>` inside `.mastra/output` if they need to opt scripts back in for a one-off install.

I'd consider the carry-over (read the user's `allowBuilds` / `onlyBuiltDependencies` and copy them into the output yaml) as the natural follow-up to this PR — flagged in the changeset note. Happy to do it in a follow-up once the shape of this PR is reviewed.

## Test plan

- [x] New unit tests in `packages/deployer/src/services/deps.test.ts`:
  - `does NOT pass --ignore-workspace to pnpm install (#16613)`
  - `still passes --loglevel=error for pnpm add`
  - `writes an empty packages list when no workspace yaml is present`
  - `does not overwrite an existing pnpm-workspace.yaml`
  - `is idempotent across repeated calls`
- [x] All `packages/deployer/src/{build,bundler,services}` tests pass (278 tests total).
- [x] `pnpm exec tsc --noEmit` reports only the pre-existing `@mastra/hono` / `@mastra/server` errors (unrelated to this change) and no new errors.

## Not in scope

- Migrating `supportedArchitectures` out of `package.json > pnpm` into `pnpm-workspace.yaml` (v11 still reads it from `package.json > pnpm` for the time being; we can move it when v11 explicitly deprecates that location).
- Carrying over `allowBuilds` / `onlyBuiltDependencies` from the user's workspace yaml. Worth doing as a follow-up; the user is best positioned to tell us whether they want their parent allow list to apply to the deployer output or not.

---
Written by an agent (Claude Code, claude-opus-4-7).

## ELI5

mastra was breaking when using pnpm v11 because pnpm changed how it finds workspace settings. This PR makes the build put a tiny workspace file in the output folder so pnpm treats the output as its own project and installs dependencies correctly.

---

### Summary
Fixes `mastra build` failures on pnpm v11 caused by incorrect pnpm metadata and the use of `--ignore-workspace`. Instead of writing a now-ignored/unwanted pnpm field into the generated package.json, the deployer now creates a local pnpm-workspace.yaml (packages: []) in the output directory so pnpm treats the output as an install root. The pnpm install command no longer passes `--ignore-workspace`. The implicit "allow every build script" bypass is removed; users must opt packages in via pnpm workspace settings if needed.

Key implementation notes verified in changed files:
- Removed writing `pnpm: { neverBuiltDependencies: [] }` from generated .mastra/output/package.json and replaced it with explanatory comments in packages/deployer/src/bundler/index.ts.
- Added Deps.ensurePnpmWorkspaceYaml(dir) which atomically writes pnpm-workspace.yaml with flag 'wx' (silently ignoring EEXIST/EISDIR) to avoid races and to preserve existing files.
- ensurePnpmWorkspaceYaml is invoked before running pnpm install; getPackageManagerCommand no longer includes `--ignore-workspace` for pnpm.
- Comments document why the historical bypass was removed and the rationale for the new approach.

### Files Changed
- .changeset/deployer-pnpm-11-support-16613.md — Changeset describing the fix and user guidance.
- packages/deployer/src/bundler/index.ts — Stop emitting pnpm neverBuiltDependencies workaround; add explanatory comments.
- packages/deployer/src/services/deps.ts — Add ensurePnpmWorkspaceYaml, writePnpmConfig adjustments, remove use of `--ignore-workspace` in pnpm command generation, and integrate workspace-yaml creation into install flow.
- packages/deployer/src/services/deps.test.ts — New Vitest tests covering:
  - pnpm install command does not include `--ignore-workspace` and pnpm add keeps --loglevel=error
  - ensurePnpmWorkspaceYaml writes `packages: []\n` when absent
  - existing pnpm-workspace.yaml is preserved
  - ensurePnpmWorkspaceYaml is idempotent (preserves mtime)

### User-Visible Behavior
- The previous implicit bypass that allowed all dependency build scripts is gone. Users who relied on that must explicitly opt packages in using pnpm workspace settings (e.g., allowBuilds / onlyBuiltDependencies) inside `.mastra/output` or configure them in a pnpm-workspace.yaml in their project and re-run builds.
- Carrying user allow-lists into the output is intentionally left for a follow-up.

### Testing
- New unit tests added and included in the deployer test suite.
- All deployer tests pass locally (278 tests); TypeScript checks show no new errors.

Closes `#16613`

[Review Change Stack](https://app.coderabbit.ai/change-stack/mastra-ai/mastra/pull/16681?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


gabrielanhaia:
Hey @wardpeet 
Any plan to get this one shipped soon?
Thanks a lot 🙏

wardpeet:
I can't seem to reproduce this on latest, these are my steps

1. install `pnpm@11.5.2` 
2. Run `pnpm dlx create-mastra@latest`
3. Run `pnpm build`
4. Run `pnpm start`
5. Run `pnpm dev`

no issues here, what am i missing?

wardpeet:
I believe we solved this already

PR #16682, opened 2026-05-16
feat(mastracode): /om toggle to skip attachments when observer is text-only

CalebBarnes:
Stacked on #16671 (now merged into main).

## Why

When the main model is multimodal but the observer model is text-only (e.g. DeepSeek via OpenRouter), forwarding image/file parts to the observer causes provider errors like `No endpoints found that support image input`. #16671 added a programmatic `observeAttachments` option on `ObservationConfig`; this PR adds the matching toggle to mastracode's `/om` settings overlay so operators can switch it on the fly.

## What

- New row in `/om` settings: **Observe attachments** (On / Off).
- Persists to global settings as `omObserveAttachments` (boolean | null).
- Seeded into harness initial state at startup so the configured value applies to new threads immediately.
- Threaded through `getDynamicMemory` to the underlying `observation.observeAttachments` config, and included in the memory cacheKey so toggling rebuilds the Memory instance.
- Defaults to **On** for backwards compatibility.
- Generalised the internal `BooleanSubmenu` to accept per-row label/description overrides so the existing caveman row keeps its descriptions.

image

## Test plan

- `pnpm exec vitest run src/__tests__/index.test.ts` ✓ (incl. new startup-seed tests)
- `pnpm exec vitest run src/tui/commands/__tests__/om.test.ts src/onboarding/__tests__/settings.test.ts` ✓
- `pnpm exec tsc --noEmit` ✓
- `pnpm build:lib` ✓
- `pnpm lint` ✓
- Manually: opened `/om` overlay, toggled the new row on/off, confirmed value persisted across restart.

## ELI5
This adds a simple On/Off switch in the Observational Memory (/om) settings that decides whether images and files are sent to the Observer AI. Turn it off if your observer only handles text; the choice is saved and used for new conversations right away.

## Overview
Adds an "Observe attachments" toggle to the /om TUI that controls forwarding of image/file attachment parts to the Observer LLM to avoid provider errors with text-only observers. The preference is persisted to global settings, seeded into the harness initial state at startup, threaded into memory construction, and included in the memory cache key so changes rebuild cached Memory instances.

### Configuration & Persistence
- New global setting: models.omObserveAttachments: boolean | null (DEFAULTS: null to inherit built-in default).
- Harness state schema: state.observeAttachments: boolean (default true for backwards compatibility).
- createMastraCode seeds initialState.observeAttachments from globalSettings.models.omObserveAttachments when defined.
- Added exported helper persistOmObserveAttachments(enabled: boolean) to save the global setting.

### Memory & Caching
- getDynamicMemory reads state.observeAttachments (default true), sets observeAttachments on the constructed Observational Memory instance, and includes observeAttachments in the Memory cacheKey so toggling segregates/rebuilds cached Memory.

### UI & User Interaction
- OMSettingsComponent adds an "Observe attachments" SettingItem (On/Off) wired to callbacks.onObserveAttachmentsChange.
- BooleanSubmenu generalized to accept per-row labels/descriptions so existing rows keep their original descriptions.
- OM command handlers update harness state, persist settings via persistOmObserveAttachments, and show an info toast on change.
- Settings menu layout height increased to fit the new entry.

### Backwards Compatibility
- Default behavior remains On to avoid regressions.
- Placeholder transcript behavior for filtered attachments is preserved.

## Tests
- Vitest tests added to verify createMastraCode seeds harness initialState.observeAttachments from persisted global setting (boolean vs null).
- Test fixtures updated across TUI/onboarding tests to include models.omObserveAttachments (initialized to null).
- Manual verification: toggling the /om row persists across restart.
- CI checks reported passing: vitest unit tests, TUI/onboarding tests, tsc --noEmit, pnpm build:lib, pnpm lint.

## Review note / Follow-up requested
Reviewer requested adding a focused, non-blocking test for the command-level interaction path that isn't covered by startup-seeding tests: invoke the OM UI callback onObserveAttachmentsChange(false) and assert that (1) harness.setState({ observeAttachments: false }) is called and (2) settings.models.omObserveAttachments is persisted as false. This would catch regressions in the command callback wiring (mastracode/src/tui/commands/om.ts).

[Review Change Stack](https://app.coderabbit.ai/change-stack/mastra-ai/mastra/pull/16682?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


TylerBarnes:
From Mastra Code: Non-blocking test coverage suggestion: the PR covers startup seeding of `omObserveAttachments` from global settings into harness initial state, but I don’t see a focused test for the `/om` interaction path itself.

The path I’d like covered is:

```ts
onObserveAttachmentsChange(false)
  -> harness.setState({ observeAttachments: false })
  -> persist settings.models.omObserveAttachments = false
```

Right now, if the callback wiring regressed in `mastracode/src/tui/commands/om.ts`, the existing startup tests would still pass because they only verify that an already-persisted value is seeded correctly on boot.

A small command-level test that opens `/om` with a mocked `OMSettingsComponent` callback, invokes `onObserveAttachmentsChange(false)`, and asserts both the harness state update and persisted setting would close that gap.

TylerBarnes (approved):

PR #16691, opened 2026-05-17
fix: inherit env for MastraCode commands

TylerBarnes:
This restores MastraCode command execution so subprocesses inherit the user's environment again, while keeping traces from storing env values or obvious secrets.

Before, only allowlisted env vars reached commands, so exported values needed by tests or scripts were missing:

```ts
const env: NodeJS.ProcessEnv = {};
for (const key of SANDBOX_ENV_ALLOWLIST) {
  env[key] = process.env[key];
}
```

Now the sandbox receives the full parent env, and workspace tracing redacts env-shaped fields plus secret-looking keys before spans are recorded:

```ts
const env = {
  ...process.env,
  CI: 'true',
  NONINTERACTIVE: '1',
};
```

I added coverage for both sides: MastraCode subprocess env inheritance and core workspace trace redaction.

Verified with:

```bash
pnpm i
pnpm run prettier:changed
pnpm run lint
pnpm --dir mastracode exec vitest run src/agents/__tests__/workspace-env.test.ts
pnpm --dir packages/core exec vitest run src/workspace/tools/__tests__/tracing.test.ts
```

## ELI5

This PR lets sandboxed code run commands with all the environment variables from your computer (like API keys and config settings), instead of just a few pre-approved ones. However, to keep things secure, it hides all those environment variables from being recorded in logs and traces—so sensitive information doesn't accidentally get stored or leaked.

### MastraCode Sandbox Environment Inheritance

The `buildSandboxEnv()` function in `workspace.ts` was refactored to inherit the full parent environment instead of using an allowlist. Previously, only whitelisted keys (defined in `SANDBOX_ENV_ALLOWLIST`) were copied to the sandbox. Now, the implementation spreads all of `process.env` into the sandbox environment and explicitly sets non-interactive/CI flags (`FORCE_COLOR`, `CLICOLOR_FORCE`, `TERM`, `CI`, `NONINTERACTIVE`, `DEBIAN_FRONTEND`).

### Workspace Trace Data Sanitization

Added sanitization utilities in `packages/core/src/workspace/tools/tracing.ts` that detect and redact:
- Environment variable fields (both keys and values)
- Common credential field names (password, secret, token, key, authorization, etc.)

The `startWorkspaceSpan` function now sanitizes both input and output data before recording them in traces, ensuring sensitive information isn't stored.

### Testing

Added comprehensive test coverage:
- **`workspace-env.test.ts`**: Verifies that sandboxed commands inherit parent environment variables (tests that a custom env var set in the parent is accessible within the sandbox)
- **`tracing.test.ts`**: Verifies that `startWorkspaceSpan` properly redacts environment variables and credential-like fields from trace input and output

### Packaging

Updated the changeset (`swift-shells-trace.md`) to mark both `mastracode` and `@mastra/core` as patch releases, documenting the behavior change and security improvement.

[Review Change Stack](https://app.coderabbit.ai/change-stack/mastra-ai/mastra/pull/16691?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/.changeset/swift-shells-trace.md
+++ b/.changeset/swift-shells-trace.md
@@ -1,0 +1,6 @@
+---
+'mastracode': patch
+'@mastra/core': patch
+---
+
+Restore MastraCode local command execution to inherit parent environment variables while redacting env-shaped and secret-looking workspace trace data.
--- a/mastracode/src/agents/__tests__/workspace-env.test.ts
+++ b/mastracode/src/agents/__tests__/workspace-env.test.ts
@@ -1,0 +1,52 @@
+import * as fs from 'node:fs/promises';
+import * as os from 'node:os';
+import * as path from 'node:path';
+import { afterEach, describe, expect, it, vi } from 'vitest';
+
+vi.mock('../onboarding/settings.js', () => ({
+  loadSettings: () => ({}),
+}));
+vi.mock('../../onboarding/settings.js', () => ({
+  loadSettings: () => ({}),
+}));
+
+const originalEnv = { ...process.env };
+
+function createRequestContext(projectPath: string) {
+  return {
+    get: (key: string) =>
+      key === 'harness'
+        ? {
+            modeId: 'build',
+            getState: () => ({
+              projectPath,
+              sandboxAllowedPaths: [],
+            }),
+          }
+        : undefined,
+  };
+}
+
+afterEach(() => {
+  process.env = { ...originalEnv };
+  vi.resetModules();
+});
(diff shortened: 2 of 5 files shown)


intojhanurag (approved):

intojhanurag:
Hey @TylerBarnes , You don't think , first CodeRabbit suggestion make sense for a bit ? 
btw not a big risk , it looks good

wardpeet (approved):

PR #16692, opened 2026-05-17
fix: preserve streamed messages during stale initial sync

### LINKED ISSUE #16650: [BUG] First assistant response disappears in Mastra Studio after creating a new chat thread after using tools
### Describe the bug

In Mastra Studio, the first assistant response in a newly created chat thread appears briefly and then disappears from the UI.

If I ask the agent to repeat the answer, the repeated response remains visible, and the rest of the conversation behaves normally.

### Additional context

- I have observed this across multiple agents every time when use tools.
- The issue seems to affect only the first assistant response in a newly created thread.
- Later messages in the same thread remain visible.
- I start from a new chat route such as `/agents/<agent-id>/chat/new`.

I suspect this may be related to the transition from the temporary "new chat" state to the persisted thread view, but I have not verified the root cause.

### Steps to reproduce

1. Open Mastra Studio.
2. Open an agent chat in a new thread.
3. Send the first message command to use some tools.
4. Wait for the assistant to produce its first response.
5. Observe that the first response appears briefly and then disappears.
6. Send another message, for example: "Please repeat your previous answer."
7. Observe that the repeated response remains visible and subsequent messages in the same thread behave normally.

https://github.com/turboMe/mastra-agentic-agents-blue-green-autohealing-system

### Expected behavior

The first assistant response should remain visible in the chat history after it is generated.

### Actual behavior

The first assistant response flashes on screen and then disappears almost immediately.

This happens nearly every time when starting a new chat thread. Later responses in the same thread remain visible.

### Environment Information

```block
System:
    CPU: (16) x64 AMD Ryzen 7 9700X 8-Core Processor
    Memory: 44.09 GB / 60.44 GB
    Container: Yes
    Shell: 5.2.21 - /bin/bash

  Browsers:
    Chrome: 148.0.7778.96
    Firefox: 150.0
    Firefox Developer Edition: 150.0
  npmGlobalPackages:
    @anthropic-ai/claude-code: 2.1.5
    agent-browser: 0.16.3
    corepack: 0.33.0
    firebase-tools: 14.20.0
    npm: 10.8.2

The project uses `.nvmrc` plus a wrapper script to run Mastra commands under Node v22.20.0.
```

### Verification

nacim-coder:
## Summary
- preserve locally streamed chat messages when a same-thread `initialMessages` refresh is stale
- keep the existing reset behavior when switching threads
- add regression coverage for stale refreshes, caught-up refreshes, and thread changes

## Why
When Mastra Studio creates a new chat thread, the route moves from `/chat/new` to the persisted thread id while the first assistant response can still be streaming or not yet fully reflected in memory. A same-thread `initialMessages` refresh with fewer messages can replace the local `useChat` state and make the first assistant response disappear from the UI.

This keeps the local streamed messages if the incoming same-thread initial sync is shorter than the current local state, while still accepting fetched messages once they catch up and still resetting normally across thread changes.

Fixes #16650.

## Tests
- `corepack pnpm --filter ./client-sdks/react test src/agent/initial-messages-sync.test.ts`
- `git diff --check HEAD~1..HEAD`

## ELI5 (Explain Like I'm 5)

This PR fixes a bug where AI chat responses would flash on screen then disappear in Mastra Studio. The problem happened because the chat was being refreshed while the AI was still sending its response. The fix is simple: when the chat refreshes, it checks whether the new messages are more complete than what's already showing; if not, it keeps the current messages instead of replacing them with incomplete data.

### New module: `resolveInitialMessagesSync`

Added a new utility function in `client-sdks/react/src/agent/initial-messages-sync.ts` that intelligently decides which message list to use:
- If the thread hasn't changed and the current local messages are more complete than the incoming messages (e.g., while streaming), it preserves the local messages
- If the incoming messages have caught up or the thread has changed, it uses the incoming messages
- This prevents loss of streamed messages during stale initial-sync refreshes while maintaining normal reset behavior on thread switches

### Updated `useChat` hook

Modified `client-sdks/react/src/agent/hooks.ts` to:
- Track the current `threadId` to detect when the user switches to a different thread
- Use `resolveInitialMessagesSync` to conditionally update messages based on whether messages are stale or the thread has changed
- Update `_currentRunId` based on the resolved message list

### Test coverage

Added comprehensive test suite in `client-sdks/react/src/agent/initial-messages-sync.test.ts` covering:
- Preserving messages when same-thread refresh is stale (fewer messages than current state)
- Accepting messages when incoming data has caught up
- Normal reset behavior when switching threads

### Documentation

Added Changeset entry for `@mastra/react` (patch release) documenting the fix.

## Related Issue

Fixes `#16650`: First assistant response disappearing in newly created chat threads after using tools.

[Review Change Stack](https://app.coderabbit.ai/change-stack/mastra-ai/mastra/pull/16692?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/.changeset/preserve-streamed-messages.md
+++ b/.changeset/preserve-streamed-messages.md
@@ -1,0 +1,5 @@
+---
+@mastra/react: patch
+---
+
+Preserve locally streamed chat messages when a same-thread initial messages refresh is stale.
--- a/client-sdks/react/src/agent/hooks.ts
+++ b/client-sdks/react/src/agent/hooks.ts
@@ -8,6 +8,7 @@ import type { ChunkType, NetworkChunkType } from '@mastra/core/stream';
 import { useCallback, useEffect, useRef, useState } from 'react';
 import type { MastraUIMessage } from '../lib/ai-sdk';
 import { extractRunIdFromMessages } from './extractRunIdFromMessages';
+import { resolveInitialMessagesSync } from './initial-messages-sync';
 import { convertSignalDataToBase64String } from './signal-data';
 import type { ModelSettings } from './types';
 import { finishStreamingAssistantMessage, toUIMessage } from '@/lib/ai-sdk';
@@ -109,11 +110,19 @@ export const useChat = ({
   const baseClient = useMastraClient();
   const [isRunning, setIsRunning] = useState(false);
 
+  const _lastInitialMessagesThreadId = useRef(threadId);
+
   useEffect(() => {
     const formattedMessages = resolveInitialMessages(initialMessages || []);
-    setMessages(formattedMessages);
-    _currentRunId.current = extractRunIdFromMessages(formattedMessages);
-  }, [initialMessages]);
+    const threadChanged = _lastInitialMessagesThreadId.current !== threadId;
+    _lastInitialMessagesThreadId.current = threadId;
+
+    setMessages(currentMessages => {
+      const nextMessages = resolveInitialMessagesSync({ currentMessages, formattedMessages, threadChanged });
+      _currentRunId.current = extractRunIdFromMessages(nextMessages);
+      return nextMessages;
+    });
+  }, [initialMessages, threadId]);
 
   useEffect(() => {
     _requestContext.current = propsRequestContext;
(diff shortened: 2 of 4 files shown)


mikhael28:
This PR was originally created to address https://github.com/mastra-ai/mastra/issues/16650 but as we have not been able to recreate the issue that the OP posted, I don't have confidence that this PR will address the issue. Going to close - thank you for your contribution @nacim-coder thanks for understanding.

PR #16721, opened 2026-05-18
fix(core): include MessageList inputStep mutations in processor trace output

### LINKED ISSUE #16701: [BUG] processInputStep mutations via MessageList are applied but not reflected in trace output
This issue was created from Discord post 1505779508169539654:

[Open in Discord](https://discord.com/channels/1309558646228779139/1505779508169539654)

## Summary

When an input step processor mutates the provided `messageList` by calling `messageList.addSystem(...)` and returns that same `MessageList`, the mutation is applied and affects the next LLM inference, but the Mastra trace span for the processor shows an empty output `{}` instead of the updated `systemMessages`.

## Version

- @mastra/core: 1.35.0

## What happens

Inside `processInputStep`:

```ts
async processInputStep({ messageList }) {
  messageList.addSystem({
    role: 'system',
    content: '# Project Policy ...',
  })

  return messageList
}
The added system message is actually available to the next inference, so the processor behavior works.

However, in Mastra Studio tracing:

the input step processor: ... span appears
the processor executes successfully
the span OUTPUT is {}
This makes it look as if the processor produced no effect, even though it mutated the prompt sent to the model.

Why this seems inconsistent
For processInput, when a processor returns a MessageList, the runner expands it into:

messages: result.get.all.db(),
systemMessages: result.getAllSystemMessages()
But in the inputStep workflow runner path, a returned MessageList is normalized to:

{ messageList: result }
and the trace output builder only surfaces payload.messages / payload.systemMessages, so the mutation is not visible in the span output.

Expected behavior
If processInputStep mutates messageList, the trace span should expose the resulting systemMessages diff, just like the non-step input processor path does.

Actual behavior
The next inference receives the injected system message, but the trace span output remains {}.

Reproduction
Register an input processor with processInputStep
Inside it, call messageList.addSystem(...)
Return the same messageList
Run an agent and inspect the processor span in Mastra Studio traces
Observe that the next inference uses the injected system prompt, but the processor span output is still {}

---

![Screenshot 1](https://cdn.discordapp.com/attachments/1505779508169539654/1505779585244201031/o1f72ypB6z0A0v4txoTw31JAAAAAElFTkSuQmCC.png?ex=6a0bde05&is=6a0a8c85&hm=285b039a420507e146eb0a7f9bbec248587ab7aac7aede4bb34e9af3aed4c147&)

### PR DESCRIPTION
**Pull request by nacim-coder**
Fixes #16701

## Summary
- Preserve realized `MessageList` messages and system messages after workflow `processInputStep` processors run.
- Ensures mutations like `messageList.addSystem(...)` are reflected in processor span output when the same `MessageList` is returned.
- Adds regression coverage for both workflow processor and legacy runner paths.

## Tests
- `pnpm exec eslint src/workflows/workflow.ts src/workflows/processor-step.test.ts src/processors/process-input-step.test.ts`
- `pnpm exec vitest run src/workflows/processor-step.test.ts src/processors/process-input-step.test.ts --typecheck.enabled=false --fileParallelism=false`
- `pnpm build:core`

## ELI5

If a processor edits a shared list of messages (for example, adding system instructions) and returns that same list, those edits were used but not shown in the debug trace — this PR makes the debug trace show the actual message and system-message changes so traces match runtime behavior.

### Core Changes
packages/core/src/workflows/workflow.ts
- In the `processor` step execution path for the `inputStep` phase, the returned pass-through state now derives `messages` and `systemMessages` from the (potentially mutated) `checkedMessageList` via `checkedMessageList.get.all.db()` and `checkedMessageList.getAllSystemMessages()`.
- Ensures returned `messages` and `systemMessages` reflect the actual post-mutation state while preserving other validated fields and messageId fallback logic.

### Test Coverage
packages/core/src/processors/process-input-step.test.ts
- Added a Vitest regression that verifies system-message mutations performed on `messageList` inside a `processInputStep` processor are reflected in the tracing "processor span" output.
- The test wires a fake `currentSpan`/`processorSpan` and asserts `processorSpan.end` is called with `output.systemMessages` containing the injected system message.

packages/core/src/workflows/processor-step.test.ts
- Updated to import `MessageList` at runtime and `MastraDBMessage` as a type.
- `createMockMessageList` now tracks system-message mutations and implements `addSystem`, `replaceAllSystemMessages`, and `getAllSystemMessages`.
- Added a test verifying that when a processor’s `processInputStep` mutates the provided `MessageList` (e.g., `addSystem`), the step `execute` result includes those injected `systemMessages`.

### Changeset
.changeset/input-step-message-list-traces.md
- Adds a patch changeset documenting that input step processor traces now include system message mutations made via `MessageList`.

## Tests / Validation
- ESLint: pnpm exec eslint src/workflows/workflow.ts src/workflows/processor-step.test.ts src/processors/process-input-step.test.ts
- Unit tests: pnpm exec vitest run src/workflows/processor-step.test.ts src/processors/process-input-step.test.ts --typecheck.enabled=false --fileParallelism=false
- Build: pnpm build:core

[Review Change Stack](https://app.coderabbit.ai/change-stack/mastra-ai/mastra/pull/16721?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### CODE REVIEW

**Comment by ErKeLost** [comment]
Hello, when can this pull request (PR) be merged? I am currently dynamically inserting in the step processor. Will it not be included in the system prompt? I am currently implementing it through another API. I would like to know if this setup can work properly, because I can see an output in the system prompt on the Mastra platform

```ts
tsmessageList.addSystem(...)
return {
  systemMessages: messageList.getAllSystemMessages(),
}
```

nacim-coder:
This PR is waiting for maintainer review/approval, so I can’t give a merge timeline.

The patch should cover the case where `processInputStep` mutates and returns the same `MessageList`, including system message changes. Those mutations should now be reflected in the processor trace output.

The current red checks look unrelated to the changed core files. Final merge timing is up to the Mastra maintainers.

packages/core/src/workflows/workflow.ts
```diff
@@ -1195,12 +1195,18 @@ function createStepFromProcessor<TProcessorId extends string>(
...
+              // Preserve the realized MessageList state in return so span output and subsequent steps
+              // can observe mutations made via messageList even when the processor returns the same MessageList.
+              const { messages: _messages, systemMessages: _systemMessages, ...validatedRest } = validatedResult;
+
               return {
                 ...passThrough,
-                messages,
-                ...validatedResult,
+                messages: updatedMessages,
+                systemMessages: updatedSystemMessages,
```
wardpeet, line 1219: Should we pass in the messageList as well?
```suggestion
                systemMessages: updatedSystemMessages,
                messageList: checkedMessageList,
```

TylerBarnes: seems like a good idea

TylerBarnes: ah wait it's already in ...passthrough

ErKeLost:
@nacim-coder   hi

I want to know if what I add through the step processor is actually visible in the system prompt, even if it's not visible 

in the platform trace. I need to ensure that this will not affect the normal operation of my business. If it is indeed 

added to the system prompt, then whether I view the trace or not doesn't matter

nacim-coder:
@ErkeLost hi,

Yes, my understanding is that if your processInputStep adds system messages and returns the updated systemMessages, they should be part of the model input.

This PR is mainly fixing the mismatch where MessageList mutations were applied but not shown correctly in the trace. So it should make the trace match the actual runtime behavior, not change the normal behavior of your flow.

ErKeLost:
@nacim-coder  thanks

intojhanurag:
Hey @nacim-coder , I think it actually resolved by #16950

PR #16742, opened 2026-05-18
fix(studio): prevent assistant message remount storm during streaming

### LINKED ISSUE #15275: [BUG] "Tried to unmount a fiber that is already unmounted" in Mastra Studio when running long operations
### Describe the Bug

Once I requested supervisor agent for long task, I jump into error from title. When i refresh site and try to open this chat - the same. Happens when running studio with NODE_ENV=production and without it. Occurs also when Mastra is deployed in debian based container

### Steps To Reproduce

1. Create supervisor agent with some subagent
2. Add huge maxSteps (like 100)
3. Run some job which delegates some work to subagent
4. Wait until error is thrown

https://cannot-share.com

### Expected Behavior

No matter how long operations are, UI should be able to render them

### Environment Information

```block
Node runtime: Bun 1.3
System: Windows 11
Browser: Brave

Package.json:
"dependencies": {
    "@ai-sdk/anthropic": "^3.0.69",
    "@ai-sdk/openai-compatible": "^2.0.41",
    "@mastra/core": "^1.24.1",
    "@mastra/libsql": "^1.8.0",
    "@mastra/loggers": "^1.1.1",
    "@mastra/mcp": "^1.4.2",
    "@mastra/memory": "^1.15.0",
    "@mastra/pg": "^1.9.0",
    "@mastra/qdrant": "^1.0.2",
    "@types/pg": "^8.20.0",
    "pg": "^8.20.0",
    "playwright": "^1.56.0",
    "sharp": "^0.34.4",
    "zod": "^4.3.6"
  },
  "devDependencies": {
    "@playwright/mcp": "^0.0.42",
    "@types/node": "^24.5.2",
    "mastra": "^1.5.0",
    "typescript": "^6.0.2"
  }
```

### Verification

itskrishna21:
## Summary

Fixes #15275 — "Tried to unmount a fiber that is already unmounted" crash in Studio when running long agent operations.

## Root Cause

`WrappedAssistantMessage` was defined **inside** the `Thread` component body:

```tsx
// Before — new function reference on every render
const WrappedAssistantMessage = (props: MessagePrimitive.Root.Props) => {
  return <AssistantMessage {...props} hasModelList={hasModelList} />;
};
```

React identifies component types by reference identity. A new function on every render means React sees a **different component type** on each re-render and **unmounts + remounts the entire assistant message subtree** instead of updating it in place.

Under high-frequency streaming (e.g. `maxSteps=100`), `Thread` re-renders hundreds of times, causing a remount storm that surfaces as the internal React invariant failure: _"Tried to unmount a fiber that is already unmounted"_. The error also persists after a page refresh because the same rendering path is hit when replaying historical messages.

## Fix

Wrap the component in `useMemo` with `[hasModelList]` as the dependency, so its identity is stable across re-renders and only changes when the prop actually changes (which is rare):

```tsx
// After — stable identity across re-renders
const WrappedAssistantMessage = useMemo(
  () =>
    function AssistantMessageWrapper(props: MessagePrimitive.Root.Props) {
      return <AssistantMessage {...props} hasModelList={hasModelList} />;
    },
  [hasModelList],
);
```

## Verification

- This is a violation of [React's rule against defining components inside other components](https://react.dev/learn/your-first-component#nesting-and-organizing-components), documented to cause exactly this class of reconciliation failure
- `pnpm --filter ./packages/playground lint` passes with 0 errors
- The fix is a pure refactor — identical render output, no behavior change

## ELI5

This PR fixes a bug in Mastra Studio where streaming long agent conversations would crash. The problem was that React kept destroying and rebuilding the message display component unnecessarily. The fix tells React "keep using the same message display" instead of rebuilding it every time, which stops the crashes.

### Changed Files

1. **packages/playground/src/lib/ai-ui/thread.tsx** (+8/-4)
   - Added `useMemo` to the React imports
   - Converted `WrappedAssistantMessage` from an inline function definition to a memoized component
   - Set `hasModelList` as the dependency, ensuring the wrapper reference only updates when this value changes
   - This prevents React from treating the assistant message component as a new component on each re-render

2. **.changeset/fix-studio-fiber-unmount-crash.md** (+5/-0)
   - Added a Changesets release note documenting the fix as a patch for `mastra`

## Root Cause

The `WrappedAssistantMessage` component was defined inside the `Thread` component body, causing React to create a new function reference on every render. React would interpret this as a completely different component structure and trigger unnecessary unmount/remount cycles. During high-frequency streaming operations (especially with large `maxSteps` values), this cascade of mount/unmount events would crash the UI with the React invariant error "Tried to unmount a fiber that is already unmounted."

## Solution

By wrapping the component definition in `useMemo` with `hasModelList` as the dependency, the component reference now remains stable across re-renders. React will only remount when `hasModelList` actually changes, preventing the spurious mount/unmount cycles that were causing the crash.

[Review Change Stack](https://app.coderabbit.ai/change-stack/mastra-ai/mastra/pull/16742?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/.changeset/fix-studio-fiber-unmount-crash.md
+++ b/.changeset/fix-studio-fiber-unmount-crash.md
@@ -1,0 +1,5 @@
+---
+"mastra": patch
+---
+
+Fixed a React component identity bug in Studio that caused "Tried to unmount a fiber that is already unmounted" crashes during long-running agent operations. The assistant message component was being redefined on every render, forcing React to remount the entire message subtree instead of updating it. This was especially visible with agents using high `maxSteps` values.
--- a/packages/playground/src/lib/ai-ui/thread.tsx
+++ b/packages/playground/src/lib/ai-ui/thread.tsx
@@ -2,7 +2,7 @@ import type { MessagePrimitive } from '@assistant-ui/react';
 import { ComposerPrimitive, ThreadPrimitive, useComposer, useComposerRuntime } from '@assistant-ui/react';
 import { Avatar, Button, ButtonsGroup, cn, useAutoscroll } from '@mastra/playground-ui';
 import { ArrowUp, EyeIcon, Mic, PlusIcon } from 'lucide-react';
-import { useEffect, useRef, useState } from 'react';
+import { useEffect, useMemo, useRef, useState } from 'react';
 import { AttachFileDialog } from './attachments/attach-file-dialog';
 import { ComposerAttachments } from './attachments/attachment';
 import { BracketOverlay } from './components/bracket-overlay';
@@ -40,9 +40,13 @@ export const Thread = ({ agentName, agentId, threadId, hasMemory, hasModelList, 
   // 3. NOT currently viewing browser in sidebar
   const showThumbnailInChat = hasSession && (viewMode === 'collapsed' || viewMode === 'expanded') && !isInSidebar;
 
-  const WrappedAssistantMessage = (props: MessagePrimitive.Root.Props) => {
-    return <AssistantMessage {...props} hasModelList={hasModelList} />;
-  };
+  const WrappedAssistantMessage = useMemo(
+    () =>
+      function AssistantMessageWrapper(props: MessagePrimitive.Root.Props) {
+        return <AssistantMessage {...props} hasModelList={hasModelList} />;
+      },
+    [hasModelList],
+  );
 
   return (
     <ThreadWrapper>


mfrachet:
Hey there, thanks for you help 👋🏻 

How is this fixing the mentionned issue? Do you have a repo on which I could test the fix?

PR #16743, opened 2026-05-18
fix(observability): pause exporter uploads after auth failures

noahcrowley:
## Description

Adds an auth-failure cooldown for MastraPlatformExporter and CloudExporter so invalid credentials do not keep repeatedly uploading to the observability API.

Auth failures now stop retries immediately, drop events during the cooldown window, and probe again with exponential backoff plus jitter. Cooldown warnings and dropped-event counts are logged through the active injected exporter logger.

## Type of change

- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [ ] Test update

## Checklist

## ELI5

When an exporter tries to upload data but gets rejected because of bad credentials, it now takes a break instead of immediately trying again. It waits longer each time it fails, and during that wait time it stops sending data. Once enough time has passed, it tries again. This prevents wasting effort sending requests that will always fail.

## Overview

This PR adds an authentication-failure cooldown mechanism to the observability exporters (`MastraPlatformExporter` and `CloudExporter`) to gracefully handle repeated authorization failures. When an exporter encounters a 401 or 403 response, it enters a cooldown period during which:

- New events are dropped instead of buffered
- No upload attempts are made
- After the cooldown expires, probing resumes with exponential backoff and symmetric jitter
- A warning log includes the failure status, retry count, and cooldown duration

## Technical Changes

**New Module:** `auth-failure-cooldown.ts`
- `AuthFailureError`: Custom error class for authentication failures with HTTP status
- `isAuthFailureError`: Type guard to identify authentication failures
- `fetchWithAuthFailureHandling`: Wrapper around `fetchWithRetry` that stops retrying immediately on 401/403 responses and throws `AuthFailureError`
- `AuthFailureCooldown`: Class that tracks consecutive auth failures, manages cooldown timestamps, and counts dropped events. Provides methods to:
  - Drop events during cooldown with `dropEventIfCoolingDown()` and `dropEventsIfCoolingDown(count)`
  - Record new failures with exponential backoff (capped) plus jitter via `recordFailure()`
  - Reset counters and return dropped event count via `reset()`

**CloudExporter Updates:**
- Integrates `AuthFailureCooldown` instance
- Drops events early in signal handlers when actively cooling down
- Uses `fetchWithAuthFailureHandling` instead of `fetchWithRetry` for batch uploads
- Detects auth failures in `flushSignalBatch()` and records them with dropped event counts
- On successful flush, resets cooldown and logs enriched metadata including `droppedEventsDuringAuthCooldown`

**MastraPlatformExporter Updates:**
- Identical auth-cooldown integration pattern as CloudExporter
- Drops incoming events when cooling down
- Switches to `fetchWithAuthFailureHandling` for batch uploads
- Records auth failures with status, failed signals, and dropped batch size

**Testing:**
- Added `mockAuthFailure` helper in both test files to configure authentication failure scenarios
- Extended existing tests to assert `shouldRetryResponse` function presence in `fetchWithRetry` calls
- New auth-cooldown test blocks verify:
  - Events are dropped during in-flight cooldown, then resume after expiration
  - Retry count and cooldown state reset properly
  - Dropped event accounting works correctly
  - Events buffered during auth failure are dropped appropriately

**Documentation:**
- Added Changesets entry documenting the pause in observability uploads when invalid credentials are detected

[Review Change Stack](https://app.coderabbit.ai/change-stack/mastra-ai/mastra/pull/16743?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/.changeset/thirty-bags-begin.md
+++ b/.changeset/thirty-bags-begin.md
@@ -2,4 +2,4 @@
 '@mastra/observability': patch
 ---
 
-Improved MastraPlatformExporter and CloudExporter handling of invalid credentials to avoid repeated unauthorized upload attempts.
+Paused observability uploads after invalid credentials so exporters stop repeatedly sending unauthorized requests.
--- a/.changeset/thirty-bags-begin.md
+++ b/.changeset/thirty-bags-begin.md
@@ -1,0 +1,5 @@
+---
+'@mastra/observability': patch
+---
+
+Improved MastraPlatformExporter and CloudExporter handling of invalid credentials to avoid repeated unauthorized upload attempts.
--- a/observability/mastra/src/exporters/auth-failure-cooldown.ts
+++ b/observability/mastra/src/exporters/auth-failure-cooldown.ts
@@ -1,0 +1,116 @@
+import type { IMastraLogger } from '@mastra/core/logger';
+import { fetchWithRetry } from '@mastra/core/utils';
+
+const AUTH_FAILURE_STATUSES = new Set([401, 403]);
+const AUTH_COOLDOWN_BASE_MS = 60_000;
+// Target cap before jitter; positive jitter can exceed this to avoid synchronized retry probes.
+const AUTH_COOLDOWN_MAX_MS = 15 * 60_000;
+const AUTH_COOLDOWN_JITTER_RATIO = 0.1;
+
+export class AuthFailureError extends Error {
+  readonly status: number;
+
+  constructor(status: number, cause?: unknown) {
+    super(`Request failed with authentication status: ${status}`, { cause });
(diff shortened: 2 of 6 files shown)


epinzur (approved):
looks great.

There is also this new concept of emitting separate "droppedEvent" messages in observability itself... that dev's can connect a callback to.  Maybe we should also route the dropped events back through this? Especially for cases when the user's logs are also getting dropped because of Auth errors.

See:
- https://github.com/mastra-ai/mastra/blob/main/observability/mastra/src/exporters/default.ts#L246
- https://github.com/mastra-ai/mastra/pull/16111

PR #16760, opened 2026-05-19
draft: Add v-next Postgres observability storage domain

epinzur:
## Description

Adds a new `ObservabilityStoragePostgresVNext` domain for low-volume production observability workloads on Postgres. This is an insert-only, partitioned implementation that mirrors the ClickHouse v-next layout while adapting to Postgres semantics.

### Key Features

- **Insert-only model** with retry idempotency via `ON CONFLICT DO NOTHING`
- **Per-signal partitioned tables** (or TimescaleDB hypertables when available)
- **Root-span projection** via partial indexes instead of a separate table
- **Discovery caching** with stale-while-revalidate semantics in a dedicated table
- **Full observability surface**: spans, metrics, logs, scores, feedback with list, aggregate, breakdown, time series, and percentile operations
- **Delta-polling support** for efficient incremental reads via cursor-based pagination
- **Flexible partitioning modes**: auto-detection of TimescaleDB, pg_partman, or native Postgres partitioning

### Implementation Details

- `ddl.ts`: Raw DDL for all signal tables and discovery cache
- `index.ts`: Main domain class implementing `ObservabilityStorage` interface
- `traces.ts`: Trace/branch list reads with root-span filtering
- `tracing.ts`: Span write operations
- `metrics.ts`, `feedback.ts`, `scores.ts`, `logs.ts`: Signal-specific operations
- `discovery.ts`: Entity/environment/tag/metric discovery with caching
- `olap.ts`: Shared OLAP helpers (bucketing, aggregation, percentiles)
- `helpers.ts`: Record conversion and normalization utilities
- `filters.ts`, `sql.ts`, `polling.ts`, `partitioning.ts`: Supporting infrastructure

### Configuration

New `VNextPostgresObservabilityConfig` type allows customization of:
- Partitioning mode (auto/timescale/partman/native)
- Number of future partitions to pre-create
- Discovery cache TTL

### Guidance

⚠️ **Recommended for low-volume workloads only** (~100 calls/sec sustained max). Higher volumes should use the ClickHouse adapter. Should use a dedicated Postgres instance, not the primary application database.

## Type of change

## Checklist

https://claude.ai/code/session_01D2qnbd8baZZevs3jmwYjhv

## ELI5

This PR adds a new Postgres-based observability storage option that keeps data by appending only (no updates), splits data into daily partitions, and provides reads/OLAP/querying and discovery caching — designed for low-volume production use. It also found and mitigated a concurrency bug in incremental ("delta") polling that could silently drop rows under concurrent writers.

---

## What's New

**ObservabilityStoragePostgresVNext** — an insert-only Postgres observability domain implementing spans, metrics, logs, scores, and feedback, with:

- PostgresStoreVNext: store class composing the primary PostgresStore with a dedicated observability pool/connection and lifecycle management.
- Insert-only writes using multi-row INSERT ... ON CONFLICT DO NOTHING for idempotency.
- Per-signal partitioning with auto-detection and support for TimescaleDB hypertables, pg_partman, or native Postgres daily range partitions; pre-creation of future partitions.
- Partial indexes for root-span projection (parentSpanId IS NULL) instead of a separate root-span table.
- Discovery cache table with stale-while-revalidate semantics (mastra_observability_discovery).
- Full OLAP surface (aggregates, breakdowns, time series, percentiles) with partition-pruning when callers provide timestamp filters.
- Delta-polling cursor model for incremental reads (opt-in feature flag due to concurrency caveats).
- New modules: ddl.ts, index.ts, traces.ts, tracing.ts, metrics.ts, feedback.ts, scores.ts, logs.ts, discovery.ts, olap.ts, helpers.ts, filters.ts, sql.ts, polling.ts, partitioning.ts, plus tests.

---

## Important Behavior / Guidance

- OLAP queries without a timestamp filter will scan all partitions — callers should include a time range to enable partition pruning (aligns with ClickHouse v-next behavior).
- Observability should use an explicit, separate Postgres connection; reusing the primary connection logs a warning.

---

## Critical Bugs Found & Short-Term Mitigations

1. Delta-polling concurrency data loss:
   - Stress tests showed delta polling can silently skip rows under concurrent writers because bigserial sequence values are assigned outside transactions.
   - Loss rates rose with writer concurrency (examples: 2 writers 0.6–6% → 16 writers 58%).
   - Prototype using pg_snapshot_xmin(pg_current_snapshot()) greatly reduced loss (~58% → ~0.2%) but did not fully eliminate it; a complete fix requires snapshot-xmin plus a short post-poll drain.
   - Short-term mitigation applied in-branch: the observability-delta-polling feature is disabled by default in coreFeatures (remains opt-in). Regression tests for the issue are included on the feature branch.

2. Partition creation race:
   - CREATE TABLE IF NOT EXISTS … PARTITION OF is not atomic; concurrent init() calls could raise 42P07.
   - Fixed by swallowing duplicate-relation errors (isDuplicateRelationError) during partition creation; tests included.

3. Other fixes:
   - Fixed Date.now() per-iteration drift in ensureNativePartitions by freezing baseNowMs.
   - Discovery force-refresh now correctly forces synchronous refresh when ttlSeconds <= 0.
   - Discovery background refresh failures should be routed through the framework logger instead of console.warn (threading logger through discovery helpers remains a checklist item / follow-up).

---

## Configuration

VNextPostgresObservabilityConfig exposes:
- partitioning: { mode?: 'auto' | 'timescale' | 'partman' | 'native'; futureDays?: number; includeYesterday?: boolean } (default futureDays 14)
- discovery: { ttlSeconds?: number }

Delta polling is gated behind the observability-delta-polling feature flag (opt-in by default after mitigation).

---

## Tests & Coverage

- Expanded observability test-utils and test-suite to cover feedback, batchCreateFeedback, conditional skipping of span-update tests for insert-only strategy.
- Unit tests for polling helpers (encodeDeltaCursor, validateCursorId).
- PostgresStoreVNext integration tests assert wiring, init behavior, and partitionMode resolution.
- Regression tests for partition race and delta-polling issues included on the feature branch.

---

## Status & Follow-ups

- Draft status. Checklist items remain (thread logger into discovery helpers to replace console.warn).
- Recommended follow-ups: implement full delta-polling fix (snapshot-xmin + post-poll drain), consider coordinated cross-backend default time-window for OLAP queries, and consider splitting short-term mitigations into separate PR(s) if desired.


intojhanurag:
While reviewing this PR I ran a concurrency stress test against the delta-polling path and wanted to flag what I found, since it changes the framing a bit from what the TODO in `polling.ts` currently says.

The file already documents the cause honestly:

> We use a `bigserial` cursor. Because the sequence is assigned outside of the transaction, a late-committing row whose id falls below the cursor can be permanently skipped by a naive `WHERE cursorId > $after` read. The fix is to cap reads at a safe horizon derived from `pg_snapshot_xmin(pg_current_snapshot())`.

It also says this risk is "effectively zero" for the low-volume target (~100 calls/sec). That part is what my testing didn't bear out.

**What I ran:** N concurrent writers inserting through `createSpan`, a single delta poller draining as it goes, then comparing inserted ids vs ids the poller actually saw.

**What I observed (loss rate by writer count):**
- 1 writer → 0%
- 2 writers → ~0.6% – 6%
- 4 writers → ~1.2% – 14%
- 8 writers → ~2.4% – 19%
- 16 writers → ~58%

The trigger isn't request volume, it's **how many writes overlap in time**. Two agent runs in the same process — well under 100 calls/sec — are enough to lose data silently.

I also prototyped the `pg_snapshot_xmin()` horizon described in the TODO. It drops loss from 58% → ~0.2% but doesn't fully eliminate it; the complete fix needs the horizon plus a small post-poll drain step to pick up rows that committed below the cursor between two polls.

Because the feature flag was on by default, this would have shipped as silent data loss for anyone running >1 concurrent writer. The shortest safe change was to flip `observability-delta-polling` off in `coreFeatures` by default; callers who explicitly opt in still get the current behaviour. The real fix (safe horizon + drain) is a good follow-up PR.

I also fixed a smaller race in `ensureNativePartitions()` — `CREATE TABLE IF NOT EXISTS … PARTITION OF` isn't atomic, so two parallel `init()` calls would surface `42P07`. Wrapped it with the same `isDuplicateRelationError()` swallow already used for base tables and indexes.

Both fixes are on `claude/research-postgres-observability-w0P4C` with regression tests; happy to break them into a separate PR if that's easier to land.

NikAiyer (commented):
**A4 — Partition pruning is available but not enforced by the OLAP API contract**

The timestamp WHERE clauses here (`filters.timestamp?.start` / `filters.timestamp?.end`) are conditional on the caller passing one. In core, `metricsFilterSchema` / `scoresFilterSchema` / `feedbackFilterSchema` are all `.optional()` on the args schemas, and `timestamp` is optional inside them.

Result: any caller can invoke `getMetricAggregate({ name: 'foo' })` (no time range), and the generated SQL becomes `SELECT … FROM mastra_metric_events GROUP BY …` with no WHERE clause — the planner has nothing to prune on, and the query scans every partition that exists for that signal table. The same applies to `getMetricBreakdown` / `getMetricTimeSeries` / `getMetricPercentiles` and the score / feedback equivalents.

This isn't a correctness bug — it's a sharp edge. Once we have months of partitions on a busy instance, an unfiltered aggregate call will pull the entire signal table through the buffer cache, which is exactly what partitioning was supposed to prevent. `listLogs` / `listMetrics` are bounded by `LIMIT`, but they still scan every partition's index in order without a time predicate.

Options to discuss with the team:

1. **Enforce at the storage layer** — default to "last 24h" (or a configurable window) when no `timestamp` filter is supplied for OLAP methods. Backwards-compatible for callers that pass a range; safe by default for callers that don't.
2. **Enforce at the schema/API layer** — make `timestamp` (or at least `timestamp.start`) required on `getMetricAggregate*` / `getScoreAggregate*` / `getFeedbackAggregate*` args. Surfaces the contract change as a type error rather than a silent perf cliff.
3. **Leave as caller responsibility** — document the footgun in `postgresql.mdx` and the relevant TSDoc, and trust the playground/UI to always supply a range.

Filed as a follow-up; will discuss separately and circle back. No code change in this PR.

Refs: `stores/pg/src/storage/domains/observability/v-next/filters.ts:45-58`, `stores/pg/src/storage/domains/observability/v-next/metrics.ts:191,236,287,388` (and corresponding `scores.ts`, `feedback.ts`).

stores/pg/src/storage/domains/observability/v-next/filters.ts
```diff
@@ -0,0 +1,109 @@
...
+export function applyCommonFilters(
+  acc: FilterAccumulator,
+  filters: Record<string, any> | undefined,
+  options: { timestampColumn?: string; prefix?: string } = {},
+): void {
+  if (!filters) return;
+  const tsCol = options.timestampColumn ?? 'timestamp';
+  const prefix = options.prefix ?? '';
+
+  if (filters.timestamp?.start) {
```
NikAiyer, line 49: **A4 — Partition pruning is available but not enforced**

`filters.timestamp` is optional in `metricsFilterSchema` / `scoresFilterSchema` / `feedbackFilterSchema`, and `filters` itself is optional on the args schemas. A caller can invoke `getMetricAggregate({ name: 'foo' })` with no time range and the generated SQL becomes `SELECT ... FROM mastra_metric_events GROUP BY ...` with no WHERE clause — every partition gets scanned. Same applies to `getMetricBreakdown` / `TimeSeries` / `Percentiles` and the score / feedback equivalents. `list*` paths are bounded by `LIMIT` but still walk every partition's index.

Filing for team discussion. Options:

1. Default to a window (e.g. last 24h) at the storage layer when no `timestamp` is supplied.
2. Make `timestamp.start` required on OLAP args schemas in core.
3. Document the footgun and trust callers.

No code change in this PR.

Refs: `filters.ts:45-58`, `metrics.ts:191,236,287,388` (and `scores.ts` / `feedback.ts`).

abhiaiyer91:
**Re: A4 — partition pruning when `timestamp` filter is absent**

Decision: **Option 3 (document the footgun) for this PR.** Defer Option 1 as a coordinated follow-up.

Rationale:

- This is **not a PG-specific regression.** `timestamp` is optional in the shared `commonFilterFields`, so an OLAP query with no time range produces an empty `WHERE` and a full scan on **every** backend. The ClickHouse v-next store has identical behavior. A4 is a property of the cross-backend OLAP API contract, not a defect introduced here.
- **Reject Option 2** (make `timestamp.start` required): that's a breaking change to a shared core schema. It would break the ClickHouse store and existing callers.
- **Defer Option 1** (storage-layer default 24h window): if PG silently injects a 24h window while ClickHouse still scans all-time, identical API calls return *different* results across backends. That's a worse footgun than the one we're fixing. If we want a default window, it should be a coordinated cross-backend change, not a PG-only behavior — good follow-up PR.
- **Adopt Option 3 now**: keeps this PR consistent with the ClickHouse precedent and unblocks merge. For the target workload (~100 calls/sec, dedicated instance, daily partitions) a full scan without a time filter is tolerable.

Action for this PR: add a note in the domain guidance, e.g. *"OLAP queries without a `timestamp` filter scan all partitions; supply a time range to enable partition pruning."*

Follow-ups (separate PRs): (1) coordinated cross-backend default time window if desired; (2) the delta-polling snapshot-xmin horizon fix already tracked.

abhiaiyer91:
**Two cheap fixes / cleanups before leaving draft**

Reviewed the latest branch state. Two of the three earlier nits are already handled — calling them out so the threads can be resolved:

✅ **Already fixed — `Date.now()` per-iteration drift in `ensureNativePartitions`.** `partitioning.ts:121` now freezes `baseNowMs` once before the loop, so a partition pass can no longer cross UTC midnight mid-loop and shift day boundaries. Looks correct.

✅ **Already fixed — discovery force-refresh.** `discovery.ts:157-159` now has the explicit `if (ttlSeconds <= 0) return await refreshing;` force-refresh path. This addresses CodeRabbit's note that `refreshAllDiscoveryCaches({ ttlSeconds: 0 })` wouldn't force a synchronous refresh — that thread can be **resolved**.

🔧 **Still actionable — route discovery refresh failures through the framework logger.** `discovery.ts:104` uses `console.warn` for background refresh failures. The inline comment acknowledges this is because the discovery helpers don't currently take a logger. Before leaving draft, thread a logger through (or accept one in the discovery config) so these failures go through the framework logger like the rest of the store, instead of raw `console.warn`. Cheap, and keeps logging consistent.

Nothing here is blocking the A4 decision above.

NikAiyer (approved):

PR #16632, opened 2026-05-15
feat(observability): add stateless delta polling for observability list endpoints

epinzur:
## Description

This adds stateless delta polling to the existing observability list endpoints.

Page mode remains the default. Clients can now opt into mode=delta and resume from a deltaCursor returned by the same list endpoint.

## What changed

- added shared observability list support for:
    - mode: 'page' | 'delta'
    - after
    - shared delta limit
    - deltaCursor
    - shared page/delta validation
- updated observability list schemas for:
    - traces
    - branches
    - logs
    - metrics
    - scores
    - feedback
- extended the existing list endpoints
    - page mode stays the default when mode is omitted
    - delta mode is explicit via mode=delta
    - page responses include deltaCursor when delta polling is supported
    - delta responses include delta metadata and deltaCursor
- implemented delta-capable list behavior in:
    - in-memory observability store
    - DuckDB observability store
    - ClickHouse v-next observability store
- added runtime capability checks for delta mode
    - stores explicitly advertise which observability lists support delta polling
    - the server rejects unsupported delta requests instead of silently falling back
- added shared store test coverage for the new polling behavior and package-local tests for DuckDB and ClickHouse
- regenerated:
    - client-sdks/client-js/src/route-types.generated.ts
    - packages/cli/src/commands/api/route-metadata.generated.ts

## Behavior notes

- delta polling uses the same filters as page mode
- clients are expected to reuse the same filter set with a deltaCursor
- mode=delta without `after` starts a new delta cursor from the current filtered result set and returns no backfill
- in page mode, deltaCursor represents the newest matching row currently visible for that filter set
- in delta mode, when rows are returned, deltaCursor is the last returned row
- traces and branches only emit newly listed rows in delta mode
- logs, metrics, scores, and feedback emit newly inserted rows in delta mode

## Type of change

- [ ] Bug fix (non-breaking change that fixes an issue)
- [X] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [ ] Test update

## Checklist

- [ ] I have linked the related issue(s) in the description above
- [ ] I have made corresponding changes to the documentation (if applicable)
- [X] I have added tests that prove my fix is effective or that my feature works
- [ ] I have addressed all Coderabbit comments on this PR

## ELI5

[...]

PR #16802, opened 2026-05-19
fix(duckdb): checkpoint WAL after schema init to prevent corruption on dev server restart

mikhael28:
## Problem

On the alpha branch, restarting the dev server after initial setup causes a DuckDB WAL corruption error:

```
INTERNAL Error: Failure while replaying WAL file "mastra.duckdb.wal":
Unhandled exception: Calling DatabaseManager::GetDefaultDatabase with no default database set
```

This does **not** occur on @latest — it's a regression introduced by the delta polling feature (PR #16632).

## Root Cause

The new delta polling migrations add `ALTER COLUMN SET DEFAULT nextval('..._cursor_id_seq')` statements to all observability tables. When the dev server is killed (abrupt process termination), these statements remain in the WAL. On the next startup, DuckDB's WAL replayer tries to resolve `nextval()` sequence references before the catalog is fully initialized, causing the `GetDefaultDatabase` error.

The `executeBatch()` method runs all 50+ DDL and migration statements without transaction wrapping and without flushing the WAL afterward, leaving the database vulnerable to corruption on any non-graceful shutdown.

## Fix

Add an explicit `CHECKPOINT` call after `executeBatch()` in both `init()` and `migrateSpans()`. This flushes the WAL to the main database file immediately after schema changes, ensuring no incomplete operations survive a process kill.

## Changes

- `stores/duckdb/src/storage/domains/observability/index.ts` — Added `CHECKPOINT` after `executeBatch()` in `init()` and after `migrateSignalTables()` in `migrateSpans()`
- `stores/duckdb/src/storage/domains/observability/index.test.ts` — Updated init test to expect the new `CHECKPOINT` call

## Test Plan

- [x] Migration tests pass (`migration.test.ts` — 5/5)
- [x] Init schema test updated and passes
- [x] Existing observability tests unaffected (9 pre-existing delta polling test failures are unchanged)

## ELI5

When the database is first set up, a lot of changes are made all at once. If the server crashes before those changes are properly saved to disk, the database gets confused and crashes next time you start it. This PR makes sure the database saves its work right after making big schema changes, so sudden shutdowns don't corrupt anything.

## Problem

The dev server crashes with a DuckDB WAL (Write-Ahead Log) corruption error when restarted after initial setup: "Failure while replaying WAL file... Calling DatabaseManager::GetDefaultDatabase with no default database set". This regression was introduced by the delta polling feature, which adds ALTER COLUMN SET DEFAULT statements to observability tables. If the server is killed before the WAL is flushed, these statements remain incomplete in the WAL and cause failures when the WAL is replayed on restart.

## Root Cause

The `executeBatch()` method runs 50+ DDL/migration statements without transaction wrapping and without flushing the WAL. This leaves incomplete schema changes in the WAL that can cause failures on non-graceful shutdown.

## Solution

Added explicit `CHECKPOINT` calls to flush the WAL to the main database file immediately after schema changes:
- Added `CHECKPOINT` after `executeBatch()` in `init()`
- Added `CHECKPOINT` after `migrateSignalTables()` in `migrateSpans()`

These changes ensure the WAL is synced to disk before continuing, preventing corruption if the server crashes during or immediately after initialization.

## Changes

**stores/duckdb/src/storage/domains/observability/index.ts**
- Added explicit DuckDB `CHECKPOINT` calls with comments explaining the intent to prevent WAL replay corruption on abrupt shutdowns

**stores/duckdb/src/storage/domains/observability/index.test.ts**
- Updated the initialization/batching test to assert the new `CHECKPOINT` call after `executeBatch()`

## Testing

- Migration tests pass (5/5)
- Init schema test updated and passes
- Existing observability tests unaffected

[Review Change Stack](https://app.coderabbit.ai/change-stack/mastra-ai/mastra/pull/16802?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/stores/duckdb/src/storage/domains/observability/index.test.ts
+++ b/stores/duckdb/src/storage/domains/observability/index.test.ts
@@ -128,7 +128,8 @@ describe('ObservabilityStorageDuckDB', () => {
 
     expect(db.executeBatch).toHaveBeenCalledTimes(1);
     expect(db.executeBatch).toHaveBeenCalledWith([...ALL_DDL, ...ALL_MIGRATIONS]);
-    expect(db.execute).not.toHaveBeenCalled();
+    expect(db.execute).toHaveBeenCalledTimes(1);
+    expect(db.execute).toHaveBeenCalledWith('CHECKPOINT');
   });
 
   // ==========================================================================
--- a/stores/duckdb/src/storage/domains/observability/index.ts
+++ b/stores/duckdb/src/storage/domains/observability/index.ts
@@ -148,6 +148,11 @@ export class ObservabilityStorageDuckDB extends ObservabilityStorage {
     }
 
     await this.db.executeBatch([...ALL_DDL, ...ALL_MIGRATIONS]);
+
+    // Flush the WAL to the main database file so that incomplete WAL entries
+    // from schema migrations (especially ALTER COLUMN SET DEFAULT nextval(...))
+    // don't cause corruption on abrupt process termination (e.g. dev server restart).
+    await this.db.execute('CHECKPOINT');
   }
 
   /**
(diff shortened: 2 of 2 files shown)


epinzur:
closing in favor of https://github.com/mastra-ai/mastra/pull/16803

PR #16811, opened 2026-05-20
fix(core): stop duplicate skill versions for unchanged snapshots

### LINKED ISSUE #16388: [BUG] SkillsStorage creates duplicate versions for unchanged skill snapshots
## Description

Stored skill updates can create a new skill version even when the submitted snapshot is semantically unchanged.

`SkillsStorage.update` currently compares snapshot fields with raw `JSON.stringify` output. That is order-sensitive and does not handle storage round trips consistently for optional fields. With PostgreSQL, JSONB can return object keys in a different order, and optional values such as `license` can round-trip as `NULL`/`null` instead of `undefined`. The result is that metadata or tree entries can be detected as changed even when the skill content did not change.

In practice, repeated no-op publish/update cycles can keep incrementing stored skill versions.

## Reproduction

1. Create a stored skill with metadata such as `{ scope, origin, path }`, a `tree.entries` object with multiple keys, and no `license` value.
2. Read the latest stored version back from storage.
3. Call `skills.update` with the same semantic metadata/tree content but a different object key order, and `license: undefined`.
4. Check `countVersions()` or `getLatestVersion()`.

## Expected behavior

No new version is created for a semantically unchanged stored skill snapshot.

## Actual behavior

A new version is created even though the snapshot content is unchanged. Repeating the same no-op update can keep bumping `versionNumber`.

## Minimal executable test case

I have a fix branch with focused regression tests for InMemory, LibSQL, and PostgreSQL skill storage:

https://github.com/p-meier/mastra/tree/fix/skill-update-diff-stability

The PostgreSQL regression test reproduces the JSONB round-trip case by updating a skill with semantically equal `metadata` and `tree` objects whose key order differs after storage serialization.

intojhanurag:
## Description

`SkillsStorage.update` created a new skill version on every no-op publish/update because it compared snapshots with `JSON.stringify`, which is sensitive to object key order (e.g. PostgreSQL JSONB reordering) and to optional fields round-tripping between `undefined` and `null`. This switches the comparison to a value-based deep equality so semantically identical snapshots no longer bump the version number.

- Added `skillSnapshotFieldValuesEqual` in `@mastra/core` — deep, key-order-independent compare that treats `undefined`/`null`/missing as equal and handles arrays and `Date`.
- Wired the helper into changed-field detection for InMemory, Filesystem, LibSQL, PostgreSQL, and MongoDB skill stores, replacing the `JSON.stringify` comparisons.
- MongoDB now only creates a new version when a field actually changed (previously it always created one when any config field was present).
- Added unit tests for the comparison helper, an InMemory regression test, and a shared `createSkillsTests` suite run against the SQL/Mongo backends.
- Raised the `@mastra/core` peerDependency floor to `>=1.36.0-0` in libsql/mongodb/pg, since they now import the new helper.

## Related Issue(s)

Fixes #16388

## Type of Change

- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [ ] Test update

## Checklist

## ELI5

This PR makes sure a “publish/update” that doesn’t really change a skill won’t create a brand-new skill version. It does that by comparing skill snapshot values more intelligently, instead of using a string-based comparison that can be fooled by key reordering or `undefined` vs `null`.

---

## Overview

Fixes a bug where `SkillsStorage.update()` created duplicate skill versions for semantically unchanged snapshots. Previously, change detection relied on `JSON.stringify`, which is sensitive to object key order and to `undefined`/`null` round-tripping differences across storage backends. The PR introduces a dedicated deep equality helper and applies it across storages so only real value changes produce a new version. Fixes `#16388`.

### Core utility
- Added `skillSnapshotFieldValuesEqual(a, b)` in `@mastra/core` (`packages/core/src/storage/domains/skills/skill-snapshot-field-equal.ts`) and exported it from the skills domain.
  - Key-order independent object comparison
  - Treats `null`, `undefined`, and missing fields as equivalent (nullish semantics)
  - Supports arrays (element-wise) and `Date` (compares timestamps)

### Storage implementations updated (JSON.stringify removal)
- Replaced `JSON.stringify`-based field comparison with `skillSnapshotFieldValuesEqual` in:
  - `InMemorySkillsStorage` (`packages/core/src/storage/domains/skills/inmemory.ts`)
  - `FilesystemSkillsStorage` (`packages/core/src/storage/domains/skills/filesystem.ts`)
  - `SkillsLibSQL` (`stores/libsql/src/storage/domains/skills/index.ts`)
  - `SkillsPG` (`stores/pg/src/storage/domains/skills/index.ts`)
  - `MongoDBSkillsStorage` (`stores/mongodb/src/storage/domains/skills/index.ts`)

### MongoDB behavior change
- Adjusted MongoDB version creation so a new skill version is created **only when at least one provided config field’s value actually changes** (using the new equality), instead of treating provided config keys as changed unconditionally.

## Tests
- Added unit tests for `skillSnapshotFieldValuesEqual` including `undefined`/`null`, key-order independence, nested structures, arrays, and `Date` handling (`packages/core/src/storage/domains/skills/skill-snapshot-field-equal.test.ts`).
- Added/extended InMemory regression coverage to ensure no new version is created for semantically unchanged updates (including reordering and explicit `license: undefined`) (`packages/core/src/storage/domains/skills/inmemory.test.ts`).
- Added a shared `createSkillsTests` suite that validates the no-op update behavior across SQL and MongoDB backends (`stores/_test-utils/src/domains/skills/index.ts`) and wired it into the test factory.

## Packaging / dependency updates
- Raised `@mastra/core` peer dependency floors to support importing the new helper:
  - `stores/libsql`: `>=1.36.0-alpha.8`
  - `stores/mongodb`: `>=1.36.0-alpha.8`
  - `stores/pg`: `>=1.42.1-0`
- Added a changeset bump for `@mastra/core`, `@mastra/libsql`, `@mastra/mongodb`, and `@mastra/pg` capturing the fix.


stores/libsql/package.json
```diff
@@ -43,7 +43,7 @@
     "vitest": "catalog:"
   },
   "peerDependencies": {
-    "@mastra/core": ">=1.34.0-0 <2.0.0-0"
+    "@mastra/core": ">=1.36.0-alpha.8 <2.0.0-0"
```
wardpeet: Don't include alphas this needs to be next version. Use patch

wardpeet (changes requested):

packages/core/src/storage/domains/skills/skill-snapshot-field-equal.ts
```diff
@@ -0,0 +1,25 @@
+export function skillSnapshotFieldValuesEqual(a: unknown, b: unknown): boolean {
```
wardpeet, line 1: I wonder if https://github.com/lukeed/dequal is worth using instead of this

wardpeet (approved):

packages/core/src/storage/domains/skills/skill-snapshot-field-equal.ts
intojhanurag: May be it is worth it , didn't use till now :)

PR #16846, opened 2026-05-20
fix(core): merge thread metadata and working memory on partial updates

### LINKED ISSUE #13990: Bug: Working Memory - multiple data integrity and initialization issues
## Summary

We've identified several interrelated bugs in Working Memory that affect data integrity, initialization, and token efficiency. We're currently using pnpm patches against `@mastra/core@1.10.0` and `@mastra/memory@1.6.1` to work around these issues.

---

### 1. `generateEmptyFromSchema` fails on string schemas, nested objects, and default values

**Location:** `WorkingMemory.generateEmptyFromSchema()` in `@mastra/core`

**Problem:** The method doesn't handle:
- String-type schema input (needs `JSON.parse` first)
- Nested objects with `properties` (only handles flat structures)
- `default` values defined in the schema

**Current behavior:**
```js
// String schema → crash or returns null
// Nested object → gets empty string instead of recursive processing
// Schema with defaults → ignored
```

**Suggested fix:**
```js
generateEmptyFromSchema(schema) {
  try {
    const parsedSchema = typeof schema === 'string' ? JSON.parse(schema) : schema;
    if (!parsedSchema || parsedSchema.type !== "object" || !parsedSchema.properties) return {};

    const TYPE_DEFAULTS = {
      string: "", array: [], number: 0, integer: 0, boolean: false
    };

    const processProperty = (prop) => {
      if (prop.type === "object" && prop.properties) {
        return processProperties(prop.properties);
      }
      if (prop.default !== undefined) return prop.default;
      return TYPE_DEFAULTS[prop.type] ?? "";
    };

    const processProperties = (properties) => {
      const obj = {};
      for (const [key, prop] of Object.entries(properties)) {
        obj[key] = processProperty(prop);
      }
      return obj;
    };

    return processProperties(parsedSchema.properties);
  } catch {
    return {};
  }
}
```

---

### 2. Thread metadata is overwritten instead of merged

**Location:** `AgentLegacyHandler` and `createPrepareMemoryStep` in `@mastra/core`

**Problem:** When updating thread metadata, the existing metadata is completely replaced instead of merged:

```js
// Current (breaks existing metadata fields)
thread: { ...existingThread, metadata: thread.metadata }

// Expected (preserves existing fields)
thread: { ...existingThread, metadata: { ...existingThread.metadata, ...thread.metadata } }
```

**Impact:** Any metadata fields not included in the update are silently lost.

---

### 3. Working Memory does not support partial JSON updates (full overwrite)

**Location:** `updateWorkingMemory` tool execute in `@mastra/core` (MockMemory) and `@mastra/memory`

**Problem:** When the LLM sends a partial JSON update (only changed fields), the entire working memory is replaced. This causes data loss for unchanged fields and wastes tokens by forcing the LLM to re-send the complete JSON every time.

**Suggested fix:** Add shallow merge for JSON-format working memory:

```js
// In updateWorkingMemory execute:
let inputMemory = typeof inputData.memory === "string" 
  ? inputData.memory : JSON.stringify(inputData.memory);

try {
  const parsed = JSON.parse(inputMemory);
  if (parsed && typeof parsed === "object" && !Array.isArray(parsed)) {
    const existingWM = await memory.getWorkingMemory({ threadId, resourceId, memoryConfig });
    if (existingWM) {
      try {
        const existingParsed = JSON.parse(existingWM);
        if (existingParsed && typeof existingParsed === "object" && !Array.isArray(existingParsed)) {
          inputMemory = JSON.stringify({ ...existingParsed, ...parsed });
        }
      } catch {}
    }
  }
} catch {}
```

Also update the tool description to reflect partial update semantics:
```
"Update the working memory. For JSON format, you can send only the fields 
that changed (partial update). Unchanged fields will be preserved automatically."
```

---

### 4. `<working_memory_data>` is empty when no data exists yet (initial template not shown)

**Location:** `getWorkingMemoryToolInstruction()` in `@mastra/core`

**Problem:** When `data` is `null`/`undefined` (first interaction), the `<working_memory_data>` section is empty. The LLM has no reference for the expected structure, making the first `updateWorkingMemory` call unreliable.

**Suggested fix:**
```js
const displayData = data || (hasEmptyWorkingMemoryTemplateObject 
  ? JSON.stringify(emptyWorkingMemoryTemplateObject, null, 2) 
  : null);

// Then use displayData instead of data in the template:
// <working_memory_data>
// ${displayData}
// </working_memory_data>
```

---

### 5. `updateWorkingMemory` tool description instructs LLM to send entire content (token waste)

**Location:** `getWorkingMemoryToolInstruction()` in `@mastra/core`

**Problem:** The instruction says:
> "REMEMBER: the way you update your working memory is by calling the updateWorkingMemory tool with the **entire** JSON content."

This forces the LLM to re-send the complete working memory on every update, even when only one field changed. Combined with issue #3 (no partial merge), this wastes significant tokens.

**Suggested fix** (after implementing #3):
```
- IMPORTANT: For JSON format, you can send only the fields that changed 
  (partial update) - they will be merged with existing data.
```

---

## Affected Versions

- `@mastra/core`: confirmed in 1.0.4 and 1.10.0
- `@mastra/memory`: confirmed in 1.0.0 and 1.6.1

## Workaround

We are currently using pnpm patches to fix all of the above in our project.

## Environment

- `@mastra/core`: 1.10.0
- `@mastra/memory`: 1.6.1
- PostgreSQL: 16.x
- Node.js: 22.x

abhiaiyer91:
## Summary

Thread metadata updates silently dropped existing fields not included in the update payload. Working memory updates in MockMemory replaced all data instead of merging, causing data loss on partial JSON updates.

## Changes

**Thread metadata merge (Bug 2 from #13990)**
- `prepare-memory-step.ts` and `agent-legacy.ts`: Changed metadata updates from full replace to shallow merge (`{ ...(existing ?? {}), ...incoming }`)
- Existing metadata fields are now preserved when updating a thread with partial metadata

**Working memory partial JSON updates (Bug 3 from #13990)**
- `mock.ts`: Added merge semantics in the `updateWorkingMemory` tool for schema-based configs, matching `@mastra/memory` behavior
- Fetches existing WM, parses both as JSON, merges if both are objects, falls back to full-replace for non-JSON/template-based formats
- Fixed MockMemory constructor to preserve `workingMemory` config options (like `schema`) when `enableWorkingMemory: true`

## Test Plan

- Added metadata merge test in `memory-metadata.test.ts` — verifies existing fields survive partial updates
- Added 4 tests in `mock-working-memory-merge.test.ts` — covers schema-based merge, template-based replace, field overwrite, and first-write scenarios
- All 60 memory-related tests pass, typecheck clean

Closes #13990

## ELI5

Previously, updating a part of a conversation's saved data would accidentally erase everything else on that note. This PR makes partial updates keep the existing data and only change the parts you send—so nothing important gets lost.

## Overview

This PR fixes data-loss and initialization bugs in thread metadata and working memory by changing replace semantics into merge semantics where appropriate, improving initial template handling, and preserving working-memory configuration options.

### Thread metadata preservation
- Incoming thread metadata is now merged with existing metadata instead of replacing it, preserving unspecified fields.
- Files updated: prepare-memory-step.ts and agent-legacy.ts (shallow merge applied when persisting metadata).

### Working memory merge semantics (MockMemory)
- MockMemory.updateWorkingMemory now supports schema-based merge semantics:
  - When a schema is configured, the tool attempts to parse existing and incoming values as JSON and deep-merge plain objects (recursive merge), with these rules:
    - Nested plain objects are merged recursively.
    - Arrays are replaced entirely.
    - Setting a key to null deletes that key.
    - Primitives are overwritten directly.
  - When no schema is configured, behavior falls back to full-replace (store incoming value as string/JSON-stringified value).
- The tool's instructions/descriptions are updated to document partial JSON update behavior to reduce token usage.
- MockMemory constructor now preserves provided workingMemory options (e.g., schema, template) when enableWorkingMemory: true instead of overwriting them.

### generateEmptyFromSchema and initialization
- Improvements ensure empty templates are built robustly from string or object schemas and respect property default values so first-interaction working memory shows the expected structure.

### Tests
- Added/updated tests:
  - memory-metadata.test.ts: verifies metadata merge preserves existing fields.
  - mock-working-memory-merge.test.ts: covers schema-based deep-merge behaviors (nested merges, null deletes, array-replace), template/replace behaviors, explicit overwrites, and first-write initialization.
- All memory-related tests pass and typechecking is clean.

## Motivation / Context

Closes `#13990` — prevents data loss on partial updates, improves working-memory initialization and merge semantics so LLMs can send minimal diffs instead of full objects, and preserves configuration when enabling working memory.

[Review Change Stack](https://app.coderabbit.ai/change-stack/mastra-ai/mastra/pull/16846?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/.changeset/perky-sloths-shake.md
+++ b/.changeset/perky-sloths-shake.md
@@ -1,0 +1,9 @@
+---
+'@mastra/core': patch
+---
+
+Fixed thread metadata updates to merge with existing fields instead of replacing them. Previously, updating a thread's metadata would silently drop any fields not included in the update. Now existing metadata fields are preserved when updating.
+
+Fixed MockMemory working memory tool to support partial JSON updates when using schema-based working memory. Previously, sending a partial update would overwrite all existing data. Now for schema-based configs, unchanged fields are preserved automatically (matching @mastra/memory behavior).
+
+Fixed MockMemory constructor to preserve workingMemory config options (like schema) when enableWorkingMemory is true.
--- a/packages/core/src/agent/__tests__/memory-metadata.test.ts
+++ b/packages/core/src/agent/__tests__/memory-metadata.test.ts
@@ -186,6 +186,51 @@ function memoryMetadataTests(version: 'v1' | 'v2') {
       expect(thread?.metadata).toEqual({ client: 'updated' });
     });
 
+    it('should merge metadata with existing fields instead of replacing', async () => {
+      const mockMemory = new MockMemory();
+      const initialThread: StorageThreadType = {
+        id: 'thread-1',
+        resourceId: 'user-1',
+        metadata: { existingField: 'should-persist', client: 'initial' },
+        createdAt: new Date(),
+        updatedAt: new Date(),
+      };
+      await mockMemory.saveThread({ thread: initialThread });
+
+      const agent = new Agent({
+        id: 'test-agent',
+        name: 'Test Agent',
+        instructions: 'test',
+        model: dummyModel,
+        memory: mockMemory,
+      });
+
+      if (version === 'v1') {
+        await agent.generateLegacy('hello', {
+          memory: {
+            resource: 'user-1',
+            thread: {
+              id: 'thread-1',
+              metadata: { client: 'updated' },
+            },
+          },
+        });
+      } else {
+        await agent.generate('hello', {
+          memory: {
+            resource: 'user-1',
+            thread: {
+              id: 'thread-1',
+              metadata: { client: 'updated' },
+            },
+          },
+        });
+      }
+
+      const thread = await mockMemory.getThreadById({ threadId: 'thread-1' });
+      expect(thread?.metadata).toEqual({ existingField: 'should-persist', client: 'updated' });
+    });
+
     it('should not update metadata if it is the same using generate', async () => {
       const mockMemory = new MockMemory();
       const initialThread: StorageThreadType = {
--- a/packages/core/src/agent/agent-legacy.ts
+++ b/packages/core/src/agent/agent-legacy.ts
@@ -398,7 +398,7 @@ export class AgentLegacyHandler {
             (thread.metadata && !deepEqual(existingThread.metadata, thread.metadata))
           ) {
             threadObject = await memory.saveThread({
-              thread: { ...existingThread, metadata: thread.metadata },
+              thread: { ...existingThread, metadata: { ...(existingThread.metadata ?? {}), ...thread.metadata } },
               memoryConfig,
             });
           } else {
--- a/packages/core/src/agent/workflows/prepare-stream/prepare-memory-step.ts
+++ b/packages/core/src/agent/workflows/prepare-stream/prepare-memory-step.ts
@@ -151,7 +151,7 @@ export function createPrepareMemoryStep<OUTPUT = undefined>({
           (thread.metadata && !deepEqual(existingThread.metadata, thread.metadata))
         ) {
           threadObject = await memory.saveThread({
-            thread: { ...existingThread, metadata: thread.metadata },
+            thread: { ...existingThread, metadata: { ...(existingThread.metadata ?? {}), ...thread.metadata } },
             memoryConfig,
           });
         } else {
--- a/packages/core/src/memory/mock-working-memory-merge.test.ts
+++ b/packages/core/src/memory/mock-working-memory-merge.test.ts
@@ -1,0 +1,87 @@
+import { describe, it, expect } from 'vitest';
+import { z } from 'zod/v4';
+import { MockMemory } from './mock';
+
+describe('MockMemory working memory merge semantics', () => {
+  const threadId = 'thread-1';
+  const resourceId = 'resource-1';
+
+  async function setupMemory(useSchema: boolean) {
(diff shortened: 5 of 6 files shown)


TylerBarnes (approved):
Thread metadata change looks valid to me after tracing the call path. This is not changing `saveThread` semantics globally; the agent path loads the existing thread and then constructs the full thread payload, so it should preserve existing metadata when applying runtime thread metadata. OM should be okay because it updates thread metadata via `updateThread` and `setThreadOMMetadata(thread.metadata, ...)`, which already preserves existing metadata.

I agree with CodeRabbit’s constructor concern: `template: workingMemoryTemplate` should be conditional so an existing `options.workingMemory.template` is not overwritten with `undefined`.

One additional non-blocking thing: the MockMemory schema merge only does a shallow object merge, while `@mastra/memory`’s working memory tool uses `deepMergeWorkingMemory` semantics: nested objects merge, `null` deletes, arrays replace. Since this PR says MockMemory should match `@mastra/memory`, it’d be better to either match that behavior or narrow the claim/tests to shallow partial updates.

Approving since the core thread metadata fix is legitimate and the remaining MockMemory items are small enough for the author to decide whether to handle here or follow up.

PR #16856, opened 2026-05-20
fix(memory): prevent mutation of cached embeddings

Akash504-ai:
## Description

Fixes a mutation safety issue in semantic recall embedding caching.

Previously, embeddings stored in the global embedding cache were returned by reference. This meant downstream consumers could accidentally mutate cached vectors, causing future cache reads to return corrupted embeddings.

This PR fixes the issue by:

* storing frozen copies of embeddings in the global cache
* returning defensive copies when reading cached embeddings
* updating cache typings to use `ReadonlyArray<number>`
* adding regression test coverage to ensure cached embeddings cannot be mutated externally

## Related issue(s)

N/A

## Type of change

* [ ] Documentation update
* [ ] Code refactoring
* [ ] Performance improvement
* [x] Test update

## Checklist

* [ ] I have linked the related issue(s) in the description above

## ELI5

This PR fixes a bug where cached embeddings (arrays of numbers used for similarity searches) could be accidentally modified by code using them, which would corrupt the cache. The fix makes the code return copies of cached embeddings instead of the originals, and stores frozen copies in the cache to prevent accidental changes.

### Type safety improvement
- Updated the `globalEmbeddingCache` type definition to use `ReadonlyArray<number>` instead of `number[]`, enforcing immutability at the TypeScript level for cached embeddings.

### Defensive copying in embedding retrieval
- Modified `SemanticRecall.embedMessageContent` to return cloned embeddings (via spreading) instead of returning cached references directly, preventing downstream mutation of globally cached values.

### Safe cache storage
- When caching newly generated embeddings, the values are now stored as frozen clones using `Object.freeze([...embedding])` rather than raw arrays, adding an extra layer of protection against accidental modification.

### Regression test coverage
- Added a new unit test in the "Embedding Caching" suite that verifies cached embeddings are returned as defensive copies. The test calls the processor twice with identical input, mutates the first result, and confirms the second cached result remains unchanged (and the embedder is invoked only once).

[Review Change Stack](https://app.coderabbit.ai/change-stack/mastra-ai/mastra/pull/16856?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


abhiaiyer91:
Thanks for the contribution! We reviewed the defensive-copy approach and agree it's technically correct — returning cached embeddings by reference is a potential footgun.

However, after tracing the two consumers of `embedMessageContent()`:

1. **`vector.query()`** — takes the embedding as a read-only `queryVector` for search. No mutation.
2. **`vector.upsert()`** — collects embeddings into an array and passes them to the vector store for serialization. No mutation.

There's no code path in the repo that actually mutates a cached embedding today. The bug is theoretical, and the fix adds per-read allocation overhead (spreading 256–3072 element float arrays on every cache hit) to guard against it.

We're going to close this for now since it's solving a problem that doesn't exist in our codebase. If we ever see a code path that mutates embeddings after retrieval, we'll revisit. Appreciate you thinking about cache safety though!

PR #16911, opened 2026-05-21
fix(gcs): populate mimeType on stat() so read_file surfaces media

BrandiATMuhkuh:
## Summary

- `GCSFilesystem.stat()` was returning `FileStat` without `mimeType`, even though the GCS object's `Content-Type` was already on the metadata response. The workspace `read_file` tool gates its native media-part path on `stat.mimeType` (`packages/core/src/workspace/tools/read-file.ts:188`, with `buildMediaTypeCheck` short-circuiting on `!mimeType` at `:120`), so images and PDFs stored in GCS were returned as metadata-only text instead of being surfaced to the model. The local filesystem works because its `stat()` populates `mimeType` from the file extension (`packages/core/src/workspace/filesystem/fs-utils.ts:304`).
- Fix: in `workspaces/gcs/src/filesystem/index.ts` `stat()`, set `mimeType` from `metadata.contentType` when present (this is what `writeFile` already writes on upload at `:387`), falling back to the existing `getMimeType(path)` extension helper for objects uploaded out-of-band.
- No extra GCS round-trip — `file.getMetadata()` was already being called; we just read a field that was already on the wire.

## Repro (before this PR)

1. Upload an image to a GCS-backed workspace (e.g. `workspace.writeFile('photo.png', bytes)`).
2. Ask an agent with the `read_file` tool to read it.
3. The tool returns `"… (N bytes, undefined) — binary file not readable as text. Pass an explicit encoding…"` instead of the image as a viewable media part. The same workflow on `LocalFilesystem` returns the image correctly.

## Test plan

- [x] `pnpm --filter @mastra/gcs test:unit` — existing `stat()` unit test updated to assert `mimeType`; two new tests cover (a) Content-Type → `mimeType` for image objects, (b) extension-based fallback when Content-Type is absent.
- [x] Updated `index.integration.test.ts` "can get file stats" to assert `mimeType === 'text/plain'` and added a PNG upload case asserting `mimeType === 'image/png'` (runs against fake-gcs-server / real GCS).
- [x] `pnpm --filter @mastra/gcs lint` clean.
- [ ] Manual: read an image from a GCS workspace via an agent and confirm it now arrives as a native media part.

## ELI5

When you upload an image or PDF to Google Cloud Storage, the system now knows the file type and will show it as a picture or document instead of trying to display it as confusing text. It reads the Content-Type GCS already stores (or falls back to the file extension) and returns that from stat().

---

## Overview

This PR ensures GCS-backed workspaces expose file MIME types via GCSFilesystem.stat(), allowing the workspace read_file tool to surface images/PDFs as native media parts (same behavior as LocalFilesystem). Previously stat() omitted mimeType even when GCS object metadata included Content-Type, causing media to be treated as non-readable text.

## Changes

- workspaces/gcs/src/filesystem/index.ts
  - Populate FileStat.mimeType: use metadata.contentType when present, otherwise fall back to existing getMimeType(path) (extension-based).
  - No additional GCS API round-trip: uses the already-called file.getMetadata().
  - Added inline comment noting read_file gates media handling on stat.mimeType.

- Tests
  - workspaces/gcs/src/filesystem/index.test.ts: updated mocks and added unit tests verifying:
    - mimeType is taken from Content-Type for image objects
    - mimeType falls back to extension-based detection when Content-Type is missing
  - workspaces/gcs/src/filesystem/index.integration.test.ts: updated integration test to assert text/plain for text files and added a PNG upload case asserting image/png.

- Release metadata
  - .changeset/smooth-turkeys-matter.md: documented as a patch bump for `@mastra/gcs` describing the fix.

## Impact

- Fixes agent/tooling behavior so images and PDFs in GCS are returned as media parts rather than metadata-only text.
- No API perf regression (no extra GCS round-trip).
- Lint passes for `@mastra/gcs`. Unit and integration tests added/updated; manual agent verification listed as pending.

[Review Change Stack](https://app.coderabbit.ai/change-stack/mastra-ai/mastra/pull/16911?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


philipandersson (approved):

philipandersson:
@abhiaiyer91 I believe we have this issue for S3 and Azure as well

wardpeet (approved):

PR #16930, opened 2026-05-21
fix: order agent responses after input history

### LINKED ISSUE #16893: [BUG] Agent response message gets stale createdAt, mis-ordering the prompt and causing duplicate tool calls (1.34.0 unaffected, 1.35.0+ affected)
## Summary

In `@mastra/core` **1.35.0+** (1.34.0 is unaffected), the agent loop stamps the
assistant **response message** with `currentStep.createdAt` (the step *creation* time).
For a long prior conversation, `MessageList.generateCreatedAt` inflates input-message
timestamps past real wall-clock time, so the response message ends up with a `createdAt`
*older* than the latest input messages. `MessageList.addOne` sorts by `createdAt`, so
the response message is inserted into the **middle** of the history instead of at the
end.

On the next loop step the model's prompt no longer has its own `assistant(tool-call)` +
`tool(result)` as the latest turn — so the model **re-issues the same tool call**.

## Describe the Bug

When an agent runs a tool inside a multi-step loop with a non-trivial prior conversation:

1. Step 0: the model emits a `tool-call`; mastra builds the assistant response message
   via `buildMessagesFromChunks(...)`.
2. The response message is created with `createdAt: currentStep.createdAt`.
3. `currentStep.createdAt` was set at *step construction time* — before the LLM stream
   even started.
4. Meanwhile, every input message added without an explicit `createdAt` gets a synthetic
   `lastCreatedAt + 1` timestamp from `MessageList.generateCreatedAt`. With a long
   conversation, this 1ms-per-message inflation pushes the timeline *past* real
   wall-clock time.
5. The response message's `createdAt` is therefore older than the trailing input
   messages. `MessageList.addOne` sorts by `createdAt` and places the response message
   in the middle of the history.
6. On step 1, the prompt passed to the model has the `assistant(tool-call)` /
   `tool(result)` messages buried mid-history — not as the latest turn.
7. The model treats the user's request as still unanswered and re-issues the tool call.

This is a regression: in 1.34.0 the response message's `createdAt` was set to
`new Date()` (build time), which is at or after the input messages, so ordering was
correct.

## Affected Code

Call site — `packages/core/src/loop/workflows/agentic-execution/llm-execution-step.ts`:

```ts
// currentStep is constructed before the LLM stream starts:
const currentStep: { messageId: string; createdAt: Date; /* ... */ } = {
  messageId: currentMessageId,
  createdAt: new Date(),   // <-- step *creation* time
  // ...
};

// ...later, after the stream completes:
const builtMessages = buildMessagesFromChunks({
  chunks: collectedChunks,
  messageId: currentStep.messageId,
  responseModelMetadata: buildResponseModelMetadata(runState, currentStep.model),
  tools: currentStep.tools,
  createdAt: currentStep.createdAt,   // <-- stale timestamp passed here
});
```

`buildMessagesFromChunks` — `packages/core/src/loop/workflows/agentic-execution/build-messages-from-chunks.ts`:

```ts
export function buildMessagesFromChunks({
  chunks,
  messageId,
  responseModelMetadata,
  tools,
  createdAt = new Date(),   // <-- default would be correct; the explicit arg overrides it
}: { /* ... */ }): MastraDBMessage[] {
  // ...
  return [{ role: 'assistant', content: { /* ... */ }, createdAt }];
}
```

`MessageList.addOne` sorts `this.messages` by `createdAt`, so a stale `createdAt`
re-positions the message; `MessageList.generateCreatedAt` assigns `lastCreatedAt + 1`
to input messages lacking an explicit timestamp.

The same call site is present in the **published npm package's bundled output**
(`dist/`), so the regression affects installed packages, not just the TypeScript source.

## Historical Context

Verified against the published source at the `@mastra/core@1.34.0` and
`@mastra/core@1.35.0` git tags:

- `1.34.0`: `buildMessagesFromChunks` had **no `createdAt` parameter** at all. The
  response message's `createdAt` was hardcoded to `new Date()` inside the function
  (`build-messages-from-chunks.ts`), i.e. response build time. `currentStep` had no
  `createdAt` field, and the call site passed only `chunks` / `messageId` /
  `responseModelMetadata` / `tools` (`llm-execution-step.ts`). Ordering was correct.
- `1.35.0+`: a `createdAt = new Date()` parameter was added to
  `buildMessagesFromChunks`, `currentStep` gained a `createdAt` field (set to
  `new Date()` at step construction time), and the call site started passing
  `createdAt: currentStep.createdAt` — overriding the safe default with a stale value.

## Reproduction

**Minimal reproducible example:** https://gist.github.com/technote-space/ca34d79ecbd4558bf89ebb76f03f5ed2

Reproduces reliably on `@mastra/core@1.36.0` with an Amazon Bedrock model. The prompt
mis-ordering observed in `prepareStep` is deterministic; the resulting *duplicate* tool
call depends on how the model reacts to the mis-ordered prompt, so it does not occur on
every single run, but it occurs reliably across runs.

<details>
<summary><code>package.json</code></summary>

```json
{
  "name": "mastra-repro-createdat",
  "private": true,
  "type": "module",
  "scripts": { "start": "tsx repro.ts" },
  "dependencies": {
    "@ai-sdk/amazon-bedrock": "4.0.107",
    "@aws-sdk/credential-providers": "3.1037.0",
    "@mastra/core": "1.36.0",
    "ai": "6.0.184",
    "zod": "4.3.6"
  },
  "devDependencies": { "tsx": "^4.19.0", "typescript": "^6.0.0" }
}
```

</details>

<details>
<summary><code>repro.ts</code></summary>

```ts
/**
 * Minimal reproduction: agent response message gets a stale `createdAt`,
 * which mis-orders the conversation passed to the model on the next loop step.
 */
import { createAmazonBedrock } from "@ai-sdk/amazon-bedrock";
import { fromNodeProviderChain } from "@aws-sdk/credential-providers";
import { z } from "zod";

const MASTRA_VERSION = "1.36.0";

// Number of prior conversation messages. The bug appears once this is large
// enough that generateCreatedAt's 1ms-per-message inflation outruns real
// elapsed wall-clock time. With a short history (e.g. 2) it does NOT reproduce.
const HISTORY_COUNT = 80;
const TRIGGER = "Please echo the exact word: hello";

const bedrock = createAmazonBedrock({
  region: process.env.AWS_BEDROCK_REGION ?? "ap-northeast-1",
  credentialProvider: fromNodeProviderChain(),
});

const echoTool = createTool({
  id: "echo",
  description: "Echoes back the given message.",
  inputSchema: z.object({ message: z.string() }),
  execute: async ({ context }) => ({ echoed: context.message }),
});

const agent = new Agent({
  id: "repro",
  name: "repro",
  instructions:
    "For every user request, call the `echo` tool exactly once, then reply with a single short sentence.",
  model: bedrock(process.env.BEDROCK_MODEL_ID ?? "jp.anthropic.claude-sonnet-4-6"),
  tools: { echo: echoTool },
});

// A long prior conversation. None of these carry an explicit `createdAt`, so
// MessageList.generateCreatedAt assigns each one `lastCreatedAt + 1` — a
// synthetic timestamp — inflating the timeline past real wall-clock time.
const messages: { role: "user" | "assistant"; content: string }[] = [];
for (let i = 0; i < HISTORY_COUNT; i++) {
  messages.push({
    role: i % 2 === 0 ? "user" : "assistant",
    content: `filler conversation message #${i}`,
  });
}
messages.push({ role: "user", content: TRIGGER });

type PartLike = { type?: string; text?: string; toolName?: string };

let inspected = false;

const stream = await agent.stream(messages, {
  prepareStep: (args: { stepNumber: number; messageList?: any }) => {
    // Inspect the prompt built for the step that runs *after* the tool call.
    if (inspected || args.stepNumber < 1) return undefined;
    inspected = true;

    const msgs: { role?: string; content?: unknown }[] =
      args.messageList?.get?.all?.aiV5?.model?.() ?? [];
    const total = msgs.length;

    const hasToolCall = (m: { content?: unknown }) =>
      Array.isArray(m.content) &&
      (m.content as PartLike[]).some((p) => p.type === "tool-call");
    const isTrigger = (m: { role?: string; content?: unknown }) => {
      if (m.role !== "user") return false;
      const text = Array.isArray(m.content)
        ? (m.content as PartLike[]).map((p) => p.text ?? "").join("")
        : String(m.content ?? "");
      return text.includes("hello");
    };

    const toolCallIdx = msgs.findIndex((m) => m.role === "assistant" && hasToolCall(m));
    const triggerIdx = msgs.findIndex(isTrigger);

    console.log(`\n=== prepareStep (stepNumber=${args.stepNumber}) ===`);
    console.log(`@mastra/core              : ${MASTRA_VERSION}`);
    console.log(`prior history messages    : ${HISTORY_COUNT}`);
    console.log(`total messages in prompt  : ${total}`);
    console.log(`triggering user message   : index ${triggerIdx}`);
    console.log(`step-0 assistant tool-call: index ${toolCallIdx}  (expected: ${total - 2})`);

    if (toolCallIdx !== -1 && triggerIdx !== -1 && toolCallIdx < triggerIdx) {
      console.log(
        `\n[BUG REPRODUCED] step-0 tool-call message (index ${toolCallIdx}) is placed BEFORE` +
          ` the triggering user message (index ${triggerIdx}).`,
      );
    } else {
      console.log(
        `\n[OK] step-0 tool-call message is correctly placed after the triggering user message.`,
      );
    }
    return undefined;
  },
});

const output = await stream.getFullOutput();

console.log(`\n=== final output ===`);
console.log(`steps        : ${output.steps?.length}`);
console.log(`finishReason : ${output.finishReason}`);
const toolCallTotal = output.steps?.reduce(
  (n: number, s: { toolCalls?: unknown[] }) => n + (s.toolCalls?.length ?? 0),
  0,
);
console.log(`total tool calls across steps: ${toolCallTotal}  (expected: 1)`);
```

</details>

`repro.ts` builds an Agent with one `echo` tool, prepends 80 filler conversation
messages (none carry an explicit `createdAt`), then sends one trigger message. In
`prepareStep` (stepNumber >= 1) it inspects `messageList.get.all.aiV5.model()` and
checks whether the step-0 `assistant(tool-call)` message sits *before* the triggering
user message.

Observed output (`@mastra/core@1.36.0`):

```
=== prepareStep (stepNumber=1) ===
@mastra/core              : 1.36.0
prior history messages    : 80
total messages in prompt  : 83
triggering user message   : index 82
step-0 assistant tool-call: index 10  (expected: 81)

[BUG REPRODUCED] step-0 tool-call message (index 10) is placed BEFORE the triggering user message (index 82).

=== final output ===
steps        : 3
finishReason : stop
total tool calls across steps: 2  (expected: 1)
```

When the history is ordered correctly, the last two messages of the step-1 prompt
should be the step-0 `assistant(tool-call)` and its `tool(result)`, so the tool-call
message belongs at index `total - 2` (= 81 here). Instead it landed at index 10, and
the model issued the `echo` tool call **twice** (`total tool calls: 2`, `steps: 3`).

Lowering `HISTORY_COUNT` to a small value (e.g. 2) makes the bug disappear — confirming
the cause is timestamp inflation, not the conversation content.

## Environment

- `@mastra/core`: 1.35.0, 1.36.0 (1.34.0 unaffected)
- Model: Amazon Bedrock — `jp.anthropic.claude-sonnet-4-6` (via `@ai-sdk/amazon-bedrock`)
- AI SDK: v6
- Node.js: v24.15.0
- OS: macOS 15.7.4 (Darwin 24.6.0)

## Expected Behavior

The assistant response message should be ordered *after* the input messages that
preceded it, so the model sees its own tool call as the latest turn and does not
re-issue it.

## Proposed Fix

Remove the explicit `createdAt: currentStep.createdAt` from the `buildMessagesFromChunks`
call in `llm-execution-step.ts`, letting the function use its `createdAt = new Date()`
default. This restores the build-time ordering behavior used in 1.34.0.

Verified locally: removing that single line from the bundled `dist/` output of the
installed 1.35.0 package eliminated the mis-ordering and, in this repro, eliminated the
duplicate tool call.

Rohithmatham12:
## Summary
- stop passing stale step-creation timestamps into `buildMessagesFromChunks`
- let response messages use build-time timestamps so they sort after prior input history
- add a regression test for long synthetic histories where input timestamps can outrun step creation time
- add a patch changeset for `@mastra/core`

Fixes #16893

## Tests
- `./node_modules/.bin/vitest run src/loop/workflows/agentic-execution/llm-execution-step.test.ts` from `packages/core`
- `./node_modules/.bin/vitest run src/loop/workflows/agentic-execution/build-messages-from-chunks.test.ts src/agent/message-list/tests/message-list-ordering.test.ts` from `packages/core`

## ELI5 - What This PR Does

The PR fixes a bug where assistant messages in conversations were getting "stuck" in the middle of the message list instead of appearing at the end. This happened because they were given old timestamps (from when the step started), so they sorted before newer input messages. As a result, the AI couldn't see its own recent responses and would keep trying the same tool calls over and over. The fix makes response messages use the current time when they're built, so they naturally appear at the end where they belong.

### Changeset
- **`.changeset/fix-agent-response-createdat.md`** — Adds a patch-level changeset for `@mastra/core` documenting the fix for agent response message ordering by timestamping responses at build time.

### Test Updates
- **`packages/core/src/loop/workflows/agentic-execution/llm-execution-step.test.ts`** — Adds regression test `orders response messages after long input history with synthetic timestamps` that validates correct message ordering with 80+ alternating user/assistant messages. Uses fake timers to verify that assistant responses are appended after the triggering input message even when synthetic timestamps for prior history have advanced past real time.

## Root Cause & Fix
The regression was introduced in `@mastra/core` 1.35.0 when `llm-execution-step.ts` began passing `currentStep.createdAt` (set at step creation time) to `buildMessagesFromChunks()`. This stale timestamp prevented response messages from sorting after later input messages, breaking message order and causing the model to re-issue tool calls. The fix removes this timestamp parameter, allowing response messages to receive build-time timestamps that correctly place them at the end of the conversation history.


Rohithmatham12:
Opened this as a focused fix for #16893 and rebased it onto the latest `main` after resolving the nearby test conflict.

Local checks run from `packages/core`:

- `./node_modules/.bin/vitest run src/loop/workflows/agentic-execution/llm-execution-step.test.ts` — 18 tests passed
- `./node_modules/.bin/vitest run src/loop/workflows/agentic-execution/build-messages-from-chunks.test.ts src/agent/message-list/tests/message-list-ordering.test.ts` — 46 tests passed

This should be ready for maintainer review. Could someone please approve the pending workflows when convenient?

Rohithmatham12:
Rebased this branch onto the latest `main` and resolved the conflict in `packages/core/src/loop/workflows/agentic-execution/llm-execution-step.test.ts` by keeping both the upstream streamed-tool transform regression and this PR's long-history response ordering regression.

Local checks run from `packages/core`:

- `./node_modules/.bin/vitest run src/loop/workflows/agentic-execution/llm-execution-step.test.ts` — 1 file passed, 19 tests passed
- `./node_modules/.bin/vitest run src/loop/workflows/agentic-execution/build-messages-from-chunks.test.ts src/agent/message-list/tests/message-list-ordering.test.ts` — 2 files passed, 47 tests passed
- `git diff --check upstream/main..HEAD` — passed

This should be ready for maintainer review again.

LekoArts:
Fixed by https://github.com/mastra-ai/mastra/pull/16913

PR #16937, opened 2026-05-21
feat(channels): tool display modes, streaming text, adaptive typing status, and driver refactor

CalebBarnes:
## Summary

A focused overhaul of the agent channels UX surface and its implementation. The user-facing change is a richer set of knobs for how an agent renders into chat platforms (Slack, Discord, etc.); the internal change is splitting the previously-monolithic `consumeAgentStream` into two focused drivers behind a thin dispatch.

### `toolDisplay` modes

New `ChannelAdapterConfig.toolDisplay` controls how tool calls render:

- `'cards'` (default) — per-tool running/result cards in rich Block Kit.
- `'text'` — per-tool running/result messages as plain text (replaces the old `cards: false`).
- `'timeline'` — every tool gets its own task row in a live streaming widget with status icons and args. Requires `streaming: true`.
- `'grouped'` — all tools in the run collapse into a single streaming widget; args fold inline into the title and successful results are suppressed for an at-a-glance summary (errors keep their full text). Requires `streaming: true`.
- `'hidden'` — tools run silently; only the typing indicator shows work.
- **Function form (`ToolDisplayFn`)** — `(event, ctx) => { kind: 'post', message } | { kind: 'stream', chunk } | undefined` for full control. `'post'` results post a discrete message (closing/reopening the streaming session when needed); `'stream'` results push a `task_update`/`plan_update` into the active streaming widget; `undefined` skips the event.

Invalid combinations warn once and downgrade: `streaming: false` + `'timeline'`/`'grouped'` → `'cards'`. `streaming: true` + `'cards'`/`'text'` is valid and uses the streaming driver's close/post/reopen lifecycle so cards still appear cleanly between streamed text.

### Streaming text

Opt-in per-adapter `streaming` flag (`boolean | { updateIntervalMs?: number }`) that pushes the agent's text deltas into the platform message progressively via the Chat SDK.

### Adaptive typing indicator

The platform typing status now reflects what the agent is doing:

- `is working…` at run start
- `is thinking…` during reasoning
- `is typing…` while generating text
- `is calling {toolName}…` while a tool runs
- `is saving to memory…` / `is recalling memory…` during memory work
- `is requesting approval for {toolName}…` while a tool is suspended

Coalesced so the platform API isn't called on every delta, and skipped while a streaming session is open (the widget itself conveys progress). Built-in channel tools (`add_reaction`, `update_status`, etc.) are filtered so they never surface as `is calling add_reaction…`.

New `ChannelAdapterConfig.typingStatus` (`boolean | (chunk, ctx) => string | false`, default `true`) lets you suppress all typing indicators or set custom copy per chunk. Compose with the exported `defaultTypingStatus` helper:

```ts
typingStatus: (chunk, ctx) => {
  if (chunk.type === 'tool-call' && chunk.payload.toolName === 'searchDocs') {
    return 'is searching docs…';
  }
  return defaultTypingStatus(chunk, ctx);
},
```

### Observational-memory lifecycle in streaming widgets

`data-om-buffering-*` and `data-om-activation` chunks are routed into the active streaming session in `'timeline'`/`'grouped'` modes as their own task rows (e.g. `Saved to memory (10x)` with `12.4k message tokens → 1.2k memory tokens`), so memory work is visible alongside tool calls.

### Slack provider defaults

`@mastra/channel-slack` now defaults to `streaming: true` + `toolDisplay: 'grouped'` for an opinionated, polished out-of-box experience. Overridable per-adapter.

### Bug fixes along the way

- **Parallel same-tool approval** — two parallel calls to the same tool with `requireApproval: true` no longer clobber each other (pending entries keyed by `toolCallId` instead of `toolName`).
- **Tool error rendering** — failing tools emit a closing task update in `'timeline'`/`'grouped'` modes (previously the row stayed `in_progress` and rendered as ⚠ at session close) and edit their card in `'cards'`/`'hidden'` modes. The error text is inlined into the task `details` (with a ⚠ glyph) while the task itself stays `status: 'complete'` so a single tool failure doesn't flip the overall plan header to an error state.
- **Slack DM tool-approval routing** — clicks on approval buttons in DMs now route to the correct thread (previously could land in the wrong thread when multiple DMs were open).
- **Stuck typing status at run end** — cleared on Slack at finish/error/abort.
- **Signal-aware message boundaries** — when a `data-user-message` signal echoes mid-reply, any in-flight text is flushed first so the agent's response renders as a new message after the user's signal.
- **Stronger stay-silent prompt** — the channel input processor's non-DM system message explicitly calls out anti-patterns (bracketed status notes, "Got it"/"Noted" acknowledgments, apologizing for silence) and points the model at `add_reaction` for silent acknowledgments.
- **Logger propagation** — the Mastra logger is now propagated into `AgentChannels` on register so channel-level logs flow through the configured logger.

## Internal refactor

Adding the streaming-text path could have meant bolting a second rendering pipeline (live `StreamingPlan` widget) onto the existing static `for await` loop in `consumeAgentStream` (discrete `sdkThread.post`/`editMessage` calls), with branches on `streaming` and `toolDisplay` at every chunk type. Instead this PR splits the consumer into two focused drivers behind a thin dispatch, so neither path is muddied by the other:

- **`chat-driver-streaming.ts`** — consumes `AgentChunkType`, drives a `StreamingPlan`. Lazy session open via `pushToSession`, OM lifecycle widgets, approval-card pause-and-post, error/tripwire/file out-of-band posting.
- **`chat-driver-static.ts`** — consumes `AgentChunkType`, posts/edits discrete Block Kit cards. Handles `'cards'`/`'text'`/`'hidden'` modes.
- **`stream-helpers.ts`** — `ToolTracker` (cross-chunk correlation keyed by `toolCallId`), `extractErrorMessage`, `renderBuiltInToolEvent`, `chunkToFallbackMessage`, and shared `postStreamError` / `postTripwire` / `postFileAttachment` helpers (both drivers honor `formatError`).
- **`om.ts`** — OM data part types + `renderOmTaskUpdate` + `formatTokens`. OM types pulled out of `types.ts` for cohesion.
- **`inline-media.ts`** — inline-link/media helpers extracted from `agent-channels.ts` so the class file holds only the `AgentChannels` class.
- **`types.ts`** — consumer-facing types (`PostableMessage`, `ChannelAdapterConfig`, `ToolDisplay*`, `ChannelHandlers`, etc.) extracted from `agent-channels.ts`.

`consumeAgentStream` itself is now a ~30-line dispatch: resolve config, wrap the stream with a typing-status pass-through async generator, hand off to the chosen driver.

`agent-channels.ts` shrunk from ~1940 lines to ~1527 lines (driver loop + helpers + types extracted).

## Breaking changes

- `ChannelAdapterConfig.cards: boolean` removed. Migrate `cards: false` → `toolDisplay: 'text'`.
- `ChannelAdapterConfig.formatToolCall` removed. Migrate `formatToolCall: (info) => msg` → `toolDisplay: (event) => event.kind === 'result' ? { kind: 'post', message: msg } : undefined`.

## Example

```ts
import { defaultTypingStatus } from '@mastra/core/channels';

const agent = new Agent({
  name: 'support-bot',
  channels: {
    slack: {
      streaming: true,
      toolDisplay: 'grouped',
      typingStatus: false, // widget already conveys progress
    },
    discord: {
      typingStatus: (chunk, ctx) => {
        if (chunk.type === 'tool-call' && chunk.payload.toolName === 'searchDocs') {
          return 'is searching docs…';
        }
        return defaultTypingStatus(chunk, ctx);
      },
      toolDisplay: event => {
        if (event.kind !== 'result') return undefined;
        return { kind: 'post', message: `🛠 ${event.toolName} → ${event.resultText}` };
      },
    },
  },
});
```

## Test plan

- `pnpm --filter @mastra/core build:lib` — clean
- `cd packages/core && pnpm exec vitest run src/channels` — **122/122 pass**
- `cd channels/slack && pnpm build && pnpm exec vitest run` — **63/63 pass**
- Smoke tested in Slack across all `toolDisplay` modes, streaming on/off, parallel tool calls, tool approval flow, OM buffering/activation, and tool errors.

## Changesets

- `@mastra/core` — minor
- `@mastra/channel-slack` — minor

## ELI5

This PR makes agent chat channels show tool actions more clearly, lets agents stream their text as they generate it, and displays smarter typing/status indicators so users can see what the agent is doing in real time. Internally it splits the streaming vs static rendering logic to simplify the code and enable these features.

### Tool Display Modes
- Adds ChannelAdapterConfig.toolDisplay with modes: 'cards' (default), 'text', 'timeline', 'grouped', and 'hidden'.
- Adds a function form ToolDisplayFn returning { kind: 'post', message } | { kind: 'stream', chunk } | undefined for custom per-tool rendering; unsupported mode/streaming combos warn and gracefully downgrade.
- Deprecates ChannelAdapterConfig.cards (migrate: cards: false → toolDisplay: 'text'; cards: true → toolDisplay: 'cards').

### Streaming Text Deltas
- Per-adapter streaming option: boolean | { updateIntervalMs?: number } to push progressive text deltas via the Chat SDK.
- Static (buffered) behavior preserved when streaming is disabled.
- `@mastra/channel-slack` defaults changed to streaming: true and toolDisplay: 'grouped' (overridable).

### Adaptive Typing Status
- Typing indicators now reflect agent activity (thinking, typing, tool calls, memory work, requesting approval) and coalesce updates to reduce API calls.
- ChannelAdapterConfig.typingStatus accepts boolean or (chunk, ctx) => string | false; defaultTypingStatus helper and TypingStatusContext exported.
- Typing-state emission centralized via a typingGate/withTypingStatus mechanism and resets on run boundaries.

### Observational Memory (OM) Integration
- OM buffering/activation chunks render into active streaming sessions as task rows; repeated activations coalesced into aggregated rows; sessions may show an initial "Updating memory" title.
- OM buffering completion emits terminal updates on session close to avoid spurious error states.

### Provider & Manifest Updates
- Slack provider accepts top-level adapter overrides (cors, gateway, formatError, streaming, typingStatus, toolDisplay) which override the deprecated adapterConfig fallback.
- Slack manifest now includes assistant:write bot scope and an assistant_view feature; added manifest tests.
- Slack DM/thread approval routing fixed so approvals resume on the correct Mastra thread.

### Bug Fixes & Improvements
- Fixes: parallel same-tool approval keying, tool error rendering across display modes, Slack DM approval routing, clearing stuck typing at run end, signal-aware message boundaries (data-user-message), stronger stay-silent system guidance, logger propagation into AgentChannels.
- Improved handling of tripwires, abort/retry semantics, file-chunk posting (flush pending text before files), and robust fallback when streaming post fails.

## Internal Refactor

- consumeAgentStream split into a short dispatcher plus two drivers:
  - chat-driver-streaming.ts — streaming driver creating StreamingPlan sessions (timeline/grouped/hidden), buffering/coalescing pieces, OM/task handling, streaming-aware tool rendering, and plan lifecycle handling.
  - chat-driver-static.ts — static driver that posts/edits discrete messages, buffers text, handles approval flows, files, errors, and terminal boundaries.
- Shared helpers/types moved to stream-helpers.ts (ToolTracker, extractErrorMessage, editOrPost, post helpers, renderBuiltInToolEvent), om.ts (OM chunk rendering), inline-media.ts (URL/MIME helpers), typing-status.ts (defaultTypingStatus), and types.ts (expanded public types).
- New ToolTracker class for correlating tool lifecycle across chunks and pendingApprovalCards stash for approval resumption.
- AgentChannels reduced in size; resolveToolDisplay, typing gating, and approval handling reorganized; logger propagation added.

## Breaking Changes & Migrations
- Removed ChannelAdapterConfig.cards and ChannelAdapterConfig.formatToolCall — migrate to toolDisplay (string or function) and ToolDisplayFn respectively.
- SlackProviderConfig restructured to expose explicit top-level adapter fields (streaming, toolDisplay, typingStatus, etc.); adapterConfig remains only as a deprecated fallback.
- Update callers using old cards/formatToolCall to the new toolDisplay API.

## Testing
- Extensive new tests for consumeAgentStream covering streaming/static drivers, toolDisplay modes, OM behavior, typing-status transitions, approvals, errors, files, tripwires, and message boundaries.
- Added tests: Slack manifest assistant:write/assistant_view, per-adapter streaming preservation, logger propagation, and stream-helpers (ToolTracker, extractErrorMessage).
- Core and Slack unit tests passed; Slack manually smoke-tested across modes.

## Documentation & Changesets
- Updated Channels docs and reference to document toolDisplay, streaming, typingStatus, the deprecations and migrations, and TypeScript examples.
- Changesets added for `@mastra/core` and `@mastra/channel-slack` describing UX changes and new defaults.

[Review Change Stack](https://app.coderabbit.ai/change-stack/mastra-ai/mastra/pull/16937?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


CalebBarnes:
@coderabbitai review

packages/playground/src/pages/metrics/index.tsx
```diff
@@ -52,7 +52,7 @@ const ANALYTICS_OBSERVABILITY_TYPES = new Set([
   'ObservabilityStorageClickhouseVNext',
   'ObservabilityStorageDuckDB',
   'ObservabilityInMemory',
-  "ObservabilitySpanner",
+  'ObservabilitySpanner',
```
CalebBarnes, line 55: unrelated lint fix

docs/src/content/en/reference/sidebars.js
```diff
@@ -464,13 +464,13 @@ const sidebars = {
         { type: 'doc', id: 'storage/convex', label: 'Convex Storage' },
         { type: 'doc', id: 'storage/duckdb', label: 'DuckDB Storage' },
         { type: 'doc', id: 'storage/dynamodb', label: 'DynamoDB Storage' },
+        { type: 'doc', id: 'storage/spanner', label: 'Google Cloud Spanner Storage' },
```
CalebBarnes, line 467: unrelated lint fix

TylerBarnes:
Thanks for the thorough channel UX work here — the driver split, `ToolTracker` keyed by `toolCallId`, Slack-specific defaults, and the new tests all look directionally good. I verified the core channels tests, Slack package checks, docs sidebar/remark checks, and playground build locally.

One issue I think should be fixed before merge:

`TypingStatusContext.toolCalls` is documented as a live read-only view of tools called so far, useful for branching on concurrent tools or the most recently called tool (`packages/core/src/channels/typing-status.ts:18-23`). But `withTypingStatus()` currently passes a fresh empty map on every callback invocation (`packages/core/src/channels/agent-channels.ts:1305-1311`).

That means custom `typingStatus` callbacks relying on `ctx.toolCalls` will never see in-flight tools, despite the public API promising that behavior. The built-in resolver does not use it, so default behavior still works.

Could we either maintain a run-scoped map in `withTypingStatus()` by observing tool-call/result/error/approval chunks, or remove `toolCalls` from the public context until it’s implemented?

abhiaiyer91 (approved):

CalebBarnes:
Good catch — and the right takeaway is that `toolCalls` shouldn't be on `TypingStatusContext` at all.

I went to wire it up and realized:
- The built-in `defaultTypingStatus` doesn't read it (only `chunk` + `ctx.channelTools`)
- No tests rely on it
- No identified use case actually needs a map vs. the current chunk's `payload.toolName`
- And anyone who genuinely wants in-flight tracking can do it in ~3 lines in their own closure (set on `tool-call`, delete on `tool-result`/`tool-error`) with full control over eviction policy

So I dropped the field from `TypingStatusContext` in 8b4913d9f8. Smaller honest surface, and it's non-breaking to add back later (optional field) if a concrete use case surfaces.

TylerBarnes (approved):
Looks good from my review. The TypingStatusContext surface issue was resolved by removing the unused toolCalls field.

PR #16651, opened 2026-05-15
Add FGA route policy coverage controls

graysonhicks:
## Description

This PR makes Mastra's FGA path safer and easier to extend for hosted server routes. ELI5: if auth says "this route is protected," FGA should either know what resource/permission to check or clearly fail/warn instead of silently letting the request through.

Before this change, configuring `server.fga` did not guarantee protected resource routes were FGA-protected because `checkRouteFGA` only ran when each route had explicit `route.fga` metadata. This adds provider-level policy coverage controls (`requireForProtectedRoutes`, `auditProtectedRoutes`, `resolveRouteFGA`, and `validatePermissions`), built-in FGA metadata resolution for Mastra resource routes, and startup validation for emitted permissions.

Example: a protected `POST /workflows/:workflowId/start` route can now derive `{ resource: { type: 'workflow', id: workflowId }, permission: 'workflows:execute' }` without each adapter route manually mutating metadata. Built-in coverage also applies to agents, workflows, tools, MCP tools, memory threads, responses, and conversations, while custom providers can still override route mapping globally with `resolveRouteFGA`.

The built-in resolver is now the source of truth for those resource routes, so redundant per-route `fga` blocks were removed from the server handlers while keeping special tool resource IDs for agent-scoped and MCP tools.

Follow-up review fixes also align direct execution FGA checks with the same resource scope used at runtime, guard agent resume paths (`resumeStream` / `resumeGenerate`), and preserve stored resource metadata on partial updates when `metadata` is omitted.

This also hardens the WorkOS provider boundary: when WorkOS reports an FGA resource is missing (`entity_not_found`/404), `check()` now returns denial and `require()` throws `FGADeniedError` instead of leaking a provider-specific error that can become a 500. The docs and focused tests were updated to cover the new route policy behavior and WorkOS denial semantics.

## Related Issue(s)

N/A

## Type of Change

- [x] Bug fix (non-breaking change that fixes an issue)
- [x] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [x] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [x] Test update

## Checklist

- [x] I have made corresponding changes to the documentation (if applicable)
- [x] I have added tests that prove my fix is effective or that my feature works

[...]

PR #16935, opened 2026-05-21
feat: v1 ToolProvider extensions — Studio frontend (PR 2 of 2)

YujohnNattrass:
## Summary

PR 2 of 2 split from #16837. Stacks on top of #16934 (backend). Adds the Studio Connections UI, surface-scoped model-policy provider, and CMS/agent-builder layout wiring.

## Scope

`packages/playground` only — 76 files.

- **tool-providers domain**
  - `ConnectionPicker` (716 LOC) + 38 tests: authorize OAuth, list / pin / unpin / reauth / disconnect, delete failed, dynamic pre-auth fields, owner badges, required label on create, auto-stamp `caller-supplied` sentinel.
  - `ToolProvidersSection`: mounts a picker per toolkit; required `scope` prop (no visibility radio — surface-locked).
  - `HealthPill` + `use-agent-health` hook + tests.
  - Hooks: `use-all-provider-tools`, `use-authorize`, `use-connection-*`, `use-disconnect-connection`, `use-existing-connections`, `use-infinite-connections` (server-side scope filter), `use-tools`.
  - Form mappers with round-trip tests; Zod schemas.
  - Deletes the legacy modal UX (`toolkit-list`, `tool-list`, `integration-tools-section`, `tool-provider-dialog`, `selected-tool-list`, `use-provider-tools`).

- **agent-builder integration**
  - Hybrid Tools/Connections tabs: integration tools render as `type:'integration'` rows in the Tools tab; 'Set up connection' deep-links to the Connections tab when no pin exists.
  - `use-tool-providers-bridge`: `addIntegrationTool` / `removeIntegrationTool` / `needsConnectionSetup`.
  - `use-set-agent-tools-tool` routes `type:'integration'` entries through `routeToolInputToFormKeys` into `toolProvidersFragment` and shallow-merges into `form.toolProviders`.
  - `use-agent-builder-tool` gates the `SET_AGENT_TOOLS` client tool on `useAllProviderTools().isLoading` so the LLM does not pick from an empty enum.
  - Builder pins hard-locked to `scope='per-author'`.

- **CMS / model policy**
  - `ModelPolicyProvider` context keyed by surface (`'builder' | 'editor'`).
  - `LLMProviders` / `LLMModels` read policy via `useModelPolicy()` — the builder allowlist no longer leaks into the CMS editor, chat composer, or other host surfaces.
  - CMS agent create/edit layouts wrapped in `<ModelPolicyProvider surface="editor">`.
[...]

PR #16944, opened 2026-05-22
fix(server): restore agent-builder permissions in generated catalog

YujohnNattrass:
## Summary

The `agent-builder:read`, `agent-builder:write`, `agent-builder:execute`, and `agent-builder:*` permissions silently dropped out of `packages/core/src/auth/ee/interfaces/permissions.generated.ts`, which made typed `RoleMapping` configs that reference them fail at compile time. Runtime FGA still works (it derives the permission from path/method), so this is a **type-only regression**.

## Root cause

#16651 (FGA route policy coverage) regenerated the permission catalog from `SERVER_ROUTES`. `AGENT_BUILDER_ROUTES` is deliberately *not* spread into `SERVER_ROUTES` so that importing the routes barrel does not eagerly load `@mastra/agent-builder` (Cloudflare Workers compat from #16499 — see `agent-builder-loading.test.ts`).

The generator therefore never saw agent-builder routes.

## Fix

`packages/server/scripts/permission-generator.ts` now walks `AGENT_BUILDER_ROUTES` in addition to `SERVER_ROUTES` when deriving the permission catalog. The runtime `SERVER_ROUTES` array is unchanged, so the lazy-load guarantee is preserved.

## Test plan

- `pnpm --filter @mastra/server generate:permissions` — regenerates with `agent-builder:*` restored
- `pnpm --filter @mastra/server test` — all 1585 tests pass, including `agent-builder-loading.test.ts` which asserts `SERVER_ROUTES` still does not contain `/agent-builder` paths
- Manual: `RoleMapping` configs in example apps that reference `'agent-builder:*'` now typecheck

## ELI5 (Explain Like I'm 5)

The permission system lost track of "agent-builder" permissions when generating type definitions—making TypeScript angry even though the runtime still worked. The fix was to update the permission generator to also look at agent-builder routes (in addition to the main routes) when creating the type definitions, without changing how those routes are lazily loaded.

---

## Summary of Changes

This PR restores `agent-builder` permissions (`agent-builder:read`, `agent-builder:write`, `agent-builder:execute`, and `agent-builder:*`) to the generated permission catalog.

### Problem

Commit `#16651` regenerated the permission catalog from `SERVER_ROUTES`, but `AGENT_BUILDER_ROUTES` is deliberately omitted from `SERVER_ROUTES` to preserve lazy-loading behavior (avoiding eager imports of `@mastra/agent-builder` for Cloudflare Workers compatibility). This caused the generator to silently drop agent-builder permissions from `PERMISSION_PATTERNS`, breaking TypeScript type-checking for `TypedRoleMapping` configs that referenced those permissions. Runtime FGA continued to derive them correctly from path/method, so this was a type-only regression.

### Solution

Updated `packages/server/scripts/permission-generator.ts` to walk both `SERVER_ROUTES` and `AGENT_BUILDER_ROUTES` when deriving the permission catalog. This restores agent-builder permissions to the generated types without changing the lazy-loading guarantee (SERVER_ROUTES itself remains unchanged).

### Files Modified

- **packages/core/src/auth/ee/interfaces/permissions.generated.ts** — Restored agent-builder permissions (plus auth and infrastructure) to `RESOURCES`, `PERMISSION_PATTERNS`, `PERMISSIONS`, and `MastraFGAPermissions`
- **packages/server/scripts/permission-generator.ts** — Now imports and includes `AGENT_BUILDER_ROUTES` in permission derivation
- **.changeset/restore-agent-builder-permissions.md** — Changeset documentation

### Testing & Verification

- All 1585 server tests pass
- `agent-builder-loading.test.ts` verifies `SERVER_ROUTES` still excludes `/agent-builder` paths
- Developers regenerate permissions via `pnpm --filter `@mastra/server` generate:permissions`
- `TypedRoleMapping` configs referencing `agent-builder:*` now typecheck correctly

[Review Change Stack](https://app.coderabbit.ai/change-stack/mastra-ai/mastra/pull/16944?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/.changeset/restore-agent-builder-permissions.md
+++ b/.changeset/restore-agent-builder-permissions.md
@@ -1,0 +1,6 @@
+---
+'@mastra/server': patch
+'@mastra/core': patch
+---
+
+Restore `agent-builder:*` permissions in the generated permission catalog. The FGA route policy coverage refactor (#16651) regenerated `permissions.generated.ts` from `SERVER_ROUTES`, but `AGENT_BUILDER_ROUTES` is intentionally excluded from that array (deferred-import for Cloudflare Workers compatibility, see #16499). As a result `agent-builder:read`, `agent-builder:write`, `agent-builder:execute`, and `agent-builder:*` were silently dropped from `PERMISSION_PATTERNS`, breaking typed `RoleMapping` configs that reference them. The permission generator now also walks `AGENT_BUILDER_ROUTES` so the catalog matches runtime FGA behavior, without changing the lazy-loading guarantee.
--- a/packages/core/src/auth/ee/interfaces/permissions.generated.ts
+++ b/packages/core/src/auth/ee/interfaces/permissions.generated.ts
@@ -13,12 +13,15 @@
  */
 export const RESOURCES = [
   'a2a',
+  'agent-builder',
   'agents',
+  'auth',
   'background-tasks',
   'channels',
   'datasets',
   'embedders',
   'experiments',
+  'infrastructure',
   'logs',
   'mcp',
   'memory',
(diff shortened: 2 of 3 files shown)


YujohnNattrass:
Moving this fix into PR #16935 instead so it ships with the marquess merge.

Issue #9969, opened 2025-11-11
Is there a way to process the System Prompt using Input processor at the Agent Level?


daneatmastra:
This issue was created from Discord post 1437556709823479999:

[![Open in Discord](https://img.shields.io/badge/Open_in_Discord-5865F2?style=for-the-badge&logo=discord&logoColor=white)](https://discord.com/channels/1309558646228779139/1437556709823479999)

Currently, I only see role: user that passed in the argument.
Well, in the documentation it says Messages in general, so I was expecting to receive all the messages regardless of roles

Or perhaps, I missed something, Still continue reading

---

![Screenshot 1](https://cdn.discordapp.com/attachments/1437556709823479999/1437556710108823694/Screenshot_2025-11-11_at_5.33.37_AM.png?ex=69145544&is=691303c4&hm=86178b7df2c3f45bb702f11355e444b4de5829f928bc323928e981af6d28bfa0&)

spo0nman:
Looking at the mastra code, InputProcessors only receive user messages (messageList.clear.input.db()), not system messages. The semantic recall is added as a system message, so InputProcessors can't access it.

Flow:
Line 227: Semantic recall added as system message
Line 237: InputProcessors run, but only receive user messages
Line 483 (llm-execution-step.ts): Final prompt assembled from messageList

From runner.ts line 289:
const userMessages = messageList.clear.input.db(); 


Semantic recall is added as a system message (line 227 in prepare-memory-step.ts), so InputProcessors never see it.

spo0nman:
https://discord.com/channels/1309558646228779139/1432282762265755819

roaminro:
Thanks @spo0nman ! Unfortunately it's not possible at the moment, but we;re working on a new processors implementation that will give you more flexibility. stay tuned ;)

Issue #16947, opened 2026-05-22
ChatChannelProcessor overwrites Observational Memory system messages


panliji:
### Problem

When Observational Memory is enabled together with Channels, active observations are stored and OM activation runs, but the final model input does not contain the OM context.

### Repro / Root Cause

In `ObservationalMemoryProcessor`, OM adds tagged system messages to the current `MessageList` via `messageList.addSystem(..., "observational-memory")`.

Later, the built-in `ChatChannelProcessor` (`id = "chat-channel-context"`) returns a full `systemMessages` array:

```js
return { systemMessages: [...args.systemMessages, systemMessage] };
```

The `ProcessorRunner` then calls `replaceAllSystemMessages(...)` for results that contain `systemMessages`, which clears previously added tagged system messages. As a result, OM is present in storage and activation, but disappears from the prompt before the model call.

### Impact

This reproduces in our Feishu Channel integration after enabling official Mastra Channels. It appears to affect any channel flow where `chat-channel-context` runs after OM.

### Expected Behavior

Channel context should preserve previously added system messages, including OM observations, rather than replacing them with a stale snapshot.

### Suggested Fix

`ChatChannelProcessor` should merge against the current `messageList.getAllSystemMessages()` (or otherwise preserve existing tagged system messages), instead of returning a replacement array derived from `args.systemMessages`.

### Notes

We confirmed the issue in a TEST Feishu group thread where OM was active but absent from the final model input.

daneatmastra:
Thank you for reporting this issue! We have labeled it for trio-tb and we will look into it as soon as possible.

If you're experiencing an error, please provide a [minimal reproducible example](https://github.com/mastra-ai/mastra/blob/main/CONTRIBUTING.md#minimal-reproduction) whenever possible to help us resolve it quickly.

intojhanurag:
Hey @panliji I can think about stale or duplicate OM message , but I am not able to understand how are you not seeing OM context . 
Can you give me a reproduction ?

panliji:
Here is a minimal reproduction from our side:

### Setup
- Mastra Agent with `memory.observationalMemory.enabled = true`
- `memory.workingMemory.enabled = true`
- Official Channels enabled on the agent
- A Feishu channel thread context so `chat-channel-context` runs
- A thread that has already triggered OM activation after idle

### Steps
1. Send enough messages in the same thread to let Observational Memory activate.
2. Wait until OM has stored active observations.
3. Send a new message in the same thread.
[...]

PR #16950, opened 2026-05-22
fix(core): preserve tagged system messages when processors return systemMessages

intojhanurag:
## Description

Returned `systemMessages` from processors used to call `replaceAllSystemMessages()`, which cleared both untagged system messages and tagged system messages owned by other processors. That meant a later processor returning a system message array could strip tags such as `observational-memory`, breaking later refresh/clear behavior and causing stale or duplicated memory context across steps.

This PR updates `ProcessorRunner` so returned `systemMessages` replace only untagged system messages. Tagged system messages are preserved for the processors that own them, and returned entries that correspond to previously tagged messages are not duplicated as untagged copies.

- Preserves tagged system messages when regular input processors return `systemMessages`
- Preserves tagged system messages when `processInputStep` processors return `systemMessages`
- Adds regression coverage for OM-like tagged system messages with channel context and for runner-level returned `systemMessages` behavior

## Context

#17168 already fixed the channel-specific symptom by switching `ChatChannelProcessor` from returning `{ systemMessages: [...] }` to calling `messageList.addSystem(msg, this.id)`. Its PR description explicitly defers the general fix to this PR:

> "the bug is general to any processor that returns `systemMessages`… The runner-side fix (general, broader) is being pursued separately in #16950."

The bug class is still live for any other processor that returns a `systemMessages` array. The clearest example today is `BrowserContextProcessor` (`packages/core/src/browser/processor.ts:85`):

```ts
const systemMessages = [...args.systemMessages, { role: 'system' as const, content: lines.join(' ') }];
return { messages: args.messages, systemMessages };
```

An agent using Observational Memory together with a browser tool would hit the same tag-wipe behavior today. This PR closes that gap once at the runner level so every current and future processor returning `systemMessages` is safe.

## Type of Change

- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [ ] Test update

## Checklist

## ELI5
When one processor updates the system messages (notes/instructions), it was accidentally deleting all system messages from other processors too—including special labeled ones that shouldn't be touched. This fix makes sure that only regular, unlabeled messages get replaced, while labeled messages (like "observational-memory" tags) stay safe and protected.

---

## Overview
This PR fixes a bug in `ProcessorRunner` where processors returning `systemMessages` were inadvertently clearing **tagged** system messages owned by other processors. Previously, `replaceAllSystemMessages()` would strip all system messages regardless of ownership tags, breaking functionality that relies on those tags (like observational memory refresh/clear behavior). The fix preserves tagged system messages while allowing processors to replace only untagged system messages.

### Core Logic Updates (`packages/core/src/processors/runner.ts`)
- Added internal helper functions to manage system message merging:
  - `getTaggedSystemMessages`: Extracts system messages that have tags by comparing all system messages against untagged ones
  - `applyReturnedSystemMessages`: Intelligently applies processor-returned system messages, preserving tagged ones while replacing untagged ones
- Updated `runInputProcessors`: Processors returning `systemMessages` now call `applyReturnedSystemMessages()` instead of `replaceAllSystemMessages()`, preserving tags
- Updated `runProcessInputStep`: Applies the same preservation logic when step processors return `systemMessages`

### Type Documentation (`packages/core/src/processors/index.ts`)
- Clarified JSDoc comments for `processInput` and `processInputStep` return types to explicitly document that returned `systemMessages` replace only untagged messages while preserving tagged system messages owned by other processors

### Test Coverage
- **`process-input-step.test.ts`** (+86 lines): Added two regression tests in the `systemMessages modification` suite:
  - Verifies that appending untagged system messages preserves existing tagged messages (e.g., `observational-memory` tags)
  - Validates ordering semantics when untagged messages are prepended: tagged messages remain in original positions while untagged messages appear in expected sequence
- **`runner.test.ts`** (+11/-6 lines): Updated existing Issue `#9969` regression test to assert that after a processor returns modified `systemMessages`, only untagged messages are reflected in the prompt, while tagged system messages remain unchanged and retain original content

### Release Documentation
- Added `.changeset` entry documenting the patch-level fix for `@mastra/core`

## Impact
This change prevents tagged system messages (used by observational memory and similar features) from being inadvertently dropped when subsequent processors return updated system messages, ensuring consistent and predictable behavior across processor chains.

[Review Change Stack](https://app.coderabbit.ai/change-stack/mastra-ai/mastra/pull/16950?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


intojhanurag:
Hey @panliji , can you review this PR and can confirm my claim is right or not ?

packages/core/src/channels/processor.ts
```diff
@@ -52,6 +52,8 @@ export class ChatChannelProcessor {
     }
 
     const systemMessage: CoreMessageV4 = { role: 'system', content: lines.join('\n') };
-    return { systemMessages: [...args.systemMessages, systemMessage] };
+    args.messageList.clearSystemMessages(this.id);
+    args.messageList.addSystem(systemMessage, this.id);
+    return { messageList: args.messageList };
```
TylerBarnes: instead of this we should fix what happens when systemMessages array is returned

TylerBarnes: we need to restore tagged system messages, and returned system messages here should probably not be considered the same as tagged ones anyway

TylerBarnes (changes requested):

panliji:
@intojhanurag @Bickio Follow-up from our latest debugging on our TEST environment:

We re-tested the Feishu channel case and can now confirm that our original diagnosis was correct.

In our trace:
- Observational memory exists in storage for the thread.
- `ObservationalMemoryProcessor` adds the observation text into the processor input/context.
- `chat-channel-context` runs later.
- After that processor step, the final `model_step` input no longer contains the observational-memory block.

So this does not appear to be caused by our app-side storage or memory-key logic. The observed behavior matches the issue described in #16947: a later processor returning `systemMessages` causes previously tagged system messages, including `observational-memory`, to be dropped.

The approach in this PR looks aligned with our latest findings. Preserving tagged system messages at the `ProcessorRunner` level seems better than fixing only `ChatChannelProcessor`, because it protects OM and other tagged processor-owned system messages from the same class of bug.

TylerBarnes (changes requested):
Thanks, this fixes the main legacy runner issue and the approach is directionally right. Preserving tagged system messages at the processor-runner level is the right shape for the observational-memory/channel-context bug.

I found one remaining workflow-adapter edge case worth tightening before merge.

In `createStepFromProcessor`, after `applyReturnedSystemMessages()` updates the shared `MessageList`, the step still returns the processor’s raw `typedResult.systemMessages` / `validatedResult.systemMessages`. In a chained processor workflow, the next processor step receives `args.systemMessages` from that returned object, so it can miss tagged messages that were preserved on `messageList`.

Examples:
- `packages/core/src/workflows/workflow.ts:1141-1146`
- `packages/core/src/workflows/workflow.ts:1205-1215`
- Same pattern in `packages/core/src/workflows/evented/workflow.ts:1163-1168` and `1224-1234`

I think the safer fix is to return the canonical `messageList.getAllSystemMessages()` value after applying the helper:

```ts
applyReturnedSystemMessages(checkedMessageList, typedResult.systemMessages);

return {
  ...passThrough,
  messages: typedResult.messages,
  systemMessages: checkedMessageList.getAllSystemMessages(),
};
```

For the `validatedResult` branches, make sure `systemMessages` is overridden after spreading `validatedResult`, e.g.:

```ts
return {
  ...passThrough,
  messages,
  ...validatedResult,
  systemMessages: checkedMessageList.getAllSystemMessages(),
  ...(currentMessageId ? { messageId: validatedResult.messageId ?? currentMessageId } : {}),
};
```

Specific design suggestion: instead of exporting `applyReturnedSystemMessages()` as a processor helper, I’d update `MessageList.replaceAllSystemMessages()` itself to preserve existing tagged system-message ownership.

Proposed semantics for `replaceAllSystemMessages(messages)`:

1. Preserve existing tagged system-message buckets.
2. Replace the untagged system-message bucket with the returned messages.
3. If the returned array includes content that already exists in a tagged bucket, do not re-add that content as untagged.

That makes the safer behavior the default and keeps the processor code simple:

```ts
messageList.replaceAllSystemMessages(result.systemMessages);
```

I think that is the right default because tags are ownership boundaries between processors. A replacement array should be able to update the caller-owned/untagged system context, but it should not silently flatten tagged messages into untagged messages or wipe another processor’s tagged context. If a truly destructive “clear every tagged bucket and flatten everything” reset is needed later, that can be added as a separate explicit API, but it should not be the normal behavior used by processor-returned `systemMessages`.

Suggested regression coverage:
- Create a two-step processor workflow.
- Seed `messageList` with one untagged system message and one tagged memory/observational-memory system message.
- Step 1 returns a replacement untagged `systemMessages` array.
- Step 2 asserts its `args.systemMessages` still includes the tagged memory message.

Local validation I ran:
- `pnpm build:core`
- `pnpm --filter ./packages/core exec vitest run src/processors/runner.test.ts src/processors/process-input-step.test.ts src/workflows/processor-step.test.ts --bail 1 --reporter=dot`
- `pnpm --filter ./packages/core exec vitest run src/workflows/evented/evented-workflow.test.ts --bail 1 --reporter=dot`

TylerBarnes (changes requested):
After thinking about this more, I think this is still patching the symptom rather than fixing the contract.

The root issue is that processor `args.systemMessages` is populated from `messageList.getAllSystemMessages()`, so it mixes untagged system messages with tagged processor-owned messages. But returning `{ systemMessages }` is then treated as a replacement write. That means tagged messages are being round-tripped through a mutable untagged replacement API.

I don’t think tagged system messages should enter `args.systemMessages` in the first place. `systemMessages` should represent the mutable untagged bucket only. Tagged messages should remain separate `MessageList` state and only be assembled with untagged messages when building the final model input.

The current `replaceAllSystemMessages()` change prevents data loss, but it does so by reconstructing tag ownership after processors have already received and echoed tagged messages through the wrong channel. That feels like a symptom patch.

Suggested direction:
- pass `messageList.getSystemMessages()` as processor `systemMessages`
- keep tagged buckets out of the processor-returned `systemMessages` round trip
- keep final model calls using `messageList.getAllSystemMessages()`
- add regression coverage proving tagged messages are not passed into processor `args.systemMessages`, but are still present in the final model input

TylerBarnes (changes requested):
Thanks for updating the core runner/workflow paths to make `args.systemMessages` untagged-only. That’s the right contract, but I found a few remaining places that still either use the old flattened behavior or rely on the old assumption.

I don’t think we should merge until these are addressed:

1. `client-sdks/ai-sdk/src/middleware.ts` still passes tagged messages into processor args:

```ts
systemMessages: messageList.getAllSystemMessages(),
```

This should use:

```ts
systemMessages: messageList.getSystemMessages(),
```

Otherwise processors invoked through the AI SDK middleware still see the old tagged+untagged array.

2. `workflows/inngest/src/index.ts` appears to still have the same old contract in the processor workflow adapter. It passes `systemMessages` through to `processInput`, `processInputStep`, and `processOutputStep`, and returns `getAllSystemMessages()` from several `MessageList` return paths. This should match the updated core workflow adapters: processor args/phase return values should be untagged-only via `getSystemMessages()`, while tagged messages stay owned by `messageList`.

3. `TokenLimiterProcessor` now under-counts prompt tokens. It reads `args.systemMessages`, but the final LLM prompt still includes tagged system messages. Since token limiting budgets against the final prompt, it should probably count `messageList.getAllSystemMessages()` internally, while still preserving the public `args.systemMessages` contract as untagged-only.

4. `packages/core/src/channels/processor.test.ts` still manually builds processor args with `messageList.getAllSystemMessages()` and has a stale comment saying the runner flattens tagged messages. That should be updated to the new contract.

The core direction is good now, but these missed entry points/consumers mean tagged system messages can still leak into `args.systemMessages` or be ignored for final-prompt budgeting.

TylerBarnes (approved):
Rechecked the latest head. The untagged-only systemMessages contract now covers the core runner/workflow paths, the AI SDK middleware, and the Inngest adapter; TokenLimiter now budgets against the full final system-message set; channel tests and changesets were updated. Targeted core checks passed locally, and the remaining live failures I see are Vercel preview statuses. Looks good to me.

PR #16954, opened 2026-05-22
fix(mcp): close transport on connection close to prevent SSE leaks (fixes #16693)

### LINKED ISSUE #16693: SSE transport leak in InternalMastraMCPClient — implicit reconnect path orphans EventSource inst
This issue was created from Discord post 1505521220542791680:

[Open in Discord](https://discord.com/channels/1309558646228779139/1505521220542791680)

Hi! We're seeing a real session-leak bug in `@mastra/mcp@1.6.0` that caused our MCP server's session map to grow to ~30K entries over a few days of normal use. Traced
  it down to a specific path in `InternalMastraMCPClient`. Happy to send a PR — wanted to confirm the diagnosis first.

  ## The bug

  When the MCP server triggers a connection close (server restart, network blip, idle timeout — anything that fires `client.onclose`), `InternalMastraMCPClient` leaks
  the previous transport. The orphaned `EventSource` continues its built-in 3-second retry loop, opening fresh SSE handshakes against the server indefinitely.

  In `@mastra/mcp/dist/index.js`:

  - The `onclose` handler set inside `connect()` (lines 688–695) clears `this.isConnected = null` but does **not** call `this.transport.close()` and does **not** clear
  `this.transport`.
  - The next `connect()` runs `connectHttp()` (lines 628 / 650), which assigns `this.transport = streamableTransport` / `this.transport = sseTransport` without first
  closing the previous transport.
  - Because nothing calls `SSEClientTransport.close()` on the orphan, its underlying `EventSource.readyState` stays `CONNECTING`, and `scheduleReconnect_fn` keeps firing
   every `reconnectInterval` (3 s default per the W3C spec / `eventsource` polyfill). Each retry is a fresh HTTP GET against the SSE URL.
  - Compare with `forceReconnect()` (lines 791–806), which correctly does `await this.transport.close()` before clearing — so the **explicit** reconnect path is
  leak-free. Only the **implicit** `onclose → connect()` path leaks.

STiFLeR7:
Closes #16693

## Summary
This PR fixes an SSE transport resource leak in the MCP client by properly closing the transport when the server initiates connection closure.

### Changes
packages/mcp/src/client/client.ts
Modified the this.client.onclose disconnect handler in the InternalMastraMCPClient.connect() method to:
1. Explicitly call this.transport.close() when the transport exists
2. Catch and log any errors that occur during transport closure
3. Clear the this.transport reference
4. Preserve the original behavior of resetting this.isConnected to null
5. Call any pre-existing onclose handler if one was defined

## ELI5

The PR fixes a resource leak where the MCP client wasn't properly cleaning up connections when the server closed them. Think of it like leaving a door open after someone tells you to shut it — the client would keep trying to reconnect through the old door instead of closing it, which wasted server resources and caused the server to accumulate thousands of unused sessions. The fix makes sure the client properly closes and cleans up the old connection when the server tells it to close.

## Changes

**packages/mcp/src/client/client.ts**

Modified the `this.client.onclose` disconnect handler in `InternalMastraMCPClient.connect()` to:
- Explicitly call `this.transport.close()` when the transport exists
- Catch and log any errors thrown during transport closure
- Clear the `this.transport` reference after closing
- Preserve the existing behavior of resetting `this.isConnected` to `null`
- Still invoke any pre-existing `onclose` handler if defined

This ensures that orphaned EventSource instances don't remain in a CONNECTING state with an active retry loop, preventing the repeated SSE connections that were leaking server sessions.

**Changelog**

Added a Changesets entry for `@mastra/mcp` marking a patch release documenting the fix for the SSE transport resource leak (fixes `#16693`).

[Review Change Stack](https://app.coderabbit.ai/change-stack/mastra-ai/mastra/pull/16954?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/.changeset/fix-mcp-sse-leak.md
+++ b/.changeset/fix-mcp-sse-leak.md
@@ -1,0 +1,5 @@
+---
+"@mastra/mcp": patch
+---
+
+fix(mcp): close transport on connection close to prevent SSE leaks (fixes #16693)
--- a/packages/mcp/src/client/client.ts
+++ b/packages/mcp/src/client/client.ts
@@ -500,6 +500,17 @@ export class InternalMastraMCPClient extends MastraBase {
         this.client.onclose = () => {
           this.log('debug', `MCP server connection closed`);
           this.isConnected = null;
+
+          // Cleanup transport on server-initiated close to prevent SSE leaks (fixes #16693)
+          if (this.transport) {
+            this.transport.close().catch(e => {
+              this.log('debug', 'Error closing transport on connection close (ignored)', {
+                error: e instanceof Error ? e.message : String(e),
+              });
+            });
+            this.transport = undefined;
+          }
+
           if (typeof originalOnClose === 'function') {
             originalOnClose();
           }


abhiaiyer91:
Closing in favor of https://github.com/mastra-ai/mastra/pull/17326

PR #17016, opened 2026-05-25
feat(agent): allow workflows to define explicit toolId for agent

### LINKED ISSUE #17001: [FEATURE] Allow workflows attached to agents to define explicit tool IDs
## Problem Statement

When a workflow is attached to an agent, Mastra currently exposes that workflow to the model as a generated tool name using `workflow-${registrationKey}`.

That makes the model-facing tool name depend on the JavaScript object key used to register the workflow, and it always adds an automatic prefix. Tool names are part of the model contract, so users need a way to make them stable, descriptive, and compatible with their chosen naming convention, such as snake_case or kebab-case.

Example today:

```ts
const ingestBrandCatalog = createWorkflow({
  id: 'ingest-brand-catalog',
  // ...
});

const agent = new Agent({
  workflows: { ingestBrandCatalog },
});
```

The model sees a generated tool name like:

```txt
workflow-ingestBrandCatalog
```

## Proposed Solution

Allow workflows to define an explicit model-facing tool ID for the agent-tool exposure path.

For example:

```ts
const ingestBrandCatalog = createWorkflow({
  id: 'ingest-brand-catalog',
  options: {
    toolId: 'ingest_brand_catalog',
  },
  // ...
});
```

When this workflow is attached to an agent, the model should see exactly:

```txt
ingest_brand_catalog
```

If `options.toolId` is omitted, Mastra should keep the existing behavior for backwards compatibility:

```txt
workflow-${registrationKey}
```

This keeps existing projects working while giving users explicit control when they need it.

## Component

Agents, Tools, Workflows, Studio

## Alternatives Considered

The current workaround is to choose a registration object key that approximates the desired tool name, but that still leaves the automatic `workflow-` prefix in place and couples model-facing naming to JavaScript registration structure.

Another option would be to change the default generated workflow tool name to use the workflow `id`, but that would be a breaking change for existing agents and saved traces. An explicit opt-in `toolId` is safer.

## Example Use Case

As a developer building a multi-agent system, I want workflow tools exposed to models with exact, stable names like `ingest_brand_catalog`, `read_brand_profile`, or `classify_source_url`, so that tool naming is consistent across agents, prompts, Studio, traces, and provider-specific tool constraints.

This is especially useful in larger systems where tools and workflow-backed tools need to follow the same naming convention and where registration keys are internal implementation details rather than model-facing API.

## Additional Context

It would also be useful for Studio and server list responses to distinguish between:

- the workflow's canonical `id`
- the registration key used in a Mastra config object
- the model-facing tool ID used when the workflow is attached to an agent

Those are related but distinct concepts, and making them explicit would reduce confusion when debugging agent tool calls.

## Verification

nehaprasad-dev:
#### description

- Allow workflows attached to agents to define an explicit model-facing tool ID via options.toolId. When set, the agent uses this value as the tool name instead of the default workflow-${registrationKey}. If omitted, existing behavior is preserved.

#### Related issue(s)
Fixes #17001

#### Type of change

#### Checklist

## ELI5

Workflows attached to agents now get to choose what name they're called by the AI model. Instead of always getting an auto-generated name like "workflow-myWorkflow123", they can specify their own name (like "send-email" or "check-inventory") to be clearer and more stable across different parts of the system.

### Overview

This PR adds support for workflows to explicitly define their model-facing tool ID via a new optional `options.toolId` parameter. When provided, the agent uses this exact name as the tool identifier instead of the auto-generated `workflow-${registrationKey}` format. If omitted, the existing behavior is preserved for backwards compatibility.

### Core Changes

**WorkflowOptions Type (`packages/core/src/workflows/types.ts`)**
- Added optional `toolId?: string` field to the `WorkflowOptions` export
- Documented as the explicit model-facing tool ID used when the workflow is exposed as a tool on an agent

**Workflow Constructor (`packages/core/src/workflows/workflow.ts`)**
- Updated the private `#options` object to include `toolId: options.toolId`
- Preserves the option alongside existing configuration like `validateInputs`, `shouldPersistSnapshot`, `tracingPolicy`, and lifecycle hooks

**Agent Tool Resolution (`packages/core/src/agent/agent.ts`)**
- Modified `listWorkflowTools()` to derive `toolName` from `workflow.options?.toolId` 
- Falls back to the existing `workflow-${workflowName}` format when `toolId` is not specified
- Applies the resolved `toolName` consistently across:
  - Tool creation via `createTool({ id })`
  - Tool options naming
  - Workflow tool registration keys

### Backwards Compatibility

The implementation maintains full backwards compatibility by defaulting to the existing `workflow-${workflowName}` naming scheme when `toolId` is not provided. Existing workflows will continue to function without any modifications.

### Benefits

- **Stable naming**: Tool names are no longer coupled to internal JavaScript registration keys
- **Descriptive identifiers**: Workflows can use meaningful names that clearly describe their purpose
- **Multi-agent systems**: Consistent tool naming across different agents in complex deployments
- **Convention compliance**: Supports naming conventions required by specific AI models or integration standards

[Review Change Stack](https://app.coderabbit.ai/change-stack/mastra-ai/mastra/pull/17016?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/packages/core/src/agent/agent.ts
+++ b/packages/core/src/agent/agent.ts
@@ -4721,6 +4721,8 @@ export class Agent<
     const workflows = await this.listWorkflows({ requestContext });
     if (Object.keys(workflows).length > 0) {
       for (const [workflowName, workflow] of Object.entries(workflows)) {
+        const toolName = workflow.options?.toolId || `workflow-${workflowName}`;
+
         // Build input/output schemas as JSONSchema7 to avoid Zod composition issues
         // when workflow schemas are StandardSchemaWithJSON wrappers (e.g. from storage)
         const inputDataJsonSchema: JSONSchema7 = workflow.inputSchema
@@ -4769,7 +4771,7 @@ export class Agent<
         };
 
         const toolObj = createTool({
-          id: `workflow-${workflowName}`,
+          id: toolName,
           description: workflow.description || `Workflow: ${workflowName}`,
           inputSchema: extendedInputSchema,
           outputSchema,
@@ -4933,7 +4935,7 @@ export class Agent<
         });
 
         const options: ToolOptions = {
-          name: `workflow-${workflowName}`,
+          name: toolName,
           runId,
           threadId,
           resourceId,
@@ -4948,7 +4950,7 @@ export class Agent<
           tracingPolicy: this.#options?.tracingPolicy,
         };
 
-        convertedWorkflowTools[`workflow-${workflowName}`] = makeCoreTool(
+        convertedWorkflowTools[toolName] = makeCoreTool(
           toolObj,
           options,
           undefined,
--- a/packages/core/src/workflows/types.ts
+++ b/packages/core/src/workflows/types.ts
@@ -479,6 +479,12 @@ export interface WorkflowOptions {
    * Errors thrown in this callback are caught and logged, not propagated.
    */
   onError?: (errorInfo: WorkflowErrorCallbackInfo) => Promise<void> | void;
+
+  /**
+   * Explicit model-facing tool ID used when this workflow is exposed as a tool on an agent.
+   * If set, the agent uses this value as the tool name instead of the default `workflow-${registrationKey}`.
+   */
+  toolId?: string;
 }
 
 export type WorkflowInfo = {
--- a/packages/core/src/workflows/workflow.ts
+++ b/packages/core/src/workflows/workflow.ts
@@ -1709,6 +1709,7 @@ export class Workflow<
       onFinish: options.onFinish,
       onError: options.onError,
       sharePubsub: options.sharePubsub,
+      toolId: options.toolId,
     };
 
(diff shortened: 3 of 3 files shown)


intojhanurag:
Hey @nehaprasad-dev, thanks for the PR, but this needs more work before it can land. Closing for now.

The workflow- prefix is checked in three other places that this PR doesn't update:
- packages/core/src/loop/workflows/agentic-execution/tool-call-step.ts:462 — breaks suspend/resume of workflow tools with a custom toolId
- packages/core/src/tools/tool-builder/builder.ts:88 — schema injection for suspendedToolRunId is skipped
- packages/core/src/background-tasks/resolve-config.ts:92 — background-task config lookup gets the wrong key

Also missing: tests, changeset, docs, and a duplicate-toolId guard (CodeRabbit flagged this).

PR #17023, opened 2026-05-25
fix(voice-google-gemini-live): route tool Zod schemas through GoogleSchemaCompatLayer

### LINKED ISSUE #17018: [BUG] @mastra/voice-google-gemini-live cannot connect to any current Gemini Live native-audio model
## Describe the Bug

`@mastra/voice-google-gemini-live@0.11.4` cannot establish a session with any current Gemini Live native-audio model. Every `voice.connect()` call fails with a WebSocket close before `setupComplete` is received, surfaced to the caller as a 30 s `Session creation timeout`.

The default model `gemini-2.0-flash-exp` was [shut down on 2025-12-09](https://ai.google.dev/gemini-api/docs/changelog) — so the zero-config default path is unconnectable. Overriding `model` to any current native-audio model (`gemini-3.1-flash-live-preview`, `gemini-2.5-flash-native-audio-preview-12-2025`, `gemini-live-2.5-flash-preview-native-audio-09-2025`) doesn't help, because the setup-message payload itself is the root cause.

Root cause is in `sendInitialConfig()` at `voice/google-gemini-live-api/src/index.ts:1731`. Three distinct wire-shape bugs:

1. **Missing `generation_config.response_modalities`.** Native-audio models reject the setup with `1007 Cannot extract voices from a non-audio request.` when the field is absent. Verified at wire level — see the repro's `VALIDATION.md`.

2. **camelCase setup keys.** The library sends `systemInstruction`; Gemini's Live API wire format expects `system_instruction`. This is consistent with the existing snake_case shape already used by `UpdateMessage` in this very package (`voice/google-gemini-live-api/src/types.ts:311`), Google's own [plain-js native-audio demo](https://github.com/GoogleCloudPlatform/generative-ai/blob/main/gemini/multimodal-live-api/native-audio-websocket-demo-apps/plain-js-demo-app/frontend/geminilive.js), and the snake_case keys the library already emits in its `session.update` path (`voice/google-gemini-live-api/src/index.ts:595`).

3. **Tools wire shape — silent breakage.** The library emits `tools: [{ functionDeclarations: [...] }]` — camelCase, with **one container per tool**. The wire format expects `tools: [{ function_declarations: [...] }]` — snake_case, with **a single container** holding every declaration. Gemini accepts the malformed shape at setup but never emits tool_call frames back to the client, so the model "talks about" using a tool without ever calling it. This is independently observable on agents that include tools.

A fourth issue lives in `toolResponse` handling: when a tool returns an array or a primitive, the session closes with `1007 Unknown name "response": Proto field is not repeating` because the `response` proto field is a struct, not repeating. Tool returns need to be wrapped in `{ result: ... }` for non-struct payloads. This is in scope for the same fix.

The repro repository linked below isolates each bug at both the library level and the WebSocket wire level.

## Steps To Reproduce

```bash
git clone https://github.com/petrchiriboga/mastra-gemini-live-native-audio-repro
cd mastra-gemini-live-native-audio-repro
pnpm install
cp .env.example .env   # paste a Gemini API key from https://aistudio.google.com/apikey
pnpm repro:mastra      # library-level — every model fails with Session creation timeout
pnpm repro:wire        # wire-level — variant A fails, variants B+C succeed
```

`repro-wire.ts` sends three setup-message payloads over a raw WebSocket:

- **A**: current upstream shape (camelCase, no `generation_config`) — closes with 1007 / 1011 depending on model
- **B**: snake_case + `response_modalities: ["AUDIO"]`, **no** `voice_config` — succeeds
- **C**: B plus `voice_config.prebuilt_voice_config.voice_name: "Puck"` — succeeds

`VALIDATION.md` in the repro records the exact verdicts on `gemini-3.1-flash-live-preview` and `gemini-2.5-flash-native-audio-preview-12-2025`.

https://github.com/petrchiriboga/mastra-gemini-live-native-audio-repro

## Expected Behavior

`voice.connect()` on any current native-audio model should resolve and emit a `connected` session-state event. The setup message should:

- Use snake_case keys for `system_instruction`, `generation_config`, `response_modalities`, `speech_config`, `voice_config`, `prebuilt_voice_config`, `voice_name` — matching the existing `UpdateMessage` shape in the same package and Google's native-audio reference demo
- Emit `generation_config.response_modalities: ["AUDIO"]` by default — this is a voice library, so audio output is the only sensible session-level modality (per-turn override remains available via `GeminiLiveVoiceOptions.responseModalities`)
- Emit `voice_config.prebuilt_voice_config.voice_name` only when the constructor's `speaker` option is set, so Gemini's server-side default applies when the user doesn't configure one
- Emit tools as a single `{ function_declarations: [...] }` container in snake_case
- Wrap non-struct tool results in `{ result: ... }` before sending as `functionResponses[].response`, and include `functionResponses[].name`

The default model should also be bumped from the shut-down `gemini-2.0-flash-exp` to a current native-audio model (Google's quickstart currently uses `gemini-3.1-flash-live-preview`).

## Environment Information

```
System:
  OS: macOS 15.3.1
  CPU: arm64 Apple M4 Pro
Binaries:
  Node: 24.13.0
  npm: 11.6.2
  pnpm: 10.29.3
npmPackages:
  @mastra/core: 0.24.4
  @mastra/voice-google-gemini-live: 0.11.4
  ws: 8.18.0
  dotenv: 16.4.5
```

LLM Provider: Google Gemini (Live API, `generativelanguage.googleapis.com`, v1alpha)

Also tested with Node v22.13.0 — same behavior. The bug is wire-shape, not Node-version-dependent.

## Verification

### LINKED ISSUE #17020: [BUG] @mastra/voice-google-gemini-live hand-rolled Zod-to-JSON-Schema converter drops most Zod types
## Describe the Bug

`@mastra/voice-google-gemini-live@0.11.4` ships a hand-rolled `convertZodSchemaToJsonSchema` / `convertZodDefToJsonSchema` at `voice/google-gemini-live-api/src/index.ts:2071-2165`. The converter supports only a hardcoded short list of Zod types — `ZodString`, `ZodNumber`, `ZodBoolean`, `ZodArray`, `ZodObject`, `ZodOptional`, `ZodEnum` — and falls through to a `default` branch that returns `{ type: 'object', properties: {} }` for every other Zod type, stripping all type information from the tool parameter schemas before they reach the Gemini Live setup message.

Tools registered via `voice.addTools(...)` with any of the following Zod patterns end up exposing an empty parameter schema to Gemini, so the model receives no signal about what arguments to populate and tool args come back empty at call time:

- `z.union([...])` — common for accepting multiple shapes
- `z.discriminatedUnion(...)` — tagged-union pattern, very common
- `z.literal('...')` — used as discriminators or constrained strings
- `z.string().nullable()` / `z.nullable(...)`
- `z.record(z.string())` — open-shape maps
- `z.tuple([...])` — fixed-arity arrays
- `z.date()`
- `z.number().default(10)` / any `.default(...)`
- `z.string().refine(...)` / `z.string().transform(...)` — refinements and transforms (`ZodEffects`)
- `z.lazy(...)` — recursive schemas

The converter also reads `def.typeName`, which is the **Zod v3** internal API. Zod v4 restructured `_def` so this access pattern fails even for the supported types — every Zod 4 schema currently degrades to the empty-object fallback.

The package declares `zod-to-json-schema` as a dependency at `voice/google-gemini-live-api/package.json:53` but the import is never used; it's effectively a dead dep.

The rest of the Mastra monorepo uses `zodToJsonSchema` from `@mastra/schema-compat/zod-to-json` (re-exported by `packages/core/src/zod-to-json.ts`) — for example `packages/core/src/datasets/dataset.ts:136`. Replacing the hand-rolled converter with this canonical one would cover the full Zod surface area, work on Zod 3 and Zod 4, drop the dead dep, and align this package with the rest of the monorepo.

This is **distinct from #10161** (closed by #10226). #10161 was about inbound `tool_call` argument extraction; this bug is about outbound JSON-Schema generation at registration time. The two compound — even a fully-fixed inbound parser cannot recover args the model never produced because no schema was advertised in setup.

## Steps To Reproduce

```bash
git clone https://github.com/petrchiriboga/mastra-voice-google-gemini-live-zod-converter-repro
cd mastra-voice-google-gemini-live-zod-converter-repro
pnpm install
pnpm repro
```

The repro calls the private `convertZodSchemaToJsonSchema` method directly with nine representative Zod schemas and prints the JSON Schema produced for each. No network calls, no API key needed.

https://github.com/petrchiriboga/mastra-voice-google-gemini-live-zod-converter-repro

## Expected Behavior

Every Zod schema passed to `addTools(...)` should be converted to a faithful JSON Schema before it reaches the Gemini Live setup message. `z.union`, `z.discriminatedUnion`, `z.literal`, `z.string().nullable()`, `z.record`, `z.tuple`, `z.date`, `z.string().refine(...)`, `z.number().default(10)`, and `z.lazy(...)` should all produce schemas that describe their actual shape, not the empty `{ type: 'object', properties: {} }` fallback.

The fix is a one-import swap. Replace the hand-rolled converter at `voice/google-gemini-live-api/src/index.ts:2071-2165` with:

```ts
import { zodToJsonSchema } from '@mastra/schema-compat/zod-to-json';
```

…and drop the unused `zod-to-json-schema` dependency. After the swap, all nine cases in the repro should print faithful JSON Schemas.

## Environment Information

```
System:
  OS: macOS 15.3.1
  CPU: arm64 Apple M4 Pro
Binaries:
  Node: 24.13.0
  pnpm: 10.30.2
npmPackages:
  @mastra/voice-google-gemini-live: 0.11.4
  @mastra/core: 0.24.4
  zod: 3.23.8
```

LLM Provider: Google Gemini (Live API).

The repro deliberately uses Zod v3 to stay within the converter's nominal compatibility window. Zod v4 schemas additionally fail because the `def.typeName` access pattern itself is gone.

## Verification

### LINKED ISSUE #17057: [BUG] @mastra/schema-compat: GoogleSchemaCompatLayer breaks Gemini REST tool calls for `z.discriminatedUnion`, `z.lazy`, and `z.tuple` inputs
## Describe the Bug

`GoogleSchemaCompatLayer` emits JSON Schema 2020-12 keywords that aren't part of the OpenAPI 3.0 `Schema` Object declared by `@google/genai`'s `FunctionDeclaration.parameters` typedef. Three Zod patterns fail against Gemini REST tool-calling as a result. All three reproduce on `gemini-2.5-flash` and `gemini-2.5-pro` via the standalone repro at https://github.com/petrchiriboga/mastra-schema-compat-google-openapi-3-repro.

| Zod input | Layer emits (out-of-spec) | REST symptom |
|---|---|---|
| `z.discriminatedUnion(...)` | `oneOf` + string `const` | **Silent mis-fill** — model uses wrong discriminator name, wrong field names, wrong types; AI SDK rejects as `AI_InvalidToolInputError`; `finishReason: tool-calls` but `execute()` never runs |
| `z.lazy(...)` (recursive) | `$ref` + `definitions` | Model can't follow `$ref` (silent — same pattern as `oneOf`) |
| `z.tuple([...])` | `items: [array]` (Draft-4 form) | **Hard HTTP 400** — `Invalid JSON payload received. Unknown name "items" at ... 'Proto field is not repeating, cannot start list.'` |

Same union shape expressed with `z.union` (which emits `anyOf`) works correctly — isolating the `discriminatedUnion` failure to the `oneOf` keyword itself, not the union semantics:

| Case | Layer emits | flash | pro |
|---|---|---|---|
| `z.discriminatedUnion` | `oneOf` | ❌ FAIL | ❌ FAIL |
| `z.union` (control, same shape) | `anyOf` | ✅ PASS | ✅ PASS |

Raw failure for `z.discriminatedUnion` (flash):

```jsonc
{
  "invalid": true,
  "input": { "shape": { "radius": "5cm", "type": "circle" } },  // schema declared `kind` / `r`
  "error": { "name": "AI_InvalidToolInputError",
             "cause": { "message": "invalid_union: No matching discriminator. discriminator: \"kind\"" } }
}
```

Raw failure for `z.tuple` (flash REST API):

```
400 Bad Request — Invalid JSON payload received.
Unknown name "items" at 'tools[0].function_declarations[0].parameters.properties[0].value':
Proto field is not repeating, cannot start list.
```

Beyond these three, the layer also emits `$schema`, root `additionalProperties: false`, `propertyNames`, sub-schema `additionalProperties`, and string `const` — all tolerated by REST today but rejected by Gemini Live's stricter setup validator ([#17020 verdict table](https://github.com/mastra-ai/mastra/issues/17020)). The proposed fix below covers both surfaces.

## Steps To Reproduce

```bash
git clone https://github.com/petrchiriboga/mastra-schema-compat-google-openapi-3-repro
cd mastra-schema-compat-google-openapi-3-repro
pnpm install
GOOGLE_GENERATIVE_AI_API_KEY=... pnpm repro       # PROBE_MODEL=gemini-2.5-pro to switch
```

Without the key, dumps the layer output only (still shows `oneOf` for `z.discriminatedUnion`, `$ref` + `definitions` for `z.lazy`, `items: [array]` for `z.tuple`). With the key, also runs the live REST probe.

https://github.com/petrchiriboga/mastra-schema-compat-google-openapi-3-repro

## Expected Behavior

Layer output should match the OpenAPI 3.0 `Schema` Object declared by `@google/genai`:

- `z.discriminatedUnion` → `anyOf` with `enum: [literal]` discriminators (no `oneOf`, no `const`)
- `z.literal('x')` → `{ type: 'string', enum: ['x'] }`
- `z.lazy(...)` recursive → outer ref inlined one level, inner self-refs collapsed to `{ type: 'object' }`, no `$ref`, no `definitions`
- `z.tuple([A, B])` → `{ type: 'array', items: { anyOf: [A, B] } }` (loses position info, stays in spec)
- `z.nullable(X)` → `{ ...X, nullable: true }` on both `processToJSONSchema` and `processToAISDKSchema` paths (today only the latter collapses it)
- `z.record(K, V)` → `{ type: 'object' }` (no `propertyNames`, no sub-schema `additionalProperties`)
- Root objects: no `$schema`, no `additionalProperties: false` default

In particular, `z.discriminatedUnion`, `z.lazy`, and `z.tuple` tool inputs should produce valid args that reach `execute()`.

## Environment Information

```
mastra: main @ 5b22ffd74c
@mastra/schema-compat: 1.2.10
zod: 4.3.6
@ai-sdk/google: 2.0.74
ai: 5.0.192
LLM Provider: Google Gemini (REST generateContent)
Models confirmed affected: gemini-2.5-flash, gemini-2.5-pro
Node: 24.13.0
```

## Proposed Fix

All in [`packages/schema-compat/src/provider-compats/google.ts`](https://github.com/mastra-ai/mastra/blob/main/packages/schema-compat/src/provider-compats/google.ts):

1. **In `postProcessJSONNode`**, recursively strip / rewrite per node:
   - `delete $schema`, `delete propertyNames`, `delete additionalProperties` (both sub-schema and `false` forms)
   - `oneOf → anyOf` (**load-bearing REST correctness fix**)
   - string `const → enum: [const]`
   - `items: [array]` (tuple form) → `items: { anyOf: [...] }` (single-schema form) — fixes the HTTP 400
2. **New `inlineRefsAndDropDefinitions` post-pass** — expand recursive `$ref`s one level (model sees the outer shape), collapse further self-refs to `{ type: 'object' }` (no infinite loop), drop the top-level `definitions` map. Applied on both `processToJSONSchema` and `processToAISDKSchema` paths.
3. **Run `fixAISDKNullableUnionTypes`'s null-branch collapse on the `processToJSONSchema` path too** — today it only runs on `processToAISDKSchema`.

After this, PR #17023's downstream `gemini-schema-sanitizer.ts` becomes unnecessary.

**Test gap:** `google.e2e.test.ts`'s `allSchemas` had no `z.literal`, `z.discriminatedUnion`, or `z.tuple` — none of the three failing patterns were exercised, which is how the bugs slipped through CI. Adding all three cases is part of the fix.

**Risk:** verify against existing `google.e2e.test.ts` and any structured-output e2e — REST structured-output may depend on `additionalProperties: false` being present today.

A PR with the fix + tests + changeset will follow. Sibling architectural pattern to #17051 / #17052.

## Verification

petrchiriboga:
## Description

`@mastra/voice-google-gemini-live` shipped a hand-rolled `convertZodSchemaToJsonSchema` /
`convertZodDefToJsonSchema` (~95 lines) that only understood a hardcoded short list of Zod 3
types — `ZodString`, `ZodNumber`, `ZodBoolean`, `ZodArray`, `ZodObject`, `ZodOptional`,
`ZodEnum` — and dropped every other Zod type to an empty `{ type: 'object', properties: {} }`
fallback. Tools registered via `voice.addTools(...)` with common patterns (`z.union`,
`z.discriminatedUnion`, `z.literal`, `.nullable()`, `z.record`, `z.tuple`, `z.date`,
`.refine()`, `.default()`, `z.lazy()`) reached Gemini Live with no schema, so the model got no
signal about tool arguments and returned empty args. The converter also read `def.typeName`
(a Zod 3 internal), so every Zod 4 schema fell through too.

This PR routes tool Zod schemas through `@mastra/schema-compat`'s `applyCompatLayer` +
`GoogleSchemaCompatLayer` — the in-repo Google pipeline the rest of the monorepo uses. It
covers the full Zod 3/4 surface and emits the **OpenAPI 3.0 Schema Object** shape Gemini
Live's setup validator requires (`tools[].function_declarations[].parameters` is OpenAPI 3.0
per `@google/genai`'s `Schema` typedef, not JSON Schema 2020-12), so no Live-specific
post-processing is needed in this package.

**Merge order — top of a four-PR chain:**
1. **#17052** — patches a Zod v4 `z.record()` crash in schema-compat's standard-schema adapter (pre-processing prerequisite for `applyCompatLayer`).
2. **#17060** — aligns `GoogleSchemaCompatLayer` output with the OpenAPI 3.0 Schema Object (fixes #17057; post-processing).
3. **#17019** — native-audio setup + tool wire shape (fixes #17018).
4. **#17023** (this PR) — the Zod-routing fix.

Held as **draft** until #17052, #17060, and #17019 merge. This branch currently carries
#17019's native-audio commit as its base; once #17019 lands in `main` and this branch is
rebased, that commit drops out and the diff narrows to the pure Zod-routing change.

Side effects:
- Added `@mastra/schema-compat` as a dependency; removed the dead `zod-to-json-schema` devDependency (declared, never imported).
- Removed the stale `vi.mock('zod-to-json-schema', ...)` block in `index.test.ts` that masked the bug.
- Added a `Zod schema → JSON Schema for tool parameters` describe block in `index.test.ts`.

## Related issue(s)

Fixes #17020.

## Type of change

- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [ ] Test update

## Checklist

- [ ] I have made corresponding changes to the documentation (if applicable) — N/A: no public API surface or behavior contract changes.
- [ ] I have addressed all Coderabbit comments on this PR — will address as feedback arrives.

### DIFF
--- a/.changeset/shaky-singers-teach.md
+++ b/.changeset/shaky-singers-teach.md
@@ -1,0 +1,7 @@
+---
+'@mastra/voice-google-gemini-live': patch
+---
+
+Tools registered via `voice.addTools(...)` with rich Zod schemas now reach Gemini Live with a faithful schema. The package routes Zod through `@mastra/schema-compat`'s `applyCompatLayer` + `GoogleSchemaCompatLayer` — the in-repo Google pipeline — so the tool parameters land in the OpenAPI 3.0 shape Gemini Live's setup validator accepts.
+
+Fixes #17020.
--- a/.changeset/wide-cities-jog.md
+++ b/.changeset/wide-cities-jog.md
@@ -1,0 +1,15 @@
+---
+'@mastra/voice-google-gemini-live': patch
+---
+
+**Fixed** Gemini Live sessions now connect successfully when using native-audio models. Previously the connection failed during session setup.
+
+**Fixed** tools are now invoked correctly. Previously tool calls were silently ignored even when tools were registered during setup.
+
+**Fixed** tool results of any shape (arrays, primitives, objects) are now accepted. Previously, non-object tool return values caused sessions to close unexpectedly.
+
+**Fixed** the `speaker` option is now honored when passed at the `VoiceConfig` root alongside `realtimeConfig`, not only when passed in the flat config shape.
+
+**Changed** default model from `gemini-2.0-flash-exp` (shut down 2025-12-09) to `gemini-3.1-flash-live-preview` (Google's current Live API quickstart model). If you weren't explicitly setting `model`, your sessions will start connecting again.
+
+Fixes #17018.
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -6807,6 +6807,9 @@ importers:
       '@google/genai':
         specifier: ^1.45.0
         version: 1.52.0(@modelcontextprotocol/sdk@1.29.0(@cfworker/json-schema@4.1.1)(zod@4.4.3))(bufferutil@4.1.0)
+      '@mastra/schema-compat':
+        specifier: workspace:*
+        version: link:../../packages/schema-compat
       google-auth-library:
         specifier: ^10.6.1
         version: 10.6.2
@@ -6853,9 +6856,6 @@ importers:
       zod:
         specifier: 'catalog:'
         version: 4.4.3
-      zod-to-json-schema:
-        specifier: ^3.25.1
-        version: 3.25.1(zod@4.4.3)
 
   voice/inworld:
     dependencies:
--- a/voice/google-gemini-live-api/package.json
+++ b/voice/google-gemini-live-api/package.json
@@ -33,6 +33,7 @@
   "license": "Apache-2.0",
   "dependencies": {
     "@google/genai": "^1.45.0",
+    "@mastra/schema-compat": "workspace:*",
     "google-auth-library": "^10.6.1",
     "ws": "^8.20.0"
   },
@@ -49,8 +50,7 @@
     "tsx": "^4.21.0",
     "typescript": "catalog:",
     "vitest": "catalog:",
-    "zod": "catalog:",
-    "zod-to-json-schema": "^3.25.1"
+    "zod": "catalog:"
   },
   "peerDependencies": {
     "@mastra/core": ">=1.0.0-0 <2.0.0-0"
--- a/voice/google-gemini-live-api/src/index.test.ts
+++ b/voice/google-gemini-live-api/src/index.test.ts
@@ -1,5 +1,6 @@
 import { PassThrough } from 'node:stream';
 import { describe, it, expect, vi, beforeEach, afterEach } from 'vitest';
+import { z } from 'zod';
 import { GeminiLiveVoice } from './index';
 
 // Mock WebSocket
@@ -45,15 +46,6 @@ vi.mock('google-auth-library', () => {
 
   return { GoogleAuth: MockGoogleAuth };
 });
-
-// Mock zod-to-json-schema
-vi.mock('zod-to-json-schema', () => ({
-  zodToJsonSchema: vi.fn().mockImplementation(() => ({
-    type: 'object',
-    properties: {},
-    $schema: 'http://json-schema.org/draft-07/schema#',
-  })),
-}));
 
 describe('GeminiLiveVoice', () => {
   let voice: GeminiLiveVoice;
@@ -1111,6 +1103,96 @@ describe('GeminiLiveVoice', () => {
       expect(usage.outputTokens).toBe(2);
       expect(usage.totalTokens).toBe(3);
       expect(['audio', 'text', 'video']).toContain(usage.modality);
+    });
+  });
+
+  describe('Zod schema → JSON Schema for tool parameters', () => {
+    async function captureToolParameters(zodSchema: unknown): Promise<any> {
+      const v = new GeminiLiveVoice({ apiKey: 'k' });
+      v.addTools({
+        probe: {
+          id: 'probe',
+          description: 'probe tool',
+          inputSchema: zodSchema,
+          execute: vi.fn() as any,
+        },
+      });
+
+      vi.spyOn((v as any).connectionManager, 'waitForOpen').mockResolvedValue(undefined as any);
+      (v as any).waitForSessionCreated = vi.fn().mockResolvedValue(undefined);
+
+      await v.connect();
+
+      const wsSent = ((v as any).connectionManager.getWebSocket() as any).send as any;
+      const payloads = wsSent.mock.calls.map((c: any[]) => JSON.parse(c[0]));
+      const setupMsg = payloads.find((p: any) => p.setup);
+      const toolEntry = setupMsg.setup.tools?.[0];
+      // The wire shape (camelCase `functionDeclarations` vs snake_case `function_declarations`) is
+      // tracked separately in #17018; pick whichever container the current setup builder produced.
+      const declarations = toolEntry?.functionDeclarations ?? toolEntry?.function_declarations;
+      return declarations?.[0]?.parameters;
+    }
+
+    it('converts ZodUnion to a real JSON Schema (not the empty fallback)', async () => {
+      const params = await captureToolParameters(z.object({ value: z.union([z.string(), z.number()]) }));
+      const value = params.properties.value;
+      // The old hand-rolled converter dropped unions to `{ type: 'object', properties: {} }`.
+      // zod-to-json now emits a union via `anyOf` / `oneOf`. We don't pin the exact keyword —
+      // both forms describe "string or number" faithfully and are what we care about.
+      const branches = value.anyOf ?? value.oneOf;
+      expect(branches).toBeDefined();
+      expect(branches.map((b: any) => b.type).sort()).toEqual(['number', 'string']);
+    });
+
+    it('converts ZodLiteral to a const-style JSON Schema', async () => {
+      const params = await captureToolParameters(z.object({ mode: z.literal('strict') }));
+      const mode = params.properties.mode;
+      // Faithful conversion preserves the literal value via `const` or single-element `enum`.
+      const literalValue =
+        mode.const ?? (Array.isArray(mode.enum) && mode.enum.length === 1 ? mode.enum[0] : undefined);
+      expect(literalValue).toBe('strict');
+    });
+
+    it('converts ZodDiscriminatedUnion (tagged union) to a branched JSON Schema', async () => {
+      const params = await captureToolParameters(
+        z.object({
+          event: z.discriminatedUnion('kind', [
+            z.object({ kind: z.literal('click'), x: z.number() }),
+            z.object({ kind: z.literal('hover'), seconds: z.number() }),
+          ]),
+        }),
+      );
+      const event = params.properties.event;
+      const branches = event.anyOf ?? event.oneOf;
+      expect(branches).toBeDefined();
+      expect(branches).toHaveLength(2);
+      // Each branch must be an object with its discriminator + its own payload field — i.e. the
+      // type information actually survived the conversion.
+      for (const branch of branches) {
+        expect(branch.type).toBe('object');
+        expect(Object.keys(branch.properties)).toContain('kind');
+      }
+    });
+
+    it('converts a nullable field to a faithful nullable JSON Schema', async () => {
+      const params = await captureToolParameters(z.object({ note: z.string().nullable() }));
+      const note = params.properties.note;
+      // Accept any faithful encoding (anyOf [string, null]; `type: ['string','null']`; or `nullable: true`).
+      const acceptableNullable =
+        (Array.isArray(note.type) && note.type.includes('null')) ||
+        note.nullable === true ||
+        (Array.isArray(note.anyOf) && note.anyOf.some((b: any) => b.type === 'null'));
+      expect(acceptableNullable).toBe(true);
+    });
+
+    it('passes through a tool that already supplies a JSON Schema in inputSchema unchanged', async () => {
+      const jsonSchema = {
+        type: 'object',
+        properties: { q: { type: 'string' } },
+        required: ['q'],
+      };
+      const params = await captureToolParameters(jsonSchema);
+      expect(params).toEqual(jsonSchema);
     });
   });
 });
--- a/voice/google-gemini-live-api/src/index.test.ts
+++ b/voice/google-gemini-live-api/src/index.test.ts
@@ -812,7 +812,251 @@ describe('GeminiLiveVoice', () => {
       const setupMsg = payloads.find((p: any) => p.setup);
       expect(setupMsg).toBeDefined();
       expect(setupMsg.setup.model).toBe('models/gemini-2.0-flash-exp');
-      expect(setupMsg.setup.systemInstruction.parts[0].text).toBe('You are test');
+      expect(setupMsg.setup.system_instruction.parts[0].text).toBe('You are test');
+    });
+
+    it('connect() should default model to gemini-3.1-flash-live-preview when none is supplied', async () => {
+      const v = new GeminiLiveVoice({ apiKey: 'k' });
+
+      vi.spyOn((v as any).connectionManager, 'waitForOpen').mockResolvedValue(undefined as any);
+      (v as any).waitForSessionCreated = vi.fn().mockResolvedValue(undefined);
+
+      await v.connect();
+
+      const wsSent = ((v as any).connectionManager.getWebSocket() as any).send as any;
+      const payloads = wsSent.mock.calls.map((c: any[]) => JSON.parse(c[0]));
+      const setupMsg = payloads.find((p: any) => p.setup);
+      expect(setupMsg.setup.model).toBe('models/gemini-3.1-flash-live-preview');
+    });
+
+    it('connect() should always emit generation_config.response_modalities: ["AUDIO"]', async () => {
+      const v = new GeminiLiveVoice({ apiKey: 'k' });
+
+      vi.spyOn((v as any).connectionManager, 'waitForOpen').mockResolvedValue(undefined as any);
+      (v as any).waitForSessionCreated = vi.fn().mockResolvedValue(undefined);
+
+      await v.connect();
+
+      const wsSent = ((v as any).connectionManager.getWebSocket() as any).send as any;
+      const payloads = wsSent.mock.calls.map((c: any[]) => JSON.parse(c[0]));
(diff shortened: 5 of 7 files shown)


petrchiriboga:
**Heads up on merge order with #17019.** This PR was branched off `main` and rewrites the two `this.convertZodSchemaToJsonSchema(tool.inputSchema)` call sites that exist there. #17019 introduces a private `buildToolDeclarations` helper that absorbs both call sites into a single function which still calls `this.convertZodSchemaToJsonSchema`. If #17019 lands first (it's approved by @intojhanurag), this PR's rebase will hit a structural conflict: the call sites this PR targets no longer exist on `main`, but the new helper still references the deleted method.

Concrete patch for the rebase, for reviewers' confidence:

```diff
   private buildToolDeclarations(
     configTools: GeminiToolConfig[] | undefined,
     registeredTools: ToolsInput | undefined,
   ): Array<{ name: string; description?: string; parameters?: unknown }> {
     ...
     for (const [toolName, tool] of Object.entries(registeredTools)) {
       try {
         let parameters: unknown;
         if ('inputSchema' in tool && tool.inputSchema) {
           if (typeof tool.inputSchema === 'object' && 'safeParse' in tool.inputSchema) {
-            parameters = this.convertZodSchemaToJsonSchema(tool.inputSchema);
+            parameters = zodToJsonSchema(tool.inputSchema);
           } else {
             parameters = tool.inputSchema;
           }
         }
         ...
```

I'll rebase and push as soon as #17019 merges. Until then this PR is intentionally held in draft to avoid a conflicting double-merge.

Separately, before merge I want to e2e-validate against `gemini-3.1-flash-live-preview` (or whatever default ships in #17019) with a real Zod-typed tool: schema-compat injects `additionalProperties: false` (v4 path) and `$schema` (v3 path) on the produced JSON Schema, and Gemini Live's `function_declarations[].parameters` accepts only an OpenAPI subset. The hand-rolled converter never emitted either key, so the wire-acceptance of the new keys is currently unverified. I'll post results in a follow-up comment.

petrchiriboga:
End-to-end probing reveals the simple converter swap is **insufficient to fix #17020 as advertised** — Gemini Live's `tools[].function_declarations[].parameters` is OpenAPI 3.0 Schema, not JSON Schema 2020-12, and rejects more than just `\$schema`.

Procedure: built the rebased state (this PR on top of #17019), tested four Zod-schema scenarios via `voice.connect()` against `gemini-3.1-flash-live-preview` — all FAIL with `Session creation timeout` because Gemini's setup validator closes the WS with 1007 before `setupComplete` arrives. Then narrowed by raw-WebSocket wire-probe to isolate which keys block:

| # | Wire shape | Verdict |
|---|---|---|
| A | Minimal `{ type: 'object', properties: {q: {type:'string'}}, required: ['q'] }` (known good from #17019) | **PASS** |
| B | A + `\$schema: '<draft 2020-12>'` (schema-compat v4 path) | FAIL — `Unknown name \"\$schema\" at 'setup.tools[0].function_declarations[0].parameters'` |
| C | A + `additionalProperties: false` (schema-compat v4 strict-mode injection) | FAIL — `Unknown name \"additionalProperties\"` |
| D | A + both | FAIL — `\$schema` (first error) |
| E | A + `\$schema: '<draft-07>'` (schema-compat v3 path) | FAIL — same `\$schema` rejection |
| F | `additionalProperties: { type: 'string' }` (the `z.record` form) | FAIL — `Unknown name \"additionalProperties\"` |
| G | `anyOf: [...]` (the `z.union` form) | **PASS** |
| H | `oneOf: [...]` with `const` (the `z.discriminatedUnion` form) | FAIL — `Unknown name \"const\"` |
| I | `type: ['string','null']` (the `.nullable()` form) | FAIL — `Unknown name \"type\"` (the array form) |
| J | `const: 'X'` (the `z.literal` form) | FAIL — `Unknown name \"const\"` |

Gemini Live's parameter validator rejects: `\$schema` (any URL), `additionalProperties` (any form — boolean or sub-schema), `oneOf`, `const`, and `type: ['T', 'null']`. Only `anyOf` survives intact.

This means schema-compat's output is incompatible with Gemini for nearly every Zod feature beyond `z.string()` / `z.number()` / `z.object()` / `z.union()`. The patterns that motivated #17020 (`z.union`, `z.discriminatedUnion`, `z.literal`, `.nullable()`, `z.record`) all still produce rejected payloads after this PR — just at the wire instead of in the local converter.

**Required to actually fix #17020**: a Gemini-specific sanitizer that converts JSON Schema → OpenAPI 3.0 Schema after `zodToJsonSchema(...)`. At minimum:
- strip `\$schema`
- strip `additionalProperties` recursively (record semantics become open shapes)
- rewrite `oneOf` → `anyOf`
- rewrite `const: 'X'` → `enum: ['X']`
- rewrite `type: ['T', 'null']` → `type: 'T'` (and possibly `nullable: true` if Gemini accepts it on Live — TBD by probe)

I'll hold this PR in draft and refile #17020 with the expanded scope, or push a follow-up commit here that adds the sanitizer + probes nullable handling. Open to either path.

Repro:
- E2E (against the FIXED build, rebased onto #17019): https://github.com/petrchiriboga/mastra-voice-google-gemini-live-zod-converter-repro (probe results in `VALIDATION.md`)
- Wire-level isolation (raw WebSocket): bundled in the same repro

petrchiriboga:
Closed #17020  is closed

PR #17033, opened 2026-05-25
fix(core): use `extends object` for structured output detection in AgentExecutionOptions

### LINKED ISSUE #16732: [BUG] AgentExecutionOptions<undefined> requires structuredOutput under strictNullChecks: false
### Describe the Bug

`@mastra/core@1.35.0` defines:

```ts
// packages/core/src/agent/agent.types.ts
export type AgentExecutionOptions<OUTPUT = unknown> = AgentExecutionOptionsBase<OUTPUT> &
  (OUTPUT extends {} ? { structuredOutput: StructuredOutputOptions<OUTPUT> } : { structuredOutput?: never });
```

`AgentConfig.defaultOptions` resolves to `AgentExecutionOptions<TOutput>` where `TOutput` defaults to `undefined` (per the `Agent` / `AgentConfig` generic defaults — established in #11741 / #10710).

The conditional `OUTPUT extends {}` resolves differently across TypeScript modes:

- **`strictNullChecks: true`** — `undefined extends {}` is `false` → `{ structuredOutput?: never }` → optional → matches Mastra's own templates (`defaultOptions: { maxSteps: 50 }`, `defaultOptions: { stopWhen: stepCountIs(1) }`, etc.).
- **`strictNullChecks: false`** — `undefined extends {}` is `true` → `{ structuredOutput: required }` → consumers omitting `structuredOutput` get:

```
TS2322: Type '{ providerOptions: {…} }' is not assignable to type 'DynamicArgument<AgentExecutionOptions<undefined>, unknown>'.
  Property 'structuredOutput' is missing in type '{ providerOptions: {…} }'
  but required in type '{ structuredOutput: StructuredOutputOptions<undefined>; }'.
```

The pattern the consumer writes is the same one used in `templates/template-github-review-agent/src/mastra/agents/workflow-review-agent.ts` and `examples/agent-v6/src/mastra/agents/index.ts` — those compile clean under strict mode but fail under non-strict.

### Steps To Reproduce

```bash
git clone <minimal-repro>     # see gist link below
npm install
npx tsc --project tsconfig.strict.json     # passes
npx tsc --project tsconfig.nostrict.json   # fails (TS2322, see above)
```

Minimal source (also in `src/repro.ts` in the gist):

```ts

new Agent({
  id: 'demo',
  name: 'demo',
  instructions: 'demo',
  model: {} as any,
  defaultOptions: {
    providerOptions: { openai: { reasoningEffort: 'low' as const } },
  },
});
```

https://gist.github.com/rares04/478a8df96b21170404eda91cc6e21a69

The gist contains the full setup (package.json, two tsconfigs, the repro file, and a `proposed-fix.ts` empirically verifying the two fix variants below).

### Expected Behavior

`AgentExecutionOptions<undefined>` (and by extension `AgentConfig.defaultOptions` with the default `TOutput = undefined`) should accept the same shapes Mastra's own templates use — `{ providerOptions }`, `{ maxSteps }`, `{ stopWhen }` — without requiring `structuredOutput`, regardless of `strictNullChecks` setting.

### Suggested fix

Either of the two changes below — both verified empirically (see `src/proposed-fix.ts` in the gist) — fix non-strict mode without regressing strict mode. Same patch applied to `AgentExecutionOptions` and `PublicAgentExecutionOptions` (they share the pattern).

**Option B — explicit nullish check first:**

```diff
 // packages/core/src/agent/agent.types.ts
 export type AgentExecutionOptions<OUTPUT = unknown> = AgentExecutionOptionsBase<OUTPUT> &
-  (OUTPUT extends {} ? { structuredOutput: StructuredOutputOptions<OUTPUT> } : { structuredOutput?: never });
+  ([OUTPUT] extends [undefined | null | void]
+    ? { structuredOutput?: never }
+    : OUTPUT extends {}
+      ? { structuredOutput: StructuredOutputOptions<OUTPUT> }
+      : { structuredOutput?: never });
```

**Option D — `NonNullable` check (uses a built-in helper):**

```diff
 export type AgentExecutionOptions<OUTPUT = unknown> = AgentExecutionOptionsBase<OUTPUT> &
-  (OUTPUT extends {} ? { structuredOutput: StructuredOutputOptions<OUTPUT> } : { structuredOutput?: never });
+  ([NonNullable<OUTPUT>] extends [never]
+    ? { structuredOutput?: never }
+    : OUTPUT extends {}
+      ? { structuredOutput: StructuredOutputOptions<OUTPUT> }
+      : { structuredOutput?: never });
```

Tuple wrappers (`[OUTPUT] extends [...]`) disable union distribution — they're there so the check stays well-defined when someone instantiates with a union like `string | undefined`.

Also worth considering: documenting `strict: true` (or at least `strictNullChecks: true`) as a hard requirement for consumers, since Mastra's own templates and examples all use strict mode. Either response is fine from the consumer side; would just be useful to have an authoritative answer.

### Environment Information

```block
System:
  OS: macOS 15.7.4
  CPU: arm64 Apple M4 Pro
Binaries:
  Node: 24.15.0
  npm: 11.12.1
npmPackages:
  @mastra/core: 1.35.0
  typescript: 5.9.2
```

LLM Provider: N/A (type-only bug, no runtime needed).

### Verification

serhiizghama:
## Problem

`AgentExecutionOptions<undefined>` (the default when no `TOutput` generic is provided) compiles incorrectly under `strictNullChecks: false`.

The conditional `OUTPUT extends {}` resolves differently across TypeScript modes:
- **`strictNullChecks: true`**: `undefined extends {}` → `false` → `structuredOutput` is optional ✓
- **`strictNullChecks: false`**: `undefined extends {}` → `true` → `structuredOutput` becomes **required** ✗

Since `Agent<TOutput = undefined>` uses `undefined` as the default, this causes a TS2322 error for any project with non-strict settings:

```ts
// Fails under strictNullChecks: false
new Agent({
  name: 'my-agent',
  instructions: '...',
  model: myModel,
  defaultOptions: { maxSteps: 50 }, // TS2322: missing required 'structuredOutput'
});
```

Closes #16732

## Solution

Replace `OUTPUT extends {}` with `OUTPUT extends object` in the four locations within `AgentExecutionOptions`, `PublicAgentExecutionOptions`, `InnerAgentExecutionOptions`, and `NetworkOptions`.

`{}` includes `null` and `undefined` under `strictNullChecks: false`, while `object` consistently excludes all primitives and `undefined`/`null` in both strict and non-strict modes.

| Type | `extends {}` (strict=false) | `extends object` |
|------|---------------------------|-----------------|
| `undefined` | ✅ true (broken) | ❌ false (correct) |
| `null` | ✅ true (broken) | ❌ false (correct) |
| `string` | ✅ true | ❌ false |
| `{ text: string }` | ✅ true | ✅ true |
| `unknown` | ❌ false | ❌ false |

## Testing

This is a pure TypeScript type change — no runtime behavior is affected. The fix can be verified by creating a `tsconfig.json` with `"strictNullChecks": false` and checking that `new Agent({ defaultOptions: { maxSteps: 50 } })` compiles without errors.

## ELI5

TypeScript has different "strictness" modes. In less-strict mode, a code check (`OUTPUT extends {}`) was incorrectly telling TypeScript that every agent needs to provide a `structuredOutput` configuration, even when it shouldn't. The fix changes that check to (`OUTPUT extends object`) so it works correctly whether you're in strict or less-strict mode, letting developers use agents without explicitly providing this optional setting.

---

### Type System Fix for AgentExecutionOptions

Changed the structured output detection logic in `AgentExecutionOptions`, `PublicAgentExecutionOptions`, and `NetworkOptions` types from using `OUTPUT extends {}` to `OUTPUT extends object`. This ensures that the optional nature of `structuredOutput` is correctly preserved across TypeScript compilation modes, particularly when `strictNullChecks` is disabled.

The root issue: Under `strictNullChecks: false`, the condition `undefined extends {}` evaluates to `true`, incorrectly making `structuredOutput` a required property for agents without explicit output types. The `object` type constraint consistently excludes primitives, `undefined`, and `null` across all TypeScript modes, fixing this behavior.

### Files Modified

- **packages/core/src/agent/agent.types.ts**: Updated conditional type logic in `NetworkOptions<OUTPUT>`, `PublicAgentExecutionOptions<OUTPUT>`, and `AgentExecutionOptions<OUTPUT>` to use `OUTPUT extends object` for determining `structuredOutput` requirement.
- **.changeset/three-ravens-cross.md**: Added changeset entry documenting the patch fix to `@mastra/core`.

### Impact

This is a purely type-level fix with no runtime impact. Untyped agents (using the default `undefined` for `OUTPUT`) now correctly treat `structuredOutput` as optional regardless of TypeScript strictness settings. This resolves compilation errors when using common patterns like `{ defaultOptions: { maxSteps: 50 } }` in projects with `strictNullChecks: false`.

[Review Change Stack](https://app.coderabbit.ai/change-stack/mastra-ai/mastra/pull/17033?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/.changeset/three-ravens-cross.md
+++ b/.changeset/three-ravens-cross.md
@@ -1,0 +1,5 @@
+---
+'@mastra/core': patch
+---
+
+Fix `AgentExecutionOptions` to use `extends object` instead of `extends {}` for structured output detection. The `{}` condition resolves differently under `strictNullChecks: false`, causing `undefined` (the default `TOutput`) to be treated as a concrete output type and making `structuredOutput` required in `defaultOptions`. Using `extends object` correctly excludes `undefined` and `unknown` from requiring structured output in both strict and non-strict projects.
--- a/packages/core/src/agent/agent.types.ts
+++ b/packages/core/src/agent/agent.types.ts
@@ -406,7 +406,7 @@ export type NetworkOptions<OUTPUT = undefined> = {
    * const result = await stream.object;
    * ```
    */
-  structuredOutput?: PublicStructuredOutputOptions<OUTPUT extends {} ? OUTPUT : never>;
+  structuredOutput?: PublicStructuredOutputOptions<OUTPUT extends object ? OUTPUT : never>;
 
   /** Callback fired after each LLM step within a sub-agent execution */
   onStepFinish?: LoopConfig<OUTPUT>['onStepFinish'];
@@ -608,14 +608,14 @@ export type AgentExecutionOptionsBase<OUTPUT> = {
  * Use this type for public method signatures.
  */
 export type PublicAgentExecutionOptions<OUTPUT = unknown> = AgentExecutionOptionsBase<OUTPUT> &
-  (OUTPUT extends {} ? { structuredOutput: PublicStructuredOutputOptions<OUTPUT> } : { structuredOutput?: never });
+  (OUTPUT extends object ? { structuredOutput: PublicStructuredOutputOptions<OUTPUT> } : { structuredOutput?: never });
 
 /**
  * Internal agent execution options that require StandardSchemaWithJSON.
  * Use this type internally after converting from PublicSchema.
  */
 export type AgentExecutionOptions<OUTPUT = unknown> = AgentExecutionOptionsBase<OUTPUT> &
-  (OUTPUT extends {} ? { structuredOutput: StructuredOutputOptions<OUTPUT> } : { structuredOutput?: never });
+  (OUTPUT extends object ? { structuredOutput: StructuredOutputOptions<OUTPUT> } : { structuredOutput?: never });
 
 export type InnerAgentExecutionOptions<OUTPUT = unknown> = AgentExecutionOptionsBase<OUTPUT> & {
   outputWriter?: OutputWriter;
@@ -629,4 +629,4 @@ export type InnerAgentExecutionOptions<OUTPUT = unknown> = AgentExecutionOptions
     snapshot: any;
   };
   toolCallId?: string;
-} & (OUTPUT extends {} ? { structuredOutput: StructuredOutputOptions<OUTPUT> } : { structuredOutput?: never });
+} & (OUTPUT extends object ? { structuredOutput: StructuredOutputOptions<OUTPUT> } : { structuredOutput?: never });


serhiizghama:
Closing — issue #16732 was resolved on 2026-05-31 and PR #17131 (`fix(core): AgentExecutionOptions<undefined> no longer requires structuredOutput`) was merged via commit 0bf2d932 with the same fix. Our PR is fully superseded.

PR #17038, opened 2026-05-25
chore: migrate workspace to pnpm v11

TylerBarnes:
Migrates the workspace to pnpm 11.3.0 and moves pnpm config out of package.json into pnpm-workspace.yaml, which is where pnpm v11 reads these settings.

It also enables the global virtual store for faster git worktree installs:

```yaml
# Share pnpm's virtual store across git worktrees: https://pnpm.io/git-worktrees
enableGlobalVirtualStore: true
```

I also moved the standalone e2e fixture pnpm settings into local pnpm-workspace.yaml files so their overrides and build approvals still apply under pnpm v11.

Verified with pnpm 11.3.0 install --frozen-lockfile and git diff --check.

## ELI5

This PR upgrades the repository to pnpm v11 (11.3.0), moves workspace pnpm settings out of package.json into pnpm-workspace.yaml so pnpm v11 reads them, and enables a shared global virtual store to speed up installs (useful for git worktrees). It also updates tests, fixtures, and CI to use pnpm-specific env vars and workspace-aware installs.

### Core changes
- Bumped required pnpm in root package.json: engines.pnpm → >=11.0.0 and updated packageManager to pnpm@11.3.0 (with new integrity).
- Added/expanded top-level pnpm-workspace.yaml:
  - enableGlobalVirtualStore: true to share pnpm’s virtual store across git worktrees.
  - Reworked patchedDependencies to versioned patch entries (e.g., `@changesets/get-dependents-graph`@2.1.3/2.1.4, tsup@8.5.1).
  - Added packageExtensions (e.g., `@uiw/codemirror-theme-dracula`@*).
  - Extended trustPolicyExclude and expanded overrides to constrain many deps (ssri, qs, zod variants, etc.).
  - Broadened allowBuilds to explicitly disable builds for many packages.

### E2E & fixture workspace changes
- Removed pnpm/resolutions and top-level pnpm blocks from many package.json files and introduced local pnpm-workspace.yaml files in fixture/e2e directories (commonjs, create-mastra, deployers, monorepo, no-bundling, type-check, memory, agent-builder, mcp, playground kitchen-sink, pkg-outputs, etc.).
- Those local pnpm-workspace.yaml files reintroduce necessary overrides (e.g., ssri >=6.0.2, qs ^6.15.1) and allowBuilds settings so fixture-specific build allowances and overrides continue to apply under pnpm v11.

### Tests & tooling updates
- Updated many e2e tests and setup scripts to export pnpm-specific env vars (pnpm_config_registry, pnpm_config_minimum_release_age, etc.) instead of npm_config_* so pnpm receives registry/minimum-release-age configs.
- Extended pnpm-specific env var allowlist (packages/_types-builder) with additional pnpm_config_* keys.
- Small TypeScript typings and build tweaks (tsup externals, various Droppable typing updates).
- Patch update: changesets-get-dependents-graph patch updated to construct Range with includePrerelease: true.

### CI & GitHub Actions
- Composite action explicitly sets up Node 22 via actions/setup-node@v6 and continues to install pnpm via pnpm/action-setup@v4.
- Turbo action now runs turborepo via pnpm dlx (pnpm dlx turbo ...) and disables telemetry.
- Multiple CI steps and workflows changed pnpm install --ignore-workspace → pnpm install to be workspace-aware where appropriate.

### Commit / migration notes
- Added commit(s) to preserve nested pnpm workspace configs for fixtures, add missing packageExtensions, and make other adjustments required by pnpm v11 (including explicit allowBuilds in nested workspaces so frozen installs pass).
- Author verified migration with pnpm 11.3.0 using install --frozen-lockfile and git diff --check, but noted further testing/investigation will follow (primary motivation: enable worktree/global virtual store behavior).

## Impact & reviewer notes
- Requires pnpm >=11.0.0 locally and in CI (declared 11.3.0). Update local tooling if needed.
- Workspace overrides, patched deps, and build allowances were relocated to pnpm-workspace.yaml — reviewers should confirm fixture behavior, package linking, and build approvals still match expectations.
- enableGlobalVirtualStore changes dependency layout/hoisting for worktree installs; test git worktree installs and CI runs.
- Verify e2e fixtures, CI jobs, and any environment-specific registry/settings propagate correctly via pnpm_config_* env vars.

[Review Change Stack](https://app.coderabbit.ai/change-stack/mastra-ai/mastra/pull/17038?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


e2e-tests/commonjs/commonjs.test.ts
```diff
@@ -17,7 +17,7 @@ describe('commonjs', () => {
     fixturePath = await mkdtemp(join(tmpdir(), 'mastra-commonjs-test-'));
     console.log('fixturePath', fixturePath);
 
-    process.env.npm_config_registry = registry;
+    process.env.pnpm_config_registry = registry;
```
wardpeet, line 20: i'm not sure if this actually works, nomrally npm_config_registry still works

e2e-tests/create-mastra/create-mastra.test.ts
```diff
@@ -28,8 +28,8 @@ describe('create mastra', () => {
           stdio: ['inherit', 'inherit', 'inherit'],
           env: {
             ...process.env,
-            npm_config_registry: registry,
-            npm_config_minimum_release_age: '0',
+            pnpm_config_registry: registry,
```
wardpeet, line 31: same with these

wardpeet (commented):
I don't really trust this diff

TylerBarnes:
> I don't really trust this diff

@wardpeet fair, I'll do some more testing and looking into it, really just want to be able to use the worktree linking feature 😄

TylerBarnes:
@wardpeet I followed the migration guide here https://pnpm.io/migration

e2e-tests/commonjs/commonjs.test.ts
TylerBarnes: Reese says: @wardpeet I double-checked this against pnpm 11.3.0 locally, and `npm_config_registry` no longer feeds pnpm config here, while `pnpm_config_registry` does.

Verification:

`env npm_config_registry=https://example.invalid corepack pnpm@11.3.0 config get registry`
returned the default registry.

`env pnpm_config_registry=https://example.invalid corepack pnpm@11.3.0 config get registry`
returned `https://example.invalid/`.

Same result for `minimum_release_age`: the `npm_config_*` env var was ignored, and the `pnpm_config_*` env var was honored. So I think this change is needed for pnpm v11.

Love, Reese

e2e-tests/create-mastra/create-mastra.test.ts
TylerBarnes: Reese says: @wardpeet same reasoning as the other thread. I verified with pnpm 11.3.0 that `npm_config_registry` / `npm_config_minimum_release_age` are ignored as pnpm config inputs, while `pnpm_config_registry` / `pnpm_config_minimum_release_age` are honored.

So I think these should stay as `pnpm_config_*` for the pnpm v11 migration.

Love, Reese

packages/playground/e2e/kitchen-sink/package.json
```diff
@@ -24,16 +24,5 @@
...
-      "@mastra/core": "link:../../../../packages/core",
-      "@mastra/loggers": "link:../../../../packages/loggers",
-      "@mastra/memory": "link:../../../../packages/memory",
-      "@mastra/libsql": "link:../../../../stores/libsql",
-      "@mastra/editor": "link:../../../../packages/editor",
-      "@mastra/mcp": "link:../../../../packages/mcp"
-    }
-  },
-  "packageManager": "pnpm@10.29.3"
+  "packageManager": "pnpm@11.3.0"
```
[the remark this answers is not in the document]
TylerBarnes: Reese says: @coderabbitai this was valid against the earlier commit, but the latest branch has since fixed the pnpm v11 workspace/build fallout.

Validation now passes locally with pnpm 11.3.0:

`corepack pnpm@11.3.0 install --frozen-lockfile`
`corepack pnpm@11.3.0 turbo --filter "!./examples/**/*" --filter "!./docs/" --filter "!./explorations/**/*" build`
`corepack pnpm@11.3.0 build:playground-ui`
`corepack pnpm@11.3.0 --filter ./packages/playground-ui test -- --run --reporter=dot --bail 1`

The filtered build includes the playground package build/typecheck and no longer hits the `@mastra/react` resolution error.

Love, Reese

e2e-tests/commonjs/commonjs.test.ts
wardpeet: checke the migration guide and yeah you're right

e2e-tests/deployers/pnpm-lock.yaml
```diff
@@ -5,7 +5,13 @@ settings:
   excludeLinksFromLockfile: false
 
 overrides:
```
wardpeet: we should not link this defeats the purpose of the verdaccio install which actually tests the packages like a real install. This is necessary to check type errors and compilation problems

wardpeet (changes requested):
overall looks good just the link in the e2e test need to be cleaned up

packages/playground/src/pages/metrics/index.tsx
```diff
@@ -52,7 +52,7 @@ const ANALYTICS_OBSERVABILITY_TYPES = new Set([
   'ObservabilityStorageClickhouseVNext',
   'ObservabilityStorageDuckDB',
   'ObservabilityInMemory',
-  "ObservabilitySpanner",
+  'ObservabilitySpanner',
```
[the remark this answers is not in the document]
TylerBarnes: @coderabbitai fixed in https://github.com/mastra-ai/mastra/pull/17038/commits/bad4600efd35e9c26f6fbb21e24f0a7af5e4a741 — both the empty-state description and the in-memory notice now include Spanner alongside ClickHouse and DuckDB.

Love, Reese

e2e-tests/deployers/pnpm-lock.yaml
TylerBarnes: Reese says: @wardpeet fixed in https://github.com/mastra-ai/mastra/pull/17038/commits/1259194681d2d1e1b7ef6bca348f6b765880cfc7b — removed the entire overrides block from the deployer fixture's pnpm-workspace.yaml and regenerated the lockfile. The fixture no longer has link: overrides, so Verdaccio tests now install published tarballs as intended.

Love, Reese

TylerBarnes: Agreed — the deployers fixture link overrides were already removed in the latest commit (`d617c2cb59`), along with identical patterns in the other three Verdaccio-backed fixtures (no-bundling, type-check, monorepo). The deployer pnpm-workspace.yaml now has only `packages`, `allowBuilds`, and no overrides.

Verified no remaining `link:../../packages` or `link:../../stores` overrides exist in any e2e fixture workspace config, and all three fixture lockfiles pass `pnpm install --frozen-lockfile` with pnpm v11.3.0.

packages/cli/src/commands/create/utils.ts
```diff
@@ -259,6 +259,19 @@ export const createMastraProject = async ({
       );
     }
 
+    // Write pnpm workspace config for pnpm v11
+    if (pm === 'pnpm') {
+      await fs.writeFile(
+        'pnpm-workspace.yaml',
+        `packages:
+  - '.'
+allowBuilds:
```
wardpeet, line 268: does this work for older versions as well?

packages/deployer/src/services/deps.ts
```diff
@@ -91,19 +91,37 @@ export class Deps extends MastraBase {
...
-        cpu: options.cpu || [],
-        libc: options.libc || [],
-      },
-    };
+    const workspaceYamlPath = path.join(dir, 'pnpm-workspace.yaml');
+
+    const lines: string[] = [
+      'packages:',
+      "  - '.'",
+      'allowBuilds:',
```
wardpeet, line 99: i'm guessing there is no allow all?

wardpeet (approved):
some questions but good for me

packages/cli/src/commands/create/utils.ts
TylerBarnes: Reese says: Good call. `allowBuilds` is the pnpm v11 config and it also works in pnpm >=10.26, but pnpm 10.25 does not honor it. I added the older `onlyBuiltDependencies` list alongside `allowBuilds` in the generated create-mastra `pnpm-workspace.yaml`, so generated projects work across pnpm 10.25, 10.26+, and 11.

Verified with fresh installs using:

- `corepack pnpm@10.25.0 install --config.minimum-release-age=0`
- `corepack pnpm@10.26.0 install --config.minimum-release-age=0`
- `corepack pnpm@11.3.0 install --config.minimum-release-age=0`

Also ran `pnpm --filter ./packages/cli typecheck` and `pnpm build:cli`.

Love, Reese

packages/deployer/src/services/deps.ts
TylerBarnes: Reese says: There is an allow-all option: `dangerouslyAllowAllBuilds: true`. I tested that it works with pnpm v11, but I avoided it here intentionally because it allows every current and future transitive dependency install script to run in deploy output.

For the deploy path I think the safer behavior is the explicit list of native/build deps we know can appear in bundled output: `bcrypt`, `esbuild`, `sharp`, `protobufjs`, `workerd`, `bufferutil`, and `utf-8-validate`. That keeps deploy installs working without turning the generated project into an open-ended install-script allow-all.

Love, Reese

Issue #7184, opened 2025-08-29
[DISCORD:1410546986150789150] InferUITool does not work with the new mastra version


wardpeet:
This issue was created from Discord post: https://discord.com/channels/1309558646228779139/1410546986150789150

Here is my mastra version:

```json
{
  "ai": "^5.0.26",
  "@ai-sdk/anthropic": "^2.0.9",
  "@ai-sdk/openai": "latest",
  "@ai-sdk/openai-compatible": "^1.0.13",
  "@ai-sdk/react": "^2.0.26",
  "@mastra/core": "0.15.2",
  "@mastra/loggers": "^0.10.6",
  "@mastra/memory": "0.14.2",
  "@mastra/mongodb": "^0.13.5",
  "mastra": "0.11.2",
}
```

With the `ai-sdk-v5`  version of the mastra sdk this was working:

```ts
import { InferUITools } from "ai";
import { createTool } from "@mastra/core/tools";

const mastraTools = {
  "some-tool": createTool(...)
}

type MastraUITools = InferUITools<typeof mastraTools>
```

Currently this does not work anymore, I got this ts error 

```Types of property 'execute' are incompatible.```

can someone help me?

Also there is a way to get this type from a workflow?

TylerBarnes:
Followed up in discord but I don't see this issue on latest package versions:

my code
<img width="633" height="570" alt="Image" src="https://github.com/user-attachments/assets/852f0b77-1956-4a0a-94d1-1fb6682aedf1" />

inspecting the type:
<img width="742" height="644" alt="Image" src="https://github.com/user-attachments/assets/8626242a-1f99-4939-8af5-e726f15b707f" />

TylerBarnes:
Ok a bit more digging and I do see an issue still:

<img width="392" height="297" alt="Image" src="https://github.com/user-attachments/assets/eadd5a2a-971c-4230-bf99-504a37fabc5c" />

TylerBarnes:
Currently publishing a PR snapshot with a fix! There's a new `import { InferUITools } from "@mastra/core/tools"` you can use.
The snapshot version will be `@mastra/core@infer-ui-tools` once this action is finished https://github.com/mastra-ai/mastra/actions/runs/17473775247

PR #17039, opened 2026-05-25
fix(core): `InferUITools` no longer falls through to `never` for concrete `outputSchema`

### LINKED ISSUE #17036: InferUITools from @mastra/core/tools produces `never` for tools with concrete outputSchema (follow-up to #7184)
## Summary

The `InferUITools` helper added in [#7184](https://github.com/mastra-ai/mastra/issues/7184) (`@mastra/core/tools`) produces `{ input: never; output: never }` for any tool whose `outputSchema` resolves to a concrete type rather than `unknown`. This includes effectively every tool authored via the documented `createTool({ outputSchema: z.object(...) })` pattern.

Verified empirically against `@mastra/core@1.36.0`. The bug is structural — it lives in `InferToolInput<T>`'s `extends Tool<infer I, unknown, unknown, unknown>` constraint, where the `unknown` placeholders in slots 2-4 are exact-match in TypeScript's type system. A concrete `outputSchema` (or any other concretely-typed generic slot) fails the exact-match and falls the whole mapped type to `never`.

The [Mastra v4→v5 migration guide](https://mastra.ai/guides/migrations/ai-sdk-v4-to-v5) explicitly documents this pattern as working — the `calculator` example shows `output: { result: number }`, not `never`. The example happens to satisfy the `unknown` exact-match because of how the generics narrow in that simplified context, but production tools with richer typing don't.

## Minimal repro

```ts
import { createTool, type InferUITools } from "@mastra/core/tools";
import { z } from "zod";

const echoTool = createTool({
  id: "echo",
  description: "echo back the input",
  inputSchema: z.object({ x: z.string() }),
  outputSchema: z.object({ y: z.number() }),
  execute: async ({ x }) => ({ y: x.length }),
});

const tools = { echo: echoTool };

type UITools = InferUITools<typeof tools>;
//   ^? { echo: { input: never; output: never } }
```

Expected (per migration guide): `{ echo: { input: { x: string }; output: { y: number } } }`.

## Source of the bug

`node_modules/@mastra/core/dist/tools/ui-types.d.ts:12`:

```ts
export type InferToolInput<T> = T extends Tool<infer I, unknown, unknown, unknown> ? I : never;
export type InferToolOutput<T> = T extends Tool<unknown, infer O, unknown, unknown> ? O : never;
```

The `unknown` in slots 2-4 is the issue. Any `Tool<I, O, TContext, TSchemaIn>` where `O`, `TContext`, or `TSchemaIn` are anything other than literally `unknown` falls through to `never`. Mastra's own tool types fill those slots concretely (`outputSchema` carries the schema's inferred type, etc.), so the helper degrades for the canonical case.

## Suggested fix

Widen the placeholders to `any` for bivariant inference (matches AI SDK's `InferUITools` approach in `@ai-sdk/provider-utils`):

```ts
export type InferToolInput<T> = T extends Tool<infer I, any, any, any> ? I : never;
export type InferToolOutput<T> = T extends Tool<any, infer O, any, any> ? O : never;
```

`any` in the unused slots is the documented TS pattern for "don't constrain this slot during inference." The intent is to extract slot 1 (or 2) without caring what the others contain — `unknown` is too strict for that.

## Local workaround we ship today

We hand-rolled a `InferLocalUITools` shim with the `any` placeholders. Identical shape, ~25 LOC. Worked around the bug to ship our PR but it's pure boilerplate that should live upstream:

```ts
type InferMastraToolInput<T> = T extends Tool<infer I, any, any, any> ? I : never;
type InferMastraToolOutput<T> = T extends Tool<any, infer O, any, any> ? O : never;
type InferLocalUITools<TOOLS extends Record<string, Tool>> = {
  [NAME in keyof TOOLS & string]: {
    input: InferMastraToolInput<TOOLS[NAME]>;
    output: InferMastraToolOutput<TOOLS[NAME]>;
  };
};
```

Swapping our shim for the upstream `InferUITools` breaks 15+ tool-result React components (every consumer destructuring `.input` / `.output` from a discriminated `tool-<name>` UIMessage part). Reverting to the shim restores typecheck.

## Environment

- `@mastra/core`: `1.36.0`
- `ai`: `6.0.182`
- `zod`: `4.4.3`
- `typescript`: `5.x` (project's `strict: true`)

## Related

- [#7184](https://github.com/mastra-ai/mastra/issues/7184) — original report; closed when `InferUITools` was added to `@mastra/core/tools`. The shipped helper has the bug described here.
- [Mastra v4→v5 migration guide](https://mastra.ai/guides/migrations/ai-sdk-v4-to-v5) — documents `InferUITools` with the `calculator` example that implies it works for concrete output schemas.

Happy to PR the fix if useful — it's a one-character-per-line change in `ui-types.d.ts`.

abshirahmed:
## Why

Fixes #17036.

`InferToolInput` / `InferToolOutput` / `InferUITool` / `InferUITools` from `@mastra/core/tools` fall through to `never` for any tool authored via the canonical `createTool({ outputSchema: z.object(...) })` shape. The bug is structural — the previous definitions exact-matched `unknown` in the trailing generic slots:

```ts
// before
export type InferToolInput<T> = T extends Tool<infer I, unknown, unknown, unknown> ? I : never;
export type InferToolOutput<T> = T extends Tool<unknown, infer O, unknown, unknown> ? O : never;
```

TypeScript treats `unknown` in `extends Tool<infer I, unknown, ...>` as an exact-match constraint. Any tool whose `outputSchema` / context / suspend-resume slot is concretely typed fails the match and the whole conditional resolves to `never`. This closes the gap left by #7184 (the helpers were added but didn't cover the concrete-output case — verified empirically that even the post-#7184 version reproduces the bug).

## What

Replace the `unknown` constraints with `infer _` placeholders. Same bivariant inference as the AI SDK's `Tool<infer INPUT, any>` pattern, but without an `any`-cast or lint disable:

```ts
// after
export type InferToolInput<T> = T extends Tool<infer I, infer _O, infer _C, infer _S> ? I : never;
export type InferToolOutput<T> = T extends Tool<infer _I, infer O, infer _C, infer _S> ? O : never;
```

Concrete-output tools now type correctly end-to-end:

```ts
const echo = createTool({
  id: 'echo',
  inputSchema: z.object({ x: z.string() }),
  outputSchema: z.object({ y: z.number() }),
  execute: async ({ x }) => ({ y: x.length }),
});

type UI = InferUITool<typeof echo>;
// before: { input: never; output: never }
// after:  { input: { x: string }; output: { y: number } }
```

## Test

Added `packages/core/src/tools/ui-types.test-d.ts` — 5 `expectTypeOf` assertions covering both concrete-output and no-outputSchema cases for `InferToolInput`, `InferToolOutput`, `InferUITool`, and `InferUITools`. Verified locally: all 5 pass with the fix; all 5 fail on `main` (confirming the regression).

## Empirical verification against a real consumer

Patched a real Next.js + Mastra agent project (~19 native tools, all with concrete `outputSchema` per the canonical pattern) and swapped its local `InferLocalUITools` workaround shim for the upstream `InferUITools` after this fix. Result: project's `tsc --noEmit` clean, **873 tests passing**, 15+ tool-result React components that consume discriminated `tool-<name>` UIMessage parts continue to narrow correctly. Without the fix, swapping back to upstream breaks every consumer (`Property 'input' does not exist on type 'never'`).

## Risk

Type-only change. No runtime behaviour. Bivariant inference is the canonical TypeScript pattern for "infer one slot, ignore the others" — same approach AI SDK uses in `@ai-sdk/provider-utils`. The fix preserves backwards compatibility with the `unknown`-output case (covered by the new test).

## Linked

- Fixes #17036
- Closes the gap left by #7184 (helper was added but didn't cover concrete-output)

## ELI5

TypeScript was incorrectly saying a tool’s input/output were “never” even when the tool clearly declared them. This change teaches the type helpers how to correctly “read” those declared schemas so the UI/tool types come out right.

## Problem

`InferToolInput`, `InferToolOutput`, `InferUITool`, and `InferUITools` in `@mastra/core/tools` evaluated to `never` for tools authored with a concrete `outputSchema` (for example, `createTool({ outputSchema: z.object(...) })`), breaking downstream type inference for consumers.

## Root Cause

The conditional types used `unknown` in unused generic slots (e.g., `Tool<infer I, unknown, unknown, unknown>`). TypeScript treats those `unknown` positions as exact-match constraints, so when the real tool used concrete types in those slots, the conditional failed to match and fell through to `never`.

## Solution

Replace the `unknown` placeholders with inferred wildcard placeholders so unused generic slots don’t constrain matching:

- `InferToolInput<T>`: `T extends Tool<infer I, infer _O, infer _C, infer _S> ? I : never`
- `InferToolOutput<T>`: `T extends Tool<infer _I, infer O, infer _C, infer _S> ? O : never`

This preserves the existing “unknown output” behavior while correctly inferring concrete output types.

## Changes Made

- `packages/core/src/tools/ui-types.ts`: Updated `InferToolInput` and `InferToolOutput` to use `infer _O/_C/_S` instead of `unknown` (+2/-2).
- `packages/core/src/tools/ui-types.test-d.ts`: Added type-level tests covering concrete-output and no-`outputSchema` scenarios (+49/-0).
- `.changeset/fix-infer-ui-tools-never-for-concrete-output-schemas.md`: Documented the TypeScript inference fix for `@mastra/core` (+20/-0).

## Testing

Added Vitest `.test-d.ts` type-level tests that fail on main and pass with this fix. Also validated via a real Next.js + Mastra agent project (typecheck + test suite passes) after swapping a local shim for the upstream fix.

## Compatibility & Impact

Type-only change with no runtime impact. Fixes the reported inference regression (e.g., Issue `#17036`) for concrete `outputSchema` tools while keeping the prior behavior for the unknown-output case.


mikhael28:
@abshirahmed thank you for submitting this PR - I am running it against our full test suite now.

TheIsrael1 (approved):

PR #17042, opened 2026-05-25
fix: stabilize MastraCode Harness v1 startup and resumes

TylerBarnes:
This fixes a few Harness v1 issues that showed up while trying MastraCode locally. It's still not enough for it to work properly, but it gets it past a few issues.

The shared workspace factory now gets MastraCode runtime state, so startup no longer fails when workspace setup needs `projectPath`:

```ts
workspace: ({ requestContext }) =>
  config.workspace!({
    requestContext: this.withRuntimeHarnessContext(requestContext),
    mastra: this.mastra,
  })
```

The internal v1 Mastra instance also matches legacy Harness logging behavior, so expected storage capability noise does not leak into the TUI.

Yolo mode and MastraCode task tools now bypass v1 approval prompts correctly. `task_write`, `task_update`, `task_complete`, and `task_check` stay allowed even if permission rules are stricter.

`request_access` now uses the native v1 sandbox-access resume path, and resume rebuilds the request context with model, memory, toolsets, yolo state, and workspace context so approving sandbox access continues the original run instead of leaving it idle.

Validated with focused mastracode/core tests and typechecks.

### DIFF
--- a/mastracode/src/harness/runtime.ts
+++ b/mastracode/src/harness/runtime.ts
@@ -19,6 +19,7 @@ import type {
 import { convertStoredMessageToHarnessMessage, defaultDisplayState } from '@mastra/core/harness';
 import {
   Harness as HarnessV1,
+  HarnessSessionLockedError,
   getHarnessWorkspaceActionPathInput,
   isHarnessWorkspaceFileMutationTool,
 } from '@mastra/core/harness/v1';
@@ -48,7 +49,11 @@ import {
   toHarnessV1Subagents,
   toModelInfo,
 } from './config.js';
-import type { MastraCodeModelInfo, MastraCodeRuntimeConfig } from './config.js';
+import type {
+  LeaseRecoveryAction,
+  MastraCodeModelInfo,
+  MastraCodeRuntimeConfig,
+} from './config.js';
 import { MastraCodeHarnessEventProjector } from './events.js';
 import { emptyOMProgress, getOMModelState } from './observational-memory.js';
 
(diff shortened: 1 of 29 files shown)


mastracode/src/harness/runtime.ts
```diff
@@ -1793,13 +1826,15 @@ export class MastraCodeHarnessRuntime<TState extends Record<string, unknown>> {
...
       ...Object.entries(rules.categories ?? {})
         .filter(([category]) => TOOL_CATEGORIES.includes(category as ToolCategory))
         .map(([category, policy]) => session.permissions.setPolicy({ category: category as ToolCategory, policy })),
-      ...Object.entries(rules.tools ?? {}).map(([toolName, policy]) =>
-        session.permissions.setPolicy({ toolName, policy }),
-      ),
+      ...Object.entries(rules.tools ?? {})
+        .filter(([toolName]) => !alwaysAllowTools.has(toolName))
+        .map(([toolName, policy]) => session.permissions.setPolicy({ toolName, policy })),
+      ...ALWAYS_ALLOW_TOOL_NAMES.map(toolName => session.permissions.setPolicy({ toolName, policy: 'allow' })),
```
mbenhamd, line 2149: Can we double-check this path? `syncPermissionRules()` now forces the always-allowed tools to `allow` on `session.permissions`, but `prepareActiveToolsStep -> filterActiveTools() -> isToolDenied()` still seems to read `this.state.permissionRules` directly.

If raw state contains something like `tools.task_write = 'deny'`, `task_write` can still be removed from `activeTools` before the model can call it, even though the session permission policy was corrected to `allow`.

Maybe `isToolDenied()` should also exempt `ALWAYS_ALLOW_TOOL_NAMES`, or active-tool filtering should use the same effective policy as the permission sync path. A regression test with `permissionRules.tools.task_write = 'deny'` and `filterActiveTools(['task_write'])` should catch it.

mbenhamd (commented):
Left two small concerns to double-check around always-allowed tool filtering and per-turn yolo across resume.

packages/core/src/harness/v1/session.ts
```diff
@@ -7185,7 +7187,22 @@ export class Session {
...
     try {
       assertResumedTurnNotDeleted();
+      const mode = this._harness._getMode(resumeModeId);
+      const toolsets = this._buildToolsets(mode, undefined);
+      const requestContext = await Promise.race([
+        this._buildRequestContext({
+          modeId: resumeModeId,
+          modelId: resumeRuntimeDependencies.modelId ?? this._record.modelId,
+          abortSignal: turnAbortController.signal,
+          yolo: ((this._record.state ?? {}) as { yolo?: unknown }).yolo === true,
```
mbenhamd: Can we also check the per-turn `yolo` resume case? The live message/queue paths pass `yolo` into the request context, but I don’t see that value being persisted on `PendingResume`.

On resume, this rebuilds context from `state.yolo === true` only and calls `_shouldRequireToolApproval()` without the original per-turn flag. So a turn started with `{ yolo: true }` can resume under non-yolo approval semantics unless yolo was also set in session state.

Maybe the pending resume record should carry the effective yolo value and feed it back into `_buildRequestContext()` / `_shouldRequireToolApproval()`. A regression test should suspend a `yolo: true` turn, resume it, then verify the next gated tool does not prompt.

abhiaiyer91 (approved):

packages/core/src/harness/v1/session.ts
TylerBarnes: sounds good, feel free to make changes too

PR #15832, opened 2026-04-28
fix(core): isolate subagent tool results from supervisor context

mbenhamd:
Fixes #15823.

## Summary

This keeps subagent nested tool results out of the supervisor model context by default. The supervisor still receives the subagent text for later iterations, while application and UI code can keep reading the raw delegation result, including `subAgentToolResults`, from the tool result payload.

This also adds an explicit escape hatch for workflows that intentionally want the full subagent result in the supervisor model context:

```ts
delegation: {
  includeSubAgentToolResultsInModelContext: true,
}
```

## Why

Subagent tool results can contain large intermediate payloads, nested tool arguments, and implementation details that are useful to application code but usually harmful as implicit supervisor prompt context. Sending only the subagent text by default keeps the next supervisor model call focused and avoids accidental context bloat.

The opt-in keeps compatibility for prompt-organization workflows that depend on the supervisor model seeing the full raw subagent result.

## Changes

- Adds text-only `toModelOutput` for subagent delegation tools by default.
- Adds `delegation.includeSubAgentToolResultsInModelContext` to restore full raw subagent result context when needed.
- Preserves raw nested tool results in `result.toolResults` for application and UI consumers.
- Adds regression coverage for both default isolation and opt-in behavior with output processors.
- Updates supervisor-agent docs and the `@mastra/core` changeset.

## Validation

- `pnpm --filter ./packages/core exec vitest run src/agent/__tests__/supervisor-integration.test.ts src/harness/subagent-tool.test.ts src/harness/display-state.test.ts src/agent/message-list/tests/message-list-v5.test.ts`
- `pnpm --filter ./packages/core check`
- `pnpm --filter ./packages/core lint`
- `pnpm --dir docs validate`
- `pnpm --dir docs exec remark --no-stdout --frail --quiet --ext mdx src/content/en/docs/agents/supervisor-agents.mdx`
- `pnpm --dir docs exec prettier --check src/content/en/docs/agents/supervisor-agents.mdx`

## ELI5 (Explain Like I'm 5)

Subagents used to pour all their internal steps into the supervisor's brain and make it cluttered. This PR keeps only the subagent's final text in the supervisor's context by default, while still saving the full raw subagent result separately for apps/UIs; you can opt-in to include the full raw result in the supervisor context if needed.

---

## Summary of Changes

Fixes #15823 by preventing nested subagent tool results from being merged into the supervisor model context by default while preserving the raw delegation payload for application/UI consumption.

### Key changes
- Code
[...]

PR #17066, opened 2026-05-26
fix(core,ci): isolate legacy sub-agent tool output

damien-schneider:
## Summary
- bridge legacy v1 tool `toModelOutput` into the AI SDK v4 `experimental_toToolResultContent` path so supervisor model input stays text-only by default
- keep the raw sub-agent tool result shape available to callers, including `subAgentThreadId`, `subAgentResourceId`, and `subAgentToolResults`
- update the focused core E2E test and cassette so exact replay verifies both sides of that contract
- include root `package.json` in PR Triage sparse checkouts so `setup-pnpm-node` can read the pinned `packageManager` version

## Why
#15825 intentionally made the v1 raw sub-agent result richer, but the legacy v1 path was not applying the existing model-output projection before sending the tool result back to the supervisor model. This restores the split introduced by #15832: application-visible tool results can include sub-agent metadata, while model-facing tool output remains text-only unless a tool opts into something else.

The PR Triage failures seen on #17062 are separate from the core E2E failure. They happen before tests because the sparse checkout omitted root `package.json`, so `pnpm/action-setup@v4` could not infer the pnpm version. Adding `package.json` to those sparse checkouts keeps the existing shared setup action behavior and does not change the automation scripts.

## Checks
- `pnpm prettier --check .github/workflows/pr-triage.yml`
- `git diff --check`
- `node -p "require('./package.json').packageManager"`
- `CI=true LLM_TEST_MODE=replay pnpm vitest run packages/core/src/agent/__tests__/tool-handling.e2e.test.ts --project e2e:packages/core`
- previously on this branch: `CI=true pnpm vitest run packages/core/src/agent/__tests__/tool-handling.e2e.test.ts --project e2e:packages/core`
- previously on this branch: `pnpm --filter ./packages/core check`
- previously on this branch: `pnpm --filter ./packages/core lint`
- previously on this branch: `pnpm prettier --check . '!docs/**/*' '!examples/**/*' '!explorations/**/*'`
- previously on this branch: `pnpm test:core`

CodeRabbit CLI local review could not run because this machine is not authenticated for CodeRabbit (`coderabbit auth login` required).

## ELI5 Explanation

Think of this PR as a mailroom problem: when a sub-agent sends back a response with extra details (metadata), we want those details delivered to the person who asked for it, but NOT passed along in the next message to the supervisor. This PR fixes that by cleaning up the supervisor's message—removing the extra details so it only sees the essential text—while still preserving all those details for the original caller.

---

## Overview

This PR isolates how legacy AI SDK v1 tools handle sub-agent results to prevent metadata leakage. It separates the **raw tool result** (with all sub-agent metadata visible to callers) from the **model-facing output** (sanitized text-only content for the supervisor), ensuring backward compatibility with v1 tool consumers while maintaining clean supervisor context.

### Core Implementation (packages/core/src/agent/agent-legacy.ts)
Added legacy tool result-content normalization to handle the `experimental_toToolResultContent` hook:
- Introduced helper functions to stringify legacy content and normalize model outputs into the standard `{ type: 'text' | 'image', ... }[]` format
- Added image/media handling and safe fallbacks for unexpected structures
- Updated the legacy agent tool conversion flow to automatically wrap converted tools with `applyLegacyToolModelOutput(...)` at two key points:
  1. After the initial `convertTools(...)` call
  2. After step 0 of `__runProcessInputStep(...)` returns updated tools

This ensures v1 tools receive and expose raw results (including sub-agent metadata like `subAgentThreadId`, `subAgentResourceId`, `subAgentToolResults`, and `usage`) while the supervisor model only sees the model-facing output with metadata stripped.

### Test Updates (packages/core/src/agent/__tests__/tool-handling.e2e.test.ts)
- Initialized the gateway mock with `exactMatch: true` for stricter test matching
- Enhanced the "agents as tools" assertion to explicitly verify that agent tool results contain the expected sub-agent identifiers and metadata fields
- Uses version-dependent paths to extract and assert the full agent tool result object

### Test Cassette Refresh (packages/core/__recordings__/core-src-agent-__tests__-tool-handling.e2e.json)
- Updated the recorded tool-call output to convert a single-object response form (`{"text":"Dummy response"}`) to the AI SDK v4 tool-result content array form (`[{"type":"text","text":"Dummy response"}]`)
- This reflects the application of `toModelOutput` and ensures sub-agent metadata no longer leaks into supervisor model requests
- Updated hash values and reformatted parameter arrays for consistency

### Build/Workflow Updates
- Updated `.github/workflows/pr-triage.yml` sparse-checkout lists in four jobs (`issue-link-nudge`, `complexity`, `summarize`, `close-stale-unlinked`) to include `package.json`
- Added a Changesets entry for `@mastra/core` documenting the patch-level fix

## Context
This PR addresses a regression introduced in PR `#15825`, which enriched v1 raw sub-agent result fields for caller visibility. However, those fields were inadvertently leaked into the supervisor model's next request because the legacy v1 path did not apply `toModelOutput`. This fix applies the necessary output transformation to maintain the surface separation: callers continue to receive full metadata while the model only sees sanitized, model-facing content.

[Review Change Stack](https://app.coderabbit.ai/change-stack/mastra-ai/mastra/pull/17066?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/.changeset/sweet-doodles-clap.md
+++ b/.changeset/sweet-doodles-clap.md
@@ -1,0 +1,5 @@
+---
+'@mastra/core': patch
+---
+
+Fix legacy agent tools so model-facing sub-agent results stay text-only by default while raw metadata remains available.
--- a/.github/workflows/pr-triage.yml
+++ b/.github/workflows/pr-triage.yml
@@ -35,6 +35,7 @@ jobs:
           persist-credentials: false
           sparse-checkout: |
             .github/actions
+            package.json
             scripts/pr-issue-link
           sparse-checkout-cone-mode: false
 
@@ -81,6 +82,7 @@ jobs:
           persist-credentials: false
           sparse-checkout: |
             .github/actions
+            package.json
             scripts/pr-complexity
           sparse-checkout-cone-mode: false
 
@@ -126,6 +128,7 @@ jobs:
           persist-credentials: false
           sparse-checkout: |
             .github/actions
+            package.json
             scripts/pr-automation-comment
           sparse-checkout-cone-mode: false
 
@@ -168,6 +171,7 @@ jobs:
           persist-credentials: false
           sparse-checkout: |
             .github/actions
+            package.json
             scripts/pr-issue-link
           sparse-checkout-cone-mode: false
 
--- a/packages/core/__recordings__/core-src-agent-__tests__-tool-handling.e2e.json
+++ b/packages/core/__recordings__/core-src-agent-__tests__-tool-handling.e2e.json
@@ -208,7 +208,7 @@
       }
     },
     {
-      "hash": "e805b7bf457a5dc0",
+      "hash": "ee7f80b59dfaedef",
       "request": {
         "url": "https://api.openai.com/v1/chat/completions",
         "method": "POST",
@@ -241,7 +241,7 @@
             {
               "role": "tool",
               "tool_call_id": "call_PT0d286WNzsYZW8lnwi268G7",
-              "content": "{\"text\":\"Dummy response\"}"
+              "content": "[{\"type\":\"text\",\"text\":\"Dummy response\"}]"
             }
           ],
           "tools": [
@@ -539,7 +539,14 @@
                   "resumeData"
                 ],
                 "additionalProperties": false,
-                "x-optional": ["threadId", "resourceId", "instructions", "maxSteps", "suspendedToolRunId", "resumeData"]
+                "x-optional": [
+                  "threadId",
+                  "resourceId",
+                  "instructions",
+                  "maxSteps",
+                  "suspendedToolRunId",
+                  "resumeData"
+                ]
               },
               "strict": false
             }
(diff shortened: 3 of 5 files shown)


damien-schneider:
CI note: the current red checks on this PR are failing before repository code runs. The logs show GitHub Actions checkout/action-download failures such as `remote: Your account is suspended` / HTTP 403 and failed action archive downloads. GitHub Status currently has an unresolved Actions/Pages incident for authentication failures when starting Actions runs and downloading actions.\n\nI reran the failed workflows once and they hit the same setup-layer failures. Local verification for the code path is listed in the PR body. Also note that the PR Triage workflow change is in a `pull_request_target` workflow, so this PR's own PR Triage run still uses the current base-branch workflow until the fix is merged to `main`; after that, new PR Triage runs on other PRs will include root `package.json` in the sparse checkout.

Issue #9868, opened 2025-11-07
Error: Failed to download asset


daneatmastra:
This issue was created from Discord post 1436055586151071824:

[![Open in Discord](https://img.shields.io/badge/Open_in_Discord-5865F2?style=for-the-badge&logo=discord&logoColor=white)](https://discord.com/channels/1309558646228779139/1436055586151071824)

Possibly linked to this? https://github.com/mastra-ai/mastra/pull/9790

Error message:

```
Error executing model gemini-2.5-pro, attempt 1==== MastraError: Failed to download asset
    at downloadFromUrl (file:///var/task/mastra.mjs:35469:11)
    at process.processTicksAndRejections (node:internal/process/task_queues:105:5)
    at async file:///var/task/mastra.mjs:62378:20 {
  id: 'DOWNLOAD_ASSETS_FAILED',
  domain: 'LLM',
  category: 'USER',
  details: {},
  [cause]: Error: Request failed with status: 400 Bad Request
      at fetchWithRetry (file:///var/task/mastra.mjs:27313:15)
      at process.processTicksAndRejections (node:internal/process/task_queues:105:5)
      at async downloadFromUrl (file:///var/task/mastra.mjs:35449:22)
      at async file:///var/task/mastra.mjs:62378:20
}
```

abhiaiyer91:
Answered in discord!

PR #9209, opened 2025-10-24
Fix: Don't download unsupported media

BCook98:
## Description

`isUrlSupportedByModel` is being calculated, however it is not being used.  

If media is not supported, it is still being downloaded and resulting in errors.

Also fixes potential data corruption issues due to index misalignment.

## Type of Change

- [x] Bug fix (non-breaking change that fixes an issue)
- [ ] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [ ] Test update

## Checklist

- [ ] I have made corresponding changes to the documentation (if applicable)
- [ ] I have added tests that prove my fix is effective or that my feature works

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed an issue where unsupported media types were being unnecessarily downloaded.

PR #17069, opened 2026-05-26
fix(core): include failing URL in DOWNLOAD_ASSETS_FAILED error from downloadFromUrl

rares04:
## Description

When \`downloadFromUrl\` (in \`packages/core/src/agent/message-list/prompt/download-assets.ts\`) fails — either on a non-OK HTTP response or on a fetch rejection — the resulting \`MastraError\` discards the failing URL:

- \`text\` is the fixed string \`'Failed to download asset'\`
- \`details\` is \`{}\` (defaults)
- the inner \`cause\` from \`fetchWithRetry\` only includes the HTTP status (\`Request failed with status: <status> <statusText>\`)

So the URL the SDK was trying to fetch is **nowhere in the error chain**. This makes it effectively impossible to:
- identify which URL failed when a user message has multiple image/file parts
- drive downstream recovery (e.g. an input processor that replaces the dead URL with a text placeholder on retry, so the next attempt doesn't replay the same poisoned URL against every fallback model — exhausting all of them with the same payload)
- group metrics by host / asset / customer

This change surfaces the URL in two complementary places, matching the convention used elsewhere in the codebase for \`MastraError\` (e.g. \`packages/core/src/evals/base.ts\` uses \`details\` for structured context alongside a human-readable \`text\`):

\`\`\`ts
throw new MastraError({
  id: 'DOWNLOAD_ASSETS_FAILED',
  text: \`Failed to download asset: \${urlText}\`,
  domain: ErrorDomain.LLM,
  category: ErrorCategory.USER,
  details: { url: urlText },
});
\`\`\`

\`text\` keeps any existing string-based log line / regex extractor working; \`details.url\` gives structured access without re-parsing. Both error throw sites (non-OK response, caught fetch rejection) are updated identically. The original error is still preserved on \`cause\` so HTTP status and network error code are still available.

## Why this matters in practice

We hit this in production today on \`@mastra/core@1.35.0\`. An agent received an email with an attached image whose CDN-signed URL had expired. Mastra correctly threw \`DOWNLOAD_ASSETS_FAILED\`, our retry policy correctly classified it as transient, but the retry replayed the exact same poisoned URL because the URL wasn't in the error message for us to mark as dead. Three retries × three fallback models = nine identical 403s and a hard fail back to the user, even though the right behavior was \"drop this part, keep the rest, continue\". Same root cause described in #9868.

## Related issue(s)

Related to #9868 (closed; reported the same \`Failed to download asset\` symptom with empty \`details: {}\`). This PR makes the error actionable for downstream consumers without changing any of the existing retry / supported-URL routing logic.

Adjacent fixes in this area that this PR builds on top of: #9209 (don't download unsupported media), #9790 (respect provider URL support), #10122 / #10079 (download assets gating). None of those exposed the URL in the error.

## Type of change

- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [ ] Test update

## Checklist

- [x] I have made corresponding changes to the documentation (n/a — no public API surface change; \`MastraErrorDefinition.details\` already accepts arbitrary records)
- [x] I have added tests that prove my fix is effective — two new cases in \`download-assets.test.ts\` cover both the non-OK response path and the fetch rejection path, asserting \`error.message\` includes the URL, \`error.details.url\` is set, and the original cause is preserved
- [ ] I have addressed all Coderabbit comments on this PR (will do on review)

## Compatibility

- **No public API change.** \`MastraErrorDefinition\` already declares \`details?: Record<string, Json<Scalar>>\`, and \`text\` is already a string. Consumers that read \`error.message\` literally as \`'Failed to download asset'\` will need to switch to a contains-check, but the previous string was already lossy for any real diagnostic use.
- **\`category: 'USER'\` is preserved.** The semantics (\"the user gave us a URL we can't reach\") are unchanged; the error now just carries the URL needed to act on that classification.
- **No changes to \`downloadAssetsFromMessages\` flow** — the supported-URL fast path, concurrency, retry count, and \`p-map\` orchestration are all untouched.

## Changeset

\`@mastra/core\` — patch.

## ELI5

When the app fails to download a file, the error now tells you exactly which web address failed (without showing secret query tokens), so you can find and fix the broken file more easily.

## Overview

This PR improves error reporting in downloadFromUrl so download failures include the failing asset URL in the thrown MastraError. Human-readable error text is redacted to origin+pathname (query string and fragment removed) to avoid leaking signed-token secrets to logs, while the full original URL (including query and fragment) is preserved in error.details.url for programmatic recovery and diagnostics. The original fetch cause is preserved on the error for debugging.

## Changes

- packages/core/src/agent/message-list/prompt/download-assets.ts
  - Added redactUrlForLog(url: URL) that returns origin+pathname.
  - On non-OK HTTP responses and fetch rejections, throw MastraError with:
    - text: `Failed to download asset: <origin><pathname>` (query + fragment stripped)
    - details: { url: "<full original URL>" } (full URL retained)
    - original fetch error preserved as cause
  - No changes to public APIs, error category (USER), or download concurrency/retry logic.

- packages/core/src/agent/message-list/prompt/download-assets.test.ts
  - Added tests asserting:
    - Non-OK response and fetch rejection both produce a MastraError with id DOWNLOAD_ASSETS_FAILED, message containing the redacted URL, details.url set to the full URL, and the fetch cause preserved.
    - A signed URL test verifies message redacts query+fragment while details.url preserves the full signed URL.
    - Existing behavior for retrying server errors remains validated.

- .changeset/include-url-in-download-assets-error.md
  - Changeset updated to describe the user-visible outcome (error includes failing URL; message redacts query/fragment; details keep full URL).

## Impact

- Improves diagnosability and enables programmatic recovery (e.g., tombstoning failing message parts) without leaking sensitive query tokens into logs.
- Backwards compatible: no public type or API changes and retry/concurrency behavior unchanged.

[Review Change Stack](https://app.coderabbit.ai/change-stack/mastra-ai/mastra/pull/17069?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


rares04:
Thanks for the review @coderabbitai. Pushed [82453b0](https://github.com/mastra-ai/mastra/pull/17069/commits/82453b07e3) addressing both comments:

**1. URL sanitization** — Scanned the codebase first. There's no existing `sanitizeUrl` / `redactUrl` helper, and other URL-in-error sites (e.g. `attachmentsToParts`) currently pass URLs through verbatim. The closest precedents are:
- `SENSITIVE_KEYS` in `packages/core/src/tools/validation.ts` — redacts by key name in error context, applied at logging time
- `redactHeaders` in server config — masks named HTTP headers, applied at logging time

Both follow "redact at the log boundary, keep raw in storage." Applied the same model here:
- **`error.message` / `text`** is now `Failed to download asset: <origin><pathname>` — query string and fragment stripped, so `X-Amz-Signature`, WhatsApp media tokens, GCS `X-Goog-Signature` etc. don't leak to any logger that captures `error.message`.
- **`error.details.url`** keeps the **full URL** — this is the structured field that programmatic consumers (e.g. an input processor matching the failing URL back to a specific message part for recovery) need exact equality on. If we sanitized this too, the very recovery use case that motivated the PR would break: a sanitized URL no longer equals the original `part.image` / `part.data` in the message, so callers can't reliably tombstone it.

Added a dedicated test using a representative signed URL with `X-Amz-Signature`, `X-Amz-Expires`, `token`, and a `#fragment` that asserts both surfaces.

If the project would prefer details.url scrubbed too, happy to switch — but flagging that it would close off the main downstream recovery path that motivated this change. An alternative would be `details.urlPath` (redacted) + `details.url` (raw) as two separate fields, but that feels like over-engineering for what `details.url` already does for a careful consumer.

**2. Changeset wording** — Rewrote to a single user-outcome sentence with a brief note about the redaction split, dropping the implementation prose.

abhiaiyer91 (approved):

PR #17082, opened 2026-05-26
fix(core,mastracode): managed browser ownership so /browser propagation survives

mbenhamd:
## Summary

Closes the browser-ownership half of [MASTRA-4340](https://linear.app/mastra/issue/MASTRA-4340). Stacked on #17068.

- Adds `Agent.__setManagedBrowser()` to `@mastra/core` — an internal setter that updates the agent's browser without claiming explicit ownership (parallel to `Agent.__setWorkspace()` and `#setBrowserFromWorkspace`). A separate `#hasManagedBrowser` flag makes the controller-set browser sticky against workspace reconciliation in `getWorkspace()` and `#execute`.
- Updates `Harness.setBrowser()` and `Harness.propagateRuntimeServicesToAgent()` to route through the managed setter and dedupe per call, so harness-level browser updates stay consistent when one agent is shared across modes.
- Rewires `mastracode`'s `/browser` slash command and `MastraCodeHarnessRuntime.setBrowser()` to use the new managed path, eliminating the self-poisoning loop where `setBrowser()` flipped `hasOwnBrowser()` to true and blocked subsequent global propagations.

End-user outcome: `/browser on`, `/browser off`, and the interactive setup flow now preserve agents that were configured with their own browser, dedupe across modes, and keep working across repeated `/browser on → off → on` cycles. Runtime-level browser updates inherit the same stability.

## Review history

Three rounds of `codex exec --sandbox read-only`, one Claude subagent review, and three rounds of `coderabbit review`. The reviewers found (and this PR addresses):

- The first round's mastracode-local WeakSet approach was conceptually leaky (codex pass 3 + Claude). Pivoted to a core-side method symmetric with `__setWorkspace`.
- The pivot's first cut left a lifecycle bug: `__setManagedBrowser` updated `#browser` without flipping any flag, so the next `getWorkspace()` could clear or overwrite the controller-set browser. Fixed with `#hasManagedBrowser` checked in `#setBrowserFromWorkspace`, both no-workspace branches in `getWorkspace()`, and the workspace-fallback site in `#execute`.
- CodeRabbit caught two follow-ups: the truthy guard in `propagateRuntimeServicesToAgent` dropped `setBrowser(undefined)` clears, and `#execute`'s workspace fallback could revive a managed `undefined`. Both fixed.

## Test plan

- [ ] CI green on `Lint`, `Build`, `UNIT Test Shards 1/3/4` (Shard 2 mcp test is a known pre-existing flake)
- [ ] Reviewer can verify locally:
  - `pnpm --filter @mastra/core exec vitest run src/agent/__tests__/browser.test.ts` — 14 pass
  - `pnpm --filter @mastra/core exec vitest run src/harness/` — 1291 pass
  - `pnpm --filter mastracode exec vitest run src/tui/commands/__tests__/browser.test.ts src/harness/runtime.test.ts` — 54 pass
  - `pnpm --filter mastracode check && pnpm --filter mastracode lint` — clean

### DIFF
--- a/mastracode/src/harness/lease-recovery-prompt.ts
+++ b/mastracode/src/harness/lease-recovery-prompt.ts
@@ -1,0 +1,62 @@
+import { createInterface } from 'node:readline';
+
+import type { LeaseRecoveryAction, LeaseRecoveryPromptHandler, LeaseRecoveryPromptInfo } from './config.js';
+
+/**
+ * Stdio-driven prompt invoked during `MastraCodeHarnessRuntime.init()` when a
+ * stale Harness v1 session lease is held by another process. The handler
+ * resolves to `wait` immediately when stdin is not a TTY so headless callers
+ * stay non-interactive without an explicit `MASTRACODE_LEASE_RECOVERY` value.
+ */
+export function defaultStdioLeaseRecoveryPrompt(): LeaseRecoveryPromptHandler {
+  return async (info: LeaseRecoveryPromptInfo): Promise<LeaseRecoveryAction> => {
+    // Re-check TTY state at invocation time so tests / embedders that flip
+    // `process.stdin.isTTY` after construction get the right behavior.
+    if (!process.stdin.isTTY) return 'wait';
+    const expires =
+      Number.isFinite(info.expiresAt) && info.expiresAt > 0
+        ? new Date(info.expiresAt).toLocaleTimeString()
+        : 'an unknown time';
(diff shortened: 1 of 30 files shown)


mastracode/src/harness/lease-recovery-prompt.ts
wardpeet, line 3: If i undestand correctly this is to make sure only one instance is alive at once? This shouldn't be necessary when multiple consumers can listen to a thread now using signals

Issue #17065, opened 2026-05-26
[FEATURE] Cancel a running agent execution from a Channel (e.g. Slack "stop" reply)


rodrigojranastacio:
## Summary

There's currently no first-class way to **interrupt an in-progress agent execution that was started via a Channel adapter** (Slack, Discord, etc.) from the same channel. Once a user mentions a bot and the agent starts working (specially with long-running tool calls), the user has no way to say "stop, never mind" — they have to wait for the run to finish (or fail).

## Context

Today the building blocks exist, but none of them are wired up to the Channel layer:

- `agent.generate(..., { abortSignal })` and `agent.stream(..., { abortSignal })` accept an `AbortSignal` and propagate it to tools.
- `workflow.run.cancel()` cancels a workflow run by ID.
- Channels (Slack, etc.) handle the inbound webhook and call into the agent, but **the adapter does not expose the running execution's `AbortController` (or a `runId`) anywhere the user-land code can reach it** to abort it later from a follow-up message in the same thread.

Related open / closed issues that touch the same surface but don't solve this:

- #11579 — Resumable Agent Streams (lifecycle of streams via Channels is a known gap)
- #14148 — Thread resume fails after interrupted streams
- #11063 — AbortSignal not propagated to sub-workflows
- #10874 — Propagate AbortSignal through Agent Networks
- #13646 — `@mastra/mcp` does not forward AbortSignal (fixed)

## Use case

We run ~25 Mastra agents on Slack at Aprova Digital (production, internal support + sales + engineering). Common scenario:

> User mentions the agent with the wrong request, or realizes mid-execution that the agent is going down the wrong path (e.g. running a long search/RAG pipeline, or about to write to an external system). Right now the only escape hatch is waiting it out or restarting the process.

## Proposed API (sketch — open to other shapes)

Two ergonomic options that would fit the current Channels design:

### Option A — keyword/reaction-driven cancel handled by the adapter

```ts
slackChannel({
  cancel: {
    keywords: ["stop", "cancel", "para"],   // text in same thread cancels
    reactions: ["x", "no_entry"],            // reactions cancel
  },
})
```

The adapter would internally track `threadId → AbortController` for runs it started and call `.abort()` when a cancel signal arrives. On abort, post a configurable message (e.g. "Stopped by user").

### Option B — expose the run handle and a `cancelByThread` helper

```ts
// inside onMention / onMessage handlers
const run = await thread.runAgent(agent, { ... });   // returns { runId, abort }

[...]

PR #17089, opened 2026-05-26
feat(core): expose abort signal + abort() to custom channel handlers via ctx arg

CalebBarnes:
## Summary

Custom channel handlers (`onDirectMessage`, `onMention`, `onSubscribedMessage`) now receive a 4th `ctx` argument that gives user code first-class access to the active agent run on the current thread.

The `ctx` object exposes:

- `threadId` / `resourceId` — the Mastra thread + resource ids resolved for this chat thread (memory & history scope)
- `abortSignal: AbortSignal` — fires when the active run on this thread is aborted from anywhere (this handler, another handler on the same thread, `agent.abortRunStream()`, or `channels.close()`)
- `abort(reason?): boolean` — abort the currently active run on this thread; returns `true` if a run was aborted
- `activeRunId(): string | undefined` — the active runId on this thread, or `undefined` if idle

This unblocks the user request in #17065 — a `'stop'` command handler is now this short:

```ts
channels: {
  handlers: {
    onDirectMessage: async (thread, message, defaultHandler, ctx) => {
      if (message.text.trim().toLowerCase() === 'stop') {
        if (ctx.abort()) await thread.post('⏹️ stopped')
        return
      }
      await defaultHandler(thread, message)
    },
  },
}
```

`ctx` is also a natural place to pass `abortSignal` into any async work a handler kicks off (`fetch`, model calls, etc.) so the work stops cleanly when the run is canceled.

## How it works

- `AgentChannels` resolves the Mastra thread + ensures the thread subscription **before** handler dispatch (hoisted into a new `resolveHandlerContext()` private method), so the handler and `defaultHandler` see the same `ctx`.
- A per-invocation `AbortController` is registered in `handlerAbortControllers: Map<mastraThreadId, Set<AbortController>>`.
- The stream wrapper in `ensureThreadSubscription` taps `chunk.type === 'abort'` chunks before drivers consume them, and fans out abort to every in-flight handler controller on that thread.
- `close()` aborts every in-flight handler controller.
- A `try/finally` around handler dispatch releases its controller (success or failure) — no leaks.

## Backward compatibility

The 4th `ctx` arg is optional from the caller's TypeScript perspective. Existing 3-arg handlers (`(thread, message, defaultHandler) => …`) compile and run unchanged. A regression test covers this.

## Tests

7 new tests under `describe('handler ctx')` in `packages/core/src/channels/__tests__/agent-channels.test.ts`:

1. `ctx` is passed to custom `onDirectMessage` with the right shape
2. `ctx.abort()` forwards to `AgentThreadSubscription.abort()` when a run is active
3. `ctx.abort()` returns `false` when no run is active
4. `ctx.abortSignal` fires when the underlying run aborts via `agent-run-aborted` fan-out
5. `ctx.abortSignal` does NOT fire across threads (per-thread isolation)
6. `release()` cleans up the controller from `handlerAbortControllers`
7. `close()` aborts all in-flight handler controllers
8. Compile-time + runtime check that 3-arg handler signatures still match `ChannelHandler`

All 141 channels tests pass.

## Test plan

- [x] `pnpm --filter @mastra/core typecheck`
- [x] `pnpm --filter @mastra/core lint`
- [x] `pnpm --filter @mastra/core test src/channels/__tests__/` — **141/141 passing**
- [x] `pnpm build:core`

## Docs + changeset

- Updated `docs/src/content/en/reference/agents/channels.mdx` with a "Stopping the active run" section and the full `ChannelHandlerContext` type signature.
- Added `.changeset/channel-handler-ctx-abort.md` (minor bump for `@mastra/core`).

Refs #17065


suzunn:
I noticed a small API consistency issue while reading through the diff: the PR body and docs describe `activeRunId()` as returning `undefined` when idle, but the exported `ChannelHandlerContext` type currently says `() => string | null`.

I would line those up before merge, ideally by matching the underlying subscription return type and documenting the same sentinel everywhere. Otherwise handler authors may end up writing defensive checks for both `null` and `undefined`, and this becomes harder to tighten later because it is part of the public handler context.

PR #17095, opened 2026-05-26
docs: clarify example dev command

DanielSLew:
This updates AGENTS.md to tell agents to run examples with `pnpm mastra dev` instead of `mastra dev`, so local example runs use the linked workspace version rather than any globally installed CLI.

## ELI5 (Explain Like I'm 5)

This PR updates the project's instructions to tell people how to run example projects the right way. Instead of typing `mastra dev` (which might accidentally use an old copy installed on their computer), it tells them to use `pnpm mastra dev` (which always uses the newest version from the project). This makes sure everyone is testing with the same, correct version.

## Changes Made

**Examples Documentation (`examples/AGENTS.md`)**
- Updated example running instructions to specify `pnpm mastra dev` instead of `mastra dev` to ensure the workspace version of the CLI is used

**CLI Init Utils (`packages/cli/src/commands/init/utils.ts`)**
- Enhanced the `generateAgentsMarkdown` function to include a new guidance bullet instructing users to use package scripts (e.g., `npm run dev`) instead of running `mastra dev` directly

**Changeset (`changeset/dull-peaches-pump.md`)**
- Created a patch-level changeset documenting that generated `AGENTS.md` guidance now prefers package scripts over global Mastra commands

## Release Impact

This is a documentation-only change marked as a patch release. No code logic changes are included, and the updates apply both to static example documentation and the dynamically generated `AGENTS.md` scaffold that guides new project setup.

[Review Change Stack](https://app.coderabbit.ai/change-stack/mastra-ai/mastra/pull/17095?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


junydania (approved):

AGENTS.md
```diff
@@ -5,6 +5,7 @@ For work in packages read package local packages/<name>/AGENTS.md first
 turborepo pnpm workspace
 packages use strict TypeScript
 vitest tests are colocated with source
+When running examples locally, use `pnpm mastra dev` instead of `mastra dev` so the linked workspace version is used instead of any global install.
```
TylerBarnes: maybe we should add this into a new examples/AGENTS.md? since they autoload based on if the agent is working in that dir

TylerBarnes (commented):
one thought ☝️

AGENTS.md
TylerBarnes: ah yeah looks like we'll have to to keep in our token budget for this file 😆 
image

we could also add this into an AGENTS.md in scaffolded projects with create-mastra

PR #17125, opened 2026-05-27
Run memory tests on integration manifest changes

wardpeet:
## Summary
- Treat packages/memory/integration-tests/package.json changes as memory test-triggering changes.
- Preserve the existing turbo-changed package/dependency hash check while adding an explicit standalone manifest check.

## Test plan
- Parsed .github/workflows/prebuild.yml with Ruby YAML loader.
- Verified the branch diff only changes .github/workflows/prebuild.yml.

## ELI5 (Explain Like I'm 5)

This PR makes memory tests run whenever either the memory package code changes or the integration-tests package.json file is changed, so configuration updates won't be missed.

## Summary of changes

Modified .github/workflows/prebuild.yml to expand the trigger conditions for memory tests in the memory-check-changes job.

Key changes:
- Updated the job-level output so memory-check-changes.outputs.changed is true when either the existing Turborepo/dorny paths change detection (steps.changes.outputs.changed) is true OR when a new explicit manifest check detects changes (steps.integration-test-manifest.outputs.changed).
- Added a new step integration-test-manifest that runs a git diff between the PR base SHA and head SHA to set changed=true if packages/memory/integration-tests/package.json appears in the diff.

Result:
Memory tests will now trigger when either:
1. The `@mastra/memory` package or its dependencies change (existing Turborepo logic), or
2. The integration test manifest file (packages/memory/integration-tests/package.json) is modified.

This ensures changes to integration test configuration are detected and cause the memory test suite to run.

[Review Change Stack](https://app.coderabbit.ai/change-stack/mastra-ai/mastra/pull/17125?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/.github/workflows/prebuild.yml
+++ b/.github/workflows/prebuild.yml
@@ -198,7 +198,7 @@ jobs:
     runs-on: starsling-ubuntu-24.04
     timeout-minutes: 10
     outputs:
-      changed: ${{ steps.changes.outputs.changed }}
+      changed: ${{ steps.changes.outputs.changed == 'true' || steps.integration-test-manifest.outputs.changed == 'true' }}
     permissions:
       contents: read
     steps:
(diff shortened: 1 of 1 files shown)


.github/workflows/prebuild.yml
```diff
@@ -219,6 +219,15 @@ jobs:
           tasks: 'build,test'
           base-ref: 'origin/main'
 
+      - name: Check for memory integration test manifest changes
+        id: integration-test-manifest
+        run: |
+          if git diff --name-only origin/main...HEAD | grep -qx 'packages/memory/integration-tests/package.json'; then
```
wardpeet:
```suggestion
          if git diff --name-only "${{ github.event.pull_request.base.sha }}" "${{ github.event.pull_request.head.sha }}" | grep -qx 'packages/memory/integration-tests/package.json'; then
```

PR #17097, opened 2026-05-26
fix(core): end AGENT_RUN span when stream suspends for tool approval

intojhanurag:
## Description

When an agent stream terminated via `tool-call-approval` or `tool-call-suspended` (HITL flows), the `AGENT_RUN` observability span was never `.end()`-ed, so traces never reached Langfuse, Braintrust, Datadog, or any other backend that exports on `SPAN_ENDED`. This PR fires the existing `onFinish` callback from the suspend switch arm with a `finishReason: 'suspended'` payload, mirroring the pattern used for the prior error/abort fixes in #14850 and #14661.

image

image

image

## Related Issue(s)

Fixes #17074

## Type of Change

- [x] Bug fix (non-breaking change that fixes an issue)
- [ ] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [ ] Test update

## Checklist

- [ ] I have made corresponding changes to the documentation (if applicable)
- [x] I have added tests that prove my fix is effective or that my feature works
- [ ] I have addressed all Coderabbit comments on this PR

## ELI5

When an AI agent pauses because a tool needs human approval or a tool suspended itself, the monitoring trace could stay open and look like the run never finished. This PR makes sure those runs are explicitly closed and labeled "suspended" with the reason so observability shows what happened.

---

### Core Observability & Stream Finish Payloads

**packages/core/src/stream/base/output.ts**
- Ensure `options.onFinish` is invoked at most once per stream via a new `#finishCallbackSent` guard.
- When the stream suspends (`tool-call-approval` or `tool-call-suspended`), call `options?.onFinish` with a synthesized suspended payload containing:
  - `finishReason: 'suspended'`
  - `suspendReason` (e.g. `'tool-call-approval'` or `'tool-call-suspended'`)
  - `toolName` and `toolCallId` when available
  - empty/default placeholders for other fields (e.g., `text: ''`, empty arrays/objects for `toolCalls`, `toolResults`, `steps`, `usage`, `totalUsage`, and a safe `response: {}`) to avoid reading partial/unavailable response state
- Add helper `#createSuspendedOnFinishPayload` to build the safe suspended callback argument and ensure both suspended and final `finish` branches respect the single-callback guard.

**packages/core/src/agent/workflows/prepare-stream/map-results-step.ts**
[...]

PR #17134, opened 2026-05-27
fix(core): populate resumed AGENT_RUN span input and link to original trace

### LINKED ISSUE #17075: [FEATURE] agent.resumeStream() should surface resumeData as the AGENT_RUN span input
## Problem Statement

When `agent.resumeStream()` is called (e.g. after a HITL tool-call approval or rejection), the resulting `AGENT_RUN` span always has an **empty input** (`[]`). This is because `resumeStream()` passes `messages: []` to `#execute()` internally — the actual decision (approved/rejected, which tool, optional reason) lives in `resumeData` but is never propagated to the span.

In an observability backend like Langfuse, this means every resumed trace — whether it's an approval, a rejection, or a generic resume — looks identical: blank Input column. There is no way to distinguish what triggered the resumed run or what decision was made without digging into child spans.

Relevant code path: [`agent.resumeStream()` in `packages/core/src/agent/agent.ts`](https://github.com/mastra-ai/mastra/blob/main/packages/core/src/agent/agent.ts) passes `messages: []` to `#execute`, which then passes that empty array as `span.input`.

## Proposed Solution

The `AGENT_RUN` span's `input` should reflect the `resumeData` passed to `resumeStream()` / `resume()`. This would let users see at a glance:

- **Which tool** was being approved/rejected
- **The decision**: approved or rejected
- **Optional reason** (if the user provided one on rejection)
- **For generic resumes**: the full `resumeData` payload

Concretely, in `#execute()` (or `getOrCreateSpan()`), when `messages` is empty but `resumeData` is present, the span input should be set to the `resumeData` (or a structured summary of it) instead of the empty array.

For example, after an approval, the span input might look like:
```json
{
  "type": "tool-call-approval",
  "toolName": "deleteCampaign",
  "toolCallId": "call_abc123",
  "approved": true
}
```

After a rejection:
```json
{
  "type": "tool-call-approval",
  "toolName": "deleteCampaign",
  "toolCallId": "call_abc123",
  "approved": false,
  "reason": "User declined deletion"
}
```

This is purely additive — it only changes span metadata for resumed runs, where the input is currently always empty.

## Component

Agents, Observability

## Alternatives Considered

We currently work around this with a Hono middleware that:
1. Intercepts the request body on resume/approval/decline paths
2. Parses the approval decision from the body
3. Injects it into `requestContext` via `requestContext.set('resumeInput', payload)`
4. Uses a `customSpanFormatter` to rewrite the empty `span.input` from the `requestContext` data on `SPAN_ENDED`

This works but is fragile — it requires body-cloning middleware, path-matching heuristics, and formatter logic that duplicates what the framework should do natively.

## Example Use Case

As a developer using Mastra with HITL (human-in-the-loop) tool approval, I want to see in my Langfuse dashboard which tool was approved or rejected for each resumed agent run, so that I can audit decision patterns, debug approval flows, and understand why a particular run was resumed.

## Additional Context

The `resumeData` is already available inside `resumeStream()` — it just needs to be forwarded to the span as input. The data structures for approval (`{ approved, toolCallId, toolName }`) and rejection (`{ approved: false, reason }`) are well-defined in `@mastra/ai-sdk`'s `extractV6NativeApproval`.

This pairs well with (but is independent of) the orphan span closing bug (#17074) — that issue covers spans never ending, this one covers spans starting with empty input.

### Verification

intojhanurag:
## Description

Resumed agent runs (`agent.resumeStream()`, `agent.resumeGenerate()`, `approveToolCall`, `declineToolCallGenerate`) previously created an `AGENT_RUN` span with an empty `input`, and the resumed span was orphaned from the original suspended trace. This made HITL approval flows hard to interpret in tracing backends (Langfuse, Braintrust, Datadog). The resumed span now uses the resume payload — enriched with the suspended `toolName`/`toolCallId` when available — as its input, names itself `agent run: '<id>' (resumed)`, and re-parents itself to the persisted suspended span when one exists. Caller-supplied `tracingOptions` still take precedence. Mirrors the existing `Workflow.resume()` tracing pattern.

image

Pairs with #17097 (which ended the suspended span). That PR ensured the suspended trace closes; this PR ensures the resumed run is meaningful and stitched back to it.

## Related Issue(s)

Fixes #17075

## Type of Change

- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [ ] Test update

## Checklist

## ELI5 (Explain Like I'm 5)

When an agent pauses to ask a human and then resumes, the resumed run used to lose the reason and connection to the original paused trace. This PR makes resumed runs record the resume payload (approval/rejection and tool info) and reconnects the resumed span to the original suspended trace so the whole flow is clear in traces.

## Summary

This PR fixes observability for resumed agent executions by ensuring AGENT_RUN spans created during resume operations (agent.resumeStream(), agent.resumeGenerate(), approveToolCall(), declineToolCall()) are populated with the resume payload as the span input, are named to indicate they were resumed, and are re-parented to the persisted suspended span when available. Caller-supplied tracingOptions (traceId/parentSpanId) still take precedence.

### Key Changes

- packages/core/src/agent/agent.ts
  - Added `#getSuspendedToolInfo`(existingSnapshot) to extract suspended tool identifiers (toolName, toolCallId) from saved snapshots (supports requireToolApproval and generic suspend shapes).
  - Added `#getResumeSpanInput`(resumeData, suspendedToolInfo) to merge resumeData with suspended tool info (ensuring approved/declined, toolName, toolCallId, optional reason or full resumeData are surfaced).
  - Updated `#execute`() to:
    - Detect resumes and build resumeTracingOptions that prefer caller-supplied traceId/parentSpanId but fall back to persisted suspended tracing context.
    - Use a resumed span name `agent run: '<id>' (resumed)`.
    - Populate AGENT_RUN span input with resume-shaped data instead of empty input.
    - Add metadata flags resumed: true and resumedFromSpanId when re-parenting to a persisted suspended span.
    - Preserve entityVersionId when present.
- packages/core/src/agent/agent.types.ts
  - Narrowed resume snapshot typing to WorkflowRunState in InnerAgentExecutionOptions to reflect expected resume shape.
- Tests
  - Added tests at packages/core/src/agent/__tests__/resume-span-tracing.test.ts covering:
    - Approve (stream) and decline (generate) flows produce resumed AGENT_RUN spans with input.approved, toolName, toolCallId.
    - Generic resumeData is preserved in resumed span input.
    - Trace continuity: traceId preserved and parentSpanId links to suspended span when caller doesn't override.
    - Caller-provided tracingOptions override persisted tracing context.
- Changeset: `@mastra/core` patch. Type: Bug fix (non-breaking).

### Related

- Pairs with PR `#17097` (which ends the suspended span).
- Fixes issue `#17075`.

[Review Change Stack](https://app.coderabbit.ai/change-stack/mastra-ai/mastra/pull/17134?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


packages/core/src/agent/agent.ts
```diff
@@ -5686,13 +5727,45 @@ export class Agent<
...
+          | { traceId?: string; spanId?: string; parentSpanId?: string }
+          | undefined)
+      : undefined;
+
+    // Only fall back to persisted traceId/parentSpanId when the caller didn't provide
+    // their own. This prevents cross-trace parentage if the caller is explicit.
+    const userProvidedTraceId = options.tracingOptions?.traceId;
+    const userProvidedParentSpanId = options.tracingOptions?.parentSpanId;
+    const effectiveTraceId =
+      userProvidedTraceId ?? (!userProvidedParentSpanId ? persistedTracingContext?.traceId : undefined);
```
epinzur, line 5762: from claude:

> perhaps this should just be 
> 
> ```
> const effectiveTraceId = userProvidedTraceId ?? persistedTracingContext?.traceId;
> ```
> 
> to match the version in workflows

packages/core/src/agent/agent.ts
```diff
@@ -5498,6 +5498,47 @@ export class Agent<
...
+      return resumeData;
+    }
+
+    const resumeInput =
+      resumeData && typeof resumeData === 'object' && !Array.isArray(resumeData)
+        ? { ...(resumeData as Record<string, unknown>) }
+        : { resumeData };
+
+    return {
+      ...resumeInput,
```
epinzur: from claude:

> `#getResumeSpanInput` silently overwrites caller fields:
> 
> If `resumeData` already contains `toolName/toolCallId`, the spread order makes `suspendedToolInfo` win. In practice that's fine (they should match), but if a caller deliberately passes a different `toolName`, it disappears without warning. Consider letting resumeData's own fields win, or at least preserving them under a different key.

packages/core/src/agent/agent.ts
```diff
@@ -5498,6 +5498,47 @@ export class Agent<
     return undefined;
   }
 
+  #getSuspendedToolInfo(existingSnapshot: any): { toolCallId?: string; toolName?: string } | undefined {
```
epinzur: can we define any sort of type for `existingSnapshot` instead of `any`?

even `unknown` would be better.

epinzur: https://dev.to/sachinchaurasiya/type-safety-in-typescript-unknown-vs-any-55c0

packages/core/src/agent/agent.ts
```diff
@@ -5498,6 +5498,47 @@ export class Agent<
...
+      const payload = step.suspendPayload;
+      if (!payload) continue;
+
+      if (payload.requireToolApproval) {
+        return {
+          toolCallId: payload.requireToolApproval.toolCallId,
+          toolName: payload.requireToolApproval.toolName,
+        };
+      }
+      if (payload.toolCallSuspended || payload.toolName) {
```
epinzur: should there also be an OR on `payload.toolCallId`?

epinzur (approved):

packages/core/src/agent/agent.ts
intojhanurag: Left this one as-is intentionally. `workflow.ts:4077-4078` actually uses the same complex form, with the `!userProvidedParentSpanId` guard there to prevent cross-trace parentage when a caller passes only `parentSpanId` without `traceId`. The comment at `workflow.ts:4074-4076` explains the reasoning. Wanted to keep agent.ts symmetric with that. Happy to revisit if I'm misreading the workflows side.

PR #17171, opened 2026-05-27
feat(core): add opt-out for default channels context system message

CalebBarnes:
## What

Adds `channels.threadContext.addSystemMessage: boolean` (defaults to `true`) to opt out of the built-in channel system message.

```ts
new Agent({
  channels: {
    adapters: { slack: createSlackAdapter() },
    threadContext: {
      addSystemMessage: false, // skip the built-in channel system message
    },
  },
});
```

## Why

Issue #16947 has two parts:

1. The built-in channel system message can't be customized or removed.
2. `ChatChannelProcessor` was clobbering other processors' tagged system messages (already fixed in #17168).

For (1), the main demand is *opt-out*. Static and dynamic customization were considered but rejected:

- A function form (`(ctx) => string`) is a prompt-cache footgun — easy to vary content per request, breaking caching for every step.
- A static `string` form is marginal over the agent's existing `instructions` field.
- For real customization, users can already register their own input processor with `id: 'chat-channel-context'` — it shadows the built-in one via the existing `hasProcessor` check in `getInputProcessors`.

So this PR ships only the boolean.

### Why `threadContext.addSystemMessage` and not `addProcessor`

The name describes what the user sees (a system message) rather than how it's built (a processor). It also clusters with the existing `threadContext.maxMessages` field — both are about how the agent picks up context about the current thread.

## How

- `ChannelConfig.threadContext.addSystemMessage?: boolean` added in `packages/core/src/channels/types.ts`.
- `AgentChannels.getInputProcessors()` returns `[]` when `threadContext.addSystemMessage === false`; otherwise falls through to existing default + escape-hatch logic.
- `ChatChannelProcessor` itself is unchanged from main.

## Tests

- Agent channels wire-up tests (`agent-channels.test.ts`) cover: default behavior, `addSystemMessage: false` skip, user-supplied processor escape hatch.
- 145/145 channel tests pass, 45/45 process-input-step tests pass, typecheck clean.

## Docs

- `docs/src/content/en/docs/agents/channels.mdx`: new "Disabling the channel system message" section.
- `docs/src/content/en/reference/agents/channels.mdx`: `threadContext` type updated.

## Changeset

Minor bump on `@mastra/core`.

## ELI5

This PR lets developers turn off an automatic message that Mastra adds to tell the agent which channel or platform a request came from (like whether it's a direct message or public channel). Before, there was no way to disable this message—now you can set `threadContext: { addSystemMessage: false }` in your agent config to skip it.

### Configuration & Types
- Added `addSystemMessage?: boolean` option to `ChannelConfig.threadContext` in `packages/core/src/channels/types.ts`. Defaults to `true` to preserve existing behavior.

### Implementation
- Updated `AgentChannels.getInputProcessors()` in `packages/core/src/channels/agent-channels.ts` to skip injecting the built-in `ChatChannelProcessor` when `threadContext.addSystemMessage === false`.
- Preserved the existing escape-hatch behavior: if a caller provides a custom processor with id `chat-channel-context`, the built-in processor is not added.

### Tests
- Added `getInputProcessors()` test suite in `packages/core/src/channels/__tests__/agent-channels.test.ts` covering:
  - Default behavior (chat-channel-context processor included)
  - Opt-out behavior when `addSystemMessage: false`
  - Escape-hatch verification (custom processor with same id prevents built-in injection)

### Documentation
- Updated `.changeset/chat-channel-thread-system-message.md` with changeset entry describing the new option.
- Added "Thread context" note in `docs/src/content/en/docs/agents/channels.mdx` explaining the system message and how to disable it.
- Updated `threadContext` type documentation in `docs/src/content/en/reference/agents/channels.mdx` to include the new `addSystemMessage` option alongside existing `maxMessages`.

## Design Notes
The implementation uses a simple boolean flag rather than more complex alternatives (dynamic functions or string customization) because users can already register custom input processors if they need full control, avoiding prompt-caching concerns and configuration duplication.

[Review Change Stack](https://app.coderabbit.ai/change-stack/mastra-ai/mastra/pull/17171?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


CalebBarnes:
@coderabbitai review

TylerBarnes (approved):

PR #17176, opened 2026-05-27
fix(core): keep internal stream workflows transient

graysonhicks:
## Problem

Users can see this debug log during `agent.stream()` / `agent.generate()`:

```text
Cannot get workflow run. Mastra storage is not initialized
```

## Verified User Context

The reported case is a supervisor/sub-agent pattern. The customer's code describes Elle as the Coordinator/supervisor and explicitly says Elle's `agents:` map is the delegation surface where Elle picks up sub-agents as tools. The attached support note also reports the logs as coming from the sub-agent/supervisor setup.

That matches the current core path: while assembling supervisor tools, `Agent.listAgentTools()` derives background config for each sub-agent, and `deriveSubAgentBackgroundConfig()` calls `subAgent.getToolsForExecution({ requestContext })` for full `Agent` sub-agents. So this issue is showing up when the supervisor streams/generates with sub-agents configured.

Important nuance: the storage log itself was attributed in the user's platform logs to the supervisor's synthesized internal stream workflow spans, not to sub-agent spans directly. So the pattern that exposes the issue is supervisor/sub-agent delegation, but the specific `Cannot get workflow run...` log is emitted by the supervisor's internal `execution-workflow` plumbing.

## Technical Problem

The log comes from the transient internal agent stream workflow (`execution-workflow`). That workflow is created inside agent execution as an implementation detail. It is not a user-registered workflow, and attaching it directly to the user's Mastra instance has downsides because it can expose or persist internal `execution-workflow` runs in user workflow storage/observability.

The specific call path is that `Agent` creates the internal stream workflow, calls `executionWorkflow.createRun()`, and `Workflow.createRun()` performs a workflow-run recovery lookup via `getWorkflowRunById()`. Since the transient workflow has no Mastra storage attached, that lookup emits the misleading storage debug log even though the run can continue from the in-memory workflow run.

## Current Draft Implementation

This draft currently tries to keep the internal stream workflow transient by:

- marking the internal `execution-workflow` as non-persistent with `shouldPersistSnapshot: () => false`
- changing `Workflow.createRun()` so it skips the initial storage recovery lookup when snapshot persistence is disabled
- adding a regression test that asserts internal `execution-workflow` runs are not written to workflow storage

## Uncertainty / Review Needed

I am not fully confident this is the right abstraction boundary.

The concern is that the current draft changes the meaning of `shouldPersistSnapshot`. Today that option is about whether snapshots are persisted. This draft also makes it affect whether `createRun()` performs the initial storage recovery lookup. That may be a reasonable interpretation for non-persistent workflows, but it is still a broader workflow-runtime behavior change than the original agent log issue.

Specific questions for a domain reviewer:

- Should `shouldPersistSnapshot: false` also imply "do not perform storage recovery lookup"?
- Could any existing workflow rely on `shouldPersistSnapshot` returning false for the initial pending state while still expecting `createRun({ runId })` to recover state from storage?
- Should internal agent stream workflows have a more explicit internal/transient marker instead of reusing snapshot persistence behavior?
- Would a narrower fix be better, such as suppressing the storage debug log only when `getWorkflowRunById()` can return an in-memory run fallback?
- Is attaching the internal stream workflow to Mastra definitely wrong because of storage/observability pollution, or is that acceptable and expected for internal workflows?
- Should supervisor/sub-agent tool-probing paths tag or suppress internal workflow/storage logs differently from normal user workflow execution?

## Alternatives Considered

1. Attach/create the internal stream workflow with the active Mastra instance. This fixes the log, but risks polluting user workflow storage and observability with internal `execution-workflow` runs.
2. Suppress the storage debug log for internal workflows or for successful in-memory fallback. This is narrower behaviorally, but may hide useful diagnostics if applied too broadly.
3. Add a private internal/transient workflow concept. This may be cleaner long-term, but is larger than the immediate bug fix.

## Validation Run

- `pnpm test packages/core/src/agent/__tests__/working-memory-context.test.ts`
- `pnpm test packages/core/src/workflows/workflow.test.ts`
- `pnpm prettier --check packages/core/src/agent/agent.ts packages/core/src/agent/workflows/prepare-stream/index.ts packages/core/src/workflows/workflow.ts packages/core/src/agent/__tests__/working-memory-context.test.ts .changeset/odd-hands-juggle.md`

CodeRabbit CLI is installed locally, but `coderabbit doctor` reports it is not signed in, so I could not run a local CodeRabbit review without an auth step.


abhiaiyer91:
Closing, reopen when ready

PR #17185, opened 2026-05-27
perf(core): speed up channel message handling by dropping blocking storage writes

CalebBarnes:
## Summary

Removes two awaited storage round-trips from the channels per-message dispatch path. Each chat message dispatch now blocks on zero storage operations before `sendSignal` returns — down from two (`chatThread.subscribe()` write, then `memoryStore.getThreadById()` read).

## What changed

In `processChatMessage`:

- `await chatThread.subscribe()` → fire-and-forget `void chatThread.subscribe().catch(...)`. The subscribe metadata write now lands in parallel with (or after) signal dispatch.
- `memoryStore.getThreadById()` + the `threadForRun` snapshot refresh → deleted entirely.
- `streamOptions.memory.thread: threadForRun` (the refreshed snapshot) → `streamOptions.memory.thread: mastraThread.id` (plain id). The agent re-reads the thread from storage inside `prepareMemoryStep` anyway, so passing the id is functionally identical and removes any reason for the channels layer to refresh.

Net: −14 / +10 in `agent-channels.ts`, plus a changeset.

## Why it's safe

The original ordering existed to guard against a metadata clobber: an earlier version of `prepareMemoryStep` did `metadata: thread.metadata`, which would overwrite the freshly-written `channel_subscribed: 'true'` with the stale pre-subscribe snapshot. That clobber was later fixed to a merge:

```ts
// prepare-memory-step.ts
metadata: { ...(existingThread.metadata ?? {}), ...thread.metadata }
```

Walking the two real scenarios:

- **Cold thread, first message.** The `mastraThread` snapshot from `getOrCreateThread` does not contain `channel_subscribed` at all. `prepareMemoryStep` re-reads the thread from storage (which by then either does or doesn't have `channel_subscribed: 'true'` depending on whether the fire-and-forget write has landed yet — usually it has, since the agent's memory step runs many awaits into the run). The merge spreads `existingThread.metadata` first, so whatever is in storage wins. Nothing gets clobbered either way.
- **Warm thread, subsequent message.** Both `existingThread.metadata` and `mastraThread.metadata` already have `channel_subscribed: 'true'`. `deepEqual` is equal, no write happens. Same as before.

So the workaround was guarding against a bug that has already been fixed elsewhere.

## What this does and doesn't fix

**Fixes:** Two storage round-trips per message. Felt on every dispatch, not just cold starts. The wall-clock impact scales linearly with storage backend latency — a few ms on local SQLite, tens of ms on remote Postgres.

**Does not fix:**

- `isSubscribed()` round-trip earlier in the path (in-process cache is a separate change).
- `fetchThreadHistory` on cold non-DM mentions (predates the signals migration; out of scope).
- Cross-instance coordination on stateless serverless (still needs shared PubSub).
- Cold-lambda module init (`getChatModule()` dynamic import).

## Test plan

- `pnpm --filter ./packages/core test -- src/channels` — 145/145 passed
- `pnpm --filter ./packages/core test -- src/processors/process-input-step` — 45/45 passed
- `pnpm --filter ./packages/core test -- src/agent/__tests__/memory-metadata.test.ts` — 13/13 passed (covers the metadata-merge behavior this relies on)

## ELI5

This PR makes sending chat messages faster by not waiting for two slow storage reads/writes during each message. It starts saving subscription info in the background and passes just the thread ID so the agent can re-read the thread when needed.

### `packages/core/src/channels/agent-channels.ts`

- Replaced awaited `await chatThread.subscribe()` with fire-and-forget `void chatThread.subscribe().catch(...)`, allowing the subscribe metadata write to run concurrently with signal dispatch.
- Removed the `memoryStore.getThreadById()` call and the `threadForRun` snapshot refresh.
- Pass `streamOptions.memory.thread: mastraThread.id` (thread id) instead of a refreshed thread snapshot; the agent re-reads the thread in `prepareMemoryStep`.

Net diff: −14 / +10.

### `.changeset/channels-cold-start-fire-and-forget-subscribe.md`

- Added a patch changeset noting reduced per-message latency by removing two awaited storage round-trips.

## Rationale and Safety

- `prepareMemoryStep` was fixed to merge metadata (`metadata: { ...(existingThread.metadata ?? {}), ...thread.metadata }`), preventing stale snapshots from clobbering `channel_subscribed: 'true'`.
- Cold threads: `prepareMemoryStep` re-reads storage and the merge ensures stored values win regardless of subscribe timing.
- Warm threads: metadata already contains `channel_subscribed: 'true'`, so no conflicting write occurs.
- Effectively removes two awaited storage operations per message (subscribe write await and getThreadById read) without changing other round-trips.

## Tests

- Core channel tests: 145/145 passed
- Processors process-input-step tests: 45/45 passed
- memory-metadata unit tests: 13/13 passed (covers metadata-merge)

[Review Change Stack](https://app.coderabbit.ai/change-stack/mastra-ai/mastra/pull/17185?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


CalebBarnes:
@coderabbitai review

abhiaiyer91 (approved):

PR #17190, opened 2026-05-28
fix(gateway): support custom gateways without apiKeyEnvVar

### LINKED ISSUE #17139: [FEATURE] Support custom gateways that don't necessarily use an API key env var
### Problem Statement

Some custom gateways might have custom authentication systems that might not use an env variable (e.g. Application default credentials), but currently:
1. `apiKeyEnvVar` is required, along with it being required around in other places in the API, like probably the providers list API.
2. `connected` in the providers list is set according to the env var, and not according to whether the `getApiKey` or any other method returns anything... So you get a warning about it in studio, even if the provider otherwise works.

P.S.  I would expect only `fetchProviders` and `resolveLanguageModel` to really be needed in the interface, with not all providers/gateways actually needing `getApiKey`, `buildUrl`, leaving API key and URL building decisions to `resolveLanguageModel` instead of hardcoding a call to `getApiKey` in Mastra itself, and buildUrl seems to really only be called inside resolveLanguageModel anyhow...

Also it seems that the models array returned in each provider needs to also include the provider ID to work correctly, despite already having the provider ID

### Proposed Solution

Modify the `MastraModelGateway` interface and API to accommodate this. For example, let the gateway/provider decide whether it is `connected`, etc.

### Component

Studio, Server, Other

### Alternatives Considered

Ignore the warning and go along with it :P 

### Example Use Case

Custom gateway that uses Azure auth, and uses Azure default credentials if nothing else is set.

### Additional Context

I specifically have a case where I want to pass an API key from a config file directly into my custom gateway when Mastra is embedded in another server, instead of requiring users to also set it in an env var.

### Verification

Akash504-ai:
## Description

Supports custom gateways/providers that do not require an `apiKeyEnvVar`.

This updates provider auth handling to allow gateways that use alternative authentication flows (e.g. Azure default credentials, embedded config auth, custom runtime auth) without being incorrectly marked as disconnected.

Changes include:

* Making `apiKeyEnvVar` optional in `ProviderConfig`
* Treating providers without `apiKeyEnvVar` as connected by default
* Preventing embedding router failures when no env var is configured
* Updating provider connection logic in server handlers
* Updating provider schema typings
* Adding regression tests for gateways/providers without `apiKeyEnvVar`

This preserves backward compatibility for existing providers using env-based API keys.

## Related issue(s)

Fixes #17139

## Type of change

* [x] Test update

## Checklist

* [x] I have linked the related issue(s) in the description above

## ELI5

This PR lets gateway providers work without requiring a secret stored in an environment variable. If a provider doesn't declare an API-key env var, it's treated as connected and can handle auth itself.

## Overview

Make ProviderConfig.apiKeyEnvVar optional and update runtime, router, server handlers, schemas, and tests so providers that don't declare env-based API keys are considered connected by default. When env-var mappings are present, checks now treat the provider as connected if any configured env var is set. Embedding/router logic no longer forces env-var lookups when none are declared, and missing env-vars cause errors only when an env-var mapping was expected.

## Key Changes

- packages/core/src/llm/model/gateways/base.ts
  - ProviderConfig.apiKeyEnvVar changed from required (string | string[]) to optional (apiKeyEnvVar?: string | string[]).

- packages/core/src/harness/harness.ts
  - getCurrentModelAuthStatus() and listAvailableModels() now normalize apiKeyEnvVar (support string | string[]), treat providers with no declared env vars as having auth by default, and expose the first normalized env var (or undefined) per model.

- packages/core/src/llm/model/embedding-router.ts
  - Only attempts env var lookups when providerConfig.apiKeyEnvVar is defined. If no env-var mapping exists, the router will not throw an “API key not found” error and may fall back to an empty-string key or provider-supplied key.

- packages/server/src/server/handlers/agents.ts
  - isProviderConnected() now:
    - Returns true when provider.apiKeyEnvVar is missing/falsy or normalizes to no truthy entries (providers treated as connected by default).
    - When env-var names are declared, normalizes/filter them and returns true if any of the remaining env vars is set in process.env (previous behavior required all).
  - Server-side provider listing uses undefined for envVar when absent.

- packages/server/src/server/schemas/agents.ts
  - providersResponseSchema: envVar field changed from required to optional so provider responses may omit envVar.

- Tests
  - packages/core/src/llm/model/gateways/custom-gateway.test.ts: Added NoEnvKeyCustomGateway and tests ensuring fetchProviders() and ModelRouterLanguageModel work when apiKeyEnvVar is omitted.
  - packages/server/src/server/handlers/agents.test.ts: Added tests verifying isProviderConnected returns true for providers without apiKeyEnvVar and that GET_PROVIDERS_ROUTE.handler marks such providers connected while returning envVar as undefined.

## Compatibility & Behavior Notes

- Backward compatibility preserved for providers that still declare apiKeyEnvVar.
- Providers that omit apiKeyEnvVar must implement their own auth logic (e.g., getApiKey, resolveLanguageModel, embedded config, ADC, or runtime flows).
- Server-side connectivity checks now consider providers with any configured env var set as connected (logical OR), and consider providers without env-var mappings connected by default.

[Review Change Stack](https://app.coderabbit.ai/change-stack/mastra-ai/mastra/pull/17190?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/packages/core/src/harness/harness.ts
+++ b/packages/core/src/harness/harness.ts
@@ -722,7 +722,7 @@ export class Harness<TState = {}> {
       const providerConfig = registry[provider];
       const envVars = providerConfig?.apiKeyEnvVar;
       const apiKeyEnvVar = Array.isArray(envVars) ? envVars[0] : envVars;
-      if (apiKeyEnvVar && process.env[apiKeyEnvVar]) {
+      if (!apiKeyEnvVar || (apiKeyEnvVar && process.env[apiKeyEnvVar])) {
         return { hasAuth: true };
       }
       return { hasAuth: false, apiKeyEnvVar: apiKeyEnvVar || undefined };
@@ -767,7 +767,7 @@ export class Harness<TState = {}> {
         const providerConfig = registry[provider];
         const envVars = providerConfig?.apiKeyEnvVar;
         const apiKeyEnvVar = Array.isArray(envVars) ? envVars[0] : envVars;
-        const hasEnvKey = apiKeyEnvVar ? !!process.env[apiKeyEnvVar] : false;
+        const hasEnvKey = apiKeyEnvVar ? !!process.env[apiKeyEnvVar] : true;
 
         let hasApiKey = hasEnvKey;
         if (!hasApiKey && this.config.modelAuthChecker) {
--- a/packages/core/src/harness/harness.ts
+++ b/packages/core/src/harness/harness.ts
@@ -721,11 +721,14 @@ export class Harness<TState = {}> {
       const registry = PROVIDER_REGISTRY as Record<string, { apiKeyEnvVar?: string | string[] }>;
       const providerConfig = registry[provider];
       const envVars = providerConfig?.apiKeyEnvVar;
-      const apiKeyEnvVar = Array.isArray(envVars) ? envVars[0] : envVars;
-      if (!apiKeyEnvVar || (apiKeyEnvVar && process.env[apiKeyEnvVar])) {
-        return { hasAuth: true };
-      }
-      return { hasAuth: false, apiKeyEnvVar: apiKeyEnvVar || undefined };
+      const normalizedEnvVars = Array.isArray(envVars) ? envVars.filter(Boolean) : envVars ? [envVars] : [];
+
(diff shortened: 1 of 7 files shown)


wardpeet:
I've solved it differently and it's part of Gatewat now. Thanks for the contribution!

PR #17192, opened 2026-05-28
fix(core): register Mastra on execution workflow before createRun()

### LINKED ISSUE #17137: [BUG] "Cannot get workflow run. Mastra storage is not initialized" debug log spam on every agent.generate() despite storage configured
### Describe the Bug

Calling `agent.generate()` on an agent registered to a Mastra instance with `storage` configured produces 5+ debug logs `Cannot get workflow run. Mastra storage is not initialized` per call.

Root cause traced in `@mastra/core@1.32.1` dist :

1. `Workflow.getWorkflowRunById` at `dist/chunk-QPZ35KK2.cjs:13878` :
```js
async getWorkflowRunById(runId, options = {}) {
  const storage = this.#mastra?.getStorage();
  if (!storage) {
    this.logger.debug("Cannot get workflow run. Mastra storage is not initialized");
    return this.#getInMemoryRunAsWorkflowState(runId);
  }
  // ...
}
```

2. `this.#mastra` is undefined on the internal `agenticLoopWorkflow` that each `agent.generate()` creates. The wiring at `dist/chunk-QPZ35KK2.cjs:20143-20162` :
```js
const agenticLoopWorkflow = createAgenticLoopWorkflow({ ...rest });
if (rest.mastra) {
  agenticLoopWorkflow.__registerMastra(rest.mastra);
}
```
`rest.mastra` is falsy in this code path even though the parent agent was correctly registered via `new Mastra({ agents, storage })`. So `__registerMastra` is never called, and every `createRun()` on the internal workflow logs the debug warning before falling back to in-memory state.

The fallback works — `agent.generate()` returns the expected output — so this is **debug-noise only, no functional failure**. But the noise scales linearly with `.generate()` call volume (live scorers at sampling rate 1, dataset experiments with many items), and is misleading when actually debugging real storage misconfigurations.

Related : the closed issue #11116 reports a similar symptom with Inngest, claimed fixed in v1.0. The warning still fires in 1.32 with a vanilla `LibSQLStore` setup, no Inngest, no NestJS — just plain agent + storage.

### Steps To Reproduce

1. Clone the minimal repro repo : https://github.com/obigroup/mastra-storage-warning-repro
2. `cd mastra-storage-warning-repro`
3. `npm install`
4. `npm test`
5. Observe in the Vitest stdout : at least 5 lines of
   `Cannot get workflow run. Mastra storage is not initialized`
   emitted during the single `agent.generate()` call.

No API key is required to reproduce the bug : the warning fires inside `createRun()` before any LLM call. The repro intentionally catches the downstream LLM error so the focus remains on the storage warning.

https://github.com/obigroup/mastra-storage-warning-repro

### Expected Behavior

No debug log about uninitialized storage when both conditions hold :
- The agent is registered to a Mastra instance via `new Mastra({ agents })`.
- That Mastra instance has `storage` configured (here : `LibSQLStore`).

Either :
1. The internal `agenticLoopWorkflow` created by `agent.generate()` should systematically receive the parent agent's mastra reference, so `__registerMastra` is always called and the internal workflow can access storage.
2. Or the `logger.debug` line should be removed / scoped to genuinely unexpected uninitialized-storage scenarios (currently it fires on the happy path).

### Environment Information

```block
System:
    OS: Linux 7.0 Ubuntu 26.04 LTS 26.04 LTS (Resolute Raccoon)
    CPU: (24) x64 AMD Ryzen AI 9 HX 370 w/ Radeon 890M
    Memory: 30.77 GB / 60.66 GB
    Container: Yes
    Shell: 5.3.9 - /bin/bash
  Binaries:
    Node: 25.2.1 - /home/vev/.nvm/versions/node/v25.2.1/bin/node
    npm: 11.6.2 - /home/vev/.nvm/versions/node/v25.2.1/bin/npm
    Deno: 2.6.10 - /home/vev/.deno/bin/deno
  Browsers:
    Chromium: 148.0.7778.167
    Firefox: 151.0
    Firefox Developer Edition: 151.0
  npmPackages:
    @mastra/core: 1.32.1 => 1.32.1 
    @mastra/libsql: 1.10.0 => 1.10.0 
    @types/node: ^22 => 22.19.19 
    typescript: ^5.6 => 5.9.3 
    vitest: ^4.1.7 => 4.1.7 
  npmGlobalPackages:
    npm: 11.6.2
    opencode-ai: 1.2.27
```

### Verification

KrishavRajSingh:
## Description

Wire the internal prepare-stream workflow to the agent's Mastra instance so agent generate/stream can use configured storage instead of falling back to in-memory state and emitting debug noise.

## Related issue(s)

Fixes #17137 

## Type of change

- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [ ] Test update

## Checklist

## ELI5
The PR fixes annoying debug log messages that kept appearing when you asked an agent to generate responses. The agent was supposed to use a storage system you gave it, but the internal workflow it creates wasn't being told about that storage, so it complained about not having it (even though it was fine). This fix simply tells the internal workflow "hey, here's the storage" before it starts working, so it stops complaining.

### Core Fix
- **packages/core/src/agent/agent.ts**: Modified `Agent.#execute` method to register the Mastra instance on the internal execution workflow via `executionWorkflow.__registerMastra(this.#mastra)` when Mastra is present, immediately before creating the workflow run via `createRun()`.

### Documentation
- **.changeset/eight-rules-pull.md**: Added changeset documenting a patch-level fix for `@mastra/core` that eliminates debug log noise from agent generate/stream operations by ensuring the internal execution workflow uses configured Mastra storage.

## Impact
- **Fixes**: Issue `#17137` — eliminates spurious "Cannot get workflow run. Mastra storage is not initialized" debug logs when calling `agent.generate()` or `agent.stream()` on agents registered to Mastra instances with configured storage
- **Behavior**: The internal agent execution workflow now properly receives the parent Mastra reference, allowing it to use configured storage instead of falling back to in-memory state while logging warnings
- **No breaking changes**: The fallback mechanism remains functional; this only prevents misleading debug noise

[Review Change Stack](https://app.coderabbit.ai/change-stack/mastra-ai/mastra/pull/17192?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/.changeset/eight-rules-pull.md
+++ b/.changeset/eight-rules-pull.md
@@ -1,0 +1,5 @@
+---
+'@mastra/core': patch
+---
+
+Fixed internal execution workflow not using Mastra storage during agent generate and stream, which caused debug log noise when storage was configured.
--- a/packages/core/src/agent/agent.ts
+++ b/packages/core/src/agent/agent.ts
@@ -5821,6 +5821,10 @@ export class Agent<
       drainPendingSignals: runId => agentThreadStreamRuntime.drainPendingSignals(runId, threadStreamPubSub),
     });
 
+    if (this.#mastra) {
+      executionWorkflow.__registerMastra(this.#mastra);
+    }
+
     const run = await executionWorkflow.createRun();
     const observabilityContext = createObservabilityContext({ currentSpan: agentSpan });
     const result = await run.start({ ...observabilityContext });


abhiaiyer91:
Thanks for digging into #17137 and tracing the root cause — the diagnosis here is spot on, and registering Mastra on the internal execution workflow is the right instinct.

One thing to flag before this lands: registering Mastra on `execution-workflow` on its own has a side effect. That workflow has no `shouldPersistSnapshot` option, so it defaults to `() => true`. Once it can see storage, `createRun()` will persist a `pending` snapshot row to the user's storage on **every** `agent.generate()` / `agent.stream()` call — under the fixed `workflowName: 'execution-workflow'`.

I verified this locally: applying this change produces 1 persisted `execution-workflow` row per `generate()`, where `main` produces 0. So it trades the debug-log noise for storage-write noise on the same hot paths the issue calls out (scorers at sampling 1, dataset experiments).

The sibling internal workflows (`agentic-loop`, `executionWorkflow`) deliberately set `shouldPersistSnapshot` to control exactly this. The complete fix is to also opt this workflow out of persistence (`shouldPersistSnapshot: () => false` in `createPrepareStreamWorkflow`), since it's internal and non-resumable and shouldn't write any rows.

I've opened #17344 with that approach plus two regression tests (one for the original log bug, one that fails specifically when Mastra is registered without the persistence opt-out). Happy to credit your diagnosis there. Thanks again for the clear write-up on the issue — it made tracking this down much faster.

PR #17209, opened 2026-05-28
[codex] Improve Studio startup error handling

damien-schneider:
## Summary
- move the pre-React Studio startup fallback out of `index.html` and into a small `bootstrap.ts` entrypoint
- catch only the app entry import/direct start failure path, so ordinary runtime errors still flow through React, route error handling, console reporting, and Vite's dev overlay
- preserve the original caught startup error with `console.error(...)` before rendering the fallback
- show raw startup error details in development, or in production-style builds only when `VITE_MASTRA_STUDIO_SHOW_STARTUP_ERROR_DETAILS=true` is explicitly set
- add focused jsdom coverage for console preservation, detailed output, hidden production-style details, and the guard that avoids replacing an already-rendered app

Before: (empty white screen)

image

After: with the error message

image

## Security note
Raw startup errors can include implementation details such as stack frames, import URLs, local paths, or accidentally sensitive thrown values. This PR keeps those details dev-only by default while still avoiding a blank screen in production-style builds.

## Validation
- `pnpm --filter ./packages/playground exec vitest run src/bootstrap.test.ts src/startup-error.test.ts`
- `pnpm --filter ./packages/playground typecheck`
- `pnpm --filter ./packages/playground build`
- `pnpm --filter ./packages/playground lint` (passes with the existing 63 warnings)
- `pnpm dev:ui` from `examples/agent`, then local browser load of `http://localhost:5173/agents`: renders `Mastra Studio failed to start` with the original `@standard-schema/spec` startup import error instead of a blank screen
- CodeRabbit CLI doctor: installed and reachable, but not signed in, so no local review run

## ELI5

When Mastra Studio fails to start, this PR ensures users see a clear error screen instead of a blank page. It moves the startup fallback into small, tested files so the page stays minimal and error details are shown only in dev (or when explicitly enabled).

## Overview

This PR extracts the pre-React startup fallback out of index.html into a tiny bootstrap entrypoint (packages/playground/src/bootstrap.ts) and a DOM helper (packages/playground/src/startup-error.ts). The bootstrap only catches failures that occur while importing or calling the app's entry (startStudio); normal runtime errors continue to be handled by React, route boundaries, console reporting, and Vite's dev overlay.

## Key Changes

- index.html: load ./src/bootstrap.ts as the module entry instead of inlining startup logic.
- bootstrap.ts: new minimal entrypoint that dynamically imports ./main and calls startStudio(); on import/startup failure it logs the original error and calls renderStartupError(error).
- main.tsx: refactored to export startStudio(), moving prior top-level initialization into that function.
- startup-error.ts: new renderStartupError(error, { showDetails }) helper that:
  - computes a readable message from unknown errors,
  - renders a full-page alert into `#root` when startup fails before React mounts,
  - shows raw error details in dev (or when VITE_MASTRA_STUDIO_SHOW_STARTUP_ERROR_DETAILS=true),
  - avoids replacing content if `#root` is missing or already has children.
- Added changeset: .changeset/studio-startup-error-ui.md documenting the behavior change.

## Testing

- New Vitest/JSDOM tests:
  - bootstrap.test.ts: verifies console.error is preserved and renderStartupError is invoked on startup failure.
  - startup-error.test.ts: verifies detailed vs. hidden output, and that the renderer does not replace an already-rendered app.
- Validation performed: targeted vitest runs, typecheck/build/lint for playground package, manual dev run showing the fallback UI with original import error, and CodeRabbit CLI doctor noted.

## Security

Raw error details (stacks, import URLs, local paths, thrown values) are potentially sensitive. The PR keeps these details visible only in development by default; production-style builds hide details unless VITE_MASTRA_STUDIO_SHOW_STARTUP_ERROR_DETAILS=true is explicitly set.

## Notes

- Reviewer comment: add a changeset (a changeset was added in this PR).

[Review Change Stack](https://app.coderabbit.ai/change-stack/mastra-ai/mastra/pull/17209?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/.changeset/studio-startup-error-ui.md
+++ b/.changeset/studio-startup-error-ui.md
@@ -1,0 +1,5 @@
+---
+'@internal/playground': patch
+---
+
+Improved how the Studio surfaces startup failures before React mounts. Import errors and direct boot failures now render a clear fallback in the page instead of leaving a blank screen, while normal runtime errors stay with React, route error states, and Vite's dev overlay.
--- a/packages/playground/index.html
+++ b/packages/playground/index.html
@@ -89,10 +89,6 @@
 
   <body class="overflow-hidden">
     <div id="root"></div>
-    <script type="module">
-      // Use relative import to work with any base path
-      // Vite transforms this during build to reference compiled output with correct paths
-      import('./src/main.tsx');
-    </script>
+    <script type="module" src="./src/bootstrap.ts"></script>
   </body>
 </html>
--- a/packages/playground/src/bootstrap.test.ts
+++ b/packages/playground/src/bootstrap.test.ts
@@ -1,0 +1,29 @@
+// @vitest-environment jsdom
+import { afterEach, describe, expect, it, vi } from 'vitest';
+
+afterEach(() => {
+  vi.resetModules();
+  vi.doUnmock('./main');
+  vi.doUnmock('./startup-error');
+  vi.restoreAllMocks();
+});
+
+describe('bootstrap', () => {
+  it('preserves startup errors in the console before rendering the fallback', async () => {
+    const startupError = new Error('broken startup import');
+    const renderStartupError = vi.fn();
+    const consoleError = vi.spyOn(console, 'error').mockImplementation(() => {});
+
+    vi.doMock('./main', () => ({
+      startStudio: () => {
+        throw startupError;
+      },
+    }));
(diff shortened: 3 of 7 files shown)


wardpeet (approved):

wardpeet:
@damien-schneider don't forget a changeset

PR #17238, opened 2026-05-28
03 feat(client): add message helper APIs

TylerBarnes:
## Stack position

Stack 3 of 8 in the signals API cleanup.

- Previous PR: #17237
- Next PR: #17240

## Summary

Adds SDK-level message helpers on top of the server routes from Stack 2 and migrates the React chat path to use the message-first API when available.

- Adds `agent.sendMessage()` and `agent.queueMessage()` to `@mastra/client-js`
- Adds generated route types and exported helper parameter types
- Preserves object payloads, attributes, metadata, provider options, resource/thread IDs, and idle stream options
- Updates `@mastra/react` chat sending to prefer `sendMessage()` while keeping the existing legacy `/signals` fallback for older servers
- Keeps `data-user-message` stream compatibility while accepting canonical `type: 'user'` signal metadata
- Updates client-js reference docs and package changesets

## Test plan

- `pnpm --filter ./client-sdks/client-js test src/resources/agent.test.ts`
- `pnpm --filter ./client-sdks/client-js build`
- `pnpm --filter ./client-sdks/react test src/lib/ai-sdk/utils/toUIMessage.test.ts`
- React build was run; remaining errors were pre-existing unrelated package errors noted during implementation

## ELI5 (Explain Like I'm 5)

This PR adds two simple helpers so apps can give messages to agents: sendMessage() to deliver a message now and queueMessage() to save a message for the next turn. It makes sending user input simpler while keeping fallbacks so older servers and the existing signal API still work.

## Overview

Adds experimental, message-first helper APIs to `@mastra/client-js` and updates `@mastra/react` to prefer them for user-authored thread input. Includes implementation, exported types, React integration changes with safe fallbacks, docs, tests, and changesets.

## New Features

- Agent methods
  - agent.sendMessage(params): send user-authored input to an active agent run or idle thread (POST /agents/:agentId/send-message)
  - agent.queueMessage(params): enqueue user-authored input for the next thread turn (POST /agents/:agentId/queue-message)
  - Both preserve object payloads, attributes, metadata, provider options, resource/thread IDs, and ifIdle.streamOptions; return { accepted: true, runId, signal? }.

- Exported types
  - SendAgentMessageParams
  - QueueAgentMessageParams

## Implementation Details

- Added private helpers prepareSignalRouteBody and requestSignalRoute to centralize request preparation, ifIdle.streamOptions transformation (including processed clientTools), runtime option tracking, and cleanup on failure.
- sendSignal now delegates to requestSignalRoute to remove duplication.
- sendMessage and queueMessage reuse the same signal request preparation so idle stream options and serialized client-tool/runtime options are preserved for server handling and subscribed continuations.
- Tests assert payload preservation (strings and objects), parts array preservation for queueMessage, clientTools processing for idle continuations, and runtime option caching behavior.

## React Integration

- useChat now prefers agent.sendMessage() for thread input, computes messageContents and reusable streamOptions, and extracts echoedSignalId from responses.
- On unsupported-server responses from sendMessage, it calls onSignalSent with resolvedSignalId, retries via agent.sendSignal() with prepared contents/streamOptions, and falls back to legacy /signals streaming if still unsupported. Non-unsupported errors follow adjusted echo and rethrow behavior.
- Adjusts pending-tool-approval and echo handling to use sendMessage results; keeps legacy paths where needed.

## Documentation

- Added docs for Agent.sendMessage() and Agent.queueMessage() in client-js reference.
- Updated sendSignal docs to direct user-authored input to the new helpers and expanded signal.type union to include 'user' and added optional tagName.
- Guides updated to use sendMessage in examples and note compatibility/fallback behavior for older servers.

## Testing

- client-js tests: new vitest cases for sendMessage and queueMessage request wiring, payload preservation, ifIdle.streamOptions transformation, and runtime option caching.
- React tests: updated useChat tests to mock/assert sendMessage calls and verify per-send and hook-provided clientTools and continuation options.
- Build/test commands included in PR test plan; unrelated pre-existing build errors were observed during implementation.

## Backward Compatibility

- sendSignal remains available; stored signal rows and older clients continue to load via compatibility layers.
- React integration retains fallback to legacy /signals route for older servers.
- Maintains data-user-message stream compatibility and accepts canonical signal metadata type 'user'.

## Files Changed (high level)

- Changesets: .changeset/client-js-message-helpers.md, .changeset/react-message-helpers.md
- Client implementation & types: client-sdks/client-js/src/resources/agent.ts, client-sdks/client-js/src/types.ts
- Client tests: client-sdks/client-js/src/resources/agent.test.ts
- React hook & tests: client-sdks/react/src/agent/hooks.ts, client-sdks/react/src/agent/hooks.test.ts
- Docs: docs/src/content/en/docs/agents/signals.mdx, docs/src/content/en/reference/client-js/agents.mdx


abhiaiyer91:
Only bit is we should call it message and queue to align with the naming exercise we did on Harness v1

CalebBarnes (approved):

TylerBarnes:
@abhiaiyer91 I'm not sure because it can also be used inside processors 

```ts
({ sendSignal }) => {}
```

but we also have `sendNotificationSignal`, `sendStateSignal`, `sendToolApproval`, etc

The idea was "send" prefix was being used to denote anything subscription related, as in you're sending it into the agent loop, and it might not be processed in place like you would get with `stream`/`generate`. In harness v1 we don't really have that issue because there's only 1 way to do it, but on the agent class there's the old way and new way.

Also with harness you have message/queue which are verbs, but with signals many of these are actually nouns and send is the verb

PR #17239, opened 2026-05-28
04 docs: audit message-first signal docs

TylerBarnes:
## Stack position

Stack 4 of 8 in the signals API cleanup.

- Previous PR: #17238
- Next PR: #17240

## Summary

Performs the docs audit for the message-first signal API after the core, server, and SDK helpers are in place.

- Ensures the conceptual signals docs lead with `sendMessage()` / `queueMessage()`
- Keeps `sendSignal()` documented as the lower-level signal API for reactive, state, and notification primitives
- Updates client reference wording around `subscribeToThread()` to mention message helpers alongside signal helpers
- Fixes heading structure in the compatibility section
- Confirms the earlier stacks have package-scoped changesets

## Test plan

- `pnpm run prettier:changed`


TylerBarnes:
Folded into PR #17238 — these were small doc clarifications tied directly to the client message helper APIs, so they belong in that PR. The two doc edits are now part of #17238.

PR #17224, opened 2026-05-28
feat: v1 ToolProvider extensions — backend

YujohnNattrass:
Add the v2 ToolProvider surface used by Agent Builder to manage OAuth-backed integrations.

Surface additions:
- ToolProviderCapabilities (authorize, connections, listToolkits, etc.)
- Auth round-trip on the ToolProvider interface (authorize, getAuthStatus, getConnectionStatus)
- Per-author / shared / caller-supplied connection scoping with resolveConnectionAuthorId() and request-context plumbing
- v2 list/resolve methods (listTools, resolveTools) alongside the v1 catalog

Storage:
- New tool_provider_connections storage domain (in-memory base + LibSQL impl, plus shape additions for ClickHouse / Cloudflare KV)

Server / client:
- Server routes under /tool-providers/* (authorize, auth-status, connection-status, connections list, disconnect, usage, fields, health)
- Matching @mastra/client-js ToolProvider resource methods
- Stored-agent toolProviders config shape (connections, tools) and runtime merge in editor.agent.applyStoredOverrides

Additive — no breaking changes. Studio UI ships separately.

## Type of change

- [ ] Bug fix (non-breaking change that fixes an issue)
- [ ] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [ ] Test update

## Checklist

- [ ] I have linked the related issue(s) in the description above
- [ ] I have made corresponding changes to the documentation (if applicable)
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] I have addressed all Coderabbit comments on this PR

PR #17248, opened 2026-05-28
feat(tool-provider): v1 ToolProvider runtime + server + client SDK — PR 2/3 of #17224 split

YujohnNattrass:
Wires up the v1 ToolProvider runtime, server routes, `@mastra/client-js` resource, and editor integration so stored agents can pin OAuth-backed connections per toolkit and resolve them at run time. The Composio provider is rewritten on the new surface; Studio UI ships separately.

A stored agent's config now carries a `toolProviders` block (`{ providerId: { connections, tools } }`) that the editor merges into the agent definition when it materializes. From the client SDK:

```ts
const composio = client.toolProvider('composio');
const { items } = await composio.listConnections({ toolkit: 'gmail' });
```

`assertModelAllowed` is intentionally dropped from `POST /stored/agents` and `PATCH /stored/agents/:id` as part of the surface-scoped model-policy plan from earlier — the picker-side enforcement is the source of truth and the save-path check was redundant.

A follow-up commit (`eab90298d6`) also hardens the runtime against two soft failures that came out of review: a one-shot warn when caller-supplied connection scope falls back to the shared bucket, and capability mismatches now log-and-skip the offending toolkit instead of throwing and disabling the rest of the providers.

Stacked on #17247; #17249 stacks on this. Replaces #17224.


NikAiyer (approved):

YujohnNattrass:
Thanks @coderabbitai — addressed in 23e2d8cf6a:

- ✅ **Critical: fail closed on unknown connectionId** — disconnect + usage handlers now return 403 when storage is configured but no row matches and the caller isn't an admin.
- ✅ **Major: SDK toolkit forwarding** — `disconnectConnection` now forwards `params.toolkit` and `getConnectionUsage` accepts `{ toolkit }`; added `GetToolProviderConnectionUsageParams` alias.
- ✅ **Minor: authorize label validation** — aligned with stored config rules (`min(1)`, `max(32)`, `/^[A-Za-z0-9 _-]+$/`).
- ⏭️ **Minor: model token placeholders** — declined. The flagged test files already use literal model IDs (`'gpt-4'`, `'gpt-5.5'`, etc.) extensively in their existing fixtures, and the rest of the repo (evals, core stream tests, background-tasks tests) consistently uses literal model names in tests. The remark-model-tokens plugin only operates on docs. Switching just these new lines would be inconsistent with the rest of the file.

Issue #17262, opened 2026-05-29
[FEATURE] Per-session voice scope for realtime/STS agents (Agent voice is a shared mutable singleton)


petrchiriboga:
## Problem Statement

On an `Agent`, every configurable field is request-scoped through
`DynamicArgument<T, TRequestContext>` and resolved per request via a
`get*({ requestContext })` accessor — `instructions`, `tools`, `agents`,
`workflows`, `scorers`, `model`, `memory`, and the processors
(`packages/core/src/agent/agent.ts:311-336`). **Voice is the single exception.**

It is stored as a plain singleton (`#voice: MastraVoice`, `agent.ts:330`; the
config type is `voice?: MastraVoice`, `agent/types.ts:440`), seeded once in the
constructor (`agent.ts:465`). `Agent.getVoice({ requestContext })` then returns
that one shared instance and **mutates it on every call**:

```ts
// packages/core/src/agent/agent.ts:1656-1662
public async getVoice({ requestContext }: { requestContext?: RequestContext } = {}) {
  if (this.#voice) {
    const voice = this.#voice;
    voice?.addTools(await this.listTools({ requestContext }));
    const instructions = await this.getInstructions({ requestContext });
    voice?.addInstructions(this.#convertInstructionsToString(instructions));
    return voice;
  }
  return new DefaultVoice();
}
```

This is fine for one-shot TTS (`speak()` resolves before the next call). It is
**structurally broken for STS / realtime**, where multiple live sessions overlap
on one deployed agent. The realtime providers store everything per-instance, so
a shared instance has exactly one slot for each piece of session state. For
`@mastra/voice-google-gemini-live`:

- `addTools(tools)` → `this.tools = tools` — a full overwrite, not a merge
  (`voice/google-gemini-live-api/src/index.ts:1965-1968`).
- `addInstructions(s)` → `this.options.instructions = s` (`index.ts:2082`).
- `ws`, `state`, `contextManager`, `sessionId`, `requestContext` are single
  per-instance fields (`index.ts:105-132`) — one WebSocket and one of each,
  shared across every concurrent caller.

So when session B starts while session A is still live, B's `getVoice()` call
overwrites A's tools, instructions, and request context on the shared object —
and they share one ws. This affects every realtime provider, not just Gemini:
`@mastra/voice-openai-realtime`, `@mastra/voice-xai-realtime`, and
`@mastra/voice-aws-nova-sonic` follow the same single-instance shape.

## Proposed Solution

[...]

PR #17267, opened 2026-05-29
feat(core): persist realtime voice turns to agent memory

### LINKED ISSUE #17264: [FEATURE] Persist realtime voice transcripts to Memory
## Problem Statement

`agent.generate()` and `agent.stream()` automatically write user/assistant turns to `mastra_messages` through the agent's configured Memory (`packages/core/src/agent/agent.ts:1416` → `Memory.saveMessages()`). Realtime voice runs through a disjoint path — the provider's WebSocket handler — and turns reach only a per-session, in-memory context window. For example, `GeminiLiveVoice.addToContext()` writes to `this.contextManager` (`voice/google-gemini-live-api/src/index.ts:1060`), never to Memory.

As a result a voice conversation leaves **no persistent history**: it isn't recallable, can't be resumed, and doesn't sit alongside text turns in the same thread. The abstract `MastraVoice` surface (`packages/_internals/voice/src/voice/voice.ts`) has zero memory-aware methods, and the agent — which owns `#voice` (`agent.ts:330`) and already configures it in `getVoice()` (`agent.ts:1656`) — never subscribes to its events. Because recent work extracted voice internals out of core and forbids voice provider packages from importing core, any bridge must live in core (on `Agent`), not in providers.

## Proposed Solution

_(proposal for maintainer feedback, not a finished design)_

A persistence bridge owned by `Agent`, consuming a standardized voice event:

1. **A payload-bearing turn event on the voice contract.** Add a completed-turn event to `VoiceEventMap` in `@internal/voice` with a provider-safe, framework-agnostic shape (e.g. `{ role: 'user' | 'assistant', text: string }`). The existing chunked `writing` event is unchanged. The event is **optional** — providers that don't emit it simply don't persist, so no provider is forced to change.

2. **Agent-side bridge.** `getVoice()` gains an optional memory scope, mirroring `generate()`:

   ```ts
   const voice = await agent.getVoice({ memory: { thread: 't1', resource: 'u1' } });
   await voice.connect();
   // completed turns now persist to the same thread agent.generate() would use
   ```

   When a scope is supplied, the agent subscribes to the turn event and, per completed turn, calls its own `Memory.saveMessages()` (engaging semantic-recall indexing). The thread is auto-created on the first turn; `resourceId` is required. With no scope, nothing persists — fully non-breaking.

3. **Provider emit (Gemini).** `GeminiLiveVoice` emits the event on turn completion (it already detects this boundary at `index.ts:1484`; today's `turnComplete` event carries only a timestamp). Text-in user turns (`index.ts:597`) and accumulated assistant text turns are both available cleanly today.

**Scope:** realtime WS sessions only; one-shot `speak()`/`listen()` are stateless and out of scope. Persistence is live per-turn; an interrupted (barge-in) turn is persisted as-final.

**Not covered here** (called out so the boundary is explicit, not as a commitment): tool calls/results, working-memory updates, providers beyond Gemini, and transcription of spoken user audio (see Additional Context).

## Component

Memory, Agents, Other (Voice)

## Alternatives Considered

- _Memory methods on `MastraVoice`_ — enlarges the provider surface and pulls memory concepts toward the provider layer.
- _Standalone adapter (not Agent)_ — adds a new component yet still needs the thread/resource scope the agent already owns.
- _Buffer chunked `writing` + standardize the existing textless `turnComplete`_ — pushes turn-reassembly state (chunk interleaving, interrupts) into core; a payload-bearing event removes that.

## Example Use Case

> As a developer building a voice agent, I want spoken/typed conversations saved to the same Memory thread as my text conversations, so a user can start a task by voice and continue it in chat (and vice versa) with shared history and semantic recall — without manually wiring transcript capture into `Memory.saveMessages()` myself.

## Additional Context

**Possible extensions.** The turn event payload could generalize to structured parts — `{ role, parts: Array<{type:'text'} | {type:'tool-call'; toolName; args; id} | {type:'tool-result'; id; result}> }` — to reach full `generate()` parity (persisting tool calls/results), with the agent mapping parts to `MastraMessageV2`; the same optional event would let OpenAI Realtime and other providers participate. The `{ role, text }` shape above is a forward-compatible subset.

**Spoken-audio user transcripts.** Gemini Live supports input transcription (JS SDK: `inputAudioTranscription` in session config; delivered at `serverContent.inputTranscription.text`), but this integration never enables it — the setup message type (`index.ts:1722`) has no `input_audio_transcription` field and `handleServerContent` (`:1363`) only parses assistant `modelTurn.parts`. So a pure-audio session currently produces no user-role transcript; text-in user turns and assistant text-out are what's available. Enabling input transcription (add the setup field + handle the new server message) is a small, self-contained piece of provider work that this feature would build on for the spoken-user case.

**Grounding (current `main`).** Abstract surface `packages/_internals/voice/src/voice/voice.ts`; Gemini context-only writes `voice/google-gemini-live-api/src/index.ts:1060`, turn boundary `:1484`, text-in user turn `:597`; agent memory path `agent.ts:1416` and `agent-legacy.ts:400`; agent owns `#voice` `agent.ts:330` and configures it in `getVoice()` `agent.ts:1656`. Import boundary: the Gemini provider imports only `@internal/voice`, so the new event type is defined there; mapping to `MastraMessageV2` stays in core.

nehaprasad-dev:
#### Description
- Persist realtime voice turns to Memory via getVoice({ memory: { thread, resource } }).

#### Related issue(s)
Fixes #17264

#### Type of change
- Test update

#### Checklist

## ELI5

This PR lets voice conversations be saved and remembered by agents, just like text conversations already are. When someone talks to an agent using voice, their back-and-forth exchanges are now stored in the agent's memory so they can be referenced later, and developers can opt into this feature by passing memory details when setting up voice.

---

## Overview

This PR introduces voice turn persistence to the agent memory system, bridging realtime voice conversations with the persistent memory layer. A new `turn` event is added to the voice event contract, carrying provider-safe payloads (`{ role: 'user' | 'assistant', text: string }`). The `Agent.getVoice()` method is extended to accept an optional `memory` scope, allowing completed voice turns to be automatically persisted to agent memory. Thread creation is automatic on first turn, and `resourceId` is required when memory persistence is enabled.

---

### Voice Contract & Types

**`packages/_internals/voice/src/voice/voice.ts`**
- Added `VoiceTurnRole` type (`'user' | 'assistant'`)
- Added `VoiceTurnEvent` interface with `role` and `text` fields
- Extended `VoiceEventMap` with new `turn` event type

**`packages/core/src/voice/index.ts`**
- Re-exported new types `VoiceTurnEvent` and `VoiceTurnRole` from `@internal/voice`

### Agent Voice Memory Integration

**`packages/core/src/agent/agent.ts`**
- Extended `Agent.getVoice()` method signature to accept optional `memory` parameter (`{ thread, resource }`)
- Implemented voice turn persistence logic:
  - Registers `voice.on('turn', ...)` handler when memory scope is provided
  - Converts each completed voice turn to a `MastraDBMessage` and saves via `Memory.saveMessages()`
  - Auto-creates memory thread on first turn with provided IDs
  - Tracks voice instance and handler via private `#voiceMemoryPersistence` field to prevent duplicate listeners

**`packages/core/src/agent/durable/durable-agent.ts`**
- Updated `getVoice()` override to accept and forward optional `options` parameter to wrapped agent
- Maintains parity with base `Agent.getVoice()` signature

### Provider Implementation

**`voice/google-gemini-live-api/src/index.ts` & `voice/google-gemini-live-api/src/types.ts`**
- Implemented `turn` event emission in `GeminiLiveVoice`:
  - Emits `turn` event with user text when `speak()` is called
  - Accumulates assistant response text during streaming
  - Emits `turn` event with complete assistant text when `turnComplete` is received
- Updated `GeminiLiveEventMap` to include `turn` event type

### Testing

**`packages/core/src/agent/__tests__/voice-memory.test.ts`** (New)
- Added comprehensive test suite for voice memory persistence:
  - Validates that completed user voice turns are saved to memory with correct threading
  - Confirms `MockMemory.createThread` is called with `saveThread: true`
  - Tests assistant turn persistence via separate save call
  - Validates error handling for missing `memory.resource` parameter

---

## Implementation Details

The feature is entirely opt-in; providers that do not emit the new `turn` event remain unaffected. Thread creation happens automatically on the first voice turn when a memory scope is provided, and the `resourceId` is required to establish message ownership. The voice memory persistence handler is managed to prevent multiple listeners on the same voice instance, with automatic cleanup when switching between different voice instances.

---

## Backward Compatibility

This is a non-breaking change. The `memory` parameter is optional in `Agent.getVoice()`, and existing voice provider implementations continue to work unchanged. Providers are not required to implement the new `turn` event, though doing so enables voice history persistence for agents that opt in.

[Review Change Stack](https://app.coderabbit.ai/change-stack/mastra-ai/mastra/pull/17267?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/packages/_internals/voice/src/voice/voice.ts
+++ b/packages/_internals/voice/src/voice/voice.ts
@@ -3,10 +3,18 @@ import type { ToolsInput } from '@internal/core/types';
 
 export type VoiceEventType = 'speaking' | 'writing' | 'error' | string;
 
+export type VoiceTurnRole = 'user' | 'assistant';
+
+export interface VoiceTurnEvent {
+  role: VoiceTurnRole;
+  text: string;
+}
+
 export interface VoiceEventMap {
   speaker: NodeJS.ReadableStream;
   speaking: { audio?: string };
   writing: { text: string; role: 'assistant' | 'user' };
+  turn: VoiceTurnEvent;
   error: { message: string; code?: string; details?: unknown };
   [key: string]: unknown;
 }
--- a/packages/core/src/agent/__tests__/voice-memory.test.ts
+++ b/packages/core/src/agent/__tests__/voice-memory.test.ts
@@ -1,0 +1,124 @@
+import { MockLanguageModelV2 } from '@internal/ai-sdk-v5/test';
+import { describe, expect, it, vi } from 'vitest';
+import { MockMemory } from '../../memory/mock';
+import { MockStore } from '../../storage/mock';
+import type { MastraVoice, VoiceTurnEvent } from '../../voice';
+import { Agent } from '../agent';
+
+function createMockVoice(): MastraVoice & {
+  emitTurn: (turn: VoiceTurnEvent) => void;
+} {
+  const listeners = new Map<string, Set<(data: unknown) => void>>();
+
+  return {
+    addTools: vi.fn(),
+    addInstructions: vi.fn(),
+    speak: vi.fn(),
+    listen: vi.fn(),
+    updateConfig: vi.fn(),
+    connect: vi.fn(),
+    send: vi.fn(),
+    answer: vi.fn(),
+    close: vi.fn(),
+    getSpeakers: vi.fn().mockResolvedValue([]),
+    getListener: vi.fn().mockResolvedValue({ enabled: false }),
+    serializeForSpan: vi.fn().mockReturnValue({ component: 'VOICE' }),
+    on: vi.fn((event: string, callback: (data: unknown) => void) => {
+      if (!listeners.has(event)) {
+        listeners.set(event, new Set());
+      }
+      listeners.get(event)!.add(callback);
+    }),
+    off: vi.fn((event: string, callback: (data: unknown) => void) => {
+      listeners.get(event)?.delete(callback);
+    }),
+    emitTurn: (turn: VoiceTurnEvent) => {
+      for (const callback of listeners.get('turn') ?? []) {
+        callback(turn);
+      }
+    },
+  } as MastraVoice & { emitTurn: (turn: VoiceTurnEvent) => void };
+}
+
+describe('Agent voice memory persistence', () => {
+  it('persists completed voice turns when memory scope is provided', async () => {
+    const mockVoice = createMockVoice();
+    const memoryStore = new MockStore();
+    const memory = new MockMemory({ storage: memoryStore });
+
+    const saveMessages = vi.spyOn(memory, 'saveMessages');
+    const createThread = vi.spyOn(memory, 'createThread');
+
+    const agent = new Agent({
+      id: 'voice-memory-agent',
+      name: 'Voice Memory Agent',
+      instructions: 'You are helpful.',
+      model: new MockLanguageModelV2({
+        doGenerate: async () => ({
+          finishReason: 'stop',
+          usage: { inputTokens: 1, outputTokens: 1, totalTokens: 2 },
+          content: [{ type: 'text', text: 'ok' }],
+          warnings: [],
+        }),
+      }),
+      memory,
+      voice: mockVoice,
+    });
+
+    const voice = await agent.getVoice({
+      memory: { thread: 'thread-voice-1', resource: 'user-1' },
+    });
+
+    mockVoice.emitTurn({ role: 'user', text: 'Hello from voice' });
+    await vi.waitFor(() => expect(saveMessages).toHaveBeenCalled());
+
+    expect(createThread).toHaveBeenCalledWith(
+      expect.objectContaining({
+        threadId: 'thread-voice-1',
+        resourceId: 'user-1',
+        saveThread: true,
+      }),
+    );
+
+    const savedMessages = saveMessages.mock.calls[0]![0].messages;
+    expect(savedMessages).toHaveLength(1);
+    expect(savedMessages[0]).toMatchObject({
+      role: 'user',
+      threadId: 'thread-voice-1',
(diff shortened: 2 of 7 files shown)


petrchiriboga:
Heads up — #17345 ("per-request voice resolution on agents", merged May 31, `Fixes #17262`) landed after this branch and reworks the same `getVoice()` / `voice` config surface this PR builds on. Two implications:

1. **Rebase needed** — `voice` is now `DynamicArgument<MastraVoice>` and `getVoice()` resolves a fresh, session-owned instance per call. This branch still has `voice?: MastraVoice` / `#voice: MastraVoice`, so it'll conflict in `getVoice()` and `agent/types.ts`.

2. **Persistence should move onto the resolved per-session instance.** Currently it lives in a single agent-level slot (`#voiceMemoryPersistence`) and `getVoice()` detaches the previous handler each call. With #17345 each session resolves its own voice, so a shared slot means only the most recent `getVoice()` call keeps a `turn` listener — earlier concurrent sessions stop persisting. Suggest attaching the `turn`→`saveMessages` handler to the instance returned by the resolver, scoped to that call's `thread`/`resource`, with teardown on the caller's `disconnect()`.

FWIW the core bridge logic works — I built the branch and a `turn`→`saveMessages` round-trip persists and reads back correctly with a static voice. It's specifically the multi-tenant/per-session path that the shared-slot approach doesn't cover.

intojhanurag:
closing this ! we will handle this by own :)

Issue #17284, opened 2026-05-29
mastra_workflow_snapshot rows grow to 50+ MB with HITL suspend/resume --> causes Postgres CPU spike


daneatmastra:
This issue was created from Discord post 1509895368081539123:

[![Open in Discord](https://img.shields.io/badge/Open_in_Discord-5865F2?style=for-the-badge&logo=discord&logoColor=white)](https://discord.com/channels/1309558646228779139/1509895368081539123)

**Context**
We're building a multi-turn financial advice agent using:
- Harness with Observational Memory (OM)
- agentic-loop → executionWorkflow architecture
- HITL via suspend/resume on an askUser tool — the agent suspends, the user answers, the workflow resumes on the same run_id
- The agent calls many tools with large payloads under the hood, large outputs are shrunk using `toModelOutput`
- PostgreSQL storage (@mastra/pg)

A typical session involves 20+ askUser suspensions

**Problem**
Our mastra_workflow_snapshot table has 42 rows totalling 350 MB. Individual rows reach 45 MB compressed / 91 MB uncompressed for a single conversation. On a modest DB (1 vCPU / 1 GB RAM), every resume request causes a 100% CPU spike and takes 3–99 seconds per snapshot query, leading to ~82-second delays before the agent resumes after each user answer. Obviously this is completely breaking for our use case.

We profiled the snapshots in detail using pg_column_size queries and direct JSON analysis and we found that the AI SDK result object (~7.7 MB) is stored 8 times across pipeline stages because each step stores both payload (input) and output. Total: ~54 MB on a single object.
*(maybe i did smth wrong here, but I believe there is for sure duplication in the jsonb data)*

**The main questions**
1. Do you at Mastra recommend not doing HITL for many turns? Or should we rework to work with user messages?
2. Is this a known issue? The docs state to keep snapshots for workflows small, but the harness dumps the entire messageList into the snapshot, so it would require me to filter the entire messageList.

daneatmastra:
Thank you for reporting this issue! We have labeled it for trio-tnt and we will look into it as soon as possible.

If you're experiencing an error, please provide a [minimal reproducible example](https://github.com/mastra-ai/mastra/blob/main/CONTRIBUTING.md#minimal-reproduction) whenever possible to help us resolve it quickly.

frits-kvdb:
Probably good to know for this issue: it is `executionWorkflow.context` which is growing like crazy.
Example from my local db:

```
        key         |  bytes   |    size
[...]

PR #17296, opened 2026-05-29
fix(core): reduce workflow snapshot duplication

mbenhamd:
## Description

This reduces workflow snapshot growth during agentic-loop HITL suspend/resume runs.

The issue was that AI SDK step results kept cumulative `response.messages` in each step result, and workflow snapshots persisted those cumulative histories repeatedly across execution workflow context entries. In long HITL sessions, this caused the workflow snapshot JSON's `context` field to grow very large and made resume reads expensive.

**How it works.** Persisted workflow snapshots now store AI SDK step results with only the response messages added by that step, via a `Symbol.for`-keyed serializer hook on `DefaultStepResult` (`Symbol.for` keeps the hook working across duplicated copies of `@mastra/core` and never collides with user-defined properties). The runtime/persistence boundary is explicit:

- `mergeWorkflowStepResult` writes the serialized view into the snapshot that storage adapters persist, but returns the raw merged result for engine runtime state, so serialized deltas never replace live values mid-run.
- The agentic loop rebuilds the cumulative per-step view (`restoreCumulativeStepResponseMessages`) before LLM request processing **and** before emitting `step-finish`/`finish` stream chunk payloads — on the evented engine the loop body's output round-trips through snapshot persistence between iterations, so stream consumers would otherwise see deltas.
- Resumed runs restore the cumulative view from the rehydrated message history by content matching; when matching fails (processor-mutated messages, trimmed history), steps conservatively keep their stored messages.

Resume edge cases fixed along the way:

- explicit falsy resume payloads like `false`, `0`, `null`, and `""` are preserved and validated
- evented step execution receives explicit resume state through the worker/server path via a new optional `isResuming` body field (additive; old clients and servers are unaffected)
- evented foreach failure handling no longer treats user output shaped like `{ status: "failed" }` as a failed step result

## Behavioral changes (beyond snapshot size)

- **Foreach merge now requires the internal `__mastra_foreach__` marker** for element-wise output merging. Plain array step outputs replace wholesale; previously any array output was element-merged.
- **dsql, dynamodb, mssql, mysql, redis and spanner switched from last-write-wins to the shared element-wise merge** in `updateWorkflowResults`, so concurrent foreach iterations no longer clobber each other on those stores.
- **A failed foreach iteration fails the workflow immediately** (persists `failed`, publishes `workflow.fail` from the step-end path). Late-arriving successful iterations still fill their output slot for observability; late suspends are dropped and cannot resurrect a failed run.
- **Terminal runs drop late step-end events** (except the fill case above).
- **Bulk foreach resume** now writes per-index pending-marker resets through the storage layer's atomic merge (previously one bulk write built from a possibly-stale read), and on a failed resume-event publish it restores the suspended slot and rethrows instead of swallowing the error.
- **Fresh (non-resume) step executions no longer inherit fields from a prior result** (`resumePayload`, `suspendPayload`, stale `error` from a retried attempt). Resumed executions keep the original `payload`.
- Workflow snapshot values are serialized with JSON semantics before persistence: `toJSON` is honored, functions are dropped (objects) or nulled (arrays), circular references are stored as `"[Circular]"` instead of failing the write, and `Date` instances are passed through for the store driver to handle. In-memory storage now holds the serialized form too, matching persistent stores.
- `__mastra_foreach_completed_indexes__` is internal bookkeeping persisted inside in-flight foreach step results; it is stripped when the foreach completes and from failure results/events.

## Known limitations

- Bloated `suspendPayload` is **not** fully solved here: default-engine foreach can still store `__workflow_meta.foreachOutput`, and evented foreach intentionally preserves suspend payload fields such as stream state. If remaining snapshot growth comes from suspended approval payloads, that is a separate follow-up.
- Stores that merge server-side (MongoDB aggregation pipeline, Upstash Lua, Convex server runtime) mirror the core merge logic. Their `updateWorkflowResults` return hands runtime callers the raw step result for non-foreach entries (`withRuntimeStepResult`), matching core semantics; foreach entries keep the server-merged stored view, which the agentic loop's restore handles. The MongoDB merge pipeline uses `$getField` and therefore requires MongoDB 5.0+.

## Performance

Ad-hoc local Postgres measurement using the repo PG16 perf container, one `jsonb` column, synthetic snapshots shaped after the duplicated cumulative `response.messages` pattern from issue #17284, 25 measured samples per old/new side after 5 warmups. Old = pre-fix cumulative step result objects; new = current snapshot serialization path.

| Iterations | Approx msg size | Stage copies | Old JSON | New JSON | Old `jsonb` column | New `jsonb` column | Column reduction | Old insert p50 | New insert p50 | Insert speedup | Old read p50 | New read p50 | Read speedup | Insert p95 old/new | Read p95 old/new |
|---:|---:|---:|---:|---:|---:|---:|---:|---:|---:|---:|---:|---:|---:|---:|---:|
| 20 | 2 KB | 4 | 4.07 MB | 0.44 MB | 0.15 MB | 0.02 MB | 8.86x | 33.50 ms | 4.24 ms | 7.89x | 29.70 ms | 3.26 ms | 9.11x | 36.80 / 4.98 ms | 34.59 / 4.50 ms |
| 20 | 8 KB | 4 | 16.38 MB | 1.61 MB | 0.49 MB | 0.10 MB | 4.79x | 128.08 ms | 13.47 ms | 9.51x | 122.47 ms | 10.33 ms | 11.85x | 135.78 / 15.03 ms | 129.78 / 11.30 ms |
| 50 | 2 KB | 4 | 24.51 MB | 1.10 MB | 0.67 MB | 0.04 MB | 17.30x | 196.71 ms | 10.09 ms | 19.49x | 176.01 ms | 7.42 ms | 23.71x | 207.88 / 13.17 ms | 200.10 / 8.64 ms |

Notes: this is a synthetic local microbenchmark of the duplicated-messages pattern — directional only, not an end-to-end production HITL resume latency measurement. `jsonb` sizes use `pg_column_size(context)`; timings are local p50/p95 wall-clock for insert into `jsonb` and `select context` through the Node `pg` client. Cost side: serialization adds a pre-pass on each step persist (same order of work as the JSON stringify stores already do), and the evented step-end handler performs its own snapshot read (in addition to the dispatch-level one) so terminal checks and foreach merge bases are as fresh as possible.

## Related issue(s)

Mitigates #17284 — fixes the dominant growth source reported there (duplicated cumulative `response.messages` across step results). Suspend-payload size for suspended approval flows is tracked separately (see Known limitations).

## Type of change

- [ ] Documentation update
- [ ] Code refactoring
- [x] Performance improvement
- [x] Test update

## Checklist


mbenhamd:
Benchmark addendum after the foreach merge/data-loss fixes:

The original Postgres table above still describes the snapshot-size issue from #17284: persisted AI SDK step results no longer duplicate cumulative message history, reducing `jsonb` size and read/insert cost.

I also reran a separate production-code benchmark for the foreach merge path added while hardening this PR. This benchmark imports the real workflow snapshot merge code; on this PR branch it also uses the real evented foreach marker helper. It replays a stale concurrent sibling write after one foreach iteration completed with a legitimate `null` output.

| Foreach slots | origin/main data-safe? | PR data-safe? | origin/main stale result | PR stale result | origin/main stale sibling us/op | PR stale sibling us/op | PR final JSON bytes |
|---:|---:|---:|---|---|---:|---:|---:|
| 100 | No | Yes | `["stale-old-0", "done-1", ...]` | `[null, "done-1", ...]` | 46.707 | 124.025 | 4,137 |
| 1,000 | No | Yes | `["stale-old-0", "done-1", ...]` | `[null, "done-1", ...]` | 476.100 | 1172.610 | 43,737 |
| 5,000 | No | Yes | `["stale-old-0", "done-1", ...]` | `[null, "done-1", ...]` | 2449.435 | 6393.091 | 231,737 |

Interpretation: the PR adds CPU work in this narrow merge path because it tracks completed foreach indexes and rejects stale sibling overwrites. That is intentional correctness work. The previous behavior is faster in this microbenchmark because it does not preserve a completed `null` value and can resurrect stale data. The PR preserves the user-visible output and cleans internal completed-index metadata before the final snapshot result is stored.

Focused local validation after this benchmark:

- `pnpm --filter ./packages/core exec vitest run src/storage/workflow-snapshot.test.ts src/workflows/evented/evented-workflow.test.ts src/workflows/evented/step-executor.test.ts`
  - `288 passed | 5 skipped`, no type errors
- `pnpm --filter ./stores/convex exec vitest run src/server/storage.test.ts src/storage/domains/workflows/workflows.test.ts`
  - `69 passed`
- `pnpm --filter ./stores/mongodb exec vitest run src/storage/index.test.ts`
  - `481 passed | 39 skipped`
- `pnpm --filter ./stores/upstash exec vitest run src/storage/index.test.ts`
  - `230 passed | 269 skipped`

PR #17305, opened 2026-05-30
fix(evals): read tool calls from V2 content.parts in extractToolCalls and extractToolResults

serhiizghama:
## Problem

`extractToolCalls()` and `extractToolResults()` in `@mastra/evals` only read `message.content.toolInvocations` when extracting tool data for scorers.

This misses valid tool calls when an assistant message stores them only in V2 `content.parts` as `tool-invocation` entries — which is the format used when observable memory is enabled.

This is the same class of bug as #15438 / PR #15439, which fixed `extractTrajectory()` in `@mastra/core` but did not address the equivalent functions in `@mastra/evals`.

### Impact

When observable memory is enabled:
- **Hallucination scorer** with `getContext` using `extractToolResults()` always receives empty context → scores all responses as 100% hallucinated (score=1.0)
- **Tool-usage scorer** using `extractToolCalls()` always sees 0 tools → scores all tool-using responses as 0

## Solution

Applied the same fallback pattern already used in `extractTrajectory()` (from PR #15439) to both functions: prefer `content.toolInvocations` when present, fall back to `content.parts` filtered for `tool-invocation` entries when not.

## Testing

Added regression tests covering:
- `extractToolResults`: V2 parts fallback, precedence over `toolInvocations`
- `extractToolCalls`: legacy path, V2 parts fallback, multiple V2 parts, precedence, empty case

Existing tests continue to pass unchanged.

## ELI5
The code that scores AI responses was looking for tool usage information in the wrong place when a feature called "observable memory" was turned on. This PR fixes two functions to look in both the old location AND the new location for that tool information, so the scoring works correctly.

### Updated Functions: `extractToolCalls` and `extractToolResults`

Both functions in `packages/evals/src/scorers/utils.ts` now support dual formats for reading tool invocation data:

**`extractToolCalls`**: Now extracts tool calls from two sources:
- Prefers legacy `message.content.toolInvocations` array when present
- Falls back to filtering `message.content.parts` for `tool-invocation` entries and mapping them to the `toolInvocation` format
- Records tool names and generates default `toolCallId` values (`${messageIndex}-${invocationIndex}`) when not provided
- Includes tool calls in both 'result' and 'call' states

**`extractToolResults`**: Similarly updated to extract tool results:
- Prefers legacy `message.content.toolInvocations` array when present
- Falls back to extracting from `message.content.parts` entries of type `tool-invocation`
- Only includes invocations with `state === 'result'` and a defined `result` property

Both functions apply the same fallback pattern already used in `extractTrajectory()`, ensuring consistent handling across the codebase.

### Test Coverage

Comprehensive regression tests were added to `packages/evals/src/scorers/utils.test.ts`:

**`extractToolResults` tests**:
- Extract tool results from V2 content.parts when legacy toolInvocations is absent
- Prefer legacy toolInvocations over content.parts when both are present

**`extractToolCalls` tests**:
- Extract from legacy toolInvocations (existing behavior)
- Extract from V2 content.parts when toolInvocations is absent
- Extract multiple tool calls within a single message (including non-result states like 'call')
- Prefer legacy toolInvocations over parts when both exist
- Return empty arrays when no tool calls are present

### Changeset

A patch-level changeset for `@mastra/evals` documents the fix for improved transparency in release notes.

## Impact

Fixes scoring bugs where:
- The hallucination scorer received empty tool context and incorrectly scored all responses as fully hallucinated
- The tool-usage scorer reported zero tools when observable memory was enabled

The changes enable proper scoring when messages use V2 content format (the format used by observable memory) while maintaining backward compatibility with the legacy format.

[Review Change Stack](https://app.coderabbit.ai/change-stack/mastra-ai/mastra/pull/17305?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/.changeset/spicy-jokes-nail.md
+++ b/.changeset/spicy-jokes-nail.md
@@ -1,0 +1,5 @@
+---
+'@mastra/evals': patch
+---
+
+Fixed `extractToolCalls` and `extractToolResults` in `@mastra/evals` to also read tool invocations from V2 `content.parts` when `toolInvocations` is absent. Previously both functions only checked `message.content.toolInvocations`, which missed tool calls stored in `content.parts` as `tool-invocation` parts — the format used when observable memory is enabled. This caused hallucination and tool-usage scorers to receive empty tool data, producing incorrect scores.
--- a/packages/evals/src/scorers/utils.test.ts
+++ b/packages/evals/src/scorers/utils.test.ts
@@ -15,6 +15,7 @@ import {
   getReasoningFromRunOutput,
   createTestMessage,
   createToolInvocation,
+  extractToolCalls,
   extractToolResults,
   extractTrajectory,
   compareTrajectories,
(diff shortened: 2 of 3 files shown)


abhiaiyer91:
Closed and addressed

Issue #17012, opened 2026-05-24
Agents chat not working in studio


daneatmastra:
This issue was created from Discord post 1508223433677209902:

[![Open in Discord](https://img.shields.io/badge/Open_in_Discord-5865F2?style=for-the-badge&logo=discord&logoColor=white)](https://discord.com/channels/1309558646228779139/1508223433677209902)

Hi, the agents chat UI is not working for me. I type a message and I get no response, no errors, nothing

Current versions:
```
@mastra/client-js@1.13.2
@mastra/core@1.35.0
@mastra/duckdb@1.3.0
@mastra/hono@1.4.2
@mastra/libsql@1.8.0
@mastra/loggers@1.1.1
@mastra/mcp@1.4.1
@mastra/memory@1.15.0
@mastra/observability@1.12.0
mastra@1.9.0
```

The screenshot attached is what I see after typing "hello" in the chatbox and pressing "enter"

---

![Screenshot 1](https://cdn.discordapp.com/attachments/1508223433677209902/1508223434008563863/Screenshot_2026-05-24_at_5.40.58_PM.png?ex=6a14c208&is=6a137088&hm=3bf7f14dcbb23061ffd94f0f9d793f111e910a84cdaf2745836596192221e041&)

daneatmastra:
Thank you for reporting this issue! We have labeled it for trio-tb and we will look into it as soon as possible.

If you're experiencing an error, please provide a [minimal reproducible example](https://github.com/mastra-ai/mastra/blob/main/CONTRIBUTING.md#minimal-reproduction) whenever possible to help us resolve it quickly.

Praanjaljain29:
Please assign this issue to me

intojhanurag:
It's a version-skew issue: mastra@1.9.0's bundled Studio UI sends a request body that @mastra/core@1.35.0 rejects (HTTP 400 "Invalid input"). Bump both to latest together — mastra@1.10.0 + @mastra/core@1.36.0 — and it works (tested with your exact other deps). Upgrade mastra and @mastra/core in lockstep. Closing as resolved.

Issue #16444, opened 2026-05-11
Issue with MCP Tools with “cost”-constrained fields


daneatmastra:
This issue was created from Discord post 1503449123033059441:

[![Open in Discord](https://img.shields.io/badge/Open_in_Discord-5865F2?style=for-the-badge&logo=discord&logoColor=white)](https://discord.com/channels/1309558646228779139/1503449123033059441)

Issue: MCP tools with const-constrained fields fail validation

MCP API specs commonly use @type discriminator fields (e.g. "@type": { "const": "com.cvent.EventSpec" }) to guide server-side deserialization into the correct Java/typed model. These are marked required in the JSON schema. The LLM either omits them or provides the wrong value — it has no reason to know the correct constant. validateToolInput fails with "eventSpec. @type: must be equal to constant" and the tool never executes. As a workaround we can recursively strip const-constrained fields from required arrays in the schema, then inject the correct constant values in the execute wrapper but that's essentially a hack and is super fragile.

Proposed fix for Mastra: Add a step to the validateToolInput pipeline (in packages/core/src/tools/validation.ts) that uses the already-imported standardSchemaToJSONSchema to find all const-constrained properties and auto-inject their values before validation runs. Since a const field has exactly one valid value, the LLM can never "choose" correctly — Mastra should own this injection. This would fix both issues generically for all tools.

daneatmastra:
Thank you for reporting this issue! We have labeled it for trio-tb and we will look into it as soon as possible.

If you're experiencing an error, please provide a [minimal reproducible example](https://github.com/mastra-ai/mastra/blob/main/CONTRIBUTING.md#minimal-reproduction) whenever possible to help us resolve it quickly.

mikhael28:
Asked the user on Discord if they would be interested in submitting a PR/reproducible example for us to take a look at - will be interested to see what their response will be!

daneatmastra:
## 💬 Discord Activity

<details>
<summary><strong>3 messages</strong> synced from Discord</summary>

<details>
<summary><a href="https://discord.com/channels/1309558646228779139/1503449123033059441/1503449123033059441">zacdavidsonatx</a> · May 11</summary>

Issue: MCP tools with const-constrained fields fail validation

[...]

PR #17308, opened 2026-05-30
fix(observability): OpenRouter pricing miss for dotted-version models

### LINKED ISSUE #17189: [BUG] OpenRouter vendor-prefixed models with a dotted version never match the pricing registry (estimatedCost null)
## Describe the Bug

When an Agent calls a model through the OpenRouter gateway, the `MODEL_GENERATION` span records the model as the full route id (e.g. `google/gemini-2.5-flash`). The embedded pricing registry keys that model as `gemini-2-5-flash` — vendor prefix dropped, dots flattened to dashes. For any route whose version contains a dot, `estimateCosts` never finds a match and returns `no_matching_model`, so `estimatedCost` is null and Studio's "Total Model Cost" / "Model Usage & Cost" panels stay empty even though token metrics are recorded fine.

`getModelVariants()` (`observability/mastra/src/metrics/pricing-registry.ts`) does try both transforms, just never in the same pass:

- flatten dots → `google/gemini-2-5-flash` (vendor prefix still there)
- strip vendor → `gemini-2.5-flash` (dots still there)

Neither equals `gemini-2-5-flash`. PR #16206 added the vendor-strip variant, which fixed dotless ids like `xiaomi/mimo-v2-pro`, but the dotted-version case slipped through. This hits most Google/Gemini routes, `meta-llama/llama-3.1-*`, and similar.

#16281 reported this exact case and was closed as a duplicate; this issue narrows it to the still-unfixed dotted-version variant and adds a runnable repro. It's a different root cause from #16584 (word ordering on Anthropic routes).

## Steps To Reproduce

Deterministic, no API key (reads the pricing data shipped inside the package):

1. `npm install` in the linked repro
2. `node repro.mjs`
3. Observe: the bundled `pricing-data.jsonl` contains `openrouter::gemini-2-5-flash`, but the variants generated for `google/gemini-2.5-flash` are `[google/gemini-2.5-flash, google/gemini-2-5-flash, google-gemini-2-5-flash, gemini-2.5-flash]` — none equal `gemini-2-5-flash`, so the lookup misses.

End-to-end (with a key): an Agent configured with `model: 'openrouter/google/gemini-2.5-flash'` plus an OLAP observability store; after a `generate()` call the model-cost metric has `estimatedCost` null while token metrics are populated.

https://github.com/domuk-k/mastra-openrouter-cost-repro

## Expected Behavior

`estimateCosts({ provider: 'openrouter', model: 'google/gemini-2.5-flash', ... })` resolves the `gemini-2-5-flash` pricing row and returns a non-null `estimatedCost`, so Studio's "Total Model Cost" reflects OpenRouter usage.

## Environment Information

```
@mastra/observability: 1.14.0
@mastra/core: 1.37.0
Node: 22.19.0
OS: macOS 26.2
LLM Provider: OpenRouter (google/gemini-2.5-flash)
Observability store: ClickHouse (also reproducible against the in-package pricing data, see repro)
```

## Verification

- [x] I have searched the existing issues to make sure this is not a duplicate (references #16281 / #16584 above)

SaarthurR:
`getModelVariants()` never combined vendor-prefix stripping and dot-to-dash normalization in the same pass. For a model like `google/gemini-2.5-flash`, it produced `gemini-2.5-flash` (dots intact) as the vendor-stripped variant — but the registry stores it as `gemini-2-5-flash`. Added one line to also generate the dot-flattened form of the after-slash substring.

Closes #17189

## ELI5 (Explain Like I'm 5)

The code missed one of the alternate names the billing system uses for some models (it didn't turn dots into dashes after removing the vendor name), so prices couldn't be found. This PR adds that missing name so the correct price is looked up.

## Summary

Fixes a pricing lookup bug for OpenRouter-style model IDs that include a vendor prefix and dotted version numbers (e.g., google/gemini-2.5-flash). getModelVariants() now also emits the vendor-stripped variant with dots converted to dashes (e.g., gemini-2-5-flash), allowing matches against existing pricing registry keys like openrouter::gemini-2-5-flash.

## Changes Made

- observability/mastra/src/metrics/pricing-registry.ts
  - getModelVariants(): when a model contains a slash, extract the substring after the slash (afterSlash) and add both afterSlash and afterSlash.replace(/\./g, '-') to the variants Set (in addition to existing variants). Date-stripped variants are still added via stripDateSuffix and the Set deduplicates collisions.

- .changeset/openrouter-dotted-model-pricing-fix.md
  - Patch release note describing the fix for OpenRouter dotted-version model pricing lookups.

## Impact

- estimateCosts will now match pricing rows such as openrouter::gemini-2-5-flash for routes like openrouter/google/gemini-2.5-flash, returning a non-null estimatedCost.
- Studio cost panels will correctly display costs for affected OpenRouter models (Google Gemini, Llama 3.1 variants, etc.).
- Existing pricing registry tests should remain green; Set-based deduplication prevents duplicate lookups and the change is conservative.

[Review Change Stack](https://app.coderabbit.ai/change-stack/mastra-ai/mastra/pull/17308?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


abhiaiyer91:
Closing this in favor of splitting it up. This PR is scoped/titled as the OpenRouter dotted-version pricing fix (#17189), but it actually bundles 5 unrelated changes closing #17189, #17265, #16444, #17443, and #17012 — including a `minor` public-API addition (`validators` on `registerApiRoute` + new `@hono/standard-validator` dep) that warrants its own review.

Each fix is fine on its own, but they should ship as separate PRs so they can be reviewed and released independently. I'll re-open the pricing fix as a standalone PR (ideally with a regression test covering the dotted `gemini-2.5-*` → `gemini-2-5-*` case, which currently has no coverage).

PR #17314, opened 2026-05-30
fix(core): fix TS2589 depth error, storage debug spam, and MCP const-field validation

### LINKED ISSUE #16444: Issue with MCP Tools with “cost”-constrained fields

### LINKED ISSUE #17047: TS2589: Type instantiation excessively deep when using MastraMessagePart with @hono/zod-openapi (regression since 1.24.0)
### Description

Since `@mastra/core@1.24.0`, using `convertMessages(messages).to('AIV5.UI')` inside a `@hono/zod-openapi` route handler causes TypeScript error **TS2589: Type instantiation is excessively deep and possibly infinite** on the entire handler (lines spanning the `openapi()` call).

This worked fine in `@mastra/core@1.12.0`.

### Root Cause

The regression was introduced by #14592, which expanded `MastraMessagePart` to support the AI SDK v6 approval flow. The new definition uses inline complex types:

```typescript
export type MastraMessagePart =
  | PartWithProviderMetadata<
      Exclude<UIMessageV4['parts'][number], { type: 'tool-invocation' | 'source' | 'step-start' }>
    >
  // ...
  | PartWithProviderMetadata<AIV5Type.DataUIPart<AIV5.UIDataTypes>>;
```

When TypeScript resolves `MastraMessagePart` in the context of `@hono/zod-openapi`'s deep response-type validation, the inline `PartWithProviderMetadata<Exclude<...>>` and `PartWithProviderMetadata<DataUIPart<...>>` chains push TypeScript past its type instantiation depth limit.

### Minimal Reproduction

```typescript
import { OpenAPIHono } from '@hono/zod-openapi';
import { convertMessages } from '@mastra/core/agent';
import type { HonoBindings, HonoVariables } from '@mastra/hono';

const router = new OpenAPIHono<{ Bindings: HonoBindings; Variables: HonoVariables }>();

router.openapi(someRoute, async (c) => {
  const result = await memory.recall({ threadId: 'id', perPage: 50 });
  const uiMessages = convertMessages(result.messages).to('AIV5.UI');
  //                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  // TS2589: Type instantiation is excessively deep and possibly infinite.
  return c.json({ ...result, uiMessages }, 200);
});
```

### Fix

Extract the inline complex types into named intermediate type aliases. TypeScript caches named types, breaking the deep instantiation chain:

```typescript
// Before
export type MastraMessagePart =
  | PartWithProviderMetadata<
      Exclude<UIMessageV4['parts'][number], { type: 'tool-invocation' | 'source' | 'step-start' }>
    >
  | ...
  | PartWithProviderMetadata<AIV5Type.DataUIPart<AIV5.UIDataTypes>>;

// After
type MastraStandardUIV4Part = Exclude<UIMessageV4['parts'][number], { type: 'tool-invocation' | 'source' | 'step-start' }>;
type MastraStandardUIV4PartWithMetadata = PartWithProviderMetadata<MastraStandardUIV4Part>;
type MastraDataUIPart = PartWithProviderMetadata<AIV5Type.DataUIPart<AIV5.UIDataTypes>>;

export type MastraMessagePart =
  | MastraStandardUIV4PartWithMetadata
  | ...
  | MastraDataUIPart;
```

A PR with this fix is attached.

### Versions

- `@mastra/core`: 1.24.0 (broken), 1.12.0 (working)
- TypeScript: 5.x
- `@hono/zod-openapi`: latest

### LINKED ISSUE #17137: [BUG] "Cannot get workflow run. Mastra storage is not initialized" debug log spam on every agent.generate() despite storage configured

SaarthurR:
## Summary
- Fix TS2589 "type instantiation is excessively deep" error when `MastraMessagePart` is used alongside deeply-generic libraries like `@hono/zod-openapi` — extract the `Exclude<UIMessageV4['parts'][number], ...>` union into a named type alias so TypeScript caches the expansion instead of re-evaluating it per use site. (#17047)
- Fix `"Mastra storage is not initialized"` debug spam appearing on every `agent.generate()` call when no storage backend is configured — guard `shouldPersistSnapshot` in `createAgenticLoopWorkflow` with a storage availability check before attempting persistence. (#17137)
- Fix MCP tools with `const`-constrained or single-value `enum` discriminator fields always failing validation — auto-inject these predetermined values into tool input before the validation pipeline runs so the LLM does not need to supply them. (#16444)

## Test plan
```
pnpm --filter @mastra/core exec vitest run src/tools/validation.test.ts --bail 1 --reporter=dot
pnpm --filter @mastra/core typecheck
```

Closes #17047
Closes #17137
Closes #16444

## ELI5

This PR fixes three separate bugs in the Mastra core library: (1) a TypeScript error that happens when Mastra is used with certain other libraries (fixed by reorganizing type definitions), (2) annoying repeated debug messages appearing when storage isn't set up (fixed by only checking for storage when needed), and (3) MCP tools failing validation when they have required fixed-value fields (fixed by automatically filling in those fixed values before validation runs).

### Type System Optimization (TS2589 Fix)
- **File:** `packages/core/src/agent/message-list/state/types.ts`
- Introduces a new internal type alias `UIV4NonMastraPart` to extract the complex `Exclude<UIMessageV4['parts'][number], ...>` expression
- Updates `MastraMessagePart` to use the precomputed alias instead of the inline complex expression
- Resolves TypeScript's "type instantiation is excessively deep" error (TS2589) when Mastra is used alongside deeply generic libraries like `@hono/zod-openapi`

### Storage Debug Spam Fix
- **File:** `packages/core/src/loop/workflows/agentic-loop/index.ts`
- Adds a storage-availability check before attempting to persist snapshots in `shouldPersistSnapshot` logic
- Prevents repeated "Mastra storage is not initialized" debug messages when no storage backend is configured
- Updated inline documentation to reflect the new persistence guard

### MCP Tool Const-Field Auto-Injection
- **File:** `packages/core/src/tools/validation.ts`
- Adds new `injectConstValues` helper function that auto-injects JSON-schema `const` values and single-value `enum` values into tool input before validation
- Runs as a normalization step (Step 2.5) in the validation pipeline
- Ensures discriminator fields with fixed values are automatically populated, allowing MCP tools with `const` or single-value enum fields to pass validation without requiring the LLM to provide the predetermined values

### Changelog Entries
- Three changeset files added to document these fixes for `@mastra/core` patch release

[Review Change Stack](https://app.coderabbit.ai/change-stack/mastra-ai/mastra/pull/17314?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/.changeset/mcp-const-field-auto-inject.md
+++ b/.changeset/mcp-const-field-auto-inject.md
@@ -1,0 +1,5 @@
+---
+'@mastra/core': patch
+---
+
+Fixed MCP tools with `const`-constrained or single-value `enum` discriminator fields (e.g. `"@type": { "const": "com.SomeSpec" }`) always failing validation. Mastra now automatically injects these predetermined values before validation runs.
--- a/.changeset/storage-not-initialized-debug-spam.md
+++ b/.changeset/storage-not-initialized-debug-spam.md
@@ -1,0 +1,5 @@
+---
+'@mastra/core': patch
+---
+
+Fixed "Mastra storage is not initialized" debug messages appearing on every `agent.generate()` call when no storage backend is configured.
--- a/.changeset/ts2589-mastra-message-part-type.md
+++ b/.changeset/ts2589-mastra-message-part-type.md
@@ -1,0 +1,5 @@
+---
+'@mastra/core': patch
+---
+
+Fixed a TypeScript TS2589 "type instantiation is excessively deep" error when using Mastra alongside deeply-generic libraries such as @hono/zod-openapi.
--- a/packages/core/src/agent/message-list/state/types.ts
+++ b/packages/core/src/agent/message-list/state/types.ts
@@ -73,13 +73,18 @@ export type MastraSourceUrlPart = Omit<LegacySourcePart, 'providerMetadata'> & {
   createdAt?: number;
 };
 
+// Named alias so TypeScript caches the Exclude expansion and avoids TS2589 when
+// MastraMessagePart is used alongside deeply-generic libraries like @hono/zod-openapi.
+type UIV4NonMastraPart = Exclude<
+  UIMessageV4['parts'][number],
+  { type: 'tool-invocation' | 'source' | 'step-start' }
+>;
+
 // Canonical stored part type. It starts from the v4 UI part model and extends
 // it with provider metadata, AI SDK v5 data parts, and v6-only persisted parts
 // such as approval-aware tool invocations and source documents.
 export type MastraMessagePart =
-  | PartWithProviderMetadata<
-      Exclude<UIMessageV4['parts'][number], { type: 'tool-invocation' | 'source' | 'step-start' }>
-    >
+  | PartWithProviderMetadata<UIV4NonMastraPart>
   | MastraStepStartPart
   | MastraToolInvocationPart
   | MastraSourceUrlPart
--- a/packages/core/src/loop/workflows/agentic-loop/index.ts
+++ b/packages/core/src/loop/workflows/agentic-loop/index.ts
@@ -64,10 +64,10 @@ export function createAgenticLoopWorkflow<Tools extends ToolSet = ToolSet, OUTPU
         internal: InternalSpans.WORKFLOW,
       },
       shouldPersistSnapshot: params => {
-        // We need a persisted snapshot record to support `resumeStream()`.
-        // - Create the initial record early ("pending")
-        // - Update it when execution is suspended ("paused"/"suspended")
-        // Avoid persisting "running" snapshots so we don't overwrite an existing suspended snapshot.
+        // Skip when no storage is configured to avoid debug spam on every generate() call.
+        if (!rest.mastra?.getStorage()) return false;
+        // Persist pending/suspended snapshots to support resumeStream(); skip "running"
+        // to avoid overwriting an existing suspended snapshot.
         return (
           params.workflowStatus === 'pending' ||
           params.workflowStatus === 'paused' ||
--- a/packages/core/src/tools/validation.ts
+++ b/packages/core/src/tools/validation.ts
@@ -201,6 +201,34 @@ function normalizeNullishInput(schema: StandardSchemaWithJSON<any>, input: unkno
 }
 
 /**
+ * Auto-injects const-constrained field values into input before validation (GitHub #16444).
+ * JSON Schema `const` and single-value `enum` properties have a fixed predetermined value
+ * that LLMs cannot know to supply (e.g. discriminator fields like `"@type": {"const": "..."}`).
+ * Injecting them here lets the tool schema validate successfully without requiring the LLM
+ * to guess opaque constant values.
+ */
+function injectConstValues(schema: StandardSchemaWithJSON<unknown>, input: unknown): unknown {
+  if (!isPlainObject(input)) return input;
+  const jsonSchema = standardSchemaToJSONSchema(schema, { io: 'input' });
+  if (jsonSchema.type !== 'object' || !jsonSchema.properties) return input;
+
+  let injected = false;
+  const result: Record<string, unknown> = { ...(input as Record<string, unknown>) };
+  for (const [key, rawProp] of Object.entries(jsonSchema.properties)) {
+    if (result[key] !== undefined) continue;
+    const prop = rawProp as Record<string, unknown>;
+    if ('const' in prop) {
+      result[key] = prop['const'];
+      injected = true;
+    } else if (Array.isArray(prop['enum']) && prop['enum'].length === 1) {
+      result[key] = prop['enum'][0];
+      injected = true;
+    }
+  }
+  return injected ? result : input;
+}
+
+/**
  * Checks if a value is a plain object (created by {} or new Object()).
  * This excludes class instances, built-in objects like Date/Map/URL, etc.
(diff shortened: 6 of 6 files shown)


abhiaiyer91:
Closing, we will accept this if reopened 1 PR per issue

PR #17320, opened 2026-05-30
fix(inngest): prevent duplicate step IDs in persistStepUpdate

### LINKED ISSUE #17319: [BUG] @mastra/inngest: persistStepUpdate generates duplicate Inngest step IDs, triggering AUTOMATIC_PARALLEL_INDEXING warning
## Description

When running a simple linear workflow via `@mastra/inngest`, the server logs consistently show:

```
Duplicate step ID "workflow.<id>.run.<uuid>.path.[0].stepUpdate" detected across parallel chains
{
  code: "AUTOMATIC_PARALLEL_INDEXING",
  explanation: "Using the same ID for steps in different parallel chains can cause unexpected behaviour. Your function is still running.",
  action: "Use a unique ID for each step, especially those in parallel.",
}
```

## Root Cause

`persistStepUpdate` in `@mastra/core` generates its Inngest step ID using only the `executionPath`:

```ts
// packages/core/src/... (chunk-2365SJTK.js:12107 in the compiled output)
const operationId = `workflow.${workflowId}.run.${runId}.path.${JSON.stringify(executionContext.executionPath)}.stepUpdate`;
await engine.wrapDurableOperation(operationId, async () => { ... });
```

`wrapDurableOperation` calls `inngestStep.run(operationId, fn)` internally. The problem is that `persistStepUpdate` is called **twice** with the **same `operationId`** within a single Inngest function execution for the same step:

1. **Inside `executeStep`** — before the step runs, to record `status: "running"`
2. **Inside the `execute` main loop** — after `executeEntry` returns, to record the final `status: "success"` / `"failed"` / etc.

Both calls share the same `executionContext.executionPath` (e.g. `[0]` for the first step), so they produce an identical `operationId`. Inngest's `resolveStepIdCollision` detects the collision and fires the `AUTOMATIC_PARALLEL_INDEXING` warning.

The fix would be to differentiate the two calls, for example by appending a phase suffix:
- `...path.[0].stepUpdate.start` (before execution)
- `...path.[0].stepUpdate.end` (after execution)

## Reproduction

Any workflow with at least one step via `@mastra/inngest` reproduces this. Minimal example:

```ts
import { init } from "@mastra/inngest";
import { z } from "zod";
import { inngest } from "./inngest-client";

const { createWorkflow, createStep } = init(inngest);

const myStep = createStep({
  id: "my-step",
  inputSchema: z.object({ value: z.string() }),
  execute: async ({ inputData }) => ({ ok: true }),
});

export const myWorkflow = createWorkflow({
  id: "my-workflow",
  inputSchema: z.object({ value: z.string() }),
})
  .then(myStep)
  .commit();
```

Trigger the workflow once and observe the `AUTOMATIC_PARALLEL_INDEXING` warning in logs.

## Impact

- Inngest resolves the collision automatically (appends `:1` to the second call's ID), so the workflow **continues to run correctly**.
- However, the second `persistStepUpdate` call (final status) now runs under a different memoized step key than intended. Depending on whether Mastra storage is configured, the final workflow snapshot may or may not be persisted correctly.
- The warning is noisy and could mask real parallelism bugs.

## Environment

- `@mastra/core`: 1.37.1
- `@mastra/inngest`: 1.4.1
- `inngest`: 4.5.0
- Runtime: Bun 1.3.14

Aayush-engineer:
## Description

`persistStepUpdate` was called twice per step with the same `operationId` — once in `step.ts` before execution (`status: running`) and once in `entry.ts` after execution (final status). Both calls shared the same `executionContext.executionPath`, producing identical Inngest step IDs and triggering `AUTOMATIC_PARALLEL_INDEXING` warnings.

Added an optional `phase` field to `PersistStepUpdateParams`. When provided, it is appended to the `operationId` as a suffix. The pre-execution call in `step.ts` now passes `phase: 'start'`. All other call sites omit `phase`, keeping full backward compatibility with already-running workflows.

## Related issue(s)

Fixes #17319

## Type of change

## Checklist

## ELI5

A workflow step was being saved twice with the exact same ID (once before running, once after), which caused confusing warnings. The pre-run save now gets a "start" label so each save has a unique ID and the warnings stop.

### Changeset Documentation
- Added `.changeset/fix-inngest-duplicate-step-ids.md` documenting a patch bump for `@mastra/core` and the fix for duplicate Inngest step IDs.

### Type Definitions
- **packages/core/src/workflows/handlers/entry.ts**: Added optional `phase?: string` to `PersistStepUpdateParams`. `persistStepUpdate` appends the phase as a suffix to the durable operation ID when provided (e.g., `.start`), preserving the original operation ID when `phase` is omitted.

### Step Execution
- **packages/core/src/workflows/handlers/step.ts**: `executeStep` now passes `phase: 'start'` to `engine.persistStepUpdate()` for the pre-execution (status: running) persistence call. All post-execution call sites continue to omit `phase` for backward compatibility with in-flight workflows.

## Motivation

Fixes duplicate Inngest step IDs and the resulting AUTOMATIC_PARALLEL_INDEXING warnings by ensuring the two persistStepUpdate invocations for the same step (pre-execution and post-execution) generate distinct operation IDs via an optional phase suffix, while maintaining compatibility with existing executions.


Aayush-engineer:
@graysonhicks ready for review when you get a chance!

TheIsrael1 (approved):

PR #17328, opened 2026-05-30
fix(memory): include role, content, and created_at in vector metadata

### LINKED ISSUE #16380: Memory.saveMessages writes incomplete vector metadata (drops role/content/created_at)
## Summary

`Memory.saveMessages` (and two adjacent inline-embedding paths in the same file) upsert vector metadata that omits `role`, `content`, and `created_at`. Only `message_id`, `thread_id`, `resource_id` are written. This breaks any downstream consumer that builds search results from vector metadata alone — they receive matches with empty content fields and cannot tell who said what.

The agent path (`@mastra/core` `SemanticRecall.processOutputResult`) writes the full metadata, so the inconsistency surfaces only when messages are persisted via `Memory.saveMessages` directly — most commonly through the built-in `POST /api/memory/save-messages` HTTP route, which is the path third-party agents (e.g. LiveKit voice) use to write messages without going through `agent.generate/stream`.

## Versions

- `@mastra/memory@1.17.5` (latest stable)
- `@mastra/memory@1.18.0-alpha.3` (latest alpha) — same bug
- Confirmed on `main` at `packages/memory/src/index.ts` lines ~1138-1142

## Where

`packages/memory/src/index.ts`, three locations:

1. `saveMessages` — main inline-embedding loop:

```ts
embeddingData.push({
  embeddings: result.embeddings,
  metadata: result.chunks.map(() => ({
    ...threadMetadata,
    message_id: message.id,
    thread_id: message.threadId,
    resource_id: message.resourceId,
  })),
});
```

2. The update path that re-embeds modified messages (around the `existingMessage` branch).
3. `rememberMessages` / re-embedding helper near the end of the file.

All three need the same fields added.

## Compare to the agent path

`@mastra/core` `SemanticRecall.processOutputResult` writes:

```ts
metadataList.push({
  message_id: message.id,
  thread_id: threadId,
  resource_id: resourceId || '',
  role: message.role,
  content: textContent,
  created_at: message.createdAt.toISOString(),
});
```

## Effect we observed

Production data: 19% of one user's vector index entries (50/262) had `role`/`content`/`created_at` missing from `pgvector` metadata — all 50 came from threads written via the LiveKit voice agent, which posts to `/api/memory/save-messages`. Older threads written via the agent's generate/stream pipeline have full metadata. The empty-metadata vectors still match queries, eat top-K slots, and produce empty-content rows in our search-memory tool's output.

Cosine similarity between query embeddings and these empty-metadata vectors is identical to what it would be for the corresponding properly-saved vector — the embedding itself is correct, only the metadata is incomplete.

## Repro

1. Configure `Memory` with `semanticRecall: { topK: 3, scope: 'resource' }` and a `PgVector`.
2. Call `memory.saveMessages({ messages: [{ id, threadId, resourceId, role: 'user', content: { format: 2, parts: [{ type: 'text', text: 'hello' }] }, createdAt: new Date() }] })` directly.
3. Inspect the vector store. The new row's metadata is `{ message_id, thread_id, resource_id }` only.
4. Same row written via `agent.stream()` writes metadata with `role`, `content`, `created_at` populated.

## Suggested fix

In all three locations in `packages/memory/src/index.ts`, extend the metadata object with the missing fields:

```ts
metadata: result.chunks.map(() => ({
  ...threadMetadata,
  message_id: message.id,
  thread_id: message.threadId,
  resource_id: message.resourceId,
  role: message.role,
  content: textForEmbedding, // already computed in scope
  created_at: message.createdAt instanceof Date ? message.createdAt.toISOString() : String(message.createdAt),
}))
```

`textForEmbedding` is the same string that was just embedded, so no extra work. The schema in `pgvector`'s metadata is `jsonb`, so adding fields is non-breaking for existing rows; consumers should already tolerate missing keys.

## Workaround

Patching `@mastra/memory` locally via `pnpm patch` to add the three fields at all three sites.

Happy to send a PR if a maintainer confirms the proposed shape is correct (matching `SemanticRecall.processOutputResult` exactly).

SaarthurR:
## Summary

- Write `role`, `content`, and `created_at` (alongside the existing `message_id`/`thread_id`/`resource_id`) into the metadata stored with each message's vector embedding, so consumers reading directly from vector search results get complete message data instead of only the lookup ids.
- Applied consistently across all four embedding paths (`saveMessages`, `updateMessages`, `indexMessagesList`, `embedClonedMessages`) via a shared `buildVectorMessageMetadata` helper that also normalizes `created_at` to an ISO string. Matches the metadata already written by the semantic-recall processor in `@mastra/core`.

## Test plan

- `pnpm --filter ./packages/memory test` (see new `vector metadata completeness` tests in `src/index.test.ts`)

Closes #16380

## ELI5

When messages are saved with vector embeddings (a way to make text searchable), the system used to only store the message's ID number in the metadata. Now it also stores the actual message content, who wrote it, and when it was created—so someone searching through vectors can get the complete message information without having to look it up separately.

## Overview

This PR enriches vector embedding metadata for messages in the `@mastra/memory` package by including role, content, and created_at fields alongside existing identifiers. Previously, vector metadata contained only message_id, thread_id, and resource_id, leaving consumers unable to reconstruct complete message data from vector metadata alone. The fix aligns the package with what the semantic-recall processor in `@mastra/core` expects.

### Core Implementation (packages/memory/src/index.ts)

- Introduced a new `buildVectorMessageMetadata` helper function that standardizes the construction of vector embedding metadata
- The helper enriches metadata with:
  - `message_id`, `thread_id`, `resource_id` (existing identifiers)
  - `role` (message sender role)
  - `content` (the text used for embedding)
  - `created_at` (normalized to ISO string format)

- Applied this metadata generation consistently across all embedding paths:
  - `saveMessages`: When saving messages for semantic recall
  - `updateMessages`: When updating messages and syncing semantic-recall vectors
  - `indexMessagesList`: When indexing a list of messages directly
  - `embedClonedMessages`: When embedding cloned thread messages

### Tests (packages/memory/src/index.test.ts)

- Added a new test suite "vector metadata completeness (issue `#16380`)" with 88 new lines
- Tests validate that `Memory.saveMessages` and `Memory.updateMessages` correctly include all required metadata fields
- `updateMessages` test verifies that metadata fields are sourced from the stored message (not the partial update input)
- Tests use mocked vector client and embedder (384-dim)

### Changeset

- Updated `@mastra/memory` with a patch version bump documenting the functional change

## Technical Details

- Metadata is stored as JSONB in pgvector, making this change backward-compatible
- Existing consumers that only use identifiers will continue to work
- The metadata shape now matches what `@mastra/core`'s SemanticRecall processor already expects
- `created_at` is normalized to ISO string format for consistency

## Related Issue

Closes issue `#16380`

[Review Change Stack](https://app.coderabbit.ai/change-stack/mastra-ai/mastra/pull/17328?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/.changeset/blue-crabs-repeat.md
+++ b/.changeset/blue-crabs-repeat.md
@@ -1,0 +1,5 @@
+---
+'@mastra/memory': patch
+---
+
+Include role, content, and created_at in vector embedding metadata when saving and updating messages, so consumers that build search results directly from vector metadata get complete message data instead of only the lookup ids.
--- a/packages/memory/src/index.test.ts
+++ b/packages/memory/src/index.test.ts
@@ -1766,6 +1766,94 @@ describe('Memory', () => {
     });
   });
 
+  describe('vector metadata completeness (issue #16380)', () => {
+    const embeddingDim = 384;
+    const fakeEmbedding = new Array(embeddingDim).fill(0.1);
+
+    function setup() {
+      const mockVector: MastraVector = {
+        createIndex: vi.fn().mockResolvedValue(undefined),
+        upsert: vi.fn().mockResolvedValue(undefined),
+        query: vi.fn().mockResolvedValue([]),
+        listIndexes: vi.fn().mockResolvedValue([]),
+        deleteVectors: vi.fn().mockResolvedValue(undefined),
+        describeIndex: vi.fn().mockResolvedValue({ dimension: embeddingDim }),
+        id: 'mock-vector',
+      } as any;
+      const mockEmbedder = {
+        doEmbed: vi.fn().mockResolvedValue({ embeddings: [fakeEmbedding] }),
+        modelId: 'mock-384-embedder',
+        specificationVersion: 'v1',
+        provider: 'mock',
+      } as any;
+      const memory = new Memory({
+        storage: new InMemoryStore(),
+        vector: mockVector,
+        embedder: mockEmbedder,
+        options: { semanticRecall: { scope: 'thread' }, lastMessages: 10, generateTitle: false },
+      });
+      return { memory, mockVector };
+    }
+
+    it('saveMessages writes role, content, and created_at to vector metadata', async () => {
+      const { memory, mockVector } = setup();
+      await memory.createThread({ threadId: 'meta-thread', resourceId: 'meta-resource' });
+
+      await memory.saveMessages({
+        messages: [
+          {
+            id: 'meta-msg-1',
+            role: 'user',
+            content: { format: 2, parts: [{ type: 'text', text: 'hello world' }], content: 'hello world' },
+            createdAt: new Date('2024-01-01T00:00:00.000Z'),
+            threadId: 'meta-thread',
+            resourceId: 'meta-resource',
+          } as MastraDBMessage,
+        ],
+      });
+
+      expect(mockVector.upsert).toHaveBeenCalled();
+      const metadata = vi.mocked(mockVector.upsert).mock.calls[0]![0].metadata[0] as Record<string, unknown>;
+      expect(metadata.message_id).toBe('meta-msg-1');
+      expect(metadata.thread_id).toBe('meta-thread');
+      expect(metadata.resource_id).toBe('meta-resource');
+      expect(metadata.role).toBe('user');
+      expect(metadata.content).toBe('hello world');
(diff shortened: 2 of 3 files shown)


abhiaiyer91:
There are already PRs opened for this. I see you are opening PRs while I thank you for trying to contribute opening PRs for issues with PRs already isn't helping!

PR #17351, opened 2026-05-31
fix(pg): batch memory saveMessages inserts

### LINKED ISSUE #17350: [BUG] Postgres saveMessages inserts each message in a separate round trip
### Describe the Bug

### Steps To Reproduce

2. Create a thread.

### Expected Behavior

### Environment Information

### Additional context

mbenhamd:
## Description

Batches `MemoryPG.saveMessages` inserts into multi-row Postgres upserts instead of awaiting one insert per message. The insert path now writes one statement per bind-parameter-safe chunk, while preserving the existing transaction, conflict update columns, and returned message behavior.

The implementation also de-dupes duplicate message ids before each batch so Postgres does not reject a multi-row `ON CONFLICT` statement that would affect the same row twice. This keeps the previous sequential behavior: later duplicate messages update mutable fields, while the original `createdAt` remains preserved.

Local Docker Postgres benchmarking showed the end-to-end `saveMessages` median improving from 24.32ms to 4.66ms for 100 messages and from 112.68ms to 15.24ms for 500 messages. The stable behavior change is reducing insert statements from `N` per-message upserts to `ceil(N / 8191)` chunked upserts.

## Related issue(s)

Fixes #17350

## Type of change

- [ ] Documentation update
- [ ] Code refactoring
- [x] Performance improvement
- [x] Test update

## Checklist

## ELI5

Instead of saving each message with a separate database trip, this PR groups many messages into batched multi-row upserts so the database is contacted far fewer times — making saves several times faster while preserving behavior.

### Performance Optimization
- Batch Postgres message inserts into multi-row INSERT ... VALUES (...) ON CONFLICT upserts, chunked to respect Postgres bind-parameter limits (POSTGRES_MAX_BIND_PARAMETERS = 65535 → MESSAGE_INSERT_BIND_PARAMETERS = 8 → MAX_MESSAGES_PER_INSERT = floor(65535 / 8)).
- Reduces DB round trips from N single-row upserts to ceil(N / MAX_MESSAGES_PER_INSERT) statements.

### Key Implementation Details (stores/pg/src/storage/domains/memory/index.ts)
- Added constants for bind-parameter limits and computed MAX_MESSAGES_PER_INSERT.
- Implemented dedupeMessagesForSave to collapse duplicate message ids within the incoming batch, preserving the first-seen message’s createdAt for that id while allowing later duplicates to overwrite mutable fields.
- Validates messages before starting a transaction (ensures required threadId and resourceId are present and verifies each referenced thread exists); mixed-thread batches are supported.
- Generates dynamic multi-row VALUES placeholders and emits one upsert per bind-safe chunk while preserving transaction atomicity.
- De-duplicates before each chunk to avoid Postgres rejecting multi-row ON CONFLICT statements that would affect the same row twice.
- Error handling adjusted: re-throws MastraError unchanged; wraps other errors in a MastraError.

### Preserved Behaviors
- ON CONFLICT (id) updates for thread_id, content, role, type, and resourceId.
- Preservation of createdAt/createdAtZ on conflict (first-seen createdAt kept when duplicate ids occur within the save).
- Transaction atomicity (including thread updatedAt/updatedAtZ updates).
- Duplicate-id semantics match prior behavior: createdAt preserved from the first-seen message for an id; later messages in the batch update mutable fields.

### Tests
- Unit tests (stores/pg/src/storage/domains/memory/save-messages.test.ts) verify:
  - Multi-row upsert usage and per-thread updatedAt updates.
  - Duplicate-id handling (createdAt preservation and last-write-wins for mutable fields).
  - Proper chunking across bind-parameter limits with one upsert per chunk.
  - JSON content round-trip behavior.
  - Validation errors for missing threadId/resourceId or missing thread references occur before transaction queries.
  - Mixed-thread batches update each touched thread (allowed).
- Integration test addition (stores/pg/src/storage/test-utils.ts) ensures saving 8192 messages persists all rows and exercises chunking logic.

### Performance Impact
- Local Docker Postgres benchmarks (median end-to-end saveMessages):
  - 100 messages: 24.32ms → 4.66ms
  - 500 messages: 112.68ms → 15.24ms

### Other
- Changeset (.changeset/twelve-frogs-kneel.md) bumps `@mastra/pg` patch and documents the performance improvement.
- Categorized as bug fix / performance improvement.
- Fixes issue `#17350`.


danielnc (commented):
Reviewed this against latest `main` and ran the new `save-messages.test.ts` locally (9/9 green). The approach is sound and the coverage is strong — multi-row upsert, bind-param chunking at 8191 rows, mixed-thread, and the reject-before-tx validation paths are all exercised. A few observations, none blocking:

**Correctness (✅ confirmed):** Deduping by `id` *before* chunking is the load-bearing detail — without it a multi-row `ON CONFLICT (id) DO UPDATE` containing the same id twice raises Postgres `cardinality_violation` ("ON CONFLICT DO UPDATE command cannot affect row a second time"). `dedupeMessagesForSave` correctly collapses duplicates across the whole batch before slicing, so the same id can never land in two chunks. The last-write-wins-content / first-write-wins-`createdAt` semantics are reasonable and tested. Computing `createdAt` once and reusing it for both `createdAt` and `createdAtZ` also quietly fixes a latent micro-skew in the old code (which called `new Date()` twice).

**Minor — redundant first-message guard:** the early `const threadId = messages[0]?.threadId; if (!threadId) throw 'Thread ID is required'` now overlaps the per-message loop check that throws `Expected to find a threadId for message…`. Two different messages for the same condition on `messages[0]`, and `threadId` is otherwise unused now that the tx keys off the `threadIds` set. Could drop the early guard (or unify the message) for clarity.

**Scope note for a maintainer:** this generalizes `saveMessages` to mixed-thread batches (validates every thread, updates each touched thread). That's beyond the issue's single-thread N+1 ask — it's a nice superset and is tested, just flagging so it's a deliberate API decision rather than incidental.

**Optional optimization (defer-able):** thread existence is validated with N sequential `getThreadById` calls (one per unique thread). The common single-thread path is unchanged (1 query, no regression); for large mixed-thread batches it could collapse to a single `SELECT id … WHERE id = ANY($1)`. Not worth blocking on.

**Test-depth nit:** the new unit tests assert SQL *string shape* via a recording mock rather than executing real upsert semantics (e.g. that `ON CONFLICT` preserves `createdAt`). The combined-store (real-PG) suite covers the integration path and is green here, so coverage is adequate — just noting the new file is construction-level, not behavioral.

Nice perf win overall (the benchmark in the description matches what I'd expect). 👍

abhiaiyer91 (approved):

PR #17352, opened 2026-05-31
fix(pg): batch message inserts into multi-row INSERT statements in saveMessages

### LINKED ISSUE #17350: [BUG] Postgres saveMessages inserts each message in a separate round trip

xxiaoxiong:
## Problem

\`@mastra/pg\` persists \`memory.saveMessages\` by awaiting one INSERT ... ON CONFLICT per message inside a transaction. For a batch of N messages, this makes the insert path require N sequential database round trips plus the thread timestamp update.

## Root Cause

The \`saveMessages\` method in \`stores/pg/src/storage/domains/memory/index.ts\` uses a \`for\` loop with individual \`INSERT ... VALUES (, , ...) ON CONFLICT\` statements for each message, rather than batching multiple rows into a single INSERT statement.

## Changes

- Replaced the sequential row-by-row INSERT loop in \`saveMessages\` with batched multi-row INSERT statements
- Messages are chunked into batches of 8192 (well below Postgres' 65536 bind-parameter limit)
- Preserves all existing semantics:
  - \`ON CONFLICT (id) DO UPDATE SET\` for thread_id, content, role, type, and resourceId
  - \`createdAt\` / \`createdAtZ\` preservation on conflict (not in UPDATE SET)
  - Transaction atomicity with the thread timestamp update
  - Duplicate-id handling within a single \`saveMessages\` call

Performance improvement: reduces N sequential round trips to ceil(N / 8192) round trips.

Closes #17350

# ELI5

Instead of saving messages one-by-one (many slow database trips), this PR groups messages into large batches and inserts each batch with a single SQL statement — making saves much faster while keeping duplicate protection and creation timestamps intact.

---

### `stores/pg/src/storage/domains/memory/index.ts`

- Refactored `MemoryPG.saveMessages` to:
  - First validate all messages (threadId/resourceId) in a single pass.
  - Persist messages using batched multi-row `INSERT ... VALUES (...), (...)` statements with `ON CONFLICT (id) DO UPDATE` instead of per-message UPSERTs.
  - Chunk batches using `BATCH_SIZE = 8191` rows (reduced from 8192 in a commit) to keep total bind parameters below PostgreSQL's limit.
  - Build per-batch SQL placeholder tuples and a flattened values array, executing one UPSERT per batch.
- Preserves prior upsert semantics:
  - On conflict (id) updates: `thread_id`, `content`, `role`, `type`, and `resourceId`.
  - Deliberately omits `createdAt` / `createdAtZ` from the UPDATE set to preserve original creation timestamps.
- Maintains transaction atomicity: the batched inserts and the subsequent thread `updatedAt` / `updatedAtZ` update occur in the same transaction.
- Correctly handles duplicate ids that appear within a single `saveMessages` call.
- Fix: moved batching logic outside the validation loop to avoid repeated batching per message and fixed variable shadowing (inner `message` renamed to `msg`).

Performance impact: Reduces N sequential database round trips to ceil(N / 8191) round trips.

---

## Related Issue
Closes `#17350`

### DIFF
--- a/stores/pg/src/storage/domains/memory/index.ts
+++ b/stores/pg/src/storage/domains/memory/index.ts
@@ -1226,8 +1226,10 @@ export class MemoryPG extends MemoryStorage {
           }
         }
 
-        // Batch messages for multi-row INSERT (reduces round trips from N to ⌈N/8192⌉)
-        const BATCH_SIZE = 8192;
+        // Batch messages for multi-row INSERT (reduces round trips from N to ⌈N/8191⌉)
+        // Postgres supports at most 65535 bind parameters per query;
+        // each row uses 8 parameters, so max batch size is 65535/8 ≈ 8191.
+        const BATCH_SIZE = 8191;
         const batches: { placeholders: string[]; values: any[] }[] = [];
         let currentBatch: { placeholders: string[]; values: any[] } | null = null;
 
--- a/stores/pg/src/storage/domains/memory/index.ts
+++ b/stores/pg/src/storage/domains/memory/index.ts
@@ -1224,16 +1224,19 @@ export class MemoryPG extends MemoryStorage {
               `Expected to find a resourceId for message, but couldn't find one. An unexpected error has occurred.`,
             );
           }
-          await t.none(
-            `INSERT INTO ${tableName} (id, thread_id, content, "createdAt", "createdAtZ", role, type, "resourceId")
-             VALUES ($1, $2, $3, $4, $5, $6, $7, $8)
-             ON CONFLICT (id) DO UPDATE SET
-              thread_id = EXCLUDED.thread_id,
-              content = EXCLUDED.content,
-              role = EXCLUDED.role,
-              type = EXCLUDED.type,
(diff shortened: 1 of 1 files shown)


xxiaoxiong:
Fixed: moved the batching logic outside the message validation loop, so batching runs once instead of N times. Also fixed the variable shadowing (renamed inner  to ). The Vercel failures appear to be pre-existing authorization issues unrelated to this change.

xxiaoxiong:
Fixed: moved the batching logic outside the message validation loop, so batching runs once instead of N times. Also fixed the variable shadowing (renamed inner `message` to `msg`). The Vercel failures appear to be pre-existing authorization issues unrelated to this change.

xxiaoxiong:
Fixed: reduced BATCH_SIZE from 8192 to 8191 (Math.floor(65535 / 8)) to stay under Postgres 65535 bind parameter limit. Updated in commit 6f78b7a.

abhiaiyer91:
Closing already resolved

PR #17358, opened 2026-05-31
test(memory): guard assistant text persistence after tool call with observational memory

### LINKED ISSUE #15078: [BUG] Coordinator agent text response not persisted when observational memory is enabled
### Describe the Bug

When a coordinator agent (an agent with sub-agents via the agents property) has observationalMemory: true, the coordinator's text response is generated by the LLM but never persisted to the database. Only the tool invocations (sub-agent delegations) and data-om-status markers are saved.

### Steps To Reproduce

Create a coordinator agent with sub-agents and observationalMemory: true:

```
const coordinator = new Agent({
  id: 'coordinator',
  instructions: '...',
  model: 'google/gemini-3-flash-preview',
  agents: { subAgent },
  memory: new Memory({
    storage: pgStore,
    options: {
      observationalMemory: true,
      workingMemory: { enabled: true, schema: z.object({ ... }) },
    },
  }),
});
```

Send a message that causes the coordinator to delegate to a sub-agent and then respond with text.
Query the stored messages for the coordinator's thread:

```
SELECT id, role, content::text
FROM mastra.mastra_messages
WHERE thread_id = '<thread_id>' AND role = 'assistant';
```
Expected: The assistant message contains both tool-invocation parts AND a text part with the coordinator's response.

[mastra_messages_rows.json](https://github.com/user-attachments/files/26490078/mastra_messages_rows.json)

Actual: The assistant message contains only data-om-status, tool-invocation, and step-start parts. No text part is stored. The content top-level field is also missing.

**Root cause analysis:**

The issue is in ObservationalMemoryProcessor's interaction with ObservationTurn during multi-step LLM responses.

When the coordinator's LLM response spans multiple steps (step 0: tool call → step 1: text after tool result), the following occurs:

Step 0: LLM generates a tool call. Assistant message M1 is created in the MessageList with [tool-invocation] parts, tracked in the "response" source bucket.

Step 1 processInputStep: Because stepNumber > 0, the processor calls messageList.clear.response.db(), which:

Returns M1 from the "response" bucket
Persists M1 to the database (with only tool invocations, no text yet)
Moves M1 to the "memory" source bucket via messageList.add(msg, "memory")
Step 1 execution: The LLM generates the text response. The text parts are appended to M1 in-memory, but M1 is now in the "memory" bucket, not "response".

turn.end(): Checks only messageList.get.response.db() and messageList.get.input.db() for unsaved messages. The "response" bucket is empty (M1 was moved to "memory" in step 2). Nothing is re-saved. The text generated in step 1 is lost.

The database retains the version of M1 from step 2 — containing tool invocations and OM status markers but no text.

Verification: Langfuse traces confirm the coordinator LLM does generate text (visible in the trace output and as chunk: 'text' spans under step 1). The text is produced but not persisted.

Why it works without observational memory: When observationalMemory is disabled (the default — new Memory() without the option), the ObservationalMemoryProcessor is not registered. Without it, there is no step-level message clearing, so all parts (tool invocations + text) remain in the "response" bucket and are saved together at the end.

Workaround: Set observationalMemory: false in the coordinator agent's memory config.

Suggested fix: ObservationTurn.end() should detect messages in the "memory" bucket that have been modified since their last persist (e.g., by comparing part counts or a dirty flag) and re-save them. Alternatively, the step transition in processInputStep could avoid clearing the response bucket until the full turn is complete.

https://github.com/00md13/mastra-memory-issue

### Expected Behavior

The assistant message contains both tool-invocation parts AND a text part with the coordinator's response.

### Environment Information

```block
@mastra/core: 1.21.0
@mastra/memory: 1.13.0
@mastra/pg: 1.8.5
@mastra/server: 1.21.0
```

### Verification

SaarthurR:
## Summary

Adds an end-to-end regression guard for #15078: with observational memory enabled, a multi-step turn (step 0 tool call → step 1 text) must persist the assistant text response, not just the tool invocation.

The reported data loss was filed against `@mastra/core` 1.21.0 and is **already resolved on `main`**: when step 1's text is merged into the assistant message, `MessageList` re-pushes it to its source bucket (`message-list.ts` `pushMessageToSource(existingMessage, messageSource)`), so the updated message lands back in the `response` bucket and `turn.end()` persists it with the text intact.

This PR adds a test that locks in that behavior so it can't silently regress. It covers the "quiet" path where no observation fires this turn (so there is no response-message-id rotation) and step 1's text appends to the assistant message that the step transition already moved into the `memory` bucket — the exact condition described in #15078.

## Test plan

- `pnpm vitest run src/processors/observational-memory/__tests__/om-error-and-persistence.test.ts` — 7 passed.
- The new test fails if the step-1 text is dropped from the persisted assistant message.

Closes #15078

### DIFF
--- a/packages/memory/src/processors/observational-memory/__tests__/om-error-and-persistence.test.ts
+++ b/packages/memory/src/processors/observational-memory/__tests__/om-error-and-persistence.test.ts
@@ -548,6 +548,67 @@ describe('OM Persistence', () => {
     expect(hasOmParts).toBe(true);
   });
 
+  it('should persist the assistant text response that follows a tool call (issue #15078)', async () => {
+    const threadId = 'test-coordinator-text-thread';
+    const resourceId = 'test-resource';
+
+    const quietStore = new InMemoryStore();
+    const quietMemory = new Memory({
+      storage: quietStore,
+      options: {
+        observationalMemory: {
+          enabled: true,
+          observation: {
+            model: createMockObserverModel() as any,
+            messageTokens: 1_000_000,
+            bufferTokens: false,
+          },
+          reflection: {
+            model: createMockReflectorModel() as any,
+            observationTokens: 50_000,
+          },
+        },
+      },
+    });
+    const quietAgent = new Agent({
+      id: 'test-quiet-coordinator',
+      name: 'Test Quiet Coordinator',
+      instructions: 'You are a helpful assistant. Always use the test tool first.',
+      model: createMockOmModel(longResponseText) as any,
+      tools: { test: omTriggerTool },
+      memory: quietMemory,
+    });
+
+    const response = await quietAgent.stream('Please use the test tool and then answer.', {
+      memory: {
+        thread: threadId,
+        resource: resourceId,
+      },
+    });
+
+    const reader = response.fullStream.getReader();
+    try {
+      while (true) {
+        const { done } = await reader.read();
+        if (done) break;
+      }
+    } finally {
+      reader.releaseLock();
+    }
+
+    const memoryStore = await quietStore.getStore('memory');
+    const result = await memoryStore!.listMessages({ threadId });
+    const assistantMessages = result.messages.filter((m: any) => m.role === 'assistant');
+
+    const persistedText = assistantMessages
+      .flatMap((m: any) => m.content?.parts ?? [])
+      .filter((p: any) => p.type === 'text')
+      .map((p: any) => p.text)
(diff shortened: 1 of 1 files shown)


SaarthurR:
Closing — this is already fixed on `main` (see my note on #15078), so a test-only regression guard isn't worth the added review/maintenance surface. The existing OM persistence tests already cover this path. Leaving #15078 for a maintainer to close.

PR #17362, opened 2026-05-31
fix(playground): keep reasoning text visible after a chat reload

damien-schneider:
## Summary

PR 1 of a small split of chat-rendering fixes for #17208 (MastraDBMessage unification).

Reasoning blocks went **blank after reloading a thread**. Persisted reasoning parts store an empty `reasoning` field with the text inside `details` (AIV5Adapter writes `reasoning: '', details: [...]`, and core's own reader falls back to `details`). The Studio chat converter read only `part.reasoning`, so the reasoning text disappeared on reload.

## Scope

- `to-assistant-ui-message.ts`: the `reasoning` branch falls back to the joined `details` text. No cast needed — the V4 reasoning part already types `details`.
- Tests: reload shape (empty `reasoning` + `details`) renders the text; live streaming (`reasoning` field) still renders.

## Verification

- `pnpm --filter @internal/playground typecheck` — clean
- Converter tests pass (9, incl. 2 new reasoning cases)

## Intentionally excluded (split follow-ups)

- File/image rendering, `dynamic-tool` status, and the Observational-Memory badge type-safety fix ship as separate small PRs.
- SDK-side fixes (network sub-agent approvals/suspension, sub-agent `childMessages`, reload approvals/`suppressFeedback`) will be raised as comments on #17208 since they live in `@mastra/react`.

PR #17364, opened 2026-05-31
refactor(playground): drop redundant `as ContentPart` casts (re-enable type-checking)

damien-schneider:
## Summary

Small type-safety cleanup of the Studio chat converter (`to-assistant-ui-message.ts`). **Stacked on #17362** (same file).

Every branch of `toContentPart` asserted its return `as ContentPart`. Empirically, those casts hid **nothing**: removing all five type-checks cleanly (the `source-document` branch already returned without one). They only **suppressed checking**.

Removing them means the converter's content-part returns are now checked by inference — a future field typo or shape mismatch (e.g. `image: ...` with the wrong value type) is caught at compile time instead of being silently cast away.

## Scope

- Drop the 5 redundant `as ContentPart` casts in `toContentPart`.

## Verification

- `pnpm --filter @internal/playground typecheck` — clean (proves the literals already satisfy `ContentPart`)
- Converter tests pass (9)
- No runtime change — `as` casts are compile-time only.

## Honest note

This does **not** fix a hidden bug — there wasn't one behind these casts. It re-enables type-checking on these returns. The deeper converter type smell (the V4-anchored `MastraMessagePart` vs the V5 runtime shape the accumulator emits) is upstream in `@mastra/core` and out of scope here.

### DIFF
--- a/.changeset/quiet-converter-casts.md
+++ b/.changeset/quiet-converter-casts.md
@@ -1,0 +1,4 @@
+---
+---
+
+Internal: remove redundant `as ContentPart` casts in the Studio chat converter so its content-part returns are type-checked by inference. No behavior change.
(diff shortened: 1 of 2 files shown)


mfrachet:
Thanks 🙏🏻

PR #17369, opened 2026-05-31
fix: replace Function() eval with CSP-safe runtime Zod converter

### LINKED ISSUE #16992: [BUG] Self-hosted Studio requires 'unsafe-eval' CSP due to Function() in resolveSerializedZodOutput
## Describe the Bug

Self-hosted Mastra Studio cannot run under a strict Content Security Policy (CSP) that does not allow `'unsafe-eval'`. As soon as a user opens any page that renders a schema-driven form (e.g. `/tools/:tool-name`), the page fails to load with:

```
EvalError: Evaluating a string as JavaScript violates the following Content Security Policy
directive because 'unsafe-eval' is not an allowed source of script:
script-src 'self' 'sha256-<hash-of-index.html-inline-script>'
```

This is a hard blocker for any deployment where the CSP is owned/enforced by a security team and `'unsafe-eval'` is not permitted. There is no workaround other than weakening the CSP.

## Root Cause

`packages/core/src/utils.ts` exposes `resolveSerializedZodOutput`, which materializes a Zod schema by **evaluating a string of JavaScript at runtime** via the `Function()` constructor:

```ts
// packages/core/src/utils.ts (~L325)
export function resolveSerializedZodOutput(schema: string): z.ZodType {
  // Creates and immediately executes a new function that takes 'z' as a parameter
  // The function body is a string that returns the serialized zod schema
  // When executed with the 'z' parameter, it reconstructs the zod schema in the current context
  return Function('z', `"use strict";return (${schema});`)(z);
}
```

The input `schema` is the **string output** of `jsonSchemaToZod(...)` from `packages/schema-compat/src/json-to-zod.ts`, which wraps `json-schema-to-zod`. That upstream library is a **code generator**: it emits JavaScript source code (e.g. `"z.object({ name: z.string() })"`) rather than a live `z.ZodType` instance. Studio therefore round-trips through `eval`-equivalent semantics to turn that string back into a runtime Zod schema.

Browsers gate `Function(string)` under the same CSP directive as `eval`, so any policy without `'unsafe-eval'` rejects this code path — and the calling React component throws synchronously, blanking the page.

The intermediate string form is fragile in other ways too. There is already a workaround in `packages/schema-compat/src/json-to-zod.ts` for TypeScript syntax leaking into the generated source:

```ts
// Fix: The upstream json-schema-to-zod generates TypeScript syntax `reduce<z.ZodError[]>`
// in parseOneOf which fails when evaluated at runtime with Function().
// See: https://github.com/mastra-ai/mastra/issues/11610
return result.replace(/\.reduce<[^>]+>/g, '.reduce');
```

That regex fix-up is a symptom of the same underlying design: the playground depends on stringified JS that must remain valid in a browser eval context.

## Impact / Affected Surfaces

`resolveSerializedZodOutput` (re-exported by the playground at `packages/playground/src/lib/form/utils.ts`) is consumed across most schema-driven UI in `packages/playground`:

- `packages/playground/src/domains/tools/components/ToolPanel.tsx` — tool input forms (`/tools/:tool-name`)
- `packages/playground/src/domains/agents/components/AgentToolPanel.tsx` — agent tool forms
- `packages/playground/src/domains/agents/components/agent-playground/agent-playground-request-context.tsx` — agent request-context variables
- `packages/playground/src/domains/mcps/components/MCPToolPanel.tsx` — MCP tool forms
- `packages/playground/src/domains/workflows/workflow/use-workflow-trigger.ts` — workflow trigger + state schemas
- `packages/playground/src/domains/workflows/workflow/workflow-suspended-steps.tsx` — resume schemas for suspended steps
- `packages/playground/src/domains/workflows/workflow/workflow-time-travel-form.tsx` — time-travel step/state forms
- `packages/playground/src/domains/datasets/utils/csv-validation.ts` — dataset CSV row validation
- `packages/playground/src/domains/request-context/components/request-context-schema-form.tsx` — request-context schema form
- `packages/core/src/utils.ts:466` — `isZodType(schema) ? schema : resolveSerializedZodOutput(jsonSchemaToZod(schema))`

Under a strict CSP, **all** of these pages either crash on render or silently fail to produce a usable form.

## Steps To Reproduce

1. Create a new Mastra project with at least one tool exposing a non-trivial input schema and build the app (`mastra build`) so the playground is served as part of the standalone server (the bundled Studio, not `mastra dev`).
2. Serve Studio behind a reverse proxy / hosting layer that injects a strict CSP header, for example:
   ```
   Content-Security-Policy:
     default-src 'self';
     script-src 'self' 'sha256-<hash-of-the-inline-script-in-index.html>';
     style-src 'self' 'unsafe-inline';
     connect-src 'self';
   ```
   Notably **without** `'unsafe-eval'` in `script-src`.
3. Open Studio in the browser, navigate to `/tools/<any-tool>` (or any of the affected pages listed above).
4. Observe the page fails to render. The browser devtools console shows an `EvalError` referencing the CSP `script-src` directive, originating from `Function()` inside `resolveSerializedZodOutput`.

Disabling the CSP, or adding `'unsafe-eval'` to `script-src`, makes the page render normally — confirming the cause.

## Expected Behavior

Mastra Studio should render schema-driven forms (tools, MCP tools, agent tools, workflow trigger/resume/state, request context, dataset validation) under a strict CSP that does **not** allow `'unsafe-eval'`. Studio should not require runtime `eval`/`Function(string)` to function.

A reasonable security baseline for self-hosted Studio is roughly:

```
script-src 'self';
connect-src 'self';
style-src 'self' 'unsafe-inline';
```

This should be achievable without policy relaxations specific to Studio.

## Suggested Fix

Replace the codegen-then-eval pipeline (`jsonSchemaToZod` → `string` → `Function()` → `z.ZodType`) with a **runtime converter** that builds a live `z.ZodType` directly from a JSON Schema object. Concretely:

1. Introduce a runtime `jsonSchemaToZodRuntime(schema: JsonSchema): z.ZodType` in `packages/schema-compat` that recursively constructs Zod types in-memory (no string emission). This can be implemented in-tree without a heavy new dependency, since the playground only exercises a bounded subset of JSON Schema (objects, arrays, primitives, enums, unions/`oneOf`/`anyOf`, `$ref` within the same doc, common formats).
2. Reimplement `resolveSerializedZodOutput` in `packages/core/src/utils.ts` so it no longer calls `Function()`:
   - Preferred: deprecate the function and update all call sites in `packages/playground` (listed above) plus `packages/core/src/utils.ts:466` to use the new runtime converter directly, passing the JSON Schema object instead of a stringified Zod expression.
   - Back-compat shim (if needed): keep `resolveSerializedZodOutput(schemaString)` exported but route it through a safe parser (e.g. a deterministic AST walk of the generated expression) rather than `Function()`. This is harder to keep correct than path (1) and is not recommended long term.
3. Remove the `.reduce<…>` regex workaround in `packages/schema-compat/src/json-to-zod.ts` once nothing depends on the generated string form.
4. Add a regression test that imports a representative tool input schema and asserts that no call path uses `Function`, `eval`, or `new Function` (e.g. by stubbing them in a `vitest` test and ensuring they are never invoked while rendering `ToolPanel`).

Once the eval path is gone, document that Studio is CSP-strict by default and add a Studio docs page describing the recommended CSP header.

There is no viable workaround that keeps the current implementation: `Function(string)` is gated by `'unsafe-eval'` in every modern browser. Whitelisting hashes does not help because `Function()` does not execute parsed inline script tags.

## Environment Information

Reproduces against the latest published `mastra` packages at the time of writing (Studio served by the built standalone server, not `mastra dev`).

- LLM Provider: N/A — the failure is in the Studio UI before any agent call is made.
- Browser: Chromium-based and Firefox both reproduce; the violated directive is enforced by the browser, not the server.

## Verification

sekerderya:
Fixes #16992

## Problem
`resolveSerializedZodOutput` was using `new Function()` to evaluate serialized Zod schemas at runtime. This is equivalent to `eval()` and is blocked by Content Security Policy (CSP) headers, causing Mastra Studio to fail in CSP-enforcing environments.

## Solution
Added `jsonSchemaToZodRuntime()` to `@mastra/schema-compat` that converts JSON Schema directly to a Zod type at runtime without any code generation or eval.

- `packages/schema-compat/src/json-to-zod.ts`: Added `jsonSchemaToZodRuntime` and `parseSchemaRuntime` functions
- `packages/core/src/utils.ts`: Updated `convertVercelToolParameters` to use `jsonSchemaToZodRuntime` directly
- `packages/playground/src/lib/form/utils.ts`: Updated `resolveSerializedZodOutput` to delegate to `jsonSchemaToZodRuntime`

## ELI5

This PR fixes a security issue where Mastra Studio was using dangerous code evaluation (Function/eval) that violates strict Content Security Policies. Instead of generating JavaScript code as strings and executing them to build schema validators, the code now builds them directly in memory without any string evaluation.

## Overview

Replaces unsafe `Function()`-based evaluation with a CSP-compliant runtime converter for materializing Zod schemas from JSON Schema objects. This eliminates Content Security Policy violations that were causing Mastra Studio to fail in restrictive CSP environments.

## Problem Addressed

The original implementation used `resolveSerializedZodOutput`, which constructed Zod schemas via `new Function('z', `"use strict";return (${schema});`)(z)` — equivalent to eval() and blocked by strict CSPs disallowing 'unsafe-eval'. This caused schema-driven forms to fail across multiple surfaces: tools, agent tool forms, MCP tools, workflows, dataset validation, and request-context forms.

### New Runtime Converter
**packages/schema-compat/src/json-to-zod.ts** (+127 lines)
- Added `jsonSchemaToZodRuntime(schema: JsonSchema): z.ZodType` — new public API for CSP-safe schema conversion
- Implemented `parseSchemaRuntime()` — recursive converter building Zod types directly from JSON Schema
- Handles:
  - Boolean schemas (`true` → `z.any()`, `false` → `z.never()`)
  - `$ref` resolution from local `#/$defs` / `#/definitions`
  - Combinators: `oneOf`/`anyOf` as `z.union()`, `allOf` as chained `.and()`
  - Keywords: `enum` → `z.enum()`, `const` → `z.literal()`
  - Primitive types: string, number, integer, boolean, null, array, object with constraints
  - Object properties with required/optional support and `additionalProperties` handling
  - Metadata: description via `.describe()`, defaults via `.default()`
- Existing `jsonSchemaToZod` string-generation API remains unchanged

### Updated Call Sites
**packages/core/src/utils.ts** (+2/-2)
- Updated `convertVercelToolParameters` to use `jsonSchemaToZodRuntime` directly instead of the serialized string reconstruction path
- Replaced import from `jsonSchemaToZod` with `jsonSchemaToZodRuntime`

**packages/playground/src/lib/form/utils.ts** (+2/-2)
- Updated `resolveSerializedZodOutput` to delegate to `jsonSchemaToZodRuntime`
- Removed unsafe `Function()` evaluation and direct `z` import

### Changeset
**.changeset/wild-cloths-dig.md** (+6/-0)
- Marked `@mastra/core` and `@mastra/schema-compat` as patch-level updates
- Documented fix: "Replace Function() eval with CSP-safe runtime Zod converter to fix CSP violations in Mastra Studio"

## Impact

- Eliminates CSP violations caused by runtime code evaluation
- Enables Mastra Studio to operate in strict CSP environments without requiring 'unsafe-eval' permission
- Maintains API compatibility while replacing unsafe implementation
- All schema-driven form surfaces now function safely under restrictive policies

### DIFF
--- a/.changeset/wild-cloths-dig.md
+++ b/.changeset/wild-cloths-dig.md
@@ -1,0 +1,6 @@
+---
+'@mastra/core': patch
+'@mastra/schema-compat': patch
+---
+
+Replace Function() eval with CSP-safe runtime Zod converter to fix CSP violations in Mastra Studio
--- a/packages/core/src/utils.ts
+++ b/packages/core/src/utils.ts
@@ -1,6 +1,6 @@
 import { createHash } from 'node:crypto';
 import type { CoreMessage } from '@internal/ai-sdk-v4';
-import { jsonSchemaToZod } from '@mastra/schema-compat/json-to-zod';
+import { jsonSchemaToZodRuntime } from '@mastra/schema-compat/json-to-zod';
 import { z } from 'zod/v4';
 import type { MastraPrimitives } from './action';
 import type { ToolsInput } from './agent';
@@ -465,7 +465,7 @@ function convertVercelToolParameters(tool: VercelTool): z.ZodType {
     schema = schema();
   }
 
-  return isZodType(schema) ? schema : resolveSerializedZodOutput(jsonSchemaToZod(schema));
+  return isZodType(schema) ? schema : jsonSchemaToZodRuntime(schema);
 }
 
 /**
--- a/packages/playground/src/lib/form/utils.ts
+++ b/packages/playground/src/lib/form/utils.ts
@@ -1,6 +1,6 @@
 import type { FieldConfig } from '@autoform/core';
 import { buildZodFieldConfig } from '@autoform/react';
-import { z } from 'zod';
+import { jsonSchemaToZodRuntime } from '@mastra/schema-compat/json-to-zod';
 import type { FieldTypes } from './auto-form';
 
 // @ts-expect-error - TODO
@@ -55,5 +55,5 @@ export function removeEmptyValues<T extends Record<string, any>>(values: T): Par
  * @returns resolved zod object
  */
 export function resolveSerializedZodOutput(obj: any) {
-  return Function('z', `"use strict";return (${obj});`)(z);
+  return jsonSchemaToZodRuntime(obj);
 }
--- a/packages/schema-compat/src/json-to-zod.ts
+++ b/packages/schema-compat/src/json-to-zod.ts
@@ -7,6 +7,7 @@ import jsonSchemaToZodOriginal, {
   parseOneOf,
   parseSchema,
 } from 'json-schema-to-zod';
+import { z } from 'zod';
 
 function parseObject(objectSchema: JsonSchemaObject & { type: 'object' }, refs: Refs): string {
   let properties: string | undefined = undefined;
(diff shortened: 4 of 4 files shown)


sekerderya:
Thanks for the review — I’m going to close this for now as I don’t have time to address the changes right now.

PR #17376, opened 2026-05-31
fix(cli): add --skip-install flag to mastra migrate for immutable lockfile environments

### LINKED ISSUE #16957: `mastra migrate` runs its own npm/yarn/pnpm/bun install, bypassing immutable/locked dependency policies
### Describe the Bug

When running `mastra migrate` in CI, the command internally **re-bundles** the Mastra app and then **runs a package-manager install** in the generated `.mastra/output` directory. This behavior bypasses immutable/locked dependency policies (e.g. Yarn 4 `--immutable` / `YARN_ENABLE_IMMUTABLE_INSTALLS=true`) and can result in new or compromised packages being installed at migration time, even if the main workspace install is fully locked down.

**This is a supply‑chain/security concern for CI pipelines that rely on immutable installs and lockfiles to prevent unexpected dependency changes.**

## Affected code paths
mastra migrate CLI wiring

   The CLI command constructs a MigrateBundler and calls bundle:

   ```ts
   const bundler = new MigrateBundler(env);
   bundler.__setLogger(logger);

   // Prepare the output directory
   await bundler.prepare(dotMastraPath);

   // Bundle the project with migration entry
   const discoveredTools = bundler.getAllToolPaths(mastraDir, []);
   await bundler.bundle(entryFile, dotMastraPath, {
     toolsPaths: discoveredTools,
     projectRoot: rootDir,
   });
   ```

   [[migrate.ts](https://github.com/mastra-ai/mastra/blob/main/packages/cli/src/commands/migrate/migrate.ts#L83-L143)]

   MigrateBundler extends the base Bundler from @mastra/deployer and does not override installDependencies:
   ```ts
   export class MigrateBundler extends BuildBundler {
     // ...
     async bundle(
       entryFile: string,
       outputDirectory: string,
       { toolsPaths, projectRoot }: { toolsPaths: (string | string[])[]; projectRoot: string },
     ): Promise<void> {
       return this._bundle(this.getEntry(), entryFile, { outputDirectory, projectRoot }, toolsPaths);
     }
   }
```   

   [[MigrateBundler](https://github.com/mastra-ai/mastra/blob/main/packages/cli/src/commands/migrate/MigrateBundler.ts)]

Bundler _bundle method installs dependencies

   The base Bundler implementation in @mastra/deployer always installs dependencies as part of _bundle:

```ts
   this.logger.info('Installing dependencies');
   await this.installDependencies(outputDirectory, projectRoot);
   this.logger.info('Done installing dependencies');

   this.logger.info('Generating package-lock.json for deploy');
   await this.generateNpmLockfile(join(outputDirectory, this.outputDir));
   this.logger.info('Done generating package-lock.json');
```

   [[Bundler _bundle](https://github.com/mastra-ai/mastra/blob/main/packages/deployer/src/bundler/index.ts#L426-L494)]

   When anything in this stage fails (including the install), _bundle wraps it as:

```ts
   throw new MastraError(
     {
       id: 'DEPLOYER_BUNDLER_BUNDLE_STAGE_FAILED',
       text: `Failed during bundler bundle stage: ${message}`,
       domain: ErrorDomain.DEPLOYER,
       category: ErrorCategory.SYSTEM,
     },
     error,
   );
```

   [[Bundler _bundle](https://github.com/mastra-ai/mastra/blob/main/packages/deployer/src/bundler/index.ts#L304-L421)]

   This matches the error we see in CI:

   
`Error: Failed during bundler bundle stage: Process exited with code 1`

installDependencies delegates to DepsService.install

   The default installDependencies implementation is:
```ts
   protected async installDependencies(outputDirectory: string, rootDir = process.cwd()) {
     const deps = new DepsService(rootDir);
     deps.__setLogger(this.logger);

     await deps.install({ dir: join(outputDirectory, this.outputDir) });
   }
```

   [[Bundler deps](https://github.com/mastra-ai/mastra/blob/main/packages/deployer/src/bundler/index.ts#L99-L205)]

   DepsService.install (in @mastra/deployer) detects the package manager from lockfiles and runs the appropriate install command in the given directory. [[Deps.install](https://github.com/mastra-ai/mastra/blob/main/packages/deployer/src/services/deps.ts#L135-L189)]

   In other words, mastra migrate:

   
Writes a package.json into .mastra/output with the external dependencies it detected. [[Bundler _bundle](https://github.com/mastra-ai/mastra/blob/main/packages/deployer/src/bundler/index.ts#L304-L421)]
Runs npm|yarn|pnpm|bun install in that directory, resolving versions (sometimes falling back to 'latest' when it can’t resolve a version from the project). [[Bundler _bundle](https://github.com/mastra-ai/mastra/blob/main/packages/deployer/src/bundler/index.ts#L304-L421)]

Why this is a problem
In CI we:

- Use Yarn 4 with immutable installs (--immutable / YARN_ENABLE_IMMUTABLE_INSTALLS=true).
- Rely on a checked‑in lockfile to ensure no new packages or versions are resolved during CI.

However, mastra migrate:

- Performs a second install in .mastra/output, outside the main workspace.
- May resolve versions independently (including 'latest' fallbacks).
- Fails under Yarn 4 immutable mode, which is why we had to set YARN_ENABLE_IMMUTABLE_INSTALLS=false around the migration step.

This means that even if the main workspace install is fully locked and immutable, the migration step can still:

- Pull in new versions of dependencies.
- Potentially install compromised packages if the registry is attacked between runs. ‼️

From a supply‑chain/security perspective, this undermines the guarantees provided by immutable installs and lockfiles.

## Current workaround
Our only viable workaround today is to disable immutable installs for the mastra migrate step, e.g.:
```
- name: Run Mastra migrations
  run: npx mastra migrate --yes
  env:
    YARN_ENABLE_IMMUTABLE_INSTALLS: 'false'
    # plus DB/app env vars...
```

This is not ideal, because it re‑introduces the risk of unexpected dependency changes during CI.

## Summary

- mastra migrate currently always runs a package-manager install in .mastra/output via the shared Bundler implementation. [[Bundler _bundle](https://github.com/mastra-ai/mastra/blob/main/packages/deployer/src/bundler/index.ts#L426-L494); [Bundler deps](https://github.com/mastra-ai/mastra/blob/main/packages/deployer/src/bundler/index.ts#L99-L205)]
- This conflicts with immutable install policies and can bypass lockfile‑based supply‑chain protections.
- There is no documented way to disable or control this behavior for migrations.
- A --skip-install / “reuse existing deps” option (plus docs) would allow secure CI pipelines to use mastra migrate without weakening their dependency security posture.

### Steps To Reproduce

1. Run mastra migrate in GitHub actions

N/A

### Expected Behavior

It would be very helpful to have one or more of the following:

### A --skip-install (or similar) flag for mastra migrate
Skip the installDependencies step entirely.
Assume that all required dependencies are already available from the main workspace install.

### A “reuse existing deps” mode
Allow mastra migrate to bundle against the existing node_modules without writing a new package.json + running a fresh install in .mastra/output.

### Stronger lockfile integration
Optionally force the bundler to respect the existing workspace lockfile and fail if it would need to resolve any new versions (instead of falling back to 'latest').

### Documentation update
Explicitly document that mastra migrate runs its own package-manager install in .mastra/output.
Provide guidance for secure CI setups (Yarn 4, pnpm, etc.) and how to avoid or mitigate this behavior.

### Environment Information

```block
mastra CLI:          1.8.1
  @mastra/core:        1.32.1   (patched:
  ~/.yarn/patches/@mastra-core-npm-1.32.1-a05d167ef4.patch)
  @mastra/deployer:    1.32.1   ← contains the bundler that runs `yarn install`
  @mastra/memory:      1.17.5   (patched:
  ~/.yarn/patches/@mastra-memory-npm-1.17.5-05bea81bdc.patch)
  @mastra/observability: 1.11.1
  @mastra/pg:          1.10.0
  @mastra/mcp:         1.7.0
  @mastra/rag:         2.2.1
  @mastra/redis:       1.1.0
  @mastra/posthog:     1.0.23
  @mastra/duckdb:      1.3.0
  @mastra/loggers:     1.1.1   (transitive)
  @mastra/schema-compat: 1.2.9 (transitive)
  @mastra/server:      1.32.1  (transitive)

  Node:                v24.11.1
  Yarn:                4.13.0
  yarnrc nodeLinker:   node-modules
  OS:                  Ubuntu 24.04 (GitHub Actions ubuntu-24.04 runner)

  For the issue, the load-bearing file:lines you'll want to cite:
  - node_modules/@mastra/deployer/dist/chunk-4KIRQ3MQ.js:354-356 — base
  Bundler.bundle() calls installDependencies() (inherited by MigrateBundler).
  - node_modules/mastra/dist/index.js:1232-1383 — the migrate() function that
  re-bundles via MigrateBundler.
```

### Verification

SaarthurR:
Fixes #16957

`mastra migrate` always runs package installation internally, which breaks CI/CD pipelines that enforce immutable lockfiles (Yarn 4 `--immutable`, `npm ci`, `pnpm --frozen-lockfile`).

This adds a `--skip-install` flag and auto-detects common immutable-install environments:

- `--skip-install` CLI flag: explicit opt-out
- `YARN_ENABLE_IMMUTABLE_INSTALLS=true`: Yarn Berry's own env var
- `npm_command=ci`: set by npm v7+ when running `npm ci`
- `npm_config_ci=true`: manual fallback for older npm setups

When skipping, `MigrateBundler.installDependencies()` is a no-op and logs an info message.

## ELI5

This PR adds a `--skip-install` flag to the `mastra migrate` command so that it won't try to install packages during migration when you're in a strict CI environment that locks your dependencies (like Yarn 4 or npm ci). It also automatically detects these strict environments, so you don't even have to use the flag if you don't want to.

## Overview

This PR adds a `--skip-install` CLI flag to `mastra migrate` that prevents the command from running dependency installation in immutable lockfile environments, addressing CI/CD failures when strict lockfile policies are enforced.

### CLI Interface
- Added `--skip-install` flag to the `migrate` command in the CLI
- Updated the `migrate` command interface to accept an optional `skipInstall` boolean parameter

### Core Implementation
- Extended `MigrateBundler` to accept a `skipInstall` constructor flag
- Modified `MigrateBundler.installDependencies()` to become a no-op when:
  - The `--skip-install` flag is explicitly set, OR
  - Immutable lockfile environments are detected via environment variables:
    - `YARN_ENABLE_IMMUTABLE_INSTALLS=true` (Yarn 4 immutable mode)
    - `npm_command=ci` (npm ci command)
    - `npm_config_ci=true` (npm ci configuration)
  - When skipping, logs an informational message
- Updated the `migrate` runner to pass the `skipInstall` flag through to `MigrateBundler`

### Testing
- Added comprehensive test suite for `MigrateBundler.installDependencies()` covering:
  - Cases where installation should be delegated to parent implementation
  - Cases where installation should be skipped based on the `skipInstall` flag
  - Auto-detection of immutable lockfile environments via each of the three environment variable patterns
  - Verification that informational logging occurs when skipping

### Documentation
- Updated changeset to document the new `--skip-install` flag for the `migrate` command

## Addresses

Fixes issue `#16957` by providing an explicit opt-out mechanism and auto-detection of immutable install modes to prevent `mastra migrate` from bypassing workspace lockfile guarantees and potential supply-chain issues in strict CI environments.

### DIFF
--- a/.changeset/plain-aliens-cover.md
+++ b/.changeset/plain-aliens-cover.md
@@ -1,0 +1,5 @@
+---
+'mastra': patch
+---
+
+Add --skip-install flag to migrate command to skip dependency installation in immutable lockfile environments (Yarn 4 --immutable, npm ci)
--- a/packages/cli/src/commands/actions/migrate.ts
+++ b/packages/cli/src/commands/actions/migrate.ts
@@ -1,7 +1,14 @@
 import { analytics, origin } from '../..';
 import { migrate as runMigrate } from '../migrate/migrate';
 
-export const migrate = async (args: { dir?: string; root?: string; env?: string; debug?: boolean; yes?: boolean }) => {
+export const migrate = async (args: {
+  dir?: string;
+  root?: string;
+  env?: string;
+  debug?: boolean;
+  yes?: boolean;
+  skipInstall?: boolean;
+}) => {
   await analytics.trackCommandExecution({
     command: 'mastra migrate',
     args: { ...args },
@@ -12,6 +19,7 @@ export const migrate = async (args: { dir?: string; root?: string; env?: string;
         env: args?.env,
         debug: args?.debug ?? false,
         yes: args?.yes ?? false,
+        skipInstall: args?.skipInstall,
       });
     },
     origin,
--- a/packages/cli/src/commands/migrate/MigrateBundler.test.ts
+++ b/packages/cli/src/commands/migrate/MigrateBundler.test.ts
@@ -11,6 +11,7 @@ vi.mock('@mastra/deployer/build', () => ({
 vi.mock('../build/BuildBundler.js', () => ({
   BuildBundler: class MockBuildBundler {
     protected platform = 'node';
+    protected logger: { info: (...args: unknown[]) => void } = { info: vi.fn() };
     constructor(_options?: { studio?: boolean }) {}
     __setLogger(_logger: unknown) {}
     getAllToolPaths(_dir: string, _extra: unknown[]) {
@@ -23,6 +24,9 @@ vi.mock('../build/BuildBundler.js', () => ({
       return Promise.resolve(new Map());
     }
     protected _bundle(_entry: string, _entryFile: string, _options: unknown, _toolsPaths: unknown): Promise<void> {
+      return Promise.resolve();
+    }
+    protected installDependencies(_outputDirectory: string, _rootDir?: string): Promise<void> {
       return Promise.resolve();
     }
   },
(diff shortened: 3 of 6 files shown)


abhiaiyer91:
I closed this as won’t fix. We don’t want to open PRs for closed issues

PR #17379, opened 2026-05-31
fix(observability): use configured model alias for OpenRouter pricing lookup

### LINKED ISSUE #16584: [BUG] OpenRouter responseModel uses different word ordering than pricing-registry (Anthropic models still miss after PR #16206)

serhiizghama:
## Problem

When using Anthropic models through OpenRouter (e.g. `openrouter/anthropic/claude-sonnet-4-6`), estimated cost shows as `null` in Studio's Model Usage & Cost panel with `costMetadata: { error: 'no_matching_model' }`.

Closes #16584

Root cause: `emitUsageMetrics` in `auto-extract.ts` prefers `attrs.responseModel` over `attrs.model` for the pricing registry lookup. For OpenRouter routes, the two diverge:

- `attrs.model` (configured alias): `anthropic/claude-sonnet-4-6`
- `attrs.responseModel` (downstream internal id): `anthropic/claude-4.6-sonnet-20260217`

The variants generated from the response id (`claude-4-6-sonnet`, `claude-4.6-sonnet`, …) never match `claude-sonnet-4-6` — the word ordering is different, and no textual transform bridges it. Lookup returns `no_matching_model`.

## Solution

For OpenRouter, the configured `model` alias is what the pricing registry stores (it's also what OpenRouter bills the user for). Using the response model id is a layer-mixing error: we're trying to price a downstream provider's internal snapshot name against OpenRouter's alias-keyed pricing data.

One-line change in `emitUsageMetrics`:

```typescript
// Before (broken for OpenRouter Anthropic models)
const model = attrs.responseModel ?? attrs.model;

// After
const model =
  provider === 'openrouter'
    ? (attrs.model ?? attrs.responseModel)
    : (attrs.responseModel ?? attrs.model);
```

For all non-OpenRouter providers the existing behaviour (prefer `responseModel`, fall back to `model`) is preserved.

## Testing

Two regression tests added to `estimator.test.ts`:

1. **Shows the failure**: looking up `anthropic/claude-4.6-sonnet-20260217` (the response id) returns `no_matching_model` — confirming the root cause.
2. **Shows the fix**: looking up `anthropic/claude-sonnet-4-6` (the configured alias) resolves correctly and returns `estimatedCost > 0`.

```
✓ fails to resolve OpenRouter Anthropic models by responseModel id (word-order mismatch)
✓ resolves OpenRouter Anthropic models when the configured alias is used
```

## ELI5

When you use Anthropic models through OpenRouter in Studio, the cost wasn't showing up correctly. This happened because the system was looking up pricing information using the wrong name for the model—it was using an internal backend ID instead of the name you actually configured. This PR fixes it by making the system use the correct configured name for lookups, so costs now display properly.

## Problem

When using Anthropic models via OpenRouter, Studio's Model Usage & Cost panel displayed `estimatedCost` as `null` with error metadata `{ error: 'no_matching_model' }`. The root cause: `emitUsageMetrics` was preferring `attrs.responseModel` over `attrs.model` for pricing lookups. For OpenRouter routes, these differ—`attrs.model` contains the configured alias (e.g., `anthropic/claude-sonnet-4-6`), while `attrs.responseModel` contains the downstream internal ID with different word ordering (e.g., `anthropic/claude-4.6-sonnet-20260217`). Variants generated from the response ID don't match the alias keys in the pricing registry, causing lookup failures.

## Solution

Modified `emitUsageMetrics` to select the model identifier based on provider:
- **For OpenRouter**: Use `attrs.model` (the configured alias) as the primary option, falling back to `attrs.responseModel` if needed
- **For other providers**: Maintain existing behavior—use `attrs.responseModel` first, then fall back to `attrs.model`

This preserves backward compatibility for non-OpenRouter providers while using the configured alias for OpenRouter, which matches both the pricing registry keys and what OpenRouter actually bills.

## Changes

**observability/mastra/src/metrics/auto-extract.ts**
- Updated `emitUsageMetrics` to conditionally select model identifier based on provider, with inline documentation explaining why the configured alias is preferred for OpenRouter

**observability/mastra/src/metrics/__fixtures__/pricing-data-test.jsonl**
- Added four new test pricing records for OpenRouter models, including `anthropic/claude-sonnet-4-6`

**observability/mastra/src/metrics/estimator.test.ts**
- Added two regression tests:
  1. Confirms that pricing lookup fails when using the responseModel ID (demonstrating the word-order mismatch issue)
  2. Confirms that pricing lookup succeeds and returns `estimatedCost > 0` when using the configured alias

**.changeset/brave-meals-add.md**
- Changeset documenting the fix and marking packages for patch version bump

## Testing

Two regression tests verify the fix:
- Test 1: Demonstrates lookup failure with responseModel ID (`anthropic/claude-4.6-sonnet-20260217`), resulting in `no_matching_model` error
- Test 2: Confirms successful lookup with configured alias (`anthropic/claude-sonnet-4-6`), returning proper estimatedCost

Closes `#16584`.


TheIsrael1:
Closing this as a duplicate of #16585, which covers the same OpenRouter pricing lookup issue and is the PR we are continuing with.

PR #17381, opened 2026-05-31
fix(pg): batch saveMessages into chunked multi-row INSERT

### LINKED ISSUE #17350: [BUG] Postgres saveMessages inserts each message in a separate round trip

SaarthurR:
Fixes #17350

`saveMessages()` was executing one `INSERT` per message inside a transaction, causing N sequential round trips for N messages.

This replaces the per-message loop with chunked multi-row `INSERT ... VALUES (...), (...), ...`. Chunk size is `floor(65535 / 8) = 8191` rows (Postgres bind-parameter cap at 8 params per row). For most workloads this reduces N round trips to 1.

The validation loop (checking `threadId`/`resourceId` on each message) still runs before the batch INSERT. The thread `updatedAt` UPDATE is unchanged.

## ELI5

Instead of telling the database "insert this message" one at a time (which is slow because each message needs its own trip to the database), the code now groups messages together and says "insert all of these at once" in batches. This is much faster because it reduces the number of trips from many down to just a few.

## Summary

This PR optimizes the `saveMessages` method in the Postgres adapter by replacing per-message `INSERT ... ON CONFLICT` statements with chunked multi-row batch inserts.

### Changes

**Memory Storage Optimization** (`stores/pg/src/storage/domains/memory/index.ts`)
- Replaced sequential single-row INSERT statements inside a transaction with batched multi-row `INSERT ... VALUES (...), (...) ON CONFLICT (id) DO UPDATE` statements
- Chunk size set to 8,191 rows (computed as `floor(65535 / 8)`) to respect Postgres bind-parameter limits while maintaining 8 parameters per row
- Per-message validation for `threadId` and `resourceId` preserved before batch insertion
- Thread timestamp (`updatedAt`) updates remain unchanged
- Reduces database round trips from N (one per message) to approximately 1 for typical workloads

**Test Coverage** (`stores/pg/src/storage/domains/memory/save-messages-batch.test.ts`)
- Added comprehensive test suite validating batch insert behavior
- Confirms single `t.none` call for 10 messages with correct parameter binding
- Validates chunking logic when messages exceed chunk size (multiple `t.none` calls)
- Verifies `CHUNK_SIZE` constant equals 8,191
- Confirms `ON CONFLICT` clause preservation with correct column ordering
- Tests edge cases including empty message arrays

**Release Notes** (`.changeset/better-buses-turn.md`)
- Patch release entry documenting the optimization

### Impact

- Preserves existing upsert semantics (`ON CONFLICT (id)` updates for thread_id, content, role, type, and resourceId)
- Maintains transaction atomicity including thread timestamp updates
- Handles duplicate ids within single `saveMessages` calls correctly
- Reduces database latency for batch message persistence operations

### DIFF
--- a/.changeset/better-buses-turn.md
+++ b/.changeset/better-buses-turn.md
@@ -1,0 +1,5 @@
+---
+'@mastra/pg': patch
+---
+
+Fix Postgres saveMessages to use a single batched INSERT instead of one round trip per message
--- a/stores/pg/src/storage/domains/memory/index.ts
+++ b/stores/pg/src/storage/domains/memory/index.ts
@@ -1212,7 +1212,6 @@ export class MemoryPG extends MemoryStorage {
     try {
       const tableName = getTableName({ indexName: TABLE_MESSAGES, schemaName: getSchemaName(this.#schema) });
       await this.#db.client.tx(async t => {
-        // Insert messages sequentially to avoid concurrent queries on the same pg client
         for (const message of messages) {
           if (!message.threadId) {
             throw new Error(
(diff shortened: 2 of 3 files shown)


abhiaiyer91:
Another issue that already has a fix. Don’t open PRs on closed issues

PR #17386, opened 2026-06-01
fix(schema-compat): strip typeless optional properties from Gemini tool schemas

### LINKED ISSUE #17325: [BUG] supervisor agent fails using gemini model through openrouter
### Describe the Bug

error occurs when using supervisor agent with openrouter gemini model

this is nearly identical to this issue: https://github.com/mastra-ai/mastra/issues/13988 and was marked as fixed by https://github.com/LiorVainer/data-israel/commit/652250434a52d44fbce7bd536e7a1d3d45edf4c0

the supervisor agent works with gemini models when using `google/gemini-3.1-flash-lite` and `vercel/google/gemini-3.1-flash-lite` but fails when using `openrouter/google/gemini-3.1-flash-lite`

```typescript

const supervisorAgent = new Agent({
  id: 'supervisor-agent',
  name: 'Supervisor Agent',
  instructions: 'You are a helpful assistant',
  // model: 'google/gemini-3.1-flash-lite', // works
  // model: 'vercel/google-gemini-3.1-flash-lite', // works
  model: 'openrouter/google/gemini-3.1-flash-lite', // error: INVALID_ARGUMENT, GenerateContentRequest.tools[0].function_declarations[0].parameters.required[0]: property is not defined
  agents: { subAgent }
})
```

produces the error: 

```json
{
  "error": {
    "message": "Provider returned error",
    "code": 400,
    "metadata": {
      "raw": {
        "error": {
          "code": 400,
          "message": "* GenerateContentRequest.tools[0].function_declarations[0].parameters.required[0]: property is not defined",
          "status": "INVALID_ARGUMENT"
        }
      },
      "provider_name": "Google AI Studio",
      "is_byok": false
    }
  },
  "user_id": "omitted"
}
```

### Steps To Reproduce

see https://github.com/alexbatis/mastra-gemini-supervisor-repro/blob/933fdde2b9eb127bf6aec4d0c22601ede39ac3b1/README.md

https://github.com/alexbatis/mastra-gemini-supervisor-repro

### Expected Behavior

the agent should generate a response

### Environment Information

```block
- LLM Provider: google gemini model used through openrouter

System:
  OS: macOS
Binaries:
  bun: 1.3.11
npmPackages:
  @mastra/core: ^1.37.1 => 1.37.1 
  @types/bun: latest => 1.3.14
```

### Verification

serhiizghama:
## Problem

Supervisor agents using `openrouter/google/gemini-*` models fail with HTTP 400 INVALID_ARGUMENT:

```
GenerateContentRequest.tools[0].function_declarations[0].parameters.required[0]: property is not defined
```

The error is misleading — `prompt` (the valid required property) is reported as "not defined" even though it's perfectly typed as `{ type: 'string' }`.

**Root cause**: `CoreToolBuilder` injects `resumeData: z.any().optional()` into every resumable tool's input schema (including `agent-*` delegation tools). `z.any()` serialises to an empty JSON schema `{}` — no `type`, `anyOf`, or any other keyword. Gemini's OpenAPI 3.0 Schema validator (reached via OpenRouter) requires every property to have a valid schema. When it encounters `resumeData: {}`, it considers the entire function declaration invalid, so it reports the first required property as "not defined".

**Why direct `google/gemini-*` works**: The `@ai-sdk/google` SDK processes schemas internally before sending to the Gemini REST API and tolerates or strips empty schemas. OpenRouter passes the schema as-is, exposing the invalid property.

## Solution

`GoogleSchemaCompatLayer` now removes optional properties that have no Gemini-compatible schema keyword (`type`, `anyOf`, `oneOf`, `allOf`, `$ref`, `enum`) from:

1. `fixNullableUnionTypes` — covers the `processToAISDKSchema` code path  
2. `postProcessJSONNode` — covers the `processToJSONSchema` code path

Required properties (present in `required`) are never removed.

## Testing

- Added regression test: `'removes z.any() optional properties from supervisor agent tool schemas (OpenRouter Gemini fix)'` in `google.test.ts` — validates the exact schema shape from issue #17325.
- Updated existing `'drops optional properties with no representable Gemini type'` test — reflects correct behavior (property removed rather than left as `{}`).
- Updated existing agent delegation pattern test to assert `resumeData` is stripped.
- All 876 `@mastra/schema-compat` tests pass.

Fixes #17325

## ELI5

Gemini rejects tool fields without a type. This PR gives those fields a flexible schema instead of removing them, so OpenRouter requests work without breaking suspend/resume support.

## Problem

Supervisor agents using Gemini through OpenRouter can receive HTTP 400 `INVALID_ARGUMENT` errors when `z.any().optional()` creates a typeless schema.

## Solution

`GoogleSchemaCompatLayer.postProcessJSONNode` now:

- Converts typeless schemas into nullable `anyOf` schemas.
- Supports string, number, integer, boolean, object, and array values.
- Preserves fields such as `resumeData`.
- Preserves descriptions and nullable handling.
- Converts multi-type schemas into `anyOf` variants.
- Keeps required properties available for validation.

## Changes

- Updated `packages/schema-compat/src/provider-compats/google.ts`.
- Added regression coverage in `packages/schema-compat/src/provider-compats/google.test.ts`.
- Added a patch changeset for `@mastra/schema-compat`.

## Testing

- The schema-compatibility suite passes with 824 tests.
- Direct-Google request recordings may require regeneration because SDK normalization adds a redundant `nullable` field.

Fixes `#17325`.


serhiizghama:
Rebased on main. The compat layer was reworked quite a bit since I opened this, so I re-homed the fix to fit the new pipeline: the typeless-optional-property strip now runs in `preProcessJSONNode` (before type arrays are collapsed) rather than the old `fixNullableUnionTypes` spot.

That ordering is deliberate — a property that starts as `type: ['string', 'null']` still has a type at pre-process time, so it survives and your `drops type and nullable for multi-non-null type arrays` test keeps passing, while a genuinely typeless `z.any().optional()` `{}` gets stripped. Re-added the #17325 regression under the `processToJSONSchema — OpenAPI 3.0 compat` suite. Full schema-compat suite is green (823 passing).

serhiizghama:
You're right, and the A/B evidence is convincing — I reproduced the core e2e failures. Deleting `resumeData` does break suspend/resume: the recording shows the live model filling it in, and `shouldApply()` matching any `google`/`gemini-` id means this hits direct Gemini too, not just the OpenRouter path I was chasing. The CI green was misleading because the changed-file gate scoped it to schema-compat and never ran core e2e.

I'm on board with rewriting typeless properties into the permissive `anyOf` + `nullable: true` form in `postProcessJSONNode` instead of dropping them — that keeps the field functional, fixes OpenRouter (which forwards the schema literally instead of letting `@ai-sdk/google` rewrite the type array), and matches the shape Gemini already accepts in the recording.

The one thing I want a maintainer's call on before I push it: this flips `drops type and nullable for multi-non-null type arrays` from #17057 — that rule ("drop `type` entirely") is itself what produces the typeless property Gemini chokes on, so "emit `anyOf`" would supersede it rather than sit alongside. Is that the intended direction, or does #17057's drop-type behavior need to hold for some transport I'm not seeing? Once that's settled I'll redo the diff, cover the required-multi-type case you flagged (typeless-after-postprocess still slips through today), and re-run the core Gemini e2e rather than leaning on the scoped gate.

serhiizghama:
Went with your suggested direction. Typeless properties are now rewritten into the permissive `anyOf` in `postProcessJSONNode` instead of being deleted, so `resumeData` (and any other `z.any()` field the model is expected to fill) stays in the tool contract. I also dropped the equivalent silent-delete I'd added to the AI-SDK path, so that path is back to letting `@ai-sdk/google` type the empty schema.

On the open question I flagged — whether emit-`anyOf` should supersede #17057's "drop `type` for multi-non-null arrays" — I went ahead with it, since it's the only shape that actually fixes the OpenRouter path (a bare typeless property is exactly what Gemini rejects there). The two assertions you predicted would flip are updated (this PR's own test and `drops type and nullable for multi-non-null type arrays`), plus I added coverage for the required-multi-type case you flagged as still slipping through. schema-compat is green at 824 passed.

One heads-up on the core Gemini e2e: the compat layer now emits `resumeData` as an explicit `anyOf` + `nullable: true` — I verified it produces exactly the shape in the recording. On the direct-Google path `@ai-sdk/google` re-normalizes that and drops the now-redundant top-level `nullable`, so the recorded request hashes no longer match; the only diff is that one `nullable`. The field is present and suspend/resume works — it's a stale recording, not the functional regression from before. Those need regenerating with live keys, which I can't run here. Happy for CI to re-record, or if you'd rather I gate the rewrite so the direct-Google request stays byte-identical, just say.

serhiizghama:
Good catch on the array case — z.any() accepts arrays but the fallback anyOf only had scalar/object variants, so an array-valued resumeData would've failed Gemini's schema validation. Added an array branch with a permissive items schema, plus a test asserting it's there. Also reworded the changeset away from the wildcard model pattern.

Worth noting: this fallback schema doesn't actually drive runtime traversal (#resolveAnyOf just grabs the first non-null anyOf branch for #traverse, so array/object values already passed through untouched) — the array branch matters purely for Gemini's own schema validation on the request side, not for our coercion logic.

abhiaiyer91 (approved):
Did a deep review of this one — applied the diff locally and verified beyond the included tests. Approving.

**What I verified locally:**
- Full schema-compat suite with the diff applied: 834 passed / 9 skipped, including the direct-Google e2e replays (recordings unchanged — no behavior shift for schemas that already worked). Typecheck clean.
- Both serialization paths are covered: `mode: 'jsonSchema'` and `mode: 'aiSdkSchema'` converge on `processToJSONSchema` → `postProcessJSONNode`, and `fixAISDKNullableUnionTypes` doesn't undo the permissive `anyOf` (no `type: 'null'` variant inside it, so the collapse branch never fires).
- Edge probes: nested `z.any()` (inside objects and array items) gets sanitized at all depths, and nodes left typeless by the `additionalProperties` stripping (raw JSON / MCP record-style schemas) are now fixed too — a bonus beyond the reported issue.

**Why I like the approach:** the catch-all's guard means it only matches nodes Gemini was *already rejecting* — previously-working schemas can't reach it, so the regression surface is essentially zero. And keeping fields like `resumeData` in the contract as a permissive `anyOf` (rather than stripping them, as an earlier revision did) preserves tool suspend/resume, which is the right call. This also correctly reverses the earlier "drop `type` for multi-type arrays" resolution, which left a typeless node Gemini rejects — that was the residual half of this bug class.

**Non-blocking notes:**
- The PR title/description still describe the earlier "strip" approach — the final code rewrites into `anyOf`, never strips. Mind updating the title? We'll make sure the squash commit message is accurate either way (the changeset is already correct).
- `not`-only nodes still leak through untouched (pre-existing, out of scope — noting for posterity).

Thanks for the thorough work here, including the well-commented tests pinning each Gemini constraint. 🙏

PR #17399, opened 2026-06-01
perf(pg): batch Postgres storage inserts

### LINKED ISSUE #17398: [Storage] Batch Postgres storage inserts into fewer round trips
## Summary

`PgDB.batchInsert` currently runs one `INSERT` statement per record inside a transaction. For callers that already pass batches, such as Postgres observability span creation, a batch of N records still costs N sequential PostgreSQL statements.

## Why this matters

Observability span volume scales with agent, tool, workflow, retry, and tracing activity. In busy server apps, span batches can become sustained write volume. Paying one database round trip per record adds avoidable connection and query overhead.

## Current behavior

`stores/pg/src/storage/db/index.ts` loops through `records` and awaits `executeInsert` for each record.

That preserves behavior, but it means a 1,000-record batch still sends 1,000 insert statements.

## Expected behavior

`batchInsert` should insert compatible rows with multi-row parameterized `INSERT` statements while preserving existing behavior:

- keep the existing transaction boundary,
- preserve column filtering for migrated or partially upgraded schemas,
- preserve span `ON CONFLICT ("traceId", "spanId")` behavior,
- chunk statements below PostgreSQL bind-parameter limits,
- handle duplicate span IDs in the same input batch safely.

## Notes

Duplicate `(traceId, spanId)` values inside one input batch need special handling because PostgreSQL cannot update the same conflict target twice in one multi-row `ON CONFLICT DO UPDATE` statement.

mbenhamd:
## Description

Updates `PgDB.batchInsert` to group compatible records into multi-row parameterized `INSERT` statements instead of sending one insert statement per record.

This targets batched Postgres storage writes, especially observability span ingestion. The implementation keeps the existing transaction boundary and reuses the same record preparation path as single-row inserts.

The batching logic:

- filters records against known table columns before building SQL
- canonicalizes column order so equivalent records batch together
- chunks statements below PostgreSQL bind-parameter limits
- preserves span `ON CONFLICT ("traceId", "spanId")` upsert behavior
- falls back to serial span inserts when the same `(traceId, spanId)` appears more than once in a single input batch

Local Docker Postgres benchmark:

The old path sends one insert statement per record, so the record count is also the old statement count.

| records | new statements | old median | new median |
| ---: | ---: | ---: | ---: |
| 100 | 1 | 22.0ms | 2.9ms |
| 1,000 | 1 | 192.2ms | 12.9ms |
| 5,000 | 1 | 906.1ms | 62.6ms |
| 12,001 | 2 | 2138.2ms | 134.2ms |

## Related issue(s)

Fixes #17398

## Type of change

- [ ] Documentation update
- [ ] Code refactoring
- [x] Performance improvement
- [x] Test update

## Checklist

## Validation

- `pnpm --filter ./stores/pg exec vitest run src/storage/db/transaction-client.test.ts`
- `pnpm --filter ./stores/pg exec vitest run src/storage/migration.test.ts -t "batch insert"`
- `pnpm --filter ./stores/pg build:lib`
- `pnpm --filter ./stores/pg lint`
- `pnpm prettier --check stores/pg/src/storage/db/index.ts stores/pg/src/storage/db/transaction-client.test.ts .changeset/pg-batch-insert-throughput.md`
- `TZ=UTC pnpm --filter ./stores/pg test`

## ELI5

Instead of saving each record with its own database request (like sending one letter per truck), this PR groups compatible records and inserts many at once (like loading many letters into one truck), making bulk saves much faster.

## Overview

This PR implements batching for Postgres storage inserts in PgDB.batchInsert: compatible records are grouped into multi-row parameterized INSERT statements rather than issuing one INSERT per record. The change preserves transaction boundaries, existing column-filtering behavior, and span upsert semantics, and it chunks statements to stay under PostgreSQL bind-parameter limits.

Performance (local Docker Postgres median latencies):
- 100 records: 22.0ms → 2.9ms (1 statement)
- 1,000 records: 192.2ms → 12.9ms (1 statement)
- 5,000 records: 906.1ms → 62.6ms (1 statement)
- 12,001 records: 2138.2ms → 134.2ms (2 statements)

Fixes: `#17398`

## Changes

stores/pg/src/storage/db/index.ts
- Added bind-parameter-limit-aware batching path and refactored single-row & batch insert flow into a shared "prepare → build query → execute" pipeline.
- New helpers: prepareInsertRecord, buildInsertQuery; introduced BATCH_INSERT_MAX_PARAMETERS and helpers for filtering/normalizing/sorting record columns and values.
- Filters records to known table columns and canonicalizes column order so equivalent records batch together.
- Generates multi-row INSERTs with:
  - ON CONFLICT ("traceId","spanId") DO UPDATE when non-PK columns exist,
  - DO NOTHING when only PK/conflict columns are present.
- Chunks multi-row statements to keep total bind parameters under PostgreSQL limits.
- For spans, detects duplicate (traceId, spanId) within the same input batch and falls back to serial inserts (after ordering by conflict key) to avoid invalid multi-row ON CONFLICT updates.
- executeInsert updated to use the new pipeline and skip inserts when no known columns remain; prepareValuesForInsert now operates on sorted key/value entries.
- No exported/public API changes.

stores/pg/src/storage/db/transaction-client.test.ts
- Expanded and revised tests to assert multi-row INSERT shapes, placeholder grouping, total bound parameter counts, filtering by cached table columns, batching across mixed/incompatible record shapes, parameter-limit splitting, and span-specific upsert behaviors.
- Tests for ordering by conflict keys, DO NOTHING behavior when only conflict-key columns are present, and fallback to serial inserts on duplicate span keys.
- Enhanced mock DbClient/TxClient harness with configurable cached tableColumns and query/tx spies to capture executed SQL and bound parameters.
- Test-only changes; no public API changes.

stores/pg/src/storage/migration.test.ts
- Added Vitest cases covering:
  - batchCreateSpans handling duplicate (traceId, spanId) within a single batch (deduplication, last-write retention).
  - Concurrent batch upserts with overlapping spans and differing record order resolving to a valid outcome.
- Test-only changes.

.changeset/pg-batch-insert-throughput.md
- Added changeset documenting a patch bump for `@mastra/pg` and noting reduced Postgres round trips from batched inserts.

## Impact

- Performance: Significant throughput improvement for bulk inserts (observed up to ~16x in local benchmarks).
- Backward compatibility: Transaction semantics, column filtering, and span upsert behavior preserved.
- Safety: Batches are chunked to respect parameter limits; duplicate span keys within a single input batch trigger a safe serial-insert fallback.

### DIFF
--- a/.changeset/pg-batch-insert-throughput.md
+++ b/.changeset/pg-batch-insert-throughput.md
@@ -1,0 +1,5 @@
+---
+'@mastra/pg': patch
+---
+
+Reduced Postgres round trips when batch inserting multiple storage records.
--- a/stores/pg/src/storage/db/index.ts
+++ b/stores/pg/src/storage/db/index.ts
@@ -22,6 +22,10 @@ import type { DbClient, QueryValues, TxClient } from '../client';
 import { PoolAdapter } from '../client';
 import { buildConstraintName } from './constraint-utils';
 
+// PostgreSQL's extended query protocol supports at most 65,535 bind parameters.
+// Keep chunked batch inserts below that limit with some headroom.
+const BATCH_INSERT_MAX_PARAMETERS = 60_000;
+
 // Re-export DbClient for external use
 export type { DbClient } from '../client';
 
@@ -163,6 +167,11 @@ function mapToSqlType(type: StorageColumn['type']): string {
       return getSqlType(type);
   }
 }
+
+type PreparedInsertRecord = {
+  columns: string[];
+  values: QueryValues;
+};
 
 export function generateTableSQL({
   tableName,
@@ -441,8 +450,8 @@ export class PgDB extends MastraBase {
   /**
    * Prepares values for insertion, handling JSONB columns by stringifying them
    */
-  private prepareValuesForInsert(record: Record<string, any>, tableName: TABLE_NAMES): any[] {
-    return Object.entries(record).map(([key, value]) => {
+  private prepareValuesForInsert(entries: [string, any][], tableName: TABLE_NAMES): QueryValues {
+    return entries.map(([key, value]) => {
       const schema = TABLE_SCHEMAS[tableName];
       const columnSchema = schema?.[key];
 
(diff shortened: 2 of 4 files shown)


abhiaiyer91:
This increases complexity for unknown gain, we should try and see if this is valuable and simplify how its done

PR #17404, opened 2026-06-01
fix(core): preserve whitespace in text parts when no working memory tags are stripped

### LINKED ISSUE #15880: filterMessagesForPersistence trims leading/trailing whitespace from individual text parts

Oxygen56:
## Description

`filterMessagesForPersistence` unconditionally trims whitespace from all text parts via `.trim()` after `removeWorkingMemoryTags()`. When the LLM tokenizer splits text at word boundaries during streaming, individual text parts may contain meaningful leading whitespace (e.g. `" access"`). The unconditional trim strips these spaces, causing concatenated output to lose word boundaries.

## Fix

The `.trim()` is now only applied when `removeWorkingMemoryTags()` actually modified the text. For text parts with no working memory tags, the original whitespace is preserved.

### Before
```ts
text: removeWorkingMemoryTags(text).trim(),  // strips all whitespace
```

### After
```ts
const cleaned = removeWorkingMemoryTags(text);
text: cleaned !== text ? cleaned.trim() : text,  // only trim when tags were removed
```

Fixes #15880

## Testing

A changeset has been included for a `@mastra/core` patch release.

## ELI5 (Explain Like I'm 5)

When assembling a sentence from smaller pieces, this PR stops removing spaces around each piece unless they actually need to be removed—so words don't accidentally get glued together.

## Problem

filterMessagesForPersistence was unconditionally calling .trim() on every text part after removing working memory tags. With span-based persistence (one text part per streaming span), trimming removed meaningful leading/trailing whitespace (e.g., " access"), causing concatenated persisted messages to lose spaces and produce glued words (["I have"," access"," to"] → ["I have","access","to"]).

## Root Cause

.trim() ran even when removeWorkingMemoryTags() made no changes, stripping single-space token boundaries produced by span-split text parts.

## Solution

Apply trimming only when removeWorkingMemoryTags() actually changes the text:
- Compute const cleaned = removeWorkingMemoryTags(text)
- If cleaned !== text, use cleaned.trim(); otherwise preserve the original text (and its whitespace)

This is applied to both the main content string and individual text parts.

## Changes

- packages/core/src/processors/memory/message-history.ts: conditionally trim cleaned text instead of unconditionally trimming (preserves span-boundary whitespace).  
- packages/core/src/processors/memory/message-history.test.ts: added regression tests verifying preservation of whitespace for parts without working memory tags and trimming when tags are present.  
- .changeset/curvy-pianos-whisper.md: patch changeset for `@mastra/core` documenting the bug fix.

## Impact

Fixes the regression that caused persisted/ recalled messages to lose token-boundary spaces when content was split into span-based parts; prevents concatenated strings from missing spaces.

## References

Fixes issue `#15880`


Oxygen56:
CodeRabbit review passed with no issues. Small whitespace fix (+9/-2), ready for human review when someone has time.

abhiaiyer91 (approved):

PR #17411, opened 2026-06-01
feat(core): compose Harness v1 session state

wardpeet:
This adds Harness v1 state and workspace context support, then updates Mastra Code compatibility so session-owned model values come from the active session instead of mutable harness state.

Harness state still holds app-level config like project paths and permission rules. Session-derived fields like `currentModelId`, `modeId`, and the default `subagentModelId` are composed by `HarnessCompat.getState()` so existing Mastra Code callers can keep reading the same shape.

```ts
const state = harnessCompat.getState();
state.currentModelId; // from session.getModelId()
state.modeId; // from session.getMode().id
state.subagentModelId; // from session.getSubagentModelId()
```

Verified with focused Harness v1 tests, HarnessCompat tests, core typecheck, and the core build.

## ELI5

This PR moves session-specific settings (which model/mode a session uses and the session workspace) into each session instead of a global harness state, while keeping the old state shape for existing callers by composing those session values when requested.

## Overview

Adds typed per-session state and workspace support to Harness v1 and updates Mastra Code compatibility so session-owned model/mode values are sourced from the active session. Harness continues to hold app-level configuration (project paths, permission rules). HarnessCompat.getState() composes session-derived fields into the legacy state shape so callers see the same fields:

- state.currentModelId ← session.getModelId()
- state.modeId ← session.getMode().id
- state.subagentModelId ← session.getSubagentModelId()

setState routing now directs session-owned fields (currentModelId, modeId) to the v1 Session (including mode switching) while applying other keys to both session and legacy harness state. Observability was adjusted to stop recording subagentModelId in spans.

## Architecture Changes

- Harness and Session are generic over TState and accept optional stateSchema and initialState; workspace can be an instance, factory, or config object.
- Session<TState> manages typed in-memory state with public APIs: getState(), setState(), updateState(). Updates are serialized, validated against schema, produce frozen snapshots, and emit a state_changed event with changed keys.
- Harness threads stateSchema/initialState/workspace into session creation and exposes getWorkspace() when workspace is concrete.
- HarnessCompat composes session-derived model/mode fields, special-cases getState/setState/switchThread/cloneSession to preserve session model selection across thread/session operations, and adds useSkill passthrough.

## Key Changes by Component

- packages/core/src/harness/v1
  - Session<TState>: added typed state handling, getState/setState/updateState, validation, serialized updates, emits StateChangedEvent.
  - Harness<MODES, TState>: new generic, accepts stateSchema/initialState/workspace, session APIs return Session<TState>, added getWorkspace() and various lookups (subagents/skills/models/agents/tool categories).
  - New StateChangedEvent added to HarnessEvent union; RESERVED_EVENT_TYPES updated.
  - New v1 tooling: buildSessionToolsets, permission/tool types, skills/subagent typings and helpers.
  - Tests: harness.state-workspace.test.ts, harness.tools.test.ts, skills.test.ts, tools.test.ts, and harness.tools tests covering resolution/overrides.

- mastracode
  - HarnessCompat: composes session-derived fields into legacy state, routes model/mode updates to v1 session (including switchMode), preserves selected model across thread switches, and is generic-aware for cloning.
  - Schema/types: split MastraCodeSessionState (currentModelId, modeId) and MastraCodeComposedState = MastraCodeState & MastraCodeSessionState; MastraCodeState no longer contains currentModelId.
  - Agents/tools/instructions now use MastraCodeComposedState and derive state via requestContext.get('harness').getState().
  - createMastraCode: unified typed initialState/workspace construction and removed subagentModelId from observability requestContextKeys.
  - Tests: HarnessCompat.test.ts verifies composition, setState routing, thread switching preservation, and subagent override behavior.

## Backward Compatibility

Existing Mastra Code callers keep the same state shape because HarnessCompat composes session values into returned state. Legacy subagent-model selection behavior is preserved where needed; legacy subagent shapes are not forwarded into the new v1 subagent registry (legacy subagents remain executed by HarnessCompat/v0 until v1 runtime is implemented).

## Public API / Type Changes

- New/changed generics and signatures:
  - Harness<MODES, TState> and HarnessConfig<MODES, TState>
  - Session<TState> and SessionConfig<TState>
  - Added Session methods: getState(), setState(), updateState()
  - HarnessCompat constructor/methods made generic-aware; added getState()/setState() overrides and useSkill passthrough
  - Added MastraCodeSessionState and MastraCodeComposedState types
  - Added StateChangedEvent to harness events
  - New types: PermissionPolicy, ToolCategory, ToolCategoryResolver, SubagentDefinition, SubagentRegistryConfig, ModelResolver, Skill re-exports, and HarnessSkillNotFoundError
  - New utility: buildSessionToolsets and related types

## Tests & Verification

Focused Harness v1 tests, HarnessCompat tests, core typecheck, and core build were run and verified.


wardpeet (commented):
## Review: feat(core): compose Harness v1 session state

### Overall

This is a well-structured PR that adds typed, validated session state and lazy workspace resolution to the Harness v1 `Session` class, and then composes that state into the legacy `HarnessCompat` bridge so MastraCode can derive `currentModelId` / `modeId` from the active session rather than mutable harness state. Clean separation of concerns.

### Core `Session<TState>` — looks solid

- **State initialization**: schema defaults → `initialState` overlay is the right pattern (line 50–53).
- **Serialization queue**: `#stateUpdateQueue` chain for `setState` / `updateState` is correct. The `.then(() => undefined, () => undefined)` prevents unhandled-rejection propagation from earlier failures — good.
- **Schema validation**: `#applyStateUpdates` validates the merged state, not just the delta, which prevents partial-invalidation states. The `state_changed` event only fires after validation passes.
- **`updateState` transaction**: The updater receives a frozen snapshot, returns `{ updates, events, result }`, and events fire after state is committed — correct ordering.
- **`getState()` frozen copy**: Good defensive design.

### `HarnessCompat` state composition — one concern

**`getState()` try/catch on private field access** (line 47–51):

```ts
try {
  session = this.#session;
} catch {
  session = undefined;
}
```

This `try/catch` wrapping a private field read is unusual. Private field access only throws if `this` doesn't have that field (which can't happen for a properly constructed instance). If the intent is to handle cases where `#session` hasn't been set yet, you can just read it directly — it's `undefined` by default. Same pattern repeated in `setState()` (lines 68–72).

**`setState()` dual-write** (lines 83–88): Both `session.setState()` and `super.setState()` receive the same harness updates. This is intentional for backwards compat, but worth calling out that if either fails, the other may have already committed. Not a blocker for this PR since this is a compat bridge.

**`switchThread()` default model override** (lines 108–111): When switching threads, the session's model is reset to the mode's `defaultModelId`. This is a behavior change — previously when switching threads the model persisted. Intentional?

### Events

- `StateChangedEvent` with `changedKeys` is useful for subscribers to do diff-based reactions. Reserved correctly. Good.

### Workspace resolution

- Lazy resolution via function + caching on `#resolvedWorkspace` / `#workspaceResolved` is clean.
- `getWorkspace()` correctly returns `undefined` for dynamic workspaces since they need a request context.

### Tests

- Good unit test coverage for state initialization, validation, transactions, and workspace resolution.
- `HarnessCompat.test.ts` properly mocks the legacy harness and verifies routing of `currentModelId`/`modeId` to session methods.

### CI

The Memory Tests failure is in `working-memory.test.ts` (gpt-4o retry exhaustion, unrelated to this PR). All other checks pass.

### Minor nits

1. **`#getSchemaDefaults()` catch block** (session.ts:232): silently swallows errors from `jsonSchema.output()`. Consider at least a `console.warn` for debugging, or document that this is intentional.
2. **Dead commented code**: `session.ts` lines 31–32 (`parentSessionId`, `subagentDepth`) and `session.types.ts` lines 28–36. If these are planned future fields, a brief `// TODO` comment would help. If not, consider removing.
3. **`harness.types.ts`**: Lots of commented-out sections (queue, files, subagents, goals, permissions). This predates the PR but is worth noting — it makes the file harder to scan.

PR #17413, opened 2026-06-01
fix(duckdb): release file lock on shutdown to fix dev hot-reload lock conflict

DanielSLew:
Running `mastra dev` with a DuckDB store would fail on the second hot reload with `Could not set lock on file ...duckdb: Conflicting lock is held`. DuckDB keeps an exclusive file lock for the whole process lifetime, but nothing released it on shutdown and the server had no signal handlers, so when a hot reload SIGINT'd the old process the lock lingered and the restarted process couldn't reopen the file.

This adds a `close()` to `DuckDBStore` that releases the lock, and wires graceful shutdown into the server so storage backends actually clean up on SIGINT/SIGTERM.

Before, on the second reload:

```
Error: Could not set lock on file "mastra-observability.duckdb": Conflicting lock is held by node (PID 93003)
```

After, hot reloads release the lock cleanly:

```ts
// DuckDBStore now releases the native file lock on shutdown
async close(): Promise<void> {
  await this.db.close();
}

// server installs handlers so mastra.shutdown() runs before exit
process.once('SIGINT', () => void shutdown('SIGINT'));
process.once('SIGTERM', () => void shutdown('SIGTERM'));
```

The shutdown step feature-detects `mastra.shutdown()` so it stays compatible with older `@mastra/core` versions, and is guarded against double-invocation. Verified with repeated hot reloads (zero lock conflicts) plus a new test covering lock release and idempotency.

## ELI5
When developing with DuckDB, restarting the dev server left a lock on the database file so the new process couldn't open it. This PR ensures the database connection is closed on shutdown and the server runs a graceful shutdown so hot-reloads no longer fail with lock conflicts.

## Changes Overview
Fixes a hot-reload failure in `mastra dev` caused by DuckDB’s native file lock persisting across process shutdowns and preventing the restarted process from reopening the DB (error: `Could not set lock on file "...duckdb": Conflicting lock is held`).

## Key Changes

**DuckDB Storage (stores/duckdb/src/storage/index.ts)**
- Added `async close(): Promise<void>` on `DuckDBStore` that calls the underlying `this.db.close()` to release the native DuckDB file lock.
- `close()` is safe to call multiple times (idempotent).

**Server Graceful Shutdown (packages/deployer/src/server/index.ts)**
- `createNodeServer` installs `SIGINT`/`SIGTERM` handlers to run shutdown logic before exiting.
- Shutdown sequence:
  - Uses a `shuttingDown` guard to prevent re-entry and stop accepting new requests.
  - Calls `server.close()` and feature-detects `mastra.shutdown()` (for compatibility with older `@mastra/core`), racing it against a 5s timeout; warns and proceeds if shutdown hangs.
  - Catches and logs shutdown errors, then exits the process.
- Handlers are registered with `process.once` and guarded against double-invocation.

**Tests (stores/duckdb/src/storage/close.test.ts)**
- Added tests verifying:
  - `close()` releases DuckDB’s native file lock so a second `DuckDBStore` can reopen the same database.
  - `close()` is idempotent (calling it twice resolves cleanly).

**Changesets / Docs**
- Added changeset notes documenting the `@mastra/deployer` graceful-shutdown behavior and the `@mastra/duckdb` lock-release fix; adjusted deployer changeset wording to remove DuckDB-specific phrasing.

## Result
Hot-reloads in `mastra dev` now release DuckDB file locks on shutdown and complete cleanly; graceful shutdown is bounded (5s) and robust to older core versions and repeated invocations.

### DIFF
--- a/packages/deployer/src/server/index.ts
+++ b/packages/deployer/src/server/index.ts
@@ -602,6 +602,7 @@ export async function createNodeServer(mastra: Mastra, options: ServerBundleOpti
   // native file lock) before the process exits. On `mastra dev` hot reloads
   // the old process is sent SIGINT; without this the lock can linger and the
   // restarted process fails with "Conflicting lock is held".
+  const SHUTDOWN_TIMEOUT_MS = 5000;
   let shuttingDown = false;
   const shutdown = async (signal: NodeJS.Signals) => {
     if (shuttingDown) {
@@ -614,10 +615,21 @@ export async function createNodeServer(mastra: Mastra, options: ServerBundleOpti
     // Feature-detect for older @mastra/core versions without shutdown().
     const lifecycle = mastra as unknown as { shutdown?: () => Promise<void> };
     if (typeof lifecycle.shutdown === 'function') {
+      // Bound the wait so a hanging shutdown can't block process exit.
+      let timeout: ReturnType<typeof setTimeout> | undefined;
+      const timedOut = Symbol('shutdown-timeout');
+      const timeoutPromise = new Promise<typeof timedOut>(resolve => {
+        timeout = setTimeout(() => resolve(timedOut), SHUTDOWN_TIMEOUT_MS);
+      });
       try {
-        await lifecycle.shutdown();
+        const result = await Promise.race([lifecycle.shutdown(), timeoutPromise]);
+        if (result === timedOut) {
+          logger.warn('Mastra shutdown timed out; forcing exit', { timeoutMs: SHUTDOWN_TIMEOUT_MS });
+        }
       } catch (error) {
         logger.error('Error during Mastra shutdown', { error });
+      } finally {
+        clearTimeout(timeout);
       }
     }
     process.exit(0);
--- a/packages/deployer/src/server/index.ts
+++ b/packages/deployer/src/server/index.ts
@@ -598,5 +598,32 @@ export async function createNodeServer(mastra: Mastra, options: ServerBundleOpti
     await workerLifecycle.startEventEngine();
   }
 
+  // Graceful shutdown so storage backends release resources (e.g. DuckDB's
+  // native file lock) before the process exits. On `mastra dev` hot reloads
+  // the old process is sent SIGINT; without this the lock can linger and the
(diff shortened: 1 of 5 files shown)


packages/deployer/src/server/index.ts
```diff
@@ -598,5 +598,32 @@ export async function createNodeServer(mastra: Mastra, options: ServerBundleOpti
     await workerLifecycle.startEventEngine();
   }
 
+  // Graceful shutdown so storage backends release resources (e.g. DuckDB's
+  // native file lock) before the process exits. On `mastra dev` hot reloads
+  // the old process is sent SIGINT; without this the lock can linger and the
+  // restarted process fails with "Conflicting lock is held".
+  let shuttingDown = false;
+  const shutdown = async (signal: NodeJS.Signals) => {
```
abhiaiyer91, line 607: i have a feeling we already call shutdown somewhere?

DanielSLew: AI says: Good instinct — I went and verified it across the runtime path rather than guessing. Grepping `core/src/mastra`, `hono/src`, `server/src`, and `deployer/src` (excluding tests):

- The only runtime call to `mastra.shutdown()` is the one this PR adds in the server entry (`packages/deployer/src/server/index.ts`). The other match in core is a JSDoc `@example` comment.
- The only `SIGINT`/`SIGTERM` handlers in those packages are the two this PR adds; the one in `core/src/mastra/index.ts` is also just inside a doc example.

What exists nearby but doesn't cover this case:
- `packages/cli/src/commands/dev/dev.ts` only forwards the signal to the child via `serverProcess.kill('SIGINT')` — it never shuts down the in-process Mastra instance inside the bundle.
- `WorkerBundler.getEntry()` installs its own SIGINT/SIGTERM handlers, but that's the worker bundle and only calls `stopWorkers()`, not the API server.

So before this PR nothing called `mastra.shutdown()` when the API server process got SIGINT on a hot reload, which is why DuckDB's file lock lingered. Happy to consolidate the worker + server handlers into a shared helper later if you'd prefer.

Mastra Code

abhiaiyer91 (approved):

PR #17435, opened 2026-06-02
fix(schema-compat): accept ai-sdk Schema type in createTool without cast

### LINKED ISSUE #16384: [BUG] createTool({ inputSchema }) type signature is too narrow — rejects FlexibleSchema (jsonSchema(), StandardSchema) that asSchema() accepts at runtime
## Describe the Bug

`createTool({ inputSchema })` types `inputSchema` as `SchemaWithValidation<T> = ZodLikeSchema<T>` (per `@mastra/core/dist/stream/base/schema.d.ts`), which requires `.parse` and `.safeParse` methods.

But at runtime, mastra calls `asSchema(sdkTool.inputSchema)` from `@ai-sdk/provider-utils` (in `chunk-YC6Z75K2.js` around line 703), and `asSchema` accepts `FlexibleSchema = Schema | LazySchema | ZodSchema | StandardSchema`.

Passing a `Schema` object (e.g. one returned by `jsonSchema()` from `@ai-sdk/provider-utils`) works correctly at runtime — wire schema is correct, tool calls succeed — but TypeScript rejects it. Users have to cast `as never` to bypass the type error, even though the runtime is identical.

This matters as an escape hatch for the OpenAI strict-mode issue (#16383): users who want to bypass `@mastra/schema-compat` and pre-render their JSON Schema directly need to pass a `Schema` object to `createTool`. Today they have to `as never` to do so. Pure DX issue, but it's the difference between "drop-in workaround" and "ugly cast that future readers will question".

**Suggested fix**: widen `SchemaWithValidation<T>` (or directly the `inputSchema?` / `outputSchema?` field types on `Tool` / `createTool`) to `FlexibleSchema<T>` from `@ai-sdk/provider-utils`. Same type that `asSchema()` accepts. Runtime is unchanged.

## Steps To Reproduce

1. Clone the repro and `cd` in:
   ```sh
   git clone https://github.com/bluepnume/mastra-bug-repros
   cd mastra-bug-repros/createtool-type-signature
   ```
2. Install + run:
   ```sh
   npm install
   npm run repro   # runs tsc --noEmit
   ```
3. Observe: code that works correctly at runtime fails type-checking unless cast `as never`. The repro file has both forms side-by-side with `@ts-expect-error` showing the type error.

https://github.com/bluepnume/mastra-bug-repros/tree/main/createtool-type-signature

## Expected Behavior

`inputSchema` (and `outputSchema`) on `Tool` / `createTool` should accept the same `FlexibleSchema<T>` type that `asSchema()` accepts internally. Passing a `Schema` from `jsonSchema()` should type-check without a cast.

## Environment Information

```
System:
  OS: macOS 26.4
  CPU: arm64
Binaries:
  Node: 22.22.1
npmPackages:
  @mastra/core: 1.10.0
  @ai-sdk/provider-utils: 4.0.20
  typescript: 5.9.3
```

SaarthurR:
Fixes #16384

When users pass a `Schema<T>` object from `@ai-sdk/provider-utils` (e.g. via `jsonSchema<T>(...)`) to `createTool({ inputSchema })`, TypeScript was rejecting it or requiring an `as never` cast even though the runtime worked fine.

**Root cause:** `PublicSchema` matched `Schema<T>` only through the versioned `SchemaV4/V5/V6` branches, which require matching unique symbols from specific vendored packages. A user's own `@ai-sdk/provider-utils` has different symbols than the internal vendored copies.

**Fix:** Added `AISdkSchemaLike<Output> = { _type: Output }` as a structural branch in `PublicSchema`. Any schema with a `_type` property (all ai-sdk `Schema<T>` versions regardless of symbol identity) now matches cleanly.

**Test:**
```
pnpm --filter @mastra/core exec vitest run --typecheck src/tools/createtool-input-types.test-d.ts
```

## ELI5
TypeScript is being picky about where a “schema” object comes from, even though the objects behave the same at runtime. This PR teaches `createTool` to accept any schema object that has the right internal “shape” (a `_type` field), so users don’t need annoying `as never` casts.

---

## Problem
Passing a `Schema<T>` from `@ai-sdk/provider-utils` (for example via `jsonSchema<T>(...)`) into `createTool({ inputSchema })` caused TypeScript type errors or forced users to cast, even though it worked at runtime.

This happened because `PublicSchema` only matched ai-sdk schemas through specific versioned branches (`SchemaV4`/`SchemaV5`/`SchemaV6`) that depended on symbol identity from vendored packages—those identities didn’t match the user-installed types.

## Solution
Added a structural fallback type:

- `AISdkSchemaLike<Output> = { _type: Output }`

and included it in `PublicSchema`, so any schema object with a `_type` field is accepted, regardless of symbol identity.

## Changes
- **packages/schema-compat/src/schema.types.ts**: Introduced `AISdkSchemaLike<Output>` and added it to the `PublicSchema` union.
- **packages/core/src/tools/createtool-input-types.test-d.ts**: Added type-level tests confirming `createTool` accepts:
  - `inputSchema` produced by `jsonSchema<...>()` (no `as never`)
  - plain `JSONSchema7`-shaped object literals (no `as never`)
- **.changeset/lemon-boxes-itch.md**: Documented the patch release fix for `@mastra/schema-compat` / `@mastra/core`.

## Testing
Type checking can be verified via:
```sh
pnpm --filter `@mastra/core` exec vitest run --typecheck src/tools/createtool-input-types.test-d.ts
```


wardpeet:
Good catch! Question: would still need 
```
  | SchemaV4<Output>
  | SchemaV5<Output>
  | SchemaV6<Output>
```

cc @coderabbitai wdyt?

abhiaiyer91 (approved):

PR #17439, opened 2026-06-02
fix(core): use stable key-order-independent stringify in CacheKeyGenerator to fix duplicate rs_* on tool-approval resume

### LINKED ISSUE #17432: [BUG] Duplicate OpenAI reasoning item (rs_*) on tool-approval resume — jsonb snapshot vs text message key order defeats message dedup
### Describe the Bug

With `@mastra/core@1.37.1` / `@mastra/memory@1.20.0`, an agent using **Observational Memory** with **PostgreSQL storage** (`@mastra/pg`) can send the **same `rs_*` reasoning item twice** in one OpenAI Responses request after a tool-call approval suspend/resume, failing with:

```
AI_APICallError: Duplicate item found with id rs_...
```

On resume, two copies of the same message land in the `MessageList` and differ **only by JSON object key order**:

- the copy restored from the **workflow snapshot** — `mastra_workflow_snapshot.snapshot` is **`jsonb`**, so Postgres normalizes key order (shortest key first, then bytewise);
- the copy reloaded from messages — `mastra_messages.content` is **`text`**, so it preserves the original insertion order.

`messagesAreEqual` hashes `data-*` parts with order-sensitive `JSON.stringify(data)` (`CacheKeyGenerator.fromDBParts`), so a `data-om-status` part whose keys were reordered by jsonb makes the two copies compare **unequal**. `MessageList.addOne` then mints a second copy with a new id but the **same parts** (same `rs_*`), and history conversion emits that `rs_*` twice.

### Steps To Reproduce

1. Agent with `Memory` + an Observational Memory processor, backed by `@mastra/pg`.
2. A tool whose call requires user approval (suspend/resume).
3. Let OM run for a few turns so an assistant message carries a `data-om-status` part.
4. Send a turn that triggers the approval-gated tool → suspend → approve → resume.
5. Resumed request contains a duplicated `rs_*` → `AI_APICallError: Duplicate item found`.

N/A — this is a framework-internal storage/dedup bug. The mechanism is pinned to specific code paths (see Additional context) and reproduces from the steps above with `@mastra/pg` + an Observational Memory processor + an approval-gated tool and a suspend→resume on a thread that already has `data-om-status` parts.

**Confirmed locally** (`@mastra/core@1.37.1`, `@mastra/memory@1.20.0`, `@mastra/pg`): a single approve→resume on an affected thread crashes with `Duplicate item found with id rs_…`. Instrumenting `messagesAreEqual` showed 10 assistant messages comparing **unequal solely due to key order** (every one diverging at `data-om-status.windows.buffered`, `reflection`↔`observations`); a key-order-independent comparison reports them equal. Of those, the 3 sitting at/before the sealed boundary were minted as duplicates, and one carried the `rs_*` that OpenAI rejected.

### Expected Behavior

Reloading an already-present message on resume should not create a second copy. Two stored representations of the same message that differ only in JSON key order (jsonb workflow snapshot vs text message row) should compare equal, so each `rs_*` reasoning item is sent to OpenAI exactly once and the request succeeds.

### Environment Information

```
@mastra/core: 1.37.1
@mastra/memory: 1.20.0
@mastra/pg (PostgreSQL storage)
LLM Provider: OpenAI (Responses API)
```

Also reproduces on `main`.

### Verification

---

### Additional context

**Code path** (paths on `main`):

`CacheKeyGenerator.fromDBParts` hashes `data-*` parts order-sensitively — `packages/core/src/agent/message-list/cache/CacheKeyGenerator.ts`:

```ts
if (part.type.startsWith('data-')) {
  const data = (part as AIV5Type.DataUIPart<AIV5.UIDataTypes>).data;
  key += JSON.stringify(data); // order-sensitive
}
```

`messagesAreEqual` (MastraDBMessage branch) relies on that key — `packages/core/src/agent/message-list/conversion/utils.ts`:

```ts
return (
  oneMM2.id === twoMM2.id &&
  CacheKeyGenerator.fromDBParts(oneMM2.content.parts) === CacheKeyGenerator.fromDBParts(twoMM2.content.parts)
);
```

When unequal, `MessageList.addOne` mints a duplicate — `packages/core/src/agent/message-list/message-list.ts:1519`:

```ts
} else if (existingIsAtOrBeforeSealedBoundary) {
  messageV2.id = this.generateMessageId?.({ idType: 'message', source: 'memory' }) ?? randomUUID();
  // ...
  this.messages.push(messageV2); // second copy, same parts incl. same rs_*
}
```

The double-load happens because OM reloads DB messages on top of the snapshot-restored list with no id-based skip — `packages/memory/src/processors/observational-memory/observation-turn/load-memory-context.ts:18` (`MessageHistory` does skip existing ids):

```ts
for (const msg of ctx.messages) {
  if (msg.role !== 'system') messageList.add(msg, 'memory');
}
```

**Byte-level evidence** — same message/values, different key order. `data-om-status.data.windows.buffered`:

```jsonc
// text / mastra_messages copy (source builder order: observations before reflection)
"buffered":{"observations":{"chunks":0,"messageTokens":0,"projectedMessageRemoval":0,"observationTokens":0,"status":"idle"},
            "reflection":{"inputObservationTokens":0,"observationTokens":0,"status":"idle"}}

// jsonb / mastra_workflow_snapshot copy (Postgres: shortest key first, then bytewise)
"buffered":{"reflection":{"status":"idle","observationTokens":0,"inputObservationTokens":0},
            "observations":{"chunks":0,"status":"idle","messageTokens":0,"observationTokens":0,"projectedMessageRemoval":0}}
```

The reorder follows jsonb's rule exactly (`reflection`(10) before `observations`(12); `status`(6) < `observationTokens`(17) < `inputObservationTokens`(22); etc.). The substantive parts (`reasoning`/`rs_*`, `tool-invocation`, `text`) are byte-identical — only the `data-om-*` payload key order differs, and that alone flips the dedup key.

**Proposed fix.** Replace the order-sensitive `JSON.stringify(data)` in `fromDBParts` with a key-order-independent stable stringify, so functionally-equal `data-*` payloads always produce the same key:

```ts
if (part.type.startsWith('data-')) {
  const data = (part as AIV5Type.DataUIPart<AIV5.UIDataTypes>).data;
  key += stableStringify(data); // was JSON.stringify(data)
}
```

`safe-stable-stringify` is the best engine (deterministic, circular/BigInt-safe, no intermediate allocation) and is already in the dependency tree via pino.

**Related (different duplicate-item paths):** #15617 (OM async-buffering "Duplicate item found", claimed fixed in 1.26.0 — different trigger), #15454 (streaming assembly, merged), #15897 (agent-loop snapshot re-append, merged), #15774 (OM rerun boundaries, merged), #15622 / #14395 (closed, not merged — symptom-level id dedup), #15476 (open — suspended tool approvals don't round-trip on reload, same lifecycle/different symptom). None address the jsonb-vs-text key-order mechanism here.

Aayush-engineer:
## Description

When an agent using PostgreSQL storage (`@mastra/pg`) resumes after a tool-approval suspend, OpenAI rejects the request with:

**Root cause:** `CacheKeyGenerator.fromDBParts` used `JSON.stringify` to hash `data-*` parts, which is sensitive to object key order. PostgreSQL's `jsonb` column (`mastra_workflow_snapshot`) normalizes key order, while the `text` column (`mastra_messages`) preserves insertion order. This caused functionally-equal messages to compare unequal in `messagesAreEqual`, causing `MessageList.addOne` to mint a duplicate copy with the same `rs_*` reasoning id — which OpenAI rejects.

**Fix:**
- Extract a shared `stableStringify` util (key-order-independent `JSON.stringify`) into `agent/message-list/cache/stable-stringify.ts`
- Use it in `CacheKeyGenerator.fromDBParts` instead of `JSON.stringify` for `data-*` parts
- Remove the duplicate `stableStringify` from `response-cache.ts` and import the shared version

**Before/After:**
```typescript
// Before: jsonb-reordered and text-preserved keys compared unequal
JSON.stringify({ a: 1, b: 2 }) !== JSON.stringify({ b: 2, a: 1 }) // true — bug

// After: both produce the same key regardless of storage backend key order
stableStringify({ a: 1, b: 2 }) === stableStringify({ b: 2, a: 1 }) // true — fixed
```

## Related issue(s)

Fixes #17432

## Type of change

## Checklist

## ELI5

Some messages were stored with the same fields in different orders, which made the app think they were different and create duplicate entries. This PR makes message comparison ignore the order of object keys so identical messages are treated the same and duplicates stop appearing.

## Root Cause

CacheKeyGenerator.fromDBParts used JSON.stringify for `data-*` parts; JSON.stringify is sensitive to key order. PostgreSQL jsonb normalizes object key order while the messages table (text) preserves insertion order, so functionally-identical messages restored from different backends produced different cache keys. That caused messagesAreEqual to return false and MessageList.addOne to create duplicate messages with identical rs_* reasoning IDs, leading OpenAI errors like "Duplicate item found."

## Changes Made

- Added packages/core/src/agent/message-list/cache/stable-stringify.ts: exported stableStringify(value: unknown): string which deterministically orders object keys at every level before stringifying.
- Updated packages/core/src/agent/message-list/cache/CacheKeyGenerator.ts to use stableStringify instead of JSON.stringify for `data-*` parts in fromDBParts.
- Re-exported stableStringify from packages/core/src/agent/message-list/cache/index.ts.
- Removed duplicate stableStringify in packages/core/src/processors/processors/response-cache.ts and imported the shared utility instead.
- Updated .changeset/fix-jsonb-key-order-dedup.md to publish a patch release for `@mastra/core`.
- Added a regression test (CacheKeyGenerator-tool-invocation-guard.test.ts) ensuring reordered object keys produce the same cache key.

## Impact

Cache keys for messages are now key-order-independent, so messages that differ only by object key ordering (e.g., jsonb vs text storage) compare equal and will not produce duplicate messages or duplicate rs_* reasoning IDs during resume flows. This prevents the OpenAI "Duplicate item found" errors in resumed tool-approval / observational memory scenarios.

## Related

Fixes `#17432`

### DIFF
--- a/.changeset/fix-jsonb-key-order-dedup.md
+++ b/.changeset/fix-jsonb-key-order-dedup.md
@@ -2,13 +2,6 @@
 "@mastra/core": patch
 ---
 
-Fix duplicate OpenAI reasoning item (`rs_*`) on tool-approval resume caused by jsonb vs text key-order mismatch.
+Fixes a crash where OpenAI rejects resumed tool-approval requests with `AI_APICallError: Duplicate item found with id rs_*`.
 
-`CacheKeyGenerator.fromDBParts` used `JSON.stringify` to hash `data-*` parts, which is sensitive to object key order. PostgreSQL's `jsonb` column (workflow snapshot) normalizes key order, while the `text` column (messages) preserves insertion order. This caused functionally-equal messages to compare unequal, minting duplicate copies with the same `rs_*` reasoning id — which OpenAI rejects with `AI_APICallError: Duplicate item found`.
-
-The fix replaces `JSON.stringify` with a stable, key-order-independent stringify so jsonb-reordered and text-preserved representations of the same message always produce the same cache key.
-
-```typescript
-// Before: { a: 1, b: 2 } and { b: 2, a: 1 } produced different cache keys
-// After: both produce the same key regardless of key order
-```
+The same message stored in PostgreSQL's `jsonb` column (workflow snapshot) and `text` column (messages) could have different JSON key orders, causing the deduplication check to treat them as different messages and send the same reasoning item twice. Both representations now compare equal regardless of key order.
--- a/.changeset/fix-jsonb-key-order-dedup.md
+++ b/.changeset/fix-jsonb-key-order-dedup.md
@@ -1,0 +1,14 @@
+---
+"@mastra/core": patch
+---
+
+Fix duplicate OpenAI reasoning item (`rs_*`) on tool-approval resume caused by jsonb vs text key-order mismatch.
+
+`CacheKeyGenerator.fromDBParts` used `JSON.stringify` to hash `data-*` parts, which is sensitive to object key order. PostgreSQL's `jsonb` column (workflow snapshot) normalizes key order, while the `text` column (messages) preserves insertion order. This caused functionally-equal messages to compare unequal, minting duplicate copies with the same `rs_*` reasoning id — which OpenAI rejects with `AI_APICallError: Duplicate item found`.
+
+The fix replaces `JSON.stringify` with a stable, key-order-independent stringify so jsonb-reordered and text-preserved representations of the same message always produce the same cache key.
+
+```typescript
+// Before: { a: 1, b: 2 } and { b: 2, a: 1 } produced different cache keys
+// After: both produce the same key regardless of key order
+```
--- a/packages/core/src/agent/message-list/cache/CacheKeyGenerator-tool-invocation-guard.test.ts
+++ b/packages/core/src/agent/message-list/cache/CacheKeyGenerator-tool-invocation-guard.test.ts
@@ -52,6 +52,60 @@ describe('CacheKeyGenerator tool-invocation null guard (#16756)', () => {
     expect(() => CacheKeyGenerator.fromDBParts(parts as any)).not.toThrow();
   });
 
+  it('fromDBParts should generate the same key for data parts with reordered object keys', () => {
+    const messagePartsFromTextStorage = [
+      {
+        type: 'data-om-status' as const,
+        data: {
+          windows: {
+            buffered: {
+              observations: {
+                chunks: 0,
+                messageTokens: 0,
+                projectedMessageRemoval: 0,
+                observationTokens: 0,
+                status: 'idle',
+              },
+              reflection: {
+                inputObservationTokens: 0,
+                observationTokens: 0,
+                status: 'idle',
+              },
+            },
+          },
+        },
+      },
+    ];
+
+    const messagePartsFromJsonbSnapshot = [
+      {
+        type: 'data-om-status' as const,
+        data: {
+          windows: {
+            buffered: {
+              reflection: {
+                status: 'idle',
+                observationTokens: 0,
+                inputObservationTokens: 0,
+              },
+              observations: {
+                chunks: 0,
+                status: 'idle',
+                messageTokens: 0,
+                observationTokens: 0,
+                projectedMessageRemoval: 0,
+              },
+            },
+          },
+        },
+      },
+    ];
+
+    expect(CacheKeyGenerator.fromDBParts(messagePartsFromTextStorage as any)).toBe(
+      CacheKeyGenerator.fromDBParts(messagePartsFromJsonbSnapshot as any),
+    );
+  });
+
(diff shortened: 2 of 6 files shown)


TylerBarnes (approved):

TylerBarnes:
Thanks for this fix! really good find, appreciate it a lot 🙌

PR #17474, opened 2026-06-02
fix(core): preserve failed tool results through agent loop

### LINKED ISSUE #15569: Tool execution errors stored as successful results — lost on history reload

intojhanurag:
## Description

Fixes the failed-tool agent loop regression that reproduced in MastraCode when a tool returns an error result. This PR makes failed tool results flow through the loop as real error outputs, preserves them in prompt/history state, and prevents downstream memory/stream consumers from dropping or choking on that failed-tool context.

Root cause:

The reintroduced tool-error persistence path stored failed tool calls as `output-error`, but Observational Memory token counting did not handle that state. When MastraCode hit a failed tool call, the failed result could display in the UI, but the follow-up model step either froze or lost the error context because downstream consumers were not consistently treating it as an error result.

What changed:

- Treat structured failed tool results with `isError: true` as failed outputs in the agentic loop.
- Persist failed tool results as `output-error` with the tool error text.
- Preserve legacy `toolInvocations` with `output-error` when converting history back into AI SDK v5 prompt/UI parts.
- Count `output-error` tool invocations in Observational Memory token counting instead of throwing.
- Avoid treating continued step-level `finish` chunks as terminal in harness/subscribed stream consumers.
- Preserve streamed tool errors in harness message state.
- Mark failed structured tool-result rows as `error` in channel timeline rendering instead of `complete`.

Manual MastraCode validation:

```text
Use the find_files tool with path "/Users/anuragojha" and maxDepth 1. Do not use execute_command. After the tool fails, respond with one sentence summarizing the tool error, then stop.
```

Before rebuilding with the fix, the tool error displayed but the next prompt lost context and responded as if there was no prior tool result. After applying the fix and rebuilding `@mastra/core` and `@mastra/memory`, MastraCode correctly summarized the failed `find_files` permission error, preserved the failed result for follow-up questions, and completed the run instead of freezing.

## Related issue(s)

Fixes #15569.

## Type of change

- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [x] Test update

## Checklist

## ELI5

When a tool fails while an agent is working on a task, this PR makes sure the error is properly saved and shown throughout the system instead of being lost or causing the agent to get stuck.

## Overview

This PR fixes a regression where tool execution errors could be lost or cause the agent loop to hang. The fix ensures that failed tool results are treated as error outputs (marked with `isError: true` and `errorText`) throughout the stack—from live streaming through persistent history to UI rendering.

### Core Message & Invocation Handling
- **`buildMessagesFromChunks.ts`**: Now converts `tool-result` chunks with `isError: true` into `tool-invocation` parts with `state: 'output-error'` and `errorText`, using `safeStringify` for non-string error payloads.
- **`convert-to-mastra-v1.ts`**: Added `hasTerminalToolOutput` and `toToolResultPart` helpers to include error tool invocations (`output-error` state) when building tool-result message content, rather than skipping them.
- **`AIV4Adapter.ts`** and **`AIV5Adapter.ts`**: Updated to preserve `output-error` invocations in UI message conversions and map `errorText` to displayable `result` fields for backward compatibility.

### Agent Loop & Execution
- **`llm-mapping-step.ts`**: Enhanced to detect error conditions via both `toolCall.error` and an `isError` flag in `toolCall.result`. Changed to emit `tool-result` chunks with `isError: true` instead of `tool-error` chunks for consistent error handling.
- **`llm-execution-step.ts`**: Extended to update `messageList` for both successful and error tool results, persisting error state as `output-error` with `errorText`.

### Stream Processing
- **`thread-stream-runtime.ts`** and **`harness.ts`**: Added `isTerminalStreamPart` / `isTerminalFinishChunk` helpers to avoid treating continued-step `finish` chunks as terminal, allowing the agent loop to continue after handling tool errors.
- **`harness.ts`**: Enhanced `tool-error` stream handling to record errors as `tool_result` content in the assistant message (instead of only emitting events) and emit corresponding `message_update`.

### UI & Timeline Rendering
- **`chat-driver-streaming.ts`**: Changed task updates for failing tools to set status to `'error'` when `isError` is true, instead of always using `'complete'`.

### Memory & Token Counting
- **`token-counter.ts`**: Extended token counting to include `tool-invocation` parts with `state: 'output-error'`, treating error text similarly to regular tool results.

### Utilities
- **`utils.ts`**: Updated `safeStringify` to always return a string by falling back to `String(value)` when `JSON.stringify` returns `undefined`.

## Test Coverage

Added comprehensive tests across multiple files:
- **Agent loop error recovery** (`test-utils/options.ts`): Tests verify tool failure termination, error details in next-step prompts, recovery with structured error results, message list persistence of error state, and mixed success/failure handling.
- **Stream conversion** (`build-messages-from-chunks.test.ts`): Tests verify `isError: true` tool results are converted to `output-error` invocations with appropriate `errorText`.
- **Provider errors** (`llm-execution-step.test.ts`): Tests verify persisted error state for provider-executed tool failures.
- **LLM mapping** (`llm-mapping-step.test.ts`): Tests verify error tool results are persisted and continue the loop appropriately.
- **Signal messages** (`signal-messages.test.ts`): Tests verify continued finish handling and structured tool error flows through the subscribed thread stream.
- **Channel consumption** (`consume-agent-stream.test.ts`): Tests verify timeline task updates transition to `error` status for failed tools.
- **Message list conversion** (`message-list-v5.test.ts`): Tests verify legacy `output-error` invocations are preserved through V5 conversion.
- **Token counting** (`token-counter.test.ts`): Tests verify error tool invocations are counted.

## Related Issues

Fixes `#15569`. Related to regressions/reverts `#16976`, `#17403`, `#17422`.

## Changeset

Marks `@mastra/core` and `@mastra/memory` for patch version updates.

### DIFF
--- a/.changeset/bumpy-doors-crash.md
+++ b/.changeset/bumpy-doors-crash.md
@@ -3,4 +3,4 @@
 '@mastra/memory': patch
 ---
 
-Fixed failed tool calls being treated as successful results in agent history, live harness streams, and observational memory token counting. Tool errors are now preserved as error outputs when reloading memory, and continued finish chunks no longer stop the harness before the agent can respond to a failed tool call. Fixes #15569.
+Fixed failed tool calls being treated as successful results in agent history, live harness streams, and observational memory. Tool errors are now preserved as error outputs end-to-end: they survive memory reload, the observational-memory observer and recall transcripts include the error so the agent keeps the failure context, the token limiter and tool-call filter account for failed results, and tool-accuracy evals record a failed tool as an unsuccessful step. Continued finish chunks no longer stop the harness before the agent can respond to a failed tool call. Fixes #15569.
--- a/.changeset/bumpy-doors-crash.md
+++ b/.changeset/bumpy-doors-crash.md
@@ -2,4 +2,4 @@
 '@mastra/core': patch
 ---
 
-Fixed tool execution errors being stored as successful results, which caused failed tool calls to reappear as successes when a conversation was reloaded from history. When a tool throws (or a provider-executed tool reports a failure), the result is now persisted as `state: "output-error"` with `errorText` instead of `state: "result"`, so the error survives history reload via `toAISdkV5Messages()` and matches the live streaming behavior. Fixes #15569.
+Fixed failed tool calls showing up as successful when a conversation is reloaded from history. Previously a tool that threw an error displayed correctly while the agent was running, but reappeared as a successful result once the conversation was loaded from memory. The error is now preserved and displayed on reload, matching what you see during the live run. Fixes #15569.
--- a/.changeset/bumpy-doors-crash.md
+++ b/.changeset/bumpy-doors-crash.md
@@ -1,5 +1,6 @@
 ---
 '@mastra/core': patch
+'@mastra/memory': patch
 ---
 
-Fixed failed tool calls showing up as successful when a conversation is reloaded from history. Previously a tool that threw an error displayed correctly while the agent was running, but reappeared as a successful result once the conversation was loaded from memory. The error is now preserved and displayed on reload, matching what you see during the live run. Fixes #15569.
+Fixed failed tool calls being treated as successful results in agent history, live harness streams, and observational memory token counting. Tool errors are now preserved as error outputs when reloading memory, and continued finish chunks no longer stop the harness before the agent can respond to a failed tool call. Fixes #15569.
--- a/.changeset/bumpy-doors-crash.md
+++ b/.changeset/bumpy-doors-crash.md
@@ -1,0 +1,5 @@
+---
+'@mastra/core': patch
+---
+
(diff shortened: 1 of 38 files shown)


intojhanurag:
## During Hang
image

## After fix 
image

Issue #17216, opened 2026-05-28
[Auth/FGA] Trusted system-actor signal for cron / background workflows (membership-only bypass)


tiffanybalc:
## Summary

`requireFGA` has no way to recognize a trusted, server-constructed system
actor (cron jobs, scheduled workflows, background snapshots). Today these
hit `WorkOSFGAMembershipResolutionError` (or a generic `FGADeniedError`)
because there's no human membership to authorize against. Downstream apps
are reinventing the same subclass workaround.

This issue proposes adding a **trusted-call signal** to the FGA call path
(not to provider user shapes) that lets a server-side caller assert
"this invocation is a system actor — skip the per-membership FGA check."

**Strict bound on the proposal**: this is a *membership-resolution* bypass
only. It is NOT a tenant-scope bypass and MUST NOT allow cross-org access.
Acceptance criteria below.

## Use case

Mastra workflows are increasingly invoked from background jobs (Inngest crons,
queued tasks, snapshot pipelines). In our product these run nightly per-org
across hundreds of organizations, fan out per-team and per-user, and execute
agents/workflows with no human actor in the loop. Tenant scope flows
(`organizationId` on `RequestContext` + Prisma RLS downstream), but there is
no membership to authorize against — because there is no member.

This isn't WorkOS-specific. The 9 provider packages (`auth/auth0`,
`auth/better-auth`, `auth/clerk`, `auth/cloud`, `auth/firebase`, `auth/okta`,
`auth/studio`, `auth/supabase`, `auth/workos`) all share the same
user-centric `IFGAProvider` shape and will hit this when downstream apps add
background workflows.

## Where the gap is

`packages/core/src/auth/ee/fga-check.ts:84-118` (`requireFGA`) — the single
chokepoint all 5 call sites in core funnel through (`Agent.generate`/`stream`,
`Workflow.execute`, agentic tool-call step, agent-bound tools, memory/thread):

```ts
if (!fgaProvider) return;
if (!user) throw new FGADeniedError(user, resource, permission, 'authenticated user is required');
await fgaProvider.require(user, ...);
```

No way for a *trusted caller* to opt out per call.
`MastraFGAWorkos.require` at `auth/workos/src/fga-provider.ts:173-200`
hardcodes `{ strictMembershipResolution: true }` (line 184), so any present
user without a resolvable membership throws — even when the call is
legitimately a system actor.

The existing `trustJwtClaims` + `jwtClaims.organizationMembershipId` path
[...]

PR #17483, opened 2026-06-02
feat(core): add system actor FGA signal

graysonhicks:
## Stack

This is **1 of 3** for #17216.

- **1 of 3:** #17483 - core FGA signal (this PR)
- **2 of 3:** #17484 - direct FGA call-site propagation
- **3 of 3:** #17487 - agent propagation and WorkOS guidance

## What changed

This adds the core FGA primitive for trusted server-side system actors. `requireFGA` and `checkFGA` now accept an explicit `systemActor` option, require an `organizationId` in the request context before bypassing provider membership resolution, and emit FGA telemetry with `actor_kind: "system"` plus null user and membership IDs.

This PR intentionally stays limited to the core primitive. The follow-up PRs in this stack plumb it through direct invocation boundaries and agent/background paths.

## Validation

- `pnpm --filter ./packages/core exec vitest run src/auth/ee/__tests__/fga-check.test.ts src/auth/ee/__tests__/ee-telemetry.test.ts`
- `git diff --check`
- `pnpm build:core`
- `coderabbit review --plain --type uncommitted --base origin/main --dir packages/core/src/auth/ee`

## ELI5

This PR lets trusted server-side processes (system actors) perform authorization checks without going through normal user membership lookup, but only when an organization ID is present so it can't be misused; it also logs these special accesses for auditing.

## Overview

First of three PRs implementing core Fine-Grained Authorization (FGA) support for system actors. Adds a minimal, explicit signal for trusted server-side checks so core FGA primitives can bypass provider membership resolution safely and emit appropriate telemetry. Follow-ups will propagate this signal across call boundaries and agents/background jobs.

## Key Changes

- New exported type: SystemActorSignal
  - export type SystemActorSignal = true | { actorKind: 'system'; sourceWorkflow?: string }

- FGA function updates
  - checkFGA and requireFGA accept an optional systemActor option.
  - When a valid systemActor is provided:
    - requireFGA/checkFGA derive the requestContext (must include organizationId) and will throw if organizationId/tenant scope is missing.
    - Provider membership resolution is bypassed (no provider.require call).
    - Enterprise telemetry is emitted with actor_kind: "system", and user_id and organization_membership_id set to null.
    - sourceWorkflow is taken from systemActor.sourceWorkflow if present, otherwise from requestContext.metadata.sourceWorkflow.

- Non-system behavior
  - Telemetry for non-system checks is tagged actor_kind: "user" and normalizes absent user or membership IDs to null.
  - Malformed or user-controlled systemActor values do not trigger the trusted bypass and follow the normal provider path.

- Type/API adjustments
  - requestContext was moved from RequireFGAOptions into CheckFGAOptions.
  - CheckFGAOptions gains systemActor?: SystemActorSignal.
  - RequireFGAOptions now inherits requestContext via CheckFGAOptions.
  - index.ts re-exports SystemActorSignal.

## Tests & Validation

- Added and updated tests in packages/core/src/auth/ee/__tests__:
  - fga-check.test.ts: covers tenant-scoped workflow system actor bypass, missing organizationId rejection, malformed/system-like-but-not-trusted values, and ensuring user-supplied values are not trusted signals.
  - ee-telemetry.test.ts: validates emitted ee_feature_used telemetry for system actors (actor_kind: "system", null user/membership IDs, correct source_workflow resolution).
- Changeset added and standard validation steps included (build, vitest subset, git diff/check, coderabbit review).

## Scope & Next Steps

Scope limited to the core FGA primitive. Follow-up PRs will:
- Propagate the system actor signal at direct FGA call sites (`#17484`)
- Support agent/background propagation and WorkOS guidance (`#17487`)


tiffanybalc (approved):
good set up for https://github.com/mastra-ai/mastra/pull/17484/changes

abhiaiyer91 (approved):

rphansen91 (approved):
LGTM. Design is well-scoped, fail-closed semantics preserved, and the spoof-resistance is solid — `systemActor` is a function argument only, never read from `user` or `requestContext`, and the `organizationId` requirement keeps the bypass tenant-scoped. Tests cover the real attack surface.

One thing to keep in mind for the follow-ups (#17484, #17487): the entire trust boundary lives at the call sites that set `systemActor`. The signal needs to come from run-origin / internal runtime state — never copied from a request body, user object, request context value, or any caller-supplied options bag that crosses a trust boundary. Worth being explicit about that as those PRs land.

Nice setup for the rest of the stack.

PR #17489, opened 2026-06-02
feat: serve agent-builder model availability and auth permission patterns over the API

mfrachet:
## Summary

Studio's browser bundle was importing server-only Enterprise (EE) helpers directly — `@mastra/core/agent-builder/ee/allowlist` and `@mastra/core/auth/ee/permissions`. Those imports dragged Node-only modules (`fs`, `os`, `path`, `module`, `stream`) into the client, so the build needed browser-safe core subpaths, a static provider-registry reader, and Vite stubs just to keep Studio from crashing at module-eval.

This PR removes that leak at the source by moving the two browser-facing EE concerns behind API endpoints, so EE logic stays server-only and the browser consumes plain JSON.

### New server endpoints (`@mastra/server`)
- **`GET /editor/builder/models/available`** — returns the provider/model list already filtered by the active builder model policy. Auth: `requiresAuth: true`, `stored-agents:read`.
- **`GET /auth/permission-patterns`** — returns the authoritative permission-pattern vocabulary. Auth: `requiresAuth: true` only — the response is the non-sensitive route-permission vocabulary every authenticated user needs to gate their own sidebar/redirects, and no narrower permission fits.

Both endpoints lazily/dynamically import the EE code on the server, per existing convention.

### client-js (`@mastra/client-js`)
- Adds `getBuilderAvailableModels()` and `getPermissionPatterns()` plus response types, so type safety holds at the client boundary.

### Playground
- The model picker and staleness checks now render the server-filtered available-models list and derive "model not allowed" via **set-membership** instead of running the allowlist matcher in the browser.
- `MODEL_NOT_ALLOWED_CODE` is a Playground-local literal (`'MODEL_NOT_ALLOWED'`, the wire string the server already emits on 422).
- `route-permissions` becomes hook-driven (`useRoutePermissions` + `usePermissionPatterns`): the hardcoded route→permission literals are validated at runtime against the fetched pattern set, preserving the old compile-time typo protection.
- The route-permission guard is now **defensive while patterns load**: it shows a spinner instead of leaking protected content before gating can run.

### Core revert
With no browser code importing `@mastra/core/*/ee/*` anymore, the earlier browser-safe core splits are reverted: the `./auth/ee/permissions` and `./agent-builder/ee/allowlist` exports, the static provider-registry reader, and the tsup/`package.json` plumbing are removed. The server still uses the full Node-capable registry and EE barrels via dynamic import. These core additions were branch-local and never released, so there's no breaking-change note.

## Test plan
- **Core**: `pnpm --filter ./packages/core check` — `tsc` + 128 tests pass; confirmed `dist/auth/ee/permissions.*` and `dist/llm/model/provider-registry.static.*` are gone and no new export paths remain.
- **Server**: handler tests for both endpoints (policy-inactive ⇒ full list; active allowlist ⇒ filtered; permission-patterns returns the full list); `build:server` + `tsc` clean.
- **client-js**: builds; both endpoint paths present in the bundle.
- **Playground**: full suite passes (131 files); `tsc -p tsconfig.build.json` clean. New MSW-driven test covers the guard's loading-spinner behavior and the gated render path.
- Repo-wide grep confirms **zero** source imports of `@mastra/core/agent-builder/ee/allowlist` or `@mastra/core/auth/ee/permissions`.

## ELI5 (Explain Like I'm 5)
The browser was accidentally importing server-only code, which made the client heavier and risky. This PR moves those server-only checks behind two server endpoints so the browser simply fetches JSON (which models are allowed and what permission strings exist).

---

## Overview
Moves Enterprise/server-only agent-builder allowlist and auth permission-pattern logic out of client bundles and behind authenticated server APIs. Client and Playground now fetch authoritative JSON from the server; server handlers lazily import EE code so EE modules remain server-only.

---

## Key Changes

Server
- New endpoints:
  - GET /editor/builder/models/available — returns providers/models filtered by the active builder model policy. (requiresAuth: true, stored-agents:read)
  - GET /auth/permission-patterns — returns authoritative permission-pattern vocabulary. (requiresAuth: true)
- Handlers dynamically import EE/server-capable code at runtime.
- Added response schemas, route metadata, and server tests.
- Added buildProvidersList helper and ProviderListItem typing.

Client SDK (client-js)
- MastraClient additions:
  - getBuilderAvailableModels(): Promise<BuilderAvailableModelsResponse>
  - getPermissionPatterns(): Promise<PermissionPatternsResponse>
- New types: BuilderAvailableModelsResponse, PermissionPatternsResponse, PermissionPattern.
- Route contract/type generation updated for both endpoints.

Playground / Browser
- useAgentBuilderAllowedModels now queries getBuilderAvailableModels() (React Query) and exposes providers + flattened models.
- Model pickers, staleness checks, and filtering use server-provided allowed models (set-membership on normalized provider:model keys).
- MODEL_NOT_ALLOWED_CODE defined locally in Playground to avoid EE imports.
- New usePermissionPatterns hook fetches permission patterns; RoutePermissionsGate fetches and validates route literals at app start (shows spinner while loading).
- RoutePermissionGuard shows spinner while permissions load and guards against redirect loops.
- AppSidebar and related UI hide permission-gated items while patterns/capabilities are loading.
- Pages that create/edit agents prefetch/seed the builder-available-models cache for users with write access.

Core (`@mastra/core`)
- Removed browser-facing EE exports (`./auth/ee/permissions`, `./agent-builder/ee/allowlist`) and related packaging plumbing; reverted prior browser-safe split.
- Server continues to dynamically import EE code when needed.

---

## Tests & Validation
- Server handler tests added for both new endpoints; server build/typecheck clean.
- Playground test suites updated; new MSW tests validate permission-pattern loading and guard behavior; many tests adjusted to mock getBuilderAvailableModels.
- Client-js builds and includes new endpoint paths and types.
- Changeset updated for minor releases for `@mastra/server` and `@mastra/client-js`.
- Repo grep confirmed no remaining source imports of the two EE entrypoints.

---

## Why this matters
- Prevents Node/EE modules from leaking into the browser bundle.
- Centralizes policy and permission pattern truth on the server for consistent client behavior.
- Simplifies client code and test setup by replacing compile-time EE imports with runtime HTTP-backed data.


damien-schneider (dismissed):

packages/core/src/llm/model/provider-registry.static.ts
```diff
@@ -0,0 +1,57 @@
+/**
```
wardpeet: is it possible to create an API endpoint instead? Maybe under agentbuilder that returns this list?

packages/playground/src/domains/auth/components/route-permissions-gate.tsx
```diff
@@ -0,0 +1,57 @@
...
+      </div>
+    );
+  }
+
+  // Only validate when RBAC is actually in effect (non-empty pattern set).
+  // RBAC-off resolves with an empty set and gating is a no-op.
+  if (patterns.size > 0) {
+    for (const literal of collectRouteLiterals(ROUTE_PERMISSIONS)) {
+      if (!patterns.has(literal)) {
+        throw new Error(
```
wardpeet, line 59: Do we want to crash the app?

packages/playground/src/domains/auth/components/route-permissions-gate.tsx
```diff
@@ -0,0 +1,57 @@
...
+      <div className="flex h-screen w-full items-center justify-center">
+        <Spinner />
+      </div>
+    );
+  }
+
+  // Only validate when RBAC is actually in effect (non-empty pattern set).
+  // RBAC-off resolves with an empty set and gating is a no-op.
+  if (patterns.size > 0) {
+    for (const literal of collectRouteLiterals(ROUTE_PERMISSIONS)) {
```
wardpeet: isn't this the same?
```suggestion
    for (const literal of ALL_SIDEBAR_PERMISSIONS) {
```

packages/playground/src/domains/auth/components/route-permissions-gate.tsx
mfrachet: That was the initial behaviour (see the P function), it will lift up to the react error boundary

YujohnNattrass (approved):

packages/server/src/server/handlers/auth.ts
```diff
@@ -16,6 +16,7 @@ import type {
   SSOCallbackResult,
 } from '@mastra/core/auth';
 import type { IRBACProvider, IFGAProvider, EEUser } from '@mastra/core/auth/ee';
+import { PERMISSION_PATTERNS } from '@mastra/core/auth/ee';
```
YujohnNattrass: lets make this a dynamic import, looks like the convention in this file is set that way. Most likely to prevent import errors when bundling and mismatch deps.

PR #17505, opened 2026-06-03
fix(playground): preserve completed OM badge state

damien-schneider:
## Summary

- Keep completed/failed/activated Observational Memory cycles from being marked disconnected when a later stream is canceled or errors.
- Make synthetic buffering-end injection idempotent so repeated buffer-status handling does not add duplicate OM completion markers.
- Restore OM sidebar progress when persisted initial messages arrive after the runtime provider mounts.

## Verification

- pnpm --filter ./packages/playground exec vitest run src/services/__tests__/om-parts-converter.test.ts src/services/__tests__/mastra-runtime-provider.test.tsx --bail 1 --reporter=dot
- pnpm --filter ./packages/playground typecheck
- git diff --check

### DIFF
--- a/packages/playground/src/services/mastra-runtime-provider.tsx
+++ b/packages/playground/src/services/mastra-runtime-provider.tsx
@@ -207,19 +207,22 @@ export function MastraRuntimeProvider({
   };
 
   // Helper to update progress from streamed data-om-status parts
-  const handleProgressUpdate = (data: any) => {
-    // Ignore progress from a different thread (e.g., if user switched threads mid-stream)
-    if (data.threadId && data.threadId !== threadId) {
-      return;
-    }
-    setStreamProgress({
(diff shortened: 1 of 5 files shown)


packages/playground/src/services/mastra-runtime-provider.tsx
```diff
@@ -272,8 +275,7 @@ export function MastraRuntimeProvider({
     if (lastProgress) {
       handleProgressUpdate(lastProgress);
     }
-    // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, []); // Only run once on mount
+  }, [handleProgressUpdate, initialMessages, markCycleIdActivated]);
```
mfrachet, line 278: We should use useEffectEvent instead of those and the callback above I think

PR #17513, opened 2026-06-03
fix(mcp): scope MCP server FGA checks independently of agent tool map

### LINKED ISSUE #17508: MCPServer: scope tool-authorization (FGA) independently of the instance-level agent `tool` mapping
### Summary

`MCPServer` tool-authorization checks (`tools/list` / `tools/call`) and a Mastra instance's **internal agent/workflow tool execution** both resolve the `tool` resource through the *same* instance-level `resourceMapping.tool`. There is no way to scope the MCP tool gate independently from the agent tool gate.

This is a follow-up to #17475 (thanks for `mapAuthInfoToUser` — it solves the "no `user` in the MCP request context" half of the problem).

### Background

When an `MCPServer` is registered on a Mastra instance that carries an FGA provider (`server.fga`), `@mastra/mcp` enforces `TOOLS_EXECUTE` on every `tools/list` / `tools/call` by reading `requestContext.get('user')` and checking the `tool` resource. The resource type/id are resolved by the FGA provider's `resourceMapping.tool` (e.g. `MastraFGAWorkos` → `{ fgaResourceType, deriveId }`).

That exact same `resourceMapping.tool` is also applied to **internal tool execution inside agent/workflow runs** (core checks `{ type: 'tool' }` / `TOOLS_EXECUTE` during a run). So a single mapping is forced to serve two surfaces with very different authorization shapes.

### Why this is a problem

These two surfaces frequently need different scoping:

- **Agent/workflow tool execution** is often **team/tenant-scoped** — e.g. `tool` → `team`, `deriveId: user => user.teamId`.
- **MCP callers** are commonly **org + user scoped and span many teams**. A `tools/list` response isn't team-specific, and there's no single "current team" for an MCP session.

Because both go through one `resourceMapping.tool`, you're forced into a bad trade:

1. Loosen the global `tool` mapping (e.g. → `user`/`org`) so MCP passes — which **also relaxes the agent/workflow tool gate**; or
2. Synthesize a `user` in `mapAuthInfoToUser` that satisfies the agent-shaped mapping (e.g. pick an arbitrary team the caller belongs to) — which works but is semantically muddy and couples the MCP gate to an unrelated mapping.

`deriveId` can change the resolved *id* per request, but not the *resource type*, so the type mismatch can't be resolved from the mapping alone.

### Request

Let an `MCPServer` scope its tool authorization independently of the instance-level agent `tool` mapping. Any one of these would solve it:

1. **Per-server FGA resource mapping / permission override** on `MCPServer` (e.g. `fga: { resourceMapping, permissionMapping }`), consulted for that server's `tools/list` / `tools/call` instead of the instance default; or
2. **A distinct resource type for MCP tool checks** (e.g. `mcpTool`) so it can be mapped separately from agent `tool`; or
3. A documented hook to fully resolve the FGA `{ resource, permission }` for an MCP tool check (the MCP analogue of route-level `fga` metadata).

`mapAuthInfoToUser` is the natural companion to this — together they'd let an OAuth-protected MCP server be authorized coherently without disturbing the agent/workflow tool gate.

### Illustrative example

```ts
const server = new MCPServer({
  name: 'My Server',
  version: '1.0.0',
  tools: { getData },
  mapAuthInfoToUser: ({ authInfo }) => mapToUser(authInfo),
  // proposed: scope this server's tool checks independently of the
  // instance-level agent `tool` mapping
  fga: {
    resourceMapping: { tool: { fgaResourceType: 'user', deriveId: ({ user }) => user.id } },
    permissionMapping: { 'tools:execute': 'read' },
  },
});
```

### Environment

- `@mastra/mcp` `1.9.0`, `@mastra/core` `1.38.0`

nehaprasad-dev:
#### Description

- MCP server tools/list and tools/call FGA checks now use the mcpTool resource type by default, separate from instance-level tool mapping used by agent/workflow execution. MCPServer accepts optional fga.resourceMapping and fga.permissionMapping overrides, merged via withConfigOverrides on the FGA provider.

#### Related issue(s)
Fixes #17508

#### Type of change
- Bug fix (non-breaking change that fixes an issue)
- Documentation update

#### Checklist
-  I have made corresponding changes to the documentation (if applicable)
-  I have addressed all Coderabbit comments on this PR

## ELI5

Currently, when an MCP server (a type of tool extension) checks if someone is allowed to use its tools, it uses the same authorization rules as regular Mastra tools, making it impossible to restrict MCP tool access separately. This PR lets each MCP server have its own authorization rules, so you can control who can use MCP tools independently from who can use regular tools.

## Summary

This PR fixes MCP server tool authorization (FGA checks) to use a dedicated `mcpTool` resource type by default, rather than sharing the instance-level `tool` resource mapping used by agent/workflow execution. It introduces optional FGA configuration overrides (`resourceMapping` and `permissionMapping`) on `MCPServer` to allow per-server customization of authorization checks.

### Key Changes

**FGA Provider Enhancement**
- Added `withConfigOverrides(overrides: FGAProviderConfigOverrides)` method to `MastraFGAWorkos`, enabling callers to create a new provider instance with merged resource and permission mappings
- Extended `getResourceMapping()` to recognize and handle the `mcpTool` resource type alongside existing aliases (`mcpTools`)

**MCP Server FGA Configuration**
- Introduced `MCPServerFGAConfig` interface in `packages/core/src/mcp/types.ts` with optional properties: `resourceType`, `resourceMapping`, and `permissionMapping`
- Extended `MCPServerConfig` to accept optional `fga?: MCPServerFGAConfig`
- Updated `MCPServer` to store and utilize per-server FGA configuration when enforcing tool execution authorization

**MCP-Specific Authorization Utilities**
- Added `MCP_SERVER_TOOL_FGA_RESOURCE_TYPE` constant as the default resource type for MCP tools
- Added `getMCPServerToolFGAResourceType()` to derive resource type from optional config
- Added `resolveMCPServerFGAProvider()` to conditionally wrap the FGA provider with config overrides when server-level mappings are provided

**Updated Authorization Logic**
- Modified tool execution authorization (`enforceToolExecutionFGA`) to dynamically derive the FGA resource type using MCP-specific helpers instead of hardcoding `"tool"`
- Error messages now include the computed resource type and tool resource ID for better debugging

**Documentation**
- Added enforcement point documentation for MCP server `tools/list` and `tools/call` with `mcpTool` resource type

### Non-Breaking
This change is backward compatible. Existing MCP servers without explicit FGA configuration will automatically use the new `mcpTool` resource type for authorization, while allowing new per-server overrides via `MCPServerConfig.fga` for fine-grained control.

### DIFF
--- a/auth/workos/src/fga-provider.ts
+++ b/auth/workos/src/fga-provider.ts
@@ -23,6 +23,7 @@ import type {
 import { FGADeniedError } from '@mastra/core/auth/ee';
 import { WorkOS } from '@workos-inc/node';
 
+import type { FGAProviderConfigOverrides } from '@mastra/core/auth/ee';
 import type { MastraFGAWorkosOptions, FGAResourceMappingEntry, WorkOSUser } from './types';
 
 const FILTER_ACCESSIBLE_CHECK_CONCURRENCY = 5;
@@ -112,6 +113,8 @@ export class MastraFGAWorkos implements IFGAManager<WorkOSUser> {
   readonly resolveRouteFGA?: MastraFGAWorkosOptions['resolveRouteFGA'];
   readonly validatePermissions?: MastraFGAWorkosOptions['validatePermissions'];
 
+  private readonly options: MastraFGAWorkosOptions;
+
   constructor(options: MastraFGAWorkosOptions) {
     const apiKey = options.apiKey ?? process.env.WORKOS_API_KEY;
     const clientId = options.clientId ?? process.env.WORKOS_CLIENT_ID;
@@ -123,6 +126,7 @@ export class MastraFGAWorkos implements IFGAManager<WorkOSUser> {
       );
     }
 
+    this.options = options;
     this.workos = new WorkOS(apiKey, { clientId });
     this.organizationId = options.organizationId;
     this.resourceMapping = options.resourceMapping ?? {};
@@ -131,6 +135,20 @@ export class MastraFGAWorkos implements IFGAManager<WorkOSUser> {
     this.auditProtectedRoutes = options.auditProtectedRoutes;
     this.resolveRouteFGA = options.resolveRouteFGA;
     this.validatePermissions = options.validatePermissions;
+  }
+
+  withConfigOverrides(overrides: FGAProviderConfigOverrides): MastraFGAWorkos {
+    return new MastraFGAWorkos({
+      ...this.options,
+      resourceMapping: {
+        ...this.resourceMapping,
+        ...overrides.resourceMapping,
+      },
+      permissionMapping: {
+        ...this.permissionMapping,
+        ...overrides.permissionMapping,
+      },
+    });
   }
 
   // ──────────────────────────────────────────────────────────────
@@ -537,7 +555,9 @@ export class MastraFGAWorkos implements IFGAManager<WorkOSUser> {
           ? ['workflow', 'workflows']
           : resourceType === 'tool'
             ? ['tool', 'tools']
-            : resourceType === 'thread'
+            : resourceType === 'mcpTool'
+              ? ['mcpTool', 'mcpTools']
+              : resourceType === 'thread'
               ? ['thread', 'threads', 'memory']
               : [resourceType];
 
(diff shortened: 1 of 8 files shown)


LekoArts:
Fixed by https://github.com/mastra-ai/mastra/pull/17529

PR #17516, opened 2026-06-03
fix(acp): persist tool sessions across executions

wardpeet:
This fixes ACP tool session persistence so a single `createACPTool` instance reuses its ACP connection across executions by default. `persistSession: false` still keeps prompts isolated by creating and disconnecting a fresh connection each time.

Before, repeated calls to the same ACP tool started a new ACP connection even though sessions were documented as persistent. After this change, repeated calls through the same tool instance share the session/process unless persistence is disabled.

Test plan:
- `pnpm --filter ./packages/acp test`
- `pnpm --filter ./packages/acp build:lib`

## ELI5

Imagine you're using a tool that needs to talk to an AI assistant. Before this change, every time you used the tool, it would hang up and make a brand new phone call. Now, it stays on the same call the whole time (unless you specifically ask it to hang up), which is faster and lets the assistant remember what you've been talking about.

## Overview

This PR fixes ACP tool session persistence to ensure a single `createACPTool` instance reuses its ACP connection across multiple executions by default, rather than creating a new connection each time.

### Core Implementation

**New `ACPToolSession` class** (`agent-sdks/acp/src/session.ts`):
- Manages connection lifecycle for ACP tools
- `getConnection(workspace)` returns a fresh connection when `persistSession === false`, otherwise lazily creates and caches a connection on first use and reuses it for all subsequent calls
- `createConnection(workspace)` helper constructs the `ACPConnection` with proper workspace resolution

**Updated `createACPTool` function** (`agent-sdks/acp/src/tool.ts`):
- Now instantiates an `ACPToolSession` from provided options
- Tool's `execute` handler calls `session.getConnection(workspace)` instead of directly creating new `ACPConnection` instances

### Testing

**New test coverage** (`agent-sdks/acp/src/__tests__/tool.test.ts`):
- Verifies default session reuse: repeated `execute` calls with the same tool instance reuse the same `sessionId`
- Verifies non-persistent mode: when `persistSession: false`, each `execute` creates a fresh connection and calls process `kill`
- Confirms ACP file access uses the runtime Mastra workspace by mocking `mastra.getWorkspace().filesystem.readFile`

### Documentation

**Changeset** (`.changeset/free-doors-laugh.md`):
- Patch version bump for `@mastra/acp`
- Documents fix ensuring ACP tools keep their default session alive across executions

## Impact

Fixes prior behavior where repeated calls to the same ACP tool started a new ACP connection despite sessions being documented as persistent. After this change, repeated calls through the same tool instance share the session/process by default, unless `persistSession: false` is explicitly set.

### DIFF
--- a/agent-sdks/acp/README.md
+++ b/agent-sdks/acp/README.md
@@ -115,7 +115,7 @@ const codeAgentTool = createACPTool({
 
 ## Session and workspace behavior
 
-`createACPTool` and `AcpAgent` start the configured command on first use and create an ACP session. Sessions persist across calls by default. Set `persistSession: false` to stop the ACP process after each prompt.
+`createACPTool` and `AcpAgent` start the configured command on first use and create an ACP session. Sessions persist across calls for the same tool or agent instance by default. Set `persistSession: false` to isolate each prompt and stop the ACP process after it completes.
 
 ```typescript
 const codeAgent = new AcpAgent({
--- a/agent-sdks/acp/README.md
+++ b/agent-sdks/acp/README.md
@@ -115,7 +115,7 @@ const codeAgentTool = createACPTool({
 
 ## Session and workspace behavior
 
-`createACPTool` and `AcpAgent` start the configured command on first use and create an ACP session. Sessions persist across calls for the same tool or agent instance by default. Set `persistSession: false` to isolate each prompt and stop the ACP process after it completes.
+`createACPTool` and `AcpAgent` start the configured command on first use and create an ACP session. Sessions persist across calls by default. Set `persistSession: false` to stop the ACP process after each prompt.
 
 ```typescript
 const codeAgent = new AcpAgent({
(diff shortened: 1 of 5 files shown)


agent-sdks/acp/README.md
```diff
@@ -115,7 +115,7 @@ const codeAgentTool = createACPTool({
 
 ## Session and workspace behavior
 
-`createACPTool` and `AcpAgent` start the configured command on first use and create an ACP session. Sessions persist across calls by default. Set `persistSession: false` to stop the ACP process after each prompt.
+`createACPTool` and `AcpAgent` start the configured command on first use and create an ACP session. Sessions persist across calls for the same tool or agent instance by default. Set `persistSession: false` to isolate each prompt and stop the ACP process after it completes.
```
wardpeet:
```suggestion
`createACPTool` and `AcpAgent` start the configured command on first use and create an ACP session. Sessions persist across calls by default. Set `persistSession: false` to stop the ACP process after each prompt.
```

abhiaiyer91 (approved):

PR #17540, opened 2026-06-03
fix(datadog): forward score events from DatadogBridge to LLM Observability

### LINKED ISSUE #17527: DatadogBridge does not forward eval/score events (onScoreEvent) — intended, or planned?
**Package:** `@mastra/datadog@1.2.3` (current `latest`), `@mastra/observability@1.12.0`, `@mastra/core@1.33.x`.

**Context:** We're wiring production LLM Observability for an agent/workflow app. We chose `DatadogBridge` specifically for its live APM correlation — auto-instrumented HTTP/DB calls inside tools nest correctly under their Mastra spans, and the LLM Obs span shares the APM trace (clickable APM ↔ LLM Obs). That part works great.

We also want Mastra scorer results to surface in Datadog LLM Observability as `@evaluation.*` annotations. Reading the package source, this appears **unsupported by the bridge**:

- `DatadogExporter` implements `onScoreEvent` (calls `tracer.llmobs.submitEvaluation(...)`).
- `DatadogBridge` does **not** implement `onScoreEvent` (or `addScoreToTrace`) — no score/evaluation handling anywhere in the class.
- `BaseExporter` (in `@mastra/observability`) doesn't define `onScoreEvent` either, and the observability bus only dispatches score events to handlers where `handler.onScoreEvent` exists.

Net effect: with `DatadogBridge` as the exporter, `ScoreEvent`s are **silently dropped** — they never reach Datadog.

**Questions:**

1. Is eval/score forwarding (`onScoreEvent` → `submitEvaluation`) **intended to be added to `DatadogBridge`**? If so, is there a rough timeline?
2. Is the recommended way to get evals into Datadog *today* to use `DatadogExporter` instead — accepting the loss of live APM correlation / in-tool span nesting that the bridge provides?
3. Is running **both** `DatadogBridge` (for spans + APM correlation) **and** `DatadogExporter` (for scores only) in the same `Observability` config a supported pattern? In our reading both export spans on `span_ended` (→ duplicate spans), and the exporter's `onScoreEvent` is coupled to its own span-export pipeline (it looks up the dd-trace IDs *it* assigned), so a "scores-only exporter" subclass that no-ops span export would drop every score. Is there a supported score-only mode we're missing?

intojhanurag:
## Description

DatadogBridge previously did not implement `onScoreEvent`, so `ScoreEvent`s were silently dropped when the bridge was used instead of `DatadogExporter`. This PR adds score/evaluation forwarding to Datadog LLM Observability via `tracer.llmobs.submitEvaluation`, matching `DatadogExporter` semantics .

image

## Related Issue(s)

Fixes #17527

## Type of Change

- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [ ] Test update

## Checklist

## ELI5

DatadogBridge now sends evaluation scores to Datadog so scorer results show up in Datadog LLM Observability instead of being silently dropped.

## Overview

Implements ScoreEvent handling in DatadogBridge by adding onScoreEvent(event: ScoreEvent) that forwards scorer results to tracer.llmobs.submitEvaluation (metricType: 'score') when the tracer LLMObs API is available. The bridge converts hex span IDs produced by its dd-trace spans into Datadog's decimal span IDs before submission, logs and drops events missing required IDs, and guards against submit/export errors. Tests were added to validate forwarding and related edge cases.

## Changes

- DatadogBridge (observability/datadog/src/bridge.ts)
  - Added public method: onScoreEvent(event: ScoreEvent): Promise<void> to forward ScoreEvent payloads via tracer.llmobs.submitEvaluation.
  - Introduced toDatadogSpanId helper to convert 16-hex span IDs to decimal BigInt strings; uses converted IDs when submitting evaluations.
  - Refactored constructor/config assignment to explicitly set mlApp, site, apiKey, agentless, env.
  - onScoreEvent warns and drops events missing traceId/spanId and logs errors if submitEvaluation throws.

- Tests (observability/datadog/src/bridge.test.ts)
  - Extended dd-trace mock with mockSubmitEvaluation and adjusted mock span context to return hex or decimal IDs depending on the hex flag.
  - Added createScoreEvent helper for typed score payloads.
  - New tests verify:
    - onScoreEvent forwards scores to submitEvaluation using bridge-created trace/span IDs.
    - Forwarding works when no local finished-span context exists by using event-provided identifiers.
    - Edge behaviors: missing IDs are dropped; submission errors are logged and do not crash.

- Changeset (changeset/plenty-radios-drum.md)
  - Patch release note documenting that DatadogBridge now forwards scores to Datadog LLM Observability.

## Notes / Behavior changes

- Score forwarding requires available traceId/spanId (either bridge-produced decimal IDs or event-provided IDs); missing IDs result in a dropped event with a warning.
- The bridge no longer relies on a finished-span cache for score submission — it submits evaluations using decimal span IDs derived from dd-trace spans to preserve live APM correlation while surfacing evaluations in Datadog LLM Observability.

### DIFF
--- a/observability/datadog/src/bridge.ts
+++ b/observability/datadog/src/bridge.ts
@@ -510,9 +510,17 @@ export class DatadogBridge extends BaseExporter implements ObservabilityBridge {
       });
     }
 
-    const exported = tracer.llmobs?.exportSpan ? tracer.llmobs.exportSpan(ddSpan) : undefined;
-    if (exported?.traceId && exported?.spanId) {
-      this.rememberFinishedSpanContext(span.traceId, span.id, exported);
+    try {
+      const exported = tracer.llmobs?.exportSpan ? tracer.llmobs.exportSpan(ddSpan) : undefined;
+      if (exported?.traceId && exported?.spanId) {
+        this.rememberFinishedSpanContext(span.traceId, span.id, exported);
+      }
+    } catch (error) {
+      this.logger.error('[DatadogBridge] Failed to export dd span for score lookup', {
+        error,
+        spanId: span.id,
+        spanName: span.name,
+      });
     }
 
     try {
--- a/observability/datadog/src/bridge.ts
+++ b/observability/datadog/src/bridge.ts
@@ -23,6 +23,7 @@ import type {
   ModelStepAttributes,
   ObservabilityBridge,
   CreateSpanOptions,
+  ScoreEvent,
   SpanType,
   SpanIds,
 } from '@mastra/core/observability';
@@ -62,6 +63,13 @@ interface TraceContext {
   userId?: string;
   sessionId?: string;
 }
+
+interface FinishedSpanContext {
+  traceId: string;
+  spanId: string;
+}
+
+const DEFAULT_FINISHED_SPAN_CONTEXT_CACHE_MAX_ENTRIES = 10_000;
 
 function flushApmExporter(): Promise<void> {
   const exporterFlush = (tracer as any)?._tracer?._exporter?.flush;
@@ -152,6 +160,11 @@ export interface DatadogBridgeConfig extends BaseExporterConfig {
    * LLM Observability tags instead of being nested in annotations.metadata.
    */
   requestContextKeys?: string[];
+
+  /**
+   * Maximum number of finished span contexts to retain for score forwarding.
+   */
+  finishedSpanContextCacheMaxEntries?: number;
 }
 
 /**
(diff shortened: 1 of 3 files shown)


observability/datadog/src/bridge.ts
```diff
@@ -291,6 +315,53 @@ export class DatadogBridge extends BaseExporter implements ObservabilityBridge {
...
+      this.logger.warn(
+        'Datadog bridge: dropping score for span that has not been finished and exported to dd-trace yet',
+        {
+          traceId: score.traceId,
+          spanId: score.spanId,
+          scorerId: score.scorerId,
+        },
+      );
+      return;
+    }
```
epinzur: i don't think we should maintain a cache of finished spans or check to see if a span is finished before sending the score to llmObs.

We should just send the score to llmObs.  If there is an error in the send, then perhaps we maintain a cache of failed-sends and retry later and/or on flush().

epinzur (changes requested):
pls remove cache of completedSpans for score tracking

observability/datadog/src/bridge.ts
```diff
@@ -291,6 +299,40 @@ export class DatadogBridge extends BaseExporter implements ObservabilityBridge {
...
+  async onScoreEvent(event: ScoreEvent): Promise<void> {
+    if (this.isDisabled || !(tracer as any).llmobs?.submitEvaluation) return;
+
+    const { score } = event;
+    if (!score.traceId || !score.spanId) {
+      this.logger.warn('Datadog bridge: dropping score with no traceId/spanId', {
+        scorerId: score.scorerId,
+      });
+      return;
+    }
```
epinzur, line 311: i think this will drop trace-level scores... `spanId` would be undefined in this case.

epinzur: Actually seems I'm incorrect. Seems we do set `spanId` for trace level scoring.

epinzur (approved):

PR #17469, opened 2026-06-02
feat(core): route Gateway embedding models

arctic-char:
## Summary
- route `mastra/provider/model` embedding model IDs through Mastra Gateway
- add embedding-router coverage for Gateway routing
- document Gateway embedding model usage
- keep observational memory local for Gateway-routed actor models, so users opt in with explicit observer/reflector model config instead of implicit Gateway-managed memory behavior

## Test plan
- `pnpm --filter @mastra/core test:unit embedding-router.e2e.test.ts`
- `pnpm --filter @mastra/core typecheck`
- `pnpm vitest run packages/memory/src/processors/observational-memory/__tests__/gateway-skip.test.ts --reporter=dot`
- `pnpm --filter ./packages/memory check`

Depends on platform support for `/v1/embeddings`.

## ELI5

This PR adds support for routing embedding model requests through Mastra's Gateway service. Users can now use special model IDs like `mastra/openai/text-embedding-3-small` to send embeddings through the Gateway instead of directly to providers. The PR also ensures that local memory processing continues to work correctly and prevents accidental interference with platform-managed memory.

## Overview

Introduces first-class Gateway embedding routing support by enabling `ModelRouterEmbeddingModel` to accept `mastra/provider/model` identifiers, which resolve to upstream Gateway endpoints. Includes comprehensive test coverage, documentation updates, and related observational-memory refinements.

### Core Gateway Embedding Support
- Extended `ModelRouterEmbeddingModel` to handle `mastra/provider/model` format in model ID parsing
- When `providerId` resolves to `mastra`, the constructor now requires `MASTRA_GATEWAY_API_KEY` (with `config.apiKey` as an override option)
- Initializes OpenAI-compatible embedding models using the Gateway base URL (defaults to `https://gateway-api.mastra.ai/v1` or respects explicitly configured `MASTRA_GATEWAY_URL`)
- Normalizes configured Gateway URLs to `/v1` form and includes `MASTRA_USER_AGENT` in request headers
- Maintains backward compatibility for existing non-Mastra provider routing

### Testing & Documentation
- Added comprehensive e2e test suite (`embedding-router.e2e.test.ts`) covering:
  - Mastra-prefixed ID normalization and model ID rewriting
  - Gateway base URL handling (explicit, blank/whitespace fallback, normalization)
  - OpenAI-compatible client initialization with correct headers and configuration
  - Error handling when `MASTRA_GATEWAY_API_KEY` is missing
[...]

PR #17543, opened 2026-06-03
fix(core,memory): keep observational memory local for Gateway models

arctic-char:
## Summary

Split out from #17469 so the Gateway embedding routing change can land independently.

Two related fixes that ensure agents using Mastra Gateway actor models still get **locally-configured** Observational Memory instead of silently deferring to platform-managed Gateway memory:

1. **`@mastra/memory`** — Stop short-circuiting OM for Gateway-routed actor models. The processor now runs locally for `provider: 'mastra'` actor models and records the actor model context (`{ provider, modelId }`) so observer/reflector models are invoked as configured.

2. **`@mastra/core`** — Stop auto-forwarding internal `_internal.threadId` / `_internal.resourceId` as `x-thread-id` / `x-resource-id` headers to the Gateway. Those headers are how the platform activates its server-side memory, and forwarding them implicitly bypassed local OM. Explicit user / model-settings headers are still respected.

## Behavior change

Agents whose actor model is a Gateway-routed ID (e.g. `mastra/openai/gpt-5`) will now run the locally configured `observationalMemory.observation` and `observationalMemory.reflection` models, instead of deferring to whatever the Gateway might do.

If you were previously relying on the Gateway to manage memory implicitly via internal thread/resource forwarding, you can opt back in by setting `x-thread-id` / `x-resource-id` explicitly via `modelSettings.headers`.

## Test plan

- `packages/memory` — new `gateway-skip.test.ts` asserts OM begin/end turns fire for Gateway actor models and that actor context is recorded.
- `packages/core` — `llm-execution-step.test.ts` updated to assert internal memory context is **not** auto-forwarded as Gateway headers, while explicit user headers still are.
- Focused suites pass locally:
  - `packages/core/src/loop/workflows/agentic-execution/llm-execution-step.test.ts` — 18/18
  - `packages/memory/src/processors/observational-memory/__tests__/gateway-skip.test.ts` — 2/2

## ELI5 Explanation

When agents use Mastra's Gateway models, they should use their own configured memory tools (like custom observers) instead of automatically relying on the Gateway's built-in memory system. This PR fixes two things to make that happen: (1) it makes sure the agent's local memory processor actually runs for Gateway models, and (2) it stops automatically sending internal memory context to the Gateway that would trigger its memory handling.

## Key Changes

**Memory Package (`@mastra/memory`)**
- Updated `ObservationalMemoryProcessor` to process Gateway-routed actor models locally instead of short-circuiting
- Local OM processor now runs and records actor model context (`{ provider, modelId }`) so configured observer and reflector models execute as intended
- Added regression test coverage in `gateway-skip.test.ts` verifying local OM processing occurs for Gateway models

**Core Package (`@mastra/core`)**
- Modified `llm-execution-step.ts` to stop auto-forwarding internal `_internal.threadId` and `_internal.resourceId` as `x-thread-id` and `x-resource-id` headers to the Gateway
- Updated header merge semantics so `modelSettings.headers` override model config headers
- Gateway-specific headers are no longer automatically injected; explicit user or model headers remain respected
- Updated corresponding test assertions in `llm-execution-step.test.ts` to verify headers are not auto-generated from internal context

## Behavior Changes

Gateway-routed actor models (e.g., `mastra/openai/gpt-5`) will now execute locally-configured `observationalMemory.observation` and `observationalMemory.reflection` instead of deferring to platform-managed Gateway memory. Users who previously relied on implicit Gateway memory handling can opt back in by explicitly setting `x-thread-id` and `x-resource-id` via `modelSettings.headers`.

## Testing

- New test file `packages/memory/src/processors/observational-memory/__tests__/gateway-skip.test.ts` added with comprehensive coverage of local OM processing for Gateway models
- `packages/core/src/loop/workflows/agentic-execution/llm-execution-step.test.ts` updated to verify non-generation of internal-context headers and correct header merge behavior

### DIFF
--- a/.changeset/gateway-memory-handling.md
+++ b/.changeset/gateway-memory-handling.md
@@ -1,0 +1,6 @@
+---
+'@mastra/core': patch
+'@mastra/memory': patch
+---
+
+Updated observational memory processing so agents using Mastra Gateway actor models still run the locally configured observer and reflector models instead of implicitly deferring to Gateway-managed memory behavior. Internal thread/resource context is no longer forwarded as Gateway request headers, preventing platform-managed Gateway memory from being activated by local agent memory context.
--- a/packages/core/src/loop/workflows/agentic-execution/llm-execution-step.test.ts
+++ b/packages/core/src/loop/workflows/agentic-execution/llm-execution-step.test.ts
@@ -653,7 +653,7 @@ describe('createLLMExecutionStep gateway provider tools', () => {
     });
   });
 
-  it('preserves model config headers when modelSettings adds non-conflicting headers', async () => {
+  it('preserves model config headers when modelSettings adds non-conflicting headers without internal memory context', async () => {
     const doStream = vi.fn(async () => ({
       stream: convertArrayToReadableStream([
         {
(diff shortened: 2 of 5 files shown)


abhiaiyer91 (approved):

abhiaiyer91:
closing for now

PR #17560, opened 2026-06-04
fix(playground-ui): make popup portal container resolution null-safe

damien-schneider:
## Summary
Follow-up hardening to #17556. Base UI's `FloatingPortal` treats `container={null}` as *"not ready — render nothing"*, while `undefined` means *"portal to `document.body`"*. Our `PortalContainerContext` defaulted to `null`, so popups opened outside a `SideDialog` (the model & provider pickers live in the chat composer, which has no `PortalContainerProvider` ancestor) resolved their portal container to `null` and mounted to no DOM node — they opened in state but rendered nowhere.

## Change
- `usePortalContainer(container?)` now resolves `explicit → provider → undefined` and **never returns `null`**; the "no container" sentinel is `undefined` to match Base UI's contract.
- Dropped the scattered `?? undefined` band-aids in `Combobox`, `Select`, `Popover`, and `DropdownMenu` — they now just call `usePortalContainer(container)`.

## Tests
- `portal-container.test.tsx` guards the resolver contract (never `null`; normalizes a `null` provider; explicit wins).
- `combobox.test.tsx` asserts the popup portals into `document.body`.
- Verified these fail under the old `null`-leaking behavior (combobox open/select tests go red) and pass with the fix.

## ELI5

Imagine a magical doorway that decides where popup menus (like dropdowns and options) should appear on the screen. This PR fixes a broken doorway that prevented popups from showing up when opened outside certain panels—now the doorway always works and shows popups in the right place (the main document area) as a fallback.

## Problem

Base UI's FloatingPortal treats `container={null}` as "not ready — render nothing" while `undefined` means "portal to document.body". The PortalContainerContext was defaulting to `null`, which caused popups opened outside a SideDialog (such as model and provider pickers in the chat composer) to fail rendering because there was no valid DOM node to attach to.

## Solution

The `usePortalContainer` hook was redesigned to:
- Accept an optional `container` parameter that can explicitly override the default behavior
- Resolve the final portal target with precedence: explicit argument → context provider value → `undefined`
- Never return `null`; always return `undefined` as the "no container" sentinel to match Base UI's expectations

The PortalContainerContext now treats `undefined` as the default (fall back to document.body) and normalizes any incoming `null` values to `undefined`.

## Changes Made

- **portal-container.tsx**: Added `PortalContainer` type, updated `usePortalContainer` to accept an optional container override and exclude null from its return type, and normalized null values in the provider context
- **Combobox, Select, Popover, and DropdownMenu**: Replaced manual fallback logic (`container ?? portalContainer ?? undefined`) with direct calls to `usePortalContainer(container)`
- **portal-container.test.tsx**: Added comprehensive tests asserting the resolver contract (never returns null, normalizes null provider to undefined, explicit values override provider values)
- **combobox.test.tsx**: Added test verifying popups mount into `document.body` when no portal container provider is present

## Testing

Tests were added to verify the contract of the portal-container resolver and confirm that popups now mount correctly when opened outside a SideDialog. Tests were verified to fail under the previous null-leaking behavior and pass with the fix.

### DIFF
--- a/.changeset/proud-gifts-appear.md
+++ b/.changeset/proud-gifts-appear.md
@@ -1,0 +1,5 @@
+---
+'@mastra/playground-ui': patch
+---
+
+Fixed dropdown, combobox, select, and popover popups silently failing to render when opened outside a side panel (most visibly the model and provider pickers in the chat composer). The shared portal-container resolver now always falls back to the document body instead of leaking an unrenderable value.
--- a/packages/playground-ui/src/ds/components/Combobox/combobox.test.tsx
+++ b/packages/playground-ui/src/ds/components/Combobox/combobox.test.tsx
@@ -45,6 +45,20 @@ describe('Combobox', () => {
     expect(screen.getByRole('option', { name: 'Google' })).toBeTruthy();
   });
 
+  it('portals the popup into document.body when there is no portal container provider', async () => {
+    const { container } = renderCombobox();
+
+    fireEvent.click(screen.getByRole('combobox'));
+
+    // The regression: outside a SideDialog the portal container resolved to
+    // `null`, which Base UI's FloatingPortal reads as "render nothing", so the
+    // popup never mounted. It must land in document.body, outside the trigger's
+    // own subtree.
+    const option = await screen.findByRole('option', { name: 'OpenAI' });
+    expect(document.body.contains(option)).toBe(true);
+    expect(container.contains(option)).toBe(false);
+  });
+
   it('selects an item and fires onValueChange with the selected value', async () => {
     const onValueChange = vi.fn();
     renderCombobox({ onValueChange });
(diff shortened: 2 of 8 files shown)


damien-schneider:
Some context on the root cause: base-ui's `FloatingPortal` treats `container={null}` and `container={undefined}` differently. `null` hits an early-return that intentionally renders nothing (the comment says "Wait for the container to be resolved if explicitly null"): https://github.com/mui/base-ui/blob/master/packages/react/src/floating-ui-react/components/FloatingPortal.tsx#L94-L129. That branch is for lazy refs, a container that isn't mounted yet, so it avoids falling back to body in the meantime. `undefined` is the "no container, use document.body" path.

We were the ones emitting `null`: `PortalContainerContext` defaulted to `null` (#17479's `createContext<HTMLElement | null>(null)`) and `container ?? portalContainer` forwarded it, so outside a SideDialog the popup rendered `<X.Portal container={null}>` and never mounted.

This fixes it at the source: the sentinel is `undefined` and `usePortalContainer` never returns `null`. We also collapse `null` to `undefined` in the provider since we don't rely on base-ui's lazy-wait (popups open after the drawer host resolves).

wardpeet (approved):

PR #14852, opened 2026-03-30
fix(core): supervisor/sub-agent abort signal

arctic-char:
## Description

Ensures `AbortSignal` sent to supervisor agent is passed to sub-agents, aborts propagate to children.

## Notes
Draft PR as issue contains "Needs Discussion", these changes are intended to facilitate discussion.

## Related Issue(s)

Fixes #14820 

## Type of Change

- [x] Bug fix (non-breaking change that fixes an issue)
- [ ] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [ ] Test update

## Checklist

- [x] I have made corresponding changes to the documentation (if applicable)
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] I have addressed all Coderabbit comments on this PR

## Summary by CodeRabbit

* **New Features**
  * Supervisor agents now propagate abort signals to delegated sub-agents across all execution types (generate, stream, resume operations). When a supervisor stops via AbortController or client disconnect, in-flight sub-agent work is automatically cancelled instead of continuing independently.

* **Documentation**
  * Updated documentation to clarify abort signal propagation behavior in supervisor-to-sub-agent delegation patterns.

PR #17561, opened 2026-06-04
fix(core): forward parent abortSignal to delegated sub-agents

### LINKED ISSUE #14820: how to abort subagents executions in supervisor pattern ?
This issue was created from Discord post 1487828672361795694:

[Open in Discord](https://discord.com/channels/1309558646228779139/1487828672361795694)

So far i only managed stopping the main agent using abortcontroller but no idea how to propagate this down to the subagent ?

KiraPC:
The sub-agent delegation tool gets the parent's `abortSignal` on its execution context but never passed it down to the sub-agent's `stream()`/`generate()` calls. So aborting a supervisor only stopped the supervisor itself — each delegated sub-agent kept looping and calling its own tools and the LLM for several seconds after the abort, observing a detached signal that never fired.

This matters beyond correctness: when a user cancels (or an HTTP client disconnects), the delegated sub-agents keep making LLM calls and running tools that nobody is waiting for, wasting compute and cost. Forwarding the signal stops that work at the next step.

The fix forwards `context.abortSignal` to every delegation path (`stream`, `generate`, `resumeStream`, `resumeGenerate`, and the legacy variants), so aborting the parent cancels in-flight sub-agents at their next step. It also covers `streamUntilIdle`, which delegates through the same tool. The delegation already runs sub-agents inline (`disableBackgroundTasks: true`), so this is just the missing signal forwarding and changes no background-task behavior.

Before:

```typescript
const stream = await supervisor.stream('...', { abortSignal: controller.signal })
controller.abort() // supervisor stops, but delegated sub-agents keep calling the LLM
```

After:

```typescript
const stream = await supervisor.stream('...', { abortSignal: controller.signal })
controller.abort() // supervisor and any in-flight sub-agents stop
```

Added regression tests for the stream and generate paths (they fail without the change — the sub-agent tool sees an `undefined` signal), plus a Cancellation section in the supervisor docs and a note on the `abortSignal` parameter in the generate/stream references.

Note: this is aware of the earlier #14852, which proposed the same forwarding but was closed for inactivity before it landed. This revives that fix on the current codebase with regression tests and docs. Credit to the original author for the approach.

Fixes #14820.

## ELI5

Imagine you have a manager (supervisor agent) who delegates tasks to workers (sub-agents). If you tell the manager to stop working, the workers were ignoring you and kept working anyway. This PR fixes that by passing the "stop" signal all the way down to the workers so they stop too when you cancel the supervisor.

### Core Fix
The delegation tool in `Agent.ts` now forwards the parent's `abortSignal` to all delegated sub-agent calls, ensuring that cancelling a supervisor also stops all in-flight sub-agent executions. The signal is propagated across all code paths: `stream`, `generate`, `resumeStream`, `resumeGenerate`, and legacy variants (`streamLegacy`, `generateLegacy`).

### Testing
A new regression test suite was added (`supervisor-integration.test.ts`) with test cases covering both `stream()` and `generate()` delegation paths. The tests verify that:
- The `abortSignal` is successfully forwarded to delegated sub-agents
- When the parent's `AbortController` is aborted, the delegated sub-agent receives and observes the aborted signal

### Documentation Updates
- **Supervisor agents guide**: Added a "Cancellation" section explaining how passing an `abortSignal` to `supervisor.stream()` or `supervisor.generate()` propagates to delegated subagents, with a TypeScript example
- **Agent.generate() reference**: Clarified that aborting terminates ongoing operations and any in-flight subagent runs via delegation
- **Agent.stream() reference**: Updated to clarify that aborting includes cancellation of delegated subagent runs
- **Changeset entry**: Documented the fix with a TypeScript example demonstrating supervisor cancellation

## Impact
Prevents wasted compute and costs by ensuring sub-agents stop execution when the parent supervisor is cancelled, rather than continuing to make LLM calls and run tools after the parent request is aborted.


abhiaiyer91 (approved):

KiraPC:
Any news on this guys? I see it is approved. Do you need anything?

TheIsrael1:
Thanks for contributing @KiraPC !

PR #17584, opened 2026-06-04
Add hosted source storage Studio UI

DanielSLew:
## Summary
- Surface source-provider capabilities in Studio and show provider-aware save/history UI
- Inject hosted Platform context through playground HTML, CLI studio serving, and deployer serving
- Add MSW coverage for source-provider and unavailable source-storage UI states

## Test plan
- pnpm --filter ./packages/deployer exec vitest run src/build/utils.test.ts
- pnpm --filter ./packages/playground exec vitest run src/domains/agents/components/__tests__/agent-playground-version-bar.msw.test.tsx
- pnpm --filter ./packages/playground-ui build
- pnpm --filter ./packages/deployer build:lib
- pnpm --filter ./packages/playground build
- git diff --check

Note: playground typecheck currently has an unrelated existing error in src/services/mastra-runtime-provider.tsx for isAwaitingToolApproval.

Stack: PR 3 of 4 in the Mastra code-source stack. Base is #17583.

## ELI5 (Explain Like I'm 5)

This PR teaches the Studio interface where your code changes will get saved. Instead of always saving to a file, the Studio can now save to different places (like a code provider or a file system), and it tells you which one based on what the platform provides. The PR also makes sure the right configuration information is passed to the Studio so it knows where to save things.

---

## Overview

This PR adds hosted source-provider capabilities to Studio, enabling provider-aware save/history UI and injecting Platform context through multiple serving paths (playground HTML, CLI studio server, and deployer server). It includes MSW test coverage for source-provider and unavailable source-storage UI states.

---

### Configuration & Environment Setup
- **Release metadata**: Updated `.changeset/hosted-code-source-provider.md` to mark `@mastra/deployer` and `mastra` as `minor` releases, with release notes documenting hosted source-provider context injection
- **Platform context injection**: CLI studio command and deployer server now read `MASTRA_PLATFORM_PROJECT_ID` from environment and inject it into Studio HTML via `%%MASTRA_PLATFORM_PROJECT_ID%%` placeholder
- **HTML templating**: Playground `index.html` now includes `window.MASTRA_PLATFORM_PROJECT_ID` global variable

### Platform Detection Logic
- **App.tsx**: Updated platform detection to read `MASTRA_CLOUD_API_ENDPOINT` into a variable and verify it doesn't start with placeholder prefix `%%` before treating it as valid
[...]

PR #17585, opened 2026-06-04
Add propose and inspect source change UX

DanielSLew:
## Summary
- Replace the hosted source-change action with a “Propose change” flow
- Add optional commit-message entry and defer opening the pull request until “Inspect change”
- Wire custom source-change messages through the Studio callback and verify the interaction with MSW

## Test plan
- pnpm --filter ./packages/playground exec vitest run src/domains/agents/components/__tests__/agent-playground-version-bar.msw.test.tsx
- pnpm --filter @mastra/server exec vitest run src/server/handlers/stored-agents.test.ts
- pnpm --filter @mastra/client-js exec vitest run src/resources/stored-agent.test.ts
- pnpm --filter ./packages/playground build
- pnpm --filter @mastra/client-js build:lib
- git diff --check

Stack: PR 4 of 4 in the Mastra code-source stack. Base is #17584.

## ELI5 (Explain Like I'm 5)

This PR changes the code-change workflow from immediately opening a pull request to a two-step process: users can now click "Propose change," optionally enter a commit message, and then later click "Inspect change" to actually open the PR. This gives users more time to review their proposed changes before the PR is created.

### Changeset
Added a changeset documenting the new "Propose change" workflow in code-mode, which replaces the previous hosted source-change action with a deferred PR-opening flow.

### UI Components

**Agent Playground Version Bar Component** (`agent-playground-version-bar.tsx`)
- Replaced direct "Open PR" button with a two-step "Propose change" flow
- Added internal state to capture an optional commit/proposal message via a new `Dialog`
- Updated `onOpenPr` prop signature to accept `changeMessage?: string` and return `{ url: string }` (the PR URL)
- Stores the returned PR URL and provides a dropdown option to "Inspect change" (which opens the PR in a new tab using `window.open`)
- Shows a loading spinner while the proposal is being submitted

**Agent Playground View** (`agent-playground-view.tsx`)
- Updated `AgentPlaygroundViewProps.onOpenPr` signature to accept optional `changeMessage` and return `{ url: string } | void`

**Agent Playground Page** (`agent-playground/index.tsx`)
- Now retrieves `mastraPlatformUserName` from `useMastraPlatform`
- Updated `handleOpenPrClick` callback to accept and pass the optional `changeMessage` and `userName` to `handleOpenPr`

**Edit Layout Page** (`cms/agents/edit-layout.tsx`)
- Added UI state for the propose dialog, commit message input, and PR URL storage
- Replaced the direct "Open PR" action with a "Propose change" primary button and a dropdown "Inspect change" option
- Integrated the new two-step flow with visual feedback (loading state during proposal)

### Hook

**use-agent-cms-form Hook**
- Updated `handleOpenPr` to accept `platformOptions` (including optional `userName`) and optional `changeMessage`
- Changed behavior from directly opening a PR URL in a new tab to calling `openChangeRequest` with the message and username, then returning the result
- Success toast now displays "Change proposed" instead of "Pull request opened"

### Tests

**Agent Playground Version Bar MSW Test** (`agent-playground-version-bar.msw.test.tsx`)
- Added a comprehensive test case that exercises the full propose-and-inspect flow:
  - Mocks the PR-opening handler
  - Stubs `window.open`
  - Simulates clicking "Propose change" and entering a custom commit message
  - Verifies the proposal is submitted with the message
  - Verifies "Inspect change" opens the PR URL with proper window parameters


DanielSLew:
https://github.com/coderabbitai review

DanielSLew:
@coderabbitai review

mfrachet (approved):

abhiaiyer91:
## Review: Request changes

**Verdict:** Request changes — the branch is `CONFLICTING` against its base, and the auto-fired discovery request deserves a second look before merge.

### What I verified (head branch only — not verified against current base)

Ran locally on `85ddc71`, credentials stripped:

| Check | Result |
| --- | --- |
| `packages/playground` full unit suite | ✅ 141 files / 1009 tests |
| `agent-playground-version-bar.msw.test.tsx` | ✅ 4 tests |
| `@mastra/server` `stored-agents.test.ts` | ✅ 71 passed / 1 skipped |
| `@mastra/client-js` `stored-agent.test.ts` | ✅ 27 tests |
| `@mastra/deployer` `build/utils.test.ts` | ✅ 85 tests |
| `packages/playground` `tsc --noEmit` | ✅ clean |
| `eslint` on the 4 changed playground files | ✅ clean |

Effective diff against the base tip is 17 files (+474/−48). Nothing in the diff touches CI workflows, install scripts, or network/credential paths. The new `MASTRA_PLATFORM_USER_NAME` global is escaped before HTML injection in both the deployer (`escapeForHtml`) and the CLI (`escapeJsonForHtml`), and `sanitize()` drops the unreplaced `%%…%%` placeholder — that part is solid.

### Blocking

**1. Merge conflicts with the base branch.** `mergeStateStatus: DIRTY`. A trial merge against `feat/code-source-04-studio-provider-ui` conflicts in 7 files: `deployers/vercel/src/index.ts`, `packages/cli/src/commands/studio/studio.ts`, `packages/deployer/src/build/utils.ts`, `packages/deployer/src/build/utils.test.ts`, `packages/deployer/src/server/index.ts`, `packages/playground/index.html`, `packages/playground/src/domains/agents/hooks/use-agent-cms-form.ts`, plus an add/add on the MSW test. The conflicts look mechanical — the base branch was rebased onto newer `main` and the stack commits were rewritten (base tip `2565683f2c` duplicates this branch's `654d0e9e91`). A rebase onto the current base should resolve them, but the results above only hold for the head branch as it stands.

**2. Discovery fires a write-shaped POST on every mount.** Both `agent-playground-version-bar.tsx:177` and `edit-layout.tsx:304` run an effect that calls `openChangeRequest({ inspectOnly: true })` as soon as `canOpenPr` is true — before the user has expressed any intent. On the server that path does more than read: it calls the provider with `files: []`, then `agentsStore.useProviderRef(storedAgentId, result.ref ?? headRef)` and `editor.agent.clearCache(...)`. `useProviderRef` drops the hydrated agent, its loaded history and cached provider versions, and re-hydrates from the change-request ref. So simply opening the agent playground on a hosted source project repoints that agent's active source ref at the proposal branch, and the client then invalidates `stored-agent` / `agent` / `agent-versions`.

That's a meaningful behavior change for a page load. Two things I'd want addressed:
- Should discovery be a read? A `GET`-style query (React Query `useQuery`) against a read-only "does a change request exist" endpoint would avoid the ref mutation and get caching/dedup for free. As written, two mounted consumers (playground + CMS edit layout) each fire their own POST.
- The handler destructures `inspectOnly` out of the body and never forwards it to `provider.openChangeRequest`, even though `SourceChangeRequestInput` declares the flag. The provider only sees a change request with zero files, so whether this discovers or *creates* is entirely up to broker behavior we can't see from here. (That handler lives on the base branch, so it's out of scope for this PR to fix — but this PR is the first caller, so it's the first time the ambiguity actually matters.)

### Should fix

**3. The new wiring has no test coverage.** The two added MSW tests stub `onOpenPr` / `onInspectPr` as props, so they exercise the version bar's local state machine but never `use-agent-cms-form`. The genuinely new contract — `userName` sourced from `useMastraPlatform()`, `changeMessage` passthrough, `inspectOnly: true` on the discovery call — is not asserted anywhere in this PR. Per `.claude/skills/playground-msw-tests`, the intended approach is to drive the real `@mastra/client-js` + React Query stack through MSW handlers; an MSW handler on `POST /stored/agents/:id/change-request` asserting the request body would cover this directly.

**4. The propose dialog is duplicated verbatim.** `showProposeDialog` / `proposeMessage` / `isProposingChange` / `proposedChangeUrl`, the discovery effect, `handleProposeChange`, `handleInspectChange`, and ~60 lines of dialog JSX are copy-pasted between `agent-playground-version-bar.tsx` and `edit-layout.tsx`. Two copies of a state machine that talks to the server will drift. A shared hook plus a shared `<ProposeChangeDialog>` would collapse most of it.

**5. Failures silently discard the user's commit message.** `handleOpenPr` catches its own errors and returns `undefined`, so `handleProposeChange` can't distinguish success from failure — it closes the dialog and clears `proposeMessage` either way. The user sees a "Failed to propose change" toast with their typed message already gone. Keeping the dialog open on failure would be a small change.

### Nits

**6. Changeset misses `@mastra/deployer-vercel`.** `deployers/vercel/src/index.ts` changes here, and `StudioInjectionConfig` gains a *required* `platformUserName` field, so the vercel deployer must ship in lockstep. Runtime is safe if it lags (the placeholder survives and `sanitize()` strips it), but the feature just won't work on Vercel deployments until it's released.

**7. Brittle button selection in the test** (`agent-playground-version-bar.msw.test.tsx:100`) — CodeRabbit already flagged this. `getAllByRole('button', { name: /Propose Change/i })` then indexing by `length - 1` assumes DOM order between the toolbar trigger and the dialog submit. `within(dialog).getByRole(...)` is stable.

**8. Generated-file drift.** `route-types.generated.ts` and `route-metadata.generated.ts` pick up unrelated routes (`POST /agents/:agentId/threads/abort`, `POST /agents/:agentId/send-tool-approval`) that have nothing to do with this feature — regeneration noise from the stale base. The rebase in (1) should clear it.

### Prior signal

@mfrachet approved at `7d84b23e`, a commit that is no longer an ancestor of this head — the branch was rewritten after that approval, and two commits (`9bf28f54`, `85ddc716`) landed afterward, including the auto-discovery effect in (2). Treating that approval as covering the current head would be a stretch. CodeRabbit's only substantive note is (7).

The core UX — deferring PR creation behind an explicit "Propose change" with an optional commit message, then a separate "Inspect change" — is a clear improvement over the old fire-and-open-a-tab button, and the platform-user-name plumbing is clean end to end. Items (1) and (2) are what I'd want resolved before this merges.

PR #17594, opened 2026-06-04
feat(core): return ownerStream from sendSignal/sendMessage

CalebBarnes:
## Why

`Agent.sendSignal` (and the `sendMessage` / `queueMessage` / `sendStateSignal` / `sendNotificationSignal` wrappers) has no way to hand the caller the run it just woke. When the call transitions a thread from idle into a new run, the runtime fires `agent.stream(...)` as `void agent.stream(...).catch(...)` and the promise is unreachable.

That makes it impossible for a caller to:

- drive the run to completion before doing follow-up work
- attach its own consumer to the run's `MastraModelOutput`
- decide whether to keep its host process alive while the run drains

This PR exposes that promise so callers can opt into any of the above.

## What this PR does

Threads an optional `ownerStream` through the signals return types:

- `SendAgentSignalResult`
- `SendAgentMessageResult`
- `QueueAgentMessageResult`
- `SendAgentStateSignalResult`
- `SendAgentNotificationSignalResult`

`ownerStream` is `Promise<MastraModelOutput<OUTPUT>>` and is **only set when the call wakes a new run from idle**. For deliver / join-remote / queue-into-active-run / persist / discard paths it is `undefined`.

Internally, the wake path in `thread-stream-runtime.sendSignal` already calls `void agent.stream(...)` and detaches it. This PR captures the promise instead and returns it. The existing `.catch` for cleanup is preserved, and an internal no-op `.catch` is attached so callers that don't opt in still don't see unhandled rejection warnings.

`MastraModelOutput` is lazy, so awaiting `ownerStream` alone just resolves to the wrapper. To drive the run to completion, await `consumeStream()` (or any of the delayed-promise getters like `.finishReason` / `.text`, which kick off consumption automatically). `consumeStream()` is idempotent and safe to call even if another consumer is already draining.

```ts
const result = await agent.sendMessage(message, target);
if (result.ownerStream) {
  const ownerStream = await result.ownerStream;
  await ownerStream.consumeStream();
}
```

The generics on `Agent.sendMessage` / `queueMessage` / `sendSignal` / `sendStateSignal` / `sendNotificationSignal` are threaded through so `ownerStream` carries the agent's `OUTPUT` type.

## Scope

Pure addition to the signals return type. No behavior changes for existing callers that ignore the new field.

## Test plan

- [x] `pnpm build:core`
- [x] `pnpm --filter ./packages/core check`

### DIFF
--- a/.changeset/signal-owner-stream.md
+++ b/.changeset/signal-owner-stream.md
@@ -1,0 +1,18 @@
+---
+"@mastra/core": minor
+---
+
+Add `ownerStream` to the return value of `Agent.sendSignal`, `Agent.sendMessage`, `Agent.queueMessage`, `Agent.sendStateSignal`, and `Agent.sendNotificationSignal`.
+
+When the call wakes a new agent run from idle, `ownerStream` is the `Promise<MastraModelOutput<OUTPUT>>` for that run. When the signal is delivered to an already-running stream, joins a remote run, is persisted, queued, or otherwise does not start a new stream, `ownerStream` is `undefined`.
+
+Awaiting `ownerStream` keeps the caller alive until the run completes, which is required for serverless environments (Vercel, AWS Lambda, etc.) where the handler would otherwise exit immediately after responding and kill the run mid-flight.
+
+```ts
+const result = await agent.sendMessage(message, target);
+if (result.ownerStream) {
+  // This call woke a new run. Await it so the run can drive any stream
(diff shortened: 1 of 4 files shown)


CalebBarnes:
Superseded by #17723, which combines the ownerStream return with the PubSub lease API for cross-process signal coordination.

PR #17595, opened 2026-06-04
feat(core/channels): await ownerStream in AgentChannels webhook

CalebBarnes:
**Stacked on top of #17594.** Targets `caleb/agent-channels-await-completion` until that lands, then will be retargeted to `main`.

## Why

`AgentChannels.processChatMessage` fires `agent.sendMessage(...)` and returns immediately. The agent run is driven by a side-effect subscription (`ensureThreadSubscription` + `chatThread.subscribe()`). That works in `mastra dev` because the Node event loop keeps consumer microtasks alive, but it breaks under serverless runtimes (Vercel, AWS Lambda, etc.) where the function process is terminated as soon as the handler returns. The agent stream is killed mid-flight and the user sees nothing, partial replies, or duplicate Slack retries piling up.

See #17065 for user-reported symptoms.

## What this PR does

In `processChatMessage`, capture the `sendMessage` result and, when it carries an `ownerStream` (added in #17594), await `ownerStream.consumeStream()` before returning from the webhook handler:

```ts
const result = this.agent.sendMessage(...);

if (result.ownerStream) {
  try {
    const ownerStream = await result.ownerStream;
    await ownerStream.consumeStream();
  } catch (err) {
    this.log('debug', 'ownerStream consume failed', err);
  }
}
```

`ownerStream` is only present when the call wakes a new run from idle. For deliver / join-remote / queue / persist / discard paths it's `undefined` and the handler returns as before.

`consumeStream()` is idempotent and safe to call alongside the existing per-thread subscription consumer that forwards chunks to the chat platform.

Error handling on the await is intentionally permissive (debug-log + swallow), because the run's own error path is already handled via the stream pipeline.

## What this PR does NOT do

- Does not move platform-forwarding (Slack/Discord API calls) from per-process subscriptions into a stream processor. That's a follow-up that removes the need for cross-Lambda subscription coordination entirely.
- Does not change anything about pubsub. Multi-Lambda deployments still require `@mastra/redis-streams` (or another shared `PubSub`) for signals to cross instances.

## Test plan

- [x] `pnpm build:core`
- [x] `pnpm --filter ./packages/core check`
- [ ] Manual: deploy to Vercel + Slack repro and confirm bot replies arrive end-to-end.

## Related

- #17594 (parent PR adding `ownerStream` to the signals return type)
- #17065

### DIFF
--- a/.changeset/agent-channels-await-owner-stream.md
+++ b/.changeset/agent-channels-await-owner-stream.md
@@ -1,0 +1,9 @@
+---
+"@mastra/core": patch
+---
+
+`AgentChannels` now awaits the agent run to completion when an incoming chat message wakes a new run.
+
+Previously the webhook handler fired `agent.sendMessage(...)` and returned immediately, relying on a side-effect subscription to drive the agent stream. That works in long-lived Node processes but breaks in serverless runtimes (Vercel, AWS Lambda, etc.) where the function exits as soon as it returns and kills the run mid-flight, producing missing or partial responses.
+
+When `sendMessage` returns an `ownerStream` (i.e. the call woke a new run from idle), the handler now awaits `ownerStream.consumeStream()` before returning, keeping the host alive for the duration of the run. Calls that join an in-flight run, queue, persist, or discard are unaffected.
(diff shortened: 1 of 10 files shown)


CalebBarnes:
Superseded by #17832 (channels consumer + waitUntil + output processor + multi-step fix, stacked on #17723).

PR #17596, opened 2026-06-04
feat(pg): add Agent Builder storage parity with libSQL

YujohnNattrass:
PostgreSQL was missing some storage that Agent Builder relies on, so projects on PG couldn't persist tool provider connections or agent tool providers, and a few builder tables were left out of the exported schema. This brings the PG adapter in line with libSQL.

It adds a tool-provider-connections domain with full CRUD, persists agent version toolProviders across save and load, and gives the remaining builder domains (blobs, skills, workspaces, mcp-clients, mcp-servers) a getExportDDL so all 20 domains show up in exportSchemas(). Verified against a live Postgres instance with new and existing integration tests.

## ELI5

This PR brings PostgreSQL database support up to speed with libSQL by adding the ability to save and manage "tool provider connections" (configurations that let agents use external tools). It also makes sure that when you save an agent, its list of available tools gets saved too, so the agent remembers which tools it should use when you load it back up.

## Overview

This PR introduces full Agent Builder storage parity with libSQL in the PostgreSQL adapter. The changes span three main areas:

1. **New tool provider connections domain** – A complete CRUD implementation for persisting tool provider connection configurations
2. **Agent version tool provider persistence** – Ensures agent versions maintain their associated tool providers across save/load cycles
3. **Export DDL implementations** – Adds schema export support for blobs, skills, workspaces, mcp-clients, and mcp-servers domains

### New Tool Provider Connections Storage
- Introduced `ToolProviderConnectionsPG` class with full lifecycle management
- Implements core operations: `getConnectionById`, `upsertConnection`, `listConnectionsByAuthor`, and `deleteConnection`
- Uses composite primary keys `(authorId, providerId, connectionId)` with upsert-on-conflict semantics
- Includes comprehensive test suite validating insert, retrieve, update, list, and delete flows
- Properly manages timestamps (`createdAt`, `updatedAt`) and connection scope normalization

### Agent Version Tool Provider Persistence
- Extended `AgentsPG` to persist `toolProviders` field on agent versions
- `agents_versions` table altered to include `toolProviders` column during initialization
- Legacy migration updated to copy `toolProviders` from old snapshot data
- `createVersion` and `parseVersionRow` methods updated to handle tool provider serialization/deserialization

### Schema Export DDL Methods
Added `getExportDDL(schemaName?: string)` static methods to:
- `BlobsPG`
- `SkillsPG` 
- `WorkspacesPG`
- `MCPClientsPG`
- `MCPServersPG`

These methods generate DDL statements for their respective tables with proper schema prefixing for non-`public` schemas.

### Index Definitions Refactoring
Multiple domains refactored default index definitions into schema-aware static helpers (`getDefaultIndexDefs(schemaPrefix: string)`), enabling both runtime index creation and DDL export while supporting custom schema configurations.

### Integration
- Updated `stores/pg/src/storage/index.ts` to import, export, and wire `ToolProviderConnectionsPG` into the `PostgresStore`
- Extended `ALL_DOMAINS` ordering to include the new domain in schema exports
- Verified against live PostgreSQL instances using new and existing integration tests


stores/pg/src/storage/domains/blobs/index.ts
```diff
@@ -18,6 +18,17 @@ export class BlobsPG extends BlobStore {
     this.#schema = schemaName || 'public';
   }
 
+  static getExportDDL(schemaName?: string): string[] {
```
YujohnNattrass, line 21: looks like some domains are missing this, so it was added here

abhiaiyer91 (approved):

PR #17602, opened 2026-06-05
fix(core): preserve Anthropic signed thinking across memory replay

### LINKED ISSUE #17457: Opus 4.8 + adaptive thinking: replayed thinking blocks lose text but keep signature → 400 "thinking blocks cannot be modified" on multi-turn
### Summary

With the model router pointed at `anthropic/claude-opus-4-8` and **adaptive** extended thinking, multi-turn conversations backed by `Memory` fail on the Anthropic API with a 400:

```
messages.1.content.8: `thinking` or `redacted_thinking` blocks in the latest assistant message
cannot be modified. These blocks must remain as they were in the original response.
```

(`invalid_request_error`, HTTP 400, from `AnthropicMessagesLanguageModel.doStream`)

### Root cause (observed)

When a prior assistant turn is replayed from Memory on a follow-up request, its thinking blocks are sent back with an **empty `thinking` text but the original `signature` retained**:

```json
{ "type": "thinking", "thinking": "", "signature": "EtcDCmMIDhgC…<truncated>==" }
```

Anthropic re-validates the signature against the (now empty) text and rejects the block as "modified". This only triggers on a **continuation** — a follow-up message into a thread that already contains a signed-thinking assistant turn. Fresh single-shot runs are fine.

### Versions

- `@mastra/core` 1.37.0
- `@mastra/memory` 1.20.0 (PostgresStore via `@mastra/pg`)
- `ai` 6.0.x
- model: `anthropic/claude-opus-4-8` (model router)
- `providerOptions.anthropic`: `{ thinking: { type: "adaptive" }, effort: "high" }` → serialized to `thinking: { type: "adaptive" }` + `output_config: { effort: "high" }`

### Repro

1. `Agent` with `Memory` (Postgres), tools enabled, and the adaptive-thinking provider options above on `claude-opus-4-8`.
2. Run a turn that produces thinking + tool calls; let it complete and persist to the thread.
3. Send a follow-up message in the **same** thread.
4. The next request 400s as above.

### Notes

- The same agent on `claude-opus-4-6` with `thinking: { type: "enabled", budgetTokens: 10000 }` does **not** reproduce — multi-turn works fine. The regression appeared only after switching model + thinking mode to 4.8/adaptive.
- Appears analogous to the Gemini `thoughtSignature` round-trip fixes (#14972, #14900), but for Anthropic-native thinking signatures: `ProviderHistoryCompat` preserves the block, but the reasoning **text** is dropped on persist→replay while the signature is kept.
- No mention of `opus-4-8` / adaptive thinking in any changelog through `1.38.0-alpha.6`.

### Expected

Persisted Anthropic thinking blocks should round-trip verbatim (text **and** signature), or be safely stripped from completed prior turns before send so the signature/text invariant can't break.

Akash504-ai:
## Description

Fixes an issue where Anthropic signed thinking blocks could lose their reasoning text during persistence/replay while retaining the original signature.

When replayed in a multi-turn conversation, Anthropic validates the signature against the original thinking content and rejects modified blocks with:

> thinking blocks cannot be modified

This change:

* Preserves Anthropic signed reasoning text during persistence/replay
* Rehydrates reasoning text from either `reasoning` or `details` for backward compatibility
* Adds a defensive compatibility rule that strips invalid signed reasoning blocks with empty text before Anthropic replay
* Adds regression tests covering round-trip persistence and replay behavior

## Related issue(s)

Fixes #17457

## Type of change

* [ ] Documentation update
* [ ] Code refactoring
* [ ] Performance improvement
* [ ] Test update

## Checklist

* [x] I have linked the related issue(s) in the description above

## ELI5

When Anthropic conversation history is saved and reused, the system now keeps the reasoning text with its signature. If old history has a signature but no text, the system removes that invalid block before sending it back.

## Changes

- Preserve Anthropic signed reasoning during memory persistence and replay.
- Rehydrate reasoning text from `reasoning` or `details` for compatibility.
- Remove empty signed reasoning blocks before Anthropic requests.
- Preserve non-Anthropic reasoning and Gemini `thoughtSignature` metadata.
- Add regression tests for adapter, memory, multi-turn, and provider-history flows.
- Add a test that verifies concatenated reasoning text and Anthropic signatures remain intact.
- Update the changeset for the compatibility fix.


graysonhicks:
Verifying this request for changes:

- **Warning — stale conflict can regress non-Anthropic reasoning.** The PR conditionally persists reasoning only when an Anthropic signature exists ([lines 676 and 915](https://github.com/mastra-ai/mastra/blob/a0c1b3e5e7661016f025367221843a595bb76bea/packages/core/src/agent/message-list/adapters/AIV5Adapter.ts#L673-L676)). Current `main` now persists it unconditionally via [#17803](https://github.com/mastra-ai/mastra/commit/fe98ef2e66dbfcbd7d645c88c9ee1e67b458a136). Rebase and resolve both conflicts in `main`’s favor, otherwise non-Anthropic reasoning is lost again.

- **Info — the stream finalization change is obsolete.** Current `main` accumulates `ref.reasoning` on every delta via [#18534](https://github.com/mastra-ai/mastra/commit/b8375c1f8fe905df8ae2ae9a893bb365f17aec4e), so the PR’s Anthropic-only recomputation at `reasoning-end` is redundant. Remove its two helpers and source change; the regression assertion can remain.

- **Info — narrow the changeset.** Forward reasoning persistence has already shipped. The release note should describe only the still-needed legacy empty-signed-block compatibility rule.

cc: @intojhanurag

abhiaiyer91:
AI says: @graysonhicks, the requested changes from your review have been addressed:

- merged current `main` and resolved both `AIV5Adapter.ts` conflicts in `main`'s favor, preserving reasoning unconditionally
- removed the superseded Anthropic-only adapter and stream-finalization helpers/source changes
- retained the stream regression assertion
- kept the legacy empty-signed-block compatibility rule and round-trip coverage
- narrowed the changeset to the legacy-history sanitization behavior

Verification on the updated branch: 80/80 targeted tests passed, core typecheck passed, core build passed, and format/lint passed. CodeRabbit has also completed successfully, and there are no unresolved review threads.

Could you please re-review when convenient?

— Mastra Code

PR #17610, opened 2026-06-05
fix(mcp): evaluate resource and template providers per request

### LINKED ISSUE #17609: [BUG] MCPServer caches resources/list on the shared instance, leaking resources across authenticated callers (cross-tenant)
### Describe the Bug

# Summary

MCPServer caches the result of the first resources/list call on the shared, long-lived server instance (`this.definedResources`) and replays it to every subsequent caller, ignoring per-request auth. For servers whose resource lists are per-user / per-tenant (resolved from `extra.authInfo`), this leaks one caller's resource index (titles + URIs) to the next caller.

# Affected versions

Confirmed in @mastra/mcp@1.9.0 and @mastra/mcp@1.9.1.

# Where it happens
        
In the `ListResources` request handler (registered via `registerResourceHandlersOnServer`):

```typescript
serverInstance.setRequestHandler(ListResourcesRequestSchema, async (_request, extra) => {
  if (this.definedResources) {
    return { resources: this.definedResources };   // serves the first caller's list to everyone
  }
  const resources = await capturedResourceOptions.listResources({ extra });
  this.definedResources = resources;               // cached on the shared instance
  return { resources: this.definedResources };
});
``` 

`this.definedResources` lives on the shared `MCPServer` instance. Even in serverless/stateless mode `(startHTTP({ options: { serverless: true } }))`, `createServerInstance()` produces a per-request transient server, but its handlers close over the shared this, so the cache is process-wide. The same cache is also used by the ReadResource handler (to resolve URIs) and by the `public MCPServer.listResources()` method. It is only ever cleared on an external resource-list-change notification — never per request or per caller.
    
# Impact

- A dynamic `listResources({ extra })` provider that returns auth-scoped resources will leak resource metadata (names + URIs) across authenticated users/tenants.
- The first authenticated caller to hit resources/list populates the cache; all later callers receive that cached list regardless of
their identity.
- Resource content reads may still be protected if getResourceContent re-checks auth, but the resource listing is leaked.

# Suggested fix

- Do not populate/read `this.definedResources` for the dynamic-provider path; always call `listResources({ extra })` with the current request's extra.
- Alternatively, only cache when resources are supplied as a static list (no per-request extra dependency), and document that dynamic providers are evaluated per request.

# Workaround

Consumers can neutralize the shared cache by redefining `definedResources` as a no-op sink on the instance so every resources/list re-runs the provider — but this relies on a private field name and is fragile across releases, hence this issue.

### Steps To Reproduce

1. Create an `MCPServer` with a dynamic resource provider whose `listResources({ extra })` returns different resources depending on `extra.authInfo` (e.g. tenant-scoped).
2. Call resources/list as tenant A.
3. Call resources/list as tenant B.
4. Expected: tenant B receives tenant B's resources.
5. Actual: tenant B receives tenant A's cached resources; the provider is invoked only once.

https://github.com/usepalette/palette-monorepo/pull/1461 - fix

The public `listResources()` passes `extra = {}`, so this isolates the caching defect itself; in production the same `definedResources` field is shared by the `ListResources` request handler, which receives the real per-caller extra.authInfo —that's where it becomes a cross-tenant leak.

```typescript
import { MCPServer } from "@mastra/mcp";

  let tenant = "tenant-A";
  const server = new MCPServer({
    name: "repro", version: "1.0.0", tools: {},
    resources: {
      listResources: async () => [
        { uri: `app://${tenant}/doc`, name: `Doc for ${tenant}`,
  mimeType: "text/plain" },
      ],
      getResourceContent: async ({ uri }) => ({ text: uri }),
    },
  });

  console.log((await server.listResources()).resources[0].name); //
  "Doc for tenant-A"
  tenant = "tenant-B";
  console.log((await server.listResources()).resources[0].name); //
  BUG: still "Doc for tenant-A"
``` 

### Expected Behavior

When resources is configured as a dynamic provider (callback-based `listResources`/`getResourceContent`), the result must not be cached on the shared instance across requests. The provider should be invoked per request with the current extra, or the cache should be keyed by auth/request and never shared across callers. A static resource array can still be cached safely.

### Environment Information

```block
System:
    OS: macOS 26.2
    CPU: (14) arm64 Apple M4 Pro
    Memory: 177.92 MB / 24.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 24.15.0 - /Users/tiffanylozenski/.local/state/fnm_multishells/64524_1780650152277/bin/node
    npm: 11.12.1 - /Users/tiffanylozenski/.local/state/fnm_multishells/64524_1780650152277/bin/npm
    pnpm: 10.30.3 - /opt/homebrew/bin/pnpm
  Browsers:
    Chrome: 149.0.7827.53
    Firefox: 150.0
    Safari: 26.2
  npmPackages:
    @biomejs/biome: ^2.4.14 => 2.4.14 
    @commitlint/cli: ^20.4.3 => 20.4.3 
    @commitlint/config-conventional: ^20.4.3 => 20.4.3 
    @next/eslint-plugin-next: ^16.1.6 => 16.1.6 
    @palette/design-system: workspace:* => 0.0.0 
    @semantic-release/exec: ^7.1.0 => 7.1.0 
    @types/node: catalog: => 25.3.3 
    @typescript-eslint/eslint-plugin: catalog: => 8.59.1 
    @typescript-eslint/parser: catalog: => 8.59.1 
    dotenv-cli: ^11.0.0 => 11.0.0 
    eslint: catalog: => 9.39.3 
    eslint-plugin-import: ^2.32.0 => 2.32.0 
    eslint-plugin-react: ^7.37.5 => 7.37.5 
    eslint-plugin-react-hooks: ^7.0.1 => 7.0.1 
    eslint-plugin-sonarjs: ^4.0.1 => 4.0.1 
    husky: ^9.1.7 => 9.1.7 
    knip: ^5.86.0 => 5.86.0 
    lint-staged: ^16.3.2 => 16.3.2 
    madge: ^8.0.0 => 8.0.0 
    npm-check-updates: ^19.6.3 => 19.6.3 
    semantic-release: ^25.0.3 => 25.0.3 
    simple-git: ^3.36.0 => 3.36.0 
    ts-api-utils: ^2.4.0 => 2.4.0 
    typescript: catalog: => 5.9.3 
    typescript-eslint: catalog: => 8.59.1 
  npmGlobalPackages:
    @codacy/analysis-cli: 0.9.0
    corepack: 0.34.6
    npm: 11.12.1
    ntn: 0.14.1
```

### Verification

tiffanybalc:
## Description

`MCPServer` cached the first `resources/list` result on the shared, long-lived instance and replayed it to every later caller. For a dynamic provider that scopes resources per user/tenant via `extra.authInfo`, that leaked one caller's resource index (names + URIs) to the next. The same stale cache also backed `resources/read` URI resolution and the public `listResources()` method, and `resources/templates/list` had the identical bug.

Before, the first caller populated the cache and everyone got that same list:

```ts
serverInstance.setRequestHandler(ListResourcesRequestSchema, async (_request, extra) => {
  if (this.definedResources) return { resources: this.definedResources };
  const resources = await listResources({ extra });
  this.definedResources = resources; // shared across callers
  return { resources };
});
```

After, the provider runs on every request with the current caller's `extra`:

```ts
serverInstance.setRequestHandler(ListResourcesRequestSchema, async (_request, extra) => {
  const resources = await listResources({ extra });
  return { resources };
});
```

One side effect worth flagging for a follow-up: `resources/read` now re-invokes `listResources({ extra })` to resolve the requested URI's metadata instead of reading the cache, so each read costs one provider call. A reasonable optimization later would be resolving a single URI without listing everything (e.g. an optional `getResource({ uri, extra })` lookup) so expensive providers don't pay the full-list cost on every read.

Verified end-to-end by driving a real MCP client over a transport: switching the provider's tenant between calls returns fresh data each time for list, read, and templates, and the per-request `extra` (including `authInfo`) reaches the provider.

## Related issue(s)

Fixes #17609

## Type of change

- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [ ] Test update

## Checklist

- [x] I have made corresponding changes to the documentation (N/A — internal behavior fix, no public API or docs change)

## ELI5

The server used to remember the first person's resource list and then gave that same list to everyone else. This PR makes the server ask the configured providers on every request so each caller gets a fresh, personalized resource list.

## Problem

MCPServer stored the first results of resources/list and resources/templates/list on the shared server instance and reused them for subsequent callers. Dynamic providers that depend on per-request context (e.g., extra.authInfo/tenant) could therefore leak one caller’s resource index (titles + URIs) to others. resources/read and the public listResources() also relied on that cached index for URI resolution.

## Solution

Remove the shared definedResources / definedResourceTemplates caching and invoke resource and template providers per request with the current caller’s extra context. Handlers (ListResources, ReadResource URI resolution, ListResourceTemplates) now call providers each request and log provider errors; ServerResourceActions no longer expects or calls cache-clear callbacks and simply sends the list-changed notification. A construction-time hasUiResources flag was added to record whether appResources exist for advertising the io.modelcontextprotocol/ui extension.

Key changes:
- No shared instance cache for dynamic resource/template providers; providers are invoked per request.
- ListResources, ReadResource (URI lookup), and ListResourceTemplates now use current extra for each call.
- ServerResourceActions dependency interface simplified (removed clearDefinedResources/clearDefinedResourceTemplates) and notifyListChanged() no longer clears caches.
- MCPServer records hasUiResources at construction to decide UI extension advertising.
- Note: resources/read currently resolves URIs by re-invoking the list provider; a potential follow-up is adding a targeted getResource({ uri, extra }) to avoid full-list costs.

## Testing

Added Vitest regression tests (src/server/__tests__/server-resource-cache.test.ts) that:
- Verify dynamic providers respect extra.authInfo.subject and return tenant-specific lists and templates.
- Assert listResources(), resources/read, and templates/list re-invoke providers per caller.
- Confirm the public server.listResources() invokes the provider on repeated calls (prevents regression).

## Related Issue

Fixes `#17609`.


tiffanybalc:
@coderabbitai review

Aayush-engineer:
I independently implemented the same fix removing the shared definedResources cache and invoking the provider per request. Happy to close my branch since you got there first. One thing I'd suggest adding: a regression test that explicitly calls listResources() twice with a dynamic provider and asserts the callback is invoked both times, to prevent this from silently regressing. Let me know if I can help get this merged.

tiffanybalc:
Addressed the nitpick in e7ef09cb95 — collapsed the redundant outer conditional to the single `hasUiTools || this.hasUiResources` check that determines whether the UI extension is advertised. No behavior change; 102 tests + typecheck pass locally.

Note: I followed the suggested diff rather than the "Prompt for AI Agents" text, which proposed `hasUiTools || this.resourceOptions || this.hasUiResources` — that would advertise the UI extension for any resources, changing behavior.

tiffanybalc:
> I independently implemented the same fix removing the shared definedResources cache and invoking the provider per request. Happy to close my branch since you got there first. One thing I'd suggest adding: a regression test that explicitly calls listResources() twice with a dynamic provider and asserts the callback is invoked both times, to prevent this from silently regressing. Let me know if I can help get this merged.

@Aayush-engineer thanks, I added a regression test

tiffanybalc:
@Aayush-engineer I may need your help getting approval for workflows

Aayush-engineer:
> @Aayush-engineer I may need your help getting approval for workflows

Unfortunately I don't have write access to approve workflows only maintainers can do that. Try tagging @mikhael28 @abhiaiyer91  or @daneatmastra directly.

tiffanybalc:
@daneatmastra @abhiaiyer91 @mikhael28 please approve the workflows to unblock this critical bug fix 🙏

abhiaiyer91 (approved):

PR #17622, opened 2026-06-05
Reduce Agent Builder instruction churn

NikAiyer:
## Summary
- Tune Studio Agent Builder's generated-instructions guidance so prompts stay concise without exposing internal hard-limit/counting language to the model.
- Keep the runtime over-length rejection guard in the client tool, but return model-facing guidance as a concise target instead of a hard-limit protocol.
- Add deterministic Agent Builder Playwright coverage for all starter templates plus a complex freeform prompt.
- Fix filesystem stored-agent handling so dynamically created Agent Builder agents are treated as stored agents, not unresolved code agents.

## Why
The previous Agent Builder prompt wording made the model churn through repeated `set-agent-instructions` calls and often produce near-limit instructions. The narrowed guidance keeps the same quality checklist but avoids teaching the model to count against the internal limit.

The new E2E coverage also exercises the actual starter/freeform create flow that exposed the filesystem stored-agent edge case: user-created stored agents are not present in `mastra.getAgentById()`, so code-agent detection must treat unknown IDs as normal stored agents instead of throwing.

## Test plan
- `pnpm build --force`
- `pnpm --filter ./packages/playground test:e2e -- e2e/tests/agent-builder/deterministic-builder.spec.ts`
- Manual scaffold smoke: Support triage starter completed with `set-agent-instructions` called 2 times and final instructions length `2249` chars.

## ELI5
This PR improves the Agent Builder tool (used to create AI agents through a UI) by making the instructions clearer and simpler for the AI model to understand (removing confusing internal rules), adding comprehensive tests to verify everything works correctly, and fixing how agents saved to disk are recognized and handled.

## Overview
This PR makes several improvements to the Agent Builder feature and its E2E testing coverage:

### Instruction Guidance Refinement
The Agent Builder now provides clearer, more concise guidance to the model when generating agent instructions. The system no longer exposes internal hard-limit counting language to the model. Instead, it presents a concise target (1,200–2,000 characters) focusing on essential sections (trigger, capabilities, source rules, response format, completion criteria) while avoiding worked examples, FAQs, and exhaustive lists. The client-side tool retains runtime rejection for over-length submissions, but the model-facing guidance is simplified to avoid instruction churn.

### Filesystem Agent Storage Fix
Fixed how dynamically created agents stored in the filesystem are recognized. Previously, agents created through the Agent Builder and persisted to disk were incorrectly treated as unresolved code agents. Now, unknown agent IDs are properly treated as stored agents rather than throwing errors. The implementation adds an `isAgentNotFoundError` utility to distinguish expected "agent not found" errors (which are suppressed) from unexpected failures (which are rethrown).

### Deterministic E2E Test Coverage
Added comprehensive Playwright E2E tests covering:
- All starter templates (support, standup, PR reviewer, onboarding)
- A complex freeform prompt scenario
- Regression guards ensuring each deterministic tool-call step appears exactly once
- Validation that configured form fields match expected values

Tests reset storage before each run and verify agent configuration through the UI to completion.

### Infrastructure Updates
- Created new E2E fixture module with standardized agent-builder fixture scenarios
- Updated E2E storage to use file-backed SQLite (instead of in-memory) for test persistence
- Extended `Fixtures` type and route interception patterns to support agent-builder fixtures
- Updated request interception to handle `stream-until-idle` and `send-message` endpoints

### Documentation
Added changesets documenting patch releases for:
- `mastra`: Update to Studio Agent Builder instruction generation guidance
- `@mastra/core`: Filesystem-stored agents no longer treated as code agents unless declared in code

### Tool Input Preservation
The tool implementation ensures client inputs are preserved across continuation and reconstruction paths during retries.

### Test Updates
Updated test assertions to verify:
- Over-limit instruction rejection returns human-readable character-range messaging (not hard limit values)
- Builder output includes expected agent name/description/instructions configuration
- Generated instructions follow the concise target structure without exposing internal limits


TylerBarnes (approved):
I didn’t find any blocking issues.

The only thing I’d double-check is that `FilesystemAgentsStorage` now treats any `getAgentById()` throw as “not a code agent.” That fixes the stored-agent miss path, but it also means unrelated lookup failures would silently fall back to shared stored-agent persistence. If that’s intentional/acceptable, this looks good to me.

I also verified the focused unit tests and the new Agent Builder canaries locally.

PR #17625, opened 2026-06-05
fix: apply ownership filtering to GET /agents endpoint

YujohnNattrass:
## Summary

The `GET /agents` endpoint was returning ALL stored agents without ownership filtering, leaking private agents from other users into the agent edit form's tools picker. This PR applies unconditional ownership filtering to the stored agents section of `LIST_AGENTS_ROUTE`, matching the pattern used by `GET /stored/agents`.

### `packages/server/src/server/handlers/agents.ts`
- Added import for `matchesAuthorFilter` and `resolveAuthorFilter` from `./authorship`
- Added `authorId` and `visibility` query parameters to `LIST_AGENTS_ROUTE`
- Call `resolveAuthorFilter()` with `requestContext`, `resource`, and optional query params
- Pass `authorId` and `visibility` to `editor.agent.list()` based on filter kind:
  - `exact` → `authorId`
  - `publicOnly` → `visibility: 'public'`
  - `ownedOrPublicOthers` → `authorId` + `visibility: 'public'`
- Apply `matchesAuthorFilter()` as post-filter for defense-in-depth

### `packages/server/src/server/handlers/agents.test.ts`
- Added 8 test cases covering ownership filtering scenarios:
  1. Non-admin user doesn't see private agents from other users
  2. Admin user sees all stored agents
  3. No caller (auth not configured) sees all stored agents
  4. Code-defined agents are always included regardless of ownership
  5. Legacy unowned agents (`authorId: null`) are visible to all users
  6. `?visibility=public` returns only public agents
  7. Non-admin `?authorId=other-user` returns only that user's public agents
  8. Admin `?authorId=other-user` returns all of that user's agents

### `packages/server/src/server/schemas/__tests__/route-contracts.typetest.ts`
- Updated type assertion for `ListAgentsQuery` to include new query params

## Test Plan

- [x] TypeScript compilation passes (`tsc --noEmit`)
- [x] Server tests pass (1758 tests, 4 new ownership tests added)
- [x] Playground tests pass (1007 tests)

## Behavior Changes

**Before**: `GET /agents` returned ALL stored agents regardless of ownership  
**After**: `GET /agents` returns only agents the caller is authorized to see

This is a **security fix**, not a breaking change. Any code that was relying on seeing unauthorized agents was already broken from a security perspective.

## Consistency

This makes `GET /agents` consistent with:
- `GET /stored/agents` (unconditional ownership filtering)
- `GET /stored/workflows` (same pattern)
- `GET /stored/skills` (same pattern)

All stored resource routes now apply ownership filtering by default.

### DIFF
--- a/packages/server/src/server/handlers/agents.test.ts
+++ b/packages/server/src/server/handlers/agents.test.ts
@@ -1955,3 +1955,244 @@ describe('extractVersionOptions', () => {
     expect(extractVersionOptions(undefined, {})).toBeUndefined();
   });
 });
+
+// =============================================================================
+// LIST_AGENTS_ROUTE stored agents ownership filtering
+// =============================================================================
+
+describe('LIST_AGENTS_ROUTE stored agents ownership', () => {
+  const MASTRA_USER_KEY = 'mastra__user';
+  const MASTRA_USER_PERMISSIONS_KEY = 'mastra__userPermissions';
+
+  function createCallerContext(userId: string, permissions: string[] = []) {
+    const ctx = new RequestContext();
+    ctx.set(MASTRA_RESOURCE_ID_KEY, userId);
+    ctx.set(MASTRA_USER_KEY, { id: userId });
+    if (permissions.length > 0) {
+      ctx.set(MASTRA_USER_PERMISSIONS_KEY, permissions);
+    }
+    return ctx;
+  }
+
+  function createMockEditor(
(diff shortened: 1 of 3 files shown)


YujohnNattrass:
closing, was comparing different models for this ticket  and seemed like it opened up a PR.

here is the PR I will be going with: https://github.com/mastra-ai/mastra/pull/17626/changes

Issue #17307, opened 2026-05-30
BUG: PostgresStore: SSL config not passed to internal sub-stores (MemoryPG, WorkflowsPG, etc.)


suspect-47:
### Describe the Bug

Bug: PostgresStore accepts an ssl config option, but internal sub-stores (MemoryPG, WorkflowsPG, ScoresPG, etc.) create their own pg connections that don't inherit the SSL config. This causes unable to get local issuer certificate errors when connecting to databases with self-signed CA certificates.

Version: @mastra/pg@1.11.1

Environment: Vercel (Node.js serverless), Next.js 16 with Turbopack

Database: InsForge BaaS (Postgres with self-signed CA cert)

What happens:

Error: unable to get local issuer certificate
  id: 'MASTRA_STORAGE_PG_CREATE_TABLE_FAILED'
  details: { tableName: 'mastra_threads' }
  code: 'UNABLE_TO_GET_ISSUER_CERT_LOCALLY'
Config:

new PostgresStore({
  id: "my-app",
  connectionString: "postgresql://...",
  ssl: { ca: MY_CA_CERT, rejectUnauthorized: true },
  max: 3,
})
What I tried (all failed on Vercel):

Passing ssl: { ca, rejectUnauthorized: true } to PostgresStore — sub-stores ignore it
Passing a pre-created pg.Pool with ssl config — sub-stores still create their own connections
Patching pg/lib/defaults.ssl globally — Turbopack bundles pg inline, patch doesn't reach bundled code
Setting PGSSLMODE=no-verify env var — Turbopack reads env at build time, not runtime
Stripping sslmode=require from connection string + passing ssl config — still fails
Root cause: In createPool(), the ssl config IS passed to pg.Pool. But internal domain stores (MemoryPG, WorkflowsPG, etc.) appear to create their own connections or pools that don't inherit the parent pool's SSL configuration.

Expected: All internal pg connections should use the same ssl config passed to PostgresStore.

Workaround: None found for Vercel + Turbopack. Works on local dev (macOS trusts more CAs). Agents function without persistent storage — they just lose conversation memory between sessions.

### Steps To Reproduce

1. Configure PostgresStore with a self-signed CA cert via ssl: { ca: cert, rejectUnauthorized: true }
2. Deploy to Vercel with Turbopack (Next.js 16)
3. Connect to a Postgres database that uses a self-signed CA
4. Mastra init fails with "unable to get local issuer certificate" on mastra_threads table creation
[...]

PR #17650, opened 2026-06-06
fix(pg): honor explicit ssl over sslmode in PgVector connection string

### LINKED ISSUE #17649: [BUG] PgVector ignores explicit `ssl` option when connection string contains `sslmode=`
`PgVector` does not honor an explicit `ssl` option when the connection string also
contains an `sslmode=` (or `ssl=`) query parameter. The explicit `ssl` object is
silently discarded, producing `UNABLE_TO_GET_ISSUER_CERT_LOCALLY` /
"self-signed certificate" errors against self-signed or internal CAs — even though
the caller explicitly asked to skip verification.

**Root cause**

In `stores/pg/src/vector/index.ts`, the connection-string branch of the `PgVector`
constructor passes both `connectionString` and `ssl` straight into `new pg.Pool(...)`:

```ts
poolConfig = {
  connectionString: config.connectionString,
  ssl: config.ssl,
  // ...
};
```

`node-postgres` re-parses `connectionString` via `pg-connection-string` and
`Object.assign`s the URL-derived result over the explicit options, so an
`sslmode=`/`ssl=` query param overrides the explicit `ssl` object.

This is the **same bug that was already fixed for `PostgresStore`** in #17356
(issue #17307) via a `buildConnectionStringPoolConfig` helper. That fix only
touched the storage code path — the parallel vector code path was missed.

### Steps To Reproduce

1. Stand up a Postgres instance that uses a self-signed certificate (or any
   internal CA), reachable with `sslmode=require`.
2. Create a `PgVector` with a connection string that includes `?sslmode=require`
   and an explicit `ssl` that skips verification:

   ```ts
   import { PgVector } from '@mastra/pg';

   const vector = new PgVector({
     id: 'my-vector',
     connectionString: 'postgresql://user:pass@host:5432/db?sslmode=require',
     ssl: { rejectUnauthorized: false },
   });

   await vector.listIndexes();
   ```
3. The connection fails with `UNABLE_TO_GET_ISSUER_CERT_LOCALLY` (self-signed
   certificate in certificate chain), because `{ rejectUnauthorized: false }` is
   ignored.

The equivalent `PostgresStore` configuration connects correctly, which confirms
the inconsistency.

https://github.com/mastra-ai/mastra (see `stores/pg/src/vector/index.ts`)

### Expected Behavior

An explicit `ssl` option should take precedence over `sslmode=`/`ssl=` in the
connection string, exactly like `PostgresStore`. Connection-string-only SSL
(`?sslmode=require` with no explicit `ssl`) should keep working unchanged.

### Environment Information

```block
- Package: `@mastra/pg` (PgVector)
- Affects all OS / Node versions; logic bug in connection-pool configuration.
```

### Verification

Abuhaithem:
Fixes #17649

## Problem

`PgVector` did not honor an explicit `ssl` option when the connection string also
contained an `sslmode=` (or `ssl=`) query parameter. `node-postgres` re-parses the
connection string and `Object.assign`s the URL-derived result over the explicit
options, so the explicit `ssl` object was silently discarded. This produced
`UNABLE_TO_GET_ISSUER_CERT_LOCALLY` / "self-signed certificate" errors against
self-signed or internal CAs even when the caller asked to skip verification.

This is the same class of bug fixed for `PostgresStore` in #17356 (issue #17307).
That fix introduced a `buildConnectionStringPoolConfig` helper but only wired it
into the storage path; the parallel `PgVector` code path was missed.

## Solution

- Moved the existing `buildConnectionStringPoolConfig` helper from
  `storage/pool-config.ts` to `shared/pool-config.ts` so both `PostgresStore` and
  `PgVector` share a single, tested implementation.
- Updated the `PgVector` constructor's connection-string branch to build its pool
  config through that helper, preserving the vector-specific `connectionTimeoutMillis`
  and `pgPoolOptions` behavior.
- Connection-string-only SSL (`?sslmode=require` with no explicit `ssl`) is
  unchanged.

### Before

```ts
import { PgVector } from '@mastra/pg';

// Throws UNABLE_TO_GET_ISSUER_CERT_LOCALLY — the explicit ssl is ignored
const vector = new PgVector({
  id: 'my-vector',
  connectionString: 'postgresql://user:pass@host:5432/db?sslmode=require',
  ssl: { rejectUnauthorized: false },
});
```

### After

```ts
import { PgVector } from '@mastra/pg';

// Connects — the explicit ssl option wins over sslmode= in the URL
const vector = new PgVector({
  id: 'my-vector',
  connectionString: 'postgresql://user:pass@host:5432/db?sslmode=require',
  ssl: { rejectUnauthorized: false },
});
```

## Tests

- Relocated the existing `buildConnectionStringPoolConfig` unit tests alongside the
  helper.
- Added DB-free tests mirroring the `PgVector` connection-string composition
  (explicit `ssl` precedence, URL-only `sslmode` preserved, `pgPoolOptions`
  overrides applied).

Run:

```bash
pnpm --filter @mastra/pg test src/shared/pool-config.test.ts
```

## Changeset

Added a `patch` changeset for `@mastra/pg`.

## ELI5

When you tell a database connection how to use security (SSL) in two different ways—once as a direct instruction and once hidden in the connection web address—the code was ignoring your direct instruction. This PR fixes that by making sure your explicit security option always wins over the hidden one in the web address.

## Summary

This PR fixes a bug where PgVector ignored an explicit SSL option when a connection string contained `sslmode=` or `ssl=` query parameters. The root cause was that node-postgres re-parses the connection string and overwrites explicit options with URL-derived ones.

### Changes

1. **Shared pool configuration helper**: Moved `buildConnectionStringPoolConfig` from `stores/pg/src/storage/pool-config.ts` to `stores/pg/src/shared/pool-config.ts` to establish a single, shared implementation for both PostgresStore and PgVector. This ensures consistent SSL handling across both code paths.

2. **Updated import in PostgresStore**: Modified `stores/pg/src/storage/index.ts` to import `buildConnectionStringPoolConfig` from the new shared location instead of the local storage module.

3. **Updated PgVector connection-string handling**: Modified `stores/pg/src/vector/index.ts` to import and use `buildConnectionStringPoolConfig` when constructing pools for connection-string configurations. This ensures explicit SSL options take precedence over `sslmode=`/`ssl=` parameters in the connection string, while preserving connection-string-only SSL behavior when no explicit SSL option is provided.

4. **Documentation**: Added clarifying comments and a changeset note documenting the SSL precedence behavior shared between PostgresStore and PgVector, with examples showing the corrected behavior for self-signed/internal CA scenarios.

### Testing

- Unit tests for `buildConnectionStringPoolConfig` are co-located with the helper in the shared location
- Added DB-free tests for PgVector connection-string behavior covering: explicit SSL precedence, URL-only sslmode preservation, and pgPoolOptions overrides

This brings PgVector's SSL handling into parity with PostgresStore, fixing certificate verification errors (e.g., UNABLE_TO_GET_ISSUER_CERT_LOCALLY) that occurred with self-signed or internal CAs.


abhiaiyer91 (approved):

Abuhaithem:
@abhiaiyer91 I couldn't merge it because there are some pending checks, and 5 workflows are still awaiting approval.

PR #17656, opened 2026-06-07
fix(schema-compat): apply `.default()` / `.optional()` semantics when models return `null`

### LINKED ISSUE #17655: [BUG] Tool-call & structured-output validation fails for `.default()` / `.optional()` fields when the model returns `null` (most providers)
### Describe the Bug

When a tool's `inputSchema` (or an agent's `structuredOutput` / `experimental_output` schema) contains a Zod field declared with `.default()` or `.optional()`, validation of the model's response fails with:

```
expected <type>, received null
```

In strict structured-output / tool-calling mode, providers return `null` for a field the model decided not to fill, rather than omitting the key. `@mastra/schema-compat` doesn't consistently translate that `null` back into the "absent" value Zod expects, so the original schema rejects it and the tool call / structured output fails.

The breakage depends on the provider and the modifier:

| Field modifier | OpenAI | Google (Gemini) | Anthropic | DeepSeek | Meta |
| --- | :---: | :---: | :---: | :---: | :---: |
| `.default(x)` | ❌ | ❌ | ❌ | ❌ | ❌ |
| `.optional()` | ✅ | ❌ | ❌ | ❌ | ❌ |

`.default()` is broken on **every** provider; `.optional()` is broken on every provider **except** OpenAI.

**Why it matters:** `z.string().default(...)` and `z.number().optional()` are extremely common in tool parameter definitions. When the model legitimately omits such a field, the tool call throws during result validation instead of falling back to the default / `undefined`, so the agent "randomly" fails on valid tool calls and declared defaults are never applied.

**Root cause:**
1. `processToAISDKSchema` builds the model-facing schema from Zod's *output* projection (`standardSchemaToJSONSchema(compat)` defaults to `io: 'output'`), where `.default()` fields are `required`/non-nullable and never tagged `x-optional` — contradicting the documented intent of `processToJSONSchema`/`toJSONSchema` (*"Use input mode so fields with defaults are optional"*).
2. Only the OpenAI compat layer's `#traverse` converts the strict-mode `null` back to `undefined`; Google, Anthropic, DeepSeek and Meta never do.
3. The null-handling tests live only in `createOpenAISuite`, so the gap was invisible for the other providers.

### Steps To Reproduce

```
1. In a project with `@mastra/schema-compat` and `zod` installed, add the file below.
2. Run it with: `npx tsx repro.ts`
3. See the validation fail with "expected number, received null".
```

```ts
// repro.ts
import { GoogleSchemaCompatLayer } from '@mastra/schema-compat';
import { z } from 'zod';

const schema = z.object({
  query: z.string(),
  limit: z.number().default(10),
});

const model = {
  provider: 'google',
  modelId: 'gemini-2.0-flash',
  modelVersion: 'v2',
  supportsStructuredOutputs: true,
};

const aiSchema = new GoogleSchemaCompatLayer(model).processToAISDKSchema(schema);

// In strict mode the model returns `null` for the omitted, defaulted field:
console.log(aiSchema.validate({ query: 'cats', limit: null }));
// → { success: false, error: ZodError: expected number, received null }
// Expected: { success: true, value: { query: 'cats', limit: 10 } }
```

End-to-end, this fires for any agent tool / structured output that uses defaults or optionals, because `CoreToolBuilder` and `MastraLLM` build the AI SDK schema via `applyCompatLayer({ mode: 'aiSdkSchema' })` → `processToAISDKSchema`.

https://github.com/mastra-ai/mastra/blob/9d7188b6c3251eab62f5f568e2f24aee76f42a59/packages/schema-compat/src/provider-compats/google.ts#L247

### Expected Behavior

For all providers, a `null` returned for a non-required (optional or defaulted) field should be treated as "absent" so Zod applies `.optional()` / `.default()` semantics:

- `.optional()` field → value becomes `undefined` and validation passes.
- `.default(x)` field → value falls back to the declared default `x`.
- `.nullable()` (required) field → explicit `null` is preserved.

So `aiSchema.validate({ query: 'cats', limit: null })` should return `{ success: true, value: { query: 'cats', limit: 10 } }`.

### Environment Information

```block
Affected package: @mastra/schema-compat (impacts @mastra/core tool calling & structured output)
Reproduced against the OpenAI, Google, Anthropic, DeepSeek and Meta compat layers.
```

### Verification

Abuhaithem:
Fixes #17655

## Problem

In strict structured-output / tool-calling mode, providers return `null` for a non-required field
the model chose not to fill. Mastra’s schema-compat layer didn’t consistently translate that `null`
back into the “absent” value Zod expects, so validation threw `expected <type>, received null`.

- `.default()` fields broke on **every** provider.
- `.optional()` fields broke on every provider **except OpenAI**.

Two root causes:

1. `processToAISDKSchema` built the model-facing schema from Zod’s **output** projection
   (`standardSchemaToJSONSchema(compat)` defaults to `io: 'output'`), where `.default()` fields are
   required and not tagged `x-optional`. This contradicts the documented intent of
   `processToJSONSchema` / `toJSONSchema` (*“Use input mode so fields with defaults are optional”*).
2. Only OpenAI’s `#traverse` normalized the strict-mode `null` back to `undefined`.

The null-handling tests lived in `createOpenAISuite` (OpenAI only), so the gap was invisible for the
other providers.

## Fix

- **Use the input projection** for AI-facing schemas: pass `{ io: 'input' }` to
  `standardSchemaToJSONSchema` in every provider `processToAISDKSchema` and in
  `applyCompatLayer`’s `jsonSchema` branch. Tool *output* schemas keep `io: 'output'`.
- **Add a shared `convertOptionalNullsToUndefined` helper** that converts `null → undefined` for
  optional fields, honoring both conventions — `x-optional` (OpenAI strict) and absence from
  `required` (standard JSON Schema) — while preserving explicit `null` for `.nullable()` fields. It
  resolves nullable `anyOf` wrappers and recurses into arrays/nested objects.
- **Wire the helper** into Google, Anthropic, DeepSeek and Meta validation (after the existing
  date-converting `#traverse`). OpenAI already normalizes inline and only needed the projection fix.
- **Close the coverage gap**: move the null-handling round-trip into the universal `createSuite`
  (runs for all providers) and add unit tests for the helper.

## Before / after

```ts
const schema = z.object({ query: z.string(), limit: z.number().default(10) });
const aiSchema = new GoogleSchemaCompatLayer(model).processToAISDKSchema(schema);

aiSchema.validate({ query: 'cats', limit: null });
// before → { success: false, error: "expected number, received null" }
// after  → { success: true, value: { query: 'cats', limit: 10 } }
```

## Provider matrix (manual repro, `{ field: null }` → validate)

| | optional | default | nested default | array default | nullable keeps null |
| --- | :---: | :---: | :---: | :---: | :---: |
| OpenAI | ✅ | ✅ | ✅ | ✅ | ✅ |
| Google | ✅ | ✅ | ✅ | ✅ | ✅ |
| Anthropic | ✅ | ✅ | ✅ | ✅ | ✅ |
| DeepSeek | ✅ | ✅ | ✅ | ✅ | ✅ |
| Meta | ✅ | ✅ | ✅ | ✅ | ✅ |

(Every cell was ❌ for `.default()` and ❌ for non-OpenAI `.optional()` before this PR.)

## Tests

- `pnpm --filter @mastra/schema-compat test` → **863 passed** (+61 new), 9 skipped.
- New shared block in `createSuite` exercises optional/default/nullable null handling for all providers.
- New unit tests for `convertOptionalNullsToUndefined` (both optionality conventions, nullable
  preservation, anyOf resolution, array/nested recursion).
- Lint + typecheck clean.

## Notes for reviewers

- Behavior-only change; no public API additions beyond exporting the new helper.
- Downstream `@mastra/core`: model-facing JSON Schema for `.default()` fields now reads as optional
  rather than required (strictly more correct). Snapshot tests that capture generated tool schemas
  with defaults may need regeneration — no runtime behavior regresses.

## Changeset

`.changeset/schema-compat-optional-default-null.md` (`@mastra/schema-compat`, patch).

## ELI5

When AI models are asked to fill in structured forms (like tool inputs), sometimes they skip optional fields by sending back `null` to mean "I didn't provide this." The validation system was treating that `null` as a bad value instead of recognizing it as "missing," so default values weren't being applied. This PR teaches the system to understand that `null` from the model means "this field is missing" for optional or defaulted fields, so defaults work properly.

---

## Overview

This PR fixes cross-provider validation failures in `@mastra/schema-compat` when AI models return `null` for non-required fields in strict structured-output and tool-calling scenarios. The fix consists of two main strategies: (1) using the input projection when generating AI-facing JSON schemas so that `.default()` and optional fields are properly recognized, and (2) introducing a shared utility to convert model-returned `null` to `undefined` for optional/default fields before Zod validation, allowing `.default()` and `.optional()` semantics to work correctly.

## Changes by File

**packages/schema-compat/src/index.ts:**
- Exports new utility `convertOptionalNullsToUndefined` alongside existing `wrapSchemaWithNullTransform`

**packages/schema-compat/src/null-to-undefined.ts:**
- Introduces `resolveNullableVariant` helper to unwrap nullable `anyOf` schemas
- Adds new exported `convertOptionalNullsToUndefined` function that mutates objects/arrays by converting `null` to `undefined` for fields treated as optional (identified via `required` array absence or `x-optional` marker), while preserving `null` for explicitly required or `.nullable()` fields
- Preserves existing `wrapSchemaWithNullTransform` function

**packages/schema-compat/src/provider-compats/openai.ts:**
- Updates `processToAISDKSchema` to explicitly use input projection (`{ io: 'input' }`) when generating the JSON schema
- Ensures `.default()` fields are treated as optional in the schema sent to the model

**packages/schema-compat/src/provider-compats/google.ts:**
- Uses input projection in `processToAISDKSchema`
- Applies `convertOptionalNullsToUndefined` during `jsonSchema` validation to normalize `null` values back to `undefined` before Zod parsing

**packages/schema-compat/src/provider-compats/anthropic.ts:**
- Uses input projection in `processToAISDKSchema`
- Integrates `convertOptionalNullsToUndefined` alongside date-handling traversal during validation

**packages/schema-compat/src/provider-compats/deepseek.ts:**
- Uses input projection and applies `convertOptionalNullsToUndefined` during JSON schema validation

**packages/schema-compat/src/provider-compats/meta.ts:**
- Uses input projection in `processToAISDKSchema`
- Applies `convertOptionalNullsToUndefined` after traversal but before Zod validation

**packages/schema-compat/src/utils.ts:**
- Updates `applyCompatLayer` to use input projection when calling `standardSchemaToJSONSchema` in `jsonSchema` mode for both primary compat layer and fallback paths

**packages/schema-compat/src/provider-compats/test-suite.ts:**
- Adds comprehensive test coverage for null-handling across all providers via a new test block in `createSuite`
- Tests verify: `null` is treated as `undefined` for optional fields, `.default()` values apply when model returns `null`, `null` is preserved for required nullable fields, nested object and array defaults work correctly, and provided non-null values aren't overwritten

**.changeset/schema-compat-optional-default-null.md:**
- Adds changeset documentation describing the fix and before/after behavior with a TypeScript example

## Results

- Schema-compat test suite: 863 passed (+61), 9 skipped
- All providers (OpenAI, Google, Anthropic, DeepSeek, Meta) now consistently handle optional/default fields when models return `null`
- Linting and type checking pass

### DIFF
--- a/.changeset/schema-compat-optional-default-null.md
+++ b/.changeset/schema-compat-optional-default-null.md
@@ -4,9 +4,14 @@
 
 Fixed tool-call and structured-output validation failing with "expected <type>, received null" when a tool or output schema used `.default()` or `.optional()` fields and the model left them out.
 
-In strict structured-output / tool-calling mode, providers return `null` for a non-required field instead of omitting it. The AI SDK schema produced by `processToAISDKSchema` was built from Zod's *output* projection, where `.default()` fields are required, and only the OpenAI compat layer normalized the returned `null` back to `undefined`. As a result `.default()` fields broke on every provider, and `.optional()` fields broke on Google, Anthropic, DeepSeek, and Meta.
+In strict structured-output and tool-calling mode, validation now handles missing optional and defaulted fields consistently across providers.
 
-The schema is now built from the *input* projection (so defaulted fields are optional, matching what the model is asked to produce), and a shared `convertOptionalNullsToUndefined` helper normalizes `null` → `undefined` for optional/defaulted fields across all providers, while preserving explicit `null` for `.nullable()` fields.
+Fixed behavior:
+- `.optional()` fields now treat model-returned `null` as missing input.
+- `.default()` fields now apply fallback values when models return `null`.
+- `.nullable()` fields still preserve explicit `null`.
+
+This fix applies to OpenAI, Google, Anthropic, DeepSeek, and Meta.
 
 ```ts
 // Before: this tool failed when the model omitted `limit`
--- a/.changeset/schema-compat-optional-default-null.md
+++ b/.changeset/schema-compat-optional-default-null.md
@@ -1,0 +1,23 @@
+---
+'@mastra/schema-compat': patch
+---
+
+Fixed tool-call and structured-output validation failing with "expected <type>, received null" when a tool or output schema used `.default()` or `.optional()` fields and the model left them out.
+
+In strict structured-output / tool-calling mode, providers return `null` for a non-required field instead of omitting it. The AI SDK schema produced by `processToAISDKSchema` was built from Zod's *output* projection, where `.default()` fields are required, and only the OpenAI compat layer normalized the returned `null` back to `undefined`. As a result `.default()` fields broke on every provider, and `.optional()` fields broke on Google, Anthropic, DeepSeek, and Meta.
+
+The schema is now built from the *input* projection (so defaulted fields are optional, matching what the model is asked to produce), and a shared `convertOptionalNullsToUndefined` helper normalizes `null` → `undefined` for optional/defaulted fields across all providers, while preserving explicit `null` for `.nullable()` fields.
+
+```ts
+// Before: this tool failed when the model omitted `limit`
+const search = createTool({
+  id: 'search',
+  inputSchema: z.object({
+    query: z.string(),
+    limit: z.number().default(10), // model returns null → "expected number, received null"
+  }),
+  // ...
+});
+
+// After: `limit` correctly falls back to 10
+```
--- a/packages/schema-compat/src/index.ts
+++ b/packages/schema-compat/src/index.ts
@@ -30,7 +30,7 @@ export { SchemaCompatLayer } from './schema-compatibility';
 
 // Utility functions
 export { convertZodSchemaToAISDKSchema, applyCompatLayer, convertSchemaToZod, isZodType } from './utils';
-export { wrapSchemaWithNullTransform } from './null-to-undefined';
+export { wrapSchemaWithNullTransform, convertOptionalNullsToUndefined } from './null-to-undefined';
 export { ensureAllPropertiesRequired, prepareJsonSchemaForOpenAIStrictMode } from './zod-to-json';
 
 // Standard Schema compatibility utilities
--- a/packages/schema-compat/src/null-to-undefined.ts
+++ b/packages/schema-compat/src/null-to-undefined.ts
@@ -46,6 +46,87 @@ export function transformNullToUndefined(value: unknown, jsonSchema: Record<stri
 }
 
 /**
+ * Resolve a nullable wrapper (`anyOf: [<schema>, { type: 'null' }]`) down to its
+ * non-null variant, matching the resolution the provider `#traverse` helpers do.
+ */
+function resolveNullableVariant(schema: Record<string, unknown>): Record<string, unknown> {
+  if (Array.isArray(schema.anyOf)) {
+    const nonNull = (schema.anyOf as unknown[]).find(
+      (s): s is Record<string, unknown> =>
+        !!s && typeof s === 'object' && (s as Record<string, unknown>).type !== 'null',
+    );
+    if (nonNull) {
+      return nonNull;
+    }
+  }
+  return schema;
+}
+
+/**
+ * Recursively convert `null` → `undefined` for the optional fields of a
+ * processed schema.
+ *
+ * Providers return `null` for an "absent" optional/defaulted field, but the
+ * different compat layers express optionality two ways:
+ *
+ * - **Standard JSON Schema** (Anthropic, Google, DeepSeek, Meta): the field is
+ *   simply omitted from the object's `required` array.
+ * - **OpenAI strict mode**: every property must appear in `required`, so the
+ *   originally optional / defaulted fields are instead tracked in `x-optional`.
+ *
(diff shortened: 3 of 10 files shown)


wardpeet (commented):
I don't belive this is the correct fix, waiting your answer on the issue. We should only replace it if the model made a mistake which you don't really know with your approach

wardpeet:
Closing this is not the correct fix, it needs to be captured differently

PR #17660, opened 2026-06-08
fix(core): don't downgrade settled tool-invocation parts in updateToolInvocation

### LINKED ISSUE #17659: [BUG] Background-dispatched tool calls are stripped from the model prompt and re-dispatched (duplicate sub-agents)
### Describe the Bug

When a tool is dispatched as a **background task** (`backgroundTasks` + agent-as-tool delegation), the model frequently **re-dispatches the same task**, spawning a second sub-agent instance for the same work.

The root cause is a destructive interaction between two behaviors that are each fine in isolation:

1. **`onExecution` destructively downgrades the tool-invocation.** After dispatch, the initial turn writes a placeholder result (`state:'result'` / `output-available`) so the model can see "task dispatched, running in background". But the async `onExecution` hook in `packages/core/src/loop/workflows/agentic-execution/tool-call-step.ts` (~L1061) then calls `updateToolInvocation` with `state:'call'` purely to attach `startedAt`/`taskId` metadata. Because `updateToolInvocation` replaces the part wholesale (`packages/core/src/agent/message-list/message-list.ts` ~L1098), it overwrites the part back to `input-available` and **drops the placeholder result**:

   ```ts
   msg.content.parts[i] = {
     ...inputPart,                      // state:'call', no result
     toolInvocation: { ...inputPart.toolInvocation, args: part.toolInvocation.args },
     ...
   };
   ```

2. **The incomplete-tool-call filter then strips it from the prompt.** When building the request to the model (`get.all.aiV5.prompt()`, which runs `sanitizeV5UIMessages` with `filterIncompleteToolCalls=true` by default), `input-available` client-side tool calls are dropped to preserve the provider tool-call/tool-result invariant — `packages/core/src/agent/message-list/conversion/output-converter.ts` (~L173):

   ```ts
   if (filterIncompleteToolCalls) {
     if (p.state === 'output-available' || p.state === 'output-error') return true;
     if (p.state === 'input-available' && p.providerExecuted && assistantTurnStillOpen) return true;
     return false; // input-available client-side tool call → dropped
   }
   ```
   A dispatched background tool call is a legitimate "pending" call, but here it is indistinguishable from an orphaned/aborted call, so it gets dropped.

**Net effect:** the just-dispatched delegation disappears from the assistant turn the model sees (it becomes a reasoning-only "shell" with no `tool_calls` and no paired tool result), so the model thinks it never delegated and delegates again — starting a duplicate sub-agent for the same task.

This is intermittent because it is a race: `onExecution` runs asynchronously (when the worker picks up the task). If it fires before the next request is built, the placeholder is already clobbered and the call is stripped; if the worker is queued behind another task, `onExecution` fires later and the placeholder survives that request. In one observed run the two delegations differed only by `onExecution` timing (one fired ~330ms before the next request → re-dispatched; the other ~13s after → survived).

### Steps To Reproduce

This reproduces in a single `agent.stream()` agentic loop (no `streamUntilIdle` required):

1. Configure an agent with a sub-agent exposed as a tool and `backgroundTasks` enabled, so the delegation tool runs in the background.
2. Ask it to delegate a task to the sub-agent.
3. Observe the request sent to the model on the **next** step: the assistant turn that delegated is now a reasoning-only shell — the delegation `tool_calls` entry and its tool result are both gone.
4. The model delegates again; two sub-agent instances claim the same task.

Both delegations end up persisted in storage as `state=input-available` with **no output**, while synchronous tools in the same stream (e.g. `skill`, `createTask`) are correctly `output-available` and survive.

A colocated unit test reproduces it deterministically (`packages/core/src/agent/message-list/tests/background-task-tool-invocation-stripping.test.ts`). It asserts the desired behavior, so it currently **fails**:

```
× onExecution must not downgrade a dispatched background call away from its placeholder result
  AssertionError: expected 'call' to be 'result'

× a dispatched background delegation stays visible to the model after onExecution fires
  expected '[...]' to contain 'call_RY9990'
  Received: [{"role":"user",...},
             {"role":"assistant","content":[{"type":"reasoning","text":"delegating to biExecutorAgent ..."}]}]
             // ^ reasoning-only shell, no tool_calls — exactly the production symptom
```

1

### Expected Behavior

A dispatched (running or suspended) background tool call should stay visible to the model across turns — with its placeholder result — so the model does not re-dispatch it.

Concretely:

- `onExecution` should only merge `backgroundTasks` metadata (`startedAt`/`suspendedAt`/`taskId`) and must **not** downgrade an existing `output-available`/result part back to `input-available`. (Either pass `state:'result'` preserving the placeholder, or add a "don't downgrade a part that already has a result" guard in `updateToolInvocation`.)
- Separately (design gap, can be a follow-up): `sanitizeV5UIMessages` could recognize that an `input-available` client tool carrying `backgroundTasks` metadata is intentionally pending, and keep it (synthesizing a placeholder tool result) instead of dropping it like an orphan.

### Environment Information

```block
- `@mastra/core`: 1.40.x (also reproduces on 1.38–1.40)
- Feature: `backgroundTasks` + agent-as-tool delegation
- Provider: provider-agnostic (first observed with DeepSeek; the stripping happens in core message conversion, not in any provider)
```

### Verification

Aayush-engineer:
## Description
When a tool is dispatched as a background task, `onExecution` calls `updateToolInvocation` with `state: 'call'` to attach `startedAt`/`taskId` metadata. Because `updateToolInvocation` replaced the part wholesale, it overwrote an already-settled `output-available` placeholder back to `input-available`, dropping the result.

`sanitizeV5UIMessages` then correctly filtered out the `input-available` client-side call (to preserve the provider tool-call/tool-result invariant), making the dispatched delegation invisible to the model — which then re-dispatched the same task, spawning a duplicate sub-agent.

**Fix:** Added an `isAlreadySettled` guard in `updateToolInvocation` — if the existing part is already `result`, `output-available`, `output-denied`, or `output-error`, the part is not replaced. Only the `backgroundTasks` metadata is merged (which was already handled correctly below the replacement block).

The two root causes described in the issue are actually one cascading failure: fixing the downgrade keeps the part as `output-available`, which `sanitizeV5UIMessages` correctly preserves, so the model sees the placeholder result and does not re-dispatch.

## Related issue(s)
Fixes #17659

## Type of change

## Checklist

## ELI5

This PR makes sure that once a background tool call has produced (or is already safely “done enough” to use) its output placeholder, we don’t accidentally overwrite it with an older “still pending” version. That way, the model keeps seeing the task and doesn’t try to run it again.

## Problem

When a tool is dispatched as a background task, `onExecution` later calls `updateToolInvocation` to attach metadata like `startedAt`/`taskId` (and related fields). Previously, `updateToolInvocation` could overwrite an already-settled tool-invocation part (e.g. `result` / `output-available`), downgrading it back to an “input-available”-style placeholder. Then `sanitizeV5UIMessages` could remove that downgraded call from the prompt, and the model could re-dispatch the same background task.

## Solution

`MessageList.updateToolInvocation` now adds an `isAlreadySettled` guard:

- If the existing `toolInvocation.state` on the matching part is already settled (`result`, `output-available`, `output-denied`, `output-error`), it **skips replacing** that part—preventing the downgrade.
- If not settled, it updates the part as before.
- Even when skipping the destructive overwrite, it **still merges** incoming `backgroundTasks` metadata (and preserves applicable runtime/provider fields like `providerExecuted` and merged `providerMetadata` when applicable).

## Changes

- `packages/core/src/agent/message-list/message-list.ts`
  - Updated `updateToolInvocation` to avoid clobbering already-settled tool-invocation parts while still merging `backgroundTasks` metadata.
  - Lines changed: **+24 / -12**
- `.changeset/cold-shoes-read.md`
  - Adds a **patch** changeset for `@mastra/core` documenting the behavior change.
- Tests
  - Added coverage in `update-tool-invocation.test.ts` to verify:
    - settled `result` / `output-available` parts are not downgraded
    - `backgroundTasks` metadata still merges on settled parts
    - non-settled parts continue updating normally

## Effect

Prevents a “downgrade cascade” so background-dispatched tool-call placeholders remain visible to `sanitizeV5UIMessages`, avoiding duplicate re-dispatch/re-start behavior.

## Related issue

Fixes **`#17659`**

### DIFF
--- a/.changeset/cold-shoes-read.md
+++ b/.changeset/cold-shoes-read.md
@@ -1,0 +1,5 @@
+---
+'@mastra/core': patch
+---
+
+Don't downgrade settled tool-invocation parts in updateToolInvocation, preventing duplicate background task dispatch
--- a/packages/core/src/agent/message-list/message-list.ts
+++ b/packages/core/src/agent/message-list/message-list.ts
@@ -1138,7 +1138,7 @@ export class MessageList {
           const existingState = part.toolInvocation?.state;
           const isAlreadySettled =
             existingState === 'result' ||
-            existingState === 'output-available' ||
+            (existingState as string) === 'output-available' ||
             existingState === 'output-denied' ||
             existingState === 'output-error';
           if (!isAlreadySettled) {
--- a/packages/core/src/agent/message-list/message-list.ts
+++ b/packages/core/src/agent/message-list/message-list.ts
@@ -1131,18 +1131,30 @@ export class MessageList {
                 } as AIV5Type.ProviderMetadata)
               : undefined;
 
-          msg.content.parts[i] = {
-            ...inputPart,
-            toolInvocation: {
-              ...inputPart.toolInvocation,
-              args: part.toolInvocation.args,
-            },
-            // Preserve providerExecuted from original call if not in result
-            ...(originalPart.providerExecuted !== undefined && inputPartWithMeta.providerExecuted === undefined
-              ? { providerExecuted: originalPart.providerExecuted }
-              : {}),
-            ...(mergedProviderMetadata !== undefined ? { providerMetadata: mergedProviderMetadata } : {}),
-          };
+          // Do not downgrade a part that already has a result — only merge metadata.
+          // This guards against onExecution (which always passes state:'call') clobbering
+          // an output-available placeholder written by a background-dispatched tool call,
+          // which would make the call disappear from the model prompt and cause re-dispatch.
+          const existingState = part.toolInvocation?.state;
+          const isAlreadySettled =
+            existingState === 'result' ||
+            existingState === 'output-available' ||
+            existingState === 'output-denied' ||
+            existingState === 'output-error';
+          if (!isAlreadySettled) {
+            msg.content.parts[i] = {
+              ...inputPart,
+              toolInvocation: {
+                ...inputPart.toolInvocation,
+                args: part.toolInvocation.args,
+              },
+              // Preserve providerExecuted from original call if not in result
+              ...(originalPart.providerExecuted !== undefined && inputPartWithMeta.providerExecuted === undefined
+                ? { providerExecuted: originalPart.providerExecuted }
+                : {}),
+              ...(mergedProviderMetadata !== undefined ? { providerMetadata: mergedProviderMetadata } : {}),
+            };
+          }
           this.lastCreatedAt = Math.max(this.lastCreatedAt || 0, Date.now());
           this.updateLastCreatedAt(msg);
 
--- a/packages/core/src/agent/message-list/tests/update-tool-invocation.test.ts
+++ b/packages/core/src/agent/message-list/tests/update-tool-invocation.test.ts
@@ -703,4 +703,272 @@ describe('MessageList.updateToolInvocation', () => {
       mastra: { modelOutput: true },
     });
   });
+
+  describe('isAlreadySettled guard (background task fix)', () => {
+    it('should not downgrade a result part back to call when onExecution fires', () => {
+      const messageList = new MessageList();
+
+      const msg = makeAssistantMessage([
+        {
+          type: 'tool-invocation',
+          toolInvocation: {
+            state: 'result',
+            toolCallId: 'tc-bg',
+            toolName: 'delegate_agent',
+            args: { task: 'summarise' },
+            result: { output: 'task dispatched' },
+          },
+        },
+      ]);
+      messageList.add(msg, 'response');
+
+      // onExecution always passes state:'call' to attach metadata — must not downgrade
+      messageList.updateToolInvocation({
+        type: 'tool-invocation',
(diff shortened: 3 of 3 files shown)


Aayush-engineer:
@TylerBarnes @CalebBarnes ready for review when you get a chance! Gentle ping after 2 weeks — CodeRabbit has no actionable comments, all checks pass. Happy to rebase on main or make any changes needed.

Aayush-engineer:
Rebased onto latest main and added 4 tests covering the isAlreadySettled guard directly in update-tool-invocation.test.ts — verifies that result and output-available parts are not downgraded back to call, that backgroundTasks metadata still merges correctly on settled parts, and that non-settled parts still update normally. All 22 tests passing.

Aayush-engineer:
Closing — the same fix was independently implemented and merged in #18556. The isAlreadySettled guard approach and the metadata-only update path both solve the same root cause.

PR #17661, opened 2026-06-08
feat(core): add workspace cleanup API and shutdown cascade

### LINKED ISSUE #16748: [FEATURE] Proper builtin sandbox workspaces cleanup
### Problem Statement

The current Mastra server doesn't do anything to cleanup sandbox workspaces, so they just accumulate on the server until they remotely time out, or just exist locally forever.

This requires users of Mastra to wire something themselves without this caveat being mentioned anywhere in Mastra's docs. So this missing feature, as it causes a resource leak, can also be considered a bug.

### Proposed Solution

Add a builtin configurable policy for cleaning up workspaces as part of the Mastra server and possibly to standalone agent classes. e.g. When the agent is not used for X time, or immediately. Might also want policy for just stop/resume when supported instead of full destroy/recreate.

It's important that this also runs cleanly on server shutdown (From a signal for example), so that sandboxes don't leak when 

### Component

Workspaces

### Alternatives Considered

Wire something up by myself that somehow tracks which workspaces are active and will need to decide when to stop/destroy them.

### Example Use Case

Just using workspaces normally, for any workspace that actually has a lifecycle.

### Additional Context

There are lifecycle methods on some sandboxes, but only `workspace.destroy()` calls them, which doesn't seem to be called by anything either...

### Verification

intojhanurag:
## Description

Mastra workspaces had no way to be cleaned up: there was no API to remove a workspace from the registry, `mastra.shutdown()` did not destroy registered workspaces, and the docs did not explain workspace lifecycle. This PR closes that gap so workspace resources (sandbox processes, filesystem handles, LSP, browser) are released on shutdown and through an explicit API.

- Add `mastra.removeWorkspace(id, { destroy? })` to remove a workspace from the registry, optionally destroying it first; on destroy failure the workspace stays registered so the caller can recover
- Cascade `mastra.shutdown()` to destroy all registered workspaces between `stopWorkers()` and `storage.close()`; best-effort, unregisters only successfully destroyed entries, failed ones remain visible for debugging
- Touch `workspace.lastAccessedAt` in `resolveEffectiveWorkspace()` so workspace tool execution updates activity, not just search operations
- Document workspace lifecycle: new `removeWorkspace` reference page, expanded `workspace.destroy()` section in `workspace-class.mdx`, new "Lifecycle and cleanup" section in workspace overview, sidebar entry, and updated `mastra.shutdown()` JSDoc
- Add focused tests in `packages/core/src/mastra/workspace-cleanup.test.ts` covering remove, destroy-failure rollback, global slot clearing, shutdown happy-path, and partial-failure shutdown; extend dynamic filesystem/sandbox tool tests to assert `lastAccessedAt` advances on tool use

## Related Issue(s)

Fixes #16748

## Type of Change

- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [ ] Test update

## Checklist

ELI5
Mastra makes it safe to clean up workspaces: you can remove a workspace (optionally destroying it first), shutdown now tries to destroy all registered workspaces, and Mastra records when a workspace was last used so idle-cleanup policies can work.

Overview
Adds workspace lifecycle and cleanup primitives, docs, and tests: mastra.removeWorkspace(id, { destroy? }), an enhanced mastra.shutdown() that best-effort destroys registered workspaces before closing storage/observability, and lastAccessedAt updates during tool execution so activity can drive idle policies. Focused tests and documentation verify behavior and this PR addresses resource-leak goals from issue `#16748` while leaving follow-ups scoped separately.

Key Changes

- New API: mastra.removeWorkspace(id, { destroy?: boolean })
  - New method Mastra.prototype.removeWorkspace(id, options?): Promise<boolean>.
  - If options.destroy is true, awaits workspace.destroy() before unregistering; if destroy() throws the call rejects and the workspace remains registered so callers can recover/debug.
  - Returns true when removed, false when ID not found.
  - Clears Mastra's global workspace slot if it referenced the removed workspace.

- mastra.shutdown() behavior
  - Shutdown enumerates registered workspace IDs and calls removeWorkspace(id, { destroy: true }) for each (via Promise.allSettled) between stopWorkers() and storage.close().
  - Workspace destruction is best-effort: per-workspace errors are logged and shutdown proceeds; only successfully destroyed workspaces are unregistered, failed ones remain registered for debugging.
  - Ensures workspace destruction runs before storage.close() and observability shutdown.

- Activity tracking
  - resolveEffectiveWorkspace() now updates workspace.lastAccessedAt whenever the effective workspace is resolved.
  - Tests updated to assert filesystem and sandbox tool executions advance lastAccessedAt, enabling idle-timeout/reaper policies to rely on tool usage.

- Workspace.destroy() robustness
  - destroy() made idempotent under concurrent calls by tracking an internal _destroyPromise and moving teardown logic into a private _performDestroy(), so overlapping destroy() calls await or short-circuit appropriately.

Documentation
- New reference doc: docs/src/content/en/reference/core/removeWorkspace.mdx (.removeWorkspace() usage, params, semantics).
- Added "Lifecycle and cleanup" section to workspace overview.
- Expanded Workspace.destroy() docs with shutdown/cleanup order and provider ownership notes.
- Added .removeWorkspace() to Core reference sidebar.
- Changeset: .changeset/workspace-cleanup-lifecycle.md

Tests
- New suite: packages/core/src/mastra/workspace-cleanup.test.ts verifying:
  - Successful removal/unregister and destroy:true path.
  - Rollback when destroy rejects (workspace remains registered).
  - Clearing the global workspace slot.
  - Shutdown happy-path and partial-failure shutdown behavior.
  - That multiple workspace destroy operations begin in parallel (concurrency).
- Tool tests updated to assert lastAccessedAt advances:
  - packages/core/src/workspace/tools/__tests__/dynamic-filesystem.test.ts
  - packages/core/src/workspace/tools/__tests__/dynamic-sandbox.test.ts

Files Modified (high level)
- packages/core/src/mastra/index.ts — add removeWorkspace() and enhanced shutdown logic (calls removeWorkspace for registered workspaces).
- packages/core/src/workspace/tools/tools.ts — update lastAccessedAt in resolveEffectiveWorkspace().
- packages/core/src/workspace/workspace.ts — make destroy() idempotent; add _performDestroy().
- packages/core/src/mastra/workspace-cleanup.test.ts — new tests.
- packages/core/src/workspace/tools/__tests__/dynamic-filesystem.test.ts — updated assertion.
- packages/core/src/workspace/tools/__tests__/dynamic-sandbox.test.ts — updated assertion.
- docs/... — new/updated docs and sidebar.
- .changeset/workspace-cleanup-lifecycle.md — release note.

Follow-ups (author notes)
1. Thin HTTP DELETE /workspaces/:workspaceId wrapper and client-js workspace.remove() in a follow-up PR.
2. Implement an opt-in idle-timeout reaper (workspaces: { idleTimeoutMs, checkIntervalMs }) that sweeps by lastAccessedAt and calls removeWorkspace(id, { destroy: true }). The author suggests keeping issue `#16748` open until the reaper ships.


intojhanurag:
Filing follow-ups so this PR stays scoped to primitives and #16748 can fully close:

1. **HTTP `DELETE /workspaces/:workspaceId`** : thin wrapper over `mastra.removeWorkspace()` + client-js `workspace.remove()` method (addresses @segevfiner's comment).
2. **Idle-timeout reaper** — opt-in `workspaces: { idleTimeoutMs, checkIntervalMs }` config that sweeps via `lastAccessedAt` and calls `removeWorkspace(id, { destroy: true })`; modeled after `background-tasks/manager.ts`. This is the headline ask of #16748.

Suggest keeping #16748 open until (2) ships. Workspace-level pause/resume left out, no concrete demand and provider-level `sandbox.stop()` already exists as an escape hatch.

packages/core/src/mastra/index.ts
```diff
@@ -4311,6 +4343,15 @@ export class Mastra<
...
     await this.stopWorkers();
+
+    for (const id of Object.keys(this.#workspaces)) {
+      try {
+        await this.removeWorkspace(id, { destroy: true });
+      } catch (error) {
+        this.#logger?.error('Failed to destroy workspace during shutdown', { workspaceId: id, error });
+      }
+    }
+
```
NikAiyer: I'm thinking we should switch to using `Promise.allSettled` for this to parallelize this cleanup.

NikAiyer (commented):
hey @intojhanurag, I think there are some changes we should add to this PR, and then some we should add as followups:

For this PR:
1. **Shutdown ordering test.** We should add a test to ensure we destroy registered workspaces before closing storage, and ensure we don't do that out of order. The changeset promises this behavior but nothing pins it — a regression that flipped the order would pass every current test. A `vi.fn()` + recorded-call-order assertion in `workspace-cleanup.test.ts` is enough.
2. **Parallel teardown.** I added a comment below, but I think we should change the teardown to be parallel. The current serial `for…of await` makes shutdown latency `N × slowest sandbox destroy`, and sandbox destroys (e2b/daytona) can take seconds. `Promise.allSettled(...)` parallelises cleanly and the "failed entries stay registered" semantic translates directly — rejected entries just don't get deleted, so the existing "continues after failure" test still passes

For a follow-up PR (before the reaper lands): 
1. **In-flight operation tracking on `Workspace`.** Today `removeWorkspace(id, { destroy: true })` will happily destroy a workspace mid-tool-call — the sandbox process gets killed under a running command and the tool throws garbage. The reaper makes this worse because eviction is automatic. Needs a refcount (or AsyncLocalStorage) wired through `wrapTool` + `wrapWithReadTracker`, plus a small helper for direct `workspace.*` access (search, skills, LSP, agent memory). `destroy()` should drain before tearing down LSP/browser/sandbox, with a timeout. Both the reaper and manual `removeWorkspace` should respect it (reaper skips and retries; manual either errors or accepts `force: true`). Two traps worth calling out: background processes started by `execute_command` legitimately outlive the tool call (track at tool-execute granularity, not process-lifetime), and the drain has to land before `clearSandboxCache()` and the LSP shutdown, not after. 
2. **Ownership/source check on `removeWorkspace`.** Once we're touching its contract for the above, we should define what happens for `source: 'agent'` entries. The discriminator already exists on the registry entry `removeWorkspace` just ignores it today. Probably: error by default, `force: true` to override. 
3. **Test for `agent.generate()` after the agent's workspace was removed.** 
(1) `removeWorkspace(agentWorkspaceId)` errors by default, 
(2) `removeWorkspace(agentWorkspaceId, { force: true })` succeeds and the subsequent `agent.generate()` fails with a sensible error rather than `Cannot read properties of undefined`. 
4. **`listWorkspaces()` enrichment.** Surface `lastAccessedAt` and a `status` field per entry so the reaper has something clean to consume and operators can introspect registry health.                                                                                                                                                       

For the reaper PR itself: 
1. **Lifecycle docs paragraph** covering "what happens to an agent/workflow whose workspace was removed" — user-facing story is the same as the reaper's eviction story.

NikAiyer (approved):
LGTM! But looks like there are some CI checks failing

intojhanurag:
> LGTM! But looks like there are some CI checks failing

I think that is unrelated to the PR

PR #17682, opened 2026-06-08
fix(playground): clear stale pending-signal badge in Studio chat

### LINKED ISSUE #17681: [BUG] Studio chat: 'pending' signal indicator lingers above the composer after the agent replies
### Describe the bug

In Studio agent chat, the `pending: …` shimmer indicator shown above the message input (the element with `data-testid="pending-signal-message"`) can **fail to clear**. After you send a follow-up message and the agent replies, the `pending: <your text>` badge stays above the composer. It can **stack across multiple messages**, and only disappears on a **page refresh**.

Notably, the sent message itself behaves correctly — it appears in the conversation as a normal user message and the agent replies to it. Only the `pending` badge is stranded.

Image

### Reproduction

1. Open an agent chat with memory enabled (thread signals are the default path).
2. Send a message; you briefly see `pending: <your text>` above the composer.
3. The agent replies and the message renders in history as expected.
4. The `pending: <your text>` badge **does not clear**.
5. Send more messages → multiple stale `pending: …` badges accumulate. A page refresh clears them (the state is ephemeral client-side React state).

### Expected behavior

The `pending` indicator should clear once the signal is acknowledged (echoed back), as it does in the normal path.

### Actual behavior

The `pending: …` indicator lingers indefinitely (until a page refresh) and can pile up.

### Root cause (analysis)

The pending badge is **added** via `onSignalSent` (fired after the send-message HTTP response resolves) and **removed** via `onSignalEcho` (fired from the thread-subscription stream when the `data-user-message` echo arrives) — two independent async channels with **no ordering guarantee**.

On the idle path the server starts the run and emits the `data-user-message` echo over the already-open subscription **before** the send-message HTTP response returns. So `onSignalEcho` runs **before** `onSignalSent` has added the pending entry: the remove is a no-op, and the late add then inserts a badge that no further echo will ever clear.

Relevant code:
- `packages/playground/src/services/mastra-runtime-provider.tsx` (`addPendingSignal` / `removePendingSignal` / `pendingSignals`)
- `client-sdks/react/src/agent/hooks.ts` (`onSignalSent` after `agent.sendMessage` resolves; `onSignalEcho` from `processStreamChunk` on `data-user-message`)

### Environment

- Area: UI / Dev Playground (Mastra Studio), agent-signals subscription chat path
- Reproduced locally in dev Studio against a real agent.

### Additional context

A fix is on the way (provider-side, order-independent bookkeeping). A more thorough, longer-term fix for related edges (a narrow reset-vs-late-send re-race and unbounded growth of the premature-echo tracking from foreign/duplicate echoes on a shared thread) likely belongs at the `@mastra/react` `useChat` SDK boundary, where the ordering hazard originates.

aisensiy:
Closes #17681

## Summary

Studio chat's `pending: …` indicator above the message input could linger after the agent finished replying — and pile up across messages — clearing only on a page refresh.

## Root cause

The badge is **added** via `onSignalSent` (the send-message HTTP response) and **removed** via `onSignalEcho` (the thread-subscription stream) — two async channels with no ordering guarantee. On the idle path the server starts the run and emits the `data-user-message` echo over the already-open subscription **before** the send-message response returns, so `onSignalEcho` fired before `onSignalSent` had added the pending entry: the remove ran as a no-op, then the late add orphaned a badge that no further echo would ever clear. (The echo itself arrives fine — the user message renders in history — only the badge is stranded.)

## Fix

Make the pending bookkeeping order-independent using two synchronous id mirrors, so add/remove are commutative regardless of arrival order:

- `pendingSignalIdsRef` — mirrors the currently-visible badge ids
- `prematurelyEchoedSignalIdsRef` — remembers echoes that landed before their add, so the late add is skipped instead of resurrecting a badge

Both are routed through a `clearPendingSignals()` helper at the three reset sites (thread/agent switch, `onThreadSignalsUnsupported`, `onCancel`).

## Testing

- New **deterministic** unit tests force the echo-before-add ordering and assert the badge clears (fails before the fix, passes after), plus normal-order and per-id independence cases.
- **Validated live** in a dev Studio against a real agent: the lingering `pending` no longer appears.
- `pnpm --filter ./packages/playground test src/services` → 54/54, plus typecheck and eslint pass.

## Review follow-ups included here

- Add `afterEach(cleanup)` to the provider test — this package runs vitest with globals disabled, so React Testing Library's auto-cleanup never registers and rendered providers were leaking between tests; also hardened `getSignalCallbacks`.
- Reuse the exported `PendingSignalMessage` type instead of an inline literal.

## Known follow-ups (not in this PR)

A fully-robust fix for two related, lower-severity edges — a narrow reset-vs-late-send re-race, and unbounded growth of the premature-echo set from foreign/duplicate echoes on a shared thread — belongs at the `@mastra/react` `useChat` SDK boundary, where the ordering hazard originates (both callbacks fire from inside the hook). Tracking separately rather than band-aiding per consumer.

## ELI5

When you send a message in Studio chat, a "pending: ..." badge appears while the message is being processed. This badge *should* disappear once the server confirms the message was received, but sometimes it sticks around and won't go away even after you get a reply—you have to refresh the page to clear it. This happens because the confirmation and the acknowledgment can arrive out of order, and the code only knew how to clean up if they arrived in the right order.

## Summary

This PR fixes a race condition in Studio's chat pending signal indicator where the `pending: …` badge above the message input could linger after the agent replied and accumulate across messages.

**Root cause:** The pending badge is added when the send-message HTTP response arrives (`onSignalSent`) and removed when the echo is received from the thread-subscription stream (`onSignalEcho`). Since these two async channels have no ordering guarantee, when an echo arrives before the send response completes, the premature remove becomes a no-op. A subsequent add then leaves an orphaned badge that never clears.

**Solution:** The provider now maintains order-independent bookkeeping using two synchronous `Set` refs:
- `pendingSignalIdsRef`: tracks currently-visible pending badge IDs
- `prematurelyEchoedSignalIdsRef`: tracks echoes that arrived before their corresponding send confirmation

A new `clearPendingSignals()` helper centralizes cleanup logic and ensures both refs are cleared at reset points (thread/agent switch, losing signal support, cancelling a run). The logic skips late adds if an echo was already seen for that signal ID, making the add/remove operations commutative regardless of arrival order.

**Changes:**
- `mastra-runtime-provider.tsx`: Added the two ref-based tracking sets and `clearPendingSignals()` helper; updated thread/agent switch, signal-unsupported handler, and cancel handlers to use centralized cleanup
- `mastra-runtime-provider.test.tsx`: Added `cleanup()` between tests to prevent state leakage; added three new unit tests validating signal lifecycle under normal ordering, reverse ordering (echo-before-send), and per-ID independence
- `.changeset/studio-pending-signal-badge.md`: Documents the patch release for `@internal/playground`

**Testing:** Deterministic unit tests now force the problematic echo-before-add ordering (previously failed, now pass), plus normal-order and per-id independence cases; validated live in dev Studio; package tests and linters pass.

### DIFF
--- a/.changeset/studio-pending-signal-badge.md
+++ b/.changeset/studio-pending-signal-badge.md
@@ -1,0 +1,5 @@
+---
+'@internal/playground': patch
+---
+
+Fixed Studio chat leaving a stale "pending" signal indicator above the message input. Sending a follow-up message while the agent was idle could leave one or more shimmering "pending: …" badges that lingered after the reply finished (and piled up across messages), only disappearing on a page refresh. The badge now clears reliably regardless of whether the send confirmation or its echo arrives first.
--- a/packages/playground/src/services/__tests__/mastra-runtime-provider.test.tsx
+++ b/packages/playground/src/services/__tests__/mastra-runtime-provider.test.tsx
@@ -1,9 +1,9 @@
 // @vitest-environment jsdom
 import type { MastraDBMessage } from '@mastra/core/agent/message-list';
 import { useChat } from '@mastra/react';
-import { act, render, waitFor } from '@testing-library/react';
+import { act, cleanup, render, waitFor } from '@testing-library/react';
 import type { ReactNode } from 'react';
-import { beforeEach, describe, expect, it, vi } from 'vitest';
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
 
 const mocks = vi.hoisted(() => ({
   cancelRun: vi.fn(),
@@ -143,6 +143,12 @@ describe('MastraRuntimeProvider', () => {
     delete (window as any).MASTRA_AGENT_SIGNALS;
   });
 
+  // This package runs vitest with globals disabled, so @testing-library/react's
+  // auto-cleanup never registers. Unmount between tests explicitly, otherwise
+  // every render() leaks a mounted provider and getSignalCallbacks/threadRuntimeState
+  // (both last-writer-wins sinks) would read a stale instance.
+  afterEach(() => cleanup());
+
   it('opts Playground into thread signals by default', () => {
     render(
       <MastraRuntimeProvider agentId="agent-1" threadId="thread-1" initialMessages={[]} modelVersion="v2">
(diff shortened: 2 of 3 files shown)


TheIsrael1:
Hi @aisensiy your PR has some conflicts, kindly resolve them so the PR can be reviewed

aisensiy:
Closing as superseded. The `pending: …` signal badge this PR fixed was removed entirely by the inline-pending-bubble refactor (#17688 / #17774) — `mastra-runtime-provider.tsx` and the whole `pendingSignals` mechanism no longer exist on `main`, which is the source of the merge conflict here. The add-vs-echo ordering race is structurally gone in the new model: the user turn is appended synchronously as a single `status: 'pending'` bubble before the run starts and reconciled by `clientMessageId` when the `data-user-message` echo arrives, so the echo can't beat the add. Nothing left to rebase onto. Thanks @TheIsrael1 for the flag. Resolved via #17681.

PR #17684, opened 2026-06-08
feat(clickhouse): add opt-in replicated table support

### LINKED ISSUE #15618: [@mastra/clickhouse] Support Replicated* table engines for multi-replica ClickHouse clusters

NikAiyer:
## Summary

Adds opt-in replicated table support for ClickHouse-backed Mastra storage.

- Adds a top-level `replication` config for ClickHouse storage
- Rewrites Mastra-owned table DDL from MergeTree/ReplacingMergeTree to Replicated* engines when enabled
- Adds `ON CLUSTER` to Mastra-owned CREATE/ALTER/DROP DDL when a cluster is configured
- Fails fast when existing local tables are present under replication config to avoid mixed local/replicated storage
- Threads replication through generic ClickHouse domains and v-next observability
- Documents the new replication config in the ClickHouse README

## Related Issue(s)
Fixes https://github.com/mastra-ai/mastra/issues/15618

## Test plan

- `pnpm --filter @mastra/clickhouse lint`
- `pnpm --filter @mastra/clickhouse build:lib`
- `docker compose up -d --wait`
- `pnpm --filter @mastra/clickhouse exec vitest run src/storage/db/replication.test.ts src/storage/migration.test.ts src/storage/index.vnext.test.ts src/storage/index.test.ts`

Draft for comparison/review; changeset intentionally omitted for now.

## ELI5: What's This PR About?

This PR adds an opt-in setting so Mastra can safely create and manage ClickHouse tables across multiple replicas: it rewrites plain MergeTree engines to replicated variants, runs cluster-aware DDL, and prevents mixing non-replicated tables with a replication configuration.

---

## Overview

Adds opt-in replicated table support for ClickHouse-backed Mastra storage. When enabled, Mastra:
- Rewrites Mastra-owned table engines from MergeTree / ReplacingMergeTree to Replicated* variants,
- Injects ON CLUSTER '<cluster>' into Mastra-owned DDL (CREATE/ALTER/DROP/TRUNCATE/OPTIMIZE and supported SYSTEM operations) when a cluster is configured,
- Validates and fails fast if existing local (non-replicated) Mastra tables are present under a replication config to avoid mixed local/replicated deployments.

---

## Key Changes

Configuration & types
- New ClickhouseReplicationConfig: optional fields cluster?, zookeeperPath?, replicaName?.
- ClickhouseConfig gained optional replication?: ClickhouseReplicationConfig and ClickhouseReplicationConfig is re-exported.

Validation & utilities (stores/clickhouse/src/storage/db/replication.ts)
- validateReplicationConfig(): rejects empty/unsafe values (whitespace/quotes/backslashes).
- isReplicatedOrSharedEngine(): detects already-replicated/shared engines.
- buildReplicatedTableEngine(): rewrites MergeTree/ReplacingMergeTree to Replicated* with safe balanced-parentheses parsing and templated zookeeperPath/replicaName.
- addOnClusterToDDL(): idempotently injects ON CLUSTER '<cluster>' into supported DDL.
- applyReplicationToDDL(): applies engine rewrite then ON CLUSTER injection.
- buildLocalTableReplicationError(): constructs user-facing error when existing local tables conflict with replication config.

Database layer (ClickhouseDB and helpers)
- resolveClickhouseConfig and ClickhouseDB accept and store replication config; ClickhouseDB constructor updated.
- assertExistingTableCompatibleWithReplication(): queries system.tables and fails init/create when local engines conflict.
- createTable()/alterTable()/dropTable()/clearTable()/optimizeTable()/materializeTtl() now wrap SQL with applyReplicationToDDL/addOnClusterToDDL as appropriate.
- SYSTEM STOP/START MERGES remain non-clustered by design; truncate/drop/etc. use ON CLUSTER when configured.
- migrateSpansTableSortingKey (automatic span-table sorting-key migration) is blocked when replication is enabled and throws a MastraError instructing manual migration.

Domain wiring
- All ClickHouse domain storages (BackgroundTasks, Memory, Observability, Scores, Workflows, v-next observability) now pass replication from resolveClickhouseConfig into ClickhouseDB; v-next observability is replication-aware for bootstrapping, reconciliation, and operational DDL.

V-Next observability behavior
- init() validates discovery/signal tables for engine compatibility under replication.
- Discovery MV/table reconciliation and MV refresh/wait operations use ON CLUSTER when replication is enabled.
- Additive migrations and MODIFY TTL run via ON CLUSTER when replication is enabled.
- migrateSpans() is disabled under replication (requires manual migration first).
- dangerouslyClearAll() truncates signal tables with ON CLUSTER to clear all replicas.

Testing
- New unit tests (replication.test.ts) cover engine mapping, idempotent ON CLUSTER insertion, nested-parentheses handling, zookeeperPath/replica templating, config validation, and ClickhouseDB.createTable() behavior (detection of existing local engines and emitted replicated DDL).
- v-next observability integration tests verify init() rejects when local tables exist under replication and that created signal tables use Replicated/Shared engines.
- Some unrelated v-next tests were removed during test rework.

Documentation & changeset
- ClickHouse README and reference docs updated with replication config docs, default zookeeperPath/replicaName patterns, guidance, and operational notes (including explicit guidance not to set replication on ClickHouse Cloud).
- Changeset added documenting opt-in replicated support and operational constraints.

---

## Migration & Operational Notes

- Enabling replication requires pre-existing Mastra tables be migrated to Replicated* engines first; init() will fail if local non-replicated Mastra tables exist.
- v-next automatic signal-table migrations are blocked when replication is enabled; perform manual migration before enabling replication.
- Default zookeeperPath and replicaName templates are provided but can be overridden to match cluster conventions.
- ON CLUSTER is applied to Mastra-owned DDL; SYSTEM STOP/START MERGES are intentionally not clustered.


intojhanurag:
LGTM ! 
/cc @epinzur

intojhanurag (approved):

abhiaiyer91 (approved):

PR #17710, opened 2026-06-09
fix(core): preserve tool execution error state

### LINKED ISSUE #15569: Tool execution errors stored as successful results — lost on history reload

littlewwwhite:
## Summary
- Persist failed tool executions as `output-error` history parts instead of successful `result` parts.
- Keep legacy AI SDK v4 and Mastra v1 conversion paths compatible by mapping stored tool errors back to legacy tool-result messages when needed.
- Add regression coverage for persisted tool errors, v4 sanitization, v1 conversion, and the agent-network final result shape.

Fixes #15569.

## Testing
- `pnpm --filter ./packages/core exec vitest run src/loop/workflows/agentic-execution/llm-mapping-step.test.ts`
- `pnpm --filter ./packages/core exec vitest run src/agent/message-list/conversion/output-converter-provider-executed.test.ts`
- `pnpm --filter ./packages/core exec vitest run src/agent/message-list/prompt/convert-to-mastra-v1.test.ts`
- `pnpm --filter ./packages/core exec vitest run src/agent/agent-network.test.ts --testNamePattern "should NOT store redundant toolCalls"`
- `pnpm --filter ./packages/core exec vitest run src/loop/loop.test.ts -u`
- `pnpm --filter ./packages/core exec vitest run src/loop/loop.test.ts`
- `pnpm test:core`
- `pnpm --filter ./packages/core lint`
- `pnpm build:core`
- `git diff --check`

CodeRabbit CLI was not available locally (`coderabbit` and `cr` were not found).

## ELI5

When a tool breaks and throws an error, the system now correctly saves that failure to the conversation history instead of marking it as a successful result. When the conversation is reloaded later, the error stays properly marked as an error—with all the error details preserved—instead of being confusingly converted to a successful output.

## Summary

This PR fixes a data persistence issue where tool execution errors were incorrectly stored and loaded from conversation history. Previously, when a tool's `execute()` method threw an error, the error was persisted with `state: "result"` and the error text in the `result` field. Upon reloading from history, this appeared as a successful tool execution with the error message in the output, losing the semantic meaning that the tool had failed.

## Changes Overview

**Core Fix — llm-mapping-step.ts**  
Modified the tool error handling in `createLLMMappingStep` to persist failed tool executions with the correct state and error field. When a tool throws, it now calls `messageList.updateToolInvocation` with:
- `state: 'output-error'` (instead of `'result'`)
- `errorText: reifiedError.message` (instead of putting the error in a `result` field)

**Compatibility Layer — output-converter.ts**  
Extended `sanitizeAIV4UIMessages` to handle `output-error` tool invocations when converting to legacy AI SDK v4 format. Introduced `toAIV4CompatibleToolInvocation` helper that maps `output-error` invocations back to `state: 'result'` with the `errorText` moved to the `result` field, ensuring backward compatibility. Also updated `toolInvocations` filtering to preserve both successful results and errors.

**V1 Conversion — convert-to-mastra-v1.ts**  
Added two helper functions to support tool errors in Mastra v1 message format:
- `hasToolOutcome`: Returns true for tool invocations with `state: 'result'` (with a result field) or `state: 'output-error'`
- `toToolResultPart`: Maps both successful and error tool invocations to v1 tool-result messages, converting `output-error` by using `errorText` as the result content

These helpers ensure that both successful and failed tool executions are properly included in tool-result messages during v1 conversion.

**Tests**  
- Added regression test in `llm-mapping-step.test.ts` verifying that tool errors are persisted with `state: 'output-error'` and `errorText`
- Added test in `output-converter-provider-executed.test.ts` validating v4 sanitization converts `output-error` to legacy `result` format
- Added test in `convert-to-mastra-v1.test.ts` confirming v1 conversion handles `output-error` tool invocations
- Updated test in `agent-network.test.ts` to accept both `tool-result` and `tool-error` entries in final results

**Changeset**  
Added `@mastra/core` patch release note documenting that tool execution errors are now properly preserved in reloaded conversation history (closes `#15569`).

### DIFF
--- a/.changeset/soft-bushes-crash.md
+++ b/.changeset/soft-bushes-crash.md
@@ -1,0 +1,5 @@
+---
+'@mastra/core': patch
+---
+
+Fixed tool execution errors so reloaded history preserves failed tool calls instead of treating them as successful results. Fixes #15569.
--- a/packages/core/src/agent/agent-network.test.ts
+++ b/packages/core/src/agent/agent-network.test.ts
@@ -274,11 +274,11 @@ describe('Agent - network - finalResult token efficiency', () => {
     // But the tool call data should still be present in the messages array
     const messagesInFinalResult = parsedContent.finalResult.messages || [];
     const toolCallMessages = messagesInFinalResult.filter((m: any) => m.type === 'tool-call');
-    const toolResultMessages = messagesInFinalResult.filter((m: any) => m.type === 'tool-result');
+    const toolOutcomeMessages = messagesInFinalResult.filter((m: any) => m.type === 'tool-result' || m.type === 'tool-error');
 
-    // Verify tool calls are preserved in messages
+    // Verify tool calls and their result/error outcomes are preserved in messages
     expect(toolCallMessages.length).toBeGreaterThan(0);
-    expect(toolResultMessages.length).toBeGreaterThan(0);
+    expect(toolOutcomeMessages.length).toBeGreaterThan(0);
   });
 });
 
--- a/packages/core/src/agent/message-list/conversion/output-converter-provider-executed.test.ts
+++ b/packages/core/src/agent/message-list/conversion/output-converter-provider-executed.test.ts
@@ -1,6 +1,11 @@
 import { describe, expect, it } from 'vitest';
 import type { AIV5Type } from '../types';
-import { addStartStepPartsForAIV5, aiV5UIMessagesToAIV5ModelMessages, sanitizeV5UIMessages } from './output-converter';
+import {
+  addStartStepPartsForAIV5,
+  aiV5UIMessagesToAIV5ModelMessages,
+  sanitizeAIV4UIMessages,
+  sanitizeV5UIMessages,
+} from './output-converter';
 
 /**
  * Tests for provider-executed tool handling in sanitizeV5UIMessages.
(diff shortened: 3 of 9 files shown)


abhiaiyer91 (changes requested):
must respond to PR bots

abhiaiyer91:
Closing as superseded. The linked issue #15569 was resolved by merged PRs #16976 and #17403, which preserve tool execution errors as output-error across persistence and history reload. Thanks for the contribution.

PR #17711, opened 2026-06-09
fix(core,cli): surface channel startup errors

### LINKED ISSUE #17692: [BUG] Agent channels silently fail to start under `mastra start`
### Describe the Bug

A code-defined agent configured with `channels` (per the [Channels Quickstart](https://mastra.ai/docs/agents/channels)) connects correctly under `mastra dev`, but **silently never connects under `mastra start`** (the built `.mastra/output` bundle / deployed server). No socket connects, no webhook handler becomes ready, and **no error is logged** — the server otherwise boots fine and serves `/health`.

**Root cause (traced in `@mastra/core@1.41.0`):**

1. `Mastra.addAgent()` registers each agent and, for agents with channels, fires the init **fire-and-forget with no error handling**:

   ```js
   const agentChannelsInstance = mastraAgent.getChannels();
   if (agentChannelsInstance) {
     // ...getWebhookRoutes()...
     void agentChannelsInstance.initialize(this);   // no await, no .catch
   }
   ```

   `addAgent` is invoked from the `Mastra` constructor (`config.agents.forEach(... this.addAgent(...))`).

2. `AgentChannels.initialize(mastra)` (`packages/core/src/channels/agent-channels.ts`) reads storage as its very first step:

   ```js
   const memoryStore = storage ? await storage.getStore("memory") : undefined;
   if (!memoryStore) {
     throw new Error("Channels require storage to be configured on the Mastra instance. Configure a storage provider like LibSQLStore.");
   }
   ```

   Under the production bundle, `mastra.getStorage()` is **empty at this point** (storage isn't wired onto the instance yet when `addAgent` runs during construction), so it throws — *before* the Chat SDK instance is created.

3. Because step 1 is `void initialize(this)` with no `.catch`, the rejection is **swallowed**. The server boots normally; the channel is simply dead.

`mastra dev` wins the timing race (storage is ready when init runs — it logs `Using MastraStateAdapter…` then `Slack socket mode connected`); the `mastra start` bundle loses it and logs neither.

> **Distinct from #15948** (the 1.29.0 startup crash where channel webhook routes collided with `/api` prefix validation, since fixed): that was a *visible crash* from route validation. This is a *silent* failure from storage init-ordering plus a swallowed `void initialize()`.

**Suggested fixes:**

1. Initialize agent channels **after** the `Mastra` instance has finished wiring storage (or `await` storage readiness inside the channel-init path) so `mastra.getStorage()` is populated when `AgentChannels.initialize()` runs.
2. At minimum, **stop swallowing the error**: wrap the `void agentChannelsInstance.initialize(this)` in `addAgent` with `.catch(err => logger.error(...))`, mirroring the instance-level `#channels` init path which already uses `try/catch`. The silent failure is what makes this so hard to diagnose in production.

### Steps To Reproduce

1. Configure a code-defined agent with a Slack channel (socket/gateway mode, the default) plus instance storage — the documented Quickstart:

   ```ts
   // agents/support-agent.ts
   import { Agent } from '@mastra/core/agent'
   import { createSlackAdapter } from '@chat-adapter/slack'

   export const supportAgent = new Agent({
     id: 'support-agent',
     name: 'Support Agent',
     model: 'openai/gpt-5.5',
     channels: { adapters: { slack: createSlackAdapter() } },
   })

   // mastra/index.ts
   import { Mastra } from '@mastra/core'
   import { LibSQLStore } from '@mastra/libsql'

     agents: { supportAgent },
     storage: new LibSQLStore({ url: process.env.DATABASE_URL }),
   })
   ```

2. Run `mastra dev` → logs `Using MastraStateAdapter (subscriptions persist across restarts)` then `[chat-sdk:slack] Slack socket mode connected`. ✅ Works.
3. Run `mastra build && mastra start` (same env) → server boots and serves `/health`, but logs **none** of the channel/chat lines and the Slack socket never connects — and **no error is printed**. ❌ Broken, silently.
4. Confirm it's `getStorage()` emptiness (not an uninitialized store): pre-`await storage.getStore("memory")` *before* `new Mastra(...)` returns a ready store, yet the channel still doesn't connect under `mastra start`.

https://github.com/berrydev-ai/mastra-channels-prod-repro

### Expected Behavior

Agent channels should initialize under `mastra start` (production) exactly as they do under `mastra dev` — the Slack socket connects (or the webhook handler becomes ready). At minimum, any failure during channel initialization should be surfaced/logged rather than silently swallowed by the un-awaited `void initialize()`.

### Environment Information

```block
@mastra/core:        1.41.0
mastra (CLI):        1.12.2
@mastra/libsql:      1.12.1
@chat-adapter/slack: 4.29.0
Runtime:             Bun 1.3.14 / Node.js 22
Storage:             LibSQL (DATABASE_URL)
Repro command:       `mastra build && mastra start` (production) — works under `mastra dev`
```

### Verification

littlewwwhite:
## Summary
- Defer agent-level channel initialization until after the `Mastra` constructor finishes.
- Log agent channel initialization failures through the Mastra logger instead of leaving rejected promises silent/unhandled.
- Stream child server stderr live from `mastra start`, matching `mastra worker`, so channel/adapter startup errors are visible while the server keeps running.
- Add regression coverage for constructor-time deferral, initialization failure logging, and live stderr forwarding.

Fixes #17692.

## Testing
- Red test confirmed before the core fix: `pnpm --filter ./packages/core exec vitest run src/channels/__tests__/integration.test.ts`
- Red test confirmed before the CLI fix: `pnpm --filter ./packages/cli exec vitest run src/commands/start/start.test.ts`
- `pnpm --filter ./packages/core exec vitest run src/channels/__tests__/integration.test.ts`
- `pnpm --filter ./packages/core exec vitest run src/channels/__tests__/agent-channels.test.ts src/channels/__tests__/integration.test.ts src/channels/__tests__/state-adapter.test.ts`
- `pnpm --filter ./packages/core exec vitest run src/mastra/processor-workflow-registration.test.ts src/mastra/workspace-registration.test.ts src/mastra/config-spread.test.ts`
- `pnpm test:core`
- `pnpm --filter ./packages/core lint`
- `pnpm build:core`
- `pnpm --filter ./packages/cli exec vitest run src/commands/start/start.test.ts`
- `pnpm build:cli`
- `pnpm --filter ./packages/cli typecheck`
- `pnpm --filter ./packages/cli lint`
- `pnpm test:cli`
- `git diff --check`

CodeRabbit CLI was not available locally (`coderabbit` and `cr` were not found).

## ELI5

This PR fixes a bug where the Mastra server was silently ignoring errors that happened when starting up agent channels. The problem was that initialization was happening too early (before everything was ready) and errors weren't being shown to anyone. The fix makes sure initialization happens at the right time and that any errors get logged and displayed so developers can see what went wrong.

## Summary

This PR addresses issue `#17692` by fixing how agent channel initialization errors are handled in Mastra's production setup (`mastra start`). The core issue was that channels were being initialized during the Mastra constructor (before storage was wired), causing initialization failures to be silently swallowed as unhandled rejected promises.

### Changes

**Core initialization logic** (`packages/core/src/mastra/index.ts`):
- Added a private `#initializeAgentChannels()` helper that asynchronously initializes agent channels and logs any errors through the Mastra logger instead of leaving rejected promises unhandled
- Updated `addAgent()` to defer channel initialization by routing through this new helper after the Mastra constructor completes

**CLI server output** (`packages/cli/src/commands/start/start.ts`):
- Modified the `start` command to stream child server stderr in real-time to the parent process, ensuring startup and channel initialization errors are visible immediately during server boot (matching the behavior of `mastra worker`)

**Testing** (`packages/core/src/channels/__tests__/integration.test.ts` and `packages/cli/src/commands/start/start.test.ts`):
- Added integration test verifying that `channels.initialize()` is deferred until after the Mastra constructor completes
- Added integration test validating that channel initialization failures are logged via the Mastra logger
- Added unit test for stderr forwarding in the start command

**Documentation** (`.changeset/` entries):
- Added changelog entries for both `@mastra/core` and `mastra` packages documenting these fixes

### Impact

With these changes, channel initialization failures that previously caused the server to boot without active channel connections (and without any visible error logging) will now:
1. Initialize at the correct time when storage is available
2. Have their errors properly logged through the Mastra logger
3. Be visible in real-time during server startup via stderr forwarding

### DIFF
--- a/.changeset/orange-pugs-obey.md
+++ b/.changeset/orange-pugs-obey.md
@@ -1,0 +1,5 @@
+---
+'@mastra/core': patch
+---
+
+Fixed agent channels so initialization runs after Mastra construction and startup failures are logged instead of being silently swallowed. Fixes #17692.
--- a/.changeset/orange-wasps-laugh.md
+++ b/.changeset/orange-wasps-laugh.md
@@ -1,0 +1,5 @@
+---
+'mastra': patch
+---
+
+Fixed `mastra start` so server stderr is shown while the process is running, making channel startup errors visible in production. Fixes #17692.
--- a/packages/cli/src/commands/start/start.test.ts
+++ b/packages/cli/src/commands/start/start.test.ts
@@ -89,4 +89,29 @@ describe('start command - customArgs handling', () => {
 
     expect(commands).toEqual(['index.mjs']);
   });
+
+  it('should stream server stderr while the process is still running', async () => {
+    const stderrWrite = vi.spyOn(process.stderr, 'write').mockImplementation(() => true);
+    let stderrHandler: ((data: Buffer) => void) | undefined;
+    const mockProcess = {
+      stderr: {
+        on: vi.fn((event: string, handler: (data: Buffer) => void) => {
+          if (event === 'data') {
+            stderrHandler = handler;
+          }
+        }),
+      },
+      on: vi.fn(),
+      kill: vi.fn(),
+    };
+    spawnMock.mockReturnValue(mockProcess);
+    const { start } = await import('./start');
+
+    await start();
+    stderrHandler?.(Buffer.from('channel failed\n'));
+
+    expect(stderrWrite).toHaveBeenCalledWith(Buffer.from('channel failed\n'));
+
+    stderrWrite.mockRestore();
+  });
 });
--- a/packages/cli/src/commands/start/start.ts
+++ b/packages/cli/src/commands/start/start.ts
@@ -45,6 +45,7 @@ export async function start(options: StartOptions = {}) {
     let stderrBuffer = '';
     server.stderr.on('data', data => {
       stderrBuffer += data.toString();
+      process.stderr.write(data);
     });
 
     server.on('exit', code => {
--- a/packages/core/src/channels/__tests__/integration.test.ts
+++ b/packages/core/src/channels/__tests__/integration.test.ts
@@ -1,5 +1,5 @@
 import { MockLanguageModelV2, convertArrayToReadableStream } from '@internal/ai-sdk-v5/test';
-import { describe, expect, it } from 'vitest';
+import { describe, expect, it, vi } from 'vitest';
 
 import { Agent } from '../../agent';
 import { Mastra } from '../../mastra';
@@ -57,6 +57,10 @@ function createTestAgent(id: string, options?: { channels?: ChannelConfig }) {
     }),
     channels: options?.channels,
   });
+}
+
+function waitForChannelInit() {
+  return new Promise(resolve => setImmediate(resolve));
 }
 
 describe('Mastra Channel Integration', () => {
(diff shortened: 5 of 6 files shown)


littlewwwhite:
CI triage update:

- The failing `Combined Store Tests / test (pinecone)` job is failing before any assertions because `PINECONE_API_KEY` is empty in the fork PR environment (`PineconeConfigurationError: apiKey`). This looks environment/secret related rather than related to this PR's CLI/core stderr forwarding changes.
- The failing `Unit and E2E Tests / E2E Test (packages/*)` job failed in `packages/mcp` on `configuration.e2e.test.ts > should subscribe and unsubscribe from a resource on a specific server` with a transient `weather` fixture connection failure. I reran the same focused MCP test locally and it passed.
- Focused local validation for this PR still passes:
  - `pnpm --filter ./packages/cli exec vitest run src/commands/start/start.test.ts --reporter=dot`
  - `pnpm --filter ./packages/core exec vitest run src/channels/__tests__/integration.test.ts --reporter=dot`
  - `pnpm --filter ./packages/mcp exec vitest run src/client/configuration.e2e.test.ts -t "should subscribe and unsubscribe from a resource on a specific server" --reporter=dot`

I tried to rerun the failed Actions jobs, but GitHub returned: `Must have admin rights to Repository`. Could a maintainer rerun the failed jobs and authorize the Vercel checks if needed?

LekoArts:
Fixed by https://github.com/mastra-ai/mastra/pull/17720

PR #17717, opened 2026-06-09
fix(durable-agent): populate MASTRA_THREAD_ID_KEY on step requestContext for input processors

### LINKED ISSUE #17714: [BUG] ToolSearchProcessor loaded tools never become available under DurableAgent (thread id missing in input-processor requestContext)
## Summary
An Agent configured with a ToolSearchProcessor input processor and wrapped in a DurableAgent cannot use cold-loaded tools. The model correctly calls search_tools → load_tool (which reports success), but on the next turn the loaded tool is still not in the tool set, so the agent re-searches/re-loads in a loop and the run ultimately fails. Cold tools are effectively unreachable.

## Root-cause signal
ToolSearchProcessor tracks loaded tools in an in-memory Map keyed by requestContext.get(MASTRA_THREAD_ID_KEY) ("mastra__threadId"). Wrapping processInputStep to log per step, we observed:

- RequestContext.get(MASTRA_THREAD_ID_KEY) is undefined inside the processor, even though .stream() was called with memory: { resource, thread }. The processor therefore falls back to the 'default' bucket.
- Manually doing requestContext.set('mastra__threadId', threadId) before .stream() makes the processor see the real id — but loaded tools still don't appear on later turns, which suggests the in-memory loaded-tool Map also isn't carried across DurableAgent step boundaries.

## Actual
1. MASTRA_THREAD_ID_KEY is undefined in the input-processor requestContext under DurableAgent.
2. Loaded tools never appear on subsequent turns; the agent loops; the run fails.

## Workaround
Removed ToolSearchProcessor and registered all tools warm.

  Questions
1. Is ToolSearchProcessor intended to work under DurableAgent?
2. Should MASTRA_THREAD_ID_KEY be populated in input-processor requestContext from memory.thread?
3. Is the processor's in-memory loaded-tool Map expected to survive durable step boundaries/replays? If not, should loaded-tool state be persisted (e.g., to working memory) so cold-loading works durably?

### Steps To Reproduce

  Minimal reproduction
```ts
  import { DurableAgent } from '@mastra/core/agent/durable';
  import { ToolSearchProcessor } from '@mastra/core/processors';
  import { z } from 'zod';

  const searchable = {
    getThing: createTool({
      id: 'getThing',
      description: 'Fetch a thing by id',
      inputSchema: z.object({ id: z.string() }),
      execute: async () => ({ ok: true }),
    }),
    // a few more so search/load is genuinely exercised
  };

  const base = new Agent({
    id: 'demo',
    model: 'google/gemini-2.5-flash',
    instructions:
      'If you lack a needed capability, call search_tools, then load_tool, then use the tool.',
    inputProcessors: [new ToolSearchProcessor({ tools: searchable, search: { topK: 5 } })],
    memory: new Memory({ storage: new PostgresStore({ connectionString }) }),
  });

  const agent = new DurableAgent({ agent: base });

  await agent.stream([{ role: 'user', content: 'Fetch thing 123' }], {
    memory: { resource: 'user-1', thread: 'thread-1' },
  });

  Add a thin wrapper around the processor to expose the problem:
  const tsp = new ToolSearchProcessor({ tools: searchable, search: { topK: 5 } });
  const traced = {
    id: tsp.id,
    name: tsp.name,
    processInputStep: async (args) => {
      const r = await tsp.processInputStep(args);
      console.log('thread=', args.requestContext?.get('mastra__threadId'),
                  'available=', Object.keys(r.tools ?? {}));
      return r;
    },
  };
  // use `traced` as the inputProcessor
```

  Observed: thread= logs undefined, and getThing never appears in available= after load_tool
  succeeds.

n/a

### Expected Behavior

1. The thread id passed via memory.thread should be visible to input processors as MASTRA_THREAD_ID_KEY.
2. A tool loaded via load_tool should be present on the next turn within the same run/thread.

### Environment Information

```block
Environment

  - @mastra/core 1.41.0
  - @mastra/memory 1.20.2
  - @mastra/pg 1.12.1
  - @mastra/mcp 1.9.1
  - mastra (CLI) 1.12.2
  - Node 24.11.x, run via mastra dev
  - Model: google/gemini-* (via @ai-sdk/google)
  - Storage/memory: PostgresStore + Memory, called with a threadId
```

### Verification

Aayush-engineer:
## Description

`ToolSearchProcessor` keys its loaded-tool state by `requestContext.get(MASTRA_THREAD_ID_KEY)`. Under `DurableAgent` this was always `undefined`, so every thread shared the same `'default'` bucket — meaning cold-loaded tools from one thread leaked into others, and on the agent's next turn the loaded tools were missing because the bucket was keyed wrong.

**Root cause:** Two places hand off a `RequestContext` to input processors, and neither was setting `MASTRA_THREAD_ID_KEY`:

1. `preparation.ts` — runs input processors once before the durable workflow starts. `threadId` was extracted from `execOptions.memory.thread` but never written onto `requestContext`.
2. `llm-execution.ts` — runs input processors on every subsequent durable step. The workflow engine injects a **fresh** `RequestContext` per step, so even if `preparation.ts` had set it, the key would be missing again on turns 2, 3, ...

**Fix:** Mirror what the non-durable agent path already does — set `MASTRA_THREAD_ID_KEY` and `MASTRA_RESOURCE_ID_KEY` on `requestContext` from the resolved `threadId`/`resourceId` before handing off to processors, in both places.

**Not addressed:** The secondary issue raised in #17714 — loaded-tool state lost across true durable suspend/resume (different worker process or after TTL expiry) — requires persisting `ToolSearchProcessor.threadLoadedTools` to working memory and is a separate architectural change.

## Related issue(s)

Fixes #17714

## Type of change

## Checklist

## ELI5 (Explain Like I'm 5)

When the durable agent ran multiple conversations at the same time, tools that were loaded in one conversation would accidentally get mixed up and appear in other conversations. The fix makes sure the system tells the tool processor which conversation it's working on, so tools stay in their own conversation instead of being shared with everyone.

## Changes

This PR fixes a critical bug in DurableAgent where input processors (specifically `ToolSearchProcessor`) were unable to identify which thread they were operating on, causing all threads to share a single bucket of loaded tools instead of maintaining per-thread state.

### Root Cause
The workflow engine injects a fresh `RequestContext` for each durable step, but two code paths were not populating `MASTRA_THREAD_ID_KEY` and `MASTRA_RESOURCE_ID_KEY`:
1. `preparation.ts` — runs input processors once before the durable workflow starts
2. `llm-execution.ts` — runs input processors on every subsequent durable step

Without these keys, `ToolSearchProcessor` falls back to a 'default' bucket, causing cold-loaded tools to leak across threads and disappear on subsequent agent turns within the same thread.

### Solution
Both files were updated to mirror the non-durable agent behavior:

- **preparation.ts**: Now sets `MASTRA_THREAD_ID_KEY` and `MASTRA_RESOURCE_ID_KEY` in `requestContext` when `threadId` and `resourceId` are resolved from `execOptions.memory`

- **llm-execution.ts**: Now ensures `requestContext` has these keys set from `typedInput.state` before passing to input processors

- **Changeset**: Documents the fix and notes that a separate issue—persisting `ToolSearchProcessor.threadLoadedTools` across durable suspend/resume—remains unaddressed and requires architectural changes to persist state to working memory

### Lines Changed
- +35/-0 across affected files (low code complexity)

### DIFF
--- a/.changeset/fix-durable-agent-tool-search-thread-id.md
+++ b/.changeset/fix-durable-agent-tool-search-thread-id.md
@@ -1,0 +1,22 @@
+---
+"@mastra/core": patch
+---
+
+fix(durable-agent): populate MASTRA_THREAD_ID_KEY on step requestContext for input processors
+
+The workflow engine injects a fresh RequestContext into each durable step,
+so MASTRA_THREAD_ID_KEY was undefined when ToolSearchProcessor (and other
+input processors) read it. The processor fell back to the 'default' bucket,
+causing all threads to share one loaded-tool set and cold-loaded tools to
+never appear on subsequent turns under DurableAgent.
+
+Fix: before running input processors in the LLM execution step, copy
+threadId and resourceId from typedInput.state onto the step's requestContext
+if not already set — mirroring what preparation.ts does before the initial
+processor run.
+
+Note: the secondary issue (loaded-tool state lost across process boundaries
+on true durable suspend/resume) requires persisting ToolSearchProcessor state
+to working memory and is not addressed here.
+
+Fixes #17714
--- a/packages/core/src/agent/durable/preparation.ts
+++ b/packages/core/src/agent/durable/preparation.ts
@@ -6,7 +6,7 @@ import type { MastraMemory } from '../../memory/memory';
 import type { MemoryConfig, MemoryConfig as _MemoryConfig, StorageThreadType } from '../../memory/types';
 import type { InputProcessorOrWorkflow, OutputProcessorOrWorkflow, ErrorProcessorOrWorkflow } from '../../processors';
 import type { ProcessorState } from '../../processors/runner';
-import { RequestContext, MASTRA_VERSIONS_KEY, mergeVersionOverrides } from '../../request-context';
+import { RequestContext, MASTRA_VERSIONS_KEY, MASTRA_THREAD_ID_KEY, MASTRA_RESOURCE_ID_KEY, mergeVersionOverrides } from '../../request-context';
 import type { VersionOverrides } from '../../request-context';
 import type { CoreTool, ToolHooks } from '../../tools/types';
 import type { Workspace } from '../../workspace';
@@ -143,6 +143,8 @@ export async function prepareForDurableExecution<OUTPUT = undefined>(
     typeof execOptions?.memory?.thread === 'string' ? { id: execOptions.memory.thread } : execOptions?.memory?.thread;
   const threadId = thread?.id;
   const resourceId = execOptions?.memory?.resource;
+  if (threadId) requestContext.set(MASTRA_THREAD_ID_KEY, threadId);
+  if (resourceId) requestContext.set(MASTRA_RESOURCE_ID_KEY, resourceId);
   let threadObject: StorageThreadType | undefined;
   let threadExists = false;
 
--- a/packages/core/src/agent/durable/workflows/steps/llm-execution.ts
+++ b/packages/core/src/agent/durable/workflows/steps/llm-execution.ts
@@ -236,6 +236,16 @@ export function createDurableLLMExecutionStep(_options?: DurableLLMExecutionStep
 
             const registryEntry = globalRunRegistry.get(runId);
             const executionAbortSignal = (registryEntry as any)?.abortSignal ?? abortSignal;
+            // Ensure threadId/resourceId are set on the step's requestContext so input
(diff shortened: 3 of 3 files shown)


roaminro:
Hey @Aayush-engineer ! Thanks for the contribution, but this PR wouldn't fix the issue which is that when restarting the server, the tools would not be rehydrated because they're currently stored in RAM. 
Closing in favor of https://github.com/mastra-ai/mastra/pull/17691

PR #17722, opened 2026-06-09
feat(auth): add dual auth system for Studio vs API

rphansen91:
## Summary

Adds **dual auth system** to Mastra, allowing separate authentication providers for:
- **Studio UI** → internal team members (SSO, dashboard access)
- **API** → external customers (tokens, API keys)

## Why This Matters

Previously, the same auth provider handled both Studio and API requests. This caused confusion because:
1. Internal team using SSO to access Studio !== external apps using API tokens
2. Different audiences need different auth patterns (OAuth for humans, tokens for machines)
3. RBAC/FGA rules for "what can this teammate do in Studio" differs from "what can this API consumer do"

### Core (`@mastra/core`)
- Add `StudioConfig` type with `auth`, `rbac`, and `fga` options
- Add `studio?: StudioConfig` to main `Config` interface
- Add `mastra.getStudio()` accessor

### Server (`@mastra/server`)
- Add `MASTRA_AUTH_MODE_KEY` constant to track auth provider used
- Add `getEffectiveAuthConfig()` to route based on `x-mastra-client-type` header
- Add `getEffectiveRBACProvider()` and `getEffectiveFGAProvider()` for dual RBAC/FGA
- Make all auth routes studio-aware (`/auth/capabilities`, `/auth/me`, SSO, credentials, etc.)

### Server Adapters (Hono, Express, Fastify, Koa)
- Update `checkRoutePermission()` calls to pass `requestContext` for dual auth routing

## Security

When `x-mastra-client-type: studio` header is present but `studio.auth` is not configured, the server returns **401 Unauthorized** instead of falling back to server auth. This prevents external users from spoofing the studio header to access Studio UI.

## Usage Example

```typescript
const mastra = new Mastra({
  // API authentication for external customers
  server: {
    auth: new MastraAuthWorkos({ ... }),
    rbac: new MastraRBACWorkos({ ... }),
  },
  // Studio authentication for internal team
  studio: {
    auth: new MastraAuthOkta({ ... }),
    rbac: new StaticRBACProvider({
      roles: DEFAULT_ROLES,
      getUserRoles: (user) => [user.role],
    }),
  },
});
```

## Test Plan

- [x] All 1771 server tests pass
- [x] All 53 auth-workos tests pass
- [x] Verified SimpleAuth blocking unauthenticated requests
- [x] Verified WorkOS SSO flow working correctly
- [x] Build succeeds for core, server, and all adapters

## Related

- Cherry-picked from `auth-vnext` branch commits: `627370c`, `7ccaed8`, `c00bf66`
- Foundation for team/users management UI (next PR)

## ELI5

Mastra now supports two separate login systems: one for internal team members using the Studio dashboard (which can use SSO), and another for external customers using the API (which can use API tokens). The server detects whether a request is from Studio or the API and uses the appropriate auth/authorization configuration.

## Overview

This PR implements an opt-in dual authentication and authorization system that lets Studio UI and API traffic use separate providers for auth, RBAC, and FGA. Studio-specific configuration is supported via a new StudioConfig; when studio.auth is configured Studio requests use it exclusively, otherwise the system preserves backward compatibility by falling back to server (API) auth. The adapter plumbing tags each request with an auth mode so downstream permission checks pick the correct RBAC/FGA provider.

### Core (`@mastra/core`)
- New StudioConfig type (auth?, rbac?, fga?) to hold Studio-specific auth/authorization.
- Added optional studio?: StudioConfig to the main Config type.
- Mastra now stores studio config and exposes getStudio(): StudioConfig | undefined.

### Server (`@mastra/server`)
- Added MASTRA_AUTH_MODE_KEY and MastraAuthMode ('studio' | 'server') to track which auth mode was selected per request.
- Implemented request-aware helpers to choose providers:
  - getEffectiveAuthConfig() / equivalent logic to select Studio vs server auth
  - getEffectiveRBACProvider() and getEffectiveFGAProvider() to choose RBAC/FGA based on auth mode
- Auth handlers (/auth/capabilities, /auth/me, SSO flows, logout/refresh, credentials sign-in/sign-up) now select providers based on whether the request is a Studio request (x-mastra-client-type header).
- Dual-auth behavior is opt-in: when studio.auth exists Studio requests use it exclusively; otherwise Studio requests fall back to server auth, preserving backward compatibility.

### Server Adapter (`@mastra/server`)
- Detects Studio requests via x-mastra-client-type and stores selected auth mode in RequestContext for downstream use.
- checkRouteAuth uses the effective auth config; checkRoutePermission now accepts an optional RequestContext and selects the RBAC provider according to the stored auth mode.
- Re-exports MastraAuthMode and related constants/types for adapter consumers.

### Server Adapter Implementations (Express, Fastify, Hono, Koa)
- Updated to capture a local requestContext and pass it to checkRoutePermission() and FGA checks:
  - Express: use res.locals.requestContext and pass it to checkRoutePermission/checkRouteFGA.
  - Fastify: use request.requestContext and pass it through.
  - Hono: read requestContext from Hono context and pass it through.
  - Koa: use ctx.state.requestContext and pass it to permission/FGA checks.
- No public API signature changes in adapters; behavior is backward compatible.

## Security Notes
- Studio auth is isolated when configured: Studio requests use Studio auth exclusively (no fallback).
- If studio.auth is not configured, legacy behavior is preserved: Studio requests fall back to server auth (opt-in dual-auth model).
- The adapter records auth mode in RequestContext to prevent ambiguous authorization decisions and to ensure RBAC/FGA use the correct provider.

## Testing & Build
- Server tests: 1771 passed.
- Auth WorkOS tests: 53 passed.
- SSO and SimpleAuth flows verified; core, server, and adapters build successfully.


rphansen91:
## Response to CodeRabbit: Dual Auth Fallback Behavior

CodeRabbit flagged `getAuthProvider()` falling back to `server.auth` for Studio requests as a security concern.

**We're intentionally keeping the fallback for backward compatibility**, but it's now **opt-in**:

| `studio.auth` configured? | Studio request uses |
|---------------------------|---------------------|
| **No** (existing users) | `server.auth` — preserves existing behavior |
| **Yes** (opt-in dual auth) | `studio.auth` exclusively — strict isolation, no fallback |

This means:
- **Existing users** who only have `server.auth` won't be broken
- **Users who want isolation** can configure `studio.auth` to get strict separation

The security concern is valid, but only applies when dual auth is explicitly enabled — and in that case, we DO enforce strict isolation with no fallback.

abhiaiyer91 (approved):

PR #17726, opened 2026-06-09
perf(pg): batch storage inserts into multi-row statements (#17398)

### LINKED ISSUE #17398: [Storage] Batch Postgres storage inserts into fewer round trips

rodboev:
Fixes #17398

## Summary

`PgDB.batchInsert()` already receives batches, but it still sends one `INSERT` per record inside a transaction. This PR converts those sequential inserts into bounded multi-row statements, preserving existing conflict behavior while cutting round trips for high-volume Postgres writes like observability spans.

## What Changed

- `stores/pg/src/storage/db/index.ts`: `batchInsert()` now emits parameterized multi-row inserts with chunking
- Postgres storage tests: cover chunking, span conflicts, duplicate conflict targets inside one input batch, and single schema-column metadata lookup for the batch path

## Why It Matters

Busy observability workloads already batch rows before calling storage. Turning those batches into actual multi-row inserts removes avoidable network and query overhead without changing the storage API.

## Verification

- `pnpm --filter ./stores/pg exec tsc --noEmit -p tsconfig.json` passed. Covers the package typecheck and the reviewed `pendingColumns` narrowing path.
- `pnpm --filter ./stores/pg run build:lib` passed. Covers the `@mastra/pg` library build.
- `pnpm --filter ./stores/pg exec vitest run src/storage/db/transaction-client.test.ts` passed, 4/4 tests. Covers multi-row chunking, duplicate span conflict splitting, transaction ownership, and metadata lookup caching.
- `pnpm --filter ./stores/pg exec eslint src/storage/db/index.ts src/storage/db/transaction-client.test.ts` passed. Covers the changed production and test files.

## ELI5

This PR sends many database records in fewer PostgreSQL requests. This reduces database overhead while preserving existing transaction and conflict behavior.

## What changed

- `PgDB.batchInsert()` now uses parameterized multi-row `INSERT` statements.
- Inserts are chunked to stay below PostgreSQL bind-parameter limits.
- Existing transaction boundaries and column filtering remain unchanged.
- Schema metadata is cached before chunking.
- Span inserts preserve `ON CONFLICT ("traceId", "spanId")` behavior.
- Duplicate span conflict targets are split across statements.
- A patch changeset was added for `@mastra/pg`.

## Tests

Updated tests verify:

- Multi-row insert chunking.
- SQL placeholders and bind-value counts.
- Duplicate span conflict targets.
- Span conflict handling.
- Transaction statement recording.

## Impact

High-volume PostgreSQL writes use fewer queries and round trips. The public storage API remains unchanged.

### DIFF
--- a/.changeset/plenty-lamps-sip.md
+++ b/.changeset/plenty-lamps-sip.md
@@ -1,0 +1,5 @@
+---
+'@mastra/pg': patch
+---
+
+Improved Postgres batch inserts to use fewer queries while keeping span conflict handling unchanged.
--- a/stores/pg/src/storage/db/index.ts
+++ b/stores/pg/src/storage/db/index.ts
@@ -25,6 +25,8 @@ import { buildConstraintName } from './constraint-utils';
 // Re-export DbClient for external use
 export type { DbClient } from '../client';
 
+const POSTGRES_MAX_BIND_PARAMETERS = 65_535;
+
 /**
  * Configuration for standalone domain usage.
  * Accepts either:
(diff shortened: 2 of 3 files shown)


rodboev:
Addressed CodeRabbit feedback: the parallel normalization is safe because the schema metadata cache is warmed before chunking, so normalizeForInsert hits the in-memory cache on every call.

rodboev:
Thanks for catching the package build failure. You're right that `hasPendingColumns` can't preserve narrowing for mutable `pendingColumns` across the captured `flush` closure.

I'll keep the change focused on the blocker:

1. Capture `pendingColumns` in a type-safe local `currentPendingColumns` snapshot before the compatibility checks.
2. Derive `incompatibleColumns` from that snapshot and remove the redundant `hasPendingColumns` branch term, so the undefined check and `join()` use the same stable value.

Batching, chunking, span conflict splitting, sequential normalization, and SQL generation stay unchanged.

abhiaiyer91:
Thanks for putting this together! #17398 was resolved on `main` by #24085, so this PR is no longer needed. Closing it to keep the queue focused, but we really appreciate the work here. If you find a case that #24085 doesn't cover, please open a new issue with a repro and we'll pick it up.

PR #17737, opened 2026-06-09
feat(datasets): add tool replay to experiments for deterministic agent evals

### LINKED ISSUE #17466: Datasets/Experiments: native tool replay/mocking for agent targets (deterministic eval of agents with side-effecting tools)
## Context

Datasets + Experiments are great, and the building blocks for point-in-time eval already exist: per-item `requestContext`, SCD-2 item versioning, and `extractTrajectoryFromTrace` (which exposes each `tool_call`'s `toolArgs`/`toolResult`).

## Problem

For `targetType: "agent"`, an experiment re-runs `agent.generate()` with the item's input + `requestContext`, but the agent's **tools execute live**. There's no native way to replay/mock recorded tool outputs. For agents that read/write **mutable external state** (support tickets, CRM, etc.) this means:

- **Non-deterministic**: the external resource drifts between capture and replay → same item, different result.
- **Real side effects**: tool writes hit production during an eval (e.g., a support agent creates a real ticket comment on every run).
- **Can't faithfully re-run a recorded trace**, which is the standard "record once, replay forever" pattern for agent evals.

`requestContext` doesn't solve this on its own: it's a request-scoped key-value bag passed to the run; each tool must be hand-coded to read it and short-circuit its own I/O. That's a per-tool workaround, not a framework mechanism.

## What we do today (workaround)

We stuff the recorded read responses (extracted via `extractTrajectoryFromTrace`) into the item's `requestContext`, and our generic HTTP tool checks for it: reads return the snapshot, writes become no-ops. Works, but it's custom glue every tool/project has to reinvent.

## Proposal (any of, in order of preference)

1. **Native tool replay in experiments**: an option on `startExperiment`/the item to supply recorded tool outputs keyed by `(toolName, args)` that the runtime returns instead of executing the tool. The data shape already exists in `extractTrajectoryFromTrace` (`toolArgs`/`toolResult`).
2. **Replay-from-trace mode**: given a captured trace, auto-stub tool execution from that trace's tool spans, with a configurable miss policy (error / passthrough).
3. **Minimum**: a documented per-run **tool-execution interception hook**, so users implement replay once instead of monkeypatching each tool.

## Prior art

"Record once, replay forever" is the common practice for deterministic, side-effect-free agent evals; the framework already has the capture (`extractTrajectoryFromTrace`) and storage (`requestContext`, SCD-2) — only the replay consumption is missing.

Happy to contribute a PR if there's a preferred direction.

mbenhamd:
## Description

Adds native tool replay and tool mocks to dataset experiments. Experiments with agent targets can replay tool outputs recorded in a prior traced run instead of executing live tools — and/or mock individual tools directly — making evals of agents with side-effecting tools deterministic and side-effect free.

```typescript
// 1. Record: any traced run works — experiments already persist a traceId per item result
const baseline = await dataset.startExperiment({
  targetType: 'agent',
  targetId: 'support-agent',
})

// 2. Replay: tool calls return the recorded outputs, no live tools run
const replayed = await dataset.startExperiment({
  targetType: 'agent',
  targetId: 'support-agent',
  toolReplay: { fromExperimentId: baseline.experimentId, matching: 'strict' },
})

// 3. Or mock tools without any recording
const mocked = await dataset.startExperiment({
  targetType: 'agent',
  targetId: 'support-agent',
  toolMocks: {
    weatherInfo: { output: { temp: 20 } },                                  // static stub
    paymentApi:  { error: { message: 'timed out' } },                       // failure injection
    createTicket: { expect: { args: { priority: 'high' }, calledTimes: 1 } }, // assertion (observe-only)
    geoLookup: { cases: [                                                   // args-conditional stub
      { args: { city: 'Paris' }, output: { lat: 48.85 } },
      { args: { city: 'Tokyo' }, error: { message: 'region offline' } },
    ] }, // no matching case → item fails (TOOL_MOCK_ARGS_MISMATCH) or runs live via onNoMatch
    searchDocs:  async ({ input }) => fakeIndex.search(input),              // replace execute (code-only)
  },
})

// 4. Study what actually happened
console.log(replayed.results[0].toolReplay)
// { sourceTraceId, totalRecorded, replayedCount, misses, unconsumed,
//   argMismatches, calls, mocks, expectations }
```

**Design**

- **The trace is the recording.** Replay events are derived from trace tool spans at replay time — no new "record mode". Each item resolves its source trace from `replayTraceId` (inline items), `metadata.replayTraceId` (stored items), or the `itemId → traceId` mapping of `toolReplay.fromExperimentId`'s results.
- **Interception reuses the per-execution `hooks` option** (`ToolHooks.beforeToolCall` short-circuit), so it covers every tool source (assigned tools, toolsets, MCP, memory) without touching the agent loop.
- **Two matching policies.** `'fifo'` (default) serves per-tool events in recorded order with a deterministic spanId tie-break; arg drift is replayed anyway and reported in `argMismatches`. `'strict'` serves an event only on an exact-args match (canonicalized; runtime-injected nullish keys ignored) — anything else is a miss, and recorded calls left unconsumed fail the item with `TOOL_REPLAY_UNCONSUMED` (not retried). In strict mode the tape is a contract in both directions: unexpected calls fail, and missing expected calls fail too.
- **Tool mocks ride the same hook**, with a clear precedence per call: mock → replay queue → miss policy. Data mocks stub outputs, inject errors, and/or assert calls (`expect: { args?, calledTimes? }` — unsatisfied assertions fail the item with `TOOL_MOCK_EXPECTATION_FAILED`, not retried; `calledTimes: 0` asserts a tool must NOT be called, optionally only with specific args). Function mocks replace `execute` entirely (code-only; not exposed over HTTP). Conditional data mocks (`cases: [{args, output|error}]`, first canonicalized-args match answers; `onNoMatch: 'error' | 'passthrough'`, default fails the item with `TOOL_MOCK_ARGS_MISMATCH`, never retried) cover the issue's literal (toolName, args) → output table without code. Mock keys that normalize to the same tool name are rejected at setup (`TOOL_MOCK_NAME_COLLISION`) instead of silently dropping an assertion, and structurally invalid `cases` are rejected upfront (`TOOL_MOCK_INVALID_CASES`).
- **`onMiss: 'error' | 'passthrough'`** (default `'error'` — never silently call a live tool). Misses abort the run via a per-attempt abort signal and fail the item without retries (`TOOL_REPLAY_MISS`). No recording at all → `TOOL_REPLAY_NO_RECORDING`; failed trace fetch → `TOOL_REPLAY_LOAD_FAILED` (fails only the affected item). Error-code precedence within one attempt is deterministic and documented: `TOOL_REPLAY_MISS` (run aborted) > `TOOL_MOCK_EXPECTATION_FAILED` > `TOOL_REPLAY_UNCONSUMED` — the report always carries the full picture regardless of which code wins.
- **Divergence is signal, not failure.** Every item carries a `toolReplay` report — replayed count, misses, unconsumed, arg mismatches, recording-quality signals (`redactedPayloadCount`, `staleRecording`), mock usage, expectation outcomes, and `calls[]`: the run's actual tool-call flow in order (replayed / mocked / missed / live, with the consumed tape position and arg-drift flags). The same outcomes are stamped on the run's own trace as flagged synthetic tool spans, keeping timelines and trajectory views readable. The report persists in a **dedicated `toolReplay` column** on experiment results (all six storage adapters), so scorers never see it and replay runs score identically to baselines.
- **Runs are honestly labeled.** Replay runs and runs whose mocks suppress live execution are stamped (`metadata.toolReplay` marker with `fromExperimentId`/`onMiss`/`matching` and/or `mockedTools`) and refused as replay sources — their traces lack real tool spans. Expect-only mock runs execute tools live, so they are not stamped and stay eligible. The marker also persists the data-mock configs (`mockConfigs`, function mocks as `{ function: true }` placeholders), making mock runs auditable and faithfully re-runnable. It is stamped on both the sync and async (pre-created record) trigger paths.
- **Subset runs.** `itemIds` restricts an experiment to selected dataset items — re-run a single divergent item with replay without re-running the whole dataset.

**Limitations (documented)**

- Agent targets only; workflow targets and inline `task` functions are rejected at setup.
- Tools inside sub-agents (agent-as-tool) or workflow-as-tool internals are not intercepted; the top-level call replays as a single recorded output.
- Approval-required tools suspend before the replay hook runs — not supported with replay.
- Replayed/mocked calls don't execute the real tool. Their traces stay readable: each served call emits a synthetic `TOOL_CALL` span flagged `metadata.toolReplay.synthetic` (with `outcome` and consumed `sequence`), and replay extraction ignores flagged spans — so a replay/mock run still can't serve as a recording (belt and braces with the marker + source refusal above).
- Function mocks are code-only (`startExperiment` in code); HTTP/Studio can send data mocks. A throwing function mock propagates like a throwing live tool.
- Recordings inherit tracing's sensitive-data filtering: redacted inputs surface as `argMismatches`; redacted outputs replay as `[REDACTED]`.
- Traces export asynchronously — flush before replaying a just-recorded run in the same process.

HTTP exposure ships in stacked PR #17754, client-js typing in #17771, Studio UI in the follow-up playground PR.

**Docs:** "Replaying tool calls", "Matching policy", and "Tool mocks" sections in the running-experiments guide; `toolReplay`/`toolMocks`/`itemIds` parameter docs and the result `toolReplay` field in the dataset references. Changesets included (`@mastra/core` minor, storage adapters patch).

## Related issue(s)

Fixes #17466

## Type of change

- [x] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [ ] Test update

## Validation

- `pnpm --filter ./packages/core exec vitest run src/datasets/experiment src/storage/domains/experiments` — 237 passed (55 replay unit: extraction ordering/filtering/tie-breaks/name normalization incl. synthetic-span filtering, FIFO + strict matching, miss policies, mock ladder incl. error-wins precedence, arg-filtered negative assertions, name-collision rejection, call-flow recording incl. function-mock rejections; 43 integration through the real runner: conditional cases (first-match, canonicalization, onNoMatch both ways), persisted mockConfigs round-trip, replay + mocks + strict, error-code precedence, async-path marker stamping, expect-only source eligibility, strict-unconsumed failure, itemIds subset runs, retry exclusions; plus the experiments storage round-trip of the new column). The suite keeps a baseline test demonstrating the original issue.
- True end-to-end tests in `@mastra/observability` exercise the REAL tracing pipeline (CoreToolBuilder spans → SensitiveDataFilter → storage exporter → replay), including the synthetic spans a replay/mock run writes to its own trace and the guarantee that chaining a replay at such a trace serves nothing.
- `pnpm --filter ./packages/core typecheck` / eslint / Prettier / docs remark lint — clean.

## Checklist


BeamusWayne:
I've built record/replay reliability gates for tool-using agents ([trace-vault](https://github.com/BeamusWayne/trace-vault)), so this problem space is close to home. The core calls here look right to me: deriving the recording from trace spans instead of a new artifact, intercepting via `beforeToolCall` so MCP/toolsets/memory are covered uniformly, fresh `ToolReplayState` per attempt, and breaking the retry loop on `TOOL_REPLAY_MISS` (a deterministic failure — retrying can only waste tokens). A few things from lessons learned building the same machinery:

**1. Per-tool FIFO can pair the wrong observation when the model reorders same-tool calls — and the item still "passes".**

Recorded run: `search("flights")` → `search("hotels")`. Replay run (live LLM, free to diverge): the model asks about hotels first → FIFO hands it the *flights* result. The run completes, scorers grade an answer built on swapped observations, and the only signal is a diagnostic `argMismatches` entry. Combined with `onMiss: 'passthrough'`, a run can also silently mix replayed and live state. Two cheap mitigations, neither blocking:

- a docs note that `argMismatches.length > 0` (and any `passthrough` miss) usually means "inspect before trusting the scores" — the report has exactly the right fields for this already;
- longer-term, an opt-in keyed mode: match on `(toolName, normalized args)` first, fall back to FIFO on miss. Was keyed-then-FIFO considered and rejected? FIFO-only is a defensible default given volatile args (see next point), but for read-heavy tools keyed matching eliminates the swap failure mode entirely.

**2. Volatile arg fields will make `argMismatches` permanently noisy for some tools.**

Args frequently embed values from prior outputs — pagination cursors, request IDs, timestamps. `deepEqual(input, event.input)` will then flag a mismatch on every replay of those tools, forever, and a report that always has entries stops being read. An optional per-experiment normalizer (or ignore-paths) applied *only* to the diagnostic compare — matching stays FIFO — keeps the report signal-dense. In trace-vault, volatile-field normalization was the single biggest reducer of false-positive divergence.

**3. `startedAt` ties make queue order adapter-dependent.**

`extractToolReplayEvents` sorts by `startedAt` (ms resolution). Parallel calls to the same tool can tie, and `Array.prototype.sort` being stable means ties resolve to adapter return order — which the comment above the sort correctly calls non-contractual. Same trace, different storage adapter → different FIFO order → different replay pairing. A deterministic tie-break (any monotonic per-span field, even `spanId` lexicographic as a last resort) would pin it down.

**4. The `FGADeniedError` carve-out is correct today — consider making it un-driftable.**

I checked the loop: `tool-call-step.ts:1171` is currently the only name-based re-throw, so the denylist in `buildReplayHooks` is complete. But it's a string literal duplicated far from its source of truth; if the loop ever special-cases another name (a tripwire/abort variant, say), replayed errors will quietly behave differently from live ones. Exporting a shared `RETHROWN_TOOL_ERROR_NAMES` constant from the loop and consuming it in `replay.ts` removes the failure mode.

Happy to run this branch against traces recorded from our agents and report back if more real-world divergence data would help.

mbenhamd:
Thanks @BeamusWayne, this is exactly the kind of review the feature needed, and the trace-vault experience shows.

On points 3 and 4 (startedAt ties, FGADeniedError drift): both are fixed in bc54bb96afc. The extractor now breaks ties by spanId using plain ordinal comparison (we avoided localeCompare, since locale sensitivity would trade one nondeterminism for another). The error name carve-out now lives in a shared `RETHROWN_TOOL_ERROR_NAMES` constant consumed by both the tool call step and the replay hooks, so the two cannot drift apart.

On point 1a (divergence as a trust signal): agreed. The docs now state it explicitly: arg mismatches mean the agent received re-paired recorded observations, passthrough misses mean live observations entered the run, and either way the report should be inspected before trusting scores. Related hardening from the same pass: scorers never see the report at all. It merges into the stored output only after scoring, so replay metadata cannot skew score comparisons against baselines.

On point 1b (keyed then FIFO): it was considered and deliberately deferred rather than rejected. Arg volatility (cursors, request IDs, timestamps embedding prior outputs) made order based matching the safer default. Keyed first matching also inherits exactly the false mismatch problem your point 2 describes, which is why I think the two land together: a volatile field normalizer is what makes keyed matching viable, not just a way to quiet the diagnostics. Both are on the follow up list.

On point 2 (normalizer): agreed on the failure mode. A report that always has entries stops being read. Same follow up bucket.

Real world divergence data from your agents would be genuinely useful, especially for deciding what the default report should surface. Yes please.

BeamusWayne:
Hey @mbenhamd — nice hardening pass in bc54bb9. I dug into the three checks that went red again on the re-run; two of them are one-line fixes inside this PR, the third is an unrelated flake. Details + suggested diffs below.

## 1. Quality assurance (run 27263407959) — eslint `import/order` in `tool-call-step.ts`

`@mastra/core#lint` fails on the new import added in bc54bb9:

```
packages/core/src/loop/workflows/agentic-execution/tool-call-step.ts
  28:1  error  `./rethrown-error-names` import should occur after import of `../schema`  import/order
```

The repo's `import/order` config (`packages/_config/src/eslint.js`) orders groups `parent` before `sibling`, so the `./rethrown-error-names` import has to come after the `../` imports. This is exactly what `eslint --fix` produces:

```diff
--- a/packages/core/src/loop/workflows/agentic-execution/tool-call-step.ts
+++ b/packages/core/src/loop/workflows/agentic-execution/tool-call-step.ts
@@ -25,9 +25,9 @@ import { ensureSerializable } from '../../../utils';
 import type { SuspendOptions } from '../../../workflows/step';
 import { createStep } from '../../../workflows/workflow';
 import type { OuterLLMRun } from '../../types';
-import { RETHROWN_TOOL_ERROR_NAMES } from './rethrown-error-names';
 import { serializeToolError, ToolNotFoundError } from '../errors';
 import { toolCallInputSchema, toolCallOutputSchema } from '../schema';
+import { RETHROWN_TOOL_ERROR_NAMES } from './rethrown-error-names';
 
 type AddToolMetadataOptions = {
   toolCallId: string;
```

Two notes on the rest of this workflow:

- Yesterday's Format-step failure (run 27240040405) was not yours — it tripped on 3 unformatted `.claude/skills/react-best-practices/` files that came in from main via #17736 and were reformatted on main in b7f5a4d3. Fresh merge-ref runs won't hit it again.
- Today the Format step was skipped (lint failed first). I pre-checked the 4 TS files this PR changed in bc54bb9 against the root prettier 3.8.3 config — all clean, so Format should pass once lint is green.

## 2. Lint Docs (run 27263408069) — prettier in `startExperiment.mdx`

`pnpm prettier --check .` (in `docs/`) flags `src/content/en/reference/datasets/startExperiment.mdx`. The culprit is the new `toolReplay` description: it's a single-quoted JS string containing an escaped apostrophe (`` `fromExperimentId`\'s ``). The docs prettier setup (3.8.3, `singleQuote: true`, plus `prettier-mdx.mjs` which formats `PropertiesTable` JSX expressions with the babel parser) rewrites strings containing a single quote to double quotes to drop the escape:

```diff
--- a/docs/src/content/en/reference/datasets/startExperiment.mdx
+++ b/docs/src/content/en/reference/datasets/startExperiment.mdx
@@ -136,7 +136,7 @@ console.log(`Status: ${summary2.status}`)
     type: 'object',
     isOptional: true,
     description:
-      'Replay recorded tool outputs from prior traced runs instead of executing live tools. Agent targets only. Each item resolves its source trace from `item.replayTraceId` (inline data), then `item.metadata.replayTraceId` (storage-backed items), then the itemId → traceId mapping of `fromExperimentId`\'s results. See [Replaying tool calls](/docs/evals/datasets/running-experiments#replaying-tool-calls).',
+      "Replay recorded tool outputs from prior traced runs instead of executing live tools. Agent targets only. Each item resolves its source trace from `item.replayTraceId` (inline data), then `item.metadata.replayTraceId` (storage-backed items), then the itemId → traceId mapping of `fromExperimentId`'s results. See [Replaying tool calls](/docs/evals/datasets/running-experiments#replaying-tool-calls).",
     properties: [
       {
         type: 'object',
```

Equivalently: `pnpm prettier --write src/content/en/reference/datasets/startExperiment.mdx` from `docs/`. I verified with the exact CI toolchain (prettier 3.8.3 + `docs/.prettierrc` + `docs/prettier-mdx.mjs`) that the current file fails `--check`, the diff above passes, and it's byte-identical to what `--write` produces. This check failed on both runs for the same reason — this file wasn't touched by bc54bb9.

## 3. Prebuild (run 27263408477) — unrelated flaky test, no PR change needed

`UNIT Test (Shard 1)` failed on:

```
FAIL unit:packages/core src/events/__tests__/per-thread-pubsub-multiprocess.test.ts
  > UnixSocketPubSub - multi-process per-thread isolation
  > elects a new broker and preserves IPC subscriptions after broker process exits
Error: Timeout waiting for "status" from worker   (test file line 56)
```

This isn't yours: the PR doesn't touch `packages/core/src/events` at all, the sibling tests in that same file passed in the same run, and the file has a flake history (it last changed in "test(core): stabilize PubSub failover stress test" #17164, May 27). For context, yesterday's Prebuild run failed on a *different* unrelated flake (`client-sdks/react` `use-speech-recognition.test.ts`), which has since been de-flaked on main — two consecutive runs, two different out-of-footprint flakes. The red "Merge Test Reports" job is purely downstream of the shard failure (it just re-aggregates the same failing test).

Pushing the two fixes above will re-trigger CI on a fresh merge ref anyway; if this broker-election test times out again, it deserves a re-run (or a flake report to maintainers) rather than a change in this PR.

**TL;DR:** one import reorder in `tool-call-step.ts` + one quote flip in `startExperiment.mdx` should turn Quality assurance and Lint Docs green; Prebuild's failure is an out-of-scope flake that a re-run should clear.

mbenhamd:
Thanks @BeamusWayne

PR #17768, opened 2026-06-10
fix(core,pg): reduce workflow snapshot bloat on tool-approval suspends

### LINKED ISSUE #17738: Memory Exhaustion with tool approval enabled during large payload steps @mastra/pg state persistence
This issue was created from Discord post 1513993643269296299:

[Open in Discord](https://discord.com/channels/1309558646228779139/1513993643269296299)

Hi Team! 

We are building an AIDLC platform with Mastra and hit a severe memory bottleneck specifically when enabling tool approval on long-running workflow steps.

**The Context & Evidences:**

**Without Tool Approval:** Memory fluctuates completely normally between 205MB and ~300MB.

**With Tool Approval:** We originally found a slightly higher baseline utilization of memory just by having tool approval enabled. We observed that memory usage gradually increases after every tool approval iteration (such as for file write and edit operations).

**The Trigger:** After bringing in our Figma integration, this progressive growth blew up exponentially. Rather than a stable ceiling, memory now accumulates aggressively after each step tool approval, rapidly climbing from 205MB to over 3GB and causing standard Node OOM crashes.

The Profiler Evidences:
I ran a clinic-heapprofiler on our Node process, and the heap snapshot points directly to the state-saving routine triggered during the approval phase:

parseRow (pg/lib/result.js): ~266.15 MB (Self Size)

persistWorkflowSnapshot (@mastra/pg/dist/index.cjs): ~143.46 MB (Self Size)

The data indicates that when the workflow engine serializes the context state to Postgres via @mastra/pg during a suspension, the heap experiences massive garbage collection pressure that builds up over time. I noticed a similar memory leak issue on GitHub (#6322) was closed a few months ago.

Appreciate any advice or structural patterns on how to handle heavy tool outputs when human-in-the-loop approvals are enabled!

intojhanurag:
## Description

Tool-approval workflows suspend on every approval cycle and persist the agent stream-state into the workflow snapshot. With large tool payloads (e.g. Figma) the snapshot grew tens of MB per step and `@mastra/pg` read-modified-wrote the whole row on every update, producing the `parseRow` (~266 MB self) and `persistWorkflowSnapshot` (~143 MB self) heap hotspots in #17738 and causing Node OOM crashes. This PR shrinks what gets serialized into the snapshot and removes the full-row round-trip on per-step updates.

- Trim `__streamState` in `packages/core/src/stream/base/output.ts` so buffered steps persist per-step request/response deltas instead of cumulative history; the full view is rebuilt on deserialize for resume.
- Deduplicate large tool-result payloads in serialized stream state by replacing matching `toolResults`, `dynamicToolResults`, and tool-result content-part outputs with `{ __toolResultRef: toolCallId }` references that rehydrate from the single message-list copy on resume; non-matching (transformed) outputs are kept as-is.
- In `stores/pg/src/storage/domains/workflows/index.ts`, switch `updateWorkflowResults` to a single in-place `jsonb_set` UPDATE for non-foreach step results and switch `updateWorkflowState` to a `snapshot || $3::jsonb` merge, so per-step updates no longer SELECT, parse, merge and rewrite the full snapshot row.
- Keep the transactional read-merge-write path as a fallback for foreach element-wise array merges and for the first-write-creates-row case; sanitize only the fragment being written for invalid JSON escapes / null chars.
- Add size-regression and dedup-rehydration tests in `output.test.ts`, plus pg integration tests covering foreach element-wise merge, snapshot auto-create on first `updateWorkflowResults`, and in-place unicode sanitization.

## Related Issue(s)

Fixes #17738

## Type of Change

- [ ] Documentation update
- [ ] Code refactoring
- [x] Performance improvement
- [ ] Test update

## Checklist

## ELI5
When a workflow pauses for tool approval, it saves “where it left off” so it can continue later. This PR makes that saved progress much smaller and easier to update, so memory doesn’t keep growing with every approval—and Postgres updates don’t repeatedly rewrite huge snapshots.

### What changed
- **`@mastra/core`**
  - Reduced workflow snapshot bloat by persisting **only per-step stream deltas** (avoids repeatedly saving cumulative `request`, `dbMessages`, `uiMessages`).
  - Reduced serialized size further by **deduplicating large tool results** into reference markers and **rehydrating** them on resume.
  - Updated serialization/deserialization logic to rebuild buffered steps and restore tool-result payloads from refs; **stops serializing `bufferedTextChunks`** (defaults on deserialize instead).
- **`@mastra/pg`**
  - Optimized workflow snapshot persistence to avoid full **read/parse/merge/rewrite** for common updates.
  - Added an **in-place fast path** using PostgreSQL `jsonb_set`/merge-style updates for:
    - `updateWorkflowResults` when the step output is **not** a foreach/array merge
    - `updateWorkflowState` when `snapshot.context` is a JSON object
  - Kept the **transactional read-merge-write fallback** for foreach array merges and for cases where the snapshot row must be created.
- **Release notes (changesets)**
  - Added changesets for both `@mastra/core` and `@mastra/pg` documenting the memory/snapshot bloat mitigation.

### Tests added/updated
- **Core unit tests** for:
  - snapshot size regression / linear JSON growth
  - buffered-step delta serialization round-trips (including legacy behavior)
  - tool-result dedup + rehydration
  - foreach-related delta behavior
  - unicode sanitization
- **PG integration tests** for:
  - in-place step-result updates
  - foreach element-wise merge semantics
  - implicit workflow snapshot creation on first update
  - unicode sanitization
- Fixed a test-fixture regression involving **a NUL byte** in the unicode sanitization test (replaced with a space during rebase), restoring CI.

### DIFF
--- a/.changeset/workflow-snapshot-suspend-bloat-core.md
+++ b/.changeset/workflow-snapshot-suspend-bloat-core.md
@@ -2,8 +2,8 @@
 '@mastra/core': patch
 ---
 
-Reduced workflow snapshot bloat when tool approval suspends with large tool payloads.
+Fixed memory exhaustion during tool-approval workflows with large tool payloads.
 
-When tool approval is enabled, every approval cycle suspends the agent loop and persists a stream-state snapshot. The serialized state now keeps per-step message deltas instead of cumulative `request`/`response` per buffered step, and replaces large tool-result payloads in step content with references to the message-list copy that get rehydrated on resume. Snapshots written by older versions still resume.
+Tool-approval snapshots now stay significantly smaller — large tool results (e.g. Figma exports) are no longer duplicated across the snapshot, and message history no longer grows quadratically with each approval cycle. This prevents Node OOM crashes in long-running approval flows. Snapshots written by older versions still resume without changes.
 
 Fixes [#17738](https://github.com/mastra-ai/mastra/issues/17738).
--- a/.changeset/workflow-snapshot-suspend-bloat-core.md
+++ b/.changeset/workflow-snapshot-suspend-bloat-core.md
@@ -1,0 +1,9 @@
+---
+'@mastra/core': patch
+---
+
+Reduced workflow snapshot bloat when tool approval suspends with large tool payloads.
+
+When tool approval is enabled, every approval cycle suspends the agent loop and persists a stream-state snapshot. The serialized state now keeps per-step message deltas instead of cumulative `request`/`response` per buffered step, and replaces large tool-result payloads in step content with references to the message-list copy that get rehydrated on resume. Snapshots written by older versions still resume.
+
+Fixes [#17738](https://github.com/mastra-ai/mastra/issues/17738).
--- a/.changeset/workflow-snapshot-suspend-bloat-pg.md
+++ b/.changeset/workflow-snapshot-suspend-bloat-pg.md
@@ -1,0 +1,9 @@
+---
+'@mastra/pg': patch
+---
+
+Cut per-step workflow snapshot update cost in the PostgreSQL storage adapter.
+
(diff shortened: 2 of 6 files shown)


intojhanurag:
closing this

PR #17774, opened 2026-06-10
Fix streamed user attachment reconciliation

mfrachet:
## Summary

Fixes streamed chat reconciliation in `@mastra/react` so user messages with attachments keep their text and files together while streaming, and stale empty assistant placeholders do not interfere with echoed user messages.

This is PR 1 of the stacked split. It is intentionally limited to the React chat stream accumulator/reconciliation behavior and its changeset.

## Test plan

- `pnpm --filter ./client-sdks/react test`
- `pnpm --filter ./client-sdks/react typecheck`

## ELI5 (Explain Like I'm 5)

When you send a message with text and a file, it should stay as one message bubble while the assistant responds. This PR makes sure text+attachments stay together as a single pending user bubble while streaming and that empty assistant placeholders don't break matching the server's echoed user message.

---

## Changes Overview

Fix streamed chat reconciliation in `@mastra/react` so a user turn containing attachments (text + files) is represented as a single pending message during streaming and server-echo reconciliation, and so empty trailing assistant placeholders are removed so they don't interfere with echoed user messages.

### Key Changes

- Message merging & normalization
  - Added fromCoreUserMessagesToMastraDBMessage to merge multiple CoreUserMessage inputs (text + attachments) into one MastraDBMessage.
  - Normalized file parts to the canonical V4 shape ({ type: 'file', mimeType, data }) and improved conversion to derive data, mimeType, and filename correctly.
  - Re-exported the new merge helper from the mastra-db index.

- Optimistic UI & send path
  - sendMessage now creates a single pending DB user message for an entire user turn (text + attachments) with a client-set id, mode: 'stream', status: 'pending', and CLIENT_MESSAGE_ID_KEY for correlation — replacing the prior multi-bubble optimistic behavior.
  - resolveInitialMessages now strips the transient CLIENT_MESSAGE_ID_KEY (and pending status) more reliably during reload/normalization.

- Stream reconciliation & accumulator
  - Added a helper that clears transient pending state while preserving the clientMessageId correlation key for optimistic→server id adoption.
  - data-user-message echo reconciliation now preserves the correlation id during id adoption and removes empty trailing assistant messages before appending echoed user messages.
  - finishStreamingAssistantMessage drops empty trailing assistant shells instead of leaving them.

### Tests & Validation

- Added/updated tests covering:
  - Multi-part user sends collapsing into a single pending bubble (text + file parts) with correlation id parity.
  - Reconciliation: correlation id retention across id adoption, removal of empty assistant shells when user echoes arrive, and clearing pending while keeping correlation id.
  - fromCoreUserMessagesToMastraDBMessage merging parity with single-message conversion.
  - MessageFactory reasoning streaming-state propagation.
- Test commands:
  - pnpm --filter ./client-sdks/react test
  - pnpm --filter ./client-sdks/react typecheck

### Notable Files Changed

- client-sdks/react
  - src/agent/hooks.ts (+tests)
  - src/ui/MessageFactory/types.ts (+test)
  - src/lib/mastra-db/fromCoreUserMessage.ts (+tests)
  - src/lib/mastra-db/accumulator.ts (+tests)
  - src/lib/mastra-db/index.ts
- .changeset entries for `@mastra/react` (patch)
- Numerous playground UI test and renderer additions/adjustments to exercise stream/reconciliation behavior

---

## Scope & Risk

- Scoped as PR 1 of a stacked split: changes focus on React chat stream accumulator/reconciliation behavior and associated tests/types.
- Minimal external API surface changes: one new mastra-db export is re-exported; other runtime/export signatures mostly unchanged.
- Areas requiring careful review: optimistic send/reconciliation logic and accumulator changes (medium–high review effort); tests added to validate correctness.


client-sdks/react/src/lib/mastra-db/fromCoreUserMessage.test.ts
```diff
@@ -37,17 +37,28 @@ describe('fromCoreUserMessageToMastraDBMessage', () => {
...
 
-    expect(out.content.parts).toEqual([{ type: 'file', mediaType: 'image/png', url: 'https://example.com/cat.png' }]);
+    expect(out.content.parts).toEqual([{ type: 'file', mimeType: 'image/png', data: 'https://example.com/cat.png' }]);
   });
 
-  it('defaults image mediaType to image/* when mimeType is missing', () => {
+  it('defaults image mimeType to image/* when mimeType is missing', () => {
     const input: CoreUserMessage = {
       role: 'user',
       content: [{ type: 'image', image: 'https://example.com/x' }],
```
damien-schneider, line 46: Maybe not fully related to this PR, but naming the fields image and data (see the other comment) and having them accept mainly URLs can be confusing and misleading for consumers of those objects. I'm not sure if they also accept raw data like base64, but it might be worth adding a follow-up task to either rename to imageUrl and dataUrl if the fields are URL-only, or, if they can accept multiple formats, to clarify or create two fields.

(Note: I’m writing this here to avoid losing track.)

client-sdks/react/src/lib/mastra-db/fromCoreUserMessage.test.ts
```diff
@@ -37,17 +37,28 @@ describe('fromCoreUserMessageToMastraDBMessage', () => {
     };
     const out = fromCoreUserMessageToMastraDBMessage(input);
 
-    expect(out.content.parts).toEqual([{ type: 'file', mediaType: 'image/png', url: 'https://example.com/cat.png' }]);
+    expect(out.content.parts).toEqual([{ type: 'file', mimeType: 'image/png', data: 'https://example.com/cat.png' }]);
```
damien-schneider, line 40: here for the `data`

damien-schneider (approved):

PR #17791, opened 2026-06-10
fix(core): background-dispatched sub-agents no longer send null tool content

### LINKED ISSUE #17798: Background-dispatched sub-agent delegations send null tool-message content (provider 500)
## Description

Dispatching a **sub-agent** delegation (`agent-<name>` tool) as a background task makes the supervisor's next model request fail with a provider `500`, because the placeholder tool result is serialized into a `role: "tool"` message with `null` content.

`@mastra/core` version: `1.41.0`

## Root cause

When a sub-agent invocation is dispatched as a background task, the agentic loop returns a placeholder result instead of running the tool synchronously:

```ts
// packages/core/src/loop/workflows/agentic-execution/tool-call-step.ts
return {
  result: `Background task started. Task ID: ${task.id}. The tool "${inputData.toolName}" is running in the background. ...`,
  ...inputData,
};
```

But the sub-agent tool defines a `toModelOutput` that reads `output.text` (the `agentOutputSchema` shape):

```ts
// packages/core/src/agent/agent.ts
: (output) => ({ type: 'text', value: output.text });
```

The placeholder has no `text` field, so `value` is `undefined`, producing a tool message with `null` content. Providers that validate tool content (e.g. Anthropic) reject it with `500`.

The malformed request (captured by intercepting the provider call) — note message `#6`:

```
#5 assistant  tool_calls=[Us6 (agent-opsAgent, backgrounded), LSN (a normal tool)]
#6 tool       content=NULL             tool_call_id=Us6
#7 tool       content={"success":true}  tool_call_id=LSN
```

Reproduces with and without `untilIdle`, and is unaffected by `includeSubAgentToolResultsInModelContext`.

## Reproduction

```ts
const supervisor = new Agent({
  // ...
  agents: { opsAgent },
  backgroundTasks: { tools: { opsAgent: { enabled: true } } },
});
// mastra has backgroundTasks: { enabled: true } and storage configured
await supervisor.stream('delegate something to ops', { memory: { resource, thread } });
// -> the continuation request 500s; the supervisor turn emits an `error` chunk with no text.
```

## Expected

The supervisor should acknowledge the dispatch and continue (the tool message must carry non-empty content) while the sub-agent runs in the background.

A fix is proposed in #17791.

fedesapuppo:
Fixes #17798

## Summary

Backgrounding a **sub-agent** delegation makes the supervisor's next model request fail with a provider `500`, because the placeholder tool result is serialized into a `role: "tool"` message with `null` content.

## Root cause

When a sub-agent invocation (an `agent-<name>` tool) is dispatched as a background task, the agentic loop returns a placeholder result instead of running the tool synchronously:

```ts
// packages/core/src/loop/workflows/agentic-execution/tool-call-step.ts
return {
  result: `Background task started. Task ID: ${task.id}. The tool "${inputData.toolName}" is running in the background. ...`,
  ...inputData,
};
```

But the sub-agent tool defines a `toModelOutput` that reads `output.text` (the `agentOutputSchema` shape):

```ts
// packages/core/src/agent/agent.ts
const toModelOutput = delegation?.includeSubAgentToolResultsInModelContext
  ? undefined
  : (output) => ({ type: 'text', value: output.text });
```

The background placeholder has **no `text` field**, so `value` is `undefined`, and the supervisor's next request carries a tool message with `null` content. Providers that validate tool content (e.g. Anthropic) reject it:

```json
{ "error": { "message": "Internal Server Error", "code": 500 } }
```

The exact malformed request (captured by intercepting the provider call) — note message `#6`:

```
#5 assistant  tool_calls=[Us6 (agent-opsAgent, backgrounded), LSN (a normal tool)]
#6 tool       content=NULL            tool_call_id=Us6
#7 tool       content={"success":true} tool_call_id=LSN
```

The normal tool (`#7`) serializes fine; only the backgrounded sub-agent (`#6`) is null. It reproduces with and without `untilIdle`, and is not affected by `includeSubAgentToolResultsInModelContext`.

## Fix

Fall back to the placeholder's `result` (then to an empty string) so the tool message always carries non-empty content:

```ts
value: output.text ?? (output as { result?: string }).result ?? '',
```

## Reproduction

A supervisor with a sub-agent opted into background tasks:

```ts
const supervisor = new Agent({
  // ...
  agents: { opsAgent },
  backgroundTasks: { tools: { opsAgent: { enabled: true } } },
});
// mastra has backgroundTasks: { enabled: true } and storage configured
await supervisor.stream('delegate something to ops', { memory: { resource, thread } });
// -> the continuation request 500s; the supervisor turn emits an `error` chunk with no text.
```

## Verification

Verified end-to-end in a downstream app (`@mastra/core@1.41.0`) by applying this change via `patch-package`:

- Before: the supervisor turn emits an `error` chunk (provider `500`), no acknowledgement text.
- After: all requests `200`; the supervisor acknowledges immediately (e.g. "running in the background, I'll let you know when it's done"), the background sub-agent completes, and its result is delivered out-of-band via `backgroundTaskManager.stream()`.

## Notes

I couldn't run the full monorepo test suite locally for this change. Happy to add a regression test if you can point me at the preferred layer — the cleanest spot is probably an assertion that the generated `agent-<name>` tool's `toModelOutput` handles the background-placeholder shape (`{ result }`) as well as the normal `{ text }` shape.

## ELI5 Explanation

When one agent asks another to run in the background, the supervisor sometimes got an empty reply and errored; this change ensures the supervisor always receives a non-empty placeholder string when a sub-agent is backgrounded so the supervisor continues normally.

---

### Problem
When a sub-agent (an `agent-<name>` tool) is dispatched as a background task, the agent loop passed a placeholder string (e.g., "Background task started...") to the sub-agent tool's `toModelOutput`. `toModelOutput` previously assumed `output` was an object with `text`, so for the placeholder string it produced `undefined` (serialized as `null`) for tool message content, causing provider validation failures (500) and breaking the supervisor continuation.

### Root Cause
`toModelOutput` assumed the runtime shape matched `agentOutputSchema` ({ text }). Background placeholders are handed as a plain string by the agent loop, so reading `output.text` returned undefined.

### Fix
`toModelOutput` now accepts `z.infer<typeof agentOutputSchema> | string` and uses the placeholder string directly when `output` is a string; otherwise it falls back to `output.text ?? ''`. This guarantees the serialized `role: "tool"` message always carries non-null, non-empty content when a sub-agent is backgrounded.

Example: value: typeof output === 'string' ? output : (output.text ?? '')

### Files Changed
- .changeset/fix-subagent-background-null-tool-content.md — changeset entry describing the behavioral fix
- packages/core/src/agent/agent.ts — widen `toModelOutput` input type and use placeholder-string fallback in delegated sub-agent mapping
- packages/core/src/agent/__tests__/subagent-background-tool-content.test.ts — new Vitest regression test covering backgrounded and inline delegation paths

### Verification
Locally applied (via patch-package) — before the fix supervisor continuation requests failed with 500 and emitted an error chunk; after the fix supervisor requests succeed (200), the supervisor acknowledges the background dispatch, and the sub-agent result is delivered out-of-band.

### Tests
Added a regression test that:
- Configures a supervisor with `backgroundTasks.tools.helper.enabled = true`.
- Captures supervisor prompts and asserts the backgrounded delegation produces non-empty, string tool content containing "Background task started".
- Includes a contrast test for inline delegation asserting the real sub-agent response ("Hello from the sub-agent.").


abhiaiyer91:
Thanks for this @fedesapuppo , can you add a test!

fedesapuppo:
Done, added a regression test in `packages/core/src/agent/__tests__/subagent-background-tool-content.test.ts` and read CONTRIBUTING/DEVELOPMENT.

The test wires a supervisor with one sub-agent opted into background dispatch (`backgroundTasks.tools`), runs the delegation, and asserts the supervisor's continuation turn carries non-empty `tool` content for the backgrounded delegation. It fails on `main` (content is undefined -> null) and passes with the fix. A second test covers the inline (non-background) path as a contrast. Verified locally with `pnpm build:core` + `vitest run` on the file (and it stays green alongside `supervisor-integration` and `stream-until-idle`).

Writing the test surfaced that `toModelOutput` actually receives the background placeholder as a plain string ("Background task started..."), not a `{ result }` object, so I simplified the fix to use the string directly: `value: typeof output === 'string' ? output : (output.text ?? '')`. Updated the changeset to match.

TheIsrael1:
## 🟠 High: Maintainer notes

### Scope
`@mastra/core` Agent sub-agent delegation when the delegated `agent-<name>` tool is dispatched as a background task.

### Context
PR #17791 directly closes #17798. The linked issue includes a minimal reproduction where the unpatched release sends a tool-result with `type: "text"` but no `value`, which can surface as provider 500s on the continuation request.

### Evidence checked
- PR diff touches only the core agent mapper, a focused regression test file, and a core patch changeset.
- `packages/core/src/agent/agent.ts` preserves background placeholder strings as text output while keeping current `SubAgentToolOutput` typing from `main`.
- `packages/core/src/agent/__tests__/subagent-background-tool-content.test.ts` covers both the background-dispatched placeholder path and the inline sub-agent path.
- Focused regression passed locally: `pnpm --filter ./packages/core exec vitest run src/agent/__tests__/subagent-background-tool-content.test.ts --exclude '**/tool-builder/**'`.
- Package check currently fails on unrelated `src/workflows/workflow-schema-types.test-d.ts:52` (`inputData.count` possibly undefined).
- CodeRabbit review threads are resolved; only unrelated Vercel status contexts were failing during triage.

### Observations
- The fix matches the linked root cause and keeps the behavior change localized to the sub-agent tool model-output adapter.
- Regression coverage exercises both the failing background path and the non-background path.
- Maintainer conflict resolution has been applied and pushed in merge commit `998b37a495`.

taofeeq-deru (approved):

Issue #17789, opened 2026-06-10
DurableAgent: run terminates when every tool call in a step errors — error results never reach the model


BrandiATMuhkuh:
## Package & version

`@mastra/core` 1.38.0 (verified unchanged through 1.41.0 and current `main`).

## Bug

In the durable agentic loop, the LLM-mapping step terminates the run when **all** tool calls of a step errored — and since most steps issue a single tool call, one throwing tool kills the whole run:

https://github.com/mastra-ai/mastra/blob/cff40d473842969a020616dad7b9ccc80fc391dd/packages/core/src/agent/durable/workflows/steps/llm-mapping.ts#L125-L127

```ts
const allToolsErrored = toolResults.length > 0 && toolResults.every(r => r.error !== undefined);
const allToolsNotFound = allToolsErrored && toolResults.every(r => r.error?.name === 'ToolNotFoundError');
const isContinued = llmOutput.stepResult.isContinued && (!allToolsErrored || allToolsNotFound);
```

The error results are already added to the message list a few lines above (`Error: ${toolResult.error.message}` tool-invocation parts) — but because `isContinued` is forced `false`, the dowhile loop in `createDurableAgenticWorkflow` exits and **the model never sees them**. The stream finishes \"normally\" with the assistant message ending on a tool part and no final text.

This diverges from the regular (non-durable) `Agent.stream()` loop, which feeds tool error results back to the model so it can retry or self-correct — one of the core agentic behaviours. Wrapping an agent with `createDurableAgent` silently changes error semantics: any transient tool failure (model-chosen bad argument, upstream 500, timeout) ends the turn mid-work instead of letting the model recover.

`ToolNotFoundError` is already exempted (`allToolsNotFound`), so the loop trusts the model to recover from calling a nonexistent tool — but not from a tool that exists and threw. The asymmetry seems unintentional.

## Repro

1. Wrap any agent with `createDurableAgent` and give it a tool that throws (e.g. validation error on a bad argument).
2. Prompt so the model calls that tool as the only tool of a step.
3. The run ends right after the tool result: no further LLM call, no final text, `FINISH` emitted as if complete.

With the same agent un-wrapped, `Agent.stream()` continues: the model receives the `Error: …` result, retries with corrected arguments, and finishes with text.

## Expected

[...]

PR #17794, opened 2026-06-10
fix(core): honor wrapped agent defaultOptions in DurableAgent

### LINKED ISSUE #17790: DurableAgent.stream() ignores the wrapped agent's defaultOptions — maxSteps silently falls to 5, providerOptions dropped
## Package & version

`@mastra/core` 1.38.0 (verified unchanged through 1.41.0 and current `main`).

## Bug

The regular `Agent.stream()/generate()` deep-merges the agent's `defaultOptions` into every call:

```ts
const defaultOptions = await this.getDefaultOptions({ requestContext: options?.requestContext });
const mergedOptions = deepMerge(defaultOptions, options ?? {});
```

`DurableAgent.stream()` does not: it passes the caller's per-request options straight into `prepareForDurableExecution`, which serializes only `execOptions?.maxSteps`, `execOptions?.providerOptions`, etc. — the wrapped agent's `defaultOptions` are never consulted:

https://github.com/mastra-ai/mastra/blob/cff40d473842969a020616dad7b9ccc80fc391dd/packages/core/src/agent/durable/durable-agent.ts

So for an agent configured with e.g.

```ts
const base = new Agent({
  // ...
  defaultOptions: {
    maxSteps: 250,
    providerOptions: { anthropic: { thinking: { type: "adaptive" }, effort: "high" } },
  },
});
export const agent = createDurableAgent({ agent: base });
```

wrapping with `createDurableAgent` silently changes behaviour:

1. `maxSteps` falls back to `DurableAgentDefaults.MAX_STEPS = 5` — runs stop mid-work after 5 LLM iterations (we hit this in production: every chat turn needing >5 steps was cut off).
2. `providerOptions` are dropped — the Anthropic thinking config stopped being sent, with no error or warning.

`DurableAgent` also doesn't delegate `getDefaultOptions()` to the wrapped agent (the delegate list covers `getModel`/`getInstructions`/`listTools`/`getMemory`/`getVoice`), so even route-level code that merges `agent.getDefaultOptions()` gets `{}`.

## Expected

`DurableAgent.stream()` (and `prepareForDurableExecution`) should merge `wrappedAgent.getDefaultOptions()` under the per-request options, like the regular execution paths do — or at minimum `DurableAgent` should delegate `getDefaultOptions()` to the wrapped agent so callers can do the merge themselves.

## Workaround

Pass `maxSteps` explicitly to `createDurableAgent({ agent, maxSteps })` and duplicate `maxSteps`/`providerOptions` into the route-level `defaultOptions` of `chatRoute` (`@mastra/ai-sdk`), which does merge them into the per-request options.

JigyasuRajput:
## Description

The regular `Agent.stream()` and `generate()` deep-merge the agent's `defaultOptions` into every call. `DurableAgent` did not, so wrapping an agent with `createDurableAgent` silently dropped its configured defaults. `maxSteps` fell back to the durable default of 5 (runs that need more than 5 steps get cut off mid-work) and `providerOptions` such as the Anthropic thinking config stopped being sent, with no error or warning.

This wires the durable path up to behave like the normal one:

- `prepareForDurableExecution` now resolves the wrapped agent's `defaultOptions` and merges them under the per-request options before it serializes the workflow input. Per-request options still win, and nested objects like `providerOptions` merge instead of getting clobbered. This is the single place both `stream()` and `prepare()` flow through, so both are covered.
- `DurableAgent` now delegates `getDefaultOptions()` to the wrapped agent (it already delegates `getModel`, `getInstructions`, `listTools`, and friends), so route-level code that reads `agent.getDefaultOptions()` gets the real values back instead of an empty object.

Before:

```ts
const base = new Agent({
  model,
  defaultOptions: {
    maxSteps: 250,
    providerOptions: { anthropic: { thinking: { type: 'adaptive' }, effort: 'high' } },
  },
});

const agent = createDurableAgent({ agent: base });
// runs stopped after 5 steps and providerOptions were never sent
```

After this change the defaults are honored, and an explicit per-request value still takes precedence over them.

One thing worth calling out: because the merge now happens up front, `defaultOptions.versions` also flows into the request-context version merge, which lines the durable path up with the non-durable one. There are tests covering that too.

### How it was tested

Added `durable-agent-default-options.test.ts`, which checks the `getDefaultOptions` delegation, that `maxSteps` and `providerOptions` make it into the serialized workflow input, per-request precedence, nested `providerOptions` merging, function-valued `defaultOptions`, the version-override path, and an end-to-end run that drives the agentic loop past 5 steps using only `defaultOptions.maxSteps`. I also confirmed these fail without the fix (the e2e one caps at 5) and pass with it. The full durable suite stays green.

## Related issue(s)

Fixes #17790

## Type of change

- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [x] Test update

## Checklist

## ELI5

DurableAgent was ignoring the settings (like `maxSteps` and `providerOptions`) that its wrapped agent wanted to use, causing those settings to silently disappear. This PR fixes it by making DurableAgent properly respect and use those settings, the same way a regular Agent does.

## Overview

This PR fixes DurableAgent to properly honor the wrapped agent's `defaultOptions` during durable workflow execution. Previously, these defaults were silently dropped, causing observable bugs like premature termination at 5 steps and loss of provider-specific configuration.

### Core Fixes

**`packages/core/src/agent/durable/durable-agent.ts`**
- Added `getDefaultOptions()` method that delegates to the wrapped agent
- Aligns durable agent behavior with the underlying agent's default options exposure

**`packages/core/src/agent/durable/preparation.ts`**
- Modified `prepareForDurableExecution` to resolve and deep-merge wrapped agent's `defaultOptions` with per-request options
- Per-request options maintain precedence while nested objects (e.g., `providerOptions`) are merged rather than replaced
- Ensures `defaultOptions.versions` flows into request-context version merge

### Testing

**`packages/core/src/agent/durable/__tests__/durable-agent-default-options.test.ts`** (new)
- Verifies `getDefaultOptions()` delegation from DurableAgent to wrapped agent
- Tests serialization of wrapped-agent `defaultOptions` (including `maxSteps` and nested `providerOptions`) into workflow input
- Validates per-request option precedence through deep merge semantics
- Confirms function-valued `defaultOptions` resolve correctly with RequestContext
- Tests version override merging behavior
- Includes end-to-end streaming test demonstrating agentic loops exceeding the previous 5-step durable limit

### Documentation

**`.changeset/tired-tables-start.md`**
- Documents the fix with before/after examples showing `maxSteps` and provider configuration now being honored

## Impact

- Removes need for workarounds (duplicating defaults in `createDurableAgent` or route-level defaults)
- Achieves parity between DurableAgent and regular Agent paths for option handling
- Fixes silent failures where maxSteps and providerOptions were dropped during durable execution

Fixes `#17790`


JigyasuRajput:
Hey @taofeeq-deru @wardpeet, when you get a chance, would you mind reviewing these two DurableAgent bug fixes? CodeRabbit flagged you as the suggested reviewers. cc @roaminro, who triaged both linked issues.

- #17794 honor the wrapped agent's defaultOptions in DurableAgent (fixes #17790)
- #17801 let DurableAgent recover from tool errors instead of ending the run (fixes #17789)

Both have green CI and are linked to their issues. The only red checks are the two Vercel deploy previews.

No rush, just flagging since they've been open about a week. Happy to make any changes you'd want.

abhiaiyer91 (approved):

JigyasuRajput:
btw @abhiaiyer91 hey!

just curious...
in case you like my work, are you currently hiring?

PR #17796, opened 2026-06-10
fix(agent): apply toModelOutput transform to client tool results in sendToolApproval continuation

### LINKED ISSUE #17792: Can client tools (the `clientTools` passed to `agent.generate()`) return multimodal output?
This issue was created from Discord post 1514319098966966312:

[Open in Discord](https://discord.com/channels/1309558646228779139/1514319098966966312)

I've been looking into multimodal function responses from [Gemini](https://ai.google.dev/gemini-api/docs/function-calling#multimodal) . This seems to work fine when using **server-side** tool definitions and a small `toModelOutput()` helper. However, the same tool defined under `clientTools` doesn't seem to work. 

i.e. a server-tool call like this would produce the following output* when dumping chat history - 

```js
const imageTool = createTool({
  id: "aToolThatReturnsAnImage",
  execute: async () => {
    const jpeg = await getAnImage("...");
    return { ok: true, _b64: jpeg.toString("base64") };
  },
  toModelOutput: (output) =>
    ({
      type: "content",
      value: [
        {
          type: "image-data",
          data: (output as { _b64: string })._b64,
          mediaType: "image/jpeg",
        },
      ],
    }),
});
```

**truncated for brevity*
```json
"functionResponse": {
  "id": "l9brzb2y",
  "name": "aToolThatReturnsAnImage",
  "parts": [
    {
      "inlineData": {
        "data": "<TRUNCATED>",
        "mimeType": "image/jpeg"
      }
    }
  ],
  "response": {
    "content": "Here is the current screenshot.",
    "name": "aToolThatReturnsAnImage"
  }
}
}```

A **client tool** (same shape, executed client-side and returned over HTTP) never produces `inlineData`. The whole content array ends up stringified inside `functionResponse.response.content` as opaque JSON, so the model just sees text.

**Client Tool**
```json
"functionResponse": {
  "id": "o7pu6myi",
  "name": "aToolThatReturnsAnImage",
  "response": {
    "content": [
      {
        "text": "Here is the current screenshot.",
        "type": "text"
      },
      {
        "data": "<TRUCATED>",
        "mediaType": "image/jpeg",
        "type": "image-data"
      }
    ],
    "name": "aToolThatReturnsAnImage"
  }
}```

Aayush-engineer:
## Description

Fixes a bug where `toModelOutput` was never applied to client tool results, causing multimodal output (images, files) to be stringified as opaque JSON instead of being sent as proper structured content (e.g. `inlineData`) to the LLM.

### Root Cause

For **server tools**, the flow goes through `llm-mapping-step.ts` → `getProviderMetadataWithModelOutput()` which runs `toModelOutput` and stores the result in `providerMetadata.mastra.modelOutput`. The message-list then applies this when building LLM context.

For **client tools**, there is no server-side execution. The client executes the tool and sends the result back via `sendToolApproval` as pre-formed messages with `output: { type: 'json', value: result }`. These messages go directly into `continueWithMessages` → `agent.stream()` — completely bypassing the `toModelOutput` transform.

This is why the same tool defined as a server tool produces correct `inlineData` in the Gemini function response, but defined as a `clientTool` produces a stringified content array instead.

### Fix

Added `#applyToModelOutputToMessages()` — a private method on `Agent` that:
1. Walks incoming messages looking for `tool-result` parts
2. Looks up each tool by name from the agent's tool registry (using the caller's `requestContext` so function-based tool configs are resolved correctly)
3. If the tool has `toModelOutput`, runs it on the raw result value
4. Replaces the `output` field with the transformed multimodal value

This method is called in `sendToolApproval` before `continueWithMessages`, intercepting client tool results at the only point where they enter the server-side pipeline.

### Behavior

- Tools **without** `toModelOutput`: unchanged, pass through as-is
- Tools **with** `toModelOutput`: transform applied, multimodal output reaches LLM correctly
- If `toModelOutput` throws: error is caught silently, original part returned (best-effort, non-breaking)
- Works for both `agent.stream()` and `agent.generate()` with `clientTools`

### Files Changed

- `packages/core/src/agent/agent.ts` — added `#applyToModelOutputToMessages()`, called in `sendToolApproval`
- `.changeset/client-tool-toModelOutput-continuation.md` — patch changeset for `@mastra/core`

## Related issue(s)

Fixes #17792

## Type of change

## Checklist

## ELI5
When a client tool returns an image/file, the system used to accidentally turn it into a plain JSON string, so the AI couldn’t understand it as an actual image/file. This PR makes sure the same “convert tool output into model-ready multimodal format” logic runs for client tools too.

## Summary
This PR fixes multimodal outputs (images/files) from `clientTools` by applying each tool’s `toModelOutput` transform to `tool-result` messages before they re-enter the server-side LLM pipeline. It prevents multimodal payloads from being stringified as JSON and ensures the model receives structured multimodal content (e.g., Gemini `inlineData`-compatible shapes).

## Root cause
Server-side tool execution runs `toModelOutput` during the `llm-mapping-step` path, producing normalized multimodal output. Client tools execute on the client and send back tool results already packaged in messages, bypassing that server-side transform—so multimodal content was being delivered to the model as opaque JSON text.

## Fix
- Added `Agent.#applyToModelOutputToMessages(...)`:
  - Walks incoming `MessageListInput` for `tool-result` parts.
  - Resolves the corresponding tool by name from the agent registry, merging per-call `streamOptions.clientTools` (“extraTools”) with registry tools.
  - Extracts the raw result (supports `{ type: 'json', value }` wrappers or direct values).
  - Runs `computeToolModelOutput` (which applies `toModelOutput` and normalizes multimodal shapes) and replaces the message part’s `output` when a transformed value is produced.
  - If the tool has no `toModelOutput`, the result is nullish, or transformation throws, it preserves the original part (best-effort, non-breaking).
- Updated `sendToolApproval` to call this transformer before `continueWithMessages`, covering both `agent.stream()` and `agent.generate()` flows for `clientTools`.
- Refactored shared logic into a new `packages/core/src/tools/model-output.ts` module:
  - `normalizeModelOutput` and `computeToolModelOutput` are now used by both server mapping and the new client-tool message continuation path.
- Updated `llm-mapping-step.ts` to delegate to `computeToolModelOutput` (removing duplicated normalization/transform code).
- Added `model-output.test.ts` to validate normalization, transform behavior, and observability span handling.

## Behavior and impact
- Client tools **without** `toModelOutput` behave unchanged.
- Client tools **with** `toModelOutput` now produce properly structured multimodal content for the model instead of JSON-stringified payloads.
- `toModelOutput`/normalization errors are caught during continuation; the original tool result is kept to avoid breaking execution.
- Includes release notes for the fix.

## Files changed
- `packages/core/src/agent/agent.ts`
- `packages/core/src/loop/workflows/agentic-execution/llm-mapping-step.ts`
- `packages/core/src/tools/model-output.ts`
- `packages/core/src/tools/model-output.test.ts`
- `.changeset/client-tool-toModelOutput-continuation.md`


Aayush-engineer:
Fixed in the latest commit. `#applyToModelOutputToMessages` now accepts an `extraTools` parameter and merges `streamOptions.clientTools` with the agent registry tools before lookup, so per-call client tools with `toModelOutput` are correctly resolved.

Aayush-engineer:
@abhiaiyer91 @TylerBarnes CodeRabbit comments addressed ready for review when you get a chance!

Aayush-engineer:
@abhiaiyer91 @TylerBarnes wanted to flag that #17825 (Toby's client-js PR) and this PR approach the problem from different layers. This PR fixes the server-side sendToolApproval continuation path in @mastra/core. #17825 fixes the client-js SDK execution paths. Happy to close this if the team prefers the client-side-only approach, or keep both if the server-side path also needs fixing.

wardpeet (dismissed):

PR #17800, opened 2026-06-10
fix(memory): preserve tool-call messages when client-side tool result is pending

### LINKED ISSUE #15244: [BUG] Client-side tool results lost when Observational Memory is enabled
### Describe the Bug

When Observational Memory (OM) is enabled, client-side tool (tool without `execute` function) execution results are lost on the next request. The agent cannot see the tool result and acts as if it never arrived.

OM's `filterObservedMessages` removes messages that have been "observed", but has no awareness of tool-call / tool-result pairing. On the second request (when the client sends back the tool result), OM removes the tool-call message because its ID is in `observedMessageIds`, leaving the tool-result orphaned — the LLM sees a result without the corresponding call.

This differs from #14282 which only fixed **provider-executed tools** (e.g. Anthropic `web_search`) by deferring observation when `state: 'call'` is detected mid-stream. Client-side tools return in a separate HTTP request where the call already has `state: 'result'`, so that guard doesn't help.

### Steps To Reproduce

1. Create an agent with OM enabled and a client-side tool:

```ts
const agent = new Agent({
  name: 'test-agent',
  model: openai('gpt-4o'),
  tools: {
    myTool: createTool({
      id: 'my-tool',
      description: 'A client-side tool',
      inputSchema: z.object({ query: z.string() }),
      // no execute → client-side tool
    }),
  },
  memory: new Memory({
    options: { observationalMemory: { enabled: true } },
  }),
});
```

2. Start a conversation. The agent generates a tool call (`state: 'call'`) and returns it to the client.

3. Client executes the tool locally and sends the result back in a new request (messages include the tool-call + tool-result).

4. **Expected**: Agent sees the tool result and continues the conversation.

5. **Actual**: Agent does not see the tool result — it's effectively lost due to OM filtering.

/

### Expected Behavior

The agent should see the client-side tool result and continue the conversation normally. `filterObservedMessages` (`packages/memory/src/processors/observational-memory/message-utils.ts`) should check for tool-call/tool-result pair integrity before removing messages. If a message containing a `tool-invocation` part is about to be removed, the filter should verify that no remaining message has a `tool-result` part referencing the same `toolCallId`. If an orphaned result would be created, either keep the tool-call message or remove both.

**Workaround**: Converting client-side tools to server-side tools (adding an `execute` function) avoids this, since results are resolved within the same request's agentic loop before OM's cross-request filtering kicks in.

### Environment Information

```block
- `@mastra/core`: >= 1.22.0
- `@mastra/memory`: >= 1.13.0
```

### Verification

rodboev:
## Summary

When ObservationalMemory is enabled, client-side tools (tools without an `execute` function) break on the second request. The assistant message containing the tool-call is in `observedMessageIds`, so `filterObservedMessages` removes it — but the client's tool-result reply arrives in the same request and has no matching call to anchor it. The model receives an orphaned `tool-result` and loses the entire tool exchange silently.

## Root cause

`filterObservedMessages` in `packages/memory/src/processors/observational-memory/message-utils.ts` has two pruning paths, and both could split a tool-call/result pair.

- In the record-based fallback path, observed messages were removed by ID with no awareness of a matching `state: 'result'` part that was still present in a non-observed message.
- In the marker-boundary path used on step 0, messages before the latest completed observation marker were removed by position, and the marker-bearing message itself was trimmed to post-marker parts only. That could still orphan a call when the matching result lived after the marker, either in a later message or in the same marker-bearing message.

On the second request for a client-side tool, the previous turn's assistant message can already be observed while the current turn's tool-result is still live. After pruning, the model sees the result without its matching call, and the exchange is lost.

The fix teaches both pruning paths to preserve call/result pairs until the matching result is no longer pending.

## Changes

- `packages/memory/src/processors/observational-memory/message-utils.ts`: add shared helpers that collect pending tool-result IDs, preserve observed call messages that still have a live matching result, scope marker-mode preservation to post-marker content, and keep paired pre-marker call parts when trimming the marker-bearing message
- `packages/memory/src/processors/observational-memory/__tests__/observational-memory.test.ts`: add marker-pruning and record-fallback regressions for pending-result preservation, already-observed-result removal, fallback cursor pruning, and fallback time pruning

## What is NOT changed

Behavior for text-only messages, the `preserveMessageIds` escape hatch, and all other `filterObservedMessages` callers is unchanged. No exported type or function signatures change.

## Out of scope

This PR does not change how ObservationalMemory decides whether to observe, buffer, or activate. It only fixes how already-observed content is pruned from the live message context once tool-call/result pairs are present.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## Checklist

- [x] Linked related issue
- [x] Added or updated tests

## Test plan

- `pnpm exec vitest run src/processors/observational-memory/__tests__/observational-memory.test.ts` — 455 passing, 1 todo. Covers the marker-pruning and record-fallback tool-call/result regressions, including fallback cursor and time pruning.

## ELI5

When the assistant asks a tool for information, it must keep the question with the tool's answer. This PR prevents memory pruning from removing the question while the answer is still pending.

## Problem

`filterObservedMessages` could remove an observed client-side tool-call message while its matching tool result remained unobserved. This left the agent with an orphaned result.

## Solution

- Preserve tool-call/result pairs when the result is still pending.
- Apply this behavior to marker-boundary pruning and record-based fallback pruning.
- Retain paired call parts when trimming marker-bearing messages.
- Preserve existing text-only behavior, `preserveMessageIds`, callers, and exported APIs.

## Changes

- Updated `message-utils.ts` with internal helpers and pair-aware pruning logic.
- Added regression tests for marker pruning, cursor pruning, timestamp pruning, observed pairs, and standalone tool calls.
- Added a patch changeset for `@mastra/memory`.

## Impact

- Prevents pending client-side tool results from losing their matching tool calls.
- Provider-executed tool behavior remains unchanged.
- Tests: 455 passing and 1 todo.

## Upstream

Closes #15244.
Reported by @Onweekendd.


rodboev:
Addressed CodeRabbit’s fallback pruning finding. The record-based fallback now applies the same pending tool-result pair guard before cursor and lastObservedAt pruning, so it no longer removes a tool-call message while its matching result is still live. I also added focused regression coverage for those fallback branches.

Verification:
`pnpm exec vitest run src/processors/observational-memory/__tests__/observational-memory.test.ts` — 455 passing, 1 todo.

abhiaiyer91 (approved):
Thanks @rodboev — excellent fix for a nasty high-impact bug (#15244). Review notes:

**Root cause & fix are exactly right.** Both pruning paths in `filterObservedMessages` could orphan a client-side tool-result by removing its observed call message. The fix makes pruning pair-aware in both:
- **Marker-boundary path:** pending result IDs are collected from the marker message's unobserved parts + everything after the marker; pre-marker call messages with a live matching result survive, and the marker-bearing message keeps paired pre-marker call parts ahead of its unobserved parts (call-before-result ordering preserved).
- **Record-fallback path:** the pending-pair guard is applied consistently to all three removal branches (observed-ID, cursor, and time-based) — a naive fix would've only patched the first.

**No retention leak:** preserved call messages are only kept while the result is unobserved; once OM observes the result the guard stops matching and normal pruning resumes — covered by the removal regressions.

**Maintainer changes pushed to this branch:** brought the branch up to date with main (clean merge, no conflicts), and a small style refactor replacing `as any` part access with typed narrowing via a shared `getToolInvocationId` helper, matching neighboring OM code. No behavior change.

Verified locally on current main: OM suite 471 passed / 1 todo (including the 6 new regressions), `tsc --noEmit` clean. LGTM 🚀

PR #17801, opened 2026-06-10
fix(core): let DurableAgent recover from tool errors instead of ending the run

### LINKED ISSUE #17789: DurableAgent: run terminates when every tool call in a step errors — error results never reach the model

JigyasuRajput:
## Description

In the durable agentic loop, the LLM-mapping step ended the run whenever every tool call in a step errored. Because most steps issue a single tool call, one throwing tool killed the whole turn. The error result was already written to the message list, but `isContinued` was forced to `false`, so the `dowhile` loop in `createDurableAgenticWorkflow` exited before the model ever saw it. The stream finished "normally" with the assistant message ending on a tool part and no final text.

This is different from how the regular `Agent.stream()` loop behaves. That one keeps going and feeds tool error results back to the model so it can retry or self-correct, which is one of the core agentic behaviours. Wrapping an agent with `createDurableAgent` quietly changed the error semantics, so any transient tool failure (a model-chosen bad argument, an upstream 500, a timeout) ended the turn mid-work. `ToolNotFoundError` was already exempted, so the loop trusted the model to recover from calling a missing tool but not from a tool that exists and threw, which looks unintentional.

The fix keeps the LLM step's own `isContinued` (it already respects `finishReason`) instead of overriding it on tool errors:

```ts
// before
const allToolsErrored = toolResults.length > 0 && toolResults.every(r => r.error !== undefined);
const allToolsNotFound = allToolsErrored && toolResults.every(r => r.error?.name === 'ToolNotFoundError');
const isContinued = llmOutput.stepResult.isContinued && (!allToolsErrored || allToolsNotFound);

// after
const isContinued = llmOutput.stepResult.isContinued;
```

The error results are already in the message list, so the model sees them on the next step and can correct itself, matching the non-durable loop. `maxSteps` still bounds a tool that keeps failing, so this does not open the door to runaway loops.

### How it was tested

Added `durable-agent-tool-error-recovery.test.ts` with two end-to-end cases that drive the durable stream with a mock model and a throwing tool:

- A tool errors on the first step, and the run continues: the model is called again, the error message shows up in the prompt of that second call, and the run finishes with the model's recovery text instead of ending on the tool part.
- A tool that always throws is bounded by `maxSteps` rather than looping forever.

Both fail against the current code (the loop stops after the first errored step) and pass with the fix. The full durable suite stays green.

## Related issue(s)

Fixes #17789

## Type of change

- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [x] Test update

## Checklist

## ELI5

Imagine you're asking an assistant to do a task, and it tries using a tool but the tool fails. The old version of DurableAgent would just give up right there and stop. The new version lets the assistant see what went wrong and try again (or ask for help differently), just like the regular Agent does—but it still won't retry forever, so it doesn't get stuck.

## Problem

The DurableAgent loop was forcing the "continue" flag to `false` whenever all tool calls in a step failed, causing the run to terminate immediately. This prevented the model from seeing the error message and attempting recovery or self-correction. The behavior diverged from the non-durable `Agent.stream()` loop, which feeds tool error results back to the model. While `ToolNotFoundError` was already exempted, other tool errors would end the durable run prematurely.

## Solution

The fix removes the logic that overrides the LLM step's `isContinued` decision based on tool error outcomes. Now the step's "continue" flag is taken directly from `llmOutput.stepResult.isContinued`, allowing error results to flow back to the model on the next iteration. The `maxSteps` parameter remains as a safeguard against infinite retry loops, ensuring bounded behavior while permitting recovery from transient tool failures.

## Changes

- **`.changeset/fine-trees-deny.md`**: Documents the patch to `@mastra/core` fixing premature run termination on tool errors.

- **`packages/core/src/agent/durable/__tests__/durable-agent-tool-error-recovery.test.ts`** (new file): Adds regression test coverage for issue `#17789` with two test cases:
  - Verifies the durable agent continues after a tool error and feeds the error back to the model for recovery
  - Verifies bounded behavior when a tool persistently fails by confirming `maxSteps` prevents infinite loops

- **`packages/core/src/agent/durable/workflows/steps/llm-mapping.ts`**: Simplifies the continuation logic in `createDurableLLMMappingStep` to use `llmOutput.stepResult.isContinued` directly without gating based on tool error presence or type.

## Related Issue

Fixes `#17789`


JigyasuRajput:
lmk if there's anything from my end 

thanks 🙌🏻

TheIsrael1:
## 🟠 Maintainer's Triage Note

**Current Phase:** Reviewed — close as superseded

**Triage:**
- Type: bug — DurableAgent can terminate after a step where every tool call errors, preventing the model from seeing tool error results and recovering.
- Maintainer read: Issue #17789 reports a core DurableAgent behavior difference from regular `Agent.stream()`; PR #17801 was directionally correct and targeted the right durable LLM mapping path.
- Next Steps: Close PR #17801 as superseded by #18881. Issue #17789 has been commented as fixed on main.
- Severity: 🟠 high — A transient or model-correctable tool error can end a durable agent run mid-turn in a core agentic workflow; the issue is labeled `impact:high`.
- Confidence: High

**Review:**
- Status: complete
- Findings: PR #17801 conceptually fixes the reported issue by allowing DurableAgent to continue after tool errors so the model can see the error result and recover, with focused tests for recovery and `maxSteps` bounding. However, the PR is currently `DIRTY`, and newer merged work in #18881 already fixed the same durable tool-error recovery path on main with broader durable scenario coverage and a core changeset.
- Confidence: High
- Follow-up: No approval path needed for #17801 unless the author rebases to contribute only distinct regression coverage not already covered by #18881.

**Approve:**
- Status: not approved
- Final approver: Not applicable
- Notes: Recommended outcome is close as superseded by #18881, not merge as-is.

TheIsrael1 (commented):
Thanks for the contribution — the direction here is right, and the added regression coverage matches the bug reported in #17789.

I don't think we should merge this PR as-is now because it has been superseded by newer durable-agent parity work on main. GitHub currently reports this PR as dirty, and #18881 already changed the same durable LLM mapping path to keep the agentic loop going after tool errors () with broader durable scenario coverage and a core changeset for tool-error recovery.

CI is otherwise green for the relevant code checks; the Vercel failures are authorization-required preview failures rather than code failures.

Recommendation: close this as superseded by #18881, unless you want to rebase and keep only any targeted regression test coverage that is still distinct from the newer scenario coverage.

PR #17802, opened 2026-06-10
feat(playground): auto-grow chat composer input up to max height

damien-schneider:
The chat composer textarea in Studio was stuck at two lines, so longer messages were cramped into a tiny scrolling box.

It now grows with the content via `field-sizing-content` and is wrapped in the design system `ScrollArea`, which caps the height at ~8 lines and fades the clipped edges once the content overflows. The initial two-line height is preserved with a `min-h`, and since sizing is pure CSS the input shrinks back automatically when the message is sent. Browsers without `field-sizing` support keep the previous fixed-height behavior.

## ELI5

The PR makes the chat message input box in Mastra Studio grow taller as you type longer messages, instead of keeping it cramped at a fixed size. Once it reaches a maximum height of about eight lines, it becomes scrollable with a nice fade effect at the edges to show there's more text below.

## Summary

This PR improves the chat composer user experience in Mastra Studio by implementing auto-growing textarea behavior. 

**Key Changes:**
- The message input textarea now automatically grows with its content using the CSS `field-sizing-content` property, expanding to fit multi-line messages
- The textarea is wrapped in a `ScrollArea` component with a `maxHeight` of 212px (~8 lines), which caps the visible height and applies a subtle fade effect to clipped edges when content overflows
- A `min-h-17` class preserves the original two-line minimum height
- The sizing is implemented entirely in CSS, allowing the input to automatically shrink back to its minimum height when a message is sent
- For browsers that don't support `field-sizing`, the behavior gracefully falls back to the previous fixed-height behavior

**Files Modified:**
- `.changeset/composer-textarea-autogrow.md` – Added changeset entry documenting the improvement as a patch release for `@internal/playground`
- `packages/playground/src/lib/ai-ui/thread.tsx` – Updated the `Composer` component to import and use `ScrollArea`, wrapping the textarea with appropriate styling and inline documentation

The change maintains all existing functionality (placeholder text, permission gating, send-on-Enter key handling) while providing a better user experience for composing longer messages.


mfrachet (approved):

damien-schneider:
Note on the red check: `E2E kitchen-sink (1/3)` › `tool stream` is a known main-wide failure, unrelated to this PR. The kitchen-sink `weatherInfo` fixture fetches the real open-meteo API, which currently fails on CI runners, so `tool-result` never renders. It fails identically on unrelated branches and a fix (mocked weather response) is already in flight in #17774. This check is not required; all required checks (Lint, Merge Test Reports) pass.

PR #12570, opened 2026-02-01
fix(core): support Zod .transform() on createTool outputSchema

abhiaiyer91:
Fixes #12426

When using `createTool` with a Zod `.transform()` on `outputSchema`, TypeScript was forcing `execute()` to return the post-transform type — requiring `as any` to compile. The runtime was already correct (transforms are applied via `safeParse` during validation), so this was purely a typing issue.

**Before:**
```ts
const tool = createTool({
  id: 'list-users',
  outputSchema: z.object({ data: z.array(...) }).transform(resp => ({
    activeUsers: resp.data.filter(u => u.acc_status === 'A')
  })),
  execute: async () => {
    return { data: [...] } as any; // required to avoid type error
  },
});
```

**After:**
```ts
const tool = createTool({
  id: 'list-users',
  outputSchema: z.object({ data: z.array(...) }).transform(resp => ({
    activeUsers: resp.data.filter(u => u.acc_status === 'A')
  })),
  execute: async () => {
    return { data: [...] }; // works — execute accepts pre-transform type
  },
});
```

Two changes make this work: relaxed `ZodLikeSchema` to accept `ZodEffects` (from `.transform()`), and added a `TSchemaOutInput` type parameter to `ToolAction` so `execute()` independently infers the pre-transform return type. Non-transformed schemas are unaffected since input and output types are identical.

## Release Notes

* **Bug Fixes**
  * Resolved issue where createTool required type assertions for Zod schemas with .transform() on outputSchema.
  * Enhanced type inference for tool execution to correctly handle pre-transform and post-transform type differences.

* **Tests**
  * Added comprehensive test coverage for output schema transformations, including both type-level and runtime validation scenarios.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

PR #17807, opened 2026-06-10
fix(core): use pre-transform type for createTool execute return

### LINKED ISSUE #12426: [BUG] createTool execute() expects POST-transform type instead of PRE-transform type
### Describe the Bug

When using `createTool` with a Zod schema that includes `.transform()`, TypeScript incorrectly expects the `execute` function to return the POST-transform type (OUTPUT) instead of the PRE-transform type (INPUT). This forces developers to use `as any` to bypass the false type error, even though the runtime behavior is correct.

**Why transforms are crucial for LLM agents:** Many APIs return data with cryptic codes and abbreviated field names that make it difficult for LLMs to reason about:
- Cryptic status codes like `"A"`, `"I"`, `"S"` instead of `"Active"`, `"Inactive"`, `"Suspended"`
- Abbreviated keys like `f_name`, `l_name`, `acc_status` instead of clear names
- Generic keys like `data`, `result`, `items` with no semantic meaning

Transforms allow developers to convert these API responses into semantically clear, LLM-friendly formats. For example:
- `"A"` → `"Active"` (expand cryptic codes into readable words)
- `f_name` + `l_name` → `fullName` (combine abbreviated fields)
- `data` → `activeUsers` (semantically meaningful context)

This significantly improves agent performance by giving the LLM clear, unambiguous data to work with instead of cryptic codes.

**Runtime behavior:** The code works correctly at runtime - transforms are applied during validation as expected. This is purely a TypeScript typing issue that results in false errors on valid code.

**Current workaround:** Developers must use `as any` to suppress the incorrect type error, which bypasses type safety and makes the code harder to maintain.

### Steps To Reproduce

1. Create a tool that fetches data from an API and transforms cryptic codes into LLM-friendly values:

```typescript
import { z } from 'zod';
import { createTool } from '@mastra/core';

// API returns cryptic codes like "A" for active, abbreviated field names
const apiResponseSchema = z
  .object({
    data: z.array(z.object({
      user_id: z.string(),
      f_name: z.string(),      // Abbreviated
      l_name: z.string(),      // Abbreviated  
      acc_status: z.string(),  // Cryptic: "A", "I", "S"
    })),
  })
  .transform(response => ({
    // Transform to clear, LLM-friendly format
    activeUsers: response.data
      .filter(u => u.acc_status === 'A')
      .map(user => ({
        userId: user.user_id,
        fullName: `${user.f_name} ${user.l_name}`,
        accountStatus: 'Active',  // "A" → "Active"
      })),
  }));

const tool = createTool({
  id: 'list-users',
  description: 'Get active users',
  outputSchema: apiResponseSchema,
  execute: async () => {
    // Must return raw API response with cryptic codes
    return {
      data: [
        { user_id: '123', f_name: 'John', l_name: 'Doe', acc_status: 'A' },
      ],
    };
  },
});
```
2. Observe TypeScript errors:
    - On outputSchema: Type mismatch between schema and inferred output type
    - On execute: Type '{ data: ... }' is not assignable to type '{ activeUsers: ... }'

3. Run the code - it works correctly at runtime, LLM receives clear output with "Active" instead of "A"

https://github.com/tareksanger/mastra/blob/demo-create-tool-issue/packages/core/src/tools/__test__/transform-typing-issue.test.ts

### Expected Behavior

TypeScript should understand that:

1. The `execute()` function returns the **PRE-transform type** (raw API response with cryptic codes)
2. The validation flow is: `execute returns INPUT → schema.parse(INPUT) → OUTPUT to LLM`
3. The transform converts cryptic API data into semantically clear LLM-friendly data

**Expected type flow:**
```typescript
// Raw API with cryptic codes → Transform → Clear output for LLM
execute: () => Promise<{ data: Array<{user_id, f_name, l_name, acc_status: "A"|"I"|"S"}> }>  // ✅ Should accept this

// After transform, LLM receives clear data:
result: { activeUsers: Array<{userId, fullName, accountStatus: "Active"}> }
```

**Real-world benefit:**
Without transforms, the LLM sees confusing data:

```json
{ "data": [{"user_id": "123", "f_name": "John", "l_name": "Doe", "acc_status": "A"}] }
```
What does "A" mean? What's "f_name"? The LLM has to guess.

With transforms, the LLM gets clear, actionable data:

```json
{ "activeUsers": [{"userId": "123", "fullName": "John Doe", "accountStatus": "Active"}] }
```

The LLM immediately understands: "Active" users, not cryptic "A" codes. "fullName" is clear, not abbreviated "f_name"/"l_name".

No workarounds (like as any) should be necessary for this pattern.

### Environment Information

```block
- LLM Provider: OpenAI

Environment:
  OS: macOS 26.2 (ARM64 - Apple M3 Max)
  Node: 22.21.1
  pnpm: 10.25.0

npmPackages:
  mastra: 1.0.1
  @mastra/core: 1.0.4
  @mastra/ai-sdk: 1.0.2
  @mastra/rag: 2.0.0
  @mastra/evals: 1.0.0
  @mastra/mcp: 1.0.0
  @mastra/langsmith: 1.0.0
  @mastra/memory: 1.0.0
  @mastra/observability: 1.0.0
  ai: 6.0.50
  openai: 6.16.0
  langsmith: 0.3.87
  typescript: 5.9.3
  vitest: 3.2.4
  zod: 4.1.13
```

### Verification

rodboev:
Fixes #12426

## Summary

`createTool` executes before output validation and transformation, but its callback was typed to return the post-transform output. With an output schema that transforms `{ name: string }` into `{ name: string; upper: string }`, this required callers to construct a value that the runtime creates later.

The callback now accepts the schema input shape while the returned `Tool` continues to expose the validated post-transform output. The implementation follows the current schema inference and tool execution contracts in `main`.

## Root cause

`InferPublicSchema` extracts the output type from several schema representations, including AI SDK schemas and JSON Schema. The original input helper inferred `Input` from the `PublicSchema` union with `PublicSchema<any, infer Input>`. Output-only schema members could satisfy that union with `unknown`, so `execute` lost type checking for typed AI SDK and JSON Schema output schemas.

The input helper now uses the schema representations' explicit input metadata, then falls back to `InferPublicSchema<T>` for schemas that expose output information without a separate input type. `createTool` uses this input type only for the callback return. The `Tool` result, output hooks, `toModelOutput`, and `transform` continue to use the post-transform output type.

## Changes

- `packages/schema-compat/src/schema.types.ts`: add explicit input inference for Zod v4, Standard Schema, and Zod v3, with output inference as the fallback. Preserve the existing AI SDK and JSON Schema output inference.
- `packages/core/src/tools/tool.ts`: keep `ToolExecuteFunction`, including its request-context contract, and use the pre-transform type only for the `execute` output argument. Remove the constructor `as any` cast so `new Tool(opts)` remains checked.
- `packages/core/src/tools/createtool-output-types.test-d.ts`: cover transformed callbacks, an invalid callback assignment, typed AI SDK JSON Schema output, no-output tools, and the returned post-transform `Tool` output type.
- `packages/schema-compat/src/schema.ts`, `packages/schema-compat/src/index.ts`, and `packages/core/src/schema/index.ts`: preserve the public `InferPublicSchemaInput` re-exports.
- `.changeset/kind-chairs-unite.md`: document the corrected callback contract and the compile-time impact for code relying on the old annotation.

## What is NOT changed

- Runtime execution remains `execute` return, output validation, schema transformation, then the post-transform result.
- `ToolAction` and the public tool interface remain unchanged.
- Current tool request-context, background, resume, and validation behavior remains unchanged.
- Input schema inference remains unchanged.
- Provider-specific JSON Schema coercion and default handling remain outside this change.

## Prior work

PR #12570 explored the same issue. This fix keeps the change at the `createTool` boundary rather than adding a type parameter to the shared `ToolAction` interface.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## Checklist

- [x] Linked related issue
- [x] Added or updated tests
- [ ] Addressed all Coderabbit comments

## Test plan

- `pnpm exec vitest --config vitest.config.ts --project typecheck:packages/core packages/core/src/tools/createtool-output-types.test-d.ts`: 4/4 passing. Covers pre-transform callback acceptance, invalid callback typing, typed JSON Schema output, no-output behavior, and post-transform `Tool` output.
- `pnpm exec vitest --config vitest.config.ts --project typecheck:packages/core packages/core/src/tools/createtool-input-types.test-d.ts packages/core/src/tools/request-context-schema-types.test-d.ts packages/core/src/tools/suspend-execute-return-types.test-d.ts`: 15/15 passing. Covers input, request-context, and suspend-return regressions.
- `pnpm --filter ./packages/core test src/tools/createtool-types.test.ts`: 10/10 passing. Covers runtime createTool type behavior.
- `pnpm --filter ./packages/core check`: passed.
- `pnpm --filter ./packages/core lint`: passed.
- `pnpm --filter ./packages/schema-compat lint`: passed.
- `pnpm --filter ./packages/schema-compat test`: 829 passed, 9 skipped.
- `pnpm --filter ./packages/core test`: 772 passed, 22 failed, 1 skipped. A clean current-main run reported 772 passed, 21 failed, 1 skipped. The failures are in unrelated Windows socket, path, LLM-recorder, and fixture surfaces; the additional worktree failure was `MastraCompositeStore editor shorthand`, outside the changed files.

## Upstream

Reported by @tareksanger.

## ELI5

`createTool.execute` returns data before the schema cleans it up. This PR fixes TypeScript so `execute` accepts that raw data while the returned tool still exposes the cleaned-up data.

## Overview

This PR fixes `createTool.execute` typing for transformed `outputSchema` values. It adds input-schema inference, updates public exports, and adds type-level regression tests. Runtime behavior remains unchanged.

## What changed

- Added `InferPublicSchemaInput<T>` to infer a schema’s pre-transform input type.
- Re-exported `InferPublicSchemaInput` from `@mastra/schema-compat` and `@mastra/core/schema`.
- Updated `createTool` so `execute` returns the pre-transform input type.
- Preserved post-transform output types for `Tool`, `toModelOutput`, and `transform`.
- Preserved `ToolAction`, request-context typing, and input-schema inference.
- Added coverage for:
  - Transformed outputs.
  - Negative transformed-output cases.
  - Plain schemas.
  - Missing output schemas.
  - Typed AI SDK `jsonSchema`.
  - Returned-tool output types.
- Added a patch changeset for `@mastra/schema-compat` and `@mastra/core`.

## Impact

- Runtime validation and transformation do not change.
- `execute` can return raw API response shapes without unsafe casts.
- Type inference supports Zod v3, Zod v4, Standard Schema, typed `jsonSchema`, and fallback inference.
- Consumers that relied on the previous post-transform `execute` return annotation may need a compile-time adjustment.

### DIFF
--- a/.changeset/kind-chairs-unite.md
+++ b/.changeset/kind-chairs-unite.md
@@ -1,0 +1,6 @@
+---
+'@mastra/schema-compat': patch
+'@mastra/core': patch
+---
+
+Fixed `createTool.execute` typing to accept the schema input shape before output validation and transformation. Code relying on the previous post-transform callback annotation may need a type adjustment after upgrading. Runtime validation and transformation remain unchanged.
--- a/packages/core/src/schema/index.ts
+++ b/packages/core/src/schema/index.ts
@@ -5,6 +5,7 @@ import type { ZodType as ZodTypeV4 } from 'zod/v4';
 export type {
   PublicSchema,
   InferPublicSchema,
+  InferPublicSchemaInput,
   StandardSchemaWithJSON,
   InferStandardSchemaOutput,
   StandardSchemaIssue,
--- a/packages/core/src/tools/createtool-output-types.test-d.ts
+++ b/packages/core/src/tools/createtool-output-types.test-d.ts
@@ -1,0 +1,106 @@
+import type { Schema as AISdkSchema } from '@internal/ai-sdk-v4';
+import { jsonSchema } from '@mastra/schema-compat';
+import type { InferPublicSchemaInput } from '@mastra/schema-compat';
+import { describe, expectTypeOf, it } from 'vitest';
+import { z } from 'zod/v4';
+
+import { createTool } from './tool';
+import type { InferToolOutput } from './ui-types';
+
+/**
+ * Regression tests for issue #12426: `createTool`'s `execute` callback return type
+ * was typed as the post-transform output schema type (`z.output<T>`) instead of the
+ * pre-transform type (`z.input<T>`). This is incorrect because transforms are applied
+ * during validation after execute returns. These are type-level assertions only.
+ */
+describe('createTool execute return type inference (issue #12426)', () => {
+  it('execute return type matches pre-transform shape when outputSchema uses .transform()', () => {
+    const outputSchema = z.object({ name: z.string() }).transform(d => ({ ...d, upper: d.name.toUpperCase() }));
+
+    const tool = createTool({
+      id: 'transform-output',
+      description: 'Test',
+      outputSchema,
+      execute: async () => ({ name: 'test' }),
+    });
+
+    expectTypeOf<InferToolOutput<typeof tool>>().toEqualTypeOf<{
+      name: string;
+      upper: string;
+    }>();
+
+    // The callback returns the input shape, so post-transform fields are rejected.
+    createTool({
+      id: 'transform-output-invalid',
+      description: 'Test',
+      outputSchema,
+      execute: async () => {
+        // @ts-expect-error - `upper` is a post-transform field.
+        return { name: 'test', upper: 'TEST' } satisfies InferPublicSchemaInput<typeof outputSchema>;
+      },
+    });
+
+    createTool({
+      id: 'transform-output-explicit',
+      description: 'Test',
+      outputSchema,
+      execute: async () => ({ name: 'test' }),
+    });
+  });
+
+  it('execute return type matches the plain shape when outputSchema has no transform', () => {
+    const outputSchema = z.object({
+      id: z.string(),
+      status: z.enum(['active', 'inactive']),
+    });
+
+    createTool({
+      id: 'plain-output',
+      description: 'Test',
+      outputSchema,
+      execute: async () => ({ id: '123', status: 'active' as const }),
+    });
+  });
+
+  it('keeps typed AI SDK schema output input checking', () => {
+    const outputSchema: AISdkSchema<{ a: string }> = jsonSchema<{ a: string }>({
+      type: 'object',
+      properties: { a: { type: 'string' } },
+      required: ['a'],
+    });
+
+    createTool({
+      id: 'typed-json-schema-output',
+      description: 'Test',
+      outputSchema,
+      execute: async () => ({ a: 'value' }),
+    });
+
+    createTool({
+      id: 'typed-json-schema-output-invalid',
+      description: 'Test',
+      outputSchema,
+      // @ts-expect-error - the callback must return the typed schema shape.
+      execute: async () => ({ wrong: 1 }),
+    });
+  });
+
+  it('execute allows void/undefined returns when outputSchema is not provided', () => {
+    createTool({
+      id: 'no-output',
+      description: 'Test',
+      execute: async () => {
+        // Should allow void/undefined
+        return undefined;
+      },
+    });
+
+    createTool({
+      id: 'no-output-implicit',
+      description: 'Test',
+      execute: async () => {
+        // Should allow returning nothing
+      },
+    });
+  });
+});
--- a/packages/core/src/tools/createtool-output-types.test-d.ts
+++ b/packages/core/src/tools/createtool-output-types.test-d.ts
@@ -1,6 +1,5 @@
 import type { Schema as AISdkSchema } from '@internal/ai-sdk-v4';
 import { jsonSchema } from '@mastra/schema-compat';
-import type { InferPublicSchemaInput } from '@mastra/schema-compat';
 import { describe, expectTypeOf, it } from 'vitest';
 import { z } from 'zod/v4';
 
@@ -29,15 +28,13 @@ describe('createTool execute return type inference (issue #12426)', () => {
       upper: string;
     }>();
 
-    // The callback returns the input shape, so post-transform fields are rejected.
+    // The callback is checked against the pre-transform shape rather than unknown.
     createTool({
       id: 'transform-output-invalid',
       description: 'Test',
       outputSchema,
-      execute: async () => {
-        // @ts-expect-error - `upper` is a post-transform field.
-        return { name: 'test', upper: 'TEST' } satisfies InferPublicSchemaInput<typeof outputSchema>;
-      },
+      // @ts-expect-error - `name` remains a string in the pre-transform shape.
+      execute: async () => ({ name: 42 }),
     });
 
     createTool({
--- a/packages/core/src/tools/tool.ts
+++ b/packages/core/src/tools/tool.ts
@@ -2,7 +2,7 @@ import type { ToolBackgroundConfig } from '../background-tasks';
 import type { Mastra } from '../mastra';
 import { RequestContext } from '../request-context';
 import { toStandardSchema } from '../schema';
-import type { PublicSchema, StandardSchemaWithJSON, InferPublicSchema } from '../schema';
+import type { PublicSchema, StandardSchemaWithJSON, InferPublicSchema, InferPublicSchemaInput } from '../schema';
 import type { SuspendOptions } from '../workflows';
 import type {
   McpMetadata,
@@ -554,6 +554,7 @@ export class Tool<
  */
 type SchemaLike = PublicSchema<any> | undefined;
 type InferSchema<T extends SchemaLike> = T extends PublicSchema<any> ? InferPublicSchema<T> : unknown;
+type InferSchemaInput<T extends SchemaLike> = T extends PublicSchema<any> ? InferPublicSchemaInput<T> : unknown;
 
 type CreateToolOpts<
   TId extends string,
@@ -579,7 +580,7 @@ type CreateToolOpts<
   outputSchema?: TOutputSchema;
   suspendSchema?: TSuspendSchema;
   resumeSchema?: TResumeSchema;
-  execute?: ToolExecuteFunction<InferSchema<TInputSchema>, InferSchema<TOutputSchema>, TContext, TRequestContext>;
+  execute?: ToolExecuteFunction<InferSchema<TInputSchema>, InferSchemaInput<TOutputSchema>, TContext, TRequestContext>;
 };
 export function createTool<
   TId extends string = string,
--- a/packages/schema-compat/src/index.ts
+++ b/packages/schema-compat/src/index.ts
@@ -49,7 +49,7 @@ export { type JSONSchema7, type Schema, jsonSchema } from './json-schema';
 
 // Re-export standard schema types and functions from the schema.ts subpath
 // These are also available directly from the main entry for convenience
-export type { PublicSchema, InferPublicSchema } from './schema.types';
+export type { PublicSchema, InferPublicSchema, InferPublicSchemaInput } from './schema.types';
 
 export type {
   StandardSchemaWithJSON,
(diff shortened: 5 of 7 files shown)


abhiaiyer91 (changes requested):
## Review: fix(core): use pre-transform type for createTool execute return

The diagnosis is right and the goal is worth shipping: `execute` runs *before* validation/transform, so typing its return as `z.output<T>` is wrong. I verified this end-to-end locally, including porting the change onto current `main`. Two blocking issues, both surfaced by rebasing.

### Verification performed

- `pnpm install` + `pnpm turbo build --filter '@mastra/core^...'` on both the PR head and `main`.
- PR head: `vitest --project typecheck:packages/core createtool-output-types` → **3/3 pass**; reverting only the source changes (keeping the new test file) makes the transform test fail with `Property 'upper' is missing in type '{ name: string; }'`. The test genuinely gates the fix.
- Ported the change onto current `main` and ran the full type-test project (**154 type tests**), `src/tools` unit suite (**311 passed / 4 skipped**), and `tsc --noEmit` on `@mastra/core` (clean).
- Probed the resolved types directly with scratch `.test-d.ts` files (removed afterwards).

---

### 1. Blocking — `InferPublicSchemaInput` silently degrades non-Zod output schemas to `unknown`

The new type mirrors the *old* shape of `InferPublicSchema`:

```ts
export type InferPublicSchemaInput<T extends PublicSchema> =
  T extends PublicSchema<any, infer Input> ? Input : never;
```

But `main` replaced that naive form in #18608 precisely because it does not infer reliably across the `PublicSchema` union. `SchemaV4/V5/V6`, `AISdkSchemaLike` and `JSONSchema7` are parameterized by `Output` only — there is no `Input` inference site — so the conditional resolves to `unknown`.

Measured on current `main` with the PR's approach applied:

| output schema | `InferPublicSchema` | `InferPublicSchemaInput` |
| --- | --- | --- |
| `z.object({...}).transform(...)` | `{ name: string; upper: string }` | `{ name: string }` ✅ |
| `z.object({ id: z.string() })` | `{ id: string }` | `{ id: string }` ✅ |
| `jsonSchema<{ a: string }>(...)` (ai-sdk `Schema`) | `{ a: string }` | `unknown` ❌ |
| `JSONSchema7` | — | `unknown` ❌ |

Because `execute`'s return type is now driven by this type, tools using an ai-sdk `Schema` or raw JSON Schema as `outputSchema` lose return-type checking entirely — this regresses the inference added in #17435. I confirmed it with a negative test: with the PR's type, `execute: async () => ({ wrong: 1 })` against `jsonSchema<{ a: string }>(...)` compiles clean, and `@ts-expect-error` reports "Unused '@ts-expect-error' directive".

Fix that I verified works (note the fallback must be `InferPublicSchema<T>`, and the Zod inference sites must be explicit — using `T extends PublicSchema<any, infer Input>` with a fallback still yields `unknown`, because the union match succeeds with `Input = unknown`):

```ts
export type InferPublicSchemaInput<T extends PublicSchema> =
  T extends { _input: infer Input }
    ? Input
    : T extends { '~standard': { types: { input: infer I } } }
      ? I
      : T extends z4.ZodType<any, infer Input>
        ? Input
        : T extends z3.Schema<any, z3.ZodTypeDef, infer Input>
          ? Input
          : InferPublicSchema<T>;
```

With this version: the 3 new tests pass, ai-sdk `Schema` output schemas keep full `execute` return checking, and the whole 154-test type project plus `tsc --noEmit` stay clean.

### 2. Blocking — merge conflicts are semantic, not mechanical

Both conflicting files are the two files this PR changes:

- **`packages/core/src/tools/tool.ts`** — `main` now uses `execute?: ToolExecuteFunction<InferSchema<TInputSchema>, InferSchema<TOutputSchema>, TContext, TRequestContext>` (#19680, which made `requestContext` non-optional inside the callback). Taking this PR's side drops that fix; taking `main`'s side drops this one. The correct resolution keeps `ToolExecuteFunction` and only swaps the output type parameter:

  ```ts
  execute?: ToolExecuteFunction<InferSchema<TInputSchema>, InferSchemaInput<TOutputSchema>, TContext, TRequestContext>;
  ```

- **`packages/schema-compat/src/schema.types.ts`** — conflicts with the rewritten `InferPublicSchema` from #18608 (see item 1).

Good news from the rebase: **`new Tool(opts as any)` is no longer needed.** With the `ToolExecuteFunction` resolution above, `return new Tool(opts)` (as on `main`) type-checks and `tsc --noEmit` on `@mastra/core` passes with no cast. Please drop the `as any` rather than carrying it forward with a comment — it currently disables checking of every field on `opts` at the constructor call site.

---

### Non-blocking

**Test coverage.** The suite only asserts the positive path, so it can't distinguish "correct pre-transform type" from "widened to `unknown`/`any`" — which is exactly the failure mode in item 1. Worth adding:

- a negative case, e.g. `// @ts-expect-error` on `execute: async () => ({ name: 'test', upper: 'TEST' })`;
- a case with an ai-sdk `Schema` / JSON-Schema `outputSchema` asserting `execute`'s return is still checked;
- an assertion that the returned `Tool`'s output type stays *post*-transform (verified locally as `Tool<unknown, { name: string }, ...>` on the PR head — the `Tool` type parameter is unchanged, which is correct, but nothing pins it).

Also, `expectTypeOf(result).toEqualTypeOf<{ name: string }>()` asserts on a locally annotated `const`, so it's vacuous — the real assertion is the `return` statement itself. I'd drop those lines or replace them with assertions on the inferred signature.

**Lint.** `prettier --check` fails locally on `packages/core/src/tools/createtool-output-types.test-d.ts` and `packages/schema-compat/src/schema.types.ts`; `pnpm prettier --write` on those two files clears it. This matches the failing Lint job.

**Changeset.** Reads well after c72cff2. Since the old types were wrong, some users will see *new* errors after upgrading (their `as any` workarounds become unnecessary, and any code relying on the post-transform return type will need adjusting). One extra sentence naming that would help the release notes.

### Summary

Correct fix, real bug, good instinct on the test file. Blocking on: (1) `InferPublicSchemaInput` must fall back to output inference so ai-sdk/JSON-Schema output types don't silently become `unknown`, and (2) a rebase that preserves #19680's `ToolExecuteFunction` signature — which also lets the `as any` go away.

rodboev:
The `createTool.execute` contract now reflects the value returned before output validation and transformation. The two blocking issues are addressed as follows:

1. `InferPublicSchemaInput` now follows current main's `InferPublicSchema` structure instead of inferring from `PublicSchema<any, infer Input>`. It explicitly handles Zod v4 `_input`, Standard Schema input, Zod v4 and Zod v3 inputs, then falls back to `InferPublicSchema<T>`. Typed AI SDK schemas and JSON Schema output checking therefore remain precise instead of widening to `unknown`.

2. `packages/core/src/tools/tool.ts` retains current main's `ToolExecuteFunction` signature, including the `requestContext` typing and current tool fields. Only the callback output argument changes to `InferSchemaInput<TOutputSchema>`. The returned `Tool` continues to use the post-transform output type, and `createTool` constructs it with `new Tool(opts)` so the constructor checks remain active.

3. The type-level coverage now includes a negative transformed-output case, a typed AI SDK `jsonSchema` case, and an assertion that the returned tool still exposes the post-transform output. The vacuous assertions on locally annotated constants are removed while the existing plain-schema and no-output cases remain. The reported formatting and lint issues in the touched type files are corrected.

The public re-exports stay in place. Consumers relying on the old post-transform callback annotation may need a compile-time adjustment after upgrading, while runtime validation and transformation remain unchanged.

PR #17816, opened 2026-06-11
fix(workspace): normalize file URI keys in LSP diagnostics map to fix Windows mismatch

### LINKED ISSUE #17813: [BUG] Workspace lsp:true returns no diagnostics on Windows - file URI mismatch (stored file:///c%3A/... vs looked-up file:///C:/...)
### Summary
On Windows, `Workspace` LSP diagnostics (`lsp: true`) always come back empty after the `waitForDiagnostics` timeout, even when the language server publishes non-empty diagnostics. Root cause: a Windows `file:` URI canonicalization mismatch between how diagnostics are **stored** vs **looked up**.

### Environment
- `@mastra/core@1.41.0` (also present in 1.36.0) · Windows 11 · Node 22
- Server: `lua-language-server`, registered via custom LSP config (#14828 / #14969)

### Where — `packages/core/src/workspace/lsp/client.ts` (added in #13441)
```ts
// store — keyed by the SERVER's published URI
this.connection.onNotification("textDocument/publishDiagnostics",
  (params) => { this.diagnostics.set(params.uri, params.diagnostics); });

// helper — no canonicalization
function toFileUri(fsPath) { return pathToFileURL(fsPath).toString(); }

// lookup — keyed by pathToFileURL → always misses on Windows
async waitForDiagnostics(filePath) { const uri = toFileUri(filePath); /* this.diagnostics.get(uri) */ }
```

### Steps to reproduce
1. On Windows, create a `Workspace` with `lsp: true` + a custom registration for `lua-language-server`.
2. Trigger diagnostics for a Lua file containing an error.
3. The server publishes non-empty diagnostics, but the call returns `[]` after the full `diagnosticTimeout` (default 5s).

### Expected vs actual
- **Expected:** published diagnostics are returned.
- **Actual:** `[]` after a 5s timeout, every time.

### Root cause
Store key = server's published URI; lookup key = `pathToFileURL(path).toString()`. On Windows these differ:
```js
require('node:url').pathToFileURL('C:\\Users\\me\\res\\client.lua').toString();
// => 'file:///C:/Users/me/res/client.lua'      (uppercase drive, literal colon)

require('vscode-uri').URI.file('C:\\Users\\me\\res\\client.lua').toString();
// => 'file:///c%3A/Users/me/res/client.lua'     (lowercased drive, %3A colon) — what lua-language-server emits
```
Different strings → `Map.get` miss → loop runs to `timeoutMs` → returns `[]`.

### Impact
Affects any LSP server emitting VS Code-style canonical URIs on Windows (lua-language-server confirmed; likely others). `lsp: true` diagnostics are unusable on Windows for those servers. Same class as #17670 (Windows path handling in `workspace/`).

### Suggested fix
Canonicalize before the Map store/lookup — simplest is to key the diagnostics Map by the OS path (`fileURLToPath(uri)`) on **both** the `publishDiagnostics` store and the `waitForDiagnostics` lookup; comparing normalized fs paths sidesteps drive-case and percent-encoding. Happy to open a PR.

Aayush-engineer:
## Description

Fixes a bug where `Workspace` LSP diagnostics (`lsp: true`) always returned `[]` on Windows after the full timeout, even when the language server published non-empty diagnostics.

### Root Cause

The `diagnostics` Map was keyed by raw URI strings, but Windows LSP servers emit different URI formats for the same path:

- Node's `pathToFileURL` produces: `file:///C:/Users/me/res/client.lua` (uppercase drive, literal colon)
- VS Code-style LSP servers (e.g. lua-language-server) emit: `file:///c%3A/Users/me/res/client.lua` (lowercase drive, percent-encoded colon)

Since these are different strings, `Map.get` always missed — the loop ran to `timeoutMs` and returned `[]` every time.

### Fix

Introduced `diagnosticsKey(uriOrPath)` — a normalization helper that:
1. Converts any `file:` URI to an OS path via `fileURLToPath`
2. Strips any leading slash before the drive letter (handles `/C:/...` edge case)
3. Lowercases the drive letter on Windows paths

Applied consistently to **all** `diagnostics` Map operations:
- `publishDiagnostics` store (line 203) — server URI normalized before storing
- `notifyOpen` delete (line 308) — normalized before clearing
- `waitForDiagnostics` lookup (line 342) — normalized before querying
- `notifyClose` delete (line 378) — normalized before clearing

This ensures store and lookup keys always match regardless of which URI form the server emits.

### Tests Added

4 new unit tests in `client.test.ts` covering:
- Windows URIs differing in drive-letter case and colon encoding produce the same key
- Drive letter is always lowercased
- POSIX paths pass through unchanged
- Non-file-URI strings pass through unchanged

All 157 LSP tests pass.

## Related issue(s)

Fixes #17813

## Type of change

- [x] Test update

## Checklist

## ELI5
Windows can represent the same file path in different URI forms (e.g. file:///C:/... vs file:///c%3A/...), which caused diagnostics to be stored under one form but looked up under another. This PR normalizes file URIs/paths before storing and looking up diagnostics so they always match.

## Overview
Fixes a bug where Workspace LSP diagnostics returned empty arrays on Windows after waitForDiagnostics timed out because diagnostic Map keys differed between server-published URIs and lookup URIs.

## Changes
- Added exported helper `diagnosticsKey(uriOrPath: string): string` in packages/core/src/workspace/lsp/client.ts that:
  - Converts file: URIs to filesystem paths via fileURLToPath (falls back to input on error),
  - Strips any leading slash before a Windows drive letter,
  - Lowercases Windows drive letters (so C: and c: normalize the same).
- Applied diagnosticsKey to all diagnostics Map operations in LSPClient:
  - publishDiagnostics store: diagnostics.set(diagnosticsKey(params.uri), ...)
  - notifyOpen delete
  - waitForDiagnostics lookup
  - notifyClose delete
- Added four unit tests in packages/core/src/workspace/lsp/client.test.ts to verify:
  - Windows URI variations (drive-letter case and colon encoding) produce the same key,
  - Drive-letter lowercasing,
  - POSIX file:// path pass-through,
  - Non-file-URI/path pass-through.
- Updated changeset documenting the user-facing fix.

## Testing
All existing LSP tests (157) pass and 4 new tests validate diagnosticsKey normalization behavior.

## Related Issues
Fixes `#17813`


Aayush-engineer:
Fixed in latest commit — rewrote the changeset to focus on user-facing outcome rather than implementation details.

Aayush-engineer:
@wardpeet — ready for review when you get a chance. CodeRabbit is clean and all 157 + 4 new LSP tests pass.

TheIsrael1 (approved):

PR #17818, opened 2026-06-11
fix(core): semantic recall not injected for signal-delivered user messages

### LINKED ISSUE #17797: Semantic recall (scope: resource) retrieves but does not inject into prompt in Studio playground (works via API)
## Description

Semantic recall with `scope: 'resource'` retrieves messages correctly (they show up in the Studio **Semantic Recall** side panel), but the retrieved messages are **not injected into the LLM prompt** when chatting through the **Studio playground**. The agent answers "I don't have that information" even though the panel clearly shows the relevant past messages were recalled.

The exact same memory configuration works **correctly via the HTTP API** (both `POST /api/agents/:id/generate` and `/stream`) with explicit `memory: { resource, thread }`. So the retrieval + injection pipeline itself is fine — the issue appears specific to how the **playground** constructs the chat request.

## Expected behavior

When semantic recall retrieves messages (cross-thread, `scope: 'resource'`), those messages should be injected into the model prompt (as a `<remembered_from_other_conversation>` system block) so the agent can use them — the same way it works via the API.

## Actual behavior

In the Studio playground:
- The **Semantic Recall** panel shows the retrieved messages (retrieval works).
- The agent's answer claims it has no memory of them.
- Inspecting the trace, the `model_generation` prompt contains only `[system, user]` — **no recalled-context block**.

Via the API the same query yields a prompt with the `<remembered_from_other_conversation>` system block, and the agent answers correctly.

## Steps to reproduce

Agent with semantic recall enabled:

```ts
const memory = new Memory({
  embedder: /* any embedder */,
  vector: new LibSQLVector({ id: 'agent-vector', url: 'file:./vector.db' }),
  options: {
    lastMessages: 20,
    semanticRecall: { topK: 3, messageRange: 2, scope: 'resource' },
  },
});
```

**In the Studio playground:**
1. New chat → "Remember: my secret code word is Orchid-7."
2. New chat (same agent) → "What was my secret code word?"
3. Observe: the Semantic Recall panel shows the recalled message, but the agent answers that it doesn't know.

**Via the API (works correctly — contrast):**
```bash
# plant
curl -X POST localhost:4111/api/agents/<id>/generate -H 'Content-Type: application/json' \
  -d '{"messages":[{"role":"user","content":"Remember: my secret code word is Orchid-7."}],"memory":{"resource":"u1","thread":"t-A"}}'
# cross-thread query — answers "Orchid-7" correctly
curl -X POST localhost:4111/api/agents/<id>/generate -H 'Content-Type: application/json' \
  -d '{"messages":[{"role":"user","content":"What was my secret code word?"}],"memory":{"resource":"u1","thread":"t-B"}}'
```

## Ruled out during debugging

- **Not a model issue** — reproduced with both a local model (Qwen via LM Studio) and `openai/gpt-5-mini`.
- **Not reasoning/thinking** — same with thinking on/off.
- **Not the streaming path** — both `/generate` and `/stream` inject correctly via API.
- **Not embedder/vector** — retrieval works (panel shows hits), and API injection works with the same stores.
- Working memory (`scope: 'resource'`) *does* persist across playground chats, suggesting the playground uses a stable `resourceId` — so cross-thread recall should be in scope.

## Environment

- `@mastra/core` 1.41.0
- `@mastra/memory` 1.20.2
- `@mastra/libsql` 1.12.1
- Storage: instance-level `MastraCompositeStore` (memory inherits instance storage; vector via `LibSQLVector`)
- Node 24, macOS

## Note

Our storage is configured at the instance level via `MastraCompositeStore` and inherited by the agent's `Memory` (no explicit `storage` on the `Memory`). Adding explicit `storage` to the `Memory` did **not** change the playground behavior. Flagging in case the playground resolves memory storage differently from the API request path.

Sandy-1711:
## Description

Semantic recall retrieves messages but never injects them into the prompt when chatting through the Studio playground, while the exact same memory config works via the HTTP API (`/generate` and `/stream`). The Studio Semantic Recall panel shows the recalled messages, but the agent answers as if it has no memory of them.

Root cause: the playground sends chat turns through the agent signal pipeline (`agent.sendMessage` → thread-stream runtime → `agent.stream(signal, ...)`). User turns delivered this way are stored as `role: 'signal'` rows and only get projected to `role: 'user'` at prompt time, after input processors have already run. `SemanticRecall.extractUserQuery()` only matched `role === 'user'`, so on this path it found no query and silently skipped the search and injection — retrieval still worked elsewhere (the side panel does its own `/api/memory/search` call with the raw text), which is why the panel showed hits while the prompt got nothing.

Fix: treat user-type signals as user messages when extracting the recall query. Non-user signals (state, notification, system-reminder) still don't trigger recall.

Tests: added a `Signal-Delivered User Messages` block to the `SemanticRecall` suite covering the plain user message path, the signal-delivered path (which fails without this fix — the embedder is never called), and a guard that non-user signals don't run recall.

## Related issue(s)

Fixes #17797

## Type of change

- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [ ] Test update

## Checklist

## ELI5 (Explain Like I'm 5)
The AI agent has a memory system that looks back at previous conversations to answer questions better. When you chat through the Studio playground interface, the memory feature correctly found relevant old conversations but forgot to actually use them when answering—kind of like looking up information in a book but then not reading it. The fix teaches the memory system to recognize messages that arrive through the playground's different communication channel so it can properly use them when generating answers.

## Summary

This PR fixes a bug where semantic recall (agent memory that retrieves relevant past messages) would successfully find matching messages in the Studio playground interface but fail to inject them into the LLM prompt for the model to use when generating responses.

**Problem:** The Studio playground delivers user input through a different internal mechanism (the agent signal pipeline) than the HTTP API routes do. When messages arrive this way, they're temporarily stored with `role: 'signal'` instead of `role: 'user'`. The semantic recall extraction logic only looked for `role: 'user'` messages, so it skipped signal-delivered user turns entirely, even though the HTTP API worked correctly.

**Solution:** Updated `SemanticRecall.extractUserQuery()` to recognize signal messages that represent user input (`user-message` signals) as valid queries for semantic recall, while continuing to exclude other signal types (system-reminder, state, notification).

**Changes:**
- Modified `packages/core/src/processors/memory/semantic-recall.ts` to add `isMastraSignalMessage` import and a helper method `isUserMessageSignal()` that detects signal messages representing user input
- Updated `extractUserQuery()` to match either traditional `role: 'user'` messages or recognized user-type signals
- Added comprehensive test suite `Signal-Delivered User Messages` in `semantic-recall.test.ts` that verifies recall works for plain user input, signal-delivered user messages, and correctly skips non-user signal types
- Added changeset entry documenting the patch fix for `@mastra/core`

**Related:** Fixes issue `#17797`


TylerBarnes (approved):

TylerBarnes:
thanks for this! good catch

PR #17819, opened 2026-06-11
feat(core): propagate typed errors from processors and workflow steps via FatalError

### LINKED ISSUE #17808: [FEATURE] `fatal` exit to terminate agent with a propagated error
### Problem Statement

There's no way to terminate the agent execution due to a non-retryable / internal errors that occur from within a processor or workflow step. `abort()` bails cleanly but swallows the error entirely, nothing reaches the caller (e.g., translate the error within `onError` and rewrite it to avoid leaking internal details to external clients). 

The only option is to throw, but Mastra catches it at every layer and reconstructs a new error from the message string, so whatever typed context you had (error codes, custom fields, cause chains) is gone by the time it surfaces.

This makes it impossible to propagate a meaningful, structured error from inside the execution pipeline to the code that invoked it. You end up either parsing Mastra's internal message format at the call site or maintaining out-of-band state to move context across the boundary — both of which are workarounds for something the framework would ideally handle natively.

### Proposed Solution

A `fatal` mechanism on the abort controller, or a `fatal` callback function, available in the processor and step context that terminates execution immediately and delivers the provided error to the caller without wrapping.

```ts
// In a processor
processInputStep({ messages, abort }) {
    if (detectTerminalCondition(messages)) {
        abort.fatal(new AppError('TENANT_DATA_UNAVAILABLE'));
    }
}

// In a workflow step
execute({ inputData, abort }) {
    abort.fatal(new AppError('QUOTA_EXCEEDED'));
}
```

The caller receives the original error as-is:

```ts
// agent.generate()
try {
    await agent.generate(messages);
} catch (err) {
    // err is AppError, not a wrapped plain object
}

// handleChatStream onError
onError: (error) => {
    // same — original error, not a wrapper
}
```

The contract: fatal is non-resumable (same semantics as `abort()`), but the error is the terminal value propagated to the caller rather than discarded.

### Component

Agents, Workflows

### Alternatives Considered

- Throw and parse the wrapper message at the call site — couples application code to Mastra's internal message formatting (`"step processor:<id>: <message>"`), which is an implementation detail that can change across releases.
- `AsyncLocalStorage` side-channel — store the error before throwing, read it back at the call site. Works but is application-level plumbing for what is fundamentally a framework concern.
- Encode context in `error.message` — since .message is the one field Mastra preserves, serialize what you need and deserialize at the call site. Functional, but not how anyone should be writing code.

### Example Use Case

As a developer building agents over workflows, when a step or processor encounters a non-retryable condition (unavailable data source, exhausted quota, auth failure), I want to propagate a structured error to the caller so I can handle it programmatically and emit a single, fully-contextualized log — without reverse-engineering Mastra's internal error wrapping or maintaining out-of-band state.

### Additional Context

_No response_

### Verification

intojhanurag:
## Description

Adds a way for processors and workflow steps to terminate execution with a typed error that propagates to the caller without being wrapped, re-classed, or serialized. Previously, custom errors thrown from processors got re-classed as `MastraError` and `abort(reason)` swallowed typed context entirely, so callers could not use `instanceof` or read custom properties.

- New `FatalError` class and `isFatalError` / `attachFatal` helpers in `packages/core/src/agent/fatal-error.ts`
- New `abort.fatal(err)` method on the processor abort callback via `ProcessorAbortFn` interface
- All 9 inline abort definitions in `processors/runner.ts`, plus `workflow.ts` and `evented/workflow.ts`, wrapped with `attachFatal()`
- Catch blocks in `agent.ts` (input processors + 4 outer agent boundaries), `runner.ts` (workflow-as-processor + output processor), `workflows/default.ts` (`executeStepWithRetry` + `fmtReturnValue`), and `workflows/handlers/control-flow.ts` (parallel/race) detect `FatalError` and re-throw the original cause unwrapped
- `fatal?: Error` threaded through `StepFailure` and `FormattedWorkflowResult` so nested workflow failures preserve the fatal marker
- 4 dedicated tests in `fatal-error.test.ts` covering input processor, output processor, workflow step, and `isFatalError`
- Fully opt-in and backward compatible: existing `abort(reason)` and `throw new Error(...)` behavior is unchanged

## Related Issue(s)

Fixes #17808

## Type of Change

- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [ ] Test update

## Checklist

## ELI5

Normally, when an error happens inside a processor or workflow step, the framework wraps it in its own error type, which makes it hard for the caller to know what really went wrong or handle it properly. This PR adds a new way to throw errors (`abort.fatal()` or `throw new FatalError(...)`) that skips the wrapping and delivers the original error directly to the caller, so it keeps its type, custom properties, and cause chain intact.

## Overview

This PR introduces a mechanism for processors and workflow steps to propagate structured, typed errors to callers without framework wrapping. It adds:
- A new `FatalError` sentinel class for marking errors that should bypass wrapping
- An `abort.fatal(err)` method on processor abort callbacks
- Support for `throw new FatalError(err)` in workflow steps
- Preservation of original error class identity, custom properties, and cause chains

The feature is entirely opt-in; existing `abort(reason)` and `throw new Error(...)` behavior remains unchanged.

## Key Changes

**New Types & Utilities:**
- `FatalError` class that wraps a user error while preserving its original form for unwrapping
- `isFatalError(err)` type guard for detecting fatal errors across realm boundaries
- `ProcessorAbortFn<TTripwireMetadata>` interface supporting both `abort(reason)` and `abort.fatal(error)` methods
- `StepFatalInfo` interface in workflow types to track fatal error metadata

**Agent Layer:**
- `agent.generate()`, `stream()`, `resumeStream()`, and `resumeGenerate()` now detect and rethrow fatal errors from processor failures and workflow step failures without additional wrapping
- Input processor error handling recognizes `FatalError` and rethrows it directly instead of wrapping in `MastraError`

**Processor Layer:**
- Abort callbacks are now enhanced via `attachFatal()` to provide both standard `abort(reason)` and `abort.fatal(error)` paths
- Processor runners detect fatal errors and preserve them through the execution pipeline

**Workflow Layer:**
- Workflow execution engine captures fatal errors from step execution and includes them in step failure metadata
- Control flow handlers (`executeParallel`, `executeConditional`) preserve fatal metadata from failed branches
- Fatal error formatting integrates with the result serialization to keep fatal errors separate from standard tripwire/error handling

## Testing

New comprehensive test suite (`fatal-error.test.ts`) covers:
- Input processor calling `abort.fatal(...)` causing `agent.generate()` to reject with the original error
- Workflow step throwing `FatalError` resulting in failed run with `result.fatal` capturing the original error and its properties
- `isFatalError` behavior and cross-realm detection
- Preservation of custom error properties (code, retryAfterSeconds, etc.)

All 267 existing tests pass; typecheck is clean.


intojhanurag:
Not good , closing this .

PR #17820, opened 2026-06-11
feat(core): make task tools agent-agnostic with durable thread-state storage + state-signal projection

abhiaiyer91:
## What

Makes the built-in task tools (`task_write` / `task_update` / `task_complete` / `task_check`) work on any Mastra `Agent`, not just inside the Harness/TUI, and makes the task list **durable** so it survives a process restart. This is the third PR in the stack making harness-coupled built-in tools agent-agnostic (after `ask_user` and `submit_plan`).

## Why

The task tools previously stored the task list in **Harness session state** and injected the live list into the **cached system prompt** on every mutation, which busted the prompt-cache prefix on each task update. Neither mechanism exists for a plain `Agent`, so the tools only worked inside the Harness. The session-state list was also in-memory only, so on a thread reopen after restart the pinned task UI (rebuilt from message replay) and the agent's working list diverged — the agent would report "the list was cleared" for tasks still shown on screen.

## How

- **`threadState` storage domain** — a new generic, thread-scoped domain keyed by `(threadId, type)`. The task list lives under `type: "task"`; the domain is intentionally generic so other agent-scoped state (e.g. `"goal"`) can reuse it. Tools reach it via `context.mastra.getStorage().getStore('threadState')`, scoped by the run's `threadId`. This is the single source of truth.
- **Durable backend** — the composite store wires `InMemoryThreadStateStorage` by default (works out of the box, no config), and `@mastra/libsql` ships `ThreadStateLibSQL` for durability. Because mastracode's storage is LibSQL-backed, the task list now persists across restarts — fixing the reopen/divergence bug.
- **State-signal projection (delta-first)** — `TaskStateProcessor` projects the list into the model context as a `tasks` state-signal lane. The first projection (and each compaction) is a full `<current-task-list>` **snapshot**; subsequent mutations emit small `<task-list-update>` **deltas**. It re-snapshots from the `threadState` store when observational-memory truncation drops the base snapshot, so the list survives context-window truncation without re-sending the whole list every turn.
- **Memory required** — with no thread/resource the tools no-op cleanly (single mechanism, no fallback path).
- **Prompt-cache fix** — the task list is removed from harness session state and from the cached system prompt, so task updates no longer invalidate the prompt-cache prefix. The display-only `task_updated` event and thread-replay reconstruction are preserved for the TUI.
- **One-liner setup** — `TaskSignalProvider` bundles the four tools + `TaskStateProcessor` behind a single `signals: [new TaskSignalProvider()]` registration so they can't get out of sync.
- **Completed-task receipt removed** — once a list is fully completed the pinned list hides itself and the agent narrates completion, so the inline receipt only read as a confusing second live list. The explicit "Tasks cleared" receipt is kept.

## Test plan

- `pnpm --filter @mastra/core check` — clean
- `pnpm --filter @mastra/core test` (task-tools, task-state-processor incl. delta/compaction/OM re-emit, signal-provider, thread-state in-memory, e2e, harness display-bridge) — pass
- `pnpm --filter @mastra/libsql test` — `ThreadStateLibSQL` CRUD + durability (fresh store instance over the same DB sees persisted tasks) — pass
- `pnpm --filter ./mastracode test` — processor/provider registration, prompt-cache assertion, render-messages/event-dispatch receipt behavior — pass
- Full turbo build verified green for `@mastra/libsql`, `@mastra/clickhouse`, `@mastra/cloudflare` (new table added to exhaustive `TABLE_NAMES` maps)
- `pnpm --filter @mastra/core lint` + prettier — clean
- Manual MC smoke (scenarios verified): CRUD lifecycle, single-in-progress invariant, cross-turn continuity, completion clears pinned list with no lingering receipt, explicit clear shows cleared receipt, **thread reopen after restart now keeps the agent's working list in sync with the pinned UI**, no state-signal XML in transcript.

> Stacked on `feat/submit-plan-native-suspend`.

## ELI5

This PR lets any AI agent using Mastra handle task lists (like a to-do list), not just the special Harness/TUI version. Instead of storing tasks in the agent's session state, tasks are now saved in a persistent memory store and shown to the agent as a "state signal" that updates as tasks change—similar to how you'd show a to-do list to a human to keep them focused on what needs doing.

---

## Summary

This PR refactors task tools (`task_write`, `task_update`, `task_complete`, `task_check`) to be agent-agnostic by decoupling them from Harness-specific state management. Tasks are now stored in a generic, thread-scoped **`ThreadStateStorage`** domain keyed by `(threadId, type)` with tasks persisted under `type: "task"`. The composite store wires an `InMemoryThreadStateStorage` by default, and durable persistence is provided by `@mastra/libsql`'s `ThreadStateLibSQL` implementation.

A new `TaskStateProcessor` input processor projects the task list as a **state-signal** (using `stateId: "tasks"`) to the agent's model context via a delta-first approach: full `<current-task-list>` snapshots are emitted on initial population or after observational-memory compaction, followed by `<task-list-update>` deltas for incremental changes. This keeps the system prompt stable for prompt caching while allowing tasks to survive context-window truncation through re-snapshotting. Tools read and write task state synchronously through `context.mastra.getStorage().getStore("threadState")` scoped by the run's `threadId`, and no-op gracefully if the run lacks a memory backing (`threadId`/`resourceId`).

A new `TaskSignalProvider` signal provider bundles the four task tools and `TaskStateProcessor` for easy registration on any agent's `signals` array, ensuring synchronized behavior. The Harness implementation no longer stores the task list in session state but continues to emit the `task_updated` display event and reconstruct tasks from storage on thread replay. The completed-task inline receipt is removed—completion is now narrated by the agent and indicated by the pinned task list state change—while an explicit "Tasks cleared" receipt is preserved.

---

### Core Storage & State Management
- Added `ThreadStateStorage` abstract base class defining the CRUD contract for thread-scoped state keyed by `(threadId, type)`
- Added `InMemoryThreadStateStorage` in-memory implementation with cloned read/write semantics to prevent mutation of internal state
- Added `ThreadStateLibSQL` LibSQL-backed implementation for durable per-thread state persistence
- Integrated `threadState` domain into `MastraCompositeStore` with `InMemoryThreadStateStorage` as the default
- Extended `StorageDomains` type to include optional `threadState` domain

### Task Tools & State Projection
- Created new `task-tools.ts` module implementing agent-agnostic task CRUD: `taskWriteTool`, `taskUpdateTool`, `taskCompleteTool`, `taskCheckTool`
- Task tools enforce invariants (e.g., single `in_progress` task per list), support deterministic ID assignment via `assignTaskIds` with ID reuse when content matches, and no-op without memory backing
- Tools read/write via thread-scoped `threadState` store and emit optional Harness `task_updated` events when a bridge is present
- Added `TaskStateProcessor` input processor that publishes tasks as a state-signal using snapshot+delta projections with compaction and re-snapshotting on context-window truncation
- Removed inline completed-task receipt; completion is now narrated by the agent
- Fixed state-signal XML output: `TaskStateProcessor` now uses `tagName: "current-task-list"` and returns only inner markup lines so the framework emits `<current-task-list count="N">...</current-task-list>` with proper self-closing for empty lists

### Signal Provider & Integration
- Added `TaskSignalProvider` signal provider that bundles task tools and `TaskStateProcessor` for convenient one-shot registration
- Updated `mastracode` to register `TaskSignalProvider` in the code-agent configuration
- Added documentation for `TaskSignalProvider` with usage examples

### Documentation & Migration
- Updated Harness documentation to reflect that task state is no longer stored in Harness session state but in thread-scoped storage
- Clarified in task-management docs that task tools are now agent-agnostic, require memory backing, and project tasks via state signals to avoid prompt-cache pollution
- Added Changeset entries for `@mastra/core` and `@mastra/libsql` detailing the new thread-state domain and task tool agent-agnosticism

### TUI/Harness Updates
- Removed `renderCompletedTasksInline` callback from `EventHandlerContext` (no longer needed; task completion is narrated by agent)
- Updated `mastra-tui.ts` and event-dispatch handlers to filter out the tasks state-signal from inline rendering (handled by pinned task list UI instead)
- Updated `task_updated` event handling to render "cleared tasks" receipt only when transitioning to an empty list
- Refactored task-tools test in Harness to verify "display bridge" effect: `task_updated` events trigger `harness.getDisplayState().tasks` updates while not populating `harness.getState().tasks`

### Code Generation (mastracode)
- Removed task list from `buildFullPrompt` generated system prompt to avoid prompt-cache invalidation
- Updated prompt-build tests to assert tasks are no longer injected into cached prompts
- Removed `formatTaskPromptValue` helper

### Test Coverage
- Added comprehensive task-tools unit tests covering memory gating, ID assignment, persistence, cross-turn continuity, and within-turn propagation
- Added `TaskStateProcessor` tests validating snapshot/delta emissions, XML markup format (with no double-escape), compaction, and re-snapshotting on OM truncation
- Added end-to-end task-tools test simulating multi-turn task mutations in an agent run
- Added `TaskSignalProvider` tests verifying tool bundling and processor registration
- Updated `InMemoryThreadStateStorage` and `ThreadStateLibSQL` tests to verify core CRUD and durability semantics
- Updated existing TUI render/event-dispatch tests to reflect removed completed-task receipts and state-signal filtering

---

## Compatibility

- Tasks now require memory backing (`threadId` + `resourceId`) to function; plain agents without memory will see task tools no-op with explanatory results
- Agents must register `TaskSignalProvider` (or manually wire `TaskStateProcessor` and task tools) to enable task tracking
- Harness behavior is preserved (display events, thread replay reconstruction) but task state is sourced from `threadState` storage, not session state
- Completed-task inline receipts are removed; task completion is indicated by the agent's narration and the pinned task list UI state

---

## Testing

- `@mastra/core`: 468 tests pass (including task-tools, task-state-processor, e2e, harness display-bridge)
- `mastracode`: 1224 tests pass (processor registration, prompt-cache stability, render/event-dispatch receipts)
- Lint/Prettier: clean
- Manual smoke tests verified CRUD lifecycle, single-in-progress invariant, cross-turn continuity, completion behavior, explicit clear behavior, and thread reopen/replay


CalebBarnes:
This looks good!

One issue in `task-state-processor.ts` though.

`renderTaskList()` (lines 34–41) hand-wraps the contents in `<current-task-list>...</current-task-list>`, but the state-signal framework already wraps `contents` in the tag named by the signal's `tagName` field. Since this processor sets `tagName: 'state'`, the LLM ends up seeing a double-wrapped, XML-escaped payload.

`signalToXmlMarkup()` in `packages/core/src/agent/signals.ts:184-192` is the wrapping/escaping layer applied to every state signal:

```ts
return `<${tagName}${attributesXml}>${escapeXml(signal.contents)}</${tagName}>`;
```

Note the `escapeXml(signal.contents)` — contents are escaped, not embedded as raw XML. So a snapshot from this processor currently reaches the model as:

```xml
<state type="tasks" count="3">&lt;current-task-list&gt;
  ✓ [completed] {id: t1} First task
  ▸ [in_progress] {id: t2} Second task
&lt;/current-task-list&gt;</state>
```

…rather than what I'm guessing is the intent:

```xml
<current-task-list count="3">
  ✓ [completed] {id: t1} First task
  ▸ [in_progress] {id: t2} Second task
</current-task-list>
```

The other in-tree state-signal processors (`WorkingMemoryStateProcessor`, `BrowserContextProcessor`) both pass plain text in `contents` and let the framework wrap via `tagName` — worth a look as reference.

**Suggested fix** (two reasonable options):

1. **Match the browser pattern** — keep `tagName: 'state'`, drop the inline tag from `renderTaskList`, let `attributes.type: 'tasks'` distinguish the lane.
2. **Match the working-memory pattern** — promote `current-task-list` to be the signal's own `tagName`, drop `type: 'tasks'` from attributes, and have `renderTaskList` return only the inner lines. Probably preferable because:
   - It mirrors the prior `<current-task-list>` wrapper that used to be injected into the cached system prompt, so the model sees a familiar surface and existing prompt instructions stay aligned.
   - `signalToXmlMarkup` handles the empty case naturally as `<current-task-list count="0" />` when `contents` is an empty string.

Either way, the test should be updated to assert against the rendered signal (or at minimum assert plain-text `contents` plus `tagName`/`attributes`) so this contract is locked in.

abhiaiyer91:
Good catch — fixed in 4544c81. Went with option 2: `current-task-list` is now the signal's own `tagName` (dropped the `type: "tasks"` attribute, kept `count`), and `renderTaskList` returns only the inner lines so the framework wraps once instead of double-wrapping/escaping. The model now sees `<current-task-list count="N">…</current-task-list>` directly, and an empty list renders as `<current-task-list count="0" />`. Updated `task-state-processor.test.ts` to assert against the rendered `signalToXmlMarkup()` output (incl. a no-double-escape check and the empty self-closing case) so the contract is locked in.

CalebBarnes (approved):
LGTM 👍

A few non-blocking nits:

- Stale "tasks storage domain" comment in `task-tools.ts` after the rename to `threadState`.
- `stableTasksCacheKey` joins fields with `:` / `|` — content containing either could collide. Low risk, but a length-prefixed fingerprint or one regression test would be cheap insurance.

abhiaiyer91:
Thanks @CalebBarnes — both nits addressed in 4e615fd112:

- **`stableTasksCacheKey` collision**: each field is now length-prefixed (`<len>:<value>`) before joining, so content containing `:`/`|` can no longer shift a boundary into a different list. Added a regression test (`produces distinct cacheKeys when delimiter chars in content could collide`).
- **Stale "TaskStore" comments** in `task-tools.ts` updated to the current `threadState` store naming.

Also dropped the `@mastra/libsql` sentence from the `@mastra/core` changeset (CodeRabbit nit) — the `ThreadStateLibSQL` export is already documented in its own `thread-state-libsql.md` changeset.

PR #17829, opened 2026-06-11
fix(memory): prevent OM observer cross-agent deadlock

### LINKED ISSUE #17827: Blocking Observational Memory observation deadlocks in waitForCrossAgentThreadRun (circular wait with main agent run)
## Bug Description

When **blocking (synchronous) Observational Memory observation** is triggered during a multi-step agent run, the stream hangs indefinitely. The main agent run never reaches a terminal state; an 8-minute inactivity watchdog eventually cancels it.

This is a **pure JavaScript circular wait** inside `@mastra/core`, not a model timeout, database lock, or network stall.

## Affected Versions

- `@mastra/core@1.35.0` (confirmed)
- `@mastra/memory@1.18.2` (confirmed)
- Still present on `main` as of `@mastra/core@1.41.0` (checked `packages/core/src/agent/thread-stream-runtime.ts`)

## Root Cause

1. Main agent run reaches a step boundary where unobserved message tokens exceed the threshold → **blocking observation** starts. The main run is paused waiting for observation to complete.

2. OM's internal **Observer** agent calls `agent.stream()` and **inherits the main run's `requestContext`**, which includes `MASTRA_THREAD_ID_KEY` (and `MASTRA_RESOURCE_ID_KEY`).

3. At the start of `agent.stream()`, `AgentThreadStreamRuntime.waitForCrossAgentThreadRun()` sees that the same thread already has an active run (the main agent) owned by a **different** agent id.

4. The observer waits for that active run to finish. The active run is blocked waiting for observation. **Circular wait → permanent deadlock.**

Relevant upstream code (`thread-stream-runtime.ts`, `main` branch):

```ts
async waitForCrossAgentThreadRun(agent, options, pubsub?) {
  const { threadId, resourceId } = this.#getThreadTarget(options);
  if (!threadId) return;
  // ...
  while (true) {
    const activeRecord = state.threadRunsById.get(activeRunId);
    if (activeRecord) {
      if (activeRecord.agent.id === agent.id || !this.#isThreadBlockingRun(state, activeRecord)) {
        return;
      }
      await activeRecord.output._waitUntilFinished().catch(() => {});
      continue;
    }
    // ...
  }
}
```

OM internal agent ids (from `@mastra/memory`):

- `observational-memory-observer`
- `multi-thread-observer`
- `observational-memory-reflector`

## Field Evidence (production)

When the hang occurs:

- Last stream part: `data-om-observation-start`
- `OM_DEBUG` last line: `[OM:sync-obs] sealed newest message ... for ratio-aware observation`
- **No** subsequent OM logs (`observer.call`, model invocation, etc.)
- **No** in-flight HTTP or PostgreSQL activity (Node inspector: idle CPU, no pending requests)
- PostgreSQL `pg_stat_activity` shows no blocked queries
- Hang lasts until stream inactivity timeout (we observed 8 minutes)

With diagnostic probes, the last log before hang is consistently:

```
[om-probe] observer.call agent.stream invoking
```

The model (`doStream`) is never reached.

## When It Happens

- Blocking observation path (`bufferTokens: false`, or `blockAfter` safety threshold exceeded)
- Multi-step runs (`stepNumber > 0`) with enough unobserved tokens
- Caller sets `MASTRA_THREAD_ID_KEY` / `MASTRA_RESOURCE_ID_KEY` on `requestContext` (typical for session-scoped agents)

Async buffering observation often avoids the deadlock because observation runs outside the blocking step boundary — but blocking observation is the documented safety net and is hit under real workloads.

## Related but Different Fixes

These upstream fixes address **other** OM deadlock/contention issues, not this one:

- PR #13436 — PostgreSQL row lock ordering when parallel agents share `resourceId` without `threadId`
- PR #17024 — reserved runs in `subscribeToThread`

Neither changes the cross-agent wait when OM internal agents inherit the main thread's `requestContext`.

## Suggested Fix

Skip `waitForCrossAgentThreadRun` for OM-internal agents. They are designed to execute **inside** the active main run's processor pipeline, not as independent cross-thread competitors.

```ts
const OM_INTERNAL_AGENT_IDS = new Set([
  "observational-memory-observer",
  "multi-thread-observer",
  "observational-memory-reflector",
]);

async waitForCrossAgentThreadRun(agent, options, pubsub?) {
  if (OM_INTERNAL_AGENT_IDS.has(agent.id)) return;
  // ... existing logic
}
```

Alternative: do not propagate `MASTRA_THREAD_ID_KEY` into the observer/reflector `requestContext` when calling `agent.stream()` from OM.

## Minimal Reproduction

Conditions:

1. `Memory` with `observationalMemory`, `bufferTokens: false` (force blocking path), low `messageTokens` threshold
2. Multi-step agent run (tool call → second step) to reach `stepNumber > 0`
3. `requestContext` sets `MASTRA_THREAD_ID_KEY` and `MASTRA_RESOURCE_ID_KEY` (same as production session wiring)

A regression test that reproduces the deadlock without the fix (30s vitest timeout) and passes with the one-line guard above is available in a downstream project. Happy to contribute it upstream if helpful.

## Workaround

Patch `@mastra/core` locally (e.g. `pnpm patch`) with the `OM_INTERNAL_AGENT_IDS` early return in `waitForCrossAgentThreadRun`.

## Environment

- Node.js (production + local repro)
- `@mastra/core@1.35.0`, `@mastra/memory@1.18.2`
- `PostgresStore` in production; `InMemoryStore` sufficient for local repro
- AI SDK v6 / `MockLanguageModelV3` for deterministic local test

0xjackey:
## Problem

Observational Memory (OM) can run observer and reflector agents from inside another agent run. Those internal agents were receiving the parent run's `RequestContext`, including the parent `threadId`.

That made the internal OM run look like another run on the same active thread. Core's cross-agent thread coordination could then wait for the parent run to finish, while the parent run was already blocked waiting for the OM observation to finish. The result was a circular wait:

```text
main agent run
  └─ blocking OM observe()
      └─ OM observer/reflector agent.stream()
          └─ inherits parent threadId
              └─ waits for active parent thread run
                  └─ parent is waiting for OM observe()
```

This caused blocking OM observations to hang indefinitely.

## Solution

Isolate OM's internal agent runs at the stream call boundary.

When OM invokes an internal observer or reflector agent, it now clones the incoming `RequestContext` and derives an internal thread id from the parent thread id plus the OM agent id:

```text
<parentThreadId>-<omAgentId>
```

This preserves the rest of the request-scoped context while preventing nested OM runs from reusing the parent run's active thread id.

This avoids adding OM-specific agent-id allowlists to `@mastra/core`; the fix lives where the internal OM `agent.stream()` calls are made.

## Changes

- Added an OM internal request-context helper that derives an isolated thread id for internal OM agent runs.
- Updated the single-thread observer, multi-thread observer, and reflector stream calls to use the derived internal request context.
- Removed the previous core-level OM agent id special case.
- Added memory-package regression coverage for the derived thread identity passed into OM internal stream calls.
- Added a changeset for `@mastra/memory` / `@mastra/core` patch release metadata.

## Testing

Verified locally:

- `pnpm --filter ./packages/memory test src/processors/observational-memory/__tests__/internal-request-context.test.ts`
- `pnpm build:core`
- `pnpm --filter ./packages/memory build:lib`

## Related

Fixes #17827

## ELI5

Observational Memory (OM) sometimes pauses the main agent to run “extra” internal work. The bug was that these internal OM workers accidentally reused the parent’s “currently running” identifier, so the main agent and the OM workers waited on each other forever. This PR gives each internal OM worker its own unique thread ID so they can’t deadlock.

### Problem
When OM performs blocking observations, OM’s internal observer/reflector agent runs inherited the parent run’s `RequestContext` (including `threadId`). Core’s cross-agent thread coordination then treated the parent run as still active on that same thread and waited for it, while the parent run was waiting for the OM observation—creating a circular wait and permanent deadlock.

### Solution
In `@mastra/memory`, isolate OM’s internal agent runs at the stream-call boundary by deriving an internal request context with a new thread ID:
- Added `withOmInternalThreadId(requestContext, omAgentId)`:
  - returns `undefined` if `requestContext` is missing
  - returns the original `requestContext` if the parent `threadId` is missing or invalid
  - otherwise clones the context and sets `MASTRA_THREAD_ID_KEY` to `${parentThreadId}-${omAgentId}` (preserving other request-scoped entries like `tenantId` and resource identifiers)

### Changes Made
- **New helper:** `packages/memory/src/processors/observational-memory/internal-request-context.ts`
  - Implements `withOmInternalThreadId(...)`.
- **Updated OM internal runners to use derived request context:**
  - `packages/memory/src/processors/observational-memory/observer-runner.ts`
    - Uses derived `requestContext` for both single-thread (`call`) and multi-thread (`callMultiThread`) observer streams.
  - `packages/memory/src/processors/observational-memory/reflector-runner.ts`
    - Uses derived `requestContext` for reflector `agent.stream(...)`.
- **Regression tests:**
  - `packages/memory/src/processors/observational-memory/__tests__/internal-request-context.test.ts`
    - Verifies correct derived thread ID formation, preservation of key request-scoped fields, and that internal `agent.stream` calls receive the derived context while the parent context remains unchanged.
  - `packages/memory/src/processors/observational-memory/__tests__/observational-memory.test.ts`
    - Adjusts observer payload tests to pass a real `RequestContext` (including `MASTRA_THREAD_ID_KEY`) rather than a plain object.
- **Release metadata:** `.changeset/fix-om-internal-thread-ids.md` (patch release for `@mastra/memory`).

### Key Design Decision
Avoid hard-coding special-cases in `@mastra/core` by handling thread isolation inside `@mastra/memory` where OM triggers its internal observer/reflector streams.


TylerBarnes:
Hey @0xjackey — thanks again for tracking this down and opening the PR. The root cause analysis in the issue/PR was right: OM's internal observer/reflector agents were inheriting the parent run's `RequestContext`, including the active `threadId`, which let core's cross-agent thread coordination see the parent run as active and wait on it while the parent was waiting on OM.

I pushed a follow-up commit to your branch that keeps the same intended fix — preventing OM internal runs from waiting on the parent run — but moved the implementation out of `@mastra/core` and into the OM call sites in `@mastra/memory`.

What changed:

- Removed the core-level OM agent id allowlist from `thread-stream-runtime.ts`.
- Added a memory-side helper that clones the incoming `RequestContext` and derives an internal OM thread id:
  ```text
  <parentThreadId>-<omAgentId>
  ```
- Updated the observer, multi-thread observer, and reflector `agent.stream()` calls to use that derived request context.
- Added memory-package tests that verify those internal stream calls receive the derived thread id.
- Added a changeset and updated the PR title/description to reflect the new approach.

The reason for the change is to avoid coupling `@mastra/core` to hard-coded Observational Memory agent ids. Conceptually, the issue is not that these specific agent ids should be exempt from coordination; it's that OM is starting nested/internal agent runs and should not reuse the parent run's active thread id. Handling that at the `agent.stream()` boundary keeps core generic and makes the fix less fragile if the OM agent ids ever change.

Local verification from the follow-up:

- `pnpm --filter ./packages/memory test src/processors/observational-memory/__tests__/internal-request-context.test.ts`
- `pnpm build:core`
- `pnpm --filter ./packages/memory build:lib`

Latest commit on the PR branch is `b6d6d6c4d74` (`fix(memory): isolate OM internal thread runs`).

TylerBarnes (approved):

PR #17832, opened 2026-06-11
feat(channels): serverless-safe channel rendering with per-run output processor and waitUntil

CalebBarnes:
## Summary

Channels on serverless platforms (Vercel, Cloudflare Workers, Netlify, AWS Lambda) can receive the same webhook event across multiple concurrent invocations. Without coordination, each invocation renders the agent reply independently — producing duplicate messages, partial renders, and streams killed mid-flight when the Lambda freezes.

This PR makes `AgentChannels` serverless-safe by:

1. **Awaiting run ownership** — `processChatMessage` awaits `result.accepted` and only consumes the stream when the invocation wins the lease (`action === "wake"`). Losers resolve cleanly without rendering.

2. **Per-run output processor** — A new `ChatChannelOutputProcessor` renders stream chunks to the chat platform inline through the agent output processor pipeline, replacing the old background subscription approach. Registered via `Agent#listResolvedOutputProcessors`.

3. **Multi-step rendering** — The render session stays open when `stepResult.isContinued === true`, so multi-step runs render every step instead of dropping all but the first.

4. **`waitUntil` forwarding** — Channels and `SlackProvider` resolve `waitUntil` automatically for Cloudflare Workers and Netlify, with an explicit config option for Vercel/Lambda. Keeps the invocation alive until background rendering finishes.

5. **Concurrent chat-sdk strategy** — Switches from `queue` to `concurrent`, deferring ordering to the signals/lease layer and avoiding stale-lock issues in frozen serverless invocations.

6. **Unified approval rendering** — Tool approval/decline now flows through the output processor (same path as regular messages) instead of the legacy `ensureThreadSubscription` + `consumeAgentStream` path. The old subscription-based rendering code is removed; `close()` is deprecated as a no-op.

7. **Tool approval chunk visibility** — `tool-call-approval` chunks are routed through `outputWriter` so output processors (including `ChatChannelOutputProcessor`) can render approval buttons. Previously these chunks bypassed the processor pipeline entirely.

## Changed files

**Core channels:**
- `agent-channels.ts` — Refactored `processChatMessage` to await ownership + consume stream; unified approval/decline through output processor; removed `ensureThreadSubscription`, `consumeAgentStream`, `threadSubscriptions`; deprecated `close()`.
- `output-processor.ts` (new) — `ChatChannelOutputProcessor` with `createChunkQueue` (single-producer/single-consumer async queue), `RenderSession`, and `processOutputStream`.
- `wait-until.ts` (new) — `resolveWaitUntil(c)` helper for Cloudflare/Netlify auto-detection.
- `types.ts` — Added `waitUntil`, `resolveWaitUntil` to `ChannelConfig`.
- `agent.ts` — Channel output processors run before user-configured processors.

**Slack:**
- `provider.ts` — Resolves and forwards `waitUntil` for webhooks and slash commands; removed `existing?.close()` call.
- `types.ts` — Added `waitUntil`, `resolveWaitUntil` to `SlackProviderConfigBase`.

**Loop:**
- `tool-call-step.ts` — Routes `tool-call-approval` chunks through `outputWriter` (both `requireApproval` and `requireToolApproval` paths).

**Tests:**
- `agent-channels.test.ts` — Updated for output processor path; added wake/non-wake consumption assertions.
- `output-processor.test.ts` (renamed from `consume-agent-stream.test.ts`) — 52 tests covering streaming/static drivers, multi-step, error/abort boundaries, typing status, approval rendering through the processor.

**Docs:**
- `channels.mdx` — Added `waitUntil` and `resolveWaitUntil` parameter documentation.

## Test plan

- `pnpm build:core` ✅
- `pnpm --filter @mastra/core typecheck` ✅
- 198 channel tests pass (8 files)
- 8/8 manual smoke tests via Slack (DM, streaming, all toolDisplay modes, tool approval approve/deny flow, static mode)
- Verified no accidental reverts of recently-merged channel PRs (#18109, #17471, #17185, #17171)


hermes-leobrival (approved):
Approved as requested by Léo via Hermes.

CalebBarnes:
@coderabbitai review

CalebBarnes:
@coderabbitai full review

CalebBarnes:
@coderabbitai full review

TylerBarnes:
Nice! Only thing I noticed is this fixes duplicate rendering via run ownership, but channel thread creation is still list-then-save. If two first-ever messages for the same external thread race before a mapping exists, thread splitting may still be possible unless storage uniqueness/upsert handles it elsewhere.

CalebBarnes:
Good catch — that's pre-existing. The list-then-save race in `getOrCreateThread` exists on main today. The signals layer deduplicates run ownership (so two racing invocations won't both render), but thread creation itself could still split if two first-ever messages race before any mapping exists. Worth a follow-up with a storage-level upsert or unique constraint, but not something this PR introduces or makes worse.

abhiaiyer91 (approved):

PR #17836, opened 2026-06-11
feat(core): make agentic-loop serialization-safe on the evented engine

taofeeq-deru:
## Summary

Makes the agentic-loop's evented-engine path structurally serialization-safe. Step inputs/outputs are guaranteed JSON-safe and non-serializable runtime state (live class instances, closures, transport refs, save queues, memory handles) lives in a per-run scratch space (`RunScope`) that never crosses the wire. A new codec layer at the `UnixSocketPubSub` boundary preserves rich types (`Date`, `Error`, `Map`, `Set`, `RegExp`, `URL`, `BigInt`, `undefined`, registered classes like `DefaultGeneratedFile`) on round-trip across processes.

Before this change, the agentic-loop relied on a convention: keep non-serializable values out of step I/O and pass them via a `_internal` bag carried through closures. That convention was easy to break, and any rich value that reached `UnixSocketPubSub.publish` was either silently flattened or crashed the publisher.

### Before / after on the wire

This was validated against a hermetic Mastra project linked to dist tarballs built from both `main` HEAD and this branch. Both runs published the same payload over a real `UnixSocketPubSub` from a broker process to a forked client process.

| Payload field on the wire     | BEFORE (`main` dist)                          | AFTER (this branch dist)                |
| ----------------------------- | --------------------------------------------- | --------------------------------------- |
| `BigInt`                      | **`TypeError: Do not know how to serialize a BigInt` — publisher crashes** | preserved as `bigint`                   |
| `DefaultGeneratedFile`        | same crash + tree-shaken from dist            | instance preserved, `.base64` getter intact |
| `Date`                        | silently → ISO string on receiver             | `Date` instance                         |
| `RegExp`                      | silently → `{}` on receiver                   | `/source/flags` preserved               |
| `URL`                         | silently → string on receiver                 | `URL` instance                          |
| `Set` (size 3)                | silently → `{}` (size 0) on receiver          | `Set` size 3, entries preserved         |
| `Map` (size 2)                | silently → `{}` (size 0) on receiver          | `Map` size 2, entries preserved         |
| `Error` with custom field     | silently → plain object, `.message` lost      | `Error` instance, message + custom fields preserved |
| In-process publish (same proc)| works (in-memory shortcut, no serialization)  | works (same shortcut, codec bypassed)   |

Single-process workflows and Studio chat aren't affected either way — `UnixSocketPubSub.publish` short-circuits same-process publishes through `#deliverLocal` and never reaches the codec. The breakage was, and the fix is, scoped to the cross-process wire.

### Approach

**`RunScope` per-run scratch space**

- `RunScope` (`Map<symbol, unknown>`) stores non-serializable runtime state (live instances, closures, transport refs, memory handles, background task managers) keyed by `runId`, addressed by typed `RunScopeKey<T>` symbols
- Refcounted lifecycle: allocated by `__registerInternalWorkflow`, released by `__unregisterInternalWorkflow`, swept by the existing TTL sweeper, and kept alive across suspend/resume on the evented engine
- Never persisted to storage, never published over pubsub
- Bridges the wire boundary: step I/O stays JSON-safe; live handles ride in scope

**Codec at the `UnixSocketPubSub` boundary**

- `serializeFrame` / `readFrames` now wrap / unwrap rich types using tagged envelopes (`CODEC_TAG = '__m_codec__'`) with a closed `EnvelopeTag` discriminator
- Supports `Date`, `Error` (incl. `cause` chains and enumerable own props), `Map`, `Set`, `RegExp` (with source length cap + flag whitelist as ReDoS defense), `URL`, `BigInt`, `undefined`, and registered classes via a `ClassCodec` registry
- Cycle-safe via `WeakSet` — repeat references encode as `null` (matches `JSON.stringify`'s "first visit wins" behavior without throwing)
- In-process `EventEmitterPubSub` and same-process `UnixSocketPubSub.#deliverLocal` bypass the codec — object references pass through unchanged for local subscribers

**Workflow migration**

- `prepare-stream`, `agentic-execution`, `agentic-loop` migrated to read/write via `readScoped` / `writeScoped` helpers
- `StreamInternal` marked `@deprecated` per-field; bootstrap values hydrated once at loop start via `hydrateRunScopeFromInternal`. Runtime-write fields (`stepTools`, `stepActiveTools`, `stepWorkspace`, `_delegationBailed`) are intentionally skipped — they're produced during execution, not bootstrap input
- Step output schemas are now JSON-safe by construction; all live handles (memory, save queues, transport refs, tool registries, processor state) live in `RunScope`

### Implementation details

**New modules**

- `run-scope.ts` — `RunScope` interface, `MapRunScope` implementation, typed `RunScopeKey<T>` factory
- `loop/run-scope-keys.ts` — 22 typed keys for the agentic loop (`NOW_KEY`, `GENERATE_ID_KEY`, `STEP_TOOLS_KEY`, `DRAIN_PENDING_SIGNALS_KEY`, etc.)
- `prepare-stream/run-scope-keys.ts` — 5 typed keys for prepare-stream (`MESSAGE_LIST_KEY`, `CONVERTED_TOOLS_KEY`, `PROCESSOR_STATES_KEY`, etc.)
- `loop/run-scope-access.ts` — `getRunScope`, `readScoped`, `writeScoped` helpers (read prefers scope, falls back to `_internal`; write mirrors to `_internal` for backward compat)
- `loop/hydrate-run-scope.ts` — bootstraps scope from `_internal` once at loop start
- `events/codec/{codec,tags,registry,error,registrations}.ts` — codec primitives + built-in class codecs (`DefaultGeneratedFile`, `DefaultGeneratedFileWithType`, `DefaultStepResult`)

**Mastra core**

- `Mastra` manages `#runScopes` and `#runScopeRefcounts`
- New internal methods: `__getRunScope(runId)`, `__createRunScope(runId)`, `__releaseRunScope(runId)`
- TTL sweeper releases scopes when evicting stale internal workflow registrations
- `__registerInternalWorkflow` / `__unregisterInternalWorkflow` create / release scope refcounts as part of their existing lifecycle

**Tree-shaking fix**

The class codec registry (`DefaultGeneratedFile`, `DefaultGeneratedFileWithType`, `DefaultStepResult`) was originally activated by a side-effect import (`import './registrations'`) from the codec barrel. With `"sideEffects": false` in `packages/core/package.json`, the bundler dropped that import, and the registrations were missing from the published dist — `DefaultGeneratedFile` instances would lose their prototype on cross-process publish even with the codec in place. The dist reproducer from the smoke test confirmed this: `instanceof DefaultGeneratedFile` was `true` on send and `false` on receive.

Fix: registrations live in a named export (`BUILTIN_CODECS_REGISTERED`) that `codec.ts` imports and references with `typeof BUILTIN_CODECS_REGISTERED`, which the bundler retains. Verified by inspecting `dist/events/index.js` (4 `registerClass` calls present, was 0) and by the cross-process `DefaultGeneratedFile` round-trip test that loads from dist.

### Test coverage

**Codec round-trip**
- `codec.test.ts` — every envelope type (Date, Error, Map, Set, RegExp, URL, BigInt, undefined, Class) with cycles, `toJSON`, hostile flags, oversized source, malformed source fallback, class registry, registry-collision safety
- `unix-socket-pubsub-codec.test.ts` — same types over a real same-process broker+client
- `cross-process-codec-roundtrip.test.ts` — same types over a real `fork()`ed broker+client loaded from dist, including `DefaultGeneratedFile` (which catches the tree-shaking bug)

**RunScope lifecycle**
- `run-scope.test.ts` — storage semantics, key collision safety, refcounting, idempotent `__createRunScope`, register/unregister pairing, TTL eviction
- `run-scope-helpers.test.ts` — hydrate field contract (populated vs intentionally skipped), `readScoped` scope-then-`_internal` fallback, `writeScoped` mirror
- `runscope-leak.test.ts` — parallel runs (per-run scope isolation), TTL sweep on abandoned suspend, many sequential runs (no accumulation), for both default and evented engines
- `agentic-loop-snapshot-lifecycle.test.ts` — scope release across suspend/resume/decline/fail, engine-aware (default holds across suspend, evented releases at suspend boundary)

**Serialization invariants**
- `serialization-invariants.test.ts` — step output schemas exclude forbidden handle/config keys; representative payloads do not leak Class/Function envelopes at the codec boundary
- `prepare-stream/run-scope.test.ts` — step output schemas are JSON-safe; `RunScope` carries non-wire-safe handles; codec boundary never sees them
- `step-output-type-preservation.test.ts` — full default + evented engine validation that Date, Error, Map, Set, RegExp, URL, BigInt, `DefaultGeneratedFile`, `DefaultStepResult`, and cycles preserve types through workflow snapshot/restore

**Transport ↔ consumer**
- `unix-socket-pubsub-redelivery-budget.test.ts` — `MAX_LOCAL_REDELIVERIES` >= `WorkflowEventProcessor.MAX_DELIVERY_ATTEMPTS` so the consumer always gets its full retry budget
- `redelivery-end-to-end.test.ts` — real WEP + real `UnixSocketPubSub`: K-1 transient failures then success (pubsub redelivers, WEP eventually acks, no `workflow.fail`) and K+1 transient failures (WEP trips `MAX_DELIVERY_ATTEMPTS`, publishes `workflow.fail` exactly once, pubsub stops)

### Backward compatibility

- `_internal` field on `streamVNext()` options continues to work; values are hydrated into `RunScope` once at loop start and marked `@deprecated`
- `readScoped` / `writeScoped` mirror writes to `_internal` so legacy code reading from the bag still observes mutations
- In-process `EventEmitterPubSub` bypasses codec; continues to pass object references as before
- Same-process `UnixSocketPubSub` publishes (same broker + same client process) short-circuit through `#deliverLocal` and never touch the codec
- No changes to exported workflow APIs; migration is transparent to callers

### Documentation

- `packages/core/AGENTS.md` — documents `runScope` contract, lifecycle, and the JSON-safety constraint on step I/O
- `.changeset/pink-poets-joke.md` — notes resolution of intermittent "getFullOutput is not a function" and "Workflow not found" errors on multi-instance deployments

### Validation

- Core typecheck passes
- Core tests: 9807 pass, 115 skipped, 19 todo (1 unrelated pre-existing flake on `resume-api.test.ts` `cachingPubsub.getHistory` race)
- Smoke-tested locally against a hermetic create-mastra project linked to dist tarballs from both `main` and this branch:
  - **Option A**: single-process evented workflow with codec-typed step output — passes on both branches (in-memory shortcut, doesn't exercise codec)
  - **Option B**: forked broker + client over `UnixSocketPubSub` with `DefaultGeneratedFile` + `BigInt` — `main` **crashes** at `serializeFrame` with `TypeError: Do not know how to serialize a BigInt`; this branch preserves all 7 type assertions including `instanceof DefaultGeneratedFile` and `.base64`
  - **Option E**: forked broker + client with `Date`/`RegExp`/`URL`/`Set`/`Map`/`Error` (no BigInt, so `main` doesn't crash) — `main` silently flattens every value on the receiver; this branch preserves all of them
  - **Option C**: real `gpt-5-mini` agentic chat in Studio with `MASTRA_EVENTED_EXECUTION=true` invoking a tool that returns the full codec-soak payload — completes with `SUCCESS`, no `AGENT_GENERATE_MALFORMED_RESULT`, snapshot persisted, trace recorded

## ELI5
When the agentic loop runs on the evented setup, it may pause/resume and also send workflow updates between processes. This PR ensures only “safe data” crosses the wire (so types aren’t corrupted), while “live stuff” like class instances and runtime handles stays in a per-run in-memory space.

---

## Overview
This PR makes the agentic-loop evented-engine path serialization-safe by:
1) introducing a per-run `RunScope` scratch space for runtime-only state that must never be serialized/published, and  
2) adding a codec layer at the `UnixSocketPubSub` boundary so JSON-safe frames can round-trip rich values across processes.

---

### Core Infrastructure: per-run `RunScope` scratch space
- Added a `RunScope` abstraction using typed `RunScopeKey<T>` symbol keys backed by a `Map` for runtime-only values.
- `Mastra` now manages `RunScope` lifecycles per `runId` using refcounting:
  - allocated on internal workflow registration
  - released on unregistration
  - released by TTL eviction using stored metadata (no fragile key parsing)
- Added `hydrateRunScopeFromInternal(mastra, runId, internal?)` to bootstrap selected inputs from the legacy `StreamInternal` `_internal` bag into `RunScope` when runs start/resume.
- Added typed helpers:
  - `readScoped(ctx, key, internalField)` (prefer `RunScope`, fallback to `_internal`)
  - `writeScoped(ctx, key, internalField, value)` (write to `RunScope` and mirror to `_internal` for legacy behavior)

### Codec at the `UnixSocketPubSub` boundary
- Implemented tagged, JSON-safe `encode`/`decode` for pubsub frames to preserve:
  - `Date`, `Error` (including `cause` chains and enumerable custom fields), `Map`, `Set`, `RegExp`, `URL`, `BigInt`, and explicit `undefined`
  - registered class instances via a class codec registry
- Hardened codec behavior:
  - cycle-safe encoding (recurrence handled to match JSON-style semantics)
  - bounded/validated `RegExp` decoding to avoid hostile reconstruction
  - malformed/unknown envelopes decode safely as plain data
- Updated `UnixSocketPubSub` to route frames through `encode`/`decode` wrappers (`serializeFrame` / `readFrames`).
- Kept local/same-process delivery behavior from unnecessarily rehydrating objects by bypassing codec where appropriate.
- Refactored the local redelivery budget constant to module scope (`MAX_LOCAL_REDELIVERIES`).

### Tree-shaking fix for built-in codec registrations
- Ensured built-in codec registrations are not dropped by bundlers with `sideEffects: false` by gating them behind a named export:
  - `BUILTIN_CODECS_REGISTERED`
- Codec encoding validates that built-in registrations are present before use.

---

### `prepare-stream`: move shared state into RunScope keys
- Migrated internal step sharing to typed `RunScope` keys (`MESSAGE_LIST_KEY`, `CONVERTED_TOOLS_KEY`, `PROCESSOR_STATES_KEY`, `INITIAL_SIGNAL_ECHOES_KEY`, `LOOP_OPTIONS_KEY`) using `runScope.get/set`.
- Refactored `PrepareStreamRunScope` to be an alias of the shared `RunScope` type.
- Marked `StreamInternal` usage/dependencies as deprecated guidance.

### Agentic execution / agentic loop: scoped state via `readScoped` / `writeScoped`
- Migrated multiple steps to store and read runtime values from `RunScope` instead of mutating `_internal`, including:
  - `llm-execution-step`, `tool-call-step`, `llm-mapping-step`, `signal-drain-step`, `background-task-check-step`, `is-task-complete-step`
- Updated agentic loop control flow to consult/update scoped keys (e.g., pending signals, messageId generation, delegation-bailed flag) rather than mutating `_internal`.
- Preserved backward compatibility by mirroring writes to `_internal` when present and hydrating supported bootstrap inputs into the scope at loop start.

---

## Deprecation / Compatibility
- `_internal` / `StreamInternal` remains supported, with `@deprecated` JSDoc steering new logic toward `RunScope`.
- Scoped helpers bridge `RunScope` and `_internal` so existing expectations continue to work.

---

## Tests Added/Expanded
- Codec boundary tests: comprehensive round-trips and resilience cases (Date/Error/Map/Set/RegExp/URL/BigInt/undefined, cycles, hostile inputs, envelope/tag collisions, class registry preservation).
- Unix socket end-to-end tests: codec round-trips over a real broker↔client connection, including `DefaultGeneratedFile` and nested mixed payloads.
- Cross-process regression: detects codec/class-registry loss caused by tree-shaking.
- RunScope invariants: lifecycle, refcounting, register/unregister pairing, TTL eviction, and isolation/leak-vector behavior.
- Workflow serialization/type preservation:
  - serialization invariants for agentic execution/loop schemas (ensuring only JSON-safe fields are in advertised output schemas)
  - end-to-end payload type preservation across default vs evented engines
  - invariant test tying UnixSocketPubSub local redelivery budget to consumer retry attempts.


packages/core/src/events/codec/codec.ts
```diff
@@ -0,0 +1,166 @@
...
+    const env = value as Envelope;
+    switch (env[CODEC_TAG]) {
+      case 'Undefined':
+        return undefined;
+      case 'Date':
+        return new Date(env.v);
+      case 'BigInt':
+        return BigInt(env.v);
+      case 'RegExp':
+        return new RegExp(env.v.source, env.v.flags);
```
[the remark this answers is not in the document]
taofeeq-deru: Addressed in e5989c34f3.

The codec runs at the intra-host `UnixSocketPubSub` frame boundary so the trust model is process-local, but a corrupted/hostile frame still shouldn't be able to crash decoding or construct an unexpected regex. Two layers of defense:

1. `isEnvelope` now bounds `flags` to the spec-defined set (`d g i m s u y`, no duplicates). Anything else short-circuits envelope detection and the payload is preserved as user data — `new RegExp(...)` is never called.
2. `decode` wraps `new RegExp(...)` in a try/catch and returns the raw envelope when `source` is unparseable, instead of throwing.

Tests added in `codec.test.ts` cover both hostile flags and unparseable source.

.changeset/pink-poets-joke.md
```diff
@@ -0,0 +1,10 @@
...
+- Introduced a per-run `RunScope` on `Mastra`, keyed by `runId` and refcounted alongside `__registerInternalWorkflow`. It holds non-serializable runtime state (`MessageList`, processor states, converted tools, loop options, `SaveQueueManager`, `BackgroundTaskManager`, `MastraMemory`, `StreamTransportRef`, dynamic `ToolSet`, `Workspace`, drain-signal closures, …) that must never cross the wire. The scope is never persisted, never published over pubsub, and dies with the run.
+- Migrated the `prepare-stream`, `agentic-execution`, and `agentic-loop` step factories to read and write this state through typed `RunScopeKey<T>` keys instead of the legacy `_internal` closure bag. The corresponding `StreamInternal` fields are marked `@deprecated` for external callers (the type still accepted by `loop()` for back-compat; `loop()` hydrates the scope from it once at the single bootstrap point).
+- Added a serialization-invariants test that asserts none of the agentic step output schemas advertise live handle keys and that encoded sample outputs never produce class or function codec envelopes.
```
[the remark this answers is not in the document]
taofeeq-deru: Rewritten in 6428d07e44 as outcome-first developer-facing notes (no more `__registerInternalWorkflow` / `_internal` internals), with the multi-instance evented hang as the headline fix line.

packages/core/src/events/codec/codec.ts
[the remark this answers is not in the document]
taofeeq-deru: Addressed in e5989c34f3.

Two-layer defense:
- `isEnvelope` bounds `flags` to the spec-defined set (`d g i m s u y`, no duplicates) so malformed/hostile flags short-circuit envelope detection and the payload is kept as raw user data — `new RegExp(...)` is never reached.
- `decode` wraps `new RegExp(...)` in a try/catch and returns the raw envelope when `source` is unparseable.

Tests for both paths in `codec.test.ts`.

packages/core/src/loop/workflows/serialization-invariants.test.ts
```diff
@@ -0,0 +1,167 @@
...
+        // Unwrap common wrappers
+        const def = (s as any)._def ?? (s as any).def;
+        if (!def) return;
+        const inner =
+          def.innerType ?? def.schema ?? def.type ?? def.element ?? def.valueType ?? def.left ?? def.options;
+        if (Array.isArray(inner)) {
+          for (const i of inner) visit(i as z.ZodTypeAny);
+        } else if (inner) {
+          visit(inner as z.ZodTypeAny);
+        }
```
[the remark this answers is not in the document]
taofeeq-deru: Addressed in 6428d07e44. `collectKeys` now walks every child container (`innerType`, `schema`, `type`, `element`, `valueType`, `keyType`, `left`, `right`, `options`, `items`) so unions, intersections, records, and tuples are fully traversed.

packages/core/src/mastra/index.ts
```diff
@@ -2592,6 +2667,11 @@ export class Mastra<
       if (now - registeredAt > Mastra.INTERNAL_WORKFLOW_TTL_MS) {
         delete this.#internalMastraWorkflows[key];
         this.#runScopedWorkflowTimestamps.delete(key);
+        // key shape is `${workflowId}:${runId}` — release the matching scope.
+        const colon = key.lastIndexOf(':');
+        if (colon !== -1) {
+          this.#releaseRunScope(key.slice(colon + 1));
+        }
```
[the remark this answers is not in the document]
taofeeq-deru: Good catch — addressed in 6428d07e44.

`#runScopedWorkflowTimestamps` now stores `{ registeredAt, runId }`, and the TTL sweep releases scopes using the stored `runId` directly instead of re-parsing it from the composite key. Added a regression test in `run-scope.test.ts` using a tenant-style runId containing `:`.

packages/core/src/events/codec/codec.ts
```diff
@@ -0,0 +1,183 @@
...
+      case 'RegExp': {
+        // `isEnvelope` already constrained `flags` to the spec-defined set. A
+        // malformed or hostile `source` is escaped so untrusted input is
+        // treated literally; if construction still fails, fall back to the raw
+        // envelope so a hostile peer cannot crash frame decoding.
+        try {
+          return new RegExp(escapeRegExpLiteral(env.v.source), env.v.flags);
+        } catch {
+          return value;
+        }
```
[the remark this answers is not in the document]
taofeeq-deru: Good catch — addressed in cf2d901f4b.

You're right that the CodeQL autofix overshot. Dropped `escapeRegExpLiteral` from the decode path; the existing defense layers from e5989c34f3 are sufficient:
- `isEnvelope` already constrains `flags` to the spec-defined set (`d g i m s u y`, no duplicates), so hostile flags never reach the constructor.
- `decode` keeps the try/catch fallback, so a malformed/hostile `source` returns the raw envelope instead of crashing frame decoding.

Added a regression test covering `/foo.*/gi`, `/\d+-\d+/`, `/[a-z]{2,4}/i`, `/^(yes|no)$/`, `/a\.b\\c/` with both source/flags equality and behavioural match parity, so any future "literalize on decode" attempt gets caught.

packages/core/src/events/codec/registrations.ts
```diff
@@ -0,0 +1,14 @@
...
+
+registerClass<DefaultGeneratedFile, GeneratedFileData>('DefaultGeneratedFile', {
+  toData: f => ({ data: f.base64, mediaType: f.mediaType }),
+  fromData: d => new DefaultGeneratedFile({ data: d.data, mediaType: d.mediaType }),
+});
+
+registerClass<DefaultGeneratedFileWithType, GeneratedFileData>('DefaultGeneratedFileWithType', {
+  toData: f => ({ data: f.base64, mediaType: f.mediaType }),
+  fromData: d => new DefaultGeneratedFileWithType({ data: d.data, mediaType: d.mediaType }),
+});
```
NikAiyer: **[P0] Register `DefaultStepResult` here.**

This is the only real behavior gap I found. Only `DefaultGeneratedFile` and `DefaultGeneratedFileWithType` are registered today. `DefaultStepResult` instances flow through `steps: z.array(z.any())` on the workflow output schema, and the in-memory storage path (`storage/domains/workflows/inmemory.ts`) already has an `instanceof DefaultStepResult` consumer pattern. On the evented engine, consumers expecting a class instance will receive plain data instead.

The in-process clone path (`deepCloneForRun`) preserves prototype via direct prototype linking, but that doesn't help the cross-process case.

`DefaultStepResult` is a flat data-holder — the codec should be trivial.

**Follow-up (not blocking):** brief audit pass for any other classes that cross step boundaries via `z.any()`-typed schema fields.

packages/core/src/events/codec/codec.ts
```diff
@@ -0,0 +1,197 @@
...
+ * purpose is to round-trip a pattern, so metacharacters must reach
+ * `new RegExp(...)` verbatim. Safety comes from the bounded length, the
+ * flag whitelist, and the `try/catch` fallback, not from escaping.
+ */
+function decodeRegExpEnvelope(v: unknown): RegExp {
+  if (!v || typeof v !== 'object') return /(?:)/;
+  const candidate = v as { source?: unknown; flags?: unknown };
+  if (typeof candidate.source !== 'string') return /(?:)/;
+  if (typeof candidate.flags !== 'string') return /(?:)/;
+  if (candidate.source.length > 1024) return /(?:)/;
```
NikAiyer: **[P1] Import `MAX_REGEXP_SOURCE_LENGTH` from `tags.ts` instead of the literal `1024`.**

The constant is already declared in `tags.ts:16` (`MAX_REGEXP_SOURCE_LENGTH = 1024`) but this site checks the literal directly. Two sources of truth for the same value — easy drift target if someone tightens or relaxes the cap in one place and not the other.

One-line DRY fix: import it and replace the literal.

packages/core/src/loop/hydrate-run-scope.ts
```diff
@@ -0,0 +1,58 @@
...
+  if (internal.transportRef) scope.set(TRANSPORT_REF_KEY, internal.transportRef);
+  if (internal.backgroundTaskManager) scope.set(BACKGROUND_TASK_MANAGER_KEY, internal.backgroundTaskManager);
+  if (internal.agentBackgroundConfig) scope.set(AGENT_BACKGROUND_CONFIG_KEY, internal.agentBackgroundConfig);
+  if (internal.backgroundTaskManagerConfig)
+    scope.set(BACKGROUND_TASK_MANAGER_CONFIG_KEY, internal.backgroundTaskManagerConfig);
+  if (internal.skipBgTaskWait !== undefined) scope.set(SKIP_BG_TASK_WAIT_KEY, internal.skipBgTaskWait);
+  if (internal.drainPendingSignals) scope.set(DRAIN_PENDING_SIGNALS_KEY, internal.drainPendingSignals);
+  if (internal.initialSignalEchoes) scope.set(INITIAL_SIGNAL_ECHOES_KEY, internal.initialSignalEchoes);
+  if (internal.toolPayloadTransform) scope.set(TOOL_PAYLOAD_TRANSFORM_KEY, internal.toolPayloadTransform);
+}
```
NikAiyer, line 66: **[P2] Add an inline comment listing the skipped `StreamInternal` fields and why.**

This function copies most fields from `_internal` into the scope but intentionally skips `stepTools`, `stepActiveTools`, `stepWorkspace`, and `_delegationBailed`. I confirmed the skip list is correct — those are runtime-written outputs of step execution, not bootstrap inputs, so hydrating them would copy stale/empty values that get overwritten on the first step. And the one case that looks like a bootstrap input (`stepTools` from `ToolSearchProcessor` via `resolveInternalState`) still works because `readScoped` falls back to `_internal[field]` when the scope slot is `undefined`.

The top-level docstring covers the bootstrap-once pattern, but a short inline comment naming the four skipped fields would make the invariant ("scope keys are produced by step writes, not seeded at boot") obvious to the next reader.

packages/core/src/events/codec/tags.ts
```diff
@@ -0,0 +1,69 @@
+import type { SerializedError } from '../../error/utils';
+
+/**
+ * Discriminator key for codec-tagged envelopes. Long, namespaced, and unlikely
+ * to collide with user data. Plain objects that happen to carry this key but
+ * do not match an envelope shape are preserved as-is by the decoder.
+ */
+export const CODEC_TAG = '__m_codec__';
```
NikAiyer, line 16: **[P2] Add a short doc note about the `__m_codec__` reservation contract.**

If user data contains a key literally named `__m_codec__` and its shape happens to match an envelope, the decoder will reconstruct it as that envelope type. In practice this is very low risk thanks to (a) the namespaced key, (b) `isEnvelope`'s conservative shape check that narrows the blast radius to decoder-reconstructable types, and (c) `toJSON()` being skipped for objects already carrying `CODEC_TAG`. Encoder-side escaping would add complexity for a near-zero-probability case.

Worth documenting the namespace contract here so downstream consumers aren't surprised — "`__m_codec__` is reserved on the wire; do not use as a user key on objects that cross pubsub."

packages/core/src/mastra/run-scope.ts
```diff
@@ -0,0 +1,90 @@
+/**
```
NikAiyer, line 1: **[P3, follow-up] Consider a runtime-warning log when the TTL sweep evicts a still-registered run.**

The refcounted lifecycle (register/unregister + `__createRunScope`/`__releaseRunScope`) is correct, and the `finally` block in `stream.ts` together with the safety unregister in `agent.ts` cover the resume-failure path. A scope can still be evicted by the TTL sweep if a run is suspended and then garbage-collected (long-suspended runs that no resume ever comes back for).

Not blocking for this PR, but a warning log in `#sweepStaleRunScopedWorkflows` when it evicts a registration would give operators a signal for this case rather than silent cleanup.

NikAiyer (commented):
## Review notes

I walked the codec, RunScope lifecycle, and the agentic-loop migration end-to-end. Overall the design is sound — the codec is well-defended (RegExp narrowing, conservative envelope shape check, cycle resilience), the RunScope refcount lifecycle is correct, and the migration from `_internal` to RunScope preserves backward compat cleanly.

I'm leaving inline comments on the handful of items worth addressing, and a few non-action notes for things I checked and concluded were fine.

### Priority order

| Priority | Item | Why |
|---|---|---|
| **P0** | Register `DefaultStepResult` in `registrations.ts` | Only actual behavior gap. Cheap to land, expensive to debug later as flaky cross-process behavior. |
| **P1** | Import `MAX_REGEXP_SOURCE_LENGTH` in `codec.ts` | One-line DRY fix, prevents future drift. |
| **P2** | Skip-list comment in `hydrate-run-scope.ts` | Makes the runtime-write-vs-bootstrap invariant explicit. |
| **P2** | Namespace doc note in `tags.ts` | Documents the `__m_codec__` reservation contract. |
| **P3** | TTL-sweep eviction warning log | Operator visibility for long-suspended runs; not blocking for this PR. |
| **P3** | Audit other classes crossing step boundaries | Single-pass review of `steps`/`chunks`/`metadata` payloads beyond `DefaultStepResult`. |

Things I checked and concluded were fine (no inline comment needed): codec cycle handling returning `null` (intentional, JSDoc already documents it); the `writeScoped` mirror to `_internal` (purely a test affordance — `_internal` is built fresh per call in `stream-step.ts`); scope leak on resume failure (the `finally` block in `stream.ts` + safety unregister in `agent.ts` cover it); the `__`-prefixed method naming (already an established `Mastra` convention).

NikAiyer (approved):
Approving. Leaving three small follow-ups as inline comments (test fixture path length on macOS, CodeQL suppression rationale, MaxListenersExceededWarning in runscope-leak) — none blocking.

packages/core/src/loop/test-utils/aimock/scenarios/evented-unix-pubsub.scenario.test.ts
```diff
@@ -0,0 +1,112 @@
...
+describe('AIMock loop scenario: evented + UnixSocketPubSub', () => {
+  const getMock = useLoopScenarioAimock();
+  const pubsubs: UnixSocketPubSub[] = [];
+
+  afterEach(async () => {
+    await Promise.allSettled(pubsubs.splice(0).map(p => p.close()));
+  });
+
+  it('round-trips a tool result containing Date/Map/Error across the evented pubsub', async () => {
+    const socketPath = join(tmpdir(), `aimock-evented-${randomUUID()}.sock`);
```
NikAiyer, line 33: Nit / followup: on macOS hosts with a long `TMPDIR` (e.g. `/var/folders/.../T/`), this resolves to a path of ~109 chars and exceeds the 104-byte `sun_path` limit — `server.listen()` then throws `EINVAL`. The other unix-socket tests in this PR pass on the same host because their paths happen to be shorter.

Suggest `mkdtempSync("/tmp/aim-")` or a shorter prefix + trimmed UUID so the test runs portably regardless of `TMPDIR` length. Not a regression in the PR's runtime code — purely a fixture issue.

packages/core/src/events/codec/codec.ts
```diff
@@ -0,0 +1,212 @@
...
+  if (!v || typeof v !== 'object') return /(?:)/;
+  const candidate = v as { source?: unknown; flags?: unknown };
+  if (typeof candidate.source !== 'string') return /(?:)/;
+  if (typeof candidate.flags !== 'string') return /(?:)/;
+  if (candidate.source.length > MAX_REGEXP_SOURCE_LENGTH) return /(?:)/;
+  const flags = candidate.flags;
+  if (!/^[dgimsuvy]*$/.test(flags)) return /(?:)/;
+  if (new Set(flags).size !== flags.length) return /(?:)/;
+  try {
+    // lgtm[js/regex-injection] -- payload is narrowed (length cap, flag whitelist,
```
NikAiyer, line 150: Nit: the suppression is correct — safety comes from the source-length cap, flag whitelist, and try/catch fallback in `isEnvelope`/`decodeRegExpEnvelope`, not from escaping (which is why dropping `escapeRegExpLiteral` in 1938a77d37 was the right call).

Worth a one-line comment here naming those three structural gates, since a future reader scanning for `lgtm[` won't see the `isEnvelope` checks on the same screen and might be tempted to "fix" it by reintroducing escaping.

packages/core/src/agent/__tests__/runscope-leak.test.ts
```diff
@@ -0,0 +1,262 @@
...
+    );
+
+    // Every scope released — proves refcounted register/unregister fires
+    // per-run with no cross-run interference.
+    for (const runId of runIds) {
+      expect(mastra.__getRunScope(runId)).toBeUndefined();
+    }
+  }, 30000);
+
+  it('does not accumulate scopes across many sequential runs', async () => {
```
NikAiyer, line 102: Followup (not blocking): when this test runs, Node prints

```
MaxListenersExceededWarning: Possible EventEmitter memory leak detected.
11 workflows-finish listeners added to [EventEmitter].
```

The scope refcount assertions still hold, so it's not a RunScope leak. But it suggests `workflows-finish` listeners are accumulating on a shared emitter across the 20 sequential runs and being GC'd lazily rather than unsubscribed at run terminal. Worth a peek to confirm there's no slow leak on something this test doesn't assert.

PR #17856, opened 2026-06-11
fix(schema-compat): publish json-schema type dependency

### LINKED ISSUE #17826: [BUG] createTool inputData is `any` because @mastra/schema-compat declares @types/json-schema only in devDependencies
### Describe the Bug

`createTool({ inputSchema, execute })` types the `execute` callback's `inputData` parameter as `any` instead of `z.infer<inputSchema>`, so there is **no type-checking on tool inputs**. (The `outputSchema`-derived return type is unaffected, which makes the problem easy to overlook.)

The trigger is **not** the TypeScript version, the Zod version, or `@mastra/core`'s own type declarations. It is a **dependency-scoping bug** in `@mastra/schema-compat`: its public `.d.ts` imports `JSONSchema7` from the untyped `json-schema` runtime package, while `@types/json-schema` (which actually provides that type) is declared only in **`devDependencies`**. When nothing else in the consumer tree pulls `@types/json-schema` in transitively, `JSONSchema7` resolves to `any`, which poisons the `PublicSchema` union and collapses `InferPublicSchema` to `any`.

This is related to — but distinct from — the closed #16528, which attributes the `any` to a TS contextual-typing limitation / Zod v4 class hierarchy. This report shows the result is purely a function of whether `@types/json-schema` is present (TS and Zod versions held constant), and pins the cause to a `package.json` dependency scope.

### Steps To Reproduce

```bash
git clone https://github.com/oschcz/mastra-tool-inference-repro
cd mastra-tool-inference-repro
npm install

# 1) BUG: typecheck PASSES with no error, even though src/tool.ts calls the
#    number method `.toFixed()` on a string (inputData.location). Proves `any`.
npm run check        # -> exit 0, no output   (BUG)

# 2) Install the missing types package and re-check:
npm i -D @types/json-schema
npm run check        # -> src/tool.ts: error TS2551: Property 'toFixed' does not
                     #    exist on type 'string'.   (inference restored)
```

`src/tool.ts`:

```ts
import { z } from "zod";

createTool({
  id: "get-weather",
  description: "Get weather",
  inputSchema: z.object({ location: z.string() }),
  execute: async (inputData) => {
    // inputData is `any` — this compiles with NO error even though `.toFixed`
    // is a number method on a string. Should error if inputData were typed.
    inputData.location.toFixed(2);
    return { ok: true };
  },
});
```

https://github.com/oschcz/mastra-tool-inference-repro

### Expected Behavior

`inputData` should be inferred as `z.infer<typeof inputSchema>` (here `{ location: string }`) automatically, so that `inputData.location.toFixed(2)` is a compile-time type error.

### Environment Information

```
  System:
    OS: Windows 11 10.0.26200
  Binaries:
    Node: 24.16.0
    npm: 11.12.1
    pnpm: 11.5.2
  npmPackages:
    @mastra/core: 1.41.0
    @mastra/schema-compat: 1.2.11   (transitive)
    @mastra/ai-sdk: 1.4.4
    @mastra/libsql: 1.12.1
    @mastra/memory: 1.20.2
    @mastra/observability: 1.14.1
    @mastra/loggers: 1.1.2
    typescript: 6.0.3
    zod: 4.4.3
  tsconfig: strict: true, moduleResolution: "bundler", module: "esnext", skipLibCheck: true
```

---

### Root cause (verified)

`@mastra/schema-compat`'s public types entry (`dist/schema.d.ts`, the `"./schema"` export) contains:

```ts
// @mastra/schema-compat/dist/schema.types.d.ts
import type { JSONSchema7 } from 'json-schema';
export type PublicSchema<Output = unknown, Input = Output> =
  | z4.ZodType<Output, Input>
  | z3.Schema<Output, z3.ZodTypeDef, Input>
  | SchemaV4<Output> | SchemaV5<Output> | SchemaV6<Output>
  | JSONSchema7                       // <-- becomes `any` when @types/json-schema is absent
  | StandardSchemaWithJSON<Input, Output>;
export type InferPublicSchema<T extends PublicSchema> =
  T extends PublicSchema<infer Output> ? Output : never;
```

- The runtime `json-schema` package ships JS only — no `.d.ts`, no `JSONSchema7` type.
- `JSONSchema7` comes from `@types/json-schema`, which `@mastra/schema-compat@1.2.11` declares only in **`devDependencies`**, so it is not installed for consumers.
- With `skipLibCheck: true` (a common setting) the broken import resolves to `any` **silently**. With `skipLibCheck: false` TS reports it explicitly:

```
error TS7016: Could not find a declaration file for module 'json-schema'.
  '.../node_modules/json-schema/lib/validate.js' implicitly has an 'any' type.
  Try `npm i --save-dev @types/json-schema` if it exists ...
```

When `JSONSchema7` is `any`, `PublicSchema<Output>` has an `any` member, so `T extends PublicSchema<infer Output>` matches the `any` branch and `Output` collapses to `any` → `execute`'s input is `any`.

**Evidence — same `@mastra/core@1.41.0` + `zod@4.4.3`, only toggling `@types/json-schema`:**

| `@types/json-schema` | `skipLibCheck` | `inputData` | TS reports `json-schema` error? |
|---|---|---|---|
| absent  | true  | **`any`** (silent) | no |
| absent  | false | **`any`**          | **yes (TS7016)** |
| present | true  | `{ location: string }` ✅ | no |
| present | false | `{ location: string }` ✅ | n/a |

**Why it also looks like a 1.38 → 1.39 regression:** bisecting `@mastra/core`, `1.6.0`–`1.38.0` infer correctly and `1.39.0`+ break. This is not a code change in core — overlaying the full `1.39.0` `dist/` onto a `1.38.0` install still infers correctly, while `1.39.0`'s **dependency tree** breaks a `1.38.0` `dist/`. The `1.38.0` tree happened to pull `@types/json-schema` transitively (via `@standard-community/standard-json` / `openapi-types` / `quansync`); `1.39.0`+ dropped those, exposing the latent scoping bug.

### Proposed fix

Move `@types/json-schema` from `devDependencies` to **`dependencies`** in `packages/schema-compat/package.json` — a published `.d.ts` that imports from a `@types/*` package must depend on it as a regular dependency. Consider the same audit for `@mastra/core`, whose bundled ai-sdk provider `_types` also `import ... from 'json-schema'`.

Alternatively, vendor a local `JSONSchema7` type so `PublicSchema` doesn't rely on a `@types/*` package being hoisted into the consumer tree.

### Workarounds (consumer side)

1. `npm i -D @types/json-schema` (or `pnpm add -D @types/json-schema`) — fully restores automatic inference on current versions. ✅ verified
2. Annotate each tool: `execute: async (inputData: z.infer<typeof schema>) => { ... }`.
3. Pin `@mastra/core` to `1.38.0` (not recommended — only works because that tree happens to pull `@types/json-schema`).

### Related

- #16528 (closed) — same symptom, different attributed cause; still reproducible on `@mastra/core@1.41.0`.
- #14194 / #14195 — prior "untyped import breaks createTool inference" class of bug.
- #17435, #17807 (open) — ongoing `createTool` / `schema-compat` type work.

### Verification

rodboev:
title: fix(schema-compat): publish json-schema type dependency

Fixes #17826

## Summary

`@mastra/schema-compat` publishes public declaration files that import `JSONSchema7` from `json-schema`, but the package only declares `@types/json-schema` as a development dependency. Consumers without that type package elsewhere in their tree can see `createTool` input inference collapse to `any`.

## Root cause

`PublicSchema` includes `JSONSchema7`, and that type comes from `@types/json-schema`, not from the runtime `json-schema` package itself. Because the declaration provider is dev-only, downstream installs can load the public `.d.ts` through an implicit-any import; that `any` member poisons the schema union and makes `InferPublicSchema` infer `any` for tool input schemas.

## Changes

- `packages/schema-compat/package.json`: move `@types/json-schema` into published dependencies (~2 lines)
- `pnpm-lock.yaml`: reflect the dependency-scope change
- `packages/schema-compat/src/schema.types.test.ts`: add a dependency-scope regression for the public `JSONSchema7` type import (~15 lines)
- `packages/core/src/tools/createtool-input-types.test-d.ts`: keep existing inference coverage unchanged and rerun it as a smoke check (~0 lines expected)

## What is NOT changed

`PublicSchema`, `InferPublicSchema`, and JSON Schema runtime conversion behavior are unchanged. This PR only fixes the package metadata that makes the existing public types safe for consumers.

## Out of scope

The issue also suggests auditing other generated declaration surfaces that mention `json-schema`. This PR stays on the confirmed `@mastra/schema-compat` package boundary that causes the `createTool` inference regression.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## Checklist

- [x] Linked related issue
- [x] Added or updated tests

## Test plan

- `pnpm --filter ./packages/schema-compat test src/schema.types.test.ts` — passed, `2/2` tests. Covers: the new package-metadata regression that asserts `@types/json-schema` is published in `dependencies` and absent from `devDependencies`, plus the package's existing colocated `schema.types` test file included by the Vitest run.
- `pnpm --filter ./packages/core test src/tools/createtool-input-types.test-d.ts` — failed before reaching a change-specific regression signal. Covers: existing `createTool` input inference smoke coverage when the package type-test suite is healthy. Current blocker: unrelated pre-existing `packages/core` type-test failures in workflow and workspace files, including missing `@internal/ai-sdk-v4`, missing `logger` properties, and unrelated `InferPublicSchema`/`workflow-schema-types.test-d.ts` errors.

## ELI5 (Explain Like I'm 5)

This PR ensures the package that provides schema types actually ships the JSON Schema type definitions it references so TypeScript won't treat those types as "any" for people who install the package.

## Problem Statement

Published declaration files in `@mastra/schema-compat` import JSONSchema7 from 'json-schema', but `@types/json-schema` was only a devDependency. Consumers without that types package could have JSONSchema7 resolve to any, collapsing type inference (e.g., createTool input types) to any.

## Changes Made

1. Moved `@types/json-schema` from devDependencies to dependencies in packages/schema-compat/package.json (now at "^7.0.15").
2. Updated pnpm-lock.yaml to reflect the dependency scope change (included in the PR).
3. Added a Vitest regression test at packages/schema-compat/src/schema.types.test.ts that asserts:
   - package.json lists `@types/json-schema` in dependencies with version '^7.0.15'
   - `@types/json-schema` is absent from devDependencies
4. Added a changeset patch (.changeset/kind-lobsters-drum.md) documenting the fix.
5. Re-ran an existing createtool input inference type-test in packages/core as a smoke check (noting unrelated pre-existing failures in packages/core remain).

## Impact

- No runtime or exported API changes (PublicSchema, InferPublicSchema, and JSON Schema runtime conversion unchanged).
- Packaging fix ensures consumers receive proper JSONSchema7 types, restoring TypeScript inference for createTool and preventing InferPublicSchema from collapsing to any.
- Tests: schema-compat type tests pass locally (2/2); new regression test included to prevent future regressions.

## Notes / Alternatives

- Alternatives considered: vendor a local JSONSchema7 type or audit other declaration surfaces referencing 'json-schema' for similar scoping errors; chosen fix is minimal and low-risk.

### DIFF
--- a/.changeset/kind-lobsters-drum.md
+++ b/.changeset/kind-lobsters-drum.md
@@ -1,0 +1,5 @@
+---
+'@mastra/schema-compat': patch
+---
+
+Fixed a packaging issue where `JSONSchema7` type imports could resolve to `any` for consumers without `@types/json-schema`, breaking type inference in `createTool` and other schema utilities.
--- a/packages/schema-compat/package.json
+++ b/packages/schema-compat/package.json
@@ -98,6 +98,7 @@
   "author": "",
   "license": "Apache-2.0",
   "dependencies": {
+    "@types/json-schema": "^7.0.15",
     "json-schema-to-zod": "^2.7.0",
     "zod-from-json-schema": "^0.5.2",
     "zod-from-json-schema-v3": "npm:zod-from-json-schema@^0.0.5",
@@ -118,7 +119,6 @@
     "@internal/test-utils": "workspace:*",
     "@internal/types-builder": "workspace:*",
     "@standard-schema/spec": "^1.1.0",
-    "@types/json-schema": "^7.0.15",
     "@types/node": "22.19.15",
     "@vitest/coverage-v8": "catalog:",
     "@vitest/ui": "catalog:",
--- a/packages/schema-compat/src/schema.types.test.ts
+++ b/packages/schema-compat/src/schema.types.test.ts
@@ -1,0 +1,19 @@
+import { readFileSync } from 'node:fs';
+
+import { describe, expect, it } from 'vitest';
+
+type PackageJson = {
+  dependencies?: Record<string, string>;
+  devDependencies?: Record<string, string>;
+};
+
+describe('schema-compat package metadata', () => {
+  it('publishes @types/json-schema for consumers', () => {
+    const packageJson = JSON.parse(
+      readFileSync(new URL('../package.json', import.meta.url), 'utf8')
+    ) as PackageJson;
+
+    expect(packageJson.dependencies?.['@types/json-schema']).toBe('^7.0.15');
+    expect(packageJson.devDependencies?.['@types/json-schema']).toBeUndefined();
+  });
+});
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -5016,6 +5016,9 @@ importers:
 
   packages/schema-compat:
     dependencies:
+      '@types/json-schema':
+        specifier: ^7.0.15
+        version: 7.0.15
       json-schema-to-zod:
         specifier: ^2.7.0
         version: 2.7.0
@@ -5062,9 +5065,6 @@ importers:
       '@standard-schema/spec':
         specifier: ^1.1.0
         version: 1.1.0
-      '@types/json-schema':
-        specifier: ^7.0.15
-        version: 7.0.15
       '@types/node':
         specifier: 22.19.15
         version: 22.19.15


TheIsrael1:
Linked Issue fixed by #17877

PR #17867, opened 2026-06-12
fix(core): add AGENT_RUN root span to DurableAgent.stream() for observability traces

### LINKED ISSUE #17863: [BUG] Durable agents (createDurableAgent) produce no observability traces — no AGENT_RUN root span is created
## Problem Statement

Agents wrapped with `createDurableAgent` (`DurableAgent`) emit **no observability traces at all**, while plain `Agent`s in the same Mastra instance trace correctly. Tool calls, processors, model calls, and workflows all show up in the trace UI/storage — but any run that goes through the durable execution path is completely absent.

This is not a storage/exporter misconfiguration: the same `Observability` config produces traces for every non-durable agent. The difference is solely the `createDurableAgent` wrapper.

## Root Cause

The durable execution path never opens an `AGENT_RUN` span, and the workflow it runs inside is fully marked internal — so the resulting span tree has no exportable root and is dropped.

**1. The normal path opens an `AGENT_RUN` root span.**
`Agent.stream()` calls (in `@mastra/core` dist, `chunk-*.js`):
```js
const agentSpan = getOrCreateSpan({ type: "agent_run" /* AGENT_RUN */, ... });
```
Every LLM/tool/processor span is a child of that non-internal root, so the whole trace exports.

**2. The durable path never does.**
`DurableAgent.stream()` → `prepareForDurableExecution()` → `executeWorkflow()` → `run.start()`. Grepping the entire `dist/agent/durable/index.js` shows **no** `getOrCreateSpan({ type: AGENT_RUN })` anywhere. The run executes inside the durable agentic-loop workflow.

**3. Both durable workflow levels are marked internal** (`dist/agent/durable/index.js`):
```js
// singleIterationWorkflow (AGENTIC_EXECUTION) and the outer AGENTIC_LOOP
options: {
  tracingPolicy: { internal: 1 /* WORKFLOW */ }, // "hide workflow spans"
}
```

**4. With no non-internal ancestor, the trace is dropped.** The exporter resolves a root via `getRootExportSpan`:
```js
function getRootExportSpan(span) {
  let rootExportSpan = span.isInternal ? void 0 : span;
  while (current?.parent) { current = current.parent; if (!current.isInternal) rootExportSpan = current; }
  return rootExportSpan; // → undefined when every ancestor is internal
}
```
The span tree for a durable run is `workflow_run (internal)` → `workflow_step (internal)` → model/tool spans. No `AGENT_RUN`, every ancestor internal → `getRootExportSpan` returns `undefined` → nothing is exported.

**5. No userland workaround.** `executeWorkflow` calls `run.start({ inputData, requestContext })` with **no** `tracingOptions`, so a caller-supplied `traceId`/`parentSpanId` passed to `durableAgent.stream()` is dropped before it reaches the workflow. There is no way to inject a non-internal parent from outside.

Secondary symptom in the same direction: the durable LLM step rebuilds a model span only from `inputData.modelSpanData` (`durable/index.js` ~L1564), which is solely seeded *after* a tool-calling iteration (`modelSpanData: hasToolCalls ? modelSpan?.exportSpan?.() : void 0`, ~L1828). It is never seeded on the first iteration, so even the model-span chain never starts.

## Expected Behavior

A durable agent run should produce the same `AGENT_RUN`-rooted trace as a non-durable agent run, with LLM/tool/processor spans nested under it. The internal durable workflow spans can stay hidden, but they must hang off an exportable `AGENT_RUN` root.

## Reproduction

```ts
import { createDurableAgent } from "@mastra/core/agent/durable";
import { Observability, MastraStorageExporter } from "@mastra/observability";
import { anthropic } from "@ai-sdk/anthropic";

const base = new Agent({ name: "demo", model: anthropic("claude-haiku-4-5"), instructions: "..." });
const durable = createDurableAgent({ agent: base });

const mastra = new Mastra({
  agents: { plain: base, durable },
  storage: /* any storage backend */,
    configs: { default: { serviceName: "repro", exporters: [new MastraStorageExporter()] } },
  }),
});

// Run both:
await mastra.getAgent("plain").stream("hello");    // → trace appears
await mastra.getAgent("durable").stream("hello");  // → NO trace appears
```

## Environment

- `@mastra/core`: 1.41.0 (also reproduced by source inspection in `1.42.0-alpha.3`, the `tracing-hotfix` build `0.0.0-tracing-hotfix-20260319165043`, and the latest main snapshot `0.0.0-channels-on-lease-20260611222452` — all four still lack any `agent_run` span in `dist/agent/durable/index.js`)
- `@mastra/observability`: (current)
- Storage: Postgres + ClickHouse observability domain (also reproduces with any storage exporter)
- Node: 26

## Component

- Agents
- Observability (AI Telemetry)

## Verification

- [x] I have provided sufficient context for the team to understand the request.

Aayush-engineer:
## Summary
Fixes #17863

DurableAgent.stream() never created an AGENT_RUN root span, so all
durable workflow spans had no non-internal ancestor and were dropped
by getRootExportSpan(). Plain Agent.stream() correctly creates this span.

## Changes
- Added getOrCreateSpan(AGENT_RUN) in DurableAgent.stream() matching the non-durable path
- Stored agentSpan in globalRunRegistry so executeWorkflow() can access it
- Pass agentSpan.getTracingContext() into run.start() so LLM/tool/processor
  child spans are correctly parented under the AGENT_RUN root
- End/error the span in onFinish/onError callbacks
- Added tracingOptions and tracingContext to DurableAgentStreamOptions
- Added agentSpan field to RunRegistryEntry

## Trace chain
stream() → agentSpan (AGENT_RUN, non-internal) → run.start(tracingContext)
→ workflow steps → llm-execution params.tracingContext → child spans

## ELI5

Before this, durable agents didn’t create the “top” trace span that monitoring tools look for, so their runs were often invisible. This PR makes durable agents create the same `AGENT_RUN` root span as regular agents and then hangs the rest of the model/tool spans under it so exporters can export complete traces.

## Overview

Updates the durable execution path to create an exportable, non-internal `AGENT_RUN` root span and to consistently propagate tracing context through `run.start()`, model execution, and tool calls—matching the trace shape of non-durable `Agent.stream()`.

## Changes

- **`packages/core/src/agent/durable/durable-agent.ts`**
  - **`DurableAgent.stream()`** now creates an **`AGENT_RUN` root span** via `getOrCreateSpan({ type: SpanType.AGENT_RUN, tracingContext: options?.tracingContext })`.
  - Stores the created **`agentSpan`** in `globalRunRegistry` (alongside `messageList`) so workflow steps can parent to it.
  - Ensures the span is **ended/errored** in the `onFinish`/`onError` lifecycle callbacks.
  - **`executeWorkflow()`** reads `registryEntry.agentSpan` and passes it as `tracingContext` into `run.start(...)` so durable workflow spans attach under the `AGENT_RUN` root.

- **`packages/core/src/agent/durable/types.ts`**
  - `RunRegistryEntry` adds `agentSpan?: Span<SpanType.AGENT_RUN>` for trace propagation into workflow steps.
  - `DurableToolCallInput` adds `stepSpanData?: unknown` so tool calls can be parented under the correct model step span.

- **`packages/core/src/agent/durable/workflows/steps/llm-execution.ts`**
  - Improves span creation/closing across iterations (e.g., creating the first `MODEL_GENERATION` child span when missing, and ending `MODEL_GENERATION` appropriately when tool calls exist so the next iteration can start cleanly).
  - Ensures streaming finish handling results in correct `MODEL_STEP`/`MODEL_INFERENCE` closure.
  - Produces per-tool-call output that includes optional `stepSpanData` for correct tool-call parenting.

- **`packages/core/src/agent/durable/workflows/steps/tool-call.ts`**
  - When `stepSpanData` is present, reconstructs the parent span (via observability span rebuilding) and passes the derived `tracingContext` into `toolOptions`, so `TOOL_CALL` spans nest under the intended `model_step` span rather than incorrectly rooting elsewhere.

- **`.changeset/giant-cars-smoke.md`**
  - Documents that durable agents now emit complete observability traces consistent with the non-durable path.

## Observability Outcome

Durable trace structure becomes:

`stream()` → **`AGENT_RUN` (root, non-internal)** → `run.start(tracingContext)` → workflow/model/tool spans (with tool calls correctly nested under the relevant `model_step`), ensuring exporters can find/export the trace root.

## Impact

- Durable agent runs now produce **exportable, `AGENT_RUN`-rooted traces** like non-durable agents.
- Caller-provided tracing context is preserved through the durable execution pipeline.
- Model/tool step span hierarchy and closing behavior are corrected so traces are complete across tool-calling iterations.

### DIFF
--- a/packages/core/src/agent/durable/durable-agent.ts
+++ b/packages/core/src/agent/durable/durable-agent.ts
@@ -392,7 +392,9 @@ export class DurableAgent<
     const workflow = this.getWorkflow();
     const registryEntry = globalRunRegistry.get(runId);
     const requestContext = registryEntry?.requestContext;
-    const spanTracingContext = (registryEntry?.agentSpan as any)?.getTracingContext?.();
+    const spanTracingContext = registryEntry?.agentSpan
+      ? { currentSpan: registryEntry.agentSpan }
+      : undefined;
     const run = await workflow.createRun({ runId, pubsub: this.pubsub });
     const result = await run.start({ inputData: workflowInput, requestContext, tracingContext: spanTracingContext });
 
--- a/packages/core/src/agent/durable/durable-agent.ts
+++ b/packages/core/src/agent/durable/durable-agent.ts
@@ -12,6 +12,7 @@ import type { MessageListInput } from '../message-list';
 import type { ToolsInput } from '../types';
 
 import { AGENT_STREAM_TOPIC } from './constants';
+import { EntityType, SpanType, getOrCreateSpan } from '../../observability';
 import { runDurableStreamUntilIdle } from './durable-stream-until-idle';
 import { prepareForDurableExecution } from './preparation';
 import { ExtendedRunRegistry, globalRunRegistry } from './run-registry';
@@ -64,6 +65,10 @@ export interface DurableAgentStreamOptions<OUTPUT = undefined> {
   structuredOutput?: AgentExecutionOptions<OUTPUT>['structuredOutput'];
   /** Version overrides for sub-agent delegation */
   versions?: AgentExecutionOptions<OUTPUT>['versions'];
+  /** Tracing options for observability */
+  tracingOptions?: AgentExecutionOptions<OUTPUT>['tracingOptions'];
+  /** Tracing context for span propagation */
+  tracingContext?: AgentExecutionOptions<OUTPUT>['tracingContext'];
   /** Callback when chunk is received */
   onChunk?: (chunk: ChunkType<OUTPUT>) => void | Promise<void>;
   /** Callback when step finishes */
@@ -385,10 +390,11 @@ export class DurableAgent<
    */
   protected async executeWorkflow(runId: string, workflowInput: DurableAgenticWorkflowInput): Promise<void> {
     const workflow = this.getWorkflow();
-    const requestContext = globalRunRegistry.get(runId)?.requestContext;
-
+    const registryEntry = globalRunRegistry.get(runId);
+    const requestContext = registryEntry?.requestContext;
+    const spanTracingContext = (registryEntry?.agentSpan as any)?.getTracingContext?.();
     const run = await workflow.createRun({ runId, pubsub: this.pubsub });
-    const result = await run.start({ inputData: workflowInput, requestContext });
+    const result = await run.start({ inputData: workflowInput, requestContext, tracingContext: spanTracingContext });
 
     if (result?.status === 'failed') {
       const error = new Error((result as any).error?.message || 'Workflow execution failed');
@@ -434,7 +440,22 @@ export class DurableAgent<
     messages: MessageListInput,
     options?: DurableAgentStreamOptions<TOutput>,
   ): Promise<DurableAgentStreamResult<TOutput>> {
-    // 1. Prepare for durable execution (non-durable phase)
+    // 1. Create AGENT_RUN root span so durable traces have an exportable root,
+    //    matching the non-durable Agent.stream() path.
+    const agentSpan = getOrCreateSpan({
+      type: SpanType.AGENT_RUN,
+      name: `agent run: '${this.#wrappedAgent.id}'`,
+      entityType: EntityType.AGENT,
+      entityId: this.#wrappedAgent.id,
+      entityName: this.#wrappedAgent.name,
+      input: messages,
+      tracingOptions: options?.tracingOptions,
+      tracingContext: options?.tracingContext,
+      requestContext: options?.requestContext,
+      mastra: this.#mastra,
+    });
+
+    // 2. Prepare for durable execution (non-durable phase)
     const preparation = await prepareForDurableExecution<TOutput>({
       agent: this.#wrappedAgent as Agent<string, any, TOutput>,
       messages,
@@ -448,7 +469,7 @@ export class DurableAgent<
 
     // 2. Register non-serializable state (both local and global registries)
     this.#runRegistry.registerWithMessageList(runId, registryEntry, messageList, { threadId, resourceId });
-    globalRunRegistry.set(runId, { ...registryEntry, messageList });
+    globalRunRegistry.set(runId, { ...registryEntry, messageList, agentSpan });
 
     // Track cleanup state to avoid double cleanup
     let cleanedUp = false;
@@ -487,10 +508,13 @@ export class DurableAgent<
       onStepFinish: options?.onStepFinish,
       onFinish: async result => {
         await options?.onFinish?.(result);
+        agentSpan?.end?.({ output: result.output });
         scheduleAutoCleanup();
       },
       onError: async error => {
         await options?.onError?.(error);
+        agentSpan?.error?.({ error });
+        agentSpan?.end?.({});
         scheduleAutoCleanup();
       },
       onSuspended: options?.onSuspended,
--- a/packages/core/src/agent/durable/durable-agent.ts
+++ b/packages/core/src/agent/durable/durable-agent.ts
@@ -595,6 +595,19 @@ export class DurableAgent<
     };
 
     const globalEntry = globalRunRegistry.get(runId);
+    // Create a new AGENT_RUN span for the resumed run.
+    // The original span is gone (globalRunRegistry is in-memory and doesn't
+    // survive suspensions), so we open a fresh root span marked as a resume.
+    const resumeAgentSpan = getOrCreateSpan({
+      type: SpanType.AGENT_RUN,
+      name: `agent run: '${this.#wrappedAgent.id}' (resumed)`,
+      entityType: EntityType.AGENT,
+      entityId: this.#wrappedAgent.id,
+      entityName: this.#wrappedAgent.name,
+      metadata: { runId, resumed: true },
+      requestContext: globalEntry?.requestContext,
+      mastra: this.#mastra,
+    });
     const resumeModel = globalEntry?.model as any;
 
     const {
@@ -616,10 +629,13 @@ export class DurableAgent<
       onStepFinish: options?.onStepFinish,
       onFinish: async result => {
         await options?.onFinish?.(result);
+        resumeAgentSpan?.end?.({ output: result.output });
         scheduleAutoCleanup();
       },
       onError: async error => {
         await options?.onError?.(error);
+        resumeAgentSpan?.error?.({ error });
+        resumeAgentSpan?.end?.({});
         scheduleAutoCleanup();
       },
       onSuspended: options?.onSuspended,
@@ -631,7 +647,10 @@ export class DurableAgent<
     ready
       .then(async () => {
         const run = await workflow.createRun({ runId, pubsub: this.pubsub });
-        const result = await run.resume({ resumeData, requestContext });
+        const spanTracingContext = resumeAgentSpan
+          ? { currentSpan: resumeAgentSpan }
+          : undefined;
+        const result = await run.resume({ resumeData, requestContext, tracingContext: spanTracingContext });
         if (result?.status === 'failed') {
           const error = new Error((result as any).error?.message || 'Workflow resume failed');
           void this.emitError(runId, error);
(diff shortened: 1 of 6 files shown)


packages/core/src/agent/durable/workflows/steps/tool-call.ts
```diff
@@ -662,6 +680,8 @@ export function createDurableToolCallStep() {
           result,
         };
       } catch (error) {
+        toolSpan?.error({ error: error instanceof Error ? error : new Error(String(error)) });
+        toolSpan?.end({});
```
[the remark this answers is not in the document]
Aayush-engineer: Done, removed redundant toolSpan?.end({}) since error() ends the span by default.

packages/core/src/agent/durable/workflows/steps/tool-call.ts
```diff
@@ -640,8 +643,23 @@ export function createDurableToolCallStep() {
         }
       }
 
+      // Rebuild model_step span so the tool call spans nest under it
+      const observability = (mastra as any)?.observability?.getSelectedInstance({ requestContext });
+      const stepSpanData = (typedInput as any).stepSpanData as ExportedSpan<SpanType.MODEL_STEP> | undefined;
+      const stepSpan = stepSpanData && observability
+        ? observability.rebuildSpan(stepSpanData)
+        : undefined;
+      const toolSpan = stepSpan?.createChildSpan({
```
intojhanurag: This creates a second TOOL_CALL span. The tool builder https://github.com/mastra-ai/mastra/blob/76926ad154ab7bd9fb14b567cbae8f0dbf9bff63/packages/core/src/tools/tool-builder/builder.ts#L743 already makes one per execution from toolOptions.tracingContext,  but toolOptions
https://github.com/mastra-ai/mastra/blob/76926ad154ab7bd9fb14b567cbae8f0dbf9bff63/packages/core/src/agent/durable/workflows/steps/tool-call.ts#L310 has no tracingContext, so its span ends up orphaned at the root. Testing a tool call against a real model, I get 2 tool spans (one nested here, one floating) vs 1
for a plain agent.

Aayush-engineer:
Good catch! Removed the duplicate span now passing tracingContext derived from stepSpanData into toolOptions so the tool builder's existing TOOL_CALL span parents correctly under model_step.

intojhanurag:
Hey @Aayush-engineer , did you trace suspend/resume , which is the defining feature of durable agent ?

Aayush-engineer:
Good point  haven't traced suspend/resume yet. The agentSpan lives in globalRunRegistry (in-memory) so it won't survive a suspension. On resume the root span would be missing. Looking into how to handle this properly.

Aayush-engineer:
Added a fresh AGENT_RUN span in resume() with resumed: true metadata and propagated its tracingContext into run.resume(). Since globalRunRegistry is in-memory and doesn't survive suspensions, creating a new root span on resume is the right approach.

packages/core/src/agent/durable/workflows/steps/tool-call.ts
```diff
@@ -304,13 +307,22 @@ export function createDurableToolCallStep() {
...
       }
 
+      // Build tracingContext from stepSpanData so builder's TOOL_CALL span
+      // nests under model_step instead of floating at AGENT_RUN root.
+      const observability = (mastra as any)?.observability?.getSelectedInstance({ requestContext });
+      const stepSpanData = (typedInput as any).stepSpanData as ExportedSpan<SpanType.MODEL_STEP> | undefined;
+      const stepSpan = stepSpanData && observability
+        ? observability.rebuildSpan(stepSpanData)
+        : undefined;
+      const toolTracingContext = stepSpan ? { currentSpan: stepSpan } : undefined;
```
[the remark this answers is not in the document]
Aayush-engineer: Done, removed the unnecessary as any cast.

Aayush-engineer:
@intojhanurag all concerns addressed including suspend/resume — ready for re-review when you get a chance!

packages/core/src/agent/durable/workflows/steps/llm-execution.ts
```diff
@@ -383,8 +394,25 @@ export function createDurableLLMExecutionStep(_options?: DurableLLMExecutionStep
...
             // Wrap the base stream with ModelSpanTracker to create MODEL_STEP and MODEL_CHUNK spans
             const baseStream = outputStream._getBaseStream();
-            const trackedStream = modelSpanTracker?.wrapStream(baseStream) ?? baseStream;
-
+            // Convert finish → step-finish before wrapStream so the ModelSpanTracker
+            // sees step-finish and correctly closes MODEL_STEP/MODEL_INFERENCE spans.
+            // The durable LLM stream emits finish but never step-finish; wrapStream
+            // only closes spans on step-finish, so without this transform the spans
+            // are never ended and never exported.
+            const stepFinishStream = modelSpanTracker
```
intojhanurag, line 402: this finish→step-finish is now gated on `modelSpanTracker`, but it used to emit to pubsub unconditionally. with no observability there's no tracker, so the client never gets `step-finish` and `getFullOutput().text`/`steps` come back empty. tested with real anthropic: base returns the text + 1 step, this branch returns "" + 0 steps. can we keep the pubsub emit unconditional?

packages/core/src/agent/durable/workflows/steps/llm-execution.ts
```diff
@@ -596,6 +617,18 @@ export function createDurableLLMExecutionStep(_options?: DurableLLMExecutionStep
                   });
                 }
               }
+              // End model_generation span — no tool calls means this is the final iteration
+              modelSpan?.end({
+                output: { text: textDeltas.join('') },
+                attributes: { usage },
+              });
+            } else {
```
intojhanurag, line 625: the `model_step` span never gets ended on the tool-calling iteration here (only model_generation does). so tool_call + model_inference orphan to the root instead of nesting under model_step, opposite of what the PR says. confirmed on a real tool run: step:0 stays incomplete and tool_call.parentSpanId points at it. need to end the model_step in this branch too.

packages/core/src/agent/durable/durable-agent.ts
```diff
@@ -487,10 +510,13 @@ export class DurableAgent<
       onStepFinish: options?.onStepFinish,
       onFinish: async result => {
         await options?.onFinish?.(result);
+        agentSpan?.end?.({ output: result.output });
         scheduleAutoCleanup();
       },
       onError: async error => {
         await options?.onError?.(error);
+        agentSpan?.error?.({ error });
+        agentSpan?.end?.({});
```
intojhanurag, line 519: `error()` already ends the span by default, so this `end({})` double-ends it (2 SPAN_ENDED for the same agent_run on a real errored run). same thing you fixed for the tool span in a4935cb. drop the extra `.end({})`. same at line 638 in resume().

packages/core/src/agent/durable/durable-agent.ts
```diff
@@ -487,10 +510,13 @@ export class DurableAgent<
...
         scheduleAutoCleanup();
       },
       onSuspended: options?.onSuspended,
```
intojhanurag, line 522: on suspend we never end `agentSpan`, so a suspended run has no exportable root again. the non-durable path ends it on suspend (map-results-step.ts:289).

packages/core/src/agent/durable/durable-agent.ts
```diff
@@ -569,6 +595,19 @@ export class DurableAgent<
     };
 
     const globalEntry = globalRunRegistry.get(runId);
+    // Create a new AGENT_RUN span for the resumed run.
+    // The original span is gone (globalRunRegistry is in-memory and doesn't
+    // survive suspensions), so we open a fresh root span marked as a resume.
+    const resumeAgentSpan = getOrCreateSpan({
```
intojhanurag, line 601: this resume span starts a brand new disconnected trace. the framework already persists tracing context in the snapshot and relinks on resume (workflow.ts:4054, agent.ts:6269) , passing a new currentSpan into run.resume actually overrides that. suspend/resume isn't in #17863's scope, might be cleaner to pull this out into its own PR.

intojhanurag (commented):
simple-stream fix looks great, durable matches the plain agent now and persists fine. couple blockers: getFullOutput().text/steps break when observability is off (default case, regression), and tool calls orphan because the tool-iteration model_step never closes. also the error path double-ends agent_run. suspend/resume feels out of scope for this issue. details inline.

intojhanurag:
> @intojhanurag all concerns addressed including suspend/resume — ready for re-review when you get a chance!

if you are testing in observability , you can attach ss for Before and After.

intojhanurag:
Thanks for the contribution @Aayush-engineer ! we will handle this by own .

PR #17873, opened 2026-06-12
feat(core): export Middleware type from server entrypoint

Sandy-1711:
## Description

The `Middleware` type was defined in `server/types` but never re-exported from the `@mastra/core/server` entrypoint, so you couldn't import it:

```ts
import type { Middleware } from '@mastra/core/server'
```

This made it awkward to declare middleware in its own file. Inline middleware inside `new Mastra({ ... })` gets its types from the config, but once you move it to a separate module there's no exported type to annotate it with, so the handler params fall back to `any` in strict projects.

This PR re-exports `Middleware` from the server entrypoint so you can do this:

```ts
import type { Middleware } from '@mastra/core/server'

export const authMiddleware: Middleware = {
  path: '/*',
  handler: async (c, next) => {
    const authHeader = c.req.header('authorization')
    if (!authHeader) {
      return new Response('Unauthorized', { status: 401 })
    }
    await next()
  },
}
```

No runtime changes, this is purely the type export. I also added a type test and updated the middleware docs to show the import.

While looking into this I noticed `ServerConfig` and `Methods` have the same gap (and `ApiRoute`, which is already exported, references `Methods`). I kept this PR scoped to `Middleware` since that's what the issue asks for, but I'm happy to open a separate issue/PR for those if the team wants them exported too.

## Related issue(s)

Closes #17868

## Type of change

- [x] New feature (non-breaking change that adds functionality)
- [x] Documentation update

## Checklist

- [x] I have linked the related issue(s) in the description above
- [x] I have made corresponding changes to the documentation (if applicable)
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] I have addressed all Coderabbit comments on this PR

## ELI5

This PR makes it easier to write middleware files in TypeScript. Instead of losing type information when you create middleware in a separate file, you can now import a `Middleware` type from the Mastra library to make sure your middleware has the right format. It's like getting a template that tells you exactly what shape your code needs to be.

### Type Exports
- **Re-exported `Middleware` type** from `@mastra/core/server` entrypoint (in `packages/core/src/server/index.ts`), enabling imports like `import type { Middleware } from '`@mastra/core/server`'`

### Documentation
- **Updated middleware documentation** (`docs/src/content/en/docs/server/middleware.mdx`) with a practical example showing how to declare middleware in a separate module using the newly exported `Middleware` type
- Example demonstrates path-scoped middleware with proper TypeScript typing for the handler function

### Testing
[...]

Issue #17868, opened 2026-06-12
[FEATURE] export Middleware type from server entrypoint


dszdsxc19:
#### Description

  Mastra already supports global server middleware through `server.middleware`, and the docs describe the option as `Middleware | Middleware[]`. However, when middleware is declared in a separate module, users currently cannot import the `Middleware` type from the public
  server entrypoint:

  ```ts
  import type { Middleware } from '@mastra/core/server'
```

  The type exists internally in server/types, but it is not re-exported by server/index, so users either need to inline middleware inside the new Mastra(...) config or derive the type indirectly from Config['server']['middleware'].

  This makes separately declared middleware awkward in strict TypeScript projects.

<img width="1624" height="370" alt="Image" src="https://github.com/user-attachments/assets/d5b0a755-c334-4ce5-8427-b30baf0670dc" />

  #### Example

this does not work:

```ts
  import type { Middleware } from '@mastra/core/server'

  export const authMiddleware: Middleware = {
    path: '/*',
    handler: async (c, next) => {
      const authHeader = c.req.header('authorization')
      if (!authHeader) {
        return new Response('Unauthorized', { status: 401 })
      }

      await next()
    },
  }
```

#### Changes

  - Re-export Middleware from @mastra/core/server
  - Keeps runtime behavior unchanged
  - Improves type ergonomics for middleware declared outside the Mastra config object

  #### Why

  Inline middleware receives contextual typing from the Mastra config:

```ts
  new Mastra({
    server: {
      middleware: [
        {
          path: '/*',
          handler: async (c, next) => {
            await next()
          },
        },
      ],
    },
  })
```

  But when middleware is declared in another file, TypeScript cannot infer the expected handler parameter types from the later import site. A public Middleware type gives users the intended way to annotate reusable middleware.

### Component

Server

### Example Use Case

```ts
import type { Middleware } from '@mastra/core/server'

  export const authMiddleware: Middleware = {
    path: '/*',
    handler: async (c, next) => {
      const authHeader = c.req.header('authorization')
      if (!authHeader) {
        return new Response('Unauthorized', { status: 401 })
      }

      await next()
    },
  }
```

### Verification

- [x] I have searched the existing issues to make sure this is not a duplicate
- [x] I have provided sufficient context for the team to understand the request

daneatmastra:
Thank you for reporting this issue! We have labeled it for the team and we will look into it as soon as possible.

[...]

PR #17875, opened 2026-06-12
feat(core): export ServerConfig and Methods types from server entrypoint

### LINKED ISSUE #17874: [FEATURE] export ServerConfig and Methods types from server entrypoint
### Problem Statement

Following #17868 / #17873 (which add the `Middleware` re-export), two related types defined in `server/types` are still not re-exported from `@mastra/core/server`:

- `ServerConfig` — the shape of the `server` config object. Users who want to build their server config in a separate module can't annotate it.
- `Methods` — the HTTP method union (`'GET' | 'POST' | 'PUT' | 'DELETE' | 'PATCH' | 'ALL'`). `ApiRoute` is already exported and references `Methods`, so today you can consume `ApiRoute` but can't name the type of its `method` field.

Neither of these imports works today:

```ts
import type { ServerConfig, Methods } from '@mastra/core/server'
```

### Proposed Solution

Re-export `ServerConfig` and `Methods` from `@mastra/core/server`, the same way #17873 does for `Middleware`. Runtime behavior is unchanged — this only improves type ergonomics for server config and routes declared outside the `new Mastra(...)` constructor.

### Component

Server

### Alternatives Considered

Deriving the types indirectly, e.g. `Config['server']` or `NonNullable<ServerConfig['apiRoutes']>[number]['method']`, but that's awkward and not discoverable.

### Example Use Case

As a developer, I want to declare my server config and route methods in a separate file with proper types so that I can keep my Mastra setup modular.

```ts
import type { ServerConfig, Methods } from '@mastra/core/server'

const protectedMethods: Methods[] = ['POST', 'PUT', 'DELETE']

export const serverConfig: ServerConfig = {
  port: 4111,
  middleware: async (c, next) => {
    await next()
  },
}
```

### Additional Context

Implementation is a one-line change in the same export block as #17873 (packages/core/src/server/index.ts).

### Verification

Sandy-1711:
## Description

Follow-up to #17873 (which re-exports `Middleware`, closing #17868). This one does the same for two related types that have the same gap:

- `ServerConfig` — the shape of the `server` config object, so you can build it in a separate file and annotate it.
- `Methods` — the HTTP method union. `ApiRoute` is already exported and references `Methods`, so right now you can use `ApiRoute` but can't name the type of its `method` field.

Neither of these worked before:

```ts
import type { ServerConfig, Methods } from '@mastra/core/server'
```

Now you can do this:

```ts
import type { ServerConfig, Methods } from '@mastra/core/server'

const protectedMethods: Methods[] = ['POST', 'PUT', 'DELETE']

export const serverConfig: ServerConfig = {
  port: 4111,
  middleware: async (c, next) => {
    await next()
  },
}
```

No runtime changes, this is purely the type exports. I also added a type test for both.

## Related issue(s)

Closes #17874

## Type of change

## Checklist

### DIFF
--- a/.changeset/tidy-server-types.md
+++ b/.changeset/tidy-server-types.md
@@ -1,0 +1,18 @@
+---
+'@mastra/core': patch
+---
+
+Added the `ServerConfig` and `Methods` type exports to `@mastra/core/server`. Server config and HTTP methods can now be declared in a separate file with full type safety. This complements the `Middleware` export and lets you name the type of `ApiRoute['method']`, which was previously unreachable.
+
+```ts
+import type { ServerConfig, Methods } from '@mastra/core/server';
+
+const protectedMethods: Methods[] = ['POST', 'PUT', 'DELETE'];
+
+export const serverConfig: ServerConfig = {
+  port: 4111,
+  middleware: async (c, next) => {
+    await next();
+  },
+};
+```
--- a/packages/core/src/server/index.ts
+++ b/packages/core/src/server/index.ts
@@ -12,6 +12,8 @@ export type {
   ContextWithMastra,
   CorsOptions,
   ApiRoute,
+  Methods,
+  ServerConfig,
   HttpLoggingConfig,
   ValidationErrorContext,
   ValidationErrorResponse,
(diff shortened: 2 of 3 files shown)


roaminro:
closing in favor of https://github.com/mastra-ai/mastra/pull/17882

PR #17877, opened 2026-06-12
fix(types-builder): bundle @types fallbacks for declarations

Aayush-engineer:
This fixes generated type declarations so packages can bundle matching `@types/*` declarations when the runtime package root cannot be resolved.

For `@mastra/schema-compat`, this means generated declarations no longer leak bare `json-schema` type imports. They now point at the bundled `@types/json-schema` declarations under `dist/_types`, so consumers do not need a separate `json-schema` type dependency for those public declarations.

Verified with:

```bash
pnpm --filter @mastra/schema-compat build
rg "from ['\"]json-schema['\"]|import\(['\"]json-schema['\"]\)" packages/schema-compat/dist --glob "*.d.ts"
pnpm --filter @mastra/schema-compat test
```

The declaration grep returns no bare `json-schema` imports after build.

## ELI5
When TypeScript generates `.d.ts` files, it needs to point to the right type packages for everything your code references. This PR updates the type-declaration builder so packages like `@mastra/schema-compat` can bundle the matching `@types/*` fallbacks (notably JSON Schema) themselves, instead of forcing consumers to install separate `json-schema` types.

### Files Modified

**`.changeset/fixed-schema-compat-json-schema-types.md`**
- Added a patch Changesets entry for `@mastra/schema-compat`, documenting the JSON Schema type declaration bundling fix.

**`packages/_types-builder/src/replace-types.js`**
- Refactored `replaceBundledReferences()` to scope `pkgJson` correctly and compute `typesFiles` only in the appropriate branch.
- Ensures the `@types/*` fallback path works cleanly when the runtime package root can’t be resolved, so generated declarations for `@mastra/schema-compat` reference bundled `@types/json-schema` under `dist/_types` rather than leaking bare `json-schema` imports.

**`client-sdks/client-js/src/utils/zod-to-json-schema.ts`**
- Added an explicit return type: `JSONSchema7 | T`.
- Added `import type { JSONSchema7 } from '`@mastra/schema-compat`'` so the function’s public typings use the bundled JSON Schema type.

**`client-sdks/client-js/src/utils/zod-to-json-schema.test.ts`**
- Added Vitest coverage for `zodToJsonSchema`, including:
  - Zod object conversion (with correct `required`)
  - Pass-through for non-Zod inputs
  - Primitive passthrough
  - Nested schema conversion (format + arrays)
  - Optional-field handling (only non-optional keys in `required`)

## Verification
- Built `@mastra/schema-compat`.
- Checked generated `.d.ts` files to confirm no bare `json-schema` imports remain.
- Ran `@mastra/schema-compat` tests.


Aayush-engineer:
I see PR #17856 was submitted earlier with the same fix. Happy to close this as a duplicate — submitting independently confirmed the root cause diagnosis.

abhiaiyer91 (approved):

Aayush-engineer:
@abhiaiyer91 thanks for the approval! Could you approve the CI workflows so the checks can run? Happy to close in favor of #17856 if that's simpler, just want to make sure the fix lands.

wardpeet:
I made some changes to actually fix the issue

wardpeet (approved):

PR #17887, opened 2026-06-12
feat(studio): configurable chat attachment file types

emilcardell-lg:
### What

Makes the Studio chat composer's accepted attachment file types configurable via a new `MASTRA_STUDIO_ATTACHMENT_TYPES` env var — a comma-separated MIME allowlist with `type/*` wildcards:

```bash
MASTRA_STUDIO_ATTACHMENT_TYPES="image/*,application/pdf,text/csv,application/vnd.openxmlformats-officedocument.spreadsheetml.sheet" mastra dev
```

When unset, behavior is unchanged (all file types accepted). When set:

- the file picker gets an `accept` attribute, so the native dialog filters to allowed types;
- files/URLs with disallowed content types are rejected **with visible feedback** in the attach dialog (previously a disallowed pick disappeared silently);
- the allowlist is shown in the rejection message so users know what's accepted.

Two supporting changes:

- **Binary non-PDF attachments (spreadsheets, docs, …) are now sent as `file` parts with their own MIME type** (new `file` kind) instead of falling through to the read-as-text path, which produced mojibake for binary formats. Textual application types (`application/json`, XML, YAML, NDJSON) still go through the text path.
- The `EXTENSION_TO_MIME` map in `@mastra/playground-ui` gains common document/spreadsheet/text types (csv, xlsx, xls, docx, doc, pptx, ppt, txt, md, json, xml) so URL attachments to those resolve a sensible content type.

### Why

Teams use Studio agents on workflows that revolve around documents the current composer can't meaningfully accept — e.g. security questionnaires arriving as `.xlsx`/`.csv`. Today the accepted set is hardcoded and a disallowed file silently fails to attach, which reads as a bug (this PR started from exactly that confusion). A configurable allowlist lets deployments decide what their models/tooling can handle, without changing the default.

### How

Follows the existing studio config pattern (`MASTRA_EXPERIMENTAL_FEATURES`, `MASTRA_AGENT_SIGNALS`, …): env var → `%%MASTRA_STUDIO_ATTACHMENT_TYPES%%` template placeholder in `index.html` → `window.MASTRA_STUDIO_ATTACHMENT_TYPES` → read by a small new `accepted-types.ts` module in the playground.

- `packages/playground/src/lib/ai-ui/attachments/accepted-types.ts` (new): parse the allowlist (`null` = unconfigured = accept all), match content types (exact / `type/*` / `*`), build the `accept` attribute value.
- `composer-attachments.tsx`: `addFiles` filters against the allowlist and returns rejected file names; `addUrl` resolves `false` for disallowed content types; new `file` attachment kind for non-PDF binary types, reusing the PDF data-URI file-part path with the attachment's own MIME type.
- `attach-file-dialog.tsx`: sets `input.accept`, shows a rejection message listing the rejected names and the allowed types.
- `packages/cli/src/commands/studio/studio.ts` + `packages/playground/index.html` + `packages/playground/vite.config.ts`: placeholder plumbing for built and standalone-dev studio.
- `packages/playground-ui/src/lib/file/constants.ts`: extension→MIME additions.

### Tests

- New `__tests__/accepted-types.test.ts`: parsing (unset, unreplaced placeholder, whitespace), matching (exact, wildcard, parameters/casing, `*/*`), accept-attribute building.
- Extended `__tests__/composer-attachments.test.tsx`: unconfigured = accept all; filtering returns rejected names and keeps allowed files; configured extra types accepted with `file` kind; URL rejection via mocked HEAD content type; xlsx → file part with correct data-URI MIME.
- `pnpm vitest run src/lib/ai-ui/attachments` → 20/20 pass.

### Backward compatibility

- Unset env var → no change for files that previously worked (images, PDFs, text).
- Behavior change (arguably a fix): unknown **binary** types previously became garbled text content; they are now proper file parts. Models that don't support that MIME type will error explicitly rather than receive mojibake.
- `addFiles`/`addUrl` signatures gain return values (`void` → `string[]`, `Promise<void>` → `Promise<boolean>`); existing callers that ignore the return value are unaffected (the attach dialog is the only caller in-repo).

### Future work (out of scope)

- A `studio.attachments.accept` field on the Mastra instance's `StudioConfig` (core types) flowing through the dev server, for config-file rather than env-var control.
- A spreadsheet adapter converting csv/xlsx client-side into markdown-table text parts for models without native spreadsheet/file support.

## ELI5

Admins can now restrict which file types users may attach in the Studio chat by setting a simple environment variable; disallowed files are clearly shown to the user instead of disappearing. Non-text binary files (e.g., spreadsheets, docs) are now attached correctly with their real file type.

---

## Overview

Adds a configurable allowlist for attachment MIME types via MASTRA_STUDIO_ATTACHMENT_TYPES (comma-separated MIME types; supports type/* wildcards). When set the file input’s accept attribute is applied, disallowed local files and URLs are rejected with visible feedback in the attach dialog; when unset, prior behavior (accept everything) remains.

---

## Key Changes

- Environment & HTML plumbing
  - CLI server and Vite standalone plugin inject a sanitized MASTRA_STUDIO_ATTACHMENT_TYPES into the frontend HTML (window.MASTRA_STUDIO_ATTACHMENT_TYPES). Sanitization restricts characters to MIME-legal set to prevent XSS at injection sites.

- Attachment allowlist utilities
  - New packages/playground/src/lib/ai-ui/attachments/accepted-types.ts:
    - getAcceptedAttachmentTypes(): parses window flag, returns string[] or null (null = accept all).
    - isAcceptedAttachmentType(contentType, accepted): exact + type/* wildcard matching, case- and parameter-insensitive, treats */* as permissive.
    - acceptAttributeValue(accepted): builds file input accept attribute or undefined when unrestricted.

- UI behavior
  - attach-file-dialog.tsx: sets input.accept, tracks rejectedNames, clears stale rejection feedback when reopened, displays an alert listing rejected filenames and the allowlist when applicable, and prevents closing the dialog on rejection.
  - composer-attachments.tsx: filters files/URLs according to the allowlist; addFiles now returns rejected filenames (string[]); addUrl returns Promise<boolean> indicating success.

- Attachment processing
  - Adds ComposerAttachmentKind 'file' (image | pdf | text | file).
  - Non-PDF binary attachments (spreadsheets, docs, etc.) are emitted as 'file' parts with their original MIME types; textual application/* types (JSON, XML, YAML, NDJSON) remain text parts.
  - File parts in core user messages use att.contentType-derived MIME instead of hardcoded application/pdf.

- MIME mappings
  - EXTENSION_TO_MIME expanded to include: txt, md, csv, json, xml, xls, xlsx, doc, docx, ppt, pptx.

- Tests & changeset
  - New/updated tests for accepted-types parsing/matching/accept attribute, composer-attachments behavior (filtering, URL HEAD content-type rejection, binary file-part handling). Local vitest: 20/20 for the attachment suite.
  - Changeset updated with usage example and formatting.

---

## Backward Compatibility

- Unset MASTRA_STUDIO_ATTACHMENT_TYPES preserves prior behavior (accept all).
- Binary files that were previously misread as text are now correct file parts — this may expose explicit model errors for unsupported MIME types.
- addFiles/addUrl now return values; callers that ignore them remain compatible.


emilcardell-lg:
Addressed all review feedback in bc4475b:

- **XSS (critical)** — `MASTRA_STUDIO_ATTACHMENT_TYPES` is now sanitized to MIME-legal characters (`[a-zA-Z0-9.+*/, -]`) at **both** injection sites (CLI server and the vite standalone plugin) before landing in the HTML shell's JS string literal. I went with a character allowlist rather than `JSON.stringify` + `escapeJsonForHtml` because the value is a constrained comma-separated MIME list: the allowlist removes every character that could break out of the string literal (quotes, backslashes, `<`, `>`), keeps `index.html` consumption unchanged, and also drops garbage input that isn't a valid MIME character. Happy to switch to the escape-and-parse approach if maintainers prefer consistency with `MASTRA_REQUEST_CONTEXT_PRESETS`.
- **Stale rejection alert** — `rejectedNames` now resets via `useEffect` when the dialog opens, per the suggested fix.
- **Changeset** — rewritten with a usage example and bullet formatting.

`pnpm vitest run src/lib/ai-ui/attachments` → 20/20 passing.

PR #17889, opened 2026-06-12
Native Agent goal mechanism (replaces MastraCode's between-turn judge loop)

abhiaiyer91:
## What this is

This makes **goals a first-class capability of the Agent** rather than a MastraCode-specific feature.

A *goal* is a durable, thread-scoped objective: a standing instruction that the agent keeps working toward across iterations until a judge model decides it is satisfied (or a run budget is exhausted). Previously this only existed inside MastraCode's TUI as a hand-rolled loop that re-invoked the agent between full turns and judged the result with a separate judge agent. That approach was MastraCode-only and, critically, could not evaluate a goal when a new message arrived in the middle of an already-running turn.

## How it works now

**Core (`@mastra/core`)**

- The objective lives in the existing `threadState` storage under a new `goal` slot (`GoalObjectiveRecord`), so it survives reloads and is serializable mid-run — exactly like the task list.
- A new **goal step** runs inside the agentic execution loop, right after `isTaskComplete`. On a real candidate answer it scores the conversation against the objective and gates the loop: not satisfied → continue, satisfied → stop and mark the objective `done`, budget exhausted → stop but stay resumable.
- Evaluation surfaces to consumers as a typed **`goal` stream chunk**, so the UI updates mid-run. This is what lets a message delivered into a live run still be judged against the goal.
- The **judge model is the activation switch**: if no judge resolves (neither the per-objective override nor the agent's `goal.judge` config), the goal step is a complete no-op — no scoring, no chunk, no budget consumed.
- `goal.judge` accepts a **resolver function**, so a host app can inject provider credentials and read the currently-selected judge at runtime. Bare model ids are resolved through the Mastra model router so gateways/auth apply.
- Settings precedence: per-objective record value wins, otherwise the agent's `goal` config default (`maxRuns` default 50, default judge prompt ported from MastraCode).
- New `Agent` methods: `setObjective`, `getObjective`, `clearObjective`, `updateObjectiveOptions` (all no-op without storage + thread).
- A goal that stops at its run budget stays `active` so raising `maxRuns` can resume it; a later run on the same thread is guarded against re-scoring (it stops and emits a terminal `goal` chunk without burning another judge call).

**MastraCode**

- `GoalManager` is now a thin adapter over the Agent's objective methods; all `/goal` behavior (status / pause / resume / clear, plan-started goals, status line, keyboard shortcuts, input lock) is preserved.
- The judge display is driven by the `goal` stream chunk instead of the old between-turn judge agent, which is removed.
- Judge configuration moved from a standalone `/judge` command to a **`/goal judge` subcommand**, since the judge only matters in the context of a goal.
- The goal judge model resolves through MastraCode's gateway (`getGoalJudgeModel`) so the configured judge authenticates with credentials from auth storage rather than env vars.

**`@mastra/react`**

- The accumulator passes the new `goal` chunk through as a UI-only lifecycle signal (no stray DB message), since the core goal step already injects any feedback into message history.

## Test plan

- **Core**: goal agent, goal-step, goal-scorer, goal-state-processor, and goal-signal-provider suites all pass; broader agentic-execution / is-task-complete / supervisor suites pass with no regressions.
- **MastraCode**: full suite passes (goal-manager, `/goal` command, agent-lifecycle, event-dispatch, command-dispatch, keyboard-shortcuts, queueing).
- **`@mastra/react`**: accumulator passthrough test for the `goal` chunk.
- **Builds, lint, typecheck**: `pnpm build:core`, `pnpm --filter ./packages/core check`, and full lint pass for both `@mastra/core` and `mastracode`.

## ELI5

This PR gives Agents a built-in, durable "goal" feature so you can set an objective for a conversation, have the agent automatically check progress using a judge model each turn, and either continue or stop based on whether the goal is met—persisting that objective across restarts and mid-run serialization.

---

## Overview

Adds a native, durable goal mechanism to Agent: thread-scoped objectives stored in threadState as GoalObjectiveRecord, projected into model context, evaluated in-loop by an LLM judge, and emitted as typed `goal` stream chunks so UIs can update mid-run. Replaces MastraCode’s between-turn judge loop with an integrated in-loop evaluation step and preserves MastraCode user-facing `/goal` behaviors via a thin adapter.

---

## Core features

- Agent-level config: optional `goal` on Agent with fields like `judge` (resolver or model id), `maxRuns`, and `prompt`.
- New Agent public methods:
  - setObjective({ threadId, ... })
  - getObjective({ threadId })
  - clearObjective({ threadId })
  - updateObjectiveOptions({ threadId, ... })
  - __getGoalConfig() (internal accessor)
- Durable objective type: `GoalObjectiveRecord` stored under `type: "goal"` in thread-state (objective, status, runsUsed, maxRuns, judgeModelId, prompt, timestamps).
- In-loop goal step (runs after isTaskComplete):
  - Skips for background tasks, tool-only iterations, or when no active objective/judge resolves.
  - Enforces budget: if runsUsed >= maxRuns → emits terminal goal chunk (maxRunsReached) and stops continuation but leaves objective `active` (resumable).
  - Scoring: resolves judge model (per-objective override → agent config → no-op), builds/uses scorer (custom or default), runs LLM judge, converts result to pass/fail + reason, updates runsUsed/status/updatedAt, sets isContinued accordingly, emits assistant feedback and a typed `goal` chunk with iteration/maxRuns/passed/results/timing/suppressFeedback.
  - Protects budget-exhausted goals from re-scoring on later runs.
- Goal state projection:
  - `GoalStateProcessor` emits `<current-objective>` state-signal snapshots for active objectives only; uses stable cache keys to avoid redundant re-emits.
  - `GoalSignalProvider` exposes the processor and is auto-registered when an agent has `goal` config (no-op without memory-backed thread support).
- Streaming & events:
  - New `GoalEvaluationPayload` and `goal` chunk arm in stream types.
  - Harness emits `goal_evaluation` events; Mastra React accumulator treats `goal` chunks as UI-only lifecycle signals (not DB-stored).
- Types & exports:
  - Re-exports and new barrel modules expose goal constants, processors, providers, scorer helper, and types across `@mastra/core` tools/signals/stream and thread-state exports.
- Tests:
  - Comprehensive tests added for core goal behavior, signal/provider/processor, MastraCode adapter, queueing/plan-mode integration, and React accumulator. CI reported passing for core, MastraCode, and react.

---

## MastraCode integration & refactor

- `/judge` top-level command removed; replaced by `/goal judge` subcommand wired to judge defaults UI and persistence.
- `GoalManager` refactored into a thin adapter over Agent objective APIs (set/get/clear/update) and retains UI behaviors (status/pause/resume/clear, plan-start flow, input lock, shortcuts).
- New handler `handleGoalEvaluation(ctx, payload)` updates judge display, mirrors evaluation to `state.goalManager.applyEvaluation`, clears `state.activeGoalJudge` when appropriate, and triggers plan-mode switching for matching plan-started goals.
- MastraCode’s judge display and model resolution now route through the MastraCode resolver to inject provider credentials when configured.
- Raw `<current-objective>` state snapshots are suppressed from the transcript; judge display driven by `goal` chunks.

---

## Settings & precedence

Effective judge/model/prompt resolution precedence:
1. Per-objective stored override (GoalObjectiveRecord)
2. Agent `goal` config (defaults can be set via MastraCode `/goal judge`)
3. Built-in defaults (e.g., DEFAULT_GOAL_MAX_RUNS = 50, DEFAULT_GOAL_JUDGE_PROMPT)

`goal.judge` accepts a resolver function so hosts can inject credentials or return `undefined` (to disable judging). Bare model id strings resolve through Mastra’s model router.

---

## Backwards compatibility & breaking changes

- Additive overall: new APIs, types, and chunk kind added. Existing workflows continue to function. MastraCode’s standalone `/judge` is removed (now `/goal judge`).

---

## Noted review items / open concerns

1. Changeset example hard-coded a model id; must use placeholder token `__GATEWAY_OPENAI_MODEL__`.
2. `getGoalJudgeModel` in MastraCode should thread `settingsPath` (or settings) consistently to avoid divergence from runtime settings.
3. `/goal pause` flow may race: pause should persist (await save) before clearing `activeGoalJudge` to prevent reload of an active objective.
4. Auto-wiring gap: `GoalSignalProvider` is only registered when agent `config.goal` exists—if threads contain per-objective judge config but agent lacks `goal`, the mechanism silently no-ops; consider auto-registration when storage+thread support exists or document agent-level requirement.
5. Remaining TODOs: document `goal.judge` resolver signature, verify `suppressFeedback` usage, remove payload casts so goal chunk payloads are typed end-to-end, consider less-generic `GOAL_STATE_ID`, and confirm single source-of-truth for `DEFAULT_GOAL_JUDGE_PROMPT`.

---

## Files / surface changes (high level)

- Core: new goal subsystem under packages/core/src/agent/goal (objective, scorer, state-processor, signal-provider), loop goal-step, Agent methods, types, stream chunk types, and tests.
- MastraCode: rewired to use Agent goal APIs; refactored goal-manager, command/handler updates, judge-display mapping, and tests updated.
- React SDK: accumulator updated to treat `goal` chunks as UI-only signals; tests added.
- Thread-state storage: new `GoalObjectiveRecord` type and export.
- Misc: pin vitest version in pnpm-workspace.yaml to 4.1.8 to resolve test versioning mismatch.


CalebBarnes:
Re-reviewed at `029b1dee03`. Design holds up. CodeRabbit already filed 10 actionable inline comments — flagging only the items not covered there, plus prior-pass items still open.

**Not covered by CodeRabbit**

1. **Changeset hard-codes a model id.** `.changeset/goal-mechanism-core.md:20` has `judge: 'openai/gpt-5.5'`. Per `AGENTS.md` line 8, model ids in changesets must use a placeholder token from `docs/src/plugins/remark-model-tokens/models.ts` — here `__GATEWAY_OPENAI_MODEL__`. Verified no other pending changeset hard-codes a model id, so this is also the in-practice convention.

2. **`getGoalJudgeModel` settings-path drift.** `mastracode/src/agents/model.ts:421-424` calls `loadSettings()` with no args, while `createMastraCode` uses `loadSettings(config?.settingsPath)`. With a non-default settings file, the judge model + maxTurns silently disagree with the rest of the runtime. Thread the resolved `settingsPath` (or the loaded settings) into the resolver.

3. **`/goal pause` isn't persisted before returning.** `mastracode/src/tui/setup.ts:218-221` calls `state.goalManager.pause()` and clears `activeGoalJudge`, but the paused record isn't awaited-saved. A subsequent `syncInitialThreadState` can reload the prior `active` objective.

4. **Per-objective-only activation gap.** Auto-wire of `GoalSignalProvider` only fires on `config.goal`. A thread whose `GoalObjectiveRecord` carries its own `judgeModelId` + `prompt` but with no agent-level `goal` config will silently no-op — even though the documented precedence (record > config > defaults) implies records alone should be able to activate goals. Either auto-wire when storage+thread are present, or document explicitly that agent-level `goal` config is required to enable the mechanism.

**Still open from prior pass**

- Document `goal.judge`'s resolver-form signature so typings don't drift.
- `suppressFeedback` on the chunk payload looks unused — verify it isn't dead.
- Payload casts in `goal-step.ts` chunk emission — payload could be typed end-to-end.
- Confirm `DEFAULT_GOAL_JUDGE_PROMPT` in core is the single source of truth (no stale duplicate in MastraCode).

CalebBarnes:
Re-reviewed at `b2af861656`. All 4 items from my last pass are addressed, and CodeRabbit's 10 actionable items are too. Nice turnaround.

**Verified locally** (after `pnpm build:core` to refresh the dist that re-exports `GOAL_STATE_ID`):
- `packages/core` goal tests: 13/13 (including the 3 new snapshot-retraction tests).
- `mastracode` full suite: 1305/1305.
- `tsc --noEmit` clean in both packages.
- CI on this head: 74 SUCCESS, 0 FAILURES, rest in-progress.

**One small follow-up (nit, non-blocking):**

`mastracode/src/tui/__tests__/mastra-tui-queueing.test.ts:859` — the new `persistedGoal` fixture hard-codes `judgeModelId: 'openai/gpt-5.5'`. `AGENTS.md:8` puts tests in scope for the placeholder-token rule, so it should pull from `docs/src/plugins/remark-model-tokens/models.ts` like the changeset now does.

CalebBarnes (approved):
Approving — all prior review items and CodeRabbit's actionable comments addressed in b2af861656 and follow-ups. CI green (0 failures), local tests pass (core goal 13/13, agent goal 14/14, mastracode 1305/1305), typecheck clean. Docs and changeset look good. Nice work.

Issue #17900, opened 2026-06-12
[BUG] Memory embedding cache grows without bound and can return another message's embeddings (32-bit hash collisions)


Abuhaithem:
### Describe the Bug

`Memory.embedMessageContent()` caches embeddings in a plain `Map` with **no eviction and no size limit**. The code comment says hashing is used so "we will [not] store all messages in memory for the life of the process" — but only the *key* is a hash; the cached *value* contains `chunks` (the full message text) **and** the full embedding vectors. So a long-running server with semantic recall enabled accumulates every distinct message (and every recall query string) it has ever embedded, in memory, forever. With a 1536-dim float embedding (~12 KB) plus raw text, that is roughly 12–20 KB per message — hundreds of MB over a few weeks of production traffic.

Separately, the cache key is a **32-bit** xxhash (`h32`). By the birthday bound, after ~77k distinct contents there is a >50% chance of at least one collision. On collision, `embedMessageContent` returns the *other* message's embeddings and chunks — silently storing wrong vectors in the vector index (corrupting recall for that message) or querying with the wrong vector.

`packages/memory/src/index.ts:971-1037`:

```ts
private hasher = xxhash();

// embedding is computationally expensive so cache content -> embeddings/chunks
private embeddingCache = new Map<
  number,
  {
    chunks: string[];
    embeddings: Awaited<ReturnType<typeof embedMany>>['embeddings'];
    usage?: { tokens: number };
    dimension: number | undefined;
  }
>();
...
protected async embedMessageContent(content: string) {
  const key = (await this.hasher).h32(content);   // 32-bit key
  const cached = this.embeddingCache.get(key);
  if (cached) return cached;
  ...
  this.embeddingCache.set(key, result);           // never evicted
  return result;
}
```

There is no `delete`, no LRU, and no max size anywhere in the class — the cache grows with total unique messages + unique recall queries.

Suggested fix: replace the `Map` with a bounded LRU (the core SemanticRecall cache already uses `lru-cache` with `max: 1000`), and switch `h32` → `h64` so collisions become negligible.

### Steps To Reproduce

Memory growth:

1. Create `new Memory({ vector, embedder, options: { semanticRecall: true } })`.
2. Save many short unique messages through `saveMessages` (or run a soak test calling `recall` with unique query strings).
3. Heap-snapshot the process: the embedding cache retains all chunk strings and embedding arrays; RSS grows linearly and never shrinks.

[...]

PR #17901, opened 2026-06-12
fix/memory-embedding-cache-unbounded-and-h32-collision

Abuhaithem:
Solves: 

The in-process embedding cache used by semantic recall (`Memory.embedMessageContent`) had two defects:

1. **Unbounded growth.** It stored results in a plain `Map` with no eviction. Each entry holds the content's chunk text **and** its embedding vectors, so a long-running `Memory` instance accumulated every message and recall query it ever embedded for the lifetime of the process — directly contradicting the code's own comment about keeping memory usage low (only the *key* was hashed; the *values* dominate memory).

2. **32-bit hash collisions corrupt recall.** Cache keys came from `xxhash.h32` (32 bits). By the birthday bound, distinct contents start colliding after tens of thousands of entries. On a collision, `embedMessageContent` returned **another message's** cached chunks and embeddings — silently indexing/recalling the wrong vectors with no error.

### Changes

- Replace the unbounded `Map` with an `LRUCache` (from `lru-cache`, already a direct dependency of `@mastra/memory`) capped at **1000** entries — matching the default used by the core `SemanticRecall` embedding cache (`packages/core/src/processors/memory/embedding-cache.ts`).
- Key the cache by `xxhash.h64` (64-bit) instead of `h32`, making collisions negligible. The cache key type changes from `number` to `bigint`.
- Make the cache-hit debug log BigInt-safe (`key.toString()`).
- Add regression tests.

No public API changes. Semantic recall results for correctly cached content are unchanged.

**Files changed**
- `packages/memory/src/index.ts` — LRU cache + `h64` keying
- `packages/memory/src/embedding-cache.test.ts` — new regression tests
- `.changeset/memory-embedding-cache-bounds.md` — patch changeset

## Related issue(s)

_No tracking issue yet — needs one filed and linked before submitting (see note above)._

## Type of change

- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [ ] Test update

## How it was tested

- **Reproduced both bugs on `main`** before fixing:
  - Unbounded growth: `embeddingCache.size === 2000` after embedding 2000 distinct contents.
  - Collision: with a brute-forced colliding pair (`"msg-4246"` and `"msg-268273"` both hash to `2346541822` under `h32`), `embedMessageContent("msg-268273")` returned `["msg-4246"]`'s chunks.
- **New regression tests** (`embedding-cache.test.ts`) — verified they **fail on the unpatched `index.ts`** and **pass with the fix**:
  - reuses cached embeddings for repeated content (cache hit)
  - bounds the cache via LRU eviction (1500 inserts → size ≤ 1000)
  - does not return wrong cached embeddings for the `h32`-colliding pair
- **No regressions**: full `@mastra/memory` unit suite — **32 files, 1134 passed, 1 todo**.
- `tsc --noEmit`, ESLint, and Prettier all clean on the changed files.

## Checklist

- [ ] I have linked the related issue(s) in the description above _(pending — file an issue first)_
- [x] I have made corresponding changes to the documentation (if applicable) _(none needed; internal behavior only)_
- [ ] I have addressed all Coderabbit comments on this PR _(after PR is opened)_

## Diff summary

```diff
+import { LRUCache } from 'lru-cache';

+// Max number of distinct contents whose embeddings are kept in the in-process cache.
+const DEFAULT_EMBEDDING_CACHE_MAX_SIZE = 1000;

-  private embeddingCache = new Map<
-    number,
+  private embeddingCache = new LRUCache<
+    bigint,
     { chunks: string[]; embeddings: ...; usage?: { tokens: number }; dimension: number | undefined }
-  >();
+  >({ max: DEFAULT_EMBEDDING_CACHE_MAX_SIZE });

-    const key = (await this.hasher).h32(content);
+    const key = (await this.hasher).h64(content);
-      this.logger.debug('Embedding cache hit', { contentHash: key, ... });
+      this.logger.debug('Embedding cache hit', { contentHash: key.toString(), ... });
```

Solves: #17900

## ELI5

This PR fixes two memory problems in the embedding cache used for semantic search: it was growing forever (using up more and more memory) and sometimes returning the wrong cached data because of poor hash key collision detection. The fix uses a fixed-size cache that automatically deletes old entries, and switches to better hashing so different messages never accidentally share cached results.

## Overview

Fixes two defects in the in-process embedding cache (`Memory.embedMessageContent`) used by semantic recall:

**Unbounded Memory Growth**: The cache previously used a plain JavaScript `Map` that accumulated every embedded message and recall query indefinitely, causing unbounded memory usage. It is now replaced with an LRUCache (from the lru-cache package) with a fixed limit of 1000 entries, matching the SemanticRecall default configuration.

**32-bit Hash Collisions Corrupting Results**: Cache keys were computed using xxhash.h32 (32-bit hashing), which caused collisions that led to returning another message's cached chunks and embeddings. The cache now uses xxhash.h64 (64-bit hashing) with bigint keys, making collisions negligible and ensuring cache isolation between different messages. Debug logging was updated to safely handle bigint keys via `toString()`.

## Changes

- **packages/memory/src/index.ts**: Replaces `Map` with `LRUCache` as the backing store, migrates cache keys from h32 (number) to h64 (bigint), adds configurable max cache size, and updates logging for bigint compatibility.
- **packages/memory/src/embedding-cache.test.ts**: Adds regression tests covering cache hits, LRU eviction (1500 inserts capped at ≤1000 entries), and correctness for messages that collide under h32 but not h64.
- **.changeset/memory-embedding-cache-bounds.md**: Patch-level changeset documenting the fix.

## Testing & Verification

- Both bugs were reproduced on main and verified fixed.
- New regression tests confirm cache hit behavior, LRU eviction correctness, and h32 collision separation.
- Full `@mastra/memory` unit suite passes: 1134 tests passed, 1 todo (32 test files).
- Type checking, ESLint, and Prettier validation pass on all modified files.

## Impact

No public API changes. Semantic recall results remain unchanged for correctly cached content.


Abuhaithem:
@abhiaiyer91 could u check this branch !

PR #17903, opened 2026-06-12
Document TrustedRouter OpenAI-compatible routing

jperla:
This pull request adds a short README note that TrustedRouter can be used with Mastra through an OpenAI-compatible base URL. TrustedRouter is an open-source, verifiable attested router for privacy-sensitive agent workflows with zero prompt and output logging by default.

Users can connect with `https://api.trustedrouter.com/v1` and a TrustedRouter API key while keeping the standard model routing interface.

Verification: `git diff --check`.

## ELI5

This PR updates the README to tell users about a special privacy-focused tool called TrustedRouter that they can use with Mastra. It's like adding a note to the instruction manual saying "by the way, you can also use this other trusted system if you care a lot about keeping your data private!"

## Summary

This pull request adds documentation to the README highlighting TrustedRouter as an option for privacy-sensitive agent workflows. The addition mentions that TrustedRouter can be used through Mastra's model routing interface via an OpenAI-compatible base URL (`https://api.trustedrouter.com/v1`), and notes its key feature of zero prompt and output logging by default. TrustedRouter is described as an open-source and verifiable attested router.

The change is integrated into the "Model routing" highlight section of the README's feature list, maintaining consistency with the existing documentation style.

**Lines changed:** +1/-1

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,7 +18,7 @@ Purpose-built for TypeScript and designed around established AI patterns, Mastra
 
 Some highlights include:
 
-- [**Model routing**](https://mastra.ai/models) - Connect to 40+ providers through one standard interface. Use models from OpenAI, Anthropic, Gemini, and more.
+- [**Model routing**](https://mastra.ai/models) - Connect to 40+ providers through one standard interface. Use models from OpenAI, Anthropic, Gemini, and more. For privacy-sensitive agents, [TrustedRouter](https://trustedrouter.com) can be used through the OpenAI-compatible base URL `https://api.trustedrouter.com/v1`; it is an open-source and verifiable attested router with zero prompt and output logging by default.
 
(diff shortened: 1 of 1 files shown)


abhiaiyer91:
Thanks for this PR but we will not accept this change

jperla:
Thanks for reviewing. Understood that this PR is not acceptable as written. Is there any format you would accept for this, such as a neutral OpenAI-compatible router guide, an integration recipe in docs, or an issue/discussion first? TrustedRouter is built as an end-to-end encrypted, open-source router, and I am happy to remove any wording that reads as promotional.

jperla:
One clarification on why I am asking about an acceptable format: TrustedRouter is built to be the #1 end-to-end encrypted open-source router, not just another provider listing. The value for users is a verifiable private OpenAI-compatible route for sensitive prompts. If this PR format is not acceptable, would a neutral custom OpenAI-compatible router recipe or issue be a better fit?

PR #17907, opened 2026-06-13
fix(deployer): default dev server bind to localhost (closes #17906)

### LINKED ISSUE #17906: [BUG] Mastra binds to `0.0.0.0`, log says it's binding to `localhost`
### 🎯  Log Message Mismatch
When `serverOptions.host` is `undefined` and `MASTRA_HOST` env var is unset, there're 2 variables with different values:

| `bindHost` | `host` |
|-------|-----|
| used in the hono server initialization to start the server |  used in log message, falls back to `localhost` when `bindHost` is `undefined` |
| Becomes `undefined` | Becomes `localhost` | 
* `bindHost` => used in the hono server initialization to start the server => Becomes `undefined`
* `host` => used in log message, falls back to `localhost` when `bindHost` is `undefined` => Becomes `localhost`

The default runtime behavior for nodejs when a bindhost is `undefined`, it binds to `0.0.0.0`

As a result, if you don't specify a hostname, the mastra server will bind to `0.0.0.0` (at least on nodejs) while the log says it's running on `localhost`, which is misleading.

### 🔓  Security-wise: if hostname is undefined, it causes unnecessary exposure to all networks directly reachable by the host
IMO, setting an app to all network interfaces should be a conscious decision by the developer, not a default behavior. An unaware developer may unnecessarily expose the app to more networks than needed, while thinking that it's just running on `localhost`

### Steps To Reproduce

1. Start a mastra project
2. Keep `serverOptions.host` undefined
4. Keep the `MASTRA_HOST` env var unset
5. Run `npm run dev` to trigger `mastra dev`
6. Observe the log message: 
  ```
│ Studio: http://localhost:4111
│ API:    http://localhost:4111/api
```
7. Verify that mastra is reachable via localhost AND other network interfaces: run `ifconfig` to get your LAN IP, and curl against it: `curl http://<LAN_IP>:4111`, you will get the HTML response

Not required

### Expected Behavior

* The log message should indicate the correct hostname
* Fallback behavior of defining no host should fallback to localhost, avoiding unnecessary network exposure
* Another reason why localhost as a fallback is favorable: Since it's not straightforward to infer what the runtime's default host binding is, I recommend resetting the fallback to localhost, while printing a warning that for production, or containerized environments, you need to set the host to the appropriate values.

### Environment Information

```block
System:
    OS: Linux 6.18 Ubuntu 24.04.3 LTS 24.04.3 LTS (Noble Numbat)
    CPU: (8) x64 Intel(R) Core(TM) i7-8550U CPU @ 1.80GHz
    Memory: 3.73 GB / 7.69 GB
    Container: Yes
    Shell: 5.9 - /usr/bin/zsh
  Binaries:
    Node: 26.3.0 - /home/hazem_wsl/.nvm/versions/node/v26.3.0/bin/node
    npm: 11.16.0 - /home/hazem_wsl/.nvm/versions/node/v26.3.0/bin/npm
  Browsers:
    Brave Browser: 143.1.85.117
  npmPackages:
    @mastra/core: ^1.42.0 => 1.42.0 
    @mastra/duckdb: ^1.4.2 => 1.4.2 
    @mastra/evals: ^1.3.0 => 1.3.0 
    @mastra/libsql: ^1.13.0 => 1.13.0 
    @mastra/loggers: ^1.1.2 => 1.1.2 
    @mastra/memory: ^1.20.3 => 1.20.3 
    @mastra/observability: ^1.14.1 => 1.14.1 
    @types/node: ^25.9.3 => 25.9.3 
    mastra: ^1.13.0 => 1.13.0 
    typescript: ^6.0.3 => 6.0.3 
    zod: ^4.4.3 => 4.4.3 
  npmGlobalPackages:
    npm: 11.16.0
```

### Verification

tsushanth:
## Summary

Closes #17906.

When neither `serverOptions.host` nor `MASTRA_HOST` was set, `bindHost` was `undefined`. Hono's `serve()` passes that through to Node's `http.Server.listen`, which treats `undefined` as the unspecified address and binds to every reachable IPv4 interface on the host (LAN, VPN, container bridge, etc.). At the same time the "Mastra API running" log substituted `'localhost'` for that case, so the running server silently exposed every interface while telling the operator it was bound to loopback. That's both a log-vs-bind mismatch and a security default issue, which the reporter laid out clearly.

Fix is to fall back to `'localhost'` explicitly when nothing else is set, so the bind matches the log and `mastra dev` doesn't default to a network-reachable surface. Folks who want all-interfaces binding can still opt in via `serverOptions.host` or `MASTRA_HOST=0.0.0.0` (the existing "uses MASTRA_HOST when it is set" test continues to cover that path).

## Test commands

```
pnpm build:core
pnpm --filter @mastra/server build
cd packages/deployer && npx vitest run src/server/__tests__/node-server-host.test.ts
```

The `node-server-host.test.ts` suite has two cases for this code path:

- `defaults hostname to localhost when no host is configured (closes #17906)` — new contract, asserts `hostname: 'localhost'` (was `hostname: undefined`).
- `uses MASTRA_HOST when it is set` — unchanged, asserts `hostname: '0.0.0.0'` when env is `0.0.0.0`.

The opt-in path (`MASTRA_HOST=0.0.0.0` or `serverOptions.host = '0.0.0.0'`) keeps the previous behavior verbatim.

## Notes

The reporter pinpointed the exact mismatch between `bindHost` (line 553) and `host` (line 554) in `packages/deployer/src/server/index.ts`, and the studio-injection block at line 453 had the same pattern; both are corrected. Changeset added at `.changeset/thirty-knives-worry.md`.

## ELI5

Imagine a security guard who tells people the front door is locked, but actually forgets to lock it—letting anyone nearby walk in. This PR fixes that: the Mastra dev server was saying it's only available locally (localhost) but was actually open to everyone on the network. Now it actually stays locked locally unless you explicitly ask to open it to the network.

## Overview

This PR fixes a host binding mismatch in the Mastra deployer where the dev server would advertise binding to `localhost` in logs but actually bind to `0.0.0.0` (all network interfaces), unintentionally exposing the development server to external networks. The fix ensures the server explicitly defaults to binding on `localhost` when neither `serverOptions.host` nor the `MASTRA_HOST` environment variable is configured, making the actual bind behavior match what is logged and preventing accidental exposure.

### Core Logic (`packages/deployer/src/server/index.ts`)
- Updated host resolution logic to explicitly default `bindHost` to `'localhost'` when both `serverOptions.host` and `MASTRA_HOST` are unset
- This ensures the value passed to the Node.js `serve()` function and the Studio HTML injection logic (host/protocol/port placeholders) consistently use `localhost` rather than `undefined`

### Tests (`packages/deployer/src/server/__tests__/node-server-host.test.ts`)
- Updated test assertions to expect `hostname: 'localhost'` when no explicit host configuration is provided
- Added clarifying comments explaining the expected default behavior
- Preserved existing tests for explicit configuration (e.g., `MASTRA_HOST=0.0.0.0`) to ensure opt-in all-interfaces binding still works

### Documentation (`.changeset/thirty-knives-worry.md`)
- Added changeset documenting the patch-level fix to the `@mastra/deployer` package
- Noted the default bind address change and instructions for explicitly enabling external access

## Behavior
- **Before**: No explicit host configured → server binds to `0.0.0.0` while logs report `localhost` (mismatch)
- **After**: No explicit host configured → server binds to `localhost` (matches logs)
- **Opt-in unchanged**: `serverOptions.host='0.0.0.0'` or `MASTRA_HOST=0.0.0.0` still allows external access

### DIFF
--- a/.changeset/thirty-knives-worry.md
+++ b/.changeset/thirty-knives-worry.md
@@ -1,0 +1,5 @@
+---
+'@mastra/deployer': patch
+---
+
+Fixed Mastra dev server defaulting to bind on `0.0.0.0` (all network interfaces) while logging that it was running on `localhost`. The dev server now defaults the bind address to `localhost`, matching the log message and avoiding silent network exposure. Set `serverOptions.host` or `MASTRA_HOST=0.0.0.0` when the server needs to be reachable beyond the loopback adapter. Closes #17906.
--- a/packages/deployer/src/server/__tests__/node-server-host.test.ts
+++ b/packages/deployer/src/server/__tests__/node-server-host.test.ts
@@ -102,13 +102,16 @@ describe('createNodeServer host binding', () => {
     delete process.env.MASTRA_HTTPS_CERT;
   });
 
-  it('leaves hostname undefined when no host is configured', async () => {
+  it('defaults hostname to localhost when no host is configured (closes #17906)', async () => {
     await createNodeServer(mockMastra, { tools: {} });
 
     expect(serveMock).toHaveBeenCalledOnce();
+    // Previously the bind was left undefined, which Node interprets as the
+    // unspecified address (binds to every interface) while the log claimed
+    // localhost. Default explicitly so the bind matches what we log.
     expect(serveMock.mock.calls[0]?.[0]).toMatchObject({
       port: 4111,
-      hostname: undefined,
+      hostname: 'localhost',
     });
     expect(logger.info).toHaveBeenCalledWith('Mastra API running', { url: 'http://localhost:4111/api' });
   });
--- a/packages/deployer/src/server/index.ts
+++ b/packages/deployer/src/server/index.ts
@@ -450,8 +450,8 @@ export async function createHonoServer(
       // Inject the server configuration into index.html placeholders
       const port = serverOptions?.port ?? (Number(process.env.PORT) || 4111);
       const hideCloudCta = process.env.MASTRA_HIDE_CLOUD_CTA === 'true';
-      const bindHost = serverOptions?.host ?? process.env.MASTRA_HOST;
-      const host = bindHost ?? 'localhost';
+      const bindHost = serverOptions?.host ?? process.env.MASTRA_HOST ?? 'localhost';
+      const host = bindHost;
       const key =
         serverOptions?.https?.key ??
         (process.env.MASTRA_HTTPS_KEY ? Buffer.from(process.env.MASTRA_HTTPS_KEY, 'base64') : undefined);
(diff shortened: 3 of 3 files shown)


TheIsrael1:
Thanks for the PR, We would have to close it though -  more info here https://github.com/mastra-ai/mastra/issues/17906#issuecomment-4706693772

PR #17910, opened 2026-06-13
fix(stores): memory list reads throw on backend failure instead of returning empty

### LINKED ISSUE #17783: [BUG] listThreads/listMessages in libsql and pg stores return empty on DB errors instead of throwing
### Describe the Bug

The paginated memory read methods in the libsql store catch database errors, log them, and then return an empty result like `{ threads: [], total: 0, hasMore: false }` instead of re-throwing. So a failed read looks exactly like a read that genuinely found nothing.

That's a problem because a transient `SQLITE_BUSY`, a locked table, or a dropped connection gives the caller the same response as "this resource has no threads/messages." Nothing is thrown, so the caller has no way to tell the store failed. There's only a log line.

Where I noticed the practical impact: an agent reading conversation history while the DB is briefly locked gets back `[]` and acts like the user has no history. If something downstream then writes state based on that empty read, real data can get dropped, and it all happens silently.

Affected code in `stores/libsql/src/storage/domains/memory/index.ts`:

- `listMessages`, catch block at lines 443-465
- `listThreads`, catch block at lines 1149-1175
- `listMessagesByResourceId`, same pattern

```ts
} catch (error) {
  const mastraError = new MastraError({ /* THIRD_PARTY ... */ }, error);
  this.logger?.error?.(mastraError.toString());
  this.logger?.trackException?.(mastraError);
  return { threads: [], total: 0, page, perPage: perPageForResponse, hasMore: false };
}
```

This already looks like a known issue that just wasn't applied everywhere. The spanner store throws here, and it has a regression test that spells out the intended behavior:

```ts
// stores/spanner/src/storage/index.test.ts
describe('error propagation (no empty-on-error)', () => {
  // Regression guard: backend failures must propagate as MastraError so
  // callers can distinguish "no data" from "store unavailable". The
  // prior implementation swallowed errors and returned {total:0,...}.
```

So the behavior is currently inconsistent across stores for these same list methods:

| Throws (correct) | Returns empty on error |
|---|---|
| spanner, clickhouse, cloudflare-d1, lance | libsql, pg, mongodb, upstash, dynamodb, mssql |

Worth noting this is only about the paginated list methods (`listThreads`, `listMessages`, `listMessagesByResourceId`). Point reads like `getMessagesById` return empty even in spanner, and those are out of scope here.

### Steps To Reproduce

1. Set up an agent or storage instance using `@mastra/libsql` (or `@mastra/pg`) for memory.
2. Call `storage.listThreads({ ... })` or `storage.listMessages({ ... })` while the underlying client fails once. The easiest way is to mock the libsql client's `execute` to reject a single time with `new Error('simulated backend outage')`.
3. The call resolves successfully with `{ threads: [], total: 0, hasMore: false }`.
4. No error is thrown, so the caller cannot tell the difference between a real failure and an empty result.

This is verifiable directly from the catch blocks at the lines listed above, and reproducible with a one-line mock that makes the client reject once. The repo already has the reference behavior in `stores/spanner/src/storage/index.test.ts` under `error propagation (no empty-on-error)`, and the libsql/pg stores currently would not pass an equivalent test.

### Expected Behavior

Backend failures in these list methods should propagate as a `MastraError`, the same way the spanner store does it, so callers can tell "no data" apart from "store unavailable." Validation errors (USER category) should keep throwing like they do now. Only a genuinely empty result should return an empty payload.

### Environment Information

```
System:
  OS: macOS 26.2
  CPU: (10) arm64 Apple M4
  Shell: 5.9 - /bin/zsh
Binaries:
  Node: 20.17.0
  npm: 10.8.2
  Yarn: 1.22.22

Affected packages:
  @mastra/libsql: 1.12.1
  @mastra/core: 1.42.0-alpha.3
  Same pattern also in: @mastra/pg, @mastra/mongodb, @mastra/upstash, @mastra/dynamodb, @mastra/mssql
Reference (already throws): @mastra/spanner, @mastra/clickhouse, @mastra/cloudflare-d1, @mastra/lance

LLM Provider: N/A (storage layer)
Browser: N/A
```

### Verification

I'm working on this issue and will raise a PR for the same!

JigyasuRajput:
## Description

The paginated memory list reads (`listThreads`, `listMessages`, `listMessagesByResourceId`) across most storage adapters caught backend DB failures, logged them, and then returned an empty payload like `{ threads: [], total: 0, hasMore: false }`. That made a transient backend failure (a locked table, a dropped connection, a one-off timeout) indistinguishable from a genuinely empty result: nothing was thrown, so a caller could not tell "the store is down" apart from "there is no data."

The practical impact is on agents: an agent reading conversation history during a brief outage gets back `[]`, assumes the user has no history, and can silently overwrite real state.

This PR makes those reads re-throw the failure as a `MastraError` instead of swallowing it, matching the `spanner` store, which already behaved this way and had a regression test for it.

What stays the same:

- USER-category validation errors still throw as before (they are re-thrown first, so a bad argument is never reclassified as a backend failure).
- A genuinely empty result still returns an empty payload from inside the `try`.
- Point reads like `listMessagesById` are out of scope and still return empty.

Stores updated: `libsql`, `pg`, `mongodb`, `upstash`, `dynamodb`, `mssql`, `cloudflare-d1`, `clickhouse`, `lance`, `dsql`, `mysql`, `redis`. (`spanner` already behaved correctly and is the reference; `convex` already lets errors propagate.) Each store gets a regression test that simulates a one-off backend rejection and asserts the read throws with the right error id and keeps the original error as the cause.

## Related issue(s)

Fixes #17783

## Type of change

- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [ ] Test update

## Checklist

## ELI5

When a memory store fails while loading threads or messages, it previously looked like there was no data. This PR reports the failure instead, so callers can handle the outage correctly.

## Overview

This PR updates paginated memory list reads across multiple storage adapters. Backend failures now throw `MastraError` instead of returning empty results.

Updated stores include libsql, pg, mongodb, upstash, dynamodb, mssql, cloudflare-d1, clickhouse, lance, dsql, mysql, redis, and spanner. Convex already propagated errors.

## Behavior Changes

- Affected methods include `listThreads`, `listMessages`, and `listMessagesByResourceId` where implemented.
- Backend failures now throw operation-specific `MastraError` values.
- `USER`-category validation errors remain unchanged.
- Successful empty results remain unchanged.
- Point reads remain unchanged.

## Tests

Each updated store includes regression coverage that:

- Simulates a backend failure.
- Verifies the expected error ID.
- Verifies that the original error is preserved as the cause or message.
- Confirms that failures do not return empty payloads.

## Migration Notes

Callers that invoke these methods directly must handle thrown errors with `try/catch` or rejection handlers.

## Related Issues

- Fixes `#17783`


JigyasuRajput:
btw heads up on the diff size: this is the same one-line bug repeated across stores. 

And listThreads/listMessages/listMessagesByResourceId are abstract per adapter, so each store has its own catch that swallows the error and returns empty. There's no shared chokepoint to fix it in one place. Also the change only touches the error path. Every method's success path is unchanged, and each store has its own regression test.

TheIsrael1:
## 🟠 High: Maintainer notes

### Scope

PR #17910 fixes #17783 by changing paginated memory list reads across storage adapters to throw backend failures instead of returning empty results.

### Context

The linked issue describes a real memory correctness problem: a transient storage outage could look identical to “no threads/messages,” so agent memory reads could proceed without existing history. The PR updates the repeated adapter-level catch blocks for `listThreads`, `listMessages`, and `listMessagesByResourceId` where applicable.

### Evidence checked

- PR body, linked issue #17783, issue comments/labels, changed files, representative diff hunks, added regression tests, merge/check status, and PR comments/review comments.
- Merge conflict was resolved in `stores/upstash/src/storage/index.test.ts` by keeping both the PR's memory error-propagation tests and `main`'s workflow snapshot coverage.
- Verification passed: `pnpm --filter @mastra/upstash test -- src/storage/index.test.ts`.
- Checks seen before the maintainer fix: labeler, issue-link-nudge, complexity, summarize, Socket Security, and CodeRabbit succeeded. Vercel docs/playground UI checks failed but do not appear relevant to these storage changes.

### Observations

- The behavior change matches the issue: backend failures now become `MastraError`s while USER validation errors and genuine empty successful reads remain unchanged.
- The broad diff is mostly mechanical across adapter-specific implementations; there does not appear to be a shared storage-layer chokepoint to make this a one-file fix.
- The changeset documents the caller-visible behavior change and direct-caller migration guidance.
- Maintainer-side conflict fix was applied and verified; no lint/check fix or inline suggestion application was found during triage.

JigyasuRajput:
lmk if there's anything on my end!
thanks for the maintainer notes @TheIsrael1 🙌🏻

mikhael28:
@coderabbitai please re-review

JigyasuRajput:
@mikhael28 none of the red checks look related to this change.

mongodb: all tests passed, then the job hung for ~12 min with no output and got cancelled on timeout. My added tests use a mocked `connectorHandler` and never open a `MongoClient`, so nothing to leak. Same error-propagation tests on pg/mysql/libsql/clickhouse passed, so it reads like flaky teardown on the live replica-set suite. A re-run should do it.

@coderabbitai can you please re-review and show that here (it got cutoff in your last review)

PR #17912, opened 2026-06-13
fix(core): interrupt workflow sleep on cancel (closes #17908)

### LINKED ISSUE #17908: [BUG] Workflow `cancel()` does not interrupt `sleep()` / `sleepUntil()` — the run keeps executing and overwrites the canceled status
### Describe the Bug

Calling `Run.cancel()` while a workflow run is parked in `.sleep(ms)` or `.sleepUntil(date)` does **not** interrupt the wait.

`cancel()` aborts the run's `AbortController` and writes `status: 'canceled'` to storage, but the default execution engine's sleep primitives are plain `setTimeout` promises that never observe the abort signal (`packages/core/src/workflows/default.ts`):

```ts
async executeSleepDuration(duration: number, _sleepId: string, _workflowId: string): Promise<void> {
  await new Promise(resolve => setTimeout(resolve, duration < 0 ? 0 : duration));
}

async executeSleepUntilDate(date: Date, _sleepUntilId: string, _workflowId: string): Promise<void> {
  const time = date.getTime() - Date.now();
  await new Promise(resolve => setTimeout(resolve, time < 0 ? 0 : time));
}
```

As a result, a run canceled mid-sleep:

1. **keeps the Node timer alive for the full duration** — a `.sleepUntil(+24h)` holds the run in memory for 24 hours after cancellation;
2. **wakes up and overwrites the canceled status** — `packages/core/src/workflows/handlers/entry.ts` unconditionally calls `persistStepUpdate({ workflowStatus: 'running' })` after the sleep, replacing the `'canceled'` status that `cancel()` just persisted; and
3. **publishes a sleep-step `success` event** before the engine finally notices the aborted signal at the end of the entry handler.

So observers (Studio, `watch()`, storage queries) see the run flip `canceled` → `running` → `canceled`, and in-process the cancellation simply does not take effect until the timer fires. The handlers already receive `abortController`, but never use it for the wait itself.

### Steps To Reproduce

2. Add the following workflow:

   ```ts
   import { createWorkflow, createStep } from '@mastra/core/workflows';
   import { z } from 'zod';

   const a = createStep({ id: 'a', inputSchema: z.object({}), outputSchema: z.object({}), execute: async () => ({}) });
   const b = createStep({
     id: 'b',
     inputSchema: z.object({}),
     outputSchema: z.object({}),
     execute: async () => { console.log('b ran'); return {}; },
   });

   export const wf = createWorkflow({ id: 'w', inputSchema: z.object({}), outputSchema: z.object({}) })
     .then(a)
     .sleep(60_000)
     .then(b)
     .commit();
   ```

3. Start the run and cancel it ~1s in, during the `sleep(60_000)`:

   ```ts
   const run = await wf.createRun();
   const result = run.start({ inputData: {} });
   setTimeout(() => run.cancel(), 1000);
   await result;
   ```

4. Observe:
   - `run.start()` does **not** resolve until ~60s later (the timer is never cleared);
   - at wake-up the run row briefly reads `running` again, and the sleep step is recorded/published as `success`;
   - step `b` is still reached after the canceled sleep.

Reproducible directly against `@mastra/core` with the workflow above (default engine + an in-memory store), calling `run.cancel()` during the sleep. A self-contained reproduction/regression test is included in the fix at `packages/core/src/workflows/cancel-sleep.test.ts`.  https://github.com/mastra-ai/mastra

### Expected Behavior

`cancel()` should interrupt an in-flight sleep immediately:

- the timer is cleared (and does not keep the process alive),
- the run settles as `canceled` right away,
- the step after the sleep never runs, and
- no `running` status or sleep-step `success` is persisted/published after cancellation.

### Environment Information

```block
System:
  OS: Linux 6.17 Ubuntu 24.04.4 LTS
  Memory: 3.92 GB / 15.49 GB
  Shell: /bin/bash
Binaries:
  Node: 22.22.2
  npm: 10.9.7
  pnpm: 11.5.1
npmPackages:
  @mastra/core: 1.42.0

Affected package: @mastra/core (packages/core/src/workflows/default.ts, handlers/entry.ts, handlers/sleep.ts)
```

### Verification

tsushanth:
Closes #17908.

`Run.cancel()` aborts the run's `AbortController` and writes `status: 'canceled'` to storage, but the default engine's `executeSleepDuration` / `executeSleepUntilDate` were plain `setTimeout` promises that never observed the abort signal. Three downstream effects:

1. the Node timer stayed alive for the full sleep duration; a `.sleepUntil(+24h)` held the run in memory for 24h after cancellation
2. when the timer eventually fired, the entry handler unconditionally called `persistStepUpdate({ workflowStatus: 'running' })` and published a sleep-step `success` event, overwriting the `'canceled'` status that `cancel()` had persisted
3. the step after the sleep ran anyway

## Fix

Wire the run's abort signal through to the sleep primitives and skip the post-sleep "running" status flip / success event when the signal is already aborted.

- `executeSleepDuration` / `executeSleepUntilDate` take an optional `signal?: AbortSignal`. The default implementation uses a small `abortableTimeout` helper that clears the underlying timer and resolves the promise as soon as the signal fires (and short-circuits when the signal is already aborted on entry). The parameter is optional so existing subclasses (e.g. Inngest) are not broken by the signature change.
- `handlers/sleep.ts` passes `abortController.signal` to both calls.
- `handlers/entry.ts`, for both `sleep` and `sleepUntil` branches, checks `abortController.signal.aborted` after the sleep returns. When aborted, it records `execResults` as `{ status: 'canceled', startedAt, endedAt }` and skips the persistStepUpdate('running') + sleep-step success/finish publishes. The existing terminal block at the end of `executeEntry` then persists the canceled status and publishes the workflow-canceled event exactly as it did before.

## Tests

`packages/core/src/workflows/cancel-sleep.test.ts`:
- a 60s `executeSleepDuration` whose signal aborts after 50ms must return in under 1s
- a 60s `executeSleepDuration` whose signal is pre-aborted must return in under 50ms
- a 60s `executeSleepUntilDate` whose signal aborts after 50ms must return in under 1s
- without a signal, a 150ms `executeSleepDuration` must still run for ~150ms (no regression)

All four pass against the patched engine and would fail against `main`. The full `pnpm --filter @mastra/core test workflows` suite (37 files, 949 tests) stays green.

## ELI5

When you tell a workflow to cancel while it's sleeping, it should wake up right away and stop. Previously, it would ignore the cancel command and finish sleeping anyway. This PR fixes that by making the sleep function listen for a "stop" signal, so cancellation interrupts the sleep immediately and the workflow shuts down cleanly.

## Changes Overview

This PR fixes workflow cancellation behavior during sleep operations by introducing interrupt-aware sleep implementations:

### Core Changes

**1. Abortable Sleep Helper** (`packages/core/src/workflows/default.ts`)
- Added internal `abortableTimeout(ms, signal?)` helper that clears timers and resolves immediately when an `AbortSignal` fires
- Updated `executeSleepDuration` and `executeSleepUntilDate` to accept optional `signal?: AbortSignal` parameter and use abortable waits
- Preserves backward compatibility by making the signal parameter optional; existing behavior (clamping negative durations/times to 0) unchanged

**2. Signal Propagation** (`packages/core/src/workflows/handlers/sleep.ts`)
- `executeSleep` and `executeSleepUntil` now pass `abortController?.signal` to the underlying engine sleep methods
- Error handling, span lifecycle, and duration computation remain unchanged

**3. Post-Sleep Cancellation Handling** (`packages/core/src/workflows/handlers/entry.ts`)
- After sleep completes, the entry handler now checks `abortController.signal.aborted`
- When aborted: sets `execResults` to `{ status: 'canceled', startedAt, endedAt }` and skips post-sleep persist and event publishing
- When not aborted: preserves existing behavior (persists running status, publishes sleep-step success/finish events)
- Existing terminal block still persists canceled status and publishes workflow-canceled event as before

### Test Coverage

New test suite (`packages/core/src/workflows/cancel-sleep.test.ts`) with four test cases:
- `executeSleepDuration` with signal aborting after 50ms resolves in <1s
- `executeSleepDuration` with pre-aborted signal returns in <50ms
- `executeSleepUntilDate` with signal aborting after 50ms resolves in <1s
- `executeSleepDuration` without signal still waits ~150ms (regression test)

### Outcomes

- Canceling a workflow during sleep immediately interrupts the timer instead of waiting for completion
- Timers are cleared, preventing memory leaks from long-duration sleeps (e.g., `sleepUntil(+24h)`)
- Run status settles as `canceled` immediately without overwriting to `running`
- Subsequent workflow steps do not execute after cancellation
- No running status or sleep-step success events are published/persisted after cancellation

All tests pass; full core workflow test suite remains green.

### DIFF
--- a/.changeset/curly-rocks-matter.md
+++ b/.changeset/curly-rocks-matter.md
@@ -1,0 +1,5 @@
+---
+'@mastra/core': patch
+---
+
+Fixed `run.cancel()` not interrupting `sleep()` / `sleepUntil()` in the default workflow execution engine. The pending `setTimeout` was previously held until it fired naturally, which kept the run in memory for the full sleep duration, then flipped its status from 'canceled' back to 'running' and recorded the sleep step as 'success'. The sleep primitives now observe the run's abort signal: when `cancel()` is called the timer is cleared, the run settles as 'canceled' immediately, and the next step does not execute (closes #17908).
--- a/packages/core/src/workflows/cancel-sleep.test.ts
+++ b/packages/core/src/workflows/cancel-sleep.test.ts
@@ -1,0 +1,54 @@
+import { describe, expect, it } from 'vitest';
+import { DefaultExecutionEngine } from './default';
+
+describe('DefaultExecutionEngine sleep + cancel (#17908)', () => {
+  it('executeSleepDuration resolves immediately when its abort signal fires', async () => {
+    const engine = new DefaultExecutionEngine({ mastra: undefined });
+    const abortController = new AbortController();
+
+    const startedAt = Date.now();
+    const sleeping = engine.executeSleepDuration(60_000, 'sleep-1', 'wf', abortController.signal);
+
+    setTimeout(() => abortController.abort(), 50);
+
+    await sleeping;
+
+    const elapsed = Date.now() - startedAt;
+    expect(elapsed).toBeLessThan(1_000);
+  });
+
+  it('executeSleepDuration resolves immediately when its signal is already aborted', async () => {
+    const engine = new DefaultExecutionEngine({ mastra: undefined });
+    const abortController = new AbortController();
+    abortController.abort();
+
+    const startedAt = Date.now();
+    await engine.executeSleepDuration(60_000, 'sleep-1', 'wf', abortController.signal);
+    expect(Date.now() - startedAt).toBeLessThan(50);
+  });
+
+  it('executeSleepUntilDate resolves immediately when its abort signal fires', async () => {
+    const engine = new DefaultExecutionEngine({ mastra: undefined });
+    const abortController = new AbortController();
+    const target = new Date(Date.now() + 60_000);
+
+    const startedAt = Date.now();
+    const sleeping = engine.executeSleepUntilDate(target, 'sleepUntil-1', 'wf', abortController.signal);
+
+    setTimeout(() => abortController.abort(), 50);
+
+    await sleeping;
+
+    expect(Date.now() - startedAt).toBeLessThan(1_000);
+  });
+
+  it('executeSleepDuration still completes its full duration without a signal', async () => {
+    const engine = new DefaultExecutionEngine({ mastra: undefined });
+
+    const startedAt = Date.now();
+    await engine.executeSleepDuration(150, 'sleep-1', 'wf');
+    const elapsed = Date.now() - startedAt;
+
+    expect(elapsed).toBeGreaterThanOrEqual(140);
+  });
+});
--- a/packages/core/src/workflows/default.ts
+++ b/packages/core/src/workflows/default.ts
@@ -48,6 +48,21 @@ import type {
 // Re-export ExecutionContext for backwards compatibility
 export type { ExecutionContext } from './types';
 
+function abortableTimeout(ms: number, signal?: AbortSignal): Promise<void> {
+  if (signal?.aborted) return Promise.resolve();
+  return new Promise<void>(resolve => {
+    const timer = setTimeout(() => {
+      signal?.removeEventListener('abort', onAbort);
+      resolve();
+    }, ms);
+    const onAbort = () => {
+      clearTimeout(timer);
+      resolve();
+    };
+    signal?.addEventListener('abort', onAbort, { once: true });
+  });
+}
+
 /**
  * Default implementation of the ExecutionEngine
  */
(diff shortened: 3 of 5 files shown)


tsushanth:
Same situation as #17911 — @Abuhaithem opened #17909 about 2 hours before I shipped this, and I missed it because the `Fixes:` keyword wasn't there. Closing in favor of #17909; sorry for the duplicate.

PR #17894, opened 2026-06-12
feat(mcp): request-scoped progress notifications in serverless mode

noahjohnhay:
Closes #17926

## Summary

`MCPServer.handleServerlessRequest()` creates a transient `StreamableHTTPServerTransport` with `enableJsonResponse: true` hard-coded. That buffers each request into a single JSON-RPC response and silently drops any `notifications/progress` a tool emits via `extra.sendNotification()`. As a result, serverless MCP tools can call `sendNotification()` but the client's `onprogress` handler never fires.

This adds an opt-in `serverlessStreaming: true` flag that handles the request with **request-scoped SSE streaming** (`enableJsonResponse: false`) instead, so progress notifications reach the client before the final result — while staying fully stateless.

```ts
await server.startHTTP({
  url, httpPath: '/mcp', req, res,
  options: { serverless: true, serverlessStreaming: true },
});
```

## Why

This reconciles two existing capabilities that were mutually exclusive in serverless mode:

- #9585 added serverless mode by disabling streaming (`enableJsonResponse: true`, `sessionIdGenerator: undefined`).
- #10637 added progress-notification support, which serverless mode then dropped.

The MCP SDK already supports stateless per-request SSE when `sessionIdGenerator: undefined` and `enableJsonResponse: false` — Mastra just needed to stop forcing JSON-response mode for every serverless request. This also aligns with the direction in modelcontextprotocol/modelcontextprotocol#2575.

## Changes

- `startHTTP()` options gain `serverlessStreaming?: boolean` (defaults to `false`).
- `startHTTP()` computes `enableJsonResponse` (`serverlessStreaming === true` → `false`; otherwise honors an explicit `enableJsonResponse`, defaulting to `true`) and passes it through.
- `handleServerlessRequest()` now receives and applies `enableJsonResponse` instead of hard-coding `true`.
- Reference docs (`reference/tools/mcp-server.mdx`) document the flag and its scope.
- Changeset added (`@mastra/mcp` minor).

## Behavior & compatibility

- **Default unchanged.** Without the flag, serverless still returns buffered JSON responses, so existing serverless users are unaffected.
- **Still stateless.** No `mcp-session-id` is required or persisted in either mode.
- **Scope.** This enables only request-scoped notifications such as **progress**. Elicitation, resource subscriptions, and out-of-request resource/list-change notifications still require session state or a different protocol model, and remain unavailable in serverless mode.

## Tests

Added to `packages/mcp/src/server/server.test.ts`:

[...]

PR #9585, opened 2025-10-31
fix: add flag to skip sessions and streaming in serverless mcp

graysonhicks:
## Description

- Added a serverless: true option to the startHTTP() method in MCPServer.
- When enabled, the server bypasses session and streaming logic, using a transient StreamableHTTPServerTransport with enableJsonResponse: true and no sessionIdGenerator.
- The in-memory maps for session tracking (streamableHTTPTransports, httpServerInstances) are skipped in this mode.
- Default behavior (persistent sessions and streaming) is unchanged.

### Purpose

- Addresses the problem that persistent MCP sessions and streaming transports do not work reliably in serverless or edge runtimes, where connections are short-lived and state cannot be maintained.
- Introduces a stateless request-response mode suitable for serverless environments.
- Preserves full compatibility with traditional long-running server deployments.

### Trade-Offs

- Sessions are not maintained; each request starts fresh.
- Streaming output is disabled; responses are sent only after completion.
- Subscriptions, logs, and other live updates are not supported.
- Diverges slightly from the MCP spec’s stateful connection model.
- Best suited for single, stateless tool calls rather than multi-turn interactive flows.

## Type of Change

- [ ] Bug fix (non-breaking change that fixes an issue)
- [ x] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [ ] Test update

## Checklist

- [ x] I have made corresponding changes to the documentation (if applicable)
- [ ] I have added tests that prove my fix is effective or that my feature works

## Summary by CodeRabbit

* **New Features**
  * Added serverless mode for HTTP transport — stateless, per-request MCP operation for serverless platforms (no session persistence; some persistent-connection features unavailable).

* **Documentation**
  * Updated docs with serverless configuration, usage examples (Edge/Workers/Vercel/Deno), integration patterns, limitations, and default session ID behavior.

* **Chores**
  * Added release changeset for this patch.

PR #17927, opened 2026-06-15
feat(mcp): request-scoped progress notifications in serverless mode

### LINKED ISSUE #17926: [BUG] @mastra/mcp serverless mode drops notifications/progress (forces enableJsonResponse: true)
### Description

In serverless mode (`MCPServer.startHTTP({ options: { serverless: true } })`), `handleServerlessRequest()` creates a transient `StreamableHTTPServerTransport` with `enableJsonResponse: true` hard-coded:

```ts
const tempTransport = new StreamableHTTPServerTransport({
  sessionIdGenerator: undefined,
  enableJsonResponse: true,
});
```

This buffers each request into a single JSON-RPC response, so any `notifications/progress` a tool emits via `extra.sendNotification()` are silently discarded. A serverless tool can call `sendNotification()` without error, but the client's `onprogress` handler **never fires**.

This reconciles two existing capabilities that are currently mutually exclusive in serverless mode:

- #9585 added serverless mode by disabling streaming.
- #10637 added progress-notification support, which serverless mode then drops.

The MCP SDK already supports stateless per-request SSE when `sessionIdGenerator: undefined` and `enableJsonResponse: false`, so the transport behavior is available — Mastra just forces JSON-response mode for every serverless request.

### Steps to reproduce

1. Start an `MCPServer` with `startHTTP({ options: { serverless: true } })`.
2. Add a tool whose `execute` calls `context.mcp.extra.sendNotification({ method: 'notifications/progress', params: { progressToken, progress, total } })`.
3. From an MCP client, call the tool with an `onprogress` handler.

### Expected behavior

The client's `onprogress` handler receives the progress notifications before the final result.

### Actual behavior

`onprogress` is never called; only the final buffered result is returned.

### Proposed fix

Add an opt-in flag (e.g. `serverlessStreaming: true`) so serverless requests can be handled with request-scoped SSE streaming (`enableJsonResponse: false`), while keeping the JSON-response default for backward compatibility and remaining fully stateless. Tracked in #17894.

### Environment

- `@mastra/mcp` (current `main`)

noahjohnhay:
Closes #17926

## Summary

`MCPServer.handleServerlessRequest()` creates a transient `StreamableHTTPServerTransport` with `enableJsonResponse: true` hard-coded. That buffers each request into a single JSON-RPC response and silently drops any `notifications/progress` a tool emits via `extra.sendNotification()`. As a result, serverless MCP tools can call `sendNotification()` but the client's `onprogress` handler never fires.

This adds an opt-in `serverlessStreaming: true` flag that handles the request with **request-scoped SSE streaming** (`enableJsonResponse: false`) instead, so progress notifications reach the client before the final result — while staying fully stateless.

```ts
await server.startHTTP({
  url, httpPath: '/mcp', req, res,
  options: { serverless: true, serverlessStreaming: true },
});
```

## Why

This reconciles two existing capabilities that were mutually exclusive in serverless mode:

- #9585 added serverless mode by disabling streaming (`enableJsonResponse: true`, `sessionIdGenerator: undefined`).
- #10637 added progress-notification support, which serverless mode then dropped.

The MCP SDK already supports stateless per-request SSE when `sessionIdGenerator: undefined` and `enableJsonResponse: false` — Mastra just needed to stop forcing JSON-response mode for every serverless request. This also aligns with the direction in modelcontextprotocol/modelcontextprotocol#2575.

## Changes

- `startHTTP()` options gain `serverlessStreaming?: boolean` (defaults to `false`).
- `startHTTP()` computes `enableJsonResponse` (`serverlessStreaming === true` → `false`; otherwise honors an explicit `enableJsonResponse`, defaulting to `true`) and passes it through.
- `handleServerlessRequest()` now receives and applies `enableJsonResponse` instead of hard-coding `true`.
- Reference docs (`reference/tools/mcp-server.mdx`) document the flag and its scope.
- Changeset added (`@mastra/mcp` minor).

## Behavior & compatibility

- **Default unchanged.** Without the flag, serverless still returns buffered JSON responses, so existing serverless users are unaffected.
- **Still stateless.** No `mcp-session-id` is required or persisted in either mode.
- **Scope.** This enables only request-scoped notifications such as **progress**. Elicitation, resource subscriptions, and out-of-request resource/list-change notifications still require session state or a different protocol model, and remain unavailable in serverless mode.

## Tests

Added to `packages/mcp/src/server/server.test.ts`:

1. Serverless JSON-response mode (default) still returns final tool results.
2. Serverless streaming mode delivers `notifications/progress` to a real MCP SDK `Client.callTool(..., { onprogress })` handler, and still returns the final result.
3. Serverless mode does not require or persist an `mcp-session-id`.

All 83 server tests pass; `eslint` and `prettier` clean.

## ELI5

In serverless mode, Mastra’s MCP server was buffering everything into one big JSON response, so progress updates (like “I’m working on it…”) never reached the client. This PR adds an opt-in `serverlessStreaming: true` flag that makes progress notifications stream to the client in real time—without needing any session storage.

---

### Core Implementation
- **server.ts**
  - Added `serverlessStreaming?: boolean` to `MCPServer.startHTTP()` options.
  - Computes the transport setting with precedence:  
    **`enableJsonResponse = options?.enableJsonResponse ?? !options?.serverlessStreaming`**
  - Passes `enableJsonResponse` into serverless handling instead of hard-coding JSON mode.
  - When `serverlessStreaming: true` (and `enableJsonResponse` isn’t explicitly set), serverless uses SSE streaming (`enableJsonResponse: false`) so MCP tool **`notifications/progress`** reaches the client’s **`onprogress`** handler before the final result.
  - Default behavior remains unchanged: serverless continues using buffered JSON responses unless `serverlessStreaming: true` is set.
  - Keeps the server fully stateless: no `mcp-session-id` is required/persisted; only request-scoped progress can be streamed.

### Tests
Updated/extended `packages/mcp/src/server/server.test.ts` with coverage for:
- **Default serverless JSON-response mode** still returns buffered tool results.
- **Serverless streaming mode** surfaces progress notifications to the MCP SDK client via `onprogress`.
- **Explicit `enableJsonResponse` overrides `serverlessStreaming`** (e.g., `{ serverlessStreaming: true, enableJsonResponse: true }` buffers instead of streaming progress).
- **Stateless serverless semantics**: `mcp-session-id` remains `undefined` and serverless flows still work across requests.

### Documentation / Release Note
- **mcp-server.mdx**: Documented `serverlessStreaming: true` for `MCPServer.startHTTP()` with an example, and clarified session-dependent MCP features are still unavailable in serverless mode.
- **changeset/serverless-mcp-progress-streaming.md**: Added a release note explaining the new opt-in behavior, the progress-notification fix, and the `enableJsonResponse` precedence.

---

## Design Decisions
- **Backward compatible by default** (still JSON-response mode unless `serverlessStreaming: true` is enabled).
- **Explicit precedence**: `enableJsonResponse` always wins over `serverlessStreaming`.
- **Stateless implementation**: enables request-scoped progress streaming only; session-dependent features still require session state.

---

## Testing
- All server tests pass (including the new serverless streaming/progress and precedence/statelessness assertions).

---
_Supersedes #17894, which GitHub would not let me reopen after it was auto-closed by the `needs-issue` gate. Same branch and commits (`f6405fd`), now opened with the linked issue in place._


noahjohnhay:
Thanks @coderabbitai — both addressed in e418cff:

- **Explicit `enableJsonResponse` precedence:** reworked to `options?.enableJsonResponse ?? !options?.serverlessStreaming`, so an explicit `enableJsonResponse` always wins and `serverlessStreaming` only acts as the default. Added a test (`should let an explicit enableJsonResponse override serverlessStreaming`) verifying that `{ serverlessStreaming: true, enableJsonResponse: true }` buffers in JSON mode and delivers no streamed progress.
- **Changeset code example:** added the `startHTTP({ … serverlessStreaming: true })` snippet per the coding guideline.

All 5 serverless tests pass; eslint/prettier clean.

abhiaiyer91 (approved):

PR #17934, opened 2026-06-15
fix(stores): updateThread now bumps the thread's updatedAt

### LINKED ISSUE #17932: [BUG] libsql updateThread does not bump updatedAt
## Describe the Bug

In the libsql storage adapter, `updateThread` updates the thread title and metadata but never touches `updatedAt`. The timestamp is missing from the SQL SET clause, and the returned object just spreads the stale value from the thread that was read in.

This goes against how it is supposed to work. The in-memory reference implementation bumps `updatedAt` on every update, and so do the siblings in the same file (`saveMessages` and `updateMessages` both run `UPDATE mastra_threads SET updatedAt = ?`). The pg and cloudflare-d1 adapters also bump it in their own `updateThread`. libsql is the odd one out.

Source: `stores/libsql/src/storage/domains/memory/index.ts:1218-1233`

```ts
const updatedThread = {
  ...thread,        // updatedAt here is the old value
  title,
  metadata: { ...thread.metadata, ...metadata },
};

await this.#client.execute({
  sql: `UPDATE ${TABLE_THREADS} SET title = ?, metadata = jsonb(?) WHERE id = ?`,
  args: [title, JSON.stringify(updatedThread.metadata), id],
});

return updatedThread;
```

The existing tests do not catch this because they only check that the returned object matches what comes back from a read. Both carry the same stale timestamp, so the test passes on broken code.

Heads up for whoever picks this up: mongodb and upstash have the same miss, so it may be worth fixing all three together.

## Steps To Reproduce

1. Create a thread with a libsql store.
2. Wait a moment, then call `updateThread` to change the title.
3. Read the thread back with `getThreadById`.
4. Check `updatedAt` and notice it still equals `createdAt`.

N/A, this is reproducible directly against the libsql adapter and is visible in the source linked above.

## Expected Behavior

After `updateThread`, the thread's `updatedAt` should move to the current time, both in the database and on the returned object. This is what the in-memory reference and the other adapters do. Without it, edited threads never resurface in `listThreads`, which sorts by `updatedAt` by default, and callers get a stale timestamp.

## Environment Information

```
Affects @mastra/libsql storage adapter. Not environment specific.
```

JigyasuRajput:
## Description

The libsql `updateThread` only updated a thread's title and metadata and left `updatedAt` alone. It was missing from the SQL SET clause, and the returned object just carried over the old value from the thread that was read in. mongodb and upstash had the same miss.

The effect was that editing a thread never moved its `updatedAt`, so edited threads did not resurface when listing threads sorted by most recently updated (the default), and callers got back a stale timestamp.

This was already the expected behavior everywhere else. The in-memory reference bumps `updatedAt` on update, the siblings in the same files (`saveMessages`, `updateMessages`) bump it, and the pg and cloudflare-d1 adapters already do it in their own `updateThread`. So this just brings libsql, mongodb and upstash in line.

What changed:

- libsql: added `updatedAt` to the SQL SET clause and set it on the returned object, using the same ISO string format the rest of the file uses.
- mongodb: added `updatedAt` to the `$set` and the returned object.
- upstash: set `updatedAt` before saving the thread.
- Strengthened the shared `updateThread` test. It was tautological before (it compared two objects that were both equally stale, so it passed on broken code). It now asserts that `updatedAt` actually moves forward, both on the returned object and after reading the thread back. Also removed a stray `console.log`.

I verified the before and after against libsql on Node 22: the strengthened test fails on the old code (`expected X to be greater than X`) and passes with the fix, and the full Threads suite stays green. mongodb and upstash were typechecked and formatted but not run locally since they need live services. The shared test covers them in CI.

## Related issue(s)

Fixes #17932

## Type of change

- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [x] Test update

## Checklist

## ELI5

When you edit a thread (change its title or details), the system should remember when you did that so it can show you the edited thread first when you ask for your most recently updated threads. This PR fixes a bug where three of the storage systems (libsql, mongodb, and upstash) weren't actually updating that "last edited time" when threads were modified, so edited threads would stay stuck at the bottom of the list.

### Changeset
- Bumped patch versions for `@mastra/mongodb`, `@mastra/upstash`, and `@mastra/libsql`
- Records fix where `updateThread` now refreshes a thread's `updatedAt` timestamp when thread title or metadata is edited, ensuring edited threads resurface correctly when listing threads sorted by most recently updated

### Test Improvements
The shared `updateThread` test in `stores/_test-utils/src/domains/memory/threads.ts` was strengthened:
- Now captures the original `updatedAt` timestamp before calling `updateThread`
- Asserts that the returned thread's `updatedAt` is later than the original
- After retrieving the thread from storage to verify persistence, confirms the retrieved thread's `updatedAt` also advanced
- Removed a stray `console.log` statement

### libsql Adapter
In `stores/libsql/src/storage/domains/memory/index.ts`, `updateThread` now:
- Creates a single `now` timestamp once
- Sets `updatedAt` on the returned `updatedThread` object
- Updates the database with `updatedAt = ?` in the SQL SET clause, passing the ISO string formatted timestamp as a parameter

### MongoDB Adapter
In `stores/mongodb/src/storage/domains/memory/index.ts`, `updateThread` now:
- Creates a single `now = new Date()` value
- Uses it for both the returned `updatedThread.updatedAt` field and the MongoDB `$set.updatedAt` update operation

### Upstash Adapter
In `stores/upstash/src/storage/domains/memory/index.ts`, `updateThread` now:
- Captures a single `now` timestamp at the start
- Sets `updatedAt` on the returned `updatedThread` object using this timestamp

These changes bring libsql, mongodb, and upstash in line with the in-memory reference implementation and the pg and cloudflare-d1 adapters, which were already correctly updating `updatedAt` on thread modifications.

### DIFF
--- a/.changeset/heavy-paws-smoke.md
+++ b/.changeset/heavy-paws-smoke.md
@@ -1,0 +1,7 @@
+---
+'@mastra/mongodb': patch
+'@mastra/upstash': patch
+'@mastra/libsql': patch
+---
+
+Fixed updateThread not refreshing a thread's updatedAt timestamp. Editing a thread's title or metadata now bumps updatedAt, so edited threads correctly resurface when listing threads sorted by most recently updated.
--- a/stores/_test-utils/src/domains/memory/threads.ts
+++ b/stores/_test-utils/src/domains/memory/threads.ts
@@ -143,6 +143,11 @@ export function createThreadsTest({ storage }: { storage: MastraStorage }) {
       const thread = createSampleThread();
       await memoryStorage.saveThread({ thread });
 
+      const originalUpdatedAtTime = new Date(thread.updatedAt).getTime();
+
+      // Wait a small amount to ensure a different timestamp
+      await new Promise(resolve => setTimeout(resolve, 10));
+
       const newMetadata = { newKey: 'newValue' };
       const updatedThread = await memoryStorage.updateThread({
         id: thread.id,
@@ -155,11 +160,12 @@ export function createThreadsTest({ storage }: { storage: MastraStorage }) {
         ...thread.metadata,
         ...newMetadata,
       });
+      expect(new Date(updatedThread.updatedAt).getTime()).toBeGreaterThan(originalUpdatedAtTime);
 
       // Verify persistence
       const retrievedThread = await memoryStorage.getThreadById({ threadId: thread.id });
-      console.log('retrievedThread', retrievedThread);
       expect(retrievedThread).toEqual(updatedThread);
+      expect(new Date(retrievedThread!.updatedAt).getTime()).toBeGreaterThan(originalUpdatedAtTime);
     });
 
     it('should return consistent timestamps from getThreadById and listThreads (issue #11496)', async () => {
--- a/stores/libsql/src/storage/domains/memory/index.ts
+++ b/stores/libsql/src/storage/domains/memory/index.ts
@@ -1225,6 +1225,7 @@ export class MemoryLibSQL extends MemoryStorage {
       });
     }
 
+    const now = new Date();
     const updatedThread = {
       ...thread,
       title,
@@ -1232,12 +1233,13 @@ export class MemoryLibSQL extends MemoryStorage {
         ...thread.metadata,
         ...metadata,
       },
+      updatedAt: now,
     };
 
     try {
       await this.#client.execute({
-        sql: `UPDATE ${TABLE_THREADS} SET title = ?, metadata = jsonb(?) WHERE id = ?`,
-        args: [title, JSON.stringify(updatedThread.metadata), id],
+        sql: `UPDATE ${TABLE_THREADS} SET title = ?, metadata = jsonb(?), updatedAt = ? WHERE id = ?`,
+        args: [title, JSON.stringify(updatedThread.metadata), now.toISOString(), id],
       });
 
       return updatedThread;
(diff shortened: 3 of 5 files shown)


wardpeet (approved):
LGTM, just checking if this isnt breaking anything

wardpeet (approved):
Reviewed the affected store changes and shared thread coverage. The libsql SQL update now persists `updatedAt` with the correct parameter ordering, MongoDB includes `updatedAt` in `$set`, and Upstash returns/saves the new timestamp consistently via `saveThread`. The shared thread test now verifies both the returned and fetched thread timestamps advance.

Local verification:
- `pnpm build:core`
- `pnpm_config_verify_deps_before_run=false pnpm --filter ./stores/libsql test -- src/storage/index.test.ts`
- `docker compose up -d --wait && pnpm_config_verify_deps_before_run=false pnpm exec vitest run src/storage/index.test.ts` in `stores/mongodb`
- `docker compose up -d && pnpm_config_verify_deps_before_run=false pnpm exec vitest run src/storage/index.test.ts` in `stores/upstash`

Current CI failures I checked appear unrelated to this PR: Lance vector cleanup, Pinecone missing `PINECONE_API_KEY`, and Vercel authorization/deploy statuses.

JigyasuRajput:
Thanks for the thorough review, really appreciate it 🙌🏻

PR #17940, opened 2026-06-15
fix(playground-ui): refine drawer overlay affordances

damien-schneider:
Updates the Drawer content bundle so it ships with a default top-right close button, while still letting consumers hide it with `showCloseButton={false}`. It also hides the floating side drag handle at desktop widths, where click-outside is the primary close affordance, while preserving the visible overlay backdrop and native right-swipe dismissal path.

Validated with:
- `pnpm --filter ./packages/playground-ui exec vitest run src/ds/components/Drawer/drawer.test.tsx`
- `pnpm build:playground-ui`
- `pnpm --filter ./packages/playground-ui typecheck`
- `pnpm --filter ./packages/playground-ui exec eslint src/ds/components/Drawer/drawer.tsx src/ds/components/Drawer/drawer.test.tsx`
- Storybook browser check on the visible overlay story


damien-schneider:
Closing this duplicate branch/PR. The unique overlay-affordance changes from this branch were moved into #17770 (codex/drawer-floating-variant).

PR #17942, opened 2026-06-15
fix(workflow):workflow-cancel-interrupt-sleep

### LINKED ISSUE #17908: [BUG] Workflow `cancel()` does not interrupt `sleep()` / `sleepUntil()` — the run keeps executing and overwrites the canceled status

Abuhaithem:
Fixes:  #17908
Makes `Run.cancel()` interrupt an in-flight `.sleep(ms)` / `.sleepUntil(date)` immediately, instead of waiting for the full duration and then overwriting the canceled status.

Fixes the workflow-cancel-during-sleep bug #17908 .

## Why

The default execution engine's sleep primitives were plain `setTimeout` promises that never observed the run's abort signal. A run canceled mid-sleep would:

1. keep the Node timer alive for the full duration (`.sleepUntil(+24h)` held the run for 24h after cancel);
2. wake up and overwrite the persisted `'canceled'` status with `'running'`; and
3. publish a sleep-step `success` event.

So observers saw the run flip `canceled` → `running` → `canceled`, and cancellation didn't actually take effect until the timer fired.

## How

- **`workflows/utils.ts`** — new `abortableSleep(ms, signal)` helper. It races the timer against the abort signal, clears the timer on abort, and `unref`s it so a long sleep never keeps the Node event loop alive on its own.
- **`workflows/default.ts`** — `executeSleepDuration` / `executeSleepUntilDate` now accept an optional `AbortSignal` and use `abortableSleep`. The parameter is optional, so existing engine overrides (e.g. Inngest) remain source-compatible.
- **`workflows/handlers/sleep.ts`** — threads `abortController?.signal` into those engine calls.
- **`workflows/handlers/entry.ts`** — after the wait, if the signal is aborted, the handler skips the `'running'` persist and the sleep-step `success` / `finish` events and settles the result as `canceled`. The existing tail handler then persists `canceled` and publishes `workflow-canceled`. (The per-step result is left as its prior `'waiting'` value, since `'canceled'` is not a valid `StepResult` status — cancellation is tracked on the run, matching existing behavior.)

### Before / After

```ts
const run = await workflow.createRun();
const result = run.start({ inputData: {} });
setTimeout(() => run.cancel(), 1000); // during .sleep(60_000)

await result;
// Before: resolves after ~60s; status briefly flips canceled → running → canceled
// After:  resolves promptly with status 'canceled'; sleep step never reports success
```

## Tests

- `packages/core/src/workflows/cancel-sleep.test.ts` — new regression test covering both `.sleep()` and `.sleepUntil()`: asserts the wait is interrupted (<5s), the downstream step never runs, the result is `canceled`, and storage stays `canceled` (not flipped to `running`/`success`). Fails on `main` (times out), passes with this fix.
- `packages/core/src/workflows/utils.test.ts` — unit tests for `abortableSleep` (resolves on duration, resolves early on abort / already-aborted signal, handles non-positive durations).

### Checks run

- `vitest run` for the new/affected suites (regression + unit + `default.test.ts` + evented sleep tests): all pass.
- `tsc --noEmit -p tsconfig.build.json`: clean.
- `eslint` on changed files: clean.
- Changeset added: `.changeset/fix-workflow-cancel-interrupts-sleep.md` (`@mastra/core`: patch).

## Affected package

`@mastra/core`

## ELI5

When you tell a workflow to stop while it's sleeping, it now actually stops right away instead of waiting for the full sleep duration. Before, if a workflow was supposed to sleep for 60 seconds and you canceled it at 1 second, it would still make you wait the full 60 seconds — and even worse, it would "forget" that it was canceled after the sleep finished.

## Problem

The default execution engine's sleep functions (`setTimeout`-based promises) didn't listen to cancellation signals. This caused three issues when `Run.cancel()` was called during a sleep:

1. **Long waits persisted**: Long-duration sleeps (e.g., `.sleepUntil(+24 hours)`) kept Node.js timers alive for the entire duration even after cancellation, preventing clean process shutdown
2. **Status corruption**: After the timer fired, the run would wake up and overwrite the persisted `'canceled'` status with `'running'`
3. **Phantom sleep success**: The sleep step was recorded as `success` before the run noticed it was aborted

## Solution

The fix introduces an abort-aware sleep mechanism that races the timer against a cancellation signal:

- **`abortableSleep` utility** (`workflows/utils.ts`): A new Promise-based sleep helper that respects `AbortSignal`, clears the timer on abort, and uses `unref()` to prevent long sleeps from keeping the event loop alive
- **Signal threading**: `executeSleepDuration` and `executeSleepUntilDate` methods in `DefaultExecutionEngine` now accept an optional `AbortSignal` parameter (backward compatible) and use `abortableSleep` instead of plain `setTimeout`
- **Signal propagation**: The abort signal is threaded through `workflows/handlers/sleep.ts` into engine calls
- **Status preservation**: When a signal is aborted in `workflows/handlers/entry.ts`, the run skips the `'running'` persist and sleep-step success events, settling directly to `'canceled'`

## Changes

**`workflows/utils.ts`**: Added `abortableSleep(ms, signal?)` function that immediately resolves if `ms <= 0` or signal is already aborted; otherwise, creates a timeout that resolves early when the signal aborts, with `unref()` to prevent event loop lock.

**`workflows/default.ts`**: Updated `executeSleepDuration` and `executeSleepUntilDate` to accept optional `AbortSignal` parameter and use `abortableSleep` instead of plain `setTimeout` promises.

**`workflows/handlers/sleep.ts`**: Passes `abortController?.signal` into engine sleep methods.

**`workflows/handlers/entry.ts`**: After sleep completes, branches on `abortController?.signal?.aborted`: if aborted, skips the `'running'` persist and success/result/finish events, leaving the step in `waiting` state and settling the run to `canceled`; if not aborted, persists normally and publishes step result events.

**Test coverage**: 
- New regression test suite (`workflows/cancel-sleep.test.ts`) with tests for both `.sleep()` and `.sleepUntil()` that verify cancellation interrupts within 5 seconds, downstream steps don't execute, and run status remains `canceled` in both in-memory and persisted state
- New unit tests in `workflows/utils.test.ts` covering `abortableSleep` behavior for normal completion, pre-aborted signals, mid-sleep abort, and non-positive durations

## Impact

- Canceling a workflow during any sleep operation now interrupts immediately instead of forcing the process to wait
- Long sleeps no longer keep Node.js timers alive after cancellation
- Run status transitions to and remains `'canceled'` without intermediate `'running'` state oscillations
- Downstream workflow steps scheduled after a canceled sleep do not execute
- Behavior is consistent both in-process and in persisted storage
- All changes are backward compatible; the `AbortSignal` parameter is optional


intojhanurag:
## 🟠 Maintainer's Triage Note

**Current Phase:** Awaiting Approval
**Next Step:** Await final approval

**Triage:**

- Type: bug — Fixes workflow cancellation not interrupting `.sleep()` / `.sleepUntil()` and allowing canceled runs to resume execution.
- Maintainer read: PR #17942 directly routes open issue #17908, which reports that canceling a workflow during sleep leaves the timer alive, briefly restores `running`, publishes sleep success, and may continue to the next step. The changed area is core workflow sleep/cancel handling with regression tests included.
- Route: Review PR #17942
- Severity: 🟠 high — Workflow cancellation can be delayed and persisted state can be overwritten on a core workflow path.

**Review:**

- Status: complete
- Findings: Default engine fix is coherent: abortable sleep clears timers, handlers pass the abort signal, and entry handling avoids repersisting running/success after cancellation. Tests cover `.sleep()` and `.sleepUntil()`. Main risk is that the Inngest execution engine overrides the same sleep methods and appears to ignore the new abort signal, so Inngest-backed workflow sleeps may remain non-interruptible unless intentionally out of scope.
- Confidence: 3/5
- Follow-up: Confirm whether Inngest-backed workflows are in scope for this fix; if yes, update `workflows/inngest/src/execution-engine.ts` or add an explicit follow-up before approval.

**Approve:**

- Status: waiting for final approval
- Final approver: @abhiaiyer91
- Notes: Final approval requested from @abhiaiyer91 based on CODEOWNERS for `packages/core/src/workflows/` and recent workflow-area context. Approver should decide whether the Inngest sleep override gap must be handled before approval or explicitly tracked as out of scope/follow-up.

abhiaiyer91:

We're currently testing out our agentic factory, and it implemented a fix for the linked issue (#17908), which has since been closed. So I'm closing this PR out as no longer needed.

PR #17985, opened 2026-06-15
fix(agent-browser): install Chromium Linux dependencies before launch

arctic-char:
## Summary

Fixes Agent Browser server deployments where Chromium exists but fails to start because the Linux runtime is missing Playwright system libraries such as `libglib-2.0.so.0`.

- Adds `playwright-chromium` as a direct `@mastra/agent-browser` dependency so the framework owns the Chromium/Playwright CLI dependency instead of requiring each template to add it.
- Runs `playwright-chromium install-deps chromium` before local browser launch in both shared-scope and thread-scope AgentBrowser paths.
- Skips dependency installation for remote CDP/browser sessions, non-Linux runtimes, non-root runtimes, or when `BROWSER_SKIP_INSTALL_DEPS=true`.
- Adds `installLinuxDependencies: false` as an opt-out for images that already include the required system packages.

## Test plan

- `pnpm --filter ./browser/agent-browser exec vitest run src/__tests__/agent-browser.test.ts src/__tests__/thread-manager.test.ts -t "Playwright Linux dependencies"`
- `pnpm --filter ./packages/core build:lib`
- `pnpm --filter ./browser/agent-browser build`
- `pnpm exec prettier --check browser/agent-browser/package.json browser/agent-browser/src/agent-browser.ts browser/agent-browser/src/thread-manager.ts browser/agent-browser/src/types.ts browser/agent-browser/src/playwright-deps.ts browser/agent-browser/src/__tests__/agent-browser.test.ts browser/agent-browser/src/__tests__/thread-manager.test.ts .changeset/new-candles-ring.md`

Note: running the full two-file Agent Browser test selection currently exposes an unrelated existing recording-tools failure: `createBrowserRecordingTools is not a function`. The focused new regression tests pass.

## ELI5

If your Linux server is missing some system libraries, Chromium can’t start. This PR makes Agent Browser ensure Playwright’s Chromium “install-deps” runs right before launching local browsers (when it’s safe to do so), and skips that step for remote/unsupported environments.

## Overview

Agent Browser deployments could fail on Linux because Chromium launched by Playwright lacked required system libraries (e.g., `libglib-2.0.so.0`). This PR makes the framework handle Playwright/Chromium dependency setup itself, so templates don’t need to manage it.

## Changes

- **Own the Playwright Chromium CLI dependency**
  - Added `playwright-chromium@^1.60.0` to `@mastra/agent-browser` dependencies.
  - Also added it to `peerDependencies` as an optional peer (so apps can still control inclusion while Agent Browser uses it when available).
  - Updated workspace config to allow builds for `playwright-chromium`.

- **Pre-launch Linux dependency installation (best-effort, safe to skip)**
  - Added `installLinuxDependencies` and `installLinuxDependenciesTimeoutMs` to Agent Browser config.
  - Implemented `installPlaywrightLinuxDeps({ cdpUrl, enabled, timeoutMs })`:
    - Resolves the local Playwright CLI (`playwright-chromium` or `playwright`) and runs `cli.js install-deps chromium`.
    - Runs **only on Linux** and **only as root**.
    - **Skips** when:
      - Connecting via a **remote CDP/browser session** (`cdpUrl` present)
      - `BROWSER_SKIP_INSTALL_DEPS=true`
      - `enabled` is false (via `installLinuxDependencies: false`)
      - Non-Linux runtime
      - Non-root user contexts
    - Enforces a timeout (with `SIGTERM` on expiry) and uses a cached single-flight promise so concurrent launches don’t duplicate installs; failures clear the cache for retry.
    - Includes a test helper to reset the cached state.

- **Wired into both launch paths**
  - Shared-scope: installs before `sharedManager.launch`.
  - Thread-scope: installs during session creation before the thread-scoped `BrowserManager` launch.

- **Release publishing**
  - Added a changeset to publish `@mastra/agent-browser` with a patch bump and note the Playwright Chromium install-deps behavior.

## Testing

- Updated unit tests to mock the Linux dependency installer and verify call ordering (install happens before launch).
  - Shared scope install behavior
  - Thread scope install behavior
  - Disabling via `installLinuxDependencies: false`
- Note: the full test suite still has an unrelated pre-existing failure involving `createBrowserRecordingTools`, but the new regression tests pass.


abhiaiyer91 (approved):

browser/agent-browser/package.json
```diff
@@ -30,7 +30,8 @@
   },
   "license": "Apache-2.0",
   "dependencies": {
-    "agent-browser": "0.19.0"
+    "agent-browser": "0.19.0",
+    "playwright-chromium": "^1.60.0"
```
wardpeet: we should make these optional peerdeps

wardpeet (changes requested):

PR #288, opened 2024-09-10
add scopes and filter out invalid scope

Joshuafolorunsho:

PR #17990, opened 2026-06-15
fix(harness,libsql): keep agent loop alive after tool suspension

abhiaiyer91:
## Summary

Two fixes for the reported bug where **MastraCode stops mid-task right after an `ask_user` (or other interactive-tool) suspension**, even though the agent says it will keep working.

### 1. `harness`: resume was capped at the agent's default step budget
`handleToolResume` called `agent.resumeStream()` **without `maxSteps`**. `resumeStream` merges per-call options over the agent's `getDefaultOptions()`, whose `maxSteps` is small (~5). So a resumed run took only a few LLM steps and ended with `reason:"complete"` — looking like the agent quit mid-task.

The initial stream sets `maxSteps: 1000`; the resume didn't. The fix extracts a single `buildSharedRunOptions()` helper (`maxSteps`, `savePerStep`, `modelSettings`, fable-5 fallback `providerOptions`, `requireToolApproval`) used by **both** the initial stream and `resumeStream`, so the two paths can't drift again. A `HARNESS_MAX_STEPS` constant documents why.

### 2. `libsql`: SQLITE_BUSY storms under the evented engine
Concurrent writers on a single-connection LibSQL client triggered `SQLITE_BUSY` retry-exhaustion that could stall runs. Bump `@libsql/client` to `^0.17.4` and thread a configurable connection timeout into `createClient` so `PRAGMA busy_timeout` **survives connections created after `transaction()`** (Turso `libsql-client-ts` #288/#345). SQLite now waits for locks instead of immediately throwing.

## Tests
- `harness-tool-suspension.test.ts`: new regression test asserting the resume forwards the full step budget (`maxSteps: 1000`, `savePerStep: false`); existing yolo-mode resume assertion still holds.
- `stores/libsql/.../busy-timeout.test.ts`: covers default/custom timeout and post-`transaction()` survival.

## Verification
- harness suspension + ask-user suites: 12 passed
- libsql busy-timeout + write-lock suites: 8 passed

## Notes
All `[suspend-trace]` debug instrumentation used during investigation was stripped from this branch; only the two fixes + tests + dep bumps remain.

## ELI5

This PR fixes two ways an AI agent could “get stuck.” After the agent pauses to ask the user something, it was restarting with too few “work steps” left—so it could look like it quit early. Also, the database layer could error when multiple writes happened at once; the fix makes SQLite wait for locks instead of failing immediately.

---

## Issue 1: Harness step budget lost during tool resume

**Problem:**  
When a MastraCode agent hit an interactive tool suspension (e.g., `ask_user`), it resumed with the wrong (too-small) step budget. That made the resumed run stop after only a handful of LLM steps, appearing like the agent quit prematurely.

**Solution:**  
- Added `HARNESS_MAX_STEPS` to document the intended 1000-step budget.
- Introduced `buildSharedRunOptions()` to centralize shared harness run configuration (including `maxSteps`, `savePerStep`, `modelSettings`, provider option fallback, and `requireToolApproval`).
- Refactored initial streaming option construction and tool-resume handling to both use the exact same shared run options when calling `agent.resumeStream()`.

**Tests:**  
- Added `harness-tool-suspension.test.ts` to verify that after a tool suspension, the harness resumes with `resumeOptions.maxSteps === 1000` (and `resumeOptions.savePerStep === false`), preventing early termination.

---

## Issue 2: LibSQL `SQLITE_BUSY` contention under concurrent writes

**Problem:**  
Concurrent writes using a single-connection LibSQL client could hit `SQLITE_BUSY` retry-exhaustion errors, stalling execution.

**Solution:**  
- Bumped `@libsql/client` to `^0.17.4`.
- Added optional `connectionTimeoutMs` configuration (with exported `DEFAULT_CONNECTION_TIMEOUT_MS = 5000`) to apply SQLite `busy_timeout`.
- Updated `resolveClient()` to pass `{ timeout }` to `createClient()` for local `file:` / `:memory:` URLs so the setting applies to connections created after `transaction()`.
- Updated local initialization pragmas to set `busy_timeout` from `connectionTimeoutMs` (instead of a hardcoded value).

**Tests:**  
- Added `stores/libsql/.../busy-timeout.test.ts` to cover default vs custom timeout and that `busy_timeout` persists for connections created after `client.transaction()`.

---

## Files modified

| File | Changes |
|---|---|
| `packages/core/src/harness/harness.ts` | Added `HARNESS_MAX_STEPS`; added `buildSharedRunOptions()`; ensured both initial streaming and `resumeStream` use identical shared run options |
| `packages/core/src/harness/__tests__/harness-tool-suspension.test.ts` | New regression test asserting resumed runs keep `maxSteps: 1000` and `savePerStep: false` |
| `stores/libsql/package.json` | Updated `@libsql/client` to `^0.17.4` |
| `stores/libsql/src/storage/db/index.ts` | Added `connectionTimeoutMs`, exported `DEFAULT_CONNECTION_TIMEOUT_MS`, and applied `{ timeout }` for local connections |
| `stores/libsql/src/storage/index.ts` | Added `connectionTimeoutMs` to store config and updated `busy_timeout` pragmas |
| `stores/libsql/src/storage/busy-timeout.test.ts` | New tests for default/custom timeout and post-`transaction()` persistence |
| `pnpm-workspace.yaml` | Added release-gate exclusions for pinned `@libsql/client@0.17.4` and `@libsql/core@0.17.4` |
| `.changeset/keep-agent-loop-alive-after-suspension.md` | Documented the harness resume + LibSQL `busy_timeout` fixes |

### DIFF
--- a/.changeset/keep-agent-loop-alive-after-suspension.md
+++ b/.changeset/keep-agent-loop-alive-after-suspension.md
@@ -1,0 +1,11 @@
+---
+"@mastra/core": patch
+"@mastra/libsql": patch
+---
+
+Keep the agent loop alive after a tool suspension (e.g. `ask_user`, `request_access`).
+
+Two related fixes that stop the agent from appearing to quit mid-task right after a suspended tool resumes:
+
+- **Harness resume step budget**: `resumeStream()` was called without a step budget, so a resumed run merged over the agent's small default `maxSteps` and stopped after a single step. The harness now threads a shared set of run options (`maxSteps`, `savePerStep`, `modelSettings`, etc.) into both the initial stream and the resume via one helper, so the two paths can no longer drift.
+- **LibSQL `busy_timeout` survival**: bump `@libsql/client` to `^0.17.4` and add a configurable connection timeout so `PRAGMA busy_timeout` survives connections created after `transaction()` (libsql-client-ts#288/#345). This reduces `SQLITE_BUSY` retry-exhaustion stalls under the evented engine's concurrent writes.
--- a/packages/core/src/harness/__tests__/harness-tool-suspension.test.ts
+++ b/packages/core/src/harness/__tests__/harness-tool-suspension.test.ts
@@ -348,4 +348,72 @@ describe('Harness: tool suspension and resumption', () => {
     // Yolo mode should disable tool approval gating on resume, matching sendMessage's behavior
     expect(resumeOptions.requireToolApproval).toBe(false);
   });
+
+  it('should forward the full run budget (maxSteps) to resumeStream so the resumed run does not stop mid-task', async () => {
+    // Regression: resumeStream previously omitted maxSteps, so the resumed run
+    // merged over the agent's small default budget and ended with reason
+    // "complete" after a few steps — the agent stopped mid-task after ask_user.
+    const confirmTool = createTool({
+      id: 'confirm-action',
+      description: 'Confirms an action with the user',
+      inputSchema: z.object({ action: z.string() }),
+      execute: async (input: { action: string }, context?: any) => {
+        // Resume-aware: continue instead of re-suspending once resumeData arrives,
+        // so the resumed run can actually complete.
+        const resumeData = context?.agent?.resumeData ?? context?.workflow?.resumeData ?? context?.resumeData;
+        if (resumeData) {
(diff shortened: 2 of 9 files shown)


TylerBarnes (approved):
looks good! just needs a changeset

PR #18014, opened 2026-06-16
fix(auth-workos): fix OAuth PKCE regression in handleCallback

rphansen91:
## Summary

Fixes a regression in the WorkOS auth provider where OAuth callback was failing with "PKCE verifier cookie missing" error.

## Problem

The AuthKit `handleCallback()` method requires PKCE cookies to be set during the login redirect, but the `ISSOProvider` interface separates `getLoginUrl()` and `handleCallback()` across different HTTP requests. Since cookies aren't shared between these requests, the PKCE verifier cookie was never found.

This regression was introduced in commit `6e4d4f5cf3` (Jan 2026) when someone changed the working `authenticateWithCode()` implementation to use AuthKit's `handleCallback()` with dummy Request/Response objects.

## Solution

Reverted to using the WorkOS SDK's `authenticateWithCode()` directly, which performs server-side code exchange without needing PKCE cookies. Server-side code exchange is secure because the server can safely store the client secret.

PKCE is primarily designed for public clients (mobile apps, SPAs) that can't securely store secrets. Server-side applications don't need PKCE when they have a client secret.

## Changes

1. **fix(auth-workos)**: Use `authenticateWithCode()` instead of AuthKit `handleCallback()`
2. **feat(examples)**: Add dual-auth example demonstrating Studio SSO + API JWT pattern
3. **docs(auth)**: Add dual auth system documentation and demo script

## Testing

- [x] All 53 auth-workos tests pass
- [x] Verified OAuth login flow works end-to-end in browser
- [x] Dual-auth example dev server starts and serves correctly

## Related

- Fixes broken WorkOS OAuth login that was sitting on main for ~3 months
- Related to dual auth system (PR #17722) which was merged to main

## ELI5

This PR fixes WorkOS “sign in” by making the server handle the OAuth callback in a way that doesn’t depend on PKCE cookies (which were getting lost between requests). It also adds a dual-auth example: Studio users sign in with WorkOS SSO, while API customers use JWTs.

### What broke
OAuth callbacks were failing with **“PKCE verifier cookie missing”** after the implementation switched to AuthKit’s `handleCallback()` using dummy `Request`/`Response` objects.

### Why it broke
PKCE requires a verifier cookie created during the login redirect, but the `ISSOProvider` interface separates `getLoginUrl()` and `handleCallback()` across different requests—so the cookie wasn’t available during callback handling.

### What changed
In `auth/workos/src/auth-provider.ts`, `MastraAuthWorkos.handleCallback()` was changed to:
- Exchange the OAuth `code` directly via `workos.userManagement.authenticateWithCode()` (server-side; no PKCE cookies needed)
- Map the SDK response into the provider’s `WorkOSUser` (including `organizationId` and `workosId`)
- Build session data (access/refresh tokens, user, `organizationId`, `impersonator`)
- Encrypt/seal the session data with `sessionEncryption.sealData()` using `this.config.cookiePassword`
- Manually set the `Set-Cookie` value(s) with consistent attributes (`Path=/`, `HttpOnly`, `SameSite` defaulting to `Lax`, and `Secure` only when `NODE_ENV=production`)
- Omit `cookies` entirely when `cookiePassword` is not configured

### Example app
Added `examples/dual-auth/` with:
- **Studio (`studio.auth`)**: `MastraAuthWorkos` for WorkOS SSO (optionally using `WORKOS_SSO_PROVIDER`)
- **API (`server.auth`)**: `authenticateToken` that verifies an HMAC-SHA256 JWT using `JWT_SECRET`, including constant-time signature comparison and `exp` expiry checks

### Documentation
- `docs/dual-auth-demo.md`: end-to-end demo guide (env vars, Studio login flow, API curl examples, permission differences, and security notes)
- `docs/src/content/en/docs/server/auth/dual-auth.mdx`: explains dual auth configuration and routing using `x-mastra-client-type: studio`, including that the header only selects the provider and **cannot bypass authentication** without valid `studio.auth` credentials

## Testing / verification
- All **53 auth-workos tests pass**
- WorkOS OAuth login flow verified end-to-end in the browser
- Dual-auth example dev server starts and serves correctly

### DIFF
--- a/auth/workos/src/auth-provider.ts
+++ b/auth/workos/src/auth-provider.ts
@@ -520,31 +520,61 @@ export class MastraAuthWorkos
   /**
    * Handle the OAuth callback from WorkOS.
    *
-   * Uses AuthKit's handleCallback for proper session creation.
+   * Uses WorkOS SDK's authenticateWithCode directly instead of AuthKit's handleCallback.
+   * AuthKit's handleCallback requires PKCE cookies that must be set during getLoginUrl()
+   * and read during handleCallback(), but our ISSOProvider interface separates these
+   * calls across different requests without cookie propagation.
+   *
+   * This approach was the original implementation before commit 6e4d4f5cf3 introduced
+   * a regression by switching to AuthKit's handleCallback with dummy Request/Response
+   * objects that couldn't provide the required PKCE cookies.
    */
   async handleCallback(code: string, _state: string): Promise<SSOCallbackResult<EEUser>> {
-    // Use AuthService's handleCallback for session creation
-    const result = await this.authService.handleCallback(
-      new Request('http://localhost'), // Dummy request, not used
-      new Response(), // Dummy response to get headers
-      { code, state: _state },
-    );
+    // Use WorkOS SDK directly to exchange code for tokens (server-side, no PKCE required)
+    const authResponse = await this.workos.userManagement.authenticateWithCode({
+      clientId: this.clientId,
+      code,
+    });
 
     const user: WorkOSUser = {
-      ...mapWorkOSUserToEEUser(result.authResponse.user),
-      workosId: result.authResponse.user.id,
-      organizationId: result.authResponse.organizationId,
+      ...mapWorkOSUserToEEUser(authResponse.user),
+      workosId: authResponse.user.id,
+      organizationId: authResponse.organizationId,
     };
 
-    // Extract session cookie from headers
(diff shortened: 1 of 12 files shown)


abhiaiyer91 (approved):

rphansen91:
Closing in favor of #18035 which contains only the PKCE fix with a proper changeset. The dual auth docs/example will be a separate PR.

PR #18016, opened 2026-06-16
fix(core): make the native Agent goal mechanism robust (tools, waiting, budget, judge-failure)

abhiaiyer91:
## Summary

Makes the native Agent goal mechanism robust, restoring behavior that regressed when goal handling moved into core (PR #17889). Addresses Tyler's five review recommendations plus an infinite-loop fix found while verifying them.

## What changed

**1. Tool-capable judges** — Scorer judge configs accept an optional `tools` set the internal judge agent can call before producing its score. The default goal scorer forwards `goal.tools`, so it can verify the agent's work against reality instead of grading prose alone. MastraCode wires its read-only workspace tools (`view`, `search_content`, `find_files`, `file_stat`, `lsp_inspect`) into `goal.tools` via a per-request resolver.

**2. Tri-state waiting** — The default goal scorer emits `done`/`continue`/`waiting`. A `waiting` decision — only when the goal text explicitly asks to stop for the user — parks the objective as `paused` instead of looping. The judge still self-resolves "wait for the judge" instructions.

**3. Budget-exhaustion pause** — Reaching `maxRuns` without completing now parks the objective as `paused` with a clear reason, resumable by raising `maxRuns` and reactivating (`updateObjectiveOptions({ maxRuns, status: 'active' })`), instead of silently leaving it `active` but stuck.

**4. Judge-failure pause (fixes an infinite loop)** — *Any* failure while evaluating the goal now pauses the objective and stops the loop, surfacing the cause. This includes failures during **judge-model and `goal.tools` resolution**, not just the scorer run — that resolution path was previously unguarded, so a judge error (e.g. a gateway "Bad Request") escaped the goal step after the model had already produced output but before `isContinued` was set to `false`, causing the loop to re-run the model against the broken judge every turn until the user hit esc.

**5. Structured-output retry** — `tryGenerateWithJsonFallback` now retries with `jsonPromptInjection` when the judge resolves *without* a parseable object, not only when it throws. Previously an empty/malformed object crashed downstream. Mirrors the streaming path.

The goal evaluation chunk now carries `pausedReason` and `judgeFailed`, and `reason` falls back to `pausedReason` so consumers (e.g. the TUI judge display) always show a cause rather than a blank "paused".

## Testing

- `@mastra/core`: 129 tests across goal, scorer, goal-step, validation, utils, and evals/base — all passing.
- `mastracode`: judge-display + goal-judge-tools — all passing.
- Both packages typecheck clean; core builds clean.
- Regression tests include: stops on the first judge failure under a large budget (no march to `maxRuns`), and pauses on a **resolution-time** throw without escaping the step.

## Follow-up (not in this PR)

**Transient judge-failure retry.** This PR guarantees a failing judge never loops forever — it pauses cleanly with the cause. It does **not** add bounded retry + backoff for *transient* failures (429/5xx/network), with immediate pause on deterministic errors (400/401/403). That's resilience polish on top of the correctness fix and worth doing deliberately; deferred intentionally.

## ELI5
This PR improves how the system double-checks whether the agent did the “goal” correctly—using safe, read-only workspace facts—without getting stuck. If it needs to wait, runs out of tries, or the checker/judge crashes, the goal is paused with a clear explanation so it can resume later.

## Overview
The native Agent goal mechanism was restored/strengthened to make judging more robust and control-flow safer. It focuses on five areas:

### Tool-capable judges
- Judge/scorer configurations now accept an optional `tools` set.
- The default goal scorer forwards `goal.tools` into the internal judge agent so the judge can verify claims against real workspace info (not just text).
- MastraCode provides per-request read-only workspace tools via a new `getGoalJudgeTools` resolver, filtered by an allowlist (returns `undefined` if the workspace can’t be resolved).
- Only the allowlisted inspection tools are exposed to judges.

### Tri-state waiting (and correct pausing)
- The default goal scorer now returns `done | continue | waiting`.
- `waiting` is used when the goal text explicitly asks to stop for the user; when that happens, the objective is parked as `paused` instead of continuing the loop.
- “Wait for the judge” style instructions are handled by judge logic without necessarily causing an objective pause.

### Budget-exhaustion pause (resumable)
- If `maxRuns` is reached without completing, the objective is set to `paused` with a clear budget-related `pausedReason`.
- The objective can be resumed by raising `maxRuns` and reactivating via `updateObjectiveOptions({ maxRuns, status: 'active' })`.

### Judge-failure pause (prevents infinite loops)
- Any failure during goal evaluation now pauses the objective and stops the loop.
- This includes failures during judge-model execution and during `goal.tools` resolution, preventing repeated re-runs against a broken judge setup.

### Structured-output retry improvements
- `tryGenerateWithJsonFallback` now retries with `jsonPromptInjection` when the first attempt resolves but produces no parseable structured object (not only when it throws).

### Richer evaluation payload
- Goal evaluation chunks now include:
  - `pausedReason` (only when paused)
  - `judgeFailed` (true when the scorer/judge itself errored)
  - `reason`, which falls back to `pausedReason` for display.

## Additional robustness notes (implementation + control)
- Judge evaluation was switched from `agent.generate` to `agent.stream` for compatibility with stream-only models (judges always need the final structured decision object, not user-visible streaming).
- MastraCode goal TUI commands now refresh the status line immediately:
  - `/goal pause` and `/goal resume` call `ctx.updateStatusLine()` after saving state.
  - `/goal clear` aborts in-flight agent turns (when the harness is running/pending) and clears in-flight ask-user/inline-question state, setting `userInitiatedAbort`.

## Changes by file (high level)
- **Core goal judging/scoring**
  - `packages/core/src/agent/goal/objective.ts`: adds `GOAL_SCORE_WAITING` and tightens the default judge prompt to require exactly one of `done/continue/waiting`.
  - `packages/core/src/agent/goal/scorer.ts`: extends scorer to accept optional `tools`, updates judge output to `{ decision, reason }`, implements tri-state mapping, and augments judge instructions when tools exist.
  - `packages/core/src/agent/goal/index.ts`: exports `GOAL_SCORE_WAITING`.
- **Goal step control flow**
  - `packages/core/src/loop/workflows/agentic-execution/goal-step.ts`: handles waiting/judge failure/budget exhaustion via `pausedReason` + `judgeFailed`, stops looping on these conditions, and includes pause details in emitted goal chunks/transcript feedback.
- **Types/storage**
  - `packages/core/src/storage/domains/thread-state/base.ts`: adds `pausedReason?: string` to `GoalObjectiveRecord`.
  - `packages/core/src/stream/types.ts`: adds `pausedReason?: string` and `judgeFailed?: boolean` to `GoalEvaluationPayload`, and documents `reason` fallback.
- **MastraCode integration**
  - `mastracode/src/agents/workspace.ts`: adds `getGoalJudgeTools` with read-only allowlisting and `undefined` on workspace resolution failure.
  - `mastracode/src/index.ts`: wires `goal.tools` into the native goal judge configuration.
  - `mastracode/src/tui/commands/goal.ts`: updates status line immediately; `/goal clear` aborts in-flight turns and clears in-flight user/inline question state.
- **Eval infrastructure + structured-output retry**
  - `packages/core/src/evals/base.ts`: introduces `ScorerJudgeConfig` with optional `tools`, ensures step/judge tools resolution works, and uses the JSON fallback path with streaming-capable retries.
  - `packages/core/src/agent/utils.ts`: ensures the fallback retry triggers even when `agent.generate/stream` resolves without a structured `result.object`.

## Testing
129 passing tests across `@mastra/core` and `mastracode`.


packages/core/src/storage/domains/thread-state/base.ts
```diff
@@ -38,6 +38,12 @@ export interface GoalObjectiveRecord {
   judgeModelId?: string;
   /** Extra judge guidance. Falls back to agent `goal.prompt` (default = built-in goal judge prompt). */
   prompt?: string;
+  /**
+   * Why the objective is `paused`, when applicable. Distinguishes the pause
+   * causes for the UI / resume flow: an explicit user checkpoint requested by
+   * the goal ("waiting"), or a judge/scorer failure. Unset for `active`/`done`.
+   */
+  pausedReason?: string;
```
wardpeet, line 47:
```suggestion
  statusReason?: string;
```
Should it be more generic, we could fill this when an failure is happenign as well

TylerBarnes: I think we should keep it as-is, looks like it's already used for failure, but this is really the reason that it paused. statusReason could be taken as any status (like continue or waiting for user input)

PR #18018, opened 2026-06-16
fix(schema-compat): convert null to undefined for optional/default fields in non-OpenAI providers

### LINKED ISSUE #17655: [BUG] Tool-call & structured-output validation fails for `.default()` / `.optional()` fields when the model returns `null` (most providers)

nuthalapativarun:
## What changed

When a model returns `null` for a field defined as `.optional()` or `.default(value)`, Zod validation now passes correctly for Google, Anthropic, DeepSeek, and Meta providers.

**Before:** Validation failed with a Zod error because `null` was passed directly to schemas like `z.string().optional()` (which expects `undefined`, not `null`) or `z.number().default(1)`.

**After:** `null` is converted to `undefined` for optional fields (allowing Zod's default to apply for `.default()` schemas), matching the behavior that already existed in the OpenAI compat layer.

Nullable fields (explicitly `.nullable()`) continue to accept `null` unchanged.

## Root cause

`processToAISDKSchema` in the non-OpenAI providers used `io: 'output'` when building the traversal schema, which strips default markers and makes default fields appear required. The `#traverse` method also had no logic to convert `null` → `undefined` for fields absent from the JSON Schema `required` array.

**Fix:** Changed `processToAISDKSchema` to use `io: 'input'` (so optional/default fields correctly appear non-required in the schema) and added a `required` array check in each provider's `#traverse` — if a field is absent from `required` and the value is `null`, it is set to `undefined`. Zod then applies `.default()` automatically when it receives `undefined`.

## Tests

Added three tests to `createSuite()` (the universal test suite run for all providers):
- `null` → `undefined` for `.optional()` fields
- `null` → default value for `.default()` fields
- `null` preserved for `.nullable()` fields

All 812 unit tests pass (only the 4 e2e tests that require live API keys are excluded).

Fixes #17655

## Checklist

- [x] Linked issue
- [x] Tests added
- [x] No docs update needed (internal validation logic)
- [x] Changeset added

## ELI5

This PR fixes a bug where certain AI providers (Google, Anthropic, DeepSeek, and Meta) incorrectly rejected data when a model returned `null` for fields marked as optional or having default values. The fix teaches these providers to convert `null` to `undefined` so the validation rules can work correctly, just like the OpenAI provider already does.

## Overview

This PR addresses null handling for optional and default fields across non-OpenAI providers (Google, Anthropic, DeepSeek, and Meta). When models return `null` for fields defined with `.optional()` or `.default(value)`, validation now passes by converting `null` to `undefined`.

## Problem

Previously, validation failed because:
1. `null` was passed directly to Zod schemas like `z.string().optional()` (which expects `undefined` rather than `null`)
2. `null` was passed to schemas with defaults like `z.number().default(1)` without triggering the default value logic
3. The `processToAISDKSchema` function used `io: 'output'`, which stripped default markers and made optional/default fields appear required
4. The `#traverse` method in non-OpenAI providers had no logic to handle this conversion

Fields explicitly defined with `.nullable()` were unaffected and continue to accept `null` correctly.

## Solution

Two key changes were made:

1. **Schema Generation**: Modified `processToAISDKSchema` to use `{ io: 'input' }` when calling `standardSchemaToJSONSchema()`. This ensures optional and default fields correctly appear as non-required in the generated JSON schema.

2. **Value Traversal**: Added a `required` array check in each provider's `#traverse` method. For object properties not listed in the schema's `required` array, if the runtime value is `null`, it is converted to `undefined`, allowing Zod to apply `.default()` or `.optional()` correctly.

These changes bring the non-OpenAI providers into alignment with the existing OpenAI compatibility layer behavior.

## Files Changed

**packages/schema-compat/src/provider-compats/**:
- `anthropic.ts`: Updated schema generation and traversal logic (+7/-2)
- `deepseek.ts`: Updated schema generation and traversal logic (+7/-2)
- `google.ts`: Updated schema generation and traversal logic (+7/-2)
- `meta.ts`: Updated schema generation and traversal logic (+7/-2)

**packages/schema-compat/src/provider-compats/test-suite.ts**:
- Added a new "Null Coercion" test block with three validation tests:
  - `null` converts to `undefined` for `.optional()` fields
  - `null` converts to the default value for `.default()` fields  
  - `null` is preserved for `.nullable()` fields
- Changes: +41/-0

**.changeset/polite-kiwis-look.md**:
- Changeset entry documenting the fix (+5/-0)

## Testing

Three new tests were added to the universal test suite verifying:
1. Optional field null coercion to undefined
2. Default field null coercion to default value
3. Nullable field null preservation

All 812 unit tests pass.


nuthalapativarun:
Friendly bump — this has been mergeable for about a week with CodeRabbit clean. The Vercel preview failure is just blocked on team auth, not a real failure. Let me know if there's anything else needed to move this forward. Thanks!

nuthalapativarun:
Hi team — following up on PR #18018 (schema-compat null→undefined fix, closes #17655). Mergeable; CodeRabbit clean; Vercel preview is auth-gated (not a real failure). 2nd bump. Happy to address any feedback. Thanks!

wardpeet:
This is not really an issue when testing, models that do not require fields will omit the values and default will kick in

PR #18026, opened 2026-06-16
fix(memory): catch unhandled promise rejection in async buffered observation

### LINKED ISSUE #14651: [BUG] ObservationalMemory can produce unhandled promise rejection

nuthalapativarun:
## What changed

`ObservationalMemory.startAsyncBufferedObservation` fires a background async operation (`runAsyncBufferedObservation`) and stores it in the `BufferingCoordinator.asyncBufferingOps` mutex map. If a subsequent call arrives before the operation completes it will `await` the existing promise and absorb any rejection. But if **no** subsequent call arrives, the rejection goes unhandled, surfacing as an `UnhandledPromiseRejection` error in Node.js.

There were two unguarded paths:

1. The `asyncOp` chain had `.finally()` but no `.catch()` — so a rejection from `runAsyncBufferedObservation` would propagate past `.finally()` as an unhandled rejection.
2. Inside `runAsyncBufferedObservation`, `this.messageHistory.persistMessages(...)` was called with `await` but no try/catch — a failure there would bubble up uncaught and hit the same gap.

## Fix

- Added `.catch(err => omError(...))` **before** `.finally()` in `startAsyncBufferedObservation` so any rejection is logged and swallowed before cleanup runs.
- Wrapped `persistMessages` in `runAsyncBufferedObservation` with `try/catch` — on failure, logs the error and returns early from the current cycle rather than crashing the async chain.

## Testing

- `pnpm --filter ./packages/memory test` — 217 tests pass (16 pre-existing failures from an unbuilt `@internal/ai-sdk-v5/test` dev package; unrelated to this change).

## Checklist

- [x] Linked issue: Fixes #14651
- [x] Changeset added (`@mastra/memory` patch)
- [ ] Docs updated (no public API change)
- [ ] Tests added (logic covered by existing OM test suite; the race condition is hard to unit-test without mocking internal async state)

## ELI5

The code was trying to do some cleanup work in the background but didn't properly listen when things went wrong. When that background work failed and nobody else was waiting for it, the program would crash with an error. The fix adds "listeners" to catch these failures and log them instead of letting them crash.

## Summary of Changes

This PR fixes an unhandled promise rejection bug in the `ObservationalMemory` class where background async operations could fail silently and cause Node.js to crash.

### Root Causes

1. **Missing catch handler**: The `startAsyncBufferedObservation` method chains a `.finally()` cleanup but has no `.catch()` handler, allowing promise rejections from `runAsyncBufferedObservation` to propagate unhandled.

2. **Unprotected storage call**: The `persistMessages` call inside `runAsyncBufferedObservation` lacks error handling, allowing storage failures to propagate up the async chain.

### Fixes Applied

**In `startAsyncBufferedObservation`:**
- Added an explicit `.catch()` handler before the existing `.finally()` to log any failures with the message `[OM] async buffering observation failed`
- The catch handler ensures rejections are suppressed before cleanup runs

**In `runAsyncBufferedObservation`:**
- Wrapped the "persist sealed messages before buffering" step in a `try/catch` block
- If persisting fails, the code logs `[OM] Failed to persist sealed messages before buffering — skipping observation cycle` and returns early, skipping the rest of that observation cycle instead of crashing

### Why This Matters

During normal multi-step agent execution, the bug was often masked because subsequent `runAsyncBufferedObservation` calls would accidentally consume previous rejections. However, when a failing operation is the last one created, no subsequent call exists to handle it, leaving an unhandled rejection that crashes the process.

### Testing & Changeset

- All 217 tests in the memory package pass
- A patch-level changeset has been added for `@mastra/memory`
- No public API changes were made


nuthalapativarun:
Friendly bump — this has been mergeable for about a week with CodeRabbit clean. The Vercel preview failure is just blocked on team auth, not a real failure. Let me know if there's anything else needed to move this forward. Thanks!

TheIsrael1 (approved):

PR #18034, opened 2026-06-16
fix(core): don't persist abort-cancelled tool calls as completed results

### LINKED ISSUE #17995: Server-side Tool Execution Aborted When Mixed with Client-side Tool Calls in the Same Assistant Turn
This issue was created from Discord post 1516305403934281819:

[Open in Discord](https://discord.com/channels/1309558646228779139/1516305403934281819)

JigyasuRajput:
## Description

When a request is aborted while a server-side tool (`createTool({ execute })`) is still running — e.g. a turn parks waiting on a client-side tool and the SSE response closes, or the user hits Stop — the cancelled tool was persisted as a **completed** tool call whose `result` was the abort message (`"The operation was aborted."`). On resume, that cancelled (possibly half-finished) operation read as a successful tool call.

**Root cause:** `CoreToolBuilder` wraps the tool's `AbortError` in a `TOOL_EXECUTION_FAILED` `MastraError`; the tool-call step returns it as `{ error }`; and `llm-mapping-step` records any tool `{ error }` as a `state:'result'` history entry (intended for genuine errors so the model self-corrects). Aborts flowed through that same path.

**Fix:** when the request is aborted, flag the tool `aborted` (no result/error) instead of `{ error }` — keyed off the abort *signal*, not the error type (the wrapper hides the underlying `AbortError`). The mapping step leaves it an incomplete call, and the to-LLM prompt path drops it on resume. Genuine tool errors on a live request still surface to the model; provider-executed deferred tools and the deliberate `onFinish` skip-on-abort are unchanged.

**Notes:**
- Under the opt-in `filterIncompleteToolCalls: false`, an aborted call now orphans (a recoverable error on the next provider call) instead of fake-completing. Those users already have no orphan protection for any incomplete call (`sanitizeOrphanedToolPairs` is gated behind `filterIncompleteToolCalls=true`); happy to make it run unconditionally as a separate change if preferred.
- Out of scope: the durable-agent path (`agent/durable/workflows/steps/tool-call.ts`) still records aborts as errors, but is unreachable from `handleChatStream`.

## Related issue(s)

Closes #17995

## Type of change

## Checklist

## ELI5

When a task stops before it finishes, the system must not record it as successful. This PR marks stopped server-side tool calls as incomplete so they are not treated as completed when work resumes.

## Summary

This PR fixes persistence of server-side tool calls interrupted by an aborted request. Aborted calls no longer store `"The operation was aborted."` as a successful result.

### Changes

- Added an optional `aborted` field to `toolCallOutputSchema`.
- Updated `createToolCallStep` to detect request aborts through `abortSignal.aborted`.
- Aborted tool calls now return `{ aborted: true }` without a result or error.
- Updated HITL detection to ignore aborted tool calls.
- Excluded aborted calls from the model prompt during resume.
- Preserved tool-use and tool-result pairing.
- Kept genuine tool errors, provider-executed deferred tools, and `onFinish` abort behavior unchanged.

### Testing

Added regression coverage for:

- Aborted and genuine tool errors.
- HITL behavior with aborted tools.
- Resume prompt filtering.
- Schema parsing.
- The complete aborted tool-call pipeline.


JigyasuRajput:
@coderabbitai review

blah-mad:
This is the right distinction. An abort is not a tool result. I’d also persist a tiny cancellation receipt outside the model path: toolCallId, abort source, started_at, side-effect boundary if known, and resume behavior. Otherwise a stopped write can look like completed work later.

JigyasuRajput:
@blah-mad 

agreed. The receipt is a good follow-up but out of scope here - this PR just keeps aborts from being persisted as completed results. The durable receipt (esp. the side-effect boundary, which the tool can't reliably report today) needs its own design tbh.

mikhael28:
@coderabbitai please re-review

JigyasuRajput:
@mikhael28 Lint was a real miss on my side, Prettier formatting in `llm-mapping-step.ts`. Fixed and pushed.
lmk if there's anything else from my side. 

@coderabbitai can you review pls?

PR #18036, opened 2026-06-16
feat(datasets): item-level static tool mocks — engine + storage (1/3)

YujohnNattrass:
## Summary

This is PR 1 of 3 split from #17991.

Adds item-level static tool mocks for deterministic agent experiments: the core engine, types, and storage persistence.

## Scope

- Core engine: `toolMocks` on dataset items (toolName + args + output + `matchArgs` strict/ignore), composed `beforeToolCall` hook, ordered consumption per (toolName, args), abort-on-mismatch/exhaustion, per-item `toolMockReport`.
- `Agent.getConfiguredToolHooks()` accessor so experiment hooks compose with user-configured hooks.
- Storage columns + adapters: libsql, pg, mongodb, spanner persist; **mysql explicitly rejects** rather than dropping silently.
- Shared storage test-utils: new `supportsToolMocks` capability so rejecting adapters assert rejection instead of round-trip.

## Intentionally excluded

- Server API + client-js types → PR 2/3 (`feat/tool-mocks-api`)
- Studio UI + docs + example → PR 3/3 (`feat/tool-mocks-studio`)

## Verification

- `pnpm build:core` ✓
- core tool-mocks tests 26/26 ✓
- mysql suite 533 ✓ (round-trip skipped, rejection asserted)
- libsql suite 796 ✓ (round-trip persists)

## ELI5

This PR lets each dataset item define “pretend tool results,” so running an agent experiment doesn’t depend on calling real tools. Instead, the experiment engine checks the pre-recorded answers, returns the mocked output when they match, and records exactly what happened—especially when a tool is called with the wrong arguments.

---

## Summary

This PR introduces **item-level static tool mocks** for deterministic agent experiments, including execution behavior, per-item diagnostics, and persistence across supported storage adapters (with **MySQL explicitly rejecting** mock persistence).

### Core engine: matching, consumption, and reporting

**`ToolMockMatcher` (`packages/core/src/datasets/experiment/tool-mocks.ts`)**
- Dataset items can declare `toolMocks?: ItemToolMock[]`, where each mock includes `{ toolName, args, output, matchArgs? }`.
- `matchArgs` modes:
  - `strict` (default): deep-equality on `args` (object key order doesn’t matter; array order matters; no coercion).
  - `ignore`: matches by `toolName` only (doesn’t compare `args`).
- Mocks are consumed **in declared order** for the relevant `(toolName, args)` set.
- Tool call resolution:
  - `serve`: a matching unconsumed mock exists → return `output` and mark it consumed.
  - `live`: no matching mock exists → tool runs normally.
  - `fail`:
    - `TOOL_MOCK_EXHAUSTED` when matching mocks exist but are fully consumed.
    - `TOOL_MOCK_MISMATCH` when args don’t match any unconsumed mock.
- Tracking + deterministic failure semantics:
  - `toolMockReport.served`: served mocks (in consumption order)
  - `toolMockReport.unconsumed`: declared mocks left unused (**does not fail** the item)
  - `toolMockReport.liveCalls`: tools that ran live
  - `toolMockReport.failure`: first mismatch/exhaustion that causes item failure
  - After the first failure, subsequent resolutions return failures and do not add further served/live calls.

### Executor integration: intercept tool calls and abort on failure

**`executeTarget` / `executeAgent` (`packages/core/src/datasets/experiment/executor.ts`)**
- `executeTarget` accepts `options.toolMocks?: ItemToolMock[]` for agent targets.
- When `toolMocks` are present:
  - Creates a fresh `ToolMockMatcher` per item execution.
  - Forces deterministic behavior with `runOptions.toolCallConcurrency = 1`.
  - Composes newly built mock interception hooks with user hooks.
  - On mismatch/exhaustion, uses an abort mechanism to stop the run immediately and prevents later tool calls.
  - Returns deterministic failures with `ExecutionResult.error.code` set to `TOOL_MOCK_MISMATCH` or `TOOL_MOCK_EXHAUSTED`, and attaches `ExecutionResult.toolMockReport`.
- When `toolMocks` are empty: tools run live and no `toolMockReport` is attached.

### Hook composition support in `Agent`

**`Agent.getConfiguredToolHooks()` (`packages/core/src/agent/agent.ts`)**
- Adds an accessor to expose the agent’s statically configured tool hooks so experiment code can compose mock-interception hooks with any user-provided hooks.

### Experiment orchestration + retry behavior

**`packages/core/src/datasets/experiment/index.ts`**
- Adds `toolMocks?: ItemToolMock[]` to `ExperimentItem` and propagates it for both inline and dataset-backed item resolution (agent targets only).
- Propagates `toolMockReport` into both in-memory item results and persisted experiment results.
- Updates retry logic to **avoid retrying deterministic tool-mock failures** (`TOOL_MOCK_MISMATCH`, `TOOL_MOCK_EXHAUSTED`).

### Storage persistence across adapters (and MySQL rejection)

**Schema additions**
- Adds persistence fields for:
  - dataset items: `toolMocks`
  - experiment results: `toolMockReport`

**Adapters that persist**
- **libsql, PostgreSQL, MongoDB, Spanner**: full read/write support for `toolMocks` and `toolMockReport`.

**MySQL**
- Explicitly rejects tool mock persistence:
  - dataset item writes with non-empty `toolMocks` throw `MYSQL_DATASET_TOOL_MOCKS_UNSUPPORTED`
  - experiment result writes with `toolMockReport` throw `MYSQL_EXPERIMENT_TOOL_MOCK_REPORT_UNSUPPORTED`

**Test capability flag**
- Shared storage tests introduce `supportsToolMocks` (derived from `capabilities.toolMocks !== false`) to either run round-trip persistence checks or assert rejection (MySQL).

### Types and exports

- Core types extended:
  - `DataItem.toolMocks?: ItemToolMock[]`
  - `ItemResult.toolMockReport?: ToolMockReport`
- Experiment module re-exports tool-mock matcher constants/types.

### Tests

- Unit tests cover strict/ignore matching, consumption order, exhaustion/mismatch failure codes, “first failure only” behavior, and that unconsumed mocks are report-only.
- Execution/scenario tests verify end-to-end determinism:
  - served mocks skip live tool execution
  - argument mismatch aborts immediately (later tools never run)
  - ignore-mode mocks still consume in order and fail on exhaustion
  - tool-mock reports are attached/omitted appropriately.

### Out of scope

The PR excludes the server API, client-js types, and Studio UI (planned for later parts), along with documentation/examples (planned separately).


NikAiyer:
## Critique — `feat(datasets): item-level static tool mocks — engine + storage (1/3)`

**TL;DR:** Strong foundation. Engine is well-factored, failure semantics are deterministic, and the storage divergence is handled honestly. A few UX/clarity items worth addressing before merging the follow-up PRs.

### Verified locally
- ✅ `pnpm build:core` (12 tasks, cached + clean)
- ✅ `packages/core` — `tool-mocks.test.ts` (16) + `tool-mocks-execution.test.ts` (11) + 5 scenario suites (12) → **39/39 passing**
- ✅ `stores/libsql` full suite → **796 passing / 19 skipped** (round-trip tool-mocks tests included via shared `_test-utils`)
- ✅ Confirmed MySQL `#rejectToolMocks` is wired into all three mutation paths: `_doAddItem`, `_doUpdateItem`, `_doBatchInsertItems` (no escape hatch for adding mocks to an existing item).

### What's good

1. **`ToolMockMatcher` as a pure state machine.** The engine is decoupled from the executor, easy to unit test, and the failure codes (`TOOL_MOCK_MISMATCH`, `TOOL_MOCK_EXHAUSTED`) are stable enough to gate behavior on.
2. **Deterministic retry skip** in `runExperiment` (index.ts:355-376). Retrying a mock failure can never produce a different result, so skipping is correct and saves cycles.
3. **Belt-and-suspenders abort.** Hook short-circuits the tool (`proceed: false`), `mockAbort.abort()` fires immediately, `toolCallConcurrency: 1` is forced, and the post-execution fallback (executor.ts:297) catches the race where the model finishes a step before abort propagates. Load-bearing but correct.
4. **`Agent.getConfiguredToolHooks()`** is a minimal, intentional escape hatch — much better than duplicating hook-composition logic in the executor.
5. **MySQL rejection is explicit** with stable, namespaced error codes (`MYSQL_DATASET_TOOL_MOCKS_UNSUPPORTED`, `MYSQL_EXPERIMENT_TOOL_MOCK_REPORT_UNSUPPORTED`). Good migration path UX.
6. **Capability-driven test suite** (`supportsToolMocks` in `_test-utils`) is the right pattern for divergent backends.

### Concerns

1. **Silent drop for inline `task` targets.** In `runExperiment` (index.ts:200-222), an item's `toolMocks` flows through to `ExperimentItem` but is never passed to `execFn` when `targetType !== 'agent'`. A user reusing a dataset across `agent` and `task` targets will see mocks silently ignored on the task path. Recommend either:
   - Logging a warning when `item.toolMocks?.length > 0 && targetType !== 'agent'`, or
   - Documenting the constraint explicitly in the `@mastra/core` changeset and the `ItemToolMock` JSDoc.

2. **`matchArgs: 'ignore'` naming.** Reading the tests it means "match on tool name only, accept any args" — but the literal `'ignore'` could be misread as "ignore specific keys". Consider `'any'` or `'name-only'`, or add a JSDoc example. Worth catching now before downstream PRs solidify the API.

3. **`unconsumed` semantics on mismatch.** In `mock-mismatch-aborts.scenario.test.ts` the report asserts `unconsumed.length === 1` against a 1-mock fixture, so it's ambiguous whether the *mismatched* mock counts as unconsumed or only the trailing tail. A scenario with `[mockA, mockB, mockC]` where `mockA` mismatches would make the contract observable. (My read of `ToolMockMatcher` is: the mismatched mock stays in `unconsumed` because it was never "served" — please confirm.)

4. **`AddExperimentResultInput.toolMockReport: DatasetToolMockReport | null`.** The `null` vs `undefined` distinction isn't documented. Is `null` meaningful ("ran with mocks but produced no report") vs `undefined` ("no mocks configured")? Worth a JSDoc.

5. **Changeset bump for `@mastra/mysql` is `patch`.** This changes runtime behavior by rejecting input shapes that other backends accept. `minor` is more honest — a user upgrading to this patch with a mixed-backend setup could see new throws.

### Nits

- `buildToolMockHooks` returns `afterToolCall: context => userHooks?.afterToolCall?.(context)`. If `userHooks?.afterToolCall` is undefined this still creates a no-op wrapper that the agent calls. Either pass `userHooks?.afterToolCall` directly (preserving undefined) or add a comment explaining why the wrapper is intentional.
- `recordingTool()` uses `z.object({}).passthrough()` — fine, but a one-liner comment that this is deliberately permissive for fixture purposes would help future readers.
- The `scriptedModel()` helper implementing both `doGenerate` and `doStream` is great — consider exporting it from `@internal/test-utils` so the follow-up PRs can reuse it without duplication.

### Approval

LGTM with the items above as **non-blocking follow-ups** — most are doc/naming clarifications. Concern #1 (silent drop on `task` target) is the one I'd most like addressed before users hit it.

YujohnNattrass:
Thanks for the thorough review, @NikAiyer — really appreciate it. Pushed `c1e26360f6` addressing the follow-ups.

### Fixed
1. **Silent drop on non-agent targets (concern #1)** — `runExperiment` now logs a one-time `warn` when a dataset carrying `toolMocks` is run against a `task`/`workflow`/`scorer` target (emitted once per run, not per item). Also documented the agent-only constraint in the `ItemToolMock` JSDoc.
2. **`matchArgs` clarity (concern #2)** — added a JSDoc example contrasting `'strict'` (deep-equal args) vs `'ignore'` (tool-name-only, for LLM-authored sub-agent args). Kept the `'ignore'` name — see skip note below.
4. **`toolMockReport` null/undefined (concern #4)** — JSDoc on `AddExperimentResultInput.toolMockReport`: `null`/`undefined` both mean "no report"; a present report means the item ran with mocks.
5. **MySQL changeset bump (concern #5)** — `patch` → `minor`. Agreed, rejecting input other backends accept is new runtime behavior.
- **`afterToolCall` no-op wrapper (nit #1)** — now passes `userHooks?.afterToolCall` through directly, preserving `undefined` so the agent skips the call when none is configured.

### Skipped (with reasons)
3. **3-mock unconsumed-on-mismatch scenario (concern #3)** — confirmed your read: a mismatched mock is never marked `consumed`, so it stays in `unconsumed` (`tool-mocks.ts` `resolve()` only sets `consumed` on a `serve`). The contract is correct as-is; the existing `mock-mismatch-aborts` test covers it. Happy to add the 3-mock case if you'd like it observable, but it's not exercising new behavior.
- **Rename `'ignore'` → `'any'`/`'name-only'` (concern #2)** — the literal is already shipped across core, server schemas, client-js types, generated route types, and docs. Renaming is cross-stack churn for a naming preference; the JSDoc example resolves the ambiguity.
- **`recordingTool()` passthrough comment (nit #2)** — test-fixture cosmetic, no behavioral value.
- **Export `scriptedModel` from `@internal/test-utils` (nit #3)** — agreed it's the right move, but premature until the follow-up PRs actually reuse it. Will do it when PR2/PR3 need it.

NikAiyer:
Reviewed `c1e26360f6` locally — `pnpm build:core` clean, all 39 focused tool-mock tests pass, CI green. Engine invariants and storage split look solid. A few things worth addressing before merge:

### Should fix before merge

1. **Changeset doesn't mention the sequential-tool-calls implication.** `.changeset/tool-mocks-core.md` advertises deterministic execution but doesn't call out that any item with `toolMocks` forces `toolCallConcurrency: 1` for that item run. Users authoring large mocked datasets may be surprised by the latency hit. Suggest one sentence: "Items with `toolMocks` execute tools sequentially within that item run to guarantee ordered consumption; items without mocks are unaffected."

2. **CodeRabbit's open re-review comment** — looks already addressed by this commit's scope (warn-on-non-agent, JSDoc, `afterToolCall` pass-through, MySQL minor bump). Worth a "resolved by this commit?" reply so the bot dismisses it before merge.

### Worth considering (non-blocking)

3. `Agent.getConfiguredToolHooks()` is marked `@internal` but lives on the exported `Agent` class — anything doing `keyof Agent` will see it. Fine for PR 1/3, but **PR 2/3 should double-check** that server agent-shape DTOs / `client-js` types don't accidentally surface it.

4. The non-agent warning fires once per `runExperiment` (good — no spam), but consider including the item count in the message so users can tell a stray one-off from a fully misrouted dataset.

5. Doc nit in `tool-mocks.ts`: `matchArgs: 'ignore'` is the only way to mock sub-agent delegations (`agent-*`) since their `prompt` is free text. Worth a one-line example in the changeset since that's not obvious from the type name.

### Heads-up for PR 2/3

PR 2 should add an integration test that the **server API** rejects `toolMocks` cleanly (4xx) when the configured storage adapter is MySQL — otherwise users will hit the engine-layer error as a 500.

Otherwise LGTM 🚀

YujohnNattrass:
Thanks @NikAiyer — pushed `c9e22cbf3e`.

### Should fix
1. **Sequential-execution note in changeset** — added to `.changeset/tool-mocks-core.md`: *"Items that declare `toolMocks` run their tools sequentially (`toolCallConcurrency: 1`) within that item run to guarantee ordered consumption. Items without mocks are unaffected."*
2. **CodeRabbit re-review** — confirmed: the open findings are addressed by `c1e26360f6` (warn-on-non-agent, JSDoc, `afterToolCall` pass-through, MySQL minor bump). Resolving the bot comment now.

### Non-blocking
4. **Item count in the warning** — done. The non-agent warning now reads `"… but N of M dataset items declare toolMocks …"` so a stray one-off is distinguishable from a fully misrouted dataset.
5. **`matchArgs` example** — the core changeset's code sample already carries an inline note on `matchArgs` (`'strict'` deep-compares args; `'ignore'` matches tool-name-only for LLM-authored sub-agent calls), so the rename rationale is visible in the changelog.
3. **`getConfiguredToolHooks()` `@internal` leakage** — noted; I'll verify in PR 2/3 that the server agent DTOs and `client-js` types don't surface it.

### Heads-up for PR 2/3
Agreed — PR 2 will add an integration test that the server API rejects `toolMocks` with a 4xx (not a 500) when the configured storage is MySQL. Tracking it for that PR.

NikAiyer (approved):

PR #18041, opened 2026-06-16
fix(@mastra/core): fix parallel sub-agent delegations requiring approval

### LINKED ISSUE #18031: Durable agent: suspended run cannot be resumed after a process restart (prepare() ignores runId; rehydrated resume drops suspendPayload.args)
## Summary

A durable agent (`EventedAgent`) run that suspends for tool approval cannot be resumed after its in-memory run registry is gone (idle eviction, or a process restart) — even though the full workflow snapshot, **including the suspended tool's args**, is durably persisted to storage. Two gaps in the durable-agent layer prevent resuming a persisted, suspended run from a cold process.

## Environment

- `@mastra/core` **1.42.0**
- Storage: `@mastra/pg` `PostgresStore`
- A durable `EventedAgent` wrapping an `Agent` with a `requireApproval: true` tool

## What works

On suspend, the snapshot is fully persisted to `mastra_workflow_snapshot` for both the inner `durable-agentic-execution` and outer `durable-agentic-loop` workflows, with `status: "suspended"` and the suspend payload including the tool args:

```json
"suspendPayload": { "args": { "note": "…" }, "type": "approval", "toolName": "danger_action", "toolCallId": "…" }
```

So everything needed to resume is on disk.

## What fails

After the process that started the run exits (or after the registry TTL elapses), `resume(runId, …)` throws:

```
Error: No registry entry found for run <id>. Cannot resume.
```

because `resume()` reads the in-memory `#runRegistry` (populated by the original `stream()`), which is gone.

`prepare()` looks like the intended rehydration primitive (it rebuilds the registry entry from the agent + a request context), but two issues prevent using it to resume a specific run.

### Gap 1 — `prepare()` ignores `options.runId`

`stream()` forwards the run id to the preparation:

```js
// agent/durable/index.js — stream()
const preparation = await prepareForDurableExecution({
  agent: this.#wrappedAgent, messages, options,
  runId: options?.runId,                 // ← forwarded
  requestContext: options?.requestContext, mastra: this.#mastra,
});
```

but `prepare()` does not:

```js
// agent/durable/index.js — prepare()
const preparation = await prepareForDurableExecution({
  agent: this.#wrappedAgent, messages, options,
  // runId NOT forwarded → prepareForDurableExecution does `providedRunId ?? crypto.randomUUID()`
  requestContext: options?.requestContext, mastra: this.#mastra,
});
```

So `prepare(messages, { runId })` registers a **random** run id, and a subsequent `resume(runId)` still can't find it. Adding `runId: options?.runId` to `prepare()` (mirroring `stream()`) fixes this: `prepare()` then registers the correct id, and `resume()` successfully loads the persisted snapshot and resumes the workflow across a restart.

### Gap 2 — rehydrated resume drops the suspended tool's args

With Gap 1 patched, `resume()` loads the snapshot and runs, but the approved tool executes with **undefined args**:

```
Error: Cannot read properties of undefined (reading 'note')
  at Object.execute (.../createTool input handling)
  at .../agent/durable/index.js  (durable tool step — `const cleanedArgs = { ...args };`, where `args` is undefined)
```

The args are present in the persisted snapshot (`suspendPayload.args`, shown above), but they are not restored into the tool step on a rehydrated (cold) resume. In an in-process resume the args survive because the step state is still in memory; from a cold process they're dropped.

## Expected behavior

It should be possible to resume a persisted, suspended durable run after a process restart — rehydrating the registry from the agent + a fresh request context and resuming from the durable snapshot, with the suspended tool's args restored from `suspendPayload.args`. Either:

- make `prepare()` forward `runId` (Gap 1) **and** restore `suspendPayload.args` on rehydrated resume (Gap 2), or
- provide a dedicated "resume a persisted run by id" API that does both.

## Minimal repro

Two Node processes sharing one `PostgresStore`:

1. **Process A:** `EventedAgent` over an `Agent` with a `requireApproval: true` tool. `agent.stream([...], { runId, memory })`; consume the stream until the approval suspends; confirm `mastra_workflow_snapshot` has rows for the run id; then **exit the process** (simulating a restart).
2. **Process B (fresh):** `resume(runId, { approved: true })` → `No registry entry found`. Patch `prepare()` to forward `runId`, then `await agent.prepare([], { runId, requestContext, memory })` and `resume(runId, { approved: true })` → the snapshot loads and the workflow resumes, but the tool throws `Cannot read properties of undefined (reading 'note')`.

**Runnable two-process reproduction:** https://github.com/bijeebuss/mastra-durable-resume-repro

roaminro:
Fixes a bug where a supervisor agent delegating the same sub-agent twice in one step (e.g. two parallel refunds) only ran the first delegation. Approving them one at a time left the second stuck with `AGENT_RESUME_NO_SNAPSHOT_FOUND`, and on a page refresh the second approval was lost entirely.

Root cause: suspended-tool tracking was keyed by tool name, so two parallel calls to the same tool collapsed into one entry and overwrote each other's suspended run id.

The fix keys the tracking by `toolCallId` (with a tool-name fallback for old data) so each delegation keeps its own suspended run, and also carries the run id in the workflow suspend payload so live in-process resumes don't depend on the shared per-message metadata.

Before:

```ts
await supervisor.approveToolCall({ runId, toolCallId: callA }); // runs refund A
await supervisor.approveToolCall({ runId, toolCallId: callB }); // error: AGENT_RESUME_NO_SNAPSHOT_FOUND
```

After:

```ts
await supervisor.approveToolCall({ runId, toolCallId: callA }); // runs refund A
await supervisor.approveToolCall({ runId, toolCallId: callB }); // runs refund B
```

Added a deterministic regression test covering live resume, cold reload (page refresh), and approval emission. Re-recorded the affected `tool-approval.e2e` recordings since the persisted message shape changed. `agent.ts` is untouched.

## ELI5
When a supervisor asks the same sub-agent two times at once, the old code mixed up the “two separate requests” and only remembered one of them. This PR tags each delegation with its own tracking ID, so both approvals resume correctly—without errors and even after a page refresh.

## Overview
This PR fixes a bug where a supervisor delegating to the same sub-agent multiple times in parallel would only run the first delegation. The second would fail with `AGENT_RESUME_NO_SNAPSHOT_FOUND`, and after a page refresh the second approval could be lost.

## Root cause
Suspended-tool tracking used the *tool name* as the key. When the same tool was suspended twice in parallel, the second suspension overwrote the first suspended run context, leaving only one snapshot to resume.

## Fix / Solution
- Key suspended-tool tracking by `toolCallId` (with `toolName` fallback for backward compatibility with older persisted data)
- Update metadata handling to safely mutate `lastAssistantMessage.content.metadata` in-place to avoid lost writes during concurrent suspensions
- Carry the correct `suspendedToolRunId` inside the workflow suspend payload, and have resume logic prefer that per-`toolCallId` run ID for collision-free parallel resumes
- Add/adjust matching and cleanup logic to remove approval/suspension metadata using `{ toolCallId, toolName }` (preferring `toolCallId`, falling back to `toolName`)
- Ensure Studio’s suspended-tool payload resolution uses `toolCallId` with `toolName` fallback, matching the underlying persisted metadata behavior

## Tests
- Added `parallel-delegation-approval.test.ts` (deterministic Vitest) covering:
  - live resume (approving both pending delegations runs both)
  - cold reload (re-instantiating the agent over the same store still resumes both)
  - approval emission (verifies both distinct approval requests are emitted)

## Test recordings
- Re-recorded the affected `tool-approval.e2e` snapshots to reflect the persisted message shape changes related to `toolCallId`-keyed suspension/approval metadata.

## Documentation
- Added a changeset documenting the fix for approving the same sub-agent delegation twice in parallel within a single supervisor step.


taofeeq-deru (commented):
looks good, but will need to update this line too - https://github.com/mastra-ai/mastra/blob/6581e28a65d72cdb5f0e8d8f892220a5c27160fe/packages/playground/src/lib/ai-ui/tools/tool-card.tsx#L155, to reference toolCallId too since it can either be saved as toolName or toolCallId now.
so something like `suspendedTools?.[toolName] ?? suspendedTools?.[toolCallId]`

wardpeet (commented):
Do you need to update the recordings? Also do we still need toolName fallbacks?

abhiaiyer91 (approved):

PR #18077, opened 2026-06-17
feat(memory): add onTitleGenerated callback to generateTitle config

### LINKED ISSUE #18076: [FEATURE] Expose a callback/hook when a thread title is generated
### Problem

`generateTitle` runs asynchronously and there is no way to be notified when the generated title is available. The only interception point is the storage layer's `saveThread`/`updateThread` method, which means consumers must wrap a specific storage adapter to react to a new title — a significant coupling for what should be a simple notification.

A concrete use case: we stream responses via SSE and want to push a `"title"` event to the client when the title is ready. Today the only working approach is to detect the title inside a custom storage adapter:

```ts
// Inside a custom storage adapter — the only place to hook today
async saveThread(args) {
  const result = await this.inner.saveThread(args);
  if (result.title && !args.existingTitle) {
    notifyClient(result.threadId, result.title); // our SSE push
  }
  return result;
}
```

This breaks down as soon as you want to support multiple storage backends, or use a first-party adapter like `LibSQLStore` or `InMemoryStore`, neither of which expose an extension point. These ultimately require `saveThread` wrapper, and in all cases if you want to ensure that `notifyClient` is only called when the title is first generated, a memory `getThreadById` lookup is also needed in the above implementation.

### Proposed API

Add an `onTitleGenerated` callback to the `generateTitle` option:

```ts
generateTitle: {
  model: myModel,
  // `threadId` argument subscribes caller to being notified only for that thread
  onTitleGenerated?: (threadId: string, callback: (title: string) => void | Promise<void>);
}
```

This would be called once, after the title is generated, regardless of which storage backend is in use. An alternative to the callback could be:

```ts
  getGeneratedTitle?: (threadId: string) => Promise<string>;
```

In its current form, I'm simply forgoing the use of Mastra's auto title generation and instead rolling that myself.

Aayush-engineer:
## Description

`generateTitle` runs asynchronously after the agent responds, but there was no way to be notified when the title was ready. The only intercept point was the storage layer, forcing consumers to subclass or wrap storage adapters (e.g. `LibSQLStore`, `InMemoryStore`) just to react to a new title — significant coupling for what should be a simple notification.

**Fix:** Added an optional `onTitleGenerated` callback to the `generateTitle` object config. It is invoked once after the title is successfully written to storage, regardless of which storage backend is in use.

```ts
memory: new Memory({
  options: {
    generateTitle: {
      model: openai("gpt-4o-mini"),
      onTitleGenerated: (threadId, title) => {
        sseEmitter.push(threadId, { type: "title", title });
      },
    },
  },
}),
```

Fully backward-compatible — the `boolean` form of `generateTitle` is unchanged.

## Related issue(s)

Fixes #18076

## Type of change

## Checklist

## ELI5
If your app asks the system to automatically make a chat thread title, this PR lets you also provide a small “tell me when it’s ready” callback. Once the title has actually been saved, your callback gets the `threadId` and the `title`.

## Overview
This PR adds an optional `onTitleGenerated` callback to the `generateTitle` configuration used by `Memory`. It provides a built-in notification point when a thread title is generated **asynchronously** and has been successfully **persisted to storage**, without needing to customize or wrap storage adapters.

## Problem Being Solved
Before this change, there was no standard way to react right when an async generated title becomes available. Consumers typically had to detect title creation inside storage adapter methods (e.g., `saveThread` / `updateThread`) by subclassing or wrapping storage implementations—creating tight coupling and sometimes requiring extra logic to ensure the event only fires for the first generated title.

## Solution
When `generateTitle` is provided in object form, you can now pass:

- `onTitleGenerated?: (threadId, title) => void | Promise<void>`

The callback is called **once** after the generated title has been written to storage, and it receives the `threadId` and the generated `title`. This works consistently across storage backends.

Example:
```ts
memory: new Memory({
  options: {
    generateTitle: {
      model: openai("gpt-4o-mini"),
      onTitleGenerated: (threadId, title) => {
        sseEmitter.push(threadId, { type: "title", title });
      },
    },
  },
}),
```

## Changes Made
- **Public types/config**
  - Updated `packages/core/src/memory/types.ts` to add `onTitleGenerated?: (threadId: string, title: string) => void | Promise<void>` to the `generateTitle` object shape.
- **Execution flow**
  - Updated `packages/core/src/agent/agent.ts`:
    - `resolveTitleGenerationConfig(...)` now accepts and returns `onTitleGenerated` when provided.
    - During `#executeOnFinish`, after a newly generated title is persisted, the callback is invoked (if present).
- **Release/docs**
  - Added a changeset to publish the patch and document the new callback.

## Backward Compatibility
This is fully backward-compatible:
- The existing boolean form of `generateTitle` is unchanged.
- `onTitleGenerated` is optional and only used when `generateTitle` is configured as an object.

### DIFF
--- a/packages/core/src/agent/agent.ts
+++ b/packages/core/src/agent/agent.ts
@@ -6903,6 +6903,7 @@ export class Agent<
           model: titleModel,
           instructions: titleInstructions,
           minMessages,
+          onTitleGenerated,
         } = this.resolveTitleGenerationConfig(
           config?.generateTitle as
             | boolean
@@ -6910,6 +6911,7 @@ export class Agent<
                 model?: DynamicArgument<MastraModelConfig, TRequestContext>;
                 instructions?: DynamicArgument<string>;
                 minMessages?: number;
+                onTitleGenerated?: (threadId: string, title: string) => void | Promise<void>;
               }
             | undefined,
         );
@@ -6940,6 +6942,9 @@ export class Agent<
                       title,
                       metadata: thread.metadata,
                     });
+                    if (typeof onTitleGenerated === 'function') {
+                      await onTitleGenerated(thread.id, title);
+                    }
                   } catch (error) {
                     this.logger.error('Error persisting generated title:', error);
                   }
@@ -8822,6 +8827,7 @@ export class Agent<
           model?: DynamicArgument<MastraModelConfig, TRequestContext>;
           instructions?: DynamicArgument<string>;
           minMessages?: number;
+          onTitleGenerated?: (threadId: string, title: string) => void | Promise<void>;
         }
       | undefined,
   ): {
@@ -8829,6 +8835,7 @@ export class Agent<
     model?: DynamicArgument<MastraModelConfig, TRequestContext>;
     instructions?: DynamicArgument<string>;
     minMessages?: number;
+    onTitleGenerated?: (threadId: string, title: string) => void | Promise<void>;
   } {
     if (typeof generateTitleConfig === 'boolean') {
       return { shouldGenerate: generateTitleConfig };
@@ -8840,6 +8847,7 @@ export class Agent<
         model: generateTitleConfig.model,
         instructions: generateTitleConfig.instructions,
         minMessages: generateTitleConfig.minMessages,
+        onTitleGenerated: generateTitleConfig.onTitleGenerated,
       };
     }
 
--- a/.changeset/expose-on-title-generated-callback.md
+++ b/.changeset/expose-on-title-generated-callback.md
@@ -1,0 +1,5 @@
+---
+"@mastra/core": patch
+---
+
+Add `onTitleGenerated` callback to the `generateTitle` config. Called once after the generated title is written to storage, enabling use cases like pushing SSE events to clients without wrapping or subclassing a storage adapter.
--- a/packages/core/src/memory/types.ts
+++ b/packages/core/src/memory/types.ts
@@ -1023,6 +1023,24 @@ type BaseMemoryConfig = {
          * Can be static or a function that receives request context for dynamic customization.
          */
         instructions?: DynamicArgument<string>;
+        /**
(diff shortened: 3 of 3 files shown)


packages/core/src/agent/agent.ts
```diff
@@ -6601,13 +6601,15 @@ export class Agent<
...
           minMessages,
+          onTitleGenerated,
         } = this.resolveTitleGenerationConfig(
           config?.generateTitle as
             | boolean
             | {
                 model?: DynamicArgument<MastraModelConfig, TRequestContext>;
                 instructions?: DynamicArgument<string>;
                 minMessages?: number;
+                onTitleGenerated?: (threadId: string, title: string) => void | Promise<void>;
```
ericbiewener, line 6914: While this works, I proposed an API that takes a callback as the second argument so that `onTitleGenerated` acts as a subscription mechanism for the specific `threadId`. That strikes me as the real use-case: wanting to emit an event to a specific response stream for the given thread ID. Obviously I can build this on top of the implementation you have here. I'm just unsure of what the use case would be for _not_ having a thread-specific subscription.

Aayush-engineer: That's a fair point. The current design passes threadId as the first argument to the callback precisely so you can build thread-specific subscriptions on top of it — e.g. looking up a per-thread emitter from a Map<string, SSEEmitter> keyed by threadId. Keeping it as a single callback on the config (rather than a per-thread subscription method) avoids adding a separate subscription API surface while still covering the use case. Happy to revisit if the maintainers feel a subscription pattern belongs in core.

Aayush-engineer:
Rebased onto latest main and resolved the conflict — onTitleGenerated is now invoked inside the try/catch block that upstream added, so callback errors are caught without crashing the agent. CI re-triggered, ready for review.

intojhanurag:
We will handle this by own , thanks !

PR #18078, opened 2026-06-17
feat(playground): add workflow trigger run options

mfrachet:
## Summary
- Add workflow request context, run options, and tracing dialogs to the trigger flow
- Add workflow input type toggle, JSON dialog, map config dialog, and nested graph dialog
- Wire selected/hovered workflow steps into graph cards and expand focused workflow tests

## Verification
- pnpm --filter ./packages/playground test -- src/domains/workflows/workflow/__tests__/workflow-cancel-button.test.tsx src/domains/workflows/workflow/__tests__/workflow-input-data.test.tsx src/domains/workflows/workflow/__tests__/workflow-map-config-dialog.test.tsx src/domains/workflows/workflow/__tests__/workflow-nested-graph-dialog.test.tsx src/domains/workflows/workflow/__tests__/workflow-graph-node.test.tsx src/domains/workflows/workflow/components/__tests__/workflow-card-components.test.tsx --run

## Stack
Base: split/workflow-run-context-layout
Previous: #18081


packages/playground/src/domains/workflows/workflow/components/types.ts
```diff
@@ -16,6 +16,10 @@ export interface WorkflowStepCardViewProps {
   displayStatus?: WorkflowCardDisplayStatus;
   hasStep?: boolean;
   isNestedWorkflowStep?: boolean;
+  stepKey?: string;
+  isSelected?: boolean;
+  isHovered?: boolean;
```
damien-schneider: Do we need hover as React state here? If this only changes the card border / hovered styling, I’d keep it in CSS only

damien-schneider (approved):

packages/playground/src/domains/workflows/workflow/components/types.ts
mfrachet: It's syncronized between the timeline and the chart step card. When you hover one, the other is hovered too (needs JS)

PR #18092, opened 2026-06-17
fix(upstash): preserve workflow run createdAt on re-persist

### LINKED ISSUE #18091: [BUG] Upstash workflow run createdAt resets on re-persist
### Describe the Bug

`WorkflowsUpstash.persistWorkflowSnapshot` does a blind Redis SET with `createdAt: createdAt ?? new Date()`. When the caller omits `createdAt` on a re-persist, the run's creation timestamp is overwritten.

Core never passes `createdAt` on re-persist:
- Default engine: every step via `persistStepUpdate` in `packages/core/src/workflows/handlers/entry.ts`
- Evented engine: `start()` / `resume()` in `packages/core/src/workflows/evented/workflow.ts`

With `@mastra/upstash`, `createdAt` drifts to the latest persist time instead of when the run was first created. This breaks `listWorkflowRuns({ fromDate, toDate })`, sorting by `createdAt`, and Studio display of run start time.

`@mastra/redis` already preserves existing `createdAt` on re-persist. Same class of bug as libsql #18003 / PR #18004. `updateWorkflowResults` (Lua) already preserves `createdAt`; only `persistWorkflowSnapshot` is affected.

### Steps To Reproduce

1. Use `@mastra/upstash` as workflow storage (Upstash Redis URL + token).
2. Call `persistWorkflowSnapshot` with an explicit `createdAt`:
   - workflowName: `my-workflow`
   - runId: `run-1`
   - createdAt: `2024-01-15T10:00:00.000Z`
3. Call `persistWorkflowSnapshot` again for the same run **without** `createdAt` (matches core behavior on step updates).
4. Call `getWorkflowRunById({ runId: 'run-1', workflowName: 'my-workflow' })`.
5. Observe `run.createdAt` is ~now, not `2024-01-15T10:00:00.000Z`.

Minimal code:
```ts
const originalCreatedAt = new Date('2024-01-15T10:00:00.000Z');
await workflows.persistWorkflowSnapshot({ workflowName: 'my-workflow', runId: 'run-1', snapshot, createdAt: originalCreatedAt, updatedAt: originalCreatedAt });
await workflows.persistWorkflowSnapshot({ workflowName: 'my-workflow', runId: 'run-1', snapshot: { ...snapshot, status: 'success' } });
const run = await workflows.getWorkflowRunById({ runId: 'run-1', workflowName: 'my-workflow' });
// run.createdAt !== originalCreatedAt
```

https://github.com/piyush0049/mastra/blob/fix/upstash-workflow-createdat-repersist/stores/upstash/src/storage/domains/workflows/index.test.ts

### Expected Behavior

Re-persisting a workflow snapshot without `createdAt` should preserve the existing `createdAt` from the first persist. Behavior should match `@mastra/redis` (`stores/redis/src/storage/domains/workflows/index.ts`), which reads the existing record before SET when `createdAt` is omitted.

### Environment Information

```block
Package: @mastra/upstash (workflow storage)
Storage: Upstash Redis (REST)
Node: 22.x
OS: Windows 11

Repro level: storage domain unit test (in-memory Redis mock — no Docker required)

Not LLM-dependent — this is a storage-layer timestamp bug on workflow snapshot upsert.
```

### Verification

Piyush0049:
## Description

`WorkflowsUpstash.persistWorkflowSnapshot` was doing a blind Redis `SET` with `createdAt: createdAt ?? new Date()`, so re-persisting a snapshot without an explicit `createdAt` reset the run's creation timestamp.

Core calls `persistWorkflowSnapshot` without `createdAt` on step updates and on evented engine start/resume, so this affected real workflow runs stored in Upstash.

This PR reads the existing record before insert when `createdAt` is omitted, matching `@mastra/redis` behavior, and adds a regression test.

## Related issue(s)

Fixes #18091

## Type of change

- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [x] Test update

## Checklist

## Test plan

- [x] `pnpm --filter ./stores/upstash exec vitest run src/storage/domains/workflows/index.test.ts`

## ELI5

When saving a workflow's progress to Redis, the system was accidentally overwriting the original "creation time" with the current time every time it saved. This PR fixes that by checking what the original creation time was before saving, and keeping that original time instead of replacing it.

### Bug Fix
Fixed a critical bug in `WorkflowsUpstash.persistWorkflowSnapshot` where re-persisting a workflow snapshot without an explicit `createdAt` parameter was resetting the run's creation timestamp. The method now implements a read-before-insert pattern: when `createdAt` is omitted, it reads the existing record to preserve the original timestamp, ensuring `createdAt` is maintained across multiple persist calls while `updatedAt` correctly reflects the latest update time.

### Impact
This bug was affecting real workflow runs because the core engine calls `persistWorkflowSnapshot` without the `createdAt` parameter during:
- Step updates via `persistStepUpdate` in the default engine
- Engine start/resume operations in the evented engine

The timestamp drift was breaking critical functionality including:
- Date-range filtering via `listWorkflowRuns({ fromDate, toDate })`
- Sorting by creation timestamp
- Displaying run start times in Studio

### Implementation Details
The fix reads from `TABLE_WORKFLOW_SNAPSHOT` by `namespace`, `workflow_name`, and `run_id` to retrieve the existing record, then derives `finalCreatedAt` from the stored value. If no existing record is found, it defaults to `new Date()`. This aligns with the existing behavior in `@mastra/redis` and matches the pattern already used in the Lua script for `updateWorkflowResults`.

### Testing
Added a regression test suite that verifies when persisting a workflow snapshot with an explicit `createdAt`, then re-persisting with the same `runId`/`workflowName` while omitting `createdAt`, the stored `createdAt` remains unchanged while `updatedAt` is updated to the new timestamp.

### Files Changed
- `.changeset/quick-dryers-taste.md` - Changesets entry marking a patch release
- `stores/upstash/src/storage/domains/workflows/index.ts` - Implementation fix with read-before-insert pattern
- `stores/upstash/src/storage/domains/workflows/index.test.ts` - New regression test with in-memory Redis mock

### DIFF
--- a/.changeset/quick-dryers-taste.md
+++ b/.changeset/quick-dryers-taste.md
@@ -1,0 +1,5 @@
+---
+'@mastra/upstash': patch
+---
+
+Fixed workflow run createdAt being reset when re-persisting snapshots without an explicit createdAt timestamp.
--- a/stores/upstash/src/storage/domains/workflows/index.test.ts
+++ b/stores/upstash/src/storage/domains/workflows/index.test.ts
@@ -1,0 +1,69 @@
+import { randomUUID } from 'node:crypto';
+import { createSampleWorkflowSnapshot } from '@internal/storage-test-utils';
+import type { Redis } from '@upstash/redis';
+import { describe, expect, it } from 'vitest';
+
+import { WorkflowsUpstash } from './index';
+
+function createInMemoryRedisClient(): Redis {
+  const store = new Map<string, unknown>();
+
+  const matchKeys = (pattern: string) => {
+    const regex = new RegExp(`^${pattern.replace(/[.+?^${}()|[\]\\]/g, '\\$&').replace(/\*/g, '.*')}$`);
+    return [...store.keys()].filter(key => regex.test(key));
+  };
+
+  return {
+    set: async (key: string, value: unknown) => {
+      store.set(key, value);
+      return 'OK';
+    },
+    get: async <T>(key: string) => (store.has(key) ? (store.get(key) as T) : null),
+    del: async (...keys: string[]) => {
+      let deleted = 0;
+      for (const key of keys) {
+        if (store.delete(key)) deleted++;
+      }
+      return deleted;
+    },
+    scan: async (_cursor: string, opts?: { match?: string }) => ['0', matchKeys(opts?.match ?? '*')] as [string, string[]],
+    pipeline: () => {
+      const commands: Array<() => Promise<unknown>> = [];
+      return {
+        get: (key: string) => {
+          commands.push(async () => store.get(key) ?? null);
+        },
+        exec: async () => Promise.all(commands.map(command => command())),
+      };
+    },
+    eval: async () => null,
+  } as unknown as Redis;
+}
+
+describe('WorkflowsUpstash.persistWorkflowSnapshot', () => {
+  it('preserves createdAt across subsequent upserts when caller omits it', async () => {
+    const workflows = new WorkflowsUpstash({ client: createInMemoryRedisClient() });
+    const { snapshot, runId } = createSampleWorkflowSnapshot('running');
+    const workflowName = `wf-${randomUUID()}`;
+    const originalCreatedAt = new Date('2024-01-15T10:00:00.000Z');
+
+    await workflows.persistWorkflowSnapshot({
+      workflowName,
+      runId,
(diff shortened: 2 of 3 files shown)


TheIsrael1:
The upstash fixes have now been applied in #18004 thanks!

PR #18094, opened 2026-06-17
fix(github-signals): strip bot machine-state from PR comments and stop persisting full comment body

DanielSLew:
## Summary

GitHub Signals notifications were ballooning in size because review-bot comment bodies (notably CodeRabbit) carry a large hidden `<!-- internal state start -->...<!-- internal state end -->` machine-state blob — often well over 100KB — plus collapsed `<details>` walkthrough sections. That raw body was stored verbatim in the PR snapshot and forwarded into notification metadata, bloating payloads and risking agent context-window overflow when notifications are listed.

This change sanitizes comment bodies at the snapshot-ingestion chokepoint and stops persisting the full body downstream:

- Strip XML/HTML-like markup generically (not bot-specific markers): HTML comments and `<details>` sections are removed **including their inner content**, unterminated blocks are dropped through end-of-string so a missing closing marker can't smuggle the payload through, remaining standalone tags are stripped, and any leftover stray `<` is removed so no partial markup survives. Cleaned text flows to every downstream consumer (`latestCommentBody` and each `latestComments[].body`).
- **Preserve Markdown code** — inline code spans and fenced code blocks are stashed before stripping and restored after, so human-authored code like `` `<Component>` `` or fenced JSX/TSX is kept intact while bot markup elsewhere is removed.
- Stop persisting the full `latestCommentBody` in notification `metadata.github`; the existing 240-char `latestCommentExcerpt` is retained for display/summary.
- The exported helper is `sanitizeCommentText(body: string): string`.

### ReDoS safety

Block removal uses `indexOf`-based scanning (no backtracking). The only regex in the markup path is a non-backtracking single-tag matcher (`<\/?[a-zA-Z][^<>]*>`). Adversarial input (e.g. tens of thousands of repeated `<!--` / `<!---` openers) is covered by bounded-runtime tests.

## Test plan

- `pnpm vitest run src/index.test.ts` in `signals/github` → **54/54 passing**, including the `sanitizeCommentText` suite: hidden internal-state blob removal, `<details>` content removal, unterminated-block-through-EOF, no stray `<` / partial markup, adversarial repeated-opener guards, and Markdown code preservation (inline `` `<Component>` `` and fenced JSX/TSX).
- `tsc --noEmit` and `eslint` → clean.

A changeset is included (`@mastra/github-signals` patch).

## ELI5
Some GitHub bot comments hide huge “machine state” blobs inside HTML-looking text (often in collapsed `<details>`), which makes our Signals notifications far too large. This PR cleans those hidden parts out before we save/forward notifications, while keeping real Markdown code examples so humans still see useful code.

## Summary
This PR prevents bloated GitHub Signals notification payloads by sanitizing GitHub PR comment bodies during snapshot ingestion and by ensuring notification metadata retains only a short excerpt (not the full comment body).

### Changes Made
**Comment-body sanitization (snapshot ingestion)**
- Added and exported `sanitizeCommentText(body: string): string`.
- Preserves human-authored Markdown code by temporarily stashing and restoring:
  - fenced code blocks (e.g., ```tsx … ```)
  - inline code spans, including multi-backtick variants (e.g., `` ``<Component ...>` ``)
- Strips bot/markup content outside preserved code regions:
  - removes large hidden HTML/XML-like blocks, including `<!-- ... -->` comment bodies
  - removes `<details ...>...</details>` sections including collapsed inner content
  - strips remaining standalone tags
  - removes unterminated markup fragments through end-of-string and removes leftover lone `<` characters
- Normalizes prose whitespace (while explicitly not collapsing blank lines inside preserved fenced code blocks).
- Adds internal `sanitizeCommentBody(body?: string)` wrapper that returns `undefined` when sanitization yields an empty result.
- Updates `getCommentExcerpt()` to generate excerpts from sanitized text (whitespace-normalized) and uses that for notification summaries.

**Downstream storage + notification metadata**
- During Gitcrawl PR snapshot construction, sanitizes:
  - `latestCommentBody`
  - each `latestComments[].body`
- During GitHub notification payload construction, stops persisting the full `latestCommentBody` into notification metadata and keeps only `latestCommentExcerpt`.

**Performance/safety**
- Uses sentinel-token stash/restore for code regions to prevent markup stripping from corrupting code.
- Sanitization is designed to be safe against adversarial inputs (including repeated opener patterns) without catastrophic backtracking (ReDoS).

### Tests
- Expanded `signals/github/src/index.test.ts` with a comprehensive `describe('sanitizeCommentText')` suite covering:
  - removal of large HTML-comment “internal state” blobs
  - `<details>` removal including inner collapsed content
  - standalone tag stripping (e.g., `<br/>`)
  - unterminated comment/tag fragments removed through end-of-string
  - output containing no stray `<` after unterminated markup
  - adversarial/performance cases (repeated comment openers) verified with a `Date.now()` bound
  - code preservation:
    - inline angle-bracket code inside backticks (including multi-backtick)
    - JSX/TSX inside fenced code blocks
    - blank lines inside fenced code blocks preserved verbatim
  - contextual prose with `<` kept as text (e.g., `coverage < 80%`)
- Updated notification-related assertions to verify excerpt-only persistence.

### Release / Changeset
- Added/updated a changeset for `@mastra/github-signals` documenting:
  - markup stripping behavior (including HTML comments and `<details>`)
  - Markdown code preservation
  - notification metadata now keeping only excerpts instead of full sanitized comment bodies.


abhiaiyer91 (approved):

signals/github/src/index.ts
```diff
@@ -453,8 +453,38 @@ function getMergedNotificationSummary(label: string): string {
...
+  return (
+    body
+      // CodeRabbit / similar machine-state block (may appear with or without a matching end marker).
+      .replace(/<!--\s*internal state start\s*-->[\s\S]*?<!--\s*internal state end\s*-->/gi, '')
+      .replace(/<!--\s*internal state start\s*-->[\s\S]*$/gi, '')
+      // Collapsed walkthrough/details sections used by review bots for verbose machine output.
+      .replace(/<details>[\s\S]*?<\/details>/gi, '')
+      // Generic auto-generated comment markers left behind after the blocks above are removed.
+      .replace(/<!--[\s\S]*?-->/g, '')
+      .replace(/\n{3,}/g, '\n\n')
```
TylerBarnes: we should actually  jusut strip any xml-like info in general, comments, tags, etc

TylerBarnes: I think we don't need/want specific tags like this, just any/all of them

TylerBarnes: and stripBotMachineState should probably be something like sanitizeCommentText or similar

TylerBarnes (changes requested):

TylerBarnes:
Thanks — this looks like the right direction overall. Sanitizing comment bodies at Gitcrawl snapshot ingestion is the right chokepoint, and omitting full `latestCommentBody` from notification metadata in favor of `latestCommentExcerpt` directly addresses the oversized bot-comment/context-window issue.

One suggestion: before stripping XML/HTML-like tags with:

```ts
.replace(/<\/?[a-zA-Z][^>]*>/g, '')
```

consider temporarily preserving Markdown code spans/fenced code blocks, then restoring them after sanitization. The current regex will also strip legitimate code-ish content like `` `<Component>` ``, generic type examples, or JSX in fenced code blocks. Since the UI renders Markdown code blocks/spans, preserving those keeps human-authored code examples intact while still removing bot markup elsewhere.

One lightweight way to do it:

```ts
function preserveMarkdownCode(text: string): { text: string; restore: (text: string) => string } {
  const preserved: string[] = [];

  const stash = (match: string) => {
    const token = `\u0000CODE${preserved.length}\u0000`;
    preserved.push(match);
    return token;
  };

  const protectedText = text
    // Fenced code blocks first so inline-code matching does not touch their contents.
    .replace(/```[\s\S]*?```/g, stash)
    // Simple inline code spans.
    .replace(/`[^`\n]*`/g, stash);

  return {
    text: protectedText,
    restore: sanitized =>
      sanitized.replace(/\u0000CODE(\d+)\u0000/g, (_, index) => preserved[Number(index)] ?? ''),
  };
}

export function sanitizeCommentText(body: string): string {
  const { text, restore } = preserveMarkdownCode(body);

  return restore(
    text
      .replace(/<!--[^-]*-+(?:[^->][^-]*-+)*>/g, '')
      .replace(/<\/?[a-zA-Z][^>]*>/g, '')
      .replace(/<!--/g, '')
      .replace(/</g, '')
      .replace(/\n{3,}/g, '\n\n')
      .trim(),
  );
}
```

Suggested coverage:
- ``sanitizeCommentText('Use `<Component>` here')`` preserves `` `<Component>` ``
- fenced JSX/TSX code preserves `<Component />`
- real markup like `<details open>...</details>` is still stripped outside code

Also, the PR body looks slightly stale: it mentions `stripBotMachineState()`, but the exported helper is now `sanitizeCommentText()`, and it describes removing `<details>` sections while the current behavior strips tags but preserves inner text.

TylerBarnes (approved):
Looks good. The latest commits address the sanitizer concerns: hidden comment/details blocks are removed with ReDoS-safe scanning, full comment bodies stay out of notification metadata, and Markdown inline/fenced code is preserved while markup outside code is still stripped. Approving assuming the remaining pending CI checks stay green.

PR #18105, opened 2026-06-17
fix(core): preserve signal transcript ordering

TylerBarnes:
This fixes signal turns getting attached to the previous assistant response in agent loops.

Before, some signal drains wrote directly into the input list or added state signals without first closing the current assistant response. That meant persisted memory could reload like this:

```txt
assistant: tool calls + final answer
state: browser snapshot
state: browser delta
```

After this, signal drains go through the canonical signal transcript path and rotate the response message id before adding follow-up signals:

```txt
assistant: tool call
state: browser snapshot
assistant: next tool call
state: browser delta
assistant: final answer
```

Also makes processor `sendSignal`/state-signal paths use the same boundary + rotation behavior, so stream order and persisted transcript order stay aligned.

Verified with focused core tests for agent signals, message-list ordering, processor runner, and agentic execution signal paths.

## ELI5
This PR fixes a bug where “signals” (extra updates like snapshots/deltas) were being saved under the *wrong* assistant message during agent loops. Now, the code closes out the current assistant message and switches to a fresh message ID before saving signals, so each signal always ends up in the right spot in the transcript.

---

## Overview
The fix corrects agent-loop signal drain behavior by ensuring all pending signals are persisted through the canonical signal transcript path, with consistent response-message boundary marking and response-message ID rotation. This prevents signals from being incorrectly grouped under the previous assistant response.

### Response boundary + message-id rotation (core loop)
- **`packages/core/src/loop/loop.ts` + `packages/core/src/loop/types.ts`**: Added per-run `currentResponseMessageId` and a `rotateResponseMessageId()` function, exposed on `LoopRun`.
- **`packages/core/src/loop/workflows/agentic-execution/llm-execution-step.ts`**: Uses the loop-provided `rotateResponseMessageId()` when draining pre-run signals.
- **`packages/core/src/loop/workflows/agentic-execution/signal-drain-step.ts`**: When pending signals exist, uses `rotateResponseMessageId()` to choose the next message ID (returned as the step’s `messageId`) instead of ad-hoc/generated IDs.
- **`packages/core/src/loop/workflows/agentic-loop/index.ts`**: Reworked pending-signal handling to:
  - mark a response boundary on `messageList`
  - rotate and write back the new response message ID into `typedInputData` (`messageId` / `stepResult.messageId`)
  - enqueue transcript-ready signal parts via `messageList.addSignal()` (canonical path), instead of pushing raw messages
  - refresh `typedInputData.messages` views from `messageList`

### Canonical signal transcript path + aligned processor behavior
- **`packages/core/src/processors/send-signal.ts`**: Now marks the response-message boundary and rotates the response message ID before transcript writes. It also records transcript data using `messageList.addSignal(signal)` (returning `signalForTranscript`) rather than writing transcript data from the original signal.
- **`packages/core/src/processors/runner.ts`**: Threads optional `rotateResponseMessageId` into state-signal processing by adding a `beforeAddStateSignal` hook that:
  - calls `messageList.markResponseMessageBoundary()`
  - rotates the response message ID at insertion time  
  This hook is wired into both the internal `sendStateSignal` path and computed state delta insertion, and also into the workflow-based input-step path.
- **`packages/core/src/agent/state-signals.ts`**: Extended `applyStateSignal` with an optional `beforeAddSignal?: () => void` callback, invoked immediately before adding the signal to `messageList` / writing it.

## Tests / verification
- Updated mocks/expectations around `MessageList` signal handling and boundaries (notably `addSignal` and `markResponseMessageBoundary`) in:
  - **`packages/core/src/loop/workflows/agentic-execution/goal-step.test.ts`**
  - **`packages/core/src/workflows/processor-step.test.ts`**
- Changeset marks **`@mastra/core`** as a patch release with notes about the corrected agent signal drain recording/order behavior.


abhiaiyer91 (approved):

CalebBarnes (approved):
Reviewed the diff, checked out the branch, built core, and ran the focused signal/ordering/runner/agentic tests — all green (`tsc --noEmit` clean too).

The approach is clean: collapsing every signal path onto the same `markResponseMessageBoundary` → `rotateResponseMessageId` → `addSignal` protocol is the right unification, and threading `rotateResponseMessageId` through the loop/processor/durable paths is consistent. The before/after transcript example in the description matches what the code now does.

One suggestion (non-blocking): this fixes transcript *ordering*, but there's no test asserting the new interleaved order. The existing signal/ordering tests pass but didn't catch the original bug, so they won't guard against it regressing. A small test that drains a pending signal mid-loop and asserts the resulting message sequence (assistant → signal → assistant, with rotated ids) would lock in the fix.

LGTM otherwise.

Issue #16251, opened 2026-05-06
[Windows] mastra core build error


daneatmastra:
This issue was created from Discord post 1501476129683279933:

[![Open in Discord](https://img.shields.io/badge/Open_in_Discord-5865F2?style=for-the-badge&logo=discord&logoColor=white)](https://discord.com/channels/1309558646228779139/1501476129683279933)

I pulled the latest `upstream/main` and `@mastra/core` still does not build locally.

Commit tested:
feeeca9845

Commands run:
git fetch upstream
git merge upstream/main
pnpm install
cd packages/core
pnpm run build:lib --watch

Result:
`tsup --silent --config tsup.config.ts --no-dts --watch` fails with 45 TypeScript errors in `packages/core`.

Current failing areas:
- `src/agent/agent.ts`
  - `result.result` is inferred as `unknown`
  - `unknown` not assignable to `MastraModelOutput<OUTPUT>`
- `src/evals/base.test-utils.ts`
  - properties such as `inputLength`, `outputLength`, `inputFromAnalyze`, `outputFromAnalyze`, `reformattedInput`, `reformattedOutput` are missing on `{}`
- `src/harness/tools.ts`
  - implicit `any` parameters (`o`, `t`)
- `src/loop/workflows/stream.ts`
  - spread from non-object type
  - `executionResult.result` is `unknown`
- `src/stream/base/output-format-handlers.ts`
  - `ValidateAndTransformFinalResult<unknown>` not assignable to `ValidateAndTransformFinalResult<OUTPUT>`
  - `finalResult` possibly `undefined`
  - `value` / `error` property access errors
- `src/tools/validation.ts`
  - `unknown` not assignable to `Record<string, any> | T`

Environment:
- Windows
- pnpm 10.29.3
- TypeScript 6.0.3

I had earlier also seen Windows `tsup`/`esbuild` startup issues, but after re-fetching and retrying, the current reproducible blocker is the TypeScript build failure above on latest `main`.

If helpful, I can provide the full error output, but the failure is reproducible from:
`packages/core -> pnpm run build:lib --watch`

daneatmastra:
Thank you for reporting this issue! We have labeled it for trio-wp and we will look into it as soon as possible.

If you're experiencing an error, please provide a [minimal reproducible example](https://github.com/mastra-ai/mastra/blob/main/CONTRIBUTING.md#minimal-reproduction) whenever possible to help us resolve it quickly.

daneatmastra:
[...]

PR #18125, opened 2026-06-18
fix(core): source harness v1 runId/traceId from the session

### LINKED ISSUE #18126: [BUG] Harness v1 ask_user/submit_plan tools read runId/traceId from non-existent context fields (tsc build fails)
## Describe the Bug

The Harness v1 built-in tools `ask_user` and `submit_plan` (`packages/core/src/harness/v1/tools.ts`) read `runId` and `traceId` from fields that don't exist on their context types:

```ts
runId: context.agent?.runId ?? context.workflow?.runId ?? null,
traceId: context.traceId ?? null,
```

- `AgentToolExecutionContext` has no `runId` property.
- `ToolExecutionContext` has no top-level `traceId` property.

So `tsc` fails with 4 errors when building `@mastra/core`:

```
src/harness/v1/tools.ts(219,35): error TS2339: Property 'runId' does not exist on type 'AgentToolExecutionContext<unknown, unknown>'.
src/harness/v1/tools.ts(220,30): error TS2339: Property 'traceId' does not exist on type 'ToolExecutionContext<unknown, unknown, unknown>'.
src/harness/v1/tools.ts(240,35): error TS2339: Property 'runId' does not exist on type 'AgentToolExecutionContext<unknown, unknown>'.
src/harness/v1/tools.ts(241,30): error TS2339: Property 'traceId' does not exist on type 'ToolExecutionContext<unknown, unknown, unknown>'.
```

This is **platform-independent** (fails on macOS/Linux/Windows). It isn't caught by CI because the test suite type-checks via Vitest (esbuild transpile, no `tsc` emit) — the full `tsc` `.d.ts` build only runs at publish time. At runtime esbuild strips the types, so `context.agent?.runId` / `context.traceId` silently evaluate to `undefined`, meaning these fields were effectively **always `null` and never populated**.

Likely introduced in #17534 ("Refine Harness v1 session records and tools").

## Steps To Reproduce

1. Clone the monorepo and install deps (`pnpm install`).
2. Run `pnpm --filter @mastra/core exec tsc --noEmit -p tsconfig.build.json` (or `pnpm build:core`).
3. Observe the 4 `TS2339` errors in `src/harness/v1/tools.ts` shown above.

Reproduces directly in this repository — no separate project is needed. The `tsc` build of `@mastra/core` on any platform surfaces the errors above (`pnpm --filter @mastra/core exec tsc --noEmit -p tsconfig.build.json`).

## Expected vs Actual

- **Expected:** `@mastra/core` builds cleanly under `tsc`; the pending-item `runId` / `traceId` reflect the active run.
- **Actual:** `tsc` fails with 4 `TS2339` errors; at runtime the values were always `null`.

## Suggested fix

The harness session already tracks the active run via the public `getCurrentRunId()` / `getCurrentTraceId()` getters. Sourcing the values from the session fixes the type error and correctly populates them (proposed in a PR).

## Environment

- `@mastra/core` (monorepo build), TypeScript 6.0.3, all platforms.

praveenpke:
## Description

A pre-existing, platform-independent structural type bug in the Harness v1 built-in tools. It was surfaced while getting the monorepo to build on native Windows (#16251), but it fails the `tsc` type-emit on **all** platforms — only undetected because CI type-checks via Vitest (no `tsc`).

`packages/core/src/harness/v1/tools.ts` (the `ask_user` and `submit_plan` tools) read:

```ts
runId: context.agent?.runId ?? context.workflow?.runId ?? null,
traceId: context.traceId ?? null,
```

But `AgentToolExecutionContext` has no `runId` and `ToolExecutionContext` has no top-level `traceId`, so this is a real type error (4× TS2339). At runtime (esbuild strips types) `context.agent?.runId` and `context.traceId` were always `undefined`, so `runId`/`traceId` were **never populated** — both resolved to `null`.

```ts
// after
runId: session.getCurrentRunId(),
traceId: session.getCurrentTraceId(),
```

The harness session already tracks the active run (`#markRunning(runId)`), exposed via the public `getCurrentRunId()` / `getCurrentTraceId()` (both `string | null`, matching `HarnessPendingItemRecord.runId?` / `traceId?`).

### Behavior & safety (all platforms)

- These tools fire only inside `session.signal()` (the agent path), where `#markRunning` has already set the active run **before** the tools are built. So `getCurrentRunId()` now returns the **real active runId** — previously always lost as `null`. `getCurrentTraceId()` returns `null` today (identical to the old value), now correctly wired.
- Worst case (tool fires while not running): the getters return `null` — exactly the old value. **No regression on any platform.**
- The values are write-only metadata for the `question` / `plan-approval` item kinds these tools create; the only `item.runId` consumer is gated to `kind === 'tool-approval'`, and `item.traceId` is read nowhere.

### Verification

- `@mastra/core` typecheck: removes the final 4 errors (→ **0**) on top of the Windows path-separator fixes (#16251 companion PR).
- Harness v1 test suite green (34 files / 349 tests); no test asserts these values.

### Changeset

`@mastra/core` — patch.

## Related issue(s)

Fixes #18126. Surfaced while fixing #16251 (the Windows build). This is a separate, all-platform `tsc` type error; both this and the path-separator PR (#18124) are required for a fully green build.

## Type of change

## Checklist

## ELI5

This PR fixes some tools in the Harness system that were accidentally looking for information (`runId` and `traceId`) in the wrong place - like looking for your keys in a drawer when they're actually on the table. The fix tells them to look in the session object instead, where that information actually exists.

## Summary

This PR fixes structural type errors in the Harness v1 built-in tools (`ask_user` and `submit_plan`) in `packages/core/src/harness/v1/tools.ts`. The tools were attempting to access `runId` and `traceId` properties from the tool execution context, but these properties don't exist at those locations (`AgentToolExecutionContext` lacks `runId` and `ToolExecutionContext` lacks a top-level `traceId`).

### Changes

- Updated the tools to retrieve `runId` via `session.getCurrentRunId()` and `traceId` via `session.getCurrentTraceId()` instead of attempting to read them from the context object
- The session object already has access to these values through its `#markRunning(runId)` method, which is called before the tools are constructed

### Impact

- **Type safety**: Resolves four TS2339 type-check errors that were previously undetected in CI
- **Runtime behavior**: No functional changes when tools execute within the agent path; the tools now provide the actual `runId` instead of `null`. When firing outside an active run, the methods return `null`, matching previous behavior
- **Build**: Marked as a patch-level fix for `@mastra/core` and resolves all remaining type-check errors in the core package

### DIFF
--- a/.changeset/crisp-squids-sleep.md
+++ b/.changeset/crisp-squids-sleep.md
@@ -1,0 +1,5 @@
+---
+'@mastra/core': patch
+---
+
+Fixed Harness v1 `ask_user` and `submit_plan` tools reading `runId` and `traceId` from fields that do not exist on the tool execution context, which broke the type build. They now read the current run and trace IDs from the harness session.
--- a/packages/core/src/harness/v1/tools.ts
+++ b/packages/core/src/harness/v1/tools.ts
@@ -216,8 +216,8 @@ export function buildHarnessBuiltInTools(session: AnySession): ToolsInput {
             id: randomUUID(),
             kind: 'question',
             status: 'pending',
-            runId: context.agent?.runId ?? context.workflow?.runId ?? null,
-            traceId: context.traceId ?? null,
+            runId: session.getCurrentRunId(),
+            traceId: session.getCurrentTraceId(),
             payload: input,
           });
           return { isError: false, pendingItemId: pending.id, status: pending.status };
@@ -237,8 +237,8 @@ export function buildHarnessBuiltInTools(session: AnySession): ToolsInput {
             id: randomUUID(),
             kind: 'plan-approval',
             status: 'pending',
-            runId: context.agent?.runId ?? context.workflow?.runId ?? null,
-            traceId: context.traceId ?? null,
+            runId: session.getCurrentRunId(),
+            traceId: session.getCurrentTraceId(),
             payload: { ...input, transitionModeId: session.getMode().transitionsTo },
           });
           return { isError: false, pendingItemId: pending.id, status: pending.status };


abhiaiyer91:
We will be removing Harness v1, so closing this

PR #18129, opened 2026-06-18
fix(core): require approval by default for execute_command on a local sandbox

### LINKED ISSUE #18130: [BUG] workspace execute_command runs without an approval gate by default on an un-isolated local sandbox
### Describe the Bug

**Affected component:** `@mastra/core` — workspace sandbox command execution.

When a `Workspace` is configured with the built-in `LocalSandbox` in its default isolation mode (`isolation: 'none'`), the auto-exposed `mastra_workspace_execute_command` tool runs model-selected commands directly on the **host shell** with no OS-level filesystem or network isolation. The built-in workspace tool configuration resolves `requireApproval` to `false` by default, so for an agent driven by model output the tool call can reach the host shell with **no human-in-the-loop approval** the moment such a workspace is exposed.

Concretely, `createWorkspaceTools` registers `execute_command` whenever `workspace.sandbox.executeCommand` exists, and `resolveToolConfig` resolves the built-in `requireApproval` default to `false`. `LocalSandbox` defaults `isolation` to `'none'`, which executes on the host with `shell: true`, so a model-chosen command string flows to the host shell with no approval step in the default configuration.

Decisive sink (un-isolated `LocalSandbox` invoking the host shell on the wrapped, model-supplied command):

- `packages/core/src/workspace/sandbox/local-process-manager.ts:258`

I'm filing this as a defense-in-depth / fail-safe-default concern rather than a crash bug: the dangerous default is the missing approval gate, and the practical risk appears only once an application chooses to expose a workspace whose sandbox runs un-isolated on the host. There is no `SECURITY.md` / private channel in this repo, so I'm reporting it here as a neutral hardening request; happy to move it to a private channel if maintainers prefer.

### Steps To Reproduce

1. Create a Mastra project (`npx create-mastra@latest`).
2. Define a `Workspace` backed by the built-in `LocalSandbox` using default options (i.e. `isolation` left at its default of `'none'`), and attach it to an agent so the workspace tools are exposed:
   ```ts
   import { Workspace } from '@mastra/core/workspace';
   import { LocalSandbox } from '@mastra/core/workspace'; // built-in local sandbox

     sandbox: new LocalSandbox({ workingDirectory: './ws' }),
   });
   // workspace tools (including mastra_workspace_execute_command) are now exposed to the agent
   ```
3. Inspect the resolved tool configuration produced by `createWorkspaceTools`, or run the agent and let the model invoke `mastra_workspace_execute_command` with a harmless probe command.
4. Observe that `mastra_workspace_execute_command` resolves to `requireApproval: false`, so the command is dispatched to the host shell (via the sandbox process manager) with no approval prompt — there is no human-in-the-loop step in the default configuration.

(No weaponized command is needed to demonstrate the issue: the point is that a single model-driven tool call — including one induced by prompt-injected or otherwise untrusted content the model is processing — reaches the host shell un-gated by default.)

https://github.com/mastra-ai/mastra/pull/18129

### Expected Behavior

For an un-isolated, host-backed sandbox (`LocalSandbox` with `isolation: 'none'`), `execute_command` should be **fail-safe by default**: it should require approval unless an application explicitly opts out. In other words, the default for a sandbox that runs directly on the host with no isolation should be `requireApproval: true`, raising the floor while leaving the behavior fully configurable.

Suggested hardening (default-only, backward-compatible):

- Default `mastra_workspace_execute_command` to `requireApproval: true` **only** when the configured sandbox runs un-isolated on the host (detectable via `sandbox.isolation === 'none'`, which only the built-in `LocalSandbox` reports).
- Keep this as a default that merely raises the floor: an explicit top-level `tools.requireApproval: false`, or a per-tool `mastra_workspace_execute_command: { requireApproval: false }`, should still win, so applications that intentionally run unattended keep working unchanged.
- Leave isolated and remote sandboxes unaffected: `LocalSandbox` with `seatbelt`/`bwrap`, and cloud/container providers (E2B, Docker, Modal, …), keep the existing non-gated default.

This preserves the tool's documented general-shell semantics (pipes, redirects, `&&`/`||`/`;` chaining) — it adds an approval gate rather than restricting the command surface.

I've prepared a fix along these lines and can open a PR (it changes one source file plus regression tests in `workspace-safety.test.ts`).

### Environment Information

```block
- Affected package: @mastra/core (workspace sandbox command execution)
- Decisive sink: packages/core/src/workspace/sandbox/local-process-manager.ts:258
- Configuration: Workspace + built-in LocalSandbox with default isolation ('none')
- LLM Provider: any (the issue is in the default approval configuration, not provider-specific)
- OS: any host on which an un-isolated LocalSandbox executes commands

_(`npx envinfo --system --browsers --binaries --npmPackages --npmGlobalPackages` output to be attached from the reproduction project.)_
```

### Verification

Jiangrong-W:
Fixes #18130

## Description

When a `Workspace` is configured with the built-in `LocalSandbox` in its default
isolation mode (`isolation: 'none'`), the auto-exposed `mastra_workspace_execute_command`
tool runs model-selected commands on the **host shell** (via `execa` with `shell: true`)
with no OS-level filesystem or network isolation. The built-in workspace tool config
defaulted `requireApproval` to `false`, so for an agent driven by model output a single
tool call — including one induced by prompt-injected or otherwise untrusted content —
could reach the host shell with **no human-in-the-loop gate** the moment such a workspace
is exposed.

This PR makes that default path **fail-safe** while keeping the tool fully configurable
and backward-compatible. It changes a default only; it does not restrict the command
surface or alter the execution path.

**Root cause.** `createWorkspaceTools` registers `execute_command` whenever
`workspace.sandbox.executeCommand` exists, and `resolveToolConfig` resolved the built-in
`requireApproval` default to `false`. `LocalSandbox` defaults `isolation` to `'none'`,
which executes on the host with `shell: true`, so the model-reachable command string flows
to the host shell with no approval step in the default configuration.

**The change (minimal, one source file).** In
`packages/core/src/workspace/tools/tools.ts`:
- Add `sandboxRunsUnisolatedOnHost(sandbox)` — duck-typed on `sandbox?.isolation === 'none'`.
  Only the built-in `LocalSandbox` reports `isolation`; cloud/container providers
  (E2B, Docker, Modal, …) do not, so they are treated as isolated. No `LocalSandbox`
  import is introduced, so the browser-safe module boundary is preserved.
- Add an optional `ToolConfigDefaults` and a 4th optional `defaults` parameter to
  `resolveToolConfig`; the built-in `requireApproval` default becomes
  `defaults?.requireApproval ?? false` — this only **raises the floor**. Host top-level
  and per-tool config is applied afterward and still wins.
- Thread `defaults` through `addTool`, and when registering `execute_command` pass
  `{ requireApproval: true }` if the sandbox is un-isolated.

Net effect for the default (un-isolated `LocalSandbox`): `execute_command.requireApproval`
flips from `false` to `true`, so the call is suspended for approval before it can reach
`LocalProcessManager.spawn`/`execa`. The execution-time approval gate already consumes the
tool's `requireApproval` flag, so no change to the loop/agent layer is needed. This single
default closes the one path below.

**Affected sites (decisive sink):**
- `packages/core/src/workspace/sandbox/local-process-manager.ts:258` — un-isolated
  `LocalSandbox` invokes `execa` in shell mode on the wrapped, model-supplied command.

The general shell semantics of the tool (pipes, redirects, `&&`/`||`/`;` chaining) are a
documented feature and are intentionally preserved — this adds an approval gate rather than
restricting what commands may run, so the tool's purpose is unchanged.

## Related issue(s)

N/A — no existing tracking issue is linked yet; a maintainer may request one.

## Type of change

- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [ ] Test update

**Compatibility.** No public API break — only an optional parameter with a safe default is
added; signatures stay source/binary compatible. The previous behavior is fully restorable
per workspace:

```ts
new Workspace({
  sandbox: new LocalSandbox({ workingDirectory: './ws' }),
  tools: {
    // opt out globally …
    requireApproval: false,
    // … or just for command execution:
    // mastra_workspace_execute_command: { requireApproval: false },
  },
});
```

Isolated and remote sandboxes are unaffected: `LocalSandbox` with `seatbelt`/`bwrap`, and
all cloud/container providers, keep `requireApproval` defaulting to `false`. For
untrusted/model-driven flows that need unattended execution, the recommended hardening is
native isolation, e.g. `new LocalSandbox({ isolation: 'seatbelt' /* or 'bwrap' */ })`, which
keeps `execute_command` non-gated because the sandbox is isolated.

## Checklist

### Testing

Regression coverage lives in `packages/core/src/workspace/workspace-safety.test.ts`:
- The existing default-config test now asserts that the un-isolated `execute_command`
  resolves to `requireApproval: true`, while filesystem/search tools stay `false`.
- A new `describe('execute_command host-isolation approval default')` adds four cases:
  the new default (un-isolated → `true`), the top-level opt-out
  (`tools.requireApproval: false` → `false`), the per-tool opt-out (→ `false`), and the
  unchanged isolated-sandbox behavior (`isolation: 'seatbelt'` → `false`).

The tests are hermetic — they assert on the tool config produced by `createWorkspaceTools`
and spawn no real process or network. They fail on the pre-fix tree (the un-isolated default
resolved to `requireApproval: false`) and pass with this change (24 passed). `eslint` passes
on the changed files.

### Notes for maintainers

There is a second, inlined `resolveToolConfig` in
`packages/server/src/server/constants.ts` used only for tool listing/enablement in the server
handlers; it is not on the execution-time approval path, so it is intentionally left unchanged
here. Happy to align the two helpers in a follow-up if you'd prefer them to stay in lockstep.

## ELI5
Mastra’s “workspace” can run shell commands on your machine. This PR adds a safety switch so that when the sandbox isn’t truly isolated, the AI must get your explicit approval before running those shell commands (to prevent prompt-injection from triggering them).

## Summary
This PR fixes a security vulnerability in Mastra workspace tool defaults by changing the built-in `mastra_workspace_execute_command` tool to require human approval by default when the workspace uses the built-in `LocalSandbox` in its default un-isolated mode (`isolation: 'none'`). Previously, the tool defaulted to `requireApproval: false`, so a single malicious or prompt-injected tool call could execute host shell commands with no approval gate.

### The Problem
With the default `LocalSandbox` (`isolation: 'none'`), `execute_command` runs model-selected commands directly on the host shell (via `execa` with `shell: true`). Because the default configuration allowed `requireApproval: false`, untrusted model input could reach the host shell without a human-in-the-loop approval step.

### The Fix
- Adds a helper `sandboxRunsUnisolatedOnHost()` that detects un-isolated host execution via `sandbox?.isolation === 'none'`.
- Introduces a small “built-in defaults” mechanism (`ToolConfigDefaults` and an extended `resolveToolConfig(..., defaults?)`) so built-in tool registrations can set safe fail-safe defaults without changing user config semantics.
- When registering the `EXECUTE_COMMAND` tool for an un-isolated host sandbox, the PR sets `requireApproval: true` by default.
- Isolated local sandbox backends and remote/container providers remain unchanged (their defaults stay `requireApproval: false`).

### Backward Compatibility / Opt-out
The change is fail-safe by default but fully configurable:
- Global opt-out: `tools.requireApproval: false`
- Per-tool opt-out: `mastra_workspace_execute_command: { requireApproval: false }`
User/host configuration precedence is preserved over built-in defaults.

### Code and Test Changes
- **`packages/core/src/workspace/tools/tools.ts`**: implement conditional default approval for un-isolated host sandboxes, via `sandboxRunsUnisolatedOnHost`, `ToolConfigDefaults`, and `resolveToolConfig` default plumbing.
- **`packages/core/src/workspace/workspace-safety.test.ts`**: update and extend tests to verify:
  - default approval is required for `isolation: 'none'`
  - both host-level and per-tool opt-outs still work
  - isolated sandboxes keep `requireApproval: false`
- **Changeset**: records the patch release note for the safer approval default.


intojhanurag (approved):
Add changeset during PR ! I added one for you, Thanks .

wardpeet (approved):

PR #18138, opened 2026-06-18
Refactor topics into signals

mfrachet:
## NOTE:

- Behind a feature flag
- I need it on main so I can work on Daniel's branch
- More iterations to come

CleanShot 2026-06-24 at 09 19 18@2x
CleanShot 2026-06-24 at 09 19 28@2x
CleanShot 2026-06-24 at 09 19 36@2x

## Summary

Introduces the **Signals** observability experience (replacing Topics) and moves its reusable UI into the shared `@mastra/playground-ui` package so other apps can consume it, while keeping `packages/playground` as a thin route/app shell.

### Signals experience
- Replace the Topics app routes with Signals routes and navigation.
- Signals overview with facet metric cards (trace-share %) and detail CTAs.
- Signals detail page with top-level **Trace list / Chart** tabs, facet filtering, and a dedicated trace-details route.
- The trace-by-id side panel is scoped to the **Trace list** tab only; switching to a chart tab hides it.
- Chart-tab facet controls reuse the trace-list sidebar pattern, each row acting as a toggle. Unselected facets are muted and hide their color pill while keeping the layout space reserved so rows stay aligned.

### Shared package boundary
- Move reusable Signals UI/data/chart logic into `packages/playground-ui/src/ee/signals` behind an explicit EE boundary; the app keeps only route wrappers and breadcrumb glue.
- Share `ScatterPlotChart` and `ThreadList` from `@mastra/playground-ui`; the app shell imports them instead of local copies.

### Runtime feature flag
- Gate the Signals route + sidebar entry behind a server-injected runtime flag (`window.MASTRA_SIGNALS_UI` via the `%%MASTRA_SIGNALS_UI%%` placeholder), wired through the CLI Studio HTML injection, the deployer injection config/callsites, and the playground dev `transformIndexHtml`. Defaults off unless `MASTRA_SIGNALS_UI=true`.

## Verification
- pnpm --filter ./packages/playground-ui test
- pnpm --filter ./packages/playground-ui typecheck
- pnpm build:playground-ui
- pnpm --filter ./packages/playground test
- pnpm --filter ./packages/deployer test
- pnpm --filter mastra test src/commands/studio/studio.test.ts
- pnpm build (full monorepo) — passing

## ELI5
This change swaps the old Topics observability area for a new Signals experience, but only when a feature flag is turned on. It also moves the reusable UI pieces into a shared package so the rest of the app can use them too.

### What changed
- Replaced the Topics observability flow with Signals routes, pages, and navigation.
- Added a Signals overview page and a Signals detail page with trace list/chart tabs, cluster filtering, and trace detail support.
- Added shared UI building blocks like `ScatterPlotChart` and updated `ThreadList` for reuse.
- Moved the Signals/Topics EE logic into `packages/playground-ui` and re-exported it through package barrels.
- Added a runtime flag, `MASTRA_SIGNALS_UI`, to gate the Signals route and sidebar entry.

### Flag wiring
- The flag is injected into Studio HTML as `window.MASTRA_SIGNALS_UI`.
- It’s passed through the CLI Studio server, deployer HTML injection, Vercel deployer config, and playground Vite HTML transform.
- It defaults to off unless `MASTRA_SIGNALS_UI=true`.

### Tests and fixtures
- Added/updated tests for the new Signals UI, shared chart component, thread list behavior, and HTML flag injection.
- Added shared fixtures for traces, signals, datasets, and tool-mock reports.


packages/playground-ui/src/ee/signals/components/signal-details-page.tsx
```diff
@@ -0,0 +1,273 @@
...
+  selectedTraceId,
+  selectedChartClusterIds,
+  activeTab,
+  onActiveTabChange,
+  onClusterSelect,
+  onChartClusterToggle,
+  onTraceSelect,
+}: SignalClusterTabsProps) {
+  return (
+    <Tabs<SignalTab>
```
damien-schneider, line 177: Might need a fix on this 😅

packages/playground/src/domains/agents/components/chat-threads.tsx
```diff
@@ -10,7 +10,7 @@ import {
   ThreadListItems,
   ThreadListNewItem,
   ThreadListSeparator,
-} from '@/components/thread-list';
+} from '@mastra/playground-ui';
```
damien-schneider: We might want to move to de-barrel imports in those

damien-schneider (commented):
Can you add / update stories (especially for the scatterPlotChart) please

packages/playground-ui/src/ee/signals/components/signal-details-page.tsx
mfrachet: this is how we do generics in react haha

damien-schneider (approved):

PR #18139, opened 2026-06-18
fix: Add execCommand fallback for clipboard copy in strict-permission browsers

### LINKED ISSUE #18112: Unable to copy message in Studio
This issue was created from Discord post 1516917197715279882:

[Open in Discord](https://discord.com/channels/1309558646228779139/1516917197715279882)

For some reason I'm unable to use the copy button at the bottom of any conversation and have it go to my clipboard. I'm not sure if it's a browser thing, I'm using the ARC browser, however, I believe it used to work and now it does not.

---

Screenshot 1

okxint:
Copy button on assistant messages silently failed in Arc and browsers with strict clipboard permissions. `navigator.clipboard.writeText` rejects when `clipboard-write` isn't granted — the catch only showed a toast, nothing ended up in the clipboard.

Added a `copyViaExecCommand` fallback: on permission rejection, it creates an off-screen textarea, selects the content, calls `document.execCommand('copy')`, and cleans up. Error toast only shows if both paths fail.

Closes #18112

## ELI5

The copy button on chat messages was broken in some stricter browsers like Arc. The fix tries the modern way to copy first, and if the browser says "no," it switches to an older backup method that still works. Now users can copy messages no matter which browser they're using.

### Changeset
Added a new changeset entry for `@mastra/playground-ui` documenting a `patch` version bump for fixing the copy button in browsers with strict clipboard permissions.

### Hook Refactoring (`useCopyToClipboard`)
Refactored the clipboard functionality to implement a two-tier copy strategy:

**Primary mechanism:** Uses the modern `navigator.clipboard.writeText()` API as the first approach.

**Fallback mechanism:** Added a new `copyViaExecCommand` helper function that:
- Creates an off-screen textarea element
- Populates it with the text to copy
- Selects the text and invokes `document.execCommand('copy')`
- Cleans up the temporary element via a `finally` block

**Improved error handling:** Centralized success behavior (toast notification, `isCopied` state toggle, and 2-second timeout reset) into a reusable `onSuccess` callback. Error toast notifications now only display when both the Clipboard API and the `execCommand` fallback fail, eliminating unnecessary error messages when the fallback succeeds.

The exported function signature remains unchanged, maintaining backward compatibility.

### DIFF
--- a/.changeset/soft-kind-clear.md
+++ b/.changeset/soft-kind-clear.md
@@ -1,0 +1,5 @@
+---
+'@mastra/playground-ui': patch
+---
+
+Fixed copy button silently failing in browsers with strict clipboard permissions (Arc, Firefox). Adds a document.execCommand fallback so copy works without needing the Clipboard API permission.
--- a/packages/playground-ui/src/hooks/use-copy-to-clipboard.ts
+++ b/packages/playground-ui/src/hooks/use-copy-to-clipboard.ts
@@ -6,28 +6,56 @@ type UseCopyToClipboardProps = {
   copyMessage?: string;
 };
 
+function copyViaExecCommand(text: string): boolean {
+  const textarea = document.createElement('textarea');
+  textarea.value = text;
+  // Keep the textarea off-screen so it doesn't flash visible
+  textarea.style.position = 'fixed';
+  textarea.style.top = '-9999px';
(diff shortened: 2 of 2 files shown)


TheIsrael1:
This is now addressed in https://github.com/mastra-ai/mastra/pull/18268, Thanks!

PR #18152, opened 2026-06-18
fix(core): preserve thread metadata written mid-run by processors

### LINKED ISSUE #16216: saveThread upsert replaces thread metadata instead of merging

intojhanurag:
## Description

Thread metadata written during an agent run (by working memory or a processor via `updateThread`) was silently lost when the run finished. At finish, the agent re-saved the thread using a stale in-memory snapshot taken before the run, overwriting anything written during it. The agent now re-reads the latest persisted thread before that save, so mid-run metadata survives. This fixes the bug at the root in `@mastra/core`, so it applies to every storage backend (Postgres, LibSQL, MySQL, etc.).

- Replaced the "re-read only when working memory was used" branch in `#executeOnFinish` (`agent.ts`) and its legacy twin (`agent-legacy.ts`) with an unconditional re-read of the latest thread, falling back to the in-memory snapshot when nothing is persisted.
- This covers both finish-time save sites (new-thread persistence and async title generation) and both working memory and processor writers.
- Skipped the extra read in `readOnlyMemory` mode (`agent.ts`), where nothing is saved, so there is no wasted query.
- Removed the now-unused working-memory detection (`messageListResponses` scan and `isWorkingMemoryToolName` import).
- Added a regression test using a serializing memory mock that models real-database persistence (the default in-memory mock shares object references and cannot reproduce the staleness). The test fails without the fix and passes with it; verified end-to-end against real Postgres and LibSQL.
- Cost: one extra primary-key read per run on memory-backed agents, on a path already dominated by the LLM call.

## Related Issue(s)

Fixes #16216

## Type of Change

- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [ ] Test update

## Checklist

## ELI5

When an agent finishes running, it saves the thread back to storage. Previously it used a “snapshot” of the thread from before the run started, which could overwrite new metadata added during the run. Now, right before saving, it checks what the thread looks like in storage and saves that—so the new metadata sticks.

## Overview

This PR fixes a bug where thread metadata written during an agent run was silently lost at completion. The overwrite happened because end-of-run saving used a stale in-memory snapshot (`threadAfter`) captured before execution began.

## Root Cause

During `#executeOnFinish`, the agent re-saved the thread based on `threadAfter`—often missing or having `undefined` metadata—by calling `memory.createThread()`. That operation upserted/replaced metadata in storage, overwriting changes made mid-run by processors or working memory (via `updateThread`).

## Solution

Updated `#executeOnFinish` in both `agent.ts` and `agent-legacy.ts` to re-read the latest persisted thread before the finish-time save:
- In `agent.ts`: unconditionally re-read the latest persisted thread whenever `!readOnlyMemory` and `threadId` are available; fall back to the in-memory snapshot only if no persisted thread exists. Also skips the extra read in `readOnlyMemory` mode.
- In `agent-legacy.ts`: always re-read the latest thread from storage when `threadId` is available (falling back when a re-read isn’t possible).
- Removed the working-memory detection logic that previously determined whether to re-read (including the `messageListResponses` scan) and removed the now-unused `isWorkingMemoryToolName` import.

## Changes Made

- **packages/core/src/agent/agent.ts**
  - Replaced conditional refresh logic with an unconditional persisted-thread re-read at finish (when not `readOnlyMemory`).
  - Removed the `isWorkingMemoryToolName` import.
- **packages/core/src/agent/agent-legacy.ts**
  - Removed working-memory detection logic and changed legacy “after” handling to always re-read the latest persisted thread when `threadId` exists.
- **packages/core/src/agent/__tests__/memory-metadata.test.ts**
  - Added a `SerializingMockMemory` that simulates persistence by `structuredClone`-ing the thread before saving.
  - Regression test: verifies that metadata updates done mid-run by an input processor (via `updateThread`) are preserved and merged correctly when the agent finishes a newly created thread (covering both `generateLegacy` and `generate` paths).
- **.changeset**
  - Added a patch note documenting the metadata-loss fix and referencing issue `#16216`.

## Impact

- Prevents thread metadata loss across storage backends (e.g., Postgres, LibSQL, MySQL) by fixing the root cause in `@mastra/core`.
- Performance: adds one additional primary-key read per run for memory-backed agents on the finish path; `readOnlyMemory` skips the extra read since nothing is persisted.


intojhanurag:
I closed this PR https://github.com/mastra-ai/mastra/pull/18117 approach that was for only one provider .

TheIsrael1 (approved):

PR #18181, opened 2026-06-18
fix(mastracode): improve long-thread rendering

TylerBarnes:
This improves Mastra Code responsiveness in long threads by reducing synchronous chat spacing work during streaming and keeping the initial rendered message window bounded.

Before, spacing reconciliation could repeatedly scan a large rendered chat tree during streaming updates.

After, spacing reconciliation precomputes the next compact tool group in a linear pass, and text-only streaming updates avoid reconciliation unless the component starts participating in spacing.

The startup render window is also capped at 200 messages so opening a large thread stays responsive.

Verified with:
`pnpm --filter mastracode test -- src/tui/render-messages.test.ts src/tui/handlers/__tests__/message.test.ts --bail 1 --reporter=dot`
`pnpm build:mastracode`

## ELI5
This PR makes Mastra Code’s chat feel snappier in long conversations by doing less “re-layout” work while messages stream in. It also limits how much chat history is loaded at startup so opening a huge thread doesn’t slow everything down.

## Overview
Improves responsiveness of the Mastra Code TUI when rendering long chat threads by reducing expensive spacing reconciliation during streaming updates and capping the initial rendered message window.

### 1. Reduced spacing reconciliation work during streaming
- **File:** `mastracode/src/tui/chat-boundary-reconciliation.ts`
- Removed repeated forward scanning by eliminating `findNextSpacingComponentInList`.
- Added a linear precomputation (reverse pass) of `nextCompactToolGroupKeys`, so reconciliation can use `nextCompactToolGroupKeys[i]` for O(1) lookups instead of repeatedly traversing the rendered list.
- Keeps the same spacer/label reconciliation behavior while avoiding repeated list traversal.

### 2. Reconciliation only when needed for streaming updates
- **File:** `mastracode/src/tui/handlers/message.ts`
- Added `createdStreamingComponent` tracking in `handleMessageUpdate`.
- `handleMessageUpdate` now calls `reconcileChatBoundarySpacers(...)` only when:
  - a new streaming component was created, **or**
  - an existing streaming component transitions from “not participating” to “participating” in chat spacing.
- `handleMessageEnd` removes redundant reconciliation for the final assistant update path (when trailing parts are non-empty / end conditions like `aborted` or `error`).

### 3. Bounded initial render window for large threads
- **File:** `mastracode/src/tui/render-messages.ts`
- Increased `STARTUP_MESSAGE_WINDOW_SIZE` from **40** to **200**, so initial/history rendering uses `listActiveMessages({ limit: 200 })` instead of requesting a smaller window.

## Testing
- Updated expectations in **`src/tui/render-messages.test.ts`** for the new `{ limit: 200 }` behavior.
- Verified with the mastracode test suite including **`src/tui/handlers/__tests__/message.test.ts`**.
- Confirmed the `mastracode` package build completed successfully.

## Formatting / lint fix
- **File:** `mastracode/src/tui/handlers/message.ts`
- Applied the requested Prettier formatting adjustment for a long guard condition in `handleMessageUpdate` (no logic change), resolving the formatting failure reported by the review.
 
## Release Notes
- Added a changeset (**`bright-points-move.md`**) to publish a patch release for `mastracode`, noting improved responsiveness in large threads via faster chat rendering and a bounded initial render window.

### DIFF
--- a/.changeset/bright-points-move.md
+++ b/.changeset/bright-points-move.md
@@ -1,0 +1,5 @@
+---
+'mastracode': patch
+---
+
+Improved Mastra Code responsiveness in large threads by speeding up chat rendering and keeping the initial thread render window bounded.
--- a/mastracode/src/tui/chat-boundary-reconciliation.ts
+++ b/mastracode/src/tui/chat-boundary-reconciliation.ts
@@ -29,14 +29,6 @@ export function insertChatComponentWithBoundarySpacing(
   reconcileChatBoundarySpacers(chatContainer);
 }
 
-function findNextSpacingComponentInList(components: Component[], index: number): Component | undefined {
-  for (let i = index + 1; i < components.length; i++) {
-    const child = components[i];
-    if (child && getChatSpacingKind(child)) return child;
-  }
-  return undefined;
-}
-
 /**
  * Rebuild the spacing layout for a chat container.
  *
@@ -56,6 +48,16 @@ export function reconcileChatBoundarySpacers(chatContainer: Container): void {
   // where possible, reducing object churn.
   const spacerPool = children.filter(isChatBoundarySpacer);
   let poolIndex = 0;
+
+  const nextCompactToolGroupKeys = new Array<string | undefined>(components.length);
+  let nextSpacingComponentGroupKey: string | undefined;
+  for (let i = components.length - 1; i >= 0; i--) {
+    nextCompactToolGroupKeys[i] = nextSpacingComponentGroupKey;
+    const component = components[i];
+    if (component && getChatSpacingKind(component)) {
+      nextSpacingComponentGroupKey = (component as CompactToolGroupingParticipant).getCompactToolGroupKey?.();
+    }
+  }
 
   const nextChildren: Component[] = [];
   let previousCompactToolGroupKey: string | undefined;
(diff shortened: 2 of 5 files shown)


CalebBarnes (approved):
Reviewed — LGTM, approving.

The three changes are each well-scoped:
- `reconcileChatBoundarySpacers` going O(n²) → O(n) via the reverse-pass `nextCompactToolGroupKeys` precompute is the big win, and it preserves the existing per-component logic exactly.
- Gating the update-path reconcile behind `createdStreamingComponent` / the non-participant→participant transition is the right condition — text-only growth on an already-rendered component no longer forces a full reconcile.
- The `handleMessageEnd` reconcile removal looks safe: `getChatSpacingKind()` is purely a function of content presence, the trailing-parts branch always updates with non-empty content (so participation can't flip), and the empty case skips `updateContent`. The transition that matters is already covered in `handleMessageUpdate`.

Verified locally (HEAD `9f6e22ae`): full mastracode unit suite (1319) passed, `tsc --noEmit` clean, `pnpm build:mastracode` succeeded, eslint on changed files clean.

CalebBarnes:
Heads up — the **Lint / Format** check is failing on this PR (not eslint, which passes). Prettier flagged one file:

- `mastracode/src/tui/handlers/message.ts`

It's just the long guard line in `handleMessageUpdate` (around line 524) that needs wrapping — Prettier wants:

```ts
if (
  createdStreamingComponent ||
  (!wasSpacingParticipant && state.streamingComponent.getChatSpacingKind() !== undefined)
) {
  reconcileChatBoundarySpacers(state.chatContainer);
}
```

Running `pnpm prettier --write mastracode/src/tui/handlers/message.ts` fixes it. No logic change — purely formatting.

PR #18182, opened 2026-06-18
fix(core): restore state signals without thread scans

TylerBarnes:
This fixes long-thread state signal restore by loading tracked signal messages directly by id instead of scanning thread history.

Before:
```ts
await memoryStore.listMessages({
  threadId,
  resourceId,
  perPage: false,
});
```

After:
```ts
await memoryStore.listMessagesById({ messageIds: trackedSignalIds });
```

If tracked ids or id-based lookup are unavailable, we now fall back to local in-memory state instead of doing an unbounded thread scan.

Verified with:
`pnpm --filter ./packages/core test -- src/processors/runner.test.ts --bail 1 --reporter=dot`
`pnpm --filter ./packages/core check`

## ELI5

Imagine you have a very long list of messages and need to find a few specific ones. The old way was to look through every single message in the list one by one until you found what you needed. The new way is much smarter: it says "give me these specific messages by their ID" instead, skipping all the others. This makes it much faster when threads get really long.

## Overview

This PR improves the performance of state signal restoration in long-running threads by eliminating expensive full-thread message scans. Instead of iterating through all messages in a thread's history, the system now uses a targeted ID-based lookup to fetch only the tracked signal messages it needs.

## Changes

**Core Logic Update** (`packages/core/src/agent/state-signals.ts`)
- Modified `resolveStateSignalHistory` to collect tracked signal message IDs from `tracking.activeCopies` and `tracking.lastSnapshotSignalId` upfront
- Replaced `memoryStore.listMessages({ threadId, resourceId, perPage: false })` (full thread scan) with `memoryStore.listMessagesById({ messageIds })` (targeted lookup)
- Removed the `resourceId` parameter from the function signature since it's no longer needed for the more efficient ID-based lookup
- Graceful degradation: if `listMessagesById` is unavailable or tracking yields no IDs, the function returns the local in-memory state without attempting expensive fallbacks

**Caller Update** (`packages/core/src/processors/runner.ts`)
- Removed the now-unnecessary `resourceId` argument from the `resolveStateSignalHistory` call

**Test Updates** (`packages/core/src/processors/runner.test.ts`)
- Updated mock to provide both `listMessages` and `listMessagesById` methods
- Changed test expectations to verify that only `listMessagesById` is called (with the specific message IDs) and `listMessages` is not called

## Impact

This optimization prevents performance degradation as thread histories grow longer, especially in long-running conversation scenarios where state signals need to be restored. The changes were verified against existing test suites in the core package.

### DIFF
--- a/.changeset/cuddly-ears-grab.md
+++ b/.changeset/cuddly-ears-grab.md
@@ -1,0 +1,5 @@
+---
+'@mastra/core': patch
+---
+
+Improved state signal restoration so long threads can resume without scanning every message.
--- a/packages/core/src/agent/state-signals.ts
+++ b/packages/core/src/agent/state-signals.ts
@@ -150,14 +150,12 @@ export async function resolveStateSignalHistory({
   messageList,
   memory,
   threadId,
-  resourceId,
   stateId,
   tracking,
 }: {
   messageList: MessageList;
   memory: MastraMemory;
   threadId: string;
-  resourceId: string;
   stateId: string;
   tracking?: StateSignalTracking;
 }): Promise<StateSignalHistory> {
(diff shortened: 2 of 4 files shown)


CalebBarnes (approved):
Reviewed — LGTM. Clean fix for the long-thread restore path.

Swapping the full-thread `listMessages({ perPage: false })` scan for a targeted `listMessagesById({ messageIds })` built from `tracking.activeCopies` + `lastSnapshotSignalId` is the right call, and the fallback to local history avoids any unbounded scan.

A few things I checked that hold up:
- **Ordering** is preserved even though `orderBy: createdAt ASC` was dropped — `mergeStateSignals` and `deriveStateSignalHistory` both re-sort via `sortStateSignals`, so the id-lookup result is fine regardless of adapter return order (pg/libsql actually return `DESC`).
- **Dropping `resourceId`** is safe: `listMessagesById` looks up purely by id, the var is still used elsewhere in `runner.ts`, and `dbMessagesToStateSignals` still gets `threadId` for scoping.
- **Adapter coverage:** `listMessagesById` is an abstract method on `MemoryStoreBase` and is implemented across the storage adapters (confirmed real queries in pg + libsql), so the fast path engages in practice — the `typeof !== 'function'` guard is just defensive cover for old/custom stores.

One non-blocking nit: `trackedSignalIds.size === 0` is effectively dead — the `!tracking.lastSnapshotSignalId` early-return above guarantees the set always has ≥1 entry. Harmless, just noting it.

Verified locally (HEAD `a343a76f`): `runner.test.ts` (75) passed with no type errors, `tsc --noEmit` clean, prettier clean on changed files.

PR #18191, opened 2026-06-19
fix(core): prevent duplicate stream events when observing an in-progress run

### LINKED ISSUE #18148: Durable & Evented agents: observe() delivers duplicate stream events when attaching to an in-progress agent.stream() run
### Describe the Bug

When `eventedAgent.stream(prompt, { runId })` is started fire-and-forget and a separate request calls `eventedAgent.observe(runId, { offset: 0 })` a short time later, the consumer's `output.textStream` delivers the initial text chunks **twice** - interleaved with themselves - before settling into normal single delivery.

I observed it during MCP App development: one HTTP request starts the agent, a second request (e.g. a rendered MCP App widget) attaches to the stream. The duplicated prefix makes the output unusable without post-processing.

**What's duplicated:** only the events already published before `observe()`'s subscription was established. Events published after the subscription is live arrive once. The assistant message in `mastra_messages` is stored correctly — the duplication is only in `output.textStream` on the consumer side.

Reproduces with both `createDurableAgent` and `createEventedAgent`.

### Steps To Reproduce

See the minimal reproduction repo: https://github.com/andreykrupskii/mastra-events-dedup-bug

The script (`src/index.ts`) does three things:
1. Calls `agent.stream(prompt, { runId })` fire-and-forget (producer)
2. Waits `DELAY_MS` ms to simulate the gap before the consumer connects
3. Calls `agent.observe(runId, { offset: 0 })` and pipes `output.textStream` to stdout (consumer)

```
npm install
ANTHROPIC_API_KEY=… npm start
```

The duplicate prefix appears reliably at `DELAY_MS=2000`. At `DELAY_MS=0` it disappears because the subscription wins the race before any token is published.

https://github.com/andreykrupskii/mastra-events-dedup-bug

### Expected Behavior

`output.textStream` yields each text-delta exactly once, in order. The stream output should match the assistant message stored in `mastra_messages`.

### Environment Information

```block
- `@mastra/core`: 1.43.0 (also reproduces on 1.39.0, 1.40.0, 1.42.0)
- Node.js: v24.14.1
- LLM Provider: Anthropic (`claude-opus-4-8`)
- Single Node.js process (no separate worker/server)

Sample stdout at `DELAY_MS=2000`:

TheThe night sky has captivated human imagination since the dawn of our species. When night sky has captivated human imagination since the dawn of our species. When darkness falls…


First word is doubled (TheThe). From night sky onward the stream is clean.
```

### Verification

Sandy-1711:
## Description

Attaching to a running durable/evented agent with `agent.observe(runId)` (or reconnecting to a stream with replay) could deliver the buffered text twice before settling into normal single delivery. This fixes both causes so a late observer receives each event exactly once.

## What you saw

A producer runs an agent fire-and-forget; a second request attaches to watch it:

```ts
agent.stream(prompt, { runId });                              // producer
const { output } = await agent.observe(runId, { offset: 0 }); // consumer attaches later
for await (const t of output.textStream) process.stdout.write(t);
```

The consumer's `textStream` came back with the buffered portion duplicated and interleaved (`"...a spectacleAs daylight surrenders... a spectacle of profound depth..."`), then corrected itself once it caught up. The stored assistant message was always fine — only the observer's stream was wrong.

## Why it happened

There were two separate problems, and the reported repro hit the first.

### Cause 1 — the agent double-wraps an already-caching pubsub (primary)

The repro gives Mastra a pubsub that already does caching:

```ts
const cache = new InMemoryServerCache();
const mastra = new Mastra({ pubsub: withCaching(new EventEmitterPubSub(), cache), cache });
```

On registration the agent adopts `mastra.pubsub` as its transport and then wraps it **again** in its own `CachingPubSub`, sharing the same `cache` and key prefix. Now two caching layers write to the same list with the same counter, so every event is stored twice.

This is easy to see in isolation. Publishing 3 chunks through that same double-wrapped topology and then reading the cache back gives (3 chunks in, 6 stored):

```
historyLength:  6
historyChunks:  [0, 0, 1, 1, 2, 2]   // each chunk stored twice
historyIndices: [0, 1, 2, 3, 4, 5]   // the two copies even get different indices
replayedCount:  6                    // so a late observer receives 6 events, not 3
```

`observe()` replays that doubled list — which is why the whole buffered prefix comes back duplicated, and why it never happens when there's no buffered history yet (e.g. `DELAY=0`).

### Cause 2 — replay/live dedup compared a non-stable field (secondary)

Independently of the wrapping, `CachingPubSub` deduplicates the brief overlap between cache replay and live delivery by `event.id`. But the transport regenerates `id` on publish, so the cached copy and the live copy of the *same* publish carry different ids. The same single event, compared as it sits in the cache vs as it's delivered live to a subscriber:

```
cache copy:  { id: "…0001", index: 0 }   // id the cache stored it under
live  copy:  { id: "…0002", index: 0 }   // id the transport delivered it with
```

Deduping on `id` can never match these, so the overlapping event slips through twice. This is a narrower race than Cause 1, but it's a real defect on the single-cache path too. Note that `index` is identical across both copies.

## The fix

1. **Don't double-wrap.** If the inner pubsub is already a `CachingPubSub`, reuse it instead of adding a second caching layer. In `#ensurePubsubInitialized` (`packages/core/src/agent/durable/durable-agent.ts`). This removes the duplicate storage behind Cause 1.
2. **Dedupe on `index`, not `id`.** `index` is assigned by `CachingPubSub` and preserved across every transport, so both copies of an event share it. In the `dedupKey` helper (`packages/core/src/events/caching-pubsub.ts`), used by `subscribeWithReplay` and `subscribeFromOffset`. Events without an index fall back to `id` and can't be replay/live duplicated, since they're never cached.

## Testing

- `packages/core/src/agent/durable/__tests__/no-double-wrap-pubsub.test.ts` — an agent handed a `CachingPubSub` via `mastra.pubsub` reuses it (each event cached once, not twice); a plain pubsub is still wrapped exactly once. Fails before the fix (the agent creates a second wrapper).
- `packages/core/src/events/caching-pubsub.test.ts` — forces the replay/live race for `subscribeWithReplay` and `subscribeFromOffset` by publishing an event in the window between the live subscribe and the history replay, then asserts single delivery. The previous dedup test never created a live re-delivery, so it passed without actually exercising the path.
- Verified end-to-end against the original repro: doubled before the fix, clean after.

Full `caching-pubsub`, `durable-agent`, and `resumable-streams` suites pass.

## Risk

`#ensurePubsubInitialized` only changes behavior when the inner pubsub is already a `CachingPubSub`; the common path (a plain inner pubsub) is unchanged. The dedup change is contained to `CachingPubSub`'s two replay methods plus a private helper. No public API changes; backward compatible.

## Related issue(s)

Fixes #18148

## Type of change

- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [ ] Test update

## Checklist

## ELI5

When you observe an agent working in real-time, it sends you its buffered past messages followed by new live messages. The bug was that the buffered messages appeared twice before switching to single delivery. This happened because the system was accidentally creating two layers of event buffering instead of one, and the deduplication logic couldn't recognize when the same event appeared in both buffers because each time got a new ID.

## Overview

This PR fixes a bug where observing an in-progress durable or evented agent run delivers buffered text events twice before settling into normal single delivery. The issue occurs when calling `agent.observe(runId)` on an agent that's already in the middle of streaming, causing the cached event history to be duplicated in the output.

## Root Causes & Fixes

**Double-wrapping of CachingPubSub (Primary Cause)**

When a `CachingPubSub` is passed to `new Mastra({ pubsub })`, the agent was adopting it as its transport and then wrapping it again in its own `CachingPubSub`, both sharing the same underlying cache. This caused every event to be stored twice with consecutive indices. When `observe()` replayed the cached history, it delivered the entire buffered prefix duplicated.

*Fix:* In `DurableAgent`'s `ensurePubsubInitialized`, detect if the inner pubsub is already a `CachingPubSub` and reuse it instead of adding a second caching layer.

**Unstable Deduplication Keys (Secondary Cause)**

`CachingPubSub` deduplicates the overlap between cache replay and live delivery using `event.id`. However, inner pubsub implementations regenerate `id` on publish, so the cached copy and live copy of the same event carried different ids, breaking deduplication.

*Fix:* In the `dedupKey` helper, deduplicate on the stable sequential `index` (assigned by `CachingPubSub` and preserved across transports) instead of `id`.

## Changes

- **caching-pubsub.ts**: Updated replay/live deduplication logic to use a stable `dedupKey` derived from `event.index` with fallback to `event.id`. Both `subscribeWithReplay` and `subscribeFromOffset` now properly deduplicate events during the replay/live overlap window.

- **durable-agent.ts**: Modified lazy pubsub initialization to detect and reuse existing `CachingPubSub` instances instead of wrapping them again, preventing duplicate event buffering.

- **caching-pubsub.test.ts**: Added regression tests covering replay/live race conditions for both standard and offset-based subscriptions to ensure events aren't duplicated during the overlap window.

- **no-double-wrap-pubsub.test.ts**: Added test suite validating that when a `CachingPubSub` is provided to `Mastra`, the agent correctly reuses it without double-wrapping, and that history doesn't get duplicated.

- **.changeset/fluffy-spies-matter.md**: Changelog entry documenting the fixes.

## Backward Compatibility

Changes are fully backward compatible. The common path (plain inner pubsub) remains unchanged, there are no public API modifications, and both fixes apply transparently.


intojhanurag (approved):

intojhanurag (changes requested):
Hey @Sandy-1711 , move the getHistory() assertion in resume-api.test.ts to before the cleanup() call. It is failing . And also review the CI failure properly .

Sandy-1711:
Thanks @intojhanurag — fixed in 3a743d4474.

Moved the `getHistory()` assertion in `resume-api.test.ts` to before `initialCleanup()`. Root cause: this PR makes the durable agent reuse an externally provided `CachingPubSub` instead of double-wrapping it, so `cleanup()` → `#clearPubsubTopic` now clears the *same* topic the test reads afterward, leaving history empty and failing the `toBeGreaterThan(0)` assertion. Reading before cleanup keeps the cached events intact.

I also reordered the adjacent `should deduplicate events during resume replay` test to replay before `cleanup()` for the same reason — otherwise the topic was cleared and the replay was empty, making the uniqueness check pass vacuously.

On the rest of the CI: the remaining red checks are unrelated to this change — the pinecone store job fails on a missing `apiKey` secret (not available for fork PRs) and the E2E jobs fail on `changeset pre exit can only be run when in pre mode` (release tooling). All durable-agent / caching-pubsub unit suites pass locally.

intojhanurag (approved):

TheIsrael1 (approved):

taofeeq-deru (approved):

PR #18223, opened 2026-06-19
refactor(core): move the agent run engine onto the Session (multi-session phase 1)

abhiaiyer91:
## What this ships

Two steps toward making a Harness Session a self-contained unit of interaction, so one Harness can eventually serve many concurrent sessions (multiplayer hosts — e.g. a Slack channel adapter where each thread is a session).

### 1. The `SessionMachinery` injection boundary

Formalizes the narrow set of Harness-owned capabilities a Session needs to drive an agent run: resolve the current agent, build run/stream options + toolsets + request context, persist token usage, generate ids, open a thread subscription. The Harness injects this into each Session it constructs via `session.setMachinery(...)`, wired in `#wireSession` alongside the existing resolver/store injections.

### 2. The agent run engine moves onto the Session

The stream loop that consumes an agent's event stream — folding chunks into display messages and token usage, driving tool approval/suspension, and finalizing the run — moves out of the Harness into a per-session `SessionRunEngine` owned by the Session and driven through the injected `SessionMachinery`. The pure chunk→message content transforms (signal/notification/state mappers, usage parsing, finish-reason and server-side-fallback descriptions) move to a shared `stream-content` module imported by both the engine and the Harness signal paths.

`resolveToolApproval()` and the suspension-clearing `abort()` also move onto the Session — both are pure session state plus the already-injected category resolver.

## Why

In the multi-user host the run loop, run state, and thread stream are inherently per-session and cannot be shared; they belong on the Session. *How* a run is produced (which agent answers, the config-backed builders) is shared infrastructure the Harness owns and hands to each session. This PR draws that boundary and relocates the engine across it.

## Behavior

Behavior-preserving. The only surface changes are previously-private (`as any`) internals: `harness.session.processStream(...)` and `session.resolveToolApproval(...)` replace the Harness-private equivalents.

## Test plan

- `pnpm --filter mastracode exec tsc --noEmit` — clean
- `pnpm build:core` — succeeds
- 32/32 harness test files pass
- Full e2e suite (`MC_E2E_VITEST_SCENARIOS=all`): **101 passed, 19 skipped, 0 failed**

Stacked on #18213.


packages/core/src/harness/session.ts
```diff
@@ -120,6 +122,58 @@ export interface ThreadDataStore {
...
+  /** The run budget every initial stream and resume must carry (maxSteps, provider fallbacks, …). */
+  buildSharedRunOptions(): Record<string, unknown>;
+  /** Resolve the toolset (built-in harness tools + user/subagent tools) for a run. */
+  buildToolsets(requestContext: RequestContext): Promise<ToolsetsInput>;
+  /** Resolve the effective request context for a run, layering harness defaults. */
+  buildRequestContext(requestContext?: RequestContext): Promise<RequestContext>;
+  /** Persist the session's running token usage to thread metadata. */
+  persistTokenUsage(): Promise<void>;
+  /** Generate a new id (thread ids, message ids) using the host's id strategy. */
+  generateId(): string;
```
wardpeet, line 191: You should add a similar generatId as we have in mastra with like options that people can add their own

wardpeet (approved):

PR #18239, opened 2026-06-19
fix: attribute browser click navigations to agent

TylerBarnes:
This fixes browser state attribution when an assistant click causes navigation. Previously, a `browser_click` could return the new URL, but the state signal still fell back to `user` because the processor did not connect that click result to the URL delta.

Before:
```txt
browser_click -> { success: true, url: "https://mastra.ai/pricing" }
State delta: browser
  changed: user changed active tab URL to https://mastra.ai/pricing
```

After:
```txt
browser_click -> { success: true, url: "https://mastra.ai/pricing" }
State delta: browser
  changed: agent changed active tab URL to https://mastra.ai/pricing
```

It also keeps manual navigations attributed to the user, including when the URL happens to match an older click result. Added regression coverage for click results from current steps, persisted message history, stale click history after a newer browser state signal, and preserving the URL change source in deltas.

Tested with the focused browser processor tests, typecheck, `pnpm build:core`, and a live smoke test against mastra.ai.

## ELI5

When the assistant clicks something and the page URL changes, Mastra needs to remember “the assistant did that,” not blame it on the user. This PR fixes the browser state attribution so URL changes caused by successful assistant clicks are credited to the agent, and it also avoids using old/stale click results to misattribute later manual navigations.

### Browser State Attribution for Assistant Clicks
- Updated the browser processor so when a `browser_click` tool result returns a new `url` that matches the updated `activeUrl`, the delta marks the source as `activeUrlChangeSource: 'agent'` (instead of incorrectly attributing to `user`).

### Stale Click Result Prevention (Manual Navigation Edge Case)
- Added logic so older click history isn’t reused once newer browser state signals exist—preventing cases where a user’s manual navigation could be incorrectly attributed to the agent just because the new URL matched a prior click result.

### Correct Delta Reporting
- Ensured `getChangedBrowserState` includes `activeUrlChangeSource` in the returned `changed` delta when `activeUrl` changes and a source is set.

### Implementation / Test Infrastructure
- Added defensive helpers to find the most recent successful `browser_click` URL by scanning:
  - current `steps` tool results (reverse), and
  - prior non-`<state id="browser">` messages in reverse (fallback)
- Refactored `computeStateSignal` result construction for clearer, more reliable attribution behavior.
- Expanded the `processor.test.ts` coverage by enhancing the `computeStateSignal` test helper to seed message history and adding regression tests for:
  - attribution from current click steps
  - persisted history behavior
  - ignoring stale click history
  - preserving/overriding attribution based on whether the latest click URL matches the `activeUrl`
  - retaining URL change source in deltas

## Testing
- Focused browser processor tests
- Typecheck
- `pnpm build:core`
- Live smoke test against mastra.ai

## Changelog / Formatting
- Added a changeset bump for `@mastra/core` (patch).
- Applied Prettier formatting to the browser processor changes to satisfy repo lint checks.

### DIFF
--- a/.changeset/plain-tools-double.md
+++ b/.changeset/plain-tools-double.md
@@ -1,0 +1,5 @@
+---
+'@mastra/core': patch
+---
+
+Fixed browser state updates to attribute click-driven navigation to the assistant when the browser click result reports the new URL.
--- a/packages/core/src/browser/processor.test.ts
+++ b/packages/core/src/browser/processor.test.ts
@@ -83,19 +83,24 @@ describe('BrowserContextProcessor', () => {
     const createStateArgs = (
       browserCtx?: BrowserContext,
       activeStateSignals: any[] = [],
-      options: { contextWindow?: { hasSnapshot?: boolean }; lastSnapshot?: any } = {},
+      options: {
+        contextWindow?: { hasSnapshot?: boolean };
+        lastSnapshot?: any;
+        steps?: any[];
+        messageListMessages?: MastraDBMessage[];
+      } = {},
     ) => {
       const requestContext = new RequestContext();
       if (browserCtx) requestContext.set('browser', browserCtx);
       return {
         messages: [],
-        messageList: createMockMessageList() as any,
+        messageList: createMockMessageList(options.messageListMessages) as any,
         requestContext,
         state: {},
         abort: vi.fn() as any,
         retryCount: 0,
         stepNumber: 0,
-        steps: [],
+        steps: options.steps ?? [],
         resourceId: 'resource-1',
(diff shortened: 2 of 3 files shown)


CalebBarnes (approved):
**Reviewed — looks good.**

Clean, well-tested fix. Connecting the `browser_click` result URL to the active-URL delta (and propagating `activeUrlChangeSource` through `getChangedBrowserState`) is the right approach, and the "stop at last browser-state signal" guard for stale clicks is sound. Ran locally on `25700d3a58`: 28 tests pass, tsc/eslint/prettier all clean — the formatting commit clears the Lint failure.

LGTM.

PR #18240, opened 2026-06-19
fix(mastracode): expand ${VAR} env references in MCP server headers

### LINKED ISSUE #18177: mastracode: ${ENV_VAR} not expanded in HTTP MCP server headers
## Summary

Environment variable references like `${MY_API_KEY}` in `.mastracode/mcp.json` HTTP server headers are sent as literal strings instead of being expanded. This breaks any HTTP MCP server that authenticates via a header-based API key.

## Config that fails

\`\`\`json
{
  "mcpServers": {
    "my-service": {
      "url": "https://api.example.com/mcp",
      "headers": {
        "x-api-key": "${MY_API_KEY}"
      }
    }
  }
}
\`\`\`

The string `"${MY_API_KEY}"` is sent verbatim as the header value, regardless of whether `MY_API_KEY` is set in the environment.

## Error

\`\`\`
Failed to connect to MCP server my-service: Error: Streamable HTTP error: Error POSTing
to endpoint: {"error": "Invalid API key"}
\`\`\`

The error is confusing because the key is correctly exported in the shell — the problem is that mastracode never resolves the reference.

## Expected behavior

`${VAR}` (and optionally `$VAR`) in header values should be expanded from the process environment at connection time, matching how Claude Code handles `.mcp.json` headers.

## Workaround

Use `mcp-remote` as a stdio wrapper so the shell expands the variable:

\`\`\`json
{
  "mcpServers": {
    "my-service": {
      "command": "sh",
      "args": ["-c", "exec npx -y mcp-remote https://api.example.com/mcp --header \"x-api-key: $MY_API_KEY\" --transport http-only"]
    }
  }
}
\`\`\`

Related: #13997 (dynamic auth for HTTP MCP servers)

## Component

- CLI (Mastra Code / MCP config)

## Verification

- [x] Searched existing issues — not a duplicate

VihaanAgarwal:
## Description

`${VAR}` references in MCP server HTTP headers were never resolved. `validateConfig` copied the `headers` object through verbatim, so a value like `"x-api-key": "${MY_API_KEY}"` in `mcp.json` was sent as the literal string. Header-authenticated HTTP servers then failed with a confusing `Invalid API key`, even though the variable was correctly exported in the shell.

This expands `${VAR}` and `${VAR:-default}` references in header values from the environment when the config is loaded, matching how Claude Code reads `.mcp.json`. A reference to an unset variable with no default expands to an empty string. Non-`${...}` values, and values without references, are left untouched.

The expansion is scoped to header values, which is what the issue reports. The new `expandEnvVars` helper is exported so the same resolution can be reused for other config fields later if needed.

## Related issue(s)

Fixes #18177

## Type of change

## Checklist

## ELI5
If your MCP server reads an API key from an HTTP header, you can put `${MY_API_KEY}` (or `$MY_API_KEY`) in `.mastracode/mcp.json`. This PR fixes a bug where those values were sent as the literal text `${...}` instead of being replaced with the real environment variable when the config is loaded.

### Core Fix: Expand env vars inside HTTP header values
- Added an exported helper `expandEnvVars(value, env?)` that resolves:
  - `${VAR}`
  - `${VAR:-default}` (uses `default` when `VAR` is unset/empty)
  - bare `$VAR`
- `${VAR}` and `$VAR` expand to `''` when the variable is unset/empty and no `:-default` is provided.
- Only valid env references are expanded; a `$` not followed by an identifier character (e.g., `$5`) is left unchanged.
- Added an internal `expandHeaderEnvVars()` to apply `expandEnvVars` to string-valued entries in an MCP HTTP server `headers` object.
- Updated `validateConfig()` so MCP HTTP server header values are expanded at config-load time rather than sent verbatim.

### Tests
- Updated `mastracode/src/mcp/__tests__/config.test.ts` with:
  - Assertions that `${VAR}` inside HTTP header values is expanded using `process.env` during validation.
  - Unit tests for `expandEnvVars` covering `${VAR}`, embedded substitutions, `${VAR:-default}`, unset-to-`''` behavior, bare `$VAR` expansion, and the “invalid `$...`” pass-through case.

### Changeset
- Added `.changeset/mcp-header-env-expansion.md` documenting that MCP HTTP header value interpolation now supports `${VAR}`, `${VAR:-default}`, and bare `$VAR`, expanded from environment variables at config-load time.


VihaanAgarwal:
Addressed the CodeRabbit feedback:

- Dropped the unused `afterEach` import (lint) and made the header test restore the prior `MC_TEST_MCP_KEY` value in a `finally` instead of always deleting it, in the previous commit.
- Added bare `$VAR` expansion in this commit. The issue's expected behavior lists both `${VAR}` and `$VAR`, matching how Claude Code reads `.mcp.json`. A `$` not followed by an identifier is left untouched, so values like `$5` pass through unchanged. Covered by new tests.

All 31 tests in `config.test.ts` pass locally.

wardpeet (approved):

PR #18251, opened 2026-06-20
chore: add Studio Vercel preview example

damien-schneider:
## Summary
- add a private `examples/studio-preview` Mastra project for Vercel PR previews
- configure the Vercel deployer with Studio enabled and a minimal working agent/tool
- make the example follow the examples validator contract: no workspace deps, local pnpm overrides, local lockfile
- add the missing `examples/inngest/pnpm-lock.yaml` so the examples validator passes

## Testing
- `node .github/scripts/validate-examples.js`
- `corepack pnpm@10.29.3 --dir examples/studio-preview typecheck`
- `corepack pnpm@10.29.3 --dir examples/studio-preview build`

## Notes
- no changeset: this only adds a private example package
- CodeRabbit local review was attempted, but the CLI is not signed in for non-interactive auth (`coderabbit doctor` reports Authentication: Not signed in)

## ELI5
This PR adds a small “Studio Preview” demo app you can deploy on Vercel for each pull request. When you open it, Mastra Studio and a tiny API let you check whether the preview looks “ready,” without needing heavy setup.

---

## Overview
- Adds a new **private** Vercel PR-preview example at `examples/studio-preview` with **Mastra Studio enabled** (via the Vercel deployer) and a minimal agent + tool.
- Updates the example to match the examples validator expectations for self-contained behavior (no workspace dependency reliance; uses a local lockfile and pnpm overrides).

### New example: `examples/studio-preview`
- **Vercel preview setup**
  - `vercel.json` installs with `pnpm install --frozen-lockfile --ignore-workspace` and builds with `pnpm build`.
  - `scripts/ignore-vercel-build.sh` skips preview builds for `VERCEL_ENV=production`, and otherwise only builds for configured preview refs.
- **Docs + config**
  - `README.md` documents the “Studio Preview” target, expected routes (Studio at `/`, API under `/api/*`), required env vars, and reviewer workflow/security notes.
  - `.env.example` includes placeholders for `OPENAI_API_KEY` and `MASTRA_PREVIEW_MODEL`.
  - `.gitignore` ignores `.env`, `.mastra`, `.vercel`, and `node_modules`.
- **TypeScript**
  - `tsconfig.json` enables strict typechecking (no emit) with bundler-style module resolution.

### Minimal Mastra Studio + QA API
- **Agent:** `studioPreviewAgent`
  - Uses `process.env.MASTRA_PREVIEW_MODEL` for model selection.
  - Registers the preview status tool.
- **Tool:** `previewStatusTool`
  - Validates input `area` (`studio | agent | api | vercel`) with Zod.
  - Returns `{ area, status: 'ready', checks: string[], generatedAt }`.
- **App wiring:** `src/mastra/index.ts`
  - Exports a configured `mastra` instance with `studio: true`, `VercelDeployer` configured for Studio preview, and Swagger/OpenAPI enabled for the server build output.

## Verification
- `node .github/scripts/validate-examples.js`
- `corepack pnpm@10.29.3 --dir examples/studio-preview typecheck`
- `corepack pnpm@10.29.3 --dir examples/studio-preview build`

## Notes
- No changeset was added because this PR introduces a private example package.
- A CodeRabbit local review attempt could not be completed due to missing non-interactive CLI authentication.


examples/studio-preview/scripts/ignore-vercel-build.sh
```diff
@@ -0,0 +1,26 @@
+#!/usr/bin/env bash
+
+set -u
+
+allowed_branches=(
```
mfrachet: do we need that? Shouldn't we say vercel to only do that for prev envs?

examples/inngest/pnpm-lock.yaml
mfrachet, line 1: not sure you need that

examples/studio-preview/package.json
```diff
@@ -0,0 +1,37 @@
...
+  "name": "examples-studio-preview",
+  "type": "module",
+  "private": true,
+  "version": "0.0.0",
+  "scripts": {
+    "build": "pnpm build:deps && pnpm mastra:build",
+    "build:deps": "turbo --cwd ../.. build --filter mastra --filter @mastra/deployer-vercel",
+    "dev": "pnpm build:deps && pnpm mastra:dev",
+    "mastra:build": "node ../../packages/cli/dist/index.js build",
+    "mastra:dev": "node ../../packages/cli/dist/index.js dev",
```
mfrachet, line 11: Would that work? I thought we only wanted to deploy the current example project, and not the entire mastra OSS

examples/studio-preview/README.md
```diff
@@ -0,0 +1,61 @@
...
+Remove or broaden that branch check when the project should create previews for other PR branches.
+
+Add these environment variables for Preview deployments:
+
+```text
+OPENAI_API_KEY=...
+MASTRA_PREVIEW_MODEL=__AI_SDK_OPENAI_MODEL_BASE__
+```
+
+Vercel will use the generated `.vercel/output` folder. Studio is served at `/`, and the Mastra API is served under `/api/*`.
```
mfrachet, line 50: I don't think we put the playground in the build artifacts

mfrachet (changes requested):
Just to make sure we're on the same page

examples/studio-preview/scripts/ignore-vercel-build.sh
damien-schneider: It’s the “less risky” option I found because the build may require several iterations to succeed, and I can’t configure the project to work only on a dedicated branch (so it is just to test first on specific branch I've created). I set it up to run on a specific branch so it won’t fail or trigger on others. If you have another approach in mind, let me know, I tried a few things but didn’t find a perfect solution better than that 😄

examples/studio-preview/package.json
damien-schneider: Yes it should work, the root turbo step only builds the linked workspace deps used by this example so the preview runs against the current branch (I've renamed the scripts in the package.json to make it clearer)

examples/studio-preview/README.md
damien-schneider: For the normal Mastra output yes, here relying on VercelDeployer({ studio: true }) copies the studio assets into .vercel/output/static. I verified locally and the assets are generated, with /api routed to the function and other routes falling back to Studio.
So it should work on Vercel, (but not 100% sure there until we try on vercel of course 😅)

examples/inngest/pnpm-lock.yaml
damien-schneider: Not needed for the Studio preview itself but the examples validator fails because `examples/inngest` was already missing a `pnpm-lock.yaml`. I added it to keep CI green otherwise this PR fails validation for an existing example issue

damien-schneider: Saying that I didn't review the CI itself if it makes sense, doing it right now

damien-schneider: Correction / final state: this `examples/inngest/pnpm-lock.yaml` file is not needed for the Studio preview and has been removed from the PR. I initially thought it was required because the examples validator reported the existing missing lockfile, but after rechecking the validator and fixing `examples/studio-preview` with its own lockfile + overrides, CI passes without adding anything under `examples/inngest`. The only lockfile we should keep here is `examples/studio-preview/pnpm-lock.yaml`, because the Vercel preview installs that example as a standalone project with `pnpm install --frozen-lockfile --ignore-workspace`.

examples/studio-preview/scripts/ignore-vercel-build.sh
mfrachet: let's remove it and only check previews on vercel, we'll be fine 😊

examples/studio-preview/README.md
mfrachet: let's try it out, we can still remove it anyways haha

mfrachet (approved):

examples/studio-preview/scripts/ignore-vercel-build.sh
damien-schneider: Done. Removed the Vercel ignore script and the `ignoreCommand` from `vercel.json` in `72de0ef`, and updated the README to say that preview-only behavior should be controlled in the Vercel project settings instead of this repository example.

PR #18253, opened 2026-06-20
fix(observability): emit metrics independently of span export filtering

intojhanurag:
## Description

Auto-extracted metrics (duration, token usage, cost) were silently dropped when spans were filtered via `excludeSpanTypes` or `spanFilter`. Users who excluded span types to reduce per-span costs in platforms like Langfuse unknowingly lost their aggregate metrics.

- Decoupled `emitAutoExtractedMetrics` from the span export guard in `emitSpanEnded` so metrics fire for all valid non-internal spans regardless of export filtering
- Preserved the `isInternal` guard to avoid double-counting with the existing usage rollup mechanism for hidden `MODEL_GENERATION` spans
- Added 4 tests: two proving metrics survive `excludeSpanTypes` and `spanFilter`, one proving no double-count with rollup, one proving no regression for exported spans

## Related Issue(s)

Fixes #15155

## Type of Change

- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [ ] Test update

## Checklist

## ELI5
Some settings can hide individual trace “spans” to save money, but the system also accidentally hid the “summary numbers” those spans produce (tokens, latency, cost). This PR makes sure the summary numbers still get emitted even when the detailed spans are filtered out.

## Summary
This PR fixes a bug where auto-extracted aggregate metrics (duration, token usage, and cost) were only emitted when spans successfully passed span export filtering. When spans were filtered out via `excludeSpanTypes` or `spanFilter`, their derived metrics were silently dropped.

### What changed
- In `emitSpanEnded`, auto-extracted metrics are now emitted independently of whether the span is later exported or dropped by span filtering.
- The `isInternal` guard is preserved to prevent double-counting with the existing usage rollup logic (notably around hidden/excluded `MODEL_GENERATION` spans).
- When `MODEL_GENERATION` is excluded by `excludeSpanTypes`, the span `end()` wrapper captures excluded usage/provider/model information before end attributes are discarded, so token/cost metrics can still be produced.
- Span processing/export flow was refactored so export logic can use either the already-processed span (`getProcessedSpanForExport`) or the raw span, keeping processor behavior consistent.

### Tests added/updated
`observability/mastra/src/span-filtering.test.ts` adds coverage that metrics survive filtering:
- Duration metrics are emitted even when spans are excluded by `excludeSpanTypes` (i.e., absent from exported tracing events).
- Duration metrics are emitted even when spans are dropped by `spanFilter` (i.e., absent from exported tracing events).
- Internal-span/rollup scenarios don’t double-count token metrics.
- Regression coverage confirms exported spans still produce their expected duration metrics.
- Verifies `spanOutputProcessors` (e.g., `SensitiveDataFilter`) affect both exported span metadata and auto-extracted duration metric metadata (e.g., `apiKey` redaction).

### Documentation/Release note
A changeset updates `@mastra/observability` with a patch release documenting this non-breaking bug fix that decouples metric emission from span export filtering.

### DIFF
--- a/.changeset/rich-dingos-matter.md
+++ b/.changeset/rich-dingos-matter.md
@@ -1,0 +1,5 @@
+---
+'@mastra/observability': patch
+---
+
+Fixed auto-extracted metrics (duration, token usage, cost) being silently dropped when spans are filtered via `excludeSpanTypes` or `spanFilter`. Previously, excluding a span type to reduce per-span costs in platforms like Langfuse also suppressed its aggregate metrics. Metrics are now emitted independently of span export filtering.
--- a/observability/mastra/src/instances/base.ts
+++ b/observability/mastra/src/instances/base.ts
@@ -695,23 +695,19 @@ export abstract class BaseObservabilityInstance extends MastraBase implements Ob
    * then routes the exported tracing event through the ObservabilityBus.
    */
   protected emitSpanEnded(span: AnySpan): void {
-    const exportedSpan = this.getSpanForExport(span);
-
-    if (exportedSpan) {
+    // Emit metrics independently of export filtering: users who exclude span
+    // types (e.g. MODEL_STEP) to cut per-span costs should still get duration,
+    // token, and cost metrics. Only skip metrics for invalid or internal spans.
+    if (span.isValid && !(span.isInternal && !this.config.includeInternalSpans)) {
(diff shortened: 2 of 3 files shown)


epinzur (approved):
nice work!

PR #18257, opened 2026-06-20
fix(core): stop SaveQueueManager from silently dropping messages on s…

### LINKED ISSUE #18256: [BUG] SaveQueueManager silently drops conversation messages when storage save fails
### Describe the Bug

The agent's message-persistence pipeline (`SaveQueueManager`) drains unsaved messages from the `MessageList` **before** attempting the storage write, and then swallows any error from `memory.saveMessages()`. The failed messages are never re-queued, the promise returned to callers resolves successfully, and the agent run completes as if everything was persisted. The result is **permanent, silent loss of conversation history** whenever the storage backend has a transient failure (connection drop, constraint violation, timeout, pool exhaustion).

There is a second, related defect in the same class: when `debounceSave()` is called again within the debounce window, the previous timer is cleared but the previous caller's `Promise` is abandoned — its `resolve`/`reject` are never invoked, so any code `await`ing that earlier `batchMessages()` call hangs forever.

`packages/core/src/agent/save-queue/index.ts:60-74` — the error is caught, logged, and the chain **continues as success**:

```ts
private enqueueSave(threadId: string, messageList: MessageList, memoryConfig?: MemoryConfigInternal) {
  const prev = this.saveQueues.get(threadId) || Promise.resolve();
  const next = prev
    .then(() => this.persistUnsavedMessages(messageList, memoryConfig))
    .catch(err => {
      this.logger?.error?.('Error in enqueueSave', { err, threadId });
    })
    .then(() => { ... });
  ...
}
```

`packages/core/src/agent/save-queue/index.ts:95-103` — messages are drained out of the list before the write:

```ts
private async persistUnsavedMessages(messageList: MessageList, memoryConfig?: MemoryConfigInternal) {
  const newMessages = messageList.drainUnsavedMessages();
  if (newMessages.length > 0 && this.memory) {
    await this.memory.saveMessages({ messages: newMessages, memoryConfig });
  }
}
```

`packages/core/src/agent/message-list/message-list.ts:880-885` — `drainUnsavedMessages()` clears the `newUserMessages` / `newResponseMessages` sets immediately, so after a failed save there is nothing left to retry. Callers rely on `flushMessages()` for durability at critical transitions (e.g. `prepare-stream/map-results-step.ts:95` at the end of a stream, and `durable/workflows/steps/tool-call.ts:93` before suspension) — all of them are told "success" even when the write failed.

Suggested fix: re-queue the drained messages on failure (or drain only after a successful write), propagate the error to the `flushMessages()` caller, and make superseded debounce promises settle when the batched save completes.

### Steps To Reproduce

1. Configure an agent with `Memory` backed by any real storage (e.g. `@mastra/pg`).
2. Make `saveMessages` fail once (kill the DB connection mid-run, or monkey-patch `memory.saveMessages` to reject on the first call).
3. Call `agent.generate("hello", { memory: { thread, resource } })`.
4. The call completes normally (no thrown error; only an `error`-level log line). Inspect the thread: the user message and assistant response are missing, and they are never retried — subsequent turns save fine, leaving a hole in the history.
5. For the dangling promise: call `batchMessages()` twice within 100 ms for the same thread and `await` the first call — it never settles.

Reproducible on a fresh `npx create-mastra@latest` project with a memory-enabled agent by forcing `saveMessages` to reject once (see steps above).   https://github.com/mastra-ai/mastra

### Expected Behavior

A failed save should either propagate the error to the caller of `flushMessages()` (so the agent run can surface/retry it) or re-queue the drained messages so the next flush retries them. Every promise returned by `batchMessages()` should settle — superseded debounce calls should resolve when the batched save completes.

### Environment Information

```block
System:
    Memory: 6.14 GB / 15.49 GB
    Container: Yes
    Shell: 5.2.21 - /bin/bash
  Binaries:
    Node: 20.19.6 - /home/abuhaithem/.nvm/versions/node/v20.19.6/bin/node
    Yarn: 1.22.22 - /home/abuhaithem/.nvm/versions/node/v20.19.6/bin/yarn
    npm: 10.8.2 - /home/abuhaithem/.nvm/versions/node/v20.19.6/bin/npm
    pnpm: 10.22.0 - /usr/local/bin/pnpm
```

### Verification

Abuhaithem:
## Description

`SaveQueueManager` drained unsaved messages from the `MessageList` **before** attempting the storage write, and then swallowed any error from `memory.saveMessages()` (`packages/core/src/agent/save-queue/index.ts`). The drained messages were never re-queued, the promise returned to callers resolved successfully, and the agent run completed as if everything was persisted. The result was **permanent, silent loss of conversation history** whenever the storage backend had a transient failure (connection drop, constraint violation, timeout, pool exhaustion):

```ts
private enqueueSave(threadId, messageList, memoryConfig) {
  const next = prev
    .then(() => this.persistUnsavedMessages(messageList, memoryConfig))
    .catch(err => {
      this.logger?.error?.('Error in enqueueSave', { err, threadId }); // swallowed
    })
    .then(() => { ... });
}
```

Callers rely on `flushMessages()` for durability at critical transitions (end of a stream in `prepare-stream/map-results-step.ts`, before suspension in `durable/workflows/steps/tool-call.ts`) — all of them were told "success" even when the write failed, and the messages were already gone from the list so nothing retried them.

There was a second, related defect: when `debounceSave()` was called again within the debounce window, the previous timer was cleared but the previous caller's `Promise` was abandoned — its `resolve`/`reject` were never invoked, so any code `await`ing that earlier `batchMessages()` call hung forever.

### Changes

- **Persist, then clear.** `persistUnsavedMessages()` now reads the unsaved messages with a new `MessageList.getUnsavedMessages()` (no clear) and only drops them via `clearUnsavedMessages()` **after** a successful write. A failed save leaves the messages queued for the next flush. Messages added during the in-flight save are matched by id and remain unsaved.
- **Propagate the error.** `enqueueSave()` returns a result promise that rejects on failure (so `flushMessages()` callers can surface/retry), while the internal per-thread queue chain still settles successfully so a single failure does not stall later saves for the same thread.
- **No more hanging promises.** Debounce callers are tracked per thread and all settle when the batched save completes; a flush that supersedes a pending debounce adopts its waiting callers.
- Add regression tests covering all four behaviors.

No public API changes. The success path is unchanged.

### Before / after (public API behavior)

```ts
// Before: a transient storage failure was swallowed — flushMessages() resolved
// "successfully", the messages were dropped, and nothing retried them.
await agent.generate('hello', { memory: { thread, resource } });

// After: the failure propagates so the run can surface/retry it, and the
// messages stay queued for the next flush instead of being lost.
try {
  await agent.generate('hello', { memory: { thread, resource } });
} catch (error) {
  // handle the transient storage failure
}
```

**Files changed**
- `packages/core/src/agent/save-queue/index.ts` — clear messages only after a successful write; propagate save errors to `flushMessages()` callers; settle superseded debounce promises
- `packages/core/src/agent/message-list/message-list.ts` — add `getUnsavedMessages()` (peek) and `clearUnsavedMessages()` (clear-on-success)
- `packages/core/src/agent/message-list/state/MessageStateManager.ts` — add `clearUnsavedMessages(ids)` to drop specific messages by id
- `packages/core/src/agent/save-queue/save-queue.test.ts` — new regression tests
- `.changeset/save-queue-silent-message-loss.md` — patch changeset

## Related issue(s)

Fixes: #18256

## Type of change

- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [ ] Test update

## How it was tested

- **New regression tests** (`save-queue.test.ts`) assert that:
  - a save failure propagates to `flushMessages()` callers (instead of resolving as success),
  - failed messages are re-queued and persisted by a later flush,
  - a failed save does not stall later saves on the same thread, and
  - superseded debounce promises settle when the batched save completes (instead of hanging).
- Ran `src/agent/save-queue`, `src/agent/message-list`, and `src/agent/__tests__/save-and-errors.test.ts` for `@mastra/core` — all green (490 passed), no type errors.
- `tsc` typecheck (run by Vitest) is clean on the changed files.

## Checklist

- [ ] I have linked the related issue(s) in the description above _(pending — replace #XXXX once the issue is filed)_
- [x] I have made corresponding changes to the documentation (if applicable) _(none needed; internal behavior only)_
- [ ] I have addressed all Coderabbit comments on this PR _(after PR is opened)_

## Diff summary

```diff
-  private async persistUnsavedMessages(messageList, memoryConfig) {
-    const newMessages = messageList.drainUnsavedMessages();
-    if (newMessages.length > 0 && this.memory) {
-      await this.memory.saveMessages({ messages: newMessages, memoryConfig });
-    }
-  }
+  private async persistUnsavedMessages(messageList, memoryConfig) {
+    const newMessages = messageList.getUnsavedMessages();
+    if (newMessages.length > 0 && this.memory) {
+      await this.memory.saveMessages({ messages: newMessages, memoryConfig });
+      // Only clear after a successful write so a failed save can be retried.
+      messageList.clearUnsavedMessages(newMessages);
+    }
+  }

   private enqueueSave(threadId, messageList, memoryConfig): Promise<void> {
     const prev = this.saveQueues.get(threadId) || Promise.resolve();
-    const next = prev
-      .then(() => this.persistUnsavedMessages(messageList, memoryConfig))
-      .catch(err => { this.logger?.error?.('Error in enqueueSave', { err, threadId }); })
+    const result = prev.then(() => this.persistUnsavedMessages(messageList, memoryConfig));
+    const next = result
+      // Swallow on the queue chain only so later saves still run; the error is
+      // propagated to the caller via `result`.
+      .catch(err => { this.logger?.error?.('Error in enqueueSave', { err, threadId }); })
       .then(() => { if (this.saveQueues.get(threadId) === next) this.saveQueues.delete(threadId); });
     this.saveQueues.set(threadId, next);
-    return next;
+    return result;
   }
```

## ELI5

This PR makes sure chat messages don’t “vanish” when the system tries to save them but something goes wrong (like a temporary database/network issue). It also makes sure callers waiting on a debounced save don’t get stuck forever—promises are now correctly settled when the batch finishes.

## Summary

This PR fixes two issues in `SaveQueueManager` that could (1) permanently lose conversation history and (2) cause hanging promises.

### 1) Critical Issue: Silent message loss on storage save failures
Previously, `persistUnsavedMessages()` would remove (drain) unsaved messages from the `MessageList` before calling `memory.saveMessages()`, then swallow any save error—so failed writes meant messages were neither retried nor reported.

**Fix**
- `MessageList` now provides:
  - `getUnsavedMessages()` to read an immutable snapshot of unsaved messages (deep-cloned for persistence).
  - `clearUnsavedMessages(messages)` to clear only the specific snapshot entries that were successfully persisted, using ID comparisons plus `messagesAreEqual` to avoid clearing messages that changed since the snapshot.
- `persistUnsavedMessages()` now:
  - reads unsaved messages via `getUnsavedMessages()` (no clearing up front),
  - clears them only after `memory.saveMessages()` succeeds,
  - leaves them queued when the save fails, enabling retry on the next flush.

### 2) Secondary Issue: Debounced callers could hang indefinitely
When `debounceSave()` was called again inside the debounce window, the old timer was replaced and earlier callers’ promises were not properly settled.

**Fix**
- `SaveQueueManager` now tracks all waiting debounce callers per thread via a `pendingDebounceResolvers` structure.
- When the debounced timer fires, it enqueues the save and resolves/rejects all pending callers with the same result.
- If `flushMessages()` happens while a debounce is pending, the flush adopts and settles the pending debounce waiters using the flush’s `enqueueSave(...)` promise.

### Supporting changes / verification
- `enqueueSave()` was adjusted to return the actual result promise (so failures reject correctly), while the internal per-thread queue chain still settles in a way that prevents one failure from blocking later saves.
  - error propagation from `flushMessages()`,
  - re-queuing after failed saves so later flushes retry,
  - avoiding stalling later saves after an earlier failure,
  - ensuring superseded debounce promises settle (no hangs).
- Added `MessageStateManager.clearUnsavedMessages(ids)` to support selective clearing of persisted snapshot IDs.
- Includes a patch changeset note describing both behavioral fixes (`@mastra/core`).

### DIFF
--- a/.changeset/save-queue-silent-message-loss.md
+++ b/.changeset/save-queue-silent-message-loss.md
@@ -1,0 +1,11 @@
+---
+'@mastra/core': patch
+---
+
+Fixed silent loss of conversation history when saving messages to storage failed.
+
+Previously, if a storage write failed during an agent run (transient connection drop, timeout, constraint violation, pool exhaustion), the unsaved messages were already drained from the message list and the error was swallowed. The agent run completed as if everything was persisted, leaving a permanent hole in the thread history that was never retried.
+
+Now a failed save keeps the messages queued so the next flush retries them, and the error is propagated to the caller of `flushMessages()` so the agent run can surface or handle it. A single failure no longer stalls later saves for the same thread.
+
+Also fixed a related issue where rapidly calling `batchMessages()` within the debounce window left the superseded call's promise hanging forever. Every promise now settles when the batched save completes.
--- a/packages/core/src/agent/message-list/message-list.ts
+++ b/packages/core/src/agent/message-list/message-list.ts
@@ -884,6 +884,26 @@ export class MessageList {
     return messages.map(message => this.transformMessageForTranscript(message));
   }
 
+  /**
+   * Returns the unsaved user/response messages (transformed for persistence)
+   * without clearing them from the unsaved tracking. Pair with
+   * {@link clearUnsavedMessages} to only drop messages once a storage write has
+   * succeeded, so a failed save can be retried instead of silently lost.
+   */
+  public getUnsavedMessages(): MastraDBMessage[] {
+    const messages = this.messages.filter(m => this.newUserMessages.has(m) || this.newResponseMessages.has(m));
+    return messages.map(message => this.transformMessageForTranscript(message));
+  }
+
+  /**
+   * Marks the given messages (matched by id) as saved by removing them from the
+   * unsaved user/response tracking. Messages added after {@link getUnsavedMessages}
+   * was called remain unsaved and will be picked up by the next save.
+   */
+  public clearUnsavedMessages(messages: MastraDBMessage[]): void {
+    this.stateManager.clearUnsavedMessages(new Set(messages.map(m => m.id)));
+  }
+
   private transformToolStateDataForTranscript(data: unknown, phase: 'approval' | 'suspend'): unknown {
     if (!data || typeof data !== 'object') {
       return data;
--- a/packages/core/src/agent/message-list/message-list.ts
+++ b/packages/core/src/agent/message-list/message-list.ts
@@ -889,19 +889,38 @@ export class MessageList {
    * without clearing them from the unsaved tracking. Pair with
    * {@link clearUnsavedMessages} to only drop messages once a storage write has
    * succeeded, so a failed save can be retried instead of silently lost.
+   *
+   * The returned messages are deep-cloned so they are an immutable snapshot of the
+   * content at read time: streaming that appends more parts to the same message id
+   * during the in-flight save cannot mutate the payload being written, and
+   * {@link clearUnsavedMessages} can reliably detect whether the live message
+   * changed since this snapshot.
    */
   public getUnsavedMessages(): MastraDBMessage[] {
     const messages = this.messages.filter(m => this.newUserMessages.has(m) || this.newResponseMessages.has(m));
-    return messages.map(message => this.transformMessageForTranscript(message));
+    return messages.map(message => structuredClone(this.transformMessageForTranscript(message)));
   }
 
   /**
-   * Marks the given messages (matched by id) as saved by removing them from the
-   * unsaved user/response tracking. Messages added after {@link getUnsavedMessages}
-   * was called remain unsaved and will be picked up by the next save.
+   * Marks the given messages (the snapshots returned by {@link getUnsavedMessages})
+   * as saved by removing them from the unsaved user/response tracking. A message is
+   * only cleared when the live message still matches the snapshot that was persisted;
+   * if its content changed while the save was in flight (e.g. streaming appended more
+   * parts to the same id), it stays tracked so the newer content is written by the
+   * next save instead of being silently dropped. Messages added after
+   * {@link getUnsavedMessages} was called also remain unsaved.
    */
   public clearUnsavedMessages(messages: MastraDBMessage[]): void {
-    this.stateManager.clearUnsavedMessages(new Set(messages.map(m => m.id)));
+    const savedIds = new Set<string>();
+    for (const saved of messages) {
+      const live = this.messages.find(m => m.id === saved.id);
+      // Drop from tracking only if nothing changed since the snapshot was taken.
+      // A missing live message means it was already removed, so clearing is safe.
+      if (!live || messagesAreEqual(this.transformMessageForTranscript(live), saved)) {
+        savedIds.add(saved.id);
+      }
+    }
+    this.stateManager.clearUnsavedMessages(savedIds);
   }
 
   private transformToolStateDataForTranscript(data: unknown, phase: 'approval' | 'suspend'): unknown {
--- a/packages/core/src/agent/message-list/state/MessageStateManager.ts
+++ b/packages/core/src/agent/message-list/state/MessageStateManager.ts
@@ -183,6 +183,20 @@ export class MessageStateManager {
   }
 
   /**
+   * Remove specific unsaved user/response messages (matched by id) from the new
+   * message tracking. Used to mark messages as saved only after a successful
+   * storage write, so a failed save can be retried instead of silently dropped.
+   */
+  clearUnsavedMessages(ids: Set<string>): void {
+    for (const message of this.newUserMessages) {
+      if (ids.has(message.id)) this.newUserMessages.delete(message);
+    }
+    for (const message of this.newResponseMessages) {
+      if (ids.has(message.id)) this.newResponseMessages.delete(message);
+    }
+  }
+
+  /**
    * Clear all messages from all sources (but not persisted tracking)
    */
   clearAll(): void {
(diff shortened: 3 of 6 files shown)


wardpeet:
Thanks for taking the time to dig into this and for identifying a real problem in the save queue behavior.

That said, we're going to close this PR rather than move it forward.

`message-list` is one of the most fragile parts of the agent runtime, and we do not want contributors making changes there unless the work is being directly coordinated with the maintainers. We've seen a lot of subtle bugs come from that area, especially around streaming, persistence, message identity, and unsaved-message tracking.

There are also a few concrete concerns with this PR:

- The save-queue issue itself is real: the current destructive `drainUnsavedMessages()` path can drop messages when storage fails.
- The proposed direction of reading unsaved messages non-destructively and only clearing them after a successful write is directionally reasonable.
- However, the implementation changes message-list semantics in a high-risk area and needs maintainer-owned design/review.
- There is also a promise-settlement edge case around `clearDebounce()`: directly clearing a debounce can leave callers waiting on `batchMessages()` promises unless those pending resolvers are explicitly settled or `clearDebounce()` is made private/internal to `flushMessages()`.

If we take a fix here, it should be a small maintainer-led PR limited to the save queue + message-list persistence path, with focused tests for storage failure retry, in-flight streaming mutations, debounce promise settlement, and queue continuation after failure.

Closing this for now.

wardpeet:
Closing per maintainer review. See the comment above for details.

PR #18262, opened 2026-06-20
fix(workflow): enforce FGA on Run resume paths

### LINKED ISSUE #18165: [BUG] Workflow run resume APIs bypass internal FGA check
### Describe the Bug

Workflow execution FGA enforcement is inconsistent between the workflow-level internal execute path and run-level resume paths.

`Workflow.execute()` checks `mastra.getServer()?.fga` and calls `requireFGA` with `workflows:execute` before continuing execution. However, `Run.resume()`, `Run.resumeAsync()`, and `Run.resumeStream()` load the persisted workflow snapshot by `workflowId` + `runId` and continue execution without an equivalent internal FGA check.

The standard HTTP server adapter appears to add route-level FGA for built-in `/workflows/:workflowId/.../resume...` routes, and the handlers also validate stored `resourceId` ownership. The gap is still valid at the core API boundary and for any custom/internal caller that reaches `createRun({ runId }).resume*()` directly or otherwise bypasses the built-in adapter middleware.

Relevant code paths:

- `packages/core/src/workflows/workflow.ts`: `Workflow.execute()` performs `requireFGA` for `workflows:execute`.
- `packages/core/src/workflows/workflow.ts`: `Run._resume()` loads the snapshot and calls `executionEngine.execute(...)` without a matching FGA check.
- `packages/core/src/workflows/evented/workflow.ts`: evented `EventedRun.resume()` has the same snapshot-load-and-execute shape without local FGA enforcement.
- `packages/server/src/server/handlers/workflows.ts`: built-in resume handlers call `_run.resume*({ ...params, requestContext })` after route/ownership checks.

### Steps To Reproduce

1. Configure a Mastra instance with `server.fga` enabled and a provider that denies `workflows:execute` for a given user/workflow.
2. Create and suspend a workflow run.
3. In code, bypass the built-in server route adapter and call the core run API directly:

```ts
const run = await workflow.createRun({ runId: suspendedRunId, resourceId: existingRunResourceId });
await run.resume({ resumeData, requestContext: deniedUserRequestContext });
```

4. Observe that the resume path proceeds to snapshot loading and execution without invoking `fgaProvider.require` / `requireFGA` for the workflow.

The built-in HTTP route path may still be blocked by route-level FGA; this report is about the lower-level core resume boundary being asymmetric with `Workflow.execute()`.

Not available yet. This is a code-level issue found in the current repository; the reproduction path above should be enough to turn into a focused unit test in `packages/core/src/workflows`.

### Expected Behavior

All workflow execution entry points should enforce the same `workflows:execute` FGA requirement when an FGA provider is configured, including resume APIs.

Possible fixes:

- Add a shared workflow FGA helper and call it from `Run._start()` / `Run._resume()` and the evented equivalents, or
- Route all public/custom resume execution through an existing workflow-level method that performs the check.

The behavior should continue to fail closed when FGA is configured and no authenticated user is available, matching the existing `requireFGA` behavior.

### Environment Information

Repository inspection only; no external app runtime involved.

- Repo: `mastra-ai/mastra`
- Branch inspected: current workspace based on `origin/main`
- Packages involved: `packages/core`, `packages/server`
- Area: Workflows, FGA authorization

### Verification

tsushanth:
Fixes #18165.

## What

`Workflow.execute()` already gates execution behind a `workflows:execute` FGA check, but the resume paths — `Run.resume`, `Run.resumeAsync`, `Run.resumeStream` (all funnelling through `Run._resume`), and the evented `EventedRun.resume` — load the persisted snapshot and continue execution without an equivalent internal check.

The built-in server route adapter still enforces FGA on the `/workflows/:workflowId/.../resume*` HTTP path, so this isn't reachable from a stock deployment, but any caller hitting the core API directly or replacing/wrapping the route adapter bypasses the gate. The reporter's filing has the full path-by-path trace.

## How I changed it

Both resume entry points now mirror `Workflow.execute`:

- `packages/core/src/workflows/workflow.ts` — `Run._resume` runs the FGA check before loading the snapshot, using the same `{ type: 'workflow', id: workflowId }` resource, the same `workflows:execute` permission, and the same `resourceId` / `requestContext` / `metadata` shape the execute path uses. `Run.resume`, `Run.resumeAsync`, and `Run.resumeStream` all funnel through `_resume`, so a single insertion covers all three. The `actor` param from `_resume` is also threaded through so trusted-system-actor flows behave consistently across execute / resume.
- `packages/core/src/workflows/evented/workflow.ts` — `EventedRun.resume` gets the matching block; same shape, no `actor` since the evented public resume API doesn't currently expose one.

When no FGA provider is configured, behavior is unchanged (`requireFGA` no-ops). When FGA is configured but no authenticated user is available, resume now fails closed with the same `authenticated user is required` error initial execution already raises — same threat model, same response.

## Tests

Three regression cases added to the existing `FGA checks` describe block in `packages/core/src/workflows/workflow.test.ts`, tagged `@see #18165`:

1. `checks workflows:execute on resume with request context metadata` — asserts `fgaProvider.require` is called exactly once on resume with the expected `{ resource, permission, context }` shape.
2. `fails closed on resume when FGA is configured and no user is available` — asserts the resume rejects with `authenticated user is required` and never reaches the provider.
3. `no-ops on resume when no FGA provider is configured` — asserts existing happy path is untouched.

All 6 FGA tests pass under `pnpm vitest run src/workflows/workflow.test.ts -t FGA`. `pnpm typecheck` clean.

## Changeset

`.changeset/dark-rice-tickle.md` (`@mastra/core`: patch) describes the fix in terms of the user-visible change (resume now respects the same FGA gate as execute).

## Notes

- Kept the fix scoped per CLAUDE.md ("Bug fixes ... MUST include tests" + "Do not change public API behavior beyond the fix"). The resume signatures are unchanged; the new check is purely additive and only takes effect when an FGA provider is configured.
- The reporter also notes the standard HTTP server adapter validates stored `resourceId` ownership — that path is preserved; this fix is the defense-in-depth at the core API boundary.

## ELI5

This PR adds a “bouncer check” to workflow resume calls. Previously, you could sometimes resume a suspended workflow without proving you’re allowed to—now resume also enforces the same `workflows:execute` permission that the normal workflow start path uses.

## Summary

This PR fixes inconsistent Fine-Grained Authorization (FGA) enforcement at the core API boundary for workflow resume operations. `Workflow.execute()` already gates execution with `workflows:execute`, but the core resume paths (`Run._resume`, public `Run.resume`/`resumeAsync`/`resumeStream`, and the evented `EventedRun.resume`) could load and continue from a persisted “suspended” snapshot without performing the equivalent FGA check—allowing direct/custom callers (bypassing the standard server route adapter) to bypass authorization.

## Changes

**FGA enforcement added to resume before snapshot load**
- `packages/core/src/workflows/workflow.ts`
  - `Run._resume()` now checks `workflows:execute` with the same workflow resource shape and authorization context used by `Workflow.execute()`.
  - The actor passed into resume is forwarded into the FGA check.
  - If an FGA provider is configured but no authenticated user is present in `params.requestContext`, resume fails closed with `authenticated user is required`.
  - If no FGA provider is configured, resume proceeds unchanged.

- `packages/core/src/workflows/evented/workflow.ts`
  - `EventedRun.resume()` now performs the same `workflows:execute` FGA gate before loading the snapshot.
  - The FGA gate includes the same workflow resource, request context, context (`resourceId`), and metadata (`workflowId`, `runId`, `resourceId`).
  - When FGA isn’t configured, the gate is skipped.

**Regression tests added**
- `packages/core/src/workflows/workflow.test.ts`
  - Added tests covering:
    - authorized resume calls (verifies `fgaProvider.require` invocation and expected permission/metadata)
    - unauthenticated resume when FGA is enabled (expects `authenticated user is required`, and ensures `fgaProvider.require` is not called)
    - resume when no FGA provider is configured (proceeds successfully)

**Changeset**
- `.changeset/dark-rice-tickle.md`
  - Documents the closed authorization bypass and the fail-closed behavior when FGA is enabled without an authenticated user.

### DIFF
--- a/.changeset/dark-rice-tickle.md
+++ b/.changeset/dark-rice-tickle.md
@@ -1,0 +1,9 @@
+---
+'@mastra/core': patch
+---
+
+**Fixed:** Closed an authorization bypass where Workflow run resume APIs (`Run.resume`, `Run.resumeAsync`, `Run.resumeStream`, and the evented `EventedRun.resume`) executed persisted snapshots without an internal FGA check. The same `workflows:execute` permission that gates `Workflow.execute` now also gates every resume path, so callers reaching the core API directly (or any handler that skips the built-in server route adapter) cannot continue a suspended run that the configured FGA provider would deny.
+
+When no FGA provider is configured, behavior is unchanged. When FGA is configured and no authenticated user is available, resume now fails closed with the same `authenticated user is required` error the initial execution already raised.
+
+Bug report: https://github.com/mastra-ai/mastra/issues/18165
--- a/packages/core/src/workflows/evented/workflow.ts
+++ b/packages/core/src/workflows/evented/workflow.ts
@@ -2133,6 +2133,33 @@ export class EventedRun<
       includeResumeLabels?: boolean;
     };
   }): Promise<WorkflowResult<TState, TInput, TOutput, TSteps>> {
+    // FGA authorization check — mirrors Workflow.execute() so the evented
+    // resume path goes through the same `workflows:execute` gate as the
+    // initial execution. Without this, callers reaching EventedRun.resume
+    // directly bypass the internal check while the server route adapter
+    // still enforces it on the HTTP path.
+    const fgaProvider = this.mastra?.getServer()?.fga;
+    if (fgaProvider) {
+      const user = params.requestContext?.get('user' as any);
+      const { getWorkflowFGAResourceId, requireFGA } = await import('../../auth/ee/fga-check');
+      const { MastraFGAPermissions } = await import('../../auth/ee');
+      await requireFGA({
+        fgaProvider,
+        user,
+        resource: { type: 'workflow', id: getWorkflowFGAResourceId(this.workflowId) },
+        permission: MastraFGAPermissions.WORKFLOWS_EXECUTE,
+        requestContext: params.requestContext,
+        context: {
+          resourceId: this.resourceId,
+        },
+        metadata: {
+          workflowId: this.workflowId,
+          runId: this.runId,
+          resourceId: this.resourceId,
+        },
+      });
+    }
+
     const workflowsStore = await this.mastra?.getStorage()?.getStore('workflows');
     if (!workflowsStore) {
       throw new Error('Cannot resume workflow: workflows store is required');
--- a/packages/core/src/workflows/workflow.test.ts
+++ b/packages/core/src/workflows/workflow.test.ts
@@ -819,6 +819,101 @@ describe('Workflow (Default Engine Specifics)', () => {
       expect(result).toEqual({ value: 'ok' });
       expect(fgaProvider.require).not.toHaveBeenCalled();
     });
+
+    /**
+     * @see https://github.com/mastra-ai/mastra/issues/18165
+     */
+    describe('Run.resume FGA enforcement', () => {
+      function createSuspendableFGAWorkflow() {
+        const suspendStep = createStep({
+          id: 'fga-suspend-step',
+          inputSchema: z.object({ value: z.string() }),
+          outputSchema: z.object({ value: z.string() }),
+          suspendSchema: z.object({ message: z.string() }),
+          resumeSchema: z.object({ confirm: z.boolean() }),
+          execute: async ({ inputData, resumeData, suspend }) => {
+            if (!resumeData?.confirm) {
+              await suspend({ message: 'awaiting confirm' });
+            }
+            return inputData;
+          },
+        });
+        return createWorkflow({
+          id: 'fga-resume-workflow',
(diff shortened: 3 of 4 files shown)


abhiaiyer91:
Dont use dynamic imports

tsushanth:
Addressed — replaced both dynamic `await import('../auth/ee/fga-check')` call sites with a single static import at the top of the file. Both symbols were already re-exported from `../auth/ee` (index.ts:34,37).

abhiaiyer91:

We're currently testing out our agentic factory, and it implemented a fix for the linked issue (#18165), which has since been closed. So I'm closing this PR out as no longer needed.

PR #18276, opened 2026-06-21
feat(core): expand AIMock regression scenario suite to 166 tests across 78 files

abhiaiyer91:
## Summary

Expands the AIMock-based BDD regression test suite for the core agentic loop (`packages/core/src/loop`) from 72 files / 152 tests to **78 files / 166 tests**, covering agent features documented in the official docs including suspend/resume flows, approval paths, thread persistence, and snapshot integrity.

### Harness improvements
- Added `createSharedAgent()` helper and `sharedAgent` option to `RunLoopScenarioOptions` so scenarios can persist the same agent + Mastra storage across multiple `runLoopScenario` calls. This unlocks suspend/resume scenarios that were previously marked infeasible.
- Forwarded `memoryOptions`, `abortSignal`, and `modelSettings` to `agent.stream()` via the harness.

### New scenario files (16 files, ~14 new tests)
| File | What it covers |
| --- | --- |
| `auto-resume-suspended-tools` | `autoResumeSuspendedTools: true` injects suspended-tool info into the system prompt on the next call |
| `resume-stream` | Manual resumption via `agent.resumeStream(resumeData, { runId })` |
| `generate-approval-path` | Non-streaming `approveToolCallGenerate()` / `declineToolCallGenerate()` with `finishReason: 'suspended'` |
| `resume-after-decline` | Declined tool can be retried in a subsequent turn with fresh approval |
| `multi-call-thread-persistence` | Message accumulation across calls; thread and resource isolation |
| `suspended-snapshot-integrity` | Complex nested args preserved exactly through suspend → resume cycle |
| `delegation-complete-bail` | `onDelegationComplete` bail() stops supervisor loop immediately |
| `include-subagent-tool-results` | `includeSubAgentToolResultsInModelContext` flag controls subagent result visibility |
| `tool-runtime-suspension` | `context.suspend()` with `resumeSchema` / `resumeData` |
| `on-error-callback` | `onError` fires for API errors but not tool errors |
| `maxsteps-long-chains` | `maxSteps` + `stopWhen` interaction in long tool chains |
| `structured-output-error-strategy` | `strict` / `warn` / `fallback` error strategies on schema validation |
| `request-context-isolation` | `requestContext` passthrough across execution steps |
| `request-context-mutation` | Tool mutations do NOT persist between tool executions |
| `empty-turn` | Model returns text immediately without tool calls |
| `abort-structured-output` | Abort signal during structured output streaming |

### Regression-injection proofs
Performed on:
- `autoResumeSuspendedTools` system prompt injection path (disabled check → scenario catches missing prompt modification)
- `tool-call-suspended` chunk type (renamed to `tool-call-paused` → both resume tests fail)
- `!resumeData.approved` decline path (always-true → approved tool also gets "not approved" result)
- Suspended snapshot args preservation

### Verification
- All 166 tests pass across 78 scenario files
- Typecheck clean
- Zero changes to core loop source code (harness + scenarios only)

## Test plan
- `pnpm vitest run packages/core/src/loop/test-utils/aimock/scenarios` — 78 files / 166 tests
- `pnpm --filter ./packages/core check` — typecheck clean
- Regression injections performed on 4 paths (see above)

## ELI5
This PR adds lots more “can’t break” tests for Mastra’s core AI agent loop—especially the tricky parts where the agent pauses for tool approval and later resumes, while keeping thread/memory data consistent. It only changes test code and the test harness, so it helps catch bugs early without changing how the real agent works.

---

## Test Suite Expansion
- Expands the AIMock-based BDD regression suite for `packages/core/src/loop` from **72 files / 152 tests** to **78 files / 166 tests**.
- Adds scenarios covering suspend/resume flows, approval paths (approve/decline + retries), thread persistence/isolation, and suspended snapshot integrity.

## New Testing Infrastructure / Harness Updates
- Adds a `createSharedAgent()` helper and a `sharedAgent` option in `RunLoopScenarioOptions`, letting multiple `runLoopScenario()` calls reuse the same `Agent` + Mastra storage (required for realistic suspend/resume across turns).
- Adds `aimock-scenario.ts` utilities:
  - `useLoopScenarioAimock()` to run a shared AIMock server for the test suite
  - `runLoopScenario()` to drive loop runs and optionally collect chunks
  - `runApprovalScenario()` to drive tool-approval suspend/resume flows and collect approval logs

## Key Option / Plumbing Improvements
- Forwards scenario inputs into `agent.stream()`:
  - `memoryOptions`
  - `abortSignal`
  - `modelSettings`

## Scenario Coverage Highlights (New/Expanded)
- Approval + resume behaviors:
  - `autoResumeSuspendedTools` prompt/system injection on follow-up calls
  - manual resumption via `agent.resumeStream()` with `resumeData`
  - `approveToolCallGenerate()` / `declineToolCallGenerate()` paths, including decline-then-retry
  - `tool-call-suspended` chunk validation and suspended snapshot integrity (including nested argument preservation)
- Thread/message correctness:
  - message accumulation and thread isolation across multiple runs
  - nested delegation and shared-storage suspend/resume flows
- Structured output + reliability:
  - strict/fallback/warn validation error strategies and repair behavior
  - abort-signal handling during structured output streaming
  - empty/no-tool turn behavior with text-only model responses
- Request context + mutation semantics:
  - `requestContext` passthrough into tool execution
  - requestContext isolation (including “tool mutations don’t leak” across tool executions)
- Signals / PubSub edge cases:
  - subscribe/unsubscribe delivery correctness, dedup/cacheKey behavior, and wake vs persist semantics
- Additional loop behaviors exercised via new scenarios:
  - delegation completion with `bail()`
  - subagent tool result visibility control
  - long tool chains bounded by `maxSteps` and `stopWhen` interaction
  - lifecycle/error callback execution patterns
  - multi-tool parallel execution and streamed chunk ordering

## Regression-Proofing + Documentation / Release Notes
- Improves test falsifiability via hardened assertions (tight request counts/orderings, concrete `finishReason` checks, stronger validation of approval/abort behavior).
- Adds changeset/changelog notes, contributor guidance, and a comprehensive AIMock scenario README.
- Updates `packages/core/package.json` devDependencies to include `@copilotkit/aimock` for the enhanced AIMock test harness.


abhiaiyer91:
## CodeRabbit review — addressed in 6d9c402

Thanks for the thorough pass. All 14 findings handled; the theme was "tests that pass regardless of loop behavior." Fixes below; full aimock suite is green (83 files / 179 tests), typecheck clean, no loop source changes.

**Made falsifiable / strengthened**
- **actor-identity** — now captures `context.actor` inside the tool and asserts it equals the passed actor (and is `undefined` when none is provided).
- **error-processor-retry-exhaustion** & **error-processor** — replaced the `expect(true).toBe(false)` sentinel. Note: `runLoopScenario` does **not reject** on terminal API errors — it resolves with `finishReason: 'error'`. So `.rejects.toThrow()` would fail; instead we assert `finishReason === 'error'` plus the processor side effects (retryCount sequence, call order, state). This is what actually masked the original behavior.
- **maxsteps-long-chains** — exact counts at the `maxSteps` cap and the `stopWhen` boundary instead of `<=` ceilings. Dropped the `stopWhen + maxSteps` combination because it does **not** bound at the intended step (effective bound was `maxSteps`); flagging that separately rather than encoding it as expected.
- **provider-metadata** — the OpenAI v5 provider does not serialize `providerOptions` into the captured chat-completions body, so a body-level assertion isn't possible. Re-scoped to pair `providerOptions` with an observable `modelSettings.temperature` and assert that lands in the body; documented the limitation in the file header.
- **structured-output-repair** — re-scoped to validation-failure surfacing (assert the `error` chunk's validation message) and valid streamed-JSON assembly. AIMock returns each body as one block, so a genuinely torn/partial JSON stream can't be scripted here (covered by output-format-handlers unit tests); documented.
- **processor-retry** — re-scoped the describe/docstring to the observation/transformation/sequencing it actually verifies (not retry); now asserts exact processor order and request counts.

**Signals**
- Best-effort subscriber delivery in the in-test pubsub (a throwing subscriber no longer aborts delivery or bubbles as an uncaught async error), in both signal files.
- `withTimeout` now clears the timer whether the wrapped promise or the timeout wins (previously leaked when the promise won).
- Awaited the async `unsubscribe()` calls; fixed the misleading state-signal dedup comment (the cacheKey changed, which is why it's accepted).

**skills-same-name-disambiguation** — normalize path separators before path assertions so they hold on Windows.

**Abort timing (already addressed before this review)** — `abort-during-tool-execution` and `abort-structured-output` were rewritten to abort deterministically from inside tool `execute` (no `setTimeout` races); verified they go red when abort wiring is disabled.

wardpeet (approved):

PR #18286, opened 2026-06-22
feat(voice-google-gemini-live): add sendContext() for context seeding

### LINKED ISSUE #18285: [FEATURE] @mastra/voice-google-gemini-live: public API to send incremental client_content turns
  (turnComplete: false) for context seeding
## Problem Statement

`@mastra/voice-google-gemini-live` has no public way to send a multi-turn `client_content` frame with `turnComplete: false` into a freshly connected Gemini Live session.

This is the wire shape Gemini Live exposes for *incremental content* — loading conversation history into a session without triggering a model response — and it's the documented way to seed prior context on a cold connect (see [Gemini Live API docs § sendClientContent](https://ai.google.dev/gemini-api/docs/live-api/capabilities)).

The two public send paths on `GeminiLiveVoice` today only cover the response-triggering case:

- `speak(text)` ([`voice/google-gemini-live-api/src/index.ts`](https://github.com/mastra-ai/mastra/blob/main/voice/google-gemini-live-api/src/index.ts)) always sends a single `user` turn with `turnComplete: true`. The model immediately generates a reply.
- `send(audioStream | Int16Array)` is realtime audio, not text turns.
- `resumeSession(handle, context)` accepts `Array<{ role, content }>` but only runs on the resumption path (requires a prior handle, opens a `session_resume` frame).
- The lower-level `sendEvent('client_content', …)` that would carry an arbitrary multi-turn / non-completing frame is `private` on the provider class.

Apps that store conversation history outside the live session (e.g. Mastra `Memory` with text + voice sharing one `threadId`) currently have to reach into the private `sendEvent` to replay history on cold connect. That works but breaks every minor version bump.

## Proposed Solution

Add a public method on `GeminiLiveVoice` for incremental context, roughly:

```ts
type IncrementalTurn = { role: 'user' | 'assistant'; content: string };

await voice.sendContext(turns: IncrementalTurn[], options?: { turnComplete?: boolean }): Promise<void>;
```

Semantics:

- Maps to a single `client_content` frame with role-tagged `turns` (`user` → `user`, `assistant` → `model`) and `turnComplete` defaulting to `false`.
- When `turnComplete: false`, the server loads the turns into context **without** generating a response. The next real client turn that completes is what the model replies to. This matches Gemini Live's documented incremental-content behaviour.
- Must be safe to call after `connect()` resolves and before the first `send(audio)` / `speak()` call.
- Should also append to the provider's internal context history (same path `addToContext` already uses) so `getContextHistory()` stays consistent.

A thinner alternative is to expose `sendEvent` (or a narrow `sendClientContent({ turns, turnComplete })`) as public — same effect, smaller surface, but less ergonomic for the common "seed N turns" case.

## Component

Voice

## Alternatives Considered

1. **Reach into `private sendEvent` from app code.** Works today but is not a supported surface; any internal refactor of the provider's send pipeline breaks consumers silently.
2. **Fold history into `instructions` / system prompt before `connect()`.** Rejected by callers because it pushes untrusted user text into the instruction channel (prompt-injection escalation) and conflates roles. Replaying as role-tagged `user` / `model` turns keeps untrusted content in the same channel a live call already uses.
3. **Extend `speak()` to accept `turnComplete: false` and multi-turn input.** Possible, but overloads a method whose name implies the model will speak. A separate `sendContext()` keeps intent clear.
4. **Wait for `resumeSession()` to cover this.** Resumption only restores a *prior* Gemini Live session from a server-side handle, within Gemini's resumption window. It cannot carry history from other surfaces (e.g. text chat sharing the same thread) into voice, and does nothing on a true cold start. Resumption and seeding are complementary, not substitutes.

## Example Use Case

As an app developer building a voice agent that shares a thread with a text chat surface, I want to prime each fresh Gemini Live session with the recent conversation history so a returning user doesn't have to re-explain context — without the agent greeting them on connect, and without folding user-supplied text into the system instructions.

```ts
import { GeminiLiveVoice } from '@mastra/voice-google-gemini-live';

const voice = new GeminiLiveVoice({ /* ... */ });
await voice.connect();

const history = await memory.recall({ threadId, resourceId, selectBy: { last: 20 } });

// Replay prior turns into the live session without triggering a reply.
await voice.sendContext(
  history.messages
    .filter(m => m.role === 'user' || m.role === 'assistant')
    .map(m => ({ role: m.role, content: m.content })),
);

// Agent stays silent until the user actually speaks.
await voice.send(micStream);
```

## Additional Context

- Gemini Live wire reference: <https://ai.google.dev/gemini-api/docs/live-api/capabilities> (sendClientContent with `turnComplete: false`).
- Related issue (not a duplicate): #18119 covers fixing `resumeSession()` itself. Resumption restores a prior Gemini Live session from a handle; this request covers the cold-connect / cross-surface case that resumption can't address.
- Provider entry point that would host the new method: [`voice/google-gemini-live-api/src/index.ts`](https://github.com/mastra-ai/mastra/blob/main/voice/google-gemini-live-api/src/index.ts) — the `private sendEvent('client_content', …)` and existing `speak()` / `resumeSession()` paths show the shape this would slot into.

## Verification

intojhanurag:
## Description

Adds a public `sendContext()` method to `GeminiLiveVoice` that sends conversation history into a Gemini Live session without triggering a model response. This replaces the need to reach into the private `sendEvent` method, which broke on every minor version bump.

- Maps turns to a single `client_content` frame with `turnComplete` defaulting to `false`
- Converts `assistant` role to `model` for Gemini's wire format
- Syncs seeded turns to the provider's internal context history via `addToContext()`
- Exports `IncrementalTurn` type for consumers
- Verified against real Gemini Live API with `gemini-2.5-flash-native-audio-preview-12-2025` (full user+assistant history) and `gemini-3.1-flash-live-preview` (user-only history)

## Related Issue(s)

Fixes #18285

## Type of Change

- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [ ] Test update

## Checklist

## ELI5

This PR lets your Gemini voice session start with “already-discussed chat history,” without making Gemini talk immediately. It’s like preloading the AI’s memory card so it stays quiet until the user speaks.

## Overview

Adds a public `sendContext()` method to `GeminiLiveVoice` so developers can seed previous conversation turns on a cold Gemini Live connect, replacing a previous workaround that relied on the private `sendEvent()` API.

## Key Changes

**New Public Method: `sendContext()`**
- Signature: `sendContext(turns: IncrementalTurn[], options?: { turnComplete?: boolean }): Promise<void>`
- Input: an array of `{ role: 'user' | 'assistant'; content: string }` (via the new `IncrementalTurn` type)
- Sends **one** Gemini `client_content` frame containing all provided turns
- Defaults `turnComplete` to `false`, so seeding does **not** trigger an immediate model response
- Applies Gemini wire-format requirements by mapping `assistant` → `model` within `client_content.turns`
- Keeps provider state consistent by also appending the seeded turns into internal context via `addToContext()` (so `getContextHistory()` reflects what was sent)
- Connection/edge-case behavior:
  - Rejects with `Not connected` when not connected
  - No-ops when `turns` is empty (no WebSocket `send`)
  - Throws a standardized `GeminiLiveError` using `AUDIO_PROCESSING_ERROR` if sending or context update fails

**New Public Type Export**
- Exports `IncrementalTurn` for consumers.

**Documentation & Release Notes**
- Adds reference docs for `sendContext()` (including a `connect() → sendContext([...]) → send(micStream)` style example)
- Adds a changeset entry documenting the new capability.

## Testing

Added tests covering:
- Default behavior sends `turnComplete: false`
- Correct formatting of the outgoing `client_content.turns` (including assistant→model normalization)
- `turnComplete: true` override behavior
- Internal context history updates after seeding
- `sendContext([])` causes no WebSocket send
- Rejection when the voice session is disconnected

## Model Notes (Role Support Variations)

Gemini Live models vary in what they accept for seeded `client_content`: some support both user and assistant roles, while others only accept user-role turns. `sendContext()` follows Gemini’s required wire format, so behavior can differ based on the selected Gemini Live endpoint.

### DIFF
--- a/.changeset/itchy-frogs-type.md
+++ b/.changeset/itchy-frogs-type.md
@@ -2,7 +2,7 @@
 '@mastra/voice-google-gemini-live': minor
 ---
 
-Added `sendContext()` method to `GeminiLiveVoice` for seeding conversation history into a live session without triggering a model response. This maps to Gemini Live's `client_content` frame with `turnComplete: false`, letting apps replay prior turns (e.g. from Mastra Memory) on a cold connect so the model has context before the user speaks.
+Added `sendContext()` method to `GeminiLiveVoice` for seeding conversation history into a fresh voice session without triggering a model response. This lets apps replay prior turns from Mastra Memory (or any external store) on a cold connect so the model has full context before the user speaks — enabling seamless handoff between text chat and voice on a shared thread.
 
 **Usage:**
 
--- a/.changeset/itchy-frogs-type.md
+++ b/.changeset/itchy-frogs-type.md
@@ -1,0 +1,19 @@
+---
+'@mastra/voice-google-gemini-live': minor
+---
+
+Added `sendContext()` method to `GeminiLiveVoice` for seeding conversation history into a live session without triggering a model response. This maps to Gemini Live's `client_content` frame with `turnComplete: false`, letting apps replay prior turns (e.g. from Mastra Memory) on a cold connect so the model has context before the user speaks.
+
+**Usage:**
+
+```ts
+await voice.connect();
+
+await voice.sendContext([
+  { role: 'user', content: 'What is the weather?' },
+  { role: 'assistant', content: 'It is 72°F in San Francisco.' },
+]);
+
+// Model stays silent until the user actually speaks
+await voice.send(micStream);
+```
--- a/docs/src/content/en/reference/voice/google-gemini-live.mdx
+++ b/docs/src/content/en/reference/voice/google-gemini-live.mdx
@@ -280,6 +280,53 @@ Converts text to speech and sends it to the model. Can accept either a string or
 
 Returns: `Promise<void>` (responses are emitted via `speaker` and `writing` events)
 
+### `sendContext()`
+
+Sends conversation history into the live session without triggering a model response. Use this to seed prior turns (e.g. from Mastra Memory) on a cold connect so the model has context before the user speaks.
+
+```typescript
+await voice.sendContext([
+  { role: 'user', content: 'What is the weather?' },
+  { role: 'assistant', content: 'It is 72°F in San Francisco.' },
+])
+
+// Model stays silent until the user actually speaks.
+await voice.send(micStream)
+```
+
+<PropertiesTable
+  content={[
+  {
+    name: 'turns',
+    type: 'IncrementalTurn[]',
+    description: 'Prior conversation turns to seed into the session. Each turn has a `role` ("user" or "assistant") and `content` string.',
+    isOptional: false,
+  },
+  {
+    name: 'options',
+    type: 'object',
+    description: 'Optional configuration.',
+    isOptional: true,
+    properties: [
+      {
+        type: 'object',
+        parameters: [
+          {
+            name: 'turnComplete',
+            type: 'boolean',
+            description: 'Whether to mark the turn as complete and trigger a model response.',
+            isOptional: true,
+            defaultValue: 'false',
+          },
+        ],
+      },
+    ],
+  },
+]}
+/>
+
+Returns: `Promise<void>`
+
 ### `listen()`
 
 Processes audio input for speech recognition. Takes a readable stream of audio data and returns the transcribed text.
--- a/docs/src/content/en/reference/voice/google-gemini-live.mdx
+++ b/docs/src/content/en/reference/voice/google-gemini-live.mdx
@@ -299,7 +299,8 @@ await voice.send(micStream)
   {
     name: 'turns',
     type: 'IncrementalTurn[]',
-    description: 'Prior conversation turns to seed into the session. Each turn has a `role` ("user" or "assistant") and `content` string. Both roles are supported on newer models (e.g. `gemini-2.5-flash-native-audio-preview-12-2025`). Some older models only accept user-role turns.',
+    description:
+      'Prior conversation turns to seed into the session. Each turn has a `role` ("user" or "assistant") and `content` string. Both roles are supported on newer models (e.g. `gemini-2.5-flash-native-audio-preview-12-2025`). Some older models only accept user-role turns.',
     isOptional: false,
   },
   {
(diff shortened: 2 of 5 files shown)


intojhanurag:
Hey @petrchiriboga ,  Full user+assistant history seeding works on newer models (e.g. gemini-2.5-flash-native-audio-preview-12-2025). Older models like gemini-3.1-flash-live-preview only accept user-role turns in client_content, this is a Google API model-level constraint.

wardpeet (approved):

petrchiriboga:
> Hey @petrchiriboga ,  Full user+assistant history seeding works on newer models (e.g. gemini-2.5-flash-native-audio-preview-12-2025). Older models like gemini-3.1-flash-live-preview only accept user-role turns in client_content, this is a Google API model-level constraint.

@intojhanurag Hmm interesting, thanks for the heads up.
 
Though, isn't 3.1 the latest gemini live model?

intojhanurag:
> > Hey @petrchiriboga ,  Full user+assistant history seeding works on newer models (e.g. gemini-2.5-flash-native-audio-preview-12-2025). Older models like gemini-3.1-flash-live-preview only accept user-role turns in client_content, this is a Google API model-level constraint.
> 
> Hmm interesting, thanks for the heads up.
> 
> Though, isn't 3.1 the latest gemini live model?

I just checked it was not working , not sure about it is new or not . But there was another solution was related to history_config but it was breaking the whole architecture so I avoided that and implemented this with some limitation with models :)

petrchiriboga:
> > > Hey @petrchiriboga ,  Full user+assistant history seeding works on newer models (e.g. gemini-2.5-flash-native-audio-preview-12-2025). Older models like gemini-3.1-flash-live-preview only accept user-role turns in client_content, this is a Google API model-level constraint.
> > 
> > Hmm interesting, thanks for the heads up.
> > 
> > Though, isn't 3.1 the latest gemini live model?
> 
> I just checked it was not working , not sure about it is new or not . But there was another solution was related to history_config but it was breaking the whole architecture so I avoided that and implemented this with some limitation with models :)

Thank you so much! It should be alright!

PR #18302, opened 2026-06-22
fix(libsql,pg,mysql): drop dynamic import in MemoryStorage.init() so build output doesn't deadlock

### LINKED ISSUE #18298: mastra build output deadlocks: self-referential import('./index.mjs') in LibSQL/PG memo
This issue was created from Discord post 1518556818220126268:

[Open in Discord](https://discord.com/channels/1309558646228779139/1518556818220126268)

**Summary**
A built Mastra app hangs forever during storage initialization. The bundler splits the app into an index.mjs entry chunk and a mastra.mjs chunk; the LibSQL/Postgres memory stores live in mastra.mjs but load OBSERVATIONAL_MEMORY_TABLE_SCHEMA via await import('./index.mjs'). Since index.mjs statically imports mastra.mjs, that dynamic import can never resolve when storage initializes during module evaluation. dev works because it's unbundled.

**Environment**
```
 - @mastra/core 1.42.0, @mastra/libsql 1.13.0, @mastra/pg (same catalog), @mastra/memory 1.20.3
 - Node 22.13.0, ESM output ("type": "module"), mastra build → mastra start
 - Storage: LibSQLStore (local file:) and PostgresStore — both affected
```

**Symptoms**
 - mastra dev (mastra dev --dir src): works perfectly -> chatting with harness
 - mastra build + mastra start: server boots and logs Mastra API running, but the first request that triggers harness/storage init hangs indefinitely. No error, no timeout — just silence.
 - Process is idle (event loop free, awaiting an unresolved promise) with one referenced timer; not CPU-bound, no network activity.

**Reproduction**
1. Mastra app using a Harness with Memory (observational memory enabled) and LibSQLStore (local file:) or PostgresStore, where new Mastra({ storage }) is constructed at module top level.
2. mastra build then mastra start.
3. Send a request that builds the harness / initializes storage.
4. Observe the hang. Add logging around MemoryLibSQL.init() in the built mastra.mjs — it reaches the await import('./index.mjs') line and never returns.

**Workaround I will use for now (confirmed working)**
Mark the storage/memory packages as bundler externals so they load from node_modules (like dev):

```
new Mastra({
  bundler: { externals: ['@mastra/libsql', '@mastra/pg', '@mastra/memory'] },
});
```

tsushanth:
Fixes #18298.

## What

`LibSQLStore`, `PostgresStore`, and `MySQLStore` each loaded the observational-memory table schema inside `MemoryStorage.init()` via

```ts
const { OBSERVATIONAL_MEMORY_TABLE_SCHEMA } = await import('@mastra/core/storage');
```

@daneatmastra's analysis is the right read: `mastra build` splits the user's app into an entry `index.mjs` and a `mastra.mjs` chunk, the memory stores end up in `mastra.mjs`, and the dynamic import gets rewritten to target the entry chunk (because `@mastra/core` is hoisted there). When `new Mastra({ storage })` runs at module top level the storage init dynamically re-imports the entry chunk that's currently still being evaluated — top-level import cycle, never resolves. The server boots, logs `Mastra API running`, then hangs on the first storage-touching request.

`mastra dev` doesn't hit it because it doesn't bundle.

## Why the fallback is safe to drop

The dynamic import was guarded with `try { … } catch {}` for compatibility with older `@mastra/core` versions that didn't export `OBSERVATIONAL_MEMORY_TABLE_SCHEMA`. But all three stores declare:

```json
"peerDependencies": { "@mastra/core": ">=1.42.1-0 <2.0.0-0" }
```

and `OBSERVATIONAL_MEMORY_TABLE_SCHEMA` has been exported from `@mastra/core/storage` since `1.42.1` — the lower bound of every store's peerDep. So no resolvable npm graph can produce a `@mastra/core` without the constant. The fallback is dead code.

## How

- `stores/libsql/src/storage/domains/memory/index.ts` — add `OBSERVATIONAL_MEMORY_TABLE_SCHEMA` to the existing top-of-file static import from `@mastra/core/storage`. Read the table entry directly in `init()`.
- `stores/mysql/src/storage/domains/memory/index.ts` — same.
- `stores/pg/src/storage/domains/memory/index.ts` — reuse the module-level `_omTableSchema` that's already set via the top-of-file `createRequire` shim (lines 60-68). The dynamic import in `init()` was a duplicate of work already done at module load.

## Test plan

- `pnpm --filter ./stores/libsql --filter ./stores/pg --filter ./stores/mysql run build:lib` — all three build clean against the local `@mastra/core` build.
- `pnpm --filter ./stores/libsql run test --run -t init` — 14 passed / 0 failed (the suite under `init`).
- `pnpm prettier:changed` clean. `pnpm --filter ./stores/libsql --filter ./stores/pg --filter ./stores/mysql run lint` clean.

I did not stand up a full `mastra build` + `mastra start` cycle locally — flagging that explicitly since the failure is build-bundling specific. The static-import side of the change is mechanical and the failing path it removes is exactly the dynamic import @daneatmastra called out. Happy to add an integration test that asserts on the bundled output if you want — let me know which test suite that should live in.

## Notes

- The reporter's `bundler.externals: ['@mastra/libsql', '@mastra/pg', '@mastra/memory']` workaround is no longer required after this lands.
- No behavior change for `@mastra/core` versions ≥ `1.42.1` (the entire peerDep range). The only behavior change vs the old code path: a forced install of a `@mastra/core` < `1.42.1` (not resolvable via npm peer-dep checks) would now fail at top-level import instead of silently no-op-ing the OM table init. That's strictly better diagnostics if anyone forces it.
- Touches three packages so the changeset bumps `@mastra/libsql`, `@mastra/pg`, and `@mastra/mysql` together. Did not include `@mastra/core` in the changeset since the fix is store-side.

## ELI5 (Explain Like I’m 5)

When you bundle a Mastra app, it splits code into multiple files. Those files ended up accidentally “asking each other for help” in a circle, so the app froze the first time it tried to use observational memory storage. This PR fixes it by loading the needed observational-memory table schema up front, instead of importing it later at startup.

## Problem

Built Mastra applications with `LibSQLStore`, `PostgresStore`, or `MySQLStore` using observational memory could hang indefinitely on the first request that accessed storage. The stores dynamically imported `OBSERVATIONAL_MEMORY_TABLE_SCHEMA` from `@mastra/core/storage` inside `MemoryStorage.init()`. During bundling, that dynamic import was rewritten in a way that created a circular import: the bundled entry chunk statically imports the store chunk, while the store chunk tries to dynamically re-import the entry chunk during module initialization—so the import never finishes. The app could start and log “Mastra API running” but then freeze when storage was first used. This didn’t affect `mastra dev` because dev mode doesn’t bundle.

## Solution

Replaced the runtime/dynamic import with module-level static data loading:

- **`libsql` and `mysql` stores**: Added `OBSERVATIONAL_MEMORY_TABLE_SCHEMA` to the existing top-level static import from `@mastra/core/storage` and read the schema directly during `init()`, removing the dynamic import + `try/catch` fallback.
- **`pg` store**: Reused an existing module-scoped `_omTableSchema` value set via a `createRequire` shim, avoiding duplicate schema lookup work.

This is safe because these stores rely on a `@mastra/core` peer dependency version where `OBSERVATIONAL_MEMORY_TABLE_SCHEMA` is available (>= `1.42.1`). With that guarantee, the old compatibility guard around the dynamic import is no longer needed.

## Impact

- Removes the circular-import deadlock that could hang built apps on first storage access (observational memory).
- No public API changes.
- Makes the reporter’s bundler externals workaround unnecessary.


tsushanth:
Thanks @coderabbitai — tightened the changeset in [379b0e7](https://github.com/mastra-ai/mastra/pull/18302/commits/379b0e7bab) to user-focused outcome only (build no longer hangs, no app-side workaround needed). Implementation rationale stays in the commit body / PR description.

abhiaiyer91 (approved):

PR #18334, opened 2026-06-23
fix(core): inject mastra into tool execution context for voice providers

### LINKED ISSUE #18283: [BUG] Agent does not inject `mastra` into tool execution context for voice providers
### Describe the Bug

When an `Agent` is configured with a static voice provider (e.g. `voice: new GeminiLiveVoice({ ... })` or any `MastraVoice` subclass), `Agent` hands its tools to the voice provider via `voice.addTools(tools)` **without** wrapping those tools to inject the `mastra` instance into their execution context.

As a result, any tool that reads `context.mastra` — e.g. `context.mastra.getVector(...)`, `context.mastra.getAgent(...)`, `context.mastra.getWorkflow(...)` — silently breaks when the same agent is invoked over voice, even though the tool works correctly over chat.

This is **not** Gemini-Live-specific. It affects every realtime voice provider in the ecosystem because the gap is in `Agent`, not in the voice providers themselves.

### Steps To Reproduce

```bash
git clone https://github.com/petrchiriboga/mastra-voice-tool-mastra-context-repro
cd mastra-voice-tool-mastra-context-repro
pnpm install
pnpm repro
```

The repro is a single ~130-line Node script (`repro.ts`) that:

1. Defines a ~20-line `StubVoice` (a `MastraVoice` subclass). Its `invokeTool` does exactly what every realtime voice provider does on a function call:
   ```ts
   await tool.execute(args, { requestContext: this.requestContext });
   ```
2. Defines two probe tools whose `execute` records what `ctx.mastra` it sees.
3. Creates a `Mastra` instance with one agent that has both tools and a `StubVoice` voice provider attached.
4. Invokes the same shape of tool call along two paths:
   - **Chat path**: builds the tool-execution context the way `Agent`'s chat loop already does today — `{ requestContext, mastra }` — and calls `tool.execute(args, ctx)`.
   - **Voice path**: calls `await agent.getVoice()` to resolve the voice instance, then invokes a tool on it. This matches exactly how `Agent.getVoice()` hands tools to a static voice provider via `voice.addTools(...)` at `packages/core/src/agent/agent.ts:1978`.
5. Prints both `ctx.mastra` values side by side.

No API keys, no network, no Gemini — only `@mastra/core@1.45.0`.

The repro exits with code `1` when the bug is present, so it can be wired into CI as a regression guard once a fix lands.

https://github.com/petrchiriboga/mastra-voice-tool-mastra-context-repro

### Expected Behavior

A tool's `execute(args, context)` receives the same `context.mastra` regardless of whether the agent is invoked over chat or over voice. Tools written against the documented `ToolExecutionContext` shape should work identically in both paths.

Actual output today:

```
chat  path: ctx.mastra is <_Mastra> ✅
voice path: ctx.mastra is undefined ❌
```

#### Where the gap lives (for maintainers)

In `packages/core/src/agent/agent.ts`, two places hand tools to a static voice instance via `voice.addTools(...)`:

- Constructor seed: around `agent.ts:528`
- `getVoice()`: around `agent.ts:1978`

Both call `voice.addTools(tools)` without wrapping each tool's `execute` to inject `mastra` into the execution context. The resolver-function voice path at `agent.ts:1969–1974` does receive `mastra`, but only because the user's resolver factory can hand it to the voice instance manually — which still requires per-provider knowledge today.

By contrast, when `Agent` executes tools itself, the tool-execution context is built by the agent and includes `mastra: this.#mastra` — the same injection appears repeatedly throughout `agent.ts` (e.g. lines 1716, 1788, 1925, 2114, 2129, 2177). The voice path should follow the same convention.

#### Proposed fix (Agent-side only, no public API change)

At the two `voice.addTools(...)` call sites in `Agent`, wrap each tool's `execute` so that:

```ts
execute: (input, ctx) => originalExecute(input, { ...ctx, mastra: this.#mastra })
```

This is the same shape `Agent` already uses for tool wrapping (`wrapToolWithHooks` at `agent.ts:5637`) and matches the `mastra: this.#mastra` injection pattern used everywhere else in the agent.

- **All voice providers benefit automatically.** No per-provider changes.
- **No public API change.** `MastraVoice`, `IMastraVoice`, and provider constructors untouched.
- **No user-facing migration.** `voice: new GeminiLiveVoice({ ... })` continues to work and now correctly delivers `mastra` to tool execution.

#### Workaround currently used by downstream consumers

A `bindMastraToVoiceTools` wrapper that walks the agent's tools at app-startup time and rewrites each tool's `execute` to inject `mastra` manually. This works but every consumer of every realtime voice provider needs the same wrapper — the right place for the fix is `Agent` itself.

### Environment Information

```
  System:
    OS: macOS
    Node: v22.13+
  npmPackages:
    @mastra/core: 1.45.0
```

- LLM Provider: not applicable — repro uses a stub voice provider and does not call a real model.
- Browser: not applicable — Node.js script.

### Verification

Aayush-engineer:
## Description

When an `Agent` is configured with a static voice provider (e.g. `GeminiLiveVoice`), it calls `voice.addTools(tools)` at two places — the constructor and `getVoice()` — without wrapping each tool's `execute` to inject the `mastra` instance into the execution context. As a result, any tool that reads `context.mastra` (e.g. `context.mastra.getAgent()`, `context.mastra.getVector()`) gets `undefined` when invoked over voice, even though the exact same tool works correctly over chat.

The root cause is that the agent's chat/stream path builds the tool execution context itself and always includes `mastra: this.#mastra`, but the voice path handed raw unwrapped tools directly to the voice provider.

**Fix:** At both `voice.addTools(...)` call sites in `agent.ts`, wrap each tool's `execute` to spread the incoming context and inject `mastra: this.#mastra`, exactly matching the pattern already used throughout the agent for chat/stream tool execution. No public API change — `MastraVoice`, `IMastraVoice`, and all provider constructors are untouched. Every voice provider benefits automatically.

## Related issue(s)
Fixes #18283

## Type of change

## Checklist

## ELI5 (Explain Like I'm 5)

Imagine your agent has helper tools that need to know about their parent agent to do their job. When you talk to the agent through chat, the helpers get that information. But when you talk through voice, the helpers weren't getting that information and would fail. This PR fixes it by making sure helpers get that information regardless of whether you're using voice or chat.

## Summary

This PR fixes a bug where tools invoked through voice providers fail to receive the `mastra` instance in their execution context, while the same tools work correctly when invoked through chat. Tools that depend on `context.mastra` (such as `context.mastra.getAgent()` or `context.mastra.getVector()`) would fail silently when called over voice.

### Root Cause

The agent's chat and stream execution paths build tool execution context internally and inject `mastra: this.#mastra`, but the voice path was passing unwrapped raw tools directly to voice providers without this context enrichment.

### Solution

The fix wraps each tool's `execute` method at the two `voice.addTools(...)` call sites in `agent.ts` (in the constructor and in the `getVoice()` method) to inject `mastra: this.#mastra` into the tool execution context. This matches the pattern already used throughout the agent for chat and stream tool execution.

### Impact

- **Non-breaking**: No modifications to public APIs (`MastraVoice`, `IMastraVoice`, or provider constructors)
- **Universal**: All voice providers automatically benefit from the fix without requiring per-provider modifications
- **Consistent**: Tools now receive the same `context.mastra` instance regardless of invocation path (chat or voice)

### Files Changed

- **`.changeset/fix-agent-voice-tool-mastra-context.md`**: Changeset documenting the patch to `@mastra/core`
- **`packages/core/src/agent/agent.ts`**: Tool wrapping logic added at both `voice.addTools()` call sites to inject mastra context

### DIFF
--- a/.changeset/fix-agent-voice-tool-mastra-context.md
+++ b/.changeset/fix-agent-voice-tool-mastra-context.md
@@ -1,0 +1,5 @@
+---
+"@mastra/core": patch
+---
+
+Agent now injects the `mastra` instance into the tool execution context when handing tools to a static voice provider, matching the same injection already done for chat/stream paths. Tools that access `context.mastra` now work correctly over voice (e.g. `GeminiLiveVoice`) without any workarounds.
--- a/packages/core/src/agent/agent.ts
+++ b/packages/core/src/agent/agent.ts
@@ -525,7 +525,15 @@ export class Agent<
       // where its session-owned instance is configured, so we must not touch it here.
       if (typeof this.#voice !== 'function') {
         if (typeof config.tools !== 'function') {
-          this.#voice.addTools(this.#tools as TTools);
+          const wrappedToolsForVoice = Object.fromEntries(
+            Object.entries(this.#tools as TTools).map(([name, tool]: [string, any]) => [
+              name,
+              typeof tool?.execute === 'function'
+                ? { ...tool, execute: (input: unknown, ctx: any) => tool.execute(input, { ...ctx, mastra: this.#mastra }) }
+                : tool,
+            ]),
+          ) as TTools;
+          this.#voice.addTools(wrappedToolsForVoice);
         }
         if (typeof config.instructions === 'string') {
           this.#voice.addInstructions(config.instructions);
@@ -1967,7 +1975,16 @@ export class Agent<
     }
 
     const voice = this.#voice;
-    voice?.addTools(await this.listTools({ requestContext }));
+    const resolvedTools = await this.listTools({ requestContext });
+    const wrappedToolsForVoice = Object.fromEntries(
+      Object.entries(resolvedTools as Record<string, any>).map(([name, tool]) => [
+        name,
+        typeof tool?.execute === 'function'
+          ? { ...tool, execute: (input: unknown, ctx: any) => tool.execute(input, { ...ctx, mastra: this.#mastra }) }
+          : tool,
+      ]),
+    ) as typeof resolvedTools;
+    voice?.addTools(wrappedToolsForVoice);
     const instructions = await this.getInstructions({ requestContext });
     voice?.addInstructions(this.#convertInstructionsToString(instructions));
     return voice;


petrchiriboga:
Thanks for picking this up @Aayush-engineer. One scope clarification before this lands:

Our runtime uses the **resolver form** of `voice`, not a static instance:

```ts
voice: ({ requestContext, mastra }) => new GeminiLiveVoice({ ... })
```

Required for multi-user websocket sessions — each connection gets its own voice instance instead of mutating a shared one.

Both hunks in this PR are gated to the static branch (`typeof this.#voice !== 'function'`). The resolver branch in `getVoice()` (lines 1972–1977) returns `resolved` immediately with no wrapping. The resolver receives `mastra` as a factory argument, but that doesn't flow into `tool.execute(args, ctx)` later — providers build `ctx` themselves at function-call time, so `ctx.mastra` stays `undefined`. The repro still fails after this PR if you change `voice: new StubVoice()` to `voice: () => new StubVoice()`.

Two adjacent things worth flagging:

1. **Instructions parity.** Static branch calls `addInstructions` after `addTools` (line 1983); resolver branch doesn't. Same class of bug.
2. **Tool-source clobbering.** If the agent calls `addTools` on a resolver-returned instance, it overwrites any tools the user wired up in the factory — `addTools` is a full replacement in every provider (e.g. `voice/aws-nova-sonic/src/index.ts:1670`, Gemini Live).

Question for maintainers / @Aayush-engineer — preferred shape: extend this PR to wrap tools (and lift `addInstructions`) on the resolver branch too for parity in one change, or land this PR as a narrow static-voice fix and track the resolver path in a follow-up issue with an adapted repro?

Aayush-engineer:
Thanks for the thorough analysis @petrchiriboga. I attempted to extend the fix to the resolver branch but hit an existing test at src/agent/__tests__/voice.test.ts — does not post-mutate a resolver-produced instance — which explicitly asserts that addTools and addInstructions should NOT be called on a resolver-produced instance. That test was added for issue #17262 and represents an intentional design decision to avoid clobbering tools the user wired up in the factory.
So the resolver path fix has a real conflict: if we call addTools on the resolver instance we break the no-mutate contract and clobber user-configured tools (as you flagged in point 2). The right fix there is likely a different approach — e.g. wrapping at the provider's invokeTool level, or passing a mastraToolWrapper into the resolver factory — which needs a maintainer decision on the preferred shape.
I'd suggest landing this PR as the narrow static-voice fix and opening a follow-up issue for the resolver path with the adapted repro. Happy to open that follow-up if maintainers agree.

intojhanurag:
Hey @Aayush-engineer , we will handle this by own . Thanks for the PR .

PR #18336, opened 2026-06-23
fix(pg): make PostgresStore.init() concurrency-safe and honor disableInit

### LINKED ISSUE #18282: [BUG] How to use `PostgresStore.init()` Concurrent PostgresStore.init() fails on shared store with RoutingDbClient already has a pinned client
## Description

We are using Mastra in a Next.js/Fly.io server environment.

Each chat request creates a request-scoped Mastra app instance because the runtime needs request-specific options such as:

- model
- user API key
- sandbox
- callbacks
- pubsub
- worker lifecycle

However, all request-scoped Mastra instances share a singleton `PostgresStore` / `Memory` runtime to avoid creating a new `pg.Pool` per request.

Under concurrent requests, multiple Mastra instances appear to call `storage.init()` on the same shared `PostgresStore` at the same time. With `@mastra/pg@1.14.0`, this can fail because `PostgresStore.init()` pins a `RoutingDbClient`, and the second concurrent init attempts to pin again.

## Error

```txt
Storage init failed; will retry on next storage call {
  error: Error: RoutingDbClient already has a pinned client
      at PostgresStore.init (...)
      at async ...
  id: 'MASTRA_STORAGE_PG_INIT_FAILED',
  domain: 'STORAGE',
  category: 'THIRD_PARTY',
  [cause]: Error: RoutingDbClient already has a pinned client
      at RoutingDbClient.pin (...)
      at PostgresStore.init (...)
}
```

## Environment

- `@mastra/core`: `1.45.0`
- `@mastra/pg`: `1.14.0`
- `@mastra/memory`: `1.21.0`
- `mastra` CLI/package: `1.15.0`
- Runtime: Next.js server runtime on Fly.io
- Database: Supabase Postgres
- Storage: `PostgresStore`
- `disableInit: true` is set because schema/migrations are managed outside the app runtime

## Our architecture

We intentionally reuse the same `PostgresStore` instance:

```ts
const storage = new PostgresStore({
  id: 'howone-memory',
  connectionString,
  schemaName: 'agent',
  disableInit: true,
  ssl,
  max,
  idleTimeoutMillis,
})

const memory = new Memory({
  storage,
  vector,
  options,
})

// Stored globally and reused across request-scoped Mastra instances.
globalThis.__memoryRuntime ??= { storage, memory, vectors }
```

Each request creates a Mastra instance using that shared memory/storage:

```ts
const mastra = new Mastra({
  agents,
  storage: memoryRuntime.storage,
  vectors: memoryRuntime.vectors,
  // request-specific model/sandbox/pubsub/callbacks...
})
```

The reason we create a request-scoped Mastra instance is that each request may have different model/user API key/sandbox/callbacks.

The reason we share `PostgresStore` / `Memory` is to avoid creating a new `pg.Pool` per request.

## Important question about `init()`

In our application, we do not want Mastra to manage database schema creation or migrations at runtime.

We already create/manage the database schema separately, and the runtime database user is not intended to perform DDL. That is why we set:

```ts
disableInit: true
```

So we want to clarify the intended lifecycle:

1. If `disableInit: true`, is `PostgresStore.init()` still expected to run?
2. Does `PostgresStore.init()` do anything required besides schema/table/index/migration setup?
3. In our architecture, where schema is managed externally and we only need runtime memory/background-task persistence, do we need `PostgresStore.init()` at all?
4. Is it safe/recommended to make `init()` a no-op when `disableInit: true`, or does Mastra rely on `init()` for non-DDL storage readiness?
5. If `init()` must still run, should it be single-flight/concurrency-safe on the same `PostgresStore` instance?

## Expected behavior

One of these should be true:

1. `PostgresStore.init()` should be safe to call concurrently on the same store instance.
2. `disableInit: true` should avoid the pinned init path entirely if no DDL/migration is needed.
3. The docs should recommend a different lifecycle pattern for server apps that need request-scoped Mastra instances but shared DB-backed memory/storage.

## Actual behavior

Concurrent request-scoped Mastra instances sharing the same `PostgresStore` can trigger overlapping `storage.init()` calls.

The second init fails with:

```txt
RoutingDbClient already has a pinned client
```

## Current workaround

We are considering wrapping `PostgresStore.init()` with a single-flight guard in our app:

```ts
class AppPostgresStore extends PostgresStore {
  private initPromise: Promise<void> | null = null

  override init(): Promise<void> {
    this.initPromise ??= super.init().catch((error) => {
      this.initPromise = null
      throw error
    })

    return this.initPromise
  }
}
```

This ensures concurrent calls share the same initialization promise.

## Questions

1. Is sharing one `PostgresStore` / `Memory` instance across multiple request-scoped Mastra instances the recommended pattern for Next.js/server environments?
2. Should `PostgresStore.init()` itself be single-flight / concurrency-safe?
3. Should `disableInit: true` skip the pinned-client section in `PostgresStore.init()`?
4. Does a runtime app with externally managed schema need `PostgresStore.init()` at all?
5. Is the workaround above safe, or is there a better recommended lifecycle pattern?
6. Should each Mastra instance get its own `PostgresStore`, even though that creates more `pg.Pool` instances and risks exhausting DB connections?

## Related context

This looks related to the pinned-client changes from:

https://github.com/mastra-ai/mastra/pull/17833

That PR appears to pin DDL to a single backend during `PostgresStore.init()`, but it may not guard against concurrent `init()` calls on the same store instance.

### Steps To Reproduce

1. Create a singleton `PostgresStore` and `Memory` runtime at module/global scope.

2. Reuse that same `PostgresStore` across multiple request-scoped `Mastra` instances.

3. Trigger two or more requests concurrently so both `Mastra` instances attempt to use the shared storage at startup.

4. Mastra internally calls `storage.init()` for each instance.

5. The concurrent `PostgresStore.init()` calls overlap.

none

### Expected Behavior

none

### Environment Information

```block
none
```

### Verification

mdkulkarni2005:
Fixes #18282

## Problem

  When a single `PostgresStore` is shared across concurrent request-scoped
  Mastra instances, multiple `init()` calls race past the `isInitialized`
  check and both try to pin the routing client, throwing:

  > RoutingDbClient already has a pinned client

  `disableInit: true` also still ran the connect-and-pin path, even though
  the schema is managed externally.

  ## Changes

  - **Single-flight `init()`** — concurrent calls now share one cached
    `#initPromise` instead of each pinning the client, so the second caller
    no longer throws.
  - **`disableInit` is honored** — `init()` returns early (no connect, no
    pin, no DDL) when `disableInit: true` or `MASTRA_DISABLE_STORAGE_INIT=true`.
  - **Retryable on failure** — `connect()` now runs inside the `try`, so a
    failed connection resets the cached promise instead of permanently
    poisoning the store.

  ## Tests

  - Concurrent `init()` calls resolve without error and create tables once.
  - `disableInit: true` performs zero `pool.connect()` calls and no DDL.
  - A transient `connect()` failure rejects, and a later `init()` retries
    successfully.

## ELI5
If many requests try to “set up” the same shared Postgres-backed storage at once, they can conflict and crash during startup. This PR makes `PostgresStore.init()` run only once per store instance (others wait), and it can fully skip initialization when `disableInit: true` is configured.

---

### PostgresStore concurrency fix
- **Single-flight init**: `PostgresStore.init()` caches an in-progress initialization promise (`#initPromise`), so concurrent calls on the same instance coalesce into one initialization.
- **Avoid pinned-client conflicts**: refactors the pinned-client/DDL path into `#runPinnedInit()` so only one caller acquires/pins/releases the routing DB client at a time, preventing “RoutingDbClient already has a pinned client”.
- **Retry after transient failures**: if pinned initialization fails, `#initPromise` is cleared so later `init()` calls can retry.

### Honor `disableInit`
- `init()` is now a **true no-op** when `disableInit: true` (or `MASTRA_DISABLE_STORAGE_INIT === 'true'`), skipping the connect/pin/schema-setup behavior entirely.

### Test coverage (issue `#18282`)
- Adds regression tests that:
  - concurrent `store.init()` calls on the same store instance resolve without throwing and still create the expected schema/table (e.g., `threads`);
  - `disableInit: true` prevents `pool.connect()` from being called and prevents the `threads` table from being created;
  - a transient one-time `pool.connect()` failure doesn’t permanently break initialization (a subsequent `init()` succeeds).

### Changeset
- Updates the `@mastra/pg` package with the concurrency-safe `PostgresStore.init()` and `disableInit` no-op behavior.


intojhanurag (changes requested):
One test is broken though:

`stores/pg/src/storage/index.test.ts:721` , `.resolves.toBeDefined()` always fails because `init()` returns `Promise<void>` (resolves to `undefined`). The concurrent test on line 659 passes by accident since `Promise.all()` returns an array, which *is* defined.

Replace with just `await store.init();` or `.resolves.toBeUndefined()` and this is good to go.

intojhanurag (approved):

mdkulkarni2005:
---
  Hi @intojhanurag, thanks for the review! I've pushed the fix — replaced .resolves.toBeDefined() with .resolves.toBeUndefined() on the
  init() retry test so it correctly matches the Promise<void> return.

  Regarding the failing CI checks — they look unrelated to this PR, which only touches stores/pg:

  - test (pinecone) — fails with PineconeConfigurationError: must have required property: apiKey. This is a missing CI secret, not a code
  issue.
  - test (lance) — fails with a Table ... was not found error during test cleanup, which looks like a pre-existing flaky test.
  - @mastra/redis-streams#lint — passes clean locally (eslint . → exit 0) on this branch; this package isn't modified by the PR.

  None of these are in stores/pg. Could you re-run the failed jobs (or let me know if anything on my end needs adjusting)? Happy to make
  any further changes. Thanks!

mdkulkarni2005:
---
  Update on the failing checks — I dug into each and they're all unrelated to this PR (which only touches stores/pg):

  - Lint (@mastra/redis-streams) — the floating-promise error in agent-thread-worker.entry.ts:83 was introduced by #17723 (June 19), not
  this branch. The file is identical on main, so main and other open PRs are hitting the same failure.
  - test (pinecone) — fails on PineconeConfigurationError: must have required property: apiKey, i.e. a missing CI secret.
  - test (lance) — flaky Table ... was not found during test cleanup.

  None are in stores/pg. Could you re-run / advise whether anything on my side is needed? Happy to help. Thanks!

  ---

PR #18350, opened 2026-06-23
fix(chroma): return per-metric similarity score from query()

### LINKED ISSUE #18316: [BUG] Chroma vector store computes `score = 1 - distance` for every metric — wrong (negative/unbounded) for euclidean, and missing distances become perfect scores
### Describe the Bug

`ChromaVector.createIndex()` supports three metrics (`cosine`, `euclidean` → `l2`, `dotproduct` → `ip`), but `query()` converted distance to score with the cosine-only formula `1 - distance` unconditionally. For an `euclidean` collection, Chroma returns squared-L2 distances in `[0, ∞)`, so Mastra returned scores in `(-∞, 1]` — e.g. a near match at distance `0.5` scored `0.5` while a far match at distance `25` scored `-24`.

Ordering happens to survive (the conversion is monotonic), but:

- The documented, portable "score in 0..1, higher is better" contract is broken — `minScore`-style threshold code and rerank weighting (the `vector` weight multiplies the score in `@mastra/rag`'s `rerank`) produce garbage on negative/unbounded values.
- The same query on `@mastra/pg` (euclidean → `1/(1+d)`) or `@mastra/duckdb` yields a completely different score range, so store portability silently breaks.

Additionally, `const distance = results.distances?.[0]?.[index] || 0` maps a *missing* distance to `0`, i.e. a **perfect score of 1** — absence of information becomes maximal similarity.

`stores/chroma/src/vector/index.ts` (in `query()`):

```ts
return (results.ids[0] || []).map((id: string, index: number) => {
  // Chroma returns distances (lower = more similar), convert to similarity scores (higher = more similar)
  // For cosine: similarity = 1 - distance
  const distance = results.distances?.[0]?.[index] || 0;
  const score = 1 - distance;
  ...
});
```

The metric is available at query time (`createIndex` maps it via `spaceMappings` / `hnsw: { space }`, and `describeIndex` recovers the space from `collection.configuration.hnsw?.space`). The correct per-metric pattern already exists in-repo at `stores/duckdb/src/vector/index.ts` and `stores/pg/src/vector/index.ts`.

### Steps To Reproduce

```ts
import { ChromaVector } from '@mastra/chroma';

const chroma = new ChromaVector({ id: 'repro' /* + client args */ });
await chroma.createIndex({ indexName: 'i', dimension: 2, metric: 'euclidean' });
await chroma.upsert({
  indexName: 'i',
  vectors: [
    [0, 0], // id 'near'
    [3, 4], // id 'far'
  ],
  ids: ['near', 'far'],
});

const results = await chroma.query({ indexName: 'i', queryVector: [0, 0], topK: 2 });
console.log(results.map(r => ({ id: r.id, score: r.score })));
```

1. Run the snippet above.
2. Observe `near` has `score: 1` and `far` has `score: -24` (`1 - 25`, squared-L2).
3. Any consumer filtering `score > 0.5`, or feeding scores into `rerank`, misbehaves on the negative value.
4. The same data/metric on `@mastra/pg` yields `near: 1`, `far: ~0.038`.

Reproducible directly against `@mastra/chroma` with the euclidean create / upsert / query sequence above.  https://github.com/mastra-ai/mastra

### Expected Behavior

Scores normalized per metric, consistent with pg/duckdb:

- cosine → `1 - distance`
- euclidean → `1 / (1 + distance)` (maps into `(0, 1]`)
- dotproduct → `1 - distance` (Chroma's `ip` distance is `1 - <a, b>`, so this recovers the inner product)

A missing distance should yield a missing/0 score, not a perfect 1.

### Environment Information

```block
System:
    Memory: 4.10 GB / 15.49 GB
    Container: Yes
    Shell: 5.2.21 - /bin/bash
  Binaries:
    Node: 20.19.6 -
    Yarn: 1.22.22 - 
    npm: 10.8.2 -
```

### Verification

rachit367:
## Summary

`ChromaVector.query()` converted Chroma's distance to a `score` using the cosine-only formula `1 - distance` for **every** metric. This is only correct for cosine:

- **euclidean** collections returned scores in `(-∞, 1]` (Chroma's `l2` space returns squared distances, so a far match at distance 25 scored `-24`).
- **dotproduct** collections were mis-scaled.
- The `results.distances?.[0]?.[index] || 0` fallback turned a *missing* distance into a **perfect score of 1**.

Ordering happened to survive for euclidean (the conversion is monotonic), but the documented "score in 0..1, higher is better" contract was broken — `minScore` thresholds, rerank `vector` weighting in `@mastra/rag`, and cross-store portability all produce garbage on negative/unbounded values.

Closes #18316.

## Fix

The metric is already available on the collection (`configuration.hnsw.space`), so `query()` now derives the score per-metric, matching the convention used across the other Mastra vector stores (and the recently merged LanceDB fix #18104):

| metric | conversion |
|---|---|
| cosine | `1 - distance` |
| dotproduct | `1 - distance` (recovers the dot product, matching `@mastra/pg`) |
| euclidean | `1 / (1 + sqrt(distance))` (Chroma `l2` returns squared distance) |

A missing distance now scores `0` instead of `1`. The conversion lives in a small pure helper (`distance-to-score.ts`) so it's unit-testable without a live Chroma.

## Tests

- New unit tests for `distanceToScore` (cosine / euclidean / dotproduct + monotonicity).
- New integration tests asserting the closer vector outscores the far one on both cosine and euclidean indexes, and that euclidean scores stay non-negative.
- Full `@mastra/chroma` suite green against a live Chroma container (171 passed, 6 skipped). Typecheck, eslint, and prettier all clean.

## ELI5
Chroma gives you a “distance” (how far things are), but Mastra expects a “score” where higher means more similar. This PR fixes the math so each distance type gets converted with the right formula—so scores stay in the 0..1 range and missing distances don’t pretend to be perfect matches.

## Summary
This PR fixes `ChromaVector.query()` to correctly convert Chroma distances into normalized similarity scores (0..1, higher is better). Previously, the code always used `score = 1 - distance`, which is only correct for cosine distance; for euclidean (squared L2) it produced unbounded negative scores, and for dotproduct it produced incorrectly scaled scores. It also treated missing distances as `0`, which became a perfect score of `1`, effectively turning “missing” into “best match.”

The fix:
- Determines the metric configured for the Chroma collection from `configuration.hnsw.space` / `configuration.spann.space` (defaulting to `cosine`)
- Converts distances to scores with metric-specific logic via a new helper:
  - **cosine**: `1 - distance`
  - **dotproduct**: `1 - distance`
  - **euclidean** (squared L2): `1 / (1 + sqrt(distance))` (keeps scores bounded in `[0, 1]`)
- Treats `null`/`undefined` distances as score **0** (not 1)

This restores the documented “score in 0..1, higher is better” contract, improving correctness for downstream uses like `minScore` filtering and reranking.

## Changes
- **`stores/chroma/src/vector/distance-to-score.ts`**: Added pure `distanceToScore(distance, metric)` helper
- **`stores/chroma/src/vector/distance-to-score.test.ts`**: Added unit tests covering cosine/euclidean/dotproduct conversions and euclidean monotonicity/non-negativity
- **`stores/chroma/src/vector/index.ts`**: Updated `query()` to resolve the configured metric from collection configuration and convert each returned distance with `distanceToScore` (missing → 0)
- **`stores/chroma/src/vector/index.test.ts`**: Added tests asserting correct higher-is-better ordering for cosine and euclidean
- **`.changeset/chroma-per-metric-score.md`**: Updated changeset to document the user-visible scoring fix with a patch bump

## Testing
Full test suite validation: **171 passed, 6 skipped**.

### DIFF
--- a/.changeset/chroma-per-metric-score.md
+++ b/.changeset/chroma-per-metric-score.md
@@ -1,0 +1,5 @@
+---
+'@mastra/chroma': patch
+---
+
+Fixed `ChromaVector.query()` returning wrong similarity scores for non-cosine collections. Scores were always computed as `1 - distance`, which is only correct for cosine. Euclidean collections returned unbounded negative scores, breaking `minScore` thresholds and rerank weighting and making scores incomparable with other vector stores. Scores are now derived from the collection's actual metric (`euclidean` → `1 / (1 + sqrt(distance))` since Chroma's `l2` space returns squared distances, `dotproduct` → `1 - distance`, `cosine` → `1 - distance`), matching the convention used across the other Mastra vector stores. A missing distance now scores `0` instead of a perfect `1`.
--- a/.changeset/chroma-per-metric-score.md
+++ b/.changeset/chroma-per-metric-score.md
@@ -2,4 +2,9 @@
 '@mastra/chroma': patch
 ---
 
-Fixed `ChromaVector.query()` returning wrong similarity scores for non-cosine collections. Scores were always computed as `1 - distance`, which is only correct for cosine. Euclidean collections returned unbounded negative scores, breaking `minScore` thresholds and rerank weighting and making scores incomparable with other vector stores. Scores are now derived from the collection's actual metric (`euclidean` → `1 / (1 + sqrt(distance))` since Chroma's `l2` space returns squared distances, `dotproduct` → `1 - distance`, `cosine` → `1 - distance`), matching the convention used across the other Mastra vector stores. A missing distance now scores `0` instead of a perfect `1`.
+Fixed similarity scoring in `ChromaVector.query()` for non-cosine indexes.
+
+- Euclidean indexes now return bounded, positive scores instead of unbounded negative ones.
+- Dotproduct indexes now use the correct score conversion.
+- A missing distance now scores `0` instead of a perfect `1`.
+- `minScore` filtering and rerank weighting now behave consistently with the other Mastra vector stores.
--- a/stores/chroma/src/vector/distance-to-score.test.ts
+++ b/stores/chroma/src/vector/distance-to-score.test.ts
@@ -1,0 +1,29 @@
+import { describe, expect, it } from 'vitest';
+
+import { distanceToScore } from './distance-to-score';
+
+describe('distanceToScore', () => {
+  it('converts cosine distance with 1 - distance', () => {
+    expect(distanceToScore(0, 'cosine')).toBe(1);
+    expect(distanceToScore(0.25, 'cosine')).toBeCloseTo(0.75);
+    expect(distanceToScore(2, 'cosine')).toBe(-1);
+  });
+
+  it('maps euclidean (squared L2) distance into a bounded 0..1 score', () => {
+    expect(distanceToScore(0, 'euclidean')).toBe(1);
+    expect(distanceToScore(1, 'euclidean')).toBeCloseTo(0.5);
+    expect(distanceToScore(25, 'euclidean')).toBeCloseTo(1 / 6);
+  });
+
+  it('keeps euclidean scores monotonic and never negative', () => {
+    const near = distanceToScore(0.5, 'euclidean');
+    const far = distanceToScore(25, 'euclidean');
+    expect(near).toBeGreaterThan(far);
+    expect(far).toBeGreaterThan(0);
+  });
+
+  it('recovers the dot product from dotproduct distance with 1 - distance', () => {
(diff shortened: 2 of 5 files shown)


rachit367:
Thanks @coderabbitai — addressed in f7a262e. Rewrote the changeset into short, outcome-focused bullets per the `.changeset` path guidelines, dropping the formula/implementation detail from the release notes.

abhiaiyer91 (approved):

PR #18389, opened 2026-06-23
docs(agents): clarify when to use backgroundTasks vs EventedAgent vs DurableAgent

### LINKED ISSUE #17996: [DOCS] Clarify when to use EventedAgent vs DurableAgent/backgroundTasks
## Describe the Improvement Needed

The current docs mention `createEventedAgent` as "fire-and-forget durable execution" and explain that it uses `startAsync()`, but it is still hard to understand when an application should choose it over `DurableAgent.stream()` or Mastra `backgroundTasks`.

From a product integration perspective, `DurableAgent.stream()` already returns a `runId`, supports streaming output, and provides `observe(runId)` for reconnecting to an existing run. For long-running work inside an agent conversation, `backgroundTasks` also seem like a natural fit.

Could the docs clarify the intended decision boundary?

Questions that would help:

1. What concrete scenarios should prefer `EventedAgent` over `DurableAgent.stream()`?
2. What user-facing behavior, if any, does `EventedAgent` enable that `DurableAgent` does not?
3. Is `EventedAgent` mainly intended for fire-and-forget whole-agent runs, rather than interactive chat flows?
4. When should developers use `EventedAgent` instead of Mastra `backgroundTasks` for long-running or externally-triggered work?

A short comparison table or scenario guide for `Agent.stream()`, `DurableAgent.stream()`, `EventedAgent`, and `backgroundTasks` would make this much easier to reason about.

## Documentation Type

- Unclear Explanation
- Missing Information

aryanbarde80:
## What does this PR do?

Adds a comparison section to `docs/agents/background-tasks.mdx` clarifying when to use `backgroundTasks`, `EventedAgent`, and `DurableAgent`.

Closes #17996

## Problem

The three async execution APIs overlap in purpose but solve very different problems. Users were confused about when to reach for each one. Issue #17996 explicitly requested this clarification.

## Change

Added a `## Background tasks vs EventedAgent vs DurableAgent` section before the existing `## Quickstart` in `background-tasks.mdx`:

- **Comparison table** on 5 dimensions: scope of what runs async, user-visible streaming, process-restart survival, external engine requirement, and run reconnection
- **Prose** explaining the right use case for each API
- **Note** clarifying that `createEventedAgent` is a factory helper for existing `Agent` instances

## Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Test coverage improvement
- [x] Documentation update

## Notes

Docs only — 1 file changed, 24 lines added, no changeset required. All facts verified against source code in `packages/core/src/agent/durable/`.

## ELI5

This PR adds helpful guides to the documentation so developers can understand which tool to use when they need something to keep running in the background. It explains the differences between three similar but different tools (backgroundTasks, EventedAgent, and DurableAgent) and when to pick each one, plus it adds information about setting up separate security settings for the Studio interface versus the API.

### Background Tasks vs EventedAgent vs DurableAgent Documentation

Added a new "Background tasks vs EventedAgent vs DurableAgent" section to `docs/agents/background-tasks.mdx` that directly addresses user confusion about choosing between three overlapping async execution APIs. The section includes:

- A comparison table evaluating the three approaches across five key dimensions:
  - Scope of async execution
  - User-visible streaming capability
  - Process-restart survival
  - External engine requirement
  - Run reconnection capability

- Prose guidance explaining the appropriate use case for each API
- A clarifying note that `createEventedAgent` is a factory helper for wrapping existing `Agent` instances with `EventedAgent`, rather than a separate execution mode

(+24 lines)

### Studio and API Authentication Documentation

Added a new "Separate Studio and API authentication" section to `docs/studio/auth.mdx` explaining how to configure different authentication providers for Studio versus the API. The section covers:

- How `server.auth` normally secures both Studio and API routes
- How `studio.auth` can override this for Studio-specific authentication
- Configuration example showing external-customer authentication on `server.auth` and internal-team authentication on `studio.auth`
- Fallback behavior when `studio.auth` is not configured
- Note that any `server.auth` provider can also be used for `studio.auth`

(+41 lines)

### DIFF
--- a/docs/src/content/en/docs/agents/background-tasks.mdx
+++ b/docs/src/content/en/docs/agents/background-tasks.mdx
@@ -28,6 +28,30 @@ Background tasks require a configured [storage](/docs/memory/storage) backend on
 
 :::
 
+## Background tasks vs EventedAgent vs DurableAgent
+
+These three APIs all run work asynchronously but solve different problems:
+
+| | `backgroundTasks` | `EventedAgent` | `DurableAgent` |
+| - | - | - | - |
+| **What runs async** | A single tool call | The whole agent run | The whole agent run |
+| **User sees streaming output** | Yes, via `streamUntilIdle()` | Yes, via pubsub | Yes, via pubsub |
+| **Survives process restart** | No | No | Yes (external engine required) |
+| **External engine needed** | No | No | Yes (e.g. Inngest) |
+| **Reconnect to existing run** | No | Yes, via `observe(runId)` | Yes, via `observe(runId)` |
+
+**Use `backgroundTasks`** when one specific tool call is slow but the agent should keep responding to the user. The rest of the agentic loop runs normally and `streamUntilIdle()` keeps the stream open until the tool result is processed.
+
+**Use `EventedAgent`** when the entire agent run should be fire-and-forget and you do not need a durable external engine. `EventedAgent` uses the built-in evented workflow engine — execution is non-blocking via `startAsync()` and events are streamed through the internal pubsub. Runs are resumable via `observe(runId)` but do **not** survive a process restart.
+
+**Use `DurableAgent`** when the agent run must survive process restarts. This requires an external durable workflow engine such as [Inngest](https://inngest.com). Runs are checkpointed and resume automatically after a server crash.
+
+:::note
+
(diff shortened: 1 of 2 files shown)


intojhanurag:
there is some factual error .

intojhanurag:
we will handle by own .

PR #18393, opened 2026-06-23
fix(mongodb): hardening and data-modeling improvements of storage and vector domains

### LINKED ISSUE #18584: MongoDB store: multiple data-integrity and silent-failure bugs in storage and vector domains
A production readiness review of `stores/mongodb/src/` identified several correctness bugs across the storage and vector domains. These affect any deployment using the MongoDB store adapter. All are addressed in PR #18393.

---

## 1. Multi-document operations are not atomic — partial writes leave orphaned records

Creating or deleting a versioned entity (agents, MCP clients/servers, prompt-blocks, scorer-definitions, skills, workspaces, schedules) involves writes to two or more collections. If the process crashes or an error is thrown between writes, the store is left in an inconsistent state with no recovery path except manual cleanup:

- `agents.create()` — agent row inserted, version insert fails → orphaned agent with no version
- `agents.delete()` — versions deleted, agent `deleteOne` fails → versions gone, agent row remains
- Same pattern for MCP clients/servers, prompt-blocks, scorer-definitions, skills, workspaces, and schedules
- `memory.saveMessages()` — messages bulk-written, thread `updatedAt` update fails → messages with no timestamp refresh

`memory.deleteThread()` and `datasets.deleteDataset()` also involve multi-collection cascade deletes (messages + thread row; versions + items + dataset row). These are intentionally **not** made atomic in the fix: the child collections are unbounded, and wrapping an unbounded `deleteMany` in a transaction would cause large threads or datasets to hit MongoDB's 60-second `transactionLifetimeLimitSeconds` limit, making them permanently undeletable. Instead the fix deletes children first and the parent row last, so a crash mid-cascade leaves the parent in place and a subsequent call completes the job idempotently.

---

## 2. `batchCreateSpans` silently drops spans on duplicate key

`batchCreateSpans` calls `insertMany` with the default `ordered: true`. When a duplicate-key error (code 11000) occurs — expected in at-least-once delivery — MongoDB aborts the remainder of the batch. Every span after the first duplicate is silently dropped with no error surfaced to the caller.

---

## 3. `saveMessages` leaves stale `updatedAt` on all but the first thread in a mixed-thread batch

When `saveMessages` is called with messages belonging to more than one thread, only `messages[0].threadId`'s `updatedAt` is refreshed. All other threads in the batch silently retain stale timestamps, breaking recency ordering for those threads.

---

## 4. `dangerouslyClearAll` does not clear observational memory

`mastra_observational_memory` is absent from the `deleteMany` set in `dangerouslyClearAll`. A full storage reset leaves stale observational memory records in place.

---

## 5. `updateWorkflowState` silently no-ops for snapshots without `context`

The `findOneAndUpdate` filter includes `'snapshot.context': { $exists: true }`. Any workflow snapshot that lacks a `context` field matches no document, so the update silently returns `undefined`. The workflow state is lost with no error raised.

---

## 6. `getCollection` caches phantom handles for non-existent collections

`getCollection` in the vector store caches the returned handle unconditionally, including when `collectionExists = false`. A subsequent call with `throwIfNotExists: true` retrieves the cached phantom handle and returns it silently rather than throwing, causing callers to operate on a handle that refers to a non-existent collection.

caseyclements:
## Description

Production readiness review and hardening of `stores/mongodb/src/` covering the
vector store, memory, workflows, and observability storage domains. The review
identified 17 findings (denoted F1, F2, ..). The PR addresses the ones with clear correct behavior
and a deterministic test signal. Three concurrency races (F3, F4) and one
volume-dependent failure (F6) are documented as known issues with test stubs for
follow-up.

A follow-on pass (Tranches A+B) hardens the memory domain's data modeling — index
design and native metadata storage — and adds topology-aware transactions so
multi-document writes are atomic on a replica set while still working on a
standalone server.

**Storage domain fixes:**

- **F5 — `batchCreateSpans` batch abort on duplicate key**: Changed `insertMany`
  from default `ordered: true` to `ordered: false` and added handling for
  `MongoBulkWriteError` to skip duplicate-key errors (code 11000). In
  at-least-once delivery, a single duplicate previously caused all subsequent
  spans in the batch to be silently dropped.

- **F10 — `saveMessages` stale `updatedAt` on multi-thread batches**: When a
  single `saveMessages` call contained messages for more than one thread, only
  `messages[0].threadId`'s `updatedAt` was refreshed. Now collects all distinct
  thread IDs in the batch and updates each one.

- **F13 — `dangerouslyClearAll` missing `OM_TABLE`**: The observational memory
  table (`mastra_observational_memory`) was absent from the reset, leaving stale
  OM records after a full storage clear. Added to the parallel `deleteMany` set.

- **F14 — `updateWorkflowState` silent drop**: The `findOneAndUpdate` filter
  included `'snapshot.context': { $exists: true }`, causing any workflow
  snapshot without a `context` field to silently return `undefined` instead of
  updating. Removed the filter and added a post-update check that throws a
  `MastraError` when the returned snapshot has no `context`.

**Vector domain fixes:**

- **F15 — `getCollection` phantom handle caching**: `getCollection` previously
  cached collection handles unconditionally, including when
  `collectionExists = false`. A subsequent call with `throwIfNotExists: true`
  would return the cached phantom handle silently instead of throwing. Now only
  caches handles for collections that actually exist.

- **F12 — `createIndex` async index lifecycle (documented)**: Atlas Search
  indexes transition through PENDING → BUILDING → READY after `createIndex`
  returns. Added JSDoc to `createIndex` explaining that callers must call
  `waitForIndexReady` before querying. The existing `createIndexAndWait` test
  helper already does this correctly.

- **`getCollection` docstring**: Replaced confusing inline comments with a
  JSDoc that explains the Mastra index ↔ MongoDB collection mapping (they are
  distinct concepts) and the caching invariant.

**TypeScript fix:**

- `MongoBulkWriteError.writeErrors` is typed as `OneOrMore<WriteError>`
  (`T | ReadonlyArray<T>`), which does not expose `.filter()`. Normalized via
  `Array.isArray` before filtering.

**Data modeling & transactions (memory domain — Tranches A+B):**

- **A1 — default index audit (15 → 9)**: Audited `getDefaultIndexDefinitions()`
  against the actual query shapes (`listThreads`, `listMessages`,
  `listMessagesByResourceId`, and the resource/OM operations). Dropped redundant
  single-field indexes whose prefixes are already covered by a compound, dropped
  unused sort indexes, and **added** the missing `{ resourceId: 1, createdAt: 1 }`
  that `listMessagesByResourceId` needs. Net: lower write amplification with
  better query coverage. (`createDefaultIndexes` only creates indexes, so pruning
  the now-redundant indexes from existing deployments is a follow-up drop
  migration.)

- **A3 — resource metadata as native BSON**: `saveResource`/`updateResource` no
  longer `JSON.stringify` resource metadata, matching how thread metadata is
  already stored and enabling dot-notation queries. The read path keeps a
  string-or-object fallback for back-compat with existing rows.

- **A4 — fail loud on index-creation failure**: `createDefaultIndexes` now throws
  a `MastraError` instead of logging a warning and continuing, so a missing index
  (a latent performance regression at scale) surfaces at `init()`. The
  `skipDefaultIndexes` flag remains the escape hatch for self-managed indexes.

- **B1–B3 — topology-aware transactions**: Added `supportsTransactions()`
  (probes `hello`, cached only on success) and `withTransaction()` to the
  connector. `deleteThread` (delete messages + thread) and `saveMessages`
  (message bulk-write + per-thread `updatedAt`) now run atomically inside a
  transaction on a replica set or sharded cluster, and degrade to sequential
  best-effort on a standalone server — no replica set is required to run the
  store.

**Test coverage:** Each F-finding fix has a corresponding test in the `Hardening: NODE-7556`
describe block that fails without the fix and passes with it. Several `test.todo`
stubs document the known issues that cannot be
exercised deterministically in CI against atlas-local. The Tranches A+B work adds
`hardening-data-modeling.test.ts` and `hardening-transactions.test.ts`, which
exercise both the transaction path (atlas-local replica set, port 27018) and the
degrade path (standalone, port 27017).

> This PR was developed with assistance from an AI coding agent (Claude Code).
> All code was reviewed and verified by Casey Clements.

## Related issue(s)

Fixes #18584

## Type of change

- [ ] Documentation update
- [ ] Code refactoring
- [x] Performance improvement
- [x] Test update

## Checklist

## ELI5
This PR makes the MongoDB storage layer act more like a “save first, or save nothing” system, so errors don’t leave half-written data behind. It also fixes a handful of tricky edge cases (timestamps, cleanup completeness, vector index readiness, and memory indexing) so behavior is consistent and observable.

## What changed
- **Topology-aware transactions:** Added connector support to run multi-document operations in a MongoDB transaction on replica sets/sharded clusters, with safe sequential fallback on standalone servers.
- **Atomic hardening across storage domains:** Refactored key create/update/delete flows (agents, MCP clients/servers, workspaces, prompt blocks, scorer definitions, skills, datasets, etc.) to perform related writes inside a shared transaction/session.
- **Best-effort cascade deletes:** Kept large cascade-delete paths out of transactions, deleting children first (e.g., threads/messages, dataset items) and parent last so retries remain safe after mid-delete failures.
- **Observability memory cleanup:** Expanded `dangerouslyClearAll()` to also clear observational memory.
- **Span creation duplicate-key safety:** Hardened `batchCreateSpans` bulk inserts so duplicate-key errors don’t silently drop data; duplicate handling is explicitly detected.
- **Correct `saveMessages` timestamping:** Updated `saveMessages` so **all touched threads** in mixed-thread batches get their `updatedAt` refreshed.
- **Workflow state correctness:** Updated `updateWorkflowState` to **fail loudly** with a user error when a stored snapshot exists but has no `context`, instead of silently no-oping.
- **Vector store hardening:**
  - Improved `getCollection` caching so non-existent collections aren’t cached as “phantom” handles.
  - Strengthened Atlas Search index readiness and validation (including throwing on missing/misconfigured vector fields).
  - Normalized query/filter behavior (metadata filtering vs document filtering) and added optional `numCandidates` control.
- **Memory domain hardening:**
  - Reduced/audited default index definitions, added the missing `{ resourceId: 1, createdAt: 1 }` index, and verified expected compound index shapes.
  - Stored resource `metadata` as native BSON objects (with backward-compatible parsing for legacy JSON-string metadata).
  - Made default index creation **fail fast** (throw) instead of logging-and-continuing.
  - Added/validated rollback-safe transaction behavior for multi-step memory operations.

## Tests
- Added/updated hardening tests for:
  - `batchCreateSpans` duplicate-key handling
  - `saveMessages` updating `updatedAt` for all touched threads
  - `dangerouslyClearAll` clearing observational memory
  - `updateWorkflowState` failing when snapshot `context` is missing
  - topology-aware transaction behavior (rollback on replica sets; graceful persistence on standalone)
  - dataset/agent cascading + rollback scenarios via injected failures
  - memory index definitions + native BSON metadata persistence + legacy metadata parsing
  - vector index edge cases around deletion/update of Atlas Search index metadata


stores/mongodb/src/vector/index.ts
```diff
@@ -366,31 +395,22 @@ export class MongoDBVector extends MastraVector<MongoDBVectorFilter> {
...
-
         // pre-filter for candidate document IDs
         const candidateIds = await collection
-          .aggregate([{ $match: filterWithExclusion }, { $project: { _id: 1 } }])
+          .aggregate([{ $match: combinedFilter }, { $project: { _id: 1 } }])
           .map(doc => doc._id)
           .toArray();
 
         if (candidateIds.length > 0) {
           vectorSearch.filter = { _id: { $in: candidateIds } };
```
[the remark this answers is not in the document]
caseyclements: This is an issue that was identified in this work. It is noted as F6 in a test.todo. [Issue #118587](https://github.com/mastra-ai/mastra/issues/18587) has been opened on mastra/core to make allow us to filter on metadata fields during vector search. For now, we moved the documentFilter into the vectorSearch so this pre-filter $match aggregation will happen if and only if metadata filter is provided.

caseyclements:
I believe that I've addressed all of CodeRabbit's comments now. Would you please kick off workflows? All tests related to this change pass locally.

.changeset/chatty-clowns-push.md
```diff
@@ -0,0 +1,75 @@
+---
```
abhiaiyer91, line 1: Why do we have all these changesets?

abhiaiyer91 (changes requested):
lets fix the extraneous snapshots and we'll be good

TheIsrael1:
@caseyclements this PR looks like the right remediation path for #18584 and the diff covers the reported MongoDB storage/vector areas. Before maintainer re-review, please address Abhi's requested cleanup around the extraneous snapshots and confirm the targeted MongoDB storage/vector tests pass in a replica-set-capable environment, especially transaction rollback, mixed-thread `saveMessages`, duplicate span insertion, workflow state update, and vector phantom collection cache coverage.

.changeset/chatty-clowns-push.md
caseyclements: Good catch. Only `'@mastra/rag': minor` should be in there, and the change is tiny.
Everything else is in ` '@mastra/mongodb': patch`.
I've pushed the fix.

caseyclements:
> @caseyclements this PR looks like the right remediation path for #18584 and the diff covers the reported MongoDB storage/vector areas. Before maintainer re-review, please address Abhi's requested cleanup around the extraneous snapshots and confirm the targeted MongoDB storage/vector tests pass in a replica-set-capable environment, especially transaction rollback, mixed-thread `saveMessages`, duplicate span insertion, workflow state update, and vector phantom collection cache coverage.

I've just rerun all the storage and vector tests onsite. Note that we run a nightly CI of our integrations into upstream repos to ensure they don't silently drift from main.

abhiaiyer91 (approved):

PR #18396, opened 2026-06-24
fix(core): deliver tool-result images to AI SDK v6 provider models

CalebBarnes:
## Summary

Images returned from tools were silently dropped when an agent's model ran on an AI SDK v6 provider (`@ai-sdk/*` v3, e.g. `@ai-sdk/anthropic@3`). A tool that returns an image through `toModelOutput` — for example a browser screenshot — would reach the model as an empty tool result ("Tool ran without output or errors") instead of an actual image, so the model never saw the picture.

This restores image delivery for v6 providers without regressing v5.

## Why this happened

AI SDK v5 and v6 use mutually exclusive content shapes inside a tool result:

- **v5** (model spec version `v2`) accepts only `media`
- **v6** (model spec version `v3`) accepts only `image-data` / `file-data` — it has no `media` case

Core builds the model prompt with the vendored v5 SDK and therefore emits `media`. v6 normally translates `media` into `image-data`/`file-data` during its own message conversion, but that translation never runs in the v5 conversion path. So the `media` part reached the v6 provider untranslated and was discarded.

## What changed

- Added a `media` → `image-data`/`file-data` translation in the message-list `llmPrompt()` builder, applied after the stored `modelOutput` is reapplied so override-injected media is also covered.
- The translation is gated on the consuming model's `specificationVersion` and runs **only** for `v3` (v6) providers. v5 providers still receive `media` and are unaffected.
- The model's spec version is threaded through from both the evented and durable execution paths.

## Test plan

- New unit tests assert tool-result image parts become `image-data` and non-image media becomes `file-data` for a `v3` model, while the existing v5 test confirms `media` is left untouched.
- `pnpm build:core`, `pnpm --filter ./packages/core check` (typecheck), and the message-list suite pass.
- Verified end-to-end in a v6 (`@ai-sdk/anthropic@3`) agent: a screenshot tool result now reaches the model as a viewable image instead of an empty result.

## ELI5
When an agent’s tool returns an image (or a file), this PR makes sure newer AI SDK providers don’t accidentally turn that tool result into an empty payload. It converts the tool output into the exact “image-data” (for images) or “file-data” (for other files) format the model expects.

## Summary
- Fixed tool-result image/file drop for AI SDK v6 providers (`@ai-sdk/*` with spec `v3`) by adding `media` → `image-data` / `file-data` translation when building the LLM prompt message list.
- The translation is gated by the consuming model’s `specificationVersion` and only runs for `v3` flows, preserving existing `v5`/`v2` behavior that keeps `media`.
- Ensured the media translation happens after stored `modelOutput` is reapplied, so override-injected media is handled correctly.
- Threaded `specificationVersion` through both durable workflow and loop/evented execution paths to select the right prompt builder (`aiV6` path for `v3`, `aiV5` otherwise).
- Implemented a dedicated conversion helper (`aiV5PromptToAIV6Prompt`) that rewrites tool-result content parts from `media` into v6-compatible `image-data`/`file-data`.
- Added tests verifying:
  - image media becomes `image-data`
  - non-image media becomes `file-data`
  - existing `v5` behavior still preserves `media`.
- Added a changeset entry for a patch release of `@mastra/core`.


packages/core/src/agent/message-list/message-list.ts
```diff
@@ -699,6 +586,145 @@ export class MessageList {
...
 
+  /**
+   * Shared implementation behind aiV5.llmPrompt / aiV6.llmPrompt.
+   *
+   * `transformModelMessages`, when provided, runs after the stored-modelOutput
+   * override and before asset download / V2 prompt mapping. aiV6 uses it to
+   * translate tool-result `media` parts into the `image-data`/`file-data` shape
+   * that v6 (spec 'v3') providers require; aiV5 omits it so `media` is preserved.
+   */
+  async #buildLlmPrompt(
```
TylerBarnes: I think we should make the code style of this match the rest of message list, each conversion function handles a single type of conversion and you just map through multiple conversions, so for ex you could convert to v5 and then a second function goes v5->v6

this file is already too complex so I'd avoid adding new patterns

TylerBarnes: @CalebBarnes looks much better but we should also inline this back on line 541 and remove `#buildLlmPrompt`

CalebBarnes: Done in 31d181946c — inlined the body back into `aiV5.llmPrompt` and removed `#buildLlmPrompt`. `aiV6.llmPrompt` stays as the chained `aiV5PromptToAIV6Prompt(await aiV5.llmPrompt(...))`. Build + typecheck clean, 495 message-list tests and both v5/v6 tool-result image e2e tests pass.

TylerBarnes (approved):

PR #18432, opened 2026-06-24
feat(core): surface step providerMetadata to output-step processors

### LINKED ISSUE #18435: Surface step providerMetadata to output processors (e.g. Bedrock guardrail trace on content-filter blocks)
### What

Output-step processors (`processOutputStep`) have no access to the finishing model step's `providerMetadata`. On a content-filter block this means the provider's assessment is unreachable from a processor.

### Context

When an AWS Bedrock Guardrail intervenes, the AI SDK Bedrock provider returns the full guardrail assessment in the provider stream's terminal `finish` part, under `providerMetadata.bedrock.trace.guardrail` (which policy/topic/filter blocked, the action, the confidence).

Today Mastra drops this at the Mastra → output-processor boundary. When `processOutputStep` runs for a `finishReason: "content-filter"` step:

- `args.steps` is empty — the content-filtered step is not recorded into the completed-steps array, and
- there is no `providerMetadata` anywhere on `ProcessOutputStepArgs`.

So observability/guardrail processors can detect _that_ a block happened (via `finishReason`) but **cannot attribute it** to the responsible policy — even though the provider already returns that data end-to-end. It is only lost at this boundary.

Additionally, when a processor calls `abort()` (a tripwire), the stream is torn down before any `finish`/`step-finish` chunk carrying `providerMetadata` reaches the agent stream consumer, so the route layer cannot recover it downstream either.

### Why it matters

Without the assessment, a guardrail block is an opaque "content safety" message. Operators can't distinguish a true positive from a false positive (e.g. a benign domain question tripping a denied topic), which is the whole point of running with `trace: "enabled"`.

### Expected

`processOutputStep` should be able to read the finishing step's `providerMetadata` (including on a `content-filter` block, where `steps` is empty), in both streaming and non-streaming generation. Behaviour should be unchanged for steps/providers that emit no provider metadata.

### Notes

PR with a proposed minimal fix (an optional `providerMetadata` field on `ProcessOutputStepArgs`) follows. Not Bedrock-specific — any provider that emits step `providerMetadata` benefits.

KiraPC:
Closes #18435

## What

Adds an optional `providerMetadata` field to `ProcessOutputStepArgs`, surfacing the finishing model step's provider-specific metadata to output-step processors.

## Why

When an AWS Bedrock Guardrail intervenes, the AI SDK Bedrock provider returns the full guardrail assessment in the provider stream's terminal `finish` part, under `providerMetadata.bedrock.trace.guardrail` (which policy/topic/filter blocked, the action, the confidence).

Today Mastra drops this at the Mastra → output-processor boundary. On a `finishReason: "content-filter"` step:

- `args.steps` is empty — the content-filtered step is not recorded into the completed-steps array, and
- there is no `providerMetadata` anywhere on `ProcessOutputStepArgs`.

As a result, observability/guardrail processors can detect _that_ a block happened (via `finishReason`) but **cannot attribute it** to the responsible policy — which is exactly the data the provider already returns end-to-end. It is only lost at this boundary. And because a processor `abort()` (tripwire) tears the stream down before any `finish`/`step-finish` chunk carrying `providerMetadata` reaches the agent stream consumer, the route layer can't recover it downstream either.

This matters for guardrail observability: without the assessment, a block is an opaque "content safety" message and operators can't tell a true positive from a false positive (e.g. a benign domain question tripping a denied topic).

## How

- `MastraModelOutput` captures the step `providerMetadata` from the terminal `finish` chunk — in the same handler that sets the finish reason, so it's available before any processor `abort()` runs — and exposes it via an internal `_getImmediateProviderMetadata()`.
- The LLM execution step passes it into `runProcessOutputStep`, which threads it through to `processOutputStep`.
- Works in both streaming and non-streaming generation. Behaviour is unchanged for steps without provider metadata (the field is simply `undefined`).

```ts
class MyProcessor implements Processor {
  async processOutputStep({ finishReason, providerMetadata }: ProcessOutputStepArgs) {
    if (finishReason === 'content-filter') {
      const guardrail = providerMetadata?.bedrock?.trace?.guardrail;
      // attribute the block to the responsible policy / topic / filter
    }
  }
}
```

## Tests

- `MastraModelOutput` surfaces the guardrail `providerMetadata` on a `content-filter` block where `steps` is empty, and leaves it `undefined` when the step produced none.
- `processOutputStep` receives `providerMetadata` on a `content-filter` block (and `undefined` when absent).

## Notes / open questions

We went with the minimal, additive Option A (a new optional field on `ProcessOutputStepArgs`). An alternative we considered was also pushing the content-filtered step into `args.steps` so existing consumers that read `steps[last].providerMetadata` work uniformly.

We're of course open to suggestions on the API shape — field name/placement, whether to also populate `args.steps`, or any concerns about the streaming capture point. Happy to adjust to match maintainers' preferences.

## ELI5
This change lets output processors see extra info from the model about why a response ended, even in cases where there isn’t any normal finished-step data. That means tools like guardrail and observability processors can still tell which provider policy or filter caused a block, instead of losing that info.

## Summary
- Added optional `providerMetadata` to `ProcessOutputStepArgs` and threaded it through output-step processing in both streaming and non-streaming paths.
- Captured provider metadata from the final model `finish` chunk and exposed it through `MastraModelOutput`.
- Updated processor runner/workflow plumbing so `processOutputStep` receives this metadata, including `content-filter` cases with empty `steps`.
- Added tests covering provider metadata propagation and the undefined fallback when no metadata exists.
- Documented the new field in the processor API reference.


intojhanurag (approved):
LGTM , Thanks ! Added a minor docs improvement .

PR #18446, opened 2026-06-24
feat(server,client-js): enrich harness session HTTP surface

abhiaiyer91:
## What

Enriches the harness session HTTP surface (and matching `@mastra/client-js` reads) so a client can render a status line, read agent behavior settings, and scope threads per git worktree — plus a fix for an SSE double-framing bug.

These additions build on the harness routes/SDK already on `main`; this PR is the server + client-js slice extracted from the MastraCode Web work so it can be reviewed on its own.

### Server (`@mastra/server`)

- `GET /harness/:harnessId/sessions/:resourceId` now also returns:
  - **`omProgress`** — status-line slice of observational-memory progress (pending tokens vs. observation threshold, accumulated observations vs. reflection threshold, projected message removal / reflection savings)
  - **`tokenUsage`** — cumulative token usage for the current thread
  - **`settings`** — `yolo`, `thinkingLevel`, `notifications`, `smartEditing`
- `GET .../threads` accepts an optional **`projectPath`** query param. A single `resourceId` can be shared across git worktrees of the same repo (the id is derived from the git URL), so `projectPath` scopes the list to threads tagged for that working directory (mirrors the TUI's worktree-strict selection). Each returned thread now also includes its `projectPath` metadata.
- `.../stream` now enqueues **raw event objects** and lets the server adapter handle SSE framing, fixing a double-framing bug where events were wrapped twice (`data: "data: {...}\n\n"\n\n`) and couldn't be parsed by clients.

### Client (`@mastra/client-js`)

- New `HarnessOMProgress` and `HarnessSessionSettings`, surfaced on `HarnessSessionState` (`omProgress`, `tokenUsage`, `settings`).
- New `display_state_changed` event in `KnownHarnessEvent` carrying the status-line figures.
- `HarnessSession.listThreads()` now accepts either a number (back-compat) or `{ limit?, projectPath? }`.

## Test plan

- `@mastra/server`: typecheck + build clean, `check:permissions` passes (no permission changes — these routes already existed), 18 harness handler tests pass (incl. new `projectPath` filter + raw-event-shape assertion).
- `@mastra/client-js`: build clean, 14 harness resource tests pass.
- `route-types.generated.ts` matches a fresh `generate:route-types` run (no drift).

## ELI5
This PR makes harness sessions show more useful “what’s happening” info (progress, token counts, and behavior settings), and it lets you fetch only the threads for the right workspace/worktree. It also fixes the event streaming so the client receives clean event objects instead of confusing “double-framed” SSE data.

## Summary
- **Server (`@mastra/server`)**
  - **`GET /harness/:harnessId/sessions/:resourceId`** now includes:
    - `omProgress` (status-line observational-memory progress figures)
    - `tokenUsage` (cumulative token usage)
    - `settings` (`yolo`, `thinkingLevel`, `notifications`, `smartEditing`)
  - **`GET .../threads`** now supports scoping with an optional **JSON-encoded `tags` query param** (e.g., `tags: { projectPath: ... }`), and each returned thread includes its `tags`.
  - **`.../stream`** now enqueues **raw event objects** for the server adapter to apply SSE framing, fixing a double-framing parsing bug.
  - **Session creation (`POST /harness/:harnessId/sessions`)** now accepts optional `tags` in the request body.
  - Prevents internal/reserved thread-metadata keys from leaking into user-facing tags and from being used in tag filtering by introducing a shared reserved-key predicate.

- **Client (`@mastra/client-js`)**
  - Adds types `HarnessOMProgress` and `HarnessSessionSettings`, exposed on `HarnessSessionState` as `omProgress`, `tokenUsage`, and `settings`.
  - Extends `KnownHarnessEvent` with `display_state_changed` (including status-line figures and optional telemetry).
  - Updates `HarnessSession.create()` to accept optional `{ tags?: Record<string, string> }`.
  - Updates `HarnessSession.listThreads()` to accept either a legacy `number` or `{ limit?: number; tags?: Record<string, string> }`, encoding `tags` as JSON in the query string.

- **Session/thread scoping via tags**
  - `harness.createSession` and thread resumption now use **match-all tag semantics** so sessions only resume/create the correct thread for the intended scope (e.g., worktrees sharing a `resourceId` don’t “cross-resume”).
  - New tests cover tag-based thread selection and stamping tags onto created threads.

- **Types/tests/generation**
  - Route metadata and generated route types were updated for the new `tags` request/query fields and the expanded session response shape.
  - Handler and client resource tests were updated for the new stream payload shape and tag-based thread filtering.


packages/server/src/server/handlers/harness.ts
```diff
@@ -110,6 +138,7 @@ const listThreadsResponseSchema = z.object({
       id: z.string(),
       title: z.string().optional(),
       updatedAt: z.string().optional(),
+      projectPath: z.string().optional(),
```
wardpeet: Should we consider adding `projectPath` as a recognized parameter within the harness? Currently, it functions as a custom state or is treated as a `resourceId`. It seems beneficial to use a combination of `ownerId` and `resourceId`. However, there's a challenge since `ownerId` is stored in the session, not the thread. Therefore, we should likely list sessions instead of threads.

wardpeet (approved):

PR #18448, opened 2026-06-24
feat(core): add Devin-style chat channels to Harness

abhiaiyer91:
Connects a Harness-backed bot to chat platforms like Slack. Inbound messages route into a Harness Session and the session's event stream renders back to the thread: streaming replies edit one message in place, tool calls render as cards that update to a result, and approvals post Approve/Deny buttons that resume the run on a click.

The thread-to-session mapping is automatic and Devin-style: a channel-root @mention starts a new session, replying in that thread continues the same session, and @mentioning elsewhere starts a fresh one. An opt-in `acknowledge` config adds a visible signal when a new session starts — a reaction on the triggering message and/or a "session started" message in the thread. It fires only on the first message of a session, is best-effort, and never blocks the user's message. Off by default, so nothing changes unless you configure it.

Channels are configured directly on the Harness, mirroring how agents do it. When the Harness is hosted on a Mastra server its webhook routes register automatically. A standalone `HarnessChannels` class is also exported for manual lifecycle management.

```typescript
const harness = new Harness({
  id: 'support',
  modes,
  channels: {
    adapters: { slack: createSlackAdapter({ botToken, signingSecret }) },
    acknowledge: { reaction: 'eyes', sessionStartMessage: '🧵 Started a new session.' },
  },
});
export const mastra = new Mastra({ harnesses: { support: harness } });
```

No changes to AgentChannels, the HarnessEvent/AgentChunkType types, or the existing session-binding model. Covered by colocated vitest tests (290 channel tests pass).

## ELI5
This PR lets a Harness bot “talk” inside chat apps like Slack. When someone mentions it, it can start or continue a conversation, show live replies and tool results, and let people approve or deny actions right in the thread.

## Summary
- Added `HarnessChannels` to connect Harness sessions to chat platforms and render session events back into threads.
- Implemented automatic thread-to-session mapping, plus optional first-message acknowledgements (reaction and/or starter message).
- Added chat event rendering for streaming assistant replies, tool cards, approvals, errors, and info messages.
- Wired channel support into `Harness`, including a new `channels` config option and `getChannels()` accessor.
- Registered harness channel webhook routes automatically in Mastra.
- Exported the new channel/renderer APIs from `packages/core/src/channels`.
- Added docs for the new channel system and updated Harness API references.
- Covered the new behavior with colocated Vitest tests.


CalebBarnes:
Nice to see Harness get first-class channels — the adapter config parsing, idempotent `initialize()`, and webhook `initPromise`/503 gating all match how `AgentChannels` does it, and it's clearly been studied carefully. Two architectural things I'd want resolved before this lands, both stemming from the fact that `AgentChannels` moved on after the model this PR is based on:

### 1. This reintroduces the subscribe-and-consume model that #17832 just deleted

The implementation's own doc table frames it as the analogue of `AgentChannels`: `agent.subscribeToThread()` → `session.subscribe(listener)`, with a `bindings = new Map<>()` cache and a long-lived `handleHarnessEvent` consumer per thread.

That's exactly the model #17832 (`86623c1adf`) **removed** from `AgentChannels`. On current `main`, `ensureThreadSubscription`/`consumeAgentStream`/`threadSubscriptions` are gone — rendering now happens *inline with the agent run* via `ChatChannelOutputProcessor` (stash a per-run render context on `requestContext`, then `consumeStream()` to drive the run). The whole point was serverless lifetime: a long-lived in-process subscriber dies when the webhook Lambda returns, so only the worker that wins the wake/signals race renders, and it renders inside the run that survives via `waitUntil`.

`HarnessChannels` keeps the in-memory `bindings` Map + `session.subscribe()` consumer, which has the lifetime problem we just designed out: when the webhook function returns on Vercel/Cloudflare, the binding and its subscription go away before the run finishes streaming. Has the serverless case been considered here, or is this intentionally scoped to long-lived processes for now? If the latter, worth saying so explicitly in the docs/changeset.

### 2. The renderer doesn't use the platform's native streaming API

`handleHarnessEvent` renders via `editOrPostMessage` — post a placeholder, then `adapter.editMessage(...)` repeatedly as `message_update` snapshots arrive. That's edit-in-place (e.g. a series of `chat.update` calls on Slack), not the platform's streaming primitive.

`AgentChannels` renders through `StreamingPlan` → `chatThread.post(plan)`, which the adapter drives through the platform's *native* streaming API (Slack `startStream`/`appendStream`/`stopStream`, etc.) with SDK-managed batching. So as written, Harness messages will look and behave differently (and worse) than `AgentChannels` messages on the same platform — no native streaming indicator, and it burns edit-rate-limit budget.

There's also a lossy round-trip baked in: the agent emits `text-delta`s → `session-run-engine` accumulates them into full `message_update` snapshots → the renderer diffs the snapshots back into deltas (`fullText.length - lastTextLength`) to feed `editOrPostMessage`. That's re-deriving the increments `StreamingPlan` consumes directly. A few ways to converge:
- **Cheapest:** have the renderer feed re-derived text deltas into a `StreamingPlan` instead of `editMessage`, so at least text gets native streaming.
- **Better:** factor the shared render machine out of `runStreamingDriver` so both `AgentChannels` and Harness emit through one path (text + tool cards + `data-*`/OM), instead of a second hand-rolled renderer.
- **Fallback:** a `HarnessEvent → AgentChunkType` adapter so `runStreamingDriver` can be reused wholesale (keeps the round-trip, but no duplicate renderer).

### Smaller stuff
- Changeset uses `createSlackAdapter(...)` but the docs use `new SlackAdapter(...)` — pick one.
- A few `as any` casts (`(sdkInstance as any).webhooks?.[platform]`, etc.) worth typing if the SDK surface allows.
- `mastra/index.ts` swallows harness-channel init failures with `console.error` — `AgentChannels` uses the logger; worth matching.
- CI is red (Combined Store Tests/pg, Lint docs, Mastra Code E2E, Superagent Security Scan) — some may be flakes, but Lint docs is probably the changeset/docs mismatch above.

PR #18458, opened 2026-06-25
chore(playground): debarrel cn imports

damien-schneider:
## Summary
- Add a narrow `@mastra/playground-ui/utils/cn` public utility subpath and build/export wiring.
- Move existing `cn` imports in `packages/playground` from the root `@mastra/playground-ui` barrel to `@mastra/playground-ui/utils/cn`, preserving unrelated mixed root imports.
- Add the playground ESLint ratchet for `cn` and a patch changeset for `@mastra/playground-ui`.

## Why
This is mechanical import hygiene to reduce root barrel use and avoid accidental performance regressions.

## Validation
- `pnpm build:playground-ui`
- `pnpm --filter ./packages/playground-ui typecheck`
- `pnpm --filter ./packages/playground typecheck`
- `pnpm --filter ./packages/playground lint` (passes with 41 existing warnings)
- `pnpm prettier:changed` (passed; reverted out-of-scope example formatter spillover)
- `pnpm lint` (passes with existing warning baseline)
- Parser audit: zero `cn` imports remain from `@mastra/playground-ui` in `packages/playground/src`
- `npx -y react-doctor@latest . --verbose --diff` (exit 0; existing warnings only)
- `coderabbit doctor` (CLI is not signed in, so local review was not run)

## ELI5
This PR lets you import the `cn` helper from a tiny, specific path (like `@mastra/playground-ui/utils/cn`) instead of grabbing it from the whole “playground UI” bundle. It keeps behavior the same, but makes the import targets more precise and prevents accidentally going back to the old path.

## Summary
- Added a public `@mastra/playground-ui/utils/cn` subpath export in `packages/playground-ui` and wired it into the build/export output.
- Updated `packages/playground-ui/src/utils/cn.ts` to re-export `cn` from the shared utils implementation.
- Enhanced `packages/playground-ui/vite.config.ts` to generate `src/utils`-based utility entrypoints so the new subpath is actually built and shipped.
- Switched `cn` imports throughout `packages/playground` from the root `@mastra/playground-ui` barrel to `@mastra/playground-ui/utils/cn`, while preserving unrelated root imports (e.g., `Icon`, other components).
- Added an ESLint restriction in `packages/playground/eslint.config.js` to ratchet against reintroducing `cn` imports from the root playground UI barrel.
- Updated the `@mastra/playground-ui` patch changeset to document the developer-facing import path: `import { cn } from '`@mastra/playground-ui/utils/cn`';`.

### DIFF
--- a/.changeset/gold-buttons-shop.md
+++ b/.changeset/gold-buttons-shop.md
@@ -2,4 +2,8 @@
 '@mastra/playground-ui': patch
 ---
 
-Added a direct utility import for cn.
+Added a direct utility import for the class name helper so applications can avoid the root Playground UI barrel.
+
+```ts
+import { cn } from '@mastra/playground-ui/utils/cn';
+```
--- a/.changeset/gold-buttons-shop.md
+++ b/.changeset/gold-buttons-shop.md
@@ -1,0 +1,5 @@
+---
+'@mastra/playground-ui': patch
+---
+
+Added a direct utility import for cn.
--- a/packages/playground-ui/package.json
+++ b/packages/playground-ui/package.json
@@ -43,6 +43,16 @@
         "default": "./dist/utils.cjs.js"
       }
     },
+    "./utils/*": {
+      "import": {
+        "types": "./dist/utils/*.d.ts",
+        "default": "./dist/utils/*.es.js"
+      },
+      "require": {
+        "types": "./dist/utils/*.d.ts",
+        "default": "./dist/utils/*.cjs.js"
+      }
+    },
     "./hooks/*": {
       "import": {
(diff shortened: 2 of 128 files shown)


mfrachet (approved):

damien-schneider:
Handled the CodeRabbit review follow-up.

- Updated the changeset in fff24da7a8 to describe the developer-facing outcome and include the public API example:

  `import { cn } from '@mastra/playground-ui/utils/cn';`

- Confirmed the package export is already present in `packages/playground-ui/package.json` as `./utils/*`, using the same nested `import` / `require` condition style as the existing `./utils` and `./tokens` exports. No code change was needed for that thread.

PR #18467, opened 2026-06-25
feat(core/harness): move workspace and browser ownership to sessions

wardpeet:
Moves workspace and browser from Harness-level shared resources to per-session ownership.

`createSession` now accepts optional `workspace` and `browser` overrides that are passed directly to the `Session` constructor. These overrides are static instances only — use the Harness-level `DynamicArgument` config for per-request factory functions. When no override is provided, the Harness-level config is used as a fallback. The `Session` constructor validates that `workspace` is a `Workspace` instance and throws if missing.

This is a breaking change — `HarnessConfig.workspace` no longer accepts a `WorkspaceConfig` shorthand, and the following `Harness` methods are removed: `getWorkspace`, `resolveWorkspace`, `hasWorkspace`, `isWorkspaceReady`, `destroyWorkspace`.

**Before:**
```ts
const harness = new Harness({
  name: 'my-harness',
  workspace: { name: 'my-workspace' }, // WorkspaceConfig shorthand
});
const session = await harness.createSession({ resourceId: 'user-1' });
const ws = harness.getWorkspace();
```

**After:**
```ts
const harness = new Harness({
  name: 'my-harness',
  workspace: new Workspace({ name: 'my-workspace' }),
});

// static override per session
const session = await harness.createSession({
  resourceId: 'user-1',
  workspace: new Workspace({ name: 'ws-1' }),
});

// dynamic per-request factory stays at the harness config level
const harness2 = new Harness({
  name: 'my-harness',
  workspace: ({ requestContext }) => new Workspace({ name: `ws-${requestContext.session.id}` }),
});
```

All 36 harness test files (359 tests) pass and `tsc --noEmit` is clean.

## ELI5
Before, the “Harness” picked and owned the `workspace`/`browser` that all sessions shared. Now, each `Session` is responsible for its own `workspace` (and optional `browser`), so setups don’t accidentally leak across sessions.

## Summary
- Shifted workspace/browser ownership from Harness-level shared state to per-session ownership.
- `Harness.createSession` now accepts optional per-call overrides:
  - `workspace?: Workspace` (static instance only)
  - `browser?: MastraBrowser`
  - `requestContext?: RequestContext`
  These are applied during session creation and passed into the `Session` constructor.
- Dynamic/per-request workspace factories are still supported, but only via the Harness-level `HarnessConfig.workspace` `DynamicArgument`; per-session overrides are static instances.
- `Session` now requires a valid `workspace: Workspace` at construction time and throws if it’s missing/invalid.
- Workspace lifecycle events are handled per session:
  - During session creation, if the resolved `workspace` is a `Workspace` instance, the harness calls `workspace.init()` and emits workspace lifecycle events on the session.
  - `SessionBus` buffers the latest workspace lifecycle event group and replays it to subscribers that attach after session creation.

## Breaking changes / API updates
- `HarnessConfig.workspace` no longer accepts a `WorkspaceConfig` shorthand; it is now limited to the `DynamicArgument<Workspace | undefined>` form.
- `Harness.destroyWorkspace()` was removed; `Harness.destroy()` no longer tears down the workspace.
- Workspace API behavior changes:
  - `getWorkspace()` returns only the Harness-level static workspace (dynamic factory setups don’t resolve via this method).
  - `resolveWorkspace()` now requires a `session` parameter and resolves dynamic workspace factories against that session’s request context (and is used to support code paths outside the normal request flow).

## Docs and tests
- Updated docs and reference pages to reflect the new session-level `workspace`/`browser` resolution model and the changed workspace APIs.
- Updated/expanded tests across Harness, Session, and server/SSE coverage to construct Harnesses with explicit mock `Workspace` instances where needed.
- `tsc --noEmit` passes, and the harness test suite runs cleanly.

## Changeset
- Added a breaking changeset documenting the workspace/browser ownership move from `Harness` to `Session`, including the revised `createSession` override behavior and workspace lifecycle expectations.

### DIFF
--- a/.changeset/strict-feet-tan.md
+++ b/.changeset/strict-feet-tan.md
@@ -1,10 +1,8 @@
 ---
-'@mastra/core': major
+'@mastra/core': minor
 ---
 
-Moved workspace and browser ownership from the Harness to individual sessions.
-
-`createSession` now accepts optional `workspace` and `browser` overrides that are passed directly to the `Session` constructor. These overrides are static instances only — use the Harness-level `DynamicArgument` config for per-request factory functions. When no override is provided, the Harness-level config is used as a fallback. Workspace is now required at session construction time — the `Session` constructor validates it is a `Workspace` instance and throws if missing.
+Moved workspace and browser ownership from the Harness to individual sessions. `createSession` now accepts optional `workspace` and `browser` overrides that are passed directly to the `Session` constructor. When no override is provided, the Harness-level config is used as a fallback.
 
 **Breaking changes:**
 
--- a/.changeset/strict-feet-tan.md
+++ b/.changeset/strict-feet-tan.md
@@ -11,7 +11,8 @@ Moved workspace and browser ownership from the Harness to individual sessions.
 - `Session` constructor now requires `workspace: Workspace` and accepts optional `browser?: MastraBrowser`
 - `HarnessConfig.workspace` no longer accepts a `WorkspaceConfig` object — pass a `Workspace` instance or a dynamic factory function
 - `createSession` overrides accept static `Workspace` / `MastraBrowser` instances only (not `DynamicArgument`)
-- Removed `Harness.getWorkspace()`, `Harness.resolveWorkspace()`, `Harness.hasWorkspace()`, `Harness.isWorkspaceReady()`, and `Harness.destroyWorkspace()` — workspace is now accessed via the session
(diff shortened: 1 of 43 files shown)


abhiaiyer91 (approved):

.changeset/strict-feet-tan.md
```diff
@@ -0,0 +1,58 @@
+---
+'@mastra/core': major
```
wardpeet:
```suggestion
'@mastra/core': minor
```

PR #18473, opened 2026-06-25
fix(ai-sdk,core): deliver background-task-started chunk to the UI stream and onChunk

### LINKED ISSUE #18464: background-task-started lifecycle chunk is dropped before reaching the UI stream in handleChatStream
# `background-task-started` lifecycle chunk is dropped before reaching the UI stream in `handleChatStream`

## Summary

First off — thank you for the background tasks feature, it's a genuinely great
fit for "fire and keep chatting" UX. We're running into one issue that currently
makes it hard to consume the task lifecycle from a web-chat UI.

When an agent dispatches a tool as a background task, Mastra emits a structured
`background-task-started` chunk on the agent stream carrying the real `taskId`
(`{ taskId, toolName, toolCallId }`). In a standard web-chat setup that routes
through `@mastra/ai-sdk`'s `handleChatStream()` → `agent.stream()` →
`toAISdkStream()`, this chunk appears to be **dropped on two independent paths**
and never reaches the UI. As a result consumers cannot obtain the `taskId` from
the documented lifecycle event, and are forced to regex-parse it out of the
tool-result string instead.

This is meant as a constructive report — happy to refine it or provide more
repro detail if that helps.

## Why we need `taskId` (real-world impact)

Background tasks are explicitly designed for the "fire and keep chatting" UX:
the agent dispatches a long-running tool, returns an immediate acknowledgement,
and the user keeps talking in the same thread while the task runs. To make that
UX actually work, the frontend needs the `taskId` at dispatch time for several
concrete things — all of which are currently blocked without the chunk:

1. **Render a live progress card at dispatch time.** We show a "task running"
   card the moment the task is dispatched so the user sees something is
   happening and can keep chatting. Without `taskId`, the card cannot be created
   until the task *finishes* — which for a multi-minute research task defeats the
   entire point of running it in the background.

2. **Subscribe to the task's lifecycle out-of-band.** Because the agent's stream
   returns immediately after dispatch (it does not block on the task), the only
   way for the UI to observe `running → output → completed/failed` is to open a
   separate subscription scoped to the task — e.g. `backgroundTaskManager.stream({ taskId })`
   / `getTask(taskId)` over a dedicated SSE endpoint. That subscription is keyed
   on `taskId`; no `taskId`, no way to follow the task.

3. **Show streamed output / the final report.** Our deep-research task streams
   its output incrementally and writes a report on completion. Both arrive via
   the task-scoped manager stream as `background-task-output` /
   `background-task-completed`, which the UI can only receive once it has the
   `taskId`.

4. **Render a terminal state instead of an infinite spinner.** When a background
   task fails, the failure is reported as `background-task-failed` on the
   manager stream. Without a `taskId`-driven subscription, the UI never learns
   the task died and the card spins forever.

In short: **`taskId` is the handle that makes a background task observable from
the UI at all.** The whole value proposition of background tasks (non-blocking,
keep-chatting, live progress, eventual result) depends on the frontend getting
that handle at dispatch time — which is exactly what `background-task-started`
was meant to provide.

## Versions

- `@mastra/core`: `1.42.0`
- `@mastra/ai-sdk`: `1.4.5`
- `ai`: `6.0.188`

### 1. Mastra does emit the chunk

In the agent's tool-execution loop, when a background task is dispatched
(`fallbackToSync === false`), the `background-task-started` chunk is enqueued
onto the stream controller with a structured payload, and the tool returns the
human-readable acknowledgement as its result:

```js
const { task, fallbackToSync } = await bgTask.dispatch();
if (!fallbackToSync) {
  safeEnqueue(controller, {
    type: "background-task-started",
    runId,
    from: "AGENT",
    payload: {
      taskId: task.id,
      toolName: inputData.toolName,
      toolCallId: inputData.toolCallId,
    },
  });
  return {
    result: `Background task started. Task ID: ${task.id}. The tool "${inputData.toolName}" is running in the background. You will be notified when it completes.`,
    ...inputData,
  };
}
```

So the chunk exists, carries exactly the structured `{ taskId, toolName, toolCallId }`
a UI needs, and is placed onto the agent stream controller.

### 2. Path A — `onChunk` is not invoked for it

In the agent loop, the `onChunk` option is only invoked for a fixed allowlist of
chunk types:

```js
default:
  safeEnqueue(controller, chunk);   // ← chunk IS enqueued to the stream...
}
if ([
  "text-delta",
  "reasoning-delta",
  "source",
  "tool-call",
  "tool-call-input-streaming-start",
  "tool-call-delta",
  "tool-call-input-streaming-end",
  "raw",
].includes(chunk.type)) {
  if (chunk.type === "raw" && !includeRawChunks) {
    continue;
  }
  await options?.onChunk?.(chunk);   // ← ...but onChunk is NOT called for background-task-started
}
```

So a consumer passing `onChunk` to `handleChatStream` / `agent.stream()`
expecting to observe `background-task-started` never receives it.

### 3. Path B — `toAISdkStream` transformer discards it

`@mastra/ai-sdk`'s `convertFullStreamChunkToUIMessageStream` returns `null` for
any chunk type it doesn't explicitly handle and that isn't `data-` prefixed:

```js
default: {
  if (isDataChunkType(payload)) {
    // ... pass through data-* chunks
  }
  return null;   // ← background-task-started falls through here and is dropped
}
```

There's no handling of any `background-task-*` chunk type anywhere in
`@mastra/ai-sdk` (a grep for `background-task` returns no matches in the
package), so the chunk — even though present on the agent `fullStream` — is
filtered out before reaching the AI SDK UI message stream.

## Current (fragile) workaround

Because the chunk is unreachable, we're currently extracting the `taskId` by
regex-matching the tool-result **string**:

```ts
// tool-result is a plain string, set by Mastra at dispatch time:
// "Background task started. Task ID: <id>. The tool \"...\" is running in the background."
const match = output.match(/Background task started\.\s+Task ID:\s+([^\s.]+)/i);
const taskId = match?.[1]?.trim();
```

This works today but is brittle — it depends on a natural-language string and
its exact wording, spacing, and punctuation, and it throws away the structured
payload the chunk already carries. It would be great to not have to rely on it.

## Expected behavior

One (or both) of these would resolve it cleanly:

1. **`@mastra/core`**: include `background-task-started` (and ideally the other
   agent-stream background-task chunks) in the `onChunk` allowlist so the
   callback fires for them.
2. **`@mastra/ai-sdk`**: have `toAISdkStream` / `convertFullStreamChunkToUIMessageStream`
   pass `background-task-*` chunks through (e.g. as `data-` data parts) so they
   survive into the UI message stream the AI SDK UI consumes.

Either change would let a web-chat consumer obtain the structured
`{ taskId, toolName, toolCallId }` from the lifecycle event directly, instead of
regex-parsing the tool-result string.

## Reproduction

```ts
import { handleChatStream } from "@mastra/ai-sdk";

// 1. Configure an agent with backgroundTasks.tools enabled for some tool,
//    and backgroundTasks.enabled on the Mastra instance.
// 2. Route a web chat through handleChatStream (version: "v6").
// 3. Trigger a prompt that makes the model call the background-eligible tool.

const stream = await handleChatStream({
  mastra,
  agentId,
  version: "v6",
  params: { messages, memory: { thread: tId, resource: rId } },
  defaultOptions: {
    onChunk: (chunk) => {
      console.log("chunk:", chunk.type);
      // You'll see tool-call, tool-result, text-delta, reasoning-delta...
      // ...but never "background-task-started", even though the task was
      // dispatched and is visible via backgroundTaskManager.getTask(taskId).
    },
  },
});
```

Querying `mastra_background_tasks` (or `backgroundTaskManager.getTask`) confirms
the task was created and started, so dispatch definitely happened — the chunk
simply never propagates to the UI layer.

## Additional context

- The manager API `backgroundTaskManager.stream()` does emit `background-task-running`
  / `-completed` / etc., which we already use for the post-dispatch lifecycle.
  But those arrive on a long-lived stream that stays open until the caller's
  `AbortSignal` fires, and `running` fires when the task starts executing (not
  when it's enqueued). Wiring that into a request-scoped web-chat stream
  introduces timing/ordering complexity that `background-task-started` on the
  agent stream was presumably meant to avoid.
- The embedded background-tasks doc (`docs-agents-background-tasks.md`) lists
  `background-task-started` as an agent-stream chunk type with payload
  `{ taskId, toolName, toolCallId }`, which reads as though it's meant to be
  consumable from the agent's own stream.
- Related (different root cause, same area): #18451 (tool-level `background`
  config silently ignored) and #17798 (background sub-agent delegations send
  null tool-message content).

Thanks again for the great work on Mastra, and please let me know if there's
anything I can clarify or test on our side. 🙏

Sandy-1711:
## Description

Fixes #18464.

When an agent dispatches a tool as a background task, it emits a `background-task-started` chunk that carries the `taskId` (plus `toolName` and `toolCallId`). That `taskId` is the handle a UI needs to show a progress card and subscribe to the task while the user keeps chatting. The problem: the chunk never made it out to consumers, so people were left scraping the `taskId` out of the tool-result string with a regex.

Digging in, it turned out the chunk was getting dropped in two unrelated places. Both needed fixing.

### 1. The AI SDK UI stream silently dropped it (`@mastra/ai-sdk`)

This is the one that actually matters for web-chat UIs. `convertFullStreamChunkToUIMessageStream` only knows how to translate a fixed set of chunk types into UI parts; anything else has to be `data-` prefixed or it returns `undefined` and the chunk disappears. `background-task-started` is neither, so `handleChatStream` / `useChat` never saw it.

The fix is to forward any `background-task-*` chunk as a `data-background-task-*` part so it survives with its payload intact:

```ts
if (typeof partType === 'string' && partType.startsWith('background-task-')) {
  const { type, ...data } = part;
  const toolCallId = data.toolCallId;
  return {
    type: `data-${type}`,
    ...(typeof toolCallId === 'string' ? { id: toolCallId } : {}),
    data,
  };
}
```

I matched on the whole `background-task-*` family rather than just `-started`, so the other lifecycle chunks survive too if they reach this stream, and I set the part `id` to `toolCallId` so a UI can reconcile updates for the same task.

### 2. `onChunk` never fired for it (`@mastra/core`)

The issue (and an earlier investigation comment) suggested adding `background-task-started` to the `onChunk` allowlist in `llm-execution-step.ts`. I tried that first and it does nothing — and it's worth explaining why, because it's a bit subtle.

That allowlist lives inside the loop that reads the *model's* output stream (`outputStream._getBaseStream()`). But `background-task-started` isn't produced by the model. It's emitted by `tool-call-step`, which runs later in the workflow (`.then(llmExecution).foreach(toolCall)`) and writes straight to the stream controller. By the time it's emitted, the model loop has already finished, so it never sees the chunk — adding it to the allowlist gives the check something to match that never arrives.

So the fix has to live where the chunk is actually emitted. `options.onChunk` is already in scope there, so it's a one-line addition next to the existing enqueue:

```ts
safeEnqueue(controller, backgroundTaskStartedChunk);   // already there
await options?.onChunk?.(backgroundTaskStartedChunk);  // new
```

One caveat I'd flag for reviewers: in `handleChatStream`, `onChunk` runs on the server and doesn't cross the wire by itself, so fix #1 is the one a browser actually depends on. Fix #2 is for server-side consumers (logging, forwarding, etc.). Both are real, so I did both.

### Verifying it

I confirmed the bug against `main` rather than just trusting the new tests — reverted the two source files to their pre-fix state and ran the same checks:

- **Path B:** on `main` the UI part comes out `undefined` (dropped); with the fix it comes out as `data-background-task-started` carrying `{ taskId, toolName, toolCallId }`.
- **Path A:** on `main`, `onChunk` sees `tool-call`, `tool-result`, `text-delta`… but not `background-task-started`; with the fix it shows up. Checked on both the `normal` and `evented` engines. (This also confirms the allowlist approach wouldn't have worked — `tool-call` shows up because the model emits it through that loop, but the background-task chunk doesn't.)

## Related issue(s)

Fixes #18464

## Type of change

## Checklist

### Tests added

- `client-sdks/ai-sdk/src/helpers.test.ts` — `background-task-*` chunks map to `data-*` parts with the structured payload.
- `background-task-agent-level.scenario.test.ts` — an agent-level scenario asserting `onChunk` receives `background-task-started`.

## ELI5
When an agent starts a long-running “background task,” the app should immediately hear about it so it can show progress. This PR makes sure the `background-task-started` event isn’t dropped: it reaches both the UI stream and any `onChunk` listeners, with the real task details already structured.

## Summary
- In `@mastra/ai-sdk`, `convertFullStreamChunkToUIMessageStream` now forwards all `background-task-*` lifecycle chunks into the UI stream as typed parts named `data-background-task-*`, preserving the structured payload and using `toolCallId` as the part `id` when available.
- In `@mastra/core`, `background-task-started` now invokes `options.onChunk` at the moment the chunk is emitted (instead of depending on a path/allowlist that never receives it), so agent-level consumers can observe `{ taskId, toolName, toolCallId }` immediately.
- Adds tests covering:
  - UI-stream passthrough/conversion of `background-task-*` chunks into `data-background-task-*` parts with the expected payload.
  - Agent-level `onChunk` delivery of `background-task-started` when background execution is enabled.
- Adds changesets documenting the fixes for `@mastra/ai-sdk` and `@mastra/core`.


graysonhicks:
Closing in favor of https://github.com/mastra-ai/mastra/pull/18628

PR #18479, opened 2026-06-25
fix(core): correct replay ordering on pull-mode pubsub transports

### LINKED ISSUE #18409: [BUG] CachingPubSub.subscribeFromOffset / subscribeWithReplay have correctness issues on persistent pubsub transports (e.g. Redis Streams)
## What happens

When a subscriber attaches to an in-progress agent run via `subscribeWithReplay` / `subscribeFromOffset` — for example, reconnecting an SSE consumer after a network blip, or attaching with `agent.observe(runId)` — events can be delivered:

- **Out of order**, with live re-replay arriving before the cached history catches up
- **Duplicated**, with the same chunk reaching the consumer twice
- **Without honoring the offset**, with events strictly before the requested index still reaching the consumer

The symptoms are most visible on transports that re-replay the entire stream on a fresh subscribe (Redis Streams, where a fresh consumer group starts at id `0` and re-reads everything in the stream). On the in-memory `EventEmitterPubSub` used by the framework's own tests, none of these are reproducible — which is part of why they've stayed hidden.

## The issues

All three live in `packages/core/src/events/caching-pubsub.ts`. The two replay-aware subscribe methods — `subscribeWithReplay` and `subscribeFromOffset` — currently carry near-identical logic in two separate implementations, and both exhibit issues 1 and 2 independently. Issue 3 applies only to `subscribeFromOffset` (since `subscribeWithReplay` has no offset parameter).

**1. Out-of-order delivery on pull-mode transports**

The current code subscribes to the live transport first, then `await`s `getHistory`, then iterates the cached events. On pull-mode transports (Redis Streams), the live consumer starts re-reading the stream from its beginning as soon as `inner.subscribe` resolves. During the `await getHistory()` window, the transport hands re-replayed events to the wrapped callback. The wrapped callback delivers them to the consumer immediately — *before* the synchronous history loop runs.

The result is that a consumer expecting "history first, live second" can see events arrive in stream order from the live path, interleaved with or preceding the cached history. Streaming UI consumers see a later chunk render before an earlier one.

A fix here needs to defer live delivery during the bootstrap window: subscribe live, buffer everything that arrives, then deliver history in order, then drain the buffer (suppressing duplicates that history already covered), then flip to passthrough for steady-state.

**2. Dedup key is not stable across transports**

`CachingPubSub.publish` assigns the event's `id` and caches the event with it. The inner pubsub (`EventEmitterPubSub`, `RedisStreamsPubSub`, …) then regenerates `id` inside its own `publish`. So the cached copy and the live copy of the same event carry different ids, and the current `seen.has(event.id)` dedup never matches a duplicate.

On in-memory transports this is mostly invisible because the live path typically only delivers events *after* the subscribe — there are no duplicates to dedup. On pull-mode transports, the consumer group re-replays everything, so every history event also arrives via live, and they all bypass dedup.

The framework already assigns a sequential `event.index` in `CachingPubSub.publish`, and inner transports preserve it. Dedup keyed on `event.index` (with a fallback to `event.id` for events that have no index — these are never cached and so can't be replay/live duplicated) is stable across both paths.

This is the same root cause as #18191.

**3. `offset` parameter is only honored on the history path**

`getHistory(topic, offset)` correctly returns only events with index ≥ offset. But the wrapped callback that handles live deliveries never checks `event.index < offset`. On pull-mode transports that re-replay from the start, events strictly before the requested offset still reach the consumer even though they explicitly asked to skip them.

A one-line guard at the top of the wrapped callback — `if (typeof event.index === 'number' && event.index < offset) return` — closes this.

## Reproductions

- For (1): subscribe with `subscribeWithReplay` or `subscribeFromOffset(topic, 0, cb)` to a `RedisStreamsPubSub`-backed topic that has cached events. The chunks arrive in transport order interleaved with history.
- For (2): #18148 has a clean reproduction (duplicate buffered prefix when observing an in-progress run).
- For (3): `subscribeFromOffset(topic, 50, cb)` on a `RedisStreamsPubSub` with a populated cache; events 0–49 still reach `cb`.

## Suggested fix direction

Rewrite `subscribeFromOffset` with the algorithm below, and make `subscribeWithReplay` a thin delegate (`return this.subscribeFromOffset(topic, 0, cb)`) — collapsing two near-duplicate implementations into one. The unified function does:

1. Subscribe live first; while bootstrapping, push every delivery into a local buffer instead of forwarding to the consumer callback.
2. Fetch history and deliver it in order; record each event's stable dedup key in a `seen` set.
3. Drain the buffer; entries whose key is already in `seen` are skipped (they're the same events arriving via live re-replay).
4. After the drain, replace the `seen` set with a `lastDelivered` index check — bounded memory in the steady state, suppresses any later re-deliveries from the transport.
5. Filter `event.index < offset` inside the wrapped callback so the live path honors the offset.

## Related

- #18148 — original bug report covering the duplicate-events case
- #18191 — fix that addresses the dedup key (issue 2) plus a separate double-wrapping bug; doesn't address issues 1 or 3

## Environment

- Reproduced on `@mastra/core` 1.45 with `RedisStreamsPubSub` as the inner transport
- Not reproducible against the in-memory transport used in the framework's test suite

---

Happy to open a PR with the unified `subscribeFromOffset` if you agree with the issue and the suggested direction.

intojhanurag:
## Description

Fixes three correctness bugs in `CachingPubSub.subscribeFromOffset` / `subscribeWithReplay` that caused out-of-order delivery, duplicate events, and offset violations when backed by pull-mode transports like Redis Streams. These methods are used when a late consumer attaches to an in-progress agent stream (e.g. `agent.observe(runId)` or SSE reconnect).

- Buffer live deliveries during bootstrap instead of forwarding immediately, so history is always delivered before live events regardless of transport behavior
- Add `event.index < offset` guard in the live callback so events before the requested offset are never delivered to the consumer
- Suppressed duplicate events are silently dropped without acking (transport-level cleanup is the adapter's responsibility), which is safe when multiple subscribers share one ack handle (e.g. Google Cloud Pub/Sub)
- Preserve `nack` handles through the buffer so consumers can trigger redelivery on processing failures
- Collapse `subscribeWithReplay` into a one-line delegate to `subscribeFromOffset(topic, 0, cb)`, eliminating near-duplicate implementations
- Use a bounded `lastDelivered` index watermark for steady-state dedup instead of nulling out the `seen` Set

## Related Issue(s)

Fixes #18409

## Type of Change

- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [ ] Test update

## Checklist

## ELI5
When you start listening to a stream in the middle (or reconnect), the pubsub layer can accidentally send old “past” events too late or send the same event twice. This PR fixes the subscription flow so it first delivers the cached history, then safely forwards any live events that weren’t already replayed.

## Summary
- Fixed replay/bootstrap correctness for pull-mode pubsub transports (e.g., Redis Streams) in `CachingPubSub.subscribeFromOffset`.
- Replaced `subscribeWithReplay(topic, cb)` with a delegate to `subscribeFromOffset(topic, 0, cb)`.
- During subscription bootstrap:
  - Subscribes to live events immediately but buffers them until cached history is delivered.
  - Drops any live event whose `event.index < offset` (so requested offsets are honored on the live path too).
  - Delivers cached history first, then drains buffered live events while suppressing duplicates seen in history.
  - Suppressed duplicates are dropped (they are not acked via the buffered ack/nack handles).
- In steady state:
  - Uses a bounded `lastDelivered` watermark for dedup instead of an unbounded “seen set”.
  - Skips events with `event.index <= lastDelivered`, except when `event.deliveryAttempt > 1` (so `nack` redeliveries can be retried).
  - Ensures the watermark only advances on higher indices, so retry delivery doesn’t regress dedup state.
- Preserved consumer `ack`/`nack` semantics for buffered live events that are delivered after history replay (including correct ack/nack handle passing and retry behavior).
- Added rollback on bootstrap failure: if `getHistory` throws, the wrapper callback is unsubscribed and removed from `callbackMap` so subsequent live events won’t reach the consumer.
- Expanded tests, including a pull-mode `PullModePubSub` mock to verify ordering (history-before-live), no duplicates at the replay/live boundary, offset filtering, bounded watermark dedup behavior, `nack` retry delivery, ack/nack handle preservation for buffered live events, and cleanup when replay bootstrap fails.
- Added a changeset documenting the patch-level replay ordering fix for `@mastra/core`.

### DIFF
--- a/.changeset/calm-bags-agree.md
+++ b/.changeset/calm-bags-agree.md
@@ -1,0 +1,5 @@
+---
+'@mastra/core': patch
+---
+
+Fixed replay ordering on pull-mode transports (e.g. Redis Streams): history events are now delivered before live events, offsets are enforced on the live path, suppressed duplicates are acknowledged so persistent transports stop redelivering them, and ack/nack handles are preserved for buffered events.
--- a/packages/core/src/events/caching-pubsub.test.ts
+++ b/packages/core/src/events/caching-pubsub.test.ts
@@ -792,7 +792,6 @@ describe('CachingPubSub', () => {
 
       // Only events with index >= 3 should be delivered
       expect(received).toEqual([3, 4]);
-      expect(pullInner.getAckedIndices()).toEqual([0, 1, 2, 3, 4]);
     });
 
     it('delivers events published during getHistory bootstrap without duplication', async () => {
@@ -843,23 +842,6 @@ describe('CachingPubSub', () => {
       await pullCaching.publish(topic, { type: 'e2', runId: 'r', data: {} });
 
       expect(received).toEqual([0, 1, 2]);
-    });
-
-    it('acks suppressed duplicate live deliveries so persistent transports do not redeliver them', async () => {
-      const pullInner = new PullModePubSub();
-      const pullCaching = new CachingPubSub(pullInner, cache);
-      const topic = 'pull-ack-suppressed';
-
-      await pullCaching.publish(topic, { type: 'e0', runId: 'r', data: {} });
-      await pullCaching.publish(topic, { type: 'e1', runId: 'r', data: {} });
-
-      const received: number[] = [];
-      await pullCaching.subscribeWithReplay(topic, (event: Event) => {
-        received.push(event.index!);
-      });
-
-      expect(received).toEqual([0, 1]);
-      expect(pullInner.getAckedIndices()).toEqual([0, 1]);
     });
 
     it('allows nack-redelivered events through even when index <= lastDelivered', async () => {
--- a/packages/core/src/events/caching-pubsub.test.ts
+++ b/packages/core/src/events/caching-pubsub.test.ts
@@ -862,6 +862,64 @@ describe('CachingPubSub', () => {
       expect(pullInner.getAckedIndices()).toEqual([0, 1]);
     });
 
+    it('allows nack-redelivered events through even when index <= lastDelivered', async () => {
+      const pullInner = new PullModePubSub();
+      const pullCaching = new CachingPubSub(pullInner, cache);
+      const topic = 'pull-nack-retry';
+
+      await pullCaching.publish(topic, { type: 'e0', runId: 'r', data: {} });
+      await pullCaching.publish(topic, { type: 'e1', runId: 'r', data: {} });
+
+      const received: Array<{ index: number; attempt: number | undefined }> = [];
+      await pullCaching.subscribeWithReplay(topic, (event: Event) => {
+        received.push({ index: event.index!, attempt: event.deliveryAttempt });
+      });
+
+      expect(received).toEqual([
+        { index: 0, attempt: undefined },
+        { index: 1, attempt: undefined },
+      ]);
+
+      // Simulate a nack redelivery: same index, deliveryAttempt > 1
+      pullInner.emitLiveOnly(topic, {
+        id: 'retry-id',
+        type: 'e1',
+        runId: 'r',
+        data: {},
+        createdAt: new Date(),
+        index: 1,
+        deliveryAttempt: 2,
+      });
+
+      expect(received).toHaveLength(3);
+      expect(received[2]).toEqual({ index: 1, attempt: 2 });
+    });
+
+    it('cleans up wrappedCb when replay bootstrap fails', async () => {
+      const pullInner = new PullModePubSub();
+      const pullCaching = new CachingPubSub(pullInner, cache);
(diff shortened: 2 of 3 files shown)


ChristianOlsson:
Thanks for creating a PR @intojhanurag . I think the changes looks good, is there anything I can do to help, maybe help validate the fix? We are using Durable Agents and Redis in our system

taofeeq-deru (dismissed):

taofeeq-deru (approved):

PR #18483, opened 2026-06-25
fix(slack): normalize trailing slash in provider baseUrl

CalebBarnes:
## Summary

The Slack provider builds all of its callback URLs by joining the configured `baseUrl` with paths like `/slack/oauth/callback`, `/slack/events/<id>`, and `/slack/commands/<id>`. When `baseUrl` carries a trailing slash — which happens naturally with values like `MASTRA_BASE_URL=https://example.com/` — the result is a malformed double-slash URL (`https://example.com//slack/oauth/callback`). Slack stores that exact string in the app manifest, breaking the OAuth flow and webhook delivery.

This normalizes the `baseUrl` by stripping trailing slashes at the single point where it is read, so every callback URL is well-formed regardless of how the value is configured.

## Changes

- Add a `stripTrailingSlash()` helper in `channels/slack/src/provider.ts`.
- Apply it in `#getBaseUrl()` (the choke point every callback URL flows through) and in `setBaseUrl()`, so both explicitly-configured and programmatically-set base URLs are normalized.
- Add unit tests covering single/multiple trailing slashes, no-op, path preservation, and a clean OAuth-callback join.

## Test plan

- `pnpm --filter @mastra/slack test` — 68 passing (5 new)
- `pnpm --filter @mastra/slack typecheck` — clean

## ELI5
This change fixes Slack links so they don’t accidentally get two slashes in a row when your app’s base URL ends with `/`. That helps Slack login and webhook links keep working normally.

## Summary
- Normalizes Slack `baseUrl` values by stripping trailing slashes when they are read or set.
- Prevents malformed callback URLs like `https://example.com//slack/oauth/callback`.
- Adds unit tests covering:
  - single and multiple trailing slashes
  - URLs without trailing slashes
  - preserving path segments
  - correct OAuth callback joining
- Includes a changeset for `@mastra/slack` documenting the patch release.


CalebBarnes:
@coderabbitai full review

abhiaiyer91 (approved):

PR #18490, opened 2026-06-25
fix(mastracode): make plan mode use a thread-scoped working file and stop cleanly on reject

DanielSLew:
Plan mode wasn't reliable: writes outside the working plan file weren't actually blocked, the plan prompt could show a placeholder path instead of the real one, and clicking "Request Changes" didn't stop the run so the model kept talking after a rejection.

The write guard was reading the harness off the wrong shape, so its mode check always came back undefined and it silently bailed:

```ts
// before — these don't exist on HarnessRequestContext, so the guard never enforced
const modeId = harness?.session?.mode?.get?.();
const threadId = harness?.session?.thread?.getId?.();

// after — the real shape
const modeId = harness?.session?.modeId;
const threadId = harness?.threadId;
```

Rejection used to only resume the tool and lean on an in-loop processor to abort, which raced the next model step. Now it awaits the resume boundary (so the rejection result is persisted) and then aborts host-side, which is deterministic:

```ts
onReject: () => {
  void (async () => {
    try {
      await state.session.respondToToolSuspension({ toolCallId, resumeData: { action: 'rejected' } });
    } finally {
      state.planRejectionAbort = true;
      state.session.abort();
    }
  })();
  resolve();
}
```

Plan mode now uses one thread-scoped working file at `.mastracode/plans/threads/<threadId>/current-plan.md`, the prompt renders that real path, `submit_plan` takes just a title, and approved plans get archived to a flat human-readable file. All covered by unit tests plus the plan-approval e2e scenarios.

## ELI5
In “plan mode,” the assistant now writes the plan to a real markdown file on disk (per thread), instead of keeping the whole plan text in chat. When you click “Request Changes” (reject the plan), the run stops right away so the assistant can’t keep talking after the rejection.

## What changed
- Plan mode now uses thread-scoped plan files on disk (a “current plan” file per thread) and submits/reads via filesystem paths.
- `submit_plan` is now path-based: it submits `{ path }` (no inline `{ title, plan }`), and the host loads the plan from disk.
- The plan prompt renders the real plan path (no placeholder).
- Plan-mode workspace write guards were fixed and tightened to enforce that plan writes are limited to allowed markdown plan locations under `.mastracode/plans/` (using the correct harness/tool-hook context shape).
- Approved plans are archived to a human-readable location, and revision diffs are computed from the real saved plan file.
- After a “Request Changes” rejection, the run aborts deterministically (waiting for the suspension/resume result to persist, then aborting host-side to prevent any trailing model output).

## Tests and fixtures
- Added/updated unit tests for plan-file write guarding, plan diff generation/size gating, and robust `submit_plan` rejection abort detection across multiple message/tool-result shapes.
- Updated TUI and E2E fixtures to `write_file` the plan markdown to disk first, then call `submit_plan` using only the saved file `path`.
- Updated prompt generation and tool-schema expectations to match the new `{ path }` contract.

## Other notes
- Plan diffs were refactored to an LCS-based line diff with a “show diff only if small” heuristic.
- Added support for custom thread/message ID generation via `MastraCodeConfig.idGenerator`.


mastracode/src/agents/prompts/plan.ts
```diff
@@ -15,7 +31,7 @@ This mode is **read-only for project files**. You must NOT modify the project.
 - Do NOT run commands that change state (no git commit, no npm install, no file creation)
 - Do NOT run build commands, tests, or scripts that have side effects
 
-The ONE exception is your plan file: you CAN write to \`.mastracode/plans/\` to create and edit your plan.
+The ONE exception is your thread-scoped plan file: you CAN write to \`${planPath}\` to create and edit your plan.
```
TylerBarnes: I think current-plan.md isn't exactly right, cause I'll often run multiple different plans in the same thread, it should probably just be .mastracode/plans/$plan-name.md and we just encourage the agent to continue using the same one without updating the title, until it's time to make a totally new plan

TylerBarnes: the issue I found was the diff is all wrong if it starts making a completely new plan later, but also we probably want to be able to look at each individual plan that was made, not just the latest one

mastracode/src/agents/prompts/plan.ts
```diff
@@ -61,24 +77,23 @@ For each step:
...
 2. **Reading**: Use \`view\` to read the plan file.
 3. **Editing**: Use \`string_replace_lsp\` for targeted edits to specific sections.
 
 ## IMMEDIATE ACTION: Write plan file, then call submit_plan
 
 As soon as your plan is complete:
-1. Write it to \`.mastracode/plans/<slug>.md\` (slugified from your title) using \`write_file\`
-2. Call \`submit_plan\` with the plan content
+1. Write it to \`${planPath}\` using \`write_file\`
+2. Call \`submit_plan\` with only a short \`title\` — never the plan body or a path
```
TylerBarnes: why not just remove the ability to pass the body to the tool? The tool should really just take the path on disk to the plan file (so there can be more than one plan over time)

mastracode/src/tui/handlers/prompts.ts
```diff
@@ -272,46 +287,88 @@ function formatPlanGoalObjective(title: string, plan: string): string {
...
+  const oldLines = previousPlan.split('\n');
+  const newLines = plan.split('\n');
+  const maxLines = Math.max(oldLines.length, newLines.length);
+
+  let changed = 0;
+  for (let i = 0; i < maxLines; i++) {
+    if (oldLines[i] !== newLines[i]) changed++;
+  }
+
+  return changed / newLines.length <= 0.5;
```
TylerBarnes: what do we use for making the diff? I don't think this check is good enough for many cases, we should generate the diff then compare the diff length to the plan length, instead of trying to predict what the diff will be (line index checks will be totally off from a real diff)

packages/core/src/tools/builtin/submit-plan.ts
```diff
@@ -96,14 +97,17 @@ export const submitPlanTool = createTool({
 
       const suspend = context?.agent?.suspend;
       if (suspend) {
-        await suspend({ title: resolvedTitle, plan });
+        // The agent writes the plan to its working file; the host reads that file to
+        // render the approval UI. The tool only knows the title, so it suspends with an
+        // empty `plan` placeholder for the host to fill from disk.
+        await suspend({ title: resolvedTitle, plan: '' });
```
TylerBarnes: I think this means we need to change how the actual plan tool works or make a new one

TylerBarnes (changes requested):

mastracode/src/agents/prompts/plan.ts
DanielSLew: AI says: Reworked plan mode along the lines you suggested across the threads here. Summary of what changed (pushed in `2b9141b23d`):

- **`submit_plan` now takes a `path`, not a title/body.** It points at the markdown file the agent wrote, so multiple plans can exist over time and the host reads/diffs the exact file the user reviews. (r3477292726, r3477318465)
- **Named plan files.** The agent writes each plan to its own `.mastracode/plans/<name>.md` and reuses it across revisions, only creating a new file for a genuinely new plan. No more single `current-plan.md` — every plan stays on disk to look back at, and diffs stop breaking when a new plan starts. The write guard now allows any `.md` directly inside `.mastracode/plans/` (and nothing else). On approval the local file is kept and also archived globally. (r3477283280, r3477287921)
- **Real diff for the size gate.** Replaced the line-index prediction with an actual LCS diff and gate on the real diff size (changed lines vs the whole diff) instead of guessing what the diff will be, so an inserted/removed line no longer marks everything after it as changed. (r3477309291)

Tests, typecheck, and lint pass, and the plan-approval + tool-schema e2e scenarios are green.

Daniel

DanielSLew: Addressed in `a1d259fbd1`. Plan mode now writes stable named files under `.mastracode/plans/<slug>.md`, reuses the same title/file for revisions, and keeps each individual plan file on disk so later plans do not overwrite earlier ones.

mastracode/src/agents/prompts/plan.ts
DanielSLew: Addressed in `a1d259fbd1`. I adjusted the contract to the intended title-only version: `submit_plan` now takes only `{ title }`; Mastra Code derives `.mastracode/plans/<slug(title)>.md` and reads that file. The plan body and path are no longer accepted as model-controlled tool args.

mastracode/src/tui/handlers/prompts.ts
DanielSLew: Addressed in `a1d259fbd1`. Diff display now uses the real LCS-based unified diff and gates on actual diff size relative to the plan, instead of predicting changes by line index.

packages/core/src/tools/builtin/submit-plan.ts
DanielSLew: Addressed in `a1d259fbd1`. The built-in `submit_plan` tool now has a title-only schema and carries submitted plan snapshots through resume results for rendering/replay.

mastracode/src/tui/handlers/prompts.ts
```diff
@@ -275,43 +292,56 @@ function formatPlanGoalObjective(title: string, plan: string): string {
...
-  const previousPlan = state.previousPlanSnapshot?.plan;
-
-  // Save a snapshot of this submission and write to local plans dir
-  state.previousPlanSnapshot = { title, plan };
+  // submit_plan carries only the path to the plan file. Read the real content
+  // from disk so the approval UI renders and diffs the actual plan.
   const projectPath = (state.session.state.get() as any)?.projectPath as string | undefined;
-  if (projectPath) {
-    savePlanSnapshot({ title, plan, projectPath }).catch(() => {});
+  const planPath = projectPath ? resolvePlanPath(projectPath, submittedPath) : undefined;
```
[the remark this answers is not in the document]
DanielSLew: Addressed in `a1d259fbd1`. The untrusted `path` arg was removed from `submit_plan`; the host derives the plan file path from the submitted title, so this arbitrary-read/path-traversal path is no longer exposed.

mastracode/src/tui/render-messages.ts
```diff
@@ -915,19 +915,29 @@ export async function renderExistingMessages(state: TUIState): Promise<void> {
...
             }
 
-            if (args?.title && args?.plan) {
+            if (args?.path) {
+              // The plan body lives in the submitted file, not in the tool args.
+              // Recover it from disk when it still exists; derive the title from
+              // the file's leading heading and fall back to the path otherwise.
+              const sessionState = state.session.state.get() as any;
+              const projectPath = sessionState?.projectPath as string | undefined;
+              const absPath = projectPath ? resolvePlanPath(projectPath, args.path) : undefined;
```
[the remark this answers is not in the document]
DanielSLew: Addressed in `a1d259fbd1`. History replay no longer trusts a submitted path arg; it uses the title-derived path and, when available, the persisted submitted plan snapshot from the tool result.

mastracode/src/tui/render-messages.ts
```diff
@@ -915,19 +915,29 @@ export async function renderExistingMessages(state: TUIState): Promise<void> {
...
               state.chatContainer.addChild(planResult);
               replacedWithInline = true;
-              // Restore previousPlanSnapshot so that if the agent resubmits after
-              // a restart, the diff can be computed against the last known plan.
-              state.previousPlanSnapshot = { title: args.title, plan: args.plan };
+              // Restore previousPlanSnapshot (keyed by path) so that if the agent
+              // resubmits after a restart, the diff can be computed against the
+              // last known plan.
+              state.previousPlanSnapshot = { path: args.path, plan: planBody };
             }
```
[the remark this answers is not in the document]
DanielSLew: Addressed in `a1d259fbd1`. The submitted plan body/path/title are persisted in the `submit_plan` tool result on approval/rejection, and replay restores `previousPlanSnapshot` from that persisted body instead of rereading the mutable plan file.

mastracode/src/utils/plan-diff.ts
```diff
@@ -0,0 +1,70 @@
...
+
+export interface DiffEntry {
+  type: 'added' | 'removed' | 'context';
+  text: string;
+}
+
+/** Produce structured diff entries between two plan texts using an LCS diff. */
+export function generatePlanDiff(oldText: string, newText: string): DiffEntry[] {
+  const oldLines = oldText.split('\n');
+  const newLines = newText.split('\n');
```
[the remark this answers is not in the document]
DanielSLew: Addressed in `a1d259fbd1`. `generatePlanDiff()` now splits with `/\r?\n/`, and I added a CRLF/LF normalization test.

TylerBarnes (approved):

PR #18512, opened 2026-06-26
fix(e2b): use per-mount credentials file for S3 and GCS mounts

### LINKED ISSUE #18513: S3 and GCS mounts share a single credentials file, racing on concurrent mounts
## Description

`mountS3` and `mountGCS` write the mount credentials to a single fixed path
(`/tmp/.passwd-s3fs` and `/tmp/gcs-key.json`) and rewrite it with a non-atomic
`sudo rm -f` → `files.write` → `chmod` sequence on **every** mount
(`workspaces/e2b/src/sandbox/mounts/s3.ts`, `gcs.ts`).

When two mounts run concurrently in the same sandbox — e.g. mounting several
buckets at once, or re-establishing FUSE mounts after a pause/resume — those
rewrites interleave. One mount's `write`/`chmod` races another's `rm -f`, which
yields an `EACCES` and a failed mount, or (worse) a mount that ends up reading
another mount's credentials.

`mountAzure` already avoids this by deriving a per-mount path from a hash of the
mount point (`/tmp/.blobfuse2-config-${mountHash}.yaml`); S3 and GCS never
adopted that pattern.

## Expected

Concurrent S3/GCS mounts in the same sandbox should not contend on a shared
credentials file.

## Suggested fix

Give each mount its own credentials file by deriving the path from a hash of the
mount point, matching `mountAzure`.

## Affected

`@mastra/e2b` — `workspaces/e2b/src/sandbox/mounts/s3.ts`, `gcs.ts`

IslandHanX:
Closes #18513

## What

`mountS3` and `mountGCS` write the mount credentials to a single fixed path
(`/tmp/.passwd-s3fs` and `/tmp/gcs-key.json`) and rewrite it with a non-atomic
`sudo rm -f` → `files.write` → `chmod` sequence on **every** mount.

## Why

When two mounts run concurrently in the same sandbox — e.g. mounting several
buckets at once, or re-establishing FUSE mounts after a pause/resume — those
rewrites interleave. One mount's `write`/`chmod` races another's `rm -f`, which
yields an `EACCES` and a failed mount, or (worse) a mount that ends up reading
another mount's credentials.

`mountAzure` already avoids this by deriving a per-mount path from a hash of the
mount point:

```ts
const mountHash = createHash('md5').update(mountPath).digest('hex').slice(0, 8);
const configPath = `/tmp/.blobfuse2-config-${mountHash}.yaml`;
```

## What this PR does

Applies the same per-mount pattern to S3 and GCS, so each mount gets its own
credentials file and the shared resource (and the race) is gone:

- `/tmp/.passwd-s3fs` → `/tmp/.passwd-s3fs-${mountHash}`
- `/tmp/gcs-key.json` → `/tmp/gcs-key-${mountHash}.json`

`mountHash` is deterministic per mount point, so re-mounting the same path reuses
the same file. Azure is unchanged (already correct).

## Testing

The mount helpers need a live E2B sandbox and have no unit-test harness in the
repo, so this was verified by reasoning through the rewrite sequence and by
matching the existing `mountAzure` implementation. A changeset (`@mastra/e2b`
patch) is included.

## ELI5
When the sandbox mounts S3 or GCS, it temporarily writes credential files to `/tmp`. This PR makes those files unique per mount point, so two mounts running at the same time can’t overwrite each other’s credentials and cause failures.

## Summary
- Updated `mountS3` to write credentials to a deterministic per-mount temp path derived from `mountPath` (replacing `/tmp/.passwd-s3fs` with `/tmp/.passwd-s3fs-${mountHash}`).
- Updated `mountGCS` to write service account keys to a deterministic per-mount temp path derived from `mountPath` (replacing `/tmp/gcs-key.json` with `/tmp/gcs-key-${mountHash}.json`).
- Added a changeset for a `patch` release of `@mastra/e2b` documenting the concurrency/race fix.
- Left `mountAzure` unchanged since it already uses per-mount hashed config paths.

## Why
Previously, concurrent mounts could interleave the `rm -f` → write → `chmod` steps on shared credential files, leading to `EACCES`, failed mounts, or credentials being read from the wrong mount. Using per-mount credential file paths removes that shared-file contention.

### DIFF
--- a/.changeset/e2b-per-mount-credential-files.md
+++ b/.changeset/e2b-per-mount-credential-files.md
@@ -1,0 +1,15 @@
+---
+'@mastra/e2b': patch
+---
+
+Fix concurrent S3/GCS mounts racing on a shared credentials file
+
+`mountS3` and `mountGCS` wrote credentials to a single fixed path
+(`/tmp/.passwd-s3fs`, `/tmp/gcs-key.json`) and rewrote it (`rm` → `write` →
+`chmod`) on every mount. When two mounts run concurrently in the same sandbox
+(e.g. mounting several buckets, or re-establishing mounts after a pause/resume),
+their rewrites interleave — one mount's `write`/`chmod` races another's `rm` —
+causing `EACCES` and a failed mount, or a mount reading another mount's
+credentials. Each mount now derives a per-mount path from a hash of its mount
+point (the same approach `mountAzure` already uses), removing the shared
+resource. Azure was already unaffected.
--- a/.changeset/e2b-per-mount-credential-files.md
+++ b/.changeset/e2b-per-mount-credential-files.md
@@ -2,14 +2,6 @@
 '@mastra/e2b': patch
 ---
 
-Fix concurrent S3/GCS mounts racing on a shared credentials file
+Make concurrent S3 and GCS mounts reliable in the same sandbox
 
-`mountS3` and `mountGCS` wrote credentials to a single fixed path
-(`/tmp/.passwd-s3fs`, `/tmp/gcs-key.json`) and rewrote it (`rm` → `write` →
-`chmod`) on every mount. When two mounts run concurrently in the same sandbox
-(e.g. mounting several buckets, or re-establishing mounts after a pause/resume),
-their rewrites interleave — one mount's `write`/`chmod` races another's `rm` —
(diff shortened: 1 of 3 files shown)


intojhanurag (approved):
LGTM , Thanks  !

PR #18552, opened 2026-06-27
feat(core): thread agent and thread identity through the processor surface

CalebBarnes:
## Summary

Processors today have no first-class way to learn which agent, thread, or resource a run belongs to. Consumers that need this (for example the channels output processor) are forced to depend on request-scoped side channels that are not always populated. This PR adds `agentId`, `agentName`, `threadId`, and `resourceId` as flat optional fields on `ProcessorContext`, so every processor method receives them directly.

The fields are optional and `undefined` for non-agent or threadless pipelines, so existing processors keep working unchanged.

```ts
const myProcessor: Processor = {
  id: 'my-processor',
  processOutputStream: async ({ part, agentId, threadId, resourceId }) => {
    // agentId, threadId, resourceId now available here
    return part;
  },
};
```

## What changed

- Added the four identity fields to `ProcessorContext` (propagates to all method arg types).
- Added `agentId` to the `ProcessorRunner` constructor alongside `agentName`; `threadId`/`resourceId` are passed per-call into the public entry methods.
- Injected the identity fields into every hand-built processor args object — both the in-process regular-processor objects and the workflow-as-processor inline objects.
- Added the identity fields to the workflow step schemas and types so they survive the workflow serialization boundary, and spread them into `baseContext` in both the default and evented workflow executors.
- Wired `agentId` (and proper `agentName`/`threadId`/`resourceId` where available) into all non-test `ProcessorRunner` construction sites.

All four fields are plain strings, so they cross the workflow step serialization boundary safely. No live handles or closures were added to any step schema.

## Test plan

- New cross-path symmetry test in `runner.test.ts`: a processor receives identical `agentId`/`agentName`/`threadId`/`resourceId` whether it runs in-process or as a workflow.
- `pnpm build:core` — passing.
- `pnpm --filter ./packages/core check` (typecheck) — passing.
- Focused processor tests (`src/processors/`, 38 files) — passing.
- Loop, agent, and workflow tests (125 files) — passing.

This is PR #1 of a stack. It is independently useful and unblocks a follow-up that makes the channels output processor self-sufficient.

### DIFF
--- a/.changeset/two-papayas-taste.md
+++ b/.changeset/two-papayas-taste.md
@@ -1,0 +1,17 @@
+---
+'@mastra/core': minor
+---
+
+Added agent and thread identity to processors. Every processor method now receives `agentId`, `agentName`, `threadId`, and `resourceId` so processors can tell which agent, thread, and resource a run belongs to without inspecting `requestContext`. The same fields arrive identically whether the processor runs in-process or as a workflow.
+
+```ts
+const myProcessor: Processor = {
+  id: 'my-processor',
+  processOutputStream: async ({ part, agentId, threadId, resourceId }) => {
+    // agentId, threadId, resourceId now available here
+    return part;
+  },
+};
+```
+
+The fields are optional and undefined for non-agent or threadless pipelines, so existing processors keep working unchanged.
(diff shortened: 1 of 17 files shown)


CalebBarnes:
Superseded by #18630. The self-sufficient channels output processor (the only consumer of this identity work) instead sources the run's threadId from the framework-populated `MastraMemory` request-context key via `parseMemoryRequestContext`, so threading agentId/agentName/threadId/resourceId through the entire processor surface is no longer required for this work. Closing to avoid a stale stack.

PR #18553, opened 2026-06-27
fix(core): broadcast channel output for runs without inbound request context

CalebBarnes:
## What this ships

Agent runs on a channel-backed thread (Slack, Discord, etc.) now post their output back to the channel **even when the run did not start from an inbound platform message**. Heartbeat-triggered runs, Studio runs, and custom-UI / user-code runs against a channel thread previously produced no channel output at all — they were silently dropped.

## Why

`ChatChannelOutputProcessor` only rendered to the channel when `AgentChannels` stashed a render context on `requestContext` during an inbound webhook (`processChatMessage`, approve/decline). Any run that did not originate from a platform message had no such context, so the processor early-returned and nothing reached the channel. Channel delivery was effectively coupled to the inbound request path.

## How

- `ChatChannelOutputProcessor` is now bound to its owning `AgentChannels` at construction.
- When no inbound render context is present, the processor reconstructs it from the run's `threadId` (provided by the processor-identity work) via a new `AgentChannels.buildRenderContextForThread(threadId)`. That method reads the thread's persisted channel coordinates (`channel_platform`, `channel_externalThreadId`), materializes the live channel thread handle, and delegates to the **same** `_buildRenderContext` the inbound paths use — so both paths produce an identical render context (single source of truth).
- The inbound `requestContext` path is unchanged and still takes precedence (fast path).
- Non-channel threads pass through untouched. The resolved render context (or `null`) is cached on `state` so it is built once per run, not per chunk.

Nothing new is persisted: every non-serializable field (adapter, closures, logger) is rebuilt live from the bound `AgentChannels`.

## Stacking

Stacked on #18552 (processor context identity), which threads `threadId` to the output processor. This PR targets that branch and stays a draft until #18552 merges.

## Test plan

- `pnpm build:core`
- `pnpm --filter ./packages/core check` (typecheck clean)
- `pnpm --filter ./packages/core test -- src/channels/` — all 238 channels tests pass, including 5 new fallback tests in `output-processor.test.ts`:
  - broadcasts on a channel-backed thread with no `requestContext` key
  - passes through when the thread has no channel metadata
  - passes through when the thread does not exist in storage
  - passes through when no `threadId` is provided
  - prefers the inbound `requestContext` render context over the fallback (no regression)

### DIFF
--- a/.changeset/wide-signs-send.md
+++ b/.changeset/wide-signs-send.md
@@ -1,0 +1,5 @@
+---
+'@mastra/core': patch
+---
+
+Fixed channel broadcasting so agent runs on a channel-backed thread post back to the channel even when they did not start from an inbound platform message. Previously only runs triggered by an incoming Slack/Discord/etc. message would render to the channel; heartbeat, Studio, and custom UI runs were silently dropped. The channels output processor now reconstructs the channel destination from the thread itself, so any run on a channel-backed thread delivers its output.
--- a/packages/core/src/channels/__tests__/output-processor.test.ts
+++ b/packages/core/src/channels/__tests__/output-processor.test.ts
@@ -1,5 +1,6 @@
 import { describe, it, expect, vi, beforeAll } from 'vitest';
 
+import { InMemoryStore } from '../../storage/mock';
 import { AgentChannels } from '../agent-channels';
 import { getChatModule } from '../chat-lazy';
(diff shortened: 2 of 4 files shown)


CalebBarnes:
Superseded by #18630, which collapses this two-PR stack into a single self-contained PR. Instead of threading new identity fields through the whole processor surface (old #18552), #18630 reads the run's threadId from the framework-populated `MastraMemory` request-context key via the validated `parseMemoryRequestContext` accessor — no processor API changes needed. The channels fix (this PR's content) is unchanged.

PR #18561, opened 2026-06-27
fix(core): keep non-text parts when flushing BatchPartsProcessor batches

### LINKED ISSUE #18560: BatchPartsProcessor drops non-text parts when emitOnNonText is false
## Description

`BatchPartsProcessor` with `emitOnNonText: false` buffers non-text parts (tool-call, tool-result, object, reasoning, …) into its batch alongside text deltas. When the batch is flushed (on `batchSize`, `maxWaitTime`, or `flush()`), `flushBatch` combines the text deltas and then clears the **entire** batch (`state.batch = []`), silently discarding every buffered non-text part whenever the batch also contains text.

## Reproduction

With `emitOnNonText: false`, feed a `text-delta`, a `tool-call`, then a `text-delta` through `processOutputStream` (shared state), then drain via `flush()`:

- **Actual:** only the combined `text-delta` is emitted — the `tool-call` is dropped.
- **Expected:** `text-delta`, `tool-call`, `text-delta` (order preserved, nothing dropped).

Verified against the real processor in `packages/core/src/processors/processors/batch-parts.test.ts` (a new test fails on `main` and passes after the fix).

## Fix

Emit the batch one part at a time, preserving order, instead of combining text and clearing the batch. A PR is ready.

Osamaali313:
Fixes #18560

## Bug

`BatchPartsProcessor` with `emitOnNonText: false` buffers non-text parts (tool-call, tool-result, object, reasoning, …) into the batch alongside text deltas (`processOutputStream` falls through to `state.batch.push(part)`). But `flushBatch` combined the text deltas and then cleared the **whole** batch:

```ts
const textChunks = state.batch.filter(p => p.type === 'text-delta');
if (textChunks.length > 0) {
  // ...combine text...
  state.batch = []; // <-- silently drops every buffered non-text part
  return combinedChunk;
}
```

So a stream like `text, tool-call, text` emerges as just the combined text — the `tool-call` is lost.

## Reproduction (real processor, vitest)

A new test drives `text-delta, tool-call, text-delta` through `processOutputStream` with `emitOnNonText: false`, then drains via `flush()`:

```
RED  (main):  emitted ['text-delta']                          # tool-call dropped
GREEN (fix):  emitted ['text-delta','tool-call','text-delta'] # preserved, ordered
```

Full `batch-parts.test.ts` suite: **17 passed** (16 existing + 1 new), no regressions.

## Fix

Emit the batch one part at a time, preserving order: if the batch leads with a non-text part, emit it directly; otherwise combine the **leading contiguous run** of text deltas and keep the remainder for subsequent flushes. All-text batches still combine into a single chunk (existing behavior unchanged).

Includes a changeset (`@mastra/core` patch). Commit is signed (Verified).

## ELI5
When Mastra batches streaming output, it sometimes mixed “normal text” with other messages (like tool calls). On flush, the older code combined the text but accidentally threw away the non-text messages. This fix makes sure everything is kept and emitted in the right order.

## Summary
- Fixed `BatchPartsProcessor.flushBatch` so that when `emitOnNonText: false`, non-text parts (tool-call, tool-result, object, reasoning, etc.) are not dropped during batching.
- Preserves order by emitting either:
  - the first non-text part directly if the batch starts with non-text, or
  - a single combined `text-delta` only from the leading contiguous run of text, leaving later parts buffered for subsequent flushes.
- Added a regression test asserting that a buffered `tool-call` is retained and emitted in order between text deltas (`text-delta`, `tool-call`, `text-delta`), including the expected `toolCallId`.
- Included a `@mastra/core` changeset documenting the user-visible fix (non-text stream parts no longer silently disappear when `emitOnNonText: false`).


Osamaali313:
Thanks @coderabbitai — updated the changeset to be outcome-first per the repo's changeset guidelines (a26c3f7). It now leads with the user-visible fix (non-text stream parts are no longer dropped when `emitOnNonText: false`; tool calls, tool results, objects, and reasoning stay in order with the text) instead of the internal implementation detail.

PR #18563, opened 2026-06-28
feat(memory): expose providerMetadata on observational memory hooks

### LINKED ISSUE #18562: Expose providerMetadata on Observational Memory ObserveHooks (onObservationEnd / onReflectionEnd)
### Summary

Observational Memory's lifecycle hooks — `ObserveHooks.onObservationEnd` and `onReflectionEnd` — currently surface only token `usage` (`{ inputTokens, outputTokens, totalTokens }`). The OM model call's `providerMetadata` — where providers such as the Vercel AI Gateway report per-call economics like `cost` and `generationId` (under `providerMetadata.gateway`) — is discarded before the hook fires.

### Problem

A consumer who wants to account for the cost of the observer/reflector model calls has no way to do it from the hooks. The only workaround is to wrap the OM observer/reflector models in a `LanguageModelMiddleware` purely to read `providerMetadata` off the model result — duplicating accounting plumbing for data the framework already has in hand at the hook boundary.

### Proposed solution

Pass the OM model call's `providerMetadata` through to the hook results, alongside the existing `usage`:

\`\`\`ts
onObservationEnd?: (result: { usage?: ObserveHookUsage; error?: Error; providerMetadata?: ProviderMetadata }) => void;
onReflectionEnd?:  (result: { usage?: ObserveHookUsage; error?: Error; providerMetadata?: ProviderMetadata }) => void;
\`\`\`

`providerMetadata` already rides on the `getFullOutput()` result the runners read `usage` from, so it threads through with the canonical `ProviderMetadata` type from `@mastra/core/stream` and zero cast. The change is additive and optional, so existing hook consumers are unaffected, and it is provider-agnostic — the field carries the whole `providerMetadata` (the Gateway's cost/generationId live under `providerMetadata.gateway`), not a hardcoded `cost`.

I have a PR ready that implements this with regression tests.

brennanmceachran:
## Summary

`ObserveHooks.onObservationEnd` and `onReflectionEnd` surfaced only token `usage`. The OM model call's `providerMetadata` — where providers such as the Vercel AI Gateway report per-call economics (`cost`, `generationId`) under `providerMetadata.gateway` — was dropped before the hook fired, so consumers couldn't account for OM model cost without wrapping the observer/reflector models in a `LanguageModelMiddleware`.

This threads the OM model call's `providerMetadata` to both hook results, along the exact path `usage` already travels:

```ts
onObservationEnd?: (result: { usage?: ObserveHookUsage; error?: Error; providerMetadata?: ProviderMetadata }) => void;
onReflectionEnd?:  (result: { usage?: ObserveHookUsage; error?: Error; providerMetadata?: ProviderMetadata }) => void;
```

- **Provider-agnostic** — exposes the whole `providerMetadata`, not a hardcoded `cost`.
- **Zero-cost typing** — `ProviderMetadata` from `@mastra/core/stream` is the exact static type of `getFullOutput().providerMetadata`; no new dependency, no cast.
- **Additive + optional** — existing hook consumers are unaffected; the field is simply absent when the provider emits no metadata (the live hook-fire sites use a conditional spread, so the key never appears as `undefined`).
- **Batched/retry semantics** — for resource-scoped batched observations and multi-attempt reflections it reflects the last batch/attempt that emitted provider metadata (`usage` is still summed; per-call provider metadata can't be meaningfully merged). Documented in JSDoc on the hook types.

## How it works

`usage` is read off the observer/reflector model's `getFullOutput()` result in `observer-runner` / `reflector-runner`. `providerMetadata` sits next to it on the same result, so it's lifted verbatim and carried through `ObserverOutput` → `ObservationRunResult` (and the resource-scoped and reflection accumulators) to the hook. The async-failure path leaves it absent (no model result).

## Testing

- Three new tests in `observational-memory-api.test.ts`: observer `providerMetadata` reaches `onObservationEnd`; reflector `providerMetadata` reaches `onReflectionEnd`; and it stays **absent** when the model emits none (zero-breaking-change guard). Verified failing before the change and passing after.
- Full observational-memory unit suite green (899 tests); `tsc --noEmit`, `eslint`, and `prettier --check` clean.

Closes #18562

---
🤖 This PR was authored with [Claude Code](https://claude.com/claude-code) using Claude Opus 4.8.

## ELI5
This PR lets the observation/reflection hook callbacks get extra information from the model provider (like cost or a generation ID), not just token usage. It’s optional—your hook code won’t see it unless the provider actually sends it.

## Summary
- Added an optional `providerMetadata?: ProviderMetadata` field to `ObserveHooks.onObservationEnd` and `ObserveHooks.onReflectionEnd` (using the canonical `ProviderMetadata` type from `@mastra/core/stream`).
- Threaded `providerMetadata` through the observational-memory observation and reflection pipelines, including:
  - resource-scoped/batched observations (the last emitted batch’s `providerMetadata` is surfaced)
  - multi-attempt reflections (the last emitted attempt’s `providerMetadata` is surfaced, while `usage` behavior remains summed as before).
- Kept the change additive and backwards-compatible:
  - when the provider doesn’t emit metadata, `providerMetadata` is omitted entirely from the hook result (not just set to `undefined`).
- Documented the behavior in a new changeset.
- Added regression tests verifying:
  - `providerMetadata` reaches both `onObservationEnd` and `onReflectionEnd` when provided
  - `providerMetadata` is absent from hook results when omitted by the provider
  - the batched/resource-scoped observer path passes through the last-emitted `providerMetadata`.


brennanmceachran:
Addressed the CodeRabbit review in d99fe35:

- **Changeset** now leads with a short `onObservationEnd` / `onReflectionEnd` example reading `providerMetadata`, with the implementation detail trimmed.
- The "absent when not emitted" test now asserts `not.toHaveProperty('providerMetadata')`, pinning the omission contract (the live hook-fire sites conditionally spread the field) rather than merely an `undefined` value.
- Added coverage for the batched **resource-scoped** observer path (`callMultiThread` → `lastBatchProviderMetadata` → `onObservationEnd`), which the single-thread tests didn't exercise. The multi-batch / multi-attempt *last-emitted-wins* distinctness is type-checked and documented in the hook JSDoc — glad to add deterministic distinctness tests (stateful per-call metadata + forced batching / degenerate-retry) if you'd like them in this PR.

`tsc`, `eslint`, `prettier --check`, and the observational-memory unit suite (now 149) are green.

TylerBarnes (approved):

PR #9048, opened 2025-10-21
Improve stream workflow output

wardpeet:
## Description

Adds `fullstream` to the modeloutput and simplifies the code for observeStreamVnext

### Additional Changes

- Fixed TypeScript signature compatibility issues in `EventedRun` class
  - Updated `watch` and `watchAsync` methods to use proper function overloads instead of generic conditional types
  - This ensures compatibility with the base `Run` class signatures
  - Resolves type assignment errors where conditional types were incompatible with overloaded signatures

## Type of Change

- [x] Bug fix (non-breaking change that fixes an issue)
- [ ] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation update
- [x] Code refactoring
- [ ] Performance improvement
- [ ] Test update

## Checklist

- [ ] I have made corresponding changes to the documentation (if applicable)
- [ ] I have added tests that prove my fix is effective or that my feature works

PR #18571, opened 2026-06-28
fix(core): isolate fullStream consumer cancel and surface workflow stream pipeline errors

### LINKED ISSUE #18570: [BUG] WorkflowRunOutput: cancelling one `fullStream` consumer detaches ALL consumers, and pipeline errors are swallowed by a bare `console.log`
### Describe the Bug

`WorkflowRunOutput.fullStream` is a getter that returns a **new** replayable `ReadableStream` per access, all fed by one shared `EventEmitter` so several consumers can read the same run at once (the buffered-replay design exists precisely for this). Two defects break that contract:

1. **`cancel()` on any one consumer breaks every other consumer.** The stream's `cancel()` callback calls `self.#emitter.removeAllListeners()`, which removes the `chunk`/`finish` listeners of *all* other `fullStream` instances. Those consumers stop receiving chunks and — because their `finish` listener is also gone — **never close**: their readers await forever.

2. **Stream pipeline errors are swallowed.** Both the constructor and `resume()` end their `pipeTo()` chain with `console.log(' something went wrong', reason)`. When the source stream rejects (e.g. a provider/transport failure mid-run), `close()` never runs, the `finish` event never fires, the delayed `result`/`usage` promises are never settled, and `fullStream` consumers hang — the only trace is a stray `console.log` (not even the framework logger).

`packages/core/src/stream/RunOutput.ts` (the per-call stream + the destructive cancel):

```ts
get fullStream(): ReadableStream<WorkflowStreamEvent> {
  const self = this;
  return new ReadableStream<WorkflowStreamEvent>({
    start(controller) {
      self.#bufferedChunks.forEach(chunk => controller.enqueue(chunk));
      // ...
      self.#emitter.on('chunk', chunkHandler);
      self.#emitter.on('finish', finishHandler);
    },
    // ...
    cancel() {
      self.#emitter.removeAllListeners();   // <-- nukes every other consumer's handlers
    },
  });
}
```

The error swallow is in the constructor's `pipeTo().catch()` and the identical copy in `resume()`:

```ts
.catch(reason => {
  // eslint-disable-next-line no-console
  console.log(' something went wrong', reason);
});
```

The `close()` handler is the only place `finish` is emitted and the delayed promises settle — it is skipped entirely when `pipeTo` rejects.

### Steps To Reproduce

```ts
const run = await workflow.createRunAsync();
const output = run.streamAsync(...); // WorkflowRunOutput

const a = output.fullStream.getReader();
const b = output.fullStream.getReader();

await a.cancel();                 // consumer A opts out
// consumer B now receives no further chunks and never closes:
while (true) {
  const { done, value } = await b.read();   // hangs after A's cancel
  if (done) break;
}
```

1. Attach two `fullStream` consumers to the same run.
2. Cancel one of them — the other stops receiving chunks and never closes.
3. For defect 2: make the source stream error mid-run (e.g. provider abort) — observe `' something went wrong'` on stdout while `await output.result` / `output.usage` and any `for await (const c of output.fullStream)` loops never terminate.

Reproducible on a fresh `npx create-mastra@latest` project by attaching two `fullStream` consumers to a streamed run and cancelling one, or by making the run's stream error mid-flight.  https://github.com/mastra-ai/mastra

### Expected Behavior

- Cancelling one `fullStream` consumer should detach only that consumer's listeners; other consumers continue and close on `finish`.
- A pipeline error should mark the run `failed`, emit a terminal `workflow-finish` event (so consumers close), settle/reject the delayed `result`/`usage` promises with the error, and log through a real error channel instead of a bare `console.log`.

### Environment Information

```block
System:
    Memory: 4.94 GB / 15.49 GB
    Container: Yes
    Shell: 5.2.21 - /bin/bash
  Binaries:
    Node: 20.19.6
    Yarn: 1.22.22
    npm: 10.8.2
```

### Verification

Abuhaithem:
`WorkflowRunOutput.fullStream` returns a new replayable `ReadableStream` per access, all fed by one shared `EventEmitter`, so several consumers can read the same run at once. When the underlying stream pipeline errored, that contract broke:

**Stream pipeline errors were swallowed.** Both the constructor and `resume()` ended their `pipeTo()` chain with `console.log(' something went wrong', reason)`. When the source stream rejected (e.g. a provider/transport failure mid-run), `close()` never ran, so `finish` never fired, the delayed `result`/`usage` promises never settled, and `fullStream` consumers hung — the only trace was a stray `console.log`.

```ts
const result = await run.stream(input).result;
// Before: hangs forever when the underlying stream errors.
// After:  rejects with the error; the run is marked `failed` and consumers close.
```

> **Note:** This PR originally also fixed cancel-isolation (cancelling one `fullStream` consumer detached all of them). That fix landed on main independently via #19745, and this branch has been reconciled with it — this PR now only ships the pipeline-error fix, plus a regression test for cancel isolation.

### Changes

- **Finalize on pipeline error.** Added a private `#finalizeWithError(reason)` that mirrors the `close()` path but marks the run `failed`: it records the error, emits a terminal `workflow-finish` chunk (with the error in `metadata`) so consumers close, rejects any still-pending delayed promises (`result`/`usage`) with the error, and logs via `console.error` instead of the swallowing `console.log`. Both `pipeTo().catch()` blocks (constructor and `resume()`) now call it. It follows the same "reject pending promises on stream error" design `MastraModelOutput` already uses.
- **Tests** covering pipeline-error propagation on the initial stream (`result`/`usage` reject and consumers receive a terminal failed `workflow-finish` instead of hanging), pipeline-error propagation after `resume()` (the re-armed promises reject and the post-resume consumer closes), and a cancel-isolation regression test complementing the one from #19745. Each test reads through a timeout race so a regression fails fast instead of hanging the suite.

No public API changes; behavior only changes for the previously-broken case (erroring streams).

**Files changed**
- `packages/core/src/stream/RunOutput.ts` — add `#finalizeWithError`; wire both `pipeTo().catch()` blocks to it
- `packages/core/src/stream/RunOutput.test.ts` — pipeline-error tests (constructor + resume paths) and cancel-isolation regression test
- `.changeset/upset-goats-judge.md` — patch changeset (scoped to the pipeline-error fix)

## Related issue(s)

Fixes #18570

## Type of change

- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [ ] Test update

## How it was tested

- **New unit tests** (`RunOutput.test.ts`): a source-stream error rejects `result`/`usage` with the error and delivers a terminal failed `workflow-finish` to consumers — on both the constructor and `resume()` pipe wirings; cancelling one `fullStream` consumer still delivers post-cancel chunks to another consumer and closes it.
- Existing `WorkflowRunOutput` tests still pass (no regressions).
- Run locally with:
  ```bash
  pnpm --filter ./packages/core exec vitest run src/stream/RunOutput.test.ts
  ```
  → 6 passed, no type errors. `pnpm --filter ./packages/core check` clean.

## Checklist

## ELI5

This PR keeps workflow streams working for multiple readers. If one reader stops, the others continue. If the stream fails, all readers receive a clear failure and do not wait forever.

## Summary

- Isolated `fullStream` consumer cancellation so other consumers continue receiving chunks and close normally.
- Handled pipeline errors during construction and `resume()`:
  - Marks the run as failed.
  - Emits a terminal `workflow-finish` event.
  - Rejects pending `result` and `usage` promises.
  - Logs errors with `console.error`.
- Added tests for cancellation isolation and pipeline-error propagation.


abhiaiyer91 (approved):

abhiaiyer91 (approved):
Reviewed in depth. The core fix is solid: `#finalizeWithError` mirrors the `close()` path, follows the same reject-pending-on-error design `MastraModelOutput` already uses, is guarded against double-finalization, and is wired into both `pipeTo().catch()` blocks (constructor and `resume()`). Tests cover the contract on both paths — `result`/`usage` reject and consumers receive a terminal failed `workflow-finish` instead of hanging forever.

Housekeeping was applied on the branch: the changeset and description are now scoped to the pipeline-error fix (the cancel-isolation half landed independently via #19745), and a `resume()`-path error test was added.

Verified locally: all 6 `RunOutput.test.ts` tests pass, core typecheck clean.

One deliberate call, on the record: `#finalizeWithError` force-sets status to `failed` even over `suspended`/`canceled`. That matches the existing `rejectResults()` convention, and storage remains authoritative for resume decisions, so it's a cosmetic edge at worst.

abhiaiyer91 (approved):
Did a final deep review of this on top of current main.

**Verified locally** (diff applied to latest main):
- RunOutput suite: 5/5 pass, including both new tests
- Full stream suite: 221 tests pass; full workflows suite: 1144 tests pass; core typecheck clean
- Confirmed the bug is still live on main: `controller.error()` in the pipeline leaves `result`/`usage` pending forever — with this PR they reject with the real error and `fullStream` consumers get a terminal failed `workflow-finish`.

**Why this is the right shape:**
- The swallowed error dates back to #9048 (`pipeTo().catch(console.log)`) — this finally surfaces it.
- `#finalizeWithError` faithfully mirrors the existing `close()` semantics (`#streamFinished` guard, late-subscriber behavior), and `resume()` resets the guard so error finalization works on resumed runs too.
- The timeout-raced `drain()` helper in the tests is exactly right for hang-class bugs — a regression fails fast instead of freezing CI.

Notes (non-blocking):
- The cancel-isolation half of the original PR already shipped via #19745; the branch was correctly reconciled and the changeset now only documents the pipeline-error fix. The PR title still mentions the cancel fix — worth updating for the merge commit, but cosmetic.
- The extra cancel test overlaps the #19745 regression test but exercises slightly different mechanics; fine to keep.

LGTM.

PR #18580, opened 2026-06-28
feat(core): add TemperatureDeprecatedHandler for deprecated sampling params

### LINKED ISSUE #16247: [BUG] claude-opus-4-7 deprecates `temperature` — agent.stream() returns 400, no framework-level handling (parity with PrefillErrorHandler)
### Describe the Bug

Anthropic's 2026 breaking change to `claude-opus-4-7` (and likely the wider 4.7 family) **deprecates the `temperature` parameter** (along with `top_p` / `top_k`). Any request that includes `temperature` returns:

```
400 — `temperature` is deprecated for this model.
```

Mastra's `agent.stream()` carries a default `temperature` (or whatever the user sets in `defaultOptions` / per-call `callSettings`) straight through to `@ai-sdk/anthropic`, which forwards it to the API verbatim. The result is a hard 400 in production with no graceful path.

This is the same shape of breaking change that motivated #13969 / #14435 (prefill removed in 4.5/4.6) — **a model in a previously-supported family quietly drops a parameter**, and the framework has no capability layer to absorb it.

#### Live failure

Production agent on `claude-opus-4-7` with `temperature: 0.3` → 400 mid-stream → run fails for the user with no automatic recovery. We currently have no way to ship 4-7 without forking Mastra or sanitising `callSettings` outside the framework.

### Steps To Reproduce

1. Create a Mastra `Agent` configured with `model: anthropic('claude-opus-4-7')` and any non-zero `temperature` (or rely on the SDK default).
2. Call `agent.stream({ messages: [{ role: 'user', content: 'hello' }] })`.
3. Anthropic returns `400 'temperature' is deprecated for this model.`
4. The error surfaces to the user; there is no `processAPIError` hook that fires for this since the `PrefillErrorHandler` regex doesn't match, and no other built-in processor strips the param.

### Expected Behavior

Pick one of these (in order of preference):

1. **Auto-strip dropped-capability params** for known models. Same shape as the auto-injected `PrefillErrorHandler` — a built-in `inputProcessor` (or call-settings sanitiser) consults a capability matrix per `provider+modelId` and removes params the model doesn't accept (`temperature`, `top_p`, `top_k`, prefill-trailing-assistant, etc.) before dispatch.
2. **Expose a public capability matrix users can extend.** e.g. `setModelCapabilities('claude-opus-4-7', { supportsTemperature: false })` so adding a new Anthropic / OpenAI / xAI release becomes a one-row catalog edit instead of a code search across every agent definition.
3. **At minimum: a built-in `TemperatureDeprecatedHandler`** sibling to `PrefillErrorHandler`, registered the same way (auto-injected via `errorProcessors`), that on the 400 strips `temperature` (and `top_p`/`top_k`) from `runState.callSettings` and signals retry.

### Why this matters / context

The ecosystem is hitting this same shape across frameworks each time Anthropic ships a breaking change:

- 4.5 / 4.6 prefill: #13969 (fixed via #14435 — auto-injected `PrefillErrorHandler`)
- 4.7 temperature: this issue
- LiteLLM has the same bug open: BerriAI/litellm#26444
- Roo-Code: RooCodeInc/Roo-Code#12162
- anything-llm: Mintplex-Labs/anything-llm#5461

Each framework currently solves it independently with its own per-model deny-list. Mastra is uniquely positioned to handle this at the `provider + processor` layer so users don't have to maintain their own `model_capabilities.ts`.

The `processAPIError` infrastructure that #14435 added is exactly the right primitive — it just needs (a) more matchers and (b) a way for the matcher to mutate `callSettings`, not just `messageList`. From the #14435 PR description, processors today can mutate `messageList` and signal `{ retry: true }`; if `processAPIError` were extended to allow mutating `callSettings` (or returning a patched options bag), a stock `TemperatureDeprecatedHandler` becomes ~20 lines.

### Reference

- #13969 — sibling issue (4.6 prefill, fixed in #14435)
- #14435 — `processAPIError` + auto-injected `PrefillErrorHandler` (the pattern to follow)
- BerriAI/litellm#26444 — same root cause in LiteLLM

### Verification

- [x] I have searched the existing issues to make sure this is not a duplicate (no issue mentions `claude-opus-4-7` + `temperature` as of filing).

abhay-codes07:
## What

Some models drop support for `temperature` (and `top_p` / `top_k`) within a previously supported family. `claude-opus-4-7` is the current example: it returns `400 temperature is deprecated for this model` when the parameter is sent. Mastra passes the default or configured temperature straight through to the provider, so the run fails mid stream with no way to recover short of forking or sanitising call settings outside the framework.

## Change

Adds `TemperatureDeprecatedHandler`, an opt-in processor that recovers from this automatically. It is the sampling parameter sibling of `PrefillErrorHandler`.

How it works:

1. `processAPIError` detects a 400 that names a sampling parameter together with a deprecated or unsupported reason, flags shared processor state, and signals a retry. It only acts on the first attempt, so it never loops.
2. On the retry, `processInputStep` reads that flag and strips `temperature`, `topP`, and `topK` from `modelSettings` before the request is dispatched. The flag stays set for the rest of the run so later steps keep stripping.

The two hooks run from different processor lists, so register the same instance in both `inputProcessors` and `errorProcessors`. They share state by processor id.

```ts
const handler = new TemperatureDeprecatedHandler()

new Agent({
  // ...
  inputProcessors: [handler],
  errorProcessors: [handler],
})
```

This needs no changes to the core loop or the processor interfaces. It only uses existing hooks.

## Tests

- Unit tests for error detection, retry signalling, the retry once guard, and the `modelSettings` stripping.
- Integration tests that drive a real agent through `generate()` and `stream()` with a mock model that rejects `temperature`, asserting the retried call has the sampling params removed and succeeds, including the `top_p` / `top_k` case.

All new tests pass, and typecheck and lint are clean. The two unrelated failing suites in `packages/core/src/processors` (agents-md autoload and tool-result-reminder) already fail on a clean checkout and are not touched by this change.

## Docs and changeset

- New reference page with a sidebar entry, plus a short mention on the processors overview.
- Minor changeset for `@mastra/core`.

Closes #16247

---

cc @TylerBarnes since this follows the `PrefillErrorHandler` pattern from #14435, and @mikhael28 who is assigned the issue. Happy to adjust the shape if you would prefer auto injection or single list registration.

## ELI5
This adds a helper that notices when a model says “I don’t accept those settings” for things like `temperature`, `top_p`, or `top_k`. If that happens, it removes those settings and tries again so requests can still succeed instead of failing.

### What changed
- Added `TemperatureDeprecatedHandler` to detect 400 errors for deprecated/unsupported sampling parameters and retry once after stripping them.
- The handler removes `temperature`, `topP`, and `topK` from `modelSettings` on the retry.
- It must be registered in both `inputProcessors` and `errorProcessors` so the retry state is shared.
- Exported the new processor from the core processors index.
- Added unit tests and integration tests covering error detection, retry behavior, and streamed/non-streamed recovery.
- Updated processor docs, added a reference page, updated the sidebar, and included a changeset.

### DIFF
--- a/.changeset/temperature-deprecated-handler.md
+++ b/.changeset/temperature-deprecated-handler.md
@@ -1,0 +1,5 @@
+---
+'@mastra/core': minor
+---
+
+Add `TemperatureDeprecatedHandler`, an opt-in processor that recovers from models which dropped support for `temperature`, `top_p`, or `top_k` (for example Anthropic's `claude-opus-4-7`, which returns a 400 when `temperature` is sent). On that rejection it strips the unsupported sampling parameters from the call settings and retries the same model once. Register the same instance in both `inputProcessors` and `errorProcessors` to enable it. Closes #16247.
--- a/docs/src/content/en/docs/agents/processors.mdx
+++ b/docs/src/content/en/docs/agents/processors.mdx
@@ -948,6 +948,8 @@ export class ContextLengthHandler implements Processor {
 
 Mastra includes a built-in [`PrefillErrorHandler`](/reference/processors/prefill-error-handler) that automatically handles the Anthropic "assistant message prefill" error. This processor is auto-injected and requires no configuration.
 
+Mastra also ships a [`TemperatureDeprecatedHandler`](/reference/processors/temperature-deprecated-handler) for models that dropped support for `temperature` / `top_p` / `top_k` (such as `claude-opus-4-7`). It strips the rejected sampling parameters and retries. Register the same instance in both `inputProcessors` and `errorProcessors` to enable it.
+
 ## Related documentation
 
 - [Guardrails](/docs/agents/guardrails): Security and validation processors
--- a/docs/src/content/en/reference/processors/temperature-deprecated-handler.mdx
+++ b/docs/src/content/en/reference/processors/temperature-deprecated-handler.mdx
@@ -1,0 +1,83 @@
+---
+title: "Reference: TemperatureDeprecatedHandler | Processors"
+description: "Documentation for the TemperatureDeprecatedHandler processor in Mastra, which recovers from models that reject deprecated sampling parameters by stripping them and retrying."
+packages:
+  - "@mastra/core"
+---
+
+# TemperatureDeprecatedHandler
+
+The `TemperatureDeprecatedHandler` recovers from models that have dropped support for sampling parameters (`temperature`, `top_p`, `top_k`) within a previously-supported family. Some models reject any request that still carries these parameters with a hard `400` (for example, Anthropic's `claude-opus-4-7` returns `400 — temperature is deprecated for this model.`).
+
+When the error is detected, the processor strips the offending sampling parameters from `modelSettings` and retries the same model once. It is the sampling-parameter sibling of [`PrefillErrorHandler`](/reference/processors/prefill-error-handler).
+
+## How it works
+
+The handler uses two hooks that run from different processor lists, so it must be registered in both `inputProcessors` and `errorProcessors`:
+
+1. The LLM API call fails with a `400` that names a sampling parameter and an "unsupported"/"deprecated" reason.
+1. `processAPIError` (which runs for error processors) checks that this is the first retry attempt, flags shared state, and returns `{ retry: true }`.
+1. On the retry, `processInputStep` (which runs for input processors) reads that flag and removes `temperature`, `topP`, and `topK` from `modelSettings` before the request is dispatched.
+1. The flag stays set for the rest of the run, so later steps keep stripping the parameters.
+
+Both hooks share the same per-processor state, so register the **same instance in both lists**.
+
+## Usage example
+
+```typescript title="src/mastra/agents/my-agent.ts"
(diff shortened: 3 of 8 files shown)


intojhanurag:
I don't think it is fixing the real cause of issue .

intojhanurag:
we will handle this by own , thanks !

abhay-codes07:
Thanks for the review and for taking this on, @intojhanurag.

A bit of context on why this PR took the shape it did, in case it helps your version: a `processAPIError` handler can detect the 400 but cannot mutate `callSettings` to drop `temperature` before the retry, so it had to register the same processor in both `inputProcessors` and `errorProcessors` to share state across the two hooks. That is opt-in and awkward, so I agree the cleaner fix is automatic recovery (auto-injection) or a proactive capability matrix so unsupported params are never sent in the first place, which is option 1 in the issue.

Happy to help on the infra side if useful, for example letting `processAPIError` return modified `callSettings` so a single error handler can do this. Could you link the PR you are working on so I can follow along and learn the preferred approach?

PR #18609, opened 2026-06-29
File-based agents: config, tools, skills, workspace seeding, and subagents alongside code-registered agents

abhiaiyer91:
## What

Adds **file-based agents** to Mastra: you can define an agent purely by directory convention under `src/mastra/agents/<name>/`, alongside agents created with `new Agent()`. Both styles register into the same Mastra instance and appear together in Studio, the server, and the bundler.

A directory becomes an agent when it contains a `config.ts` or `instructions.md`. The directory name is the agent name.

```
src/mastra/agents/weather/
  config.ts          # model + config overrides (agentConfig() for typing); optional
  instructions.md    # the agent instructions
  tools/*.ts         # default-exported tools, keyed by filename
  skills/            # flat <skill>.md, packaged <skill>/SKILL.md, or createSkill() modules
  workspace.ts       # optional Workspace override
  workspace/**       # seed files mirrored into the agent's workspace
  subagents/
    <childId>/       # a declared subagent, same layout as an agent
```

This unit of work covers the full feature:

- **Config + instructions + tools** — assemble an agent from `config.ts`, `instructions.md`, and `tools/*.ts`.
- **Skills** — flat `<skill>.md`, packaged `<skill>/SKILL.md` (with `references/`), or `createSkill()` modules, inlined into the bundle.
- **Workspace** — each file-based agent gets a default workspace; `agents/<name>/workspace/**` seed files are mirrored into the bundle and used to seed it at runtime. Override with `workspace.ts` or `config.workspace`.
- **Subagents** — declare child agents under `agents/<name>/subagents/<childId>/` using the same directory layout. Each is assembled independently and wired into the parent's `agents` map, so the loop exposes it as a delegation tool named after the directory.

## How

- **`@mastra/core`** — `assembleAgentFromFsEntry` builds an `Agent` from a portable `FsAgentEntry`, with `mergeTools` / `mergeSkills` / `mergeWorkspace` / `mergeSubAgents` precedence helpers and an `agentConfig()` typing helper. File-routed agents now run as a first-class variant through the full AIMock loop scenario battery.
- **`@mastra/deployer`** — the bundler discovers `agents/<name>/` directories, assembles them, bundles their tools/skills, mirrors workspace seeds, and registers them onto the Mastra instance during build. Subagents are discovered one level deep and emitted as nested `assembleAgentFromFsEntry` calls.
- **`mastra` (CLI)** — `mastra dev` and `mastra build` pick up file-based agents and mirror workspace seeds.

## Rules & precedence

- Code-registered agents win on name collisions, so existing projects are unaffected.
- A `config.ts` that exports `new Agent()` is used as-is; its sibling `instructions.md`, `tools/`, `skills/`, and `subagents/` are ignored with a warning.
- Subagents are **one level deep** (nested `subagents/` ignored with a warning); each subagent's `config.ts` **must** set a non-empty `description` (build error otherwise); a subagent id colliding with a parent tool key or another subagent id is a build error; `config.agents` wins over a discovered subagent of the same id.
- File-based agents are discovered by the bundler and only registered when the app runs through the Mastra CLI. Importing the `mastra` instance directly as a library does **not** discover `agents/<name>/` — register those in code instead. (Documented as a caveat.)

## Docs & example

- New docs page: `docs/agents/file-based-agents` covering config, instructions, tools, skills, workspace, subagents, precedence, and the library-import caveat.
- `examples/agent/src/mastra/agents/weather-fs/` exercises every capability: config, instructions, a tool, a flat skill, a packaged skill with references, workspace seeds, and a `forecaster` subagent.
- Changesets: `@mastra/core` (minor), `@mastra/deployer` (minor), `mastra` (minor).

## Test plan

- `pnpm build:core`
- `pnpm --filter ./packages/core exec vitest run src/agent/fs-routing src/loop/test-utils/aimock/scenarios` — 86 files / 564 tests pass; file-routed agents run as a first-class scenario variant.
- `pnpm --filter ./packages/deployer exec vitest run src/build/fs-routing` — 32 tests pass (discovery, codegen, mirror, eval).
- Manual: deployer discovery + codegen against `examples/agent/weather-fs` confirms the tool, both skills (incl. inlined `references/`), and the `forecaster` subagent are discovered and emitted, with the subagent's required `description` validated.

## ELI5
This PR teaches Mastra how to spot agents that are defined by folders and files instead of only code. That means you can drop an agent into the right directory, and Mastra will find it, bundle it, and show it in dev/build tools automatically.

### What changed
- Added filesystem-based agent discovery and assembly in core.
- Added build/deploy support so `mastra dev` and `mastra build` pick up those agents.
- Added `Mastra.__registerFsAgents(...)` and updated server/schema types to accept `source: 'fs'`.
- Added loop test coverage for the new file-routed path.
- Added docs, changesets, and a `weather-fs` example showing the full layout.

### Key behavior
- File-based agents live under `src/mastra/agents/<name>/`.
- Discovery looks for `config.ts` and/or `instructions.md`, plus optional `tools/`, `skills/`, `workspace/`, and one-level `subagents/`.
- Code-registered agents still win on name collisions.
- `config.ts` exporting `new Agent()` takes precedence over sibling files.
- Subagents need a description and can’t nest more than one level.

### Notable extras
- Workspace seed files are mirrored into the runtime bundle.
- Symlink escapes in tools/skills are guarded against.
- Docs now explain the build-time-only discovery caveat for direct library imports.


packages/core/src/agent/fs-routing/index.ts
```diff
@@ -0,0 +1,414 @@
...
+/**
+ * A single tool discovered under `agents/<name>/tools/`. `key` defaults to the
+ * filename slug; `tool` is the default export of that module.
+ */
+export interface FsAgentToolEntry {
+  key: string;
+  tool: ToolsInput[string];
+}
+
+export interface FsAgentEntry {
```
wardpeet, line 44: is there a way to use the same logic as db agents from editor? So it's fs is basically a storage mechanism?

packages/deployer/src/build/fs-routing/codegen.ts
```diff
@@ -0,0 +1,188 @@
...
+ * Emit the imports for a single discovered agent into `lines` and return the
+ * source of its `assembleAgentFromFsEntry` entry object (the `{ name, config,
+ * ... }` argument). `idPath` is a dot-free, unique path index (e.g. `0` for the
+ * first top-level agent, `0_1` for its second subagent) used to keep generated
+ * identifiers unique across the parent/child tree. `workspaceName` is the
+ * slash-joined workspace key (`<parent>/<child>` for subagents) so seed files
+ * don't collide. When `allowSubagents` is true the agent's discovered subagents
+ * are emitted as a nested `subagents: [...]` field.
+ */
+async function emitAgentEntry(
```
wardpeet, line 19: We should probably move this to babel generation instead but this works for now

packages/deployer/src/build/fs-routing/mirror.ts
```diff
@@ -0,0 +1,59 @@
...
+ * `agents/<name>/workspace/` could point outside the workspace and be preserved
+ * in the bundle, letting the agent read arbitrary files at runtime. We copy only
+ * regular files and directories.
+ */
+async function rejectSymlinks(source: string): Promise<boolean> {
+  const stats = await lstat(source);
+  return !stats.isSymbolicLink();
+}
+
+async function mirrorAgentSeeds(
```
wardpeet, line 17: We could move this to rollup so it's smart enough to rebuild on changes.

wardpeet (approved):
Overall look good, we can do some rollup/babel plugins to add auto refresh etc.

Main question if we coul.d streamline database agents and fileystem agents that would be great.

PR #18612, opened 2026-06-29
fix(core): catch async title persistence failures in Agent

### LINKED ISSUE #18617: [BUG] generated thread title persistence failure can escape as an unhandled rejection
### Problem

When Mastra auto-generates a thread title in the background, a failure while persisting that title can escape as an unhandled rejection.

The risky path is in the async generated-title flow inside `Agent`:

- `genTitle(...)` resolves successfully
- the success handler then calls `await memory.createThread(...)`
- if that write rejects, the rejection is not caught locally inside the async success callback

There is an error handler attached to the outer `.then(...)`, but that only handles rejection from `genTitle(...)` itself, not a later rejection thrown inside the async fulfillment callback.

### Why this matters

If the host app treats unhandled rejections as fatal, this can crash the process on thread-title persistence failures such as SQLite / LibSQL write errors.

This is especially relevant when `generateTitle` is enabled in memory config.

### Affected area

`@mastra/core` title-generation path in `Agent` background persistence logic.

### Steps To Reproduce

1. Enable automatic thread title generation:
   ```ts
   memory: new Memory({
     options: {
       generateTitle: true,
     },
   })

2. Use a storage backend where memory.createThread(...) can be made to fail during the background title write.
For example, mock or force the title-persistence call to reject with something like:
throw new Error('sqlite write failed')

3. Trigger an agent run that causes a title to be generated.

4. Observe that the generated-title persistence failure can surface as an unhandled rejection instead of being logged and contained.

https://github.com/nemphys/mastra-title-unhandled-rejection-repro

### Expected Behavior

If title generation succeeds but persisting the generated title fails, Mastra should catch that error locally and log it, without leaking an unhandled rejection.

### Environment Information

```block
System:
    OS: macOS 26.5.1
    CPU: (12) arm64 Apple M2 Max
    Memory: 350.80 MB / 32.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 22.23.0 - /opt/homebrew/opt/node@22/bin/node
    Yarn: 1.22.22 - /opt/homebrew/bin/yarn
    npm: 10.9.8 - /opt/homebrew/opt/node@22/bin/npm
    pnpm: 11.8.0 - /opt/homebrew/bin/pnpm
    Deno: 2.8.3 - /opt/homebrew/bin/deno
  Browsers:
    Chrome: 149.0.7827.200
    Firefox: 144.0
    Safari: 26.5
  npmPackages:
    @mastra/core: 1.42.0 => 1.42.0
    @mastra/memory: 1.20.0 => 1.20.0
  npmGlobalPackages:
    @openai/codex: 0.114.0
    @redocly/cli: 1.25.3
    @vscode/vsce: 3.7.1
    jsonlint: 1.6.3
    npm: 10.9.0
    nx: 18.3.4
    openapi2pdf: 1.0.0
    pm2: 5.3.0
```

### Verification

nemphys:
## Description

- catch errors from async generated-title persistence in `Agent`
- prevent background `memory.createThread()` failures from surfacing as unhandled rejections
- add a regression test for title persistence failure handling

`genTitle()` is kicked off in the background. If title generation succeeds but
the later `memory.createThread()` call fails, that rejection can escape the async
success handler unless it is caught locally.

Fixes #18617 

## Type of change

- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [ ] Test update

## Checklist

- [X] I have linked the related issue(s) in the description above

## ELI5
If the app makes up a nicer thread title later, it now won’t crash in the background if saving that title fails. Instead, it safely logs the problem and keeps going, and there’s a test to make sure this doesn’t happen again.

## Summary
- Wrapped async generated-title persistence in `Agent` with local `try/catch` so `memory.createThread()` failures are handled instead of becoming unhandled rejections.
- Added logging for title persistence failures with the underlying error message.
- Added a regression test covering failed title persistence, asserting no unhandled rejection occurs and the original thread title remains intact.

### DIFF
--- a/packages/core/src/agent/__tests__/title-generation.test.ts
+++ b/packages/core/src/agent/__tests__/title-generation.test.ts
@@ -1,6 +1,6 @@
 import { simulateReadableStream, MockLanguageModelV1 } from '@internal/ai-sdk-v4/test';
 import { convertArrayToReadableStream, MockLanguageModelV2 } from '@internal/ai-sdk-v5/test';
-import { beforeEach, describe, expect, it } from 'vitest';
+import { beforeEach, describe, expect, it, vi } from 'vitest';
 import { noopLogger } from '../../logger';
 import { MockMemory } from '../../memory/mock';
 import { RequestContext } from '../../request-context';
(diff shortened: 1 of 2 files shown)


Sandy-1711:
Hi @nemphys, your PR description only mentions the issue number. Please use Fixes #XXXX so GitHub can automatically link the Pull Request with the issue.

abhiaiyer91 (approved):

PR #18614, opened 2026-06-29
fix(core): prevent routing signals to dead runs in multi-pod deployments

### LINKED ISSUE #18600: sendSignal routes thread messages from per-pod in-memory state → cross-pod messages delivered into completed runs and lost
## Summary
With a distributed `PubSub` (`@mastra/redis-streams`) + the Redis `LeaseProvider`, running ≥2 pods with **no** sticky routing, `AgentThreadStreamRuntime.sendSignal` can route a new user message to a run that has **already completed**, on a pod that never owned it. The signal is published as `signal-enqueued {action: "deliver"}` to that dead runId, nothing drains it, and the message is never executed, persisted, or broadcast.

## Root cause
`sendSignal` resolves the thread's active run from the **pod-local in-memory map** `state.activeThreadRunIds`, not from the (shared) lease:

```js
// thread-stream-runtime, sendSignal
const activeRunId = state.activeThreadRunIds.get(key);
activeRecord = activeRunId ? state.threadRunsById.get(activeRunId) : undefined;
if (activeRecord && !this.#isThreadBlockingRun(state, activeRecord)) {
  state.activeThreadRunIds.delete(key);
  activeRecord = undefined;
}
if (activeRecord && activeRecord.agent.id === agent.id) {
  runId = activeRecord.runId;
} else if (activeRunId && !activeRecord) {
  runId = activeRunId;            // <-- adopts a runId with NO live record on this pod
}
```

`activeThreadRunIds` is synced across pods only via best-effort `run-registered` / `run-completed` pubsub events. On completion the entry is cleared **only on the owning pod** (`#watchThreadRunCompletion`, guarded by `activeThreadStreamIds.get(key) === record.streamId`). A pod that missed/never-processed the `run-completed` event keeps a stale entry. The `else if (activeRunId && !activeRecord)` branch then adopts that dead runId; `isActiveTarget` becomes true via `state.activeThreadRunIds.get(key) === runId`, and execution reaches the deliver path:

```js
if (key && state.activeThreadRunIds.get(key) === runId) {
  ...
  this.#publish(pubsub, key, { type: "signal-enqueued", runId, ... });
  return { signal, accepted: Promise.resolve({ action: "deliver", runId }) };
}
```

No live loop is draining that run (it completed), so the message is silently dropped. Symptom for a chat UI: user message vanishes on refresh, no agent response, other subscribers see nothing — intermittently, depending on which pod the send lands on. Sticky routing hides it by co-locating the send with the owning pod (whose map self-clears on completion).

## Intent of the branch (and why it misfires)
The code comment on `main` says the branch exists for a legitimate case:
```js
} else if (activeRunId && !activeRecord) {
  // A run can be reserved before its stream record is registered. Keep the reserved
  // id so early follow-ups still attach to the run that is starting.
  runId = activeRunId;
}
```
That intent (a run reserved **on this pod** that hasn't registered its stream record yet) is fine. The bug is that the same condition `activeRunId && !activeRecord` **also** matches a stale **cross-pod** entry for an already-completed run (synced in via an old `run-registered` event, never cleared because the `run-completed` cleanup only runs on the owning pod). The branch can't tell "reserved-and-starting-here" from "dead-and-stale-from-elsewhere," so it adopts the dead runId. Still present on `main` (1.48.x), not just 1.47.

## Expected
When the thread's "active" runId has no live local record, distinguish a locally-reserved run from a stale cross-pod one — and for the stale case consult the **shared** ownership state (the lease) rather than per-pod memory: lease held → forward to the real owner; lease free → wake a new run. The lease-guarded wake path already does exactly this.

## Workaround we're using
Only adopt the reserved id when it's **locally** reserved (`threadKeysByRunId` maps it to this thread key — set synchronously at reservation); otherwise drop the stale entry and fall through to the lease-guarded wake path:
```js
} else if (activeRunId && !activeRecord) {
  if (state.threadKeysByRunId.get(activeRunId) === key) {
    runId = activeRunId;            // reserved locally, still starting — preserve intended behavior
  } else {
    state.activeThreadRunIds.delete(key);  // stale cross-pod entry — drop; lease decides
  }
}
```
`acquireLease` in the wake path then forwards to the real owner (held) or starts a fresh run (free). Fixes the lost-message symptom on a 2–3 pod no-sticky deployment, while preserving the locally-reserved-run case the branch was written for.

## Env
@mastra/core 1.47.0, @mastra/redis-streams 0.2.0, @mastra/server 1.47.0; Redis Streams pubsub + lease; Kubernetes, no session affinity.

Rudra2637:
## Description

In multi-pod deployments using a distributed pub/sub provider (such as Redis Streams) alongside a lease provider, a pod can occasionally miss the broadcasted `run-completed` event (e.g., due to a temporary network blip or pod restart). When this happens, the pod retains a stale active run mapping in its local `state.activeThreadRunIds`.

When a subsequent user message is routed to this pod, the routing logic in `AgentThreadStreamRuntime.sendSignal` adopts the stale, dead `runId` because it cannot distinguish it from a locally starting/reserved run that has not registered its stream record yet. The signal is then enqueued to a dead run that nothing is draining, causing the user's message to be silently dropped.

This PR fixes the routing bug by verifying whether the starting run belongs to the current pod (by checking `state.threadKeysByRunId`, which is a strictly local map). If the run is not local, the stale entries are cleared from `state.activeThreadRunIds` and `state.activeThreadStreamIds`, allowing the signal to safely fall back to the lease-guarded wake path. The wake path will then either forward the signal to the actual owner pod (if the lease is held) or spin up a new run (if the lease is free).

## Related issue(s)

Fixes #18600

## Type of change

- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [ ] Test update

## Checklist

## ELI5
In a multi-pod setup, one pod might keep a “run id” around even after that run finished on another pod. If it later reuses that old run id, the signal can get sent to a completed run and effectively disappear. This PR checks whether the run id is still locally reserved for the same thread key before reusing it.

## Summary
- Fixed `AgentThreadStreamRuntime.sendSignal` to avoid reusing stale `activeRunId` state in multi-pod deployments (distributed pub/sub + lease, no sticky routing).
- When an incoming signal’s target matches an `activeRunId` but there’s no local `activeRecord`, the code now verifies `state.threadKeysByRunId.get(activeRunId) === key`.
- If the run id isn’t locally mapped to the expected thread key, stale entries are removed from `state.activeThreadRunIds` and `state.activeThreadStreamIds`, causing routing to fall back to the lease-guarded wake path (forward to the current owner if the lease is held, or start a new run if free).
- Updated `agent-signals.test.ts` to explicitly acquire/release PubSub leases for the remote tuples involved in the remote/stale routing scenarios, improving regression coverage for the stale remote active-run case.
- Added a changeset for `@mastra/core` (patch) documenting the cross-pod signal routing fix.


intojhanurag (approved):

Rudra2637:
Hey @intojhanurag ! Hope you're doing well.

I had a quick question regarding my PR. Since it's already been approved, do you have an estimate of when it might be merged?

I'm asking because I'd love to contribute more, but I'm currently limited to having only two open PRs at a time. Once this one is merged, I'll be able to pick up another issue.

No rush at all , I was just curious. Thanks!

TheIsrael1:
## 🔴 critical: Maintainer notes

**Merge confidence:** 5/5 — ready to merge

**Why:**
- The PR matches issue #18600: it prevents `sendSignal` from adopting a runId that exists only in another pod/stale per-pod `activeThreadRunIds` map when there is no local live `threadRunsById` record.
- The implementation preserves the legitimate local reservation path by checking `state.threadKeysByRunId.get(activeRunId) === key`, then clears stale local `activeThreadRunIds`/`activeThreadStreamIds` so the existing lease-based wake path can either forward to the real owner or start a fresh run.
- Follow-up commit `fd50ce2aaf` adds focused regression coverage for the stale remote active-run case, and the focused core signal test passed locally.

**Why not higher:**
- None. Relevant behavior is covered by the existing cross-runtime PubSub test plus the new stale-remote-active-run regression. Remaining Vercel failures are unrelated deployment authorization failures and are non-blocking.

**Required before merge:**
- None.

**Non-blocking follow-ups:**
- None.

**Author comments:**
- None.

**Prior maintainer notes:**
- Supersedes the previous freeform maintainer note posted in this comment.

abhiaiyer91 (approved):

PR #18622, opened 2026-06-29
feat(core): strip unsupported sampling params via capability registry

### LINKED ISSUE #16247: [BUG] claude-opus-4-7 deprecates `temperature` — agent.stream() returns 400, no framework-level handling (parity with PrefillErrorHandler)

intojhanurag:
## Description

Extends the model capability registry to include a `temperature` dimension so the model router can proactively strip unsupported sampling parameters (`temperature`, `topP`, `topK`) before dispatching requests. Models like `claude-opus-4-7` and `gpt-5-pro` that report `temperature: false` on models.dev now have these parameters silently removed, preventing 400 errors with zero user configuration.

- Generalized the attachment-only capability loader into a multi-dimension system (`CapabilityDimension`) that reads both `attachment` and `temperature` from per-provider JSON files
- Added `modelSupportsTemperature()` public lookup function following the same pattern as `modelSupportsAttachments()`
- Added `stripUnsupportedSamplingParams()` in `ModelRouterLanguageModel` that checks the registry before every `doGenerate` and `doStream` call
- Extended `ModelsDevGateway.fetchProviders()` to collect `temperature` capability data from models.dev alongside attachments
- Updated `registry-generator` and `generate-providers` script to write merged capability files with both dimensions
- Added temperature data to `anthropic.json` and `openai.json` capability files sourced from the live models.dev API

## Related Issue(s)

Fixes #16247

## Type of Change

- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [ ] Test update

## Checklist

## ELI5
Some AI models only accept certain “settings” (like how random the output should be). This PR teaches Mastra to check whether a model supports `temperature`/`topP`/`topK` and automatically removes unsupported ones so requests don’t fail with 400 errors.

## Summary
- Extended the model capability registry to support a new `temperature` capability dimension (alongside existing `attachment` support), with a new `modelSupportsTemperature()` lookup exported from core.
- Generalized the registry/capability loader/generator from attachment-only to multi-dimension capability files, merging `attachment` and `temperature` per provider.
- Updated `ModelsDevGateway.fetchProviders()` to collect temperature-capable model IDs from models.dev and include them when generating the registries.
- Updated the router (`ModelRouterLanguageModel`) to strip unsupported sampling parameters before dispatch:
  - removes `temperature` and `topP` when the target model doesn’t support temperature
  - additionally removes `topK` for streaming when unsupported
- Refreshed `anthropic.json` and `openai.json` capability lists to reflect temperature support from live models.dev.
- Added tests validating that the router strips or preserves these parameters depending on `modelSupportsTemperature()`.

## Notes / known limitations
- Capability matching won’t work for custom proxy aliases (capabilities are keyed to real model IDs).
- Support is treated as a simple on/off capability and can’t express conditional “only reject in certain scenarios” behavior.

### DIFF
--- a/.changeset/twelve-candles-love.md
+++ b/.changeset/twelve-candles-love.md
@@ -1,0 +1,15 @@
+---
+'@mastra/core': minor
+---
+
+Agents using models that dropped support for `temperature`, `topP`, or `topK` (such as `claude-opus-4-7` or `gpt-5-pro`) no longer crash with a 400 error. The model router now automatically strips unsupported sampling parameters before the request is sent — no configuration or processors needed.
+
+```ts
+const agent = new Agent({
+  model: 'anthropic/claude-opus-4-7',
+  instructions: 'You are a helpful assistant.',
+});
+
+// temperature is stripped automatically — no 400 error
+await agent.generate('hello', { modelSettings: { temperature: 0.7 } });
+```
--- a/packages/core/scripts/generate-providers.ts
+++ b/packages/core/scripts/generate-providers.ts
@@ -10,7 +10,8 @@ const __dirname = path.dirname(__filename);
 
 async function generateProviderRegistry(gateways: MastraModelGateway[]) {
   // Fetch providers from all gateways
-  const { providers, models, attachmentCapabilities, failedGateways } = await fetchProvidersFromGateways(gateways);
+  const { providers, models, attachmentCapabilities, temperatureCapabilities, failedGateways } =
+    await fetchProvidersFromGateways(gateways);
 
   if (failedGateways.length > 0) {
     console.warn(
@@ -23,12 +24,26 @@ async function generateProviderRegistry(gateways: MastraModelGateway[]) {
   const srcDir = path.join(__dirname, '..', 'src', 'llm', 'model');
   const srcJsonPath = path.join(srcDir, 'provider-registry.json');
   const srcTypesPath = path.join(srcDir, 'provider-types.generated.d.ts');
-  await writeRegistryFiles(srcJsonPath, srcTypesPath, providers, models, attachmentCapabilities);
+  await writeRegistryFiles(
+    srcJsonPath,
+    srcTypesPath,
+    providers,
+    models,
+    attachmentCapabilities,
+    temperatureCapabilities,
+  );
 
   // Write registry files to dist/ (for build output)
   const distJsonPath = path.join(__dirname, '..', 'dist', 'provider-registry.json');
   const distTypesPath = path.join(__dirname, '..', 'dist', 'llm', 'model', 'provider-types.generated.d.ts');
-  await writeRegistryFiles(distJsonPath, distTypesPath, providers, models, attachmentCapabilities);
+  await writeRegistryFiles(
+    distJsonPath,
+    distTypesPath,
+    providers,
+    models,
+    attachmentCapabilities,
+    temperatureCapabilities,
+  );
 
   // Log summary
   console.info(`\nRegistered providers:`);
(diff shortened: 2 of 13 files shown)


intojhanurag:
## Known limitations (out of scope for this PR)

**Custom proxy aliases:** If a user configures a custom model alias like `my-company/fast-claude` in LiteLLM that internally routes to `claude-opus-4-7`, the capability registry won't match it — temperature passes through. Workaround: use the real model ID instead of an alias.

**Value-conditional rules:** Some models reject temperature only under specific conditions (e.g., GPT-5 rejects `temperature != 1`, gpt-5.1 only when `reasoning_effort` is set). This PR strips all sampling params or none — it can't express conditional rules. No user has reported hitting this yet.

Both could be addressed later by adding a user-facing capability override API (e.g., `setModelCapabilities('my-alias', { temperature: false })`), built on top of the same `CapabilityDimension` system this PR introduces.

abhiaiyer91 (approved):

TylerBarnes (approved):

PR #18639, opened 2026-06-29
fix(core, client-js): preserve client-tool args when a tool-result is replayed without its call

### LINKED ISSUE #16017: [BUG] AIV5Adapter.fromModelMessage fabricates empty args on resume tool-results, poisons LLM context after 3-4 cycles

praveenrav:
## Summary

Follow-up to #16981, which fixed #16017 (empty client-tool args causing a tool-retry loop) for the case where the tool-call and tool-result are converted together. Args are still lost on the **persisted-replay** path, and the loop returns.

## Why it still loops

On the client-tool recursion path the tool-result is sent back as a standalone `role: "tool"` message. When the conversation is later rebuilt, the originating tool-call is **not in the same converted window**, so the adapter takes the "synthesize a call for this orphan result" branch and must supply `args` from somewhere:

```ts
const recoveredArgs = context.dbMessages
  ? findToolCallArgs(context.dbMessages, toolResultPart.toolCallId)
  : undefined;
const fallbackArgs =
  recoveredArgs && Object.keys(recoveredArgs).length > 0
    ? recoveredArgs
    : (toolResultPart.input ?? {});
```

Both sources come up empty on this path:

- **`findToolCallArgs(context.dbMessages, …)`** — only finds the call when it's present in `dbMessages`, but `dbMessages` isn't wired into this conversion, so it returns `{}`.
- **`toolResultPart.input`** — does not survive persistence. The stored tool-result carries `args: {}` and no `input`.

So `fallbackArgs` resolves to `{}`. Once that empty-args tool-call is persisted, the model in-context-learns the pattern and re-emits tool calls with `{}` — the #16017 loop.

Observed in a client-side-tool app: a multi-step tool sequence degraded into a runaway loop (~9 of 31 tool calls reconstructed with `{}`, worsening as the thread grew). The same flow with the original args attached directly to the result stays clean.

## Fix

Make the tool-result self-describing instead of relying on a lookup that isn't always reachable:

- **`@mastra/client-js`** — attach the original args directly to the standalone tool-result (under both `args` and `input`) so they persist with the result.
- **`AIV5Adapter`** — honor args carried on the tool-result (after `findToolCallArgs`, before `input`), treating an empty object as absent so a persisted `args: {}` can't win over a usable source.

`findToolCallArgs` is unchanged and still preferred when `dbMessages` is available.

## Test

Adds a case to `split-tool-call-args.test.ts`: a standalone tool-result carrying args, with no matching call in scope, must reconstruct the tool-call with those args rather than `{}`.

## ELI5

When a tool result is replayed without the original tool request, this PR keeps the request’s arguments attached so the tool is not called again with empty data.

## What changed

- Client tool-results now preserve original arguments in both `args` and `input` across streaming paths.
- `AIV5Adapter` recovers arguments from persisted messages or the tool-result itself, ignoring empty objects before falling back to `{}`.
- Added coverage for standalone tool-result replay and argument preservation.
- Added a changeset for `@mastra/client-js` and `@mastra/core`.


.changeset/preserve-client-tool-args-on-replay.md
```diff
@@ -0,0 +1,10 @@
+---
+'@mastra/client-js': patch
+'@mastra/core': patch
+---
+
+Preserve client-tool call args when a tool-result is replayed without its call
```
TylerBarnes, line 6: haven't read the code but this sounds like a hack to me, hiding whatever the real bug is which is dropping the call

TylerBarnes (changes requested):

.changeset/preserve-client-tool-args-on-replay.md
praveenrav: will investigate this further and get back to you - brought it out of draft by mistake

TylerBarnes: no worries and thank you! 🙌

abhiaiyer91:
Closing as superseded by #16981, which merged on June 26, 2026 and closed #16017. The merged fix preserves client-tool arguments when replaying orphan tool results and includes the corresponding client recursion and adapter regression coverage. Thanks for the contribution.

PR #18644, opened 2026-06-29
Fix tool-call narration leaking into stream.text

### LINKED ISSUE #17986: Separate "narration" text from the final answer
This issue was created from Discord post 1516182997437780181:

[Open in Discord](https://discord.com/channels/1309558646228779139/1516182997437780181)

Hi!,I use mastra supervisor with `agent.stream()` , The supervisor emits text between them ("let me check...", "now I'll do…"),

i want to stream everything to the user, but only to persist main response text (not the narrator text)

both come as text-delta, so I can't tell them apart by type.
is there a way to do that?
thanks!

labifrancis:
## Summary
- Fix `stream.text` for tool-driven multi-step agent runs without memory or output processors.
- Keep all interim narration streaming through `fullStream`, while resolving `stream.text` to the final step's answer when tool calls were involved.
- Preserve existing aggregate `stream.text` behavior for non-tool continuations, including active thread signal runs.
- Add regression coverage for a no-memory agent that narrates before a tool call and then returns a final answer.

Fixes #17986

## Why
Models can emit progress or recovery narration between tool calls, such as "let me try another data source", before producing the actual final answer. Those chunks should still be streamed live to consumers, but they should not be treated as the final resolved answer text.

This keeps the streaming behavior intact and narrows the final `stream.text` resolution behavior only for tool-driven runs where step boundaries already identify the final answer step.

## Test plan
- `pnpm build:core`
- `pnpm --filter @mastra/core check`
- `pnpm test:core -- --run src/agent/__tests__/stream.test.ts src/agent/__tests__/agent-signals.test.ts src/agent/__tests__/agent-loop-snapshot-size.test.ts --reporter=verbose`

## ELI5

When an agent calls tools, it may blurt out “thinking” messages between steps (like “let me check…”). This change makes `agent.stream()`’s `stream.text` show only the final answer, while `fullStream` still includes the narration too.

## What changed

- Fixed `stream.text` for tool-driven multi-step agent runs when there’s **no memory and no output processors**: it now resolves from the **final step’s** text (so interim narration before/around tool calls isn’t included).
- Updated `MastraModelOutput` to track whether the stream entered a **“was suspended”** state via a new internal `#wasSuspended` flag, ensuring existing behavior is preserved for suspended/resumed tool-approval flows.
- Persisted/restored the new suspension flag in `serializeState()` / `deserializeState()` so behavior stays correct across serialized stream state.
- Added regression coverage for a **no-memory** agent that narrates before a tool call and then returns a final answer.
- Added a patch changeset for `@mastra/core` documenting the `stream.text` fix.
- Verified via core build, type checks, and targeted agent tests.


intojhanurag (dismissed):
Hey @labifrancis , The sub-agent tagging is fine, but the tool-calls truncation in output.ts breaks all agents that emit text before tool calls, not just supervisors.

labifrancis (commented):
Updated the PR to remove the broad tool-call-step truncation. Final text aggregation now preserves normal pre-tool-call streamed text and only excludes delegated sub-agent streamed text when it is explicitly tagged. Added regressions for both the supervisor delegation case and a regular tool-call flow.

intojhanurag:
Hey @labifrancis , I dropped some workaround for the user . If it will not work we are gonna some sort of labelling on text-chunk or some other fix .

intojhanurag (approved):

PR #18653, opened 2026-06-29
feat(memory): add observational memory extractors

TylerBarnes:
## Summary

Adds the observational memory extractor pipeline.

Extractors let OM capture structured values alongside observations, persist selected values into thread metadata, and carry extraction failures through the OM lifecycle. Extractor instructions and schemas can be static or resolved from runtime context, so extraction can adapt to the active agent and thread.

```ts
observationalMemory: {
  enabled: true,
  observation: {
    extract: [
      new Extractor({
        slug: 'priority',
        name: 'Priority',
        schema: z.enum(['low', 'medium', 'high']),
        instructions: 'Extract the user priority when it changes.',
      }),
    ],
  },
}
```

## Test plan

- `pnpm --filter ./packages/memory test:unit -- --run src/processors/observational-memory/__tests__/extractor.test.ts src/processors/observational-memory/__tests__/observational-memory-api.test.ts --reporter=dot --bail 1`
- `pnpm --filter ./packages/memory check`
- Stack rebase verified linear against #18652

## Stack

- Previous: #18652
- Next: #18654

ELI5: This PR makes “observational memory” not only remember what happened, but also automatically pull out specific structured facts (like a user’s priority), save them on the thread, and reuse them later. It also adds an `Extractor` API so you can define how those facts are detected, validated, and what happens when extraction fails.

### Summary
- Added a public `Extractor` API for Observational Memory with:
  - **inline (XML/tag-based)** and **structured (Zod/schema-based)** extraction modes
  - slug validation, runtime-resolved/dynamic `instructions` and optional `schema`
  - per-extractor `onExtracted` hooks to normalize/transform values and record failures safely.
- Extended Observational Memory configuration with extraction pipelines:
  - `observationalMemory.observation.extract` and `observationalMemory.reflection.extract`
  - built-in extractors (current task, suggested response, thread title) composed with user extractors.
- Implemented extraction end-to-end across observer/reflector:
  - prompt generation includes extractor output sections and optional “prior extracted values”
  - parsing supports extracting values from model output while stripping extractor sections
  - structured extraction retries/fallback behavior is handled, and extracted values are validated/merged.
- Persisted extracted values into thread observational memory metadata under `threadOMMetadata.extracted`, including carry-forward of prior extracted values into subsequent observer/reflector runs.
- Propagated extraction failures through the OM lifecycle (without blocking successful extractions):
  - added `extractionFailures` alongside `extractedValues` in OM end markers and marker/telemetry payloads.
- Updated exports/types and documentation:
  - new/extended types for `extractedValues` and `extractionFailures` across OM configs, runners, and storage
  - added Observational Memory docs sections for “Extractors” plus an “Extractor API” reference.
- Added tests covering extractor logic and Observational Memory persistence across:
  - extractor unit behavior (inline/structured parsing, retries, hooks, prompt composition)
  - observational-memory API behavior (observe/reflect/buffer) and thread metadata persistence
  - marker schema changes and additional observational-memory lifecycle assertions.


docs/src/content/en/docs/memory/observational-memory.mdx
```diff
@@ -179,6 +179,116 @@ OM uses fast local token estimation for this thresholding work. Text is estimate
...
+
+export const agent = new Agent({
+  name: 'assistant',
+  instructions: 'You are a helpful assistant.',
+  model: '__GATEWAY_OPENAI_MODEL_MINI__',
+  memory,
+})
+```
+
+Extractor behavior is derived from whether you provide a schema. Schema-backed extractors run as a follow-up structured output call after the Observer or Reflector finishes. The follow-up call adds model latency and usage, but it uses the same temporary in-memory Observer or Reflector thread so providers that support prompt caching can reuse the shared prompt prefix. Schema-less extractors are inline string extractors emitted directly in the Observer or Reflector response.
```
TylerBarnes: this is overly explanatory. We can just say that
- adding a schema makes it do a follow up structured output request (don't need to talk about tradeoffs)
- it should be assumed that we're preserving prompt caching, don't need to over explain internals
- yes schema less extractors are inline in the first request

docs/src/content/en/docs/memory/observational-memory.mdx
```diff
@@ -179,6 +179,116 @@ OM uses fast local token estimation for this thresholding work. Text is estimate
...
+
+```typescript title="src/mastra/agents/agent.ts"
+new Extractor({
+  name: 'Mood',
+  instructions: 'Extract the user mood as a short phrase.',
+})
+```
+
+By default, OM includes the previous extraction in later extraction prompts. Set `includePreviousExtraction: false` for values that should only come from the current OM run.
```
TylerBarnes: wording is a bit confusing. we should say something like: when its false the observer does not see the last extracted value (or something)

docs/src/content/en/docs/memory/observational-memory.mdx
```diff
@@ -179,6 +179,116 @@ OM uses fast local token estimation for this thresholding work. Text is estimate
...
+
+```typescript title="src/mastra/agents/agent.ts"
+new Extractor({
+  name: 'Latest blocker',
+  instructions: 'Extract only the latest blocker mentioned in this conversation window.',
```
TylerBarnes: weird instructions. should be more like: extract any blockers the agent is running into

docs/src/content/en/docs/memory/observational-memory.mdx
```diff
@@ -179,6 +179,116 @@ OM uses fast local token estimation for this thresholding work. Text is estimate
...
+    observationalMemory: {
+      observation: {
+        extract: [new WorkingMemoryExtractor()],
+      },
+    },
+  },
+})
+```
+
+`WorkingMemoryExtractor` uses the normal extractor pipeline. It reads the active working memory template or schema, uses structured extraction when working memory has a JSON schema, writes updates through the memory instance, and doesn't duplicate the value under OM extracted metadata.
```
TylerBarnes: this doesn't belong here, we should add a note about working memory here and start with the sugar setting, then add a note about what that configures internally. But this should be in a working memory section.

TylerBarnes: We should probably also update the working memory page linking over here

docs/src/content/en/reference/memory/observational-memory.mdx
```diff
@@ -134,6 +134,13 @@ OM performs thresholding with fast local token estimation. Text uses `tokenx`, a
             isOptional: true,
             defaultValue: 'false',
           },
+          {
+            name: 'extract',
+            type: 'Extractor<any>[]',
```
TylerBarnes: why any type?

docs/src/content/en/reference/memory/observational-memory.mdx
```diff
@@ -267,6 +274,13 @@ OM performs thresholding with fast local token estimation. Text uses `tokenx`, a
               "Custom instruction appended to the Reflector's system prompt. Use this to customize how the Reflector consolidates observations, such as prioritizing certain types of information.",
             isOptional: true,
           },
+          {
+            name: 'extract',
+            type: 'Extractor<any>[]',
```
TylerBarnes: same

TylerBarnes: also this section is duplicated?

docs/src/content/en/reference/memory/observational-memory.mdx
```diff
@@ -351,8 +365,116 @@ OM persists token payload estimates so repeated counting can reuse prior token e
...
+```typescript title="src/mastra/agents/agent.ts"
+import { Memory, WorkingMemoryExtractor } from '@mastra/memory'
+
+const memory = new Memory({
+  options: {
+    workingMemory: {
+      enabled: true,
+      injectTools: false,
+    },
+    observationalMemory: {
```
TylerBarnes: this is out of date, and not sure it even belongs on this page at all?

packages/memory/src/processors/observational-memory/temporary-memory.ts
```diff
@@ -0,0 +1,32 @@
...
+import type { MastraMemory } from '@mastra/core/memory';
+import { InMemoryStore } from '@mastra/core/storage';
+
+export interface TemporaryOmMemoryContext {
+  memory: MastraMemory;
+  options: AgentMemoryOption;
+}
+
+export async function createTemporaryOmMemoryContext(prefix: string): Promise<TemporaryOmMemoryContext> {
+  const { Memory } = await import('../../index');
```
TylerBarnes: why a dynamic import?

abhiaiyer91 (approved):

PR #18664, opened 2026-06-30
fix(mcp): lazily call createRequire to prevent Cloudflare Workers crash

### LINKED ISSUE #18659: [BUG] @mastra/mcp 1.12 crashes Cloudflare Workers at module init — createRequire(import.meta.url) for dd-trace tracer
### What version of `@mastra/mcp` are you using?

1.12.0 (latest). Regression introduced since 1.6.0.

### What environment are you running on?

Cloudflare Workers (`workerd`) — both `wrangler dev` and deployed. `nodejs_compat` enabled.

### Describe the bug

`@mastra/mcp@1.12.0` crashes at **module initialization** on Cloudflare Workers. `src/client/client.ts` has an unconditional top-level call:

```js
// @mastra/mcp/dist/index.js (src/client/client.ts)
var require2 = createRequire(import.meta.url);
```

On `workerd`, `import.meta.url` is `undefined` in the bundled worker module (this is a known, long-standing Workers behavior — `nodejs_compat` provides `createRequire` but does **not** populate `import.meta.url`). So `createRequire(undefined)` throws at the top level, before any code runs:

```
✘ [ERROR] service core:user:...: Uncaught TypeError: The argument 'path'
  must be a file URL object, a file URL string, or an absolute path string.
  Received 'undefined'
    at createRequire (node:module)
    at <module init> (index.js)   // var require2 = createRequire(import.meta.url)
✘ [ERROR] The Workers runtime failed to start.
```

Because this runs at module init, **any Worker that statically imports `MCPClient` fails to boot** — the whole worker is down, not just MCP calls.

### Root cause / why it's only needed for Datadog

`require2` is used solely to lazy-load the optional Datadog tracer:

```js
function loadDatadogTracer() {
  if (!isDatadogTracerLikelyLoaded()) return null;   // false on Workers
  try { return require2("dd-trace"); } catch { return null; }
}
```

The `dd-trace` load is already fully guarded (it never loads on Workers — `isDatadogTracerLikelyLoaded()` returns false). But the **`createRequire(import.meta.url)` that backs it is hoisted to module scope and runs unconditionally**, so it crashes the worker even though `dd-trace` is never used there.

This is new in the 1.x line: `@mastra/mcp@1.6.0` has no `createRequire` call and boots fine on Workers; the regression came in with the Datadog tracer bridge.

### Minimal reproduction

```ts
// src/index.ts (a Worker with compatibility_flags = ["nodejs_compat"])
import { MCPClient } from "@mastra/mcp"; // value import is enough
export default { fetch() { return new Response("ok"); } };
```

```bash
wrangler dev
# -> Uncaught TypeError ... createRequire ... Received 'undefined'
# -> The Workers runtime failed to start.
```

### Suggested fix

Make `require2` lazy so it's only created when Datadog is actually loaded (which never happens on Workers), e.g. inside `loadDatadogTracer()` after the `isDatadogTracerLikelyLoaded()` guard:

```js
function loadDatadogTracer() {
  if (!isDatadogTracerLikelyLoaded()) return null;
  try {
    const require2 = createRequire(import.meta.url || "file:///");
    return require2("dd-trace");
  } catch { return null; }
}
```

Or, minimally, the same `|| "file:///"` fallback Mastra already uses elsewhere for the hoisted call. This mirrors the lazy-init fixes already applied to other packages for the identical Workers module-init class:

- `@mastra/observability` — `fileURLToPath` at module init broke Workers (#12536), fixed by lazy init (#12540).
- `@mastra/schema-compat` — `createRequire` broke browser/Vite builds (#14440), fixed in #14221 / #14443.

Related open Workers/`workerd` reports: #14503 (AJV `new Function` in core 1.11+), #15926 (schema-compat `JsonSchemaWrapper`).

Happy to send a PR if helpful.

intojhanurag:
## Description

`@mastra/mcp` crashes at module initialization on Cloudflare Workers because `createRequire(import.meta.url)` runs unconditionally at the top level of `client.ts`, and `import.meta.url` is `undefined` on `workerd`. This prevents any Worker that imports `MCPClient` from booting.

- Moved `createRequire` from module scope into `loadDatadogTracer()`, after the `isDatadogTracerLikelyLoaded()` guard, so it only runs when Datadog is actually present
- Removed the `require.resolve`/`require.cache` check from `isDatadogTracerLikelyLoaded()` since it depended on the module-scope `require` and is redundant (`loadDatadogTracer` already has a try/catch)

## Related Issue(s)

Fixes #18659

## Type of Change

- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [ ] Test update

## Checklist

## ELI5
This PR prevents `@mastra/mcp` from doing a “Node-only setup” when it loads on Cloudflare Workers. Instead, it waits and only sets that up when Datadog tracing is actually needed, so importing `MCPClient` won’t crash the worker.

## Summary
- Fixes a Cloudflare Workers startup crash caused by an unconditional top-level `createRequire(import.meta.url)` in `packages/mcp/src/client/client.ts`.
- Makes `createRequire` lazy by creating it inside `loadDatadogTracer()` only after the Datadog loading guard passes.
- Simplifies Datadog detection by removing the `require.resolve` / `require.cache`-based check from `isDatadogTracerLikelyLoaded()`, since the guarded `try/catch` in `loadDatadogTracer()` already handles the safe probing.
- Updates the changeset to publish a patch release noting MCP no longer crashes Workers at module initialization.


packages/mcp/src/client/client.ts
```diff
@@ -138,12 +138,7 @@ function isDatadogTracerLikelyLoaded(): boolean {
     return true;
   }
 
-  try {
-    const resolvedPath = require.resolve('dd-trace');
-    return Boolean(require.cache[resolvedPath]);
-  } catch {
-    return false;
-  }
+  return false;
```
epinzur: pls restore this, with a fix as well. IE:

```
  // Detect dd-trace loaded programmatically (e.g. an entrypoint that imports
  // dd-trace and calls tracer.init()), which the execArgv/NODE_OPTIONS checks
  // above do not catch. createRequire is created lazily inside the try/catch so
  // this stays safe on runtimes where import.meta.url is undefined (e.g.
  // Cloudflare Workers / workerd), where createRequire throws and we fall
  // through to returning false rather than crashing.
  try {
    const req = createRequire(import.meta.url);
    const resolvedPath = req.resolve('dd-trace');
    return Boolean(req.cache[resolvedPath]);
  } catch {
    return false;
  }
```

intojhanurag: Yes done , thx !

epinzur (approved):

PR #18680, opened 2026-06-30
feat(mastracode): programmatic headless API (runMC)

abhiaiyer91:
## Summary

Turns Mastra Code's headless mode into a real **programmatic API** that works equally well from CI/Node code and from the command line. The CLI is now a thin adapter over the same runner.

Previously, headless mode was a single ~650-line file that fused argv/stdin parsing, `process.exit`, output formatting, and the run loop together, and returned only an exit code. It was unusable from code and hard to extend.

## What's new

- **`runMC({ controller, session, prompt, ... })`** — a pure async runner. It returns a handle that streams live events as an async-iterable **and** resolves to a typed `RunMCResult` (status, text, usage, tool calls, exit code). It never calls `process.exit` or writes to global streams.

  ```ts
  import { createMastraCode, runMC } from 'mastracode';

  const { controller, session } = await createMastraCode({ settingsPath });
  const run = runMC({ controller, session, prompt: 'Fix the bug' });
  for await (const event of run) { /* optional: live progress */ }
  const result = await run.result; // typed RunMCResult
  ```

- **`runMCCli`** — the CLI is now a thin adapter over `createMastraCode` + `runMC`, owning argv/stdin, output formatters, exit-code mapping, and teardown.
- **Pluggable `ResolutionPolicy`** — `autoApprovePolicy` (default, preserves prior behavior), `denyPolicy`, and `permissionModeToPolicy`, selectable via a new `--permission-mode {auto|deny}` flag.
- **`--max-turns`** flag (and `maxTurns` option) caps assistant turns and reports a `max_turns` status with exit code 1 when hit mid-task.
- **Pure output formatters** (human / json / jsonl) decoupled from sinks.
- Public API exported from `mastracode` and the new `mastracode/headless` subpath.

## Breaking changes

- The `--format` and `--output-format` flags are replaced by a single `--output` flag with values `human`, `json`, or `jsonl`.
- The old `headless.ts` module is removed in favor of the `mastracode/headless` surface.

Changeset is marked `minor` (mastracode is free to evolve/break).

## Docs

Audited every Mastra Code doc against the source and fixed drift: the slash-command table now matches the command registry exactly, the reference Returns/options tables match `createMastraCode`, fake API examples (`selectOrCreateThread`, `controller.subscribe/sendMessage`) were corrected to the real session-based API, and a new headless page documents the programmatic API.

## Test plan

- `pnpm --filter ./mastracode check` (typecheck) — clean
- Headless suite (`runMC`, policy, format, cli, scenarios) — passing
- Manual CLI smoke: piped stdin run, `--output json`/`jsonl`, `--max-turns`, `--permission-mode deny`, exit-code mapping (0/1/2), `--help`
- Manual programmatic smoke: `createMastraCode` → `runMC`, iterate events + await result, confirming no `process.exit`

## ELI5
This PR turns Mastra Code’s “headless mode” into a reusable library you can call from your own code or run from the command line. That way CI/scripts can stream progress and control behavior (like unattended tool approvals) without relying on `process.exit` or messing with global stdin/stdout.

## What changed
- Added a programmatic headless runner: `runMC`
  - Streams live events as an async-iterable and resolves to a typed `RunMCResult` (including `exitCode`)
  - Avoids `process.exit` and avoids writing to global stdin/stdout directly
  - Supports run states including `completed`, `error`, `aborted`, `timeout`, and `max_turns`
- Refactored the CLI into a thin adapter over the runner
  - New CLI entrypoint: `runMCCli`
  - `mastracode/src/main.ts` now routes headless execution to `runMCCli`
- Added configurable unattended tool resolution policies
  - `autoApprovePolicy`, `denyPolicy`, and `permissionModeToPolicy`
  - New flag: `--permission-mode {auto|deny}`
- Added execution limiting
  - New flag: `--max-turns` (and `maxTurns` option) which stops the run and returns `max_turns`
- Replaced output flags and made output formatting pure
  - Replaced `--format` / `--output-format` with a single `--output` option: `human`, `json`, `jsonl`
  - Added pure helpers: `formatHuman`, `formatJsonl`, `renderTextResult`, `renderJsonResult`
- Updated public API/exports
  - Public exports added from both `mastracode` and `mastracode/headless`
  - Added `mastracode` subpath export: `mastracode/headless`
- Removed the old headless module
  - Deleted `mastracode/src/headless.ts` and replaced it with a new `mastracode/src/headless/*` structure (runner/cli/format/policy/types)

## Docs & other changes
- Added a new documentation page for headless mode (`docs/src/mastra-code/headless.mdx`) and updated related Mastra Code docs to match the new flags/API
- Added a Vercel redirect for `/docs/mastra-code/headless`


mastracode/src/headless/cli.ts
```diff
@@ -0,0 +1,351 @@
...
+  } catch {
+    return false;
+  }
+}
+
+/**
+ * Parse CLI arguments for headless mode. Output is controlled by a single
+ * `--output <human|json|jsonl>` flag (default `human`).
+ */
+export function parseHeadlessArgs(argv: string[]): HeadlessArgs {
```
wardpeet, line 74: I wonder if we should use https://www.npmjs.com/package/citty for parsing etc like we do in create-mastra

mastracode/src/headless/cli.ts
```diff
@@ -0,0 +1,351 @@
...
+}
+
+export function printHeadlessUsage(): void {
+  process.stdout.write(`
+Usage: mastracode --prompt <text> [options]
+
+Headless (non-interactive) mode options:
+  --prompt, -p <text>           The task to execute (required, or pipe via stdin)
+  --continue, -c                Resume the most recent thread instead of creating a new one
+  --thread, -t <id|title>       Resume a specific thread by ID or title
```
wardpeet: Should we search on title? I think id is sufficient

wardpeet (approved):
LGTM, small nits

mastracode/src/headless/cli.ts
abhiaiyer91: Good thought. I checked and create-mastra actually uses `commander`, not citty, so there isn't an existing citty convention to align with here. For this PR I kept `node:util`'s `parseArgs` since it covers the headless flag set with zero new deps and the parsing is isolated to `cli.ts`. Happy to do a follow-up that standardizes the CLIs on one parser (citty or commander) repo-wide if we want consistency — felt out of scope for the headless overhaul.

mastracode/src/headless/cli.ts
abhiaiyer91: Agreed — titles aren't unique (the old code had to sort by `updatedAt` to break ties), so ID is the right contract. Dropped title resolution: `resolveThread` now matches on exact ID only, and updated the help text, `RunMCThreadOptions.id` doc, and the headless docs table accordingly. Done in the latest commit.

mastracode/src/headless/cli.ts
abhiaiyer91: Took another pass at this — you're right that the parsing was clunky regardless of citty. I refactored it into a declarative flag table (`flags.ts`): each flag carries its own type, coercion/validation, and help text, and the `parseArgs` config, validation loop, and `--help` usage block now all derive from that one table instead of three hand-maintained spots.

Concretely: enum/positive-int validators are defined once and reused, error messages are uniform (`--mode must be one of: build, plan, fast`), `--help` is now self-documented (it was previously missing from the usage), and adding a flag is a one-row change.

I kept `node:util parseArgs` rather than pulling in citty — create-mastra uses commander, so there isn't an existing citty convention to match, and this removes the boilerplate without a new dep. Happy to standardize on a parser library repo-wide as a follow-up if you'd prefer. Pushed in 31d59fcf32.

PR #18681, opened 2026-06-30
feat(playground): agent capabilities footer + memory panel at bottom

damien-schneider:
## What

First slice split out of the larger `feat/agent-version-editor-chat` branch so it reviews in isolation. This PR is purely the **agent sidebar UI refresh** — no version-editor backbone.

- **New `AgentCapabilitiesFooter`** — a collapsible row at the bottom of the agent sidebar that summarizes which features are active for the agent (Memory, Editor, Tools, Workflows, Sub-agents, Processors). Collapsed it shows a chip per capability plus an enabled/total counter; expanded it reveals per-capability detail rows linking to the relevant docs.
- **Memory panel docked at the bottom** — the Memory card moved from the top to the bottom of the sidebar (chevron direction flipped accordingly) so the thread list stays the primary view.
- **Empty state moved into the thread layer** — "Memory not enabled" now renders inside the thread layer instead of replacing the whole panel, so the capabilities footer is always visible even when the agent has no memory.
- **`isMemoryLoading` plumbing** — threaded through `AgentSidebar` → `MemorySidebar` so the footer can show a "Checking" state without flashing the empty state.
- **`useAgentVersions({ enabled })`** — optional gate so the footer only looks up versions when the editor is actually available.

## Scope notes

- All deps the footer relies on (`useAgent`, `useIsCmsAvailable`, `agent.editor`, `GetAgentResponse.tools/workflows/agents/processors`, the `Collapsible` primitive) already exist on `main`; this PR adds no new backbone.
- `memory-sidebar.tsx` was rewritten by hand (not cherry-picked) to keep only the footer + panel-repositioning, dropping the version-editor wiring that lives in the follow-up PR.

## Tests

- `memory-sidebar.test.tsx` (MSW + typed `@mastra/client-js` fixtures): registers the footer's agent/system-packages endpoints and adds a case asserting the footer renders collapsed and expands to reveal capability detail rows. **12/12 pass.**
- Lint clean on all changed files.

## Heads-up

`pnpm --filter ./packages/playground typecheck` currently reports 3 pre-existing errors in `agent-playground/index.tsx` and `cms/agents/edit-layout.tsx` (`activeVersionId: string | null` vs a `string | undefined` prop). These reproduce on a clean `main` checkout and are untouched by this PR — they're in dead/soon-removed pages handled by the version-editor branch.

## ELI5
This PR adds a small “what can this agent do?” footer to the bottom of the agent sidebar, so you can quickly see capabilities and open more details. It also makes the Memory area feel smoother while it loads by showing a “Checking” state and keeping the footer visible even when Memory isn’t enabled.

## Summary
- Added a collapsible **AgentCapabilitiesFooter** at the bottom of the agent sidebar:
  - **Collapsed:** shows capability chips plus an **enabled/total** counter.
  - **Expanded:** shows per-capability detail rows for **Memory, Editor, Tools, Workflows, Sub-agents, and Processors**, including **status text** (e.g., **“Checking”**) and **documentation links**.
- Reworked **MemorySidebar** layout/behavior so the Memory panel **docks at the bottom** (keeping the thread list the primary view):
  - Updated overlay/visibility behavior and **chevron direction**.
  - Changed the “Memory not enabled” empty state to render **inside the thread layer**, so the capabilities footer remains visible.
- Threaded loading state through the sidebar:
  - `AgentSidebar` now passes `isMemoryLoading` → `MemorySidebar`.
  - `MemorySidebar` treats `isMemoryLoading || hasMemory` as “panel present” to avoid UI flashing.
  - `AgentCapabilitiesFooter` uses `isMemoryLoading` to show **Checking** for the Memory capability.
- Improved data fetching for editor-related capability details:
  - `useAgentVersions` now accepts an optional `enabled` flag so version lookup only runs when the editor is actually available (and tests verify the skipped calls when CMS/editor is unavailable or locked).
- Updated tests and docs-related behavior:
  - Added comprehensive tests for **AgentCapabilitiesFooter** (collapsed/expanded behavior, CMS availability, editor locked/disabled, memory states, capability counting, and doc link hrefs).
  - Updated memory sidebar tests to mock agent/CMS/version endpoints needed for the footer.
  - Updated memory-disabled CTA link to the new `/docs/memory/overview` URL.
- Studio preview support:
  - Added `@mastra/editor` dependency (pinned via overrides), registered the `MastraEditor`, and introduced an `editorShowcaseAgent` to surface the **Editor** capability.
- Added a React best-practices rule to discourage nested ternaries:
  - New `structure-no-nested-ternary` rule with an example of guard-clause refactoring, plus catalog/index updates and count adjustments.


packages/playground/src/domains/agents/components/memory-sidebar/agent-capabilities-footer.tsx
```diff
@@ -0,0 +1,260 @@
...
+    {
+      id: 'processors',
+      label: 'Processors',
+      status: isAgentCapabilitiesLoading ? 'Checking' : countStatus(processorsCount),
+      description:
+        processorsCount > 0
+          ? `${processorsCount} input/output processor${processorsCount === 1 ? '' : 's'} configured.`
+          : 'No input or output processors are configured.',
+      docsHref: 'https://mastra.ai/docs/agents/processors',
+      enabled: processorsCount > 0,
```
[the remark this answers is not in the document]
damien-schneider: Fixed in 7ef22df55f. Pointed the Memory capability `docsHref` (and the matching "memory not enabled" empty-state CTA) at the canonical `https://mastra.ai/docs/memory/overview`, dropping the retired `/en/docs/agents/agent-memory` path so it lines up with the other capability links. Added a regression assertion on the link target in the new footer test.

packages/playground/src/domains/agents/components/memory-sidebar/__tests__/agent-capabilities-footer.test.tsx
```diff
@@ -0,0 +1,137 @@
...
+    registerFooterHandlers({ cmsEnabled: false });
+    renderFooter();
+
+    const footer = await screen.findByTestId('agent-capabilities-footer');
+    fireEvent.click(footer);
+
+    // Editor reads "Off" and the gated versions query stays put.
+    expect(await screen.findByRole('link', { name: 'Editor: Off' })).not.toBeNull();
+    await new Promise(resolve => setTimeout(resolve, 50));
+    expect(onVersions).not.toHaveBeenCalled();
```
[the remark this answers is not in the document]
damien-schneider: Done in 1b87606806 — dropped the `setTimeout(50)` from both negative-fetch assertions. Once `Editor: Off`/`Editor: Locked` renders, `editorEnabled` is false so the versions query is permanently disabled; asserting `onVersions` was not called right after the row renders is sufficient.

.claude/skills/react-best-practices/references/rules/structure-early-return-render-branches.md
```diff
@@ -0,0 +1,62 @@
...
+  <Panel>
+    <List items={items} />
+  </Panel>
+);
+```
+
+**Correct (one wrapper; the body early-returns bare content):**
+
+```tsx
+function PanelBody(props) {
```
mfrachet: there's something off here, it feels isLoading should be passed to "props" and that's actually what we DONT want

mfrachet: The issue is not about early breaking, but more about limiting spreading down props to defer responsibility to the children.

If a parent fetches data receives isLoading, it must handle that specific case.

Same goes for error.

packages/playground/src/domains/agents/components/memory-sidebar/agent-capabilities-footer.tsx
```diff
@@ -0,0 +1,263 @@
...
+  },
+  cyan: {
+    icon: 'text-cyan-700 dark:text-cyan-300',
+  },
+  orange: {
+    icon: 'text-orange-700 dark:text-orange-300',
+  },
+};
+
+const getCount = (value?: Record<string, unknown> | unknown[]) => {
```
mfrachet: how could it be an array of unknown?

packages/playground/src/domains/agents/components/memory-sidebar/agent-capabilities-footer.tsx
```diff
@@ -0,0 +1,263 @@
...
+
+const countStatus = (count: number) => (count > 0 ? String(count) : 'Off');
+
+function getMemoryStatus(isLoading: boolean, hasMemory: boolean, memoryType?: 'local' | 'gateway'): string {
+  if (isLoading) return 'Checking';
+  if (!hasMemory) return 'Off';
+  return memoryType === 'gateway' ? 'Gateway' : 'On';
+}
+
+function getEditorStatus(isLoading: boolean, enabled: boolean, locked: boolean, versionCount: number): string {
```
mfrachet: Feels like it should be a component

packages/playground/src/domains/agents/components/memory-sidebar/agent-capabilities-footer.tsx
```diff
@@ -0,0 +1,263 @@
...
+  if (isLoading) return 'Checking';
+  if (enabled) return versionCount > 0 ? String(versionCount) : 'Ready';
+  return locked ? 'Locked' : 'Off';
+}
+
+function getEditorDescription(enabled: boolean, locked: boolean): string {
```
mfrachet: same here => component

packages/playground/src/domains/agents/components/memory-sidebar/agent-capabilities-footer.tsx
```diff
@@ -0,0 +1,263 @@
...
+
+  const versionCount = versionsQuery.data?.total ?? versionsQuery.data?.versions.length ?? 0;
+  const toolsCount = getCount(agent?.tools);
+  const workflowsCount = getCount(agent?.workflows);
+  const subAgentsCount = getCount(agent?.agents);
+  const processorsCount = (agent?.inputProcessors?.length ?? 0) + (agent?.outputProcessors?.length ?? 0);
+  const isEditorLoading = isAgentLoading || isCmsAvailabilityLoading || (editorEnabled && versionsQuery.isLoading);
+  const isAgentCapabilitiesLoading = isAgentLoading && !agent;
+
+  const capabilities: Capability[] = [
```
mfrachet: I hate config objects lol 🫠

packages/playground/src/domains/agents/components/memory-sidebar/memory-sidebar.tsx
```diff
@@ -71,37 +72,44 @@ function ConfigBadge({ icon: Icon, label, tooltip, enabled, value, expanded }: C
...
+// clauses and returns bare content — see structure-early-return-render-branches.
+export function MemorySidebar(props: MemorySidebarProps) {
+  return (
+    <SidebarPanel>
+      <MemorySidebarBody {...props} />
+    </SidebarPanel>
+  );
+}
+
+function MemorySidebarBody({ agentId, threadId, threads, isLoading, onDelete }: MemorySidebarProps) {
```
mfrachet: isLoading shouldn't be passed down

packages/playground/src/domains/agents/components/memory-sidebar/use-memory-feature-flags.ts
```diff
@@ -0,0 +1,33 @@
+import { useMemoryConfig } from '@/domains/memory/hooks';
+
+// A memory feature may be configured as a plain boolean or as an object that
+// carries `{ enabled }` (plus extra options), so normalize both shapes.
+type ToggleConfig = boolean | { enabled?: boolean } | null | undefined;
```
mfrachet: wee should make it simpler to avoid weird type checking

packages/playground/src/domains/agents/components/memory-sidebar/use-memory-feature-flags.ts
```diff
@@ -0,0 +1,33 @@
...
+
+function isEnabled(value: ToggleConfig): boolean {
+  return typeof value === 'object' ? Boolean(value?.enabled) : Boolean(value);
```
mfrachet: (like this weird check)

packages/playground/src/domains/agents/components/memory-sidebar/use-memory-feature-flags.ts
```diff
@@ -0,0 +1,33 @@
...
+/**
+ * Reads the agent's memory config and reduces its loosely-typed feature settings
+ * to the on/off flags the sidebar renders, keeping that parsing out of the view.
+ */
+export function useMemoryFeatureFlags(agentId: string): MemoryFeatureFlags {
+  const { data: memoryConfig } = useMemoryConfig(agentId);
+  const config = memoryConfig?.config;
+
+  return {
+    lastMessages: typeof config?.lastMessages === 'number' ? config.lastMessages : undefined,
```
mfrachet: 🫠

packages/playground/src/domains/agents/hooks/use-agent-versions.ts
```diff
@@ -16,14 +16,22 @@ export type { ListAgentVersionsParams, CreateAgentVersionParams };
...
+  enabled?: boolean;
+}) => {
   const client = useMastraClient();
   const { requestContext } = usePlaygroundStore();
 
   return useQuery<ListAgentVersionsResponse>({
     queryKey: ['agent-versions', agentId, params, requestContext],
     queryFn: () => client.getStoredAgent(agentId).listVersions(params, requestContext),
-    enabled: !!agentId,
+    enabled: enabled && !!agentId,
```
mfrachet: enabled should be enough, we can pass enabled: !!agent on the caller side. let's not make implem details compelx when not needed

mfrachet (changes requested):

packages/playground/src/domains/agents/hooks/use-agent-versions.ts
damien-schneider: The issue is more on the agentId not well typed in the prop, so I've in addition to improve this added a line on an existing skill for it, I will probably do a follow up PR to clean already all the similar code smell elswhere in the repo.

packages/playground/src/domains/agents/components/memory-sidebar/memory-sidebar.tsx
damien-schneider: Addressed, I let you review the changes in the skill, I've included a hook use in the skill to better explicit but maybe there is a way to make it a bit more concise. Opened to any improvements of course

.claude/skills/react-best-practices/references/rules/structure-early-return-render-branches.md
damien-schneider: Let me know what you think about the new cersion including a hook in it to better explain

packages/playground/src/domains/agents/components/memory-sidebar/agent-capabilities-footer.tsx
damien-schneider: Addressed with a type and a helper typed from GetAgentResponse tools/workflows/agents

packages/playground/src/domains/agents/components/memory-sidebar/agent-capabilities-footer.tsx
damien-schneider: Kept this as a helper because it only derives reusable status text, not JSX, but open to see how you would handle it if you still think that it should be a component

packages/playground/src/domains/agents/components/memory-sidebar/agent-capabilities-footer.tsx
damien-schneider: same here

packages/playground/src/domains/agents/components/memory-sidebar/agent-capabilities-footer.tsx
damien-schneider: Not sure to see how you would build it cleanly without using a config object, feel free to suggest because i'm not sure to see the problem here 😅

mfrachet: `<Component label={} id={} etc.../>` instead of mapping on a predefined shape that we remap again on the component shape

.claude/skills/react-best-practices/references/rules/client-request-dedupe.md
```diff
@@ -62,4 +62,55 @@ function UpdateButton() {
 }
 ```
 
-Reference: [https://tanstack.com/query](https://tanstack.com/query)
+**For dependent params in custom hooks:**
+
+If callers often pass `id ?? ''` with `enabled: Boolean(id)`, the hook input should be `id?: string | null`. Callers should pass the real value, not a fake empty string:
```
mfrachet: don't need |null

.claude/skills/react-best-practices/references/rules/client-request-dedupe.md
```diff
@@ -62,4 +62,55 @@ function UpdateButton() {
...
+  });
+}
+```
+
+**Manual refetch-compatible query:**
+
+```tsx
+import { useQuery } from '@tanstack/react-query';
+
+function useProject(projectId?: string | null, enabled = true) {
```
mfrachet: for me, this is a codesmell. if projectId is not defined, we should be able to "useProject" at all.

packages/playground/src/domains/agents/components/memory-sidebar/agent-capabilities-footer.tsx
mfrachet: isLoading + enabled makes me feel it's not a simple helper.Up to you thogh

mfrachet (approved):

PR #18688, opened 2026-06-30
fix(schema-compat): gate v4 nullable/optional handlers on inner type

### LINKED ISSUE #18687: [BUG] schema-compat v4 nullable/optional handlers gate on wrapper type instead of inner type
### Describe the Bug

In `@mastra/schema-compat`, the Zod v4 `defaultZodNullableHandler` (`schema-compatibility-v4.ts`) gates on `value.constructor.name`, which for a `ZodNullable` is always the string `"ZodNullable"` rather than the wrapped inner type. The handler is supposed to process the inner type only when that inner type is in `handleTypes` (default `SUPPORTED_ZOD_TYPES`) and otherwise pass the schema through. Because `"ZodNullable"` is itself in `SUPPORTED_ZOD_TYPES`, the gate is effectively always true, so the inner type is always processed regardless of what it is.

The Zod v3 handler does this correctly by checking the inner type name (`value._def.innerType._def.typeName`), and the dispatcher's own optional handler in `schema-compatibility.ts` computes the inner v4 type name and converts it to the v3-style name before checking. The v4 nullable/optional handlers diverge.

The visible effect: a nullable wrapping an unsupported inner type (for example `ZodTuple`, which is in `UNSUPPORTED_ZOD_TYPES`) is processed instead of passed through, so providers that reject unsupported types throw on a schema that the v3 path accepts.

### Steps To Reproduce

```ts
const layer = new GoogleSchemaCompatLayer({ provider: 'google', modelId: 'gemini-1.5-pro', supportsStructuredOutputs: true })
layer.defaultZodNullableHandler(z.tuple([z.string()]).nullable())
```

### Expected Behavior

The nullable tuple is passed through unchanged (its inner `ZodTuple` is unsupported), matching the v3 handler. No error.

### Actual Behavior

Under Zod v4 it throws `gemini-1.5-pro does not support zod type: ZodTuple`, because the inner tuple is processed even though it should have been skipped.

### Additional Context

`defaultZodOptionalHandler` in the same file has the identical `value.constructor.name` defect. Fix is to gate on the inner type's name (mirroring the dispatcher). I have a PR ready with unit tests that run under both the v3 and v4 test projects.

abhay-codes07:
## What

The Zod v4 `defaultZodNullableHandler` (and `defaultZodOptionalHandler`) gate on `value.constructor.name`, which for a `ZodNullable` is always `"ZodNullable"` rather than the wrapped inner type. They are meant to process the inner type only when it is in `handleTypes` (default `SUPPORTED_ZOD_TYPES`) and otherwise pass through. Since `"ZodNullable"` is itself in `SUPPORTED_ZOD_TYPES`, the gate is always true and the inner type is always processed.

The v3 handler checks the inner type name correctly, and the dispatcher in `schema-compatibility.ts` already computes the inner v4 type name for its optional path. The visible effect: a nullable wrapping an unsupported inner type (e.g. a tuple, which is in `UNSUPPORTED_ZOD_TYPES`) is processed and rejected under v4, while v3 passes it through.

## Change

Gate on the inner type's name (normalized to the v3-style `Zod*` name), mirroring the dispatcher, in both the nullable and optional v4 handlers.

## Tests

Added tests (a nullable tuple passes through; a nullable string is still processed) that run under both the v3 and v4 test projects. The tuple case fails on the current v4 code and passes after the fix. The full schema-compat suite (32 files) and typecheck pass.

Closes #18687

---

Hey @abhiaiyer91, you merged my last couple of schema-compat fixes so flagging you on this one too. Thanks for taking a look.

## ELI5
This PR fixes a bug in how the library handles Zod v4 “optional” and “nullable” schemas. Instead of accidentally checking the wrapper label (optional/nullable), it now checks what’s inside—so unsupported inner schemas (like tuples) are left alone, while supported ones (like strings) still get converted correctly.

## Summary
- Fixed `@mastra/schema-compat` Zod v4 nullable/optional handlers to gate on the wrapped *inner* Zod type (not the `ZodNullable`/`ZodOptional` wrapper type).
- Added a `zodV4TypeName` helper to normalize v4 type names (from v4 internals) to the v3-style `Zod*` form, matching existing dispatcher logic.
- Preserved pass-through behavior for unsupported inner schemas (e.g., nullable/optional tuples), avoiding unnecessary provider/schema handling.
- Continued to process supported inner schemas (e.g., nullable/optional strings) as before.
- Added test coverage for both v3 and v4, including:
  - nullable tuple passthrough
  - optional tuple passthrough (including identity/instance preservation)
  - nullable/optional supported-type processing
- Documented the fix in a changeset entry.


TheIsrael1:
@abhay-codes07 thanks for the targeted fix. One small coverage request before maintainer review: since #18687 calls out the same inner-type gating defect for both nullable and optional handlers, could you add/confirm a Zod v4 regression test for an optional unsupported inner type too, e.g. `z.tuple([z.string()]).optional()` passes through without throwing? The current visible tests cover nullable tuple passthrough and nullable supported inner processing.

abhay-codes07:
Good call, done. Added two optional-handler regression tests in c0821bfd67: `z.tuple([z.string()]).optional()` passes through without throwing (and I assert it returns the same schema instance, so it's a real pass-through, not a re-wrap), plus `z.string().optional()` still gets processed. They run under both the v3 and v4 projects and are green. Thanks for the review.

abhiaiyer91 (approved):

PR #18706, opened 2026-06-30
feat(playground-ui): match navigation for trace/span search

### LINKED ISSUE #18705: Studio: cycle through matches when searching inside a trace/span
## What

In Studio's trace view, searching inside a span's **Input**, **Output**, **Metadata**, or **Attributes** sections highlights all matches and jumps to the first one, but there's no way to step through the matches one by one. With many matches you can't tell how many there are or move to the next/previous one.

## Proposal

Add browser-style "find" navigation to the span code search:

- **Enter** → go to the next match, **Shift+Enter** → previous (with wrap-around at both ends).
- Next / previous (▲ / ▼) buttons next to the search field.
- A `current / total` counter (e.g. `3 / 12`).
- The active match shown in a distinct color from the other matches and scrolled into view.

This mirrors the find bar (Ctrl/Cmd+F) behavior people already expect.

## Scope

`packages/playground-ui` — `DataCodeSection` (the search highlight extension) and `SearchFieldBlock` (counter + nav controls, kept backward compatible).

## Notes

I've already implemented this and I'm happy to open a PR. Match-finding and index cycling are extracted into pure, unit-tested helpers, and the change is covered by unit + jsdom tests and verified in Storybook.

LuisMUtec:
## Description

Adds browser-style match navigation to the search inside a span's **Input**, **Output**, **Metadata**, and **Attributes** sections in Studio (the `DataCodeSection` component).

Before, the search highlighted every match and jumped to the first one, with no way to step through them. Now it works like a browser find bar:

- **Enter** → next match, **Shift+Enter** → previous, with wrap-around at both ends.
- Next / previous (▲ / ▼) buttons next to the search field.
- A `current / total` counter (e.g. `3 / 12`).
- The active match is shown in a distinct color from the other matches and scrolled into view.

### How it works

- `DataCodeSection`: the CodeMirror highlight `StateField` now tracks the match ranges plus the active index and paints two decorations — `cm-search-match` for all matches and `cm-search-match-current` for the active one. A shared `useCodeSearch` hook drives both the inline and the expanded editors (removing the previous duplicated dispatch logic) and wires up keyboard + button navigation.
- `search-matches.ts` (new): pure `findMatchRanges` / `getNextMatchIndex` helpers (case-insensitive, literal, wrap-around) so the logic is unit-testable without a live editor.
- `SearchFieldBlock`: gains optional, backward-compatible props (`onKeyDown`, `matchCount`, `currentMatch`, `onNext`, `onPrev`) to render the counter and nav controls. Existing consumers are unaffected.

### Tests

- Unit tests for `findMatchRanges` / `getNextMatchIndex` (multiple matches, case-insensitive, overlap, literal query, wrap-around forward/back, zero matches, empty query).
- jsdom test for `DataCodeSection` covering the counter, Enter / Shift+Enter cycling, next/previous buttons, and the disabled `0/0` state.
- Verified manually in Storybook (`Composite/DataCodeSection` → `SearchNavigation`).

## Related issue(s)

Fixes #18705

## Type of change

- [x] Documentation update

## Checklist

## ELI5
Studio now lets you search inside a span like a browser “Find” bar: it highlights every match, shows which one you’re on, and lets you jump forward/backward with Enter (or Shift+Enter). The currently selected match is kept in view as you move through results.

## Summary by CodeRabbit

* **New Features**
  * Added browser-style match navigation to Studio `DataCodeSection` search across a span’s **Input**, **Output**, **Metadata**, and **Attributes**:
    * highlights **all** matches
    * applies **distinct styling** to the **active** match (outline ring)
    * shows a **current / total** match counter
    * supports stepping **forward/backward** via **Enter / Shift+Enter** (with wrap-around)
    * adds **Next match / Previous match** buttons
    * **scrolls** the active match into view
  * Introduced shared, reusable navigation primitives:
    * `useMatchNavigation` hook for wrap-around cycling + keyboard handling
    * `MatchNav` UI component for the counter and prev/next controls
  * Implemented `useCodeSearch` to centralize CodeMirror-specific searching behavior (collapsed + expanded editors), including reapplying highlights after editor remounts and updating selection/scroll for the active match.

* **Implementation Details**
  * Added pure helpers for matching/cycling:
    * `findMatchRanges` (literal, case-insensitive matching) returns stable `{ from, to }` ranges
    * `getNextMatchIndex` provides wrap-around match stepping (re-exported for a consistent import surface)
  * Reworked highlighting to a decoration-based approach driven by precomputed match ranges:
    * builds all match decorations in one pass
    * distinguishes active vs non-active matches
    * preserves decorations across document edits by mapping through CodeMirror transactions
  * Updated `SearchFieldBlock` to remain backward-compatible while optionally rendering match navigation:
    * new optional props for match count/current index and prev/next callbacks
    * added `onKeyDown` passthrough for key handling
    * adjusted right-side padding so the counter/controls don’t overlap the input’s clear button cluster
  * Counter formatting:
    * displays capped totals as `999+` while retaining exact values in accessibility labels.

* **Tests**
  * Added unit tests for match logic:
    * `findMatchRanges` correctness (non-overlapping matches, case-insensitivity, literal matching, UTF-16 offset anchoring)
    * `getNextMatchIndex` stepping + wrap-around + empty/single-match behavior
  * Added jsdom/RTL tests for `DataCodeSection` match navigation:
    * counter visibility and correct `1/3`-style values after a query
    * Enter/Shift+Enter cycling forward/backward with wrap-around
    * prev/next button behavior (including disabled state when there are no matches)
    * recalculation when `codeStr` changes during an active search
  * Added tests for `MatchNav` and `useMatchNavigation`:
    * keyboard handling and event swallowing
    * button disabled state
    * counter capping behavior (`999+`) and aria-label accuracy.

* **Documentation / Verification**
  * Updated Studio observability docs to instruct users how to search within span panels and navigate matches (Enter/Shift+Enter + next/previous controls).
  * Added Storybook coverage for both `DataCodeSection` search navigation and `MatchNav` (including large-count display and hook-driven behavior).


TheIsrael1:
## 🟡 Maintainer's Triage Note

**Current Phase:** Reviewed

**Triage:**
- Type: feature request — Adds browser-style match navigation for Studio trace/span search results.
- Maintainer read: Issue #18705 requests next/previous navigation, a current/total counter, active-match styling, and scroll-to-active-match behavior for span search in Studio; PR #18706 explicitly closes the open issue and changes the expected playground-ui/docs areas.
- Next Steps: Maintainer approval/merge once routine checks are acceptable.
- Severity: 🟡 medium — Real Studio usability gap in trace/span search, but limited to navigating highlighted matches rather than a core outage or data-loss path.
- Confidence: High

**Review:**
- Status: reviewed
- Findings: No blocking code findings from maintainer review. The PR stays in the requested scope (`DataCodeSection`, `SearchFieldBlock`, tests/story/docs/changeset), adds pure helper coverage for matching/cycling, adds component coverage for counter/navigation/wraparound/no-match/resync behavior, and includes docs plus a Storybook scenario.
- Manual review notes: Verified the follow-up handles `codeStr` changes and CodeMirror remounts by re-running/re-applying the active search. The literal `RegExp('gi')` matcher preserves UTF-16 offsets for CodeMirror selections; worth keeping an eye on non-ASCII case-folding expectations, but I did not find a concrete blocking regression in the reviewed paths.
- Verification: Attempted `pnpm --filter ./packages/playground-ui test -- DataCodeSection` in a detached PR worktree; local verification is blocked before tests run because pnpm invokes install/deps-status and the supply-chain lockfile policy check over 4345 entries is killed with SIGTERM after 120s. GitHub checks currently show CodeRabbit/security/automation checks passing; Vercel failures are authorization-required deploy failures.
- Confidence: 4/5
- Follow-up: If maintainers want extra confidence, rerun the playground-ui focused test/typecheck in CI or an environment where the pnpm deps-status check can complete.

**Approve:**
- Status: ready for CODEOWNER/maintainer approval
- Final approver: Recommend @damien-schneider or @mfrachet based on `packages/playground-ui` 
- Notes: Maintainer review found no blocking code findings. Merge readiness depends on CODEOWNER/maintainer approval plus acceptable CI signal; current blocking status appears tied to deploy/check authorization rather than reviewed code concerns.

mfrachet:
I'm almost wondering if we couldn't abstract a resuable hook / compound component set that allows to nav up / down / enter in a consistent way accross the app

LuisMUtec:
Good call — I explored this and the abstraction falls out cleanly. I split it into a generic `useMatchNavigation` hook (active index + wraparound + Enter/Shift+Enter/optional arrow keys, agnostic of what a "match" is via an `onActiveChange` callback) and a small `MatchNav` counter/prev-next component extracted from `SearchFieldBlock`. `useCodeSearch` composes the hook and keeps only the CodeMirror-specific parts (highlight dispatch + scroll). Net effect: -52 lines in the existing files, and the existing `DataCodeSection` tests pass unchanged.

There's also a concrete second consumer already in the codebase: the experiment trace timeline search (`experiment-result-trace-panel.tsx`) highlights matching spans via `featuredSpanIds` but has no counter/navigation today — it could adopt the same hook, and `SearchFieldBlock` already accepts the match-nav props.

Happy to either fold this into this PR or land the current scope and open a follow-up with the extraction (+ the experiment timeline adoption as its first second consumer) — whichever you prefer. I have the refactor working locally with tests either way.

mfrachet:
@LuisMUtec you can add the abstractions here, that's fine, thanks a lot for your help 🙏🏻

packages/playground-ui/src/ds/components/DataCodeSection/data-code-section.tsx
```diff
@@ -25,39 +26,49 @@ import {
...
 const searchHighlightTheme = EditorView.baseTheme({
+  // Non-active matches: a faint wash so the current match clearly stands out against them.
   '.cm-search-match': {
-    backgroundColor: 'color-mix(in srgb, var(--accent1) 60%, transparent)',
+    backgroundColor: 'color-mix(in srgb, var(--accent1) 35%, transparent)',
+    borderRadius: 'var(--radius-sm)',
+  },
+  // The current match the user is parked on: a strong, near-solid accent fill.
+  '.cm-search-match-current': {
+    backgroundColor: 'color-mix(in srgb, var(--accent1) 85%, transparent)',
```
damien-schneider: I think those changes cause accessibility regressions, could you update the values to work in both dark and light mode?

Image

damien-schneider (changes requested):
Thank you very much for the PR and the recent updates, sorry for the delay on the review, just caught one issue, but seems good overall

packages/playground-ui/src/ds/components/DataCodeSection/data-code-section.tsx
damien-schneider: I think having a ring instead of background change can easily solve that

packages/playground-ui/src/ds/components/FormFieldBlocks/fields/search-field-block.tsx
```diff
@@ -99,13 +116,15 @@ export function SearchFieldBlock({
...
             className={cn(
-              size === 'sm' && 'pl-8 pr-8',
-              size === 'md' && 'pl-9 pr-9',
-              (!size || size === 'default') && 'pl-10 pr-10',
-              size === 'lg' && 'pl-11 pr-11',
+              // Extra right padding when the match-nav cluster is shown so it never overlaps text.
+              size === 'sm' && (hasMatchNav ? 'pl-8 pr-[6.5rem]' : 'pl-8 pr-8'),
+              size === 'md' && (hasMatchNav ? 'pl-9 pr-28' : 'pl-9 pr-9'),
+              (!size || size === 'default') && (hasMatchNav ? 'pl-10 pr-[7.5rem]' : 'pl-10 pr-10'),
+              size === 'lg' && (hasMatchNav ? 'pl-11 pr-32' : 'pl-11 pr-11'),
```
damien-schneider: And fixed `pr-*` values break with a wide counter (searching one letter in a big payload gives e.g. `2841/2841`, which overlaps the typed text), either render `MatchNav` outside the input or cap the displayed total like `999+` it shouldn't happen that often of course, but if we can include these changes that would be nice

packages/playground-ui/src/ds/components/DataCodeSection/data-code-section.tsx
LuisMUtec: Fixed in 1ec61aa — the active match now uses a ring plus the same faint wash as other matches, so syntax colors keep their contrast in both themes.

packages/playground-ui/src/ds/components/FormFieldBlocks/fields/search-field-block.tsx
LuisMUtec: Fixed in b738f3d — displayed values cap at 999+ (aria-label keeps real counts) and the input's right padding is now computed from the rendered counter, so it can't overlap the typed text.

LuisMUtec:
Thanks for the review! Both points are addressed in the latest commits (1ec61aa, b738f3d) — replied inline with the details.

packages/playground-ui/src/ds/components/FormFieldBlocks/fields/search-field-block.tsx
```diff
@@ -43,14 +62,23 @@ export function SearchFieldBlock({
...
   isMinimized,
   onMinimizedChange,
+  matchCount,
+  currentMatch,
+  onNext,
+  onPrev,
 }: SearchFieldBlockProps) {
   const inputRef = useRef<HTMLInputElement>(null);
   const buttonSize = size === 'default' ? 'lg' : size;
+  const hasMatchNav = matchCount !== undefined;
```
mfrachet, line 78: I'm wondering if this shouldn't live in the caller side? SearchField by default does not need to know if anything outside of it is matching.

I think this should be done at the parent level, then just passing down a className adjustment (conditionally). A "Searchbar" does not need to know about "onNext", "onPrev" etc...

If we really want to have something like this, a compound component set would probably be a bit more conveniant. If not, just passing down className should do the job and leaves the parent deciding what the style of the search field should be in this context.

packages/playground-ui/src/ds/components/FormFieldBlocks/fields/search-field-block.tsx
```diff
@@ -118,24 +151,29 @@ export function SearchFieldBlock({
...
-                  onMinimizedChange?.(true);
-                }
-              }}
-              className="absolute top-1/2 right-0 -translate-y-1/2"
-            >
-              <XIcon />
-            </Button>
-          )}
+          <div className="absolute top-1/2 right-0 flex -translate-y-1/2 items-center">
+            {hasMatchNav && (
```
mfrachet, line 155: shouldn't live in the searchbar

packages/playground-ui/src/ds/components/MatchNav/match-nav.tsx
```diff
@@ -0,0 +1,55 @@
...
+  onPrevious?: () => void;
+  className?: string;
+};
+
+/**
+ * Browser-find-bar-style "current/total" counter with previous/next controls. Purely
+ * presentational — pair it with `useMatchNavigation` (or any index state) for the stepping logic.
+ * Rendered inline by `SearchFieldBlock` when match navigation props are provided.
+ */
+export function MatchNav({ current, total, onNext, onPrevious, className }: MatchNavProps) {
```
mfrachet, line 21: This should be a compound component set so we can restyle each sub component easily.

```tsx
<MatchNav>
<MatchNavNext/>
<MatchNavPrev/>
<MatchNavCount/>
</MatchNav>
```

packages/playground-ui/src/hooks/use-match-navigation.ts
```diff
@@ -0,0 +1,105 @@
...
+
+export interface MatchNavigation {
+  /** Index of the active match in `matches`, or -1 when there are none. */
+  activeIndex: number;
+  /** 1-based position of the active match for "current/total" counters; 0 when there are none. */
+  current: number;
+  /** Total number of matches. */
+  total: number;
+  /** Step to the next match, wrapping around at the end. */
+  goToNext: () => void;
```
mfrachet, line 39:
```suggestion
  onNext: () => void
```

packages/playground-ui/src/hooks/use-match-navigation.ts
```diff
@@ -0,0 +1,105 @@
...
+  /** Step to the previous match, wrapping around at the start. */
+  goToPrevious: () => void;
```
mfrachet, line 41:
```suggestion
  onPrev: () => void;
```

packages/playground-ui/src/hooks/use-match-navigation.ts
```diff
@@ -0,0 +1,105 @@
...
+  /** Keydown handler for the search input: Enter / Shift+Enter (and optionally ArrowDown/Up). */
+  onSearchKeyDown: React.KeyboardEventHandler<HTMLInputElement>;
```
mfrachet, line 43:
```suggestion
  onKeyDown: React.KeyboardEventHandler<HTMLInputElement>;
```

packages/playground-ui/src/hooks/use-match-navigation.ts
```diff
@@ -0,0 +1,105 @@
...
+  matches,
+  onActiveChange,
+  arrowKeys = false,
+}: UseMatchNavigationOptions<T>): MatchNavigation {
+  const [activeIndex, setActiveIndex] = useState(-1);
+
+  // Reset to the first match whenever a new match list arrives. Identity-compared via a ref so a
+  // re-render with the same array (or an unstable `onActiveChange`) doesn't re-trigger the reset.
+  const lastMatchesRef = useRef<T[] | null>(null);
+  useEffect(() => {
```
mfrachet, line 65: This is a code smell: useEffect should be used for syncing React and the outside world. Here, we are re-computing state from inside an effect which is not a good practice.

packages/playground-ui/src/hooks/use-match-navigation.ts
```diff
@@ -0,0 +1,105 @@
...
+      setActiveIndex(nextIndex);
+      onActiveChange?.(nextIndex, matches);
+    },
+    [matches, activeIndex, onActiveChange],
+  );
+
+  const goToNext = useCallback(() => goTo(1), [goTo]);
+  const goToPrevious = useCallback(() => goTo(-1), [goTo]);
+
+  const onSearchKeyDown = useCallback<React.KeyboardEventHandler<HTMLInputElement>>(
```
mfrachet, line 86: we don't need useCallback and useMemo in this entire suggestions. Perf is not a concern here

mfrachet (changes requested):
Thanks again for the iteration 🙏🏻 . I still have a few concern regarding the React codebase, but the feature will look awesome once in a good shape 🙏🏻

abhiaiyer91:
Thanks for the contribution! #18705 was closed after search utilities landed in the traces view on `main`, so this PR is no longer needed. Closing it to keep the queue focused. If the current search still falls short for your use case, please reopen #18705 with specifics and we'll take another look.

PR #18759, opened 2026-07-01
fix(core): stream array textStream incrementally to keep JSON valid

### LINKED ISSUE #18758: [BUG] Array textStream emits invalid JSON when the first object chunk already has elements
### Describe the Bug

`createJsonTextStreamTransformer` (`packages/core/src/stream/base/output-format-handlers.ts`) produces invalid JSON for an array `textStream` when the first `object` chunk already contains one or more elements.

Object chunks for array output carry the array-so-far and grow over time. The transformer has a first-chunk fast path that assumes the first non-empty chunk is the complete array and emits a fully closed JSON string:

```ts
if (chunkCount === 1) {
  if (chunk.object.length > 0) {
    controller.enqueue(JSON.stringify(chunk.object)); // emits a closed "[...]"
    previousArrayLength = chunk.object.length;
    hasStartedArray = true;
    return;
  }
}
```

When later chunks arrive, the transformer appends more elements and then a closing `]` on flush, so the output has already-closed array followed by more elements: broken JSON.

### Steps To Reproduce

Feed the transformer (with a `z.array(...)` schema) these progressive object chunks: `[{a:1}]`, `[{a:1},{a:2}]`, `[{a:1},{a:2},{a:3}]`.

- **Actual output:** `[{"a":1}],{"a":2},{"a":3}]` → `JSON.parse` throws "Unexpected non-whitespace character after JSON at position 9".
- **Expected output:** `[{"a":1},{"a":2},{"a":3}]`.

This is reachable whenever the first parseable `object` chunk already has an element (coarse or batched provider deltas). Fine-grained streaming (first chunk `[]`) and single-chunk generate mode happen to avoid the fast path, so they are unaffected.

### Expected Behavior

The array `textStream` always concatenates to a single valid JSON array.

### Additional Context

Fix: drop the first-chunk fast path and always stream incrementally (open `[` once, emit only newly-added elements, close `]` on flush). This is behavior-preserving for the working cases and produces valid JSON for the broken one. I have a PR ready with unit tests covering first-chunk-nonempty, first-chunk-empty, and single-chunk.

abhay-codes07:
## What

`createJsonTextStreamTransformer` emits invalid JSON for an array `textStream` when the first `object` chunk already contains elements. Object chunks carry the array-so-far and grow over time, but the transformer has a first-chunk fast path that assumes the first non-empty chunk is the complete array and writes a fully-closed `JSON.stringify(array)`. When later chunks arrive it appends more elements plus a closing `]` on flush, producing broken output.

Feeding progressive chunks `[{a:1}]`, `[{a:1},{a:2}]`, `[{a:1},{a:2},{a:3}]`:
- Actual: `[{"a":1}],{"a":2},{"a":3}]` → `JSON.parse` throws.
- Expected: `[{"a":1},{"a":2},{"a":3}]`.

This fires whenever the first parseable chunk already has an element (coarse/batched provider deltas). Fine-grained streaming (first chunk `[]`) and single-chunk generate mode avoid the fast path, so they were unaffected.

## Change

Drop the first-chunk fast path and always stream incrementally: open `[` once, emit only newly-added elements, close `]` on flush. Behavior-preserving for the empty-first-chunk and single-chunk cases (same final text), and valid for the broken case.

## Tests

Added unit tests driving the transformer with a `z.array(...)` schema for all three cases (first-chunk-nonempty, first-chunk-empty, single complete chunk). The first-chunk-nonempty case fails on the current code and passes after the fix. The full stream suite (13 files, 181 tests) still passes.

Closes #18758

---

Hey @epinzur, you were last in this file so tagging you. Happy to add a schema-level e2e test too if you'd prefer. Thanks for the look.

## ELI5
This fix makes streamed JSON arrays stay valid while they’re being sent piece by piece. Before, if the first chunk already had items, the code could accidentally close the array too early and then keep adding more, which broke the JSON.

## Summary
- Changed `createJsonTextStreamTransformer` to always stream array outputs incrementally.
- Removed the first-chunk fast path that treated the first non-empty array chunk as complete JSON.
- Now it opens `[` once, emits only newly added items as chunks arrive, and closes `]` on flush.
- Added tests for:
  - first chunk already containing elements
  - first chunk being empty
  - a single complete chunk
- Updated the patch release note to document the fix.

### DIFF
--- a/.changeset/fix-array-textstream-invalid-json.md
+++ b/.changeset/fix-array-textstream-invalid-json.md
@@ -1,0 +1,5 @@
+---
+'@mastra/core': patch
+---
+
+Fix array `textStream` emitting invalid JSON when the first streamed object chunk already contains elements. The JSON text transformer wrote a fully-closed array on the first non-empty chunk and then appended more elements plus another closing bracket, producing unparseable output. It now always streams incrementally, so the array `textStream` concatenates to a single valid JSON array. Closes #18758.
--- a/packages/core/src/stream/base/json-text-stream-array.test.ts
+++ b/packages/core/src/stream/base/json-text-stream-array.test.ts
@@ -1,0 +1,50 @@
+import { describe, expect, it } from 'vitest';
+import { z } from 'zod';
+import { toStandardSchema } from '../../schema';
+import { createJsonTextStreamTransformer } from './output-format-handlers';
+
+async function runTransformer(schema: any, objectChunks: any[]): Promise<string> {
+  const transformer = createJsonTextStreamTransformer(schema);
+  const writer = transformer.writable.getWriter();
+  const reader = transformer.readable.getReader();
+
+  const out: string[] = [];
+  const readAll = (async () => {
+    while (true) {
+      const { done, value } = await reader.read();
+      if (done) break;
+      out.push(value);
+    }
+  })();
+
+  for (const object of objectChunks) {
+    await writer.write({ type: 'object', object, runId: '1', from: 'AGENT' } as any);
+  }
+  await writer.close();
+  await readAll;
+
+  return out.join('');
+}
+
+describe('createJsonTextStreamTransformer array output', () => {
+  const schema = toStandardSchema(z.array(z.object({ a: z.number() }))) as any;
+
+  it('produces valid JSON when the first chunk already has elements', async () => {
+    const text = await runTransformer(schema, [[{ a: 1 }], [{ a: 1 }, { a: 2 }], [{ a: 1 }, { a: 2 }, { a: 3 }]]);
+
+    expect(() => JSON.parse(text)).not.toThrow();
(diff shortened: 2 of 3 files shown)


abhay-codes07:
Quick note on sequencing, @epinzur. I have two more fixes ready and tested and would like to open them next: #18771 (DurableAgent ignoring the readOnly memory config and persisting messages anyway) and #18774 (modelContent dropping step 1 content and mis-numbering later steps). Both are done with regression tests and the local suites are green.

The catch is that GitHub only lets an outside contributor keep about two PRs open at a time, and this one plus #18755 are already using those slots. Whenever you have a chance to review and merge this one, it frees a slot and I will push one of the two fixes above right away. Appreciate the review.
cc @abhiaiyer91

abhiaiyer91:
@abhay-codes07 we will review and get to them on our schedule.

abhay-codes07:
Superseded by #21559, which fixes the same issue with a buffering approach that also handles the single-chunk case cleanly. Closing this in favor of that. Thanks.

PR #18796, opened 2026-07-02
fix(auth): make auth-provider boundary types structural

### LINKED ISSUE #18682: @mastra/auth bundles a #private-branded copy of @mastra/core base class → MastraAuthProvider not assignable to server.auth
## Summary
Since the auth-internals extraction, `@mastra/auth` (pulled by every provider package, e.g. `@mastra/auth-workos`) has **no `@mastra/core` peer dependency** and instead **bundles its own copy of core's base class** under `dist/_types/@internal_auth/dist/_types/@internal_core/dist/base/index.d.ts`. That base class carries a `#private` member, so it's a **nominal** brand. `@mastra/core` bundles its *own* copy of the same base. The two copies are structurally identical but nominally distinct.

Result: a `MastraAuthProvider` produced by `@mastra/auth-workos` is **not assignable** to the `MastraAuthProvider<any>` that `Mastra`'s `server.auth` expects:

```
Type 'MastraAuthWorkos' is not assignable to type 'MastraAuthConfig<any> | MastraAuthProvider<any>'.
  Property '#private' is missing in type 'MastraAuthWorkos' but required in type 'MastraAuthProvider<any>'.
```

## Reproduction
With `@mastra/core@1.47.0` + `@mastra/auth-workos@1.6.1`:

```ts
import { MastraAuthWorkos } from '@mastra/auth-workos';

new Mastra({ server: { auth: new MastraAuthWorkos() } }); // TS2322: '#private' missing
```

## Scope (matrix tested)
- Breaks for **every `@mastra/core` >= 1.42** (the first release with the `actor` FGA signal) against **every `@mastra/auth-workos`** I tried (1.5.0 and 1.6.1).
- Works on `@mastra/core@1.38` + `@mastra/auth-workos@1.5.0` — only because `@mastra/auth@1.0.2`'s bundled base had **no `#private`** member (the check was structural). It became nominal in `@mastra/auth@1.1.x`.
- Not WorkOS-specific — all provider packages share `@mastra/auth`, so each hits it once a consumer assigns the provider to `server.auth`.

## Root cause
`@mastra/auth@1.1.1`'s `package.json` declares no `@mastra/core` dependency (peer or otherwise); the core base is inlined into the published `dist`. So there is no shared `@mastra/core` identity for the package manager to dedupe to — the brand can never match core's.

## Suggested fix
Have `@mastra/auth` (and the `@internal/*` packages it bundles) **peer-depend on `@mastra/core`** rather than inlining a private-branded base, so a single deduped base-class identity is shared across core and the provider packages. Alternatively, drop the `#private` brand from the extracted base if the nominal identity is not intended.

## Workaround (downstream)
A single `as unknown as MastraAuthProvider<...>` cast at the `server.auth` assignment — the classes are identical at runtime; it is a pure type-identity false negative.

abhiaiyer91:
Fixes #18682.

`@mastra/core` and every auth provider package ship their own bundled copy of the `MastraAuthProvider`/`MastraBase` declarations. Those copies carry nominal brands (`#private` and `protected` members), which TypeScript checks by declaration identity — so two identical copies are mutually unassignable. Assigning a provider instance to `server.auth` failed to compile under strict configs (notably `exactOptionalPropertyTypes: true`), even though runtime is fully duck-typed and unaffected.

Before:

```ts
const server = {
  auth: new MastraAuthWorkos({ apiKey, clientId }),
  // TS2322: Property '#private' is missing in type 'MastraAuthWorkos'
};
```

After:

```ts
new Mastra({
  server: { auth: new MastraAuthWorkos({ apiKey, clientId }) }, // compiles
});
```

The fix: positions that accept user-supplied providers (`server.auth`, `studio.auth`, `CompositeAuth`) now accept a structural `IMastraAuthProvider<TUser>` interface instead of the branded class, so assignability no longer depends on shared declaration identity across packages. `MastraAuthProvider` declares `implements IMastraAuthProvider` so the compiler keeps class and interface in sync, and the interface is exported from `@mastra/core/server` and `@mastra/auth`. The boundary rule (types crossing published package boundaries via `@internal/*` bundling must be structural — no `#private`/`protected`/`instanceof` reliance) is documented in the root and `packages/_types-builder` AGENTS.md.

Regression coverage: duplicate-copy assignability assertions in core's `server.test-d.ts` (simulates a bundled branded copy and asserts it stays assignable to the interface), and a new `auth.test-d.ts` in `e2e-tests/type-check` that exercises the exact broken patterns against packed artifacts installed from a local registry under `exactOptionalPropertyTypes: true`.

Test plan: core typecheck + auth test suites (43 tests), `@internal/auth` unit tests (106), builds of core/auth/auth-workos, and the full `e2e-tests/type-check` suite (9 files, 144 tests incl. 4 new auth assertions) passing against packed artifacts with the nominal brands still present in the bundled copies — confirming the interface alone resolves the issue.


packages/_types-builder/src/replace-types.js
```diff
@@ -132,6 +132,56 @@ async function resolveRelativeDeclaration(moduleSpecifier, fromFile) {
...
+ * Bundled declaration files are *copies*: multiple packages ship their own
+ * duplicate of the same class. Nominal members make those otherwise identical
+ * copies mutually unassignable (e.g. `'#private' is missing in type ...`),
+ * breaking users who pass instances across package boundaries (see #18682).
+ * A nominal brand asserts "same declaration = same type", which is false by
+ * construction in a duplicated declaration file, so we strip the brands.
+ * Members are dropped entirely (not made public) so copies stay structurally
+ * compatible regardless of visibility differences.
+ */
+export function stripNominalBrands(sourceFile) {
```
wardpeet: we shouldnt need this anymore

wardpeet (approved):

wardpeet (approved):

PR #18801, opened 2026-07-02
fix: eliminate polynomial ReDoS in string parsing paths (CodeQL)

abhiaiyer91:
## Summary

Resolves all 10 open CodeQL `js/polynomial-redos` alerts (high severity): #732–736, #779, #810–812, #815.

Two anti-patterns were flagged, both of which back-track polynomially on adversarial input:

**1. Slash/dot trim regexes** — patterns like `/\/+$/` and `/^\/+|\/+$/g` are quadratic on inputs like `"////…////x"` because the engine re-scans the run from every start position. Replaced with linear index scans in:

- `packages/core/src/workspace/skills/publish.ts` (`joinPath`)
- `packages/core/src/workspace/skills/versioned-skill-source.ts` (`#normalizePath`)
- `packages/deployer/src/server/welcome.ts` (API prefix normalization)
- `auth/okta/src/auth-provider.ts` (issuer trailing-slash trim)
- `packages/playground-ui/src/ee/signals/services/entity-learning-service.ts` (`trimTrailingSlash`)

**2. Ambiguous `\s*` runs around optional/lazy groups** — replaced with incremental string parsing in:

- `packages/core/src/stream/aisdk/v5/transform.ts` (`sanitizeToolCallInput`, processes raw LLM output)
- `packages/playground-ui/src/domains/memory/components/observation-detail-view.tsx` (observation memory parser: bullet items, `Date:` headings, `<observations>` extraction)

These matter beyond lint hygiene: the transform runs on untrusted LLM provider output and the observation parser on memory content of arbitrary size.

## Verification

- Differential testing of old regex vs new implementations across edge cases (nested parens, empty groups, missing tags, tight spacing) — behavior-equivalent everywhere. The differential test caught one subtle case during development (nested parens in `Date:` headings, where `[^)]+` rejects `)`) which the final implementation mirrors correctly.
- Adversarial 50k-char inputs that hang the old patterns now parse in <0.1ms.
- `tsc --noEmit` clean in all 4 packages.
- Tests: core `transform.test.ts` + `workspace/skills` (298 passed), auth/okta (35 passed), deployer server suite (99 passed).

## ELI5
This PR updates some “cut up text” logic so it can’t get tricked into taking extremely long time on nasty input. It keeps the same outputs, but replaces a few risky regexes with simpler, faster step-by-step string parsing.

## Summary
- Fixed CodeQL `js/polynomial-redos` high-severity alerts by removing regex backtracking hazards in untrusted/LLM-derived parsing paths:
  - Replaced slash/dot trimming regexes with linear index-based scans in:
    - `packages/core/src/workspace/skills/publish.ts`
    - `packages/core/src/workspace/skills/versioned-skill-source.ts`
    - `packages/deployer/src/server/welcome.ts`
    - `auth/okta/src/auth-provider.ts`
    - `packages/playground-ui/src/ee/signals/services/entity-learning-service.ts`
  - Replaced ambiguous `\s*` patterns around optional/lazy groups with incremental parsing in:
    - `packages/core/src/stream/aisdk/v5/transform.ts`
    - `packages/playground-ui/src/domains/memory/components/observation-detail-view.tsx`
- Preserved behavior while eliminating polynomial ReDoS risk (including cases involving large model-generated content).
- Added durable-agent E2E test recording fixtures covering background-task and foreground-tool response streaming flows.

## Verification
- Differential testing vs. prior regex behavior
- Adversarial large-input tests (including 50k-character cases)
- `tsc --noEmit`
- Package test suites passing for affected areas


auth/okta/src/auth-provider.ts
```diff
@@ -182,7 +182,11 @@ export class MastraAuthOkta
     this.clientId = clientId;
     this.clientSecret = clientSecret;
     // Normalize trailing slashes so a stray `OKTA_ISSUER=https://domain/` doesn't produce `.../oauth2//v1/...`
-    this.issuer = (issuer ?? `https://${domain}/oauth2/default`).replace(/\/+$/, '');
+    // (index scan instead of regex to avoid backtracking — CodeQL js/polynomial-redos)
+    const rawIssuer = issuer ?? `https://${domain}/oauth2/default`;
```
wardpeet: Can we use replaceAll instead?

packages/core/src/workspace/skills/publish.ts
```diff
@@ -126,15 +126,25 @@ async function walkSkillDirectory(
...
 /**
  * Join path segments using forward slashes.
  */
 function joinPath(...segments: string[]): string {
   return segments
-    .map((seg, i) => {
-      if (i === 0) return seg.replace(/\/+$/, '');
-      return seg.replace(/^\/+|\/+$/g, '');
-    })
+    .map((seg, i) => trimSlashes(seg, i > 0))
```
wardpeet, line 147: same useReplaceAll for readability

packages/core/src/workspace/skills/versioned-skill-source.ts
```diff
@@ -46,11 +46,15 @@ export class VersionedSkillSource implements SkillSource {
...
    * Normalize a path by stripping leading/trailing slashes and dots.
+   * Uses index scans instead of regex to avoid backtracking on adversarial
+   * inputs (CodeQL js/polynomial-redos).
    */
   #normalizePath(path: string): string {
-    let normalized = path.replace(/^[./\\]+|[/\\]+$/g, '');
-    if (normalized === '') return '';
-    return normalized;
+    let start = 0;
+    let end = path.length;
```
wardpeet: same here

packages/deployer/src/server/welcome.ts
```diff
@@ -1,6 +1,12 @@
 export function welcomeHtml(apiPrefix: string = '/api') {
-  // Normalize: ensure single leading slash, no trailing slash
-  const prefix = '/' + apiPrefix.replace(/^\/+|\/+$/g, '');
+  // Normalize: ensure single leading slash, no trailing slash.
+  // Index scans instead of regex to avoid backtracking on adversarial
+  // inputs (CodeQL js/polynomial-redos).
+  let start = 0;
+  let end = apiPrefix.length;
```
wardpeet: I would also make keep thie regex as it's not ussr input

packages/playground-ui/src/domains/memory/components/observation-detail-view.tsx
```diff
@@ -92,17 +92,39 @@ function parseItem(line: string): ParsedItem | null {
...
+      rest = rest.slice(timeMatch[0].length).trimStart();
+    }
+    if (rest) {
+      return { text: rest, time, priority, children: [] };
+    }
   }
   return { text: trimmed, time: null, priority: null, children: [] };
 }
 
 function parseObservations(raw: string): ParsedSection[] {
```
wardpeet, line 124: i'm also not sure if it's worht the complexity of this

packages/playground-ui/src/ee/signals/services/entity-learning-service.ts
```diff
@@ -38,7 +38,12 @@ export type EntityLearningOutlierExamplesParams = {
 
 export type EntityLearningService = ReturnType<typeof createEntityLearningService>;
 
-const trimTrailingSlash = (value: string) => value.replace(/\/+$/, '');
+// Index scan instead of regex to avoid backtracking (CodeQL js/polynomial-redos)
+const trimTrailingSlash = (value: string) => {
```
wardpeet, line 42: i would keep the old

wardpeet (approved):
I'm not really sure around these redos complains. Perhaps we should add https://es-toolkit.dev/reference/string/trim.html#trim-str-chars.

auth/okta/src/auth-provider.ts
abhiaiyer91: `replaceAll('/', '')` would strip every slash in the issuer URL (`https://domain/oauth2/default` → `https:domain oauth2default`), not just the trailing run — replaceAll can't express "trim from the end". And the trailing-run regex `/\/+$/` is exactly what CodeQL flagged (quadratic when a slash run appears mid-string). The 2-line scan was the smallest safe replacement. If you'd rather have es-toolkit's `trimEnd(str, '/')` here I'm happy to switch — it's currently only in the tree transitively via recharts, so it'd be a new direct dep.

packages/core/src/workspace/skills/publish.ts
abhiaiyer91: Same constraint as the okta one — replaceAll removes every occurrence, but `joinPath` needs to trim slash runs from the segment edges only (interior slashes must survive). That's why the four regexes got consolidated into the single `trimSlashes` helper: call sites read as `trimSlashes(seg, i > 0)` and the loop lives in one place.

packages/core/src/workspace/skills/versioned-skill-source.ts
abhiaiyer91: Same as publish.ts — this needs edge-trimming of `.`, `/`, `\\` runs, which replaceAll can't do without removing interior separators too. Kept the scan for that reason.

packages/deployer/src/server/welcome.ts
abhiaiyer91: Agreed it's config, not user input — the risk here is basically zero. But CodeQL can't see that, so keeping the regex leaves the alert open and we'd have to hand-dismiss it as false-positive (and re-dismiss if it ever resurfaces after refactors). Since behavior is identical either way, I kept the scan so the alert closes on merge. Can revert + dismiss if you feel strongly.

packages/playground-ui/src/domains/memory/components/observation-detail-view.tsx
abhiaiyer91: This is the one place I'd argue the complexity earns its keep: the input is model-generated observation text rendered client-side, and the old lazy regexes (`\s*(.+?)(?:\s*\(...\))?$`) go quadratic on crafted content — a pathological observation would freeze the playground tab on the UI thread. The new parsing is differential-tested against the old regexes, including adversarial 50k-char inputs: identical outputs, 0.04–0.06ms vs. unbounded backtracking.

packages/playground-ui/src/ee/signals/services/entity-learning-service.ts
abhiaiyer91: The old `/\/+$/` is exactly what CodeQL flagged (alert 815) — it backtracks quadratically when a slash run sits mid-string, and the base URL here comes from runtime config. The 3-line scan is the minimal regex-free fix. es-toolkit's `trimEnd(value, '/')` would make it a 1-liner if we want to take the dep (see the okta thread) — happy to do that in a follow-up if you prefer.

abhiaiyer91:
Replied to the review threads inline — the short version:

- **`replaceAll` doesn't fit these sites**: they all trim character *runs from the ends* (trailing/leading slashes), and `replaceAll('/', '')` would strip interior slashes too (e.g. mangle the okta issuer URL). The run-trimming regexes are precisely the patterns CodeQL flags as polynomial.
- **es-toolkit `trim`/`trimEnd`**: nice option, but it's currently only a transitive dep (via recharts) — adopting it means a new direct runtime dep in @mastra/core, @mastra/deployer, @mastra/auth-okta, and @mastra/playground-ui. Happy to do that as a follow-up if you think it's worth it; the scans here are dependency-free in the meantime.
- **welcome.ts / entity-learning-service**: agreed the inputs are config, not user input, but reverting re-opens the alerts and they'd need manual false-positive dismissals. Behavior is identical, so I kept the safe versions.
- **observation-detail-view**: this one is genuinely worth it — the input is model-generated text and the old lazy regexes freeze the UI thread on crafted content. Parsing is differential-tested against the old regexes (identical outputs, incl. adversarial 50k-char inputs).

auth/okta/src/auth-provider.ts
wardpeet: it's fine, it's just hard to read IMO. Adding it as a dev dep would get it bundled or mabye create a utiltiy fn so it's named properly

packages/core/src/workspace/skills/versioned-skill-source.ts
wardpeet: lets create a helper fn so it tells us that its trimming

packages/deployer/src/server/welcome.ts
wardpeet: no but maybe add a hlper fn so it's named

packages/playground-ui/src/domains/memory/components/observation-detail-view.tsx
wardpeet: Can you add a comment what it's extracting so it's more clear

auth/okta/src/auth-provider.ts
abhiaiyer91: Done in 1db447ce3e — extracted into a named `trimTrailingSlashes()` helper (skipped the dep since a one-off devDep-for-bundling felt heavier than 5 lines).

packages/core/src/workspace/skills/versioned-skill-source.ts
abhiaiyer91: Done in 1db447ce3e — extracted into `trimPathEdges()` with a doc comment; `#normalizePath` now just delegates to it.

packages/deployer/src/server/welcome.ts
abhiaiyer91: Done in 1db447ce3e — pulled the scan into a named `trimSlashes()` helper, so `welcomeHtml` is back to one readable line.

packages/playground-ui/src/domains/memory/components/observation-detail-view.tsx
abhiaiyer91: Done in 1db447ce3e — added comments to all three parsing blocks showing the line format each one extracts with an example (`* 🔴 (11:55) text`, `<observations>…</observations>`, `Date: Jul 2, 2026 (today)`) and the resulting fields.

PR #18823, opened 2026-07-02
fix(core): preserve empty scrubber instructions

tianma-if:
## Summary
- Preserve explicitly empty system prompt scrubber instructions
- Add a regression test for custom instruction handling

## Testing
- pnpm exec vitest run src/processors/processors/system-prompt-scrubber.test.ts
- pnpm --filter ./packages/core exec eslint src/processors/processors/system-prompt-scrubber.ts src/processors/processors/system-prompt-scrubber.test.ts
- pnpm --filter ./packages/core build:lib

## ELI5
This change makes the system prompt scrubber respect an empty instruction string instead of silently swapping in a default one. In other words, if someone intentionally sets the instructions to be empty, the code now keeps it that way.

## Summary
- Updated `SystemPromptScrubber` to use nullish coalescing so `''` is preserved while `null`/`undefined` still fall back to default instructions.
- Added a regression test to confirm `instructions: ''` remains unchanged.
- Included a changeset noting the patch release and behavior fix for empty custom instructions.

### DIFF
--- a/.changeset/tame-worlds-strive.md
+++ b/.changeset/tame-worlds-strive.md
@@ -1,0 +1,5 @@
+---
+'@mastra/core': patch
+---
+
+Fixed the system prompt scrubber so empty custom instructions are preserved.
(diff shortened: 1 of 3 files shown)


abhiaiyer91:
Please open issues before opening PRs. Please do not open drafts. I will close the open PRs until we have an issue attached.

PR #18824, opened 2026-07-02
fix(core): preserve empty structured instructions

tianma-if:
## Summary
- Preserve explicitly empty structured output processor instructions
- Add a regression test for custom instruction handling

## Testing
- pnpm exec vitest run src/processors/processors/structured-output.test.ts
- pnpm --filter ./packages/core exec eslint src/processors/processors/structured-output.ts src/processors/processors/structured-output.test.ts
- pnpm --filter ./packages/core build:lib

## ELI5
This change makes the code respect when someone explicitly sets instructions to an empty string instead of ignoring it and making up new ones. It also adds a test to make sure that special case keeps working.

## Summary
- Fixed `StructuredOutputProcessor` to use nullish coalescing so empty `instructions` are preserved.
- Added a regression test covering empty custom instructions.
- Included a changeset entry to publish a patch for `@mastra/core`.

### DIFF
--- a/.changeset/empty-foxes-mix.md
+++ b/.changeset/empty-foxes-mix.md
@@ -1,0 +1,5 @@
+---
+'@mastra/core': patch
+---
+
+Fixed structured output processors so empty custom instructions are preserved.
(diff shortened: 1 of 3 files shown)


abhiaiyer91:
Please open issues before opening PRs. Please do not open drafts. I will close the open PRs until we have an issue attached.

PR #18844, opened 2026-07-02
fix(core): preserve batched non-text stream parts

### LINKED ISSUE #18560: BatchPartsProcessor drops non-text parts when emitOnNonText is false

tianma-if:
Fixes #18560.

This keeps BatchPartsProcessor from dropping non-text parts when emitOnNonText is false and a mixed batch is flushed. The flush path now emits leading text deltas as one chunk, then leaves later non-text and text parts queued for the next flush.

Tested with:

pnpm --filter ./packages/core test src/processors/processors/batch-parts.test.ts

### DIFF
--- a/.changeset/modern-dogs-relate.md
+++ b/.changeset/modern-dogs-relate.md
@@ -1,0 +1,5 @@
+---
+'@mastra/core': patch
+---
+
+Fixed batch parts processing so non-text stream parts are preserved when mixed batches are flushed.
--- a/packages/core/src/processors/processors/batch-parts.test.ts
+++ b/packages/core/src/processors/processors/batch-parts.test.ts
@@ -268,6 +268,65 @@ describe('BatchPartsProcessor', () => {
       expect(result).toBeNull();
     });
 
+    it('preserves non-text parts when flushing mixed batches with emitOnNonText false', async () => {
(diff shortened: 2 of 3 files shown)


tianma-if:
Closing this draft to follow the maintainer guidance: no draft PRs, and PRs should be linked to an existing issue. I'll come back with ready, issue-linked PRs instead.

PR #18846, opened 2026-07-02
fix(core): handle leading step start in model content

### LINKED ISSUE #18774: [BUG] modelContent(stepNumber) drops step 1 content and is off by one for a leading step-start
### Describe the Bug

`StepContentExtractor.extractStepContent` (`packages/core/src/agent/message-list/conversion/step-content.ts`), which backs `messageList.get.response.aiV5.modelContent(stepNumber)`, treats a leading `step-start` marker inconsistently. Assistant steps are stored beginning with a `step-start` marker, so it is a prefix that begins a step, not a separator between steps. The extractor records every `step-start` as a boundary, which makes step 1 read as empty and shifts every later step number by one.

The two internal branches contradict each other:

- `extractLastStep` reads content after the last marker (`slice(lastStepStart + 1)`), treating `step-start` as a step prefix.
- `extractFirstStep` returns `[]` when the first marker is at index 0, treating a leading `step-start` as an empty step 1.

### Reference proving intent

- The single-step round-trip invariant: for a message with exactly one step, `modelContent(1)` must equal `modelContent(-1)`. With a stored leading `step-start` (`["step-start", text]`), `modelContent(-1)` correctly returns the text but `modelContent(1)` returns `[]`.
- `extractLastStep` in the same file is the correct sibling (step-start is a prefix), which `extractFirstStep` and the boundary model contradict.
- `modelContent(stepNumber)` is consumed with positive step numbers at `packages/core/src/loop/workflows/agentic-execution/llm-execution-step.ts` (`modelContent(steps.length)`), so wrong per-step content reaches `StepResult.content`.

### Steps To Reproduce

For a stored assistant message with parts `["step-start", "STEP1", "step-start", "STEP2"]`:

- `modelContent(1)` returns `[]` (STEP1 content is dropped).
- `modelContent(2)` returns STEP1 content (off by one).
- `modelContent(-1)` returns STEP2 content (correct).

For a single-step message `["step-start", "ONLY"]`, `modelContent(1)` returns `[]` while `modelContent(-1)` returns the content.

### Expected Behavior

`modelContent(1)` returns step 1 content, `modelContent(2)` returns step 2 content, and for a single-step message `modelContent(1)` equals `modelContent(-1)`.

### Additional Context

Fix: do not count a leading `step-start` (index 0) as a boundary, since it begins the first step rather than separating steps. Separator-style data (no leading marker) is unaffected. I have a PR ready with unit tests.

tianma-if:
Fixes #18774.

This treats a leading step-start marker as the prefix for step 1 instead of a separator before step 1, so modelContent(1) and later numbered steps are no longer shifted.

Tested with:

pnpm --filter ./packages/core test src/agent/message-list/conversion/step-content.test.ts

### DIFF
--- a/.changeset/orange-moles-reply.md
+++ b/.changeset/orange-moles-reply.md
@@ -1,0 +1,5 @@
+---
+'@mastra/core': patch
+---
+
+Fixed step-specific model content so leading step-start markers do not shift numbered steps.
--- a/packages/core/src/agent/message-list/conversion/step-content.test.ts
+++ b/packages/core/src/agent/message-list/conversion/step-content.test.ts
@@ -1,0 +1,49 @@
+import { describe, expect, it } from 'vitest';
+import { StepContentExtractor } from './step-content';
+
+describe('StepContentExtractor', () => {
+  const stepContentFn = (message?: { content?: unknown }) =>
+    Array.isArray(message?.content) ? message.content : message?.content ? [message.content] : [];
+
+  it('returns first step content when the message starts with a step-start marker', () => {
+    const uiMessages = [
+      {
+        id: 'msg-1',
+        role: 'assistant' as const,
+        parts: [{ type: 'step-start' as const }, { type: 'text' as const, text: 'Only step' }],
+      },
+    ];
+
+    expect(StepContentExtractor.extractStepContent(uiMessages, 1, stepContentFn)).toEqual([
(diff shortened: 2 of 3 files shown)


tianma-if:
Closing this draft to follow the maintainer guidance: no draft PRs, and PRs should be linked to an existing issue. I'll come back with ready, issue-linked PRs instead.

PR #18856, opened 2026-07-02
fix(core): honor readOnly memory in DurableAgent

### LINKED ISSUE #18771: [BUG] DurableAgent ignores readOnly memory config and persists messages
### Describe the Bug

A `DurableAgent` persists messages to memory even when the caller passes `memory.options.readOnly: true`. `readOnly` is documented as "read memory but won't save new messages" (`packages/core/src/memory/types.ts`), and the non-durable `Agent` honors it, but the durable path saves regardless. This is a correctness and privacy issue: ephemeral/preview runs and PII-sensitive threads get persisted against the caller's explicit instruction.

Root cause: the durable finish path saves via `saveQueueManager.flushMessages` directly, bypassing the `MessageHistory` output processor that enforces `readOnly` in the non-durable path, and its guard has no `readOnly` check:

```ts
// packages/core/src/agent/durable/workflows/create-durable-agentic-workflow.ts (map-final-output)
if (
  registryEntry?.saveQueueManager &&
  registryEntry.memory &&
  durableState?.threadId &&
  durableState?.resourceId &&
  !durableState.observationalMemory
  // no readOnly check
) {
  // ... saveQueueManager.flushMessages(...)
}
```

### Reference proving intent

- Non-durable sibling gates the whole persistence block on `!readOnlyMemory`: `packages/core/src/agent/agent.ts` `#executeOnFinish` (`if (memory && resourceId && thread && !readOnlyMemory)`).
- The `MessageHistory` output processor (the normal save mechanism) early-returns on `readOnly`: `packages/core/src/processors/memory/message-history.ts` (`const readOnly = memoryContext?.memoryConfig?.readOnly; if (!context || readOnly) return ...`).
- `memory.saveMessages` deliberately does not self-check `readOnly`; it delegates to processors, so the durable path (which bypasses them) must check it itself.
- Public contract: `packages/core/src/memory/types.ts` documents `readOnly: true` as "Agent can read memory but won't save new messages".

### Steps To Reproduce

Run a durable agent with a `MockMemory` and `readOnly: true`:

```ts
await durableAgent.stream('user input', {
  memory: { thread: 't', resource: 'r', options: { readOnly: true } },
})
// then: memory.recall({ threadId: 't', resourceId: 'r' })
```

### Expected Behavior

Nothing is saved; `recall` returns no new messages (matching the non-durable agent).

### Actual Behavior

Both the user input and the assistant response are persisted, silently violating `readOnly`.

### Additional Context

The same missing-guard pattern also exists in the durable suspension/background flush paths (`packages/core/src/agent/durable/workflows/steps/tool-call.ts` around the `flushMessagesBeforeSuspension` call and the background-task on-finish flush). This PR fixes the main finish path (which is the primary, always-hit case) with a regression test; happy to extend the same one-line guard to those two flush sites in this PR or a follow-up.

serhiizghama:
Fixes #18771. A `DurableAgent` persisted messages even when the caller passed `memory.options.readOnly: true`, so ephemeral/preview runs and PII-sensitive threads got written against the caller's explicit instruction.

The durable finish path (`create-durable-agentic-workflow.ts`, map-final-output) flushes via `saveQueueManager.flushMessages` directly, which bypasses the `MessageHistory` output processor that enforces `readOnly` in the normal save path — and its guard had no `readOnly` check. The non-durable `Agent#executeOnFinish` gates the whole persistence block on `!readOnlyMemory` (`readOnlyMemory: memoryConfig?.readOnly`), so I mirrored that by adding `!durableState.memoryConfig?.readOnly` to the guard. Added a regression test alongside the existing persist-after-completed-stream one — it fails without the guard (user + assistant get saved) and passes with it.

I left the suspension/background flush sites in `steps/tool-call.ts` alone on purpose: the non-durable sibling (`loop/workflows/agentic-execution/tool-call-step.ts`) also flushes before suspension regardless of `readOnly`, so gating those would diverge from parity rather than restore it. Happy to revisit them separately if you'd like readOnly to also suppress mid-run durability flushes.

## ELI5
If you tell `DurableAgent` to use memory in **read-only** mode, it shouldn’t secretly save new conversation messages while the run is paused for tool approval or while background tasks finish. This PR adds those missing “don’t write” checks so the saved thread stays empty and `recall()` can’t see any new messages.

## Summary

- Adds read-only guards in `packages/core/src/agent/durable/workflows/steps/tool-call.ts` to prevent durable persistence writes during:
  - tool-call suspension flows (approval / in-execution suspend) by skipping `saveQueueManager.flushMessages(...)` when `memoryConfig.readOnly` is set
  - background-task completion flushes (result/progress paths) by gating `saveQueueManager.flushMessages(...)` the same way
- Adds regression coverage in `packages/core/src/agent/durable/__tests__/durable-agent-tool-suspension.test.ts`:
  - verifies a read-only durable run emits the suspension/approval chunk
  - verifies `memory.recall({ threadId, resourceId })` returns `messages: []` (no message writes occurred)
- Includes a patch changeset for `@mastra/core` describing the durable mid-run readOnly fix follow-up to the earlier durable finish-path fix.


minh2416294 (commented):
﻿Hey, your core fix looks right.
One thing I wanted to check with you: the terminal flush is now gated, but the mid-run flushes still aren't. `saveQueueManager.flushMessages` at tool-call.ts:816-817 (background-task completion) and the `doFlush()` calls at 436/603/642 (suspend/approval) don't consult `memoryConfig?.readOnly`. So a `readOnly:true` run that suspends for approval or finishes a background tool would still write to the thread. Given the changeset says these runs are "no longer written to the thread," is it worth either gating those too or scoping the wording to the final-persistence path? I'm happy to add a suspend-then-readOnly test case if that helps pin the intended behavior.

(Also looks like it's picked up a merge conflict since you opened it, i think it  probably worth a rebase on main to get CI green.)

serhiizghama:
Good catch, thanks. Turns out the core finish-path fix already landed separately in #18921 while this sat open, so I merged main and dropped the now-duplicate bit. Went ahead and gated the two spots you pointed at — the suspend/approval flush in `flushMessagesBeforeSuspension` and both bg-task result/progress flushes — same readOnly check as the finish path. Added a test for each, verified by reverting the guard and confirming real message content leaks into recall() without it. Should be good for another look.

abhiaiyer91 (approved):
**Verdict: approve**

### What this PR does

Follow-up to #18921 (which gated the durable finish path). That fix closed the always-hit persistence site in `create-durable-agentic-workflow.ts`, but three mid-run flush sites in `packages/core/src/agent/durable/workflows/steps/tool-call.ts` still wrote straight to the thread via `saveQueueManager.flushMessages`, bypassing the `MessageHistory` output processor that is the only place `readOnly` is normally enforced. This PR adds the same guard to all three, closing out issue #18771.

### Correctness

- The guard placement is right. `flushMessagesBeforeSuspension` is the single funnel for all suspend/approval flushes in this file — all four `doFlush()` call sites route through it, so the one early-return covers approval suspension, general `suspend()` suspension, and the metadata-removal flush on resume. The two background-task callbacks (`onResult`, `onExecution`) are the only remaining direct `flushMessages` calls, and both are now gated. Grepping the file confirms no unguarded write path remains in the durable tool-call step.
- Skipping the whole function (rather than only the flush) is correct rather than over-broad: the thread-creation branch inside it is a fallback — `preparation.ts` already creates the thread during `prepare()` when memory + threadId + resourceId are present — so nothing downstream loses the thread record. The `threadExists` bookkeeping is only consumed by that same fallback.
- Resume is unaffected: durable resume is driven by the workflow snapshot and run registry, not by persisted messages. Suspended-run *discovery* (which reads `suspendedTools`/`pendingToolApprovals` from stored message metadata, per #18940) will not find readOnly runs — but that is inherent to "read memory but don't save new messages" and is already true of the finish-path guard merged in #18921. Consistent, not a new gap.
- `readOnly` semantics match the documented contract in `memory/types.ts` and the deliberate design note in `memory/memory.ts` (`saveMessages` intentionally does not self-check `readOnly`; enforcement belongs at the processor / call site), so a path that bypasses processors must check it itself.

### Tests

Both added tests are real regressions, verified by revert-and-run locally: with the three guards reverted, `does not flush messages or create a thread when suspending a readOnly run` and `does not flush the bg task result to memory when the run is readOnly` both fail with actual conversation content (tool-invocation parts, user text) surfacing through `recall()`. Restored, the full `src/agent/durable` suite passes (56 files, 547 tests) and `tsc --noEmit` on `@mastra/core` is clean.

Nit (non-blocking): the suspension test's title says "or create a thread", but it only asserts `recalled.messages` is empty, and its own inline comment explains the thread *is* created in `prepare()`. The title overstates what is asserted.

### Scope & consistency

Four files, one concern; changeset present and correctly scoped to `@mastra/core` as a patch. The earlier finish-path commit was dropped after main was merged in, so the diff is now exactly the remaining gap. The only failing checks are the two Vercel preview deploys, which are blocked on team authorization for an external contributor, not on this code.

Follow-up worth filing (not blocking this PR): the non-durable sibling `loop/workflows/agentic-execution/tool-call-step.ts` still flushes before suspension without a `readOnly` check (its local `flushMessagesBeforeSuspension` guards only on `!saveQueueManager || !threadId`). After this PR the durable path is *stricter* than the non-durable one. Given the documented contract the durable behavior is the correct one and the non-durable path carries the residual bug — but the parity gap should be tracked rather than left implicit.

serhiizghama:
Thanks for the thorough review. Fixed the test title nit — dropped the "or create a thread" part since it only asserts the messages are empty.

Filed #20341 for the non-durable sibling gap you flagged (`loop/workflows/agentic-execution/tool-call-step.ts`'s `flushMessagesBeforeSuspension` only guards on `!saveQueueManager || !threadId`, not readOnly) so it doesn't get lost.

PR #18866, opened 2026-07-02
fix(deployer): bundle transitive workspace dependencies

### LINKED ISSUE #18816: [BUG] mastra build drops transitive workspace dependencies (dep-of-a-dep) from the output bundle
## Describe the Bug

In a pnpm monorepo where the Mastra app imports a workspace package that itself
depends on another workspace package, `mastra build` omits the **transitive**
workspace package (the dependency-of-a-dependency) from the output.

The build reports success, but the deepest workspace package is left in the
bundle as an unresolved bare `import` and is **not** added to the generated
`.mastra/output/package.json` dependencies. Starting the built server therefore
crashes immediately with `ERR_MODULE_NOT_FOUND`.

The bundler follows exactly **one** level of workspace→workspace nesting: the
directly-imported workspace package and its direct workspace deps are bundled,
but anything two or more levels below the Mastra entry is dropped.

`mastra dev` works fine — the dropped package still resolves through the pnpm
`node_modules` symlinks at runtime. The failure is specific to `mastra build`,
because the standalone build output has no such symlink and never bundled the
package.

Reproduces with the **default** bundler config and also with
`bundler: { externals: true }`.

## Steps To Reproduce

Minimal reproduction (see linked repo — `pnpm install` then the commands below):

Workspace layout — a Mastra app plus a 3-package chain, where the app only ever
imports `@repo/pkg-a`, and `@repo/pkg-c` is reachable only transitively:

```
packages/
  app/      # Mastra entry, imports @repo/pkg-a
  pkg-a/    # imports @repo/pkg-b   (directly imported by the app)
  pkg-b/    # imports @repo/pkg-c
  pkg-c/    # leaf: export function greetFromC(name) { ... }   (transitive only)
```

```bash
pnpm install

# 1. dev works — prints "[startup] greeting = HELLO MASTRA, GREETINGS FROM @REPO/PKG-C!"
pnpm --filter @repo/app dev

# 2. build succeeds
pnpm --filter @repo/app build

# 3. running the build crashes
node packages/app/.mastra/output/index.mjs
```

Step 3 output:

```
Error [ERR_MODULE_NOT_FOUND]: Cannot find package '@repo/pkg-c'
imported from .../packages/app/.mastra/output/index.mjs
    at Object.getPackageJSONURL (node:internal/modules/package_json_reader:314:9)
    ...
  code: 'ERR_MODULE_NOT_FOUND'
```

`packages/app/.mastra/output/mastra.mjs` shows `@repo/pkg-a` and `@repo/pkg-b`
inlined, but `@repo/pkg-c` left as a bare, unbundled import:

```js
import { greetFromC } from '@repo/pkg-c';   // not bundled, not in output package.json
function greetFromB(name) { return greetFromC(name).toUpperCase(); }   // pkg-b, inlined
function makeGreeting(name) { return greetFromB(name); }               // pkg-a, inlined
```

https://github.com/segevfiner/mastra-transitive-ws-repro

## Expected Behavior

The built server should start and behave like `mastra dev`. `@repo/pkg-c` should
be bundled/transpiled into the output (or listed in the output
`package.json` dependencies) just like `@repo/pkg-a` and `@repo/pkg-b`, so that
running the build prints:

```
[startup] greeting = HELLO MASTRA, GREETINGS FROM @REPO/PKG-C!
```

## Environment Information

```
System:
  OS: macOS 26.5.1
  CPU: (10) arm64 Apple M1 Pro
Binaries:
  Node: 22.22.2
  npm: 11.17.0
  pnpm: 11.5.1
npmPackages:
  mastra: 1.17.0
  @mastra/core: 1.48.0
  @mastra/loggers: 1.2.0
```

- Package manager: pnpm workspaces (pnpm 11.5.1)
- LLM Provider: n/a (reproduction does not call a model)

## Verification

segevfiner:
Fixes #18816.

`mastra build` was dropping transitive workspace packages from the output bundle. If your Mastra app imported a workspace package that depended on another workspace package the app didn't import directly (app -> `@repo/a` -> `@repo/b` -> `@repo/c`), the deeper packages were left out. The build succeeded but the server crashed at runtime with `ERR_MODULE_NOT_FOUND`. It only worked in `mastra dev` because the missing package still resolved through the pnpm `node_modules` symlinks.

Two things were going on in the transitive dependency walk in `analyzeEntry.ts`. A discovered package was marked "processed" the moment it was found, so it got skipped before we ever looked at its own dependencies (discovery stopped one hop in). And transitive deps were resolved relative to the top-level entry, which can't see a dependency-of-a-dependency in strict pnpm. Now a package is marked processed when it's actually analyzed, and each transitive dep is resolved relative to the file that imports it.

Before, `@repo/c` ended up as a bare `import '@repo/c'` in the output and wasn't installed, so it threw at startup. After, it's bundled/transpiled like the rest of the chain and the server boots.

Repro: https://github.com/segevfiner/mastra-transitive-ws-repro

Heads up for the reviewer: this change was AI-generated (Claude Code). I purposefully left some of the explanatory comments the AI wrote in the diff so you can follow what it did — feel free to keep or drop whichever ones you want.

P.S. This whole custom bundling flow feels like it could be extracted into its own tool/rollup plugin at some point, this problem of packaging a package from a workspace with proper externalizing of even transitive deps is more general then Mastra. Though the current code includes a some Mastra specific hacks and logic...

## ELI5
`mastra build` was sometimes leaving out packages that were only needed “through another package” in a pnpm workspace. This fix teaches the build to follow those dependency chains all the way down, so the server doesn’t crash later with missing-module errors.

## Summary
- Fixed transitive workspace dependency discovery in `packages/deployer/src/build/analyze/analyzeEntry.ts` so deeper workspace packages aren’t skipped.
- Resolved transitive dependencies using the correct importer context (the importing file’s resolved URL), instead of always using the top-level entry.
- Updated the “processed” tracking so a workspace dependency is marked as done only after it’s actually analyzed.
- Added a nested pnpm-workspace (3-level) fixture and a new regression test to cover the multi-hop chain under strict pnpm-like resolution.
- Updated existing analyzer test expectations/caching assertions to reflect the corrected discovery behavior.
- Added a changeset documenting the `mastra build` runtime `ERR_MODULE_NOT_FOUND` scenario and the fix.

### DIFF
--- a/.changeset/swift-rings-poke.md
+++ b/.changeset/swift-rings-poke.md
@@ -1,0 +1,9 @@
+---
+'@mastra/deployer': patch
+---
+
+Fixed `mastra build` and `mastra dev` dropping transitive workspace packages from the output bundle.
+
+When a workspace package depended on another workspace package that your Mastra app did not import directly (e.g. app → `@repo/a` → `@repo/b` → `@repo/c`), the deepest packages were left out of the build. The build succeeded but the server crashed at runtime with `ERR_MODULE_NOT_FOUND`. It worked in `mastra dev` only because the missing package was still resolvable through the pnpm `node_modules` symlinks.
+
+The dependency analyzer now walks the full workspace dependency graph instead of stopping one level deep, and resolves each transitive workspace package relative to the package that imports it (so it works with strict pnpm layouts).
--- a/packages/deployer/src/build/analyze/__fixtures__/nested-workspace/apps/mastra/src/deep-transitive.ts
+++ b/packages/deployer/src/build/analyze/__fixtures__/nested-workspace/apps/mastra/src/deep-transitive.ts
@@ -1,0 +1,7 @@
+import { level1 } from '@internal/level1';
+
+// The entry imports ONLY @internal/level1. @internal/level2 (one hop) and
+// @internal/level3 (two hops) are reachable only transitively.
+export const fn = () => {
+  return `${level1}`;
+};
--- a/packages/deployer/src/build/analyze/__fixtures__/nested-workspace/packages/level1/package.json
+++ b/packages/deployer/src/build/analyze/__fixtures__/nested-workspace/packages/level1/package.json
@@ -1,0 +1,10 @@
+{
+  "name": "@internal/level1",
+  "version": "1.0.0",
+  "description": "",
+  "type": "module",
+  "main": "src/index.ts",
+  "dependencies": {
+    "@internal/level2": "workspace:*"
+  }
+}
--- a/packages/deployer/src/build/analyze/__fixtures__/nested-workspace/packages/level1/src/index.ts
+++ b/packages/deployer/src/build/analyze/__fixtures__/nested-workspace/packages/level1/src/index.ts
@@ -1,0 +1,3 @@
+import { level2 } from '@internal/level2';
+
+export const level1 = `level1 ${level2}`;
--- a/packages/deployer/src/build/analyze/__fixtures__/nested-workspace/packages/level2/package.json
+++ b/packages/deployer/src/build/analyze/__fixtures__/nested-workspace/packages/level2/package.json
@@ -1,0 +1,10 @@
+{
+  "name": "@internal/level2",
+  "version": "1.0.0",
+  "description": "",
+  "type": "module",
+  "main": "src/index.ts",
+  "dependencies": {
+    "@internal/level3": "workspace:*"
+  }
+}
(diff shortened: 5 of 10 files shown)


wardpeet:
Thank you for the pull request, but it isn't fully correct. We need to update our end-to-end tests for this. I will submit a new pull request to address this issue. Thank you for your contribution, though.

segevfiner:
That's fine. It was an attempt to let AI fix it as the disclaimer in the PR says. I wonder if there is some dependency walker code that Mastra can use instead of inventing its own, as is evident, its not all that easy with all the various shenanigans Node.js package managers do.

PR #18870, opened 2026-07-03
fix(agent-browser): prevent unhandled rejection from dangling navigation wait

### LINKED ISSUE #18868: [BUG] @mastra/agent-browser: dangling waitForNavigation rejection crashes the whole process when click/press/select fails first
## Describe the Bug

`AgentBrowser.click` / `press` / `select` start the `waitUntil` navigation wait **before** awaiting the action, and no rejection handler is attached to the navigation promise until `await navigation` is reached:

```ts
// browser/agent-browser/src/agent-browser.ts (click; press/select have the same shape)
const navigation = input.waitUntil ? page.waitForNavigation({ waitUntil: input.waitUntil, timeout }) : undefined;

await locator.click({ ... });  // ← while this is pending, `navigation` has no handler

await navigation;
```

If the navigation wait rejects while the action is still pending (or after the action threw and control jumped to the `catch` block), the rejection is **unhandled**, and with Node's default `--unhandled-rejections=throw` behavior it **kills the entire process** — in our case a `mastra dev` server hosting all agents went down mid-request.

What makes this practically likely rather than exotic: the navigation wait and the action **share the same `timeout` value**, and the navigation timer starts first. So whenever the action itself is slow or blocked long enough to time out (e.g. a click blocked by an overlay, or an element detached mid-navigation), the navigation timer always fires a moment earlier — while `await navigation` has not been reached. The tool call returns a graceful error object, but the process dies anyway.

Observed in production-like use via `webResearchAgent`-style setups: stack trace pointed at `AgentBrowser.click` → `Tool.execute` → tool-builder `setImmediate`, followed by the process exiting (`exit-hook`'s "SYNCHRONOUS TERMINATION NOTICE" from @mastra/mcp's cleanup hooks confirmed a synchronous process exit).

## Steps to Reproduce

Deterministic repro without mastra — serve a page whose link is covered by a transparent overlay (so the click waits for actionability until timeout), then click with `waitUntil`:

```ts
import http from 'node:http';
import { AgentBrowser } from '@mastra/agent-browser';

process.on('unhandledRejection', r => console.log('!!! UNHANDLED REJECTION:', String(r).slice(0, 120)));

const server = http.createServer((req, res) => {
  if (req.url === '/') {
    res.setHeader('content-type', 'text/html');
    res.end(`<a href="/slow">go</a><div style="position:fixed; inset:0"></div>`);
  } // "/slow" never responds
});
await new Promise(r => server.listen(0, '127.0.0.1', () => r(null)));
const port = server.address().port;

const browser = new AgentBrowser({ headless: true, scope: 'thread', timeout: 1500 });
await browser.goto({ url: `http://127.0.0.1:${port}/` }, 't1');
const snap = await browser.snapshot({ interactiveOnly: true }, 't1');
const ref = JSON.stringify(snap).match(/@e\d+/)[0];

const result = await browser.click({ ref, waitUntil: 'domcontentloaded' }, 't1');
console.log('tool returned gracefully:', JSON.stringify(result).slice(0, 120));
await new Promise(r => setTimeout(r, 2500)); // let the dangling rejection fire
```

Output:

```
!!! UNHANDLED REJECTION: TimeoutError: page.waitForNavigation: Timeout 1500ms exceeded.
tool returned gracefully: {"success":false,"code":"element_blocked","message":"Element @e1 is blocked by another element." ...
```

The tool call handled its own error, but the dangling navigation rejection escapes — in any process without a global `unhandledRejection` handler (like the mastra dev/server process) this is fatal.

## Expected Behavior

A failing browser action should surface as the tool's error result only. The navigation wait's rejection should never escape as an unhandled rejection / process crash.

## Environment Information

- `@mastra/agent-browser`: 0.4.0 (latest; same code on main at `browser/agent-browser/src/agent-browser.ts`)
- `@mastra/core`: 1.48.0
- Node.js: 22.x (default unhandled-rejections behavior)
- Affected methods: `click`, `press`, `select` (all three `waitUntil` sites)

## Proposed Fix

Attach a noop handler right after starting the navigation wait, so the rejection is always observed while remaining visible to the later `await navigation`:

```ts
const navigation = input.waitUntil ? page.waitForNavigation({ waitUntil: input.waitUntil, timeout }) : undefined;
navigation?.catch(() => {}); // prevent an unhandled rejection while the action is pending

await locator.click({ ... });
await navigation; // still throws the original error into the existing try/catch
```

I have a PR ready with this fix for all three sites plus a regression test.

## Verification

mstysk:
## Description

Fixes #18868

`AgentBrowser.click` / `press` / `select` start the `waitUntil` navigation wait **before** awaiting the action, and no rejection handler is attached to the navigation promise until `await navigation` is reached:

```ts
const navigation = input.waitUntil ? page.waitForNavigation({ waitUntil: input.waitUntil, timeout }) : undefined;
await locator.click({ ... });  // ← while this is pending, `navigation` has no handler
await navigation;
```

Because the navigation wait and the action **share the same `timeout`** and the navigation timer starts first, any action that is slow or blocked long enough to time out lets the navigation rejection fire first — while `await navigation` has not been reached. The rejection is unhandled and, with Node's default behavior, **kills the entire process** (in our case a `mastra dev` server hosting all agents went down mid-request), even though the tool call itself returned a graceful error object.

## Fix

Attach a noop handler right after starting the navigation wait in all three sites (`click`, `press`, `select`):

```ts
const navigation = input.waitUntil ? page.waitForNavigation({ waitUntil: input.waitUntil, timeout }) : undefined;
navigation?.catch(() => {});
```

The rejection is always observed, and `await navigation` still surfaces the original error through the existing try/catch, so error behavior of the tools is unchanged.

## Testing

- Added a regression test that drives `click` with `waitUntil` while the navigation wait rejects before the click settles, asserting no `unhandledRejection` escapes.
  - Verified the test **fails without the fix** (unhandled rejection observed) and **passes with it**.
  - Note: the test replaces the `waitForNavigation` mock with a plain function for this case — vitest attaches handlers to promises returned from `vi.fn()` mocks (for `mock.settledResults`), which would mask the unhandled rejection.
- Full package suite: `browser/agent-browser` — 12 files / 221 tests pass.
- `tsc --noEmit` and `eslint` clean for the package.
- Also reproduced the original crash outside tests with a real browser (link covered by an overlay + `waitUntil: 'domcontentloaded'`): the tool returned `element_blocked` gracefully while the dangling navigation rejection escaped as an unhandled rejection.

## Changeset

Added a `patch` changeset for `@mastra/agent-browser`.

## ELI5
When the browser tool does a **click**, **press**, or **select**, it also waits for the page navigation to finish. This fix ensures that if navigation fails first, the error is still “caught” so Node won’t crash—while the tool still returns the normal failure result.

## Summary
- Refactored `click`, `press`, and `select` to use a shared `startNavigationWait(...)` helper that immediately attaches a noop `navigation?.catch(() => {})` right after creating the `waitForNavigation` promise.
- Kept the existing behavior of awaiting the navigation promise after the DOM action, preserving the usual tool error surface.
- Added regression tests for **click**, **press**, and **select** that simulate a “dangling navigation wait” (navigation rejects while the action is still pending) and assert no `unhandledRejection` escapes.
- Added a Changesets entry for a patch release of `@mastra/agent-browser`.


mstysk:
Addressed all CodeRabbit feedback in the latest commit:

- **Changeset**: rewritten to focus on the user-facing outcome (crash → normal error result) instead of the internal fix mechanism
- **Helper extraction**: added a `startNavigationWait` private helper shared by `click`/`press`/`select`, so future navigation-based methods can't forget the noop-catch pattern
- **Test symmetry**: added the same dangling-navigation regression test for `press` and `select`

Package checks after the changes: 12 files / 223 tests pass, `tsc --noEmit` and `eslint` clean.

abhiaiyer91 (approved):

TheIsrael1:
## 🟡 Medium: Maintainer notes

**Merge confidence:** 4/5 — mergeable with non-blocking follow-up

**Why:**
- The fix is focused on `@mastra/agent-browser`: it observes the navigation wait immediately while preserving the later `await navigation` behavior, which directly addresses the dangling `waitForNavigation` rejection crash path.
- Regression coverage now covers the affected `click`, `press`, and `select` flows, including the case where navigation rejects while the action is still pending.
- I pushed a formatting-only follow-up commit (`19bb2465f4`) to fix the failed lint/format check; local verification passes:
  - `pnpm exec prettier --check browser/agent-browser/src/__tests__/agent-browser.test.ts`
  - `pnpm --filter @mastra/agent-browser lint`
  - `pnpm --filter @mastra/agent-browser exec vitest run src/__tests__/agent-browser.test.ts` — 58 tests passed

**Why not higher:**
- CI is still finishing on the latest commit, including the rerun lint/build/test gates. This is blocking until the required checks complete.
- The Vercel docs/playground-ui checks are failing due to authorization-required deploy checks; I’m treating those as non-code/non-blocking for this package change.
- I could not run the full `@mastra/agent-browser` test suite locally because the non-mocked `profile-lifecycle.test.ts` cases require Playwright browser binaries missing from this environment (`chromium_headless_shell`). I verified the affected regression test file directly instead.

**Required before merge:**
- Wait for the approved GitHub workflows on `19bb2465f4` to complete successfully, especially lint/build/affected tests.

**Non-blocking follow-ups:**
- None.

**Author comments:**
- None.

**Prior maintainer notes:**
- This comment supersedes my earlier malformed note at https://github.com/mastra-ai/mastra/pull/18870#issuecomment-4876083065.

PR #18874, opened 2026-07-03
feat!: rename heartbeats to schedules and unify with workflow schedules

abhiaiyer91:
Renames the heartbeats feature to schedules before we promote it, and merges agent + workflow scheduling into one API. "Heartbeat" reads as a liveness/health check to most programmers, but the feature is cron-based agent scheduling — so now there's a single `mastra.schedules` surface (create/get/list/update/pause/resume/run/delete) that handles both agent and workflow targets, discriminated by `agentId` vs `workflowId`.

```ts
// before
await mastra.heartbeats.create({ agentId: 'chef', cron: '0 9 * * *', prompt: 'Morning check-in' });

// after — same API also covers workflow schedules
await mastra.schedules.create({ agentId: 'chef', cron: '0 9 * * *', prompt: 'Morning check-in' });
await mastra.schedules.create({ workflowId: 'digest', cron: '0 8 * * 1' });
```

Server-side, `/api/heartbeats` is merged into `/api/schedules` (10 routes). Client-js gains `createSchedule`/`updateSchedule`/`deleteSchedule`/`runSchedule`, and the old heartbeat methods are deprecated but still work by delegating to the schedules API. Existing stored rows keep working: a read-shim in all 8 store implementations normalizes legacy `target.type: 'heartbeat'` rows to `'agent'` at deserialization. New agent schedule ids use the `agent_` prefix (old `hb_` ids still resolve when passed in full). Docs moved from `agents/heartbeats` to `agents/schedules` with a redirect.

Verified with the core/server/client-js/store test suites (typecheck clean), plus an HTTP smoke test through the hono adapter covering full CRUD, pause/resume, manual run with trigger recording, legacy-row read-shim, and `/api/heartbeats` returning 404.

## ELI5
This PR renames “heartbeats” to “schedules” and uses one unified feature for both agent cron scheduling and workflow scheduling. It also keeps old saved heartbeat data working by automatically treating it as agent schedules, while moving the HTTP/SDK APIs and docs to the new `/api/schedules` + `mastra.schedules` system.

## Summary
- Renamed the heartbeats feature to **schedules** and introduced a unified `mastra.schedules` API for both **agent** and **workflow** scheduling.
- Added `mastra.schedules` operations: **create, get, list, update, pause, resume, run, delete**, distinguishing targets by **`agentId` vs `workflowId`**.
- Consolidated server routes: **removed `/api/heartbeats`** and added **`/api/schedules`**; legacy heartbeat endpoints now return **404**.
- Updated `@mastra/client-js` with new methods: **`createSchedule`, `updateSchedule`, `deleteSchedule`, `runSchedule`** and a schedule-shaped response model; kept old heartbeat client methods as **deprecated wrappers** delegating to the schedules API.
- Implemented backward compatibility in storage: on read, all store implementations normalize legacy persisted rows with `target.type: 'heartbeat'` into `target.type: 'agent'`, and legacy **`hb_`** IDs still resolve when provided fully.
- Introduced new schedule ID semantics: **agent schedules use `agent_` prefix** (workflow schedules use their new `schedule_` prefix).
- Updated runtime orchestration to use schedule workers/topics (including scheduler + agent-schedule startup behavior) and added the “worker startup race” fix by memoizing the in-flight scheduler worker startup promise (with a regression test).
- Tightened schedule create validation: `agentId`/`workflowId` now require `.min(1)` and create schemas are `.strict()` to prevent ambiguous payload routing (clear 400s).
- Improved hookless schedule handling: schedule hook context loading is only attempted when hooks exist; lookup failures are now debug-logged instead of silently swallowed.
- Migrated documentation and examples from **heartbeats → schedules**, including new pages, sidebar updates, and Vercel redirects.
- Updated/extended tests across core, server, client-js, and storage (including an HTTP smoke test for CRUD, pause/resume, manual run, legacy-row reads, and `/api/heartbeats` → 404).
- Bumped peer dependency ranges for affected packages to align with the updated `@mastra/core`.


packages/core/src/agent/index.ts
```diff
@@ -6,19 +6,28 @@ export * from './signals';
...
-  toHeartbeat,
-  type HeartbeatInput,
-  type HeartbeatOutput,
-  type HeartbeatRunStatus,
-  type Heartbeat,
-  type CreateHeartbeatInput,
-  type UpdateHeartbeatInput,
-  type ListHeartbeatsFilter,
-} from './heartbeat';
+  AGENT_SCHEDULE_PREFIX,
```
wardpeet, line 9: not sure if you want to alias them or not

stores/libsql/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@mastra/libsql",
-  "version": "1.15.0",
+  "version": "1.15.0-alpha.0",
```
wardpeet: this should not change

stores/mongodb/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@mastra/mongodb",
-  "version": "1.12.0",
+  "version": "1.12.0-alpha.0",
```
wardpeet: same

stores/mysql/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@mastra/mysql",
-  "version": "0.3.3",
+  "version": "0.3.2",
```
wardpeet: same

stores/pg/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@mastra/pg",
-  "version": "1.15.0",
+  "version": "1.15.0-alpha.0",
```
wardpeet: same

stores/spanner/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@mastra/spanner",
-  "version": "1.2.2",
+  "version": "1.2.1",
```
wardpeet: same

docs/src/content/en/docs/long-running-agents/schedules.mdx
```diff
@@ -0,0 +1,237 @@
...
+title: "Schedules | Agents"
+description: "Run an agent on a cron schedule to deliver a recurring prompt, with optional thread delivery and lifecycle hooks."
+packages:
+  - "@mastra/core"
+  - "@mastra/client-js"
+---
+
+# Schedules
+
+**Added in:** `@mastra/core@1.46.0`
```
LekoArts: Nit: While we introduced the "Heartbeats" API in .46, we'll release the API under the new name with .50. So let's update this so people don't try to use the new name on too old version

```suggestion
**Added in:** `@mastra/core@1.50.0`
```

packages/core/src/agent/index.ts
abhiaiyer91: AI says: We intentionally dropped the old `Heartbeat*` names here rather than aliasing. Rationale: the feature is still marked beta/experimental (breaking changes allowed without a major bump — same treatment as the interval-API rename in 1.48.0), and keeping aliases would leave two names for the same surface indefinitely. For HTTP consumers, backward compatibility is preserved where it matters: legacy client-js methods are kept as deprecated wrappers and storage has a read-shim for legacy `type: "heartbeat"` rows. Happy to add deprecated type aliases if you prefer, though. Mastra Code

.changeset/tame-twin-schedules-server.md
```diff
@@ -0,0 +1,10 @@
+---
+'@mastra/server': minor
+---
+
+**Merged `/api/heartbeats` into `/api/schedules`.** The server now exposes one unified schedules API that covers both agent schedules (previously heartbeats) and workflow schedules.
+
+- `GET /api/schedules` lists both kinds and supports `agentId`, `workflowId`, `status`, `threadId`, `resourceId`, and `name` filters.
+- `POST /api/schedules` creates an agent schedule (body with `agentId`, `cron`, `prompt`) or a workflow schedule (body with `workflowId`, `cron`).
+- `PATCH`, `DELETE`, and `POST .../pause`, `.../resume`, `.../run` work for both kinds.
+- The `/api/heartbeats/*` routes were removed. Use `/api/schedules/*` instead.
```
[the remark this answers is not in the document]
abhiaiyer91: Done in ce8785e6b6 — added a curl POST example for /api/schedules to the changeset.

packages/core/src/schedules/schedules.test.ts
```diff
@@ -0,0 +1,324 @@
...
+
+    await mastra.schedules.delete(hb.id);
+    await expect(mastra.schedules.delete(hb.id)).resolves.toBeUndefined();
+    expect(await mastra.schedules.get(hb.id)).toBeNull();
+  });
+
+  it('get returns null for unknown ids and for workflow-target rows it cannot match', async () => {
+    const { mastra } = makeMastra(['a']);
+    expect(await mastra.schedules.get('agent_nope')).toBeNull();
+  });
```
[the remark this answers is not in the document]
abhiaiyer91: Done in 07b8b4d66e — the test now also asserts that an agent-prefixed lookup does not resolve a workflow-target row.

packages/playground/src/pages/workflows/schedule.tsx
```diff
@@ -64,7 +64,7 @@ export default function SchedulePage() {
     );
   }
 
-  const workflowId = schedule?.target.type === 'workflow' ? schedule.target.workflowId : undefined;
+  const workflowId = schedule?.workflowId;
```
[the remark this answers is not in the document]
abhiaiyer91: Done in 2c9b8b3d51 — the detail page now shows an "Agent" MetaItem (linking to the agent) for agent schedules and keeps the "Workflow" MetaItem for workflow schedules. Also renamed the list column header to "Target" in f6d1ab13c8.

stores/libsql/package.json
abhiaiyer91: Restored in cb439a1262 — version fields for libsql/mongodb/mysql/pg/spanner now match main; only the @mastra/core peer floor bump remains.

docs/src/content/en/docs/long-running-agents/schedules.mdx
abhiaiyer91: Fixed in 53e1083271 — now reads @mastra/core@1.50.0.

abhiaiyer91:
Addressed the remaining CodeRabbit outside-diff comments:

- **Worker startup race** — fixed in d11f79b063: `#ensureSchedulingWorkersStarted()` now memoizes the in-flight startup promise, so concurrent calls can no longer double-subscribe to the `agent-schedules` topic. Added a regression test that fails without the fix.
- **Create schema validation** — fixed in 2ca649604d: `agentId`/`workflowId` now require `.min(1)` and both create body schemas are `.strict()`, so ambiguous bodies (both ids) and empty ids are rejected with a clear 400 instead of misrouting.
- **Hookless schedule fires** — fixed in af59cf948f: the worker only loads the schedule row for hook context when schedule hooks are configured, and lookup failures are debug-logged instead of silently swallowed.
- **route-types agent fields** — already correct in the current commit: `GetSchedules_Response` includes `ifActive`, `ifIdle`, and `providerOptions` for agent schedules; no change needed.

All schedules test suites pass (core 86, server 46, playground full suite).

PR #18876, opened 2026-07-03
fix(core): decouple workflow schedules from the evented execution engine

### LINKED ISSUE #18807: Declaring `schedule` on a workflow silently switches ALL its runs to the evented engine, breaking HTTP-triggered execution on serverless
## Summary

In `@mastra/core` 1.42.0, adding the `schedule` field to a workflow makes `createWorkflow()` return an `EventedWorkflow` instead of a default `Workflow`:

```js
// dist/chunk-*.js (src/workflows)
function createWorkflow(params) {
  if (params.schedule) {
    return createWorkflow2(params); // EventedWorkflow + EventedExecutionEngine
  }
  return new Workflow(params);      // DefaultExecutionEngine (in-process)
}
```

This changes the execution engine for **every** run of that workflow, not just scheduler-fired runs. Runs started via `workflow.createRun()` + `run.start()` from a custom HTTP route (`registerApiRoute`) also go through the evented engine, which publishes events and depends on a live event processor to actually execute steps.

On a long-lived server (`mastra start`) that is fine, the processor is running. But on serverless (we hit this with `@mastra/deployer-vercel`), there is no event processor consuming the pubsub, so:

- `createRun()` persists the initial snapshot,
- the run stays in status `running` forever with **zero steps executed** (snapshot `context` stays empty),
- `await run.start()` never resolves, and the function dies at `maxDuration` (504),
- accumulated hung requests degrade the warm instance (subsequent requests on the same instance hang before producing any log output).

## What we expected

Either of:
- `schedule` only affects scheduler-fired runs (route/manual runs keep the default in-process engine), or
- a loud failure at `createRun()`/`start()` when the evented engine has no processor available in the current runtime.

Today the only guard we found is the `ATOMIC_STORAGE_OPERATIONS_NOT_SUPPORTED` error, which triggers on the storage capability (`supportsConcurrentUpdates`), not on the presence of an event processor. With `@mastra/pg` the check passes and the run silently hangs.

## Reproduction

1. Define a workflow with `schedule: { cron: "*/15 * * * *", timezone: "UTC", inputData: {...} }` and register it on the Mastra instance. Scheduler can be disabled (`scheduler: { enabled: false }`); the engine switch happens regardless.
2. Register a custom API route that does `mastra.getWorkflow(id)`, `workflow.createRun()`, `await run.start({ inputData })` (the classic cron-webhook pattern).
3. Deploy with `@mastra/deployer-vercel`, call the route.
4. Observe: HTTP 504, snapshot row in `mastra_workflow_snapshot` stuck at `"status": "running"` with empty `context`. Removing the `schedule` field (nothing else) makes the same route execute in-process and complete normally.

We reproduced this across ~30 production workflows on 2026-07-02 (40 orphaned `running` runs in one morning). Removing/conditionally declaring `schedule` restored the previous behavior immediately.

## Impact

Any project that keeps HTTP/cron-webhook triggers while adopting declarative schedules (e.g. migrating gradually from platform crons to the native scheduler) breaks in serverless the moment `schedule` lands on a workflow, with no error pointing at the cause.

## Workaround

Declare the schedule conditionally so serverless builds keep the default engine:

```ts
export function nativeSchedule<const T>(cfg: T): T | undefined {
  return process.env.MASTRA_SCHEDULER_ENABLED === "1" ? cfg : undefined;
}

createWorkflow({ id, inputSchema, outputSchema, steps, schedule: nativeSchedule({ cron: "*/15 * * * *", timezone: "UTC" }) })
```

## Environment

- `@mastra/core` 1.42.0
- `@mastra/pg` (Neon Postgres storage)
- `@mastra/deployer-vercel` (single serverless function)
- Node 24 / Vercel Fluid Compute

Rudra2637:
## Description

Previously, declaring a schedule on a workflow conditionally promoted it to an `EventedWorkflow` running on the `EventedExecutionEngine`. This forced all executions—including manual runs triggered via `run.start()` or HTTP endpoints—to use the evented engine. The evented engine relies on a pubsub queue and background event workers, which causes hangs and timeouts in serverless environments (such as Vercel) where long-running background workers do not exist.

This PR decouples scheduling configurations from the execution engine:
- `createWorkflow()` now always returns the standard `Workflow` class running on the in-process `DefaultExecutionEngine`.
- Schedule normalization, cron syntax checks, and duplicate ID validation have been moved to the base `Workflow` class constructor.
- Exposes schedule metadata via `.getScheduleConfigs()` on the base `Workflow` class so that the scheduler worker can still discover, register, and trigger workflow runs asynchronously via the pubsub queue.
- Added a regression test to verify that manual runs for scheduled workflows execute synchronously in-process via the `DefaultExecutionEngine`.

## Related issue(s)

Fixes https://github.com/mastra-ai/mastra/issues/18807

## Type of change

- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [ ] Test update

## Checklist

## ELI5
Before, just adding a schedule to a workflow could accidentally make it run in a different way. Now, scheduled workflows keep using the normal in-process runner, while the scheduler can still see the schedule and trigger it when needed.

## Summary
- Stopped `createWorkflow()` from auto-promoting scheduled workflows to the evented engine.
- Moved schedule parsing, cron validation, and duplicate/missing schedule ID checks into the base `Workflow` constructor.
- Added `getScheduleConfigs()` so schedule metadata is available to the scheduler worker.
- Updated tests to verify scheduled workflows still use the default engine for manual runs and expose schedule configs correctly.
- Added a changeset for `@mastra/core`.

### DIFF
--- a/.changeset/tame-dolls-rest.md
+++ b/.changeset/tame-dolls-rest.md
@@ -1,0 +1,5 @@
+---
+'@mastra/core': patch
+---
+
+Fixed workflow schedule engine promotion. Manual executions of scheduled workflows now run on the default in-process engine rather than evented engine, preventing hangs on serverless environments.
--- a/packages/core/src/workflows/create.ts
+++ b/packages/core/src/workflows/create.ts
@@ -31,18 +31,6 @@ export function createWorkflow<
   TSteps extends Step<string, any, any, any, any, any, DefaultEngineType>[] = Step[],
   TRequestContextSchema extends PublicSchema<any> | undefined = undefined,
 >(params: CreateWorkflowParams<TWorkflowId, TStateSchema, TInputSchema, TOutputSchema, TSteps, TRequestContextSchema>) {
-  if (params.schedule) {
-    return createEventedWorkflowImpl(params as any) as unknown as Workflow<
-      DefaultEngineType,
-      TSteps,
-      TWorkflowId,
-      InferSchemaOutput<TStateSchema>,
-      InferPublicSchema<TInputSchema>,
-      InferPublicSchema<TOutputSchema>,
-      InferPublicSchema<TInputSchema>,
-      InferSchemaOutput<TRequestContextSchema>
-    >;
-  }
   return new Workflow<
     DefaultEngineType,
     TSteps,
--- a/packages/core/src/workflows/schedule-promotion.test.ts
+++ b/packages/core/src/workflows/schedule-promotion.test.ts
@@ -2,8 +2,8 @@ import { describe, expect, it } from 'vitest';
 import { z } from 'zod/v4';
 import { createWorkflow as createDefaultWorkflow, createStep } from './index';
 
-describe('createWorkflow (default) — schedule promotion to evented', () => {
-  it('returns an evented workflow when a schedule is declared', () => {
+describe('createWorkflow (default) — schedules', () => {
+  it('retains default engine when a schedule is declared but exposes configs', () => {
     const step = createStep({
       id: 'noop',
       inputSchema: z.object({}),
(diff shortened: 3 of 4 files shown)


Rudra2637:
Hi! Just following up on this PR. Whenever you have a chance, I'd really appreciate a review. Thanks!

abhiaiyer91:
going to go in a diff direction. ty for opening this though

PR #18882, opened 2026-07-03
fix(core,observability): preserve tracingOptions.metadata across resume

### LINKED ISSUE #18875: @mastra/observability: tracingOptions.metadata (including sessionId) is silently dropped for child spans, causing orphaned Langfuse traces with no sessionId
## Bug description

`@mastra/observability` version **1.10.0** silently drops `tracingOptions.metadata` (which includes `sessionId`, `userId`, `tags`, etc.) for any span that has a parent. This means pick/input-form continuation turns — where Mastra creates a new agent run as a child span of an existing trace context — are exported to Langfuse **without a `sessionId`**, making them invisible to any query that filters by `sessionId`.

## Root cause

In `@mastra/observability/dist/index.js` (line 19744 in the built output):

```js
const tracingMetadata = !options.parent ? tracingOptions?.metadata : void 0;
```

The condition `!options.parent` means metadata (and therefore `sessionId`, `userId`, tags, etc.) is **only applied to root spans**. When a continuation turn is created as a child span (i.e. `options.parent` is set), `tracingMetadata` becomes `void 0` and the entire `tracingOptions.metadata` object is discarded.

Same pattern applies to `tags` two lines below:
```js
const tags = !options.parent ? tracingOptions?.tags : void 0;
```

## How we discovered it

We built a product-instrumentation pipeline that queries Langfuse for all traces belonging to a session (`GET /api/public/traces?sessionId=<threadId>`). For sessions that include pick/input-form widget interactions, the MessageHistory span accumulates more blocks than Langfuse returns traces — because the pick/input-form continuation traces exist in Langfuse **without** `sessionId` set.

**Concrete example:**
- Session `dc05cb8e-f7e3-4ff8-8c2e-4d878458633e` has 26 agent turns (MH span blocks)  
- Langfuse returns only **22 traces** when filtering by `sessionId=dc05cb8e-…`  
- Querying by `userId` (orgId) in the same time window finds **6 additional traces with no sessionId** — exactly the orphaned pick/input-form continuation turns

## Steps to reproduce

1. Configure a Mastra agent with `tracingOptions: { metadata: { sessionId: threadId, userId: orgId } }`
2. Run a session that includes a `pick` or `input-form` client-side tool where the agent continues after the user submits a result
3. Query Langfuse for `GET /api/public/traces?sessionId=<threadId>` — the continuation traces are missing
4. Query Langfuse for `GET /api/public/traces?userId=<orgId>` in the same time window — the continuation traces appear there **without** `sessionId`

## Expected behaviour

`tracingOptions.metadata` (and `tags`) should be propagated to all spans that represent distinct Langfuse traces, regardless of whether they have an internal parent span reference. At minimum, `sessionId` should be inherited by child spans that start new Langfuse root traces.

## Proposed fix

Change the condition from:

```js
const tracingMetadata = !options.parent ? tracingOptions?.metadata : void 0;
```

to:

```js
const tracingMetadata = tracingOptions?.metadata;
```

Or, if dropping metadata for child spans is intentional to avoid duplication on *observation*-level spans, the fix should at least ensure that spans which create **new Langfuse root traces** (not child observations within an existing trace) always receive the metadata including `sessionId`.

## Environment

- `@mastra/observability`: 1.10.0  
- `@mastra/langfuse`: (same install)  
- Langfuse exporter maps `attributes["mastra.metadata.sessionId"]` → `session.id` on the trace (in `@mastra/langfuse/dist/index.js` line ~158)

intojhanurag:
## Description

Fixes `tracingOptions.metadata` (`sessionId`, `userId`) being silently dropped when an agent resumes after a pick/input-form suspension, causing orphaned Langfuse traces with no `sessionId`.

- Extend `WorkflowStateTracingContext` with `tracingMetadata` and `tracingTags` fields
- Persist span metadata and tags into the workflow snapshot at both default and evented engine suspend points
- Restore persisted metadata/tags into `resumeTracingOptions` on agent resume, workflow resume, and durable agent resume paths (caller-provided values override persisted)

**Scoped to `@mastra/core` only** — no observability or exporter changes needed. The `!options.parent` guard in `startSpan` does not affect resumed spans (they have no live parent object), so the fix is entirely about persisting and restoring metadata across the suspend/resume boundary.

## Related Issue(s)

Fixes #18875

## Type of Change

- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [ ] Test update

## Checklist

## ELI5

When an agent pauses to wait for user input and then continues, some “tracking info” (like sessionId and userId) was getting lost, so the next trace didn’t connect to the earlier one. This PR saves that tracking info when pausing and restores it when resuming, so tools like Langfuse can still group everything correctly.

## What changed

- Persist tracing metadata and tags in workflow snapshots at suspension points.
- Restore persisted metadata and tags when resuming agents, workflows, and durable agents.
- Preserve caller-provided tracing values, while filling unspecified fields from persisted values.
- Extend `WorkflowStateTracingContext` with optional tracing metadata and tags.
- Add regression tests covering restoration and caller-overrides.
- Include a patch changeset for `@mastra/core`.

## Testing

- Focused regression tests passed: 11/11.
- Broader `@mastra/core` tests still contain unrelated existing failures and type errors.

### DIFF
--- a/.changeset/shy-snails-invite.md
+++ b/.changeset/shy-snails-invite.md
@@ -1,5 +1,4 @@
 ---
-'@mastra/observability': patch
 '@mastra/core': patch
 ---
 
--- a/.changeset/shy-snails-invite.md
+++ b/.changeset/shy-snails-invite.md
@@ -1,0 +1,6 @@
+---
+'@mastra/observability': patch
+'@mastra/core': patch
+---
+
+Fixed tracingOptions.metadata (sessionId, userId) being silently dropped on agent resume. Metadata and tags are now persisted in the workflow snapshot during suspend and restored into tracingOptions when the agent resumes, ensuring Langfuse traces retain their sessionId across pick/input-form continuations.
--- a/observability/mastra/src/instances/base.ts
+++ b/observability/mastra/src/instances/base.ts
@@ -200,8 +200,8 @@ export abstract class BaseObservabilityInstance extends MastraBase implements Ob
       traceState = this.computeTraceState(tracingOptions);
     }
 
-    // Merge tracingOptions.metadata with span metadata (tracingOptions.metadata takes precedence)
-    const tracingMetadata = tracingOptions?.metadata;
+    // Merge tracingOptions.metadata with span metadata (tracingOptions.metadata takes precedence for root spans)
+    const tracingMetadata = !options.parent ? tracingOptions?.metadata : undefined;
     const mergedMetadata = metadata || tracingMetadata ? { ...metadata, ...tracingMetadata } : undefined;
 
     // Extract metadata from RequestContext
@@ -221,7 +221,8 @@ export abstract class BaseObservabilityInstance extends MastraBase implements Ob
         ? { ...(enrichedMetadata ?? {}), environment: this.#mastraEnvironment }
         : enrichedMetadata;
 
-    const tags = tracingOptions?.tags;
+    // Tags are only passed for root spans (no parent)
+    const tags = !options.parent ? tracingOptions?.tags : undefined;
 
     // Extract traceId and parentSpanId from tracingOptions for root spans (no parent)
     // These allow nested workflows to join the parent workflow's trace
--- a/observability/mastra/src/instances/base.ts
+++ b/observability/mastra/src/instances/base.ts
@@ -200,8 +200,8 @@ export abstract class BaseObservabilityInstance extends MastraBase implements Ob
       traceState = this.computeTraceState(tracingOptions);
     }
 
-    // Merge tracingOptions.metadata with span metadata (tracingOptions.metadata takes precedence for root spans)
-    const tracingMetadata = !options.parent ? tracingOptions?.metadata : undefined;
+    // Merge tracingOptions.metadata with span metadata (tracingOptions.metadata takes precedence)
+    const tracingMetadata = tracingOptions?.metadata;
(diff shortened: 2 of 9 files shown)


intojhanurag:
## 🟡 medium: Maintainer notes

**Merge confidence:** 4/5 — ready to merge once required CI is green or unrelated failures are waived

**Why:**
- PR directly fixes the reported Langfuse/session correlation break by allowing `tracingOptions.metadata` to flow into child spans and by persisting/restoring metadata/tags across agent and workflow suspend/resume.
- Implementation is consistent across non-durable agent resume, workflow resume, evented/default workflow snapshot persistence, and durable agent resume.
- Focused regression coverage passes locally: `pnpm --dir packages/core exec vitest run src/agent/__tests__/resume-span-tracing.test.ts` passed 11/11 tests.

**Why not higher:**
- Non-blocking: only the focused regression test was cleanly verified locally; the broader `@mastra/core test:unit` invocation currently surfaces unrelated existing failures/type errors, so final merge confidence still depends on relevant CI signal or maintainer waiver of unrelated checks.

**Required before merge:**
- Ensure required CI checks are green, or confirm any remaining failures are unrelated to this tracing change.

**Non-blocking follow-ups:**
- None.

**Author comments:**
- None. CodeRabbit nitpicks appear addressed in the latest PR state.

**Prior maintainer notes:**
- None found.

PR #18902, opened 2026-07-03
feat(playground): add studio thread rail

damien-schneider:
Adds a reusable MessageScroller and ThreadRail in playground-ui, then wires Studio chat to show one rail tick per user turn with in-view state, hover/focus previews, and click-to-scroll navigation.

The Studio rail now sits outside the chat message max-width, is hidden by a container query when the thread area cannot safely fit the rail, and uses a single animated preview shell with scale, opacity, translate, and blur transitions.

Validated with `pnpm --filter ./packages/playground-ui test -- src/ds/components/MessageScroller/message-scroller.test.tsx src/ds/components/MessageScroller/thread-rail-turns.test.ts`, `pnpm --filter ./packages/playground test:run src/lib/ai-ui/__tests__/thread.test.tsx`, `pnpm build:playground-ui`, `pnpm --filter ./packages/playground-ui typecheck`, `pnpm --filter ./packages/playground typecheck`, `pnpm --filter ./packages/playground-ui lint`, `pnpm --filter ./packages/playground lint`, `npx -y react-doctor@latest packages/playground-ui/src/ds/components/MessageScroller --verbose`, and `git diff --check`.

CodeRabbit CLI was installed, but local review could not run because it is not authenticated in this non-interactive environment.

## ELI5
Studio chat now has a slim “conversation timeline” next to the messages. As you read, the tick for the current message lights up; hover/focus shows a quick preview, and clicking jumps you to that spot in the conversation.

## Summary
- Added reusable `MessageScroller` primitives and a `ThreadRail` to `playground-ui`, powered by a `MessageScrollerProvider` that:
  - registers message elements and anchors
  - tracks active/in-view turns via `IntersectionObserver` (with fallback)
  - exposes imperative navigation (`scrollToStart`/`scrollToEnd`/`scrollToMessage`) and maintains correct scroll/visibility state
- Wired Studio’s thread UI to render one rail tick per displayable user turn using `buildThreadRailTurns`, compute active/in-view state for the correct tick, and support click-to-scroll navigation.
- Implemented hover/focus previews for rail ticks with a single animated “preview shell” (scale/opacity/translate/blur) and staged transitions (entering → visible → exiting), positioned using a new `useMeasuredAutoHeight` hook.
- Kept the rail outside the chat message max-width and hid it when space isn’t safe using container-query CSS.
- Added/updated supporting infrastructure and hardening:
  - new `ThreadRail` turn-view model builder (`thread-rail-turns`)
  - scroller context hooks/exports (`message-scroller-hooks`, barrel exports)
  - improved thread/message rendering integration (`thread.tsx`, `message-row.tsx`) with more defensive typing
  - expanded Vitest test coverage for scroller behavior, preview transitions, and turn building
- Added Changesets entries documenting the new `playground-ui` patch release with usage guidance for composing the provider + rail.


packages/playground-ui/src/ds/components/ThreadRail/thread-rail.tsx
```diff
@@ -0,0 +1,382 @@
...
+          currentTurn={previewState.currentTurn}
+          previousTurn={previewState.previousTurn}
+          phase={previewState.phase}
+          top={previewState.top}
+        />
+      )}
+    </nav>
+  );
+}
+
```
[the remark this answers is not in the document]
damien-schneider: tailwind v4 accept by default duration-360, no more need duration-[360ms]. More info here https://tailwindcss.com/docs/transition-duration.

packages/playground-ui/src/hooks/use-measured-auto-height.ts
```diff
@@ -0,0 +1,72 @@
+import * as React from 'react';
+
+const useIsomorphicLayoutEffect = typeof window === 'undefined' ? React.useEffect : React.useLayoutEffect;
```
mfrachet: don't we already have something like this somewhere else?

packages/playground/src/lib/ai-ui/messages/message-row.tsx
```diff
@@ -34,6 +33,12 @@ export interface MessageRowProps {
 
 type MessagePart = MastraDBMessage['content']['parts'][number];
 
+const isRecord = (value: unknown): value is Record<string, unknown> =>
```
mfrachet, line 36: 😬

mfrachet (changes requested):

packages/playground/src/lib/ai-ui/messages/message-row.tsx
damien-schneider: I added this to remove the "as" casts below, it is not ideal but better than before in my opinion.
If you think it’s better to keep the "as" casts, I’ll revert the changes. Just let me know

packages/playground-ui/src/hooks/use-measured-auto-height.ts
damien-schneider: We don't have it as a shared hook, but I'll extract it and change the other consumers even outside the scope

mfrachet (approved):

PR #18922, opened 2026-07-03
fix(memory): add transient flag to data-om-status part

### LINKED ISSUE #18869: [BUG] data-om-status part missing transient flag — pollutes persisted message history with standalone status-only messages
## Summary

The `data-om-status` Observational Memory data part is emitted without the `transient: true` flag that **every other OM lifecycle marker** carries. As a result, Mastra's `OutputWriter` treats each `data-om-status` snapshot as a standalone assistant message and persists it to `mastra_messages`, polluting message history and crowding out real conversation messages from paginated history reads.

## Affected versions

- `@mastra/core` 1.41.0
- `@mastra/memory` 1.20.2
- Verified the bug is **still present on `main`** (see source references below).

## Root cause

OM lifecycle markers are emitted through two code paths, but only one consistently sets `transient`:

**`observational-memory.ts` — status emit (the bug):**
```ts
const statusPart: DataOmStatusPart = {
  type: 'data-om-status',
  data: { /* window usage snapshot for polling UIs */ },
};
await writer.custom(statusPart).catch(() => {});   // ← no transient flag
```

**Same file — every other OM marker (correct):**
```ts
// Stream OM lifecycle markers as transient so the OutputWriter does not persist
// standalone data-only messages; OM persists the durable marker explicitly.
void writer.custom({ ...startMarker, transient: true }).catch(() => {});
```

Mastra's persistence gate (`@mastra/core`, the `OutputWriter` chunk handler) persists any `data-*` chunk that is **not** marked transient as a standalone assistant message:

```ts
if (chunk.type.startsWith("data-") && !("transient" in chunk && chunk.transient)) {
  messageList.add({
    role: "assistant",
    content: { format: 2, parts: [{ type: chunk.type, data: chunk.data }] },
  });
}
```

`data-om-status` is intended as an ephemeral snapshot for polling UIs — it is explicitly documented as something channels intentionally do **not** render (`packages/core/src/channels/om.ts`):
> The `data-om-status` part (window usage snapshot for polling UIs) lives in `@mastra/memory` and is intentionally absent — channels don't render it.

Unlike the other OM markers, the status emit has **no** corresponding `persistMarkerToStorage(...)` call — it was never meant to be durable. It is persisted purely because the `transient` flag is missing.

## Observed impact (real data)

On a production-shaped Postgres store with Observational Memory enabled (`scope: 'thread'`, default buffering), querying `mastra.mastra_messages` for standalone assistant rows whose single part is an OM data part:

```sql
SELECT (content::jsonb -> 'parts' -> 0 ->> 'type') AS sole_part_type, count(*) AS msg_count
FROM mastra.mastra_messages
WHERE role = 'assistant'
  AND jsonb_array_length(content::jsonb -> 'parts') = 1
  AND (content::jsonb -> 'parts' -> 0 ->> 'type') LIKE 'data-om-%'
GROUP BY sole_part_type ORDER BY msg_count DESC;
```

Result:

| sole_part_type | msg_count |
| --- | --- |
| `data-om-status` | 81 |
| `data-om-buffering-start` | 35 |
| `data-om-observation-start` | 1 |
| `data-om-observation-end` | 1 |

- 17 threads affected; the worst thread accumulated **100** status-only messages.
- These status-only messages occupy real rows in the pagination window, so a `listMessages({ page, perPage })` history read returns mostly empty status snapshots and real conversation turns disappear off the first pages. This is especially harmful for long-running threads.

> Note: the `data-om-buffering-start` / observation-start rows appear to come from `persistMarkerToStorage`, which injects markers into existing assistant messages. The `data-om-status` rows are the clear regression: they have no durable intent and are persisted solely due to the missing `transient` flag.

## Suggested fix

Add `transient: true` to the `data-om-status` emit, matching every other OM lifecycle marker:

```diff
- await writer.custom(statusPart).catch(() => {});
+ await writer.custom({ ...statusPart, transient: true }).catch(() => {});
```

This is consistent with the documented design intent (the marker is an ephemeral polling snapshot, not durable memory) and with how every sibling marker is already emitted.

## Reproduction

1. Configure an agent with `observationalMemory` enabled (any scope, default buffering).
2. Run a multi-turn conversation long enough for OM to emit status snapshots (they fire once per agent loop step).
3. Query `mastra.mastra_messages` for the thread:
   ```sql
   SELECT id, role,
          (content::jsonb -> 'parts' -> 0 ->> 'type') AS first_part_type,
          jsonb_array_length(content::jsonb -> 'parts') AS part_count
   FROM mastra.mastra_messages
   WHERE thread_id = '<thread_id>' AND role = 'assistant'
   ORDER BY "createdAt" DESC;
   ```
4. Observe standalone assistant rows whose only part is `data-om-status`.

## Workaround

Until fixed, consumers reading paginated history should filter out messages whose `parts` array contains only transient OM/status part types (e.g. `data-om-status`) before applying their page window.

aryanbarde80:
## Problem

The `data-om-status` Observational Memory data part is emitted without the `transient: true` flag that every other OM lifecycle marker carries. As a result, Mastra's `OutputWriter` treats each `data-om-status` snapshot as a standalone assistant message and persists it to `mastra_messages`, polluting message history and crowding out real conversation messages from paginated history reads.

## Root Cause

In `observational-memory.ts`, the status part is emitted via:
```ts
await writer.custom(statusPart).catch(() => {});
```

While every other OM marker uses `transient: true`:
```ts
void writer.custom({ ...startMarker, transient: true }).catch(() => {});
```

## Fix

Spread `statusPart` and add `transient: true` to the custom call, matching the pattern used by all other OM lifecycle markers.

Fixes #18869

## ELI5
This PR makes `data-om-status` behave like a “temporary status ping,” not a real chat message. So the system won’t save these progress snapshots into message history where they can clutter or push real conversation turns out of paginated reads.

## Summary
- Mark `data-om-status` as `transient: true` when emitting it during `emitProgress` for Observational Memory (when streaming via `ProcessorStreamWriter`).
- Prevents `OutputWriter` from persisting `data-om-status` snapshots as standalone assistant messages.
- Adds a Changesets entry to bump `@mastra/memory` with the documented fix (aligning `data-om-status` with other Observational Memory transient lifecycle markers).

### DIFF
--- a/.changeset/data-om-status-transient.md
+++ b/.changeset/data-om-status-transient.md
@@ -1,0 +1,5 @@
+---
+'@mastra/memory': patch
+---
+
+Mark data-om-status parts as transient so they are not persisted as standalone assistant messages
--- a/packages/memory/src/processors/observational-memory/observational-memory.ts
+++ b/packages/memory/src/processors/observational-memory/observational-memory.ts
@@ -2629,7 +2629,7 @@ ${formattedMessages}
       omDebug(
         `[OM:status] step=${stepNumber} msgs=${pendingTokens}/${threshold} obs=${currentObservationTokens}/${effectiveObservationTokensThreshold} gen=${record.generationCount}`,
       );
-      await writer.custom(statusPart).catch(() => {});
+      await writer.custom({ ...statusPart, transient: true }).catch(() => {});
     }
   }
 


intojhanurag (dismissed):

aryanbarde80:
Closing this for now — refocusing contributions on higher-impact, well-scoped unit test additions. May revisit later.

PR #18925, opened 2026-07-03
feat(playground): nested-workflow tree and live run-status columns in the workflows list

### LINKED ISSUE #18924: [FEATURE] Workflows list: nested-workflow tree, live running and pending-input (HITL) counts
## Problem

The Studio `/workflows` page is a flat table with a name/description text filter. Three things are invisible at list level:

1. **Composition** — workflows that orchestrate other workflows (`.then(childWorkflow)`) look identical to leaf workflows. In real projects the list grows into dozens of entries where entry-point pipelines and their sub-workflows are indistinguishable, so teams end up encoding structure into naming conventions (`prd-ship-product`, `prd-groom-product`, …).
2. **Activity** — there is no way to see that a workflow currently has runs in flight (or how many, since the same workflow can run several times in parallel) without opening each workflow's runs tab.
3. **Human-in-the-loop** — runs suspended at a `suspend()` point waiting for `resume()` are exactly the ones that need human attention, and today they are the least visible.

## Proposal

Enhance the workflows list with composition and live status, keeping the current table look:

```
  Name                          Description                Running   Pending input   Steps
▾ prd-ship-product ⧉2          Single entry: triage…       ● 2                         3
│  ├ prd-groom-product ⧉1      Feature lane: idea…                    ⏸ 1             19
│  │   └ use-case-arch (inline) Per-use-case fan-out…                                   —
│  └ prd-fix-product            Bug pipeline: evidence…                                 5
  eng-ship-ticket ⧉3           Ship one ticket end-to-end… ● 1                         9
```

- **Nested-workflow tree**: workflows composing other workflows get a muted count indicator and a chevron; expanding reveals child rows connected by tree guide lines (`├`/`└` style). Registered children link to their own workflow pages and expand recursively (cycle-safe); nested workflows that are not registered standalone render as inline, non-link rows. The same child under different parents expands independently.
- **Running column**: live count of `status=running` runs per workflow (pulsing dot + count, blank when idle), polled every few seconds.
- **Pending input column**: count of `status=suspended` runs — the HITL state — with a distinct pause indicator, so runs waiting for a human are visible at a glance and separated from actively running ones.
- **Fit fix**: the list currently overflows horizontally (the grid is `w-max`, so long descriptions force content width and push trailing columns off-screen). An additive `fit="container"` mode on `DataList` lets the list fill the viewport and truncate descriptions instead.

## Implementation notes (validated locally)

I have this working against a real project and I am happy to contribute it as PRs following your guidance:

- **No API changes needed.** The list payload already contains everything for the tree: nested children are top-level `steps` entries flagged `isWorkflow` in `allSteps`, and step ids resolve to registry entries by key or by workflow id. Run counts use the existing `GET /api/workflows/:workflowId/runs?status=…&perPage=1` and read the returned `total`.
- **Design system**: one additive, backward-compatible prop (`fit?: 'content' | 'container'` on `DataListRoot`, default unchanged), everything else composes existing primitives — the tree connectors are 1px borders using the muted border token.
- **Tests**: Vitest + MSW with typed `@mastra/client-js` fixtures covering the list behaviors (tree expand/collapse, cycle guard, running/suspended counts, search), plus unit tests for the tree derivation.
- **Known trade-off**: counts poll per (workflow × status), i.e. 2 lightweight requests per workflow per interval. Fine for typical projects; the cleaner long-term shape is probably a server-side aggregate (active-run counts on the list endpoint or a global runs query). Happy to align with whichever direction you prefer.

## Future direction (separate follow-up)

Workflow `tags` (mirroring datasets' `tags: string[]`) with a tag filter on this page — I can open a separate request if there is interest.

leandroppf:
Fixes #18924

## Description

Enhances the Studio `/workflows` list with composition visibility and live run status, as proposed in the linked issue:

- **Nested-workflow tree** — workflows composing other workflows show a muted count indicator and a chevron; expanding reveals child rows connected by tree guide lines. Registered children link to their own pages and expand recursively (cycle-safe — mutually nested workflows can't loop); nested workflows that aren't registered standalone render as inline, non-link rows. The same child under two parents expands independently. Nested step ids are resolved to registry entries by key **or** by workflow id, so camelCase registry keys with kebab-case workflow ids work.
- **Running column** — live per-workflow count of `status=running` runs (pulsing dot + count, blank when idle), refreshed every 5s.
- **Pending input column** — count of `status=suspended` runs (the human-in-the-loop state, awaiting `resume()`), with a distinct pause indicator, separated from Running per the issue discussion.
- **Viewport fit** — new additive `fit` prop on `DataList` (`'content'` default preserves current behavior for all existing consumers; `'container'` fills the container so flexible tracks truncate). The workflows list adopts `fit="container"` with a `minmax(0, 1fr)` description track, removing the horizontal scrolling that previously pushed trailing columns off-screen.

Counts now come from a new aggregated endpoint, following the review feedback about the original client-side N+1 polling (https://github.com/mastra-ai/mastra/pull/18925#issuecomment-4892672871): `GET /api/workflows/run-counts` returns `{ running, suspended }` per workflow registry key in a single request, and the Studio polls that one endpoint every 5s regardless of how many workflows are registered. Counts are scoped to the request-context resource id the runs endpoints already enforce and filtered per user when FGA is configured; otherwise a short in-process cache keeps repeated polls cheap. Older servers that predate the endpoint respond 404 once — the count columns render blank and polling stops; transient errors keep polling so a network blip never silences the counts.

Image of my real instance with the proposed changes
CleanShot 2026-07-03 at 16 55 06@2x

## Changes by package

- `@mastra/server`: new `GET /workflows/run-counts` route (schema, registration, handler) with resource-id scoping, FGA filtering, and a short-TTL in-process cache for the unscoped case, plus handler tests.
- `@mastra/client-js`: new `listWorkflowRunCounts()` method and response type; regenerated route types.
- `mastra` (CLI): regenerated API route metadata for the new route.
- `@mastra/playground-ui`: additive `DataListRoot` `fit` prop + `DataListSkeleton` passthrough (backward compatible, default unchanged), with tests.
- `@internal/playground`: workflows list tree UI (`nested-workflows.ts` derivation utils + connector rendering), `useWorkflowsRunCounts` now issues a single aggregated query (React Query, 5s polling, 404 fallback for older servers), new columns. Also a small dev-only fix proxying `/refresh-events` to the target server in the dev playground.
- `docs`: reference entries for the new server route and the new client method.
- Changesets included for `@mastra/server` (minor), `@mastra/client-js` (minor), `@mastra/playground-ui` (minor), and `@internal/playground` (minor).

## Testing

- `packages/server`: `vitest run src/server/handlers/workflows.test.ts` → 61 tests passing, including the new run-counts coverage: counts grouped by registry key, registry-key vs workflow-id mapping, FGA filtering (and that FGA-filtered requests never serve or fill the shared cache), a genuinely suspended run produced by a real execution, and static-route registration order ahead of the parameterized workflow route.
- `packages/playground`: `vitest run src/domains/workflows/` → 21 files, 139 tests passing; includes new `useWorkflowsRunCounts` tests (polling stops after a 404 from an older server, keeps polling through server and non-HTTP transient errors, standard interval otherwise) and list tests exercising the aggregated counts endpoint via MSW.
- `tsc --noEmit` clean on `@mastra/server` and `@mastra/client-js`; lint-staged (ESLint/Prettier) clean on all changed files at commit time.
- Tests follow the repo's MSW + typed `@mastra/client-js` fixtures approach (network mocked, real client + React Query driven; no app hooks mocked).
- Validated manually against a real project (20 workflows, camelCase registry keys, multi-level nesting incl. inline sub-workflows, parallel and suspended runs), including sustained polling of the new endpoint against a live, actively-writing store.

## Checklist

- [x] Linked issue: #18924
- [x] Documentation — added reference docs for `GET /workflows/run-counts` and `listWorkflowRunCounts()`
- [x] Added tests that prove the feature works
- [x] Will address all CodeRabbit review comments

**Type of change:** New feature (non-breaking change which adds functionality)

## ELI5
The Studio’s **workflows** page can now show a workflow **inside other workflows** as a neat tree you can expand/collapse. It also shows, for each workflow, whether it currently has runs **actively running** or **paused waiting for resume()**, updating automatically every few seconds.

### What changed
- **Nested workflow tree UI:** Workflows are rendered as an expandable/collapsible tree with chevrons and visual guide lines, including **cycle-safe** expansion and proper resolution between **registry keys** and **workflow IDs**.  
  - **Registered nested workflows** link to their own pages and can expand recursively (independently under different parents).
  - **Unregistered nested workflows** render inline as non-link leaf rows.
- **Activity columns (live run counts):**
  - **Running:** count of runs with `status=running` per workflow.
  - **Pending input:** count of runs with `status=suspended` per workflow (awaiting `resume()`).
  - Counts **auto-refresh every 5 seconds** and stop polling after a **404** indicates the endpoint isn’t supported on older servers.
- **Layout/overflow fix:** Added a `fit` prop to the Studio `DataList` and used `fit="container"` for the workflows list so it fits the viewport without horizontal overflow (with long descriptions truncating instead).
- **Backend + client support for counts:** Added an aggregated endpoint **`GET /api/workflows/run-counts`** and a client method **`listWorkflowRunCounts()`** to fetch `{ running, suspended }` counts in **one request per poll**, with request-context scoping, FGA filtering, caching for unscoped requests, and **404 fallback** behavior.
- **Tests, generated types, docs, and changesets:** Added/updated tests for nested rendering/expansion, cycle prevention, search filtering, and run-count behavior (including the 404/endpoint-unsupported case). Regenerated route/client/server types, updated reference docs, and added changesets for the affected packages.


mfrachet:
Hey there 👋🏻 and thanks for suggesting this PR.

This is a great idea, but I'm not sure this is the actual way to make it.

Right now, it seems the frontend is doing a N+1 query at the workflows page level on the frontend. Meaning that for each workflow, we'll end up making one additional fetch request (one per workflow id if I get it correctly).

I'm wondering if a better approach wouldn't to prefill some metadata fields on the backend instead to avoid having that many net requests.

Again, thanks for the suggestion 🙏🏻

leandroppf:
@mfrachet  Thank you so much for the review! You're right, I'll explore the possibility of prefilling some metadata fields on the backend to get the runs 🙏🏻 . I'd really love to improve the experience on this screen, when dealing with many workflows, it makes a huge difference.

leandroppf:
@mfrachet thanks again for the nudge, you were right that the client-side N+1 wasn’t the way. While exploring the prefill idea I landed on a slightly different shape: a small dedicated endpoint (GET /api/workflows/run-counts) that returns { running, suspended } for every workflow in one request, rather than embedding the counts in the GET /workflows response.
With this approach we can have the counts refreshing every 5s without the need to load all the workflows for each refresh. 🙏🏻 
Looking forward to your feedback! Thank you so much!

PR #18930, opened 2026-07-04
fix(deployer): resolve metadata for exported packages

### LINKED ISSUE #18849: [BUG] External dependency version resolved incorrectly when the package's exports does not expose ./package.json (e.g. execa 9)
### Describe the Bug

When `mastra build` generates the `package.json` for `.mastra/output`, it resolves the version of each external (non-bundled) dependency by reading that package's `package.json` through Node module resolution (`getPackageRootPath` → `local-pkg`'s `getPackageInfo`, in `@mastra/deployer`).

If the version actually used is an ESM package whose `exports` map does **not** expose `./package.json` (common/recommended for modern ESM packages — `execa@9` is one), then resolving `<pkg>/package.json` throws `ERR_PACKAGE_PATH_NOT_EXPORTED`. `local-pkg` swallows this and falls through to whatever *other* copy of the package it can find whose `package.json` is openly readable — typically an **older, legacy version** hoisted elsewhere in the workspace (or nothing at all).

The result: `.mastra/output/package.json` pins the **wrong version**. In my case a bundled workspace package imports `execa@9` (and uses its template API, `execa(options)\`cmd\``), but the output was pinned to `execa@5.1.1` — even though `execa@9.6.1` is installed right in the app's own `node_modules`. The deployed server then crashes at runtime because the v5 API is incompatible. The version Mastra records is decided by *which copy of the package happens to have an old-style, unrestricted `package.json`*, not by what the importing code actually depends on. If **no** readable copy exists, the dep falls back to `"latest"`.

**Root cause:** `getPackageRootPath(packageName, parentPath)` calls `getPackageInfo(packageName, { paths: [parentPath] })`, which resolves via the `<packageName>/package.json` subpath. Node applies the target package's `exports` map to that subpath; if `./package.json` isn't exported, resolution throws and the correct copy is skipped. The resolver should read `package.json` directly from the resolved package **directory** via `fs` (the file always exists on disk regardless of `exports`), rather than resolving the `./package.json` subpath through the module system.

### Steps To Reproduce

Minimal reproduction (see the linked repo, `README.md`):

```bash
pnpm install
pnpm -C apps/mastra-app build
cat apps/mastra-app/.mastra/output/package.json   # -> "execa": "5.1.1"  (WRONG; installed is 9.6.1)
```

The repo also includes `repro.mjs`, a ~10-line standalone demonstration of the resolver bug (no build needed):

```
$ node repro.mjs
app's node_modules/execa -> 9.6.1  (.pnpm/execa@9.6.1/node_modules/execa)
getPackageInfo("execa") from app -> 5.1.1  (execa@5.1.1/node_modules/execa)
require.resolve("execa/package.json") -> ERR_PACKAGE_PATH_NOT_EXPORTED
```

`execa@9.6.1`'s `exports` is `{ "types": "./index.d.ts", "default": "./index.js" }` — no `./package.json`. `execa@5.1.1` has no `exports` field, so its `package.json` is freely readable, and that stale copy is what gets picked.

https://github.com/segevfiner/mastra-execa-exports-repro

### Expected Behavior

`.mastra/output/package.json` should pin `execa` to the version the bundled code actually resolves (`9.6.1`), which the deployed server can run. Instead `dependencies.execa` is `"5.1.1"` and `node .mastra/output/index.mjs` fails at runtime. The deployer's external-version resolution should read the resolved package directory's `package.json` from disk rather than resolving the `./package.json` subpath through Node's `exports` gate, so packages that don't export `./package.json` are versioned correctly.

### Environment Information

```
System:
  OS: macOS 26.5.1 (arm64)
Binaries:
  Node: 22.22.2
  pnpm: 10.33.0
npmPackages:
  mastra: 1.12.2
  @mastra/core: 1.47.0
  @mastra/deployer: 1.41.0   (transitive; getPackageRootPath / bundleExternals)
  local-pkg: 1.1.0           (transitive; getPackageInfo)
  execa (needed by bundled code): 9.6.1
  execa (wrongly resolved/pinned): 5.1.1
```

- LLM Provider: N/A (build-time / bundler bug, provider-independent)

### Verification

- [X] I have searched the existing issues to make sure this is not a duplicate
- [X] I have included sufficient information for the team to reproduce and understand the issue

tianma-if:
Fixes #18849.

This updates deployer package metadata resolution to first resolve the package entry from the importing project, then read package.json directly from the resolved package root. That keeps .mastra/output/package.json pinned to the installed package version even when the package exports map does not expose ./package.json.

I also added a regression test for a package that exports its entry point but hides package.json.

Checks run:
- pnpm --filter ./packages/deployer test src/build/package-info.test.ts src/build/plugins/node-modules-extension-resolver.test.ts
- pnpm --filter ./packages/deployer lint
- pnpm --filter ./packages/deployer build:lib

## ELI5
This change makes the deployer check the installed package’s real folder first, instead of asking for `package.json` through the package export map. That way, it records the correct version even when a package hides its `package.json` file.

## Summary
- Fixed deployer metadata resolution so external package versions are read from the resolved package root on disk.
- Avoided relying on `<pkg>/package.json` export resolution, which can fail for packages that don’t expose that subpath.
- Added a regression test for a package that exports its entry point but hides `package.json`, confirming the correct root path and version are captured.
- Updated the changeset to note the patch release and the dependency version fix.


TheIsrael1:
## 🟠 high: Maintainer notes

**Merge confidence:** 5/5 — ready to merge

**Why:**
- The PR directly addresses linked issue #18849 by resolving the installed package from the importer context, walking back to the actual package root, and reading that package's `package.json` from disk, which covers packages whose `exports` map hides `./package.json`.
- Regression coverage was added for hidden `package.json` export metadata and package subpath fallback behavior, and I additionally exercised root, exported subpath, hidden subpath, and alias/symlink packageSpec cases locally.
- Local verification passed: `pnpm --filter ./packages/deployer test src/build/package-info.test.ts src/build/plugins/node-modules-extension-resolver.test.ts`, `pnpm --filter ./packages/deployer lint`, and `pnpm --filter ./packages/deployer build:lib`.

**Why not higher:**
- Confidence is already maxed. The only failing observed remote checks are Vercel preview deploy contexts requiring authorization, which are non-code/non-blocking for this deployer change.

**Required before merge:**
- None.

**Non-blocking follow-ups:**
- None.

**Author comments:**
- None.

**Prior maintainer notes:**
- None found.

TheIsrael1 (approved):

PR #18931, opened 2026-07-04
fix(core): backfill resource id on hydrated messages

### LINKED ISSUE #18559: hydrateMastraDBMessageFields skips resourceId backfill when a message already has a threadId
## Summary

In `packages/core/src/agent/message-list/conversion/input-converter.ts`, `hydrateMastraDBMessageFields` nests the `resourceId` backfill inside the `threadId` guard:

```ts
if (!message.threadId && context.memoryInfo?.threadId) {
  message.threadId = context.memoryInfo.threadId;

  if (!message.resourceId && context.memoryInfo?.resourceId) {
    message.resourceId = context.memoryInfo.resourceId;
  }
}
```

So a `MastraDBMessage` that already carries a `threadId` but is missing `resourceId` never gets `resourceId` hydrated from `memoryInfo`, because the outer `!message.threadId` condition is false.

This contradicts the function's own docstring ("Hydrate a MastraDBMessage with missing fields (id, createdAt, threadId, resourceId)") and the sibling adapter branches in the same file (the AIV5/AIV6 paths), which set `threadId` and `resourceId` independently from `context.memoryInfo`.

## Reproduction

Call the exported `hydrateMastraDBMessageFields` with a message that already has a `threadId` but no `resourceId`:

```ts
const message = {
  id: 'msg-1',
  role: 'user',
  createdAt: new Date(),
  threadId: 'thread-1',      // already set
  // resourceId missing
  content: { format: 2, parts: [] },
};

const context = {
  memoryInfo: { threadId: 'thread-1', resourceId: 'resource-1' },
  newMessageId: () => 'generated-id',
  generateCreatedAt: () => new Date(),
  dbMessages: [],
};

const result = hydrateMastraDBMessageFields(message, context, 'memory');
// Expected: result.resourceId === 'resource-1'
// Actual:   result.resourceId === undefined
```

This path is reachable from the public conversion flow: `inputToMastraDBMessage` validates that an incoming message's `threadId` matches `memoryInfo.threadId`, so DB-restored messages commonly arrive with `threadId` already set while `resourceId` may be absent.

## Impact

Resource-scoped memory and history association is silently dropped for re-hydrated messages that have a `threadId` but no stored `resourceId`.

## Suggested fix

Backfill `threadId` and `resourceId` in two independent guards, matching the sibling branches and the docstring.

tianma-if:
Fixes #18559.

This updates MastraDB message hydration so `resourceId` is filled from `memoryInfo` independently from `threadId`. Previously, messages that already had a `threadId` skipped the `resourceId` backfill.

I added a focused regression test for a restored message that already has `threadId` but is missing `resourceId`.

Checks run:
- pnpm --filter ./packages/core test src/agent/message-list/conversion/input-converter.test.ts
- pnpm --filter ./packages/core test src/agent/message-list/conversion/input-converter.test.ts src/agent/message-list/tests/message-list.test.ts
- git diff --check

## ELI5
When MastraDB restores a saved message, it should be able to remember which “resource box” it belongs to. This PR fixes hydration so that even if the message already has its “thread/conversation ID,” it will still fill in the missing “resource ID” from `memoryInfo`.

## What changed
- Fixed `hydrateMastraDBMessageFields` so `resourceId` is backfilled independently of whether `threadId` is missing.
- Kept any existing `threadId` intact while hydrating missing `resourceId` from `context.memoryInfo?.resourceId`.
- Added a regression test for a restored message that already has `threadId` but lacks `resourceId`, asserting that `resourceId` is restored and that unnecessary ID/createdAt generation is not triggered.
- Added a changeset entry for a patch release to `@mastra/core`.

## Notes
- This targets restored DB message conversion so resource-scoped memory/history association is preserved.

### DIFF
--- a/.changeset/tired-dancers-begin.md
+++ b/.changeset/tired-dancers-begin.md
@@ -1,0 +1,5 @@
+---
+'@mastra/core': patch
+---
+
+Fixed message hydration to backfill resource IDs when thread IDs are already present.
--- a/packages/core/src/agent/message-list/conversion/input-converter.test.ts
+++ b/packages/core/src/agent/message-list/conversion/input-converter.test.ts
@@ -1,0 +1,35 @@
+import { describe, expect, it, vi } from 'vitest';
+import type { InputConversionContext } from './input-converter';
+import { hydrateMastraDBMessageFields } from './input-converter';
+import type { MastraDBMessage } from '../state/types';
+
+describe('hydrateMastraDBMessageFields', () => {
+  it('backfills resourceId when the message already has a threadId', () => {
+    const message = {
+      id: 'msg-1',
+      role: 'user',
+      createdAt: new Date('2026-01-01T00:00:00.000Z'),
+      threadId: 'thread-1',
+      content: {
+        format: 2,
+        parts: [],
+      },
+    } satisfies MastraDBMessage;
+    const context = {
+      memoryInfo: {
+        threadId: 'thread-1',
+        resourceId: 'resource-1',
+      },
+      newMessageId: vi.fn(() => 'generated-id'),
+      generateCreatedAt: vi.fn(() => new Date('2026-01-02T00:00:00.000Z')),
+      dbMessages: [],
+    } satisfies InputConversionContext;
+
+    const result = hydrateMastraDBMessageFields(message, context, 'memory');
+
+    expect(result.threadId).toBe('thread-1');
+    expect(result.resourceId).toBe('resource-1');
+    expect(context.newMessageId).not.toHaveBeenCalled();
+    expect(context.generateCreatedAt).not.toHaveBeenCalled();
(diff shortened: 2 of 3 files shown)


tianma-if:
Addressed the CodeRabbit import-order nit in `b4f442d996`.

Validation run locally:
- `pnpm --filter ./packages/core test src/agent/message-list/conversion/input-converter.test.ts`
- `pnpm --filter ./packages/core exec eslint src/agent/message-list/conversion/input-converter.test.ts`
- `git diff --check`

I also ran `pnpm --filter ./packages/core lint`; it currently fails on an unrelated existing unused type in `packages/core/src/storage/utils.test.ts`, so I validated the touched file directly with ESLint.

abhiaiyer91 (approved):

PR #18955, opened 2026-07-05
fix(duckdb): forward listTracesLight through the lazy observability facade

### LINKED ISSUE #18942: [BUG] listTracesLight missing from DuckDBStore's lazy-load delegate wrapper — throws base-class "not implemented" despite the real implementation supporting it
From the Zynda team. We're a company building a production AI chat product on Mastra. This isn't production/severity-blocking for us — we have a working `--verbose` workaround — but it does block our default `mastra api trace list` path for local-dev observability (debugging agent/workflow behavior, cost, and latency), and we spent a good amount of time isolating the root cause, so we wanted to hand that over in case it saves you the same digging.

**Environment**
- mastra CLI: 1.16.0 → 1.18.0 (reproduces on both)
- @mastra/core: 1.49.0
- @mastra/duckdb: 1.5.0

**Symptom**

`mastra api trace list` (default, lightweight endpoint) returns:
```
500 "This storage provider does not support listing lightweight traces"
```
`trace list --verbose` (the full `listTraces` endpoint) works fine.

**Root cause (confirmed by reading dist source + an isolated reproduction)**

`@mastra/duckdb`'s `index.js` exports `DuckDBStore`, whose observability domain is a lazy-loading delegate shim class (also named `ObservabilityStorageDuckDB`, `dist/index.js` ~line 560) that extends `ObservabilityStorage` directly and dynamically imports the real implementation on first use:

```js
this.loadPromise = import('./observability-M35AJUGT.js')
  .then(({ ObservabilityStorageDuckDB }) => new ObservabilityStorageDuckDB({ db }));
```

Every method on this shim forwards to the delegate with an identical 3-line wrapper, e.g.:

```js
async listTraces(...args) {
  const delegate = await this.requireDelegate();
  return delegate.listTraces(...args);
}
```

This pattern exists for `getTrace`, `getTraceLight`, `listTraces`, `listBranches`, `listLogs`, every `getMetric*`, etc. — **but not for `listTracesLight`**. The source goes straight from the `listTraces` wrapper to the `listBranches` wrapper; `listTracesLight` has no delegate method on the shim at all. So a call falls through to the abstract `ObservabilityStorage` base class's default throw, even though `observability-M35AJUGT.js`'s actual `ObservabilityStorageDuckDB` class (the delegate, once loaded) does implement `listTracesLight` correctly.

**How this was isolated (ruling out our own app as the cause)**

Wrote a standalone script constructing, in isolation:
- (A) `new DuckDBStore(...).getStore('observability')`, called directly
- (B) the same object wrapped in a `MastraCompositeStore` (Postgres default + DuckDB observability override — our actual usage pattern)
- (C) the same composite wrapped in a full `new Mastra({ storage })`

All three throw identically, and object-reference checks (`===`) confirm the exact same underlying instance flows through every layer unchanged — ruling out `MastraCompositeStore`/`Mastra` wrapping as a contributing factor. Also ruled out: stale `.mastra` build cache (full delete + rebuild, still fails), and duplicate/mismatched package versions (exactly one copy of each package in node_modules).

**Suggested fix**

Add the missing delegate wrapper to the shim class, matching the existing pattern:

```js
async listTracesLight(...args) {
  const delegate = await this.requireDelegate();
  return delegate.listTracesLight(...args);
}
```

**Repro**

1. `mastra dev` with a DuckDBStore-backed observability domain
2. `mastra api trace list --url http://localhost:4111` → 500 error above
3. `mastra api trace list --url http://localhost:4111 --verbose` → succeeds

Related: #16304 (same "storage provider does not support X" pattern, different method/backend, still open). Originating PR: #16608 (added `listTracesLight` + `--verbose`; this delegate wrapper appears to have been missed).

We'd appreciate confirmation this matches your read of it, or a heads-up if there's a supported workaround beyond `--verbose` we should be using instead.

minh2416294:
## What
Adds the missing `listTracesLight` forwarding method to `@mastra/duckdb`'s lazy-loading observability store facade.

## Why
`mastra api trace list` (the default lightweight trace-listing path) returns a 500, `"This storage provider does not support listing lightweight traces"`, for any project using a DuckDB-backed observability store, even though DuckDB fully supports the operation. `--verbose` (the full `listTraces` path) works, which makes it look inconsistent.

Reported in #18942 with a confirmed repro.

## How
The DuckDB observability domain (`stores/duckdb/src/storage/index.ts`) is a lazy-loading facade that hand-forwards each method to a dynamically-imported delegate via `requireDelegate()`. Every observability method has a wrapper, except `listTracesLight`, which was missed (the originating PR #16608 added `listTracesLight` + `--verbose` to the real implementation but not this facade wrapper). So calls fell through to the base class `ObservabilityStorage.listTracesLight`, which throws by default.

The real implementation (`stores/duckdb/src/storage/domains/observability/index.ts`) already implements `listTracesLight` correctly. I this just adds the one missing wrapper, matching the surrounding `listTraces` / `listBranches` pattern verbatim.

## Testing
Added a regression test in `stores/duckdb/src/storage/domains/observability/index.test.ts` that seeds a root span and calls `listTracesLight({})` on the facade (`store.observability`).

- On `main` (without the fix): the test **fails** with `"This storage provider does not support listing lightweight traces"`.
- With the fix: it **passes**, returning the seeded trace.

Ran the full DuckDB observability suite locally: **151 passed**. `tsc --noEmit` clean.

Fixes #18942

## ELI5
DuckDB was missing a tiny forwarding step, so one trace-listing request was getting lost and causing an error. This change connects that request to the real DuckDB trace code, so lightweight trace lists work again.

### Summary
- Added `listTracesLight` forwarding to the DuckDB lazy-loading observability facade.
- Added a regression test to verify `store.listTracesLight({})` returns seeded lightweight traces.
- Added a changeset documenting the DuckDB patch fix.

### DIFF
--- a/.changeset/duckdb-list-traces-light.md
+++ b/.changeset/duckdb-list-traces-light.md
@@ -1,0 +1,5 @@
+---
+'@mastra/duckdb': patch
+---
+
+Fixed listing lightweight traces on DuckDB storage. Calling `listTracesLight` on a DuckDB observability store previously threw "This storage provider does not support listing lightweight traces" because the lazy-loading store facade was missing the forwarding method, even though DuckDB fully supports the operation. The call now returns lightweight traces as expected. Fixes #18942.
--- a/stores/duckdb/src/storage/domains/observability/index.test.ts
+++ b/stores/duckdb/src/storage/domains/observability/index.test.ts
@@ -430,6 +430,48 @@ describe('ObservabilityStorageDuckDB', () => {
       expect(traces.spans.length).toBeGreaterThanOrEqual(1);
     });
 
+    it('listTracesLight returns seeded traces via the facade', async () => {
+      await storage.batchCreateSpans({
+        records: [
+          {
+            traceId: 'trace-light-1',
+            spanId: 'root-light',
+            parentSpanId: null,
+            name: 'workflow-run',
+            spanType: SpanType.WORKFLOW_RUN,
+            isEvent: false,
+            entityType: EntityType.WORKFLOW_RUN,
+            entityId: 'wf-light',
+            entityName: 'myWorkflow',
+            userId: null,
+            organizationId: null,
+            resourceId: null,
+            runId: null,
+            sessionId: null,
+            threadId: null,
+            requestId: null,
+            environment: 'production',
+            source: null,
+            serviceName: 'svc',
+            scope: null,
+            attributes: null,
+            metadata: null,
+            tags: ['v1'],
+            links: null,
+            input: null,
+            output: null,
+            error: null,
+            startedAt: new Date('2026-01-03T00:00:00Z'),
+            endedAt: new Date('2026-01-03T00:00:01Z'),
+          },
+        ],
+      });
+
+      // On base the facade falls through to the base-class throw; after the fix it resolves.
+      const result = await storage.listTracesLight({});
+      expect(result.spans.map(s => s.traceId)).toContain('trace-light-1');
+    });
+
     it('listTraces applies scalar prefilter and tag post-filter correctly', async () => {
       await storage.batchCreateSpans({
         records: [
--- a/stores/duckdb/src/storage/index.ts
+++ b/stores/duckdb/src/storage/index.ts
@@ -202,6 +202,13 @@ export class ObservabilityStorageDuckDB extends CoreObservabilityStorage {
     return delegate.listTraces(...args);
   }
 
+  async listTracesLight(
+    ...args: Parameters<ObservabilityStoreImpl['listTracesLight']>
+  ): ReturnType<ObservabilityStoreImpl['listTracesLight']> {
+    const delegate = await this.requireDelegate();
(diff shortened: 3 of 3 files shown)


minh2416294:
Hey, This PR is code-complete (fix + regression test + changeset) but I've had to leave it as a draft for now. My fork account has hit GitHub's open-PR limit for this repo, so the "Ready for review" button is disabled until one of my other open PRs (#18917, #18604) merges and frees a slot. I'll mark it ready as soon as that clears. Thank you guys so much

The change is a one-method fix: the DuckDB observability store is a lazy-loading facade that forwards each call to its delegate, but the `listTracesLight` wrapper was missing, so calls fell through to the base class's "not supported" throw even though DuckDB implements the method. The regression test fails on `main` with that exact error and passes with the fix

abhiaiyer91 (approved):

minh2416294:
@abhiaiyer91 Thank you so much brother for merging my PR

PR #18980, opened 2026-07-06
fix(react): keep interleaved text and tool calls in order while streaming

### LINKED ISSUE #18964: [BUG] Studio chat: live streaming view merges text segments across tool calls (order correct after refresh)
### Describe the Bug

In Studio's agent chat, when a single assistant turn interleaves text and tool calls (text → tool call → more text, as produced by a multi-step / supervisor agent loop), the live streaming view renders incorrectly:

- consecutive text segments from before and after a tool call are merged into one continuous text block, and
- the tool-call chip is displaced below text it should precede.

After refreshing the page, the same conversation re-renders from storage with the correct interleaved order (text → tool chip → text). So the persisted message parts are correct — only the live streaming renderer loses the interleaving while chunks are coming in.

Environment:
- mastra 1.18.0, @mastra/core 1.49.0 (Studio via `mastra dev`)
- Agent: supervisor/routing agent with sub-agents, tools and workspace skills; memory enabled (LibSQL storage, working memory)
- Model: deepseek/deepseek-chat
- macOS, Chrome

Expected behavior: the live streaming view preserves the same text / tool-call interleaving that the post-refresh (persisted) render shows.

### Steps To Reproduce

1. Define an agent whose turns interleave text and tool calls — e.g. a supervisor/routing agent with sub-agents and tools, and instructions that make it narrate briefly before each tool call. Enable memory.
2. Run `mastra dev`, open Studio, and chat with the agent so that one assistant turn produces: text A → tool call → text B.
3. Watch the live streaming view: text A and text B render merged into a single text block, and the tool-call chip is displaced after them (screenshot 1).
4. Refresh the page: the same turn now renders correctly as text A → tool chip → text B (screenshot 2).

https://github.com/liurongjie0/credit-risk-agent

### Expected Behavior

The live streaming view should preserve the actual part order of the assistant turn — text A, then the tool-call chip, then text B — exactly as the post-refresh render (re-hydrated from storage) already shows. Live view and persisted view should be consistent.

### Environment Information

```block
System:
  OS: macOS 26.4.1
  CPU: (10) arm64 Apple M5
  Shell: 5.9 - /bin/zsh
Binaries:
  Node: 25.9.0 - /opt/homebrew/bin/node
  npm: 11.12.1 - /opt/homebrew/bin/npm
Browsers:
  Chrome: 149.0.7827.156
  Safari: 26.4
npmPackages:
  @mastra/core: latest => 1.49.0
  @mastra/duckdb: ^1.5.0 => 1.5.0
  @mastra/editor: ^0.13.4 => 0.13.4
  @mastra/fastembed: ^1.2.0 => 1.2.0
  @mastra/libsql: ^1.15.0 => 1.15.0
  @mastra/mcp: ^1.13.0 => 1.13.0
  @mastra/memory: ^1.22.1 => 1.22.1
  @mastra/observability: ^1.15.2 => 1.15.2
  @mastra/rag: ^2.4.0 => 2.4.0
  mastra: latest => 1.18.0
  tsx: latest => 4.23.0
  typescript: latest => 6.0.3
  zod: latest => 4.4.3

- LLM Provider: DeepSeek (deepseek/deepseek-chat via Mastra model router)
- Browser: Chrome 149 (also reproduced in the same browser after refresh renders correctly)
```

### Verification

Sandy-1711:
## Description

In the Mastra Studio chat, when an assistant turn interleaves text, a tool call, and more text (`text` -> `tool call` -> `more text`), the live streaming view incorrectly merged both text blocks into one and pushed the tool-call chip below them. Reloading the page pulled the correct order from storage, meaning the bug was entirely isolated to the live stream accumulator.

image

### The Root Cause
The live view relies on the stream accumulator in `@mastra/react` to stitch incoming stream chunks into an ordered list of message parts (e.g., `[text, tool-invocation, text]`). 

The bug is triggered by providers like DeepSeek that reuse the exact same text ID (`txt-0`) for text segments that come *after* a tool call, re-emitting `text-start` with that duplicated ID.

Two factors combined to cause the text merge:
- When the post-tool text chunk arrived with `text-start(txt-0)`, the code scanned the entire message for any part matching `txt-0`. Because it found the pre-tool text block, it assumed it was a duplicate and skipped creating a new message part entirely.
- As the characters for the second text block streamed in, `text-delta` searched the entire list for `txt-0`, matched the pre-tool text block, and appended the new characters directly to it.

This forced the pre-tool and post-tool text to fuse into a single block, dropping the tool chip below them (`[text(A+B), tool]`).

### The Fix
Once a tool call appears in the stream, the text sequence preceding it is treated as closed. Any text arriving afterward belongs to a new segment. The fix enforces this boundary in two places:

- **In `text-delta`:** Before appending to a matched text part, we check if a `tool-invocation` now sits after it in the list. If it does, we don't append; we create a fresh text part instead.
- **In `text-start`:** We only deduplicate if the very last part in the message is an open text part with that ID. If a tool call has since been appended to the tail, a matching ID means a brand-new segment is starting, so we create a new part.

Because this logic relies on positioning (`Is there a tool after this text?`) rather than a provider's specific ID scheme, it cleanly handles all cases: unique IDs (which already worked), reused IDs (DeepSeek), and missing IDs.

## Demo
Before and after (live streaming in Studio, DeepSeek model):
**Before (Merged text blocks, tool out of order):**

https://github.com/user-attachments/assets/217ff869-86c7-4cb2-b43e-6e200760c43e

**After (Correct interleaving sequence):**

https://github.com/user-attachments/assets/7856f388-2786-4ef7-bd11-89b6d99ad1e7

## How I verified it

- **Live Reproduction:** Captured and verified the thread-subscription stream running through Studio via the direct DeepSeek provider. Confirmed that both distinct text segments arrived labeled as `txt-0`. Verified that the post-fix UI accurately streams chunks sequentially in real-time, matching the state on a page reload.
- **Automated Regression Tests:** Added comprehensive test suites handling interleaved `text -> tool -> text` scenarios across unique, reused, and completely missing text IDs.
## Related issue(s)

Fixes #18964

## Type of change

- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [x] Test update

## Checklist

- [x] I have linked the related issue(s) to this pull request
- [ ] I have updated the documentation accordingly
- [x] I have addressed all CodeRabbit comments

## ELI5

When the assistant streams text, a tool call, and more text, the live chat now keeps these parts in the correct order. This prevents separate text sections from merging and moving the tool chip.

## What changed

- Updated the React streaming accumulator to close text segments when any non-text part follows.
- Updated `text-start` deduplication to check only the trailing open text part.
- Updated `text-delta` handling to create a new text segment after an intervening non-text part.
- Updated `getPartKey` to include the part index with `textId`, preventing key collisions when providers reuse IDs.
- Added regression tests for distinct, reused, and missing text IDs, including tool and reasoning boundaries.
- Added a patch changeset for `@mastra/react`.


minh2416294 (commented):
﻿Hey, I have two questions.
- `getPartKey` now ties the text key to array index. That's fine today because every part addition appends, but it quietly assumes append-only ordering, so it's worth a comment so nobody later inserts a part ahead of a streaming text block and triggers a remount.
- Did you consider `text-start(t1) -> reasoning -> text-start(t1)` (same id, reasoning rather than a tool between)? The old `.some()` would dedupe that; the tail check won't, so it'd split into two text parts. Probably the right call, but the three new tests only cover text->tool->text.

I can add a text->reasoning->text case and test this against DeepSeek locally if useful.

Sandy-1711:
Thanks @minh2416294, both fair points; I've handled them.

For the key, I added a comment on `getPartKey `noting that mixing the index in assumes parts are append-only, so nobody later inserts ahead of a streaming text block and triggers a remount.

On the reasoning case, you're right that splitting is the intended behavior. Rather than special-case it, I generalized the check so a text run closes on any non-text part that follows it, not just a tool call, so `text → reasoning → text` with a reused id splits correctly. Added tests for it too, both with and without a second `text-start`.

PR #18983, opened 2026-07-06
refactor(mastracode): context-driven web chat UI with skeleton loading states

mfrachet:
Restructures the mastracode web UI so state lives in dedicated context providers instead of being drilled through a giant AppLayout, and makes every loading state render a proper skeleton instead of placeholder text or a blank screen.

Chat.tsx is now a pure composition root (`ActiveProjectProvider → ChatSessionProvider → OverlaysProvider → ChatCommandsProvider`), AppLayout/AppContent are gone in favor of a slotted ChatLayout plus chat-domain components (ChatHeader, ChatMessageList, ComposerPanel, ChatOverlays), and the sidebar sections (ProjectSwitcher, WorkspacesSection, ThreadList) consume contexts directly instead of taking a dozen props each. `relativeTime` moved into a shared date lib backed by date-fns.

Loading states before: `<Txt>Loading providers…</Txt>` or `return null` while auth is pending. After: `<SkeletonRows label="Loading providers" rows={3} />` — a playground-ui Skeleton wrapper with `role="status"` + `aria-label`, so deep links shimmer instead of flashing a blank page and tests query by role/name instead of text.

Everything was done TDD with MSW: 41 spec files / 232 tests in the web-ui suite (delayed handlers cover the pending states), unit suite green, typecheck and lint clean, changesets included.

## ELI5
This update cleans up the chat app so different parts manage their own state instead of one big screen passing everything around. It also makes loading screens nicer by showing skeleton placeholders instead of blank spaces or text.

## Summary
- Refactored the web UI around dedicated context providers for active project, chat session, overlays, and chat commands.
- Split the chat screen into smaller domain components, with `Chat.tsx` now acting as a composition root and `ChatLayout` handling slots for header, content, footer, and sidebar.
- Moved shared relative-time formatting into a date utility backed by `date-fns`.
- Replaced many loading text/blank states across chat, settings, workspaces, and auth with accessible `SkeletonRows` placeholders.
- Added and updated MSW/Vitest coverage for the new provider-based flows, overlay behavior, chat rendering, sidebar actions, and loading states.
- Added the corresponding changeset and dependency update for `date-fns`.


abhiaiyer91 (approved):

mastracode/src/web/ui/domains/chat/components/ThreadList.tsx
```diff
@@ -0,0 +1,283 @@
...
+        <Txt as="span" variant="ui-sm" className={`truncate ${thread.title ? 'text-icon6' : 'text-icon3 italic'}`}>
+          {thread.title || 'Untitled'}
+        </Txt>
+        {thread.updatedAt && (
+          <Txt as="span" variant="ui-xs" className="shrink-0 text-icon3">
+            {relativeTime(thread.updatedAt)}
+          </Txt>
+        )}
+      </button>
+      <div className="relative pr-1" ref={menuRef}>
```
damien-schneider, line 238: we should use a dropdown menu here instead of custom handling

mastracode/src/web/ui/domains/chat/hooks/useGlobalShortcuts.ts
damien-schneider, line 1: We probably want a hook in playground-ui to share the logic for shortcuts within the different apps (can be done as follow up of course, writing it to not loose it)

mastracode/src/web/ui/domains/workspaces/components/ProjectSwitcher.tsx
```diff
@@ -0,0 +1,59 @@
...
+import { useOverlays } from '../../../lib/overlays';
+import { useActiveProjectContext } from '../context/ActiveProjectProvider';
+
+/**
+ * Propless project switcher: shows the active project and opens the projects
+ * modal. Opening the modal also closes the sidebar drawer (mobile behavior).
+ */
+export function ProjectSwitcher() {
+  const { activeProject } = useActiveProjectContext();
+  const overlays = useOverlays();
```
damien-schneider, line 14: Didn't check in depth but this smells the need of using a Dialog or a Drawer instead of this, and if it is about having the overlay content then the name can be more explicit I think

mastracode/src/web/ui/lib/overlays/overlays.tsx
```diff
@@ -0,0 +1,52 @@
...
+  sidebar: false,
+  palette: false,
+  settings: false,
+  shortcuts: false,
+  projects: false,
+};
+
+const OverlaysContext = createContext<OverlaysApi | null>(null);
+
+export function OverlaysProvider({ children }: { children: ReactNode }) {
```
damien-schneider, line 31: Ok it is not a drawer then but it is more about using the sidebar component of the playground UI instead, so it can be addressed in a follow‑up I think

PR #18988, opened 2026-07-06
fix(core): resume only selected foreach iteration

### LINKED ISSUE #18961: foreach: plain resume() without forEachIndex resumes all suspended iterations
### Description
When a `.foreach()` step runs with `concurrency > 1` and multiple iterations `suspend()` simultaneously, resuming with `run.resume({ resumeData })` **without** a `forEachIndex` resumes *every* suspended iteration — each with the same `resumeData` — instead of only the intended one. The workflow jumps to `success` when it should stay `suspended`.

### Steps to reproduce
- Workflow: `.foreach(approvalStep, { concurrency: 3 })`, where `approvalStep` calls `suspend()` when it has no `resumeData`.
- `run.start({ inputData: [{ name: 'alpha' }, { name: 'beta' }, { name: 'gamma' }] })` → status `suspended` (all three suspended).
- `run.resume({ resumeData: { approved: true } })` (no `forEachIndex`).

Observed: status `success`; all three iterations resumed with the same data.
Expected: only the first suspended index resumes; status stays `suspended` for the others (matching the `forEachIndex`-based resume semantics).

### Root cause
`packages/core/src/workflows/handlers/control-flow.ts:1157`:
```ts
const isIndexSuspended = prevItemResult?.status === 'suspended' || resumeIndex === k;
if (isIndexSuspended) { resumeToUse = resume; }
```
This `else` branch runs when `resume.forEachIndex === undefined` (plain resume). `resumeIndex` is the single intended iteration (from `__workflow_meta.foreachIndex`), but the extra `prevItemResult?.status === 'suspended' ||` term makes the guard true for every still-suspended iteration, so all of them receive the same `resume`.

### Suggested fix
Match by index only: `const isIndexSuspended = resumeIndex === k;` (drop the suspended-status term), mirroring the `forEachIndex` branch just above it.

thatssoheil:
Fixes #18961.

Plain `run.resume({ resumeData })` for a suspended `.foreach()` should only resume the iteration selected by the saved workflow metadata. The previous guard also matched every still-suspended iteration, so concurrent foreach runs could reuse the same resume payload for all pending items and finish too early.

This changes the plain-resume branch to match only `resumeIndex === k` and adds a regression test with three suspended foreach iterations.

Verification run locally: focused regression red before the fix and green after; `pnpm --filter @mastra/core exec vitest run src/workflows/default.test.ts`; `pnpm --filter @mastra/core exec vitest run src/workflows/default.test.ts src/workflows/state-reader.test.ts src/workflows/nested-resume-label.test.ts`; changed-file Prettier and ESLint; `pnpm --filter ./packages/core check`. I also tried the full `pnpm --filter @mastra/core test`, but unrelated agent/workflow test workers hit Node heap OOM and teardown stalled, so I stopped that broader run.

AI assistance disclosure: I used OpenAI Codex to help draft the patch and test, with local verification before opening this PR.

## ELI5
If a workflow is paused while running the same step multiple times (`foreach`), this fix makes sure “resume” restarts only the one intended paused iteration—rather than restarting every still-paused iteration. That prevents the workflow from completing early.

## What changed
- Fixed plain `run.resume({ resumeData })` handling for suspended `.foreach()` steps: the resume selection now matches by `resumeIndex === k` only (instead of also depending on whether the persisted per-index result was still `suspended`).
- Added a regression test that suspends three `foreach` iterations and resumes without `forEachIndex`, asserting only the selected iteration restarts (`resumedItems` is `[10]`), another suspended iteration remains paused (`suspendedItems` contains `20`), and the overall result stays `suspended`.
- Updated the changeset to release this as a patch for `@mastra/core`.

## Verification
- Notes local verification with targeted Vitest runs, plus lint/format checks and `pnpm --filter ./packages/core check`.


taofeeq-deru:
Hi @thatssoheil , will be closing this PR. I've dropped a comment on the issue - #18961 
The current behaviour is intended, and you can control the number of foreach indices to resume using the concurrency option. Or better still, use forEachIndex when resuming

PR #18998, opened 2026-07-06
feat(core): add onTitleGenerated callback to agent execution options

### LINKED ISSUE #18076: [FEATURE] Expose a callback/hook when a thread title is generated

intojhanurag:
## Description

Add a per-request `onTitleGenerated` callback to the `memory` property of `agent.generate()` and `agent.stream()`. When `generateTitle` is enabled in memory config, this callback fires after the thread title is generated and persisted to storage — letting users react to the title (e.g., push it via SSE) without wrapping storage adapters.

- Added `onTitleGenerated?: (title: string) => void | Promise<void>` to `AgentMemoryOption`
- Threaded the callback through `map-results-step.ts` into `#executeOnFinish`, where it is invoked after `memory.createThread()` persists the title
- Refactored the title generation promise chain from `.then(onFulfilled, onRejected)` to `.then().catch()` to prevent unhandled rejections when the callback or persistence throws
- Added 4 unit tests covering generate, stream, existing-title skip, and callback error handling
- Added docs for `agent.generate()` and `agent.stream()` reference pages

```ts
await agent.stream('Hello', {
  memory: {
    thread: threadId,
    resource: userId,
    onTitleGenerated: (title) => {
      res.write(`event: title\ndata: ${JSON.stringify({ title })}\n\n`);
    },
  },
});
```

## Related Issue(s)

Fixes #18076

## Type of Change

- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [ ] Test update

## Checklist


.changeset/many-dancers-boil.md
```diff
@@ -0,0 +1,16 @@
...
+---
+
+Added `onTitleGenerated` callback to agent execution options. When `generateTitle` is enabled, pass `onTitleGenerated` to `agent.generate()` or `agent.stream()` to be notified when the thread title has been generated and persisted to storage. This is a per-request callback (like `onFinish`), so each request handler can use its own callback with direct access to its response stream — no storage adapter wrapping needed.
+
+**Example:**
+
+```ts
+await agent.stream('Hello', {
+  memory: { thread: threadId, resource: userId },
+  onTitleGenerated: (title) => {
```
wardpeet: Not sure about putting it here, though I thought we were putting it in memory as a trigger, but I guess we want people to act on it in the stream 🤔. Let's move this to the memory property instead, where memory thread id and resourceId are present.

intojhanurag: yeah I followed the pattern of onFinish, onStepFinish, onChunk, onError, onAbort , all per-request callbacks sitting as top-level execution options. Let me move this to memory property instead :)

abhiaiyer91 (approved):

PR #18999, opened 2026-07-06
feat(core): add readOnlyWorkspace option to local sandbox

### LINKED ISSUE #18992: [FEATURE] Add `readOnly` to `LocalSandbox` with isolation enabled
### Problem Statement

If I want to use a readonly local file system that can safely be shared across many agents/workflows that shouldn't really modify it, I'd want a readonly sandbox as well, but such an option is not exposed today.

### Proposed Solution

That should really only be a flag away to bwrap/seatbelt to just mount the working/workspace directory as read only which can be added to the `nativeSandbox` property.

### Component

Workspaces

### Alternatives Considered

Fork the `LocalSandbox` code and do it myself. Or hack it via custom bwrap/seatbelt options (This obviously sucks).

### Example Use Case

Using a readonly filesystem and sandbox to be shared across multiple agents/workflows where you don't want them to modify the files.

### Additional Context

_No response_

### Verification

Rudra2637:
## Description

This PR introduces the `readOnlyWorkspace` flag to the native sandbox configuration. This allows developers to run processes in an isolated local sandbox that mounts the workspace/working directory as read-only.

**Key Changes:**
- **Types**: Added `readOnlyWorkspace?: boolean;` to [NativeSandboxConfig](file:///C:/Users/Rudra/Desktop/open%20source/mastra/packages/core/src/workspace/sandbox/native-sandbox/types.ts).
- **Linux (Bubblewrap)**: If `readOnlyWorkspace` is true, mounts the workspace directory using `--ro-bind` instead of `--bind`.
- **macOS (Seatbelt)**: If `readOnlyWorkspace` is true, omits the `(allow file-write* (subpath ...))` rule for the workspace path in the generated SBPL profile.
- **Bug Fix**: Fixed a platform compatibility bug in `bubblewrap.ts` where resolving the node binary directory on Windows (using hardcoded forward slashes `nodePath.lastIndexOf('/')`) would resolve to `""`, causing unit tests to fail on Windows. It now handles backslashes on Windows properly.
- **Tests**: Added direct unit tests validating correct generation of bubblewrap command arguments and seatbelt profiles when the option is enabled or disabled.

## Related issue(s)

Fixes https://github.com/mastra-ai/mastra/issues/18992

## Type of change

- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [x] Test update

## Checklist

## ELI5
This PR adds a “read-only sandbox” option so programs running in the local sandbox can open and read your workspace files, but can’t modify them. It enforces this on Linux (Bubblewrap) and macOS (Seatbelt), and also fixes a Windows path issue so the sandbox correctly locates the Node directory.

### Summary
- Added `readOnly?: boolean` to `NativeSandboxConfig` (defaults to `false`; ignored when `isolation` is `'none'`)
- Linux: updated Bubblewrap command generation to mount the workspace with `--ro-bind` when `readOnly` is enabled (otherwise uses `--bind`)
- macOS: updated Seatbelt profile generation to omit the workspace write rule when `readOnly` is enabled
- Fixed Windows path handling in `bubblewrap.ts` so the Node binary directory is resolved correctly with backslashes
- Added unit tests validating both enabled/disabled behavior for Bubblewrap and Seatbelt
- Updated the changeset to document the new option and bump the `@mastra/core` minor release


CalebBarnes (dismissed):
Thanks for this @Rudra2637 — the implementation is clean and I like that you wired **both** native backends (seatbelt profile + bwrap `--ro-bind`) instead of just one, with unit tests covering each direction. A few requests before merge, mostly naming/docs:

**1. Rename `readOnlyWorkspace` → `readOnly`**

The field lives on the sandbox config, but `Workspace` is a first-class Mastra primitive (`new Workspace({ filesystem, sandbox })`), so `new LocalSandbox({ readOnlyWorkspace: true })` reads like it's reaching up to configure the `Workspace` rather than the sandbox itself. `readOnly` reads naturally as a property of the sandbox and pairs symmetrically with the existing `LocalFilesystem({ readOnly: true })`.

**2. Reword the JSDoc — avoid "Mount" and "workspace directory"**

Both terms collide with existing Mastra concepts: our filesystem mounts API and the `Workspace` class — neither is what's happening here (this is an OS-level `--ro-bind` / seatbelt rule). Suggest access-oriented wording that matches the sibling `readOnlyPaths` / `readWritePaths` docs, and that also makes the `isolation: 'none'` no-op explicit:

```ts
/**
 * Restrict the sandbox's working directory to read-only access.
 * Spawned processes can read but not write files in the working directory.
 * Enforced at the OS level (seatbelt profile / bwrap --ro-bind); ignored when isolation is 'none'.
 * @default false
 */
readOnly?: boolean;
```

**Also:** please update the changeset text to match the renamed field, and add a short code example — the repo convention is for new-feature changesets to show public-API usage.

**Optional nit:** the `bwrap` backend is only selected on Linux (`isBwrapAvailable()` returns false elsewhere), so `bubblewrap.ts` effectively only runs there — which makes the `nodePath.includes('\\')` branch unreachable, since `process.execPath` will be a POSIX path. While you're in there, the whole node-dir computation could just be `path.dirname(nodePath)` instead of the manual `lastIndexOf`/`substring` split (matches CodeRabbit's suggestion, minus the `win32` case). Not a blocker.

Rudra2637:
Thanks for the review @CalebBarnes , I will look into it and make the changes accordingly

Rudra2637:
should i look after the nitpick comments of the coderabbit as well @CalebBarnes ?

Rudra2637:
Hey @CalebBarnes ! Just checking in on my PR. Whenever you have some time, I'd really appreciate a review. No worries if you're busy—just wanted to follow up. Thanks! 🙂

packages/core/src/workspace/sandbox/native-sandbox/types.ts
```diff
@@ -56,6 +56,14 @@ export interface NativeSandboxConfig {
...
    */
   bwrapArgs?: string[];
+
+  /**
+   * Restrict the sandbox's working directory to read-only access.
+   * Spawned processes can read but not write files in the working directory.
+   * Enforced at the OS level (seatbelt profile / bwrap --ro-bind); ignored when isolation is 'none'.
+   * @default false
+   */
+  readOnly?: boolean;
```
NikAiyer, line 71: Could we document the enforcement exceptions here? `readWritePaths` and mounted filesystem targets remain writable, while non-empty `bwrapArgs` and an existing `seatbeltProfilePath` replace the generated policy and can bypass `readOnly` entirely. Since this is a public security-related option, callers should be able to understand its actual boundary from the API contract.

packages/core/src/workspace/sandbox/local-sandbox.test.ts
```diff
@@ -622,6 +629,67 @@ describe('LocalSandbox', () => {
 
       expect(info.metadata?.isolation).toBe('none');
     });
+
+    describe('readOnly option', () => {
+      it('should generate a bwrap command with --ro-bind when readOnly is true', () => {
```
NikAiyer, line 644: These assertions verify policy serialization, but not the user-facing OS-level guarantee. Could we add platform-gated behavioral coverage that runs real sandboxed commands and independently verifies:

1. reading an existing workspace file succeeds;
2. creating or overwriting a workspace file fails and leaves the host filesystem unchanged; and
3. writing inside an explicit `readWritePaths` exception succeeds?

Please use separate commands for each behavior so shell chaining or short-circuiting cannot produce false positives. I consider this blocking because `readOnly` is presented as an OS-enforced security boundary.

packages/core/src/workspace/sandbox/local-sandbox.test.ts
```diff
@@ -622,6 +629,67 @@ describe('LocalSandbox', () => {
...
+        while (index !== -1) {
+          bindIndices.push(index);
+          index = args.indexOf('--bind', index + 1);
+        }
+        for (const idx of bindIndices) {
+          expect(args[idx + 1]).not.toBe(workspacePath);
+        }
+      });
+
+      it('should generate a bwrap command with --bind when readOnly is false or undefined', () => {
```
NikAiyer, line 670: The title says this covers both `readOnly: false` and `undefined`, but the test only passes `{}`, so it verifies the undefined/default case. Could we either explicitly test both configurations or rename the test to match its assertion? The same applies to the Seatbelt default-writable test.

NikAiyer (changes requested):
Thanks for adding this. The implementation direction fits the existing native-sandbox design. The behavioral-test and public-contract concerns are noted inline.

One related lifecycle concern: when an existing `seatbeltProfilePath` is loaded, that custom profile initially replaces the generated policy, so `readOnly` may not apply. A later mount or unmount regenerates the profile from `NativeSandboxConfig`, replacing the custom profile in memory and potentially changing enforcement for subsequent processes. This behavior predates this PR, but the new option makes it more visible. Could we either address it here or track it explicitly as a follow-up?

Rudra2637:
Thanks! for pointing this out! I agree this lifecycle behavior predates the readOnly changes, and the new option makes it more apparent. I'd prefer to track the seatbelt profile path lifecycle issue as a separate follow-up PR to keep this PR focused on the new readOnly feature. I will open a follow-up issue/PR for that right after this one.

That said, I'm happy to address it in this PR instead if you think that's the better approach. What would you prefer?

Rudra2637:
Also I'll go through the inline comments and make the requested changes shortly

NikAiyer:
That split makes sense since the lifecycle behavior predates this PR. Could you open and link a follow-up issue or PR before this PR merges? I’ll review the inline changes once they’re ready.

Rudra2637:
Thanks! That sounds good. I'll open a follow-up issue and pr. Will work through the remaining inline comments

packages/core/src/workspace/sandbox/native-sandbox/types.ts
Rudra2637: I've updated the JSDoc for the readOnly property to explicitly document these exceptions (explicit readWritePaths and overrides like custom bwrapArgs / seatbeltProfilePath) so developers understand the exact security boundaries.

packages/core/src/workspace/sandbox/local-sandbox.test.ts
Rudra2637: Added platform-gated behavioral tests for both macOS (seatbelt isolation) and Linux (bwrap isolation) that execute actual commands inside the sandbox to verify the OS-level security guarantees

packages/core/src/workspace/sandbox/local-sandbox.test.ts
Rudra2637: Thanks for pointing this out ,
I've updated both the Bubblewrap and Seatbelt default-writable unit tests to explicitly test and assert both {} (undefined) and { readOnly: false } configurations.

Rudra2637:
@NikAiyer, I do want to take this (lifecycle one) on as the follow-up. At the moment I already have two open PRs, so I can't open another one until one of them is merged. Would it be okay if I open and link the follow-up issue once this PR is merged, and then submit the follow-up PR myself?

NikAiyer (approved):
Hey @Rudra2637 that's fine. Just ping me when you submit the PR and I can take a look. I also added a small change myself for the readOnly change for temp directories. Otherwise, LGTM!

Rudra2637:
Thanks! I'll ping you once the PR is up. I will make that pr when one of my two  pr is merged.

Rudra2637:
Hey @NikAiyer i have an update, I've submitted the follow-up PR here: https://github.com/mastra-ai/mastra/pull/20197. I'd appreciate it if you could take a look when you have a chance.
Issue Link: https://github.com/mastra-ai/mastra/issues/20196

NikAiyer:
hey @Rudra2637 will take a look at your other PR now!

PR #19006, opened 2026-07-06
feat(mongodb): add filterFields to createIndex for index-level filter hints

### LINKED ISSUE #18587: feat(core): add filterFields to CreateIndexParams for adapter-level index filter hints
## Problem

`CreateIndexParams` in `@mastra/core` has no way for a caller to declare which metadata fields will be used for filtering:

```ts
export interface CreateIndexParams {
  indexName: string;
  dimension: number;
  metric?: 'cosine' | 'euclidean' | 'dotproduct';
}
```

For MongoDB Atlas Vector Search, declaring a field as a `filter` field in the index definition allows `$vectorSearch` to filter on that field natively during the ANN search — without a separate `$match` stage and without materialising candidate `_id`s into an `$in` array.

When **metadata** filter fields are **not** declared in the index, the MongoDB adapter must use a two-step pre-filter:

1. Run `$match` on the metadata fields to collect matching `_id`s.
2. Pass `{ _id: { $in: candidateIds } }` to `$vectorSearch.filter`.

This approach has a hard ceiling: MongoDB BSON documents are capped at 16 MB. With UUID strings (~49 bytes each), that ceiling is hit at roughly 342 000 matching documents — a realistic limit for production datasets with heavy metadata filtering.

## Proposed API

```ts
export interface CreateIndexParams {
  indexName: string;
  dimension: number;
  metric?: 'cosine' | 'euclidean' | 'dotproduct';
  /**
   * Metadata field names that will be used as filter predicates during queries.
   * Adapters that support native index-level filter declarations (e.g. MongoDB
   * Atlas Vector Search) will register these paths in the index definition so
   * that filtered queries bypass the pre-filter materialisation step.
   *
   * Adapters that do not support this hint ignore it silently.
   */
  filterFields?: string[];
}
```

## MongoDB adapter behaviour

With `filterFields` provided, the MongoDB adapter adds a `{ type: 'filter', path: 'metadata.<field>' }` entry for each declared field in the Atlas vectorSearch index at creation time. Queries on those metadata fields are then passed directly to `$vectorSearch.filter` — no `$match`, no `$in`, no BSON ceiling.

Without `filterFields`, the adapter falls back to the current pre-filter approach for metadata fields. (The `document` field is already hard-coded as a declared filter field because it is always known at index creation time.)

## Usage example

```ts
await vectorStore.createIndex({
  indexName: 'my-index',
  dimension: 1536,
  metric: 'cosine',
  filterFields: ['category', 'tenant_id'],
  // MongoDB adapter registers these as:
  //   { type: 'filter', path: 'metadata.category' }
  //   { type: 'filter', path: 'metadata.tenant_id' }
});
```

## Affected packages

- **`@mastra/core`** — add `filterFields` to `CreateIndexParams`
- **`@mastra/mongodb`** — consume `filterFields` in `createIndex` to populate Atlas vectorSearch filter field definitions

Other vector adapters (Pinecone, pgvector, etc.) that have no equivalent concept ignore the field silently.

serhiileniv:
## Description

MongoDB Atlas Vector Search can only filter natively on fields declared in the index definition. Since metadata fields are unknown at index-creation time, the adapter pre-filters in two steps — `$match` the metadata to collect matching `_id`s, then pass `{ _id: { $in: [...] } }` to `$vectorSearch`. That `$in` array is bounded by the 16 MB BSON document limit, which caps metadata-filtered queries at roughly 342,000 matching documents.

This adds an optional `filterFields` to the MongoDB adapter's `createIndex` (via a new exported `MongoDBCreateIndexParams extends CreateIndexParams`, following the same adapter-specific extension pattern as Qdrant and Lance — no change to `@mastra/core`).

Before:

```ts
await vectorStore.createIndex({ indexName: 'my-index', dimension: 1536 });
// metadata-filtered queries always pre-filter via $match + { _id: { $in: [...] } }
```

After:

```ts
await vectorStore.createIndex({
  indexName: 'my-index',
  dimension: 1536,
  filterFields: ['category', 'tenant_id'],
});
// queries filtering on category / tenant_id are pushed straight into
// $vectorSearch.filter — no $match, no $in, no BSON ceiling
```

Details:

- **`MongoDBCreateIndexParams`** — `createIndex` registers each declared field as a `{ type: 'filter', path: 'metadata.<field>' }` entry in the Atlas vectorSearch index.
- **Pushdown safety** — `query` takes the fast path (filter passed directly to `$vectorSearch`) **only** when it can prove it's safe: every field the filter touches is a declared filter field, and every operator is one Atlas Vector Search accepts inside `filter` (`$and $or $eq $ne $gt $gte $lt $lte $in $nin`). Anything else — an undeclared field, or an operator like `$regex` / `$exists` — falls back to the existing `$match` pre-filter, so behavior is unchanged when the hint isn't (fully) applicable.
- **Readiness-gated discovery** — query-only processes hydrate the declared paths from the live index definition. Because `latestDefinition` is the most recently *requested* definition (queries may still be served by the previous one while an update builds), the definition is only trusted — and cached — once the index reports `READY`. Until then queries take the always-correct fallback, and nothing is cached so a later query re-reads the definition.
- **Independent `IndexAlreadyExists` handling** — the vector and companion full-text index creations each treat `IndexAlreadyExists` as a no-op independently, so an existing vector index no longer skips creation of the full-text index (previously a shared catch could leave the pair in a partial state).

The `document` field is still always declared, so `documentFilter` queries are unaffected.

## Related issue(s)

Fixes #18587

## Type of change

- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [ ] Test update

## Checklist

_Disclosure: I used AI assistance (Claude Code) to investigate and prepare this change._

## ELI5

You can now tell MongoDB which “metadata” fields you usually filter on, so Atlas Vector Search can apply those filters directly inside the search. If the filter can’t be safely pushed down, it falls back to the older method that first finds matching document IDs.

## What changed

- Added MongoDB-specific `filterFields` support to `MongoDBVector.createIndex()` via `MongoDBCreateIndexParams` (extends `CreateIndexParams`).
- Registers declared metadata filter fields as Atlas Vector Search filter paths (`metadata.<field>`), while keeping built-in `document` as a filter field.
- Uses native filter pushdown by sending compatible filters directly in `$vectorSearch.filter`.
- Keeps the existing fallback pipeline when:
  - a filter references undeclared fields,
  - operators/structures aren’t supported for pushdown,
  - or the index isn’t `READY` yet (prevents unsafe cache/definition usage).
- Caches declared filter paths only after the index definition reaches `READY`, and clears the cache on index deletion.
- Ensures vector and full-text index creation handle `IndexAlreadyExists` independently (so one doesn’t block the other).
- Added unit tests for index definition discovery (including building/race cases), filter validation, caching behavior, and pushdown vs fallback execution.
- Updated MongoDB vector docs and added changesets for the MongoDB-only option and index-creation fix.


stores/mongodb/src/vector/index.ts
```diff
@@ -868,6 +919,76 @@ export class MongoDBVector extends MastraVector<MongoDBVectorFilter> {
...
+    const cached = this.declaredFilterPaths.get(indexName);
+    if (cached) return cached;
+
+    const collection = await this.getCollection(indexName, true);
+    const indexNameInternal = `${indexName}_vector_index`;
+    const indexInfo: any[] = await (collection as any).listSearchIndexes().toArray();
+    const indexData = indexInfo.find((idx: any) => idx.name === indexNameInternal);
+
+    const paths = new Set<string>();
+    for (const field of indexData?.latestDefinition?.fields ?? []) {
```
NikAiyer: `latestDefinition` is the most recently requested Atlas Search definition, but it may not yet be the definition serving queries. While an index update is building, the active `mainIndex` can still use the previous definition.

If `latestDefinition` adds `metadata.category`, this code caches that path and permits direct pushdown even though the active index does not support it yet. The resulting `$vectorSearch.filter` can fail rather than taking the candidate-ID fallback.

Could we gate discovery and caching on the index being `READY`, leaving non-ready or incomplete results uncached? Please also add a regression test where a building `latestDefinition` contains a new filter field but the active definition does not; it should use the fallback until the index becomes ready.

packages/core/src/vector/types.ts
```diff
@@ -60,6 +60,15 @@ export interface CreateIndexParams {
...
   metric?: 'cosine' | 'euclidean' | 'dotproduct';
+  /**
+   * Metadata field names that will be used as filter predicates during queries.
+   * Adapters that support native index-level filter declarations (e.g. MongoDB
+   * Atlas Vector Search) will register these paths in the index definition so
+   * that filtered queries bypass the pre-filter materialisation step.
+   *
+   * Adapters that do not support this hint ignore it silently.
+   */
+  filterFields?: string[];
```
NikAiyer: Could this be a MongoDB-specific extension instead?

```ts
interface MongoDBCreateIndexParams extends CreateIndexParams {
  filterFields?: string[];
}
```

`filterFields` currently describes Atlas-specific index declarations and is consumed only by MongoDB. Other stores model filterability differently. For example, they may use typed payload indexes, inverse non-filterable keys, or externally deployed schemas. Silently ignoring this on every other adapter does not yet establish a meaningful shared contract.

Lance and Qdrant already use adapter-specific `CreateIndexParams` extensions. We can promote this to core later if multiple stores converge on portable semantics. If moved, the core changeset should also be removed.

NikAiyer (changes requested):
Thanks. The direct pushdown approach addresses the candidate-ID materialization and BSON-size problem while preserving the existing fallback for unsupported filters.

The staged-definition correctness race described inline needs readiness-gated discovery and regression coverage before merge.

I also recommend keeping `filterFields` MongoDB-specific until another adapter demonstrates compatible semantics. Separately, the broad `IndexAlreadyExists` catch around both index-creation calls can leave the companion full-text index missing; handling the two creations independently would avoid that partial state.

stores/mongodb/src/vector/index.ts
serhiileniv: Done in a4361ec6. `getDeclaredFilterPaths` now only trusts — and caches — `latestDefinition` once the index reports `READY`. For a missing or non-READY index it returns an empty set without caching, so queries take the candidate-ID fallback and a later call re-reads the definition instead of pushdown being permanently disabled.

Added the regression tests you described: a `BUILDING` `latestDefinition` containing a new filter field is not trusted or cached (unit level), and a query filtering on that field builds the fallback $match pipeline until the index becomes `READY` (query level).

packages/core/src/vector/types.ts
serhiileniv: Agreed — moved it to `MongoDBCreateIndexParams extends CreateIndexParams`, exported from `@mastra/mongodb`, mirroring the existing Qdrant/Lance adapter-specific param pattern. The `@mastra/core` change and its changeset are reverted, so the PR now only releases `@mastra/mongodb`. Happy to revisit promoting it to core if other stores converge on portable semantics.

serhiileniv:
Thanks for the review — all three points addressed in a4361ec6:

1. **Readiness-gated discovery**: `getDeclaredFilterPaths` only trusts and caches the index definition once it reports `READY`; otherwise it returns an uncached empty set, so queries take the candidate-ID fallback until the update is live. Regression tests added at both the discovery and query level (building `latestDefinition` with a new filter field → fallback pipeline).
2. **Scope**: `filterFields` moved off core `CreateIndexParams` onto `MongoDBCreateIndexParams extends CreateIndexParams` in `@mastra/mongodb`, following the Qdrant/Lance pattern. Core changeset removed — the PR now only releases `@mastra/mongodb`.
3. **`IndexAlreadyExists`**: the vector and companion full-text index creations now each treat `IndexAlreadyExists` as a no-op independently, so an existing vector index no longer skips creating the full-text index. Covered by tests and a separate patch changeset.

PR title and description updated to match the narrower scope.

NikAiyer (approved):

PR #19009, opened 2026-07-06
fix(core): compose stopWhen with maxSteps instead of replacing it

### LINKED ISSUE #19007: [BUG] Custom stopWhen is discarded when maxSteps is also set
### Describe the Bug

When both `maxSteps` and a custom `stopWhen` are passed, the user's `stopWhen` is silently discarded and replaced by `stepCountIs(maxSteps)`. The agent can no longer stop early on the user's condition and instead runs all the way to the `maxSteps` cap.

`packages/core/src/llm/model/model.loop.ts`:

```ts
let stopWhenToUse;
if (maxSteps && typeof maxSteps === 'number') {
  stopWhenToUse = stepCountIs(maxSteps); // user's stopWhen thrown away
} else {
  stopWhenToUse = stopWhen;
}
```

`maxSteps` is just sugar for the stop condition `stepCountIs(maxSteps)`, and the loop is designed to combine stop conditions with OR (`packages/core/src/loop/workflows/agentic-loop/index.ts` normalizes `stopWhen` to an array and uses `conditions.some(...)`). The agent layer forwards both `maxSteps` and `stopWhen` to the loop, so both are meant to be honored, not one replacing the other.

### Steps To Reproduce

```ts
let n = 0
await agent.stream('go', {
  maxSteps: 10,
  stopWhen: () => { n++; return n >= 2 }, // should stop after 2 steps
})
```

Use a model that always emits a tool call so it never finishes on its own.

### Expected Behavior

The loop stops after 2 model calls (the custom `stopWhen` fires first), with `maxSteps: 10` acting only as an upper safety cap.

### Actual Behavior

The custom `stopWhen` is ignored and the loop runs 10 model calls.

### Additional Context

Fix: compose the conditions when both are set, `[stepCountIs(maxSteps), ...userConditions]`, and move the default `stepCountIs(5)` so it applies only when neither `maxSteps` nor `stopWhen` is provided (otherwise a `maxSteps`-only run would be wrongly capped at 5). I have a PR ready with a scenario test.

abhay-codes07:
## What

When both `maxSteps` and a custom `stopWhen` are passed, the user's `stopWhen` is silently discarded and replaced by `stepCountIs(maxSteps)`, so the agent cannot stop early and runs all the way to the `maxSteps` cap.

```ts
// packages/core/src/llm/model/model.loop.ts
if (maxSteps && typeof maxSteps === 'number') {
  stopWhenToUse = stepCountIs(maxSteps); // user's stopWhen thrown away
} else {
  stopWhenToUse = stopWhen;
}
```

`maxSteps` is sugar for `stepCountIs(maxSteps)`, and the loop composes stop conditions with OR (`agentic-loop/index.ts` normalizes `stopWhen` to an array and uses `conditions.some(...)`). The agent forwards both, so they should combine, not replace. Example: `agent.stream('go', { maxSteps: 10, stopWhen: hasToolCall('done') })` ignores `stopWhen` and runs to 10.

## Change

Compose `[stepCountIs(maxSteps), ...userConditions]` when both are set, and move the default `stepCountIs(5)` into the `else` branch so a `maxSteps`-only run still caps at `maxSteps` (not `min(maxSteps, 5)`).

## Tests

Added a scenario test: with `maxSteps: 10` and a custom `stopWhen` that fires at step 2, the loop stops at 2 model calls; and a `maxSteps: 3`-only run still caps at 3. The custom-stopWhen case fails on current code (runs 10) and passes after the fix. The existing max-steps-edge-cases scenario (20 tests) still passes.

Closes #19007

---

Hey @abhiaiyer91, small loop control-flow fix. The reference is the OR-composition the loop already does for multiple stop conditions. Thanks for the look.

## ELI5
This change makes the loop listen to your “stop” rule even if you also set a maximum number of steps. Before, setting `maxSteps` could accidentally hide your custom stop condition; now both work together, so the loop can stop early for your reason but still has a safety limit.

### Summary
- Fixed loop control flow so `maxSteps` no longer overwrites a custom `stopWhen`.
- Combined both stop conditions when both are provided, while keeping `maxSteps` as an upper cap.
- Moved the default `stepCountIs(5)` fallback so it only applies when neither `maxSteps` nor `stopWhen` is set.
- Added scenario tests for:
  - `maxSteps` + custom `stopWhen`
  - `maxSteps` only


abhiaiyer91:
You don't need to mention me directly in your PR comment. We have CODEOWNERs for a reason, thanks for the PR we'll take a look

abhiaiyer91 (approved):

Issue #18877, opened 2026-07-03
[BUG] channels: MastraStateAdapter cache (dedupe/modal-context) is per-process — duplicate Slack replies when running multiple instances


mstysk:
## Describe the bug

When running a Mastra server with agent channels (Slack adapter) on **multiple instances behind a load balancer**, every Slack mention gets processed twice — duplicate progress messages, duplicate sub-agent calls, and duplicate final replies.

## Root cause

Slack sends **two events** for a single mention (`app_mention` + `message` via the `message.channels` event subscription). The chat SDK deliberately relies on the state adapter to collapse them (`handleIncomingMessage` dedupes via `stateAdapter.setIfNotExists("dedupe:${adapter.name}:${message.id}", ...)`).

However, the default `MastraStateAdapter` (`@mastra/core/channels`) only persists **subscriptions** to Mastra storage. The cache — which backs dedupe keys, modal contexts, and thread state — is a **per-process in-memory Map** ("In-memory ephemeral state (cache, locks, lists, queues)").

With 2+ instances, the load balancer routes the two Slack events to different instances (~45ms apart in our case). Each instance sees the message as "first time" and runs the full pipeline. Slack's 3-second-timeout retries (`x-slack-retry-num`) fall into the same hole.

Besides dedupe, `modal-context:*` keys are also per-process, so a modal submit routed to a different instance loses its context (`relatedMessage`/`relatedThread` become undefined).

## Steps to reproduce

1. Configure an agent with `channels: { adapters: { slack: ... } }` and Postgres storage
2. Run 2 server instances behind a load balancer (e.g. ECS + ALB)
3. Mention the bot in a Slack channel
4. Both instances process the message; the bot replies twice

## Expected behavior

The dedupe cache should be shared across instances — e.g. persisted through the Mastra storage layer with an atomic `setIfNotExists` (`INSERT ... ON CONFLICT`), or the docs should state that `ChannelConfig.state` must be a shared StateAdapter when running multiple instances.

## Workaround

Inject a custom `state` (the official `ChannelConfig.state` extension point) that stores the cache methods (`get`/`set`/`setIfNotExists`/`delete`) in Postgres with a conditional upsert, and delegates the rest to `MastraStateAdapter`. Locks/queues/lists are unaffected in practice since channels hardcodes `concurrency: { strategy: "concurrent" }`.

## Environment

- @mastra/core 1.48.0, chat 4.31.0, @chat-adapter/slack 4.31.0
[...]

PR #17049, opened 2026-05-26
fix(channels): isolate thread subscriptions per agent

Akash504-ai:
## Description

Fixes channel subscription state leaking across agents/bots sharing the same Mastra storage adapter.

Previously, `MastraStateAdapter` looked up subscriptions only by `channel_externalThreadId`, which caused multiple agents using the same storage backend to share subscription state unintentionally. As a result, different bots could respond to the same incoming message if they shared the same external thread ID (such as a Telegram `chatId`).

This PR scopes external thread IDs by `agentId` when storing and resolving thread subscriptions, ensuring subscriptions remain isolated per agent.

Also adds regression tests covering multi-agent isolation behavior.

## Related issue(s)

Fixes #17037

## Type of change

* [x] Bug fix (non-breaking change that fixes an issue)
* [ ] New feature (non-breaking change that adds functionality)
* [ ] Breaking change (fix or feature that would cause existing functionality to change)
* [ ] Documentation update
* [ ] Code refactoring
* [ ] Performance improvement
* [x] Test update

## Checklist

* [x] I have linked the related issue(s) in the description above
* [ ] I have made corresponding changes to the documentation (if applicable)
* [x] I have added tests that prove my fix is effective or that my feature works
* [ ] I have addressed all Coderabbit comments on this PR

## ELI5

Two different bots were accidentally sharing "who's listening" notes because they used the same storage. This change tags each bot's notes with its own name so bots no longer see each other's subscriptions.

## What Changed

This PR fixes a bug where channel thread subscription state leaked across agents that share the same Mastra storage backend by scoping stored external thread IDs to the agent that owns them. Previously, MastraStateAdapter looked up threads only by channel_externalThreadId, causing subscriptions to be treated as global across agents sharing the same storage. The adapter now prefixes external thread IDs with the agent ID (format: `${agentId}:${externalThreadId}`) when storing and resolving thread mappings so subscription state remains isolated per agent.

- Root cause: MastraStateAdapter.findThreadByExternalId (and related subscription checks) queried by raw external thread ID without scoping by agentId.
- Solution: Add agent scoping to external thread IDs and update subscription logic to use scoped IDs.

## Files Modified

- packages/core/src/channels/state-adapter.ts
  - MastraStateAdapter constructor now accepts agentId (constructor(memoryStore, agentId)).
  - Added agentId field and helper to produce scoped thread IDs.
  - subscribe(), unsubscribe(), and isSubscribed() now resolve threads using the scoped external thread ID.

[...]

PR #19014, opened 2026-07-06
fix(core): scope channel threads per agent and ignore bot messages by default

### LINKED ISSUE #17037: [BUG] Thread Subscription State Bleeds Across Agents/Bots When Sharing a Mastra Storage Adapter
When running multiple separate agents with active Telegram adapters (or other channel-based adapters) in a single Mastra application, all agents/bots share the same thread subscription state in the database. 

If a thread/chat is subscribed to by *one* bot, **every other bot** sharing the same Mastra storage instance sees itself as subscribed to that same thread ID (the external `chatId` in Telegram's case). 

Consequently, if a message is sent to *any* of the bots, all initialized bots believe they are subscribed to that thread and trigger their handlers, causing multiple bots to reply to the same message concurrently.

---

## Technical Root Cause Analysis
Mastra's channel subscription mechanism is implemented via `AgentChannels` and backed by `MastraStateAdapter` (unless a custom state adapter is specified).

Looking at the `MastraStateAdapter` implementation:

```typescript
// @mastra/core/dist/channels/state-adapter.ts
export declare class MastraStateAdapter implements StateAdapter {
    // ...
    async subscribe(threadId: string): Promise<void> {
        const thread = await this.findThreadByExternalId(threadId);
        if (!thread) return;
        await this.memoryStore.updateThread({
            id: thread.id,
            title: thread.title ?? "",
            metadata: { ...thread.metadata, channel_subscribed: "true" }
        });
    }

    async isSubscribed(threadId: string): Promise<boolean> {
        const thread = await this.findThreadByExternalId(threadId);
        if (!thread) return false;
        return thread.metadata?.channel_subscribed === "true";
    }

    private async findThreadByExternalId(externalThreadId: string) {
        const { threads } = await this.memoryStore.listThreads({
            filter: { metadata: { channel_externalThreadId: externalThreadId } },
            perPage: 1
        });
        return threads[0] ?? null;
    }
}
```

### The Problem:
1. When any adapter (such as Telegram) receives an incoming event, the thread identifier (e.g. Telegram's `chatId` or `channelId`) is passed as `threadId` to `isSubscribed(threadId)`.
2. `findThreadByExternalId(threadId)` searches Mastra's shared database for **any thread** where `channel_externalThreadId` matches the Telegram `chatId`.
3. In a multi-agent application, all agents are registered under the same `Mastra` instance and share the same storage backend (e.g., SQLite via `LibSQLStore`).
4. Since `channel_externalThreadId` is not scoped by **Agent ID** or **Bot Token/ID**, if Bot A creates a thread for `chatId: "12345678"` and subscribes to it, that thread is stored in the shared database with `channel_externalThreadId: "12345678"` and `channel_subscribed: "true"`.
5. When a message is sent to Bot B's chat (which happens to have the same `chatId: "12345678"` because it's the same Telegram user), `MastraStateAdapter.isSubscribed("12345678")` will query the shared DB, find the thread created by Bot A, see `channel_subscribed: "true"`, and return `true`.
6. Therefore, Bot B's channel processor believes it is subscribed to the thread, causing both Bot A and Bot B to respond to the incoming event simultaneously.

---

## Proposed Solutions / Mitigations
1. **Scope the Thread External ID by Agent/Bot**:
   Scope lookups and metadata keys inside `MastraStateAdapter` by prefixing the external thread ID with the `agentId` or a hash of the bot's credentials:
   ```typescript
   metadata: { 
     channel_externalThreadId: `${agentId}:${externalThreadId}`,
     channel_subscribed: "true" 
   }
   ```
2. **Allow Isolated / Custom State Adapters Per Agent**:
   Ensure `AgentChannels` can easily receive/use an isolated state adapter instance (or fall back to an in-memory/isolated adapter) so that subscription data does not automatically merge into a global, un-scoped database lookup.

***

## How to Reproduce
1. Initialize a Mastra app with a shared SQLite store (`LibSQLStore`).
2. Create two separate agents (e.g., `agentA` and `agentB`).
3. Configure each agent with its own unique Telegram bot token:
   ```typescript
   import { createTelegramAdapter } from "@chat-adapter/telegram";

   const agentA = new Agent({
     name: "agent-a",
     instructions: "You are Agent A",
     model: "google/gemini-2.5-pro",
     channels: {
       adapters: {
         telegram: createTelegramAdapter({
           botToken: process.env.TELEGRAM_BOT_TOKEN_A,
           mode: "webhook",
         })
       }
     }
   });

   const agentB = new Agent({
     name: "agent-b",
     instructions: "You are Agent B",
     model: "google/gemini-2.5-pro",
     channels: {
       adapters: {
         telegram: createTelegramAdapter({
           botToken: process.env.TELEGRAM_BOT_TOKEN_B,
           mode: "webhook",
         })
       }
     }
   });

     agents: { agentA, agentB },
     storage: new LibSQLStore({ ... }),
   });
   ```
4. Send a message to `Bot A` to trigger a thread creation and subscription.
5. Send a message to `Bot B` from the same Telegram account.
6. Observe that both `Bot A` and `Bot B` reply to the Telegram chat because they both resolve the subscription check to `true`.

n/a

### Expected Behavior

Each bot channel should have independent user/thread state.

### Environment Information

```block
System:
    OS: macOS 15.5
    CPU: (8) arm64 Apple M1
    Memory: 130.88 MB / 8.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 22.17.0 - /Users/agonbina/.nvm/versions/node/v22.17.0/bin/node
    npm: 10.9.2 - /Users/agonbina/.nvm/versions/node/v22.17.0/bin/npm
    pnpm: 10.12.4 - /Users/agonbina/.nvm/versions/node/v22.17.0/bin/pnpm
    bun: 1.2.18 - /Users/agonbina/.bun/bin/bun
  Browsers:
    Chrome: 148.0.7778.179
    Safari: 18.5
  npmPackages:
    @chat-adapter/telegram: ^4.29.0 => 4.29.0 
    @composio/core: ^0.10.0 => 0.10.0 
    @mastra/core: ^1.36.0 => 1.36.0 
    @mastra/duckdb: ^1.4.0 => 1.4.0 
    @mastra/evals: ^1.2.2 => 1.2.2 
    @mastra/libsql: ^1.11.1 => 1.11.1 
    @mastra/loggers: ^1.1.1 => 1.1.1 
    @mastra/mcp: ^1.8.0 => 1.8.0 
    @mastra/memory: ^1.19.0 => 1.19.0 
    @mastra/observability: ^1.13.0 => 1.13.0 
    @types/node: ^25.9.1 => 25.9.1 
    ai-sdk-provider-gemini-cli: ^2.0.1 => 2.0.1 
    imapflow: ^1.3.3 => 1.3.3 
    mastra: ^1.10.0 => 1.10.0 
    mongodb: ^7.2.0 => 7.2.0 
    mongodb-memory-server: ^11.1.0 => 11.1.0 
    typescript: ^6.0.3 => 6.0.3 
    zod: ^3 => 3.25.76
```

### Verification

CalebBarnes:
Fixes #17037

## Problem

When two or more agents (bots) share one Mastra storage instance and receive messages on the same platform thread (Slack thread, Telegram chat, Discord channel), they all map it to a **single** Mastra thread. `getOrCreateThread()` and `MastraStateAdapter.findThreadByExternalId()` filter only on `{ channel_platform, channel_externalThreadId, channel_externalChannelId }` — no agent scoping. Consequences:

- Shared conversation history and a `resourceId` race between agents
- The `channel_subscribed` flag is one boolean on that shared thread, so one bot subscribing makes **every** bot reply to subsequent messages (the bug in #17037)
- Broadcast/schedule delivery cross-contaminates between agents

Separately: `handleChatMessage` has no bot-author guard. The Chat SDK only filters a bot's *own* messages (`isMe`), so another bot's message in a subscribed thread wakes the agent. With per-agent threads, two subscribed bots replying to each other is a real infinite-loop risk.

## Fix

**Part 1 — per-agent thread identity.** Thread identity becomes `(platform, externalThreadId, channelId, agentId)` via a new `channel_agentId` metadata key:

- `getOrCreateThread()` stamps `channel_agentId` on new threads and includes it in lookups
- `MastraStateAdapter` takes an optional `agentId` ctor param and scopes `findThreadByExternalId()` the same way; subscription isolation follows for free since `channel_subscribed` now lives on a per-agent thread
- **Legacy migration:** on a scoped miss, an unclaimed legacy thread (no `channel_agentId`) is claimed and stamped by the first agent that touches it — single-agent deployments keep their history seamlessly. Threads already claimed by another agent are never stolen.
- `channel_externalThreadId` stays **pure** (the real platform thread ID), so `buildRenderContextForThread()` and the schedule/broadcast path keep working unchanged

**Part 2 — bot-wake guard.** New `ChannelConfig` option `respondToBots?: boolean` (default `false`). `handleChatMessage` returns early for `author.isBot === true` unless opted in. `isBot: 'unknown'`/`false` pass through so humans on adapters that can't classify are never blocked.

> ⚠️ **Behavior change:** agents no longer auto-reply to other bots' messages by default. Bot-to-bot replies are now opt-in via `respondToBots: true`.

## Relationship to #17049

#17049 addresses the same issue by mangling the stored external ID to `${agentId}:${externalThreadId}`. That breaks the self-sufficient broadcast path (`buildRenderContextForThread()` would materialize a nonexistent platform thread), orphans all existing channel threads on upgrade (no legacy fallback), and changes the meaning of `channel_externalThreadId` for everything else that reads it. This PR keeps the external ID pure and scopes identity with a separate metadata key instead.

## Test plan

- 12 new tests: cross-agent thread isolation, legacy claim + stamp, no-steal of claimed threads, subscription isolation at the state-adapter level, and the full `respondToBots` matrix (`true`/`false`/`'unknown'`/default)
- Red/green proof: copying the new test files onto a clean `origin/main` checkout yields **6 failed / 75 passed** — each failure matches the bug (shared thread, cross-agent subscription bleed, bot message waking the agent). On this branch, all 252 channels tests pass.
- `pnpm --filter ./packages/core check` clean
- Docs updated (guide + reference) with per-agent scoping, migration note, and `respondToBots`

## Out of scope

#18877 (duplicate replies on multi-instance deployments) is a different bug in the same area — the in-memory dedupe cache needs a storage-backed atomic primitive. The `agentId` handle added to `MastraStateAdapter` here is the scoping hook that follow-up will need for agent-scoped dedupe keys.

### DIFF
--- a/.changeset/per-agent-channel-threads.md
+++ b/.changeset/per-agent-channel-threads.md
@@ -1,0 +1,19 @@
+---
+'@mastra/core': minor
+---
+
+Scoped channel threads per agent so multiple agents sharing one storage instance no longer bleed conversation history and subscription state into each other's threads. Each agent now maps a platform thread (Slack, Discord, Telegram) to its own Mastra thread via a `channel_agentId` metadata key; threads created before this change are claimed by the first agent that touches them, keeping existing history. Fixes thread subscription state bleeding across agents.
+
+Added a `respondToBots` option to the channels config (default `false`). Agents now ignore messages authored by other bots in the default handlers, preventing bot-to-bot reply loops when multiple bots share a channel or thread. This is a behavior change: previously another bot's message in a subscribed thread would wake the agent. Opt back in with `respondToBots: true`:
+
+```typescript
+export const agent = new Agent({
+  id: 'agent',
+  channels: {
+    adapters: {
+      slack: createSlackAdapter(),
+    },
+    respondToBots: true,
+  },
+})
+```
--- a/docs/src/content/en/docs/agents/channels.mdx
+++ b/docs/src/content/en/docs/agents/channels.mdx
@@ -132,6 +132,28 @@ In group conversations, Mastra automatically prefixes each message with the send
 [Bob (@U456DEF)]: I have a question too.
 ```
 
+## Multiple agents on one platform
+
+Channel threads are scoped per agent. When two agents (for example, a support bot and a triage bot) are both active in the same Slack or Discord thread, each agent gets its own Mastra thread with independent conversation history and subscription state. One agent subscribing to a platform thread doesn't affect the other.
+
+Threads created before per-agent scoping have no owning agent recorded. The first agent to touch such a thread claims it and keeps its history; other agents get fresh threads.
+
+By default, agents ignore messages authored by other bots. This prevents two subscribed bots from replying to each other in a loop. Set `respondToBots: true` to opt in:
+
+```typescript
+export const agent = new Agent({
+  id: 'agent',
+  channels: {
+    adapters: {
+      slack: createSlackAdapter(),
+    },
+    respondToBots: true,
+  },
+})
+```
+
+The guard applies to the built-in handlers, so custom `handlers` that call `defaultHandler` inherit it. Handlers that do their own routing bypass it.
+
 ## Multimodal content
 
 Models like Gemini can natively process images, video, and audio. Combine `inlineMedia` and `inlineLinks` to let users share rich content with your agent across platforms:
--- a/docs/src/content/en/reference/agents/channels.mdx
+++ b/docs/src/content/en/reference/agents/channels.mdx
@@ -52,6 +52,14 @@ The `channels` property accepts a `ChannelConfig` object with the following fiel
     description: 'Override default message handlers for DMs, mentions, and subscribed threads.',
   },
   {
+    name: 'respondToBots',
+    type: 'boolean',
+    isOptional: true,
+    defaultValue: 'false',
+    description:
+      "Whether messages authored by other bots wake the agent. By default the built-in handlers ignore bot-authored messages, preventing reply loops when multiple bots share a channel or thread. Messages the adapter can't classify pass through. The guard lives in the default handler, so custom `handlers` that call `defaultHandler` inherit it.",
+  },
+  {
     name: 'inlineMedia',
     type: 'string[] | ((mimeType: string) => boolean)',
     isOptional: true,
--- a/packages/core/src/channels/__tests__/agent-channels.test.ts
+++ b/packages/core/src/channels/__tests__/agent-channels.test.ts
@@ -685,6 +685,231 @@ describe('AgentChannels', () => {
       );
     });
   });
+
+  describe('per-agent thread scoping', () => {
+    function makeMastra() {
+      const db = new InMemoryDB();
+      const memoryStore = new InMemoryMemory({ db });
+      return {
+        getStorage: () => ({ getStore: () => memoryStore }),
+        getServer: () => null,
+      } as any;
+    }
+
+    function makeChatThread(adapter: any) {
+      return {
+        id: 'channel-1:thread-1',
+        channelId: 'channel-1',
+        isDM: false,
+        adapter,
+        isSubscribed: vi.fn().mockResolvedValue(true),
+        subscribe: vi.fn().mockResolvedValue(undefined),
+        mentionUser: vi.fn((userId: string) => `<@${userId}>`),
+        messages: (async function* () {})(),
+      } as any;
+    }
+
+    const message = {
+      id: 'message-1',
+      text: 'hi',
+      author: { userId: 'user-1', userName: 'tyler', fullName: 'Tyler Barnes' },
+      attachments: [],
+    } as any;
+
+    function makeChannels(agentId: string) {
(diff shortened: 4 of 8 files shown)


CalebBarnes:
Superseded by #21288, a fresh implementation of the per-agent thread identity design against current main (channel_ownerId metadata key, scoped lookups, lazy legacy adoption). The respondToBots part of this draft was intentionally dropped: users can check author.isBot in their handlers today.

PR #19021, opened 2026-07-07
fix(core): read v5 mediaType/url on file parts across message-list read sites

### LINKED ISSUE #19020: Memory persist throws on an AI SDK v5-shaped file part (mediaType) in message-list converters
### What happened

Persisting a conversation to memory fails with `AGENT_MEMORY_PERSIST_RESPONSE_MESSAGES_FAILED` ("Error saving memory on finish") when a **user** message carries a file part in the AI SDK **v5** shape (`{ type: 'file', data, mediaType }`, no `mimeType`).

The underlying error is:

> If the attachment is not an image or text, it must specify a content type

### Root cause

AI SDK v5 renamed the media-type field `mimeType` → `mediaType`. Mastra stores messages in the v2 (AI-SDK-v4) shape and converts them on persist through v4 code that reads `part.mimeType`. For a v5-shaped file part `part.mimeType` is `undefined`, so `convertToV1Messages` builds an attachment with `contentType: undefined` (`packages/core/src/agent/message-list/prompt/convert-to-mastra-v1.ts:81`), and `attachmentsToParts` throws at `packages/core/src/agent/message-list/prompt/attachments-to-parts.ts:69`.

This is the same class of bug as #17366, which fixed the Studio renderer (`to-assistant-ui-message.ts`) to read `mediaType ?? mimeType` / `url ?? data`. The memory-persist / message-list converters were left uncovered.

The same untyped read has a second symptom in `CacheKeyGenerator.fromAIV4Part` (`packages/core/src/agent/message-list/cache/CacheKeyGenerator.ts:95`): a v5-shaped file part yields `undefined` for both `data` and `mimeType`, so two *distinct* v5 file parts derive the **same** cache key and `MessageMerger` can wrongly treat them as duplicates. Same root cause, different symptom.

### Minimal reproduction (unit)

```ts
import { convertToV1Messages } from '@mastra/core/agent'; // message-list export path

convertToV1Messages([
  {
    id: 'm1', role: 'user', createdAt: new Date(), threadId: 't1', resourceId: 'r1',
    content: {
      format: 2,
      parts: [{ type: 'file', data: 'JVBERi0xLjQ=', mediaType: 'application/pdf' } as any],
    },
  },
]);
// throws: "If the attachment is not an image or text, it must specify a content type"
```

### Expected

v5-shaped file parts persist without throwing; the media type is preserved.

### Environment

`@mastra/core` (reproduces on `1.48.0` and on `main` @ HEAD — the read sites are byte-identical).

rares04:
## Description

Message converters in `@mastra/core`'s message-list read a file part's media type from the AI SDK **v4** fields (`mimeType`/`data`). AI SDK **v5** renamed this to `mediaType`/`url`. When a **v5-shaped file part** (`{ type: 'file', data, mediaType }`, no `mimeType`) is persisted, `convertToV1Messages` builds an attachment with `contentType: undefined`, and AI-SDK-v4 `attachmentsToParts` throws:

> If the attachment is not an image or text, it must specify a content type

surfaced as `AGENT_MEMORY_PERSIST_RESPONSE_MESSAGES_FAILED` ("Error saving memory on finish"). The throw is reached via the **user** branch of `convertToV1Messages` (the sole caller of `attachmentsToParts`); "response" is the persist step, not the throwing message.

The same untyped read has a second symptom: `CacheKeyGenerator.fromAIV4Part` derives a cache key from `part.data` + `part.mimeType`, both `undefined` for a v5-shaped part, so two *distinct* v5 file parts collapse to the same key and `MessageMerger` can wrongly dedup them.

This is the same bug class as #17366 (which fixed the Studio renderer `to-assistant-ui-message.ts` to read `mediaType ?? mimeType` / `url ?? data`), on the previously-uncovered memory-persist / message-list path.

### Fix
Read both conventions at the message-list v4 file-part read sites, using the same narrow boundary cast #17366 introduced (the stored `MastraMessagePart` union only describes the v4 shape) — `mimeType ?? mediaType`, `data ?? url`:

- `prompt/convert-to-mastra-v1.ts` — **the fix for the throw**.
- `adapters/AIV4Adapter.ts` (`toUIMessage`) — same-class hardening on the render/round-trip path (attachment was built with `contentType: undefined`).
- `adapters/AIV5Adapter.ts` (v2→v5) — same-class hardening so the media type isn't lost to the `image/png` default.
- `cache/CacheKeyGenerator.ts` (`fromAIV4Part`) — fixes the cache-key collapse so distinct v5 file parts key distinctly.

Other `mimeType`/`mediaType` reads under `message-list/` operate on strongly-v4- or strongly-v5-typed contexts (not the ambiguous stored union) and are correctly out of scope.

### Tests
All are genuine fail-before / pass-after:
- `prompt/convert-to-mastra-v1.test.ts` — a v5-shaped user file part round-trips without throwing and keeps `application/pdf`.
- `adapters/AIV4Adapter-file-media-type.test.ts` — v5 file part yields the correct `experimental_attachments.contentType`; v4 parts still work.
- `adapters/AIV5Adapter-file-media-type.test.ts` — v5-shaped stored part → emitted `FileUIPart` carries the correct `mediaType` (not the `image/png` default).
- `cache/CacheKeyGenerator-file-media-type.test.ts` — two v5 file parts differing only in `mediaType`/`url` produce distinct keys via both `fromAIV4Part` and `fromDBParts` (the `MessageMerger` path).

## Related issue(s)

Fixes #19020

## Type of change

## Checklist

## ELI5
This PR teaches the system to understand file attachments stored in two slightly different formats (old vs new). That prevents saves from failing and ensures each file keeps its correct type instead of being accidentally mixed up with other files.

### What changed
- Added a shared `resolveFilePartMediaTypeAndData` helper to read **both** file-part shapes:
  - v4-style: `mimeType` + `data`
  - v5-style: `mediaType` + `url`
- Updated ambiguous “stored-part” read paths to use the helper, including **message conversion** (`convert-to-mastra-v1.ts`) and **agent-controller** logic.
- Updated adapters so attachments preserve the correct MIME/media type:
  - `AIV4Adapter.toUIMessage` now resolves from v4 (`mimeType/data`) or v5 (`mediaType/url`) and sets the resulting attachment `contentType` from the resolved MIME type (avoiding `undefined`).
  - `AIV5Adapter.toUIMessage` similarly resolves MIME type and reads the payload as `data ?? url` when needed.
- Added/used an `application/octet-stream` fallback when a MIME type can’t be determined.
- Fixed cache-key collisions by updating `CacheKeyGenerator.fromAIV4Part` to derive cache-key inputs from either v4 or v5 file fields, preventing incorrect dedup during message merging.
- Updated related typing/normalization (`normalizedUrl`) and rewrote the changeset to be user-facing.

### Tests added/updated
- Regression coverage for adapter behavior across v4/v5 file-part shapes (`AIV4Adapter` and `AIV5Adapter`).
- Conversion coverage for v5-shaped file parts (`convertToV1Messages`).
- Cache-key generation coverage to ensure distinct v5 files don’t collapse into the same key (`CacheKeyGenerator`).
- Added helper-focused tests for `resolveFilePartMediaTypeAndData`.

### Result
- Prevents `AGENT_MEMORY_PERSIST_RESPONSE_MESSAGES_FAILED` for AI SDK v5-shaped file parts.
- Preserves each attachment’s original media/MIME type and payload URL (including URL-carried files).
- Avoids incorrect file deduplication by generating distinct cache keys for distinct files.


minh2416294 (commented):
﻿Hi, I've traced this through and the fix is solid. The throw really does originate at `attachments-to-parts.ts:69`, and reading `mimeType ?? mediaType` in `convert-to-mastra-v1.ts` is what prvents `contentType` from being undefined. Ordering is safe for v4 parts too, since they don't have `mediaType`/`url`. Nice regression test on the CacheKeyGenerator dedup collapse.

One thing I not sure: `AIV5Adapter` seems to get only half the fix. You add the `mediaType` fallback, but every payload read there is still `part.data` (see lines ~389, ~396, ~410-422), while the other three sites read `data ?? url`. The canonical v5 `FileUIPart` (in `@internal/ai-sdk-v5` it's `{ type, mediaType, url, filename }`, no `data`) carries the payload in `url`, so a true v5 part would hit `createDataUri(undefined, ...)`. The AIV5Adapter test puts the payload in `data`, so it wouldn't catch this — whereas your CacheKeyGenerator test correctly uses `url`. Is AIV5Adapter reachable with a `url`-carried part, or is it always normalized upstream first? I'm happy to add a `url`-shaped test case and check against a real v5 file upload if useful.

rares04:
Thanks both, good catches. Update pushed.

@minh2416294 you're right, AIV5Adapter only got half the fix. It read `mediaType` but still took the payload from `data`, so a real v5 `url`-only part would break. It now reads `data ?? url`, and I added a url-carried test. It fails without the change with `data:...;base64,undefined`, exactly as you called it.

On the duplication: I pulled the `mimeType ?? mediaType` / `data ?? url` read into a `resolveFilePartMediaTypeAndData` helper (in `image-utils.ts`) and used it at every site that reads the ambiguous stored-part union. That's the four here plus `agent-controller.ts`, which had the same latent bug (a v5 url-only file part was silently dropped as non-string data). The strongly-typed v4/v5/v6 reads and the writes are left as-is. The `@mastra/memory` token-counter is similar but sits in another package with a different shape, so I left it as a possible follow-up rather than reaching across the boundary.

Also added the octet-stream fallback on the contentType sites, fixed the `normalizedUrl` type, and rewrote the changeset to be user-facing.

abhiaiyer91 (approved):

minh2416294 (commented):
I'm glad the url-only repro reproduced exactly (base64,undefined), ty that you added the test that fails without the change

On the @mastra/memory follow-up you flagged: i agree that it's a separate package and out of scope here, but I had a look and it is the same bug class in two spots, if it helps whoever picks it up:

token-counter.ts:1516, estimateImageAssetTokens(part, part.data, 'file') reads part.data with no ?? part.url, so a v5 url-only part counts as 0 image tokens (the media-type reads at 1359/1389 already do mediaType ?? mimeType, so it's just the payload read left).
observer-agent.ts (~636–746) reads part.mimeType / part.data throughout with no mediaType/url fallback at all, broader than the token counter

Neither crashes (they degrade to under-counting / mis-detecting the image), which is probably why it stayed latent. i can open a follow-up issue and take that one if you'd rather keep this PR scoped

PR #19025, opened 2026-07-07
fix(playground): remove cms block editor focus outline

damien-schneider:
Fixes the unwanted dashed focus outline that appears when focusing CodeMirror-backed `CodeEditor` surfaces, including the embedded CMS instruction block editors and the normal bordered editor shown in Storybook. The root cause is CodeMirror's focused editor outline rendering inside the component chrome.

The fix lives component-wise in `@mastra/playground-ui`: `CodeEditor` now suppresses CodeMirror's internal focused outline for all variants, the default bordered variant uses the same subtle `focus-within` container border treatment as the input family, and `variant="embedded"` remains available for borderless editor surfaces. The CMS inline/ref block editors use that embedded variant with only layout sizing kept locally.

Storybook now includes `Composite/CodeEditor` stories for `Focused Default` and `Focused Embedded`, and the built Storybook index includes `composite-codeeditor--focused-default` and `composite-codeeditor--focused-embedded`. The regression asserts the rendered `.cm-editor` computed outline in jsdom for both default and embedded variants instead of checking class tokens. A patch changeset is included for `@mastra/playground-ui`.

Validated with `pnpm exec prettier --check .changeset/fluffy-coins-wish.md packages/playground-ui/src/ds/components/CodeEditor/code-editor.tsx packages/playground-ui/src/ds/components/CodeEditor/__tests__/code-editor.test.tsx packages/playground-ui/src/ds/components/CodeEditor/__tests__/code-editor-styles.test.tsx packages/playground-ui/src/ds/components/CodeEditor/code-editor.stories.tsx packages/playground/src/domains/agents/components/agent-cms-blocks/agent-cms-block.tsx packages/playground/src/domains/agents/components/agent-cms-blocks/agent-cms-ref-block.tsx`, `pnpm --filter ./packages/playground-ui test -- src/ds/components/CodeEditor/__tests__/code-editor.test.tsx src/ds/components/CodeEditor/__tests__/code-editor-styles.test.tsx --runInBand`, `pnpm --filter ./packages/playground-ui typecheck`, `pnpm --filter ./packages/playground-ui build`, `pnpm --filter ./packages/playground-ui lint`, `pnpm --filter ./packages/playground-ui build-storybook`, `pnpm --filter ./packages/playground typecheck`, `pnpm --filter ./packages/playground lint`, `git diff --check`, and `npx -y react-doctor@latest . --verbose --scope changed --base origin/main`.

## ELI5
When you click into the code editor, CodeMirror was showing a distracting dashed outline. This PR removes that internal outline and instead uses a more subtle “border shows when focused” look—plus it adds an `embedded` mode for borderless editors used inside CMS blocks.

## Summary
- Removed CodeMirror’s internal dashed focus outline from `CodeEditor` by adding high-precedence focus styling extensions in `@mastra/playground-ui`.
- Introduced a `variant` prop for `CodeEditor`:
  - `default`: keeps the bordered editor look, using a subtle outer `focus-within` container border treatment for focus styling.
  - `embedded`: borderless/better blended inline mode for embedded editor surfaces.
- Updated CMS instruction/ref block editors to use `variant="embedded"` (and simplified styling to rely on local sizing like `min-h-12`).
- Added Storybook stories for focused states: `FocusedDefault`, `FocusedEmbedded`, and a wrapped `FocusedEmbeddedInBlock`; updated the Storybook index accordingly.
- Updated the regression test to assert the rendered `.cm-editor` computed outline in jsdom for both `default` and `embedded` variants.
- Added a patch changeset for `@mastra/playground-ui`.

### DIFF
--- a/packages/playground-ui/src/ds/components/CodeEditor/code-editor.tsx
+++ b/packages/playground-ui/src/ds/components/CodeEditor/code-editor.tsx
@@ -228,12 +228,16 @@ const codeEditorVariants = cva('p-1 font-mono relative overflow-hidden', {
   variants: {
     variant: {
       default: 'rounded-md bg-surface3 border border-border1',
-      embedded: 'rounded-none border-none bg-transparent [&_.cm-editor.cm-focused]:outline-hidden',
+      embedded: 'rounded-none border-none bg-transparent',
     },
   },
   defaultVariants: {
     variant: 'default',
   },
+});
+
+const embeddedEditorAttributes = EditorView.editorAttributes.of({
+  style: 'outline: none',
 });
 
 export type CodeEditorProps = {
@@ -288,6 +292,10 @@ export const CodeEditor = forwardRef<ReactCodeMirrorRef, CodeEditorProps>(
         exts.push(EditorView.lineWrapping);
       }
 
+      if (variant === 'embedded') {
+        exts.push(embeddedEditorAttributes);
+      }
+
       if (language === 'json') {
         exts.push(jsonLanguage);
       } else if (language === 'markdown') {
@@ -307,7 +315,7 @@ export const CodeEditor = forwardRef<ReactCodeMirrorRef, CodeEditorProps>(
       }
 
       return exts;
-    }, [language, highlightVariables, schema, editable, lineWrapping]);
+    }, [language, highlightVariables, schema, editable, lineWrapping, variant]);
 
     return (
       <div className={cn(codeEditorVariants({ variant }), className)} {...props}>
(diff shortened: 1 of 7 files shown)


packages/playground-ui/src/ds/components/CodeEditor/code-editor.tsx
```diff
@@ -222,6 +224,52 @@ export const useCodemirrorTheme = (): Extension => {
   return useMemo(() => (isDark ? buildDarkTheme() : buildLightTheme()), [isDark]);
 };
 
+const codeEditorVariants = cva(
+  cn(
+    'font-mono relative overflow-hidden outline-hidden focus:outline-hidden focus-within:outline-hidden',
```
mfrachet, line 229: we should keep an outline somehow for accessiblity. focus-visible eventually

mfrachet (changes requested):

mfrachet (approved):

PR #19028, opened 2026-07-07
fix(core): skip guaranteed-miss storage read in createRun for transient runs

### LINKED ISSUE #19015: Streaming does one storage read per token: processor workflows call createRun() per chunk
This issue was created from Discord post 1523798407087394886:

[Open in Discord](https://discord.com/channels/1309558646228779139/1523798407087394886)

## Environment

- `@mastra/core`: **1.49.0** (bug also verified present in `1.50.0-alpha.2` dist)
- `@mastra/pg`: 1.13.2 (`PostgresStore`, schema `mastra`)
- `@mastra/ai-sdk`: 1.4.7, `ai`: 7.0.3
- Runtime: Node 24.x on Vercel Fluid Compute (region `iad1`), Next.js 16.2.10
- Storage: Supabase Postgres, region `ap-southeast-2` (Sydney), via pooler — **~210ms RTT from the function region** (this distance is what makes the bug visible; it reproduces at any latency)

## Summary

Agents combine their output processors (including the Memory-injected `MessageHistory`) into an
internal workflow named `${agentId}-output-processor` and execute it **once per streamed part**
via `ProcessorRunner.executeWorkflowAsProcessor`:

```js
async executeWorkflowAsProcessor(workflow, input, ...) {
  const run = await workflow.createRun();   // no runId → randomUUID()
  const result = await run.start({ ... });
```

`Workflow.createRun` then performs an **unconditional storage existence check** on that freshly
minted UUID:

```js
const shouldPersistSnapshot = this.#options.shouldPersistSnapshot({ ... }); // hard-coded false for processor workflows
const existingRun = await this.getWorkflowRunById(runIdToUse, { withNestedWorkflows: false });
```

The run ID was generated one line earlier, so the read is a guaranteed miss, every time. Net
effect: **one storage round trip per streamed chunk**, serialized into the stream pipeline.

Note the ordering: `shouldPersistSnapshot` is computed *first* (and is hard-coded `false` for
processor workflows, which the code itself describes as "transient and non-resumable" per the
#17344 fix), but it only gates the *write*. The read happens regardless.

## Expected behavior

- Streaming throughput is bounded by the model and network push, with no storage round trips on
  the per-chunk path. A direct `streamText` call from the same function/region streams the same
  gateway at ~50 tok/s; an agent with processors should be in the same range.
- Creating a run handle with a framework-generated UUID does not query storage: a just-minted
  UUID cannot already exist, and processor workflows can never have snapshot rows at all
  (`shouldPersistSnapshot` is hard-coded `false` for them).

## Actual behavior

- Every streamed part executes the combined processor workflow, whose `createRun()` performs a
  guaranteed-miss `getWorkflowRunById` against storage. Token delivery serializes behind one
  storage RTT per chunk.
- Measured: metronomic **212ms inter-chunk gap** (p50 212.0 / p95 213.8 / max 220 over 130
  chunks) = **~4.7 output tokens/sec**. A 316-token generation took 67.7s; multi-specialist
  agent turns took 11+ minutes.
- Storage-side confirmation: `pg_stat_statements` showed the `getWorkflowRunById` fingerprint at
  ~11 calls/sec during a turn with two active streams; a counting proxy over `stores.workflows`
  attributed 37 of 42 reads in a short turn to `workflowName: "<agentId>-output-processor"`,
  each with a never-before-seen `runId`.
- After short-circuiting only these reads (see workaround): inter-chunk gap dropped to
  **16.9ms mean (~59 tok/s), a 12.6× speedup**, same region layout, same code otherwise.
- With co-located storage (~1ms RTT) the waste still exists but hides inside model latency,
  which is presumably why it hasn't been noticed.

## Reproduction

Any agent with at least one output processor (or Memory enabled, which injects one) + a storage
adapter with non-trivial latency:

1. Wrap `storage.stores.workflows` in a proxy that counts `getWorkflowRunById` calls and records
   `args.workflowName`.
2. Stream any agent response.
3. Observe ~1 read per streamed part with `workflowName` = `${agentId}-output-processor`, each
   with a never-before-seen `runId`. Streaming rate degrades linearly with storage RTT.

serhiileniv:
## Description

Agents run their output processors as an internal `${agentId}-output-processor` workflow, and that workflow is executed once per streamed chunk. Every one of those runs called `Workflow.createRun`, which did an unconditional `getWorkflowRunById` storage read to check whether the run already exists. But the run id is freshly generated on that same call, and these processor workflows hard-code `shouldPersistSnapshot: () => false`, so they never write a snapshot row — the read can never return anything. On a high-latency storage adapter this turned into one storage round-trip per streamed token (the issue reports token delivery dropping to ~4.7 tok/s, ~212ms between chunks, on a ~210ms-RTT Postgres).

The fix skips that existence read when the run id was just generated *and* the workflow never persists a snapshot. Persisting workflows, and any call that passes an explicit `runId` (resume), still read exactly as before, so behavior is unchanged for them.

Before:

```ts
// createRun, on every streamed chunk of an agent with output processors / memory:
const existingRun = await this.getWorkflowRunById(runIdToUse, { withNestedWorkflows: false });
// guaranteed miss — fresh runId, and the workflow never persists — but still a storage round-trip
```

After:

```ts
const existingRun =
  shouldPersistSnapshot || options?.runId
    ? await this.getWorkflowRunById(runIdToUse, { withNestedWorkflows: false })
    : undefined;
```

The same guard is applied to the evented engine's `createRun` for consistency.

## Related issue(s)

Fixes #19015

## Type of change

- [ ] Documentation update
- [ ] Code refactoring
- [x] Performance improvement
- [ ] Test update

## Checklist

_Disclosure: I used AI assistance (Claude Code) to investigate and prepare this change._

## ELI5
This change stops the app from doing a pointless “go look for something that can’t be there” storage check for temporary streaming workflows. That makes streamed agent responses faster, while keeping the old behavior for runs that are meant to be saved or resumed later.

## Summary
- Skips the `getWorkflowRunById` storage read in `Workflow.createRun()` when a run is freshly created and the workflow does not persist snapshots.
- Preserves storage lookup behavior when snapshots are persisted or when an explicit `runId` is provided for resume behavior.
- Applies the same guard in the evented workflow engine’s `createRun()`.
- Updates tests to cover the non-persisting transient workflow path, the persisting path, and explicit `runId` behavior.
- Adds a changeset documenting the `@mastra/core` patch and the streaming/storage performance fix.

### DIFF
--- a/.changeset/fair-boxes-tan.md
+++ b/.changeset/fair-boxes-tan.md
@@ -1,0 +1,5 @@
+---
+'@mastra/core': patch
+---
+
+Fixed slow agent streaming with remote storage: output processors run an internal workflow once per streamed chunk, and each run performed a storage read that could never hit (the run id is freshly generated and these transient workflows never persist a snapshot). On high-latency databases this throttled token delivery to roughly one storage round-trip per token. The guaranteed-miss read is now skipped for freshly-created transient runs. (#19015)
--- a/packages/core/src/agent/__tests__/processor-workflow-storage-noise.test.ts
+++ b/packages/core/src/agent/__tests__/processor-workflow-storage-noise.test.ts
@@ -77,11 +77,13 @@ function buildAgentWithProcessor() {
 }
 
 describe('agent processor-workflow storage noise (issue #17137 follow-up to #17344)', () => {
-  it('gives the internal processor workflow access to storage (no "storage is not initialized" branch)', async () => {
-    // The buggy processor workflow used its own un-registered logger, so spying on the Mastra
-    // logger does not see the noise. Instead, assert the root cause directly: when
-    // getWorkflowRunById runs for the processor workflow it now has storage, so the
-    // no-storage debug branch is unreachable.
+  it('does not read storage (getWorkflowRunById) for the internal processor workflow on generate', async () => {
+    // #19015 short-circuits createRun's storage existence read for transient workflows
+    // (shouldPersistSnapshot: () => false) that mint a fresh runId. The internal
+    // processor workflow is exactly that, so its createRun no longer calls
+    // getWorkflowRunById at all. This strictly subsumes the original #17137/#17344
+    // no-noise goal: a lookup that never runs can never hit the "storage is not
+    // initialized" branch.
     const seen: Array<{ id: string; hasStorage: boolean }> = [];
     const original = (Workflow.prototype as unknown as { getWorkflowRunById: (...a: unknown[]) => unknown })
       .getWorkflowRunById;
@@ -100,8 +102,7 @@ describe('agent processor-workflow storage noise (issue #17137 follow-up to #173
     }
 
     const processorLookups = seen.filter(s => s.id === PROCESSOR_WORKFLOW_ID);
-    expect(processorLookups.length).toBeGreaterThan(0);
-    expect(processorLookups.every(s => s.hasStorage)).toBe(true);
+    expect(processorLookups).toEqual([]);
   });
 
   it('does not persist a snapshot for the internal processor workflow on generate', async () => {
--- a/packages/core/src/workflows/evented/workflow.ts
+++ b/packages/core/src/workflows/evented/workflow.ts
@@ -1663,9 +1663,13 @@ export class EventedWorkflow<
       stepResults: {},
     });
 
-    const existingRun = await this.getWorkflowRunById(runIdToUse, {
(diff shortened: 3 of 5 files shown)


abhiaiyer91 (approved):

serhiileniv:
Thanks @rohitsux! Glad it matched your sketch on the issue credit to you for laying out the approach. Appreciate the review ;)

PR #19033, opened 2026-07-07
fix(core): don't serialize live workflow in map initData reducer

### LINKED ISSUE #19018: Workflow `.map(mapVariable({ initData: <workflow> }))` deep-serializes the live workflow (incl. logger) into serializedStepGraph → heap OOM at module load
## Summary

`workflow.map({ field: mapVariable({ initData: <workflowInstance>, path }) })` (the object-form `.map` + the `initData` overload) causes Mastra to **JSON.stringify the entire live workflow instance — including its `logger` and full nested step graph — into that map step's `mapConfig` inside `serializedStepGraph`**, at `.commit()`/module-load time. For a deeply-nested / self-referential workflow this materializes a **~1 GB string** and the process dies with a **heap OOM before the server binds**.

This is related to (but not the same as) #11124 / #11175 — see [Relation to #11124](#relation-to-11124) below. The truncation added in #11175 does **not** prevent this OOM.

## Environment
- `@mastra/core@1.49.0` (confirmed). The same serialization logic is byte-identical in **1.50.0** (`dist/…/chunk-JGDMZZAO.js` ~L17024–17099), so the bug is expected to persist on latest.
- Node.js 24, production server (`mastra build` output), 2 GiB instance.

## Symptom

```
FATAL ERROR: Reached heap limit Allocation failed - JavaScript heap out of memory
```

- Happens during **workflow module evaluation / `.commit()`**, before the HTTP server binds a port. RSS peaks ~2.6 GB.
- A heap snapshot shows a single **~996 MB string** retained via `string ←[mapConfig] Object ←[step] Object ←[stepsArray]`, with content beginning:
  ```
  {"shipmentId":{"initData":{"component":"WORKFLOW","logger":{ …
  ```
  i.e. a serialized **workflow instance** (note `"component":"WORKFLOW"` and the inlined `logger`) sitting inside a map step's `mapConfig`.

## Root cause (traced in installed `@mastra/core@1.49.0` dist)

1. **`mapVariable` is an identity function** — it returns the config object verbatim, so the workflow instance is retained on `.initData`:
   ```js
   // dist/…/chunk-A77V6K33.js:16312
   function mapVariable(config) { return config; }
   ```
   The typed signature *requires* a workflow object for this overload — there is no id/boolean form:
   ```ts
   // dist/workflows/workflow.d.ts:35
   export declare function mapVariable<TWorkflow extends AnyWorkflow>({ initData, path }: { initData: TWorkflow; path: ... }): ...
   ```

2. **The object-form `.map()` reducer keeps the mapping object by reference** for an `initData` entry (falls through to the final `else`), so the built config still holds the live `Workflow` instance:
   ```js
   // dist/…/chunk-A77V6K33.js:17411
   const newMappingConfig = Object.entries(mappingConfig).reduce((a, [key, mapping]) => {
     const m = mapping;
     if (m.value !== void 0) { a[key] = m; }
     else if (m.fn !== void 0) { a[key] = { fn: m.fn.toString(), schema: m.schema }; }
     else if (m.requestContextPath) { a[key] = { requestContextPath: m.requestContextPath, schema: m.schema }; }
     else { a[key] = m; }        // <-- initData entry: retains the live workflow instance
     return a;
   }, {});
   ```

3. **It is then JSON.stringified into `mapConfig`** — and the length guard truncates only *after* the full string is materialized:
   ```js
   // dist/…/chunk-A77V6K33.js:17482
   this.serializedStepFlow.push({
     type: "step",
     step: {
       id: mappingStep.id,
       mapConfig: JSON.stringify(newMappingConfig, null, 2)?.length > 1e3
         ? JSON.stringify(newMappingConfig, null, 2).slice(0, 1e3) + "...\n}"
         : JSON.stringify(newMappingConfig, null, 2)
     }
   });
   ```
   `JSON.stringify` deep-walks the retained `Workflow` instance and everything reachable from it (nested steps, sub-workflows). With a self/nested reference this expands unboundedly and OOMs **before** `.slice(0, 1000)` can run.

4. **Why the `logger` gets embedded:** `Workflow extends MastraBase`, and `MastraBase` declares `component` and `logger` as **public, enumerable, own** instance fields:
   ```js
   // dist/…/chunk-77VL4DNS.js:4
   var MastraBase = class {
     component = RegisteredLogger.LLM;
     logger;
     constructor({ component, name }) {
       this.component = component || RegisteredLogger.LLM;
       this.logger = new ConsoleLogger({ name: `${this.component} - ${this.name}` });
     }
   ```
   So `JSON.stringify` serializes the live `ConsoleLogger` inline — matching the `"logger":{…` in the heap snapshot.

**Contrast — the function-form `.map()` is safe** because it stores the *source text* rather than walking an object graph:
```js
// dist/…/chunk-A77V6K33.js:17402
mapConfig: mappingConfig.toString()  // fn form — no object graph walked
```

**The runtime is unaffected by fixing this**: at execution the map only calls `getInitData()`; it never reads `m.initData` as an object:
```js
// dist/…/chunk-A77V6K33.js:17454
const stepResult = m.initData ? getInitData() : getStepResult2(...);
```

## Relation to #11124 / #11175

- **#11124** ("Snapshots grow exponentially when using mapVariable with sub-workflows", closed) is the same root cause but for the **`mapVariable({ step: <workflow> })`** overload, and reports a **runtime snapshot** growing to ~250 MB (Mastra Studio slowness), not a heap OOM at module load.
- **#11175** (merged) "fixed" it by adding the `JSON.stringify(...).length > 1e3 ? …slice(0,1000)…` truncation shown above. **This does not prevent the OOM**: the full ~1 GB string is built by `JSON.stringify` *before* the slice runs.

This report is specifically the **`initData` overload → heap OOM at `.commit()`/module load** path, which is not currently tracked and is not addressed by post-hoc truncation.

## Minimal repro (illustrative — not executed)

> Note: the ~1 GB balloon scales with graph depth / self-reference; a trivial two-step parent serializes small. The reproduction below shows the mechanism; the real-world OOM was observed with a deeply nested, self-referential workflow. The snippet was signature-checked against `@mastra/core@1.49.0` but not run.

```ts
import { z } from 'zod';
import { createWorkflow, createStep, mapVariable } from '@mastra/core/workflows';

const parentWorkflow = createWorkflow({
  id: 'parent-workflow',
  inputSchema: z.object({ shipmentId: z.string() }),
  outputSchema: z.object({ shipmentId: z.string() }),
}).then(createStep({
  id: 'run-child',
  inputSchema: z.object({ shipmentId: z.string() }),
  outputSchema: z.object({ shipmentId: z.string() }),
  execute: async ({ inputData }) => inputData,
}));

// initData points back at the parent instance itself (self-reference):
parentWorkflow.map({
  shipmentId: mapVariable({ initData: parentWorkflow, path: 'shipmentId' }),
});
parentWorkflow.commit();

// mapConfig now embeds the whole `parentWorkflow` instance (logger + graph):
console.log(JSON.stringify(parentWorkflow.serializedStepGraph).length); // balloons; OOMs when deeply nested
```

Run with a tight heap to force the boot form deterministically:
```
node --max-old-space-size=512 --import tsx repro.ts
# FATAL ERROR: Reached heap limit Allocation failed - JavaScript heap out of memory
```

## Proposed fix

Serialize a **stable reference (the workflow id)** for `initData` (and `step`-is-a-workflow) mappings instead of the live instance — not post-hoc truncation. E.g. in the object-form `.map()` reducer, add a branch before the fallthrough `else`:

```js
else if (m.initData !== void 0) { a[key] = { initData: m.initData.id, path: m.path }; }
```

Since the execute closure only uses `getInitData()` and never reads `m.initData` as an object, this is behavior-preserving. (Alternatively: pass a `replacer` to the `JSON.stringify` that drops `logger`/`component`/functions and collapses workflow instances to `{ id }`; or make `logger`/`component` non-enumerable on `MastraBase`.)

## Workaround

Use the **function form** of `.map()` and read the source via `getInitData()`:

```ts
.map(async ({ getInitData }) => {
  const init = MyInputSchema.parse(getInitData());
  return { shipmentId: init.shipmentId, /* … */ };
}, { id: '…' })
```

This serializes as `fn.toString()`, so `serializedStepGraph` stays tiny. It's value-identical to `mapVariable({ initData })` and resolved the OOM for us.

rohitsux:
Fixes #19018

## Problem

`workflow.map({ key: mapVariable({ initData: <workflowInstance>, path }) })` deep-serializes the entire live workflow instance — including its `logger` and full nested step graph — into that map step's `mapConfig` inside `serializedStepGraph`, at `.commit()` / module-load time.

`mapVariable` is an identity function, so an `initData` mapping carries a live `Workflow` reference. In the map reducer (`packages/core/src/workflows/workflow.ts`), that mapping falls through the generic `else` branch and is retained by reference, then `JSON.stringify`'d into `mapConfig`. The `.length > 1000` truncation only runs **after** the full string is materialized, so for a deeply-nested/self-referential workflow the walk builds a huge (reported ~1 GB) string and the process OOMs before the slice — before the server even binds.

## Fix

Add an `initData` branch to the reducer that serializes a slim reference instead of the instance:

```ts
} else if (m.initData !== undefined) {
  a[key] = { initData: m.initData?.id ?? true, path: m.path };
} else {
  a[key] = m;
}
```

This is behavior-preserving: the map step's execute closure only reads `m.initData` for truthiness (it calls `getInitData()`) and never treats it as an object, so serializing the id changes nothing at runtime.

`map()` is defined once on `Workflow` and inherited (not overridden) by `EventedWorkflow`, so a single change covers both the default and evented engines.

## Test

Adds `packages/core/src/workflows/map-initdata-serialization.test.ts`: builds a workflow with `.map({ x: mapVariable({ initData: someWorkflow, path }) }).commit()` and asserts the serialized map step's `mapConfig` holds an `{ initData: <id>, path }` reference and does **not** embed the live instance (`logger` / `component` / `stepFlow`), plus that it stays small. This reproduces the size regression deterministically — it fails before the fix (mapConfig is the truncated ~1000-char inlined blob) and passes after — without needing to trigger a real OOM.

## Verification

- New regression test fails on `main`, passes with the fix.
- `@mastra/core` map/serialization tests stay green (branch-map + default suites).
- `pnpm --filter @mastra/core build` OK; prettier clean.

## ELI5
Sometimes, mapping a workflow with `mapVariable({ initData: ... })` accidentally saved the entire live workflow inside the saved config, which could make the config enormous and crash from running out of memory. This PR makes that saved config store only a tiny reference to the workflow instead.

## Summary
- Fixed `Workflow.map()` serialization for `initData` mappings so `mapConfig` stores a slim reference (`initData: m.initData?.id ?? true` plus `path`) rather than deep-serializing the full live `Workflow` instance.
- Preserved runtime behavior by still treating `initData` as present for execution (using `getInitData()` when provided), while preventing workflow internals (logger/component/nested step graphs) from being embedded in `serializedStepGraph`.
- Added a regression test for the `mapVariable({ initData, path })` object-form mapping that verifies the serialized `mapConfig` is compact (and does not include live workflow internals).
- Added a changeset for a patch release of `@mastra/core` documenting the heap OOM fix.


taofeeq-deru (approved):

TheIsrael1:
## 🟠 Maintainer's Triage Note

**Current Phase:** Awaiting Approval
**Next Step:** Await final approval

**Triage:**

- Type: bug — Fixes workflow map `initData` serialization causing live workflow instances to be embedded in `serializedStepGraph`.
- Maintainer read: PR #19033 targets open issue #19018, which reports heap OOM from deep-serializing a live workflow instance during workflow commit/module load; the changed core workflow engine path is critical and was routed to workflow owners before normal Review.
- Route: Critical-path owner review
- Severity: 🟠 high — Issue #19018 is labeled `impact:high` and describes a workflow module-load OOM/blocking failure; the fix touches the main workflow engine implementation.

**Review:**

- Status: complete
- Findings: Critical-path owner review completed by @taofeeq-deru. GitHub reports the PR review decision as approved. Implementation was not independently reviewed in this triage flow.
- Confidence: 3/5
- Follow-up: Await final approval handling from @taofeeq-deru.

**Approve:**

- Status: waiting for final approval
- Final approver: @taofeeq-deru
- Notes: Critical-path owner approval is present; maintainers can use this note for final approval/merge routing once remaining repository checks and policy requirements are satisfied.

taofeeq-deru (approved):

PR #19035, opened 2026-07-07
fix(memory): honor per-call observationalMemory:false when OM is enabled on the instance

### LINKED ISSUE #18943: [BUG] Memory recall is branch-blind: no way to scope lastMessages history to a subset/active path of a forked thread (and per-call lastMessages:false is ignored when OM is enabled)
> **From the Zynda team.** We're a company building a production AI chat product on Mastra. This is an **important / near-blocking** issue for us: it prevents us from shipping branched conversations (parallel alternatives, edit-and-regenerate) correctly, and it silently affects context fidelity for any forked thread. We'd really appreciate guidance on a supported path or a fix.

### Summary

`Memory.recall()` builds the agent's history by **flat recency only** — `listMessages({ threadId, perPage: lastMessages, orderBy: { field: 'createdAt', direction: 'DESC' } })`. There is no `selectBy` / message-id allowlist / custom selector to restrict which messages are recalled.

This makes Mastra memory unusable for **branched/forked conversations**. An app can maintain a message tree (e.g. storing a `parentMessageId` in each message's `content.metadata`) and render the active branch correctly in its own UI — but when it calls `agent.stream(input, { memory: { thread, resource } })`, **every** sibling branch in the last-N window enters the model context. Continuing one branch leaks the other branch(es) into the prompt. There's no supported way to tell recall "only load this set of message ids / this linear path."

A secondary problem makes the obvious workaround fail (details below): when `observationalMemory` is enabled on the `Memory` instance, the per-call override `memory: { ..., options: { lastMessages: false, observationalMemory: false } }` is **accepted but has no effect** — full thread history is still recalled.

### Environment

- `@mastra/core@1.45.0`, `@mastra/memory@1.21.0`, `@mastra/pg@1.14.0`
- `ai@6.0.168`, `@ai-sdk/provider@3.0.8`
- Node 22, Postgres storage

### Reproduction

```ts
import type { MastraDBMessage } from '@mastra/core/agent';
import { MockLanguageModelV3 } from 'ai/test';
import { randomUUID } from 'node:crypto';

const storage = new PostgresStore({ connectionString: process.env.DATABASE_URL! });
const memory = new Memory({ storage, options: { lastMessages: 30 } });

// capture-the-prompt model
const model = new MockLanguageModelV3({
  doStream: async (opts) => {
    console.log('PROMPT MESSAGES:', JSON.stringify(opts.prompt));
    return {
      stream: new ReadableStream({
        start(c) {
          c.enqueue({ type: 'stream-start', warnings: [] });
          c.enqueue({ type: 'text-start', id: '0' });
          c.enqueue({ type: 'text-delta', id: '0', delta: 'ok' });
          c.enqueue({ type: 'text-end', id: '0' });
          c.enqueue({ type: 'finish', finishReason: 'stop',
            usage: { inputTokens: 1, outputTokens: 1, totalTokens: 2 } });
          c.close();
        },
      }),
    };
  },
});

const agent = new Agent({ id: 'a', name: 'a', instructions: 'x', model, memory });

const threadId = `t-${randomUUID()}`;
const resourceId = `r-${randomUUID()}`;

const row = (id: string, role: 'user'|'assistant', text: string,
             parentMessageId: string|null, createdAt: Date): MastraDBMessage => ({
  id, role, type: 'text', createdAt, threadId, resourceId,
  content: { format: 2, parts: [{ type: 'text', text }] as any,
    metadata: { parentMessageId } },
});

await memory.saveThread({ thread: {
  id: threadId, resourceId, title: 't',
  createdAt: new Date(), updatedAt: new Date(), metadata: {} } });

const t = Date.now();
await memory.saveMessages({ messages: [
  row('q',  'user',      'Q: which approach?',        null, new Date(t+1)),
  row('a1', 'assistant', 'ANSWER-ALPHA (branch A)',   'q',  new Date(t+2)), // branch A
  row('a2', 'assistant', 'ANSWER-BRAVO (branch B)',   'q',  new Date(t+3)), // branch B (sibling)
]});

// Continue branch A. There is no API to say "only recall the q -> a1 path".
await agent.streamUntilIdle(
  { role: 'user', content: 'continue' },
  { memory: { thread: threadId, resource: resourceId } },
);
// Logged PROMPT contains BOTH "ANSWER-ALPHA" and "ANSWER-BRAVO".
```

**Actual:** the prompt includes both `a1` and `a2` (both branches).
**Expected (desired):** a way to restrict recall to a caller-specified subset/linear path, so only `q -> a1` is loaded.

### Secondary issue — per-call recall override ignored under OM

With `observationalMemory` enabled on the instance, this call still recalls the full flat thread:

```ts
await agent.streamUntilIdle(activePathMessages, {
  memory: {
    thread: threadId, resource: resourceId,
    options: { lastMessages: false, observationalMemory: false }, // accepted, no effect
  },
});
```

The same `{ lastMessages: false }` override **does** suppress recall on an OM-**off** instance, so the override path appears to be bypassed specifically when OM is enabled (OM owns recall and isn't gated by the per-call options). Happy to split this into its own issue if preferred.

### Why it matters / impact

Any app implementing message branching, "edit & regenerate", or parallel alternatives on top of Mastra memory cannot keep the agent's context aligned with the branch the user is actually on. The only workaround we found is to **bypass Mastra recall entirely** — call `agent.stream(explicitMessages, /* no memory */)` and persist turns manually via `memory.saveMessages(...)` — which forfeits OM, working memory, and built-in persistence.

### Possible directions

1. A `selectBy` / message-id allowlist / predicate on the per-call `memory.options` that reaches `listMessages` (e.g. `memory: { thread, resource, options: { selectBy: { include: string[] } } }`).
2. A `recallFilter` / selector hook so apps can supply the active-path id set (or walk their own `parentMessageId`).
3. At minimum: honor per-call `options.lastMessages: false` / `options.observationalMemory: false` even when OM is enabled, and document it — so the "explicit messages + no recall" pattern is officially supported.

Thanks for building Mastra — glad to provide more detail, logs, or a PR if you can point us at the right seam.

rohitsux:
Fixes #18943.

> **Scope note (please read before merge):** #18943 has two parts. This PR fixes only part (2), the per-call disable bug. Part (1), the branch-aware recall selector, is a larger feature and is **intentionally out of scope** here. If maintainers want that feature tracked separately, it should move to a fresh follow-up issue before this merges, since `Fixes #18943` will auto-close the original on merge. Happy to open that follow-up issue on request.

## The two parts of #18943

1. A branch-aware recall selector (a larger feature) — **not addressed in this PR.**
2. A bug where a per-call `observationalMemory: false` is silently ignored when Observational Memory is enabled on the Memory instance — **this is what this PR fixes.**

## Problem

`Memory.getInputProcessors` injects the Observational Memory input processor via `createOMProcessor`. It decided whether a per-call runtime override applied by branching on the *normalized* override:

```ts
const runtimeObservationalMemory = normalizeObservationalMemoryConfig(runtimeMemory?.memoryConfig?.observationalMemory);
const threadConfig = runtimeObservationalMemory ? this.getMergedThreadConfig({ ... }) : this.threadConfig;
```

But `normalizeObservationalMemoryConfig(false)` returns `undefined`, so an explicit per-call `observationalMemory: false` was indistinguishable from "no override provided" — the code fell back to `this.threadConfig` (instance config, OM enabled) and the OM processor was still added. The per-call disable was accepted by the API and silently dropped.

## Fix

Apply the per-call config whenever it carries an `observationalMemory` key (keyed on presence, not truthiness):

```ts
const hasRuntimeOMOverride = !!runtimeMemoryConfig && 'observationalMemory' in runtimeMemoryConfig;
const threadConfig = hasRuntimeOMOverride ? this.getMergedThreadConfig(runtimeMemoryConfig) : this.threadConfig;
```

`getMergedThreadConfig` deep-merges the per-call `false` over the enabled instance value; the merged `observationalMemory: false` then normalizes to `undefined` and `createOMProcessor` returns `null`, so the processor is not added. Callers that pass no override are unchanged.

`createOMProcessor` is the single OM injection point used by both the default engine (`prepare-stream` -> `getInputProcessors`) and the durable engine, so one change covers both.

## Test

Adds a regression describe block in `packages/memory/src/index.test.ts`: with OM enabled on the instance, `getInputProcessors` includes the `observational-memory` processor by default, and a per-call `RequestContext` carrying `{ memoryConfig: { observationalMemory: false } }` now correctly omits it. Fails on `main`, passes with the fix.

## Verification

- New regression test fails on `main`, passes with the fix.
- `@mastra/memory` suite stays green (index + clone-thread-om, 107 tests).
- `pnpm --filter @mastra/memory build` OK; prettier clean.

## ELI5
This change fixes a case where turning a feature off just for one request didn’t actually turn it off. Now, if a request says `observationalMemory: false`, it really stays off even when it’s enabled on the Memory instance.

## Summary
- Fixed per-call `observationalMemory: false` overrides in `@mastra/memory`
- Updated merge logic to detect the `observationalMemory` setting by key presence, not truthiness
- Kept instance-level behavior unchanged when no per-call override is provided
  - default behavior when Observational Memory is enabled on the instance
  - per-request disable behavior via `RequestContext`
- Added a changeset for a patch release

### DIFF
--- a/.changeset/fix-om-per-call-disable.md
+++ b/.changeset/fix-om-per-call-disable.md
@@ -1,0 +1,5 @@
+---
+'@mastra/memory': patch
+---
+
+Fixed disabling Observational Memory for a single call. Previously, passing `observationalMemory: false` for one request had no effect when Observational Memory was already enabled on the Memory instance — it stayed active. This per-request override is now respected correctly.
--- a/packages/memory/src/index.test.ts
+++ b/packages/memory/src/index.test.ts
@@ -2768,4 +2768,31 @@ describe('Memory', () => {
       expect(childSpan.error).not.toHaveBeenCalled();
     });
   });
+
+  describe('per-call observationalMemory override (#18943)', () => {
+    const withOMEnabled = () =>
+      new Memory({
+        storage: new InMemoryStore(),
+        options: { observationalMemory: { enabled: true } },
+      });
+
+    it('adds the observational-memory input processor when OM is enabled on the instance', async () => {
+      const memory = withOMEnabled();
+      const processors = await memory.getInputProcessors([], new RequestContext());
+      expect(processors.find(p => p.id === 'observational-memory')).toBeDefined();
+    });
+
+    it('honors a per-call observationalMemory:false and does not add the OM processor', async () => {
+      const memory = withOMEnabled();
+      const requestContext = new RequestContext();
+      // Per-call disable, even though OM is enabled on the instance. Before the
+      // fix this was silently ignored (the override normalizes to undefined, so
+      // the code fell back to the instance config) and the processor was still
+      // added.
+      requestContext.set('MastraMemory', { memoryConfig: { observationalMemory: false } });
+
+      const processors = await memory.getInputProcessors([], requestContext);
+      expect(processors.find(p => p.id === 'observational-memory')).toBeUndefined();
+    });
+  });
 });
--- a/packages/memory/src/index.ts
+++ b/packages/memory/src/index.ts
@@ -2999,14 +2999,16 @@ Notes:
     if (hasObservationalMemory) return null;
 
     const runtimeMemory = context?.get('MastraMemory') as { memoryConfig?: RuntimeMemoryConfig } | undefined;
-    const runtimeObservationalMemory = normalizeObservationalMemoryConfig(
-      runtimeMemory?.memoryConfig?.observationalMemory,
-    );
-    const threadConfig = runtimeObservationalMemory
-      ? this.getMergedThreadConfig({
-          ...runtimeMemory?.memoryConfig,
-          observationalMemory: runtimeObservationalMemory,
-        } as MemoryConfigInternal)
+    // Merge in the per-call config whenever it explicitly carries an
+    // `observationalMemory` override, keyed on presence rather than truthiness.
+    // A per-call `observationalMemory: false` normalizes to `undefined`, so
+    // branching on the normalized value would drop the override and fall back
+    // to the instance config — silently ignoring an explicit per-call disable
+    // when OM is enabled on the instance (#18943).
+    const runtimeMemoryConfig = runtimeMemory?.memoryConfig;
+    const hasRuntimeOMOverride = !!runtimeMemoryConfig && 'observationalMemory' in runtimeMemoryConfig;
+    const threadConfig = hasRuntimeOMOverride
+      ? this.getMergedThreadConfig(runtimeMemoryConfig as MemoryConfigInternal)
       : this.threadConfig;
 
     const effectiveConfig = normalizeObservationalMemoryConfig(threadConfig.observationalMemory);


rohitsux:
Thanks @coderabbitai. Addressed both:

1. Changeset wording: rewritten in plain, outcome-focused language, dropping the internal symbol names per the changeset house-style. Applied your suggested phrasing.

2. Enable-per-call test: I tried adding it and it surfaced that a bare per-call `observationalMemory: true` deep-merges into a disabled instance's `{ enabled: false }` and does not flip OM on, so the processor is (correctly) not added in that case. That means the earlier "enabling per-call still works" line in the description was inaccurate for the bare-boolean case, so I have removed that claim rather than assert behavior the code does not guarantee. This PR is scoped to the disable path, which is fully covered by the two regression tests. Happy to open a separate issue if reliable per-call enable is desired.

abhiaiyer91:
Thanks for the contribution! The linked issue (#18943) was closed as not planned, so there's no longer an open problem for this PR to address. Closing it for now. If you believe the issue should be revisited, please comment on #18943 with your reasoning and we can reopen both.

PR #19059, opened 2026-07-07
feat(mastracode): storage retention policies and /prune command

abhiaiyer91:
Mastra Code's local libsql db grows forever — mine hit 30 GB (plus a 25 GB WAL), mostly observability spans and old chat messages. This adds default retention policies (spans/logs 14d, scores/snapshots 30d, messages/threads 90d) and a `/prune` command to apply them. Nothing is deleted automatically — rows only go away when you run the command.

```
/prune          # delete rows older than the retention policies
/prune vacuum   # prune, then compact the db files to reclaim disk
```

`/prune` closes the TUI and hands the terminal to a maintenance run with live progress, so deletes never contend with a live session (running it inside the TUI froze the event loop and hit `SQLITE_BUSY`). Vacuum uses `VACUUM INTO` + an atomic swap instead of in-place `VACUUM`, which kept memory bounded on the 30 GB db (26s, 69 MB peak RSS, 30 GB → 10 GB) where in-place VACUUM froze the whole machine. Before swapping, it proves exclusive access and refuses with a clear message if another session has the file open, since a surviving connection would keep writing to the swapped-out inode. The probe reads the SQLite file header because closed `@libsql/client` connections keep their file lock until GC finalizes their statements (upstream libsql-js bug, documented in code).

Along the way: `MastraCompositeStore` domain overrides now accept `false` to disable a domain entirely (mastracode uses this so observability data stops falling through to libsql when local tracing is off), and `LibSQLVector` gains a `close()` mirroring `LibSQLStore.close()` so the vector db can be released before compaction. Covered by unit tests, a TUI e2e scenario, and verified end-to-end against the real 30 GB database (`quick_check` ok, retention boundaries exact).

## ELI5
This update gives Mastra Code a way to clean up old local data safely. You can now run `/prune` to delete expired storage rows on demand, and `/prune vacuum` to also shrink the database file afterward without risking an active session.

## Summary
- Added default retention rules for local storage data:
  - 14 days for spans/logs
  - 30 days for scores/workflow snapshots
  - 90 days for messages/threads
- Added a `/prune` command that exits the TUI and performs maintenance in the terminal with live progress.
- Added `/prune vacuum` to compact libsql databases using `VACUUM INTO` plus an atomic swap, with safety checks for disk space and exclusive access.
- Added `/prune keep-memory` to keep messages/threads while still pruning telemetry and run records.
- Allowed `false` overrides in `MastraCompositeStore` so individual domains can be disabled explicitly.
- Added per-call `prune({ retention })` overrides without changing the store’s configured retention.
- Added `LibSQLVector.close()` so local libsql files can be cleanly released before compaction.
- Updated docs and added tests for pruning, retention overrides, disabled domains, and vector close behavior.


docs/src/content/en/reference/storage/composite.mdx
```diff
@@ -96,6 +96,32 @@ export const mastra = new Mastra({
...
+  connectionString: process.env.DATABASE_URL,
+})
+
+export const mastra = new Mastra({
+  storage: new MastraCompositeStore({
+    id: 'composite',
+    default: pgStore,
+    domains: {
+      // don't persist traces and spans
+      observability: false,
```
DanielSLew, line 152: why do you have to set this to false, shouldn't it not persist traces and spans if you don't add observability to Mastra?

DanielSLew (commented):
added a question, but also had a couple other q's

- Is there a way to set a retention period that automatically prunes rather than having to remember to do it manually?
- This feels like it should live in the storage layer itself rather than in MC, setting a retention policy in the storage layer for a given domain could be very useful.

abhiaiyer91:
Retention is in core, executing prune is also in core. Doing it automatically is dangerous so users can set it up themselves 

No setting a domain defaults it to the base which LibSQL has ai spans. So false explicitly says we don’t want this domain

DanielSLew (approved):

DanielSLew:
Did a deep pass over this (already approved) — leaving the detailed notes here.

## What's strong

- The vacuum path is unusually careful: `VACUUM INTO` + swap keeps memory bounded, the free-space preflight (`live_bytes × 1.2 + 256 MB`) prevents mid-vacuum disk exhaustion, and the raw file-header probe (bytes 18/19) to prove exclusive access — instead of a pragma, because closed `@libsql/client` connections pin file locks until GC finalizes statements — is the right workaround for the upstream bug, and it's documented in-code.
- Swap ordering (`file → .old`, delete `-wal`/`-shm` sidecars, `tmp → file`) is correct; a plain atomic rename would leave stale WAL sidecars that SQLite would try to recover into the new db.
- Pruning through the *inner* store so legacy libsql observability spans are covered even with the composite domain disabled is a nice touch.
- Test coverage is genuinely good: command ordering ("TUI stops before first prune pass"), vector `close()` WAL assertions + idempotency, composite `false`-override tests, and a real terminal hand-off e2e.

## Suggested follow-ups (pre-merge for 1–2, follow-up issues fine for the rest)

1. **`threadState` can't actually be disabled via `false`** (coderabbit is right): `base.ts` `resolve('threadState') ?? new InMemoryThreadStateStorage()` silently re-enables the domain, contradicting the new JSDoc ("nothing can read from or write to it"). Since this is public core API, either exclude `threadState` from the `| false` type or document the exception.
2. **Quiesce failures are swallowed** in `prune.ts` — `Promise.allSettled` results aren't inspected, so a failed `stopWorkers()` means maintenance runs against a live writer, which is the exact contention this design exists to prevent. At minimum log rejections; arguably abort.
3. **Swap has no rollback on partial failure**: if the process dies between `rename(file, .old)` and `rename(tmp, file)`, the main db path is empty (data recoverable from `.old`/tmp, but a naive restart creates a fresh db). A try/catch restoring `.old`, or a startup check for orphaned `.old` files, would close this.
4. **TOCTOU window**: the exclusivity probe runs before `VACUUM INTO` (~30s on big DBs); a session started in that window gets its inode orphaned by the swap. Re-probing immediately before the swap would shrink it substantially.
5. **DuckDB observability has no retention story** — when local tracing is on, the growth problem will eventually reappear in that file. Follow-up issue is fine.
6. Optional: a one-line note near the deps for why both `@libsql/client` *and* native `libsql` are needed, so nobody "cleans up" the seemingly redundant one later.

## CI

The 3 failing jobs all look unrelated: 4× server-adapter `POST /schedules` failures + 6× core `request-context-mutation.scenario` failures (no adapter/loop code touched here — core tests only run because `base.ts` changed), and kitchen-sink 2/4 is a Playwright `locator.click` timeout (1/4, 3/4, 4/4 pass). Everything actually exercising this PR is green (Mastra Code E2E, Combined Store Tests, Memory Tests, lint, build). Worth confirming those repro on `main` and rerunning.

mastracode/sdk/src/utils/storage-maintenance.ts
```diff
@@ -0,0 +1,371 @@
...
+ * everything else (spans have been observed at 17+ GB):
+ * - spans/logs: short-lived debugging telemetry → 14d
+ * - scorer results & workflow snapshots: medium-lived run records → 30d
+ * - chat history (messages/threads): long-lived, user-visible → 90d
+ */
+export const DEFAULT_RETENTION: RetentionConfig = {
+  memory: {
+    messages: { maxAge: '90d' },
+    threads: { maxAge: '90d' },
+  },
```
TylerBarnes, line 42: it'd be nice to have an option to keep messages/threads because they're super valuable if you want to fine tune on them later (or create evals, or do learning over the data). But I'd love to delete the other stuff

axe312ger: **huge +1**

docs/src/content/en/reference/storage/composite.mdx
abhiaiyer91: Not adding observability to Mastra only disables tracing — the storage domain is separate. `MastraCompositeStore` resolves every domain from the `default` store, and the LibSQL default includes an observability domain (with the `mastra_ai_spans` table), so span writes still land in LibSQL even with tracing off. `observability: false` explicitly opts the domain out of that fallback. Clarified this in the composite storage docs in 143135d.

mastracode/sdk/src/utils/storage-maintenance.ts
abhiaiyer91: Shipped in 143135d: `/prune keep-memory` skips the memory domain (messages/threads/resources) while still pruning telemetry, scores, and run records. Composes with vacuum in any order, e.g. `/prune vacuum keep-memory`. Under the hood `prune()` now accepts a per-call `retention` option so callers can override the standing policies.

mastracode/tui/src/tui/commands/prune.ts
```diff
@@ -0,0 +1,57 @@
...
+  const log = (line: string) => console.info(line);
+  log('Closing the TUI to run storage maintenance…');
+  let exitCode = 0;
+  try {
+    // Quiesce background writers so maintenance has the db to itself.
+    await Promise.allSettled([
+      ctx.mcpManager?.disconnect(),
+      ctx.controller.getMastra()?.stopWorkers(),
+      ctx.controller.stopIntervals(),
+    ]);
```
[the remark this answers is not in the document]
abhiaiyer91: Fixed in 143135d — the `Promise.allSettled` results are now inspected and each quiesce failure is logged with context before maintenance proceeds. Covered by a unit test.

mastracode/sdk/src/utils/storage-maintenance.ts
```diff
@@ -0,0 +1,371 @@
...
+      db.close();
+    }
+    // Swap the compacted copy into place. The old WAL/SHM sidecars belong to
+    // the old inode — they must never be paired with the new file.
+    renameSync(file, `${file}.old`);
+    rmSync(`${file}-wal`, { force: true });
+    rmSync(`${file}-shm`, { force: true });
+    renameSync(tmp, file);
+    rmSync(`${file}.old`, { force: true });
+    results.push({ file, bytesBefore, bytesAfter: fileSizeWithWal(file) });
```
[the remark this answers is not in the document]
abhiaiyer91: Fixed in 143135d — the swap is now wrapped in try/catch and restores the original file from `.old` on partial failure, and exclusivity is re-probed immediately before the swap to shrink the TOCTOU window. Both paths have unit tests (targeted `renameSync` failure + lock acquired between probe and swap).

packages/core/src/storage/base.ts
```diff
@@ -347,9 +354,13 @@ export class MastraCompositeStore extends MastraBase {
...
+      // A `false` override disables the domain — it resolves to undefined
+      // instead of falling through to the editor/default stores.
       const resolve = <K extends keyof StorageDomains>(key: K): StorageDomains[K] | undefined => {
-        if (domainOverrides[key] !== undefined) return domainOverrides[key];
+        const override: StorageDomains[K] | false | undefined = domainOverrides[key];
+        if (override === false) return undefined;
+        if (override !== undefined) return override;
         if (editorDomainSet.has(key) && editorStores?.[key] !== undefined) return editorStores[key];
         return defaultStores?.[key];
       };
```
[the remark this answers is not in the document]
abhiaiyer91: Fixed in 143135d — the fallback is now `resolve("threadState") ?? false`, so a `false` override disables the domain instead of silently re-creating the in-memory store. Test added.

abhiaiyer91:
Thanks for the deep pass @DanielSLew — addressed in 143135d:

- **TOCTOU window**: exclusivity is now re-probed immediately before the swap (with rollback if a lock appears mid-vacuum)
- **Dep note**: comment added explaining why both `@libsql/client` and native `libsql` are needed, with a link to the upstream bug we filed (tursodatabase/libsql-js#228) so the native dep can be dropped once fixed
- **DuckDB observability retention**: agreed, tracking as a follow-up issue

Also shipped `/prune keep-memory` (Tyler's request) to preserve messages/threads while pruning telemetry. CI failures you flagged don't touch this PR's paths — will confirm they repro on main and rerun.

PR #19066, opened 2026-07-07
fix(deployer): false-positive LOCAL_STORAGE_PATH preflight error from bundler intermediates

rphansen91:
The deploy preflight check could flag local storage paths like `file:./data.db` even when they don't exist anywhere in the user's project, blocking `mastra server deploy` and forcing `--skip-preflight`.

The local-storage detector excluded only `@mastra__*` shim files under `.mastra/.build/`, but the bundler writes other intermediate modules there too — including generic `chunk-*.mjs` files that preserve JSDoc examples from `@mastra/core`:

```ts
// inside .mastra/.build/chunk-JGDMZZAO.mjs (bundled @mastra/core JSDoc)
 * const mastra = new Mastra({
 *   storage: new LibSQLStore({ id: 'mastra-storage', url: 'file:./data.db' })
 * });
```

The detector now excludes everything under `.mastra/.build/`, since nothing in that directory is user source. Local storage paths in your own files are still detected.

Verified end-to-end against a real affected project: reproduced the false positive with the published deployer, confirmed the flagged module was a generated chunk, and confirmed the fix clears it while still flagging a real `file:./data.db` added to user source. Unit tests extended to cover non-`@mastra__` build intermediates (11 passing).

## ELI5
This change stops the deploy checker from blaming your app for paths that only exist in temporary build files. It now ignores the whole `.mastra/.build/` folder, so only real local-storage paths in your source code still cause warnings.

## Summary
- Fixed a false-positive `LOCAL_STORAGE_PATH` preflight error during `mastra server deploy`.
- Updated the detector to ignore all generated files under `.mastra/.build/`, not just `@mastra__*` shims.
- Kept detection active for user-authored source files that באמת contain local-storage paths.
- Expanded tests to cover additional generated build intermediates, including non-`@mastra__` shim files.
- Verified the issue against a real affected project.

### DIFF
--- a/.changeset/modern-items-study.md
+++ b/.changeset/modern-items-study.md
@@ -1,0 +1,5 @@
+---
+'@mastra/deployer': patch
+---
+
+Fixed a false-positive LOCAL_STORAGE_PATH preflight error that flagged storage paths like `file:./data.db` that don't exist in your project. The deploy bundler's local-storage detector now excludes everything under `.mastra/.build/` (deployer-generated intermediate chunks), not just `@mastra__*` shim files. Those chunks can carry JSDoc examples from library code (for example `LibSQLStore({ url: 'file:./data.db' })` from `@mastra/core`), which previously blocked `mastra server deploy` and forced `--skip-preflight` even though the user's code had no local storage paths. Local storage paths in your own source files are still detected.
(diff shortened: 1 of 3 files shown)


rphansen91:
Folding this into #19071, which includes this fix plus the broader preflight metadata rework (env-guarded fallbacks + user env ref scoping).

PR #19078, opened 2026-07-07
test(core): add trailing assistant guard coverage

### LINKED ISSUE #18951: Add unit tests for trailing-assistant-guard processor
The `TrailingAssistantGuard` processor and `isMaybeClaude46` helper in `packages/core/src/processors/trailing-assistant-guard.ts` guard against Claude 4.6 trailing-assistant rejections but are untested.

Scope: `isMaybeClaude46` across string configs, language model objects, function configs, fallback arrays, unknown shapes; `TrailingAssistantGuard` gating (missing schema, jsonPromptInjection, model, non-assistant/empty messages) and user-message append behavior.

godququ5-code:
## Summary

- add focused coverage for Claude 4.6 detection across string, language-model object, dynamic, unknown, and fallback configurations
- verify that `TrailingAssistantGuard` only appends its continuation message for native structured output after an assistant message
- assert the exact appended message shape and preserve the original message sequence using real `MastraDBMessage` fixtures and a Zod schema

Fixes #18951.

## Why

`trailing-assistant-guard.ts` currently has no dedicated unit tests. These cases protect both the conservative model-detection behavior and the structured-output gating conditions that determine whether the guard mutates the input message list.

## Validation

- `../../node_modules/.bin/vitest --config /dev/null run src/processors/trailing-assistant-guard.test.ts` (14 tests passed)
- `../../node_modules/.bin/eslint src/processors/trailing-assistant-guard.test.ts`
- `../../node_modules/.bin/prettier --check src/processors/trailing-assistant-guard.test.ts`
- `git diff --check`

## Notes

The package-wide `tsc --noEmit` check could not complete in the filtered local install because several internal workspace packages were not built/resolvable (for example `@internal/ai-sdk-v4` and `@mastra/schema-compat`). The focused Vitest run type-checked and executed this test file successfully.

## ELI5
This adds tests for a “helper” that decides when to add one more message at the end of a conversation. The tests make sure it only does that in the right cases, so the app behaves correctly with different model setups.

### Summary
- Added Vitest coverage for `isMaybeClaude46` across:
  - string model configs
  - language model objects
  - function/dynamic configs
  - fallback arrays
  - unknown shapes and non-Claude 4.6 cases
- Added `TrailingAssistantGuard` tests for:
  - public `id` and `name`
  - appending a continuation user message after an assistant message when native structured output is enabled
  - preserving the original message sequence
  - skipping the append when structured output is missing, `jsonPromptInjection` is enabled, the model doesn’t match, the last message is not from the assistant, or there are no messages
- Uses real `MastraDBMessage` fixtures and a Zod schema to verify the exact appended message shape

### DIFF
--- a/packages/core/src/processors/trailing-assistant-guard.test.ts
+++ b/packages/core/src/processors/trailing-assistant-guard.test.ts
@@ -1,0 +1,148 @@
+import { describe, expect, it } from 'vitest';
+import { z } from 'zod/v4';
+
+import type { MastraDBMessage } from '../agent/message-list';
+import { isMaybeClaude46, TrailingAssistantGuard } from './trailing-assistant-guard';
+import type { ProcessInputStepArgs } from './index';
+
+const createMessage = (role: 'user' | 'assistant', text: string): MastraDBMessage => ({
+  id: `${role}-${text}`,
+  role,
+  content: {
+    format: 2,
+    parts: [{ type: 'text', text }],
+  },
+  createdAt: new Date(),
+  threadId: 'test-thread',
+});
+
+const makeArgs = (
+  overrides: Pick<Partial<ProcessInputStepArgs>, 'messages' | 'structuredOutput'> = {},
+): ProcessInputStepArgs =>
+  ({
+    messages: overrides.messages ?? [createMessage('assistant', 'draft response')],
+    structuredOutput:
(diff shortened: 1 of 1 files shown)


wardpeet (approved):
Thank you, this one was indeed missing and a good candiate

PR #19084, opened 2026-07-08
fix(memory): mark data-om-status part transient so it is not persisted

### LINKED ISSUE #18869: [BUG] data-om-status part missing transient flag — pollutes persisted message history with standalone status-only messages

minh2416294:
## What

`data-om-status` Observational Memory parts were being persisted into message history. This adds the `transient: true` flag to the status emit so the `OutputWriter` stops writing each snapshot as a standalone assistant message.

Fixes #18869.

## Why

`data-om-status` is broken because the emit omits `transient`. Every OM lifecycle marker is streamed with `transient: true` so the `OutputWriter` doesn't persist it as a standalone data-only message, except the status snapshot. In `observational-memory.ts`, the three sibling markers all set it:

```ts
void writer.custom({ ...startMarker, transient: true }).catch(() => {});
```
but the status emit did not:
```ts
await writer.custom(statusPart).catch(() => {});   // ← no transient flag
```

The core persistence gate writes any non-transient data-* chunk as an assistant message, so each status snapshot became a durable row. data-om-status is a window-usage snapshot for polling UIs. It has no persistMarkerToStorage call and is explicitly non-rendered in channels, so it was never meant to persist. On long-running threads this crowds real conversation turns out of paginated listMessages reads.

## How
One-line change matching the sibling markers:

```ts
- await writer.custom(statusPart).catch(() => {});
+ await writer.custom({ ...statusPart, transient: true }).catch(() => {});
```

## Testing
Added a regression test: `data-om-status transient flag (regression for #18869)` in `observational-memory.test.ts`. It calls `emitProgress` with a capturing writer and asserts the emitted `data-om-status` part carries `transient: true`.

- Fails on main (without the fix `transient` is `undefined`), passes with this change.
- Full OM suite green: 453 passed.

```
node_modules/.bin/vitest run packages/memory/src/processors/observational-memory/__tests__/observational-memory.test.ts -t "transient flag"
```

## Impact / regression risk
Scoped to the OM status emit only. The snapshot is still streamed to clients (polling UIs unaffected). It's just no longer written to storage, matching every other OM marker. No change for any other data part.

## Deviations
I commented on the issue first before opening this. Prior PR [#18922](https://github.com/mastra-ai/mastra/issues/18922) shipped the same one-line fix but was self-closed by its author without a regression test. This PR adds that missing test so the fix lands with coverage. I'd happy to defer if @bkidy or @aryanbarde80 would rather take it

## ELI5
This PR stops Observational Memory “status update” snapshots from being saved into the chat’s stored message history. They still stream to the UI as they happen, but they won’t show up later as fake assistant messages.

## What changed
- Updated `emitProgress` so `data-om-status` snapshots are emitted with `transient: true`.
- Added a regression test asserting that the emitted `data-om-status` part includes `transient: true` (fix for `#18869`), ensuring `OutputWriter` won’t persist these snapshots.

## Why
- `data-om-status` is ephemeral UI/polling state (a window-usage snapshot), not durable conversation content.
- Without `transient: true`, these snapshots polluted persisted message history and displaced real conversation turns in paginated reads.

### DIFF
--- a/.changeset/om-status-transient.md
+++ b/.changeset/om-status-transient.md
@@ -1,0 +1,5 @@
+---
+'@mastra/memory': patch
+---
+
+Fixed Observational Memory writing `data-om-status` snapshots into persisted message history. The status part (a window-usage snapshot meant only for polling UIs) was emitted without the `transient` flag that every other OM lifecycle marker sets, so the stream writer persisted each snapshot as a standalone assistant message. On long-running threads this crowded real conversation turns out of paginated history reads. The status part is now marked transient and is no longer persisted. Fixes #18869.
--- a/packages/memory/src/processors/observational-memory/__tests__/observational-memory.test.ts
+++ b/packages/memory/src/processors/observational-memory/__tests__/observational-memory.test.ts
@@ -7377,6 +7377,66 @@ describe('Locking Behavior', () => {
   });
 });
 
+describe('data-om-status transient flag (regression for #18869)', () => {
+  const makeCapturingWriter = () => {
+    const customCalls: any[] = [];
+    const writer = {
+      custom: async (part: any) => {
+        customCalls.push(part);
+      },
+      write: async () => {},
+      close: async () => {},
+    } as any;
+    return { writer, customCalls };
+  };
+
+  it('emits data-om-status with transient: true so the OutputWriter does not persist it', async () => {
+    const storage = createInMemoryStorage();
+    const threadId = 'om-status-transient-thread';
+    const resourceId = 'om-status-transient-resource';
+
+    const om = new ObservationalMemory({
+      storage,
+      scope: 'thread',
+      observation: {
+        messageTokens: 100000,
+        model: createStreamCapableMockModel({ defaultObjectGenerationMode: 'json' }) as any,
+      },
+      reflection: {
+        observationTokens: 40000,
+        model: createStreamCapableMockModel({ defaultObjectGenerationMode: 'json' }) as any,
+      },
+    });
+
+    const record = await storage.initializeObservationalMemory({
+      threadId,
+      resourceId,
+      scope: 'thread',
+      config: {},
+    });
+
+    const { writer, customCalls } = makeCapturingWriter();
+
+    await om.emitProgress({
+      record,
+      pendingTokens: 10,
+      threshold: 100,
+      effectiveObservationTokensThreshold: 100,
+      currentObservationTokens: 0,
+      writer,
+      stepNumber: 0,
+      threadId,
+      resourceId,
+    });
+
+    const statusParts = customCalls.filter(part => part?.type === 'data-om-status');
+    expect(statusParts).toHaveLength(1);
+    // Without transient: true the OutputWriter persists each status snapshot as a
+    // standalone assistant message (#18869). Every sibling OM marker sets this flag.
+    expect(statusParts[0].transient).toBe(true);
+  });
+});
+
 describe('Reflection with Thread Attribution', () => {
   it('should create a new record after reflection', async () => {
     const storage = createInMemoryStorage();
--- a/packages/memory/src/processors/observational-memory/observational-memory.ts
+++ b/packages/memory/src/processors/observational-memory/observational-memory.ts
@@ -2664,7 +2664,8 @@ ${formattedMessages}
       omDebug(
         `[OM:status] step=${stepNumber} msgs=${pendingTokens}/${threshold} obs=${currentObservationTokens}/${effectiveObservationTokensThreshold} gen=${record.generationCount}`,
       );
-      await writer.custom(statusPart).catch(() => {});
+      // Stream OM lifecycle markers as transient so the OutputWriter does not persist standalone data-only messages; the status snapshot is ephemeral (polling UIs only) and has no durable intent.
+      await writer.custom({ ...statusPart, transient: true }).catch(() => {});
     }
   }
 


bkidy:
No worries, please go ahead. Thanks for adding the regression test. @minh2416294

aryanbarde80:
Thanks for picking this up @minh2416294 and adding the regression test. Glad to see #18869 getting fixed!

abhiaiyer91 (changes requested):
## Review: request changes

The one-line fix is internally consistent with the sibling OM markers and the tests/CI are green, but `data-om-status` is not write-only — three Studio code paths read it back out of **persisted** message history. Dropping persistence silently changes their behavior, and that isn't acknowledged in the PR body, the changeset, or the tests.

### Verification I ran

- `pnpm install` + `pnpm turbo build --filter ./packages/memory` — clean
- `vitest run observational-memory.test.ts` — **453 passed**, 1 todo
- `vitest run ... -t "transient flag"` — passes
- `gh pr checks 19084` — build + test green; no merge conflicts

So this is not about correctness of the diff as written; it's about scope.

### Blocking 1 — Studio loses OM progress after a page reload

`packages/playground/src/lib/ai-ui/chat/chat-provider.tsx:202-211` scans persisted messages on mount, with the comment: *"On initial load, scan messages for activation markers + last progress so buffering badges show as activated and token counts are accurate on reload."* The scanner it calls reads exactly this part — `packages/playground/src/services/om-parts-converter.ts:586` (`part?.type === 'data-om-status'`). After this change `lastProgress` is always `undefined` on reload, so `streamProgress` stays `null`.

Most consumers survive that, because `getObservationWindowTokens` falls back to the OM record (`memory-detail-view.tsx:82-87`, `agent-observational-memory.tsx:452-459`). One does not: `packages/playground/src/domains/agents/components/memory-sidebar/memory-sidebar.tsx:110-115` derives the collapsed memory card's percentage from `streamProgress` **only**, with no record fallback — so that indicator goes blank after reload.

### Blocking 2 — memory studio timeline loses its status series

`packages/playground-ui/src/domains/memory/lib/extract-markers.ts:73-81` maps persisted `data-om-status` parts into `status` markers, which are consumed as time-series points by `flame-graph.tsx:60-92` and preferred over `omRecords` in `thread-context-window-state.ts:32-43`. Records only update when an observation actually runs, so the pending-message-token and active-observation curves become materially sparser between observations. (The headline numbers stay correct — `memory-detail-view.tsx:98-103` passes explicit `contextWindow` values that take precedence — the graph fidelity is what degrades.)

Neither path has test coverage that would catch this: the playground-ui fixtures hand-construct `data-om-status` parts (`__tests__/fixtures/memory-studio.ts:53`), so they keep passing regardless of what the emitter does.

### Non-blocking — the regression test restates the diff

The new test asserts the emitted part carries `transient: true`, which is the change itself. #18869 is about a row landing in `mastra_messages`. A test that fails on `main` for the reported reason would assert through the persistence gate (`packages/core/src/loop/workflows/agentic-execution/llm-execution-step.ts:118`) or at storage level that no standalone assistant message is written.

### Non-blocking — half of the sibling pattern

The sibling markers pair `transient: true` with an explicit `persistMarkerToStorage(...)` (`observational-memory.ts:2138`, `:3066`) — transient for the writer, durable on purpose. This PR adopts the first half only. Given the readers above, some durable status history does appear to be relied upon.

Also worth flagging: on #18869 both @bkidy and you explicitly asked a maintainer whether persisting `data-om-status` was intentional, and neither question was answered. The in-repo signals cut both ways — `channels/om.ts:196` says channels intentionally skip it, but the Studio reload and timeline code reads it back from storage.

### Suggested path forward

Either:

1. Keep the fix and update the readers to source status from the OM record — `AgentController.loadOMProgress()` (`packages/core/src/agent-controller/agent-controller.ts:1387`) already reconstructs status from the record and emits `om_status`, so there's precedent for a record-backed restore; or
2. Get maintainer confirmation that losing persisted status snapshots is acceptable, and call the Studio behavior change out in the changeset.

Happy to re-review quickly once one of those is settled — the underlying bug is real and worth fixing.

minh2416294:
thanks for digging into @abhiaiyer91. you're right that the PR as written is incomplete, and i went and verified the reader analysis against the tree. two refinements, one of which affects your suggested path.

the blast radius is larger than studio. AgentController.loadOMProgress() (packages/core/src/agent-controller/agent-controller.ts:1387) also rebuilds OM state by scanning persisted messages for data-om-status, and its called from the mastracode TUI (mastra-tui.ts:654, event-dispatch.ts:213). it pre-seeds only messageTokens/observationTokens/thresholds from the record (1410-1414), so bufferedObs, bufferedRef, generationCount and stepNumber have no record fallback. post-fix foundStatus is never true, so a thread thats reflected a few times with buffered chunks pending reopens reporting generation 0 and an idle all-zero buffer. silent, no error.

which means loadOMProgress isnt the precedent for option 1, its another consumer. re-pointing readers at the record would have to cover core as well as the two playground packages, so option 1 is a three-package change rather than a follow-up to a one line fix.

one narrowing on the timeline point. the ThreadContextProgress bars dont actually degrade, memory-detail-view.tsx:98-103 passes explicit contextWindow values that take precedence. what does degrade is flame graph resolution between observations, and standalone MemoryStudioPanel consumers that omit the contextWindow prop, those lose the numeric readout rather than just fidelity.

on the test, agreed, thats a fair hit. asserting transient: true restates the diff. ill rewrite it to assert the reported symptom through the persistence gate, that no standalone assistant message lands in storage for a status emit.

on scoping, i looked at the "adopt the full sibling pattern" idea (transient: true plus persistMarkerToStorage) since it would fix the standalone rows without a cross package migration. it doesnt work as is, for two reasons worth recording:

- persistMarkerToStorage dedups on data.cycleId (~line 1176) and the status part has no cycleId, it carries recordId/stepNumber/generationCount. so alreadyPresent is always false and it would append a status part on every call.
- emitProgress runs once per input step, unconditionally (processor.ts:312), not once per cycle like the lifecycle markers. so that append plus its listMessages+persistMessages round trip would happen every step.

that would just move the bloat from standalone rows into one assistant messages parts array, and add a db read per step. it could be made to work with replace-in-place semantics instead of append (readers only ever want the latest snapshot, scanOmInitialState does lastProgress = part.data), and persisting on a turn boundary rather than per step, but thats a design change rather than adopting an existing pattern.

so which way do you want this scoped?

1. full reader migration across core + playground + playground-ui. id want your sign off before touching core.
2. accept losing persisted status snapshots, and i document the studio and TUI behaviour change explicitly in the changeset.
3. status persisted with replace-in-place on a turn boundary, keeping consumers working, as sketched above.

happy to implement whichever. 3 feels closest to right to me but its the most design so its your call.

CodingButter (changes requested):
---

You correctly diagnosed the asymmetry, the sibling markers all set `transient: true` and the status emit does not, and the history pollution on long threads is real. Verified your regression test is genuinely red on main and green on the branch. Unfortunately the one-liner cannot land as-is, because one consumer reads `data-om-status` back out of persisted history.

## Must fix

1. **`loadOMProgress` restores OM status from the persisted part; this PR silently breaks that.** `packages/core/src/agent-controller/agent-controller.ts:1404-1495` lists recent messages on thread switch/reload and scans persisted assistant parts for the most recent `data-om-status`, pulling the real configured thresholds, both `buffered` blocks, `generationCount`, and `stepNumber` out of it before emitting `om_status` to the UI. With the part transient, that path finds nothing on new threads: thresholds fall back to the hardcoded `30_000`/`40_000`, buffered observation/reflection state resets to idle, and step/generation counters zero out. The degradation is silent (old threads still have persisted parts, new ones do not), which is the worst kind. There is an existing regression test guarding exactly this class of UI bug (the "memory X/0k instead of X/40k" test in the OM suite), so the fidelity was deliberate.

   The sibling precedent the fix copies is not fully analogous either: the buffering marker is transient *because* a durable marker is persisted separately via `persistMarkerToStorage` right before it. The status part has no durable counterpart, it was doing double duty as both stream signal and restore source.

   Two ways forward, maintainer's call:
   - Give `loadOMProgress` a durable source for the fields it cannot get from the OM record (persist a compact status snapshot on the OM record itself, or via `persistMarkerToStorage` on state transitions rather than every step), then make the stream part transient. This fixes the pollution without losing restore fidelity.
   - Or accept the degraded fallback explicitly, state it in the PR body and changeset, and update or remove the restore path so the behavior is chosen rather than accidental.

2. **The test asserts the changed line, not the behavior.** It captures the writer and checks `transient === true`, which restates the diff. It cannot see the `loadOMProgress` regression, and it would pass even if another layer still persisted the part. A meaningful test drives a turn through the OM processor with real storage and asserts `listMessages` contains no assistant message whose only part is `data-om-status` (the harness exists, the mock-agent and long-session OM tests run full turns), paired with a test pinning whatever restore behavior gets chosen for point 1.

## Notes, not blocking

- The changeset says only "no longer persisted"; whichever direction point 1 goes, the user-visible consequence (status restore fidelity after reload) belongs in it.
- The OM reference docs document `data-om-status` as a stream part without addressing persistence; a line stating it is stream-only (once true) would prevent the next person from building on the persisted shape.
- The `{ ...statusPart, transient: true }` spread itself is fine, `statusPart` is a fresh local and the `.catch` swallow predates you. If you end up keeping the line, setting the flag in the object literal at the construction site reads better than bolting it on at the call.

The issue you filed against is legitimate and worth fixing, this just needs the restore path handled in the same change rather than left to break quietly. Happy to re-review.

minh2416294:
Thanks for the careful reading, and for testing things out both ways.

on your point 1 - I agree, and actually raised this exact question with @abhiaiyer91 on jul 31. The one who breaks is the consumer, specifically the fields that don't have record fallback are the buffered blocks, generationCount and stepNumber. worth noting the sibling asymmetry you mention is the same conclusion he and I arrived at independently, the buffering marker gets to be transient because persistMarkerToStorage writes a durable equivalent, and the status part never had one.

on your point 2 - it's fair, the test reasserts the difference. the version I'd write would turn the OM processor against InMemory storage, and assert that listMessages has no assistant message with only a data-om-status part, which is precisely the symptom #18869 is documenting, plus a test pinning whichever restore behaviour we decide on.

The blocker is that both you and abhiaiyer91 have declared the direction a maintainer's call, and I'm yet to get an answer. Rather than ask a third time, here's what I'm planning to do, unless anyone tells me otherwise:

persist a compact status snapshot on state transitions, rather than every step (the current emit is once per input step, and unconditional, see processor.ts:312), make the streamed part transient, and point loadOMProgress to that durable source. That should eliminate the standalone records while retaining the ability to restore from the durable pointer, within packages/memory.

@abhiaiyer91 if you think that taking the easy way out and documenting the degraded fallback in the changeset, while modifying the restore path to use it by default rather than accident, would be preferable, just say the word, and I'll do that instead. either way, I'd rather do what you want than assume.

One unrelated request - there are six workflow runs stuck on "Approve and run", from the outside-contributor bot, so CI has never actually completed on this branch. If anyone has a maintainer account and can approve those, it'd be much appreciated.

minh2416294:
superseded by #21604, which carries the same transient fix plus the loadOMProgress restore work that @abhiaiyer91 and @CodingButter correctly flagged as missing here. closing this one.

thanks @abhiaiyer91 and @CodingButter for the reviews - the loadOMProgress dependency was the thing that made this more than a one-liner, and #21604 handles it properly. i've left review notes there on the playground readers, which look like they still need the same treatment.

PR #19085, opened 2026-07-08
test(core): add unit tests for message-list cache and utils helpers

### LINKED ISSUE #19087: Add unit tests for message-list cache and utils helpers
Several small, pure helper modules under `packages/core/src/agent/message-list/` have no direct unit test coverage:

- `cache/stable-stringify.ts` (`stableStringify`) — deterministic JSON key-ordering used for cache-key generation; the source doc comment explicitly calls out a jsonb-vs-text-column storage scenario that depends on this being correct.
- `utils/stamp-part.ts` (`stampPart`, `stampMessageParts`) — assigns `createdAt` to message parts, with a `'memory'` source short-circuit and a falsy-but-valid `createdAt = 0` case that's easy to break with a naive falsy check.
- `utils/tool-name.ts` (`sanitizeToolName`) — regex allow-list guard for tool names before they're used downstream.
- `utils/response-item-metadata.ts` (`getResponseProviderItemId(s)`, `getResponseProviderItemKey(s)`) — implements a specific azure/openai provider namespace-collision merge rule.

Scope: add `vitest` unit tests for each module's exported functions, covering the documented edge cases above plus standard null/undefined/empty-input handling.

aryanbarde80:
Fixes #19087

## What

Adds unit test coverage for four small, previously-untested pure helper modules under `packages/core/src/agent/message-list/`:

- `cache/stable-stringify.ts` — `stableStringify`
- `utils/stamp-part.ts` — `stampPart`, `stampMessageParts`
- `utils/tool-name.ts` — `sanitizeToolName`
- `utils/response-item-metadata.ts` — `getResponseProviderItemId(s)`, `getResponseProviderItemKey(s)`

All four are pure functions with no I/O, so the tests require no mocking beyond faking `Date.now()` for the `stamp-part` timestamp assertions.

## Why

These modules had no direct test file even though they sit on paths that matter for correctness:

- `stableStringify` is explicitly documented as critical for cache-key generation across storage backends that normalize JSON key order differently (jsonb vs text columns) — an untested regression here would silently break cache dedup.
- `stampPart`/`stampMessageParts` control whether/when message parts get a `createdAt` timestamp, including a `'memory'` source short-circuit and a falsy-but-valid `createdAt = 0` edge case that's easy to break with a naive falsy check.
- `sanitizeToolName` is a security-relevant allow-list guard for tool names; tests pin down the exact fallback behavior for unsafe/unicode input.
- `response-item-metadata` helpers implement a specific azure/openai namespace-collision merge rule that's non-obvious from a quick read and worth locking in with tests.

## Coverage

54 test cases across 4 files, covering: happy paths, documented edge cases from source comments, falsy-vs-nullish distinctions, empty/undefined inputs, and the provider-collision merge behavior.

## Testing

- `npx vitest run` on all four new files: **54/54 passing**
- `npx eslint` on all four new files: clean

## ELI5
This PR adds tests for a few small helper functions that do important “housekeeping” work, like keeping cache keys consistent, adding timestamps, cleaning up tool names, and matching response metadata. These tests help make sure the helpers keep behaving the same way over time, especially in edge cases.

## Summary
- Added Vitest coverage for `stableStringify` to verify deterministic object serialization, stable nested key ordering, array handling, `undefined`/`null` behavior, and realistic payload cases.
- Added tests for `stampPart` and `stampMessageParts` to confirm timestamp stamping rules, `memory` source behavior, preservation of existing timestamps (including `0`), and safe handling of missing/empty parts.
- Added tests for `sanitizeToolName` to validate allowed names, fallback behavior for invalid or non-string inputs, and the fallback sentinel value.
- Added tests for response-item metadata helpers to verify provider/item ID extraction, Azure/OpenAI collision merging rules, key formatting, and ordering behavior.

No production code was changed.

### DIFF
--- a/packages/core/src/agent/message-list/cache/stable-stringify.test.ts
+++ b/packages/core/src/agent/message-list/cache/stable-stringify.test.ts
@@ -1,0 +1,113 @@
+/**
+ * Tests for packages/core/src/agent/message-list/cache/stable-stringify.ts
+ *
+ * `stableStringify` is a pure function with no I/O and no async behaviour.
+ * The behaviour under test is the deterministic-key-order guarantee that
+ * makes it safe to use for cache-key generation, as described in the
+ * function's own doc comment.
+ */
+import { describe, expect, it } from 'vitest';
+
+import { stableStringify } from './stable-stringify';
+
+describe('stableStringify', () => {
+  it('produces identical output regardless of key insertion order', () => {
+    const a = { a: 1, b: 2 };
+    const b = { b: 2, a: 1 };
+
+    expect(stableStringify(a)).toBe(stableStringify(b));
+  });
+
+  it('sorts keys alphabetically at the top level', () => {
+    const result = stableStringify({ z: 1, a: 2, m: 3 });
+
+    expect(result).toBe('{"a":2,"m":3,"z":1}');
+  });
+
+  it('sorts keys at every nested level, not just the top level', () => {
+    const nested = { outer2: 1, outer1: { innerB: 1, innerA: 2 } };
+
+    expect(stableStringify(nested)).toBe('{"outer1":{"innerA":2,"innerB":1},"outer2":1}');
+  });
+
+  it('produces identical output for deeply nested objects with different key orders', () => {
+    const a = { x: { y: { c: 1, b: 2, a: 3 } } };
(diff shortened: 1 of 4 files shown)


aryanbarde80:
Superseded by #19089 (same changes, with the linked issue #19087 included from the start).

PR #19086, opened 2026-07-08
test(core): add unit tests for processor-workflow guard and prompt content helpers

aryanbarde80:
## What

Adds unit test coverage for three previously-untested pure modules:

- `processors/is-processor-workflow.ts` — `isProcessorWorkflow`
- `agent/message-list/prompt/invalid-content-error.ts` — `InvalidDataContentError`
- `agent/message-list/prompt/data-content.ts` — `convertDataContentToBase64String`

## Why

- `isProcessorWorkflow` is a type guard used to tell a real `Workflow` apart from a `Processor` (this repo's own doc comment notes it's factored out specifically to avoid ESM init-time cycles from the processors barrel). Its correctness hinges entirely on excluding every `process*`-prefixed method — tests pin down each of the 7 documented exclusions (`processInput`, `processInputStep`, `processOutputStream`, `processOutputResult`, `processOutputStep`, `processLLMRequest`, `processAPIError`) individually, plus the required-field and non-object negative cases.
- `InvalidDataContentError` uses a `Symbol.for(...)` marker-based `isInstance` check instead of `instanceof` specifically so it still works across bundler/realm boundaries — that's exactly the kind of non-obvious design choice worth locking in with a test, along with the default-message `typeof` interpolation and the message/cause override paths.
- `convertDataContentToBase64String` fans out on 3 different runtime input shapes (string / ArrayBuffer / Uint8Array). Tests cover all three branches, empty-input edge cases for each, that `Buffer` correctly falls through the `Uint8Array` branch, and a full binary round-trip (0x00/0xff bytes) to catch any future encoding regression.

## Coverage

39 test cases across 3 files: happy paths, all documented branch/exclusion conditions, and null/undefined/primitive/empty edge cases.

## Testing

- `npx vitest run` on all three new files: **39/39 passing**
- `npx eslint` on all three new files: clean

## ELI5
This PR adds tests for three small helper pieces of code so we can be more sure they work the way we expect. It checks how they recognize valid workflows, how a custom error behaves, and how different kinds of data get turned into base64 text.

### Summary
- Added coverage for `isProcessorWorkflow`, including valid workflow shapes, missing/invalid required fields, non-objects, and `process*` method exclusions.
- Added coverage for `InvalidDataContentError`, including default/custom messages, `cause`, stored content, and marker-based instance detection.
- Added coverage for `convertDataContentToBase64String`, including string passthrough, empty input handling, `ArrayBuffer`, `Uint8Array`/`Buffer`, and binary round-trip behavior.
- No production code changes.
- 39 new test cases across 3 files.

### DIFF
--- a/packages/core/src/agent/message-list/prompt/data-content.test.ts
+++ b/packages/core/src/agent/message-list/prompt/data-content.test.ts
@@ -1,0 +1,62 @@
+/**
+ * Tests for packages/core/src/agent/message-list/prompt/data-content.ts
+ *
+ * `convertDataContentToBase64String` is a pure function with no I/O. It
+ * fans out on the runtime type of its input (string passthrough,
+ * ArrayBuffer, or Uint8Array/Buffer), so coverage exercises each branch
+ * plus the documented edge cases (empty input, Buffer-is-a-Uint8Array).
+ */
+import { describe, expect, it } from 'vitest';
+
+import { convertDataContentToBase64String } from './data-content';
+
+describe('convertDataContentToBase64String', () => {
+  it('returns a string input unchanged (no re-encoding)', () => {
+    expect(convertDataContentToBase64String('already-base64==')).toBe('already-base64==');
+  });
+
+  it('returns an empty string input unchanged', () => {
+    expect(convertDataContentToBase64String('')).toBe('');
+  });
+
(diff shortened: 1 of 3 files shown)


aryanbarde80:
Superseded by #19090 (same changes, with the linked issue #19088 included from the start).

PR #19090, opened 2026-07-08
test(core): add unit tests for processor-workflow guard and prompt content helpers

### LINKED ISSUE #19088: Add unit tests for processor-workflow guard and prompt content helpers
Three small, pure modules used in the agent/processor pipeline have no direct unit test coverage:

- `processors/is-processor-workflow.ts` (`isProcessorWorkflow`) — type guard that tells a real `Workflow` apart from a `Processor`; its own doc comment notes it was factored out specifically to avoid ESM init-time cycles from the processors barrel, and its correctness depends on excluding every `process*`-prefixed method (`processInput`, `processInputStep`, `processOutputStream`, `processOutputResult`, `processOutputStep`, `processLLMRequest`, `processAPIError`).
- `agent/message-list/prompt/invalid-content-error.ts` (`InvalidDataContentError`) — uses a `Symbol.for(...)` marker-based `isInstance` check instead of `instanceof` so it still works across bundler/realm boundaries; this non-obvious design choice is currently unverified by any test.
- `agent/message-list/prompt/data-content.ts` (`convertDataContentToBase64String`) — fans out on 3 different runtime input shapes (string / ArrayBuffer / Uint8Array/Buffer) with no coverage of any branch.

Scope: add `vitest` unit tests for each module's exported functions/classes, covering the documented edge cases above plus standard null/undefined/empty-input handling.

aryanbarde80:
Fixes #19088

## What

Adds unit test coverage for three previously-untested pure modules:

- `processors/is-processor-workflow.ts` — `isProcessorWorkflow`
- `agent/message-list/prompt/invalid-content-error.ts` — `InvalidDataContentError`
- `agent/message-list/prompt/data-content.ts` — `convertDataContentToBase64String`

## Why

- `isProcessorWorkflow` is a type guard used to tell a real `Workflow` apart from a `Processor` (this repo's own doc comment notes it's factored out specifically to avoid ESM init-time cycles from the processors barrel). Its correctness hinges entirely on excluding every `process*`-prefixed method — tests pin down each of the 7 documented exclusions (`processInput`, `processInputStep`, `processOutputStream`, `processOutputResult`, `processOutputStep`, `processLLMRequest`, `processAPIError`) individually, plus the required-field and non-object negative cases.
- `InvalidDataContentError` uses a `Symbol.for(...)` marker-based `isInstance` check instead of `instanceof` specifically so it still works across bundler/realm boundaries — that's exactly the kind of non-obvious design choice worth locking in with a test, along with the default-message `typeof` interpolation and the message/cause override paths.
- `convertDataContentToBase64String` fans out on 3 different runtime input shapes (string / ArrayBuffer / Uint8Array). Tests cover all three branches, empty-input edge cases for each, that `Buffer` correctly falls through the `Uint8Array` branch, and a full binary round-trip (0x00/0xff bytes) to catch any future encoding regression.

## Coverage

39 test cases across 3 files: happy paths, all documented branch/exclusion conditions, and null/undefined/primitive/empty edge cases.

## Testing

- `npx vitest run` on all three new files: **39/39 passing**
- `npx eslint` on all three new files: clean

## ELI5
This PR adds missing unit tests for three helper functions that help the system (1) tell whether something is a real “workflow”, (2) produce consistent errors when message content is invalid, and (3) convert binary data into base64 text. The goal is to prevent subtle bugs by locking in the expected behavior for both normal and tricky inputs.

## Summary
- Added Vitest coverage for `isProcessorWorkflow`, including:
  - Positive detection of workflow-like objects
  - Excluding any object that looks like a processor via documented `process*` methods (`processInput`, `processInputStep`, `processOutputStream`, `processOutputResult`, `processOutputStep`, `processLLMRequest`, `processAPIError`)
  - Negative cases for `null`/`undefined`, primitives, arrays, empty objects, and missing/wrong-typed required fields
  - Confirmation that extra unrelated properties don’t affect results
- Added Vitest coverage for `InvalidDataContentError`, including:
  - Error subclassing behavior
  - Default message formatting (including checks based on `typeof` of invalid content)
  - Proper propagation/override of `message` and `cause`
  - `Symbol.for(...)` marker-based `isInstance` behavior across boundaries (instead of relying on `instanceof`)
- Added Vitest coverage for `convertDataContentToBase64String`, including:
  - String passthrough behavior
  - Empty-input handling
  - Base64 conversion for `ArrayBuffer`, `Uint8Array` (and `Buffer` via the typed-array path), including empty typed/buffer inputs
  - Cross-type equivalence between `Uint8Array` and `ArrayBuffer`
  - Correct round-tripping for arbitrary binary data

## Testing
- 39 Vitest cases passing
- Lint clean

### DIFF
--- a/packages/core/src/agent/message-list/prompt/data-content.test.ts
+++ b/packages/core/src/agent/message-list/prompt/data-content.test.ts
@@ -1,0 +1,62 @@
+/**
+ * Tests for packages/core/src/agent/message-list/prompt/data-content.ts
+ *
+ * `convertDataContentToBase64String` is a pure function with no I/O. It
+ * fans out on the runtime type of its input (string passthrough,
+ * ArrayBuffer, or Uint8Array/Buffer), so coverage exercises each branch
+ * plus the documented edge cases (empty input, Buffer-is-a-Uint8Array).
+ */
+import { describe, expect, it } from 'vitest';
+
+import { convertDataContentToBase64String } from './data-content';
+
+describe('convertDataContentToBase64String', () => {
+  it('returns a string input unchanged (no re-encoding)', () => {
+    expect(convertDataContentToBase64String('already-base64==')).toBe('already-base64==');
+  });
+
+  it('returns an empty string input unchanged', () => {
+    expect(convertDataContentToBase64String('')).toBe('');
+  });
+
+  it('encodes a Uint8Array to a base64 string', () => {
+    const bytes = new Uint8Array([104, 101, 108, 108, 111]); // "hello"
+
+    expect(convertDataContentToBase64String(bytes)).toBe('aGVsbG8=');
+  });
+
+  it('encodes an ArrayBuffer to a base64 string', () => {
+    const buffer = new Uint8Array([104, 101, 108, 108, 111]).buffer; // "hello"
+
+    expect(convertDataContentToBase64String(buffer)).toBe('aGVsbG8=');
+  });
+
+  it('produces the same base64 output for equivalent Uint8Array and ArrayBuffer input', () => {
+    const bytes = new Uint8Array([1, 2, 3, 4]);
+    const buffer = bytes.buffer;
+
+    expect(convertDataContentToBase64String(bytes)).toBe(convertDataContentToBase64String(buffer));
+  });
+
+  it('encodes an empty Uint8Array to an empty base64 string', () => {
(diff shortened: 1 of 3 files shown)


aryanbarde80:
@coderabbitai good catch, fixed in 66db036 — simplified the description to `returns false for an array` since the test just uses an empty array.

intojhanurag:
closed issue , thanks !

PR #19103, opened 2026-07-08
fix(core): forward toolPayloadTransform through loop internal state

### LINKED ISSUE #19102: loop() drops _internal.toolPayloadTransform, silently disabling tool payload transforms
### Describe the Bug

`loop()` silently drops the tool payload transform policy. When an agent is configured with a `transform` policy, the non-durable stream path sets `_internal.toolPayloadTransform` and passes it into `loop()`. But `loop()` rebuilds the `_internal` bag into a fresh `internalToUse` object and copies every consumed field except `toolPayloadTransform`:

`packages/core/src/loop/loop.ts`:

```ts
const internalToUse: StreamInternal = {
  now: _internal?.now || (() => Date.now()),
  // ... every other field copied ...
  drainPendingSignals: _internal?.drainPendingSignals,
  initialSignalEchoes: _internal?.initialSignalEchoes ? [..._internal.initialSignalEchoes] : undefined,
  // toolPayloadTransform is never copied
};
```

This rebuilt `internalToUse` is the authoritative bag: it becomes `workflowLoopProps._internal`, and `workflows/stream.ts` hydrates the run scope from it via `hydrateRunScopeFromInternal`. Because the field is stripped before hydration, the scope slot for `TOOL_PAYLOAD_TRANSFORM_KEY` is never set. The consumer reads it with `readScoped(scopeCtx, TOOL_PAYLOAD_TRANSFORM_KEY, 'toolPayloadTransform')`, whose fallback also reads the same stripped bag, so the policy is `undefined` everywhere.

The sibling hydrator copies the field (`hydrate-run-scope.ts`: `if (internal.toolPayloadTransform) scope.set(TOOL_PAYLOAD_TRANSFORM_KEY, internal.toolPayloadTransform)`), and there is a round-trip test for it in `run-scope-helpers.test.ts`, so the intent is clearly that the field flows through.

### Steps To Reproduce

Stream from an agent configured with a `transform` (tool payload transform) policy, using a model that emits a tool call. The transform is never applied to any tool-call / tool-result / tool-call-delta chunk.

The unit tests that cover the transform (`llm-execution-step.test.ts`) do not catch this because they construct the step factory directly with `_internal.toolPayloadTransform`, bypassing `loop()`'s reconstruction. Only an end-to-end `loop()` run exercises the drop.

### Expected Behavior

The configured tool payload transform policy is applied to tool payload chunks, exactly as when the policy is read directly by the step.

### Actual Behavior

The policy is dropped by `loop()`, so it silently no-ops for every non-durable agent stream that configured one. The durable path is unaffected because its steps read the policy from their registry entry directly.

### Additional Context

One-line fix: add `toolPayloadTransform: _internal?.toolPayloadTransform` to `internalToUse`. I have a PR ready with an end-to-end regression test that drives `loop()` with a tool call and asserts the policy is invoked.

abhay-codes07:
## What

`loop()` silently dropped the tool payload transform policy. When an agent is configured with a `transform` policy, the non-durable stream path sets `_internal.toolPayloadTransform` and hands it to `loop()`. But `loop()` rebuilds the `_internal` bag into a fresh `internalToUse` object and copies every consumed field except this one:

```ts
const internalToUse: StreamInternal = {
  now: _internal?.now || (() => Date.now()),
  // ... every other field copied ...
  initialSignalEchoes: _internal?.initialSignalEchoes ? [..._internal.initialSignalEchoes] : undefined,
  // toolPayloadTransform was never copied
};
```

That rebuilt bag is the authoritative one: it becomes `workflowLoopProps._internal`, and `workflows/stream.ts` hydrates the run scope from it via `hydrateRunScopeFromInternal`. Since the field is stripped before hydration, the scope slot for `TOOL_PAYLOAD_TRANSFORM_KEY` is never set, and the consumer's `readScoped(scopeCtx, TOOL_PAYLOAD_TRANSFORM_KEY, 'toolPayloadTransform')` fallback reads the same stripped bag, so the policy is `undefined` everywhere. The transform silently no-ops for the whole run.

The sibling hydrator already copies the field (`hydrate-run-scope.ts`: `if (internal.toolPayloadTransform) scope.set(...)`), with a round-trip test in `run-scope-helpers.test.ts`, so the field is meant to flow through. The durable path is unaffected because its steps read the policy from their registry entry directly.

## Fix

Forward `toolPayloadTransform` in the `internalToUse` reconstruction, matching every other consumed field.

## Why it was not caught

The existing unit tests for the transform (`llm-execution-step.test.ts`) construct the step factory directly with `_internal.toolPayloadTransform`, bypassing `loop()`'s reconstruction. Only an end-to-end `loop()` run exercises the drop.

## Tests

Added `tool-payload-transform-forwarding.test.ts`, which drives a real `loop()` run with a model that emits a tool call and a transform policy, then asserts the policy is actually invoked. It fails on the current code (`transformInvoked` stays false) and passes with the fix. The surrounding suites stay green (`loop.test.ts` 149 passed, `run-scope-helpers` + `llm-execution-step` 43 passed).

Closes #19102

## ELI5
This change makes sure a tool-payload transform setting doesn’t get lost when `loop()` starts a run. In plain terms, if you’ve told the system to reshape tool-related data, it now keeps doing that instead of forgetting.

## What changed
- `loop()` now carries `_internal.toolPayloadTransform` into its rebuilt internal state.
- This matches the behavior of the sibling run-scope hydrator, so tool payload transforms continue to apply during non-durable agent streams.
- Added an end-to-end regression test that runs `loop()` with a tool call and verifies the transform policy is actually invoked.

## Why
- Fixes the bug where `loop()` dropped the configured transform policy and tool payload transformations stopped working for the run.
- Covers the real execution path so this regression is caught going forward.


intojhanurag:
Thanks for opening this PR.

This changes `packages/core/src/loop/loop.ts`, which is a critical core agent execution path covering ordering, streaming, tool calls, and resume behavior. For external contributors, changes to this path require internal ownership, so this PR can’t proceed through normal review.

The linked issue #19102 remains the right place for maintainer follow-up on the reported behavior.

intojhanurag (approved):

PR #19135, opened 2026-07-08
feat(memory): one-shot conversation summarization — summarizeConversation() / Memory.summarizeThread()

mikhael28:
## Description

This is PR 1 of 3 split from #19019 (as [requested](https://github.com/mastra-ai/mastra/pull/19019#issuecomment-4910853761) by @TylerBarnes). Stack: **memory summarization (this PR)** → livekit worker configuration (#19136) → voice-agent example (#19137).

Adds one-shot conversation summarization to `@mastra/memory`: a standalone `summarizeConversation()` (takes messages) and `Memory.summarizeThread()` (loads the thread for you). Both distill a whole conversation using the same Observer plumbing that powers Observational Memory — **without** attaching OM to an agent and **without** writing anything back to memory. They return `{ summary, extracted }`, and each `Extractor` (typed via Zod schema) gets its `onExtracted` hook called, so structured data (sentiment, requested services, follow-ups) lands in the application's own database:

```ts
await memory.summarizeThread({
  model: 'openai/gpt-5-mini',
  threadId,
  resourceId,
  instructions: "Summarize this call for the contractor's office: who called, what they wanted, any follow-up promised.",
  extract: [callSummaryExtractor], // Zod-typed extraction → app storage via onExtracted
});
```

This is the third iteration of the design, landed after maintainer feedback on the original PR: `observe({ force: true })`, then a buffer-path `flush()`, were both OM-lifecycle solutions to what is really a distillation use case. The final shape leaves the OM engine **untouched** (`observational-memory.ts` diff vs main is empty).

## Scope

- `summarizeConversation()` + `Memory.summarizeThread()` (`packages/memory`)
- Reference docs page (`reference/memory/summarizeThread`) + sidebar entry
- `summarize.test.ts` — 6 tests: summary output, instructions, inline + structured extractors, empty-thread no-op, `summarizeThread` round-trip
- Changeset: `@mastra/memory` minor

## Intentionally excluded

- `@mastra/livekit` worker `configuration` (greeting/disclosure, consent, endCall, stt/tts resolvers) and the `MastraLLM` plugin — PR 2
- `examples/voice-agent` updates (permissive default + regulated demo) — PR 3

## Verification

- `summarize.test.ts` — 6/6 passing
- `pnpm --filter ./packages/memory check` — clean

## Type of change

- [x] Documentation update

## ELI5
This PR lets `@mastra/memory` turn a whole conversation (or an entire saved thread) into a short summary and pull out key “facts” it finds—without saving anything back. It’s like a one-time read-and-summarize operation for callers.

### What changed
- Added `summarizeConversation()` for one-shot summarization from a provided list of messages.
- Added `Memory.summarizeThread()` to load messages for a thread and then summarize them in one call.
- Reused the existing Observer/extractor pipeline (including extractor `onExtracted` hooks), but without attaching Observational Memory to an agent and without writing to memory.
- Supported bounded thread loading via `lastMessages` and `maxInputTokens`.
- Threaded `abortSignal` through the thread-loading loop and added abort checks so cancellation stops paging/loading promptly (not just the model call).

### Documentation
- Added first-class reference docs for both `reference/memory/summarizeConversation` and `reference/memory/summarizeThread`, plus a sidebar entry.

### Tests
- Added/expanded `summarize.test.ts` coverage for:
  - summary output + usage tokens
  - custom `instructions`
  - inline and structured extractors (including extractor `onExtracted`)
  - empty-thread no-op behavior returning `{ summary: '', extracted: {} }`
  - `lastMessages` / `maxInputTokens` paging + ordering behavior
  - abort-signal behavior during thread paging/loading
  - surfaced structured-extractor failures via `extractionFailures` (and omission of the failed value from `extracted`, with `onExtracted` not called)

### DIFF
--- a/docs/src/content/en/reference/sidebars.js
+++ b/docs/src/content/en/reference/sidebars.js
@@ -9,7 +9,6 @@ const sidebars = {
   referenceSidebar: [
     { type: 'doc', id: 'index', label: 'Overview' },
     { type: 'doc', id: 'configuration', label: 'Configuration' },
-    { type: 'doc', id: 'project-structure', label: 'Project Structure' },
     {
       type: 'category',
       label: 'ACP',
@@ -165,19 +164,6 @@ const sidebars = {
         { type: 'doc', id: 'client-js/tools', label: 'Tools API' },
         { type: 'doc', id: 'client-js/vectors', label: 'Vectors API' },
         { type: 'doc', id: 'client-js/workflows', label: 'Workflows API' },
-      ],
-    },
-    {
-      type: 'category',
-      label: 'Code SDK',
-      collapsed: true,
-      items: [
-        {
-          type: 'doc',
-          id: 'code-sdk/mount-agent-controller',
-          label: 'mountAgentControllerOnMastra()',
-          customProps: { tags: ['beta'] },
-        },
       ],
     },
     {
@@ -379,6 +365,7 @@ const sidebars = {
         { type: 'doc', id: 'memory/getThreadById', label: '.getThreadById()' },
         { type: 'doc', id: 'memory/listThreads', label: '.listThreads()' },
         { type: 'doc', id: 'memory/recall', label: '.recall()' },
+        { type: 'doc', id: 'memory/summarizeThread', label: '.summarizeThread()' },
       ],
     },
     {
@@ -560,12 +547,6 @@ const sidebars = {
     },
     {
       type: 'category',
-      label: 'Schedules',
-      collapsed: true,
-      items: [{ type: 'doc', id: 'schedules/overview', label: 'Overview' }],
-    },
-    {
-      type: 'category',
       label: 'Server',
       collapsed: true,
       items: [
@@ -679,7 +660,6 @@ const sidebars = {
       label: 'Tools and MCP',
       collapsed: true,
       items: [
-        { type: 'doc', id: 'tools/ask-user-tool', label: 'askUserTool' },
         { type: 'doc', id: 'tools/brightdata', label: 'Bright Data Tools' },
         {
           type: 'doc',
@@ -694,8 +674,6 @@ const sidebars = {
         { type: 'doc', id: 'tools/mcp-client', label: 'MCPClient' },
         { type: 'doc', id: 'tools/mcp-server', label: 'MCPServer' },
         { type: 'doc', id: 'tools/perplexity', label: 'Perplexity Tools' },
-        { type: 'doc', id: 'tools/submit-plan-tool', label: 'submitPlanTool' },
-        { type: 'doc', id: 'tools/task-tools', label: 'Task tools' },
         { type: 'doc', id: 'tools/tavily', label: 'Tavily Tools' },
       ],
     },
--- a/docs/src/content/en/reference/sidebars.js
+++ b/docs/src/content/en/reference/sidebars.js
@@ -373,6 +373,7 @@ const sidebars = {
         { type: 'doc', id: 'memory/memory-class', label: 'Memory Class' },
         { type: 'doc', id: 'memory/observational-memory', label: 'Observational Memory' },
         { type: 'doc', id: 'memory/serialized-memory-config', label: 'SerializedMemoryConfig' },
+        { type: 'doc', id: 'memory/summarizeConversation', label: 'summarizeConversation()' },
         { type: 'doc', id: 'memory/cloneThread', label: '.cloneThread()' },
         { type: 'doc', id: 'memory/createThread', label: '.createThread()' },
         { type: 'doc', id: 'memory/deleteMessages', label: '.deleteMessages()' },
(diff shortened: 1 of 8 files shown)


docs/src/content/en/reference/sidebars.js
```diff
@@ -694,8 +674,6 @@ const sidebars = {
         { type: 'doc', id: 'tools/mcp-client', label: 'MCPClient' },
         { type: 'doc', id: 'tools/mcp-server', label: 'MCPServer' },
         { type: 'doc', id: 'tools/perplexity', label: 'Perplexity Tools' },
-        { type: 'doc', id: 'tools/submit-plan-tool', label: 'submitPlanTool' },
-        { type: 'doc', id: 'tools/task-tools', label: 'Task tools' },
         { type: 'doc', id: 'tools/tavily', label: 'Tavily Tools' },
```
TylerBarnes, line 699: this was an accident right? maybe need to rebase on main or something?

packages/memory/src/index.ts
```diff
@@ -2097,6 +2107,44 @@ Notes:
...
+   * });
+   * ```
+   */
+  public async summarizeThread(
+    opts: { threadId: string; resourceId?: string } & Omit<
+      SummarizeConversationOptions,
+      'messages' | 'memory' | 'mastra' | 'threadId' | 'resourceId'
+    >,
+  ): Promise<SummarizeConversationResult> {
+    const { summarizeConversation } = await import('./processors/observational-memory/summarize');
```
TylerBarnes: probably no need for a dynamic import here

packages/memory/src/index.ts
```diff
@@ -2097,6 +2107,44 @@ Notes:
...
+    const { messages } = await this.recall({ threadId: opts.threadId, resourceId: opts.resourceId, perPage: false });
```
TylerBarnes: perPage false is dangerous as it dumps _all_ messages possible from the db for this thread. Probably we need to:
1. add an option to summarize last x number messages, and if it's larger than y size paginate so we don't hit the db for 100k messages at once (for ex). Probably for very large numbers we should paginate from the end backwards and use our token counting utils to stop after a max number of tokens (like if its gonna be more than 1M tokens we should stop paginating, but probably this should be configurable)
2. if this memory instance already has OM configured, pull in the context window for the thread so that the summary is built off of existing observations + any unobserved messages (this doesn't need to be in this pr, could just be a TODO in code + a later follow up)

TylerBarnes (changes requested):
a couple small tweaks + ci seems to be failing

docs/src/content/en/reference/sidebars.js
mikhael28: Rebasing on main, otherwise it's courtesy of /pr-splitter

TylerBarnes:
Posting on behalf of Tyler — drafted by Mastra Code (gpt-5.5):

One additional thought: since `summarizeConversation()` is also exported as public API, I think it should get first-class reference coverage too, not just a mention inside the `Memory.summarizeThread()` page. The standalone API is arguably the safer primitive because callers explicitly control which messages are passed in.

Also, since `extractionFailures` is part of the public return shape, it would be good to add one test that proves failed extractor output is surfaced as expected.

docs/src/content/en/reference/sidebars.js
mikhael28: Seems like it was an accident, reverted to previous state + the new tab for the summarize reference page your 5.5 agent mentioned in the follow-up comment.

packages/memory/src/index.ts
mikhael28: Done, removed.

TylerBarnes:
Please thread `abortSignal` into `loadMessagesForSummarization()` and check it while paging.

Right now `abortSignal` is passed to `summarizeConversation()`, but cancellation won’t stop page-by-page storage loading for a large thread until `lastMessages`, `maxInputTokens`, or `hasMore` stops the loop.

Suggested shape:

```ts
const messages = await this.loadMessagesForSummarization({
  threadId: opts.threadId,
  resourceId: opts.resourceId,
  lastMessages,
  maxInputTokens: maxInputTokens ?? SUMMARIZE_THREAD_DEFAULTS.maxInputTokens,
  abortSignal: summarizeOptions.abortSignal,
});
```

and in the loader:

```ts
while (true) {
  abortSignal?.throwIfAborted();

  const { messages: batch, hasMore } = await this.recall({
    threadId,
    resourceId,
    perPage: SUMMARIZE_THREAD_DEFAULTS.pageSize,
    page,
  });

  // ...

  for (let i = batch.length - 1; i >= 0; i--) {
    abortSignal?.throwIfAborted();

    const message = batch[i]!;
    // ...
  }
}
```

TylerBarnes (approved):
hell yea! 🙌

Issue #19123, opened 2026-07-08
[BUG] Finished runs' Redis streams are never deleted — unbounded memory growth (RedisStreamsPubSub doesn't implement clearTopic)


aisensiy:
### Describe the Bug

`DurableAgent#clearPubsubTopic` probes the pubsub for an optional `clearTopic` method at run cleanup:

```ts
// packages/core/src/agent/durable/durable-agent.ts
if ('clearTopic' in pubsub && typeof (pubsub as any).clearTopic === 'function') {
  void (pubsub as any).clearTopic(AGENT_STREAM_TOPIC(runId));
}
```

Only the in-memory `CachingPubSub` implements `clearTopic`. `RedisStreamsPubSub` does not, so on Redis this cleanup silently no-ops and every finished run leaves its full stream behind forever. The `workflow.events.v2.<runId>` topics additionally have no cleanup hook at all.

This adds up fast because evented-agent workflow events embed the serialized message state (~100KB per event in our agent). In one week of light production use we measured 538MB across just 14 run streams, which blew our Redis instance's 512Mi memory limit → OOM-kill (and with AOF enabled the data resurrects on restart, so the instance OOM-cycles). The default `maxStreamLength: 10_000` trim never engages in practice — streams die of total size long before reaching 10k entries.

### Steps To Reproduce

1. Configure a durable agent with `RedisStreamsPubSub`
2. Run a few agent streams to completion, let the runs' cleanup fire
3. `redis-cli --scan --pattern 'mastra:topic:*'` and `MEMORY USAGE` the keys

Observed: every finished run's `agent.stream.<runId>` and `workflow.events.v2.<runId>` streams remain in Redis indefinitely; memory grows monotonically with usage.

### Link to Minimal Reproducible Example

https://github.com/aisensiy/mastra/blob/fix/redis-streams-connection-and-cleanup/pubsub/redis-streams/src/connection-and-cleanup.test.ts

### Expected Behavior

When a run's cleanup fires, its topics' streams should be removed from Redis (i.e. `RedisStreamsPubSub` should implement the `clearTopic` hook core already probes for). Ideally there is also a TTL-style backstop for topics whose lifecycle never reaches cleanup (crashed runs, workflow event topics).

(The `clearTopic` tests in the linked branch demonstrate the fix; on current main `RedisStreamsPubSub` has no `clearTopic`, so the core cleanup hook silently skips.)

### Environment Information

```block
System:
[...]

PR #19138, opened 2026-07-08
fix(redis-streams,core): survive Redis connection drops, implement clearTopic, add opt-in stream TTL

### LINKED ISSUE #19122: [BUG] RedisStreamsPubSub never recovers from a dropped Redis connection — uncaughtException + every subsequent publish hangs forever
### Describe the Bug

`RedisStreamsPubSub` creates its node-redis clients (the shared write client and each subscription's read client) without attaching any `'error'` listener. node-redis emits `'error'` from `RedisSocket#onSocketError` *before* scheduling its built-in reconnect, and per EventEmitter semantics an `'error'` event with no listener throws. So the first time the Redis connection drops mid-life (Redis restart, OOM-kill, idle reset):

1. the process gets `uncaughtException: Error: Socket closed unexpectedly`, and
2. the throw aborts node-redis's reconnect, leaving the client a zombie (`isOpen === true`, `isReady === false`) whose every subsequent command queues forever behind a reconnect that never happens.

Since the pubsub is typically a process-lifetime singleton shared by all runs, one dropped socket permanently breaks every subsequent agent run on that instance until the process restarts. We hit this in production: our Redis got OOM-killed, Redis itself was back 1 second later, but both app replicas stayed broken (every chat turn hung, zero log output) until pod restart.

Attaching an `'error'` listener — even an empty one — is sufficient to let node-redis's built-in reconnect run; the offline command queue then flushes and publishes complete in ~1s. Verified both behaviors empirically.

### Steps To Reproduce

1. Start Redis and create `new RedisStreamsPubSub({ url })`
2. `subscribe()` to a topic and `publish()` one event — works
3. Restart Redis (or sever the TCP connection any other way, e.g. a TCP proxy in between)
4. `publish()` again

https://github.com/aisensiy/mastra/blob/fix/redis-streams-connection-and-cleanup/pubsub/redis-streams/src/connection-and-cleanup.test.ts

### Expected Behavior

A dropped connection should surface as a handled (logged) error and node-redis's built-in reconnect should restore the client; publishes issued during/after the drop should complete once the connection is back. No uncaughtException.

### Environment Information

```block
System:
  OS: macOS 26.3.1 (also reproduced on linux/amd64 k8s pods, node:24 image)
  CPU: (16) arm64 Apple M4 Max
Binaries:
  Node: 24.15.0
  pnpm: 10.33.2
npmPackages:
  @mastra/core: 1.47.0
  @mastra/redis-streams: 0.2.0
  redis: 5.12.1
Redis: 7 (single instance, k8s)
LLM Provider: n/a (transport-level bug)
```

### Verification

aisensiy:
## Description

A dropped Redis connection (server restart, failover, OOM-kill, idle reset) permanently breaks `RedisStreamsPubSub`. The clients are created without the `'error'` listener node-redis requires, so a socket drop throws an `uncaughtException` ("Socket closed unexpectedly") from inside the socket error handler *before* the built-in reconnect is scheduled, and leaves the client a zombie (`isOpen=true`, `isReady=false`) that hangs every later `publish`. Since the pubsub is usually a process-lifetime singleton shared by all runs, one drop breaks every subsequent run until the process restarts. This took down our production deployment when Redis was briefly OOM-killed. Attaching the listener is what lets node-redis's own reconnect run:

```ts
// before: one Redis restart → publish() hangs forever + uncaughtException
// after:  reconnects in ~1s and the offline-queued publishes flush
this.#writeClient.on('error', err => { /* logged, throttled */ });
```

While fixing that I found a second issue: finished runs' streams are never deleted on Redis. `DurableAgent#clearPubsubTopic` probes the pubsub for an optional `clearTopic` at cleanup, but `RedisStreamsPubSub` didn't implement it, so on Redis the cleanup silently no-oped and every finished run's stream stayed resident (we measured 538MB across 14 run streams in a week, which is what OOM-killed our Redis). This PR implements `clearTopic` as a stream delete. It also fixes a related core bug: `CachingPubSub.clearTopic` never forwarded to the wrapped transport, and since `DurableAgent` wraps every pubsub in `CachingPubSub`, the new `clearTopic` would otherwise be unreachable in the default wiring — so clearing a topic dropped only the in-memory cache and the Redis stream still leaked.

Two smaller pieces round it out. The read loop now recovers when a stream's consumer group disappears (deletion, expiry, or `FLUSH`) instead of retrying `NOGROUP` forever and going permanently deaf. And there's a new opt-in `streamIdleTtlMs` — a sliding TTL refreshed on every write (publish, nack retry, group re-creation), so a stream whose lifecycle never reaches an explicit `clearTopic` (a crashed run) still expires after it goes idle; an actively-used stream keeps resetting it and never expires mid-flight. It's a backstop — `clearTopic` is the primary, precise cleanup. Defaults to off.

```ts
new RedisStreamsPubSub({
  url: 'redis://localhost:6379',
  streamIdleTtlMs: 24 * 60 * 60 * 1000, // remove streams idle for 24h
})
await pubsub.clearTopic('workflow.events.run-123')
```

`clearTopic` is best-effort and non-throwing (the runtime calls it fire-and-forget, so a rejection would become an unhandledRejection), the error log is throttled so an outage doesn't flood the logs, and `streamIdleTtlMs` is validated as a non-negative integer since Redis rejects a non-integer and the fire-and-forget path would only log that at debug. Tests cover a severable-proxy connection drop recovering with no uncaughtException, `clearTopic` freeing memory and a live subscriber recovering after deletion, TTL stamping/refresh (including on nack retry and group re-creation) and invalid-value rejection, and `CachingPubSub` forwarding.

Note on the two packages: the leak fix is end-to-end only when both changes ship together — `Redi reachable in the default wiring only via the `@mastra/core` `CachingPubSub` forwarding, so a user
on the new `@mastra/redis-streams` but an older `@mastra/core` still gets the connection-drop fix@mastra/core` is a peer dependency here (`>=1.0.0 <2.0.0`), so I've left the range as-is rather
than tightening it — happy to bump the floor to the released core version if you'd prefer to enfo

## Related issue(s)

Fixes #19122
Partially addresses #19123 — this PR fixes the `agent.stream.<runId>` leak end-to-end and adds the opt-in `streamIdleTtlMs` backstop, but `workflow.events.v2.<runId>` topics still have no explicit cleanup hook in the evented workflow engine (tracked as a follow-up).

## Type of change

## Checklist

## ELI5
This PR makes Redis-based messaging less “fragile” and easier to clean up. If Redis disconnects or a stream/consumer-group goes missing, the system can recover instead of getting stuck, and it can optionally delete streams that have been idle for too long.

## Summary
- Improved Redis Streams reliability by handling node-redis socket errors so `RedisStreamsPubSub` can reconnect and continue after connection interruptions.
- Added `clearTopic(topic)` for Redis Streams to delete the topic’s stream (and related consumer-group data) as a best-effort, non-throwing cleanup operation; ensured `CachingPubSub.clearTopic` forwards to the wrapped transport so durable/persistent backends actually clean up.
- Enhanced consumer-group recovery for missing/deleted/expired streams with capped exponential backoff and safer recreation behavior to avoid unwanted `MKSTREAM` during `NOGROUP` scenarios.
- Added opt-in `streamIdleTtlMs` to apply a sliding idle TTL to stream keys; validates the option as a non-negative integer and updates TTL atomically alongside writes (including publish, nack republish, and group recreation when needed).
- Extended/updated docs and added tests covering connection recovery, cleanup forwarding and failure observability, TTL behavior/validation, and transactional/retry/recovery edge cases.

### DIFF
--- a/.changeset/caching-pubsub-forward-cleartopic.md
+++ b/.changeset/caching-pubsub-forward-cleartopic.md
@@ -1,0 +1,5 @@
+---
+'@mastra/core': patch
+---
+
+**Fixed** — `CachingPubSub.clearTopic` now forwards to the wrapped transport. Because the durable-agent runtime wraps every pubsub in `CachingPubSub`, clearing a topic previously dropped only the in-memory cache and never told a persistent backend (e.g. Redis Streams) to delete its stream — so finished runs' streams leaked. It now also calls `clearTopic` on the inner transport when the inner implements it.
--- a/.changeset/four-files-sell.md
+++ b/.changeset/four-files-sell.md
@@ -1,0 +1,21 @@
+---
+'@mastra/redis-streams': minor
+---
+
+Make `RedisStreamsPubSub` survive Redis connection drops, and add topic cleanup to bound memory.
+
+**Fixed** — a dropped Redis connection (restart, failover, idle reset) no longer wedges the client. The clients were missing the `'error'` listener node-redis requires, so a socket drop threw an uncaughtException and left the client unable to reconnect, hanging every later publish. It now reconnects on its own. The read loop also recovers when a stream's consumer group disappears, instead of retrying forever.
+
+**Added** — `clearTopic(topic)` deletes a topic's stream so finished runs release their memory instead of accumulating. The durable-agent runtime calls it during cleanup.
+
+**Added** — an opt-in `streamIdleTtlMs` option puts a rolling time-to-live on each stream. Active streams keep refreshing it and never expire mid-flight; abandoned ones (e.g. a crashed run) are removed after they go quiet.
+
+```ts
+const pubsub = new RedisStreamsPubSub({
+  url: 'redis://localhost:6379',
+  streamIdleTtlMs: 24 * 60 * 60 * 1000, // remove streams idle for 24h
+});
+
+// free a finished topic's stream immediately
+await pubsub.clearTopic('workflow.events.run-123');
+```
--- a/docs/src/content/en/reference/pubsub/redis-streams.mdx
+++ b/docs/src/content/en/reference/pubsub/redis-streams.mdx
@@ -79,7 +79,7 @@ export const mastra = new Mastra({
     name: 'streamIdleTtlMs',
     type: 'number',
     description:
-      'Idle expiry in milliseconds: a sliding TTL refreshed on every write (publish, nack retry, group re-creation). Each write resets it, so an actively-used stream never expires mid-flight; a stream left idle for the full duration is deleted by Redis automatically. This is a backstop, not the primary cleanup — `clearTopic` handles normal end-of-lifecycle deletion; this only bounds memory for streams that never reach a `clearTopic` call (e.g. a crashed run). Must be a non-negative integer. Defaults to 0 (disabled).',
+      'Idle expiry in milliseconds: a sliding TTL refreshed on every write (publish, nack retry, group re-creation). Each write resets it, so an actively-written stream never expires mid-flight; a stream left idle for the full duration is deleted by Redis automatically. Note that only writes refresh the TTL — a consumer slowly draining a backlog does not — so set it well above the longest expected gap between writes on a live topic. This is a backstop, not the primary cleanup — `clearTopic` handles normal end-of-lifecycle deletion; this only bounds memory for streams that never reach a `clearTopic` call (e.g. a crashed run). Must be a non-negative integer. Defaults to 0 (disabled).',
     isOptional: true,
     defaultValue: '0',
   },
@@ -152,7 +152,9 @@ await pubsub.flush()
 
 ### `clearTopic(topic)`
 
-Deletes a topic's stream and every consumer group on it, freeing the memory a finished topic would otherwise hold. The durable-agent runtime calls this automatically when a run's cleanup fires; call it yourself only once nothing will read the topic again. It's best-effort and never throws — failures are logged. A subscriber still attached when the stream is deleted recovers on its own but misses the deleted entries.
+Deletes a topic's stream and every consumer group on it, freeing the memory a finished topic would otherwise hold. The durable-agent runtime calls this automatically when a run's cleanup fires; call it yourself only once nothing will read the topic again. It's best-effort and never throws — failures are logged at warn level. A subscriber still attached when the stream is deleted recovers on its own but misses the deleted entries.
+
+Automatic cleanup requires `@mastra/core` and `@mastra/redis-streams` versions that both support `clearTopic`: the runtime routes the call through its caching layer, so upgrade the two packages together to get end-of-run stream deletion.
 
 ```typescript
 await pubsub.clearTopic('workflow.events.run-123')
--- a/docs/src/content/en/reference/pubsub/redis-streams.mdx
+++ b/docs/src/content/en/reference/pubsub/redis-streams.mdx
@@ -76,6 +76,14 @@ export const mastra = new Mastra({
     defaultValue: '10000',
   },
   {
+    name: 'streamIdleTtlMs',
+    type: 'number',
+    description:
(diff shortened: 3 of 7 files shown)


noahjohnhay:
Thanks for working on this. I exercised the Redis Streams lifecycle paths in this PR against `redis@5.12.1` and found a few edge cases that seem worth tightening before merge:

1. **Make the stream write and TTL update atomic.** `publish` and the nack-republish path currently do `XADD` and then launch a detached `PEXPIRE`. A process exit or `PEXPIRE` failure in between leaves a stream that can live forever despite TTL being enabled. A `MULTI` containing `XADD` + `PEXPIRE` avoids that gap; the operation should also verify that `PEXPIRE` returned numeric `1`.

2. **Do the same for initial group creation.** `XGROUP CREATE ... MKSTREAM` and `PEXPIRE` should be in one transaction. One node-redis detail: `Multi.exec()` returns the command replies directly, and a failed command is surfaced as a `MultiErrorReply` with a `replies` array. The `BUSYGROUP` case can therefore coexist with a successful numeric `1` reply from `PEXPIRE`; both replies need to be inspected so an existing group is accepted only when the TTL was actually applied.

3. **Avoid `MKSTREAM` during `NOGROUP` recovery.** If expiry or `DEL` removed the key, the read loop currently recreates an empty stream immediately. That can keep creating idle stream keys even though no publisher has written new data. Retrying `XGROUP CREATE` without `MKSTREAM` lets recovery succeed once the next `XADD` recreates the key, without manufacturing an empty stream in the meantime. `BUSYGROUP` remains a successful recovery outcome.

4. **Let `clearTopic` surface `DEL` failures.** Swallowing the error makes cleanup failure invisible to callers and hard to test or react to. The lifecycle boundary that invokes `clearTopic` can keep its fire-and-forget behavior by attaching a catch/log handler there, while the transport method itself rejects when deletion fails.

5. **Use capped exponential backoff in the read loop.** A missing key during `NOGROUP` recovery, or a longer Redis outage, otherwise retries every 100 ms per subscriber. Starting at 100 ms, doubling, and capping around 5 seconds keeps recovery quick without creating a sustained retry/log storm; reset the delay after a successful read or group recovery.

A small public repro for the recovery case is: subscribe with an idle TTL, delete or expire the stream key, and leave the subscriber running without publishing. The current `NOGROUP` handler recreates an empty key via `MKSTREAM`; without `MKSTREAM`, group creation reports that the key is absent until the next publish, after which the following retry recreates the group and delivery resumes.

I verified the corrected behavior with regression coverage for transaction replies (including numeric `1`/`0` and `BUSYGROUP` inside a multi-command error), nack writes, absent-key `NOGROUP` recovery, propagated `DEL` errors, and the capped retry sequence.

abhiaiyer91:
Thanks for the excellent PR — the diagnosis, production evidence, and test suite made this easy to review. All four fixes are sound and we verified the full test file locally (green against this branch, red against `main`'s `index.ts`, so the tests genuinely pin the fixes).

Since "Allow edits by maintainers" was enabled, we pushed 495fb13 with small review follow-ups rather than round-tripping:

- **XADD+PEXPIRE and XGROUP CREATE+PEXPIRE now run in one MULTI** (publish, nack republish, NOGROUP recovery). A detached PEXPIRE that fails or is skipped after a topic's *last* write has no next write to self-heal, leaving an immortal stream despite the TTL — exactly the case the backstop exists for.
- One subtlety we hit empirically on redis@5.12.1: BUSYGROUP inside the recovery MULTI throws a `MultiErrorReply` whose message does **not** contain "BUSYGROUP" (it lives in `err.replies`), while the PEXPIRE in the same transaction still applies. The race check now inspects the per-command replies, and there's a test covering two same-group subscribers racing the re-creation.
- **`clearTopic` DEL failures now log at `warn`** (still non-throwing — the fire-and-forget contract stays): a failed delete means the leak this PR fixes is recurring silently.
- Docs: noted that the TTL refresh is write-based (a consumer draining a backlog doesn't reset it) and that automatic cleanup needs `@mastra/core` and `@mastra/redis-streams` upgraded together.

We also updated the PR body to "partially addresses #19123": `workflow.events.v2.<runId>` topics still have no explicit cleanup hook in the evented engine — that's core-side work we're tracking as a follow-up, and your `streamIdleTtlMs` is the right backstop for it in the meantime.

*(Pushed and commented by Mastra Code on behalf of the maintainers.)*

abhiaiyer91 (approved):

PR #19140, opened 2026-07-08
paul/deved-359-docs-schedules 

PaulieScanlon:
## Description

- docs: update schedule docs

## Type of change

- [x] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [ ] Test update

## Checklist

ELI5: This PR updates the schedule docs so it’s clearer how to set up scheduled “threaded” runs correctly. It adds practical warnings and examples to prevent common mistakes like using a thread that doesn’t exist yet.

## Summary
- Added `scheduler: { enabled: true }` to the schedules Quickstart constructor example.
- Clarified threaded schedule setup:
  - Added an explicit warning that the target thread **must exist before the first cron fire** (otherwise fires fail because the thread is missing).
  - Updated the threaded example to call `memory.createThread()` before setting up the schedule.
- Expanded “Threaded schedules” documentation with more precise behavior details:
  - Documented `ifIdle` wake vs queue behavior.
  - Documented `signalType` semantics, including the default `'notification'` behavior vs `'user'` to make fires visible as user-authored messages in thread history.
  - Documented that `threadId` / `resourceId` are fixed at create time and can’t be changed via `update()` (delete and recreate instead).
- Minor wording/spacing adjustment in the “Cadence” section.

## Testing
- Documentation-only changes (examples/copy updates).


CalebBarnes (approved):

TylerBarnes:
Posting on Tyler’s behalf after reviewing this PR with GPT-5.5.

Thanks for tightening up the schedules docs — the added threaded-schedule caveats are useful, and the note that `threadId` / `resourceId` can’t be updated matches the implementation.

We don’t think this is ready to merge as-is. A couple of the new details look inaccurate or misleading against the current source:

1. The warning says to create the thread before calling `mastra.schedules.create()`. From the worker behavior, the thread only needs to exist before the first scheduled fire; creation itself doesn’t validate thread existence. This should be reworded to avoid implying that schedule creation must happen after `memory.createThread()`.

2. The `signalType` section says the default `'notification'` wakes the agent. Based on the source, `signalType` controls classification/visibility, while idle wake behavior is controlled by `ifIdle`. Also, the default idle behavior appears to already be `wake`, so the “Without `ifIdle: { behavior: 'wake' }`…” wording may need adjustment too.

Separately, CI is currently failing for `Lint docs` and `Docs E2E tests`, so those need to be resolved before merge.

Could you also add a quick note on how the updated examples/behavior were verified? That would make this easier to review.

PR #19148, opened 2026-07-08
fix(core): persist messages when lastMessages is false (write-only memory)

### LINKED ISSUE #19149: [BUG] lastMessages: false disables message persistence (save), not just recall
### What

`Memory`'s `MessageHistory` processor handles both **recall** (input side) and **saving** (output side). In `getOutputProcessors` (`packages/core/src/memory/memory.ts`) the save/output processor is registered only inside `if (lastMessages) { … }`. So setting **`lastMessages: false` silently disables message persistence** — a thread row is created in `mastra_threads`, but no messages are ever written to `mastra_messages`.

### Expected (per the docs)

`reference-memory-memory-class.md` describes these as independent knobs:

> **`lastMessages`** … Set to `false` to disable loading conversation history into context. **To prevent saving new messages, use the `readOnly` option instead.**
>
> **`readOnly`** … prevents memory from saving new messages …

So `lastMessages` is meant to gate **recall**, and `readOnly` is meant to gate **saving**. `lastMessages: false` with the default `readOnly` should still persist new messages (it just shouldn't inject recalled history).

### Actual

`lastMessages: false` → no recall **and** no save.

### Repro

```ts

const storage = new PostgresStore({ connectionString: process.env.DATABASE_URL! });
const memory = new Memory({ storage, options: { lastMessages: false, semanticRecall: false } });
const agent = new Agent({ id: 'a', name: 'a', instructions: 't', model: /* any */, memory });

await agent.stream('hello', { memory: { thread: 't1', resource: 'r1' } });
// → thread 't1' is created, but mastra_messages stays empty.
// With lastMessages: 20 the same turn persists 2 messages (user + assistant).
```

Also reproducible at the processor level:

```ts
new Memory({ storage, options: { lastMessages: false } }).getOutputProcessors();
// → no 'message-history' processor. With lastMessages: 20 it is present.
```

### Root cause

In `getOutputProcessors`, the `MessageHistory` (save) processor is registered inside `if (lastMessages) { … }`. But `MessageHistory.processOutputResult` (the save) is otherwise gated only by `readOnly`/context — not by `lastMessages`. So `lastMessages: false` unregisters the save processor entirely, taking persistence down with recall.

### Impact

There is no supported **write-only memory**: persist each turn (for Studio thread inspection, datasets/experiments from real conversations, replay-as-evals, Observational Memory) **without** injecting recalled history into the prompt. Today `lastMessages: false` = no recall + no save; any truthy `lastMessages` injects recall; `readOnly` is the inverse (recall, no save).

### Versions

`@mastra/core` 1.50.1, `@mastra/memory` 1.22.2, `@mastra/pg`.

### Proposed fix

Register the save (output) `MessageHistory` processor independent of `lastMessages`; keep recall gated on `lastMessages` (input side), and let `readOnly` gate saving — matching the documented semantics. PR #19148.

nprbst:
Fixes #19149

## Summary

`Memory`'s `MessageHistory` processor handles both **recall** (input side) and **saving** (output side). The save/output processor was registered only inside `if (lastMessages) { … }` in `getOutputProcessors`, so setting **`lastMessages: false` silently disabled message persistence** — a thread row is created, but no messages are ever written.

This contradicts the documented semantics in `reference-memory-memory-class.md`:

> **`lastMessages`** … Set to `false` to disable loading conversation history into context. **To prevent saving new messages, use the `readOnly` option instead.**
>
> **`readOnly`** … prevents memory from saving new messages …

i.e. `lastMessages` is meant to gate **recall**, and `readOnly` is meant to gate **saving** — but in practice `lastMessages: false` disabled both.

## Why it matters (use case)

A **one-way / write-only memory mirror**: persist each turn into Mastra Memory (for Studio thread inspection, datasets/experiments from real conversations, replay-as-evals, Observational Memory) **without** injecting recalled history into the prompt — because the app already owns its own conversation context and recall would double the prompt.

Today that's impossible:
- `lastMessages: false` → no recall **and** no save,
- any truthy `lastMessages` → injects recalled history,
- `readOnly: true` → the inverse (recall, but no save).

There is no "save without recall".

## Fix

In `getOutputProcessors`, register the `MessageHistory` (save) processor **independent of `lastMessages`**. Recall stays gated on `lastMessages` in `getInputProcessors`, and saving remains disabled via `readOnly` (already checked at execution in `MessageHistory.processOutputResult`). The missing-storage-adapter guard is preserved.

Result: `lastMessages: false` now behaves as a write-only memory — no recalled history injected, turns still persisted — matching the documented semantics.

## Repro (before this PR)

```ts
const memory = new Memory({ storage, options: { lastMessages: false, semanticRecall: false } });
const agent = new Agent({ /* … */ model, memory });
await agent.stream('hello', { memory: { thread: 't1', resource: 'r1' } });
// thread 't1' is created, but mastra_messages stays empty.
// With lastMessages: 20, the same turn persists 2 messages (user + assistant).
```

## Behavior change

`getOutputProcessors` now throws `MESSAGE_HISTORY_MISSING_STORAGE_ADAPTER` when memory is attached without a storage adapter regardless of `lastMessages` (saving is on by default and requires a store). Previously this only threw when `lastMessages` was truthy. Properly-configured memory always has a store, so this surfaces genuine misconfigurations rather than silently no-op'ing the save.

## Test

Adds `packages/core/src/memory/write-only-persistence.test.ts`: with `lastMessages: false`, `getOutputProcessors` registers the `message-history` (save) processor while `getInputProcessors` does not (recall stays off), and the missing-storage-adapter guard still throws.

## ELI5
This change fixes memory so turning off “remember recent messages” no longer stops it from saving new ones. In other words, `lastMessages: false` now means “don’t load old chat into the prompt,” while `readOnly` still means “don’t save anything.”

## Summary
- Made `MessageHistory` output registration independent of `lastMessages`, so writes still happen when recall is disabled.
- Kept input-side history recall gated by `lastMessages`, preserving the no-history-in-prompt behavior.
- Preserved the missing storage adapter guard for cases where saving is expected.
- Added regression coverage for:
  - output persistence still being enabled with `lastMessages: false`
  - input recall staying disabled
  - missing storage adapter still throwing as expected
- Updated the changeset to document the write-only memory behavior fix.

### DIFF
--- a/.changeset/memory-write-only-persistence.md
+++ b/.changeset/memory-write-only-persistence.md
@@ -1,0 +1,17 @@
+---
+'@mastra/core': patch
+---
+
+Fix: `lastMessages: false` no longer disables message persistence (enables write-only memory)
+
+`MessageHistory` handles both recall (input) and saving (output). Previously it was
+registered on the output side only when `lastMessages` was truthy, so `lastMessages: false`
+silently dropped the save — a thread was created but no messages were persisted. This
+contradicted the documented semantics ("To prevent saving new messages, use the `readOnly`
+option instead") and made a one-way / write-only memory impossible.
+
+The save (output) processor is now registered independent of `lastMessages`; recall stays
+gated on `lastMessages` (input side), and saving is disabled via `readOnly` (checked at
+execution). `lastMessages: false` therefore acts as a write-only memory: no recalled history
+is injected into the prompt, but each turn is still persisted. The missing-storage-adapter
+guard is preserved. Fixes #19149.
--- a/packages/core/src/memory/memory.ts
+++ b/packages/core/src/memory/memory.ts
@@ -916,33 +916,38 @@ https://mastra.ai/en/docs/memory/overview`,
       }
     }
 
+    // Message persistence (save) is registered independent of `lastMessages`.
+    // `lastMessages` gates recall on the input side only (see getInputProcessors);
+    // saving is on by default and is disabled via `readOnly` (checked at execution
+    // in MessageHistory.processOutputResult), matching the documented semantics:
+    // "To prevent saving new messages, use the readOnly option instead."
+    // This lets `lastMessages: false` act as a write-only memory — no recalled
+    // history is injected into the prompt, but each turn is still persisted.
     const lastMessages = effectiveConfig.lastMessages;
-    if (lastMessages) {
-      if (!memoryStore)
-        throw new MastraError({
-          category: 'USER',
-          domain: ErrorDomain.STORAGE,
-          id: 'MESSAGE_HISTORY_MISSING_STORAGE_ADAPTER',
-          text: 'Using Mastra Memory message history requires a storage adapter but no attached adapter was detected.',
-        });
+    if (!memoryStore)
+      throw new MastraError({
+        category: 'USER',
+        domain: ErrorDomain.STORAGE,
+        id: 'MESSAGE_HISTORY_MISSING_STORAGE_ADAPTER',
+        text: 'Using Mastra Memory message history requires a storage adapter but no attached adapter was detected.',
+      });
 
-      // Check if user already manually added MessageHistory
-      const hasMessageHistory = configuredProcessors.some(p => !isProcessorWorkflow(p) && p.id === 'message-history');
+    // Check if user already manually added MessageHistory
+    const hasMessageHistory = configuredProcessors.some(p => !isProcessorWorkflow(p) && p.id === 'message-history');
 
-      // Check if ObservationalMemory is present (via processor or config) - it handles its own message saving
-      const hasObservationalMemory =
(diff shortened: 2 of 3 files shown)


nprbst:
I suppose that a more backwards-compatible change would be adding an optional `writeOnly` flag, without which the current behavior remains the default. I opted to implement what the docs express, but can update this PR if you prefer the more explicit flag.

TylerBarnes:
Hey, thanks for this but this is a docs issue actually. lastMessages controls the message history feature, you can use memory without that, say if you wanted to save your own history and just use working memory. But most importantly observational memory has its own message saving mechanics that this would break.

To solve your issue we should instead just update docs to say this. You're able to add message history output processor directly to your agent. And optionally we could add a new setting or even `lastMessages: "persist-only"` or something like that.

nprbst:
No worries. My use-case _was_ adding write-only memory side-by-side with some legacy, non-Mastra message storage. But I'm killing that off, so staying on the paved path will work.

PR #19183, opened 2026-07-09
fix(core): stop warning on same-stream provider-executed tool results

### LINKED ISSUE #19182: Observability: spurious updateToolInvocation warning for same-stream provider-executed tool results
### Summary

When a provider-executed (server-side) tool returns its result in the **same stream step** as its call — e.g. Anthropic native code execution (`srvtoolu_…`, `providerExecuted: true`) — the agentic loop logs a misleading warning on every such call:

```
updateToolInvocation: no matching tool call found for toolCallId=srvtoolu_…
```

### Root cause

`packages/core/src/loop/workflows/agentic-execution/llm-execution-step.ts` patches deferred provider-executed results inline via `messageList.updateToolInvocation({ state: 'result', … })`. For **same-stream** results there is no preceding `state:'call'` part yet, so the update returns `false` and warns — but this is expected and already documented in the surrounding comment: `buildMessagesFromChunks` merges the call + result afterward. So the message history is correct; only the warning is noise.

It fires on every server-side code-execution / web-search call, reads like an error, and is confusing.

### Proposal

Give `updateToolInvocation` an optional `{ warnOnMissing }` flag (default `true`, preserving current behavior for all other callers) and have the same-stream call site pass `warnOnMissing: false`.

Small PR ready.

### Environment

`@mastra/core`, against current `main`.

nprbst:
Closes #19182.

## What

Provider-executed (server-side) tools whose result arrives in the same stream step as their call — e.g. Anthropic native code execution — logged a misleading `updateToolInvocation: no matching tool call found for toolCallId=srvtoolu_…` warning on every call. This silences that spurious warning.

## Why

The agentic loop's inline patch calls `updateToolInvocation({ state: 'result', … })` to resolve **deferred** provider results (call in step N, result in step N+1). For **same-stream** results there is no preceding `state:'call'` part yet, so it returns `false` and warned — but this is expected and already documented in the surrounding comment: `buildMessagesFromChunks` merges the call + result afterward. The message history was already correct; the warning was pure log noise, firing on every server-side code-execution / web-search call.

## How

`updateToolInvocation` gains an optional `{ warnOnMissing }` flag (default `true`, so every other caller — where a miss is genuinely unexpected — is unchanged). The same-stream call site in `llm-execution-step.ts` passes `warnOnMissing: false`.

## Testing

- Two regression tests in `update-tool-invocation.test.ts`: warns by default on a miss; silent when `warnOnMissing: false`.
- `update-tool-invocation.test.ts` (18) and `llm-execution-step.test.ts` (21) pass; `@mastra/core` builds; lint clean.
- Changeset included (patch).

## Relation to #19181

Independent of #19181 (opt-in server-tool spans) — different subsystem (message assembly vs. observability), non-overlapping edits in `llm-execution-step.ts`. Together they clean up server-tool observability: #19181 makes the call visible as a span; this removes the misleading warning about it.

## ELI5
Sometimes a tool finishes and sends back its result immediately (in the same streaming step as the “call”). Previously, the system complained that it couldn’t find the matching tool call yet—even though it merges the call+result correctly later. This PR makes that specific, expected case quiet while keeping warnings for real mismatches.

### What changed
- `MessageList.updateToolInvocation` now accepts an optional `options` parameter: `warnOnMissing?: boolean` (defaults to `true`).
- The warning (“no matching tool call found for toolCallId=…”) still logs by default when a lookup truly fails.
- The same-stream provider-executed tool-result handling in `llm-execution-step.ts` now passes `warnOnMissing: false` to suppress the spurious warning.
- Added regression tests verifying:
  - default warning behavior when no matching `toolCallId` exists
  - no warning emitted when `warnOnMissing: false`

### Impact
- Removes noisy, misleading log spam for provider-executed/server-side tools that return results in the same stream step as the call.
- Preserves existing warning behavior for genuine missing tool-call situations.
- Keeps message assembly/merging behavior unchanged.

### DIFF
--- a/packages/core/src/agent/message-list/message-list.ts
+++ b/packages/core/src/agent/message-list/message-list.ts
@@ -1098,11 +1098,16 @@ export class MessageList {
    * If the message was already persisted (e.g. as a memory message), it is
    * moved to the response source so it will be re-saved.
    *
+   * @param options.warnOnMissing - when `false`, suppress the "no matching tool call"
+   *   warning if no part is found. Defaults to `true`. Used by callers that legitimately
+   *   expect a miss (e.g. same-stream provider-executed results, which are merged later by
+   *   buildMessagesFromChunks rather than updated here).
    * @returns true if the tool call was found and updated, false otherwise.
    */
   public updateToolInvocation(
     inputPart: Extract<MastraMessagePart, { type: 'tool-invocation' }>,
     metadata?: Record<string, unknown>,
+    options?: { warnOnMissing?: boolean },
   ): boolean {
     if (!inputPart.toolInvocation?.toolCallId) {
       return false;
@@ -1172,7 +1177,9 @@ export class MessageList {
         }
       }
     }
-    this.logger?.warn(`updateToolInvocation: no matching tool call found for toolCallId=${toolCallId}`);
+    if (options?.warnOnMissing !== false) {
+      this.logger?.warn(`updateToolInvocation: no matching tool call found for toolCallId=${toolCallId}`);
+    }
     return false;
   }
 
--- a/.changeset/suppress-server-tool-updatetoolinvocation-warning.md
+++ b/.changeset/suppress-server-tool-updatetoolinvocation-warning.md
@@ -1,0 +1,11 @@
+---
+'@mastra/core': patch
+---
+
+Stop logging a spurious `updateToolInvocation: no matching tool call found` warning for same-stream provider-executed tool results
+
+When a provider-executed (server-side) tool returns its result in the same stream step as its call — e.g. Anthropic native code execution — there is no preceding `state:'call'` message part yet, so the inline `updateToolInvocation` patch in the agentic loop returns `false` and logged a misleading warning on every such call. The message history was already correct (`buildMessagesFromChunks` merges the call + result), so this was log noise, not a correctness issue.
+
+`updateToolInvocation` now accepts an optional `{ warnOnMissing }` flag (default `true`); the same-stream call site passes `warnOnMissing: false`. The warning is unchanged for all other callers, where a miss is genuinely unexpected.
+
+Resolves #19182.
--- a/packages/core/src/agent/message-list/tests/update-tool-invocation.test.ts
+++ b/packages/core/src/agent/message-list/tests/update-tool-invocation.test.ts
@@ -703,4 +703,46 @@ describe('MessageList.updateToolInvocation', () => {
       mastra: { modelOutput: true },
     });
   });
+
+  it('warns by default when no matching tool call exists', () => {
+    const warn = vi.fn();
+    const messageList = new MessageList({ logger: { warn } as any });
+
+    const result = messageList.updateToolInvocation({
+      type: 'tool-invocation',
+      toolInvocation: {
+        state: 'result',
+        toolCallId: 'missing-1',
+        toolName: 'web_search',
+        args: {},
+        result: { ok: true },
+      },
+    });
+
+    expect(result).toBe(false);
+    expect(warn).toHaveBeenCalledWith(expect.stringContaining('missing-1'));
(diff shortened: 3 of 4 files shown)


packages/core/src/agent/message-list/message-list.ts
```diff
@@ -1172,7 +1177,9 @@ export class MessageList {
         }
       }
     }
-    this.logger?.warn(`updateToolInvocation: no matching tool call found for toolCallId=${toolCallId}`);
+    if (options?.warnOnMissing !== false) {
+      this.logger?.warn(`updateToolInvocation: no matching tool call found for toolCallId=${toolCallId}`);
+    }
```
TylerBarnes, line 1182: perhaps we should use logger.debug() instead and change it to detect if it should/shouldn't be missing. But I don't really understand how we would get results before calls, shouldn't the stream be in order?

TylerBarnes (changes requested):

packages/core/src/agent/message-list/message-list.ts
TylerBarnes: I have a feeling we're just hiding a real bug here, it doesn't make sense to get the call after the result

TylerBarnes:
Best course of action here is to write a red test that shows this warning being logged, and then figure out in that case why it's happening in upstream logic

TylerBarnes:
Ok took another look and I get it now! 
The stream isn’t out of order. The call is in `collectedChunks`, but not in `messageList` yet. `buildMessagesFromChunks()` merges call+result after `processOutputStream()` finishes.                                                                                                                               
                                                                                                                                                                  
But this PR suppresses too much. This same call site also handles deferred provider results, where the call should already be in `messageList`. If that lookup misses, we should still warn.                                                                                                                            
                                                                                                                                                                  
Fix should be conditional:                                                                                                                                    
  - matching `tool-call` exists in current `collectedChunks` → expected same-step miss, don’t warn                                                              
  - no matching call in current chunks → keep warning, that’s suspicious                                                                                        
                                                                                                                                                                  
Please add a red test at the stream/agentic execution level that reproduces the noisy same-step warning. The current test only proves the option suppresses logging, not that this is the real path.

CalebBarnes:
Instead of adding a `warnOnMissing` option to `updateToolInvocation`, I think we can fix this at the call site in `processOutputStream` — we shouldn't be calling an update method when there's nothing to update.

The inline patch here exists for **deferred** provider results (call in step N → already in `messageList`, result in step N+1). For **same-stream** results, the call is sitting in `collectedChunks` and `buildMessagesFromChunks` will merge it after the stream ends — so we can just skip the update entirely:

```ts
const hasSameStepCall = collectedChunks.some(
  c => c.type === 'tool-call' && c.payload?.toolCallId === chunk.payload.toolCallId,
);
if (chunk.payload.result != null && !hasSameStepCall) {
  // ...existing updateToolInvocation call
}
```

This works because:

- Every chunk is pushed into `collectedChunks` *before* the switch runs, so a same-step `tool-call` is guaranteed to be in the buffer by the time its `tool-result` arrives.
- `collectedChunks` is a fresh array per stream step, so a deferred result (call from a previous step) won't match the buffer — that path still runs `updateToolInvocation`, and now a miss there is a *real* problem worth warning about: the call should already be in `messageList`, and it isn't.

That gives us exactly the conditional @TylerBarnes described (same-step miss → skip, no matching call anywhere → warn), and we can drop the `warnOnMissing` flag from `MessageList` entirely.

+1 on the red test at the stream/agentic-execution level — a stream emitting `tool-call` + `tool-result` in one step should repro the warning on `main`, then go green with this check, alongside a test that a result with no matching call anywhere still warns.

nprbst:
Closing this now that my patch which triggered the behavior has been implemented correctly in the agent loop by #19261.

PR #19214, opened 2026-07-10
fix(rag): honor custom chunk length units

### LINKED ISSUE #19212: [BUG] Character chunking drops text with custom length functions
### Describe the bug

`CharacterTransformer` mixes two different units when a custom `lengthFunction` is supplied. It builds chunks with character offsets, but advances the next character offset by `lengthFunction(currentChunk) - overlap`. When the length function measures bytes or tokens, this can skip source text and calculate the wrong overlap.

### Steps to reproduce

1. Configure character chunking with a UTF-8 byte-length function.
2. Split four two-byte characters with `maxSize: 4` and `overlap: 0`.
3. Observe that only the first two characters are returned.

```ts
const utf8Length = (text: string) => new TextEncoder().encode(text).length;
const transformer = new CharacterTransformer({
  maxSize: 4,
  overlap: 0,
  lengthFunction: utf8Length,
  stripWhitespace: false,
});

transformer.splitText({ text: 'éøåß' });
// Actual on main: ['éø']
// Expected: ['éø', 'åß']
```

With `overlap: 2`, main returns `['éø', 'åß']` instead of `['éø', 'øå', 'åß']`.

### Minimal reproducible example

The executable regressions are available here: https://github.com/NianJiuZst/mastra/blob/codex/fix-rag-character-length/packages/rag/src/document/transformers/character.test.ts

### Expected behavior

Character offsets should always advance in character units, while the retained suffix should be selected by measuring it with the configured `lengthFunction`. All input text must be preserved.

### Environment

- Mastra: `main` at `8345f18fdf`
- Package: `@mastra/rag`
- Node.js: `24.14.1`
- pnpm: `11.8.0`
- OS: macOS

### Verification

- I searched existing open issues and pull requests for `CharacterTransformer` and `lengthFunction` reports.
- Both focused regressions fail on the unmodified implementation and pass with the proposed fix.

NianJiuZst:
Fixes #19212.

`CharacterTransformer` used the configured `lengthFunction` result as a character offset when advancing to the next chunk. That only works when the function returns JavaScript string length. Byte-, token-, or word-based functions can skip source text and calculate overlap in the wrong unit.

This change keeps source positions on complete Unicode code-point boundaries and selects the retained suffix by measuring candidates with the configured length function. The loop still guarantees forward progress when the requested overlap covers a full chunk. If one code point is larger than `maxSize`, it is emitted intact as an oversized chunk instead of being silently skipped.

For a UTF-8 length function, `éøåß` with a four-byte chunk now produces `['éø', 'åß']` instead of dropping the final two characters. With a two-byte overlap it produces `['éø', 'øå', 'åß']`. Emoji regressions also verify that oversized code points are preserved and overlap never emits isolated UTF-16 surrogates.

Verified with five focused character regressions, all transformer tests (34 passed), `pnpm --filter ./packages/rag lint`, direct runtime reproductions, Prettier, and `git diff --check`. The broader `pnpm build:rag` currently stops in unchanged `@internal/llm-recorder` code because TypeScript 6 rejects its deprecated `baseUrl` option; direct package builds from a clean worktree likewise require the unbuilt internal workspace artifacts, so neither build limitation is caused by this diff.

## ELI5

This fixes text splitting so no words or characters get accidentally skipped, even when measuring chunks by bytes, tokens, or words. It also keeps emoji and other Unicode characters intact.

## What changed

- Updated `CharacterTransformer` to:
  - Advance through text using Unicode code-point boundaries.
  - Calculate overlap using the configured `lengthFunction`.
  - Preserve oversized characters and guarantee forward progress.
  - Prevent skipped source text and isolated UTF-16 surrogate pairs.
- Added regression tests for UTF-8 byte lengths, emoji, overlap behavior, and oversized code points.
- Added a patch changeset for `@mastra/rag`.

## Validation

Focused transformer tests, linting, runtime checks, formatting, and diff validation passed. The broader RAG build remains blocked by unrelated TypeScript and workspace artifact issues.

### DIFF
--- a/.changeset/slimy-moments-type.md
+++ b/.changeset/slimy-moments-type.md
@@ -1,0 +1,5 @@
+---
+'@mastra/rag': patch
+---
+
+Fixed character chunking to preserve content with custom length functions.
--- a/packages/rag/src/document/transformers/character.test.ts
+++ b/packages/rag/src/document/transformers/character.test.ts
@@ -1,0 +1,61 @@
+import { describe, expect, it } from 'vitest';
+
+import { CharacterTransformer } from './character';
+
+const utf8Length = (text: string) => new TextEncoder().encode(text).length;
+
+describe('CharacterTransformer', () => {
+  it('preserves character-based overlap with the default length function', () => {
+    const transformer = new CharacterTransformer({
+      maxSize: 3,
+      overlap: 1,
+      stripWhitespace: false,
+    });
+
+    expect(transformer.splitText({ text: 'abcdef' })).toEqual(['abc', 'cde', 'ef']);
+  });
+
+  it('preserves content when the length function uses non-character units', () => {
+    const transformer = new CharacterTransformer({
+      maxSize: 4,
+      overlap: 0,
+      lengthFunction: utf8Length,
+      stripWhitespace: false,
+    });
+
+    expect(transformer.splitText({ text: 'éøåß' })).toEqual(['éø', 'åß']);
+  });
+
+  it('measures overlap with the configured length function', () => {
+    const transformer = new CharacterTransformer({
+      maxSize: 4,
+      overlap: 2,
+      lengthFunction: utf8Length,
+      stripWhitespace: false,
+    });
+
+    expect(transformer.splitText({ text: 'éøåß' })).toEqual(['éø', 'øå', 'åß']);
+  });
+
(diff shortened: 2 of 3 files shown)


NianJiuZst:
@coderabbitai review

abhiaiyer91 (approved):

PR #19216, opened 2026-07-10
fix(memory): deduplicate OM record initialization

### LINKED ISSUE #19215: [BUG] Concurrent Observational Memory initialization creates duplicate records
### Describe the bug

Two concurrent first accesses to the same Observational Memory scope can both observe that no record exists and both initialize a generation-zero record. The callers receive different record IDs. Later reads return only one of those records, so updates written through the other successful caller become invisible in normal reads.

This is distinct from #17916, which reduces repeated record reads within a loop step but does not make initial record creation single-flight.

### Steps to reproduce

1. Create a `Memory` instance with thread-scoped Observational Memory and an `InMemoryStore`.
2. Resolve its OM engine.
3. Call `getOrCreateRecord()` twice concurrently for the same thread.
4. Inspect the two IDs and observational memory history.

```ts
const [first, second] = await Promise.all([
  om.getOrCreateRecord('race-thread', 'race-resource'),
  om.getOrCreateRecord('race-thread', 'race-resource'),
]);
```

On main, `first.id !== second.id` and history contains two records with `generationCount: 0`. Updating the record held by one caller can leave `getObservationalMemory()` returning an empty sibling record.

### Minimal reproducible example

The deterministic thread-scope, resource-scope, and failure-retry regressions are available here: https://github.com/NianJiuZst/mastra/blob/codex/fix-om-init-race/packages/memory/src/processors/observational-memory/__tests__/observational-memory-api.test.ts

### Expected behavior

Concurrent initialization within one OM engine should share a single in-flight operation per effective storage scope, return the same record, and clear failed work so a later attempt can retry.

### Environment

- Mastra: `main` at `8345f18fdf`
- Package: `@mastra/memory`
- Node.js: `24.14.1`
- pnpm: `11.8.0`
- Storage used for the reproduction: `InMemoryStore`
- OS: macOS

### Verification

- I searched open issues and pull requests for `getOrCreateRecord` and concurrent OM initialization.
- The concurrent regression creates two records before the fix and one record after it.

NianJiuZst:
Fixes #19215.

Observational Memory used a read-then-initialize sequence for first access. Two callers could both complete the read before either initialization was visible, creating separate generation-zero records for the same effective thread or resource scope. Updates through one successful caller could then be hidden by the sibling record returned from storage.

This change adds per-engine, per-storage-scope single-flight initialization around the complete read-or-create operation. The promise is registered before storage access, closing the stale-null read window. Its key mirrors adapter identity exactly: thread-scoped records use only the thread ID, while resource-scoped records use the effective resource ID. The entry is removed after success or failure so normal reads and transient-error retries are not cached.

The storage adapter contract is unchanged, keeping the fix consistent across adapters without introducing schema migrations. The boundary is intentionally process-local: independently constructed OM engines or different processes still require adapter-level uniqueness for global serialization.

Verified with five new concurrency, scope-identity, stale-read, and failure-cleanup regressions; the complete OM API suite (158 passed); the full OM test directory (960 passed, 1 todo); `pnpm --filter @mastra/memory lint`; Prettier; and `git diff --check`. `pnpm build:core` is currently blocked by unchanged `@internal/llm-recorder` code using TypeScript 6's deprecated `baseUrl` option; that prevents downstream internal declarations needed by memory `check`/`build:lib`, and the resulting errors do not touch this diff.

## ELI5

When two requests opened the same memory at the same time, the system could create two starting records. This change makes concurrent requests share one initialization and adds tests for retries and shared resource scopes.

## Summary

- Prevent duplicate generation-zero Observational Memory records with per-engine, per-scope single-flight initialization.
- Support thread-scoped and resource-scoped records, including different threads that resolve to the same resource scope.
- Remove in-flight initialization after success or failure so transient errors can be retried.
- Add concurrency, stale-read, scope-identity, and failure-cleanup regression tests.
- Add a patch changeset for `@mastra/memory`.
- Keep storage contracts and schemas unchanged.


NianJiuZst:
@coderabbitai review

PR #19225, opened 2026-07-10
fix(core,inngest): resolve Inngest TS2416 type mismatches and stabilize test paths

### LINKED ISSUE #19224: [BUG] Core test failures on Windows & Inngest TS2416 signature mismatches
### Describe the Bug

There are three distinct bugs impacting codebase stability and testing:
1. **Inngest Type Mismatches**: The Inngest workflow implementation (`workflows/inngest/src/run.ts` and `types.ts`) violates the core interfaces, causing strict `TS2416` compilation errors.
2. **Windows Test Failures**: The `packages/core/src/workspace/tools/__tests__/helpers.test.ts` file uses `node:path` resolution which resolves to Windows slashes (`\`), but the test assertions strictly expect POSIX slashes (`/`), causing the test suite to instantly fail for any Windows contributor.
3. **Default Engine Cancellation**: In `packages/core/src/workflows/default.ts`, an aborted workflow does not explicitly check the `abortController.signal` to transition the status to `'canceled'` in the backend storage, leaving workflows orphaned in a running state.

### Steps To Reproduce

1. **To reproduce the Inngest TS2416 error:**
   - Run `pnpm tsc --noEmit` inside the `workflows/inngest` package. 
   - Note the `TS2416: Property is not assignable to the same property in base type` errors.
2. **To reproduce the Windows Test Failure:**
   - Clone the repo on a Windows machine.
   - Run `pnpm --filter @mastra/core test:unit src/workspace/tools/__tests__/helpers.test.ts`
   - Observe `AssertionError: expected "vi.fn()" to be called with arguments: [ '/project/src/app.ts', 'code' ]` but received `[ 'D:\\project\\src\\app.ts', 'code' ]`.
3. **To reproduce the Default Engine state bug:**
   - Trigger a workflow using the default engine and immediately abort the signal.
   - Check the DB storage for the workflow status—it will not correctly reflect `'canceled'`.

The reproduction steps above utilize the existing Mastra codebase and internal unit test suite directly. No external reproducible example is needed as the bugs exist inherently within the core packages.

### Expected Behavior

1. The `inngest` package should compile cleanly without any `TS2416` signature mismatch errors.
2. The `helpers.test.ts` suite should handle cross-platform pathing correctly so tests pass consistently on both macOS/Linux and Windows environments.
3. The Default workflow engine should accurately propagate the `'canceled'` status to persistent storage when an abort signal is received.

### Environment Information

```block
System:
  OS: Windows 11
  CPU: (x64) Intel / AMD
Binaries:
  Node: 20.x.x
  npm: 10.x.x
  pnpm: 9.x.x
```

### Verification

Piyush0049:
## Description
This PR addresses several critical stability and typing issues in the core engine and test suite:

1. **Inngest Type Mismatches**: Resolves `TS2416` type signature mismatches in the Inngest workflow engine (`run.ts`, `types.ts`, `workflow.ts`) by correctly passing the `TRequestContext` generic parameter to match the base `Workflow` class.
2. **Cross-Platform Test Pathing**: Fixes cross-platform path resolution errors in `packages/core/src/workspace/tools/helpers.ts` by using `path.posix.resolve` for virtual LSP paths, ensuring the core test suite passes on Windows environments.
3. **Default Engine Cancellation**: Adds explicit logic to the default workflow engine (`default.ts`) to properly transition workflow states to `'canceled'` in persistent storage when the `abortController.signal` fires, preventing orphaned workflows.

## Related issue(s)
Fixes #19224

## Type of change
- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [x] Test update

## Checklist

## ELI5

When a workflow gets stopped, this PR makes sure it always finishes in a clean “canceled” state (so it doesn’t look like it’s still running) and formats its results correctly. It also fixes Windows-vs-Mac/Linux path handling so virtual LSP diagnostics don’t break, and updates Inngest types so typed `requestContext` workflows compile.

## What changed

- **Canceled workflows now end cleanly and consistently:** The default execution engine detects `abort` during the main step loop, persists the workflow-step update with `workflowStatus: 'canceled'`, marks the execution span as canceled, triggers lifecycle callbacks with `status: 'canceled'`, and returns an immediately normalized canceled result.
- **Canceled-result formatting/deduping is corrected:** Added/extended tests to ensure canceled payloads and metadata are deduplicated and normalized consistently between the main result and `onFinish` results, including stripping nested run IDs in `step.metadata` while preserving other metadata.
- **Fixed Windows vs POSIX virtual LSP paths:** Updated `getEditDiagnosticsText` to choose path resolution correctly based on whether the LSP root is a Windows absolute/UNC path (using `workspace.filesystem.resolveAbsolutePath` when available, otherwise falling back to `path.win32.resolve` vs `path.posix.resolve`).
- **Resolved Inngest `TS2416` type mismatches:** Introduced/threaded a typed `TRequestContext` generic through Inngest workflow/run/config types and ensured `requestContext` parameters are consistently typed as `RequestContext<TRequestContext>` across `start/resume/timeTravel/stream` APIs.
- **Added regression tests:**
  - Default-engine cancellation `onFinish` contract test (asserts `status: 'canceled'`, payload normalization, and persistence behavior).
  - Windows path resolution cases for drive-letter, `c:/...`, and UNC roots.
  - Compile-time TypeScript regression verifying `init<CustomContext>` preserves `TRequestContext` through `createRun` method signatures.
- **Release/update metadata:** Added changesets for `@mastra/core` (cancellation/formatting fix) and `@mastra/inngest` (request-context typing fix).


packages/core/src/workflows/default.ts
```diff
@@ -772,6 +772,62 @@ export class DefaultExecutionEngine extends ExecutionEngine {
...
+          attributes: {
+            status: 'canceled',
+          },
+        });
+
+        await this.invokeLifecycleCallbacks({
+          status: 'canceled',
+          result: undefined,
+          error: undefined,
+          steps: stepResults,
```
NikAiyer: Could we format the canceled result through `fmtReturnValue` before invoking lifecycle callbacks and returning? Other terminal paths persist raw `stepResults`, but expose `result.steps` publicly. Passing raw steps here can leak `metadata.nestedRunId`, omit top-level `input`, and skip payload deduplication. Please also add a regression test asserting that the returned steps and `onFinish` steps are identically normalized.

workflows/inngest/src/run.ts
```diff
@@ -990,8 +991,10 @@ export class InngestRun<
     const self = this;
     const stream = new ReadableStream<WorkflowStreamEvent>({
       async start(controller) {
-        // TODO: fix this, watch doesn't have a type
-        const unwatch = self.watch(async ({ type, from = ChunkFrom.WORKFLOW, payload }) => {
+        const unwatch = self.watch(async (event: WorkflowStreamEvent) => {
+          const type = event.type;
+          const from = event.from ?? ChunkFrom.WORKFLOW;
+          const payload = 'payload' in event ? (event as any).payload : undefined;
```
NikAiyer: Is this guard and `any` cast necessary? `watch()` already declares `WorkflowStreamEvent`, and every current member of that union contains `payload`. Direct destructuring was type-safe and avoided weakening the contract. The same change appears in the time-travel stream path below.

packages/core/src/workspace/tools/helpers.ts
```diff
@@ -97,9 +97,10 @@ export async function getEditDiagnosticsText(workspace: Workspace, filePath: str
     // Use the filesystem's path resolution to get the real disk path.
     // This correctly handles contained: true (virtual paths → basePath)
     // and contained: false (absolute paths used as-is).
+    // Use posix resolve to ensure LSP paths (which are URIs/POSIX) don't get Windows drive letters appended in tests
     const absolutePath =
       workspace.filesystem?.resolveAbsolutePath?.(filePath) ??
-      path.resolve(lspManager.root, filePath.replace(/^\/+/, ''));
+      path.posix.resolve(lspManager.root, filePath.replace(/^\/+/, ''));
```
NikAiyer: Please add a platform-independent regression test for the path behavior this fixes. `LSPManager.root` is not currently constrained or normalized to POSIX format, and `path.posix.resolve('C:\\project', ...)` treats that root as relative. Either the fallback needs to support that case or the POSIX-root invariant should be made explicit.

NikAiyer (changes requested):
Thanks for working through these fixes. Before merge, please:

- Add focused regression tests for the canceled return/`onFinish` contract and typed Inngest request-context compatibility.
- Add a changeset for the runtime cancellation behavior change.

The final-step abort semantics appear pre-existing and can be tracked separately.

Piyush0049:
@NikAiyer Thank you for the review. I have pushed the commit to fix the issues and a merge conflict.
Please do let me know if any changes are required.

packages/core/src/workflows/default.ts
```diff
@@ -806,6 +772,70 @@ export class DefaultExecutionEngine extends ExecutionEngine {
...
+          runId,
+          workflowId,
+          resourceId,
+          input,
+          requestContext: currentRequestContext,
+          state: lastState,
+          stepExecutionPath,
+        });
+
+        return {
```
NikAiyer, line 862: This still manually reconstructs the result and omits normalized fields such as top-level `input`. Please return the complete formatted result:

```ts
return {
  ...formattedResult,
  runId,
  ...(params.outputOptions?.includeState ? { state: lastState } : {}),
};
```

Persistence should continue using raw `stepResults`; callbacks and the public return should use the formatted result.

packages/core/src/workflows/default.test.ts
```diff
@@ -1405,45 +1325,65 @@ describe('DefaultExecutionEngine.deserializeRequestContext', () => {
...
+    expect(result.status).toBe('canceled');
+    expect(onFinishResult).toBeDefined();
+    expect(onFinishResult?.status).toBe('canceled');
 
-    expect(calls).toBe(4);
-    expect(result.ok).toBe(false);
-    if (!result.ok) {
-      expect(result.error.nonRetryable).toBeUndefined();
-    }
+    // The returned steps and onFinish steps should be identically normalized
```
NikAiyer, line 1488: This does not prove the normalization described above: the fixture never creates `nestedRunId`, so that assertion is vacuous.

Please exercise and assert the actual contract: real metadata removal, top-level input, payload deduplication, one `onFinish` call with the returned steps, skipped subsequent execution, run ID/state, and raw persistence data.

workflows/inngest/src/index.test.ts
```diff
@@ -217,6 +217,42 @@ async function resetInngest(expectedFnIds: string[] = []) {
...
+      id: 'step1',
+      inputSchema: z.object({}),
+      outputSchema: z.object({ user: z.string() }),
+      execute: async ({ requestContext }) => {
+        // requestContext should be typed as RequestContext<CustomContext>
+        return { user: requestContext.get('userId') as string };
+      },
+    });
+
+    const workflow = createWorkflow<any, any, any, any, CustomContext>({
```
NikAiyer: CodeRabbit covers the invalid fifth generic argument, but correcting it alone will not propagate `CustomContext`: the factory does not expose an independent request-context generic.

Please thread `TRequestContext` through `init().createWorkflow()` and test typed run methods with `RequestContext<CustomContext>`. A base-`Workflow` assignment would also directly guard the TS2416 substitutability issue.

packages/core/src/workspace/tools/helpers.ts
```diff
@@ -97,9 +97,15 @@ export async function getEditDiagnosticsText(workspace: Workspace, filePath: str
     // Use the filesystem's path resolution to get the real disk path.
     // This correctly handles contained: true (virtual paths → basePath)
     // and contained: false (absolute paths used as-is).
```
NikAiyer, line 99: Please remove the duplicated comment. Also consider `path.win32.isAbsolute(lspManager.root)` instead of the drive-letter regex so UNC roots are handled too.

NikAiyer (changes requested):
Thanks for the update. Please address the unresolved CodeRabbit comments and the additional issues noted inline.

Piyush0049:
@NikAiyer I have pushed the commit to fix the issues that were highlighted.
Please do let me know if any changes are required.

packages/core/src/workspace/tools/helpers.ts
```diff
@@ -97,9 +97,14 @@ export async function getEditDiagnosticsText(workspace: Workspace, filePath: str
...
+    // Use posix resolve to ensure POSIX LSP paths don't get Windows drive letters appended,
+    // but fall back to win32 resolve if the root is explicitly a Windows absolute path.
+    const isWindowsRoot = path.win32.isAbsolute(lspManager.root);
```
NikAiyer: `path.win32.isAbsolute('/project')` is also `true`, so this misclassifies POSIX roots and produces `\project\src\app.ts`. Three existing workspace tests fail as a result.

Our earlier suggestion to use `path.win32.isAbsolute()` was incorrect—sorry about that. Please detect Windows-specific drive and UNC roots instead:

```ts
const isWindowsRoot =
  /^[A-Za-z]:[\\/]/.test(lspManager.root) ||
  /^(?:\\\\|\/\/)[^\\/]+[\\/][^\\/]+/.test(lspManager.root);
```

Please retain `/project` coverage and add cases for `C:\project`, `C:/project`, and `\\server\share`.

workflows/inngest/src/index.ts
```diff
@@ -1120,10 +1120,16 @@ export function init(inngest: Inngest) {
         InngestEngineType
       >[],
     >(params: InngestWorkflowConfig<TWorkflowId, TState, TInput, TOutput, TSteps>) {
```
NikAiyer: Please pass `TRequestContext` as the sixth `InngestWorkflowConfig` generic. The config currently defaults to `unknown` while the constructed workflow expects `TRequestContext`, causing the TypeScript build error at line 1132:

```ts
InngestWorkflowConfig<
  TWorkflowId,
  TState,
  TInput,
  TOutput,
  TSteps,
  TRequestContext
>
```

workflows/inngest/src/index.test.ts
```diff
@@ -217,6 +217,42 @@ async function resetInngest(expectedFnIds: string[] = []) {
...
+      id: 'typed-context-workflow',
+      inputSchema: z.object({}),
+      outputSchema: z.object({ user: z.string() }),
+      steps: [step1],
+    });
+
+    workflow.then(step1).commit();
+
+    // Assigning to the base Workflow type explicitly guards against TS2416 substitutability issues
+    const workflowAsBase: Workflow<any, any, any, any, any, any, any, CustomContext> = workflow as any;
```
NikAiyer: `workflow as any` bypasses the assignability check this test is intended to protect. Please assign `workflow` directly, provide a typed `requestContextSchema`, and verify that the resulting run’s argument type accepts `RequestContext<CustomContext>`.

NikAiyer (changes requested):
Thanks for iterating on this—the earlier cancellation, changeset, and stream concerns look addressed. I found two remaining blockers and one related regression-test gap, noted inline.

Piyush0049:
@NikAiyer I have pushed a commit for the changes you highlighted. Please do let me know if any other changes are required.

Piyush0049:
Hey @NikAiyer, I've got the `TS2416` type mismatch fix ready locally (just needed to pass `runId` into `FormattedWorkflowResult` inside `fmtReturnValue`). I have a commit ready to push to fix it. Shall I push it?

NikAiyer:
Hey @Piyush0049, I pushed a different fix for the type mismatch. I removed runId from FormattedWorkflowResult while keeping it on the final workflow execution results, since fmtReturnValue() doesn’t currently receive or use the run ID. Was there a particular reason you intended FormattedWorkflowResult itself to include it?

Piyush0049:
I was only threading `runId` through to make the `TS2416` compiler error go away. Your fix is definitely the cleaner approach since `fmtReturnValue` doesn't actually need it. No specific reason on my end, I've dropped my local changes and pulled yours.

NikAiyer (approved):
hey @Piyush0049 thanks for making this PR! I will merge once checks pass.

PR #19242, opened 2026-07-10
fix(core): preserve empty output processor results

### LINKED ISSUE #19240: [BUG] Output processors cannot clear final agent text
### Describe the bug

An output processor can clear the final assistant text in the response message, but `Agent.generate()` and `Agent.stream()` still expose the original text through their final result. The response message is correctly empty; the stale value comes from a truthy fallback that treats the valid empty string as if no processed result existed.

This is an empty-string edge case following #9916: non-empty processed text is returned correctly, but a processor cannot safely redact all text.

### Steps to reproduce

1. Configure an agent with an output processor that replaces every assistant text part with `''`.
2. Use a model that returns a known string such as `sensitive response`.
3. Call `agent.generate()` or `agent.stream()` and collect the final output.

```ts
const redactor = {
  id: 'redactor',
  name: 'Redactor',
  async processOutputResult({ messages }) {
    return messages.map(message => ({
      ...message,
      content: {
        ...message.content,
        parts: message.content.parts.map(part => (part.type === 'text' ? { ...part, text: '' } : part)),
      },
    }));
  },
};

const result = await agent.generate('test');
```

### Actual behavior

`result.response.messages` contains the empty processed text, but `result.text`, the final streamed text, and the last step's `text` still contain `sensitive response`.

### Expected behavior

All final result views should return the processor's intentional empty string. A processor used for redaction must not leak the pre-processed text through a secondary result field.

### Verification

- Reproduced with deterministic generate and stream regressions on `upstream/main` at `8345f18fdf`.
- The unmodified fallback uses `outputText || originalText`, which makes the valid empty string unreachable.
- The focused output and agent-processor test files pass with the proposed repair (92 tests).
- I searched existing issues and pull requests; #9916 is related but closed and does not cover this empty-string boundary.

NianJiuZst:
Fixes #19240.

When an output processor intentionally replaced the final assistant text with an empty string, the processed response message was empty but `result.text`, streamed final text, and the final step fell back to the original text. That could expose content that a redaction processor had removed.

This change treats an empty string as a real processed result when the response text changed. It preserves the existing retry fallback only when the response was already empty before processing or when there is no response message to inspect.

The regression coverage exercises both `Agent.generate()` and `Agent.stream()`, and also keeps the empty-response retry path intact. Verified with the focused output and agent-processor suites (92 tests), `pnpm --filter @mastra/core lint`, Prettier, and `git diff --check`.

`pnpm --filter @mastra/core check` is currently blocked by unchanged schema type errors in `src/llm/model/model.ts` and `src/tools/tool-builder/builder.ts`. The broader core unit suite also has existing failures in Unix-socket and evented-workflow tests; neither failure touches this diff.

## ELI5

When a tool intentionally removes an agent’s response, the agent now correctly returns an empty answer instead of accidentally showing the original text again. Existing retry behavior remains unchanged.

## What changed

- Updated final output handling to preserve processed empty strings.
- Ensured `generate()` and `stream()` consistently expose cleared text in:
  - Response messages
  - Result text
  - Streamed final text
  - Final step text
- Preserved fallback behavior for initially empty responses and missing response messages.
- Added regression tests for generate, stream, and retry scenarios.

### DIFF
--- a/.changeset/fancy-eels-occur.md
+++ b/.changeset/fancy-eels-occur.md
@@ -1,0 +1,5 @@
+---
+'@mastra/core': patch
+---
+
+Fixed final agent results leaking original text after an output processor clears it.
--- a/packages/core/src/agent/agent-processor.test.ts
+++ b/packages/core/src/agent/agent-processor.test.ts
@@ -24,6 +24,43 @@ const createMessage = (text: string, role: 'user' | 'assistant' = 'user'): Mastr
   createdAt: new Date(),
 });
 
+const createTextClearingProcessor = (): Processor => ({
+  id: 'text-clearing-processor',
+  name: 'Text Clearing Processor',
+  async processOutputResult({ messages }) {
+    return messages.map(message => ({
+      ...message,
+      content: {
+        ...message.content,
+        parts: message.content.parts.map(part => (part.type === 'text' ? { ...part, text: '' } : part)),
+      },
+    }));
+  },
+});
+
+const createSensitiveResponseModel = () =>
+  new MockLanguageModelV2({
+    doGenerate: async () => ({
+      content: [{ type: 'text', text: 'sensitive response' }],
+      finishReason: 'stop',
+      usage: { inputTokens: 2, outputTokens: 2, totalTokens: 4 },
+      rawCall: { rawPrompt: null, rawSettings: {} },
+      warnings: [],
+    }),
+    doStream: async () => ({
+      stream: convertArrayToReadableStream([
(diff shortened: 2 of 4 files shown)


abhiaiyer91 (changes requested):
Factory comments

abhiaiyer91:
Thanks for taking this on, and sorry it sat for a while. Issue #19240 has since been fixed by #20998, which has merged, so this one is now superseded. Really appreciate you spotting the problem and putting up a fix — please do keep the contributions coming.

PR #19246, opened 2026-07-10
fix(rag): record rerank scoring errors

### LINKED ISSUE #19245: [BUG] RAG rerank observability span is not marked as errored when scoring fails
### Describe the Bug

`rerank()` and `rerankWithScorer()` create a `RAG_ACTION` observability span for rerank operations, but the span is only marked as errored for
  invalid weight configuration.

  If scoring fails later, for example when `scorer.getRelevanceScore()` rejects for one of the retrieved results, the error bubbles out of
  `executeRerank()` but the rerank span is never ended with error status.

  Current code marks the span as errored for invalid weights:

  ```ts
  if (!sum.eq(1)) {
    const err = new Error(`Weights must add up to 1. Got ${sum} from ${weights}`);
    rerankSpan?.error({ error: err, endSpan: true });
    throw err;
  }

  But scoring errors inside Promise.all(...) are not reported to the span:

  const scoredResults = await Promise.all(
    results.map(async (result, index) => {
      let semanticScore = 0;
      if (result?.metadata?.text) {
        semanticScore = await scorer.getRelevanceScore(query, result?.metadata?.text);
      }

      // ...
    }),
  );

  This makes failed rerank operations harder to debug in traces because the operation fails from the caller’s perspective, but the span does not
  record the failure.

### Steps To Reproduce


  ```ts
  import { rerankWithScorer } from '@mastra/rag';

  const scorer = {
    async getRelevanceScore() {
      throw new Error('scorer failed');
    },
  };

  const results = [
    {
      id: '1',
      metadata: { text: 'document text' },
      score: 0.8,
    },
  ];

  await rerankWithScorer({
    results,
    query: 'test query',
    scorer,
    options: {
      topK: 1,
      // provide an observabilityContext with a currentSpan
    },
  });


 Minimal repro included inline in the Steps To Reproduce section.

### Expected Behavior


  ```md
  When rerank scoring fails, the rerank `RAG_ACTION` span should be ended with error status before the error is rethrown.

  Expected trace behavior:

  - `RAG_ACTION` span is created for rerank.
  - If `scorer.getRelevanceScore()` throws or rejects, the span records the error.
  - The original error still propagates to the caller.
  - Existing successful rerank behavior remains unchanged.
  - Existing invalid weight handling continues to mark the span as errored.

### Environment Information

```block
System:
    OS: macOS 26.5.1
    CPU: (10) arm64 Apple M5
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 24.14.1
    npm: 11.11.0
    pnpm: 11.8.0
  npmPackages:
    @mastra/rag: source/main
    @mastra/core: source/main
    vitest: 4.1.9
    typescript: 6.0.3
```

### Verification

GautamSharma99:
Fixes #19245

  Summary

  This PR fixes RAG rerank observability so scoring failures are recorded on the rerank RAG_ACTION span before the error is rethrown.

  Previously, executeRerank() marked the span as errored for invalid weight configuration, but failures from scorer.getRelevanceScore() were not
  recorded on the span. The caller still received the error, but traces did not show the rerank span as failed.

  Changes

  - Wrap rerank scoring in a narrow try/catch.
  - Call rerankSpan.error({ error, endSpan: true }) when scoring fails.
  - Rethrow the original error so caller behavior is unchanged.
  - Keep the existing successful rerankSpan.end(...) behavior unchanged.
  - Add regression coverage for scorer failure recording the span error.

  Validation

  pnpm --filter ./packages/rag test src/rerank/index.test.ts
  pnpm test:rag
  pnpm --filter ./packages/rag lint
  pnpm build:rag
  pnpm prettier --check packages/rag/src/rerank/index.ts packages/rag/src/rerank/index.test.ts

## ELI5

If reranking fails, the system now records that failure in tracing before passing the error along, making it easier to understand what went wrong without changing successful behavior.

## What changed

- Wrapped rerank scoring in a narrow `try/catch`.
- Records scoring errors on the rerank span and ends it before rethrowing.
- Preserves successful span completion and existing invalid-weight handling.
- Added regression coverage for scorer failures.
- Bumped `@mastra/rag` with a patch changeset.

### DIFF
--- a/.changeset/little-kids-bake.md
+++ b/.changeset/little-kids-bake.md
@@ -1,0 +1,5 @@
+---
+'@mastra/rag': patch
+---
+
+Fixed RAG rerank observability spans to record scoring failures before rethrowing.
--- a/packages/rag/src/rerank/index.test.ts
+++ b/packages/rag/src/rerank/index.test.ts
@@ -2,7 +2,7 @@ import { cohere } from '@ai-sdk/cohere';
 import { describe, it, expect, vi, beforeEach } from 'vitest';
 import { CohereRelevanceScorer } from './relevance';
 
-import { rerank } from '.';
+import { rerank, rerankWithScorer } from '.';
 
 vi.spyOn(CohereRelevanceScorer.prototype, 'getRelevanceScore').mockImplementation(async () => {
   return 1;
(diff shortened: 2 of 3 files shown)


abhiaiyer91 (approved):

GautamSharma99:
Hey @abhiaiyer91 , this PR fixes the rerank observability gap from the linked issue.

The change is intentionally small: it only wraps the scoring step so failures from getRelevanceScore() are recorded on the rerank RAG_ACTION span before rethrowing the same error. Successful rerank behavior and the existing invalid-weights error path are unchanged.
I added a focused regression test for the scorer failure case and ran the RAG test/lint/build checks.

PR #19255, opened 2026-07-10
fix(background-tasks): distinguish `cancelled` tasks from `completed` tasks in continuation prompt

### LINKED ISSUE #19190: Background Task Cancellations sending Notifications
This issue was created from Discord post 1524875298015805511:

[Open in Discord](https://discord.com/channels/1309558646228779139/1524875298015805511)

So I'm implementing background tasks in my chatbot, and wanted to introduce an api to cancel the background task. This was simple enough, we send a cancel signal that stops the bg task correctly. However, what comes after isn't quite what I expected. I have a simple tool to run in the background, and this is the response I get when starting and cancelling

After start:
Another 10-second countdown is running in the background — I'll let you know when it finishes.

After cancel:
The countdown from 10 to 0 just finished successfully.

I was looking at the mastra code, and it looks like `background-task-cancelled` is treated as one of the terminal background-task chunks consumed by the `untilIdle` continuation path, alongside completed/failed tasks. The directive for completed tasks is `Background task(s) you previously dispatched have completed...` which doesn't differentiate between cancelled and success/failed. I was wondering if there was a reason behind this or if it's a bug, because I would expect if a bg task was cancelled, it would just be dropped and the agent wouldn't need to be aware of it at all! I can see how it could think the bg task is still running though, but there should be a better way (perhaps writing to memory?) to show that it was cancelled rather than sending a notifcation that the background task was completed.

A minimum reproduction would just be to build a simple bg task and cancel it as above.

JayaKrishnaNamburu:
fixes https://github.com/mastra-ai/mastra/issues/19190

## Description

You can use the setup here to reproduce the same https://github.com/JayaKrishnaNamburu/split-background-statuses/blob/main/src/mastra/index.ts. 

By registering a cancel route and triggering to cancel it.
https://github.com/JayaKrishnaNamburu/split-background-statuses/blob/d58cfafaaa90a937735df7a13ed367444f96d0b2/src/mastra/index.ts#L49-L69

When passing status of the background tasks to the LLM, all the tasks are sent under one list. So, according to LLM anything other than `suspended` considered as completed.
https://github.com/mastra-ai/mastra/blob/1b2f2cca30f04f2cb8e20df661733a1367d779a0/packages/core/src/loop/shared/stream-until-idle-helpers.ts#L96-L122

A fix is to separate  the `taskId`'s w.r.t to the statuses. 
https://github.com/JayaKrishnaNamburu/mastra/blob/d4b95b2ae5dc9fc2a5b4267e8ac80cdb60c2d46f/packages/core/src/loop/shared/stream-until-idle-helpers.ts#L124-L149

While checking this, i noticed that when a `prevResult` is not preset for the execution engine is seems to throw an error as the engine is trying to read status on top of it. Normalised that too.
Here is the stack-trace

```
background-task workflow start failed for 6b268dcd-dcd8-4c9f-9a23-3ef2e095a478: TypeError: Cannot read properties of undefined (reading 'status')
    at EventedExecutionEngine.execute (file:///Users/jk/Documents/personal/test/zod-repro/node_modules/.pnpm/@mastra+core@1.50.1_express@5.2.1_zod@4.4.3/node_modules/@mastra/core/dist/chunk-OE4IEL7C.js:19024:31)
    at process.processTicksAndRejections (node:internal/process/task_queues:104:5)
```

## Current Behaviour
Screen Recording 2026-07-11 at 4 04 36 PM

## Behaviour after the fix
Screen Recording 2026-07-11 at 3 57 21 PM

## Type of change

- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [x] Test update

## Checklist

## ELI5
The system was telling the AI that “cancelled” background work had actually “finished,” which could lead to wrong next steps. This PR keeps cancelled vs completed outcomes separate, and also makes sure terminal workflow events always include the right status even when there’s no previous result.

## Changes
- Updated the continuation prompt to build separate tool-call ID groups (and instructions) for **completed**, **failed**, **cancelled**, and **suspended** tasks, so cancelled tasks aren’t treated as completed.
- Normalized workflow terminal handling when `prevResult` is missing:
  - In `WorkflowEventProcessor.endWorkflow`, it now creates a non-undefined `normalizedPrevResult`.
  - Persists and publishes the normalized result so `status` is correctly set (e.g., `'canceled'`).
- Added/updated a test to verify `endWorkflow('canceled', undefined)` emits exactly one `workflow.end` event and that both the published payload and persisted snapshot contain `prevResult.status: 'canceled'`.
- Added a `@mastra/core` patch changeset documenting the fixes.

### DIFF
--- a/.changeset/brave-clocks-fly.md
+++ b/.changeset/brave-clocks-fly.md
@@ -1,0 +1,5 @@
+---
+'@mastra/core': patch
+---
+
+Fixed background-task cancellation so terminal workflow events always carry a valid `prevResult`, and split the continuation prompt so completed, failed, cancelled, and suspended background tasks each get clearer LLM instructions.
--- a/.changeset/brave-clocks-fly.md
+++ b/.changeset/brave-clocks-fly.md
@@ -2,4 +2,4 @@
 '@mastra/core': patch
 ---
 
-Fixed background-task cancellation so terminal workflow events always carry a valid `prevResult`, and split the continuation prompt so completed, failed, cancelled, and suspended background tasks each get clearer LLM instructions.
+Fixed background-task cancellation so cancelled tasks no longer look completed to the agent, terminal workflow events still surface a valid result when cancellation happens before one is produced, and completed, failed, cancelled, and suspended background tasks each get clearer continuation instructions.
--- a/packages/core/src/loop/shared/stream-until-idle-helpers.ts
+++ b/packages/core/src/loop/shared/stream-until-idle-helpers.ts
@@ -76,7 +76,7 @@ export async function resolveScope(
 
 /**
  * Build the ephemeral user-prompt text that tells the LLM which tool-call
- * IDs just completed. The directive stops the LLM from (a) re-processing
+ * IDs just completed / failed / canceled. The directive stops the LLM from (a) re-processing
  * results already handled on a prior continuation and (b) mimicking the
  * prior assistant ack text ("I'm running it in the background") and
  * re-dispatching the same tool.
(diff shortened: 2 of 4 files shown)


JayaKrishnaNamburu:
Will share the repro project, and the fix recordings in a short while.

JayaKrishnaNamburu:
@coderabbitai review

PR #19265, opened 2026-07-10
chore(deps): update security updates [security] (major)

daneatmastra:
> ℹ️ **Note**
> 
> This PR body was truncated due to platform limits.

This PR contains the following updates:

| Package | Type | Update | Change |
|---|---|---|---|
| [@hono/node-server](https://redirect.github.com/honojs/node-server) | devDependencies | major | [`^1.19.14` → `^2.0.0`](https://renovatebot.com/diffs/npm/@hono%2fnode-server/1.19.14/2.0.5) |
| [brace-expansion@<1.1.13](https://redirect.github.com/juliangruber/brace-expansion) | pnpm-workspace.overrides | major | [`1.1.13` → `5.0.8`](https://renovatebot.com/diffs/npm/brace-expansion@<1.1.13/1.1.13/5.0.8) |
| [brace-expansion@>=2.0.0 <2.0.3](https://redirect.github.com/juliangruber/brace-expansion) | pnpm-workspace.overrides | major | [`2.1.2` → `5.0.8`](https://renovatebot.com/diffs/npm/brace-expansion@>=2.0.0 <2.0.3/2.1.2/5.0.8) |
| [js-yaml](https://redirect.github.com/nodeca/js-yaml) | pnpm-workspace.overrides | major | [`^3.15.0` → `^5.2.2`](https://renovatebot.com/diffs/npm/js-yaml/3.15.0/5.2.2) |
| [pnpm](https://pnpm.io) ([source](https://redirect.github.com/pnpm/pnpm/tree/HEAD/pnpm11/pnpm)) | packageManager | major | [`10.29.3+sha512.498e1fb4cca5aa06c1dcf2611e6fafc50972ffe7189998c409e90de74566444298ffe43e6cd2acdc775ba1aa7cc5e092a8b7054c811ba8c5770f84693d33d2dc` → `11.5.3`](https://renovatebot.com/diffs/npm/pnpm/10.29.3/11.5.3) |
| [react-router](https://redirect.github.com/remix-run/react-router) ([source](https://redirect.github.com/remix-run/react-router/tree/HEAD/packages/react-router)) | dependencies | major | [`^7.13.1` → `^8.0.0`](https://renovatebot.com/diffs/npm/react-router/7.18.0/8.3.0) |

---

### Node.js Adapter for Hono: Path traversal in `serve-static` on Windows via encoded backslash (`%5C`)
[GHSA-frvp-7c67-39w9](https://redirect.github.com/advisories/GHSA-frvp-7c67-39w9)

<details>
<summary>More information</summary>

#### Details
The same as the `hono` core [Path traversal in `serve-static` on Windows via encoded backslash (`%5C`)](https://redirect.github.com/honojs/hono/security/advisories/GHSA-wwfh-h76j-fc44).

##### Summary

On Windows hosts, an encoded backslash (`%5C`) in the request path decodes to `\`, which the Windows path resolver treats as a separator. `serve-static` then resolves a single URL segment such as `admin\secret.txt` into a nested file under the root and serves it, letting an attacker read static files meant to be protected behind prefix-mounted middleware. Directory escape (`..`) remains blocked.

##### Details

The router splits paths only on `/`, so `/admin%5Csecret.txt` is one segment and middleware on `/admin/*` does not run. The `serve-static` guard rejects `.`/`..` and consecutive separators but lets a lone `\` through; on Windows the file resolver re-splits it into the protected subtree.

This affects Windows hosts serving static files via the Node, Bun, or Deno adapters that guard a static subtree with prefix-mounted middleware.

##### Impact

An unauthenticated attacker can read static files under a middleware-guarded prefix on Windows hosts. The read stays within the configured root; escape outside the root is not possible.

#### Severity
- CVSS Score: 5.9 / 10 (Medium)
- Vector String: `CVSS:3.1/AV:N/AC:H/PR:N/UI:N/S:U/C:H/I:N/A:N`

#### References
- [https://github.com/honojs/node-server/security/advisories/GHSA-frvp-7c67-39w9](https://redirect.github.com/honojs/node-server/security/advisories/GHSA-frvp-7c67-39w9)
- [https://github.com/honojs/node-server/commit/cd076e117cfe4cb8d31f9eb11d2e8f38a6cb8faf](https://redirect.github.com/honojs/node-server/commit/cd076e117cfe4cb8d31f9eb11d2e8f38a6cb8faf)
- [https://github.com/honojs/node-server/releases/tag/v2.0.5](https://redirect.github.com/honojs/node-server/releases/tag/v2.0.5)
- [https://github.com/advisories/GHSA-frvp-7c67-39w9](https://redirect.github.com/advisories/GHSA-frvp-7c67-39w9)

This data is provided by the [GitHub Advisory Database](https://redirect.github.com/advisories/GHSA-frvp-7c67-39w9) ([CC-BY 4.0](https://redirect.github.com/github/advisory-database/blob/main/LICENSE.md)).
</details>

---

### Node.js Adapter for Hono: Path traversal in `serve-static` on Windows via encoded backslash (`%5C`)
[GHSA-frvp-7c67-39w9](https://redirect.github.com/advisories/GHSA-frvp-7c67-39w9)

<details>
<summary>More information</summary>

#### Details
The same as the `hono` core [Path traversal in `serve-static` on Windows via encoded backslash (`%5C`)](https://redirect.github.com/honojs/hono/security/advisories/GHSA-wwfh-h76j-fc44).

##### Summary

On Windows hosts, an encoded backslash (`%5C`) in the request path decodes to `\`, which the Windows path resolver treats as a separator. `serve-static` then resolves a single URL segment such as `admin\secret.txt` into a nested file under the root and serves it, letting an attacker read static files meant to be protected behind prefix-mounted middleware. Directory escape (`..`) remains blocked.

##### Details

The router splits paths only on `/`, so `/admin%5Csecret.txt` is one segment and middleware on `/admin/*` does not run. The `serve-static` guard rejects `.`/`..` and consecutive separators but lets a lone `\` through; on Windows the file resolver re-splits it into the protected subtree.

This affects Windows hosts serving static files via the Node, Bun, or Deno adapters that guard a static subtree with prefix-mounted middleware.

##### Impact

An unauthenticated attacker can read static files under a middleware-guarded prefix on Windows hosts. The read stays within the configured root; escape outside the root is not possible.

#### Severity
- CVSS Score: 5.9 / 10 (Medium)
- Vector String: `CVSS:3.1/AV:N/AC:H/PR:N/UI:N/S:U/C:H/I:N/A:N`

#### References
- [https://github.com/honojs/node-server/security/advisories/GHSA-frvp-7c67-39w9](https://redirect.github.com/honojs/node-server/security/advisories/GHSA-frvp-7c67-39w9)
- [https://github.com/honojs/node-server/commit/cd076e117cfe4cb8d31f9eb11d2e8f38a6cb8faf](https://redirect.github.com/honojs/node-server/commit/cd076e117cfe4cb8d31f9eb11d2e8f38a6cb8faf)
- [https://github.com/honojs/node-server](https://redirect.github.com/honojs/node-server)
- [https://github.com/honojs/node-server/releases/tag/v2.0.5](https://redirect.github.com/honojs/node-server/releases/tag/v2.0.5)

This data is provided by [OSV](https://osv.dev/vulnerability/GHSA-frvp-7c67-39w9) and the [GitHub Advisory Database](https://redirect.github.com/github/advisory-database) ([CC-BY 4.0](https://redirect.github.com/github/advisory-database/blob/main/LICENSE.md)).
</details>

---

### brace-expansion: DoS via exponential-time expansion of consecutive non-expanding {} groups
[CVE-2026-13149](https://nvd.nist.gov/vuln/detail/CVE-2026-13149) / [GHSA-3jxr-9vmj-r5cp](https://redirect.github.com/advisories/GHSA-3jxr-9vmj-r5cp)

<details>
<summary>More information</summary>

##### Summary
brace-expansion's expand() exhibits exponential-time - O(2ⁿ) - behavior in the number of consecutive non-expanding {} groups. A short, all-ASCII input (~90 bytes/30 groups) blocks the calling thread for minutes; a slightly longer input hangs it effectively indefinitely. Because the dominant consumers run on Node's single-threaded event loop, one small input can fully stall a worker/process.

In `expand_`, `post` is computed unconditionally at the top of the function, before the early-return branches that don't use it:
```js
const post = m.post.length ? expand_(m.post, max, false) : [''];   // always recurses
  ...
if (!isSequence && !isOptions) {
  if (m.post.match(/,(?!,).*\}/)) {
    str = m.pre + '{' + m.body + escClose + m.post;
    return expand_(str, max, true); // restart — `post` discarded
  }
  return [str];
}
```

For input like a{},{},…, the first {} is non-expanding, so control reaches the {a},b} rewrite branch - but `expand_` has already recursed into post over the entire remaining tail, only to throw the result away.
Each level therefore spawns two recursive expansions over essentially the same remaining work: `T(n) = 2·T(n−1) ⇒ O(2ⁿ)`.

The max option does not mitigate this: max only bounds the output-building loops; neither the post recursion nor the rewrite recursion consults it.
  
Measured on 5.0.6:

| groups (n) | input bytes | time |
|---|---|---|
| 20 | 60 | 130 ms |
| 24 | 72 | 1.9 s |
| 26 | 78 | 7.8 s |
| 30 (PoC) | 90 | ~2 min |

##### Proof of concept
```js
const { expand } = require('brace-expansion');
// 30 non-expanding groups, ~90 bytes — blocks for minutes:
expand('a{},{},{},{},{},{},{},{},{},{},{},{},{},{},{},{},{},{},{},{},{},{},{},{},{},{},{},{},{},{}');
```

##### Impact

Any application that passes attacker-influenced strings to brace-expansion.expand() - directly or transitively via minimatch/glob brace patterns - can be driven into a multi-minute-to-indefinite CPU hang by a tiny request, denying service on that thread/process.

##### Remediation

Upgrade to a patched release. The fix:
1. Defers computing post until after the early-return branches (and computes it locally in the $-suffix branch), so post is only expanded when a brace set actually expands and the value is used. This alone removes the exponential.
1. Converts the {a},b} rewrite from recursion to an in-function loop, so a long run of rewrites cannot grow the call stack.

Verified: the PoC drops from ~2 min to 0.55 ms, 5,000 groups complete in ~344 ms, and output is identical to 5.0.6 across a behavioral-equivalence suite (sequences, padding, $-prefix, a{},b}c, {},a}b, x{{a,b}}y, etc.). Post-fix complexity is ~O(n²) on this input class - acceptable for the security fix; a linear rewrite can be a non-urgent follow-up.

If immediate upgrade isn't possible, avoid passing untrusted input to expand() / glob brace patterns, or run such expansion under a timeout/worker.

#### Severity
- CVSS Score: 7.7 / 10 (High)
- Vector String: `CVSS:4.0/AV:N/AC:L/AT:N/PR:N/UI:N/VC:N/VI:N/VA:H/SC:N/SI:N/SA:N/E:P/S:N/AU:Y/R:U/V:D/RE:M/U:Amber`

#### References
- [https://github.com/juliangruber/brace-expansion/security/advisories/GHSA-3jxr-9vmj-r5cp](https://redirect.github.com/juliangruber/brace-expansion/security/advisories/GHSA-3jxr-9vmj-r5cp)
- [https://nvd.nist.gov/vuln/detail/CVE-2026-13149](https://nvd.nist.gov/vuln/detail/CVE-2026-13149)
- [https://github.com/juliangruber/brace-expansion/pull/122](https://redirect.github.com/juliangruber/brace-expansion/pull/122)
- [https://github.com/juliangruber/brace-expansion/pull/123](https://redirect.github.com/juliangruber/brace-expansion/pull/123)
- [https://github.com/juliangruber/brace-expansion/commit/835d6be91201122d9adffb0c0c8c094189ace265](https://redirect.github.com/juliangruber/brace-expansion/commit/835d6be91201122d9adffb0c0c8c094189ace265)
- [https://github.com/juliangruber/brace-expansion/commit/c7e33ec13ac1a684c116720843ce24e208611754](https://redirect.github.com/juliangruber/brace-expansion/commit/c7e33ec13ac1a684c116720843ce24e208611754)
- [https://github.com/juliangruber/brace-expansion/commit/d74e63030c012e3b7ae81657b8d665619cd51b95](https://redirect.github.com/juliangruber/brace-expansion/commit/d74e63030c012e3b7ae81657b8d665619cd51b95)
- [https://github.com/juliangruber/brace-expansion/releases/tag/v1.1.16](https://redirect.github.com/juliangruber/brace-expansion/releases/tag/v1.1.16)
- [https://github.com/juliangruber/brace-expansion/releases/tag/v2.1.2](https://redirect.github.com/juliangruber/brace-expansion/releases/tag/v2.1.2)
- [https://github.com/juliangruber/brace-expansion/releases/tag/v5.0.7](https://redirect.github.com/juliangruber/brace-expansion/releases/tag/v5.0.7)
- [https://www.npmjs.com/package/brace-expansion](https://www.npmjs.com/package/brace-expansion)
- [https://github.com/advisories/GHSA-3jxr-9vmj-r5cp](https://redirect.github.com/advisories/GHSA-3jxr-9vmj-r5cp)

This data is provided by the [GitHub Advisory Database](https://redirect.github.com/advisories/GHSA-3jxr-9vmj-r5cp) ([CC-BY 4.0](https://redirect.github.com/github/advisory-database/blob/main/LICENSE.md)).
</details>

---

### brace-expansion: DoS via unbounded expansion length causing an out-of-memory process crash
[CVE-2026-14257](https://nvd.nist.gov/vuln/detail/CVE-2026-14257) / [GHSA-mh99-v99m-4gvg](https://redirect.github.com/advisories/GHSA-mh99-v99m-4gvg)

<details>
<summary>More information</summary>

##### Summary

`expand()` bounds the *number* of results it produces (the `max` option,
`100_000` by default) but not their *length*. By chaining many brace groups,
an attacker keeps the result count under `max` while making every result grow
with the number of groups. Building `max` long results — plus the intermediate
arrays combined at each brace group — exhausts memory and crashes the Node
process with an **uncatchable** out-of-memory error. `try/catch` around
`expand()` does not help: the fatal error terminates the process.

A ~7.5 KB input (`'{a,b}'.repeat(1500)`) is enough to crash a default Node
process.

##### Details

For `N` chained brace groups such as `'{a,b}'.repeat(N)`:

- the result count is `2^N`, immediately capped at `max` (`100_000`), so the
  `max` protection appears to hold, but
- each result is `N` characters long, so the total output size is
  `max × N` characters, which grows without bound in `N`.

`expand_` combines each brace set with the fully-expanded tail:

```js
const post = m.post.length ? expand_(m.post, max, false) : ['']
...
for (let j = 0; j < N.length; j++) {
  for (let k = 0; k < post.length && expansions.length < max; k++) {
    const expansion = pre + N[j] + post[k]   // grows one group longer per level
    ...
    expansions.push(expansion)
  }
}
```

The loop guard `expansions.length < max` limits how many strings are built, but
nothing limits how long they get. Each recursion level materializes another
array of up to `max` strings, one character longer than the level below, and —
because V8 represents `pre + N[j] + post[k]` as a cons-string (rope) that
references `post[k]` — those intermediate strings stay reachable through the
whole chain. Memory therefore scales with `max × N`.

Measured on `5.0.7` (`'{a,b}'.repeat(N)`, default `max`):

| groups (N) | input bytes | result count | peak RSS |
|---|---|---|---|
| 20 | 100 | 100,000 | ~80 MB |
| 50 | 250 | 100,000 | ~214 MB |
| 100 | 500 | 100,000 | ~409 MB |
| 300 | 1,500 | 100,000 | ~1,148 MB |
| 1500 | 7,500 | — | **OOM crash** |

##### Proof of concept

```js
const { expand } = require('brace-expansion')

// ~7.5 KB input — crashes the process with a fatal, uncatchable OOM:
//   FATAL ERROR: ... JavaScript heap out of memory
try {
  expand('{a,b}'.repeat(1500))
} catch (e) {
  // never reached — the process is already dead
}
```

##### Impact

Any application that passes attacker-influenced strings to
`brace-expansion.expand()` — directly, or transitively via `minimatch` / `glob`
brace patterns — can be crashed by a small request. Because the failure is a
fatal V8 out-of-memory error rather than a thrown exception, it cannot be caught
and it takes down the whole worker/process, denying service.

##### Remediation

Upgrade to a patched release. The fix bounds the total number of characters a
single `expand()` call may accumulate (`EXPANSION_MAX_LENGTH`, default
`4_000_000`, configurable via a new `maxLength` option), applied inside the
output-building loops so intermediate arrays are bounded too. Once the limit is
reached, output is truncated — consistent with how `max` already truncates —
instead of growing without bound. The limit sits well above any realistic
expansion (100,000 results hitting `max` measure ~1M characters), so legitimate
input is unaffected.

After the fix, `'{a,b}'.repeat(1500)` returns a bounded, truncated result in
~0.7 s using ~340 MB and never crashes, including under a constrained 512 MB
heap.

The fix bounds memory but the algorithm still rebuilds intermediate arrays at
each level (roughly `O(N × maxLength)` work on this input class). A streaming
rewrite that produces output in `O(total output size)` can be a non-urgent
follow-up.

If immediate upgrade isn't possible, avoid passing untrusted input to
`expand()` / glob brace patterns, or pass a small explicit `max` **and**
`maxLength`.

#### Severity
- CVSS Score: 7.5 / 10 (High)
- Vector String: `CVSS:3.1/AV:N/AC:L/PR:N/UI:N/S:U/C:N/I:N/A:H`

#### References
- [https://github.com/juliangruber/brace-expansion/security/advisories/GHSA-mh99-v99m-4gvg](https://redirect.github.com/juliangruber/brace-expansion/security/advisories/GHSA-mh99-v99m-4gvg)
- [https://nvd.nist.gov/vuln/detail/CVE-2026-14257](https://nvd.nist.gov/vuln/detail/CVE-2026-14257)
- [https://github.com/juliangruber/brace-expansion/commit/a1bd33999ea75262c4749fff3bbb0d1372bd07b5](https://redirect.github.com/juliangruber/brace-expansion/commit/a1bd33999ea75262c4749fff3bbb0d1372bd07b5)
- [https://github.com/juliangruber/brace-expansion](https://redirect.github.com/juliangruber/brace-expansion)
- [https://www.npmjs.com/package/brace-expansion](https://www.npmjs.com/package/brace-expansion)
- [https://github.com/advisories/GHSA-mh99-v99m-4gvg](https://redirect.github.com/advisories/GHSA-mh99-v99m-4gvg)

This data is provided by the [GitHub Advisory Database](https://redirect.github.com/advisories/GHSA-mh99-v99m-4gvg) ([CC-BY 4.0](https://redirect.github.com/github/advisory-database/blob/main/LICENSE.md)).
</details>

---

### brace-expansion: DoS via exponential-time expansion of consecutive non-expanding {} groups
[CVE-2026-13149](https://nvd.nist.gov/vuln/detail/CVE-2026-13149) / [GHSA-3jxr-9vmj-r5cp](https://redirect.github.com/advisories/GHSA-3jxr-9vmj-r5cp)

<details>
<summary>More information</summary>

##### Summary
brace-expansion's expand() exhibits exponential-time - O(2ⁿ) - behavior in the number of consecutive non-expanding {} groups. A short, all-ASCII input (~90 bytes/30 groups) blocks the calling thread for minutes; a slightly longer input hangs it effectively indefinitely. Because the dominant consumers run on Node's single-threaded event loop, one small input can fully stall a worker/process.

In `expand_`, `post` is computed unconditionally at the top of the function, before the early-return branches that don't use it:
```js
const post = m.post.length ? expand_(m.post, max, false) : [''];   // always recurses
  ...
if (!isSequence && !isOptions) {
  if (m.post.match(/,(?!,).*\}/)) {
    str = m.pre + '{' + m.body + escClose + m.post;
    return expand_(str, max, true); // restart — `post` discarded
  }
  return [str];
}
```

For input like a{},{},…, the first {} is non-expanding, so control reaches the {a},b} rewrite branch - but `expand_` has already recursed into post over the entire remaining tail, only to throw the result away.
Each level therefore spawns two recursive expansions over essentially the same remaining work: `T(n) = 2·T(n−1) ⇒ O(2ⁿ)`.

The max option does not mitigate this: max only bounds the output-building loops; neither the post recursion nor the rewrite recursion consults it.
  
Measured on 5.0.6:

| groups (n) | input bytes | time |
|---|---|---|
| 20 | 60 | 130 ms |
| 24 | 72 | 1.9 s |
| 26 | 78 | 7.8 s |
| 30 (PoC) | 90 | ~2 min |

##### Proof of concept
```js
const { expand } = require('brace-expansion');
// 30 non-expanding groups, ~90 bytes — blocks for minutes:
expand('a{},{},{},{},{},{},{},{},{},{},{},{},{},{},{},{},{},{},{},{},{},{},{},{},{},{},{},{},{},{}');
```

##### Impact

Any application that passes attacker-influenced strings to brace-expansion.expand() - directly or transitively via minimatch/glob brace patterns - can be driven into a multi-minute-to-indefinite CPU hang by a tiny request, denying service on that thread/process.

##### Remediation

Upgrade to a patched release. The fix:
1. Defers computing post until after the early-return branches (and computes it locally in the $-suffix branch), so post is only expanded when a brace set actually expands and the value is used. This alone removes the exponential.
1. Converts the {a},b} rewrite from recursion to an in-function loop, so a long run of rewrites cannot grow the call stack.

Verified: the PoC drops from ~2 min to 0.55 ms, 5,000 groups complete in ~344 ms, and output is identical to 5.0.6 across a behavioral-equivalence suite (sequences, padding, $-prefix, a{},b}c, {},a}b, x{{a,b}}y, etc.). Post-fix complexity is ~O(n²) on this input class - acceptable for the security fix; a linear rewrite can be a non-urgent follow-up.

If immediate upgrade isn't possible, avoid passing untrusted input to expand() / glob brace patterns, or run such expansion under a timeout/worker.

#### Severity
- CVSS Score: 7.7 / 10 (High)
- Vector String: `CVSS:4.0/AV:N/AC:L/AT:N/PR:N/UI:N/VC:N/VI:N/VA:H/SC:N/SI:N/SA:N/E:P/S:N/AU:Y/R:U/V:D/RE:M/U:Amber`

#### References
- [https://github.com/juliangruber/brace-expansion/security/advisories/GHSA-3jxr-9vmj-r5cp](https://redirect.github.com/juliangruber/brace-expansion/security/advisories/GHSA-3jxr-9vmj-r5cp)
- [https://nvd.nist.gov/vuln/detail/CVE-2026-13149](https://nvd.nist.gov/vuln/detail/CVE-2026-13149)
- [https://github.com/juliangruber/brace-expansion/pull/122](https://redirect.github.com/juliangruber/brace-expansion/pull/122)
- [https://github.com/juliangruber/brace-expansion/pull/123](https://redirect.github.com/juliangruber/brace-expansion/pull/123)
- [https://github.com/juliangruber/brace-expansion/commit/835d6be91201122d9adffb0c0c8c094189ace265](https://redirect.github.com/juliangruber/brace-expansion/commit/835d6be91201122d9adffb0c0c8c094189ace265)
- [https://github.com/juliangruber/brace-expansion/commit/c7e33ec13ac1a684c116720843ce24e208611754](https://redirect.github.com/juliangruber/brace-expansion/commit/c7e33ec13ac1a684c116720843ce24e208611754)
- [https://github.com/juliangruber/brace-expansion/commit/d74e63030c012e3b7ae81657b8d665619cd51b95](https://redirect.github.com/juliangruber/brace-expansion/commit/d74e63030c012e3b7ae81657b8d665619cd51b95)
- [https://github.com/juliangruber/brace-expansion](https://redirect.github.com/juliangruber/brace-expansion)
- [https://github.com/juliangruber/brace-expansion/releases/tag/v1.1.16](https://redirect.github.com/juliangruber/brace-expansion/releases/tag/v1.1.16)
- [https://github.com/juliangruber/brace-expansion/releases/tag/v2.1.2](https://redirect.github.com/juliangruber/brace-expansion/releases/tag/v2.1.2)
- [https://github.com/juliangruber/brace-expansion/releases/tag/v5.0.7](https://redirect.github.com/juliangruber/brace-expansion/releases/tag/v5.0.7)
- [https://www.npmjs.com/package/brace-expansion](https://www.npmjs.com/package/brace-expansion)

This data is provided by [OSV](https://osv.dev/vulnerability/GHSA-3jxr-9vmj-r5cp) and the [GitHub Advisory Database](https://redirect.github.com/github/advisory-database) ([CC-BY 4.0](https://redirect.github.com/github/advisory-database/blob/main/LICENSE.md)).
</details>

---

### brace-expansion: DoS via unbounded expansion length causing an out-of-memory process crash
[CVE-2026-14257](https://nvd.nist.gov/vuln/detail/CVE-2026-14257) / [GHSA-mh99-v99m-4gvg](https://redirect.github.com/advisories/GHSA-mh99-v99m-4gvg)

<details>
<summary>More information</summary>

##### Summary

`expand()` bounds the *number* of results it produces (the `max` option,
`100_000` by default) but not their *length*. By chaining many brace groups,
an attacker keeps the result count under `max` while making every result grow
with the number of groups. Building `max` long results — plus the intermediate
arrays combined at each brace group — exhausts memory and crashes the Node
process with an **uncatchable** out-of-memory error. `try/catch` around
`expand()` does not help: the fatal error terminates the process.

A ~7.5 KB input (`'{a,b}'.repeat(1500)`) is enough to crash a default Node
process.

##### Details

For `N` chained brace groups such as `'{a,b}'.repeat(N)`:

- the result count is `2^N`, immediately capped at `max` (`100_000`), so the
  `max` protection appears to hold, but
- each result is `N` characters long, so the total output size is
  `max × N` characters, which grows without bound in `N`.

`expand_` combines each brace set with the fully-expanded tail:

```js
const post = m.post.length ? expand_(m.post, max, false) : ['']
...
for (let j = 0; j < N.length; j++) {
  for (let k = 0; k < post.length && expansions.length < max; k++) {
    const expansion = pre + N[j] + post[k]   // grows one group longer per level
    ...
    expansions.push(expansion)
  }
}
```

The loop guard `expansions.length < max` limits how many strings are built, but
nothing limits how long they get. Each recursion level materializes another
array of up to `max` strings, one character longer than the level below, and —
because V8 represents `pre + N[j] + post[k]` as a cons-string (rope) that
references `post[k]` — those intermediate strings stay reachable through the
whole chain. Memory therefore scales with `max × N`.

Measured on `5.0.7` (`'{a,b}'.repeat(N)`, default `max`):

| groups (N) | input bytes | result count | peak RSS |
|---|---|---|---|
| 20 | 100 | 100,000 | ~80 MB |
| 50 | 250 | 100,000 | ~214 MB |
| 100 | 500 | 100,000 | ~409 MB |
| 300 | 1,500 | 100,000 | ~1,148 MB |
| 1500 | 7,500 | — | **OOM crash** |

##### Proof of concept

```js
const { expand } = require('brace-expansion')

// ~7.5 KB input — crashes the process with a fatal, uncatchable OOM:
//   FATAL ERROR: ... JavaScript heap out of memory
try {
  expand('{a,b}'.repeat(1500))
} catch (e) {
  // never reached — the process is already dead
}
```

##### Impact

Any application that passes attacker-influenced strings to
`brace-expansion.expand()` — directly, or transitively via `minimatch` / `glob`
brace patterns — can be crashed by a small request. Because the failure is a
fatal V8 out-of-memory error rather than a thrown exception, it cannot be caught
and it takes down the whole worker/process, denying service.

##### Remediation

Upgrade to a patched release. The fix bounds the total number of characters a
single `expand()` call may accumulate (`EXPANSION_MAX_LENGTH`, default
`4_000_000`, configurable via a new `maxLength` option), applied inside the
output-building loops so intermediate arrays are bounded too. Once the limit is
reached, output is truncated — consistent with how `max` already truncates —
instead of growing without bound. The limit sits well above any realistic
expansion (100,000 results hitting `max` measure ~1M characters), so legitimate
input is unaffected.

After the fix, `'{a,b}'.repeat(1500)` returns a bounded, truncated result in
~0.7 s using ~340 MB and never crashes, including under a constrained 512 MB
heap.

The fix bounds memory but the algorithm still rebuilds intermediate arrays at
each level (roughly `O(N × maxLength)` work on this input class). A streaming
rewrite that produces output in `O(total output size)` can be a non-urgent
follow-up.

If immediate upgrade isn't possible, avoid passing untrusted input to
`expand()` / glob brace patterns, or pass a small explicit `max` **and**
`maxLength`.

#### Severity
- CVSS Score: 7.5 / 10 (High)
- Vector String: `CVSS:3.1/AV:N/AC:L/PR:N/UI:N/S:U/C:N/I:N/A:H`

#### References
- [https://github.com/juliangruber/brace-expansion/security/advisories/GHSA-mh99-v99m-4gvg](https://redirect.github.com/juliangruber/brace-expansion/security/advisories/GHSA-mh99-v99m-4gvg)
- [https://nvd.nist.gov/vuln/detail/CVE-2026-14257](https://nvd.nist.gov/vuln/detail/CVE-2026-14257)
- [https://github.com/juliangruber/brace-expansion/commit/a1bd33999ea75262c4749fff3bbb0d1372bd07b5](https://redirect.github.com/juliangruber/brace-expansion/commit/a1bd33999ea75262c4749fff3bbb0d1372bd07b5)
- [https://github.com/juliangruber/brace-expansion](https://redirect.github.com/juliangruber/brace-expansion)
- [https://www.npmjs.com/package/brace-expansion](https://www.npmjs.com/package/brace-expansion)

This data is provided by [OSV](https://osv.dev/vulnerability/GHSA-mh99-v99m-4gvg) and the [GitHub Advisory Database](https://redirect.github.com/github/advisory-database) ([CC-BY 4.0](https://redirect.github.com/github/advisory-database/blob/main/LICENSE.md)).
</details>

---

### js-yaml: YAML merge-key chains can force quadratic CPU consumption
[CVE-2026-59869](https://nvd.nist.gov/vuln/detail/CVE-2026-59869) / [GHSA-52cp-r559-cp3m](https://redirect.github.com/advisories/GHSA-52cp-r559-cp3m)

<details>
<summary>More information</summary>

##### Impact

js-yaml can spend quadratic CPU time parsing a document whose size grows only linearly. The issue is triggered by a chain of mappings where each mapping merges the previous one:

```yaml
a0: &a0 { k0: 0 }
a1: &a1 { <<: *a0, k1: 1 }
a2: &a2 { <<: *a1, k2: 2 }
a3: &a3 { <<: *a2, k3: 3 }
...
b: *aN
```

For each new mapping, the loader has to enumerate the keys inherited from the previous mapping. With N chained mappings, this results in roughly 1 + 2 + ... + N merged-key visits, i.e., O(N^2) work for O(N) input size.

##### PoC

From N = 4000 delay become > 1s (doc size < 100K)

```js
import { performance } from 'node:perf_hooks'
import { Buffer } from 'node:buffer'
import { load, YAML11_SCHEMA } from 'js-yaml'

const n = Number(process.argv[2] || 4000)

function makeMergeChain (count) {
  const lines = ['a0: &a0 { k0: 0 }']

  for (let i = 1; i < count; i++) {
    lines.push(`a${i}: &a${i} { <<: *a${i - 1}, k${i}: ${i} }`)
  }

  lines.push(`b: *a${count - 1}`)
  return `${lines.join('\n')}\n`
}

const source = makeMergeChain(n)

console.log(source.split('\n').slice(0, 8).join('\n'))
console.log('...')
console.log(source.split('\n').slice(-4).join('\n'))
console.log()
console.log(`N: ${n}`)
console.log(`YAML size: ${Buffer.byteLength(source)} bytes`)

const started = performance.now()
const result = load(source, { schema: YAML11_SCHEMA })
const elapsed = performance.now() - started

console.log(`parse time: ${elapsed.toFixed(1)} ms`)
console.log(`top-level keys: ${Object.keys(result).length}`)
console.log(`b keys: ${Object.keys(result.b).length}`)
```

##### Patches

Fix released. The most robust protection is to limit the total number of merged keys per parse call. This should close all past and future edge cases with merge. The default 10K-key limit should be okay in most cases.

#### Severity
- CVSS Score: 7.5 / 10 (High)
- Vector String: `CVSS:3.1/AV:N/AC:L/PR:N/UI:N/S:U/C:N/I:N/A:H`

#### References
- [https://github.com/nodeca/js-yaml/security/advisories/GHSA-52cp-r559-cp3m](https://redirect.github.com/nodeca/js-yaml/security/advisories/GHSA-52cp-r559-cp3m)
- [https://nvd.nist.gov/vuln/detail/CVE-2026-59869](https://nvd.nist.gov/vuln/detail/CVE-2026-59869)
- [https://github.com/nodeca/js-yaml/commit/24f13e79ee1343a7e30bd6f6c9d9cdbf0ac9b2b7](https://redirect.github.com/nodeca/js-yaml/commit/24f13e79ee1343a7e30bd6f6c9d9cdbf0ac9b2b7)
- [https://github.com/nodeca/js-yaml/commit/59423c6f8cdc78742ac00e25a4dd39ef16b702e4](https://redirect.github.com/nodeca/js-yaml/commit/59423c6f8cdc78742ac00e25a4dd39ef16b702e4)
- [https://github.com/nodeca/js-yaml/releases/tag/3.15.0](https://redirect.github.com/nodeca/js-yaml/releases/tag/3.15.0)
- [https://github.com/nodeca/js-yaml/releases/tag/4.3.0](https://redirect.github.com/nodeca/js-yaml/releases/tag/4.3.0)
- [https://github.com/advisories/GHSA-52cp-r559-cp3m](https://redirect.github.com/advisories/GHSA-52cp-r559-cp3m)

This data is provided by the [GitHub Advisory Database](https://redirect.github.com/advisories/GHSA-52cp-r559-cp3m) ([CC-BY 4.0](https://redirect.github.com/github/advisory-database/blob/main/LICENSE.md)).
</details>

---

### JS-YAML: Quadratic-complexity DoS in merge key handling via repeated aliases
[CVE-2026-53550](https://nvd.nist.gov/vuln/detail/CVE-2026-53550) / [GHSA-h67p-54hq-rp68](https://redirect.github.com/advisories/GHSA-h67p-54hq-rp68)

<details>
<summary>More information</summary>

##### Summary
A crafted YAML document can trigger algorithmic CPU exhaustion in `js-yaml` merge-key processing (`<<`) by repeating the same alias many times in a merge sequence.  
This causes quadratic parse-time behavior relative to input size and can block a Node.js worker/event loop for seconds with a relatively small payload (tens of KB), resulting in denial of service.

##### Details
The issue is in merge handling inside `lib/loader.js`:

- `storeMappingPair(...)` iterates every element of a merge sequence when key tag is `tag:yaml.org,2002:merge`.
- For each element, it calls `mergeMappings(...)`.
- `mergeMappings(...)` computes `Object.keys(source)` and performs `_hasOwnProperty.call(destination, key)` checks for each key.

When input is of the form:

a: &a {k0:0, k1:0, ..., kK:0}
b: {<<: [*a, *a, *a, ... repeated M times ...]}
all *a entries refer to the same anchored object. After the first merge, subsequent merges are semantically no-ops, but the parser still reprocesses all keys each time.
Resulting work is O(K * M), while input size is O(K + M), giving quadratic scaling as payload grows.
Relevant code path:
lib/loader.js in storeMappingPair(...) merge branch (keyTag === 'tag:yaml.org,2002:merge')
lib/loader.js mergeMappings(...)

##### Root cause
      lib/loader.js
Function:   storeMappingPair(state, _result, overridableKeys, keyTag, keyNode,
                             valueNode, startLine, startLineStart, startPos)
Lines:      ~359-366

    if (keyTag === 'tag:yaml.org,2002:merge') {
      if (Array.isArray(valueNode)) {
        for (index = 0, quantity = valueNode.length; index < quantity; index += 1) {
          mergeMappings(state, _result, valueNode[index], overridableKeys);
        }
      } else {
        mergeMappings(state, _result, valueNode, overridableKeys);
      }
    }

When the merge value is a sequence (YAML 1.1 <<: [ *a, *a, ... ]), each element
is handed to mergeMappings() without deduplication. mergeMappings() then does

    sourceKeys = Object.keys(source);
    for (index = 0; index < sourceKeys.length; index += 1) {
      key = sourceKeys[index];
      if (!_hasOwnProperty.call(destination, key)) {
        setProperty(destination, key, source[key]);
        overridableKeys[key] = true;
      }
    }

Every alias reference in the sequence resolves (by design) to the SAME object
via state.anchorMap. After the first merge, every subsequent merge of that same
reference is a pure no-op semantically, but still performs:

  * one Object.keys(source) call (O(K))
  * K _hasOwnProperty.call checks on the destination

Total: M * K hasOwnProperty checks + M Object.keys allocations, while the final
object and all observable side effects are identical to a single merge.

YAML semantics for `<<:` are idempotent and commutative over duplicate sources,
so collapsing duplicates preserves behavior exactly; this isn't a spec trade-off.

##### PoC
Environment:
js-yaml version: 4.1.1
Node.js: v24.5.0
Platform: arm64 macOS (reproduced consistently)
Reproduction script:
Create many keys in one anchored map (&a).
Merge that same alias repeatedly via <<: [*a, *a, ...].
Measure parse time and compare with control payload using single merge (<<: *a).
Observed repeated runs (same machine):
K=M=1000, input 9,909 bytes: ~33–36 ms
K=M=2000, input 20,909 bytes: ~121–123 ms
K=M=4000, input 42,909 bytes: ~524–537 ms
K=M=6000, input 64,909 bytes: ~1,608–1,829 ms
K=M=8000, input 86,909 bytes: ~3,395–3,565 ms
Control (single merge, similar key counts):
K=2000: ~1–2 ms
K=4000: ~3 ms
K=8000: ~5 ms
Also verified: repeated-merge output equals single-merge output (same key count and same JSON), confirming excess time is redundant computation.

##### Impact
This is a denial-of-service vulnerability (CPU exhaustion / algorithmic complexity).
Any service parsing untrusted YAML with js-yaml can be impacted, including API backends, CI tools, config processors, and automation services. An attacker can submit crafted YAML to significantly increase CPU time and reduce availability.

##### Suggested fix:
Dedupe the merge source list by reference before invoking mergeMappings. Any of
the following are minimal and preserve YAML 1.1 merge semantics:

dedupe in storeMappingPair:

    if (keyTag === 'tag:yaml.org,2002:merge') {
      if (Array.isArray(valueNode)) {
        var seen = new Set();
        for (index = 0, quantity = valueNode.length; index < quantity; index += 1) {
          var src = valueNode[index];
          if (seen.has(src)) continue;   // idempotent; skip redundant alias
          seen.add(src);
          mergeMappings(state, _result, src, overridableKeys);
        }
      } else {
        mergeMappings(state, _result, valueNode, overridableKeys);
      }
    }

#### Severity
- CVSS Score: 5.3 / 10 (Medium)
- Vector String: `CVSS:3.1/AV:N/AC:L/PR:N/UI:N/S:U/C:N/I:N/A:L`

#### References
- [https://github.com/nodeca/js-yaml/security/advisories/GHSA-h67p-54hq-rp68](https://redirect.github.com/nodeca/js-yaml/security/advisories/GHSA-h67p-54hq-rp68)
- [https://nvd.nist.gov/vuln/detail/CVE-2026-53550](https://nvd.nist.gov/vuln/detail/CVE-2026-53550)
- [https://github.com/advisories/GHSA-h67p-54hq-rp68](https://redirect.github.com/advisories/GHSA-h67p-54hq-rp68)

This data is provided by the [GitHub Advisory Database](https://redirect.github.com/advisories/GHSA-h67p-54hq-rp68) ([CC-BY 4.0](https://redirect.github.com/github/advisory-database/blob/main/LICENSE.md)).
</details>

---

### pnpm: Tarball hash of GitHub git dependencies is not stored in lockfile
[CVE-2026-48995](https://nvd.nist.gov/vuln/detail/CVE-2026-48995) / [GHSA-hg3w-7f8c-63hp](https://redirect.github.com/advisories/GHSA-hg3w-7f8c-63hp)

<details>
<summary>More information</summary>

##### Summary

A malicious `codeload.github.com` server can serve whatever tarball it wants and pnpm will install it regardless of the lockfile.

##### Details

The lockfile does not store the hash of the dependencies from https://codeload.github.com

This means that if this server was compromised or a person's machine configuration was compromised, pnpm would download and install these dependencies.

##### PoC

```sh
> pnpm -v     
10.28.2
```

Given the following package.json:

```json
{
  "dependencies": {
    "add": "git://github.com/dsherret/npm-git-dep.git#b3eeb9b"
  }
}
```

This produces a lockfile like so:

```yaml
lockfileVersion: '9.0'

settings:
  autoInstallPeers: true
  excludeLinksFromLockfile: false

importers:

  .:
    dependencies:
      add:
        specifier: git://github.com/dsherret/npm-git-dep.git#b3eeb9b
        version: https://codeload.github.com/dsherret/npm-git-dep/tar.gz/b3eeb9b

packages:

  add@https://codeload.github.com/dsherret/npm-git-dep/tar.gz/b3eeb9b:
    resolution: {tarball: https://codeload.github.com/dsherret/npm-git-dep/tar.gz/b3eeb9b}
    version: 1.0.0

snapshots:

  add@https://codeload.github.com/dsherret/npm-git-dep/tar.gz/b3eeb9b: {}
```

Notice that there is no hash. The `b3eeb9b` is not sufficient because I can configure my machine to resolve a compromised tarball from that url (I tested it out and pnpm just installs it).

##### Impact

Anyone relying on github git dependencies.

#### Severity
- CVSS Score: 4.8 / 10 (Medium)
- Vector String: `CVSS:4.0/AV:N/AC:H/AT:N/PR:N/UI:A/VC:H/VI:H/VA:H/SC:N/SI:N/SA:N/E:U`

#### References
- [https://github.com/pnpm/pnpm/security/advisories/GHSA-hg3w-7f8c-63hp](https://redirect.github.com/pnpm/pnpm/security/advisories/GHSA-hg3w-7f8c-63hp)
- [https://nvd.nist.gov/vuln/detail/CVE-2026-48995](https://nvd.nist.gov/vuln/detail/CVE-2026-48995)
- [https://github.com/pnpm/pnpm](https://redirect.github.com/pnpm/pnpm)

This data is provided by [OSV](https://osv.dev/vulnerability/GHSA-hg3w-7f8c-63hp) and the [GitHub Advisory Database](https://redirect.github.com/github/advisory-database) ([CC-BY 4.0](https://redirect.github.com/github/advisory-database/blob/main/LICENSE.md)).
</details>

---

### pnpm: Unsafe default behavior breaks integrity check
[CVE-2026-50573](https://nvd.nist.gov/vuln/detail/CVE-2026-50573) / [GHSA-54hh-g5mx-jqcp](https://redirect.github.com/advisories/GHSA-54hh-g5mx-jqcp)

<details>
<summary>More information</summary>

#### Details
While it is unclear whether this should be classified as a vulnerability, it is being reported through this channel because the current behavior may represent an unsafe default.

##### Summary

`pnpm install` in non-frozen mode can accept new remote package content after detecting that the downloaded tarball does not match the integrity recorded in `pnpm-lock.yaml`.

When a package is already locked with an `integrity` value, and the registry later serves different metadata and tarball content for the same package name and version, pnpm initially reports an integrity mismatch. However, plain `pnpm install` then performs a resolution repair, accepts the registry's new integrity, updates the lockfile, installs the new content, and exits successfully.

This means the lockfile integrity check does not act as a hard stop by default.

##### Reproduction Scenario

1. Run a local npm-compatible registry.
2. Publish or serve `example-package@1.0.0` with tarball content `v1`.
3. Install it with pnpm:

```bash
pnpm add example-package@1.0.0 --registry=http://127.0.0.1:48741
```

4. Confirm `pnpm-lock.yaml` contains the `v1` integrity:

```yaml
packages:
  example-package@1.0.0:
    resolution:
      integrity: sha512-...v1...
```

5. Change the registry metadata and tarball for the same `example-package@1.0.0` to content `v2`.
6. On a clean store/cache, run:

```bash
pnpm install --registry=http://127.0.0.1:48741
```

##### Observed Behavior

pnpm detects the checksum mismatch:

```text
WARN Got unexpected checksum for "http://127.0.0.1:48741/example-package/-/example-package-1.0.0.tgz".
Wanted "sha512-...v1..."
Got "sha512-...v2...".

ERR_PNPM_TARBALL_INTEGRITY The lockfile is broken! Resolution step will be performed to fix it.
```

However, the install still succeeds:

```text
INSTALL_RC=0
INSTALLED=v2-replaced
```

The lockfile is then rewritten to trust the new remote integrity:

```yaml
packages:
  example-package@1.0.0:
    resolution:
      integrity: sha512-...v2...
```

##### Expected Behavior

If a downloaded tarball does not match the integrity recorded in `pnpm-lock.yaml`, the install should fail by default.

The lockfile integrity should be treated as authoritative unless the user explicitly requests lockfile repair or dependency update behavior.

##### Security Impact

This behavior weakens the protection normally expected from a committed lockfile.

If a registry is compromised and an attacker overwrites the metadata and tarball for an existing package version, a new environment without the old pnpm store/cache may install the attacker's replacement package even though the project already has a lockfile with the original integrity.

Examples of affected new or clean environments include:

- an engineer setting up the project on a new machine
- a new team member onboarding to the project

In this situation, pnpm first detects that the downloaded tarball does not match the integrity stored in `pnpm-lock.yaml`. However, instead of failing by default, plain `pnpm install` performs a resolution repair, trusts the current remote registry metadata, updates the lockfile to the new integrity, and installs the new registry content.

In other words, when the lockfile and registry disagree, the default non-frozen behavior can end up trusting the remote registry over the content previously recorded in the lockfile.

This is especially relevant for:

- private registries that allow overwriting or republishing the same version
- registry mirrors or proxies that can serve changed metadata and tarballs
- compromised public or private registries
- compromised registry proxy infrastructure

The behavior is also surprising because the command reports an integrity error but still exits successfully after resolution repair.

This issue does not occur when `--frozen-lockfile` is enabled. In frozen mode, the same integrity mismatch fails the install and does not install the changed package content.

However, since the lockfile already records an integrity value, the integrity for the same package version should normally not change. If it does change, one likely explanation is that the server or registry has been compromised or is serving mutated package content. Under normal package publishing workflows, changed package content should be published as a new version instead of replacing an existing version.

For that reason, it may be safer for pnpm's default behavior to be closer to frozen mode for this specific case. At minimum, pnpm should not automatically repair the lockfile and trust the registry after an integrity mismatch. It should fail and let the user explicitly decide whether to discard the locked integrity, re-resolve the package from the remote registry, and update the lockfile.

##### Comparison

In the same scenario, `npm install` with an existing `package-lock.json` fails with `EINTEGRITY` and does not install the changed tarball.

`pnpm install --frozen-lockfile` also fails as expected:

```text
ERR_PNPM_TARBALL_INTEGRITY
```

The issue is specific to the default non-frozen behavior of plain `pnpm install` in non-CI environment.

#### Severity
- CVSS Score: 6.8 / 10 (Medium)
- Vector String: `CVSS:3.1/AV:N/AC:H/PR:N/UI:R/S:U/C:H/I:H/A:N`

#### References
- [https://github.com/pnpm/pnpm/security/advisories/GHSA-54hh-g5mx-jqcp](https://redirect.github.com/pnpm/pnpm/security/advisories/GHSA-54hh-g5mx-jqcp)
- [https://nvd.nist.gov/vuln/detail/CVE-2026-50573](https://nvd.nist.gov/vuln/detail/CVE-2026-50573)
- [https://github.com/pnpm/pnpm](https://redirect.github.com/pnpm/pnpm)

This data is provided by [OSV](https://osv.dev/vulnerability/GHSA-54hh-g5mx-jqcp) and the [GitHub Advisory Database](https://redirect.github.com/github/advisory-database) ([CC-BY 4.0](https://redirect.github.com/github/advisory-database/blob/main/LICENSE.md)).
</details>

---

### pnpm binds unscoped user-level npm auth credentials to a repository-selected registry
[CVE-2026-50017](https://nvd.nist.gov/vuln/detail/CVE-2026-50017) / [GHSA-cjhr-43r9-cfmw](https://redirect.github.com/advisories/GHSA-cjhr-43r9-cfmw)

<details>
<summary>More information</summary>

##### Summary

pnpm can send user-level unscoped npm authentication credentials to a registry chosen by a repository-local `.npmrc` file.

In the reproduced case, the user's npm config contains a default registry and an unscoped `_authToken`. The repository does not provide a token-bearing auth line. It only sets `registry=` to a different registry URL. During normal pnpm metadata/install workflows, pnpm binds the user-origin unscoped credential to the repository-selected registry and sends it as an `Authorization` header.

This was reproduced with fake credentials and loopback registries only. No third-party registry or real token was used.

##### Affected Behavior Observed

Observed affected:

- pnpm `10.33.2`: `pnpm install --ignore-scripts` sends the user-level unscoped `_authToken` to the repository-selected registry.
- pnpm `11.1.3`: `pnpm install --ignore-scripts` sends the user-level unscoped `_authToken` to the repository-selected registry.
- pnpm `11.2.1` (`next-11` dist tag at testing time): `pnpm install --ignore-scripts` sends the user-level unscoped `_authToken` to the repository-selected registry.
- pnpm `11.1.3`: `pnpm view` also sends user-level unscoped `_authToken`, `_auth`, and `username` / `_password` credentials to the repository-selected registry in the local loopback replay.

Control:

- npm `10.9.7` rejects the same unscoped user `_authToken` configuration with `ERR_INVALID_AUTH` and does not send an `Authorization` header to the  repository-selected registry.
- URL-scoped registry token controls held in the local loopback replay: tokens scoped to the trusted registry URL were not sent to the attacker registry.

##### Threat Model

Victim:

- developer or CI job with user-level npm registry credentials configured;
- runs `pnpm install`, `pnpm view`, or an equivalent pnpm metadata/restore command in a repository.

Attacker:

- controls repository-local package manager configuration, such as `.npmrc`;
- can set `registry=` to a registry endpoint they control;
- does not need to provide a token-bearing auth line for the strong case.

Boundary:

Credentials from a higher-trust user configuration should not be rebound to a lower-trust repository-selected registry unless the credential is explicitly scoped to that registry.

##### Minimal Reproduction

The reproducer below starts two loopback HTTP registries:

- a trusted registry URL used in the isolated user `.npmrc`;
- an attacker registry URL used in the repository-local `.npmrc`.

The isolated user `.npmrc` contains:

```ini
registry=<trusted-loopback-registry>
_authToken=PR166_FAKE_REGISTRY_TOKEN
```

The repository-local `.npmrc` contains:

```ini
registry=<attacker-loopback-registry>
```

The repository `package.json` depends on a toy package served by the loopback registry. The script then runs:

```text
pnpm install --ignore-scripts
npm install --ignore-scripts
```

##### Expected Safe Behavior

pnpm should not send the user-level unscoped `_authToken` to the repository-selected registry. A safe behavior would be to reject or ignore the unscoped credential in this lower-trust registry-rebinding situation and require the credential to be URL-scoped to the selected registry.

##### Observed Behavior

pnpm `10.33.2`, pnpm `11.1.3`, and pnpm `11.2.1` send:

```http
Authorization: Bearer PR166_FAKE_REGISTRY_TOKEN
```

to the attacker loopback registry during install. npm `10.9.7` rejects the same config and sends no `Authorization` header.

##### Security Impact

This can disclose npm registry credentials from user-level configuration to a registry endpoint selected by an untrusted repository. The leak occurs before package lifecycle scripts run and does not depend on package code execution.

##### Non-Claims

This report does not claim:

- remote code execution;
- registry account compromise by itself;
- leakage of URL-scoped tokens for a different registry;
- npm CLI impact;
- impact from a repository explicitly committing its own token-bearing auth
  line.

##### Source-Level Notes

In pnpm's config/auth-header flow, unscoped/default credentials are parsed from the merged auth config and stored as default credentials. The auth-header logic then maps those default credentials to the effective default registry. Because repository-local `.npmrc` can change the effective default registry, higher-trust default credentials can be applied to a lower-trust registry choice.

##### Suggested Fix Direction

The conservative fix direction is to reject or contain unscoped/default auth credentials when a lower-trust workspace/repository config changes the default registry. A compatibility-preserving fix could track the source layer of both the default registry and the default credentials, then only bind default credentials to a registry selected by the same or higher-trust source. A stricter npm-compatible fix would reject unscoped auth and require URL-scoped
credentials.

This needs maintainer semantic review and compatibility control because some legacy workflows may intentionally rely on default/unscoped auth.

##### Runnable Reproducer

Save the following as `repro.py` and run it with Python 3 in an environment with pnpm and npm available. To force a specific pnpm version through Corepack, set `PR166_PNPM_SPEC`, for example `PR166_PNPM_SPEC=11.2.1`.

```python
import base64
import contextlib
import hashlib
import http.server
import io
import json
import os
import shutil
import subprocess
import sys
import tarfile
import tempfile
import threading
from pathlib import Path

"""Standalone loopback reproducer.

It creates only temporary directories and loopback HTTP servers. Cleanup is handled by TemporaryDirectory context managers and registry shutdown handlers; no persistent state is expected outside the package-manager cache directories inside the temporary home. Non-claims: this does not use real credentials, third-party registries, package scripts, or remote services. Failure paths return exit 1 or exit 2 through sys.exit(main()).
"""

TOKEN = "PR166_FAKE_REGISTRY_TOKEN"
PACKAGE_TGZ = None

class RegistryHandler(http.server.BaseHTTPRequestHandler):
    requests = []

    def do_GET(self):
        self.requests.append(
            {
                "method": self.command,
                "path": self.path,
                "authorization": self.headers.get("Authorization"),
            }
        )
        if self.path.endswith(".tgz"):
            payload = make_package_tgz()
            self.send_response(200)
            self.send_header("Content-Type", "application/octet-stream")
            self.send_header("Content-Length", str(len(payload)))
            self.end_headers()
            self.wfile.write(payload)
            return

        payload = make_package_tgz()
        body = json.dumps(
            {
                "name": "@&#8203;private/probe",
                "dist-tags": {"latest": "1.0.0"},
                "versions": {
                    "1.0.0": {
                        "name": "@&#8203;private/probe",
                        "version": "1.0.0",
                        "dist": {
                            "tarball": f"http://127.0.0.1:{self.server.server_port}/private/@&#8203;private/probe/-/probe-1.0.0.tgz",
                            "shasum": hashlib.sha1(payload).hexdigest(),
                            "integrity": "sha512-"
                            + base64.b64encode(hashlib.sha512(payload).digest()).decode("ascii"),
                        },
                    }
                },
            }
        ).encode("utf-8")
        self.send_response(200)
        self.send_header("Content-Type", "application/json")
        self.send_header("Content-Length", str(len(body)))
        self.end_headers()
        self.wfile.write(body)

    def log_message(self, fmt, *args):
        return

@&#8203;contextlib.contextmanager
def registry():
    handler = type("RecordingRegistryHandler", (RegistryHandler,), {"requests": []})
    server = http.server.ThreadingHTTPServer(("127.0.0.1", 0), handler)
    thread = threading.Thread(target=server.serve_forever, daemon=True)
    thread.start()
    try:
        yield server, handler.requests
    finally:
        server.shutdown()
        thread.join(timeout=5)
        server.server_close()

def make_package_tgz():
    global PACKAGE_TGZ
    if PACKAGE_TGZ is not None:
        return PACKAGE_TGZ
    bio = io.BytesIO()
    with tarfile.open(fileobj=bio, mode="w:gz") as tf:
        data = b'{"name":"@&#8203;private/probe","version":"1.0.0"}\n'
        info = tarfile.TarInfo("package/package.json")
        info.size = len(data)
        tf.addfile(info, io.BytesIO(data))
    PACKAGE_TGZ = bio.getvalue()
    return PACKAGE_TGZ

def write_text(path, text):
    path.parent.mkdir(parents=True, exist_ok=True)
    path.write_text(text, encoding="utf-8", newline="\n")

def run_install(tool, trusted_url, attacker_url):
    exe = shutil.which(tool)
    if exe is None:
        return {"tool": tool, "error": "missing"}
    cmd = [exe, "install", "--ignore-scripts"]
    if tool == "pnpm" and os.environ.get("PR166_PNPM_SPEC"):
        corepack = shutil.which("corepack")
        if corepack is None:
            return {"tool": tool, "error": "corepack missing"}
        cmd = [corepack, f"pnpm@{os.environ['PR166_PNPM_SPEC']}", "install", "--ignore-scripts"]

    with tempfile.TemporaryDirectory(prefix=f"pr166-min-{tool}-") as td:
        root = Path(td)
        home = root / "home"
        project = root / "project"
        home.mkdir()
        project.mkdir()
        userconfig = home / ".npmrc"

        write_text(userconfig, f"registry={trusted_url}\n_authToken={TOKEN}\n")
        write_text(project / ".npmrc", f"registry={attacker_url}\n")
        write_text(
            project / "package.json",
            '{"name":"pr166-probe","version":"1.0.0","dependencies":{"@&#8203;private/probe":"1.0.0"}}\n',
        )

        env = os.environ.copy()
        env.update(
            {
                "HOME": str(home),
                "USERPROFILE": str(home),
                "NPM_CONFIG_USERCONFIG": str(userconfig),
                "npm_config_userconfig": str(userconfig),
                "NPM_CONFIG_CACHE": str(home / "cache"),
                "npm_config_cache": str(home / "cache"),
                "NPM_CONFIG_STORE_DIR": str(home / "store"),
                "npm_config_store_dir": str(home / "store"),
                "XDG_CACHE_HOME": str(home / "xdg-cache"),
                "XDG_DATA_HOME": str(home / "xdg-data"),
                "NO_COLOR": "1",
            }
        )

        proc = subprocess.run(
            cmd,
            cwd=str(project),
            env=env,
            text=True,
            encoding="utf-8",
            errors="replace",
            stdout=subprocess.PIPE,
            stderr=subprocess.STDOUT,
            timeout=60,
        )
        return {"tool": tool, "returncode": proc.returncode, "output_tail": proc.stdout[-2000:]}

def summarize(tool, result, attacker_requests):
    auth_hits = [r for r in attacker_requests if r.get("authorization")]
    return {
        "tool": tool,
        "result": result,
        "attacker_auth_hits": auth_hits,
        "attacker_request_count": len(attacker_requests),
    }

def tool_version(tool):
    exe = shutil.which(tool)
    if exe is None:
        return "missing"
    cmd = [exe, "--version"]
    if tool == "pnpm" and os.environ.get("PR166_PNPM_SPEC"):
        corepack = shutil.which("corepack")
        if corepack is None:
            return "corepack missing"
        cmd = [corepack, f"pnpm@{os.environ['PR166_PNPM_SPEC']}", "--version"]
    proc = subprocess.

> ✂ **Note**
> 
> PR body was truncated to here.

 - [ ] <!-- rebase-check -->If you want to rebase/retry this PR, check this box

### CODE REVIEW

**Comment by hamchowderr** [comment]
Reporting a downstream experience with the `brace-expansion` part of this, in case the data is useful.

We build on the Mastra CLI and hit the same advisory. Here's what we observed and what we tried.

### The flagged path in our tree on `mastra@1.20.2`

```
mastra@1.20.2 → serve-handler@6.1.7 → minimatch@3.1.5 → brace-expansion@1.1.16
```

### What we tried

We attempted a `brace-expansion` override to `5.0.8` in our own project and backed it out, because `5.0.8` changed its export shape:

```js
// brace-expansion@2.1.2
require('brace-expansion')             // → function

// brace-expansion@5.0.8
require('brace-expansion')             // → { EXPANSION_MAX, EXPANSION_MAX_LENGTH, expand }
require('brace-expansion')('a{b,c}d')  // → TypeError: ... is not a function
```

`minimatch@3.1.5` (which `serve-handler@6.1.7` pins exactly) does `const expand = require('brace-expansion')` and then calls `expand(...)`, so it throws under the override. We saw the same shape change in `minimatch@10` — `require('minimatch')` returns `{ minimatch, sep, GLOBSTAR, ... }` rather than a callable — so forcing that didn't work for us either.

Both were verified by installing the packages in a scratch directory and inspecting the resolved exports.

### What we found upstream of `serve-handler`

- `serve-handler@6.1.7` is the latest published release and pins `minimatch: 3.1.5` exactly.
- `serve@14.2.6` is the latest published release and pins `serve-handler: 6.1.7` exactly.
- Neither `vercel/serve` nor `vercel/serve-handler` has an open issue referencing this.

### Where we landed

We left the advisory in place rather than override it. In our case the affected packages are dev/build dependencies only and aren't in our production dependency tree, so we accepted it.

Passing this along purely as a downstream report — we haven't looked at what's failing in this PR's CI and aren't suggesting a direction.

daneatmastra:
### ⚠️ Artifact update problem

Renovate failed to update an artifact related to this branch. You probably do not want to merge this PR as-is.

♻ Renovate will retry this branch, including artifacts, only when one of the following happens:

 - any of the package files in this branch needs updating, or 
 - the branch becomes conflicted, or
 - you click the rebase/retry checkbox if found above, or
 - you rename this PR's title to start with "rebase!" to trigger it manually

The artifact failure details are included below:

##### File name: packages/playground/vercel-preview/package.json

```
Command failed: corepack use pnpm@11.5.3
[WARN] The "pnpm" field in package.json is no longer read by pnpm. The following keys were ignored: "pnpm.overrides". See https://pnpm.io/settings for the new home of each setting.

```

daneatmastra:
### Edited/Blocked Notification

Renovate will not automatically rebase this PR, because it does not recognize the last commit author and assumes somebody else may have edited the PR.

You can manually request rebase by checking the rebase/retry box above.

 ⚠️ **Warning**: custom changes will be lost.

PR #19274, opened 2026-07-10
fix(schema-compat): strip propertyNames from OpenAI schemas

### LINKED ISSUE #19273: [BUG] OpenAI schema-compat leaks propertyNames (z.record), rejected by Structured Outputs strict mode
### Describe the Bug

The OpenAI schema compatibility layer leaks the `propertyNames` JSON Schema keyword, which OpenAI Structured Outputs strict mode rejects. Any tool or structured-output schema that uses `z.record(...)` produces a schema OpenAI refuses with an "Invalid schema ... 'propertyNames' is not permitted" error, so the whole request fails.

`packages/schema-compat/src/provider-compats/openai.ts`, `postProcessJSONNode` forces `additionalProperties: false` on object nodes (as strict mode requires) but never removes `propertyNames`:

```ts
if (isObjectSchema(schema)) {
  schema.additionalProperties = false;
  // propertyNames is never deleted
  ...
}
```

`z.record(z.string(), z.string())` emits `propertyNames: { type: 'string' }`, which survives into the final schema handed to the OpenAI API.

The Google layer already strips this keyword for the same reason (`google.ts`: `delete s['propertyNames']`, with a test in `google.test.ts` asserting it is gone). OpenAI, which is also a restricted/strict provider, has no equivalent strip. The OpenAI reasoning layer inherits the same `postProcessJSONNode`, so it is affected too.

### Steps To Reproduce

Convert a record schema through the OpenAI compat layer:

```ts
import { z } from 'zod';

const schema = z.object({ metadata: z.record(z.string(), z.string()) });
// OpenAI layer output still contains "propertyNames":
// { type: 'object', properties: { metadata: { type: 'object', propertyNames: { type: 'string' }, additionalProperties: false, required: [] } }, ... }
```

Then use that schema as a tool input schema or structured output with a `gpt-4o` (or reasoning) model. The request fails with an invalid-schema error.

### Expected Behavior

`propertyNames` is stripped (as the Google layer does), so the schema is strict-mode legal and the request succeeds. Record fields work with OpenAI.

### Actual Behavior

`propertyNames` leaks into the schema and OpenAI Structured Outputs strict mode rejects the whole request.

### Additional Context

The fix is one line, mirroring the Google layer: `delete schema['propertyNames']` in the object branch of the OpenAI `postProcessJSONNode`. I have a PR ready with a regression test that asserts the emitted schema for a `z.record` field no longer contains `propertyNames`, for both the standard and reasoning layers.

abhay-codes07:
## What

The OpenAI schema compatibility layer leaked the `propertyNames` JSON Schema keyword, which OpenAI Structured Outputs strict mode rejects. Any tool or structured-output schema that used `z.record(...)` produced a schema OpenAI refused with an "Invalid schema ... 'propertyNames' is not permitted" error, failing the whole request.

`postProcessJSONNode` forced `additionalProperties: false` on object nodes (as strict mode requires) but never removed `propertyNames`:

```ts
if (isObjectSchema(schema)) {
  schema.additionalProperties = false;
  // propertyNames was never deleted
  ...
}
```

`z.record(z.string(), z.string())` emits `propertyNames: { type: 'string' }`, which survived into the final schema. The Google layer already strips this exact keyword for the same reason (`google.ts`: `delete s['propertyNames']`, with a test locking it in), and OpenAI reasoning models inherit this same handler, so they were affected too.

## Fix

Strip `propertyNames` in the object branch of the OpenAI `postProcessJSONNode`, mirroring the Google layer. One line.

## Why it was not caught

The OpenAI suite has no `z.record` / `propertyNames` assertion (only Google does), and the OpenAI e2e tests do not feed a record schema, so the rejected keyword never reached a strict request in CI.

## Tests

Added two assertions in `openai.test.ts` (standard and reasoning layers): convert `z.object({ metadata: z.record(z.string(), z.string()) })` and assert the emitted JSON no longer contains `propertyNames`. Both fail on the current code (the schema emits `"propertyNames":{"type":"string"}`) and pass with the fix. The full schema-compat suite stays green (824 passed) with no snapshot changes.

Closes #19273

## ELI5

OpenAI does not understand one part of the JSON format produced for `z.record(...)` schemas. This change converts those records into compatible shapes, then restores them during validation so tools and structured outputs continue to work correctly.

## What changed

- Removed unsupported `propertyNames` from OpenAI schemas, including strict-mode, nested, optional, root-level, and reasoning-model paths.
- Converted enum-keyed records into closed objects.
- Converted arbitrary string-key records into key/value arrays and folded them back into records during validation.
- Added standard and reasoning-layer regression and end-to-end coverage.
- Refreshed OpenAI recordings and added a package changeset.


PiedPiper911:
Thanks for the fix @abhay-codes07! The `postProcessJSONNode` change in `openai.ts` correctly covers the compat layer path (used for tool schemas and structured output via `processToCompatSchema`/`processToJSONSchema`).

There's a second path where `propertyNames` can also leak through: `prepareJsonSchemaForOpenAIStrictMode()` in `packages/schema-compat/src/zod-to-json.ts` (called from `packages/core/src/stream/aisdk/v5/execute.ts` line ~180). This function runs `ensureAllPropertiesRequired()` and `ensureAdditionalPropertiesFalse()` — neither strips `propertyNames`. So if a user's schema hits the strict-mode path via `execute.ts` (rather than the compat layer), the keyword still reaches OpenAI.

The same `delete result['propertyNames']` could be added inside `ensureAllPropertiesRequired()` (since it already recurses into every node) or as a third pass in `prepareJsonSchemaForOpenAIStrictMode()`.

Would you like to add that to this PR, or should I open a follow-up? Happy to help either way.

abhiaiyer91:
🤖 Comment from Mastra Code (on behalf of the reviewing maintainer)

Thanks for the fix! While reviewing we found that stripping `propertyNames` alone had a hidden side effect: the layer already forces `additionalProperties: false` on every object, so after the strip a `z.record(...)` field became `{ type: 'object', additionalProperties: false, required: [] }` — an object that admits no keys. Requests stopped erroring, but every record field silently came back as `{}`.

We pushed a follow-up commit (2003f1a753) onto this branch that keeps your strip as the fallback and adds a proper rewrite so record semantics survive OpenAI strict mode:

- **Enum keys** (`z.record(z.enum([...]), V)`): expanded losslessly into a closed object with one property per key.
- **Arbitrary string keys** (`z.record(z.string(), V)`): the model is asked for an array of `{ key, value }` pairs (tagged `x-record`), which the validation layer folds back into a plain object — tools and structured outputs still receive a real `Record<string, V>`.
- **Root-level records** and other inexpressible cases: fall back to your `propertyNames` strip, mirroring the Google layer.

Also included: `parentSchema` forwarded to the `pre/postProcessJSONNode` hooks in the base class (needed to distinguish root vs. nested records), 9 additional tests (pair shape, fold-back, nested records, enum keys, optional, zod v3, reasoning layer, root fallback), and an updated changeset.

Verified locally: schema-compat suite 833 passing (both zod v3/v4 projects, no snapshot changes), `tsc --noEmit` clean, and downstream `packages/core` tool-builder tests passing.

abhiaiyer91:
🤖 Comment from Mastra Code (on behalf of the reviewing maintainer)

@PiedPiper911 Good catch — confirmed and fixed in 723b8b6a0c. `prepareJsonSchemaForOpenAIStrictMode()` now deletes `propertyNames` in the same pass that sets `additionalProperties: false`, so the `execute.ts` strict-mode path no longer leaks it either.

One note on why this path gets the strip rather than the key/value-pair rewrite used in the compat layer: there's no validation layer between the model response and the caller on this path, so there'd be nothing to fold a pair array back into a record — the strip is the only safe behavior here.

Added 3 tests in `zod-to-json.test.ts` (raw JSON Schema, `z.record`, nested/anyOf) that fail without the fix and run under both the zod v3 and v4 test projects. Full schema-compat suite: 846 passing, tsc clean.

abhay-codes07:
Thanks for the thorough review and for pushing the follow-ups directly onto the branch.

The empty-record side effect is a real catch. I went with the minimal strip and missed that forcing additionalProperties: false on a stripped record leaves an object that cannot hold any keys, so record fields would have silently come back as {}. The key/value-pair rewrite for arbitrary string keys, plus the lossless enum-key expansion, is a much better answer than just dropping the constraint, and keeping the strip as the root-level fallback makes sense given there is no validation layer to fold a pair array back on that path.

Good catch on the second leak path through prepareJsonSchemaForOpenAIStrictMode as well. I had only traced the compat-layer path via processToCompatSchema/processToJSONSchema and did not check the strict-mode helper that execute.ts calls directly. Agreed that the strip is the only safe behavior there, since nothing folds the model response back into a record on that path.

Happy with where the branch landed. Thanks for taking it the rest of the way to a complete fix.

abhiaiyer91:
Closing because the linked issue #19273 is closed and was resolved by merged PR #20977. Thanks for the contribution.

PR #19295, opened 2026-07-11
feat(behaviors): add durable behavior state trees

TylerBarnes:
## Summary

- add the reusable `@mastra/behaviors` package
- normalize TypeScript and filesystem behavior definitions into one validated immutable graph
- provide authoritative in-memory and LibSQL runtime stores with atomic per-thread updates
- add guarded and judged transitions, stale-result protection, migrations, state projection, and thread-state mirroring
- document behavior authoring and public runtime APIs

## Test plan

- `pnpm --filter @mastra/behaviors exec vitest run --reporter=dot --bail 1` — 18 passed
- `pnpm --filter @mastra/behaviors check`
- `pnpm --filter @mastra/behaviors build:lib`
- `pnpm validate` from `docs/` — 566 files passed
- scoped Remark and Vale checks for both behavior docs
- `git diff --check feat/behavgoal-plugin-signals...HEAD`

## Scope

This stacked PR contains the reusable runtime only. Intent/tool enforcement and the Mastra Code adapter follow in the next PR. No `packages/core` files changed.

### DIFF
--- a/.changeset/shaggy-mirrors-do.md
+++ b/.changeset/shaggy-mirrors-do.md
@@ -1,0 +1,5 @@
+---
+'@mastra/behaviors': minor
+---
+
+Added the @mastra/behaviors package for defining durable agent state trees from TypeScript or filesystem manifests. Includes transactional in-memory and LibSQL stores, guarded and judged transitions, migration safety, and state-signal projection.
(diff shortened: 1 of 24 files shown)


TylerBarnes:
## State-snapshot correction

The realistic Studio proof exposed that the behavior signal cache key included the internal runtime revision. Periodic bookkeeping and same-state transitions could therefore emit duplicate full snapshots even when the model-facing behavior projection was unchanged.

The cache key now reflects only model-facing identity: definition version, active state, and current intent. A regression test increments only the runtime revision and confirms no new snapshot is emitted. Compaction still causes a fresh snapshot when the prior base leaves the context window.

Verification: 5 behavior test files / 32 tests passed; package check and build passed.

TylerBarnes:
Closing after extracting the behavior-specific runtime, integration, documentation, and proofs into the standalone private behaviors repository. The OSS stack will retain only the generic plugin signal-provider contract in #19294.

PR #19296, opened 2026-07-11
feat(behaviors): enforce intent and integrate Mastra Code

TylerBarnes:
## Summary\n\n- enforce behavior intent before and during governed tool execution while preserving tool names, schemas, and wrapper identity across transitions\n- route state-specific models, acting instructions, and skills through host resolvers without exposing judge instructions\n- add stable behavior control tools and a thin Mastra Code plugin adapter backed by the shared provider\n- cover ordinary-provider behavior, plugin loading, and the real Mastra Code TUI path\n\n## Test plan\n\n- pnpm --filter @mastra/behaviors exec vitest run --reporter=dot --bail 1\n- pnpm --filter @mastra/behaviors check\n- pnpm build:mastracode\n- pnpm --filter ./mastracode/sdk exec vitest run src/plugins --reporter=dot --bail 1\n- MC_E2E_VITEST_SCENARIOS=behaviors pnpm --filter ./mastracode/tui exec vitest run --config e2e/vitest.config.ts --reporter=dot --bail 1\n- pnpm validate (from docs/)\n- live proof transcript ends with PROOF: GREEN\n

### DIFF
--- a/.changeset/shy-behaviors-dream.md
+++ b/.changeset/shy-behaviors-dream.md
@@ -1,0 +1,6 @@
+---
+'@mastra/behaviors': minor
+'@mastra/code-sdk': patch
+---
+
+Add transition-stable intent enforcement, state-specific model and instruction routing, behavior control tools, and a thin Mastra Code behavior plugin adapter.
--- a/docs/src/content/en/docs/long-running-agents/behaviors.mdx
+++ b/docs/src/content/en/docs/long-running-agents/behaviors.mdx
@@ -124,3 +124,24 @@ The behavior store is authoritative. The provider can mirror committed records i
 Deterministic guards run before the optional judge. A failed guard prevents the judge call. The engine commits a judged transition only when the runtime revision still matches the revision that the judge reviewed.
 
 `JUDGE.md` and `judgeInstructions` are sent only to the judge. They aren't included in the acting agent's state signal.
+
+## Require intent for tools
+
+The provider adds an `intent` field to each governed tool once when it attaches to the agent. Tool names, schemas, and wrapper identities stay stable when the behavior changes state, preserving prompt-cache compatibility.
+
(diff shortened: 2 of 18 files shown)


TylerBarnes:
## Studio integration correction

The original Studio proof exposed a real host integration bug: standard memory-backed Agent runs provide thread identity to `computeStateSignal`, but not as a custom `requestContext.get('threadId')` value. Behavior tools therefore could not select a behavior in Studio.

This branch now carries the memory thread identity observed by the state processor through the provider's request-scoped context and reuses it for behavior tools, routing, and enforcement. A regression test proves both `behavior_select` and the resulting `current-behavior` snapshot in the ordinary Agent path.

Verification:
- `pnpm --filter @mastra/behaviors exec vitest run --reporter=dot --bail 1` — 5 files, 32 tests passed
- `pnpm --filter @mastra/behaviors check` — passed
- `pnpm --filter @mastra/behaviors build:lib` — passed
- `pnpm --filter ./mastracode/sdk exec vitest run src/plugins --reporter=dot --bail 1` — 8 files, 66 tests passed

TylerBarnes:
## Real-world proof follow-up

The replacement issue-fixing proof revealed that models may reuse a thread ID as every transition's `attemptId`, causing later transitions to be correctly deduplicated. The transition and exit schemas now explicitly describe `attemptId` as a unique idempotency key that must be fresh for each attempt.

With that schema guidance, a real `openai/gpt-4.1-mini` run autonomously completed investigate → write-test → implement → verify → exit from a single ordinary issue report.

TylerBarnes:
Closing after extracting the behavior-specific runtime, integration, documentation, and proofs into the standalone private behaviors repository. The OSS stack will retain only the generic plugin signal-provider contract in #19294.

PR #19299, opened 2026-07-11
fix: preserve Cloudflare execution context through the custom-route bridge

### LINKED ISSUE #19285: [BUG] Channels on Cloudflare Workers: agent replies dropped after 200 OK —  auto-detection fails ( undefined in channel route)
## Environment
- `@mastra/core` (current), `@mastra/deployer-cloudflare`, `@chat-adapter/slack`
- Channel SDK: `chat` (loaded via `@mastra/core`'s obfuscated `const mod = "chat"; await import(mod)`)
- Hono `4.12.x`
- Deploy target: Cloudflare Workers (module syntax)
- Platform: Slack (`chat.webhooks.slack`)

## Summary
When a Slack DM is sent to an agent deployed to Cloudflare Workers, the webhook returns `200 OK` but the agent never replies, and **no error is logged**. The background agent run is killed when the Worker isolate freezes after the response is sent, because `waitUntil` is never registered with the Workers `ExecutionContext`.

Mastra's `ChannelConfig` docs state Cloudflare/Netlify users "typically don't need" to set `waitUntil` because it auto-resolves from `c.executionCtx` / `c.env.context`. That auto-resolution **fails for channel webhooks on Cloudflare Workers**, so the documented out-of-the-box behavior does not hold.

## Reproduction
1. Deploy an agent with a Slack channel to Cloudflare Workers (via `@mastra/deployer-cloudflare`).
2. Send a DM to the bot.
3. Observe: webhook logs `200`, but no reply appears in Slack, and no error appears in `wrangler tail`.

## Evidence
Instrumentation of the request path shows `waitUntilType: 'undefined'` at every layer:
- Slack adapter `handleWebhook` → `processMessage` receives `options.waitUntil === undefined`
- Mastra's `resolveWaitUntil(c)` (core `chunk` `function resolveWaitUntil(c)`) returns `undefined`, so `chat.webhooks.slack(c.req.raw, undefined)` is called
- `chat`'s `processMessage` then skips `if (options?.waitUntil) options.waitUntil(task)` — the agent run is a fire-and-forget promise that the Worker drops after `200`.

Because the run is dropped (not errored), nothing is ever logged — which is why the failure is silent.

## Root cause analysis
- The generated Worker entry already forwards the execution context correctly: `app.fetch(request, env, context)`.
- Hono's `fetch` forwards the 3rd arg as `executionCtx` (`#dispatch(request, executionCtx, env, method)` → `new Context(request, { executionCtx })`), so `c.executionCtx` *should* be populated.
- Yet inside the channel route handler `c.executionCtx` is `undefined`, so `resolveWaitUntil(c)` finds neither `c.executionCtx?.waitUntil` nor `c.env?.context?.waitUntil`.
- Consequently the agent run is never registered with `ctx.waitUntil` and the isolate freezes after the `200`.

Related: this is the same channels + Cloudflare Workers defect class as #19254 (the `No such module "chat"` bundling error). 

Note: a commonly-suggested fix `resolveWaitUntil: (ctx) => ctx?.waitUntil` does **not** work — the `ctx` Mastra passes is the Hono context, which has no `waitUntil` property (only `c.executionCtx`, which is itself undefined here). `ctx.executionCtx?.waitUntil` also fails for the same reason.

## Workaround (until upstream fix)
Capture the Worker's `ExecutionContext.waitUntil` at the entry (the only place it is reliably available) and expose it to the channel:

```js
// generated entry (index.mjs), patched post-build:
globalThis.__CF_WAIT_UNTIL = context?.waitUntil ? context.waitUntil.bind(context) : undefined;
```

```ts
// agent config:
channels: {
  adapters: { slack: createSlackAdapter() },
  resolveWaitUntil: () => (globalThis as any).__CF_WAIT_UNTIL,
}
```

This keeps the background run alive after the `200` and the reply is delivered.

## Suggested upstream fix
Ensure `c.executionCtx` is available on the Hono context inside channel webhook route handlers on Cloudflare Workers (or have the deployer-generated entry explicitly surface `waitUntil` to Mastra's channel `resolveWaitUntil` default), so Cloudflare users don't need manual `waitUntil` configuration as the docs promise.

Sandy-1711:
## Description

Agents exposed through a channel (Slack, Discord, …) on Cloudflare Workers never reply — the webhook returns `200 OK`, nothing is posted back, and nothing is logged.

**Root cause**

The reply is generated after the webhook responds, so it needs `ExecutionContext.waitUntil` to outlive the `200`. It never gets it. Channel webhooks register as custom API routes, which the Hono adapter re-dispatches through a second, internal Hono sub-app in `handleCustomRouteRequest` — a shared bridge so Express/Koa/Fastify/Next reuse one implementation. That hop forwarded only two of `app.fetch`'s three arguments:

```ts
// packages/server/.../server-adapter/index.ts (before)
this.customRouteHandler = async (request, env) => app.fetch(request, env);
```

`executionCtx` is the optional third argument. The Worker entry and the outer Hono context both carry it; it's dropped at the bridge:

```
workerd → fetch(request, env, context)             executionCtx present
  → outer Hono app  c.executionCtx                  executionCtx present
  → handleCustomRouteRequest(...)                   (no param to carry it)
  → customRouteHandler → sub-app.fetch(req, env)    dropped (2 args)
  → route handler  c.executionCtx → undefined       gone → run killed after 200
```

Downstream, `resolveWaitUntil()` returns `undefined`, the run is fire-and-forget, and the isolate freezes on response and kills it mid-flight — no error surfaces, which is why it's silent.

**Fix**

Forward `executionCtx` through the bridge:

```ts
// packages/server/.../server-adapter/index.ts (after)
this.customRouteHandler = async (request, env, executionCtx) =>
  app.fetch(request, env, executionCtx as Parameters<typeof app.fetch>[2]);
```

- `handleCustomRouteRequest` takes an optional trailing `executionCtx` and passes it to `customRouteHandler`.
- The Hono adapter reads `c.executionCtx` (guarded — the getter throws on Node) and passes it in.
- Express/Koa/Fastify/Next call unchanged and pass `undefined`; those runtimes have no execution context.

Reproduced in `workerd` (`wrangler dev`): before the fix the context is present at the entry and outer context, then absent in the sub-app; after, it reaches the handler.

## Related issue(s)

Fixes #19285

## Type of change

- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [ ] Test update

## Checklist

## ELI5

Cloudflare was sometimes ending the worker too soon, so channel replies never got sent even though the webhook said “success.” This change keeps the work alive until the agent reply is delivered.

## Summary

- Forward Cloudflare’s `executionCtx` through the custom-route bridge into Hono.
- Safely handle environments without an execution context, including Node, Express, Koa, Fastify, and Next.
- Add regression tests verifying `waitUntil` is preserved and Node runtimes remain safe.
- Add patch changesets for `@mastra/hono` and `@mastra/server`.


Sandy-1711:
@coderabbitai review

packages/server/src/server/server-adapter/index.ts
```diff
@@ -921,7 +921,12 @@ export abstract class MastraServer<TApp, TRequest, TResponse> extends MastraServ
...
     app.notFound(() => new Response(null, { status: 404, headers: { [NOT_FOUND_HEADER]: 'true' } }));
 
-    this.customRouteHandler = async (request, env) => app.fetch(request, env);
+    // Forward the platform execution context (e.g. Cloudflare Workers' `waitUntil`)
+    // into the internal app so custom route handlers can keep background work
+    // alive after the response is sent. Without this third arg, serverless
+    // runtimes freeze the invocation on return and kill any in-flight work.
+    this.customRouteHandler = async (request, env, executionCtx) =>
+      app.fetch(request, env, executionCtx as Parameters<typeof app.fetch>[2]);
     return true;
```
NikAiyer, line 934: Could we use Hono’s exported `ExecutionContext` type instead of `unknown` plus a cast?

```ts
executionCtx?: ExecutionContext
```

The Hono adapter can use `ExecutionContext | undefined`, allowing this call without an assertion:

```ts
app.fetch(request, env, executionCtx)
```

Both packages already depend on Hono, and I verified this change builds successfully.

Sandy-1711: Thanks for the review! I will make the change and push a commit shortly

server-adapters/hono/src/__tests__/execution-context.test.ts
```diff
@@ -0,0 +1,88 @@
...
+    const fakeExecutionCtx = {
+      waitUntil: vi.fn(),
+      passThroughOnException: vi.fn(),
+    };
+
+    // Mirrors a Cloudflare Worker entry: `fetch: (request, env, ctx) => app.fetch(request, env, ctx)`.
+    const response = await app.fetch(
+      new Request('http://localhost/probe-execution-ctx', { method: 'POST' }),
+      {},
+      fakeExecutionCtx as any,
```
NikAiyer: Looks good. One small cleanup: could the test add `props: {}` and use `satisfies ExecutionContext` instead of `as any`? This would validate the same contract used by the implementation.

Sandy-1711: Yeah, you are right. Will do this asap and push.

Sandy-1711: done

Sandy-1711:
Tightened the seenWaitUntil as well with:
```ts
ExecutionContext['waitUntil'] | undefined
```

NikAiyer (approved):

NikAiyer:
@Sandy-1711 thanks again for making this change!

PR #19305, opened 2026-07-11
fix(observability): report stable storage capabilities

### LINKED ISSUE #19300: [BUG] @mastra/pg@1.15.x breaks Studio Metrics/Logs detection for PostgresStoreVNext (constructor name regression)
## Summary
With PostgresStoreVNext as the observability store, Studio shows "Metrics/Logs are not available with your current storage" even though the store fully supports metrics/logs. It's a regression in @mastra/pg@1.15.0.

## Versions

`@mastra/pg: 1.15.0 and 1.15.1 (broken); 1.14.3 (works)
mastra (Studio): 1.17.0 and 1.18.2 — both affected
@mastra/core: 1.49.0`

## Root cause
Studio's system-info endpoint reports the observability storage type as:
`observabilityStorageType = storage.stores.observability?.constructor.name`

and the Studio bundle only enables Metrics/Logs when that value is in an allowlist:
`new Set(["ObservabilityStorageClickhouseVNext","ObservabilityStorageDuckDB","ObservabilityInMemory","ObservabilitySpanner","ObservabilityStoragePostgresVNext"])`

But @mastra/pg's built output changed the class emission:

- @mastra/pg@1.14.3: var ObservabilityStoragePostgresVNext = class extends … → constructor.name === "ObservabilityStoragePostgresVNext" ✅
- @mastra/pg@1.15.0/1.15.1: var ObservabilityStoragePostgresVNext = class _ObservabilityStoragePostgresVNext extends … → constructor.name === "_ObservabilityStoragePostgresVNext" ❌ (esbuild adds the leading underscore because the class gained a static self-reference)
So Set.has("_ObservabilityStoragePostgresVNext") is false → metrics/logs reported as unsupported.

## Impact
Any deployed app using PostgresStoreVNext for observability on @mastra/pg@1.15.x loses Studio Metrics + Logs, with no error — just the "not available with your current storage" empty state.

### Steps To Reproduce

Shows the regression directly via the class name:

The Studio allowlist expects "ObservabilityStoragePostgresVNext", so 1.15.x fails the Set.has(...) check.

Full end‑to‑end repro (Studio UI)
1. Create a Mastra app with mastra@1.17.0 (or 1.18.2) and @mastra/core@1.49.0.
2. Install @mastra/pg@1.15.1 and point observability at PostgresStoreVNext:
```
import { Mastra } from "@mastra/core/mastra";
import { PostgresStoreVNext } from "@mastra/pg";
import { Observability, MastraStorageExporter } from "@mastra/observability";

const pg = new PostgresStoreVNext({
  id: "app",
  connectionString: process.env.DATABASE_URL!,
  observability: { connectionString: process.env.DATABASE_URL! },
});

  storage: pg,
    configs: { default: { serviceName: "app", exporters: [new MastraStorageExporter()] } },
  }),
});
```
3. Run the app (mastra dev, or mastra build + start the built server) against a real Postgres.
4. Run an agent once so spans/metrics get written.
5. Open Studio → Metrics (and Logs) tab.
5.1. Expected: metrics dashboard renders.
5.2. Actual: "Metrics are not available with your current storage."
6. Downgrade to @mastra/pg@1.14.3, rebuild/restart, reload Studio → Metrics/Logs now work. No other changes.

NA

### Expected Behavior

NA

### Environment Information

```block
NA
```

### Verification

bkidy:
## What changed

- extend the observability feature contract with explicit `metrics` and `logs` capabilities
- make Postgres vNext advertise both capabilities independently of its runtime constructor name
- expose stable observability capabilities from `/system/packages` and regenerate the client route type
- make Studio prefer the capability response while retaining the class-name fallback for older servers

## Why

Bundling can rename `ObservabilityStoragePostgresVNext` to `_ObservabilityStoragePostgresVNext`, so using `constructor.name` as a capability check incorrectly hides the Metrics page. The explicit capability contract is stable across bundlers and can be adopted by other storage adapters over time.

Fixes #19300

## Testing

- PG vNext focused regression: 1 passed
- Server system handler: 17 passed
- Studio MSW capability hook: 3 passed
- ESLint on all changed TypeScript files: passed
- regenerated `client-sdks/client-js/src/route-types.generated.ts`: 390 routes

Full Playground typecheck is not included as a passing local signal because this shallow checkout does not contain generated declaration outputs for unrelated internal workspace packages; the focused Studio test and changed-file lint pass.

## ELI5

Studio used to guess whether Metrics/Logs should show up by looking at a storage class name—but bundlers can rename that class and make the guess wrong. This PR teaches the server to explicitly advertise “metrics” and “logs” support, so Studio can trust that answer (and still work with older servers).

## What changed

- Extended the observability storage feature contract to include explicit `metrics` and `logs` capabilities (in addition to `delta-polling`).
- Updated the server’s `GET /system/packages` response to optionally include `observabilityStorageCapabilities: { metrics, logs }`, derived from `observabilityStorage.getFeatures()`.
- Updated Postgres vNext to always report `metrics` and `logs` via `getFeatures()` (with `delta-polling` still conditional).
- Added a client-side hook (`useObservabilityStorageCapabilities`) and updated Studio to:
  - Prefer the capability response when present
  - Fall back to legacy class-name detection for older servers
- Updated the generated client route types to include `observabilityStorageCapabilities` and refreshed the route type generation (390 routes).
- Added/updated regression tests and MSW fixtures to cover:
  - Bundler/renamed Postgres vNext storage class behavior
  - Older server compatibility (no capabilities returned)
  - Cases where the server explicitly disables metrics (without incorrectly overriding it)
- Added a changeset to release patch updates for the affected packages (`@mastra/core`, `@mastra/pg`, `@mastra/server`, `@mastra/client-js`, plus `@internal/playground`).


mfrachet (approved):
Frontend wise, looks chill

PR #19307, opened 2026-07-11
fix(core): discover suspended runs persisted by durable/evented agents

### LINKED ISSUE #19302: listSuspendedRuns() / sendToolApproval() storage discovery returns [] for createEventedAgent — queries workflowName "agentic-loop" but evented agents persist as "durable-agentic-loop"
## Summary

`agent.listSuspendedRuns({ threadId, resourceId })` returns an empty `runs` array for a `createEventedAgent`, even when the run is genuinely suspended for a tool approval and its snapshot is on disk with `status: "suspended"`.

Cause: `listSuspendedRuns()` queries `listWorkflowRuns({ workflowName: "agentic-loop" })`, but the evented (durable) agent persists its workflow snapshots under **`durable-agentic-loop`** / **`durable-agentic-execution`**. The hardcoded name matches nothing, so discovery returns `[]`.

This also breaks `sendToolApproval()`'s storage-backed fallback: when no in-memory run is found for the thread, it calls `listSuspendedRuns()` to discover the suspended run — which comes back empty — so the approval can't be routed. (The docs present both as the way browser UIs rediscover pending approvals after a reload / across instances.)

## Environment

- `@mastra/core` **1.50.1**
- Storage: `@mastra/pg` `PostgresStore`
- `createEventedAgent()` wrapping an `Agent` with a `requireApproval` tool

## Relevant source (1.50.1)

```js
// listSuspendedRuns() — dist chunk, ~line 56018
const { runs } = await workflowsStore.listWorkflowRuns({
  workflowName: "agentic-loop",   // <-- hardcoded
  status: "suspended", fromDate, toDate,
});
```

```js
// but the evented/durable agent persists under these names — dist: agent/durable/index.js
const DurableStepIds = {
  AGENTIC_EXECUTION: "durable-agentic-execution",  // line 60
  AGENTIC_LOOP:      "durable-agentic-loop",        // line 62
};
```

So the query workflow name and the persisted workflow name never match for durable/evented agents. There is no option to override the name passed to `listSuspendedRuns`.

## Minimal reproduction (runnable)

```js
// repro-listsuspended.mjs
// Run: ANTHROPIC_API_KEY=... DATABASE_URL=postgres://... node repro-listsuspended.mjs
// Deps: @mastra/core@1.50.1 @mastra/pg @ai-sdk/anthropic zod pg
import { createEventedAgent } from '@mastra/core/agent/durable';
import { anthropic } from '@ai-sdk/anthropic';
import { z } from 'zod';
import pg from 'pg';

const log = (...a) => console.log(...a);
const SCHEMA = 'mastra_repro_ls';
const pool = new pg.Pool({ connectionString: process.env.DATABASE_URL });

const askTool = createTool({
  id: 'ask_user',
  description: 'Ask the user a question. You MUST call this tool before answering.',
  inputSchema: z.object({ question: z.string() }),
  outputSchema: z.object({ answer: z.string() }),
  requireApproval: true,
  execute: async ({ context }) => ({ answer: `answered: ${context?.question ?? '?'}` }),
});
const base = new Agent({
  id: 'repro_agent', name: 'Repro Agent',
  instructions: 'Call ask_user exactly once to ask a clarifying question, then answer briefly.',
  model: anthropic('claude-haiku-4-5-20251001'),
  tools: { ask_user: askTool }, defaultOptions: { maxSteps: 8 },
});
const mastra = new Mastra({
  agents: { repro_agent: createEventedAgent({ agent: base }) },
  storage: new PostgresStore({ id: 'repro', connectionString: process.env.DATABASE_URL, schemaName: SCHEMA }),
});
const agent = mastra.getAgent('repro_agent');

async function status(runId) {
  try {
    const { rows } = await pool.query(
      `SELECT workflow_name, snapshot->>'status' AS status FROM ${SCHEMA}.mastra_workflow_snapshot WHERE run_id = $1`, [runId]);
    return rows;
  } catch (e) { if (e?.code === '42P01') return []; throw e; }
}
async function waitSuspended(runId, ms = 20000) {
  const end = Date.now() + ms;
  while (Date.now() < end) { const r = await status(runId); if (r.some(x => x.status === 'suspended')) return r; await new Promise(r => setTimeout(r, 300)); }
  return await status(runId);
}

const memory = { thread: 'repro-ls-' + Math.floor(Date.now() / 1000), resource: 'user-1' };
const s = await agent.stream([{ role: 'user', content: 'Help me plan a trip.' }], { memory });
const runId = s.runId;
(async () => { try { for await (const _ of s.output.fullStream) {} } catch {} })();
const rows = await waitSuspended(runId);
log('persisted workflow snapshots:', JSON.stringify(rows));

const res = await agent.listSuspendedRuns({ threadId: memory.thread, resourceId: memory.resource });
log('listSuspendedRuns runs:', res?.runs?.length ?? 0);
await pool.end(); process.exit(0);
```

### Output

```
persisted workflow snapshots: [{"workflow_name":"durable-agentic-execution","status":"suspended"},{"workflow_name":"durable-agentic-loop","status":"suspended"}]
listSuspendedRuns runs: 0
```

The run is suspended on disk under `durable-agentic-loop`, but `listSuspendedRuns` returns 0.

## Expected

`listSuspendedRuns()` (and the `sendToolApproval()` storage fallback that uses it) should discover suspended runs for durable/evented agents, i.e. query the workflow name the durable agent actually persists under (`durable-agentic-loop`) rather than a hardcoded `"agentic-loop"`.

## Impact

Any HITL approval UI that follows the docs and uses `listSuspendedRuns()` / `sendToolApproval()` / the `GET /agents/:agentId/suspended-runs` endpoint to rediscover a pending approval after a page reload (or on a second server instance) gets nothing back for evented agents.

## Note

Related to #19301 (cold-resume after restart). Different root cause: #19301 is `resume()` needing the in-memory registry; this one is storage *discovery* querying the wrong workflow name. They compound — `sendToolApproval()` first can't *find* the run (this issue), and even when a run id is supplied, the resume it delegates to can't rehydrate (#19301).

VihaanAgarwal:
## Description

`agent.listSuspendedRuns()` returns an empty list for agents created with `createEventedAgent()`, even when a run is genuinely suspended for tool approval and its snapshot is in storage. This also breaks `sendToolApproval()`'s storage fallback: after a restart (or on another instance) it discovers the suspended run via `listSuspendedRuns()`, gets nothing back, and the approval can't be routed.

Two gaps, both in discovery:

1. It queried `listWorkflowRuns({ workflowName: 'agentic-loop' })`, but the durable/evented loop persists snapshots under `DurableStepIds.AGENTIC_LOOP` (`'durable-agentic-loop'`), so the query matched nothing. Discovery now queries both names; the existing per-snapshot agent-id filter keeps results scoped regardless of which engine wrote them.
2. Even with the right name, durable snapshots were filtered out: the in-process loop stamps `__agentId` and stream state (with thread/resource info) into suspend payloads, while the durable loop's payload is `{ type: 'approval', toolCallId, toolName, args }` with the agent id and memory info serialized in the workflow input instead. Discovery now falls back to `context.input` for the owning agent id and memory info, and reports top-level approval payloads as approval-requiring tool calls.

The default-deny behavior for legacy snapshots is preserved: in-process loop snapshots carry no `agentId` on their workflow input (verified empirically), so a snapshot stripped of `__agentId` still matches no agent.

The regression test runs a real `createEventedAgent` to suspension against `InMemoryStore`, first pins the premise (the snapshot lands under `durable-agentic-loop` with the run suspended), then asserts `listSuspendedRuns()` finds the run with the right thread, resource, and tool call. It fails on main with `total = 0`.

## Related issue(s)

Fixes #19302

## Type of change

## Checklist

## ELI5

When an agent pauses to ask for permission, it can now be found again after a restart or from another server. The fix checks both storage formats and restores the information needed to resume the pending action.

## What changed

- Updated `listSuspendedRuns()` to search both `agentic-loop` and `durable-agentic-loop` snapshots.
- Added fallback parsing for durable workflow inputs to recover:
  - Owning agent ID
  - Thread and resource information
  - Pending tool calls requiring approval
- Preserved default-deny behavior for legacy snapshots.
- Added a regression test covering suspended evented-agent runs with `InMemoryStore`.


VihaanAgarwal:
Closing this: #20529 landed the same fix on main. It queries both `agentic-loop` and `DurableStepIds.AGENTIC_LOOP` in `listSuspendedRuns()` and adds the same `context.input` fallbacks for the agent id and the message-list memory info, so there is nothing left here to merge.

PR #19310, opened 2026-07-12
fix(braintrust): show tool name instead of unknown_tool in Messages tab

jaymalve:
## Description

Tool spans (`TOOL_CALL` and `MCP_TOOL_CALL`) were logged with raw args/result objects, so Braintrust's Messages renderer found no `function.name` and labelled every tool call `unknown_tool`. This chat-shapes tool span input/output into OpenAI messages so Braintrust shows the real tool name.

The tool name comes from the span's `entityName` (falling back to the name inside the span title, e.g. `tool: 'getWeather'`), and the call is paired with its result by the span id.

Before, the tool span input was logged as-is and the renderer couldn't read a name:

```json
{ "city": "Paris" }
```

After, it's an assistant message with a `tool_calls` entry:

```json
[
  {
    "role": "assistant",
    "content": "",
    "tool_calls": [
      {
        "id": "<span-id>",
        "type": "function",
        "function": { "name": "getWeather", "arguments": "{\"city\":\"Paris\"}" }
      }
    ]
  }
]
```

The result is shaped into a matching tool message (`{ role: "tool", content, tool_call_id }`).

This is the companion to #19311 (independent, can merge in any order). They touch different code paths and share the same reproduction.

## Related issue(s)

Fixes #19308

## Type of change

- [x] Bug fix (non-breaking change that fixes an issue)
- [ ] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [ ] Test update

## Checklist

- [x] I have linked the related issue(s) in the description above
- [ ] I have made corresponding changes to the documentation (if applicable)
- [x] I have added tests that prove my fix is effective or that my feature works
- [ ] I have addressed all Coderabbit comments on this PR

PR #19311, opened 2026-07-12
fix(core,braintrust): show tool results in the Braintrust Thread view

### LINKED ISSUE #19309: [BUG] @mastra/braintrust: Thread view never shows tool results (TOOL_CALL span created without toolCallId)
### Describe the Bug

When a Mastra agent calls a tool and traces are exported to Braintrust with `BraintrustExporter`, the **Thread** view shows each tool call (from the model step) but **never the tool's result**. The exporter's thread reconstruction silently drops every result.

**Root cause:** `accumulateToolCallResult` keys pending tool results by `span.input?.toolCallId` (`observability/braintrust/src/tracing.ts`). But `@mastra/core` creates the `TOOL_CALL` span with `input: args` — the validated tool arguments only, with no `toolCallId` (see `getOrCreateSpan({ type: SpanType.TOOL_CALL, ..., input: args })` in `packages/core/src/tools/tool-builder/builder.ts`). `execOptions.toolCallId` is in scope at that call site but never reaches the span, so `pendingToolResults` stays empty and `reconstructThreadOutput` attaches no results.

### Steps To Reproduce

1. Clone the minimal reproduction: https://github.com/jaymalve/mastra-braintrust-unknown-tool-repro
2. `npm install`, then `cp .env.example .env` and add `OPENAI_API_KEY` + `BRAINTRUST_API_KEY`.
3. `npm run repro` (the agent calls one tool; spans flush to Braintrust).
4. In Braintrust, open the newest trace → **Thread** view.
5. The tool call appears, but its result is never shown.

https://github.com/jaymalve/mastra-braintrust-unknown-tool-repro

### Expected Behavior

The **Thread** view shows both the tool call and its result (the `role: "tool"` message), so the full conversation flow is visible.

### Environment Information

```
Reproduced with:
- OS: macOS (Darwin 25.5.0)
- Node: v25.8.1
- Package manager: npm 11.11.0
- @mastra/core: 1.50.1
- @mastra/braintrust: 1.2.3
- @mastra/observability: 1.16.0
- LLM Provider: OpenAI (openai/gpt-4o-mini via the model router)

Also present across @mastra/braintrust 1.1.0 → 1.2.3 and @mastra/core 1.33.0 → 1.50.1.
```

### Screenshot

_(Thread-view screenshot showing the missing result attached in a comment below)_

### Companion issue

One of two related Braintrust tool-span rendering bugs. See #19308 for the `unknown_tool` label in the Messages tab. Both share this same reproduction.

### Verification

jaymalve:
## Description

The Braintrust Thread view showed each tool call but never its result. The exporter keyed pending tool results by `span.input.toolCallId`, but core creates the `TOOL_CALL` span with `input` set to the validated tool arguments only, so the id was never present and every result was dropped during thread reconstruction.

Core now sets `toolCallId` on the `TOOL_CALL` and `MCP_TOOL_CALL` span attributes (the value is already in scope as `execOptions.toolCallId`), and the exporter reads it from there, falling back to `input` for older core versions.

```ts
// core, when creating the tool span
attributes: {
  toolDescription: options.description,
  toolType: logType || 'tool',
  toolCallId: execOptions?.toolCallId, // new
}
```

```ts
// braintrust exporter, when accumulating the result
const toolCallId = span.attributes?.toolCallId ?? span.input?.toolCallId;
```

This is the companion to #19310 (independent, can merge in any order). They touch different code paths and share the same reproduction.

## Related issue(s)

Fixes #19309

## Type of change

- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [ ] Test update

## Checklist

## ELI5

Braintrust now shows both a tool request and the tool’s answer in Thread views. The change adds a shared ID so the two can be matched, including when using older core versions.

## What changed

- Added optional `toolCallId` tracing attributes to `TOOL_CALL` and `MCP_TOOL_CALL` spans.
- Updated the Braintrust exporter to match tool results using the span attribute, with a fallback to `span.input.toolCallId`.
- Added regression and tracing tests covering regular and MCP tools.
- Added patch release changesets for `@mastra/core` and `@mastra/braintrust`.

### DIFF
--- a/.changeset/icy-icons-grin.md
+++ b/.changeset/icy-icons-grin.md
@@ -1,0 +1,5 @@
+---
+'@mastra/braintrust': patch
+---
+
+Fixed tool results missing from the Braintrust Thread view. The exporter now reads the tool call ID from the span attributes to pair each tool result with its call.
--- a/.changeset/nine-heads-boil.md
+++ b/.changeset/nine-heads-boil.md
@@ -1,0 +1,5 @@
+---
+'@mastra/core': patch
+---
+
+Added the tool call ID to tool execution spans (`TOOL_CALL` and `MCP_TOOL_CALL`) so observability exporters can correlate a tool call with its result.
--- a/observability/braintrust/src/tracing.test.ts
+++ b/observability/braintrust/src/tracing.test.ts
@@ -861,6 +861,110 @@ describe('BraintrustExporter', () => {
     );
   });
 
+  /**
+   * Regression test for issue #19309: the Braintrust Thread view never showed
+   * tool results because core creates the TOOL_CALL span with input = args only
+   * (no toolCallId). The toolCallId now lives on the span attributes, and the
+   * exporter reads it from there to pair each result with its call.
+   *
+   * @see https://github.com/mastra-ai/mastra/issues/19309
+   */
+  it('attaches tool results using the toolCallId from span attributes (issue #19309)', async () => {
+    const traceId = 'tool-result-attributes-trace';
+
+    const llmSpan = createMockSpan({
+      id: 'model-gen-span',
+      traceId,
+      name: 'gpt-4-call',
+      type: SpanType.MODEL_GENERATION,
+      isRoot: true,
+      input: [{ role: 'user', content: 'What is 2+2?' }],
+      output: { text: 'The answer is 4.' },
+      attributes: { model: 'gpt-4', provider: 'openai' },
+    });
+
+    const modelStep0Span = createMockSpan({
+      id: 'model-step-0-span',
+      traceId,
+      name: 'Model Step 0',
+      type: SpanType.MODEL_STEP,
+      isRoot: false,
+      input: {},
+      output: {
+        text: '',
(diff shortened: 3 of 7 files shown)


epinzur:
Thanks again, @jaymalve — this one nails #19309 too. Pairing the tool result back to
its call via a stable `toolCallId` is exactly the right idea, and reading it from the
span attributes with an `input` fallback is a sensible, backward-compatible shape.

We're closing this for the same reason as #19310: the `toolCallId` piece is the part
we specifically want to unify across the whole codebase rather than in a single
exporter. Right now the tool-call ID lives in different places depending on which code
path created the span — some set `metadata.toolCallId`, this PR introduces
`attributes.toolCallId`, and a couple of producers (like the Inngest durable-agent
path) don't set it at all even though the value is right there. We'd rather pick one
canonical home for it, wire every producer to set it, and update the exporters that
can expose it (Braintrust, Sentry, OTel) in one coordinated change.

That's a cross-package refactor touching `@mastra/core`, so we're going to take it
internally. Your PR is the blueprint for it. Thank you for the thorough work and the
tests — genuinely helpful, and please keep them coming.

Issue #18341, opened 2026-06-23
[BUG] Durable agents — input/output processor spans still orphan after #18083; MEMORY_OPERATION spans missing


ChristianOlsson:
## Problem

[#18083](https://github.com/mastra-ai/mastra/pull/18083) (in `@mastra/core@1.45.0`) closed [#17863](https://github.com/mastra-ai/mastra/issues/17863) by opening an `AGENT_RUN` root span for durable runs. In-loop spans (`MODEL_GENERATION`, `MODEL_STEP`, `TOOL_CALL`, step-processors) reparent correctly.

**Agent-level input/output processors don't.** They still receive `{} as any` as their observability context, so:

- Their `processor_run` spans emit with no parent — root rows on every durable turn.
- The `MEMORY_OPERATION` spans inside `MessageHistory.processInput` / `processOutputResult` are skipped entirely (the processor bails when `currentSpan` is undefined), so there's no visibility into `storage.listMessages` / `storage.saveMessages` latency.

```
agent run: '<id>'                   ✓ from #18083
input processor: message-history    ✗ orphan
output processor: message-history   ✗ orphan
```

## Root Cause

Two `{} as any` casts. The cast itself signals these were knowingly deferred.

**1. `packages/core/src/agent/durable/preparation.ts:254`** — input processors run before `agentSpan` is created (the span is constructed at line 342):

```ts
await runner.runInputProcessors(messageList, {} as any, requestContext, 0);
```

**2. `packages/core/src/agent/durable/workflows/create-durable-agentic-workflow.ts:285`** — output-processor step doesn't destructure `tracingContext` from `params`:

```ts
async params => {
  const { inputData, mastra, requestContext } = params;   // tracingContext missing
  // ...
  await runner.runOutputProcessors(outputMessageList, {} as any, requestContext ?? new RequestContext(), 0);
},
```

`params.tracingContext` is populated by the workflow engine — another step in the same file already uses it (line 381 destructures; line 452 passes `createObservabilityContext(tracingContext)`).

## Reproduction

```ts
import { Mastra } from "@mastra/core";
import { Agent } from "@mastra/core/agent";
import { createDurableAgent } from "@mastra/core/agent/durable";
import { Memory } from "@mastra/memory";
import { Observability, DefaultExporter } from "@mastra/observability";
import { PostgresStore } from "@mastra/pg";
import { anthropic } from "@ai-sdk/anthropic";

const storage = new PostgresStore({
  id: "mastra",
[...]

PR #19313, opened 2026-07-12
fix(core): pass tracingContext to runProcessOutputStep so durable output-step processor spans nest under AGENT_RUN

### LINKED ISSUE #19312: [BUG] Durable agents: per-step output processor spans orphan (runProcessOutputStep missing tracingContext in llm-execution.ts)
### Describe the Bug

In the durable/evented agent execution path, the framework-created `processor_run` span for an **output** processor's `processOutputStep` is created **without a parent**, so it detaches from the `agent_run` trace and becomes its own root span — one orphaned root per LLM step. A plain (non-durable) `Agent` parents the same span correctly under `MODEL_STEP`.

**Root cause:** In `createDurableLLMExecutionStep` (`packages/core/src/agent/durable/workflows/steps/llm-execution.ts`), the `runProcessOutputStep(...)` call omits `tracingContext`, while its sibling `runProcessInputStep(...)` in the same file passes it. On `main`:

- **Line 363** — `runProcessInputStep({ …, tracingContext: modelSpanTracker?.getTracingContext() ?? tracingContext, … })` ✅
- **Lines 1328–1340** — `runProcessOutputStep({ …, requestContext, writer })` ❌ no `tracingContext`

Without it, `ProcessorRunner.runProcessOutputStep` → `resolveObservabilityContext(args)` falls back to `noOpTracingContext` (`currentSpan: undefined`), so `parentSpan` resolves to `undefined` and the `processor_run` span is created parentless.

> Note: this is a **distinct code path** from #18341 (closed). That fixed the batch `runOutputProcessors` calls in `create-durable-agentic-workflow.ts` and `preparation.ts`. This is the **per-LLM-step** `runProcessOutputStep` in `steps/llm-execution.ts`, which still lacks `tracingContext` on `main`. Affects both `createEventedAgent` and `createDurableAgent` (shared execution step).

### Steps To Reproduce

Minimal reproduction — no external DB (spans captured via an in-memory Observability exporter; storage is in-memory libsql). Runs the **same agent + output processor both ways** in separate processes (two Mastra instances in one process interfere).

```bash
mkdir repro && cd repro
npm init -y && npm pkg set type=module
npm i @mastra/core@1.50.1 @mastra/observability@1.16.0 @mastra/libsql@1.15.1 @mastra/memory@1.22.2 @ai-sdk/anthropic@4.0.12
# save repro.mjs below
ANTHROPIC_API_KEY=sk-... node repro.mjs
```

<details><summary><code>repro.mjs</code></summary>

```js
import { spawn } from 'node:child_process';

const mode = process.argv[2];

if (!mode) {
  const self = fileURLToPath(import.meta.url);
  const runOne = (m) => new Promise((resolve) => {
    let out = '';
    const p = spawn(process.execPath, [self, m], { env: process.env });
    p.stdout.on('data', (d) => { out += d; process.stdout.write(d); });
    p.stderr.on('data', () => {});
    p.on('close', () => {
      const line = out.split('\n').find((l) => l.startsWith('RESULT '));
      resolve(line ? JSON.parse(line.slice(7)) : { mode: m, procCount: 0, orphaned: 0 });
    });
  });
  const plain = await runOne('plain');
  const durable = await runOne('durable');
  console.log('\n================ SUMMARY ================');
  console.log(`plain   (non-durable Agent):     ${plain.orphaned}/${plain.procCount} orphaned`);
  console.log(`durable (createEventedAgent):    ${durable.orphaned}/${durable.procCount} orphaned`);
  const ok = plain.procCount > 0 && plain.orphaned === 0 && durable.procCount > 0 && durable.orphaned === durable.procCount;
  console.log(ok ? '\n✅ BUG REPRODUCED' : '\n❓ inconclusive');
  process.exit(ok ? 0 : 1);
}

const { Mastra } = await import('@mastra/core/mastra');
const { Agent } = await import('@mastra/core/agent');
const { createEventedAgent } = await import('@mastra/core/agent/durable');
const { Observability } = await import('@mastra/observability');
const { TracingEventType } = await import('@mastra/core/observability');
const { LibSQLStore } = await import('@mastra/libsql');
const { Memory } = await import('@mastra/memory');
const { anthropic } = await import('@ai-sdk/anthropic');

const spans = new Map();
const exporter = {
  name: 'collector', async init() {}, async shutdown() {},
  async onTracingEvent(event) {
    const s = event?.exportedSpan;
    if (!s?.id) return;
    const prev = spans.get(s.id) ?? {};
    spans.set(s.id, { name: s.name ?? prev.name, type: s.type ?? prev.type, spanId: s.id,
      parentSpanId: prev.parentSpanId ?? s.parentSpanId ?? null, traceId: s.traceId ?? prev.traceId });
  },
};

const base = new Agent({
  id: 'probe', name: 'probe', instructions: 'Reply "done" in one short sentence.',
  model: anthropic('claude-haiku-4-5-20251001'), memory: new Memory(),
  outputProcessors: [{ id: 'my-output-processor', name: 'my-output-processor', async processOutputStep() { return undefined; } }],
});
const mastra = new Mastra({
  agents: { probe: mode === 'durable' ? createEventedAgent({ agent: base }) : base },
  storage: new LibSQLStore({ id: 'repro', url: ':memory:' }),
  observability: new Observability({ configs: { default: { serviceName: 'repro', exporters: [exporter] } } }),
});

const result = await mastra.getAgent('probe').stream(
  [{ role: 'user', content: 'Say done.' }], { memory: { thread: 't1', resource: 'r1' } });
for await (const _ of (result.output?.fullStream ?? result.fullStream)) { /* drain */ }
await new Promise((r) => setTimeout(r, 3000));

const all = [...spans.values()];
const byId = new Map(all.map((s) => [s.spanId, s]));
const proc = all.filter((s) => s.type === 'processor_run' && s.name === 'output step processor: my-output-processor');
const parentOf = (s) => (!s.parentSpanId ? 'ROOT (orphan!)' : (byId.get(s.parentSpanId)?.type ?? `ext:${s.parentSpanId.slice(0, 6)}`));
console.log(`\n===== mode=${mode} =====`);
console.log(`output-step processor spans: ${proc.length}`);
for (const s of proc) console.log(`  parent: ${parentOf(s)}`);
const orphaned = proc.filter((s) => !s.parentSpanId).length;
console.log(`RESULT ${JSON.stringify({ mode, procCount: proc.length, orphaned })}`);
process.exit(0);
```
</details>

Observed output:
```
===== mode=plain =====     output-step processor span → parent: model_step
===== mode=durable =====   output-step processor span → parent: ROOT (orphan!)
SUMMARY: plain 0/1 orphaned, durable 1/1 orphaned
✅ BUG REPRODUCED
```

https://gist.github.com/devangpadhiyar/b39e57c82abdc93ba0998200e4258bbc

### Expected Behavior

The durable agent's output-step `processor_run` span should nest under `MODEL_STEP` (→ `MODEL_GENERATION` → `AGENT_RUN`), identical to the non-durable path — matching the intent of the existing `durable-agent-tracing.test.ts` ("parents agent-level output processor spans to AGENT_RUN").

**Fix** — add the same argument the sibling call already uses, at `llm-execution.ts:1328`:

```ts
await outputStepRunner.runProcessOutputStep({
  …,
  requestContext,
  tracingContext: modelSpanTracker?.getTracingContext() ?? tracingContext, // ← add
  writer: outputStepWriter,
});
```

### Environment Information

```block
System:  macOS 26.5 | Apple M1 Pro | Node 22.15.1 | npm 10.9.2
npmPackages:
  @mastra/core: 1.50.1        (also present on main / 1.51.0-alpha.5)
  @mastra/observability: 1.16.0
  @mastra/libsql: 1.15.1
  @mastra/memory: 1.22.2
  @ai-sdk/anthropic: 4.0.12
LLM Provider: Anthropic (claude-haiku-4-5)
```

### Verification

devangpadhiyar:
## Description

In the durable/evented agent execution path, the framework-created `processor_run` span for an **output** processor's `processOutputStep` was created **without a parent**, so it detached from the `AGENT_RUN` trace and became its own root span — one orphaned root per LLM step. A plain (non-durable) `Agent` parents the same span correctly under `MODEL_STEP`.

**Root cause:** In `createDurableLLMExecutionStep` (`packages/core/src/agent/durable/workflows/steps/llm-execution.ts`), the `runProcessOutputStep(...)` call omitted `tracingContext`, while its siblings in the same file (`runProcessInputStep`, `runProcessLLMRequest`) all pass `tracingContext: modelSpanTracker?.getTracingContext() ?? tracingContext`. Without it, `ProcessorRunner.runProcessOutputStep` → `resolveObservabilityContext(args)` falls back to `noOpTracingContext` (`currentSpan: undefined`), so `parentSpan` resolves to `undefined` and the `processor_run` span is created parentless.

**Fix:** add the same argument the sibling calls already use (one line). Both `createEventedAgent` and `createDurableAgent` are affected (shared execution step).

Note: this is a **distinct code path** from #18341 (closed), which fixed the batch `runOutputProcessors` calls in `create-durable-agentic-workflow.ts` / `preparation.ts`. This is the **per-LLM-step** `runProcessOutputStep` in `steps/llm-execution.ts`.

## Related issue(s)

Fixes #19312

## Type of change

## Checklist

## Notes

- Added a regression test in `durable-agent-tracing.test.ts` (`parents agent-level output STEP processor spans to AGENT_RUN`) mirroring the existing input-processor test — it uses a `processOutputStep` processor and asserts the resulting `processor_run` span (name `output step processor: <id>`) is created as a child rather than a parentless root.
- The behaviour was validated independently in an isolated repro (see #19312): plain agent → `parent: model_step`; durable agent → `ROOT (orphan!)` without the fix, `model_step` with it.

## ELI5
This PR fixes a tracing bug where “output processing” pieces of an agent run were showing up as their own unrelated trace instead of being attached to the actual agent run. It passes the right tracing info into the output-processing step, and adds a test to prevent the bug from coming back.

## Changes
- In `createDurableLLMExecutionStep`, pass the correct `tracingContext` into `runProcessOutputStep`:
  - uses `modelSpanTracker?.getTracingContext() ?? tracingContext`
- Ensures durable and evented agent output-processor `processor_run` spans are properly nested under the expected agent/model run spans (instead of becoming orphan root spans).
- Updates/extends durable agent tracing regression coverage to assert correct span parenting for the output-step processor.
- Adds a `@mastra/core` patch changeset documenting the tracing fix.


packages/core/src/agent/durable/__tests__/durable-agent-tracing.test.ts
```diff
@@ -237,6 +237,47 @@ describe('DurableAgent observability tracing', () => {
...
+      const outputStepProcessorSpanCall = agentSpans[0].createChildSpan.mock.calls.find(
+        (call: any[]) =>
+          call[0]?.type === 'processor_run' &&
+          call[0]?.name === 'output step processor: test-output-step-processor',
+      );
+      expect(outputStepProcessorSpanCall).toBeDefined();
+
+      cleanup();
+    } finally {
+      spy.mockRestore();
```
intojhanurag, line 289: This test fails,  the mock `createTracker().getTracingContext()` (line 79) returns `{}`. By the time `runProcessOutputStep` runs, `modelSpanTracker` exists, so `modelSpanTracker?.getTracingContext() ?? tracingContext` gives `{}` (truthy, no fallback) → `currentSpan` is undefined → no span created.

intojhanurag (commented):
One test fix needed, see inline comment.

packages/core/src/agent/durable/__tests__/durable-agent-tracing.test.ts
devangpadhiyar: Good catch, and thanks for the fix in `3cc46b8` — confirmed. The mock's `getTracingContext()` returning `{}` was truthy, so `modelSpanTracker?.getTracingContext() ?? tracingContext` never fell back and `currentSpan` stayed `undefined`. Returning `{ currentSpan: span }` mirrors the real `ModelSpanTracker` (`observability/mastra/src/model-tracing.ts`: `currentSpan: this.#currentStepSpan ?? this.#modelSpan`).

One extra thing worth flagging on the assertion: because the output-step processor now runs through a wrapping **processor-workflow**, its `processor_run` span is not a direct child of `agent_run` — it lands several levels down (`workflow_step → workflow_run → …`). The `findParent('agent_run')` walk in your version handles that correctly. I verified the gate holds: reverting the one-line production fix makes `findParent('agent_run')` return `undefined` (parentless orphan) and the test goes red; with the fix it resolves to `agentSpans[0]`.

intojhanurag (approved):

PR #19315, opened 2026-07-12
fix(core): generate thread title for durable agents (createEventedAgent / Inngest)

### LINKED ISSUE #19314: [BUG] Memory `generateTitle` never fires for durable agents (createEventedAgent / Inngest)
## Summary

`Memory({ options: { generateTitle: ... } })` works for a plain `Agent` (both `.generate()` and `.stream()`), but **never fires for a durable agent** created with `createEventedAgent({ agent })` — or, by the same code path, the Inngest durable engine. The run completes, messages persist, but the thread `title` stays `null` forever.

## Root cause

The durable finish step reimplements the non-durable `#executeOnFinish` memory logic but only ports the message-persistence half of it, omitting the entire title-generation branch.

**Non-durable** — `packages/core/src/.../agent` `#executeOnFinish` does:
`resolveTitleGenerationConfig(config?.generateTitle)` → guard `shouldGenerate && !thread.title && messages.length >= minMessages` → `genTitle(userMessage, …)` → `memory.createThread({ title })`.

**Durable** — [`packages/core/src/agent/durable/workflows/create-durable-agentic-workflow.ts`](https://github.com/mastra-ai/mastra/blob/main/packages/core/src/agent/durable/workflows/create-durable-agentic-workflow.ts), the `map-final-output` step, block commented `// Memory persistence (executeOnFinish equivalent)`:

```ts
if (
  registryEntry?.saveQueueManager &&
  registryEntry.memory &&
  durableState?.threadId &&
  durableState?.resourceId &&
  !durableState.observationalMemory &&
  !durableState.memoryConfig?.readOnly
) {
  const memoryMessageList = new MessageList();
  memoryMessageList.deserialize(state.messageListState);

  if (!durableState.threadExists) {
    await registryEntry.memory.createThread?.({ threadId, resourceId, memoryConfig });
  }
  await registryEntry.saveQueueManager.flushMessages(memoryMessageList, threadId, memoryConfig);
}
```

There is no `resolveTitleGenerationConfig` / `genTitle` call anywhere in this file — the title-generation branch that exists in the non-durable `#executeOnFinish` was simply never carried over. `saveQueueManager` and `memory` are both already in scope here (the registry entry exposes them), so the fix is to append the same guarded `genTitle` → `createThread({ title })` logic to this block.

Note the `!durableState.observationalMemory` guard: when Observational Memory is enabled, this whole block is skipped (OM handles its own persistence), so any title fix should be reachable on the OM path too, or title generation should be lifted out of the `!observationalMemory` guard.

## Reproduction

Same `Memory({ options: { generateTitle: { model } } })` config on three agents sharing one `PostgresStore`; trivial one-shot prompt (no tools → clean completion). Probe `mastra_threads.title` immediately and after 8s.

```
[plain-generate]  after 8s: title="Fun fact about otters"           msgs=2
[plain-stream]    after 8s: title="Otter fun facts and sleeping habits" msgs=2
[evented-stream]  after 8s: title=""                                  msgs=2   ← never generated
```

All three runs completed and persisted 2 messages; only the evented/durable one has no title. It's not a suspend, drain, or generate-vs-stream difference — the durable finish step just doesn't call title generation.

## Environment

- `@mastra/core` 1.50.1, `@mastra/memory` 1.22.2, `@mastra/pg` 1.15.1
- Also present on `main` (verified against `create-durable-agentic-workflow.ts` @ `cf7831a`)
- Node 22, PostgresStore

## Related

- #18076 (feature: hook when a title is generated) presumes titles are being generated — moot on the durable path until this is fixed.
- #18617 (title persistence unhandled rejection) — same `#executeOnFinish` title branch, non-durable.

## Workaround

Generate the title explicitly after the first turn (own `generateText` call keyed off the first user message) and write it to the thread + your own linked records, rather than relying on `Memory.generateTitle`.

devangpadhiyar:
## Description

`Memory` `generateTitle` works for a plain `Agent` (both `.generate()` and `.stream()`) but **never fires for a durable agent** (`createEventedAgent`, and by the same code path the Inngest engine). The run completes and messages persist, but the thread `title` stays `null`.

**Root cause:** the durable finish step in `create-durable-agentic-workflow.ts` (the `map-final-output` step, block commented `// Memory persistence (executeOnFinish equivalent)`) reimplemented only the message-persistence half of the non-durable `#executeOnFinish`. The title-generation branch (`resolveTitleGenerationConfig` → `genTitle` → `createThread({ title })`) was never carried over, so `generateTitle` silently no-ops on the durable path.

**Fix:** a `generateThreadTitle` closure is parked on the `RunRegistryEntry` during `prepareForDurableExecution` — where the agent instance is in scope — and invoked from the durable finish step. It:

- resolves the merged `generateTitle` config,
- guards on first turn + no existing title + `minMessages` (mirroring `#executeOnFinish`),
- generates the title via the agent's `genTitle`, and
- persists it onto the thread.

It is kept **outside** the `!observationalMemory` guard so title generation also runs when Observational Memory is enabled (OM handles its own message persistence, but title generation is orthogonal). Like the other registry closures (`drainPendingSignals`, `goal`, `onIterationComplete`, …), it is non-serializable, so it fires for in-process durable runs and degrades cleanly on cross-process engines (e.g. Inngest after a worker restart).

I first tried resolving the agent inline via `mastra.getAgentById(agentId)` in the finish step, but that no-ops whenever the run isn't attached to a resolvable `Mastra` instance (the common durable harness case) — which is precisely the bug. Parking the closure at preparation time, where the real agent is already in hand, is the reliable approach and matches the existing registry-closure pattern.

## Related issue(s)

Fixes #19314

## Type of change

## Checklist

## Tests

Added two cases to `durable-agent-memory.test.ts` (`describe('generateTitle')`), mirroring the non-durable `title-generation.test.ts` harness (MockMemory + a dedicated title model wired via `getMergedThreadConfig`):

- generates a thread title from the first message after a completed durable stream;
- does not overwrite an existing thread title.

Verified locally against the pinned commit:

- `durable-agent-memory.test.ts` — **22 passed** (incl. the 2 new)
- all `src/agent/durable/` tests — **547 passed**, 0 failed
- non-durable `title-generation.test.ts` — **43 passed** (no regression on the shared title path)

## ELI5

Durable agents can now automatically name a chat when they finish—just like non-durable agents. If a title is already set, it won’t be overwritten, and this works even when Observational Memory is enabled.

## What changed

- During durable execution preparation, the run registry now stores a `generateThreadTitle` closure (in-process only) that knows how to generate and persist a thread title later.
- When the durable workflow finishes, it calls `registryEntry.generateThreadTitle` to generate/persist the title (best-effort; failures only log a warning).
- Title generation is guarded so it only runs when:
  - `generateTitle` is enabled in the merged memory/thread config,
  - the thread doesn’t already have a title,
  - `minMessages` requirements are satisfied,
  - and the title is generated from the first relevant user turn/message.
- The generated title is persisted independently of Observational Memory message persistence:
  - uses `memory.updateThread` when a thread record already exists (preserving existing thread metadata like `createdAt`),
  - otherwise uses `memory.createThread` for first-time persistence.
- If memory is missing (or durable is cross-process), the registry closure is not set, so the durable engines skip this step.
- Read-only memory skips title generation.
- Added a changeset documenting the behavioral fix for durable agents created via `createEventedAgent` (including the Inngest path).

## Tests

- 22 durable-agent memory tests passed.
- 547 durable-agent tests passed.
- 43 non-durable title-generation tests passed.


intojhanurag (approved):
this only covers in-process runs, so Inngest after a worker restart still skips title  generation (existing registry-closure tradeoff, fine as-is).

PR #19316, opened 2026-07-12
fix(core): normalize V4 file parts

### LINKED ISSUE #19298: Agent loop sends flat file parts to LanguageModelV4 providers — serialized as null on the wire → OpenAI/Azure 400
### What happened

Any `agent.stream()` / `agent.generate()` call whose prompt contains a **file part** (e.g. an image sent as a data URI) fails with an OpenAI/Azure 400 when the model is a **LanguageModelV4** provider (AI SDK v7, `@ai-sdk/openai` 4.x):

> 400 `invalid_type`: "Invalid type for 'input[N].content[1]': expected an object, but got null instead."

The file part is serialized as literal `null` on the wire. In production this bricks every thread that has an image anywhere in its history — each run replays the file part from memory and dies before the first token.

### Root cause

The V4 provider spec requires `LanguageModelV4FilePart.data` to be a **tagged union** (`SharedV4FileData`: `{type: 'data' | 'url' | 'reference' | 'text', ...}`). Mastra builds the LLM prompt with **flat** file parts — `{type: 'file', data: 'data:image/webp;base64,…', mediaType}` (the V1/V2 shape, where `data` is a plain string/URL/Uint8Array):

- `packages/core/src/agent/message-list/conversion/to-prompt.ts` (`aiV4CoreMessageToV1PromptMessage`, `file` case) keeps `data` flat — correct for the V1 prompt shape it targets, but that shape is what ultimately reaches V4 models.
- `packages/core/src/llm/model/aisdk/v7/model.ts` (`AISDKV7LanguageModel.doStream/doGenerate`) forwards call options to the wrapped V4 model with **only `remapToolsToV4`** — tools get version-remapped, file parts don't. The comment on `remapToolsToV4` already names the pattern ("Tools may arrive in V2 format when prepared upstream … without knowing the final model version"); file parts have the exact same problem and no remap.

`@ai-sdk/openai`'s content conversion then switches on `part.data.type`, which is `undefined` for a flat string, falls through every case, returns `undefined`, and `JSON.stringify` turns it into `null` on the wire.

**Why this doesn't reproduce with plain `streamText`:** the `ai` package normalizes flat file data into the tagged union in `convertToLanguageModelPrompt` before calling `doStream`. Mastra's agent loop calls `doStream` directly and skips that normalization — so the same messages work via `streamText` and die via the Agent.

### Minimal reproduction

Verified against `@mastra/core` 1.50.1 + `@ai-sdk/openai` 4.0.11 (`ai` 7.0.22):

```js

const captured = [];
const openai = createOpenAI({
  apiKey: 'test-key',
  fetch: async (_url, init) => {
    captured.push(JSON.parse(init.body));
    return new Response(JSON.stringify({ error: { message: 'captured' } }), { status: 400 });
  },
});

const agent = new Agent({
  name: 'repro',
  instructions: 'You are a test agent.',
  model: openai.responses('gpt-5.2'),
});

const stream = await agent.stream([
  {
    role: 'user',
    content: [
      { type: 'text', text: 'What is in this image?' },
      {
        type: 'file',
        data: 'data:image/webp;base64,UklGRhoAAABXRUJQVlA4TA0AAAAvAAAAEAcQERGIiP4HAA==',
        mimeType: 'image/webp',
      },
    ],
  },
]);
for await (const _ of stream.fullStream) { /* drain */ }

const userMsg = captured[0].input.find(m => m.role === 'user');
console.log(JSON.stringify(userMsg.content));
// → [{"type":"input_text","text":"What is in this image?"},null]
//                                                          ^^^^ the file part
```

Expected: `{"type":"input_image","image_url":"data:image/webp;base64,…"}` (or equivalent) as the second content part.

### Impact

- Confirmed on `@ai-sdk/openai` / `@ai-sdk/azure` 4.x (Responses API). Every V4-spec provider that reads `part.data.type` is presumably affected the same way.
- Fresh messages with file parts fail immediately; threads with an image in retained history fail on **every subsequent run** (the part replays from memory), with zero tokens generated — from the user's perspective the thread is permanently dead.

### Suggested fix

Normalize file-part `data` to the `SharedV4FileData` tagged union where tools are already remapped — `AISDKV7LanguageModel` (`remapToolsToV4` has a natural sibling `remapFilePartsToV4`): `string` data URI / http(s) URL → `{type: 'url', url}` or `{type: 'data', data: base64}`, `Uint8Array` → `{type: 'data', data}`, `URL` → `{type: 'url', url}`, already-tagged objects passed through.

Same bug class as #19020 / #17366 (file-part shape drift across AI SDK spec versions at an uncovered conversion site), different site: those are read/persist paths, this is the prompt-emission path into `doStream`.

We're currently working around it downstream with a `wrapLanguageModel` middleware that rewrites flat file parts into the tagged union in `transformParams` — happy to share it or send a PR if you'd take one.

bkidy:
Fixes #19298.

Agent loops now normalize flat legacy file data before calling V4 providers, so data URI attachments no longer serialize as `null`.

Before: `{ type: "file", data: "..." }`
After: `{ type: "file", data: { type: "data", data: "..." } }`

Regression coverage includes Agent MessageList prompts and direct V4 model calls.

## ELI5

Some AI agents were sending “file attachments” in an older format, and newer model providers ended up receiving `null` instead of the file (leading to errors). This PR converts those attachments into the exact newer V4 “tagged” format before the provider is called, so images/files reliably work.

## What changed

- Fixed `stream()` and `generate()` paths so legacy prompt message `file` parts are normalized into the LanguageModelV4 tagged format before calling LanguageModelV4 providers (preventing wire-time `null` attachments and OpenAI/Azure 400s).
- Added detection for already-V4-tagged file data and preserves it unchanged.
- Converted legacy “flat” file shapes into the V4 tagged union structure (including `{ type: 'data', data: ..., mediaType? }`) via the existing `convertToDataContent` flow.
- Applied the remapping to both fresh prompts and retained message history by scanning `user`/`assistant` messages for `file` parts.
- Restored the prompt type after a TypeScript widening change in `remapFilePartsToV4`, mirroring the existing cast approach used by the tools remapper.
- Updated `AISDKV7LanguageModel` so `doGenerate` and `doStream` use the combined remapper (tools + file-part normalization).
- Added regression tests verifying remapping for base64/data URIs, remote URLs, `Uint8Array`, tagged/reference file objects, and different user vs assistant roles.
- Added a test for malformed data URIs: the remapping now fails client-side with `MastraError` code `INVALID_DATA_URL_FORMAT` and prevents the provider call.
- Added a patch changeset for `@mastra/core` documenting the fix.


abhiaiyer91:
Pushed be17a38 (via Mastra Code agent): `pnpm build:lib` in `packages/core` was failing with TS2322 at `model.ts:100` — the `.map()` in `remapFilePartsToV4` widens the message union across roles, so the rebuilt prompt wasn't assignable to `LanguageModelV4Prompt`. Restored the prompt type with a cast, mirroring the existing `as typeof options.tools` pattern in `remapToolsToV4`. Build, lint, and all 6 tests pass locally. Thanks for the fix @bkidy!

abhiaiyer91:
Also pushed 10a26b2 (via Mastra Code agent): added a test covering the malformed data URI path — `convertToDataContent` throws a client-side `MastraError` (`INVALID_DATA_URL_FORMAT`) instead of letting the bad part reach the provider as a 400. This documents the intentional error-surface change this fix introduces. 7/7 tests passing.

PR #19322, opened 2026-07-12
fix(core): prevent fatal crash when LSP server dies mid-write

### LINKED ISSUE #19323: Fatal crash (write EPIPE) when an LSP server dies while a request is being written
## Describe the bug

When an LSP server process dies (or closes its stdin) while the workspace LSP client is writing a request to it, the entire host process crashes with a fatal unhandled rejection:

```
Fatal error: write EPIPE
    at LocalProcessHandle.sendStdin (...)
```

In Mastra Code this means any language server crashing mid-session (tsserver OOM, malformed input, etc.) takes the whole CLI down, losing the session and any in-flight work.

## Root cause

`vscode-jsonrpc`'s `Connection.sendRequest` re-throws stream write errors inside an `async` promise executor that nothing awaits. The throw rejects the executor's own implicit promise instead of the request promise, so the write failure surfaces as an unhandled rejection and Node kills the process.

All response/notification write paths in `connection.js` have `.catch()` on the write — `sendRequest` is the one path that leaks.

## Steps to reproduce

1. Start a workspace with an LSP server configured
2. Have the server process exit or close its stdin while a request (hover, definition, etc.) is being written
3. Host process dies with `write EPIPE`

Deterministic minimal repro: a fake LSP server that completes the initialize handshake, then calls `fs.closeSync(0)` while staying alive — the next request write hits EPIPE and crashes the client process.

## Expected behavior

The LSP request should fail cleanly (the client already wraps every request in a timeout) and the host process should survive. A dying language server should never be fatal to the application embedding `@mastra/core`.

## Environment

- mastracode v0.30.0 / @mastra/core (latest)
- Node v25.2.1, Linux

CodingButter:
When an LSP server process dies (or stops reading its stdin) while a request is being written to it, the whole host process crashes with a fatal `write EPIPE`. We hit this repeatedly in Mastra Code — any language server crashing mid-session takes the entire CLI down with it.

The root cause is in how vscode-jsonrpc's `Connection.sendRequest` handles write failures: it re-throws the stream write error inside an `async` promise executor that nothing awaits, so the failure surfaces as an unhandled rejection instead of rejecting the request promise.

This guards the message writer in `LSPClient.initialize` so write rejections are swallowed — the error still reaches the connection's error handler, and every request is already wrapped in a timeout, so callers now get a clean timeout error instead of a crash.

```ts
// before: server dies mid-write → unhandled rejection → fatal "write EPIPE"
// after:  queryHover(...) rejects with "LSP request timed out" and the process survives
```

Includes a regression test with a fake LSP server that completes the initialize handshake and then closes its stdin while staying alive, reproducing the exact EPIPE write path (fails without the fix, passes with it).

Fixes #19323

## ELI5

Sometimes an app sends a message to a language server, but the server dies or stops accepting input mid-message. This PR makes the app handle that safely, so it doesn’t crash—instead the request fails cleanly (via the existing timeout).

## Summary

- Updated `LSPClient.initialize` to swallow failed JSON-RPC stream writes (preventing host-crashing unhandled write rejections) while still routing errors through the connection’s existing error handling.
- Kept the existing per-request timeout behavior so affected requests resolve as a clean timeout error rather than crashing the host.
- Added a regression test with a fake LSP server that completes `initialize`, then closes stdin while staying alive; asserts hover requests reject without emitting `unhandledRejection` events.
- Added a patch changeset for `@mastra/core` describing the user-visible “crash to timeout error” outcome.

### DIFF
--- a/.changeset/lsp-write-epipe-crash.md
+++ b/.changeset/lsp-write-epipe-crash.md
@@ -1,0 +1,5 @@
+---
+'@mastra/core': patch
+---
+
+Fixed a fatal crash (write EPIPE) when a language server process exits or stops reading while a request is being sent to it. LSP requests now fail with a clean timeout error instead of crashing the host process.
--- a/.changeset/lsp-write-epipe-crash.md
+++ b/.changeset/lsp-write-epipe-crash.md
@@ -2,4 +2,4 @@
 '@mastra/core': patch
 ---
 
-Fixed a fatal crash (write EPIPE) when a language server process exits or stops reading while a request is being sent to it. LSP requests now fail with a clean timeout error instead of crashing the host process.
+Fixed a crash that could happen when a language server process exited or stopped responding while a request was being sent to it. The request now fails with a clean timeout error instead of crashing the host process.
--- a/packages/core/src/workspace/lsp/client.test.ts
+++ b/packages/core/src/workspace/lsp/client.test.ts
@@ -1,5 +1,13 @@
-import { describe, it, expect } from 'vitest';
-import { diagnosticsKey } from './client';
+import { existsSync } from 'node:fs';
+import { mkdtemp, rm, writeFile } from 'node:fs/promises';
+import { tmpdir } from 'node:os';
+import path from 'node:path';
+
+import { describe, it, expect, beforeEach, afterEach, vi } from 'vitest';
+
+import { LocalSandbox } from '../sandbox/local-sandbox';
+import { LSPClient, diagnosticsKey } from './client';
+import type { LSPServerDef } from './types';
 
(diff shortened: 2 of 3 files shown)


CodingButter:
@coderabbitai Addressed in 2bb7351:

- **Changeset wording**: reworded to describe the user-visible outcome; removed the internal "write EPIPE" reference.
- **Debug log for swallowed write errors**: skipping — `LSPClient` has no logger field, so `this.logger?.debug?.(...)` would be a no-op. The failure still reaches the connection's error handler via the writer's `handleError`, and adding a logger to the class is out of scope for this fix.

abhiaiyer91 (approved):

PR #19351, opened 2026-07-13
fix(core): auto-start background task workers in library mode

### LINKED ISSUE #19339: Background agents without mastra start
This issue was created from Discord post 1526153336309547070:

[Open in Discord](https://discord.com/channels/1309558646228779139/1526153336309547070)

Hi,

I'm trying to do background subagents (first time using background tasks).
The delegation goes well, but then the subagents never seem to run.

According to Claude, this is because I'm running mastra as a lib, not as a server, so the (undocumented) `.startWorkers()` never runs.

```ts
export function createChatRouter(): Router {
  const router = Router();

  router.post("/", async (req, res) => {
    // auth and context
    ...

    await runWithEventContext(eventContext, async () => {
      const mastra = getMastra();


      const stream = await handleChatStream({
        mastra,
        agentId: "supervisor",
        version: "v6",
        params: {
          ...
          memory: {
            thread: xxx,
            resource: xxx,
          },
        },
      });

      pipeUIMessageStreamToResponse({
        response: res,
        stream: throttled(stream),
      });
    });

  });
}
```

and

```ts
export function getMastra(): Mastra {
  if (!mastra) {
    observabilityInstance = buildObservability();
    mastra = new Mastra({
      logger: new PinoMastraLogger(),
      storage: getMastraStorage(),
      agents: {
        ...
      },
      ...(observabilityInstance && { observability: observabilityInstance }),
    });
  }
  return mastra;
}
```

Is there still a way to run full mastra as a lib with the new orchestration?

intojhanurag:
## Description

Fixes background tasks never completing when Mastra is used as a library (`new Mastra(...)` without `mastra start`). Without `mastra start`, nothing calls `startWorkers()`, so dispatched background tasks — including backgrounded sub-agent delegations — were picked up but hung in `running` forever.

- Lazily call `startWorkers()` in the agent execution path when a `backgroundTaskManager` exists and workers haven't been started yet (`agent.ts`)
- Expose an internal `__workersStarted` getter on `Mastra` to guard the idempotent lazy start (`mastra/index.ts`)
- Add a dedicated regression test that verifies a backgrounded sub-agent delegation completes without any manual `startWorkers()` call
- Update existing test comment to reflect the new lazy-start behavior

## Related Issue(s)

Fixes #19339

## Type of Change

- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [ ] Test update

## Checklist

## ELI5
If you use Mastra as a library (without running `mastra start`), background jobs could get stuck because no worker processes were ever started. This PR makes Mastra automatically start the background-task worker machinery the first time a background task needs to run, so tasks (including background sub-agent delegations) can finish on their own.

## Changes
- Added lazy, single-flight background-worker startup in `BackgroundTaskManager` (`#ensureWorkersStarted()`), guarded by `#workersStartPromise`, so concurrent first dispatches share one startup.
- Wired lazy startup into background-task entry points (`dispatch()` and `resume()`), so task enqueue/recovery/restart/resume paths start background execution workers when needed.
- Added `Mastra.__startBackgroundTaskWorkers()` and `Mastra.__workersStarted` so only the background-task execution/orchestration pieces are started in library mode (without booting scheduler/agent-schedule workers or subscribing user event listeners).
- Ensured `stopWorkers()` resets the lazy-start state (`BackgroundTaskManager.__onWorkersStopped()`), allowing later dispatches to restart workers.
- Expanded regression coverage in `subagent-background-library-mode.test.ts` to verify:
  - backgrounded sub-agent delegation completes without manually calling `startWorkers()`,
  - directly dispatched background tasks complete in library mode,
  - stale tasks recovered on restart complete,
  - concurrent cold dispatch starts `__startBackgroundTaskWorkers()` only once,
  - workers restart correctly after `stopWorkers()`.
- Updated the background sub-agent delegation test comment in `subagent-background-tool-content.test.ts`.
- Added a patch changeset documenting the `@mastra/core` library-mode background-task fix (fixing issue `#19339` behavior).


packages/core/src/agent/agent.ts
```diff
@@ -6837,6 +6837,16 @@ export class Agent<
...
       // without events). Avoids requestContext serialisation loss.
       executionWorkflow.__registerMastra(effectiveMastra);
+
+      // Library-mode users who instantiate Mastra directly (without `mastra start`)
+      // never call startWorkers(). When background tasks are configured, dispatched
+      // tasks get picked up by the manager but hang in `running` forever because
+      // the workers that drive them to completion were never started. Lazily start
+      // them here, mirroring the evented path above. startWorkers() is idempotent —
+      // safe to call if already started.
+      if (effectiveMastra?.backgroundTaskManager && !effectiveMastra.__workersStarted) {
```
abhiaiyer91: Nice find, and the test is a faithful repro of #19339. One suggestion on placement: this guard covers the agent-delegation producer, but not the other library-mode producer — direct `mastra.backgroundTaskManager` + `createBackgroundTask()`/`enqueue()`. I verified locally that a direct-enqueue sibling of your new test still hangs at `running` with this fix in place.

`manager.enqueue()` is the natural choke point — it already awaits `initPromise` for the same "caller arrived before infra is ready" reason (`manager.ts:219-224`). Moving the guard there covers every producer with one hook, and your new test still passes unchanged:

```ts
// manager.ts, enqueue(), after `if (this.initPromise) await this.initPromise;`
if (this.#mastra && !this.#mastra.__workersStarted) {
  await this.#mastra.startWorkers();
}
```

If you take this, the `agent.ts` hunk can be dropped entirely — the evented branch already calls `startWorkers()` itself.

abhiaiyer91 (commented):
Review from Mastra Code (agent) on behalf of abhiaiyer91 — verified locally against this branch.

**Non-blocking, mostly pre-existing:** lazy `startWorkers()` isn't serialized, so two concurrent first requests can both enter it (the `#workersStarted` flag is only set at the end). Consequences are minor but real — user event listeners can double-subscribe (the `alreadySubscribed` check races), and the OrchestrationWorker can leak a duplicate transport (`#running` is set only after the awaited `transport.start()`). The evented branch above (`agent.ts` ~L6827) has the same race today. There's precedent for the fix in `#schedulingWorkersStartPromise` — worth a follow-up to wrap `startWorkers()` in the same single-flight pattern.

packages/core/src/agent/__tests__/subagent-background-library-mode.test.ts
```diff
@@ -0,0 +1,131 @@
...
+            { type: 'text-delta', id: 'x', delta: 'Hello from the sub-agent.' },
+            { type: 'text-end', id: 'x' },
+            { type: 'finish', finishReason: 'stop', usage: { inputTokens: 1, outputTokens: 1, totalTokens: 2 } },
+          ]),
+        }),
+      }),
+    });
+  }
+
+  it('runs a backgrounded sub-agent delegation to completion', async () => {
```
abhiaiyer91, line 96: Suggest adding a second case for the direct-enqueue path — it fails with the current placement and passes with the choke-point one, so it pins down the full library-mode contract:

```ts
it('runs a directly-enqueued task to completion', async () => {
  mastra = new Mastra({ logger: false, storage: new MockStore(), backgroundTasks: { enabled: true } });
  const bgTask = createBackgroundTask(mastra.backgroundTaskManager!, {
    toolName: 'my-tool', toolCallId: 'call-1', args: {}, agentId: 'a1', runId: 'r1',
    context: { executor: { execute: async () => ({ data: 'hello' }) } },
  });
  const { task } = await bgTask.dispatch();
  // ...same poll-to-terminal-state loop as above, expect 'completed'
});
```

packages/core/src/background-tasks/manager.ts
```diff
@@ -731,7 +733,24 @@ export class BackgroundTaskManager {
...
+   * manager's own initPromise when sharing Mastra's manager.
+   */
+  async #ensureWorkersStarted(): Promise<void> {
+    if (this.#mastra && !this.#mastra.__workersStarted) {
+      await this.#mastra.startWorkers();
+    }
+  }
+
   private async dispatch(task: BackgroundTask, isRestart?: boolean): Promise<void> {
+    await this.#ensureWorkersStarted();
```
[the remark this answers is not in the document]
intojhanurag: Deliberately leaving this as-is — a startup failure here doesn't strand tasks: the task is persisted as `pending` before dispatch, so it self-heals via the existing recovery paths (`recoverStaleTasks()` on the next init re-dispatches pending tasks, and `drainPending()` picks them up once a run terminates). `recoverStaleTasks()` already wraps its dispatches in try/catch and logs. And `dispatch()` could always throw (e.g. `pubsub.publish` failure), so awaiting `startWorkers()` doesn't introduce a new failure class — callers see the same semantics as before. Wrapping it in a catch would silently swallow a real startup error and leave the task hanging with no signal.

intojhanurag:
Quick summary of where this PR landed after the review rounds, since the diff grew a bit from the original one-liner idea.

**The problem** is really a lifecycle question: in library mode nothing calls `startWorkers()`, so dispatched background tasks get picked up but never run to completion.

**Final design:**

- `BackgroundTaskManager` lazily starts workers from `dispatch()` and `resume()` — the choke points every producer goes through (enqueue, direct `createBackgroundTask()`, stale-task recovery, restart). Single-flight, so concurrent first dispatches on a cold instance share one start.
- The lazy path calls a new internal `Mastra.__startBackgroundTaskWorkers()` instead of the full `startWorkers()`. It starts only the execution machinery (orchestration + backgroundTasks workers, plus the push-mode workflow subscription). Dispatching a task no longer boots the scheduler / agent-schedule workers or subscribes user event listeners as a side effect. Mirrors the existing `#startSchedulingWorkers()` pattern that `schedules.create()` uses.
- `stopWorkers()` resets the manager's lazy-start guard, so a dispatch after stopping workers restarts them instead of hanging.

Each of these came out of an empirically reproduced failure, and each has its own regression test (6 total in `subagent-background-library-mode.test.ts`). Full core suite passes locally (12k+ tests, no type errors).

Happy to adjust if you'd rather see a different shape here.

PR #19362, opened 2026-07-13
fix(react): emit DB-shaped file parts so toAISdkMessages does not crash on attachments

### LINKED ISSUE #19356: [BUG] @mastra/react emits AIV5-shaped file parts (url/mediaType) that crash toAISdkMessages — attachment messages break the chat until refresh
## Describe the Bug

`useChat` (with `enableThreadSignals: true`) builds file parts in the **AIV5 UI shape** — `{ type: 'file', url, mediaType }` — inside `format: 2` DB messages, but `toAISdkMessages` from `@mastra/ai-sdk/ui` reads DB-shaped file parts — `{ type: 'file', data, mimeType }`. The adapter finds `part.data === undefined` and crashes:

```
Uncaught TypeError: Cannot read properties of undefined (reading 'startsWith')
    at createDataUri (@mastra/core, src/agent/message-list/prompt/image-utils.ts)
    at AIV5Adapter.toUIMessage
    at AIV6Adapter.toUIMessage
    at MessageList.toAIV6UIMessages
```

Two producers in `packages/react` emit the mismatched shape:

1. **`signalContentsToUserMessages` → `toMessagePart`** (`src/lib/…`): when a `data-user-message` signal echoes a sent message that carried an attachment, the file content `{ type: 'file', data, mediaType }` is mapped to `{ type: 'file', mediaType, url }` — the `data` key is dropped.
2. **The `case 'file'` stream-chunk handler**: assistant `file` chunks are appended as `{ type: 'file', mediaType, url, providerMetadata }` — same missing `data`.

The crash path in `@mastra/core`'s `AIV5Adapter.toUIMessage` file branch: `typeof part.data === 'string'` is false → `categorized = { type: 'raw', … }` → `filePartData = part.data` (`undefined`) → `createDataUri(undefined)` → `.startsWith('data:')` throws.

Because the **stored** message is written in the correct DB shape, the crash is transient: every render throws while the live echo is in `chat.messages`, then a refresh (which reloads stored history) renders fine. In our app this took down the whole chat surface for the duration of the first run after sending a screenshot.

Note the correct DB shape *is* produced elsewhere in `packages/react` — `coreUserMessageToParts` in `fromCoreUserMessage.ts` emits `{ type: 'file', mimeType, data }` — so the fix is likely aligning the two producers above with it.

Still present in `@mastra/react@1.2.5-alpha.6` (checked the published dist today).

## Steps To Reproduce

In-app:

1. React app with `useChat({ agentId, threadId, resourceId, enableThreadSignals: true })`, rendering via `toAISdkMessages(chat.messages, { version: 'v6' })`.
2. Send a message with an attachment through the agent's `queueMessage` (contents `[{ type: 'text', text }, { type: 'file', data: '<url or data-uri>', mediaType: 'image/png' }]`).
3. When the `data-user-message` signal echoes into `chat.messages`, every render throws the `TypeError` above.
4. Refresh the page — stored messages load in the DB shape and render fine, confirming only the live echo is malformed.

Library-only repro (no app needed) — feed `toAISdkMessages` the exact message shape `useChat` produces after the echo:

```ts
import { toAISdkMessages } from '@mastra/ai-sdk/ui'

// what useChat's signalContentsToUserMessages/toMessagePart puts in chat.messages
const messages = [
  {
    id: 'signal-1',
    role: 'user',
    createdAt: new Date(),
    content: {
      format: 2,
      parts: [
        { type: 'text', text: 'have a look at this' },
        { type: 'file', mediaType: 'image/png', url: 'https://example.com/shot.png' }, // AIV5 UI shape — no `data`
      ],
    },
  },
]

toAISdkMessages(messages, { version: 'v6' })
// TypeError: Cannot read properties of undefined (reading 'startsWith')
```

https://github.com/harjot1singh/mastra-react-file-part-crash-repro — `bun install && bun repro.ts` throws the TypeError; `bun test` holds the expected-behavior assertion (red today).

## Expected Behavior

A sent attachment renders as a file part in the converted UI messages — `{ type: 'file', url, mediaType }` after conversion — with no crash. Internally, `useChat` should emit DB-shaped file parts (`{ data, mimeType }`, as `coreUserMessageToParts` already does) so the `@mastra/ai-sdk` converters can read them; alternatively the AIV5/AIV6 adapters could tolerate the `url` shape defensively.

## Environment Information

```
System:
  OS: macOS 15.5
  CPU: (10) arm64 Apple M4
  Shell: 5.9 - /bin/zsh
Binaries:
  Node: 25.2.1
  npm: 11.6.2
  bun: 1.3.5
npmPackages:
  @mastra/react: ^1.2.4 => 1.2.4
  @mastra/client-js: ^1.31.1 => 1.31.1
  @mastra/ai-sdk: 1.6.1
  @mastra/core: 1.50.1 (server + hoisted client dep)

LLM Provider: Anthropic (not relevant — pure client-side conversion bug)
Browser: Chrome (production Next.js 15 app, Turbopack)
```

## Verification

clementantonyk:
## Description

When `useChat` is used with `enableThreadSignals: true`, user file attachments stored in `format: 2` messages were accumulated using the AI SDK v5 UI shape (`{ mediaType, url }`).

Applications rendering those messages with `toAISdkMessages(chat.messages, { version: 'v6' })` would crash when a `data-user-message` signal echoed a newly sent attachment. The v6 conversion path (`AIV5Adapter`) expects the database shape (`{ mimeType, data}`), so `part.data` was `undefined` and `createDataUri` threw.

This only affected the live accumulation path. Reloaded history continued to work because the server persisted attachments using the correct database shape.

This PR makes both live accumulation paths consistent with `fromCoreUserMessage`, which already emits the correct structure:

- **Signal echo path** (`toMessagePart`): file/image parts now use `{ mimeType, data }`
- **Stream chunk path** (`case 'file'`): assistant file chunks use the same shape

## Related issue(s)

Fixes #19356

## Type of change

- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [x] Test update

## Checklist

## Test plan

- [x] `npx vitest run src/lib/mastra-db/accumulator.test.ts` in `client-sdks/react` (111 tests pass)
- [x] Regression: signal file parts convert through `MessageList` → AI SDK v6 UI without crashing

## ELI5
When chats stream in, the app was building attachment “file parts” in the wrong internal shape. That mismatch could make the UI crash during live updates; this PR fixes the shape and adds tests so it won’t regress.

## Changes
- Fixed live attachment handling for `useChat` with `enableThreadSignals: true` by emitting canonical DB-shaped file parts (`{ type: 'file', mimeType, data, filename? }`) instead of the older `{ mediaType, url }` form:
  - Signal echo handling via `toMessagePart`
  - Assistant streaming file chunks via the `file` accumulator case
- Updated stream `file` chunk accumulation to build `data:<mime>;base64,...` strings (including converting `Uint8Array` payloads via `uint8ArrayToBase64`) and to include `providerMetadata` on emitted file parts.
- Added/updated regression tests to ensure signal-derived file parts can flow through `MessageList`, convert to AI SDK v6 UI messages without crashing, and preserve their `url` and `mediaType`.
- Added a patch Changeset for `@mastra/react`.
- Verified the accumulator test suite passes 111 tests.


abhiaiyer91:
🤖 Note from Mastra Code (on behalf of maintainer @abhiaiyer91):

We pushed a small follow-up commit onto this branch (01b08c3e73) — hope that's okay, maintainer edits were enabled.

**What it does:** now that the accumulator emits the canonical DB shape `{ mimeType, data }`, the playground's `UserFilePartRenderer` no longer needs its dual-shape fallback (`url ?? data`, `mediaType ?? mimeType`) or the comment saying the accumulator emits the V5 shape. The commit removes that dead code and updates the renderer tests to cover the DB shape only (the V5-shape test cases with unique coverage were converted rather than deleted).

**Verification:** all 10 renderer tests pass, typecheck and lint are clean. No change to your `@mastra/react` fix itself — thanks for the thorough root-cause work and the regression test that exercises the real `MessageList` conversion path!

abhiaiyer91 (approved):

PR #19370, opened 2026-07-13
feat(storage): support caller-defined dataset IDs

NikAiyer:
Allows callers to provide a durable dataset ID through the existing dataset creation API. Storage adapters atomically insert the dataset or return the compatible record that already owns the ID, so concurrent retries converge without a separate resolution API.

```ts
const dataset = await mastra.datasets.create({
  id: durableDatasetId,
  name: 'Candidate evaluation',
  organizationId,
  projectId,
});
```

Reusing an ID with different immutable tenancy or candidate identity fields now throws `DATASET_ID_CONFLICT`. Mutable fields are neither compared nor overwritten, ordinary generated-ID creation is unchanged, and deleting a dataset releases its ID for a fresh generation.

The shared storage contract covers concurrent creation, compatible retries, null/undefined normalization, immutable conflicts, and delete/recreate behavior across the built-in adapters.

Verified with the core dataset suite, core typecheck, adapter builds and lint, and the full LibSQL integration suite. PostgreSQL, MySQL, MongoDB, and Spanner integration tests require a running Docker daemon and were not available locally.

## ELI5

You can now choose the “dataset ID” when creating a dataset. If you accidentally create the same dataset again with the same ID, Mastra will just give you back the one it already saved—unless that same ID is being used for a different (immutable) owner/identity, in which case it stops and reports a conflict.

## Summary

- Extended `mastra.datasets.create()` / `DatasetsManager.create` to accept an optional caller-provided durable `id` for idempotent dataset creation.
- Updated `CreateDatasetInput` so storage adapters atomically insert-or-return the existing dataset when `id` is provided (no separate resolution API needed).
- Added validation for empty `id` values (`DATASET_INVALID_ID`) and immutable conflict handling when reusing an ID with different immutable tenancy/identity fields, throwing `DATASET_ID_CONFLICT` (mutable fields are not compared/overwritten).
- Preserved existing generated-ID behavior; deleting a dataset releases its ID for reuse (delete/recreate resets versioning/timestamps).
- Implemented adapter-level compatible retry behavior by detecting “already exists/duplicate” errors and then loading the existing record by `id` and resolving it via the shared immutable-field logic.
- Added shared contract tests covering concurrency, compatible retries, null/undefined normalization, immutable conflicts, and delete/recreate across built-in adapters (including core suite and adapter/typecheck/Lint + full LibSQL integration suite).
- MySQL adapter internals: introduced an insert-only statement path (`insertOnly` / `prepareInsertOnlyStatement`) to decouple ID-conflict handling from upsert SQL formatting.
- Updated release notes and bumped storage package peer dependency floors to `@mastra/core >= 1.51.0-0`.


packages/core/src/storage/domains/datasets/base.ts
```diff
@@ -21,6 +22,8 @@ import type {
 } from '../../types';
 import { StorageDomain } from '../base';
 
+const DATASET_IMMUTABLE_FIELDS = ['organizationId', 'projectId', 'candidateKey', 'candidateId'] as const;
```
YujohnNattrass, line 25: are candidatekey and candidateid actual fields? I believe they shouldn't be here

YujohnNattrass (approved):

abhiaiyer91 (approved):

PR #19371, opened 2026-07-13
feat(core): add durable flag to AgentConfig

taofeeq-deru:
Lets users opt an Agent into DurableAgent execution via config instead of wrapping every agent with createDurableAgent. Setting durable: true (or an options object) on new Agent({...}) makes Mastra.addAgent auto-wrap the agent with createDurableAgent at registration time, reusing the existing DurableAgentLike registration path (workflow registration, __setMastra, channel routes).

The base Agent class stays free of durable/workflow/pubsub imports — it just stores the raw option and exposes a getter. Wrapping is centralized in Mastra.addAgent so it runs exactly once per registration and does not double-wrap already-durable agents.

To break the resulting load-time cycle
(agent -> mastra -> create-durable-agent -> durable-agent -> agent), the Mastra value import inside agent.ts is now lazy (dynamic import inside the async ephemeral-mastra path). The type import is preserved.

## Type of change

- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [ ] Test update

## Checklist

## ELI5
You can now set `durable: true` on an `Agent` and, when you add it to `Mastra`, Mastra will automatically “wrap” it for durable execution. That way the agent can keep working reliably (e.g., after restarts) without you manually setting up the durable wrapper.

## Summary
- Added `AgentConfig.durable`, allowing `durable: true` or a durable options object to opt agents into durable execution.
- Updated `Mastra.addAgent()` to auto-wrap truthy `durable` agents with `createDurableAgent` during registration, reusing the existing durable registration path (workflow registration, `__setMastra`, and channel routes) and avoiding double-wrapping for already-durable agents.
- Enhanced the base `Agent` to store the raw durable option and expose it via a `durable` getter, while avoiding durable/workflow/pubsub load-time imports (including a lazy runtime `Mastra` import in `agent.ts` to prevent a dependency cycle).
- Extended durable options to support `cleanupTimeoutMs` (forwarded into `DurableAgentConfig`) via `CreateDurableAgentOptions`.
- Introduced the exported `AgentDurableOption` type and documented the new `durable` config flag in the durable agent reference docs (“Using the `durable` config flag”).
- Added tests covering auto-wrapping, option forwarding, correct no-op when omitted/false, idempotency (no double wrapper), `mastra.addAgent(new Agent({ durable: true }))`, and the `Agent.durable` accessor.
- Added release documentation via changesets for `@mastra/core` (new durable agent config flag) and a separate `@mastra/dynamodb` patch note fixing `hasMore` pagination behavior when included/context messages already cover the remaining thread messages.


packages/core/src/agent/agent.ts
```diff
@@ -6542,6 +6568,7 @@ export class Agent<
     if (this.#ephemeralMastra) {
       return this.#ephemeralMastra;
     }
+    const { Mastra: MastraClass } = await import('../mastra');
```
abhiaiyer91: what is this? was it already here?

taofeeq-deru: no, i added it. there was a circular import issue introduced when I added the auto-wrap logic to Mastra.addAgent, I imported createDurableAgent statically at the top of `packages/core/src/mastra/index.ts`. That created this cycle:
 ```
    agent.ts
      → imports { Mastra as MastraClass } from '../mastra'   (was static)
    mastra/index.ts
      → imports { createDurableAgent } from '../agent/durable/create-durable-agent'
    create-durable-agent.ts
      → imports { DurableAgent } from './durable-agent'
    durable-agent.ts
      → extends Agent   ← Agent module is still mid-loading, so `Agent` is `undefined`
  ```
  
  At runtime this blew up in ESM with:

  ```
    TypeError: Class extends value undefined is not a constructor or null
  ```

taofeeq-deru: The whole block where MastraClass is being used will be removed soon though, when we undo evented workflows in regular agent

abhiaiyer91 (approved):

PR #19375, opened 2026-07-13
fix(core): persist processor data chunks

roaminro:
Custom data chunks emitted from output processors were streamed to clients but skipped by message persistence, so data such as trace references disappeared from thread history.

Previously, this streamed the chunk without saving it:

\`\`\`ts
await writer.custom({
  type: 'data-trace-context',
  data: { traceId, spanId },
})
\`\`\`

With this change, the same chunk is stored with the assistant message. Chunks explicitly marked with \`transient: true\` remain stream-only.

This also preserves response message IDs through processor workflows, covers direct and tool-derived processor chunks with regression tests, and documents persistence and transient usage.

## ELI5

When an AI response processor adds “custom data chunks,” those chunks now get saved into the conversation (as part of the assistant message) instead of only being sent to the client during streaming. Chunks marked `transient: true` still stream live, but they’re not saved.

## What changed

- Persist non-transient custom `data-*` chunks emitted by output processors onto the assistant message parts in `MessageList` (durable storage), rather than streaming only.
- Keep message identity consistent by threading `messageId` through processor/workflow writer paths so the processor persistence logic can reuse the correct IDs.
- Apply the same persistence behavior to both:
  - direct processor-emitted custom chunks (during output streaming and final output processing), and
  - custom chunks produced while processing tool/data parts (with `transient: true` still remaining stream-only).
- Update documentation to clarify `writer.custom()` behavior and how `memory` affects persistence (saved unless `transient: true`).

## Tests & documentation

- Added regression tests covering persistence and recall for custom chunks from `processOutputStream` / `processOutputResult`, plus tool-derived processor chunks via `processDataParts`, including verification of `transient: true` behavior.
- Updated processor interface reference docs to reflect the persistence/transient rules and the `writer.custom()` output shape expectations.
- Added a patch changeset for `@mastra/core` describing the persistence fix.

### DIFF
--- a/.changeset/jolly-ears-spend.md
+++ b/.changeset/jolly-ears-spend.md
@@ -1,0 +1,5 @@
+---
+'@mastra/core': patch
+---
+
+Fixed custom data chunks emitted by processors so they are saved with thread messages unless marked transient.
--- a/docs/src/content/en/reference/processors/processor-interface.mdx
+++ b/docs/src/content/en/reference/processors/processor-interface.mdx
@@ -646,7 +646,7 @@ processLLMRequest?(
     name: 'writer',
     type: 'ProcessorStreamWriter',
     description:
-      'Stream writer for emitting custom data chunks during streaming. Use `writer.custom()` to send transient UI signals.',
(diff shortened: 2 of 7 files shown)


sgarfinkel:
@roaminro any chance we can get a review on this?

PR #19383, opened 2026-07-13
feat(core): bound Retry-After delays in StreamErrorRetryProcessor

YujohnNattrass:
Adds bounded `Retry-After` handling to `StreamErrorRetryProcessor` and keeps retryable scorer judge failures inside the failed request. Provider and transport failures no longer trigger structured-output JSON recovery, so completed scorer steps, trace targets, and score writes are not rerun. Legacy judge adapters retain their separate `generateLegacy()` retry settings.

Configure bounded provider pacing with:

```ts
new StreamErrorRetryProcessor({
  maxRetries: 2,
  maxRetryAfterMs: 30_000,
})
```

Verified with `pnpm build:core`, focused scorer and processor tests (111 passing), core type checking, and documentation validation. `pnpm test:core` still has unrelated live-provider and environment failures.

## ELI5

If a model call fails for a moment, the system now waits a little (but not too long) and retries only that specific failed piece, instead of redoing work that already finished. It also makes sure it doesn’t switch into the wrong kind of “JSON recovery” when the failure is really a provider/transport/network problem.

## What changed

- Added bounded `Retry-After` handling to `StreamErrorRetryProcessor`.
  - New `maxRetryAfterMs` option (default: 30 seconds) caps provider-controlled delays.
  - Parses `Retry-After` from the error/cause chain (supports numeric values and HTTP-date formats).
  - Computes the wait as the max of configured backoff delay vs capped `Retry-After`, and stops promptly on abort.
- Scoped scorer judge retries so only the failed judge request is retried.
  - Completed scorer steps, trace target writes, score writes, and sibling batch targets are not rerun.
  - Legacy `generateLegacy()` adapters keep separate retry settings/behavior.
- Tightened structured-output JSON fallback to only formatting-type failures.
  - Fallback/retry occurs for `JSONParseError`, `NoObjectGeneratedError`, and `TypeValidationError`.
  - Provider/transport/network errors are rethrown without triggering structured-output JSON recovery.
- Updated docs to clarify retry budgets, abort behavior, and the legacy vs current judge error-processor behavior.
- Added/expanded tests covering retry timing, `Retry-After` parsing/capping/abort behavior, scorer-step isolation (no reruns of completed work), and structured-output fallback vs rethrow rules.


docs/src/content/en/docs/agents/processors.mdx
```diff
@@ -862,10 +862,16 @@ const agent = new Agent({
 
 The retry mechanism:
 
-- Only works in `processOutputStep()` and `processInputStep()` methods
+- Works in `processOutputStep()` and `processInputStep()` methods
 - Replays the step with the abort reason added as context for the LLM
 - Tracks retry count via the `retryCount` parameter
-- Respects `maxProcessorRetries` limit on the agent
+- Requires an explicit `maxProcessorRetries` limit on the agent or call
```
YujohnNattrass, line 868: Documentation clarification only: this does not change input/output processor retry behavior. `abort(..., { retry: true })` already required `maxProcessorRetries`; the adjacent section documents the separate `processAPIError()` default of 10 retries.

abhiaiyer91 (approved):

packages/core/src/processors/stream-error-retry-processor.ts
```diff
@@ -36,9 +36,16 @@ export type StreamErrorRetryProcessorOptions = {
    * existing behavior for consumers that do not configure a delay.
    */
   delayMs?: StreamErrorRetryDelayMs;
+  /**
+   * Maximum provider-controlled Retry-After delay in milliseconds. Invalid values
```
NikAiyer, line 40: Non-blocking: Could we explicitly document that setting `maxRetryAfterMs` to `0` ignores provider-supplied `Retry-After` delays while still retaining delays configured through `delayMs`? The current text implies this, but the zero-value behavior is useful enough to state directly.

packages/core/src/processors/stream-error-retry-processor.ts
```diff
@@ -240,9 +296,10 @@ export class StreamErrorRetryProcessor implements Processor<'stream-error-retry-
...
     const effectiveDelay = policy.delayMs ?? this.#delayMs;
-    if (effectiveDelay !== undefined) {
-      await waitDelay(effectiveDelay, args, abortSignal);
-    }
+    const configuredDelayMs = effectiveDelay === undefined ? 0 : await resolveDelayMs(effectiveDelay, args);
+    const retryAfterMs = getRetryAfterMs(error);
+    const providerDelayMs = retryAfterMs === undefined ? 0 : Math.min(retryAfterMs, this.#maxRetryAfterMs);
+    await waitDelay(Math.max(configuredDelayMs, providerDelayMs), abortSignal);
 
     return { retry: true };
```
NikAiyer, line 304: Non-blocking question: If the signal aborts during `waitDelay()`, this continues and returns `{ retry: true }`. I assume the Agent runtime observes the same aborted signal and guarantees that another provider request is not issued. Could you confirm that contract? This appears to predate the `Retry-After` change, so I don’t consider it blocking here.

NikAiyer (approved):
The retry behavior remains bounded by both processor and Agent retry budgets, terminal errors remain excluded, and `Retry-After` only influences the abort-aware delay. Scorer retries are correctly scoped to the failed judge request, and the narrowed JSON fallback now handles only errors it can repair. The test coverage is strong, including retry exhaustion and avoiding reruns of completed scorer steps.

The currently failing CI checks appear unrelated to these changes, but required checks should still be green or formally waived before merging.

PR #19413, opened 2026-07-14
fix(core): clean up ephemeral Mastra hooks after standalone Agent execution

### LINKED ISSUE #19404: [BUG] Standalone Agent ephemeral Mastra leaks scorer hook after stream()
### Describe the Bug

Calling `stream()` on a short-lived standalone `Agent` creates an ephemeral
`Mastra`. The ephemeral instance registers an `onScorerRun` handler on Mastra's
module-level hook emitter.

When the Agent is discarded without later being registered on another Mastra,
the ephemeral instance is never torn down and its scorer hook is never
unregistered. The module-level emitter therefore keeps the ephemeral Mastra
reachable, causing linear heap growth when an application creates one Agent per
request.

This reproduces on `@mastra/core@1.50.1`.

### Root cause

`Mastra` registers a scorer hook in its constructor:

```ts
this.#onScorerHook = createOnScorerHook(this);
registerHook(AvailableHooks.ON_SCORER_RUN, this.#onScorerHook);
```

A standalone Agent creates an ephemeral Mastra during execution:

```ts
const ephemeral = new MastraClass({
  logger: false,
  storage: new InMemoryStore(),
  pubsub: new EventEmitterPubSub(),
});

await ephemeral.startWorkers();
this.#ephemeralMastra = ephemeral;
```

The hook is unregistered only if a real Mastra is subsequently attached through
`Agent.__registerMastra()`:

```ts
if (this.#ephemeralMastra) {
  this.#ephemeralMastra.__unregisterHooks();
  void this.#ephemeralMastra.stopWorkers().catch(() => {});
  this.#ephemeralMastra = undefined;
}
```

A short-lived standalone Agent never reaches this cleanup path. There is also no
public `Agent.dispose()` or equivalent lifecycle method available to the caller.

### Observed results

100 sequential standalone Agents, with a fully consumed mocked stream and forced
GC after every 20 iterations:

```text
mode=standalone iterations=100
baseline           heap=61.2 MB rss=159.9 MB
after 20           heap=73.8 MB rss=194.3 MB
after 40           heap=76.4 MB rss=201.2 MB
after 60           heap=79.2 MB rss=205.0 MB
after 80           heap=81.9 MB rss=209.5 MB
after 100          heap=84.7 MB rss=212.1 MB
final              heap=84.7 MB rss=212.1 MB
```

The equivalent control using one shared Mastra stabilizes after warm-up:

```text
mode=shared iterations=100
baseline           heap=61.2 MB rss=162.5 MB
after 20           heap=70.8 MB rss=189.4 MB
after 40           heap=70.6 MB rss=194.1 MB
after 60           heap=70.7 MB rss=196.0 MB
after 80           heap=70.8 MB rss=196.7 MB
after 100          heap=70.8 MB rss=197.0 MB
final              heap=70.8 MB rss=197.0 MB
```

A second control lets each Agent create its ephemeral Mastra and calls
`__registerMastra()` only after the stream. This triggers the existing
`__unregisterHooks()` cleanup branch and also stabilizes the heap.

### Relation to #18762 / #18763

This appears to be a remaining lifecycle case related to #18762 and its fix in
#18763.

That fix added hook deregistration when an internal or ephemeral Mastra is
explicitly torn down. It works when `__registerMastra()` is called later, but a
discarded standalone Agent does not trigger that teardown.

### Scheduler clarification

In `@mastra/core@1.47.0`, this scenario also retained one 10-second scheduler
interval per Agent.

That timer accumulation no longer reproduces in `1.50.1`, because idle ephemeral
Mastras no longer start the scheduler. The memory growth described here remains
present through the scorer-hook lifecycle.

### Steps To Reproduce

1. Clone the reproduction:

   ```bash
   git clone https://github.com/ffesseler/mastra-ephemeral-agent-memory-leak-repro.git
   cd mastra-ephemeral-agent-memory-leak-repro
   ```

2. Install the exact dependencies:

   ```bash
   npm install
   ```

3. Run the standalone-Agent reproduction:

   ```bash
   npm run repro
   ```

4. Observe that forced-GC heap usage continues increasing after warm-up.

5. Run the shared-Mastra control:

   ```bash
   npm run control
   ```

6. Observe that heap usage stabilizes after warm-up.

7. Run the late-cleanup control:

   ```bash
   npm run late-cleanup
   ```

8. Observe that triggering the existing ephemeral-Mastra cleanup branch also
   stabilizes heap usage.

No API key or external request is required. The Anthropic SSE response is mocked
locally.

https://github.com/ffesseler/mastra-ephemeral-agent-memory-leak-repro

### Expected Behavior

After a standalone Agent finishes streaming and is no longer referenced, its
internally owned ephemeral Mastra should be eligible for garbage collection.

Any global hooks registered by that ephemeral Mastra should be removed
automatically, or callers should have a supported public lifecycle method such
as `agent.dispose()`.

Repeated creation and disposal of standalone Agents should not cause linear heap
growth.

### Environment Information

```block
System:
  OS: macOS 26.5.2
  CPU: (14) arm64 Apple M4 Pro
  Memory: 3.43 GB / 48.00 GB
  Shell: 5.9 - /bin/zsh

Binaries:
  Node: 22.19.0
  npm: 10.9.3
  pnpm: 10.27.0
  bun: 1.3.6
  Deno: 2.9.2

npmPackages:
  @ai-sdk/anthropic: 4.0.2
  @mastra/core: 1.50.1

LLM Provider:
  Anthropic-compatible mocked SSE response; no external API call or API key.
```

### Verification

PiedPiper911:
## Problem

Calling `stream()` on a short-lived standalone `Agent` (one not registered on a `Mastra` via `{ agents }`) creates an ephemeral `Mastra` that registers an `onScorerRun` handler on the module-level mitt emitter. When the Agent is discarded without later being registered on a real Mastra, the ephemeral instance is never torn down and its scorer hook is never unregistered. The module-level emitter keeps the ephemeral Mastra reachable, causing linear heap growth.

The cleanup path only exists in `Agent.__registerMastra()`, which standalone Agents never reach.

Closes #19404

## Solution

Two complementary fixes:

**1. Automatic cleanup in `#execute()` finally block**

After each standalone Agent execution completes, clean up the ephemeral Mastra's scorer hook and stop its workers. The next call to `#getOrCreateEphemeralMastra()` lazily rebuilds a fresh ephemeral Mastra.

```ts
if (!this.#mastra && this.#ephemeralMastra) {
  this.#ephemeralMastra.__unregisterHooks();
  void this.#ephemeralMastra.stopWorkers().catch(() => {});
  this.#ephemeralMastra = undefined;
}
```

**2. Public `Agent.dispose()` method**

Explicit teardown API for users who create one Agent per request:

```ts
const agent = new Agent({ ... });
const result = await agent.stream('hello');
// ... use result ...
agent.dispose(); // releases hook + workers
```

After `dispose()`, the Agent can still accept new calls — a fresh ephemeral Mastra is created on demand.

## Changes

| File | Change |
|------|--------|
| `packages/core/src/agent/agent.ts` | Add hook cleanup in `#execute()` finally block + `dispose()` public method |
| `packages/core/src/agent/__tests__/standalone-ephemeral-leak.test.ts` | 4 regression tests |

## Testing

New test suite (`standalone-ephemeral-leak.test.ts`):
- Standalone `stream()` cleans up scorer hook after completion
- `dispose()` explicitly releases hooks and workers
- Sequential standalone Agents do not accumulate hooks
- Mastra-attached Agents do not create ephemeral Mastra (no leak vector)

All 4 new tests pass. Existing related tests (`scorer-hook-teardown.test.ts` — 3 tests, `runscope-leak.test.ts` — 6 tests) remain green.

## Notes for Reviewer

- The `#execute()` cleanup is the primary safety net — it covers the common "create Agent, call stream once, discard" pattern without requiring users to remember to call `dispose()`.
- `dispose()` is the explicit opt-in for users who want deterministic teardown timing (e.g., in a `finally` block of their request handler).
- Both paths are idempotent — `__unregisterHooks()` is already idempotent, and we set `#ephemeralMastra = undefined` to prevent double cleanup.
- The ephemeral Mastra rebuild on next call adds negligible overhead (~1 Mastra construction + `startWorkers`), acceptable for standalone usage which is already non-performance-critical.

## ELI5

When you create a standalone `Agent` (not attached to a `Mastra`), it sometimes created a temporary (“ephemeral”) `Mastra` that kept extra hidden hook registrations around. This PR prevents those ephemeral instances from registering the global scorer hook, so they can be garbage-collected and you don’t see memory grow over time.

## What changed

- Ephemeral `Mastra` instances are now marked internally with `__ephemeral: true` and **skip registering** the module-level `ON_SCORER_RUN` scorer hook at construction time.
- Updated `Agent` behavior around ephemeral Mastra handling so that switching from ephemeral to a real `Mastra` still cleans up the cached ephemeral reference, but without relying on an explicit `Agent.dispose()` lifecycle API.
- Added test-only helper `__hookHandlerCount(hook)` to read module-level hook handler counts.
- Added a regression test suite for `#19404` that verifies:
  - Standalone `Agent.stream()` does **not** increase `ON_SCORER_RUN` handler count.
  - Repeating standalone `Agent` creation/execution does **not** accumulate handlers.
  - Post-execution rebuild paths (e.g., title generation after memory execution) also do **not** change handler count.
  - A real `Mastra` that has an attached agent registers exactly one `ON_SCORER_RUN` handler, and `mastra.__unregisterHooks()` returns it to baseline.
- Added a `@mastra/core` patch changeset describing the memory leak fix.

### DIFF
--- a/.changeset/sour-points-tell.md
+++ b/.changeset/sour-points-tell.md
@@ -1,0 +1,7 @@
+---
+'@mastra/core': patch
+---
+
+Fixed a memory leak where every discarded standalone agent (an `Agent` used directly without being registered on a `Mastra` instance) stayed reachable for the lifetime of the process. The internal Mastra instance created for standalone execution no longer registers a module-level scorer hook it can never use, so standalone agents are garbage-collected once discarded. Applications that create one agent per request no longer see linear heap growth.
+
+Fixes [#19404](https://github.com/mastra-ai/mastra/issues/19404).
--- a/packages/core/src/agent/__tests__/standalone-ephemeral-leak.test.ts
+++ b/packages/core/src/agent/__tests__/standalone-ephemeral-leak.test.ts
@@ -1,21 +1,26 @@
 /**
- * Regression test for #19404: standalone Agent ephemeral Mastra leaks
+ * Regression tests for #19404: standalone Agent ephemeral Mastra leaked its
  * scorer hook on the module-level mitt emitter.
  *
- * A standalone Agent (not attached to a Mastra via `{ agents }`) creates
- * an ephemeral Mastra on first `stream()`/`generate()` call.  Before the
- * fix, the scorer hook registered by the ephemeral Mastra was never
- * unregistered, causing the module-level emitter to retain a reference
- * and prevent garbage collection.
+ * A standalone Agent (not attached to a Mastra via `{ agents }`) creates an
+ * ephemeral Mastra on first `stream()`/`generate()` call. Before the fix, the
+ * ephemeral Mastra's constructor registered an ON_SCORER_RUN handler on the
+ * module-level emitter that was never deregistered, so the emitter retained
+ * the whole ephemeral graph for the process lifetime — one leaked Mastra per
+ * discarded standalone Agent.
  *
- * This suite verifies:
- * 1. Hooks are cleaned up automatically after #execute() completes
- * 2. The public dispose() method cleans up hooks explicitly
- * 3. Multiple sequential standalone Agents do not accumulate hooks
+ * The fix suppresses hook registration for ephemeral instances entirely
+ * (`__ephemeral: true`): the hook could never resolve a scorer on a
+ * registry-less instance anyway. These tests assert on the module-level
+ * handler count (`__hookHandlerCount`), so they fail if any standalone code
+ * path — including post-execution rebuilds like title generation — registers
+ * a handler it doesn't release.
  */
 import { afterAll, beforeAll, describe, expect, it, vi } from 'vitest';
-import { AvailableHooks, executeHook } from '../../hooks';
+import { __hookHandlerCount, AvailableHooks } from '../../hooks';
+import { MockMemory } from '../../memory/mock';
 import { Agent } from '../agent';
+import type { AgentExecutionOptions } from '../agent.types';
 import { convertArrayToReadableStream, MockLanguageModelV2 } from './mock-model';
 
 function makeMockModel() {
@@ -35,8 +40,15 @@ function makeMockModel() {
   });
 }
 
-async function flushHooks() {
-  await new Promise(resolve => setTimeout(resolve, 50));
+function scorerHookCount() {
+  return __hookHandlerCount(AvailableHooks.ON_SCORER_RUN);
+}
+
+async function streamAndDrain(agent: Agent, prompt: string, options?: AgentExecutionOptions) {
+  const result = options ? await agent.stream(prompt, options) : await agent.stream(prompt);
+  for await (const _ of result.fullStream) {
+    // drain
+  }
 }
 
 describe('standalone Agent ephemeral Mastra hook leak (#19404)', () => {
@@ -47,7 +59,9 @@ describe('standalone Agent ephemeral Mastra hook leak (#19404)', () => {
     vi.unstubAllEnvs();
   });
 
-  it('clean up scorer hook after standalone stream() completes', async () => {
+  it('standalone stream() does not register a module-level scorer hook', async () => {
+    const baseline = scorerHookCount();
+
     const agent = new Agent({
       id: 'standalone-leak-test',
       name: 'Standalone Leak Test',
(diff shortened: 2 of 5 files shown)


abhiaiyer91:
Thanks @PiedPiper911 — great diagnosis, and the repro in #19404 checked out exactly as described. We verified the leak locally (linear heap growth with one standalone agent per request) and confirmed the scorer hook on the module-level emitter was the pin.

While reviewing, we found a few things and pushed a follow-up commit (225cb280a2) on top of yours rather than asking for another round trip. What's different and why:

**1. The fix moved from teardown to the constructor.** The finally-block cleanup in `#execute()` handles the main path, but the ephemeral Mastra is lazily *rebuilt* by post-execution code: title generation (`generateTitleFromUserMessage` → `#getOrCreateEphemeralMastra()`), snapshot loading, and `listSuspendedRuns()`. Title generation is fire-and-forget and runs *after* the finally block, so a standalone agent **with memory** re-registered a hook that was never released — one leaked Mastra graph per request, same as before. Instead of chasing every rebuild site, the ephemeral Mastra now passes an internal `__ephemeral: true` flag and the constructor skips hook registration entirely. Nothing is lost: `findScorer()` resolves via agent/editor/scorer registries, which are all empty on an ephemeral instance, so the hook could only ever throw `MASTRA_SCORER_NOT_FOUND`.

**2. The tests were rewritten to actually detect the leak.** The `expect(() => executeHook(...)).not.toThrow()` pattern can't fail — `executeHook` defers via `setImmediate`, so handler exceptions never reach the expect. We confirmed 3 of the 4 original tests passed with the fix reverted. The new tests assert on the module-level handler count (via a test-only `__hookHandlerCount()` export) and include the memory + title-generation case. With the flag removed, 4 of 5 fail; with it, all pass.

**3. `dispose()` was dropped.** With no global registration, the ephemeral graph holds no external references (we checked: no timers, no process pins, workers only subscribe to the ephemeral's own pubsub) — discarded standalone agents are fully GC-collectable with zero ceremony. That means no new public API is needed for this bugfix, and users won't cargo-cult `finally { dispose() }` unnecessarily. If a lifecycle API is wanted later, it can be designed deliberately.

Also added the missing `@mastra/core` patch changeset.

Your commit and test scaffolding are kept in the history — thanks for tracking this down and for the clean writeup. 🙏

*(Posted by Mastra Code on behalf of @abhiaiyer91)*

abhiaiyer91 (approved):

PR #19419, opened 2026-07-14
fix(playground): keep trace panels independently scrollable

### LINKED ISSUE #19185: [FEATURE] Constrain Studio trace detail panels to viewport height
### Problem Statement

The **Studio trace detail page** can grow far beyond the browser viewport when a trace has many spans.

The **Trace Timeline** card and **Span Details** card should each scroll inside the available page space. Instead, the parent layout can expand to the full height of the timeline content.

This makes the page hard to use because the browser page becomes extremely tall instead of behaving like two fixed-height panels.

### Proposed Solution

Update the **trace detail page layout** so the content row is constrained to the remaining viewport height.

The desired behavior:

- The **main page area** stays within the browser viewport.
- The **Trace Timeline** card scrolls independently.
- The **Span Details** card scrolls independently.
- Selecting a span does not make the whole page thousands of pixels tall.

Implementation shape:

- Use `grid-rows-[auto_minmax(0,1fr)]` for full-height page layouts.
- Add `min-h-0` where scrollable grid or flex children need to shrink.
- Change the trace detail two-column grid from `items-start` to `items-stretch`.
- Give the right-side details column `h-full`.
- Ensure `DataPanel.Content` can shrink before scrolling.

### Component

Observability, Studio

### Alternatives Considered

A live browser-only CSS patch was tested first to confirm the behavior before making a persistent code change.

That proved the layout issue was not caused by trace data rendering itself. It was caused by the parent grid row expanding to fit child content.

### Example Use Case

As a developer, I want to inspect a long trace in Mastra Studio so that I can compare the timeline and selected span details without losing my place.

For example:

1. Open a trace with many spans (more than visible without scrolling).
2. Select a span in the timeline.
3. Scroll the left timeline while keeping the right span details visible.
4. Scroll the right details panel without moving the whole page.

### Additional Context

Here are the instructions provided by my AI session on how to implement the change:
https://gist.github.com/coderberry/34571e742029186b98df7a9d3ea9a5b2

I also created a chrome extension to resolve this without modifying the Mastra code:
https://github.com/berrydev-ai/mastra-studio-ui-enhancements

### Verification

damien-schneider:
Fixes #19185.

Long traces could make the trace page grow beyond Studio's viewport. Opening a deep span moved the whole Studio content area and hid both panel headers. This keeps the trace layout within the viewport and lets the timeline and span details scroll independently. The containment is scoped to this route and does not add an outer overflow-hidden.

Inspired by the [mastra-studio-ui-enhancements](https://github.com/berrydev-ai/mastra-studio-ui-enhancements) repository shared by @coderberry and its [integration gist](https://gist.github.com/coderberry/34571e742029186b98df7a9d3ea9a5b2). Thanks for reporting this and sharing the initial direction.

Before, using a 61-span reproduction on main:

Current main after revealing the final span, with both trace panel headers above the viewport

The page moved to reveal span 60, pushing both panel headers out of view.

After:

Fixed trace route with both panel headers visible and the timeline scrolled to span 60

Both headers remain visible and the timeline scrolls to the selected span.

After scrolling the span details:

Fixed trace route with the details panel scrolled while the timeline remains on span 60

Only the details panel moves. The timeline stays at the same position.

A focused Chromium regression covers the overflowing layout and independent scrolling. The playground and playground-ui builds, lint, and typechecks also pass.

## ELI5

Long trace pages no longer make the whole Studio page grow endlessly. The timeline and span details now fit within the screen and scroll separately.

## What changed

- Constrained the trace detail layout to the available viewport height.
- Added shrinkable grid and flex containers so panels remain within their layout.
- Enabled independent scrolling for the timeline and span details.
- Updated `DataPanelContent` to support constrained scrolling.
- Added deterministic Chromium coverage for long traces with 60 tool-call spans.
- Regression tests verify panel visibility, independent scrolling, and no page-level overflow.
- Added patch changesets for the affected packages.

### DIFF
--- a/.changeset/sunny-breads-kiss.md
+++ b/.changeset/sunny-breads-kiss.md
@@ -1,0 +1,5 @@
+---
+'@internal/playground': patch
+---
+
+Fixed trace detail pages so long timelines and span details scroll independently within the viewport.
--- a/.changeset/tender-lizards-push.md
+++ b/.changeset/tender-lizards-push.md
@@ -1,0 +1,5 @@
+---
+'@mastra/playground-ui': patch
+---
+
+Fixed data panel content so it scrolls correctly inside constrained layouts.
--- a/packages/playground-ui/src/ds/components/DataPanel/data-panel-content.tsx
+++ b/packages/playground-ui/src/ds/components/DataPanel/data-panel-content.tsx
@@ -7,7 +7,7 @@ export interface DataPanelContentProps {
 
 export function DataPanelContent({ children, ref }: DataPanelContentProps) {
   return (
-    <div ref={ref} className="flex-1 overflow-y-auto p-4">
+    <div ref={ref} className="min-h-0 flex-1 overflow-y-auto p-4">
       {children}
     </div>
   );
--- a/packages/playground/e2e/tests/traces/$traceId/page.spec.ts
+++ b/packages/playground/e2e/tests/traces/$traceId/page.spec.ts
@@ -1,9 +1,85 @@
+import type { MastraClient } from '@mastra/client-js';
+import { SpanType } from '@mastra/core/observability';
 import { test, expect } from '@playwright/test';
 import type { Page } from '@playwright/test';
 import { resetStorage } from '../../__utils__/reset-storage';
 import { expectBreadcrumbLink, expectCurrentBreadcrumb, expectRouteDocsLink } from '../../__utils__/route-header';
 
 const FAKE_TRACE_ID = 'trace-does-not-exist';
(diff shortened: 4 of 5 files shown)


damien-schneider:
Thanks @coderberry for reporting this and sharing the [mastra-studio-ui-enhancements](https://github.com/berrydev-ai/mastra-studio-ui-enhancements) chrome extension repo 🙏
It gave us a very useful starting point for reproducing the long-trace case and bringing the fix into Studio

abhiaiyer91 (approved):

PR #19421, opened 2026-07-14
fix(deployer): thread transpilePackages through build and dev bundler paths

### LINKED ISSUE #19380: [BUG] bundler.transpilePackages is silently ignored by mastra build (and mastra dev)
### Describe the Bug

`bundler.transpilePackages` is a documented config option, but setting it currently has no effect in either mode:

- **`mastra build`**: `_bundle` constructs the internal bundler options from only `sourcemap`, `externals`, `enableEsmShim`, and `dynamicPackages` (`packages/deployer/src/bundler/index.ts`, `internalBundlerOptions`). The internal `BundlerOptions` interface (`packages/deployer/src/build/types.ts`) has no `transpilePackages` field at all, so the value structurally cannot reach `bundleExternals` — whose `transpilePackages` parameter silently defaults to `[]`.
- **`mastra dev`**: `DevBundler.watch` (`packages/cli/src/commands/dev/DevBundler.ts`) reads the user's bundler options but forwards only `sourcemap` into `getWatcherInputOptions`. PR #15630 threaded the option through for dev and was closed without landing; there's an unanswered follow-up comment on that PR asking why.

The reason most people never notice: all workspace package names are automatically added to the transpile set regardless (`bundleExternals.ts`: `new Set([...transpilePackages, ...workspacePackagesNames])`). So for monorepo-internal packages the auto-transpilation covers you — but that also means the user-facing option does nothing in every mode, while the types and docs suggest it works.

What this blocks: depending on a regular npm package that ships TypeScript source (not a workspace package). The documented knob for that case is a no-op, so the build fails on TS syntax with no supported recourse. It also creates false confidence — teams set the option, see green builds (because auto-transpilation happened to cover their workspace packages), and only discover the gap when a non-workspace TS dependency enters the graph.

### Steps To Reproduce

1. Add a dependency that ships `.ts` source to a Mastra project (any non-workspace package whose entry points at TypeScript).
2. Set `bundler: { transpilePackages: ['that-package'] }` in the `Mastra` config.
3. Run `mastra build` — the option never reaches the bundler (a breakpoint in `bundleExternals` shows `transpilePackages: []`), and the build fails on the untranspiled source.
4. Same for `mastra dev`.

The drop is also verifiable by inspection at the three seams listed above (`internalBundlerOptions`, the `BundlerOptions` interface, `DevBundler.watch`).

Inline above; the option loss is statically visible in the code paths referenced. Happy to push a repro repo if the team prefers one.

### Expected Behavior

One of:

1. The option is threaded through both `mastra build` and `mastra dev` (reviving the intent of #15630), so non-workspace TS-shipping dependencies can be transpiled; or
2. The option is validated and the CLI warns loudly that it is ignored (or redundant, when every listed package is a workspace package already covered by auto-transpilation).

Either is a big improvement over documented config that silently does nothing. I'm happy to contribute whichever direction maintainers prefer — flagging it as an issue first since workspace auto-transpilation may be the intended long-term answer (#6852).

### Environment Information

```
@mastra/deployer: current main (2026-07-13)
Node: 20.x
OS: macOS 15 / Linux (both)
```

### Verification

shreyas1903:
Fixes #19380

## Problem
`transpilePackages` is a documented `Config['bundler']` option but had no 
effect in either `mastra build` or `mastra dev`.

## Root cause (three structural drop points)
1. `BundlerOptions` (`build/types.ts`) had no `transpilePackages` field, so 
   `internalBundlerOptions` (`bundler/index.ts`) couldn't carry the value 
   even though `getUserBundlerOptions()` already returns it.
2. `analyzeBundle`'s parameter type explicitly `Pick<BundlerOptions, ...>` a 
   fixed list of fields that excluded `transpilePackages` — so it was 
   dropped at this boundary even after fixing (1).
3. `DevBundler.watch` only forwarded `{ sourcemap }` to 
   `getWatcherInputOptions`, never `bundlerOptions` at all — despite 
   `getInputOptions` already accepting and forwarding a full `BundlerOptions` 
   object through to `analyzeBundle`.

All workspace packages are auto-added to the transpile set regardless 
(`bundleExternals.ts`), which is why this was easy to miss — the option 
only matters for non-workspace npm packages that ship TypeScript source, 
where it silently did nothing.

## Fix
- Add `transpilePackages?: string[]` to `BundlerOptions`
- Include `'transpilePackages'` in `analyzeBundle`'s `Pick<...>`
- Forward it in `internalBundlerOptions` (`mastra build` path)
- Build a full `bundlerOptions` object in `DevBundler.watch` instead of 
  only `sourcemap` (`mastra dev` path)

## Tests
Added a regression test in `analyze.test.ts` asserting `analyzeBundle` 
accepts `transpilePackages` in `bundlerOptions` without dropping it 
(type-level regression guard against #19380 recurring).

**Note:** I could not run the full local test suite in this checkout — 
`@mastra/core` fails to build via `turbo` due to a pre-existing, unrelated 
TS 6.0 `baseUrl` deprecation error in `@internal/llm-recorder` (reproduces 
identically on unmodified `main`, confirmed via `git stash`). I verified 
the change instead via isolated `tsc` checks against all four edited files 
(no new errors beyond pre-existing, unrelated `logger` typing gaps) and by 
manually tracing all three drop points end to end.

## ELI5

Some npm packages contain TypeScript that needs to be converted before Mastra can use it. This change makes `bundler.transpilePackages` work consistently in both `mastra build` and `mastra dev`, and adds a test to prevent the issue from returning.

## Changes

- Added `transpilePackages` to `BundlerOptions`.
- Forwarded the option through build and dev bundling paths.
- Updated `analyzeBundle` to accept the option.
- Added a regression test for option propagation.
- Added a changeset for `@mastra/deployer` and `mastra`.

## Verification

- Isolated TypeScript checks were used.
- Full test execution was blocked by an unrelated TypeScript 6.0 `baseUrl` deprecation error.

### DIFF
--- a/.changeset/fix-transpile-packages-threading.md
+++ b/.changeset/fix-transpile-packages-threading.md
@@ -1,0 +1,6 @@
+---
+'@mastra/deployer': patch
+'mastra': patch
+---
+
+Fix `bundler.transpilePackages` being silently ignored by `mastra build` and `mastra dev`. The option is now threaded through `BundlerOptions`, `analyzeBundle`, and `DevBundler.watch` so non-workspace npm packages that ship TypeScript source can be transpiled as documented.
--- a/packages/cli/src/commands/dev/DevBundler.ts
+++ b/packages/cli/src/commands/dev/DevBundler.ts
@@ -102,6 +102,13 @@ export class DevBundler extends Bundler {
       },
       {
         sourcemap: sourcemapEnabled,
+        bundlerOptions: {
+          enableSourcemap: sourcemapEnabled,
+          enableEsmShim: true,
+          externals: bundlerOptions?.externals ?? true,
+          dynamicPackages: bundlerOptions?.dynamicPackages,
+          transpilePackages: bundlerOptions?.transpilePackages,
+        },
         analysisEntries: [entryFile, devServerAnalysisEntry],
       },
     );
--- a/packages/deployer/src/build/analyze.test.ts
+++ b/packages/deployer/src/build/analyze.test.ts
@@ -269,6 +269,55 @@ describe('external dependency versions', () => {
   }, 15000);
 });
 
+describe('transpilePackages threading (issue #19380)', () => {
+  it('accepts transpilePackages in bundlerOptions without dropping it', async () => {
+    await mkdir(tempRoot, { recursive: true });
+    const tempDir = await mkdtemp(join(tempRoot, 'mastra-transpile-packages-'));
+    tempDirs.push(tempDir);
+
+    const appDir = join(tempDir, 'apps', 'app');
+    const entryFile = join(appDir, 'index.ts');
+    const outputDir = join(appDir, '.mastra', '.build');
+
+    await mkdir(outputDir, { recursive: true });
+    await writeFile(join(tempDir, 'package.json'), JSON.stringify({ name: 'test-workspace', version: '1.0.0' }));
+    await writeFile(join(appDir, 'package.json'), JSON.stringify({ name: 'app', version: '1.0.0', type: 'module' }));
+    await writeFile(entryFile, `export const value = 1;`);
+
+    const originalCwd = process.cwd();
+    process.chdir(tempDir);
+    try {
+      // Before the fix, `transpilePackages` was outside the `Pick<BundlerOptions, ...>`
+      // accepted by analyzeBundle's type, and internalBundlerOptions
+      // (packages/deployer/src/bundler/index.ts) never read it off the user's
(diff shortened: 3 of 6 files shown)


LekoArts:
Hi, thanks for the PR! We'll handle this ourselves though as this is a delicate area of the codebase.

shreyas1903:
Understood, thanks for taking a look! Happy to have flagged the issue and the three drop points either way,  let me know if a smaller, scoped piece (e.g. just the BundlerOptions type addition, or the regression test) would still be useful as a separate PR.

PR #19338, opened 2026-07-13
feat(core): provider-aware actor authorizer for system-agent least privilege

tiffanybalc:
## Description

System / non-user actors (cron jobs, autonomous agents) currently pass the tenant-scope check but then skip provider authorization. In `requireFGA`, an `ActorSignal` returns after tenant validation without consulting the provider, so providers cannot express per-agent least privilege or deny an actor's access to a specific resource.

This adds an opt-in, backward-compatible hook so providers can authorize system actors:

- `IFGAProvider` gains optional `requireActor(actor, params)` (throws `FGADeniedError` to deny).
- `ActorSignal` gains optional `agentId`, `permissions` (`MastraFGAPermissionInput[]`, the actor analog of a user's resolved permissions), and `scope`.
- `requireFGA` delegates to `requireActor` when the provider implements it, after the existing tenant-scope check. When it does not, the legacy trusted-actor behavior is preserved.

The PR also closes related authorization gaps identified during review:

- Public agent routes normalize top-level and nested execution options so clients cannot supply a trusted `actor` or override `requestContext.organizationId`.
- Durable `stream`, `generate`, and `resume` authorize the effective merged defaults, run context, memory resource, and actor used by execution.
- Durable approval and decline routes validate run ownership and the suspended `toolCallId`, and preserve trusted authorization context through resume.
- Tool execution uses the canonical authorization wrapper instead of an additional check against a different resource ID.

Before — system actors had tenant scoping but no provider-defined per-agent authorization:

```ts
await agent.generate('run nightly report', {
  actor: { actorKind: 'system', sourceWorkflow: 'nightly' },
});
```

After — a provider can enforce least privilege on the acting agent:

```ts
import { FGADeniedError } from '@mastra/core/auth/ee';
import type { IFGAProvider } from '@mastra/core/auth/ee';

class MyFga implements IFGAProvider {
  // ...existing check / require / filterAccessible...
  async requireActor(actor, { resource, permission }) {
    const agentId = actor === true ? undefined : actor.agentId;
    // Resolve authoritative grants from a trusted source keyed by agentId.
    const granted = await this.grantsForAgent(agentId);
    // `permission` may be a single value or an array (needs ANY one of them).
    const required = Array.isArray(permission) ? permission : [permission];
    if (!required.some(p => granted.includes(p))) {
      throw new FGADeniedError(null, resource, permission, 'actor lacks required permission');
    }
  }
}

await agent.generate('run nightly report', {
[...]

PR #19429, opened 2026-07-14
fix(inngest): forward actor and unify durable trigger/resume event fields

### LINKED ISSUE #19426: createInngestAgent: actor signal dropped when triggering/resuming the durable agentic workflow
### Background

`@mastra/inngest@1.8.0`'s workflow engine (`run.ts`, `workflow.ts`, `execution-engine.ts` — added in #18674) forwards a trusted `ActorSignal` from `event.data.actor` through `engine.execute()` and across nested `inngestStep.invoke()` calls, so `Run.start()`/`.resume()` on the general Workflow API correctly threads a trusted `actor` across the durable boundary.

`createInngestAgent`'s durable agent wrapper (`workflows/inngest/src/durable-agent/create-inngest-agent.ts`) drives the same underlying agentic-loop workflow, but it doesn't go through `run.ts` — it sends its own bespoke Inngest events directly via `inngest.send()`, and #18674 never touched this file.

### Bug

In `create-inngest-agent.ts`:

- `triggerWorkflow()` (used by `stream()`/`generate()`) never included `actor` in the outbound event's `data`, even though `workflowInput.options.actor` is correctly populated during `prepareForDurableExecution`. The workflow function handler (`workflow.ts`, `let { ..., actor } = event.data`) reads `actor` straight off the top-level event data, so it was always `undefined` for this call path.
- `resume()` had no way to accept `actor` at all: `InngestAgentResumeOptions` didn't declare an `actor` field, and `resume()`'s own `inngest.send()` call likewise omitted it. There was no public API surface to re-supply `actor` on a resumed run — even though `actor` is documented (see the comment in `run.ts`) as a per-call trust signal that must be re-supplied on every resume, not persisted from the original run's snapshot.

### Impact

A consumer using `createInngestAgent` with a trusted system `actor` (e.g. `{ actorKind: 'system', sourceWorkflow: '...' }`) to bypass per-user FGA membership checks loses that signal on this path. Most concretely: there's no way to pass `actor` back in on `resume()` at all, which breaks any human-in-the-loop flow that suspends for tool approval and needs to resume as a trusted system actor. Depending on how an FGA provider is wired (see `requireFGA` in `packages/_internals/auth/src/ee/fga-check.ts`), a missing/incomplete actor+tenant-scope pairing can surface as `FGADeniedError: trusted actor requires organizationId / tenant scope` or similar.

### Repro

1. Wrap an agent with `createInngestAgent({ agent, inngest })`.
2. Call `durableAgent.generate(messages, { actor: { actorKind: 'system', sourceWorkflow: 'my-workflow' } })`.
3. The run suspends for tool approval; call `durableAgent.resumeGenerate(runId, resumeData)`.
4. There is no `actor` field on `InngestAgentResumeOptions` to pass it back in, and even a manual cast wouldn't help — `resume()`'s `inngest.send()` never forwarded it.

### Fix

I've implemented and tested a fix: forward `actor` in both `triggerWorkflow()`'s and `resume()`'s event payloads, and add an `actor` field to `InngestAgentResumeOptions`. Opening a PR against this issue shortly.

### Related

- #18674 — added engine-level `actor` threading to `@mastra/inngest`. This issue is the piece of that work that never made it into `createInngestAgent`'s own trigger/resume paths.
- #19064 — a related but distinct proposal about auto-propagating `actor` into agent/tool calls made *inside* a workflow step's own code. Not the same bug: that one assumes the engine-level path already works (true for `Run.start()`/`.resume()`), just not for `createInngestAgent`.

### LINKED ISSUE #19428: @mastra/inngest: durable trigger/resume event payloads are built twice (run.ts vs createInngestAgent) and keep drifting out of sync
`InngestRun` (the general Workflow API — `start`/`startAsync`/`resume`/`timeTravel` in `run.ts`) and `createInngestAgent`'s durable-agent wrapper (`triggerWorkflow`/`resume` in `durable-agent/create-inngest-agent.ts`) both fire the same `workflow.<id>` event via `inngest.send()`, as two completely independent implementations. Neither delegates to the other, even though the wrapper's workflow object is itself an `InngestWorkflow` with a working `createRun()`.

That duplication is why the same class of bug has shipped twice:

- `actor` (`ActorSignal`) was added to `run.ts`'s event payload in #18674. `create-inngest-agent.ts` was never updated to match — `triggerWorkflow()` never included it, and `resume()`'s options type had no `actor` field at all. (Previously reported narrowly as #19426.)
- `requestContext` was hardcoded to `{}` in both of `create-inngest-agent.ts`'s send calls, fixed narrowly in #19223 — but that fix only restored the persisted-context replay. `run.ts`'s resume already merges persisted context with fresh caller-supplied values (`_resumeAndSendEvent`); `create-inngest-agent.ts`'s `resume()` had no way to accept fresh `requestContext` at all until now.

Neither fix caught the other because each was scoped to exactly the field named in its own report, not to the duplication underneath. This isn't a timing accident either — `startAsync()` (the reusable fire-and-forget path on `Run`) has existed since 2025-12-11, four months before the durable-agent wrapper landed (2026-04-30), so the wrapper was hand-rolled independently rather than reusing what already existed.

**Fix:** pull the `requestContext`/`actor` payload construction — including the resume-merge logic — into one shared module (`durable-event-signals.ts`) that both `run.ts` and `create-inngest-agent.ts` call. This doesn't touch snapshot timing or control flow, just makes "which per-call signals ride on this event, and how" a single implementation instead of two, so a future addition can't silently miss one side again. PR incoming.

We considered having `createInngestAgent` delegate to `Run.startAsync()`/`.resume()` outright instead, which would prevent this even more thoroughly, but didn't do it here: `createRun()` does its own storage round-trip (existence check + conditional snapshot persist) and `startAsync()` does another unconditional snapshot persist, both adding latency to a path that today does neither, and neither has a clean seat for the wrapper's ordering requirement (spans created before trigger, wait for pubsub-subscription-ready before send). Flagging it in case there's appetite for that as a separate, bigger change.

tika:
## Description

Durable Inngest workflows were inconsistent about forwarding per-call `actor` and `requestContext` signals across start and resume boundaries.

`createInngestAgent` omitted `actor` from trigger events and could not accept a fresh actor or request context on resume. Actors are per-execution-segment trust signals: start uses the actor supplied for that start, resume uses only the actor resolved for the current resume, and tools consume that current workflow-segment actor. Persisted request context is still merged with fresh resume values, with fresh values winning.

After #19338 merged, this branch was updated against its FGA-aware core durable-agent implementation. Duplicate core implementation and its changeset were removed; this PR now focuses on the Inngest event boundary while retaining stronger regression coverage for actor replacement and omission at the core tool/FGA boundary.

## Changes

- Added shared `buildDurableTriggerFields()` and `buildDurableResumeFields()` helpers for Inngest event payloads.
- Routed the Workflow API and `createInngestAgent` start/resume paths through those helpers.
- Added `actor` and `requestContext` to `InngestAgentResumeOptions`.
- Forwarded the current actor on Inngest trigger and resume events without restoring an actor from persisted state.
- Added tests for request-context merging, actor replacement, actor omission, and the actual tool/FGA boundary.
- Added an `@mastra/inngest` patch changeset.

## Example

```ts
await inngestAgent.resume(runId, { approved: true }, { actor, requestContext });
```

## Related issue(s)

Fixes #19426
Fixes #19428

## Type of change

- [ ] Breaking change (fix or feature that would cause existing functionality change)
- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [x] Test update

## Checklist

- [N/A] I have made corresponding changes to the documentation (if applicable)

## ELI5

Durable Inngest workflows now use the actor supplied for each start or resume segment. They carry persisted request context forward while allowing fresh resume values to override it.

## What changed

- Forwarded actor signals when triggering and resuming durable Inngest workflows.
- Added `actor` and `requestContext` support to `InngestAgentResumeOptions`.
- Merged fresh resume request context over persisted context.
- Centralized durable Inngest event payload construction in shared helpers.
- Kept regression coverage for current-segment actor handling at the core tool/FGA boundary.
- Added an `@mastra/inngest` patch changeset.


graysonhicks:
One blocking issue: `resumeOptions.actor` reaches the Inngest event, but `createDurableToolCallStep()` still passes the persisted initial actor to tools. This ignores a new resume actor and can reuse trusted actor state when resume omits it. Please use the current workflow-segment actor (ensuring core/Inngest start + resume both supply it) and cover the actual tool/FGA boundary. Also add a short `resume(..., { actor, requestContext })` changeset example.

abhiaiyer91 (changes requested):
## Review: `fix(inngest): forward actor and unify durable trigger/resume event fields`

**Verdict: Request changes** — one concrete docs gap to close. The code itself is correct and I verified it end to end.

### What I verified

I checked out the branch and ran the affected suites in a clean sandbox:

- `@mastra/inngest` — `durable-event-signals.test.ts` + `__tests__/create-inngest-agent.test.ts`: **32 passed**, stable across 3 consecutive runs.
- `@mastra/core` — `durable-agent-actor.test.ts` + `tool-call-actor.test.ts`: **6 passed**, no type errors.
- `tsc --noEmit` on `@mastra/inngest`: clean.
- No merge conflicts; no install-time or test-time execution surface changed (no `package.json` scripts, lockfile, or CI workflow edits).

I also traced the full signal path rather than trusting the tests: `buildDurableTriggerFields()` / `buildDurableResumeFields()` results are spread into the `inngest.send()` payloads at all six call sites (`run.ts` 371/442/576/822, `create-inngest-agent.ts` 618/1014), and `workflow.ts` destructures `actor` off `event.data` at line 314 and hands it to `engine.execute()` at line 382. The wiring is genuinely closed, not just asserted.

### On the design

The centralization is the right call, and the module doc comment on `durable-event-signals.ts` explains *why* the duplication existed — that context is what stops this bug from recurring, and it's more valuable than the helpers themselves.

Three behavior details I checked specifically:

- **Trigger paths are semantically identical to before.** The old `requestContext: x ? Object.fromEntries(x.entries()) : {}` and `actor: x` pairs collapse into the spread with the same keys and same values, including `actor: undefined` still being an own property. No accidental key-shape change.
- **`serializeRequestContext()` duck-types on `.entries`,** which is what lets `create-inngest-agent.ts:619` pass the already-plain `workflowInput.requestContextEntries` through unchanged while `run.ts` passes a live `RequestContext`. That asymmetry is real and the helper handles it correctly.
- **`actor` is deliberately not read from the snapshot** on resume, and `packages/core/src/workflows/workflow.ts` `_resume` does the same. Persisting a membership-bypass signal into durable storage would be the actual bug; the negative test ("does not reuse the initial trusted actor when the resumed segment omits actor", asserting FGA rejects with *authenticated user is required*) is the most important test in this PR because it pins the security property rather than the happy path.

`create-inngest-agent.ts:618` reads `actor` from `workflowInput.options?.actor`, which `preparation.ts:614` populates — so top-level `actor` is redundant with `inputData.options.actor`. That redundancy is required, since the workflow handler only reads the top-level field. Not a problem, just confirming it's intentional and not a double-source-of-truth risk.

### Blocking: `resume()` docs are now inaccurate

`docs/src/content/en/reference/agents/inngest-agent.mdx:205` states:

> The third argument accepts `threadId` and `resourceId` in addition to the lifecycle callbacks

That sentence is now wrong, and the `PropertiesTable` below it (lines 207–220) omits both new public fields. `InngestAgentResumeOptions` gained `actor` and `requestContext` in this PR — these are user-facing API additions with non-obvious semantics (`actor` must be re-supplied every resume; `requestContext` merges over persisted with fresh winning). Those are exactly the rules a caller cannot guess, and the changeset is not a substitute for reference docs.

Repo `AGENTS.md` requires docs updates for feature work. Please add both rows and adjust the lead-in sentence. `docs/.../durable-agent.mdx:455-460` already has a good `actor` row to mirror:

```
{
  name: 'actor',
  type: "AgentExecutionOptions['actor']",
  isOptional: true,
  description: 'Per-call actor signal forwarded to FGA checks and tool execution. Not rehydrated from the suspended run — re-supply it on every resume.',
},
{
  name: 'requestContext',
  type: 'RequestContext',
  isOptional: true,
  description: 'Fresh request-context values merged over the suspended run\'s persisted context. Fresh values win on key collision.',
},
```

Note that the `Stream options` section (line 322) already inherits by reference via "the same agent execution options as `DurableAgent.stream()`", so `actor` on the start path is covered. Only the `resume()` table needs the change.

### Non-blocking: a pre-existing failure in this exact area

`src/actor-signal.test.ts > forwards actor into the nested-workflow RESUME invoke payload` fails on this branch (`expected 'failed' to be 'success'` at line 408). **This is not caused by this PR** — I confirmed it by restoring `run.ts` and `create-inngest-agent.ts` to their base-commit versions and removing `durable-event-signals.ts`, and the test still fails identically. It reproduces on `main`.

Worth flagging because it sits in the same actor-forwarding surface this PR hardens, and CI's affected-test detection didn't surface it. Please open a separate issue — it should not be fixed here and should not hold up this merge.

### Prior review signal

CodeRabbit's changeset finding (missing a usage example) is **already addressed** — `.changeset/many-readers-wish.md:10-14` now carries the `resume(runId, { approved: true }, { actor, requestContext })` example. Its colocation nitpick on `__tests__/create-inngest-agent.test.ts` doesn't apply: that file predates this PR, and adding cases to an existing suite is the right move over splitting them into a new colocated file. The two failing Vercel checks are `Authorization required to deploy`, not code failures.

---

Add the two `resume()` doc rows and this is good to merge.

abhiaiyer91:
Closing because the linked issue #19426 is closed. Thanks for the contribution.

PR #19454, opened 2026-07-15
fix(core): populate toolResults in onIterationComplete

### LINKED ISSUE #19453: [BUG] onIterationComplete returns empty toolResults after successful tool execution
## Describe the Bug

With `@mastra/core@1.50.1`, a server-side tool executes successfully, but the `onIterationComplete` callback receives the tool call with an empty `toolResults` array.

The completed step itself contains the tool result. Only the callback context loses it, which conflicts with the public `IterationCompleteContext.toolResults` contract.

This was observed with an Anthropic-backed agent, but the defect appears provider-independent because it is in the shared agentic-loop callback mapping.

## Steps To Reproduce

Mastra's existing agent-loop regression scenario is a minimal reproduction:

https://github.com/mastra-ai/mastra/blob/37f13b800239414a578cba578e2e64bfc791ba0d/packages/core/src/loop/test-utils/aimock/scenarios/iteration-complete.scenario.test.ts

In the first test, immediately after the existing `toolCalls` assertions, add:

```ts
expect(iterations[0].toolResults).toHaveLength(1);
expect(iterations[0].toolResults[0]).toMatchObject({
  id: 'call_add_1',
  name: 'add',
  result: { sum: 5 },
});
```

Run that scenario for the existing engines. The tool executes and the loop proceeds using its result, but the callback's `toolResults` is empty.

## Source-level cause

The loop already derives the actual results from newly-added message content:

```ts
const toolResultParts = currentContent.filter(
  part => part.type === 'tool-result',
);
```

It correctly assigns those parts to the completed step:

```ts
toolResults: toolResultParts as StepResult<Tools>['toolResults'],
```

However, `iterationContext` is constructed from a different field:

```ts
toolResults: (typedInputData.output.toolResults || []).map(...)
```

In this path, `typedInputData.output.toolResults` is empty even though `toolResultParts` contains the executed results.

Relevant implementation:

https://github.com/mastra-ai/mastra/blob/37f13b800239414a578e2e64bfc791ba0d/packages/core/src/loop/workflows/agentic-loop/index.ts#L125-L160

https://github.com/mastra-ai/mastra/blob/37f13b800239414a578e2e64bfc791ba0d/packages/core/src/loop/workflows/agentic-loop/index.ts#L180-L197

The callback contract explicitly documents results:

https://github.com/mastra-ai/mastra/blob/37f13b800239414a578e2e64bfc791ba0d/packages/core/src/agent/agent.types.ts#L203-L222

## Expected Behavior

`onIterationComplete(context)` should receive one normalized result for each tool that completed during that iteration. The callback should use the same actual result collection used by `currentStep.toolResults`, rather than `typedInputData.output.toolResults` when that field is not populated.

Concurrent tool calls should remain correlated by `toolCallId`.

## Environment Information

```
@mastra/core: 1.50.1
Node.js: 22
Execution API: Agent.generate
Tool type: server-side createTool
Observed provider: Anthropic
Also present on upstream main at commit 37f13b800239414a578cba578e2e64bfc791ba0d
```

## Verification

- [x] I searched open and closed issues for `onIterationComplete`, empty/missing `toolResults`, and the affected source expression.
- [x] The existing Mastra AIMock scenario provides a minimal reproduction; the missing assertion is included above.
- [x] No application or customer data is included in this report.

vedhsaka:
## Description

`onIterationComplete` read results from `typedInputData.output.toolResults`, which the agentic loop does not populate. Successful executions therefore appeared in `toolCalls` while `toolResults` remained empty.

This uses the current iteration's `tool-result` parts already collected by the loop, preserving each call ID and tool name while normalizing the AI SDK output envelope to the callback's documented `result` value.

The existing all-engine AIMock scenario now verifies the successful result on the tool iteration and an empty result list on the following text-only iteration.

## Related issue(s)

Fixes #19453

## Type of change

- Bug fix (non-breaking change that fixes an issue)
- Test update

## Checklist

## ELI5

When a tool finishes during an AI loop, the “iteration complete” callback should tell you what the tool actually returned. This fix makes that callback use the results from the just-finished iteration (so you don’t get an empty list), and it keeps each result tied to the correct tool call.

## Changes

- Fixed `onIterationComplete` to populate `iterationContext.toolResults` from the iteration’s collected `tool-result` parts (instead of `typedInputData.output.toolResults`, which could be empty/unset).
- Preserved correlation by keeping each item’s `toolCallId` and tool `name`.
- Normalized the tool output shape so the callback receives the documented `result` value.
- Updated the AIMock regression test to assert:
  - the first iteration includes the expected tool result (`call_add_1` → `{ sum: 5 }`)
  - the next iteration (text-only) returns an empty `toolResults` array.
- Added a patch changeset for `@mastra/core` documenting the fix.


vedhsaka:
@coderabbitai review

vedhsaka:
@coderabbitai full review

abhiaiyer91 (approved):

PR #19502, opened 2026-07-15
refactor: use Web Crypto randomUUID instead of node:crypto imports

### LINKED ISSUE #19501: Unify on Web Crypto crypto.randomUUID and drop node:crypto imports of randomUUID
## Problem Statement

The codebase generates UUIDs two different ways, often within the same package:

- 93 source files import `randomUUID` from `node:crypto` (or bare `crypto`)
- 35 files in `@mastra/core` alone already call the Web Crypto global (`crypto.randomUUID()` / `globalThis.crypto.randomUUID()`), alongside 27 core files that use the Node import

The two forms are functionally identical on every supported platform — the global has been available since Node 19, and the repo's engines floor is Node >= 22 — but the `node:crypto` import form has a real cost: it marks the importing module (and, through bundler chunking, every chunk it lands in) as Node-only. For edge/isolate runtimes that provide Web Crypto but no Node builtin modules (Convex default runtime, Cloudflare Workers without `nodejs_compat`), each of these imports is a bundling failure.

This was fixed for `@mastra/convex` in #19497 / #19498, which removed that package's only Node builtin dependency by switching to the global. The same inconsistency remains everywhere else. Distribution of the remaining `randomUUID`-from-`node:crypto` files:

| Area | Files |
|---|---|
| `packages/core` | 27 |
| `packages/memory` | 12 |
| `stores/*` (mongodb 12, spanner 5, mysql 4, pg/mssql/libsql…) | ~25 |
| `packages/server`, `agent-sdks/*`, `workflows/inngest`, `pubsub/*`, others | ~29 |

In `packages/core`, 10 of the importing files are on the agent inference path (`agent/agent.ts`, `mastra/index.ts`, `llm/model/router.ts`, `stream`/`loop` files), so the import form alone currently prevents bundling `Agent` for isolate runtimes even where no other Node functionality is used.

## Proposed Outcome

One convention repo-wide: UUID generation uses the Web Crypto global, and `randomUUID` is no longer imported from `node:crypto`. Other `node:crypto` APIs (`createHash`, HMAC/cipher usage in auth and channels) are out of scope — they have different portability trade-offs and deserve separate consideration.

A guard (lint rule or test, like the isolate-safety test added to `@mastra/convex` in #19498) would keep the convention from regressing.

## Component

**Core** (plus memory, stores, server — mechanical change repo-wide)

## Additional Context

Follow-up to #19497 / #19498. Part of making the agent runtime path bundleable for V8-isolate runtimes; related adapter-parity precedent: #17871.

ediskandarov:
## Description

Unifies UUID generation repo-wide on the Web Crypto global. Before this change the codebase used two forms side by side: 167 files imported `randomUUID` from `node:crypto` (or bare `crypto`), while 35 files in `@mastra/core` alone already called the global `crypto.randomUUID()`. The forms are identical on every supported platform — the global has existed since Node 19 and the engines floor is Node 22.

The import form has a real cost: it marks the importing module — and, through bundler chunking, every dist chunk it lands in — as Node-only. For V8-isolate runtimes that provide Web Crypto but no Node builtin modules (Convex default runtime, Cloudflare Workers without `nodejs_compat`), each such import is a bundling failure. This was fixed for `@mastra/convex` in #19498; this PR applies the same convention everywhere.

**The change is mechanical:**

- Remove `randomUUID` from `node:crypto`/`crypto` import specifiers (dropping the import entirely where it was the only specifier; other specifiers like `createHash` are kept).
- Rewrite bare `randomUUID(...)` call sites to `crypto.randomUUID(...)`.
- `packages/core/src/workflows/branch-map-bug.test.ts` mocked the `crypto` module for UUID determinism; that mock is dead once the source uses the global, and determinism is already provided by `@internal/test-utils/setup` (which stubs the global `crypto.randomUUID` with a per-test counter), so the local mock is removed.
- One stale JSDoc example import removed in `packages/mcp/src/server/server.ts`.

Out of scope, deliberately: other `node:crypto` APIs (`createHash`, HMAC/cipher usage in auth and channels) — different portability trade-offs, tracked separately.

## Related issue(s)

Fixes #19501

## Type of change

- [ ] Documentation update
- [x] Code refactoring
- [ ] Performance improvement
- [x] Test update

## Testing

- No leftover imports: repo-wide grep for `randomUUID` from `node:crypto`/`crypto` returns zero runtime matches (one type-only `import type { UUID }` remains, which is erased at build time).
- `pnpm turbo build` for `@mastra/core`, `@mastra/memory`, `@mastra/server`, `@mastra/mcp` (plus their dependency graphs): clean — proves every rewritten call site typechecks.
- `packages/memory` `tsc --noEmit`: clean.
- Targeted test runs: the rewritten `branch-map-bug.test.ts` (2 passed) and a `@mastra/memory` slice exercising UUID-generating paths (19 passed). Deterministic-UUID behavior in tests is unchanged since the shared setup already stubs the global.
- Changeset covers all 31 published packages touched (patch bumps).

## AI disclosure

Implemented with Claude Code (Fable 5) as a reviewed codemod; human-reviewed.

### DIFF
--- a/.changeset/sour-planets-post.md
+++ b/.changeset/sour-planets-post.md
@@ -1,0 +1,39 @@
+---
+'@mastra/voice-google-gemini-live': patch
+'@mastra/google-cloud-pubsub': patch
+'@mastra/agent-builder': patch
+'@mastra/nestjs': patch
+'@mastra/observability': patch
+'@mastra/redis-streams': patch
+'@mastra/cloudflare-d1': patch
+'@mastra/voice-aws-nova-sonic': patch
+'@mastra/claude': patch
+'@mastra/cursor': patch
+'@mastra/openai': patch
+'@mastra/deployer': patch
+'@mastra/cloudflare': patch
+'@mastra/inngest': patch
+'@mastra/couchbase': patch
+'@mastra/vectorize': patch
+'@mastra/editor': patch
+'@mastra/memory': patch
+'@mastra/server': patch
+'@mastra/acp': patch
+'@mastra/code-sdk': patch
+'mastracode': patch
+'@mastra/github-signals': patch
+'@mastra/mongodb': patch
+'@mastra/spanner': patch
+'@mastra/upstash': patch
+'@mastra/core': patch
+'@mastra/libsql': patch
+'mastra': patch
+'@mastra/mcp': patch
+'@mastra/rag': patch
+'@mastra/mssql': patch
+'@mastra/mysql': patch
+'@mastra/redis': patch
+'@mastra/pg': patch
+---
+
+Switched UUID generation to the Web Crypto global (crypto.randomUUID) instead of importing randomUUID from node:crypto. No behavior change on Node.js; removes an unnecessary Node-only module dependency so packages bundle more cleanly for edge and V8-isolate runtimes such as the Convex default runtime and Cloudflare Workers.
--- a/agent-sdks/acp/src/agent.ts
+++ b/agent-sdks/acp/src/agent.ts
@@ -1,4 +1,3 @@
-import { randomUUID } from 'node:crypto';
 import { ReadableStream } from 'node:stream/web';
 
 import type { ModelInfo, SessionUpdate } from '@agentclientprotocol/sdk';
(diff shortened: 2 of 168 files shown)


ediskandarov:
Closing: over the 150-file review limit. Split into two PRs (packages/* and adapters) — links follow.

ediskandarov:
Split: #19503 (packages/*, 85 files) and #19504 (stores/agent-sdks/pubsub/etc., 84 files).

PR #19504, opened 2026-07-15
refactor(adapters): use Web Crypto randomUUID instead of node:crypto (2/2)

### LINKED ISSUE #19501: Unify on Web Crypto crypto.randomUUID and drop node:crypto imports of randomUUID

ediskandarov:
## Description

Part 2 of 2 of the repo-wide UUID unification (split from #19502, which exceeded the 150-file review limit; part 1 covers `packages/*`). This part covers `stores/*`, `agent-sdks/*`, `pubsub/*`, `mastracode/`, `workflows/inngest`, `voice/*`, `signals/github`, `server-adapters/nestjs`, and `observability/` — 83 files.

Same mechanical change and rationale as part 1: drop `randomUUID` from `node:crypto`/`crypto` imports and call the identical Web Crypto global, removing an unnecessary Node-only module dependency that breaks bundling for V8-isolate runtimes (Convex default runtime, Cloudflare Workers without `nodejs_compat`). Follows the `@mastra/convex` fix in #19498.

One non-trivial spot: the datasets storage domains in `stores/mongodb`, `stores/mysql`, and `stores/spanner` passed `randomUUID` as a function reference to `planDatasetItemBatch`; those now pass `() => crypto.randomUUID()`.

## Related issue(s)

Fixes #19501

## Type of change

- [x] Code refactoring
- [ ] Performance improvement
- [x] Test update

## Testing

- Zero leftover runtime `randomUUID`-from-crypto imports in the covered directories (repo-wide grep).
- `pnpm turbo build` for `@mastra/mongodb`, `@mastra/mysql`, `@mastra/spanner`, `@mastra/openai` and their dependency graphs: clean (this caught and verified the function-reference fix above).
- Deterministic test UUIDs unaffected: `@internal/test-utils/setup` stubs the global `crypto.randomUUID`.
- Changeset patch-bumps the 25 published packages touched here.

## AI disclosure

Implemented with Claude Code (Fable 5) as a reviewed codemod; human-reviewed.

## ELI5

We replaced Node-specific UUID generation with the browser-compatible Web Crypto version across the codebase. This helps the same code run in edge environments like Cloudflare Workers and Convex without extra Node compatibility settings.

## Summary

- Updated UUID generation across storage, SDKs, workflows, voice, pubsub, signals, adapters, Mastra Code, and observability to use `crypto.randomUUID()`.
- Updated MongoDB, MySQL, and Spanner dataset batching to receive UUID generators.
- Preserved deterministic UUID stubs in tests and other `node:crypto` APIs.
- Added validation to prevent covered directories from reintroducing `randomUUID` imports.
- Added patch changesets for 25 published packages.

### DIFF
--- a/.changeset/web-crypto-uuid-adapters.md
+++ b/.changeset/web-crypto-uuid-adapters.md
@@ -1,0 +1,29 @@
+---
+'@mastra/acp': patch
+'@mastra/claude': patch
+'@mastra/cloudflare': patch
+'@mastra/cloudflare-d1': patch
+'@mastra/code-sdk': patch
+'@mastra/couchbase': patch
+'@mastra/cursor': patch
+'@mastra/github-signals': patch
+'@mastra/inngest': patch
+'@mastra/libsql': patch
+'@mastra/mongodb': patch
+'@mastra/mssql': patch
+'@mastra/mysql': patch
+'@mastra/nestjs': patch
+'@mastra/observability': patch
+'@mastra/openai': patch
+'@mastra/pg': patch
+'@mastra/google-cloud-pubsub': patch
+'@mastra/redis-streams': patch
+'@mastra/spanner': patch
+'@mastra/upstash': patch
+'@mastra/vectorize': patch
+'@mastra/voice-aws-nova-sonic': patch
+'@mastra/voice-google-gemini-live': patch
+'mastracode': patch
+---
+
+Switched UUID generation to the Web Crypto global (crypto.randomUUID) instead of importing randomUUID from node:crypto. No behavior change on Node.js; removes an unnecessary Node-only module dependency so packages bundle more cleanly for edge and V8-isolate runtimes such as the Convex default runtime and Cloudflare Workers.
--- a/agent-sdks/acp/src/agent.ts
+++ b/agent-sdks/acp/src/agent.ts
@@ -1,4 +1,3 @@
-import { randomUUID } from 'node:crypto';
 import { ReadableStream } from 'node:stream/web';
 
 import type { ModelInfo, SessionUpdate } from '@agentclientprotocol/sdk';
@@ -111,7 +110,7 @@ export class AcpAgent<
       },
       toolResults: [],
       finishReason: 'stop',
-      runId: options?.runId ?? randomUUID(),
+      runId: options?.runId ?? crypto.randomUUID(),
     };
   }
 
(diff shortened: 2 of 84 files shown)


ediskandarov:
Closing: pivoting to package-scoped PRs per review feedback. Will re-submit per package after the @mastra/core slice.

PR #19537, opened 2026-07-16
feat(memory): capture scoped knowledge with Subconscious OM

TylerBarnes:
Adds experimental Subconscious configuration, deterministic capture, and semantic indexing. Observation cycles extract scoped knowledge nodes and items, code stamps provenance and timestamps, persists node content and item text, and drains semantic-outbox work through the configured embedder/vector infrastructure. Missing semantic dependencies fail explicitly rather than degrading silently.

Also includes the packed-package proof scaffold used by later stack segments.

Tested with focused capture/config/semantic-index suites, Memory typechecking/builds, and LibSQL-backed integration coverage.

Stack: 2 of 14 · Previous: #19536 · Next: #19538

## ELI5

This PR helps Memory save useful facts from conversations, keep them separated by organization, resource, and thread, and find them using meaning-based search. It also adds configuration, storage, recovery, and integration tests for this experimental feature.

## Summary

- Added experimental `Subconscious` configuration and public exports.
- Added deterministic knowledge capture with:
  - Scoped nodes and records.
  - Provenance and timestamps.
  - Capture ceilings.
  - Learned guidance.
  - Custom capture hooks.
- Added semantic indexing for knowledge documents:
  - Embedding and vector storage.
  - Scoped outbox draining.
  - Bounded batches and deduplication.
  - Stale-vector deletion.
  - Retry and release-on-failure handling.
- Updated Memory to:
  - Accept `experimental_subconscious`.
  - Register observation extractors without duplicates.
  - Validate knowledge and semantic-index dependencies.
  - Expose `getKnowledgeSemanticIndex()` and `drainKnowledgeSemanticIndex()`.
- Updated in-memory and LibSQL knowledge stores to:
  - Apply visibility at the record scope.
  - Serialize semantic-outbox work for successive document versions.
- Added focused configuration, capture, semantic-index, and LibSQL integration tests.
- Added a packed-package proof scaffold for later stack segments.

### DIFF
--- a/.changeset/little-eyes-beg.md
+++ b/.changeset/little-eyes-beg.md
@@ -1,5 +1,0 @@
----
-'@mastra/memory': patch
----
-
-Fixed Subconscious capture to discard disconnected entities that have no facts.
--- a/.changeset/little-eyes-beg.md
+++ b/.changeset/little-eyes-beg.md
@@ -1,0 +1,5 @@
+---
+'@mastra/memory': patch
+---
+
+Fixed Subconscious capture to discard disconnected entities that have no facts.
--- a/.changeset/major-fans-roll.md
+++ b/.changeset/major-fans-roll.md
@@ -1,0 +1,20 @@
+---
+'@mastra/memory': minor
+'@mastra/core': patch
+'@mastra/libsql': patch
+---
+
+Added Subconscious capture configuration for Observational Memory. Conversations can now extract scoped entities and facts into durable knowledge storage and reconcile them into a semantic vector index.
+
+```ts
+const memory = new Memory({
+  storage,
+  vector,
+  embedder,
+  options: {
+    observationalMemory: {
+      subconscious: new Subconscious(),
+    },
+  },
+});
+```
--- a/.changeset/tall-mangos-leave.md
+++ b/.changeset/tall-mangos-leave.md
@@ -1,0 +1,5 @@
+---
+'@mastra/memory': patch
+---
+
+Fixed Observational Memory nested agents so their temporary thread context cannot leak into later parent-agent processors.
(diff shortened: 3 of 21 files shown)


TylerBarnes:
The capture path is directionally sound, but I don’t think this stack PR is independently ready yet. I’d like these runtime contracts proven before merge:

1. Add an integration test covering the public `Agent → Memory/OM → capture → LibSQL knowledge → semantic vector` flow, including sibling-thread isolation.
2. Add concurrent capture coverage for the same normalized entity and define/test recovery when entity persistence succeeds but a later fact write fails.
3. Add cross-scope semantic-outbox tests: rescope a document between visibility scopes, then overlap scoped and broader drains. The test should prove pending work cannot remain indefinitely stale merely because only a broader worker can claim the blocking operation.
4. Align the public configuration with executable behavior. `remind`, `curate`, `learn`, and related settings currently resolve successfully but do nothing in this PR. Either reject/defer unsupported configuration or add tests documenting an intentional staged contract.
5. Add user-facing docs for the newly exported `Subconscious` configuration and clearly identify which capabilities are available in this release.

The existing unit and file-backed integration coverage is useful, especially for scope ceilings, provenance, mentions, idempotent vector updates, and crash recovery. The missing cases above are the boundaries most likely to produce silent partial knowledge, cross-scope staleness, or misleading public behavior.

CodingButter (approved):
Walked capture.ts, semantic-index.ts, and the Subconscious config surface. Extractor failure isolation verified against extractor.ts/extracted-values.ts - a scope-context throw degrades to a per-extractor failure, never breaks the main conversation. Outbox claim/complete/release protocol and drain dedup are clean. Approving.

PR #19538, opened 2026-07-16
feat(memory): surface scoped Subconscious activity

TylerBarnes:
Publishes bounded, scope-filtered Subconscious knowledge activity on the state-signal lane. Activity snapshots use unified node/item records, stable cache keys, visibility checks, and bounded previews so hidden scopes and storage IDs do not leak.

Tested with focused activity and capture suites plus Memory typechecking/builds.

Stack: 3 of 14 · Previous: #19537 · Next: #19539

## ELI5

This PR adds a safe way to show recent knowledge activity to the active agent and provide knowledge browsing tools in Factory and TUI interfaces. It limits data to visible scopes, removes duplicates, and keeps previews bounded.

## Changes

- Added bounded Subconscious activity snapshots with scope filtering, visibility checks, stable cache keys, recent-update limits, hot-node summaries, bounded errors, and duplicate suppression.
- Propagated `sendStateSignal` through observational-memory capture, observation strategies, turns, and extractor hooks.
- Added capture-time node scopes, pinning, metadata, capture reasons, and `maxScope` enforcement.
- Added curation, learning, reminder, pinned-state, and scoped knowledge read/write tools.
- Added knowledge storage support for LibSQL, PostgreSQL, MySQL, and MongoDB, including metadata, pagination, activity, optimistic concurrency, and semantic outbox operations.
- Added Factory knowledge APIs with organization, project, thread, and scope isolation.
- Added SDK `KnowledgeInspector` support with scoped handles, cursors, pagination, visibility checks, and bounded content.
- Added feature-gated Factory, TUI, and React knowledge browsers with graph layouts, node details, activity views, navigation, and polling controls.
- Added focused unit, integration, storage, Factory, SDK, UI, and TUI test coverage.

## Testing

- Added focused Subconscious activity, capture, curation, pinning, learning, reminder, and guardrail tests.
- Added LibSQL integration coverage for observation, activity publication, scope filtering, cache deduplication, pinning, curation, and knowledge-tool visibility.
- Added storage contract and adapter tests for PostgreSQL, MySQL, and MongoDB.
- Added Factory, SDK, React UI, and TUI tests and end-to-end scenarios.

PR #19539, opened 2026-07-16
feat(memory): add scope-bound knowledge retrieval tools

TylerBarnes:
Adds scope-bound knowledge read tools over unified nodes and items. Tools support opaque cursor pagination, exact reads, relationship traversal, and deterministic hybrid lexical/semantic retrieval while failing closed when organization scope or semantic infrastructure is unavailable.

Tested with cursor, semantic retrieval, visibility, and unavailable-index coverage plus Memory/Core/LibSQL builds.

Stack: 4 of 14 · Previous: #19538 · Next: #19540


TylerBarnes:
The scope revalidation, private-parent masking, deterministic cursor contract, and shared InMemory/LibSQL coverage are strong. I don’t think this is ready to merge yet because several runtime contracts remain unclear or unproven:

1. **Define `readOnly` behavior.** `readOnly` skips the OM mutation lifecycle and suppresses the working-memory write tool, but still registers `knowledge_search`, which claims and completes semantic outbox entries. Please either prevent that mutation, suppress the relevant tool, or explicitly define and test that maintenance writes are permitted in read-only mode.

2. **Add real Agent integration coverage.** Current tests call `memory.listTools()` and tool executors directly. Please prove an Agent receives these default-on tools through normal Memory wiring and can execute one with runtime organization/resource/thread context while preserving sibling isolation.

3. **Document the public behavior.** Subconscious users now receive three model-visible tools by default. Please document `tools: false`, required scope IDs, vector/embedder requirements, default-on prompt impact, and semantic indexing/failure behavior.

4. **Cover hydration races.** Add a regression test where a semantic candidate is deleted or rescoped after vector retrieval but before hydration, proving final storage revalidation prevents stale or cross-scope output.

The underlying retrieval design looks coherent, but these contracts should be settled before making the tools default-on.

packages/memory/src/processors/observational-memory/subconscious/knowledge-tools.ts
```diff
@@ -0,0 +1,283 @@
...
+      additionalProperties: false,
+    } satisfies JSONSchema7,
+    execute: async (input, context) => {
+      const { query, limit: requestedLimit } = input as { query: string; limit?: number };
+      const scope = resolveScope(context as KnowledgeToolContext);
+      const limit = normalizeLimit(requestedLimit);
+      const store = await getKnowledgeStore(memory);
+      const semanticCandidates = await memory
+        .getKnowledgeSemanticIndex()
+        .then(index => index.search(query, scope, limit * 2));
```
CodingButter, line 170: Design question rather than a defect: if the semantic search throws here (broken embedder, StaleKnowledgeSemanticIndexError from a backed-up outbox), the whole knowledge_search call fails even though the lexical results on the next line could still serve the request. In-repo precedent leans toward not offering a mode you cannot serve rather than runtime fallback, so if fail-hard is intended here it is defensible - but worth stating on purpose. An alternative is catching the semantic failure and returning lexical-only results with a degraded flag.

CodingButter (approved):
Read tools reviewed in full. Three-layer scope enforcement (vector filter per ancestor key, metadata check, store re-verification) is solid defense in depth; RRF hybrid merge is correct. One design question left inline about semantic-failure behavior in knowledge_search - not blocking. Approving.

packages/memory/src/processors/observational-memory/subconscious/knowledge-tools.ts
TylerBarnes: Fail-hard is intentional. The signed v1 contract requires hybrid retrieval and explicitly forbids silently degrading semantic requests to lexical-only behavior; a stale or unavailable semantic index must remain actionable.

PR #19540, opened 2026-07-16
feat(memory): add targeted Subconscious reminders

TylerBarnes:
Adds targeted Subconscious reminders on the reactive signal lane. Reminder selection reads visible knowledge items, applies freshness and scope rules, avoids repeated delivery, and keeps reminder generation isolated from ordinary observation state.

Tested with focused reminder suites and LibSQL-backed integration coverage.

Stack: 5 of 14 · Previous: #19539 · Next: #19541


TylerBarnes:
The hook-mode extractor fits the OM architecture, and the resource-scoped thread targeting is a good direction. I don’t think the reminder runtime contract is safe enough to merge yet:

1. **Require verified provenance.** If the model does not cite a valid retrieved record, please suppress the reminder rather than assigning the first candidate IDs. The current fallback can make hallucinated output look grounded. Add coverage for invalid, missing, and tool-only citations.
2. **Prevent echo and feedback loops.** `capture` runs immediately before `remind`, so current-cycle same-thread facts can be returned as reminders. A persisted `remembered` signal can also re-enter observation and trigger another reminder. Please exclude recent/current-cycle self-knowledge and Subconscious-origin signals, with regression tests.
3. **Define model/runtime requirements.** Reminder execution currently depends on `mainAgent`; document and test the resolution order for per-agent config, OM model, and main-agent fallback, plus behavior when no main agent exists.
4. **Prove real lifecycle behavior.** Current tests invoke `omEngine.observe()` directly, and the resource test mocks `Agent.sendSignal()`. Please add Agent-driven coverage for active delivery, idle persistence and later consumption, async-buffer/finalization, cancellation, and exactly-once hook execution.
5. **Document the public behavior.** This activates the default `remind` agent and exports new configuration. Please document when it runs, required organization/resource/thread context, scope behavior, disabling/configuring it, model/tool dependencies, signal delivery semantics, and non-fatal failure behavior.

The file-backed LibSQL coverage is useful, but it currently proves the orchestration helpers rather than the complete Agent → OM → reminder → signal runtime path.

TylerBarnes:
Fixed the concrete provenance fallback in `ed33f8771b0`: reminder output is now suppressed unless it cites an ID from the exact scoped candidate set. Missing and invented citations have regression coverage, and the focused reminder suite passes 6/6.

The broader Agent-driven lifecycle matrix, current-cycle echo exclusion, model-without-main-agent contract, and cancellation/exactly-once semantics are design/runtime expansions rather than mechanical review fixes; I have kept those as explicit blockers for the final review gate rather than claiming this comment fully resolves them.

packages/memory/src/processors/observational-memory/subconscious/remind.ts
```diff
@@ -0,0 +1,150 @@
...
+            },
+          );
+          const reminder = result.text.trim();
+          if (!reminder || /^<no-reminder\s*\/>$/i.test(reminder)) {
+            return;
+          }
+
+          const candidateIds = [...new Set(sources.flatMap(source => [source.id, source.recordId]))];
+          const sourceIds = candidateIds.filter(id => reminder.includes(id)).slice(0, 5);
+          if (sourceIds.length === 0) {
```
CodingButter, line 160: The grounding filter is the right call, but when the model writes a good reminder in prose without echoing any candidate id verbatim, the reminder is suppressed with zero trace. A debug log on this branch (reminder text + candidate ids) would make selectivity tunable; today a mis-grounded model is indistinguishable from nothing worth reminding about.

CodingButter (approved):
Remind path reviewed. Grounded-citation filtering, the no-reminder sentinel, and the error-rethrow-after-activity-publish pattern all verified. Left an inline note suggesting a debug log on silent grounding suppression to aid tuning - not blocking. E2E reds at tip are outage-window install failures, rerun kicked. Approving.

packages/memory/src/processors/observational-memory/subconscious/remind.ts
```diff
@@ -0,0 +1,150 @@
...
+  'there',
+  'they',
+  'this',
+  'user',
+  'what',
+  'when',
+  'where',
+  'which',
+  'with',
+]);
```
TylerBarnes: what's this about? seems odd

packages/memory/src/processors/observational-memory/subconscious/remind.ts
TylerBarnes: The reminder is deliberately suppressed unless the model cites one of the current candidate IDs. This prevents plausible prose from surfacing knowledge that was not actually retrieved. We are keeping quiet no-result behavior; model-output logging would add an observability/privacy contract beyond this segment.

packages/memory/src/processors/observational-memory/subconscious/remind.ts
TylerBarnes: This is the grounding safeguard: the reminder model must include a retrieved candidate ID, and the runtime strips the ID before delivery. If it produces uncited prose, we suppress it rather than risk an ungrounded reminder. The bot was suggesting debug logging for suppressed output; I am not adding that because it would log model-generated/user-derived content and expand the observability contract.

PR #19541, opened 2026-07-16
feat(memory): curate durable Subconscious knowledge

TylerBarnes:
Adds durable Subconscious curation over unified knowledge nodes and items. The curator receives bounded source-thread worklists, writes node content and items through scope-aware tools, uses CAS for mutable node content, soft-removes items, and advances persistent item cursors only after successful work.

Tested with curator/write-tool suites, cursor persistence coverage, and Memory/Core/LibSQL builds.

Stack: 6 of 14 · Previous: #19540 · Next: #19542


TylerBarnes:
The storage-side foundations here are strong: curator tools stamp trusted provenance, enforce visibility and ceilings, use CAS where appropriate, expose soft deletion without restoration, and delegate semantic/activity consistency to the knowledge domain.

I don’t think the orchestration contract is ready yet:

1. **Cursor advancement does not prove durable prefix completion.** Any ID in the delivered worklist is accepted, so the model can acknowledge a later item without proving earlier items or intended mutations succeeded. Tool mutations commit independently; if generation later fails or emits an invalid marker, those writes remain while the cursor stays behind. Please define and test partial-success/retry semantics, including a successful mutation followed by failure, out-of-order acknowledgement, and retry idempotency.

2. **Scope authorization needs stronger runtime proof.** Tools reload and visibility-check records before issuing scope-unaware mutations, leaving potential concurrent rescope/merge authorization races. Newly synthesized node content may also be promoted to resource/org scope when configuration permits, without an evidence ceiling equivalent to `KnowledgeItem.maxScope`. Please either make authorization atomic at storage mutation time or add executable concurrent-rescope coverage, plus tests preventing thread-specific evidence from being laundered into broader durable knowledge.

3. **The lifecycle integration test can hide broken wiring.** It manually calls `maybeReflect()` when `observe()` does not trigger curation. We need a real Agent/OM reflection test where the normal production path reliably invokes curation, advances the cursor, and reports failures—without fallback. Please cover synchronous threshold reflection, buffered activation, cancellation/callback failure, and concurrent curator runs on one source thread.

4. **Model resolution is surprising.** A configured curator model still requires `context.mainAgent` because resolution always routes through `mainAgent.getModel()`. Please establish and test the intended fallback/override contract, including direct/manual reflection contexts where no main agent exists.

5. **This changes default public behavior.** `new Subconscious()` now enables a mutating `curate` phase by default. Please document when it runs, what it may write/delete/merge/rescope, model requirements, scope behavior, retry/error semantics, and how to disable or customize it.

The focused tool tests and LibSQL mechanics are valuable, but they currently prove individual safeguards more strongly than the end-to-end curation contract.

packages/memory/src/processors/observational-memory/subconscious/knowledge-write-tools.ts
```diff
@@ -0,0 +1,230 @@
...
+        requireVisible(node.scope, options, 'Knowledge node');
+        return store.updateNode({
+          id: node.id,
+          version: value.expectedVersion,
+          name: value.name,
+          kind: value.kind,
+        });
+      },
+    }),
+    knowledge_merge_nodes: createTool({
```
CodingButter, line 133: knowledge_merge_nodes is the only write tool that does not reject already-merged nodes: add_item and update_node both guard mergedInto, this one checks existence and visibility only. Because the curation batch replays on a failed ack, merge A into B followed by a replayed or confused merge B into A creates a cycle, and resolveTerminalNode then throws 'Knowledge merge cycle detected' on every subsequent resolveNode for that name, permanently. A one-line guard (if source.mergedInto or target.mergedInto, throw) closes it.

packages/memory/src/processors/observational-memory/subconscious/knowledge-write-tools.ts
```diff
@@ -0,0 +1,230 @@
...
+  expandKnowledgeScope,
+  isKnowledgeScopeVisible,
+  knowledgeScopeKey,
+} from '@mastra/core/storage';
+import type { ToolAction } from '@mastra/core/tools';
+import { createTool } from '@mastra/core/tools';
+import type { JSONSchema7 } from 'json-schema';
+
+const CURATOR_IDENTITY = 'subconscious:curate';
+const MAX_GUIDANCE_LENGTH = 8_000;
```
CodingButter: Cap mismatch: MAX_GUIDANCE_LENGTH here is 8000, but capture.ts slices the same capture-guidance node at MAX_CAPTURE_GUIDANCE_LENGTH 4000 - a curator can write guidance the capture agent will silently truncate in half. Worth aligning or sharing the constant.

CodingButter (approved):
Curator write tools + cursor protocol reviewed. CAS on update/merge, soft-delete-only removal, and the visibility-then-ceiling assert order on rescope are all correct. Two inline notes (merge_nodes missing a mergedInto guard, 4k/8k guidance cap mismatch) - worth fixing but neither blocks the rung. Approving.

packages/memory/src/processors/observational-memory/subconscious/knowledge-write-tools.ts
TylerBarnes: The storage boundary now rejects this replay cycle after resolving the target alias (82758f4fbb2), with InMemory and LibSQL coverage. The write tool can rely on that canonical invariant instead of duplicating alias resolution.

packages/memory/src/processors/observational-memory/subconscious/knowledge-write-tools.ts
TylerBarnes: Fixed in 2c0bfdb57de: curator guidance now enforces the same 4,000-character limit consumed by capture. Focused knowledge write-tool tests and the Core build pass.

PR #19542, opened 2026-07-16
feat(memory): learn reusable skills from Subconscious reflection

TylerBarnes:
Adds the durable Subconscious learner. Reflection-derived `KnowledgeItem` evidence feeds a bounded learner agent that can create or update reusable skills without recursively enabling Subconscious work.

Tested with learner skill-evidence, cursor, and cancellation coverage plus Memory typechecking/builds.

Stack: 7 of 14 · Previous: #19541 · Next: #19543


TylerBarnes:
This is a solid foundation for durable skill evidence, but I don’t think the runtime contract is ready to merge yet.

The focused tests cover the learner internals well, but both unit and LibSQL integration tests invoke the learner handler directly. Please add real Agent → Memory/OM → reflection → learner coverage for synchronous and buffered reflection. That should prove the pre-reflection observations are correct, recursion is prevented, cancellation is honored, activity failures are surfaced, and cursor advancement occurs through the production lifecycle.

Retry and concurrency semantics also need to be explicit and tested. Skill-node creation, evidence writes, and cursor advancement are separate mutations. Please cover:

- A successful evidence write followed by failure before cursor advancement
- Replaying the identical source items after that failure
- A retry that proposes conflicting procedure content or scope
- Concurrent learners creating/updating the same skill
- Out-of-order completion markers and partial worklist acknowledgement

Deterministic evidence IDs help, but accepting an existing ID without verifying its parent, content, scope, and ceiling makes conflicting retries first-writer-wins rather than demonstrably safe.

I’m also concerned about scope and authority. A broadly visible skill may summarize narrower evidence, and the learner receives general knowledge write tools that can bypass `knowledge_record_skill`’s two-source, deterministic-ID, and one-skill safeguards. Please either restrict mutation authority to the specialized path or prove with mixed-scope tests that broader retrieval cannot expose private source-derived content.

Finally, an explicitly configured learner still depends on `context.mainAgent` for model resolution. Please define and test model resolution when no main agent is present. Since `learn` changes from an inert default to default-on model-driven durable writes, public documentation should cover activation, disabling, model/context requirements, costs, write/scope behavior, and partial-failure/retry semantics.

The source-bound provenance, deterministic IDs, isolated learner memory, and per-handler failure isolation are good building blocks. The missing part is executable proof that they remain safe through the actual reflection lifecycle and under retries, concurrency, and mixed scopes.

packages/memory/src/processors/observational-memory/subconscious/learn.ts
```diff
@@ -0,0 +1,252 @@
...
+      input.state.recordedName = normalizedName;
+      const nodeScope = expandKnowledgeScope(input.scope, input.defaultScope);
+      let node = await input.store.resolveNode({ name: normalizedName, scope: input.scope });
+      if (node && node.kind !== 'skill') throw new Error(`Knowledge node is not a skill: ${normalizedName}`);
+      node ??= await input.store.createNode({ name: normalizedName, kind: 'skill', scope: nodeScope });
+      const evidence = [];
+      for (const sourceId of sourceIds) {
+        const id = evidenceItemId(sourceId, normalizedName);
+        const existing = await input.store.getItem({ id });
+        if (existing) {
```
CodingButter, line 107: The deterministic evidence id makes replays idempotent (nice), but it also means a refined procedure for the same skill/source pair hits this existing-item check and continues - the improved text is never written and the tool still reports success, while the tool description advertises 'or evidence update'. Either fold a content hash into the id or route procedure improvements to the node content; today update is a silent no-op the model cannot detect.

CodingButter (approved):
Learn rung reviewed. Deterministic evidence-id derivation with race-tolerant idempotent appends is well designed; two-evidence minimum and one-skill-per-reflection gates verified. Inline note on record_skill silently dropping procedure updates for an existing (skill, source) pair. CI reds at tip are outage-window infra, rerun kicked. Approving.

packages/memory/src/processors/observational-memory/subconscious/learn.ts
TylerBarnes: Procedure refinement semantics are a design-sized learner follow-up. v1 intentionally keys evidence deterministically for crash/replay idempotency; changing IDs or mutating skill content requires a signed update/conflict policy rather than silently changing this segment.

PR #19543, opened 2026-07-16
feat(memory): harden Subconscious execution boundaries

TylerBarnes:
Hardens Subconscious execution boundaries: recursion guards, cancellation propagation, typed background-error events, origin tracking, temporary-agent isolation, and configuration checks. Stable-looking `subconscious` configuration remains rejected; the capability is exposed through `experimental_subconscious`.

Tested with focused hardening suites, Memory typechecking/builds, and packed-package proof gates.

Stack: 8 of 14 · Previous: #19542 · Next: #19544


TylerBarnes:
Thanks for hardening these execution boundaries. The centralized filtering point, abort propagation, and typed background errors are the right areas to address, but I don’t think the isolated PR meets its runtime guarantees yet.

**Persisted Subconscious signals bypass the new filter.** `stripSubconsciousSignals()` recognizes transient `data-signal` parts, while `MessageList.addSignal()` persists signals as `role: "signal"` rows with origin under `content.metadata.signal`. `ObservationStrategy.run()` applies the filter directly to prepared messages without first converting those rows to `data-signal`. I reproduced `createSignal() → MessageList.addSignal() → get.all.db() → stripSubconsciousSignals()` and the Subconscious reminder remained. Activity and error state signals use the same persistence path. Please support both durable and transient representations and add a persistence/reload or real next-observation-cycle regression test.

**Cancellation also needs post-await enforcement.** `Agent.generate()` can resolve with empty output after cancellation rather than throw, so catch-only abort checks are insufficient. Remind can treat cancellation as quiet success, while curate/learn may continue parsing completion markers or advancing durable cursors. Please check cancellation after awaited model/tool/storage work and before parsing results, emitting signals/errors/activity, or advancing cursors. Tests should cover in-flight cancellation and prove cursors remain unchanged and no error/state output is written after cancellation.

The current tests are primarily helper-level: fabricated `data-signal` parts, a pre-aborted handler that explicitly throws, and isolated error publication. Please add lifecycle coverage through real sync, buffered, and resource-scoped OM paths, including durable signal reload, writer failure/closure, and cancellation during active work.

There are also release/description mismatches:

- The PR body credits temporary-agent isolation and configuration validation, but those are earlier-stack changes, not changes in this isolated diff.
- The changeset releases `@mastra/core` and `@mastra/libsql` despite this diff changing only Memory files. Please justify those releases or remove them.
- The Documentation label is present, but this PR has no docs for the new typed error contract.

Once the persisted representation and post-await cancellation cases are covered end-to-end, this should provide the execution boundary the PR intends.

TylerBarnes:
Fixed the durable-signal bypass in `099ab25480d`. `stripSubconsciousSignals()` now recognizes persisted `role: "signal"` rows via `content.metadata.signal.metadata.origin` / structured source attributes, while retaining unrelated persisted signals. Regression coverage uses canonical `createSignal().toDBMessage()` representations for remembered and activity signals; focused guardrail tests pass 2/2 and Memory typecheck/lint/format pass.

The separate post-await cancellation matrix and full sync/buffered/resource lifecycle coverage remain design-sized runtime work and are still blockers rather than being treated as resolved by this mechanical fix.

CodingButter (approved):
This rung fixes three findings I flagged on earlier rungs: signal stripping via origin metadata, error surfacing through writer.custom data parts, and abort propagation into reflection handlers. Guardrails test covers the stripping contract. Clean and green. Approving.

PR #19544, opened 2026-07-16
feat(storage): add Subconscious knowledge adapter parity

TylerBarnes:
Implements the unified knowledge node/item storage contract for MongoDB, MySQL, and PostgreSQL after the LibSQL path is green. Adapter parity covers literal scope visibility, independent item visibility, node content search, item provenance, mentions, activity, curation cursors, CAS updates, and recoverable semantic-outbox writes.

Tested with the shared adapter contract and package builds for all four production adapters.

Stack: 9 of 14 · Previous: #19543 · Next: #19549


TylerBarnes:
Thanks for extending the shared knowledge contract across MongoDB, MySQL, and PostgreSQL. The shared conformance suite and adapter-specific CAS/outbox/schema tests are a strong baseline, but I found several blocking parity and recoverability failures:

1. **MongoDB standalone/custom-handler writes are not recoverable.** `withTransaction()` intentionally degrades to sequential writes when transactions are unavailable. I reproduced failures after the primary node/item insert leaving durable records without mentions, activity, or semantic-outbox entries. Retrying with the same ID returns the existing record without repairing derived state; without an explicit item ID, the generated ID is lost and retrying creates a duplicate. Either transaction support must be required for this domain or retries must reconcile every derived write. Please add failure-injection tests for each write boundary on non-transactional deployments.

2. **MySQL/PostgreSQL do not implement literal query semantics.** Both `namePrefix` and lexical search pass unescaped `%`/`_` through `LIKE`. I reproduced `%` matching unrelated names and `_` matching unrelated search content, while in-memory uses literal `startsWith`/`includes` and MongoDB escapes regex input. Please escape `LIKE` metacharacters and add shared conformance tests for `%`, `_`, and the escape character.

3. **Concurrent same-name node creation leaks raw database errors.** On fresh schemas, concurrent normalized-name creation produced raw MySQL `ER_DUP_ENTRY` and PostgreSQL `23505` failures, whereas MongoDB rereads and returns the canonical winner. Catch only the expected identity uniqueness conflict, reread the winner, and add this race to the shared suite.

4. **The Memory configuration restriction is unrelated and lacks release treatment.** This PR removes `model`/`maxSteps` from exported capture/custom-observation configuration and rejects them at runtime, but includes no `@mastra/memory` changeset or documentation. Please split this into a focused change, or add the appropriate Memory release, migration guidance, and explanation of the new Observer-model contract.

Before merge, I’d also like file-backed contention coverage for direct cursor advancement and backend-specific long-name/collation coverage. The shared tests currently miss the adapter-specific semantics that exposed these failures.

One non-blocking parity issue: MongoDB merge emits an `upsert` for a deleted mention-source item where the SQL adapters emit `delete`. The semantic-index loader currently prevents resurrection by converting the missing/deleted document into a delete, but MongoDB should still emit the correct operation and have regression coverage.

TylerBarnes:
Fixed the literal-query parity blocker in `dad79b71746`. LibSQL, MySQL, and PostgreSQL now escape `%`, `_`, and the escape character for both `namePrefix` and lexical search; the shared adapter contract covers all three metacharacters. The LibSQL contract passes 16/16, all three SQL adapter builds pass, and format/lint checks are clean.

The MongoDB non-transactional recovery model, concurrent same-name race normalization, post-write failure injection matrix, and configuration split/release treatment remain design-sized blockers; this commit only resolves item 2 and does not claim the full review is addressed.

stores/mongodb/src/storage/domains/knowledge/index.ts
```diff
@@ -0,0 +1,919 @@
...
+        .limit(limit - results.length)
+        .toArray();
+      for (const fact of facts) {
+        const parent = await this.#resolveTerminalNode(fact.parentNodeId);
+        const parentVisible = parent && isKnowledgeScopeVisible(parent.scope, scope);
+        results.push({
+          type: 'item',
+          id: fact.id,
+          recordId: fact.parentNodeId,
+          name: parentVisible ? parent.name : '(private entity)',
```
CodingButter: Tiny parity leftover from the records-to-nodes rename: every other adapter renders an invisible parent as '(private node)'; mongo says '(private entity)'.

CodingButter (approved):
Adapter parity reviewed across pg/mysql/mongodb against the libsql reference: LIKE escaping (ESCAPE clause + escapeLikePattern), per-adapter scope-prefix SQL, and mongo transactions all verified; contract suites run against each. One vocabulary nit inline (mongo's '(private entity)' vs '(private node)'). Approving.

stores/mongodb/src/storage/domains/knowledge/index.ts
TylerBarnes: Fixed during the KnowledgeRecord terminology migration; MongoDB now emits `(private node)`, matching the other adapters.

PR #19549, opened 2026-07-16
feat(mastracode): enable Subconscious memory for dogfooding

TylerBarnes:
Enables experimental Subconscious memory in Mastra Code when vector and knowledge storage are available. The SDK supplies project-scoped identity, configured observation/reflection models, semantic infrastructure, and deterministic request-context propagation while preserving non-Subconscious behavior for unsupported storage.

Tested with Mastra Code SDK memory tests and package builds.

Stack: 10 of 14 · Previous: #19544 · Next: #19557


TylerBarnes:
I don’t think this is ready to merge yet. The Mastra Code wiring is focused, but enabling Subconscious globally introduces lifecycle and isolation behavior that the current tests do not cover.

**1. Internal goal-judge and delegated-agent content can enter shared knowledge**

Goal judges inherit the parent `MastraMemory` and retain the parent resource ID. Their synthetic thread IDs isolate transcript history, but do not isolate resource-scoped Subconscious knowledge. Capture can therefore process internal prompts, judge decisions, approval state, and user follow-ups and persist them as project-visible knowledge. Delegated agents that inherit memory have a similar exposure.

Please define an explicit policy for internal actors and enforce it at the lifecycle boundary—for example, disable Subconscious capture/curation for `goalJudge` and other internal threads unless deliberately enabled. Add a real Agent → goal judge/subagent → OM/Subconscious regression proving internal deliberation cannot enter parent-visible knowledge.

**2. The failing E2E represents changed runtime behavior**

The goal-judge isolation E2E is not failing only because an expected count became stale. The existing observer fixture returns XML, while Subconscious capture expects structured knowledge output. This triggers capture retries and reminder calls; locally the scenario produced 21 model requests with 10 unmatched requests.

Please update the fixture for the intended topology, eliminate unintended calls, and make the test assert which actor and model initiated every OM/Subconscious request. The PR should not ship with this E2E failing.

**3. Capability gating does not match the documented/released contract**

The changeset and docs say activation requires vector, embedder, and knowledge storage, but `getDynamicMemory()` visibly gates Subconscious only on vector availability. Please explicitly verify all required capabilities before enabling it and test the unsupported-storage cases. Otherwise users can enter a partially configured lifecycle contrary to the stated contract.

**4. Current unit tests prove wiring, not behavior**

The added test mocks `Subconscious` and checks constructor arguments. It does not execute real Memory or prove capture, durable persistence, semantic indexing, unsupported-storage fallback, request-context isolation, or internal-actor behavior. This default-on integration needs lifecycle-level coverage using the real processor/storage path.

**5. Documentation is stale**

The docs describe `observationalMemory.subconscious`, while runtime uses `experimental_subconscious`, and they use the old entities/facts/pages terminology rather than nodes/items. The documented per-agent model/`maxSteps` behavior also needs reconciliation with the current capture/reminder configuration contract.

**6. Owner/resource identity assumptions are unstable in this stack position**

This PR derives organization identity from `session.ownerId`; later Factory changes introduce authoritative `factoryOrgId` and use `factoryProjectId` as `knowledgeResourceId`. Please either include the final identity contract here or rebase/restructure the stack so this PR does not temporarily persist knowledge under identities that subsequent PRs reinterpret.

Before merge, I’d like to see the internal-thread policy implemented, the E2E passing with explicit model-call assertions, capability gating aligned with the release contract, real lifecycle tests, corrected docs, and stable organization/resource identity semantics.

CodingButter (approved):
Dogfooding rung reviewed. Subconscious enabled only when a vector store is present, matching the documented requirement; org resolution from the controller session degrades to per-extractor failure when missing. Docs section verified accurate against actual code behavior. Approving.

docs/src/content/en/docs/memory/observational-memory.mdx
```diff
@@ -854,6 +854,76 @@ If you're using working memory to store conversation summaries or ongoing state
 
 In practical terms, OM replaces both working memory and message history, and has greater accuracy (and lower cost) than Semantic Recall.
 
+## Subconscious memory
```
TylerBarnes: we should leave docs out for now I think

TylerBarnes: maybe we can put them in a .md file in source code until we're ready to formally launch this feat?

mastracode/sdk/src/agents/memory.ts
```diff
@@ -112,6 +116,12 @@ export function getDynamicMemory(storage: MastraCompositeStore, vector?: MastraV
           enabled: true,
           temporalMarkers: true,
           retrieval: vector ? { vector: true } : true,
+          experimental_subconscious: vector
+            ? new Subconscious({
+                defaultScope: 'resource',
+                maxScope: 'resource',
+              })
+            : undefined,
```
TylerBarnes: there should be an env var opt in for this

docs/src/content/en/docs/memory/observational-memory.mdx
TylerBarnes: Removed the unreleased Subconscious section from the public OM guide in 473ac11711f. We will keep launch documentation private until the feature is ready.

mastracode/sdk/src/agents/memory.ts
TylerBarnes: Added an explicit Mastra Code opt-in in ae01a4c8b8f. Subconscious now stays disabled unless MASTRACODE_EXPERIMENTAL_SUBCONSCIOUS=1 is set; regular OM remains enabled.

PR #19552, opened 2026-07-16
fix(mastracode-web): polish Factory board entry

TylerBarnes:
## Summary

- send signed-in users with persisted Factory work to the board while preserving explicit safe return paths and keeping empty projects on the new-chat screen
- add an accessible Intake action that opens the canonical GitHub repository's new-issue page in a safe new tab
- position the board at its first populated lane once without overriding manual scrolling or later query refetches

## Test plan

- `cd mastracode/web && pnpm exec vitest run --config e2e/web-ui/vitest.config.ts src/web/ui/domains/auth/hooks/__tests__/useWebAuth.msw.test.tsx src/web/ui/domains/factory/__tests__/FactoryPages.msw.test.tsx src/web/ui/__tests__/routing.msw.test.tsx --reporter=dot --bail 1`
- `cd mastracode/web && pnpm check`
- `bash .mastracode/plans/factory-feedback-series.proof/demo.sh board-entry --app-root "$(git rev-parse --show-toplevel)" --expect green`
- live browser check against `TylerBarnes/mastracode-web-signals-proof-20260713`

## ELI5
If you’re signed in and you already have Factory work saved, the app takes you back to your Factory board (instead of starting a fresh chat). It also adds a safe “Create GitHub issue” link in the Intake lane and makes the board auto-scroll to the first real content only once—without fighting your own scrolling.

## Changes
- Updated root (`/`) landing: for authenticated GitHub projects with persisted Factory work, route to `/factory/board`; for empty projects, route to `/new`.
- If persisted Factory work fails to load, keep the existing root landing decision (don’t treat it like “empty” and redirect to `/new`).
- Improved auth redirects:
  - When sending users to `/signin`, preserve a safe `returnTo`.
  - On `/signin`, route authenticated users back via `returnTo` (or fall back to `/new` if the destination is unsafe).
  - Added an auth-loading skeleton to avoid flashes/bounce behavior during `/auth/me` resolution.
- Added Intake “Create GitHub issue” action on the Factory board:
  - Generates the canonical `https://github.com/<owner>/<repo>/issues/new` URL only for valid canonical `owner/repo` inputs.
  - Hides the link when GitHub/Intake isn’t configured or when the repo name is invalid.
  - Opens the link in a safe new tab.
- Board auto-positioning:
  - Scrolls the board to its first populated lane after initial load.
  - Tracks user manual horizontal scrolling so later automatic positioning won’t override it.
  - Prevents the initial programmatic scroll from triggering subsequent scroll handling/refetch-driven jumps.
- Expanded routing and Factory board regression tests to cover the above behaviors, including delayed work-items readiness and persisted-work failure cases.

### DIFF
--- a/mastracode/web/src/web/ui/domains/factory/BoardPage.tsx
+++ b/mastracode/web/src/web/ui/domains/factory/BoardPage.tsx
@@ -1,10 +1,10 @@
-import { Button } from '@mastra/playground-ui/components/Button';
+import { Button, buttonVariants } from '@mastra/playground-ui/components/Button';
 import { DropdownMenu } from '@mastra/playground-ui/components/DropdownMenu';
 import { Notice } from '@mastra/playground-ui/components/Notice';
 import { Txt } from '@mastra/playground-ui/components/Txt';
-import { CircleDot, EllipsisVertical, GitPullRequest, MessageSquare } from 'lucide-react';
+import { CircleDot, EllipsisVertical, GitPullRequest, MessageSquare, Plus } from 'lucide-react';
 import type { ComponentType, DragEvent } from 'react';
-import { useMemo, useState } from 'react';
+import { useEffect, useMemo, useRef, useState } from 'react';
 import { useNavigate } from 'react-router';
 
 import { useApiConfig } from '../../../../shared/api/config';
@@ -42,6 +42,11 @@ const NEEDS_APPROVAL_LABEL = 'needs-approval';
 
 function hasLabel(labels: readonly string[], label: string): boolean {
   return labels.some(item => item.toLowerCase() === label);
+}
+
+function githubNewIssueUrl(repoFullName: string): string | undefined {
+  if (!/^[A-Za-z0-9_.-]+\/[A-Za-z0-9_.-]+$/.test(repoFullName)) return undefined;
+  return `https://github.com/${repoFullName}/issues/new`;
 }
 
 function metadataLabels(metadata: Record<string, unknown>): string[] {
@@ -381,6 +386,7 @@ function Board({ project }: { project: Project & { githubProjectId: string } }) 
     'github-prs' as const,
     ...(linearReady ? (['linear'] as const) : []),
   ];
+  const newIssueUrl = config && githubIntakeActive ? githubNewIssueUrl(project.name) : undefined;
   const [intakeSource, setIntakeSource] = useState<IntakeSource>('github');
   const showIntakeSourceSwitch = availableIntakeSources.length > 1;
   const activeIntakeSource: IntakeSource | null = availableIntakeSources.includes(intakeSource)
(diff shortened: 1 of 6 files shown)


mastracode/web/src/web/ui/domains/factory/BoardPage.tsx
```diff
@@ -438,6 +452,31 @@ function Board({ project }: { project: Project & { githubProjectId: string } })
...
+    configQuery.isPending ||
+    linearStatusQuery.isPending ||
+    triageIssues.isPending ||
+    (activeIntakeSource === 'github' && issues.isPending) ||
+    (activeIntakeSource === 'github-prs' && pulls.isPending) ||
+    (activeIntakeSource === 'linear' && linearIssues.isPending);
+
+  useEffect(() => {
+    if (boardDataPending || positionedProjectRef.current === project.id) return;
+    positionedProjectRef.current = project.id;
```
mfrachet: what's the diff between positionedProjectRef and userPositionedProjectRef?

mastracode/web/src/web/ui/domains/factory/BoardPage.tsx
```diff
@@ -475,13 +514,44 @@ function Board({ project }: { project: Project & { githubProjectId: string } })
...
+        className="flex min-h-0 flex-1 gap-3 overflow-x-auto pb-2"
+        aria-label="Board columns"
+        onPointerDown={() => {
+          userPositionedProjectRef.current = project.id;
+        }}
+        onWheel={() => {
+          userPositionedProjectRef.current = project.id;
+        }}
+        onScroll={() => {
+          if (positionedProjectRef.current !== project.id) userPositionedProjectRef.current = project.id;
```
mfrachet: NIT: this condition is probably not needed

mfrachet (approved):
small nits, but lgtm

mastracode/web/src/web/ui/domains/factory/BoardPage.tsx
TylerBarnes: Renamed positionedProjectRef to autoPositionedProjectRef. It records our one-time initial positioning; userPositionedProjectRef records pointer/wheel/manual scrolling so a user's chosen lane is not overridden later.

mastracode/web/src/web/ui/domains/factory/BoardPage.tsx
TylerBarnes: The condition filters the onScroll event emitted by our own scrollTo call; without it, automatic positioning would be recorded as user positioning. I renamed the ref and documented that distinction inline.

mastracode/web/src/web/ui/router.tsx
```diff
@@ -47,21 +45,37 @@ function AuthPendingSkeleton() {
...
   return <SignInPage />;
 }
 
+function RootLanding() {
+  const { activeProject } = useActiveProject();
+  const githubProjectId = activeProject?.source === 'github' ? activeProject.githubProjectId : undefined;
+  const workItems = useWorkItemsQuery(githubProjectId);
+
+  if (githubProjectId && workItems.isPending) return <AuthPendingSkeleton label="Loading Factory board" />;
+  return <Navigate to={githubProjectId && (workItems.data?.length ?? 0) > 0 ? '/factory/board' : '/new'} replace />;
```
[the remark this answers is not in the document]
TylerBarnes: Fixed in a69ba36a078. RootLanding now renders the work-items error and remains on `/`; it only chooses `/factory/board` versus `/new` after a successful query. Added a routing regression and reran routing + Factory MSW (64/64).

PR #19557, opened 2026-07-16
feat(mastracode): inspect subconscious knowledge

TylerBarnes:
Adds a scoped Mastra Code knowledge inspector and `/knowledge` TUI browser over unified knowledge nodes and items. It preserves opaque handles, bounded previews, content reads, relationship counts, directional navigation, activity pagination, graph-role grouping, and fail-closed scope checks.

Tested with SDK inspector tests, TUI knowledge-browser tests, the focused browser E2E scenario, and Mastra Code builds/typechecks.

Stack: 11 of 14 · Previous: #19549 · Next: #21384


TylerBarnes:
Thanks—this is a useful inspection surface, and the opaque handles, bounded reads, identity invalidation, and activity-pagination corrections are directionally strong. I found two blockers before merge:

1. **Structured activity snapshots leak identifiers across the visibility boundary.** `buildSubconsciousActivitySnapshot()` removes an inaccessible target’s name, but retains `recordId`, `targetId`, and `sourceThreadId`. The full snapshot is then persisted in the state signal’s `metadata.value`; deleting those fields in the TUI parser only prevents rendering and does not prevent storage or delivery to other consumers. Please redact inaccessible identifiers before constructing/persisting the snapshot and define when `sourceThreadId` is authorized. The privacy test currently asserts that hidden record IDs remain present, which appears contrary to the stated fail-closed behavior.

2. **The specialized TUI renderer does not accept the runtime snapshot schema.** Memory now emits `node`/`item` record types and `node` hot targets, while `parseSubconsciousActivitySnapshot()` and its fixtures still require `entity`/`fact`/`page`. Real snapshots therefore fail parsing and fall back to generic state rendering. Please migrate the parser/types/fixtures to the current node/item contract.

The current tests exercise Memory and TUI separately, which allowed both issues through. Please add a persisted end-to-end test covering:

- Memory activity creation
- state-signal persistence/reconstruction
- MastraCode parsing/rendering with the real snapshot type
- a target moved outside visible scope, proving its name and identifiers never reach the consumer

The inspector’s fail-closed claim also depends on each storage adapter correctly filtering list, relationship, and activity results. The SDK tests use only `InMemoryKnowledgeStorage`; please add at least one real-adapter integration path or narrow the guarantee accordingly.

Finally, the `@mastra/core` minor changeset says this PR adds structured state-signal values/deltas to AgentController messages, but the isolated diff appears to add only MastraCode’s DB-message projection; the underlying Core capability comes from earlier stack work. Please justify, relocate, or remove that changeset so release notes match this PR.

With those addressed, the remaining inspector, `/knowledge`, and pagination design looks reasonable.

CodingButter (approved):
Inspector + TUI browser reviewed. Activity snapshots now strip record identifiers with leak-assertion tests covering it - resolves my earlier rendering nit. Stale-handle guard and scope-bound reads in KnowledgeInspector verified; the five-adapter listActivity pagination-direction fix is correct for DESC order. Reds at tip are outage infra + a known duckdb hook-timeout flake. Approving.

.changeset/curvy-bats-know.md
```diff
@@ -0,0 +1,5 @@
+---
+'@mastra/code-sdk': minor
+---
+
+Added directional relationship previews and relevance sorting to the scoped knowledge inspector.
```
TylerBarnes: there should be an env var opt in for knowledge in mastracode (I commented this on the prev PR too)

TylerBarnes: Added the Mastra Code opt-in in ae01a4c8b8f and propagated it through this branch. Knowledge stays disabled unless MASTRACODE_EXPERIMENTAL_SUBCONSCIOUS=1 is set.

CodingButter:
Stealth-launch check: the `/knowledge` command is visible and functional even when `MASTRACODE_EXPERIMENTAL_SUBCONSCIOUS` is unset.

The runtime gate (`mastracode/sdk/src/agents/memory.ts:111`) only controls whether the Subconscious instance is constructed. The TUI surface does not consult it:

- Command + help entry are registered unconditionally (`mastracode/tui/src/tui/setup.ts:367`, `mastracode/tui/src/tui/components/help-overlay.ts:41`).
- `knowledge-command.ts:6` only checks that `ctx.knowledgeInspector` exists.
- `createKnowledgeInspector` (`mastracode/sdk/src/knowledge-inspector.ts:771`) gates only on `storage.getStore('knowledge')` being truthy, and the knowledge domain always exists: it is in `DOMAIN_KEYS` (`packages/core/src/storage/base.ts:107`) and LibSQL constructs `KnowledgeLibSQL` unconditionally (`stores/libsql/src/storage/index.ts:228`).

Net effect with the flag off: users see `/knowledge` in help and can open an empty read-only browser. No data leaks, but the feature's existence is revealed, which defeats the stealth intent of the opt-in gate. Suggest gating the command registration (or inspector creation) on the same env flag.

For the sibling leaks on the factory side (ungated `KnowledgeRoutes` registration and the `requiresAuth: false` graph endpoint), those live on #21498 and we are fixing them there.

PR #19562, opened 2026-07-16
fix trucination of listMessages

### LINKED ISSUE #19550: [BUG] @mastra/dynamodb listMessages returns only the first DynamoDB page (drops newest messages on large threads)
### Describe the Bug

`MemoryStorageDynamoDB.listMessages` in `@mastra/dynamodb` fetches a thread's messages with a bare ElectroDB `query.go()` — **without `{ pages: 'all' }`** — so it returns only the **first DynamoDB query page (~1 MB)**. For any thread whose stored messages exceed one page, `listMessages` silently returns a subset.

Because the `byThread` GSI is sorted by `createdAt` **ascending**, the page that comes back is the **oldest** messages, and the **newest turns are dropped**. The method then computes `total` and `hasMore` from that truncated page and returns `hasMore: false`, so callers receive no signal that anything was omitted.

Source (`main`, `@mastra/dynamodb@1.1.2` — currently the latest published version) — `stores/dynamodb/src/storage/domains/memory/index.ts`:

```ts
// listMessages (~L424-425)
const query = this.service.entities.message.query.byThread({ entity: 'message', threadId });
const results = await query.go();          // ← no { pages: 'all' } → first page only
// ...then map/filterByDateRange/sort/slice happen IN MEMORY on that first page,
//    and `total`/`hasMore` are derived from it (hasMore: false).
```

For contrast, `listThreads` in the same file correctly paginates:

```ts
// listThreads (~L673)
...byResource({ ... }).go({ pages: 'all' })
```

`_getIncludedMessages` (~L788-789) has the same bare `query.go()`.

**Impact.** `listMessages` (and everything built on it — `memory.query`/recall) returns incomplete history for any thread larger than ~1 MB, and always drops the *most recent* messages. Ordering by `createdAt` DESC does not help: the in-memory sort only reorders the already-truncated oldest page, so the newest messages are never fetched.

This is especially damaging with **Observational Memory**. OM recall fetches raw messages via `listMessages` and filters them by `createdAt > lastObservedAt` (the observation watermark). Once a thread grows past one page, every message on the returned (oldest) page is *older* than the watermark, so recall returns **zero** recent raw messages — the agent keeps the observation summary but loses all recent conversation, and the state persists because the thread only keeps growing.

This is related to #14613 (lastMessages not respected under OM) but has a **distinct, storage-layer root cause**: even a correct OM/recall layer cannot surface messages that `listMessages` never fetched.

### Steps To Reproduce

Minimal repro (also in the gist below):

1. Create a `Memory` (`@mastra/memory`) backed by a `@mastra/dynamodb` `DynamoDBStore` (DynamoDB Local is fine).
2. Save ~400 messages to a single thread with a few KB of content each, so the thread’s total stored size exceeds one ~1 MB DynamoDB query page (≈1.6 MB).
3. Read them back:
   ```ts
   const res = await store.stores.memory.listMessages({
     threadId,
     perPage: false,
     orderBy: { field: 'createdAt', direction: 'ASC' },
   });
   ```
4. Observe:
   - `res.messages.length` is far less than 400 (only the first ~1 MB page).
   - `res.total` equals that truncated count and `res.hasMore === false` (no signal of truncation).
   - The returned messages are the **oldest**; the newest are absent (also true with `direction: 'DESC'`).
5. Confirm the rows are actually stored (not a write bug): a direct
   `entities.message.query.byThread({ entity: 'message', threadId }).go({ pages: 'all' })`
   returns all 400.

https://gist.github.com/nish932603/03a712bb74985303591fe22c8eaf2d31

### Expected Behavior

`listMessages` should return the thread's messages consistent with the requested pagination — not a silently truncated first page. Either:

- fetch all pages via `.go({ pages: 'all' })` (matching `listThreads`), then apply the existing in-memory filter/sort/paginate; and/or
- push `filter.dateRange` / limits into the DynamoDB query (e.g. a `createdAt` range on the `byThread` sort key + a real `Limit`, ordered DESC when the caller wants the newest N) so only the needed messages are read;
- and compute `total`/`hasMore` from the full result set rather than the first page.

At minimum, `listMessages` must never drop the newest messages while reporting `hasMore: false`. The same fix should apply to `_getIncludedMessages`.

### Environment Information

```block
@mastra/dynamodb: 1.1.2   (latest published; defect also present on `main`)
@mastra/memory:   1.22.2
@mastra/core:     1.50.1
Storage:          DynamoDB (reproducible on DynamoDB Local)
Node.js:          v24.x
OS:               macOS (darwin)
LLM Provider:     N/A — storage-layer bug, provider-independent
```

### Verification

connect-abdullah:
## Description

`MemoryStorageDynamoDB.listMessages()` and `_getIncludedMessages()` queried
ElectroDB with a bare `query.go()`, which only returns DynamoDB's first
Query response page (~1MB). For any thread whose messages exceeded that
size, the method silently returned a partial result set:

- Because the `byThread` GSI is sorted by `createdAt` ascending, the
  truncated page was always the **oldest** messages — the newest turns
  were dropped entirely, with no error and no signal in the response.
- `total` and `hasMore` were computed from the already-truncated set,
  so `hasMore` reported `false` even when hundreds of messages were
  missing.
- This broke Observational Memory recall: once a thread grew past one
  page, every returned message was older than the recall watermark, so
  OM lost all recent conversation while still keeping the summary.

`listThreads` in the same file already handled this correctly via
`.go({ pages: 'all' })`; `listMessages` and `_getIncludedMessages` did not.

**Fix:**

- **`listMessages`**: rewrote the DynamoDB query path to push
  `dateRange` filtering, `resourceId` filtering, and ordering into the
  ElectroDB query itself (`.gte/.lte/.between` on `createdAt`,
  `.where()` on `resourceId`) instead of fetching everything and
  filtering in memory. `total` is computed via a lightweight `id`-only
  `pages: 'all'` read; the requested page is fetched via a bounded
  query so large threads don't require loading the whole thread into
  memory just to return one page.
- **Multi-thread queries** (`threadId` as an array): DynamoDB can't
  query multiple partitions in a single request, so this is handled as
  a dedicated branch — fetch + `pages: 'all'` per thread in parallel,
  then merge/sort/paginate in memory. Covered by a dedicated test.
- **`_getIncludedMessages`**: replaced per-include full-thread loads
  with direct bounded queries around the target message's `createdAt`
  (`lte`/`gt` + ordered, count-limited fetch for previous/next
  context), so context windows resolve correctly even when the target
  message lives beyond the first DynamoDB page.

Added `domains/memory/index.test.ts` against DynamoDB Local, covering:
a 400-message/~1.6MB thread returned in full via `perPage: false`;
DESC-ordered pagination returning the correct newest-first page;
`include` context resolution for a target beyond the first page;
`dateRange` filtering with correct `total` and page contents;
multi-thread `listMessages` returning messages from all threads; and a
bounded (non-`pages:'all'`) page request where `offset + perPage`
crosses a DynamoDB page boundary, to confirm the paginated path itself
doesn't silently truncate. All tests pass.

**Note for reviewers:** the `total` calculation does a full
`pages: 'all', attributes: ['id']` scan of the thread on every call —
`id`-only, but still scales with thread size. Fine for now; worth
revisiting with a maintained counter if this becomes a hot path for
very large threads.

## Related issue(s)

Fixes #19550

## Type of change

- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [ ] Test update

## Checklist

## ELI5

Sometimes DynamoDB would only return the *first chunk* of a long conversation, so older/newer messages (and the `total`/`hasMore` page info) were wrong. This PR makes DynamoDB fetch the correct ranges/pages and, when you ask for “included” nearby messages, it fetches the right surrounding context even deep in the thread.

## What changed

- **Fixes truncated pagination** in `MemoryStorageDynamoDB.listMessages()` by pushing **date-range**, **resource**, and **ordering** filters into DynamoDB queries instead of filtering/sorting in memory.
- **Computes correct `total`/`hasMore`**: totals are calculated using an **all-pages `id`-only query**, then only the requested page’s message bodies are fetched.
- **Correct multi-thread behavior**: when `threadId` is an array, it fetches each thread in **parallel**, merges results, then **sorts/paginates** the combined set.
- **Repairs included-context retrieval** in `_getIncludedMessages()`:
  - avoids loading entire thread histories
  - fetches **bounded previous/next windows** around the target message
  - deduplicates merged included results
  - works correctly beyond the first DynamoDB page
- **Adds/updates DynamoDB Local integration tests** for:
  - large threads spanning multiple DynamoDB pages
  - pagination boundaries (`offset`/`perPage`)
  - ascending/descending ordering
  - date-range filtering including **exclusive endpoints**
  - `include` context retrieval past the first DynamoDB page
  - multi-thread list+pagination correctness
- **Adds a patch changeset** for `@mastra/dynamodb` documenting the `listMessages()` pagination + included-context fix.


abhiaiyer91 (approved):

connect-abdullah:
@abhiaiyer91 implemented code rabbit's actionable comments.

PR #19563, opened 2026-07-16
feat(playground): reshape signals around theme flow

mfrachet:
CleanShot 2026-07-17 at 14 43 08@2x

https://github.com/user-attachments/assets/1586d36c-6421-476e-a283-3a915a92db8e

Reworks the Signals experience around a focused onboarding state that explains how traces become Outcome, Goal, Behavior, and Sentiment analysis. It removes the legacy details and entity-learning UI surface, then adds the Agent Learning theme-flow hooks and Sankey composition used by the playground.\n\nVerified with the focused playground and playground-ui Signals tests, both package typechecks, the playground-ui build, and git diff validation.

## ELI5
This PR changes the **Signals** experience so it first shows a simple walkthrough of how **traces become signal analysis**, and it only shows the Sankey diagram once there’s enough **theme-flow** data. It also rebuilds the Sankey chart so labels/percentages and link weights are computed reliably.

## Summary
- **Replaced the legacy Signals drill-down UI with a guided empty/onboarding state**
  - Added **`SignalsEmptyState`** (pipeline visuals: Traces → Mastra Engine → Signal analysis, signal definitions, “waiting for traces” aside, docs/internal links, optional `actionSlot`)
  - Refactored **`SignalsOverviewPage`** to render **only** the new empty/onboarding + theme-flow visualization flow
  - Removed older surfaces including **`SignalDetailsPage`**, signal crumbs, and entity-filtering/components

- **Introduced an Agent Learning “theme-flow” data path**
  - Added theme-learning client + hooks in the playground:
    - `useThemeEntities`, `useThemeSnapshots`, `useThemeFlow` (plus a config guard)
  - Added **`SankeySignals`** to fetch snapshots/flow and render:
    - loading state, error state, and “insufficient/no data” empty state
    - the Sankey diagram when at least two stages are available
  - Added **`themeFlowToSankeyData`** to convert theme-flow stages/links into Sankey `columns` and `records`

- **Updated Sankey chart behavior for correctness + clearer labeling**
  - **`SankeyChart`**
    - Added configurable **`margin`** prop
    - Computes per-node totals/percentages and renders node labels as `value (xx%)`
    - Renders column labels above nodes with consistent `textAnchor="middle"`
  - **Weighted Sankey graph generation**
    - `buildSankeyChartGraph(..., getRecordWeight?)` and `Sankey`/`SankeyChart` now support `getRecordWeight`
    - Skips invalid weights (non-finite/negative) to avoid broken/extra links
  - Updated/added unit tests to cover node percentage labels and margin-based node positioning

- **Simplified app routing/exports**
  - Updated the `/signals` route to render only the new overview/Sankey experience
  - Adjusted exports and removed re-export surfaces for the deleted details/learning modules
  - Simplified Signals tests to match the new static/conditional rendering behavior


damien-schneider:
- Can we remove the overflow hidden on the chart, or the container of that chart to not cut the labels please.
- And also the black border on the text inside the chart isn't necessary I think, it feels very cheap, we can either add a much smoother and larger dropshadow as a safe zone, or reducing a bit more the opacity of the chart connection (idk how you call that zone haha)
- It also feels very big being full width, we should probably add a max-w-5xl with mx-auto around it imo
image

packages/playground-ui/theme.css
```diff
@@ -395,6 +395,45 @@ html.light {
   --shadow-glow-accent2: 0 0 20px oklch(0.637 0.237 25.331 / 0.25);
   --shadow-focus-ring: 0 0 0 3px oklch(0.723 0.219 149.579 / 0.12);
 
+  --animate-signals-connector: signals-connector 2.8s ease-in-out infinite;
+  --animate-signals-engine-pulse: signals-engine-pulse 3.2s ease-in-out infinite;
+  --animate-signals-chip: signals-chip 3.6s ease-in-out infinite;
```
damien-schneider: Instead of having them in the theme.css file we should create one for the component and import it only in the component

packages/playground-ui/src/ee/signals/components/signals-empty-state.tsx
damien-schneider, line 1: Can you try using either https://github.com/anthropics/claude-code/blob/main/plugins/frontend-design/skills/frontend-design/SKILL.md or at least maybe just removing the "Cards" container on these

Image

damien-schneider (changes requested):

packages/playground-ui/src/ds/components/SankeyChart/sankey-chart-utils.ts
```diff
@@ -111,16 +112,17 @@ export function buildSankeyChartGraph(
...
-        existingLink.value += 1;
+        existingLink.value += weight;
         existingLink.records.push(record);
       } else {
         linksById.set(id, {
           id,
           source: source.index,
           target: target.index,
-          value: 1,
+          value: weight,
```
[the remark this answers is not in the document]
mfrachet: AI says: Fixed in 927ec678278. `buildSankeyChartGraph` now rejects non-finite and negative weights before creating nodes or links, with regression coverage for NaN, Infinity, and negative values. Valid finite non-negative weights keep the existing aggregation behavior.

Mastra Code

damien-schneider (approved):

PR #19565, opened 2026-07-16
feat(playground): center tool execution page and add read-more for tool descriptions

damien-schneider:
https://github.com/user-attachments/assets/3996faa1-0cac-43e5-b387-394999480973

Reworks the tool execution page (Studio > Tools / agent tools / MCP tools) which previously used a two-column layout with a narrow sidebar form and a full-bleed editable CodeMirror result.

The page is now a single centered column (max-w-3xl): tool info, pill tabs, then the input form. The result only renders after running the tool, read-only via the design-system `CodeEditor` (copy button included), and slides in as a right column on desktop — the form gets pushed left with a smooth max-width transition. On mobile the result stacks below the form.

Tool descriptions are now clamped to 2 lines with a "Read more" toggle. That comes from a new `@mastra/playground-ui` component, `ClampedText`: the toggle only appears when the clamp actually cuts content, detected by measuring the rendered element (re-checked on resize and after fonts load) rather than counting characters.

```tsx
<ClampedText variant="ui-sm" className="text-neutral3" lines={2}>
  {toolDescription}
</ClampedText>
```

`Txt` also accepts a `ref` now (React 19 ref-as-prop), which `ClampedText` uses for the measurement.

## ELI5

This makes the tool playground cleaner and easier to use on phones and computers. Long descriptions can be expanded, and tool results are shown in a read-only box with a copy button.

## What changed

- Reworked the tool execution page into a centered, responsive layout.
- Displayed results only after execution, beside the form on desktop and below it on mobile.
- Replaced the editable result pane with the design-system `CodeEditor`.
- Added `ClampedText` with measured “Read more”/“Show less” behavior.
- Added `useIsClamped`, including resize and font-loading checks.
- Updated `Txt` to support React 19 refs.
- Added Storybook stories and tests for clamping behavior.

### DIFF
--- a/.changeset/yummy-llamas-fry.md
+++ b/.changeset/yummy-llamas-fry.md
@@ -1,0 +1,5 @@
+---
+'@internal/playground': patch
+---
+
+Simplified the tool execution page: form and result now stack in a single centered column, the result appears only after running the tool, and the result view is read-only with a copy button.
--- a/packages/playground/src/domains/tools/components/ToolExecutor.tsx
+++ b/packages/playground/src/domains/tools/components/ToolExecutor.tsx
@@ -1,11 +1,8 @@
-import { jsonLanguage } from '@codemirror/lang-json';
 import type { MCPToolType } from '@mastra/core/mcp';
-import { useCodemirrorTheme } from '@mastra/playground-ui/components/CodeEditor';
-import { CopyButton } from '@mastra/playground-ui/components/CopyButton';
+import { CodeEditor } from '@mastra/playground-ui/components/CodeEditor';
 import { MainContentContent } from '@mastra/playground-ui/components/MainContent';
 import { Tabs, Tab, TabList } from '@mastra/playground-ui/components/Tabs';
 import { cn } from '@mastra/playground-ui/utils/cn';
-import CodeMirror from '@uiw/react-codemirror';
 import { useState } from 'react';
 import type { ZodType } from 'zod';
 import {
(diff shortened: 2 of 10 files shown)


.changeset/yummy-llamas-fry.md
```diff
@@ -0,0 +1,5 @@
+---
+'@internal/playground': patch
```
mfrachet, line 2: not needed

packages/playground/src/domains/tools/components/ToolExecutor.tsx
```diff
@@ -40,44 +37,48 @@ const ToolExecutorContent = ({
...
               onSubmit={data => {
                 handleExecuteTool(data, schemaValues);
               }}
-              className="h-auto pb-7"
+              className="h-auto"
             />
           </div>
           {requestContextSchema && (
-            <div className={cn('p-5 overflow-y-auto', selectedTab !== 'request-context' && 'hidden')}>
+            <div className={cn(selectedTab !== 'request-context' && 'hidden')}>
```
mfrachet, line 67: removing the oveflow will not break the scrolling behaviour when we have a big schema?

mfrachet (commented):
Few comments, otherwise, lgtm

packages/playground/src/domains/tools/components/ToolExecutor.tsx
damien-schneider: No because the scroll is handled by the window directly now that it is one page, we'll later maybe add separated scroll behaviour for the result and for the tool inputs / request context, but for now it is fine

.changeset/yummy-llamas-fry.md
damien-schneider: Not sure to get why? Don't we add anymore the changeset for the playground or should we do it an other way?

mfrachet: for internal package we can skip them

mfrachet (approved):

.changeset/yummy-llamas-fry.md
damien-schneider: Doesn't it help to track the release changes, helping the growth team?

PR #14642, opened 2026-03-24
feat(channels): agent-level chat channels via Chat SDK adapters

CalebBarnes:
## Summary

Enable agents to communicate over messaging platforms (Slack, Discord, Telegram) using the [Vercel Chat SDK](https://github.com/vercel/chat) adapter ecosystem.

```ts
const agent = new Agent({
  channels: {
    adapters: {
      discord: createDiscordAdapter(),
      slack: createSlackAdapter(),
      telegram: createTelegramAdapter(),
    },
  },
});
```

## Architecture

Each agent creates a single `AgentChannels` instance that wraps one Chat SDK `Chat` instance managing all adapters. Configuration uses a nested structure on the `Agent` constructor.

### What `AgentChannels` handles

- **Event routing** — `onDirectMessage`, `onNewMention`, `onSubscribedMessage` handlers feed into `agent.stream()`
- **Thread mapping** — Maps platform threads to Mastra threads via storage metadata (no schema migration)
- **Webhook routes** — Auto-generates `POST /api/agents/{agentId}/channels/{platform}/webhook` per adapter
- **Tool generation** — Creates `add_reaction` and `remove_reaction` tools from each adapter
- **Streaming responses** — Consumes `agent.stream()` fullStream, posting text on step boundaries with timing-aware formatting
- **Tool approval** — Tools with `requireApproval: true` render as interactive cards with Approve/Deny buttons
- **Multimodal support** — Forwards image/file attachments and model-generated files (e.g. Gemini image output)
- **Multi-user awareness** — Prefixes messages with `[Author (@mention)]` so the agent can distinguish speakers
- **Gateway support** — Starts persistent WebSocket listeners (e.g. Discord Gateway) with auto-reconnection
- **Thread context** — Fetches recent platform messages on first mention to provide conversation context
- **Concurrency** — Uses Chat SDK queue strategy for overlapping messages

### `ChatChannelProcessor`

Input processor injecting channel context into agent prompts:
- Stable context (platform, DM status, bot identity) as a system message
- Per-request context (message ID, event type) prepended as `<system-reminder>` to the user message
- In group threads, instructs the agent to stay silent unless mentioned

### `MastraStateAdapter`

`StateAdapter` backed by Mastra storage. Persists thread subscriptions across restarts. Ephemeral data (locks, dedup cache, queues) stays in-memory. Falls back to `@chat-adapter/state-memory` when no storage is configured.

### Per-adapter configuration

```ts
channels: {
  adapters: {
[...]

PR #19570, opened 2026-07-16
fix(core): make lazy chat import visible to bundlers (channels on Cloudflare Workers)

### LINKED ISSUE #19254: [BUG] Channels + Cloudflare Workers — No such module "chat"
# Bug: Channels + Cloudflare Workers — `No such module "chat"`

## Summary
When deploying a Mastra agent that uses a channel adapter (e.g. `@chat-adapter/slack`)
to **Cloudflare Workers** via `@mastra/deployer-cloudflare`, the Worker fails to start /
initialize channels with:

```
Error: No such module "chat"
Failed to initialize channels for agent <id>
```

## Minimal reproduction
1. `npm create mastra@latest` (or any `@mastra/core@1.22.0+` project).
2. Add a channel adapter to an agent:
   ```ts
   import { createSlackAdapter } from '@chat-adapter/slack';
   new Agent({ id: 'weather-agent', /* ... */ channels: { adapters: { slack: createSlackAdapter() } } });
   ```
3. Configure `CloudflareDeployer` and `npm run build && wrangler deploy`.
4. The deployed Worker throws `No such module "chat"`.

## Root cause
`@mastra/core` loads the Chat SDK (`chat`) lazily through an **intentionally
obfuscated dynamic import** in `getChatModule()`:

```js
const mod = "chat";
const chatModule2 = await import(/* @vite-ignore */ /* webpackIgnore: true */ mod);
```

The docstring states this hides `chat` from bundlers because it is ESM-only and
breaks Vitest's module-graph resolution when imported statically. That works on
Node-based runtimes (Node resolves `node_modules` at runtime), but **Cloudflare
Workers require every module to be statically bundled at build time**. A dynamic
`import()` with a non-literal specifier cannot be followed by rollup or wrangler's
esbuild, so `chat` is never included.

This obfuscation also defeats Mastra's own deploy dependency detection: the
bundler scans `output.dynamicImports` for bare specifiers, but a variable specifier
is not reported, so `chat` is omitted from the deploy's `package.json` unless the
user manually adds `bundler.dynamicPackages: ['chat']` — and even then the runtime
`import(mod)` still fails on Workers.

## Expected behavior
Mastra claims Cloudflare Workers as a first-class deployment target **and** supports
channels. The two should compose without a user workaround. For the Workers target
the deployer should rewrite the obfuscated `import(mod)` into a static
`import("chat")` (or otherwise statically bundle `chat`).

## Workaround (current)
- Add `bundler: { externals: true, dynamicPackages: ['chat'] }` to `new Mastra(...)`
  so the package is installed into `.mastra/output/node_modules`.
- After `mastra build`, patch the built output to convert the obfuscated import into
  a static `await import("chat")` so wrangler bundles it. Example script:

  ```js
  // scripts/patch-chat-import.mjs
  import { readFile, writeFile, glob } from 'node:fs/promises';
  import { join } from 'node:path';
  const outputDir = fileURLToPath(new URL('../.mastra/output', import.meta.url));
  const PATTERN = /const\s+(\w+)\s*=\s*"chat";\s*const\s+(\w+)\s*=\s*await\s+import\(\s*(?:\/\*[^*]*\*\/\s*)*\1\s*\)/g;
  for await (const entry of glob('node_modules/**/*.{js,cjs,mjs}', { cwd: outputDir })) {
    const filePath = join(outputDir, entry);
    const code = await readFile(filePath, 'utf8');
    if (!code.includes('"chat"')) continue;
    PATTERN.lastIndex = 0; if (!PATTERN.test(code)) continue; PATTERN.lastIndex = 0;
    const patched = code.replace(PATTERN, (_m, _v, c) => `const ${c} = await import("chat")`);
    if (patched !== code) { await writeFile(filePath, patched); console.log('patched', entry); }
  }
  ```

## Environment
- `@mastra/core`: (installed version, e.g. 1.50.1)
- `@chat-adapter/slack`: 4.33.0
- `@mastra/deployer-cloudflare`: (installed version)
- Node: 24.x
- Deploy target: Cloudflare Workers (`wrangler deploy`)

intojhanurag:
## Description

Agents that use channels (Slack/Discord/Telegram) and deploy to Cloudflare Workers via `@mastra/deployer-cloudflare` fail at channel initialization:

```
Failed to initialize channels for agent <id>: Error: No such module "chat"
```

and channel webhooks respond with `{"error":"Chat not initialized"}`. Reproduced end-to-end with published packages (`@mastra/core@1.51.0`, `@mastra/deployer-cloudflare@1.2.7`) on `workerd` via `wrangler dev`.

**Root cause:** `getChatModule()` in `src/channels/chat-lazy.ts` loaded the ESM-only Chat SDK through a runtime-constructed specifier:

```ts
const mod = 'chat';
await import(/* @vite-ignore */ /* webpackIgnore: true */ mod);
```

A variable specifier is invisible to every bundler. Two consequences:

1. The deployer's dependency analysis (`analyzeEntry.ts` reads Rollup's `output.dynamicImports`) never sees `chat`, so it's omitted from the generated deploy `package.json` — Node deploy outputs need a manual `bundler.dynamicPackages: ['chat']` workaround.
2. Cloudflare Workers can't do runtime module resolution at all (no `node_modules` at runtime), so channels are unusable there regardless of configuration — despite the channels docs advertising Workers support (`waitUntil` auto-detection, #17832).

**Why the obfuscation existed — and why it's safe to remove:**

The pre-squash commits of #14642 (`98c37d72`, `4f256efd`) show the intent: prevent tsup from emitting `require('chat')` in the CJS build (`chat` has no CJS exports, so that's a crash — the original CI failure was a *static* top-level import, correctly fixed by making the import dynamic; the variable trick was added on top as an extra safeguard).

That safeguard is unnecessary in this package's build pipeline, verified empirically:

- Vanilla tsup 8.5.1 (no `treeshake`) **does** rewrite a literal `import('chat')` to `Promise.resolve().then(() => require("chat"))` in `.cjs` output — the concern was legitimate in general.
- But `@mastra/core` builds with `treeshake: { preset: 'smallest' }` (since `aac1667dd0`, Feb 2025), which emits CJS through Rollup — and Rollup preserves native `import()` in `.cjs` output. Confirmed in the actual built artifact: the `.cjs` chunk keeps `await import('chat')`, and no `require('chat')` appears anywhere in `dist`.
- The docstring's Vitest concern also doesn't reproduce: `/* @vite-ignore */` alone keeps Vite/Vitest from pre-resolving the import (511 core tests + typecheck green).

**Verification on workerd (before/after):**

| | before | after |
|---|---|---|
| `mastra build` output | `chat` missing from bundle and `package.json` | Rollup statically inlines `chat` into the bundle |
| `wrangler dev` boot | `No such module "chat"`, channels dead | channels initialize cleanly |
| Slack webhook | `{"error":"Chat not initialized"}` | reaches the Slack adapter (correctly rejects unsigned test request) |

## Changes

- `src/channels/chat-lazy.ts`: literal `import('chat')` instead of the variable specifier; docstring rewritten to document the real constraints (must stay dynamic for CJS, must stay literal for bundlers/serverless)
- Changeset: patch for `@mastra/core`

## Test plan

- [x] `vitest run src/channels` — 239 passed, typecheck clean (the suite the obfuscation was protecting)
- [x] `vitest run src/mastra` — 272 passed
- [x] `pnpm build:core` green; `.cjs` chunk verified to keep native `import('chat')`; no `require('chat')` anywhere in `dist`
- [x] E2E on workerd: fresh project + Slack adapter + `CloudflareDeployer`, patched core → `wrangler dev` → channels initialize, no module error
- [ ] Full CI across dependent packages

## Notes for reviewers

- Related but separate: #19285 (replies dropped after `200 OK` on Workers — `executionCtx` lost at the custom-route bridge) has its own fix in #19299. A full Slack round-trip on Workers needs both PRs; this one makes channels *load*.
- @CalebBarnes — this removes the runtime-constructed specifier from #14642. The CJS concern it guarded against is real in vanilla tsup (reproduced), but core's `treeshake` build path never performs that rewrite. If the `treeshake` option is ever removed from core's tsup config, the rewrite would come back — flagging in case you'd like a build-time guard for that as a follow-up.

Fixes #19254

## ELI5

Cloudflare Workers couldn’t “see” the Chat SDK when you used channel-based agents (so startup failed with `No such module "chat"`). This PR makes the SDK load in a bundler-detectable way, while still keeping it lazy.

## Changes

- Updated `getChatModule()` to use a literal dynamic import (`import('chat')`) instead of a runtime-constructed module specifier, so bundlers can statically include the ESM-only `chat` package on Cloudflare Workers.
- Reworked the docs/comments to clarify why the import must remain dynamic (and to keep CJS consumers from ending up with `require('chat')`).
- Added a patch changeset for `@mastra/core` documenting that Slack/Discord/Telegram channel agents now work on Cloudflare Workers without extra `dynamicPackages` configuration.


CalebBarnes (approved):
✅ LGTM — approving.

I checked the history: the runtime-constructed specifier came from #14642 (`98c37d724` → `4f256efd1`) to stop tsup emitting `require('chat')` in the CJS build, since `chat` is ESM-only. That rewrite doesn't happen on core's current build path — I confirmed in the built artifact that the `.cjs` chunk keeps `await import('chat')` with no `require('chat')` anywhere in `dist`, and the channels suite (239 tests) passes. So the literal specifier is safe and correctly unhides the dep for bundlers/Workers.

Agree with CodeRabbit's changeset nit (https://github.com/mastra-ai/mastra/pull/19570/changes#r3595038458) — worth trimming lines 7-9 to the outcome (channels deploy to Workers with no manual bundler config; unused = still lazy) and dropping the bundler internals, per the changeset styleguide.

Non-blocking follow-up: since safety hinges on core keeping `treeshake` in its tsup config, a build-time guard that fails if `require('chat')` appears in `dist` would be cheap insurance. Fine as a separate PR.

PR #19575, opened 2026-07-16
fix(core): emit durable step-start chunks in the canonical ChunkType shape

### LINKED ISSUE #19574: [BUG] Durable agents emit `step-start` chunks in a non-canonical shape; `@mastra/ai-sdk` crashes on every durable `stream()`/`observe()`
## Describe the Bug

The durable stream adapter publishes `step-start` with its fields flat on the chunk. Every other producer emits the canonical `ChunkType` shape with a `payload` wrapper. Consumers that follow the `ChunkType` contract crash on every durable stream.

`emitStepStartEvent` (`packages/core/src/agent/durable/stream-adapter.ts`):

```ts
await pubsub.publish(AGENT_STREAM_TOPIC(runId), {
  type: AgentStreamEventTypes.STEP_START,
  runId,
  data: { type: 'step-start', ...data },   // ← flat: { type, stepId, request, warnings }
});
```

The canonical contract (`packages/core/src/stream/types.ts`):

```ts
| (BaseChunkType & { type: 'step-start'; payload: StepStartPayload })
```

where `BaseChunkType = { runId: string; from: ChunkFrom; metadata?: ... }`. The regular (non-durable) engine produces exactly that (`loop/workflows/agentic-execution/llm-execution-step.ts`):

```ts
{ runId, from: ChunkFrom.AGENT, type: 'step-start',
  payload: { request: request || {}, warnings: warnings || [], messageId: currentStep.messageId } }
```

The observe-side consumer (`stream-adapter.ts`, `STEP_START` case) enqueues the event's `data` verbatim through an unchecked cast (`streamEvent.data as ChunkType<OUTPUT>`). The cast is why TypeScript never caught the drift:

```ts
case AgentStreamEventTypes.STEP_START: {
  const chunk = streamEvent.data as ChunkType<OUTPUT>;
  if (chunk && 'type' in chunk) safeEnqueue(controller, chunk);
  break;
}
```

The malformed chunk reaches every `ChunkType` consumer. `@mastra/ai-sdk`'s converter (`client-sdks/ai-sdk/src/helpers.ts`) destructures `chunk.payload` for `step-start`:

```ts
case 'step-start':
  const { messageId: _messageId, ...rest } = chunk.payload;   // ← payload is undefined
```

and crashes with:

```
TypeError: Cannot destructure property 'messageId' of 'chunk.payload' as it is undefined.
```

`emitStepFinishEvent` avoids this bug: the `STEP_FINISH` case never enqueues. It invokes the `onStepFinish` callback, whose contract is `AgentStepFinishEventData` rather than `ChunkType`. `STEP_START` is the only event that crosses into the chunk stream.

## Steps to Reproduce

1. Wrap any agent with `createDurableAgent()` / `createEventedAgent()`.
2. `stream()` (or `observe()`) a run and pipe the chunks through `@mastra/ai-sdk`'s stream conversion (e.g. `toUIMessageStream` / the chunk converter), as a chat route does.
3. As soon as the first LLM step starts, the converter throws `Cannot destructure property 'messageId' of 'chunk.payload'`.

Any other consumer written against the `ChunkType` contract (`chunk.payload`) fails the same way. The flat chunk also lacks the `runId`/`from` fields `BaseChunkType` mandates.

## Expected Behavior

Durable `step-start` chunks have the same shape as the regular engine's: `{ type: 'step-start', runId, from: 'AGENT', payload: { messageId, request, warnings, ... } }`.

## Environment

- `@mastra/core`: 1.51.0 (still present on `main` as of `837c8c3037`)
- Affects: `createDurableAgent`, `createEventedAgent` (any durable engine that goes through `emitStepStartEvent`)

## Additional context

We run the evented agent in production behind a `pnpm patch` that wraps the payload. PR incoming: emit the canonical shape from `emitStepStartEvent` (payload wrapper, `runId`/`from`, regular-engine defaults) and pass `messageId` from the durable llm-execution step for full parity.

aisensiy:
## Description

`emitStepStartEvent` published `step-start` with its fields flat on the chunk (`{ type: 'step-start', stepId, request, warnings }`) instead of the canonical `ChunkType` shape (`{ type, runId, from, payload }`) that the regular engine emits and `stream/types.ts` mandates. The observe-side `STEP_START` case enqueues the event's `data` verbatim through an unchecked cast, so the malformed chunk reached every `ChunkType` consumer. `@mastra/ai-sdk`'s converter destructures `chunk.payload` for `step-start` and crashed with `Cannot destructure property 'messageId' of 'chunk.payload' as it is undefined` on every durable `stream()`/`observe()`.

Changes:

- `emitStepStartEvent` now builds a typed canonical chunk (`Extract<ChunkType, { type: 'step-start' }>`): `payload` wrapper, `runId`, `from: ChunkFrom.AGENT`, and the regular engine's defaults (`request: {}`, `warnings: []`). Typing the literal means the next shape drift fails `tsc` before it reaches a consumer at runtime.
- The durable llm-execution call site now passes `messageId` (in scope as `currentMessageId`) for full parity with the regular engine's `step-start` payload.
- Regression tests: unit tests on the emitted event shape, plus an end-to-end durable stream test that asserts `step-start` chunks match the canonical engine shape and survive the exact destructuring `@mastra/ai-sdk` performs. All three fail on the previous code.

This PR leaves `emitStepFinishEvent` alone: `STEP_FINISH` never enqueues into the chunk stream. It feeds the `onStepFinish` callback, whose contract is `AgentStepFinishEventData` rather than `ChunkType`.

## Related issue(s)

Fixes #19574

## Type of change

## Checklist

- [ ] I have made corresponding changes to the documentation (if applicable — none needed)

## ELI5
Durable agents sent “step-start” updates in the wrong format. This fix uses the standard format so consumers can safely read `chunk.payload`.

## Summary
- Updated durable `step-start` events to use the canonical `ChunkType` structure.
- Added `runId`, `from: ChunkFrom.AGENT`, nested `payload`, `messageId`, `request`, and `warnings`.
- Passed the current `messageId` from durable LLM execution.
- Added regression tests for event shape, durable stream output, and `@mastra/ai-sdk` compatibility.
- Added a changeset for the fix.


aisensiy:
This PR has been rebased onto the latest `main` and is ready for maintainer review.

Although #19671 added a defensive fallback in `@mastra/ai-sdk`, the durable producer in `@mastra/core` still emits a non-canonical flat `step-start` chunk. This PR fixes the producer-side contract for all consumers, propagates `messageId`, and includes focused regression tests plus the required changeset.

There is some overlap with #19815, but this PR was opened earlier and includes the complete call-site update and regression coverage. If this approach matches the intended `ChunkType` contract, I would appreciate a review and merge. Thank you!

aisensiy:
Gentle bump — still conflict-free against `main`. The producer in `stream-adapter.ts` on current `main` still emits the flat `step-start` shape, so the destructuring crash in chunk consumers remains reproducible. Related durable-agent fix awaiting review: #19916 (now independently confirmed in production).

abhiaiyer91 (changes requested):
Need to respond to review

abhiaiyer91 (approved):
Re-reviewed against today's `main` — approving, this replaces my earlier changes-requested (both CodeRabbit comments were addressed in c6df2409: test colocated, changeset rewritten action-oriented).

**What I verified locally (head merged with today's main):**
- The bug is still live on `main`: `emitStepStartEvent` still publishes the flat `{ type: 'step-start', ...data }` shape, so this producer-side fix remains necessary even after the defensive consumer fallback in #19671.
- Applied the diff to current `main`: all 75 durable test files pass (680 tests) including the 3 new regression tests, and full core typecheck is clean.
- Consumer path confirmed: the observe-side `STEP_START` handler enqueues `event.data` verbatim, so the canonical `{ type, runId, from, payload }` shape flows straight to `fullStream` — and the top-level `type` guard it checks is preserved.
- `messageId` propagation from the llm-execution call site matches what `client-js` reads (`chunk.payload.messageId`) to keep the assistant message id stable — nice catch including it.

**Why this is the right fix:** it corrects the layer that violated the contract (the durable producer) instead of patching each consumer. Typing the chunk as `Extract<ChunkType, { type: 'step-start' }>` means any future drift from the canonical shape is a compile error, and mirroring the regular engine's `request ?? {}` / `warnings ?? []` defaults keeps shape parity exact.

Thanks for the patience on this one and for the focused regression coverage. 🙏

PR #19578, opened 2026-07-16
fix(core): resolve ToolSearchProcessor meta-tools in DurableAgent

### LINKED ISSUE #19571: [BUG]  DurableAgent: ToolSearchProcessor meta-tools (search_tools / load_tool) throw ToolNotFoundError — durable tool-call step resolves from static registry, not per-step processor tools
### Describe the Bug

When ToolSearchProcessor is attached to an Agent that is run through the durable path (createDurableAgent / DurableAgent.stream), the processor correctly injects the search_tools and load_tool meta-tools into the per-step tool list the LLM sees, but the durable tool-call step cannot execute them. It resolves tools only from the static run-registry (registryEntry.tools, populated once at run start from getToolsForExecution()), which never contains the processor-injected meta-tools. The model therefore sees search_tools, calls it, and receives:

Tool "search_tools" not found. Available tools: <the agent's static eager tools only>. Call tools by their exact name only — never add prefixes, namespaces, or colons.
The non-durable Agent path is unaffected because its tool-call step resolves from the per-step stepTools map (which includes processor-injected tools). This is an in-process parity gap — no cross-process resume, no suspend, no empty registry involved.

Related (same ToolSearchProcessor + DurableAgent parity family, but distinct failure mode): #17714 — there the search_tools/load_tool meta-tools execute but loaded tools don't persist across turns (fixed by #17717, thread-id keying). In contrast, here the meta-tools themselves throw ToolNotFoundError on first call. Notably #17714 (on 1.41.0) reports the meta-tools executing successfully, whereas on 1.49.0/main they don't — possible regression in the durable tool-call step's tool resolution. #19330 is the cross-process flavor of the same registry-resolution gap.

### Steps To Reproduce

2. Install deps: `npm i @mastra/core@1.49.0 @ai-sdk/anthropic zod`.
3. Add the following single file as `repro.mjs` at the project root:

```js
// repro.mjs — DurableAgent + ToolSearchProcessor parity bug.
// No network, no API key. Mocks the Anthropic API to deterministically
// emit a `search_tools` tool-call, then runs the SAME agent through both
// the regular Agent path and the DurableAgent path and compares results.
//
// Run:  node repro.mjs
// Expected: non-durable path executes search_tools successfully;
//           durable path throws ToolNotFoundError for search_tools.

import { createAnthropic } from "@ai-sdk/anthropic";
import { createDurableAgent } from "@mastra/core/agent/durable";
import { ToolSearchProcessor } from "@mastra/core/processors";
import { InMemoryStore } from "@mastra/core/storage";
import { z } from "zod";

// --- Mock Anthropic fetch -------------------------------------------------
// Call 1 -> search_tools tool_use; later calls -> "Done." text.
// Detects `stream` in the request body to serve SSE vs JSON.
function toolUsePayload({ id, name, input }) {
  return {
    id, type: "message", role: "assistant", model: "claude-3-5-sonnet",
    content: [{ type: "tool_use", id: "toolu_" + id, name, input }],
    stop_reason: "tool_use", stop_sequence: null,
    usage: { input_tokens: 10, output_tokens: 15 },
  };
}
function textPayload({ id, text }) {
  return {
    id, type: "message", role: "assistant", model: "claude-3-5-sonnet",
    content: [{ type: "text", text }],
    stop_reason: "end_turn", stop_sequence: null,
    usage: { input_tokens: 10, output_tokens: 5 },
  };
}
function sseBytes(events) {
  let out = "";
  for (const { event, data } of events) {
    out += `event: ${event}\n` + `data: ${JSON.stringify(data)}\n\n`;
  }
  return new TextEncoder().encode(out);
}
function toSSE(payload) {
  const events = [{
    event: "message_start",
    data: { type: "message_start", message: { ...payload, content: [], stop_reason: null, stop_sequence: null, usage: { input_tokens: 10, output_tokens: 1, cache_creation_input_tokens: 0, cache_read_input_tokens: 0 } } },
  }];
  for (const [i, block] of payload.content.entries()) {
    if (block.type === "tool_use") {
      events.push({ event: "content_block_start", data: { type: "content_block_start", index: i, content_block: { type: "tool_use", id: block.id, name: block.name, input: {} } } });
      events.push({ event: "content_block_delta", data: { type: "content_block_delta", index: i, delta: { type: "input_json_delta", partial_json: JSON.stringify(block.input) } } });
    } else {
      events.push({ event: "content_block_start", data: { type: "content_block_start", index: i, content_block: { type: "text", text: "" } } });
      events.push({ event: "content_block_delta", data: { type: "content_block_delta", index: i, delta: { type: "text_delta", text: block.text } } });
    }
    events.push({ event: "content_block_stop", data: { type: "content_block_stop", index: i } });
  }
  events.push({ event: "message_delta", data: { type: "message_delta", delta: { stop_reason: payload.stop_reason, stop_sequence: null }, usage: { output_tokens: payload.usage.output_tokens } } });
  events.push({ event: "message_stop", data: { type: "message_stop" } });
  return sseBytes(events);
}
function makeMockFetch() {
  let call = 0;
  return async (url, options) => {
    call += 1;
    const payload = call === 1
      ? toolUsePayload({ id: "msg_01", name: "search_tools", input: { query: "echo" } })
      : textPayload({ id: "msg_0" + call, text: "Done." });
    let stream = false;
    try { stream = JSON.parse(options?.body ?? "{}").stream === true; } catch {}
    let body, contentType;
    if (stream) {
      body = new ReadableStream({ start(c) { c.enqueue(toSSE(payload)); c.close(); } });
      contentType = "text/event-stream";
    } else {
      const bytes = new TextEncoder().encode(JSON.stringify(payload));
      body = new ReadableStream({ start(c) { c.enqueue(bytes); c.close(); } });
      contentType = "application/json";
    }
    return new Response(body, { status: 200, headers: { "content-type": contentType } });
  };
}

// --- Agent setup ----------------------------------------------------------
const searchable = {
  echo: createTool({
    id: "echo",
    description: "Echo a message back to the user.",
    inputSchema: z.object({ msg: z.string() }),
    execute: async ({ msg }) => ({ echoed: msg }),
  }),
};

function buildAgent() {
  const anthropic = createAnthropic({
    apiKey: "sk-ant-test",
    baseURL: "http://localhost/v1",
    fetch: makeMockFetch(),
  });
  return new Agent({
    id: "repro",
    name: "repro",
    instructions: "If you need a tool you don't have, call search_tools with a keyword query, then load_tool, then call the tool.",
    model: anthropic("claude-3-5-sonnet-20241022"),
    tools: {},
    inputProcessors: [new ToolSearchProcessor({ tools: searchable, search: { topK: 5 } })],
  });
}

// --- Run both paths -------------------------------------------------------
async function collectStream(agent, threadId) {
  const result = await agent.stream(
    [{ role: "user", content: "Use the echo tool to say hi." }],
    { threadId },
  );
  const events = [];
  const stream = result.stream ?? result.fullStream;
  for await (const chunk of stream) {
    const t = chunk?.type;
    if (t === "tool-call" || t === "tool-result" || t === "tool-error") {
      events.push({
        type: t,
        toolName: chunk?.payload?.toolName ?? chunk?.toolName,
        result: chunk?.payload?.result ?? chunk?.result,
        error: chunk?.payload?.error ?? chunk?.error,
      });
    }
  }
  return events;
}

const nondurable = await collectStream(buildAgent(), "nondurable-1");

const base = buildAgent();
const durable = createDurableAgent({ agent: base });
new Mastra({ agents: { repro: durable }, storage: new InMemoryStore() });
const durableEvents = await collectStream(durable, "durable-1");

console.log("\n=== NON-DURABLE Agent (control) ===");
for (const e of nondurable) {
  console.log(`  - ${e.type} toolName=${e.toolName}`);
  if (e.type === "tool-result" && e.toolName === "search_tools") {
    console.log("    result:", JSON.stringify(e.result));
  }
}

console.log("\n=== DURABLE Agent ===");
for (const e of durableEvents) {
  console.log(`  - ${e.type} toolName=${e.toolName}`);
  if (e.type === "tool-error") console.log("    error:", JSON.stringify(e.error));
}

const ndOk = nondurable.some((e) => e.type === "tool-result" && e.toolName === "search_tools" && e.result?.results?.length > 0);
const dErr = durableEvents.some((e) => e.type === "tool-error" && e.toolName === "search_tools");
console.log("\n=== VERDICT ===");
console.log("nonDurable search_tools executed:", ndOk);
console.log("durable search_tools errored:", dErr);
if (ndOk && dErr) {
  console.log("REPRODUCED: search_tools executes on the non-durable path but throws ToolNotFoundError on the durable path.");
  process.exitCode = 0;
} else {
  console.log("NOT REPRODUCED as expected.");
  process.exitCode = 1;
}

Run node repro.mjs.
See error: the durable path prints tool-error toolName=search_tools with {"name":"ToolNotFoundError","message":"Tool \"search_tools\" not found.."}, while the non-durable control path prints tool-result toolName=search_tools with the successful search results.

can't provide blocked by company security

### Expected Behavior

The search_tools meta-tool injected by ToolSearchProcessor should be executable on the DurableAgent path exactly as it is on the regular Agent path. Both paths run the same agent config (same ToolSearchProcessor over the same searchable tools), so when the model emits a search_tools tool-call, the durable tool-call step should resolve and execute that tool and emit a tool-result containing the matching tools — instead of throwing ToolNotFoundError.

Concretely, for the repro above, the DURABLE Agent output should match the NON-DURABLE control output:

tool-call toolName=search_tools
tool-result toolName=search_tools result: {"results":[{"name":"echo","description":"...","score":5.41}], "message":"Found 1 tool(s). Use load_tool ... to make them available."}
…and process.exitCode should be 1 (NOT REPRODUCED), not 0.

### Environment Information

```block
System: OS: Darwin 25.5.0 (macOS) Node: v22.22.3 pnpm: 11.1.2

npmPackages (installed, repro-pinned): @mastra/core: 1.49.0 (also confirmed present on main / 1.51.0 via source) @ai-sdk/anthropic: 3.0.75 zod: 3.25.76

LLM Provider: Anthropic (claude-3-5-sonnet-20241022) Note: the repro mocks the Anthropic HTTP layer, so no API key or network is required — the failure is independent of the provider.

Browser: N/A (server-side agent execution)

Storage backend used in repro: InMemoryStore (production uses a persistent store; the bug reproduces with InMemoryStore, so it is not storage-specific)
```

### Verification

avj21:
## Description

`DurableAgent` threw `ToolNotFoundError` whenever the model called a `ToolSearchProcessor` meta-tool (`search_tools` / `load_tool`), even though the regular `Agent` runs them fine with the same config.

`ToolSearchProcessor` injects those meta-tools per step via `processInputStep`. The regular `Agent` shows the tools to the model and executes them in the same step, so the per-step tool map is enough. The `DurableAgent` instead executes tool calls in a separate workflow step that resolves tools from the run registry — and the processor-injected tools were never written back there, so the lookup missed and the call rejected.

The durable LLM step now converts any raw Mastra tools returned by input processors into `CoreTool` form (the same `makeCoreTool` conversion the non-durable `llm-execution-step` already does) and merges them into the run registry, so the tool-call step can resolve and execute them.

```js
const agent = new Agent({
  name: 'my-agent',
  tools: {}, // no eager tools
  inputProcessors: [new ToolSearchProcessor({ tools: manyTools })],
});
const durable = createDurableAgent({ agent, pubsub });

await durable.stream('What is the weather in NYC?');
// before: search_tools -> ToolNotFoundError
// after:  search_tools -> executes and returns matching tools
```

Added `durable-agent-tool-search-processor.test.ts`, which runs the same setup through both paths and asserts they agree. It fails on `main` (durable case) and passes with the fix.

## Related issue(s)

Fixes #19571

## Type of change

## Checklist

## ELI5

Durable agents can now use tools that get added for a specific step while the agent is running (like a “right helper at the right moment”). This fixes crashes where meta-tools such as `search_tools` or `load_tool` couldn’t find the injected tools and threw `ToolNotFoundError`.

## What changed

- Updated the durable LLM execution step to convert processor-injected Mastra tools into `CoreTool` instances and wire the correct `outputWriter`.
- Stored an exact per-step converted tool snapshot into the durable run registry so the durable tool-call step can resolve and execute injected meta-tools (`search_tools`, `load_tool`) correctly.
- Added a regression test that runs `DurableAgent` with `ToolSearchProcessor` and scripted `search_tools` → `load_tool`, asserting successful `tool-result` payloads (no `tool-error`) and matching tool-result behavior vs the regular (non-durable) `Agent`.
- Added a patch changeset for `@mastra/core` documenting the `DurableAgent` `ToolNotFoundError` fix.


packages/core/src/agent/durable/workflows/steps/llm-execution.ts
```diff
@@ -420,6 +424,52 @@ export function createDurableLLMExecutionStep(_options?: DurableLLMExecutionStep
...
+                          name,
+                          runId,
+                          threadId: typedInput.state?.threadId,
+                          resourceId: typedInput.state?.resourceId,
+                          logger: boundLogger,
+                          mastra: mastra ? createMastraProxy({ mastra, logger: boundLogger }) : undefined,
+                          memory: registryEntry?.memory,
+                          agentName: typedInput.agentName ?? agentId,
+                          requestContext,
+                          workspace: registryEntry?.workspace,
```
taofeeq-deru, line 457: `outputWriter` is missing here. You can handle it the way it's handled in tool-call.ts - https://github.com/avj21/mastra/blob/fbf919d7126ada760b12176d55d9857fde0ceabe/packages/core/src/agent/durable/workflows/steps/tool-call.ts#L725

taofeeq-deru (changes requested):

packages/core/src/agent/durable/workflows/steps/llm-execution.ts
avj21: Done. wired it same way.

taofeeq-deru (approved):

PR #19579, opened 2026-07-16
docs: document build-time entry configuration

### LINKED ISSUE #19382: [DOCS] Document the entry file contract: which new Mastra({...}) options must be object literals
### Describe the Improvement Needed

Build-time options like `bundler` are extracted **statically** from the entry file's AST (`packages/deployer/src/build/shared/extract-mastra-option.ts`). This works:

```ts
  bundler: { externals: true },
});
```

These silently do not:

```ts
export const mastra = new Mastra(buildMastraOptions()); // bundler config lost

const options = { bundler: { externals: true } };
export const mastra = new Mastra({ ...options }); // also lost
```

When extraction fails, the deployer falls back to default bundler options and logs only at **debug** level — so a perfectly reasonable refactor (extracting config into a factory function, or spreading a shared options object) silently changes production bundling behavior with no visible signal.

This constraint was confirmed as by-design in #6621 and #8861, and a dedicated docs page was endorsed by a maintainer in #6621 — but that page still doesn't exist. Proposal: a page under the deployment docs ("The Mastra entry file contract") covering:

- the requirement that the entry file contains `export const mastra = new Mastra({ <object literal> })`
- which options are AST-extracted at build time (`bundler`, `server`, ...) and therefore must be literal
- which patterns break extraction — factory calls, spreads, `let` reassignment — with links to #6621, #6750, #8861
- a cross-link from the monorepo docs page, where builds depend most heavily on `bundler` config being honored

With the contract written down, teams can protect it in code review and lint rules instead of discovering it in production. A possible follow-up (separate, code change): upgrade the debug log to a `logger.warn` when extraction fails and the entry source mentions `bundler`, pointing at the new docs page — but the docs page is valuable on its own.

### Documentation Type

Missing Information

AkhilTrivediX:
Closes #19382

Documents the entry-file contract for `bundler`, `deployer`, and `server`, including supported and unsupported shapes. Adds a link from the monorepo deployment guide so the constraint is visible where workspace bundling is configured.

Validated with `pnpm --dir docs build`, Prettier, and focused MDX lint.

## ELI5

Mastra now explains which configuration settings it can read while building your app and shows the correct way to write them. The deployment guide also links to these rules so users can avoid configurations that silently fall back to defaults.

## Changes

- Documented build-time extraction for `bundler`, `deployer`, and `server`.
- Explained the required direct object-literal shape in `new Mastra({ ... })`.
- Documented unsupported patterns, including factory calls, object spreads, and reassigned `let` values.
- Added guidance about fallback build behavior and linked the configuration reference.
- Cross-linked the monorepo deployment guide, with emphasis on keeping `bundler` as a direct property.
- Validated with the docs build, Prettier, and focused MDX lint.

### DIFF
--- a/docs/src/content/en/docs/deployment/mastra-server.mdx
+++ b/docs/src/content/en/docs/deployment/mastra-server.mdx
@@ -70,6 +70,41 @@ Read the [`mastra start`](/reference/cli/mastra#mastra-start) reference for all 
 
 ## Build configuration
 
+### Build-time configuration
+
+Mastra reads the `bundler`, `deployer`, and `server` options while it builds your application. Keep those options as direct properties of the object passed to `new Mastra()` so the build can extract them.
+
+The following entry-file shape works:
+
+```typescript title="src/mastra/index.ts"
+import { Mastra } from '@mastra/core/mastra'
+
+export const mastra = new Mastra({
+  bundler: {
+    externals: ['sharp'],
+  },
+  server: {
+    port: 4111,
+  },
+})
+```
+
+The value of each option can come from a variable, import, or function call. The option itself must remain a direct property. Do not hide build-time options behind a factory call or an object spread:
+
+```typescript title="src/mastra/index.ts"
+const options = {
+  bundler: {
+    externals: ['sharp'],
+  },
+}
+
+// These patterns prevent Mastra from extracting `bundler` during the build.
+export const mastra = new Mastra(createMastraOptions())
+export const otherMastra = new Mastra({ ...options })
+```
+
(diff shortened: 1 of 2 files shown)


AkhilTrivediX:
Thanks for the suggestion. I checked the current extractor before changing the docs.

The build-time option must be a direct property of the object passed to `new Mastra()`. A direct identifier value, including one declared with `let` and reassigned at top level, is retained by the generated config export and is evaluated when the extracted config is loaded. I therefore kept the examples focused on the patterns that prevent extraction: a factory call for the whole Mastra config and an object spread that hides the option property.

The PR already links #19382 in its description. I kept the page focused on the supported entry-file contract rather than adding implementation-issue links for readers.

LekoArts (approved):
Thanks 👍

PR #19600, opened 2026-07-16
fix(core): prevent retained output from blocking the event loop

roaminro:
Bounded process output currently rescans almost the entire retained buffer whenever a small streamed chunk crosses the limit. Once a command reaches the 1 MiB cap, that synchronous work can delay unrelated requests on the Node.js event loop.

Before:

```ts
trimToMaxBytes(firstChunk.data, firstChunk.bytes - overflowBytes)
```

After:

```ts
advanceStartByUtf8Bytes(firstChunk.data, firstChunk.start, overflowBytes)
```

The retained buffer now advances a chunk-local offset by only the newly dropped bytes. It keeps Unicode boundaries and dropped-byte metadata intact, periodically compacts chunks, and detaches retained suffixes from oversized source strings so the memory bound remains effective.

The focused tests cover complexity, repeated multibyte trimming, compaction, and oversized chunks. In a local benchmark, 200 post-cap 1 KiB appends dropped from about 2.3 seconds to about 1 millisecond while retaining the same byte counts.

## ELI5

When a command produces lots of output, the system now removes old text without repeatedly rereading everything it has kept. This prevents slowdowns while preserving complete characters and accurate dropped-byte counts.

## Summary

- Reworked bounded process-output trimming to incrementally advance UTF-8 offsets instead of rescanning retained data.
- Preserved Unicode boundaries, retained byte limits, and dropped-byte metadata.
- Added periodic chunk compaction and detached suffixes for oversized source chunks to limit memory retention.
- Added regression tests for multibyte output, compaction, oversized chunks, and trimming complexity.
- Added a `@mastra/core` patch changeset.

### DIFF
--- a/.changeset/bright-ends-dress.md
+++ b/.changeset/bright-ends-dress.md
@@ -1,0 +1,5 @@
+---
+'@mastra/core': patch
+---
+
+Fixed bounded process output stalling other requests while high-volume commands continue streaming.
--- a/packages/core/src/workspace/sandbox/process-manager/process-handle.test.ts
+++ b/packages/core/src/workspace/sandbox/process-manager/process-handle.test.ts
@@ -1,6 +1,6 @@
 import { once } from 'node:events';
 
-import { describe, expect, it } from 'vitest';
+import { describe, expect, it, vi } from 'vitest';
 
 import type { MastraSandbox } from '../mastra-sandbox';
 import type { CommandResult } from '../types';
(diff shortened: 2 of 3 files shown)


NikAiyer (approved):
The incremental UTF-8 offset approach fixes the retained-output rescanning stall while preserving Unicode boundaries, byte accounting, callback/reader delivery, compaction, and bounded backing-store memory. Focused ProcessHandle tests and the core TypeScript check pass locally. The unrelated MongoDB timeout should still receive a green rerun before merge.

PR #19618, opened 2026-07-17
fix(core): support TypeScript 7 native LSP in workspace lsp_inspect

### LINKED ISSUE #19601: LSP: built-in TypeScript server definition broken with TypeScript 7 -- tsserver.js no longer exists
### Describe the Bug

When lsp: true is set on a Workspace that runs in a TypeScript 7 project, the lsp_inspect tool returns "LSP isn't available for this file type." The built-in TypeScript server definition hardcodes typescript/lib/tsserver.js in its binary resolution, but TypeScript 7 removed this file as part of the migration to ts_ls.

### Steps To Reproduce

1. Create a TypeScript 7 project (tsc --version → 7.x)
2. Configure a Mastra workspace with lsp: { searchPaths: [projectDir] }
3. Have an agent call the lsp_inspect tool on any .ts file
4. Observe: "LSP isn't available for this file type"
5. Confirm node_modules/typescript/lib/tsserver.js does not exist (TS7 removed it)

<script src="https://gist.github.com/dingonewen/63fc816dcce180fd380145dc5a21abe5.js"></script>

### Expected Behavior

LSP should work with TypeScript 7 projects. The TypeScript server definition needs to support the new ts_ls binary path instead of the deprecated tsserver.js.

### Environment Information

```block
@mastra/core: 1.50.1
TypeScript: 7.0.2
typescript-language-server: 5.x

The ecosystem migration from tsserver to ts_ls is well underway — lspconfig dropped tsserver in 0.2.0, and typescript-language-server v5.x has been updated. This is likely just a matter of updating the built-in server definition's binary resolution to look for the new path.
```

### Verification

intojhanurag:
## Description

Fixes workspace LSP support for TypeScript 7 projects. TS 7 removed `typescript/lib/tsserver.js` (which the built-in server definition hardcoded), so `lsp_inspect` reported "LSP isn't available for this file type" for any TS 7 project. The built-in TypeScript server now detects the installed version and launches TS 7's native Go-based server (`tsc --lsp --stdio`), and the LSP client supports pull diagnostics, which TS 7's server uses exclusively.

- Server resolution branches on `tsserver.js` presence: TS ≤6 keeps the `typescript-language-server` wrapper (unchanged), TS 7+ uses `tsc --lsp --stdio`
- Version guard: the `tsc` path is only used when the resolved `typescript` package is v7+, so a stale TS ≤6 `tsc` on PATH can never be handed `--lsp` (previously caused a 10s init hang per inspect call in projects without TypeScript)
- Monorepo/hoisted installs: the `tsc` binary is resolved next to wherever the `typescript` package resolves, not just the project root's `node_modules/.bin`
- LSP client detects `diagnosticProvider` in the initialize response and issues `textDocument/diagnostic` pull requests instead of polling for push notifications; also handles `client/registerCapability` requests sent by TS 7 during init
- Tests are deterministic (fake typescript installs + `process.cwd()` stub so the repo's own TS 5 can't leak in via the cwd resolution fallback) and include a fake pull-diagnostics server subprocess test; all new fix tests verified to fail against main's implementation
- Verified end-to-end against a real `typescript@7.0.2` install: pull diagnostics in ~38ms, hover working; TS ≤6 push path behavior unchanged

## Related Issue(s)

Fixes #19601

## Type of Change

- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [ ] Test update

## Checklist

## ELI5

TypeScript 7 started launching its language server differently, and that could make workspace “code help” (like diagnostics) stop working. This PR figures out which TypeScript you have installed, starts the right kind of server, and teaches the LSP client how to get TypeScript 7 diagnostics using the new “pull” approach.

## What changed

- **Workspace TypeScript server startup now branches by installed TS version**
  - **TypeScript ≤6:** uses the `typescript-language-server` wrapper (via `tsserver.js`)
  - **TypeScript 7+:** uses native LSP via `tsc --lsp --stdio`
- **More reliable command resolution**
  - Resolves `tsc` relative to the *installed* `typescript` package (avoids relying on potentially stale/incorrect PATH binaries)
  - Supports monorepo/hoisted installs and pnpm-style layouts
  - Returns `undefined` when the required pieces aren’t verifiably present (instead of guessing)
- **LSP client support for TypeScript 7 “pull diagnostics”**
  - During initialization, registers handling for `client/registerCapability`
  - Detects whether the server advertises `diagnosticProvider`
  - If supported, `waitForDiagnostics` sends `textDocument/diagnostic` and returns results immediately (TypeScript 7+ behavior)
  - If not supported, preserves the existing **push/poll-style** diagnostics behavior (TypeScript ≤6)
- **Tests**
  - Added deterministic Vitest coverage for:
    - TypeScript server command/initialization resolution across simulated install shapes (including hoisted/pnpm-like scenarios)
    - `buildServerDefs({ searchPaths })` effects on TypeScript/`tsc` resolution
    - Correct **TypeScript 7 pull diagnostics** behavior using a fake diagnostic server
  - Includes an end-to-end verification against **TypeScript 7.0.2**


packages/core/src/workspace/lsp/servers.ts
```diff
@@ -234,14 +234,37 @@ export function buildServerDefs(config?: LSPConfig): Record<string, LSPServerDef
...
+        if (!pkg) return undefined;
+        const version = (pkg.require(pkg.resolved) as { version?: string }).version ?? '';
+        if (!(parseInt(version, 10) >= 7)) return undefined;
+        // Prefer the tsc shim that sits next to the resolved typescript package
+        // (handles hoisted/monorepo installs where root has no node_modules/.bin).
+        const siblingBin = join(dirname(dirname(pkg.resolved)), '.bin', 'tsc');
+        if (existsSync(siblingBin)) return `${siblingBin} --lsp --stdio`;
+        // No PATH fallback: a bare `tsc` on PATH cannot be verified to belong
+        // to the typescript install whose version we just validated.
+        const tscBin = resolveNodeBin(root, 'tsc', searchPaths);
```
NikAiyer: **The fallback `tsc` is not guaranteed to belong to the TypeScript package whose version was validated.**

`resolveRequireFromPaths()` and `resolveNodeBin()` use different resolution flows. For example, if `typescript@7` is resolved from `searchPaths`, its adjacent `.bin/tsc` is absent, and the workspace root contains a TypeScript 5 shim, this fallback launches the root shim with `--lsp` even though the version check was performed against the search-path package.

The generic `.bin` fallback is needed for pnpm layouts, where the resolved package may live in the pnpm store while the usable shim lives at the workspace root. Could we verify that the selected shim targets the resolved package's `bin.tsc` entry before launching it?

Please also add a regression test with conflicting installations: TypeScript 7 resolved from a search path, an unrelated earlier `tsc` in the workspace root, and no adjacent search-path shim. A positive pnpm-style test where the root shim targets the resolved TS7 package would protect the legitimate hoisted case.

packages/core/src/workspace/lsp/types.ts
```diff
@@ -70,7 +70,7 @@ export interface LSPConfig {
    * Extra directories to search for both language server binaries and Node.js modules.
    * Each entry should be a directory whose node_modules contains the required packages.
    * Searched after project root and process.cwd() — for binaries in node_modules/.bin/,
-   * and for modules like typescript/lib/tsserver.js.
+   * and for modules like typescript/lib/tsserver.js (TS ≤6).
```
NikAiyer: Could we mention the TS7 lookup here too? `searchPaths` is now also used to resolve `typescript/package.json` and the `tsc` shim for TypeScript 7+, while this public API documentation only describes the TS ≤6 module path. The internal comment in `servers.ts` already describes both branches.

.changeset/three-months-appear.md
```diff
@@ -0,0 +1,7 @@
+---
+'@mastra/core': patch
+---
+
+Fixed workspace LSP support for TypeScript 7 projects. TypeScript 7 removed `tsserver.js`, which broke code inspection (diagnostics, hover) in workspaces using it. The built-in TypeScript server now detects the installed TypeScript version and starts the right language server automatically, and the LSP client supports pull diagnostics for servers that require them (TypeScript 7's native server and newer ESLint language servers). No configuration changes are needed — existing `lsp: true` setups keep working, and TypeScript ≤6 behavior is unchanged.
```
NikAiyer: The TypeScript 7 claim is well covered, but the newer-ESLint claim is not exercised by this PR's tests. Since this changeset is primarily for #19601, would you mind either keeping it focused on TS7 or pointing to the ESLint server behavior that verifies this path?

packages/core/src/workspace/lsp/servers.ts
intojhanurag: Rather than verifying shim provenance, I removed shims from the equation: we now read the validated package's own bin.tsc entry and run node <pkgDir>/bin/tsc --lsp --stdio  directly

NikAiyer (approved):
The package/compiler association issue is resolved by launching the validated TypeScript package's own bin entry, with focused coverage for conflicting installs and pnpm-style symlinked layouts. The searchPaths docs and changeset were also updated as requested. Verified with pnpm build:core and the focused LSP server tests (87/87 passing).

PR #19673, opened 2026-07-17
feat(mastracode): enable prompt caching on the Amazon Bedrock path

### LINKED ISSUE #19672: [BUG] mastracode sends no prompt-cache breakpoints on the Amazon Bedrock path (~10x input cost on long threads)
### Bug

mastracode never sends prompt-cache breakpoints on the **Amazon Bedrock** path, so long agentic threads re-pay full-price input on every turn. On a real workload this produced a ~10x cost blow-up (near-100% cache hit → ~30-50% the day Bedrock became the active provider).

### Root cause (verified in source, `mastracode@0.31` / current `main`)

Prompt caching is wired **only for the Anthropic paths**, via `providerOptions.anthropic.cacheControl`:

- `mastracode/sdk/src/agents/mastracode-gateway.ts` applies `promptCacheMiddleware` in exactly two places: `anthropicApiKeyProvider` and the Anthropic-OAuth branch.
- `openaiApiKeyProvider` ships `middleware: []`.
- `mastracode/sdk/src/providers/amazon-bedrock-gateway.ts` returns a bare `createAmazonBedrock()(modelId)` — **no cache middleware at all** (`grep -c cacheControl` → 0).

Two independent reasons Bedrock gets no caching:
1. The Bedrock gateway wraps no cache middleware.
2. Even the existing middleware wouldn't help: it writes `providerOptions.anthropic.cacheControl`, but `@ai-sdk/amazon-bedrock` reads **`providerOptions.bedrock.cachePoint`** (`{ type: 'default' }`) — a different namespace/shape (see `getCachePoint()` in the provider, which reads `providerMetadata.bedrock.cachePoint`).

### Impact

Bedrock users on Opus/Sonnet with long threads pay standard input ($15/M for Opus) instead of cache-read ($1.50/M) for the re-sent prefix on every turn. Real bill: uncached input jumped from ~$3/day to $1,000-3,000/day when Bedrock became active, cache hit-rate fell from ~100% to ~30-50%.

### Reproduction / proof of fix

A `bedrock.cachePoint` middleware (mark last system message + last message) makes the cache engage. Live Opus-4.8-on-Bedrock, identical system prefix across turns:

```
with middleware:  turn1 cacheWriteInputTokens=8509   turn2 cacheWriteInputTokens=8   (8509-token prefix served from cache)
no middleware:    no cache usage on any turn
```

### Proposed fix

Add a Bedrock cache-point middleware (parallel to `promptCacheMiddleware`) and apply it in `AmazonBedrockGateway.resolveLanguageModel`:

```ts
const bedrockCacheMiddleware: LanguageModelMiddleware = {
  specificationVersion: 'v3',
  transformParams: async ({ params }) => {
    const prompt = [...params.prompt];
    const cp = { bedrock: { cachePoint: { type: 'default' as const } } };
    const mark = (m: any) => ({ ...m, providerOptions: { ...m.providerOptions, ...cp } });
    for (let i = prompt.length - 1; i >= 0; i--) if (prompt[i].role === 'system') { prompt[i] = mark(prompt[i]); break; }
    prompt[prompt.length - 1] = mark(prompt[prompt.length - 1]);
    return { ...params, prompt };
  },
};
```

(Guard for models that don't support cache points; Anthropic + a few others on Bedrock do.)

### Additional context

Found while auditing a Bedrock bill. Happy to open a PR. Related: #17937 (added the Bedrock provider without caching), #19664 (custom-provider SigV4/responses — the OpenAI-compatible custom path has the same no-cache gap, though `/responses` caches server-side).

Sri01728:
Fixes #19672

## What

Enables prompt caching on the Amazon Bedrock path. mastracode only wired caching for its Anthropic providers, so long agentic threads on Bedrock re-paid full-price input on every turn instead of the cache-read rate — a ~10x input-cost increase on large threads.

## Root cause

Prompt caching was applied only via `providerOptions.anthropic.cacheControl` in the Anthropic API-key and OAuth branches. The Bedrock gateway (`amazon-bedrock-gateway.ts`) returned a bare `createAmazonBedrock()(modelId)` with no cache middleware — and `@ai-sdk/amazon-bedrock` reads a **different** key, `providerOptions.bedrock.cachePoint` (`{ type: 'default' }`), so even the existing middleware wouldn't have applied.

## Fix

`bedrockProvider()` now wraps its model with `bedrockCacheMiddleware`, which inserts a Bedrock cache point at the last system message and the most recent message (the same two-breakpoint placement the Anthropic path uses). Models that don't support cache points ignore the field.

## Verification

Live against Bedrock Opus 4.8, raw Converse `usage` on an identical prefix across two calls:

```
call 1:  cacheWriteInputTokens = 9013,  cacheReadInputTokens = 0
call 2:  cacheReadInputTokens  = 9013,  inputTokens = 2
```

Call 2 served the 9,013-token prefix from cache (billed at the cache-read rate) instead of re-billing it as standard input. Without the middleware, no cache usage is reported on any call.

- New unit test `amazon-bedrock-cache.test.ts` (3 tests): breakpoint placement with/without a system message, and preservation of existing `providerOptions`.
- Existing Bedrock tests unchanged and green; `tsc --noEmit` and `eslint` clean.

## Notes

- Type cast on `bedrock(modelId)`: `@ai-sdk/amazon-bedrock` returns `LanguageModelV2` while `wrapLanguageModel` is typed for V3 (wrapped at runtime for both) — same bridge the gateway uses elsewhere.
- Found while auditing a real Bedrock bill where cache hit-rate fell from ~100% to ~30-50% the day Bedrock became the active provider.

## ELI5

Amazon Bedrock can now remember parts of long conversations, so repeated prompts cost less and work faster. The change adds cache markers while preserving existing settings.

## What changed

- Added Bedrock prompt-cache middleware using `providerOptions.bedrock.cachePoint`.
- Places cache points on the last system message and newest message.
- Preserves existing provider options.
- Wraps Bedrock models with the middleware; unsupported models can ignore the metadata.
- Added unit tests and a changeset.

## Verification

- Confirmed a 9,013-token prefix was served from cache on a second live Bedrock call.
- TypeScript and ESLint checks pass.

### DIFF
--- a/.changeset/mastracode-bedrock-prompt-cache.md
+++ b/.changeset/mastracode-bedrock-prompt-cache.md
@@ -1,0 +1,9 @@
+---
+'@mastra/code-sdk': patch
+---
+
+Enable prompt caching on the Amazon Bedrock path.
+
+mastracode only wired prompt caching for its Anthropic providers (via `providerOptions.anthropic.cacheControl`); the Bedrock path had none. Because `@ai-sdk/amazon-bedrock` reads a different key (`providerOptions.bedrock.cachePoint`), long agentic threads on Bedrock re-paid full-price input on every turn instead of the cache-read rate — a ~10x input-cost increase on large threads.
+
+`AmazonBedrockGateway` now wraps its models with a middleware that inserts Bedrock cache points at the last system message and the most recent message (mirroring the Anthropic path's two breakpoints). Models that don't support cache points ignore the field.
--- a/mastracode/sdk/src/providers/__tests__/amazon-bedrock-cache.test.ts
+++ b/mastracode/sdk/src/providers/__tests__/amazon-bedrock-cache.test.ts
@@ -1,0 +1,49 @@
+import { describe, expect, it } from 'vitest';
+import { bedrockCacheMiddleware } from '../amazon-bedrock-gateway.js';
+
+// The middleware marks cache breakpoints with `providerOptions.bedrock.cachePoint`
+// so Bedrock bills the re-sent prefix at the cache-read rate. Bedrock reads this
+// key (not Anthropic's `cacheControl`), which is why the general prompt-cache
+// middleware never applied on the Bedrock path.
+const CACHE_POINT = { cachePoint: { type: 'default' } };
+
+async function transform(prompt: Array<{ role: string; content: unknown; providerOptions?: unknown }>) {
+  const result = await bedrockCacheMiddleware.transformParams!({
+    type: 'generate',
+    params: { prompt } as never,
+    model: {} as never,
+  });
+  return result.prompt as Array<{ role: string; providerOptions?: { bedrock?: unknown } }>;
+}
+
+describe('bedrockCacheMiddleware', () => {
+  it('marks the last system message and the most recent message as cache points', async () => {
+    const out = await transform([
+      { role: 'system', content: 'you are helpful' },
+      { role: 'user', content: 'first' },
+      { role: 'assistant', content: 'ok' },
+      { role: 'user', content: 'latest' },
+    ]);
+
+    expect(out[0]!.providerOptions?.bedrock).toEqual(CACHE_POINT); // system
+    expect(out[3]!.providerOptions?.bedrock).toEqual(CACHE_POINT); // most recent
+    expect(out[1]!.providerOptions?.bedrock).toBeUndefined(); // untouched middle
+    expect(out[2]!.providerOptions?.bedrock).toBeUndefined();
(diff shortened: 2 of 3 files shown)


abhiaiyer91:
Thanks for the detailed report and implementation. We confirmed the Bedrock caching gap, but we're going to replace this with a maintainer-owned PR so we can preserve existing Bedrock provider options, add the missing regression coverage, and validate the provider behavior end to end. Closing this attempt in favor of that focused follow-up.

PR #11755, opened 2026-01-09
Added execution path and remove extra verbose logs

intojhanurag:
## Description

This PR addresses inefficiencies in verbose tool calls within Mastra workflows. Currently, the model receives the entire workflow execution log with all intermediate values, which leads to:

- Duplication in execution JSON, where the output of one step is repeated as the input of the next.
- Verbose outputs that are still incomplete, missing the actual execution path.

The proposed changes aim to map verbose outputs to a more compact and user-friendly text format, reducing context usage and making workflows more efficient.

## Related Issue(s)

- #8951 

## Type of Change

- [ ] Bug fix (non-breaking change that fixes an issue)
- [x] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation update
- [ ] Code refactoring
- [x] Performance improvement
- [ ] Test update

## Checklist

- [ ] I have made corresponding changes to the documentation (if applicable)
- [ ] I have added tests that prove my fix is effective or that my feature works

## Summary by CodeRabbit

* **New Features**
  * Execution-path (stepExecutionPath) tracking added across workflow runs and exposed to time-travel, restart, and resume flows.

* **Improvements**
  * Trimmed redundant step payloads during formatting to reduce replay/snapshot size.
  * Step result payloads made optional for more flexible state handling.
  * Lifecycle callbacks, persisted state, and formatted results now include execution-path info for accurate replays.

* **Documentation**
  * Added changelog entry describing execution-path propagation and payload optimization.

PR #19676, opened 2026-07-18
fix(core): bound the step-payload dedup comparison in fmtReturnValue

### LINKED ISSUE #19373: [BUG] Streaming agent runs re-serialize the full run context once per chunk in `fmtReturnValue` — O(n²) CPU, pins a core on long streams
### Describe the Bug

`DefaultExecutionEngine.fmtReturnValue` deduplicates workflow step payloads with a `JSON.stringify` structural comparison. For user-authored workflows this runs once per run and is fine. But agent **output processors also execute on the workflow engine — once per stream chunk** — and each processor step's `payload`/`output` carries the agentic loop's execution context, including the accumulated `streamParts` array (every stream part emitted so far). So every chunk of a streaming agent run re-serializes a multi-megabyte, ever-growing object roughly twice per processor step.

Cost is O(chunks × accumulated context size) — quadratic in output length. On a 1–2 vCPU container this pins a full core for the entire run and starves the event loop. `agent.generate()` is unaffected (no per-chunk pipeline), which makes the problem easy to misattribute to app code.

The comparison, in [`packages/core/src/workflows/default.ts` (`fmtReturnValue`, ~L580)](https://github.com/mastra-ai/mastra/blob/main/packages/core/src/workflows/default.ts#L580-L589):

```ts
// Remove payload if it matches the output of the previous step (structural comparison
// handles deserialized data where reference equality would fail)
let payloadMatchesPrevious = false;
if (hasPreviousOutput) {
  try {
    payloadMatchesPrevious =
      optimizedStep.payload === previousOutput ||
      JSON.stringify(optimizedStep.payload) === JSON.stringify(previousOutput);
  } catch {
```

Introduced in #11755 (shipped in `@mastra/core` 1.9.0) as a context-size optimization for workflow execution logs (#8951). The references never match at this call site (values are cloned between steps), so both sides serialize in full on every call.

**Measurements** (chat agent with 10 output processors incl. memory processors, one streamed run producing a ~60 KB reply; instrumented live via a CDP logpoint on `fmtReturnValue` plus a sampling CPU profile):

- `fmtReturnValue` called **424×** (≈ once per chunk)
- the `stepResults` object it stringifies grew to **~33 MB per call** by late-run (10 steps × `payload`+`output` ≈ 1.8 MB each, dominated by `streamParts` ≈ 1.7 MB)
- net: **≈13 GB of `JSON.stringify` work for a 60 KB reply**
- CPU profile: `fmtReturnValue` **4.21 s self-time — the only hot application frame** in the process
- same prompt via `generate()`: ~49 engine completions vs ~440, near-idle

Production impact (Cloud Run, 1 vCPU): long multi-agent streaming runs pinned the instance at ~100% CPU for their full duration, and the starved event loop degraded unrelated features (e.g. `resumable-stream`'s 1 s resume ack window). With the comparison bounded (workaround below), the same class of audit run went from **60+ minutes to 4 minutes**, peaking at **23% CPU** — the pin was also inflating the wall-time of every LLM/tool round-trip.

| build                                | dedup self-time per streamed run |
| ------------------------------------ | -------------------------------- |
| stock 1.50.0                         | 4.21 s, grows with output length |
| stringify aborted past 32 KB (patch) | 0.15 s, hard-bounded             |

**Suggested fix directions**

1. Bound the comparison: abort the stringify once a size estimate passes a threshold and treat the payload as non-matching. Keeps the log-size optimization for the small payloads it was designed for; oversized payloads are simply retained, which is safe (the dedup is conditional, so consumers already handle payload-present).
2. Skip the dedup for the internal per-chunk processor pipeline entirely — its result is consumed internally, not an execution log anyone reads.
3. Longer term: avoid carrying `streamParts`/the full loop context in every processor step's `payload`/`output`.

<details>
<summary><strong>Workaround we're running in production</strong> (pnpm patch on 1.50.0 — happy to turn into a PR)</summary>

Replaces the two bare `JSON.stringify` calls with a bounded serializer; a counting replacer throws a sentinel past ~32 KB:

```js
const PATCH_DEDUP_MAX_CHARS = 32768;
const PATCH_TOO_BIG = {};
const patchBoundedStringify = value => {
  let approxChars = 0;
  try {
    return JSON.stringify(value, (key, v) => {
      approxChars += key.length + 3 + (typeof v === 'string' ? v.length + 2 : 8);
      if (approxChars > PATCH_DEDUP_MAX_CHARS) {
        throw PATCH_TOO_BIG;
      }
      return v;
    });
  } catch (e) {
    if (e === PATCH_TOO_BIG) return PATCH_TOO_BIG;
    throw e;
  }
};

// in fmtReturnValue:
if (optimizedStep.payload === previousOutput) {
  payloadMatchesPrevious = true;
} else {
  const payloadJson = patchBoundedStringify(optimizedStep.payload);
  const previousJson =
    payloadJson === PATCH_TOO_BIG ? PATCH_TOO_BIG : patchBoundedStringify(previousOutput);
  payloadMatchesPrevious =
    payloadJson !== PATCH_TOO_BIG && previousJson !== PATCH_TOO_BIG && payloadJson === previousJson;
}
```

Under the cap this is byte-for-byte the stock comparison (small workflow-as-tool results still dedup); over it, per-call work is hard-bounded.

</details>

Related: #19210 looks like the same growth pattern surfacing through `@mastra/inngest`. Separately, `@mastra/ai-sdk` re-emits cumulative `data-tool-agent`/`-network`/`-workflow` parts on every nested delta — a smaller, independent O(n²) on the SSE path; report to follow.

### Steps To Reproduce

1. Create an agent with one or more output processors (memory processors count) and stream a response of a few thousand tokens via `agent.stream()`.
2. Watch `process.cpuUsage()` (or a sampling profiler) during the run: CPU grows superlinearly with output length, and the profile lands on `fmtReturnValue`.
3. Run the same prompt via `agent.generate()`: near-idle. The stream/generate asymmetry is the fingerprint.
4. To see the exact per-call cost without modifying code: open the inspector on the running server (`kill -USR1 <pid>`), set a `Debugger.setBreakpointByUrl` logpoint on `fmtReturnValue`'s first line with a condition that logs `JSON.stringify(stepResults).length` and returns `false` — call count ≈ chunk count, and the size grows every call.

N/A — needs a streaming agent with output processors and a long response, so not meaningfully reducible below "any real agent"; call-site analysis, live measurements, and a validated patch are included above instead.

### Expected Behavior

Per-completion result formatting should do bounded work — it shouldn't re-serialize the run's accumulated streaming context, so streaming CPU scales linearly with output length (as `generate()` already does), instead of quadratically until a core is pinned.

### Environment Information

```block
- `@mastra/core`: 1.41.0 and 1.50.0 (verified on both; the call site is unchanged in 1.50.1)
- `ai`: 6.0.158
- Node.js: 22.x
- OS: Linux (Google Cloud Run) and macOS (local repro/profiling)
```

### Verification

xianjianlf2:
## Description

`fmtReturnValue` dedupes step payloads with an unbounded `JSON.stringify` structural comparison. Agent output processors run on the workflow engine once per stream chunk, and their step payload/output carry the accumulated streaming context, so every chunk of a streaming agent run re-serializes a multi-megabyte, ever-growing object — O(chunks × context size) CPU that pins a core on long streams (`generate()` is unaffected, which makes it easy to misattribute).

This PR bounds the comparison: a counting replacer aborts the stringify once a size estimate passes 32 KB and treats the value as non-matching. Under the cap the comparison is byte-for-byte identical to the current behavior, so the log-size optimization from #11755 still applies to the small payloads it was designed for. Oversized payloads are simply retained — safe, since the dedup is conditional and consumers already handle payload-present. Per-call work is now hard-bounded, so streaming CPU scales linearly with output length.

Tests: an oversized structurally-equal payload is retained; a toJSON-counting payload proves bounded work (~457 leaf visits vs 100,000 unbounded); small structurally-equal payloads still dedup. Changeset included (`@mastra/core` patch).

## Related issue(s)

Fixes #19373

## Type of change

- [x] Performance improvement

## Checklist

## ELI5

When streaming produces lots of updates, the system was repeatedly checking bigger and bigger data, using too much CPU. This change stops those checks after about 32 KB, while still removing duplicate small results.

## What changed

- Added bounded serialization for step payload comparisons.
- Retains oversized payloads instead of fully serializing them.
- Preserves deduplication for small, structurally equivalent payloads.
- Added tests for payload retention, bounded serialization, and small-payload deduplication.
- Included a `@mastra/core` patch changeset for issue `#19373`.

### DIFF
--- a/.changeset/jolly-badgers-divide.md
+++ b/.changeset/jolly-badgers-divide.md
@@ -1,0 +1,5 @@
+---
+'@mastra/core': patch
+---
+
+Fixed high CPU usage during agent streaming: the workflow engine's step payload deduplication re-serialized the accumulated streaming context on every chunk, which could pin a CPU core on long streamed responses. The comparison is now size-bounded, so streaming CPU scales linearly with output length. Fixes [#19373](https://github.com/mastra-ai/mastra/issues/19373).
--- a/packages/core/src/workflows/default.test.ts
+++ b/packages/core/src/workflows/default.test.ts
@@ -1362,6 +1362,86 @@ describe('DefaultExecutionEngine.fmtReturnValue stepExecutionPath and payload de
 
     expect(result.steps.step1.payload).toBe(circular);
   });
+
+  describe('oversized payloads (#19373)', () => {
+    // Builds a payload whose leaves count how many times they get JSON-serialized,
+    // so the tests can assert the dedup comparison does bounded work instead of
+    // serializing multi-megabyte step payloads in full on every call.
+    function makeCountingPayload(leafCount: number, counter: { count: number }) {
+      return {
+        parts: Array.from({ length: leafCount }, (_, i) => ({
+          toJSON() {
+            counter.count++;
+            return `leaf-${i}-${'x'.repeat(64)}`;
+          },
+        })),
+      };
+    }
+
+    it('should retain (not dedup) an oversized payload even when it structurally matches the previous output', async () => {
+      // ~1.4 MB of identical content on both sides — structurally equal, far past any
+      // reasonable dedup bound. The optimization exists to shrink small execution-log
+      // entries; for oversized values retaining the payload is the safe outcome.
+      const bigParts = Array.from({ length: 20_000 }, (_, i) => `part-${i}-${'x'.repeat(64)}`);
+      const stepResults: Record<string, StepResult<any, any, any, any>> = {
+        input: { parts: bigParts } as any,
+        step1: {
+          status: 'success',
+          output: { value: 2 },
+          payload: { parts: [...bigParts] },
+          startedAt: 1,
+          endedAt: 2,
+        },
+      };
+      const lastOutput: StepResult<any, any, any, any> = stepResults.step1!;
+
+      const result = await engine.fmtReturnValuePublic(pubsub, stepResults, lastOutput, undefined, ['step1']);
+
+      expect(result.steps.step1.payload).toEqual({ parts: bigParts });
+    });
+
+    it('should do bounded serialization work per step instead of stringifying huge payloads in full', async () => {
+      const counter = { count: 0 };
+      const leafCount = 50_000;
+      const stepResults: Record<string, StepResult<any, any, any, any>> = {
+        input: makeCountingPayload(leafCount, counter) as any,
+        step1: {
+          status: 'success',
+          output: { value: 2 },
+          payload: makeCountingPayload(leafCount, counter),
+          startedAt: 1,
+          endedAt: 2,
+        },
+      };
+      const lastOutput: StepResult<any, any, any, any> = stepResults.step1!;
+
+      await engine.fmtReturnValuePublic(pubsub, stepResults, lastOutput, undefined, ['step1']);
+
+      // Unbounded comparison serializes every leaf on both sides (100k+ toJSON calls).
+      // A bounded comparison aborts once the size cap is hit, so only a small prefix
+      // of each side is ever visited.
+      expect(counter.count).toBeLessThan(5_000);
+    });
+
+    it('should still dedup small structurally-equal payloads below the bound', async () => {
+      const small = { items: ['a', 'b', 'c'], nested: { value: 1 } };
+      const stepResults: Record<string, StepResult<any, any, any, any>> = {
+        input: JSON.parse(JSON.stringify(small)) as any,
+        step1: {
+          status: 'success',
+          output: { value: 2 },
+          payload: JSON.parse(JSON.stringify(small)),
+          startedAt: 1,
+          endedAt: 2,
+        },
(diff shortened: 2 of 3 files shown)


xianjianlf2:
Checked the red checks here. They are both Vercel authorization gates, not test failures:

- `Vercel – mastra-docs-1.x`
- `Vercel – mastra-playground-ui`

Both target URLs point to Vercel asking a Mastra team member to authorize the deployment. The GitHub Actions check on this PR is green, and CodeRabbit reported no actionable comments.

xianjianlf2:
This is superseded by #20340, which merged an updated approach on current main. Closing this one to avoid carrying forward the older bounded-stringify strategy and its conflicting tests.

PR #19707, opened 2026-07-18
fix(inngest,core): deliver resumeData when resuming a tool-suspended durable agent

### LINKED ISSUE #19699: [BUG] @mastra/inngest: resuming a tool-suspended durable agent re-executes the step with resumeData=undefined — run stays suspended
### Describe the Bug

On the `@mastra/inngest` durable agent (`createInngestAgent`), resuming a run that suspended inside a tool (`suspend()` / `agent.resume` / the `resume-stream` route) does not deliver `resumeData` to the suspended tool. Instead the durable workflow re-enters the same execution path, the agent LLM re-plans from scratch, the tool's `execute` is called a second time with `resumeData === undefined`, and it suspends again. The workflow snapshot stays `status: "suspended"` indefinitely — resume is effectively a no-op.

The plain (non-durable) `Agent` resumes correctly with the same tool and the same resume payload, so this is specific to the durable/Inngest execution engine.

Human-in-the-loop / approval gates (the primary reason to use suspend/resume) are therefore unusable on the durable agent.

## Root cause analysis

`persistStepUpdate` (packages/core/src/workflows/handlers/entry.ts) builds the Inngest step id as:

    const operationId = `workflow.${workflowId}.run.${runId}.path.${JSON.stringify(executionContext.executionPath)}.stepUpdate${phase ? `.${phase}` : ''}`;

#17319 added the `phase` suffix to separate the pre-exec (`start`) and post-exec persists within a single run (fix present in 1.51.0). But on RESUME, the resumed run re-enters the same `executionPath` (`[1]`) and emits the final (phase-less) `...path.[1].stepUpdate` again — colliding with the pre-suspend run's already-memoized step of the same id. Inngest's `AUTOMATIC_PARALLEL_INDEXING` fires, and the resume's `resumePayload` never reaches the suspended step: the nested `inngest:durable-agentic-execution` step re-executes the agentic loop instead of resuming the suspended tool, so `resumeData` arrives as `undefined`.

The persisted state is correct (`suspendedPaths: {"inngest:durable-agentic-execution":[1]}`, the step's `status:"suspended"` with the full agent payload) — it's the resume dispatch/propagation into the nested agentic-execution step that fails to inject `resumePayload`.

Note: this is distinct from #17319 (that was the start/end collision within one execution). This is the resume re-entering the same path collision. Reproduced with both `agent.resume(...)` and the standard `resume-stream` route.

### Steps To Reproduce

1. Create a durable agent with createInngestAgent, giving it one tool that suspends (request_approval-style: on first execute it calls suspend(); if resumeData is present it returns it). Log resumeData at the top of execute.
2. Trigger a suspend: stream a prompt that makes the agent call the tool. The stream emits `tool-call-suspended`; the snapshot persists status:"suspended", suspendedPaths:{"inngest:durable-agentic-execution":[1]}. Tool logs `resumeData=undefined` (expected).
3. Resume: agent.resume(runId, { approved: true }, { threadId, resourceId })  (also reproduced via the resume-stream route).
4. Observe: the resume call returns 200/normally, but the tool's execute is called a SECOND time with resumeData=undefined, the agent re-plans and re-suspends, and the snapshot stays status:"suspended" forever.

Worker logs on resume:

    Duplicate step ID "workflow.inngest:durable-agentic-loop.run.<runId>.path.[1].stepUpdate" detected across parallel chains
    { code: 'AUTOMATIC_PARALLEL_INDEXING', explanation: 'Using the same ID for steps in different parallel chains...', action: 'Use a unique ID for each step...' }

Baseline: the same code with a PLAIN Agent (no createInngestAgent) resumes correctly — the tool logs resumeData={"approved":true} and the run finishes.

https://gist.github.com/devangpadhiyar/21e4949e9c45a7cb23a1eed847c89d83

### Expected Behavior

On resume, the suspended tool's execute should receive resumeData = { approved: true } (the payload passed to resume), return it, and the agent should continue to completion — snapshot status:"success". This is exactly what the plain Agent does with identical code.

### Environment Information

```block
- @mastra/core: 1.51.0 (includes the #17319 phase:'start' fix)
- @mastra/inngest: 1.8.2
- inngest: 4.13.0
- Storage: @mastra/pg (PostgresStore); also reproduced with @mastra/libsql
- Runtime: Node 22
- Inngest: self-hosted Inngest Dev Server + Connect worker (connect())
- LLM Provider: Anthropic
```

### Verification

devangpadhiyar:
Fixes #19699

Resuming a durable agent (`createInngestAgent`) that suspended inside a tool didn't work: the nested agentic loop re-ran from scratch, the tool re-executed without its resume data, its memoized suspended step-update collided (`AUTOMATIC_PARALLEL_INDEXING`), and the run stayed `suspended` forever. Plain agents were fine, so it was specific to the Inngest engine.

Two small changes fix it. In `@mastra/inngest`, when resuming into a nested workflow whose parent only knows the nested workflow's id (`resume.steps` has a single entry, so `slice(1)` is empty), we now fall back to the nested snapshot's own `suspendedPaths` to figure out which inner step to resume and where. In `@mastra/core`, the resumed step's terminal `persistStepUpdate` now carries a `resume` phase so its durable operation id no longer collides with the memoized suspended write.

Before, on resume the suspended tool was re-invoked with `resumeData = undefined` and the run never finished. After:

```ts
const durableAgent = createInngestAgent({ agent, inngest });
const { runId } = await durableAgent.stream(messages, { runId }); // suspends in a tool
await durableAgent.resume(runId, { approved: true });             // tool now gets { approved: true }, run completes
```

Added an end-to-end regression test in `workflows/inngest` that drives a real durable agent through suspend-in-tool then resume against the Inngest dev server, asserting the tool receives the payload, emits a tool result (no re-suspend), and the agent finishes.

Verified locally: the new test passes; the core workflow suite (284), `persist-step-update` unit tests (15), durable-agent suspension domain (16), and `resume-api` (29) stay green. One pre-existing `index.test.ts` case ("basic suspend and resume flow with async await syntax with state") is flaky in my single-dev-server setup and fails identically on unpatched `main`, so it's unrelated to this change.

## ELI5

When a tool pauses an agent, the agent can now continue correctly after approval. The fix preserves the approval data and prevents the agent from pausing again.

## What changed

- Updated nested workflow resumption in `@mastra/inngest` to use the nested snapshot’s `suspendedPaths` when needed.
- Added a `resume` phase to resumed `persistStepUpdate` operations in `@mastra/core`.
- Prevented durable operation ID collisions with memoized suspended updates.
- Added unit and end-to-end regression tests for suspend-in-tool and resume behavior.
- Verified that resumed tools receive `resumeData`, emit a `tool-result`, reach `finish`, and produce `All done.`.


abhiaiyer91:
Closing, the factory will handle it

PR #19724, opened 2026-07-19
feat(mastracode): show transient connection retries

### LINKED ISSUE #19722: [FEATURE] Show Mastra Code retry progress for dropped provider connections
## Problem Statement

Mastra Code automatically retries some transient provider stream failures, but the retries are currently invisible. After the retry budget is exhausted, users only see the final error with no indication that earlier attempts occurred. This makes errors such as `write EPIPE`, `other side closed`, `ECONNRESET`, and `socket hang up` look like a single immediate failure and encourages users to manually send `continue`.

The current two-retry budget is also too small for intermittent provider connection failures. Increasing it to 10 aligns Mastra Code with the retry budget used by Claude Code and gives unstable connections more time to recover.

The generic provider retry policy can also take precedence over Mastra Code's explicit transient-connection policy. That prevents the application-specific retry budget and status reporting from being applied to provider errors carrying retryable metadata.

## Proposed Solution

Allow explicit `StreamErrorRetryProcessor` matcher policies to take precedence over generic provider retry metadata and add a matcher-level retry callback that runs before the retry delay.

Use that callback in Mastra Code to:

- Retry recognized dropped-connection failures up to 10 times.
- Use exponential backoff beginning at 500ms and capped at 30 seconds.
- Emit retry attempt, maximum retry count, and delay metadata to the TUI.
- Render progress such as `(retry 1/10 in 0.5s)` alongside the transient error.

## Component

CLI

## Alternatives Considered

- Keep retries silent. This makes successful recovery difficult to distinguish from a stalled or terminal run.
- Rely only on generic provider retry metadata. This limits Mastra Code's ability to apply a connection-specific retry budget and UI feedback.
- Add a Mastra Code-only wrapper around provider streams. This duplicates retry ownership and does not provide a reusable processor callback for other applications.

## Example Use Case

As a developer using Mastra Code with OpenAI Responses, I want a transient socket failure to show that Mastra Code is retrying, including the current attempt and delay, so I can let the run recover instead of interrupting it or manually prompting `continue`.

## Additional Context

A live trace of intermittent OpenAI failures showed seven affected attempts ending with `step-start` followed by an error before any response metadata, content, tool call, or finish event. Each incomplete attempt was discarded and followed by a successful retry, confirming these were pre-completion transport failures rather than harmless socket closure after a completed response.

The implementation includes focused processor and SDK tests plus a deterministic real-TUI scenario against the OpenAI Responses endpoint.

## Verification

iulspop:
## Description

Mastra Code previously retried dropped provider connections only twice and did not display those attempts. If both retries failed, users saw only the final error with no indication that recovery had already been attempted.

This change displays each retry with its attempt number and delay, so the final error has clear retry history. It also increases the transient-connection budget from 2 to 10 retries, matching Claude Code's retry budget, with exponential backoff from 500ms capped at 30 seconds.

Explicit stream-error matchers now override generic provider retry metadata, and a matcher-level callback reports progress such as `(retry 1/10 in 0.5s)` in the TUI.

Runtime investigation confirmed the reproduced OpenAI `other side closed` failures occurred before response metadata, content, tool calls, or a finish event. Incomplete attempts were discarded before successful retries.

Verification:

- Core processor tests: 61 passed
- Full Core suite: 12,160 passed, 110 skipped
- Code SDK suite: 1,012 passed, 2 skipped
- Real-TUI stream retry scenario: 1 passed
- Mastra Code build: 29/29 tasks passed
- Core, SDK, and TUI typechecks and lint passed; Core lint reports one unrelated existing warning
- Full TUI suite has three unrelated failures in `shell-output.test.ts`, reproducible in isolation

## Related issue(s)

Closes #19722

## Type of change

- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [x] Test update

## Checklist


iulspop:
tested this locally and i think it looks good!
image

abhiaiyer91 (approved):

iulspop:
Follow-up: #20388 adds configured retries for transient provider server errors and ensures retry timing is rendered only when a retry is actually scheduled. Tracks #20387.

PR #19728, opened 2026-07-20
fix(deployer): resolve Missing "." specifier for subpath-only workspace exports

### LINKED ISSUE #19379: [BUG] mastra build with externals: true fails with Missing "." specifier for workspace packages that only declare subpath exports
### Describe the Bug

In a pnpm monorepo, `mastra build` crashes during the analyze step when both of these are true:

1. `bundler: { externals: true }` is set — which is currently the only way to ship native `.node` addons (realtime voice SDKs, sqlite drivers, sharp, ...), and
2. any **transitive** workspace dependency declares an `exports` map with only subpath entries and no `"."` entry — a common convention for internal utility packages:

```jsonc
// packages/leaf/package.json
{
  "name": "@scope/leaf",
  "exports": {
    "./sub/*": "./src/*.ts"
  }
}
```

The build fails with:

```
Failed to analyze: Missing "." specifier in "@scope/leaf" package
```

even though no code ever imports the package root.

**Root cause chain** (paths in `packages/deployer` on current `main`):

1. The transitive workspace walk in `src/build/analyze/analyzeEntry.ts` (`checkTransitiveDependencies`) registers every transitive workspace dependency as a root-level virtual module (`exports: ['*']`) — even when the app only ever imports subpaths of it.
2. The `alias-optimized-deps` plugin in `src/build/analyze/bundleExternals.ts` (~line 224) then resolves that root entry via `resolve.exports(pkgJson, '.')` with no try/catch. `resolve.exports` **throws** `Missing "." specifier` when an `exports` map exists but has no `"."` entry, so the `pkgJson.main ?? 'index.js'` fallback on the very next line is unreachable for exactly this case.
3. The error names only the leaf package — not which package imported it — so in a large workspace you end up bisecting the dependency graph by hand to find the chain.

Two notes for whoever picks this up:

- `resolve.exports`' `unsafe: true` option is **not** an escape hatch: in resolve.exports 2.0.3 the throw on a missing entry is unconditional (`unsafe` only skips the condition defaults).
- This only fires with `externals: true` or in dev (`noBundling` mode), and likely surfaced when #18879 (merged 2026-07-06) extended what the transitive walk covers.

Fixing this lets monorepo teams keep internal packages exactly as they are — subpath-only exports maps are a deliberate, widespread pattern — while deploying Mastra apps that need `externals: true` for native modules. Today the workaround is adding stub `"."` entries (pointing at an empty `src/index.ts`) to every affected package, one build failure at a time.

### Steps To Reproduce

1. Create a pnpm workspace with a Mastra app and two workspace packages, so the failing one is transitive: app depends on `@scope/mid`, `@scope/mid` depends on `@scope/leaf`.
2. Give `@scope/leaf` an exports map with only subpaths (see JSON above) and import only a subpath from `@scope/mid`: `import { foo } from '@scope/leaf/sub/foo'`.
3. Set `bundler: { externals: true }` in the `Mastra` config.
4. Run `mastra build` — it fails with `Failed to analyze: Missing "." specifier in "@scope/leaf" package`.
5. Confirm the trigger: either removing `externals: true` or adding a `"."` entry to the leaf's exports map makes the build pass.

The setup above is complete and minimal. I'm preparing a fix PR that includes an e2e fixture in `e2e-tests/monorepo` reproducing this exact shape; happy to push a standalone repro repo as well if that's preferred.

### Expected Behavior

- `mastra build` with `externals: true` succeeds when a workspace package only declares subpath exports: a package whose root is never imported should not need a root export. The walk could skip registering the root virtual module when the package has no root entry (the subpaths actually imported are already captured per-specifier), and/or the resolver could guard the `resolve.exports` call and fall back to `main`/`index.js`.
- When a package genuinely cannot be resolved, the error should include the importer chain (`app -> @scope/mid -> @scope/leaf`) so it's actionable in a large workspace.

### Environment Information

```
@mastra/deployer: current main (2026-07-13); also reproduced on 0.0.0-livekit-beta-v3-20260709031728
Node: 20.x
OS: macOS 15 / Linux (both)
```

### Verification

rajarshidattapy:
## Type of change

  - [ ] Documentation update
  - [ ] Code refactoring
  - [ ] Performance improvement
  - [ ] Test update

  ## Checklist

## Description

`mastra build` with `bundler: { externals: true }` crashed during analysis with `Failed to analyze: Missing "." specifier in "@scope/leaf" package` whenever a transitive workspace dependency declared an `exports` map with no `"."` entry — a common convention for internal utility packages — even though nothing imported the package root.

Fixes both links in the chain the issue identified:

**1. `analyzeEntry.ts` — root cause.** `checkTransitiveDependencies` registered every transitive workspace dependency as a root-level `exports: ['*']` virtual module, including packages only ever reached through a subpath. It now skips packages that have no root export. The subpaths actually imported are already keyed separately in `depsToOptimize` by full specifier (`@scope/leaf/sub/foo`), so the root entry was pure redundancy that happened to be fatal — nothing is lost by dropping it.

**2. `bundleExternals.ts` — latent bug.** The `resolve.exports(pkgJson, '.')` call in the `alias-optimized-deps` plugin had no try/catch, which made the `pkgJson.main ?? 'index.js'` fallback on the very next line unreachable. `resolve.exports` throws rather than returning `undefined` for a missing entry, so the fallback the author clearly intended never ran. Now guarded, with a comment recording that `unsafe: true` is *not* an escape hatch here — the throw is unconditional in resolve.exports 2.0.3, that option only skips condition defaults.

**Supporting changes:** `hasRootExport()` in `utils.ts` implements Node's rule for disambiguating a subpath map (`{"./sub/*": ...}`) from a conditions map (`{"import": ..., "require": ...}`, which does apply to the root). `WorkspacePackageInfo` now carries `exports`/`main` — optional so existing fixtures compile unchanged, and free at runtime since `workspace.package` is the already-parsed manifest, which keeps the walk synchronous.

Internal packages can now keep subpath-only exports maps exactly as they are, instead of needing stub `"."` entries pointing at empty files, discovered one build failure at a time.

## Related issue(s)

Closes #19379

## Type of change

- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [ ] Test update

## Checklist

- [ ] I have made corresponding changes to the documentation (if applicable) — n/a, no user-facing API change

## Testing

- **New regression test** in `analyzeEntry.test.ts` reproducing the issue's exact shape (app → `@internal/a` → `@internal/shared`, where the leaf declares subpath-only exports and is never root-imported), plus a companion asserting packages *with* a `"."` entry are still registered.
- **Red/green verified**: stashed the fix and confirmed the regression test fails with exactly the reported behavior, then passes with it restored — the test pins the bug rather than merely passing alongside it.
- **`hasRootExport` unit matrix** in `utils.test.ts`: no exports field, string/array exports, conditions-only map, subpath map with and without `"."`, empty map, and `main` alongside a subpath-only map (which does *not* resurrect the root, since an exports map fully replaces `main`).
- 146 tests pass across `utils`, `analyzeEntry`, `bundleExternals`, `bundlerOptions`. `tsc --noEmit` clean.

## Not included

- **The `e2e-tests/monorepo` fixture.** The issue reporter mentioned they were preparing one; separately, my environment cannot create symlinks (`EPERM`), which is how those fixtures construct workspaces, so I could not run one I wrote. Happy to add it if a maintainer can verify it, or if CI is the intended proving ground.
- **Importer-chain diagnostics** (`app -> @scope/mid -> @scope/leaf`), the issue's second "Expected Behavior" bullet. That needs provenance threaded through `DependencyMetadata` and is a distinct enhancement; with this fix the error no longer escapes at all. Suggest a follow-up issue rather than bundling speculative plumbing into a bug fix — happy to do it here instead if preferred.

## ELI5

Some workspace packages only work through specific paths, not from their package root. This change stops the build from incorrectly treating them as root-importable and adds a fallback so `mastra build` no longer crashes.

## Summary

- Skips root-level registration for workspace packages with subpath-only `exports`.
- Safely handles failed `resolve.exports` lookups so `main` or `index.js` fallback resolution can run.
- Adds `hasRootExport` to detect whether a package supports root imports.
- Stores `exports` and `main` metadata in `WorkspacePackageInfo`.
- Adds regression and unit tests for subpath-only and root exports.
- Adds a Changesets patch release entry for `@mastra/deployer`.
- Verified with 146 passing tests and successful `tsc --noEmit`.

### DIFF
--- a/.changeset/subpath-only-workspace-exports.md
+++ b/.changeset/subpath-only-workspace-exports.md
@@ -1,0 +1,21 @@
+---
+'@mastra/deployer': patch
+---
+
+Fixed `mastra build` failing with `Missing "." specifier` when a workspace package declares only subpath exports.
+
+In a monorepo, a build using `bundler: { externals: true }` crashed during analysis if any transitive workspace dependency declared an `exports` map with no `"."` entry — a common convention for internal utility packages:
+
+```jsonc
+// packages/leaf/package.json
+{
+  "name": "@scope/leaf",
+  "exports": {
+    "./sub/*": "./src/*.ts"
+  }
+}
+```
+
+The build failed with `Failed to analyze: Missing "." specifier in "@scope/leaf" package` even though no code imported the package root.
+
+The transitive workspace walk was registering every workspace dependency as a root-level import, including packages only ever used through a subpath. The subpaths that are actually imported were already tracked separately, so the redundant root entry is now skipped for packages that have no root export. Internal packages can keep subpath-only exports maps as they are, instead of needing stub `"."` entries pointing at empty files.
--- a/packages/deployer/src/build/analyze/analyzeEntry.test.ts
+++ b/packages/deployer/src/build/analyze/analyzeEntry.test.ts
@@ -360,6 +360,103 @@ describe('analyzeEntry', () => {
     expect(analyzeCache.size).toBe(1);
   });
 
+  it('should not register a root export for transitive workspace packages that only declare subpath exports', async () => {
+    const root = join(import.meta.dirname, '__fixtures__', 'nested-workspace');
+    const entryFilePath = join(root, 'apps', 'mastra', 'src', 'shared-transitive.ts');
+    vi.spyOn(process, 'cwd').mockReturnValue(join(root, 'apps', 'mastra'));
+
+    const workspaceMap = new Map<string, WorkspacePackageInfo>([
+      [
+        '@internal/a',
+        {
+          location: `${root}/packages/a`,
+          dependencies: {
+            '@internal/shared': '1.0.0',
+          },
+          version: '1.0.0',
+        },
+      ],
+      [
+        '@internal/b',
+        {
+          location: `${root}/packages/b`,
+          dependencies: {
+            '@internal/shared': '1.0.0',
+          },
+          version: '1.0.0',
+        },
+      ],
+      [
+        '@internal/shared',
+        {
+          location: `${root}/packages/shared`,
+          dependencies: {},
+          version: '1.0.0',
+          // Only subpath entries, no "." — asking a resolver for the root here throws.
+          exports: {
+            './sub/*': './src/*.ts',
+          },
+        },
+      ],
+    ]);
+
+    const result = await analyzeEntry({ entry: entryFilePath, isVirtualFile: false }, '', {
+      shouldCheckTransitiveDependencies: true,
+      logger: noopLogger,
+      sourcemapEnabled: false,
+      workspaceMap,
+      projectRoot: root,
+    });
+
+    // The packages actually imported at their root are still registered...
+    expect(result.dependencies.get('@internal/a')?.exports).toEqual(['a']);
+    expect(result.dependencies.get('@internal/b')?.exports).toEqual(['b']);
+    // ...but the subpath-only package gets no root entry, so nothing asks for its ".".
+    expect(result.dependencies.has('@internal/shared')).toBe(false);
+    expect(result.dependencies.size).toBe(2);
+  });
+
+  it('should still register transitive workspace packages whose exports map has a root entry', async () => {
+    const root = join(import.meta.dirname, '__fixtures__', 'nested-workspace');
+    const entryFilePath = join(root, 'apps', 'mastra', 'src', 'shared-transitive.ts');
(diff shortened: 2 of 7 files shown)


wardpeet:
Thank you for opening this pull request (PR). The deployer logic is quite complex due to the many edge cases involved. For now, I will be closing this PR, but I have created another one to address this issue: https://github.com/mastra-ai/mastra/pull/19808.

Issue #19725, opened 2026-07-19
[BUG] A replayed `run-registered` with no terminal event permanently wedges `agent.stream()` — `#waitForRemoteRunToFinish` has no timeout and ignores the thread lease, and `abortRun` is a no-op for remote runs


Bickio:
### Summary

If a server process dies mid-run (SIGTERM/OOM after publishing `run-registered` but before any terminal event lands on the thread topic), that thread becomes **permanently unable to accept new turns** when using `RedisStreamsPubSub`:

1. The retained stream now contains a `run-registered` with no matching `run-completed` / `run-failed` / `run-aborted`.
2. Every fresh `subscribeToThread` replays it and applies it to the pubsub-shared runtime state — `state.activeThreadRunIds.set(key, data.runId)` (`thread-stream-runtime.ts` ~line 1392). This happens even if that subscriber's stream later errors; the state mutation has already landed. Replay re-poisons the state on every subscribe, so restarting the process (or scaling to a single instance) does not help.
3. The next `agent.stream()` / `generate()` on the thread persists the user message, then enters `waitForCrossAgentThreadRun` (~line 1239). The corpse runId has no local record, so it falls through to `#waitForRemoteRunToFinish` (~line 1268), which subscribes to the topic and waits for a terminal event for that runId — **with no timeout and no liveness check**. It never arrives. The call parks forever.

The failure is completely silent: the user's message is persisted, no run starts, no model call is made, nothing is logged. From the outside the agent just stops answering that one thread.

### The lease exists but is never consulted

The runtime has a cross-process thread lease (`LeaseProvider`, acquired per run, TTL'd, released on completion/abort). A run whose owner died has an expired (or never-renewed) lease — exactly the signal needed to declare it dead. But `waitForCrossAgentThreadRun` / `#waitForRemoteRunToFinish` only wait for a terminal *event*:

```ts
async #waitForRemoteRunToFinish(pubsub, key, runId) {
  const topic = this.#threadTopic(key);
  await new Promise(resolve => {
    const onEvent = event => {
      const data = event.data;
      if ((data?.type === 'run-completed' || data?.type === 'run-aborted' || data?.type === 'run-failed')
          && data.runId === runId) { ...resolve(); }
    };
    void resolvedPubSub.subscribe(topic, onEvent).catch(() => resolve());
  });
}
```

So an expired lease changes nothing — the wait is unbounded.

### There is no API escape hatch either

[...]

Issue #19723, opened 2026-07-19
[BUG] `subscribeToThread` replay is divergent across instances on Redis Streams — sourceId self-suppression drops completed runs asymmetrically, and unpersisted runs replay as phantom messages


Bickio:
### Summary

Run two or more Mastra server processes sharing a `RedisStreamsPubSub`, and `agent.subscribeToThread()` returns a **different replay backlog depending on which process handles the subscribe**. Behind a load balancer, a client reconnecting to the same thread flip-flops between two disjoint views of history. A second effect of the same mechanism: runs that never persisted to storage (mid-stream model errors, aborts) re-stream their partial content to every fresh subscriber served by a non-origin process — phantom messages that reappear on every reconnect.

A single-process deployment fully masks the bug, which is likely why it hasn't been reported: the in-memory `EventEmitterPubSub` has no replay at all, and one process suppresses *all* of its own history (so steady-state reconnects replay nothing).

### The two interacting code sites

1. **Replay-from-0 is hardwired** — `pubsub/redis-streams/src/index.ts` (~line 305 at `a440c311c`): `subscribe()` without `options.group` creates a fresh private fan-out group anchored at stream id `'0'`:

   ```ts
   const group = options?.group ?? `__fanout-${randomUUID()}`;
   await this.#writeClient.xGroupCreate(streamKey, group, '0', { MKSTREAM: true });
   ```

   So every fresh subscriber replays the topic's full retained history (MAXLEN-bounded, ~10k events). Documented fan-out behavior on its own.

2. **sourceId self-suppression is asymmetric on replay** — `packages/core/src/agent/thread-stream-runtime.ts`. The runtime stamps every `stream-part` it publishes with a per-runtime-instance id (`#getSourceId()` — a `randomUUID()` per process), and its `subscribeToThread` handler drops incoming parts from itself (line ~1401, same for `signal-enqueued` at ~1418):

   ```ts
   if (data.sourceId === this.#id) return;
   ```

[...]

PR #19729, opened 2026-07-20
fix: handle both usage paths in convertMastraChunkToAISDKBase for persisted signal chunks

### LINKED ISSUE #19721: [BUG] `convertMastraChunkToAISDKBase` throws `TypeError` on `finish` chunks synthesized by `#broadcastPersistedSignal` (`payload.output.usage` vs `payload.usage`)
### Describe the Bug

`@mastra/core`'s thread-stream runtime broadcasts persisted signals to thread subscribers as a synthetic `start` / data-part / `finish` sequence. The synthesized `finish` part puts `usage` directly on `payload`, but `@mastra/ai-sdk`'s chunk converter reads `chunk.payload.output.usage`. The result is that **any `subscribeToThread` stream piped through the AI SDK converter throws as soon as a persisted-signal broadcast passes through it**, killing the subscription.

```
TypeError: Cannot read properties of undefined (reading 'usage')
```

In an SSE route this surfaces as `controller.error(...)` → the response terminates with zero bytes (HTTP/2 RST; Chrome reports "Error in input stream").

With `RedisStreamsPubSub` as the pubsub this is not transient: fresh subscribes replay retained topic history (fan-out consumer groups are created at stream id `0`), so once a signal broadcast is in a thread's stream, **every future subscribe to that thread crashes deterministically** until the stream key is deleted. Client reconnect loops then hammer the server (we saw a sustained 1 error/sec in production).

### The two mismatched code sites

Producer — `packages/core/src/agent/thread-stream-runtime.ts` (`#broadcastPersistedSignal`, ~line 731 at `a440c311c`) synthesizes:

```ts
const parts: any[] = [
  { type: 'start', runId },
  { ...signal.toDataPart(), runId },
  {
    type: 'finish',
    runId,
    payload: {
      stepResult: { reason: 'stop' },
      usage: { inputTokens: 0, outputTokens: 0, totalTokens: 0 },  // <-- payload.usage
    },
  },
];
```

Consumer — `client-sdks/ai-sdk/src/helpers.ts` (`convertMastraChunkToAISDKBase`, `finish` case, ~line 178):

```ts
case 'finish': {
  return {
    type: 'finish',
    finishReason: normalizeFinishReason(chunk.payload.stepResult.reason),
    ...
    totalUsage: normalizeUsage(chunk.payload.output.usage),  // <-- payload.output is undefined
  };
}
```

A real model run's `finish` chunk carries the full `FinishPayload` (`payload.output.usage`, `payload.metadata`, …); the synthesized signal part carries neither `output` nor `metadata`. Note the `step-finish` case (~line 331) has the same reads (`chunk.payload.output.usage`, destructuring `chunk.payload.metadata`), so any other producer of abbreviated step-finish parts would hit the same class of crash.

### Steps To Reproduce

Minimal, converter-level:

```ts
import { convertMastraChunkToAISDKBase } from '@mastra/ai-sdk'; // or via toAISdkStream

convertMastraChunkToAISDKBase({
  type: 'finish',
  runId: 'r1',
  payload: {
    stepResult: { reason: 'stop' },
    usage: { inputTokens: 0, outputTokens: 0, totalTokens: 0 },
  },
} as any);
// TypeError: Cannot read properties of undefined (reading 'usage')
```

End-to-end: give an agent a signal configuration that persists an incoming signal on an idle thread without waking the agent, open `agent.subscribeToThread(...)` and pipe it through the AI SDK stream conversion, then deliver the signal — the subscription stream errors on the synthesized `finish` part. (The broadcast is served both to live local subscribers and, on Redis Streams, to every later subscriber via replay.)

N/A See converter-level repro above

### Expected Behavior

Persisted-signal broadcasts should convert cleanly (or be skippable) — a feature of core (`sendSignal` / persisted signals) should not produce chunks that core's own first-party converter crashes on.

### Environment Information

```block
System:
    OS: macOS 15.4.1
    CPU: (10) arm64 Apple M4
    Memory: 111.56 MB / 32.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 22.19.0 - /Users/bickio/.nvm/versions/node/v22.19.0/bin/node
    npm: 10.9.3 - /Users/bickio/.nvm/versions/node/v22.19.0/bin/npm
    pnpm: 11.13.0 - /Users/bickio/.nvm/versions/node/v22.19.0/bin/pnpm
  Browsers:
    Chrome: 150.0.7871.124
    Firefox: 152.0.5
    Safari: 26.1
  npmGlobalPackages:
    @google/gemini-cli: 0.16.0
    @zendesk/zcli: 1.0.0-beta.54
    corepack: 0.34.0
    mastracode: 0.14.0
    npm: 10.9.3
```

### Verification

PiedPiper911:
## Summary

Fixes #19721

This PR resolves a `TypeError: Cannot read properties of undefined (reading 'usage')` that crashes any `subscribeToThread` stream piped through the AI SDK converter when a persisted-signal broadcast passes through it.

The fix makes usage path resolution defensive in both the `finish` and `step-finish` cases of `convertMastraChunkToAISDKBase`, so that both normal model-run chunks and synthesized persisted-signal chunks are handled correctly.

## Root Cause Analysis

There are two distinct sources of `finish` chunks that flow through `convertMastraChunkToAISDKBase`:

1. **Normal model runs** — The agent runtime emits a full `FinishPayload` where usage lives at `chunk.payload.output.usage` (nested inside the `output` object alongside `metadata`, etc.).

2. **Persisted-signal broadcasts** — `#broadcastPersistedSignal` in `packages/core/src/agent/thread-stream-runtime.ts` synthesizes a minimal `start` / data-part / `finish` sequence to notify thread subscribers. The synthesized `finish` part places `usage` directly at `chunk.payload.usage` — there is no `output` wrapper:

```ts
// thread-stream-runtime.ts (~line 731)
{
  type: 'finish',
  runId,
  payload: {
    stepResult: { reason: 'stop' },
    usage: { inputTokens: 0, outputTokens: 0, totalTokens: 0 },  // <-- payload.usage
  },
}
```

The converter unconditionally reads `chunk.payload.output.usage`, so when `payload.output` is `undefined` (as in the synthesized chunk), it throws:

```
TypeError: Cannot read properties of undefined (reading 'usage')
```

With `RedisStreamsPubSub`, this is especially severe: retained topic history replays the offending chunk to every future subscriber, causing a deterministic crash loop.

## Fix

Applied defensive path resolution in two cases within `convertMastraChunkToAISDKBase`:

### `finish` case
```ts
// Before
totalUsage: normalizeUsage(chunk.payload.output.usage),

// After
totalUsage: normalizeUsage(chunk.payload.output?.usage ?? chunk.payload.usage),
```

### `step-finish` case
```ts
// Before
const { request: _request, providerMetadata, ...rest } = chunk.payload.metadata;
usage: normalizeUsage(chunk.payload.output.usage),

// After
const { request: _request, providerMetadata, ...rest } = chunk.payload.metadata ?? {};
usage: normalizeUsage(chunk.payload.output?.usage ?? chunk.payload.usage),
```

The `?? {}` guard on `chunk.payload.metadata` in `step-finish` prevents the same class of crash if a synthesized chunk omits metadata entirely.

**Existing behavior is preserved**: for normal model-run chunks, `chunk.payload.output` is defined, so `chunk.payload.output?.usage` resolves as before. The fallback to `chunk.payload.usage` only activates when `output` is absent.

## Files Changed

- `client-sdks/ai-sdk/src/helpers.ts` — Defensive usage path resolution in `finish` and `step-finish` cases of `convertMastraChunkToAISDKBase`

## Verification

The fix can be verified with the minimal converter-level repro from the issue:

```ts
import { convertMastraChunkToAISDKBase } from '@mastra/ai-sdk';

// Previously threw TypeError, now returns a valid finish chunk
const result = convertMastraChunkToAISDKBase({
  chunk: {
    type: 'finish',
    runId: 'r1',
    payload: {
      stepResult: { reason: 'stop' },
      usage: { inputTokens: 0, outputTokens: 0, totalTokens: 0 },
    },
  } as any,
  normalizeUsage: (u: any) => u,
  normalizeFinishReason: (r: any) => r,
  normalizeWarnings: (w: any) => w || [],
});

console.log(result);
// { type: 'finish', finishReason: 'stop', totalUsage: { inputTokens: 0, outputTokens: 0, totalTokens: 0 } }
```

End-to-end: deliver a persisted signal on an idle thread with `subscribeToThread` piped through the AI SDK stream conversion — the subscription no longer errors on the synthesized `finish` part.

---

Happy to adjust the approach if maintainers prefer a different strategy (e.g., normalizing the synthesized chunk shape at the producer side in `thread-stream-runtime.ts`). The defensive consumer-side fix was chosen here for minimal blast radius and backwards compatibility with any existing streams that already contain these chunks.

## ELI5

Some saved stream messages have a slightly different shape than live messages, which could crash the stream. This change teaches the converter to handle both shapes safely so subscribers keep receiving updates.

## Changes

- Resolve usage data from either the nested output or top-level payload for `finish` and `step-finish` chunks.
- Default missing `step-finish` metadata to an empty object.
- Preserve existing behavior for normal model-run chunks while supporting persisted-signal broadcasts.

### DIFF
--- a/client-sdks/ai-sdk/src/helpers.ts
+++ b/client-sdks/ai-sdk/src/helpers.ts
@@ -171,11 +171,14 @@ export function convertMastraChunkToAISDKBase<OUTPUT = undefined>({
       };
 
     case 'finish': {
+      // Normal model runs carry usage at payload.output.usage (full FinishPayload).
+      // Synthesized finish chunks from #broadcastPersistedSignal carry usage directly
+      // at payload.usage without an output wrapper. Resolve defensively to support both.
       return {
         type: 'finish',
         finishReason: normalizeFinishReason(chunk.payload.stepResult.reason) as FinishReason,
         ...(includeRawFinishReason ? { rawFinishReason: chunk.payload.stepResult.reason } : {}),
-        totalUsage: normalizeUsage(chunk.payload.output.usage),
+        totalUsage: normalizeUsage(chunk.payload.output?.usage ?? chunk.payload.usage),
       };
     }
     case 'reasoning-start':
@@ -319,7 +322,9 @@ export function convertMastraChunkToAISDKBase<OUTPUT = undefined>({
         providerMetadata: chunk.payload.providerMetadata,
       };
     case 'step-finish': {
-      const { request: _request, providerMetadata, ...rest } = chunk.payload.metadata;
+      // Defensive: synthesized chunks (e.g. from #broadcastPersistedSignal) may omit
+      // payload.metadata and carry usage at payload.usage instead of payload.output.usage.
+      const { request: _request, providerMetadata, ...rest } = chunk.payload.metadata ?? {};
       return {
         type: 'finish-step',
         response: {
@@ -328,7 +333,7 @@ export function convertMastraChunkToAISDKBase<OUTPUT = undefined>({
           modelId: (rest.modelId as string) || '',
           ...rest,
         },
-        usage: normalizeUsage(chunk.payload.output.usage),
+        usage: normalizeUsage(chunk.payload.output?.usage ?? chunk.payload.usage),
         finishReason: normalizeFinishReason(chunk.payload.stepResult.reason) as FinishReason,
         ...(includeRawFinishReason ? { rawFinishReason: chunk.payload.stepResult.reason } : {}),
         providerMetadata,


PiedPiper911:
Hi @abhiaiyer91! Following up on this streaming usage fix. I see the changeset bot flagged a missing changeset - I can add one if you would like. Also happy to address any CodeRabbit feedback. Thanks!

CalebBarnes:
Thanks for this PR and the detailed root cause analysis, and sorry for the slow turnaround here. Your diagnosis was exactly right: persisted-signal broadcasts place `usage` at `chunk.payload.usage` while normal model runs nest it under `chunk.payload.output.usage`. Our review also verified your patch resolved the crash end-to-end.

In the meantime a larger fix for the multi-instance thread-subscription issues (#19721, #19723, #19725) was in flight and just merged in #19806. It includes the same defensive usage-path handling in `convertMastraChunkToAISDKBase` along with unit tests for both chunk shapes and cross-process integration tests, so this change is now covered on main and will go out in the next release.

Closing this as superseded by #19806. We appreciate the contribution and would be glad to see more from you.

PR #19735, opened 2026-07-20
fix(memory): exclude context messages from observational memory

### LINKED ISSUE #19730: [BUG] Observational memory buffering persists ephemeral context-option messages to storage as durable user messages
### Summary

Messages passed to `agent.stream()` / `generate()` via the **`context` option** are supposed to be ephemeral, per-run input: `MessageList` tracks them under the `context` source, and the normal save path (`drainUnsavedMessages`) deliberately excludes them, so they are never written to storage.

Observational memory's buffering pipeline breaks that contract. When an OM buffering cycle fires, it collects its candidate window from **`messageList.get.all.db()`** — which includes `context`-sourced messages — seals the candidates, and persists them via `Memory.persistMessages`, a raw upsert whose only exclusion is `role === 'system'` ([`packages/memory/src/index.ts:1657-1660`](https://github.com/mastra-ai/mastra/blob/main/packages/memory/src/index.ts)). Since context messages are typically `role: 'user'`, they get written to the messages table as **durable user messages** with a server-generated UUID.

From then on the leaked message is indistinguishable from a real user turn: it comes back from `listMessages`/recall on every later run, and renders as a user bubble in any UI that hydrates from stored history.

### Evidence from my testing

Stored rows for the affected turn (Postgres, `mastra_messages`):

| id | role | content (head) | createdAt |
|---|---|---|---|
| `r0X2Ngcc…` (client id) | user | "Suggest a collar" | 03:50:37.531Z |
| `74296514-…` (UUID) | user | `<client-context>\n{"page":{"url":…}}\n</client-context>` | 03:50:37.546Z |
| `9a24a87b-…` | assistant | om-status, reasoning, text, tool-invocation, step-start, **`data-om-buffering-start`** | 03:50:37.603Z |
| `b22a296e-…` | assistant | om-status, reasoning, text, tool-invocation, … | 03:50:43.452Z |

Fingerprints tying it to OM buffering: the leaked row is `sealed: true`; the `data-om-buffering-start` marker sits exactly at the step boundary where the leak happened; and the same seal split the assistant output into two messages (the post-tool-call step can't merge into the sealed step-1 message) — earlier turns on the same thread each produced a single assistant row and no context row, despite sending the same context attachment every turn.

### Root cause

All of OM's buffering/observation entry points derive their message window from `messageList.get.all.db()`, which includes `context`-sourced messages, and none of the persist paths filter them (only `role === 'system'` is excluded). Sites on `main` that can persist a context message:

1. **Step-boundary deferred check** — [`observation-turn/step.ts:145-180`](https://github.com/mastra-ai/mastra/blob/main/packages/memory/src/processors/observational-memory/observation-turn/step.ts): `allMessages = messageList.get.all.db()` → candidates → `sealMessagesForBuffering(candidates)` → `this.turn.memory.persistMessages(candidates)` → `removeByIds` + re-add as `'memory'`. *(Our incident.)*
2. **Sync observation seal** — `step.ts:389-413`: picks the newest observed message from `get.all.db()`, seals and persists it — can select a context message.
3. **Async buffer cycle** — [`observational-memory.ts:2037-2114`](https://github.com/mastra-ai/mastra/blob/main/packages/memory/src/processors/observational-memory/observational-memory.ts) (`runAsyncBufferedObservation`): seals `messagesToBuffer` and persists via `messageHistory.persistMessages`.
4. **`turn.end()` idle buffering** — [`observation-turn/turn.ts:204-216`](https://github.com/mastra-ai/mastra/blob/main/packages/memory/src/processors/observational-memory/observation-turn/turn.ts): feeds `get.all.db()`-derived unobserved messages into `om.buffer(...)`, reaching site 3. (Note `end()`'s own direct save at `turn.ts:191-195` is correct — it uses `get.input.db()` + `get.response.db()`.)

Additionally `om.getStatus(...)` counts `get.all.db()` tokens, so context messages participate in threshold accounting — relevant to whichever fix you choose.

The intended contract lives in core: `MessageStateManager.addToSource` routes `'context'` into `userContextMessages`, and `drainUnsavedMessages` only drains `newUserMessages`/`newResponseMessages` — i.e. the persistence layer already treats context messages as never-persist. OM bypasses that by reading `all.db()` and saving directly.

### Suggested fix

Two plausible options:

- **(a) Exclude `context`-sourced messages from OM entirely** (window, sealing, persistence, and token accounting). Consistent with how OM already treats system messages and with core's persistence contract: `context` is per-run ephemeral input. `MessageList.makeMessageSourceChecker()` (or `MessageStateManager.isContextMessage`) provides the seam. This is our preferred option.
- **(b) Observe context messages but never persist them.** Keeps page/app state visible to the Observer, but needs care: the deferred check's `persistMessages → removeByIds → re-add as 'memory'` rotation assumes candidates are durable — removing an unpersisted context message would silently drop it from the live prompt mid-run.

### Workaround

We currently wrap our storage's memory-domain `saveMessages` to drop messages matching our context-attachment wire shape, so OM buffering can't make them durable.

### Steps To Reproduce

1. Configure an agent with memory + `observationalMemory` (defaults; a low `observation.bufferTokens` makes it trip faster).
2. On every chat turn, pass a per-turn context message:
   ```ts
   agent.stream(messages, {
     context: [{ role: 'user', content: '<client-context>…page state…</client-context>' }],
     memory: { thread, resource },
   });
   ```
   (We do this through `@mastra/ai-sdk`'s `chatRoute`, which forwards `body.context`.)
3. Chat until the unobserved window crosses the buffering threshold **mid-turn** (a multi-step turn with tool calls makes the step-boundary check fire).
4. Read the thread's stored messages.

Result: Once the buffering cycle fires, the context message is upserted into storage as a `role: 'user'` message with a generated UUID and `metadata.mastra.sealed: true`. Turns before the threshold don't leak; the turn where buffering first fires does.

N/A (sorry)

### Expected Behavior

Only real user turns and assistant responses are stored; the `context` message is never persisted (matching `drainUnsavedMessages` behavior).

### Environment Information

```block
- `@mastra/memory` 1.23.0 (also present on current `main`)
- `@mastra/core` 1.51.0
- `@mastra/pg` 1.16.0 (PostgresStore)
- OM config: `observationalMemory: { model: 'google/gemini-3-flash-preview', observation: { messageTokens: 30_000 }, reflection: { observationTokens: 40_000 } }` (async buffering defaults on)


  System:
    OS: macOS 15.4.1
    CPU: (10) arm64 Apple M4
    Memory: 111.56 MB / 32.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 22.19.0 - /Users/bickio/.nvm/versions/node/v22.19.0/bin/node
    npm: 10.9.3 - /Users/bickio/.nvm/versions/node/v22.19.0/bin/npm
    pnpm: 11.13.0 - /Users/bickio/.nvm/versions/node/v22.19.0/bin/pnpm
  Browsers:
    Chrome: 150.0.7871.124
    Firefox: 152.0.5
    Safari: 26.1
  npmGlobalPackages:
    @google/gemini-cli: 0.16.0
    @zendesk/zcli: 1.0.0-beta.54
    corepack: 0.34.0
    mastracode: 0.14.0
    npm: 10.9.3
```

### Verification

serhiileniv:
## Description

Messages passed through the `context` option of `agent.stream()` / `agent.generate()` are per-run ephemeral input. Core already treats them that way: `MessageStateManager` routes them into `userContextMessages`, and `drainUnsavedMessages` only drains input and response messages, so the normal save path never writes them.

Observational Memory bypassed that. Every OM entry point built its window from `messageList.get.all.db()`, which includes context messages, and then sealed and persisted candidates through raw upserts whose only exclusion is `role === 'system'` (`Memory.persistMessages`, and `filterMessagesForPersistence` in core). Since context messages are typically `role: 'user'`, once a buffering cycle fired they were written to the messages table as durable user messages with a generated UUID — after which they were indistinguishable from a real user turn, came back from `listMessages`/recall on every later run, and rendered as user bubbles in any UI hydrating from stored history.

This takes option (a) from the issue: exclude `context`-sourced messages from OM entirely.

**Before**

```
mastra_messages after one buffered turn
  r0X2Ngcc…  user       "Suggest a collar"
  74296514…  user       <client-context>{"page":…}</client-context>   ← leaked, sealed:true
  9a24a87b…  assistant  …
```

**After** — the context message stays in the prompt for its own run and is never persisted.

### What changed

Added `getObservableMessages(messageList)` in `@mastra/memory`, which drops context-sourced messages using the id set from `messageList.makeMessageSourceChecker().context`, and used it for every OM read of the live message list:

- **window + seal + persist** — the step-boundary buffering path, the sync-observation seal, and `turn.end()` idle buffering. These are the paths that actually reached storage.
- **token accounting** — `getStatus`, `activate`, and the processor's pending-token write, so context tokens no longer push OM over its thresholds or land in durable record state.
- **marker attachment** — `persistMarkerToMessage` (both copies) can no longer mutate an OM marker onto a context-supplied assistant message and upsert it.
- **`cleanupMessages` / `filterObservedMessages` / temporal markers / last-model detection** — same rule, so a context message can't be chosen as a pruning boundary or as "the last activity/model".

Deliberately left unfiltered: the two repro-capture sites (`processor.ts` snapshot and `repro-capture.ts`), which id-diff against each other — filtering one side would make the capture report every context message as "added". `cleanupMessages`' return value also stays unfiltered, since that value is "what remains in the prompt", and context messages do remain in the prompt.

Neither persist function can be fixed in place: they never receive the `MessageList`, so they have no notion of message source. The exclusion has to happen where the arrays are built.

## Related issue

Fixes #19730

## Type of change

## Checklist

- [x] Added a changeset (`@mastra/memory` patch)
- [x] Tests pass locally

### Testing

- New regression test in `idle-buffering.test.ts`: a context message must not appear in the window handed to `buffer()`. Verified red/green — without the fix it fails with `['msg-0','msg-1','msg-2','ctx-1']` vs expected `['msg-0','msg-1','msg-2']`.
- New unit tests for `getObservableMessages` in `message-utils.test.ts`.
- `packages/memory` unit suite: 37 files / 1220 tests pass. `tsc --noEmit`, eslint and prettier clean.

<sub>Used AI assistance while writing this fix; I reviewed and tested the change and can explain every line.</sub>

🦔

## ELI5

Temporary context messages support one run only, but Observational Memory was saving them permanently. This PR keeps them available to the model and excludes them from memory processing and persistence.

## What changed

- Added `getObservableMessages()` to exclude `context`-sourced messages from Observational Memory.
- Applied filtering to observation windows, buffering, sealing, persistence, token counting, marker handling, temporal markers, cleanup, and last-model detection.
- Kept context messages available in active prompts and cleanup results.
- Added regression and unit tests for filtering and idle buffering.
- Added a changeset for the `@mastra/memory` patch release.


itskrishna21:
@serhiileniv  / maintainers, could you please resolve the merge conflicts and rebase this branch onto main so it can be merged?

serhiileniv:
@itskrishna21 Done — rebased onto `main` and force-pushed (`afd5f1e1`). The PR is mergeable again.

Two conflicts, both from changes that landed in the OM code while this sat:

- **`observation-strategies/base.ts`** — #20846 changed `persistMarkerToMessage` to return `Promise<boolean>` so callers can fall back to `persistMarkerToStorage`. Kept the new signature; the filter just applies to the message read.
- **`observation-turn/step.ts`** — #20853 added `reflectionHooks` alongside the `lastActivityAt` lines this PR touched (kept both), and #20846 hoisted the three separate window reads into a single `observableMessages` const that excludes the seeded step-0 response message. Rather than reinstating the filter at each of the now-collapsed sites, I folded it into that hoist.

Worth flagging for review: #20846 also added a **new** observation window at `step.ts:80` (`step0Messages`, feeding `om.activate()`) that reads the unfiltered list on current `main`. The rebase carried the filter onto it, which is where it belongs — that path would otherwise be a fresh instance of the same leak this PR fixes.

I re-audited every remaining `messageList.get.all.db()` in the OM directory against current `main`. The only unfiltered production sites left are the two repro-capture paths (they id-diff against each other) and `cleanupMessages`' return value (that value is "what remains in the prompt", and context messages do remain in the prompt) — both as described in the PR body. Same 11 files, 142 insertions; no scope creep from the rebase.

Verified locally after the rebase:

- `packages/memory` unit suite: **38 files / 1283 tests pass** (1 todo)
- `tsc --noEmit`, `oxlint`, `eslint`: clean
- `oxfmt --check` on the changed files: clean

The two red `Vercel – …` checks are `Authorization required to deploy`, which needs a Mastra team member to authorize the fork deploy — they were failing before the rebase too and aren't related to the change.

itskrishna21 (approved):

Issue #17646, opened 2026-06-06
[BUG] Windows fails to clone templates due to shell-quote escaping in create-mastra


DawitMengistu:
## Problem

On Windows, `npx create-mastra@latest` fails when cloning templates.

The generated git command becomes: git clone https://github.com/mastra-ai/template-chat-with-pdf

This causes Git to fail with:

fatal: Too many arguments

## Root cause

In `packages/cli/src/utils/clone-template.ts`, commands are built using:

- `shell-quote.quote()`
- `child_process.exec()`

This causes incorrect escaping of URLs on Windows.

## Reproduction

Run:

npx create-mastra@latest --template chat-with-pdf

On Windows, the clone step fails.

## Suggested fix

Replace `shell-quote + exec` with `execFile` and pass arguments as an array to avoid shell escaping issues.

## Environment

- Windows 10/11
- Node.js (any recent version)

### Steps To Reproduce

Reproduction
Run:

npx create-mastra@latest --template chat-with-pdf

### Link to Minimal Reproducible Example

https://github.com/mastra-ai/mastra/

### Expected Behavior

On Windows, the clone step fails. 

npx create-mastra@latest --template chat-with-pdf
│
◇  What is your project name?
│  chat-with-pdf
│
◇  Select a default provider:
│  OpenAI
│
◇  Initialize a new git repository?
│  Yes
✖ Failed to clone template: Failed to clone repository: Command failed: git clone https\://github.com/mastra-ai/template-chat-with-pdf 'C:\Users\Sisay M\Desktop\loadcell\chat-with-pdf'
fatal: Too many arguments.

### Environment Information

```block
None
```

### Verification

- [x] I have searched the existing issues to make sure this is not a duplicate
- [x] I have included sufficient information for the team to reproduce and understand the issue

daneatmastra:
Thank you for reporting this issue! We have labeled it for trio-wp and we will look into it as soon as possible.

If you're experiencing an error, please provide a [minimal reproducible example](https://github.com/mastra-ai/mastra/blob/main/CONTRIBUTING.md#minimal-reproduction) whenever possible to help us resolve it quickly.

tehfreak:
Also hit this on Windows. I put together a reproduction with a windows/ubuntu CI matrix, so the platform split is pinned by the runners rather than just described:
https://github.com/tehfreak/repro-mastra-clone-template-windows

Notes on the open PR, including why the `execFile` approach still fails for `npx` and `npm` on Windows:
[...]

PR #19738, opened 2026-07-20
fix(cli): write the api key to .env without a shell

### LINKED ISSUE #19736: [BUG] create-mastra writes a corrupted API key to .env on Windows
### Describe the Bug

`writeAPIKey` appends the key to `.env` by shelling out:

`packages/cli/src/commands/init/utils.ts:671-681`

```ts
const escapedKey = shellQuote.quote([key]);
const escapedApiKey = shellQuote.quote([apiKey ? apiKey : 'your-api-key']);
await exec(`echo ${escapedKey}=${escapedApiKey} >> ${envFileName}`);
```

On Windows `child_process.exec` runs `cmd.exe`, and two things go wrong:

1. `shell-quote` escapes `=` as `\=`. That's correct POSIX and meaningless to `cmd.exe`, so the backslashes are written into the file literally.
2. `cmd.exe` `echo` keeps the space before the redirect, so every value picks up a trailing space.

Any key containing `=` arrives corrupted, which covers base64-padded keys. The scaffold exits 0 with no warning, and it surfaces later as an auth error with nothing pointing back at the CLI.

### Steps To Reproduce

On Windows:

```
npx create-mastra@latest myapp --llm openai --llm-api-key "AIzaSyBx9K2mQ==" -c agents,tools --no-example --dir src/
```

Read the raw bytes, since `type` hides the trailing space:

```powershell
[System.IO.File]::ReadAllText("myapp\.env") | ConvertTo-Json
```

Don't add `--default`, it drops `--llm-api-key` entirely and writes `.env.example` instead, which hides this.

https://github.com/tehfreak/repro-mastra-writeapikey-windows

### Expected Behavior

The key is written to `.env` byte for byte.

```
key passed in      ubuntu-latest (correct)              windows-latest (actual)
sk-proj-abc123     OPENAI_API_KEY=sk-proj-abc123\n      OPENAI_API_KEY=sk-proj-abc123 \r\n
sk-proj-a+b/c=d    OPENAI_API_KEY=sk-proj-a+b/c=d\n     OPENAI_API_KEY=sk-proj-a+b/c\=d \r\n
AIzaSyBx9K2mQ==    OPENAI_API_KEY=AIzaSyBx9K2mQ==\n     OPENAI_API_KEY=AIzaSyBx9K2mQ\=\= \r\n
```

The repro runs the published CLI on both platforms in CI, so the split is pinned by the runners. Mastra's own workflows are ubuntu only, which is why this doesn't show up here.

`fs.appendFile` instead of the shell would cover both defects.

### Environment Information

```
System:
  OS: Windows 11 10.0.26200
  CPU: (16) x64 12th Gen Intel(R) Core(TM) i5-12600K
Binaries:
  Node: 22.22.2
  npm: 10.9.7
  pnpm: 11.9.0
create-mastra: 1.19.0
```

### Verification

tehfreak:
## Description

`writeAPIKey` appended the key to `.env` by shelling out:

```ts
const escapedKey = shellQuote.quote([key]);
const escapedApiKey = shellQuote.quote([apiKey ? apiKey : 'your-api-key']);
await exec(`echo ${escapedKey}=${escapedApiKey} >> ${envFileName}`);
```

On Windows `child_process.exec` runs `cmd.exe`, where `shell-quote`'s POSIX escaping means nothing, so `=` was written as `\=`. `cmd.exe` `echo` also keeps the space before the redirect, so every value picked up a trailing space:

```
key passed in      ubuntu (correct)                   windows (before this PR)
sk-proj-abc123     OPENAI_API_KEY=sk-proj-abc123\n    OPENAI_API_KEY=sk-proj-abc123 \r\n
sk-proj-a+b/c=d    OPENAI_API_KEY=sk-proj-a+b/c=d\n   OPENAI_API_KEY=sk-proj-a+b/c\=d \r\n
AIzaSyBx9K2mQ==    OPENAI_API_KEY=AIzaSyBx9K2mQ==\n   OPENAI_API_KEY=AIzaSyBx9K2mQ\=\= \r\n
```

Any key containing `=` arrived unusable, which covers base64-padded Google and Azure credentials. The scaffold exited 0 with no warning, so it surfaced later as an auth error with nothing pointing back at the CLI.

`fs.appendFile` writes the value verbatim on every platform. It also removes the shell interpolation of the key, and makes the function testable with the `memfs` mock the rest of `utils.test.ts` already uses. `fs` was already imported here.

The tests fail against the previous implementation and pass with this change. Worth noting why they fail: the old version shelled out, so it wrote real files next to the package instead of into `memfs`, which is what made this function untested until now.

`shell-quote` stays a dependency of the package, since `clone-template.ts` still uses it. That file has its own Windows problem and its own PR (#17645), so it's untouched here.

## Related issue(s)

Fixes #19736

Reproduction with a windows/ubuntu CI matrix: https://github.com/tehfreak/repro-mastra-writeapikey-windows

## Type of change

- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [ ] Test update

## Checklist

## ELI5

Previously, the CLI used the computer’s shell to “echo” API keys into `.env`, which behaved differently on Windows and could mangle keys (especially ones containing `=`). Now it writes the `.env` line directly as text, so the key is saved exactly as entered.

## Changes

- Fixed `writeAPIKey` to append `KEY=value` to `.env` using `fs.appendFile` (no shell `echo`), preserving special characters byte-for-byte across platforms (Windows included).
- Updated fallback behavior for missing keys to match existing style (`apiKey || 'your-api-key'`).
- Added `writeAPIKey` tests using the existing `memfs` mock to verify:
  - keys (including ones containing `=` and base64-padded values) are written verbatim,
  - no `.env` is created when no key is provided, and `.env.example` gets the placeholder,
  - new entries are appended without overwriting existing `.env` content.
- Added a changeset documenting that `mastra init` on Windows previously wrote a corrupted API key to `.env`.
- Removed now-unused shell/execution-related imports from `utils.ts` (kept `shell-quote` for other existing usages).

### DIFF
--- a/.changeset/true-months-sing.md
+++ b/.changeset/true-months-sing.md
@@ -1,0 +1,5 @@
+---
+'mastra': patch
+---
+
+Fixed `create-mastra` writing a corrupted API key to `.env` on Windows. The key was appended with a shell `echo`, so `cmd.exe` left a stray backslash before every `=` and a trailing space on the value. Keys containing `=`, including base64-padded Google and Azure credentials, were written unusable and the scaffolded project failed to authenticate with no warning from the CLI.
--- a/.changeset/true-months-sing.md
+++ b/.changeset/true-months-sing.md
@@ -2,4 +2,4 @@
 'mastra': patch
 ---
 
-Fixed `create-mastra` writing a corrupted API key to `.env` on Windows. The key was appended with a shell `echo`, so `cmd.exe` left a stray backslash before every `=` and a trailing space on the value. Keys containing `=`, including base64-padded Google and Azure credentials, were written unusable and the scaffolded project failed to authenticate with no warning from the CLI.
+Fixed `mastra init` writing a corrupted API key to `.env` on Windows. The key was appended with a shell `echo`, so `cmd.exe` left a stray backslash before every `=` and a trailing space on the value. Keys containing `=`, including base64-padded Google and Azure credentials, were written unusable, and the project failed to authenticate with no warning from the CLI.
--- a/packages/cli/src/commands/init/utils.test.ts
+++ b/packages/cli/src/commands/init/utils.test.ts
@@ -28,7 +28,8 @@ vi.mock('../auth/orgs.js', () => ({
   resolveCurrentOrg: vi.fn(),
 }));
 
-const { getAPIKey, promptForObservability, writeObservabilityEnv, writeIndexFile } = await import('./utils');
+const { getAPIKey, promptForObservability, writeAPIKey, writeObservabilityEnv, writeIndexFile } =
+  await import('./utils');
 const prompts = await import('@clack/prompts');
 const { getToken, loadCredentials } = await import('../auth/credentials.js');
 const { resolveCurrentOrg } = await import('../auth/orgs.js');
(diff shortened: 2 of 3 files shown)


tehfreak:
Both addressed in 563b507.

**Changeset frontmatter mismatch.** Fair catch. The change lives in `packages/cli`, which publishes as `mastra`, and `writeAPIKey` is reached through `mastra init`, so the description now says that instead of naming `create-mastra`. `create-mastra` picks the fix up through its `workspace:^` dependency rather than through a change of its own, so a second changeset for it would describe a package that has no source change here.

**Fallback style.** Switched to `apiKey || 'your-api-key'`. Identical behaviour for `undefined` and for the empty string, and it matches the rest of the file.

wardpeet (approved):

LekoArts:
Thanks for the PR on this! I'm working on a major refactor of the CLI and incorporated your changes into my PR. I added you as co-author on that commit: https://github.com/mastra-ai/mastra/pull/19825

tehfreak:
@LekoArts thanks, glad it helped. The carry-over looks right: the `fs.appendFile` change and the round-trip test are both intact in #19825.

One thing worth folding in while this flow is open. `clone-template.ts` still has the sibling of that same bug. In #19825 lines 113, 139 and 189 build the degit, git and install commands with `shell-quote` and run them through `child_process.exec`:

```ts
const degitCommand = shellQuote.quote(['npx', 'degit', repoWithBranch, targetPath]);
await exec(degitCommand, { ... });
```

On Windows `exec` runs `cmd.exe`, where `shell-quote`'s POSIX output is malformed the same way it was for the API key: the URL colon becomes `https\://...` and the target path keeps its single quotes. So `create-mastra --template` can't clone on Windows. That's #17646.

One trap before fixing it: the plain `exec` -> `execFile` swap (what the open PR #17645 does) fixes the git call but breaks the other two. `npx` and `npm` are `.cmd` shims on Windows, and `execFile` with no shell can't spawn them:

```
execFile('npx', ['degit', ...])   spawn npx ENOENT
execFile('npm', ['install'])      spawn npm ENOENT
execFile('git', [...])            ok
```

`execa` (already a dependency of `packages/cli`) takes an args array, needs no shell, and resolves the `.cmd` shims, so it covers all three. Same spirit as the `appendFile` change: drop the shell, pass args directly.

Mastra CI is ubuntu-only, so none of this shows up there. I have two repros with a windows/ubuntu matrix that pin the split, if useful:

- clone path: https://github.com/tehfreak/repro-mastra-clone-template-windows
- key path: https://github.com/tehfreak/repro-mastra-writeapikey-windows

PR #19739, opened 2026-07-20
fix(server): handle Agent Controller background failures

### LINKED ISSUE #19734: [BUG] Agent Controller routes can crash on rejected background session operations
## Description

The Agent Controller messages, steer, and follow-up HTTP routes start asynchronous session methods with void and immediately return { ok: true }. If a session operation rejects, its promise is unobserved.

Session.sendMessage deliberately rejects when signal submission fails before a stream starts. The existing core test at packages/core/src/agent-controller/__tests__/signal-messages.test.ts covers that behavior. On supported Node versions, the route can therefore acknowledge the request and then terminate the server with an unhandled rejection.

## Reproduction

1. Configure an AgentController session whose signal acceptance rejects before a stream starts.
2. Call the messages endpoint, or call steer/follow-up while the session is idle.
3. The endpoint returns success while the rejected session promise has no catch handler.

## Expected behavior

The endpoint should retain its immediate acknowledgement semantics, but rejected background session work must be handled and logged rather than causing an unhandled rejection. Add regression coverage for the three route handlers.

aadityakumarsah:
Fixes #19734.\n\nAgent Controller message, steer, and follow-up routes still acknowledge requests immediately, but now catch and log rejected session operations. This avoids unhandled promise rejections that can terminate a Node process after a successful HTTP response.\n\nTests:\n- pnpm --filter @mastra/server exec vitest run src/server/handlers/agent-controller.test.ts\n- targeted ESLint and Prettier checks\n- pnpm --filter @mastra/server build:lib

## ELI5

The server now quietly handles background session failures instead of letting them crash the app, while still quickly telling callers that their request was accepted.

## What changed

- Added error logging for rejected background `sendMessage`, `steer`, and `followUp` operations.
- Preserved immediate `{ ok: true }` responses for all three routes.
- Added regression tests verifying failures are logged without causing unhandled promise rejections.
- Added a patch changeset for `@mastra/server`.
- Server build checks and regression tests passed.

### DIFF
--- a/packages/server/src/server/handlers/agent-controller.ts
+++ b/packages/server/src/server/handlers/agent-controller.ts
@@ -497,7 +497,9 @@ export const SEND_AGENT_CONTROLLER_MESSAGE_ROUTE = createRoute({
       // Forward the server middleware's requestContext so identity injected in
       // `server.middleware` reaches dynamic instructions and tools (same as the
       // plain agent message route).
-      void session.sendMessage({ content: message, files, requestContext });
+      void session.sendMessage({ content: message, files, requestContext }).catch(error => {
+        mastra.getLogger().error('Failed to send Agent Controller message', { controllerId, resourceId, error });
+      });
       return { ok: true };
     } catch (error) {
       return handleError(error, 'error sending controller message');
@@ -602,7 +604,9 @@ export const STEER_AGENT_CONTROLLER_SESSION_ROUTE = createRoute({
     try {
       const controller = getAgentControllerOrThrow(mastra, controllerId);
       const session = await getSession(controller, resourceId, { scope: sessionScope });
-      void session.steer({ content: message, requestContext });
+      void session.steer({ content: message, requestContext }).catch(error => {
+        mastra.getLogger().error('Failed to steer Agent Controller session', { controllerId, resourceId, error });
+      });
       return { ok: true };
     } catch (error) {
       return handleError(error, 'error steering controller session');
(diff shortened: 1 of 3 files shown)


packages/server/src/server/handlers/agent-controller.ts
```diff
@@ -497,7 +497,9 @@ export const SEND_AGENT_CONTROLLER_MESSAGE_ROUTE = createRoute({
       // Forward the server middleware's requestContext so identity injected in
       // `server.middleware` reaches dynamic instructions and tools (same as the
       // plain agent message route).
-      void session.sendMessage({ content: message, files, requestContext });
+      void session.sendMessage({ content: message, files, requestContext }).catch(error => {
+        mastra.getLogger().error('Failed to send Agent Controller message', { controllerId, resourceId, error });
```
NikAiyer, line 501: Could we normalize the caught value before including it in the structured metadata? Native `Error` properties are non-enumerable, so JSON-backed loggers—and Pino when the value is under `error` rather than its recognized `err` key—can emit `error: {}`. `DualLogger` also only normalizes an `Error` passed as a separate argument, not one nested here.

Since this change is intended to improve failure observability, please preserve at least `name`, `message`, and `stack` as plain fields, with `String(error)` as the fallback for non-`Error` rejections. This applies to the steer and follow-up handlers as well, and the regression tests should assert the normalized payload.

NikAiyer (changes requested):
The branch currently conflicts with `main`. When resolving, please retain the newer request-context propagation and exact thread/session binding behavior alongside these rejection handlers.

abhiaiyer91:
Closing as superseded by merged PR #21093, which resolved linked issue #19734 by reporting background Agent Controller failures instead of crashing the server. Thanks for the contribution.

PR #19752, opened 2026-07-20
fix(inngest): restore nested workflow resume path

graysonhicks:
## Description

Nested Inngest workflows received an empty child resume path when the parent was resumed with only the nested workflow ID. This restores the suspended child step and its path from the nested snapshot before invoking the child workflow, so the original tool receives the resume payload instead of starting a fresh execution.

This addresses Issue 1 in the upstream defect report submitted directly by Interfere. It is independent of #19429 and does not change actor propagation or durable trigger/resume event fields.

Validation: the focused execution-engine and actor-signal suites pass with 12 tests, package lint passes, and CodeRabbit reported no findings. The package build remains blocked by existing generated-type errors involving `RequestContext` and `EntityType` across unchanged Inngest files.

## Related issue(s)

No public GitHub issue; reported directly by Interfere as Issue 1 of their upstream defect report.

## Type of change

- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [x] Test update

## Checklist

## ELI5

If one workflow pauses inside another, resuming it now reliably continues the inner workflow from the exact spot it stopped. It does this by reloading the saved “where it was paused” info from the inner workflow’s snapshot, instead of starting fresh or guessing.

## Summary

- Restores nested Inngest workflow resume-path handling for “resume with only the nested workflow id”.
- When nested resume details are incomplete, `executeWorkflowStep`:
  - Loads the nested workflow snapshot (`suspendedPaths`) to reconstruct the suspended child step list.
  - Sets the nested `resumePath` from that snapshot.
- Rejects ambiguous nested resumes:
  - Errors if no suspended child steps exist in the nested snapshot.
  - Errors if multiple suspended child steps exist (requiring the caller to specify the `step` parameter), and avoids invoking the nested workflow.
- Adds/updates tests covering:
  - Execution-engine restoration of the suspended child path and forwarding of the correct nested resume payload.
  - Actor-signal threading through nested workflow resume (ensuring `actor` is present on the nested `RESUME` invoke payload).
- Updates the `@mastra/inngest` changeset for a patch release.

## Validation

- 12 focused tests passing.
- Package lint passing.
- Package build remains blocked by existing generated-type errors involving `RequestContext` and `EntityType`.

### DIFF
--- a/workflows/inngest/src/execution-engine.ts
+++ b/workflows/inngest/src/execution-engine.ts
@@ -523,6 +523,15 @@ export class InngestExecutionEngine extends DefaultExecutionEngine {
           runId: runId,
         });
 
+        const nestedResumeSteps = resume.steps.slice(1);
+        if (nestedResumeSteps.length === 0) {
+          const suspendedStepId = Object.keys(snapshot?.suspendedPaths ?? {})[0];
+          if (suspendedStepId) {
+            nestedResumeSteps.push(suspendedStepId);
+          }
+        }
+        const nestedResumeStepId = nestedResumeSteps[0];
+
         const invokeResp = (await this.inngestStep.invoke(`workflow.${executionContext.workflowId}.step.${step.id}`, {
           function: step.getFunction(),
           data: {
@@ -532,10 +541,10 @@ export class InngestExecutionEngine extends DefaultExecutionEngine {
             runId: runId,
             resume: {
               runId: runId,
-              steps: resume.steps.slice(1),
+              steps: nestedResumeSteps,
               stepResults: snapshot?.context as any,
               resumePayload: resume.resumePayload,
-              resumePath: resume.steps?.[1] ? (snapshot?.suspendedPaths?.[resume.steps?.[1]] as any) : undefined,
+              resumePath: nestedResumeStepId ? (snapshot?.suspendedPaths?.[nestedResumeStepId] as any) : undefined,
             },
             outputOptions: { includeState: true },
             perStep,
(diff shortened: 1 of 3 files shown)


workflows/inngest/src/execution-engine.ts
```diff
@@ -523,6 +523,15 @@ export class InngestExecutionEngine extends DefaultExecutionEngine {
           runId: runId,
         });
 
+        const nestedResumeSteps = resume.steps.slice(1);
+        if (nestedResumeSteps.length === 0) {
+          const suspendedStepId = Object.keys(snapshot?.suspendedPaths ?? {})[0];
```
NikAiyer: **Non-blocking:** Can `suspendedPaths` contain multiple entries here? General nested workflows can suspend parallel children, so choosing the first key may route the resume payload to an arbitrary branch. Core’s top-level resume detection rejects this ambiguity and requires an explicit target. Should this fallback only apply when exactly one suspended child exists, with a test for the multi-path case?

NikAiyer (approved):
The nested resume-path restoration looks correct for the reported single-child case, and the focused regression test covers the repaired contract well.

workflows/inngest/src/execution-engine.ts
graysonhicks: Addressed and pushed.

The fallback now applies only when exactly one child is suspended. Zero or multiple suspended children return an explicit error without invoking an arbitrary branch. Added focused coverage for single, empty, and multi-path snapshots.

PR #19754, opened 2026-07-20
test(inngest): cover resume rehydration after cleanup

### LINKED ISSUE #19330: @mastra/inngest: workspace skill/filesystem tools (`skill`, `skill_read`, `skill_search`, `mastra_workspace_*`) throw ToolNotFoundError when durable steps run on a cross-process connect() worker
## Package & version

- `@mastra/core` 1.50.1
- `@mastra/inngest` 1.8.1
- `inngest` 4.12.1
- `mastra` 1.18.2
- Node v22.15.1

## Bug

A `createInngestAgent` durable agent whose (Mastra-instance) `Workspace` has `skills` configured **cannot call its `skill` tool once the durable steps execute on a separate Inngest `connect()` worker process.** The model emits a correct `skill({ name: "greeting" })` tool call and the durable **tool-call** step rejects it with:

```
ToolNotFoundError: Tool "skill" not found. Call tools by their exact name only — never add prefixes, namespaces, or colons.
```

The **same agent, same config, same prompt** works perfectly when the run executes **in-process** (no connect worker). The only variable is whether execution crosses a process boundary.

This affects every tool that is a per-request closure over the `Workspace`:
- `skill`, `skill_read`, `skill_search`
- all `mastra_workspace_*` filesystem tools

Static `createTool` tools (e.g. a plain `request_approval`) resolve fine cross-process — only the workspace/skill-derived tools drop.

## Expected vs actual

| | Behavior |
|---|---|
| **Control — in-process (no connect worker)** | Agent calls `skill` → loads the skill → replies with the skill content ✅ |
| **Cross-process (connect worker up)** | Agent calls `skill` → `ToolNotFoundError`; the turn ends with no answer ❌ |

## Root cause

The durable agentic loop runs as two separate steps per iteration:

1. **`durable-llm-execution`** decides the model's tools. Cross-process it calls `resolveRuntimeDependencies` (`@mastra/core/dist/agent/durable/index.js`), which **correctly rebuilds the full toolset** — including the workspace `skill`/`mastra_workspace_*` tools — via `mastra.getAgentById(agentId).getToolsForExecution(...)`. This is why the model *sees* and *calls* `skill`.
2. **`durable-tool-call`** *executes* the call. It resolves the tool **only** from `globalRunRegistry.get(runId)?.tools?.[toolName]`, with a fallback to `resolveTool(toolName, mastra)` / `mastra.listTools()` that only sees **Mastra-instance-level** tools. **It never calls `resolveRuntimeDependencies`.**

`globalRunRegistry` is a **per-process** in-memory `TTLCache`. On a connect worker the run was prepared in a *different* process, so the worker's registry entry is the empty placeholder that `@mastra/inngest` seeds (`@mastra/inngest/dist/index.js`):

```js
existingEntry = {
  // Minimal placeholder fields. The durable LLM step recreates tools
  // and model from the workflow input; this slot exists primarily to
  // carry the abort controller across the resumed segment.
  tools: {},
  model: void 0
};
globalRunRegistry.set(runId, existingEntry);
```

So in the tool-call step `registryEntry.tools["skill"]` is `undefined`, and the `mastra.listTools()` fallback does not contain the workspace/skill tools (they are per-request closures built in `listSkillTools`/`listWorkspaceTools`, never registered at the Mastra-instance level) → `ToolNotFoundError`.

The asymmetry is the bug: the LLM-execution step rebuilds the full toolset cross-process, but the sibling tool-call step that actually needs it does not — it trusts a per-process registry that is empty on the worker.

### Secondary symptom (same root cause)

`resolveRuntimeDependencies` also does not rebuild `inputProcessors`, and the LLM-execution step reads them via `registryEntry?.inputProcessors ?? []` — empty cross-process. So on the worker the `SkillsProcessor` and `WorkspaceInstructionsProcessor` never run, meaning the system prompt is also missing the available-skills list and workspace instructions. Both symptoms stem from the per-process `globalRunRegistry` + an incomplete cross-process rebuild.

## Minimal reproduction

A self-contained ~40-line vanilla repro (canonical `createInngestAgent` + instance-level `Workspace` with `skills`, plus a standalone `connect()` worker). Steps:

```bash
npm install
cp .env.example .env          # set OPENAI_API_KEY=sk-...  (keep INNGEST_DEV=1)

# terminal 1 — self-hosted Inngest dev server pointed at the Mastra serve route
npx inngest-cli@latest dev -u http://localhost:4333/inngest/api -p 8299

# terminal 2 — Mastra server (serves the agent stream route + /inngest/api)
INNGEST_BASE_URL=http://localhost:8299 npm run dev

# CONTROL: run WITHOUT a connect worker → it works (skill resolves in-process)
npm run repro         # → ✅ skill tool executed

# terminal 3 — start the cross-process connect worker (this triggers the bug)
INNGEST_BASE_URL=http://localhost:8299 npm run worker

# re-run: durable steps now execute on the worker → ToolNotFoundError
npm run repro         # → ❌ BUG REPRODUCED
```

The agent:

```ts
  filesystem: new LocalFilesystem({ basePath: projectRoot }),
  skills: ['skills/greeting'],            // <root>/skills/greeting/SKILL.md
});
const agent = new Agent({
  id: 'greeter',
  instructions: 'When asked how to greet, first call the `skill` tool with { name: "greeting" }, then answer.',
  model: openai('gpt-4o-mini'),
});
const durable = createInngestAgent({ agent, inngest });
export const mastra = new Mastra({ agents: { greeter: durable }, workspace, server: { /* /inngest/api serve route */ } });
```

I can push the full repro to a public repo if that's easier to consume — happy to link it here.

## Fix direction

The durable **tool-call** step (or the `@mastra/inngest` worker path) should rebuild the agent toolset via `resolveRuntimeDependencies` / `getToolsForExecution` when the per-run registry entry is missing/empty — the same way the LLM-execution step already does — so workspace/skill tools resolve cross-process. Rebuilding `inputProcessors` in `resolveRuntimeDependencies` (and writing the rebuilt tools + inputProcessors back into `globalRunRegistry` on the worker) would fix both the tool lookup and the missing `SkillsProcessor`/`WorkspaceInstructionsProcessor`.

## Related

- #17789 — DurableAgent: run terminates when every tool call in a step errors (same file, same "durable diverges from in-process" theme)
- #19056 — Durable agents: RUNNING runs unrecoverable after restart (confirms `globalRunRegistry` is a per-process in-memory `TTLCache`, which this bug also stems from)

## Verification

- [x] I searched existing issues — no match for cross-process workspace/skill tool drop
- [x] Minimal reproduction provided (control passes in-process, bug triggers cross-process)

graysonhicks:
## Description

Adds production-shaped regression coverage for Issue 2 in the durable-agent defect report. The test starts an Inngest durable stream, waits for its trigger, cleans up the initial in-memory run registry, and resumes the same run so the adapter creates its cross-process placeholder entry.

It then exercises the runtime resolver used by the durable LLM step and verifies that the persisted request context is restored while the live model and tools are rebuilt from the registered Mastra agent and the placeholder is marked as hydrated.

The runtime implementation for this defect already landed in #19331, which fixed #19330. This PR intentionally adds only the missing lifecycle regression and does not duplicate that implementation. It also shares no changed paths with #19338, #19429, or #19752.

Validation: the focused regression passes, package lint passes, and CodeRabbit reported no findings after review. No changeset is included because this test-only follow-up does not change published behavior.

## Related issue(s)

Follow-up to #19331 and #19330. Also covers Issue 2 from the upstream defect report submitted directly by Interfere.

## Type of change

- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [x] Test update

## Checklist

## ELI5

This adds a safety check proving that a paused workflow can continue on another worker with its saved information and tools restored correctly. It tests the existing runtime fix without changing published behavior.

## Summary

- Adds regression coverage for durable-agent resume runtime rehydration.
- Verifies persisted request context, including `tenantId`, is restored.
- Confirms the registered Mastra agent rebuilds the model and execution tools.
- Verifies the cross-process placeholder registry entry becomes hydrated.
- Includes cleanup for initial and resumed workflow runs.
- Adds no runtime changes, public API changes, or changeset.


workflows/inngest/src/resume-runtime-rehydration.test.ts
```diff
@@ -0,0 +1,100 @@
...
+        getAgentById: () => registeredAgent,
+        getStorage: () => ({
+          getStore: async () => ({ loadWorkflowSnapshot: async () => snapshot }),
+        }),
+      };
+      (durableAgent as any).__setMastra(mastra);
+
+      const resumedResult = await durableAgent.resume(runId, { approved: true });
+      try {
+        const placeholder = globalRunRegistry.get(runId);
```
NikAiyer, line 76: Non-blocking: Since this placeholder primarily carries the resumed segment’s abort controller, could we capture `placeholder?.abortController` here and assert after resolution that its identity is preserved? That would protect the in-place hydration behavior in addition to the `isPlaceholder` transition.

workflows/inngest/src/resume-runtime-rehydration.test.ts
```diff
@@ -0,0 +1,100 @@
...
+        expect(placeholder?.isPlaceholder).toBe(true);
+
+        // The durable LLM step performs this resolution on the Inngest worker.
```
NikAiyer, line 81: Non-blocking follow-up: It may be useful to add coverage that drives this resolution through the actual Inngest worker handler. This direct call seems sufficient for the current regression, but worker-level coverage would additionally protect the integration wiring and serialization boundary.

NikAiyer (approved):
This looks good to me. The test covers the important Inngest lifecycle boundary: the initial registry entry is cleaned up, `resume()` creates a placeholder, persisted request context is restored, and the registered agent’s live model and tools replace the placeholder state. Approving.

PR #19757, opened 2026-07-20
docs(inngest): document generate suspension behavior

graysonhicks:
The `InngestAgent` reference currently describes `generate()` as a method forwarded to the underlying agent and does not document the durable `generate()` or `resumeGenerate()` methods.

This documents both methods, including that `generate()` resolves with `finishReason: 'suspended'`, and points callback-driven approval flows to `stream({ onSuspended })`. It also corrects the proxy wording in the reference.

This is documentation-only. It does not change runtime behavior and does not overlap with #19429.

Validated with targeted Prettier, Remark, and Vale checks, the docs metadata validators, and a production Docusaurus build.

## ELI5
This PR updates the Inngest agent docs to show how long-running AI tasks can “pause” for approval and then “continue” later using the same run ID. It also clarifies how methods not implemented by `InngestAgent` get forwarded to the underlying `Agent`.

## What changed
- Documented durable `generate(messages, options?)` and `resumeGenerate(runId, resumeData, options?)` on `InngestAgent`.
- Explained that when a durable `generate()` call suspends, it resolves with `finishReason: 'suspended'`, and that `generate()` can optionally create/return a `runId` for resuming.
- Directed callback/approval-based suspension handling to `stream({ onSuspended })`.
- Corrected and clarified proxy delegation wording for `createInngestAgent()` (what gets forwarded and when).
- Added/updated examples for using `runId` to resume suspended runs.

## Validation
- Targeted formatting and documentation validation completed.
- Production Docusaurus build checks completed successfully.

### DIFF
--- a/docs/src/content/en/reference/agents/inngest-agent.mdx
+++ b/docs/src/content/en/reference/agents/inngest-agent.mdx
@@ -77,7 +77,7 @@ Returns: [`InngestAgent`](#inngestagent-interface)
     type: 'Agent',
     isOptional: false,
     description:
-      'The Agent to wrap with Inngest durable execution. Agent methods (e.g., `generate()`, `listTools()`, `getMemory()`) delegate to this agent via a Proxy.',
+      'The Agent to wrap with Inngest durable execution. Methods not implemented by `InngestAgent` (e.g., `listTools()` and `getMemory()`) delegate to this agent via a Proxy.',
   },
   {
     name: 'inngest',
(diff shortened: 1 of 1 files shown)


docs/src/content/en/reference/agents/inngest-agent.mdx
```diff
@@ -253,6 +253,32 @@ The third argument accepts `threadId` and `resourceId` in addition to the lifecy
 
 Returns: [`Promise<InngestAgentStreamResult>`](#inngestagentstreamresult)
 
+#### `generate(messages, options?)`
+
+Runs a response on Inngest's durable execution engine and resolves with a single `FullOutput`. If the run suspends, `generate()` resolves with `finishReason: 'suspended'`. Continue the run with [`resumeGenerate()`](#resumegeneraterunid-resumedata-options). Use [`stream()`](#streammessages-options) with `onSuspended` when the caller needs a suspension callback.
+
+```typescript
+const runId = crypto.randomUUID()
```
NikAiyer: Would it be clearer to resume using the `runId` returned in `result`, or mention that providing one upfront is optional? `generate()` automatically creates and returns a run ID when one isn’t supplied, so the current example could imply that callers must generate it themselves.

NikAiyer (approved):
The documentation now matches the durable InngestAgent implementation and correctly explains buffered suspension and resume behavior. I left one non-blocking clarity suggestion about the optional run ID.

docs/src/content/en/reference/agents/inngest-agent.mdx
graysonhicks: 👍

PR #19762, opened 2026-07-20
fix(workflows): prevent duplicate durable operation IDs

graysonhicks:
## Description

Repeated and resumed durable workflow steps reused static Inngest operation IDs, allowing loop replay to trigger `AUTOMATIC_PARALLEL_INDEXING`. This scopes step lifecycle, persistence-start, execution, nested invoke, and nested result IDs with the execution path plus loop/foreach iteration and resume phase. Ordinary one-shot steps retain their existing IDs.

The focused regression runs two loop iterations, suspends iteration two, resumes it, and verifies every recorded durable operation ID is unique. Before the fix, 18 operations collapsed to 6 unique IDs.

This complements #19707 without duplicating its terminal `stepUpdate.resume` fix. The branch now includes the nested resume-path restoration from #19752 and the runtime-rehydration coverage from #19754. The `execution-engine.ts` conflict was resolved by retaining snapshot-backed child resume-path restoration while applying the scoped operation ID to nested invokes and result persistence. The actor-signal resume fixture now supplies an explicit nested step path under that restored validation behavior.

There is no file or behavior overlap with #19429 or #19338.

## Related issue(s)

Related to #19699.

## Type of change

- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [x] Test update

## Checklist

## Validation

- Core operation-ID regression: 1 test passed
- Inngest execution-engine, actor-signal, and runtime-rehydration suites: 15 tests passed
- `@mastra/core` ESLint
- `@mastra/inngest` ESLint
- Prettier and whitespace checks

The full `@mastra/inngest` test command was stopped after its server-backed `index.test.ts` remained in repeated approximately 25-second reset/setup cycles. The deterministic focused suites above completed successfully.

## ELI5

When a workflow repeats a step or resumes after a pause, it can reuse the same ID. This PR gives each operation a distinct ID while preserving existing IDs for one-time steps.

## What changed

- Scoped durable operation IDs by execution path, loop or `foreach` iteration, and resume generation.
- Applied scoped IDs to nested workflows, spans, events, retries, persistence, and result handling.
- Preserved resume metadata across suspended results.
- Added optional `operationId` propagation to the default and Inngest execution engines.
- Preserved legacy IDs for ordinary one-shot steps.

## Regression tests

- Repeated loop resumes.
- `foreach` items with identical values.
- Nested workflow operations.
- Suspend payloads and resume metadata.
- Legacy one-shot operation IDs.
- A case where 18 operations previously produced only 6 unique IDs.
- Updated actor-signal and execution-engine tests.

## Validation

- Focused workflow tests.
- Inngest execution-engine, actor-signal, and runtime-rehydration tests.
- ESLint, Prettier, and whitespace checks.
- The full server-backed Inngest test command stopped after repeated reset and setup cycles.


packages/core/src/workflows/handlers/step.ts
```diff
@@ -67,6 +67,44 @@ export interface ExecuteStepParams extends ObservabilityContext {
...
+  if (iterationCount !== undefined) {
+    operationParts.push(`iteration.${iterationCount}`);
+    persistenceParts.push(`iteration.${iterationCount}`);
+  }
+  if (executionContext.foreachIndex !== undefined) {
+    operationParts.push(`foreach.${executionContext.foreachIndex}`);
+    persistenceParts.push(`foreach.${executionContext.foreachIndex}`);
+  }
+  if (isResumingStep) {
+    operationParts.push('resume');
```
NikAiyer: Could this collide if the same step occurrence suspends again while being resumed? In loops, `currentResume` remains set when the resumed execution returns `suspended`, so a second resume of the same iteration appears to reuse the same `.iteration.N.resume` identity. Do we need a persisted resume generation here, such as `.resume.1` and `.resume.2`, to keep each occurrence replay-stable and unique?

NikAiyer (changes requested):
The operation-ID scoping looks consistent across step lifecycle operations, retries, persistence, and nested workflows. I’d like to see a little more regression coverage before approving:

- the same loop iteration suspending and resuming more than once;
- nested Inngest invoke and result operation IDs;
- foreach executions, including concurrent items or repeated values;
- confirmation that ordinary one-shot steps retain their legacy IDs.

The repeated re-suspension case is the important one because it would validate whether the boolean `.resume` suffix is sufficient. The other cases would confirm that the new identity is propagated as intended and that the compatibility behavior is protected.

PR #19791, opened 2026-07-20
fix(core): preserve schedules for stored agents

roaminro:
Agent schedules previously treated a stored agent as removed when it had not yet been hydrated after a cold start:

```ts
const agent = mastra.getAgentById(agentId);
```

The worker now falls back to the editor before cleaning up the schedule:

```ts
agent = (await mastra.getEditor()?.agent.getById(agentId)) ?? null;
```

If editor resolution fails temporarily, the fire is reported as failed but the schedule is preserved so a later run can retry. Schedules are still deleted when both the registry and editor confirm the agent is missing.

Covered by focused worker tests for successful stored-agent hydration, confirmed missing agents, and transient editor failures.

## ELI5
After a restart, scheduled tasks may reference agents that are not loaded into memory. This PR checks the editor before deleting those schedules and preserves them when the editor is temporarily unavailable.

## Changes
- Added `resolveAgentById` for registry lookup with editor fallback.
- Distinguished `found`, `missing`, and `error` resolution results.
- Preserved schedules when agent lookup fails temporarily.
- Deleted schedules only when both the registry and editor confirm that the agent is missing.
- Updated scheduler readiness checks to support asynchronous predicates.
- Updated agent schedule execution and scorer hooks to use `resolveAgentById`.
- Added tests for stored-agent hydration, transient editor failures, and confirmed missing agents.
- Updated the `@mastra/core` changeset.


roaminro:
AI says: Thanks for the thorough analysis — confirmed the gap. The registry-only predicate in `scheduler-worker.ts` deleted stored-agent rows after the grace window before `agent-schedule.fire` was ever published, so the execute-path fallback never ran in the single-process cold-start case.

Addressed in 63d187b12e, following the suggested shape:

- `SchedulerConfig.isTargetReady` now accepts `boolean | Promise<boolean>` and `#ensureTargetReady` awaits it.
- The `SchedulerWorker` predicate falls back to `await mastra.getEditor?.()?.agent.getById(...)` on a registry miss. A throwing editor lookup counts as ready (no grace miss burned on transient storage errors); only a confirmed `null` counts as a miss.
- Added three end-to-end tests in `scheduler-integration.test.ts`: a stored-only agent schedule survives ticks past `missesBeforeDelete` and fires, a throwing editor never burns misses (verified with `missesBeforeDelete: 1`), and an editor-confirmed-missing agent is still reclaimed.

Verified the first two tests fail against the old predicate. Full scheduler/schedules/worker suites pass (113 tests) and `tsc --noEmit` is clean.

Mastra Code

roaminro:
AI says: Pushed a follow-up commit (6025346) addressing the review-adjacent cleanup:

- **Shared resolution helper** — the bug existed because "registry, then editor fallback" was reimplemented per call site and one was missed. `resolveAgentById` (`packages/core/src/mastra/resolve-agent.ts`) now encodes the lookup and the confirmed-miss vs. transient-error distinction once; the scheduler readiness predicate, the agent-schedule execute path, and the scorer hook all use it. The server handler intentionally keeps its own resolution since it layers version options, stored-config overrides, and 404 mapping on top.
- **Documented trade-off** — during a sustained editor outage, the tick loop treats resolution errors as "ready", so each due tick publishes a fire that fails at execute time and records a failed trigger row. Noisy, but a schedule is never deleted on ambiguity. Now documented at the predicate.
- **Changeset** — expanded to describe both deletion paths (tick-loop reclaim and fire-path self-clean).

All scheduler/schedules/worker/hook suites pass (123 tests), `tsc --noEmit` and lint clean.

Mastra Code

abhiaiyer91 (approved):

PR #19794, opened 2026-07-20
fix(stored-agents): preserve drafts unless auto-publishing

YujohnNattrass:
Classic Studio's Save Draft was activating every newly created agent version, so runtime reads changed before Publish. This makes stored-agent PATCH updates draft-only by default and preserves the active version during code-source cleanup.

Agent Builder opts into its existing immediate-publication behavior through the new client option:

```ts
await client.getStoredAgent('agent-id').update({
  instructions: 'Updated instructions',
  autoPublish: true,
});
```

The server lifecycle tests cover draft, activation, auto-publish, and code-source cleanup. The Agent Builder MSW test covers the real client transport, and the dataset experiment test covers explicit agent-version resolution.

Verified the full server and playground test suites, server, playground, and client builds, TypeScript checks, ESLint, and server core-import checks.

### DIFF
--- a/.changeset/frank-olives-pick.md
+++ b/.changeset/frank-olives-pick.md
@@ -1,0 +1,5 @@
+---
+'@mastra/server': patch
+---
+
+Fixed stored agent draft saves to preserve the published version unless automatic publication is requested.
--- a/.changeset/hot-melons-wave.md
+++ b/.changeset/hot-melons-wave.md
@@ -1,0 +1,12 @@
+---
+'@mastra/client-js': minor
+---
+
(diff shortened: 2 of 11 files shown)


NikAiyer:
Non-blocking follow-up: `handleAutoVersioning` creates a specific `versionId` but returns only `versionCreated`, so the `autoPublish` path subsequently activates whichever version is latest. Overlapping PATCHes could therefore cause one request to activate another request’s version. This is pre-existing and this PR reduces its scope, so I don’t think it should block this fix, but returning the created version identity would make publication deterministic.

NikAiyer (approved):

PR #19795, opened 2026-07-20
fix(voyageai): correct multimodal input shape and add baseUrl

### LINKED ISSUE #19801: @mastra/voyageai: multimodal input shape rejected by API + no baseUrl passthrough
## Summary

`@mastra/voyageai` cannot embed against a provider-hosted Voyage endpoint. Two problems combine: multimodal requests are serialized in a shape the API rejects, and there is no way to set a custom base URL.

## 1. Multimodal requests return HTTP 400

`VoyageMultimodalEmbeddingModel.doEmbed` serializes text content to bare strings, producing `inputs: [["text"]]`. The Voyage multimodal API, and the SDK's own `MultimodalEmbedRequestInputsItem` type, require `inputs: [{ content: [{ type: 'text', text }] }]`. The bare-string shape returns:

```
HTTP 400: "Expected object. Received list"
```

So multimodal text embedding does not work at all.

## 2. No `baseUrl` passthrough

The embedder, reranker, and contextualized models always construct `new VoyageAIClient({ apiKey })`. There is no way to point the client at a provider-hosted endpoint (for example MongoDB Atlas's `https://ai.mongodb.com/v1`), even though the underlying `voyageai` SDK accepts `baseUrl`.

## Reproduction

```ts
import { voyageMultimodalEmbedding } from '@mastra/voyageai';

const model = voyageMultimodalEmbedding({ model: 'voyage-multimodal-3.5', apiKey: process.env.VOYAGE_API_KEY });
await model.doEmbed({ values: [{ content: [{ type: 'text', text: 'hello' }] }] });
// -> HTTP 400: Expected object. Received list
```

## Related

There is also a packaging problem that blocks any ESM app: `@mastra/voyageai` depends on `voyageai@^0.2.1`, whose ESM build has a broken directory import (`export * from "../api"` with no `/index.js`) that Node's native ESM resolver rejects under `tsx` and `vitest`. 0.3.1 ships a working ESM build with the same client API.

## Fix

I have a PR ready (was #19795, closed for lacking a linked issue): correct the multimodal input shape, add optional `baseUrl` to all four config types with a conditional spread so existing `{ apiKey }`-only construction is byte-identical, and bump `voyageai` to `^0.3.1`. Verified live against the MongoDB-hosted endpoint (returns a 1024-dimension vector where the old shape 400s).

mohammaddaoudfarooqi:
Fixes #19801

## What this fixes

Two problems block `@mastra/voyageai` from working against a provider-hosted Voyage endpoint (for example MongoDB Atlas's `https://ai.mongodb.com/v1`):

**1. Multimodal requests are rejected by the API.** `VoyageMultimodalEmbeddingModel.doEmbed` serialized text content to bare strings, producing `inputs: [["text"]]`. The Voyage multimodal API, and the SDK's own `MultimodalEmbedRequestInputsItem` type, require `inputs: [{ content: [{ type: 'text', text }] }]`. The old shape returns HTTP 400 ("Expected object. Received list"), so multimodal text embedding could not be used at all.

**2. No `baseUrl` passthrough.** The embedder, reranker, and contextualized models always constructed `new VoyageAIClient({ apiKey })`, with no way to point at a provider-hosted endpoint even though the underlying SDK accepts `baseUrl`.

## Changes

- `toSdkInput`/`toSdkContent` now emit `{ content: [{ type: 'text', text }, ...] }`. Text content is a typed object rather than a bare string.
- Added optional `baseUrl?: string` to all four Voyage config types (text, multimodal, contextualized, reranker) and pass it through with a conditional spread. When `baseUrl` is absent, the client is still constructed with exactly `{ apiKey }`, so existing behavior is byte-identical.
- Bumped the `voyageai` dependency from `^0.2.1` to `^0.3.1`. The 0.2.1 ESM build has a broken directory import (`export * from "../api"` with no `/index.js`) that Node's native ESM resolver rejects, which breaks any ESM app importing this package under `tsx` or `vitest`. 0.3.1 ships a working ESM build with the same client API. This surfaced while wiring the package into a live ESM app.
- Documented `baseUrl` in `models/embeddings.mdx` and the package README.

## Backward compatibility

`baseUrl` is optional and omitted from the client options when unset. The multimodal shape change corrects a payload the API always rejected, so no working call changes behavior. The `voyageai` bump is within a caret range and both 0.2.1 and 0.3.1 expose the same five client methods this package calls (`embed`, `multimodalEmbed`, `rerank`, `contextualizedEmbed`, `tokenize`).

## Testing

Unit tests assert the serialized multimodal request shape and the `baseUrl` present/absent construction. The patched embedder was also run against the live MongoDB-hosted Voyage endpoint and returned a 1024-dimension vector, where the old shape returned HTTP 400.

## One thing to flag

The `as any` cast at the `multimodalEmbed` call site stays, and it is deliberate: the SDK's generated request type uses camelCase keys for image and video content (`imageUrl`) while the wire API and this package's public `VoyageMultimodalContent` types use snake_case (`image_url`). Text content is fully typed. Aligning the image/video key casing to remove that cast is a reasonable follow-up; I kept it out of scope so this PR stays focused on the text-400 bug and `baseUrl`.

### DIFF
--- a/.changeset/deep-eggs-invite.md
+++ b/.changeset/deep-eggs-invite.md
@@ -1,0 +1,23 @@
+---
+'@mastra/voyageai': minor
+'@mastra/deployer': patch
+'@mastra/core': patch
+'@mastra/mcp': patch
+---
+
+Fixed VoyageAI multimodal embeddings sending text as bare strings (the API rejected the payload) and added a `baseUrl` option so you can point the embedder, reranker, and contextualized models at a provider-hosted Voyage endpoint.
+
+**Before**
+
+```ts
+// multimodal text was serialized as inputs: [["text"]] -> HTTP 400
+```
+
+**After**
+
+```ts
+const embedder = voyage.multimodalEmbedding({
+  model: 'voyage-multimodal-3.5',
+  baseUrl: 'https://ai.mongodb.com/v1',
+});
+```
--- a/.changeset/deep-eggs-invite.md
+++ b/.changeset/deep-eggs-invite.md
@@ -1,8 +1,5 @@
 ---
 '@mastra/voyageai': minor
-'@mastra/deployer': patch
-'@mastra/core': patch
-'@mastra/mcp': patch
 ---
 
 Fixed VoyageAI multimodal embeddings sending text as bare strings (the API rejected the payload) and added a `baseUrl` option so you can point the embedder, reranker, and contextualized models at a provider-hosted Voyage endpoint.
--- a/docs/src/content/en/models/embeddings.mdx
+++ b/docs/src/content/en/models/embeddings.mdx
@@ -81,6 +81,7 @@ const customModel = voyageEmbedding({
   model: "voyage-3.5",
   inputType: "query", // or 'document'
   outputDimension: 512, // 256, 512, 1024, or 2048
+  baseUrl: "https://ai.mongodb.com/v1", // Optional: custom endpoint (e.g. MongoDB-hosted Voyage)
 });
 
 const { embeddings: customEmbeddings } = await customModel.doEmbed({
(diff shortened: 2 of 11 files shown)


mohammaddaoudfarooqi:
Superseded by #19804 (same branch, rebased onto current main, now with a linked issue #19801).

PR #19796, opened 2026-07-20
test(core): cover storage lazy-init rejection handling

mohammaddaoudfarooqi:
## What this adds

Two regression tests in `packages/core/src/storage/storageWithInit.test.ts` that lock in the storage lazy-init behavior from #16427 and #19697:

1. When init fails and the caller awaits its own storage call, no rejection floats from the internal cached init promise.
2. Constructing an augmented store and never calling anything fires no background init, so nothing floats.

## Why

An app that builds a `MongoDBStore` against an unreachable or placeholder URI used to see a process-level unhandled rejection from the background `createDefaultIndexes()` path after its test run finished. The fix landed upstream; these tests guard against regressing it. This is the exact failure that forced a downstream `unhandledRejection` filter as a workaround.

## Note on test design

These do not use a caller-side fire-and-forget (`void store.method()`). In JavaScript, `void <rejecting-async-call>()` always produces an `unhandledRejection` regardless of library code, because the caller discarded the promise, so asserting otherwise would be impossible to satisfy. The tests assert the properties the fix actually guarantees.

No production code changes, no changeset. Open only if the extra coverage is wanted; otherwise it documents that the fix holds.

### DIFF
--- a/packages/core/src/storage/storageWithInit.test.ts
+++ b/packages/core/src/storage/storageWithInit.test.ts
@@ -281,5 +281,43 @@ describe('augmentWithInit', () => {
       expect(init).toHaveBeenCalledTimes(2);
       expect(listMessages).toHaveBeenCalledTimes(3);
     });
+
+    it('does not float a rejection from the internal cache when init fails and the caller handles its call', async () => {
+      const unhandled: unknown[] = [];
+      const on = (r: unknown) => unhandled.push(r);
+      process.on('unhandledRejection', on);
+      try {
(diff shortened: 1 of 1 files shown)


mohammaddaoudfarooqi:
Closing: this is test-only coverage guarding the already-merged storage lazy-init fix (#16427, #19697), with no user-facing change. Keeping it out of the review queue. The branch is preserved and can be reopened if the storage-init behavior is revisited upstream.

PR #19797, opened 2026-07-20
feat(mongodb): let MongoDBVector index an existing operational collection

### LINKED ISSUE #19802: MongoDBVector: index an existing operational collection (bring-your-own-collection)
## Problem

`MongoDBVector` owns its collection: the collection name equals the index name, and vectors are stored in a managed collection with a `{ metadata, document, embedding }` shape. Applications that already have an operational collection (their own documents, their own `_id`s and top-level fields) cannot put a vector index on those same documents. They have to duplicate data into a separate managed collection.

A common ask for MongoDB users is "same documents, one collection": the vector index lives on the operational collection the app already reads and writes.

## Proposal

Add an opt-in bring-your-own-collection (BYO) mode to `MongoDBVector`, leaving the managed path unchanged.

```ts
// Index an existing operational collection
await store.createIndex({
  indexName: 'precedents',
  dimension: 1024,
  collectionName: 'transactions',       // existing operational collection
  searchIndexName: 'transactions_vec',  // optional, controls the Atlas index name
  filterFields: ['status'],
});

// Query it, getting the full source document back
const hits = await store.query({
  indexName: 'precedents',
  queryVector,
  topK: 5,
  metadataMode: 'document',             // returns the source doc as metadata, filters on root fields
  filter: { status: { $in: ['approved', 'rejected'] } },
});
```

Key points:

- `collectionName` / `searchIndexName` are optional; omitting them is the current managed behavior.
- `metadataMode: 'document'` returns the full source document as `metadata` and filters on root fields; `'field'` (default) keeps the managed `metadata.<field>` projection.
- Native `ObjectId` `_id`s are supported (results coerce to the string `QueryResult.id`; delete/update accept that string).
- `deleteIndex` on a BYO collection drops only the search index, never the caller's collection.

## Backward compatibility and open questions

Fully backward compatible: every new parameter is optional and defaults to today's behavior. Two things need a maintainer's decision, both flagged in the PR: `listIndexes()` changes from "all collection names" to "logical index names," and BYO classification is persisted in a new store-owned `__mastra_vector_indexes__` collection so it survives a process restart.

## Status

Implemented and validated end to end against a live Atlas cluster and inside a real application (provision, seed with real embeddings, query). PR is ready to link once this feature request has maintainer feedback (was #19797).

mohammaddaoudfarooqi:
Fixes #19802

## Motivation

Today `MongoDBVector` owns its collection: the collection name equals the index name, and vectors live in a managed collection. A common need is for the vector index to live on the same operational documents the app already reads and writes, rather than a separate managed copy. This PR adds an opt-in bring-your-own-collection (BYO) mode and keeps the managed path unchanged.

## What's new

- `createIndex({ indexName, dimension, collectionName?, searchIndexName? })`. `collectionName` points the index at an existing operational collection (defaults to `indexName`); `searchIndexName` overrides the derived vector-index name. A BYO collection is never created or dropped by the store.
- `query({ ..., metadataMode?: 'field' | 'document' })`. `'document'` returns the full source document as `metadata` (via `$$ROOT`) for operational docs that have their own shape; the embedding field is omitted by default and included when `includeVector` is set. `'field'` (the default) is the existing managed projection.
- In `'document'` mode, filters operate on root document fields, so `filter: { status: {...} }` matches a top-level `status`. In `'field'` mode, bare fields are still rewritten to `metadata.<field>`.
- Native `ObjectId` `_id`s are supported: query results coerce `_id` to a string (the `QueryResult.id` contract), and `deleteVector`/`updateVector`/`deleteVectors` accept that string and match the underlying `ObjectId` document.
- Durable index registry. Each logical index's target (collection name, search-index name, `isByo`, dimension, metric) is persisted to a store-owned `__mastra_vector_indexes__` collection and hydrated on demand, so classification and routing survive a process restart. Without this, a fresh process would misclassify a BYO index and `deleteIndex` could drop the caller's operational collection. `createIndex` persists the entry before provisioning the search index, so a provisioning failure cannot leave a BYO collection unprotected.
- `deleteIndex` reads the persisted `isByo`: for BYO it drops only the Atlas search index, never the collection; for managed it drops the collection as before.
- `listIndexes()` returns logical index names (registered indexes, plus managed collections that have a `${name}_vector_index` for back-compat).

## Backward compatibility

A caller who never passes `collectionName` sees identical behavior: `resolveIndexTarget` returns the managed default (`isByo: false`), `createIndex` still creates the collection, `deleteIndex` still drops it. Every new parameter is optional and defaults to the prior behavior.

Two behavior notes worth a maintainer's eye:

- `listIndexes()` previously returned every collection name in the database. It now returns logical index names, so it no longer lists unrelated non-vector collections that happen to share the database. For a managed-only user the vector index names are the same.
- `createIndex`/`createSearchIndex` write to a new `__mastra_vector_indexes__` collection. It is excluded from `listIndexes()` and never touches caller data, but it is a new artifact in the database.

## Testing

Docker atlas-local integration: createIndex on an operational collection (no managed collection created), document-mode query returns the full source doc as metadata, describeIndex, and deleteIndex preserving the operational collection and documents while dropping only the search index. Mocked unit tests cover the BYO-vs-managed delete classification, including the `collectionName === indexName` edge case.

Also verified against a live Atlas cluster: one instance creates a BYO index, then a fresh instance (a simulated process restart) sees the logical index name via `listIndexes()`, and `deleteIndex` preserves all operational documents while dropping only the search index. The full flow was then run inside a real app on live Atlas: provision, seed with real embeddings, and query all passed.

## Open questions for review

- Parameter naming: `collectionName` / `searchIndexName` / `metadataMode: 'field' | 'document'`. Happy to match conventions you prefer.
- `document` mode returns the full source doc as `metadata`; it excludes the embedding by default to avoid large payloads.
- The registry collection is named `__mastra_vector_indexes__`. Open to a different name or namespacing.

### DIFF
--- a/.changeset/dry-emus-create.md
+++ b/.changeset/dry-emus-create.md
@@ -1,8 +1,5 @@
 ---
 '@mastra/mongodb': minor
-'@mastra/deployer': patch
-'@mastra/core': patch
-'@mastra/mcp': patch
 ---
 
 MongoDBVector can now index an existing (operational) collection instead of a managed one. Pass `collectionName` (and optionally `searchIndexName`) to `createIndex`, and query with `metadataMode: 'document'` to get the full source document back as metadata.
--- a/.changeset/dry-emus-create.md
+++ b/.changeset/dry-emus-create.md
@@ -1,0 +1,15 @@
+---
+'@mastra/mongodb': minor
+'@mastra/deployer': patch
+'@mastra/core': patch
+'@mastra/mcp': patch
+---
+
+MongoDBVector can now index an existing (operational) collection instead of a managed one. Pass `collectionName` (and optionally `searchIndexName`) to `createIndex`, and query with `metadataMode: 'document'` to get the full source document back as metadata.
+
+**Example**
+
+```ts
+await vectorStore.createIndex({ indexName: 'precedents', dimension: 1024, collectionName: 'transactions' });
+const hits = await vectorStore.query({ indexName: 'precedents', queryVector, metadataMode: 'document' });
+```
--- a/docs/src/content/en/reference/vectors/mongodb.mdx
+++ b/docs/src/content/en/reference/vectors/mongodb.mdx
@@ -122,6 +122,20 @@ Creates a new vector index (collection) in MongoDB.
     description:
       'Metadata field names to declare as filter fields in the Atlas vectorSearch index (registered as `metadata.<field>`). Queries that filter only on declared fields are pushed directly into `$vectorSearch` instead of pre-filtering candidate `_id`s, avoiding the 16 MB BSON limit on large result sets. Filters that reference an undeclared field, or use an operator `$vectorSearch` does not support, fall back to the pre-filter automatically.',
   },
+  {
+    name: 'collectionName',
+    type: 'string',
+    isOptional: true,
+    description:
+      'Store the vectors on an existing (operational) collection instead of a managed collection named after the index. The collection is never created or dropped by this store when set. Defaults to `indexName`.',
+  },
+  {
+    name: 'searchIndexName',
+    type: 'string',
+    isOptional: true,
+    description:
+      'Name for the Atlas vectorSearch index created on the collection. Defaults to `${indexName}_vector_index`.',
+  },
 ]}
 />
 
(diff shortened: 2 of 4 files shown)


mohammaddaoudfarooqi:
Note: #19798 (hybrid + full-text search) is stacked on this branch and should be reviewed after this one.

mohammaddaoudfarooqi:
Superseded by #19805 (same branch, rebased onto current main, now with a linked issue #19802).

PR #19798, opened 2026-07-20
feat(mongodb): add createSearchIndex, textQuery, and hybridQuery to MongoDBVector

### LINKED ISSUE #19803: MongoDBVector: full-text and hybrid search (createSearchIndex, textQuery, hybridQuery)
## Problem

`MongoDBVector` supports `$vectorSearch` only. MongoDB also offers full-text search (`$search`/BM25) and server-side hybrid fusion (`$rankFusion`), which are two of its more useful retrieval features and are not reachable through the store today. Applications that want lexical or hybrid retrieval have to drop to the raw driver.

## Proposal

Add three methods to `MongoDBVector`:

```ts
// Provision a full-text (BM25) index
await store.createSearchIndex({ indexName: 'precedents', fields: ['note'] });

// Full-text search
const textHits = await store.textQuery({ indexName: 'precedents', query: 'shell company', paths: ['note'], topK: 10 });

// Hybrid: vector + text fused server-side with $rankFusion
const hybridHits = await store.hybridQuery({
  indexName: 'precedents',
  queryVector,
  query: 'shell company',
  paths: ['note'],
  topK: 10,
});
```

- `hybridQuery` fuses a `$vectorSearch` and a `$search` pipeline with `$rankFusion` and reads the fused RRF score from `{ $meta: 'score' }`.
- Requires MongoDB 8.0 or later. On 8.0.x it may need a MongoDB support case to enable; where enabled, such as Atlas 8.0.x, it runs. The store gates at 8.0+ and lets the server be the final authority.
- `createSearchIndex` supports a field-restricted mapping and a custom index name, both persisted so `textQuery`/`hybridQuery` resolve them. `waitForSearchIndexReady` blocks until the async index build reports READY.

## Dependency and open question

This builds on the bring-your-own-collection work (separate feature request / PR), reusing its index registry and document mode. It should be reviewed after that lands.

One naming question for the maintainers: `hybridQuery` and `textQuery` mirror MongoDB's own `query`, while the Chroma store uses `hybridSearch`. Happy to match whichever convention you prefer.

## Status

Implemented and validated on docker atlas-local 8.2 and against a live Atlas 8.0.27 cluster, then inside a real application (vector, text, and hybrid retrieval all return correct results). PR is ready to link once this feature request has maintainer feedback (was #19798).

mohammaddaoudfarooqi:
> Stacked on #19805 (`mongodb/vector-byo-collection`). Review that one first. This branch contains its commits too, so until #19805 merges the diff below also shows the BYO changes; the hybrid/text work is the top four commits.

Fixes #19803

## Motivation

`@mastra/mongodb` supported `$vectorSearch` only. This adds MongoDB's full-text (`$search`/BM25) and server-side hybrid fusion (`$rankFusion`).

## What's new

- `createSearchIndex({ indexName, fields?, searchIndexName? })` provisions an Atlas Search (BM25) index on the index's collection. With `fields`, a field-restricted mapping is created under a name unique to the logical index, so two logical indexes on one collection do not collide.
- `textQuery({ indexName, query, paths, topK?, filter?, metadataMode? })` runs a BM25 `$search`, reusing the existing filter translator and projection.
- `hybridQuery({ indexName, queryVector, query, paths, topK?, filter?, weights?, numCandidates?, metadataMode? })` fuses a `$vectorSearch` and a `$search` pipeline server-side with `$rankFusion`. It reads the fused RRF score from `{ $meta: 'score' }`. The vector branch reuses `query()`'s declared-paths pushdown-vs-fallback, so undeclared-field filters on BYO collections do not hard-error.
- Version guard for `$rankFusion`: requires MongoDB 8.0 or later, memoized after one `buildInfo` probe. On 8.0.x it may need a MongoDB support case to enable; where enabled, such as Atlas 8.0.x, it runs, and the server is the final authority.
- The text-search-index name is persisted, so a custom name set at `createSearchIndex` is reachable by `textQuery`/`hybridQuery`.
- `waitForSearchIndexReady()` (and a `waitUntilReady` option on `createSearchIndex`) blocks until the full-text index reports READY, since it builds asynchronously and an immediate query can otherwise miss it.

## Backward compatibility

Additive. Existing `query`/`createIndex`/etc. behavior is unchanged. A shared `buildProjection` helper was factored out of `query` and reused; the field-mode projection is byte-identical to before.

## Testing

On docker atlas-local 8.2 (which supports `$rankFusion`): `createSearchIndex`, `textQuery`, and `hybridQuery` all pass, with the hybrid test asserting rank-ordered positive RRF scores. Against a live Atlas 8.0.27 cluster: `textQuery` verified, and `hybridQuery` runs there too (the guard admits 8.0+). The whole path was then exercised inside a real app on live Atlas: vector, text, and hybrid retrieval all returned correct results.

The `$rankFusion` guard started at 8.1 and was lowered to 8.0 after the live run showed it running on a real Atlas 8.0.27 cluster. A version-boundary test covers this (8.0.x, 8.1, 9.0 admitted; 7.x and 6.0 rejected).

## Open question for review

Method naming: `hybridQuery` and `textQuery` mirror MongoDB's own `query`. Chroma's store uses `hybridSearch`. Happy to rename to match whichever convention you prefer.


mohammaddaoudfarooqi:
Recreating as a ready-for-review PR from the same (rebased) branch; this draft could not be flipped to ready. See the replacement below.

mohammaddaoudfarooqi:
Consolidated into #19805, which now contains this branch's full history (the two changes are safety-interdependent: the durable index registry added here is what makes BYO deletion restart-safe in #19805). Additions since this draft: BYO collections are now **read-only by default** with an explicit `allowWrites` opt-in persisted in the registry (addressing the concern that Mastra must not modify a user's original collection), a circular-metadata guard that prevents a bson@7.x `calculateObjectSize` process hang, ObjectId-seeded hybrid/text integration tests, and small review cleanups. Validated green: 217 vector tests on docker atlas-local 8.2, plus 23/23 end-to-end checks on a live Atlas 8.0.28 cluster. Closing this draft; please review #19805.

PR #19805, opened 2026-07-20
feat(mongodb): index existing operational collections (read-only by default) + full-text & hybrid search ($rankFusion)

### LINKED ISSUE #19802: MongoDBVector: index an existing operational collection (bring-your-own-collection)

### LINKED ISSUE #19803: MongoDBVector: full-text and hybrid search (createSearchIndex, textQuery, hybridQuery)

mohammaddaoudfarooqi:
Fixes #19802
Fixes #19803

> **Note:** this PR consolidates #19805 and #19798 (previously stacked) into one reviewable unit, since the two changes are safety-interdependent: the durable index registry introduced for hybrid search is also what makes bring-your-own-collection deletion safe across process restarts. #19798 is closed in favor of this PR.

## Motivation

Two closely related gaps in `@mastra/mongodb`:

1. **`MongoDBVector` owns its collection.** Apps with an existing operational collection (their own documents, `ObjectId` `_id`s, top-level fields) cannot put a vector index on those documents — they must duplicate data into a managed copy (#19802).
2. **`$vectorSearch` only.** MongoDB's full-text (`$search`/BM25) and server-side hybrid fusion (`$rankFusion`) are unreachable through the store (#19803).

### Bring-your-own collection (BYO) — opt-in, read-only by default

- `createIndex({ indexName, dimension, collectionName?, searchIndexName?, allowWrites? })`. `collectionName` points the index at an existing operational collection; it is **never created or dropped** by the store.
- **The store never mutates caller-owned documents by default.** `upsert` / `updateVector` / `deleteVector` / `deleteVectors` throw a clear USER-category error on a BYO index unless it was created with `allowWrites: true`. Managed collections are unaffected (always writable). This is a deliberate answer to the design concern that Mastra should not modify or delete a user's original data collection: BYO mode out of the box is *indexing + querying* only, and the write path is an explicit, persisted opt-in.
- **Durable index registry.** Each logical index's target (collection, search-index names, `isByo`, `allowWrites`, dimension, metric) is persisted in a store-owned `__mastra_vector_indexes__` collection and hydrated on demand, so classification survives a process restart: a fresh process's `deleteIndex` drops only the Atlas search indexes, never the caller's collection. The entry is persisted *before* provisioning, so a provisioning failure cannot leave a BYO collection unprotected, and hydration **fails closed** (a BYO entry without a write policy is treated as read-only).
- `query({ ..., metadataMode: 'field' | 'document' })`. `'document'` returns the full source document as `metadata` (embedding omitted unless `includeVector`), and filters operate on **root** document fields. `'field'` (default) is the existing managed projection, byte-identical to before.
- Native `ObjectId` `_id`s: results coerce `_id` to the `QueryResult.id` string contract, and (opt-in) write ops accept that string and match the underlying `ObjectId` document.
- `listIndexes()` returns **logical** index names (registered indexes plus registry-less managed collections for back-compat), so its output round-trips into `deleteIndex`/`describeIndex`.

### Full-text and hybrid search

- `createSearchIndex({ indexName, fields?, searchIndexName?, waitUntilReady? })` provisions an Atlas Search (BM25) index. On BYO collections this is **opt-in** (no billable index is auto-created); field-restricted mappings get a distinct per-logical-index name so two logical indexes on one collection don't collide.
- `textQuery({ indexName, query, paths, topK?, filter?, metadataMode? })` runs a BM25 `$search`, reusing the existing filter translator and projections.
- `hybridQuery({ indexName, queryVector, query, paths, topK?, filter?, weights?, numCandidates?, metadataMode? })` fuses `$vectorSearch` + `$search` server-side with `$rankFusion` (RRF score read from `{ $meta: 'score' }`). Version-gated at MongoDB >= 8.0 (memoized `buildInfo` probe; verified working on live Atlas 8.0.x). The vector branch reuses `query()`'s pushdown-vs-fallback filter logic, so undeclared-field filters on BYO collections don't hard-error.
- `waitForSearchIndexReady()` (and `waitUntilReady` on `createSearchIndex`) blocks until the async text-index build reports READY.

### Robustness fix surfaced by testing

`upsert`/`updateVector` now reject metadata containing **circular references** with a clear USER error. Without this, `bson@7.x`'s iterative `calculateObjectSize` (no cycle detection) spins forever at 100% CPU inside `bulkWrite` — a process hang, not an exception (reproduced standalone against bson 7.3.1). Legal shared (DAG) references are still allowed.

## Backward compatibility

A caller who never passes `collectionName` sees identical behavior: managed `createIndex` still creates the collection and companion dynamic text index, `deleteIndex` still drops it, `query` projections are unchanged, writes need no opt-in. All new parameters are optional. Two notes for a maintainer's eye:

- `listIndexes()` previously returned every collection in the database; it now returns logical index names only (managed vector collections still appear).
- The store persists registrations to a new `__mastra_vector_indexes__` collection (excluded from `listIndexes()`, never touches caller data).

## Testing

- **Local (docker atlas-local 8.2 + mongod 7.0):** full package suite green — 217 vector tests (unit + integration: BYO lifecycle, write-guard refusal/opt-in/fail-closed hydration, restart durability across instances, logical `listIndexes`, document-mode projection, ObjectId `_id` coercion, `createSearchIndex`/`textQuery`/`hybridQuery` with rank-ordered positive RRF scores, circular-metadata guard) plus 620 storage/filter tests.
- **Live Atlas (8.0.28):** end-to-end validation of the exact flows above against a real cluster — BYO index on an ObjectId-keyed operational collection, write guard enforced same-instance and from a fresh instance (simulated restart), `textQuery` BM25, `hybridQuery` `$rankFusion` fusion with correct winner and positive RRF scores, `deleteIndex` preserving all operational documents while dropping both search indexes and clearing the registry, and unchanged managed-path behavior. 23/23 checks passed.
- The text/hybrid integration tests seed **native `ObjectId` `_id`s** (what real operational collections use), so the id-coercion contract is exercised end to end.

## Open questions for review

- Naming: `collectionName` / `searchIndexName` / `allowWrites` / `metadataMode: 'field' | 'document'`; `textQuery`/`hybridQuery` mirror `query` (Chroma uses `hybridSearch`). Happy to match preferred conventions.
- The registry collection is named `__mastra_vector_indexes__`; open to different namespacing.
- If maintainers prefer BYO writes to be removed entirely rather than opt-in, the guard makes that a one-line change.

## ELI5

MongoDBVector can now “plug into” MongoDB collections you already have (BYO), so it won’t copy or risk deleting your real data. It also adds safer search (BM25 + hybrid vector/text) and smarter controls for what it returns and whether it’s allowed to write to your collections.

## What changed

- Added opt-in BYO collection support via `createIndex({ collectionName, searchIndexName? })` (managed behavior unchanged when omitted).
- Made BYO collections **read-only by default**: `upsert`, `updateVector`, `deleteVector(s)` throw unless the index was created with `allowWrites: true`; the write policy is persisted so it “fails closed” after restarts.
- Added durable, restart-safe index classification/routing via a Mastra-owned registry (rehydrates BYO vs managed behavior and policy on demand).
- Added `metadataMode`:
  - `'field'` (default) keeps returning managed metadata fields,
  - `'document'` returns the full source document as `metadata` (while omitting the embedding from `metadata` by default; keep it with `includeVector: true`),
  - avoids clobbering a real source field named `score`.
- In `'document'` mode, metadata filters apply to **root document fields** (no `metadata.` prefix); `'field'` mode keeps `metadata.<field>` semantics.
- Added native MongoDB **`ObjectId`** support for BYO `_id`s while keeping `QueryResult.id` as a **string**, including correct matching for vector mutations/deletes.
- Improved lifecycle safety:
  - `deleteIndex()` for BYO removes only the Atlas search index (and companion full-text index when applicable) and never drops caller operational collections/documents,
  - `listIndexes()` returns logical Mastra index names so results round-trip with `deleteIndex()`/`describeIndex()`,
  - prevents orphaning/retarget conflicts by rejecting attempts to reuse a logical index name for a different registered collection.
- Added Atlas full-text + hybrid retrieval:
  - `createSearchIndex` + `textQuery` for BM25,
  - `hybridQuery` using MongoDB 8.0+ `$rankFusion` (documented as potentially requiring a support case on 8.0.x),
  - introduced readiness handling via `waitForSearchIndexReady()` / `waitUntilReady` so immediate queries don’t intermittently fail.
- Added safety hardening for write operations: rejects circular metadata during upsert/update.
- Updated docs and examples to cover BYO indexing, `metadataMode` behavior (including root-field filtering and embedding omission rules), hybrid search requirements, and BYO delete semantics.
- Expanded unit/integration/live Atlas test coverage for BYO read-only/write opt-in, projection correctness, readiness polling, full-text/hybrid behavior, deletion safety, and restart durability.


mohammaddaoudfarooqi:
@coderabbitai Both findings were valid — fixed in cb8413b8f4:

1. **Field-mapped default index name**: the `createSearchIndex` JSDoc and the reference-doc prose now say `${collectionName}_${indexName}_search_fields_index`, matching the implementation (the name is unique per logical index so two logical indexes on one collection don't collide).
2. **`hybridQuery` version claim**: JSDoc now states MongoDB >= 8.0, matching `assertRankFusionSupported()`. Also swept the docs for the same stale claim and fixed three additional `>= 8.1` call-outs (reference `hybridQuery()` section, the version note, and `rag/vector-databases.mdx`), keeping the 8.0.x support-case caveat.

mohammaddaoudfarooqi:
@coderabbitai All five outside-diff findings were verified against the code and were valid — addressed in c27eca8eff:

1. **Persist text-index target before provisioning** — `createSearchIndex` now writes the complete registry entry (collection/vector-index/isByo/allowWrites + textSearchIndexName) *before* creating or updating the Atlas index, mirroring `createIndex`'s persist-before-provision ordering. A crash between the steps now leaves a registered (possibly not-yet-existing) index rather than an unregistered physical index leaked onto a caller-owned collection; a failed provision converges on retry since deletion is ignore-missing. Order asserted by test R5-4.

2. **Vector/text name collisions** — `createIndex` rejects a vector `searchIndexName` that collides with the companion text-index name (which would otherwise get silently swallowed as IndexAlreadyExists), and `createSearchIndex` rejects a text name equal to the vector index name (which would otherwise *overwrite the vectorSearch definition* via `updateSearchIndex`). For shared defaults, `deleteIndex` now preserves any physical index still referenced by a sibling registry entry on the same collection (tests R5-3/3b/5).

3. **`upsert` ObjectId duplication** — upsert now uses dual-form `_id` matching (string + ObjectId for 24-hex ids) with `$setOnInsert` pinning the requested string id when no document matches (an operator filter can't infer `_id` on insert), so opted-in BYO upserts update the existing ObjectId document instead of inserting a string-keyed duplicate. Plain string ids keep the exact previous equality-filter behavior. Unit tests R5-7/7b plus a live-collection integration regression.

4. **Externally dropped BYO collection** — root cause is that `getCollection(..., false)` always returns a handle, so the drop path was reachable with a missing namespace. `dropSearchIndexIgnoringMissing` now also treats `NamespaceNotFound` (26) as success, so `deleteIndex` proceeds to clear the stale registry entry (test R5-6). I kept this narrower than the suggested existence pre-check: the ignore-missing semantics already make the flow idempotent without a TOCTOU window.

5. **Same-name BYO reclassification (critical)** — BYO classification is now **sticky**: an existing registry entry's `isByo` is preserved across idempotent re-`createIndex` calls (reclassifying requires an explicit `deleteIndex` first), and the write policy is preserved on refresh unless explicitly re-specified. For the registry-missing fallback, `deleteIndex` now **fails closed**: it only drops a collection that verifiably carries the legacy managed `${indexName}_vector_index` search index, so a repeated delete after a completed BYO deletion (registry entry gone) refuses instead of dropping the preserved caller collection (tests R5-1/1b/2).

Validation: full vector suite **228 passed / 0 failed** on docker atlas-local 8.2, and the end-to-end live-Atlas (8.0.x) validation of BYO lifecycle, write guard, text/hybrid ($rankFusion), and delete-safety flows passed **23/23** after these changes.

abhiaiyer91 (approved):

PR #19844, opened 2026-07-21
feat(mastracode-web): per-stage automation metrics for the Factory board

smthomas:
The Factory Metrics page now tells stakeholders how much of the board is actually automated, stage by stage. For each stage you see the share of completed passes that were fully automated — entered *and* exited by an automation actor, first visit, no human touch — and how those automated passes ended up: done, discarded, reworked, or still in flight. A global "Automated moves" stat surfaces the human-vs-total transition split that was already computed but never rendered.

To make that measurable, stage history entries now stamp `exitedBy` when a stage closes (entries written before this ship conservatively classify as human-exited, so automation % never over-reports), and there's a canonical automation actor classifier: the sentinel ids (`factory`/`system`/`automation`), the rules-engine system ids (`factory-rule-dispatcher`/`factory-tool-result-rule`), agent bindings (`agent:*`), and webhook actors (`github:*`) all count as automation — a `github:*` transition means a human acted on GitHub, but the board move itself was automated. All automation stage moves flow through the governed `FactoryTransitionService` (rules evaluation, idempotent ingress, revision checks); HTTP routes still always stamp the authenticated user, and no client can supply an actor.

The board also gains a `Discarded` terminal column so discard outcomes are trackable. Discarded items leave WIP/aging/queue-health like done items do, but never count toward throughput or cycle time (those stay done-only).

The `discarded` stage is part of the governed rules vocabulary (`FACTORY_RULE_STAGES`), so board drags to Discarded are accepted by the transition endpoint and discarded items keep a canonical stage (they can be revived back onto the board).
The metrics API response gains one additive field:

```jsonc
"stageAutomation": [
  { "stage": "triage", "exits": 4, "automated": 2,
    "outcomes": { "done": 1, "discarded": 0, "reworked": 1, "inFlight": 0 } }
]
```

Partial-board automation works out of the box — e.g. automated triage with human-approved planning reads as `triage: 50% automated, planning: 0%`. A pass automated at a stage but later sent back through it counts as `reworked` (deliberately, even if the item eventually merged — a redo is an automation failure).

Test plan: unit coverage for the classification rules (first-visit-only, missing `exitedBy`, window boundaries, all four outcomes, rework precedence), route tests asserting the field over HTTP, MSW tests for the new stat/section/empty states and the Discarded board column, plus a runnable proof demo that seeds a board through the real store functions and asserts the numbers served by the real metrics route. All existing suites pass unchanged (`transitions` shape is untouched; throughput/cycleTime stay done-only).

## ELI5

The Factory Metrics page can now spot when a work item moved out of a stage automatically (not by a human), and it summarizes what happened next. It also adds a “Canceled” terminal outcome so canceled work is handled correctly in metrics while still allowing items to be brought back later.

## Summary

- Added per-stage automation metrics via a new `stageAutomation` breakdown (stage exits, automated passes, and outcome buckets: done/canceled/reworked/in-flight).
- Added an “Automated moves” statistic derived from transition counts (human vs total).
- Renamed the terminal stage from “discarded” to **“canceled”** end-to-end, including board column labeling, rules vocabulary/identifiers, metrics outcomes, tests, and proof demo.
- Extended Factory stage history to record the exit actor using an optional `exitedBy` field; entries without `exitedBy` are treated as human-exited.
- Implemented automated-actor classification via `isAutomationActor(...)` and used it to compute automation vs human transitions; HTTP routes continue to use the authenticated user and reject client-supplied actor information.
- Updated metrics and queue-health logic to treat **done and canceled** as terminal for “in-flight/open” detection, while keeping throughput/cycle time done-only; canceled is excluded from overall in-flight/WIP-aging-style aggregates but still appears in per-stage WIP counts, and canceled items can be revived.
- Expanded UI: the metrics API type includes `stageAutomation`, the Metrics page renders an “Automation by stage” section, and terminal stages are omitted from the per-stage list display.
- Added/expanded test coverage across compute logic, API/routes, MSW/UI, storage unit tests, queue-health behavior, and proof-demo-related rendering.


mastracode/web/src/web/factory/rules/types.ts
```diff
@@ -1,6 +1,6 @@
 export type WorkItemSource = 'github-issue' | 'github-pr' | 'linear-issue' | 'manual';
 
-export const FACTORY_RULE_STAGES = ['intake', 'triage', 'planning', 'execute', 'review', 'done'] as const;
+export const FACTORY_RULE_STAGES = ['intake', 'triage', 'planning', 'execute', 'review', 'done', 'discarded'] as const;
```
TylerBarnes: not a big deal either way but I wonder if it'd be better to call it "canceled" instead of discarded

TylerBarnes (approved):
looks good! one optional comment/question but LGTM

PR #19848, opened 2026-07-21
fix(memory): filter cross-thread messages before title generation

### LINKED ISSUE #19833: Thread title generation uses entire messageList instead of the computed most-recent user message (cross-thread contamination with resource-scoped memory)
## Package/version

`@mastra/core` `1.51.0` (dist bundle), interacting with resource-scoped memory (`@mastra/memory` `1.23.0`).

## Expected behavior

With `generateTitle` enabled, a brand-new thread's title should be derived from that thread's own first user message.

## Actual behavior (with source evidence)

In dist `chunk-77JDY5O7.js`:

- The post-response handler (lines ~56213–56230) collects `uiMessages = messageList.get.all.ui()` — the FULL message list — and calls `this.genTitle(userMessage, ..., uiMessages)`.
- `genTitle` (lines ~53351–53365) ignores the passed single `userMessage` whenever `uiMessages.length > 0` and calls `generateTitleFromUserMessage({ messages: uiMessages, ... })` instead — discarding the most-recent-user-message it just computed.
- `generateTitleFromUserMessage` (lines ~53293–53298) then feeds `formatMessagesForTitle(messages)` (lines ~53253–53268), which concatenates ALL messages (every role) into one blob as the "user content" for the title model.

Under resource-scoped memory the message list of a fresh thread legitimately contains raw history from OTHER threads of the same resource (that is the point of resource scope). The title model therefore sees mostly foreign-thread content, and the new thread gets titled after an unrelated older conversation. Reproduced twice live: a new thread whose only message was about topic A received a title in a different language about topic B from a prior thread.

## Impact (sanitized)

Any deployment combining `generateTitle: true` with resource-scoped memory produces visibly wrong sidebar titles for new threads — titles describe old, unrelated conversations (including wrong-language titles). Cross-thread messages retain their original `threadId` on the message objects, so the correct source message is identifiable at the call site.

## Suggested direction

In `genTitle`, prefer the computed most-recent user message (or filter `uiMessages` to the active thread via each message's `threadId`) before calling `generateTitleFromUserMessage`, so titles are derived from the thread being titled. Downstream we currently work around this by disabling native `generateTitle` for resource scope and running our own selection with the same `generateTitleFromUserMessage` API.

unfitcoder101:
## Summary
With resource-scoped memory, `messageList.get.all.ui()` returns messages from ALL threads of the same resource. `genTitle` was receiving the full unfiltered list and using foreign-thread messages to generate titles — causing new threads to get titled after unrelated older conversations (including wrong-language titles).

Fix: filter `uiMessages` to the active `threadId` before passing to `genTitle`. Messages without a `threadId` pass through unchanged.

Fixes #19833

## Why
The issue is in `agent.ts` at the call site:
```ts
// Before
const uiMessages = messageList.get.all.ui();

// After  
const allUiMessages = messageList.get.all.ui();
const uiMessages = allUiMessages.filter((m: any) => !m.threadId || m.threadId === thread.id);
```

## Testing
Added regression test covering cross-thread contamination scenario with resource-scoped memory.

## Checklist
- [x] Regression test added
- [x] Lint passes

## ELI5

This PR keeps messages from different conversations from getting mixed up when naming a new conversation. It also safely handles unusual data when saving dataset items, preventing crashes and confusing storage errors.

## What changed

- Filtered title-generation messages to the active thread while retaining messages without a `threadId`.
- Added regression coverage for resource-scoped cross-thread title generation.
- Added dataset payload validation for circular, lossy, non-JSON-safe, and non-plain-object values, returning HTTP 400 for invalid requests.
- Added the public `safeStringify` utility and used it across playground trace, score, and tool-mock editors to safely serialize circular data.
- Ensured dataset item writes persist only caller-provided `requestContext` entries, excluding internal server keys.
- Added tests covering dataset validation, request context handling, and safe serialization.

### DIFF
--- a/.changeset/great-rockets-judge.md
+++ b/.changeset/great-rockets-judge.md
@@ -1,0 +1,5 @@
+---
+'@mastra/server': patch
+---
+
+Fixed dataset item endpoints to return HTTP 400 when payloads contain circular values, silently lossy JSON values (nested `undefined`, functions, symbols, bigints, and non-finite numbers), or non-plain objects (`Date`, `Map`, `Set`, class instances, and custom `toJSON()` objects) that cannot be serialized faithfully. Also fixed the add/update dataset item endpoints to persist the caller-provided `requestContext` entries instead of the live server request context instance, which contained internal server state and could fail storage serialization.
--- a/.changeset/rare-waves-argue.md
+++ b/.changeset/rare-waves-argue.md
@@ -1,0 +1,15 @@
+---
+'@mastra/core': patch
+---
+
+Fixed dataset item writes to reject circular payloads with a clear validation error before storage, instead of failing with database-specific serialization errors. Silently lossy JSON values (nested `undefined`, functions, symbols, bigints, and non-finite numbers) and non-plain objects (`Date`, `Map`, `Set`, class instances, and custom `toJSON()` objects) are now also rejected with the offending path, so identical `externalId` retries can no longer conflict with the persisted payload. Added a public `safeStringify` utility for serializing values that may contain circular references.
+
+```ts
+import { safeStringify, ensureSerializable } from '@mastra/core/utils/safe-stringify';
+
+const value: Record<string, unknown> = { prompt: 'hello' };
+value.self = value;
+
+safeStringify(value); // '{"prompt":"hello","self":"[Circular]"}'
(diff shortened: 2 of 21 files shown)


intojhanurag (changes requested):
it is pretty unrelated changes , we are gonna close . Thanks !

PR #19851, opened 2026-07-21
feat: add typed stored workflow access for Studio

NikAiyer:
Adds the typed client and React Query boundary needed for persisted workflow authoring in Studio.

`@mastra/client-js` now exposes list, get, upsert, and delete operations for stored workflow definitions using the generated server contracts. Studio gets MSW-tested hooks for the same lifecycle and keeps both persisted-definition queries and the canonical runtime workflow cache synchronized after mutations.

This is an intentionally early draft stacked on `dynamic-workflow-steps`. The editor-owned builder capability and Studio authoring pages are not implemented yet.

Test plan:

- `pnpm --filter @mastra/client-js generate:route-types`
- `pnpm --filter @mastra/client-js exec vitest run src/resources/stored-workflow.test.ts --reporter=dot --bail=1`
- `pnpm --filter @mastra/client-js build:lib`
- `pnpm --filter @internal/playground exec vitest run src/domains/workflows/hooks/use-stored-workflows.msw.test.tsx --reporter=dot --bail=1`
- `pnpm --filter @internal/playground typecheck`


NikAiyer:
Upstream ownership note for the persisted-workflow work stacked on #19492:

The following Core defects were discovered while exercising Workflow Builder, but belong to the `dynamic-workflow-steps` persistence/runtime layer rather than Studio:

- Declarative agent execution used registration-key lookup for persisted canonical `agentId` values. Fixed on this branch by resolving through `getAgentById(...)`, with default/evented coverage where the registration key differs from the agent ID (`c19cc6c6ab`).
- Persisted mapping descriptors treated workflow-local source step IDs as global tool/agent dependency IDs during rehydration. Fixed by resolving mappings against previously rehydrated local steps, with tool and agent regressions where local IDs differ from dependency IDs (`0d614181d9`).
- Persisted parallel/conditional child reconstruction did not preserve local entry IDs separately from dependency IDs. Covered and fixed in the executable-draft remediation (`9f9d35fe00`).
- After the latest `SingleStepEntry` refactor, persisted nested workflow call-site IDs cannot be preserved separately from the referenced workflow ID. This branch now rejects `id !== workflowId` recursively during normalization, Studio finalization, and strict Save/publication, with guidance to use the workflow ID for both fields. This is a compatibility guard for the current base runtime, not a Studio product requirement.

Base PR: #19492 (`dynamic-workflow-steps`, Tony/rase-). These should be reviewed for upstreaming there so the runtime/storage invariants do not depend on the Studio stack.

Issue #19317, opened 2026-07-12
[BUG] @mastra/inngest ignores toolCallConcurrency — tool calls always run sequentially on the Inngest engine


devangpadhiyar:
### Describe the Bug

The `@mastra/inngest` durable execution engine ignores `toolCallConcurrency` and always
executes an agent's parallel tool calls sequentially, even when the agent has no
suspending/approval tools.

Root cause: the Inngest engine has its own copy of the agentic-execution workflow
(`createInngestDurableAgenticWorkflow`) which builds the tool-call `foreach` WITHOUT a
concurrency option:

  node_modules/@mastra/inngest/dist/index.js
    ).foreach(toolCallStep).map(...)     // ← no options → foreach defaults to concurrency: 1

Compare with @mastra/core's `createAgenticExecutionWorkflow`, which correctly passes it:

  packages/core/src/loop/workflows/agentic-execution/index.ts
    .foreach(toolCallStep, { concurrency })   // concurrency = requireToolApproval ? 1 : (toolConcurrency ?? 10)

This is the same bug that was fixed for @mastra/core in #9704 (merged Dec 2025), but the
fix was never ported to the @mastra/inngest engine's duplicate workflow. As a result, a
`createInngestAgent` agent runs every tool call one-at-a-time regardless of
`toolCallConcurrency` / `defaultOptions.toolCallConcurrency`.

I verified with the trace: N tool calls in one turn each become a separate `durable-tool-call`
step, but they execute back-to-back (each starts exactly when the previous ends) instead of
overlapping.

### Steps To Reproduce

1. Create a durable agent with createInngestAgent({ agent, inngest }) where `agent` has a
   `sleep` tool (await setTimeout) and toolCallConcurrency set:
       defaultOptions: { maxSteps: 20, toolCallConcurrency: 5 }
   The agent must have NO suspending / requireApproval tools.
2. Register it on a Mastra instance and run a self-hosted Inngest dev server (`inngest dev`).
3. Send ONE stream() turn asking the LLM to call the sleep tool 5 times in parallel, each 20s:
       "Call the sleep tool FIVE times in one turn, each for exactly 20 seconds."
4. Measure total wall time / inspect the run trace in the Inngest dashboard.

Observed: total ≈ 100s (5 × 20s, sequential). The 5 `durable-tool-call` steps run
back-to-back — each starts when the previous ends.

Confirming it's the missing foreach option: patching dist/index.js to
[...]

PR #19855, opened 2026-07-21
fix(core,inngest): run finish-time side effects on remote durable workers — persist memory, output processors, thread title

### LINKED ISSUE #19839: [BUG] Inngest durable runs skip all finish-time side effects: no memory persistence, no output processors, no thread title
### Describe the Bug

A durable agent running on the `@mastra/inngest` engine completes its run but skips **every finish-time side effect** when the agentic loop executes on a `connect()` worker — the normal production topology:

- **No messages are persisted to memory.** A completed run leaves zero rows in the messages table and no thread record update. The next turn on the same thread recalls nothing — the agent behaves as if the conversation never happened.
- **`processOutputResult` output processors never run** at finish (including the built-in MessageHistory save).
- **No thread title is generated**, even with `generateTitle` enabled (the cross-process half of #19314 — the fix in #19315 parks a closure on the run registry, which only exists in the process that called `stream()`).

Two root causes stack:

1. `resolveRuntimeDependencies` (`packages/core/src/agent/durable/utils/resolve-runtime.ts`) rebuilds `memory` and the processor pipeline on the worker and writes them back into `globalRunRegistry` — but builds the `SaveQueueManager` **after** the write-back, so the registry entry never carries it and the finish-time persistence guard (`registryEntry?.saveQueueManager && …`) always fails cross-process.
2. The Inngest agentic workflow (`workflows/inngest/src/durable-agent/create-inngest-agentic-workflow.ts`) **has no finish-time block at all**: core's `createDurableAgenticWorkflow` runs output processors → memory persistence → thread title in its terminal map, but the Inngest variant's `map-final-output` only closes spans and emits the finish event. Even with (1) fixed, nothing on the Inngest path ever calls the persistence logic.

Same class of per-process-registry bug as #19330 (tool drop) and #19712 (suspend metadata): every `registryEntry?.…` guard silently no-ops on the worker.

### Steps To Reproduce

1. Build an agent with `memory: new Memory({ storage, options: { generateTitle: true } })` and wrap it with `createInngestAgent({ agent, inngest })`.
2. Start a **separate** `connect()` worker process (this is the part that matters — an in-process/serve-mode run cannot reproduce it, because the calling process's registry is populated there).
3. `await durableAgent.stream([{ role: 'user', content: 'My name is Zebra. Remember it.' }], { memory: { thread, resource } })` and let the run complete.
4. Query storage: the thread has **0 messages** and no title.
5. Send a second turn "What is my name?" on the same thread: the model's prompt contains no history — recall returns nothing.

https://gist.github.com/devangpadhiyar/9b8f33152258f5311ddc404e1922fb1b

### Expected Behavior

Parity with the non-durable agent and with in-process durable runs: after a completed run the user + assistant messages are persisted, `processOutputResult` processors have run, the thread title is generated, and the next turn on the thread recalls prior history.

### Environment Information

```block
- @mastra/core: 1.52.0-alpha.8 (main)
- @mastra/inngest: 1.8.3-alpha.0 (main)
- Node.js: 22.15.1
- OS: macOS (darwin 25.5.0)
- Topology: Inngest dev server + separate connect() worker process
```

### Verification

devangpadhiyar:
## Description

A durable agent running on the `@mastra/inngest` `connect()` worker — the normal production topology — finishes its run without ever executing the finish-time side effects. Nothing is persisted to memory (so the next turn on the same thread recalls nothing and the agent acts like the conversation never happened), `processOutputResult` output processors never run, and no thread title is generated.

Two things stack up to cause this. `resolveRuntimeDependencies` rebuilds `memory` and the processor pipeline on the worker and writes them back into the per-process run registry, but it constructed the `SaveQueueManager` *after* that write-back — so the registry entry never carries it and the finish-time persistence guard (`registryEntry?.saveQueueManager && …`) always fails cross-process. And even with that fixed, the Inngest workflow's terminal step simply has no finish-time blocks at all: core's `createDurableAgenticWorkflow` runs output processors → memory persistence → thread title in its terminal map, but `createInngestDurableAgenticWorkflow` only closes spans and emits the finish event. Same engine-drift family as #19317.

The fix moves the finish-time side effects into a shared `runDurableFinishSideEffects` helper (exported from `@mastra/core/agent/durable`) that both engines' terminal steps call. It first rebuilds the runtime dependencies from the Mastra instance when the process-local registry entry is missing or incomplete — which also covers the terminal step landing on a different worker than the LLM step — then runs the same output-processor, memory-persistence, and title blocks as the in-process engine, before the finish event goes out. Finish-time spans (`processor_run`, `memory_operation`) are parented under the run's rebuilt `AGENT_RUN` span; the terminal step's own tracing context is an internal workflow span that never exports, which left them orphaned.

Title generation is extracted into `generateDurableThreadTitle`: preparation's registry closure now delegates to it, and the finish helper calls it directly with agent + memory rebuilt from Mastra when the closure is absent — closing the cross-process limitation I noted in #19315.

The regression test spawns a real connect worker in a child process (same harness as #19713 — an in-process test can't reproduce this because the registry is populated there) and asserts messages + thread row persist, the title gets generated, and turn 2 recalls turn 1 via a deterministic `recall:yes` probe model. Fails on `main`, passes with the fix. The inngest build config also now excludes `src/__tests__/**` from type generation so test fixtures with dev-only deps don't break `pnpm build`.

Minimal repro: https://gist.github.com/devangpadhiyar/9b8f33152258f5311ddc404e1922fb1b

## Related issue(s)

Fixes #19839

## Type of change

## Checklist

Verified locally: new cross-process test 2/2; core typecheck, inngest tsc, and eslint all clean. The lockfile delta is just the `@mastra/memory` devDependency link.


codeXLinkX:
Cross-linking from Nomerra: we’re temporarily carrying a `pnpm` patch on `@mastra/inngest` that includes a finish-time memory flush workaround for #19839, plus a separate `cancelOn` runId scoping fix (see #19869 / https://github.com/mastra-ai/mastra/pull/19882).

We’ll drop the memory-flush patch hunk when this PR lands in a released `@mastra/inngest` / `@mastra/core`.

weilueluo:
what happens if saveMessages fails once? looks like the run might still finish without saving anything

devangpadhiyar:
@weilueluo 
> what happens if saveMessages fails once? looks like the run might still finish without saving anything

Good catch — as pushed, yes: a failed save was just a logger.warn and the run finished anyway.

Pushed 4bbfbcd to address this: persist/rebuild failures at finish are now tracked MastraErrors (logger.error + trackException, same AGENT_MEMORY_PERSIST_RESPONSE_MESSAGES_FAILED id the non-durable agent uses), so they surface in error tracking instead of a warn line.

Failing the run so Inngest retries the step is deliberately out of scope: it matches the non-durable contract (a persist failure there doesn't fail the run either), and retries wouldn't help today — SaveQueueManager drains the unsaved set before saving and swallows persist errors internally, so a retry would flush nothing. That's a pre-existing core issue affecting the non-durable path too; I'll file it separately.

abhiaiyer91:
Closing as superseded by merged PR #20926, which resolved linked issue #19839 by finalizing durable runs on remote workers. Thanks for the detailed investigation and implementation.

Issue #19301, opened 2026-07-11
Durable/Evented agent: resume() / approveToolCall() / sendToolApproval() don't auto-rehydrate via prepare(), so cold resume after restart still throws "No registry entry found"


devangpadhiyar:
## Summary

Follow-up to #18031 (fixed by #18113). That fix made `prepare([], { runId })` forward the `runId`, so a **manual** `prepare()` → `resume()` sequence can resume a persisted, suspended durable run across a process restart.

But the **documented, first-class resume APIs still don't perform that rehydration themselves**, so on `@mastra/core` **1.50.1** a user who follows the docs still hits the cold-resume wall after a restart (or after the in-process registry TTL elapses):

- `agent.resume(runId, data)` → `Error: No registry entry found for run . Cannot resume.`
- `agent.approveToolCall({ runId })` / `agent.declineToolCall({ runId })` → route through `resumeStream()` → `resume()` → same throw.
- `agent.sendToolApproval({ threadId, resourceId, approved })` → discovers the `runId` from storage via `listSuspendedRuns()`, then hands off to `resumeStream(..., { runId })` → `resume()` → **same throw**.

So storage-backed *discovery* survives a restart, but the *resume* it delegates to does not. `sendToolApproval()`'s restart-safety is only half a feature: it finds the run but can't resume it.

This matters for any human-in-the-loop pipeline where a run legitimately sits suspended at an approval for minutes-to-hours: every deploy/restart makes in-flight approvals unresumable via the public API.

## Environment

- `@mastra/core` **1.50.1**
- Storage: `@mastra/pg` `PostgresStore`
- `createEventedAgent()` wrapping an `Agent` with a `requireApproval` tool

## Minimal reproduction (runnable)

A single file. A second `Mastra` instance sharing the same `PostgresStore` models a restart — same persisted snapshot, empty in-memory registry.

```js
// repro-19301.mjs
// Run: ANTHROPIC_API_KEY=... DATABASE_URL=postgres://... node repro-19301.mjs
// Deps: @mastra/core@1.50.1 @mastra/pg @ai-sdk/anthropic zod pg
import { Mastra } from '@mastra/core/mastra';
import { Agent } from '@mastra/core/agent';
import { createEventedAgent } from '@mastra/core/agent/durable';
import { createTool } from '@mastra/core/tools';
import { PostgresStore } from '@mastra/pg';
import { anthropic } from '@ai-sdk/anthropic';
import { z } from 'zod';
[...]

PR #19877, opened 2026-07-21
fix(inngest): restore requestContext and tracingContext on durable resume

### LINKED ISSUE #19873: [BUG] Durable (@mastra/inngest) resume loses requestContext and tracingContext — the suspend snapshot never persists them
### Describe the Bug

On the Inngest engine, resuming a suspended durable run (a tool approval / any HITL suspend) comes back with an empty `requestContext` and a brand-new trace. The core engine doesn't have either problem — it persists `requestContext` and `tracingContext` into the suspend snapshot and restores them on resume. The Inngest engine's snapshot writer omits both, and its resume-event builder also ignores any `requestContext` the caller passes to `resume()`.

This is the Inngest-side sibling of the durable cold-resume work already merged for the core path — #18031 (dropped `suspendPayload.args`) and #19301 (registry rehydration) — and it's adjacent to the tracing gap in #19842. Those fixed the core `durable-agentic-loop`; the Inngest engine was left behind.

The practical fallout: any tool that scopes its work by a `requestContext` value survives the initial run but misbehaves after an approval. In my case a workspace/VFS tool keys its filesystem prefix off `requestContext.get('team_id')`/`get('thread_id')`; after resume that read is empty, so every write lands under an `anon/default` fallback. Separately, the resumed turn shows up in Langfuse as its own disconnected trace instead of continuing the original.

The approval decision itself round-trips fine — approve/reject/edit all reach the shared `tool-call-step` and the tool executes on resume. It's only the ambient context (`requestContext`, `tracingContext`) that's lost, so all four suspend/resume variants are affected equally.

**Where it comes from:**

- `workflows/inngest/src/workflow.ts` — the suspended-snapshot persist object omits both fields. Compare `packages/core/src/workflows/handlers/entry.ts`, whose snapshot includes `requestContext: engine.serializeRequestContext(...)` and `tracingContext`.
- `workflows/inngest/src/durable-agent/create-inngest-agent.ts` — the resume path sends `requestContext: snapshot?.requestContext ?? {}` (snapshot-only) and never forwards `tracingContext`. Core's resume (`packages/core/src/workflows/workflow.ts`) instead merges caller-provided `requestContext` over the snapshot and reuses the persisted `tracingContext.traceId`/`spanId` as the resumed span's parent.
- On resume the Inngest function is re-triggered by event, so `workflow.<id>.span.start` isn't memoized from the original run and mints a fresh root span with a new `traceId`.

I have the fix and a cross-process regression test ready and am happy to open a PR.

### Steps To Reproduce

Full runnable repro (spawns a real `connect()` worker — an in-process run hides the bug) in the linked repo. In short:

1. `npm install`
2. Start the Inngest dev server: `npx inngest-cli@latest dev -p 4100`
3. Start the connect worker in a second terminal: `node worker.mjs` (wait for `[worker] ready`)
4. Run the driver: `node driver.mjs`. It starts a durable run with `requestContext = { tenant: 'team-42' }`, which suspends at a `save_note` approval; then it calls `resume(runId, { approved: true })`.
5. The `save_note` tool, on approval, writes `note.txt` to a tenant-scoped dir read from `requestContext.get('tenant')`.

Observed: the worker logs `[tool] wrote note under tenant="anon"` and the file lands in `out/anon/`; the driver exits `1`. Expected it under `out/team-42/`.

https://github.com/devangpadhiyar/mastra-inngest-resume-requestcontext-repro

### Expected Behavior

A resumed durable run should see the same `requestContext` it had on the initial run and continue the same trace — matching the non-durable engine. Concretely: after `resume(runId, { approved: true })`, `requestContext.get('tenant')` inside the tool should still be `"team-42"`, the write should land under `out/team-42/`, and the resumed turn's spans should nest under the original run's trace instead of forming a new root trace.

### Environment Information

```block
System:
  OS: macOS 15.x (darwin)
  Node: 22.x
npmPackages:
  @mastra/core: ^1.51.0
  @mastra/inngest: ^1.8.2
  @mastra/libsql: ^1.0.0   (also reproduces on @mastra/pg)
  @mastra/memory: ^1.0.0
  inngest: ^4.5.0
Storage: LibSQL (DefaultStorage) / PostgresStore
LLM Provider: mock model in the repro (provider-agnostic — the bug is in the durable resume path, not the model)
```

### Verification

devangpadhiyar:
## Description

When a suspended durable run resumes on the Inngest engine, it comes back with an empty `requestContext` and starts a fresh trace. The core engine doesn't have either problem — its `persistStepUpdate` writes `requestContext` and `tracingContext` into the suspend snapshot and restores them on resume — but the Inngest engine's snapshot writer omitted both, and its resume-event builder read `requestContext` from the snapshot only (ignoring any caller-provided context) and never carried `tracingContext` at all.

The practical fallout is that any tool scoping its work by a `requestContext` value survives the initial run but breaks after a tool approval. In my case a workspace/filesystem tool keys its path prefix off `team_id`/`thread_id`; after resume that read is empty and every write lands under an `anon` fallback. Separately, the resumed turn shows up as its own disconnected trace instead of continuing the original.

The approval decision itself round-trips fine — approve/reject/edit all reach the shared `tool-call-step` and the tool executes on resume. It's only the ambient context that's lost, so all four suspend/resume variants are affected equally.

The fix mirrors core:

- `InngestWorkflow` now persists `requestContext` (via `RequestContext.toJSON()`) and `tracingContext` (`{ traceId, spanId }` from the workflow root span) into the suspended snapshot, matching `packages/core/src/workflows/handlers/entry.ts`.
- `InngestAgent.resume()` builds the resume event's `requestContext` by merging any caller-provided context over the snapshot (caller wins, like core), and forwards the persisted `tracingContext` as `tracingOptions`. The workflow handler already threads `event.data.tracingOptions` into its `span.start` step, so the resumed root span continues the original trace (`traceId`) as a child of the span the run suspended at (`parentSpanId`) — no new handler code needed.
- `InngestAgentResumeOptions` gains a `requestContext` field for parity with core's `DurableAgent.resume()`.

This is the Inngest-side sibling of the durable cold-resume work already merged for the core path (#18031, #19301).

I verified it with the regression test below (a real cross-process `connect()` worker, since an in-process run keeps the context in memory and hides the bug): with the fix the resumed tool reads `team-42` and writes to the right place and the suspended trace is the one the run continues in; reverting either half of the fix turns it red — the snapshot loses `tracingContext` and the write falls back to `anon`.

## Related issue(s)

Fixes #19873

## Type of change

- [x] Test update

## Checklist

### DIFF
--- a/.changeset/inngest-resume-context.md
+++ b/.changeset/inngest-resume-context.md
@@ -1,0 +1,11 @@
+---
+'@mastra/inngest': patch
+---
+
+Fixed the Inngest durable engine losing `requestContext` and `tracingContext` when a suspended run is resumed. Core's engine persists both into the suspend snapshot (`persistStepUpdate`) and restores them on resume; the Inngest engine's snapshot writer omitted both, and its resume-event builder read `requestContext` from the snapshot only (ignoring any caller-provided context) and never carried `tracingContext` at all.
+
+The effect was that any tool scoping its work by a `requestContext` value survived the initial run but broke after a tool approval — e.g. a workspace/filesystem tool keyed on `team_id`/`thread_id` wrote to an `anon` fallback once resumed — and the resumed turn started a fresh root span in a new trace instead of continuing the original.
+
+`InngestWorkflow` now persists `requestContext` and `tracingContext` into the suspended snapshot, and `InngestAgent.resume()` merges caller-provided context over the snapshot (caller wins, matching core) and forwards the persisted `tracingContext` as `tracingOptions` so the resumed workflow span continues the original trace. `InngestAgentResumeOptions` gains a `requestContext` field for parity with core's `DurableAgent.resume()`.
+
+Adds a cross-process regression test (real connect() worker) asserting the resumed tool sees the original `requestContext` and that the resumed turn stays in the original trace.
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -8529,6 +8529,9 @@ importers:
       '@mastra/libsql':
         specifier: workspace:*
         version: link:../../stores/libsql
+      '@mastra/memory':
+        specifier: workspace:*
+        version: link:../../packages/memory
       '@mastra/observability':
         specifier: workspace:*
         version: link:../../observability/mastra
--- a/workflows/inngest/package.json
+++ b/workflows/inngest/package.json
@@ -57,6 +57,7 @@
     "@mastra/core": "workspace:*",
     "@mastra/deployer": "workspace:*",
     "@mastra/libsql": "workspace:*",
+    "@mastra/memory": "workspace:*",
     "@mastra/observability": "workspace:*",
     "@types/express": "^5.0.6",
     "@types/koa": "^3.0.1",
--- a/workflows/inngest/src/__tests__/durable-resume-context.test.ts
+++ b/workflows/inngest/src/__tests__/durable-resume-context.test.ts
@@ -1,0 +1,158 @@
+/**
+ * Regression test for issue #19873: a resumed durable run must restore the `requestContext` and
+ * `tracingContext` it suspended with.
+ *
+ * Two things break without the fix, both because the Inngest engine's suspend snapshot never
+ * persists these fields (core's `persistStepUpdate` does):
+ *
+ *  1. requestContext — a tool that scopes its work by a context value (here: `tenant`) sees an empty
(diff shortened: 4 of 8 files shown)


abhiaiyer91:

We're currently testing out our agentic factory, and it implemented a fix for the linked issue (#19873), which has since been closed. So I'm closing this PR out as no longer needed.

Issue #19869, opened 2026-07-21
abort() on durable agents doesn’t survive process boundaries


daneatmastra:
This issue was created from Discord post 1529135790838255746:

[![Open in Discord](https://img.shields.io/badge/Open_in_Discord-5865F2?style=for-the-badge&logo=discord&logoColor=white)](https://discord.com/channels/1309558646228779139/1529135790838255746)

I’ve been digging into how cancellation works for durable agents (`DurableAgent` and `createInngestAgent`) and found two related problems I think. Working on `@mastra/inngest 1.8.3-alpha.0`, but the code is unchanged since 1.8.1.

1. `abort()` is in-process only, whcih is fine for single-instance durableagents, but doesn't do anything for:
- load balanced deployments
- inngestagents where the step worker is basically never the caller's process

I think this is what issue #[18001 ](https://github.com/mastra-ai/mastra/issues/18001)asked for exactly this (“Stop button should truly stop the run”) and was closed as fixed by #[18508](https://github.com/mastra-ai/mastra/pull/18508), but #18508 delivered the in-process version, so the cross-process case is still open.

2. The Inngest cancel event has no run scoping.
The Inngest function registers:

```cancelOn: [{ event: `cancel.workflow.${this.id}` }],   // workflow.ts:297
```

No `match` and no `if` , so emitting that event cancels all in-flight runs of the function not the targetted one, AND since every durable agent shares the single inngest:durable-agentic-loop function, that means every durable agent run in the deployment across all runs. And only the targeted run's snapshot gets marked canceled, the rest just die.

I'm new to the world of  `durable`  thoguh so I feel like I might be missing things?

daneatmastra:
Thank you for reporting this issue! We have labeled it for trio-tb and we will look into it as soon as possible.

If you're experiencing an error, please provide a [minimal reproducible example](https://github.com/mastra-ai/mastra/blob/main/CONTRIBUTING.md#minimal-reproduction) whenever possible to help us resolve it quickly.

hemaaanth:
to repro:
```ts
import { Agent } from '@mastra/core/agent';
import { createTool } from '@mastra/core/tools';
import { createToolCallThenTextModel } from '@internal/workflow-test-utils';
import { z } from 'zod';
import { describe, it, expect, beforeAll, afterAll, vi } from 'vitest';
import { createInngestAgent } from '../durable-agent';
import { InngestDurableStepIds } from '../durable-agent/create-inngest-agentic-workflow';
[...]

PR #19882, opened 2026-07-21
fix(inngest): scope cancelOn to workflow runId

### LINKED ISSUE #19883: [BUG] @mastra/inngest cancelOn is not scoped by runId — one cancel kills all concurrent runs
### Describe the Bug

`@mastra/inngest` registers Inngest functions with:

```ts
cancelOn: [{ event: `cancel.workflow.${this.id}` }]
```

and **no** `if` / match expression. Inngest therefore cancels every in-flight run of that function when `cancel.workflow.<id>` is emitted.

`InngestRun.cancel()` already sends the intended scope:

```ts
await this.inngest.send({
  name: `cancel.workflow.${this.workflowId}`,
  data: { runId: this.runId },
});
```

…but without `if: 'async.data.runId == event.data.runId'`, that payload is ignored for matching. For durable agents that share a single workflow id (`inngest:durable-agentic-loop`), cancelling one run aborts every concurrent durable-agent run in the deployment. Only the targeted run’s snapshot is marked canceled; the others are killed without a proper cancel snapshot update.

This is the same blast-radius problem described as item 2 in #19869. Opening a focused issue so a narrow PR can use `Fixes #<this>`.

### Steps To Reproduce

1. Start two concurrent Inngest workflow runs (or two `createInngestAgent` runs) that share the same workflow function id and block in a long step.
2. Call `run.cancel()` / emit `cancel.workflow.<id>` with `{ runId }` for only the first run.
3. Observe the second run is also canceled/killed.

See the concurrent repro on #19869.

### Expected Behavior

Only the run whose `runId` matches `event.data.runId` is canceled. Sibling concurrent runs continue.

### Environment Information

- `@mastra/inngest` main / 1.8.x (confirmed unscoped `cancelOn` still present in `workflows/inngest/src/workflow.ts`)
- Inngest SDK v4 `cancelOn.if` expression support

### Proposed Fix

```ts
cancelOn: [{
  event: `cancel.workflow.${this.id}`,
  if: 'async.data.runId == event.data.runId',
}]
```

for both the workflow function and the cron function.

PR: https://github.com/mastra-ai/mastra/pull/19882

### Temporary workaround note (Nomerra)

We currently carry a local `pnpm` patch on `@mastra/inngest` for:
1. this `cancelOn` runId scoping fix, and
2. a finish-time memory flush workaround tracked by #19839 / #19855

We’ll drop the memory hunk when #19855 lands, and drop the cancelOn hunk when this fix ships in a stable `@mastra/inngest` release.

### Verification

codeXLinkX:
## Description

`@mastra/inngest` registers:

```ts
cancelOn: [{ event: `cancel.workflow.${id}` }]
```

without an `if` / match expression. Inngest treats that as “cancel every in-flight run of this function,” so one `InngestRun.cancel()` (which already sends `{ runId }`) can kill all concurrent runs — including every durable agent sharing `inngest:durable-agentic-loop`.

This PR scopes cancellation with Inngest’s cancellation `if`:

```ts
if: 'async.data.runId == event.data.runId'
```

for both the workflow function and the cron function. `InngestRun.cancel()` already sends `data: { runId }`, so no cancel-path API change is required.

## Related issue(s)

Fixes #19883

Also related: #19869 (broader abort/cancel discussion; this PR addresses only the unscoped `cancelOn` blast radius).

Temporary-workaround context from Nomerra:

- Memory finish-time side effects: #19839 / #19855 — we currently carry a temporary `pnpm` patch for both the cancelOn scoping fix and a memory-flush workaround, and will drop the memory hunk when #19855 lands.

## Type of change

## Checklist

## Test plan

- [x] Unit test spies `inngest.createFunction` and asserts both workflow + cron `cancelOn` include `if: 'async.data.runId == event.data.runId'`
- [ ] Optional: run the concurrent cancel blast-radius repro from #19869 and confirm bystander runs finish

### DIFF
--- a/.changeset/inngest-cancelon-runid-scope.md
+++ b/.changeset/inngest-cancelon-runid-scope.md
@@ -1,0 +1,5 @@
+---
+'@mastra/inngest': patch
+---
+
+Scope Inngest `cancelOn` to the target workflow `runId` so one cancel no longer aborts every concurrent run of the shared function.
--- a/workflows/inngest/src/cancel-on-runid-scope.test.ts
+++ b/workflows/inngest/src/cancel-on-runid-scope.test.ts
@@ -1,0 +1,99 @@
+import { Mastra } from '@mastra/core/mastra';
+import { MockStore } from '@mastra/core/storage';
+import { Inngest } from 'inngest';
+import { beforeEach, describe, expect, it, vi } from 'vitest';
+import { z } from 'zod';
+
+import { init } from './index';
+
+describe('InngestWorkflow cancelOn runId scope', () => {
+  let inngest: Inngest;
+  let createFunctionSpy: ReturnType<typeof vi.spyOn>;
+
+  beforeEach(() => {
+    inngest = new Inngest({ id: 'cancel-on-scope-test' });
+    createFunctionSpy = vi.spyOn(inngest, 'createFunction');
+  });
+
+  it('registers cancelOn with runId match for the workflow function', () => {
+    const { createWorkflow, createStep } = init(inngest);
+
+    const step1 = createStep({
+      id: 'step1',
+      execute: async () => ({ result: 'done' }),
+      inputSchema: z.object({}),
+      outputSchema: z.object({ result: z.string() }),
+    });
+
+    const workflow = createWorkflow({
+      id: 'scoped-cancel-workflow',
+      inputSchema: z.object({}),
+      outputSchema: z.object({ result: z.string() }),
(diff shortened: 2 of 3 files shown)


abhiaiyer91:
Thanks for this, and apologies it went stale. The run-scoping fix landed in #21003 (with #21009 covering the cross-process half of #19869), so I'm closing this as superseded. You identified the missing cancelOn match correctly — appreciate the contribution.

PR #19888, opened 2026-07-21
fix(core): skip memory persistence when agent generate fails before assistant output

### LINKED ISSUE #19887: [BUG] v2 agent persists user message to memory when generate() fails before any assistant output
### Describe the Bug

When an agent with memory calls `agent.generate()` and the model fails before producing any assistant output (immediate throw, API error, etc.), v2 still persists the **user message** to thread storage via the `MessageHistory` output processor. There is no assistant reply for that turn.

**Impact:**
- Orphan user messages in thread history after failed API calls
- Retries look like a new turn instead of a retry of a failed one
- `memory.recall()` returns user text with no matching assistant message

**Root cause (suspected):** The agentic loop emits a `finish` chunk with `stepResult.reason === 'error'`, but the outer `MastraModelOutput` still runs `runOutputProcessors()`. `MessageHistory.processOutputResult` saves `[...newInput, ...newOutput]`, so the user input is persisted even with empty output.

**v1 behavior:** Failed runs typically save nothing (legacy path ties persistence to successful completion).

**Note:** `map-results-step` already skips `executeOnFinish` on error, but that does not cover the outer-stream output processor path.

### Steps To Reproduce

1. Clone `mastra-ai/mastra` and check out a version **before** the fix (e.g. `main` without the output-processor gating change).
2. From `packages/core`, run the existing regression tests (or use the snippet below):

```
ts
import { MockMemory } from '@mastra/core/memory';
import { MockLanguageModelV2 } from '@internal/ai-sdk-v5/test';

const memory = new MockMemory();
const agent = new Agent({
  id: 'repro',
  name: 'Repro',
  instructions: 'test',
  memory,
  model: new MockLanguageModelV2({
    doGenerate: async () => { throw new Error('Simulated error during response'); },
    doStream: async () => { throw new Error('Simulated error during response'); },
  }),
});

try {
  await agent.generate('trigger error', {
    memory: { resource: 'user-1', thread: { id: 'thread-1' } },
  });
} catch {}

const recalled = await memory.recall({ threadId: 'thread-1', resourceId: 'user-1' });
console.log('message count:', recalled.messages.length); // 1 (user only) — bug
console.log('roles:', recalled.messages.map(m => m.role)); // ['user']
```

3. Alternatively, from packages/core:
`pnpm exec vitest run src/agent/__tests__/save-and-errors.test.ts -t "should save thread but not messages if error occurs"`

On affected versions, memory.recall() shows a persisted user message even though generate() threw and no assistant output was produced.

4. Compare with v1 / legacy generateLegacy on the same scenario: no messages are persisted.

Inline repro (no separate repo): see Steps To Reproduce.

Fix PR: #19888 

### Expected Behavior

When `agent.generate()` fails before any assistant output is produced:

1. **No messages** should be persisted for that turn (same as v1).
2. The call should still **throw/reject** with the model error.
3. The **thread** may still be created upfront (for stores that require thread existence before saves) — that is fine — but **message history must remain empty** until a successful turn completes.
4. After a **successful** generate, user + assistant messages should still be saved normally.

### Environment Information

```
System:
    OS: Windows 11 10.0.26200
    CPU: (12) x64 13th Gen Intel(R) Core(TM) i7-1355U
    Memory: 2.19 GB / 15.68 GB
  Binaries:
    Node: 24.18.0 - D:\New Node\node.EXE
    npm: 11.16.0 - D:\New Node\npm.CMD
    pnpm: 11.8.0 - C:\Users\cleme\AppData\Roaming\npm\pnpm.CMD
    bun: 1.3.14 - C:\Users\cleme\.bun\bin\bun.EXE
  npmPackages:
    vitest: catalog: => 4.1.10 
  npmGlobalPackages:
    @hubspot/cli: 7.6.2
    pnpm: 10.34.1
    vercel: 50.1.2
```

### Verification

clementantonyk:
## Description

Fixes a v2 regression where `agent.generate()` and the streaming path persisted the **user message** to thread memory even when generation failed before producing any assistant output.

Previously, output processors were still executed when generation ended with an error, causing `MessageHistory` to save the user message despite the run failing.

### Fix

* Skip `runOutputProcessors()` for the outer stream when generation ends with an error (`stepResult.reason === 'error'`) or the stream has already recorded an error (`self.#error`).
* Preserve the existing behavior of creating the thread upfront to avoid Postgres race conditions.

### Tests

Added and updated tests to verify that:

* Failed `generate()` calls do not persist conversation messages.
* v1 and v2 now have consistent behavior (0 messages after a failed generation).
* Streaming failures before any assistant output also do not persist messages.
* Message persistence is verified by spying on `storage.saveMessages`, which is used internally by `MessageHistory`.

## Related issue

Fixes #19887

## Type of change

- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [x] Test update

## Checklist

- [x] I have linked the related issue(s) above.
- [ ] I have made corresponding documentation changes (if applicable).
- [x] I have added tests that verify the fix.
- [ ] I have addressed all CodeRabbit comments.

## Test plan

```bash
cd packages/core

pnpm exec vitest run src/agent/__tests__/save-and-errors.test.ts -t "should not save any message if interrupted|should save thread but not messages if error occurs"

pnpm exec vitest run src/agent/__tests__/stream.test.ts -t "interrupted before any part|error occurs during streaming"
```

## ELI5

When generation fails before the agent produces a response, the user message is not saved as an orphaned message. The thread is still created, and successful runs still save messages normally.

## What changed

- Fixed a v2 regression in `agent.generate()` and streaming runs.
- Skipped outer-stream output processors when generation ends with an error or records an internal stream error.
- Preserved upfront thread creation without saving messages for failed turns.
- Added a changeset for the `@mastra/core` patch release.

## Tests

- Verified that failed `generate()` calls save no messages and still reject with the model error.
- Covered streaming errors and interruptions before assistant output.
- Verified consistent v1 and v2 behavior.
- Spied on the underlying storage `saveMessages` method to confirm that failed turns do not persist messages.


clementantonyk:
@abhiaiyer91 Just following up on this PR. The fix includes regression tests covering both `generate() `and streaming error paths, ensuring no orphan user messages are persisted when generation fails before any assistant output while preserving existing behavior for successful runs. I'd appreciate a review when someone has time. Thanks!

clementantonyk:
Rebased/merged with main and resolved the agent.test.ts conflict. The fix is unchanged stepResult.reason !== 'error' guard still covers the finish-without-error-chunk case that #19716 misses. Ready for re-review. @abhiaiyer91

CodingButter (changes requested):
Thanks for sticking with this. The storage-level spies and stronger empty-history assertions are worth keeping. They check the actual persistence path rather than relying on `Memory.saveMessages` being called.

The production fix has been handled on main. #19716 added orphan-input protection, and #21370 moved it into `MessageHistory` so other output processors could still run on failed turns. I also checked the finish-without-error-chunk case you mentioned: main passes that finish reason through to `MessageHistory`, which skips saving an input-only failed turn. Local checks exercised `generate()` and `stream()` separately.

I'd like to keep this as a test-only PR. The changes I'd ask for are:

1. Update against current main and drop the two added conditions in `output.ts` that skip output processors on errors. Keep main's production behavior unchanged.
2. Keep the storage-level spies and the assertions that failed turns with no assistant output leave history empty. Remove any stale TODO or v1-only assertion gate that says this behavior is still broken in modern agents. Assert that `generate()` rejects on failure with `await expect(...).rejects.toThrow()`, rather than a catch-only assertion that could pass if nothing throws.
3. Keep the existing provider-throw processor check in `agent-processor.test.ts`. Extend the existing finish-error fixtures to assert that output processors run and see `finishReason: 'error'` when a finish carries `reason: 'error'` but no error chunk. Check `generate()` and `stream()` independently, including empty backing storage and no save calls for these input-only failures. Reuse the model fixtures rather than duplicating them.
4. Remove the bug-fix changeset, since this would no longer change shipped behavior, and update the title and description to describe the test improvements. Run the affected tests against main.

I transplanted the production guard onto the main checkout I tested, and locally added assertions in two tests failed because output processors stopped running on error terminals. That was not a full test run of your branch. The requests above are about preserving the behavior main now supports, not asking you to revive the old fix.

If you'd like to make that update, we'll check back on Thursday, September 10, before starting a separate change. No pressure if you don't have time. If there are no updates or signs that you're working on it then, we'll carry the useful test changes into our own PR with a co-author credit and an explicit link back to your work here. If work is already underway, we'll avoid duplicating it. We'll review the revised tests and share our findings with the maintainers.

PR #19906, opened 2026-07-22
fix(core): honor Retry-After on agent model-call retries

### LINKED ISSUE #19885: [BUG] Agent model-call retry ignores Retry-After header (fixed exponential backoff)
## Describe the Bug

When an `Agent` retries a failed model call (via `maxRetries`), it backs off on a
**fixed exponential schedule and ignores the `Retry-After` response header**.
Calling the underlying AI SDK (`ai`) `generateText`/`streamText` directly, with the
same model and the same `maxRetries`, **does** honor `Retry-After`. So wrapping a
model in a Mastra `Agent` silently downgrades the retry behavior the AI SDK provides.

For rate-limited providers this is the difference between recovering and failing: a
provider that says "retry after 30s" is retried on a `1s, 2s, 4s` schedule instead —
so all attempts land inside the still-throttled window and every one fails.

**Root cause (from the published bundle):** `@mastra/core` bundles two retry
implementations — `retryWithExponentialBackoffRespectingRetryHeaders`
(+ `getRetryDelayInMs`, which reads `retry-after` / `retry-after-ms` from
`error.responseHeaders`), used by the bundled `ai` `prepareRetries`; and a **plain**
`retryWithExponentialBackoff` whose `prepareRetries` returns
`retry: retryWithExponentialBackoff({ maxRetries })` with
`delayInMs: backoffFactor * delayInMs` and no header inspection. The agent's
model-call path resolves `maxRetries` (`this.maxRetries ?? 0`) and runs the call
through the **plain** variant, so `Retry-After` is dropped.

## Steps To Reproduce

Minimal repro (also as a gist — see link below). A local mock "provider" always
returns `HTTP 429` + `Retry-After: 5` and logs the gap between requests; the same
model is called with `maxRetries: 3` two ways.

`package.json`:

```json
{
  "name": "mastra-agent-retry-after-repro",
  "private": true,
  "type": "module",
  "dependencies": {
    "@ai-sdk/open-responses": "1.0.26",
    "@mastra/core": "1.50.1",
    "ai": "6.0.197",
    "zod": "4.4.3"
  }
}
```

`repro.mjs`:

```js
import { createServer } from 'node:http';
import { generateText } from 'ai';
import { createOpenResponses } from '@ai-sdk/open-responses';

const PORT = 8787;
const RETRY_AFTER = 5;
let lastHit = null;
const gaps = [];

const server = createServer((req, res) => {
  if (req.method !== 'POST') return void res.writeHead(405).end();
  const now = Date.now();
  gaps.push(lastHit == null ? 0 : (now - lastHit) / 1000);
  lastHit = now;
  req.resume();
  req.on('end', () => {
    const body = JSON.stringify({ error: { message: 'rate limited', type: 'rate_limit_error' } });
    res.writeHead(429, {
      'content-type': 'application/json',
      'content-length': Buffer.byteLength(body),
      'retry-after': String(RETRY_AFTER),
    });
    res.end(body);
  });
});
await new Promise((r) => server.listen(PORT, r));

const model = createOpenResponses({ name: 'mock', url: `http://127.0.0.1:${PORT}/v1/responses`, apiKey: 'sk-test' })
  .languageModel('test');

async function run(label, fn) {
  lastHit = null; gaps.length = 0;
  const t0 = Date.now();
  try { await fn(); } catch { /* exhausts retries against an always-429 mock */ }
  console.log(`${label}: ${gaps.length} attempts, gaps [${gaps.map((x) => x.toFixed(1) + 's').join(', ')}], total ${((Date.now() - t0) / 1000).toFixed(1)}s`);
}

console.log(`Provider always returns HTTP 429 with "Retry-After: ${RETRY_AFTER}". maxRetries=3.\n`);
await run('ai.generateText', () => generateText({ model, maxRetries: 3, prompt: 'hi' }));
const agent = new Agent({ id: 't', name: 't', instructions: 'x', model, maxRetries: 3 });
await run('Agent.generate ', () => agent.generate('hi'));
server.close();
```

Run:

```bash
npm install
node repro.mjs        # summary on stdout; the final exhausted-retry error is on stderr
```

Output:

```
Provider always returns HTTP 429 with "Retry-After: 5". maxRetries=3.

ai.generateText: 4 attempts, gaps [0.0s, 5.0s, 5.0s, 5.0s], total 15.1s   # honors Retry-After
Agent.generate : 4 attempts, gaps [0.0s, 1.0s, 2.0s, 4.0s], total 7.1s    # fixed exponential
```

https://gist.github.com/segevfiner/a60331487856303db21cd427ad8ea337

## Expected Behavior

`Agent.stream()` / `Agent.generate()` retries should honor `Retry-After` (and
`Retry-After-Ms`) on retryable errors, the same way `ai`'s `generateText` /
`streamText` do — i.e. use `retryWithExponentialBackoffRespectingRetryHeaders`
(or a `getDelayInMs` that reads the retry headers) on the agent model-call path.

## Environment Information

```
System:
  OS: macOS 26.5.2
  CPU: (10) arm64 Apple M1 Pro
Binaries:
  Node: 22.23.1
  npm: 12.0.1
  pnpm: 11.15.1
npmPackages:
  @ai-sdk/open-responses: 1.0.26 => 1.0.26
  @mastra/core: 1.50.1 => 1.50.1
  ai: 6.0.197 => 6.0.197
  zod: 4.4.3 => 4.4.3

LLM Provider: OpenAI-compatible gateway (reproduced with a local mock returning 429 + Retry-After)
```

## Verification

kimlj:
Fixes #19885

## Problem

When an `Agent` retries a failed model call via `maxRetries`, it backs off on a fixed exponential schedule and ignores the `Retry-After` response header. A provider that replies `429` with `Retry-After: 5` is retried after 1s, 2s and 4s, so every attempt lands inside the window the provider is still throttling.

Verified on `main` (`@mastra/core@1.52.0-alpha.10`) with a mock provider returning `429` + `Retry-After: 5` and `maxRetries: 2`:

```
before: calls=3  gaps=[1.0s, 2.0s]  total=3.1s
after:  calls=3  gaps=[5.0s, 5.0s]
```

## Root cause

This differs from the diagnosis in the issue, which attributes the behavior to the AI SDK's `retryWithExponentialBackoff` and asks for `retryWithExponentialBackoffRespectingRetryHeaders`. Neither runs on this path.

`stream/aisdk/v5/execute.ts` strips `maxRetries` from the model settings before the call:

```ts
const filteredModelSettings = omit(modelSettings || {}, ['maxRetries', 'headers']);
```

That disables the AI SDK's own retry (which does honor the header), and the model call is instead wrapped in `p-retry`, which was given only `retries`, `signal` and `shouldRetry`. `p-retry` has no `Retry-After` awareness, and its defaults (`minTimeout: 1000`, `factor: 2`) produce exactly the 1s/2s/4s schedule reported.

## Fix

`onFailedAttempt` now reads `Retry-After` from the error and waits for it. Because `p-retry` applies its own exponential delay after that hook resolves, the hook waits only the remainder, so the total delay lands on `max(backoff, min(Retry-After, cap))`.

`minTimeout` and `factor` are now passed explicitly. Both match `p-retry`'s defaults, so scheduling is unchanged; passing them makes the scheduled backoff computable rather than assumed.

The honored delay is capped at 30s, reusing `DEFAULT_MAX_RETRY_AFTER_MS`, so a large or hostile `Retry-After` cannot wedge a run. This is the same bounded rule `StreamErrorRetryProcessor` applies (#19383), rather than the unbounded behavior the issue requests. The wait is abort-aware, and no delay is added after the final attempt.

## Shared helper

The `Retry-After` parsing in `StreamErrorRetryProcessor` was module-private. Rather than duplicate it, it moves to `utils/retry-after.ts` and both paths import it, so the two retry surfaces cannot drift apart.

That change is a pure move: `stream-error-retry-processor.ts` is +1/-81 with one import replacing the definitions, no logic and no call sites touched. Its existing `Retry-After` tests cover the relocated code unchanged. Happy to split this into a separate commit or PR if you would prefer to keep them apart.

## Testing

`packages/core/src/agent/__tests__/model-retry-after.test.ts` covers three cases: a numeric `Retry-After` is waited out instead of the shorter backoff, a value above the cap is clamped to 30s, and exponential backoff is unchanged when no header is sent. The first two fail on `main` and pass with this change; the third passes both ways as a guard against altering default behavior.

`pnpm --filter ./packages/core test src/processors src/stream src/agent/__tests__`:

```
main:        46 failed | 2186 passed | 44 skipped (2276)
this branch: 46 failed | 2189 passed | 44 skipped (2279)
```

Same 46 pre-existing failures (workspace tools / structured output, unrelated to this path), plus the 3 new tests. `pnpm --filter ./packages/core check` reports the same 612 pre-existing errors before and after, none in the changed files. ESLint is clean on the changed files.

## ELI5
When an API says “try again later,” Mastra agents now wait for the requested time before retrying. They still use exponential backoff when no retry delay is provided, and they never wait more than 30 seconds.

## What changed
- Updated Agent model-call retries to honor `Retry-After` and `Retry-After-Ms`.
- Added shared utilities to parse retry delays, follow error causes, cap delays, and support abort-aware waits.
- Made `Retry-After-Ms` take precedence over `Retry-After`.
- Preserved exponential backoff when no provider delay is available.
- Skipped retry delays for non-retryable errors and final attempts.
- Reused the shared utilities in `StreamErrorRetryProcessor`.
- Added tests for numeric and millisecond delays, header precedence, the 30-second cap, non-retryable errors, and exponential backoff.


kimlj:
`getRetryAfterMs` now reads `retry-after-ms` and prefers it over `retry-after` as the more precise of the two. This also applies to `StreamErrorRetryProcessor`, which shares the parser and previously read only `retry-after`.

PR #19912, opened 2026-07-22
fix(core): stop llmPrompt clobbering tool-result output with a nullish stored modelOutput

aisensiy:
## Summary

A durable Agent using an OpenRouter v6 (`@openrouter/ai-sdk-provider`) model hard-crashes the model request whenever thread history contains a text-only tool result, with:

```
Error processing LLM stream { error: Cannot read properties of undefined (reading 'type') }
   at getToolResultContent (@openrouter/ai-sdk-provider … input.output.type)
   at convertToOpenRouterChatMessages
   at OpenRouterChatLanguageModel.getArgs / doStream
```

The provider reads `output.type` on a `tool-result` whose `output` is `undefined`. The bad `output` is produced by `@mastra/core` itself while building the prompt — the stored DB message is well-formed — so once a thread holds such a message, **every subsequent turn crashes and the thread is permanently poisoned**.

## Root cause

`MessageList`'s `get.all.aiV5.llmPrompt()` (used by the durable agent via `get.all.aiV6.llmPrompt()` — see `agent/durable/workflows/steps/llm-execution.ts`) has a `storedModelOutputs` pass that restores a tool's rich model output onto the converted model message:

```ts
// message-list.ts
if (
  part.type === 'tool-invocation' &&
  part.toolInvocation?.state === 'result' &&
  part.providerMetadata?.mastra &&
  typeof part.providerMetadata.mastra === 'object' &&
  'modelOutput' in part.providerMetadata.mastra          // <-- key presence only
) {
  storedModelOutputs.set(part.toolInvocation.toolCallId, part.providerMetadata.mastra.modelOutput);
}
// ...later, unconditionally:
modelMsg.content[i] = { ...part, output: storedModelOutputs.get(part.toolCallId) };
```

The durable llm-mapping step persists that metadata from the tool's `toModelOutput`:

```ts
// agent/durable/workflows/steps/llm-mapping.ts
let modelOutput = await tool.toModelOutput(toolResult.result);
modelOutput = normalizeModelOutput(modelOutput);
// ...
mastra: { ...existingMastra, modelOutput },
```

When `toModelOutput` returns `undefined` (the common case for a **text-only** tool result — e.g. a workspace file read whose `toModelOutput` only maps media), this persists `mastra: { modelOutput: undefined }` — the **key is present but the value is nullish**. (A JSON dump of the DB row shows `mastra: {}` because `JSON.stringify` drops undefined-valued keys, but the key still exists in-memory within a running workflow, which is exactly when `llmPrompt` builds the next request.)

The guard keys off `'modelOutput' in mastra`, so it collects the `undefined`, and the override then **clobbers the already-correct converted `output` with `undefined`**. `aiV5PromptToAIV6Prompt` passes it through, and the provider throws on `output.type`.

This override exists **only** on the `llmPrompt` path — not `.prompt()` / `.model()` — which is why the crash is specific to the durable prompt build.

## Fix

Only override when the stored `modelOutput` is non-nullish. One-line guard addition; a nullish `modelOutput` now leaves the correctly-converted `output` intact.

## Test

Adds `packages/core/src/agent/message-list/tests/tool-result-output-undefined.test.ts`, which reproduces the crash by adding a faithful poisoned assistant message and asserting the `role:"tool"` model message from `get.all.aiV6.llmPrompt()` has a defined `output` carrying the result text. It fails on `main` (`output` is `undefined`) and passes with this change. Full `message-list` suite (42 files / 525 tests) stays green.

## Related

Same "tool-result loses its content on outbound conversion" family as #16017 (fabricated `args: {}` on stand-alone tool-results) and #17798 (background sub-agent placeholder → `null` tool content → provider 500); this is the `output`-side sibling, and the first that produces a hard `reading 'type'` crash rather than a validation/500.

## ELI5

A saved tool result was sometimes replaced with an empty value, causing agents to crash later. This change keeps the correct result and adds a regression test.

## What changed

- Restore durable tool output only when the stored value is non-nullish.
- Preserve converted tool-result output when `modelOutput` is `undefined`.
- Add a regression test covering text-only tool results through `aiV6.llmPrompt()`.
- Add a changeset patch note.

### DIFF
--- a/.changeset/tool-result-output-nullish-clobber.md
+++ b/.changeset/tool-result-output-nullish-clobber.md
@@ -1,0 +1,5 @@
+---
+'@mastra/core': patch
+---
+
+Fixed durable Agent prompt build clobbering a valid tool-result `output` with `undefined`. `MessageList`'s `llmPrompt()` restores stored tool outputs by keying off `'modelOutput' in providerMetadata.mastra` (key presence) and then unconditionally overwriting the already-converted `output`. When a tool's `toModelOutput` returns `undefined` (e.g. a text-only result), the durable llm-mapping step still persists `mastra: { modelOutput: undefined }`, so the override replaced the correct output with `undefined` — producing a tool-result whose `output` is missing. Providers that read `output.type` (e.g. `@openrouter/ai-sdk-provider`) then threw `Cannot read properties of undefined (reading 'type')`, aborting the request and poisoning the thread on every subsequent turn. The override now only applies when the stored `modelOutput` is non-nullish.
--- a/packages/core/src/agent/message-list/message-list.ts
+++ b/packages/core/src/agent/message-list/message-list.ts
@@ -578,7 +578,16 @@ export class MessageList {
               part.toolInvocation?.state === 'result' &&
               part.providerMetadata?.mastra &&
               typeof part.providerMetadata.mastra === 'object' &&
-              'modelOutput' in (part.providerMetadata.mastra as Record<string, unknown>)
+              'modelOutput' in (part.providerMetadata.mastra as Record<string, unknown>) &&
+              // Only override with a real stored output. A tool's `toModelOutput` can
+              // return `undefined` (e.g. a text-only result), which the durable
+              // llm-mapping step still persists as `mastra: { modelOutput: undefined }`
+              // (key present, value nullish). Without this guard the override below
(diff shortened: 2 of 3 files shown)


aisensiy:
Superseded by #19916 (reopen wasn't possible after a post-close force-push). #19916 carries the same fix on the same branch, cleaned up, with a linked issue #19915.

PR #19913, opened 2026-07-22
fix(deployer): support workspace subpath exports and transitive subpath imports under externals: true

### LINKED ISSUE #19379: [BUG] mastra build with externals: true fails with Missing "." specifier for workspace packages that only declare subpath exports

### LINKED ISSUE #19909: [BUG] mastra build with externals: true leaks transitive workspace subpath imports as unresolved bare specifiers (ERR_MODULE_NOT_FOUND)
### What version of Mastra are you running?

`@mastra/deployer` 1.52.0-alpha.8 (behavior present on `main`; verified at `e433b388`).

### What's the bug?

In a pnpm/Nx monorepo built with `bundler: { externals: true }`, a **workspace subpath** import that is pulled in **transitively** — i.e. workspace package `@scope/a` imports `@scope/b/sub`, and the app depends on `@scope/a` but not directly on `@scope/b` — is emitted into the bundle as an **unresolved bare specifier** (`import ... from '@scope/b/sub'`) that is **not** registered in the generated `.mastra/output/package.json`. The build reports success, but the container fails at runtime with `ERR_MODULE_NOT_FOUND`.

The build even prints a non-fatal `Error: Package @scope/b not found` to stderr during bundling but exits 0, so the leak is silent unless you inspect the output.

Companion to #19379 (subpath-only exports crash) — same `externals: true` monorepo path; the #19379 fix is a prerequisite to even reach this leak (a subpath-only package crashes there first).

### Mechanism (root cause)

The transitive-dependency discovery is **manifest-name-based, not source-aware**. When it walks transitive workspace dependencies it registers each workspace package's **root** only (`checkTransitiveDependencies` in `packages/deployer/src/build/analyze/analyzeEntry.ts` sets `exports: ['*']` per dependency **name** from `package.json`). It never reads a workspace package's **source** to learn which **subpaths** that package imports from other workspace packages. So a transitively-imported subpath (`@scope/b/sub`) is never added to `analyzedBundleInfo.dependencies` and never compiled.

At main-bundle time (`packages/deployer/src/build/bundler.ts`, `externalsPreset` branch), the `alias-optimized-deps` plugin only resolves ids present in `analyzedBundleInfo.dependencies`. The uncaptured subpath misses that check and falls through to `nodeModulesExtensionResolver` (`packages/deployer/src/build/plugins/node-modules-extension-resolver.ts`), where `getPackageJSON` throws `Package <pkg> not found` (the app has no direct dep on `@scope/b` to resolve from); the plugin **catches it, `console.error`s, and returns `null`** (`node-modules-extension-resolver.ts:80-83`). Rollup then treats the subpath as an unresolved bare import and externalizes it.

Two other spots ensure the leaked package never re-enters the manifest even if it were tracked:

- `packages/deployer/src/build/analyze.ts:552-561` skips workspace packages from `externalDependencies`.
- `packages/deployer/src/bundler/index.ts:356-363` skips any `workspaceMap` dep from `dependenciesToInstall`.

Net: the specifier is neither **bundled** (compiled inline) nor **installed** (registered in `package.json`) → `ERR_MODULE_NOT_FOUND`.

Note this is broader than packages with subpath-only `exports` maps (the crash in #19379) — the transitive walk compiles only roots, so *any* transitively-imported workspace subpath can leak. In a subpath-only package it just hits #19379 first.

### Minimal reproduction

pnpm workspace:

- `packages/leaf/package.json`: `{ "name": "@scope/leaf", "exports": { "./sub": "./src/sub.ts" } }`, with `src/sub.ts` exporting a value.
- `packages/mid/package.json`: `{ "name": "@scope/mid", "main": "src/index.ts", "dependencies": { "@scope/leaf": "workspace:*" } }`, and `src/index.ts` does `import { x } from '@scope/leaf/sub'; export const midValue = x;`
- `apps/app`: depends on `@scope/mid` **only** (not `@scope/leaf`); the Mastra entry imports `@scope/mid` and uses `midValue` in a registered API route (so it survives tree-shaking):

```ts
  server: { apiRoutes: [/* route using midValue */] },
  bundler: { externals: true },
});
```

`mastra build` in `apps/app`. Observed:

- stderr logs `Error: Package @scope/leaf not found` (non-fatal), build exits 0;
- `.mastra/output/index.mjs` contains a bare `import ... from '@scope/leaf/sub'`;
- `.mastra/output/package.json` has **no** `@scope/leaf` entry;
- running the output → `ERR_MODULE_NOT_FOUND`.

### Expected behavior

Either the transitive workspace subpath is **compiled into the bundle** (consistent with what `externals: true` already does for workspace roots), or it is **registered** as an installable dependency in the generated `package.json`. The output must be self-contained: copy `.mastra/output` alone, `npm ci`, `node index.mjs` should boot.

### Proposed fix

Resolve an uncaptured workspace subpath (recognized via `workspaceMap`) to its source through the package's `exports` map at main-bundle time and return it non-external so the bundler compiles it inline — no new analysis pass, so the existing manifest-based transitive-discovery behavior is preserved. I have a PR that does exactly this, alongside the #19379 fix, and will link it here.

### Environment

- Node 22, pnpm workspaces (also applies to Nx monorepos), `bundler.externals: true` (required for consumers with native `.node` deps, e.g. LiveKit).

mikhael28:
## Summary

Two related, additive fixes to `@mastra/deployer` for pnpm/Nx monorepos built with `bundler: { externals: true }` (the mode required when a consumer has native `.node` bindings, e.g. LiveKit). Both are behavior-preserving: any build that already succeeds is unchanged.

- Fixes #19379 — `mastra build`/`mastra dev` crashed with `Missing "." specifier in "<pkg>" package` when a workspace package declares only subpath exports (no `.` entry).
- Fixes #19909 — a workspace subpath imported transitively (by another workspace package) leaked out of the bundle as an unresolved bare specifier, unregistered in the generated `package.json` → `ERR_MODULE_NOT_FOUND` at runtime.

## Changes

**1. Subpath-only workspace exports (#19379)** — `build/analyze/analyzeEntry.ts`, `build/analyze/bundleExternals.ts`

The transitive-dependency walk registered every transitive workspace package as a synthetic root `export * from '<pkg>'` module — even for packages that cannot be imported at their root. In `noBundling` mode (`externals: true` / dev), the `alias-optimized-deps` resolver then called `resolve.exports(pkgJson, '.')`, which throws for a subpath-only `exports` map.

- `analyzeEntry.ts`: skip the synthetic root registration for workspace packages with no root entry (no `.` export, no `exports` map, or string sugar). The subpaths actually imported are captured independently, so nothing previously bundled is dropped.
- `bundleExternals.ts`: guard the `resolve.exports` root lookup so it falls back to `main`/`index` instead of throwing.

**2. Transitive workspace subpath imports (#19909)** — `build/bundler.ts`

Transitive-dependency discovery is manifest-name-based: it registers workspace package *roots*, but never reads a workspace package's *source* to learn which *subpaths* it imports from other workspace packages. Such a subpath (`@scope/a` importing `@scope/b/sub`, app depends on `@scope/a` only) was never captured, fell through the resolver, and leaked as a bare specifier.

- `bundler.ts`: in the main bundle, resolve an uncaptured workspace subpath (recognized via `workspaceMap`) to its source through the package's `exports` map and return it non-external, so the bundler compiles it inline. No new analysis pass is added, so the existing manifest-based transitive-discovery behavior is preserved.

## Why additive

- #19379 fix: for packages with a root, the register branch is always taken and the `resolve.exports` `try` always succeeds — byte-identical behavior. The change only affects the case that currently crashes.
- #19909 fix: only engages for a workspace subpath that is otherwise unresolved (would leak); every already-resolvable import is untouched.

## Testing

- New unit tests: subpath-only root-entry detection + skip logic (`analyzeEntry.test.ts`), the `resolve.exports` guard (`bundleExternals.test.ts`), and the workspace-subpath resolver (`bundler.workspace-subpath.test.ts`).
- Full `packages/deployer` build suite green (`vitest run src/build`, 420 tests), `tsc --noEmit` clean, `eslint` clean.
- Verified end-to-end on a reproduction monorepo: unfixed → `Missing "." specifier` crash; with the fixes → `mastra build` exits 0, the transitive subpath is compiled inline, no bare workspace specifier leaks into the output, and the generated `package.json` is self-contained.

Changeset included (`@mastra/deployer` patch).

## ELI5

Some workspace packages only provide specific subpaths instead of a main entry point. This update prevents deployer crashes and makes sure those packages are bundled correctly so the final application can run without missing-module errors.

## Summary

- Safely handles workspace packages with subpath-only exports when `bundler.externals: true`.
- Resolves transitive workspace subpath imports to source files and bundles them inline.
- Adds guarded root-package resolution with `main`/`index.js` fallbacks.
- Adds regression tests and a patch changeset.
- Build tests, TypeScript checks, and linting pass.


mikhael28:
@coderabbitai please review

wardpeet:
The above PR doesnt completely solve theissue, it marks deps as external when we it's better to collect the externals and use that instead of assuming the default import https://github.com/mastra-ai/mastra/pull/19808

Issue #19918, opened 2026-07-22
[BUG] MCPClient discovers tools from servers serially; one slow server stalls all discovery


rares04:
## Describe the Bug

`MCPClient` discovers tools from every configured server **serially**. `listToolsWithErrors()` (and `listToolsetsWithErrors()`, which back the public `listTools()` / `listToolsets()`) loop over the configured servers with a `for … await` loop, connecting to and listing tools from one server at a time:

```ts
// packages/mcp/src/client/configuration.ts — listToolsWithErrors()
for (const serverName of Object.keys(this.serverConfigs)) {
  try {
    const tools = await this.getToolsForServer(serverName); // connect + tools/list, awaited before the next server starts
    ...
  } catch (error) { ... }
}
```

Source: [configuration.ts#L1094-L1116](https://github.com/mastra-ai/mastra/blob/6193d6d4ae62ad68daaaf450992198e9e49493f1/packages/mcp/src/client/configuration.ts#L1094-L1116) (and the same pattern in `listToolsetsWithErrors`).

Because each server is awaited before the next one begins, discovery time is **additive** and one slow server blocks all the others:

- Every server contributes its full connect + `tools/list` latency to the total. With real remote MCP servers we observe ~500–600 ms per server in production, so total discovery time scales linearly with the number of servers.
- A single slow or unresponsive server **blocks every server behind it**. Each server is bounded only by its own request timeout (`DEFAULT_REQUEST_TIMEOUT_MSEC`, 60 s by default — see `client.ts`, where `connect()` and `tools()` both pass `{ timeout: this.serverConfig.timeout ?? this.timeout }`), so the worst case is roughly `timeout × serverCount`: one hung server can push total discovery toward 60 s even though every other server would have answered in milliseconds.

This is on a hot path — it runs on every agent construction that passes `tools: await mcp.listTools()`.

The sibling multi-server aggregation methods `resources.list()`, `resources.templates()`, and `prompts.list()` have the same serial pattern. Note that `disconnect()` already tears servers down in parallel via `Promise.allSettled`, so the serial discovery paths are an inconsistency within the same file.

## Steps To Reproduce

1. Configure an `MCPClient` with three or more servers, at least one of which is slow to respond (e.g. a remote server, or a local stdio server that sleeps before answering `tools/list`):

   ```ts
   import { MCPClient } from '@mastra/mcp';

   const mcp = new MCPClient({
     servers: {
       a: { url: new URL('https://server-a.example/mcp') },
[...]

PR #19920, opened 2026-07-22
perf(mcp): list resources and prompts from all MCP servers concurrently

rares04:
## Description

Follow-up to the tool-discovery parallelization (#19919), completing the pattern class. The same serial `for … await` loop exists in the other multi-server aggregation methods on `MCPClient`:

- `resources.list()`
- `resources.templates()`
- `prompts.list()`

Each connects to every configured server one at a time, so the same additive-latency and head-of-line-blocking behavior applies — a slow server delays the aggregate for every server behind it. This PR converts them to concurrent `Promise.all` fan-out, identical in shape to the tool-discovery change and to the parallel teardown `disconnect()` already uses.

Behavior is unchanged: results are keyed by server and folded back in configuration order, and per-server failures are still logged and skipped (these methods return only the servers that succeeded and never throw for a single server's failure).

Left intentionally untouched:

- `disconnect()` — already parallel via `Promise.allSettled`.
- `getServerInstructions()`, `sessionIds`, `toMCPServerProxies()` — synchronous reads of already-connected client state (no per-server I/O).
- Single-server methods (`resources.read/subscribe/...`, `prompts.get`, `elicitation.onRequest`, `progress.onUpdate`) — no multi-server iteration.

## Related issue(s)

Relates to #19918

## Type of change

- [ ] Documentation update
- [ ] Code refactoring
- [x] Performance improvement

## Checklist

- [x] I have made corresponding changes to the documentation (if applicable) — n/a, no API change
- [x] I have added tests that prove my fix is effective or that my feature works — covered by the existing resource/prompt suite; the concurrency pattern itself is proven by the tests in #19919

## Notes for reviewers

- Independent of #19919 in intent, but both edit `packages/mcp/src/client/configuration.ts`, so whichever merges second will need a trivial rebase.
- Split out from the tool-discovery PR to keep each change small and focused: that PR targets the measured hot path (`listTools` / `listToolsets`, called on every agent construction) and carries the concurrency-proof tests; resources/prompts listing is lower-traffic and mechanically identical.

## ELI5

Instead of asking each MCP server one at a time, the client now asks all servers at once, making resource and prompt discovery faster. Results stay organized in configuration order, and failed servers are still skipped without stopping everything.

## Changes

- Parallelized:
  - `resources.list()`
  - `resources.templates()`
  - `prompts.list()`
- Preserved per-server error logging and failure handling.
- Added a patch changeset for `@mastra/mcp`.
- Left single-server operations, `disconnect()`, and synchronous client-state methods unchanged.

### DIFF
--- a/.changeset/mcp-parallel-resource-prompt-discovery.md
+++ b/.changeset/mcp-parallel-resource-prompt-discovery.md
@@ -1,0 +1,7 @@
+---
+'@mastra/mcp': patch
+---
+
+List resources and prompts from all MCP servers concurrently in `MCPClient.resources.list()`, `resources.templates()`, and `prompts.list()`.
+
+These methods previously queried each configured server one at a time in a serial loop, so a slow or unresponsive server delayed the aggregate result for every server behind it. They now fan out with `Promise.all`, matching the tool-discovery methods and the parallel teardown in `disconnect()`. Results are keyed by server and assembled in configuration order, and per-server failures are still logged and skipped, so behavior is otherwise unchanged.
--- a/packages/mcp/src/client/configuration.ts
+++ b/packages/mcp/src/client/configuration.ts
@@ -322,24 +322,35 @@ To fix this you have three different options:
        */
       list: async (): Promise<Record<string, Resource[]>> => {
         const allResources: Record<string, Resource[]> = {};
-        for (const serverName of Object.keys(this.serverConfigs)) {
-          try {
-            const internalClient = await this.getConnectedClientForServer(serverName);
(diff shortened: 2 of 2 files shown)


rares04:
Leaving this closed — the automation closed it because `Relates to #19918` isn't a recognized closing keyword, so it never got a linked issue.

Rather than reopen it as a second PR against the same issue, I've folded these resources/prompts changes into #19919 (which already `Fixes #19918`). Both were the same mechanical `Promise.all` change to the same file, so a single reviewable PR under one changeset is cleaner and avoids the second-PR rebase. Continuing there.

PR #19921, opened 2026-07-22
fix(observability): preserve reported Gateway costs

### LINKED ISSUE #19895: [FEATURE] Preserve provider-reported per-step cost on MODEL_GENERATION spans
> **From the Zynda team.** We're building a production AI chat product on Mastra. This is not currently blocking us because an application-level bridge is possible, and #19513 restores registry estimates for Gateway models. However, Mastra still discards the provider-reported per-call cost, so Studio cannot distinguish or display the amount reported by the Gateway. We'd appreciate guidance on the intended framework boundary and are happy to contribute a PR.

### Summary

Vercel AI Gateway returns a monetary cost on completed AI SDK steps as `providerMetadata.gateway.cost`. Mastra-managed agent generations receive that metadata, but the resulting `MODEL_GENERATION` span and automatic metrics do not preserve it.

The recently merged #19513 fixes the separate `gateway` -> `vercel` pricing-registry mismatch, so Gateway generations can receive **snapshot-based estimates**. Its PR description explicitly notes that costs remain snapshot-based. The provider-reported amount is still dropped before span/metric export.

This matters beyond accuracy: a multi-step agent generation can make several model calls. Mastra aggregates token usage across the steps, while top-level `providerMetadata` represents only the final step. A naive final-metadata parser would therefore undercount tool loops.

### Environment

- `@mastra/core@1.50.1`
- `@mastra/observability@1.16.0`
- `ai@6.0.168`
- `@ai-sdk/gateway@2.0.82`
- Node 22

I also reviewed merged PR #19513; it changes pricing-provider resolution only and does not consume provider-reported Gateway cost.

### Current behavior

The underlying AI SDK result contains a parseable Gateway cost:

```ts
import { createGateway } from '@ai-sdk/gateway';
import { generateText } from 'ai';

const gateway = createGateway();
const result = await generateText({
  model: gateway('xiaomi/mimo-v2.5-pro'),
  prompt: 'Reply with exactly: ok',
  maxOutputTokens: 8,
});

console.log(result.providerMetadata?.gateway?.cost);
// Example: "0.000028875"
```

For a Mastra-managed agent call using the same Gateway model:

- The `MODEL_GENERATION` span receives aggregate token usage.
- The framework passes `providerMetadata` into `ModelSpanTracker.endGeneration()`.
- `endGeneration()` uses that metadata for usage normalization but does not derive or attach a `costContext` from it.
- Automatic metric extraction reads an existing `attributes.costContext` first, otherwise it falls back to `PricingRegistry` estimation.
- Therefore the Gateway-reported amount is not present in the exported span or metrics. On current packages the metric can be missing; after #19513 it becomes a registry estimate, not the provider-reported value.

### Why the application workaround is brittle

An application can capture each step's `providerMetadata.gateway.cost` in `onStepFinish`, temporarily correlate it by run/response ID, then inject a `costContext` from a `SpanOutputProcessor` before automatic metric extraction.

That requires every Mastra application to implement same-process correlation, callback deduplication, expiry, multi-step aggregation, durable-run behavior, and duplicate-accounting safeguards for a span lifecycle that Mastra already owns.

### Suggested framework direction

The natural boundary appears to be model-generation finalization, where Mastra already owns the span and has access to steps, aggregate usage, and provider metadata:

1. Preserve a small per-step completion summary through generation finalization: usage, model/provider identity, response identity where available, and provider metadata needed for cost extraction.
2. Add an extensible provider-cost extractor in observability, with a built-in Vercel AI Gateway extractor once the metadata contract is considered stable.
3. Resolve cost once per completed step, then aggregate once onto the generation span/metric.
4. Prefer an explicitly supplied cost contribution, then provider-reported cost, then registry estimation; never count multiple sources for the same step.
5. Treat explicit zero as valid, but reject missing, malformed, negative, or non-finite values rather than converting them to zero.
6. Avoid persisting raw `providerMetadata`; export only amount, currency, provenance, and safe provider/model dimensions.

Conceptually, an additive contract could look like:

```ts
interface ProviderReportedCost {
  amount: number;
  currency: string;
  source: string;
  provider?: string;
  model?: string;
}

interface ModelCostExtractor {
  id: string;
  extract(input: {
    providerMetadata?: ProviderMetadata;
    usage?: LanguageModelUsage;
    provider?: string;
    model?: string;
    responseId?: string;
  }): ProviderReportedCost | undefined;
}
```

The existing `CostContext.estimatedCost` field can technically carry the number with provenance metadata, but a distinct `providerReportedCost` field would make the semantics clearer:

```ts
interface CostContext {
  providerReportedCost?: number;
  estimatedCost?: number;
  costUnit?: string;
  costMetadata?: Record<string, unknown>;
}
```

Studio could then label the value **Reported**, **Estimated**, or **Mixed**. "Reported" is safer than "actual" because the runtime amount may still differ from final invoice settlement.

### Multi-step and streaming semantics

- AI SDK aggregate usage covers all completed steps, but top-level provider metadata is final-step metadata. Provider cost therefore needs per-step handling.
- Each step must contribute at most once, including callback retries/re-entry.
- If some steps are reported and others estimated, provenance should be `mixed` rather than presenting the total as wholly reported.
- If a stream ends without terminal provider metadata, Mastra should not fabricate zero or block span finalization on a follow-up Gateway request.
- An existing caller-supplied `costContext` should win so manually traced/raw-SDK calls are not double-counted.

### Why this belongs in Mastra

Mastra owns the `MODEL_GENERATION` span, receives the AI SDK step metadata, aggregates usage, ends the generation tracker, and extracts the metrics consumed by Studio. The relevant data reaches that framework-owned lifecycle and is then discarded for cost purposes, so application callbacks and a shared map should not be the default integration.

The AI SDK/Gateway package should ideally document and type the currency/semantics of `providerMetadata.gateway.cost`, but an AI SDK change alone cannot attach the value to Mastra's span and metrics.

### Related

- #19461 / #19513: restores Vercel Gateway **registry estimates**, but intentionally remains snapshot-based
- #18515: equivalent request for OpenRouter provider-reported cost; currently closed although comments indicate the native-cost implementation may still be absent

Happy to provide the live probe, exported span/metric samples, or work on a PR once the preferred contract is confirmed.

Abdo-Eid:
Fixes #19895.

Mastra now forwards provider metadata for every completed AI SDK step and uses the Vercel AI Gateway reported cost when every step provides a valid amount. Existing caller-supplied cost contexts still take precedence, and missing, malformed, or negative reported costs fall back to the pricing registry rather than producing a partial total.

A live Gateway probe confirmed that a numeric-string providerMetadata.gateway.cost is preserved unchanged on the completed MODEL_GENERATION span. Also validated with @mastra/core type checking, focused model-tracing and auto-extraction tests, the full observability suite (975 passed, 2 skipped), package lint, and the observability build.

## ELI5

Mastra now keeps the cost that Vercel AI Gateway reports for each AI generation step. It adds valid costs across all steps and uses estimates only when reported costs are incomplete or invalid.

## What changed

- Preserved provider metadata for each completed AI SDK step.
- Added Gateway cost parsing and validation:
  - Supports numbers and numeric strings.
  - Accepts zero.
  - Rejects missing, malformed, negative, and non-finite values.
- Aggregated valid reported costs across multi-step generations.
- Preserved caller-supplied `costContext` over provider-reported costs.
- Prevented partial totals and final-step-only totals.
- Preserved cost provenance and reported-step metadata for automatic metrics.
- Added regression tests for aggregation, validation, fallback behavior, metadata scope, and precedence.
- Added automatic metrics documentation for provider-reported totals.
- Added patch changesets for `@mastra/core` and `@mastra/observability`.

## Validation

Validated with Gateway probing, type checking, model-tracing and automatic-extraction tests, the observability suite, linting, and the observability build.


abhiaiyer91 (dismissed):
**Verdict: Request changes** — one correctness gap in the non-`stepProviderMetadata` fallback, plus a doc statement this PR makes inaccurate. Everything else (design, precedence, aggregation rules, tests) looks solid.

## What I verified

- `pnpm turbo build --filter ./observability/mastra` — success
- `pnpm --filter ./observability/mastra test` — 976 passed, 2 skipped (34 files)
- `pnpm --filter ./observability/mastra exec tsc --noEmit` — clean
- `pnpm --filter ./packages/core exec tsc --noEmit` — clean
- `pnpm --filter ./observability/mastra lint` — clean
- CI: all functional checks pass; the two failing Vercel checks are `Authorization required to deploy` (team-side, not this PR)

The core design is right: aggregate per-step, refuse partial totals, treat explicit `0` as valid, and never overwrite a caller-supplied `costContext`. The five new tests cover each of those rules, and the precedence test against the Claude-SDK-style `costContext` is a good guard.

## Requested changes

### 1. `stepProviderMetadata ?? [providerMetadata]` can report a silent undercount as an authoritative total

`observability/mastra/src/model-tracing.ts:61` falls back to the single top-level `providerMetadata` when the caller does not supply per-step metadata. That top-level value is the **last step's** metadata, not a merged view — see `packages/core/src/stream/base/output.ts:1055-1058`, where the `onFinish` payload takes `providerMetadata` from `baseFinishStep` (the final buffered step).

Inside this repo the fallback is unreachable, because `packages/core/src/llm/model/model.loop.ts:317` always passes `stepProviderMetadata`. But `observability/mastra/package.json:49` declares `"@mastra/core": ">=1.16.0-0 <2.0.0-0"`, so a released `@mastra/core` 1.16.x that predates this PR paired with the new `@mastra/observability` is a supported install. In that combination, a multi-step tool loop through the AI Gateway produces:

- `estimatedCost` = final step's cost only
- `costMetadata.scope = 'query_total'`, `source = 'provider_reported'`, `reportedStepCount: 1`

and because a provided `costContext` short-circuits registry estimation (`observability/mastra/src/metrics/auto-extract.ts:75-78`), the registry estimate that would have covered all steps is suppressed. The result is a cost that is both wrong and labelled as provider-reported truth — exactly the tool-loop undercount #19895 warned about.

This file already establishes the pattern for handling that skew: `supportsModelInference()` at `observability/mastra/src/model-tracing.ts:37-39` exists specifically because a user's `@mastra/core` may predate a feature. So the skew is a real, acknowledged case here, not a hypothetical.

Suggested fix: fail closed — require `stepProviderMetadata` and drop the `[providerMetadata]` fallback, so older cores keep getting registry estimates instead of a mislabelled partial total. The "preserves an explicit zero" test would move to `stepProviderMetadata: [{ gateway: { cost: 0 } }]`, and it would be worth adding a case asserting that top-level-only `providerMetadata` yields no `costContext`.

### 2. Docs statement is now inaccurate

`docs/src/content/en/reference/observability/metrics/automatic-metrics.mdx:72` says cost context is attached "when the embedded pricing registry has a matching entry for the provider and model." After this PR, Gateway spans can carry cost context sourced from the provider instead of the registry, and the cost lands on a single carrier metric as a query total rather than per-meter estimates. Please update that section to describe provider-reported cost and the `source` / `scope` / `reportedStepCount` fields, per the repo rule that behavior changes need matching docs updates.

## Non-blocking nits

- `.changeset/hip-glasses-stop.md`: "Preserve completed-step provider metadata when ending model generations" describes the internal handoff. The changeset guidance asks for developer-facing outcomes — something like "Multi-step model generations now keep per-step provider metadata, so traces can report the provider's own cost for the whole run." Same point CodeRabbit raised.
- Test coverage: `parseGatewayCost` already rejects `NaN`, `Infinity`, and malformed strings, and the negative-cost case is covered. Adding one table-driven case over `['abc', NaN, Infinity]` would lock that in cheaply.
- `parseGatewayCost` keys off `providerMetadata.gateway` without checking the span's provider. Only the Vercel AI Gateway is expected to populate that namespace today, so this is fine, but it is worth keeping in mind if another provider ever uses a `gateway` namespace with different semantics.
- Summing floats across steps can yield values like `0.0030000000000000004`. Not worth changing on its own, just noting that consumers should format rather than compare exactly.

## Adversarial check

I tried to argue the fallback is out of scope — that `@mastra/core` and `@mastra/observability` always move together, since this PR's changesets bump both. The peer range and the existing `supportsModelInference()` gate in the same file are what convinced me otherwise: this package explicitly supports running against an older core, and here it degrades to a wrong number rather than to no number.

I also checked the opposite risk — that requiring `stepProviderMetadata` loses value for single-step calls under an older core. It does, but the lost case is one where the registry estimate is still available and reasonable, while the retained case avoids publishing an incorrect total under a `provider_reported` label. Failing closed is the better trade for cost data.

Abdo-Eid:
@abhiaiyer91 Addressed the requested changes in 483b85d.
- Removed the top-level providerMetadata fallback, so installations with older @mastra/core versions fall back to registry estimates instead of publishing a final-step cost as a query total.
- Added regression coverage for top-level-only metadata and table-driven invalid reported costs: malformed string, negative value, NaN, and Infinity.
- Updated the automatic metrics reference for provider-reported totals and their source, scope, and reportedStepCount metadata.
- Reworded the core changeset around the user-facing multi-step outcome.\n\nValidation:\n- @mastra/observability: 980 passed, 2 skipped.
- Live Mastra Agent generation through Vercel AI Gateway: the completed MODEL_GENERATION span included a provider_reported query_total costContext with reportedStepCount: 1.
- Please re-review when convenient.

abhiaiyer91 (approved):
Verified the complete provider-reported Gateway cost flow and the fail-closed behavior for partial or invalid step metadata. Added aggregate overflow validation, strengthened caller-precedence coverage, and documented per-step aggregation and precedence.

Verification:
- pnpm build:core
- pnpm build:observability
- pnpm test:observability (1708 passed, 7 skipped)
- focused model-tracing tests (51 passed)
- observability oxlint/oxfmt
- MDX format, Remark, and Vale

The full core suite reached 13,258 passing tests but remains blocked by an unrelated timeout in the existing streamLegacy cleanup test; the same focused test also times out and this PR does not touch that workflow path.

PR #19929, opened 2026-07-22
fix: explain locked Mastra Code storage startup

TylerBarnes:
## Summary

- detect SQLite lock errors hidden inside startup and cleanup failures
- preserve the fail-closed journal transition
- tell users to close all running Mastra Code sessions once and restart
- preserve existing behavior for unrelated startup and cleanup errors

## Test plan

- `pnpm --filter ./mastracode/sdk exec vitest run src/__tests__/index.test.ts --reporter=dot --bail 1`
- `pnpm --filter ./mastracode/sdk check`
- `pnpm --filter ./mastracode/sdk lint`
- `pnpm build:mastracode`
- isolated live CLI startup against a temporary WAL database held by a separate process

## ELI5

Mastra Code now recognizes when another session is using its local database during startup. It gives users clear instructions to close other sessions and restart, while preserving the existing handling for unrelated errors.

## What changed

- Added SQLite busy/locked error detection, including nested causes and errors.
- Improved startup cleanup failure handling by preserving both startup and cleanup failures.
- Added recovery guidance for local storage journaling failures caused by another process.
- Added tests for locked, busy, nested, and unrelated cleanup failures.
- Added a patch changeset for `@mastra/code-sdk`.

## Validation

- SDK tests
- Type checking
- Linting
- Mastra Code build
- Isolated CLI startup test using a temporary WAL database held by another process

### DIFF
--- a/.changeset/lemon-news-matter.md
+++ b/.changeset/lemon-news-matter.md
@@ -1,0 +1,5 @@
+---
+'@mastra/code-sdk': patch
+---
+
+Show actionable recovery instructions when another Mastra Code process blocks the local storage safety transition.
--- a/mastracode/sdk/src/__tests__/index.test.ts
+++ b/mastracode/sdk/src/__tests__/index.test.ts
@@ -414,6 +414,47 @@ describe('createMastraCode', () => {
     expect(closeVector).toHaveBeenCalledOnce();
   }, 10_000);
 
+  it('explains how to recover when another process locks local storage during startup', async () => {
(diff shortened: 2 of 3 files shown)


abhiaiyer91 (approved):

TylerBarnes:
Closing because #19931 reverted the underlying local storage safety transition after startup and live Observational Memory regressions. This messaging follow-up no longer applies while the storage rollout is paused.

PR #19940, opened 2026-07-22
fix(core): keep sequential allow tool results after auto-approved resume

### LINKED ISSUE #19814: [BUG] AgentController loses the second auto-approved allow tool result across sequential resumes
### Describe the Bug

In a non-yolo `AgentController` session, the controller sets core `requireToolApproval: true` for every agent call. When the effective permission policy for a tool is `allow`, the controller consumes the emitted approval chunk and immediately calls `approveToolCall()` itself.

The first auto-approved tool call resumes correctly. When the model makes a second sequential call in the same run, the second tool executes successfully, but the resumed `MessageList` was hydrated from the previous suspended snapshot. `MessageList.updateToolInvocation()` therefore returns `false` for the second `toolCallId` (and emits `updateToolInvocation: no matching tool call found ...` when its logger is attached). The next model request contains neither the matching assistant tool-call message nor the tool result. With normal `toolChoice: "auto"` behavior, the model cannot decide based on the result and can repeat the tool sequence.

This reproduces on current Mastra `main`, not only on `@mastra/core` 1.51.0:

| Revision / store | Result |
| --- | --- |
| `@mastra/core` 1.51.0 + `@mastra/pg` 1.16.0 + Postgres | Reproduced |
| Mastra `main` `88a28f3c10c6111013b09b91ad3261c08dcdc28b`, core `1.52.0-alpha.8`, `InMemoryStore` | Reproduced |
| Same current `main`, `PostgresStore` | Reproduced |

No artificial storage delay is required. Postgres-like timing makes the underlying race easier to recognize, but the focused current-main test also fails with `InMemoryStore`.

The minimal conditions are two sequential tool calls in one non-yolo AgentController run with the tool's effective policy set to `allow`. Postgres, memory, an external model, and an artificial delay are not required for the smallest failing upstream test.

### Steps To Reproduce

1. Clone the minimal reproduction and install dependencies:

```bash
pnpm install --frozen-lockfile
```

2. Start its isolated Postgres container:

```bash
pnpm db:up
```

3. Run the deterministic mocked-model reproduction:

```bash
pnpm repro
```

4. Observe that the tool executes twice and the model is called three times, with no user-facing `tool_approval_required` event. The relevant output is:

   ```json
   {
     "toolExecutions": 2,
     "modelCalls": 3,
     "emittedUserApprovalEvent": false,
     "invocationUpdates": [
       { "toolCallId": "allow-call-1", "updated": true },
       { "toolCallId": "allow-call-2", "updated": false }
     ]
   }
   ```

5. Inspect the recorded model prompts. Before the second tool result is mapped, model call 2 is correct:

   ```text
   system
   user
   assistant: tool-call allow-call-1
   tool: tool-result allow-call-1 = RESULT_REACHED_MODEL_7F3A
   ```

   After the second tool executes, model call 3 is incorrect:

   ```text
   system
   user
   ```

   The expected `assistant: tool-call allow-call-2` and matching sentinel tool result are both absent.

6. Optionally rerun with snapshot tracing:

   ```bash
   TRACE_STORAGE=1 corepack pnpm repro
   ```

   The decisive ordering is:

   ```text
   persist suspended [allow-call-1]
   load    suspended [allow-call-1]
   load    suspended [allow-call-1]
   load    suspended [allow-call-1]
   persist suspended [allow-call-1, allow-call-2]
   ```

7. Stop and remove the isolated test database:

   ```bash
   corepack pnpm db:down
   ```

### Current-main verification commands

From a fresh Mastra clone at `88a28f3c10c6111013b09b91ad3261c08dcdc28b`:

```bash
corepack pnpm install --frozen-lockfile
corepack pnpm turbo build --filter ./packages/core
corepack pnpm --filter ./packages/core test src/agent-controller/__tests__/agent-controller-allow-result-retention.repro.test.ts
corepack pnpm --filter ./stores/pg exec vitest run src/storage/agent-controller-allow-result-retention.repro.test.ts --reporter=verbose
```

Both focused tests fail because model prompt 3 does not contain `allow-call-2`. The in-memory diagnostic also records:

```json
{"invocationUpdates":[{"toolCallId":"allow-call-1","updated":true},{"toolCallId":"allow-call-2","updated":false}]}
```

The focused tests disable normal logger output, so they do not print the warning text. The `false` result is the exact branch at `message-list.ts:1175` that calls `logger.warn("updateToolInvocation: no matching tool call found ...")` when that resumed list has a logger, matching the reported production warning.

### Current control flow

1. `AgentController.buildSharedRunOptions()` sets `requireToolApproval: !isYolo` for both initial and resumed non-yolo runs, irrespective of the effective per-tool policy ([agent-controller.ts lines 1562-1574](https://github.com/mastra-ai/mastra/blob/88a28f3c10c6111013b09b91ad3261c08dcdc28b/packages/core/src/agent-controller/agent-controller.ts#L1562-L1574)).
2. On `tool-call-approval`, the run engine resolves the session policy. `allow` immediately invokes `session.approveToolCall()` and does not emit a user approval event ([session-run-engine.ts lines 549-570](https://github.com/mastra-ai/mastra/blob/88a28f3c10c6111013b09b91ad3261c08dcdc28b/packages/core/src/agent-controller/session-run-engine.ts#L549-L570)). Policy precedence is implemented in [session.ts lines 2859-2885](https://github.com/mastra-ai/mastra/blob/88a28f3c10c6111013b09b91ad3261c08dcdc28b/packages/core/src/agent-controller/session.ts#L2859-L2885).
3. `approveToolCall()` calls `agent.sendToolApproval()` with the same run ID and `requireToolApproval: !isYolo` ([session.ts lines 3380-3411](https://github.com/mastra-ai/mastra/blob/88a28f3c10c6111013b09b91ad3261c08dcdc28b/packages/core/src/agent-controller/session.ts#L3380-L3411)).
4. Core emits the approval chunk before it flushes messages and calls `suspend()` with serialized stream state ([tool-call-step.ts lines 513-560](https://github.com/mastra-ai/mastra/blob/88a28f3c10c6111013b09b91ad3261c08dcdc28b/packages/core/src/loop/workflows/agentic-execution/tool-call-step.ts#L513-L560)). This lets AgentController begin the immediate resume before the new snapshot is persisted.
5. `waitForSuspendedSnapshot()` stops as soon as any snapshot for the run has `status === "suspended"` ([workflows/utils.ts lines 602-620](https://github.com/mastra-ai/mastra/blob/88a28f3c10c6111013b09b91ad3261c08dcdc28b/packages/core/src/workflows/utils.ts#L602-L620)). On the second suspension, the previous call's snapshot already satisfies that predicate.
6. Current core recognizes this stale-snapshot possibility in `#validateSuspendedToolCallTarget()`: it polls until a snapshot containing the requested `toolCallId` appears. However, it retains only a boolean and discards `latestSnapshot` ([agent.ts lines 6513-6542](https://github.com/mastra-ai/mastra/blob/88a28f3c10c6111013b09b91ad3261c08dcdc28b/packages/core/src/agent/agent.ts#L6513-L6542)).
7. `resumeStream()` consequently passes the original stale `existingSnapshot` into `resumeContext`, even after validation observed the fresh snapshot ([agent.ts lines 8448-8476](https://github.com/mastra-ai/mastra/blob/88a28f3c10c6111013b09b91ad3261c08dcdc28b/packages/core/src/agent/agent.ts#L8448-L8476), [lines 8513-8527](https://github.com/mastra-ai/mastra/blob/88a28f3c10c6111013b09b91ad3261c08dcdc28b/packages/core/src/agent/agent.ts#L8513-L8527)).
8. The agent loop selects `__streamState` from that suspended snapshot ([loop.ts lines 129-139](https://github.com/mastra-ai/mastra/blob/88a28f3c10c6111013b09b91ad3261c08dcdc28b/packages/core/src/loop/loop.ts#L129-L139)), and `MastraModelOutput.deserializeState()` replaces its `MessageList` with the serialized list ([output.ts lines 1907-1928](https://github.com/mastra-ai/mastra/blob/88a28f3c10c6111013b09b91ad3261c08dcdc28b/packages/core/src/stream/base/output.ts#L1907-L1928)). This is where the assistant call-2 message should enter the resumed list; it is absent because call-1 state was deserialized instead.
9. The mapping step tries to attach the result to the corresponding assistant invocation ([llm-mapping-step.ts lines 503-539](https://github.com/mastra-ai/mastra/blob/88a28f3c10c6111013b09b91ad3261c08dcdc28b/packages/core/src/loop/workflows/agentic-execution/llm-mapping-step.ts#L503-L539)). `MessageList.updateToolInvocation()` searches backward for a matching assistant part and returns `false`/warns when none exists ([message-list.ts lines 1095-1176](https://github.com/mastra-ai/mastra/blob/88a28f3c10c6111013b09b91ad3261c08dcdc28b/packages/core/src/agent/message-list/message-list.ts#L1095-L1176)).

### Relationship to #16158, #17015, and #19377

This is the same underlying snapshot-visibility race family as [issue #16158](https://github.com/mastra-ai/mastra/issues/16158), but it is an uncovered repeated-suspension edge case in [fix #17015](https://github.com/mastra-ai/mastra/pull/17015), not a wholly distinct regression.

#17015 added a bounded poll and a test where storage returns `null`, `null`, then a suspended snapshot. That covers the first suspension, when no prior suspended row is available. It does not cover a second sequential suspension on the same run, where the stale prior row is already `suspended` and therefore passes the helper immediately.

[PR #19377](https://github.com/mastra-ai/mastra/pull/19377), included in `@mastra/core` 1.51.0, explicitly added polling for a newly surfaced sequential tool-call target. Its validator successfully sees the fresh snapshot but does not return it to `resumeStream()`/`resumeGenerate()`. The target check can therefore pass while execution still hydrates from the stale snapshot. That makes #19377 relevant evidence and the clearest current code location for the incomplete handoff.

### Smallest proposed Mastra-owned regression test

Add one `AgentController` unit test using `InMemoryStore`; Postgres is not required for the smallest proof:

1. Create a non-yolo controller session and one ordinary tool with effective policy `allow`.
2. Use a deterministic mock model that emits `allow-call-1`, then on its next invocation asserts the first call/result are visible and emits `allow-call-2`, then records the third prompt.
3. Return `RESULT_REACHED_MODEL_7F3A` from the tool.
4. Assert two executions, three model calls, and no `tool_approval_required` event.
5. Critically, assert model prompt 3 contains an assistant tool call with `toolCallId: "allow-call-2"` and the matching tool result containing the sentinel. Tool execution alone is not a sufficient assertion.

A PostgresStore version of the same test is useful integration coverage, but the in-memory test is the smallest deterministic regression test and currently fails on `main`.

### Likely fix boundary (hypothesis)

The fix belongs at the core resume snapshot-freshness boundary. The snapshot that proves the requested `toolCallId` is currently suspended must also be the snapshot passed into `resumeContext` and used to hydrate `MessageList`. One plausible shape is for `#validateSuspendedToolCallTarget()` (or a generalized `waitForSuspendedSnapshot()` predicate) to return the latest matching snapshot, then have `resumeStream()` and `resumeGenerate()` use that returned snapshot for authorization metadata and resume state.

Merely polling for `status === "suspended"` cannot distinguish successive suspensions on one run. A target-aware predicate, snapshot revision, or equivalent freshness token is needed. Separately, AgentController could avoid core suspension for effective `allow` policies by using a policy-aware `requireToolApproval` function, but that is a broader controller design change and is not required to fix the core stale-resume defect demonstrated here.

This section is a hypothesis about the fix boundary only; no fix has been implemented.

https://github.com/robo-trh/mastra-bug-agentcontroller-resume/

## Expected Behavior

An effective `allow` policy should require no user interaction. After each permitted tool executes, the next model prompt should retain the matching assistant tool-call message and tool result. With `toolChoice: "auto"`, the model should then decide whether to call another tool or finish; the controller should not treat the tool as terminal.

For the second call in this reproduction, model prompt 3 should be:

```text
system
user
assistant: tool-call allow-call-2
tool: tool-result allow-call-2 = RESULT_REACHED_MODEL_7F3A
```

Retaining older call/result pairs too is acceptable, but the immediately preceding matching call/result pair must be present.

### Environment Information

```block
System:
  OS: macOS 15.7.3
  CPU: (16) arm64 Apple M4 Max
  Memory: 6.08 GB / 64.00 GB
  Shell: 5.9 - /bin/zsh
Binaries:
  Node: 23.6.0
  npm: 11.11.1
Browsers:
  Chrome: 150.0.7871.125
  Safari: 18.6
npmPackages:
  @mastra/core: 1.51.0
  @mastra/memory: 1.23.0
  @mastra/pg: 1.16.0
  tsx: 4.22.4
  zod: 4.4.3
LLM Provider: deterministic MastraLanguageModelV2Mock (no external provider)
Database: pgvector/pgvector:0.8.4-pg16 in Docker

Fresh upstream verification:
  Mastra commit: 88a28f3c10c6111013b09b91ad3261c08dcdc28b
  packages/core version: 1.52.0-alpha.8


Generated with:


corepack pnpm dlx envinfo --system --browsers --binaries --npmPackages --npmGlobalPackages
```

### Verification

Souravrajvi0:
## Description

When a non-yolo `AgentController` auto-approves sequential tool calls with an effective `allow` policy, the second resume could hydrate `MessageList` from the previous suspended snapshot. Validation already saw the newer snapshot (the one containing `allow-call-2`), but `resumeStream` / `resumeGenerate` still passed the stale one into `resumeContext`.

That meant `updateToolInvocation()` could miss the second `toolCallId`, so the next model prompt lost both the assistant tool-call message and the tool result.

`#validateSuspendedToolCallTarget()` now returns the matching suspended snapshot, and resume uses that snapshot for hydration.

## Related issue(s)

Fixes #19814

## Type of change

- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [ ] Test update

## Checklist

## ELI5

When tools run one after another, the agent could forget the latest tool call and its result. This fix uses the newest saved state, so every tool call and result remains available.

## Changes

- Return the suspended snapshot that contains the resumed `toolCallId`.
- Use the validated snapshot for `resumeContext` and `MessageList` hydration in `resumeStream` and `resumeGenerate`.
- Preserve sequential auto-approved tool calls and results in later model prompts.
- Avoid relying only on `status === "suspended"` when checking snapshot freshness.
- Add regression coverage for two sequential `allow`-policy tool calls:
  - Two tool executions and three model calls
  - No approval event
  - Retention of the second tool call and its result
- Add a patch changeset for `@mastra/core`.


abhiaiyer91 (approved):
## Verdict: Approve

The fix is correct, minimal, and covered by a regression test that provably fails without it.

### What the change does

`#validateSuspendedToolCallTarget` already polled storage for a fresh snapshot when the target `toolCallId` was not yet suspended in the snapshot loaded at the top of `resumeStream`/`resumeGenerate` — but it discarded the fresher snapshot and only kept a boolean. Both call sites then hydrated the resume from the *stale* `existingSnapshot`. With sequential auto-approved (`allow`) tool calls, that snapshot only contains the first suspension, so `updateToolInvocation()` never resolves the second `toolCallId` and the next model prompt loses both the assistant tool-call message and its result.

The method now returns the snapshot that actually contains the suspension, and both call sites pass it into `resumeContext.snapshot`.

### Findings

None blocking.

Behavioral risk is low and well-bounded: when the initially loaded snapshot already contains the target `toolCallId` (the common single-approval and parallel-approval paths), `resumeSnapshot === snapshot` and behavior is byte-identical to before. The new value is only used on the path that previously proceeded with a known-stale snapshot. Downstream consumers of `resumeContext.snapshot` (`loop.ts` `__streamState` hydration, `#getSuspendedToolInfo`, tracing context) all get *more* correct input than before.

Non-blocking nit (not worth a round trip): the regression test asserts prompt contents via `JSON.stringify(prompt).includes(...)`, which is a loose containment check rather than a structural assertion on the tool-call/tool-result parts. It is unambiguous enough here given the sentinel value, and it correctly fails on the unfixed code.

### Verification

Executed locally on PR head `4e7375d`:

- `pnpm install --frozen-lockfile`, `pnpm build:core` — clean.
- New regression test with the fix: **1 passed**.
- New regression test with `packages/core/src/agent/agent.ts` reverted to the base commit: **1 failed** — `expected false to be true` at the `allow-call-2` assertion. The test genuinely guards the fix.
- Full `src/agent-controller` suite: **43 files / 398 tests passed**, no type errors.
- Resume/approval suites in `src/agent/__tests__` (`parallel-delegation-approval`, `resume-span-tracing`, `resume-stream-no-snapshot`, `suspended-run-discovery`, `tool-approval-recall-roundtrip`, `tool-approval-standalone-repro`, `tool-approval`): **54 tests passed**. `parallel-delegation-approval` specifically covers the `AGENT_RESUME_TOOL_CALL_NOT_SUSPENDED` rejection paths that share this method — unaffected.
- `pnpm --filter ./packages/core check` (`tsc --noEmit`): clean.
- No merge conflicts against `origin/main` (`git merge-tree` produced zero conflict markers). Branch is behind main but merges cleanly.

Pre-execution security inspection of the diff found no `postinstall`/`prepare`/`pretest` script changes, no CI workflow changes, no test-config changes, no network calls, and no credential access. All commands were run with `GH_TOKEN`/`GITHUB_TOKEN` stripped.

### Existing review signal

- **CodeRabbit** — completed, "No actionable comments were generated." Nothing to triage.
- **dane-ai-mastra** — informational triage; linked-issue check passed (#19814), complexity `low`.
- **changeset-bot** — informational; the 23-package count reflects all changesets on the branch, not this PR. This PR adds exactly one changeset, correctly scoped to `@mastra/core: patch`.
- **Vercel (docs-1.x, playground-ui)** — `Authorization required to deploy`. Infrastructure/auth, not a code defect.
- All other checks (Socket Security ×2, Superagent, Contributor trust, labeler) pass.
- No human reviews and no inline review threads exist on the PR.

### Adversarial check

The strongest case for requesting changes is that the returned snapshot is used for resume hydration while `snapshotMemoryInfo` and the FGA authorization check are still derived from the older `existingSnapshot` — an apparent inconsistency. It fails: both snapshots belong to the same `runId` and the same agentic-loop run, so thread/resource identity is invariant across them; deriving authorization from the pre-validation snapshot is also the safer ordering, since authorization must not depend on state that appears after the caller was authorized. The second-strongest case — that a scoped fix like this should be validated end-to-end rather than at the unit level — fails against the executed evidence above, including the confirmed red/green flip on the unfixed code.

### Requested changes

None.

### Assumptions

- Vercel deployment failures are treated as environment authorization, not PR defects, consistent with them failing identically on unrelated PRs.
- The `*.repro.test.ts` filename follows existing precedent in this package (`tool-approval-standalone-repro.test.ts`).

### Open questions

None.

thomashoneyman:
Also looking for this to land, as currently we're using `yolo` mode with the agent controller until this fix is in.

PR #19953, opened 2026-07-22
fix(memory): preserve OM continuation hints after buffered activation

### LINKED ISSUE #19809: [BUG] Observational memory: activation clobbers thread continuation hints, so the first post-activation observer pass loses "## Prior Thread Metadata" (and the main agent loses <current-task>)
### Summary

The OM observer's `<current-task>` / `<suggested-response>` output is stripped from the observation text and stored in exactly one durable place: `thread.metadata.mastra.om` ([`observer-agent.ts:1461-1470`](https://github.com/mastra-ai/mastra/blob/main/packages/memory/src/processors/observational-memory/observer-agent.ts) — "stored separately in thread metadata and injected dynamically"). Two consumers read it back:

1. the **next observer pass**, as the `## Prior Thread Metadata` continuity section ([`sync.ts:104-115`](https://github.com/mastra-ai/mastra/blob/main/packages/memory/src/processors/observational-memory/observation-strategies/sync.ts) → [`observer-agent.ts:1383-1403`](https://github.com/mastra-ai/mastra/blob/main/packages/memory/src/processors/observational-memory/observer-agent.ts)), and
2. the **main agent**, whose context gets `<current-task>` / `<suggested-response>` injected from the same fields (`observational-memory.ts` `buildContextSystemMessages`, ~L2513).

Buffered-observation **activation destroys that state**. `activate()` ([`observational-memory.ts:~3344-3366`](https://github.com/mastra-ai/mastra/blob/main/packages/memory/src/processors/observational-memory/observational-memory.ts)) unconditionally writes the last activated chunk's hints into thread metadata:

```ts
const newMetadata = setThreadOMMetadata(thread.metadata, {
  suggestedResponse: lastActivated.suggestedContinuation, // always undefined since #13678
  currentTask: lastActivated.currentTask,                 // always undefined since #13678
  threadTitle: chunkThreadTitle,
});
```

Buffered chunks carry no hints *by design* — the buffering strategy calls the observer with `skipContinuationHints: true` ([`async-buffer.ts:61-68`](https://github.com/mastra-ai/mastra/blob/main/packages/memory/src/processors/observational-memory/observation-strategies/async-buffer.ts)) — and `setThreadOMMetadata` shallow-merges (`{...existingOM, ...omMetadata}`), so the explicit `undefined`s clobber the persisted values. There is no fallback to the previous metadata.

Net effect, on **every** activation (mid-turn threshold, turn-start step-0, TTL, provider-change):

- the first observer pass of the next cycle receives **no** `## Prior Thread Metadata` and must re-derive conversational state from the truncated `## Previous Observations` block (default `previousObserverTokens` ≈ 2000) plus whatever raw messages are in its batch;
- the main agent's injected `<current-task>` / `<suggested-response>` vanish at the same moment;
- whatever the stateless observer pass invents is then persisted and becomes the new "truth" for both consumers.

### Real-world impact (how we found it)

Production Slack agent, one long research turn (heavy web-fetch output → several observation cycles in a single turn). Seven observer passes ran; `## Prior Thread Metadata` was present in the intra-cycle passes and absent in exactly the two passes that immediately followed a `data-om-activation` event. Timeline of the observer's state for the turn:

- passes 1–4: current-task "Analyzing/finalizing regulatory advice…", suggested-response "the assistant should **inform/confirm to the user**…" — correct; a reply was still owed.
- pass 5 (first pass after activation, no prior metadata, batch = pure tool results): re-derived state from compressed observations ending in "✅ … audit completed" and invented *"Monitoring user feedback… the assistant should **wait for the user to respond**"* — the reply had never been sent.
- pass 6 faithfully carried the hallucination forward; the main agent's injected `<suggested-response>` said "wait for the user", and the agent ended its turn with our no-reply tool, echoing that language almost verbatim. The user chased twice ("it has been 10 minutes and no reply") before getting an answer ~20 minutes late, after the agent re-did the entire research.

The failure needs no misbehaving model: a compliant observer given no continuity state and only "✅ completed" summaries makes exactly this inference.

### Why we believe this is a bug, not design

The wipe itself was deliberate — twice:

- #13606 ("clear stale OM continuation hints when omitted") introduced always-write-even-undefined at activation, with a test pinning it (`__tests__/observational-memory.test.ts:10665`).
- #13678 ("skip continuation hints during async buffered observations") stopped buffered chunks generating hints (they were stale by activation time) and explicitly relied on the `undefined` overwrite.

But #13568 ("observer context optimization") landed **ten days after** #13678 and made the same metadata fields the observer's continuity state (`## Prior Thread Metadata`) — necessary because that PR also truncated the observer's view of previous observations. The interaction was never revisited: #13568's tests never cross an activation boundary, and no test asserts the section's absence post-activation.

Two further points:

- Post-#13678, chunks structurally never carry hints, so the original rationale ("clear stale hints from older chunks") protects against nothing — the only thing the wipe removes today is the **fresh** state from the last full observation pass, which was live in the main agent's context the instant before activation.
- "The observer doesn't need it because the main agent has it" doesn't hold: the main-agent injection reads the *same* wiped fields, so after activation no copy survives anywhere.

The codebase itself treats preservation as the norm everywhere else: `reflect()` (`observational-memory.ts:~3530`), `reflector-runner.ts:93-98`, and `resource-scoped.ts:410-414` all use `metadataUpdate.x ?? previous?.x`. `activate()` is the one writer without the fallback.

### Suggested fix

Mirror the `reflect()` pattern in `activate()`:

```ts
const previousOmMetadata = getThreadOMMetadata(thread.metadata);
const newMetadata = setThreadOMMetadata(thread.metadata, {
  suggestedResponse: lastActivated.suggestedContinuation ?? previousOmMetadata?.suggestedResponse,
  currentTask: lastActivated.currentTask ?? previousOmMetadata?.currentTask,
  threadTitle: chunkThreadTitle ?? previousOmMetadata?.threadTitle,
});
```

This can't regress #13678's stale-hint bug (designed-path chunks never carry hints; legacy chunks that do still win under `??`, preserving "latest chunk wins"). It requires reworking the pinned test at `observational-memory.test.ts:10665` — its "Stale task before activation" fixture is precisely the continuity that should survive. Optional hardening: also pass `priorCurrentTask`/`priorSuggestedResponse` read-only into the buffering observer (`async-buffer.ts:61-68`, keeping `skipContinuationHints`), so buffered observations are written with knowledge of the ongoing task.

### Steps To Reproduce

1. Configure OM (thread scope) with async buffering, e.g. `observation: { messageTokens: 600, bufferTokens: 300, bufferActivation: 0.8 }` (small thresholds for speed; prod equivalent 30000/6000/0.8).
2. Run a full observer pass — its `<current-task>`/`<suggested-response>` are persisted to thread OM metadata.
3. Run a second full pass: its prompt contains `## Prior Thread Metadata` with `- prior current-task:` / `- prior suggested-response:`. Carry-forward works.
4. Let async buffering create a chunk (hint-free by design), then let it activate — the same `om.activate()` the agent loop calls mid-turn.
5. Run the next full observer pass.

**Actual:** the section is missing entirely; the thread metadata keys were overwritten with `undefined` at activation.

There is a repro test on my fork, linked below

https://github.com/Bickio/mastra/blob/om-prior-metadata-repro/packages/memory/src/processors/observational-memory/__tests__/activation-prior-metadata.test.ts

### Expected Behavior

**Expected:** the post-activation pass still receives the last known current-task/suggested-response as `## Prior Thread Metadata`.

### Environment Information

```block
- `@mastra/memory` 1.23.0, `@mastra/core` 1.51.0 (also present on current `main`)
- OM config: thread scope, `observation: { messageTokens: 30_000 }`, defaults elsewhere (`bufferTokens: 6000`, `bufferActivation: 0.8`)
```

### Verification

---

## Consolidated from #22361 (closed as duplicate)

#22361 reported the same activation-clobbers-`currentTask` behaviour from a different angle. Details from that report that are not covered above:

**Also reproduced on newer versions:** `@mastra/memory` 1.27.0 / `@mastra/core` 1.60.0 (the report above was on 1.23.0 / 1.51.0). The bug is still present.

**Explicit `continuationHints` config does not help.** Setting

```ts
observationalMemory: {
  observation: {
    messageTokens: 1_000,
    bufferTokens: 200,
    bufferOnIdle: true,
    continuationHints: {
      currentTask: true,
      suggestedResponse: false,
    },
  },
}
```

still yields a missing `metadata.mastra.om.currentTask` after activation, because `AsyncBufferObservationStrategy` overrides it with `skipContinuationHints: true` and `filterObserverExtractors()` in `observer-runner.ts` drops the `current-task` / `suggested-response` extractors regardless of user config.

**Documentation mismatch.** The packaged Observational Memory docs state:

> Buffered observations also include continuation hints, a suggested next response and the current task.

The implementation contradicts this: buffered chunks never carry hints (`skipContinuationHints: true`), and `activate()` is a pure storage operation that cannot regenerate them. Whichever fix lands, the docs or the behaviour need to be reconciled.

**Alternative fixes proposed in #22361** (in addition to the `??` fallback in `activate()` suggested above):

- Keep `current-task` enabled during async buffering while optionally skipping only `suggested-response`, if stale-response risk is the concern; or
- Regenerate continuation hints during activation from the latest activated chunk plus the remaining raw tail.

**Test expectation:** a regression test should cover async buffering followed by activation and verify the next context contains both the activated observations and a non-empty current task. (The repro test linked above already covers the `## Prior Thread Metadata` side of this.)

Souravrajvi0:
## Description

Buffered observation chunks are observed with `skipContinuationHints: true`, so they do not carry `currentTask` or `suggestedResponse`. During activation, those missing fields were still written into thread metadata as `undefined`, which overwrote the persisted continuation hints used by:

- the next observer pass (`## Prior Thread Metadata`)
- the main agent context injection (`<current-task>` / `<suggested-response>`)

Activation now falls back to the existing thread OM metadata when an activated chunk does not provide new hint values.

## Related issue(s)

Fixes #19809

## Type of change

- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [x] Test update

## Checklist

## ELI5

When saved observation chunks are activated, they now keep the task and response guidance already stored for the conversation instead of accidentally erasing it. New guidance still replaces old guidance when provided.

## Changes

- Updated observational memory activation to fall back to existing OM metadata when buffered chunks omit continuation hints.
- Preserved `currentTask`, `suggestedResponse`, and `threadTitle` during activation when no new values are available.
- Added regression coverage for preserving continuation hints.
- Added a patch changeset for `@mastra/memory`.

### DIFF
--- a/.changeset/dry-carrots-occur.md
+++ b/.changeset/dry-carrots-occur.md
@@ -1,0 +1,5 @@
+---
+'@mastra/memory': patch
+---
+
+Fixed observational memory activation clearing thread continuation hints when buffered chunks have no hints (#19809)
--- a/packages/memory/src/processors/observational-memory/__tests__/observational-memory.test.ts
+++ b/packages/memory/src/processors/observational-memory/__tests__/observational-memory.test.ts
@@ -1,6 +1,7 @@
 import { MockLanguageModelV2, convertArrayToReadableStream } from '@internal/ai-sdk-v5/test';
 import type { MastraDBMessage, MastraMessageContentV2 } from '@mastra/core/agent';
 import { coreFeatures } from '@mastra/core/features';
+import { getThreadOMMetadata } from '@mastra/core/memory';
 import { MASTRA_THREAD_ID_KEY, RequestContext } from '@mastra/core/request-context';
 import { InMemoryMemory, InMemoryDB } from '@mastra/core/storage';
 import { describe, it, expect, beforeEach, afterEach, vi } from 'vitest';
@@ -9735,6 +9736,68 @@ describe('Async Buffering Processor Logic', () => {
       expect(result.activated).toBe(true);
       expect(result.record).toBeDefined();
       expect(result.record.activeObservations).toContain('Important observation');
+    });
+
+    it('should preserve existing continuation hints when activating buffered chunks without hints (#19809)', async () => {
+      const storage = createInMemoryStorage();
+      const threadId = 'thread-1';
+      const resourceId = 'resource-1';
+      const om = new ObservationalMemory({
+        storage,
+        scope: 'thread',
+        model: createStreamCapableMockModel({ defaultObjectGenerationMode: 'json' }),
+        observation: {
+          messageTokens: 50000,
+          bufferTokens: 10000,
+          bufferActivation: 1,
+        },
+        reflection: { observationTokens: 20000, bufferActivation: 0.5 },
+      });
+
+      await storage.saveThread({
+        thread: {
+          id: threadId,
+          resourceId,
+          title: 'Test Thread',
+          createdAt: new Date(),
+          updatedAt: new Date(),
+          metadata: {
+            mastra: {
+              om: {
+                currentTask: 'Finalize regulatory advice',
+                suggestedResponse: 'Confirm the summary with the user.',
+              },
+            },
+          },
+        },
+      });
+
+      const record = await storage.initializeObservationalMemory({
+        threadId,
+        resourceId,
+        scope: 'thread',
+        config: {},
+      });
+
+      await storage.updateBufferedObservations({
+        id: record.id,
+        chunk: {
+          observations: '- Buffered observation without continuation hints',
+          tokenCount: 100,
+          messageIds: ['msg-1'],
+          messageTokens: 45000,
+          lastObservedAt: new Date('2026-02-05T10:00:00Z'),
+          cycleId: 'cycle-1',
+        },
+      });
+
+      const result = await om.activate({ threadId, resourceId });
+
+      expect(result.activated).toBe(true);
+      const updatedThread = await storage.getThreadById({ threadId });
+      const omMetadata = getThreadOMMetadata(updatedThread?.metadata);
+      expect(omMetadata?.currentTask).toBe('Finalize regulatory advice');
+      expect(omMetadata?.suggestedResponse).toBe('Confirm the summary with the user.');
     });
 
     it('should skip activation when pending tokens are below threshold (checkThreshold)', async () => {
--- a/packages/memory/src/processors/observational-memory/observational-memory.ts
+++ b/packages/memory/src/processors/observational-memory/observational-memory.ts
@@ -3349,10 +3349,11 @@ ${formattedMessages}
       const lastActivated = activatedChunks[activatedChunks.length - 1];
       if (lastActivated) {
         const chunkThreadTitle = lastActivated.threadTitle;
+        const previousOmMetadata = getThreadOMMetadata(thread.metadata);
         const newMetadata = setThreadOMMetadata(thread.metadata, {
-          suggestedResponse: lastActivated.suggestedContinuation,
-          currentTask: lastActivated.currentTask,
-          threadTitle: chunkThreadTitle,
+          suggestedResponse: lastActivated.suggestedContinuation ?? previousOmMetadata?.suggestedResponse,
+          currentTask: lastActivated.currentTask ?? previousOmMetadata?.currentTask,
+          threadTitle: chunkThreadTitle ?? previousOmMetadata?.threadTitle,
         });
         const oldTitle = thread.title?.trim();
         const newTitle = chunkThreadTitle?.trim();


Souravrajvi0:
Closing as no longer needed.

PR #19982, opened 2026-07-22
feat(create-factory): add --skip-verify and --local-workspace to sync-template

rphansen91:
## Summary

Adds two escape-hatch flags to `mastracode/mastra-factory/scripts/sync-template.mjs` so the Software Factory template can be generated locally when a linked dependency has not yet been published to npm.

Today, the sync script hard-fails at the `npm view <pkg>@<version> version` step if any linked dep is missing from npm. That's the right default for the CI-driven publish flow, but it makes local iteration impossible whenever a brand-new package (e.g. `@mastra/factory` before its first release-train publish) or an unreleased alpha bump lands in the monorepo. There was no way to say "I know it's not on npm yet, generate the template anyway so I can smoke-test it."

## New flags

| Flag | Behavior |
| --- | --- |
| `--skip-verify` | Bypasses only the `npm view` publish check. Caret ranges are still written from the local monorepo versions, so `npm install` inside the generated template will fail if a dep is truly missing from npm. The script now logs `! <pkg>@^<ver> (WARNING: not on npm)` for each affected dep so the caller can see exactly what will break. |
| `--local-workspace` | Implies `--skip-verify`. For every dep whose local monorepo version isn't on npm, writes a `file:` spec pointing at the monorepo package instead of a caret range. The resulting template installs and runs against the local package set — useful for smoke-testing a template end-to-end without waiting for a publish. Never valid for the softwarefactory-template repo (documented in the docstring): `file:` paths only resolve inside this monorepo. |

Default behavior is unchanged: the publish check still runs, still fails loudly, and the error message now points at the new flags:

```
Error: sync-template: @mastra/factory@0.1.0-alpha.1 (local monorepo version) is
not published on npm — publish it (or wait for the release train), or sync with
--tag latest. For local iteration, use --skip-verify or --local-workspace.
```

## CI safety

- The `sync-softwarefactory-template.yml` workflow keeps calling the script with **no new flags**, so it still requires every dep to be on npm before pushing to the template repo.
- The `--local-workspace` output emits `file:` specs, which are explicitly rejected by the existing "unresolved monorepo spec" guard for every other mode. That guard now only relaxes when `--local-workspace` is set, so a caller who wanted `--skip-verify` without `file:` fallback can never accidentally ship one.
- `--skip-verify` and `--tag` are mutually exclusive (the flags have opposite intents); combining them exits with a clear error.

## Latent bug caught by the new mode

The manifest transform aliases `@mastra/memory` to `@mastra/core`'s spec verbatim (memory tracks core's version and is used as a transitive peer). Under `--local-workspace` that would make memory resolve to `packages/core` when core became a `file:` path. The alias now resolves memory against `packages/memory` directly when core is a `file:` spec, using the same publish check the rest of the script uses. Behavior in the default (caret-range) path is unchanged.

## Tests

Adds four cases to `mastracode/mastra-factory/src/sync-template.test.ts` that install a fake `npm` on `$PATH` which returns 404 for every `npm view`:

- **default mode fails when a linked dep is not on npm** — asserts the error message calls out `--skip-verify` so users don't have to grep the source.
- **--skip-verify writes caret ranges anyway** — confirms the manifest still ends up with caret ranges and the peer alias still resolves.
- **--local-workspace rewrites unpublished deps to file: paths** — walks every mastra dep in the emitted manifest, confirms each `file:` spec resolves to a real `package.json` inside the monorepo, and asserts the memory peer alias no longer duplicates core's path.
- **rejects --skip-verify combined with --tag** — guards the mutual-exclusion check.

All 35 factory tests pass. Prettier and lint clean.

## Motivation

Unblocks local end-to-end validation of unreleased template changes — most immediately for `@mastra/factory`, which was added in #19945 but hasn't been through a release train yet, so the sync workflow currently can't run. Once the next release train publishes `@mastra/factory`, the automated workflow starts working again without any further change; this PR only adds a manual escape hatch for local iteration.

## Test plan

- `pnpm --filter ./mastracode/mastra-factory exec vitest run sync-template` — 8/8 pass
- `pnpm --filter ./mastracode/mastra-factory exec vitest run` — 35/35 pass
- `node mastracode/mastra-factory/scripts/sync-template.mjs --out /tmp/x` (default mode) — fails with the improved error message
- `node mastracode/mastra-factory/scripts/sync-template.mjs --out /tmp/x --skip-verify` — succeeds, warns for missing deps
- `node mastracode/mastra-factory/scripts/sync-template.mjs --out /tmp/x --local-workspace` — succeeds, writes `file:` specs for `@mastra/factory`, `@mastra/hono`, and `mastra`

## ELI5

This adds two ways to create the Software Factory template when some local packages are not published yet: skip checking npm, or link directly to local packages. Normal behavior and CI checks stay unchanged.

## What changed

- Added `--skip-verify` to bypass npm publication failures while retaining caret version ranges.
- Added `--local-workspace`, which implies `--skip-verify` and uses `file:` paths for unpublished local dependencies.
- Prevented `--skip-verify` and `--tag` from being used together.
- Improved warnings and error messages for unpublished dependencies.
- Fixed `@mastra/memory` to resolve to its own package path in local-workspace mode.
- Added four tests covering failure behavior, both new flags, and flag conflicts.

### DIFF
--- a/.changeset/factory-sync-template-local-flags.md
+++ b/.changeset/factory-sync-template-local-flags.md
@@ -1,0 +1,5 @@
+---
+'create-factory': patch
+---
+
+Add `--skip-verify` and `--local-workspace` flags to `sync-template.mjs` for local template iteration when a linked dependency has not yet been published to npm. `--skip-verify` bypasses the publish check (caret ranges still written from local monorepo versions). `--local-workspace` implies `--skip-verify` and additionally rewrites any unpublished dep to a `file:` path anchored on the monorepo package so the generated template installs and runs without needing an npm publish first. Default and CI behavior is unchanged: the publish check still runs and still fails loudly when a dep is missing on npm, and error messages now point users to the new flags.
--- a/mastracode/mastra-factory/scripts/sync-template.mjs
+++ b/mastracode/mastra-factory/scripts/sync-template.mjs
@@ -22,6 +22,27 @@
  *
  * Usage:
  *   node scripts/sync-template.mjs [--out <dir>] [--tag latest]
+ *                                  [--skip-verify] [--local-workspace]
+ *
+ * Flags:
+ *   --out <dir>          Where to write the generated template
+ *                        (default: `template-out/` next to this package).
+ *   --tag <name>         Pin caret ranges to that dist-tag on npm instead of
+ *                        the local monorepo versions.
+ *   --skip-verify        Skip the `npm view` publish check. Caret ranges are
+ *                        still written from the local monorepo versions, so
+ *                        `npm install` inside the template will fail if any
+ *                        dep is not yet on npm. Intended for local iteration
+ *                        on the sync script itself, not for publishing.
+ *   --local-workspace    Implies `--skip-verify`. For any dep whose local
+ *                        monorepo version is not yet on npm, write a `file:`
+ *                        spec pointing at the monorepo package instead of a
+ *                        caret range. The resulting template installs and
+ *                        runs against the local package set — useful for
+ *                        smoke-testing an unreleased template end-to-end.
+ *                        NEVER use this for a template that will be pushed
+ *                        to the softwarefactory-template repo: `file:` paths
+ *                        only resolve inside this monorepo.
  *
  * Output defaults to `template-out/` next to this package (gitignored).
  * Publish flow: automated — the sync-softwarefactory-template workflow runs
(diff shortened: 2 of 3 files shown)


rphansen91:
Closing — wrong approach. The real problem isn't 'let me force-sync with an unpublished alpha version', it's that sync-template pins the current monorepo alpha version at all. Every other create-* template pins `latest` (or the requested `--tag`) from npm; this one should do the same. Will open a new PR that makes `--tag latest` the default so the sync workflow no longer cares whether the monorepo is mid-alpha.

PR #19990, opened 2026-07-22
fix(mastracode): keep sidebar visible during factory creation

damien-schneider:
Keep the standard MastraCode sidebar visible on `/factories/create`.

The create route now restores the source factory context and reuses the normal app shell. Sidebar thread links use that active factory directly, so they keep working even though the create URL has no `:factoryId`.

Verified with the focused routing and session recovery tests, `pnpm check:ui`, and desktop/mobile browser QA.

## ELI5

The Create Factory page now keeps the normal sidebar and remembers which factory you came from. This lets users navigate between factory threads and the creation page without losing their place.

## Summary

- Added `CreateFactoryLayout` to restore the active factory context on `/factories/create`.
- Reused the standard application shell with sidebar and chat header.
- Updated sidebar thread navigation to use the active factory directly.
- Preserved the source location when reopening factory creation.
- Expanded routing, session recovery, and navigation test coverage.


mfrachet:
my next PR will change that, let's keep it here for now

abhiaiyer91 (approved):

PR #19992, opened 2026-07-22
fix(mastracode): make repository rows clickable

damien-schneider:
Turns each available repository row into the link action and removes the nested Link button. The full row now has native button semantics, preserves the disabled and loading states, and works with keyboard input.

Validated with the focused GitHub MSW journey, MastraCode web UI typecheck, Prettier, and React Doctor.

## ELI5

Each repository row is now one big clickable button, making it easier to select and link repositories. It also clearly shows when linking is in progress and prevents other rows from being clicked during that time.

## Changes

- Replaced the nested “Link” button with full-row button semantics.
- Preserved disabled, loading, and keyboard interaction states.
- Added an inline “Linking…” indicator for the active repository.
- Validated with the focused GitHub MSW journey, web UI typecheck, Prettier, and React Doctor.


mastracode/web/src/web/ui/domains/workspaces/components/ConnectRepositoriesPanel.tsx
```diff
@@ -149,15 +152,8 @@ export function ConnectRepositoriesPanel({ factory }: { factory: ServerFactory }
...
-                  <Button
-                    variant="ghost"
-                    size="sm"
-                    disabled={busyRepoId !== null}
-                    onClick={() => linkRepository.mutate({ factoryProjectId, repo })}
-                  >
-                    {busyRepoId === repo.id ? 'Linking…' : 'Link'}
-                  </Button>
-                </div>
+                  {busyRepoId === repo.id && <span className="text-ui-sm text-icon3">Linking…</span>}
```
mfrachet, line 155: should we do disabled + spinner?

mfrachet (approved):

PR #20003, opened 2026-07-22
feat(playground): configure dataset item timeouts

YujohnNattrass:
PR 4 of 4 for [MASTRA-4509](https://linear.app/kepler-crm/issue/MASTRA-4509/oss-experiments-per-item-timeout-overrides), stacked on #20138.

Adds Studio controls for setting and editing per-item timeout overrides. Persisted values are displayed in the dataset item detail panel, and the shared form on both the detail panel and item-history page accepts only positive integer milliseconds up to `1_800_000` (30 minutes).

After this change, users can configure an override while creating an item, edit it from either supported surface, and confirm the persisted value after reloading. Because the update API does not currently support clearing a stored override, the form prevents an existing value from being replaced with a blank value.

The core, storage, and API behavior is implemented in #20136, #20137, and #20138.

Verified with 25 MSW tests, typecheck, lint, the CLI build, and the 8-test dataset Playwright suite.

### DIFF
--- a/packages/playground/src/domains/datasets/components/dataset-detail/dataset-item-form.tsx
+++ b/packages/playground/src/domains/datasets/components/dataset-detail/dataset-item-form.tsx
@@ -1,17 +1,38 @@
 'use client';
 
+import type { DatasetItem, DatasetItemToolMock } from '@mastra/client-js';
 import { Button } from '@mastra/playground-ui/components/Button';
 import { CodeEditor } from '@mastra/playground-ui/components/CodeEditor';
+import { Input } from '@mastra/playground-ui/components/Input';
 import { Label } from '@mastra/playground-ui/components/Label';
+import { toast } from '@mastra/playground-ui/utils/toast';
 import { Pencil } from 'lucide-react';
+import { useState } from 'react';
+import type { ReactNode } from 'react';
+import { useDatasetMutations } from '../../hooks/use-dataset-mutations';
 
-/** Schema validation error from API */
-export interface SchemaValidationError {
+interface SchemaValidationError {
   field: 'input' | 'groundTruth' | 'toolMocks';
   errors: Array<{ path: string; message: string }>;
 }
 
-/** Displays field-level validation errors */
+function parseValidationError(error: unknown): SchemaValidationError | null {
+  if (!(error instanceof Error)) return null;
+
+  const match = error.message.match(/- ({.*})$/);
+  if (!match) return null;
+
+  try {
+    const parsed = JSON.parse(match[1]);
+    if (parsed.field && Array.isArray(parsed.errors)) {
+      return { field: parsed.field, errors: parsed.errors };
+    }
+  } catch {
+    // Not valid JSON
+  }
+  return null;
+}
+
 function ValidationErrors({ field, errors }: { field: string; errors: Array<{ path: string; message: string }> }) {
   if (!errors.length) return null;
 
@@ -30,9 +51,6 @@ function ValidationErrors({ field, errors }: { field: string; errors: Array<{ pa
   );
 }
 
-/**
- * Editable form view for updating dataset item
- */
 export interface EditModeContentProps {
   inputValue: string;
   setInputValue: (value: string) => void;
@@ -50,6 +68,7 @@ export interface EditModeContentProps {
   onSave: () => void;
   onCancel: () => void;
   isSaving: boolean;
+  children?: ReactNode;
 }
 
 export function EditModeContent({
@@ -69,6 +88,7 @@ export function EditModeContent({
   onSave,
   onCancel,
   isSaving,
+  children,
 }: EditModeContentProps) {
   return (
     <>
@@ -126,6 +146,8 @@ export function EditModeContent({
           )}
         </div>
 
+        {children}
+
         <div className="space-y-2">
           <Label>Request Context (JSON, optional)</Label>
           <CodeEditor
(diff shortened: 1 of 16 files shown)


YujohnNattrass:

- #20136
    - #20137
        - #20138
            - #20003 ◀

packages/playground/src/domains/datasets/components/dataset-detail/dataset-item-form.tsx
```diff
@@ -158,3 +180,187 @@ export function EditModeContent({
...
+      setToolMocksValue={handleToolMocksValueChange}
+      requestContextValue={requestContextValue}
+      setRequestContextValue={setRequestContextValue}
+      validationErrors={validationErrors}
+      onSave={handleSave}
+      onCancel={onCancel}
+      isSaving={updateItem.isPending}
+    >
+      <div className="space-y-2">
+        <Label htmlFor="edit-item-timeout">Item timeout (ms, optional)</Label>
```
NikAiyer, line 357: **Blocking:** Could we expose the item timeout on the other supported edit surface as well?

The dedicated dataset-item page is still reachable through “Go to item versions history” and version-comparison views, but its edit form does not render this timeout field. That means the same item can edit its timeout from the list-side panel but not from its dedicated page. Please keep these item-edit surfaces consistent, ideally by sharing the timeout control and validation rather than implementing it independently.

packages/playground/src/domains/datasets/components/items/dataset-items.tsx
```diff
@@ -231,6 +231,7 @@ export function DatasetItems({
 
   const detailPanelSlot = featuredItem ? (
     <DatasetItemPanel
+      key={featuredItem.id}
```
NikAiyer, line 234: **Non-blocking:** `DatasetItemEditForm` initializes all item-derived drafts only on mount, so switching-item correctness now implicitly depends on this ancestor key. Could we add coverage proving that selecting another item while editing discards the previous item’s drafts, while a same-item refetch preserves active drafts? Alternatively, putting `key={item.id}` directly on `DatasetItemEditForm` would keep the lifecycle dependency adjacent to the component that requires it.

NikAiyer (changes requested):
The timeout flow is well covered through the client/React Query path and by the reload E2E. I found one product-consistency blocker: timeout editing is only available in the list-side panel even though the dedicated item page remains an active, linked edit surface. I also left a non-blocking lifecycle/test-coverage suggestion. The current CI failures appear unrelated to this PR-specific diff, but should still be rerun before merge. This user-facing change also needs the repository-required changeset unless the stack is explicitly exempt.

packages/playground/src/domains/datasets/components/dataset-detail/dataset-item-form.tsx
YujohnNattrass: Addressed in `e161a7a17e`. The dedicated item-history page now renders the same `DatasetItemEditForm` as the list-side panel, and item-history mapping carries the persisted timeout into that form. The dataset Playwright journey now updates and reloads the timeout through both edit surfaces.

packages/playground/src/domains/datasets/components/items/dataset-items.tsx
YujohnNattrass: Addressed in `e161a7a17e`. The identity key now sits directly on `DatasetItemEditForm`. Added focused coverage proving a different item resets drafts while a same-item refetch preserves the active draft.

.changeset/ready-nights-wonder.md
```diff
@@ -0,0 +1,5 @@
+---
+'@mastra/pg': patch
+---
+
+Added persistence for dataset item `timeout` values, including batch inserts, updates, and historical dataset versions.
```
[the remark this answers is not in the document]
YujohnNattrass: Addressed in the lower stores layer, #20137, by `7ddd16c44e`. The existing PostgreSQL changeset now includes a concise `dataset.addItem({ input, timeout })` example and explicitly states that PostgreSQL preserves the override through updates and historical versions.

packages/core/src/datasets/experiment/index.ts
```diff
@@ -360,38 +391,57 @@ export async function runExperiment(mastra: Mastra, config: ExperimentConfig): P
...
+            execResult = executionFailure(itemSignal.reason);
+            break;
           }
 
+          retryCount++;
           execResult = await execFn(item, itemSignal);
+          if (signal?.aborted) {
+            throw signal.reason ?? new DOMException('Aborted', 'AbortError');
+          }
         }
```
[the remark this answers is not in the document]
YujohnNattrass: Addressed in the core layer, #20136, by `6511310a7f`. Abort classification now uses a dedicated `ExecutionResult.aborted` flag instead of message substring matching, and the regression test proves that a transient `Network request aborted by upstream` failure is retried successfully.

NikAiyer (approved):
YJ’s follow-up addresses both findings from my initial review:

- The dedicated item-history page now uses the shared dataset item edit form, including timeout editing.
- The form’s item-identity reset boundary is now colocated with the state owner, with focused coverage for cross-item navigation and same-item refetches.

The focused MSW suites pass locally (25/25), and the stack-level changeset now documents timeout usage and persistence. The remaining local typecheck issue is an unrelated missing build artifact for `@mastra/ai-sdk/ui`.

abhiaiyer91 (approved):
**Verdict: approve**

### Findings

**Correctness**
- The shared `DatasetItemEditForm` is the right consolidation. Both edit surfaces (list detail panel, item versions page) now mount it conditionally behind `isEditing` and keyed on the item id, so each edit session starts from freshly derived item state rather than a stale draft. This removes the hand-rolled `formDefaults` / `versionKey` / `prevVersionKey` reset logic the item page previously carried.
- The item page's `editableItem` is derived from `latestVersion` while the read-only `displayItem` follows the selected version. That is the correct split: browsing an older version does not let you save an edit against it.
- Blocking the clear of a persisted timeout is a genuine API constraint, not a UI shortcut. `packages/server/src/server/schemas/datasets.ts` types the update field as `experimentTimeoutSchema.optional()` with `min(1)` and no `null` branch, and `UpdateDatasetItemParams.timeout` is `number | undefined`, so there is no wire representation for "unset". Failing fast with a clear message is better than sending a request the API silently ignores.
- Client-side validation (finite, integer, `> 0`, `<= 1_800_000`) matches the server schema bounds exactly, so the UI rejects the same values the API would.

**Tests**
- Behaviour is covered, not just rendering. The panel tests assert the mutation payload on the wire (`PATCH .../items/item-1` with `timeout: 1_800_000`), assert that out-of-range values `['0','-1','1.5','1800001']` and a blanked persisted timeout produce a toast *before* any request is made, and cover the draft lifecycle (different item resets drafts, same-item refetch preserves the active draft, cancel remounts the original value).
- The add-item tests distinguish "timeout provided" from "timeout omitted" at the payload level — the empty case omits the key entirely rather than sending `null`/`undefined`, which matches the optional schema.
- The Playwright journey exercises the property that motivated the last round of changes: the same override is editable from both surfaces and survives a reload.

**Scope**
- Confined to Studio. The diff is 15 files, all under `packages/playground`, with no changes to install scripts, lockfiles, test configuration, or CI workflows.

**Pattern consistency**
- `MAX_EXPERIMENT_ITEM_TIMEOUT_MS` is now defined in three places (core validation, server schemas, Studio constants) with the same value. Studio cannot reach into core validation internals, and the server already duplicates it, so this follows the existing precedent rather than introducing a new one. Non-blocking, but the value is a release-coupled invariant worth consolidating if a shared constants surface ever exists.
- Non-blocking, pre-existing: `useDatasetItemVersions` declares `requestContext` on `DatasetItemVersion` but never maps it from the history response, so the item versions page opens the edit form with an empty Request Context field. This PR added `timeout` to that same mapping and did not introduce the gap — saving sends `requestContext: undefined`, so nothing is cleared — but the omission is now shared by both edit surfaces and is worth a one-line fix with coverage.

### Verification

- `pnpm install` — succeeded.
- `pnpm turbo build --filter ./packages/playground-ui` — succeeded; needed because the dataset tests resolve `@mastra/playground-ui` subpath exports from `dist`.
- `pnpm turbo build --filter ./client-sdks/ai-sdk` — succeeded; needed to resolve `@mastra/ai-sdk/ui` for typecheck.
- `pnpm --filter ./packages/playground typecheck` (`tsc --noEmit`) — passed.
- `pnpm --filter ./packages/playground exec vitest run src/domains/datasets` — 14 files, 81 tests, all passing.
- `pnpm --filter ./packages/playground lint` — 0 errors, 37 warnings, none in the files this PR touches (all pre-existing `react-refresh` / `exhaustive-deps` warnings elsewhere in the package).
- CI on head `ed8c936`: Changed Test Gate, Major Version Check, Lint Docs, Socket Security, Superagent Security Scan, and the Vercel preview builds all green.
- Verified working tree is clean at the reviewed head commit before running any of the above.

### Existing review disposition

- **CodeRabbit — `setTimeout` state setter shadows `window.setTimeout` in `add-item-dialog.tsx`: addressed.** The dialog now uses `timeoutValue` / `setTimeoutValue`, matching the edit form.
- **CodeRabbit — changesets lacked a public API usage example: addressed.** The PostgreSQL changeset in the stack now carries a `dataset.addItem({ input, timeout })` example.
- **Reviewer — changesets missing API usage examples: addressed** in the API PR of the stack.
- **Reviewer — capability-gated test utils asserted a MySQL-specific error id: addressed**; the utils now use generic rejection assertions while keeping the version checks.
- **Reviewer — executor retry loop matched abort state by error message substring: addressed**; the loop now reads a dedicated `aborted` flag on the execution result, with a regression test proving transient network errors are still retried.
- **Reviewer — timeout editing was missing from the item versions page: addressed** in `e161a7a`; that page now renders the same shared edit form and the Playwright journey covers both surfaces.
- **Reviewer — no coverage for the form's draft lifecycle across item switches: addressed** in `e161a7a`; the identity key moved onto the form itself and focused tests now cover reset-on-different-item and preserve-on-same-item-refetch.
- No substantive prior finding remains confirmed and unaddressed. No review bot is still working on the head commit.

### Adversarial check

The strongest case for requesting changes was the missing Studio changeset: `@internal/playground` is explicitly un-ignored in `.changeset/config.json`, is version-locked with `mastra` / `create-mastra`, and prior Studio-only work shipped its own entry — so the user-facing half of this feature would have gone unmentioned in the changelog. It fails as a blocker because it does not affect runtime behaviour, the feature's public API is already documented by the stack's core/server/client-js/storage changesets, and the fix is a single markdown file. Rather than bounce an otherwise fully verified PR, I opened #20525 against this branch to add it.

### Assumptions

- Reviewed as the fourth PR of a stack: findings already fixed in the earlier API PRs of the same stack are treated as addressed for this PR, since those commits are contained in this branch's base.
- The absence of a changeset in this PR's own diff was judged against repository convention for Studio-only changes rather than against the changeset bot comment alone.
- Playwright E2E was not executed in the sandbox; that journey is judged from the spec source and the green Changed Test Gate on CI. Vitest and typecheck were run locally.

### Open questions

- Should the API gain a way to remove a per-item timeout override (for example accepting `null` on update)? Studio currently has to refuse the operation with an error message, which is the correct behaviour today but leaves users unable to revert an item to the experiment-level fallback without recreating it.

PR #20005, opened 2026-07-22
feat(factory): add shared supervisor workspace

TylerBarnes:
## Stack

Two-part stack — merge in order:

1. #19970 — supervisor approval foundations (base: `main`)
2. **#20005 — shared supervisor workspace** ← you are here (base: `feat/factory-supervisor-approvals`)

This PR builds on #19970 and **must not merge before it**. Its base is #19970's branch, so the diff shown here is only the supervisor half; GitHub will retarget this to `main` once #19970 merges, at which point it needs a rebase.

Reviewing this in isolation is fine — the approval primitives it consumes (durable approvals, `pending_approval`, attributed messages) all land in #19970.

## Summary

- add one tenant-scoped Factory supervisor session per project with deterministic `${factoryProjectId}-supervisor` thread identity
- add bounded Factory state, work-item inspection, worker signaling, and transition approval tools with durable lifecycle notifications
- add the `/factory/supervisor` workspace, shared attributed chat transcript, live state summary, and approve/reject controls
- isolate repository-free supervisor runs from project workspaces and prevent the generic chat shell from claiming the supervisor resource before canonical session setup

## Test plan

- `pnpm --filter @mastra/factory test -- --reporter=dot --bail 1`
- `pnpm --dir mastracode/sdk test -- --reporter=dot --bail 1`
- `pnpm --dir mastracode/web web:ui:test -- --reporter=dot --bail 1`
- `pnpm --dir mastracode/web web:build`
- live authenticated browser proof with a configured real model: query Factory state, verify attributed transcript, create an agent-only pending approval, approve it, and observe the automatic intake-to-execute move

### DIFF
--- a/.changeset/hungry-flies-heal.md
+++ b/.changeset/hungry-flies-heal.md
@@ -1,0 +1,11 @@
+---
+'create-factory': patch
+'@mastra/client-js': patch
+'@mastra/factory': patch
+'@mastra/server': patch
+'@mastra/code-sdk': patch
+'@mastra/core': patch
+'mastra': patch
+---
+
+Added Factory supervisor identity fields to Mastra Code session state.
--- a/.changeset/jolly-bats-push.md
+++ b/.changeset/jolly-bats-push.md
@@ -1,0 +1,11 @@
+---
+'@mastra/factory': minor
+'create-factory': patch
+'@mastra/client-js': patch
+'@mastra/server': patch
+'@mastra/code-sdk': patch
+'@mastra/core': patch
+'mastra': patch
+---
+
+Added a shared Factory supervisor session with tenant-scoped state, work-item signaling, approval tools, and durable lifecycle notifications.
--- a/.changeset/olive-pugs-shake.md
+++ b/.changeset/olive-pugs-shake.md
@@ -1,0 +1,6 @@
+---
+'@mastra/factory': minor
+'@mastra/code-sdk': patch
+---
+
+Wake the Factory supervisor after a server restart when work is still open. A restart kills in-flight runs without emitting a run-completion event, so idle worker observation cannot see them and the affected work items stall silently. Factory now enqueues one durable supervisor notification per tenant that still has a work item outside the `done` and `canceled` stages, asking the supervisor to find stalled work and resolve it one item at a time. Repeated restarts inside a 15-minute window collapse into a single wake, and delivery is exactly-once across replicas. Opt out with `supervisor.checkInOnBoot: false` in the Factory rules.
--- a/.changeset/quiet-moons-repeat.md
+++ b/.changeset/quiet-moons-repeat.md
@@ -1,0 +1,6 @@
+---
+'@mastra/factory': patch
+'@mastra/code-sdk': patch
+---
+
+Give the Factory supervisor its tools on server-initiated runs. The supervisor resolved its tenant identity from the authenticated user on the request context, but runs the server starts itself — boot check-ins, idle worker observations, and approval notifications — build a fresh request context that carries no user. Those runs therefore woke the supervisor with none of its Factory tools registered, leaving it unable to query state or act on the very event that woke it. Tenant identity now comes from the supervisor session's own state, which is durable and pinned to the session's canonical resource and thread. A request user, when present, must still belong to the session's tenant and is used to attribute the resulting audit entries.
--- a/mastracode/factory/README.md
+++ b/mastracode/factory/README.md
@@ -41,7 +41,7 @@ Resolver and tenant identity always come from server authentication, not request
 
 ## Supervisor state and lifecycle signals
 
-Opening the canonical supervisor session emits a full `factory-state` snapshot. The snapshot is bounded to board and stage counts plus at most 50 pending approvals with work-item title, role, requested stage, revision, reason, summary, and age. It never includes credentials, storage handles, repository paths, or raw integration payloads. A stable cache key suppresses unchanged snapshots.
+Opening the canonical supervisor session emits a full `factory-state` snapshot. The snapshot is bounded to board and stage counts, at most 50 pending approvals with work-item title, role, requested stage, revision, reason, summary, and age, and live worker activity per active run binding — `running` (a run is in flight), `idle` (the bound session is live between runs), or `offline` (no in-process session owns the binding). It never includes credentials, storage handles, repository paths, or raw integration payloads. A stable cache key suppresses unchanged snapshots.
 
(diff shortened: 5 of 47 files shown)


abhiaiyer91 (approved):
**Verdict: Approve.** The supervisor workspace is coherent, fails closed on tenant identity at every seam I probed, and its behavior is covered by tests that assert the security properties rather than just the happy path. The one outstanding scanner finding is a scalability concern rather than a confidentiality one, and it matches a pattern this file already established on the base branch.

## Existing review signal — disposition

**superagent-security P2, "Cross-tenant data read in boot check-in work-item scan"** (`mastracode/factory/src/storage/domains/work-items/base.ts:1970`, fingerprint `b2d7fc79…`) — **refuted as a confidentiality issue, confirmed as a scalability concern, non-blocking.**

The scanner is factually right that `listTenantsWithOpenWorkItems` issues `findMany<WorkItemDbRow>('work_items', {})` and filters `done`/`canceled` in memory. Three things move it off the blocking list:

1. **No tenant boundary is crossed in output.** The method has exactly one caller (`FactoryBootCheckIn.run`, `supervisor/boot-check-in.ts:53`), which runs at boot *before any request establishes a tenant* — enumerating tenants is the function's purpose. Each tenant's rows only ever produce that tenant's own notification and audit record. Nothing from tenant A reaches tenant B, and the method is unreachable from any request-scoped route.
2. **It is not a new pattern.** `#claimLeases` on the merge base (`0072afbf0b`, base.ts:994) already does `findMany<GovernanceDbRow>(table, {}, …)` — a full unfiltered read of the governance tables — and it runs on **every dispatcher tick**, not once per boot. Requesting a bounded query here while a strictly hotter pre-existing path stays unbounded would be inconsistent.
3. **The tradeoff is documented at the call site** (base.ts:1961-1966), and the real fix — an indexed `has_open_work` column or a materialized view — is a schema change to the storage seam, not a local edit.

Recommend tracking the bounded-query work as a follow-up covering both this scan and `#claimLeases` together, rather than blocking this PR.

## Adversarial check

Strongest case for requesting changes: the boot scan is a new unbounded full-table read on a table that grows without limit, so a large multi-tenant deployment could OOM on restart — exactly when it is least able to absorb it. It fails because the same process already reads three governance tables unbounded on every tick; this PR does not move the ceiling, and fixing it properly requires the schema change described above.

## Verification

All run on head `4d7ac696d6` (matches the PR head OID), clean checkout, credentials stripped.

| Command | Result |
| --- | --- |
| `pnpm turbo build --filter ./mastracode/factory --filter ./mastracode/web` | 30/30 tasks pass |
| `pnpm --filter @mastra/factory test` | **1092 pass / 78 files**, 0 fail |
| `pnpm --filter @mastra/factory check` (`tsc --noEmit`) | clean |
| `pnpm --filter @mastra/code-sdk test` | **1052 pass**, 4 fail → see note |
| `pnpm --filter @mastra/code-sdk check` | clean |
| `mastracode/web` `vitest run` (unit) | **77 pass / 13 files** |
| `mastracode/web` `vitest run --config e2e/web-ui/vitest.config.ts` | **294 pass / 58 files** |
| `mastracode/web` `pnpm run check` (`tsc --noEmit` ×2, incl. UI project) | clean |

The 4 SDK failures in `src/plugins/__tests__/install.test.ts` are **not a regression** — they are `ERR_MODULE_NOT_FOUND` on `mastracode/plugin` because the `mastracode` package had no `dist/`. After `pnpm turbo build --filter mastracode`, the file passes 13/13. Same root cause for the initial web-ui run: 50 files failed to resolve `@mastra/playground-ui/*` until `pnpm run prebuild` produced the dist artifacts, after which all 58 pass. Neither is touched by this diff.

The 5 new supervisor UI tests pass individually and assert real behavior — attributed send through the deterministic session, approval resolution with cache invalidation, the stale-revision path reporting without moving state optimistically, and an unknown project bouncing to the landing route.

## What I checked beyond the tests

Tenant identity fails closed at each layer independently, which is the property that matters most here:

- `SupervisorRoutes.#resolveProject` requires an authenticated user, requires `orgId`, validates the project UUID shape, and resolves the project *scoped to the caller's org* — so a cross-tenant project id 404s before the service is touched. `supervisor.test.ts` asserts exactly this (404 cross-tenant, 401 anonymous, and `ensureSession` never called).
- `#ensureCanonicalSession` refuses to adopt a live resource whose state does not match the requesting org, project, and supervisor flag, and `#configureSession` independently re-validates the *thread's* `resourceId` and metadata before writing anything. Two independent checks, both throwing rather than coercing.
- `resolveSupervisorContext` derives tenant from durable session state rather than the request user, which is what lets server-initiated runs (boot check-in, idle observation) carry tools at all — and when a request user *is* present it must belong to the session's org. The `quiet-moons-repeat` changeset explains this clearly.
- Supervisor sessions get a repo-less, read-only isolated workspace, and `isolatedWorkerRequestContext()` keeps the supervisor's context from leaking into worker reopens. `tools.test.ts` has a dedicated case that signaling a worker does not de-canonicalize the supervisor.

Rule validation, defaults, and the `checkInOnBoot` opt-out are all wired consistently, including `assertExactKeys`. Four changesets, correctly scoped, `@mastra/factory` minor.

## Non-blocking observations

These are follow-up material, not requested changes:

1. **Duplicated approval polling.** `SupervisorWorkspace` polls `/supervisor/state` every 5s (which already carries `pendingApprovals`) *and* `/approvals?status=pending` every 5s, in two different shapes (`FactorySupervisorApprovalState` vs `FactorySupervisorApproval`). The sidebar renders the second. Worth collapsing to one source.
2. **Error text forwarded to the client.** The session route's 409 returns `error.message` verbatim. Today those are curated strings, so it is fine; it becomes a leak vector if a throw site downstream changes.
3. **Asymmetric error handling.** The session route wraps `ensureSession` in try/catch; `/supervisor/state` has none, so a storage error surfaces as an unhandled 500 rather than the bounded shape the client's `requestJson` expects.
4. **Deterministic thread id is squattable in principle.** `{factoryProjectId}-supervisor` is guessable, but `#configureSession` throws instead of adopting a mismatched thread, so the residual risk is denial-of-service shaped, not disclosure shaped — and it requires knowing a target org's project UUID. Correct call as built; noting it so it is on the record.

## Assumptions

- Reviewed the head branch in isolation. This PR targets `feat/factory-supervisor-approvals` (#19970), which is still **open** against `main`, so the approval primitives it builds on are not yet merged. I treated #19970 as sound per the author's stated stacking plan and reviewed only this diff's 47 files.
- Treated the failing Superagent check as *reported and dispositioned* rather than *pending* — the bot posted its finding on Jul 26 and the check is red because of that finding, which I have addressed above.

## Open questions for a human

- Retargeting to `main` after #19970 merges will change the effective diff. Worth a re-check at that point, though the supervisor half should be unaffected.
- Who picks up the bounded-query follow-up, and does it cover `#claimLeases` in the same pass?

PR #20015, opened 2026-07-22
fix(server-adapters): enforce body size limit on DELETE requests

### LINKED ISSUE #20013: [BUG] Body size limit is not enforced on DELETE requests in Express/Fastify/Hono server adapters
## Summary
Found this while poking at the body-limit middleware. `bodyLimit()` (and the per-adapter equivalents) only gets wired up for a fixed set of HTTP methods, and `DELETE` isn't one of them - not in Express, not in Fastify, not in Hono. So if you've got a route that takes a `DELETE` with a body, it reads that body straight into memory with zero size checking, no matter what you've set `bodySizeLimit` to.

## Severity
Low for now - nothing shipped actually hits this - but it's baked into the adapter code itself, not something a deployer could misconfigure their way into. Anyone who adds a `DELETE`-with-body route (bulk delete by ID list is the obvious one) inherits the hole automatically.

## Root Cause
Each adapter checks the request method against a hardcoded list before deciding whether to attach the size-limiting wrapper - something like `['POST', 'PUT', 'PATCH'].includes(method)`. `DELETE` never made it onto that list in any of the three adapters. Nothing stops a `DELETE` from having a body (HTTP allows it, and it's a pretty normal REST pattern), so the handler just reads it through unchecked.

## Why this matters
- **One request can take the whole process down.** A big enough body on a single `DELETE` forces Node to buffer it all into the heap. Hit the memory ceiling and the process dies — taking every other in-flight request with it, not just this one route.
- **It's a cheaper DoS than the ones people actually defend against.** `POST`/`PUT`/`PATCH` are guarded. `DELETE` isn't. Less bandwidth needed to do more damage, since it's the one nobody thought to lock.
- **Whoever configured `bodySizeLimit` thinks they're covered.** They're not, and there's no error or warning telling them so — it just quietly doesn't apply here.

## Who this actually affects
- Anyone registering a custom `DELETE`-with-body route via `server.apiRoutes[]`.
- Anyone embedding the Express/Fastify/Hono adapters directly - the gap lives in shared code, not app-specific config.
- Any future built-in route that ends up being `DELETE`-with-body (a bulk-delete endpoint, most likely) - it'd inherit this silently.

## Suggested fix
Add `'DELETE'` to the method allowlist in all three adapters' body-limit wiring. Same check, one more string in each file. Nothing else needs to change.

## One more note
No route shipped today defines a `bodySchema` on `DELETE`, so this isn't live yet. But it's exactly the kind of thing that turns into a real incident the day someone adds one convenience endpoint without knowing this blind spot exists.

## How to reproduce it
1. Register a route through `server.apiRoutes[]` with `method: 'DELETE'`, and have the handler actually read the body (`c.req.json()` in Hono, for example).
2. Set `bodySizeLimit` to something small — 1MB is fine for testing.
3. Send that route a `DELETE` with a body well over the limit — 200MB, say.
4. It goes through. The limit does nothing for this method.

https://github.com/shramanb113/mastra-issues

### Expected Behavior

The bug shows up as soon as the body crosses 100 bytes:

- **Expected:** both routes should reject a body over 100 bytes the same way,
  regardless of method.
- **Actual:** `POST` visibly changes behavior once the body exceeds 100 bytes
  (the body-limit middleware kicks in and the response degrades to an empty
  `200` instead of the normal `400` validation error — see note below).
  `DELETE` shows no change in behavior at all, at any size tested: it always
  reaches the handler and returns the same `400` validation error, meaning
  the configured 100 byte limit is never applied to it.

### Environment Information

```block
System:
  OS: Windows 11 10.0.26200
  CPU: (12) x64 12th Gen Intel(R) Core(TM) i5-12450H
  Memory: 2.32 GB / 15.68 GB
Binaries:
  Node: 22.13.1 - C:\Program Files\nodejs\node.EXE
  npm: 11.17.0 - C:\Program Files\nodejs\npm.CMD
  pnpm: 10.18.0 - C:\Users\SHRAMAN\AppData\Roaming\npm\pnpm.CMD
  bun: 1.2.23 - C:\Users\SHRAMAN\AppData\Roaming\npm\bun.CMD
Browsers:
  Chrome: 150.0.7871.128
  Edge: Chromium (150.0.4078.83)
  Internet Explorer: 11.0.26100.8115
npmPackages:
  @mastra/core: ^1.51.0 => 1.51.0
  @mastra/libsql: ^1.16.0 => 1.16.0
  @mastra/memory: ^1.23.0 => 1.23.0
  @types/node: ^26.1.1 => 26.1.1
  mastra: ^1.19.0 => 1.19.0
  typescript: ^7.0.2 => 7.0.2
  zod: ^4.4.3 => 4.4.3
npmGlobalPackages:
  @colbymchenry/codegraph: 0.9.7
  @nestjs/cli: 11.0.14
  @nubjs/nub: 0.1.12
  bun: 1.2.23
  npm: 11.17.0
  pnpm: 10.18.0
```

### Verification

shramanb113:
## Description

`server.bodySizeLimit` (and the per-route `maxBodySize` override) was only being enforced for `POST`, `PUT`, and `PATCH` requests in the Hono, Express, and Fastify adapters' `registerRoute()`. DELETE was left out of that check.

The catch is that `getParams()` in all three adapters already reads and JSON-parses the request body for DELETE requests too (HTTP allows a body on DELETE, and it's a pretty normal shape for something like a bulk-delete-by-ID-list endpoint). So the body still gets fully buffered into memory on a DELETE request, it just never gets checked against the configured limit before that happens. A POST with an oversized body gets rejected with a 413 like you'd expect; the exact same body sent as a DELETE goes straight through.

In terms of impact, this is a memory-exhaustion issue more than anything else. Because the body is read into memory whether or not the route's handler even uses it, someone can send a single DELETE request with an arbitrarily large body (a few hundred MB, a few GB, whatever) to any DELETE endpoint and the server will try to hold the whole thing in the Node process heap before anything checks its size. Since this runs on the same process handling every other request, a large enough body (or a handful of concurrent ones) can push memory high enough to slow the whole server down or crash the process under GC pressure - which takes out every other in-flight request, not just the one that sent the oversized body. It's a fairly low-effort way to cause an outage on a server that already thought it had a body size limit configured.

The fix itself is a one-line change per adapter: add `'DELETE'` to the `shouldApplyBodyLimit` method allowlist in `server-adapters/hono/src/index.ts`, `server-adapters/express/src/index.ts`, and `server-adapters/fastify/src/index.ts`, so the configured limit is enforced consistently across POST, PUT, PATCH, and DELETE.

Added a regression test in `server-adapters/hono/src/__tests__/hono-adapter.test.ts` that registers the same route with the same `bodyLimitOptions` for both POST and DELETE with an identical oversized payload - the POST case already passed, the DELETE case failed before this fix and passes now.

I also put together a minimal reproduction project (a fresh `create-mastra` app hitting the built-in `/api/schedules` and `/api/memory/threads/:id` routes) that shows the same body-size threshold change POST's response but never changes DELETE's - linked in the issue.

## Related issue(s)

Fixes #20013

## Type of change

- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [ ] Test update

## Checklist

## ELI5
Some web requests use a `DELETE` method but still include data in the request body. Before this change, the server adapters could accept huge `DELETE` bodies without checking their size—wasting memory. Now, `DELETE` bodies are size-limited the same way as other body-carrying methods.

## Summary
- Updated the Hono, Express, and Fastify server adapters so body-size limit enforcement also applies to `DELETE` requests (previously it only covered `POST`, `PUT`, and `PATCH`).
- Ensured oversized `DELETE` bodies are rejected with **HTTP 413** when they exceed the global `server.bodySizeLimit` and any per-route `maxBodySize`, preventing unbounded buffering/JSON parsing for `DELETE`-with-body routes.
- Expanded regression coverage by adding a shared test suite (`server-adapters/_test-utils`) and running it across the adapters for oversized `POST` and `DELETE` requests.
- Included additional adapter fixes noted during implementation:
  - Corrected Fastify body-limit configuration to use the supported top-level route option.
  - Adjusted Hono’s custom `onError` handling to return a proper HTTP **413** response.
- Added changesets releasing patch updates for `@mastra/express`, `@mastra/fastify`, and `@mastra/hono` documenting the security fix, plus a separate patch changeset addressing the CLI/Studio UI prebuild versioning issue.


NikAiyer (changes requested):
Thanks for fixing this. The production change correctly aligns DELETE with the other body-bearing methods, and the Hono regression test demonstrates the reported bypass.

Before merging, could you please address two gaps:

1. **Cover all three adapters using the shared test infrastructure.** The PR changes Hono, Express, and Fastify, but currently tests only Hono. Since each adapter enforces body limits differently, please add a shared body-limit test suite under `server-adapters/_test-utils` and run it against all three adapters. At minimum, it should verify that an oversized DELETE body returns 413. The existing oversized POST assertion can also live there to verify the established behavior.

2. **Add patch changesets** for the three affected published packages: `@mastra/hono`, `@mastra/express`, and `@mastra/fastify`.

With those additions, the fix should be ready.

**Non-blocking:** Fastify determines whether to apply the body limit before expanding an `ALL` route into individual methods. As a result, POST, PUT, PATCH, and DELETE requests registered through `ALL` still appear to bypass the normal limit. This predates the PR, so it could be handled separately, but it would be good to open a follow-up issue.

shramanb113:
@NikAiyer thank you for reviewing my PR -  appreciate the detailed feedback.

1. Added a shared body-limit test suite under `server-adapters/_test-utils` and ran it against all three adapters (Hono, Express, Fastify), each asserting an oversized DELETE body returns 413. Moved the existing oversized POST assertion in there too, so both are now covered by the same suite going forward.

2. Added patch changesets for `@mastra/hono`, `@mastra/express`, and `@mastra/fastify`.

On the Fastify `ALL` route bypass - agreed it predates this PR and is worth handling separately. I'll open a follow-up issue for it rather than folding it into this one.

Let me know if there's anything else before merging.

wardpeet (approved):

shramanb113:
Did exactly what mastra-platform bot suggested , dropped the unrelated changes and rebased it into main 
@NikAiyer also did the changes that you mentioned in your review, would appreciate a feedback

shramanb113 (commented):
Would appreciate a detailed feedback on the changes I made and if any more fixes are required

abhiaiyer91 (approved):

PR #20046, opened 2026-07-23
feat(mastracode-web): direct GitHub App fallback

damien-schneider:
## What

When the platform-backed GitHub integration isn't in play (self-hosted or local deployments), wire a `GithubIntegration` from a complete `GITHUB_APP_*` env group so the app still gets a real GitHub connection — "Connect GitHub" in onboarding, the repository picker, and webhooks.

## Why

On `main`, the `mastracode-web` entry only gets GitHub through the platform-backed default. A self-hosted or local deployment that already has a GitHub App configured via env (`GITHUB_APP_ID`, `GITHUB_APP_PRIVATE_KEY`, `GITHUB_APP_CLIENT_ID`, `GITHUB_APP_CLIENT_SECRET`, `GITHUB_APP_SLUG`) still shows **"GitHub is not configured for this deployment"** in onboarding and can't create a first factory. This falls back to a directly-configured `GithubIntegration` built from that env group.

## Behavior

- **Complete `GITHUB_APP_*` group** → integration active: Connect GitHub, repo picker, webhooks.
- **Partial group** → stays disabled, so the status route reports exactly what's missing (matches the existing "boots when partially configured" behavior).
- **Auth wiring is unchanged** — this only adds to the `integrations` slot of `MastraFactory`.

## Test

Added a colocated case in `src/mastra/index.test.ts` asserting the entry registers `/auth/github/connect` — a route only the `GithubIntegration` adds — when the full group is configured.

## ELI5

Self-hosted deployments can now connect to GitHub directly when the platform connection is unavailable, as long as all required GitHub App settings are provided. Incomplete settings remain disabled and report what is missing.

## Changes

- Added a direct `GithubIntegration` fallback using `GITHUB_APP_*` environment variables.
- Supports GitHub connections, repository selection, and webhooks.
- Preserves existing authentication wiring and partial-configuration behavior.
- Added a test confirming `/auth/github/connect` is registered with complete configuration.

### DIFF
--- a/mastracode/web/src/mastra/index.test.ts
+++ b/mastracode/web/src/mastra/index.test.ts
@@ -74,6 +74,28 @@ describe('platform entry (src/mastra/index.ts)', () => {
     );
 
     it(
+      'registers the direct GitHub App integration when the full group is configured',
+      { timeout: 60_000 },
+      async () => {
+        vi.resetModules();
+        // No platform identity in this env, so the only source of a GitHub
+        // connection is the direct GITHUB_APP_* group wired in the entry.
+        vi.stubEnv('MASTRA_PLATFORM_SECRET_KEY', '');
+        vi.stubEnv('GITHUB_APP_ID', '12345');
+        vi.stubEnv('GITHUB_APP_PRIVATE_KEY', 'test-private-key');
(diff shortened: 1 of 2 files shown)


damien-schneider:
Folded into #20051 (cherry-picked as c87e3f0776) so the batch ships as a single PR on latest main, per review discussion.

PR #20073, opened 2026-07-23
Wire create-factory into the shared mastra CLI PostHog analytics

smthomas:
## Summary

Wires the `create-factory` CLI (`npm create factory`) into the same PostHog analytics used by `create-mastra`.

Previously `create-factory` shipped a bespoke minimal PostHog wrapper. It now reuses the shared `PosthogAnalytics` class from the `mastra` CLI (`mastra/dist/analytics/index.js`), which means:

- Same PostHog project, host, and anonymous distinct id stored in `~/.mastra/analytics.json`
- Same opt-out: set `MASTRA_TELEMETRY_DISABLED=1` (or `true`/`yes`) and no client is created and nothing is sent
- Standard `cli_session_start` and `cli_command` events with system properties, command duration, and success/error status (command name: `create-factory`)
- Only privacy-safe args are reported: whether the default template was used (never a custom template URL), `no_platform`, whether an org was passed, and the region
- The existing `sf_create_started` / `sf_create_completed` events keep working through the shared class

The direct `posthog-node` dependency is dropped since it now comes via the `mastra` package.

Note: `mastra factory dev` was already tracked by the mastra CLI (`startDevServer` reports `command: 'factory dev'`), so no changes were needed there.

## Test plan

- `pnpm check`, `pnpm lint`, and `pnpm test` (40 tests) pass in `mastracode/mastra-factory`
- Package builds with tsup; `node dist/index.js --help` runs (exercises the analytics bootstrap at module top-level)
- Runtime check: with `MASTRA_TELEMETRY_DISABLED=1`, `PosthogAnalytics` creates no client (empty distinct id) and `trackEvent`/`shutdown` are no-ops

## ELI5

The `create-factory` CLI now uses the same privacy-friendly analytics system as `create-mastra`. This keeps telemetry settings and anonymous identifiers consistent while removing duplicate analytics code.

## What changed

- Integrated shared `PosthogAnalytics`.
- Added session and command tracking with privacy-safe metadata.
- Preserved `sf_create_started` and `sf_create_completed` events.
- Respects `MASTRA_TELEMETRY_DISABLED`.
- Shares anonymous IDs with `create-mastra`.
- Removed the custom analytics wrapper and `posthog-node`.
- Updated tests and added a patch changeset.

## Validation

Checks, linting, tests, package build, CLI startup, and telemetry-disabled runtime behavior pass.


mastracode/mastra-factory/src/index.ts
```diff
@@ -4,14 +4,21 @@ import path from 'node:path';
 import { fileURLToPath } from 'node:url';
 import { Command } from 'commander';
 
-import { Analytics } from './analytics.js';
+import { PosthogAnalytics, setAnalytics } from 'mastra/dist/analytics/index.js';
```
wardpeet, line 7: this seems broken

abhiaiyer91: Verified this import end-to-end — it works, and it's the same pattern `create-mastra` already uses on main ([packages/create-mastra/src/index.ts#L4](https://github.com/mastra-ai/mastra/blob/main/packages/create-mastra/src/index.ts#L4)):

- `mastra` (packages/cli) exposes `"./dist/*"` in its exports map, ships `dist/` in `files`, and `dist/analytics/index.js` (+ `.d.ts`) is emitted by its build.
- Here `mastra` is a runtime dep of `create-factory` (it already was, pre-PR) and tsup keeps deps external, so the import resolves from the installed `mastra` package at runtime. `posthog-node` comes via `mastra`'s own deps.
- Smoke-checked the built bin: `node dist/index.js --help` / `--version` boot fine (exercises the top-level analytics bootstrap), and with `MASTRA_TELEMETRY_DISABLED=1` no client is created.
- `tsc --noEmit`, lint, and the package tests pass on the branch merged with current main (just pushed the merge; the one failing test, `sync-template.test.ts`, reproduces identically on clean main and is unrelated).

If the concern was about deep-importing `dist` paths in general, happy to add a proper `mastra/analytics` subpath export instead — but that'd change `create-mastra` too, so probably a separate PR.

mastracode/mastra-factory/src/index.ts
```diff
@@ -32,14 +39,27 @@ program
...
+        await create({
+          projectName: projectNameArg,
+          template: args.template,
+          // commander's `--no-platform` flips `platform` to false when present, true by default.
+          noPlatform: args.platform === false,
+          org: args.org ? String(args.org) : undefined,
+          region: args.region ? String(args.region) : undefined,
+          analytics,
+        });
+      },
```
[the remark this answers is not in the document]
abhiaiyer91: Fixed in 125fe8c14a. `create()` failures are now rethrown through `redactError()` so the analytics copy of `error.message` never carries user-supplied values, while stderr keeps the original message (retry hints, org list). Redacted: URL credentials (`//user:token@` anywhere in the message), the custom template URL including its degit (`https://github.com/`-stripped) and credential-stripped variants, `--org`, an invalid `--region`, the project name, and the enumerated org list in the `--org` mismatch error. Also stopped sending a raw invalid `--region` in the args payload — it now reports `invalid`. 10 regression tests in `src/utils/redact.test.ts` cover the clone-failure and validation paths.

Issue #15481, opened 2026-04-17
[Parity] Evented workflow engine: parallel multi-suspend coordination (skipped shared tests, processWorkflowSuspend active-path TODO)


h30s:
### Describe the Bug

There is a documented parity gap between the default workflow engine and the evented engine around parallel workflows where multiple branches can `suspend()`.

In the monorepo, this shows up as:
- `packages/core/src/workflows/evented/evented-workflow.test.ts` skips `resumeParallelMulti` and `resumeMultiSuspendError`, with a comment that parallel suspend has a race-related concern (each branch can publish `workflow.suspend` independently).
- `processWorkflowSuspend` in `packages/core/src/workflows/evented/workflow-event-processor/index.ts` contains a TODO: "if there are still active paths don't end the workflow yet" before publishing `workflows-finish`, without an `activeSteps`-style guard analogous to the foreach path (which waits / aggregates before emitting suspend-related events in the "check if all iterations are complete" logic).

I do **not** have a minimal user script where the **final** workflow result differs between engines for the basic parallel double-suspend scenario. A small public repro completes with `[OK]` and matching numeric results on both `--engine=default` and `--engine=evented`. The concern is **event ordering / coordination / parity** and un-skipping the shared suite, not a simple "wrong final output" in that script.

Related (not duplicates): [#6418](https://github.com/mastra-ai/mastra/issues/6418) / [#6423](https://github.com/mastra-ai/mastra/pull/6423) addressed **resume** completion semantics for parallel steps; this report is about **evented** suspend/finish **coordination** and **shared test** coverage.

### Steps To Reproduce

1. Clone the minimal repro (no Mastra monorepo required):

   git clone https://github.com/h30s/mastra_repo.git
   cd mastra_repo
   npm install

2. Run:

   npx tsx src/repro.ts --engine=evented
   npx tsx src/repro.ts --engine=default

3. For **internal** context (maintainers / contributors), open:
   - `packages/core/src/workflows/evented/evented-workflow.test.ts`  `skipTests.resumeParallelMulti` and `resumeMultiSuspendError`
   - `packages/core/src/workflows/evented/workflow-event-processor/index.ts`  `processWorkflowSuspend` and the TODO above `workflows-finish`

### Link to Minimal Reproducible Example

https://github.com/h30s/mastra_repo

### Expected Behavior

[...]

PR #20117, opened 2026-07-23
feat(core): opt-in 'called' tool-call concurrency strategy

### LINKED ISSUE #20100: [FEATURE] Opt-in toolCallConcurrency strategy to parallelize safe tool calls when an approval/suspend tool is registered but not called
# [FEATURE] Opt-in `toolCallConcurrency` strategy to parallelize safe tool calls when an approval/suspend tool is registered but *not called*

## Problem Statement

When an agent registers **any** tool with `requireApproval` or a `suspendSchema`, the agentic loop forces **every** tool-call batch to run sequentially (`concurrency: 1`) — even for a step whose emitted tool calls are all safe and never touch the approval/suspend tool.

The decision is made from the step's *effective active tool set*, not from the tools the model actually called:

`packages/core/src/loop/workflows/agentic-execution/tool-call-concurrency.ts`
```ts
export function effectiveToolSetRequiresSequentialExecution({ requireToolApproval, tools, activeTools }) {
  if (requireToolApproval) return true;
  // ...
  const activeToolEntries = activeTools === undefined ? Object.entries(tools) : /* filtered */;
  return activeToolEntries.some(([, tool]) => Boolean(tool.hasSuspendSchema || tool.requireApproval));
}
```

Because the check is over *available* tools rather than *called* tools, a long-lived agent that keeps a single approval/suspend tool registered across a run (e.g. `request_approval`, `ask_user`, `submit_plan`) can never run any parallel tool calls — the LLM's parallel `tool_use` blocks are executed one at a time.

This is the tool-level variant of the harness/subagent slowdown reported in #15887, and it's what PR #18243 tried to fix (later reverted in #18275 to preserve the conservative default). The conservatism exists to avoid a suspension racing a concurrent sibling — but a tool that the model **did not call this step cannot suspend this step**, so serializing a batch that never touched it buys no safety.

## Steps To Reproduce

Drop this test into `packages/core/src/agent/__tests__/` and run it against `main`. It uses the same mock-model + peak-concurrency harness as the existing `tool-concurrency-active-tools.test.ts`. **On `main` it fails: `peak === 1` (should be 2).**

```ts
// packages/core/src/agent/__tests__/repro-called-strategy.test.ts
import { convertArrayToReadableStream, MockLanguageModelV2 } from '@internal/ai-sdk-v5/test';
import { describe, expect, it } from 'vitest';
import { z } from 'zod/v4';
import { createTool } from '../../tools';
import { delay } from '../../utils';
import { Agent } from '../agent';

function twoParallelToolCalls(toolNames: string[]) {
  return new MockLanguageModelV2({
    doStream: async () => ({
      stream: convertArrayToReadableStream([
        { type: 'stream-start', warnings: [] },
        { type: 'response-metadata', id: 'id', modelId: 'm', timestamp: new Date() },
        ...toolNames.map((toolName, i) => ({
          type: 'tool-call' as const, toolCallType: 'function' as const,
          toolCallId: `call-${i}`, toolName, input: JSON.stringify({ data: `d${i}` }),
        })),
        { type: 'finish', finishReason: 'tool-calls', usage: { inputTokens: 1, outputTokens: 1, totalTokens: 2 } },
      ] as any),
      rawCall: { rawPrompt: null, rawSettings: {} }, warnings: [],
    }),
  });
}

it('serializes safe parallel calls because an approval tool is merely registered', async () => {
  const tracker = { running: 0, peak: 0 };
  const tracked = (id: string) => createTool({
    id, description: id, inputSchema: z.object({ data: z.string() }),
    execute: async () => { tracker.running++; tracker.peak = Math.max(tracker.peak, tracker.running); await delay(50); tracker.running--; return { ok: true }; },
  });
  const approval = createTool({
    id: 'request_approval', description: 'approve', inputSchema: z.object({ data: z.string() }),
    requireApproval: true, execute: async () => ({ ok: true }),
  });

  const agent = new Agent({
    id: 'repro', name: 'repro', instructions: 'x',
    model: twoParallelToolCalls(['tool-1', 'tool-2']),        // batch calls ONLY the safe tools
    tools: { 'tool-1': tracked('tool-1'), 'tool-2': tracked('tool-2'), request_approval: approval },
  });

  const stream = await agent.stream('go', { maxSteps: 1, toolCallConcurrency: 10 });
  for await (const _ of stream.fullStream) { /* drain */ }

  // Expected: 2 (the batch never calls request_approval, so it can't suspend).
  // Actual on main: 1 — request_approval is registered/active, forcing sequential.
  expect(tracker.peak).toBe(2);
});
```

## Expected Behavior

A step whose emitted tool calls are all safe should run at the configured concurrency, even while an approval/suspend tool stays registered. A step that **actually calls** an approval/suspend tool should still run sequentially, and a run-wide `requireToolApproval` policy should still force sequential.

## Proposed Solution

Add an **opt-in** strategy to `toolCallConcurrency` (default behavior unchanged, so existing users are unaffected):

```ts
// number (unchanged): limit, 'available' strategy
toolCallConcurrency: 8

// object form: pick the limit and/or strategy
toolCallConcurrency: { limit: 8, strategy: 'called' }
```

- `strategy: 'available'` (default): current behavior — any approval/suspend tool available in the step forces sequential.
- `strategy: 'called'`: resolve concurrency from the tools the model **actually called this step**. A pure-safe batch parallelizes; a batch that calls an approval/suspend tool still returns `1`; run-wide `requireToolApproval` still returns `1`.

The resolver already has the called tool names at the recompute point — plain path: `toolCalls` in the `map-tool-calls` step (`index.ts`); durable path: `inputData[].toolName` in the per-run foreach resolver added in #19329. The default path stays byte-identical.

**Safety:** this does not increase concurrent suspensions. The only way multiple suspends run in parallel today is via tools that aren't statically suspendable (e.g. subagent delegations) — unchanged here. Any batch that calls a statically-suspendable tool still serializes, so it stays within the existing envelope of the open multi-suspend issues (#16468, #15481).

I have a branch implementing this across the standard loop and the durable (`@mastra/inngest`/`@mastra/temporal`) engine, with unit + agent-level integration tests (safe batch → peak 2; batch that calls a suspend tool → peak 1) and a changeset. Happy to open the PR.

## Component

Agents, Tools

## Alternatives Considered

- **`activeTools` narrowing** — works, but it's per-`stream()`/`generate()` call. Long-lived pipeline agents run the whole pipeline in one turn (`maxSteps` high) and need the approval tool available across gate steps, so you can't drop it for the generation steps.
- **`prepareStep` returning narrowed tools** — viable per-step, but requires the caller to know, before the LLM runs, which step will fan out; for skills/VFS-driven pipelines the phase isn't cleanly available in `prepareStep`.
- **`toolCallConcurrency: N` alone** — ineffective; the number is ignored while any approval/suspend tool is available.

## Example Use Case

A video-generation agent registers `request_approval` (a `suspendSchema` HITL tool) for human gates, and generates N scenes by emitting N parallel `generate_video` calls in one step. Today every scene renders sequentially — N× the wall-clock — purely because `request_approval` is registered, even though that step never calls it. With `toolCallConcurrency: { limit: 8, strategy: 'called' }` the scene batch renders in parallel while the approval gates (separate steps that *do* call `request_approval`) still suspend correctly.

## Additional Context

Lineage: origin complaint #15887; parallel-exec + configurable concurrency #7230 / #9704 / #12698; active-tools narrowing #15977 / #15978; batch-aware attempt + revert #18243 / #18275; durable-engine parity #19317 / #19329. This request is the explicit, opt-in version of #18243 — it preserves the conservative default that #18275 restored, so it doesn't reopen that decision.

## Verification

devangpadhiyar:
## Description

Adds an opt-in `'called'` tool-call concurrency strategy so batches of purely safe tool calls run in parallel even when an approval/suspending tool stays registered on the agent.

Today the loop decides sequential-vs-parallel from the step's *effective active tool set*: any registered-and-active tool with `requireApproval` or a `suspendSchema` forces the whole batch to `concurrency: 1`, even when the model's actual tool calls are all safe and never touch it. A long-lived agent that keeps a single approval tool registered across a run therefore can never parallelize any tool calls. A tool the model did not call this step cannot suspend this step, so serializing that batch buys no safety.

`toolCallConcurrency` now accepts an object form in addition to a number:

```ts
toolCallConcurrency: 8                                // unchanged: limit, 'available' strategy (default)
toolCallConcurrency: { limit: 8, strategy: 'called' } // NEW: decide from the tools actually called this step
```

- `strategy: 'available'` (default, unchanged): any approval/suspend tool available in the step forces sequential.
- `strategy: 'called'`: only the tools the model actually called this step are checked. A pure-safe batch parallelizes; a batch that calls an approval/suspend tool still returns `1`; a run-wide `requireToolApproval` policy still returns `1`.

This is the explicit, opt-in version of #18243 — the conservative `'available'` default that #18275 restored is preserved byte-for-byte, so it doesn't reopen that decision.

Implementation notes:
- Standard loop: `resolveToolCallConcurrency` gains an optional `calledToolNames`; the `map-tool-calls` step passes the called tool names when the strategy is `'called'` (the recompute already runs after the LLM output, where `toolCalls` are known).
- Durable engine (`@mastra/inngest` / `@mastra/temporal`): `resolveDurableToolCallConcurrency` narrows to the called tools via `inputData[].toolName` in the per-run foreach resolver added in #19329 (leak-safe, no shared mutable state).
- `toolCallConcurrency` widened to `number | { limit?: number; strategy?: 'available' | 'called' }` and threaded through the existing option types (union flows through the existing pass-throughs unchanged).

**Safety re: multi-suspend:** this does not increase concurrent suspensions. The only way multiple suspends run in parallel today is via tools that aren't statically suspendable (e.g. subagent delegations), which is unchanged. Any batch that calls a statically-suspendable tool still serializes, so it stays within the existing envelope of #16468 / #15481.

## Related issue(s)

Fixes #20100

## Type of change

- [ ] Documentation update
- [ ] Code refactoring
- [x] Performance improvement
- [x] Test update

## Checklist

Tests added:
- `loop/workflows/agentic-execution/tool-call-concurrency.test.ts` — `'called'` strategy unit cases + object-config parsing.
- `agent/durable/workflows/shared/tool-call-concurrency.test.ts` — durable `'called'` strategy unit cases.
- `agent/__tests__/tool-concurrency-active-tools.test.ts` — agent-level integration: safe batch with an approval tool registered → peak concurrency 2; a batch that calls a suspending tool → peak 1.

### DIFF
--- a/.changeset/batch-aware-tool-concurrency.md
+++ b/.changeset/batch-aware-tool-concurrency.md
@@ -1,0 +1,20 @@
+---
+'@mastra/core': minor
+---
+
+Add an opt-in `'called'` tool-call concurrency strategy so batches of purely safe tool calls run in parallel even when an approval/suspending tool stays registered on the agent.
+
+`toolCallConcurrency` now accepts an object form in addition to a number:
+
+```ts
+// number (unchanged): concurrency limit, 'available' strategy
+toolCallConcurrency: 8
+
+// object form: pick the limit and/or strategy
+toolCallConcurrency: { limit: 8, strategy: 'called' }
+```
+
+- `strategy: 'available'` (default, unchanged): any approval/suspending tool *available* in the step forces the whole batch sequential, even if the model did not call it. Conservative — a suspend can never race a sibling.
+- `strategy: 'called'`: only the tools the model *actually called* this step are checked. An available-but-uncalled approval/suspend tool no longer serializes a batch of safe calls; a batch that *does* call one still runs sequentially (concurrency 1), and a run-wide `requireToolApproval` policy still forces sequential. Useful for agents (e.g. multi-step generation pipelines) that keep an approval tool registered across a run but never mix it into the same batch as parallelizable calls.
+
+Applies to both the standard loop and the durable (`@mastra/inngest` / `@mastra/temporal`) engine.
--- a/client-sdks/client-js/src/route-types.generated.ts
+++ b/client-sdks/client-js/src/route-types.generated.ts
@@ -7,22 +7,22 @@
 
 export type Simplify<T> = { [K in keyof T]: T[K] } & {};
 
-type Shared_Auxiliary_290 =
+type Shared_Auxiliary_294 =
   | string
   | number
   | boolean
   | null
-  | Shared_Auxiliary_290[]
+  | Shared_Auxiliary_294[]
   | {
-      [key: string]: Shared_Auxiliary_290;
+      [key: string]: Shared_Auxiliary_294;
     };
 
-type Shared_Auxiliary_1159 = {
+type Shared_Auxiliary_1163 = {
   id?: string | undefined;
   name: string;
   type: 'file' | 'folder';
   content?: string | undefined;
-  children?: Shared_Auxiliary_1159[] | undefined;
+  children?: Shared_Auxiliary_1163[] | undefined;
 };
 
 type Shared_Type_0 = {
@@ -90,7 +90,15 @@ type Shared_Type_4 = {
   tracingOptions?: Shared_Type_3 | undefined;
   requireToolApproval?: boolean | undefined;
   autoResumeSuspendedTools?: boolean | undefined;
-  toolCallConcurrency?: number | undefined;
+  toolCallConcurrency?:
+    | (
+        | number
+        | {
+            limit?: number | undefined;
+            strategy?: ('available' | 'called') | undefined;
+          }
+      )
+    | undefined;
   includeRawChunks?: boolean | undefined;
   [x: string]: unknown;
 };
@@ -976,7 +984,7 @@ type Shared_Type_49 = {
   providerOptions?:
     | {
         [key: string]: {
-          [key: string]: Shared_Auxiliary_290;
+          [key: string]: Shared_Auxiliary_294;
         };
       }
     | undefined;
@@ -990,7 +998,7 @@ type Shared_Type_50 = {
   providerOptions?:
     | {
         [key: string]: {
-          [key: string]: Shared_Auxiliary_290;
+          [key: string]: Shared_Auxiliary_294;
         };
       }
     | undefined;
@@ -1005,13 +1013,13 @@ type Shared_Type_51 = {
     | undefined;
   metadata?:
     | {
-        [key: string]: Shared_Auxiliary_290;
+        [key: string]: Shared_Auxiliary_294;
       }
     | undefined;
   providerOptions?:
     | {
         [key: string]: {
-          [key: string]: Shared_Auxiliary_290;
+          [key: string]: Shared_Auxiliary_294;
         };
       }
     | undefined;
@@ -1099,7 +1107,7 @@ type Shared_Type_54 = {
   createdAt?: (string | Date) | undefined;
   metadata?:
     | {
-        [key: string]: Shared_Auxiliary_290;
+        [key: string]: Shared_Auxiliary_294;
       }
     | undefined;
   attributes?:
@@ -1113,7 +1121,7 @@ type Shared_Type_54 = {
   providerOptions?:
     | {
         [key: string]: {
-          [key: string]: Shared_Auxiliary_290;
+          [key: string]: Shared_Auxiliary_294;
         };
       }
     | undefined;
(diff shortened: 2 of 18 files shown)


abhiaiyer91:
Closing this out — the linked issue (#20100) has since been closed, so this change no longer has an active issue behind it.

Thanks for the contribution, and sorry for the round trip. If you believe the underlying problem still exists, please reopen this PR or comment here and we'll take another look.

PR #20128, opened 2026-07-23
fix(pg): avoid transactions for distributed locks

arctic-char:
## Summary

Replace PostgreSQL transaction-scoped advisory locks with session-scoped advisory locks for Factory distributed locking.

GitHub project operations still need serialization because ensure, commit, push, and PR flows can mutate shared checkout state. Dropping the lock would allow concurrent replicas to interleave those operations. Adding Redis would introduce a new dependency without addressing the existing PostgreSQL implementation directly.

The previous implementation held an open transaction for the entire critical section. When sandbox or GitHub I/O stalled, PostgreSQL showed an idle transaction waiting in ClientRead while the advisory lock remained held. The new implementation checks out a dedicated pool connection, acquires pg_advisory_lock, runs the critical section without BEGIN, and explicitly unlocks before returning the connection. PostgreSQL still releases the session lock automatically if the connection is lost.

This keeps cross-replica serialization while removing the long-lived transaction. It does still reserve one pool connection per active distributed lock, which is required for a session-scoped PostgreSQL lock, but the connection remains idle rather than idle-in-transaction.

## Test plan

- Factory GitHub project-lock scenario tests: 4 passed
- ESLint on all changed TypeScript files: passed
- git diff --check: passed
- Added a PostgreSQL regression test that inspects pg_locks and pg_stat_activity during the critical section and verifies the lock-holding session is idle, not idle in transaction
- The focused PostgreSQL test/build is currently blocked before reaching these tests by unresolved @mastra/core workspace dist imports; the focused Factory scenario test passes

### DIFF
--- a/.changeset/chubby-ducks-stop.md
+++ b/.changeset/chubby-ducks-stop.md
@@ -1,0 +1,5 @@
+---
+'@mastra/factory': patch
+---
+
+Fixed GitHub project locking so sandbox and network operations remain serialized without holding database transactions open.
--- a/.changeset/clear-hornets-roll.md
+++ b/.changeset/clear-hornets-roll.md
@@ -1,0 +1,5 @@
+---
+'@mastra/pg': patch
+---
+
+Fixed distributed locks so long-running critical sections no longer keep PostgreSQL transactions open.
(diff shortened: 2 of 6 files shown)


arctic-char:
Closing because session-scoped advisory locks still reserve a pool connection for the full critical section and, with Neon's idle session timeout disabled, can retain the connection and lock indefinitely when external I/O hangs. We'll replace this with a bounded lease-based design that does not hold database connections across slow I/O.

PR #20132, opened 2026-07-23
fix(factory): record named steps under project lock for diagnostic breadcrumbs

rphansen91:
> Stacked on #20129. Review that PR first; merge base is `fix/factory-lock-and-fetch-timeouts`.

## Summary

Answer Charlie's question from #20129: *"what if the slow call is one of our own platform calls, not a third party?"* Right now `ProjectLockTimeoutError` only tells us "60s elapsed" — not whether Fleet was slow, our sandbox HTTP was slow, GitHub App API was slow, or our own storage was slow. That's not enough to diagnose the next incident.

This PR adds a `LockStepRecorder` that `fn()` receives as its second argument. Callers wrap named boundaries with `recorder.step('sandbox.commitAll', () => commitAll(...))`; the recorder captures start time, duration, and outcome for each step (cheap: no I/O, just `Date.now()` bookkeeping). Two consumers:

1. **On timeout** the recorded steps and the still-running step are attached to `ProjectLockTimeoutError.steps` and `.currentStep`, and the error message includes the current step name. So instead of `Project lock critical section for "proj123:userA" exceeded 60000ms` we get `... exceeded 60000ms (currentStep=sandbox.pushBranch)` plus the full breadcrumb array on the error.
2. **On non-timeout slow completion** — when the total exceeds `MASTRACODE_PROJECT_LOCK_SLOW_WARN_MS` (default 5s) — a structured `warn` is emitted with the same breadcrumbs. This catches platform-call regressions long before they cascade into a full timeout.

All four GitHub route lock sites are wired: commit, push, PR create, sandbox teardown. Named steps cover the platform-call surface that actually runs under the lock: `fleet.resolveSandbox`, `fleet.reattachSandbox`, `sandbox.commitAll`, `sandbox.pushBranch`, `sandbox.teardown`, `github.getRepositoryAccess`, `github.createPullRequest`, `auth.ensureUser`, `storage.subscribeToPullRequest`, and the corresponding audit emits.

## What changed

- `mastracode/factory/src/integrations/github/project-lock.ts`
  - New `LockStepRecorder` interface (and internal `MutableLockStepRecorder`) with `step(name, fn)` + `entries` + `currentStep`.
  - `ProjectLockTimeoutError` now carries `steps: ReadonlyArray<LockStepEntry>` and `currentStep: string | null`; message includes `(currentStep=…)`.
  - `runWithTimeout` builds the recorder, passes it to `fn(signal, recorder)`, snapshots on timeout, and on successful completion emits a structured `warn` when `totalMs >= slowWarnThresholdMs()`.
  - `withProjectLock` / `withDbAdvisoryLock` accept an optional `logger?: LockLogger` so callers (and tests) can capture the slow-lock warn without touching `console`.
  - Ring buffer capped at 32 entries per invocation (typical is <10). Uses object-identity to replace the reserved "running" slot on completion, so even if the ring buffer evicts an old entry the outcome is preserved as a tail append.
- `mastracode/factory/src/integrations/github/routes.ts`
  - The 4 `withProjectLock` call sites now use `fn: async (_signal, steps) => { await steps.step('…', () => …) }` for every awaited call inside the lock — including audit emits, so a slow logging backend also shows up in the breadcrumbs.

## Example output

Timeout error:
```
ProjectLockTimeoutError: Project lock critical section for "proj123:userA" exceeded 60000ms (currentStep=sandbox.pushBranch)
  steps: [
    { name: 'fleet.resolveSandbox',       startedAtMs: 0,   durationMs: 42,    outcome: 'ok'      },
    { name: 'github.getRepositoryAccess', startedAtMs: 42,  durationMs: 118,   outcome: 'ok'      },
    { name: 'sandbox.pushBranch',         startedAtMs: 160, durationMs: 59840, outcome: 'running' },
  ]
```

Slow-lock warn (total 7.2s under a 5s threshold):
```
[project-lock] slow critical section for "proj123:userA" took 7213ms
  meta: {
    key: 'proj123:userA',
    totalMs: 7213,
    thresholdMs: 5000,
    steps: [
      { name: 'fleet.resolveSandbox', durationMs: 38,   outcome: 'ok' },
      { name: 'sandbox.commitAll',    durationMs: 7160, outcome: 'ok' },
      { name: 'audit.git.commit',     durationMs: 15,   outcome: 'ok' },
    ],
  }
```

## Test plan

- 5 new tests in `project-lock-scenario.test.ts`:
  - Records completed steps in order with durations and `ok` outcomes.
  - Records `error` outcome for a failing step and rethrows.
  - Attaches step history and `currentStep` to `ProjectLockTimeoutError`, including the still-running step with a live duration.
  - Emits a slow-lock warn when total exceeds the threshold; meta payload includes key, `totalMs`, `thresholdMs`, and the step names + outcomes.
  - Does *not* warn when the critical section stays under the threshold.
- Existing 8 scenario tests still pass (cross-replica serialization, ignore-signal, `withDbAdvisoryLock` direct usage, etc.).
- `pnpm --filter ./mastracode/factory exec vitest run src/integrations/github` — 15 files, **271 tests passed**.
- Typecheck on `mastracode/factory` — clean.
- Lint on `mastracode/factory` — clean.

## Notes for reviewers

- Stacked on #20129 because the recorder extends the timeout infrastructure introduced there (`ProjectLockTimeoutError`, `runWithTimeout`, `AbortSignal` on `fn`). Merging in order: #20129 first, then this.
- The `fn(signal, recorder)` shape means the recorder is *always* live under the lock, even for callers that don't opt in — they simply produce empty `steps` arrays. There is no runtime cost when unused.
- Threshold is env-tunable (`MASTRACODE_PROJECT_LOCK_SLOW_WARN_MS`), settable to `0` to disable warns entirely for local dev.
- No production log-format lock-in: the warn goes through the existing `LockLogger` shape, so we can swap in structured logging (pino, etc.) at the caller boundary without touching `project-lock.ts`.

### DIFF
--- a/.changeset/factory-lock-step-breadcrumbs.md
+++ b/.changeset/factory-lock-step-breadcrumbs.md
@@ -1,0 +1,5 @@
+---
+'@mastra/factory': patch
+---
+
+Add a step recorder to `withProjectLock` so we can tell which platform call was in flight when a critical section runs long or times out. `fn` now receives a `LockStepRecorder` as its second argument; wrapping each named boundary (e.g. `fleet.resolveSandbox`, `sandbox.commitAll`, `github.getRepositoryAccess`, `sandbox.pushBranch`) captures start time, duration, and outcome. On timeout the recorded steps and the currently-running step are attached to `ProjectLockTimeoutError`. When a critical section completes but exceeds `MASTRACODE_PROJECT_LOCK_SLOW_WARN_MS` (default 5s), a structured `warn` is emitted with the same breadcrumbs so platform-call regressions are visible before they cascade into a full timeout. All four GitHub route lock sites (commit / push / pr / sandbox teardown) are wired.
--- a/.changeset/factory-lock-timeout.md
+++ b/.changeset/factory-lock-timeout.md
@@ -1,0 +1,5 @@
+---
+'@mastra/factory': patch
+---
+
+Bound the `withProjectLock` / `withDbAdvisoryLock` critical section with an `AbortSignal` timeout (default 60s, configurable via `timeoutMs`). Previously, an unbounded outbound call inside the lock could keep the transaction open for up to Neon's `idle_in_transaction_session_timeout` (5 minutes), pinning the pool connection and the advisory lock the entire time. On timeout the wrapper aborts the `fn`'s signal, rolls the transaction back, releases the connection, and throws `ProjectLockTimeoutError`.
--- a/.changeset/sdk-anthropic-oauth-timeout.md
+++ b/.changeset/sdk-anthropic-oauth-timeout.md
@@ -1,0 +1,5 @@
+---
+'@mastra/code-sdk': patch
+---
+
+Added a 15s `AbortSignal` timeout to the Anthropic OAuth token-exchange and refresh fetches so an unresponsive upstream cannot pin the caller indefinitely.
--- a/mastracode/factory/src/integrations/github/project-lock-scenario.test.ts
+++ b/mastracode/factory/src/integrations/github/project-lock-scenario.test.ts
@@ -1,6 +1,12 @@
 import { afterEach, beforeEach, describe, expect, it } from 'vitest';
 import type { LockClient, LockPool } from './project-lock.js';
-import { __resetProjectLocksForTests, hashKey, withDbAdvisoryLock, withProjectLock } from './project-lock.js';
+import {
+  __resetProjectLocksForTests,
+  hashKey,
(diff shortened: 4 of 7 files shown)


rphansen91:
Superseded by #20135, which deletes the project-lock advisory-lock machinery entirely (including `runWithTimeout` / `ProjectLockTimeoutError` on the git routes). The breadcrumb recorder was built on top of that machinery, so there's nothing left to attach it to. Closing.

PR #20136, opened 2026-07-23
feat(core): support per-item experiment timeouts

YujohnNattrass:
PR 1 of 4 for [MASTRA-4509](https://linear.app/kepler-crm/issue/MASTRA-4509/oss-experiments-per-item-timeout-overrides).

Adds optional timeout overrides to inline and persisted dataset items. An item's `timeout` takes precedence over the experiment's `itemTimeout` fallback.

```ts
await dataset.addItem({ input, timeout: 5_000 })

await dataset.startExperiment({
  targetType: 'agent',
  targetId: 'agent',
  itemTimeout: 30_000,
})
```

Both fields accept positive integer milliseconds up to `1_800_000` (30 minutes). Invalid values fail before persistence or execution instead of disabling the deadline or leaking a platform `RangeError`.

The effective timeout is one budget across execution, retry backoff, and retries. External cancellation remains authoritative, and inline tasks are hard-raced against the same signal.

This PR contains the core contract, in-memory persistence, and execution behavior. External storage support continues in #20137.

Verified with 203 focused tests, lint, and the core build.

## ELI5

Each experiment item can have its own time limit. The item limit overrides the experiment default and stops execution, retries, and retry delays when it expires.

## What changed

- Added optional `timeout` fields to inline and persisted dataset items.
- Added `itemTimeout` as the fallback for experiment items.
- Validated timeout values before execution or persistence.
- Limited timeouts to positive integers up to 30 minutes.
- Persisted `timeout` through supported storage adapters and SCD-2 history.
- Rejected timeout writes in MySQL, which does not support this feature.
- Applied one timeout budget to execution, retries, and retry backoff.
- Preserved external cancellation as authoritative.
- Added abort-aware execution for inline tasks, agents, and workflows.
- Exported `raceWithSignal` and added `ExecutionResult.aborted`.
- Added regression tests for timeout, cancellation, retry, and persistence behavior.

Verified with 203 focused tests, lint, and the core build.


YujohnNattrass:

- #20136 ◀
    - #20137
        - #20138
            - #20003

packages/core/src/datasets/experiment/index.ts
```diff
@@ -360,38 +392,57 @@ export async function runExperiment(mastra: Mastra, config: ExperimentConfig): P
         }
 
         const itemStartedAt = new Date();
-        // Compose per-item signal (timeout + run-level abort)
+        // Compose one per-item signal before the first attempt so timeout is a whole-item budget.
+        const effectiveTimeout = item.timeout ?? itemTimeout;
         let itemSignal: AbortSignal | undefined = signal;
-        if (itemTimeout) {
-          const timeoutSignal = AbortSignal.timeout(itemTimeout);
+        if (effectiveTimeout) {
```
NikAiyer: `DatasetItemPayload.timeout` is part of the public core API, but core currently accepts any finite number. This truthy check makes `timeout: 0` silently disable the deadline, while negative or fractional values reach `AbortSignal.timeout()` and throw a `RangeError` outside `execFn`, potentially failing the experiment instead of producing an item failure. The server and Studio follow-ups define this as a positive whole-number millisecond value, but direct core consumers bypass those boundaries. Could we validate that invariant in core for both inline and persisted items, with coverage for `0`, negative, and fractional values?

NikAiyer (changes requested):
The per-item deadline composition, abort provenance, retry accounting, persistence/history handling, and focused test coverage look solid. I found one core contract issue around invalid timeout values that should be addressed before merge; details inline. The currently non-green CI jobs appear unrelated to this PR.

packages/core/src/datasets/experiment/index.ts
YujohnNattrass: Addressed in `bae7c9ca7f`. Core now validates both per-item `timeout` and run-level `itemTimeout` before persistence or execution. Values must be positive integer milliseconds no greater than `1_800_000` (30 minutes); invalid values throw `EXPERIMENT_TIMEOUT_INVALID`. I also changed the composition check to `effectiveTimeout !== undefined` and added coverage for `0`, `-1`, `1.5`, `MAX + 1`, and the accepted maximum.

NikAiyer (approved):
Thanks for addressing this. The shared validation now consistently covers experiment-level configuration, inline and persisted item overrides, and dataset storage writes. The focused experiment and dataset tests pass locally. I also investigated the affected-unit failure; it is an unrelated durable-agent parity failure for the previously introduced `experimentalTransform` option. Approved.

packages/core/src/datasets/experiment/__tests__/runExperiment.test.ts
```diff
@@ -627,6 +628,184 @@ describe('runExperiment', () => {
...
+          return 'done';
+        },
+      });
+
+      const byName = (name: string) => result.results.find(item => (item.input as { name: string }).name === name);
+      expect(byName('short')?.error?.message).toMatch(/timeout/i);
+      expect(byName('inherited')?.error?.message).toMatch(/timeout/i);
+      expect(byName('long')?.error).toBeNull();
+    });
+
```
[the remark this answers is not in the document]
YujohnNattrass: This is not present in the published restacked source. The current file has a single `const byName` declaration at line 731, followed directly by the three assertions. The core build and full core verification also passed, so there is no redeclaration to remove. This appears to have been analyzed from an intermediate conflict state.

PR #20137, opened 2026-07-23
feat(stores): persist dataset item timeouts

YujohnNattrass:
PR 2 of 4 for [MASTRA-4509](https://linear.app/kepler-crm/issue/MASTRA-4509/oss-experiments-per-item-timeout-overrides), stacked on #20136.

Persists dataset item timeout overrides across LibSQL, PostgreSQL, MongoDB, and Spanner, including updates, batches, tombstones, and SCD-2 history.

```ts
const item = await dataset.addItem({ input, timeout: 5_000 })

await dataset.updateItem({
  itemId: item.id,
  timeout: 15_000,
})
```

Supported stores preserve the override across reads and dataset versions. MySQL rejects timeout writes before mutation with `MYSQL_DATASET_ITEM_TIMEOUT_UNSUPPORTED` rather than silently dropping the value.

API and documentation support continues in #20138.

Verified with all five adapter builds and linters, supported-adapter round trips, and MySQL integration and unit rejection tests.


YujohnNattrass:

- #20136
    - #20137 ◀
        - #20138
            - #20003

stores/_test-utils/src/domains/datasets/index.ts
```diff
@@ -573,6 +574,77 @@ export function createDatasetsTests({
...
+            )?.timeout,
+          ).toBe(1_000);
+          expect(
+            (await datasetsStorage.getItemsByVersion({ datasetId: ds.id, version: 4 })).find(
+              versionItem => versionItem.id === item.id,
+            )?.timeout,
+          ).toBe(4_000);
+
+          const history = await datasetsStorage.getItemHistory(item.id);
+          expect(history.map(row => row.timeout)).toEqual([4_000, 1_000, 1_000]);
```
NikAiyer, line 720: Non-blocking: could we also assert that `timeout` is preserved on both single-item and batch deletion tombstones? Tombstone behavior is already tested generally, but field-specific assertions would directly cover the PR’s stated deletion guarantee.

YujohnNattrass: Addressed in `7ddd16c44e`. The shared capability-gated adapter test now verifies that `deleteItem` retains the replaced 4,000 ms timeout on its newest tombstone and that `batchDeleteItems` retains the original 2,000/3,000 ms timeouts on both newest tombstones. Verified through LibSQL, PostgreSQL, MongoDB, and Spanner.

NikAiyer (approved):
Re-reviewed after the follow-up. The shared adapter coverage now directly verifies timeout preservation across both single-item and batch deletion tombstones. The original concern is resolved, and the PR is ready to merge.

PR #20138, opened 2026-07-23
feat(datasets): expose item timeouts through APIs

YujohnNattrass:
PR 3 of 4 for [MASTRA-4509](https://linear.app/kepler-crm/issue/MASTRA-4509/oss-experiments-per-item-timeout-overrides), stacked on #20137.

Exposes dataset item `timeout` overrides and the experiment-level `itemTimeout` fallback through the server and client-js APIs.

```ts
await client.addDatasetItem({
  datasetId,
  input,
  timeout: 5_000,
})

await client.triggerDatasetExperiment({
  datasetId,
  targetType: 'agent',
  targetId: 'agent',
  itemTimeout: 30_000,
})
```

Server schemas enforce the core range of 1 to `1_800_000` milliseconds (30 minutes). The generated client route types and CLI metadata are updated from those schemas, and the documentation explains the limit, timeout precedence, retry-budget behavior, and cancellation semantics.

Studio support continues in #20003.

Verified with server and client builds, tests, and lint, plus generated-contract, formatting, Remark, and documentation validation checks.


YujohnNattrass:

- #20136
    - #20137
        - #20138 ◀
            - #20003

packages/server/src/server/schemas/datasets.ts
```diff
@@ -313,6 +314,7 @@ export const updateItemBodySchema = z.object({
   groundTruth: z.unknown().optional().describe('Expected output for comparison'),
   expectedTrajectory: trajectoryExpectationSchema,
   toolMocks: toolMocksSchema,
+  timeout: z.number().int().positive().optional().describe('Execution timeout override in milliseconds'),
```
NikAiyer: Non-blocking question: once an item has a `timeout`, is there an intended way to remove that override? `updateItem()` treats omission as preserving the current value, while this schema rejects `null`, so the API supports setting or replacing a timeout but not clearing it. This may be an existing core contract limitation rather than something this PR should address, but I wanted to confirm that the behavior is intentional.

NikAiyer (approved):
The API exposure is consistent across server validation, handlers, generated contracts, client types, tests, and docs. The stacked core coverage also verifies precedence and whole-item timeout budgeting. I only have one non-blocking question inline.

packages/server/src/server/schemas/datasets.ts
YujohnNattrass: Confirmed: clearing a persisted timeout is intentionally outside this stack. Existing update semantics treat omission as preservation, and the contract does not accept `null`; adding clearing would require coordinated core, storage, and API semantics. The Studio form reflects that limitation by preventing a persisted override from being blanked.

packages/server/src/server/schemas/datasets.ts
```diff
@@ -377,6 +389,7 @@ export const datasetItemResponseSchema = z.object({
   groundTruth: z.unknown().optional(),
   expectedTrajectory: z.unknown().optional(),
   toolMocks: toolMocksSchema,
+  timeout: experimentTimeoutSchema.optional(),
```
NikAiyer, line 403: Could we keep the 30-minute cap on request schemas only? Core and storage still allow values above this limit, so reusing the capped schema for responses means an item created directly through `@mastra/core`, imported, or already persisted with `timeout > 1_800_000` no longer conforms to this API response contract (and the same applies to `itemVersionResponseSchema`). A separate representational response schema, such as an unrestricted number, would preserve compatibility. The core reference docs should likewise avoid presenting this API-only cap as a constraint of `dataset.addItem()`, `updateItem()`, and `startExperiment()` unless core adopts the same invariant.

NikAiyer (changes requested):
The API input cap makes sense, but the same capped schema is now used for persisted-item responses even though core/storage do not enforce that invariant. This can make the API contract reject or misrepresent items created through core, imports, or existing storage with larger timeout values. The core reference docs also currently describe this server-only limit as a core method constraint.

packages/server/src/server/schemas/datasets.ts
YujohnNattrass: Addressed in the rewritten core layer (#20136, commit `d62e17da35`). The 30-minute limit is now a core/storage invariant rather than request-only validation:

- `validateExperimentTimeout()` enforces a positive integer no greater than `1_800_000` ms in `packages/core/src/datasets/validation/timeout.ts`.
- `DatasetStorageBase` applies it to update and batch-insert writes in `packages/core/src/storage/domains/datasets/base.ts`.
- `runExperiment()` validates both the run-level `itemTimeout` and every resolved inline or stored item's `timeout` before execution in `packages/core/src/datasets/experiment/index.ts`.

Core tests cover `0`, negative, fractional, and over-cap rejection plus acceptance at the maximum. Direct `@mastra/core` writes and experiment execution therefore cannot create or use values outside the response schema's range, so the capped response schema and core reference docs now describe the same contract.

abhiaiyer91:
## Factory review — approve

Reviewed the full diff, the stacked base, prior review history, and ran verification locally on a fresh install with `@mastra/core` built from this stack.

### What this PR does
Exposes the per-item `timeout` override and the experiment-level `itemTimeout` fallback through the server route schemas/handlers and client-js, bounded to `1..1_800_000` ms, plus regenerated contracts (client route types, CLI route metadata), docs, and changesets. Scope is correctly limited to the API-exposure layer — persistence and precedence logic land in the stacked bases (#20136 / #20137).

### Correctness
- `experimentTimeoutSchema` is applied consistently across `addItemBodySchema`, `updateItemBodySchema`, `batchInsertItemsBodySchema`, `triggerExperimentBodySchema`, and the item/version response schemas. Handlers destructure and forward `timeout` / `itemTimeout` to `addItem`, `updateItem`, `addItems`, and the experiment trigger; client-js passes them through the existing body spread, so no serialization special-casing was needed.
- Docs claims match the implementation: `packages/core/src/datasets/experiment/index.ts` composes one per-item `AbortSignal.timeout(item.timeout ?? itemTimeout)` **before the first attempt**, so the timeout is a single whole-item budget covering retries and backoff, and the deadline abort is explicitly non-retryable. The "takes precedence over the experiment's `itemTimeout`" wording is likewise accurate.

### Re: the earlier changes-requested concern (capped response schemas)
Confirmed resolved, on two independent grounds:
1. Core now enforces the same cap on every write path — `validateExperimentTimeout()` is called from `DatasetStorageBase.addItem/updateItem/batchInsertItems` and from `runExperiment()` — so out-of-range values cannot be persisted through core or imports.
2. `responseSchema` is consumed only by the OpenAPI/manifest/type generation layer (`openapi-utils.ts`, `api-schema-manifest.ts`); routes do not validate outbound payloads against it, so even a legacy out-of-range value could not cause a runtime rejection.

### Verification run locally
- `packages/server` → `vitest run src/server/handlers/datasets.test.ts`: 26/26 passed
- `client-sdks/client-js` → `vitest run src/client.test.ts`: 49/49 passed
- `tsc --noEmit` clean in client-js; clean in server apart from pre-existing `@mastra/agent-builder` module-resolution errors from unbuilt sibling packages (unrelated to this PR, same in the full-suite collection failures)
- Re-ran both generators (`generate:route-types`, `generate:api-cli-route-metadata`) — **zero diff**, so the committed generated artifacts are genuinely in sync with the schemas
- All GitHub checks green

### Non-blocking nit (follow-up, not a merge blocker)
`packages/server/src/server/schemas/datasets.ts:14` redefines `MAX_EXPERIMENT_ITEM_TIMEOUT_MS = 30 * 60 * 1000` locally, while core already exports the same constant from `@mastra/core/datasets`. Importing it would remove the drift risk of the two bounds diverging later. Left alone here since the server doesn't currently import from that subpath anywhere.

**Verdict: approve.** No blocking findings.

damien-schneider (approved):

PR #20141, opened 2026-07-23
fix(core): send each conversation message to the title generation model exactly once

TylerBarnes:
Thread title generation was sending every conversation message to the title model twice. The title call sites read messages through the deprecated AIV4 UI shape, which carries the same text in both `content` and a text `part` — and `formatMessagesForTitle` emitted both:

```
User: hi
User: hi
Assistant: Hello! How can I help?
Assistant: Hello! How can I help?
```

That duplicated greeting transcript confused small title models into replying to the conversation instead of titling it, so threads could end up with refusal text as their title.

Now the formatter skips `content` only when a text part carries the exact same text (falling back to `content` for messages with no matching text part, so content-only messages still emit):

```
User: hi
Assistant: Hello! How can I help?
```

Also refreshed the default title instructions — they still said "based on the first message", but the input has been the full conversation transcript since April. The new text describes the role-prefixed transcript and tells the model never to reply to it. Three docs pages that described the input as "the user's first message" were updated to match.

Regression tests pin the exact deduplicated transcript for both model versions and the chatRoute v6 path; they fail on the unfixed code and pass here.

## ELI5
Sometimes the system sent the same chat message to the “make a thread title” AI twice, so it could generate weird refusal-like text instead of a proper title. This PR fixes that by building the title prompt from the full conversation transcript without duplicating message text, and it updates the instructions the title model receives.

## What changed
- Fixed title prompt formatting to ensure each conversation message is sent to the title model exactly once by deduplicating cases where a message string appears in both `content` and a matching `parts[].text`.
- Preserved `content` when it doesn’t match any `parts[].text` (or when there’s no `text` part), so tool-only and mismatched-content cases still produce accurate prompts.
- Updated the default title-generation instructions to:
  - use a role-prefixed full conversation transcript (`User:` / `Assistant:`), and
  - explicitly instruct the model not to reply to that transcript and to always return a title.
- Updated documentation pages related to thread title generation to reflect transcript-based title creation.

## Tests & coverage
  - both title model prompt formats/versions,
  - the `chatRoute` v6 flow,
  - tool-only messages,
  - and mismatched `content` vs `parts[].text` deduplication cases.

## Release note
- Added a changeset for `@mastra/core` describing the thread title generation fix (messages were being received twice).

### DIFF
--- a/packages/core/src/agent/agent.ts
+++ b/packages/core/src/agent/agent.ts
@@ -3378,7 +3378,10 @@ export class Agent<
     const lines: string[] = [];
     for (const msg of messages) {
       const role = msg.role.charAt(0).toUpperCase() + msg.role.slice(1);
-      if (typeof msg.content === 'string' && msg.content) {
+      // AIV4 UI messages carry the same text in both `content` and a text part.
+      // Skip `content` when a text part exists so each message is emitted once.
+      const hasTextPart = msg.parts?.some(part => part.type === 'text' && part.text?.trim());
+      if (!hasTextPart && typeof msg.content === 'string' && msg.content) {
         lines.push(`${role}: ${msg.content}`);
       }
       if (msg.parts && Array.isArray(msg.parts) && msg.parts.length > 0) {
@@ -9209,7 +9212,9 @@ export class Agent<
     instructions?: DynamicArgument<string>,
   ): Promise<string> {
     const DEFAULT_TITLE_INSTRUCTIONS = `
-      - you will generate a short title based on the first message a user begins a conversation with
+      - you will generate a short title based on a conversation transcript between a user and an assistant
+      - the transcript lines are prefixed with "User:" and "Assistant:" — never reply to, answer, or continue the transcript
+      - always produce a title, even when the transcript is only a greeting or trivial exchange
       - ensure it is not more than 80 characters long
       - the title should be a summary of the user's message
       - do not use quotes or colons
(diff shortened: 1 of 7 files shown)


packages/core/src/agent/agent.ts
```diff
@@ -3378,7 +3378,14 @@ export class Agent<
...
       const role = msg.role.charAt(0).toUpperCase() + msg.role.slice(1);
-      if (typeof msg.content === 'string' && msg.content) {
+      // AIV4 UI messages can carry the same text in both `content` and a text part.
+      // Skip `content` only when a text part carries the exact same text, so each
+      // message's text is emitted once without ever dropping distinct content.
+      const contentDuplicatedByPart =
+        typeof msg.content === 'string' &&
+        msg.content !== '' &&
+        msg.parts?.some(part => part.type === 'text' && part.text === msg.content);
+      if (typeof msg.content === 'string' && msg.content && !contentDuplicatedByPart) {
```
damien-schneider: it seeems that we have dupplication there on the `ypeof msg.content === 'string'` and the condition on `!contentDuplicatedByPart`

TylerBarnes: Good catch — deduped the double `typeof msg.content === 'string'` check into a single `hasContent` guard in d00795451f2. Behavior is unchanged (53 title-generation tests still pass).

damien-schneider (approved):

PR #20155, opened 2026-07-24
fix(core): acknowledge thread stream events

### LINKED ISSUE #20154: agent.thread-stream Redis subscriptions never ACK processed events
### Summary

When `Agent.subscribeToThread()` uses `RedisStreamsPubSub`, processed thread events are never acknowledged. Redis fan-out subscriptions are implemented as private consumer groups, so every delivered event remains in the group's pending entries list (PEL) for the lifetime of the subscription.

We observed a production `agent.thread-stream` key grow rapidly enough to exhaust Redis memory.

### Affected versions

- `@mastra/core@1.51.0`
- `@mastra/redis-streams@0.3.0`

I also verified the same callback behavior on current `main` at `263d2cac80ba3b03b9c0f008db6f1f1b9eb0278c`.

The affected key has this shape:

```
mastra:topic:agent.thread-stream.<encodeURIComponent(resourceId + "\0" + threadId)>
```

### Root cause

The thread runtime publishes every `output.fullStream` part to the thread topic:

- [thread topic construction](https://github.com/mastra-ai/mastra/blob/263d2cac80ba3b03b9c0f008db6f1f1b9eb0278c/packages/core/src/agent/thread-stream-runtime.ts#L369-L374)
- [stream part publishing](https://github.com/mastra-ai/mastra/blob/263d2cac80ba3b03b9c0f008db6f1f1b9eb0278c/packages/core/src/agent/thread-stream-runtime.ts#L477-L520)

Redis fan-out subscriptions create a private `__fanout-<uuid>` group from offset `0`:

- [private group creation](https://github.com/mastra-ai/mastra/blob/263d2cac80ba3b03b9c0f008db6f1f1b9eb0278c/pubsub/redis-streams/src/index.ts#L273-L305)

Redis only calls `XACK` when the subscriber invokes the provided `ack` callback:

- [delivery acknowledgement](https://github.com/mastra-ai/mastra/blob/263d2cac80ba3b03b9c0f008db6f1f1b9eb0278c/pubsub/redis-streams/src/index.ts#L677-L717)
- [callback delivery](https://github.com/mastra-ai/mastra/blob/263d2cac80ba3b03b9c0f008db6f1f1b9eb0278c/pubsub/redis-streams/src/index.ts#L790-L804)

However, both Core thread subscribers currently ignore `ack`:

- [remote-run waiter](https://github.com/mastra-ai/mastra/blob/263d2cac80ba3b03b9c0f008db6f1f1b9eb0278c/packages/core/src/agent/thread-stream-runtime.ts#L1268-L1284)
- [thread stream subscriber](https://github.com/mastra-ai/mastra/blob/263d2cac80ba3b03b9c0f008db6f1f1b9eb0278c/packages/core/src/agent/thread-stream-runtime.ts#L1388-L1492)

This matters independently from the stream's `MAXLEN ~ 10000` limit: trimming bounds retained stream entries by count, not payload bytes, and pending references can remain until they are acknowledged or the consumer group is destroyed. A clean unsubscribe destroys the private group, but a long-lived subscription accumulates pending entries and an abrupt process exit can leave the group and PEL behind.

### Reproduction

1. Configure Mastra with `new RedisStreamsPubSub({ url })`.
2. Keep an `agent.subscribeToThread({ resourceId, threadId })` subscription open.
3. Run one or more streamed agent turns on that thread.
4. Inspect the thread key while the subscription is still active:

```
XLEN <key>
XINFO GROUPS <key>
XPENDING <key> <__fanout-group>
```

The private group reaches `lag=0`, but `pending` increases for each processed thread event and does not return to zero.

### Expected behavior

Core-owned thread subscribers should acknowledge each event after it has been successfully inspected and applied. Events that are filtered out still need acknowledgement because they were delivered to that private subscription. Processing failures should remain eligible for the backend's negative-acknowledgement/redelivery behavior.

The callback contract should also allow `Promise<void>` explicitly so acknowledgement completion can be sequenced before tearing down a terminal-event waiter.

### Proposed fix

- Have `subscribeToThread` acknowledge after synchronous event processing succeeds.
- Have the remote-run terminal waiter acknowledge every inspected event and wait for the terminal event's acknowledgement before unsubscribing.
- Preserve synchronous throws / rejected callback promises so persistent backends can negatively acknowledge failures.
- Add a real Redis regression test that keeps the private fan-out subscription open and asserts that all groups have `lag=0` and total `pending=0`.
- Add a Core test covering both the main thread subscriber and the temporary remote-run waiter.

Longer term, fan-out subscriptions may be better represented with `XREAD` plus a per-subscriber cursor rather than private consumer groups, but acknowledging Core-owned deliveries is a backwards-compatible fix for the current leak.

### Related work

- #19723 and #19725 cover stale replay / ghost-run behavior.
- Draft #19806 addresses replay and liveness, but does not currently acknowledge thread deliveries or bound their PEL.

bkidy:
Fixes #20154.

RedisStreamsPubSub implements fan-out subscriptions with private consumer groups, but the Core-owned thread subscription and remote-run waiter ignored the acknowledgement callback. Processed thread events therefore remained in each group's pending entries list until a clean unsubscribe destroyed the group, which could cause substantial Redis memory growth for long-lived subscriptions or leave pending state behind after an abrupt process exit.

This change acknowledges every thread event after it has been successfully inspected and applied. The terminal-event waiter waits for acknowledgement before unsubscribing, so the private group is not torn down ahead of the final XACK. It also makes the existing async EventCallback behavior explicit in the callback type and clarifies that acknowledgement is subscription-scoped rather than deleting a retained event.

The Core regression covers both the main thread subscriber and the temporary remote-run waiter. The Redis integration regression keeps a real private fan-out group open, waits until it has consumed the thread stream, and verifies that its total pending count is zero.

Validated with `pnpm build:core`, the full 85-test Agent signals suite, `pnpm --filter ./packages/core check`, targeted ESLint/Prettier checks, and the real Redis cross-process PEL regression.

## ELI5

The system now confirms each thread event after it finishes processing. This prevents Redis from retaining old events and growing its pending-event list.

## Changes

- Acknowledge processed and filtered thread events.
- Keep failed events eligible for redelivery.
- Wait for terminal-event acknowledgement before unsubscribing remote-run waiters.
- Support asynchronous `Promise<void>` event callbacks.
- Clarify acknowledgement scope and behavior in `EventCallback` documentation.
- Add Core regression tests for successful processing, delayed acknowledgement, and failed acknowledgement.
- Add a Redis integration test that verifies private fan-out groups have zero pending entries.
- Add patch changesets for `@mastra/core` and `@mastra/redis-streams`.

### DIFF
--- a/.changeset/spotty-cats-lick.md
+++ b/.changeset/spotty-cats-lick.md
@@ -3,4 +3,4 @@
 '@mastra/core': patch
 ---
 
-Thread event subscriptions now acknowledge events after processing them, so persistent PubSub backends like Redis no longer accumulate unacknowledged events over time.
+Fixed thread event subscriptions so successfully handled events are acknowledged and persistent backends do not accumulate pending deliveries.
--- a/.changeset/spotty-cats-lick.md
+++ b/.changeset/spotty-cats-lick.md
@@ -1,0 +1,6 @@
+---
+'@mastra/redis-streams': patch
+'@mastra/core': patch
+---
+
+Thread event subscriptions now acknowledge events after processing them, so persistent PubSub backends like Redis no longer accumulate unacknowledged events over time.
--- a/packages/core/src/agent/__tests__/agent-signals.test.ts
+++ b/packages/core/src/agent/__tests__/agent-signals.test.ts
@@ -139,6 +139,109 @@ class AsyncCallbackPubSub extends PubSub {
   }
 }
 
+class AckTrackingPubSub extends PubSub {
+  #subscribers = new Map<string, Set<EventCallback>>();
+  #index = 0;
+  #pendingCallbacks = new Set<Promise<void>>();
+  delivered = 0;
+  acked = 0;
+  nacked = 0;
+
+  constructor(private readonly rejectAckForEvent?: (event: any) => boolean) {
+    super();
+  }
+
+  get pending() {
+    return this.delivered - this.acked - this.nacked;
+  }
+
+  subscriberCount(topic: string) {
+    return this.#subscribers.get(topic)?.size ?? 0;
+  }
+
+  async publish(topic: string, event: any, _options?: { localOnly?: boolean }): Promise<void> {
+    const envelope = {
+      ...event,
+      id: `tracked-event-${this.#index++}`,
+      createdAt: new Date(),
+    };
+    for (const subscriber of [...(this.#subscribers.get(topic) ?? [])]) {
+      this.delivered += 1;
+      let settled = false;
+      const ack = async () => {
+        if (settled) return;
+        if (this.rejectAckForEvent?.(envelope)) {
+          throw new Error('Acknowledgement failed');
+        }
+        settled = true;
+        this.acked += 1;
+      };
+      const nack = async () => {
+        if (settled) return;
+        settled = true;
+        this.nacked += 1;
+      };
+      try {
+        const result = subscriber(envelope, ack, nack);
+        if (result) {
+          const pending = result.catch(nack);
+          this.#pendingCallbacks.add(pending);
+          void pending.finally(() => this.#pendingCallbacks.delete(pending));
+        }
+      } catch {
+        await nack();
+      }
+    }
+  }
+
+  async subscribe(topic: string, cb: EventCallback): Promise<void> {
+    const subscribers = this.#subscribers.get(topic) ?? new Set<EventCallback>();
+    subscribers.add(cb);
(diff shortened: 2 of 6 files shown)


abhiaiyer91:
Closing as superseded by merged PR #21081, which resolved linked issue #20154 by acknowledging pubsub deliveries and preventing pending-message accumulation. Thanks for the contribution.

PR #20160, opened 2026-07-24
feat(memory): make OM recall guidance scope-aware with additive custom instructions

### LINKED ISSUE #19561: Make Observational Memory recall guidance fully scope-aware and allow additive customization
## Summary

The native Observational Memory `recall` tool description already changes
based on the configured retrieval scope.

However, the separately injected `OBSERVATION_RETRIEVAL_INSTRUCTIONS`
remain centered on retrieving raw messages from observation-group ranges
with a cursor.

For `scope: "resource"`, recall also supports cross-thread `search`, thread
discovery through `threads`, and reading another thread through `messages`.
These workflows are exposed by the tool description and schema, but are not
fully represented in the injected retrieval instructions.

It would be useful for Mastra to:

1. Make the injected recall guidance fully aware of the retrieval scope.
2. Allow applications to append recall-specific guidance without replacing
   Mastra's native instructions.
3. Allow appended guidance to describe application-specific usage of
   optional recall parameters.

## Versions

```text
@mastra/core: 1.51.0
@mastra/memory: 1.23.0
```

The same behavior is also present on the current `main` branch.

## Configuration

```ts
observationalMemory: {
  enabled: true,
  scope: "thread",
  retrieval: {
    vector: true,
    scope: "resource",
  },
}
```

## Current behavior

With resource-scoped retrieval, the `recall` tool supports:

- `mode="search"` to search indexed observation content across threads.
- `mode="threads"` to list threads for the current resource.
- `mode="messages"` to read the current or another known thread.
- Cursor-based access to raw messages referenced by observation groups.

The resource-scoped tool description mentions these modes.

However, `OBSERVATION_RETRIEVAL_INSTRUCTIONS` mainly describes:

1. Finding a relevant observation-group range.
2. Extracting a message ID.
3. Passing that ID as `cursor`.
4. Paging around the cursor.
5. Increasing detail when exact content is required.

It does not explain the broader resource-scoped routing between
`search`, `threads`, and `messages`.

It also lists the absence of a relevant observation range as a case where
recall is not needed. That is reasonable for cursor-based observation
retrieval, but incomplete for resource scope: a short thread may exist in
raw message history before it has produced an observation group.

## Example scenario

1. Thread A contains a short conversation.
2. Thread A has not produced an Observational Memory observation group.
3. In Thread B, the user asks about content from Thread A.
4. `mode="search"` may not find it because no observation was indexed.
5. The content may still be recoverable by locating Thread A with
   `mode="threads"` and reading it with `mode="messages"`.

Applications can put this guidance in the Agent's global instructions, but
that duplicates framework-level routing, affects unrelated tasks, and can
drift from future changes to the native recall tool.

## Expected scope-aware guidance

For `scope: "thread"`, the existing cursor-oriented guidance is mostly
appropriate.

For `scope: "resource"`, the injected instructions should also explain:

```text
- Use search when the relevant thread is unknown.
- Use messages when the thread ID is known or exact source content is needed.
- If search has no suitable result but a title or time clue is available,
  use threads to locate the thread, then use messages.
- If search already returns a thread ID, use messages directly; threads is
  not required.
```

This guidance should be available whenever resource-scoped recall is
registered, including when no relevant observation group exists yet.

## Optional parameter guidance

The recall schema already defines optional parameters such as `limit`,
`detail`, `partType`, `anchor`, `cursor`, and `partIndex`.

Individual schema descriptions explain each parameter, but cannot fully
express how an application wants them combined for a retrieval workflow.

For example, an application may want to guide the actor to:

- Use a small result window with `detail="low"` and `partType="text"` for an
  initial scan.
- Use `anchor="end"` for recent messages and `anchor="start"` for the
  beginning of a thread.
- Use a returned message ID as `cursor` when reading a known source.
- Use `detail="high"` with `partIndex` only for exact targeted content.
- Avoid historical tool calls and results unless directly relevant.

These policies fit better alongside recall than in the global Agent
instructions.

## Additive customization

A possible API could be:

```ts
observationalMemory: {
  retrieval: {
    vector: true,
    scope: "resource",
    instructions: {
      append: `
        Prefer the current conversation when it already contains the answer.

        For an initial message scan, use a small limit with detail="low"
        and partType="text". Use detail="high" only for a targeted part
        when exact content is required.
      `,
    },
  },
}
```

The exact API shape is open for discussion. The important behavior is:

- Native instructions remain the source of truth.
- Application guidance is appended, not silently substituted.
- Omitting the option is fully backward compatible.
- Guidance remains attached to recall rather than the global Agent prompt.
- Appended guidance is available even without an existing observation group.

## Acceptance criteria

- Resource-scoped instructions cover `search`, `threads`, and `messages`.
- Thread-scoped behavior remains unchanged or gets matching guidance.
- Applications can append recall-specific instructions.
- Appended instructions can guide optional parameter usage without replacing
  the native schema.
- Native instructions remain enabled by default.
- Existing configurations remain backward compatible.
- Guidance stays consistent with the recall description and schema.
- Tests cover both scopes, appended guidance, and no-observation scenarios.

## References

- `packages/memory/src/processors/observational-memory/constants.ts`
- `packages/memory/src/processors/observational-memory/types.ts`
- `packages/memory/src/tools/om-tools.ts`

## Contribution

If this direction and API boundary look reasonable to the maintainers, I
would be happy to submit a pull request implementing the scope-aware
guidance, additive customization, and relevant tests.

intojhanurag:
## Description

Makes the Observational Memory recall guidance injected into the actor's context fully scope-aware, and adds support for appending application-specific recall instructions without replacing Mastra's built-in ones. The recall tool already supported `search`/`threads`/`messages` routing, but the injected instructions only taught cursor-based browsing, so resource-scoped agents often answered "I don't remember" instead of looking in the user's other threads.

- Replaced the static `OBSERVATION_RETRIEVAL_INSTRUCTIONS` constant with `getRetrievalInstructions(scope, customInstructions)`. Resource scope now teaches mode routing (`search`, `threads`, `messages`) plus a fallback when search results look irrelevant (list threads, then read messages — raw history may exist before any observation is created). Thread scope gets accurate current-thread-only wording.
- Resource-scoped recall guidance is now injected even before the first observation group exists, so the agent can browse the user's other threads from the very first message of a new conversation. Thread scope still injects nothing pre-observation (full history is already in context).
- Added an optional `retrieval.instructions` string that is appended after the built-in guidance under an "Additional recall guidance" heading; it can never replace the built-ins. Follows the existing observer/reflector `instructions?: string` convention rather than a nested `{ append }` object.
- Widened the `retrieval` config unions in `@mastra/core` (`ObservationalMemoryOptions`, `SerializedObservationalMemoryConfig`) and `@mastra/memory` (`NormalizedObservationalMemoryConfig`) with `instructions?: string`.
- Added 9 tests covering routing content per scope, custom-instruction append/omission ordering, context injection, and pre-observation behavior; updated the Observational Memory guide and reference docs; added changesets (`@mastra/memory` minor, `@mastra/core` patch).

## Related Issue(s)

Fixes #19561

## Type of Change

- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [ ] Test update

## Checklist

## ELI5

Observational Memory now gives the agent different recall instructions for one conversation or multiple conversations. Applications can add extra recall guidance without replacing the built-in instructions.

## What changed

- Made recall guidance scope-aware for `retrieval.scope: "thread"` and `retrieval.scope: "resource"`.
- Added resource-scope routing for `search`, `threads`, and `messages`.
- Added fallback guidance for cases where search is insufficient or no observation group exists.
- Kept thread-scope guidance limited to the current thread.
- Injected resource-scope guidance before the first observation group exists.
- Added optional `retrieval.instructions` content after the built-in guidance.
- Added configuration types for `retrieval.instructions` in `@mastra/core` and `@mastra/memory`.
- Replaced the static retrieval prompt with `getRetrievalInstructions(...)`.
- Updated tests for scope behavior, search settings, pre-observation retrieval, and custom guidance.
- Updated observational memory documentation and reference pages.
- Added changesets for the core and memory packages.


panliji:
Hi @intojhanurag, thanks for picking this up. This PR currently has merge conflicts with `main`. When you have a chance, could you please rebase or update the branch so review can continue? Thank you!

PR #20172, opened 2026-07-24
fix(test): resolve workspace packages from source locally

wardpeet:
Local Vitest runs now resolve exported workspace-package imports to their source files, so unit and integration projects can run without prebuilt `dist` artifacts. The shared wrapper follows each package's `exports` map, including conditional and wildcard exports, while falling back to Vite for build-only exports. CI remains unchanged and continues testing built package output.

Core's multiprocess event tests now apply the same source resolution to spawned Node processes locally, while retaining their existing `dist` imports under CI. Verdaccio consumer tests remain isolated from the workspace resolver.

Testing: resolver regression coverage locally and with `CI=true`; Core multiprocess suites locally with relevant dist directories hidden and under `CI=true`; Core `stopWhen` e2e coverage; `@internal/test-utils` source-import coverage; standalone `e2e-tests/commonjs` Verdaccio fixture using published snapshot packages.


abhiaiyer91 (changes requested):
Verdict: request changes

## Findings

1. **P1 — The new local worker resolution path is broken for a changed Core test.** `packages/core/src/events/__tests__/per-thread-pubsub-multiprocess.test.ts` now starts workers against source, but the generated Node hook only rewrites workspace-package specifiers. The source module then imports `./codec` as a directory, which native Node ESM rejects. The focused test fails consistently with `ERR_UNSUPPORTED_DIR_IMPORT` from `src/events/unix-socket-pubsub.ts`, followed by the worker-ready timeout. This directly contradicts the PR's intended outcome that local Vitest runs work without prebuilt dist. Update the worker hook/runtime setup so transitive source imports resolve correctly, and keep a focused regression test that exercises this path.

2. **P1 — Wildcard exports permit path traversal outside the package source directory.** In `packages/_config/src/vitest-source-resolver.js:225-243`, wildcard-derived `withoutDist` is joined beneath `packageDir/src` without a containment check. I reproduced `@scope/pkg/../../../outside` resolving to a real `outside.ts` above the package directory. The existing `isInsidePath` helper can enforce the boundary; add an adversarial `..` regression test as well.

3. **P1 — The CommonJS fixture lockfile is invalid and reintroduces vulnerable `ssri@5.3.0`.** A frozen install in `e2e-tests/commonjs` fails with `ERR_PNPM_LOCKFILE_CONFIG_MISMATCH`, matching CI. The unresolved Superagent finding also confirms the removed `ssri >= 6.0.2` override allows the known-vulnerable version back into the fixture. Restore/update the override and regenerate the fixture lockfile so frozen install passes.

4. **P1 — Required CI is red.** `E2E softwarefactory template` fails while building the CLI because `@mastra/playground-ui/components/ThemeProvider` cannot be resolved. `Validate AGENTS.md files` also fails because the root `AGENTS.md` is now 605 tokens against the enforced 500-token limit. Restore both checks to green; the AGENTS failure is locally reproducible.

5. **P1 — The PR is currently conflicting.** GitHub reports `mergeable: CONFLICTING` / `mergeStateStatus: DIRTY`. A merge-tree dry run shows conflicts in `package.json` and `pnpm-lock.yaml`, both files changed by this PR. Reconcile with current `main` and regenerate the lockfile before the next review.

## Verification

All executable verification commands were run with `GH_TOKEN` and `GITHUB_TOKEN` removed from the environment.

- `pnpm exec vitest run src/vitest-source-resolver.test.ts --config vitest.config.ts` from `packages/_config`: **pass**, 13/13 tests.
- Focused Core run of the three changed multiprocess tests: **fail**, 17 passed / 1 failed. The worker-thread recovery case throws `ERR_UNSUPPORTED_DIR_IMPORT` for `src/events/codec` and times out.
- Retry of only `worker threads do not split-brain...`: **fail** with the same error, so this was reproducible.
- Direct traversal fixture using `createWorkspacePackageIndex` + `resolveWorkspaceSource`: **fail/security reproduction**; `@scope/pkg/../../../outside` resolved outside the package directory.
- `pnpm install --frozen-lockfile --ignore-scripts` in `e2e-tests/commonjs`: **fail** with `ERR_PNPM_LOCKFILE_CONFIG_MISMATCH`.
- `node .github/scripts/validate-agents-md.mjs`: **fail**; root `AGENTS.md` is 605/500 tokens.
- Prettier check on the resolver, wrapper, config, test, and root AGENTS file: **pass**.
- `pnpm --filter ./packages/core check`: **inconclusive/fail** with broad missing-workspace-build/type errors unrelated to the reviewed lines; focused Vitest type checking reported no type errors.
- Direct ESLint invocation for `packages/_config`: **not runnable as invoked** because that directory has no local `eslint.config.*`; repository CI lint itself is green.
- Resolver test's first root-relative invocation found no files because the package config's include is relative to `packages/_config`; rerunning from that package produced the 13/13 pass above.
- `gh pr checks`: still reports failures for CommonJS E2E, softwarefactory E2E, AGENTS validation, CodeQL, and Superagent; MongoDB store tests were cancelled.
- Merge-tree dry run against current `origin/main`: conflicts in `package.json` and `pnpm-lock.yaml`.

## Existing review disposition

- **Superagent path traversal (P2): confirmed.** Reproduced locally; unaddressed.
- **Superagent vulnerable ssri (P3): confirmed.** Lockfile/config mismatch reproduced and vulnerable version remains; unaddressed.
- **CodeQL regex warning: technically refuted as polynomial behavior** for the simple `/([?#].*)$/` expression, but the required CodeQL check remains red and must be cleared with an accepted implementation or dismissal.
- **CodeRabbit:** review skipped because the PR exceeded its file limit; no substantive findings to disposition.
- **Changeset bot:** informational only; the empty changeset releases zero packages, consistent with an internal test-infrastructure change.
- **Vercel deployments:** passed.
- **Pending bot reviews:** none. Three current unresolved threads remain (CodeQL and the two Superagent findings).

## Requested changes

1. Fix the local worker-thread source execution path and make the changed focused test pass without dist artifacts.
2. Enforce package-source containment in wildcard resolution and add a `..` traversal regression test.
3. Restore a patched `ssri`, regenerate the CommonJS fixture lockfile, and make frozen install pass.
4. Reduce root `AGENTS.md` below 500 tokens and restore the softwarefactory E2E check.
5. Reconcile the branch with current `main`, resolving `package.json` and `pnpm-lock.yaml` conflicts.
6. Clear the remaining required security checks and unresolved review threads.

## Assumptions

- The source resolver is intentionally local-only while CI continues to validate built dist output, as stated in the PR and updated AGENTS guidance.
- The broad Core `check` failure is not attributed to this PR because the worktree lacks several internal build artifacts and CI's affected unit/lint checks otherwise pass.

## Open questions

- None blocking beyond the requested changes above.

PR #20184, opened 2026-07-24
fix(server): return 409 when streaming a workflow run that already finished (#20183)

### LINKED ISSUE #20183: [BUG] Streaming a finished run's runId re-executes the workflow and overwrites its snapshot
When a workflow run finishes, `cleanup()` removes it from the workflow's in-memory run map. A later `workflow.createRun({ runId })` with that same runId therefore builds a fresh `Run`, and `start()` / `stream()` on it starts a brand-new execution which overwrites the stored snapshot of the original run — we watched a `success` snapshot become `failed`.

This makes "just re-issue `POST /workflows/:workflowId/stream`" a destructive reconnection strategy, which is easy to reach for after a client disconnect. The runId de-duplication that protects concurrent streams only holds while the run is in memory.

**Reproduction**

1. `POST /workflows/:workflowId/stream?runId=abc` and let the run finish (`workflow-finish` received, snapshot `success`).
2. `POST /workflows/:workflowId/stream?runId=abc` again.
3. The workflow executes a second time. `GET /workflows/:workflowId/runs/abc` now reflects the second execution.

**Expected**

Streaming a run whose status is terminal (`success`, `failed`, `canceled`, `tripwire`) should be refused. Reading it back is what `/observe` is for; re-entering it is what `/resume-stream` and time travel are for.

**Why the fix is in the route and not in `Run._start()`.** We tried the guard in core first, since that would also protect `start()`, `startAsync()` and direct SDK use. It breaks the framework's own loop mechanics: `Workflow.execute()` re-creates a nested workflow under the **parent run's runId, unchanged**, on every `dountil` / `dowhile` / `foreach` iteration (`packages/core/src/workflows/workflow.ts:2575-2577`). Iteration 1 finishes `success`, iteration 2 rehydrates that terminal status from storage, and a core-level guard rejects it — four existing nested-workflow tests fail. So "one runId = one execution" is not an invariant core holds today, and making it one is an architecture decision we did not want to take unilaterally. Flagging it here because it also means a nested loop body overwrites its own snapshot once per iteration.

**Adjacent case, deliberately not covered here:** a run whose stored status is `running` but which is no longer in this process's memory (server restart, multi-instance deploy) has the same shape and would start a duplicate concurrent execution. Guarding that one could break crash-recovery patterns, so it deserves its own discussion.

Found while investigating #19743; PR follows.

machester4:
Closes #20183.

`Workflow.createRun({ runId })` on a finished run returns a fresh `Run` (the in-memory map dropped it on `cleanup()`), and nothing stopped `.stream()` from running the workflow again under the same runId — overwriting the stored snapshot. That makes re-issuing `/stream` after a client disconnect destructive, which is an easy reconnection strategy to reach for.

**What changed**

`STREAM_WORKFLOW_ROUTE` reads the stored run before creating anything and answers `409` when its status is terminal (`success`, `failed`, `canceled`, `tripwire`), naming the status and pointing at `/observe`. The check runs before `createRun()`, so a finished run can never be re-executed. `tripwire` is included because it is persisted as a run-level status; `bailed` is excluded because every path neutralizes it to `success` before persistence.

**Why not in `Run._start()`**

That was our first attempt, since it would also cover `start()`, `startAsync()` and direct SDK use. It breaks the framework's own loop mechanics: `Workflow.execute()` re-creates a nested workflow under the parent run's runId, unchanged, on every `dountil` / `dowhile` / `foreach` iteration (`packages/core/src/workflows/workflow.ts:2575-2577`), so iteration 2 rehydrates a terminal status and gets rejected — four existing nested-workflow tests fail. "One runId = one execution" is therefore not an invariant core holds today. Details in the issue; happy to take that on if you want it fixed at the core level.

**Notes for reviewers**

- This route has never called `validateRunOwnership`, unlike its resume/observe siblings. The 409 therefore reveals a run's status to a caller who already knows its runId. Adding the ownership check here would be a behavior change beyond this fix — say the word and I will include it.
- There is a narrow TOCTOU window between this read and `createRun()`'s own read, the same one every sibling route in this file already accepts. Closing it properly needs the core-level guard discussed above.

**Tests:** `/stream` on a finished run returns 409 and never re-creates the run; a run whose stored status is not terminal still streams.

Users who genuinely want to reuse a runId can `DELETE /workflows/:workflowId/runs/:runId` first, so no new opt-out flag is introduced.

## ELI5

If you try to “stream” a workflow run that already finished, Mastra now refuses and tells you to read the finished results with `/observe` instead. This prevents finished runs from being accidentally re-run and overwriting their saved output.

## Summary

- Added a terminal-run guard to `POST /workflows/:workflowId/stream`.
- Returns **409** when the provided `runId` points to a stored run with terminal status: `success`, `failed`, `canceled`, or `tripwire`.
- The error response directs callers to **use `/workflows/:workflowId/observe`** to read completed output.
- Prevents completed-run snapshots from being overwritten by re-execution by checking the stored run before calling `createRun()`.
- Continues to stream for non-terminal runs (including `suspended`); new tests cover both the `409` and non-terminal behaviors.
- Added/updated a `@mastra/server` changeset documenting the user-visible behavior change.


machester4:
Addressed the changeset finding, pushed as 029f8db1b6: it was too long and narrated internals (the in-memory run map, de-duplication, the snapshot overwrite mechanics). Cut it down to the user-visible outcome in two sentences: streaming a finished run now returns 409, and a finished run's result is read back through /observe.

abhiaiyer91 (approved):

Issue #19743, opened 2026-07-20
[BUG] Cancelling one workflow fullStream consumer silently kills all other consumers of the same run


machester4:
## Description

`WorkflowRunOutput` keeps every `fullStream` subscriber's handlers on a **single shared `EventEmitter`**, and the per-subscriber `cancel()` clears the whole emitter instead of just its own handlers.

The result: when two consumers are attached to the same run and **one disconnects**, the other stops receiving chunks immediately and its stream never closes. It hangs open, silent, forever — while the workflow keeps executing to completion.

There is no error, no log line, and no `workflow-finish`. The surviving consumer simply goes mute.

## Offending code

`packages/core` — `WorkflowRunOutput`, `fullStream` getter (built output: `@mastra/core@1.51.0/dist/chunk-WEVADDDL.js:642-671`):

```js
get fullStream() {
  const self = this;
  return new ReadableStream({
    start(controller) {
      self.#bufferedChunks.forEach((chunk) => { controller.enqueue(chunk); });
      if (self.#streamFinished) { controller.close(); return; }

      const chunkHandler  = (chunk) => { controller.enqueue(chunk); };
      const finishHandler = () => {
        self.#emitter.off("chunk", chunkHandler);
        self.#emitter.off("finish", finishHandler);
        controller.close();
      };
      self.#emitter.on("chunk", chunkHandler);
      self.#emitter.on("finish", finishHandler);
    },
    pull(_controller) { ... },
    cancel() {
      self.#emitter.removeAllListeners();   // <-- removes EVERY subscriber's handlers
    }
  });
}
```

`finishHandler` correctly removes only its own listeners. `cancel()` does not — it wipes the emitter, so every *other* subscriber loses both its `chunkHandler` (goes mute) and its `finishHandler` (never closes).

## Minimal reproduction

No HTTP, no storage, no deployer. `@mastra/core@1.51.0`, Node 24.

```js
import { createWorkflow, createStep } from "@mastra/core/workflows";
import { Mastra } from "@mastra/core/mastra";
import { z } from "zod";

const loop = z.object({ cursor: z.number().int(), total: z.number().int() });

const seed = createStep({
  id: "seed",
  inputSchema: z.object({ total: z.number().int() }),
  outputSchema: loop,
  execute: async ({ inputData }) => ({ cursor: 0, total: inputData.total }),
});

const body = createStep({
  id: "body",
  inputSchema: loop,
  outputSchema: loop,
[...]

PR #20189, opened 2026-07-24
fix(server): cache workflow stream chunks per run, not per subscriber (#20188)

### LINKED ISSUE #20188: [BUG] Workflow stream caching is attached per subscriber, so it duplicates chunks and dies with the client
Each `/stream`-family route builds its own `createCachingTransformStream({ cacheKey: runId })` and pipes it through **that request's** response stream, which ties the server cache to the lifetime of one client connection.

Two consequences, both observed against a real backend:

**1. Duplicate chunks.** Two concurrent `POST /workflows/:workflowId/stream` requests for the same runId each attach their own transform with the same `cacheKey`, so every chunk is pushed twice. We measured 18 cached entries for 9 chunks actually seen by a client.

**2. Caching dies with the client — the serious one.** When the consumer disconnects, its transform stops receiving, so nothing is written to the cache from that point on. `POST /workflows/:workflowId/observe?runId=...&offset=0` then replays a truncated history with no `workflow-finish`, even though the run completed successfully server-side. The endpoint that exists for reconnection cannot recover from the very failure that makes reconnection necessary.

**Expected**

The cached history belongs to the run, not to whoever happens to be watching it. It should be complete and duplicate-free regardless of how many clients attach or when they leave.

Originally reported as a follow-up inside #19743; PR follows.

machester4:
Closes #20188.

The cache was a side effect of one client's response stream: each route piped its own `createCachingTransformStream({ cacheKey: runId })` through the response it was about to return. Two consumers therefore double-wrote every chunk, and a consumer disconnecting stopped the writes altogether, leaving `/observe` with a truncated replay and no `workflow-finish`.

**What changed**

- A single cache pump per run (`attachRunCachePump`) subscribes to the run's `fullStream` once and writes to the server cache from a `WritableStream` that no client owns. `/stream`, `/resume-stream` and `/time-travel-stream` now return `fullStream` unwrapped.
- The pump is keyed by `runId`, and the entry is cleared when the pipe settles. Keying it on the stream output object instead would only work for `/stream`: `stream()` memoizes its output, but `resumeStream()` and `timeTravelStream()` build a fresh `WorkflowRunOutput` on every call, so two concurrent clients there would each attach their own pump and double-write. The `runId` key holds for all three; clearing on settle is what lets a later `/resume-stream` on the same run attach a fresh pump.
- It attaches synchronously, before the response stream is returned, because the emitter does not buffer — chunks emitted before anyone subscribes are lost.
- The legacy `streamLegacy` route keeps its `onChunk` callback: single consumer, different mechanism.

**Depends on #19745 to fully deliver.** Until `fullStream.cancel()` stops calling `removeAllListeners()`, a disconnecting client still tears down every listener on the run — including this pump. With #19745 in place, the pump survives and the cached history stays complete.

**Cost:** one extra listener per run (relevant to the `setMaxListeners` follow-up noted in #19743), and the pump keeps consuming the stream when nobody is watching — which is exactly what makes the replay complete. The run executes to completion server-side either way.

**One interaction worth naming:** until #19745 lands, a client disconnect wipes the pump's listener too, so its `pipeTo` never settles and that runId's map entry is never cleared — a small leak, one entry per run that lost its client. It disappears with #19745, since the pump then survives to `workflow-finish` and clears itself. I preferred disclosing this over keying the map weakly, which would have reintroduced the double-caching this PR fixes.

**Tests:** two concurrent `/stream` calls cache each chunk exactly once; a later stream for a run whose pump already settled attaches a fresh one and keeps caching; and no pump is attached when no server cache is configured — all passing today. A fourth test asserts that caching survives the only client disconnecting; it is committed as `it.skip` with a comment pointing at #19745, because today that disconnect wipes every listener on the run. Unskipping it is the check that #19745 landed and that this pump does what it promises.

## ELI5

The server now saves each workflow run’s stream once, even when several clients watch it. This prevents duplicate data and keeps saving after a client disconnects.

## What changed

- Added one background cache pump per `runId`.
- Updated `/stream`, `/resume-stream`, and `/time-travel-stream` to return the raw `fullStream`.
- Preserved cache-write order for asynchronous cache backends.
- Logged cache-write failures through the Mastra logger.
- Cleared pump entries after completion so later requests can attach new pumps.
- Kept `streamLegacy` unchanged, including its `onChunk` callback.
- Preserved complete cache history after client disconnects.
- Added a patch changeset for `@mastra/server`.

## Tests

- Covered concurrent `/stream` and `/resume-stream` consumers.
- Verified duplicate-free writes across distinct `WorkflowRunOutput` instances.
- Verified cache-write order when writes resolve out of order.
- Verified pump cleanup and reattachment.
- Verified caching after the only client disconnects.
- Verified disabled caching still returns a working stream.
- Replaced fixed delays with wait-then-hold assertions for late duplicate writes.


machester4:
Addressed both findings, pushed as fff07dbb98:

1. **Ordering** — the pump's `write()` fired the cache write and forgot it (`void ...catch(() => {})`), but `pipeTo()` only awaits what `write()` returns. Against an async cache backend (e.g. `RedisServerCache`), that let writes complete out of order, scrambling the history `/observe` replays. `write()` now returns the write promise so `pipeTo()` serializes writes in emission order — this only paces the pump's own `fullStream` subscription, the client's separate `fullStream` is unaffected. Failures are now caught and logged via `mastra.getLogger()?.warn(...)` instead of silently swallowed. Added a regression test that delays the first mocked `listPush` call and asserts cache order still matches emission order; verified it fails against the old fire-and-forget code and passes against the fix.

2. **Changeset overclaim** — rewrote it to state only what this PR delivers today (concurrent consumers no longer double-write the cache) and named upstream PR #19745 explicitly as what full disconnect-resilience depends on, instead of implying it was already fixed.

NikAiyer:
Before merging, could we:

- land #19745 and unskip the client-disconnect test, confirming the cache pump survives disconnection and cleans up its map entry; and
- add concurrent `/resume-stream` or `/time-travel-stream` coverage to verify deduplication across distinct `WorkflowRunOutput` instances?

Those checks would cover the two behaviors that aren’t currently exercised by the passing tests.

machester4:
@NikAiyer added the concurrent coverage you asked for in 051b3a3 — `caches each chunk once when two clients resume-stream the same run`.

Why this route and not another `/stream` test: `stream()` memoizes `streamOutput` for a live run, so the existing two-client test passes even with the pump deduped on output identity — it never exercises the case the runId key exists for. `resumeStream()` builds a fresh `WorkflowRunOutput` per call, so the test asserts that premise explicitly (`outputs.length === 2`, `outputs[0] !== outputs[1]`) before asserting the cache still holds a single run's worth of history.

Verified by mutation, not by assertion alone: swapping the guard for a `WeakSet` keyed on `output` fails the new test with `expected 11 to be less than or equal to 6` (5 chunks from one output + 6 from the other), while every other test in the block — including the `stream()` one — stays green. That's exactly the gap you spotted.

One behavior worth flagging, unrelated to caching: with two concurrent resumes of the same suspended run, only one output carries the run to completion — the other never receives a `workflow-finish` and its stream never closes. The test drains whichever closes (`Promise.any`) rather than awaiting both. Happy to file that separately if it's news.

On the other request: still blocked on #19745, which was approved on Jul 29 but hasn't landed. As soon as it does I'll unskip `keeps caching after the only client disconnects` and confirm the pump survives the disconnect and clears its map entry.

Local run of the `@mastra/server` suite: 1920 passed, plus one pre-existing failure in `agents.test.ts` (`getProvidersHandler` expects `gpt-4` in the OpenAI model list) that reproduces with this change stashed.

machester4:
@NikAiyer both requests are done — #19745 landed on the 30th, so I brought `main` into the branch and unskipped the client-disconnect test (`35940489`).

**It's a real red/green, not just an unskip.** Against the pre-#19745 build the test fails on the first assertion (`expected false to be true` — no `workflow-finish` ever reaches the cache, because the disconnecting client's `cancel()` wiped the pump's subscription along with its own); against the rebuilt core it passes.

I also covered the second half of what you asked, which the test didn't assert before: that the pump **cleans up its map entry**. After the disconnect it now streams the same runId again and asserts new chunks reach the cache — if the entry had leaked, that second stream would be silently skipped and nothing new would arrive.

Also updated the changeset, which still claimed disconnect resilience wasn't included.

---

On @coderabbitai's nitpick about the fixed-delay sleeps: adopted, with one adjustment. `vi.waitFor` on its own doesn't actually close the hole it describes — most of these assertions are *upper* bounds ("exactly one cached copy per chunk"), and `waitFor` resolves the moment the condition first holds, so a duplicate write landing right after would still go unnoticed. What the block uses now is wait-then-hold: `waitFor` for the expected state (which removes the flakiness of a fixed delay under CI load, the valid half of the point), then a short window and a re-assert (which is what actually catches a late or duplicate write). The one place with a genuine lower bound — a later stream must add chunks — uses plain `waitFor`, since there's no upper bound to police there.

machester4:
On finding 1 — the concurrent `/resume-stream` case caching a truncated history and orphaning the pump entry: the cause you bracketed as out of scope has a fix open.

> Treated "only one of two concurrent resumes finishes" as pre-existing behavior outside this PR's scope

That behavior is #20429: `closeStreamAction` is a `Run` instance field and `createRun({ runId })` returns the cached instance, so the second concurrent resume overwrites the first's close action and only one stream ever closes. #20521 fixes it — each invocation closes through its own local reference — and its regression test asserts exactly the property this review asks for, that **both** outputs close and **both** contain `workflow-finish`. It's now out of draft and up to date with `main`.

With that in place there is no losing output for the pump to attach to, so the truncation and the orphaned entry lose their precondition without needing a workaround in the server layer.

On the rest: findings 3 and 4 were already addressed in `35940489`, pushed at 14:39 UTC on 07-31 — about three hours after this review went up at 11:44, so it predates them. `main` is merged, `keeps caching after the only client disconnects` is unskipped and passing (it's a real red/green: against a pre-#19745 build it fails on the first assertion), the stale `RunOutput.ts:386` skip comment is gone, the changeset is rewritten, and the fixed settle windows were replaced. On the polling point I kept `vi.waitFor` but wrapped it: most of these assertions are upper bounds ("exactly one cached copy per chunk"), and `waitFor` alone resolves the moment the condition first holds, so a duplicate landing right after would slip through — it waits for the condition and then holds a short window and re-asserts.

Still on me here: the branch went `CONFLICTING` when #20184 merged (both touch `workflows.ts` and `workflows.test.ts`), and finding 2's completeness assertion. I'll take care of both — noting that the assertion goes green with #20521 and red without it, which is the ordering point above.

abhiaiyer91:
Closing in favor of factory generated PR

PR #20194, opened 2026-07-24
fix(core): clean up background tasks on shutdown

### LINKED ISSUE #20193: [BUG] Mastra shutdown does not release background task lifecycle resources
## Describe the Bug

`Mastra.shutdown()` does not shut down its `BackgroundTaskManager`. The manager can retain pub/sub subscriptions, cleanup timers, task contexts, static executor closures, and active `AbortController` instances after the rest of Mastra has shut down.

The manager is initialized asynchronously, so shutdown can also race initialization and leave subscriptions installed after teardown. On durable brokers, a shutting-down worker may acknowledge a dispatch that it deliberately skipped, preventing redelivery to another worker.

Shutdown handling must preserve the existing restart recovery behavior: retryable tasks left at `running` should remain recoverable on the next process rather than being changed to a terminal state.

## Steps To Reproduce

1. Create a `Mastra` instance with storage and `backgroundTasks.enabled: true`.
2. Start a background task whose executor remains active.
3. Call `await mastra.shutdown()`.
4. Observe that the executor's abort signal is not aborted and the manager's `background-tasks` and `background-tasks-result` subscriptions remain registered.
5. Race shutdown against a delayed `pubsub.subscribe()` and observe that initialization can install a listener after teardown.
6. With a durable grouped subscription, deliver a dispatch after shutdown starts and observe that it is acknowledged even though execution is skipped.

The deterministic regression tests are available in this branch:

https://github.com/Kkartik14/mastra/blob/fix/background-task-shutdown-lifecycle/packages/core/src/background-tasks/manager-lifecycle.test.ts

The end-to-end Mastra shutdown reproduction is here:

https://github.com/Kkartik14/mastra/blob/fix/background-task-shutdown-lifecycle/packages/core/src/mastra/shutdown.test.ts

Run:

```bash
cd packages/core
node_modules/.bin/vitest run src/background-tasks/manager-lifecycle.test.ts src/mastra/shutdown.test.ts --pool=forks --maxWorkers=1
```

## Expected Behavior

`Mastra.shutdown()` should release manager-owned timers, subscriptions, contexts, and global hooks before storage closes. It should abort active local executors, use bounded waits for remote cleanup, preserve retryable tasks for restart recovery, log cancellation failures, and leave interrupted durable dispatches unacknowledged for redelivery.

## Environment Information

```text
Mastra commit: 75f843d09f
Node.js: v26.4.0
pnpm: 11.13.1
OS: macOS 26.3 (25D2125), arm64
```

## Verification

Kkartik14:
## Description

Fixes background-task lifecycle leaks during `Mastra.shutdown()`. Shutdown now unregisters process-global hooks, aborts active local executors, releases manager-owned subscriptions and timers before storage closes, and bounds remote cleanup operations so teardown cannot wait indefinitely.

Retryable running tasks remain in a recoverable state for the next process, while non-retryable tasks retain terminal cancellation behavior. Dispatches interrupted by shutdown are left unacknowledged so durable brokers can redeliver them to a surviving worker. Standalone background-task workers also create a fresh manager on a stop/start cycle.

## Related issue(s)

Fixes #20193

## Type of change

- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [x] Test update

## Validation

- Focused shutdown and worker lifecycle tests pass.
- Related background-task tests pass.
- `pnpm --filter ./packages/core check` passes.
- ESLint and formatting checks pass.
- `pnpm build:core` passes.
- The full core suite completed with 12,253 passing tests. Five unrelated tests in `evented-workflow.test.ts` hit their shared `beforeEach` hook timeout under full-suite load; all five passed together when rerun in isolation.

## Checklist

- [x] I have made corresponding changes to the documentation (not applicable to the public API; a changeset is included)

## ELI5

When Mastra shuts down, it now stops and cleans up background tasks in the correct order. Retryable tasks remain available for recovery, and interrupted messages can be delivered again.

## Summary

- Fixed `BackgroundTaskManager` shutdown leaks and initialization races.
- Added bounded waits for remote cleanup operations.
- Aborted active executors and task `AbortController`s during shutdown.
- Released timers, subscriptions, task contexts, and executor references before storage closes.
- Preserved retryable running tasks for recovery after restart.
- Left interrupted durable dispatches unacknowledged for redelivery.
- Prevented new work and unsafe acknowledgments during shutdown.
- Updated `Mastra.shutdown()` to remove scorer hooks and stop the background-task manager earlier.
- Updated workflows to register abort controllers safely and preserve task state during shutdown.
- Updated `BackgroundTaskWorker` to recreate owned managers after `stop()` and clear manager state after failures.
- Added lifecycle, shutdown, and worker restart tests.


abhiaiyer91 (dismissed):
**Verdict: request changes**

The shutdown-lifecycle work itself is well constructed and well tested — the only blocking issue is that this branch no longer merges into `main`, and the conflict lands directly on the lines this PR exists to change, so it needs a semantic resolution plus a re-run rather than a textual one.

## Findings

### 1. Blocking — merge conflict in `packages/core/src/background-tasks/manager.ts`, overlapping this PR's own change

`mergeable: CONFLICTING` / `mergeStateStatus: DIRTY` at head `92ae903`. A dry-run merge against `origin/main` conflicts in exactly one file, `packages/core/src/background-tasks/manager.ts`, in two hunks — and both are the hunks this PR authored:

- **`#doInit` worker callback (around lines 114–133).** `main` has since introduced producer/worker mode (`const isProducerOnly = this.config.mode === 'producer'`) and now defines `workerCallback` **inside** an `if (!isProducerOnly)` block. This PR rewrote the same callback to add the `if (this.shuttingDown) return;` guard and the `handled` boolean that suppresses `ack()`. Neither side wins: after the merge the callback must live inside the mode branch **and** keep the shutdown guard.
- **`TOPIC_DISPATCH` subscribe (around lines 187–195).** `main` moved the dispatch subscribe under `!isProducerOnly`; this PR added a post-subscribe `if (this.shuttingDown) { await this.#releaseLateInitSubscriptions([[TOPIC_DISPATCH, this.workerCallback]]); return; }` immediately after it. That early-return block has to move inside the mode branch too.

One thing worth checking while resolving, because a naive resolution gets it wrong: the second late-init release (the one after `subscribe(TOPIC_RESULT, ...)`, lines 197–203) passes `[TOPIC_DISPATCH, this.workerCallback]` unconditionally. In producer mode after the merge, `this.workerCallback` is `undefined` and `TOPIC_DISPATCH` was never subscribed, so that entry needs the same `if (this.workerCallback)` guard `#shutdown()` already uses at line 821.

Because this file cannot be merged as-is, my verification below is **head branch only — not verified against the current base**. Please resolve, re-run the lifecycle suites, and push; I'll re-review the merged state.

### 2. Non-blocking — a shutdown-interrupted dispatch consumes one retry from the budget

`handleDispatch` writes `retryCount: deliveryAttempt - 1` (existing line, unchanged here) *before* the new `if (this.shuttingDown) return false;` checks can fire. A dispatch interrupted after that write is left unacked and redelivered with `deliveryAttempt + 1`, so the surviving worker records `retryCount = deliveryAttempt` even though the interrupted process never executed the tool. Since `workflow.ts:264` gates retries on `task.retryCount < task.maxRetries`, each shutdown that interrupts a task in this window costs it one retry. Not a correctness break — redelivery genuinely is a delivery attempt — but flagging it in case the intended semantics were "shutdown is free."

### Correctness / design (no changes requested)

- Single-flight `shutdown()` via cached `shutdownPromise` (line 771) is correct and the promise is never cleared, so re-entrant calls join rather than restart. `allows shutdown before initialization` covers the pre-init path.
- `registerActiveAbortController()` returning `false` and aborting immediately once `shuttingDown` is set closes the register-after-teardown race atomically, and `workflow.ts` correctly turns that into a `cancelled` outcome instead of a failure.
- Leaving retryable `running` tasks in storage while terminally cancelling `maxRetries === 0` tasks (`#cancelTaskForShutdown`, line 850) matches the recovery contract `recoverStaleTasks` already implements.
- `BackgroundTaskWorker` discarding the owned manager in `stop()`'s `finally` and rebuilding it in `start()` is the right call given the manager's shutdown is terminal; the shared-manager path correctly stays a no-op.
- Ordering in `Mastra.shutdown()` — `__unregisterHooks()` then manager shutdown, both before `stopWorkers()`/`storage.close()` — is what the issue asked for, and `__unregisterHooks()` is already idempotent.

### Scope

Focused. Every file traces to the shutdown lifecycle; the `workflow.ts` and `background-task-worker.ts` edits are consequences of the manager contract change, not drive-by work. Changeset present and accurate (`patch` on `@mastra/core`).

## Verification

All commands run with `env -u GH_TOKEN -u GITHUB_TOKEN`. Pre-execution inspection of the diff found no changes to install scripts, test setup, CI workflows, or lockfiles, so execution was safe.

| Command | Result |
| --- | --- |
| `pnpm install --frozen-lockfile` | ok |
| `pnpm turbo build --filter ./packages/core` | 14/14 tasks successful |
| `vitest run src/background-tasks/manager-lifecycle.test.ts src/mastra/shutdown.test.ts src/worker/worker.test.ts` | **24 passed**, 0 failed |
| `vitest run src/background-tasks src/worker` | **95 passed**, 1 skipped, 7 files |
| `pnpm --filter ./packages/core check` (`tsc --noEmit`) | clean |
| `git merge --no-commit --no-ff origin/main` | **CONFLICT** in `background-tasks/manager.ts` (aborted after inspection) |

Repo CI on this PR runs only Vercel/security/labeler checks — the two Vercel failures are `Authorization required to deploy`, unrelated to this change. No build or unit-test CI check exists here, so the runs above are the verification of record. They reflect the head branch before the conflicting base moved under it.

## Disposition of existing review signal

- **CodeRabbit — "worst-case cumulative shutdown time unbounded (~20s across four sequential steps)": refuted.** The budget is a single deadline, not per-step. `shutdown()` sets `this.shutdownDeadline = Date.now() + SHUTDOWN_GRACE_PERIOD_MS` once (line 773), and `#remainingShutdownBudgetMs()` (lines 909–914) returns `shutdownDeadline - Date.now()`, so every `#waitForShutdownStep` call races against the *remaining* slice of one 5s window, and a step that finds the budget spent returns immediately. `manager-lifecycle.test.ts:147` (`bounds the whole shutdown sequence with a single grace period`) pins this: with init stuck in a never-resolving subscribe *and* cancellation stubbed to never settle, shutdown is unsettled at 4,999ms and resolved at 5,000ms. The containerized-SIGKILL concern therefore does not apply.
- **CodeRabbit — "check the abort signal on all shutdown code paths": addressed.** `handleDispatch` and `handleResume` re-check `shuttingDown` after each await boundary, and `#shutdown()` runs a second `#abortActiveControllers()` sweep (line 841) for attempts that were mid-await when teardown began.
- **CodeRabbit — "brief window where the controller is deregistered but the loop has not exited": refuted as a defect.** Deregistration is idempotent and a deregistered controller has already been aborted, so the window has no observable effect.
- **CodeRabbit pre-merge checks** (linked issues, out-of-scope changes, title, description): all passed; consistent with my own scope read.
- No human reviews and no unresolved review threads on this PR.

## Requested changes

1. Merge `origin/main` into `fix/background-task-shutdown-lifecycle` and resolve the `background-tasks/manager.ts` conflict so that the producer/worker mode split **and** the shutdown semantics both survive: keep `workerCallback` (with its `shuttingDown` guard and conditional `ack`) inside `if (!isProducerOnly)`, and move the post-`subscribe(TOPIC_DISPATCH)` late-init release inside that same branch.
2. Guard the `TOPIC_DISPATCH` entry in the late-init release list after `subscribe(TOPIC_RESULT, ...)` on `this.workerCallback` being defined, so producer-mode teardown does not try to unsubscribe a callback it never registered.
3. Re-run `vitest run src/background-tasks src/worker` and `pnpm --filter ./packages/core check` on the merged result and confirm they still pass — the merged file is materially different from what I tested.

## Assumptions

- Treated the `retryCount` interaction in finding 2 as intentional-enough to be non-blocking, since `deliveryAttempt`-derived `retryCount` is pre-existing behavior this PR did not introduce.
- Treated the two Vercel deploy failures as environmental (`Authorization required to deploy`) rather than caused by this change, since the diff touches no docs or playground code.

## Open questions

None.

abhiaiyer91 (approved):
Conflict with main's producer/worker mode split (#19652) resolved semantically: shutdown-aware workerCallback (shuttingDown early-return, conditional ack) kept inside the !isProducerOnly branch, late-init TOPIC_DISPATCH release moved inside that branch, and the post-TOPIC_RESULT release list only includes the dispatch subscription when a workerCallback exists (producer mode has none). Verified on the merged result: background-tasks + worker + shutdown suites 99 passed / 1 skipped, core tsc clean, oxlint/oxfmt clean.

PR #20214, opened 2026-07-26
fix(ci): unblock the softwarefactory E2E job

damien-schneider:
Fixes the Software Factory E2E job, which failed during setup in all 18 runs and never reached its four assertions.

## Changes

- Build `mastracode/web` dependencies before the raw CLI build.
- Restore `sync-template.mjs --tag` for local-registry snapshots.
- Run the dev smoke test without auth, probe a real API route, and include the last probe outcomes on timeout.
- Match the exact `Workflows` sidebar link in the kitchen-sink auth check.

The release workflow does not pass `--tag`, so published template version resolution is unchanged.

## Verification

- All 15 `link:` dependencies resolve from `softwarefactory-e2e-test`.
- The generated project typechecks and builds.
- The focused authenticated-navigation Playwright check passes.
- The Software Factory E2E reaches and passes all four assertions.

## Known limitations

- The boot probe deliberately sets `MASTRACODE_AUTH_DISABLED=1`. It validates scaffold UI/API boot without external platform auth; the default auth-enabled bootstrap is not covered by this test.
- The final kitchen-sink check passed, but its observational-memory reload test failed once and passed on retry. A focused local run reproduced a post-reload UI gap: the API returned persisted extraction data, but the marker did not render it. This PR does not touch that path; it needs separate investigation.
- The local registry prints a non-blocking `@mastra/memory` peer-version warning. That needs a separate registry fix.

## ELI5

This PR fixes the Software Factory end-to-end test so it can reliably start the app and detect when it’s really ready. It also restores an important “pin by tag” feature for local templates and adjusts one UI assertion to click the correct “Workflows” sidebar link.

## What changed

- **More reliable setup/build**
  - In `e2e-tests/softwarefactory/softwarefactory.test.ts`, the `beforeAll` now runs a prerequisite `pnpm run prebuild` for `mastracode/web` **before** building the CLI and factory packages.

- **Restored template snapshot pinning**
  - In `mastracode/mastra-factory/scripts/sync-template.mjs`, `sync-template.mjs` now supports `--tag <dist-tag>` (validated as non-empty).
  - Linked package versions are resolved from `dist-tags.<pinTag>`.
  - `@mastra/memory` is also resolved from the provided `pinTag` (otherwise it falls back to `latest`).

- **Better dev-server readiness checks**
  - The dev-server smoke test is started **without authentication** via `MASTRACODE_AUTH_DISABLED: '1'`.
  - Readiness probing now:
    - Tracks per-endpoint probe outcomes in a `lastProbe` map.
    - Uses a **10s** `AbortSignal.timeout` for each fetch attempt.
    - Probes the UI route (`/`) and a real API route (`/api/tools`).
    - If readiness fails (or the dev process exits), the error message includes whether the process exited plus a joined summary of collected probe diagnostics.
  - After readiness, it still probes `/web/config/providers`, and failure diagnostics now include the `lastProbe` summary; it asserts the endpoint returns **HTTP 200**.

- **Fixed sidebar link assertion**
  - In `packages/playground/e2e/tests/auth/login-flow.spec.ts`, the “authenticated user sees main application content” check now matches the **exact** “Workflows” sidebar link string (instead of a case-insensitive regex).

## Verification

- Dependency resolution works for local snapshots (including tag-pinned resolution).
- Generated projects **typecheck and build** successfully.
- Authenticated navigation succeeds.
- All **four Software Factory E2E** assertions pass.

## Known limitations

- The boot probe runs with authentication disabled.
- A separate observational-memory reload issue remains.
- A non-blocking local-registry peer-version warning remains.


wardpeet:
Why do we need pintag?

damien-schneider:
> Why do we need pintag?

the e2e registry only has changeset snapshots `0.0.0-<tag>-<timestamp>`, so the normal latest/alpha matching never finds them and on the shared CI registry latest can belong to another suite

abhiaiyer91 (approved):

